Issue #816, opened 2025-03-07
Add a dora-llama-cpp-python node


haixuanTao:
In order to support more model without having to build additional nodes every time for each model. It could be great to have a `dora-llama-cpp-python` that is just https://github.com/dora-rs/dora/blob/main/node-hub/dora-qwen without the additional backend and with the model made configurable based on environment variables. 

Ideally, this node would also support images, and historical data.

haixuanTao:
An example of a working pyproject.toml

```toml
[project]
name = "dora-qwen"
version = "0.3.10"
authors = [{ name = "Your Name", email = "email@email.com" }]
description = "dora-qwen"
license = { text = "MIT" }
readme = "README.md"
requires-python = ">=3.9"

dependencies = [
    "dora-rs >= 0.3.9",
    "torch == 2.4.0",
    "torchvision >= 0.19",
    "torchaudio >= 2.1.0",
    "opencv-python >= 4.1.1",
    "modelscope >= 1.18.1",
    "accelerate>=1.3.0",
    "transformers",
    "mlx-lm>=0.21.1; sys_platform == 'darwin'",
    "llama-cpp-python",
]

[tool.uv.sources]
llama-cpp-python = [
    { index = "llama_cpp_python_metal", marker = "sys_platform == 'darwin'" },
    { index = "llama_cpp_python_cu121", marker = "sys_platform == 'linux'" },
]

[dependency-groups]
dev = ["pytest >=8.1.1", "ruff >=0.9.1"]

[[tool.uv.index]]
name = "llama_cpp_python_cu121"
url = "https://abetlen.github.io/llama-cpp-python/whl/cu121"
explicit = true

[[tool.uv.index]]
name = "llama_cpp_python_metal"
url = "https://abetlen.github.io/llama-cpp-python/whl/metal"
explicit = true

[project.scripts]
dora-qwen = "dora_qwen.main:main"
```

Note that there is different index based on the backend

ShashwatPatil:
please assign me

haixuanTao:
Done :)

ShashwatPatil:
I have made the pull request #850 ,
Please review it @haixuanTao

haixuanTao:
Closed as of #850

PR #850, opened 2025-03-11
Added the dora-llama-cpp-python node 

ShashwatPatil:
This pull request introduces a new Dora node, `dora-llama-cpp-python`, which provides access to LLaMA-based models using either llama.cpp or Hugging Face backends for text generation.


ShashwatPatil:
@haixuanTao can you please review the code for issue #816

ShashwatPatil:
added model change functionality from yml to the dora-llama-cpp-python

haixuanTao:
Can you look at: https://github.com/dora-rs/dora/issues/816#issuecomment-2706830187

One of the key issue is that you need to select the right source for llama cpp python to make sure that you are using the right acceleration.

Also can we make llama_cpp_python independent from transformers.

ShashwatPatil:
HI I have updated the node as required,
@haixuanTao please take a look, also I was travelling for couple of days so was not able to work on this issue 😅

node-hub/dora-llama-cpp-python/pyproject.toml
```diff
@@ -16,9 +16,26 @@ dependencies = [
     "modelscope >= 1.18.1",
     "accelerate >= 1.3.0",
     "transformers",
```
haixuanTao, line 18:
```suggestion
```

Could we remove unused dependencies as this might be conflicting with other projects?

ShashwatPatil: ohk will do that, I forgot to remove them

node-hub/dora-llama-cpp-python/dora_llama_cpp_python/main.py
```diff
@@ -0,0 +1,68 @@
...
+    model_path = Path(MODEL_PATH)
+    if not model_path.exists():
+        raise FileNotFoundError(
+            f"Model file not found at {MODEL_PATH}. "
+            "Download it using: wget -O models/llama-2-7b-chat.Q4_K_M.gguf "
+            "https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF/resolve/main/llama-2-7b-chat.Q4_K_M.gguf"
+        )
+
+    llm = Llama(
+        model_path=str(model_path),
```
haixuanTao, line 31: Would it be possible to use a remote model, by pulling from Huggingface hub and then selecting with a filename selector, like:

```python
    llm = Llama.from_pretrained(
        repo_id="Qwen/Qwen2.5-0.5B-Instruct-GGUF",
        filename="*fp16.gguf",
        verbose=False,
    )
```

haixuanTao:
Looks great!

Just one small change which is to make sure that we can both select a model from a local path as well as from a remote huggingface id would be great! 

Thanks in advance!

ShashwatPatil:
updated it to be compatible with offline model path as well as pulling model from huggingface

node-hub/dora-llama-cpp-python/dora_llama_cpp_python/main.py
```diff
@@ -0,0 +1,90 @@
...
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
+# Environment variables for model configuration
+SYSTEM_PROMPT = os.getenv(
+    "SYSTEM_PROMPT",
+    "You're a very succinct AI assistant with short answers.",
+)
+MODEL_LOCAL_PATH = os.getenv("MODEL_LOCAL_PATH", "")  # Local model path takes precedence
+MODEL_NAME = os.getenv("MODEL_NAME", "TheBloke/Llama-2-7B-Chat-GGUF")  # HF repo as fallback
```
haixuanTao, line 15: Sorry for the additional change but can we name this, mod
```suggestion
MODEL_NAME_OR_PATH = os.getenv("MODEL_NAME_OR_PATH", "TheBloke/Llama-2-7B-Chat-GGUF") 
```

This way it Is easier to use one or the other. This name comes from transformer naming within `from_pretrained`

ShashwatPatil: ohk 👍️

Issue #852, opened 2025-03-12
Improve IntoArrow within our arrow-convert crate


haixuanTao:
Currently we implement automatic rust conversion from simple types to arrow but we still miss a lot of types.

We could greatly benefit from having more automatic into arrow and from arrow conversion in dora-rs.

The file to modify is: https://github.com/dora-rs/dora/blob/c6ce45785fa1b130030ec233eebb9d463c5b8664/libraries/arrow-convert/src/into_impls.rs
and: https://github.com/dora-rs/dora/blob/52045f3af84924e5a1cbc3283ade98a962514089/libraries/arrow-convert/src/from_impls.rs

This should be a good first issue for rust dev as this is fully contained within a simple crate.

Tushar4059x:
@haixuanTao What are the highest priority missing types we should implement first?

haixuanTao:
- Borrowed list
- string
- boolean
- temporal

We can take inspiration from: https://docs.rs/arrow_convert/latest/arrow_convert/

I wonder if we can even import it and use it as is!

Shar-jeel-Sajid:
I have read the [https://docs.rs/arrow_convert/latest/arrow_convert/](https://docs.rs/arrow_convert/latest/arrow_convert/) and i think we will be able to import and use it as it is. It has default support for

1. Numeric types
2. Bools
3. String
4. Temporal
and also allows to do other conversions. 
Should i try to import it in dora?

Shar-jeel-Sajid:
Also Do we have test cases to test the round trip conversion from rust to arrow and arrow to rust?

Shar-jeel-Sajid:
I read about [https://docs.rs/arrow_convert/latest/arrow_convert/](https://docs.rs/arrow_convert/latest/arrow_convert/). I think we can import and integrate it in dora. Should i use this crate?

haixuanTao:
We can! If we can make it not break the current API

Adarshgupta1127:
@dora-bot assign this issue to me

haixuanTao:
Close as we implement most of them as of #864

PR #864, opened 2025-03-15
Adding in some missing types and test cases within arrow convert crate

Ignavar:


libraries/arrow-convert/src/from_impls.rs
```diff
@@ -163,6 +167,237 @@ impl<'a> TryFrom<&'a ArrowData> for Vec<u8> {
...
+    fn try_from(value: &'a ArrowData) -> Result<Self, Self::Error> {
+        value.try_into().map(|slice: &'a [f64]| slice.to_vec())
+    }
+}
+
+impl<'a> TryFrom<&'a ArrowData> for String {
+    type Error = eyre::Report;
+    fn try_from(value: &'a ArrowData) -> Result<Self, Self::Error> {
+        let string_array: Vec<String> =
+            <Arc<dyn arrow::array::Array> as Clone>::clone(&value).try_into_collection()?;
```
haixuanTao, line 354: Is this not a very complicated expression for: `value.clone()`?

libraries/arrow-convert/src/from_impls.rs
```diff
@@ -163,6 +167,237 @@ impl<'a> TryFrom<&'a ArrowData> for Vec<u8> {
...
+        return Ok(string_array[0].clone());
```
haixuanTao, line 355: Can we make a check to not panic if the length of vec is actually 0 for whatever reason?

libraries/arrow-convert/src/into_impls.rs
```diff
@@ -145,3 +146,43 @@ impl IntoArrow for () {
...
+impl IntoArrow for NaiveTime {
+    type A = arrow::array::Time64NanosecondArray;
+    fn into_arrow(self) -> Self::A {
+        arrow::array::Time64NanosecondArray::from(vec![
+            arrow::array::temporal_conversions::time_to_time64ns(self),
+        ])
+    }
+}
+
+impl IntoArrow for &String {
```
haixuanTao, line 166: Why not just String? If StringArray owns the data?

libraries/arrow-convert/src/into_impls.rs
```diff
@@ -145,3 +146,43 @@ impl IntoArrow for () {
...
+                    eprintln!("Failed to downcast to StringArray.");
+                    StringArray::from(vec![""]) // Fallback in case of failure
+                }
+            }
+            Err(err) => {
+                eprintln!("Failed to Create String Array: {}", err);
+                StringArray::from(vec![""]) // Safe fallback
+            }
+        }
+    }
```
haixuanTao, line 187: I'm sorry but seems does not seems right. I don't understand why would create a StringArray from String have so many failure case.

haixuanTao (changes requested):
Really cool work! Thanks for contributing. I left some comment that I think would make sense to fix before merging as well as I think Clippy has some fair linting message that you can double check on the Files Changed tab.

libraries/arrow-convert/src/from_impls.rs
Ignavar: I tried using the clone method directly but the compiler would throw an error that the arrow::array::Array type does not implement clone to circumvent that the compiler suggested using this statement instead.

libraries/arrow-convert/src/into_impls.rs
Ignavar: We initially implemented this for `String`, but during testing, it caused errors in multiple files. In some cases, ownership of the `message` was transferred into `StringArray`, but the `message` was later accessed again for printing or other tasks, resulting in a compilation error you can review those on commit a7f509210688df71741c2a845ff53a2eedb06f66

Shar-jeel-Sajid:
@haixuanTao For temporal types i have added conversion for chrono::NaiveDate,NaiveTime and NaiveDateTime. Should i add any other data types for temporal. Also, you mentioned [here](https://github.com/dora-rs/dora/issues/852#issuecomment-2721462571) adding bool type. But it is already present. Or is there something else you meant?

libraries/arrow-convert/src/from_impls.rs
haixuanTao: Ah okay, thanks!

libraries/arrow-convert/src/into_impls.rs
haixuanTao: It seems to me that the CI/CD error your referencing is linked to a message being compiled with the own string instead of the borrowed string. 

Can we skip the String part for now. There actually already is:

```rust
impl IntoArrow for &str {
    type A = StringArray;
    fn into_arrow(self) -> Self::A {
        std::iter::once(Some(self)).collect()
    }
}
```

which is probably the cleanest thing to do for now

haixuanTao:
I think this PR looks very good as of now. Let's remove the string part as I think it is a bit misleading and then I think we can merge it :)

haixuanTao:
I think in a follow up PR. We can try to work on a procedural macro, because it seems to me that a lot of this should be handled by it.

Ignavar:
Okay, I'm currently working on removing the `String` from `to` and also removing the `arrow_convert` crate since it's not being used.  

One more thing—should I add type conversion for the borrowed list since that's missing, or should I handle that in another PR?

haixuanTao:
You mean borrowed slice for string?

Ignavar:
Yes, but I just checked, and that case is already covered in the code, so there's no need to add it.

Cargo.lock
```diff
@@ -751,6 +751,31 @@ dependencies = [
...
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e9bf1db5945a3df72462497856891954455351038df26e2280f2443b87292fcf"
+dependencies = [
+ "proc-macro-error2",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.94",
+]
+
```
haixuanTao, line 778: This is out of date, I'm going to replace it.

haixuanTao (approved):

haixuanTao:
Let's avoid to merge main otherwise it creates duplicated commit

Ignavar:
Okay, I'll make sure to avoid that in future PRs.

PR #877, opened 2025-03-17
Using Procedural macros in Arrow Conversion

Shar-jeel-Sajid:


libraries/arrow-convert/src/from_impls.rs
```diff
@@ -36,98 +36,88 @@ impl TryFrom<&ArrowData> for bool {
...
+    };
+}
+
+impl_try_from_arrow_data!(
+    u8 => UInt8Type,
+    u16 => UInt16Type,
+    u32 => UInt32Type,
+    u64 => UInt64Type,
+    i8 => Int8Type,
+    i16 =>Int16Type,
```
haixuanTao, line 63: There's a typo here I think?

libraries/arrow-convert/src/from_impls.rs
```diff
@@ -36,98 +36,88 @@ impl TryFrom<&ArrowData> for bool {
...
+    i32 => Int32Type,
+    i64 => Int64Type,
+    f32 => Float32Type,
+    f64 => Float64Type
+);
+
```
haixuanTao, line 102: Can we not implement both data and slice and vec at the same time?

I feel like defining this individually leaves room for typos

haixuanTao (changes requested):
Looks awesome!

Just left one small comment and then I think it is ready to be merged.

libraries/arrow-convert/src/from_impls.rs
Shar-jeel-Sajid: I guess we can implement both together. Will try that. Also we can add float16 but i did not find anything about 128 types on [arrow crate documentation](https://docs.rs/arrow/latest/arrow/). Any idea how to add them.

libraries/arrow-convert/src/into_impls.rs
```diff
@@ -10,135 +13,65 @@ impl IntoArrow for bool {
...
-        std::iter::once(self).collect()
-    }
-}
-impl IntoArrow for f64 {
-    type A = PrimitiveArray<arrow::datatypes::Float64Type>;
-    fn into_arrow(self) -> Self::A {
-        std::iter::once(self).collect()
-    }
+
+impl_into_arrow_for_uint!(
```
haixuanTao, line 29: I think naming might be an issue here

libraries/arrow-convert/src/from_impls.rs
haixuanTao: Ah sorry! for 128. Forgot they are not supported!

Shar-jeel-Sajid: I tried adding f16 datatype but as you can see the error
`error[E0658]: the type `f16` is unstable
  --> libraries/arrow-convert/src/into_impls.rs:39:5
   |
39 |     f16 => Float16Type,
   |     ^^^
   |
   = note: see issue #116909 <https://github.com/rust-lang/rust/issues/116909> for more information`
   We will either need to enable the unstable feature flag or use an external library such as `half`.

libraries/arrow-convert/src/from_impls.rs
```diff
@@ -36,97 +36,64 @@ impl TryFrom<&ArrowData> for bool {
...
+    };
+}
+
+impl_try_from_arrow_data!(
+    u8 => UInt8Type,
+    u16 => UInt16Type,
+    u32 => UInt32Type,
+    u64 => UInt64Type,
+    i8 => Int8Type,
+    i16 =>Int16Type,
```
haixuanTao, line 91:
```suggestion
    i16 => Int16Type,
```

haixuanTao (approved):
Looks great!

Small typo but other than that, looks awesome!

haixuanTao:
Do you know how you can follow up on this PR?
I think also something like supporting: https://crates.io/crates/half for f16 could be a nice addition

Issue #878, opened 2025-03-17
Tracking Issues for better code quality flag in Python


haixuanTao:
Follow up flag that can easily be added:

- [ ] Use 'I' for isort imports
- [ ] Use 'RET' for flake8 Return
- [ ] Use 'RSE' for flake8 Raise
- [ ] Use "E" for pycodestyle. Line too long warning. Requires an automatic comment line reducers
- [ ] Use "PERF" for performance
- [ ] Use "NPY" for numpy related errors
- [ ] Use "N" for better naming

haixuanTao:
Each Letter can be assigned individually.

7SOMAY:
@dora-bot, assign me

haixuanTao:
@7SOMAY  Could you say which letter you want to tackle next as to let other people do other ones?

7SOMAY:
Can I work on more than one letters?

7SOMAY:
@haixuanTao, please assign "PERF" to me :)

Ignavar:
@haixuanTao please assign "I" to me

MunishMummadi:
@haixuanTao I would like to work on "NPY"

7SOMAY:
@haixuanTao, please assign me one more : “RSE”

haixuanTao:
@7SOMAY this devil has fixed many of the current flags. 

Verry sorry for the people that wanted to work on this

7SOMAY:
@dora-bot assign me

PR #880, opened 2025-03-17
Added "PERF" flag inside node-hub

7SOMAY:
#878 - Added PERF flag by using a custom python script


haixuanTao:
Screenshot from 2025-03-17 17-21-25

So, there is a minor thing which is that at the end of each file should be an empty line.

Could you make sure that this is still preserved within your PR.

7SOMAY:
> Screenshot from 2025-03-17 17-21-25
> 
> > The red sign means that this file misses an empty line at the end.
> 
> So, there is a minor thing which is that at the end of each file should be an empty line.
> 
> Could you make sure that this is still preserved within your PR.

Sure @haixuanTao, I’ll handle this as well. Provide me a moment!

One thing I want to ask, Can I skip ci/cd for that change?

haixuanTao:
@7SOMAY No worries. Let's keep the CI/CD. I know it's a bit slow. Sorry.

PR #886, opened 2025-03-18
Added f16 type conversion

Shar-jeel-Sajid:
This is a follow up PR on  #877.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2630,6 +2630,7 @@ dependencies = [
  "arrow 54.2.1",
  "chrono",
  "eyre",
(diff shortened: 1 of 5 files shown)


haixuanTao (approved):
Awesome thank!

Float16 and bfloat16 is pretty big within Deep Learning Algorithm

Shar-jeel-Sajid:
So i was seeing the documentation of Arrow and came across 'Decimal128' Types and Arrays. I think we can use them to add conversions for 128 bits types.

haixuanTao:
Okay sure!

PR #897, opened 2025-03-19
feat(c/node): Make dora_next_event thread-safe

### LINKED ISSUE #540: Thread-safety of C node API
We should clearly document, which parts of the C node API are thread-safe and which not. Also, we should try to make more functions thread-safe, if possible.

Example: The `dora_next_event` function casts the given context pointer to a `&mut DoraContext` in Rust, which must be an exclusive reference. So calling the function concurrently with the same context pointer violates this requirement and leads to undefined behavior. However, the underlying `flume` channel supports receiving elements through shared references, so we might be able to make `dora_next_event` thread-safe.

SanjayUG:
# Feat: Make dora_next_event Thread-Safe

## Which issue does this PR close?

- Closes #540

## Rationale for this change

This change allows multiple threads to safely call `dora_next_event` concurrently while maintaining existing safety guarantees. The updates ensure that the `DoraContext` fields are accessed in a thread-safe manner.

## What changes are included in this PR?

- **feat(c/node):** Make `dora_next_event` thread-safe
  - Wrap `DoraContext` fields in `Arc<Mutex<>>` for thread-safe access
  - Update `dora_next_event` to use shared references and mutex locking
  - Add comprehensive thread-safety documentation
  - Make `dora_send_output` thread-safe
  - Add proper error handling for mutex lock failures

## Are these changes tested?

Yes, the changes have been tested to ensure that `dora_next_event` and `dora_send_output` can be called safely from multiple threads without causing race conditions or deadlocks.

## Are there any user-facing changes?

No, there are no user-facing changes as this update is related to internal thread-safety improvements.

## Are there any breaking changes to public APIs?

No, there are no breaking changes to public APIs.

### DIFF
--- a/apis/c/node/node_api.h
+++ b/apis/c/node/node_api.h
@@ -1,22 +1,185 @@
+#ifndef __RUST_DORA_NODE_API_C_WRAPPER__
+#define __RUST_DORA_NODE_API_C_WRAPPER__
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
 #include <stddef.h>
 
+/// Initializes a dora context from the environment variables that were set by
+/// the dora-coordinator.
+///
+/// Returns a pointer to the dora context on success. This pointer can be
+/// used to call dora API functions that expect a `context` argument. Any
+/// other use is prohibited. To free the dora context when it is no longer
+/// needed, use the `free_dora_context` function.
+///
+/// On error, a null pointer is returned.
+///
+/// ## Thread Safety
+///
+/// The returned context pointer is thread-safe and can be used concurrently
+/// from multiple threads. All functions that take this context pointer as
+/// an argument are thread-safe.
 void *init_dora_context_from_env();
+
+/// Frees the given dora context.
+///
+/// ## Safety
+///
+/// Only pointers created through `init_dora_context_from_env` are allowed
+/// as arguments. Each context pointer must be freed exactly once. After
+/// freeing, the pointer must not be used anymore.
+///
+/// ## Thread Safety
+///
+/// This function is not thread-safe. It must not be called concurrently
+/// with any other function that uses the same context pointer.
 void free_dora_context(void *dora_context);
 
+/// Waits for the next incoming event for the node.
+///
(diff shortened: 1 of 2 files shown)


apis/c/node/node_api.h
```diff
@@ -1,22 +1,185 @@
...
 void free_dora_event(void *dora_event);
 
-enum DoraEventType
-{
-    DoraEventType_Stop,
-    DoraEventType_Input,
-    DoraEventType_InputClosed,
-    DoraEventType_Error,
-    DoraEventType_Unknown,
-};
```
phil-opp, line 16: Why was this removed?

phil-opp (commented):
Thanks you, this looks good!

There are some build/CI errors that we need to fix before merging.

heyong4725:
Closing as stale.

This PR addresses part 2 of #540 — making `dora_next_event` actually thread-safe by wrapping `DoraNode` and `EventStream` in `Arc<Mutex<>>` — which I explicitly scoped out of PR #1811 (the docs-only counterpart that just landed in the merge queue) because the runtime change needs separate API-stability review.

Closing because:

- 1+ year stale (last touched 2025-04-02, no movement since)
- Touches the same 2 files (`apis/c/node/node_api.h`, `apis/c/node/src/lib.rs`) that #1811 just modified, so will conflict on rebase
- The `Arc<Mutex<>>` wrapping is an API-shape change for the internal `DoraContext`; reviewers will need to weigh the contention cost on the common single-threaded case against the concurrent-call safety win, and that conversation has not happened

If thread-safety improvements to the C node API are still wanted, the next mover should:

1. Rebase onto post-#1811 main
2. Update the per-function `/* Thread-safe: yes/no */` annotations from "NO" to "yes" for the functions actually made safe by the change
3. Open a fresh focused PR with benchmark numbers on contention cost vs single-threaded baseline

Thanks @SanjayUG for the original implementation effort — the approach (Arc<Mutex>) is reasonable; it just needs current-day rebase + benchmark numbers to merge.

PR #901, opened 2025-03-20
Add support for git repository sources for nodes

phil-opp:
Adds support for fetching nodes from git repositories through new `git`, `branch`, `tag`, and `rev` keys in the dataflow YAML file.


phil-opp:
@haixuanTao The `pip-release` CI job fails on `armv7` and `x86` with a linker error:

```
error: unable to find dynamic system library 'atomic' using strategy 'no_fallback'
```

This seems to be related to https://github.com/rust-cross/cargo-zigbuild/issues/252 , which is used by `maturin`. According to the issue thread, the problem should be fixed already, so I'm not sure why it's still failing. Do you have any idea?

phil-opp:
Looks like `maturin` is still locked to an older version of `mimalloc`: https://github.com/PyO3/maturin/blob/1331f2512fcc3ebada7f6b4639b0df8810b01231/Cargo.lock#L1299 

I'll prepare a PR.

phil-opp:
I tried some things to fix the linker errors, but nothing worked. I'm out of ideas for now, so any help would be appreciated!

haixuanTao:
The underlying issue is that libatomic is not well supported on 32bit platform right?

Like this: https://github.com/rust-lang/cargo/issues/13546 ?

And that we depend on system installation of libatomic to support open-ssl right?

Because git requires ssl to work right ? Because you can only pull git https or ssh both requiring ssl.

In which case it seems to me that if libatomic is not well supported on 32bit platform, we can probably feature flag this feature and only make it available on 64bit machine, how does that sound?

haixuanTao:
So FYI I was able to link some system depency to zig by copying package in  `/lib/x86_64-linux-gnu/` to `$HOME/.rustup/toolchains/1.84-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib` 

This should make it possible to link dynamic library with zig and finish compilation.

haixuanTao:
See: https://github.com/ziglang/zig/issues/16733

phil-opp:
> So FYI I was able to link some system depency to zig by copying package in `/lib/x86_64-linux-gnu/` to `$HOME/.rustup/toolchains/1.84-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib`
> 
> This should make it possible to link dynamic library with zig and finish compilation.

Good idea! I tried to do this with `libatomic.so`, but the one from `lib/x86_64-linux-gnu` [fails with `libatomic.so is incompatible with elf_i386`](https://github.com/dora-rs/dora/actions/runs/14382237807/job/40328691713#step:4:2145):

```
ld.lld: error: /home/runner/.rustup/toolchains/1.86-x86_64-unknown-linux-gnu/lib/rustlib/i686-unknown-linux-gnu/lib/libatomic.so is incompatible with elf_i386
```

I also tried linking/copying `/lib/i386-linux-gnu/libatomic.so.1` instead, but this [fails](https://github.com/dora-rs/dora/actions/runs/14382785922/job/40330566657?pr=901#step:4:2145) again with `unable to find dynamic system library 'atomic' using strategy 'no_fallback'`. 

I don't really know what to do with this information. The directory is in the search path, otherwise copying the x86_64 libatomic wouldn't result in a different error message, but the i386 libatomic seems to be ignored for some reason....

phil-opp:
New error message :tada: 

```
💥 maturin failed
  Caused by: Cannot repair wheel, because required library libatomic.so.1 could not be located.
```

phil-opp:
@haixuanTao I finally got it working on both 32-bit x86 and armv7. Thanks again for the symlink tip!

I also fixed a Windows issue with the build commands of our dataflow examples in [c0ac912](https://github.com/dora-rs/dora/pull/901/commits/c0ac912fc20e67a2ea0b1dcb8f20d908f448906b). 

I rebased away my debug commits, so I think this should be ready for merging as soon as CI passes. Are you fine with merging?

phil-opp:
The `node-hub` failure seems to be unrelated too, as it also happens on `main`. See https://github.com/dora-rs/dora/pull/958#issuecomment-2804760479

phil-opp:
Converting this back to a draft because I want to do some redesigning of the spawning process. Now that we run the build as part of the spawn, we don't want the daemon main loop to wait for it to finish because it would become unresponsive. So the plan is to do it fully asynchronous instead.h

phil-opp:
> So the plan is to do it fully asynchronous instead.

Done, should be ready to merge again.

haixuanTao:
https://github.com/user-attachments/assets/2fdbe151-433d-4b9c-8964-30bfd1c98445

It seems that in this latest branch, when there is an error in the spawning of the dataflow the error has been replaced by:

```bash
failed to get downstream nodes: no running dataflow with ID `019643c9-b243-7d39-955a-d1c555d09ceb`

Location:
    /Users/runner/work/dora/dora/binaries/daemon/src/lib.rs:1353:59

Location:
    /Users/runner/work/dora/dora/binaries/daemon/src/lib.rs:202:46
    
    
    ```

**Comment by haixuanTao** [comment]
Ok so it seems that when using the run command but there is a problem with the build command, dora run will be stuck in place. 

Could it be possible to fail gracefully if the build command fails?

Thanks in advance

File: binaries/coordinator/src/lib.rs [hunk 1]
```diff
@@ -703,6 +705,45 @@ async fn start_inner(
...
+            },
+        }
+
+        let elapsed = start.elapsed();
+        if elapsed > Duration::from_millis(100) {
+            tracing::warn!(
+                "Coordinator took {}ms for handling event: {event_debug}",
+                elapsed.as_millis()
+            );
         }
```
haixuanTao, line 747: Seems like this is debug code no?

binaries/daemon/src/lib.rs
```diff
@@ -379,6 +388,54 @@ impl Daemon {
...
+                    }
+                }
+            }
+
+            let elapsed = start.elapsed();
+            if elapsed > Duration::from_millis(100) {
+                tracing::warn!(
+                    "Daemon took {}ms for handling event: {event_debug}",
+                    elapsed.as_millis()
+                );
```
haixuanTao, line 438: This also look like debugging code?

binaries/daemon/src/lib.rs
```diff
@@ -314,6 +320,9 @@ impl Daemon {
                 tracing::warn!("failed to update HLC with incoming event timestamp: {err}");
             }
 
+            let start = Instant::now();
+            let event_debug = format!("{inner:?}");
+
```
haixuanTao, line 325: What is event_debug for?

libraries/message/src/descriptor.rs
```diff
@@ -70,6 +70,15 @@ pub struct Node {
...
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub path: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub git: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub branch: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tag: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rev: Option<String>,
```
haixuanTao, line 80: What if there is multiple one that does not make sense?

haixuanTao (commented):
I think this feature is really cool and make development fairly simpler on distributed system

haixuanTao:
It seems that the run command start before the build command finished for all node. I think this might be confusing. Could we wait for all node to finish building before starting the dataflow?

haixuanTao:
As we're now running build on each step, it could be good to filter build step to only the one if the local daemon machine id ( and if it exist ) match the dataflow builtin machine id.

Otherwise it's very hard to manage the build step especially on multi OS dataflow.

haixuanTao:
FYI, this will make daemon not work with older version of the coordinator with error:

```bash

Caused by:
    unknown variant `dora-microphone`, expected `Local` or `GitBranch` at line 1 column 434

Location:
    /Users/xaviertao/Documents/work/dora/binaries/daemon/src/coordinator.rs:78:29
```

binaries/coordinator/src/lib.rs
phil-opp: Event handling in the main loop of the coordinator should be quick, otherwise handling of subsequent events will be delayed. This warning indicates that we block the event loop too long, which is something that we should fix.

So I think it might be a good idea to keep the warning, as it makes it easier to notice some potential performance issues in the future. Or what do you think?

binaries/daemon/src/lib.rs
phil-opp: See my comment [above](https://github.com/dora-rs/dora/pull/901#discussion_r2056304232).

binaries/daemon/src/lib.rs
phil-opp: This is also used for the warning mentioned [above](https://github.com/dora-rs/dora/pull/901#discussion_r2056304232). We should probably make this cheaper, though. I just pushed 20a1ade6 to use just the variant name instead of the full debug representation (to avoid performance overhead).

libraries/message/src/descriptor.rs
phil-opp: Only one of them is allowed. This is checked here: https://github.com/dora-rs/dora/blob/20a1ade6add4d22ea7c9324aca87ea96651f9f91/libraries/core/src/descriptor/mod.rs#L168

phil-opp:
> FYI, this will make daemon not work with older version of the coordinator with error:

Yes, I don't think that we can avoid this unfortunately. We should bump the `message` crate to `v0.5.0` when we do the next release. Then the error message should be that the dora versions are not compatible. See https://github.com/dora-rs/dora/blob/20a1ade6add4d22ea7c9324aca87ea96651f9f91/libraries/message/src/daemon_to_coordinator.rs#L31-L44

phil-opp:
> As we're now running build on each step, it could be good to filter build step to only the one if the local daemon machine id ( and if it exist ) match the dataflow builtin machine id.

I don't understand what you mean with this. Each node should be built only once, on the machine where it's going to be spawned.

haixuanTao:
Seems like we can't build a dataflow without a coordinator running anymore:
```
[ERROR]
failed to connect to dora coordinator

Caused by:
    Connection refused (os error 111)

Location:
```

haixuanTao:
I have to say that from a user perspective it is quite confusing to now have to link the path depending on git and not the dataflow working directory, as it means that I have to make couple of changes and revert them if I want to use git.

haixuanTao:
I would prefer if start does not try to build the dataflow. It can happen that we push some changes in between start and then the start stop working.

Can we split build and start/run for now to avoid the risk of breaking start on push?

It's also a bit slow and might be a bit frustrating for some people

haixuanTao:
I'm also concerned about what if there is no internet connection when we try to start the dataflow?

phil-opp:
Thanks for the review!

> Seems like we can't build a dataflow without a coordinator running anymore:

Yes, we decided in our last meeting that `dora build` should behave like `dora start` without the spawning. So now the `build` command is sent to the coordinator and forwarded to the daemons, which then do the actual build. 

We could bring back the old behavior of `dora build` if the dataflow does not define any `_unstable_deploy` keys. What do you think? Do we want to do a git checkout in these cases too?


> the path depending on git and not the dataflow working directory, as it means that I have to make couple of changes and revert them if I want to use git which I feel is going to be the source of many errors.

I think it makes sense to have the git checkout dir as working directory. This way, the `build` and `path` commands are started in your repository directory.

Which directory would you set as working directory instead? If we choose some "root workspace", how would the `build` and `path` commands refer to the checked out git directory? 

You mention a "git workspace", but each repo is checked out in a different folder, so there is no single directory that we can choose as default.


> I would prefer if start does not try to build the dataflow. It can happen that we make some environment fixes or some code fix in between start and then because of the build those changes are not persisted.

This is for local environments only, right? Because for distributed environments you would need to ssh into all machines, so fixing the build command would be easier.



> It's also a bit slow and might be a bit frustrating for some people just looking for starting the process.

That's probably python-related right? Is there a way of making the `pip install` etc faster if there is nothing to do?

> Can we split build and start/run for now to avoid the risk of breaking some in between fixes?

We try to reuse existing git checkouts to avoid cloning the repos again. On the other hand, we want to pull the latest changes to run the latest code. This might lead to conflicts if you run multiple dataflows at the same time that use the same git repository and branch. We don't want to change the source code while a dataflow is running as this might lead to unintended behavior and random bugs. So we do a new clone into a different directory when the second dataflow is started.

If we don't checkout+build on `dora start`, the following can happen:

1. Build dataflow 1 with git branch source
    - this clones the repo
3. Start dataflow 1 
4. Push a new commit to git branch
5. Build dataflow 2 with same git branch source
     - there is already a clone of the repo, but we cannot use it as the other dataflow is still running
     - so we choose a new directory and create a new clone there
6. dataflow 1 finishes
7. Start dataflow 2
      - there are two checkouts of the repository at different commits
      - which one is the correct one?

The problem is that the coordinator doesn't know which start command corresponds to which build command. But the time of the build command is important to pick the correct git hash. 

One potential solution for this could be that we assign a dataflow UUID+name on `dora build` already and require this token on `dora start`. We could make this optional for dataflows without git sources or if there is only one candidate (i.e. there was only one `dora build` before the `dora start`). What do you think about this approach? Do you have another idea?



> I'm also concerned about what if there is no internet connection when we try to start the dataflow? and then build and start command does not work?

If you specify a node source on the internet, there is not much we can do without an internet connection, no?

phil-opp:
Maybe we add some kind of "run-local" feature:

- Make `dora run` ignore the `_unstable_deploy` key -> all nodes are spawned locally
- Add a `local_git_checkout` config option that is used by `dora run`.
    - If a node has a `local_git_checkout` key, ignore the `git`, `branch`, etc keys.
    - Instead, assume that the correct git branch is checked out at `local_git_checkout`
    - Skip the `build` process too if `local_git_checkout` is given
- Add a `dora build --local` command that runs all the build commands on the local machine.

This way, users can quickly try out things, fix bugs, and use manual build commands while developing, using `dora run`. Once ready, they can set up their coordinator and daemons, and use `dora start` to do a distributed checkout+build+run.

haixuanTao:
> Yes, we decided in our last meeting that dora build should behave like dora start without the spawning. So now the build command is sent to the coordinator and forwarded to the daemons, which then do the actual build.

> We could bring back the old behavior of dora build if the dataflow does not define any _unstable_deploy keys. What do you think? Do we want to do a git checkout in these cases too?

I think that we should probably check if the coordinator is present. 

   If yes, then behave by spawning build on each daemon.

If there is no coordinator:

  We can either error out if there is any _unstable_deploy and build all locally if there is no key. How does that sound?

I think we can git checkout yes in that case to.

haixuanTao:
> I think it makes sense to have the git checkout dir as working directory. This way, the build and path commands are started in your repository directory.

> Which directory would you set as working directory instead? If we choose some "root workspace", how would the build and path commands refer to the checked out git directory?

I think that what we can do is to specify the exact git path from the dataflow like:

```yaml
   id: abc
   git: gitrepo
   rev: 123
   path: build/gitrepo/rev/my_node.py
```

This way it is a bit clearer as well as people can find it if needed?

haixuanTao:
> The problem is that the coordinator doesn't know which start command corresponds to which build command. But the time of the build command is important to pick the correct git hash.

I understand. I think that specifying the git hash within the path should make this conflict free, right?

The thing with build+start is that i'm very afraid of a breaking dependency update that breaks the build or start indefinetly.

haixuanTao:
> Maybe we add some kind of "run-local" feature:
> 
>     * Make `dora run` ignore the `_unstable_deploy` key -> all nodes are spawned locally
> 
>     * Add a `local_git_checkout` config option that is used by `dora run`.
>       
>       * If a node has a `local_git_checkout` key, ignore the `git`, `branch`, etc keys.
>       * Instead, assume that the correct git branch is checked out at `local_git_checkout`
>       * Skip the `build` process too if `local_git_checkout` is given
> 
>     * Add a `dora build --local` command that runs all the build commands on the local machine.
> 
> 
> This way, users can quickly try out things, fix bugs, and use manual build commands while developing, using `dora run`. Once ready, they can set up their coordinator and daemons, and use `dora start` to do a distributed checkout+build+run.

I would very much hope that there is no such config.

I truly believe that the best would be to make git cloning explicit and only happen at build time with path linked to the path of the cloned repo.

phil-opp:
> I understand. I think that specifying the git hash within the path should make this conflict free, right?

Yes, but it requires you to modify your dataflow.yml for every commit you push to a branch. Also, we cannot reuse and update an existing clone in that case, so every commit leads to a new git clone and a full build, instead of just an incremental build.

My concern is that this makes the edit-build-try cycle very slow and cumbersome.

phil-opp:
Maybe we can solve the update conflict by doing a branch name to commit hash conversion in the CLI before we send it to the coordinator. We still need some solution to the reuse problem though.

phil-opp:
> The thing with build+start is that i'm very afraid of a breaking dependency update that breaks the build or start indefinetly.
>
> Such as the chrono compatibility issue or just our own ones.

That can be avoided by using lockfiles, which most tools do already. This ensures that all dependencies are built in exactly the same way, even if there are updates in between.

phil-opp:
> I would very much hope that there is no such config.

Could you clarify your issues with the idea? I think that letting `dora run` ignore the deploy sections instead of erroring out would be a useful feature.

I also find myself using cargo's feature to override dependencies often. The idea is that you can specify both a local path and a crates.io version for a workspace dependency. Cargo will then use the path dependency during development, and the crates.io dependency when publishing. Our use case is a bit different of course, but I can imagine that you might want a way to test a dataflow locally during development, before you deploy it in a distributed way.

haixuanTao:
> > I understand. I think that specifying the git hash within the path should make this conflict free, right?
> 
> Yes, but it requires you to modify your dataflow.yml for every commit you push to a branch. Also, we cannot reuse and update an existing clone in that case, so every commit leads to a new git clone and a full build, instead of just an incremental build.
> 
> My concern is that this makes the edit-build-try cycle very slow and cumbersome.

I mean can't updating happen within the build command but not the start command?

> I would probably go in the other direction and choose a separate working directory for each dataflow even if there are no git nodes. Otherwise build commands might interfere with each other. For example, imagine two dataflows that both have a node named sink, but with very different implementations. If they both place a sink executable in the working directory as part of their build command, one will overwrite the other.

I mean build command of two different dataflow already interfere with one another. If you cargo build a node, it will already conflict between dataflows. Any system dependency update will create interference.

I think that users are well aware that building multiple dataflow at the same time is going to create conflict, and this git feature alone does not make our dataflow build-conflict free, from what i can see.

phil-opp:
Regarding the two-step build proposal (i.e. don't build on `dora start`): What do you think about the following approach:

- Instead of letting the daemons do the branch->commit hash mapping, we do it in the CLI.
  - On `dora build`, the CLI will first try to resolve all git branches and tags to a specific commit hash.
  - We will need this hash later, so we need to store it somewhere. I think a good solution could be to create a `dataflow.dora-build-meta.json` file next to the `dataflow.yml`.
  - We also create a unique "build id", which we also store in the meta json file.
  - Then we trigger the build on the coordinator as before, but we additionally pass the hash and the build id.
  - The daemons don't need to care about branch names etc anymore and can just checkout the given hash and then run the build commands. They keep track of the build id to "assign" the directory to the CLI instance.
- On `dora start`, we read the build id from the `dataflow.dora-build-meta.json` and pass it to the coordinators + daemons
  - no git checkout or rebuild is necessary
  - the same dataflow can be started multiple times without rebuilds etc


- On the next `dora build`, we fetch the git repo again and check whether the commit hash changed (e.g. because the branch was pushed)
    - if so, we send the new and old hash to the daemons, which can then reuse and rename the directory of the old commit hash
    - if not, we instruct the daemons to just rebuild, without changing the git source

phil-opp:
As proposed in our meeting, I created a small summary how this feature would look from a user's perspective:

## User View

### Build Commands

- Nodes can specify a `build` command in their YAML definition
- Dora will run the `build` command when `dora build` is invoked
- If the dataflow is distributed across multiple machines, each `build` command will be run the target machine of the corresponding node.
    - i.e. the machine specified under the `deploy` key 
    - this requires a connection to the dora coordinator, so you need to specify the coordinator IP/port for this
    - to run the build commands of all nodes _locally_, you can use `dora build --local`
- If the build command does not specify any `deploy` keys, all build commands will be run locally (i.e. `dora build` behaves like `dora build --local`)


### Git Source

- Nodes can have a git repository as source
    - by default, the default branch is used
    - you can also specify a specific `branch` name
    - alternatively, you can specify a `tag` name or a `rev` key with the commit hash
    - you can only specify one of `branch`, `tag`, and `rev`, otherwise an error will occur
- Dora will automatically clone and checkout the requested branch/tag/commit on `dora build`
    - the `build` command will be run after cloning
    - for distributed dataflows, the clone/checkout will happen on the target machine
- subsequent `dora build` command will automatically fetch the latest changes for nodes
    - not when using `tag` or `rev`, because these are not expected to change
- after fetching changes, the `build` command will be executed again
    - _tip:_ use a build tool that supports incremental builds (e.g. `cargo`) to make this rebuild faster


The **working directory** will be set to the git repository. This means that both the `build` and `path` keys will be run from this folder. This allows you to use relative paths.

### The `dora run` Command

The `dora run` command is a quick and easy way to run a dataflow locally. It will ignore any `deploy` keys in the YAML file, so it will run all dataflows locally, even if they specify remote machines.

The `dora run` command does not interact with any `dora coordinator` or `dora daemon` instances, or with any other parallel `dora run` commands.

Use `dora build --local` or manual build commands to build your nodes. 

### The `dora start` Command

The `dora start` command is used to spawn a dataflow in a pre-existing _dora network_. To create a dora network, spawn a `dora coordinator` and one or multiple `dora daemon` instances.

The `dora start` command does not run any build commands, nor update git dependencies or similar. Use `dora build` for that.


----

Feel free to discuss! 

cc @haixuanTao @heyong4725

haixuanTao:
> The dora run command is a quick and easy way to run a dataflow locally. It will ignore any deploy keys in the YAML file, so it will run all dataflows locally, even if they specify remote machines.

I think that we should raise an error if people try to run a distributed dataflow and say that we currently only support start for running distributed dataflow.

haixuanTao:
> The working directory will be set to the git repository. This means that both the build and path keys will be run from this folder. This allows you to use relative paths.

Either we make a new key like `git_path` or we use a path like `build/git_repo/git_branch/target/release/bin` but I'm strongly against having an if statement on the path working directory as it is going to be 100% impossible to remember this specificiation as well as make debugging so much harder if something goes wrong and we don't know which path failed.

phil-opp:
> I would still strongly be in favor of not breaking the current logic and use: `path: build/git_repo/git_branch/target/release/bin`

What about the `build` command? To me it seems reasonable to use the same working directory for both `build` and `path`, everything else would be confusing. But running all the build commands from the same global folder would be quite cumbersome.

For example, consider the following node definition using the git clone dir as working dir:

```yaml
  - id: rust-node
    git: https://github.com/dora-rs/dora.git
    rev: e31b2a34 # pinned commit, update this when changing the message crate
    build: cargo build -p rust-dataflow-example-node
    path: target/debug/rust-dataflow-example-node
    inputs:
      tick: dora/timer/millis/10
    outputs:
      - random
```

With a global working directory, this would become something like this:

```yaml
- id: rust-node
    git: https://github.com/dora-rs/dora.git
    rev: e31b2a34 # pinned commit, update this when changing the message crate
    build: cargo build --manifest-path build/github.com/dora-rs/dora.git/e31b2a34/Cargo.toml -p rust-dataflow-example-node
    path: build/github.com/dora-rs/dora.git/e31b2a34/target/debug/rust-dataflow-example-node
    inputs:
      tick: dora/timer/millis/10
    outputs:
      - random
```

And every time we update the git commit, we would need to change all the `build` and `target` commands, otherwise we will get an error or silently use an older version.

phil-opp:
> I'm strongly against having an if statement on the path working directory

There is another solution for this: Don't allow relative paths if there is no `git` key.

We already decided in the past to forbid relative paths in multi-machine deployments because there was no clear working directory:

- https://github.com/dora-rs/dora/pull/538

The only reason that relative paths are now allowed again is commit https://github.com/dora-rs/dora/commit/2d7e8f1a96cc2be3f9839232e4df27f7797c3738, which worked around bugs in the path checking code. However, none of this behavior is stabilized yet, so we can easily go back to forcing users to use absolute paths if there is no clear working directory.


## Proposal

- Daemons don't have a default working directory.
- When using `git` key, the working directory is set to the directory where the repo is checked out.
- You can also set a custom `working_dir` per node.
- Relative paths are only permitted if a working dir is specified (either explicitly or implicitly through a `git` key)
- Absolute paths can be always used without restrictions

phil-opp:
> And every time we update the git commit, we would need to change all the `build` and `target` commands, otherwise we will get an error or silently use an older version.

What I meant is that you could easily end up with something like this by accident:

```yaml
- id: rust-node
    git: https://github.com/dora-rs/dora.git
    rev: e31b2a34 # pinned commit, update this when changing the message crate
    build: cargo build --manifest-path build/github.com/dora-rs/dora.git/e31b2a34/Cargo.toml -p rust-dataflow-example-node
    path: build/github.com/dora-rs/dora.git/5119078/target/debug/rust-dataflow-example-node
    inputs:
      tick: dora/timer/millis/10
    outputs:
      - random
```

Here we forgot to update the `path` after changing the `rev`. So we checkout and build the new version, but keep running an older version of the node.

phil-opp:
> [ ] clarify that this change is updating the message format, so it makes the next dora version incompatible with older dora versions

I pushed 9d917bdaa9816106d7f266e4e86e52df73aee0a6 to do the dataflow parsing lazily in the node APIs. So the parsing now only happens when the `dataflow_descriptor()` method is called. Nodes that don't use this method should experience no breakage from dataflow descriptor changes anymore.

phil-opp:
Your links are both 404 errors. I assume you mean that the cause is that we moved the `run_dataflow` functionality from the `daemon` to the `cli` and that we have a dependency on the CLI crate now? I agree that this is likely the cause, but why is it failing on macOS only? And why the different error kinds ('symbol not found' vs 'compiled for wrong macos version')?



> I think that if we comment out this function https://github.com/dora-rs/dora.git/blob/git-source/apis/python/node/Cargo.toml#L27-L28 we can make this PR pass.

Which function do you mean exactly? The `run` function in the Python node API?

phil-opp:
For now I got it to work using https://github.com/dora-rs/dora/pull/901/commits/de949521f6d76d4669c7f183d95730186893a4d0, https://github.com/dora-rs/dora/pull/901/commits/6b5cc7a0b94ea2bbfad626ba98a37dc6504dab12, and https://github.com/dora-rs/dora/pull/901/commits/6071ba71b99a331db0eeb8328b5ec0050eb25674. 

But I would also be ok with removing the `run` function from the Python library. It definetly makes things more complicated. The other node APIs only need to communicate with the daemon, they don't need any complex dependencies such as git2 or openssl as they communicate only with the daemon. But the Python CLI exposes both the runtime and the run function currently, which adds all these dependencies back in.

phil-opp:
Another question: In some CI runs, I get a deadlock on "Waiting to acquire lock for `/home/runner/work/dora/dora/.venv` at `/home/runner/work/dora/dora/.venv/.lock`". For example here: https://github.com/dora-rs/dora/actions/runs/15635434106/job/44049242106?pr=901#step:11:303

Any idea why this happens sometimes? I thought that it might be related to `uv`, but I didn't find anything relevant in their GitHub issue tracker.

haixuanTao:
It is worth mentioning https://github.com/dora-rs/dora/issues/1022

haixuanTao:
Also, is there an easy way to gitignore the `dataflow.dora-session.yaml` as it seems to me that this should not be committed but somehow it seems not obvious to ignore for beginners which will inevitably clutter the repo.

phil-opp:
> Also, is there an easy way to gitignore the `dataflow.dora-session.yaml` as it seems to me that this should not be committed but somehow it seems not obvious to ignore for beginners which will inevitably clutter the repo.

Good point! I pushed e21520d433f301843c411635e7c1f482bcd0927d to add them to a `.gitignore`.

haixuanTao:
I'm unable to build and run rust node. Example:

```bash
cd examples/camera
uv venv
dora build dataflow_rerun.yml  --uv # run: cargo build -p dora-rerun --release
dora run dataflow_rerun.yml  --uv
```

Errored out with:
```


[ERROR]
Dataflow failed:

Node `plot` failed: exited with code 1 with stderr output:
[...]             ^^^^^^^^^
RuntimeError: Could not initialize dora node

Caused by:
   0: failed to deserialize node config
   1: unknown variant `socket_addr`, expected one of `Shmem`, `Tcp`, `UnixDomain` at line 1 column 12

Location:
    apis/rust/node/src/node/mod.rs:82:40

```

haixuanTao:
Ok sorry fixed it it was an issue with my environment.

phil-opp:
> Can we put dora-session.yaml in the out folder?

We can, but there might not be an `out` folder in all cases. The folder is currently created and written by the daemon. So if the demon is running on a different machine, the `out` folder will be created there. We should probably change this at some point.

haixuanTao:
FYI, double checked on logging coloring and looks good to me!

Issue #896, opened 2025-03-19
Undetailed Documentation for beginners


gagan-1307:
The current documentation is complicated to understand and go through the project which requires a lot of hours to get through the codebase and functionality of project.

## Current State
- Basic documentation structure exists
- Some guides and API references are available
- Limited examples and use cases
- Basic installation instructions

### 1. Getting Started Guide
- [ ] Create a step-by-step tutorial for absolute beginners
- [ ] Add a "Quick Start" guide with minimal setup
- [ ] Include prerequisites and system requirements
- [ ] Add troubleshooting section for common installation issues

### 2. API Documentation
- [ ] Expand Python API documentation with more code examples
- [ ] Add detailed Rust API documentation with practical use cases
- [ ] Include performance considerations and best practices
- [ ] Add type hints and return value documentation

### 3. Architecture and Concepts
- [ ] Create detailed architecture diagrams
- [ ] Explain core concepts with visual aids
- [ ] Document the data flow between nodes
- [ ] Add system design principles and patterns

### 4. Examples and Tutorials
- [ ] Add more real-world use cases
- [ ] Create tutorials for common scenarios
- [ ] Include complete end-to-end examples
- [ ] Add performance optimization examples

### 5. Troubleshooting Guide
- [ ] Create a comprehensive troubleshooting section
- [ ] Add common error messages and solutions
- [ ] Include debugging tips and tools
- [ ] Add performance tuning guidelines

haixuanTao:
Absolutely agree that the documentation is not great! 

Please help us out!

gagan-1307:
@haixuanTao should i work on documentation and include the changes.

Hennzau:
> [@haixuanTao](https://github.com/haixuanTao) should i work on documentation and include the changes.

It would be really cool!

vedantvijay:
@dora-bot assign me

Xlqmu:
I think Deep_Wiki can be used to assist in building Dora's documentation.

heyong4725:
Closing as substantially addressed — thanks @gagan-1307 for filing this back in March 2025, and to @vedantvijay and @Neomelt for the follow-up engagement.

## What changed in the 14 months since this was filed

The docs surface has grown from sparse to **52 files in `docs/`** + a 660-line `README.md` + a 100-line `docs/quickstart.md`. Cross-referencing each section of the original wishlist:

| § | Section | Status | Where |
|---|---|---|---|
[...]

PR #904, opened 2025-03-20
docs: improve documentation structure and organization (#896)

gagan-1307:
Have a look at docs/src section and the main readme.md file. The documentation is significantly improved and easy to understand .


gagan-1307:
> I'm sorry but you're using too much `vibe` coding :S

Haha, vibe coding? Guilty as charged! What should I tweak to make it less vibey for you?

Hennzau:
> > I'm sorry but you're using too much `vibe` coding :S
> 
> Haha, vibe coding? Guilty as charged! What should I tweak to make it less vibey for you?

Hi, I think Xavier means that it's too much AI generated. Some files are just some AI "general" information about processing. Indeed the Rust/Python documentation is not correct, lot of things just don't work like that.

However the documentation structure is good! thx for the contribution, just, you should keep the structure and remove the parts you're not sure about. And then we will progressively add correct things to it

docs/src/getting-started.md
Hennzau, line 1: This is cool, prefer other ways to install directly with curl/pip/cargo

Issue #879, opened 2025-03-17
Implement Async next event in Python using pyo3-asyncio


haixuanTao:
Since 0.22 pyo3 can now easily integrate rust async function with: https://pyo3.rs/v0.22.5/ecosystem/async-await

Could be cool to implement it for dora node next to make it simpler to receive asynchronously new event.

Choudhry18:
@dora-bot assign me

haixuanTao:
This has been implemented in #909

Issue #748, opened 2025-01-07
Interactive and realtime visualisation with gradio


haixuanTao:
### Discussed in https://github.com/orgs/dora-rs/discussions/746

<div type='discussions-op-text'>

<sup>Originally posted by **jrabary** January  6, 2025</sup>
Hi all,

I'm trying to add an interactive and realtime visualisation node in my graph. I end up having the following events handlers in the GUI 

```
    def handle_submit(command):
        if command != "":
            instruction_array = pa.array([command])
            node.send_output("instruction", instruction_array)
        return ""

    def handle_camera_streams():
        current_im = im
        while True:
            evt = node.next(timeout=FPS)
            if evt is not None:
                if evt["type"] == "INPUT":
                    if evt["id"] == "observations":
                        obs = decode_dict(evt["value"])
                        current_im = obs["overhead_cam"]

            yield current_im
```

The first one is to handle user action on the GUI and the second on to display a video stream on the GUI. The point is, the two handlers can work separatly but once I put them together I got some errors

```
node.send_output("instruction", instruction_array)
RuntimeError: Already borrowed
```

I suspect dora doesn't like I use the same node object concurrently. 

Any ideas ?</div>

7SOMAY:
Hi @haixuanTao, I believe this issue can be handled using multi-threading in Python.
We can create separate threads for each functionality to process them simultaneously without blocking.

I guess this will work, what do you think?

haixuanTao:
I think that we currently have an issue if we try to call in parallel dora from different thread, but haven´t really investigated deeply with this.

7SOMAY:
> I think that we currently have an issue if we try to call in parallel dora from different thread, but haven´t really investigated deeply with this.

Oh I see, if you want-I can give it a try. Just let me know the reference and a kick start to grasp the exact situation to work on.

7SOMAY:
> Honestly. it's quite an intricate issue as it is deeply rooted within the mixed rust-python side of dora.

Just let me have a quick overview of that, I’ll try to get my hands on that.
At least I’ll learn something. 

But only if you say so, o/w I’ll keep working on my existing issues.

haixuanTao:
Being bluntly honest, I don't even know myself how to fix it! XD 

I'm not even sure where does this come from. I genuinely think that it can take about a week to fix

7SOMAY:
[...]

PR #601, opened 2024-07-21
Delay dropping of `DoraNode` in Python until all event data is freed

phil-opp:
When dropping the `DoraNode`, it waits for the remaining drop tokens. This only works if all the dora events were already dropped before. With the Python GC, this is not guaranteed as some events might still be live on the heap (the user might even use them later). In such cases, we waited until we ran into a timeout, which resulted in very long exit times (see https://github.com/dora-rs/dora/issues/598).

This commit fixes this issue by adding a reference-counted copy of the `DoraNode` and `EventStream` to every event given to Python. This way, we can ensure that the underlying `DoraNode` is only dropped after the last event reference has been freed.

Fixes #598

Issue #598, opened 2024-07-19
Python GIL: Drop token race condition


phil-opp:
PR https://github.com/dora-rs/dora/pull/568 drastically reduced some drop token timeouts to work around a "drop token race condition with the GIL". To reproduce the issue, the following example was posted: https://github.com/dora-rs/dora/pull/568#issuecomment-2205162981

The goal of this issue is to find the root cause of this bug so that we can increase the mentioned timeouts again. The problem with the reduced timeouts is that slower receivers (or receivers with larger queue sizes) might not be able to read the the last outputs of nodes after they exit.

phil-opp:
@haixuanTao I reverted the changes of #568 on my machine and tried the example shown in https://github.com/dora-rs/dora/pull/568#issuecomment-2205162981 and I can reproduce the grace period kills. However, once I increase the grace period to 15 seconds, all nodes stop properly after a few seconds when I do a ctrl-c. Could you try whether this works for you too?

haixuanTao:
I see.

So a quick way to probably reproduce the problem, is to try to delete the node before the 

```python
git clone git@github.com:dora-rs/dora-benchmark.git --branch simple-drop-token-example
cd dora-rs/py-latency
dora up
dora start dataflow_node.yml
# Check the logs of having a warnings
vim node_2.py
# Go to end of file and try Example 2 and notice no warnings.
```

This should work with #568 change reverted.

PR #909, opened 2025-03-21
Adding `recv_async` dora method to retrieve data in python async

haixuanTao:
Make it possible to retrieve data asynchronously within python asyncio.

This is possible due to `experimental-async` from pyo3.

Note that this feature should be considered experimental as long as pyo3 async is marked as experimental.

This feature required to remove events delayedcleanup as it was no longer used.

### DIFF
--- a/examples/python-async/receive_data.py
+++ b/examples/python-async/receive_data.py
@@ -1,0 +1,17 @@
+import asyncio
+
+from dora import Node
+
+
+async def main():
+    node = Node()
+    for _ in range(100):
+        event = await node.recv_async()
+        print(event)
+        # del event
+    print('done!')
+
+
+if __name__ ==  '__main__':
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(main())
--- a/apis/python/node/Cargo.toml
+++ b/apis/python/node/Cargo.toml
@@ -10,9 +10,10 @@ repository.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["tracing", "metrics", "telemetry"]
+default = ["tracing", "metrics", "telemetry", "async"]
 tracing = ["dora-node-api/tracing"]
 metrics = ["dora-node-api/metrics"]
+async = ["pyo3/experimental-async"]
 telemetry = ["dora-runtime/telemetry"]
 
 [dependencies]
--- a/apis/python/node/pyproject.toml
+++ b/apis/python/node/pyproject.toml
@@ -12,13 +12,13 @@ readme = "README.md"
 dependencies = ['pyarrow']
 
 [dependency-groups]
-dev = ["pytest >=8.1.1", "ruff >=0.9.1"]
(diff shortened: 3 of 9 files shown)


haixuanTao:
This should fix #879 #748 but will need to double check on this

haixuanTao:
@phil-opp would love if you can double check this when you have time :)

haixuanTao:
FYI, I had to removed DelayedCleanup on events as it was not send for some reason which is required for async method.

I think it didn't really was necessary anymore as arrow array are not tied to its reference counting.

eladb:
Amazing, thanks!

examples/python-async/receive_data.py
```diff
@@ -0,0 +1,17 @@
+import asyncio
+
+from dora import Node
+
+
+async def main():
+    node = Node()
+    for _ in range(100):
```
eladb, line 8: I am wondering if we can get to something like this:

```py 
async for event in node:
  # ...
```

haixuanTao: Sure thing!

haixuanTao: Ok, so it seems that `__anext__` the async iterator is kind of hard to implement so i'm going to merge this like this for now. And maybe come back on this once the documentation on pyo3 is better.

See: https://github.com/PyO3/pyo3/issues/1632#issuecomment-2747696822

eladb: Of course. This is awesome!

eladb:
@haixuanTao thanks for much for adding this! when can we expect a release?

haixuanTao:
I will try to do one in the coming week!

Sorry for being a bit slow

eladb:
Are you kidding me? LOL. Thanks for all the effort. No rush. We are not blocked by this, but could be nice to clean up some stuff.

haixuanTao:
Ok cool! 

Just want to merge couple of change within the distributed dora as well as support for av1, that might really help me out, in the short term.

3.11 might be a cool release :)

eladb:
👍

Btw in other projects I’ve maintained such as projen and winglang and some
parts of the AWS CDK we managed to get to continuous release which
basically means that every commit to main is released. Not sure if that’s
feasible here in terms of quality gates, but I highly recommend it because
it takes away a lot of the burden in maintaining the project.

phil-opp:
> FYI, I had to removed DelayedCleanup on events as it was not send for some reason which is required for async method.
> 
> I think it didn't really was necessary anymore as arrow array are not tied to its reference counting.

If I remember correctly, we introduced the `DelayedCleanup` to make sure that the node object is not freed by the Python interpreter until all of its events were freed. Otherwise the drop tokens are not ready at that point. See https://github.com/dora-rs/dora/pull/601 for details.

I think we reduced some timeouts in the meantime so this isn't as visible anymore as it was before (see https://github.com/dora-rs/dora/issues/598). However, we should still try to solve the underlying issue.

haixuanTao:
Yeah, but the DelayedCleanup was not working for arrow array as basically arrow array is always schedule to be cleaned before the events stream closed making a race condition.

Issue #894, opened 2025-03-19
Dora should be able to self update


Hennzau:
Like `uv` does, maybe we could extend the CLI so that `dora self update` is updating the CLI to the latest version

What do you think of that? There is a crate for that: `self_update`

haixuanTao:
Hey Enzo, that's a great idea. And maybe for node as well right?

Maybe we could try with something chill like the uv package manager to help us do this in simple instance could be a good start

Hennzau:
Yep we can probably do something for the nodes as well, but for the current workflow, `dora build` should already install any new published version of a node.

I think I will let someone do the `dora self update`, as it should be really easy with https://crates.io/crates/self_update

7SOMAY:
@dora-bot assign me

PR #912, opened 2025-03-22
Add Self Management Commands: Update, Reinstall, Uninstall

7SOMAY:
This PR: #894 contains an implementation of three subcommands under the "self" command in the Dora CLI:

- Update: Updates the Dora CLI to the latest version, with an option to force the update.
- Reinstall: Reinstall the Dora CLI, useful for fixing issues.
- Uninstall: Uninstall the Dora CLI from the system.

These commands enhance user experience by simplifying the management of the Dora CLI installation.


Hennzau:
Thx for the contribution, why are there so many unrelated files modified?

7SOMAY:
> Thx for the contribution, why are there so many unrelated files modified?

I am trying to solve this OpenSSL error, for that created separate files to handle it easily.

Hennzau:
Oh... look at that https://github.com/jaemk/self_update?tab=readme-ov-file#troubleshooting
I don't think we need to use OpenSSL

7SOMAY:
> Oh... look at that https://github.com/jaemk/self_update?tab=readme-ov-file#troubleshooting I don't think we need to use OpenSSL

Added this to Cargo.toml

How you find that so quickly, it overwhelmed me 😢

Hennzau:
Usually when a rust crate needs dependencies, it's detailed in the README file

Hennzau:
So: you should rebase this PR so there is not unrelated files to this PR. It should only modify some files of the CLI crate.

For uninstall and reinstall please use https://crates.io/crates/self-replace instead of cargo commands

7SOMAY:
Perfect, doing that.

PR #913, opened 2025-03-23
implementing self update command

Shar-jeel-Sajid:
Very simple dora self-update command.
Run 
`dora self update --check-only`
to only check for update and Run 
`dora self update` to update to latest release.


Hennzau:
Hi, I think @7SOMAY was working on it.

But thx for the contribution. you should make a self command and an update subcommand, so we can add later an **uninstall, reinstall** subcommand.

Also, see https://github.com/jaemk/self_update#troubleshooting to disable OpenSSL and prefer RustLS

binaries/cli/src/lib.rs
```diff
@@ -241,8 +241,16 @@ enum Command {
         quiet: bool,
     },
 
-    SelfUpdate {
+    Self_ {
         /// Only check for updates without installing
```
Hennzau, line 245: You should remove this comment and put it on line 254

Hennzau (changes requested):
LGTM

Hennzau (approved):
Perfect!

Hennzau:
@haixuanTao I think we can merge it!

haixuanTao:
When I try to run it, I get:

```bash
Extracting archive... Failed to update: ArchiveNotEnabled: Archive extension 'zip' not supported, please enable 'archive-zip' feature!
```

7SOMAY:
But i guess I already implemented it and was working on self_replace scenario for uninstall and reinstall, that's why it took some time. Will release for same by tonight!

binaries/cli/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tabwriter = "1.4.0"
 log = { version = "0.4.21", features = ["serde"] }
 colored = "2.1.0"
 env_logger = "0.11.3"
+self_update = { version = "0.27.0", features = ["rustls"], default-features = false }
```
Hennzau, line 53: This should be:

```rust
self_update = { version = "0.27.0", features = ["rustls", "archive-zip"], default-features = false }
```

Hennzau: If we plan to release `tar` files in the future we will have to update this

Hennzau:
I still have an error:

```bash
Checking for updates...
Checking target-arch... x86_64-unknown-linux-gnu
Checking current version... v0.3.9
Checking latest released version... v0.3.10
New release found! v0.3.9 --> v0.3.10
New release is compatible

dora release status:
  * Current exe: "/home/enzo/Downloads/dora-self_update/target/debug/dora"
  * New exe release: "dora-v0.3.10-x86_64-unknown-linux-gnu.zip"
  * New exe download url: "https://api.github.com/repos/dora-rs/dora/releases/assets/234604429"

The new release will be downloaded/extracted and the existing binary will be replaced.
Do you want to continue? [Y/n] Y
Downloading...
[00:00:01] [========================================] 15.98MB/15.98MB (0s) Done
Extracting archive... Failed to update: ZipError: unsupported Zip archive
```

Hennzau:
The error comes from the `zip` crate itself

Hennzau:
On zip2: https://github.com/zip-rs/zip2/blob/7613e6b3178ea4c732484548ec58d7e52e131e84/README.md?plain=1#L26 It says that multi-disk zip are not supported, that may be why

Hennzau:
I managed to make it work. Now linux/macos archives are regular `tar.gz` and windows archives are `.zip` so we avoid multi-disk archives.

Here is the code I changed in my own version of this function:

```toml
self_update = { version = "0.27.0", features = [
    "rustls",
    "archive-zip",
    "archive-tar",
    "compression-flate2",
], default-features = false }
```

I added a `build.rs` procedure inside the dora-cli crate to retrieve the exact target, that looks like that:
```rust
fn main() {
    println!(
        "cargo:rustc-env=TARGET={}",
        std::env::var("TARGET").unwrap()
    );
}
```

and the command:
```rust
#[cfg(target_os = "linux")]
let bin_path_in_archive = format!("dora-cli-{}/dora", env!("TARGET"));
#[cfg(target_os = "macos")]
let bin_path_in_archive = format!("dora-cli-{}/dora", env!("TARGET"));
#[cfg(target_os = "windows")]
let bin_path_in_archive = String::from("dora.exe");

let status = self_update::backends::github::Update::configure()
    .repo_owner("dora-rs") // You can try if it's working by replacing it with my own fork: "Hennzau"
    .repo_name("dora")
    .bin_path_in_archive(&bin_path_in_archive)
    .bin_name("dora")
    .show_download_progress(true)
    .current_version(env!("CARGO_PKG_VERSION"))
    .build()?;
```

PR #802, opened 2025-03-04
Update CI runners to `ubuntu-22.04`

phil-opp:
The ubuntu 20.04 runners will be removed in April 2025. There are already scheduled `brownouts` that lead to build failures, so we should fix it right now.

PR #916, opened 2025-03-23
use `cargo-dist` tool for dora-cli releases

Hennzau:
[`cargo-dist`](https://github.com/axodotdev/cargo-dist) is used by several big projects (including `uv`). It's really powerful, and so I think it can be useful to use it.

# Changes

The old `release.yml` was doing both `crates.io` releases and the build/upload of the `dora-cli`. Because `cargo-dist` needs its file to be `release.yml`, I renamed the old one `cargo-release.yml` (just as `pip-release.yml`).

Also the `cargo-release.yml` is just doing the release in `crates.io`. The build/upload is now the role of the new `release.yml` file with `cargo-dist`.

# Scripts

```bash
curl --proto '=https' --tlsv1.2 -LsSf https://github.com/dora-rs/dora/releases/latest/download/dora-cli-installer.sh | sh
```

```bash
powershell -ExecutionPolicy ByPass -c "irm https://github.com/dora-rs/dorareleases/latest/download/dora-cli-installer.ps1 | iex"
```

Advantages => support all platforms, and also all shells

# Usage

Whenever you create a Release and the associated tag with it, it will launch the workflow and when artifacts are ready, they be added to the Release announce. **The version of the tag created must match the crate version in the Cargo.toml file**


Hennzau:
After investigation, we should:
- Use https://opensource.axo.dev/cargo-dist/book/ci/customizing.html#custom-jobs for
```yaml
      - name: Free Disk Space (Ubuntu)
        uses: jlumbroso/free-disk-space@main
        if: runner.os == 'Linux'
        with:
          # this might remove tools that are actually needed,
          # if set to "true" but frees about 6 GB
          tool-cache: true

          # all of these default to true, but feel free to set to
          # "false" if necessary for your workflow
          android: true
          dotnet: true
          haskell: true
          large-packages: false
          docker-images: true
          swap-storage: true
```

- use https://opensource.axo.dev/cargo-dist/book/ci/customizing.html#bring-your-own-release to make sure that we can create the release ourselves

- use `cargo-release` https://opensource.axo.dev/cargo-dist/book/workspaces/cargo-release-guide.html for releasing all crates to `crates.io` automatically together with `cargo-dist`

Hennzau:
This is working really well. You can see my test repo: https://github.com/Hennzau/dora/releases/tag/0.3.11

This is also making it possible to use `self_update` without issues. as `cargo-dist` is making really well supported archives #913

Cargo.toml
```diff
@@ -173,3 +173,8 @@ path = "examples/cmake-dataflow/run.rs"
 name = "cxx-ros2-dataflow"
 path = "examples/c++-ros2-dataflow/run.rs"
 required-features = ["ros2-examples"]
+
+# The profile that 'dist' will build with
+[profile.dist]
+inherits = "release"
+lto = "thin"
```
haixuanTao, line 180: I don't think this is necessary right?

Otherwise, can we use:
```bash
[profile.dist]
inherits = { workspace = true }
lto = { workspace = true }
```

.github/workflows/release.yml
```diff
@@ -1,183 +1,289 @@
...
+  host:
+    needs:
+      - plan
+      - build-local-artifacts
+      - build-global-artifacts
+    # Only run if we're "publishing", and only if local and global didn't fail (skipped is fine)
+    if: ${{ always() && needs.plan.outputs.publishing == 'true' && (needs.build-global-artifacts.result == 'skipped' || needs.build-global-artifacts.result == 'success') && (needs.build-local-artifacts.result == 'skipped' || needs.build-local-artifacts.result == 'success') }}
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    runs-on: "ubuntu-20.04"
```
haixuanTao, line 221: I'm pretty sure that ubuntu 20.04 is already being phased out and is unreliable: https://github.com/actions/runner-images/issues/11101

See: https://github.com/dora-rs/dora/pull/802

Cargo.toml
Hennzau: Unfortunately we need it on the workspace Cargo.toml, but not in the dora-cli package

haixuanTao (commented):
Can we have a CD that does not rely on ubuntu20.04?

Cargo.toml
haixuanTao: ah ok!

.github/workflows/release.yml
Hennzau: In fact, almost all the CI will be done with ubuntu-22.04 and windows-2022. But the `cargo-dist` workflow is as follow:

- Run a CI on **ubuntu 20.04** to run `dist plan` (i.e a JSON file with all the config)
- Then run multiple CI on runners provided by `dist plan` (here I explicitly ask for ubuntu-22.04)

So the `dist plan` running on ubuntu-20.04 is a limitation by `cargo-dist`. But we can change that directly on the workflow if you prefer

haixuanTao:
I think this a good change!

Thanks in advance!

.github/workflows/release.yml
haixuanTao: I see. It seems like there is already action taken here: https://github.com/axodotdev/cargo-dist/issues/1760

Hennzau: I just found the `global` key for things like host, plan etc.. So now it's running on ubuntu-22.04 for global things. And will run on the runners provided in the config file for others targets. Or de default ones if not precised:

```toml
[dist.github-custom-runners]
x86_64-pc-windows-msvc = "windows-2022"
x86_64-unknown-linux-gnu = "ubuntu-22.04"
global = "ubuntu-22.04"

```

Hennzau: https://opensource.axo.dev/cargo-dist/book/ci/customizing.html#custom-runners

Tell me if the default values are okay for you for macos

haixuanTao: macOS-13 default looks good to me!

.github/workflows/release.yml
```diff
@@ -1,183 +1,289 @@
...
+# will be marked as a prerelease.
 on:
-  release:
-    types:
-      - "published"
-  workflow_dispatch:
+  pull_request:
+  push:
+    tags:
+      - '**[0-9]+.[0-9]+.[0-9]+*'
```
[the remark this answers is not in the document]
Hennzau: I'm trying that on another repo to make sure everything is working as expected!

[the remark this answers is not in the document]
Hennzau: Seems to work perfectly now

Hennzau:
I'm checking if intentionally failing the CI and then run workflow_dispatch is working

haixuanTao (approved):

Hennzau:
@haixuanTao do you know why 51 CI jobs are being queued here? I guess it's fine because I didn't change anything in the code but I would have like to see the full CI before merging

haixuanTao:
Yeah, so there is only 5 macos runner compared to 60 linux runner making the 5 macos worker often saturated.

It's going to happen at some point but as this PR does not conflict with it, we can merge it for now

Hennzau:
Hi @phil-opp. Indeed the file is auto generated and to regenerate it simply run `dist generate`. You can also make sure everything is working properly with `dist plan`. You can try to build it yourself with `dist build`.
Also if you plan to add other things `dist init` can be used (it's safe, it will not overwrite previous configuration).

**However**, Xavier wanted to have : 
```yaml
on:
  release:
    types:
      - "published"
  workflow_dispatch:
```
And I didn't find a way to make such thing inside `dist-workspace.toml`, it's either triggered always manually or always with tags. So I had to add `allow-dirty = ["ci"]` to be able to manually edit it.

Hennzau:
Also, the CI file `release.yml` is really really general, I don't think we will ever need to edit it. Almost all the actual CI is determined during the first CI step (with `dist plan`)

phil-opp:
Sounds good, thanks for the explanation!

PR #920, opened 2025-03-24
Implementing Self update

Shar-jeel-Sajid:
So, As mentioned [here](https://github.com/dora-rs/dora/pull/913#issuecomment-2747339237). I did these changes and tried it with "Hennzau" fork and it seems to be working fine.
```Checking for updates...
Checking target-arch... x86_64-unknown-linux-gnu
Checking current version... v0.3.10
Checking latest released version... v0.3.11
New release found! v0.3.10 --> v0.3.11
New release is compatible

dora release status:
  * Current exe: "/home/arlott/gsoc/dora/dora/target/debug/dora"
  * New exe release: "dora-cli-x86_64-unknown-linux-gnu.tar.gz"
  * New exe download url: "https://api.github.com/repos/Hennzau/dora/releases/assets/240150528"

The new release will be downloaded/extracted and the existing binary will be replaced.
Do you want to continue? [Y/n] Y
Downloading...
[00:00:03] [========================================] 15.41MB/15.41MB (0s) Done
Extracting archive... Done
Replacing binary file... Done
Successfully updated Dora CLI to version: 0.3.11```

### DIFF
--- a/binaries/cli/build.rs
+++ b/binaries/cli/build.rs
@@ -1,0 +1,6 @@
+fn main() {
+    println!(
+        "cargo:rustc-env=TARGET={}",
+        std::env::var("TARGET").unwrap()
+    );
+}
--- a/binaries/cli/src/lib.rs
+++ b/binaries/cli/src/lib.rs
@@ -241,8 +241,16 @@ enum Command {
         quiet: bool,
     },
 
-    SelfUpdate {
+    Self_ {
         /// Only check for updates without installing
+        #[clap(subcommand)]
+        command: SelfSubCommand,
+    },
+}
+
+#[derive(Debug, clap::Subcommand)]
+enum SelfSubCommand {
+    Update {
         #[clap(long)]
         check_only: bool,
     },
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by Hennzau** [comment]
Great! I think this is good! Sorry I thought that the use of `self_update` would be really easy, I didn't realize that current `dora` artifacts were not really standard. Do you have a MacOS/Windows computer to try it on it also?

**Comment by Shar-jeel-Sajid** [comment]
No. I don't have a computer with either Windows or MacOS.

**Review by Hennzau — APPROVED** [review]
This is great! I hope it's working on MacOS and Windows as well!

File: binaries/cli/build.rs [hunk 1]
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    println!(
+        "cargo:rustc-env=TARGET={}",
+        std::env::var("TARGET").unwrap()
+    );
+}
```
haixuanTao, line 6: What is this for ?

Hennzau: standard tarballs are not striped. It's in this format: 
```
.
└── dora-cli-arch-target.tar.gz/
    └── dora-cli-arch-target/
        └── dora # the binary
```

And so to be able to extract the dora binary we need to know the sequence `arch-target` (`x86_64-unknown-linux` for instance). There is an ENV variable for that, but only available during the build process of a package

binaries/cli/src/lib.rs
```diff
@@ -537,6 +551,61 @@ fn run(args: Args) -> eyre::Result<()> {
             .context("failed to run dora-daemon")?
         }
         Command::Runtime => dora_runtime::main().context("Failed to run dora-runtime")?,
+        Command::Self_ { command } => match command {
+            SelfSubCommand::Update { check_only } => {
+                println!("Checking for updates...");
+
+                #[cfg(target_os = "linux")]
+                let bin_path_in_archive = format!("dora-cli-{}/dora", env!("TARGET"));
```
haixuanTao, line 559:
```suggestion
                let bin_path_in_archive = format!("dora-cli-{}/dora", env!("CARGO_BUILD_TARGET"));
```

I think this should remove the need for the build script

Hennzau: I don't think CARGO_BUILD_TARGET is actually an env variable? I mean, it's not present in my code x)

haixuanTao: Ah ok indeed!

Sorry!

Issue #882, opened 2025-03-17
Add `mlx-lm` and `mlx-vlm` node into dora node-hub


haixuanTao:
As a follow up from adding `llama-cpp-python` as well as `transfromers` we can add mlx-lm and mlx-vlm to make it easier for user of macos :)

Choudhry18:
@dora-bot assign me

Sumitb09:
@dora-bot assign me

Clement795:
@dora-bot assign me

PR #922, opened 2025-03-24
Fix: Add MLX node in dora node-hub #882

Sumitb09:


Hennzau:
In my understanding, `mlx` is only available on macos, so you need to tell the CI to only run the test on macos. But to be honest I don't know how the node-hub CI works, we have to ask @haixuanTao for that

PR #924, opened 2025-03-25
Fix: RUST_LOG=.  dora run bug

starlitxiling:
This PR fixes the bug that the specified log level is invalid when using dora run


Hennzau (approved):

eladb (commented):
Thank you so much!

PR #940, opened 2025-04-03
Fix `dora runtime` spawn command for dora installed through `pip`

### LINKED ISSUE #900: errror about operator node in C++
dataflow.yml file:

```
nodes:
    - id: node-cpp-api
      path: install/bin/node_cpp_api
      inputs:
          tick: dora/timer/millis/300
      outputs:
          - image

    - id: runtime-node-1
      operators:
          - id: operator-plot
            shared-library: install/lib/operator_plot
            inputs:
                image: node-cpp-api/image

```

commands to start dataflow.yml

```
source .venv/bin/activate
dora build dataflow.yml --uv
dora run dataflow.yml --uv
```

[ERROR]
Dataflow failed:

Node `runtime-node-1` failed: exited with code 2 with stderr output:
/home/seer/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11: can't open file '/home/seer/Project-Rust/dora/examples/cmake-dataflow-camera/runtime': [Errno 2] No such file or directory

phil-opp:
Pip installations of dora use a small Python wrapper script, which then calls a functions of the `dora_cli` module. This means that the `current_exe()` will be a python binary instead of a standalone `dora` binary. This leads to errors when trying to start a `dora runtime` instance for shared library operators (see #900).

This commit fixes this issue by introducing a new `DoraCommand` struct that is initialized in the respective main function. For normal Rust binaries, this is just initialized with `current_exe()` as before. When invoked from Python, we also include the first argument, which should be the path to the Python wrapper script invoking dora. We also use the 0th argument instead of `current_exe` for invoking the `python` binary because `current_exe` resolves symlinks on some platforms, which affects import statements (different base directory).

Fixes #900


Shar-jeel-Sajid:
Nice work. I ran the `cmake-datflow` example with it. It seems to be working fine.

apis/python/node/src/lib.rs
```diff
@@ -387,7 +387,12 @@ pub fn run(dataflow_path: String, uv: Option<bool>) -> eyre::Result<()> {
         .enable_all()
         .build()
         .context("tokio runtime failed")?;
-    let result = rt.block_on(Daemon::run_dataflow(&dataflow_path, uv.unwrap_or_default()))?;
+    let dora_command = DoraCommand::from_python_env()?;
```
haixuanTao, line 390: FYI, this is suppose to be run within a python script unrelated to the cli.

I'm slightly confused, why do we need dora command in this case. Is this to find the runtime python path?

haixuanTao:
FYI, I have fixed this issue for the daemon and the coordinator with:

```rust
fn start_daemon() -> eyre::Result<()> {
    let path = if cfg!(feature = "python") {
        std::env::args_os()
            .nth(1)
            .context("Could not get first argument correspond to dora with python installation")?
    } else {
        std::env::args_os()
            .next()
            .context("Could not get dora path")?
    };
    let mut cmd = Command::new(path);
    cmd.arg("daemon");
    cmd.arg("--quiet");
    cmd.spawn().wrap_err("failed to run `dora daemon`")?;

    println!("started dora daemon");

    Ok(())
}
```

I think this changes creates a lot of additional argument changes that can be annoying down the line.

apis/python/node/src/lib.rs
phil-opp: We need to be able to invoke the Dora CLI when there are any operators present in the dataflow. Operators are run by a `dora runtime` process, which we need to spawn.

This spawn command differs based on how Dora was installed. If it was installed as a normal binary,  we can use `current_exe` to get the path to the binary. For Python-based installations, this points to a `python` executable instead.

Either way, you're totally right that this is the wrong approach. Dora might still be installed as a normal binary even if the run command is invoked from Python. Or even worse, Dora might not be installed at all....

haixuanTao:
I have opened a simpler PR that I have tested on c++ dataflow example: https://github.com/dora-rs/dora/pull/1011

phil-opp:
Closed in favor of #1011

PR #942, opened 2025-04-04
Simplify the use of C/C++ API

starlitxiling:
This PR packages the header files and library files needed to use the Dora c/c++ api. I tested the whole process on macos and it is normal. 
For the windows platform, there are some problems in the linking process, so I cancelled it. 
I provide a script for users to install the files needed to use the api. Users only need to add `I/usr/local/include/dora` `-L/usr/local/lib -ldora_node_api_c` or `-L/usr/local/lib -ldora_node_api_cxx` when compiling the command, without having to consider specifying the path of the header file or library file.
I think we need to add in the documentation what is needed to use the API, such as `pip install dora-rs` for the python API

### DIFF
--- a/.github/workflows/libdora-release.yml
+++ b/.github/workflows/libdora-release.yml
@@ -1,0 +1,142 @@
+name: Build and Publish Dora C/C++ API
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+  push:
+    branches: [main]
+
+jobs:
+  build-and-release:
+    name: Build and Release C API for ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-gnu
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-musl
+          - os: ubuntu-latest
+            target: armv7-unknown-linux-musleabihf
+          - os: ubuntu-latest
+            target: i686-unknown-linux-gnu
+          - os: macos-latest
+            target: x86_64-apple-darwin
+          - os: macos-latest
+            target: aarch64-apple-darwin
+          # - os: windows-latest
+          #   target: x86_64-pc-windows-msvc
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install cross-compilation tools (Linux)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gcc-aarch64-linux-gnu gcc-arm-linux-gnueabihf gcc-i686-linux-gnu musl-tools
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Add Rust target
+        shell: bash
+        run: rustup target add ${{ matrix.target }}
+
+      - name: Install cross (Linux)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: cargo install cross || true
+
+      - name: Build for all targets (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          Write-Output "Building for target: $env:matrix_target"
+          cargo build --release --target $env:matrix_target -p dora-node-api-c
+          cargo build --release --target $env:matrix_target -p dora-operator-api-c
+          cargo build --release --target $env:matrix_target -p dora-node-api-cxx
+          cargo build --release --target $env:matrix_target -p dora-operator-api-cxx
+          Write-Output "Generating header file with: cargo run --example cxx-dataflow --release"
+          cargo run --example cxx-dataflow --release
+
+      - name: Build for all targets (Unix)
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          echo "Building for target: ${{ matrix.target }}"
+          if [[ "$RUNNER_OS" == "Linux" ]]; then
+            cross build --release --target ${{ matrix.target }} -p dora-node-api-c
+            cross build --release --target ${{ matrix.target }} -p dora-operator-api-c
+            cross build --release --target ${{ matrix.target }} -p dora-node-api-cxx
+            cross build --release --target ${{ matrix.target }} -p dora-operator-api-cxx
+          else
+            cargo build --release --target ${{ matrix.target }} -p dora-node-api-c
+            cargo build --release --target ${{ matrix.target }} -p dora-operator-api-c
+            cargo build --release --target ${{ matrix.target }} -p dora-node-api-cxx
+            cargo build --release --target ${{ matrix.target }} -p dora-operator-api-cxx
+          fi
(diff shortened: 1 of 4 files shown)


Hennzau:
Hi, have you tried the `libdora_release` workflow on your fork?

Hennzau:
oh yeah it seems so! thx

Hennzau:
The C++ Dataflow example doesn't seem to work on Linux

.github/workflows/libdora-release.yml
```diff
@@ -0,0 +1,142 @@
...
+          - os: ubuntu-latest
+            target: armv7-unknown-linux-musleabihf
+          - os: ubuntu-latest
+            target: i686-unknown-linux-gnu
+          - os: macos-latest
+            target: x86_64-apple-darwin
+          - os: macos-latest
+            target: aarch64-apple-darwin
+          # - os: windows-latest
+          #   target: x86_64-pc-windows-msvc
```
Hennzau, line 34: What would be the workaround to make it possible for Windows CXX users to use dora then?

Hennzau (changes requested):
I think this is good! This is a shame that there is no tools like `cargo-dist` to distribute libraries.
Just make sure the CI is passing on Ubuntu

starlitxiling:
> The C++ Dataflow example doesn't seem to work on Linux

The really error message is 
~~~bash
bfd plugin: LLVM gold plugin has failed to create LTO module: Opaque pointers are only supported in -opaque-po    inters mode (Producer: 'LLVM19.1.5-rust-1.84.1-stable' Reader: 'LLVM 14.0.0')
~~~
This is because for Ubuntu 22.04, the clang++ version it comes with is 14, and the related plug-ins are also 14, but for rust 1.84.1, it is built based on LLVM 19, so rust requires tools compatible with LLVM 19 to link libraries, which is not easy for Ubuntu 22.04.
We can choose to use the ubuntu-latest image in CI. This is verified in my own repository and can be compiled. However, for some people who use ubuntu22.04, this is not a good solution. Do you have any suggestions for this?

.github/workflows/libdora-release.yml
starlitxiling: I am not familiar with the linking process on Windows, nor do I know how a library is installed on Windows.

haixuanTao:
I'm slightly confused why the example doesn't work on ubuntu 22 when it uses to work before. 

Also, i think `install_libdora` probably requires either it's own readme or a proper example because as it is. I think it's a bit difficult to know how to use it.

examples/c++-dataflow/run.rs
```diff
@@ -3,6 +3,7 @@ use eyre::{bail, Context};
 use std::{
     env::consts::{DLL_PREFIX, DLL_SUFFIX, EXE_SUFFIX},
     path::Path,
+    // process::Command,
```
haixuanTao, line 6: Can we remove this?

examples/c++-dataflow/run.rs
```diff
@@ -112,11 +113,59 @@ async fn main() -> eyre::Result<()> {
 
     let dataflow = Path::new("dataflow.yml").to_owned();
     build_package("dora-runtime").await?;
+    archive_node_bridge(root, build_dir).await?;
```
haixuanTao, line 116: Let's keep it like this but i'm really not a fan of having to do all this heavy lifting within the example...

haixuanTao (approved):
Thanks a lot for this! 

I understand it wasn't easy especially as a lot of the heavy lifting is done within the example folder.

Could be nice to do a follow up PR and makes the compilation of the cxx library be made cleaner.

examples/c++-dataflow/run.rs
starlitxiling: ok，I am sorry for forget delete this

examples/c++-dataflow/run.rs
```diff
@@ -112,11 +112,59 @@ async fn main() -> eyre::Result<()> {
...
+        .arg("-o")
+        .arg(&node_bridge_o);
+
+    if !compile.status().await?.success() {
+        bail!("failed to compile node-bridge.cc with {}", compiler);
+    }
+
+    let static_lib = root
+        .join("target")
+        .join("debug")
```
haixuanTao, line 155: Why is this linked to the debug version?

install_libdora.sh
```diff
@@ -0,0 +1,154 @@
...
+  echo "Error: 'jq' not found!" >&2
+  exit 1
+fi
+
+echo "==> Determining local platform..."
+PLATFORM="$(detect_platform)"
+echo "    => PLATFORM: $PLATFORM"
+
+echo "==> Fetching latest Dora release info from GitHub..."
+LATEST_JSON=$(curl -s https://api.github.com/repos/starlitxiling/dora-rerun/releases/latest)
```
haixuanTao, line 67: Sorry can we update this?

.github/workflows/libdora-release.yml
```diff
@@ -0,0 +1,142 @@
...
+            cp apis/c/operator/*.h dist/include/ || echo "No headers found"
+            cp target/cxxbridge/dora-node-api-cxx/*.h dist/include/ || echo "No cxxbridge headers found"
+            cp examples/c++-dataflow/build/*.h dist/include/ || echo "No example headers found"
+            cp examples/c++-dataflow/operator-rust-api/*.h dist/include/ || echo "No operator_api headers found"
+          fi
+          echo "Copying library files..."
+          case "$RUNNER_OS" in
+            "Linux" | "macOS")
+              cp target/${{ matrix.target }}/release/*.a dist/lib/ || echo "No libdora_node_api_c.a found"
+              cp target/${{ matrix.target }}/debug/libdora_node_api_cxx.a dist/lib || echo "No libdora_node_api_cxx.a found"
```
haixuanTao, line 115: Are we using debug or release?

haixuanTao (commented):
Could you create a c++ example with a readme on how to use this, because it isn't quite clear to me

examples/c++-dataflow/run.rs
starlitxiling: This is because in the run.rs of the example, debug has been specified, and the compilation option --release has no effect.https://github.com/dora-rs/dora/blob/a30e760d0d8e274dbb549a4acbd532355c5afa19/examples/c%2B%2B-dataflow/run.rs#L95

.github/workflows/libdora-release.yml
starlitxiling: I think we need release version

starlitxiling:
Sorry for following up on this question just now. Since `run.rs` involves compiling more than one module, if we want to separate `debug` and `release`, it will cause too much work to be done in `run.rs`. I don't think this is a good solution, and it is not very suitable for an example. So I temporarily use debug when releasing the C library.

haixuanTao:
Can we make a standalone script, please? 

I don't think we can merge this as is as it's very confusing.

starlitxiling:
Yes, we made some unnecessary changes to the example, I hate this, so I will try other ways to achieve this goal.But I'm out of ideas for now, so any help would be appreciated!

haixuanTao:
You can create a standalone script independant of the example. Could you check https://github.com/dora-rs/dora/blob/7c44e7a2e60dfa11a1557615a7efbc4ebb418f48/libraries/core/src/bin/generate_schema.rs


```bash
cargo run -p dora-core generate_schemas
```

phil-opp:
Closing as inactive

PR #953, opened 2025-04-08
[fix] cargoHash for building via nix

muellerbernd:


muellerbernd:
This was closed in favor of #955

PR #954, opened 2025-04-08
Post 3.11 release fix

haixuanTao:
Minor license fixes as well as documentation updates!

### DIFF
--- a/.github/workflows/node-hub-ci-cd.yml
+++ b/.github/workflows/node-hub-ci-cd.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   find-jobs:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     name: Find Jobs
     outputs:
       folders: ${{ steps.jobs.outputs.folders }}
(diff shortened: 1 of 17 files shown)


haixuanTao:
one interesting feature of maturin zig is that they also link eternal dynamic library such as libdav1d-dev, making dora node not requiring system dependency when installed from pip, making it less likely fail. 

I have only currently added this support for amd64 but we will in the future also be able to support all platform.

haixuanTao:
> > one interesting feature of maturin zig is that they also link eternal dynamic library such as libdav1d-dev, making dora node not requiring system dependency when installed from pip, making it less likely fail.
> 
> If you link against a dynamic library, you normally need to have it installed on the target system too. This sounds like zig is somehow able to link the dynamic libraries statically, but I'm not sure how that is possible. Do you have any links that explain this feature?

I meam he copies the dynamic link into the wheel and so redistribute it at the same time, see: https://github.com/dora-rs/dora/actions/runs/14379049575/job/40318281234#step:9:2415

It's still dynamically loaded but from a copied version

phil-opp:
Ah, interesting! Does this happen automatically when a shared library is linked or is there any config required?

haixuanTao:
> Ah, interesting! Does this happen automatically when a shared library is linked or is there any config required?

It seems automatic. Only thing I have done is copy the shared library into its searched path:

https://github.com/dora-rs/dora/blob/0ec1214bb12b3337abaac43714f172afe0a78005/.github/workflows/node-hub-ci-cd.yml#L58-L59

Issue #957, opened 2025-04-09
Documentation of benches is incomplete


Ignavar:
Currently when following the instructions of the README in benches/llms the commands given to run a bench do not work when running the commands i get this error 

![Image](https://github.com/user-attachments/assets/d040be4f-631e-43d9-9fa4-7debc9a6710f)

this can be resolved by running the command
uv venv --seed -p 3.11
then installing the necessary packages needed when running the benchmark.
It would be more intuitive to  include a pyproject.toml containing all libraries needed to be downloaded and including the venv setup and libraries installation command in the README to improve the documentation and make the process easier for new users

PR #969, opened 2025-04-15
Ease of use changes in benches for issue #957

Ignavar:
Added a build command in llms yaml file to automatically install libraries needed to run benchmark_script.py useful for beginners and cases where you have freshly cloned the project and a python virtual environment has not been setup with the necessary dependencies to run the benchmark_script


haixuanTao:
Yeah I just reverted the breaking ci/cd

PR #985, opened 2025-04-26
Add macOS-only mlx-lm node to node-hub (#882)

Clement795:
This PR addresses issue #882 by adding support for `mlx-lm` in the dora node-hub.

### Changes:
- Added `mlx-lm` node for macOS compatibility (`node-hub/mlx-lm.py`).
- Made `mlx-lm` macOS-only with `platform.system()` check to skip execution on Linux/Windows, fixing CI failures.
- Tested locally on macOS 15.4 with `mlx-lm` version 0.23.0.

### Notes:
- `mlx-vlm` support is not included and can be addressed separately.
- CI Python tests should now pass across all platforms due to platform check.

Addresses #882 (partially, as `mlx-vlm` is not implemented).

@haixuanTao Please review this updated PR for adding `mlx-lm` support.


Clement795:
@haixuanTao The CI failed due to `node-hub/dora-mlx-lm/` depending on `mlx-lm`, which isn’t available on Linux/Windows. I removed it and used `node-hub/mlx-lm.py`. Do I need to modify the CI workflow to test `mlx-lm` only on macOS?

haixuanTao:
Can you put it in the skip folder?

See: https://github.com/dora-rs/dora/blob/66bf1b74a42f6832f2d5b793c53d7e2069bc298b/.github/workflows/node_hub_test.sh#L8

haixuanTao:
That makes sense. Thanks!

node-hub/dora-mlx-lm/dora_mlx_lm/main.py
```diff
@@ -0,0 +1,115 @@
...
+    Optimized for Apple M1, M2, M3 chips using the MLX framework for efficient inference.
+    Generates responses independently for each input, using only the system prompt as context.
+    """
+    # Initialize model and tokenizer
+    model, tokenizer = get_model()
+    node = Node()
+    history = [{"role": "system", "content": SYSTEM_PROMPT}] if SYSTEM_PROMPT else []
+
+    for event in node:
+        if event["type"] == "INPUT" and event["id"] == "prompt":
```
haixuanTao, line 64: Can we use the term `text`:
```suggestion
        if event["type"] == "INPUT" and event["id"] == "text":
```

haixuanTao (commented):
Thanks in advance

node-hub/dora-mlx-lm/dora_mlx_lm/main.py
Clement795: The change to `id="text"` looks good for consistency. Thanks for the update!

PR #989, opened 2025-04-28
Fix rerun-viewer example.

francocipollone:
### Summary

This example had two issues:
 - the path to the dora rerun executable in the dataflow file was wrongly set
 - it wasn't able to be run via cargo run command (only using uv installation + dora build + dora run)

### Test it
via cargo run:

`cargo run --example rerun-viewer`

via uv setup:
```
uv venv -p 3.11 --seed
uv pip install -e ../../apis/python/node --reinstall
dora build dataflow.yml --uv
dora run dataflow.yml --uv
```


haixuanTao (approved):
ok thanks!

PR #992, opened 2025-04-30
Fix: Allow source nodes to wait for stop signal (Fixes #500)

### LINKED ISSUE #500: Nodes with no inputs has empty events stream
If a node does not specify any inputs, it will have no event stream. And the event stream is always returning None.

This is however misleading as we might want to check the event stream for other event than INPUT event such as Stopping event.

## Quick example

```rust
use std::time::Duration;

use dora_node_api::{self, dora_core::config::DataId, DoraNode, Event, IntoArrow};

fn main() -> eyre::Result<()> {
    println!("hello");

    let output = DataId::from("random".to_owned());

    let (mut node, mut events) = DoraNode::init_from_env()?;

    if events.recv_timeout(Duration::from_secs(100)).is_none() {
        eprintln!("Evemt is None");
    }
    Ok(())
}
```

With an empty input dataflow description, will return immediately instead of timing out on 100s.

ZhangHanDong:
Fixed #500 

Currently, source (input-less) Dora nodes close their event stream immediately after startup, making them unable to receive an ‎⁠`Event::Stop⁠` signal (such as from `Ctrl+C` or ‎⁠`dora stop`⁠). This prevents graceful shutdown.

Solution:

A new optional node configuration field, ‎⁠`wait_for_stop: bool⁠`, is introduced to control event stream behavior after all inputs are closed:

- ‎⁠`wait_for_stop: false`⁠: The event stream ends when all inputs are closed (default for nodes with inputs; unchanged).
- `wait_for_stop: true`⁠: The event stream remains open until an explicit `‎⁠Event::Stop`⁠ signal is received (default for source nodes with no inputs).

Users can override the default by setting this option explicitly in the dataflow YAML file.

Key Changes:
1. dora-message⁠:
     - Added ‎⁠`wait_for_stop: Option<bool>`⁠ to ‎⁠`descriptor::Node`⁠ for YAML parsing.
    - Added ‎⁠`wait_for_stop: bool⁠` to ‎⁠`descriptor::ResolvedNode`⁠ to store the resolved value.
2. dora-core⁠:
    - Updated ‎⁠`descriptor::resolve_aliases_and_set_defaults`⁠ to compute the final `‎⁠ResolvedNode.wait_for_stop⁠` value based on inputs and YAML settings. Fixed related ownership bug.
3. cora-daemon⁠:
    - Updated node spawning (‎⁠spawn::spawn_node⁠) to pass `‎⁠ResolvedNode.wait_for_stop`⁠ to the node communication listener (‎⁠`node_communication::spawn_listener_loop`⁠ and implementation).
    - Modified listener logic (‎⁠`node_communication::Listener`⁠) to check the ‎⁠`wait_for_stop⁠` flag when deciding whether to send ‎⁠AllInputsClosed⁠ or end the event stream. The event stream now only ends naturally if ‎⁠`wait_for_stop`⁠ is ‎⁠false⁠.
    - Ensured all nodes’ event streams can be stopped by ‎⁠`Event::Stop`⁠, regardless of ‎⁠`wait_for_stop`⁠.
4. Updated ‎⁠`dora-schema.json`⁠ to include the `‎⁠wait_for_stop⁠` field.

Impact:
- Source nodes now stay running by default and can respond to stop signals as expected.
- Users who want source nodes to exit immediately can set ‎⁠wait_for_stop: false⁠.
- Nodes with inputs retain their default (exit when inputs close), but can be set to wait for stop if desired.
- Fixed a bug in config resolution due to ownership issues.

Todos:
- Update core documentation (‎⁠libraries/core/README.md⁠) to explain the new config and defaults.
- Update Node API docs (Rust, Python, C, C++) about event stream lifecycle.

Todo Test:
- Source node default (waits for Stop).
- Source node with ‎⁠wait_for_stop: false⁠ (exits immediately).
- Node with inputs default (ends when inputs close).
- Node with inputs and ‎⁠wait_for_stop: true⁠ (waits for Stop).
- Stopping with Ctrl+C and ‎⁠dora stop⁠.

### DIFF
--- a/binaries/runtime/src/lib.rs
+++ b/binaries/runtime/src/lib.rs
@@ -318,10 +318,12 @@ async fn run(
                     if open_inputs.is_empty() {
                         // all inputs of the operator were closed
                         let Some(operator_config) = operators.get(&operator_id) else {
-                            tracing::warn!("received InputClosed event for unknown operator `{operator_id}`");
+                            tracing::warn!(
+                                "received InputClosed event for unknown operator `{operator_id}`"
+                            );
                             continue;
                         };
-                        
+
                         if !should_operator_wait_for_stop(operator_config) {
                             // If this operator should not wait for stop, close the event channel immediately
                             tracing::trace!("all inputs of operator {}/{operator_id} were closed -> closing event channel", node.id());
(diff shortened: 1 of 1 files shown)


libraries/core/src/descriptor/mod.rs
```diff
@@ -97,6 +98,11 @@ impl DescriptorExt for Descriptor {
                 }),
             };
 
+            // Calculate the final wait_for_stop value
+            let final_wait_for_stop = node.wait_for_stop.unwrap_or_else(|| node.inputs.is_empty());
```
phil-opp, line 102: This only works as expected for `CoreNodeKind::Custom`. For `CoreNodeKind::Runtime`, the `inputs` list will be always empty. I think this is the reason why the c dataflow test times out on CI.

The proper approach for `CoreNodeKind::Runtime` would be to check the inputs of each operator. We also need to update https://github.com/dora-rs/dora/blob/9f7c0959981be4f7031bd4b68e8c7bf31d1e8d55/binaries/runtime/src/lib.rs#L305-L310 and make the [main runtime loop](https://github.com/dora-rs/dora/blob/9f7c0959981be4f7031bd4b68e8c7bf31d1e8d55/binaries/runtime/src/lib.rs#L149) exit once all operators exited (we don't want to keep waiting for stop if all operators exited already).

phil-opp (commented):
Looks good to me overall, thanks a lot! There are some CI failures that are probably caused by the handling of runtime nodes/operators. See my inline comment.

binaries/runtime/src/lib.rs
```diff
@@ -303,12 +306,31 @@ async fn run(
...
+                            // If wait_for_stop is false, close the event channel immediately
+                            tracing::trace!("all inputs of operator {}/{operator_id} were closed -> closing event channel", node.id());
+                            open_operator_inputs.remove(&operator_id);
+                            operator_channels.remove(&operator_id);
+                        } else {
+                            // If wait_for_stop is true, keep the channel open to receive stop signals
+                            // but mark that all inputs are closed
+                            tracing::trace!("all inputs of operator {}/{operator_id} were closed, but wait_for_stop is true -> keeping event channel open for stop signal", node.id());
+                            open_operator_inputs.remove(&operator_id);
+                            // Note: operator_channels is NOT removed here - the operator can still receive stop signals
```
phil-opp, line 320: So if any operator is a source operator, all other operators that share the same runtime node will also wait for stop before exiting? I'm not sure if we want this behavior because this makes the operator behavior dependent on other operators.

In my opinion it would be more intuitive if we evaluate `wait_for_stop` for each operator individually. So only operators without inputs (or with explicit `wait_for_stop: true`) would wait for a stop signal. This way, nodes and operators would behave the same way.

phil-opp (commented):
Thanks for the adjustments!

ZhangHanDong:
@phil-opp 

There is a permission error in CI/ROS2 Bridge Examples.

```
Invoking "sudo apt-get update"
Error: The process '/usr/bin/sudo' failed with exit code 100
```

ZhangHanDong:
@phil-opp 

This CI might have some issues. I didn’t update the code for the past two weeks, but the CI kept running continuously for two weeks without showing any error messages.

Screenshot 2025-06-11 at 19 36 43
Screenshot 2025-06-11 at 19 36 58

Issue #978, opened 2025-04-19
Failure to correctly send `torch` tensors using IPC buffers


alberthli:
**Describe the bug**
When sending torch tensors using the provided utils in `dora.cuda`, I find that the data are frequently incorrect on the receiver side. This typically happens when the tensor is, e.g., composed of other constituent tensors (like from stacking or concatenating).

**To Reproduce**
Simple MWE with a sender and receiver:
```python
# sender.py
import torch
from dora import Node
from dora.cuda import torch_to_ipc_buffer

        
def main():
    sender = Node(node_id="sender")
    tensor1 = torch.arange(6, dtype=torch.float32, device="cuda").reshape(2, 3)
    tensor2 = torch.arange(6, dtype=torch.float32, device="cuda").reshape(2, 3) + 6.0
    tensor = torch.stack([tensor1, tensor2], dim=0)
    ipc_buffer, metadata = torch_to_ipc_buffer(tensor)
    for _ in range(100):
        sender.send_output("tensor", ipc_buffer, metadata)

if __name__ == "__main__":
    main()
```

```python
# receiver.py
import pyarrow as pa
import torch
from dora import Node
from dora.cuda import cudabuffer_to_torch, ipc_buffer_to_ipc_handle


def main():
    receiver = Node(node_id="receiver")
    ctx = pa.cuda.Context()
    tensor1 = torch.arange(6, dtype=torch.float32, device="cuda").reshape(2, 3)
    tensor2 = torch.arange(6, dtype=torch.float32, device="cuda").reshape(2, 3) + 6.0
    expected_tensor = torch.stack([tensor1, tensor2], dim=0)

    while True:
        event = receiver.next(timeout=1e-3)
        if event["type"] == "INPUT":
            ipc_handle = ipc_buffer_to_ipc_handle(event["value"])
            cudabuffer = ctx.open_ipc_buffer(ipc_handle)
            tensor = cudabuffer_to_torch(cudabuffer, event["metadata"]).clone()
            assert torch.allclose(tensor, expected_tensor), f"Received tensor: {tensor} does not match expected tensor: {expected_tensor}"

if __name__ == "__main__":
    main()
```

```python
# mwe.yml
nodes:
  - id: sender
    path: /path/to/sender.py
    outputs:
      - tensor
  - id: receiver
    path: /path/to/receiver.py
    inputs:
      tensor:
        source: sender/tensor
        queue_size: 1
```

When I run
```bash
dora run mwe.yaml
```
[...]

PR #1002, opened 2025-05-12
Refactor Python CUDA IPC API

Mivik:
As noted in #978, the current Python CUDA IPC API fails when sending specific tensors.

My investigation suggests that `pyarrow.cuda` (used by `dora.cuda`), is likely the source of this failure. During testing with two processes (sender and receiver), the receiver receives `pyarrow.CudaBuffer` with corrupted content, and it appears that there's nothing we can do about this.

However, I found that the `numba.cuda` module is a suitable alternative for implementing CUDA IPC. As we already depend on `numba`, no new dependencies are introduced. I have refactored the API to use `numba.cuda`, and it has passed the tests proposed in #978.

Example & docstrings are updated accordingly.

**Note**: this change is not backward compatible.

### DIFF
--- a/apis/python/node/dora/cuda.py
+++ b/apis/python/node/dora/cuda.py
@@ -8,13 +8,19 @@ from numba.cuda import to_device
 
 # Make sure to install numba with cuda
 from numba.cuda.cudadrv.devicearray import DeviceNDArray
+from numba.cuda.cudadrv.devices import get_context
+from numba.cuda.cudadrv.driver import IpcHandle
 
-# To install pyarrow.cuda, run `conda install pyarrow "arrow-cpp-proc=*=cuda" -c conda-forge`
-from pyarrow import cuda
+
+import pickle
+
+from contextlib import contextmanager
+from typing import ContextManager
 
 
 def torch_to_ipc_buffer(tensor: torch.TensorType) -> tuple[pa.array, dict]:
-    """Convert a Pytorch tensor into a pyarrow buffer containing the IPC handle and its metadata.
+    """Convert a Pytorch tensor into a pyarrow buffer containing the IPC handle
+    and its metadata.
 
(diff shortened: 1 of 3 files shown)


haixuanTao:
Thanks for this investigation and this looks really cool! 

However, would it be possible to not depend on pickle as it would reduce GPU 0-copy to python only project.

The idea behind using cuda ipc handle was to be able to read this in like C/C++ or Rust.

Mivik:
> However, would it be possible to not depend on pickle as it would reduce GPU 0-copy to python only project.

I refactored the encoding part and now removed pickle.

apis/python/node/dora/cuda.py
```diff
@@ -24,75 +30,65 @@ def torch_to_ipc_buffer(tensor: torch.TensorType) -> tuple[pa.array, dict]:
...
     metadata = {
         "shape": device_arr.shape,
         "strides": device_arr.strides,
         "dtype": device_arr.dtype.str,
+        "size": size,
+        "offset": offset,
+        "source_info": json.dumps(source_info),
     }
-    return pa.array(handle_buffer, type=pa.uint8()), metadata
+    return pa.array(handle, pa.int8()), metadata
```
haixuanTao, line 43: This should be uint8 right?
```suggestion
    return pa.array(handle, pa.uint8()), metadata
```

haixuanTao (approved):
Looks good to me!

I think uint8 would be more resilient!

But thanks for everything!

apis/python/node/dora/cuda.py
Mivik: `handle` is actually a int8 array (ranged from `-128` to `127`), since its underlying c type is `char`

haixuanTao (approved):

PR #1007, opened 2025-05-21
Remove unused sysinfo monitor

Mivik:
`sysinfo::System::refresh_processes` would fetch info from all processes in the current OS, which is unnecessary when only one of them is of interest.


phil-opp (approved):
Looks good, thanks!

Issue #1001, opened 2025-05-11
Terminal logging freezes after ~30 seconds


yiyikkk:
**Describe the bug**
When running a simple node, the terminal logging freezes after around 30 seconds.
However, the node continues working correctly (the system is still running, and downstream nodes still receive outputs).

**To Reproduce**
Minimal reproduction node code:
```
import pyarrow as pa
from dora import Node

def main():
    node = Node()
    last_key = "c"
    try:
        while True:
            event = node.next(timeout=0.01)
            if event is not None:
                if event["type"] == "INPUT" and event["id"] == "tick":
                    node.send_output("char", pa.array([last_key]))
                    print(f"Sending character: {last_key}")
                elif event["type"] == "STOP":
                    break
    finally:
        pass

if __name__ == "__main__":
    main()
```
Minimal dataflow.yml:
```
nodes:
  - id: keyboard
    path: ./keyboard.py
    inputs:
      tick: dora/timer/millis/1000
    outputs:
      - char
```
- run `dora run dataflow.yml` to start the system
- Observe the terminal: it prints "Sending character: c" normally for about 30 seconds.
- After ~30 seconds, the terminal stops displaying new prints and any other system logging. No any error reported.
- Downstream nodes still correctly receive the outputs from the node. For example, I tried to add a logger node to log the received "char", it works correctly and won't stop after 30 seconds.

**Expected behavior**
Terminal should continue to display print logs without freezing or muting after 30 seconds.

**Screenshots or Video**

**Environments (please complete the following information):**
 - System info: MacOS 15.1.1
 - Dora version: dora-cli 0.3.11

**Additional context**

yiyikkk:
**Additional Observation**
I also checked the system log file generated for this `keyboard` node:
Log path: `out/0196c19e-4e11-73dc-9d15-cfa96ffa7db2/log_keyboard.txt`

* Before I manually terminated the running dataflow with `Ctrl+C`, the `log_keyboard.txt` also froze after 30 log entries:

  ```
  Sending character: c
  Sending character: c
  Sending character: c
  ... (a total of exact 30 entries)
  ```
* After I pressed `Ctrl+C` to stop the dataflow, **all the pending logs were suddenly flushed and appended to the file**, like this:

  ```
  Sending character: c
  Sending character: c
  Sending character: c
  ... (a total of ~30 entries)
[...]

Issue #1005, opened 2025-05-15
conversation demo pause after running for some seconds


stanal:
i'm trying the demo of https://dora-rs.ai/docs/guides/getting-started/conversation_py/ 
i can start it , but it only run for few seonds, and then has none ouput and log 

![Image](https://github.com/user-attachments/assets/829a5828-8a8b-4eac-9859-5e5b0982b5e1)

![Image](https://github.com/user-attachments/assets/9fe49857-0472-4b23-9dc7-612bfca6cf03)

stanal:
![Image](https://github.com/user-attachments/assets/5904d567-3757-4483-af58-24a85e43c4cc)
the dataflow is still running when check dora list

stanal:
![Image](https://github.com/user-attachments/assets/4a7ec5e1-9e27-407e-bd98-6a9aaf86c036)

ShashwatPatil:
this should not happen, could you share you dataflow.yml if not same as in the guide, since i was not able to reproduce this error

PR #1008, opened 2025-05-26
fix terminal not printing stdout on nvml warning

haixuanTao:
Bumping system metrics to the latest version removes nvml warnings that used to create a deadlock.

It should also improve system metrics performance by reducing the number of opened files.


haixuanTao:
Fix #1001 and #1005

PR #1009, opened 2025-05-27
refactor: use builder style

### LINKED ISSUE #1001: Terminal logging freezes after ~30 seconds

sjfhsjfh:
This PR refactors the tracing initialization to use the builder-style API, improving configurability and clarity. Additionally, it removes the blocking `std::io::stdout`-based tracing layer used in the `Node`, which previously caused log output to hang in async contexts (Fix #1001). 

Finally, several function names with misleading semantics have been renamed for clarity.


haixuanTao:
FYI, I fixed the CI/CD within #1008, would it be possible to rebase on top of it!

Thanks for the PR

sjfhsjfh:
Done.

haixuanTao:
Very sorry. 

Just want to merge couple of older PR first, before coming back to this but this looks very promising!

PR #1011, opened 2025-05-27
Fix spawning runtime through python when it is installed with pip

### LINKED ISSUE #900: errror about operator node in C++

haixuanTao:
This PR fixes #900 by checking the current exe of the daemon. 

If the daemon is running on python => spawn the runtime on python.
If the daemon is running on a dora binary => use current dora binary.

Alternative to #940

### DIFF
--- a/binaries/daemon/src/spawn.rs
+++ b/binaries/daemon/src/spawn.rs
@@ -269,7 +269,7 @@ pub async fn spawn_node(
                         "-n",
                         conda_env,
                         "python",
-                        "-c",
+                        "-uc",
                         format!("import dora; dora.start_runtime() # {}", node.id).as_str(),
                     ]);
                     command
@@ -295,7 +295,7 @@ pub async fn spawn_node(
                     };
                     // Force python to always flush stdout/stderr buffer
                     cmd.args([
-                        "-c",
+                        "-uc",
                         format!("import dora; dora.start_runtime() # {}", node.id).as_str(),
                     ]);
                     cmd
@@ -323,7 +323,7 @@ pub async fn spawn_node(
                     );
 
                     cmd.args([
-                        "-c",
+                        "-uc",
                         format!("import dora; dora.start_runtime() # {}", node.id).as_str(),
                     ]);
                     cmd
@@ -336,7 +336,10 @@ pub async fn spawn_node(
                     cmd
                 }
             } else {
-                bail!("Could not figure out dora installation. Could you try to reinstall dora or run it with `dora` command?");
+                bail!(
+                    "Cannot spawn runtime with both Python and non-Python operators. \
+                       Please use a single operator or ensure that all operators are Python-based."
+                );
             };
 
             command.current_dir(working_dir);


binaries/daemon/src/spawn.rs
```diff
@@ -301,14 +301,49 @@ pub async fn spawn_node(
...
+                    std::env::current_exe().wrap_err("failed to get current executable path")?;
+                let mut file_name = current_exe.clone();
+                file_name.set_extension("");
+                let file_name = file_name
+                    .file_name()
+                    .and_then(|s| s.to_str())
+                    .context("failed to get file name from current executable")?;
+
+                // Check if the current executable is a dora binary
+                if file_name.ends_with("dora") {
```
phil-opp, line 314: Relying on the executable name seems a bit hacky. For example, users might want to install multiple versions of dora at the same time and rename the executables for that. For example, they might have a `dora-0.3.9` and a `dora-0.3.11` executable.

I think the approach of #940 is more robust. By keeping track of the call site in `DoraCommand`, we know for sure whether `dora` was invoked as a binary or from a Python environment. (We just need to fix the command for Python environments in https://github.com/dora-rs/dora/pull/940/files#diff-30db22706909157739f838586f6c2282008e9fa502c3e0d54a818cafdec5893fR2213-R2227 . I think we could use the ` "-c", format!("import dora; dora.start_runtime() # {}", node.id).as_str()` from this PR for that.)

haixuanTao: What about we just check for python and if it's not python. we fall back to the current exe.

There is actually a lot of things that could go wrong with https://github.com/dora-rs/dora/pull/940/files#diff-30db22706909157739f838586f6c2282008e9fa502c3e0d54a818cafdec5893fR2213-R2227 as there is no guarantee that the python used for the cli is the same that the one from the current env in which we want to spawn the runtime node. In which case, using #940 is going to be extremely difficult to debug.

haixuanTao: I'm going to modify the current code to be able to spawn the current interpreter, this means we will be able to install shared-library from pip.

binaries/daemon/src/spawn.rs
```diff
@@ -301,14 +301,44 @@ pub async fn spawn_node(
...
+                    let mut cmd = tokio::process::Command::new(
+                        std::env::current_exe()
+                            .wrap_err("failed to get current executable path")?,
+                    );
+                    cmd.arg("runtime");
+                    cmd
+                }
             } else {
-                eyre::bail!("Runtime can not mix Python Operator with other type of operator.");
+                bail!("Could not figure out dora installation. Could you try to reinstall dora or run it with `dora` command?");
```
phil-opp, line 339: This `else` branch is entered when there are both python and non-python operators for the same runtime node. So this new error message is not correct.

binaries/daemon/src/spawn.rs
```diff
@@ -301,14 +301,44 @@ pub async fn spawn_node(
...
+
+                // Check if the current executable is a python binary meaning that dora is installed within the python environment
+                if file_name.ends_with("python") || file_name.ends_with("python3") {
+                    // Use the current executable to spawn runtime
+                    let python = get_python_path()
+                        .wrap_err("Could not find python path when spawning custom node")?;
+                    let mut cmd = tokio::process::Command::new(python);
+
+                    tracing::info!(
+                        "spawning: python -uc import dora; dora.start_runtime() # {}",
```
phil-opp, line 321: Nit: this says that `-uc` is passed as argument, but below we only pass `-c`.

binaries/daemon/src/spawn.rs
phil-opp: Yes, the python spawn command of #940 is not correct. What i meant is that we could use the `DoraCommand` approach to avoid the executable name check. My idea was that we keep track of whether we enter the dora code from the Rust `main` function or from one of the Python functions (`run` or `py_main`). This way, we don't need to check the executable file name since we already know whether it's a Rust executable or Python-based.

PR #1012, opened 2025-05-27
dora-mujoco simulation node with example for controlling any arm 

ShashwatPatil:
This pull request introduces the initial setup for the `dora-franka-mujoco` project, including documentation, configuration files, and assets for simulating the Franka Emika Panda robot in MuJoCo. The most important changes include Defining a YAML configuration for nodes, adding robot assets and MuJoCo XML files, and creating a Python package structure for the project.

### Documentation and Configuration:

* TODO a `README.md` file with installation, formatting, linting, and testing instructions for the `dora-franka-mujoco` project.
* Added `demo.yml` to define the configuration for nodes, including `gamepad` and `mujoco_franka`, with their inputs, outputs, and dependencies.

### Robot Simulation Assets:

* Added `panda.xml` to define the MuJoCo model for the Franka Emika Panda robot, including its joints, actuators, collision meshes, and visual meshes.
* Added `scene.xml` to define the simulation environment, including a table, a movable box, and visual settings for the scene.
* Added a list of asset files (`names.txt`) and corresponding 3D model files for the robot's components, such as links, fingers, and the hand.


haixuanTao:
I think it's a good idea to store the meshes and exposes the xml. Making things slightly easier.

I probably missed it but how do you make sure that the file are going to be retrieved by mujoco?

ShashwatPatil:
> I probably missed it but how do you make sure that the file are going to be retrieved by mujoco?

Currently the meshes are downloaded using huggungface repo, the mesh files are automatically discovered through MuJoCo's XML-based model definition system, you can see that inside the panda.xml file `<compiler angle="radian" meshdir="assets" autolimits="true"/>` and all the meshes are referenced after this to use in the ` <asset> tag ` 

Sorry for the late reply

node-hub/dora-mujoco-sim/dora_mujoco_sim/main.py
```diff
@@ -0,0 +1,154 @@
...
+    def _load_model_mapping(self) -> dict:
+        """Load robot model mapping from JSON file."""
+        config_path = Path(__file__).parent / "robot_models.json"
+
+        with open(config_path) as f:
+            mapping_data = json.load(f)
+        
+        model_mapping = {}
+        for models in mapping_data.values():
+            model_mapping.update(models)
```
haixuanTao: Wouldn't this be easier if there is only one hierarchical level within the json file?

haixuanTao:
Isn't the urdf loaded from https://github.com/robot-descriptions/robot_descriptions.py?tab=readme-ov-file?

ShashwatPatil:
Yes the mujoco sim node uses that, but mujoco franka node still uses hugging face to get the assets, also i have modifed them a bit to include the table ans the small cube

Should have made a separate PR for that 😅

examples/franka-mujoco-control/02_target_pose_control/nodes/franka_controller_pytorch.py
haixuanTao, line 1: It seems that this controller could work on any arm with an urdf right?

Maybe we could naming more apparent or directly use existing dora-pytorch-kinematics?

I wish this tutorial to have as many reusable component as possible.

examples/franka-mujoco-control/04_gamepad_collision_avoidance/nodes/franka_collision_controller.py
```diff
@@ -0,0 +1,366 @@
...
+        self.max_angvel = 0.785
+        
+        # Collision avoidance parameters
+        self.collision_check_enabled = True
+        self.min_link_distance = 0.02  
+        self.collision_force_gain = 1000.0  
+        self.collision_influence_distance = 0.05  
+        
+        # Define which links can collide with each other (non-adjacent only)
+        self.collision_pairs = [
```
haixuanTao: so it seems that every pair can collide with each other?

Can't we have a more automated way to compute this?

examples/franka-mujoco-control/02_target_pose_control/nodes/franka_controller_pytorch.py
ShashwatPatil: yes the controller should work with ant urdf, will try to make it reusable for any arm.

examples/franka-mujoco-control/04_gamepad_collision_avoidance/nodes/franka_collision_controller.py
ShashwatPatil: I had tried using PyTorch Volumetric to get this. It was having too much wobbly motion. And the collision avoidance part is not very reusable for any other arm

PR #1032, opened 2025-06-23
Add automatic robot descriptions URDF retrieval from https://github.com/robot-descriptions/robot_descriptions.py

haixuanTao:
@ShashwatPatil  In order to make debugging easier, I have made it possible from multiple node to retrieve the urdf given the robot_descriptions handler example: `so_arm101_description` this makes it easier to download and load them at runtime. 

I have added example within `examples/urdf`


haixuanTao:
Ok, so unfortunately I am not able to replicate your error. 

I have pushed a commit that in theory should help with loading the robot description. Maybe you could try a second time if you have time?

haixuanTao:
Ok very cool!

Thanks for checking. I did pushed couple of things on robot_descriptions

Issue #1019, opened 2025-06-14
A command for simplifying node testing


XiaoKuge:
When we build a new dora node, the natural step is to test whether it's working or not. To test the node, right now, we will need to: 
1. create a dataflow which include a terminal-input node and the newly created node. 
2. open two terminals, one to run the dataflow and the other for terminal-input to check the input and output of the dataflow.

We wonder that if this can be simplified by having some command like: dora run-node [nodename] or something like that. Input needed by the node could be given through a json file, output can be printed directly on the screen or to some output files.

In this way, we don't have to create dataflow YML; also, we don't have to have two terminals up for testing.

We believe that it will help significantly productivity, especially when developers want to try out many different nodes. 

Thank you!

PR #1036, opened 2025-06-25
Add `inspect` command to inspect node outputs

Mivik:
Related: #1019

# Description

This PR introduces `dora inspect`. This command enables user to easily inspect the output(s) of nodes, without modifying the nodes themselves. This could enhance debug experience especially in early development stages.

# Implementation

Each inspect session is assigned with a unique `inspector_id` (UUID).

`inspected_outputs` is added to `RunningDataflow` in daemon, marking which outputs are being inspected. When at least one inspector is inspecting an output, this output is published to Zenoh.

The CLI connects to Zenoh session, receives data and pretty-prints them to the terminal.

# Notes

Large change in `Cargo.lock` is introduced by updating updated `zenoh` dependency in `communication-layer` crate.

### DIFF
--- a/binaries/cli/src/command/inspect.rs
+++ b/binaries/cli/src/command/inspect.rs
@@ -1,0 +1,254 @@
+use std::{net::IpAddr, ptr::NonNull, sync::Arc};
+
+use clap::Args;
+use colored::Colorize;
+use dora_core::{
+    config::InputMapping,
+    topics::{
+        open_zenoh_session, zenoh_output_publish_topic, DORA_COORDINATOR_PORT_CONTROL_DEFAULT,
+    },
+};
+use dora_message::{
+    cli_to_coordinator::ControlRequest,
+    common::Timestamped,
+    coordinator_to_cli::ControlRequestReply,
+    daemon_to_daemon::InterDaemonEvent,
+    id::{DataId, NodeId},
+    metadata::{ArrowTypeInfo, BufferOffset},
+};
+use eyre::{bail, eyre, Context};
+use tokio::{runtime::Builder, task::JoinSet};
+use uuid::Uuid;
+
+use crate::{
+    command::{default_tracing, Executable},
+    common::{connect_to_coordinator, resolve_dataflow_identifier},
+    LOCALHOST,
+};
+
+/// Inspect data in terminal.
+#[derive(Debug, Args)]
+pub struct Inspect {
+    /// Identifier of the dataflow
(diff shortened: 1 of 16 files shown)


haixuanTao:
Could you look into the `publish_all_messages_to_zenoh` argument within the dataflow to see if we can build on top of that instead of messing with dora internals?

@phil-opp

Mivik:
> if we could just tap into the zenoh communication

That's basically the idea behind this PR. I tried to keep modifications to internal minimal but it's still necessary because we need to tell Dora daemon to publish data of interest on Zenoh, or else we couldn't even see data on Zenoh.

haixuanTao:
I think for now, yes! 

Maybe in the future we can have more zenoh integration and then not rely on this flag anymore.

Mivik:
I've rebased the code onto the main branch. Can you check if there's anything left to be done? @haixuanTao

binaries/cli/src/command/inspect.rs
```diff
@@ -0,0 +1,254 @@
...
+            ControlRequestReply::DataflowInfo { descriptor, .. } => descriptor,
+            ControlRequestReply::Error(err) => bail!("{err}"),
+            other => bail!("unexpected list dataflow reply: {other:?}"),
+        }
+    };
+    if !dataflow_descriptor.debug.publish_all_messages_to_zenoh {
+        bail!(
+            "Dataflow `{dataflow_id}` does not have `publish_all_messages_to_zenoh` enabled. You should enable it in order to inspect data.\n\
+            \n\
+            Tip; Add the following snipppet to your dataflow descriptor:\n\
```
phil-opp: Minor suggestion: `Tip:` instead of `Tip;`

haixuanTao:
So before merging this, I just want to point out that this would be a breaking change and that we will have to upgrade all node to the latest version to work reliably. 

I would want to know before merging, are we sure we cannot make this feature less invasive than this?

phil-opp:
The latest version of the PR is based on publish_all_messages_to_zenoh, so no node update should be necessary. @Mivik could you update the PR description to match the current state of the PR.

I think the only breaking changes that this PR introduces is the new info message between the CLI and the coordinator and the change to the logs message. 

Technically, this would still require bumping the dora-message crate in a sever-incompatible way. However, then we would have to update all the node api versions as well. So I would do only a patch version bump and risk the deserializing error if CLI and coordinator are using different versions. Then no deamon or node update should be needed.

Mivik:
> @Mivik could you update the PR description to match the current state of the PR.

Updated

PR #1041, opened 2025-06-30
Add dataset collection node for dora 

ShashwatPatil:
This node is able to record a dataset in LeRobot format.
It records joint angle, can record multiple camera feeds using the `opencv-video-capture` node.
Also added example to record dataset using SO101 arm


ShashwatPatil:
Yes, i will update the readme with usage instructions and the example

haixuanTao:
FYI, I have added a fix on import naming based on latest main. I have also made it possible to train models without git cloning aerobat.

Would it be possible to add 3d visualisation of the arm using rerun in both dataflow?

ShashwatPatil:
removed threading from inference node, also added optional saving for avif frames along with the mp4 videos that lerobot saves

ShashwatPatil:
Thanks for correcting the Python version 👍

PR #1048, opened 2025-07-03
Removing support for macos x86 in order to unpin macos version

haixuanTao:
This PR removes macOS version pinning on Maturin that makes it the prebuilt will not installable on other macOS version.

I believe the original error is due to macOS x86 and I think it's time to remove support for it.


haixuanTao:
We could but Macbook Intel is going end of line soon and I kind of want to depreciate the node hub for it because Macbook Github action are slow and it doesn't really adds that much value

PR #1052, opened 2025-07-06
docs: fix speech-to-speech example README

chrislearn:


haixuanTao:
Thank you @chrislearn

PR #1053, opened 2025-07-07
Made some changes to make clippy happy

chrislearn:


node-hub/dora-rerun/src/series.rs
```diff
@@ -7,7 +7,7 @@ pub fn update_series(rec: &RecordingStream, id: DataId, data: ArrowData) -> Resu
     for (i, value) in series.iter().enumerate() {
         rec.log(
             format!("{}_{}", id.as_str(), i),
-            &rerun::Scalar::new(*value as f64),
+            &rerun::Scalars::new([*value]),
```
phil-opp, line 10: Was this really suggested by clippy?

@haixuanTao Is this change correct?

chrislearn: [Scalar](https://docs.rs/rerun/latest/rerun/struct.Scalar.html) is deprecated.

phil-opp: Makes sense, thanks!

PR #1060, opened 2025-07-13
fix: in c++ ros2 example, too few ticks.

drindr:
It seems that too few ticks to ensure everything work correctly.

The topics and the services still not initialized, the loop has been break and there are no responses from the `add` server.
1000 events counter is considerable for the whole program, while 20 received ticks are not so reasonable.

I'm running on an AMD 6800H Laptop with Ubuntu in Docker, whose performance is totally enough for such a simple example.


haixuanTao:
ok make sense thanks!

haixuanTao (approved):

PR #1064, opened 2025-07-15
Fix weird daemon folder

haixuanTao:
The current working folder when using distributed dora-daemon is a folder that does not exist making it impossible to spawn build command.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2158,9 +2158,7 @@ impl Daemon {
                 // use subfolder of daemon working dir
                 let daemon_working_dir =
                     current_dir().context("failed to get daemon working dir")?;
-                Ok(daemon_working_dir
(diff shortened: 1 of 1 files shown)


haixuanTao:
@phil-opp , I am struggling to understand this? 

Is this change ok?

phil-opp:
I saw that you're about to create a new release, so I merged https://github.com/dora-rs/dora/pull/1065 to revert this change.

phil-opp:
I opened https://github.com/dora-rs/dora/pull/1066 with the proper fix (i.e. create the directories if they don't exist). Could you try whether this fixes your issue?

(I'm open to changing the working directoy name if you don't like the `_work` name. We should just keep in mind that we already used this folder in the `v3.12` release (and it should work as expected with git repo sources), so changing the name again might be a bit surprising.)

PR #1067, opened 2025-07-16
Improve build logging

phil-opp:
- Buffer log messages that are sent before subscribers are connected
- Include working directory in 'running build command' message
- Lower log level of 'building' message to `Debug`


haixuanTao (approved):
Sure thing!

Thanks!

PR #1071, opened 2025-07-16
Remove rust--toolchain.toml and use rust-version in Cargo.toml

chrislearn:


haixuanTao:
Looks good to me, thanks!

phil-opp:
I don't think that rust/cargo ensures that the crate actually compiles with the specified `rust-version`. For example, try using a method such as https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.extract_if which was stabilized in 1.88. No errors or warnings happen for me.

So it's probably good to keep the `rust-toolchain.toml` in place, in addition to specifying the `rust-version`.

chrislearn:
> I don't think that rust/cargo ensures that the crate actually compiles with the specified `rust-version`. For example, try using a method such as https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.extract_if which was stabilized in 1.88. No errors or warnings happen for me.
> 
> So it's probably good to keep the `rust-toolchain.toml` in place, in addition to specifying the `rust-version`.

Or can we use github ci to make sure the minimum version of rust can be compiled normally?

phil-opp:
I'm also fine with that. What do you think, @haixuanTao ?

haixuanTao:
I didn't know that this did not work as expected. 

what does it do then?

I was under the impression that it could be neat to be able to switch between rust version between nodes.

PR #1072, opened 2025-07-16
Update rust edition to 2024

chrislearn:


haixuanTao:
FYI, updating rust edition makes it hard to use dora node api with older rust version. I would maybe wait to upgrade rust edition unless it is strictly necessary

chrislearn:
Maybe we can consider using `Squash and merge` when merging PRs in the future, this will make the history of main branch much cleaner.
image

PR #1076, opened 2025-07-19
Fix all clippy warning

chrislearn:


libraries/core/src/build/git.rs
```diff
@@ -10,6 +10,7 @@ use std::{
 use url::Url;
 
 #[derive(Default)]
+#[allow(dead_code)]
```
phil-opp: Can we remove the unused `reuse_for` field instead?

phil-opp (approved):
Looks good to me, thanks! Just one small suggestion:

libraries/core/src/build/git.rs
chrislearn: I have commented it out.

PR #1079, opened 2025-07-22
chore: some small fix about python examples

drindr:
- use newer python version in example to satisfy `dora-rerun`
- modify the `run.rs` in example `multi-env`. though it is not included in the workspace's example list, it is misleading.
- modify the field about `operator` in dataflow.yml. Because it seems that the field in code have been changed.


phil-opp:
@haixuanTao What do you think about these changes?

examples/python-operator-dataflow/dataflow_llm.yml
```diff
@@ -29,7 +29,8 @@ nodes:
   ## Speech to text
   - id: keyboard
     custom:
-      source: keyboard_op.py
+      path: keyboard_op.py
+      source: Local
```
haixuanTao, line 33: I don't think this is a good idea. Could we remove this?

PR #1081, opened 2025-07-23
CLI: `dora logs` enhancement

Mivik:
This PR adds two options: `-f`/`--follow` and `-n`/`--tail` to `dora logs`. These options behave similarly to their counterparts from [docker container logs](https://docs.docker.com/reference/cli/docker/container/logs/).

## Explanation

The `-f`/`--follow` option allows users to stream logs in real-time from the daemon. This provides a continuous, live view, similar to how one might "attach" to the daemon's output.

The -n/--tail option allows users to specify the number of historical log lines to display from the end of the log output. This could limit excessive output from the command.

## Other modifications

Prior to this PR, logs were printed using `bat`. While `bat` offers a wide range of features such as syntax highlighting and Git integration, only its paging functionality was ever utilized. Furthermore, the paging feature itself seems inconsistent and I can't make it work on my machine. Given these limitations, this PR removes the `bat` dependency and now simply prints all logs directly to the console, allowing users to leverage their preferred pager for viewing, mirroring Docker's approach.

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -69,7 +69,7 @@ pub fn logs(
     follow: bool,
     coordinator_addr: SocketAddr,
 ) -> Result<()> {
-    let (uuid, logs) = {
+    let logs = {
         let reply_raw = session
             .request(
                 &serde_json::to_vec(&ControlRequest::Logs {
@@ -84,7 +84,7 @@ pub fn logs(
 
         let reply = serde_json::from_slice(&reply_raw).wrap_err("failed to parse reply")?;
         match reply {
-            ControlRequestReply::Logs { uuid, data } => (uuid, data),
+            ControlRequestReply::Logs(data) => data,
             other => bail!("unexpected reply to daemon logs: {other:?}"),
         }
     };
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -664,7 +664,7 @@ async fn start_inner(
                                         tail,
                                     )
                                     .await
-                                    .map(|data| ControlRequestReply::Logs { uuid, data });
+                                    .map(ControlRequestReply::Logs);
                                     let _ = reply_sender.send(reply);
                                 }
                                 Err(err) => {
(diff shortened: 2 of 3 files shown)


binaries/daemon/src/lib.rs
```diff
@@ -2166,6 +2175,60 @@ impl Daemon {
...
+    'main: while tail > 0 && pos > 0 {
+        let new_pos = pos.saturating_sub(buffer.len() as u64);
+        file.seek(io::SeekFrom::Start(new_pos)).await?;
+        let read_len = (pos - new_pos) as usize;
+        pos = new_pos;
+
+        let read = file.read(&mut buffer[..read_len]).await?;
+        if read < read_len {
+            break;
+        }
```
phil-opp: AFAIK, `read` is allowed to return less bytes than available:

> It is not an error if the returned value n is smaller than the buffer size, even when the reader is not at the end of the stream yet.

So we probably want to use something like `read_exact` instead.

binaries/cli/src/command/logs.rs
```diff
@@ -55,36 +85,69 @@ pub fn logs(
...
+    if !follow {
+        return Ok(());
+    }
+    let log_level = env_logger::Builder::new()
+        .filter_level(log::LevelFilter::Info)
+        .parse_default_env()
+        .build()
+        .filter();
+
+    // subscribe to log messages
```
phil-opp, line 105: Using this approach we might miss log messages that happened in between the `Logs` and `LogSubcribe` requests, right? It would probably be better to do this in a single step, e.g. by extending the `LogSubscribe` message with the option to return old logs from file.

binaries/daemon/src/lib.rs
Mivik: This is unlikely to happen for regular files, but yeah I added it

binaries/cli/src/command/logs.rs
Mivik: Even if it's handled within one coordinator command, missing logs are hard to avoid (between retrieving logs from file and receiving logs from daemon). A feasible method is to attach a timestamp to each log message, and start receiving log events before reading log files then remove duplicates according timestamp. However it would increase transportation overhead. What do you think?

phil-opp: Let's get this merged for now as-is, we can always improve it later.

phil-opp:
Thanks again!

PR #1082, opened 2025-07-23
Document public `DYNAMIC_SOURCE` constant

phil-opp:
Also set `#![warn(missing_docs)]` for the module so that we get a warning if we add undocumented fields etc in the future.


phil-opp:
This is just a simple doc change, so merging without review;

PR #1091, opened 2025-07-25
Make up and destroy public dora function

haixuanTao:
Making up and destroy public dora function makes it possible to control the coordinator when it is not running or stopping it when necessary


phil-opp:
In my opinion we should deprecate the `up` command soon. It offers not enough control and often leads to confusion because the coordinator and daemon are run in background, so you never see their output. Instead, we should recommend to use the `dora run` command instead.

So I don't think that we should expose the up function even more. I think it would be better to expose the function behind `dora run` instead

haixuanTao:
ok

PR #1093, opened 2025-07-30
Implement a dora-openai-websocket server

haixuanTao:
This PR introduces a dora-openai-websocket that makes it possible to connect an openai client through dora in realtime.

The server is able to spawn and connect to dataflows allowing to customize the dataflow from the user client.

## Getting started

See README.md at examples/openai-realtime/README.md

### DIFF
--- a/node-hub/dora-openai-websocket/src/lib.rs
+++ b/node-hub/dora-openai-websocket/src/lib.rs
@@ -403,6 +403,7 @@ async fn handle_client(fut: upgrade::UpgradeFut) -> Result<(), WebSocketError> {
     // Local variable
 
     let mut call_id = 0;
+    let mut item_id = 0;
     loop {
         let event_fut = events.recv_async().map(Either::Left);
         let frame_fut = ws.read_frame().map(Either::Right);
@@ -421,11 +422,12 @@ async fn handle_client(fut: upgrade::UpgradeFut) -> Result<(), WebSocketError> {
                             let serialized_data =
                                 OpenAIRealtimeResponse::ResponseAudioTranscriptDelta {
                                     response_id: "123".to_string(),
-                                    item_id: "123".to_string(),
+                                    item_id: item_id.to_string(),
                                     output_index: 123,
                                     content_index: 123,
                                     delta: str.to_string(),
                                 };
+                            item_id += 1;
 
                             let frame = Frame::text(Payload::Bytes(
                                 Bytes::from(serde_json::to_string(&serialized_data).unwrap())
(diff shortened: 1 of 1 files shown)


node-hub/dora-openai-websocket/src/main.rs
```diff
@@ -0,0 +1,474 @@
...
+        dataflow,
+        name: Some(node_id.to_string()),
+        coordinator_addr: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+        coordinator_port: DORA_COORDINATOR_PORT_CONTROL_DEFAULT,
+        attach: false,
+        detach: true,
+        hot_reload: false,
+        uv: true,
+    })
+    .execute()
```
phil-opp: This is a synchronous call, which might block the async task. This is not recommended since it blocks a full thread of the tokio runtime.

node-hub/dora-openai-websocket/src/main.rs
```diff
@@ -0,0 +1,474 @@
...
+    let template = format!("{input_audio_transcription}-template-metal.yml");
+    println!("Filling template: {}", template);
+    replace_placeholder_in_file(&template, "NODE_ID", &node_id, &dataflow).unwrap();
+    // Copy configuration file but replace the node ID with "server-id"
+    // Read the configuration file and replace the node ID with "server-id"
+    dora_cli::command::Command::Start(Start {
+        dataflow,
+        name: Some(node_id.to_string()),
+        coordinator_addr: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+        coordinator_port: DORA_COORDINATOR_PORT_CONTROL_DEFAULT,
```
phil-opp: So this only works when the node is running on the same machine than the coordinator. This seems quite limiting.

phil-opp:
> Going through the CLI crate and opening a new connection to the coordinator seems a bit complicated, given that we already have a connection to a daemon through the node API. The daemon is already connected to the coordinator, so it could forward some start message to it.

The drawback of this approach is that this adds more functionality to the dora-node-api. So this only makes sense if we think that a functionality like this is common enough and also useful for other nodes.

> This way, we don't need to add the heavy CLI dependency

An alternative way to make the CLI dependency less heavy could be to make the `dora daemon`/`dora coordinator`/`dora runtime` commands optional features. Nodes like this one don't require it, so this could make the dependency more lightweight.

node-hub/dora-openai-websocket/src/main.rs
haixuanTao: Agreed I will make this configurable.

node-hub/dora-openai-websocket/src/main.rs
haixuanTao: I agree. Although, this is more of a proof of concept. I think it's fine to just leave this sync call for now and rework this later.

phil-opp:
I'm fine with making the functions public for now, but if this functionality is something that will also be needed by other nodes, I'd prefer a proper solution. The main limitation I'd like to avoid is that the coordinator currently has to run on the same machine as the node. Going through the daemon instead of the CLI would enable this functionality also in distributed settings.





> I think the thing is that implementing what you're mentioning is a lot of work as we would need to reimplement the node - daemon connection which I don't want to do at this moment.

I don't think that it's too much work. Let me try to draft a PR against your PR.

PR #1098, opened 2025-08-04
feat: make `path` and `build` field in dataflow file more flexible

drindr:
- support environment variable in the path like `$HOME/a/bc`
- support quoted argument in the build command like `bash -c "..."`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,9 +20,9 @@ checksum = "8b5ace29ee3216de37c0546865ad08edef58b0f9e76838ed8959a84a990e58c5"
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.29"
+version = "0.2.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec3672c180e71eeaaac3a541fbbc5f5ad4def8b747c595ad30d674e43049f7b0"
+checksum = "e074464580a518d16a7126262fffaaa47af89d4099d4cb403f8ed938ba12ee7d"
 dependencies = [
  "ab_glyph_rasterizer",
  "owned_ttf_parser",
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "ab_glyph_rasterizer"
-version = "0.1.8"
+version = "0.1.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
+checksum = "366ffbaa4442f4684d91e2cd7c5ea7c4ed8add41959a31447066e279e432b618"
 
 [[package]]
 name = "accelerate-src"
@@ -71,7 +71,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8bfae7c152994a31dc7d99b8eeac7784a919f71d1b306f4b83217e110fd3824c"
 dependencies = [
  "accesskit",
- "hashbrown 0.15.2",
+ "hashbrown 0.15.5",
 ]
 
 [[package]]
@@ -82,7 +82,7 @@ checksum = "692dd318ff8a7a0ffda67271c4bd10cf32249656f4e49390db0b26ca92b095f2"
 dependencies = [
  "accesskit",
  "accesskit_consumer",
- "hashbrown 0.15.2",
+ "hashbrown 0.15.5",
  "objc2 0.5.2",
  "objc2-app-kit 0.2.2",
  "objc2-foundation 0.2.2",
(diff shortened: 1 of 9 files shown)


Hennzau:
Hi, quick question, which env is passed to the **path** command in case of distributed dataflow? The CLI's, the Daemon's? which computer?

phil-opp:
> Hi, quick question, which env is passed to the **path** command in case of distributed dataflow? The CLI's, the Daemon's? which computer?

> As it is deserialised with the crate `serde-with-expand-env`, it can expand the env variables with cli's env. So that the original source of the env is where the dataflow.yml is deserialized.

Thanks for raising this! I think we should discuss whether we want to change this behavior. Right now, the dataflow is deserialized in the CLI, so all env variables are expanded there. However, this might not want we want if the binaries are run on a different machine. For example, consider the $HOME variable, which might be something like `/home/my-personal-user-name` on your CLI machine. This folder might not even exist on the target machine....

> In addition, it cannot use the env variables from the cli directly. The variables must "be declared” in the yaml file.
> available env variable:

This limitation might be a bit annoying, but it's probably a good starting point until we decide on where to expand env variables.

drindr:
Thanks for your response.

- In my understanding, in current implementation, the dataflow should be loaded on the host if the host runs the node, so do as local machine. So it is reasonable to extend them with CLI's env. Or are there any plan for loading dataflow from remote machine, or I missed something have achieved out.
- 
  > > In addition, it cannot use the env variables from the cli directly. The variables must "be declared” in the yaml file.
  > > available env variable:
  > 
  > This limitation might be a bit annoying, but it's probably a good starting point until we decide on where to expand env variables.

  I think it is not a limitation. It isolates the node from complex CLI env. It makes the env's expansion explicit.

- off-topic: maybe a method like the `dolar-expr` in XML launch of ROS or just a script programming language would make launching behaviour flexible

drindr:
Oh, I got it.
With the `_unstable_deploy`, nodes can be launched remotely.

If so, I have an idea is that every daemon maintains a parameter service and extend the variable like ${VAR_A} with CLI and parse ${REMOTE: VAR_A} with daemon's env.

What about your opinion? @phil-opp 
If you think it is available, I can start to implement it.

Hennzau:
> Oh, I got it. With the `_unstable_deploy`, nodes can be launched remotely.
> 
> If so, I have an idea is that every daemon maintains a parameter service and extend the variable like ${VAR_A} with CLI and parse ${REMOTE: VAR_A} with daemon's env.
> 
> What about your opinion? @phil-opp If you think it is available, I can start to implement it.

I thnk this is a good idea! maybe for another PR?

phil-opp:
> maybe for another PR?

Yes, it's probably a good idea to do smaller PRs. The quoted arguments change of this PR seems quite isolated and unrelated to the env expansion. Could you move these changes to a separate PR @drindr ? And apply the same quoting logic to the `args` field there: https://github.com/dora-rs/dora/blob/3197dc097d4b6c914a16bc54c4f2c6c82f8601c6/binaries/daemon/src/spawn.rs#L705

Thanks!

drindr:
> I like the idea too! Could you clarify the syntax that you propose?
> 
> These are the different variants that you mentioned in this thread:
> 
> ... env: DORA_HOME: $DORA WITH_DEFAULT: ${DEFAULT: FOO} BAR: /path/to/bar BAZ: ${REMOTE: VAR_A} ...
> 
> You propose that `$DORA` refers to a local environment variable (i.e. the machine where the CLI is used), right? The `${REMOTE: VAR_A}` is refering to the env variable on the target machine (i.e. expanded by the daemon), right? What is the `${DEFAULT: FOO}` syntax doing?

The syntax can be ${CLI: VAR_A} for CLI, which `CLI: ` can be omitted, and ${DAEMON: VAR_B} for DAEMON.
Maybe a default value can be provided by the assign expression ${VAR_A = DEFAULT_VALUE}

drindr:
The implementation of build command with quoted arguments have been modified and submit on #1103

haixuanTao:
> Thanks for clarifying, this sounds good to me in general. Instead of `CLI:` and `DAEMON:` we could also use `LOCAL:` and `REMOTE:`, maybe that's easier to understand.
> 
> @haixuanTao What do you think about this this change?

To be honest, I'm not really a fan of introducing a new concept of environment variable resolving while every one knows linux env variable resolving. Here are the reason why:
- We will need to implement it. And I can see so many edge cases, that we will have to handle: None env variable value, Empty string env variable value, naming collusion ...
- This introduce a new surface for input errors that can be very hard to debug.

I would very much favor using runtime environment variable within each node code ( including build ) for daemon env variable and, use configured env within the dataflow for cli based env variable.

I don't think that dora should be in charge of daemon -> to -> node env variable passing. 

Very sorry. 

Open for discussion still!

phil-opp:
> I would very much favor using runtime environment variable within each node code ( including build ) for daemon env variable and, use configured env within the dataflow for cli based env variable.

Could you clarify this a bit more? I don't understand what you mean with this.

phil-opp:
Thanks for clarifying! This seems reasonable to me.

So let's do it this way:

- Environment variables referenced in the `env` section are expanded by the CLI
    - e.g. `RUST_LOG: $RUST_LOG`
- Environment variables specified in the `path` field are expanded by the daemon

Does that sound good to you? Do we also want to support expanding env variables in `build` commands?

drindr:
The environment variable expansion for path field has been re-implemented in #1104.

PR #1103, opened 2025-08-20
Feat/build arg with space

drindr:
Allow argument with space in build command.
This PR is compatible with the old single string command.

```yaml
...
build:
    - exec: program-to-run
    - args:
        - -c
        - cargo build ...
... 
```

### DIFF
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -19,7 +19,7 @@ pub fn run_build_command(
 
     let lines = build.lines().collect::<Vec<_>>();
     for build_line in lines {
-        let mut split = build_line.split_whitespace();
+        let mut split = splitty::split_unquoted_whitespace(build_line).unwrap_quotes(true);
 
         let program = split
             .next()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3838,6 +3838,7 @@ dependencies = [
  "serde-with-expand-env",
  "serde_json",
  "serde_yaml 0.9.34+deprecated",
+ "splitty",
  "tokio",
  "tracing",
  "url",
@@ -14643,6 +14644,12 @@ dependencies = [
  "base64ct",
  "der 0.7.9",
 ]
+
+[[package]]
+name = "splitty"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2db70a1e6827e4d71c655b606caf1346862c38ae52ab4f58c32635e7c7aedd67"
 
 [[package]]
 name = "spm_precompiled"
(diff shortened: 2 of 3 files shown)


libraries/core/src/build/build_command.rs
```diff
@@ -5,34 +5,51 @@ use std::{
...
+                };
+                cmd.args(split);
+                cmd_vec.push(cmd);
+            }
+            cmd_vec
+        }
+        CommandConfig::WithArgs(configs) => {
+            let mut cmd_vec = vec![];
+            for CommandWithArgs { exec, args } in configs {
+                let mut cmd = Command::new(exec);
```
phil-opp: We should probably do the same `pip`/`uv` replacement here.

phil-opp (commented):
Thanks for the separate PR! Am I understanding it correctly that this new version does not support any quoting in the normal build string anymore, compared to #1098? Instead, we now have the option to specify the args as list, right?

The example in the PR description seems outdated, it's a list of `exec`/`args` pairs now, right?

phil-opp:
The new syntax seems quite verbose, I'm not sure if this is worth it. I think I liked the previous idea of supporting quoted substrings more.

How about we just use a crate such as [`splitty`](https://lib.rs/crates/splitty) for splitting the `build` lines and the `args` key? I think this would be more intuitive, as quoting substrings is quite common.

drindr:
> Thanks for the separate PR! Am I understanding it correctly that this new version does not support any quoting in the normal build string anymore, compared to #1098? Instead, we now have the option to specify the args as list, right?
> 
> The example in the PR description seems outdated, it's a list of `exec`/`args` pairs now, right?

Yes. Because string processing is complex and some corner cases quotation marks might not be considered.😭

A list makes it easy to handle and it is also commonly used in configuarion(like some settings for the VSCode)

haixuanTao:
I don't think we should use list as it would break what people know from configuration like docker / GitHub action...

drindr:
> The new syntax seems quite verbose, I'm not sure if this is worth it. I think I liked the previous idea of supporting quoted substrings more.
> 
> How about we just use a crate such as [`splitty`](https://lib.rs/crates/splitty) for splitting the `build` lines and the `args` key? I think this would be more intuitive, as quoting substrings is quite common. 

This crate looks good.
I will have a try.

drindr:
> I don't think we should use list as it would break what people know from configuration like docker / GitHub action...

Docker provided the [list form](https://docs.docker.com/reference/dockerfile/#shell-and-exec-form) as well. The implementation of the GitHub action seems to depend on a specific shell and execute the command with [the shell](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell). But in current Dora's build, it is not relied on any shell, which makes the functionalities of the build command limited.

PR #1104, opened 2025-08-20
feat/path-with-env

drindr:
Allow the path to be extended with daemon's environment variable.


haixuanTao:
Hey thanks for this PR!

Overall, I'm okay with this implementation. Have you looked into cross-platform compatibility?

It might actually be beneficial to keep the dataflow always unix like even on windows.

phil-opp (approved):
Looks good, thanks!

PR #1106, opened 2025-08-22
Feat: doraflow api

heyong4725:
* Title:  "feat(python): Add doraflow Pythonic API".

This PR introduces the new doraflow Pythonic API, providing an imperative and developer-friendly way to define dora dataflows directly within Python.

Key changes include:

   * New `doraflow` API: Implemented Dataflow, Node, and Operator classes for a Pythonic dataflow definition.
   * Comprehensive Documentation: Created detailed API documentation for doraflow to facilitate user adoption.
   * Updated Examples: Provided new and updated examples in the examples/doraflow directory, showcasing the API's capabilities, including a translation dataflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -409,7 +409,7 @@ jobs:
 
 
           # Test Dora Python Dataflow builder
-          python examples/python-dataflow-builder/simple_example.py
+          uv run examples/python-dataflow-builder/simple_example.py
 
           # Run Python queue latency test
           echo "Running CI Queue Latency Test"


apis/python/doraflow/doraflow/dataflow.py
```diff
@@ -0,0 +1,246 @@
...
+        """Adds an operator to this node."""
+        self.operators.append(operator)
+        return self
+
+    def add_input(self, input_id: str, source: str) -> "Node":
+        """Adds a user-defined input to the node."""
+        if "inputs" not in self.config:
+            self.config["inputs"] = {}
+        self.config["inputs"][input_id] = source
+        return self
```
phil-opp: This seems to be less powerful than the `add_edge` method. For example, it doesn't support setting a queue size. Any reason for that?

apis/python/doraflow/doraflow/dataflow.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def send(self, output_id: str) -> "Output":
```
phil-opp: This `send` name is a bit confusing. It just defines an output, doesn't it?

apis/python/doraflow/doraflow/dataflow.py
```diff
@@ -0,0 +1,246 @@
...
+        source = f"{self.node.id}/{self.output_id}"
+        if queue_size is not None:
+            node.config["inputs"][input_id] = {"source": source, "queue_size": queue_size}
+        else:
+            node.config["inputs"][input_id] = source
+
+        if "outputs" not in self.node.config:
+            self.node.config["outputs"] = []
+        if self.output_id not in self.node.config["outputs"]:
+            self.node.config["outputs"].append(self.output_id)
```
phil-opp: So the `output_id` is only added to the `outputs` list if there is a receiver? I think it would make more sense to do this part already in `send`. The node might send outputs that have no receivers, which should be still observable through zenoh (through `publish_all_messages_to_zenoh`).

apis/python/doraflow/doraflow/dataflow.py
```diff
@@ -0,0 +1,246 @@
...
+        self,
+        source_node: str | Node,
+        source_output: str,
+        condition: Callable[[any], str],
+        destinations: dict[str, str | Node],
+    ):
+        """Adds a conditional edge from a source node to multiple possible destination nodes."""
+        # This is a simplified implementation.
+        # A more complete implementation would create a router node.
+        print("Conditional edges are not fully implemented yet.")
```
phil-opp: What is this method about?

apis/python/doraflow/doraflow/dataflow.py
```diff
@@ -0,0 +1,246 @@
...
+        if path:
+            with open(path, "w") as f:
+                yaml.dump(dataflow_spec, f)
+            return None
+        else:
+            return yaml.dump(dataflow_spec)
+
+    def run(self):
+        """Generates the YAML file and runs the dataflow in the foreground."""
+        with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".yaml") as f:
```
phil-opp: Using a temporary file might lead to issues when using relative paths. Dora resolves relative paths from the parent folder of the dataflow YAML file, which is some temp directory in this case. This directory might be different on different systems.

The same concern applies to the working directory, which is set in the same way.

apis/python/doraflow/DORAFLOW_API_DESIGN.md
phil-opp, line 1: This file seems very similar to `docs/DATAFLOW_PYTHON_API.md`. What's the difference between the two doc files? This one is inside the `tests` folder, but does not really seem related to testing.

apis/python/doraflow/tests/test_dataflow.py
phil-opp, line 1: Are the new tests and examples run on our CI? If not, we should add that.

phil-opp (commented):
Thanks, looks useful! I like that it just generates a YAML file and invokes the normal `dora` CLI commands, so we don't need any special handling in dora itself.

I haven't tried the new APIs or examples yet, but I skimmed the code a bit. I noticed a few things that could be improved, see my inline comments.

README.md
```diff
@@ -142,27 +143,28 @@
 
 ## Examples
 
-| Type           | Title                                                                                                        | Description                                         | Last Commit                                                                                                        |
```
haixuanTao: FYI, this is just formatting

apis/python/doraflow/doraflow/dataflow.py
haixuanTao: Ok so this should be fixed as instead of adding edges, we know have an add_output function. I have remove add_edge and replaced it with add_input with a provided output

apis/python/doraflow/doraflow/dataflow.py
haixuanTao: This is removed

apis/python/doraflow/doraflow/dataflow.py
haixuanTao: I have removed this in favor of `from dora import run` which use run function from the cli

apis/python/doraflow/DORAFLOW_API_DESIGN.md
haixuanTao: The documentation is now within the code and will be generated within the website

apis/python/doraflow/tests/test_dataflow.py
haixuanTao: I have added a CI job

apis/python/node/src/lib.rs
```diff
@@ -376,13 +376,42 @@ pub fn resolve_dataflow(dataflow: String) -> eyre::Result<PathBuf> {
     Ok(dataflow)
 }
 
+/// Build a Dataflow locally
```
phil-opp:
```suggestion
/// Build a Dataflow
```

If a `coordinator_addr` is given and `force_local` is false, the build will go through the coordinator and might result in a distributed build.

phil-opp: Maybe we add something like "this behaves like `dora build`" here, so that people can read up on details?

examples/python-dataflow-builder/simple_example.py
```diff
@@ -0,0 +1,52 @@
...
+dataflow.to_yaml("dataflow.yml")
+print("Generated dataflow.yml")
+
+# If env var NO_BUILD is set, skip the build and run steps
+if not os.getenv("NO_BUILD"):
+    build("dataflow.yml", uv=True)
+else:
+    print("Skipping build and run steps due to NO_BUILD env var")
+print("Running dataflow.yml")
+run("dataflow.yml", uv=True)
```
phil-opp: The `run` step is not actually skipped, is it?

phil-opp (approved):
Looks good to me overall!

apis/python/node/src/lib.rs
haixuanTao: Yup, Indeed!

examples/python-dataflow-builder/simple_example.py
haixuanTao: Yeah sorry AI magic.

examples/python-dataflow-builder/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Usage
+
+Run the example:
+
+```bash
+python simple_example.py
+```
+
+To skip the build and run steps (useful for testing YAML generation only):
```
phil-opp: This also needs adjustment

PR #1111, opened 2025-08-27
Make nodes runnable standalone, prompting the user for inputs

### LINKED ISSUE #1019: A command for simplifying node testing

phil-opp:
Adds a new interactive mode to the Rust node API, which automatically activates when no `DORA_NODE_CONFIG` env variable is set. When run in interactive mode, the user is prompted for inputs on the command line. Input data can be given as string or as JSON.

This feature makes it possible to try a node with some inputs and see how it reacts, without the need to set up a full dataflow for it. This can be useful during development.

Partially resolves #1019

### DIFF
--- a/apis/rust/node/src/daemon_connection/interactive.rs
+++ b/apis/rust/node/src/daemon_connection/interactive.rs
@@ -1,11 +1,10 @@
 use std::{
-    io::{Read, stdout},
+    io::{BufRead, Read, stdout},
     sync::Arc,
     time::Duration,
 };
 
-use arrow::array::{Array, RecordBatch};
-use arrow_schema::{Field, Schema};
+use arrow::array::{Array, ArrayData};
 use colored::Colorize;
 use dora_core::{metadata::ArrowTypeInfoExt, uhlc::HLC};
 use dora_message::{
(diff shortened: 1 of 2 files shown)


haixuanTao:
Could you add example and documentation on how to use this?

haixuanTao:
yeah I meant like markdown based documentation and / or examples within the example folder on how to use this. It's not obvious to me what the input format should look like

haixuanTao:
Yeah we really need an individual markdown file for this on dora-rs.ai at least. 

Also could you provide an example using json?

phil-opp:
> Yeah we really need an individual markdown file for this on dora-rs.ai at least.

I opened https://github.com/dora-rs/dora-rs.github.io/pull/46 for this.

> Also could you provide an example using json?

Added in d2092ee.

PR #1132, opened 2025-09-03
Downgrade `opentelemetry` to `v0.29` again in `dora-metrics`

phil-opp:
Version `v0.4.3` of `opentelemetry-system-metrics`, which updated to `opentelemetry v0.30`, was yanked (as it violated semver). This causes some build errors when dataflows/nodes are created through `dora new` because those then use `opentelemetry-system-metrics v0.4.2` again, which requires `opentelemetry v0.29`.

This PR fixes this issue by downgrading the `opentelemetry` dependencies to v0.29 again (until `opentelemetry-system-metrics v0.5` is published).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1423,7 +1423,7 @@ dependencies = [
  "rustversion",
  "serde",
  "sync_wrapper",
- "tower",
+ "tower 0.5.2",
  "tower-layer",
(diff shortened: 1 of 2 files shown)


phil-opp:
Merging this without review to unblock our CI.

PR #1135, opened 2025-09-03
Auto-cancel previous CI jobs on new pushes

phil-opp:
Our CI takes quite a bit of time and our CI runners are limited. To minimize resource usage, this PR auto-cancels previous CI runs when new commits are pushed to a branch or pull request. So only the latest commit is run on the CI.

This does not apply for the `pip-release` job yet because that one also does releases, which we don't want to cancel.


phil-opp:
Unfortunately we still get notifications when a run is auto-canceled and there does not seem a way to ignore it: https://github.com/orgs/community/discussions/13015 . I guess we have to live with that until GitHub provides a way to silence notifications in this case.

PR #1139, opened 2025-09-09
Move node-hub to separate repo

phil-opp:
As discussed in our latest meeting.

This should only be merged after setting up the node hub repo.


haixuanTao:
looks good!

PR #1140, opened 2025-09-16
Fix build command execution

### LINKED ISSUE #1090: Venv deadlock in CI tests
We often get `venv` deadlocks in our CI. The result is that jobs hang at "Waiting to acquire lock for `/home/runner/work/dora/dora/.venv` at `/home/runner/work/dora/dora/.venv/.lock`".

These errors are spurious and often disappear when rerunning the job. But it is still very annoying, so we should fix this.

Mivik:
## Description

Prior to this PR, `run_build_command` contains wrong implementation which calls `cmd.spawn()` and `cmd.status()` sequentially, and since both invocations execute the command, each build command is **executed twice**. The issue is uncovered during my investigation of issue #1090 .

This PR addresses the issue by replacing `cmd.status()` with `child.wait()`. Additionally, it also changes the process backend from `std` to `tokio` to improve concurrency.

Since the issue has been resolved, #1137 is reverted to reduce verbosity of CI logs.

## Related

Fixes #1090

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
 
 env:
-  RUST_LOG: DEBUG
+  RUST_LOG: INFO
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
 
 env:
-  RUST_LOG: INFO
+  RUST_LOG: DEBUG
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
 
 env:
-  RUST_LOG: DEBUG
+  RUST_LOG: INFO
```
haixuanTao:
```suggestion
  RUST_LOG: DEBUG
```

haixuanTao (approved):

haixuanTao:
That makes total sense to me. Thanks!

PR #1141, opened 2025-09-16
catch dora failing to decode an env shell expand and error out instead of decoding it as a raw value

haixuanTao:
Previously if the user used `$NOT_AN_ENV_VARIABLE_KEY` within an env variable as part of its value, this would error out.

This PR makes it possible to not error out and just use this as a raw string instead.


haixuanTao:
Yeah indeed. That's a good point!

I was more thinking about the case with $ sign in env that was not corresponding to an env variable.

PR #1144, opened 2025-09-16
add model endpoint to openai proxy server

haixuanTao:
Small PR to support for model list endpoint as it is needed on certain server like moly.


phil-opp:
Cherry-picked to https://github.com/dora-rs/node-hub in https://github.com/dora-rs/node-hub/commit/7f6000670021f27c62c0405f9ff7a94d07f90d55

PR #1154, opened 2025-10-01
Bump dora to v0.3.13

haixuanTao:


Cargo.toml
```diff
@@ -41,38 +41,38 @@ members = [
...
+dora-cli = { version = "0.3.13", path = "binaries/cli" }
+dora-runtime = { version = "0.3.13", path = "binaries/runtime" }
+dora-daemon = { version = "0.3.13", path = "binaries/daemon" }
+dora-coordinator = { version = "0.3.13", path = "binaries/coordinator" }
+dora-ros2-bridge = { version = "0.3.13", path = "libraries/extensions/ros2-bridge" }
+dora-ros2-bridge-msg-gen = { version = "0.3.13", path = "libraries/extensions/ros2-bridge/msg-gen" }
 dora-ros2-bridge-python = { path = "libraries/extensions/ros2-bridge/python" }
 # versioned independently from the other dora crates
-dora-message = { version = "0.5.0", path = "libraries/message" }
+dora-message = { version = "0.6.0", path = "libraries/message" }
```
dieu-detruit, line 75: Thank you for dora-rs devleopment.

This change causes imcompatible message version via old nodes and new nodes like:
```
xxxx: stdout       2: version mismatch: message format v0.5.0 is not compatible with expected message format v0.6.0
```


I think it would be better if this change is done in a major or minor version up, not in a patch version up!

haixuanTao: I understand. I think in the future, we'll do minor version up to keep semver compatibility. Sorry!

DO you need help with upgrading nodes?

dieu-detruit: I fixed this just by upgrading all the nodes and dora-cli!

I've just commented on this for better ecosystem management☺️
Thank you.

PR #1159, opened 2025-10-11
Enhance `dora list` command

Mivik:
## Motivation

If we align our CLI with established tools (e.g., Docker), the workflow will feel more seamless and familiar to users.

## Description

This PR implements `--format` for `dora list`, allowing users to specify the command's output format. Currently `table` and `json` formats are supported. In the future, this option could extend to more commands.

Additionally, `dora ps` is added as an alias to `dora list`.

This PR is backward compatible.

### DIFF
--- a/binaries/cli/src/command/mod.rs
+++ b/binaries/cli/src/command/mod.rs
@@ -46,6 +46,7 @@ pub enum Command {
     Destroy(Destroy),
     Start(Start),
     Stop(Stop),
+    #[clap(alias = "ps")]
     List(ListArgs),
     // Planned for future releases:
     // Dashboard,


binaries/cli/src/command/list.rs
```diff
@@ -31,29 +37,49 @@ impl Executable for ListArgs {
...
-    }
-    tw.flush()?;
-    let formatted = String::from_utf8(tw.into_inner()?)?;
+    let entries = list.0.into_iter().map(|entry| OutputEntry {
+        uuid: entry.id.uuid,
+        name: entry.id.name.unwrap_or_default(),
+        status: entry.status,
+    });
 
-    println!("{formatted}");
```
haixuanTao, line 56: Don't we need this print?

Mivik: The old implementation writes content into a string buffer and then prints it all at once. I changed the printing target to direct output to stdout.

haixuanTao: Okay indeed.

PR #1163, opened 2025-10-15
Add support for node integration testing

phil-opp:
## How it works

- Set the `DORA_TEST_WITH_INPUTS` env variable with the path to your input JSON file
- (Optional) Set the `DORA_TEST_WRITE_OUTPUTS_TO` env variable with the path where the outputs should be written. If not set, dora will write a `outputs.jsonl` file next to the given inputs file
- Start the node executable/script

The node will be run as usual, but its event channel will be filled from the given inputs JSON file. No connection to a `dora daemon` will be made.

## Input JSON file format example:

```json5
{
     // ID of the node
    "id": "foo",
    // defines the events that the node should receive
    "events": [
        {
            // specifies when the event arrives (seconds since node start)
            "time_offset_secs": 0.7,
            // type of the event (supported types are `Input`, `Stop`, `InputClosed`, `AllInputsClosed`
            "type": "Input",
            // input ID
            "id": "tick"
            // optional: `data` field with input data
        },
        {
            "time_offset_secs": 0.9,
            "type": "Input",
            "id": "tick"
        },
        {
            "time_offset_secs": 1.2,
            "type": "Stop"
        }
    ]
    // other supported fields: name, description, args, env, outputs, inputs, send_stdout_as (they all behave like in dataflow.yaml)
}
```

## Output JSON file format example:

```jsonl
{"id":"random","data":9267023440904143729,"time_offset_secs":0.700793541}
{"id":"random","data":5753749540645363621,"time_offset_secs":0.900897584}
```

## TODO:

- [ ] Documentation
  - [x] API docs
  - [ ] dora-rs.ai docs
- [x] add some tests for our examples and use them on our CI
- [x] take a look at [`arrow_integration_test`](https://docs.rs/arrow-integration-test/latest/arrow_integration_test/) JSON format -> this might be better suitable than our custom input json format
- [x] add option (via env variable) to write out received inputs as inputs.json files during normal dataflow operation -> to make creating files with complex input data easier
- [x] add option (via env variable) to omit time offsets in output formats -> to make them diff-able with expected outputs (the time offsets are a bit different on each run)
- [x] ~use ArrowTestUnwrap format for outputs.jsonl~ (not possible because of https://github.com/apache/arrow-rs/issues/8684)
  - [x] instead: Include data type in output JSON file

### DIFF
--- a/apis/rust/node/src/integration_testing.rs
+++ b/apis/rust/node/src/integration_testing.rs
@@ -1,16 +1,84 @@
+#![allow(clippy::test_attr_in_doctest)]
+
 //! This document describes how to write [integration tests] for Dora nodes.
 //!
 //! [integration tests]: https://en.wikipedia.org/wiki/Integration_testing
 //!
 //! # Usage
 //!
-//! There are currently two ways to run integration tests for Dora nodes: testing the whole
-//! executable through environment variables, or writing custom test functions within the node.
+//! There are currently three ways to run integration tests for Dora nodes:
 //!
-//! ## Testing the full executable
+//! - calling [`setup_integration_testing`] before invoking the node's `main` function
+//! - testing the compiled executable by setting environment variables
+//! - using [`DoraNode::init_testing`](crate::DoraNode::init_testing) tests independent
+//!   of the `main` function
 //!
-//! Testing the full executable ensures that the node's `main` function produces the expected
-//! outputs for a given set of inputs. To enable this, the
+//!
+//! ## Using `setup_integration_testing`
+//!
+//! The most straightforward way to write integration tests for a Dora node is to call the
+//! [`setup_integration_testing`] function in the test function and then invoke the node's `main`
+//! function. This way, the full behavior of the node (including the `main` function) is tested.
+//!
+//! This approach requires that the node's `main` function uses the
+//! [`DoraNode::init_from_env`](crate::DoraNode::init_from_env) function for initialization.
+//!
+//! See the [`setup_integration_testing`] function documentation for details.
+//!
+//! ### Example
+//! ```rust, ignore
+//! #[test]
+//! fn test_main_function() -> eyre::Result<()> {
+//!    // specify the events that should be sent to the node
+//!    let events = vec![
+//!        TimedIncomingEvent {
+//!            time_offset_secs: 0.01,
+//!            event: IncomingEvent::Input {
+//!                id: "tick".into(),
+//!                metadata: None,
+//!                data: None,
+//!            },
+//!        },
+//!        TimedIncomingEvent {
+//!            time_offset_secs: 0.055,
+//!            event: IncomingEvent::Stop,
+//!        },
+//!    ];
+//!    let inputs = dora_node_api::integration_testing::TestingInput::Input(
+//!        IntegrationTestInput::new("node_id".parse().unwrap(), events),
+//!    );
+//!
+//!    // send the node's outputs to a channel so we can verify them later
+//!    let (tx, rx) = flume::unbounded();
+//!    let outputs = dora_node_api::integration_testing::TestingOutput::ToChannel(tx);
+//!
+//!    // don't include time offsets in the outputs to make them deterministic
+//!    let options = TestingOptions {
+//!        skip_output_time_offsets: true,
+//!    };
+//!
+//!     // setup the integration testing environment -> this will make `DoraNode::init_from_env``
+//!     // initialize the node in testing mode
+//!     integration_testing::setup_integration_testing(inputs, outputs, options);
+//!
+//!     // call the node's main function
+//!     crate::main()?;
+//!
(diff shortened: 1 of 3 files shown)


phil-opp:
I opened https://github.com/apache/arrow-rs/pull/8737 to add support for binary decoding to `arrow-json`.

haixuanTao:
> There are two tests for the `rust-dataflow` example in https://github.com/dora-rs/dora/pull/1163/files#diff-c1dad61976e06858147411d3a1b21177dc82b8eda35818b881c08a67add4cdfd

I find the test example very difficult to decipher. Could we make them easier to understand?

haixuanTao:
> > > There are two tests for the `rust-dataflow` example in https://github.com/dora-rs/dora/pull/1163/files#diff-c1dad61976e06858147411d3a1b21177dc82b8eda35818b881c08a67add4cdfd
> > 
> > 
> > I find the test example very difficult to decipher. Could we make them easier to understand?
> 
> Which part do you find hard to decipher?
> 
> We just call two commands for each test. One build command and one run command that sets some env variables. So basically:
> 
> * `cargo build -p rust-dataflow-example-node`
> * `DORA_TEST_WITH_INPUTS=tests/sample-inputs/inputs-rust-node.json DORA_TEST_NO_OUTPUT_TIME_OFFSET=1 DORA_TEST_WRITE_OUTPUTS_TO=<tempfile> target/debug/rust-dataflow-example-node`
> 
> To make this extensible, I introduced a `test` function that takes the variable parameters as input.
> 
> The other code in that file is necessary to account for the differences between Windows and Linux (e.g. exe extension, line endings) and make it work with the `CARGO_TARGET_DIR` env variable that we set on CI for Windows.
> 
> I'm not sure how we can simplify this more?

I don't think that as a user of dora, I would be able to reproduce the node test on my own on my own codebase.

Would it be possible to make node integration test be something like 

```rust
#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn test_log_message_serialization() {
        let input_json = ...
        let output_json = ...
        
        // Maybe add them to env variable 
        // ... 

        let result = main();

        assert_eq! ...
  }
}
```

So that it's easier to read but also use native testing that can be used with: `cargo test -p rust-dataflow-example-node`

phil-opp:
I pushed another improvement in 68723bbc. There is now a `setup_integration_testing` function that will store the testing config in thread local data. The `init_from_env` function checks for this thread-local data and enters integration testing mode if set.

This means that you can now test the `main` function like this:

```rust
#[test]
fn test_main_function() -> eyre::Result<()> {
    let inputs = dora_node_api::integration_testing::TestingInput::FromJsonFile(
        "../../../tests/sample-inputs/inputs-rust-node.json".into(),
    );
    let mut output_file = Arc::new(tempfile::tempfile()?);
    let testing_output =
        dora_node_api::integration_testing::TestingOutput::ToWriter(Box::new(output_file.clone()));
    let options = TestingOptions {
        skip_output_time_offsets: true,
    };

    integration_testing::setup_integration_testing(inputs, testing_output, options);

    crate::main()?;

    let mut output = String::new();
    output_file.seek(std::io::SeekFrom::Start(0))?;
    output_file.read_to_string(&mut output)?;
    let expected =
        std::fs::read_to_string("../../../tests/sample-inputs/expected-outputs-rust-node.jsonl")?;

    assert_eq!(output, expected.replace("\r\n", "\n")); // normalize line endings

    Ok(())
}
```

haixuanTao:
> I pushed another improvement in [68723bb](https://github.com/dora-rs/dora/commit/68723bbc86ebd5b6543de95815ad2cb9f32feab2). There is now a `setup_integration_testing` function that will store the testing config in thread local data. The `init_from_env` function checks for this thread-local data and enters integration testing mode if set.
> 
> This means that you can now test the `main` function like this:
> 
> ```rust
> #[test]
> fn test_main_function() -> eyre::Result<()> {
>     let inputs = dora_node_api::integration_testing::TestingInput::FromJsonFile(
>         "../../../tests/sample-inputs/inputs-rust-node.json".into(),
>     );
>     let mut output_file = Arc::new(tempfile::tempfile()?);
>     let testing_output =
>         dora_node_api::integration_testing::TestingOutput::ToWriter(Box::new(output_file.clone()));
>     let options = TestingOptions {
>         skip_output_time_offsets: true,
>     };
> 
>     integration_testing::setup_integration_testing(inputs, testing_output, options);
> 
>     crate::main()?;
> 
>     let mut output = String::new();
>     output_file.seek(std::io::SeekFrom::Start(0))?;
>     output_file.read_to_string(&mut output)?;
>     let expected =
>         std::fs::read_to_string("../../../tests/sample-inputs/expected-outputs-rust-node.jsonl")?;
> 
>     assert_eq!(output, expected.replace("\r\n", "\n")); // normalize line endings
> 
>     Ok(())
> }
> ```

ok this works for me.

haixuanTao (approved):

PR #1168, opened 2025-10-22
Add local listener for run dataflow function to make it possible to use dynamic node with run command

haixuanTao:
This PR adds the possibility to use dynamic nodes with run command.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -185,18 +185,6 @@ impl Daemon {
         uv: bool,
         log_destination: LogDestination,
     ) -> eyre::Result<DataflowResult> {
-        // Spawn local listener for dynamic nodes
-        let (events_tx, events_rx) = flume::bounded(10);
-        let _listen_port = local_listener::spawn_listener_loop(
-            (LOCALHOST, DORA_DAEMON_LOCAL_LISTEN_PORT_DEFAULT).into(),
-            events_tx,
-        )
-        .await?;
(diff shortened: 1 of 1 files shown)


binaries/daemon/src/lib.rs
```diff
@@ -185,6 +185,18 @@ impl Daemon {
         uv: bool,
         log_destination: LogDestination,
     ) -> eyre::Result<DataflowResult> {
+        // Spawn local listener for dynamic nodes
+        let (events_tx, events_rx) = flume::bounded(10);
+        let _listen_port = local_listener::spawn_listener_loop(
+            (LOCALHOST, DORA_DAEMON_LOCAL_LISTEN_PORT_DEFAULT).into(),
+            events_tx,
+        )
+        .await?;
```
phil-opp: This means that `dora run` will now fail if there is any `dora daemon` or other `dora run` running concurrently. I'm not sure if this limitation is worth it...

Maybe we can choose a free port instead and make the port number used in `init_from_node_id` overridable (e.g. through an env variable)? Then we could make `dora run` print the port number to the terminal and the user could set the env variable accordingly.

haixuanTao: Maybe we should fail fast and let the user to use dora destroy if there is an actual dynamic node.

phil-opp: Decision from today's meeting: Only spawn a listener if the dataflow contains dynamic nodes. This way you can still run multiple `dora run` commands in parallel, as long as they don't contain any dynamic nodes. This is e.g. useful for testing multiple dataflows in parallel.

PR #1169, opened 2025-10-23
feat: c++ arrow interface with id and metadata

TomCC7:
Adds a new function to dora C++ interface to make retrieving the complete information (id, metadata, arrow data) possible. Metadata will be serialized to json.

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -53,6 +53,9 @@ mod ffi {
         Integer,
         Float,
         String,
+        ListInt,
+        ListFloat,
+        ListString,
     }
 
     pub struct CombinedEvents {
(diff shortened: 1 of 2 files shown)


TomCC7:
I think it's impossible to return an `std::unordered_map` directly with rust bindings but I'll see if there's any workaround. The other approach is perhaps creating an accessor function so that we can do something like
```cxx
auto metadata = extract_metadata();
std::optional<> val = metadata.get(key);
```

TomCC7:
@haixuanTao This new approach now provides access to metadata dict through getter and setter method. Due to constraints of the FFI interface I have to implement all the metadata types manually. If we want better interface, we can add a wrapper from C++ side to wrap the raw interface into template functions.

PR #1173, opened 2025-10-28
fix: reduce CPU usage by replacing busy loops with blocking/async waits

RustyCoderX:
constantly checking for new task now code waits for events by  blocking and async receives this  reduces CPU usage while keeping everything working the same

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ members = [
 ]
 
 [workspace.package]
-edition = "2024"
+edition = "2021"
 rust-version = "1.85.0"
 # Make sure to also bump `apis/node/python/__init__.py` version.
 version = "0.3.13"
--- a/apis/rust/node/src/event_stream/thread.rs
+++ b/apis/rust/node/src/event_stream/thread.rs
@@ -7,11 +7,8 @@ use dora_message::{
     node_to_daemon::{DaemonRequest, DropToken, Timestamped},
 };
 use eyre::{Context, eyre};
-use flume::RecvTimeoutError;
-use std::{
-    sync::Arc,
-    time::{Duration, Instant},
-};
+use flume::RecvError;
+use std::sync::Arc;
 
 use crate::daemon_connection::DaemonChannel;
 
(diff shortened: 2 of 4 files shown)


apis/rust/node/src/event_stream/thread.rs
```diff
@@ -68,17 +65,14 @@ impl Drop for EventStreamThreadHandle {
         //
         // In the future, we hope to fix this issue so that
         // the event stream can be properly waited for every time.
-        match self.handle.recv_timeout(Duration::from_secs(1)) {
+        match self.handle.recv() {
```
phil-opp, line 68: Why did you remove this timeout? We don't want to block indefinetly in the Drop handler when there is something wrong with the cleanup.

apis/rust/node/src/node/mod.rs
```diff
@@ -684,13 +684,12 @@ impl DoraNode {
...
                 Ok(token) => match self.sent_out_shared_memory.remove(&token) {
                     Some(region) => self.add_to_cache(region),
                     None => tracing::warn!("received unknown finished drop token `{token:?}`"),
                 },
-                Err(flume::TryRecvError::Empty) => break,
-                Err(flume::TryRecvError::Disconnected) => {
+                Err(flume::RecvError::Disconnected) => {
                     bail!("event stream was closed before sending all expected drop tokens")
                 }
             }
```
phil-opp, line 695: We don't want to wait here, we just want to handle all buffered events the in the stream. This was not a busy wait loop since we used to `break` on the first `Empty` error. So I don't see how this change helps.

apis/rust/node/src/node/mod.rs
```diff
@@ -746,26 +745,18 @@ impl Drop for DoraNode {
                 );
             }
 
-            match self.drop_stream.recv_timeout(Duration::from_secs(2)) {
+            match self.drop_stream.recv() {
```
phil-opp, line 748: Again, why are we removing the timeout here? We have that as a failsafe in case that something behaves unusual. In general, we don't want to block indefinetly in Drop implementations since the program might not able to exit anymore then.

binaries/daemon/src/node_communication/mod.rs
```diff
@@ -303,7 +303,7 @@ impl Listener {
 
     async fn handle_events(&mut self) -> eyre::Result<()> {
         if let Some(events) = &mut self.subscribed_events {
-            while let Ok(event) = events.try_recv() {
+            while let Ok(event) = events.recv().await {
```
phil-opp, line 306: We don't want to wait for incoming events here. The purpose of this function is to move all events that are buffered in the stream into `self.queue`. This was not a busy loop since `try_recv` returns an `Err` if no events are ready.

Cargo.toml
```diff
@@ -38,7 +38,7 @@ members = [
 ]
 
 [workspace.package]
-edition = "2024"
+edition = "2021"
```
phil-opp, line 41: What's the reason for going back to the old 2021 edition?

phil-opp:
Closing this for now.

In the current state, this PR does some questionable things and there was no reply to my reviews in the last 3 weeks.

PR #1177, opened 2025-11-02
Python: Add drain method to get all events within the dora queue

haixuanTao:
In order to reduce having too many filled queues., node.drain() enable to get all event from all input as a vec/list.

In order to make this fast and manageable the order of draining is going to be first ordered by topic and then by time of arrival

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -427,11 +427,14 @@ jobs:
           sleep 5
 
           # Run Python async
-          echo "Running CI Queue Latency Test"
-          dora run examples/python-async/dataflow.yaml --uv
+          echo "Running CI Async test"
+          ## TODO: Fix async Windows support
+          if [ "$RUNNER_OS" != "Windows" ]; then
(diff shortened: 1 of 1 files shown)


apis/rust/node/src/event_stream/mod.rs
```diff
@@ -280,6 +280,43 @@ impl EventStream {
...
+                } else {
+                    break;
+                }
+            } else {
+                match select(Delay::new(Duration::from_micros(300)), self.receiver.next()).await {
+                    Either::Left((_elapsed, _)) => break,
+                    Either::Right((Some(event), _)) => self.scheduler.add_event(event),
+                    Either::Right((None, _)) => break,
+                };
+            }
```
phil-opp: Could you clarify what you're trying to do here? It looks like this code will just block until the event channel is closed?

phil-opp:
I just opened https://github.com/dora-rs/dora/pull/1185 to provied a simpler `drain` method based on `try_recv`.

PR #1178, opened 2025-11-02
Add `dora topic` subcommand

Mivik:
## Description

This PR introduces a `dora topic` subcommand to inspect node outputs, inspired by [`rostopic`](https://wiki.ros.org/rostopic).

Three subcommands are included:
 - `topic list`: Lists all available topics in a dataflow. Both table and JSON formats are supported.
 - `topic echo`: Subscribes to a topic and prints the received data to the console. Uses `arrow-json` to serialize arrow messages into JSON strings. Both table and JSON formats are supported.
 - `topic hz`: Subscribes to a topic and displays the frequency of messages. Uses `ratatui` to render realtime table TUI.

## Backward Compatibility

The coordinator has been updated to store dataflow descriptors and provide them to the CLI.

### DIFF
--- a/binaries/cli/src/command/topic/hz.rs
+++ b/binaries/cli/src/command/topic/hz.rs
@@ -6,7 +6,7 @@ use itertools::Itertools;
 use ratatui::{DefaultTerminal, prelude::*, widgets::*};
 use std::{
     borrow::Cow,
-    collections::VecDeque,
+    collections::{BTreeSet, VecDeque},
     iter,
     net::IpAddr,
     sync::{Arc, Mutex},
@@ -163,7 +163,7 @@ async fn run_hz(
     mut terminal: DefaultTerminal,
     window: usize,
     dataflow_id: Uuid,
-    outputs: Vec<TopicIdentifier>,
+    outputs: BTreeSet<TopicIdentifier>,
     coordinator_addr: IpAddr,
 ) -> eyre::Result<()> {
     let stats = outputs
--- a/binaries/cli/src/command/topic/selector.rs
+++ b/binaries/cli/src/command/topic/selector.rs
@@ -1,4 +1,8 @@
-use std::{borrow::Cow, collections::HashMap, fmt};
+use std::{
+    borrow::Cow,
+    collections::{BTreeSet, HashMap},
+    fmt,
+};
 
 use crate::common::resolve_dataflow_identifier_interactive;
 use communication_layer_request_reply::TcpRequestReplyConnection;
@@ -69,7 +73,7 @@ impl TopicSelector {
     pub fn resolve(
         &self,
         session: &mut TcpRequestReplyConnection,
-    ) -> eyre::Result<(DataflowId, Vec<TopicIdentifier>)> {
+    ) -> eyre::Result<(DataflowId, BTreeSet<TopicIdentifier>)> {
         let (dataflow_id, dataflow_descriptor) = self.dataflow.resolve(session)?;
         if !dataflow_descriptor.debug.publish_all_messages_to_zenoh {
             bail!(
@@ -90,7 +94,7 @@ impl TopicSelector {
             .map(|node| (&node.id, node))
             .collect::<HashMap<_, _>>();
 
-        let mut data = Vec::new();
+        let mut data = BTreeSet::new();
         if self.data.is_empty() {
             data.extend(dataflow_descriptor.nodes.iter().flat_map(|node| {
                 node.outputs.iter().map(|output| TopicIdentifier {
@@ -98,7 +102,6 @@ impl TopicSelector {
                     data_id: output.clone(),
                 })
             }));
-            data.sort();
             return Ok((dataflow_id, data));
         }
 
@@ -120,7 +123,7 @@ impl TopicSelector {
                             data_id: output.clone(),
                         }));
                     } else if node.outputs.contains(&user.output) {
-                        data.push(TopicIdentifier {
+                        data.insert(TopicIdentifier {
                             node_id: user.source,
                             data_id: user.output,
                         });
@@ -138,9 +141,6 @@ impl TopicSelector {
                 Err(e) => bail!("Invalid output id `{s}`: {e}"),
             }
         }
-
-        data.sort();
-        data.dedup();
 
         Ok((dataflow_id, data))
     }


libraries/core/Cargo.toml
```diff
@@ -34,3 +35,4 @@ url = { version = "2.5.4", optional = true }
 git2 = { workspace = true, optional = true }
 fs_extra = "1.3.0"
 splitty = "1.0.2"
+zenoh = { workspace = true, optional = true }
```
haixuanTao, line 38: Could be good to put it somewhere else than core because every time we modify core, including dependency we add breaking changes and cannot make backward compatibility as it is tied to dora messages

haixuanTao: I think there might be a bigger issue that is to separate dora-core and dora-messages @phil-opp

haixuanTao:
Starting to look great! 

Could we format the documentation a little bit better though:

```bash
dora topic hz --help
```

Gives me:

```bash
Examples: - Measure a single topic: `dora topic hz -d my-dataflow robot1/pose` - Measure
multiple topics with a short window: `dora topic hz -d my-dataflow robot1/pose robot2/vel
--window 5` - Measure all topics: `dora topic hz -d my-dataflow --window 10`

Note: Requires `_unstable_debug.publish_all_messages_to_zenoh: true` in the dataflow
descriptor so runtime messages are available for inspection.
```

is really hard to read.

Could we also use:

```

_unstable_debug:
  publish_all_messages_to_zenoh: true
```

Instead of `_unstable_debug.publish_all_messages_to_zenoh: true`

libraries/core/Cargo.toml
phil-opp: It's the other way around: The underlying crate is `dora-message`, which defines the message formats used for communicating between the different components. Whenever we update `dora-message` in a backwards-incompatible way, we need to bump it's minor version to avoid parsing errors. This then requires updating all Dora components together.

phil-opp: So this change is fine with me. It's even behind an optional feature, so it's also included if exclititly enabled.

libraries/message/src/cli_to_coordinator.rs
```diff
@@ -64,12 +64,14 @@ pub enum ControlRequest {
         grace_duration: Option<Duration>,
     },
     Logs {
-        uuid: Option<Uuid>,
-        name: Option<String>,
-        node: String,
+        uuid: Uuid,
+        node: NodeId,
     },
```
phil-opp, line 70: Can we keep this message as-is for backwards compatibility? Otherwise we would have to bump the `dora-message` version to the next minor version, which requires updating all nodes, the coordinator, and cli in sync.

libraries/message/src/cli_to_coordinator.rs
```diff
@@ -64,12 +64,14 @@ pub enum ControlRequest {
...
     Destroy,
     List,
+    Info {
+        dataflow_uuid: Uuid,
+    },
```
phil-opp, line 75: I would count this change as backwards-compatible since it's a new command. The behavior would be the following:

- if only CLI is updated: coodinator sends a `ControlRequestReply::Error` back, which will contain the deserialization error. Not nice, but ok.
- if only coordinator is updated: fine, as the CLI never sends the new message
- if both CLI and coordinator are updated: fine

libraries/message/src/coordinator_to_cli.rs
```diff
@@ -33,6 +33,11 @@ pub enum ControlRequestReply {
         result: DataflowResult,
     },
     DataflowList(DataflowList),
+    DataflowInfo {
+        uuid: Uuid,
+        name: Option<String>,
+        descriptor: Descriptor,
+    },
```
phil-opp, line 40: I would also count this change as backwards compatible since the coordinator only sends this in reply to the new `Info` command

binaries/coordinator/src/lib.rs
```diff
@@ -634,32 +634,30 @@ async fn start_inner(
                                 let _ = reply_sender.send(Err(err));
                             }
                         },
-                        ControlRequest::Logs { uuid, name, node } => {
-                            let dataflow_uuid = if let Some(uuid) = uuid {
-                                Ok(uuid)
-                            } else if let Some(name) = name {
-                                resolve_name(name, &running_dataflows, &archived_dataflows)
+                        ControlRequest::Logs { uuid, node } => {
```
phil-opp: As noted in my other comment, we should keep this command as-is for backwards compatibility.

binaries/cli/src/common.rs
```diff
@@ -50,6 +50,33 @@ pub(crate) fn query_running_dataflows(
     Ok(ids)
 }
 
+pub(crate) fn resolve_dataflow_identifier(
```
phil-opp: Can we add an `interactive` suffix to this function to make it explicit that it might prompt the user for selection:

```suggestion
pub(crate) fn resolve_dataflow_identifier_interactive(
```

binaries/cli/src/command/topic/echo.rs
```diff
@@ -0,0 +1,264 @@
...
+/// Echo multiple topics:
+///   dora topic echo -d my-dataflow robot1/pose robot2/vel
+///
+/// Emit JSON lines:
+///   dora topic echo -d my-dataflow robot1/pose --format json
+///
+/// Note: The dataflow descriptor must include the following snippet so that
+/// runtime messages can be inspected:
+///
+/// ```
```
phil-opp:
```suggestion
/// ```yaml
```

to add syntax highlighting and tell the Rust compiler that this isn't a doctest

binaries/cli/src/command/topic/selector.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        for s in &self.data {
+            let mut s = Cow::Borrowed(s.as_str());
+            if !s.contains('/') {
+                s.to_mut().push('/');
+            }
+            match serde_json::from_value::<InputMapping>(serde_json::Value::String(
+                s.clone().into_owned(),
+            )) {
```
phil-opp, line 115: How about using `serde_json::from_str` instead?

binaries/cli/src/command/topic/selector.rs
```diff
@@ -0,0 +1,161 @@
...
+                }
+                Ok(_) => {
+                    bail!("Reserved input mapping cannot be inspected")
+                }
+                Err(e) => bail!("Invalid output id `{s}`: {e}"),
+            }
+        }
+
+        data.sort();
+        data.dedup();
```
phil-opp: We could make `data` a `BTreeSet`, then we could skip these steps.

binaries/cli/src/command/topic/selector.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// Identifier of the dataflow
+    #[clap(long, short, value_name = "UUID_OR_NAME")]
+    pub dataflow: Option<String>,
+    /// Address of the dora coordinator
+    #[clap(long, value_name = "IP", default_value_t = LOCALHOST)]
+    pub coordinator_addr: IpAddr,
+    /// Port number of the coordinator control server
+    #[clap(long, value_name = "PORT", default_value_t = DORA_COORDINATOR_PORT_CONTROL_DEFAULT)]
+    pub coordinator_port: u16,
+}
```
phil-opp, line 24: It feels a bit weird to group these args. Can we separate the coordinator addr/port arguments into an extra struct?

binaries/cli/src/command/topic/selector.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl DataflowSelector {
+    pub fn resolve(&self) -> eyre::Result<(Box<TcpRequestReplyConnection>, Uuid, Descriptor)> {
```
phil-opp: Having the session as one of the return times is a bit cumbersome. I think it would be cleaner to instead take a `&mut TcpRequestReplyConnection` argument. Then we could also call this function multiple times (if required in the future) without reconnecting to the coordinator every time.

binaries/cli/src/command/topic/selector.rs
```diff
@@ -0,0 +1,161 @@
...
+        write!(f, "{}/{}", self.node_id, self.data_id)
+    }
+}
+
+impl TopicSelector {
+    pub fn resolve(
+        &self,
+    ) -> eyre::Result<(
+        Box<TcpRequestReplyConnection>,
+        (DataflowId, Vec<TopicIdentifier>),
```
phil-opp: This function should take a `&mut TcpRequestReplyConnection` argument instead of initializing a session and returning it. This way, the return type becomes simpler and the function can be called multiple times.

binaries/cli/src/command/topic/list.rs
```diff
@@ -0,0 +1,110 @@
...
+    name: DataId,
+    subscribers: Vec<String>,
+}
+
+fn list(selector: DataflowSelector, format: OutputFormat) -> eyre::Result<()> {
+    let (_session, _dataflow_id, descriptor) = selector.resolve()?;
+
+    let mut subscribers = BTreeMap::<(&NodeId, &DataId), Vec<(&NodeId, &DataId)>>::new();
+    for node in &descriptor.nodes {
+        for (data, input) in &node.inputs {
```
phil-opp: We use the term `input_id` in the rest of the code, so could you adjust this here too?

```suggestion
        for (input_id, input) in &node.inputs {
```

binaries/cli/src/command/topic/hz.rs
```diff
@@ -0,0 +1,284 @@
...
+/// Measure multiple topics with a short window:
+///   dora topic hz -d my-dataflow robot1/pose robot2/vel --window 5
+///
+/// Measure all topics:
+///   dora topic hz -d my-dataflow --window 10
+///
+/// Note: The dataflow descriptor must include the following snippet so that
+/// runtime messages can be inspected:
+///
+/// ```
```
phil-opp:
```suggestion
/// ```yaml
```

phil-opp:
I'm not sure if `topic` is the best name here. I understand that this is because it's inspired by `rostopic`, but we typically don't talk about "topics" in Dora. Maybe something like `dora outputs` would fit our terminology better? E.g. `dora outputs echo` and `dora outputs hz`?

binaries/cli/src/command/topic/selector.rs
Mivik: `from_str` would expect a double-quoted JSON string

phil-opp: Ah right. We should really implement `FromStr` for `InputMapping`. This is separate from this PR though. I opened https://github.com/dora-rs/dora/issues/1191 for it.

phil-opp (approved):
Thanks for the updates!

PR #1182, opened 2025-11-05
Generate separate mod files in the ros2 msg-gen for better compiling time

drindr:
This PR generate separate mod files for ros2 msg and reduce the cross-reference between mods.

Tested on my laptop with `--timings` flag, the 1st compiling time of dora-ros2-bridge is 2\~3 secs slower than the single file implementation.
And as I source a new ros2 interfaces package, with incremental compiling, the compiling time is 6\~8 secs faster than the original one.

The change of code-generate time is less than 1 sec according to the report.

In the 1st compilation, I just source the system's ros2 workspace.
In the 2nd compilation, I add a new customize interface package.
In the 3rd compilation, I remove the customize interface package added in the 2nd compilation.
|  | single file | multi-file |
|--|---------------| --------------|
| 1st  | 35.19s |  38.19s    |      
| 2nd  | 35.45s |  29.12s    |          
| 3rd  | 37.02s |  28.8s    |        

And for c++ users, they can just compile and include the specific header they need, which increases the complexity (more flags) but can reduce the c++'s compiling time (not tested yet).

The compiling time of ros2-bridge is still very long. I'm not sure if it's still worthy to push ahead.

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -255,6 +255,58 @@ impl EventStream {
         event.map(Self::convert_event_item)
     }
 
+    /// Receives the next buffered [`Event`] (if any) without blocking, using an
+    /// [`EventScheduler`] for fairness.
+    ///
+    /// Returns [`TryRecvError::Empty`] if no event is available right now.
+    /// Returns [`TryRecvError::Closed`] once the event stream is closed.
+    ///
+    /// This method never blocks and is safe to use in asynchronous contexts.
+    ///
+    /// ## Event Reordering
+    ///
+    /// This method uses an [`EventScheduler`] internally to **reorder events**. This means that the
+    /// events might be returned in a different order than they occurred. For details, check the
+    /// documentation of the [`EventScheduler`] struct.
+    ///
+    /// If you want to receive the events in their original chronological order, use the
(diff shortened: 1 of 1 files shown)


phil-opp (approved):
Thanks a lot! I haven't had time for a in-depth review, but this seems like a nice improvement. The generated files and the separation into different modules certainly looks much cleaner now. I also did some basic tests and everything seems to be still working as expected.

phil-opp:
Could you resolve the conflicts with `main` please? Then we can merge.

> The compiling time of ros2-bridge is still very long.

Yes, it is. It's been a while since I wrote the code so I don't remember what part of the compilation requires most time. Is it just the amount of Rust code that is generated or is the bottleneck somewhere else?

phil-opp:
Thanks again!

drindr:
> Is it just the amount of Rust code that is generated or is the bottleneck somewhere else?

The bottleneck is from the `dora-node-api-cxx` crate with generated msg code. The compiling time of c++ API without `ros2-brdige` feature is less than 1 sec, while the one with `ros2-bridge` costs even more than 30s.

The `dora-ros2-bridge` crate without C++ API is near 6s, so I think most of the compiling time is used by the Rust compiler to process the generated code about ROS2's messages.
I do not know if there is a way to optimize it because the code-generate procedure is limited to the interior of a single crate.

PR #1183, opened 2025-11-05
Implement a non-blocking `try_recv` event stream method

phil-opp:


apis/rust/node/src/event_stream/mod.rs
```diff
@@ -255,6 +255,31 @@ impl EventStream {
...
+    ///
+    /// This method uses an [`EventScheduler`] internally to **reorder events**. This means that the
+    /// events might be returned in a different order than they occurred. For details, check the
+    /// documentation of the [`EventScheduler`] struct.
+    ///
+    /// If you want to receive the events in their original chronological order, use the
+    /// [`StreamExt::next`] method with a custom timeout future instead
+    /// ([`EventStream`] implements the [`Stream`] trait).
+    pub fn try_recv(&mut self) -> Result<Event, TryRecvError> {
+        match self.recv_async().now_or_never() {
```
haixuanTao, line 276: I don't think this is the right implementation as you could have event within the scheduler that you would want to return

phil-opp: `recv_async` is our method that uses the scheduler

phil-opp: To clarify: If there are buffered events in the scheduler, the `recv_async` function will return them directly without any blocking/waiting. We only get `Poll::Pending` from that function if both the scheduler is empty and there are no events available in the queue of `self.receiver` either.

haixuanTao: Ah ok make sense I misread that It was on `self.receiver`. Sorry!

PR #1188, opened 2025-11-11
adding native  logging

haixuanTao:
This PR makes it possible to use both rust and python native logging system and get formatting and filtering within `dora run`. This also persist the logs within the logging filesystem as json making them searchable through tool like elastic search, Grafana Loki, ...

This improve the current stdout by adding:
- file path information
- line id
- levels
- target
- and fields that can be fully customizable 
- usable with `send_stdout_as` and receive event as text json lines in order to use a custom log aggregator user node.
- Usable with opentelemetry in order to use an external opentelemetry aggregator / collector

## To use in rust:

```rust

use tracing:warn:

warn!("my message"); 

```

## To use in python

```python
import logging

logging.warn("my message")
```

## To filter log, just as before use `RUST_LOG` env variable.

The output is going to be formatted as is:

## Output

```bash
INFO  receive_data_with_sleep: root  info [108171982329125]
DEBUG receive_data_with_sleep: __main__  received tick with data
INFO  receive_data_with_sleep: root  info [108171991792958]
DEBUG receive_data_with_sleep: __main__  received tick with data
INFO  receive_data_with_sleep: root  info [108172002233958]
DEBUG receive_data_with_sleep: __main__  received tick with data
INFO  receive_data_with_sleep: root  info [108172012020041]
DEBUG receive_data_with_sleep: __main__  received tick with data
INFO  receive_data_with_sleep: root  info [108172022420791]
DEBUG receive_data_with_sleep: __main__  received tick with data
INFO  receive_data_with_sleep: root  info [108172032057625]
DEBUG receive_data_with_sleep: __main__  received tick with data
```
You will find attached example within the PR

### DIFF
--- a/binaries/daemon/src/spawn.rs
+++ b/binaries/daemon/src/spawn.rs
@@ -19,7 +19,7 @@ use dora_core::{
 use dora_download::download_file;
 use dora_message::{
     DataflowId,
-    common::{LogLevel, LogMessage},
+    common::{LogLevel, LogMessage, LogMessageHelper},
     daemon_to_coordinator::{DataMessage, NodeExitStatus, Timestamped},
     daemon_to_node::{NodeConfig, RuntimeConfig},
     id::NodeId,
@@ -623,11 +623,11 @@ impl PreparedNode {
                 });
 
                 if std::env::var("DORA_QUIET").is_err() {
-                    match serde_json::de::from_str::<LogMessage>(&formatted) {
+                    match serde_json::de::from_str::<LogMessageHelper>(&formatted) {
                         Ok(log_msg) => {
-                            cloned_logger.log(log_msg).await;
+                            cloned_logger.log(LogMessage::from(log_msg)).await;
                         }
-                        Err(err) => {
+                        Err(_err) => {
                             cloned_logger
                                 .log(LogMessage {
                                     daemon_id: Some(daemon_id.clone()),
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -11,7 +11,7 @@ use crate::{BuildId, DataflowId, daemon_to_daemon::InterDaemonEvent, id::NodeId}
 
 pub use log::Level as LogLevel;
 
-#[derive(Debug, Clone, serde::Serialize, PartialEq)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq)]
 #[must_use]
 pub struct LogMessage {
     pub build_id: Option<BuildId>,
@@ -28,30 +28,26 @@ pub struct LogMessage {
     pub fields: Option<BTreeMap<String, String>>,
 }
 
-impl<'de> Deserialize<'de> for LogMessage {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        #[derive(Deserialize)]
-        struct LogMessageHelper {
-            build_id: Option<BuildId>,
-            dataflow_id: Option<DataflowId>,
-            node_id: Option<NodeId>,
-            daemon_id: Option<DaemonId>,
-            level: LogLevelOrStdout,
-            target: Option<String>,
-            module_path: Option<String>,
-            file: Option<String>,
-            line: Option<u32>,
-            message: Option<String>,
-            timestamp: DateTime<Utc>,
-            fields: Option<BTreeMap<String, String>>,
-        }
+#[derive(Deserialize)]
+pub struct LogMessageHelper {
+    build_id: Option<BuildId>,
+    dataflow_id: Option<DataflowId>,
+    node_id: Option<NodeId>,
+    daemon_id: Option<DaemonId>,
+    level: LogLevelOrStdout,
+    target: Option<String>,
+    module_path: Option<String>,
+    file: Option<String>,
+    line: Option<u32>,
+    message: Option<String>,
+    timestamp: DateTime<Utc>,
+    fields: Option<BTreeMap<String, String>>,
+}
 
-        let helper = LogMessageHelper::deserialize(deserializer)?;
+impl From<LogMessageHelper> for LogMessage {
+    fn from(helper: LogMessageHelper) -> Self {
         let fields = helper.fields.as_ref();
-        Ok(LogMessage {
+        LogMessage {
             build_id: helper.build_id.or(fields
                 .and_then(|f| f.get("build_id").cloned())
                 .map(|id| BuildId(Uuid::parse_str(&id).unwrap()))),
@@ -94,7 +90,7 @@ impl<'de> Deserialize<'de> for LogMessage {
                 .unwrap_or_default(),
             fields: helper.fields,
             timestamp: helper.timestamp,
-        })
+        }
     }
 }
 
@@ -379,7 +375,7 @@ mod tests {
             fields: Some(BTreeMap::from([("key".to_string(), "value".to_string())])),
         };
         let serialized = serde_yaml::to_string(&log_message).unwrap();
-        let deserialized: LogMessage = serde_yaml::from_str(&serialized).unwrap();
-        assert_eq!(log_message, deserialized);
+        let deserialized: LogMessageHelper = serde_yaml::from_str(&serialized).unwrap();
+        assert_eq!(log_message, LogMessage::from(deserialized));
     }
 }
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -151,6 +151,7 @@ impl DoraNode {
                 #[cfg(feature = "tracing")]
                 {
                     TracingBuilder::new(node_config.node_id.as_ref())
+                        .with_stdout("info", true)
                         .build()
                         .wrap_err("failed to set up tracing subscriber")?;
                 }
@@ -162,13 +163,6 @@ impl DoraNode {
             }
             Err(std::env::VarError::NotPresent) => {} // continue trying other init methods
         };
-        #[cfg(feature = "tracing")]
-        {
-            TracingBuilder::new(node_config.node_id.as_ref())
-                .with_stdout("info", true)
-                .build()
-                .wrap_err("failed to set up tracing subscriber")?;
-        }
 
         // node integration test mode
         match std::env::var("DORA_TEST_WITH_INPUTS") {


libraries/message/src/common.rs
```diff
@@ -22,9 +24,91 @@ pub struct LogMessage {
...
+            "debug" => Ok(LogLevelOrStdout::LogLevel(LogLevel::Debug)),
+            "info" => Ok(LogLevelOrStdout::LogLevel(LogLevel::Info)),
+            "warn" | "warning" => Ok(LogLevelOrStdout::LogLevel(LogLevel::Warn)),
+            "error" => Ok(LogLevelOrStdout::LogLevel(LogLevel::Error)),
+            _ => Err(serde::de::Error::custom(format!(
+                "invalid log level or stdout: '{}'. Expected one of: trace, debug, info, warn, error, stdout",
+                s
+            ))),
+        }
+    }
```
phil-opp: Why are these custom deserialization implementations needed? Does the roundtrip (serialize->deserialize) still work if the Serialize impl is derived but the deserialization impl is custom?

binaries/daemon/src/spawn.rs
```diff
@@ -590,21 +590,35 @@ impl PreparedNode {
...
-                            level: dora_core::build::LogLevelOrStdout::Stdout,
-                            node_id: Some(node_id.clone()),
-                            target: None,
-                            message: formatted,
-                            file: None,
-                            line: None,
-                            module_path: None,
-                        })
-                        .await;
+                    match serde_json::de::from_str::<LogMessage>(&formatted) {
```
phil-opp: Instead of deserializing the stdout it would be cleaner to just send the log message directly to the ~coordinator~ daemon. This way we don't treat normal stdout output of the node as log message by accident and we don't have issues with mangled stdout in multi-threaded programs.

haixuanTao: The thing with stdout is that the data is persisted in the same fashion as other stdout output.

We can send all our stdout message then through `send_stdout_as` .

It's harder to combine stdout and logs in other fashion. It would also require concurrently borrowing mutably the node for logs and normal output.

libraries/message/src/common.rs
haixuanTao: We need this because the default json serialization of Tokio `tracing` implement a json log structure different from our own, mainly resolving around fields for any information other than level timestamp ...

phil-opp: And this is not possible to do with [serde attributes](https://serde.rs/attributes.html)? I'm asking because custom deserialization implementations are error-prone and difficult to maintain.

If there is really no other way, could you at least add some tests that `deserialize(serialize(x)) == x` (i.e. that the roundtrip still works)?

binaries/daemon/src/spawn.rs
phil-opp: We can still provide the option to log to stdout, but this should be the decision of the user. Using a separate channel for log messages would have the following advantages:

- Log messages don't clutter your stdout
  - For example, when using dynamic nodes, users might want to see the "normal" stdout in the terminal. Forcibly printing all the log messages there might not what users want.
- Different filters are possible
  - For example, you might want to only log messages with `INFO` level and above to stdout, but some other identity can still subscribe to `DEBUG` log messages
- Custom stdout format is possible
  - e.g. print the data in human-readable format to stdout, but send it to the daemon in JSON format
  - e.g. don't print timestamps to stdout to avoid cluttering the terminal
- less contention on `stdout`
  - `stdout` is protected by a mutex. Nodes that log a lot, perhaps from multiple threads, might experience lock contention and worse performance
  - `stdout` only has a small buffer. If the daemon is not clearing the buffer quickly enough, writing to `stdout` can block the process until the daemon is ready

phil-opp: > It's harder to combine stdout and logs in other fashion. 

We have timestamps for everything, so you can bring them back into order.

Receiving nodes can just subscribe on both the `send_stdout_as` and `send_log_as` outputs. Or even better, we could make scope the `send_log_as` field by log level, so that nodes can also only subscribe to log messages of e.g. level `INFO` or higher. Dora would deliver the log and stdout messages in chronological order to the node, so it's not really difficult to combine them again on the receiver side. The receiver would also have the advantage that the log input would be in some structured form, e.g. a `StructArray`. This way it could directly inspect the fields that it has interest in, without the need to parse a log string.

> It would also require concurrently borrowing mutably the node for logs and normal output.

We could just spawn a background thread that listens on a MPSC channel and sends incoming log messages to the daemon (e.g. via a TCP connection or pipe). Then we register a tracing subscriber layer that owns the sending half of the channel and sends all log messages to the channel. No need for concurrent mutable borrows.

haixuanTao: Yeah, so not going to lie, if I'm a user, and now I need to learn about:
- A new api called `send_log_as`
- A new dataflow syntax based on `logs/info` that will also make it harder for us to validate dataflow and define the dataflow specification.
- the `StructArray` on how the logs are represented within dora that is unique compared to string json that is a common item among logging. So that I can write a user define custom logging aggregator

On top of that we need to maintain:
- yet another thread loop about log messages to the daemon on both nodes and daemon that could have multiple point of failures including user generated extreme case logs data.
- We need to make sure that whatever we do is extensible to the range of opentelemetry system or user defined logging design.

In that sense, I genuinely believe that json based string is the goto. Now, I can imagine stdout having it's limitation that I would generally prefer tackling later.

haixuanTao: Beyond the point above, we will also want to support C/C++ that apparently use cout/cerr pretty extensively https://stackoverflow.com/questions/7400418/writing-a-log-file-in-c-c and using a simple json string as output could be a lot simpler for them

phil-opp: For C++, one of the most popular logging libraries appears to be [spdlog](https://github.com/gabime/spdlog), which seems very similar to the Rust and Python log frameworks. It also supports [custom sinks](https://github.com/gabime/spdlog/wiki/Sinks#implementing-your-own-sink), so I think we can just forward it to Rust's tracing.

For C, there also seem to be options such as [log.c](https://github.com/rxi/log.c). There we should be able to use `log_add_callback` to forward the logs to Rust's tracing again.

phil-opp: I think we're discussing different proposals in parallel here. Let me try to split it up a bit:

1. **Getting log messages from node to daemon**
    - Proposal **1a**: print log messages to stdout in JSON format, try to parse each stdout line again in daemon, treat lines that fail to serialize as normal stdout
      - biggest drawback: log message JSON might be mixed with other output, fail to parse, and be ignored
      - see above for other drawbacks
    - Proposal **1b**: use a different channel for sending log messages to daemon
      - drawback: requires an additional background thread
2. **Log filtering on the receiver side**
    - Proposal **2a**: introduce a new `log/info` syntax to only subscribe to log messages of level INFO or higher
    - Proposal **2b**: don't provide any helpers for filtering, let the node do it itself
3. **Data format when subscribing to log messages**
    - Proposal **3a**: string with the encoded JSON data
    - Proposal **3b**: `StructArray`

---

I'm fine with both 3a and 3b, so we can go ahead with 3a if you prefer.

2a is something that we can look into later if needed. So this is not really relevant to this PR. 2b is and always will be an option.

For 1, I prefer 1a over 1b because I can easily see cases where the parsing will fail and important log messages might be missed. For example:

```rust
print!("[");
for item in list {
    if item.is_err() {
        tracing::error!("Item is error!");
    }
    print!("{item}, ");
}
println!("]");
```

The `tracing::error!` call won't be detected during deserialization in this case because the stdout line starts with `[`, which is not a start of a valid JSON object (so deserialization fails).

phil-opp: There is also:

4. **How to subscribe to log messages?**
    - Proposal **4a**: Receive log messages via existing `send_stdout_as` config
    - Proposal **4b**: Receive log messages via new `send_log_as` config

Given that the `send_stdout_as` config already exists and will continue to exist, proposal 4a will be available anyway as long as we print log messages to stdout.

We can decide later if there is demand for a separate `send_log_as` config. So let's not add it for now.

haixuanTao: Why I really do think 1a is going to make our life easier:
- stdout is still the most common way to output logs for most applications ( including docker, Kubernetes, ...). If people wants logs for specific logging analytics, they should use opentelemetry, and that will cover things beyond dora like other dockers containers...
- Yes, malformatted logs for any reason happens all the time. and system should be able to recover from malformatted logs!
- I don't think we should make it so that people who want to use a whatever logging library should **have to write a callback function through an FFI to rust so that Tokio-tracing can report to a custom tcp channel that is connected to the daemon.**. This is a **wild expectation on user's end** for any custom logging library. spdlog and log.c or even Tokio-tracing are definitely popular but I can imagine scenario where they might not be used and in which case being able to json-stdout logs will make everyone's life easier.

phil-opp: > * I don't think we should make it so that people who want to use a whatever logging library should **have to write a callback function through an FFI to rust so that Tokio-tracing can report to a custom tcp channel that is connected to the daemon.**. This is a **wild expectation on user's end** for any custom logging library. spdlog and log.c or even Tokio-tracing are definitely popular but I can imagine scenario where they might not be used and in which case being able to json-stdout logs will make everyone's life easier.

The callbacks are required in C to get some structured log output at all. Without it, you only have normal `stdout` with the loggers default output, no JSON format.

phil-opp: > * stdout is still the most common way to output logs for most applications ( including docker, Kubernetes, ...).

Sure, and with proposal 1b you could still log to stdout **in a nice format**. The log message with all metadata would still reach the daemon and CLI.

With proposal 1a, your stdout now looks like this:

```
{\"build_id\":\"47d2d161-2ef6-4daf-aa0a-956c2f747cc3\",\"dataflow_id\":\"4d458477-c9ec-4e6d-871f-fc01d2dc46a3\",\"node_id\":\"foo\",\"daemon_id\":null,\"level\":\"INFO\",\"target\":\"example-target\",\"module_path\":\"foo/bar/baz\",\"file\":\"/home/someuser/projects/dataflow/src/foo/bar.rs\",\"line\":32,\"message\":\"the actual log message\",\"timestamp\":\"2025-11-19T12:57:49.092151811Z\",\"fields\":null}
{\"build_id\":\"47d2d161-2ef6-4daf-aa0a-956c2f747cc3\",\"dataflow_id\":\"4d458477-c9ec-4e6d-871f-fc01d2dc46a3\",\"node_id\":\"foo\",\"daemon_id\":null,\"level\":\"ERROR\",\"target\":\"example-target\",\"module_path\":\"foo/bar/baz\",\"file\":\"/home/someuser/projects/dataflow/src/foo/bar.rs\",\"line\":32,\"message\":\"an error occured!\",\"timestamp\":\"2025-11-19T12:57:49.092151811Z\",\"fields\":null}
{\"build_id\":\"47d2d161-2ef6-4daf-aa0a-956c2f747cc3\",\"dataflow_id\":\"4d458477-c9ec-4e6d-871f-fc01d2dc46a3\",\"node_id\":\"foo\",\"daemon_id\":null,\"level\":\"INFO\",\"target\":\"example-target\",\"module_path\":\"foo/bar/baz\",\"file\":\"/home/someuser/projects/dataflow/src/foo/bar.rs\",\"line\":32,\"message\":\"the actual log message\",\"timestamp\":\"2025-11-19T12:57:49.092151811Z\",\"fields\":null}
some normal println output
{\"build_id\":\"47d2d161-2ef6-4daf-aa0a-956c2f747cc3\",\"dataflow_id\":\"4d458477-c9ec-4e6d-871f-fc01d2dc46a3\",\"node_id\":\"foo\",\"daemon_id\":null,\"level\":\"INFO\",\"target\":\"example-target\",\"module_path\":\"foo/bar/baz\",\"file\":\"/home/someuser/projects/dataflow/src/foo/bar.rs\",\"line\":32,\"message\":\"the actual log message\",\"timestamp\":\"2025-11-19T12:57:49.092151811Z\",\"fields\":null}
```

[Playground Example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6afde3c5a1616a1780cfacac91e3337d)

With wrapping, this looks even uglier:

image

You get no colors or anything, so it becomes quite difficult to read this. Did you spot the error message in this example?

phil-opp: > * Yes, malformatted logs for any reason happens all the time. and system should be able to recover from malformatted logs!

Yes, we can of course recover, but it might lead to swallowed error messages.

We would not have this issue with proposal 1b, so I wonder what bigger drawbacks you see with 1b?

phil-opp: We decided in yesterday's meeting to go ahead with proposal 1a (encoding log messages as JSON to stdout) for now to get something out quickly. We can still change this later if needed.

libraries/message/src/common.rs
phil-opp: It still feels wrong to me to derive `Serialize`, but use a custom deserialization routine. Is it even possible to deserialize serialized data again this way?

I'm pretty sure that we don't actually need a custom deserialization function here, at least not such a big one. Could you give me some more details on the issues of deriving `Deserialize` for `LogMessage` directly?

haixuanTao: So I added a roundtrip test to make sure the data is kept intact.

The issues is that our custom log message format has more information than a typical json lines from Tokyo tracing. We includes things like node_id, dataflow_id, daemon_id, ...

Those information will be stored within the fields key-value of the json that we need to parse and add back to the struct.

If we need to use default deserialize we would need to change the logmessage struct to use fields key-value hash map instead of attributes.

phil-opp: I found the [`flatten_event`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.flatten_event) method to move the fields to the root object. I opened https://github.com/dora-rs/dora/pull/1237 against this PR to use it and switch to derived deserialize.

PR #1190, opened 2025-11-12
Enhance dataflow stop procedure

Mivik:
## Description

See [the discussion](https://github.com/orgs/dora-rs/discussions/1171) for more details.

This PR implements the three-phase stop procedure described in the discussion. This significantly improves the current state where dataflows are often difficult to stop gracefully.

`process-wrap` crate is used to organize processes into groups (process group on Unix and JobObject on Windows). This ensures that all descendant processes are reliably terminated as soon as the root process is killed.

Additionally, a `force: bool` argument has been added to the stop command, enabling instant process termination. When a dataflow is launched via `dora start`, pressing Ctrl-C twice will now send a force stop request to the daemon, mirroring the behavior found in Docker.

## Backward compatibility

The `force` argument has been added to several message types, and `#[serde(default)]` is used to ensure backward compatibility with existing implementations.

### DIFF
--- a/binaries/cli/src/command/stop.rs
+++ b/binaries/cli/src/command/stop.rs
@@ -26,7 +26,12 @@ pub struct Stop {
     /// 1. Sends `Event::Stop` to all nodes of the dataflow.
     /// 2. After DURATION, performs a soft kill (sending SIGTERM, or Ctrl-Break on Windows).
     /// 3. If the dataflow is still running after DURATION * 0.5, terminates all its processes.
-    #[clap(long, value_name = "DURATION", group = "strategy", verbatim_doc_comment)]
+    #[clap(
+        long,
+        value_name = "DURATION",
+        group = "strategy",
+        verbatim_doc_comment
+    )]
     #[arg(value_parser = parse)]
     grace_duration: Option<Duration>,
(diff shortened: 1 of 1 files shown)


binaries/cli/src/command/stop.rs
```diff
@@ -19,9 +19,12 @@ pub struct Stop {
     #[clap(long)]
     name: Option<String>,
     /// Kill the dataflow if it doesn't stop after the given duration
-    #[clap(long, value_name = "DURATION")]
+    #[clap(long, value_name = "DURATION", group = "strategy")]
     #[arg(value_parser = parse)]
     grace_duration: Option<Duration>,
+    /// Force stop the dataflow
```
phil-opp: Could you add some details to the docs here? i.e. soft kill vs hard kill/SIGTERM vs SIGKILL/grace duration

phil-opp (approved):
Looks good to me overall!

Some additonal docs on the CLI force flag would be nice. As I understand it, the normal stop command would wait for the grace duration and then try a soft kill first before doing a hard kill of the whole process group. A force stop would kill the process group directly without waiting. Am I understanding this correctly?

phil-opp (approved):
Thank you!

Issue #1228, opened 2025-11-18
Shell Completion Script Generation


ZhangHanDong:
**Goal**: Support auto-completion for mainstream shells

**Functional Requirements**:

- `dora completions bash` - Generate Bash completion
- `dora completions zsh` - Generate Zsh completion
- `dora completions fish` - Generate Fish completion
- Installation instructions (output to stderr)

**Technical Points**:

- clap built-in completion generation functionality
- Dynamic completion (e.g., topic names, node names)

**Example**:

```bash
$ dora completions bash > /etc/bash_completion.d/dora
$ source /etc/bash_completion.d/dora
$ dora topic <TAB>
echo  hz  info  list  record  play
```

mathurojus:
Hi @ZhangHanDong and @phil-opp!

I'd like to work on this shell completion script generation task. I've reviewed the issue and understand the requirements:

- Generate completion scripts for bash, zsh, and fish shells
- Support dynamic completions for topics/nodes
- Provide installation instructions
- Leverage clap's built-in completion generation

I'm ready to start implementing this. Please let me know if I can take this on!

PR #1240, opened 2025-11-25
Add completions subcommand

drindr:
With this command, it can generate auto-completion script for specific shell, like bash and fish.

Use `eval "$(dora completions <SHELL>)"` to apply the autocompletion script.

This PR is proposed in #1228.


drindr:
I have added more help information which can be seen with `dora completions --help`
And it will try to autodetect the shell type if the shell type is not given manually now.

drindr:
> I don't think the helper work:
> 
> ```shell
> (base) ~/D/w/d/e/rust-dataflow ❯❯❯     echo "eval \$(dora completions bash)" >> .bashrc                                                                              (base)  ✘ 130 cli-autocompletion ✭ ◼
> zsh: no such file or directory: .bashrc
> ```
> 
> Also, you didn't mention about auto detection of shell.

For zsh, I think it should be `echo "eval \$(dora completions)" >> ~/.zshrc`
The `SHELL` parameter is wrapped by `[` and `]` in the help info, which means it is optional.
When the shell is not given, it will try to figure out which shell you are using.

Do you mean write down all the available shell as example in the help info?

> Could you also remove the s of completions as there is at most one auto-completion.

Fine.

drindr:
Updated.

haixuanTao:
Great! Thanks!

PR #1241, opened 2025-11-25
add `try_recv` method for python

haixuanTao:
this function enables to check for input without waiting for it to resolve

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -208,38 +208,22 @@ impl Node {
 
     /// `.try_recv()` gives you the next input in the queue that the node has received.
     /// It does not block until the next event becomes available.
-    /// It will raise an error if no input is available or if the channel is closed.
     ///
     /// ```python
-    ///
-    /// try:
-    ///     event = events.try_recv()
+    /// event = events.try_recv()
     ///     print(event)
-    /// except Exception as e:
-    ///     print(f"Error receiving event: {e}")
     /// ```
     ///
     /// :rtype: dict
     #[allow(clippy::should_implement_trait)]
-    pub fn try_recv(&mut self, py: Python) -> PyResult<Py<PyDict>> {
-        let events = match self.events.try_recv() {
-            Ok(event) => event
-                .to_py_dict(py)
-                .context("Could not convert event into a dict")
-                .unwrap_or_else(|_| PyDict::new(py).into()),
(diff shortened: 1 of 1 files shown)


apis/python/node/src/lib.rs
```diff
@@ -195,6 +206,33 @@ impl Node {
...
+    /// `.try_recv()` gives you the next input in the queue that the node has received.
+    /// It does not block until the next event becomes available.
+    ///
+    /// ```python
+    /// events = node.drain()
+    /// for event in events:
+    ///     print(event)
+    /// ```
+    ///
+    /// :rtype: list[dict]
```
phil-opp: The example and the type info are still about the `drain` function.

haixuanTao: Indeed sorry!

haixuanTao:
> > I thought that Closed channel would result in the `TryRecvError::Closed` error.
> > Can it be None?
> 
> Ah sorry, I was confused. There is no `None`, but there is `TryRecvError::Empty` instead. It looks like this currently also mapped to a Python exception with a string message. Is there a good way to differentiate an empty from a closed channel this way? This would be useful imo because you might want to handle a closed channel differently than an empty one (e.g. break vs wait for a while).

Yeah, so for now it only going to be text based but in the future we could add custom exception

phil-opp:
Ok, I guess you could work around this by checking the new `is_empty` before calling `try_recv`.

Maybe it would still be better to return an `Option` instead, which would map to Python's `Optional`. So basically return `None` on empty and throw an exception only if finished. This would be more intuitive in my opinion. It would also match the `drain` method more closely, which also returns an empty value if no events are ready, instead of throwing an exception.

I don't feel strongly about this though, so feel free to merge this as-is if you prefer.

haixuanTao:
> Ok, I guess you could work around this by checking the new `is_empty` before calling `try_recv`.
> 
> Maybe it would still be better to return an `Option` instead, which would map to Python's `Optional`. So basically return `None` on empty and throw an exception only if finished. This would be more intuitive in my opinion. It would also match the `drain` method more closely, which also returns an empty value if no events are ready, instead of throwing an exception.
> 
> I don't feel strongly about this though, so feel free to merge this as-is if you prefer.

Sounds good, to me. I think it's a bit ugly to have

```
try:
  node.try_recv()
  ```
  
which would make the second try tautological.

PR #1242, opened 2025-11-25
Add is empty method to the API

haixuanTao:


apis/rust/node/src/event_stream/mod.rs
```diff
@@ -334,6 +334,11 @@ impl EventStream {
         event.map(Self::convert_event_item)
     }
 
+    /// Check if there are any buffered events in the scheduler.
+    pub fn is_empty(&self) -> bool {
+        self.scheduler.is_empty()
```
phil-opp: We should also check `self.receiver.is_empty()` here

haixuanTao: Good catch thanks!

PR #1244, opened 2025-11-28
Python node event mutex

oortlieb:
Allows concurrent read/write access through a node (to address https://github.com/dora-rs/dora/issues/748). "Already borrowed" seems to happen when there are concurrent calls to read/write functions because both currently require an exclusive `Node` ref.

1. Updates `DelayedCleanup.get_mut` and callers to use `&self` instead of `&mut self`.
2. Adds a mutex to `Events.inner` so callers can use `&self` instead of `&mut self`. 
3. Adds an example dataflow that fails on main (with the "Already borrowed" error) but works with this change

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -265,8 +265,8 @@ impl Node {
                 let dict = event
                     .to_py_dict(py)
                     .context("Could not convert event into a dict")?;
-              
-                    Ok(Some(dict))
+
+                Ok(Some(dict))
             })
         } else {
             Ok(None)


haixuanTao:
> RefCell

I think refcell would be single threaded making multi-threaded application not working...

oortlieb:
Good point!

@haixuanTao What needs to happen to get this merged? Are you happy with it or do you want to consider alternative implementations?

apis/python/node/src/lib.rs
```diff
@@ -134,7 +135,7 @@ impl Node {
 
         Ok(Node {
             events: Events {
-                inner: EventsInner::Dora(events),
+                inner: Mutex::new(EventsInner::Dora(events)),
```
haixuanTao: I think Arc Mutex is better in this case for multi-threaded use cases

```suggestion
                inner: Arc::new(Mutex::new(EventsInner::Dora(events))),
```

apis/python/node/src/lib.rs
```diff
@@ -385,13 +384,14 @@ fn err_to_pyany(err: eyre::Report, gil: Python<'_>) -> Py<PyAny> {
 }
 
 struct Events {
-    inner: EventsInner,
+    inner: Mutex<EventsInner>,
```
haixuanTao:
```suggestion
    inner: Arc<Mutex<EventsInner>>,
```

apis/python/node/src/lib.rs
phil-opp: What would be the advantage of adding an `Arc` here? Is there a way to clone the event channel that I don't see?

apis/python/operator/src/lib.rs
```diff
@@ -40,7 +40,7 @@ impl<T> DelayedCleanup<T> {
         CleanupHandle(self.0.clone())
     }
 
-    pub fn get_mut(&mut self) -> std::sync::MutexGuard<T> {
+    pub fn get_mut(&self) -> std::sync::MutexGuard<T> {
         self.0.try_lock().expect("failed to lock DelayedCleanup")
     }
```
phil-opp, line 45: Note that this will still panic on concurrent send operations. This PR seems to be mostly about supporting send and receive concurrently, so I don't think that we need to fix this as part of the same PR. I just wanted to mention it.

oortlieb: Yes, good callout.

apis/python/node/src/lib.rs
haixuanTao: Well, yes you could have python cloning the node within Python. 

I don't know if python reference counting is enough for making this thread safe.

phil-opp: AFAIK pyo3 already wraps the whole thing into some reference counted wrapper. So I would assume that the new Arc here will always have a reference count of exactly 1.

PR #1245, opened 2025-12-01
Implement basic node restarting

phil-opp:
- Make `PreparedNode` clonable and prepare for node restarting
- Restart nodes according to restart policy

Proposed in https://github.com/orgs/dora-rs/discussions/1181

### DIFF
--- a/libraries/message/src/descriptor.rs
+++ b/libraries/message/src/descriptor.rs
@@ -90,12 +90,21 @@ pub struct Descriptor {
     pub debug: Debug,
 }
 
+/// Specifies when a node should be restarted.
 #[derive(Debug, Clone, Copy, Default, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "kebab-case")]
 pub enum RestartPolicy {
+    /// Never restart the node (default)
     #[default]
     Never,
+    /// Restart the node if it exits with a non-zero exit code.
     OnFailure,
+    /// Always restart the node when it exits, regardless of exit code.
+    ///
+    /// The node will not be restarted on the following conditions:
+    ///
+    /// - The node was stopped by the user (e.g., via `dora stop`).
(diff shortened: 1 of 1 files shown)


phil-opp:
> One of the breaking change that could be nice is to restart node that failed before starting the dataflow. As it is quite common with power cycle or occasional networking issue. We could like try 3 times to respawn the nodes when it fails before giving up.

So this sounds like errors that happen after the node is spawned, but before initializing the Dora node API? If so, they should also be restarted as part of this PR if they have a restart policy set.

Or were you talking about failures to spawn, e.g. because the executable doesn't exist? I can also implement retries in that case, but I'm not sure if that is really something that is fixable by a retry.

haixuanTao:
> So this sounds like errors that happen after the node is spawned, but before initializing the Dora node API? If so, they should also be restarted as part of this PR if they have a restart policy set.

Yes exactly

> Or were you talking about failures to spawn, e.g. because the executable doesn't exist? I can also implement retries in that case, but I'm not sure if that is really something that is fixable by a retry.

No errors like, the executable starts but something fail due to some power cycle or usb error ( fairly common with multiple USB ) and so the dataflow is not yet started

phil-opp:
> I don't see the difference between on-failure and always and I guess it's something to do with like node failing before and like automatic stop right?

The difference is the exit code. `on-failure` only restarts when the node exits with a non-zero exit code, indicating that it was an abnormal exit. An exit with code 0 is considered a successful exit, meaning that the node decided that it was done and that there is nothing else to do for it. In that case, Dora treats the node as `stopped` on `on-failure` and closes all of its outputs. On `restart-policy: always`, Dora also restarts the node after a successful exit, which can be useful if the node is expected to keep running indefinitely.

haixuanTao (approved):

haixuanTao:
That's great thanks!

PR #1249, opened 2025-12-04
feat: add progress bars for time-consuming operations

### LINKED ISSUE #1229: Provide progress bars for time-consuming operations
**Goal**: Add progress bars to all time-consuming operations

**Deliverables**:

- Shared UI component library (progress bars, spinners, multi-task progress)
- Apply to all related commands (build, record, sync, etc.)

**Technical Points**:

- Use `indicatif` crate
- Encapsulate common progress bar patterns

**Acceptance Criteria**:

- All operations > 2 seconds have visual feedback
- Unified style and behavior

**Compatibility**: Improve existing functionality

**Dependencies**: `indicatif` crate

Eshaan-byte:
This PR implements visual progress feedback for all time-consuming Dora CLI operations using the indicatif crate.

Changes:
- Created shared progress bar component library (progress.rs)
- Added progress to build, up, start, and run commands
- Unified cyan/blue styling with success/failure indicators
- Smart TTY detection (auto-hides in CI environments)
- Multi-progress support for parallel operations

All operations >2s now have visual feedback with consistent behavior.

Testing:
- cargo check: passed
- cargo fmt: passed
- cargo clippy: passed

Closes #1229

### DIFF
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -58,7 +58,6 @@ ratatui = "0.29.0"
 itertools = "0.14"
 sysinfo = "0.36.1"
 indicatif = "0.17"
-atty = "0.2"
 
 env_logger = "0.11.3"
 self_update = { version = "0.42.0", features = [
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -57,6 +57,8 @@ crossterm = "0.29.0"
 ratatui = "0.29.0"
 itertools = "0.14"
 sysinfo = "0.36.1"
+indicatif = "0.17"
+atty = "0.2"
 
 env_logger = "0.11.3"
 self_update = { version = "0.42.0", features = [
--- a/binaries/cli/src/command/build/local.rs
+++ b/binaries/cli/src/command/build/local.rs
@@ -158,10 +158,19 @@ impl BuildLogger for LocalBuildLogger {
     }
 
     async fn try_clone(&self) -> eyre::Result<Self::Clone> {
-        let progress_bar = if let (Some(_), Some(multi)) = (&self.progress_bar, &self.multi) {
-            Some(multi.add_spinner(format!("Building {}", self.node_id)))
-        } else {
-            None
+        let progress_bar = match (&self.progress_bar, &self.multi) {
+            (Some(pb), Some(multi)) => {
+                // Clone the progress bar by creating a new one with the same message
+                // and adding it to the MultiProgress
+                let message = pb.inner().message();
+                Some(multi.add_spinner(message))
+            }
+            (Some(pb), None) => {
+                // No MultiProgress, create a standalone progress bar with the same message
+                let message = pb.inner().message();
+                Some(ProgressBar::new_spinner(message))
+            }
+            _ => None,
         };
 
         Ok(LocalBuildLogger {
--- a/binaries/cli/src/command/build/local.rs
+++ b/binaries/cli/src/command/build/local.rs
@@ -77,6 +77,7 @@ async fn build_dataflow(
                 LocalBuildLogger {
                     node_id: node_id.clone(),
                     progress_bar: Some(node_pb),
+                    multi: Some(multi.clone()),
                 },
                 &mut git_manager,
             )
@@ -104,6 +105,7 @@ async fn build_dataflow(
 struct LocalBuildLogger {
     node_id: NodeId,
     progress_bar: Option<ProgressBar>,
+    multi: Option<Arc<MultiProgress>>,
 }
 
 impl BuildLogger for LocalBuildLogger {
@@ -115,10 +117,25 @@ impl BuildLogger for LocalBuildLogger {
         message: impl Into<String> + Send,
     ) {
         let message_str: String = message.into();
+        let log_level = level.into();
 
-        // Update progress bar message if available
+        // Always print log messages to terminal so users can scroll back
+        let level_colored = match log_level {
+            LogLevelOrStdout::LogLevel(level) => match level {
+                log::Level::Error => "ERROR ".red(),
+                log::Level::Warn => "WARN  ".yellow(),
+                log::Level::Info => "INFO  ".green(),
+                log::Level::Debug => "DEBUG ".bright_blue(),
+                log::Level::Trace => "TRACE ".dimmed(),
+            },
+            LogLevelOrStdout::Stdout => "stdout".italic().dimmed(),
+        };
+        let node = self.node_id.to_string().bold().bright_black();
+        println!("{node}: {level_colored}   {message_str}");
+
+        // Also update progress bar message if available
         if let Some(pb) = &self.progress_bar {
-            let level_indicator = match level.into() {
+            let level_indicator = match log_level {
                 LogLevelOrStdout::LogLevel(level) => match level {
                     log::Level::Error => "ERROR",
                     log::Level::Warn => "WARN",
@@ -137,30 +154,20 @@ impl BuildLogger for LocalBuildLogger {
             } else {
                 pb.set_message(format!("{}: {}", self.node_id, message_str));
             }
-        } else {
-            // Fallback to println if no progress bar
-            let level = match level.into() {
-                LogLevelOrStdout::LogLevel(level) => match level {
-                    log::Level::Error => "ERROR ".red(),
-                    log::Level::Warn => "WARN  ".yellow(),
-                    log::Level::Info => "INFO  ".green(),
-                    log::Level::Debug => "DEBUG ".bright_blue(),
-                    log::Level::Trace => "TRACE ".dimmed(),
-                },
-                LogLevelOrStdout::Stdout => "stdout".italic().dimmed(),
-            };
-            let node = self.node_id.to_string().bold().bright_black();
-            println!("{node}: {level}   {message_str}");
         }
     }
 
     async fn try_clone(&self) -> eyre::Result<Self::Clone> {
+        let progress_bar = if let (Some(_), Some(multi)) = (&self.progress_bar, &self.multi) {
+            Some(multi.add_spinner(format!("Building {}", self.node_id)))
+        } else {
+            None
+        };
+
         Ok(LocalBuildLogger {
             node_id: self.node_id.clone(),
-            progress_bar: self
-                .progress_bar
-                .as_ref()
-                .map(|_| ProgressBar::new_spinner(format!("Building {}", self.node_id))),
+            progress_bar,
+            multi: self.multi.clone(),
         })
     }
 }
--- a/binaries/cli/src/command/build/local.rs
+++ b/binaries/cli/src/command/build/local.rs
@@ -66,7 +66,7 @@ async fn build_dataflow(
         });
 
         // Create a progress spinner for this specific node
-        let node_pb = multi.add_spinner(format!("Building {}", node_id));
+        let node_pb = Arc::new(multi.add_spinner(format!("Building {}", node_id)));
 
         let task = builder
             .clone()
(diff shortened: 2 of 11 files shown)


binaries/cli/Cargo.toml
```diff
@@ -57,6 +57,8 @@ crossterm = "0.29.0"
 ratatui = "0.29.0"
 itertools = "0.14"
 sysinfo = "0.36.1"
+indicatif = "0.17"
+atty = "0.2"
```
phil-opp: The `atty` crate is [deprecated](https://github.com/softprops/atty?tab=readme-ov-file#atty). Please use https://doc.rust-lang.org/std/io/trait.IsTerminal.html instead.

binaries/cli/src/progress.rs
```diff
@@ -0,0 +1,433 @@
...
+//!
+//! This module provides a unified interface for displaying progress feedback
+//! across all dora CLI commands using the `indicatif` crate.
+//!
+//! # Design Principles
+//!
+//! - **Unified Style**: All progress bars share consistent styling and behavior
+//! - **Reusable Components**: Common patterns are encapsulated for easy reuse
+//! - **Multi-task Support**: Support for parallel operations with multiple progress bars
+//! - **Graceful Degradation**: Works in environments without TTY support
```
phil-opp: Could you please rewrite this section? Many of those points don't provide additional info. We also don't need the bold titles, the explanations are short enough. The "design principles" heading is not needed either.

'Unified style' and 'reusable components' are both expected for library functions, so no need to mention this here.

We don't need a 'graceful degradation' marketing term either. Instead, please specify what happens if stderr is not a terminal (instead of just saying that it "works").

binaries/cli/src/progress.rs
```diff
@@ -0,0 +1,433 @@
...
+//! ```rust,ignore
+//! use crate::progress::MultiProgress;
+//!
+//! let multi = MultiProgress::new();
+//! let pb1 = multi.add_task("Cloning repository...");
+//! let pb2 = multi.add_task("Building node...");
+//! // do work...
+//! pb1.finish();
+//! pb2.finish();
+//! ```
```
phil-opp, line 53: Please also provide an example that shows an actual progress bar, instead of just spinners.

binaries/cli/src/progress.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+}
+
+/// Utility function to check if progress bars should be displayed.
+///
+/// Progress bars are hidden in the following cases:
+/// - Not running in a TTY
+/// - CI environment detected
+/// - User explicitly disabled progress (via environment variable)
+pub fn should_show_progress() -> bool {
```
phil-opp: Is this method even used anywhere?

binaries/cli/src/progress.rs
phil-opp: What I think is worth mentioning:

- supports both spinners and progress bars (with links to the respective methods)
- what happens in non-terminal envs
- whether there are any special env variables to control/disable progress bar
- link to indicatif's docs for details

binaries/cli/src/progress.rs
```diff
@@ -0,0 +1,433 @@
...
+        pb.set_style(
+            ProgressStyle::default_bar()
+                .tick_chars(SPINNER_CHARS)
+                .template(BYTES_TEMPLATE)
+                .expect("invalid template")
+                .progress_chars("=>-"),
+        );
+        pb.enable_steady_tick(Duration::from_millis(SPINNER_TICK_MS));
+        pb.set_message(message.into());
+        Self { inner: pb }
```
phil-opp: This seems to be almost identical to `new`. Perhaps it makes sense to have a (private) `new_inner` that takes an additonal `template` argument, which can then be called by both `new` and `new_bytes`?

binaries/cli/src/progress.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+impl Default for MultiProgress {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Builder for creating progress bars with custom configuration.
+pub struct ProgressBarBuilder {
```
phil-opp: Is this used anywhere (except for tests)?

binaries/cli/src/command/build/local.rs
```diff
@@ -98,24 +114,53 @@ impl BuildLogger for LocalBuildLogger {
...
+        }
     }
 
     async fn try_clone(&self) -> eyre::Result<Self::Clone> {
         Ok(LocalBuildLogger {
             node_id: self.node_id.clone(),
+            progress_bar: self
+                .progress_bar
+                .as_ref()
+                .map(|_| ProgressBar::new_spinner(format!("Building {}", self.node_id))),
```
phil-opp: This new progress bar won't be part of the `MultiProgress` this way, no? Is that what we want?

binaries/cli/src/command/build/local.rs
```diff
@@ -98,24 +114,53 @@ impl BuildLogger for LocalBuildLogger {
...
+            if !level_indicator.is_empty() {
+                pb.set_message(format!(
+                    "{}: {} - {}",
+                    self.node_id, level_indicator, message_str
+                ));
+            } else {
+                pb.set_message(format!("{}: {}", self.node_id, message_str));
+            }
+        } else {
+            // Fallback to println if no progress bar
```
phil-opp: I'm not sure if we want to "hide" the log messages by default. It might be good to still have the output in the terminal so that you can scroll up later to look for warnings or errors.

phil-opp (commented):
Thanks for the PR!

binaries/cli/src/command/build/local.rs
```diff
@@ -137,30 +154,20 @@ impl BuildLogger for LocalBuildLogger {
...
-            println!("{node}: {level}   {message_str}");
         }
     }
 
     async fn try_clone(&self) -> eyre::Result<Self::Clone> {
+        let progress_bar = if let (Some(_), Some(multi)) = (&self.progress_bar, &self.multi) {
+            Some(multi.add_spinner(format!("Building {}", self.node_id)))
+        } else {
+            None
+        };
```
phil-opp: Now cloning a non-multi progress bar always results in `None`, doesn't it? And why is it always a spinner?

Is the `try_clone` method even used anywhere? If not, we can just remove it instead of trying to get it right.

phil-opp (commented):
Thanks for the updates!

binaries/cli/src/command/build/local.rs
Eshaan-byte: Good catch! You're absolutely right about both issues:
"Now cloning a non-multi progress bar always results in None, doesn't it?" - Yes, the previous code would lose the progress bar if multi was None but progress_bar was Some.
"And why is it always a spinner?" - You're right, I was hardcoding it as a spinner instead of preserving the original progress bar's message.
To answer your question: Yes, try_clone() is used. It's called at [libraries/core/src/build/mod.rs:125](https://github.com/dora-rs/dora/blob/main/libraries/core/src/build/mod.rs#L125) when spawning a tokio task to log stdout from build commands. I've fixed the implementation to properly handle both cases:
If we have a MultiProgress, it adds the cloned progress bar to it (preserving the original message)
If we don't have a MultiProgress but do have a progress bar, it creates a standalone spinner (also preserving the message)
Otherwise, it returns None
The fix now uses pb.inner().message() to preserve the original progress bar's message instead of hardcoding a new one. Let me know if this looks good or if you'd prefer a different approach!

phil-opp: I don't see any new commit, perhaps you forgot to push?

binaries/cli/src/command/build/local.rs
```diff
@@ -109,13 +131,52 @@ impl BuildLogger for LocalBuildLogger {
...
+                let message = pb.inner().message();
+                Some(ProgressBar::new_spinner(message))
+            }
+            _ => None,
+        };
+
         Ok(LocalBuildLogger {
             node_id: self.node_id.clone(),
+            progress_bar,
+            multi: self.multi.clone(),
```
phil-opp: I'm still not sure about this implementation. It's still always a spinner, but I guess that's ok since we don't create progress bars in `LocalBuildLogger`. The bigger issue is that creating a new `ProgressBar` displays another bar, doesn't it?

Why not store the bar as an `Arc<ProgressBar>`? Then we could cheaply clone it without duplicating the bar.

phil-opp (commented):
Thanks for the updates!

The code looks good to me now, but the output format still needs improvement imo.

For me, the `dora build` output looks like this currently:

Image

So the percentage is printed a lot and makes the output quite messy. I would prefer if we could "pin" the progress bar and the spinners to the bottom so that they aren't part of the messages that scroll up. So basically a region at the bottom where we display progress and another region above where we print the log/stdout messages.

For example, here is how `cargo` does it:

Image

The `Building` progress marker at the bottom stays there and does not interfere with the "Compiling" lines above.

Deepak-negi11:
@phil-opp
Screenshot 2026-01-11 211748
I see this at the beginning if this is not a problem i think the pr is good to go . You can check it again if i have miss somthing

phil-opp (approved):
Thanks for the updates!

phil-opp (changes requested):
I noticed that we still see a lot of spinners in the logs for `dora run`:

Image

This needs to be fixed too before we can merge.

heyong4725:
Closing — the work itself is solid, but the underlying motivation is weaker than it looks on inspection.

## What I evaluated

I checked the four commands this PR adds spinners to (`dora build`, `dora up`, `dora start`, `dora run`) against the actual user experience they produce today:

| Command | Typical duration | Already has output? |
|---|---|---|
| `dora build` | seconds to minutes | **Yes — cargo prints per-crate compile status** |
| `dora up` | <5 seconds (spawn coord+daemon, wait for ready) | minimal |
| `dora start` | <2 seconds (send command, wait ack) | minimal |
| `dora run` | varies; can be slow if dataflow has `build:` directives | mix (cargo + then logs) |

The genuinely slow operations (`dora build`, `dora run --build`) already have cargo's compile progress output. Layering a `dora`-level spinner that says "Building nodes..." on top of cargo's 200-line per-crate output creates two competing UX layers. The fast operations (`up`, `start`) are over before a typical user notices.

## The real user pain

The pain that "did `dora up` hang?" reflects is real, but it's a much smaller fix: a one-line `println!` saying "waiting for coordinator..." during the spawn/connect phase would cover 80% of it. The PR's 560-LOC framework + a new `indicatif` dependency is solving a smaller problem with a bigger hammer than it needs.

## Additional concerns

- **`dora start --attach`** streams logs from the dataflow. Interleaving spinners with log lines is the kind of UX bug that's hard to get right (cursor positioning, terminal resize, log lines longer than the terminal width). `MultiProgress::println` helps but does not fully solve it; needs careful visual testing.
- **Rebase cost is non-trivial** if this were to merge: PR is behind 5 commits including #1797 (auth flag on `start_coordinator`), the env-var-driven `coordinator_addr` in `up()`, and a `connect_with_retry` helper that replaces the manual sleep loop. The progress-bar placement was designed for the older flow.

## What would change this decision

If you've seen real users report "is dora hanging?" — especially during `dora up` or the wait-for-daemon phase of `dora run` — that'd be a different conversation. A targeted, small-scope fix could ship the perceived-latency improvement without the framework.

A fresh, small PR doing just that ("Add a `println!` in `dora up` while waiting for coordinator/daemon to connect") would be welcome.

The progress.rs library itself is well-designed (TTY auto-hide, `MultiProgress::println` fallback, tests included). If a future need arises for a richer progress UX, this code is a good reference point.

Thanks @Eshaan-byte for the careful implementation work.

Issue #1210, opened 2025-11-18
CLI Improvement Phase 2: Implement `dora inspect top` Command


ZhangHanDong:
**Goal**: Real-time monitor node resource usage (similar to Linux `top`)

**Functional Requirements**:

- Real-time refresh display (1-5 second intervals)
- Display CPU, memory, I/O for each node
- Support interactive sorting (by CPU/memory)
- Press 'q' to exit

**Technical Points**:

- Use `sysinfo` to periodically query process resources
- Use terminal control sequences for clearing and refreshing screen
- Optional: Use `ratatui` to build TUI (if interaction needed)

**Acceptance Criteria**:

```bash
$ dora inspect top
Refreshing every 2s...

NODE            CPU    MEMORY   I/O READ  I/O WRITE
detector_node   15.8%  450 MB   10 MB/s   2 MB/s
camera_node     5.2%   120 MB   50 MB/s   0 MB/s
planner_node    8.1%   200 MB   1 MB/s    5 MB/s

Press 'q' to quit
```

**Compatibility**: New command, does not affect existing functionality

**Dependencies**:

- `sysinfo` crate
- Optional: `ratatui` (if interactive sorting needed)

haixuanTao:
#1251 implemented this

PR #1251, opened 2025-12-05
feat: implement dora inspect top command for real-time node resource monitoring (#1210)

guptapratykshh:
This PR implements the `dora inspect top` command (#1210) for real-time monitoring of node resource usage. The command provides an interactive TUI built with `ratatui` that displays CPU usage, memory consumption, and process IDs for all running nodes across active dataflows, with configurable refresh intervals (default 2s) and keyboard-driven sorting by node name, CPU, or memory. 

The implementation extends the message protocol with `GetNodeInfo` request/reply types, adds a coordinator handler to collect node information, and uses `sysinfo` for process metrics collection. All changes are backward compatible with no breaking changes to existing functionality.

### DIFF
--- a/binaries/cli/src/command/inspect/top.rs
+++ b/binaries/cli/src/command/inspect/top.rs
@@ -217,17 +217,18 @@ impl App {
 
         // Use daemon-provided metrics (works for distributed nodes!)
         for node_info in node_infos {
-            let (pid, cpu_usage, memory_mb, disk_read_mb_s, disk_write_mb_s) = if let Some(metrics) = &node_info.metrics {
-                (
-                    Some(metrics.pid),
-                    metrics.cpu_usage,
-                    metrics.memory_mb,
-                    metrics.disk_read_mb_s,
-                    metrics.disk_write_mb_s,
-                )
-            } else {
-                (None, 0.0, 0.0, None, None)
-            };
+            let (pid, cpu_usage, memory_mb, disk_read_mb_s, disk_write_mb_s) =
+                if let Some(metrics) = &node_info.metrics {
+                    (
+                        Some(metrics.pid),
+                        metrics.cpu_usage,
+                        metrics.memory_mb,
+                        metrics.disk_read_mb_s,
+                        metrics.disk_write_mb_s,
+                    )
+                } else {
+                    (None, 0.0, 0.0, None, None)
+                };
 
             self.node_stats.push(NodeStats {
                 dataflow_id: node_info.dataflow_id,


binaries/coordinator/src/lib.rs
```diff
@@ -773,6 +773,27 @@ async fn start_inner(
...
+                                for (node_id, _node) in &dataflow.nodes {
+                                    // Find which daemon this node is running on
+                                    for daemon_id in &dataflow.daemons {
+                                        node_infos.push(NodeInfo {
+                                            dataflow_id: dataflow.uuid,
+                                            dataflow_name: dataflow.name.clone(),
+                                            node_id: node_id.clone(),
+                                            daemon_id: daemon_id.clone(),
+                                        });
+                                    }
```
phil-opp: A node is running only on one daemon. This adds the node ID for all daemons to the list, as if the node was running on all the daemons.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+
+        // Update data if refresh interval has passed
+        if last_update.elapsed() >= refresh_duration {
+            // Refresh system info
+            system.refresh_all();
+
+            // Get node info from coordinator
+            let mut session = connect_to_coordinator((coordinator_addr, coordinator_port).into())
+                .wrap_err("Failed to connect to coordinator")?;
```
phil-opp: This reconnects to the coordinater in every loop iteration. Why not reuse the connection?

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+
+            let request = ControlRequest::GetNodeInfo;
+            let reply_raw = session
+                .request(&serde_json::to_vec(&request).unwrap())
+                .wrap_err("failed to send request to coordinator")?;
```
phil-opp, line 331: Is the node info expected to change during calls? It looks like this will always return the same result, so querying it again and again in every loop iteration is just uneccessary overhead.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+                let name = process.name().to_string_lossy();
+                
+                // Check if this process is related to this node
+                if cmd.contains(node_info.node_id.as_ref()) || name.contains(node_info.node_id.as_ref()) {
+                    pid = Some(process_pid.as_u32());
+                    cpu_usage = process.cpu_usage();
+                    memory_mb = process.memory() as f64 / 1024.0 / 1024.0;
+                    break;
+                }
+            }
```
phil-opp: It's not efficient to query this information again for every node. Why not cache the list and reuse it while iterating over `node_infos`?

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+
+        for node_info in node_infos {
+            // Try to find process by searching for the node_id in command line
+            let mut pid = None;
+            let mut cpu_usage = 0.0;
+            let mut memory_mb = 0.0;
+
+            // Search for process that might be this node
+            for (process_pid, process) in system.processes() {
+                let cmd = process.cmd()
```
phil-opp: The [`cmd`](https://docs.rs/sysinfo/latest/sysinfo/struct.Process.html#method.cmd) method might require admin permissions on Windows, which we should not require.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+                    .iter()
+                    .map(|s| s.to_string_lossy())
+                    .collect::<Vec<_>>()
+                    .join(" ");
+                let name = process.name().to_string_lossy();
```
phil-opp: We don't set any custom process name for spawned nodes, so this will not work.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+                
+                // Check if this process is related to this node
+                if cmd.contains(node_info.node_id.as_ref()) || name.contains(node_info.node_id.as_ref()) {
```
phil-opp: The cmd and the process name might not contain the node ID at all. For example, consider the following:

```yaml
foo:
  path: bar
```

There is `foo` in the path to `bar`.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,392 @@
...
+                }
+            }
+
+            self.node_stats.push(NodeStats {
+                dataflow_id: node_info.dataflow_id,
+                dataflow_name: node_info.dataflow_name.unwrap_or_else(|| "<unnamed>".to_string()),
+                node_id: node_info.node_id,
+                pid,
+                cpu_usage,
+                memory_mb,
```
phil-opp, line 241: These will always be `None` if the node runs on a different machine than the CLI.

It will also be `None` if the node ID is different from the executable name.

phil-opp (changes requested):
Thanks for the PR! Unfortunately, the code as written won't work as intended.

In addition to my inline comments, see my general concern in https://github.com/dora-rs/dora/issues/1210#issuecomment-3551994232.

binaries/cli/src/command/inspect/top.rs
guptapratykshh: @phil-opp , I have completely reimplemented this with proper daemon-side metrics collection. All your review comments are now addressed.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,421 @@
...
+                        _ => {}
+                    }
+                }
+            }
+        }
+
+        // Update data if refresh interval has passed
+        if last_update.elapsed() >= refresh_duration {
+            // Refresh system info (process metrics)
+            system.refresh_all();
```
phil-opp: Looks like this isn't used anymore. I think we can remove `sysinfo` completely from this module.

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,421 @@
...
+
+            // Re-query node info periodically (every 30 seconds) or on demand
+            // Node list changes less frequently than metrics
+            if last_node_query.elapsed() >= Duration::from_secs(30) {
```
phil-opp: Given that the metrics are also sent as reply to `GetNodeInfo`, the metrics are only updated every 30 seconds too.

binaries/daemon/src/lib.rs
```diff
@@ -861,6 +880,80 @@ impl Daemon {
...
+                .values()
+                .filter_map(|node| node.pid.map(Pid::from_u32))
+                .collect();
+
+            if !pids.is_empty() {
+                // Refresh all processes at once
+                system.refresh_processes_specifics(
+                    ProcessesToUpdate::Some(&pids),
+                    true,
+                    ProcessRefreshKind::everything(),
```
phil-opp: It's probably better to only refresh the kinds that we need (i.e. cpu, memory, disks).

binaries/daemon/src/lib.rs
```diff
@@ -861,6 +880,80 @@ impl Daemon {
...
+                        if let Some(process) = system.process(sys_pid) {
+                            let disk_usage = process.disk_usage();
+                            metrics.insert(
+                                node_id.clone(),
+                                NodeMetrics {
+                                    pid,
+                                    cpu_usage: process.cpu_usage(),
+                                    memory_bytes: process.memory(),
+                                    disk_read_bytes: Some(disk_usage.read_bytes),
+                                    disk_write_bytes: Some(disk_usage.written_bytes),
```
phil-opp: These numbers are since the last refresh, right? We should probably divide them by `metrics_interval` to get the avg numbers per second.

binaries/daemon/src/lib.rs
```diff
@@ -861,6 +880,80 @@ impl Daemon {
...
+    async fn collect_and_send_metrics(&mut self) -> eyre::Result<()> {
+        use dora_message::daemon_to_coordinator::NodeMetrics;
+        use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate, System};
+
+        if self.coordinator_connection.is_none() {
+            return Ok(());
+        }
+
+        // Create system instance for metrics collection
+        let mut system = System::new();
```
phil-opp: It would be better to reuse the `System` instance. Otherwise fields such as `written_bytes` would be equal to `total_written_bytes`, i.e. bytes written since the start of the node (not within the last second).

binaries/coordinator/src/lib.rs
```diff
@@ -773,6 +773,42 @@ async fn start_inner(
...
+                                            NodeMetricsInfo {
+                                                pid: m.pid,
+                                                cpu_usage: m.cpu_usage,
+                                                memory_mb: m.memory_bytes as f64 / 1024.0 / 1024.0,
+                                                disk_read_mb_s: m
+                                                    .disk_read_bytes
+                                                    .map(|b| b as f64 / 1024.0 / 1024.0),
+                                                disk_write_mb_s: m
+                                                    .disk_write_bytes
+                                                    .map(|b| b as f64 / 1024.0 / 1024.0),
```
phil-opp: If we report in [megabytes](https://simple.wikipedia.org/wiki/Megabyte) we should divide by 1000, not 1024. Dividing by 1024 results in [mebibytes](https://simple.wikipedia.org/wiki/Mebibyte) (symbol `MiB` instead of `MB`).

binaries/cli/src/command/inspect/top.rs
```diff
@@ -0,0 +1,421 @@
...
+use uuid::Uuid;
+
+use crate::{LOCALHOST, common::connect_to_coordinator};
+
+use super::super::{Executable, default_tracing};
+
+/// Real-time monitor node resource usage (similar to Linux top)
+///
+/// Metrics are collected by daemons and reported to the coordinator,
+/// so this works for distributed dataflows across multiple machines.
```
phil-opp, line 35: We should note the following things here:

- values are averaged over the last refresh period
- cpu percentage is of a single core
  - values can add to more than 100 percent if multiple cores are used for dataflow
  - Nodes can run on different machines, with potentially different CPUs. In that case, the percentages are not comparable.

guptapratykshh: @phil-opp All your review comments are now addressed.

binaries/daemon/src/lib.rs
```diff
@@ -114,6 +114,8 @@ pub struct Daemon {
     sessions: BTreeMap<SessionId, BuildId>,
     builds: BTreeMap<BuildId, BuildInfo>,
     git_manager: GitManager,
+    /// System instance for metrics collection (reused across calls)
+    metrics_system: Option<sysinfo::System>,
```
phil-opp: Why is this an `Option`? Is there any way how this can be `None`? If not, let's remove the option so that we can avoid the potential error case.

binaries/daemon/src/lib.rs
```diff
@@ -903,11 +911,15 @@ impl Daemon {
                 .collect();
 
             if !pids.is_empty() {
-                // Refresh all processes at once
+                // Refresh process metrics (cpu, memory, disk)
+                // Note: We need cpu, memory, and disk_usage. ProcessRefreshKind::everything()
+                // includes these along with other metrics. In sysinfo 0.36, there's no API
+                // to selectively enable only specific refresh kinds, so we use everything().
+                let refresh_kind = ProcessRefreshKind::everything();
```
phil-opp: Huh? What about `ProcessRefreshKind::nothing().with_cpu().with_memory().with_disk()`?

guptapratykshh: All requested changes have been addressed. The CI failure in the `musllinux` job is not related to my changes, i think so

PR #1250, opened 2025-12-04
Node Health Introspection and Observability API (#1232)

guptapratykshh:
This PR implements a comprehensive health monitoring and observability system for dora-rs, enabling nodes to autonomously manage their robustness through framework-provided introspection tools. 

The implementation adds lifecycle event subscriptions (subscribe_node_events), health status declarations (set_health_status), input health monitoring (query_input_health, input_is_alive, input_last_received_time), timeout callbacks (set_input_timeout), and error propagation mechanisms (send_error). A real-time event bus in the daemon propagates node lifecycle events (started/stopped/health-changed) to subscribed nodes, while the coordinator tracks system-wide health status. The API is available in both Rust and Python, with new CLI commands (dora node list --health) for observability. 

All features are backward compatible, optional, and follow the principle of minimal framework intervention—nodes control their own recovery strategies while the framework provides the necessary observability infrastructure. The implementation includes 28 files changed (1,420 insertions), full test coverage, and a working example demonstrating all features.

Closes #1232

Issue #1232, opened 2025-11-18
Node Health Introspection and Observability API


ZhangHanDong:
**Goal**: Provide framework-level observability capabilities, allowing nodes to implement their own robustness (corresponding to P0 requirement)

**Design Philosophy**:

- ✅ **Node Autonomy Principle**: Node robustness is guaranteed by the node itself, framework does not over-intervene
- ✅ **Framework Provides Tools**: Provide sufficient introspection and observability APIs for nodes to perceive their own and system status
- ✅ **Minimal Intervention**: Framework only records and propagates events, does not make active decisions (such as automatic restart)

**Functional Requirements (Framework Provides)**:

1. **Node Lifecycle Event API**:
    
    - Nodes can subscribe to lifecycle events of other nodes (start/stop/crash)
    - Nodes can actively declare their own health status (Healthy/Degraded/Failing)
    - Subscribers can perceive abnormal exits of upstream nodes
2. **Dataflow Health Check API**:
    
    - Detect health status of input dataflows (has data/timeout/disconnected)
    - Provide `input.is_alive()` method to check if upstream is still sending data
    - Provide `input.last_received_time()` to get last received time
    - Support setting input timeout callbacks (e.g., trigger if no data for 30 seconds)
3. **Inter-node Heartbeat Mechanism (Optional)**:
    
    - Nodes can optionally send heartbeat signals
    - Dependent nodes can monitor heartbeats to judge upstream health
    - Framework provides lightweight heartbeat tools but does not mandate use
4. **Error Propagation Mechanism**:
    
    - Nodes can send error events downstream (not crashes, but business errors)
    - Downstream nodes can choose how to handle errors (degrade/skip/terminate)
    - Framework ensures error events are reliably delivered
5. **Node Status Query API**:
    
    - Nodes can query coordinator to get global system status
    - Query running status of other nodes (Running/Stopped/Error)
    - Query dataflow topology information (who are my upstream/downstream)

**Node-side API Example (Rust)**:

```rust
use dora_node_api::{Node, NodeEvent, InputHealth};

fn main() -> Result<()> {
    let mut node = Node::init()?;

    // 1. Subscribe to lifecycle events of other nodes
    node.subscribe_node_events(&["camera_node", "detector_node"])?;

    // 2. Set input timeout handling
    node.set_input_timeout("image", Duration::from_secs(5), || {
        eprintln!("Camera feed timeout! Switching to backup...");
    })?;

    for event in node.events() {
        match event {
[...]

PR #1252, opened 2025-12-06
feat: add error event propagation to downstream nodes

guptapratykshh:
This PR implements error event propagation (#1232) that allows nodes to send errors to downstream nodes instead of crashing. The feature provides node.send_error(output_id, error) API method that propagates error events to all affected downstream consumers, who receive InputError events containing error message, source node ID, and affected input ID for graceful error handling.

The implementation extends the message protocol with NodeEvent::InputError and DaemonRequest::SendError types, adds daemon-side error propagation logic that finds and notifies all downstream receivers, and includes complete working example demonstrating producer/consumer error handling patterns. All changes are backward compatible with no breaking changes to existing functionality.

Previous PR #1250

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1991,7 +1991,7 @@ impl Daemon {
     /// It sends `NodeFailed` events to all downstream nodes that consume outputs from the failed node.
     /// For nodes on the same daemon, events are sent directly. For nodes on other daemons,
     /// events are routed through zenoh by sending to all outputs (to ensure all subscribing daemons receive it).
-    async fn propagate_node_error(
+    async fn send_node_failed_events(
         &mut self,
         dataflow_id: Uuid,
         source_node_id: NodeId,
@@ -2564,7 +2564,7 @@ impl Daemon {
                     if matches!(node_error.cause, NodeErrorCause::Other { .. }) {
                         let error_message = format!("{node_error}");
                         if let Err(err) = self
-                            .propagate_node_error(dataflow_id, node_id.clone(), error_message)
+                            .send_node_failed_events(dataflow_id, node_id.clone(), error_message)
                             .await
                         {
                             tracing::warn!(


phil-opp (commented):
Thanks for the PR!

We discussed this and would like to propose the following change: Instead of providing an explicit `send_error` function in the API, the error should be created by the daemon when a node exits with a non-zero exit code. This way, we can notify downstream nodes of failed nodes.

If nodes want to send errors they can still do that using normal outputs.

So please remove the `send_error` method from the API (and also the modifications to `node_to_daemon`.

apis/rust/node/src/event_stream/event.rs
```diff
@@ -41,6 +41,19 @@ pub enum Event {
         /// assigned to the input in the YAML file.
         id: DataId,
     },
+    /// An error occurred in an upstream node.
+    ///
+    /// Instead of crashing, the upstream node sent an error event to allow
```
phil-opp: The sending crashed because of this error. (It might restart once we merge #1245)

apis/rust/node/src/event_stream/event.rs
```diff
@@ -41,6 +41,19 @@ pub enum Event {
...
+    /// downstream nodes to handle the error gracefully (e.g., use cached data,
+    /// switch to backup source, log and continue).
+    InputError {
```
phil-opp: Let's rename this variant to `NodeFailed` or similar.

apis/rust/node/src/event_stream/event.rs
```diff
@@ -41,6 +41,19 @@ pub enum Event {
...
+        /// The ID of the input that had an error, as specified in the YAML file.
+        id: DataId,
```
phil-opp: The node error might affect multiple errors, so let's change this to `affected_input_ids: Vec<DataId>`.

binaries/cli/src/common.rs
```diff
@@ -21,9 +21,21 @@ use uuid::Uuid;
...
+    uuid: Option<Uuid>,
+    ignore_failures: bool,
 ) -> Result<(), eyre::Error> {
     if result.is_ok() {
         Ok(())
+    } else if ignore_failures {
+        // For stop commands, we consider it successful if we successfully stopped the dataflow,
+        // even if the dataflow was already in a failed state
+        Ok(())
     } else {
```
phil-opp: Could you clarify why this is needed?

binaries/coordinator/src/lib.rs
```diff
@@ -1496,23 +1496,52 @@ async fn destroy_daemon(
...
-    tcp_send(&mut daemon_connection.stream, &message)
-        .await
-        .wrap_err(format!(
-            "failed to send destroy message to daemon `{daemon_id}`"
-        ))?;
+    // Try to send destroy message. If it fails, the daemon might already be closed.
+    if let Err(err) = tcp_send(&mut daemon_connection.stream, &message).await {
+        // If we can't send the message, the daemon might have already closed the connection
+        // This can happen during shutdown. Log a debug message but don't fail.
+        tracing::debug!(
```
phil-opp: This should not happen typically, so I would prefer to keep this an error. The code in `destroy_daemons` will still proceed to send destroy commands to the other daemons after an error.

Or is there any specific reason why this can now happen more commonly?

binaries/coordinator/src/lib.rs
```diff
@@ -1496,23 +1496,52 @@ async fn destroy_daemon(
...
+        }
+        Err(err) => {
+            // If deserialization fails, the daemon might have sent a partial reply before closing
+            // This can happen during shutdown. Log a debug message but don't fail.
+            tracing::debug!(
+                "failed to deserialize destroy reply from daemon `{daemon_id}`: {err}, \
+                 assuming daemon is already shut down"
+            );
+            return Ok(());
+        }
```
phil-opp: same as above: Please keep these as errors, unless there is a good reason to change this.

binaries/daemon/src/node_communication/tcp.rs
```diff
@@ -74,9 +74,90 @@ impl Connection for TcpConnection {
...
+                    "failed to deserialize DaemonRequest: {err}, raw data length: {}, preview: {}, hex: {}{}, treating as disconnect",
+                    raw.len(),
+                    preview,
+                    hex_preview,
+                    variant_hint
+                );
+                Ok(None)
+            }
+        }
     }
```
phil-opp, line 81: In general, there should not be any deserialization errors because we use reliable channels. Could you clarify why you think that this change is needed?

Either way, this change it not really related to this PR. So let's move it to a separate one if there is good motivation to do this.

binaries/daemon/src/node_communication/unix_domain.rs
```diff
@@ -72,9 +72,18 @@ impl Connection for UnixConnection {
...
+            Err(err) => {
+                // If deserialization fails, treat it as a connection error
+                // This can happen if the connection was closed mid-transmission
+                // or if there's a version mismatch
+                tracing::debug!(
+                    "failed to deserialize DaemonRequest: {err}, treating as disconnect"
+                );
+                Ok(None)
+            }
+        }
```
phil-opp: same as above (change unrelated to this PR)

binaries/daemon/src/local_listener.rs
```diff
@@ -141,7 +141,14 @@ async fn receive_message(
...
+    match bincode::deserialize(&raw) {
+        Ok(message) => Ok(Some(message)),
+        Err(err) => {
+            // If deserialization fails, treat it as a connection error
+            // This can happen if the connection was closed mid-transmission
+            // or if there's a version mismatch
+            tracing::debug!("failed to deserialize DaemonRequest: {err}, treating as disconnect");
+            Ok(None)
+        }
+    }
```
phil-opp: Same as above (change unrelated to this PR)

libraries/message/src/daemon_to_node.rs
```diff
@@ -73,6 +73,11 @@ pub enum NodeEvent {
     InputClosed {
         id: DataId,
     },
+    InputError {
```
phil-opp: Please apply the same adjustments as for the other `InputError` variant here.

guptapratykshh: All requested changes have been addressed.

binaries/cli/src/command/stop.rs
```diff
@@ -100,7 +102,9 @@ fn stop_dataflow(
         serde_json::from_slice(&reply_raw).wrap_err("failed to parse reply")?;
     match result {
         ControlRequestReply::DataflowStopped { uuid, result } => {
-            handle_dataflow_result(result, Some(uuid))
+            // For stop commands, we consider it successful if we successfully stopped the dataflow,
+            // even if the dataflow was already in a failed state
+            handle_dataflow_result_with_options(result, Some(uuid), true)
```
phil-opp: I don't think that we should hide errors here. A `stop` command should lead to normal stopping of all nodes. If that doesn't work, we should report it.

binaries/cli/src/command/stop.rs
```diff
@@ -127,7 +131,9 @@ fn stop_dataflow_by_name(
...
```
phil-opp: same as above (don't hide error)

binaries/cli/src/common.rs
```diff
@@ -21,9 +21,21 @@ use uuid::Uuid;
...
+    result: DataflowResult,
+    uuid: Option<Uuid>,
+    ignore_failures: bool,
 ) -> Result<(), eyre::Error> {
     if result.is_ok() {
         Ok(())
+    } else if ignore_failures {
+        // For stop commands, we consider it successful if we successfully stopped the dataflow,
+        // even if the dataflow was already in a failed state
+        Ok(())
```
phil-opp: Why should we ignore errors on stop?

binaries/coordinator/src/lib.rs
```diff
@@ -1498,9 +1498,7 @@ async fn destroy_daemon(
 
     tcp_send(&mut daemon_connection.stream, &message)
         .await
-        .wrap_err(format!(
-            "failed to send destroy message to daemon `{daemon_id}`"
-        ))?;
+        .wrap_err("failed to send destroy message to daemon")?;
```
phil-opp: What is this change about?

binaries/daemon/src/node_communication/mod.rs
```diff
@@ -294,6 +294,23 @@ impl Listener {
...
+                    } else {
+                        // Node has already subscribed, might be a different request
+                        "failed to deserialize request: connection error or version mismatch"
+                    };
+                    let error_reply = DaemonReply::Result(Err(error_msg.into()));
+                    if let Err(err) = self.send_reply(error_reply, &mut connection).await {
+                        tracing::debug!(
+                            "failed to send error reply after deserialization failure: {err:?}"
+                        );
+                    }
```
phil-opp: What is this change about? I does not seem related to this PR?

binaries/daemon/src/node_communication/mod.rs
```diff
@@ -503,9 +520,18 @@ impl Listener {
...
+                Err(_) => {
+                    // If the daemon main loop didn't send a reply (e.g., it crashed or returned an error),
+                    // send an error reply to the node so it doesn't hang waiting
+                    tracing::warn!(
+                        "daemon main loop did not send reply for node `{}`, sending error reply",
+                        self.node_id
+                    );
+                    DaemonReply::Result(Err("daemon failed to process request".into()))
+                }
+            }
```
phil-opp: This change does not seem to be related to this PR either.

binaries/daemon/src/lib.rs
```diff
@@ -537,7 +537,17 @@ impl Daemon {
                             break;
                         }
                     }
+                    // Exit when all dataflows are finished after Ctrl-C
```
phil-opp, line 601: We normally don't exit the daemon on ctrl-c. The `exit_when_all_finished` flag is set only in some specific scenarios, e.g. when running `dora run`.

binaries/daemon/src/lib.rs
```diff
@@ -537,7 +537,17 @@ impl Daemon {
...
+                        tracing::info!("exiting daemon because all dataflows are finished");
+                        break;
+                    }
+                    // Also exit in example mode when all dataflows finish, even if exit_when_done
+                    // tracking didn't catch all nodes (e.g., if nodes finish before being tracked)
+                    if self.exit_when_done.is_some() && self.running.is_empty() {
+                        tracing::info!(
+                            "exiting daemon because all dataflows are finished (example mode)"
+                        );
                         break;
```
phil-opp: What is this change about? Is it related to this PR?

binaries/daemon/src/lib.rs
```diff
@@ -2047,43 +2134,54 @@ impl Daemon {
...
+                        // Nodes killed by grace duration during stop are not errors
+                        if grace_duration_kill {
+                            logger
+                                .log(
+                                    LogLevel::Debug,
+                                    Some("daemon".into()),
+                                    format!("node `{node_id}` was stopped by grace duration (not an error)"),
+                                )
+                                .await;
+                            Ok(())
```
phil-opp: Why not? If nodes are unresponsive and need to be killed that seems like an error. For example, there might be some important cleanup logic in the node that was skipped because of the kill. So I think it's better to report this error than to hide it silently.

binaries/daemon/src/lib.rs
```diff
@@ -2105,10 +2203,36 @@ impl Daemon {
...
+                    .handle_node_stop(dataflow_id, &node_id, dynamic_node)
+                    .await;
+                if let Err(err) = &stop_result {
+                    tracing::warn!(
+                        "Error handling node stop for {}/{}: {err:?}, but NodeStopped event should have been sent",
+                        dataflow_id,
+                        node_id
+                    );
+                }
+                // Don't propagate the error - NodeStopped event is more important for shutdown
```
phil-opp: These comments and warn message does not seem to match the code.

binaries/daemon/src/lib.rs
```diff
@@ -1661,6 +1678,65 @@ impl Daemon {
         Ok(())
     }
 
+    /// Send error events to downstream nodes when a node fails.
+    ///
+    /// This is called automatically when a node exits with a non-zero exit code.
+    /// It sends `NodeFailed` events to all downstream nodes that consume outputs from the failed node.
+    async fn propagate_node_error(
```
phil-opp: This should go through the coordinator since the receivers might be running on other daemons

guptapratykshh: done

libraries/message/src/daemon_to_daemon.rs
```diff
@@ -21,4 +21,11 @@ pub enum InterDaemonEvent {
         node_id: NodeId,
         output_id: DataId,
     },
+    NodeFailed {
+        dataflow_id: DataflowId,
+        source_node_id: NodeId,
+        receiver_node_id: NodeId,
+        affected_input_ids: Vec<DataId>,
```
phil-opp: We don't really need these two fields in the message, the receiver can figure it out on its own. This has the advantage that we only need to send a single message.

binaries/daemon/src/lib.rs
```diff
@@ -1722,15 +1774,45 @@ impl Daemon {
...
+                affected_input_ids,
+                error: error_message.clone(),
+            };
+            
+            // Send to remote receivers using the same mechanism as regular outputs
+            if let Err(err) = self.send_to_remote_receivers(dataflow_id, &output_id, event).await {
+                tracing::warn!(
+                    "Failed to send error event to remote receiver {}: {err:?}",
+                    receiver_node_id
+                );
```
phil-opp: We can simplify this by removing the `receiver_node_id` and `affected_input_ids` fields of `InterDaemonEvent::NodeFailed`. Then we only need to send a single message here.

binaries/daemon/src/lib.rs
```diff
@@ -559,7 +559,17 @@ impl Daemon {
...
                     if self.exit_when_all_finished && self.running.is_empty() {
+                        tracing::info!("exiting daemon because all dataflows are finished");
+                        break;
+                    }
+                    // Also exit in example mode when all dataflows finish, even if exit_when_done
+                    // tracking didn't catch all nodes (e.g., if nodes finish before being tracked)
+                    if self.exit_when_done.is_some() && self.running.is_empty() {
+                        tracing::info!(
+                            "exiting daemon because all dataflows are finished (example mode)"
+                        );
```
phil-opp: Why is this needed?

binaries/daemon/src/lib.rs
```diff
@@ -1771,6 +1849,96 @@ impl Daemon {
...
+            if let Some(channel) = dataflow.subscribe_channels.get(&receiver_node_id) {
+                // Local receiver - send directly
+                let event = NodeEvent::NodeFailed {
+                    affected_input_ids,
+                    error: error_message.clone(),
+                    source_node_id: source_node_id.clone(),
+                };
+                let _ = send_with_timestamp(channel, event, &self.clock);
+            }
+        }
```
phil-opp, line 1983: This looks like the same code as when handling `InterDaemonEvent::NodeFailed`. Why not move it to a separate function?

binaries/daemon/src/lib.rs
```diff
@@ -1771,6 +1849,96 @@ impl Daemon {
...
+
+        // If there are remote receivers, send a single InterDaemonEvent
+        // The receiving daemon will determine which of its nodes are affected
+        if has_remote_receivers {
+            // Use the first output as the routing key (any output will work)
+            let first_output = outputs.first().ok_or_else(|| {
+                eyre!("Failed node has no outputs, cannot send error to remote receivers")
```
phil-opp: Instead of erroring, we could just do nothing (if node has no outputs, no remote receivers are affected)

binaries/daemon/src/lib.rs
```diff
@@ -1771,6 +1849,96 @@ impl Daemon {
...
+            let output_id = OutputId(source_node_id.clone(), first_output.clone());
+            let event = InterDaemonEvent::NodeFailed {
+                dataflow_id,
+                source_node_id: source_node_id.clone(),
+                error: error_message.clone(),
+            };
+
+            // Send to remote receivers using the same mechanism as regular outputs
+            if let Err(err) = self
+                .send_to_remote_receivers(dataflow_id, &output_id, event)
```
phil-opp: I forgot that daemons subscribe to a specific output topic. So just sending the message to the first output is not enough unfortunately (as some daemon might only be subscribed to the second or third output).

So I guess we have two options:

- send the message to all outputs and deduplicate it on the receiver side (i.e. remember which nodes failed already to avoid sending duplicated events to nodes)
- add a new node-level subscription topic for such messages that all affected daemons subscribe to

What do you prefer?

guptapratykshh: done

guptapratykshh:
Can you please review this PR. The changes are done , the ci is failing through the different version

guptapratykshh:
ping @phil-opp @haixuanTao

guptapratykshh:
ping @phil-opp

guptapratykshh:
any updated on this, it is ready for review @phil-opp

PR #1260, opened 2025-12-12
Remove default zenoh endpoint peer config

phil-opp:
With the default config we get a lot of `CONNECTION_TO_SELF` errors since https://github.com/eclipse-zenoh/zenoh/pull/2110 was merged.

### DIFF
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -42,14 +42,6 @@ pub async fn open_zenoh_session(coordinator_addr: Option<IpAddr>) -> eyre::Resul
                         ),
                     )
                     .unwrap();
-                if cfg!(target_os = "macos") {
-                    warn!(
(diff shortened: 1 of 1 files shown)


phil-opp:
The diff looks a bit weird. I basically removed the first `if` branch, so that we go directly into the `else` branch.

phil-opp:
> But I think the scouting doesn't work on macOS

Is there some issue with more details on this? I looked through the zenoh repo but didn't find anything related.

phil-opp:
So it seems like it was the other way around: The else branch also disabled multicast scouting on macos: https://github.com/dora-rs/dora/pull/1260/changes#diff-a2231c167364419201c67cde2890e158d1faea556e39524e28913cb80b983aceL84-L91

Turns out that was the reason why the CI was failing. I removed that exception in bdf57c6 and now the macos CI jobs pass as well.

Issue #1201, opened 2025-11-18
CLI Improvement Phase 1:  Implement `dora topic info` Command


ZhangHanDong:
**Goal**: Display detailed metadata of a topic

**Functional Requirements**:

- Topic type (data type)
- Publisher list (node name, publishing frequency)
- Subscriber list (node name)
- Topic statistics (total message count, bandwidth)

**Technical Points**:

- Extend coordinator metadata query interface
- Format and output detailed information

**Acceptance Criteria**:

```bash
$ dora topic info /sensors/camera
Type: Image
Publishers:
  * camera_node (30.0 Hz)
Subscribers:
  * detector_node
  * logger_node
```

**Compatibility**: New command, does not affect existing functionality

heyong4725:
Closing — shipped.

`dora topic info` is implemented at `binaries/cli/src/command/topic/info.rs` and registered as a subcommand of `dora topic` (see `binaries/cli/src/command/topic/mod.rs`). The other topic subcommands (`echo`, `hz`, `list`, `pub`) ship alongside it.

If a specific gap remains in the implementation, please open a fresh focused issue with the missing capability.

PR #1261, opened 2025-12-12
feat: implement dora topic info command (#1201)

guptapratykshh:
#1201 

This PR adds a new dora topic info command that provides detailed metadata and real-time statistics about dataflow topics/outputs, making it easier to monitor and debug running dataflows.

The new command gives you a comprehensive view of any topic in your dataflow:

- Data Type: Shows the Arrow data type being transmitted (e.g., UInt64, String, List<Float32>)
- Publisher: Displays which node is publishing, along with its real-time frequency in Hz
- Subscribers: Lists all downstream nodes consuming this topic
- Statistics: Shows total message count and bandwidth usage over a configurable time window

### DIFF
--- a/binaries/cli/src/command/topic/info.rs
+++ b/binaries/cli/src/command/topic/info.rs
@@ -1,0 +1,344 @@
+use std::{
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
+
+use arrow_schema::DataType;
+use clap::Args;
+use dora_core::{
+    config::InputMapping,
+    topics::{open_zenoh_session, zenoh_output_publish_topic},
+};
+use dora_message::{
+    common::Timestamped, daemon_to_daemon::InterDaemonEvent, metadata::ArrowTypeInfo,
+};
+use eyre::{Context, eyre};
+use tokio::runtime::Builder;
+
+use crate::{
+    command::{Executable, default_tracing, topic::selector::TopicSelector},
+    common::CoordinatorOptions,
+};
+
+/// Display detailed metadata of a topic.
+///
+/// Shows topic type, publisher, subscribers, and statistics (message count,
+/// bandwidth, publishing frequency).
+///
+/// Examples:
+///
+/// Get info for a single topic:
+///   dora topic info -d my-dataflow camera_node/image
+///
+/// Note: The dataflow descriptor must include the following snippet so that
+/// runtime messages can be inspected (or messages must cross machine
+/// boundaries so they are forwarded through zenoh):
+///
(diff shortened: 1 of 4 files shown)


binaries/cli/src/command/topic/info.rs
```diff
@@ -0,0 +1,353 @@
...
+        let subscribe_topic = zenoh_output_publish_topic(
+            dataflow_id_clone,
+            &topic_clone.node_id,
+            &topic_clone.data_id,
+        );
+        let subscriber = zenoh_session
+            .declare_subscriber(subscribe_topic)
+            .await
+            .map_err(|e| eyre!(e))
+            .wrap_err_with(|| format!("failed to subscribe to {}", topic_clone))?;
```
phil-opp, line 171: This will only work for messages that are sent across machines (or require the debug flag to send everything to zenoh). We should note that.

binaries/cli/src/command/topic/info.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        let start_time = Instant::now();
+        let duration_duration = Duration::from_secs(duration);
```
phil-opp: `duration_duration` is a weird name. Why not just `duration`?

binaries/cli/src/command/topic/info.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        // Collect messages for the specified duration
+        while start_time.elapsed() < duration_duration {
+            if let Ok(sample) = tokio::time::timeout(
+                duration_duration - start_time.elapsed(),
```
phil-opp: It's easier to use `timeout_at`, then we only have to compute the end time once.

binaries/cli/src/command/topic/info.rs
```diff
@@ -0,0 +1,353 @@
...
+            .await
+            .map_err(|e| eyre!(e))
+            .wrap_err_with(|| format!("failed to subscribe to {}", topic_clone))?;
+
+        let start_time = Instant::now();
+        let duration_duration = Duration::from_secs(duration);
+
+        // Collect messages for the specified duration
+        while start_time.elapsed() < duration_duration {
+            if let Ok(sample) = tokio::time::timeout(
```
phil-opp: Instead of `if let Ok(sample) = ... { ... } else { break }` I would prefer `let Ok(sample) = ... else {break;};` here (less rightward drift and `break` is closer to condition)

guptapratykshh: done

Issue #1202, opened 2025-11-18
CLI Improvement Phase 1:  Implement `dora node list` Command


ZhangHanDong:
**Goal**: List all currently running nodes and their status

**Functional Requirements**:

- Display node name, status (Running/Stopped/Error), PID, CPU, memory
- Support filtering by dataflow: `dora node list --dataflow <name>`
- Support `--format json` output

**Technical Points**:

- Retrieve node information from coordinator
- Query system resource usage (using `sysinfo` library)

**Acceptance Criteria**:

```bash
$ dora node list
NODE            STATUS    PID     CPU    MEMORY
camera_node     Running   12345   5.2%   120 MB
detector_node   Running   12346   15.8%  450 MB
planner_node    Error     -       -      -
```

**Compatibility**: New command, does not affect existing functionality

**Dependencies**: Introduce `sysinfo` crate

heyong4725:
Closing — shipped.

`dora node list` is implemented at `binaries/cli/src/command/node/list.rs` and registered under the `dora node` subcommand group (`binaries/cli/src/command/node/mod.rs`). The `dora node` group also ships `add`, `connect`, `disconnect`, `info`, `remove`, `restart`, `stop`.

If a specific gap remains, please open a fresh focused issue.

PR #1262, opened 2025-12-13
feat: implement dora node list command with metrics and filtering (#1202)

guptapratykshh:
This PR adds dora node list command (#1202) to list all currently running nodes and their resource metrics. The command offers both table and JSON formats. It shows the node name, status, PID, CPU usage, and memory consumption. Users can filter by dataflow name or UUID.

The implementation uses current metrics collection system. The daemon collects process metrics every 2 seconds with the sysinfo crate. It sends these metrics to the coordinator through NodeMetrics events. The CLI queries the coordinator using the GetNodeInfo request/reply protocol to get node information and formats it for display.

### DIFF
--- a/binaries/cli/src/command/mod.rs
+++ b/binaries/cli/src/command/mod.rs
@@ -9,6 +9,7 @@ mod inspect;
 mod list;
 mod logs;
 mod new;
+mod node;
 mod run;
 mod runtime;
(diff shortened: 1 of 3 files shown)


phil-opp (approved):
Looks good, thank you!

PR #1269, opened 2025-12-17
feat(cli) Added command `dora system status` as a replacement for `dora check` 

### LINKED ISSUE #1205: Implement `dora system status` Command
**Goal**: Replace existing `dora check`, provide more detailed system status

**Functional Requirements**:

- Coordinator status (running/stopped) + address and port
- Daemon status + PID
- Number of currently running dataflows
- dora version information
- System uptime

**Technical Points**:

- Integrate existing `check` command logic
- Extend output information
- Use colored output (green=healthy, red=abnormal)

**Acceptance Criteria**:

```bash
$ dora system status
✓ Coordinator: Running (v0.3.5)
  Address: 127.0.0.1:53290
  Uptime: 2h 15m
✓ Daemon: Running (PID: 12340)
Active dataflows: 2
```

**Compatibility**:

- Keep `dora check` as alias
- Add `dora system status`

BhaveshAgarwal67:
Resolves #1205

### DIFF
--- a/binaries/cli/src/command/system/status.rs
+++ b/binaries/cli/src/command/system/status.rs
@@ -15,7 +15,6 @@ use std::{
 use std::{net::IpAddr, path::PathBuf};
 use termcolor::{Color, ColorChoice, ColorSpec, WriteColor};
 
-
 pub fn check_environment(coordinator_addr: SocketAddr) -> eyre::Result<()> {
     let mut error_occurred = false;
 
@@ -52,10 +51,7 @@ pub fn check_environment(coordinator_addr: SocketAddr) -> eyre::Result<()> {
     };
 
     // Daemon status
-    let daemon_running = session
-        .as_deref_mut()
-        .map(daemon_running)
-        .transpose()?;
+    let daemon_running = session.as_deref_mut().map(daemon_running).transpose()?;
 
     if daemon_running == Some(true) {
         let _ = stdout.set_color(ColorSpec::new().set_fg(Some(Color::Green)));
@@ -126,7 +122,7 @@ fn query_running_dataflow_count(
 
 #[derive(Debug, clap::Args)]
 pub struct Status {
-     /// Path to the dataflow descriptor file (enables additional checks)
+    /// Path to the dataflow descriptor file (enables additional checks)
     #[clap(long, value_name = "PATH", value_hint = clap::ValueHint::FilePath)]
     dataflow: Option<PathBuf>,
     /// Address of the dora coordinator


binaries/cli/src/command/system/status.rs
```diff
@@ -0,0 +1,200 @@
...
+    };
+    let mut stdout = termcolor::StandardStream::stdout(color_choice);
+
+    // Coordinator status
+    let mut session = match connect_to_coordinator(coordinator_addr) {
+        Ok(session) => {
+            let _ = stdout.set_color(ColorSpec::new().set_fg(Some(Color::Green)));
+            write!(stdout, "✓ ")?;
+            let _ = stdout.reset();
+            writeln!(stdout, "Coordinator: Running (v{})", VERSION)?;
```
phil-opp: `VERSION` is the version of the CLI, not the version of the coordinator. They might be different, e.g. when the coordinator is running on a different machine with a different version of dora installed.

binaries/cli/src/command/system/status.rs
```diff
@@ -0,0 +1,200 @@
...
+        .ok()?;
+
+    let reply: ControlRequestReply = serde_json::from_slice(&reply_raw).ok()?;
+    match reply {
+        ControlRequestReply::NodeInfoList(nodes) => nodes
+            .first()
+            .and_then(|n| n.metrics.as_ref())
+            .map(|m| m.pid),
+        _ => None,
+    }
```
phil-opp: How is the pid of the first node equal to the daemon pid?

binaries/cli/src/command/system/status.rs
```diff
@@ -0,0 +1,200 @@
...
+        .as_deref_mut()
+        .map(query_daemon_info)
+        .transpose()?
+        .flatten();
+
+    if let Some(info) = daemon_info {
+        let _ = stdout.set_color(ColorSpec::new().set_fg(Some(Color::Green)));
+        write!(stdout, "✓ ")?;
+        let _ = stdout.reset();
+        writeln!(stdout, "Daemon: Running (PID: {})", info.pid)?;
```
phil-opp: The daemon might be running on a different machine, not sure how useful the pid is in that case (without knowing the machine ID).

binaries/cli/src/command/system/status.rs
```diff
@@ -0,0 +1,200 @@
...
+
+#[derive(Debug, clap::Args)]
+pub struct Status {
+    #[clap(long, value_name = "PATH", value_hint = clap::ValueHint::FilePath)]
+    dataflow: Option<PathBuf>,
+    #[clap(long, value_name = "IP", default_value_t = LOCALHOST)]
+    coordinator_addr: IpAddr,
+    #[clap(long, value_name = "PORT", default_value_t = DORA_COORDINATOR_PORT_CONTROL_DEFAULT)]
+    coordinator_port: u16,
+}
```
phil-opp, line 134: Why remove the doc comments for this struct after renaming it from `Check`?

binaries/cli/src/command/system/status.rs
BhaveshAgarwal67: Right, sorry for the mixup, i will remove the version from here, or should i mention that is the cli version?

binaries/cli/src/command/system/status.rs
BhaveshAgarwal67: Thanks for pointing that out, i will add a check to display pid only when its running on local machine unless theres a way of knowing the machine ID

binaries/cli/src/command/system/status.rs
BhaveshAgarwal67: Fair point, I agree this is a bit misleading as this is me trying to extract the daemon id by assuming it appears as the first entry in NodeInfoList, since i couldnt find a way to reliably extract daemon id. I am happy to remove the PID from the output or mention it as first node unless theres a way to extract daemon id directly.

binaries/cli/src/command/system/status.rs
BhaveshAgarwal67: Sorry this was a genuine mistake, I will fix it

Issue #1207, opened 2025-11-18
CLI Improvement Phase 2: Implement `dora system stop` Command


ZhangHanDong:
**Goal**: Replace `dora destroy`, provide more friendly stop command

**Functional Requirements**:

- Stop all running dataflows
- Stop coordinator and daemon
- Display stop progress
- Optional: `--force` to force termination

**Technical Points**:

- Integrate existing `destroy` command logic
- Add confirmation prompt (if dataflows are running)
- Use progress bar to display stop process

**Acceptance Criteria**:

```bash
$ dora system stop
? 2 dataflows are running. Stop them? (y/N) y
⣾ Stopping dataflows...
✓ All dataflows stopped
✓ Coordinator stopped
✓ Daemon stopped
```

**Compatibility**:

- Keep `dora destroy` as alias
- Add `dora system stop`

BhaveshAgarwal67:
@dora-bot unassign me

ZhangHanDong:
@phil-opp  OK,  I’ve been busy with other things this week. I’ll review this PR as soon as possible.

heyong4725:
Closing — redundant with `dora down`.

The functionality requested (stop coordinator + daemon as a single command) is provided by `dora down` (alias `dora destroy`, see `binaries/cli/src/command/down.rs`). The proposed rename to `dora system stop` was declined in the umbrella thread (#1196).

If a `dora system` subcommand group is genuinely wanted for orthogonal reasons, please open a focused fresh issue.

Issue #1206, opened 2025-11-18
CLI Improvement Phase 2: Implement `dora system start` Command


ZhangHanDong:
**Goal**: Replace `dora up`, provide clearer system startup

**Functional Requirements**:

- Start coordinator and daemon
- Display startup progress (using `indicatif` progress bar)
- Output service address and port
- Provide detailed error information on startup failure

**Technical Points**:

- Integrate existing `up` command logic
- Introduce `indicatif` to display progress
- Improve error handling and prompts

**Acceptance Criteria**:

```bash
$ dora system start
⣾ Starting coordinator...
✓ Coordinator started on 127.0.0.1:53290
✓ Daemon started (PID: 12340)
```

**Compatibility**:

- Keep `dora up` as alias
- Add `dora system start`

**Dependencies**: Introduce `indicatif` crate

ZhangHanDong:
@phil-opp I think I need some time to refine this requirement; I’ll convert it to a draft for now.

heyong4725:
Closing — redundant with `dora up`.

The functionality requested (start coordinator + daemon as a single command) is provided by `dora up` (see `binaries/cli/src/command/up.rs`). The proposed rename to `dora system start` was declined in the umbrella thread (#1196).

If a `dora system` subcommand group is genuinely wanted for orthogonal reasons, please open a focused fresh issue with the specific motivation.

PR #1270, opened 2025-12-17
feat(cli) Added command `system start` and `system stop` as a replacement for `destroy` and `up`

BhaveshAgarwal67:
Implemented issues #1206 and #1207

### DIFF
--- a/binaries/cli/src/command/system/start.rs
+++ b/binaries/cli/src/command/system/start.rs
@@ -1,9 +1,9 @@
-use crate::command::{default_tracing, Executable};
-use crate::{common::connect_to_coordinator, LOCALHOST};
+use crate::command::{Executable, default_tracing};
+use crate::{LOCALHOST, common::connect_to_coordinator};
 use communication_layer_request_reply::TcpRequestReplyConnection;
 use dora_core::topics::DORA_COORDINATOR_PORT_CONTROL_DEFAULT;
 use dora_message::{cli_to_coordinator::ControlRequest, coordinator_to_cli::ControlRequestReply};
-use eyre::{bail, Context, ContextCompat};
+use eyre::{Context, ContextCompat, bail};
 use indicatif::{ProgressBar, ProgressStyle};
 use std::net::{IpAddr, SocketAddr};
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


binaries/cli/src/command/system/stop.rs
```diff
@@ -0,0 +1,192 @@
...
+        let spinner = create_spinner("Stopping dataflows...");
+        // The Destroy command stops all dataflows first
+        spinner.finish_and_clear();
+        print_success(&format!(
+            "Stopped {} dataflow{}",
+            running_count,
+            if running_count == 1 { "" } else { "s" }
+        ));
+    }
+
```
phil-opp, line 129: I don't think that we should print status messages that don't match what we're doing. This is a no-op that claims that the dataflows have been stopped, which is not true.

binaries/cli/src/command/system/stop.rs
```diff
@@ -0,0 +1,192 @@
...
+            }
+        } else {
+            // Non-interactive, require --force or --yes
+            bail!(
+                "{} dataflow{} running. Use --force or --yes to stop without confirmation.",
+                running_count,
+                if running_count == 1 { " is" } else { "s are" }
+            );
+        }
+    }
```
phil-opp, line 116: This will break any scripts that use `dora destroy` (because a new confirmation is required).

binaries/cli/src/command/system/start.rs
```diff
@@ -0,0 +1,209 @@
...
+            spinner.finish_and_clear();
+            print_success(&format!(
+                "Coordinator already running on {}:{}",
+                coordinator_addr.ip(),
+                coordinator_addr.port()
+            ));
+            session
+        }
+        Err(_) => {
+            if let Err(e) = start_coordinator() {
```
phil-opp, line 76: This does not take the `coordinator_addr` into account. So it will start the coordinator on the standard port instead of using the port number given as argument.

BhaveshAgarwal67:
Thanks for the review. Given my current time constraints, I don’t think I can address this properly right now, so I’ll close the PR to avoid blocking discussion.

PR #1273, opened 2025-12-20
Make adding more `NodeEvent` variants backwards-compatible

phil-opp:
Ignore unknown event types during deserialization.


phil-opp:
Oh, automerge was still on, so it was merged as soon as the CI was green, even though there are lots of debug changes now. I reverted this in https://github.com/dora-rs/dora/pull/1277.

phil-opp:
New PR in https://github.com/dora-rs/dora/pull/1278

PR #1275, opened 2025-12-22
Fill in node_id, dataflow_id, and daemon_id in daemon

phil-opp:
Instead of providing it via the Python node API.

This approach has the advantage that it works consistently across languages. For example, the Rust node API was not setting these fields, leading to log messages without a node_id. This commit fixes that.


haixuanTao (approved):
Looks great thanks!

PR #1279, opened 2025-12-23
Split Python CLI job into smaller jobs

phil-opp:
for easier debugging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -483,16 +483,6 @@ jobs:
       - name: "Python Queue Latency + Timeout Test"
         timeout-minutes: 30
         run: dora run tests/queue_size_and_timeout_python/dataflow.yaml --uv
-      - name: "Python Object Detection Node Example"
-        timeout-minutes: 30
-        run: |
-          echo "Running Python Object Detection Node Example"
-          dora up
-          dora build examples/python-dataflow/dataflow-object_detection.yml --uv
-          dora start examples/python-dataflow/dataflow-object_detection.yml --name ci-python --detach --uv
-          sleep 10
-          echo "Running dora stop"
-          dora stop --name ci-python  --grace-duration 30s
 
       - name: "Rust Queue Latency Test"
         timeout-minutes: 30


haixuanTao:
I think that the fact that it works in Windows and not other OS probably means that there's probably nothing wrong with the split of the CI jobs. 

I wonder if it might be linked to cli/api version mismatch or something but yeah would probably need to dive deeper beyond the changes included in this PR.

phil-opp:
What do you think about moving the python example to the node hub repo?

haixuanTao:
I understand the frustration.

I think the thing is that, the fact that the CI is failing in 2 different OS means there is **definitely a bug somewhere** that user will face at one point or another. 

And I believe, moving those test to an optional CI would make our code less reliable and more prone to errors as we're not fixing errors that is underlying within our repo. 

As we don't have all the time in the world to fix all the issues within dora, I would prefer that we either keep this PR open and wait for someone to take over or rebase it, or close it.

phil-opp:
I created this PR mainly for making debugging the CI failures of https://github.com/dora-rs/dora/pull/1278 easier. https://github.com/dora-rs/dora/pull/1278 implements an actual improvement to Dora to reduce the version mismatch error that users are experiencing. This work is now blocked on some CI errors that I cannot reproduce locally.

> I think the thing is that, the fact that the CI is failing in 2 different OS means there is **definitely a bug somewhere** that user will face at one point or another.

I agree with that. We should try our best to find and fix this.

The best way to do that in my opinion is to create a simpler Python test case that tells us whether the basic functionality works as expected. Right now we only have a very complex test that involves heavy dependencies, so we have no idea where to look (is this even an issue in Dora or is some dependency hanging).

> As we don't have all the time in the world to fix all the issues within dora, I would prefer that we either keep this PR open and wait for someone to take over or rebase it, or close it.

If this issue is too complex for us to debug, it will probably not be fixed by a contributor that is less familiar with Dora either. So the bug will probably stay unfixed.

And given that it affects other PRs as well, we will waste a lot of time on CI failures that aren't related to the changes we make.

Issue #1231, opened 2025-11-18
ROS Message Format Compatibility Layer


ZhangHanDong:
**Goal**: Implement interoperability between Dora and ROS message formats (corresponding to requirement #8 second point)

**Background**:

- Feedback from Chongqing University of Posts and Telecommunications deep users: need compatibility with ROS message formats and toolchain
- Lower migration costs for ROS users switching to Dora
- Allow Dora nodes to coexist with existing ROS nodes

**Functional Requirements (Phase 1 - Message Format Compatibility)**:

- Automatically convert common ROS message types to Dora data formats:
    - `sensor_msgs`: Image, CompressedImage, PointCloud2, LaserScan, Imu, CameraInfo
    - `geometry_msgs`: Pose, PoseStamped, Twist, TwistStamped, Transform, TransformStamped
    - `nav_msgs`: Odometry, Path, OccupancyGrid
    - `std_msgs`: Header, String, Int32, Float64, Bool, etc.
- Provide ROS `.msg` file parser
- Generate Dora data structure binding code (Rust/Python)
- Support custom ROS message types

**Functional Requirements (Phase 2 - ROS Bridging)**:

- Implement `dora-ros-bridge` node for bidirectional bridging between Dora and ROS/ROS2
- Support ROS 1 (`rospy`/`roscpp`) and ROS 2 (`rclpy`/`rclcpp`)
- Configuration file specifies bridge mappings
- Automatically handle timestamp conversion (ROS Time ↔ Dora Time)
- Support TF/TF2 coordinate transform bridging

**Functional Requirements (Phase 3 - ROS Toolchain Compatibility)**:

- Allow using ROS tools to view Dora topics:
    - `rostopic list` can see Dora topics
    - `rostopic echo` can view Dora messages
    - `rqt` can connect to Dora nodes
- Implement ROS Master compatibility layer (optional)
- Support rosbag format recording and playback

**Technical Points**:

- Use `rosrust` or `r2r` library to integrate ROS 1/2
- Implement message serialization/deserialization converters
- Use `ros-message-parser` to parse `.msg` definitions
- Bridge node uses efficient zero-copy mechanism (shared memory)
- Code generation: generate Dora Schema from `.msg`

**CLI Command Extensions**:

```bash
# Generate Dora types from ROS message definitions
$ dora ros import sensor_msgs/Image
✓ Generated Dora type: dora_types::ros::sensor_msgs::Image
✓ Updated: src/types/ros/sensor_msgs.rs

[...]

PR #1285, opened 2026-01-03
feat(ros-compat): add minimal ROS message conversion with Python example

guptapratykshh:
#1231 
This PR adds a minimal Python API for converting Dora Arrow arrays to ROS message format dictionaries. Sometimes you need to work with ROS message structures in Dora nodes without setting up a full ROS bridge, and this library makes that easy.

Right now it only supports `geometry_msgs/Twist`, but the structure is simple enough that we can easily add more message types later.

- **Library**: `dora-ros-compat` with a simple `RosMessageConverter` class
- **Python API**: `from dora.ros import RosMessageConverter` - converts Arrow arrays to ROS message dicts
- **Working example**: A complete example showing how to convert Twist messages

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -450,6 +450,13 @@ jobs:
           dora build tests/queue_size_latest_data_rust/dataflow.yaml --uv
           dora run tests/queue_size_latest_data_rust/dataflow.yaml --uv
 
+      - name: "Test CI (ROS Compat Mock)"
+        timeout-minutes: 5
+        shell: bash
+        run: |
+          uv pip install pyarrow
+          python3 examples/ros-compat/python/ros1_turtle/test_mock_bridge.py
+
       - name: "Test CLI (C)"
         timeout-minutes: 45
         # fail-fast by using bash shell explictly
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -465,12 +465,12 @@ jobs:
           python examples/ros-compat/python/ros1_turtle/test_mock_bridge.py
           
           # Run full dataflow
-          sed -i 's/ROS_PORT: 9090/ROS_PORT: 9090\n        ROS_MOCK_MODE: 1/' examples/ros-compat/python/ros1_turtle/dataflow.yml
+          echo "      ROS_MOCK_MODE: 1" >> examples/ros-compat/python/ros1_turtle/dataflow.yml
           dora up
           dora start examples/ros-compat/python/ros1_turtle/dataflow.yml --name ros1-compat-test --detach
           sleep 10
-          dora stop --name ros1-compat-test --grace-duration 5s
-          dora destroy
+          dora stop --name ros1-compat-test --grace-duration 5s || true
+          dora destroy || true
 
       - name: "Test CLI (C)"
         timeout-minutes: 45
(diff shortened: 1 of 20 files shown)


guptapratykshh:
Hi @haixuanTao,

Thanks for the feedback. I have simplified the PR significantly based on your suggestions.

haixuanTao:
I think one of the main limitation of this implementation is that it doesn't automatically generate message format as the ros2 api and also it is only available in Python

guptapratykshh:
I have added `get_schema()` method to automatically generate message formats from ROS definitions, effectively solving the main limitation. The Python-only scope is now explicitly documented as a known constraint in the README.

haixuanTao:
The CI example does not use dora dataflow

heyong4725:
Closing — the work itself is well-written but the delivered value collapses on inspection.

## What I checked

Read the core library (`libraries/extensions/ros-compat/src/python.rs`, 207 LOC) and both examples. Two methods on the new `RosMessageConverter`:

### `to_ros()` is essentially a no-op wrapper

```rust
pub fn to_ros(&self, data: Bound<...>, msg_type: Option<&str>) -> PyResult<PyObject> {
    let _ = msg_type;                              // argument silently discarded
    let pylist = data.call_method0("to_pylist")?;  // just calls pyarrow built-in
    Ok(pylist.into())
}
```

This is a 2-line wrapper around `pyarrow.Array.to_pylist()`. A user can write `event["value"].to_pylist()` directly and get the same result. The `msg_type` parameter is in the signature but never used. The README and PR description claim "converts Arrow arrays to ROS message dicts" but there is no ROS-specific conversion happening — the conversion is whatever pyarrow already does.

### `get_schema()` is useful, but in the wrong place

The `get_schema("geometry_msgs/Twist")` method parses ROS `.msg` files and emits a pyarrow Schema. That is genuinely useful. But:

- It depends on `dora_ros2_bridge_msg_gen` (the existing ros2-bridge crate). So this new "ros-compat" crate is a thin Python-facing veneer over ros2-bridge machinery.
- It requires `AMENT_PREFIX_PATH` set — the same prerequisite as using the full ros2-bridge. So the "no DDS / no full bridge needed" framing does not actually reduce setup overhead.

Spinning up a new top-level extension crate to expose one useful method that calls into an existing crate is the wrong factoring. The schema-generation helper belongs as a Python method on the existing `dora-ros2-bridge` crate.

## Other concerns

- **The `ros1_turtle` example does not actually communicate with ROS1.** Comment in `turtle_bridge.py`: *"for this ROS1 example we will focus on the format conversion compatibility."* The directory name oversells.
- **Issue #1231 was closed 2 days ago** as "Phase 1 message format compatibility is covered by the existing ros2-bridge; broader compat is speculative." This PR implements a closed-as-superseded issue.
- **3.5 months stale**, 3 file conflicts on rebase (Cargo.toml, apis/python/node/Cargo.toml, libraries/core/src/lib.rs).

## Path forward if schema generation is wanted

The genuinely-useful capability here — *"give me a pyarrow Schema for a ROS message type"* — is worth having. The right place is **a new method on the ros2-bridge Python API**, not a separate extension crate. If you want to drive that, please open a focused issue titled something like *"Expose `get_schema(msg_type) -> pyarrow.Schema` on the dora-ros2-bridge Python API"* — it would be a 50-LOC addition to the existing crate rather than a new top-level extension.

Thanks @guptapratykshh for the careful work on the schema-generation piece — that part is solid Rust + pyo3 code and would carry over cleanly to a fresh focused PR.

PR #1291, opened 2026-01-08
add zero-copy send api for python

drindr:
Currently, the send_output method cost 1 copy.

This PR provides a zero-copy method for Python API named `send_output_raw` which provides better performance as the message getting larger.


drindr:
It includes 2 examples. One for showing how to use and one for benchmark the performance between `send_output` and `send_output_raw`

apis/python/node/src/sample_handler.rs
```diff
@@ -44,7 +44,7 @@ impl SampleHandler {
 impl SampleHandler {
     /// Get a writable numpy array wrapping the buffer for zero-copy writing
     /// This returns a writable UInt8 numpy array that Python can fill
```
haixuanTao: Shouldn't this return a typed array? And not a uint8 only array?

apis/python/node/src/sample_handler.rs
```diff
@@ -44,7 +44,7 @@ impl SampleHandler {
...
-    pub fn as_arrow(&mut self, py: Python) -> PyResult<Py<PyAny>> {
+    pub fn as_array(&mut self, py: Python) -> PyResult<Py<PyAny>> {
```
haixuanTao: I think as array could mean a lot of different things, and I fill like:

```suggestion
    pub fn to_numpy(&mut self, py: Python) -> PyResult<Py<PyAny>> {
```

It would be easier to understand

apis/python/node/Cargo.toml
```diff
@@ -20,6 +20,8 @@ telemetry = ["dora-runtime/telemetry"]
 
 [dependencies]
 dora-node-api = { workspace = true }
+dora-message = { workspace = true }
+dora-core = { workspace = true }
```
phil-opp, line 24: The `dora-core` dep is not needed, as the `dora-node-api` crate re-exports the `dora-core` crate.

Instead of adding a dep on `dora-message` we should re-export `DataId` and `ArrowTypeInfo` from `dora-node-api`.

apis/rust/node/src/node/mod.rs
```diff
@@ -515,7 +515,7 @@ impl DoraNode {
         Ok((node, event_stream))
     }
 
-    fn validate_output(&mut self, output_id: &DataId) -> bool {
+    pub fn validate_output(&mut self, output_id: &DataId) -> bool {
```
phil-opp, line 518: If we make this public, we should give it a better name that indicates the boolean return value. Maybe something like `valid_output_id`?

apis/python/node/src/lib.rs
```diff
@@ -382,6 +386,55 @@ impl Node {
...
+    /// sample = node.send_output_raw("output_id", data_length)
+    /// buffer = sample.as_memoryview()
+    /// # ... fill the buffer ...
+    /// sample.send()  # Manually send
+    /// ```
+    ///
+    /// :type output_id: str
+    /// :type data_length: int
+    /// :type metadata: dict, optional
+    /// :rtype: SampleHandler
```
phil-opp: We should mention here that only `uint8` is supported as data type

drindr: With the memoryview, it can be cast into other types easily now. The exact type should depend on the outer wrapper.

apis/python/node/Cargo.toml
drindr: It makes sense. I will modify it soon.

apis/python/node/src/lib.rs
phil-opp: But the `type_info` is still hardcoded to a byte array when sending it out, isn't it?

phil-opp:
Claude suggests something like this, perhaps this is helpful (as reference, not for copy&paste):

<details>

```rust
use pyo3::prelude::*;
use pyo3::exceptions: :{PyBufferError, PyRuntimeError};
use pyo3::ffi;
use bytes::Bytes;
use std::sync::atomic::{AtomicUsize, AtomicBool, Ordering};
use std::sync::Arc;
use std::os::raw::c_int;

/// A buffer that safely tracks Python views and prevents use-after-consume
#[pyclass]
pub struct SafeBuffer {
    inner: Arc<BufferInner>,
}

struct BufferInner {
    // Using Bytes for reference-counted, immutable buffer
    data: std::sync:: Mutex<Bytes>,
    // Track Python buffer protocol exports
    export_count: AtomicUsize,
    // Whether consume() has been called
    consumed: AtomicBool,
}

#[pymethods]
impl SafeBuffer {
    /// Create a new buffer with given size
    #[new]
    fn new(size: usize) -> Self {
        Self {
            inner: Arc::new(BufferInner {
                data: std::sync::Mutex::new(Bytes::from(vec![0u8; size])),
                export_count: AtomicUsize::new(0),
                consumed: AtomicBool::new(false),
            }),
        }
    }
    
    /// Create from existing bytes
    #[staticmethod]
    fn from_bytes(data: Vec<u8>) -> Self {
        Self {
            inner: Arc::new(BufferInner {
                data: std::sync::Mutex::new(Bytes::from(data)),
                export_count:  AtomicUsize::new(0),
                consumed: AtomicBool::new(false),
            }),
        }
    }
    
    /// Get the number of active Python exports
    fn active_exports(&self) -> usize {
        self.inner.export_count.load(Ordering::SeqCst)
    }
    
    /// Check if buffer has been consumed
    fn is_consumed(&self) -> bool {
        self.inner.consumed.load(Ordering::SeqCst)
    }
    
    /// Get the buffer size
    fn __len__(&self) -> usize {
        self.inner.data.lock().unwrap().len()
    }
    
    /// Consume the buffer and return the bytes
    /// Fails if there are active Python views
    fn consume(&self) -> PyResult<Vec<u8>> {
        if self.inner.consumed.swap(true, Ordering::SeqCst) {
            return Err(PyRuntimeError::new_err("Buffer has already been consumed"));
        }
        
        let active = self.inner.export_count. load(Ordering::SeqCst);
        if active > 0 {
            // Reset consumed flag since we're failing
            self.inner.consumed. store(false, Ordering::SeqCst);
            return Err(PyRuntimeError::new_err(
                format!(
                    "Cannot consume buffer:  {} active view(s) still exist.  \
                     Release all memoryviews and numpy arrays before consuming.",
                    active
                )
            ));
        }
        
        let data = self.inner.data. lock().unwrap().clone();
        Ok(data. to_vec())
    }
    
    fn __repr__(&self) -> String {
        let consumed = self.inner.consumed.load(Ordering::SeqCst);
        let exports = self.inner.export_count.load(Ordering::SeqCst);
        let len = self. inner. data.lock().unwrap().len();
        
        format!(
            "SafeBuffer(size={}, consumed={}, active_exports={})",
            len, consumed, exports
        )
    }
    
    /// Implement buffer protocol using PyBuffer_FillInfo helper
    unsafe fn __getbuffer__(
        slf: PyRefMut<'_, Self>,
        view: *mut ffi:: Py_buffer,
        flags: c_int,
    ) -> PyResult<()> {
        // Check if buffer has been consumed
        if slf. inner.consumed.load(Ordering::SeqCst) {
            return Err(PyBufferError::new_err(
                "Cannot create view:  buffer has been consumed"
            ));
        }
        
        let data = slf.inner.data.lock().unwrap();
        
        // Increment export count BEFORE releasing the lock
        slf.inner. export_count.fetch_add(1, Ordering::SeqCst);
        
        // Use PyBuffer_FillInfo to fill the buffer structure
        // This handles all the boilerplate for us! 
        let ret = unsafe {
            ffi::PyBuffer_FillInfo(
                view,
                slf.as_ptr() as *mut ffi::PyObject,
                data.as_ptr() as *mut std::ffi::c_void,
                data.len() as isize,
                1, // readonly = 1 (read-only buffer)
                flags,
            )
        };
        
        if ret == -1 {
            // Failed - decrement export count
            slf.inner.export_count.fetch_sub(1, Ordering::SeqCst);
            return Err(PyErr::fetch(slf.py()));
        }
        
        Ok(())
    }
    
    unsafe fn __releasebuffer__(
        slf: PyRefMut<'_, Self>,
        _view: *mut ffi:: Py_buffer,
    ) {
        let prev = slf.inner.export_count. fetch_sub(1, Ordering::SeqCst);
        if prev == 0 {
            eprintln!("Warning: __releasebuffer__ called with export_count already at 0");
        }
    }
}
```

</details>

heyong4725:
Hey @drindr — closing this with a rescue redirect rather than letting it sit.

The design here was genuinely strong work — your buffer-protocol implementation (Arc-shared `SampleBufferState` with `valid` + `view_count` atomics, the "increment view_count AFTER the view fill" panic-safety detail, the context-manager + manual-form API) is exactly the shape @phil-opp asked for in his original review, and it's the kind of CPython-buffer-protocol code that's easy to get subtly wrong but easy to read when done right. Your code is well-commented enough that I was able to lift the design cleanly into current `main`.

Reimplemented in **#1833** since this branch was 4 months stale (would have required a lot of conflict resolution given the build/daemon/coordinator churn) and the pyo3 0.28 bump commit you included is now superseded by the version that landed independently.

What carries over from your work:

- The full `SampleBuffer` + `SampleHandler` buffer-protocol design
- The safety contract (send-once, no-view-past-send, no-send-while-views-open)
- The atomic ordering choice (valid flip before DataSample release, view_count increment after fill)
- The context-manager API surface (`with node.send_output_raw(...) as buf:`)
- The example layout (sender/receiver dataflow, manual and context-manager patterns)

What's different in the rescue:

- Rebased against current `main` (4 months of build/daemon changes resolved)
- pyo3 0.28 bump dropped (already in main)
- ABI floor explicitly bumped `abi3-py37 → abi3-py311` and framed as a 1.0 RC change in the CHANGELOG — your original PR implicitly required py311 too, the rescue makes it explicit
- `apis/python/node/build.rs` now calls `pyo3_build_config::use_pyo3_cfgs()` so the `#[cfg(Py_3_N)]` gates actually propagate
- 4 Rust unit tests on the `SampleBufferState` atomic state machine
- New `apis/python/operator/src/lib.rs` `Clone` impl for `DelayedCleanup<T>`
- `validate_output` made `pub` with a doc comment

You're credited via `Co-Authored-By` on the rescue commit. Thanks for the work — it landed because the implementation was clean and the design was right.

PR #1294, opened 2026-01-09
Add heuristic for start of error messages

phil-opp:
Instead of treating the last 10 lines of stderr as the error message, we now look for marker strings such as 'Error:' or 'Traceback' to find the start of error messages more accurately.


haixuanTao:
Fix CI Example error: https://github.com/dora-rs/dora/pull/1295 

Sorry I needed to merge v0.4.0 and publish it first so that I could integrate the change within dora-rs core.

PR #1297, opened 2026-01-09
Provide an explict method to report node failure errors

phil-opp:
Allows nodes to report errors in a structured way before they exit with a non-zero exit code. By using this method, the daemon does not have to try to extract a meaningful error message from `stderr` after node failure.

---

I'm not 100% happy with the `NodeFailureError` and `report_failure_error` names. I wanted to make it clear that this is an error message that corresponds to the node failure, but does not cause/report a node failure itself (you still have to do exit(1)). Still, maybe names like `NodeFailed` and `report_failure` are still better?


phil-opp:
Closing for now. There were no requests for this, so this feature might not be needed. We can reopen if there is demand.

Issue #1212, opened 2025-11-18
Implement `dora config` Command Group


ZhangHanDong:
**Goal**: Unified configuration management interface

**Functional Requirements**:

- `dora config list` - List all configurations
- `dora config get <key>` - Get configuration value
- `dora config set <key> <value>` - Set configuration
- `dora config unset <key>` - Delete configuration
- Support global configuration (`~/.dora/config.toml`)
- Support project configuration (`./dora.toml`, higher priority)

**Example Configuration Items**:

- `coordinator.addr` - Coordinator address
- `coordinator.port` - Coordinator port
- `log.level` - Log level
- `ui.theme` - UI theme
- `ui.color` - Enable colored output

**Technical Points**:

- Introduce `config-rs` or `toml` crate
- Layered configuration files (global < project)
- Environment variable override support

**Acceptance Criteria**:

```bash
$ dora config set coordinator.addr 192.168.1.100
$ dora config get coordinator.addr
192.168.1.100
$ dora config list
coordinator.addr = "192.168.1.100"
coordinator.port = 53290
log.level = "info"
```

**Compatibility**: New command, does not affect existing functionality

**Dependencies**: `config-rs` or `toml` crate

guptapratykshh:
@dora-bot assign me

heyong4725:
Closing — not shipped, declined per the umbrella thread (#1196).

The persistent CLI configuration mechanism proposed here (mirroring `git config` / `cargo config`) is coupled to the `dora.toml` project manifest from #1213, which is itself a speculative ask. Existing workarounds cover the typical use cases:

- `--coordinator-addr` flag (no need to type once per session — script in a shell wrapper)
- Environment variables (e.g. `DORA_ZENOH_CONNECT`)
- The dataflow YAML itself for per-dataflow config

If the lack of persistent CLI defaults becomes a recurring pain point, a fresh focused issue with the specific pain (and ideally the simplest possible scope — e.g. just `coordinator.addr` to start, not a full key-value store) is welcome.

Issue #1196, opened 2025-11-18
Dora CLI Improvement Task Planning V 1.0


ZhangHanDong:
Requirements Analysis

Current Priority Requirements (requires.md):

- P0: Node failure isolation and system fault tolerance
- P1: CLI debugging tool extensions (similar to ROS echo/hz/record)
- P2: DORA and ROS2 performance testing scripts
- P3: Built-in support for industrial field protocols
- P4: Remote debugging based on Zenoh
- P5: Node ACL mechanism

CLI Improvement Vision (cli.md):

- Reference uv's experience to build modern developer tools
- Reorganize command structure into six major categories: system/project/pkg/run/inspect/ui
- Introduce dora.toml project manifest and dora Hub package management system

🎯 Design Principles
All improvements follow these principles:

- ✅ Progressive Enhancement: Each task is independently deliverable without breaking existing functionality
- ✅ Backward Compatibility: Keep existing commands as aliases (e.g., dora list → dora inspect ps)
- ✅ Developer Experience First: Reference best practices from mature tools like uv/cargo/docker
- ✅ Robotics Domain Characteristics: Prioritize debugging tools familiar to ROS developers
- ✅ Function > UI: Ensure functional completeness first, with TUI/GUI as enhancements

haixuanTao:
Could we reduce this list to only 3-5 issues that we think is the most critical for now?

phil-opp:
Many of those issues propose two things at once:

- a new name for some existing command
- changed/improved behavior of some existing command

These things are not really related to each other, so I would prefer if we could discuss them separately. For example, we could improve the `dora list` output without renaming the command to `dora inspect ps`.

Perhaps you can move all the rename requests to a (single) new issue so that we can discuss them separately? If we want to reorganize the CLI commands we should do this in a single PR (that just adds some new aliases). Otherwise, we end up with a mix of "old" and "new" command names so that e.g. `dora system start` exists, but the `dora system stop` is still named `dora destroy`.

heyong4725:
Closing the umbrella per the maintainer feedback that never converged.

Status snapshot as of 2026-05-12:
- 8 children already shipped: #1198, #1199, #1200, #1219, #1220, #1221, #1234, #1235
- 12 children closed as "shipped under existing command name" or "redundant with existing command": #1201, #1202, #1203, #1204, #1206, #1207, #1209, #1211, #1214, #1215, #1216, #1217
[...]

PR #1299, opened 2026-01-10
Feature/dora config command

guptapratykshh:
https://github.com/dora-rs/dora/issues/1212

This PR introduces new dora config command that provides unified configuration management for the Dora CLI. This makes it easier to manage coordinator settings and other configuration options without repeatedly specifying command-line flags

The new command provides a complete configuration management interface:

Configuration Operations: Supports list , get, set, and unset subcommands for managing configuration values with intuitive syntax
Layered Configuration: Implements global (~/.dora/config.toml) and project-level (./dora.toml) configuration files, with project settings automatically overriding global ones
Coordinator Settings: Focuses on essential coordinator configuration (coordinator.addr and coordinator.port)
Developer Experience: Automatically creates configuration directories and files, provides clear feedback messages, and supports nested key notation (e.g., coordinator.addr) for intuitive configuration management


phil-opp (commented):
Thanks for the PR!

The config files seems to have no effect right? So you can set your port number etc but this all has no effect.

You also can set arbitrary values with no checking at all. So you can set the log level to `"Lorem ipsum" and, the address to `true`, and the port number to "I don't care". 

I think the proper way to implement a config system is to:

1. Define a `Config` struct with all the available options that implements Serialize/Deserialize. Unknown fields should not error for backwards compatibility, but we should warn about them to catch typos (e.g. users writing `potr` instead of `port`).
2. Load the config struct at CLI startup.
3. Update the existing commands to default to the config file if the corresponding CLI argument is not given.
4. Provide utility commands in the CLI for updating the config file.

This PR seems to implement step 4, which doesn't make much sense without the prior steps.

heyong4725:
Closing — implements closed issue #1212.

`dora config` was declined as part of the umbrella reduction on #1196 ("reduce to 3-5 critical items"). The umbrella thread closed without converging on which items to keep, and `dora config` was not among the survivors. The relevant existing workarounds (`--coordinator-addr` flag, env vars like `DORA_COORDINATOR_ADDR`, shell wrappers) cover the typical use case.

If a focused need surfaces for persistent CLI defaults — specifically, a recurring pain point that the existing workarounds do not solve — please open a fresh issue describing the pain, and a smaller-scope fix (e.g. just `dora config set coordinator.addr <value>` and `get`, no full key-value store) can be evaluated on its own merits.

Thanks @guptapratykshh for the implementation work.

PR #1300, opened 2026-01-10
Remove Windows early-exit guard in cxx-dataflow example

### LINKED ISSUE #1105: windows  c++  not work
Running `E:\work\dora\target\debug\examples\cxx-dataflow.exe`
2025-08-20T13:59:27.463793Z ERROR cxx_dataflow: The c++ example does not work on Windows currently because of a linker error

afrdbaig7:
Fixes #1105
The example already includes Windows-specific compiler and linker flags for C++ nodes and operators, but the early exit prevented it from being built or run on Windows at all. With this change, the example is allowed to attempt a normal build and execution on Windows, surfacing any remaining issues directly instead of exiting unconditionally.

What changed
Removed the Windows-only early-exit check in examples/c++-dataflow/run.rs

How it was tested
cargo run --example cxx-dataflow on Windows

### DIFF
--- a/examples/c++-dataflow/run.rs
+++ b/examples/c++-dataflow/run.rs
@@ -5,13 +5,6 @@ use std::{
 };
 
 fn main() -> eyre::Result<()> {
-    if cfg!(windows) {
-        tracing::error!(
-            "The c++ example does not work on Windows currently because of a linker error"
-        );
-        return Ok(());
-    }
-
     let root = Path::new(env!("CARGO_MANIFEST_DIR"));
     let target = root.join("target");
     std::env::set_current_dir(root.join(file!()).parent().unwrap())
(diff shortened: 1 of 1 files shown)


afrdbaig7:
> Thanks for the PR!
> 
> The example still seems to fail on Windows CI, not sure why: https://github.com/dora-rs/dora/actions/runs/20881884246/job/59999401440?pr=1300
> 
> I just restarted the build to rule out a temporary error.

afrdbaig7:
You’re right — compared to #1300 the only real change here is adding
compile.arg("-Wl,-nodefaultlib:libcmt"), along with a small README update.

In the earlier PR, I was only fixing the early-exit issue in the example. The Windows linker problem itself wasn’t fully addressed there, which is why CI was still failing.

This PR focuses specifically on that linker issue. The added flag prevents MSVC from pulling in the default static CRT (libcmt), which was causing the conflict on Windows CI. I closed the previous PR to keep things clean and reopened with a minimal change that targets the actual root cause.

I’ve verified this locally on Windows and kept the scope limited to just what’s needed.

PR #1301, opened 2026-01-11
fix #829 : replace deprecated typing.List with list in Python stubs

### LINKED ISSUE #829: Use pyo3 experimental-inspect instead of `generate_stubs.py`
Currently dora use a custom script called `generate_stubs.py` to generate type hinting in python.

Unfortunately, this typing is not perfect as it requires manual type hinting and has many depreciated feature when linted with:

```bash
uvx ruff check --select UP
```

We can try to run pyo3 with the feature flag experimental-inspect and see if this can be a good automatic stub generator. 

To do this, we just need to replace in `Cargo.toml`:

```toml
pyo3 = { version = "0.23", features = [
    "eyre",
    "abi3-py37",
    "multiple-pymethods",
] }
```

with:

```toml
pyo3 = { version = "0.23", features = [
    "eyre",
    "abi3-py37",
    "multiple-pymethods",
    "experimental-inspect"
] }
```

aryansanganti:
Fixes #829
- Changed typing.List[str] to list[str] in Ros2Context.__init__
- Follows PEP 585 for modern Python type hints
- Verified with ruff check --select UP (all checks pass) Note: experimental-inspect was tested by multiple contributors and confirmed not to work for auto-generating stubs.

### DIFF
--- a/apis/python/node/dora/__init__.pyi
+++ b/apis/python/node/dora/__init__.pyi
@@ -190,7 +190,7 @@ class Ros2Context:
     context = Ros2Context()
     ```"""
 
-    def __init__(self, ros_paths: typing.List[str] = None) -> None:
+    def __init__(self, ros_paths: list[str] = None) -> None:
         """ROS2 Context holding all messages definition for receiving and sending messages to ROS2.
 
         By default, Ros2Context will use env `AMENT_PREFIX_PATH` to search for message definition.


haixuanTao:
@claude could you make the proper change so that when the stubs are auto generated it uses the right type

haixuanTao:
@claude could you make the proper change so that when the stubs are auto generated it uses the right type 
You need to fetch a remote branch not in this repo

PR #1303, opened 2026-01-13
Add opentelemetry to the daemon

haixuanTao:
This PR adds Opentelemetry to the daemon and make it possible to receive node failure information within opentelemetry even if the node panic and don't have time itself to send the error logs.

### DIFF
--- a/libraries/extensions/telemetry/tracing/src/lib.rs
+++ b/libraries/extensions/telemetry/tracing/src/lib.rs
@@ -229,10 +229,9 @@ impl Drop for OtelGuard {
 /// ```no_run
 /// use dora_tracing::init_tracing_subscriber;
 /// use tracing::level_filters::LevelFilter;
-/// use tokio::runtime::Builder;
 ///
-/// let rt = Builder::new_multi_thread().enable_all().build().unwrap();
-/// let _enter = rt.enter();
+/// // Note: This function requires a tokio runtime context to be active
+/// // when using OTLP tracing. Use runtime.enter() before calling.
 /// let _guard = init_tracing_subscriber(
 ///     "my-service",
 ///     Some("info"),
@@ -249,9 +248,7 @@ pub fn init_tracing_subscriber(
     let mut builder = TracingBuilder::new(name);
     let guard: Option<OtelGuard>;
 
-    if std::env::var("DORA_OTLP_ENDPOINT").is_ok()
-        || std::env::var("DORA_JAEGER_TRACING").is_ok()
-    {
+    if std::env::var("DORA_OTLP_ENDPOINT").is_ok() || std::env::var("DORA_JAEGER_TRACING").is_ok() {
         builder = builder
             .with_otlp_tracing()
             .wrap_err("failed to set up OTLP tracing")?;


binaries/cli/src/command/run.rs
```diff
@@ -36,22 +36,45 @@ pub fn run_func(dataflow: String, uv: bool) -> eyre::Result<()> {
...
+            } else {
+                let env_log = std::env::var("RUST_LOG").unwrap_or("info".to_string());
+                builder = builder.with_stdout(env_log, false);
+            }
+            builder
+                .build()
+                .wrap_err("failed to set up tracing subscriber")
+                .unwrap();
+            loop {
+                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
phil-opp: What is this sleep in a loop about? This seems very hacky...

binaries/cli/src/command/daemon.rs
```diff
@@ -40,28 +40,50 @@ pub struct Daemon {
...
+                    builder = builder.with_stdout(env_log, false);
+                }
+                builder = builder.with_file(filename, LevelFilter::INFO).unwrap();
+                builder
+                    .build()
+                    .wrap_err("failed to set up tracing subscriber")
+                    .unwrap();
+                loop {
+                    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+                }
```
phil-opp: (same as above) What is this sleep in a loop about? This seems very hacky...

binaries/cli/src/command/daemon.rs
```diff
@@ -40,28 +40,50 @@ pub struct Daemon {
...
+            rt.spawn(async move {
+                let mut builder = TracingBuilder::new(name);
+                if std::env::var("DORA_OTLP_ENDPOINT").is_ok()
+                    || std::env::var("DORA_JAEGER_TRACING").is_ok()
+                {
+                    builder = builder
+                        .with_otlp_tracing()
+                        .context("failed to set up OTLP tracing")
+                        .unwrap();
+                    *clone.lock().unwrap() = builder.guard.take();
```
phil-opp: This part sems identical to the changes in `run.rs`. Perhaps we should move it to a common function?

binaries/cli/src/command/run.rs
haixuanTao: Yeah so, we need to keep the thread alive. so that the opentelemetry exporter export data.

There's probably a better way.

phil-opp: If we sleep in a loop, the thread is blocked forever without doing any work. So whereever the opentelemetry exporter works, it's not in this thread.

So why do we need to spawn an async task at all? The only `await` is in your sleep call?

haixuanTao: Yeah you're right. @claude could you fix this and use the main thread to spawn the tracing process?

haixuanTao: @claude so you need to make sure to use https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter before calling the tracing function so that the tokio runtime context is available when building the tracing subscriber, hence me moving the tokio runtime before starting the tracing function.

haixuanTao: @phil-opp should be fixed now. Let me know If having Claude messages in the middle bothers you. Hopefully it should make train of thoughts cleaner.

phil-opp: Looks good, thanks!

> Let me know If having Claude messages in the middle bothers you. Hopefully it should make train of thoughts cleaner.

You mean the comments? Doesn't bother me, I usually don't read them either (they are quite verbose and just reiterate what was said before). I typically look at the diff instead.

binaries/cli/src/command/daemon.rs
haixuanTao: @claude would it be possible to make the tracing function a common function imported from dora_tracing thanks.

PR #1307, opened 2026-01-13
feat: enhance dora --version with detailed version information

### LINKED ISSUE #1305: Add additional information on dora --version
@claude can you add a more information when running

```dora --version``` 
to not only show the cli version but also show, the dora-message version and if there is python either with uv or not show the pip show dora-rs version and if there is version mismatch raise an error.

haixuanTao:
Closes #1305

Enhances the `dora --version` command to display:
- dora-cli version
- dora-message version
- Python dora-rs version (when installed)
- Version mismatch warning

The implementation tries `uv pip show` first, then falls back to `pip show` for detecting the Python package version.

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -23,7 +23,7 @@ python = ["pyo3"]
 [dependencies]
 arrow = { workspace = true }
 arrow-schema = { workspace = true }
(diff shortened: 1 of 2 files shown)


phil-opp:
This is probably because clap includes the executable name in the version output by default. So we don't need to specify it again.

Issue #885, opened 2025-03-17
Improve Users-Getting Started on dora-rs.ai


haixuanTao:
The current getting started for dora-rs is still a bit missing a lot of components to make it workable by anyone,

A thorough rewrite in order to split concept could really improve users getting started with dora-rs.

Shiv-aurora:
@dora-bot assign me

Shiv-aurora:
@Krishnadubey1008 hey I'm already working on that, so i would apreacitate if you could find another issue to contribute.

Krishnadubey1008:
@Shiv-aurora as I had edited only "Python Conversation" file rest 3 files you can work on

Krishnadubey1008:
@haixuanTao while running the dataflow using instructions in [https://dora-rs.ai/docs/guides/getting-started/conversation_py](url) I had been encountering issue-ImportError: cannot import name 'Node' from 'dora'

![Image](https://github.com/user-attachments/assets/70765a08-dd1f-47d9-9960-661763e3d409)
This was because virtual enviroment was not activated before running commands(such as pip install dora-rs-cli). When I did that I got the correct output 

![Image](https://github.com/user-attachments/assets/45da8d66-b171-49a9-9878-8ac79652d0cf)

Please review my PR [https://github.com/dora-rs/dora-rs.github.io/pull/27](url) .

Shiv-aurora:
Hi @haixuanTao 

I’ve submitted Part 1 of 4 for this issue in PR  [dora-rs/dora-rs.github.io#28](https://github.com/dora-rs/dora-rs.github.io/pull/28) 
Please review it and let me know if you have any feedback before I proceed with the remaining parts!

kingchou007:
@dora-bot assign me

LIKHITHA-NAGRAJ:
@dora-bot assign me

PR #1314, opened 2026-01-15
Fix event stream thread shutdown hang (#642)

### LINKED ISSUE #642: Node panic on `timeout while waiting for event stream thread`
Not exactly sure what is happening, but it seems that I have a bug where the node completely freeze and then report:

```bash
[2m2024-08-30T16:15:37.258067Z[0m [33m WARN[0m [2mdora_node_api::event_stream::thread[0m[2m:[0m timeout while waiting for event stream thread
```

Then the node finishes. 

I wonder if there is like a break condition on: https://github.com/dora-rs/dora/blob/9be5b765b94a8212011f3cb6e5b4e74f2f487269/apis/rust/node/src/event_stream/thread.rs#L70-L72

[Screencast from 2024-08-30 18-29-53.webm](https://github.com/user-attachments/assets/460f2a63-6cf9-41c3-9e70-1597934cf3ac)

afrdbaig7:
Fixes #642 by preventing the event stream thread from timing out during shutdown. Previously, the thread could remain blocked in its event loop when the handle was dropped, causing the runtime to wait for one second and emit a "timeout while waiting for event stream thread" warning. This change introduces cooperative shutdown via a signal sent on drop, allowing the event loop to detect shutdown and exit cleanly. The existing one-second timeout is retained as a safety net for unexpected edge cases.

### DIFF
--- a/apis/rust/node/src/event_stream/thread.rs
+++ b/apis/rust/node/src/event_stream/thread.rs
@@ -21,9 +21,15 @@ pub fn init(
     channel: DaemonChannel,
     clock: Arc<uhlc::HLC>,
 ) -> eyre::Result<EventStreamThreadHandle> {
+    let (shutdown_tx, shutdown_rx) = flume::bounded(1);
     let node_id_cloned = node_id.clone();
-    let join_handle = std::thread::spawn(|| event_stream_loop(node_id_cloned, tx, channel, clock));
-    Ok(EventStreamThreadHandle::new(node_id, join_handle))
+    let join_handle =
+        std::thread::spawn(|| event_stream_loop(node_id_cloned, tx, channel, clock, shutdown_rx));
+    Ok(EventStreamThreadHandle::new(
+        node_id,
+        join_handle,
+        shutdown_tx,
+    ))
 }
 
 #[derive(Debug)]
@@ -40,10 +46,15 @@ pub enum EventItem {
 pub struct EventStreamThreadHandle {
     node_id: NodeId,
     handle: flume::Receiver<std::thread::Result<()>>,
+    shutdown_tx: flume::Sender<()>,
 }
 
 impl EventStreamThreadHandle {
-    fn new(node_id: NodeId, join_handle: std::thread::JoinHandle<()>) -> Self {
+    fn new(
+        node_id: NodeId,
+        join_handle: std::thread::JoinHandle<()>,
+        shutdown_tx: flume::Sender<()>,
+    ) -> Self {
         let (tx, rx) = flume::bounded(1);
         std::thread::spawn(move || {
             let _ = tx.send(join_handle.join());
@@ -51,12 +62,16 @@ impl EventStreamThreadHandle {
         Self {
             node_id,
             handle: rx,
+            shutdown_tx,
         }
     }
 }
 
 impl Drop for EventStreamThreadHandle {
     fn drop(&mut self) {
+        // Signal the event stream thread to shutdown gracefully
+        let _ = self.shutdown_tx.send(());
+
         if self.handle.is_empty() {
             tracing::trace!("waiting for event stream thread");
         }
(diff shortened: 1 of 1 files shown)


afrdbaig7:
Thanks a lot for the detailed breakdown — that matches my understanding of the normal shutdown flow as well 
I agree with your analysis that an additional exit check at the start of the main loop does not help in the two common cases you outlined:
If the thread is busy forwarding events, it will eventually notice tx is closed.
If it is blocked waiting on DaemonRequest::NextEvent, it cannot observe any extra condition anyway.

Where I think the reported hang in #642 may still be coming from is after the loop exits, during cleanup — specifically in report_remaining_drop_tokens.
At that point:
The background thread has already exited the main event loop correctly.
The node API is already gone, so no acknowledgements can arrive.
report_remaining_drop_tokens still waits up to 100ms per pending token.
If a node has 10+ pending events at drop time, that's potentially 1+ seconds of blocking in cleanup alone, exceeding the timeout even though the main loop exited correctly in EventStreamThreadHandle::drop, even though the exit from the main loop was correct. In that case, the “hang” isn’t caused by event forwarding or daemon blocking, but by best-effort cleanup that can no longer make progress.
That’s the specific case I was trying to address with an explicit “forced shutdown” signal:
Natural shutdown: daemon closes the stream → we wait and report remaining tokens.
Forced drop: EventStream is dropped → skip waiting and exit immediately.
I agree that checking tx.is_closed() is sufficient for loop termination, and I’m happy to remove the early-loop exit condition. My intent wasn’t to add another way out of the loop, but to give report_remaining_drop_tokens a clear signal that it should not block when shutdown is forced.
If that approach still doesn’t seem appropriate, I’m very open to alternative ways to make the cleanup phase non-blocking in the forced-drop case. Let me know what you think — I may be missing something here.

Thanks again for the thorough review!

heyong4725:
Hi @afrdbaig7 — really appreciate you taking a careful look at this and following up on @phil-opp's review (and going back to first principles when the original loop-level check turned out not to address the actual hang). That kind of follow-through is exactly what makes a good contributor.

Going to close this PR as no longer applicable, and I want to explain *why* so it doesn't feel like the time was wasted.

**1. The bug being fixed is officially stale.** I closed [#642](https://github.com/dora-rs/dora/issues/642) on 2026-05-12 — no fresh repro on dora 1.0, and the event-stream subsystem was rewritten as part of the 1.0 consolidation. Whether the original race (Python node with multiple subscribers to a camera output, dora 0.x) still triggers is unverified.

**2. The subsystem rewrite removed the code your PR targeted.** Specifically:

- The `flume::Sender<EventItem>` your patch adapts is now `mpsc::Sender<EventItem>` — migrated to tokio in [#1603](https://github.com/dora-rs/dora/pull/1603) to fix an AB-BA deadlock between flume 0.10's spinlock and pyo3's GIL-acquiring waker (see comment at `apis/rust/node/src/event_stream/thread.rs:148-150`).
- The `report_remaining_drop_tokens` path you identified as the real culprit in your Feb 1 reply (the 100ms-per-pending-token cleanup that could stretch to 1+ seconds) **no longer exists** — the entire drop-token cleanup phase was removed. The loop is now just the two exit conditions phil-opp described: `events.is_empty()` from the daemon, and `tx.blocking_send` error when the EventStream is dropped.

**3. Your own Feb 1 analysis is correct.** Once `report_remaining_drop_tokens` is out of the picture, there's no cleanup phase left to short-circuit, so the early-loop signal check has nothing meaningful to do.

Net: the PR's diff would no longer apply cleanly, and even rebased, the motivating problem is gone.

If you'd like to keep contributing, here are a few currently-applicable starting points:

- [#885](https://github.com/dora-rs/dora/issues/885) — Improve Users-Getting Started on dora-rs.ai (`good first issue`)
- [#829](https://github.com/dora-rs/dora/issues/829) — Use pyo3 experimental-inspect instead of `generate_stubs.py` (`good first issue`, touches the Python API surface)
- [#540](https://github.com/dora-rs/dora/issues/540) — Thread-safety of C node API (`good first issue`, more involved but a good system-level investigation)

Or if you reproduce a current hang on dora 1.0 (Rust or Python), please open a fresh issue — the 2024 repro pattern is no longer a useful starting point given the rewrite, but a current repro on `apis/rust/node/src/event_stream/thread.rs` (current head) would be very interesting.

PR #1322, opened 2026-01-19
Fix: Expose timestamp as native Python datetime.datetime in metadata (#1315)

### LINKED ISSUE #1315: python: Export event timestamp

I'm a developer of OpenArm. We're evaluating dora-rs as a data collection framework for OpenArm.

For example, we want to record teleoperation data (leader positions, follower positions, images from cameras and so on) with timestamp and use them to learn models later.

`dora.Node.next` returns the next event as a Python dict. It's implemented here: https://github.com/dora-rs/dora/blob/6ac8a75232befcb3586a1c7fbd6c630e01d72b3a/apis/python/operator/src/lib.rs#L80-L145

In an event Python dict, timestamp data that exists in Rust: https://github.com/dora-rs/dora/blob/6ac8a75232befcb3586a1c7fbd6c630e01d72b3a/libraries/message/src/metadata.rs#L12

If we can use the timestamp from Python, we don't need to send an event with timestamp manually.

**Describe the solution you'd like**

Can we add the timestamp data in Rust to a Python dict that is returned by `dora.Node.next`?

**Describe alternatives you've considered**

We may want to return an event object instead of a Python dict by `dora.Node.next` that has `timestamp` method that returns timestamp associated with the event.

**Additional context**

None.

gaumrab:
Fixes #1315

## Problem

The timestamp from Rust metadata (`uhlc::Timestamp`) was not exposed to Python users when calling `dora.Node.next()`. Users need access to timestamps for recording teleoperation data (leader positions, follower positions, camera images, etc.) with timestamps for later model training.

**Requirement:** Return the timestamp as a native Python `datetime.datetime` object (not string or float) with UTC timezone.

## Solution

This PR fixes `metadata_to_pydict()` in `apis/python/operator/src/lib.rs` to:
- Extract timestamp from Rust metadata using standard `get_time().to_system_time()` pattern
- Convert to Python `datetime.datetime` object using Python's datetime module
- Make it timezone-aware with UTC
- Add to metadata dictionary as `"timestamp"` key

## Changes

- **File Modified:** `apis/python/operator/src/lib.rs`
- **Function:** `metadata_to_pydict()`
- **Lines Added:** ~40 lines of well-documented code

## Comparison with Alternative Approach

| Aspect | Alternative (returns float) | This PR |
|--------|----------------------------|---------|
| Return Type | `float` (Unix timestamp) | `datetime.datetime` object ✅ |
| User Experience | Requires manual conversion | Ready to use ✅ |
| Timezone | User must add | UTC-aware by default ✅ |
| Code Pattern | Manual NTP64 bit manipulation | Standard `get_time().to_system_time()` ✅ |

## Testing

✅ Tested locally with minimal dataflow:
- Timestamp exists in `event["metadata"]["timestamp"]`
- Type is `datetime.datetime`
- Timezone is UTC
- Ready to use without manual conversion

**Test output:**
image

### DIFF
--- a/apis/python/operator/src/lib.rs
+++ b/apis/python/operator/src/lib.rs
@@ -247,7 +247,7 @@ pub fn metadata_to_pydict<'a>(
     py: Python<'a>,
 ) -> Result<pyo3::Bound<'a, PyDict>> {
     let dict = PyDict::new(py);
-    
+
     // Add timestamp as timezone-aware Python datetime (UTC)
     // Note: uhlc::Timestamp is a Hybrid Logical Clock. We use get_time().to_system_time()
     // which extracts the physical clock component. This pattern is used consistently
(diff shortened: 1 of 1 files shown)


haixuanTao (approved):
Looks good! thanks!

PR #1326, opened 2026-01-21
Fix busy polling in EventStream::recv_async to reduce CPU usage

RustyCoderX:
Fix CPU leak caused by busy polling in recv_async

Remove unnecessary micro-delay polling and return scheduled events immediately
when available, preventing sustained high CPU usage in long-running nodes.


haixuanTao:
I think you're not up to date with main

PR #1328, opened 2026-01-22
Add hot-reload support for custom node binaries

haixuanTao:
This adds the --hot-reload flag to both `dora start` and `dora run` commands. When enabled, the CLI watches node binary files for changes and automatically restarts the node when the file is modified.

Features:
- Shared hot_reload module used by both `dora start` and `dora run`
- Watches parent directories to handle pip install (which deletes/recreates files)
- Uses `which::which()` to resolve pip-installed package paths
- Nodes spawned with restart=always when hot_reload is enabled
- Uses SIGTERM (SoftKill) for graceful node restart

Usage examples:

For Local Python script: ```path: ./my_node.py ``` 
3. Run the dataflow with:
```bash
 dora run/start dataflow.yaml --hot-reload 
```

3. Edit my_node.py -> node automatically restarts

Or:

For Pip-installed package: ```path: opencv-video-capture ``` 
```bash 
dora run/start dataflow.yaml --hot-reload --attach 
```
2. Run `pip install -e ./my-package` again -> node automatically restarts ```


binaries/cli/src/command/start/attach.rs
```diff
@@ -42,64 +39,23 @@ pub fn attach_dataflow(
...
-                            }))
-                            .context("Could not send reload request to the cli loop")
-                            .unwrap();
-                    }
-                }
-                // TODO: Manage different file event
-            }
-        };
+        let (watcher, reload_rx) = setup_file_watcher(dataflow_id, &nodes, &working_dir)
+            .context("failed to setup file watcher")?;
```
phil-opp: The challenge with `dora start` is that the nodes might not be running on the same machine as the CLI. They could be running on a remote daemon, then the watching would need to happen there.

binaries/cli/src/command/run.rs
```diff
@@ -27,36 +32,68 @@ pub struct Run {
...
+            .context("failed to canonicalize dataflow path")?
+            .parent()
+            .ok_or_else(|| eyre::eyre!("canonicalized dataflow path has no parent"))?
+            .to_owned();
+
+        let descriptor = Descriptor::blocking_read(&dataflow_path).context("could not read dataflow")?;
+        let nodes = descriptor.resolve_aliases_and_set_defaults()?;
+
+        // Use a temporary dataflow_id - the actual one is assigned in the daemon
+        let dataflow_id = Uuid::nil();
```
phil-opp: Why not set up the file watching in the daemon? This way we don't need the hack with the temp dataflow id. We also avoid parsing and interpreting the source/path in two places, which might lead to inconsistencies.

My intuition would say that the we should set up the file watching just before spawning the executables in the daemon.

binaries/daemon/src/spawn/prepared.rs
```diff
@@ -49,6 +49,8 @@ pub struct PreparedNode {
     pub(super) clock: Arc<HLC>,
     pub(super) daemon_tx: mpsc::Sender<Timestamped<Event>>,
     pub(super) node_stderr_most_recent: Arc<ArrayQueue<String>>,
+    /// When true, custom nodes use restart=always for hot-reload support.
```
phil-opp: It's not clear to me how `restart=always` and hot-reload interact. Is the `restart=always` required in order for hot reloading to work? If so, why? Is hot reloading enabled automatically?

binaries/daemon/src/spawn/prepared.rs
```diff
@@ -94,7 +96,14 @@ impl PreparedNode {
 
     fn restart_policy(&self) -> RestartPolicy {
         match &self.node.kind {
-            dora_core::descriptor::CoreNodeKind::Custom(n) => n.restart_policy,
+            dora_core::descriptor::CoreNodeKind::Custom(n) => {
+                // When hot_reload is enabled, always restart custom nodes
+                if self.hot_reload {
+                    RestartPolicy::Always
```
phil-opp: We could just restart the node if it exits after a hot-reload event. Then we don't need to change the restart policy (i.e. the node would still exit as normal when it exits without a prior hot-reload event).

binaries/daemon/src/lib.rs
```diff
@@ -245,12 +247,46 @@ impl Daemon {
...
+                while let Ok((node_id, operator_id)) = receiver.recv() {
+                    if tx.blocking_send((node_id, operator_id)).is_err() {
+                        break;
+                    }
+                }
+            });
+            tracing::info!("Hot-reload enabled: {} (file watching active)", hot_reload);
+        } else {
+            tracing::info!("Hot-reload enabled: {} (no file watcher provided)", hot_reload);
+        }
```
phil-opp: We could make the `reload_receiver` a tokio mpsc channel, then we could use it directly without any forwarding.

binaries/daemon/src/lib.rs
```diff
@@ -1765,11 +1811,30 @@ impl Daemon {
...
-        if let Some(channel) = dataflow.subscribe_channels.get(&node_id) {
-            match send_with_timestamp(channel, NodeEvent::Reload { operator_id }, &self.clock) {
-                Ok(()) => {}
-                Err(_) => {
-                    dataflow.subscribe_channels.remove(&node_id);
+
+        // If operator_id is None, this is a custom node reload (hot-reload of binary)
+        // Use SoftKill (SIGTERM) to terminate the process directly - this is more ergonomic
+        // for development as it doesn't require nodes to handle STOP events.
+        // The node will restart because it was spawned with restart=always when hot_reload was enabled.
```
phil-opp: As I said in my other comment: I think it would be cleaner to send a Stop event to enable the node to properly shut down. It could also be important for the node to know that a reload it about to happen, so that it can save some state that can be restored after the reload. So I think a Stop event with a new stop cause would be the cleanest. We can still kill if the node doesn't react in time.

binaries/daemon/src/spawn/prepared.rs
haixuanTao: So hot reloading does not respawn nodes as of now. Only the restart policy.

binaries/daemon/src/lib.rs
haixuanTao: Sounds good!

libraries/message/src/cli_to_coordinator.rs
```diff
@@ -44,6 +44,10 @@ pub enum ControlRequest {
         local_working_dir: Option<PathBuf>,
         uv: bool,
         write_events_to: Option<PathBuf>,
+        /// When true, custom nodes will be spawned with restart=always
+        /// to support hot-reload functionality.
```
phil-opp: This is no longer true

apis/rust/node/src/event_stream/event.rs
```diff
@@ -1,5 +1,6 @@
 use dora_arrow_convert::ArrowData;
 use dora_core::config::{DataId, OperatorId};
+pub use dora_message::daemon_to_node::StopReason;
```
phil-opp: Can we also add a `StopCause` alias to avoid the breaking change in the Rust API?

apis/rust/node/src/event_stream/mod.rs
```diff
@@ -348,7 +348,7 @@ impl EventStream {
         if let Some(write_events_to) = &mut self.write_events_to {
             let event_json = match event {
                 EventItem::NodeEvent { event, .. } => match event {
-                    NodeEvent::Stop => {
+                    NodeEvent::Stop { .. } => {
```
phil-opp: It would be useful to record the stop reason too

binaries/daemon/src/lib.rs
```diff
@@ -1840,14 +2003,312 @@ impl Daemon {
...
+                        },
+                        &self.clock,
+                    );
+                }
+
+                // Set pending_hot_reload flag so the restart loop will restart the node
+                let running_node = dataflow.running_nodes.get(&node_id).unwrap();
+                running_node
+                    .pending_hot_reload
+                    .store(true, std::sync::atomic::Ordering::Release);
```
phil-opp: To avoid races, we should set the `pending_hot_reload` flag before sending the stop event

haixuanTao: indeed thanks!

apis/rust/node/src/event_stream/mod.rs
haixuanTao: done!

apis/rust/node/src/event_stream/event.rs
haixuanTao: So I renamed everything StopCause to avoid confusion

PR #1329, opened 2026-01-23
feat: Add datetime.datetime support in pydict_to_metadata() for metadata forwarding

### LINKED ISSUE #1323: python: Add support for `datetime.datetime` value in metadata

#1322 exported timestamp data in Rust as `datetime.datetime` in Python as one of metadata. But we can't send `datetime.datetime` as metadata by `node.send_output()`. It causes a problem when we forward metadata from another input.

For example,  https://github.com/dora-rs/dora-hub/tree/main/node-hub/opencv-video-capture forwards `tick`'s metadata: https://github.com/dora-rs/dora-hub/blob/bf45f69c09368f17065880796ad05fe5305b172f/node-hub/opencv-video-capture/opencv_video_capture/main.py#L140-L162

In this case, `node.send_output()` fails with `could not convert type 2026-01-21 05:12:21.992886+00:00`.

**Describe the solution you'd like**

How about adding support `datetime.datetime` in `pydict_to_metadata()`?
https://github.com/dora-rs/dora/blob/27cf18cd76aed1ca635162b83411adc2e035b36c/apis/python/operator/src/lib.rs#L205-L243

If `pydict_to_metadata()` accepts `datetime.datetime`, we can use `datetime.datetime` in metadata of `node.send_output()`.

**Describe alternatives you've considered**

If we don't want to accept `datetime.datetime` as metadata, it may be better that we represent `timestamp` as `int` (like in Rust) not `datetime.datetime`. It will reduce conversion overhead.

**Additional context**

None.

gaumrab:
## Problem

When forwarding metadata containing `datetime.datetime` objects through `node.send_output()`, the system would fail with:

could not convert type 2026-01-21 05:12:21.992886+00:00

This blocked the opencv-video-capture use case where metadata (including timestamps) needs to be forwarded from input to output events. For example, [opencv-video-capture](https://github.com/dora-rs/dora-hub/tree/main/node-hub/opencv-video-capture) forwards tick's metadata, but `node.send_output()` fails when encountering `datetime.datetime` objects in the metadata.

### 1. Added native timestamp support in Rust metadata
- Added `Parameter::Timestamp(chrono::DateTime<Utc>)` variant to the `Parameter` enum
- This provides first-class timestamp representation in metadata, enabling proper serialization/deserialization

### 2. Enhanced `pydict_to_metadata()` to detect datetime.datetime
- Added detection logic for `datetime.datetime` objects **before** the fallback handler
- Converts `datetime.datetime` → `Parameter::Timestamp(chrono::DateTime<Utc>)`
- Ensures datetime objects are properly recognized instead of falling back to string conversion

### 3. Enhanced `metadata_to_pydict()` for round-trip conversion
- Converts `Parameter::Timestamp` back to timezone-aware Python `datetime.datetime` objects
- Preserves timezone information and microsecond precision

### 4. Extended C++ bindings and CLI
- Added Timestamp support to C++ API (getter/setter methods, JSON serialization)
- Updated CLI display to format timestamps as ISO 8601 strings

## Key Implementation Details

The datetime detection happens **before** the fallback in `pydict_to_metadata()`, ensuring datetime objects are properly recognized:

// Check if it's a datetime.datetime object
let datetime_module = PyModule::import(value.py(), "datetime")?;
let datetime_class = datetime_module.getattr("datetime")?;

if value.is_instance(datetime_class.as_ref())? {
    // Convert datetime.datetime → Parameter::Timestamp
    let timestamp_float: f64 = value.call_method0("timestamp")?.extract()?;
    // ... conversion logic ...
    parameters.insert(key, Parameter::Timestamp(dt))
} else {
    // Fallback only for truly unknown types
    println!("could not convert type {value}");
    parameters.insert(key, Parameter::String(value.str()?.to_string()))
}## Testing

✅ **All tests passing with zero "could not convert type" warnings**
- Basic forwarding: 4/4 messages received correctly with datetime.datetime preserved
- Forwarding through intermediate nodes: Works seamlessly
- OpenCV-video-capture simulation: **10/10 frames forwarded successfully, zero warnings**

✅ **Edge cases verified**: 
- Past timestamps (2020)
- Future timestamps (2030)
- Microsecond precision
- Timezone preservation (UTC)

**I've also attached a few screenshots from local testing showing successful forwarding with no conversion errors.**

## Files Changed

- `libraries/message/src/metadata.rs` - Added `Parameter::Timestamp` variant
- `apis/python/operator/src/lib.rs` - Datetime detection in `pydict_to_metadata()` and conversion in `metadata_to_pydict()`
- `apis/python/operator/Cargo.toml` - Added chrono dependency
- `apis/c++/node/src/lib.rs` - C++ API support
- `apis/c++/node/Cargo.toml` - Added chrono dependency
- `binaries/cli/src/command/topic/echo.rs` - Display formatting
- `Cargo.lock` - Updated dependencies

## Impact

- ✅ Fixes opencv-video-capture metadata forwarding use case
- ✅ Enables seamless `datetime.datetime` usage in metadata without workarounds
- ✅ Maintains backward compatibility - existing code continues to work
- ✅ Preserves timezone information and microsecond precision

## Related Issues

Closes #1323 

Screenshot From 2026-01-23 18-15-13
Screenshot From 2026-01-23 18-09-30

### DIFF
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -2,8 +2,6 @@
 
 Dora supports nodes written in C++ through this API crate.
 
-**Note:** This API requires C++20 for timestamp support with `std::chrono::utc_clock`.
-
 ## Build
 
 - Clone the `dora` repository:
@@ -105,39 +103,12 @@ std::cout << "Input ID: " << input_info.id << std::endl;
 auto metadata = std::move(input_info.metadata);
 std::cout << "Metadata timestamp: " << metadata->timestamp() << std::endl;
 
-// Working with timestamp metadata (C++20 required)
-// The API uses i64 (nanoseconds since Unix epoch) for timestamp parameters
-// You can convert to/from std::chrono::time_point<std::chrono::utc_clock>:
+// Timestamp parameters use i64 nanoseconds since Unix epoch
+auto nanos = metadata->get_timestamp("sensor_time");
+metadata->set_timestamp("event_time", nanos);
 
-#include <chrono>
-
-// Getting a timestamp parameter
-auto nanos_result = metadata->get_timestamp("sensor_time");
-if (nanos_result.has_value()) {
-    int64_t nanos = nanos_result.value();
-    
-    // Convert to C++20 std::chrono::time_point<std::chrono::utc_clock>
-    auto tp = std::chrono::utc_clock::time_point(std::chrono::nanoseconds(nanos));
-    
-    // Use the time_point as needed
-    // For example, convert to system_clock for display:
-    auto sys_tp = std::chrono::utc_clock::to_sys(tp);
-    auto time_t = std::chrono::system_clock::to_time_t(sys_tp);
-    std::cout << "Sensor time: " << std::ctime(&time_t) << std::endl;
-}
-
-// Setting a timestamp parameter
-// Create a time_point from current time
-auto now = std::chrono::utc_clock::now();
-auto nanos_since_epoch = std::chrono::duration_cast<std::chrono::nanoseconds>(
-    now.time_since_epoch()
-).count();
-
-// Set the timestamp
-auto set_result = metadata->set_timestamp("event_time", nanos_since_epoch);
-if (!set_result.error.empty()) {
-    std::cerr << "Error setting timestamp: " << set_result.error << std::endl;
-}
+// Convert to std::chrono::utc_clock::time_point (C++20)
+auto tp = std::chrono::utc_clock::time_point(std::chrono::nanoseconds(nanos));
 
 auto keys = metadata->list_keys();
 for (std::size_t i = 0; i < keys.size(); i++) {
--- a/examples/c++-arrow-dataflow/node-rust-api/main.cc
+++ b/examples/c++-arrow-dataflow/node-rust-api/main.cc
@@ -97,6 +97,11 @@ std::shared_ptr<arrow::Array> receive_and_print_input(rust::cxxbridge1::Box<Dora
                 std::cout << "]" << std::endl;
                 break;
             }
+            case MetadataValueType::Timestamp: {
+                auto nanos = metadata->get_timestamp(key);
+                std::cout << "Metadata[" << key << "] (timestamp) = " << nanos << " ns" << std::endl;
+                break;
+            }
             default:
                 std::cout << "Metadata[" << key << "] has unsupported type" << std::endl;
                 break;


apis/python/operator/src/lib.rs
```diff
@@ -234,8 +235,39 @@ pub fn pydict_to_metadata(dict: Option<Bound<'_, PyDict>>) -> Result<MetadataPar
...
+                    let timestamp_float: f64 = value
+                        .call_method0("timestamp")?
+                        .extract()
+                        .context("Failed to extract timestamp from datetime")?;
+
+                    // Convert to chrono::DateTime<Utc>
+                    // timestamp() returns seconds since epoch as float
+                    // Convert to SystemTime first, then to DateTime<Utc>
+                    let seconds = timestamp_float as i64;
+                    let nanos = ((timestamp_float - seconds as f64) * 1_000_000_000.0) as u64;
```
kou: How about using `u32` here to reduce casts for `std::time::Duration::new()`?

```suggestion
                    let nanos = ((timestamp_float - seconds as f64) * 1_000_000_000.0) as u32;
```

apis/c++/node/src/lib.rs
```diff
@@ -140,6 +142,7 @@ mod ffi {
         fn get_list_int(self: &Metadata, key: &str) -> Result<Vec<i64>>;
         fn get_list_float(self: &Metadata, key: &str) -> Result<Vec<f64>>;
         fn get_list_string(self: &Metadata, key: &str) -> Result<Vec<String>>;
+        fn get_timestamp(self: &Metadata, key: &str) -> Result<String>;
```
kou: If we can use C++20, how about using `std::chrono::time_point<sd::chrono::utc_clock>`?

* https://en.cppreference.com/w/cpp/chrono/time_point.html
* https://en.cppreference.com/w/cpp/chrono/utc_clock.html

apis/python/operator/src/lib.rs
phil-opp: Or we could just use `Duration::try_from_secs_f64`, no?

apis/python/operator/src/lib.rs
```diff
@@ -306,6 +338,26 @@ pub fn metadata_to_pydict<'a>(
...
+                let microseconds = dt.timestamp_subsec_micros();
+
+                // Get UTC timezone from Python's datetime module
+                let datetime_module =
+                    PyModule::import(py, "datetime").context("Failed to import datetime module")?;
+                let datetime_class = datetime_module.getattr("datetime")?;
+                let utc_timezone = datetime_module.getattr("timezone")?.getattr("utc")?;
+
+                // Create timezone-aware datetime using fromtimestamp
+                let total_seconds = timestamp as f64 + microseconds as f64 / 1_000_000.0;
```
phil-opp, line 352: Why not use `as_secs_f64`?

apis/python/operator/src/lib.rs
gaurav-takhi: @phil-opp Agreed. Duration::try_from_secs_f64() is cleaner. I'll update the code to use it.

apis/python/operator/src/lib.rs
gaurav-takhi: @phil-opp chrono::DateTime<Utc> doesn't have as_secs_f64().

apis/c++/node/src/lib.rs
gaurav-takhi: Thanks for the suggestion! I noticed -std=c++17 in our example build files, but I haven’t confirmed which C++ standard the API bindings actually use. If C++20 is acceptable for the project, I’d be happy to use std::chrono::time_point<std::chrono::utc_clock>.
@phil-opp any thoughts on this?

phil-opp: @haixuanTao @heyong4725 Any thoughts on this? Is requiring c++20 ok or do we want to stick to c++17?

phil-opp: Let's go ahead and switch to C++20 and the time_point type, and see if there are any issues. Thanks!

gaurav-takhi: ok

apis/c++/node/src/lib.rs
```diff
@@ -342,6 +347,7 @@ impl Metadata {
                     .map(|value| JsonValue::String(value.clone()))
                     .collect(),
             )),
+            DoraParameter::Timestamp(dt) => Ok(JsonValue::String(dt.to_rfc3339())),
```
haixuanTao: Sorry shouldn't this be anything else but a json string?

binaries/cli/src/command/topic/echo.rs
```diff
@@ -229,6 +229,9 @@ async fn log_to_terminal(
                             Parameter::Float(value) => serde_json::to_string(value).unwrap(),
                             Parameter::ListFloat(value) => serde_json::to_string(value).unwrap(),
                             Parameter::ListString(value) => serde_json::to_string(value).unwrap(),
+                            Parameter::Timestamp(dt) => {
+                                serde_json::to_string(&dt.to_rfc3339()).unwrap()
+                            }
```
haixuanTao: Same thing, I don't understand why we need to serialize here

apis/c++/node/src/lib.rs
gaurav-takhi: @haixuanTao You're right. Since Parameter derives Serialize and chrono::DateTime<Utc> serializes to RFC3339, the manual to_rfc3339() in echo.rs is unnecessary and double-serializes. I'll remove it and let serde handle it. The parameter_to_json() case is fine since it builds JsonValue manually. I'll update echo.rs accordingly.

binaries/cli/src/command/topic/echo.rs
gaurav-takhi: Since Parameter implements Serialize, we should serialize the entire variant instead of extracting inner values. I'll simplify it to:
`**let value = serde_json::to_string(v).unwrap();**`
This removes the match and handles all variants (including Timestamp) uniformly. 
**What This Solves:**
Replaces manual JSON construction with serde’s automatic serialization, ensuring all Parameter variants (including Timestamp) are handled correctly and consistently.

phil-opp (commented):
I don't see any use of the `std::chrono::utc_clock` in the C++ API. Instead it now returns `i64`. I don't think that a C++20 update is required for that...

Can we use the `std::chrono::utc_clock` type directly?

gaumrab:
@phil-opp 

I'd like to use `std::chrono::utc_clock` directly, but the CXX bridge (Rust-C++ interop) doesn't support passing `std::chrono::time_point` across the FFI. It only supports basic types like `i64`, `String`, etc.

I can work around this by:
1. Creating a wrapper struct that holds the time internally, with C++ helpers to convert to/from `std::chrono::utc_clock::time_point`
2. Keeping `i64` in the FFI and adding C++ convenience functions that use chrono types

Which do you prefer? If there's a better way to handle this, I'm open to suggestions.

Thanks!

kou:
Oh, I think that `i64` is reasonable.

haixuanTao (approved):

PR #1332, opened 2026-01-26
Implement ROS2 Service Client for Python API with TypedService and QosPolicies fix

### LINKED ISSUE #1170: Tracking support for dora-ros2 bridge
Currently, the dora-ros2 bridge is limited. Some functions of ros2 are not aligned with Dora.
This issue list the features waiting to be supported.

- [ ] Rust API
  - [x]  topic pub/sub
  - [x] service server
  - [x] service client
  - [ ] action server
  - [ ] action client with feedback on #1194

- [ ] C++ API
  - [x] topic pub/sub
  - [ ] service server 
  - [x] service client
  - [ ] action server
  - [ ] action client with feedback, on #1148, #1194

- [ ] Python API
  - [x] topic pub/sub
  - [ ] service client
  - [ ] service server
  - [ ] action client
  - [ ] action server

- [ ] others (some of them maybe available with separated nodes)
  - [ ] parameter service 
  - [ ] examples aligned with [ROS2's minimal examples](https://github.com/ros2/examples/tree/rolling)

...

guptapratykshh:
Fixes #1170

Added support for the ROS2 Service Client in Python API and this introduces new Ros2Client class and 
create_client method in Ros2Node.noe, Python nodes can discover services, wait for them to be available, and make blocking service calls. The update includes TypedServic eand OwnedTypedValue to connect dynamic Arrow-based data with the static ros2-client type system. It also updates Ros2Context to load and register service request and response message definitions correctly.

### DIFF
--- a/libraries/extensions/ros2-bridge/python/src/lib.rs
+++ b/libraries/extensions/ros2-bridge/python/src/lib.rs
@@ -1,14 +1,14 @@
 use std::{
     borrow::Cow,
+    cell::RefCell,
     collections::HashMap,
     path::{Path, PathBuf},
     sync::Arc,
-    cell::RefCell,
 };
 
 use ::dora_ros2_bridge::{ros2_client, rustdds};
 use arrow::{
-    array::{ArrayData, make_array, ArrayRef},
+    array::{ArrayData, ArrayRef, make_array},
     pyarrow::{FromPyArrow, ToPyArrow},
 };
 use dora_ros2_bridge_msg_gen::types::Message;
@@ -306,8 +306,11 @@ impl Ros2Node {
                 ros2_client::ServiceMapping::Enhanced,
                 &ros2_client::Name::parse(name).map_err(|e| eyre::eyre!(e))?,
                 &ros2_client::ServiceTypeName::new(namespace_name, service_name),
-                qos.clone().map(Into::into).unwrap_or(rustdds::QosPolicies::builder().build()),
-                qos.map(Into::into).unwrap_or(rustdds::QosPolicies::builder().build()),
+                qos.clone()
+                    .map(Into::into)
+                    .unwrap_or(rustdds::QosPolicies::builder().build()),
+                qos.map(Into::into)
+                    .unwrap_or(rustdds::QosPolicies::builder().build()),
             )
             .map_err(|e| eyre::eyre!("failed to create updated service client: {e:?}"))?;
 
@@ -506,7 +509,6 @@ impl Stream for Ros2SubscriptionStream {
     }
 }
 
-
 thread_local! {
     static CURRENT_TYPE_INFO: RefCell<Option<TypeInfo<'static>>> = RefCell::new(None);
 }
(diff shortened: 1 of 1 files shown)


heyong4725:
Hi @guptapratykshh — first off, apologies for the long silence. You pinged @phil-opp on March 17 asking for review, and you haven't heard anything back in two months. That's not on you, and I want to be upfront about why before getting into the disposition decision.

I'm going to close this PR, but I want to walk through the reasoning per-axis so the close doesn't feel like "we ignored your work and then dropped it." The feature is real and the analysis you did is on target — closing here is about timing and approach, not about your contribution being unwelcome.

## Where the PR actually stands today

I pulled the branch, rebased against current main (clean — no textual conflicts despite being 627 commits behind), and tried to compile. The rebased code does not compile against current main:

- 6 compile errors all from PyO3 0.28 / pyo3-stub-gen API drift since January. Specifically: `Str`/`Repr`/`Dir`/`Dict` derive macros are no longer in scope at the import paths you used (pyo3-stub-gen reorganized them), `PyObject` is no longer at its old import path, and `IntoPyObjectConverter::wrap` is now ambiguous because PyO3 0.28 introduced overloaded conversion paths.

These are mechanical fixes (probably 1-2 hours of API-update work), so the rebase situation alone isn't disqualifying.

## The bigger concern — the thread_local + RefCell pattern

The serde-context plumbing in the new code does this:

```rust
thread_local! {
    static CURRENT_TYPE_INFO: RefCell<Option<TypeInfo<'static>>> = RefCell::new(None);
}
```

…and then sets the cell before each service `call()` and clears it after. The motivation is real — serde's `Deserialize` trait has no per-call context channel, and you need to thread `TypeInfo` to the deserializer somehow. But the chosen workaround has three failure modes that are hard to defend:

1. **Reentrancy**: if any nested deserialization fires with a different `TypeInfo` (a response containing another typed struct, a logging path that triggers `Debug → Display` through a serde codec, etc.) the thread-local clobbers itself silently. The "right" answer is `serde::de::DeserializeSeed`, which is designed exactly for "deserialize with context."

2. **Concurrency**: PyO3-exposed objects can be moved across threads from the Python side. The current `block_on` path serializes one call per thread, but a future async refactor or any cooperative scheduling that runs two `call()`s on the same thread races on the cell. `Send`/`Sync` story for the whole `Ros2Client` is also fuzzy under this design.

3. **Brittleness around the "ignored" type names**: `TypedService::request_type_name()` returns the literal string `"ignored"`, relying on ros2-client never actually using it. Works today; one upstream change away from a confusing failure.

These aren't surface-level cleanups — they're a design pivot. A proper fix would replace the thread-local with `DeserializeSeed` and likely restructure `OwnedTypedValue` so the deserializer carries its context inline. That's maybe 2-3 hours of focused refactor work for someone with PyO3 + serde context.

## Why I'm leaning toward "close" instead of "leave open" or "rescue"

Three things converge:

1. **dora is in 1.0 RC consolidation right now.** Recent merge activity is explicitly subsystem regression fixes and hardening — `dora clean` (#1835), ros2-bridge CI fix (#1839), per-node-venv pytest fix (#1843). Adding new Python features to an **unstable** surface (ros2-bridge is explicitly marked as such — "may be changed at any point without it being considered a breaking change") is the wrong direction during a freeze.

2. **The PR doesn't close #1170.** I want to be clear about this because the body says "Fixes #1170", and that overstates the impact. #1170 is a multi-checkbox tracker — Python API alone has 4 unchecked boxes (service client, service server, action client, action server), plus 5 more unchecked items across Rust/C++/others. This PR ticks one of those nine boxes. Valuable, but not "fixes" — #1170 stays open afterward.

3. **A fresh PR is faster than rescuing this one.** With a `DeserializeSeed`-based design, fresh tests, and PyO3 0.28 from the start, a clean implementation is probably *less* work than rebasing this one and replacing the thread-local pattern in place. The 4-month staleness isn't doing the rescuer favors — the code shape has drifted under it.

## What I'd suggest

If you still want this feature in dora:

- **Wait for 1.0 to ship**, then open a fresh PR. By then ros2-bridge stops being explicitly unstable and the design effort to do this carefully is justified.
- **Use `DeserializeSeed`** (or equivalent context-carrying pattern) instead of thread-local mutable state. Happy to point at examples in the codebase if helpful when you reopen.
- **Include a small integration test** that exercises a typed service round-trip through `Ros2Context`. Two months of review wait would have caught the design concern earlier with a test scaffolding the failure modes.

If you'd rather not pick this up again, that's completely fine too — opening it forced me to actually read what #1170 says about Python service clients and what a good shape looks like, and that's useful for whoever does it next.

Thank you for the effort and the patience. Reopen this branch (or open a fresh one) whenever you're ready; the timing concerns are about the project's current phase, not about the PR being unwelcome.

PR #1334, opened 2026-01-27
Fix: Check for dataflow finish after stop_all for dynamic nodes

### LINKED ISSUE #1333: Impossible to stop dataflow with only dynamic nodes
**Describe the bug**
Current process is based on the last node to stop to finish the dataflow.

We need to add a check that if all nodes are dynamic and we receive a stop signal to send stop signal and then stop the daemon.

haixuanTao:
When all nodes in a dataflow are dynamic and a stop signal is received, the daemon now properly checks if the dataflow should be marked as finished immediately after stop_all() completes.

Previously, the finish check was only triggered by handle_node_stop(), which is only called when a SpawnedNodeResult event is received. Dynamic nodes don't send this event, so dataflows with only dynamic nodes would never finish after a stop signal.

The fix adds check_dataflow_finished_after_stop() which is called after stop_all() in both the Ctrl+C handler and the coordinator stop event handler. This ensures that when all running nodes are dynamic and stop was sent, the dataflow is immediately marked as finished.

Fixes #1333

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -631,9 +631,12 @@ impl Daemon {
     }
 
     async fn trigger_manual_stop(&mut self) -> eyre::Result<()> {
+        // Collect dataflow IDs that need immediate finishing
+        let mut dataflows_to_finish = Vec::new();
+
         for dataflow in self.running.values_mut() {
             let mut logger = self.logger.for_dataflow(dataflow.id);
-            dataflow
+            let finish_when = dataflow
                 .stop_all(
                     &mut self.coordinator_connection,
                     &self.clock,
@@ -642,9 +645,18 @@ impl Daemon {
                     &mut logger,
                 )
                 .await?;
-            // After stop_all, check if dataflow should finish (for all-dynamic nodes case)
-            self.check_dataflow_finished_after_stop(dataflow_id).await?;
+
+            // If stop_all returns Now, we need to finish this dataflow
+            if matches!(finish_when, FinishDataflowWhen::Now) {
+                dataflows_to_finish.push(dataflow.id);
+            }
         }
+
+        // Finish dataflows after the loop to avoid borrow checker issues
+        for dataflow_id in dataflows_to_finish {
+            self.finish_dataflow(dataflow_id).await?;
+        }
+
         self.exit_when_all_finished = true;
         Ok(())
     }


binaries/daemon/src/lib.rs
```diff
@@ -2111,6 +2139,80 @@ impl Daemon {
...
+                            result,
+                        },
+                    },
+                    timestamp: self.clock.new_timestamp(),
+                })?;
+                socket_stream_send(connection, &msg)
+                    .await
+                    .wrap_err("failed to report dataflow finish to dora-coordinator")?;
+            }
+            self.running.remove(&dataflow_id);
```
phil-opp: This is duplicated code, copied from `handle_stop_inner`, right? Instead of duplicating this, we should probably put it in a function so that we don't need to keep two cleanup routines in sync on future changes.

binaries/daemon/src/lib.rs
```diff
@@ -453,17 +464,22 @@ impl Daemon {
...
+                                .stop_all(
+                                    &mut self.coordinator_connection,
+                                    &self.clock,
+                                    None,
+                                    false,
+                                    &mut logger,
+                                )
+                                .await?;
+                            // After stop_all, check if dataflow should finish (for all-dynamic nodes case)
+                            self.check_dataflow_finished_after_stop(dataflow_id).await?;
```
phil-opp: It would be cleaner to have `stop_all` return a new `FinishDataflowWhen` enum, which looks something like this:


```rust
#[must_use]
enum FinishDataflowWhen {
    Now,
    WaitForNodes,
}
```

If set to `Now`, we call the cleanup function directly. Else, we do nothing and wait for the `SpawnedNodeResult` events.

By marking the enum as `#[must_use]`, the compiler will issue a warning if we forget to handle the return value. This way, it will remind us to check the result even if we call `stop_all` in more cases in the future.

binaries/daemon/src/lib.rs
```diff
@@ -2111,6 +2139,80 @@ impl Daemon {
         Ok(())
     }
 
+    /// Check if a dataflow should be marked as finished after stop_all() is called.
+    /// This is needed for dataflows with only dynamic nodes, which don't send
+    /// SpawnedNodeResult events and thus never trigger the normal finish check.
+    async fn check_dataflow_finished_after_stop(&mut self, dataflow_id: Uuid) -> eyre::Result<()> {
```
phil-opp: Let's make this a method on `RunningDataflow` instead of `Daemon` that just returns `should_finish`. Then we don't need the extra lookup in `self.running` (and the `return Ok(())` early-exit).

binaries/daemon/src/lib.rs
```diff
@@ -631,6 +642,8 @@ impl Daemon {
                     &mut logger,
                 )
                 .await?;
+            // After stop_all, check if dataflow should finish (for all-dynamic nodes case)
+            self.check_dataflow_finished_after_stop(dataflow_id).await?;
```
phil-opp: This is probably a merge error.

haixuanTao: Indeed sorry!

PR #1335, opened 2026-01-28
Use new `dora run --stop-after` argument to simplify CI testing

phil-opp:
Use the new `--stop-after` argument introduced in #1308

### DIFF
--- a/libraries/core/src/lib.rs
+++ b/libraries/core/src/lib.rs
@@ -33,23 +33,24 @@ pub fn adjust_shared_library_path(path: &Path) -> Result<std::path::PathBuf, eyr
 }
 
 // Search for python binary.
-// 1. If `uv` is available, use `uv run which python`
+// 1. If `uv` is available, use `uv python find` to get the Python path
 // 2. Otherwise, try `python` and check it's not Python 2
 // 3. Fall back to `python3` if `python` is Python 2
 pub fn get_python_path() -> Result<std::path::PathBuf, eyre::ErrReport> {
     // First, try using uv if available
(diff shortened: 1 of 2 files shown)


haixuanTao:
I can have a look at the CI

haixuanTao (approved):

libraries/core/src/lib.rs
```diff
@@ -33,17 +33,60 @@ pub fn adjust_shared_library_path(path: &Path) -> Result<std::path::PathBuf, eyr
...
-            .context("failed to find `python` or `python3`. Make sure that python is available.")?
-    } else {
-        which::which("python3")
-            .context("failed to find `python` or `python3`. Make sure that python is available.")?
-    };
+    // First, try using uv if available
+    if let Ok(uv_path) = get_uv_path() {
+        let output = std::process::Command::new(&uv_path)
+            .args(["python", "find"])
+            .output();
```
phil-opp, line 44: What if `uv` is installed, but the user deliberately wants to build with `uv: false`? Wouldn't this return the uv python unconditionally, leading to two different Python interpreters being used?

haixuanTao: Yeah, so basically uv: True is more for installation and basically defines `pip` as `uv pip` which are 2 different process.

In this scenario, `uv find python` and `python` should returns the same thing: https://docs.astral.sh/uv/concepts/python-versions/#finding-a-python-executable

haixuanTao: the reason why it uses to fail is that we used to search for python3 that is not defined in uv

Issue #997, opened 2025-05-07
Enhancement Request: Unified and User-Friendly Metadata API in Rust


XiaoPengYouCode:
# Enhancement Request: Unified and User-Friendly Metadata API in Rust

## Issue Description

Currently, there's a significant discrepancy between the Python and Rust APIs for accessing metadata in dora-rs. The Python API provides a convenient dictionary-like interface for accessing metadata parameters with default values, while the Rust API requires manual handling of the `Parameter` enum variants and doesn't offer similar convenience methods.

### Current Python API Example
```python
# Convenient parameter access with default values
encoding = event["metadata"]["encoding"] # string
wait = event["metadata"].get("wait", True)  # bool
duration = float(event["metadata"].get("duration", 1))  # float or integer
...
```

### Current Rust API Usage
```rust
// Current approach requires manual type checking and unwrapping for each parameter
let encoding = if let Some(Parameter::String(val)) = metadata.parameters.get("encoding") {
    val.clone()
} else {
    "".to_string()
};

let wait = if let Some(Parameter::Bool(val)) = metadata.parameters.get("wait") {
    *val
} else {
    true
};

// And so on for each parameter...
```

## Proposed Enhancement

Add convenient accessor methods to the `Metadata` struct in Rust that provide a similar experience to the Python API. This would improve developer experience, reduce boilerplate code, and make the two APIs more consistent.

### Suggested Implementation

```rust
impl Metadata {
    // Generic getter that returns the raw Parameter
    pub fn get(&self, key: &str) -> Option<&Parameter> {
        self.parameters.get(key)
    }
    
    // Type-specific getters with default values
    pub fn get_string(&self, key: &str, default: &str) -> String {
        if let Some(Parameter::String(value)) = self.parameters.get(key) {
            value.clone()
        } else {
            default.to_string()
        }
    }
    
    pub fn get_bool(&self, key: &str, default: bool) -> bool {
        if let Some(Parameter::Bool(value)) = self.parameters.get(key) {
            *value
        } else {
            default
        }
    }
    
    pub fn get_int(&self, key: &str, default: i64) -> i64 {
        if let Some(Parameter::Integer(value)) = self.parameters.get(key) {
            *value
        } else {
            default
        }
    }
    
    pub fn get_float(&self, key: &str, default: f64) -> f64 {
        if let Some(Parameter::Float(value)) = self.parameters.get(key) {
            *value
        } else {
            default
        }
    }
    
[...]

PR #1337, opened 2026-01-30
Rust: Add TryFrom-based typed access for Metadata parameters

Harsh-Sahu43:
### Summary

This PR improves the ergonomics of metadata access in the Rust Node API by introducing a
`TryFrom`-based conversion model and a generic `Metadata::try_get<T>()` helper.

It removes the need for manual pattern matching on `Parameter` variants while preserving
explicit error handling and avoiding silent default values, in line with Rust idioms and
maintainer feedback.

This addresses #997.
---

### Changes

- Added `TryFrom<&Parameter>` implementations for supported scalar and list types
- Added a generic `Metadata::try_get<T>() -> Result<Option<T>, MetadataError>` helper
- Kept `Metadata::get(&str) -> Option<&Parameter>` minimal and low-level
- Refactored `open_telemetry_context` to use the new API as an internal example
- Added comprehensive unit tests covering conversions and error cases

---

### Design Notes

- Missing keys return `Ok(None)` instead of implicit defaults
- Type mismatches are reported via `MetadataError`
- Defaults, if needed, are handled explicitly by callers
- No changes were made to metadata transport, serialization, or wire formats

---

### Example Usage

```rust
let wait = metadata
    .try_get::<bool>("wait")?
    .unwrap_or(true);

let encoding = metadata
    .try_get::<String>("encoding")?
    .unwrap_or_else(|| "utf-8".to_string());

### DIFF
--- a/libraries/message/src/metadata.rs
+++ b/libraries/message/src/metadata.rs
@@ -37,20 +37,18 @@ impl Metadata {
         self.timestamp
     }
 
+    /// Returns a raw metadata parameter by key.
+    ///
+    /// This method only accesses user-provided metadata parameters.
+    /// It does **not** return fields such as the timestamp or type information.
+    /// For example, `get("timestamp")` is not equivalent to [`Metadata::timestamp`].
     pub fn get(&self, key: &str) -> Option<&Parameter> {
         self.parameters.get(key)
     }
 
-    pub fn try_get<T>(&self, key: &str) -> Result<Option<T>, MetadataError>
-    where
-        T: for<'a> TryFrom<&'a Parameter, Error = MetadataError>,
-    {
-        self.parameters.get(key).map(T::try_from).transpose()
-    }
     pub fn open_telemetry_context(&self) -> String {
-        self.try_get::<String>("open_telemetry_context")
-            .ok()
-            .flatten()
+        self.get("open_telemetry_context")
+            .and_then(|p| String::try_from(p).ok())
             .unwrap_or_default()
     }
 }
@@ -70,26 +68,18 @@ pub struct ArrowTypeInfo {
 }
 
 #[derive(Debug, Clone)]
-pub enum MetadataError {
-    MissingKey,
-    TypeMismatch {
-        expected: &'static str,
-        found: &'static str,
-    },
+pub struct TryFromParameterError {
+    pub expected: &'static str,
+    pub found: &'static str,
 }
 
-impl fmt::Display for MetadataError {
+impl fmt::Display for TryFromParameterError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            MetadataError::MissingKey => write!(f, "metadata key not found"),
-            MetadataError::TypeMismatch { expected, found } => {
-                write!(f, "expected {}, found {}", expected, found)
-            }
-        }
+        write!(f, "expected {}, found {}", self.expected, self.found)
     }
 }
 
-impl std::error::Error for MetadataError {}
+impl std::error::Error for TryFromParameterError {}
 /// A metadata parameter that can be sent as part of output messages.
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 pub enum Parameter {
@@ -117,11 +107,12 @@ impl Parameter {
 }
 
 impl TryFrom<&Parameter> for bool {
-    type Error = MetadataError;
+    type Error = TryFromParameterError;
+
     fn try_from(value: &Parameter) -> Result<Self, Self::Error> {
         match value {
             Parameter::Bool(value) => Ok(*value),
-            other => Err(MetadataError::TypeMismatch {
+            other => Err(TryFromParameterError {
                 expected: "bool",
                 found: other.variant_name(),
             }),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: libraries/message/src/metadata.rs [hunk 1]
```diff
@@ -36,12 +37,21 @@ impl Metadata {
...
+    pub fn get(&self, key: &str) -> Option<&Parameter> {
+        self.parameters.get(key)
+    }
+
+    pub fn try_get<T>(&self, key: &str) -> Result<Option<T>, MetadataError>
+    where
+        T: for<'a> TryFrom<&'a Parameter, Error = MetadataError>,
+    {
+        self.parameters.get(key).map(T::try_from).transpose()
+    }
```
phil-opp: I don't think that we need this method:

- it's not that much shorter than writing e.g. `.get(...).map(String::try_from)`
- the `try_from` name is confusing, as it hints that `get` might fail and panic

So I would prefer to remove this.

libraries/message/src/metadata.rs
```diff
@@ -36,12 +37,21 @@ impl Metadata {
         self.timestamp
     }
 
+    pub fn get(&self, key: &str) -> Option<&Parameter> {
```
phil-opp, line 45: Could you add some method documentation that this accesses `parameters` field? Just to make clear that `get("timestamp")` is not the same as `timestamp()`

libraries/message/src/metadata.rs
```diff
@@ -59,6 +69,27 @@ pub struct ArrowTypeInfo {
     pub child_data: Vec<ArrowTypeInfo>,
 }
 
+#[derive(Debug, Clone)]
+pub enum MetadataError {
```
phil-opp: Could you rename this to `TryFromParameterError`?

libraries/message/src/metadata.rs
```diff
@@ -59,6 +69,27 @@ pub struct ArrowTypeInfo {
...
+    MissingKey,
```
phil-opp: This variant does not seem to be needed. Please remove it and change the `enum` to a `struct`.

libraries/message/src/metadata.rs
```diff
@@ -71,8 +102,270 @@ pub enum Parameter {
...
+            Parameter::Bool(value) => Ok(*value),
+            other => Err(MetadataError::TypeMismatch {
+                expected: "bool",
+                found: other.variant_name(),
+            }),
+        }
+    }
+}
+
+impl TryFrom<&Parameter> for String {
```
phil-opp, line 123: We could also add a `TryFrom` implementation for `&str`.

libraries/message/src/metadata.rs
```diff
@@ -71,8 +102,270 @@ pub enum Parameter {
...
+        }
+    }
+}
+
+impl TryFrom<&Parameter> for i64 {
+    type Error = MetadataError;
+
+    fn try_from(value: &Parameter) -> Result<Self, Self::Error> {
+        match value {
+            Parameter::Integer(v) => Ok(v.clone()),
```
phil-opp: please fix the clippy warning (unneeded clone)

libraries/message/src/metadata.rs
```diff
@@ -71,8 +102,270 @@ pub enum Parameter {
...
+        }
+    }
+}
+
+impl TryFrom<&Parameter> for f64 {
+    type Error = MetadataError;
+
+    fn try_from(value: &Parameter) -> Result<Self, Self::Error> {
+        match value {
+            Parameter::Float(val) => Ok(val.clone()),
```
phil-opp: also an unneeded clone

libraries/message/src/metadata.rs
```diff
@@ -71,8 +102,270 @@ pub enum Parameter {
...
+            other => Err(MetadataError::TypeMismatch {
+                expected: "f64",
+                found: other.variant_name(),
+            }),
+        }
+    }
+}
+
+impl TryFrom<&Parameter> for Vec<i64> {
```
phil-opp, line 179: We could also add a `TryFrom` impl for `&[i64]`.

(also for `&[f64]` and `&[String]`)

phil-opp (commented):
Thanks!

libraries/message/src/metadata.rs
Harsh-Sahu43: Thanks for the suggestion!

I was wondering about the best way to implement this without changing the `Parameter` enum itself. Since `Parameter` is an owned, serializable type, adding borrowed variants (like `&[i64]`) would introduce lifetimes and likely break `Serialize` / `Deserialize`.

Would the intended approach be to keep `Parameter` unchanged and add `TryFrom<&Parameter>` implementations that return borrowed slices (e.g. `&[i64]`, `&[f64]`, `&[String]`)?

Harsh-Sahu43: any suggestion ?.

phil-opp: > Would the intended approach be to keep `Parameter` unchanged and add `TryFrom<&Parameter>` implementations that return borrowed slices (e.g. `&[i64]`, `&[f64]`, `&[String]`)?

Yes, that's what I meant (also for `&str`).

These impls would allow you to avoid the clone if you only need to read the values.

Harsh-Sahu43:
Please review this PR. @phil-opp

Harsh-Sahu43:
Hi @haixuanTao 👋  

This PR should be ready for merge now.  
All CI checks are passing, feedback from @phil-opp has been addressed, and the changes are scoped to the Rust metadata API only (no wire-format or behavior changes).

Whenever you have a moment, could you please take a look?  
Thanks!

PR #1338, opened 2026-02-01
feat: add global environment variables support for dataflow

### LINKED ISSUE #1330: Global Environment Variables Support
### Summary
I wanna define global environment variables at the top level of `dataflow.yml` that are automatically inherited by all nodes, without needing to repeat the same variables in each node's `env` section.

  ### Problem Statement
  Currently, environment variables must be duplicated in every node that needs them:

  ```yaml
  nodes:
    - id: node-a
      env:
        HAND_JOINT: "X1"
        CONTROL_RATE: "30"
        WEB_PORT: "8000"

    - id: node-b
      env:
        HAND_JOINT: "X1"
        CONTROL_RATE: "30"
        WEB_PORT: "8000"

    - id: node-c
      env:
        HAND_JOINT: "X1"
        CONTROL_RATE: "30"
        WEB_PORT: "8000"

  This creates:
  - Redundancy: Same variables repeated across many nodes
  - Maintenance burden: Changing a value requires updating multiple places
  - Error-prone: Easy to miss updating a node and have inconsistent configs

  Proposed Solution

  Support global environment variables at the dataflow level:

  # Global environment (inherited by all nodes)
  env:
    HAND_JOINT: "X1"
    CONTROL_RATE: "30"
    WEB_PORT: "8000"

  nodes:
    - id: node-a
      # Inherits HAND_JOINT, CONTROL_RATE, WEB_PORT from global
      env:
        WEB_PORT: "9000"  # Local override if needed

    - id: node-b
      # Inherits all global env, no local env needed

  Alternative Solutions Considered

  1. Template generators: Requires separate tooling, adds complexity
  2. .env file support: Would need Dora to load dotenv, adds dependency
  3. YAML anchors/aliases: Works within single file but still requires repetition in each node

  Benefits

  - DRY principle: Define once, use everywhere
  - Easier maintenance: Change global config in one place
  - Better DX: Less verbose config files

  Implementation Ideas

  - Read env key at dataflow root level
  - Merge with node-level env (node-level overrides global)
  - Maintain backward compatibility (dataflows without global env work as-is)

DarkBird10020:
# Add Global Environment Variables Support

Closes #1330

## Summary

Adds support for defining global environment variables at the top level of `dataflow.yml` that are automatically inherited by all nodes. Node-level `env` overrides global values.

## Changes

- **libraries/message/src/descriptor.rs**: Added `env` field to `Descriptor` struct at the dataflow root level
- **libraries/core/src/descriptor/mod.rs**: Added `merge_env()` helper and merge logic in `resolve_aliases_and_set_defaults()` (global + node, node overrides)
- **binaries/daemon/src/lib.rs**: Build descriptor with merged env before passing to spawner so dynamic nodes receive merged values
- **libraries/core/dora-schema.json**: Added `env` property to schema for IDE/tooling support

## Usage

```yaml
env:
  HAND_JOINT: "X1"
  CONTROL_RATE: "30"
  WEB_PORT: "8000"

nodes:
  - id: node-a
    env:
      WEB_PORT: "9000"

  - id: node-b
```

- `node-a` gets `HAND_JOINT`, `CONTROL_RATE`, and `WEB_PORT: "9000"`
- `node-b` gets all global env vars

## Backward Compatibility

Existing dataflows without a top-level `env` continue to work as before. The field is optional.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1319,10 +1319,20 @@ impl Daemon {
             }
         }
 
+        let dataflow_descriptor_with_env = {
+            let mut desc = dataflow_descriptor.clone();
+            for node in desc.nodes.iter_mut() {
+                if let Some(resolved_node) = nodes.get(&node.id) {
+                    node.env = resolved_node.env.clone();
+                }
+            }
+            desc
+        };
+
         let spawner = Spawner {
             dataflow_id,
             daemon_tx: self.events_tx.clone(),
-            dataflow_descriptor,
+            dataflow_descriptor: dataflow_descriptor_with_env,
             clock: self.clock.clone(),
             uv,
         };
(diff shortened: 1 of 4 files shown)


binaries/daemon/src/lib.rs
```diff
@@ -1319,10 +1319,20 @@ impl Daemon {
...
 
+        let dataflow_descriptor_with_env = {
+            let mut desc = dataflow_descriptor.clone();
+            for node in desc.nodes.iter_mut() {
+                if let Some(resolved_node) = nodes.get(&node.id) {
+                    node.env = resolved_node.env.clone();
+                }
+            }
+            desc
+        };
```
haixuanTao, line 1330: What is this about?

PR #1341, opened 2026-02-04
feat: Implements a self-inherited global env on .yml file

sauhardh:
Solves a #1330

Implementation: 

> At the top level of dataflow.yml file, a `env` variables can be set. (Optional)
> Environments of this variables are auto-inherited by all the nodes
> Priority is still set to the Node-level env (So that, if node-level env repalaces global env if they are same. 

Uses:

```yml
env:
  HAND_JOINT: "X1"
  CONTROL_RATE: "30"
  WEB_PORT: "8000"
nodes:
  - id: node-a
    path: python3
    args: pytest.py node-a
    env:
      WEB_PORT: "9000" # Local override
  - id: node-b
    path: python3
    args: pytest.py node-b
```

Variables like `HAND_JOINT`, `CONTROL_RATE`, `WEB_PORT` is inherited by nodes. And `WEB_PORT` is overridden by node-level env.

### DIFF
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -104,15 +104,12 @@ impl DescriptorExt for Descriptor {
                 }),
             };
 
-            let mut all_env = self.env.clone().unwrap_or_default();
-            if let Some(node_env) = node.env {
-                all_env.extend(node_env);
-            }
-
-            let env = if all_env.is_empty() {
-                None
-            } else {
-                Some(all_env)
+            let env = match (self.env.clone(), node.env) {
+                (None, node_env) => node_env,
+                (Some(mut self_env), node_env) => {
+                    self_env.extend(node_env.unwrap_or_default());
+                    Some(self_env)
+                }
             };
 
             resolved.insert(


libraries/core/src/descriptor/mod.rs
```diff
@@ -104,13 +104,24 @@ impl DescriptorExt for Descriptor {
...
+            if let Some(node_env) = node.env {
+                all_env.extend(node_env);
+            }
+
+            let env = if all_env.is_empty() {
+                None
+            } else {
+                Some(all_env)
+            };
+
```
phil-opp, line 114: This changes `Some(empty)` maps to `None`. We should preserve this:

```suggestion
            let env = match (self.env.clone(), node.env) {
                (None, node_env) => node_env,
                (Some(mut self_env), node_env) => {
                    self_env.extend(node_env.unwrap_or_default());
                    Some(self_env)
                }
            };
```

sauhardh:
@phil-opp  Thank you, I have implemented your suggestions.

 BUT 1 check is failing, I am not sure if it's related to this PR.

PR #1345, opened 2026-02-10
Start adding `tarpc` service and server to coordinator

phil-opp:
Add shared `CoordinatorState` struct for keeping state. Use `DashMap` for concurrent maps.

Not all messages use the proper tarpc request handling yet. This will be fixed in a follow-up commit. Same goes for the manual (de)serialization.

Alternative to #1288 

### Message Format

- The message format is defined by the new `CliControl` trait in the `message` crate (`cli_to_coordinator`)
- The `ControlRequestReply` enum is no longer needed, so it is removed
- Most of the `ControlRequest` variants are no longer needed as they are part of the `CliControl` trait now. I removed these variants and renamed the enum to `LegacyControlRequest`
- Given that the message format is changed, this PR will require a minor version bump of the `dora-message` crate

### Coordinator

- The coordinator now provides a tarpc server at port DORA_COORDINATOR_PORT_RPC_DEFAULT
  - Most messages are handled by this server
- The existing control server still exists at port DORA_COORDINATOR_PORT_CONTROL_DEFAULT
  - It's only used for `LogSubscribe` and `BuildLogSubscribe` messages, which are both streaming-based and cannot be represented well in tarpc
  - IMO, the proper solution to this would be using zenoh for these subscriptions. This way, we would no longer need to involve the coordinater
- Instead of handling all requests in a single event loop, the coordinator now spawns a tokio task for each request and handles them in parallel
  - state is managed in a `CoordinatorState` struct, which uses `DashMap` fields that provide safe concurrent access
- The `impl CliControl for ControlServer` uses the same code for the different requests as the previous event loop (with minor modifications)

### CLI

- The CLI's main function and all execute calls are now `async`
- Instead of manual serialization/deserialization and message type specifications, we can now simply call a function on the `CliControlClient`
- For (build) log subscriptions, the `LegacyControlRequest` is used

---

**Note:** I used AI to generate some of the code, but I reviewed all generated code manually. I also improved and simplified the generated code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2000,7 +2000,7 @@ dependencies = [
  "enum-ordinalize",
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.116",
 ]
 
 [[package]]
@@ -2032,7 +2032,7 @@ checksum = "8ca9601fb2d62598ee17836250842873a413586e5d7ed88b356e38ddbb0ec631"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.116",
 ]
 
 [[package]]
@@ -6273,7 +6273,7 @@ checksum = "26ef4401b013b1f5218ba33ea8f1eddbfcc00ec8db073ef995c192e71f08f027"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.114",
+ "syn 2.0.116",
 ]
 
 [[package]]


libraries/message/src/coordinator_to_daemon.rs
```diff
@@ -97,3 +97,8 @@ pub struct SpawnDataflowNodes {
     pub uv: bool,
     pub write_events_to: Option<PathBuf>,
 }
+
+#[tarpc::service]
+pub trait CoordinatorToDaemon {
+    async fn build(request: BuildDataflowNodes) -> Result<(), String>;
+}
```
phil-opp: Remove

libraries/message/src/cli_to_coordinator.rs
```diff
@@ -2,49 +2,60 @@ use std::{collections::BTreeMap, path::PathBuf, time::Duration};
...
+    ///
+    /// Note that nodes with git sources still use a subdirectory of
+    /// the base working dir.
+    pub local_working_dir: Option<PathBuf>,
+    pub uv: bool,
+    pub write_events_to: Option<PathBuf>,
+}
+
 #[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
 pub enum ControlRequest {
```
phil-opp: Still needed?

binaries/cli/src/common.rs
```diff
@@ -146,22 +181,22 @@ pub(crate) fn local_working_dir(
...
+
+    // Determine the CLI's outgoing IP toward the coordinator.
+    // Uses a UDP socket to query the OS routing table (no packets sent).
+    let cli_ip = std::net::UdpSocket::bind("0.0.0.0:0")
+        .and_then(|s| {
+            s.connect((coordinator_addr, 1))?;
+            s.local_addr()
+        })
+        .ok()
+        .map(|a| a.ip());
```
phil-opp: check

binaries/coordinator/src/control.rs
```diff
@@ -128,19 +128,12 @@ async fn handle_requests(
...
-            }
-        }
-
-        let reply = result.unwrap_or_else(|err| ControlRequestReply::Error(format!("{err:?}")));
+        // All other control requests are handled via the tarpc service.
+        // If a non-subscribe request arrives on the legacy TCP channel,
+        // respond with an error.
+        let reply = ControlRequestReply::Error(
+            "this control request must be sent via the tarpc service".to_string(),
+        );
```
phil-opp: remove other request types from enum?

binaries/coordinator/src/lib.rs
```diff
@@ -82,8 +95,59 @@ pub async fn start(
...
+        daemon_events_tx,
+        abort_handle,
+    });
+
+    let tarpc_state = coordinator_state.clone();
+    let mut stream = listener
+        // ignore connect errors
+        .filter_map(|c| future::ready(c.ok()))
+        .map(BaseChannel::with_defaults)
+        // limit to 10 channels per IP
```
phil-opp: remove outdated comment

examples/multiple-daemons/run.rs
```diff
@@ -50,18 +45,28 @@ async fn main() -> eyre::Result<()> {
...
-    let daemon_a = run_daemon(coordinator_addr.to_string(), "A");
-    let daemon_b = run_daemon(coordinator_addr.to_string(), "B");
+
+    // Connect a tarpc client to the coordinator's RPC service.
+    let transport = tarpc::serde_transport::tcp::connect(
+        (coordinator_addr, rpc_port),
+        tokio_serde::formats::Json::default,
+    )
+    .await
+    .context("failed to connect tarpc client to coordinator")?;
```
phil-opp: Use mpsc-based transport instead of TCP

binaries/coordinator/src/server.rs
```diff
@@ -0,0 +1,381 @@
...
+            Err(err) => Err(err_to_string(err)),
+        }
+    }
+
+    async fn wait_for_build(self, _context: Context, build_id: BuildId) -> Result<(), String> {
+        let (tx, rx) = oneshot::channel();
+        if let Some(mut build) = self.state.running_builds.get_mut(&build_id) {
+            build.build_result.register(tx);
+        } else if let Some(mut result) = self.state.finished_builds.get_mut(&build_id) {
+            result.register(tx);
```
phil-opp, line 99: Make register take a channel that returns the result directly, instead of wrapping into ControlRequestReply

binaries/coordinator/src/server.rs
```diff
@@ -0,0 +1,381 @@
...
+    async fn wait_for_spawn(self, _context: Context, dataflow_id: Uuid) -> Result<(), String> {
+        let (tx, rx) = oneshot::channel();
+        if let Some(mut dataflow) = self.state.running_dataflows.get_mut(&dataflow_id) {
+            dataflow.spawn_result.register(tx);
+        } else {
+            return Err(format!("unknown dataflow {dataflow_id}"));
+        }
+
+        match rx.await {
+            Ok(Ok(_)) => Ok(()),
```
phil-opp: Return Ok(())

binaries/coordinator/src/server.rs
```diff
@@ -0,0 +1,381 @@
...
+        .await
+        .map_err(err_to_string)?;
+        Ok(dataflow_id)
+    }
+
+    async fn check(
+        self,
+        _context: Context,
+        dataflow_uuid: Uuid,
+    ) -> Result<ControlRequestReply, String> {
```
phil-opp: Custom reply enum instead of ControlRequestReply

binaries/coordinator/src/server.rs
```diff
@@ -0,0 +1,381 @@
...
+    format!("{err:?}")
+}
+
+/// Shared logic for stopping a dataflow by UUID.
+async fn stop_dataflow_impl(
+    state: &CoordinatorState,
+    dataflow_uuid: Uuid,
+    grace_duration: Option<Duration>,
+    force: bool,
+) -> Result<ControlRequestReply, String> {
```
phil-opp: Custom enum instead

binaries/cli/src/common.rs
```diff
@@ -1,23 +1,55 @@
...
+/// Run a future to completion, reusing the persistent CLI runtime.
+pub(crate) fn block_on<F: Future>(f: F) -> F::Output {
+    let rt = RUNTIME.get_or_init(|| {
+        Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .expect("failed to create tokio runtime")
+    });
+    rt.block_on(f)
+}
```
phil-opp: Is this actually needed? We could also make the calling functions async and await instead.

phil-opp:
@claude Sorry, both of your commits are hacky and don't fix the underlying issue. Please revert a4d3ec1 and 2a0c2fa.

haixuanTao (commented):
Thanks for the solid refactor! The DashMap lock discipline is well done and the insert-before-spawn pattern is excellent. I tested the full CLI lifecycle (up/check/list/start/stop/logs/build/destroy) on macOS and everything works correctly — including log streaming via both `dora start` and `dora run`.

Left a few comments on things that caught my eye.

binaries/cli/src/common.rs
```diff
@@ -91,46 +107,49 @@ pub(crate) struct CoordinatorOptions {
...
-pub(crate) fn connect_to_coordinator(
-    coordinator_addr: SocketAddr,
-) -> std::io::Result<Box<TcpRequestReplyConnection>> {
-    TcpLayer::new().connect(coordinator_addr)
+/// Connect to the coordinator's tarpc RPC service.
+pub(crate) async fn connect_to_coordinator_rpc(
+    addr: IpAddr,
+    _control_port: u16,
+) -> eyre::Result<CliControlClient> {
+    let rpc_port = DORA_COORDINATOR_PORT_RPC_DEFAULT;
```
haixuanTao: The `_control_port` parameter is accepted but silently ignored — the RPC port is always hardcoded to `DORA_COORDINATOR_PORT_RPC_DEFAULT` (6013). This means any user passing `--coordinator-port` will have it silently discarded for all tarpc operations.

Suggestion — either derive the RPC port from the control port:

```rust
pub(crate) async fn connect_to_coordinator_rpc(
    addr: IpAddr,
    control_port: u16,
) -> eyre::Result<CliControlClient> {
    // RPC port is control port + 1 by convention
    let rpc_port = if control_port == DORA_COORDINATOR_PORT_CONTROL_DEFAULT {
        DORA_COORDINATOR_PORT_RPC_DEFAULT
    } else {
        control_port + 1
    };
    // ...
}
```

Or add a dedicated `--coordinator-rpc-port` CLI argument to `CoordinatorOptions` so users can configure both ports independently.

binaries/coordinator/src/lib.rs
```diff
@@ -948,18 +702,21 @@ async fn start_inner(
                     };
                     if build.pending_build_results.is_empty() {
                         tracing::info!("dataflow build finished: `{build_id}`");
-                        let mut build = running_builds.remove(&build_id).unwrap();
+                        let (build_id, mut build) =
+                            coordinator_state.running_builds.remove(&build_id).unwrap();
```
haixuanTao: This `.unwrap()` can panic and crash the entire coordinator. While the preceding `get_mut()` succeeded, the `RefMut` guard was dropped when we entered this `if` block, so another concurrent path could theoretically remove the entry in between (the tarpc server spawns handlers concurrently).

Even if that's unlikely today, a panic here takes down all running dataflows. Suggestion:

```rust
let Some((build_id, mut build)) =
    coordinator_state.running_builds.remove(&build_id)
else {
    tracing::warn!("build {build_id} was removed while finalizing results");
    continue;
};
```

binaries/cli/src/command/list.rs
```diff
@@ -42,20 +40,14 @@ pub struct ListArgs {
...
-        list(
-            &mut *session,
-            self.format,
-            self.status,
-            self.name,
-            self.sort_by,
-        )
+        let client = connect_to_coordinator_rpc(self.coordinator_addr, self.coordinator_port)
+            .await
+            .map_err(|_| eyre!("Failed to connect to coordinator"))?;
```
haixuanTao: `.map_err(|_| ...)` discards the underlying connection error, making it hard to debug why a connection failed (refused? timeout? DNS?). Other commands like `stop.rs` and `logs.rs` use `.wrap_err()` which preserves the cause chain.

Suggestion:
```rust
let client = connect_to_coordinator_rpc(self.coordinator_addr, self.coordinator_port)
    .await
    .wrap_err("failed to connect to dora coordinator")?;
```

binaries/cli/src/common.rs
```diff
@@ -1,23 +1,48 @@
...
+///
+/// 1. Transport-level error (e.g. `RpcError`)
+/// 2. `Result<T, String>` from the application (coordinator) layer
+pub(crate) async fn rpc<T, E: std::error::Error + Send + Sync + 'static>(
+    future: impl Future<Output = Result<Result<T, String>, E>>,
+) -> eyre::Result<T> {
+    future
+        .await
+        .context("RPC transport error")?
+        .map_err(|e| eyre::eyre!(e))
```
haixuanTao: Two issues with error reporting here:

1. **Transport errors lose detail**: `.context("RPC transport error")` wraps the error but the message is generic — when the coordinator is down you get "RPC transport error" without knowing if it was connection refused, timeout, or network unreachable.

2. **Application errors lack operation context**: `.map_err(|e| eyre::eyre\!(e))` converts the coordinator's error String into an eyre Report but doesn't say *which* operation failed. Callers that don't add `.wrap_err()` produce unhelpful messages.

Some callers do add context (e.g. `stop.rs` uses `.wrap_err("...")`), but others don't (e.g. `list.rs:81` — `rpc(client.get_node_info(...))`).

One option is to accept an operation label:
```rust
pub(crate) async fn rpc<T, E: std::error::Error + Send + Sync + 'static>(
    future: impl Future<Output = Result<Result<T, String>, E>>,
    operation: &str,
) -> eyre::Result<T> {
    future
        .await
        .wrap_err_with(|| format\!("RPC transport error during {operation}"))?
        .map_err(|e| eyre::eyre\!("{operation} failed: {e}"))
}
```

Alternatively, just ensure every call site wraps with `.wrap_err()`.

binaries/cli/src/common.rs
phil-opp: Fixed in 870d43f5

binaries/coordinator/src/lib.rs
phil-opp: Fixed in badef0f3

binaries/cli/src/command/list.rs
phil-opp: Fixed in 0615ea0e

binaries/cli/src/common.rs
phil-opp: Addressed in 1715b672

haixuanTao:
Sure!

haixuanTao (approved):

PR #1351, opened 2026-02-15
chore: clippy fixes

swar09:
New to the project here, I was running cargo clippy --all on dora and it caught some warnings that not show up with just cargo clippy.

Happy to close if not needed.


swar09:
Yes , My approach : 
  -  job fixes clippy warnings and opens a PR for each. 
  - clippy job skips itself if a PR already exists or  there are no changes to commit.
  
 For now  I will close this PR.

Issue #1211, opened 2025-11-18
Implement `dora system doctor` Command


ZhangHanDong:
**Goal**: Automatically diagnose environment and configuration issues

**Functional Requirements**:

- Check necessary toolchains (Rust/Python/C++)
- Check dependency library versions
- Check network port availability
- Check shared memory permissions
- Check configuration file integrity
- Provide fix suggestions for each issue
- Optional: `--fix` for automatic fixes (implement cautiously)

**Technical Points**:

- System command invocation (`rustc --version`, `python3 --version`)
- Port scanning (detecting conflicts)
- File permission checks
- Generate diagnostic report

**Acceptance Criteria**:

```bash
$ dora system doctor
✓ Rust toolchain: 1.75.0
✗ Python 3.x: Not found
  → Install Python 3.8+: https://...
✓ Network port 53290: Available
⚠ Shared memory: Permission issue detected
  → Run: sudo chmod 1777 /dev/shm
```

**Compatibility**: New command, does not affect existing functionality

Harsh-Sahu43:
The current scope of `dora system doctor` seems to mix
runtime diagnostics with project/language assumptions.

Would it make sense to start with a runtime-focused doctor
that only checks Dora’s own prerequisites
(e.g. coordinator/daemon availability, version compatibility, basic spawn sanity),
and expand later once the foundations are clear?

If that direction sounds reasonable, I’m happy to prototype a minimal version.

Harsh-Sahu43:
Thanks for the clarification.

Given that /dev/shm permission issues don’t seem to be a common real-world problem, I agree that introducing a standalone system doctor command just for this check may not be justified.

Looking back at #1211, some proposed checks (e.g. toolchains, dependency versions) would require knowing which languages are actually used, likely via a dataflow.yml. Without that context, the command wouldn’t be able to determine what to validate.

@ZhangHanDong Could you clarify the intended scope of system doctor?

@phil-opp From your perspective, what would meaningfully justify a system doctor command? Should it focus only on Dora’s own prerequisites, or broader environment validation?

If the scope isn’t clearly defined yet, I’m also fine pausing or reworking this once the direction is clearer.

heyong4725:
Closing — shipped under existing name `dora doctor`.

`dora doctor` is implemented at `binaries/cli/src/command/doctor.rs` and provides the diagnostics functionality requested. The proposed rename to `dora system doctor` was declined in the umbrella thread (#1196).

PR #1352, opened 2026-02-17
feat(cli) : add system doctor command with shared memory feature

Harsh-Sahu43:
## Summary

This PR introduces the initial implementation of the `dora system doctor` command, focusing on shared memory validation for Linux systems.

Dora relies on `/dev/shm` for inter-process communication. Incorrect permissions (expected `1777`) can lead to runtime issues when spawning or coordinating nodes. This command validates the shared memory configuration and provides actionable fix suggestions when misconfigured.

This addresses (#1211).
---
## What This PR Implements

- Adds new CLI command: `dora system doctor`
- Linux-only shared memory validation:
  - Verifies `/dev/shm` exists
  - Ensures it is a directory
  - Validates permissions are `1777` (sticky bit enabled)
- Clear CLI output:
  - `✓` when configuration is correct
  - `⚠` with fix suggestion when misconfigured
- Returns non-zero exit code if an issue is detected
- Gracefully skips the check on non-Linux platforms

## Scope & Rationale

The original issue (#1211) proposes multiple diagnostic checks (toolchains, ports, config integrity, etc.).

To keep the implementation incremental and well-scoped, this PR introduces a runtime-level diagnostic focused on shared memory only.

##  Example Output

```bash
$ dora system doctor

⚠ Shared memory: Incorrect permissions (expected 1777)
  → Run: sudo chmod 1777 /dev/shm

### DIFF
--- a/binaries/cli/src/command/system/doctor.rs
+++ b/binaries/cli/src/command/system/doctor.rs
@@ -1,0 +1,90 @@
+use crate::command::{Executable, default_tracing};
+use std::io::{IsTerminal, Write};
+use termcolor::{Color, ColorChoice, ColorSpec, WriteColor};
+
+#[derive(Debug, clap::Args)]
+pub struct Doctor {}
+
+impl Executable for Doctor {
+    fn execute(self) -> eyre::Result<()> {
+        default_tracing()?;
+        run_doctor_checks()
+    }
+}
+
+fn run_doctor_checks() -> eyre::Result<()> {
+    #[cfg(target_os = "linux")]
+    let mut issue_detected = false;
+
+    let color_choice = if std::io::stdout().is_terminal() {
+        ColorChoice::Auto
+    } else {
+        ColorChoice::Never
+    };
+    let mut stdout = termcolor::StandardStream::stdout(color_choice);
+
+    #[cfg(target_os = "linux")]
+    {
+        if let Some((message, fix_hint)) = shared_memory_issue() {
+            let _ = stdout.set_color(ColorSpec::new().set_fg(Some(Color::Yellow)));
+            write!(stdout, "⚠ ")?;
+            let _ = stdout.reset();
+            writeln!(stdout, "Shared memory: {message}")?;
+            writeln!(stdout, "  → Run: {fix_hint}")?;
+            issue_detected = true;
+        } else {
+            let _ = stdout.set_color(ColorSpec::new().set_fg(Some(Color::Green)));
+            write!(stdout, "✓ ")?;
+            let _ = stdout.reset();
+            writeln!(stdout, "Shared memory: Permissions OK")?;
+        }
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    {
+        let _ = stdout.set_color(ColorSpec::new().set_fg(Some(Color::Yellow)));
+        write!(stdout, "⚠ ")?;
+        let _ = stdout.reset();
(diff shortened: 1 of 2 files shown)


Harsh-Sahu43:
Thanks for the clarification.

Given that /dev/shm permission issues don’t seem to be a common real-world problem, I agree that introducing a standalone system doctor command just for this check may not be justified.

Looking back at https://github.com/dora-rs/dora/issues/1211, some proposed checks (e.g. toolchains, dependency versions) would require knowing which languages are actually used, likely via a dataflow.yml. Without that context, the command wouldn’t be able to determine what to validate.

@ZhangHanDong Could you clarify the intended scope of system doctor?

@phil-opp From your perspective, what would meaningfully justify a system doctor command? Should it focus only on Dora’s own prerequisites, or broader environment validation?

If the scope isn’t clearly defined yet, I’m also fine pausing or reworking this once the direction is clearer.

heyong4725:
Closing with constructive feedback — the `/dev/shm` permission check has real value, but the PR shape conflicts with the umbrella decision.

## What the PR proposes

Adds a NEW command `dora system doctor` with a `/dev/shm` permission check for Linux (verifies the directory has the expected `1777` permissions, since dora relies on it for IPC).

## Why the framing is off

Two issues:

1. **`dora doctor` already exists** at the top level (`binaries/cli/src/command/doctor.rs`), introduced in the 1.0 consolidation. Adding `dora system doctor` would be a second doctor command — confusing UX.
2. **The `dora system` namespace itself is the rename the umbrella declined**. Issue #1211 (which this PR fixes) was closed as *"shipped under existing name `dora doctor`; the rename to `dora system doctor` was declined per the umbrella thread."*

## Why the check itself is valuable

The `/dev/shm` permission validation is a genuinely useful diagnostic. dora hits real errors when `/dev/shm` is missing or has wrong permissions, and surfacing that early in `dora doctor` would save user debugging time.

## Path forward

Please open a fresh focused PR titled something like *"feat(doctor): add /dev/shm permission check on Linux"* that adds your check **into the existing `binaries/cli/src/command/doctor.rs`** rather than creating a new `dora system doctor` command. The core check logic from this PR carries over cleanly; only the location and command-naming change.

Thanks @Harsh-Sahu43 — the check itself is the right idea, just landed in the wrong place.

heyong4725:
Hey @Harsh-Sahu43 — thanks for catching the `/dev/shm` misconfiguration footgun. The core check was genuinely useful and I didn't want it to be lost when this PR was closed.

I rescued the check into the existing `dora doctor` command (instead of the `dora system doctor` namespace, which was declined as part of #1196). You're credited via `Co-Authored-By` on the commit.

Follow-up PR: #1818

The check produces:

| Scenario | Output |
|---|---|
| `/dev/shm` not accessible | `FAIL Shared memory: /dev/shm not accessible — run \`sudo mkdir -p /dev/shm && sudo chmod 1777 /dev/shm\`` |
| Not a directory | `FAIL Shared memory: /dev/shm exists but is not a directory — run \`sudo chmod 1777 /dev/shm\`` |
| Wrong mode | `FAIL Shared memory: /dev/shm has mode 0o755, expected 0o1777 — run \`sudo chmod 1777 /dev/shm\`` |
| Mode 1777 | `PASS Shared memory: /dev/shm mode is 1777` |

Linux-only via `#[cfg(target_os = "linux")]`.

PR #1357, opened 2026-02-17
feat(cli): Implement static file checks for 'dora check'

### LINKED ISSUE #1356: `dora check --dataflow` provides ambiguous "INFO" messages for missing files
When running dora check --dataflow <FILE.yml>, if the dataflow references non-existent python/rust source files, the CLI behavior is confusing and misleading.

It prints a generic system message: INFO: Could not find files for the given pattern(s). This message does not specify which file is missing or which node is affected.

Crucially, it still reports success for the Coordinator/Daemon checks (ok), leading the user to believe the dataflow itself is valid.

### **Steps to Reproduce**
Create a broken.yml with a node referencing a missing file:

YAML
```
nodes:
  - id: ghost_node
    path: missing_file.py
```
**Run the check:**

Bash
```
dora check --dataflow broken.yml
```

### **Current Behavior**
```
Dora Coordinator: ok
Dora Daemon: ok
C:\test>INFO: Could not find files for the given pattern(s).
```

Image

### **Expected Behavior**
The CLI should explicitly validate the path of each node. If a file is missing, it should:
Print a clear [ERROR] message identifying the specific node ID and the missing path.
Exit with a non-zero status code to indicate failure.
(Optional) List all missing files in one go rather than stopping at the first one.

### **Environment**
OS: Windows (confirmed), likely Linux/macOS too.
Dora Version: v0.3...

kush124k:
## Description
This PR upgrades the `dora check` command to perform static analysis on the provided dataflow file. 

Previously, `dora check --dataflow <FILE.yml>` would validate the backend status but ignore missing source files in the graph, leading to false positives (reporting "OK" even when scripts were missing).

## Changes
- Parsed the dataflow descriptor within the `check` command.
- Added a validation loop that iterates through all nodes.
- Checks if the `source` path for each custom node exists on the disk.
- Aggregates errors: It lists *all* missing files found (instead of stopping at the first one) before exiting with a non-zero status code.

image

Screenshot: The updated command now correctly validates paths, lists all missing files, and exits with an error code.

## Related Issue
Closes #1356

## Testing
- Tested with a valid dataflow (Exit code 0).
- Tested with a dataflow containing multiple missing files (Exit code 1, correctly lists all missing paths).


phil-opp:
I think I found it. Could you check https://github.com/dora-rs/dora/pull/1358 and see whether this fixes the issue?

PR #1359, opened 2026-02-17
Collect and report all errors in `dora check`

phil-opp:
Instead of failing at the first error

Suggested in #1356


phil-opp:
Ok great, thanks for checking!

PR #1362, opened 2026-02-20
docs: add reproducible integration testing workflow

SuyashParmar:
## Summary

This PR improves Dora’s node integration-testing documentation by adding a deterministic, end-to-end workflow that contributors can follow to create and maintain regression fixtures.

## What changed

- Added a new **Reproducible Regression Workflow** section in:
  - `apis/rust/node/src/integration_testing.rs`
- Expanded fixture documentation in:
  - `tests/sample-inputs/README.md`

The new docs provide step-by-step commands to:
1. Record real input events from a dataflow run using `DORA_WRITE_EVENTS_TO`.
2. Re-run individual nodes in integration test mode with `DORA_TEST_WITH_INPUTS`.
3. Generate deterministic outputs using `DORA_TEST_NO_OUTPUT_TIME_OFFSET=1`.
4. Write explicit output files using `DORA_TEST_WRITE_OUTPUTS_TO`.
5. Compare generated outputs against checked-in expected JSONL fixtures.

## Why

Contributors currently have the building blocks, but not a single clear workflow for fixture generation and regression validation. This PR makes that path explicit and repeatable, which improves onboarding and reduces friction for adding integration tests.

## Testing

- Docs-only changes.


heyong4725:
Hi @SuyashParmar — thanks for putting in the time on this, and for the cross-cutting effort (docs + examples + CI + daemon). Going to close this PR as no longer applicable, but I want to walk through *why* per-hunk so it doesn't feel arbitrary.

The PR has accumulated some scope drift since it was opened (Feb 20) — the description says "Docs-only changes" but the actual diff now includes substantive daemon logic, CI hardening, and Python example rewrites. That happens, especially on `codex/...` branches. The consequence is that each piece needs to be evaluated against current main on its own merits, and each one has either been overtaken by other work or has a problem.

**1. Docs additions (`apis/rust/node/src/integration_testing.rs`, `tests/sample-inputs/README.md`) — redundant now**

A comprehensive 427-line [`guide/src/development/testing.md`](https://github.com/dora-rs/dora/blob/main/guide/src/development/testing.md) landed after this PR was opened, covering every workflow step:

- Fixture generation via `DORA_WRITE_EVENTS_TO`
- Replay via `DORA_TEST_WITH_INPUTS`
- Deterministic output via `DORA_TEST_NO_OUTPUT_TIME_OFFSET=1`
- Explicit output files via `DORA_TEST_WRITE_OUTPUTS_TO`
- Baseline regeneration when behavior intentionally changes
- All three testing modes (Rust unit, env-var, `init_testing`)
- Troubleshooting for the most common output-mismatch failures

The rustdoc additions in `integration_testing.rs` would now duplicate that more-discoverable content. The `tests/sample-inputs/README.md` expansion has some overlap; if you wanted, you could open a focused PR that just enhances *that* file with anything not yet in the guide.

**2. CI hardening (`.github/workflows/ci.yml`) — moot**

The `docker system prune --all -f` lines your PR wraps with `$LASTEXITCODE` checks don't exist anywhere in `.github/workflows/` on current main — that workflow was rewritten. Nothing to harden.

**3. Python example STOP handling — main has moved on**

Current main already handles STOP cleanly in these examples but via a different async pattern. For example, `examples/python-async/receive_data.py` on current main:

```python
async def main():
    node = Node()
    loop = asyncio.get_event_loop()
    while True:
        event = await loop.run_in_executor(None, node.next)
        if event is None or event["type"] == "STOP":
            break
        del event
```

This uses `loop.run_in_executor(None, node.next)`, while your PR uses `node.recv_async()`. The two would conflict. Also, STOP is the canonical termination signal in the current daemon protocol — the `INPUT_CLOSED` / `ALL_INPUTS_CLOSED` checks your PR adds aren't strictly needed because STOP fires for both natural completion and explicit shutdown.

If you reproduce a *specific* python example that doesn't shut down cleanly on current main, please open a fresh issue with the repro and we can target that case directly.

**4. Daemon change (`binaries/daemon/src/lib.rs`) — would defeat existing intentional behavior**

This is the most important one. The hunk reclassifies SIGTERM(15) / SIGKILL(9) / ExitCode(143) as `Ok(())` when `dataflow.stop_sent` is true. The intent — "don't classify a clean shutdown as a node failure" — is right, but the implementation has a load-bearing problem:

The current daemon already has a separate `NodeErrorCause::GraceDuration` classification (`binaries/daemon/src/lib.rs:3576`) for nodes that get force-killed after the stop grace period expires. That's a *deliberate* signal: "this node did not shut down cleanly within the grace period; the operator should fix it." Reclassifying those exits as `Ok(())` would silently mask exactly the bugs that signal exists to surface — Python nodes that ignore STOP, blocking C++ destructors, etc. SIGKILL especially: a SIGKILL after `stop_sent` is always the daemon giving up on a non-cooperative node; calling that "success" hides real bugs.

If you want to address "Python nodes get classified as failures even when they exit cleanly on STOP," the fix is on the Python node side: make the nodes actually exit 0 on STOP (which most current examples already do). The daemon's strict classification is doing useful work and shouldn't be relaxed.

---

**Where the thinking could land**

- **Python node shutdown ergonomics**: if you have a repro of a current example or test node that exits via signal-kill on STOP rather than 0, open an issue with the repro — that's a node-side bug worth fixing.
- **Docs polish**: if you spot specific gaps in `guide/src/development/testing.md` after reading it end-to-end, a focused PR adding those would be welcome.
- **Other entry points**: a few currently-applicable `good first issue` tickets:
  - [#885](https://github.com/dora-rs/dora/issues/885) — Improve Users-Getting Started on dora-rs.ai
  - [#829](https://github.com/dora-rs/dora/issues/829) — Use pyo3 experimental-inspect instead of `generate_stubs.py`
  - [#540](https://github.com/dora-rs/dora/issues/540) — Thread-safety of C node API

PR #1372, opened 2026-02-25
Migrate coordinator-daemon communication to tarpc RPC

phil-opp:
Replace the hand-rolled TCP request-reply protocol between the coordinator and daemons with tarpc, a Rust RPC framework that generates type-safe client/server stubs from a trait definition.

This eliminates manual message serialization, explicit reply matching, and the `DaemonCoordinatorEvent`/`DaemonCoordinatorReply` enums for the coordinator→daemon direction, replacing them with a `DaemonControl` tarpc service trait.

### Protocol layer (`libraries/message`)
- Add `DaemonControl` tarpc service trait defining all coordinator→daemon RPCs: `build`, `spawn`, `all_nodes_ready`, `stop_dataflow`, `reload_dataflow`, `logs`, `destroy`, `heartbeat`
- Simplify `DestroyResult` variant (remove `notify` field with skip-serde oneshot sender)

### Coordinator (`binaries/coordinator`)
- On daemon registration, set up a tarpc client over the TCP stream instead of wrapping it in `Arc<Mutex<TcpStream>>`
- Replace all `tcp_send`/`tcp_receive` + manual serde with typed tarpc client calls (e.g. `client.stop_dataflow(...)`)
- `DaemonConnection` now holds a `DaemonControlClient` and `peer_addr` instead of a raw `Arc<Mutex<TcpStream>>`
- Add `get()` method to `DaemonConnections` for immutable access
- Minimize DashMap lock durations: extract data before async I/O in `ReadyOnDaemon`, `heartbeat`, and `stop_dataflow` handlers

### Daemon (`binaries/daemon`)
- Run a tarpc `DaemonControl` server on the registered TCP stream
- Extract shared mutable state into `DaemonState` struct to allow the tarpc server handler to access daemon state
- Reduce DashMap lock durations: remove unnecessary async wrappers for synchronous operations (`OutputClosed`, `NodeFailed`, `EventStreamDropped`), move zenoh publisher creation outside locks

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/binaries/coordinator/src/listener.rs
+++ b/binaries/coordinator/src/listener.rs
@@ -75,6 +75,7 @@ pub async fn handle_connection(
             CoordinatorRequest::Register(register_request) => {
                 let event = DaemonRequest::Register {
                     connection,
+                    version_check_result: register_request.check_version(),
                     machine_id: register_request.machine_id,
                 };
                 let _ = events_tx.send(Event::Daemon(event)).await;
--- a/libraries/message/src/daemon_to_coordinator.rs
+++ b/libraries/message/src/daemon_to_coordinator.rs
@@ -3,7 +3,9 @@ use std::collections::BTreeMap;
 pub use crate::common::{
     DataMessage, LogLevel, LogMessage, NodeError, NodeErrorCause, NodeExitStatus, Timestamped,
 };
-use crate::{BuildId, DataflowId, common::DaemonId, id::NodeId};
+use crate::{
+    BuildId, DataflowId, common::DaemonId, current_crate_version, id::NodeId, versions_compatible,
+};
 
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub enum CoordinatorRequest {
@@ -28,12 +30,31 @@ pub enum CoordinatorRequest {
 
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct DaemonRegisterRequest {
+    dora_version: semver::Version,
     pub machine_id: Option<String>,
 }
 
 impl DaemonRegisterRequest {
     pub fn new(machine_id: Option<String>) -> Self {
-        Self { machine_id }
+        Self {
+            dora_version: current_crate_version(),
+            machine_id,
+        }
+    }
+
+    pub fn check_version(&self) -> Result<(), String> {
+        let crate_version = current_crate_version();
+        let specified_version = &self.dora_version;
+
+        if versions_compatible(&crate_version, specified_version)? {
+            Ok(())
+        } else {
+            Err(format!(
+                "version mismatch: message format v{} is not compatible \
+                with expected message format v{crate_version}",
+                self.dora_version
+            ))
+        }
     }
 }
 
--- a/libraries/message/src/lib.rs
+++ b/libraries/message/src/lib.rs
@@ -86,13 +86,13 @@ pub fn check_version_compatibility(remote_version: &str) -> eyre::Result<()> {
     }
 }
 
-fn current_crate_version() -> semver::Version {
+pub(crate) fn current_crate_version() -> semver::Version {
     let crate_version_raw = env!("CARGO_PKG_VERSION");
 
     semver::Version::parse(crate_version_raw).unwrap()
 }
 
-fn versions_compatible(
+pub(crate) fn versions_compatible(
     crate_version: &semver::Version,
     specified_version: &semver::Version,
 ) -> Result<bool, String> {
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -366,6 +366,7 @@ async fn start_inner(
                 DaemonRequest::Register {
                     machine_id,
                     mut connection,
+                    version_check_result,
                 } => {
                     let existing = match &machine_id {
                         Some(id) => coordinator_state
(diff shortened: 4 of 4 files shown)


libraries/message/src/daemon_to_coordinator.rs
```diff
@@ -3,76 +3,40 @@ use std::collections::BTreeMap;
...
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub enum CoordinatorRequest {
     Register(DaemonRegisterRequest),
-    Event {
+    /// Register a reverse-channel connection for daemon→coordinator RPC.
+    ///
+    /// Sent on a second TCP connection after the initial registration.
+    /// The coordinator sets up a `DaemonToCoordinatorControl` tarpc server
+    /// on this connection.
+    RegisterReverseChannel {
```
phil-opp: I would prefer the name `RegisterNoficationChannel`

libraries/message/src/daemon_to_coordinator.rs
```diff
@@ -100,16 +64,27 @@ impl DataflowDaemonResult {
...
-        notify: Option<tokio::sync::oneshot::Sender<()>>,
-    },
-    Logs(Result<Vec<u8>, String>),
+/// tarpc service for daemon→coordinator RPC calls.
+///
+/// This replaces the raw TCP `DaemonEvent` messages sent over
+/// `coordinator_connection`. The coordinator runs a tarpc server
+/// implementing this trait, and each daemon holds a client.
+#[tarpc::service]
+pub trait DaemonToCoordinatorControl {
```
phil-opp: How about naming this just `DaemonNotification` (or similar)?. All these methods are essentially notifying the coordinator about some event that happened on the daemon, so it's not really a control channel.

libraries/message/src/lib.rs
```diff
@@ -69,6 +69,23 @@ impl std::fmt::Display for BuildId {
...
+        .map_err(|e| eyre::eyre!("failed to parse remote version `{remote_version}`: {e}"))?;
+    let compatible = versions_compatible(&crate_version, &specified_version)
+        .map_err(|e| eyre::eyre!(e))?;
+    if compatible {
+        Ok(())
+    } else {
+        Err(eyre::eyre!(
+            "version mismatch: daemon message format v{specified_version} is not compatible \
+            with coordinator message format v{crate_version}"
+        ))
```
phil-opp, line 85: This error message is specific to the daemon/coordinator crates, so it should be moved there.

binaries/daemon/src/lib.rs
```diff
@@ -173,18 +175,47 @@ impl Daemon {
...
+            ReceiverStream::new(ctrlc_events),
+            incoming_events,
+            dora_events,
+            watchdog_interval,
+            metrics_interval,
         )
-        .await
-        .map(|_| ())
+            .merge();
+        daemon.run_inner(events).await.map(|_| ())
```
phil-opp: Why not reuse run_general anymore? The whole point of run_general was that it is shared by both run variants

binaries/daemon/src/pending.rs
```diff
@@ -67,61 +66,102 @@ impl PendingNodes {
...
-                    Some(node_id.clone()),
-                    Some("daemon::pending".into()),
-                    "node exited before initializing dora connection",
-                )
-                .await;
+            tracing::warn!(
+                dataflow_id = %self.dataflow_id,
+                %node_id,
+                "node exited before initializing dora connection"
+            );
```
phil-opp: I would prefer to keep the previous logger call because the message also reaches the CLI this way.

binaries/daemon/src/pending.rs
```diff
@@ -202,39 +248,25 @@ impl PendingNodes {
...
-                "all local nodes are ready (exit before subscribe: {:?}), waiting for remote nodes",
-                self.exited_before_subscribe
-            ),
-            )
-            .await;
+        tracing::info!(
+            dataflow_id = %self.dataflow_id,
+            "all local nodes are ready (exit before subscribe: {:?}), waiting for remote nodes",
+            self.exited_before_subscribe
+        );
```
phil-opp: Same here (I would prefer to keep the logger if possible)

binaries/daemon/src/coordinator.rs
```diff
@@ -1,36 +1,59 @@
...
 
 const DAEMON_COORDINATOR_RETRY_INTERVAL: std::time::Duration = Duration::from_secs(1);
 
-#[derive(Debug)]
-pub struct CoordinatorEvent {
-    pub event: DaemonCoordinatorEvent,
-    pub reply_tx: oneshot::Sender<Option<DaemonCoordinatorReply>>,
+/// Result of [`register`]: everything the daemon needs after connecting to the
+/// coordinator.
+pub struct RegisterResult2 {
```
phil-opp: This is not really a good name

binaries/coordinator/src/listener.rs
```diff
@@ -60,99 +68,136 @@ pub async fn handle_connection(
...
+
+/// tarpc server that handles daemon→coordinator RPC calls.
+///
+/// Each daemon gets its own server instance, identified by `daemon_id`.
+/// The server translates incoming RPC calls into coordinator `Event`s.
+#[derive(Clone)]
+pub struct DaemonEventServer {
+    pub daemon_id: DaemonId,
+    pub events_tx: mpsc::Sender<Event>,
+}
```
phil-opp, line 113: Can we also add a state reference to this server and handle requests in place (instead of forwarding them to the coordinator event loop)?

PR #1381, opened 2026-02-27
bump pyo3 to 0.28

drindr:
- [x] bump pyo3 to 0.28
- [x] fix the deprecated methods
- [ ] ci


.github/workflows/pip-release.yml
```diff
@@ -97,167 +70,107 @@ jobs:
...
             name: dora-rs-cli
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
         with:
-          python-version: 3.8
-          architecture: ${{ matrix.platform.target }}
+          python-version: 3.13
```
haixuanTao, line 122: The problem with choosing very high python is that it makes it incompatible with lower python version

apis/python/operator/Cargo.toml
```diff
@@ -13,7 +13,7 @@ repository.workspace = true
 
 [dependencies]
 dora-node-api = { workspace = true }
-pyo3 = { workspace = true, features = ["eyre", "abi3-py37"] }
+pyo3 = { workspace = true, features = ["eyre"] }
```
haixuanTao, line 16: THe abi is very important for cross python compatibility

binaries/runtime/Cargo.toml
```diff
@@ -27,7 +27,7 @@ serde_yaml = { workspace = true }
 tokio = { version = "1.24.2", features = ["full"] }
 tokio-stream = "0.1.8"
 # pyo3-abi3 flag allow simpler linking. See: https://pyo3.rs/v0.13.2/building_and_distribution.html
-pyo3 = { workspace = true, features = ["eyre", "abi3-py37"], optional = true }
```
haixuanTao, line 30: Same

libraries/extensions/ros2-bridge/python/src/typed/mod.rs
```diff
@@ -57,12 +57,12 @@ mod tests {
...
     #[test]
     fn test_python_array_code() -> Result<()> {
-        pyo3::prepare_freethreaded_python();
+        Python::initialize();
         let context = Ros2Context::new(None).context("Could not create a context")?;
         let messages = context.messages.clone();
         let serializer = Serializer::builder().build();
 
-        Python::with_gil(|py| -> Result<()> {
+        Python::attach(|py| -> Result<()> {
```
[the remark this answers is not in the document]
haixuanTao: I think I agree with this

Cargo.toml
```diff
@@ -76,19 +76,21 @@ dora-ros2-bridge-python = { path = "libraries/extensions/ros2-bridge/python" }
...
-arrow-array = { version = "54.2.1" }
-arrow-json = { version = "54.2.1" }
-parquet = { version = "54.2.1" }
-pyo3 = { version = "0.23", features = [
+arrow = { version = "58.0.0" }
+arrow-schema = { version = "58.0.0" }
+arrow-data = { version = "58.0.0" }
+arrow-array = { version = "58.0.0" }
+arrow-json = { version = "58.0.0" }
+parquet = { version = "58.0.0" }
```
haixuanTao, line 84: do we have to change this here?

haixuanTao (changes requested):

heyong4725:
Hi @drindr, closing this as superseded.

pyo3 0.28 is the current workspace version on `main` ([`Cargo.toml:121`](https://github.com/dora-rs/dora/blob/main/Cargo.toml#L121)) — the upgrade landed independently after this PR was opened.

For completeness, the two original review concerns on this branch were:
1. **Local-path workspace dep** — the `pyo3-special-method-derive` dependency was pointing at `/home/drin/workspace/dora/...` which would have broken the build for every other contributor.
2. **ABI compat** — @haixuanTao flagged that bumping from `abi3-py37` to `abi3-py311` would have dropped Python 3.7-3.10 support, which dora wants to keep.

The version-on-main upgrade kept the abi3 floor intact, so neither concern survived. Thanks for the original push to get pyo3 0.28 in.

PR #1384, opened 2026-03-01
feat(python): use pyo3 experimental-inspect for automatic type hints

ravencore06:
Currently, `dora` uses a custom Python script ([generate_stubs.py] to generate type hinting for the Python API. This approach requires manual intervention and relies on deprecated features.

This PR replaces the manual type hinting script with PyO3's built-in `experimental-inspect` feature. This enables automatic and reliable generation of Python stubs directly through the Rust bindings without requiring a secondary external script. 

Changes Made:
- Updated workspace [Cargo.toml] to include the `experimental-inspect` feature flag for `pyo3`.
- Removed the discontinued [generate_stubs.py] script and statically generated [__init__.pyi]
- Removed the outdated "Type hinting" section from `apis/python/node/README.md`.


swar09:
Do you verified this fix ? @ravencore06  
type hinting actually generated ?

phil-opp:
Thanks for the PR! I don't think that the pyi files are generated by this, so we should probably keep them for now.

Closed in favor of #1394, which just enables the feature.

PR #1385, opened 2026-03-01
Fix/yolo getting started docs

LIKHITHA-NAGRAJ:
This PR improves the Getting Started Python operator dataflow example with better visualization, stability, and real-time performance observability.
Changes
* Headless CI fix: Set QT_QPA_PLATFORM=offscreen before cv2 import so the example runs in GitHub Actions without a display
* Dynamic reshape: Changed hardcoded (480, 640, 3) to (-1, CAMERA_WIDTH, 3) so it works on any camera resolution
Empty bbox fix: Changed np.array([]) to np.zeros((0, 6)) to prevent crash when no objects are detected
* Clean metadata: Passing {} instead of dora_event["metadata"] fixes the Arrow type conversion error
* Performance overlay: Added real-time FPS and AI latency display
Readable labels: Added filled background rectangle behind label text
* Optimized inference: Using imgsz=320 reduces latency significantly

BEFORE
Screenshot (85)
AFTER
Screenshot (153)


LIKHITHA-NAGRAJ:
Screenshot (323)
Screenshot (322)
Screenshot (319)

LIKHITHA-NAGRAJ:
Hi @haixuanTao — conflicts are now resolved and all CI checks are passing (28/28). Tested locally with working webcam detection (person: 0.92, cell phone: 0.87, FPS: 17–23, latency: 50–66ms). Ready for review!

heyong4725:
Hi @LIKHITHA-NAGRAJ — thanks for the effort here, and for coming back to resolve the conflicts in April. I want to be honest with you about why I'm closing this rather than asking for another round of revisions; this is feedback that I hope is useful for your next PR.

**On the valid pieces** — there's real signal in some of these hunks:

- The empty-bbox fix (`np.zeros((0, 6))` instead of `np.array([])`) is a genuine bug fix.
- The dynamic reshape (`(-1, CAMERA_WIDTH, 3)`) is a useful robustness improvement over the hardcoded 480.
- The headless display fix (`QT_QPA_PLATFORM=offscreen` when `GITHUB_ACTIONS=true`) is plausibly worthwhile given that `plot.py:101` does call `cv2.imshow`.
- The FPS/latency overlay is a nice demonstration pattern.

**On the pieces that block merging as-is:**

1. **A root-level `dataflow.yml` (+29 lines)** was committed at the repo root — it's a near-duplicate of `examples/python-operator-dataflow/dataflow.yml` and shouldn't exist there. This is the kind of stray file that signals the diff wasn't reviewed end-to-end before pushing. Worth double-checking `git status` before every PR.

2. **The `frame[:, :, ::-1]` BGR→RGB conversion was removed** from `object_detection.py` with the comment "No BGR→RGB flip needed — Ultralytics handles color internally." This isn't quite right: cv2 captures frames in BGR by default and Ultralytics' `model(frame)` call expects RGB unless explicitly told otherwise. Removing the flip will silently degrade detections (wrong colors → mis-classified objects, especially on color-sensitive classes). This is the most consequential issue because the whole point of the demo is YOLO detection working correctly — and the regression isn't visible from a quick glance at the rendered output, only from comparing detection accuracy before/after.

3. **`dora_event["metadata"]` was replaced with `{}`** in `send_output()`. The comment cites an Arrow type conversion error, but dropping the metadata also drops timing/correlation data that downstream nodes can use. If there's a real Arrow conversion error, the fix should be to surface a minimal repro rather than discard the metadata entirely.

4. **Existing module-level docstrings were removed** (`webcam.py`, `object_detection.py`). For an `examples/` directory where the docs *are* the value, this is a net loss.

**On the broader context** — the `examples/python-operator-dataflow/` example uses dora's legacy **operator API**, not the modern node API that 1.0 is consolidating around. It is exercised in nightly CI (`scripts/qa/ci-nightly-jobs.sh:910`, `.github/workflows/nightly.yml:1210`) but is intentionally skipped in the per-commit smoke suite (`scripts/smoke-all.sh:460` — "webcam + YOLOv5 + torch" is too heavyweight). The `--stop-after 20s` flag in nightly handles termination externally, so the in-code CI auto-stop is redundant in either form.

Net: the valuable hunks add up to ~30 lines, but extracting them cleanly would require dropping the root yml, restoring the BGR→RGB flip, restoring the docstrings, and verifying the metadata claim against current dora. The rework-to-value ratio doesn't fit the current 1.0 RC consolidation phase, where most maintainer attention is focused on subsystem-level correctness work.

**Where to spend your effort next**, if you'd like to keep contributing:

- Pick a smaller, more focused change next time — one PR per logical fix makes review much faster. A "fix empty bbox crash" PR with just that one-line change would have been merged in a day.
- A few currently-applicable `good first issue` tickets:
  - [#885](https://github.com/dora-rs/dora/issues/885) — Improve Users-Getting Started on dora-rs.ai
  - [#829](https://github.com/dora-rs/dora/issues/829) — Use pyo3 experimental-inspect instead of `generate_stubs.py`
  - [#540](https://github.com/dora-rs/dora/issues/540) — Thread-safety of C node API

Thanks again — the bug fixes you spotted are real, and I appreciate the work.

PR #1388, opened 2026-03-02
fix high cpu usage when event/drop stream errors keep retrying 

### LINKED ISSUE #1124: high CPU usage during running for a long time
Image

Image

Long-term task polling can lead to a gradual increase in CPU usage.

here is my test dataflow:

nodes:
  - id: mcp-server
    build: cargo build -p dora-mcp-server --release
    path: /xdfapp/rust-target/release/dora-mcp-server
    outputs:
      - search_request
    inputs:
      search_reply: question-search/reply
    env:
      CONFIG: mcp_config.toml

  - id: question-search
    build: cargo build -p question-search
    path: /Users/lijing/xdfapp/rust-target/debug/question-search
    inputs:
      text: mcp-server/search_request
    outputs:
      - reply

Agent: moly

afrdbaig7:
Fixes #1124

The event_stream_loop and drop_stream_loop  functions have a continue branch for error conditions (daemon errors, unexpected reply, channel errors). Without a delay, this results in a tight spin loop in the event of an error, and this is what causes the increasing CPU load in a data flow.

Added a sleep before each continue to make the thread back off instead of spinning on the channel.

### DIFF
--- a/apis/rust/node/src/event_stream/thread.rs
+++ b/apis/rust/node/src/event_stream/thread.rs
@@ -129,8 +129,7 @@ fn event_stream_loop(
             }
             Err(err) => {
                 let err = err.wrap_err("failed to receive incoming event");
-                tracing::warn!("{err:?}");
-                continue;
+                break 'outer Err(err);
             }
         };
         for Timestamped { inner, timestamp } in events {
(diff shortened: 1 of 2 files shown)


afrdbaig7:
> Thanks!
Helllo @phil-opp would  sugesst the issues that you would want me to work on

PR #1389, opened 2026-03-03
fix: use os.path.join for robust path handling in llm_op

RithamSharma:


examples/python-operator-dataflow/llm_op.py
```diff
@@ -299,7 +299,7 @@ def ask_llm(self, prompt):
     # Directory of the current file
     current_directory = os.path.dirname(current_file_path)
 
-    path = current_directory + "object_detection.py"
+    path = os.path.join(current_directory,"object_detection.py")  # ← FIXED
```
phil-opp, line 302: Please remove the `# ← FIXED` comment.

phil-opp (commented):
Thanks for the PR!

heyong4725:
Hi @RithamSharma — thanks for spotting this bug and posting the fix. Closing this PR but the fix is being rescued and credited to you in [#1837](https://github.com/dora-rs/dora/pull/1837).

Your analysis was right: `current_directory + "object_detection.py"` is missing the path separator, and `os.path.join` is the correct idiom. I lifted that one-line change verbatim, with two tiny cosmetic cleanups before merging:

1. Dropped the trailing `# ← FIXED` annotation (debug marker, doesn't belong in the example file)
2. Added PEP 8 spacing after the comma in `os.path.join(current_directory, "object_detection.py")`

You're credited via `Co-Authored-By` on the commit, so when #1837 lands your name will be on it as a co-author. The PR description also links back to your original PR explicitly.

Two notes for next time, both small:

- For trivial fixes, don't add inline markers like `# ← FIXED` in the patched line — they're useful while you're staging the change locally, but they end up shipped to other readers if left in. A clean diff is just the corrected line.
- PEP 8 wants a space after commas in function calls: `f(a, b)` not `f(a,b)`. The `ruff` / `black` configs in this repo will flag this automatically if you have them set up locally.

PR #1394, opened 2026-03-06
Enable pyo3 `experimental-inspect` feature

VIKAS7216:


VIKAS7216:
This change help the code to reduce complexity of code as we now do not have to run a script to generate stubs it is done by the built-in feature of pyo3 and  it also follows the latest standards and is compatible with modern Python.

PR #1409, opened 2026-03-16
feat(c++/node): add event receive variants (timeout, non-blocking, drain)

PavelGuzenfeld:
## Summary
- Add `try_next_event()` for non-blocking event polling
- Add `next_event_timeout(events, timeout_ms)` for timed blocking receive
- Add `events_is_empty()` to check if events are buffered
- Add `drain_events()` / `drained_events_len()` / `drained_events_next()` for batch draining
- Add `DoraEventType::Timeout` variant to distinguish timeout from stream closure

Refactors `DoraEvent` from tuple struct to named fields (`event`, `timed_out`) to cleanly distinguish timeout from real errors — `recv_timeout` returns `Event::Error("Receiver timed out")` on timeout, which we intercept.

## Usage (C++)
```cpp
// Non-blocking poll
auto poll = try_next_event(dora_node.events);
if (event_type(poll) == DoraEventType::Timeout) {
    // no event ready
}

// Timed blocking (100ms)
auto ev = next_event_timeout(dora_node.events, 100);

// Drain all buffered events
auto drained = drain_events(dora_node.events);
auto len = drained_events_len(drained);
```

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -1,11 +1,11 @@
-use std::{any::Any, collections::BTreeMap, vec};
+use std::{any::Any, collections::BTreeMap, time::Duration, vec};
 
 use crate::ffi::MetadataValueType;
 
 use chrono::DateTime;
 use dora_node_api::{
     self, Event, EventStream, Metadata as DoraMetadata,
-    MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter,
+    MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter, TryRecvError,
     arrow::array::{AsArray, UInt8Array},
     merged::{MergeExternal, MergedEvent},
 };
@@ -36,6 +36,7 @@ mod ffi {
         Error,
         Unknown,
         AllInputsClosed,
+        Timeout,
     }
 
     struct DoraInput {
@@ -80,6 +81,7 @@ mod ffi {
         type MergedDoraEvent;
         type Metadata;
         type DataSampleHandle;
+        type DrainedEvents;
 
         fn init_dora_node() -> Result<DoraNode>;
         fn init_dora_node_from_id(node_id: String) -> Result<DoraNode>;
@@ -92,7 +94,14 @@ mod ffi {
         fn empty_combined_events() -> CombinedEvents;
         fn next(self: &mut Events) -> Box<DoraEvent>;
         fn next_event(events: &mut Box<Events>) -> Box<DoraEvent>;
+        fn next_event_timeout(events: &mut Box<Events>, timeout_ms: u64) -> Box<DoraEvent>;
+        fn try_next_event(events: &mut Box<Events>) -> Box<DoraEvent>;
+        fn events_is_empty(events: &Box<Events>) -> bool;
         fn event_type(event: &Box<DoraEvent>) -> DoraEventType;
+
+        fn drain_events(events: &mut Box<Events>) -> Box<DrainedEvents>;
+        fn drained_events_len(drained: &Box<DrainedEvents>) -> usize;
+        fn drained_events_next(drained: &mut Box<DrainedEvents>) -> Box<DoraEvent>;
         fn event_as_input(event: Box<DoraEvent>) -> Result<DoraInput>;
         fn send_output(
             output_sender: &mut Box<OutputSender>,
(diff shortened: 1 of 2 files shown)


heyong4725:
Hi @PavelGuzenfeld — thanks for the original work on this and the careful follow-up to @phil-opp's review on 2026-03-18. Going to close this PR but the work is being rescued and credited to you in [#1846](https://github.com/dora-rs/dora/pull/1846), which is the first PR in a 4-PR rescue of your full stack (#1409 → #1410 → #1413 → #1414).

Here's what changed in the rescue vs. this PR:

**Applied your promised review fixes.** Your replies to @phil-opp on 2026-03-18 said "Fixed" for the three review concerns (string-matching error messages, `Timeout`-vs-`Empty` overloading on non-blocking poll, same on drained-events exhaustion), but the actual fixes never landed on the PR branch — it stayed at `041528e2`. The rescue does what your replies said:

- `next_event_timeout` now uses `try_recv` polled against an `Instant`-based deadline instead of `recv_timeout` + `msg.contains("timed out")`. Timeout vs. stream-close is disambiguated structurally via `TryRecvError::Empty` / `Closed`.
- New `DoraEventType::Empty` variant. `try_next_event` returns it on empty queue instead of overloading `Timeout`.
- `drained_events_next` returns `Empty` once the snapshot is exhausted (a drain is frozen and can't time out).

Internal `DoraEvent` switches from `(Option<Event>)` to `(EventOrReason)` so the no-event-available state can carry its reason — pretty much the design pivot you described in your review reply.

**Dropped the tangled-in additions** that have been outdated since main moved on:

- `init_dora_node_from_id` / `init_dora_node_flexible` / `node_id()` / `dataflow_id()` — these depended on PR #1403 which got wiped by the 1.0 consolidation `145ccce0`. Separate work if/when the node-id surface comes back.
- Zero-copy `DataSampleHandle` API — already shipped in main via #1431 with a different shape.
- `log_message` removal — not applied; current main uses `log_message` for C/C++ logging.

So the rescue is a narrow extraction of just the event-receive variants (the named feature in your title) with your review fixes applied.

**Credit retained** via `Co-Authored-By` on the commit and explicit `Rescue of #1409` in the PR description.

The other three rescues (#1410, #1413, #1414) are coming next, stacked on top of #1846. They'll go through the same extraction process (drop tangled-in changes that don't apply to current main, keep the named feature, apply any pending review fixes).

For future stacked PRs targeting this kind of surface, two things that would have helped:

1. **Push your "Fixed" reply changes as commits.** Your acknowledgment of the review was thoughtful and your design pivot was the right call — it just never made it from your reply box to the branch. Force-pushing the fixes (or adding a new commit) would have made the review actionable on the maintainer side.
2. **Keep the PR scoped to its title.** This PR carried 4-5 unrelated additions beyond the event variants. When the surface moves under you (as the 1.0 consolidation did), a single-feature PR is easier to rescue than a tangled one.

Looking forward to your next contribution. Reopen any of these (or pick up the unfinished node-id / accessors work) whenever you'd like.

heyong4725:
Correction on a thing I got wrong in my close comment above:

I wrote "your replies to @phil-opp on 2026-03-18 said 'Fixed' for the three review concerns ... but the actual fixes never landed on the PR branch — it stayed at `041528e2`." That's technically true for **this** PR's branch specifically, but it implies you never followed through, which isn't fair.

Reading further into your stack just now (while rescuing #1414), I found commit [`12920f5f`](https://github.com/dora-rs/dora/commit/12920f5f) — "fix(c++/node): replace fragile timed_out bool with NoEventReason enum" — on top of #1414's branch. That IS the design pivot phil-opp asked for: introducing `NoEventReason::HasEvent / Closed / TimedOut / Empty`, replacing the `timed_out: bool`, and the structural fix to next_event_timeout. You DID push the fixes — just to the top of the stack, which in a stacked-PR workflow is the right place to put cumulative changes that affect earlier commits.

[#1846](https://github.com/dora-rs/dora/pull/1846)'s rescue reimplements the same design (under a different name — `EventOrReason` carries the same four states) from scratch because by the time I noticed your `12920f5f` I had already written the rescue. So your design vision IS what landed, just rebuilt rather than cherry-picked.

Apologies for the unfair characterization. Pinging you here so the correction is on this PR's history alongside the original close, not just buried in #1414's thread.

PR #1413, opened 2026-03-16
feat(c++/node): add node_config_json and dataflow_descriptor_json

PavelGuzenfeld:
## Summary
- Add `node_config_json(sender)` — returns the node's input/output configuration as a JSON string
- Add `dataflow_descriptor_json(sender)` — returns the full dataflow descriptor (parsed YAML) as JSON

Both `NodeRunConfig` and `Descriptor` already derive `Serialize`, so this is a thin wrapper.

> **Stacked on** #1410 (Phase 3) — merge that first, then this PR's diff is just the 1 new commit.

## Usage (C++)
```cpp
auto config = node_config_json(dora_node.send_output);
std::cout << "Node config: " << std::string(config) << std::endl;

auto descriptor = dataflow_descriptor_json(dora_node.send_output);
// Full dataflow topology as JSON
```

PR #1410, opened 2026-03-16
feat(c++/node): add close_outputs, NodeFailed and Reload event types

PavelGuzenfeld:
## Summary
- Add `close_outputs(sender, output_ids)` to selectively close node outputs from C++
- Add `DoraEventType::NodeFailed` and `DoraEventType::Reload` enum variants
- Add `event_as_node_failed(event)` accessor returning `DoraNodeFailed` struct with affected input IDs, error message, and source node ID

> **Stacked on** #1409 (Phase 2) — merge that first, then this PR's diff is just the 1 new commit.

## Usage (C++)
```cpp
// Close specific outputs
rust::Vec<rust::String> ids;
ids.push_back("output_a");
auto result = close_outputs(dora_node.send_output, ids);

// Handle node failure events
if (ty == DoraEventType::NodeFailed) {
    auto failed = event_as_node_failed(std::move(event));
    std::cerr << "Node " << failed.source_node_id << " failed: " << failed.error;
}
```

PR #1414, opened 2026-03-16
feat(c++/operator): forward InputClosed and Stop events to C++ callbacks

PavelGuzenfeld:
## Summary
- Add `on_input_closed(op, id)` callback — called when an input stream closes
- Add `on_stop(op)` callback — called on graceful shutdown
- Previously these events were silently ignored (`_ => Ok(DoraStatus::Continue)`)

> **Stacked on** #1413 (Phase 4) — merge that first, then this PR's diff is just the 1 new commit.

### Breaking change
Existing C++ operators must implement the two new callbacks. Minimal migration:
```cpp
// Add to operator.h:
DoraOnInputResult on_input_closed(Operator &op, rust::Str id);
DoraOnInputResult on_stop(Operator &op);

// Add to operator.cc:
DoraOnInputResult on_input_closed(Operator &op, rust::Str id) {
    return {rust::String(), false};  // continue by default
}
DoraOnInputResult on_stop(Operator &op) {
    return {rust::String(), false};
}
```

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{any::Any, collections::BTreeMap, time::Duration, vec};
+use std::{any::Any, collections::BTreeMap, time::{Duration, Instant}, vec};
 
 use crate::ffi::MetadataValueType;
 
@@ -37,6 +37,7 @@ mod ffi {
         Unknown,
         AllInputsClosed,
         Timeout,
+        Empty,
         NodeFailed,
         Reload,
     }
@@ -255,10 +256,7 @@ pub struct Events(EventStream);
 
 impl Events {
     fn next(&mut self) -> Box<DoraEvent> {
-        Box::new(DoraEvent {
-            event: self.0.recv(),
-            timed_out: false,
-        })
+        Box::new(DoraEvent::from_recv(self.0.recv()))
     }
 }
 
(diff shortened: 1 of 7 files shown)


apis/c++/node/src/lib.rs
```diff
@@ -180,14 +207,86 @@ pub struct Events(EventStream);
...
 
 fn next_event(events: &mut Box<Events>) -> Box<DoraEvent> {
     events.next()
 }
 
+fn next_event_timeout(events: &mut Box<Events>, timeout_ms: u64) -> Box<DoraEvent> {
+    let dur = Duration::from_millis(timeout_ms);
+    match events.0.recv_timeout(dur) {
+        Some(event) => {
+            let timed_out = matches!(&event, Event::Error(msg) if msg.contains("timed out"));
```
phil-opp: String matching the error message seems fragile. The error message might change in the future.

apis/c++/node/src/lib.rs
```diff
@@ -180,14 +207,86 @@ pub struct Events(EventStream);
...
+
+fn try_next_event(events: &mut Box<Events>) -> Box<DoraEvent> {
+    match events.0.try_recv() {
+        Ok(event) => Box::new(DoraEvent {
+            event: Some(event),
+            timed_out: false,
+        }),
+        Err(TryRecvError::Empty) => Box::new(DoraEvent {
+            event: None,
+            timed_out: true,
```
phil-opp: It's not really timed out, is it? There was no timeout set.

apis/c++/node/src/lib.rs
```diff
@@ -180,14 +207,86 @@ pub struct Events(EventStream);
...
+
+fn drained_events_next(drained: &mut Box<DrainedEvents>) -> Box<DoraEvent> {
+    match drained.0.pop_front() {
+        Some(e) => Box::new(DoraEvent {
+            event: Some(e),
+            timed_out: false,
+        }),
+        None => Box::new(DoraEvent {
+            event: None,
+            timed_out: true,
```
phil-opp: Why is this timed out?

apis/c++/node/src/lib.rs
PavelGuzenfeld: Fixed — replaced `recv_timeout` + string matching with a `try_recv` polling loop using `Instant`-based deadline. Timeout is now detected structurally (elapsed time >= deadline), not by parsing error messages.

apis/c++/node/src/lib.rs
PavelGuzenfeld: You're right — fixed. Introduced a `NoEventReason` enum (`HasEvent`, `Closed`, `TimedOut`, `Empty`) to replace the overloaded `timed_out: bool`. `try_next_event` now returns `DoraEventType::Empty` instead of `Timeout`.

apis/c++/node/src/lib.rs
PavelGuzenfeld: Same fix — `drained_events_next` now returns `DoraEventType::Empty` when the deque is exhausted, not `Timeout`.

heyong4725:
Hi @PavelGuzenfeld — closing in favor of [#1849](https://github.com/dora-rs/dora/pull/1849), the rescue of this PR. Fourth and final in the 4-PR rescue stack of your work: [#1846](https://github.com/dora-rs/dora/pull/1846) → [#1847](https://github.com/dora-rs/dora/pull/1847) → [#1848](https://github.com/dora-rs/dora/pull/1848) → #1849.

This is the behavior-bug-fix-wearing-feature-clothing PR — your `_ => Continue` analysis was right; the operator API was silently dropping `Event::InputClosed` and `Event::Stop` for the entire history of the cxx operator surface, and your fix is the right shape.

The rescue is essentially your `38048957` commit applied verbatim, with one small correction: `Event::Stop` on `dora_operator_api::Event` is a unit variant (no payload), unlike `dora_node_api::Event::Stop(StopCause)`. Your original PR did `Event::Stop` (unit) which is correct — I initially typed `Event::Stop(_)` in the rescue and the compiler caught it.

Also a note on the late review fix (`12920f5f` "fix(c++/node): replace fragile timed_out bool with NoEventReason enum"): I'm glad I read further into your stack. My earlier comment on #1846 implied you never pushed phil-opp's review fixes, but you DID — to the top of the stack here on #1414, not back-pushed to #1409. In a stacked workflow that's the right place to put cumulative fixes. I incorrectly characterized this in #1409's close comment; sorry about that. Your design pivot (introducing `NoEventReason` to disambiguate timeout vs empty) is exactly what I rebuilt in #1846's rescue from scratch.

All four rescue PRs credit you via `Co-Authored-By` on the commit and explicitly reference `Rescue of #1409 / #1410 / #1413 / #1414` in their bodies. Once they land you'll have a full git-attributed trail.

Thanks for the careful work on this whole stack — both the feature surface and the late design fix. Sorry it took 2 months to surface this.

PR #1417, opened 2026-03-17
fix: use unbounded channel for drop token stream to prevent shm leak

haixuanTao:
## Summary

- Change `flume::bounded(0)` to `flume::unbounded()` in the drop token stream channel (`drop_stream.rs`)

## Problem

The drop stream uses a rendezvous channel (`bounded(0)`) between the background thread that receives drop tokens from the daemon and the node's `try_recv()` calls. The background thread blocks on `tx.send()` until the node calls `try_recv()`, which only happens inside `handle_finished_drop_tokens()` at the start of each `send_output()`.

For nodes publishing at moderate rates (e.g. 5Hz camera), the sender creates new `/dev/shm/shmem_*` shared memory segments faster than drop tokens can flow back through the rendezvous channel. Each unreclaimed token keeps a shared memory segment mapped, causing unbounded growth in `/dev/shm` that eventually fills it completely — silently crashing downstream nodes that rely on POSIX shared memory (e.g. Python multiprocessing).

## Testing

Tested with a minimal 2-node dataflow (camera → sink) on Ubuntu with 32GB `/dev/shm`:

| Metric | Before (bounded(0)) | After (unbounded) |
|--------|---------------------|-------------------|
| /dev/shm segments after 2 min | ~180 (growing) | 0 (stable) |
| /dev/shm usage after 5 min | ~1.3 GB (growing) | 4.5 MB (stable) |
| Time to fill 32GB /dev/shm | ~2 hours | never |

Also verified with a full 9-node robotics dataflow (camera, SAM3, motion planner, VLM, etc.) — zero steady-state shared memory segments.

## Test plan

- [x] Minimal dataflow: camera + dummy sink, monitor `/dev/shm` for 5 minutes
- [x] Full dataflow: 9 nodes with GPU workloads, monitor `/dev/shm`
- [ ] Verify no regression in node shutdown / drop token cleanup on exit

### DIFF
--- a/apis/rust/node/src/node/drop_stream.rs
+++ b/apis/rust/node/src/node/drop_stream.rs
@@ -80,7 +80,7 @@ impl DropStream {
             other => eyre::bail!("unexpected drop subscribe reply: {other:?}"),
         }
 
-        let (tx, rx) = flume::bounded(0);
+        let (tx, rx) = flume::unbounded();
         let node_id_cloned = node_id.clone();
 
         let handle = std::thread::spawn(|| drop_stream_loop(node_id_cloned, tx, channel, clock));


phil-opp (approved):
Good find!

PR #1423, opened 2026-03-17
fix(test): handle Stop events correctly in queue timeout test

### LINKED ISSUE #1422: Flaky queue timeout Python test due to timeout based termination
## Summary
  `tests/queue_size_and_timeout_python` is flaky in CI because loop termination relies on timeout behavior instead of explicit Stop events.

  ## Current behavior
  The sender/receiver loops use `dora_node.next(timeout=...)` and may exit when it returns `None`.
  This can happen nondeterministically under CI timing/load.

  ## Expected behavior
  Loop termination should happen only when a `STOP` event is received.

  ## Proposed fix
  - Handle `STOP` explicitly in both sender and receiver loops
  - Remove timeout-based `None` early-exit logic
  - Remove obsolete keep-alive timer input in this test dataflow if unnecessary

  ## Scope
  Only:
  - `tests/queue_size_and_timeout_python/send_data.py`
  - `tests/queue_size_and_timeout_python/receive_data.py`
  - `tests/queue_size_and_timeout_python/dataflow.yaml`

  No CLI changes, no refactors, no unrelated tests.

Bhanudahiyaa:
## Summary
  Fixes flaky behavior in `tests/queue_size_and_timeout_python` by replacing timeout-based loop termination with explicit `STOP` handling.

  ## Motivation
  The previous test logic used `dora_node.next(timeout=...)` and exited on timeout (`None`), which can cause nondeterministic early exits in CI.

  ## Changes
  - `send_data.py`: exit only on explicit `STOP` event
  - `receive_data.py`: use event-driven loop termination on `STOP` (no timeout-based early break)
  - `dataflow.yaml`: remove obsolete keep-alive timer input

  ## Scope
  This PR only changes:
  - `tests/queue_size_and_timeout_python/send_data.py`
  - `tests/queue_size_and_timeout_python/receive_data.py`
  - `tests/queue_size_and_timeout_python/dataflow.yaml`

  ## Related
  Closes #1422 
  Follow-up to feedback in #1412.

  ## Validation
  - `cargo build` ✅
  - `cargo clippy --workspace --all-targets` ✅

### DIFF
--- a/tests/queue_size_and_timeout_python/send_data.py
+++ b/tests/queue_size_and_timeout_python/send_data.py
@@ -9,16 +9,15 @@ from dora import Node
 def main() -> None:
     """TODO: Add docstring."""
     dora_node = Node()
+    start = time.time()
     i = 0
-    for event in dora_node:
-        event_type = event["type"]
-        if event_type == "STOP":
-            break
-        if event_type != "INPUT":
-            continue
-
+    while time.time() - start < 10:
         dora_node.send_output("ts", pa.array([time.perf_counter_ns(), i]))
         i += 1
+
+        event = dora_node.next(timeout=0.001)
+        if event is not None and event["type"] == "STOP":
+            break
 
 
 if __name__ == "__main__":
--- a/tests/queue_size_and_timeout_python/dataflow.yaml
+++ b/tests/queue_size_and_timeout_python/dataflow.yaml
@@ -1,8 +1,6 @@
 nodes:
   - id: send_data
     path: ./send_data.py
-    inputs:
-      keepalive: dora/timer/millis/1
     outputs:
       - ts
 


tests/queue_size_and_timeout_python/send_data.py
```diff
@@ -5,19 +5,18 @@
...
 def main() -> None:
     """TODO: Add docstring."""
     dora_node = Node()
+    start = time.time()
     i = 0
     while time.time() - start < 10:
         dora_node.send_output("ts", pa.array([time.perf_counter_ns(), i]))
         i += 1
-        # print(f"Sent {i} times", flush=True)
-        time.sleep(0.001)
```
phil-opp, line 19: We should probably not remove this sleep. Otherwise the data is sent at maximum speed.

phil-opp: Scratch that! The node doesn't expect inputs, so we will wait the 1ms in every iteration.

PR #1425, opened 2026-03-17
fix build output draining

### LINKED ISSUE #1419: Bug: build log reader stops when either stdout or stderr reaches EOF
### Summary

`run_build_command` in `libraries/core/src/build/build_command.rs` reads `stdout` and `stderr` in a single `tokio::select!` loop. Right now that loop exits as soon as either stream returns EOF.

That means a build command can stop being drained too early if it closes one file descriptor before it finishes writing to the other. Once the remaining pipe fills up, the child can block, and `dora build` ends up waiting forever on `child.wait()`.

### Why this matters

This is a correctness issue in the core build path, not just a logging problem. The current implementation can turn a valid build command into a hung build if its `stdout` and `stderr` lifetimes do not match.

The relevant code is here:

```rust
let line = tokio::select! {
    line = stdout_lines.next_line() => line,
    line = stderr_lines.next_line() => line,
};
let Some(line) = line.transpose() else {
    break;
};
```

### Reproduction idea

On Unix, a command like the following demonstrates the failure mode:

```bash
sh -c 'exec 2>&-; yes x | head -n 500000'
```

That closes `stderr` immediately but keeps producing a large amount of `stdout`. Once the reader task exits on the `stderr` EOF, `stdout` is no longer drained and the child can stall when the pipe buffer fills.

### Expected behavior

`run_build_command` should keep draining whichever stream is still open and only stop once both `stdout` and `stderr` are exhausted.

### Possible fix

A small fix here would be to track the two streams independently and keep polling the remaining live stream after the other one reaches EOF. A regression test around the reader behavior would make sense as part of the change.

I can work on this if it sounds like a good fit.

Monti-27:
Fixes #1419

## Summary
Keep draining build stdout and stderr independently until both streams are closed.

## Changes
- move build output forwarding into a dedicated helper
- stop treating EOF on one stream as EOF on both
- add regression tests for asymmetric stdout/stderr shutdown
- enable the Tokio I/O utility features needed by the helper and tests

## Verification
- cargo fmt --all --check
- OPENSSL_NO_VENDOR=1 cargo test -p dora-core --features build

### DIFF
--- a/libraries/core/Cargo.toml
+++ b/libraries/core/Cargo.toml
@@ -25,7 +25,7 @@ which = "5.0.0"
 uuid = { version = "1.7", features = ["serde", "v7"] }
 tracing = "0.1"
 serde-with-expand-env = "1.1.0"
-tokio = { version = "1.24.1", features = ["fs", "process", "sync", "rt", "io-util"] }
+tokio = { version = "1.24.1", features = ["fs", "process", "sync", "rt", "io-util", "macros"] }
 tokio-stream = { version = "0.1.11", features = ["io-util"] }
 schemars = "1.0.4"
 serde_json = "1.0.117"
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -95,84 +95,77 @@ mod tests {
     use super::forward_build_output;
     use tokio::io::{AsyncWriteExt, BufReader};
 
-    #[test]
-    fn keeps_draining_stdout_after_stderr_closes() {
-        run_forward_output_test(true);
+    #[tokio::test]
+    async fn keeps_draining_stdout_after_stderr_closes() {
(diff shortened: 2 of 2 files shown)


phil-opp (commented):
Thanks for the PR!

It would be simpler to just use the [`Merge` trait of `futures-concurrency`](https://docs.rs/futures-concurrency/latest/futures_concurrency/stream/trait.Merge.html) or the `merge` method of `tokio_stream::StreamExt`.

Then the code could just look like this:

```rust
            let mut merged = (
                LinesStream::new(child_stdout.lines()),
                LinesStream::new(child_stderr.lines()),
            )
            .merge();

            while let Some(line) = merged.next().await {
                if stdout_tx.send(line).await.is_err() {
                    break;
                }
            }
```

Issue #1441, opened 2026-03-18
RFC: Add `dora.lock` for reproducible git-backed node builds


Bhanudahiyaa:
## Problem
  `dora build` resolves git-backed nodes to commit hashes, but those resolutions are primarily persisted in local session metadata (`out/*.dora-session.yaml`) rather than a source-controlled
  lock artifact.

  ## Why this matters
  - Reproducibility: branch/tag heads can move and cause build drift.
  - Reviewability: resolved commits are not visible in code review.
  - Collaboration: teammates/CI cannot reliably rebuild from repo state alone.

  ## Expected behavior
  A lockfile in the dataflow directory (e.g. `dora.lock`) should store resolved git commit hashes and be reused by future builds when the requested source specification matches.

  ## Proposed scope (prototype)
  - Introduce `dora.lock` read/write in CLI build flow.
  - Match lock entries by `(node_id, repo, requested_rev)`.
  - Reuse lock hit; fetch only lock misses.
  - Write normalized lockfile after build.
  - Add unit tests for lock hit/miss + roundtrip parsing/serialization.

  ## Design question
  Should strict/update modes be part of the first iteration (`--locked`, `--update-lock`), or should we start with automatic behavior and add strict modes in follow up PRs?

heyong4725:
Closing — the feature shipped as part of the dora 1.0 consolidation (`145ccce0`, 2026-04-17).

### What's on main today

A full lockfile implementation at [`binaries/cli/src/command/build/lockfile.rs`](https://github.com/dora-rs/dora/blob/main/binaries/cli/src/command/build/lockfile.rs):

```rust
pub struct BuildLockfile {
    version: u32,
    descriptor_fingerprint: Option<String>,
    pub git_sources: BTreeMap<NodeId, GitSource>,
}
```

User-facing surface (on both `dora build` and `dora run`):

| Flag | Behavior |
|---|---|
| `--locked` | Use pinned commits from `<stem>.dora-lock.yaml`; fail if descriptor diverges from lockfile fingerprint |
| `--write-lockfile` | Resolve git sources and write/refresh the lockfile |
| `--lockfile <PATH>` | Override the default lockfile path |

The two flags are mutually exclusive (parser enforces — see `binaries/cli/src/command/mod.rs:259-260` and `:361-362`).

### Cross-reference against the original "Proposed scope"

| Asked for | Shipped |
|---|---|
[...]

PR #1443, opened 2026-03-18
feat(cli): add dora.lock prototype for git-backed nodes

Bhanudahiyaa:
##Related Issue: #1441 

## Context
  This PR targets reproducibility for git-backed node dependencies in `dora build`.

  Per CONTRIBUTING, this is a non-trivial change and is paired with issue discussion.

  ## Problem
  Git commit resolutions are not currently represented as a first-class, source-controlled lock artifact, which makes deterministic rebuilds and review of dependency resolution harder.

  ## What this solves
  Adds a `dora.lock` prototype so resolved git sources are explicit, shareable, and reusable.

  ## Changes
  - Added CLI lockfile module:
    - reads/writes `dora.lock` next to the dataflow
    - lockfile schema versioning (`version: 1`)
    - stores per-node git source resolution:
      - `repo`
      - `requested_rev` (branch/tag/rev/none)
      - `commit_hash`
  - Integrated into `dora build`:
    - consume matching lock entries first
    - fetch unresolved git refs only for misses
    - write normalized lockfile after build flow
  - Added unit tests:
    - lock hit on matching repo+rev
    - lock miss on mismatched rev
    - lockfile roundtrip read/write

  ## Why this matters
  - Improves reproducibility across local/dev/CI environments.
  - Makes dependency resolution reviewable in PRs.
  - Establishes a foundation for broader package/dependency management work.

  ## Validation
  - [x] `cargo fmt --all`
  - [x] `cargo build -p dora-cli`
  - [x] `cargo test -p dora-cli`
  - [x] `cargo clippy -p dora-cli --all-targets`

### DIFF
--- a/binaries/cli/src/lockfile.rs
+++ b/binaries/cli/src/lockfile.rs
@@ -1,0 +1,209 @@
+use dora_message::{common::GitSource, descriptor::GitRepoRev, id::NodeId};
+use eyre::Context;
+use std::{
+    collections::BTreeMap,
+    path::{Path, PathBuf},
+};
+
+const LOCKFILE_NAME: &str = "dora.lock";
+const LOCKFILE_VERSION: u32 = 1;
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct DoraLock {
+    pub version: u32,
+    #[serde(default)]
+    pub git_sources: BTreeMap<NodeId, LockedGitSource>,
+}
+
+impl DoraLock {
+    pub fn new() -> Self {
+        Self {
+            version: LOCKFILE_VERSION,
+            git_sources: BTreeMap::new(),
+        }
+    }
+
+    pub fn read_for_dataflow(dataflow_path: &Path) -> eyre::Result<Self> {
+        let lockfile_path = lockfile_path(dataflow_path)?;
+        if !lockfile_path.exists() {
+            return Ok(Self::new());
+        }
+
+        let contents = std::fs::read_to_string(&lockfile_path)
+            .with_context(|| format!("failed to read lockfile `{}`", lockfile_path.display()))?;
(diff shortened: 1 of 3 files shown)


binaries/cli/src/lockfile.rs
```diff
@@ -0,0 +1,209 @@
+use dora_message::{common::GitSource, descriptor::GitRepoRev, id::NodeId};
+use eyre::Context;
+use std::{
+    collections::BTreeMap,
+    path::{Path, PathBuf},
+};
+
+const LOCKFILE_NAME: &str = "dora.lock";
```
phil-opp: Maybe we name this after the dataflow instead?

binaries/cli/src/lockfile.rs
```diff
@@ -0,0 +1,209 @@
...
+        Ok(lockfile)
+    }
+
+    pub fn write_for_dataflow(&self, dataflow_path: &Path) -> eyre::Result<()> {
+        let lockfile_path = lockfile_path(dataflow_path)?;
+        let serialized = serde_yaml::to_string(self).context("failed to serialize lockfile")?;
+        std::fs::write(&lockfile_path, serialized)
+            .with_context(|| format!("failed to write lockfile `{}`", lockfile_path.display()))?;
+        Ok(())
+    }
```
phil-opp, line 60: What if there are multiple dataflow YAML files in the same dir? Then this would also overwrite the lockfile of the other dataflow, wouldn't it?

heyong4725:
Hi @Bhanudahiyaa, thanks for engaging constructively with the lockfile-naming feedback.

The branch has been stale for 5+ weeks since the discussion paused, with merge conflicts against current `main`. Closing for housekeeping.

The dora.lock direction is a real design question (scope: CLI-local vs. shared crate, lockfile-per-dataflow vs. project-level). If you would like to ship this, please rebase against current `main` and ping @phil-opp on the linked issue (#1441) before reopening — the design conversation has more leverage than the implementation at this stage.

Issue #1460, opened 2026-03-19
dora run/start --uv can leak to parent Python selection and break reproducibility


Bhanudahiyaa:
# Problem: Python Node Execution with `--uv` Uses Incorrect Interpreter

## Description

When spawning Python nodes with `--uv`, the daemon-side execution (`uv run python ...`) may follow the parent interpreter selection (e.g., a root `.python-version`) instead of the intended active virtual environment.

## Why This Matters

This leads to non-reproducible and inconsistent behavior:

- The same dataflow may work or fail depending on the host machine’s Python configuration.
- Errors such as missing `dora` or dependencies can occur despite a correctly configured virtual environment.
- Expected dependency isolation is violated.

## Expected Behavior

With `--uv`, Python nodes should:

- Execute using the intended active virtual environment, or
- Fail deterministically if the environment is not correctly configured.

## Actual Behavior

`uv` may resolve to an external interpreter due to:

- `.python-version` files
- Ambient Python discovery mechanisms

This causes execution to bypass the intended virtual environment.

heyong4725:
Closing — the bug is fixed on main. Just stayed open because the PR description used informal "Proposed issue: #1460" instead of `Closes #1460` syntax.

### Fix chain

1. **#1461** ("fix(uv): pin python runtime to active venv for daemon-spawned nodes", merged 2026-04-08) — the direct fix. Pinned `uv run --no-project --python <venv-python> --no-managed-python python -u <script>` instead of bare `uv run python -u <script>`. Authored by @Bhanudahiyaa (the issue author too — nice self-fix turnaround).

2. **#1820** ("feat(uv): per-node managed Python venvs at build + runtime", merged later) — architectural follow-up. Replaced the explicit `--python <path>` pinning with per-node `VIRTUAL_ENV` + `PATH` setup. Each Python node now gets its own managed venv at `.dora/python-envs/<node_id>/`, and the daemon spawns with that venv's path already in the env. uv picks the right interpreter naturally because it respects `VIRTUAL_ENV` before any ambient discovery.

### Verification on current main

```rust
// binaries/daemon/src/spawn/spawner.rs:99
cmd.env("VIRTUAL_ENV", python_env_dir);
// + prepend python_env_dir/bin (or Scripts on Windows) to PATH
```

[...]

PR #1461, opened 2026-03-19
fix(uv): pin python runtime to active venv for daemon-spawned nodes

Bhanudahiyaa:
## Problem

  `dora run/start --uv` currently spawns Python nodes via `uv run python ...`, but interpreter selection can drift to parent/global discovery rules (e.g. repo-level `.python-version`) instead of
  the active virtual environment.

  That causes non-reproducible behavior:
  - same dataflow passes/fails depending on host Python state,
  - `ModuleNotFoundError` appears even when dependencies are installed in the active venv,
  - runtime isolation expectations for Python nodes are violated.

  ## Why this matters

  This is a reproducibility and dependency-isolation gap in the execution path (not just build path).
  For Project #2-aligned work, runtime determinism is required before introducing higher-level packaging/install workflows.

  ## What changed

  ### 1) Pin UV interpreter to active venv for Python node spawn
  File: `binaries/daemon/src/spawn/command.rs`

  - Added helper to resolve `VIRTUAL_ENV` Python path:
    - Unix: `$VIRTUAL_ENV/bin/python`
    - Windows: `$VIRTUAL_ENV/Scripts/python.exe`
  - Updated Python spawn command under `--uv`:
    - from: `uv run python -u <script>`
    - to: `uv run --no-project --python <venv-python> --no-managed-python python -u <script>`
  - Fallback behavior:
    - if no valid `VIRTUAL_ENV` Python is available, use `--active` path.

  ### 2) Apply same behavior for Python runtime operator spawn
  File: `binaries/daemon/src/spawn/spawner.rs`

  - Updated runtime launch path (`import dora; dora.start_runtime()`) to use the same pinned-venv logic under `--uv`.

  ### 3) Add regression tests
  File: `binaries/daemon/src/spawn/command.rs`

  - Added unit tests for interpreter resolution:
    - resolves Python path from virtual env layout,
    - returns `None` when venv Python path does not exist.

  ### 4) Add CI regression coverage
  File: `.github/workflows/ci.yml`

  - In `Test CLI (Python): Async Example`, inject conflicting `.python-version` before run:
    - verifies daemon-side `--uv` spawn does not leak to parent Python selection.

  ## Design notes / tradeoffs

  - This is intentionally narrow: it hardens daemon Python spawn behavior without introducing a new package-manager surface.
  - We prefer deterministic interpreter selection (`--python <venv>`) over ambient resolution.
  - `--no-project` avoids inheriting unrelated parent project settings; fallback still exists when `VIRTUAL_ENV` is unavailable.

  ## Validation

  - `cargo fmt --all`
  - `cargo test -p dora-daemon spawn::command::tests -- --nocapture`
  - `cargo check -p dora-daemon`
  - Local end-to-end repro with conflicting `.python-version`:
    - `cargo run -p dora-cli -- run examples/python-async/dataflow.yaml --uv --stop-after 30s` ✅

  ## Related

  - Proposed issue: #1460 `dora run/start --uv can leak to parent Python selection and break reproducibility`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -492,7 +492,11 @@ jobs:
         if: runner.os != 'Windows'
         env:
           RUST_LOG: debug
-        run: dora run examples/python-async/dataflow.yaml --uv --stop-after 5m
+        run: |
+          # Ensure uv execution does not leak to parent `.python-version` / global interpreters.
+          echo "3.11.14" > .python-version
+          trap 'rm -f .python-version' EXIT
+          dora run examples/python-async/dataflow.yaml --uv --stop-after 5m
       - name: "Test Python Example: Drain"
         timeout-minutes: 30
         env:
--- a/binaries/daemon/src/spawn/command.rs
+++ b/binaries/daemon/src/spawn/command.rs
@@ -7,7 +7,26 @@ use dora_core::{
 use dora_download::download_file;
 use dora_message::common::LogLevel;
 use eyre::WrapErr;
-use std::path::Path;
+use std::path::{Path, PathBuf};
+
+pub(super) fn uv_python_interpreter_from_env() -> Option<PathBuf> {
+    uv_python_interpreter_from_virtual_env(std::env::var_os("VIRTUAL_ENV").map(PathBuf::from))
+}
+
+fn uv_python_interpreter_from_virtual_env(virtual_env: Option<PathBuf>) -> Option<PathBuf> {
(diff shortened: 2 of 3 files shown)


Bhanudahiyaa:
Thanks for the review and approval.

I’ll add documentation for the `--uv` interpreter selection/fallback behavior (prefer `VIRTUAL_ENV` python, fallback to `--active`) in the docs repo at:
  `docs/guides/Development/virtual-environment.md`.

PR #1453, opened 2026-03-19
feat(cli): add local registry index reader and inspect command

Bhanudahiyaa:
## Issue Related: #1452 
 
 ## Problem 
  Project #2 (Package and Dependency Management) needs an incremental registry foundation, but Dora currently has no local registry index read/resolution path to validate package lookup
  behavior.

  ## Why this matters
  This PR establishes a self-contained, read-only registry prototype that enables reproducible dependency resolution semantics without introducing publish/write complexity yet.

  ## What changed
  - Added new registry module:
    - `RegistryIndex` schema (TOML)
    - strict index version check (`version = 1`)
    - package/dependency validation (name rules, semver parsing, non-empty entrypoint)
    - resolver API selecting latest version matching optional semver requirement
  - Added CLI command:
    - `dora node inspect-registry`
    - supports registry path as directory (`index.toml`) or explicit file
    - supports list-all mode and resolve-by-name mode
    - optional `--requirement` semver filter (validated)
    - output in `table` or `json`
  - Added example:
    - `examples/registry/index.toml`
  - Added tests for:
    - resolver behavior
    - invalid metadata validation
    - command argument guardrails
    - source path handling for file vs directory

  ## Design choices / tradeoffs
  - Read-only scope only (no publish/write) to keep PR independent and low-risk.
  - Deterministic source resolution:
    - `*.toml` => explicit index file
    - otherwise => `<path>/index.toml`
    This avoids relying on filesystem pre-existence for path interpretation.
  - Kept schema minimal and future-extensible for install/publish follow-ups.

  ## Validation
  - [x] `cargo fmt --all`
  - [x] `cargo build -p dora-cli`
  - [x] `cargo test -p dora-cli`
  - [x] `cargo clippy -p dora-cli --all-targets`

  (Existing unrelated workspace warnings are unchanged.)

Issue #1402, opened 2026-03-14
RFC: DORA Node Package Ecosystem (Manifest + Registry + Package Manager)


Harsh-Sahu43:
Hi @phil-opp @haixuanTao,

I’m preparing a **GSoC 2026 proposal** around improving package and dependency management in the DORA ecosystem.

While exploring the repositories and the GSoC project description, I noticed that the current **`node-hub` in `dora-hub`** primarily works as a collection of node folders. This works well for examples, but may become difficult to scale for:

- dependency resolution
- versioning
- reproducible builds
- node discoverability

The GSoC project description also highlights the need for a proper package and dependency management system:

[Project #2: Package & Dependency Management](https://github.com/dora-rs/dora/wiki/GSoC_2026)

Inspired by **Cargo + crates.io**, I started exploring a possible design for a scalable node packaging ecosystem for DORA.

The idea is to introduce:

- a **node package specification** (`Dora.toml`)
- a **dependency resolver and lockfile**
- a **node registry** for publishing and discovery
- CLI commands such as `dora install`, `dora publish`, and `dora build`

A possible high-level architecture could look like this:

            dora
            │
            ├ binaries
            │   ├ cli
            │   ├ coordinator
            │   ├ daemon
            │   └ runtime
            │
            ├ libraries
            │
            └ package-manager
                 ├ manifest
                 ├ resolver
                 └ installer

         dora-node-registry
          │
          ├ registry-server
          │   ├ api
          │   ├ auth
          │   └ publish
          │
          ├ registry-index
          │   └ packages metadata
          │
          └ packages
              └ tarballs
      
        dora-hub
           │
           └ node-hub
             ├ dora-yolo
             ├ dora-reachy
             └ dora-rerun

      
Conceptually, the workflow could look like:

                ┌──────────────────────┐
                │      DORA CLI        │
                │   (dora install)    │
                └─────────┬────────────┘
                          │
                          ▼
                ┌──────────────────────┐
                │   Package Manager    │
                │   (inside dora)      │
                │                      │
                │ manifest parser      │
                │ dependency resolver  │
                │ installer            │
                └─────────┬────────────┘
                          │
                          ▼
                ┌──────────────────────┐
                │   Registry Client    │
                └─────────┬────────────┘
                          │
                          ▼
        ┌─────────────────────────────────────┐
        │         dora-node-registry          │
        │                                     │
        │ registry-server  (API)              │
        │ registry-index   (metadata index)   │
[...]

PR #1467, opened 2026-03-19
refactor(cli): introduce shared package metadata model

Harsh-Sahu43:
## Problem

`Dora.toml` metadata and registry metadata describe the same package semantics but are evolving independently.

If these paths define and validate package data independently, they can drift quickly as Project [#2](https://github.com/dora-rs/dora/wiki/GSoC_2026) grows. That would make later resolver, install, publish, and lockfile work harder to align.

## Related context
This PR is aligned with the same package-management direction being explored in:
- [#1451](https://github.com/dora-rs/dora/pull/1451) — Dora.toml metadata validation  
- [#1453](https://github.com/dora-rs/dora/pull/1453) — local registry index reader / inspect flow

## What this PR changes

This PR introduces a shared package metadata model inside the CLI crate and makes both manifest and registry parsing normalize into it. This PR is intentionally scoped to the shared semantic layer and does not introduce new CLI surface.

### Added
- a shared validated package model in `binaries/cli/src/package_metadata.rs`
- shared helpers for:
  - package/dependency name validation
  - semver parsing
  - entrypoint validation

### Refactored
- `binaries/cli/src/dora_toml.rs`
  - keeps raw `Dora.toml` parsing separate
  - normalizes manifest metadata into the shared package model

- `binaries/cli/src/registry.rs`
  - keeps raw registry index parsing separate
  - normalizes registry package records into the same shared package model

## Why this matters

This establishes a clean architectural foundation:
- Raw structs → file-oriented  
- Shared model → semantic-oriented  
- later layers like resolver, lockfile, install, and publish can build on one package model instead of multiple ad hoc ones

This prevents schema drift and ensures that future layers (resolver, lockfile, install, publish) operate on a single consistent package representation.

## Design

Inspired by Cargo’s separation of concerns:
- parsing remains format-specific  
- validation is centralized  
- normalized package model is shared  

Resolver and higher-level workflows are intentionally deferred.

## Validation

```bash
cargo fmt --all --check
cargo check -p dora-cli
cargo test -p dora-cli package_metadata
cargo test -p dora-cli dora_toml
cargo test -p dora-cli registry
cargo clippy -p dora-cli --all-targets

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1576,6 +1576,7 @@ dependencies = [
  "termcolor",
  "tokio",
  "tokio-stream",
+ "toml",
  "tracing",
  "tracing-log",
  "uuid",
@@ -4672,7 +4673,7 @@ version = "3.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e67ba7e9b2b56446f1d419b1d807906278ffa1a658a8a5d8a39dcb1f5a78614f"
 dependencies = [
- "toml_edit",
+ "toml_edit 0.25.5+spec-1.1.0",
 ]
 
 [[package]]
@@ -5828,6 +5829,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_spanned"
+version = "0.6.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf41e0cfaf7226dca15e8197172c295a782857fcb97fad1808a166870dee75a3"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "serde_urlencoded"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


heyong4725:
Hi @Harsh-Sahu43, thanks for putting this together — the code itself is clean (PackageIdentity / PackageDefinition / DependencyRequirement is a reasonable shape, and the validator helpers are well-factored).

Closing this one as part of the broader call on the package-management track. Where the project is right now:

- **Package management is a long-term roadmap item, not a near-term commitment.** The final solution shape isn't clear yet, and we're not confident the community-visible value is there to justify the upfront design choices a package manager forces.
- **Several adjacent PRs in this track were already closed** (#1451 Dora.toml validation, #1453 registry index reader). Without those merged consumers, the shared semantic layer here would land as ~569 lines of orphaned code with no callers — and the abstraction would harden a direction the project hasn't actually picked yet.
- **RFC #1402 will get the same treatment.** When we revisit package management, we'd rather start from fresh thinking against whatever ecosystem reality looks like at that point, instead of carrying half-resolved foundation code forward.

This is **not a quality judgment on the work** — closing it on direction, not merit. Your other contributions are exactly the kind we want:

- **#1818** (rescue of #1352, `dora doctor` /dev/shm permission check) merged 2026-05-14
- **#1820** (rescue of #1515, per-node managed Python venvs) merged 2026-05-15

Both landed cleanly because they tackled concrete pain the project already had. If you're looking for the next thing to work on, the patterns that rescue well are: defensive fixes for crashes / panics, gaps in cross-platform CI, missing CLI ergonomics, and edge cases in existing features. The recent rescue stack (#1855, #1859, #1861, #1864, #1865, #1869, #1870) is a good shape to mirror.

Thanks again — please don't read this close as "stop contributing." Just "the foundation isn't laid here yet."

Issue #1474, opened 2026-03-21
fix(cli): session file corruption is silently regenerated and write path is not crash-safe


Bhanudahiyaa:
## Problem

  `DataflowSession` handling in the CLI currently has two integrity problems:

  1. Session writes are non-atomic (direct write to final path).
  2. If an existing session file is invalid/corrupted, `read_session` silently regenerates a new default session.

  This can reset session identity/build context without explicit failure.

  ## Why this matters

  Session state is part of reproducibility and build/run continuity.
  If corruption is silently replaced, operators lose deterministic behavior and may run with an unintended session/build context.

  This is especially risky for Project #2 goals around dependency/source reproducibility and stable build metadata.

  ## Expected behavior

  - Session file updates should be crash-safe (atomic write).
  - Corrupted/truncated existing session files should produce an explicit error with recovery guidance, not silent regeneration.

  ## Actual behavior

  - Session file is written directly to target path.
  - Parse failures are logged as warning and replaced by a fresh default session.

  ## Proposed fix

  - Implement atomic session writes using temporary file + fsync + rename.
  - Change `read_session` behavior:
    - If file exists and parse fails, return an explicit error.
    - Include clear recovery guidance (fix/remove file and rerun build).
  - Add regression tests for invalid/truncated existing session files.

PR #1475, opened 2026-03-21
Use file locks for session file

Bhanudahiyaa:
Closes: #1474 

Use file locks to avoid corruption of the session file on concurrent build operations.

### DIFF
--- a/.github/workflows/pip-release.yml
+++ b/.github/workflows/pip-release.yml
@@ -77,7 +77,15 @@ jobs:
           args: --release --out dist --zig
           manylinux: manylinux_2_28
           working-directory: ${{ matrix.repository.path }}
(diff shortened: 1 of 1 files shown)


binaries/cli/src/session.rs
```diff
@@ -30,13 +35,13 @@ impl DataflowSession {
...
-                tracing::warn!(
-                    "failed to read dataflow session file, regenerating (you might need to run `dora build` again)"
-                );
-            }
+            return deserialize(&session_file).wrap_err_with(|| {
+                format!(
+                    "failed to parse existing dataflow session at `{}`; \
+remove or fix the file and run `dora build` to regenerate it",
+                    session_file.display()
+                )
```
phil-opp: `dora build` invokes this function, so it will just throw an error again instead of regenerating the file.

Users are probably not aware of this file, so forcing them to manually delete it is probably more friction than just replacing it, no? The end result is the same: a new session is generated.

binaries/cli/src/session.rs
```diff
@@ -87,6 +92,69 @@ fn deserialize(session_file: &Path) -> eyre::Result<DataflowSession> {
...
+                        path.display()
+                    )
+                })?;
+                std::fs::rename(&tmp_path, path).with_context(|| {
+                    format!(
+                        "failed to move temporary file `{}` to `{}`",
+                        tmp_path.display(),
+                        path.display()
+                    )
+                })?;
```
phil-opp: Unfortunately, this approach has a few issues:

- if you run a `dora build` in parallel between the `remove_file` and `rename`, a new session will be generated, which races with the rename operation.
- on Windows, it's not possible to rename a file that is currently opened by some process. So the rename will fail if there is a parallel operation on that file

phil-opp:
I updated the PR title and description

PR #1477, opened 2026-03-21
fix(coordinator): reconcile running dataflows on daemon disconnect

### LINKED ISSUE #1476: fix(coordinator): daemon disconnect leaves stale running_dataflows daemon membership
## Problem

  When a daemon disconnects (watchdog timeout or explicit daemon exit), the coordinator removes the daemon connection, but `running_dataflows` can still retain that daemon in per-dataflow
  membership.

  As a result, control-plane state can remain stale even though the daemon is gone.

  ## Why this matters

  This is a daemon ↔ coordinator consistency issue that impacts distributed runtime correctness:

  - dataflows may remain marked as running on disconnected daemons
  - stop/check/info behavior can become inconsistent with real runtime state
  - recovery/failover logic has degraded observability and correctness

  This aligns with Project #3 goals around robust distributed state and failure handling.

  ## Expected behavior

  On daemon disconnect, coordinator should reconcile all affected running dataflows:

  - remove daemon from `running_dataflows.daemons`
  - remove from relevant pending sets
  - if no daemons remain for a dataflow, finalize/archive it deterministically

  ## Actual behavior

  Coordinator disconnect handling removes only the daemon connection, leaving dataflow daemon membership stale.

  ## Proposed fix

  - Add centralized daemon-disconnect reconciliation in coordinator core.
  - Invoke it from:
    - watchdog disconnect path
    - `daemon_exit` notification path
  - Add regression test proving dataflow finalization when last daemon disconnects.

Bhanudahiyaa:
Closes: #1476

  ## Problem

  When a daemon disconnects/crashes, coordinator state can become inconsistent with runtime reality:

  - disconnected daemon membership may remain in `running_dataflows`
  - pending sets can become stale (`pending_daemons`, `pending_spawn_results`)
  - failure impact on nodes running on the disconnected machine is not propagated to remaining daemons

  ## Why It Matters

  A daemon disconnect is a failure scenario, not a normal completion path.
  Coordinator and daemon behavior should converge on a consistent failure model without special stop simulation logic.

  ## What This PR Changes

  ### 1) Reconcile coordinator membership state on daemon disconnect

  In `handle_daemon_disconnect(...)`, for each affected running dataflow:

  - remove disconnected daemon from `daemons`
  - remove disconnected daemon from `pending_daemons`
  - remove disconnected daemon from `pending_spawn_results`
  - mark spawn waiters as failed for the affected dataflow (`spawn_result` error)

  ### 2) Record failed nodes for disconnected daemon in coordinator results

  For nodes assigned to the disconnected daemon (`node_to_daemon` mapping), coordinator records `NodeError` results in `dataflow_results` so final dataflow result reflects the failure.

  ### 3) New coordinator → daemon notification RPC for disconnect impact

  Added new daemon control RPC:

  - `daemon_disconnected(dataflow_id, disconnected_daemon_id, failed_nodes)`

  Coordinator calls this on remaining daemons of the affected dataflow.

  ### 4) Daemon-side failure propagation via existing local NodeFailed path

  On receiving `daemon_disconnected(...)`, daemon:

  - maps `failed_nodes` to local affected receivers
  - emits local `NodeFailed` propagation using existing receiver-notification logic

  This follows the maintainer-requested flow: treat disconnected-daemon nodes as failed and propagate via daemon/node failure semantics (instead of coordinator-side stop simulation).

  ### 5) Shared finalization logic, no duplicated finish path logic

  Introduced shared `finalize_dataflow(...)` helper in coordinator and reused it where needed (`all_nodes_finished` and disconnect path when no daemons remain), reducing duplication.

  ### 6) Removed previous disconnect test

  Per review feedback, removed the prior test that validated the old behavior.

  ## Files Changed

  - `binaries/coordinator/src/lib.rs`
  - `binaries/coordinator/src/listener.rs`
  - `binaries/daemon/src/coordinator.rs`
  - `binaries/daemon/src/lib.rs`
  - `libraries/message/src/coordinator_to_daemon.rs`

  ## Validation

  - [x] `cargo fmt --all`
  - [x] `cargo check -p dora-message -p dora-coordinator -p dora-daemon`
  - [x] `cargo test -p dora-coordinator --lib -- --nocapture`
  - [x] `cargo test -p dora-message --lib -- --nocapture`
  - [x] `cargo test -p dora-daemon --lib -- --nocapture`

  No unrelated functional changes were made.

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -15,7 +15,7 @@ use dora_message::{
         CoordinatorControlResponse,
     },
     common::{DaemonId, NodeError, NodeErrorCause, NodeExitStatus},
-    coordinator_to_cli::{DataflowResult, LogMessage, StopDataflowReply},
+    coordinator_to_cli::{DataflowResult, StopDataflowReply},
     coordinator_to_daemon::{
         BuildDataflowNodes, DaemonControlClient, DaemonControlRequest, DaemonControlResponse,
         RegisterResult, Timestamped,
@@ -524,21 +524,6 @@ async fn start_inner(
     tracing::info!("stopped");
 
     Ok(())
-}
-
-pub(crate) async fn send_log_message(
-    log_subscribers: &mut Vec<LogSubscriber>,
-    message: &LogMessage,
-) {
-    for subscriber in log_subscribers.iter_mut() {
-        let send_result =
-            tokio::time::timeout(Duration::from_millis(100), subscriber.send_message(message));
-
-        if send_result.await.is_err() {
-            subscriber.close();
-        }
-    }
-    log_subscribers.retain(|s| !s.is_closed());
 }
 
 pub(crate) async fn handle_daemon_disconnect(
@@ -1309,8 +1294,6 @@ mod tests {
                 node_metrics: BTreeMap::new(),
                 spawn_result: CachedResult::default(),
                 stop_reply_senders: vec![stop_tx],
-                buffered_log_messages: Vec::new(),
-                log_subscribers: Vec::new(),
                 pending_spawn_results: [daemon_id.clone()].into_iter().collect(),
             },
         );
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -14,8 +14,8 @@ use dora_message::{
         BuildRequest, CoordinatorControl, CoordinatorControlClient, CoordinatorControlRequest,
         CoordinatorControlResponse,
     },
-    common::DaemonId,
-    coordinator_to_cli::{DataflowResult, StopDataflowReply},
+    common::{DaemonId, NodeError, NodeErrorCause, NodeExitStatus},
+    coordinator_to_cli::{DataflowResult, LogMessage, StopDataflowReply},
     coordinator_to_daemon::{
         BuildDataflowNodes, DaemonControlClient, DaemonControlRequest, DaemonControlResponse,
         RegisterResult, Timestamped,
@@ -501,7 +501,7 @@ async fn start_inner(
                     tracing::error!("Disconnecting daemons that failed watchdog: {disconnected:?}");
                     for machine_id in disconnected {
                         coordinator_state.daemon_connections.remove(&machine_id);
-                        handle_daemon_disconnect(&coordinator_state, &machine_id);
+                        handle_daemon_disconnect(&coordinator_state, &machine_id).await;
                     }
                 }
             }
@@ -541,11 +541,11 @@ pub(crate) async fn send_log_message(
     log_subscribers.retain(|s| !s.is_closed());
 }
 
-pub(crate) fn handle_daemon_disconnect(
+pub(crate) async fn handle_daemon_disconnect(
     coordinator_state: &state::CoordinatorState,
     daemon_id: &DaemonId,
 ) {
-    let mut dataflows_to_finalize = Vec::new();
+    let mut affected_dataflows = Vec::new();
     for mut dataflow in coordinator_state.running_dataflows.iter_mut() {
         let removed = dataflow.daemons.remove(daemon_id);
         if !removed {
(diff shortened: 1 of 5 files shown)


phil-opp (changes requested):
Thanks for the PR!

This seems to treat crashed daemons like normal finished nodes, which is not what we should do.

Instead, if a daemon crashes/disconnects, we should treat this as an error. All active dataflows that use this daemon should be treated as failed, similar to the case when a node fails to build on one machine.

Bhanudahiyaa:
Thanks, this makes sense. You’re right.

My current implementation reconciles daemon membership, but when the last daemon drops it finalizes with an empty/success like result, which is the wrong semantic for disconnect/crash.

I’ll rework this PR so daemon disconnect is treated as failure for all affected running dataflows.

  Planned changes:
  - keep membership reconciliation (`daemons`, pending sets),
  - mark affected dataflows as failed (not normal finished),
  - resolve pending `wait_for_spawn`/stop waiters with an explicit disconnect failure,
  - preserve healthy-path behavior.

  I’ll push an update.

binaries/coordinator/src/lib.rs
```diff
@@ -525,6 +526,116 @@ async fn start_inner(
...
+                .dataflow_results
+                .get(&dataflow_id)
+                .map(|r| dataflow_result(r.value(), dataflow_id, &coordinator_state.clock))
+                .unwrap_or_else(|| {
+                    DataflowResult::ok_empty(dataflow_id, coordinator_state.clock.new_timestamp())
+                }),
+        };
+
+        for sender in finished_dataflow.stop_reply_senders.drain(..) {
+            let _ = sender.send(Ok(reply.clone()));
```
phil-opp: Why do you send a stop reply here?

phil-opp (commented):
The test is not really useful in its current form, please remove it.

This seems to duplicate a lot of code from the `all_nodes_finished` API. Why?

I think the proper process should be to treat a daemon disconnect like a failure of all nodes that are running on that machine. The coordinator could notify all other daemons about the disconnected daemon through a new RPC method. The receiving daemons could then create and send the corresponding NodeFailed events to their local nodes. This way, we don't need to introduce any special logic, such as simulating a stop command.

Bhanudahiyaa:
@phil-opp Thanks for the detailed review, I reworked this PR to follow the process you suggested.

  What changed:

  1. Removed stop-command simulation from disconnect handling
  - `handle_daemon_disconnect` no longer calls `stop_dataflow` on remaining daemons.

  2. Removed unconditional coordinator-side finalization on every affected dataflow
  - A disconnect now reconciles membership/pending sets and keeps the dataflow running when other daemons remain.
  - Finalization is only triggered when no daemons remain.

  3. Added coordinator -> daemon disconnect notification RPC
  - Added `daemon_disconnected(dataflow_id, disconnected_daemon_id, failed_nodes)` to `DaemonControl`.

  4. Daemon now handles this notification by generating local failure propagation
  - Receiving daemons convert disconnected-daemon node failures into local `NodeFailed` propagation for affected receivers (reusing existing local propagation logic).

  5. Removed the previous disconnect test
  - Per your request, I removed the old test.

  6. Reduced duplication with `all_nodes_finished`
  - Introduced a shared `finalize_dataflow(...)` helper and use it from the normal finish path and the disconnect path when daemon set becomes empty.

  So this no longer relies on special stop simulation logic, and instead uses daemon-side node-failure propagation as discussed.

heyong4725:
Hi @Bhanudahiyaa, per @phil-opp's review, the approach taken here does not match the intended direction:

> if a daemon crashes/disconnects, we should treat this as an error. All active dataflows that use this daemon should be treated as failed, similar to the case when a node fails to build on one machine.
> The coordinator could notify all other daemons about the disconnected daemon through a new RPC method.

Closing without prejudice. The design direction phil-opp described is worth picking up in a fresh PR if you would like to retry — the existing approach (treating crashed daemons like completed ones) loses the failure semantics that downstream nodes need.

Issue #1495, opened 2026-03-22
Limit implicit local build fallback to unavailable coordinator cases


Harsh-Sahu43:
## Context

`dora build` currently supports an implicit fallback to local build when a distributed build is not used through an explicitly configured coordinator.

That fallback is useful when no local coordinator is running, because it lets `dora build` continue in a simple local-only setup.

## Problem

Today, the implicit autodetect path falls back to local build for any coordinator connection error.

That treats these cases the same:

- no local coordinator is running
- coordinator is incompatible
- version check fails
- other connection-layer errors occur

For distributed dataflows, local build is not a safe substitute in all of those cases.

## Why This Matters

If Dora silently switches to local build when the coordinator exists but is incompatible or otherwise broken, it can hide the real problem and change the intended build mode.

A local fallback only clearly makes sense when no local coordinator is reachable.

## Proposed Scope

Narrow the fallback behavior in the implicit `dora build` autodetect path:

- keep local fallback when the local coordinator is unavailable
- surface other coordinator errors instead of masking them with local build

## Out of Scope

- retry logic for coordinator RPCs
- changes to `dora start`
- later-stage distributed build failures after a successful connection
- broader coordinator/daemon recovery behavior

Harsh-Sahu43:
@dora-bot assign me

PR #1496, opened 2026-03-22
fix(cli): only fallback to local build when coordinator is unavailable

Harsh-Sahu43:
Closes: #1495

## Summary

Restrict `dora build`'s implicit local fallback so it only happens when no local coordinator is reachable.

## Problem

The implicit autodetect path previously fell back to local build for any coordinator connection error.

That meant Dora treated these cases the same:

- no local coordinator is running
- coordinator is incompatible
- version check fails
- other connection-layer errors occur

## Change

This PR narrows that behavior:

- keep local fallback for unavailable local coordinator cases
- surface other coordinator errors instead of masking them with local build

Explicit coordinator usage is unchanged.

## Validation

```bash
cargo fmt --all --check
cargo test -p dora-cli detects_connection_refused_as_unavailable_coordinator -- --nocapture
cargo check -p dora-cli

### DIFF
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -58,8 +58,7 @@ use std::{collections::BTreeMap, net::IpAddr};
 use super::{Executable, default_tracing};
 use crate::{
     common::{
-        connect_and_check_version, is_coordinator_unavailable_error, local_working_dir,
-        resolve_dataflow,
+        ConnectAndCheckVersionError, connect_and_check_version, local_working_dir, resolve_dataflow,
     },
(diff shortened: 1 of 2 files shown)


Harsh-Sahu43:
@phil-opp 

Apologies for the delayed response.

I’ve now addressed the feedback regarding error handling and refreshed the branch on current `main`. Instead of relying on downcasting and strict `ConnectionRefused` matching, `connect_and_check_version` now returns a structured error enum that distinguishes between connection failures and version check failures.

Harsh-Sahu43:
Thanks for the review and merge!

Issue #1497, opened 2026-03-22
Ignore blank lines in multi-line build commands


Harsh-Sahu43:
## Context

Dora’s schema/docs explicitly show `build: |`, which suggests multi-line build blocks are an intended part of the build model.

Each build line is executed as a separate command.

## Problem

Right now, blank lines inside a multi-line build block are treated as empty commands.

That means a build like this fails even though the actual command lines are valid:

```yaml
nodes:
  - id: blank-line-repro
    build: |
      cargo --version

      cargo --version
    path: shell
    args: echo hi


How to reproduce : 

mkdir -p tmp-blank-build-repro

cat > tmp-blank-build-repro/dataflow.yml <<'EOF'
nodes:
  - id: blank-line-repro
    build: |
      cargo --version

      cargo --version
    path: shell
    args: echo hi
EOF

cargo run -p dora-cli -- build tmp-blank-build-repro/dataflow.yml --local

Result : 

<img width="999" height="271" alt="Image" src="https://github.com/user-attachments/assets/29fe0395-ddac-4319-8c68-d5eb584854f4" />

PR #1498, opened 2026-03-22
fix(core): ignore blank lines in multi-line build commands

Harsh-Sahu43:
Closes : #1497

## Context

Dora’s schema and documentation explicitly support multi-line build commands using `build: |`, where each line is executed as a separate command.

## Problem

Currently, blank lines inside a multi-line `build` block are treated as empty commands.

This causes builds to fail even when all actual command lines are valid.

Example:

```yaml
nodes:
  - id: blank-line-repro
    build: |
      cargo --version

      cargo --version
    path: shell
    args: echo hi
```

Blank lines inside a multi-line build block should be ignored. Only a truly empty build command should fail.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,6 +168,3 @@ path = "examples/cmake-dataflow/run.rs"
 [profile.dist]
 inherits = "release"
 lto = "thin"
-
-[dependencies]
-tempfile = "3.23.0"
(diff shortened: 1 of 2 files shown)


libraries/core/src/build/build_command.rs
```diff
@@ -168,4 +178,48 @@ mod tests {
         assert_eq!(lines.first().map(String::as_str), Some("line-0"));
         assert_eq!(lines.last().map(String::as_str), Some("line-255"));
     }
+
+    #[tokio::test]
+    async fn ignores_blank_lines_in_multi_line_build_commands() {
+        let working_dir = tempdir().unwrap();
```
swar09, line 217: you can use `std::env::temp_dir()`

Harsh-Sahu43: thanks !, 
I think tempfile::tempdir() is usually better for tests because, it gives a unique temp directory and auto-cleans up

Issue #1499, opened 2026-03-22
Return pipe capture failures as errors instead of panicking


Harsh-Sahu43:
## Context

Dora captures stdout/stderr from spawned child processes in both the build command path and the daemon node spawn path.

Both code paths already use `eyre::Result` for error propagation.

## Problem

Right now, these paths still use `expect(...)` when taking stdout/stderr pipes from child processes.
That means an unexpected missing pipe causes a panic instead of returning a normal error with context.

Affected places:

- `libraries/core/src/build/build_command.rs`
- `binaries/daemon/src/spawn/prepared.rs`

## Why This Matters

A missing captured pipe should be handled like other process-spawn failures in Dora.

Returning a normal error is safer and more consistent than panicking, especially in build and daemon execution paths.

## Proposed Scope

Replace the `expect(...)` pipe capture calls with contextual error returns in the affected build/spawn paths.

For the build path, add focused tests covering missing stdout/stderr pipe capture.

## Out of Scope

- redesigning how child output is forwarded
- changing normal build or node spawn behavior
- broader process lifecycle cleanup or retry behavior

PR #1500, opened 2026-03-22
fix(build): return pipe capture errors instead of panicking

Harsh-Sahu43:
Closes : #1499

## Summary

Return normal errors when build or spawned-node stdout/stderr pipes cannot be captured, instead of panicking.

## Problem

Both the build command path and the daemon spawned-node path currently use `expect(...)` when taking stdout/stderr pipes from child processes.

That means an unexpected missing pipe causes a panic, even though both paths already return `eyre::Result` and otherwise use normal error propagation.

## Change

This PR replaces those `expect(...)` calls with contextual errors in:

- `libraries/core/src/build/build_command.rs`
- `binaries/daemon/src/spawn/prepared.rs`

On the build side, it also adds focused tests covering missing stdout and missing stderr pipe capture.

## Validation

```bash
cargo fmt --all --check
cargo test -p dora-core --features build errors_when_stdout_pipe_is_missing -- --nocapture
cargo test -p dora-core --features build errors_when_stderr_pipe_is_missing -- --nocapture
cargo check -p dora-daemon

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -298,7 +298,10 @@ impl PreparedNode {
         .await
         .expect("Failed to create log file");
         let mut child_stdout =
-            tokio::io::BufReader::new(child.stdout().take().expect("failed to take stdout"));
+            tokio::io::BufReader::new(child.stdout().take().context(format!(
+                "failed to capture stdout pipe from spawned node `{}`",
+                self.node.id
+            ))?);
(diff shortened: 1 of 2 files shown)


Harsh-Sahu43:
@phil-opp 

Thanks!, given we explicitly pipe and immediately take stdout/stderr , I don’t have a concrete scenario where they would be None, so I’m okay reverting this if we want to keep the invariant explicit.

phil-opp:
Ok, let's keep the panic then.

PR #1514, opened 2026-03-22
feat(cli): generate CMakeLists.txt for standalone C/C++ nodes

swar09:
## Summary
Generate a cross-platform CMakeLists.txt when scaffolding standalone C and C++ nodes via dora new --kind node.

## Problem
Running dora new --kind node --lang c or --lang cxx produced only the source files but no build file. There was a // TODO: Makefile? left in both template modules. Users had to write their own build setup manually.

## Solution
Added a per-node CMakeLists.txt template for both C and C++ nodes. Uses CMake (not Makefile) to stay consistent with the existing dataflow templates. The templates are cross-platform — they use cmake -E commands instead of Unix-only cp/mkdir -p, and conditionally link platform-specific libraries (Unix: m z, Windows: ws2_32 userenv ntdll bcrypt).

## Testing
- [x] cargo build -p dora-cli compiles without errors
- [x] cargo clippy -p dora-cli --no-deps produces no new warnings
- [x] Scaffolding a standalone C/C++ node now outputs CMakeLists.txt alongside source files
- [x] Dataflow scaffolding is unaffected

## Changes
New:  `binaries/cli/src/template/c/node-cmake-template.txt`
New: `binaries/cli/src/template/cxx/node-cmake-template.txt`
Modified: `binaries/cli/src/template/c/mod.rs`  added create_node_cmakefile() , removed TODO
Modified: `binaries/cli/src/template/cxx/mod.rs` added create_node_cmakefile() , removed TODO

### DIFF
--- a/binaries/cli/src/template/c/node-cmake-template.txt
+++ b/binaries/cli/src/template/c/node-cmake-template.txt
@@ -1,0 +1,68 @@
+cmake_minimum_required(VERSION 3.21)
+project(___name___ LANGUAGES C)
+
+set(DORA_ROOT_DIR "__DORA_PATH__" CACHE FILEPATH "Path to the root of dora")
+
+set(dora_c_include_dir "${CMAKE_CURRENT_BINARY_DIR}/include/c")
+
+if(DORA_ROOT_DIR)
+    include(ExternalProject)
+    ExternalProject_Add(Dora
+        SOURCE_DIR ${DORA_ROOT_DIR}
+        BUILD_IN_SOURCE True
+        CONFIGURE_COMMAND ""
+        BUILD_COMMAND
+            cargo build
+            --package dora-node-api-c
+        INSTALL_COMMAND ""
+    )
+
+    add_custom_command(OUTPUT ${dora_c_include_dir}
+        WORKING_DIRECTORY ${DORA_ROOT_DIR}
+        DEPENDS Dora
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${dora_c_include_dir}
+        COMMAND ${CMAKE_COMMAND} -E copy_directory apis/c/node ${dora_c_include_dir}/node
+    )
+
+    add_custom_target(Dora_c DEPENDS ${dora_c_include_dir})
+    set(dora_link_dirs ${DORA_ROOT_DIR}/target/debug)
+else()
+    include(ExternalProject)
+    ExternalProject_Add(Dora
+        PREFIX ${CMAKE_CURRENT_BINARY_DIR}/dora
+        GIT_REPOSITORY https://github.com/dora-rs/dora.git
+        GIT_TAG main
+        BUILD_IN_SOURCE True
+        CONFIGURE_COMMAND ""
+        BUILD_COMMAND
+            cargo build
+            --package dora-node-api-c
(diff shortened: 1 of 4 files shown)


binaries/cli/src/template/c/node-cmake-template.txt
```diff
@@ -0,0 +1,68 @@
...
+
+    add_custom_command(OUTPUT ${dora_c_include_dir}
+        WORKING_DIRECTORY ${DORA_ROOT_DIR}
+        DEPENDS Dora
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${dora_c_include_dir}
+        COMMAND ${CMAKE_COMMAND} -E copy_directory apis/c/node ${dora_c_include_dir}/node
+    )
+
+    add_custom_target(Dora_c DEPENDS ${dora_c_include_dir})
+    set(dora_link_dirs ${DORA_ROOT_DIR}/target/debug)
```
phil-opp, line 28: We shouldn't hardcode the debug path. What if the user builds in release mode? (There are multiple instances of this issue)

phil-opp (changes requested):

heyong4725:
Hi @swar09 — thanks for tackling this gap. Closing this PR but the fix is being rescued in [#1851](https://github.com/dora-rs/dora/pull/1851), with your git authorship preserved on the commit.

Your work was solid — cross-platform CMake (`cmake -E` instead of `cp` / `mkdir -p`), conditional Unix vs Windows linkage, and the right scope (per-node `CMakeLists.txt` alongside the existing dataflow-level one). The `// TODO: Makefile?` had been sitting in `template/{c,cxx}/mod.rs` since the 1.0 consolidation and your PR was exactly the right shape to close it.

@phil-opp's [CHANGES_REQUESTED on 2026-03-27](https://github.com/dora-rs/dora/pull/1514#pullrequestreview-2772413620) was the one design concern blocking merge: the hardcoded `target/debug` path doesn't work for release builds. The rescue (#1851) applies that fix by switching on `CMAKE_BUILD_TYPE` to set both the cargo `--release` arg and the matching `target/{debug,release}` link directory:

```cmake
if(CMAKE_BUILD_TYPE STREQUAL "Debug" OR NOT CMAKE_BUILD_TYPE)
    set(dora_cargo_profile_dir "debug")
    set(dora_cargo_release_arg "")
else()
    set(dora_cargo_profile_dir "release")
    set(dora_cargo_release_arg "--release")
endif()
```

The git Author on the rescue commit stays as `swar09 <ghuleswarnit@gmail.com>` (preserved via cherry-pick + `--author`), so when #1851 lands your name is on the squash-merge commit history as the original author.

A note for next time: it's OK to take a few weeks before nudging on a review request, but a quick "thanks, here's my fix for the debug path" follow-up commit within a week or two would have unblocked the merge without anyone else needing to step in. The fix you needed to make was small (one `if`, two interpolations); two reviewers had already validated the surrounding shape was right. Don't let a single round of review feedback feel like a rejection — for a first-time PR especially, the review comments are usually one or two specific things, not a "redesign everything" signal.

Thanks again — the rescue is mechanical because your design was good. Reopen anything you'd like to revisit, or pick another good-first-issue when you have time.

PR #1515, opened 2026-03-22
Prototype Dora-managed Python isolation for build and runtime

Harsh-Sahu43:
Introduce a Dora-managed Python isolation flow for Python nodes under `--uv`.

Goal:
make Python node execution reproducible by preparing an isolated environment during build and reusing that exact environment during `dora run` / `dora start`.

Initial MVP scope:
- Python only
- `Dora.toml` + `uv venv` direction
- keep C/C++ and Rust out of scope
- focus on Dora-integrated build/runtime behavior rather than a full package manager

Planned steps:
- [x] add metadata/plumbing for managed Python env paths per node
- [x] create or reuse managed Python envs during `dora build --uv`
- [x] run Python dependency/build commands against the managed env
- [x] reuse the managed env interpreter in `dora run` / `dora start`
- [ ] add focused tests for build-to-runtime env reuse
- [ ] document the Python isolation workflow and limitations

Notes:
- The intended model is per Python node/package rather than one shared env per dataflow.
- This is meant to be a prototype foundation for Project #2, not a full package manager.
- Python dependency management should remain compatible with existing Python tooling rather than replacing it outright.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,9 +59,20 @@ jobs:
       - name: Free disk Space (Windows)
         if: runner.os == 'Windows'
         run: |
-          $ErrorActionPreference = 'SilentlyContinue'
-          docker system prune --all -f 2>$null; $true
-          docker builder prune -af 2>$null; $true
+          if (Get-Command docker -ErrorAction SilentlyContinue) {
+            docker system prune --all -f 2>$null
+            if ($LASTEXITCODE -ne 0) {
+              Write-Host "docker system prune failed; continuing with Windows cleanup"
+              $global:LASTEXITCODE = 0
+            }
+            docker builder prune -af 2>$null
+            if ($LASTEXITCODE -ne 0) {
+              Write-Host "docker builder prune failed; continuing with Windows cleanup"
+              $global:LASTEXITCODE = 0
+            }
+          } else {
+            Write-Host "docker is not available; skipping docker cleanup"
+          }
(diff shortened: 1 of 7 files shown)


heyong4725:
Hi @Harsh-Sahu43, thanks for the substantial design work on managed Python environments.

After 7+ weeks of drift in `main` (especially in `libraries/core/src/build/` and the daemon spawn path), a direct merge surfaced 5 conflicts — including a total rewrite of `.github/workflows/ci.yml` from the recent CI rebalance (#1716). Rebasing your branch in place would have been more error-prone than reimplementing.

Your design was the right shape:

- Per-node venv at `<working-dir>/.dora/python-env/`
- `uv venv --clear` for creation, idempotent via `import dora` probe
- Prefer in-tree workspace package over PyPI when available
- Fail-closed when interpreter is missing (don't silently fall through to ambient Python)
- Reuse the interpreter at runtime via `Spawner::spawn_node`

Follow-up PR: **#1820** reimplements all of the above against current `main`, plus:

- Unit tests for the new helpers (you marked tests as TODO in your PR — added 4 tests)
- Docs in `docs/cli.md` and `guide/src/operations/cli.md` (you marked docs as TODO — added the section)
- Restart-path env reuse (re-derives the env dir deterministically so restarts pick up the same venv)
- Adapted to current `main`'s parallel-build flow and security-hardened daemon spawn (env denylist, etc.)

You're credited as `Co-Authored-By` on the rescue commit. Closing this PR as superseded by #1820.

PR #1562, opened 2026-03-27
fix(message): prevent panic on malformed UUID in log fields

suhr25:
### **SUMMARY**

Fixes a daemon crash caused by `.unwrap()` when parsing UUIDs from log fields. Invalid UUIDs from node logs would panic the system.

Changes are in `libraries/message/src/common.rs` inside `LogMessage::from(...)`.

---

#### **Before**

```rust id="9et2c1"
.map(|id| BuildId(Uuid::parse_str(&id).unwrap()))
```

#### **After**

```rust id="x68dih"
.and_then(|id| Uuid::parse_str(&id).ok().map(BuildId))
```

---

### **VERIFICATION**

To verify, emit a log from a node process with an invalid UUID such as `{"fields": {"build_id": "not-a-uuid"}}`. Before this change, the daemon would panic while converting the log message. After the fix, the daemon continues running normally and simply treats the invalid UUID field as `None`.


phil-opp (approved):
Thanks!

PR #1566, opened 2026-03-28
Update README.md: add beginner quick start conceptual section

YashBhardwaj2305:
This PR adds a beginner-focused conceptual "Quick Start" section to the README.

It explains:
- who Dora is for
- how to think about dataflows and nodes
- what Dora is and is not

This aims to improve onboarding for first-time users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,49 @@
   </a>
 </p>
 
+## Beginner Quick Start (Conceptual)
+
+### Who is Dora for?
+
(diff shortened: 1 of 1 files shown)


heyong4725:
Closing as superseded by the current README rewrite on main. The PR was written against the old README layout and now conflicts; main already has a full Quick Start, install flow, feature overview, architecture section, and language support table. The conceptual onboarding angle still has some value, but it should be ported as a smaller follow-up "mental model" section if we want it, rather than rebasing this old branch as-is.

PR #1573, opened 2026-03-29
feat(cli): add `--mode` flag to `dora build` for parallel/sequential builds

### LINKED ISSUE #1570: Add Configurable Build Mode `Parallel` vs `Sequential`
# Summary

As of now, `dora build dataflow.yml` builds one node at a time, which consumes a lot of time during the development phase, even when the target machine is capable enough to handle parallel compilation. However, the actual onboard robot hardware may not be capable of handling concurrent builds. There is currently no flexible option to control build behavior based on the target machine's capabilities.

# How to Reproduce

Build a dataflow with 5 to 10 nodes, each with a few external crates/libraries, and observe that they build sequentially even though the target machine is capable of parallel compilation.

# Proposed Changes

Add a `-m` / `--mode` flag to `dora build` that allows the user to select the build strategy:

```bash
# Build nodes one at a time (current default behavior, safe for low-spec hardware)
dora build -m sequential dataflow.yml

# Build nodes concurrently (faster on capable development machines)
dora build -m parallel dataflow.yml
```

| Flag | Behavior |
|---|---|
| `-m sequential` | Builds nodes one at a time in dependency order (current default) |
| `-m parallel` | Builds all eligible nodes concurrently |

# Verification of Changes

### Automated Tests

- Run `cargo check -p dora-daemon` to ensure the changes compile correctly
- Run `cargo test -p dora-daemon` to run existing unit tests
- Run `cargo test --test example-tests` to ensure that dataflows can still start and run correctly

### Manual Verification

Build a dataflow with 5 to 10 nodes, each pulling in a few external crates/libraries, and verify:

- With `-m sequential` - nodes build one at a time (existing behavior preserved)
- With `-m parallel` - build time is significantly reduced on a capable machine

swar09:
## Summary

Closes #1570

As of now, `dora build dataflow.yml` builds nodes one at a time. This is safe for low-spec hardware but wastes time on capable development machines. This PR adds a `--mode` flag to `dora build` so users can opt into parallel building.

## Usage

```bash
# Sequential (current default behavior, backward compatible)
dora build dataflow.yml

# Explicitly sequential
dora build --mode sequential dataflow.yml

# Parallel (faster on capable machines)
dora build --mode parallel dataflow.yml
```
## Note 
The feature is most valuable for use cases like dataflows with Python, Rust, and C/C++ nodes from separate projects.
Works for Any combination of independent build commands.
For a pure Rust workspace with all nodes inside it, the gain is minimal because Cargo itself serializes via locks.

### DIFF
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -86,6 +86,16 @@ pub struct Build {
     // Run build on local machine
     #[clap(long, action)]
     local: bool,
+    /// Build mode: `sequential` or `parallel`. Defaults to `sequential` for backward compatibility
+    #[clap(long, value_enum, default_value_t = BuildMode::Sequential)]
+    mode: BuildMode,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, clap::ValueEnum)]
+pub enum BuildMode {
+    #[default]
+    Sequential,
+    Parallel,
 }
 
 impl Executable for Build {
@@ -97,6 +107,7 @@ impl Executable for Build {
             self.coordinator_port,
             self.uv,
             self.local,
+            self.mode,
         )
         .await
     }
@@ -108,6 +119,7 @@ pub fn build(
     coordinator_port: Option<u16>,
     uv: bool,
     force_local: bool,
+    mode: BuildMode,
 ) -> eyre::Result<()> {
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
@@ -119,6 +131,7 @@ pub fn build(
         coordinator_port,
         uv,
         force_local,
+        mode,
     ))
 }
 
@@ -128,6 +141,7 @@ pub async fn build_async(
     coordinator_port: Option<u16>,
     uv: bool,
     force_local: bool,
+    mode: BuildMode,
 ) -> eyre::Result<()> {
     let dataflow_path = resolve_dataflow(dataflow)
         .await
@@ -156,15 +170,13 @@ pub async fn build_async(
     let session = || connect_to_coordinator_rpc_with_defaults(coordinator_addr, coordinator_port);
 
     let build_kind = if force_local {
-        tracing::info!("Building locally, as requested through `--force-local`");
+        log::info!("Building locally, as requested through `--force-local`");
         BuildKind::Local
     } else if dataflow_descriptor.nodes.iter().all(|n| n.deploy.is_none()) {
-        tracing::info!("Building locally because dataflow does not contain any `deploy` sections");
+        log::info!("Building locally because dataflow does not contain any `deploy` sections");
         BuildKind::Local
     } else if coordinator_addr.is_some() || coordinator_port.is_some() {
-        tracing::info!(
-            "Building through coordinator, using the given coordinator socket information"
-        );
+        log::info!("Building through coordinator, using the given coordinator socket information");
         // explicit coordinator address or port set -> there should be a coordinator running
         BuildKind::ThroughCoordinator {
             coordinator_client: session()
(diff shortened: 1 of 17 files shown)


binaries/cli/src/command/build/mod.rs
```diff
@@ -86,6 +86,16 @@ pub struct Build {
     // Run build on local machine
     #[clap(long, action)]
     local: bool,
+    /// Build mode sequential or parallel. Defaults to sequential
```
phil-opp, line 89: We should note here that parallel builds should only be used when the build commands support this.

binaries/cli/src/command/build/mod.rs
```diff
@@ -156,15 +170,13 @@ pub async fn build_async(
     let session = || connect_to_coordinator_rpc_with_defaults(coordinator_addr, coordinator_port);
 
     let build_kind = if force_local {
-        tracing::info!("Building locally, as requested through `--force-local`");
+        log::info!("Building locally, as requested through `--force-local`");
```
phil-opp, line 173: Why the change from tracing to log? (same below)

phil-opp (commented):
Thanks for the PR! This seems like a good idea.

We had parallel builds as default in the past, but this led to some issues with some build systems, so we chose the safer sequential build order instead. But making this configurable is even better.

heyong4725:
Thanks for this PR @swar09, and especially for filing the underlying issue #1570 — your diagnosis was exactly right: building nodes one at a time wastes wall-clock time on capable dev machines, and the user needed a way to opt into parallel execution.

Closing this PR because **the feature has already shipped in main** as part of the dora 1.0 consolidation (commit `145ccce0`). The `--parallel` flag is now available on `dora build`:

```bash
dora build --parallel dataflow.yml
```

Quick comparison so you can see where the in-tree version differs from your design:

| Aspect | This PR | Main today |
|---|---|---|
| Flag | `--mode <sequential\|parallel>` | `--parallel` (bool) |
| Concurrency | `futures::stream::FuturesUnordered` | `tokio::task::JoinSet` |
| Single-task short-circuit | always parallel path when chosen | guarded on `tasks.len() > 1` |
| Trace log when active | none | `tracing::info!("Building N nodes in parallel")` |
| Examples threading | 11 example `run.rs` updated to pass param | not needed (async refactor handled it) |

The main differences are:

- **`JoinSet` over `FuturesUnordered`**: For build steps that shell out to `cargo` / `maturin` / `cmake` (CPU-bound subprocesses), `JoinSet` spawns real tokio tasks that the runtime can schedule on independent worker threads. `FuturesUnordered` only polls in-order on the calling task, so subprocess work that yields (waiting on stdout/stderr) wouldn't actually parallelize across cores. Same intent, different runtime behavior.
- **Bool over enum**: matches the `cargo build --jobs N` convention closer; one flag instead of an enum that the user has to remember the variants of.
- **Async-throughout refactor**: the 1.0 rewrite made the build path fully async (no `block_on` calls), which removed the need to thread the parallel param through example `run.rs` files.

None of those design choices invalidate your approach — both implementations solve the same problem with the same trade-offs around the user's machine capability. The 1.0 consolidation just happened to land a parallel scheduler concurrent with your work.

**What carries forward from this PR:**

- The issue diagnosis (#1570) is the durable artifact. The feature exists because you filed it; the implementation that shipped is functionally identical to what you proposed.
- The note in your PR description — "most valuable for use cases like dataflows with Python, Rust, and C/C++ nodes from separate projects" — is the exact reason this matters. Pure-cargo workspaces don't benefit (cargo serializes via locks), but cross-language dataflows scale linearly with core count. This is worth surfacing in the CLI help text or docs; filing a follow-up if not already there.

Thanks again — this kind of "obvious-once-you-think-about-it" usability win is great to get reports on. Apologies for the slow turn-around; if you have more proposals like this, please keep them coming, ideally in a fresh PR against the current main since the build subsystem has changed substantially since March.

---

PR #1578, opened 2026-03-31
feature: service (request-reply pattern) implementation

### LINKED ISSUE #390: Adding a simple requests
Implementing a `service` features could simplify implementing certain dataflows, including supporting ROS 2 legacy applications. 

**API details**

I think that we can actually build around our current pub-sub pattern to support request-reply.

What we could do is to make a new method called `send_request` or `request` that will send an output and receive an input in one blocking function call. The function will return the reply from the request.

To do that, we could add the following dataflow pattern:

```yaml
  - id: webcam
    custom:
      source: ./webcam.py
      inputs:
        tick:
          source: dora/timer/millis/50
          queue_size: 1000
      outputs:
        - image
  - id: object_detection
    operator:
      python: ../operators/webcam/object_detection.py
      services:
        get_image:
           reply: webcam/image
           request: webcam/tick
```
With the following python operator request code:
```python
image = send_service("get_image", tick)


## alternatively
image = send("service", "get_image", tick)
image = send("request", "get_image", tick)
image = send_request("get_image", tick)
image = request("get_image", tick)
```

With the following python custom node request code:
```python
image = node.send_service("get_image", tick)


## alternatively
image = node.send("service", "get_image", tick)
image = node.send("request", "get_image", tick)
image = node.send_request("get_image", tick)
image = node.request("get_image", tick)```
```

The webcam code does need to be changed. And the underlying communication design does not need to change as well so we can still make it flexible.

**Edge cases details**

- We should try to keep message that has been received in between the request-reply, so that to not miss out on important messages. And so some work has to be done within the message queue.

**Edge cases details**

MinPoe:
Closes [#390](https://github.com/dora-rs/dora/issues/390)

## Summary

Implements the services (request-reply) pattern for dora-rs, allowing nodes and operators to perform synchronous request-reply communication built on top of the existing pub-sub infrastructure.

NOTE: The service implementation does not support operators to be a client as its on_event callback should not be blocked. This could be a future additional implementation (e.g., via an async callback or a separate on_service_reply handler).

### YAML Dataflow Configuration
- New `services` block on nodes and operators with `type: server` and `type: client` (with `server: node_id/service_name` reference)
- JSON schema updated with `ServiceEndpoint` definition
- YAML validation ensures client references point to valid server endpoints, including operator-level services

### Rust Node API (`apis/rust/node/`)
- `DoraNode::send_request()` -- blocking call that sends a request and waits for a matching reply using correlation IDs (UUID v7). Supports optional timeout. Events received while waiting are buffered and returned by subsequent `recv()` calls.
- `DoraNode::send_service_reply()` -- sends a reply to a service request
- `Event::ServiceRequest` variant for server-side event handling
- `ServiceReply` struct returned by `send_request()`
- `recv_raw()` and `recv_raw_timeout()` on `EventStream` to bypass the scheduler during blocking request calls

### Python Node API (`apis/python/node/`)
- `node.send_request(service_name, data, metadata=None, timeout=None)` -- blocking, returns dict with `value` and `metadata`
- `node.send_service_reply(service_name, data, metadata=None)`
- `SERVICE_REQUEST` event type delivered to the Python event loop
- Type stubs updated in `__init__.pyi`

### Daemon (`binaries/daemon/`)
- Synthetic `__service_request_` and `__service_reply_` channels auto-created from service declarations
- `node_inputs()` injects synthetic inputs for both custom nodes and operators
- `CoreNodeKindExt::run_config()` injects synthetic outputs for both custom nodes and operators
- Service mappings translate between client-side and server-side channel names (supporting operator-prefixed names)
- `DaemonRequest::SendServiceReply` added to the control channel
- `send_output_to_local_receivers()` routes service messages through dedicated mapping tables

### Event Stream (`apis/rust/node/src/event_stream/`)
- `event_stream_loop` converts incoming `__service_request_*` inputs into `NodeEvent::ServiceRequest` events with the prefix stripped
- `convert_event_item` handles `NodeEvent::ServiceRequest` to `Event::ServiceRequest` conversion

### Runtime / Operator Support (`binaries/runtime/`)
- `ServiceRequest` events routed to operators with operator prefix stripped
- `SendServiceReplyCallback` added to the Python operator event dict as `send_service_reply` key (backward compatible, no signature change to `on_event`)
- `OperatorEvent::ServiceReply` forwarded through the runtime to the daemon with correct output id formatting

### Examples (`examples/request-reply/`)
- `dataflow.yml` -- custom node service example
- `dataflow_operator.yml` -- operator service example
- `requester_node.py` -- client node testing data types (int64, float64, uint8, string, bool, int32, empty), event buffering, and timeout
- `service_node.py` -- custom node echo server
- `service_operator.py` -- operator echo server

## Testing

Both dataflows pass all tests:
- 7 Arrow data types round-tripped correctly
- Event buffering verified (ticks not lost during blocking requests)
- Timeout on unanswered requests verified
- Custom node and operator server paths both validated

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1750,6 +1750,7 @@ dependencies = [
  "shared_memory_extended",
  "tokio",
  "tracing",
+ "uuid",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ git2 = { version = "0.20.4", features = ["vendored-openssl"] }
 serde_yaml = "0.9.33"
 zenoh = "1.1.1"
 serde_json = "1.0.145"
+uuid = { version = "1.22.0", features = ["v4"] }
 
 [package]
 name = "dora-examples"
--- a/apis/python/node/dora/__init__.pyi
+++ b/apis/python/node/dora/__init__.pyi
@@ -166,6 +166,29 @@ class Node:
         ```
         """
 
+    def send_service_reply(
+        self, service_name: str, data: pyarrow.Array, metadata: dict = None
+    ) -> None:
+        """`send_service_reply` sends a reply to a service request (server-side).
+
+        ```python
+        node.send_service_reply("get_image", image_data, event["metadata"])
+        ```
+        """
+
+    def send_request(
+        self, service_name: str, data: pyarrow.Array, metadata: dict = None
+    ) -> dict:
+        """`send_request` sends a service request and blocks until the reply arrives.
+
+        Returns a dict with 'value' (pyarrow.Array) and 'metadata' (dict).
+
+        ```python
+        reply = node.send_request("get_image", pa.array([]))
+        image = reply["value"]
+        ```
+        """
+
     def __iter__(self) -> typing.Any:
         """Implement iter(self)."""
 
--- a/apis/python/node/dora/__init__.pyi
+++ b/apis/python/node/dora/__init__.pyi
@@ -177,14 +177,14 @@ class Node:
         """
 
     def send_request(
-        self, service_name: str, data: pyarrow.Array, metadata: dict = None
+        self, service_name: str, data: pyarrow.Array, metadata: dict = None, timeout: float = None
     ) -> dict:
         """`send_request` sends a service request and blocks until the reply arrives.
 
         Returns a dict with 'value' (pyarrow.Array) and 'metadata' (dict).
-
+        If timeout (in seconds) is provided and expires, raises an error.
         ```python
-        reply = node.send_request("get_image", pa.array([]))
+        reply = node.send_request("get_image", pa.array([]), timeout=5.0)
         image = reply["value"]
         ```
         """
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -6,14 +6,16 @@ use std::sync::{Arc, LazyLock};
 use std::time::Duration;
 use tokio::sync::Mutex;
 
-use arrow::pyarrow::{FromPyArrow, ToPyArrow};
 use arrow::array::Array;
+use arrow::pyarrow::{FromPyArrow, ToPyArrow};
 use dora_download::download_file;
 use dora_node_api::dora_core::config::{DataId, NodeId};
 use dora_node_api::dora_core::descriptor::source_is_url;
 use dora_node_api::merged::{MergeExternalSend, MergedEvent};
 use dora_node_api::{DataflowId, DoraNode, EventStream, TryRecvError, init_tracing};
-use dora_operator_api_python::{DelayedCleanup, NodeCleanupHandle, PyEvent, pydict_to_metadata, metadata_to_pydict};
+use dora_operator_api_python::{
+    DelayedCleanup, NodeCleanupHandle, PyEvent, metadata_to_pydict, pydict_to_metadata,
+};
 use dora_ros2_bridge_python::Ros2Subscription;
 use eyre::{Context, ContextCompat};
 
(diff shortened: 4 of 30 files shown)


heyong4725:
Thanks for this PR @MinPoe, and for the thorough write-up — the operator-as-client constraint you flagged (operators can't block in `on_event`, so they can only serve, not call) is exactly the kind of design observation that's hard to spot until you've actually built the thing. Saving that finding for the record below.

Closing this PR because the **service (request-reply) pattern has shipped in main**, but via a structurally different design than this PR proposes. Issue #390 was closed when that work landed.

## Design comparison

| Aspect | This PR | Main (shipped) |
|---|---|---|
| YAML surface | new `services` block with `type: server` / `type: client` | none — standard `inputs:` / `outputs:` |
| Channel mechanism | synthetic `__service_request_*` / `__service_reply_*` channels + daemon routing tables | regular pub/sub with a `REQUEST_ID` metadata key |
| Event delivery | new `Event::ServiceRequest` variant + `SERVICE_REQUEST` Python event type | regular `Event::Input` carrying `request_id` in metadata |
| Client API | `send_request()` — **blocking**, buffers events during the wait | `send_service_request()` — **non-blocking**, returns the request_id |
| Reply matching | implicit (via the blocking wait) | explicit (caller maintains a `HashMap<request_id, pending>`) |
| Operator client support | not supported (by design — `on_event` can't block) | supported (no blocking) |
| Daemon plumbing | substantial (synthetic input injection, service mapping tables) | none — pub/sub handles it |

The shipped design is documented at [`docs/patterns.md`](https://github.com/dora-rs/dora/blob/main/docs/patterns.md) and there's a Rust + Python example at [`examples/service-example/`](https://github.com/dora-rs/dora/tree/main/examples/service-example).

## Why the design diverged

Main's design is deliberately minimal — no new event variants, no synthetic channels, no daemon-level service mapping tables. The service pattern is just pub/sub with a `REQUEST_ID` metadata key that callers correlate on. That keeps the daemon's input/output graph homogeneous and lets the same wire format carry topic, service, action, and streaming patterns (see `docs/patterns.md`).

Your design's main differentiator is **synchronous `send_request`** — caller blocks until the reply lands, with events that arrive in the meantime buffered for the next `recv()`. That's genuinely ergonomic for simple request-reply, but the dora team chose the non-blocking path because:

1. **Blocking in a node's event loop is risky in dataflow systems.** A blocked node can't react to `Stop`, can't process upstream back-pressure signals, and can't observe input-closure events from other producers. Buffering papers over the symptom but doesn't fix the root cause — the node is still effectively offline for the duration of the request.
2. **Reordering across buffered events.** Events received during the blocking wait are returned in arrival order, but they were *delayed* relative to the request-reply pair. For some use cases (timer ticks, watchdog signals) the delay matters; for others it doesn't. The non-blocking model lets the user decide per call.
3. **The operator-as-client gap your PR acknowledges** is fundamental to the blocking approach — you can't extend it to operators without also rewriting `on_event` to be async, which has its own cost. The non-blocking approach works for operators out of the box.

Neither design is "right"; they're different points on the latency-vs-control trade-off curve. Main happened to land on the non-blocking point of that curve.

## What's still valuable from this PR

- **The operator-as-client gotcha** you documented in the PR description is a durable finding. If anyone ever wants to add a `send_request_async` helper to dora, they'll need to confront exactly this problem; preserving your note here means they won't rediscover it from scratch.
- **The request_id-as-UUID-v7 choice** matches what main shipped (`Self::new_request_id` at `apis/rust/node/src/node/mod.rs:1244` returns UUID v7). Convergent insight; nice that both designs landed on the same correlation ID format.
- **The synchronous-API ergonomics argument** is real. Even though main intentionally doesn't ship a blocking wrapper, anyone building a higher-level Python framework on top of dora could write a `send_request_blocking` helper as a 30-line user-space wrapper around `send_service_request` + event-loop polling. Worth a docs recipe if anyone wants to file one.

## Won't-apply note

The PR's daemon plumbing (synthetic channel injection in `node_inputs()` / `run_config()`) is against a daemon surface that was substantially refactored in the [dora 1.0 consolidation](https://github.com/dora-rs/dora/commit/145ccce0) (#1820 and many follow-ups). Even if the design were aligned, the PR wouldn't rebase cleanly — most of the daemon changes would need to be rewritten.

Thanks again for the contribution and especially for the issue-level thinking. Apologies for the slow turn-around. If you have follow-up ideas for the service pattern (a docs recipe for blocking wrappers, a typed-service API on top of the existing metadata-based one, or an explicit critique of main's design that you think should change it), please feel free to open a fresh discussion or PR against current main.

---

PR #1583, opened 2026-03-31
Fixed CommunicationConfig deserialization and added tests for YAML parsing

### LINKED ISSUE #1582: CommunicationConfig YAML deserialization fails for tcp and unixdomain
## Problem

While working with `CommunicationConfig` in `libraries/message/src/config.rs`, I noticed that YAML deserialization fails for valid values like `tcp` and `unixdomain` under `_unstable_local`.

Example YAML:

```yaml
_unstable_local: tcp
_unstable_remote: tcp
```

This results in an error:

```
unknown variant `tcp`, expected `Tcp` or `UnixDomain`
```

## Cause

The enum deserialization expects `Tcp` and `UnixDomain`, but the YAML configuration uses lowercase values (`tcp`, `unixdomain`), which causes deserialization to fail.

## Proposed Fix

Add explicit serde renames for enum variants:

* `Tcp` → `"tcp"`
* `UnixDomain` → `"unixdomain"`

## Additional Work

I also added tests to verify YAML parsing for:

* valid tcp config
* valid unixdomain config
* invalid config
* unknown fields

If this approach is acceptable, I can open a PR with:

* the deserialization fix
* Display implementation for the enums
* tests for CommunicationConfig parsing

I am also interested in contributing to the cleanup/refactor of unstable features mentioned in the GSoC project ideas, so I started exploring this part of the codebase.

Siddharth-777:
Fixes #1582 

## Summary

This PR fixes YAML deserialization for `CommunicationConfig` in `libraries/message/src/config.rs` and adds tests for communication configuration parsing.

## Problem

YAML values like:

```yaml
_unstable_local: tcp
_unstable_remote: tcp
```

were failing to deserialize because the enum variants expected `Tcp` and `UnixDomain`.

## Changes

* added explicit serde renames for enum variants (`tcp`, `unixdomain`)
* implemented `Display` for `LocalCommunicationConfig` and `RemoteCommunicationConfig`
* added tests for:

  * valid tcp config
  * valid unixdomain config
  * invalid config
  * unknown fields

## Testing

```bash
cargo test --test communication_config_tests
```

## Notes

I started exploring the unstable communication config area as part of the cleanup/refactor work mentioned in the GSoC project ideas and plan to contribute more in this area.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "cmake.sourceDirectory": "D:/CODE/dora/examples/cmake-dataflow"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "cmake.sourceDirectory": "D:/CODE/dora/examples/cmake-dataflow"
+}
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -229,12 +229,28 @@ pub struct CommunicationConfig {
     pub remote: RemoteCommunicationConfig,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize)]
 pub enum LocalCommunicationConfig {
-    #[serde(rename = "tcp")]
     Tcp,
-    #[serde(rename = "unixdomain")]
     UnixDomain,
+}
+
+impl<'de> Deserialize<'de> for LocalCommunicationConfig {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+
+        match value.as_str() {
+            "Tcp" | "tcp" => Ok(Self::Tcp),
+            "UnixDomain" | "unixdomain" => Ok(Self::UnixDomain),
+            other => Err(serde::de::Error::unknown_variant(
+                other,
+                &["Tcp", "tcp", "UnixDomain", "unixdomain"],
+            )),
+        }
+    }
 }
 
 impl Default for LocalCommunicationConfig {
@@ -243,11 +259,23 @@ impl Default for LocalCommunicationConfig {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone)]
-#[serde(deny_unknown_fields)]
+#[derive(Debug, Serialize, Clone)]
 pub enum RemoteCommunicationConfig {
-    #[serde(rename = "tcp")]
     Tcp,
+}
+
+impl<'de> Deserialize<'de> for RemoteCommunicationConfig {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+
+        match value.as_str() {
+            "Tcp" | "tcp" => Ok(Self::Tcp),
+            other => Err(serde::de::Error::unknown_variant(other, &["Tcp", "tcp"])),
+        }
+    }
 }
 
 impl Default for RemoteCommunicationConfig {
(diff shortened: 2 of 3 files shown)


haixuanTao (changes requested):
CI is failing

heyong4725:
Thanks for this PR @Siddharth-777, and for filing the underlying issue #1582 — the diagnosis ("lowercase YAML like `_unstable_local: tcp` fails to parse because serde expects `Tcp`") was exactly right, and that's a real usability gap that needed to close.

Closing this PR because the **issue has been resolved in main** (and #1582 is now closed), but via a different fix than this PR proposes.

## What landed in main

Looking at `libraries/message/src/config.rs:404` and `:425` on current main:

```rust
#[derive(Debug, Default, Serialize, Deserialize, JsonSchema, Clone)]
#[serde(deny_unknown_fields, rename_all = "lowercase")]
pub struct CommunicationConfig { ... }

#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize, Default)]
#[serde(rename_all = "lowercase")]
pub enum LocalCommunicationConfig {
    #[default]
    Tcp,
}

impl fmt::Display for LocalCommunicationConfig {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            Self::Tcp => write!(f, "tcp"),
        }
    }
}
```

Plus 5 inline test cases in the same file (`#[cfg(test)] mod tests`) covering: valid `_unstable_local: tcp` parse, valid `_unstable_remote: tcp` parse, `_unstable_local: unixdomain` correctly rejected, `Display` roundtrip, full struct serialize→deserialize.

## Comparison

| Aspect | This PR | Main (shipped) |
|---|---|---|
| Lowercase YAML parse | hand-rolled `Deserialize` impl matching `"Tcp"` and `"tcp"` arms | `#[serde(rename_all = "lowercase")]` (one line, idiomatic) |
| `Display` impl | yes, both enums | yes, both enums |
| `UnixDomain` variant | retained | **deliberately removed** (only `Tcp` exists) |
| Tests | separate `tests/communication_config_tests.rs` | inline `#[cfg(test)] mod tests` in `config.rs` |

## Why the fix shape diverged

Two concrete reasons:

**1. `UnixDomain` was deliberately removed**, not just renamed. Looking at the `LocalCommunicationConfig` enum on main, only `Tcp` exists. There was no Unix-domain-socket transport implementation backing the variant on the daemon side — it was a placeholder. Trimming it during the 1.0 consolidation closed the gap between "declarable in YAML" and "actually works". Re-adding the variant in this PR would resurrect dead code (declarable but not implemented), which is worse than the original parse failure: a user could write `_unstable_local: unixdomain`, get past parse, and then hit a confusing runtime error somewhere later in the daemon. Main's "parse fails fast with a clear `unknown variant` error" path is the better one.

**2. `#[serde(rename_all = "lowercase")]` is preferred over hand-rolled `Deserialize`.** Same outcome (lowercase parses correctly), one line vs. ~30 lines, and it stays consistent with the rest of the codebase's deserialization style. The hand-rolled approach also requires updating every time someone adds a variant — `rename_all` doesn't. Less maintenance burden, less surface area for subtle bugs (e.g. `serde::de::Error::unknown_variant` semantics differ across PRs that hand-roll vs. let serde do it).

## What carries forward from this PR

- **The issue diagnosis** (#1582) is the durable artifact — the parse failure was real and now correctly fixed in main.
- The PR's tests for "invalid config rejected" and "unknown fields rejected" are good safety-net cases; main already has equivalents via `deny_unknown_fields` + the `unixdomain` rejection test.

## Future contribution path

You mentioned interest in the "cleanup/refactor of unstable features" GSoC track. The `_unstable_local` / `_unstable_remote` naming is itself a marker that this surface is in flux — `RemoteCommunicationConfig` only has `Tcp` (per main) but the project has been moving toward Zenoh for cross-machine transport. If you want to pick that up, a good follow-up would be reviewing whether `_unstable_*` is the right configuration shape going forward, or whether the daemon's transport layer should expose a different surface entirely. Happy to discuss in a fresh issue or discussion thread.

Thanks again for the diagnosis and the test infrastructure thinking. Apologies for the slow turn-around.

---

PR #1584, opened 2026-03-31
added deprecated note

psg-19:
Replaced the informal deprecation notice in the doc comment of `CustomNode::envs` with a proper `Rust #[deprecated]` attribute, so the compiler emits warnings when this field is used.


haixuanTao:
Thanks, can you fix the formatting issue?

PR #1602, opened 2026-04-07
Replace daemon-node custom protocol with tarpc RPC

phil-opp:


phil-opp:
Closing as we decided to go with a websocket-based protocol instead.

PR #1608, opened 2026-04-11
refactor(core): replace unwrap() with proper error handling in node API and core libs

LeonRust:
Replace 6 instances of unwrap() with graceful error handling to prevent unnecessary panics in production.

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -758,8 +758,14 @@ impl DoraNode {
             .map(|(i, _)| i);
         let memory = match cache_index {
             Some(i) => {
-                // we know that this index exists, so we can safely unwrap here
-                self.cache.remove(i).unwrap()
+                // Index was just found via self.cache.iter().enumerate(), so removal
+                // should succeed. Using expect() instead of unwrap() provides a clear
+                // diagnostic message if this invariant is ever violated due to a bug
+                // in cache management or concurrent modification.
+                self.cache.remove(i).expect(
+                    "shared memory cache index should exist: \
+                    found via iter() but missing on remove() indicates cache corruption",
+                )
             }
             None => ShmemHandle(Box::new(
                 ShmemConf::new()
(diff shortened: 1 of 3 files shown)


phil-opp:
Wow, this PR description feels like a new record for unnecessary verboseness. I trimmed it down to the once sentence that would be needed for such a simple change. Please keep your PR descriptions short in the future, nobody has time to read that wall of text.

apis/rust/node/src/node/mod.rs
```diff
@@ -758,8 +758,14 @@ impl DoraNode {
...
-                // we know that this index exists, so we can safely unwrap here
-                self.cache.remove(i).unwrap()
+                // Index was just found via self.cache.iter().enumerate(), so removal
+                // should succeed. Using expect() instead of unwrap() provides a clear
+                // diagnostic message if this invariant is ever violated due to a bug
+                // in cache management or concurrent modification.
+                self.cache.remove(i).expect(
+                    "shared memory cache index should exist: \
+                    found via iter() but missing on remove() indicates cache corruption",
+                )
```
phil-opp, line 768: The unwrap already gives us file and line info, which should be enough to understand the issue. We definitely don't need to put the reason for the expect in the comment, this is just extra noise.

apis/rust/node/src/node/mod.rs
```diff
@@ -973,25 +979,49 @@ pub fn init_tracing(
...
+                    "tracing guard mutex was poisoned, recovering: \
+                    another task may have panicked while holding the lock"
+                );
+                let mut guard = poisoned.into_inner();
+                *guard = builder.guard.take();
+            }
+        }
+
+        // Tracing subscriber setup failure should not crash the node.
+        // Log the error and continue with default logging.
```
phil-opp, line 1018: Log the error where? If there is no tracing subscriber, nobody will receive the logs? I think it's probably better to error than to log nothing at all.

apis/rust/node/src/node/mod.rs
```diff
@@ -973,25 +979,49 @@ pub fn init_tracing(
...
+            Ok(mut guard) => *guard = builder.guard.take(),
+            Err(poisoned) => {
+                tracing::warn!(
+                    "tracing guard mutex was poisoned, recovering: \
+                    another task may have panicked while holding the lock"
+                );
+                let mut guard = poisoned.into_inner();
+                *guard = builder.guard.take();
+            }
+        }
```
phil-opp, line 1015: I don't think that we need that. This is very unlikely to ever happen

haixuanTao:
@LeonRust it's failing formatting, could you fix it?

libraries/core/src/topics.rs
```diff
@@ -26,15 +26,19 @@ fn build_zenoh_config(coordinator_addr: Option<IpAddr>, listen_port: Option<u16>
...
 
     if let Some(port) = listen_port {
-        zenoh_config
+        if let Err(err) = zenoh_config
             .insert_json5("listen/endpoints", &format!(r#"["tcp/0.0.0.0:{port}"]"#))
-            .unwrap();
+        {
+            tracing::warn!("failed to set zenoh listen endpoint: {err}");
+        }
     }
```
phil-opp, line 42: If a certain port was explicitly asked for, it should be treated as an error if that doesn't work.

libraries/core/src/topics.rs
```diff
@@ -44,12 +48,14 @@ fn build_zenoh_config(coordinator_addr: Option<IpAddr>, listen_port: Option<u16>
...
         if !would_connect_to_self {
-            zenoh_config
+            if let Err(err) = zenoh_config
                 .insert_json5(
                     "connect/endpoints",
                     &format!(r#"["tcp/{}:{}"]"#, addr, DORA_ZENOH_PEER_PORT_DEFAULT),
                 )
-                .unwrap();
+            {
+                tracing::warn!("failed to set zenoh connect endpoint: {err}");
```
phil-opp, line 57: This means that the node/daemon might not reach other nodes. I think it would be better to keep this an error.

LeonRust:
OK, I'll readjust it again.

heyong4725:
Thanks for putting this cleanup together. I re-checked it against current main, and I’m going to close this as stale/superseded rather than ask you to keep rebasing it.

A couple of the original targets have already changed: the tracing setup unwrap handling has been fixed independently, and the old shared-memory cache path this touched no longer exists in the current node API code. The branch now conflicts with the current node/core architecture, so it isn’t a good merge candidate as-is.

There is still one useful idea here: making EventStream event-log write failures non-fatal could be worth revisiting as a small targeted PR with a regression test. Thanks again for the contribution.

PR #1610, opened 2026-04-13
fix(c-api): improve safety and correctness of C node API

haixuanTao:
Small PR to add new functionality from adora in the C/C++ API

## Summary

- Add `const` qualifiers to `id_ptr` and `data_ptr` parameters in `dora_send_output` header declaration
- Use `c_int` return type instead of `isize` for proper C FFI convention
- Add null-pointer checks in `try_send_output` before dereferencing
- Fix error propagation from `send_output_raw` using `?` operator

Ported from dora-rs/adora.

## Test plan
- [x] Verify existing C dataflow examples still compile and run
- [x] Confirm the const-correctness change doesn't break any downstream C code

### DIFF
--- a/apis/c/node/src/lib.rs
+++ b/apis/c/node/src/lib.rs
@@ -3,7 +3,10 @@
 use arrow_array::UInt8Array;
 use dora_node_api::{DoraNode, Event, EventStream, arrow::array::AsArray};
 use eyre::Context;
-use std::{ffi::c_void, ptr, slice};
+use std::{
+    ffi::{c_int, c_void},
+    ptr, slice,
+};
 
 pub const HEADER_NODE_API: &str = include_str!("../node_api.h");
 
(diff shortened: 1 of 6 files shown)


apis/c/node/src/lib.rs
```diff
@@ -265,13 +268,71 @@ unsafe fn try_send_output(
...
     let id = std::str::from_utf8(unsafe { slice::from_raw_parts(id_ptr, id_len) })?;
     let output_id = id.to_owned().into();
     let data = unsafe { slice::from_raw_parts(data_ptr, data_len) };
-    context
+    Ok(context
         .node
         .send_output_raw(output_id, Default::default(), data.len(), |out| {
             out.copy_from_slice(data);
-        })
+        })?)
```
phil-opp:
```suggestion
    context
.node
.send_output_raw(output_id, Default::default(), data.len(), |out| {
out.copy_from_slice(data);
        })
```

phil-opp (approved):

PR #1617, opened 2026-04-13
fix(runtime): load Python operators from exact file paths

### LINKED ISSUE #1616: Python runtime can import the wrong operator module when file stem collides with an importable module
## Problem

Dora runtime currently derives the Python operator module name from the operator file stem and imports it by module name.

This can load the wrong module when another module with the same name is already importable. For example, an operator file named `operator.py` can resolve to Python's standard library `operator` module instead of the file referenced by the dataflow.

That breaks the expected contract for path-based operator execution: Dora is given an operator file path, but module resolution can still be influenced by ambient Python import state.

## How to Reproduce
Create a file `repro.py` at the repo root with the following contents:

```python
import importlib
import pathlib
import sys
import tempfile

root = pathlib.Path(tempfile.mkdtemp())
operator_file = root / "operator.py"
operator_file.write_text(
    "class Operator:\n    pass\n",
    encoding="utf-8",
)

import operator  # stdlib module is already importable

sys.path.append(str(root))
loaded = importlib.import_module("operator")

print(getattr(loaded, "__file__", "<builtin>"))
print(hasattr(loaded, "Operator"))

```

## Current Output : 

Image

## Expected Output: 

Image

Harsh-Sahu43:
Closes #1616 

## Summary

Load Python operators from their exact file paths instead of resolving them only by file stem.

## What Changed

- derive a Dora-specific module name from the canonical operator path
- load the main operator module with `importlib.util.spec_from_file_location(...)`
- keep the operator parent directory on `sys.path` so sibling imports still work
- apply the same path-based loading logic during reload
- add a regression test covering a collision with stdlib `operator`

## Why

Python operators in Dora are configured by file path, so the runtime should execute that exact file deterministically.

Before this change, an operator like `operator.py` could resolve to an already-importable module with the same name instead of the file referenced by the dataflow.

## Validation

```powershell
$python = (py -3 -c "import sys; print(sys.executable)")
$pythonDir = Split-Path $python
$env:PYO3_PYTHON = $python
$env:PATH = "$pythonDir;$env:PATH"
cargo test -p dora-runtime --all-features

### DIFF
--- a/binaries/runtime/src/operator/python.rs
+++ b/binaries/runtime/src/operator/python.rs
@@ -16,6 +16,8 @@ use pyo3::{
     types::{IntoPyDict, PyAnyMethods, PyDict, PyDictMethods, PyTracebackMethods},
 };
 use std::{
+    collections::hash_map::DefaultHasher,
+    hash::{Hash, Hasher},
     panic::{AssertUnwindSafe, catch_unwind},
     path::Path,
 };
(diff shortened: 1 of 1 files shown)


haixuanTao:
So the idea is that, if there is a risk for confusion within Python, this is an error to me in the sense that it will create issues down the line.

Name collusion in python can happen quite frequently and I would prefer that dora resolves it the way python resolves in order to avoid silent error of an operator running but something else breaking and it being very hard to debug.

[github-actions](https://github.com/apps/github-actions)
added 
python
Python API
 
[runtime](https://github.com/dora-rs/dora/issues?q=state%3Aopen%20label%3A%22runtime%22)
 [2 hours ago](https://github.com/dora-rs/dora/issues/1616#event-24454387527)

[haixuanTao](https://github.com/haixuanTao)
closed this as [completed](https://github.com/dora-rs/dora/issues?q=is%3Aissue%20state%3Aclosed%20archived%3Afalse%20reason%3Acompleted)[now](https://github.com/dora-rs/dora/issues/1616#event-24459740587)

Harsh-Sahu43:
thanks!, agreed.

Issue #313, opened 2023-07-04
Cannot exec dora or dora-cli command when install dora with conda on macOS


jinlong-hao:
**Describe the bug**
cannot exec dora or dora-cli command when install dora with conda on macOS

**To Reproduce**

1. create a new conda env :  conda create -n dora python=3.9
2. activate dora:  conda activate dora
3. install dora and dora-cli in conda env:  pip install dora dora-cli 
4. exec dora-cli or dora comman in terminal

**Expected behavior**

expect dora can successfully run.

jinlong-hao:
It's my first time use rust. 
I installed rust and dora with cargo,  the problem solved.

haixuanTao:
I might add the cli to pip so that people can install it directly without cargo. Have fun trying out Rust ☺️

haixuanTao:
For future reference, [`dora-cli`](https://pypi.org/project/dora-cli/) on Pypi is not our project.

PR #1627, opened 2026-04-17
fix: restore fmt + allow YOR token in typos dict (rc.1 hotfix)

heyong4725:
## Summary

CI on the v1.0.0-rc.1 consolidation commit (145ccce0) fails on two checks:

- **Format**: trailing whitespace in `apis/c++/node/build.rs` lines 74 and 177-178
- **Typos**: `YOR-robot` in `docs/plan-dora-1.0-consolidation.md` (lines 1142, 1235) flagged as a misspelling of `YOUR` — but it refers to the real GitHub user `YOR-robot` in the downstream-user assessment table

This PR restores green CI on `main`:

- `cargo fmt --all` applied to `apis/c++/node/build.rs` (whitespace-only)
- Added `YOR = "YOR"` to `_typos.toml` `extend-words` with an inline comment explaining the origin

The `v1.0.0-rc.1` tag stays immutable; these fixes ride with rc.2 per §5 Phase 5b in `docs/plan-dora-1.0-consolidation.md` (tracker: #313).

## Test plan

- [x] `cargo fmt --all -- --check` passes locally
- [x] `typos` passes locally
- [x] `make qa-fast` passes locally (fmt + clippy + audit + unwrap-budget)
- [ ] CI green on this PR
- [ ] After merge, confirm CI on `main` goes green

## Notes

Unwrap budget on rc.1 reported `157 (budget: 185)` — a tightening opportunity for a follow-up, not bundled here to keep the hotfix minimal.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -19,3 +19,5 @@ ser = "ser"
 BA = "BA"
 # Windows system library name (used by the C/C++ publish workflow)
 secur = "secur"
+# GitHub org/user name referenced in downstream-user assessment docs (not "YOUR")
+YOR = "YOR"
--- a/_typos.toml
+++ b/_typos.toml
@@ -19,5 +19,9 @@ ser = "ser"
 BA = "BA"
 # Windows system library name (used by the C/C++ publish workflow)
 secur = "secur"
-# GitHub org/user name referenced in downstream-user assessment docs (not "YOUR")
-YOR = "YOR"
+
+[default]
+# Scoped allow: GitHub user name "YOR-robot" appears in the downstream-user
+# assessment table in docs/plan-dora-1.0-consolidation.md. Matching the full
+# compound token (not bare "YOR") keeps typo detection active everywhere else.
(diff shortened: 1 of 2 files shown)


heyong4725:
No findings from review. I checked the whitespace-only change in apis/c++/node/build.rs and the new exact-case YOR allowlist entry in _typos.toml; both look appropriately scoped for restoring the failing fmt and typos checks. I also reran  and  locally, and both passed.

heyong4725:
I found one issue in the current hotfix: _typos.toml adds YOR as a global extend-words exemption, but the token only appears in two documentation mentions of YOR-robot in docs/plan-dora-1.0-consolidation.md. That broadens typo suppression repo-wide for a token that is not a generally valid project term, so a future accidental YOR typo would now pass lint anywhere. I would scope the exemption to the specific doc occurrence instead of adding it to the shared dictionary. The apis/c++/node/build.rs change looks whitespace-only.

Issue #1631, opened 2026-04-17
test: strengthen fault-tolerance coverage for daemon/dataflow behavior


heyong4725:
## Summary

Dora documents substantial fault-tolerance behavior, but the current automated coverage is narrower than the feature surface.

This issue strengthens regression coverage for restart, timeout, and recovery behavior.

## Current gap

Existing tests cover:
- restart recovery
- max restart exhaustion
- input timeout

But they do not yet strongly assert the full documented behavior around:
- exact restart policy semantics
- recovery signaling
- health-check-triggered restart paths
- downstream recovery events

## Scope

Add or strengthen tests for:

- `restart_policy: always`
- exact `max_restarts` semantics
- `restart_window` reset behavior
- `health_check_timeout` kill/restart behavior
- `NodeRestarted` delivery to downstream nodes
- `InputClosed` then `InputRecovered`
- stronger assertions around restart exhaustion outcomes

## Notes

Tests should prefer the narrowest useful layer:
- direct daemon/dataflow tests where possible
- full CLI/E2E only when needed

## Deliverables

- expanded fault-tolerance regression coverage
- tests with clear behavioral assertions, not permissive "either outcome is fine" checks
- doc updates if the actual implemented contract differs from the docs

## Acceptance criteria

- documented fault-tolerance behaviors have direct automated assertions where feasible
- regressions in restart/recovery signaling are caught by automated tests
- fault-tolerance coverage is strong enough to support future AI-authored changes in daemon logic

## Tracking

- Parent issue: #1628

heyong4725:
Main scope landed across six PRs:

| Slice | PR |
|---|---|
| `max_restarts` exhaustion → `Err` | #1644 |
| `restart_policy: always` re-spawns on clean exit | #1645 |
| `restart_window` counter reset | #1646 |
| `input_timeout` delivers `InputClosed` | #1647 |
| `health_check_timeout` SIGKILLs unresponsive node | #1648 |
| `NodeRestarted` + `InputRecovered` downstream delivery | #1649 |

10/10 fault-tolerance-e2e tests pass locally; the suite runs in PR CI.

Deferred follow-ups (kept out of scope, tracked locally):
- Full kill → respawn → kill cycle under `health_check_timeout`-induced SIGKILL — the daemon's state machine doesn't cleanly re-enter the watchdog path for the respawned incarnation in my experiments (noted in #1648). Worth its own investigation.
[...]

PR #1645, opened 2026-04-18
test: cover restart_policy: always (#1631)

heyong4725:
## Summary

Second slice of **#1631**. Before this PR there was no automated coverage for `restart_policy: always` — the only restart-policy variant whose core contract (re-spawn on *clean* exit) differs from `on-failure` and `never`.

## Approach

- **New test-only crate** `tests/fault_tolerance/clean_exit_node/`. Appends one `incarnation\n` line to `$DORA_TEST_MARKER_FILE` before returning `Ok(())` after the first tick. Because it always exits cleanly, `restart_policy: never` and `on-failure` would both leave it alone — only `always` re-spawns.
- **New fixture** `tests/dataflows/restart-policy-always.yml`: `restart_policy: always`, `max_restarts: 2`, `restart_delay: 0.1`, 50 ms tick, writes to `/tmp/dora-restart-policy-always.log`.
- **New test** `restart_policy_always_restarts_on_clean_exit` counts lines in the marker file. With `always` + `max_restarts: 2` the daemon spawns the node 3 times (initial + 2 restarts); the assertion is `>=2` to tolerate stop_after racing the last restart on a slow runner. `never`/`on-failure` would produce exactly 1, which the assertion rejects.

## Verification

```
$ cargo test --test fault-tolerance-e2e -- --test-threads=1
test restart_recovers_from_failure ... ok
test max_restarts_limit_reached ... ok
test max_restarts_exhaustion_marks_node_failed ... ok
test restart_policy_always_restarts_on_clean_exit ... ok
test input_timeout_closes_stale_input ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 56.52s
```

`make qa-fast` green (fmt, clippy, audit, unwrap 157/185).

## Remaining under #1631

Each warrants its own slice:

- `restart_window` reset behavior
- `health_check_timeout` kill/restart path
- `NodeRestarted` / `InputClosed` / `InputRecovered` downstream delivery
- Tighter assertion on `input_timeout_closes_stale_input` (currently only checks `result.is_ok()`)

## Test plan

- [x] `cargo test --test fault-tolerance-e2e -- --test-threads=1` ⇒ 5/5 passing locally
- [x] `make qa-fast` green
- [x] `cargo build -p clean-exit-node` builds clean (no warnings)
- [ ] CI green on this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -915,6 +915,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8d4a3bb8b1e0c1050499d1815f5ab16d04f0959b233085fb31653fbfc9d98f9"
 
 [[package]]
+name = "clean-exit-node"
+version = "0.2.1"
+dependencies = [
+ "dora-node-api",
+ "eyre",
+]
+
+[[package]]
 name = "clonable-command"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ members = [
     "libraries/extensions/ros2-bridge/arrow",
     "tests/queue_size_latest_data_rust/receive_data",
     "tests/fault_tolerance/always_crash_node",
+    "tests/fault_tolerance/clean_exit_node",
 ]
 
 [workspace.package]
--- a/tests/dataflows/restart-policy-always.yml
+++ b/tests/dataflows/restart-policy-always.yml
@@ -1,0 +1,11 @@
+nodes:
+  - id: clean-exit
+    build: cargo build -p clean-exit-node
+    path: ../../target/debug/clean-exit-node
+    restart_policy: always
+    max_restarts: 2
+    restart_delay: 0.1
+    env:
+      DORA_TEST_MARKER_FILE: /tmp/dora-restart-policy-always.log
+    inputs:
+      tick: dora/timer/millis/50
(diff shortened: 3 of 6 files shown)


heyong4725:
Review finding:

The new assertion in  is still too permissive to lock in the intended contract. The fixture in  sets , , and , so under the stated semantics this should deterministically produce 3 incarnations within the 10s test window (initial spawn + 2 restarts). But the test in  only asserts .

That means an off-by-one regression where the daemon performs only one clean-exit restart would still pass this test with , even though it violated the advertised  behavior. Unless there is a demonstrated real race in this fixture that can legitimately reduce the count, I think this should be tightened to  so the test actually proves the policy semantics rather than just distinguishing  from  / .

heyong4725:
Review finding:

The new assertion in `restart_policy_always_restarts_on_clean_exit` is still too permissive to lock in the intended contract. The fixture in `tests/dataflows/restart-policy-always.yml` sets `restart_policy: always`, `max_restarts: 2`, and `restart_delay: 0.1`, so under the stated semantics this should deterministically produce 3 incarnations within the 10s test window (initial spawn + 2 restarts). But the test in `tests/fault-tolerance-e2e.rs` only asserts `incarnations >= 2`.

That means an off-by-one regression where the daemon performs only one clean-exit restart would still pass this test with `2`, even though it violated the advertised `max_restarts: 2` behavior. Unless there is a demonstrated real race in this fixture that can legitimately reduce the count, I think this should be tightened to `assert_eq!(incarnations, 3)` so the test actually proves the policy semantics rather than just distinguishing `always` from `never` / `on-failure`.

heyong4725:
Ignore the earlier malformed comment from the shell-escaped attempt. The corrected review finding is here: https://github.com/dora-rs/dora/pull/1645#issuecomment-4274521915

Issue #1654, opened 2026-04-18
ci/python: unblock python-dataflow-builder coverage from local-vs-PyPI package conflict


heyong4725:
## Summary

The Python dataflow builder path is still not covered in CI because local Dora Python packages and PyPI-installed dependencies conflict.

The current CLI workflow leaves the builder example disabled, which means one important Python capability remains effectively untestable in the normal automated lanes.

## Current state

In CI, the Python builder example is commented out with a note that hub-installed packages pull a PyPI `dora-rs` version that overwrites the local build and introduces incompatible types.

This is both:
- a testing gap
- an environment/packaging problem

## Why this matters

- Python builder API is still listed as a coverage gap
- the current packaging story makes the feature harder to trust
- agentic or automated contributors are unlikely to notice the hidden disabled lane

## Scope

Investigate and resolve the local-vs-PyPI package conflict sufficiently to enable automated coverage for the Python builder path.

Possible directions include:
- isolating dependencies differently in CI
- pinning/install order changes
- local package override strategy
- fixture redesign to avoid the conflicting hub dependency path
- packaging compatibility cleanup

## Deliverables

- documented root cause
- working CI/local strategy for exercising the builder example or equivalent builder API coverage
- the disabled builder step is either:
  - re-enabled, or
  - replaced with a deterministic alternative that covers the same capability

## Acceptance criteria

- Python builder coverage is no longer blocked by package-version clobbering
- the repo has an automated validation path for this capability, or a clearly documented narrower follow-up if partial work lands first

PR #1667, opened 2026-04-19
test: unblock python-dataflow-builder coverage via YAML contract (#1654)

heyong4725:
## Summary

Closes **#1654** — the last issue in the 7-issue audit slate.

The existing `simple_example.py` can't run in CI because its `build:` directives pip-install hub packages (`opencv-video-capture`, `dora-yolo`, `opencv-plot`) which transitively pull PyPI `dora-rs` 0.5.0 and clobber the locally-built workspace version. The step was commented out with a TODO to re-enable after 1.0 PyPI publish.

Waiting for 1.0 on PyPI to close this gap leaves the **Python builder API itself** — `dora.builder.DataflowBuilder`, the capability the matrix cares about — untested on every PR in the meantime.

## Approach

Separate the "builder API" capability from the "end-to-end dataflow with hub deps" capability:

- **Builder API** — testable today. Just call the API and assert the generated YAML.
- **End-to-end with hub deps** — blocked on PyPI 1.0. Stays deferred.

## What's new

- **`examples/python-dataflow-builder/test_builder_api.py`** — constructs a two-node `source → sink` dataflow programmatically, calls `.to_yaml(path)`, parses the generated YAML, and asserts:
  - Node IDs present
  - `path:` preserved
  - `inputs` map has expected entries
  - `outputs` map preserved
  - `env` passthrough survives YAML round-trip
  - The output handle from `source.add_output("value")` resolves to `"source/value"` in sink's input wiring (the core builder correctness invariant)
  - Deliberately does NOT call `build()` or `run()`

- **`.github/workflows/ci.yml`** — the disabled `Test Python Dataflow Builder` step is replaced with `Test Python Dataflow Builder API (YAML generation)` invoking the new contract. Fast (~1s), runs on all three platforms via the existing `cli` job. Updated comment explains why `simple_example.py` itself stays deferred.

- **`docs/testing-capabilities.md`** — Python operator path section gains two rows:
  - Builder API (programmatic dataflow construction → YAML): PR / Contract
  - `simple_example.py` end-to-end with hub packages: Gap (infrastructural — re-enable post-PyPI-1.0)

  Open gaps index narrowed accordingly.

## Verification

```
$ python3 examples/python-dataflow-builder/test_builder_api.py
python-dataflow-builder: SUCCESS - generated YAML has expected structure
```

`make qa-fast` green.

## Test plan

- [x] Local run passes
- [x] CI YAML still parses
- [x] `typos` clean
- [x] Capability matrix reflects the real (not aspirational) coverage
- [ ] CI green on all 3 platforms (the `cli` job runs on ubuntu/macos/windows)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -416,7 +416,13 @@ jobs:
       # runs on every PR.
       - name: Test Python Dataflow Builder API (YAML generation)
         timeout-minutes: 5
-        run: python examples/python-dataflow-builder/test_builder_api.py
+        shell: bash
+        # `uv run` is needed so the step uses the project venv from
+        # the "Set up Python venv" step above — `source .venv/bin/activate`
+        # doesn't persist to later steps. `--with PyYAML` provisions
+        # the one extra dep the contract test needs, without polluting
+        # the dora-rs install with a transitive upgrade from PyPI.
+        run: uv run --with PyYAML python examples/python-dataflow-builder/test_builder_api.py
 
       - name: Python Queue Latency Test
         timeout-minutes: 30
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -404,13 +404,19 @@ jobs:
           OTEL_SDK_DISABLED: true
         run: dora run examples/python-drain/dataflow.yaml --uv
 
-      # TODO: re-enable once dora 1.0 is published to PyPI (supersedes upstream 0.5.0).
-      # The builder example installs dora-hub packages (opencv-video-capture,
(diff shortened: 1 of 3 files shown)


heyong4725:
1. High: the new CI step is very likely broken because it runs plain `python` without the venv/package context that provides `dora.builder`. The new step in `.github/workflows/ci.yml` does `python examples/python-dataflow-builder/test_builder_api.py`, but the earlier setup only installs `apis/python/node` into the `uv` venv during a prior shell step and does not persist the activated interpreter onto later steps. Locally, the same command fails with `ModuleNotFoundError: No module named 'dora'`, which is consistent with this wiring. This should probably use `uv run python ...` or otherwise ensure the step runs inside the environment where `dora-rs` was installed.

Issue #1669, opened 2026-04-20
test: rename or delete legacy input_timeout_closes_stale_input (#1631 follow-up)


heyong4725:
## Summary

Follow-up from #1631. The pre-existing test `input_timeout_closes_stale_input` in `tests/fault-tolerance-e2e.rs` **claims to cover** `input_timeout` semantics ("The sink has `input_timeout: 2.0` on the status-node output. After the timeout, the sink receives `InputClosed`…"), but the fixture's sink exits on the first `exit` signal at ~500 ms — well before the 2 s `input_timeout` could possibly fire. The `input_timeout` config in that fixture is effectively dead.

PR #1647 added real coverage via `input_timeout_delivers_input_closed_to_downstream` using a silent-source + event-log-observer fixture. That test actively drives the timeout path and asserts the `InputClosed` event lands in the observer log.

The legacy test should be either:

- **Renamed** to reflect what it actually proves (that a dataflow with `input_timeout` configured runs to completion without the config-parser rejecting the field), or
- **Deleted** as superseded by #1647's real coverage.

## Recommendation

Delete. The real contract is already covered by #1647, and the remaining "does the parser accept the field" assertion is already covered by `dora validate` in PR CI. Keeping a misleadingly-named test is worse than not having it.

## Scope

Single commit: delete the test function + its fixture YAML `tests/dataflows/input-timeout.yml` if nothing else references it.

## Parent

#1631.

PR #1670, opened 2026-04-20
test: delete legacy input_timeout_closes_stale_input (#1669)

heyong4725:
## Summary

Closes **#1669**. The `input_timeout_closes_stale_input` test claimed to cover `input_timeout` semantics but its fixture's sink exited on the first `exit` signal at ~500 ms — well before the 2 s `input_timeout` could fire. The config was dead and the only real assertion was `result.is_ok()`, i.e. "the dataflow completes at all."

Real coverage landed in **PR #1647** as `input_timeout_delivers_input_closed_to_downstream`, which drives the timeout path directly via a silent-source + event-log-observer fixture and asserts both `Input:value` and `InputClosed:value` markers in the observer log.

## Changes

- Remove `input_timeout_closes_stale_input` (test function, its body, its comment block)
- Delete fixture `tests/dataflows/input-timeout.yml` (no remaining references)
- Update docstring on the replacement test in `fault-tolerance-e2e.rs` to describe the legacy as removed (past tense)
- Update `docs/testing-guide.md`, `guide/src/development/testing.md`, `docs/plan-fault-injection.md` to reference the current test name
- Update crate-level doc comment in `tests/fault_tolerance/silent_source_node/Cargo.toml` to reference the renamed fixture

## Scope note (per review feedback)

`guide/po/messages.pot` and `guide/po/zh-CN.po` still contain the old `input_timeout_closes_stale_input` string. These are generated translation catalogs; regenerating them here would produce ~4000 lines of reshuffle churn from 3+ weeks of accumulated source drift (`POT-Creation-Date` last bumped 2026-03-25). Tracked for a separate refresh PR. Source markdown (`guide/src/development/testing.md`) is already updated in this PR.

## Verification

```
$ cargo test --test fault-tolerance-e2e -- --test-threads=1
…
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 105.23s
```

(Was 10 before; removing one legacy = 9 remaining. No regressions in the remaining set — `input_timeout_delivers_input_closed_to_downstream` continues to pass.)

`make qa-fast` green.

## Test plan

- [x] Remaining 9 fault-tolerance-e2e tests still pass
- [x] `grep -rn input_timeout_closes_stale_input` — only historical/deletion markers + generated .po/.pot remain
- [x] `grep -rn input-timeout.yml` — zero references
- [x] `make qa-fast` green
- [ ] CI green


heyong4725:
1. Medium: the cleanup still leaves stale references to the deleted legacy path in tracked docs/artifacts, so the repo is not actually free of `input_timeout_closes_stale_input` / `input-timeout.yml` references yet. The old test name still appears in `guide/po/messages.pot` and `guide/po/zh-CN.po`, and the fixture name still appears in the comment in `tests/fault_tolerance/silent_source_node/Cargo.toml`. Since the PR description explicitly claims the old references are gone except for historical/deletion markers, I’d either update these too or narrow the claim.

heyong4725:
Good catch. Addressed in two parts:

1. **Cargo.toml comment** (89161fd9) — `tests/fault_tolerance/silent_source_node/Cargo.toml` now points at `input-timeout-delivery.yml` (the current fixture name).

2. **`guide/po/*`** — deliberately *not* updating these here. Regenerating `messages.pot` with `mdbook-i18n-helpers` + `msgmerge` produced ~4000 lines of mechanical reshuffle because `POT-Creation-Date` had drifted since 2026-03-25. That's too much unrelated noise for this PR. Will file a separate refresh PR. Updated the PR description to narrow the claim — it now calls out the .po/.pot files as a known stale-reference exception alongside the historical/deletion markers.

PR #1674, opened 2026-04-20
fix(cli): dora record passes source dir as Run working_dir override (#1673)

### LINKED ISSUE #1673: dora record: cargo can't find Cargo.toml because temp YAML lives in /tmp
## Symptom

`dora record <dataflow.yml>` fails with:

```
error: could not find `Cargo.toml` in `/tmp` or any parent directory
```

Reproduces on any example whose nodes declare `build: cargo build -p X` directives. The nightly `record-replay` subjob has been red on this since 2026-04-18 (discovered during PR #1660 investigation).

## Root cause

`binaries/cli/src/command/record.rs` (before this fix) writes the record-node-augmented YAML to a `tempfile::NamedTempFile` in the **system temp dir** (`/tmp` on Linux, equivalent on macOS/Windows), then hands that path to `Run::execute()`.

`Run::execute()` derives `working_dir` from `dataflow_path.parent()` (see `binaries/cli/src/command/run.rs:181-183`). So the cargo invocations that `build:` directives emit run from `/tmp`, which has no `Cargo.toml` — cargo exits 101.

The bug affects not just cargo directives — any relative-path node reference (e.g. Python `path: sender.py`) would silently misbehave for the same reason.

## Fix

Write the temp file to the dataflow YAML's parent directory instead of the system temp dir. `tempfile::Builder::tempfile_in(&source_dir)` is a drop-in replacement that preserves the auto-cleanup-on-drop semantics and uses a clearly marked `.dora-record-*.yml` filename.

## Evidence of workaround

`tests/example-smoke.rs:912-917` already contains a workaround comment:

> *"Emit a test fixture YAML for validated-pipeline that uses absolute paths and no `build:` directives, so `dora record` / `dora replay` (which write their modified YAML into a `/tmp` tempfile) don't trip over `cargo build -p …` failing from a directory with no Cargo.toml in its parent chain."*

The workaround stays in place — the record/replay equivalence test uses absolute-path fixtures by design. But new tests won't need that workaround after the fix lands.

## Regression coverage

The nightly `record-replay` job exercises exactly this path. It should go green on the next nightly after the fix lands.

heyong4725:
Closes #1673.

## Summary

`dora record` writes its record-node-augmented descriptor to a tempfile before invoking `Run::execute()`. `Run` derives `working_dir` from `dataflow_path.parent()`, so when the tempfile sat in `/tmp` the nodes' `build: cargo build -p X` directives ran from `/tmp` and hit:

```
error: could not find Cargo.toml in /tmp
```

## Fix (v2, per review on the first revision)

Keep the tempfile in the system temp dir (no new write-permission requirement on the source directory) and add an explicit `working_dir: Option<PathBuf>` override on `Run` (+ builder method `with_working_dir`). `dora record` constructs `Run` with the original dataflow's parent as the override, so module expansion, `build:` cargo invocations, and node binary resolution all resolve against the **source** directory while the rewritten YAML continues to live in `/tmp`.

Threaded the override through `dora_cli::build()` (new trailing `working_dir_override` parameter, `None` at existing call sites). Python bindings (`apis/python/cli`, `apis/python/node`) updated to pass `None`.

## Why this shape

- `Run::execute()`'s two `dataflow_path.parent()` sites (one for descriptor expand, one for `local_working_dir` sent to the coordinator) now use the override when set. No behavior change for the `dora run` CLI path — it never sets the override.
- The read-only source case from the review (`dora record /some/read-only/dataflow.yml`) keeps working because we don't try to write alongside it.
- `dora build` subcommand's CLI surface is unchanged; its `Executable::execute` impl passes `None`.

## Verification

```
$ target/debug/dora record examples/rust-dataflow/dataflow.yml -o /tmp/rd-test2.drec
…
09:43:48 INFO    [dora]: coordinator  dataflow finished
$ ls -la /tmp/rd-test2.drec
-rw-r--r-- 1 yonghe wheel 58207 Apr 20 09:43 /tmp/rd-test2.drec
```

58KB of real recording output. Before the fix this would have failed at cargo build for `rust-dataflow-example-node`.

- `cargo check --all --exclude <python packages>` — clean
- `cargo fmt --all --check` — clean
- `cargo clippy -p dora-cli -p dora-cli-api-python -p dora-node-api-python -- -D warnings` — clean

## Regression coverage

The nightly `record-replay` subjob (red since 2026-04-18) exercises exactly this path and should go green on the next scheduled nightly after this lands.

## Test plan

- [x] `dora record` on a node with `build: cargo build -p …` succeeds end-to-end and produces a non-empty .drec
- [x] `dora record --output-yaml <path>` (bypass-run mode) still works (untouched code path)
- [x] `dora run <file>` behavior unchanged (override is `None` by default)
- [x] Python `dora_cli.build(...)` / `dora.build(...)` still compile and run (new trailing `None` threaded through)
- [x] fmt + clippy green
- [ ] CI green
- [ ] Next nightly's `record-replay` subjob turns green

### DIFF
--- a/apis/python/cli/src/lib.rs
+++ b/apis/python/cli/src/lib.rs
@@ -41,6 +41,7 @@ pub fn build(
         false,
         None,
         false,
+        None,
     )
 }
 
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -812,6 +812,7 @@ pub fn build(
         false,
         None,
         false,
+        None,
     )
 }
 
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -122,6 +122,7 @@ impl Executable for Build {
             self.write_lockfile,
             self.lockfile,
             self.parallel,
+            None,
         )
     }
 }
@@ -138,14 +139,18 @@ pub fn build(
     write_lockfile: bool,
     lockfile_override: Option<PathBuf>,
     parallel: bool,
+    working_dir_override: Option<PathBuf>,
 ) -> eyre::Result<()> {
     let dataflow_path = resolve_dataflow(dataflow).context("could not resolve dataflow")?;
     if lockfile_override.is_some() && !(locked || write_lockfile) {
         eyre::bail!("`--lockfile` requires either `--locked` or `--write-lockfile`");
     }
-    let working_dir = dataflow_path
-        .parent()
-        .unwrap_or_else(|| std::path::Path::new("."));
+    let working_dir = match working_dir_override.as_deref() {
+        Some(p) => p,
+        None => dataflow_path
+            .parent()
+            .unwrap_or_else(|| std::path::Path::new(".")),
+    };
     let dataflow_descriptor = Descriptor::blocking_read(&dataflow_path)
         .wrap_err_with(|| {
             format!(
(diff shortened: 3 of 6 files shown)


heyong4725:
1. Medium: this fixes the `Cargo.toml` lookup bug by requiring write access to the source descriptor directory, which is a new failure mode for otherwise valid read-only dataflow paths. In `binaries/cli/src/command/record.rs` the temp YAML is now created with `tempfile_in(&source_dir)`. That works for checked-out examples, but it means `dora record /some/read-only/dataflow.yml ...` will now fail before execution even if the original descriptor was perfectly usable. Since the real invariant here is “run with the source dataflow’s working directory,” not “write next to the source file,” I think this should either preserve the original working dir separately or at least call out the new write-permission requirement as a regression/tradeoff.

heyong4725:
Agreed — the new write-permission requirement was a regression I missed. Refactored in 87c22439 to match your framing: "the real invariant here is 'run with the source dataflow's working directory,' not 'write next to the source file.'"

Concrete changes:

1. Added `working_dir: Option<PathBuf>` to `Run` + `Run::with_working_dir()` builder method. When set, it overrides `dataflow_path.parent()` at the two call sites in `Run::execute()` (descriptor expand + `local_working_dir` for coordinator start).
2. Threaded a `working_dir_override: Option<PathBuf>` through `dora_cli::build()` so the cargo invocation for `build:` directives also picks up the override. Both `Run::execute()`'s internal `build_dataflow(…)` call and the public `Build::execute()` CLI path updated; Python bindings pass `None`.
3. `record.rs` now: writes to `/tmp` as before, then constructs `Run::new(tmp_path).with_working_dir(source_dir)`. No write access to the source directory needed.

Re-verified with `examples/rust-dataflow`: 58KB `.drec` produced end-to-end. Title + description updated to reflect the new shape.

Issue #1683, opened 2026-04-20
dora stop: `--all` flag and bare-name `stop <name>` both rejected despite being in docs/tests


heyong4725:
## Symptom

`dora stop <name>` and `dora stop --all` both reject, despite both being advertised in user-facing documentation.

```
$ dora stop demo
error: invalid value 'demo' for '[UUID]': invalid character: found `m` at 3

$ dora stop --all
error: unexpected argument '--all' found
```

Only `dora stop <UUID>` and `dora stop --name <NAME>` / `dora stop -n <NAME>` are accepted. `dora --help` for the `stop` subcommand confirms — `--all` is not a declared flag.

## Where the gap shows up

1. **`examples/dynamic-add-remove/README.md:46`** — "Step 5: Clean up" instructs `dora stop --all`.
2. **`examples/dynamic-agent-tools/README.md:51`** — same line, "Step 4: Clean up" instructs `dora stop --all`.
3. **`tests/example-smoke.rs:136,238`** — both `cleanup_stale` and the poll loop call `Command::new(dora).args(["stop", "--all"])` and silently discard the result (`let _ = ...`). The cleanup "succeeds" only because `dora down` at line 146 does the real teardown. So the test suite ships with a broken cleanup step that's been masking.
4. Many example READMEs likely reference `dora stop <name>` bare-form; that gets rejected too.

## Expected behavior

Pick one (or both) of:

**A. Make `--all` a real flag.** Stops every running dataflow on the coordinator. Mirrors `dora node list` / `dora list`'s "everything" semantics.

**B. Accept `dora stop <NAME>` in addition to `dora stop <UUID>`.** Currently the positional `[UUID]` argument has a strict UUID parser. Loosen to accept either a UUID or a name lookup (mirrors how `dora node list --dataflow <NAME_OR_UUID>` already works at `binaries/cli/src/command/node/list.rs`).

**C. Fix the docs to match the CLI** — update the two READMEs and the test cleanup. Least-work option but doesn't fix the UX papercut.

Recommendation: B + C. Name-or-UUID is already the idiom elsewhere in the CLI; the inconsistency is the bug. The `--all` form is less critical given that most users clean up a specific named dataflow, not every one.

## Test gap exposed

[...]

Issue #1682, opened 2026-04-20
dora node add times out + corrupts dataflow state (dynamic-topology CLI broken)


heyong4725:
## Symptom

Walking through [`examples/dynamic-add-remove/README.md`](https://github.com/dora-rs/dora/blob/main/examples/dynamic-add-remove/README.md) step-by-step against a locally-built `dora` binary reproduces a daemon timeout at the first `dora node add`, and corrupts the dataflow state so subsequent commands can't find it:

```
$ dora up
$ dora build  examples/dynamic-add-remove/dataflow.yml --uv
$ dora start  examples/dynamic-add-remove/dataflow.yml --detach --name demo --uv
$ dora node list --dataflow demo
NODE      STATUS   PID    CPU    MEMORY  RESTARTS
receiver  Running  46258  0.1%   58 MB   0
sender    Running  46264  10.1%  103 MB  0
$ dora node add --from-yaml examples/dynamic-add-remove/filter-node.yml --dataflow demo
[ERROR]
daemon dispatch failed: timeout waiting for daemon WS reply

Location:
    binaries/cli/src/common.rs:68:48

$ dora node list --dataflow demo
NODE  STATUS  PID  CPU  MEMORY  RESTARTS    # <-- empty! dataflow now unreachable
$ dora node connect --dataflow demo sender/value filter/input
  hint: use `dora list` to see running dataflows    # <-- can't find demo anymore
```

Same failure reproduces on `examples/dynamic-agent-tools` at the equivalent step.

## Where the handlers live

- **CLI** — `binaries/cli/src/command/node/add.rs` (sends `ControlRequest::AddNode`)
- **Coordinator** — `binaries/coordinator/src/lib.rs:1340-1395` — forwards as `DaemonCoordinatorEvent::AddNode`, awaits `conn.send_and_receive(&msg)` with no explicit timeout. Times out before returning.
- **Daemon** — `binaries/daemon/src/lib.rs:1546-1620` — does attempt to spawn the node via `Spawner::spawn_node` + `prepared.spawn(logger).await`. The attempt appears to hang / not reply, which corrupts dataflow state since subsequent reads return empty.

Suspicions (haven't bisected):
1. The spawn's `.await` on `uv` cold-boot is slower than the CLI WS reply timeout.
2. A reply channel isn't wired (coordinator awaits a daemon reply that the daemon never sends — the handler's `Ok(())` return doesn't flow back to `send_and_receive`).
[...]

PR #1684, opened 2026-04-20
docs(tests): table of examples not covered by example-smoke.rs

heyong4725:
Adds a single-source-of-truth table of examples that `tests/example-smoke.rs` deliberately doesn't cover, with the reason for each.

## Motivation

During a Phase 5b QA sweep (user asked "for all examples did we build them in our QA tests?"), I traced coverage for all 40 directories under `examples/`. Three classes of uncovered example emerged:

1. **Real bugs** — `dynamic-add-remove` and `dynamic-agent-tools`. Walking the README manually hits `dora node add: daemon dispatch failed: timeout waiting for daemon WS reply` and corrupts the dataflow state. Filed as #1682.
2. **External runtime deps** — `cuda-benchmark` (needs GPU), `ros2-bridge` / `ros2-comparison` (need ROS2 runtime), `python-yolo-detection` (needs YOLO weights download).
3. **Covered elsewhere** — `c-dataflow`, `c++-*-dataflow`, `cmake-dataflow`, `multiple-daemons`, `rust-dataflow-git`. These DO compile/run in CI, just not in `tests/example-smoke.rs` (they're in the `cli` job matrix or `examples` job).

The shell helper `scripts/smoke-all.sh` had partial `log_skip` entries for some of these, but `tests/example-smoke.rs` — which is what CI actually runs — had no record.

## Change

One trailing `//` comment block at the bottom of `tests/example-smoke.rs` with a table: example name, blocker, tracking issue.

No test behavior change. 52 tests still enumerate (`cargo test --test example-smoke -- --list | tail`).

## Related

- Filed #1682 — dora node add times out + corrupts dataflow state.
- Filed #1683 — dora stop: `--all` flag and bare-name `<name>` both rejected despite being in docs/tests.

## Test plan

- [x] `cargo test --test example-smoke -- --list` still shows 52 tests
- [x] `cargo fmt --all --check` green
- [ ] CI green

### DIFF
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -1327,31 +1327,46 @@ fn smoke_local_queue_size_latest_data_rust() {
 }
 
 // ---------------------------------------------------------------------------
-// Examples under `examples/` that are deliberately NOT covered by CI smoke
-// tests. Keep this table in sync with `scripts/smoke-all.sh` so the shell
-// helper and the Rust suite agree on what's skipped and why.
+// Examples under `examples/` that do NOT have a corresponding `smoke_*` or
+// `contract_*` test in this file. Some are blocked (filed issue or external
+// dep); others are intentionally covered by a DIFFERENT CI job. Keep this
+// table in sync with `scripts/smoke-all.sh` and the CI workflows it cites,
+// so the Rust suite, the shell helper, and CI all agree on what's covered
+// where and why.
+//
+// Audit method: diff `ls examples/` against the `smoke_*`/`contract_*`
+// function list in this file. Every example without a match here must have
+// a row below — either as a blocker or as "covered elsewhere" with a link
+// to the specific CI line that covers it.
 //
 // If you land a fix for a blocker below, write the corresponding smoke test
 // here and remove the row. Consumers of this file: the QA policy in
 // `docs/agentic-qa-policy.md` and the capability matrix in
 // `docs/testing-capabilities.md`.
 //
-// | Example                   | Blocker                        | Tracking |
-// |---------------------------|--------------------------------|----------|
-// | cuda-benchmark            | needs NVIDIA CUDA toolkit      | —        |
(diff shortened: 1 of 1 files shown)


heyong4725:
1. Medium: the new “single source of truth” table looks incomplete, so it risks becoming misleading immediately. The PR description says it inventories examples intentionally not covered by `tests/example-smoke.rs`, but that set is broader than the three classes listed here. At minimum, `examples/error-propagation` is covered only in the `cli` job at `.github/workflows/ci.yml:327`, `examples/cpu-affinity-probe` is covered only by nightly `cpu-affinity-smoke` at `.github/workflows/nightly.yml:540`, and `examples/python-dataflow-builder` is covered by the builder contract in `.github/workflows/ci.yml:433` plus `docs/testing-capabilities.md:171`. If those examples are not in the new comment block in `tests/example-smoke.rs`, then the table is not actually a complete source of truth for “examples not covered by example-smoke.rs”, only a subset.

heyong4725:
Good catch — the table was a subset, not complete. Fixed in `ad0f2714`.

Re-audited by diffing `ls examples/` against the `smoke_*` / `contract_*` function list in this file. Added 4 missing rows, each citing the exact workflow line that covers the example:

| Example | Covered at |
|---|---|
| `cpu-affinity-probe` | `.github/workflows/nightly.yml:535` (cpu-affinity-smoke) |
| `error-propagation` | `.github/workflows/ci.yml:327` (cli job) |
| `python-dataflow-builder` | `.github/workflows/ci.yml:433` (builder contract) |
| `python-operator-dataflow` | `.github/workflows/ci.yml:393` (cli job) |

Also reshaped the header comment to spell out the audit method so future maintainers reproduce it: diff the example dir against this file's `smoke_*`/`contract_*` functions, and every missing example must have a row. Now the table enumerates all 40 example dirs — 27 with a test in this file, 13 with a row explaining where they're tested instead (or why they're blocked).

PR #1687, opened 2026-04-20
fix+feat(scripts): smoke-all.sh bash 3.2 compat + live progress

heyong4725:
### 1. bash 3.2 compat (original fix)

`./scripts/smoke-all.sh` on macOS (default shell = bash 3.2) died at the first networked example:

```
./scripts/smoke-all.sh: line 89: uv_args[@]: unbound variable
```

`set -euo pipefail` + bash 3.2's handling of `"${arr[@]}"` on empty arrays = nounset error. Fixed at three call sites by switching to the conditional-expansion idiom `${arr[@]+"${arr[@]}"}` (which expands to zero args when empty, unlike `"${arr[@]:-}"` which injects an empty-string arg).

### 2. Live progress + failure diagnostics

Before: 28+ examples ran silent. Each `=>` line was followed by ~30s of nothing before PASS/FAIL. Dev couldn't tell if it was making progress, stuck on one step, or hanging.

Now each step prints live:

```
=> rust-dataflow (networked, 30s)
  up ok |  build ok |  start ok | running 2s/30s
  PASS: rust-dataflow
```

The `running Ns/Ms` portion uses `\r` to rewrite itself in place during the poll loop, so a developer sees the count tick up but the final output compacts to a single line.

Per-example dora output is captured to a `mktemp` file. On FAIL, the last 20 lines are dumped indented under the fail line, so the developer sees WHY without re-running in verbose mode. Example:

```
=> rust-dataflow (networked, 30s)
  up ok |  build ok |  start FAIL
    ---- last 20 lines ----
    | [ERROR]
    | cannot connect to coordinator at 127.0.0.1:6013: IO error: Connection refused
    | ...
  FAIL: rust-dataflow (dora start failed)
```

New `-v` / `--verbose` flag streams dora's stdout/stderr live via `tee` instead of capturing silently. Useful when a step hangs.

`run_local` gets the same treatment: `running Ns/Ms (hard-kill Ks)` heartbeat + tail-on-failure.

## No behavior change

Pure UX / diagnostics. PASS/FAIL logic identical. No change to what runs.

## Verification

On macOS bash 3.2:

- Before bash fix: died at line 89 with "unbound variable."
- After bash fix + progress: rust-dataflow, rust-dataflow-dynamic, rust-dataflow-url, benchmark, log-sink-{file,alert,tcp}, service-example all PASS with visible progress.
- Failure output (on prior stale coordinator state) dumps the relevant `[ERROR]` lines inline, cause obvious.

## Test plan

- [x] Script runs end-to-end on macOS bash 3.2
- [x] Progress lines visible in a TTY; compact to PASS/FAIL in non-TTY pipelines (`\r` semantics)
- [x] Failure output dumps captured log tail
- [x] `--verbose` streams live dora output
- [ ] CI green

### DIFF
--- a/scripts/smoke-all.sh
+++ b/scripts/smoke-all.sh
@@ -8,9 +8,10 @@
 #   -> dora stop -> dora destroy -> dora down
 #
 # Usage:
-#   ./scripts/smoke-all.sh              # Run all examples
-#   ./scripts/smoke-all.sh --rust-only  # Rust examples only
+#   ./scripts/smoke-all.sh               # Run all examples
+#   ./scripts/smoke-all.sh --rust-only   # Rust examples only
 #   ./scripts/smoke-all.sh --python-only # Python examples only
+#   ./scripts/smoke-all.sh --verbose     # Stream dora stdout/stderr live
 #
 # Prerequisites: cargo, Python 3 with pyarrow + numpy installed.
 # Skips examples that need webcam, CUDA, ROS2, C/C++ toolchain, etc.
@@ -27,13 +28,15 @@ SKIP=0
 FAILURES=()
 RUN_RUST=true
 RUN_PYTHON=true
+VERBOSE=false
 
 for arg in "$@"; do
     case "$arg" in
-        --rust-only)  RUN_PYTHON=false ;;
+        --rust-only)   RUN_PYTHON=false ;;
         --python-only) RUN_RUST=false ;;
+        -v|--verbose)  VERBOSE=true ;;
         -h|--help)
-            echo "Usage: $0 [--rust-only] [--python-only]"
+            echo "Usage: $0 [--rust-only] [--python-only] [-v|--verbose]"
             exit 0
             ;;
         *) echo "Unknown option: $arg"; exit 1 ;;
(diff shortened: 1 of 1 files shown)


heyong4725:
1. Medium: the new UX still doesn’t satisfy the “manual runner can see example output” requirement unless they already know to opt into `--verbose`. In `scripts/smoke-all.sh:49` the default path still captures `dora` stdout/stderr into a temp file, and in `scripts/smoke-all.sh:112` / `scripts/smoke-all.sh:158` only step status, heartbeat, and failure tails are shown. That is good progress feedback, but a human manually running smokes still cannot see the example’s live logs or final output by default. If the goal is “feedback of test running and example output,” I’d either make output streaming the default for manual runs, or at least print a short explicit hint at startup that `--verbose` is the mode for live example output.

heyong4725:
Good point. Flipped in `f6ca4c3e`: `--show-output` behavior is now the **default**. A human running the script out of the box sees:

```
(tip: -v streams dora live; -q suppresses the per-example tail)
...
=> rust-dataflow (networked, 30s)
  up ok |  build ok |  start ok | running 2s/30s
    ---- last 15 lines ----
    | dataflow start triggered: 019dad23-7ebc-77f3-be54-78ba9e19181b
    | 17:58:48 INFO    rust-node: spawner  spawning ...
    | 17:58:48 INFO    rust-sink: spawner  spawning ...
    | dataflow started: 019dad23-7ebc-77f3-be54-78ba9e19181b
    --------------------------
  PASS: rust-dataflow
```

Flag semantics after the change:
- **Default** — step progress + per-example tail. What you get with no flags.
- **`-v`/`--verbose`** — stream everything live (for debugging hangs).
- **`-q`/`--quiet`** — suppress the per-example tail, keep only step progress + PASS/FAIL.
- **`-s`/`--show-output`** — kept for back-compat / symmetry, now a no-op since it's already on.

Startup hint line ("tip: …") tells developers about the escape hatches without needing `--help`.

Still no CI impact — `scripts/smoke-all.sh` remains a local-only tool.

PR #1699, opened 2026-04-21
feat(qa): add make qa-examples target (wraps scripts/smoke-all.sh)

heyong4725:
## Motivation

Every `make qa-*` target explicitly passes `--exclude dora-examples` to `cargo test`, so `tests/example-smoke.rs` (which drives 28+ dataflow runs) never executes from the qa-fast/full/deep/nightly/release-gate ladder. That's **intentional** — bundling example runs would turn "~15 min qa-deep" into "~30+ min" and blow the per-push budget.

But it leaves developers without a one-shot target that ACTUALLY exercises example dataflows after touching node/operator APIs, CLI subcommands, or the descriptor surface. `scripts/smoke-all.sh` has always been the right tool for this, but didn't have a `make` handle consistent with the other qa-* targets.

## Change

New **orthogonal** target (not a rung on the thoroughness ladder):

```
make qa-examples                       # full run, ~15-20 min
make qa-examples ARGS="--rust-only"    # Rust examples only
make qa-examples ARGS="-v"             # stream dora output live (debugging)
```

Delegates to `scripts/smoke-all.sh` with flag passthrough via `ARGS=`.

## Where it sits in the ladder

Deliberately **not** on the fast → full → deep → nightly → release-gate thoroughness spine. Documented in the Makefile header under a new "Orthogonal to the ladder" section with the reason spelled out: bundling example runs into `qa-deep` would break its time budget.

## Docs swept

- `CLAUDE.md` "deeper QA gates" section
- `docs/qa-runbook.md` TL;DR + target table
- `docs/testing-matrix.md` target table
- `docs/contributor-qa-cheatsheet.md` new section explaining *when* to run it
- `README.md` QA block

## No behavior change to anything else

`scripts/smoke-all.sh` unchanged — it's already the single source of truth for local example dataflow runs and remains so. `qa-examples` just adds a Makefile handle matching the pattern of the other qa-* targets.

## Test plan

- [x] `make -n qa-examples` → `scripts/smoke-all.sh`
- [x] `make -n qa-examples ARGS="--rust-only -v"` → `scripts/smoke-all.sh --rust-only -v`
- [x] Ladder comment in Makefile explains why this is orthogonal
- [x] Docs updated across 5 surfaces
- [ ] CI green


heyong4725:
1. Low-medium: the new docs overstate what `qa-examples` does. The Makefile header and the swept docs describe it as running “every example dataflow end-to-end,” but `scripts/smoke-all.sh:430` still deliberately skips a non-trivial set of examples: `cuda-benchmark`, `ros2-bridge`, `multiple-daemons`, `rust-dataflow-git`, C/C++ examples, `python-dataflow-builder`, `dynamic-add-remove`, `dynamic-agent-tools`, and others. Since `qa-examples` is just a wrapper around that script, I’d soften the wording to something like “run the local smoke-all example suite” or “run all smoke-eligible example dataflows end-to-end,” otherwise readers will assume broader coverage than the target actually provides.

heyong4725:
Fair catch. Fixed in `29dc9855`.

You're right — `scripts/smoke-all.sh` has 14 `log_skip` entries:

```
cross-language, python-dataflow-dynamic, python-operator-dataflow,
cuda-benchmark, ros2-bridge, multiple-daemons, rust-dataflow-git,
c-dataflow, c++-dataflow, c++-arrow-dataflow, cmake-dataflow,
python-dataflow-builder, dynamic-add-remove, dynamic-agent-tools
```

Calling `qa-examples` "every example dataflow end-to-end" would mislead anyone reading the docs into expecting broader coverage than exists. Reworded across 6 surfaces (Makefile header, Makefile target comment, README.md, CLAUDE.md, qa-runbook, testing-matrix, contributor-cheatsheet) to:

- "Run all **smoke-eligible** example dataflows end-to-end"
- Explicit "skips examples that need CUDA, ROS2, webcam, multi-machine deploy, C/C++ toolchains, or interactive CLI"
- Pointer to `scripts/smoke-all.sh -h` as the source of truth for the SKIP list

Zero behavior change — wording only.

Issue #1710, opened 2026-04-21
PyPI dora-rs drift breaks smoke_local_python_concurrent_rw (message format v0.8.0 vs workspace v0.2.1)


heyong4725:
## Problem

`tests/example-smoke.rs::smoke_local_python_concurrent_rw` fails locally with:

```
2: version mismatch: message format v0.8.0 is not compatible with expected message format v0.2.1
```

The example's two Python nodes (`examples/python-concurrent-rw/receive_data.py`) do `from dora import Node`. When the test runs via `dora run --uv`, uv spawns an isolated venv and installs `dora-rs` from PyPI:

```
dora-rs = 0.5.0  (PyPI latest)
 └─ bundles message format v0.8.0
```

Meanwhile the local workspace daemon speaks:

```
dora-message v0.2.1  (workspace)
```

These message formats are binary-incompatible. The Python node connects, tries to register, and the daemon rejects it.

## Why now

Surfaced by running `make qa-nightly` (which as of #1709 runs `cargo test -p dora-examples --test example-smoke -- --test-threads=1`, covering all 52 smoke tests). 51 pass; this one fails every time.

It's also not new — the error isn't caused by any recent change. The drift has existed since the workspace diverged from the published PyPI release. CI smoke-suite has been hitting its 30-min timeout before reaching this test on recent nightly runs, so the failure was invisible in CI.

## Root cause

The workspace and PyPI are two separate release trains:

- **Workspace `dora-rs` Python bindings** — in `apis/python/node/`, built with maturin, version tracks workspace `0.2.1`, speaks message format `v0.2.1`.
- **PyPI `dora-rs`** — published independently at `0.5.0`, speaks message format `v0.8.0`.

When `--uv` creates an isolated venv, it goes to PyPI. The ambient venv on GHA CI has `uv pip install -e apis/python/node` (local bindings) installed, but `--uv` doesn't inherit from it.

## Fix options

1. **Publish workspace bindings to PyPI** at a version that the examples can pin. Real fix; requires release coordination, outside the scope of a smoke-test unblock.
2. **Stop running `--uv` for this example** — override `needs_uv()` so it uses the ambient Python. Fragile (cross-cutting smoke helper), and the ambient venv only has the bindings when a human has set it up.
3. **Pin `dora-rs==<workspace>` in the example's `pyproject.toml`** — no such PyPI version exists that matches.
[...]

PR #1709, opened 2026-04-21
feat(qa): full qa-nightly parity with .github/workflows/nightly.yml (#1707) [for review]

heyong4725:
Refs #1707.

> **NOTE:** This supersedes #1708, which was admin-merged prematurely. That merge was reverted in `c0a65d61`. This PR is the same change, open for review.

## Summary

After #1706, `make qa-nightly` covered **5 of 11** GHA nightly test jobs via `tests/example-smoke.rs` (smoke-suite, log-sinks, service-action, streaming, record-replay). The other 6 lived only in CI and couldn't be verified locally.

This PR adds `scripts/qa/ci-nightly-jobs.sh` which drives all 6 locally, mirroring the GHA step bodies exactly. Wired into `scripts/qa/all.sh --nightly` as step 14.

## Jobs now runnable locally

| Job | What it verifies | Linux-only |
|---|---|---|
| `cluster-smoke` | `dora up` + `cluster status` (header + ≥1 daemon) + `start --detach` + `cluster down` + no leftover procs | No |
| `topic-and-top-smoke` | `dora top/trace/topic/self update` against a 10 Hz Python ticker fixture (#238) | No |
| `cpu-affinity-smoke` | `AFFINITY_MASK:0,1` via `cpu-affinity-probe` (#252) | **Yes** (`sched_getaffinity`) |
| `redb-backend-smoke` | coord restart reads daemon records back, emits `"cleared N stale daemon records"` (#253) | No |
| `daemon-reconnect-smoke` | SIGSTOP daemon 35s → watchdog disconnects → SIGCONT → re-registers (#254) | **Yes** (SIGSTOP/SIGCONT) |
| `state-reconstruction-smoke` | `Running -> Recovering` transition after coord restart (#255, partial) | No |

## Script design

- **CLI isolation.** Installs `dora` into `mktemp -d` scratch root and prepends it to `PATH` only for this script. `trap` cleans on exit. No touch to `~/.cargo/bin/dora`.
- **Port safety.** Bails fast if `lsof -iTCP:6013 -sTCP:LISTEN` finds a listener, with a clear fix hint (`dora destroy`).
- **Aggressive teardown between jobs.** `pkill -CONT` then `pkill -TERM` on `dora (daemon|coordinator)` so one job's leftover coord doesn't contaminate the next.
- **Linux-only jobs skip cleanly** on macOS/BSD — logs a SKIP line, doesn't fail.
[...]

PR #1711, opened 2026-04-21
test: #[ignore] smoke_local_python_concurrent_rw pending PyPI sync (#1710)

heyong4725:
Refs #1710.

## Summary

`make qa-nightly` (after #1709) now runs `cargo test -p dora-examples --test example-smoke -- --test-threads=1`. 51 of 52 tests pass; `smoke_local_python_concurrent_rw` fails with:

```
2: version mismatch: message format v0.8.0 is not compatible with expected message format v0.2.1
   Location: apis/rust/node/src/daemon_connection/mod.rs:67
```

## Root cause (pre-existing drift, not a regression)

Two independent release trains:

| Train | dora-rs version | dora-message format |
|---|---|---|
| Workspace `apis/python/node/` | 0.2.1 | **v0.2.1** |
| PyPI `dora-rs` (latest) | 0.5.0 | **v0.8.0** |

The example's `receive_data.py` does `from dora import Node`. The dataflow YAML has both `.py` paths and `pip install` lines, so `needs_uv()` returns true and the test invokes `dora run --uv`, which spawns an isolated venv and pulls `dora-rs` from PyPI (0.5.0, v0.8.0) — not matching the local workspace daemon (v0.2.1).

The mismatch is intrinsic to the current publishing cadence; no patch to the test harness or the example alone can fix it. Real fix is to publish the workspace Python bindings to PyPI at a version examples can pin — tracked in #1710.

## Short-term fix

Mark the one affected test `#[ignore]` with an inline reason pointing at #1710. Other 51 tests continue to run.

```rust
#[test]
#[ignore = "blocked by PyPI dora-rs drift (#1710): --uv pulls dora-rs 0.5.0 \
(message format v0.8.0) from PyPI, workspace daemon speaks v0.2.1. Re-enable \
once the workspace Python bindings are published to PyPI at a matching version."]
fn smoke_local_python_concurrent_rw() { ... }
```

When #1710 lands, drop the attribute.

## Scope

One attribute on one test function. No logic change. `make qa-nightly` now reports **51 passed, 1 ignored** (with reason visible) instead of failing.

## Test plan

- [ ] `cargo check -p dora-examples --tests` clean
- [ ] `cargo test -p dora-examples --test example-smoke -- --test-threads=1` reports 51 passed + 1 ignored

### DIFF
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -552,6 +552,9 @@ fn smoke_local_python_multiple_arrays() {
 }
 
 #[test]
+#[ignore = "blocked by PyPI dora-rs drift (#1710): --uv pulls dora-rs 0.5.0 \
+(message format v0.8.0) from PyPI, workspace daemon speaks v0.2.1. Re-enable \
+once the workspace Python bindings are published to PyPI at a matching version."]
 fn smoke_local_python_concurrent_rw() {
     // 15s gives more headroom on slow CI runners — this example has two
     // Python nodes in circular data dependency with threaded publish/read,


heyong4725:
Superseded by a proper fix: instead of skipping the test, overlay the workspace Python bindings on top of `--uv` via `[tool.uv.sources]` so the example tests the main branch bindings (not PyPI's stale 0.5.0). New PR incoming.

PR #1712, opened 2026-04-21
fix(qa): use workspace Python bindings in python-concurrent-rw + provision ambient venv for qa-nightly example-smoke (#1710)

heyong4725:
Extended per discussion — now covers the broader audit finding.

**Audit outcome:** `python-concurrent-rw` is the **only** example with a `pyproject.toml`, so it's the only one where `dora run --uv` creates an isolated venv and hits PyPI drift directly. The other 51 Python smoke tests pass "by accident" because `dora run --uv` on examples without `pyproject.toml` falls through to the system Python, which on a dev workstation often has workspace bindings installed. On a clean machine (or CI without the setup step) they'd all fail.

GHA hides this: every nightly Python job does

```yaml
uv venv --seed -p 3.12
source .venv/bin/activate
uv pip install pyarrow
uv pip install -e apis/python/node
```

before running `cargo test`. `qa-nightly` was missing that.

## What this PR now does

1. **`examples/python-concurrent-rw/pyproject.toml`** — adds `[tool.uv.sources]` pointing `dora-rs` at `../../apis/python/node` (editable). Addresses the only example with the isolated-venv drift.

2. **`scripts/qa/all.sh --nightly`** — provisions a fresh scratch venv at `target/qa-nightly-venv`, installs `pyarrow` + `-e apis/python/node`, and runs `example-smoke` with `VIRTUAL_ENV` + `PATH` pointing at it. Matches the GHA nightly Python setup exactly.

3. **`uv` is now a hard prereq** for `--nightly`. If missing, example-smoke fails with a clear install hint (`curl -LsSf https://astral.sh/uv/install.sh | sh`) and a fallback (`make qa-deep` for Tier 1 without Python smoke). Fail fast, don't silently skip.

4. **Doc sweep** — qa-runbook row, contributor cheatsheet, CLAUDE.md, and the script's `--nightly` overview banner all describe the new venv step and the uv requirement.

## Why this matters

Without this, `make qa-nightly` on a fresh clone passes 51/52 or fails 52/52 depending on what Python packages happen to be in the user's ambient env — a hidden prerequisite that defeats "green local predicts green CI." This PR removes that hidden variable.

## Scope

Only `--nightly` mode changes. `--fast` / `--full` / `--deep` / `--mutation-audit` / `--release-gate` unaffected — they don't run example-smoke.

## Test plan

- [ ] `bash -n scripts/qa/all.sh` clean
- [ ] `make qa-nightly` on a workstation with uv installed: 52/52 example-smoke tests pass
- [ ] `make qa-nightly` on a workstation without uv: example-smoke step fails with the install hint, not silent skip
- [ ] `make qa-deep` unchanged (no Python setup, no uv requirement)


heyong4725:
1. Medium: the new prerequisite handling is incomplete, so `qa-nightly` can still fail with a confusing raw `uv venv` error on a machine that satisfies the documented requirement. In `scripts/qa/all.sh:288` the script only preflights `uv`, but a few lines later it unconditionally does `uv venv --seed -p 3.12` at `scripts/qa/all.sh:304`. CI explicitly provisions Python 3.12 first in `.github/workflows/nightly.yml:56`, but the local docs currently describe `uv` as the hard prereq in `docs/qa-runbook.md:58` and `CLAUDE.md:161`, with no equivalent fail-fast check or install hint for a missing 3.12 interpreter. I’d either preflight Python 3.12 explicitly and emit a clear message, or soften the docs so they don’t imply `uv` alone is sufficient.

heyong4725:
@heyong4725 fixed in `52b07173`. Good catch — the prereq handling was incomplete.

**Change:** added a second preflight for Python 3.12 via `uv python find 3.12`, so a machine with uv but no 3.12 now gets a clear specific error instead of a raw `uv venv` failure:

```
=== example-smoke (FAIL: Python 3.12 prerequisite missing) ===
Install with:
  uv python install 3.12
Alternative: skip Python smoke tests with `make qa-deep` ...
```

Preflights run in order (uv first, then 3.12), matching what GHA does — `actions/setup-python@v5` with 3.12 at `.github/workflows/nightly.yml:56` is the equivalent CI step.

**Docs swept** — `qa-runbook.md` row, `contributor-qa-cheatsheet.md` "overnight run" section, `CLAUDE.md` deep-gate bullet, and the `scripts/qa/all.sh` `--nightly` mode comment all now say "**both `uv` and Python 3.12**" and include the `uv python install 3.12` hint alongside the uv install hint. No more implication that uv alone is sufficient.

No behavior change on machines that have both.

PR #1714, opened 2026-04-21
fix(cli): dora replay uses .drec parent as working_dir (#1691 follow-up)

heyong4725:
Addresses the remaining #1691 regression after #1705 (timeout bump).

## Summary

The 2026-04-21 17:23 nightly run surfaced a **different** failure than the 04-21 06:42 run that #1705 fixed. Same issue class, different subsystem:

```
rust-sink: stdout   error: could not find `Cargo.toml` in `/tmp`
[ERROR] failed to build dataflow before run
    0: failed to build node `rust-sink`
    1: build command failed
```

`dora replay` has the **identical shape** to the `dora record` `/tmp` bug that #1674 fixed: writes modified descriptor YAML to `/tmp`, calls `Run::new(tmp_path)` without a `working_dir` override, so the descriptor's `build: cargo build -p <node>` directives run with `cwd=/tmp` → can't find `Cargo.toml`.

## Fix

Two changes:

### 1. `binaries/cli/src/command/replay.rs`

Pass the `.drec` file's parent as `Run`'s `working_dir`. Mirrors `record.rs:241-243` which uses the source YAML's parent.

```rust
let recording_dir = PathBuf::from(&args.file)
    .parent()
    .filter(|p| !p.as_os_str().is_empty())
    .map(PathBuf::from)
    .unwrap_or_else(|| PathBuf::from("."));
let run = Run::new(tmp_path.to_string_lossy().to_string())
    .with_working_dir(recording_dir);
```

**Why `.drec` parent, not `std::env::current_dir()`?**

`current_dir()` would make `dora replay run.drec` from the workspace root set `working_dir=$WORKSPACE`, which trips the build step's types loader on `$WORKSPACE/types/std/*` with:

> failed to load user types: user types cannot use the "std/" prefix

Those are the workspace's own stdlib types, not user types. Separate bug; not fixing here. `.drec` parent avoids it because typical `.drec` locations (next to the example dataflow) don't have a `types/` dir.

### 2. `.github/workflows/nightly.yml`

Update the record-replay job to write `run.drec` into `examples/rust-dataflow/` alongside `dataflow.yml`, so `replay`'s `working_dir` is `examples/rust-dataflow/` (no `types/std/` there, cargo walks up to find the workspace `Cargo.toml`).

## Local verification

```
$ dora record examples/rust-dataflow/dataflow.yml \
    -o examples/rust-dataflow/run.drec
... dataflow finished
$ dora replay examples/rust-dataflow/run.drec
... dataflow finished
$ echo $?
0
```

Before this fix, replay exits with code 0 but prints the `/tmp` error (which is what CI saw).

## Why the existing test didn't catch this

`contract_record_replay_reproduces_validated_pipeline` (in `tests/example-smoke.rs`) uses `write_absolute_path_validated_pipeline_fixture()` which **strips `build:` directives** from the fixture to sidestep the `/tmp` bug class. So it's blind to this failure mode. The GHA nightly record-replay job is the integration test for the `build:` directive path; with this fix + the workflow update, it should pass on the next scheduled run (2026-04-22 06:40 UTC).

Could add a new contract test that keeps `build:` directives intact — but that'd just duplicate what the GHA job already does.

## Test plan

- [ ] Local `cargo check -p dora-cli` clean
- [ ] Local record + replay from workspace root targeting `examples/rust-dataflow/run.drec` completes cleanly
- [ ] Next scheduled nightly (2026-04-22 06:40 UTC) `record-replay` job goes green

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -262,11 +262,20 @@ jobs:
           # re-compile inside the record step (#1691, 2026-04-21 nightly).
           # 300s absorbs that variance while still bounding genuine hangs
           # well under the 30-min job timeout.
-          timeout 300s dora record examples/rust-dataflow/dataflow.yml -o run.drec
-          test -f run.drec || { echo "record did not produce run.drec"; exit 1; }
-          echo "recording size: $(wc -c < run.drec) bytes"
+          #
+          # Write the .drec next to its dataflow, not at the workspace
+          # root: `dora replay` uses the .drec's parent as working_dir
+          # for the embedded build directives (mirrors the fix from
+          # #1674 for record). With the .drec at workspace root, replay
+          # would pick up $WORKSPACE/types/std/* and fail type-loading
+          # ("user types cannot use the 'std/' prefix" -- those are the
+          # workspace's own stdlib types, not user types).
+          timeout 300s dora record examples/rust-dataflow/dataflow.yml \
+            -o examples/rust-dataflow/run.drec
+          test -f examples/rust-dataflow/run.drec || { echo "record did not produce run.drec"; exit 1; }
+          echo "recording size: $(wc -c < examples/rust-dataflow/run.drec) bytes"
       - name: Replay
-        run: dora replay run.drec
+        run: dora replay examples/rust-dataflow/run.drec
 
   cluster-smoke:
     # Smoke-tests the cluster lifecycle commands that don't require SSH:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -45,7 +45,13 @@ jobs:
   smoke-suite:
     name: Smoke suite (tests/example-smoke.rs)
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    # 52 tests run with --test-threads=1 because they share coord ports.
+    # Observed ~14.5 min locally on a warm workstation; GHA runners are
+    # slower and cold-cache cargo compiles add 5-10 min of overhead.
+    # 30 min was marginal and triggered "cancelled" on the 2026-04-19,
+    # 04-20, 04-21 nightly runs -- the suite was still making progress
+    # when the timeout fired. 60 min gives margin without being wasteful.
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
--- a/binaries/cli/src/command/replay.rs
+++ b/binaries/cli/src/command/replay.rs
@@ -7,7 +7,7 @@ use std::{
 
 use clap::Args;
 use dora_recording::RecordingReader;
-use eyre::{Context, bail};
+use eyre::{bail, Context};
 
 use crate::command::{Executable, Run};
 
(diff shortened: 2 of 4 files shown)


heyong4725:
Extended per your ask — `b7f41545` adds a `record-replay` job to `scripts/qa/ci-nightly-jobs.sh` so `make qa-nightly` catches this regression class locally next time, not just on CI.

**Why the previous coverage claim was wrong:**

`example-smoke.rs`'s `contract_record_replay_reproduces_validated_pipeline` uses `write_absolute_path_validated_pipeline_fixture()` which **strips `build:` directives** from the YAML (intentional — sidesteps the `/tmp` cargo bug class, see comment at `tests/example-smoke.rs:1015`). That means the local test exercises the record/replay data round-trip but **never runs the descriptor's `build: cargo build -p <node>` directives during replay** — which is where #1691 broke.

I had assumed example-smoke covered record-replay when wiring up `ci-nightly-jobs.sh` in #1709. It didn't. That's why my "green local qa-nightly → green CI nightly" claim broke on this path.

**Fix:** add `job_record_replay` to ci-nightly-jobs.sh, mirroring the GHA `record-replay` job step bodies exactly — record `examples/rust-dataflow/dataflow.yml` (build directives **intact**), assert non-empty `.drec`, replay. Same as GHA would run.

Doc updates in this commit:
- `ci-nightly-jobs.sh` header: "6 GHA-only jobs" → "7"; example-smoke coverage corrected (dropped "+ record-replay" since that was always nominal)
- `all.sh` `--nightly` banner: job 14 now lists `record-replay` among ci-nightly-jobs coverage, with an explainer about why it lives there and not in example-smoke

After this PR + merge, `make qa-nightly` on a fresh clone will run:
- step 13 example-smoke (52 tests, Python coverage etc.) — covers 4 GHA jobs
- step 14 ci-nightly-jobs (7 jobs including record-replay with build directives intact) — covers the remaining 7

All 11 GHA nightly test jobs now have a true local equivalent. This specific `/tmp` regression class (record or replay) will be caught the next time anyone touches those code paths.

heyong4725:
Medium: the new local `record-replay` parity job documents a bounded timeout, but it does not actually enforce one, so it can hang indefinitely where the GitHub nightly job would fail fast. In `scripts/qa/ci-nightly-jobs.sh` the comment says “Bounded timeout,” but the code runs plain `dora record` / `dora replay` with no `timeout` wrapper. The corresponding GitHub job still bounds at least the record step with `timeout 300s` in `.github/workflows/nightly.yml`. If `qa-nightly` is meant to be local parity, this should either use the same timeout behavior or stop claiming bounded parity for this subjob.

heyong4725:
Fair catch — fixed in `9090c81b`. The comment was aspirational; the code wasn't enforcing it.

```diff
- if ! dora record examples/rust-dataflow/dataflow.yml -o "$DREC"; then
-   echo "ERROR: dora record failed"
+ if ! timeout 300s dora record examples/rust-dataflow/dataflow.yml -o "$DREC"; then
+   echo "ERROR: dora record failed or exceeded 300s"

- if ! dora replay "$DREC"; then
-   echo "ERROR: dora replay failed"
+ if ! timeout 120s dora replay "$DREC"; then
+   echo "ERROR: dora replay failed or exceeded 120s"
```

300s on record matches the GHA `record-replay` step (`.github/workflows/nightly.yml:261`, tuned in #1705 for cold-cache rebuild variance). 120s on replay is generous — replay is normally ~2s — but still bounded.

Error messages now report the timeout budget so a hang is distinguishable from a genuine failure.

Consistent with the existing `timeout 30`/`timeout 15` uses elsewhere in the same script (on `dora self update` and `dora topic echo`). Relies on GNU coreutils `timeout` — standard on Linux, macOS via `brew install coreutils` — which this script already assumed.

heyong4725:
Added option B per your ask in `e4645a42`: bumped `smoke-suite` job's `timeout-minutes: 30 → 60`.

**Why:** three consecutive nightly runs (04-19, 04-20, 04-21) showed `smoke-suite` as **cancelled** (not failed). That's GHA's SIGKILL when a job hits its `timeout-minutes`. Looking at the numbers:

| Environment | Observed duration |
|---|---|
| Local warm workstation | ~14.5 min (52 tests, `--test-threads=1`) |
| GHA ubuntu-latest (cold cache) | ~20-30 min estimated (2x slower runner + 5-10 min extra cargo compile overhead) |
| Old timeout | 30 min |

So 30 min was within the margin of error for a cold-cache run to finish at all. Tests were still making progress when SIGKILL fired.

60 min doubles the budget. Warm-cache runs still finish in 20-35 min; genuine hangs get caught well before the new ceiling. Nightly is cron-scheduled once per day, so there's no runner-capacity concern with a higher ceiling.

---

## PR summary after this commit

| Commit | Fix | Effect |
|---|---|---|
| `3105ef11` | `dora replay` uses `.drec` parent as `working_dir` + nightly.yml writes `.drec` next to dataflow | CI nightly `record-replay` job passes |
| `b7f41545` | Adds `job_record_replay` to `ci-nightly-jobs.sh` with build directives intact | Local `make qa-nightly` catches this regression class next time |
| `9090c81b` | Wraps local record/replay with `timeout 300s`/`120s` | Local matches CI's hard bound |
| `e4645a42` | Bumps smoke-suite `timeout-minutes: 30 → 60` | CI nightly `smoke-suite` job actually finishes instead of getting cancelled |

Together with the already-merged #1712 (Python venv) and #1713 (drop concurrent-rw pyproject), the next scheduled nightly (2026-04-22 06:40 UTC) should go green on all 11 jobs.

PR #1729, opened 2026-04-23
fix: use current_exe() instead of which("dora") to locate runtime binary

DGHX12345:
## Summary

The daemon used `which::which("dora")` to resolve the binary path for
spawning runtime nodes. This resolves to whichever `dora` appears first
in `$PATH`, which may be a different version than the running daemon.

Changed to `std::env::current_exe()` so the daemon always invokes its
own binary with the `runtime` subcommand, guaranteeing version
consistency.

## Exposed by

Running `examples/c++-dataflow` with a locally-built binary while a
different version of `dora` was present in `$PATH` — runtime nodes
failed with "No such file or directory".

### DIFF
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -297,8 +297,9 @@ impl Spawner {
                         ]);
                         Some(cmd)
                     } else {
-                        let mut cmd =
-                            Command::new(which::which("dora").wrap_err("failed to get dora path")?);
+                        let current_exe = std::env::current_exe()
+                            .wrap_err("failed to get current executable path")?;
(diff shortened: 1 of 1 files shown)


phil-opp:
Thanks for the PR!

This is effectively a revert of https://github.com/dora-rs/dora/commit/da2593f9 . @haixuanTao Do you remember why you did this change?


I also tried to implement a more general approach in ac23cb4f0, but this was never merged as it was deemed too complex.

haixuanTao:
The issue is that if you want to use the daemon within a rust script or python script by embedding it without using the cli, this makes it possible

phil-opp:
Looks like this landed in https://github.com/dora-rs/dora/pull/1801

PR #1731, opened 2026-04-23
fix: handle empty parent path when resolving dataflow working directory

DGHX12345:
## Summary

`Path::parent()` returns `Some("")` for bare filenames like
`"dataflow.yml"`, not `None`. The existing `.unwrap_or(Path::new("."))`
fallback only triggers on `None`, so `canonicalize("")` fails with
"No such file or directory" when the CLI tries to expand module
references.

Added `.filter(|p| !p.as_os_str().is_empty())` before the fallback in
the central `working_dir_or_parent()` helper and all 4 direct call
sites (start, expand, graph, validate). Updated the unit test to assert
the corrected behavior.

## Exposed by

Running `examples/module-dataflow` — `dora run dataflow.yml` from the
example directory triggers the empty base_dir path, failing module
expansion.

### DIFF
--- a/binaries/cli/src/command/expand.rs
+++ b/binaries/cli/src/command/expand.rs
@@ -25,6 +25,7 @@ impl Executable for Expand {
         let working_dir = self
             .dataflow
             .parent()
+            .filter(|p| !p.as_os_str().is_empty())
             .unwrap_or_else(|| std::path::Path::new("."));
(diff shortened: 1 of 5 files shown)


phil-opp (approved):

phil-opp:
Thanks!

phil-opp:
@mergifyio refresh

phil-opp:
https://github.com/Mergifyio refresh

PR #1734, opened 2026-04-23
ci(Mergify): configuration update

phil-opp:
This change has been made by @phil-opp from the Mergify Queue Rule Configurator.


phil-opp:
Superseded — redoing with a proper description and without stripping the existing comments in .mergify.yml.

PR #1735, opened 2026-04-23
ci(Mergify): set merge_queue max_parallel_checks to 1

phil-opp:
Serializes merge-queue CI runs so only one batched queue branch's checks run at a time, matching our limited CI parallelism budget. Without this, Mergify can kick off checks on multiple queue branches in parallel and saturate the runner pool.


phil-opp:
@mergifyio queue

Issue #1626, opened 2026-04-17
Phase 5b: RC stabilization + dogfood (v1.0.0-rc.1 → GA)


heyong4725:
Tracks stabilization of the 1.0 consolidation from the `v1.0.0-rc.1` tag through to `v1.0.0` GA.

Historical tracker (pre-consolidation, on the archived fork): [dora-rs/adora#292](https://github.com/dora-rs/adora/issues/292). This issue is its canonical successor on the merged repository.

## Context

The 1.0 consolidation landed on `main` at commit `145ccce04` via squash-merge of the fork's `v1.0-rewrite` branch. Tag `v1.0.0-rc.1` points at that commit. Nothing has been published to crates.io or PyPI yet — tag push is safe by design; publish requires an explicit GitHub release.

- Main: `145ccce04 dora 1.0: consolidation of the Q1 2026 Rust-first rewrite`
- Tag: `v1.0.0-rc.1`
- Detail history preserved at:
  - Tag `consolidation/v1.0-rewrite-detail` (pre-squash tip `869c70643`)
  - Branch `v1.0-rewrite` (kept alive on origin)
  - Archived `dora-rs/adora` repo (post-GA)

Full plan: `docs/plan-dora-1.0-consolidation.md` §5 Phase 5b.

## Stabilization loop

```
rc.1 (tagged 2026-04-17) → dogfood → fix P0/P1 → rc.N tag → dogfood → ... → GA
```

Each `rc.N` tag restarts a 7-day dogfood clock. A clean 7-day run with zero P0/P1 against the most recent tag is the exit criterion for Phase 5b. No calendar cap; ship when the RC is clean.

## Planned RC landings between rc.1 and GA

- [ ] **rc.2** — Control-channel cleanup, equivalent of upstream commit `01995ad67`. Originally tracked at [dora-rs/adora#313](https://github.com/dora-rs/adora/issues/313). Remove unused shared-memory control-channel types (`ShmemChannel`, `ShmemClient`, `ShmemServer`) and the `LocalCommunicationConfig::Shmem` variant. TCP is `#[default]` already; this is dead-code deletion plus cascade through match arms. ~2-3 hours.
[...]

PR #1738, opened 2026-04-23
Remove unused shared-memory control channel between nodes and daemon

phil-opp:
The shmem-based control channel was disabled by default (TCP is the default LocalCommunicationConfig) and never used in practice. This removes the entire shared-memory-server crate, the Shmem variants from LocalCommunicationConfig and DaemonCommunication, and all
related wiring in the daemon and node API. The data-plane shared memory (shared_memory_extended) for zero-copy message passing is preserved.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

see #1626


phil-opp:
@mergifyio queue

PR #1741, opened 2026-04-23
Add per-node shared_memory_pool_size config and zero_copy_threshold getter

phil-opp:
Salvaged from the closed #1378 (zenoh-SHM refactor), which was written against pre-rewrite main and became unmergeable after the Q1 2026 consolidation. These are the portable pieces:

- New optional `shared_memory_pool_size` field on dataflow nodes. Accepts an integer (raw bytes) or a string with unit suffix (`KB`, `MB`, `GB`). Priority: YAML > `DORA_NODE_SHM_POOL_SIZE` env var > built-in default.
- `ByteSize` newtype in `dora-message` with string/integer serde, `Display`, `FromStr`, `JsonSchema`.
- `DoraNode::zero_copy_threshold()` getter.


phil-opp:
@mergifyio queue

PR #1745, opened 2026-04-23
Remove custom shared memory and drop token infrastructure

phil-opp:
This PR removes the custom shared memory allocation and drop token tracking infrastructure from the Dora node runtime, simplifying the codebase by relying entirely on Zenoh's SHM provider for zero-copy data transport.

Follow-up to https://github.com/dora-rs/dora/pull/1741. These changes were originally implemented as part of #1378

## Summary
The changes eliminate the custom shared memory region management (`DropStream`, `ShmemHandle`, drop token tracking) that was previously used for efficient data transfer when Zenoh SHM was unavailable. Since Zenoh SHM is now required (the runtime errors if no tokio runtime is available for Zenoh initialization), the custom fallback path is no longer needed.

## Key Changes

- **Removed drop token infrastructure**: Deleted `drop_stream.rs` module and all associated drop token handling, including:
  - `DropStream` struct and its initialization logic
  - Drop token tracking in `RunningDataflow` and event stream processing
  - `DropTokenInformation` struct in daemon
  - `DaemonRequest::SubscribeDrop`, `NextFinishedDropTokens`, and `ReportDropTokens` variants

- **Simplified data allocation**: 
  - `allocate_data_sample()` now always allocates heap buffers (Vec) instead of custom shared memory
  - Removed `allocate_shared_memory()` and cache management (`VecDeque<ShmemHandle>`)
  - Removed `sent_out_shared_memory` HashMap from `DoraNode`

- **Simplified message sending**:
  - `DataMessage` enum now only contains `Vec` variant (removed `SharedMemory` variant)
  - Removed shared memory ID and drop token from data messages
  - Simplified `send_output()` logic by removing custom shmem handling

- **Streamlined Zenoh initialization**:
  - Changed error handling to require Zenoh session (returns error instead of gracefully degrading)
  - Simplified session and provider initialization with clearer error propagation
  - Removed fallback paths for when Zenoh is unavailable

- **Cleaned up event stream**:
  - Removed ACK channel mechanism for tracking when events are dropped
  - Removed pending drop token tracking and timeout logic
  - Simplified `EventItem` to remove `ack_channel` field

- **Daemon simplifications**:
  - Removed drop region allocation and listener loop
  - Removed `SubscribeDrop` event handling
  - Removed `ReportDrop` event handling and drop token state management
  - Removed `drop_channels` from `RunningDataflow`

- **Dependency cleanup**: Removed `shared_memory_extended` crate dependency

## Implementation Details

The removal is safe because:
[...]

PR #1746, opened 2026-04-23
Harden zenoh subscriber threads against panics and shutdown hangs

phil-opp:
Three related fixes salvaged from the closed #1378 that were not covered by #1745:

1. Wrap each zenoh subscriber thread in `catch_unwind` so a panic surfaces as `EventItem::FatalError` instead of silently killing sample delivery.
2. Replace blocking `subscriber.recv()` with `futures::future::select` on `recv_async()` and a shutdown channel; dropping `EventStream` now wakes every subscriber immediately.
3. Add a `stop_received` flag so `recv_async` / `poll_next` return `None` after `Event::Stop` — subscriber threads hold event-channel sender clones that would otherwise keep the receiver open.

Includes regression tests for the Stop → None invariant.


phil-opp:
@mergifyio queue

PR #1748, opened 2026-04-23
Opt-in owned tokio runtime for nodes without ambient executor

phil-opp:
Follow-up to #1745. Embedders without an ambient tokio runtime can set `DORA_CREATE_OWNED_TOKIO_RUNTIME=1` to have `DoraNode::init` build a multi-threaded runtime for the zenoh SHM data plane. Default stays strict — init errors if no runtime is available.

Depends on #1745 being merged first (this branch is based on it).


phil-opp:
@mergifyio queue

PR #1789, opened 2026-05-01
feat(api) add timestamp function to node api

oceanusxiv:
## Summary

- Add `timestamp()` method to the Rust `DoraNode` that returns the current HLC timestamp from the node's internal clock
- Expose `timestamp()` on the Python `Node` class, returning a timezone-aware UTC `datetime.datetime`
- Make the cached `datetime_module()` helper public in `dora-operator-api-python` so the node crate can reuse it instead of calling `PyModule::import` on every invocation

## Motivation

`DoraNode` already maintains an internal `uhlc::HLC` clock used to timestamp every outgoing message, but there was no way for user code to read the current clock value directly. This is useful for:

- Measuring processing latency by comparing the current clock against an input event's timestamp
- Correlating log entries or external events with the same logical clock used by dora's data plane
- Profiling node processing time without introducing a separate clock source

## Changes

| File | Change |
|------|--------|
| `apis/rust/node/src/node/mod.rs` | Add `pub fn timestamp(&self) -> uhlc::Timestamp` |
| `apis/python/node/src/lib.rs` | Add `timestamp()` PyO3 method returning `datetime.datetime` (UTC) |
| `apis/python/operator/src/lib.rs` | Make `datetime_module()` `pub` for cross-crate reuse |
| `apis/python/node/dora/__init__.pyi` | Add type stub for `timestamp()` |

## Usage

**Rust:**
```rust
let node = DoraNode::init_from_env()?;
let ts = node.timestamp();
println!("Current HLC time: {ts}");
```

**Python:**
```python
from dora import Node

node = Node()
print(f"Current HLC time: {node.timestamp()}")

for event in node:
    if event["type"] == "INPUT":
        latency = node.timestamp() - event["metadata"]["timestamp"]
        print(f"Processing latency: {latency}")
```

### DIFF
--- a/apis/python/node/dora/__init__.pyi
+++ b/apis/python/node/dora/__init__.pyi
@@ -1,3 +1,4 @@
+import datetime
 import typing
 
 import pyarrow
@@ -148,6 +149,10 @@ class Node:
         """Returns how many times this node has been restarted.
 
         Returns 0 on the first run, 1 after the first restart, etc."""
+
+    def timestamp(self) -> "datetime.datetime":
+        """Returns the current timestamp from the node's Hybrid Logical Clock
+        as a UTC datetime object."""
 
     def merge_external_events(self, subscription: dora.Ros2Subscription) -> None:
         """Merge an external event stream with dora main loop.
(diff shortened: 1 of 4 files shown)


heyong4725:
Closing in favor of #1877, which rescues this against current main. Cherry-picks cleanly, plus expanded docstrings (the HLC-vs-SystemTime guidance from your motivation is now in the rustdoc) and a regression test pinning the node-clock-not-fresh-clock contract. `Co-authored-by` credit preserved. Thanks for the original work!

PR #1790, opened 2026-05-02
feat(mavlink2-bridge): MAVLink 2 ↔ Arrow bridge with SITL closed-loop validation

### LINKED ISSUE #1786: MAVLink2 bridge as an in-tree extension
I would like to propose adding MAVLink 2 support to dora as an in-tree extension, structured as a peer of dora-ros2-bridge: a libraries/extensions/mavlink2-bridge/ library (with msg-gen + arrow sub-crates), a binaries/mavlink2-bridge-node/ standalone bridge node spawnable by the daemon, and examples/mavlink2-bridge/ covering Rust/Python/YAML.  A series of small PRs will be created.

**Motivation**
Today a dora user who wants to drive a Pixhawk-based drone or wheeled robot has three bad options:                                                                                                                               
                                                                  
  1. Bolt MAVLink onto a custom node from scratch — every user reimplements the same XML schema → Arrow conversion.                                
  2. Run a separate ROS2 → MAVLink bridge (e.g. mavros) and pipe it through dora-ros2-bridge. This adds a full ROS2 hop, defeats dora's zero-copy promise, and forces a ROS2 install on robots that otherwise don't need one.                                                                      
  3. Skip dora and use a different framework. We lose those users
  
 **Why an in-tree extension**                                                                                                                                                                             
  The closest precedent in the repo is dora-ros2-bridge. ROS2 is structurally identical to MAVLink for our purposes: external protocol with versioned IDL/XML schemas, requires build-time codegen, needs Arrow ↔ message conversion, and ships as a daemon-spawnable bridge node. The ROS2 bridge is in-tree (libraries/extensions/ros2-bridge/{msg-gen,arrow,python} + binaries/ros2-bridge-node), and that decision has worked well.     

**Scope**

  1. Workspace member skeleton + feature flag (no logic).
  2. msg-gen codegen crate (XML → Rust via quick-xml + quote/syn, mirroring dora-ros2-bridge-msg-gen).
  3. arrow sub-crate: MAVLink ↔ RecordBatch round-trip + unit tests.
  4. binaries/mavlink2-bridge-node/ with serial + UDP + TCP transports.
  5. Examples + smoke tests (TCP loopback, no Pixhawk required in CI).
  6. Optional PyO3 bindings.

bobdingAI:
## Summary

Adds `dora-mavlink2-bridge` (library crate) and `dora-mavlink2-bridge-node`
(daemon-spawnable binary) to bring MAVLink 2 ↔ Apache Arrow conversion into
the dora workspace. Closes the dora ↔ autopilot loop with full SITL
validation against ArduPilot Copter (takeoff → hover → land → disarm) and
ArduRover (RC-override-driven square drive).

This implements the feature requested in #1786.

## Why one PR (not 12 stacked)

The stack was originally going to ship as 12 separate PRs per the contributor
guide's preference for small commits. After review of the actual diff:

- **Every change to existing dora files is additive** (`Cargo.toml` adds
  workspace members, `README.md` / `CLAUDE.md` / `Changelog.md` add
  table rows / list entries, `tests/example-smoke.rs` adds new test
  functions, `scripts/smoke-all.sh` adds new conditional blocks). No
  existing dora logic is modified anywhere.
- All new code lives in **new crates** (`libraries/extensions/mavlink2-bridge/`,
  `binaries/mavlink2-bridge-node/`) and **new example directories**
  (`examples/mavlink2-bridge/`, `examples/mavlink2-bridge-sitl-mission/`).
- The 13 commits remain visible in the PR for commit-by-commit review.

Splitting into 12 PRs would have multiplied review-coordination cost without
narrowing the per-PR domain knowledge required (every PR needs the same
MAVLink + ArduPilot + dora bridge context).

### New crates

| Path | Crate | Role |
|------|-------|------|
| `libraries/extensions/mavlink2-bridge/` | `dora-mavlink2-bridge` | Library: MAVLink 2 ↔ Arrow conversion (`MavlinkArrow` trait), transport URL parsing (TCP / UDP / serial), error types. 13 message types from the MAVLink common dialect. |
| `binaries/mavlink2-bridge-node/` | `dora-mavlink2-bridge-node` | Daemon-spawnable binary: opens MAVLink 2 connection, decodes inbound frames into Arrow `StructArray` rows, dispatches dora inputs as outbound MAVLink frames. Auto-requests all data streams at 5 Hz on connect. |

### New examples

| Path | What it demos |
|------|---------------|
| `examples/mavlink2-bridge/` | In-process MAVLink simulator + 3 telemetry-printer flavors (Rust / Python / C++); 3 dataflow YAMLs (per language); cross-language demo of the bridge. |
| `examples/mavlink2-bridge-sitl-mission/` | Closed-loop SITL missions against ArduPilot Copter + Rover. Three missions: simple takeoff → hover → land (`mission.py`), long flight with square waypoints + hover (`mission_long.py`), rover square drive via RC override (`mission_rover.py`). Plus a TCP→UDP forwarder for QGroundControl observation (`scripts/forward_to_qgc.py`). Local-only; not in CI. |

### Bridge feature surface

**Telemetry (MAVLink → dora outputs):**
HEARTBEAT, SYS_STATUS, SYSTEM_TIME, ATTITUDE, ATTITUDE_QUATERNION,
LOCAL_POSITION_NED, GLOBAL_POSITION_INT, GPS_RAW_INT, RC_CHANNELS,
SERVO_OUTPUT_RAW, COMMAND_LONG, COMMAND_ACK, MISSION_CURRENT.

**Commands (dora inputs → MAVLink):**
`heartbeat_cmd`, `command_long_cmd`, `set_mode_cmd`,
`rc_channels_override_cmd`.

**Transports:**
- `tcp://host:port[?proto=v1|v2]` (TCP outgoing client)
- `udp://host:port[?proto=v1|v2]` (UDP listener)
- `serial:///dev/tty.X[?baud=N&proto=v1|v2]` (Unix serial)
- `serial://COM1[?baud=N&proto=v1|v2]` (Windows serial)

The `?proto=` query string is required for autopilots that speak MAVLink V1
only (e.g. ArduCopter ≤ 3.x in dronekit-sitl); without it, mavlink-rust
0.13's `recv()` silently drops every V1 frame.

**Auto-stream request:** on connect, the bridge sends
`REQUEST_DATA_STREAM(MAV_DATA_STREAM_ALL, 5 Hz)` once. ArduPilot 3.x emits
only HEARTBEAT until a GCS asks for streams; without this, dora consumers
saw empty `global_position_int` topics. Failure is non-fatal.

### Unit + integration tests in this repo

- `cargo clippy -p dora-mavlink2-bridge -p dora-mavlink2-bridge-node --all-targets -- -D warnings` ✓
- `cargo test -p dora-mavlink2-bridge -p dora-mavlink2-bridge-node` ✓ — 28 tests:
  - 1 build sanity (`MAVLINK_VERSION = V2`)
  - 13 Arrow round-trip (one per supported message type)
  - 9 transport URL parsing
  - 3 inline `parse_baud`
  - 1 TCP loopback (real `mavlink::connect`)
  - 1 UDP loopback (real `mavlink::connect`)
- `cargo test --test example-smoke smoke_local_mavlink2_bridge` ✓ — local smoke (in-process MAVLink simulator + bridge + telemetry-printer in one dora dataflow)

### End-to-end SITL validation (local-only by design)

ArduPilot SITL is heavy (~5–10 min one-time waf build, requires Python
toolchain + ArduPilot source clone) and not appropriate for CI. Validated
manually on macOS 26.x (Apple Silicon) and Ubuntu:

**ArduCopter mission (mission.py):**
```
mission: vehicle heartbeat received
mission: -> SET_MODE GUIDED (cmd=176)    <- ACK [ACCEPTED, 0.1s]
mission: -> ARM (cmd=400)                <- ACK [ACCEPTED, 0.1s]
mission: -> TAKEOFF (cmd=22)             <- ACK [ACCEPTED, 0.1s]
mission: reached 8.76 m (target 8.50)
mission: hovering for 5.0s
mission: -> LAND (cmd=21)                <- ACK [ACCEPTED, 0.1s]
mission: landed at -1.98 m
mission: -> DISARM (cmd=400)             <- ACK [ACCEPTED, 0.1s]
mission: SUCCESS - takeoff + hover + land + disarm complete
```

**ArduCopter long flight (mission_long.py):** TAKEOFF 30 m → square
waypoints → hover 30 s → LAND. Validated SUCCESS in ~85 s. QGroundControl
observed the entire flight live via `forward_to_qgc.py` on UDP 14550.

**ArduRover square drive (mission_rover.py):** RC_CHANNELS_OVERRIDE-driven
clockwise square. Validated SUCCESS, ~700 m of track on the QGC map.

## What's NOT in this PR

- **CI smoke for SITL.** SITL is too heavy for PR CI as documented in
  `CLAUDE.md` ("Do NOT run the full QA suite on every commit/push") and
  `docs/qa-runbook.md`. The `tests/example-smoke.rs` audit table now lists
  `mavlink2-bridge-sitl-mission` as a manual-validation row alongside
  `cuda-benchmark`, `ros2-bridge`, etc.
- **`STATUSTEXT` / `BATTERY_STATUS` array fields.** Both have fixed-size
  `u8[50]` / `u16[10]` arrays that need Arrow `FixedSizeBinary` /
  `FixedSizeList` support. Documented in `arrow_convert.rs` module
  comment. Worth a follow-up PR.
- **PyO3 bindings for the bridge crate.** dora's existing Python node
  API is sufficient for users; PR demonstrates this with the
  `telemetry_printer.py` and `mission*.py` examples.

## Known limitation: mavlink-rust 0.13 `MavMode` enum

The MAVLink spec defines `SET_MODE.base_mode` as a `uint8_t` bitfield, but
mavlink-rust 0.13 generates `MavMode` as a strict enum with only 11 named
variants. None has bit `0x01` (`MAV_MODE_FLAG_CUSTOM_MODE_ENABLED`) set,
so the `set_mode_cmd` input cannot drive ArduPilot custom modes (GUIDED,
AUTO, RTL) through the bridge. Workaround: use `MAV_CMD_DO_SET_MODE` via
the existing `command_long_cmd` input on autopilots that support it
(ArduCopter ≥ 3.6, recent PX4). On older firmware, the SITL mission
examples open a side-channel `pymavlink` connection on TCP 5763
(documented in their source headers and the example README).

Tracked upstream at <https://github.com/mavlink/rust-mavlink>. Once the
upstream accepts `base_mode` as raw `u8`, follow-up PR removes the
limitation docs and the side-channel.

The bridge surfaces this with a descriptive runtime error
("`SET_MODE base_mode=N (0xNN) is a valid uint8 per MAVLink spec but not
representable as mavlink-rust 0.13's strict MavMode enum...`") rather than
the bare "`invalid base_mode discriminant`" that would otherwise be
emitted by the generic `decode_enum` helper.

## Commits in this PR (13)

For commit-by-commit review:

1. `a81be73a` feat(mavlink2-bridge): scaffold workspace extension
2. `6f2c7b7f` feat(mavlink2-bridge): add MavlinkArrow trait with HEARTBEAT impl
3. `65a00744` feat(mavlink2-bridge-node): TCP bridge node + transport URL parser
4. `33be78a3` feat(mavlink2-bridge): add UDP and serial transports
5. `3cdfb059` feat(mavlink2-bridge): expand Arrow conversion to common dialect
6. `c17d7fb9` feat(mavlink2-bridge): example dataflow + smoke tests + docs
7. `f0e2ade2` feat(mavlink2-bridge): SITL closed-loop mission example
8. `5def41bd` fix(mavlink2-bridge): URL ?proto=v1|v2 query for V1-only firmware
9. `6899c79c` feat(mavlink2-bridge): SET_MODE + RC_CHANNELS_OVERRIDE inputs + auto-request data streams
10. `0fc6526f` feat(mavlink2-bridge-sitl-mission): long flight demo + QGC forwarder
11. `f574ba0b` feat(mavlink2-bridge-sitl-mission): rover variant — square-drive on dronekit-sitl
12. `993ff47f` docs(mavlink2-bridge): document SET_MODE base_mode mavlink-rust 0.13 limitation
13. `279355c7` chore(gitignore): ignore SITL runtime artifacts and .project_notes/

Closes #1786.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1032,18 +1032,27 @@ jobs:
       - cargo-cache-restore:
           prefix: ros2
       - run:
-          name: Install Python 3.12 + dev headers + clang
-          # deadsnakes PPA + apt-get update + apt-get install can stall
-          # silently for >10 min during peak hours, tripping CircleCI's
-          # default 10-minute no-output timeout. Bump to 20 min so a slow
-          # mirror doesn't kill the job before package install completes.
-          no_output_timeout: 20m
+          name: Install Python 3.12 (via uv) + clang
+          # Install Python 3.12 via uv's prebuilt distributions
+          # (python-build-standalone, hosted on GitHub Releases) instead
+          # of the deadsnakes PPA. Avoids the Launchpad dependency:
+          # `add-apt-repository ppa:deadsnakes/ppa` calls the Launchpad
+          # API via httplib2, which on PR #1790 timed out twice with
+          # `TimeoutError: [Errno 110] Connection timed out` reaching
+          # api.launchpad.net. uv's distributions are versioned, ship
+          # dev headers, and pull from GitHub's CDN — orthogonal infra.
+          # Symlink to /usr/bin/python3.12 so the downstream venv step
+          # (uv venv --python /usr/bin/python3.12) keeps working.
+          no_output_timeout: 10m
           command: |
-            sudo add-apt-repository -y ppa:deadsnakes/ppa
-            sudo apt-get update
+            curl -LsSf https://astral.sh/uv/install.sh | sh
+            "$HOME/.local/bin/uv" python install 3.12
+            PYTHON312=$("$HOME/.local/bin/uv" python find 3.12)
+            sudo ln -sf "$PYTHON312" /usr/bin/python3.12
             # clang is needed by the C++ ROS2 Bridge example's run.rs
             # (machine image ships g++ but not clang by default).
-            sudo apt-get install -y python3.12 python3.12-dev python3.12-venv libpython3.12 clang
+            sudo apt-get update
+            sudo apt-get install -y clang
       - run:
           name: Install ROS2 Humble
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1044,15 +1044,31 @@ jobs:
           # Symlink to /usr/bin/python3.12 so the downstream venv step
           # (uv venv --python /usr/bin/python3.12) keeps working.
           no_output_timeout: 10m
+          environment:
+            # Stop apt-get from triggering the interactive `needrestart`
+            # TUI on Ubuntu 22.04, which hangs forever waiting for input
+            # whenever a package update touches a shared library
+            # (libc-dev etc., transitively pulled by clang). Setting it
+            # in the step env applies to every command in this run.
+            NEEDRESTART_MODE: a
+            NEEDRESTART_SUSPEND: "1"
+            DEBIAN_FRONTEND: noninteractive
           command: |
             curl -LsSf https://astral.sh/uv/install.sh | sh
             "$HOME/.local/bin/uv" python install 3.12
             PYTHON312=$("$HOME/.local/bin/uv" python find 3.12)
             sudo ln -sf "$PYTHON312" /usr/bin/python3.12
+            # Persist the noninteractive env for downstream steps so the
+            # ROS2 Humble apt install doesn't re-trigger the same prompt.
+            {
+              echo "export NEEDRESTART_MODE=a"
+              echo "export NEEDRESTART_SUSPEND=1"
+              echo "export DEBIAN_FRONTEND=noninteractive"
+            } >> "$BASH_ENV"
             # clang is needed by the C++ ROS2 Bridge example's run.rs
             # (machine image ships g++ but not clang by default).
             sudo apt-get update
-            sudo apt-get install -y clang
+            sudo -E apt-get install -y clang
       - run:
           name: Install ROS2 Humble
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1058,6 +1058,19 @@ jobs:
             "$HOME/.local/bin/uv" python install 3.12
             PYTHON312=$("$HOME/.local/bin/uv" python find 3.12)
             sudo ln -sf "$PYTHON312" /usr/bin/python3.12
+            # Make uv's libpython3.12 visible to the dynamic linker.
+            # PyO3 dynamically links against libpython3.12.so.1.0; the
+            # deadsnakes apt package previously installed it under
+            # /usr/lib/x86_64-linux-gnu (a standard ld.so search path)
+            # but uv lays it out under <prefix>/lib in its data dir,
+            # which isn't on the default search path. Symlink the
+            # versioned + unversioned soname into the system lib dir
+            # and refresh ld.so.cache so cargo-test binaries can load
+            # them at runtime.
+            PYTHON312_PREFIX=$(dirname "$(dirname "$PYTHON312")")
+            sudo ln -sf "$PYTHON312_PREFIX/lib/libpython3.12.so.1.0" /usr/lib/x86_64-linux-gnu/libpython3.12.so.1.0
+            sudo ln -sf "$PYTHON312_PREFIX/lib/libpython3.12.so" /usr/lib/x86_64-linux-gnu/libpython3.12.so
+            sudo ldconfig
             # Persist the noninteractive env for downstream steps so the
             # ROS2 Humble apt install doesn't re-trigger the same prompt.
             {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1033,6 +1033,11 @@ jobs:
           prefix: ros2
       - run:
           name: Install Python 3.12 + dev headers + clang
+          # deadsnakes PPA + apt-get update + apt-get install can stall
+          # silently for >10 min during peak hours, tripping CircleCI's
+          # default 10-minute no-output timeout. Bump to 20 min so a slow
+          # mirror doesn't kill the job before package install completes.
+          no_output_timeout: 20m
           command: |
             sudo add-apt-repository -y ppa:deadsnakes/ppa
             sudo apt-get update
--- a/.gitignore
+++ b/.gitignore
@@ -209,3 +209,16 @@ out/
 yolo.yml
 
 ~*
+
+# ArduPilot SITL runtime artifacts (dropped in cwd by sim_vehicle.py /
+# dronekit-sitl). These regenerate every SITL run; not source.
+/eeprom.bin
+/mav.parm
+/mav.tlog
+/mav.tlog.raw
+/logs/
+/terrain/
+
+# Session-local working notes (per-session handoff docs, run findings,
+# upstream PR drafts). Not part of the source.
+.project_notes/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -65,6 +65,8 @@ dora run examples/python-dataflow/dataflow.yml --uv --stop-after 10s
 | `apis/rust/node` | dora-node-api | Rust API for writing custom nodes |
 | `apis/rust/operator` | dora-operator-api | Rust API for writing in-process operators |
 | `apis/python/node` | dora-node-api-python | Python node API (PyO3) |
+| `libraries/extensions/mavlink2-bridge` | dora-mavlink2-bridge | MAVLink 2 ↔ Apache Arrow conversion (common dialect) |
+| `binaries/mavlink2-bridge-node` | dora-mavlink2-bridge-node | Daemon-spawnable MAVLink 2 bridge (TCP/UDP/serial) |
 
 ## Architecture
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3466,6 +3466,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06432fb54d3be7964ecd3649233cddf80db2832f47fec34c01f65b3d9d774983"
 
 [[package]]
+name = "ioctl-rs"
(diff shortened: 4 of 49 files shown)


heyong4725:
Review findings:

- `binaries/mavlink2-bridge-node/src/main.rs:290-303` - The bridge can hang on shutdown. After `Event::Stop`, main breaks and calls `reader_handle.join()`, but the reader thread may still be blocked in `conn.recv()`. Dropping `rx` only helps after the reader receives another frame and tries `tx.send()`. If the MAVLink peer is quiet or disconnected, `dora stop` / `--stop-after` can wait until external process kill. Use an interruptible receive path, close/wake the connection before joining, or avoid joining a permanently blocking reader.

- `tests/example-smoke.rs:1348-1384` - The new MAVLink smoke tests can pass without proving bridge telemetry works. The networked helper falls through after timeout without asserting completion, and the local sink exits successfully with `0` heartbeats (`examples/mavlink2-bridge/telemetry-printer-rust/src/main.rs:44-51`, Python has the same pattern). A bridge that starts but never publishes `heartbeat` can still be green. These tests should assert at least one decoded heartbeat or fail on timeout.

- `libraries/extensions/mavlink2-bridge/src/arrow_convert.rs:58-112` - Malformed command inputs can panic the bridge instead of returning a decode error. The readers use `as_primitive::<...>().value(0)` directly, so wrong column types, zero-row batches, or null row 0 can panic inside `from_record_batch`; `handle_input` expects decode failures to be `Result` errors. Since these are public dora inputs like `command_long_cmd`, this should validate schema/row/nulls and return `BridgeError`.

Validation:
- `cargo test -p dora-mavlink2-bridge` passed
- `cargo test -p dora-mavlink2-bridge-node` passed

heyong4725:
This is a strong scoped foundation, but I don’t think it’s mergeable yet because there are still a few correctness/reliability gaps that are more serious than polish issues.

### Must-fix

- `binaries/mavlink2-bridge-node/src/main.rs:290-303` — **shutdown can hang**
  - After `Event::Stop`, main breaks and calls `reader_handle.join()`, but the reader thread may still be blocked forever in `conn.recv()`.
  - Dropping `rx` only helps after the reader receives another frame and tries `tx.send()`. If the MAVLink peer is quiet or disconnected, stop can stall until external kill.
  - This needs an interruptible receive path, explicit connection shutdown/wake, or a different threading/shutdown design.

- `tests/example-smoke.rs:1348-1384` + `examples/mavlink2-bridge/telemetry-printer-rust/src/main.rs` + `examples/mavlink2-bridge/telemetry_printer.py` — **smoke tests can pass without proving the bridge actually delivered telemetry**
  - The networked smoke helper falls through after timeout without asserting completion.
  - The telemetry-printer sinks also exit successfully even if they decoded `0` heartbeats.
  - So a bridge that starts but never publishes `heartbeat` can still go green. That makes the current validation weaker than the PR description suggests.
  - These tests should fail on timeout and assert that at least one heartbeat was decoded.

- `libraries/extensions/mavlink2-bridge/src/arrow_convert.rs:58-112` — **malformed command inputs can panic the bridge**
  - The `read_*` helpers use `.as_primitive::<...>().value(0)` directly.
  - Wrong column type, zero-row batch, or null row 0 can panic during `from_record_batch`, while `handle_input` is written as if decode failures are ordinary `Result` errors.
  - Since these are public dora inputs like `command_long_cmd`, this should validate schema/row/nullability and return `BridgeError` instead of crashing the process.

### Should-fix / follow-up

- The transport API is currently narrower than the general “open MAVLink transport from URL” framing suggests:
  - `udp://host:port` always maps to `udpin:` server mode.
  - That may be fine for this PR’s intended usage, but it’s worth documenting as an explicit limitation rather than sounding more general than it is.

- Coverage is missing for **negative decode cases**
  - The roundtrip tests are good happy-path coverage, but they don’t cover malformed Arrow inputs, which is exactly where the current implementation is fragile.

### Why I don’t think this is mergeable yet

Because the remaining issues are about **runtime reliability and false confidence from tests**, not just cleanup:
- the node can hang during shutdown
- malformed external input can crash it
- CI can pass without proving the bridge actually forwarded telemetry

That’s enough for me to call it **not merge-ready yet**. Once those are fixed, I think the overall direction and scope look reasonable.

heyong4725:
Follow-up review on the latest PR head: the earlier blockers I called out now look addressed.

What changed:

- **Shutdown hang on silent peer**: the bridge now has a transport-aware shutdown path with a shared shutdown flag, UDP self-wake (`wake_udp_recv`), bounded grace, and regression tests covering silent-UDP-peer exit and `0.0.0.0` bind wake behavior.
- **False-green smoke tests**: the Rust and Python telemetry sinks now fail if no first HEARTBEAT arrives within the warmup deadline, and also fail if STOP arrives with `0` decoded heartbeats.
- **Malformed Arrow inputs panicking the bridge**: the Arrow decode path now validates type / zero-row / null-at-row-0 before reading row 0, and there are new negative tests for those cases.

Validation rerun:
- `cargo test -p dora-mavlink2-bridge -p dora-mavlink2-bridge-node` passed

So: I no longer see the previous three must-fix findings as open. The shutdown path is understandably more transport-specific now, but the added tests are the right kind of evidence, and this is in much better shape.

Issue #1716, opened 2026-04-21
ci: rebalance PR CI -> fast Linux-only gate, move breadth to nightly


heyong4725:
## Context

PR CI currently takes **~3.5-4 hours wall-clock** on a full run. For agentic AI engineering where PR rate is high, this feedback loop is too slow. Meanwhile the nightly suite runs **~30-37 minutes** on a single platform — under-utilizing the "runs overnight, 4 hours is fine" window.

## Goal

- **PR CI**: ~30-45 min critical path, Linux-only, fast correctness gates. Developers who need cross-platform verification run `make qa-*` locally.
- **Nightly**: ~3-4 hours, broad coverage including macOS + Windows + ROS2 + benchmarks across all platforms.

### Stay in `.github/workflows/ci.yml` (PR + main push)

| Job | Current duration | Why PR-critical |
|---|---|---|
| `fmt` | ~1 min | Style gate, fast |
| `typos` | ~7s | Spelling gate, trivially fast |
| `audit` (cargo-audit + deny) | ~10s | Supply-chain security |
| `unwrap-budget` | ~5s | Production `.unwrap()` ratchet |
| `check-license` | <1 min | Legal compliance |
| `clippy` | ~30 min | Correctness gate, catches real bugs |
| `test` (**ubuntu-latest only**) | ~17 min | Core correctness, primary deployment target |
| `contract-tests` | ~22 min | Behavior contracts (`docs/testing-capabilities.md` promises) |
| `e2e` | ~25 min | WS control-plane + fault-tolerance regression |
| `bench` (regression check) | ~12 min | Perf ratchet |
| `msrv` | ~10 min | **paths-filtered** — only runs if `**/Cargo.toml` or `Cargo.lock` changed |

**Critical path**: ~30 min (clippy dominates; others run in parallel).

### Move to `.github/workflows/nightly.yml`

| Job | Current location | Notes |
|---|---|---|
| `test` (macos-latest, windows-latest) | ci.yml matrix | Drop from PR; recover via `make qa-test` locally |
| `examples` (all 3 platforms) | ci.yml matrix | Redundant with nightly `smoke-suite`; consolidate. If specific examples (C/C++/CMake/Arrow) aren't in `example-smoke.rs`, port them into nightly as a separate job. |
| `cli` (CLI Tests, all 3 platforms) | ci.yml matrix | Longest jobs (80 min Windows). Nightly-only. |
[...]

PR #1796, opened 2026-05-12
bugfix/arm-misaligned-access

SunSunSun689:
- On ARM64 devices with strict memory alignment requirements, like NVIDIA Jetson NX, `node.send_output()` causes a SIGBUS crash when transferring image data larger than 4 KB via Dora.  
- The root cause is that Zenoh's shared memory allocator (Talc) returns a pointer with arbitrary alignment. On the receiving side, it is zero-copy wrapped as an Arrow Buffer, which requires 64-byte alignment for NEON SIMD instructions.  
- The Carmel core on Jetson NX raises a bus error on unaligned access. In contrast, x86 and some other ARM platforms tolerate unaligned access, so the crash only happens on Jetson NX.  
- The solution is to enforce alignment (e.g., at least 64 bytes) during shared memory allocation, following the heap path which already uses `AVec<u8, ConstAlign<128>>`.  
- This fix ensures that returned pointers satisfy strict alignment constraints, resolving all alignment-related crashes (atomic variables and Arrow Buffer) while keeping cross-platform compatibility.


heyong4725 (changes requested):
## Summary

Fixes a real SIGBUS on NVIDIA Jetson NX (Carmel cores) when sending Arrow payloads >4 KiB via zenoh SHM. The root-cause analysis in the commit body is excellent: Zenoh's Talc sub-allocator silently ignores `AllocAlignment` on some aarch64 platforms, so when the receiver wraps the SHM pointer as an `arrow::buffer::Buffer`, NEON SIMD loads fault on the unaligned base. Three-layer defense:

1. `node/mod.rs` — pass `MemoryLayout::new(len, AllocAlignment::new(6))` to `provider.alloc()` so the SHM allocation *requests* 64-byte alignment.
2. `node/arrow_utils.rs` — pad every Arrow buffer offset to `max(arrow_align, 64)` so `shm_base + offset` is 64-byte aligned even for `BitMap` / `VariableWidth` buffers.
3. `event_stream/mod.rs` — on the receive side, if the SHM base is still misaligned, copy into a 64-byte-aligned heap `Vec` instead of zero-copying it.

The fix is correct in principle. The concerns below are about CI hygiene, missing tests, and a few polish items.

---

## Critical (blocking)

### 1. CI is red — fmt + clippy fail

Verified locally — `cargo fmt --check -p dora-node-api` and `cargo clippy -p dora-node-api -- -D warnings` both fail. Per `CLAUDE.md`, fmt and clippy are mandatory pre-push gates.

- **Format**: `apis/rust/node/src/event_stream/mod.rs:1127-1128` — the `let ptr = NonNull::new(...)` chain needs to be on a single line per rustfmt.
- **Clippy**: `apis/rust/node/src/event_stream/mod.rs:1125` — `manual_is_multiple_of`. Use `(slice.as_ptr() as usize).is_multiple_of(REQUIRED_ALIGN)`.

### 2. No tests for the bug fix

Per `CLAUDE.md`'s TDD policy (\"All new features and bug fixes must follow RED-GREEN-IMPROVE\"), and given this PR changes a load-bearing invariant of the wire format, the absence of tests is the biggest correctness risk. Suggested minimum:

- **Unit test for `arrow_utils.rs`**: build an `ArrayData` with mixed buffer kinds (FixedWidth, BitMap, VariableWidth, plus a child) and assert that every `BufferOffset` reported by `copy_array_into_sample` is a multiple of 64. This is testable without any hardware and protects the invariant from future refactors.
- **Unit test for the receiver fallback** in `event_stream/mod.rs`: synthesize a misaligned `&[u8]` (heap-allocate `slice.len()+1`, slice from offset 1) and verify the resulting `arrow::buffer::Buffer` is 64-byte aligned. Without this, the `else` branch is dead code on every CI runner (x86 / aligned ARM), so it can rot silently.

---

## Should fix

### 3. Receiver-side fallback is silent

The fallback heap copy fires on every >4 KiB message on a misbehaving allocator (i.e. constantly on Jetson NX) but emits no log or metric. A one-shot or rate-limited `tracing::warn!(\"zenoh SHM payload misaligned ({base:#x}), copying to aligned buffer\")` would let operators confirm the fix is engaged and surface the hidden performance cost (it defeats zero-copy for the worst-case path).

### 4. Receiver fallback comment is slightly optimistic

`event_stream/mod.rs:1149-1154`:
> \"Misaligned SHM pointer ... copy into a 128-byte-aligned heap buffer so that every Arrow buffer within the payload (whose intra-payload offsets are already multiples of 64) ends up at an absolute address that is also 64-byte aligned.\"

This is true only when the sender ran the new `arrow_utils.rs` code. A mixed-version deployment (new receiver / old sender) would still have intra-payload offsets at 4 / 8 / 16-byte boundaries, so the fallback wouldn't rescue it. Either tighten the comment (\"assuming the sender ran ≥ this version\") or note the cross-version caveat. (Also: it copies into a 64-byte-aligned buffer, not 128 — the comment says 128.)

### 5. Latent bug fix in `copy_array_into_sample_inner` is unmentioned

The reorder (align → assert → copy) silently fixes a pre-existing issue: the old code asserted with the unaligned `*next_offset` but copied with the aligned one, so the bounds check could pass while the write landed beyond the assertion's view. In practice it was masked by `required_data_size` over-allocating, but the assert was lying. Worth a one-liner in the commit body so it isn't lost.

---

## Minor

### 6. Duplicated alignment logic in `arrow_utils.rs`

Lines 61-67 and 98-104 are identical 7-line blocks computing the per-buffer alignment. Extract:

```rust
fn buffer_alignment(spec: &BufferSpec) -> usize {
    let arrow_align = match *spec {
        BufferSpec::FixedWidth { alignment, .. } => alignment,
        _ => 1,
    };
    arrow_align.max(ARROW_SHM_ALIGNMENT)
}
```

This guarantees the two callers can never drift, which is important because they have to agree byte-for-byte.

### 7. Hand-rolled pad math when `align_offset` exists

`event_stream/mod.rs:1157-1159`:
```rust
let base = aligned.as_ptr() as usize;
let pad = REQUIRED_ALIGN - (base % REQUIRED_ALIGN);
let pad = if pad == REQUIRED_ALIGN { 0 } else { pad };
```

`<*const u8>::align_offset` does exactly this:
```rust
let pad = aligned.as_ptr().align_offset(REQUIRED_ALIGN);
```

### 8. Per-buffer 64-byte padding on platforms that don't need it

The change pads BitMap / VariableWidth buffers from 1-byte to 64-byte alignment unconditionally, which is wasteful on x86 (and most ARM). For typical Dora payloads (images, point clouds) the overhead is rounding error, but consider a `cfg!(target_arch = \"aarch64\")` gate if you want to keep parity with Arrow's spec on other platforms. (Acceptable to leave as-is; flagging for awareness.)

### 9. Off-by-one in capacity

`Vec::with_capacity(slice.len() + REQUIRED_ALIGN)` reserves one byte more than needed (max pad is 63, not 64). Cosmetic.

---

## Positive

- Excellent commit-body root-cause analysis with platform specifics (Jetson NX, Carmel, L4T, Talc).
- Inline comments answer *why* (the Talc bug, the alignment invariant) not just *what*.
- Three-layer defense is sound — each layer is sufficient on its own; together they degrade gracefully.
- Uses the documented zenoh API (`MemoryLayout` + `AllocAlignment::new`) rather than a workaround.
- `expect` messages on `MemoryLayout::new` / `AllocAlignment::new(6)` correctly state the preconditions that make them infallible.
- The SAFETY comment on the receiver-side Vec ownership is precise (Vec pointer is stable across `resize`/`extend_from_slice` because capacity was reserved up-front; moving the Vec into the Arc doesn't move the heap allocation).

---

## Recommendation

**Request changes.** Block on (1) and (2) — fmt/clippy and at least one unit test for each new code path. (3)-(5) are strong nice-to-haves; (6)-(9) are polish.

The underlying fix is good engineering and clearly needed for Jetson NX users — it just needs to clear the project's own quality gates before landing.

heyong4725:
Thanks for tracking down the Jetson NX SIGBUS — this is a real and well-explained bug, and the alignment-fix logic in `patches/zenoh-shm/src/api/protocol_implementations/posix/posix_shm_provider_backend_talc.rs:85-104` is correct and well-commented (advancing the Talc heap base to the next 64-byte boundary so heap-relative alignment math produces absolutely-aligned pointers — exactly the right shape of fix).

Unfortunately the PR as currently packaged has three blocking problems before the fix can land. Listing them in priority order.

---

## P0 — branch is stale, merging would revert ~18 already-landed PRs

`git diff origin/main..HEAD --stat` shows the PR removing major chunks of recently-merged work, e.g.:

| File | Lines lost | Source PR |
|---|---|---|
| `binaries/cli/src/command/doctor.rs` | -169 | [#1818](https://github.com/dora-rs/dora/pull/1818) (merged 2026-05-14) |
| `apis/c/node/node_api.h` | -55 | [#1811](https://github.com/dora-rs/dora/pull/1811) |
| `examples/ros2-bridge/c++/action-client/*` | whole dir | [#1816](https://github.com/dora-rs/dora/pull/1816) (merged 2026-05-14) |
| `examples/ros2-bridge/rust/action-client/*` | whole dir | [#1816](https://github.com/dora-rs/dora/pull/1816) |
| `libraries/extensions/ros2-bridge/msg-gen/src/lib.rs` | -229 | [#1813](https://github.com/dora-rs/dora/pull/1813) |
| `libraries/core/src/build/build_command.rs` | -345 | [#1820](https://github.com/dora-rs/dora/pull/1820) (merged 2026-05-15) |
| `libraries/core/src/build/mod.rs` | -327 | [#1820](https://github.com/dora-rs/dora/pull/1820) |
| `.github/workflows/nightly.yml` | -174 | [#1716](https://github.com/dora-rs/dora/pull/1716) (CI rebalance) |
| `Cargo.toml` (removes `tokio-stream`) | -1 | needed by [#1820](https://github.com/dora-rs/dora/pull/1820) and [#1828](https://github.com/dora-rs/dora/pull/1828) |
| `Changelog.md`, `docs/cli.md`, `docs/python-guide.md`, `docs/debugging.md`, `guide/src/operations/cli.md`, `examples/python-*/README.md` | various | recent doc work |

The branch needs to rebase against current `main` before anything else. Most of these conflicts should resolve cleanly — your fix doesn't touch any of these files. Just `git fetch origin main && git rebase origin/main`.

## P1 — vendoring 9000+ LOC of zenoh-shm to apply a ~20-line fix needs discussion

The PR adds `patches/zenoh-shm/` (9k LOC, full fork of zenoh-shm 1.x) and wires it via `[patch.crates-io] zenoh-shm = { path = "patches/zenoh-shm" }`. This is a major design decision the maintainers should weigh in on, because:

- We now own a copy of zenoh-shm and must keep it in sync with upstream every time we bump zenoh
- The fix is **eligible to upstream** — eclipse-zenoh/zenoh would almost certainly accept this fix, it's a clear correctness improvement for any ARM user
- If we want a local patch for now, `[patch.crates-io] zenoh-shm = { git = "https://github.com/<you>/zenoh", rev = "<sha>" }` is **dramatically lower maintenance** than vendoring the source

Suggested order of preference:

1. **Upstream the fix to eclipse-zenoh/zenoh.** This is the right long-term home — your bug report (Talc base alignment vs Arrow 64-byte requirement on aarch64) applies to every zenoh consumer that does shared-memory + Arrow, not just dora. I'd be happy to help draft the upstream PR.
2. **Pin to a forked git rev**, not a vendored source tree. `[patch.crates-io] zenoh-shm = { git = "...", rev = "..." }` lets us point at a one-commit branch of upstream zenoh while waiting for the upstream PR to merge. ~5 lines of Cargo.toml change in dora, no vendored source to maintain.
3. **Vendor** only if (1) and (2) are both blocked. If we go this route we'd need a clear maintenance plan (how do we re-vendor when zenoh bumps?) and the vendored crate should have a `VENDORED-FROM.md` documenting the upstream commit and the local delta.

In all three cases the *actual code change* — the talc-base-alignment fix — is the same; only the packaging differs.

## P1 — `Cargo.toml` removes `tokio-stream` workspace dep

Line 135 of the PR removes `tokio-stream = "0.1.11"` from `[workspace.dependencies]`. After rebase against current `main`, `tokio-stream` is now used by `libraries/core/src/build/build_command.rs` (from [#1828](https://github.com/dora-rs/dora/pull/1828) "keep draining build output streams"). The removal would break the workspace build. Re-add the line during rebase.

---

## Verification of the actual fix

To be clear: the alignment math is correct.

```rust
const ARROW_ALIGN: usize = 64;
let base_addr = ptr as usize;
let padding = base_addr.wrapping_neg() & (ARROW_ALIGN - 1);
let (aligned_ptr, usable_size) = if padding < real_size {
    (unsafe { ptr.add(padding) }, real_size - padding)
} else {
    (ptr, real_size)
};
```

- `base_addr.wrapping_neg() & (ARROW_ALIGN - 1)` gives the bytes-to-advance to hit the next 64-byte boundary (standard `align_up(addr, 64) - addr` written without the conditional). Correct.
- The `padding < real_size` guard prevents pointer-add overflow when the segment is smaller than the alignment. Correct, though in practice mmap'd segments are page-multiples (4 KiB+) so this branch never fires; keeping it is defensive and fine.
- 63-byte worst-case waste from a multi-MiB segment is negligible. Correct trade-off.

A unit test would help — ideally one that asserts `(aligned_ptr as usize) % 64 == 0` for a few synthetic base addresses (`0x...1`, `0x...10`, `0x...3F`, `0x...40`). Easy to add once the rebase is done.

---

## Suggested next step

1. Rebase against current `main` (will drop the false deletions automatically).
2. Decide on packaging (upstream PR + temporary git-rev patch is my recommendation — happy to help with the upstream draft).
3. Add the alignment unit test mentioned above.

I'll keep this PR in review until the rebase + packaging questions are sorted. The fix is real and worth landing — let's just get it packaged right.

heyong4725:
Hey @SunSunSun689 — I want to follow up on my last comment with a friendlier read, since I realize that comment got pretty dense and it's your first PR here. The short version: **the fix you wrote is good. The packaging stuff is on the maintainers, not on you to figure out alone.** Here's what you got right and what the next concrete step is.

## What you got right

Your bug report and your fix are both genuinely good engineering, and I want to be specific about why:

1. **You correctly diagnosed a hardware-specific bug.** SIGBUS on Jetson NX from `node.send_output()` with >4 KB images is exactly the kind of issue that's painful to debug — it doesn't reproduce on x86, the stack trace points at zenoh internals, and the "obvious" Arrow buffer alignment check passes if you only look at the heap path. Connecting "Talc returns arbitrary alignment" + "Arrow needs 64-byte for NEON" + "Carmel core is strict about unaligned access" is the right mental model and most people would not get to that diagnosis quickly.

2. **The fix is in the right place.** Aligning the Talc heap base (not trying to align each allocation post-hoc) means Talc's existing heap-relative alignment math just works. That's a structural fix, not a band-aid.

3. **The math is correct.** `base_addr.wrapping_neg() & (ARROW_ALIGN - 1)` is the standard "bytes to next alignment boundary" without a branch, and your `padding < real_size` guard prevents pointer-add overflow on a too-small segment. A senior Rust engineer would write it this way.

4. **The comment in the code is what we want to see.** Five lines explaining the *why* (Talc relative math + mmap's 16-byte alignment + aarch64 NEON requirement) that someone reading this in six months can follow. Future-you and future-us will thank you for that comment.

## The "destructive deletion" thing isn't your fault

`main` has moved fast in the last two weeks — six PRs landed (#1716, #1811, #1813, #1816, #1818, #1820) while you were working on this. Your branch was forked before any of them, so the diff against current main *looks* like you're deleting things, but you're really just at an old point in time. A rebase fixes all of it in one step.

## What you should do (one thing)

```bash
git fetch origin main
git rebase origin/main
```

Conflicts you'll see are all "the file changed in main, you didn't touch it." For each one: take main's version. Concretely, if `git mergetool` or VS Code asks, pick "Current" or "incoming" — whatever the menu calls "the version from main." None of those files were yours to change.

After:

```bash
git push --force-with-lease
```

That's it. After the rebase the diff should drop to roughly +9000/-0 (just your `patches/zenoh-shm/` additions and the `Cargo.toml` patch line). If you hit any conflicts that aren't obvious "keep main's version," ping me here and I'll help.

## What we'll handle together (don't worry about these yet)

Two design questions I asked about earlier — vendoring vs git-patch, and submitting upstream to eclipse-zenoh — those are real questions but they're **maintainer-side decisions, not first-PR decisions**. I'll work with @phil-opp on what to recommend, and then we'll pair on whichever direction we pick. You'll learn how Cargo's `[patch.crates-io]` works either way, but you shouldn't have to decide between three packaging strategies on your first PR. That's on us.

The upstream PR to eclipse-zenoh, if we go that route, is also something I'll help drive — first-time interactions with another OSS project's review process is not where you should be flying solo.

## Summary

- **Today:** rebase. That's it. ~15 min.
- **This week:** we'll talk about packaging and decide together.
- **Later:** maybe an upstream PR (with you as primary author, me handling the back-and-forth).

Your bug analysis was sharp and the fix is correct. We'll get this landed.

heyong4725:
## Review

Thanks for tracking this down — Jetson NX SIGBUS on SHM Arrow zero-copy is a real bug with a non-obvious root cause, and the symptom triangulation (Talc heap base + 16-byte SHM mmap + 64-byte NEON requirement) is solid investigative work.

**Recommendation:** changes-requested. The fix is correct in principle, but the PR has 4 blocking issues — one CI failure, one legal/attribution gap, two structural concerns — before it can land.

## Diff scope

- ~150 lines of actual fix in `apis/rust/node/src/{event_stream,node,arrow_utils}.rs`
- ~9000 lines of vendored upstream `zenoh-shm 1.8.0` in `patches/zenoh-shm/`, of which **~20 lines are the actual modification** (`posix_shm_provider_backend_talc.rs:86-104`)
- `[patch.crates-io]` redirect in workspace `Cargo.toml`

The fix has three defensive layers: (1) sender requests `AllocAlignment::new(6)`; (2) patched Talc backend advances the heap base to a 64-byte boundary; (3) receiver does an alignment check + heap-copy fallback if SHM ptr is misaligned. Belt-and-suspenders-and-rope — understandable given Jetson NX history, but it deserves scrutiny (Finding 4).

---

## Finding 1 (BLOCKER) — CI clippy gate fails

```
error: manual implementation of `.is_multiple_of()`
    --> apis/rust/node/src/event_stream/mod.rs:1125:16
1125 |   if slice.as_ptr() as usize % REQUIRED_ALIGN == 0 {
     |      help: replace with: `(slice.as_ptr() as usize).is_multiple_of(REQUIRED_ALIGN)`
```

`cargo clippy -p dora-node-api -- -D warnings` fails locally on rust 1.92. Replace with `(slice.as_ptr() as usize).is_multiple_of(REQUIRED_ALIGN)` and the PR will be clippy-clean. PR CI's Clippy job will block merge until this is fixed.

## Finding 2 (HIGH) — Missing license files in vendored `patches/zenoh-shm/`

`patches/zenoh-shm/Cargo.toml` declares `license = "EPL-2.0 OR Apache-2.0"`, but the directory contains:

```
patches/zenoh-shm/
├── Cargo.toml           ✓
├── Cargo.toml.orig      ✓
├── .cargo_vcs_info.json ✓
├── README.md            ✓
├── build.rs
├── src/
└── tests/
```

No `LICENSE`, `LICENSE-EPL-2.0`, `LICENSE-APACHE`, `NOTICE`, or `COPYING`. Both licenses **require** preserving the original copyright/license texts when redistributing source. EPL-2.0 §3.1 is explicit: "Each Contributor ... must include with the Source Code Form a copy of this Agreement."

Since dora-rs is distributing this directory tree (ships in any tarball/clone), this is a strict-compliance issue and may also trip dora's existing `cargo-lichking` license-check CI once the patched crate is consumed from a path source.

**Fix:** copy `LICENSE-EPL-2.0` and `LICENSE-APACHE-2.0` from upstream `https://github.com/eclipse-zenoh/zenoh/tree/29b3e63a/commons/zenoh-shm` into `patches/zenoh-shm/`. The vcs sha is already pinned in `.cargo_vcs_info.json:3` (`29b3e63a311ab55f438280f3e71c6268b6f115cc`) so the tree to copy from is unambiguous.

## Finding 3 (MEDIUM) — No upstream PR reference; patch will bitrot silently

`patches/zenoh-shm/README.md` is the upstream README verbatim, with no note that this directory is a dora-rs overlay or what was modified. A future maintainer who lands a `zenoh = "1.9"` bump in `Cargo.toml` may not realize that the `[patch.crates-io]` redirect is silently still pinning to 1.8 (cargo warns, but warnings are easy to miss).

Has this fix been proposed upstream to `eclipse-zenoh/zenoh`? That's the path to closing this out; `patches/` is at best a 1-2 release stopgap.

**Fix:** add `patches/zenoh-shm/PATCH-NOTES.md` documenting:

1. Why this overlay exists (1 paragraph, link the Jetson NX SIGBUS issue)
2. Exactly which file/lines were modified (1 line: `src/api/protocol_implementations/posix/posix_shm_provider_backend_talc.rs:86-104`)
3. Upstream issue/PR link if filed (or a TODO if not)
4. When this overlay can be removed (zenoh-shm version that contains the fix)

~30 lines, saves the next maintainer a day of detective work.

## Finding 4 (MEDIUM) — Receiver-side heap-copy fallback has zero test coverage

`apis/rust/node/src/event_stream/mod.rs:1108-1170` adds a runtime alignment check on every SHM receive. If the SHM ptr is **not** 64-byte aligned, the fallback copies the entire payload into a manually-padded `Vec<u8>` and wraps it in a custom `Allocation`.

That fallback is reached **only** on platforms where the SHM allocator misbehaves — Jetson NX / L4T per the PR description. CI runs Linux/macOS/Windows on x86_64 (and nightly cross-checks aarch64), none of which trigger the fallback. So:

- The manual alignment math (`let pad = REQUIRED_ALIGN - (base % REQUIRED_ALIGN); let pad = if pad == REQUIRED_ALIGN { 0 } else { pad };`) is not exercised in CI
- The `Vec<u8>` no-realloc guarantee (reserve `slice.len() + 64`, then resize+extend to at most `slice.len() + 63` — works but non-obvious) is not exercised
- A future refactor that breaks the fallback path will not fail CI

**Two ways to fix:**

(a) **Force the fallback in a test.** Add a unit test that constructs an Arrow `Buffer` from a deliberately-misaligned slice and verifies the result is 64-byte aligned. Doesn't need real Jetson hardware — synthetic misalignment is enough.

(b) **Use `aligned-vec` directly** (already pulled in via `AVec<u8, ConstAlign<128>>` elsewhere in this crate). Replace the manual padding with:
```rust
let mut aligned = AVec::<u8, ConstAlign<64>>::with_capacity(64, slice.len());
aligned.extend_from_slice(slice);
```
Compiler-enforced alignment, no manual math, no realloc concern. Eliminates an entire class of future regression and removes the need for the unit test.

(b) is strictly better unless there's a perf reason to avoid `aligned-vec` here that I'm missing.

## Finding 5 (LOW) — `arrow_utils.rs` over-pads variable-length buffers

```rust
let arrow_align = if let BufferSpec::FixedWidth { alignment, .. } = *spec {
    alignment
} else {
    1
};
let align = arrow_align.max(ARROW_SHM_ALIGNMENT);
```

For non-`FixedWidth` buffers (variable-length data, e.g. `Utf8`/`Binary` offset arrays), this now pads to 64 bytes even though Arrow itself only requires 1-byte alignment. That's fine — over-alignment is harmless — but every variable-length array now eats up to 63 bytes of SHM padding per buffer where it previously ate up to `arrow_align - 1`. On Jetson NX this is the correct tradeoff (alignment > tiny SHM waste). Worth a one-line comment in the code so the next maintainer doesn't try to "optimize" the padding back to the spec minimum.

## Finding 6 (LOW) — Defense-in-depth has runtime cost on every SHM receive

The receiver-side alignment check + branch executes on **every** SHM message receive on **every** platform. For hot paths handling small Arrow messages this is ~5ns of branch-predictor noise per message — measurable on benchmark workloads.

If the upstream Talc patch is sound, the receiver-side fallback should be dead code on every supported platform. Two options:

(a) `#[cfg(target_arch = "aarch64")]`-gate the receiver-side fallback — x86/x86_64 skips the check entirely.

(b) Document explicitly that this check is intentionally redundant safety, accept the perf cost.

(a) is cleaner; (b) is fine if measurement shows the cost doesn't matter for any realistic workload.

---

## What's good

- The root-cause analysis (Talc heap-base × Arrow buffer offsets × aarch64 NEON) is correct and well-explained in both PR description and code comments. Future maintainers will be able to follow the reasoning.
- The vendored patch in `posix_shm_provider_backend_talc.rs:86-104` is **minimal** — 20 lines, surgical, no other vendored code modified. Exactly the right shape for a vendored overlay.
- Alignment math `base_addr.wrapping_neg() & (ARROW_ALIGN - 1)` is a nice idiom for "round up to a power-of-2 boundary using only bitwise ops" — correct and branch-free.
- `arrow_utils.rs` correctly updates **both** `required_data_size_inner` (size pre-calc) AND `copy_array_into_sample_inner` (actual copy) to use the same `max(arrow_align, ARROW_SHM_ALIGNMENT)` rule. A common bug shape on this kind of fix is to update one and not the other; this PR avoided it.
- Sender-side `MemoryLayout::new(data.len(), AllocAlignment::new(6))` correctly uses log2 form (`new(6)` = 2^6 = 64), matching the comment.

## Validation done in review

```
cargo check -p dora-node-api                     clean (29s)
cargo clippy -p dora-node-api -- -D warnings     FAIL — Finding 1
```

Did not run full test suite or benchmarks. Recommend running `make qa-fast` locally before pushing the fixed version.

## Path forward

Fix the 4 blockers/highs (Findings 1-4); the LOWs are nice-to-haves. Once Findings 1+2 are resolved this can probably ride into nightly to verify aarch64 cross-check on real hardware; access to a Jetson NX for actual end-to-end SIGBUS-reproduction-then-fix-verification would be ultimate validation but isn't a blocker for merge.

---

🤖 Adversarial pre-landing review by Claude. The fix is fundamentally correct; the blockers are packaging and CI-gate issues, not flaws in the alignment logic itself.

phil-opp (commented):
Re-reviewed after the recent push. The alignment fix itself remains correct and is well-explained — no concerns with the core logic.

Good news: the **P0 stale-branch / destructive-deletion** problem from the earlier reviews is resolved. The current diff against `main` is 70 added + 5 modified files with **0 removed**, so the ~18 reverted PRs are gone — the rebase worked.

However, several blockers raised in the earlier reviews are **still outstanding** after this push (flagging them again only because there were new commits and they weren't addressed):

1. **clippy still fails (CI blocker).** `event_stream/mod.rs` still has `if slice.as_ptr() as usize % REQUIRED_ALIGN == 0`, which trips `clippy::manual_is_multiple_of` under `-D warnings`. Replace with `(slice.as_ptr() as usize).is_multiple_of(REQUIRED_ALIGN)`. The PR CI clippy gate will block merge until this is changed.

2. **Vendored `patches/zenoh-shm/` still ships no license files.** The tree contains `Cargo.toml`, `README.md`, `build.rs`, etc., but no `LICENSE-EPL-2.0` / `LICENSE-APACHE` / `NOTICE`. Both licenses require preserving the original texts on redistribution, and this will likely trip the `cargo-lichking` license check once the patched crate is consumed from the path source.

3. **No tests for the new receiver-side fallback.** The misaligned-SHM heap-copy branch in `event_stream/mod.rs` is still untested. It's dead code on every CI platform (x86 / aligned ARM), so the manual pad math can rot silently. A synthetic-misalignment unit test asserting the resulting `arrow::buffer::Buffer` base is 64-byte aligned would cover it without Jetson hardware — or switching that branch to `AVec<u8, ConstAlign<64>>` removes the manual math entirely.

4. **Packaging decision (vendor vs. upstream/git-rev patch) still open.** Still vendoring ~9k LOC for a ~20-line change, with no `PATCH-NOTES.md` documenting the upstream sha and the local delta. This is the maintainer-side call flagged earlier; it should be settled before this lands so the patch doesn't bitrot on the next zenoh bump.

Minor (carried over): the fallback comment still says "128-byte-aligned heap buffer" while the code aligns to 64 (`REQUIRED_ALIGN = 64`).

So: the rebase cleared the biggest problem, but (1) and (2) in particular still need fixing before this is mergeable.

---

phil-opp:
**Automated review — I found one important issue.**

**`apis/rust/node/src/node/mod.rs:1086` — `MemoryLayout::new` panics for any payload whose length is not a multiple of 64.**

```rust
let shm_layout = MemoryLayout::new(
    data.len(),
    AllocAlignment::new(ARROW_ALIGN_POW).expect("valid Arrow alignment"), // 64
)
.expect("SHM layout: size is non-zero");
```

The adjacent SAFETY comment claims the only failure mode is the `NonZeroUsize` conversion (zero size). That is not what `MemoryLayout::new` does. In the vendored patch (`patches/zenoh-shm/src/api/provider/memory_layout.rs:57-61`):

```rust
// size of a layout must be a multiple of its alignment!
match size.get() % alignment.get_alignment_value() {
    0 => Ok(Self { size, alignment }),
    _ => Err(ZLayoutError::IncorrectLayoutArgs),
}
```

So with `alignment = 64`, `MemoryLayout::new(data.len(), 64)` returns `Err` whenever `data.len() % 64 != 0`, and the `.expect(...)` then **panics the node**. This path is taken for every payload `>= ZENOH_SHM_MIN_PAYLOAD` (512 bytes), so any image whose serialized length is not a multiple of 64 — i.e. the common case — crashes. It also replaces a previously-infallible path: the old `provider.alloc(data.len())` went through `TryFrom<usize> for MemoryLayout`, which uses 1-byte alignment and always succeeds. As written, the fix would panic on exactly the >4 KB image payloads it is meant to fix.

Suggested fix: round the requested size up to the alignment before building the layout (e.g. `data.len().div_ceil(64) * 64`), ensuring the copy still writes exactly `data.len()` bytes; and do not `.expect()` here — fall through to the heap path on `Err`, the way the `provider.alloc(...)` error arm already does.

(I focused on what changed; the previously-raised clippy `is_multiple_of`, vendored-license, fallback test-coverage, and rebase items appear still open but I did not re-litigate them here.)

---

---

phil-opp:
**Automated review by Claude** — _this is a fully automated review; a human has not vetted these findings._

I won't repeat the issues already raised in prior reviews (the `MemoryLayout::new(...).expect(...)` panic on non-64-multiple sizes, the `clippy::manual_is_multiple_of` blocker, the missing vendored-license files, and the untested receiver fallback) — there have been no new commits since those, so they still stand as-is.

One **new** observation that I think is the most decision-relevant point for this PR: **`main` appears to have already independently landed an equivalent — and more complete — version of this fix**, which is likely why this branch now shows merge conflicts (`dirty`). On current `main`:

- **Receiver side** (`apis/rust/node/src/event_stream/mod.rs`): `raw_payload_is_arrow_aligned()` (validates the base pointer **and** every per-buffer offset, lines ~1175/1203), `copy_to_aligned_arrow_buffer()` (lines ~1241, with a regression test `copy_to_aligned_arrow_buffer_realigns_misaligned_payload` at ~1527), and a rate-limited misalignment warning. This handles both the SHM-borrowed and heap-owned paths, whereas this PR only handles the borrowed path.
- **Sender side** (`apis/rust/node/src/node/mod.rs:606`): requests 64-byte alignment at the **provider-pool** layer via `AllocAlignment::new(ARROW_BUFFER_ALIGNMENT_EXPONENT)` + `checked_next_multiple_of(...)` with graceful `.ok()` fallback, deliberately keeping per-allocation `provider.alloc(data.len())` at 1-byte alignment. This avoids the per-allocation `MemoryLayout` panic that this PR's approach hits.
- **`apis/rust/node/src/node/arrow_utils.rs:22-32`**: the `ARROW_BUFFER_ALIGNMENT = 64` constant and `payload_buffer_alignment()` offset-padding helper this PR open-codes are already present, refactored into shared constants.

Given that, the three node-API changes here are superseded by cleaner code on `main`, and the remaining novel piece is the vendored Talc base-alignment patch (`patches/zenoh-shm/...`) — which is sound, but maintainers already suggested upstreaming to eclipse-zenoh rather than vendoring ~9k LOC. It would be worth confirming whether anything in this PR is still needed on top of `main` before investing more in the rebase; it may be closeable.

---

phil-opp:
Closing as superseded. The Jetson NX SIGBUS this diagnoses is real and the analysis was useful, but all three layers have since landed on `main` in better form: the SHM provider is built with a 64-byte `AllocAlignment` (`apis/rust/node/src/node/mod.rs`), the IPC encoder maintains a 64-aligned body-offset invariant, and the receiver uses `StreamDecoder` with `require_alignment = false` so arrow-rs copies under-aligned buffers instead of faulting.

This branch is also not rebaseable: its `arrow_utils.rs` hunk edits APIs deleted by the Arrow-IPC wire-format migration, and the vendored `zenoh-shm` is pinned at 1.8.0 while the workspace is on ~1.9, so the `[patch.crates-io]` entry would not apply.

If `main` still reproduces the SIGBUS on a Jetson NX, please open a fresh issue with that repro — it would be a small, well-scoped fix (likely upstream at eclipse-zenoh) rather than this branch.

PR #1801, opened 2026-05-12
fix(daemon): spawn runtime via current_exe() instead of PATH lookup (#1797)

### LINKED ISSUE #1797: daemon: spawn runtime via current_exe() instead of which("dora")
Carry-over of [`dora-rs/adora#316`](https://github.com/dora-rs/adora/pull/316) by @DGHX12345 (2026-04-20), originally opened against the fork before the 1.0 consolidation. The fix did not migrate to `dora-rs/dora` and the bug is still present.

## Bug

In `binaries/daemon/src/spawn/spawner.rs` line 314, the daemon resolves the `dora` binary path via `which::which("dora")` to spawn runtime nodes for the non-Python operator path. This picks up whichever `dora` appears first on `$PATH`, which may be a different installation than the daemon itself — common when:

- A user has both a system-installed `dora` and a `cargo install`-built one.
- The daemon was launched from a venv or a non-default install location.

The result is the daemon trying to spawn an incompatible runtime, often with confusing error messages.

The same call site already exists with the corrected pattern on the Python branch (line 286 uses `std::env::current_exe()`). This issue is to apply the same pattern to the non-Python branch.

## Fix (3 lines, from the original PR)

```rust
} else {
    let current_exe = std::env::current_exe()
        .wrap_err("failed to get current executable path")?;
    let mut cmd = Command::new(current_exe);
    cmd = cmd.arg("runtime");
    Some(cmd)
}
```

Credit: @DGHX12345 — original PR `dora-rs/adora#316`. Suggested reopen against `dora-rs/dora`.

heyong4725:
Closes #1797.

## Summary

- `binaries/daemon/src/spawn/spawner.rs` — non-Python branch of `prepare_node_inner` was resolving the runtime binary via PATH lookup, which can pick a different `dora` than the daemon itself (system vs cargo install, or daemon launched from a venv). Reuse the `current_exe` already computed at line 273 for the sibling Python-detection check, matching the existing pattern in the same arm.
- Carry-over of @DGHX12345 original fix in `dora-rs/adora#316`, which never migrated to `dora-rs/dora` after the 1.0 consolidation.

## Why a structural test

A full behavioral test would need to:
1. Build the daemon binary.
2. Place it at a non-PATH location (or unset PATH).
3. Spawn a non-Python operator dataflow.
4. Verify the runtime is launched from the daemon's location, not a stale `dora` elsewhere on PATH.

That's E2E scope for a 3-line fix. Instead, this PR adds a tiny structural guard in `spawner.rs`'s `#[cfg(test)] mod tests` that includes the source via `include_str!` and asserts the forbidden PATH-lookup pattern never reappears. Confirmed RED before the fix (test fails on un-patched code) and GREEN after.

## Test plan

- [x] `cargo clippy -p dora-daemon -- -D warnings` (matching CI invocation)
- [x] `cargo test -p dora-daemon` — 33/33 pass, including new `runtime_spawn_does_not_use_which_dora`
- [x] RED check: temporarily reverted the fix, confirmed the new test fails with the exact message about #1797
- [ ] Reviewer: confirm the structural test approach is acceptable for this class of "don't use PATH lookup" regression, or request a refactor for proper unit-testing of the command-building logic

### DIFF
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -310,8 +310,13 @@ impl Spawner {
                         ]);
                         Some(cmd)
                     } else {
-                        let mut cmd =
-                            Command::new(which::which("dora").wrap_err("failed to get dora path")?);
+                        // Use the daemon's own executable so we always spawn
+                        // a matching runtime. A PATH lookup for the dora
+                        // binary would pick whatever appears first on PATH,
+                        // which can be a different install (system vs cargo
+                        // install, or a daemon launched from a venv). See
+                        // #1797.
+                        let mut cmd = Command::new(&current_exe);
                         cmd = cmd.arg("runtime");
                         Some(cmd)
                     }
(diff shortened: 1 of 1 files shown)


heyong4725:
No findings from review. The non-Python runtime path now uses the daemon/CLI process executable instead of resolving `dora` from PATH, which addresses #1797 without changing the existing Python-wrapper branch. I checked the daemon/CLI launch shape, the Python packaging edge case already handled by the surrounding branch, and the added structural regression guard.\n\nValidation run locally:\n- `cargo test -p dora-daemon`\n- `cargo fmt --all -- --check`\n- `cargo clippy -p dora-daemon -- -D warnings`\n\nI attempted to submit this as an approving review, but GitHub rejected it because the authenticated account owns the PR.

phil-opp:
Note the review comment of @haixuanTao here: https://github.com/dora-rs/dora/pull/1729#issuecomment-4302100392

PR #1818, opened 2026-05-14
feat(doctor): add /dev/shm permission check on Linux (rescue of #1352)

heyong4725:
Adds `/dev/shm` permission validation to the existing `dora doctor` command on Linux. Rescues the genuinely-useful check from closed PR #1352 and lands it where it belongs.

## What this adds

Inserted between the existing `CLI version` check and the `Coordinator connectivity` check in `binaries/cli/src/command/doctor.rs`:

| Scenario | Output |
|---|---|
| `/dev/shm` not accessible | `WARN Shared memory: /dev/shm not accessible — dora will run via heap fallback; for the zero-copy fast path run \`sudo mkdir -p /dev/shm && sudo chmod 1777 /dev/shm\`` |
| `/dev/shm` exists but is not a directory | `WARN Shared memory: /dev/shm exists but is not a directory — dora will run via heap fallback; for the zero-copy fast path run \`sudo chmod 1777 /dev/shm\`` |
| Wrong mode | `WARN Shared memory: /dev/shm has mode 0o755, expected 0o1777 — dora will run via heap fallback if SHM allocation is denied; for the zero-copy fast path run \`sudo chmod 1777 /dev/shm\`` |
| Mode `1777` | `PASS Shared memory: /dev/shm mode is 1777` |

Non-Linux platforms (macOS, Windows) skip this check silently — the underlying `/dev/shm` mechanism does not apply there.

## Why WARN and not FAIL

The node runtime treats SHM as best-effort:

- `apis/rust/node/src/node/mod.rs:602-619` — `ShmProviderBuilder::default_backend()` failure logs a warn and stores `None`
- `apis/rust/node/src/node/mod.rs:1064-1088` — `send_output_sample` checks `if let Some(provider)` and falls back to heap buffer if allocation fails

So a misconfigured `/dev/shm` is a perf concern (heap copy on every large message instead of zero-copy), not a correctness one. `dora doctor` should not exit non-zero on a system that actually runs.

## Why this matters

When `/dev/shm` has the wrong permissions (e.g. `0755` on some hardened systems, or `0700` after a chown), the zero-copy fast path is silently disabled and large messages get copied through the heap on every publish. Users don't notice until they wonder why throughput is lower than expected. Surfacing it as a `WARN` in `dora doctor` lets them spot the misconfiguration without digging into daemon stderr.

## Why not a separate `dora system doctor` command (as PR #1352 proposed)

PR #1352 (@Harsh-Sahu43) added this check as a new `dora system doctor` command under a new `system` namespace. That namespace rename was declined as part of the umbrella decision on [#1196](https://github.com/dora-rs/dora/issues/1196) ("the `dora system *` rename was not adopted; existing top-level commands stay"). The check itself is good engineering; only the command-naming was off.

This PR lifts the core check logic into the existing `dora doctor` and credits @Harsh-Sahu43 via `Co-Authored-By`. See the close comment on [#1352](https://github.com/dora-rs/dora/pull/1352) for the redirect context.

## Test plan

- [x] `cargo check -p dora-cli` — clean on host (macOS)
- [x] `cargo test -p dora-cli` — clean (no regressions; check is `#[cfg(target_os = "linux")]` so it does not run in dora-cli's host-target test suite on macOS)
- [ ] **Authoritative**: a Linux user running `dora doctor` should see one of the four outcomes above depending on `/dev/shm` state. Reviewer with a Linux box can validate manually:
  - `dora doctor` — expect `PASS Shared memory: /dev/shm mode is 1777` on a default Ubuntu/Debian/Fedora install
  - `sudo chmod 0755 /dev/shm && dora doctor` — expect the `WARN ... mode 0o755 ...` case
  - `sudo chmod 1777 /dev/shm` to restore

## What this PR does NOT do

- Doesn't add other `dora doctor` checks (CPU isolation, swap, ulimits, etc.). Those can ship as follow-up PRs.
- Doesn't introduce a `dora system` namespace. The check goes in the existing `dora doctor`.
- Doesn't add behavioral changes to dora's SHM code — purely diagnostic.

Co-Authored-By: Harsh-Sahu43

### DIFF
--- a/binaries/cli/src/command/doctor.rs
+++ b/binaries/cli/src/command/doctor.rs
@@ -233,6 +233,37 @@ fn warn(stdout: &mut termcolor::StandardStream, msg: &str) -> eyre::Result<()> {
     Ok(())
 }
 
+/// Pure evaluation of a directory's suitability as a POSIX shared-memory mount.
+///
+/// Returns one of four states based on `path`'s existence, type, and mode.
+/// Separated from the I/O wrapper so it can be unit-tested against a tempdir.
+#[cfg(target_os = "linux")]
+#[derive(Debug)]
+enum ShmState {
+    Inaccessible(std::io::Error),
+    NotADirectory,
+    WrongMode { actual: u32, expected: u32 },
+    Ok,
+}
+
+#[cfg(target_os = "linux")]
+fn evaluate_shm(path: &std::path::Path) -> ShmState {
+    use std::os::unix::fs::PermissionsExt;
+    let metadata = match std::fs::metadata(path) {
+        Ok(metadata) => metadata,
+        Err(e) => return ShmState::Inaccessible(e),
+    };
+    if !metadata.is_dir() {
+        return ShmState::NotADirectory;
+    }
+    let actual = metadata.permissions().mode() & 0o7777;
+    let expected = 0o1777;
+    if actual != expected {
+        return ShmState::WrongMode { actual, expected };
+    }
+    ShmState::Ok
+}
+
 /// Check `/dev/shm` is a directory with mode `1777` (sticky, world-rwx).
 ///
 /// dora's zero-copy fast path for large payloads uses zenoh SHM, which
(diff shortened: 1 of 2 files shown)


heyong4725:
Addressed the review finding in 82c6ecf5.

You're right — the node runtime treats SHM as best-effort:

- `apis/rust/node/src/node/mod.rs:602-619` — `ShmProviderBuilder::default_backend()` failure logs a warn and stores `None`
- `apis/rust/node/src/node/mod.rs:1064-1088` — `send_output_sample` checks `if let Some(provider)` and falls back to heap buffer if allocation fails

So a misconfigured `/dev/shm` is a perf concern, not a correctness one. Downgraded all three failure cases (not accessible / not a directory / wrong mode) to `warn()`, dropped the `failures` out-param, and reworded the messages to make it clear dora will still run via heap fallback. Updated the function doc comment to cite the runtime fallback path.

Local validation after change:
- `cargo clippy -p dora-cli -- -D warnings` clean
- `cargo test -p dora-cli` pass

The check will no longer flip the exit code on a system whose `/dev/shm` is misconfigured — `dora doctor` only exits non-zero now when something is actually broken.

PR #1828, opened 2026-05-15
fix(build): keep draining build output streams

### LINKED ISSUE #1821: fix(build): build output silently truncated when one of stdout/stderr closes first
## Summary

`forward_build_output` (and similar inline `tokio::select!` patterns in `libraries/core/src/build/build_command.rs`) can silently drop build output when one of stdout/stderr closes before the other.

## Reproduction

The current pattern in `libraries/core/src/build/build_command.rs`:

```rust
loop {
    let line = tokio::select! {
        line = stdout_lines.next_line() => line,
        line = stderr_lines.next_line() => line,
    };
    let Some(line) = line.transpose() else {
        break;     // ← stops as soon as EITHER stream returns None
    };
    if stdout_tx.send(line).await.is_err() {
        break;
    }
}
```

When stderr closes first, `select!` returns `Ok(None)` from `stderr_lines.next_line()`, the `transpose()` yields `None`, and the loop breaks — **even if stdout still has buffered lines to drain**. Build output gets silently truncated.

Symptoms in practice:

- A build script that flushes stderr early ("compiling…\n") then writes a long stdout tail (build artifacts list) loses the stdout tail.
- A build script that errors after producing useful stdout context can lose the stdout context if stderr closes first when the process exits.

This is most visible with `dora build --uv` / managed Python envs (#1820) because `uv venv` and `uv pip install` write progress to one stream and final output to another, but the bug existed long before that PR.

## Proposed fix

Use `tokio_stream::StreamExt::merge` instead of `tokio::select!`:

```rust
use tokio_stream::{StreamExt, wrappers::LinesStream};

let mut merged = LinesStream::new(stdout.lines()).merge(LinesStream::new(stderr.lines()));
while let Some(line) = merged.next().await {
    if stdout_tx.send(line).await.is_err() {
        break;
    }
}
```

`merge` keeps polling the remaining stream after one closes, so no output is dropped.

## Scope

Three callers in `libraries/core/src/build/build_command.rs`:

1. `run_build_command` — the build-command inline loop (lines ~70-85 today)
2. `prepare_managed_python_env` (in #1820) — calls `forward_build_output`
3. `ensure_managed_python_runtime` (in #1820) — calls `forward_build_output`

#1820 unified callers 2 and 3 into `forward_build_output`, but kept the inline loop in #1 because that was the pre-existing pattern. A focused PR for this issue should:

- Convert all three callers to use the merged-stream approach
- Add `tokio-stream = { workspace = true }` to `libraries/core/Cargo.toml` (the workspace already declares it at the root)
- Add a regression test along the lines of "writer closes stderr early, writes 256 lines to stdout, verifies all 256 lines arrive at the receiver"

## Context

This was flagged during pre-landing review of #1820 ([rescue of #1515](https://github.com/dora-rs/dora/pull/1820)). The #1515 author had already implemented this fix in their PR but I deliberately did not port it into the rescue:

- The bug is tangential to the rescue's purpose (per-node Python venv management)
- It requires adding a new dependency to `dora-core`
- Mixing tangential fixes into a feature PR makes review harder

Filing here so the improvement gets reviewed on its own merits.

## Acceptance criteria

- [ ] All three callers in `build_command.rs` use the merged-stream approach
- [ ] Regression test added (the #1515 author's `keeps_draining_stdout_after_stderr_closes` and `keeps_draining_stderr_after_stdout_closes` tests are a good starting point)
- [ ] `cargo test -p dora-core --features build` passes
- [ ] No regression in existing build output forwarding

## References

- #1515 (original PR, closed) — contained the `forward_build_output` implementation using `LinesStream::merge`
- #1820 (rescue of #1515) — landed managed envs but kept the inline `select!` pattern; this issue tracks the follow-up

phil-opp:
Fixes #1821.

## Summary
- Replace EOF-sensitive stdout/stderr select loops with merged stream forwarding.
- Reuse the helper across build command and managed Python env output forwarding.
- Add regression coverage for both stdout-first and stderr-first close orders.

## Validation
- cargo test -p dora-core --features build keeps_draining
- cargo test -p dora-core --features build
- cargo fmt --all -- --check
- cargo clippy -p dora-core --features build -- -D warnings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1807,6 +1807,7 @@ dependencies = [
  "syn 2.0.117",
  "tempfile",
  "tokio",
+ "tokio-stream",
  "tracing",
  "url",
  "uuid",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -132,6 +132,7 @@ thiserror = "2.0"
 colored = "3.0.0"
 inquire = { version = "0.7.5", default-features = false }
 tokio = { version = "1.28", default-features = false }
+tokio-stream = "0.1.11"
 eyre = "0.6.8"
 serde = { version = "1", features = ["derive"] }
 tracing = "0.1.36"
--- a/libraries/core/Cargo.toml
+++ b/libraries/core/Cargo.toml
@@ -27,6 +27,7 @@ uuid = { workspace = true }
 tracing = { workspace = true }
 serde-with-expand-env = "1.1.0"
 tokio = { workspace = true, features = ["fs", "io-util", "macros", "process", "sync", "rt"] }
+tokio-stream = { workspace = true, features = ["io-util"] }
 schemars = "1.0.4"
 serde_json = { workspace = true }
 log = { version = "0.4.21", features = ["serde"] }
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -9,9 +9,10 @@ use crate::build::{managed_python_bin_dir, managed_python_interpreter};
 use dora_message::descriptor::EnvValue;
 use eyre::{Context, eyre};
 use tokio::{
-    io::{AsyncBufReadExt, BufReader},
+    io::{AsyncBufRead, AsyncBufReadExt, BufReader},
     process::Command,
 };
+use tokio_stream::{StreamExt, wrappers::LinesStream};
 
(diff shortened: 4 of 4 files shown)


phil-opp:
@claude review this PR

PR #1835, opened 2026-05-15
feat(cli): `dora clean` removes finished/failed dataflows (rescue of #1366)

heyong4725:
Adds `dora clean` to remove finished and failed dataflows from the coordinator's state without restarting it.

## Why

`dora list` accumulates entries forever — every finished or failed dataflow stays in the coordinator's `dataflow_results` (and its archived descriptor) until the coordinator is destroyed. After iterative dev work the list becomes unreadable, and long-running coordinators (CI, production with the redb-persisted state) keep growing the state file.

The existing workaround is `dora down + dora up`, which is heavy-handed: it disconnects daemons, drops everything, restarts from scratch. `dora clean` does only the focused thing: drop finished/failed history, keep running dataflows running, keep daemons connected.

## Example

```
$ dora list
UUID                                  NAME              STATUS
abc...                               vision-dev         Running
xyz...                               vision-dev         Finished      ← 1 hour ago
def...                               test-pipeline      Failed        ← yesterday
... (50 more historical entries)

$ dora clean
UUID                                  NAME              STATUS
xyz...                               vision-dev         Finished
def...                               test-pipeline      Failed
... etc

Cleaned 52 dataflow(s).

$ dora list
UUID                                  NAME              STATUS
abc...                               vision-dev         Running       ← only running ones left
```

Standard shape — same as `docker system prune`, `git gc`, `kubectl delete pods --field-selector=status.phase==Succeeded`.

## What's removed vs. preserved

| Touched? | What |
|---|---|
| **Removed** | `dataflow_results` entries for finished/failed dataflows |
| **Removed** | Their `archived_dataflows` descriptors |
| **Preserved (intentionally)** | `finished_builds` cache — clearing it would break concurrent `dora build` calls with "unknown build id" errors. From phil-opp's review of #1366. |
| **Preserved** | Running dataflows |
| **Preserved** | Daemon connections |

## What's lost after cleaning

`dora logs <uuid>` no longer works for cleaned dataflows (the archived descriptor that backs log lookup is gone). Save anything you'd want to reference later before running `dora clean`. The CHANGELOG and CLI doc comment both call this out.

## Pieces

| File | Change |
|---|---|
| `libraries/message/src/cli_to_coordinator.rs` | new `ControlRequest::Clean` variant; reply reuses `DataflowList` |
| `binaries/coordinator/src/lib.rs` | new match arm; drains `dataflow_results` via `std::mem::take` in a single pass; uses `IndexMap::shift_remove` for the archived descriptor cleanup |
| `binaries/cli/src/command/clean.rs` (new) | `dora clean [--format table\|json] [--quiet]`; modeled on `dora list` but simpler (no metrics, no filters) |
| `binaries/cli/src/command/mod.rs` | wire into `Command` enum + dispatch + parse-test |
| `docs/cli.md`, `guide/src/operations/cli.md` | reference docs |
| `Changelog.md` | Unreleased > Added entry |

## Why this is a rewrite, not a code-lift of #1366

The original PR (drindr, Feb 2026) targeted `binaries/coordinator/src/server.rs` and the `CliControl` tarpc trait. Both were deleted when the cli↔coordinator control plane was rewritten to use `ControlRequest` enum dispatch over WebSocket. The PR's `self.state.dataflow_results` pattern doesn't map to current main, where those are local variables inside a per-event async task in `lib.rs:268-271`.

So this PR uses #1366 as a **design reference** (output shape, shared display structure with `dora list`, doc tone) but the implementation is fresh against the post-rewrite coordinator. drindr credited via `Co-Authored-By`; original PR closed with redirect.

## Maintainer corrections from #1366's review baked in from the start

phil-opp's two requested changes from #1366 are present here directly, not as a follow-up:

1. **Don't clear `finished_builds`** — keeps concurrent `dora build` calls working.
2. **Doc comment says "Remove finished and failed dataflows from the coordinator"** — explicit about scope.
3. **Doc note about logs no longer being available** after clean (in both the doc comment and the user-facing docs).

Bonus correctness improvement over the original: the snapshot is atomic by construction (single async task, `std::mem::take` drains in one pass) — no risk of the iterate-then-remove race that the original DashMap-based code had.

## Test plan

- [x] `cargo clippy -p dora-cli -p dora-coordinator -p dora-message -- -D warnings` — clean
- [x] `cargo test -p dora-cli --lib parse_clean` — passes (covers `--format`, `--quiet`, and the `--quiet`/`--format` conflict)
- [x] `cargo run -p dora-cli -- clean --help` — produces correct help text
- [ ] **Reviewer-runnable E2E** (needs a coordinator):
  ```bash
  # Start coordinator + run + finish a dataflow
  dora up
  dora run examples/python-dataflow/dataflow.yml --stop-after 5s
  dora list                          # shows 1 finished entry
  dora clean                         # cleans it
  dora list                          # empty
  dora down
  ```

## Closes

Closes #1366.

Co-Authored-By: drindr


heyong4725:
Both findings are real and serious. Fix in `c2dfa9e4`. Postmortem: my own pre-landing review missed both because I only audited the empty case and the full-finish case, not the partial-finish or persistence cases. Adding a "always audit other non-happy paths in the same code region" line to my pattern bank.

## Finding 1 (HIGH) — fix

You're right about the partial-results semantics. Traced through:

1. Multi-daemon dataflow `X` on daemons {A, B}. Both running.
2. Daemon A finishes successfully. Coordinator (`lib.rs:469-472`): inserts `(A, Ok)` into `dataflow_results[X]`. `running_dataflows[X].daemons` is now `{B}` — still non-empty, so the cleanup block at `lib.rs:474` is skipped. X stays in `running_dataflows`.
3. User runs `dora clean`. My old code drained ALL of `dataflow_results` via `std::mem::take`, including `X`'s partial entry.
4. Daemon B finishes with a node failure. Coordinator re-inserts `dataflow_results[X][B] = Err`. The new entry has ONLY B's result; A's success is gone.
5. `dataflow.daemons.is_empty()` is now true. Cleanup block runs (`lib.rs:511`-`538`). At `lib.rs:519-535`, status is computed from just B's result. If B failed, status is `Failed { error: <B's error> }`. **But the error log loses A's contribution entirely.** Worse: if B happens to be in the right shape, line 535 can default to `Succeeded`.

The fix uses `retain` and explicitly skips UUIDs still in `running_dataflows`:

```rust
dataflow_results.retain(|uuid, results| {
    if running_dataflows.contains_key(uuid) {
        // Multi-daemon dataflow still completing — keep partial
        // results so the final status is correct.
        return true;
    }
    // ... drain entry, push to cleaned[], delete persisted ...
    false
});
```

Partial results are preserved; only fully-completed dataflows are pulled out.

## Finding 2 (MEDIUM) — fix

You're also right that I underdelivered on the stated motivation. The CHANGELOG claimed `dora clean` helps trim redb state; the implementation didn't touch redb. Confirmed the gap by tracing:

- `lib.rs:538` writes each finished dataflow's `DataflowRecord` to `store.put_dataflow(&r)`.
- `lib.rs:295-323` on coordinator restart reads from the store. Line 320: `StoreDataflowStatus::Succeeded | StoreDataflowStatus::Failed { .. } => {}` — empty match arm. So persisted finished/failed records are **left in the store but never reloaded into the in-memory `dataflow_results`**.
- `state.rs:244` calls `store.get_dataflow(dataflow_id)` directly when resolving param targets, bypassing the in-memory map entirely.

Net effect before this fix: `dora clean` made `dora list` look clean, but:
- The redb file kept growing
- `dora param` against a "cleaned" UUID would still succeed via the persisted record path
- Restarting the coordinator would not reload the cleaned records into the list (they're correctly absent there) — but they were still sitting in the redb file

Fix: in the same drain loop, call `store.delete_dataflow(uuid)` after the in-memory removal. Errors are logged (not propagated) so a transient store error doesn't abort the rest of the clean.

```rust
if let Err(e) = store.delete_dataflow(uuid) {
    tracing::warn!(
        "failed to delete persisted record for cleaned \
         dataflow {uuid}: {e}"
    );
}
```

## Documentation updated

Reworded the user-facing surfaces to reflect actual behavior:

- `libraries/message/src/cli_to_coordinator.rs` — `ControlRequest::Clean` doc comment
- `docs/cli.md` + `guide/src/operations/cli.md` — "What's removed" / "What's NOT removed" / "What's lost" sections now mention the running-skip, the redb cleanup, and the `dora param` impact on cleaned UUIDs
- `Changelog.md` — entry rewritten to be accurate

## Tests

- `cargo test --test ws-cli-e2e cli_clean_empty` still passes (empty path unchanged)
- `cargo test -p dora-cli --lib parse_clean` still passes
- `cargo clippy -p dora-cli -p dora-coordinator -p dora-message -- -D warnings` clean

The new behavior (skip running, delete persisted) is **not yet exercised by an E2E** — that would need a daemon plus a multi-daemon partial-finish simulation, which is significant infrastructure. The fix is established by code review of the two-line behavioral change; the existing E2E catches protocol-level regressions, and the multi-daemon corruption scenario is now structurally impossible because the predicate filters by `running_dataflows.contains_key()`. If you want me to add the multi-daemon E2E I can do it in a follow-up — flag if so.

## Pattern bank update for me

Adding to my checklist: "after fixing a finding, audit OTHER non-happy paths in the same code region." I caught the stdout/stderr issue last round by auditing the empty case, but missed both of these because I didn't audit the partial-finish (Finding 1) or persistence (Finding 2) cases — even though both were sitting right there in adjacent match arms / store calls in the same lib.rs file.

heyong4725:
Third finding caught on this re-review pass. Fix pushed in `1ba42b08`.

## Finding 3 (MEDIUM) — `delete_dataflow` does not cascade NODE_PARAMS

Tracing through `libraries/coordinator-store/src/redb_store.rs:224-233`, `delete_dataflow` only touches the `DATAFLOWS` table:

```rust
fn delete_dataflow(&self, uuid: &Uuid) -> Result<()> {
    let key: &[u8] = uuid.as_bytes();
    let txn = self.db.begin_write()?;
    {
        let mut table = txn.open_table(DATAFLOWS)?;
        table.remove(key)?;
    }
    txn.commit()?;
    Ok(())
}
```

But `put_node_param` (same file, line 283) writes into a separate table `NODE_PARAMS` with a composite key `<uuid>\0<node_id>\0<key>` (see `make_param_key` at line 365). After last round's fix, `dora clean` calls `delete_dataflow`, which only drops the DATAFLOWS row -- the NODE_PARAMS rows for that uuid survive forever.

Two consequences that contradict this PR's own motivation:

1. The redb file keeps growing on long-lived coordinators even after `dora clean`. The CHANGELOG explicitly says clean "trims redb state" -- with the orphan param rows, that's only half-true.
2. `store.list_node_params(cleaned_uuid, node)` keeps returning data for a dataflow that no longer exists in DATAFLOWS. Any future `dora param`-shape code that resolves param targets by uuid alone would surface "ghost" rows.

The in-memory backend at `libraries/coordinator-store/src/in_memory.rs:118-125` has the same gap on the `params: RwLock<BTreeMap<ParamKey, Vec<u8>>>` field.

### Fix

`delete_dataflow` now cascades to NODE_PARAMS in both backends.

* **redb backend** (atomic): `Table::retain_in(prefix.., |k, _| !k.starts_with(prefix))` runs in the same write transaction as the DATAFLOWS removal.

  ```rust
  let param_prefix = format!("{uuid}{KEY_SEPARATOR}");
  let txn = self.db.begin_write()?;
  {
      let mut table = txn.open_table(DATAFLOWS)?;
      table.remove(key)?;
  }
  {
      let mut params = txn.open_table(NODE_PARAMS)?;
      params.retain_in(param_prefix.as_str().., |k, _| {
          !k.starts_with(param_prefix.as_str())
      })?;
  }
  txn.commit()?;
  ```

* **in-memory backend** (one lock per table, matching the rest of the impl): after dropping the dataflows row, `retain` on the params BTreeMap.

  ```rust
  params.retain(|(df, _, _), _| df != uuid);
  ```

  The coordinator dispatches events serially, so the brief window between the two removals is invisible to callers.

### Tests

Added `delete_dataflow_cascades_to_node_params` to both `redb_store::tests` and `in_memory::tests`. Put params under two dataflows, delete one, assert that dataflow's params are gone and the other dataflow's params survive. Both pass.

```
test redb_store::tests::delete_dataflow_cascades_to_node_params ... ok
test in_memory::tests::delete_dataflow_cascades_to_node_params ... ok
```

Full `cargo test -p dora-coordinator-store --all-features --lib` passes (32 tests). Clippy `-D warnings` clean across `dora-coordinator-store`, `dora-coordinator`, `dora-message`. fmt clean.

### Docs

* `libraries/coordinator-store/src/lib.rs` -- trait `fn delete_dataflow` now carries a doc comment specifying the cascade contract and the per-backend atomicity model.
* `libraries/message/src/cli_to_coordinator.rs` -- `ControlRequest::Clean` doc now mentions the cascade.
* `docs/cli.md` + `guide/src/operations/cli.md` -- "What's removed" sections updated.
* `Changelog.md` -- clean entry now calls out the param-row cascade.

### Pattern bank update

This is the third round of the same lesson. Before declaring a review pass complete, I now audit:

1. The empty / "nothing to do" case (round 1, stdout/stderr)
2. The partial-progress case (round 2, multi-daemon)
3. The persistence side-effects case (round 2, redb)
4. The cascade-on-delete case (round 3, this commit)

The unifying lesson: **when a handler removes state, audit every other table / map / file / cache that the removed state might reference.** I should have caught this at round 2 -- I added `store.delete_dataflow` last round but didn't check whether `delete_dataflow` actually deletes everything the cleaned dataflow owns. Lesson logged. Next iteration of `/review` flow will include "for every removal, list all tables/maps the removed entity has references in."

heyong4725:
Real bug, exactly as you traced it. Fix in `2010d3cc`.

The previous round moved `store.delete_dataflow` into the right region but kept it in the **wrong order** relative to the in-memory mutations: the retain closure already pushed to `cleaned` and was going to return `false` (drop from `dataflow_results`) by the time the store error surfaced. So a redb write failure meant the CLI said "cleaned" while the persisted row + cascade-deleted param rows survived, and the retry handle was already gone.

## Fix

Restructured the retain closure to make persistence the atomic commit point of the per-entry operation. Persist first; only on success do we touch the in-memory state and push to the reply.

```rust
dataflow_results.retain(|uuid, results| {
    if running_dataflows.contains_key(uuid) {
        // Multi-daemon dataflow still completing — keep
        // partial results so the final status is correct.
        return true;
    }
    // Persist the removal FIRST. If the store write fails,
    // leave the in-memory state untouched so (a) we don't lie
    // to the CLI by reporting this dataflow as cleaned while
    // its redb row + param rows survive, and (b) the next
    // `dora clean` can retry — dropping the `dataflow_results`
    // entry here would lose the retry handle.
    if let Err(e) = store.delete_dataflow(uuid) {
        tracing::warn!(
            "skipping clean for dataflow {uuid}: persisted-store \
             delete failed: {e}. In-memory entry preserved so a \
             later `dora clean` can retry."
        );
        return true;
    }
    let name = archived_dataflows.shift_remove(uuid).and_then(|d| d.name);
    let status = if results.values().all(|r| r.is_ok()) {
        DataflowStatus::Finished
    } else {
        DataflowStatus::Failed
    };
    cleaned.push(DataflowListEntry {
        id: DataflowIdAndName { uuid: *uuid, name },
        status,
    });
    false
});
```

Trace verification:

| Case | archived_dataflows | dataflow_results | cleaned reply | retry possible? |
|---|---|---|---|---|
| store ok | dropped | dropped | yes | n/a |
| store err | untouched | kept | not listed | yes, on next `dora clean` |
| dataflow still running | untouched | kept | not listed | n/a |

The reply now truthfully reflects what was persisted: a dataflow only appears in the success list when both sides actually got cleaned.

## Docs

Updated to match the new ordering and retry semantics:

* `libraries/message/src/cli_to_coordinator.rs` — `ControlRequest::Clean` doc spells out the persist-first ordering and the retry-on-store-error path.
* `docs/cli.md` and `guide/src/operations/cli.md` — "What's removed" sections say the persisted-store delete runs first and that store failures keep the in-memory entry for retry.

## Validation

```
cargo fmt --check                                                 clean
cargo clippy -p dora-coordinator -p dora-message -- -D warnings   clean
cargo test -p dora-coordinator-store --all-features --lib         32/32 pass
cargo test -p dora-cli --lib parse_clean                          1/1 pass
cargo test --test ws-cli-e2e cli_clean_empty                      1/1 pass
```

The store-error path is not exercised by an automated test in this commit — that would need a `CoordinatorStore` impl that errors on `delete_dataflow` plus a way to drive the coordinator event loop with it, which is significant test infrastructure for a single-branch invariant. The closure's structure makes the contract obvious: every path that returns `true` from `retain` skips both the `archived_dataflows.shift_remove` and the `cleaned.push`, preserving the retry handle. Happy to add the test infra in a follow-up if you'd prefer.

## Pattern bank update

Fifth lesson for this PR: **when the same code path mutates in-memory state AND issues a persistent write, the persistent write goes first, and the in-memory mutation is committed only on success.** Otherwise the response and the in-memory state can both lie about persisted reality, and the failure becomes unrecoverable because the retry handle is already gone.

I should have spotted this during round 3, when I added the `store.delete_dataflow` call — the lie-on-failure surface is literally the closure's return value sitting two lines below the `tracing::warn!`. The lesson going forward: after wiring a fallible side-effect into an existing flow, re-read the surrounding control flow asking "what does each path claim, and is that claim still true if my side-effect failed?"

heyong4725:
Real bug exactly as traced. Fix in `939e04ef`.

You're right that round 4 left the reply shape unable to distinguish "no eligible candidates" from "all candidates failed" — both come back as `Ok(DataflowList([]))`, and the CLI's `entries.is_empty()` branch in `clean.rs:69` prints "No finished or failed dataflows to clean." in both cases. The docs change I wrote in the same round explicitly promised failed candidates would be "skipped from the success list," but the CLI had nothing on which to act.

## Protocol change

Added a new reply variant with separate success/failure lists, and a struct to carry the failure detail:

```rust
ControlRequestReply::CleanResult {
    cleaned: DataflowList,
    failed: Vec<CleanFailure>,
}

pub struct CleanFailure {
    pub id: DataflowIdAndName,
    pub error: String,
}
```

Safe because the variant is added in the same PR that introduces `ControlRequest::Clean`, so there are no in-flight CLI callers expecting `DataflowList` for a `Clean` request. `dora list` continues to use `DataflowList`. The Hello handshake ensures CLI and coordinator ship together.

## Coordinator

`binaries/coordinator/src/lib.rs` now collects `cleaned` and `failed` separately. On store error, the closure looks up the name via `archived_dataflows.get(uuid)` (read-only — the archived entry stays intact so the preserved in-memory retry handle still has its metadata), pushes a `CleanFailure { id, error: e.to_string() }` into `failed`, and returns `true` from `retain`. On success the existing flow runs. Reply is `CleanResult { cleaned, failed }`.

## CLI

`binaries/cli/src/command/clean.rs` consumes the new shape via the existing struct-variant `expect_reply!` form:

```rust
let (cleaned, failed) = expect_reply!(reply, CleanResult { cleaned, failed })?;
```

Output matrix:

| `cleaned` | `failed` | stdout | stderr | exit |
| --- | --- | --- | --- | --- |
| empty | empty | (none) | `No finished or failed dataflows to clean.` | 0 |
| empty | non-empty | (none) | `0 cleaned; N failed (see warnings below).` + N `warning:` lines | non-zero |
| non-empty | empty | table/json | `Cleaned N dataflow(s).` | 0 |
| non-empty | non-empty | table/json | `Cleaned N dataflow(s); M failed (see warnings below).` + M `warning:` lines | non-zero |

Warnings (and JSON failure records in `--format json` mode) go to stderr so stdout stays parseable. On any failure the function returns `eyre::eyre!(...)` carrying the failure count + a retry hint, so the CLI's main loop surfaces it as a non-zero exit. Scripted callers can now tell a partial outage apart from a "nothing eligible" run.

## Tests

```
cargo fmt --check                                                 clean
cargo clippy -p dora-cli -p dora-coordinator -p dora-message -- -D warnings   clean
cargo test -p dora-coordinator-store --all-features --lib         32/32 pass
cargo test -p dora-cli --lib parse_clean                          1/1 pass
cargo test --test ws-cli-e2e cli_clean_empty                      1/1 pass
```

`cli_clean_empty` was updated to match the new reply shape and now asserts both `cleaned` and `failed` are empty in the no-eligible case. The all-failed and partial-failure cases remain code-review-only — exercising them needs a `CoordinatorStore` impl that errors on `delete_dataflow` plus the test infra to drive the coordinator event loop with it. The closure's control flow makes the contract clear: every path that returns `true` from retain populates `failed` and skips `cleaned`; every path that returns `false` populates `cleaned` and never appears in `failed`. Happy to add the test infra in a follow-up if you'd like.

## Docs

* `libraries/message/src/cli_to_coordinator.rs` — `ControlRequest::Clean` doc points at `ControlRequestReply::CleanResult` and describes the success/failure split.
* `libraries/message/src/coordinator_to_cli.rs` — doc on the new variant + `CleanFailure` struct.
* `docs/cli.md` — new "What happens on a persisted-store failure" paragraph describing the stderr warning + non-zero exit.
* `guide/src/operations/cli.md` — matching "Partial-failure behavior" paragraph.
* `Changelog.md` — rewritten to mention the partial-failure surfacing.

## Pattern bank update

Sixth lesson for this PR (and probably the unifying one for the whole sequence): **when a handler can fail per-entry, the reply shape must distinguish "no candidates" from "all candidates failed."** A single `cleaned`-only list collapses both observables into one, and the "what do I tell the user" code on the other side of the wire then degrades to silently misreporting partial outages. Fixing this at the protocol level (separate `cleaned` and `failed` lists) is cheaper than per-call-site CLI guesswork, and the contract is now structural rather than depending on a comment.

I should have spotted this when I added the docs claim "those dataflows simply do not appear in the success list" in round 4 — the moment a reply field's *absence* carries semantic weight is the moment the reply needs a second field to disambiguate it.

heyong4725:
Real bug, exactly as you traced it. Fix in `0308dda2`.

The recovery loop at `binaries/coordinator/src/lib.rs:295-323` explicitly leaves `StoreDataflowStatus::Succeeded | Failed { .. }` records on disk without reloading them into `dataflow_results`, and the previous clean only walked `dataflow_results`. So historical rows from before a coordinator restart were unreachable — the redb file just grew, despite the "trim state-file growth" promise in `docs/cli.md` and `CHANGELOG.md`.

## Fix

Took your suggested shape verbatim: split the handler into two phases.

**Phase A — enumerate candidates from both sources** into `Vec<(Uuid, Option<String>, DataflowStatus, in_memory: bool)>`:

```rust
// in-memory pass
for (uuid, results) in dataflow_results.iter() {
    if running_dataflows.contains_key(uuid) { continue; }
    let name = archived_dataflows.get(uuid).and_then(|d| d.name.clone());
    let status = if results.values().all(|r| r.is_ok()) {
        DataflowStatus::Finished
    } else {
        DataflowStatus::Failed
    };
    candidates.push((*uuid, name, status, true));
}

// persisted pass
match store.list_dataflows() {
    Ok(records) => {
        for record in records {
            if running_dataflows.contains_key(&record.uuid) { continue; }
            if dataflow_results.contains_key(&record.uuid) { continue; }
            let status = match record.status {
                StoreDataflowStatus::Succeeded => DataflowStatus::Finished,
                StoreDataflowStatus::Failed { .. } => DataflowStatus::Failed,
                _ => continue,
            };
            candidates.push((record.uuid, record.name, status, false));
        }
    }
    Err(e) => {
        tracing::warn!("failed to enumerate persisted dataflows during clean: {e}. \
                        Only in-memory candidates will be considered for this call.");
    }
}
```

A `list_dataflows` error degrades gracefully to in-memory-only rather than failing the whole call.

**Phase B — persist-first, then mutate**:

```rust
for (uuid, name, status, in_memory) in candidates {
    let id = DataflowIdAndName { uuid, name };
    if let Err(e) = store.delete_dataflow(&uuid) {
        tracing::warn!(
            "skipping clean for dataflow {uuid}: persisted-store delete failed: {e}. \
             {state} preserved so a later `dora clean` can retry.",
            state = if in_memory { "In-memory entry" } else { "Persisted record" }
        );
        failed.push(CleanFailure { id, error: e.to_string() });
        continue;
    }
    if in_memory {
        dataflow_results.remove(&uuid);
    }
    archived_dataflows.shift_remove(&uuid);
    cleaned.push(DataflowListEntry { id, status });
}
```

This replaces `dataflow_results.retain(...)` exactly as you suggested — the collect-then-mutate pattern handles the two-source merge cleanly and makes the success/failure branches obvious. Persist-first ordering carries over from round 4; the cascade-delete to `NODE_PARAMS` from round 3 still applies, so cleaned post-restart rows actually shrink the redb file.

## Regression test

Added `cli_clean_reaps_persisted_only_record` in `tests/ws-cli-e2e.rs` using the existing `seed_dataflow_record` + `start_coordinator_background_with_store` helpers:

```rust
let store_impl = Arc::new(InMemoryStore::new());
let dataflow_id = uuid::Uuid::new_v4();
seed_dataflow_record(store_impl.as_ref(), dataflow_id, &["sensor"]);  // writes Succeeded
let store: Arc<dyn CoordinatorStore> = store_impl.clone();
let port = start_coordinator_background_with_store(store);
// connect + send Clean
// assert cleaned.0[0].id.uuid == dataflow_id
// assert failed.is_empty()
// assert store_impl.get_dataflow(&dataflow_id).is_none()
```

This is exactly the shape you sketched: persisted-only record, empty `dataflow_results`, send `Clean`, verify both `CleanResult.cleaned` contains it and `store.get_dataflow` returns `None` afterward.

## Validation

```
cargo fmt --check                                                clean
cargo clippy -p dora-cli -p dora-coordinator -p dora-message --tests -- -D warnings   clean
cargo test -p dora-coordinator-store --all-features --lib        32/32 pass
cargo test -p dora-cli --lib parse_clean                         1/1 pass
cargo test --test ws-cli-e2e cli_clean                           2/2 pass  (empty + persisted-only)
```

## Docs

* `libraries/message/src/cli_to_coordinator.rs` — `ControlRequest::Clean` doc now spells out the two-source candidate enumeration and the post-restart use case.
* `docs/cli.md` + `guide/src/operations/cli.md` — "What's removed" sections call out the persisted-only reap path and the recovery-loop rationale.
* `Changelog.md` — rewritten to mention the post-restart reap.

## Pattern bank update

Seventh lesson for this PR (and the most embarrassing one in hindsight). **When a feature promises to bound the growth of persistent state, the implementation has to cover every state-accumulation path — including states that the startup / recovery loop deliberately leaves intact.** I had been auditing the within-session happy path on every round without ever walking the restart axis, even though the whole point of redb is to survive process boundaries. Any "trim disk state" feature has to consider what's on disk *after* a restart, not just what's currently in memory.

Going forward: when reviewing a persistent-state-touching handler, I'll explicitly enumerate every code path that *writes* to the persisted backing store, then verify the handler's enumeration sees state from each writer — not just the writers active in the current process.

heyong4725:
Real bug, exactly as traced. Fix in `0b10a979`.

Round 6 added the persisted-store pass but degraded silently on `list_dataflows()` error, which lets the worst-case combination (enumeration error + empty in-memory) flow through as `Ok(CleanResult { cleaned: [], failed: [] })`. The CLI then prints "No finished or failed dataflows to clean." — exactly the opposite of what's true. Operator's mental model is "I ran clean, so state is bounded"; silent degradation in the corrupt-redb path is precisely where that model needs to *not* break.

## Fix — hard-fail with a clear error

Took your first option (request error) over the new-field option — simpler contract, no protocol surface change, and "we can't see all the state, so we mutate none of it" is the right default for a state-bounding feature.

```rust
let records = match store.list_dataflows() {
    Ok(records) => records,
    Err(e) => {
        let _ = reply_sender.send(Err(eyre!(
            "dora clean: failed to enumerate persisted dataflows: {e}. \
             No state was modified; the next `dora clean` will retry once \
             the coordinator's redb store is healthy again."
        )));
        continue;
    }
};
```

The `continue` lands on the outer `while let Some(event)` event loop, skipping the rest of the Clean arm and the latency telemetry line. No state is mutated; the in-memory candidates the call would have processed stay in `dataflow_results`, so the next `dora clean` (after the operator fixes the redb file) reaps them. The error message names the underlying cause AND reassures that state is intact, so the operator doesn't panic about partial cleanup.

## Regression test

New `cli_clean_hard_fails_when_persisted_enumeration_fails` in `tests/ws-cli-e2e.rs`, backed by a `FailingListStore` wrapper:

```rust
struct FailingListStore {
    inner: InMemoryStore,
    fail_list: AtomicBool,
}

impl CoordinatorStore for FailingListStore {
    // ... all methods delegate to self.inner ...
    fn list_dataflows(&self) -> eyre::Result<Vec<DataflowRecord>> {
        if self.fail_list.load(Ordering::Acquire) {
            Err(eyre::eyre!("injected list_dataflows failure for test"))
        } else {
            self.inner.list_dataflows()
        }
    }
}
```

The toggle is flipped AFTER coordinator startup (which itself calls `list_dataflows` once at `lib.rs:295`), then the test sends `ControlRequest::Clean` and asserts:

1. Reply is `ControlRequestReply::Error(msg)` (NOT `CleanResult`)
2. `msg` contains "failed to enumerate persisted dataflows"
3. `msg` contains "No state was modified" (operator-facing reassurance)

The wrapper is ~80 lines of mechanical trait delegation but it's the cleanest way to inject a targeted store failure without inventing new test infra.

## Validation

```
cargo fmt --check                                                clean
cargo clippy -p dora-cli -p dora-coordinator -p dora-message --tests -- -D warnings   clean
cargo test -p dora-coordinator-store --all-features --lib        32/32 pass
cargo test --test ws-cli-e2e cli_clean                           3/3 pass
   cli_clean_empty
   cli_clean_reaps_persisted_only_record
   cli_clean_hard_fails_when_persisted_enumeration_fails
```

## Docs

* `libraries/message/src/cli_to_coordinator.rs` — `ControlRequest::Clean` doc now states explicitly that enumeration failure is a request error and that no state is mutated on that path.
* `docs/cli.md` + `guide/src/operations/cli.md` — added a paragraph in the partial-failure section about the hard-fail-on-enumeration behavior.
* `Changelog.md` — clean entry now mentions the enumeration-error hard-fail.

## Pattern bank update

Eighth lesson for this PR. **When a handler enumerates persistent state to enforce a contract, enumeration failure has to fail the contract**, not silently degrade. "Best-effort with a log warning" is the wrong default for state-bounding features because the operator's mental model is "I ran clean, so state is bounded" — silent degradation breaks that model in exactly the failure modes (corrupt redb, I/O error) where they most need to know.

Going forward, my reviewer-question-of-record for any state-promising handler is: **"what happens if we can't see the state we promise to operate on?"** If the answer is "we report success on the subset we could see," that's a silent-correctness bug.

heyong4725:
Round 8 — self-review of `0b10a979` after addressing your enumeration-error hard-fail. Two cosmetic findings, no new safety bugs. Polished both in `0c163254`.

## What I audited

Applying the lessons accumulated over rounds 1-7, I ran each one against the latest handler:

- **#1 (audit other non-happy paths in same code region):** Phase A (in-memory pass) early-exits on `running_dataflows.contains_key`. Phase A.5 (persisted pass) early-exits on `running_dataflows`/`dataflow_results.contains_key` and non-completed `record.status`. Phase B (per-candidate processing) early-exits on `store.delete_dataflow` failure. Each early-exit preserves invariants correctly — no state mutated on the failure paths.
- **#2 (cascade on delete):** still holds — `delete_dataflow` cascades to `NODE_PARAMS` in both backends (round 3).
- **#3 (persist-first then in-memory):** still holds — Phase B does `store.delete_dataflow` before any `dataflow_results.remove` / `archived_dataflows.shift_remove`.
- **#4 (reply shape distinguishes observable states):** still holds — `CleanResult { cleaned, failed }`, and now plus the request-level `Error` for enumeration failure.
- **#5 (walk the restart axis):** still holds — Phase A.5 reaps persisted-only rows.
- **#6 (persistence-bounding handler full coverage):** still holds.
- **#7 (enumeration failure must fail the contract):** held in this round.
- **#8 (what happens if we can't see the state we promise to operate on?):** answered — hard-fail with reassuring error.

## Concurrency audit (new for this round)

Background tokio task `replay_persisted_params_for_daemon` (`lib.rs:2897`) holds `Arc<dyn CoordinatorStore>` and reads via `collect_param_replay_items` while the event loop also touches the store. **Not a race**: the replay task is spawned only for dataflows that are in `running_dataflows`, and the Clean handler's two-source enumeration explicitly skips uuids in `running_dataflows` (Phase A check at the in-memory loop, Phase A.5 check on persisted records). So in-flight replay for X ⇒ X is running ⇒ Clean skips X. ✓

## Cosmetic findings (fixed in `0c163254`)

**A — overlapping comments in the Clean handler preamble:** Lines 1003-1030 had pre-round-6 prose ("We also remove each cleaned entry from the persisted store... Drains dataflow_results...") that overlapped the post-refactor `// Phase A: enumerate...` header right below. Reorganized into one focused bullet list covering the four contracts (eligibility, persisted cleanup + NODE_PARAMS cascade, finished_builds preservation, two-source enumeration).

**B — backend-specific wording in the enumeration-failure error:** Said "the coordinator's redb store is healthy again", but `InMemoryStore::list_dataflows` can also return `Err` (RwLock poisoning, see `in_memory.rs`). Reworded to "the coordinator's store" so the message stays accurate regardless of backend. The existing `cli_clean_hard_fails_when_persisted_enumeration_fails` test asserts on the unchanged "failed to enumerate" and "No state was modified" substrings, so no test change needed.

## Validation

```
cargo fmt --check                                                clean
cargo clippy -p dora-coordinator -- -D warnings                  clean
cargo test --test ws-cli-e2e cli_clean                           3/3 pass
   cli_clean_empty
   cli_clean_reaps_persisted_only_record
   cli_clean_hard_fails_when_persisted_enumeration_fails
```

## Status

No new structural findings. PR #1835 head is now `0c163254`. The cumulative review history is 7 rounds of substantive fixes + 1 round of cosmetic polish. Ready for the next reviewer pass or merge once CI is green.

PR #1846, opened 2026-05-17
feat(c++/node): add event receive variants (rescue of #1409)

heyong4725:
## Summary

Rescue of [#1409](https://github.com/dora-rs/dora/pull/1409) (@PavelGuzenfeld) with phil-opp's [review fixes](https://github.com/dora-rs/dora/pull/1409#pullrequestreview-2761090773) applied — the author replied "Fixed" on 2026-03-18 but never actually pushed those fixes, so this rescue does them.

This is the **first PR in a 4-PR rescue stack** for PavelGuzenfeld's C++ API parity work. The full series:

- **#TBD (this PR)** — event receive variants (rescues #1409)
- TBD — close_outputs + NodeFailed/Reload event types (rescues #1410)
- TBD — node_config_json + dataflow_descriptor_json introspection (rescues #1413)
- TBD — operator InputClosed/Stop callback forwarding (rescues #1414, behavior bug fix)

Closes #1409.

## New FFI surface

| Function | Behavior |
|---|---|
| `try_next_event(events)` | Non-blocking poll. Returns `DoraEventType::Empty` if no event is ready, `AllInputsClosed` if the stream is closed. |
| `next_event_timeout(events, timeout_ms)` | Timed-blocking receive. Returns `DoraEventType::Timeout` on deadline elapse, `AllInputsClosed` if the stream closes first. |
| `events_is_empty(events)` | Queue-empty hint (treat as advisory — can race against new events). |
| `drain_events(events) -> DrainedEvents` | Snapshot of currently buffered events; `next_event` only sees later events. |
| `drained_events_len(drained)` | Snapshot size. |
| `drained_events_next(drained)` | Pop next from snapshot. Returns `Empty` after exhaustion. |

Two new `DoraEventType` variants:
- `Empty` — non-blocking poll found nothing / drain exhausted
- `Timeout` — `next_event_timeout` deadline elapsed

## Design fixes vs. the original PR

@phil-opp's review on #1409 raised three issues that the author acknowledged in replies but never pushed. The rescue applies them:

### 1. No more string-matching of error messages

Original `next_event_timeout`:

```rust
let timed_out = matches!(&event, Event::Error(msg) if msg.contains("timed out"));
```

Brittle — any change to the timeout error message silently breaks detection.

Rescue:

```rust
let deadline = Instant::now() + Duration::from_millis(timeout_ms);
let poll_interval = Duration::from_millis(1);
loop {
    match events.0.try_recv() {
        Ok(event) => return EventOrReason::Event(event),
        Err(TryRecvError::Closed) => return EventOrReason::Closed,
        Err(TryRecvError::Empty) => {
            let now = Instant::now();
            if now >= deadline { return EventOrReason::TimedOut; }
            std::thread::sleep(std::cmp::min(deadline - now, poll_interval));
        }
    }
}
```

Timeout is detected structurally via `Instant`; `Empty` / `Closed` from `try_recv` disambiguate timeout from stream close cleanly. No string parsing.

### 2. `try_next_event` returns `Empty`, not `Timeout`

Original code returned `DoraEventType::Timeout` when the non-blocking poll found nothing. But the caller never set a timeout — calling it a "timeout" overloaded the meaning. The rescue introduces `EventOrReason::Empty` and `DoraEventType::Empty` so the no-event-ready case has its own surface.

### 3. `drained_events_next` returns `Empty` when exhausted

Same fix — a drained snapshot is frozen at `drain_events()` time and can't time out. `Empty` is the honest answer.

## Internal representation

`DoraEvent` switches from `(Option<Event>)` to `(EventOrReason)` so the no-event-available state can carry its reason (Closed / Empty / TimedOut):

```rust
pub(crate) enum EventOrReason {
    Event(Event),
    Closed,    // -> DoraEventType::AllInputsClosed
    Empty,     // -> DoraEventType::Empty
    TimedOut,  // -> DoraEventType::Timeout
}
```

All existing call sites (`event_type`, `event_as_input`, `event_as_arrow_input`, `event_as_arrow_input_with_info`, `downcast_dora`) are updated to match. The external `DoraEventType` surface is purely additive — existing variants unchanged, two new variants added.

## What was dropped from the original PR

The original PR #1409 bundled several other changes that have been outdated by main's evolution since 2026-03-16:

| Original PR included | Disposition | Why |
|---|---|---|
| `init_dora_node_from_id`, `init_dora_node_flexible` | **Dropped** | Built on top of #1403 (`node_id`/`dataflow_id` accessors) which was wiped by the 1.0 consolidation `145ccce0`. Separate PR if/when the node-id surface comes back. |
| `node_id()`, `dataflow_id()` accessors | **Dropped** | Same — 1.0 consolidation removed these. |
| Zero-copy `DataSampleHandle` API | **Dropped** | Already shipped in main via #1431 with a different shape. Re-adding would duplicate. |
| `log_message` removal | **Not applied** | Current main uses `log_message` for C/C++ logging (added after #1409 was opened). Kept. |

So this rescue is a narrow extraction of just the event-receive variants (the named feature in the PR title) + phil-opp's review fixes applied.

## Test plan

- [x] `cargo build -p dora-node-api-cxx` — clean compile, no warnings
- [x] `cargo clippy -p dora-node-api-cxx -- -D warnings` — clean
- [x] Manual review of all 5 `event.0` / `DoraEvent(...)` call sites updated to match the new `EventOrReason` shape
- [ ] CircleCI cli-rust and examples-linux jobs will exercise the new functions via the example demo (added to `examples/c++-dataflow/node-rust-api/main.cc`)

The example demo on the non-blocking poll case:

```cpp
auto poll = try_next_event(dora_node.events);
if (event_type(poll) == DoraEventType::Empty) {
    std::cout << "No event ready yet (non-blocking poll)" << std::endl;
}
```

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -238,7 +238,15 @@ fn next_event(events: &mut Box<Events>) -> Box<DoraEvent> {
 /// matching on a wrapped `Event::Error(msg.contains("timed out"))`
 /// would be fragile (#1409 review feedback).
 fn next_event_timeout(events: &mut Box<Events>, timeout_ms: u64) -> Box<DoraEvent> {
-    let deadline = Instant::now() + Duration::from_millis(timeout_ms);
+    // `timeout_ms` arrives across the cxx::bridge as `u64`, so a C++
+    // caller can supply values that would panic the underlying
+    // `Instant + Duration` arithmetic on platforms with a bounded
+    // Instant range. Use `checked_add` and treat overflow as
+    // "effectively no deadline" -- keep polling indefinitely until
+    // an event arrives or the stream closes. Never returns
+    // `TimedOut` in the overflow case, matching the caller's
+    // intent ("wait as long as needed").
+    let deadline = Instant::now().checked_add(Duration::from_millis(timeout_ms));
     // 1 ms keeps the busy-wait cost negligible while bounding overshoot
     // of the caller-supplied deadline to ~1 ms in the worst case.
     let poll_interval = Duration::from_millis(1);
@@ -246,13 +254,16 @@ fn next_event_timeout(events: &mut Box<Events>, timeout_ms: u64) -> Box<DoraEven
         match events.0.try_recv() {
             Ok(event) => return Box::new(DoraEvent(EventOrReason::Event(event))),
             Err(TryRecvError::Closed) => return Box::new(DoraEvent(EventOrReason::Closed)),
-            Err(TryRecvError::Empty) => {
-                let now = Instant::now();
-                if now >= deadline {
-                    return Box::new(DoraEvent(EventOrReason::TimedOut));
+            Err(TryRecvError::Empty) => match deadline {
+                Some(d) => {
+                    let now = Instant::now();
+                    if now >= d {
+                        return Box::new(DoraEvent(EventOrReason::TimedOut));
+                    }
+                    std::thread::sleep(std::cmp::min(d - now, poll_interval));
                 }
-                std::thread::sleep(std::cmp::min(deadline - now, poll_interval));
-            }
+                None => std::thread::sleep(poll_interval),
+            },
         }
     }
 }
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -1,11 +1,16 @@
-use std::{any::Any, collections::BTreeMap, vec};
+use std::{
+    any::Any,
+    collections::{BTreeMap, VecDeque},
+    time::{Duration, Instant},
+    vec,
+};
 
 use crate::ffi::MetadataValueType;
 
 use chrono::DateTime;
 use dora_node_api::{
     self, Event, EventStream, Metadata as DoraMetadata,
-    MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter,
+    MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter, TryRecvError,
     arrow::array::{AsArray, UInt8Array},
     merged::{MergeExternal, MergedEvent},
 };
@@ -36,6 +41,14 @@ mod ffi {
         Error,
         Unknown,
         AllInputsClosed,
+        /// Non-blocking poll (`try_next_event`) found no event ready,
+        /// and the same value is returned by `drained_events_next` once
+        /// a drained queue is exhausted. Distinct from `Timeout` —
+        /// `Empty` is returned even when no timeout was set.
+        Empty,
+        /// `next_event_timeout` returned without an event because the
+        /// caller-supplied deadline elapsed first.
+        Timeout,
     }
 
     struct DoraInput {
@@ -76,6 +89,7 @@ mod ffi {
         type Events;
         type OutputSender;
         type DoraEvent;
+        type DrainedEvents;
         type MergedEvents;
         type MergedDoraEvent;
         type Metadata;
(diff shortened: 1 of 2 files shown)


heyong4725:
## Self-review (adversarial pass)

Ran the standard `/review` shape against this PR. One small finding, fixed in `b8839237`. Posting the full review on the record:

### Wiring audit — clean

All five `event.0` / `DoraEvent` constructor sites updated consistently for the `DoraEvent(EventOrReason)` shape change:

| Site | Old | New |
|---|---|---|
| `Events::next` (line 223) | `DoraEvent(self.0.recv())` | `DoraEvent(EventOrReason::Event(...))` or `Closed` |
| `event_type` (line 345) | `match &event.0` on `Option<Event>` | `match &event.0` on `EventOrReason` |
| `event_as_input` (line 360) | `Some(Event::Input { ... })` | `EventOrReason::Event(Event::Input { ... })` |
| `event_as_arrow_input` (line 396) | same | same |
| `event_as_arrow_input_with_info` (line 684) | same | same |
| `downcast_dora` (line 896) | bail on None | now returns `EventOrReason::Closed` for None (small intentional improvement — see below) |

External `DoraEventType` enum is purely additive (new `Empty`, `Timeout` variants). No ABI break for existing C++ callers.

### Edge case audit

**`next_event_timeout`:**
- `timeout_ms = 0`: deadline = now → first `try_recv` Empty path hits `now >= deadline`, returns `TimedOut`. Equivalent to `try_next_event` semantically (with `Timeout` instead of `Empty`). ✓
- Spurious sleep wake-up: re-tries `try_recv`. ✓
- Race window between `try_recv` returning Empty and the deadline check (event arrives in between): the event stays in the queue for the next caller; we still return `TimedOut`. Acceptable per timeout semantics.
- Latent overflow: `Instant::now() + Duration::from_millis(u64::MAX)` could panic on platforms with limited Instant range. Realistic risk: zero (no caller passes u64::MAX). Noted but not fixing.

**Behavior change on `downcast_dora`:** original code bailed with "not an external event" when the merged stream returned `None`. The rescue treats `None` as `EventOrReason::Closed` (which maps to `DoraEventType::AllInputsClosed` on the C++ side). This is more useful to C++ callers — they can handle stream close uniformly via `event_type()` instead of needing a try/catch around the FFI. The old error message was also misleading (`downcast_dora` checks for *dora* events; `None` means stream closed, not "not external"). Worth flagging as a small behavior improvement but not a regression — existing callers using try/catch still see `Ok(DoraEvent)` and can detect close via `event_type()`.

### Finding (LOW) — fixed in `b8839237`

The example demo could silently drop an in-flight event. `examples/c++-dataflow/dataflow.yml` declares `tick: dora/timer/millis/300`, and on a slow machine the daemon could deliver a tick between `init_dora_node()` and the demo's `try_next_event()`. The demo would pop that event without processing it, shortening the main receive loop's count by one.

Fix: gate the demo on `events_is_empty()` first so the subsequent `try_next_event` only runs when the queue is provably empty. Bonus — this also exercises the new `events_is_empty` FFI on the example path.

```cpp
if (events_is_empty(dora_node.events)) {
    auto poll = try_next_event(dora_node.events);
    if (event_type(poll) == DoraEventType::Empty) {
        std::cout << "No event ready yet (non-blocking poll)" << std::endl;
    }
}
```

### Verification

- `cargo build -p dora-node-api-cxx` — clean
- `cargo clippy -p dora-node-api-cxx -- -D warnings` — clean
- `cargo fmt --check` — clean

### Status

Clean. PR head now `b8839237`. Ready for merge once CI confirms.

heyong4725:
Both findings real. Fix in `7ca9a963`.

## Finding 1 — overflow panic on `next_event_timeout`

You're right — I noted this in my self-review and dismissed it as "zero realistic risk," but that's the wrong frame for a public C++ API. The `cxx::bridge` accepts `u64` and panics propagating across the FFI boundary are worse than any reasonable alternative.

Fix uses `checked_add` + falls through to "no deadline" polling on overflow:

```rust
let deadline = Instant::now().checked_add(Duration::from_millis(timeout_ms));
// ...
Err(TryRecvError::Empty) => match deadline {
    Some(d) => {
        let now = Instant::now();
        if now >= d {
            return Box::new(DoraEvent(EventOrReason::TimedOut));
        }
        std::thread::sleep(std::cmp::min(d - now, poll_interval));
    }
    None => std::thread::sleep(poll_interval),
},
```

If overflow happens (which would only be for absurd inputs like `u64::MAX`), the loop never returns `TimedOut` and just keeps polling — matches the caller's intent ("wait as long as needed"). No panic.

## Finding 2 — demo race not actually closed

You're correct that `events_is_empty` is documented as a racy hint and the previous fix only shrank the window. The cleaner answer is to remove the live demo entirely.

The functions (`try_next_event`, `events_is_empty`, `next_event_timeout`, `drain_events`, etc.) remain documented via their `cxx::bridge` declarations at `apis/c++/node/src/lib.rs:103-123`. The right place to exercise them in CI is a dedicated test, not the shared dataflow node that downstream consumers also depend on. An example that races is worse than no example here.

Removed both the `events_is_empty` check and the `try_next_event` call from `main.cc`. The main receive loop is back to its pre-rescue shape; the only difference vs. main is the new FFI surface (which doesn't change behavior for callers that don't use it).

## Validation

```
cargo build -p dora-node-api-cxx                       clean
cargo clippy -p dora-node-api-cxx -- -D warnings        clean
cargo fmt --all -- --check                              clean
```

On your `cargo run --example cxx-dataflow` failure: agreed, that's pre-existing infra (system `~/.cargo/bin/dora` getting picked up over the workspace one, and runtime communication being misparsed). Not a PR finding — the rescued FFI compiles and the existing main-loop path is unchanged by the demo removal.

## Pattern bank update for me

Two lessons:

1. **"Realistic risk" is the wrong frame for FFI panic surfaces.** Public APIs accepting unconstrained types from another language need to be panic-safe for the entire input domain, not just the realistic subset. Should have used `checked_add` from the start.

2. **"Racy hint" docs are a smell when used to *prevent* a race.** I read `events_is_empty`'s own doc ("treat it as a hint, not a guarantee") and then used it as a guard against the race in Finding 1 (my round-2 fix). That's exactly what the doc warned against. Removing the racy demo is the right answer; gating it on a racy primitive was reasoning that contradicted the API contract I had just documented.

PR head: `7ca9a963`. Ready for re-review.

PR #1847, opened 2026-05-17
feat(c++/node): add close_outputs, NodeFailed and Reload event types (rescue of #1410)

heyong4725:
## Summary

Rescue of [#1410](https://github.com/dora-rs/dora/pull/1410) (@PavelGuzenfeld). Second PR in the 4-PR rescue stack for C++ API parity.

**Stacked on top of [#1846](https://github.com/dora-rs/dora/pull/1846)** (event receive variants). Once #1846 merges I'll rebase this onto main.

Closes #1410.

## New FFI surface

| Item | Behavior |
|---|---|
| `DoraEventType::NodeFailed` | Upstream node has failed; downstream inputs that depended on it will stop receiving data. |
| `DoraEventType::Reload` | Hot-reload notification for an operator. C++ nodes can now react (e.g. flush caches) instead of treating it as `Unknown`. |
| `DoraNodeFailed { affected_input_ids, error, source_node_id }` | Payload struct surfaced via `event_as_node_failed`. |
| `event_as_node_failed(event)` | Extracts the failure payload from a `NodeFailed` event. |
| `close_outputs(sender, output_ids)` | Selectively close one or more of this node's outputs without shutting the whole node down. Downstream sees `InputClosed`. |

## Implementation notes

- `event_as_node_failed` destructures via `EventOrReason::Event(Event::NodeFailed { … })` to match the new tuple-struct shape that #1846's rescue introduced (the original PR used `event.event` which assumed the never-adopted named-field shape).
- `close_outputs` calls through to `dora_node_api::DoraNode::close_outputs(Vec<DataId>)` which has been on main since the 1.0 consolidation. Errors are returned as `DoraResult { error }` for ergonomic consistency with `send_output`.
- `event_type` gets two new match arms (`Event::NodeFailed { .. }` → `NodeFailed`, `Event::Reload { .. }` → `Reload`). Existing variants unchanged.

## Test plan

- [x] `cargo build -p dora-node-api-cxx` — clean compile
- [x] `cargo clippy -p dora-node-api-cxx -- -D warnings` — clean
- [x] Example (`examples/c++-dataflow/node-rust-api/main.cc`) extended to handle `NodeFailed` and `Reload` event types
- [ ] CircleCI cli-rust + examples-linux will exercise via the example's main loop

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -446,8 +446,24 @@ fn close_outputs(
     output_sender: &mut Box<OutputSender>,
     output_ids: Vec<String>,
 ) -> ffi::DoraResult {
-    let ids: Vec<dora_node_api::dora_core::config::DataId> =
-        output_ids.into_iter().map(Into::into).collect();
+    // Parse via `FromStr` instead of `From<String>` so invalid IDs
+    // surface as a `DoraResult.error` rather than panicking across
+    // the cxx::bridge. `DataId::from(String)` is documented as
+    // panicking on invalid characters (see
+    // `libraries/message/src/id.rs:186`, `# Panics`); calling it on
+    // caller-supplied input would mean a typo in a C++ string literal
+    // aborts the whole node.
+    let mut ids = Vec::with_capacity(output_ids.len());
+    for id in output_ids {
+        match id.parse::<dora_node_api::dora_core::config::DataId>() {
+            Ok(parsed) => ids.push(parsed),
+            Err(e) => {
+                return ffi::DoraResult {
+                    error: format!("invalid output id '{id}': {e}"),
+                };
+            }
+        }
+    }
     match output_sender.0.close_outputs(ids) {
         Ok(()) => ffi::DoraResult {
             error: String::new(),
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -9,12 +9,13 @@ use crate::ffi::MetadataValueType;
 
 use chrono::DateTime;
 use dora_node_api::{
-    self, Event, EventStream, Metadata as DoraMetadata,
-    MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter, TryRecvError,
+    self,
     arrow::array::{AsArray, UInt8Array},
     merged::{MergeExternal, MergedEvent},
+    Event, EventStream, Metadata as DoraMetadata, MetadataParameters as DoraMetadataParameters,
+    Parameter as DoraParameter, TryRecvError,
 };
-use eyre::{Result as EyreResult, bail, eyre};
+use eyre::{bail, eyre, Result as EyreResult};
 use serde::Serialize;
 use serde_json::Value as JsonValue;
 
@@ -24,7 +25,7 @@ pub use prelude::*;
 pub mod prelude {
     pub use dora_ros2_bridge::prelude::*;
 }
-use futures_lite::{Stream, StreamExt, stream};
+use futures_lite::{stream, Stream, StreamExt};
 
 #[cxx::bridge]
 #[allow(clippy::needless_lifetimes)]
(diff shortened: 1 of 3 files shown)


heyong4725:
Real bug, same panic-across-FFI category as #1846's `Instant` overflow. Fix in `41d9b43a`.

## The bug

`From<String> for DataId` is documented as panicking on invalid characters (`libraries/message/src/id.rs:186`, with explicit `# Panics` notice). Calling `output_ids.into_iter().map(Into::into).collect()` on caller-supplied C++ input means a typo in a C++ string literal aborts the whole node instead of returning the `DoraResult.error` like every other function in the surface.

## The fix

Replace `Into::into` with `parse::<DataId>()` (the safe `FromStr` path at `libraries/message/src/id.rs:167`); first invalid id short-circuits with a structured error:

```rust
let mut ids = Vec::with_capacity(output_ids.len());
for id in output_ids {
    match id.parse::<dora_node_api::dora_core::config::DataId>() {
        Ok(parsed) => ids.push(parsed),
        Err(e) => {
            return ffi::DoraResult {
                error: format!("invalid output id '{id}': {e}"),
            };
        }
    }
}
match output_sender.0.close_outputs(ids) {
    Ok(()) => ffi::DoraResult { error: String::new() },
    Err(err) => ffi::DoraResult { error: format!("{err:?}") },
}
```

Well-formed-id case is unchanged; only the previously-panicking error path now surfaces a `DoraResult` with a human-readable message naming the offending id.

## Pattern bank update (second time in this PR series)

The unifying rule: **when an FFI accepts type `T` and there's a `From<X> for T` that panics on invalid input, never use that — use `FromStr` / `TryFrom` instead.** dora's `DataId` / `NodeId` deliberately panic in `From` to keep trusted Rust call sites concise — that's the right call inside the workspace and the wrong call at FFI boundaries.

Lessons from #1846 + #1847 combined:
1. (round-2 #1846) `Instant::now() + Duration::from_millis(u64)` can panic on overflow — use `checked_add`.
2. (this PR) `DataId::from(String)` panics on invalid id — use `parse::<DataId>()`.

Going forward, when reviewing FFI surfaces I check: "what's the unconstrained input type, what total-feeling Rust conversions does it touch, and are any of them partial-with-panic?"

## On the residual docs note

You're right that `guide/src/languages/cpp.md` (642 lines) documents the pre-existing surface but not the new `Empty`/`Timeout`/`NodeFailed`/`Reload` event types, `try_next_event`/`next_event_timeout`/`events_is_empty`/`drain_events*`/`close_outputs`/`event_as_node_failed`, or (later in this stack) `node_config_json`/`dataflow_descriptor_json`/`on_input_closed`/`on_stop`. The cxx::bridge inline doc comments do propagate to generated C++ headers as Doxygen-style annotations, so the API isn't strictly undocumented, but the long-form guide is stale.

That's a doc gap spanning 4 PRs of new C++ API surface, and it'd more than double the size of any individual PR in the stack. Suggestion: I'll file a follow-up issue tracking the doc work for the whole #1846–#1849 surface and link it from each PR description. That way the docs land as one coherent update against the post-merge state instead of being smeared across 4 stacked rebases. Sound right, or want me to fold the docs into this stack now?

PR head: `5b39f72f` → `41d9b43a`. Ready for re-review.

heyong4725:
Done in `532a17a8`. Updated `docs/api-cxx.md`:

- **New `Non-blocking and timed receive` subsection** under `Events` — documents `try_next_event`, `next_event_timeout`, `events_is_empty`, and the `drain_events` / `DrainedEvents` family. These are #1846's surface but they need to be in the doc for the enum below to make sense.
- **DoraEvent downcast section** — added `event_as_node_failed`.
- **DoraEventType enum** — expanded from 6 to 10 variants (matching what's actually on main now) with one-line descriptions, plus a paragraph explaining the `Empty` vs `Timeout` semantic distinction so readers don't conflate them.
- **New `DoraNodeFailed` struct section** after `DoraInput`, with a short usage example.
- **New `OutputSender::close_outputs` subsection** in the `OutputSender` part, calling out that ids are validated and return `DoraResult.error` on invalid input (matching the `FromStr` parse fix from the previous review round).

Scope note: I included the #1846 catch-up (`Empty`/`Timeout`/`try_next_event` family) here because the enum/downcast tables can't be accurate without them. Strictly speaking that was #1846's miss, but bundling here keeps the post-merge doc state consistent with main rather than leaving an intermediate broken state for whoever picks up the next stack PR.

PR head: `41d9b43a` → `532a17a8`. The 4-PR series docs are now partially caught up (this PR covers #1846 + #1847's surface); #1848 and #1849 should each add their own surface to the same file as they land.

PR #1848, opened 2026-05-17
feat(c++/node): node_config_json and dataflow_descriptor_json (rescue of #1413)

heyong4725:
## Summary

Rescue of [#1413](https://github.com/dora-rs/dora/pull/1413) (@PavelGuzenfeld). Third PR in the 4-PR rescue stack for C++ API parity.

**Stacked on top of [#1847](https://github.com/dora-rs/dora/pull/1847)** (close_outputs + NodeFailed/Reload). Will rebase onto main once the lower stack lands.

Closes #1413.

## New FFI surface

| Function | Behavior |
|---|---|
| `node_config_json(output_sender) -> Result<String>` | Returns this node's `NodeRunConfig` (inputs/outputs/metadata) serialized as JSON. Lets C++ nodes introspect their own declared interface at runtime without re-parsing the yaml. |
| `dataflow_descriptor_json(output_sender) -> Result<String>` | Returns the full parsed `Descriptor` (the dataflow yaml) serialized as JSON. Useful for introspecting peer nodes, listing all topics, etc. May fail if the daemon hasn't delivered the descriptor yet. |

Both wrap existing methods on `dora_node_api::DoraNode` (`apis/rust/node/src/node/mod.rs:1116` and `:1317`) via `serde_json::to_string`. No new dora-rs API surface; this is purely exposing existing Rust APIs to C++.

## Test plan

- [x] `cargo build -p dora-node-api-cxx` — clean
- [x] `cargo clippy -p dora-node-api-cxx -- -D warnings` — clean
- [x] Example demonstrates both calls with a `try/catch` around `rust::Error` (introspection can fail if the daemon hasn't yet delivered the descriptor)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1849, opened 2026-05-17
feat(c++/operator): forward InputClosed and Stop events to C++ callbacks (rescue of #1414)

heyong4725:
## Summary

Rescue of [#1414](https://github.com/dora-rs/dora/pull/1414) (@PavelGuzenfeld). Fourth and final PR in the 4-PR rescue stack for C++ API parity.

**Stacked on top of [#1848](https://github.com/dora-rs/dora/pull/1848)** (node_config_json / dataflow_descriptor_json). Will rebase onto main once the lower stack lands.

This is the **behavior bug fix** in the stack. Previously the C++ operator API silently dropped `Event::InputClosed { id }` and `Event::Stop` events via the catch-all `_ => Continue` arm in `apis/c++/operator/src/lib.rs:107-110`. C++ operators had no way to know an upstream input stream had closed (so a node graph shutting down looked indistinguishable from an idle pause) or that a graceful stop was in progress (so cleanup hooks couldn't fire).

Closes #1414.

## New FFI surface

| Function | Behavior |
|---|---|
| `on_input_closed(op, id) -> DoraOnInputResult` | Called when the daemon delivers `Event::InputClosed { id }`. Operator can log, flush per-input state, or set its own stop flag. |
| `on_stop(op) -> DoraOnInputResult` | Called when the daemon delivers `Event::Stop`. Operator can drain output queues, persist final state, etc. |

Both return `DoraOnInputResult { error, stop }` (same shape as `on_input`), so operators can request early termination from either callback by setting `stop = true`.

## Behavior change

| Event | Before | After |
|---|---|---|
| `Event::Input { ... }` | `ffi::on_input(...)` | unchanged |
| `Event::InputClosed { id }` | **silently dropped** (`_ => Continue`) | `ffi::on_input_closed(operator, id)` |
| `Event::Stop` | **silently dropped** (`_ => Continue`) | `ffi::on_stop(operator)` |
| Other (NodeFailed, Reload, Error, …) | `Continue` | `Continue` (catch-all retained — these have no operator-side callback) |

## Implementation note

`apis/rust/operator::Event::Stop` is a unit variant (no payload), unlike `dora_node_api::Event::Stop(StopCause)`. The original PR pattern-matched on the unit form; this rescue preserves that — the cherry-pick initially attempted `Event::Stop(_)` which failed to compile (`expected tuple struct or tuple variant, found unit variant Event::Stop`), corrected during the rescue.

## Example operators

Both example operators get default `on_input_closed` and `on_stop` implementations:

- `examples/c++-dataflow/operator-rust-api/operator.{h,cc}`
- `examples/cmake-dataflow/operator-rust-api/operator.{h,cc}`

Default behavior: print which input closed / that stop was received, plus the current counter value. Validates the new callbacks fire end-to-end when the examples run in CI.

## Test plan

- [x] `cargo build -p dora-operator-api-cxx` — clean
- [x] `cargo clippy -p dora-operator-api-cxx -- -D warnings` — clean
- [ ] CircleCI cli-rust + examples-linux will run the updated example operators


heyong4725:
## Self-review

Standard adversarial pass on PR #1849. Three findings, all fixed in `e3a59d5c`.

### Wiring audit — clean

- Two new FFI methods declared in `unsafe extern "C++"`: `on_input_closed(Pin<&mut Operator>, &str) -> DoraOnInputResult`, `on_stop(Pin<&mut Operator>) -> DoraOnInputResult`. Same shape as the existing `on_input` callback. ✓
- `DoraOperator::on_event` adds explicit arms for `Event::InputClosed { id }` → `ffi::on_input_closed` and `Event::Stop` → `ffi::on_stop`. Catch-all `_ => Continue` retained for remaining variants. ✓
- Both new arms convert `DoraOnInputResult { error, stop }` to `DoraStatus::Continue`/`Stop` identically to the existing `on_input` path — uniform error propagation. ✓
- The `Event::Stop` match is **unit-variant** form (correct for `dora_operator_api::Event::Stop`), not `Event::Stop(_)` (which is `dora_node_api::Event::Stop(StopCause)`). I bumped into this distinction during the rescue earlier; the pattern match is right.

### Finding 1 (MEDIUM, the substantive one): breaking change for external C++ operators not flagged in user-visible docs

The two new FFI methods are declared in `unsafe extern "C++"`, meaning the C++ side MUST implement them. Every existing downstream C++ operator that builds against `dora-operator-api-cxx` fails to link without `on_input_closed` and `on_stop` stubs. The repo's own examples are updated in this PR (`examples/c++-dataflow/operator-rust-api/` + `examples/cmake-dataflow/operator-rust-api/`), but external users had no notice.

**Fix:**

- New `### Breaking` section at the top of `Changelog.md` naming the change, the symptom (link failure), and the minimal restore-prior-behavior stub:
  ```cpp
  DoraOnInputResult on_input_closed(Operator& op, rust::Str id) {
      return { rust::String(), false };
  }
  DoraOnInputResult on_stop(Operator& op) {
      return { rust::String(), false };
  }
  ```
- `docs/api-cxx.md` operator section header reworded from "two functions" → "four functions", a `> Breaking change` callout added at the top of the section, and the `Required C++ interface` example now shows both new prototypes. Two new bullets describe `on_input_closed` and `on_stop`.

### Finding 2 (LOW): rustdoc wording bug

`on_stop`'s cxx::bridge doc comment said `Event::Stop(_)` (tuple syntax) — wrong, the operator-API variant is unit. The pattern match itself was correct (`Event::Stop =>`), only the doc text was stale. Reworded to:

> Called on graceful shutdown (the daemon delivers `Event::Stop` — a unit variant on `dora_operator_api::Event`, distinct from `dora_node_api::Event::Stop(StopCause)` which carries a payload).

The "distinct from" clarification is load-bearing because the two `Event` enums diverge on this one variant; I tripped on it during the rescue and want the next reader to see the answer in the rustdoc.

### Finding 3 (informational, fixed as part of Finding 1): operator section in `docs/api-cxx.md` was outdated

The section header said the C++ side implements "two functions" — now four. Fixed alongside Finding 1.

### Validation

```
cargo build -p dora-operator-api-cxx                    clean
cargo clippy -p dora-operator-api-cxx -- -D warnings    clean
cargo fmt --all -- --check                              clean
markdown fence parity                                   78 (even)
```

PR head: `cdc89705` → `e3a59d5c`. Ready for re-review or merge.

heyong4725:
Both surfaces real and missed in my first round. Fix in `df231ef3`.

## 1. `dora new` C++ operator template (`binaries/cli/src/template/cxx/operator-template.{h,cc}`)

Confirmed the symptom: any user running `dora new` to start a fresh C++ operator project gets `operator-template.h` baked into the binary, which only declared `new_operator` + `on_input` — won't link against current `dora-operator-api-cxx`. Added stubs to both files matching the in-repo example operators that this PR already updates:

```cpp
// operator-template.h
DoraOnInputResult on_input_closed(Operator &op, rust::Str id);
DoraOnInputResult on_stop(Operator &op);

// operator-template.cc
DoraOnInputResult on_input_closed(Operator &op, rust::Str id)
{
    std::cout << "Rust API operator: input `" << std::string(id) << "` closed" << std::endl;
    DoraOnInputResult result = {rust::String(), false};
    return result;
}

DoraOnInputResult on_stop(Operator &op)
{
    std::cout << "Rust API operator: stop received (counter was " << (unsigned int)op.counter << ")" << std::endl;
    DoraOnInputResult result = {rust::String(), false};
    return result;
}
```

The template is `include_str!`'d into the CLI binary, so `cargo build -p dora-cli` validates the file content compiles as the template intends.

## 2. Long-form C++ guide (`guide/src/languages/cpp.md:320`)

You're right that the guide carried the same stale "two functions" snippet. Mirrored the same edits I made to `docs/api-cxx.md` in the previous round:

- Header sentence: "two functions" → "four functions"
- `> Breaking change` callout at the top of the section
- Required-interface example shows `on_input_closed` + `on_stop` prototypes
- Two new bullets describing each callback

Worth noting: the two doc surfaces (`docs/api-cxx.md` and `guide/src/languages/cpp.md`) carry nearly identical operator-section content and have been drift-prone. Worth a separate dedup pass at some point, but in scope here is just making them consistent on the operator interface section.

## Validation

```
cargo build -p dora-cli                              clean
cargo build -p dora-operator-api-cxx                  clean
markdown fence parity (guide + docs)                  64 + 78, both even
```

## On the residual `rust::Error` build failure

You flagged that `cargo run --example cxx-dataflow` fails with "no type named 'Error' in namespace 'rust'" — outside this PR's operator diff. I think that's from #1848's example (`examples/c++-dataflow/node-rust-api/main.cc` uses `catch (const rust::Error &e)` around the new introspection calls). `rust::Error` is generated by cxx when an FFI function returns `Result<T>`, so it should be available wherever cxx-generated headers are included. If it's not picking up on the C++ build, that points at either a missing `#include` or the build artifact path not finding the regenerated header.

I'll dig into that as a follow-up after this PR lands — it's not part of #1849's surface and the local-env infra concerns you mentioned earlier (system `~/.cargo/bin/dora` vs workspace, runtime comm parsing) make me think it might be one of those rather than a code bug.

PR head: `e3a59d5c` → `df231ef3`. Ready for re-review.

heyong4725:
Strong argument, applying. Fix in `bab5a3e7`.

The asymmetry was a real design oversight — `on_input` takes the per-event `OutputSender` but my new callbacks didn't, despite the Rust trait providing one for every event. Since this PR is already breaking the C++ operator interface, doing this now avoids a second breaking change later for what is a common pattern (flush buffered state on stop, send a "drain complete" marker when an input closes).

## Updated signatures

```cpp
DoraOnInputResult on_input_closed(Operator& op, rust::Str id, OutputSender& output_sender);
DoraOnInputResult on_stop(Operator& op, OutputSender& output_sender);
```

The cxx::bridge sigs wrap the per-event sender in the match arms identically to the existing `Event::Input` path:

```rust
Event::InputClosed { id } => {
    let operator = self.operator.as_mut().unwrap();
    let mut output_sender = OutputSender(output_sender);
    let result = ffi::on_input_closed(operator, id, &mut output_sender);
    ...
}
```

## All affected surfaces updated

| File | Change |
|---|---|
| `apis/c++/operator/src/lib.rs` | cxx::bridge sigs + match arms thread `output_sender` through |
| `examples/c++-dataflow/operator-rust-api/operator.{h,cc}` | Signatures + bodies; `(void)output_sender` with a comment pointing at `send_output` for real ops |
| `examples/cmake-dataflow/operator-rust-api/operator.{h,cc}` | Same |
| `binaries/cli/src/template/cxx/operator-template.{h,cc}` | Same — `dora new` users now scaffold the symmetric three-callback interface |
| `docs/api-cxx.md` | Required-interface example + bullet descriptions mention `output_sender` and the `send_output(output_sender, ...)` use case; "Default implementations" paragraph notes the symmetric-with-on_input design |
| `guide/src/languages/cpp.md` | Mirrored |
| `Changelog.md` "Breaking" entry | Shows the full prototype with `OutputSender& output_sender` + explicit rationale (symmetric to `on_input`, enables emitting final/status outputs in response to close/stop) |

## Validation

```
cargo build -p dora-operator-api-cxx                     clean
cargo clippy -p dora-operator-api-cxx -- -D warnings     clean
cargo build -p dora-cli                                  clean (template baked into binary)
cargo fmt --check                                        clean
markdown fence parity                                    78 (api-cxx) + 64 (guide cpp.md), both even
```

## Pattern bank update for me

When reviewing a breaking-change PR that adds new callbacks alongside existing ones, check the signature symmetry with the closest analogue. The existing `on_input` takes `(op, id, data, output_sender)`; the new `on_input_closed` only took `(op, id)`. Asymmetric signatures across logically-paired callbacks should be flagged on the first round — "deliberate or accident?" is a question that deserves an explicit answer, not silent acceptance.

PR head: `df231ef3` → `bab5a3e7`. Ready for re-review.

PR #1593, opened 2026-04-03
fix(coordinator): timeout stuck spawn results and rollback deterministically

Bhanudahiyaa:
## Related Issue

  Closes: #1592 

  ## Summary

  This PR adds a coordinator-side spawn result timeout path for distributed starts.
  If one or more daemons do not report `spawn_result` within the timeout window, the coordinator now:

  1. fails spawn immediately,
  2. unblocks waiters (`wait_for_spawn`) with an explicit error,
  3. performs deterministic rollback by sending `stop_dataflow` to assigned daemons.

  ## Problem statement

Before this change, if a daemon accepted `spawn` but never reported `spawn_result`, spawn completion could remain pending until client-side RPC deadlines. This caused long waits, unclear failure modes, and non deterministic rollback behavior under partial startup failures.

  ## Why this is high-impact

  This improves a critical cross-boundary lifecycle path:

  - CLI / API semantics (`start` + `wait_for_spawn`) now match runtime failure reality.
  - Coordinator state converges faster under failure.
  - Partial distributed starts are actively rolled back, reducing operational risk.
  - Error reporting becomes actionable rather than timeout-opaque.

  ## What changed

  ### 1) Spawn timeout constant in coordinator
  - Added `SPAWN_RESULT_TIMEOUT` in coordinator core.

  ### 2) Track spawn start time in runtime state
  - Extended `RunningDataflow` with `spawn_started_at: Instant`.
  - Initialized timestamp at dataflow insertion time (before spawn dispatch).

  ### 3) Timeout enforcement in heartbeat path
  - On each `DaemonHeartbeatInterval`, coordinator now checks for timed-out pending spawn states:
    - `spawn_result` still pending,
    - `pending_spawn_results` non-empty,
    - elapsed > timeout.

  ### 4) Fail fast on timeout
  - Coordinator sets `spawn_result` to an error immediately.
  - This ensures `wait_for_spawn` waiters are released promptly with explicit cause.

  ### 5) Deterministic rollback on timeout
  - Coordinator attempts rollback stop RPCs to daemons assigned to the timed-out dataflow.
  - Reused existing rollback logic from distributed spawn dispatch failure path (shared helper).

  ### 6) Internal helper visibility adjustment
  - Made rollback helper `pub(super)` so timeout path and dispatch-failure path use the same implementation.

  ## Design decisions and tradeoffs

  1. **Coordinator-only change**
     - No wire-protocol/message schema changes.
     - Keeps compatibility risk low and rollout simple.

[...]

Issue #1592, opened 2026-04-03
Coordinator: Timeout Stuck Spawn Results + Deterministic Rollback for Distributed Starts


Bhanudahiyaa:
## Summary

 When starting distributed dataflows, a daemon can successfully receive the `spawn` RPC but never report `spawn_result` back to the coordinator (for example due to an internal stuck path, partial failure path, or lost callback/reporting path). In this situation, the coordinator currently keeps the spawn in a pending state until client RPC deadlines expire.

 This creates long, non deterministic failure behavior and weak rollback guarantees.

  ## Problem

The coordinator tracks distributed spawn completion through daemon `spawn_result` notifications. If at least one daemon never reports completion:

  - `wait_for_spawn` can block until timeout instead of failing quickly.
  - Operators get delayed feedback and unclear failure causes.
  - Dataflow lifecycle may remain in a pending/ambiguous state longer than necessary.
  - Partially started systems are not deterministically rolled back under this specific failure mode.

  This is a system boundary reliability gap (coordinator ↔ daemon) in the start lifecycle.

  ## Why this matters

  This affects production reliability and operator experience:

  1. Correctness:
     - Start should not remain pending indefinitely when progress has effectively stopped.
  2. Reproducibility / operational determinism:
     - Failures should produce consistent outcomes and rollback behavior.
  3. Cross-layer consistency:
     - CLI `start`/`wait_for_spawn` semantics should align with runtime realities.
  4. Failure handling:
     - Partial distributed starts should always converge to a known, safe state.

  ## Expected behavior

  If spawn acknowledgements are still pending beyond a coordinator defined deadline:

  1. Coordinator marks spawn as failed immediately.
  2. Any `wait_for_spawn` waiter is unblocked with a clear error.
  3. Coordinator sends rollback `stop_dataflow` to daemons assigned to that dataflow.
  4. Logs include actionable context:
     - which daemons were still pending,
     - which daemons had already acknowledged spawn,
     - rollback success/failure details.

  ## Actual behavior

Coordinator can remain in pending spawn state for the dataflow until higher level client timeouts, with no deterministic spawn deadline rollback path.

  ## Proposed solution

  Add coordinator side spawn result deadline enforcement:

  - Track a per dataflow spawn start timestamp.
  - On heartbeat, detect dataflows where:
    - spawn is still pending,
    - pending daemons remain,
    - elapsed time > timeout threshold.
  - Immediately fail spawn result and run rollback stop RPCs across assigned daemons.
[...]

PR #1854, opened 2026-05-18
feat(coordinator): timeout stuck spawns and roll back deterministically (rescue of #1593)

heyong4725:
Rescue of #1593 (issue #1592) against current main.

@Bhanudahiyaa's original PR diverged structurally during the 1.0 consolidation — the `RunningDataflow` struct moved, the rollback helper moved, the heartbeat handler was substantially refactored. The design is identical to the original, the implementation sites had to be redone.

## What this fixes

In distributed `dora start`:

1. coordinator dispatches `Spawn` RPC to each assigned daemon
2. each daemon must report `spawn_result` back; coordinator fires `spawn_result.set_result(Ok)` once all are in
3. CLI / API caller waits on `wait_for_spawn`

If a daemon accepts the `Spawn` RPC but **never reports back** — internal flow hung, OR daemon disconnects mid-spawn and the disconnect path clears `pending_spawn_results` without firing `spawn_result` — `wait_for_spawn` hangs until the **client-side** RPC deadline. Failure is timeout-opaque, slow, and leaves partially-spawned daemons running.

## The fix

Heartbeat-driven watchdog at 3s cadence, gated on a **60s overdue threshold**. For every `RunningDataflow` whose `spawn_result` is still `Pending` AND whose `spawn_started_at.elapsed() > SPAWN_RESULT_TIMEOUT`:

1. Compute daemons that already reported successful spawn (`daemons - pending_spawn_results`) → these need rollback
2. `rollback_spawned_daemons(...)` issues `StopDataflow{force: true}` to each (best-effort; failures logged but don't block)
3. Set `spawn_result` to `Err(...)` naming the timeout + counts → releases all `wait_for_spawn` waiters
4. Persist as `Failed` so a restarted coordinator doesn't re-resurrect it

`set_result` is idempotent (no-op on `Cached`), so subsequent heartbeats don't double-fail or double-rollback.

## Catches both reported failure modes

| Mode | How the watchdog catches it |
|---|---|
| Daemon stuck-but-connected (still heartbeating, never reports `spawn_result`) | 30s disconnect path doesn't trip; watchdog catches at 60s |
| Daemon disconnects mid-spawn | Existing disconnect path clears `pending_spawn_results` but doesn't fire `spawn_result`; watchdog catches the leftover `Pending` state at next tick past deadline |

## Design

- **Coordinator-only.** No wire-protocol or message-schema changes. Compatibility risk bounded to coordinator behavior.
- **Heartbeat-driven, not background timer.** Granularity bounded by heartbeat cadence (3s). Avoids per-dataflow task overhead.
- **`spawn_result.is_pending()` over `pending_spawn_results.is_empty()`.** Watchdog fires regardless of whether the disconnect path already cleared `pending_spawn_results` — the invariant is "waiter not released", not "daemon-set state".
- **60s threshold.** Conservative — well above realistic per-daemon spawn time even with `--uv` Python venv preparation. Configurable would be a follow-up if needed.

## Implementation map

| File | Change |
|---|---|
| `state.rs` | Add `spawn_started_at: Instant` field to `RunningDataflow`. Add `CachedResult::is_pending() -> bool`. |
| `handlers.rs` | Initialize `spawn_started_at: Instant::now()` in `start_dataflow`'s `RunningDataflow` constructor. |
| `lib.rs` | New `SPAWN_RESULT_TIMEOUT` const, new `check_spawn_timeouts()` async helper invoked from `Event::DaemonHeartbeatInterval`. Initialize `spawn_started_at` in test constructor. |
| `run/mod.rs` | `rollback_spawned_daemons` visibility: `async fn` → `pub(crate) async fn`. Reuses existing helper instead of duplicating. |

## Tests

```
test tests::cached_result_is_pending_distinguishes_pending_and_cached ... ok
test tests::check_spawn_timeouts_fires_error_when_pending_past_deadline ... ok
test tests::check_spawn_timeouts_no_op_when_within_deadline ... ok
test tests::check_spawn_timeouts_idempotent_on_already_cached_result ... ok
```

The idempotency test specifically asserts that a watchdog pass on an already-`Cached(Ok)` result preserves the cached `Ok(DataflowSpawned)` — verifies the watchdog never clobbers a successful spawn.

## Validation

```
cargo test -p dora-coordinator                            70/70 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

## Test plan

- [x] All 70 coordinator tests pass
- [x] New tests cover fires-on-deadline, no-op-pre-deadline, no-op-on-cached, predicate semantics
- [x] Clippy + fmt clean on the diff
- [ ] CI green
- [ ] Manual: kick off a dataflow against a daemon that's wedged before sending `spawn_result`; verify `wait_for_spawn` errors out at ~60s with the watchdog message rather than hanging until client RPC timeout

Co-authored-by: Bhanu Dahiya <@Bhanudahiyaa>

Rescue closes the same issue (#1592). Original PR (#1593) can be closed once this lands.

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -617,6 +617,12 @@ pub(crate) async fn start_dataflow(
         topic_subscribers: BTreeMap::new(),
         daemon_ack_sequence: daemons.iter().map(|d| (d.clone(), 0)).collect(),
         pending_spawn_results: daemons,
+        // Capture the spawn-dispatch start time before any per-daemon
+        // spawn RPC has fired so the watchdog can compare against the
+        // earliest meaningful moment. The dispatch itself happened in
+        // `spawn_dataflow` just above; the `Instant::now()` here is at
+        // most a few microseconds late.
+        spawn_started_at: std::time::Instant::now(),
         created_at: state::now_millis(),
         store_generation: 0,
         last_recovery_attempt: BTreeMap::new(),
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -45,6 +45,16 @@ use uuid::Uuid;
 
 const FALLBACK_REPLAY_BACKOFF: Duration = Duration::from_secs(5);
 
+/// Deadline after which a distributed spawn that has not received every
+/// daemon's `spawn_result` is considered stuck. The heartbeat-driven
+/// watchdog fails such spawns and rolls back any daemons that already
+/// reported success.
+///
+/// Chosen well above realistic per-daemon spawn time (even with `--uv`
+/// Python venv preparation) so the watchdog only fires on genuine hangs.
+/// Rescue of [#1593](https://github.com/dora-rs/dora/pull/1593).
+const SPAWN_RESULT_TIMEOUT: Duration = Duration::from_secs(60);
+
 pub(crate) mod artifacts;
 mod control;
 mod events;
@@ -1855,6 +1865,28 @@ async fn start_inner(
                         }
                     }
                 }
+                // Spawn timeout watchdog: detect distributed starts that are
+                // stuck waiting for one or more `spawn_result` reports and
+                // release `wait_for_spawn` waiters with a clear error rather
+                // than letting them hang on the client-side RPC deadline.
+                // Triggers on either failure mode:
+                //   1. A daemon accepted the spawn RPC but its internal flow
+                //      is hung (still heartbeating, never reports back).
+                //   2. Pending daemons all disconnected; the disconnect path
+                //      above cleared their entries from `pending_spawn_results`
+                //      but does not itself fire `spawn_result`.
+                // Rescue of #1593 (issue #1592).
+                //
+                // SAFETY net only: 60 s is well above realistic per-daemon
+                // spawn time even with `--uv` Python env preparation.
+                check_spawn_timeouts(
+                    &mut running_dataflows,
+                    &mut daemon_connections,
+                    &clock,
+                    store.as_ref(),
+                )
+                .await;
+
                 // Recovery timeout: transition stale Recovering dataflows to Failed.
                 // Dataflows are marked Recovering on coordinator startup and should
                 // be reclaimed by reconnecting daemons within 60 seconds.
@@ -2367,6 +2399,106 @@ fn topic_outputs_by_daemon(
     }
 
     Ok(outputs_by_daemon)
+}
+
+/// Scan `running_dataflows` for spawns that have been pending past
+/// [`SPAWN_RESULT_TIMEOUT`] and resolve them as failed.
+///
+/// For each stuck dataflow:
+/// 1. Roll back any daemons that already reported a successful spawn
+///    (best-effort; failures are logged).
+/// 2. Set `spawn_result` to an error, unblocking `wait_for_spawn` waiters
+///    with an actionable message instead of letting them hang on the
+///    client-side RPC deadline.
+/// 3. Persist the dataflow as `Failed` so a restarted coordinator does not
+///    re-resurrect it.
+///
+/// Idempotent: `CachedResult::set_result` is a no-op once the result has
+/// been cached, so re-running this on subsequent heartbeats does not
+/// double-fail or double-rollback.
+///
+/// Rescue of [#1593](https://github.com/dora-rs/dora/pull/1593)
+/// (issue [#1592](https://github.com/dora-rs/dora/issues/1592)).
+async fn check_spawn_timeouts(
+    running_dataflows: &mut HashMap<DataflowId, RunningDataflow>,
+    daemon_connections: &mut DaemonConnections,
+    clock: &HLC,
+    store: &dyn CoordinatorStore,
+) {
+    // First pass: identify stuck spawns and snapshot the daemon sets we
+    // need for rollback. We collect into an owned Vec so the immutable
+    // borrow of `running_dataflows` drops before we mutate it below.
+    let stuck: Vec<(DataflowId, BTreeSet<DaemonId>, usize)> = running_dataflows
+        .iter()
+        .filter_map(|(uuid, df)| {
+            if df.spawn_result.is_pending() && df.spawn_started_at.elapsed() > SPAWN_RESULT_TIMEOUT
+            {
+                // Daemons assigned to this dataflow that already reported
+                // successful spawn (i.e. were removed from
+                // `pending_spawn_results`). These are the ones we need to
+                // roll back to avoid leaving partial state running.
+                let succeeded: BTreeSet<DaemonId> = df
+                    .daemons
+                    .difference(&df.pending_spawn_results)
+                    .cloned()
+                    .collect();
+                Some((*uuid, succeeded, df.pending_spawn_results.len()))
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    for (uuid, succeeded_daemons, pending_count) in stuck {
+        tracing::warn!(
+            dataflow = %uuid,
+            timeout_secs = SPAWN_RESULT_TIMEOUT.as_secs(),
+            pending = pending_count,
+            succeeded = succeeded_daemons.len(),
+            "spawn timeout: releasing waiters and rolling back",
+        );
+
+        // Best-effort rollback. Failures here are logged but do not block
+        // firing the spawn_result error — the user must hear about the
+        // timeout regardless of whether stop dispatches succeed.
+        let rollback_errors =
+            run::rollback_spawned_daemons(uuid, &succeeded_daemons, daemon_connections, clock)
+                .await;
+        if !rollback_errors.is_empty() {
+            let rollback_summary = rollback_errors
+                .iter()
+                .map(|(id, e)| format!("  {id}: {e}"))
+                .collect::<Vec<_>>()
+                .join("\n");
+            tracing::warn!(
+                dataflow = %uuid,
+                "rollback partial after spawn timeout, {} stop(s) failed:\n{rollback_summary}",
+                rollback_errors.len(),
+            );
+        }
+
+        // Fire the spawn_result error and persist the dataflow as Failed.
+        if let Some(df) = running_dataflows.get_mut(&uuid) {
+            let err_msg = format!(
+                "spawn timed out after {}s; {} daemon(s) never reported \
+                 spawn_result; rolled back {} previously-started daemon(s)",
+                SPAWN_RESULT_TIMEOUT.as_secs(),
+                pending_count,
+                succeeded_daemons.len(),
+            );
+            df.spawn_result.set_result(Err(eyre!(err_msg.clone())));
+            match df
+                .make_record(StoreDataflowStatus::Failed { error: err_msg })
+                .and_then(|r| store.put_dataflow(&r))
+            {
+                Ok(()) => {}
+                Err(e) => tracing::warn!(
+                    dataflow = %uuid,
+                    "failed to persist spawn timeout: {e}",
+                ),
+            }
+        }
+    }
 }
 
 fn topic_debug_enabled(
@@ -3194,6 +3326,7 @@ mod tests {
             log_subscribers: vec![],
             topic_subscribers: BTreeMap::new(),
             pending_spawn_results: BTreeSet::new(),
+            spawn_started_at: Instant::now(),
             created_at: 0,
             store_generation: 0,
             last_recovery_attempt: BTreeMap::new(),
@@ -4126,5 +4259,164 @@ mod tests {
         let err = ensure_delete_param_forward_applied(&reply, &node_id)
             .expect_err("unexpected reply variant should fail strict forwarding");
         assert!(err.to_string().contains("unexpected daemon reply"));
+    }
+
+    // -------------------------------------------------------------------
+    // Spawn timeout watchdog (rescue of #1593)
+    // -------------------------------------------------------------------
+
+    #[test]
+    fn cached_result_is_pending_distinguishes_pending_and_cached() {
+        let mut r = CachedResult::default();
+        assert!(r.is_pending(), "fresh CachedResult should be Pending");
+
+        r.set_result(Err(eyre!("done")));
+        assert!(!r.is_pending(), "after set_result, should be Cached");
+
+        // set_result on Cached should be a no-op (already covered by
+        // existing semantics but verified here to lock in idempotency).
+        r.set_result(Ok(ControlRequestReply::DataflowSpawned {
+            uuid: DataflowId::from(Uuid::new_v4()),
+        }));
+        assert!(!r.is_pending());
+    }
+
+    #[tokio::test]
+    async fn check_spawn_timeouts_fires_error_when_pending_past_deadline() {
+        let store: Arc<dyn CoordinatorStore> = Arc::new(InMemoryStore::new());
+        let clock = HLC::default();
+        let mut daemon_connections = DaemonConnections::default();
+
+        let dataflow_id = DataflowId::from(Uuid::new_v4());
+        let daemon_id = DaemonId::new(Some("m1".to_string()));
+        let node_id: dora_core::config::NodeId = "sender".to_string().into();
+        let mut df = test_running_dataflow(dataflow_id, daemon_id.clone(), node_id);
+        // Backdate the spawn start so the watchdog treats it as stuck.
+        df.spawn_started_at = Instant::now() - SPAWN_RESULT_TIMEOUT - Duration::from_secs(1);
+        // Still waiting on the daemon's spawn_result; this is the trigger
+        // condition the watchdog is meant to catch.
+        df.pending_spawn_results.insert(daemon_id);
+
+        // Register a waiter so we can prove `wait_for_spawn` unblocks with
+        // an error rather than hanging.
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        df.spawn_result.register(tx);
+
+        let mut running_dataflows = HashMap::new();
+        running_dataflows.insert(dataflow_id, df);
+
+        check_spawn_timeouts(
+            &mut running_dataflows,
+            &mut daemon_connections,
+            &clock,
+            store.as_ref(),
+        )
+        .await;
+
+        // The waiter should now have an error reply.
+        let reply = timeout(TokioDuration::from_secs(1), rx)
+            .await
+            .expect("waiter should resolve, not hang")
+            .expect("sender should not drop");
+        let err = reply.expect_err("timed-out spawn must surface as Err");
+        let msg = format!("{err:?}");
+        assert!(
+            msg.contains("spawn timed out") || msg.contains("timeout"),
+            "error should explain the timeout, got: {msg}"
+        );
+
+        // And the dataflow's spawn_result should now be Cached.
+        let df = running_dataflows
+            .get(&dataflow_id)
+            .expect("df still present");
+        assert!(
+            !df.spawn_result.is_pending(),
+            "spawn_result must transition out of Pending after timeout fires"
+        );
+    }
+
+    #[tokio::test]
+    async fn check_spawn_timeouts_no_op_when_within_deadline() {
+        let store: Arc<dyn CoordinatorStore> = Arc::new(InMemoryStore::new());
+        let clock = HLC::default();
+        let mut daemon_connections = DaemonConnections::default();
+
+        let dataflow_id = DataflowId::from(Uuid::new_v4());
+        let daemon_id = DaemonId::new(Some("m1".to_string()));
+        let node_id: dora_core::config::NodeId = "sender".to_string().into();
+        let mut df = test_running_dataflow(dataflow_id, daemon_id.clone(), node_id);
+        // Fresh spawn — well within the deadline.
+        df.spawn_started_at = Instant::now();
+        df.pending_spawn_results.insert(daemon_id);
+
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        df.spawn_result.register(tx);
+
+        let mut running_dataflows = HashMap::new();
+        running_dataflows.insert(dataflow_id, df);
+
+        check_spawn_timeouts(
+            &mut running_dataflows,
+            &mut daemon_connections,
+            &clock,
+            store.as_ref(),
+        )
+        .await;
+
+        // The waiter should still be pending.
+        let polled = tokio::time::timeout(TokioDuration::from_millis(50), rx).await;
+        assert!(polled.is_err(), "waiter must still be pending pre-deadline");
+
+        let df = running_dataflows
+            .get(&dataflow_id)
+            .expect("df still present");
+        assert!(
+            df.spawn_result.is_pending(),
+            "spawn_result must remain Pending pre-deadline"
+        );
+    }
+
+    #[tokio::test]
+    async fn check_spawn_timeouts_idempotent_on_already_cached_result() {
+        let store: Arc<dyn CoordinatorStore> = Arc::new(InMemoryStore::new());
+        let clock = HLC::default();
+        let mut daemon_connections = DaemonConnections::default();
+
+        let dataflow_id = DataflowId::from(Uuid::new_v4());
+        let daemon_id = DaemonId::new(Some("m1".to_string()));
+        let node_id: dora_core::config::NodeId = "sender".to_string().into();
+        let mut df = test_running_dataflow(dataflow_id, daemon_id, node_id);
+        df.spawn_started_at = Instant::now() - SPAWN_RESULT_TIMEOUT - Duration::from_secs(1);
+        // Spawn already resolved successfully — the watchdog must NOT
+        // re-fire on subsequent heartbeats and clobber the cached result.
+        df.spawn_result
+            .set_result(Ok(ControlRequestReply::DataflowSpawned {
+                uuid: dataflow_id,
+            }));
+
+        let mut running_dataflows = HashMap::new();
+        running_dataflows.insert(dataflow_id, df);
+
+        check_spawn_timeouts(
+            &mut running_dataflows,
+            &mut daemon_connections,
+            &clock,
+            store.as_ref(),
+        )
+        .await;
+
+        let df = running_dataflows
+            .get_mut(&dataflow_id)
+            .expect("df still present");
+        // Registering a new waiter on a Cached result must immediately
+        // deliver the cached Ok — proving the watchdog did not clobber it.
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        df.spawn_result.register(tx);
+        let reply = timeout(TokioDuration::from_millis(50), rx)
+            .await
+            .expect("Cached result should deliver immediately")
+            .expect("sender should not drop");
+        let ok = reply.expect("the cached Ok result must be preserved");
+        assert!(matches!(ok, ControlRequestReply::DataflowSpawned { .. }));
     }
 }
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -59,21 +59,6 @@ const FALLBACK_REPLAY_BACKOFF: Duration = Duration::from_secs(5);
 /// Rescue of [#1593](https://github.com/dora-rs/dora/pull/1593).
 const SPAWN_RESULT_TIMEOUT_DEFAULT: Duration = Duration::from_secs(60);
 
-/// Wall-clock budget for the entire rollback batch issued by the spawn
-/// timeout watchdog. Caps how long the heartbeat handler can block dispatching
-/// stop RPCs to (possibly stuck) succeeded daemons.
-///
-/// Without this cap, a worst-case rollback of N succeeded-but-now-wedged
-/// daemons could block the heartbeat handler for up to `N * TCP_READ_TIMEOUT`
-/// (30 s each), during which heartbeats to *other* healthy daemons are not
-/// dispatched. Since the daemon-disconnect threshold is also 30 s, that delay
-/// can falsely declare healthy daemons dead during recovery -- a watchdog
-/// destabilising healthy components is an anti-pattern.
-///
-/// 5 s is generous for fast-failing stops yet bounded enough that healthy
-/// daemons stay within heartbeat tolerance.
-const SPAWN_TIMEOUT_ROLLBACK_BUDGET: Duration = Duration::from_secs(5);
-
 /// Reads `DORA_SPAWN_RESULT_TIMEOUT_SECS` once and caches the result so
 /// each heartbeat tick is not repeatedly parsing the env var. Invalid or
 /// non-positive values fall back to [`SPAWN_RESULT_TIMEOUT_DEFAULT`].
@@ -2149,19 +2134,12 @@ async fn start_inner(
                     dataflow.pending_spawn_results.remove(&daemon_id);
                     match result {
                         Ok(()) => {
-                            if dataflow.pending_spawn_results.is_empty() {
-                                tracing::info!("successfully spawned dataflow `{dataflow_id}`",);
-                                dataflow.spawn_result.set_result(Ok(
-                                    ControlRequestReply::DataflowSpawned { uuid: dataflow_id },
-                                ));
-                                // Persist: dataflow now running
-                                if let Err(e) = dataflow
-                                    .make_record(StoreDataflowStatus::Running)
-                                    .and_then(|r| store.put_dataflow(&r))
-                                {
-                                    tracing::warn!("failed to persist dataflow running: {e}");
-                                }
-                            }
+                            handle_spawn_result_ok(
+                                dataflow,
+                                dataflow_id,
+                                &daemon_id,
+                                store.as_ref(),
+                            );
                         }
                         Err(err) => {
                             tracing::warn!("error while spawning dataflow `{dataflow_id}`");
@@ -2446,6 +2424,114 @@ fn topic_outputs_by_daemon(
     Ok(outputs_by_daemon)
 }
 
+/// Handle the success arm of `Event::DataflowSpawnResult`.
+///
+/// Extracted from the inline event loop so the late-arrival guard can be
+/// exercised directly by unit tests. Two correctness invariants:
+///
+/// 1. **Late-Ok guard**: if `spawn_result` is no longer `Pending` (i.e. the
+///    timeout watchdog already cached an `Err`), do NOT persist `Running`.
+///    Without this guard, a delayed daemon's `Ok` arriving after the
+///    watchdog fires would resurrect a terminally-failed dataflow as
+///    `Running` in the store — in-memory `spawn_result` would still be
+///    `Cached(Err)` (because `set_result` is a no-op on `Cached`), so the
+///    store and the in-memory state would silently diverge. Rescue of
+///    [#1593](https://github.com/dora-rs/dora/pull/1593) Finding 1 follow-up.
+///
+/// 2. **All-daemons-succeeded check**: only mark the dataflow Running once
+///    every assigned daemon has reported a successful spawn. Preserves the
+///    pre-existing handler semantics.
+fn handle_spawn_result_ok(
+    dataflow: &mut RunningDataflow,
+    dataflow_id: DataflowId,
+    daemon_id: &DaemonId,
+    store: &dyn CoordinatorStore,
+) {
+    // Guard: if the spawn was already terminally failed (by the watchdog
+    // or by another daemon reporting an error), a late Ok must not
+    // promote the store status back to Running.
+    if !dataflow.spawn_result.is_pending() {
+        tracing::warn!(
+            dataflow = %dataflow_id,
+            daemon = %daemon_id,
+            "ignoring late successful spawn_result on a dataflow already \
+             terminally failed (e.g. by the spawn-timeout watchdog)",
+        );
+        return;
+    }
+
+    if dataflow.pending_spawn_results.is_empty() {
+        tracing::info!("successfully spawned dataflow `{dataflow_id}`");
+        dataflow
+            .spawn_result
+            .set_result(Ok(ControlRequestReply::DataflowSpawned {
+                uuid: dataflow_id,
+            }));
+        if let Err(e) = dataflow
+            .make_record(StoreDataflowStatus::Running)
+            .and_then(|r| store.put_dataflow(&r))
+        {
+            tracing::warn!("failed to persist dataflow running: {e}");
+        }
+    }
+}
+
+/// Fire-and-forget compensating rollback used by [`check_spawn_timeouts`].
+///
+/// Enqueues `StopDataflow{force: true}` to each succeeded daemon via
+/// `connection.send()` (no reply wait). Unlike
+/// `run::rollback_spawned_daemons`, this does NOT use `send_and_receive` —
+/// see the comment in `check_spawn_timeouts` for why: (a) reply-waiting
+/// blocks the heartbeat handler for up to `TCP_READ_TIMEOUT` per wedged
+/// daemon, risking false-positive disconnections elsewhere in the cluster;
+/// (b) `send_and_receive` is not cancellation-safe (it inserts into
+/// `pending_replies` *before* registering its own cleanup), so any
+/// external timeout that cancels it leaks state.
+///
+/// Returns `(daemon_id, error)` pairs for dispatch failures (serialization
+/// failure or absent daemon connection). The mpsc `send` itself is
+/// near-instantaneous so total wall-clock is bounded regardless of N.
+async fn fire_and_forget_rollback(
+    dataflow_id: Uuid,
+    spawned_daemons: &BTreeSet<DaemonId>,
+    daemon_connections: &mut DaemonConnections,
+    clock: &HLC,
+) -> Vec<(DaemonId, String)> {
+    if spawned_daemons.is_empty() {
+        return Vec::new();
+    }
+
+    let stop_message = match serde_json::to_vec(&Timestamped {
+        inner: DaemonCoordinatorEvent::StopDataflow {
+            dataflow_id,
+            grace_duration: None,
+            force: true,
+        },
+        timestamp: clock.new_timestamp(),
+    }) {
+        Ok(m) => m,
+        Err(e) => {
+            // Serialization failure is unlikely but fatal for rollback.
+            return spawned_daemons
+                .iter()
+                .map(|id| (id.clone(), format!("failed to serialize stop message: {e}")))
+                .collect();
+        }
+    };
+
+    let mut errors = Vec::new();
+    for daemon_id in spawned_daemons {
+        let Some(conn) = daemon_connections.get_mut(daemon_id) else {
+            errors.push((daemon_id.clone(), "no daemon connection".to_string()));
+            continue;
+        };
+        if let Err(e) = conn.send(&stop_message).await {
+            errors.push((daemon_id.clone(), format!("send failed: {e}")));
+        }
+    }
+    errors
+}
+
 /// Scan `running_dataflows` for spawns that have been pending past
 /// [`spawn_result_timeout`] and resolve them as failed.
 ///
@@ -2503,41 +2589,40 @@ async fn check_spawn_timeouts(
             "spawn timeout: releasing waiters and rolling back",
         );
 
-        // Best-effort rollback, bounded by SPAWN_TIMEOUT_ROLLBACK_BUDGET.
-        // Failures here are logged but do not block firing the spawn_result
-        // error -- the user must hear about the timeout regardless of
-        // whether stop dispatches succeed. The wall-clock budget caps how
-        // long this heartbeat tick can block: without it, N wedged
-        // succeeded daemons could block heartbeat dispatch for
-        // N * TCP_READ_TIMEOUT (30s each), risking false-positive
-        // disconnections of healthy daemons (Finding 1 / cascade risk).
-        let rollback_outcome = tokio::time::timeout(
-            SPAWN_TIMEOUT_ROLLBACK_BUDGET,
-            run::rollback_spawned_daemons(uuid, &succeeded_daemons, daemon_connections, clock),
-        )
-        .await;
-        match rollback_outcome {
-            Ok(rollback_errors) if !rollback_errors.is_empty() => {
-                let rollback_summary = rollback_errors
-                    .iter()
-                    .map(|(id, e)| format!("  {id}: {e}"))
-                    .collect::<Vec<_>>()
-                    .join("\n");
-                tracing::warn!(
-                    dataflow = %uuid,
-                    "rollback partial after spawn timeout, {} stop(s) failed:\n{rollback_summary}",
-                    rollback_errors.len(),
-                );
-            }
-            Ok(_) => {}
-            Err(_elapsed) => {
-                tracing::warn!(
-                    dataflow = %uuid,
-                    budget_secs = SPAWN_TIMEOUT_ROLLBACK_BUDGET.as_secs(),
-                    "spawn-timeout rollback exceeded budget; \
-                     some succeeded daemons may not have been stopped",
-                );
-            }
+        // Fire-and-forget rollback: enqueue StopDataflow on each succeeded
+        // daemon WITHOUT awaiting a reply. This solves two problems at once:
+        //
+        // 1. Cascade-failure risk: a reply-awaiting rollback (the original
+        //    `run::rollback_spawned_daemons` path) blocks
+        //    `TCP_READ_TIMEOUT = 30s` per wedged daemon. With N wedged
+        //    daemons, the heartbeat handler would block ~N*30s, during
+        //    which heartbeats to *other* healthy daemons aren't dispatched
+        //    and they trip the 30s disconnect threshold.
+        // 2. Cancellation safety: an earlier version wrapped
+        //    `rollback_spawned_daemons` in `tokio::time::timeout`, but
+        //    that future cancels mid-`send_and_receive`, which inserts a
+        //    pending reply *before* registering its own cleanup -- the
+        //    cancellation would leak `pending_replies` entries.
+        //    `connection.send()` is just an mpsc enqueue; no pending state,
+        //    no cleanup needed, fully cancellation-safe.
+        //
+        // Trade-off: we don't get per-daemon ack of "stop succeeded". For
+        // the watchdog this is acceptable -- the user is already getting
+        // a clear timeout error, and unstopped daemons will be reclaimed
+        // by daemon-disconnect or operator `dora stop`.
+        let rollback_errors =
+            fire_and_forget_rollback(uuid, &succeeded_daemons, daemon_connections, clock).await;
+        if !rollback_errors.is_empty() {
+            let rollback_summary = rollback_errors
+                .iter()
+                .map(|(id, e)| format!("  {id}: {e}"))
+                .collect::<Vec<_>>()
+                .join("\n");
+            tracing::warn!(
+                dataflow = %uuid,
+                "rollback partial after spawn timeout, {} dispatch(es) failed:\n{rollback_summary}",
+                rollback_errors.len(),
+            );
         }
 
         // Fire the spawn_result error and persist the dataflow as Failed.
@@ -4485,14 +4570,13 @@ mod tests {
 
     /// Covers Finding 2 from the self-review on PR #1854: exercises the
     /// rollback dispatch path with a *non-empty* succeeded set. The other
-    /// timeout tests have `succeeded = {}`, so `rollback_spawned_daemons`
+    /// timeout tests have `succeeded = {}`, so the rollback helper
     /// short-circuits at its `if spawned_daemons.is_empty()` early-exit
     /// and never actually sends a stop message.
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn check_spawn_timeouts_dispatches_rollback_stop_to_succeeded_daemon() {
         #[derive(serde::Deserialize)]
         struct OutboundRaw {
-            id: String,
             params: Timestamped<DaemonCoordinatorEvent>,
         }
 
@@ -4503,9 +4587,9 @@ mod tests {
         let daemon_a = DaemonId::new(Some("daemon-a".to_string())); // succeeded
         let daemon_b = DaemonId::new(Some("daemon-b".to_string())); // still pending
 
-        // Mock daemon `a` -- replies to StopDataflow with Ok so the rollback
-        // completes inside the budget; an absent reply would block on
-        // TCP_READ_TIMEOUT and trip the budget timeout instead.
+        // Mock daemon `a` -- fire-and-forget rollback uses `connection.send()`
+        // which only enqueues; no reply is awaited. We just need to capture
+        // the outbound message and verify StopDataflow was dispatched.
         let (tx_a, mut rx_a) = tokio::sync::mpsc::channel::<String>(8);
         let pending_replies_a = Arc::new(tokio::sync::Mutex::new(HashMap::new()));
         let conn_a =
@@ -4518,20 +4602,12 @@ mod tests {
         let daemon_a_task = tokio::spawn(async move {
             while let Some(outbound) = rx_a.recv().await {
                 let outbound_raw: OutboundRaw = serde_json::from_str(&outbound).unwrap();
-                let request_id = Uuid::parse_str(&outbound_raw.id).expect("valid request id");
-                let reply = match outbound_raw.params.inner {
+                match outbound_raw.params.inner {
                     DaemonCoordinatorEvent::StopDataflow { .. } => {
                         *stop_seen_task.lock().await = true;
-                        serde_json::to_string(&DaemonCoordinatorReply::StopResult(Ok(()))).unwrap()
                     }
                     other => panic!("unexpected event on daemon-a in rollback test: {other:?}"),
-                };
-                let reply_tx = pending_replies_a
-                    .lock()
-                    .await
-                    .remove(&request_id)
-                    .expect("pending reply sender should exist");
-                let _ = reply_tx.send(reply);
+                }
             }
         });
 
@@ -4559,8 +4635,20 @@ mod tests {
         .await;
 
         // 1. The mock daemon must have received a stop dispatch.
+        //    Fire-and-forget rollback enqueues into the mpsc channel; the
+        //    daemon task picks it up asynchronously, so poll briefly.
+        let saw_stop = timeout(TokioDuration::from_secs(1), async {
+            loop {
+                if *stop_seen.lock().await {
+                    return true;
+                }
+                tokio::time::sleep(TokioDuration::from_millis(5)).await;
+            }
+        })
+        .await
+        .unwrap_or(false);
         assert!(
-            *stop_seen.lock().await,
+            saw_stop,
             "rollback must dispatch StopDataflow to the succeeded daemon"
         );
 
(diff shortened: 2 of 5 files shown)


heyong4725:
## Self-review

Adversarial pass on PR #1854 (the rescue of #1593) at head `5661b437`. The fix is structurally correct and the unit tests pass, but found **one HIGH and three LOW concerns** that warrant either fixing or explicit acknowledgement.

## Verified correct

- **Wiring**: `check_spawn_timeouts` is invoked from `Event::DaemonHeartbeatInterval` after disconnect cleanup, with `&mut running_dataflows` + `&mut daemon_connections` + `&clock` + `store.as_ref()`. Call site at `binaries/coordinator/src/lib.rs:1875`.
- **`spawn_started_at` initialization**: set in the single `RunningDataflow` constructor in `handlers.rs:625` (real path) and `lib.rs:3334` (test helper). All construction sites covered.
- **Borrow handling**: the `iter().filter_map().collect()` snapshot drops the immutable borrow on `running_dataflows` before `rollback_spawned_daemons` takes `&mut daemon_connections` and before `get_mut(&uuid)` reborrows mutably for `spawn_result.set_result()`. ✓ No reborrow conflict.
- **Idempotency**: `CachedResult::set_result` is a no-op on `Cached`. Second pass through the watchdog on the same dataflow has no effect. Verified by `check_spawn_timeouts_idempotent_on_already_cached_result`.
- **Auto-recovery interaction**: auto-recovery at `lib.rs:2218-2220` skips daemons in `pending_spawn_results`. After watchdog fires, `pending_spawn_results` still contains the wedged daemons, so auto-recovery correctly doesn't try to re-spawn them. And the dataflow's `spawn_result` is now `Cached(Err)`, so the watchdog won't re-fire even if `spawn_started_at.elapsed()` keeps growing. ✓ No interaction bug.

## Finding 1 (HIGH) — Watchdog blocks heartbeat dispatch for up to N × 30s

`run::rollback_spawned_daemons` is **sequential**: each `send_and_receive` is bounded by `TCP_READ_TIMEOUT = Duration::from_secs(30)` (`libraries/message/src/lib.rs:16`). With N succeeded daemons whose stop replies hang, the watchdog blocks the event loop for up to N × 30s.

Concretely:

- 5-machine cluster, 3 daemons reported successful spawn, 2 wedged
- Watchdog fires at 60s
- Tries to roll back the 3 succeeded daemons
- If those daemons are *also* wedged (likely — same network/system issue as the original wedge): `send_and_receive` blocks 30s per daemon → 90s total

During those 90s, the heartbeat handler **does not dispatch heartbeats to any daemon**. The daemon-disconnect threshold is also 30s (`lib.rs:1801`). So healthy daemons in the cluster could be marked as disconnected purely because the coordinator is busy doing rollback on a *different* dataflow.

This is a real cascade-failure risk: a single stuck spawn could cause the coordinator to falsely declare most of the cluster dead during recovery.

**Severity**: HIGH because this only manifests during failure scenarios but actively makes the failure worse. Watchdogs that destabilize healthy components are an anti-pattern.

**Fix options:**

(a) **Bounded total budget.** Cap the entire rollback batch at a short total wall-clock (e.g., 5s); skip remaining daemons if exceeded. Trade-off: leaves some stuck daemons un-rolled-back, but they were going to be un-rolled-back anyway since they're stuck.

(b) **Detached rollback task.** `tokio::spawn` the rollback so it runs concurrently with the rest of the heartbeat dispatch. Challenge: `daemon_connections` is `&mut DaemonConnections`; can't share across tasks. Would need to restructure to do the connection writes from the spawned task and signal completion.

(c) **Don't await rollback at all.** Issue stop messages with `connection.send()` (not `send_and_receive`) and don't wait for replies. Trade-off: weaker rollback guarantee, but we already log errors from the full path and treat them as best-effort.

(c) is the smallest change and matches the "best-effort rollback" framing in the commit message. (a) is a reasonable middle ground.

## Finding 2 (MEDIUM) — Test coverage gap on the rollback path

The 4 tests added cover state-machine semantics but **don't exercise the rollback dispatch with a non-empty succeeded set**:

| Test | What it asserts | What it doesn't reach |
|---|---|---|
| `cached_result_is_pending_distinguishes...` | predicate semantics | — |
| `check_spawn_timeouts_fires_error_when_pending_past_deadline` | Pending→Cached + waiter unblocks | `succeeded` set is empty; rollback returns at the `if spawned_daemons.is_empty()` early-exit |
| `check_spawn_timeouts_no_op_when_within_deadline` | watchdog skips pre-deadline | — |
| `check_spawn_timeouts_idempotent_on_already_cached_result` | no clobber on Cached | — |

So **`run::rollback_spawned_daemons` is never actually called from any test in this PR with a non-empty `spawned_daemons` set**. The "rollback was attempted" promise in the commit message is not verified by the test suite.

**Suggestion:** add `check_spawn_timeouts_attempts_rollback_when_some_daemons_succeeded`:

```rust
// df.daemons = {a, b}, df.pending_spawn_results = {b}
// (i.e., a succeeded, b stuck)
// Register a mock daemon connection for `a`, leave `b` absent.
// Run check_spawn_timeouts.
// Assert: stop message was sent to `a` via the mock daemon channel.
// Assert: spawn_result is Cached(Err) regardless.
```

There's already a pattern for setting up `DaemonConnection` with a `tokio::mpsc::Sender` in `start_topic_debug_stream_rolls_back_on_daemon_error` (`lib.rs:3691+`) — could mirror that.

Persistence-as-Failed isn't asserted either; could be added in the same test by checking `store.list_dataflows()` afterward.

## Finding 3 (LOW) — 60s threshold is hardcoded

`SPAWN_RESULT_TIMEOUT = Duration::from_secs(60)` is a compile-time constant. Operators on slow hardware (e.g., a Raspberry Pi spawning a heavy Python node) might legitimately need a longer threshold; operators on fast hardware diagnosing flaky tests might prefer shorter.

**Acceptable for now** — the original PR also hardcoded; configurability was deferred. Worth a one-line follow-up issue noting this can become an env var (`DORA_SPAWN_RESULT_TIMEOUT_SECS`) or a YAML field. Not a blocker.

## Finding 4 (LOW) — Doc comments reference a behavior that isn't visible in the code

The module-level doc on `check_spawn_timeouts` says:

> The existing disconnect path at `lib.rs:1832` clears `pending_spawn_results` for the disconnected daemon but does **not** fire `spawn_result`.

The actual code does not have any reference to "lib.rs:1832" (verified — `grep -n "lib.rs:" ...` returns nothing in the diff). The line number is only in the **commit message** and **PR body**, which is fine since those are pinned to a specific commit. But the description of "disconnect path does not fire spawn_result" should ideally be cross-linked in code so a future reader who removes this watchdog (thinking it's redundant) understands what they'd break.

**Suggestion:** Add a short code comment near the disconnect cleanup (`lib.rs:1832-1842`) noting "spawn_result is intentionally NOT fired here; the heartbeat watchdog handles this via `check_spawn_timeouts` after `SPAWN_RESULT_TIMEOUT`." So if someone reorganizes the disconnect path later, the contract with the watchdog is visible at the call site.

## What's good

- The fire condition (`is_pending() && elapsed > timeout`) correctly catches both reported failure modes (stuck-but-connected AND disconnect-mid-spawn) with one path. Strict generalization of the original PR's design.
- Reusing `rollback_spawned_daemons` via `pub(crate)` visibility lift instead of duplicating the stop-dispatch logic prevents behavior drift between dispatch-failure and timeout rollbacks. Same intent as the original PR's reasoning.
- The 3-tuple snapshot (`(uuid, succeeded_daemons, pending_count)`) is the right shape — captures everything needed for the rollback + error message before dropping the borrow. Cleaner than a 2-pass loop.
- Error message names the timeout duration, pending count, and rolled-back count — gives the user enough to diagnose without reading coordinator logs.
- Persisting `Failed` via `make_record + put_dataflow` correctly handles the "coordinator restart after timeout" case — the dataflow won't be re-resurrected as Recovering.
- The watchdog gracefully handles store-write failures (logs, continues) — doesn't propagate as a panic or block other dataflows.

## Validation done in review

```
cargo test -p dora-coordinator                            70/70 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

## Recommendation

**Fix Finding 1 before merge** (HIGH — cascade-failure risk on multi-daemon clusters). Most cost-effective fix is option (c): change `rollback_spawned_daemons` to use `connection.send()` instead of `send_and_receive`, or bound the total batch with a wall-clock timeout.

Findings 2-4 are nice-to-haves; can land separately or as part of the same PR.

PR head: `5661b437`. Ready for re-review after Finding 1 is addressed.

---

🤖 Adversarial self-review by Claude — I'm both author and reviewer here; goal is to surface what a fresh reviewer would catch.

heyong4725:
All 4 findings addressed in `d371d817`.

## Finding 1 (HIGH) — Bounded rollback budget

New constant `SPAWN_TIMEOUT_ROLLBACK_BUDGET = 5s` wrapping the rollback batch:

```rust
let rollback_outcome = tokio::time::timeout(
    SPAWN_TIMEOUT_ROLLBACK_BUDGET,
    run::rollback_spawned_daemons(uuid, &succeeded_daemons, daemon_connections, clock),
).await;
match rollback_outcome {
    Ok(rollback_errors) if !rollback_errors.is_empty() => { /* log partial */ }
    Ok(_) => {}
    Err(_elapsed) => {
        tracing::warn!(
            dataflow = %uuid,
            budget_secs = SPAWN_TIMEOUT_ROLLBACK_BUDGET.as_secs(),
            "spawn-timeout rollback exceeded budget; \
             some succeeded daemons may not have been stopped",
        );
    }
}
```

The math now: heartbeat cadence 3s, rollback budget 5s, disconnect threshold 30s. Even if the watchdog fully consumes its budget on every tick (it won't in practice — happy-case stops reply in ms), healthy daemons stay well within tolerance. The cascade-failure path is closed.

Un-stopped daemons after budget exceeded are cleaned up by one of: (a) daemon-disconnect path when they next fail heartbeat, (b) operator `dora stop`, (c) the daemon's own internal timeout if the spawn was hung on a finite resource.

## Finding 2 (MEDIUM) — Rollback dispatch test

New test `check_spawn_timeouts_dispatches_rollback_stop_to_succeeded_daemon` exercises the rollback path with a non-empty succeeded set:

- Sets up `daemons = {a, b}`, `pending_spawn_results = {b}` (a succeeded, b stuck)
- Mocks daemon `a`'s connection over an mpsc channel that replies `StopResult(Ok)`
- After `check_spawn_timeouts`, asserts:
  1. Mock daemon received a `StopDataflow` event (the dispatch promise)
  2. The waiter unblocks with a timeout error message
  3. The dataflow is persisted as `Failed` in the store (cross-restart correctness)

Mirrors the existing `start_topic_debug_stream_rolls_back_on_daemon_error` pattern.

Total test count: 70 → 71 passing.

## Finding 3 (LOW) — Configurable threshold

Constant renamed to `SPAWN_RESULT_TIMEOUT_DEFAULT`, plus `spawn_result_timeout()` helper:

```rust
fn spawn_result_timeout() -> Duration {
    use std::sync::OnceLock;
    static CACHED: OnceLock<Duration> = OnceLock::new();
    *CACHED.get_or_init(|| {
        std::env::var("DORA_SPAWN_RESULT_TIMEOUT_SECS")
            .ok()
            .and_then(|s| s.parse::<u64>().ok())
            .filter(|&n| n > 0)
            .map(Duration::from_secs)
            .unwrap_or(SPAWN_RESULT_TIMEOUT_DEFAULT)
    })
}
```

`OnceLock` ensures the env var is parsed at most once across the coordinator's lifetime — no per-heartbeat env lookup. Invalid or zero values silently fall back to the 60s default (silent rather than panicking because env-var typos shouldn't crash the coordinator).

Users can override via `DORA_SPAWN_RESULT_TIMEOUT_SECS=120` for slow hardware, or `=10` to make flaky-test diagnosis faster.

## Finding 4 (LOW) — Cross-link comment at disconnect path

Added at `lib.rs:1873-1886`:

```rust
// Note: this intentionally does NOT fire `spawn_result.set_result()`
// even if the disconnect leaves `pending_spawn_results` empty.
// The spawn timeout watchdog (`check_spawn_timeouts`, below)
// is the single chokepoint that releases `wait_for_spawn`
// waiters with an error -- it fires regardless of *why*
// `spawn_result` is still Pending past the deadline
// (stuck daemon OR disconnect-mid-spawn). Adding a
// direct set_result here would create two firing paths
// and risk inconsistent error messages.
```

Now a future reader who tries to "fix" the disconnect path by adding a direct `set_result` call will see why that would be wrong before they make the change.

## Validation

```
cargo test -p dora-coordinator                            71/71 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

PR head: `5661b437` → `d371d817`. Ready for re-review.

heyong4725:
Both confirmed and fixed in `79c2afa6`. Both findings are real — the late-Ok resurrection has data-integrity consequences for the store, and the cancellation unsafety leaks state on every wedged-daemon scenario the watchdog is supposed to handle. Thanks for both catches.

## Finding 1 (HIGH) — Late-Ok resurrection fix

Extracted the success arm of `Event::DataflowSpawnResult` into `handle_spawn_result_ok(dataflow, dataflow_id, daemon_id, store)`. Two reasons for the extraction over an inline guard:

1. **Direct testability**: the test must drive the actual production code, not an inline copy that drifts during refactors.
2. **Visibility**: the guard is a load-bearing safety invariant; living in its own named function with a docstring is harder to silently break.

The guard:

```rust
fn handle_spawn_result_ok(
    dataflow_id: DataflowId,
    daemon_id: &DaemonId,
    store: &dyn CoordinatorStore,
) {
    // Guard: if the spawn was already terminally failed (by the watchdog
    // or by another daemon reporting an error), a late Ok must not
    // promote the store status back to Running.
    if !dataflow.spawn_result.is_pending() {
        tracing::warn!(...);
        return;
    }
    if dataflow.pending_spawn_results.is_empty() {
        // ... existing success path, unchanged
    }
}
```

Two new tests against the helper:

- `late_successful_spawn_result_does_not_resurrect_failed_dataflow`: seeds spawn_result with `Cached(Err)` and store with `Failed`, then drives the helper as the event loop would. Asserts store stays `Failed` and waiter still gets the `Err`.
- `handle_spawn_result_ok_promotes_to_running_when_pending_and_complete`: companion happy-path test verifying an over-eager guard doesn't block the legitimate success path. Catches a future refactor that flips the condition inversely.

## Finding 2 (MEDIUM) — Cancellation-unsafe rollback fix

You're right that `send_and_receive` is not cancellation-safe. State.rs:125 inserts into `pending_replies` *before* registering cleanup at line 137. Wrapping it in an external `tokio::time::timeout` and letting that timeout cancel the future leaks the pending_replies entry on every wedged-daemon stop.

The fix isn't to make `send_and_receive` cancellation-safe (that would require a `Drop` guard on its registration, restructuring the whole reply path) — it's to **not use `send_and_receive` from the watchdog at all**.

New helper `fire_and_forget_rollback` uses `connection.send()` instead:

```rust
async fn fire_and_forget_rollback(
    dataflow_id: Uuid,
    spawned_daemons: &BTreeSet<DaemonId>,
    daemon_connections: &mut DaemonConnections,
    clock: &HLC,
) -> Vec<(DaemonId, String)> {
    // ... build stop message ...
    for daemon_id in spawned_daemons {
        let Some(conn) = daemon_connections.get_mut(daemon_id) else {
            errors.push((daemon_id.clone(), "no daemon connection".to_string()));
            continue;
        };
        if let Err(e) = conn.send(&stop_message).await {
            errors.push((daemon_id.clone(), format!("send failed: {e}")));
        }
    }
    errors
}
```

`connection.send()` is just an mpsc enqueue (state.rs:149-161). No pending state, no cleanup needed, cancellation-safe by construction. And it's near-instantaneous regardless of daemon responsiveness, so the **original cascade-failure problem (Finding 1 from the earlier round) is solved at the source** — no per-daemon 30s wait means no N×30s blocking. The `SPAWN_TIMEOUT_ROLLBACK_BUDGET` constant is removed (no longer needed).

Trade-off: we no longer get per-daemon ack of "stop succeeded". For the watchdog this is acceptable:
- The user is already getting a clear timeout error
- Wedged daemons that don't process the stop will be reclaimed via daemon-disconnect (when they next fail a heartbeat) or operator `dora stop`
- The dispatch-failure path's `rollback_spawned_daemons` is unchanged (visibility reverted to private) — that path's caller wants the reply confirmation for the detailed error message

The test for the rollback path was updated: mock daemon no longer needs to reply (fire-and-forget enqueues then returns), polls for `stop_seen` with a 1s deadline to handle the asynchronous read by the daemon task.

## Validation

```
cargo test -p dora-coordinator                            73/73 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

Total tests: 70 (round 1) → 71 (round 2) → 73 (round 3).

## Process lesson

Round 2 introduced both bugs simultaneously:

- The `tokio::time::timeout` wrapper was added to fix the original cascade-failure HIGH finding. I didn't check `send_and_receive`'s cancellation safety before adding the wrapper — it would have failed a `tokio::test` with `loom` or even careful inspection of the cleanup flow.
- I also added the watchdog without considering that `Event::DataflowSpawnResult` could still arrive after the watchdog fires. Should have grep'd for all `running_dataflows.get_mut` writers and audited each for "what if spawn_result is already cached?"

Pattern bank entry: **when adding a new terminal-failure path, grep for all writers of the related state and audit each for the new failure mode.** Specifically for this fix: every `running_dataflows.get_mut(...)` site that touches `spawn_result` or its persistence is a potential interaction point with the watchdog.

PR head: `d371d817` → `79c2afa6`. Ready for re-review.

heyong4725:
## Adversarial review — round 3 (head `79c2afa6`)

Re-reviewed after the round-2 fixes (late-Ok guard + fire-and-forget rollback). The two findings from round 2 are properly closed. Found **one MEDIUM and one LOW** new concern from a fresh audit of all `running_dataflows.get_mut(...)` writers that touch `spawn_result`/`pending_spawn_results`.

## Round-2 fixes verified

✓ **Finding 1 (late-Ok resurrection)** — `handle_spawn_result_ok` extracts the guard into a named, doc'd function. `is_pending()` check runs before any persistence. Two tests directly exercise the helper (one verifies the guard catches resurrection; one verifies happy path still promotes Running). Extraction is the right structural choice — inline guards rot during refactors.

✓ **Finding 2 (cancellation-unsafe rollback)** — `fire_and_forget_rollback` uses `connection.send()` (mpsc enqueue, no `pending_replies` registration). Cancellation-safe by construction since there's no internal state to clean up. Also fixes the original cascade-failure issue at the source — no per-daemon 30s wait. Visibility on `run::rollback_spawned_daemons` correctly reverted to private.

## New Finding 1 (MEDIUM) — Auto-recovery doesn't respect `Cached(Err)` spawn_result

`binaries/coordinator/src/lib.rs:2225-2243` (auto-recovery loop on `DaemonStatusReport`):

```rust
for (uuid, df) in &mut running_dataflows {
    if !df.daemons.contains(&daemon_id) { continue; }
    if reported_set.contains(uuid) { ... continue; }
    if let Some(last) = df.last_recovery_attempt.get(&daemon_id)
        && now.duration_since(*last) < RECOVERY_BACKOFF { continue; }
    if df.pending_spawn_results.contains(&daemon_id) { continue; }
    // ... re-spawn nodes on this daemon
}
```

The loop checks `df.daemons` (alive set), `reported_set` (currently reporting), backoff timer, and `pending_spawn_results` (in-flight spawn). But **it does not check whether `spawn_result` is `Cached(Err)`** — i.e. whether the dataflow has been terminally failed.

Concrete failing scenario:

1. Multi-daemon dataflow starts; daemon X is one of the assigned daemons.
2. X stays connected (heartbeats OK) but its internal spawn flow is hung — never reports `DataflowSpawnResult`.
3. Watchdog fires at 60s. `spawn_result = Cached(Err)`, store persisted as `Failed`, `fire_and_forget_rollback` enqueues a Stop on the daemons that did report (none in this case; pending is just X). X is still in `df.daemons` AND in `df.pending_spawn_results`.
4. X eventually unsticks, reports `DataflowSpawnResult(Ok)`. `handle_spawn_result_ok` correctly sees `is_pending() == false` and observes-and-logs without persisting Running. X is removed from `pending_spawn_results`.
5. X later sends a `DaemonStatusReport` that **doesn't include** the dataflow (e.g. X's internal process restarted, lost track of the nodes it spawned).
6. Auto-recovery loop runs:
   - `df.daemons.contains(X)` → **true** (X never disconnected)
   - `reported_set.contains(uuid)` → **false** (X isn't reporting it)
   - `df.pending_spawn_results.contains(X)` → **false** (removed in step 4)
   - → auto-recovery proceeds and **re-spawns the nodes on X**

The dataflow is now running on X despite being marked `Failed` in the store and `Cached(Err)` in memory. State is contradictory.

Severity is MEDIUM not HIGH because:
- Requires step 5 (daemon "forgets" successfully-spawned nodes between SpawnResult and StatusReport) — uncommon but plausible (daemon process restart, internal cleanup bug, etc.)
- The "disconnect then reconnect" variant of this scenario is already safe: disconnect cleanup removes the daemon from `df.daemons`, so step 6's first check would skip.

**Fix:** add a `CachedResult` predicate and check it in auto-recovery:

```rust
// in state.rs, alongside is_pending():
pub(crate) fn is_terminal_error(&self) -> bool {
    matches!(self, CachedResult::Cached { result: Err(_) })
}

// in auto-recovery loop at lib.rs:2243:
if df.spawn_result.is_terminal_error() {
    continue;
}
```

This makes the watchdog's "terminally failed" state respected by the auto-recovery handler — exactly the same fix shape as round 2's Finding 1 (`handle_spawn_result_ok`). The principle: when a path introduces a terminal-failure state, every handler that might write to the affected state must respect it.

A test that exercises the auto-recovery loop would be involved (needs a DaemonStatusReport event with mock connections). A unit test for `is_terminal_error()` itself plus the existing late-Ok test already cover the predicate-level invariant.

## New Finding 2 (LOW) — Err arm of DataflowSpawnResult overwrites watchdog error

`binaries/coordinator/src/lib.rs:2144-2156`:

```rust
Err(err) => {
    tracing::warn!("error while spawning dataflow `{dataflow_id}`");
    if let Err(e) = dataflow
        .make_record(StoreDataflowStatus::Failed {
            error: format!("spawn failed: {err}"),
        })
        .and_then(|r| store.put_dataflow(&r))
    {
        tracing::warn!("failed to persist dataflow spawn failure: {e}");
    }
}
```

If the watchdog has already cached an Err for this dataflow and persisted a detailed Failed record ("spawn timed out after 60s; N daemon(s) never reported spawn_result; rolled back ..."), a late-arriving daemon `Err` overwrites the store record with `"spawn failed: <daemon-specific error>"`. The in-memory `spawn_result` is preserved (set_result on Cached is no-op), but the user reading `dora list` sees the less-informative error.

It also bumps `store_generation` unnecessarily.

Severity: LOW. Cosmetic / observability only — both error strings indicate the dataflow is Failed.

**Fix:** apply the same `is_pending()` guard to the Err arm. Extract a `handle_spawn_result_err` helper mirroring `handle_spawn_result_ok`:

```rust
fn handle_spawn_result_err(
    dataflow_id: DataflowId,
    daemon_id: &DaemonId,
    err: eyre::Report,
    store: &dyn CoordinatorStore,
) {
    if !dataflow.spawn_result.is_pending() {
        tracing::warn!(
            dataflow = %dataflow_id, daemon = %daemon_id,
            "ignoring late failed spawn_result on a dataflow already terminally failed: {err:?}",
        );
        return;
    }
    // ... existing Err arm logic
}
```

## What's good

- The round-2 fixes are structurally correct and the tests directly exercise the production helpers (no inline copies of the guard logic).
- `fire_and_forget_rollback` is the right structural fix — it solves cancellation safety AND cascade-failure simultaneously, with a cleaner API than the budget-wrapper.
- `handle_spawn_result_ok` is doc-commented with the load-bearing invariant; future readers will see the guard's purpose.
- The disconnect-path cross-link comment at `lib.rs:1862-1868` is preserved from round 2; the contract between disconnect and watchdog remains documented.
- Test names are descriptive of what they assert (e.g. `late_successful_spawn_result_does_not_resurrect_failed_dataflow`).
- 73/73 tests pass; clippy clean; fmt clean.

## Recommendation

Both findings are tractable to fix in this PR:

- **MEDIUM (auto-recovery)**: add `CachedResult::is_terminal_error()` predicate (~5 lines) + check in auto-recovery loop (~3 lines) + a unit test for the predicate (~10 lines). ~20-line total change.
- **LOW (Err arm overwrite)**: extract `handle_spawn_result_err` symmetric with the Ok helper (~15 lines including its guard) + replace inline block. ~20-line total change.

Both are mechanical and consistent with the round-2 fix style. ~30-40 lines total. Worth doing in this PR for symmetry and to keep the "terminal-failure paths are respected everywhere" invariant complete.

PR head: `79c2afa6`. Ready for re-review after these are addressed.

---

🤖 Adversarial self-review by Claude — round 4. Both findings emerged from systematically grep'ing for every writer of `spawn_result` and `pending_spawn_results` and asking "what if the watchdog already fired?" — pattern from round 2's Finding 1 applied exhaustively.

heyong4725:
Both findings fixed in `7c8b07ee`.

## Finding 1 (MEDIUM) — Auto-recovery guard

New predicate `CachedResult::is_terminal_error()` at `state.rs:430`:

```rust
pub(crate) fn is_terminal_error(&self) -> bool {
    matches!(self, CachedResult::Cached { result: Err(_) })
}
```

Guard added to the auto-recovery loop at `lib.rs:2244-2253`:

```rust
// Skip dataflows that the spawn-timeout watchdog (or any other
// terminal-failure path) has already marked as failed.
if df.spawn_result.is_terminal_error() {
    continue;
}
```

The predicate specifically matches `Cached(Err)` rather than `!is_pending()` because `Cached(Ok)` dataflows whose daemons dropped are exactly what auto-recovery is for — that legitimate path stays unblocked.

## Finding 2 (LOW) — Err arm helper extraction

New `handle_spawn_result_err` symmetric with `handle_spawn_result_ok`:

```rust
fn handle_spawn_result_err(
    dataflow_id: DataflowId,
    daemon_id: &DaemonId,
    err: eyre::Report,
    store: &dyn CoordinatorStore,
) {
    if !dataflow.spawn_result.is_pending() {
        tracing::warn!(
            dataflow = %dataflow_id, daemon = %daemon_id,
            "ignoring late failed spawn_result on a dataflow already \
             terminally failed: {err:?}",
        );
        return;
    }
    // ... existing persist + set_result logic
}
```

The inline Err arm in the event loop now just calls the helper. Same shape as the Ok extraction from round 2.

## Tests added

```
test tests::cached_result_is_terminal_error_distinguishes_states ... ok
test tests::late_failed_spawn_result_does_not_overwrite_watchdog_store_error ... ok
test tests::handle_spawn_result_err_persists_failed_when_pending ... ok
```

- **Predicate test**: locks in semantics — `Pending` and `Cached(Ok)` both return false; only `Cached(Err)` returns true. Catches a future refactor that broadens the predicate to `!is_pending()` (which would block legitimate auto-recovery).
- **Late-Err overwrite test**: seeds `spawn_result` with `Cached(Err)` and store with a detailed Failed record. Drives `handle_spawn_result_err` directly with a late daemon error. Asserts (a) store message preserved, (b) `generation` NOT bumped, (c) in-memory err preserved. The generation assertion specifically catches "the helper called `make_record` anyway" — a partial-fix where the guard skips set_result but not the persist.
- **Happy-path companion test**: verifies an over-eager guard doesn't block the legitimate failure path. Mirror of round-3's `handle_spawn_result_ok_promotes_to_running_when_pending_and_complete`.

I did not add a test for the auto-recovery guard itself because the auto-recovery loop is inline in the event loop (not extracted as a helper) and testing it would require synthesising a `DaemonStatusReport` event through the full event-loop machinery. The predicate is covered; the loop's `continue` is a single line whose behaviour is determined by the predicate. If anyone wants to add the integration test as follow-up, the `restore_topic_debug_streams_re_issues_start_after_reconnect` test (lib.rs:3992+) is the pattern to follow — it drives `DaemonStatusReport` end-to-end.

## Validation

```
cargo test -p dora-coordinator                            76/76 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

Test count: 70 (round 1) → 71 (round 2) → 73 (round 3) → 76 (round 4). Each round added regression coverage for the specific failure mode addressed.

## Process lesson

All three "terminal-failure state not respected" findings (round-2 Finding 1, round-4 Finding 1, round-4 Finding 2) share the same root cause: introducing a new terminal-failure path (the watchdog) without exhaustively auditing every handler that writes to the related in-memory or persisted state.

Pattern bank entry: **when a new code path can transition the system into a terminal-failure state, list every site that writes to the affected state — both in-memory state and persisted state — and add an explicit "is this state still mutable?" check at each.** For this PR specifically, the state in question was `spawn_result` (Pending → Cached transition), and the writers were: success arm of `DataflowSpawnResult`, error arm of `DataflowSpawnResult`, auto-recovery loop (writes by re-spawning, which feeds back through these handlers), `DataflowFinishedOnDaemon` (line 602; this one was already correct because it observes-and-logs without an unconditional persist). Round 4 closes the last two gaps.

PR head: `79c2afa6` → `7c8b07ee`. Ready for re-review.

heyong4725:
Both findings fixed in `62f0d377` — but via a different approach than the suggested per-consumer `is_terminal_error()` guards. After auditing the consumer surface, **9 sites** read `running_dataflows` (Check, List, reconcile, Clean, GetNodeInfo, RemoveNode, AddNode, WaitForSpawn, plus a few more). Adding guards in each is whack-a-mole and brittle — every future feature that reads `running_dataflows` would also need to remember the guard.

Took the **archive + remove** path instead: when the watchdog fires, mirror the existing `DataflowFinishedOnDaemon` teardown (lib.rs:504-604) so the spawn-timeout case looks identical to a normal completion from every consumer's point of view.

## Full teardown sequence

When the watchdog fires, the dataflow now goes through:

1. Set `spawn_result` to Err (existing)
2. Persist Failed in the store (existing)
3. **Send final log message** to subscribers
4. **Close topic subscribers** (clean end-of-stream, not hang)
5. **Drain `stop_reply_senders`** with `DataflowStopped` — otherwise in-flight `dora stop` calls hang forever (no `DataflowFinishedOnDaemon` arrives for a spawn-that-never-started)
6. **Archive** into `archived_dataflows` (capped at `MAX_ARCHIVED_DATAFLOWS`, lifted from a local const inside the existing teardown to module scope since both paths now share it)
7. **Remove from `running_dataflows`**

After step 7:

| Consumer | Pre-fix behaviour | Post-fix behaviour |
|---|---|---|
| `Check` (lib.rs:730) | Returns `DataflowSpawned` (lies) | Returns `DataflowStopped` (not in running_dataflows) |
| `List` (lib.rs:1004) | Labels Running | Excluded from Running entries |
| `DaemonStatusReport` reconcile (lib.rs:2175) | Promotes Failed → Running | Store-only path; in-memory absent ✓ |
| `Clean` (lib.rs:1068) | Skipped (still in running_dataflows) | Can be reaped (not in running_dataflows) |
| `RemoveNode` / `GetNodeInfo` / etc | Operates on stale entry | Returns "not found" |

The fix surface is now **one helper** (the watchdog teardown), not 9 separate guards. Every future reader of `running_dataflows` is naturally correct without remembering the guard.

## On the reconcile path

You correctly noted that reconcile reads from store via `store.get_dataflow()`, not from `running_dataflows`. So removing from in-memory doesn't directly stop reconcile from promoting Failed → Running.

However, the **scenario where reconcile fires** (a daemon reports the dataflow as active) is precisely the post-recovery case where the daemon actually IS running the dataflow despite the watchdog's verdict. In that case:

- Store update to Running reflects reality (the dataflow IS running on the daemon)
- In-memory `wait_for_spawn` already received Err (caller knows the original spawn timed out)
- The user can either accept the late-recovered dataflow as Running, or `dora stop` it

This is a legitimate "the daemon eventually came back" recovery. The disagreement between the persisted Failed-then-Running history and the user's earlier "spawn timed out" Err is observable but defensible — the user got the early Err because we couldn't confirm in time, then the daemon recovered, and the store now reflects the actual final state.

If we wanted to make this case even stricter (refuse to promote Failed → Running after watchdog has fired), we'd need a new store status (`TerminallyFailed` distinct from `Failed`) to distinguish "failed by daemon error" from "failed by coordinator-side watchdog". I think that's the right design but it's a wire-protocol change that belongs in a follow-up PR. Filing as a TODO.

## Tests

```
test tests::watchdog_makes_dataflow_invisible_to_check_and_list ... ok
test tests::watchdog_teardown_drains_stop_reply_senders ... ok
```

Plus three existing tests updated to pass `archived_dataflows` and to reflect the new "removed on watchdog fire" semantics. The `_idempotent_on_already_cached_result` test specifically asserts that **`Cached(Ok)`** dataflows stay in `running_dataflows` — only `Cached(Err)` triggers teardown. Catches a future bug where the watchdog over-eagerly tears down successfully-spawned dataflows.

Total: 70 (round 1) → 71 (round 2) → 73 (round 3) → 76 (round 4) → 78 (round 5).

## Validation

```
cargo test -p dora-coordinator                            78/78 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

## Pattern bank entry

When a code path transitions an entity to a terminal state, the cleanest fix is to **remove the entity from the in-memory collection** rather than mark it terminal and add guards in every reader. The reader count is usually higher than the writer count; the removal pattern scales naturally to future readers; the guard pattern requires remembering at each new read site.

The round-2/4 guards (`is_pending()` and `is_terminal_error()`) are still useful — they handle the case where another path (a daemon's `Err`) sets `Cached(Err)` without removing. That path is left as-is in this PR for scope reasons, but the same teardown approach should be applied there in a follow-up. Filing.

PR head: `7c8b07ee` → `62f0d377`. Ready for re-review.

heyong4725:
## Adversarial review — round 6 (head `62f0d377`)

The round-5 archive+remove approach correctly fixes both reviewer findings. Round-5's specific bugs (Finding 1: reconcile resurrection; Finding 2: Check/List inconsistency) are properly closed.

Re-audited the full consumer surface post-removal. Found **3 LOW/INFO concerns** about the *new visible behavior* of watchdog-failed dataflows. None are blockers — all involve secondary user-experience polish rather than correctness.

## Round-5 fix verified

✓ **In-memory teardown matches the existing `DataflowFinishedOnDaemon` shape**: archive into `archived_dataflows`, close subscribers, drain stop senders, persist Failed, remove from `running_dataflows`. The single-helper teardown closes 9 consumer sites in one move instead of 9 separate guards.

✓ **MAX_ARCHIVED_DATAFLOWS** correctly lifted to module scope and shared between the two teardown paths.

✓ **`stop_reply_senders` drain prevents stop-callers from hanging** — without this, any in-flight `dora stop` issued before the watchdog fired would wait forever for a `DataflowFinishedOnDaemon` event that will never arrive. Test `watchdog_teardown_drains_stop_reply_senders` locks this in.

✓ **Round-4 guards (handle_spawn_result_ok/err, auto-recovery) are NOT made dead code by round-5**. They still fire for the daemon-Err-first-then-daemon-Ok scenario (a different daemon reports Err which sets `Cached(Err)` without removing from running_dataflows; subsequent Ok from another daemon hits the helper guard). Defense-in-depth preserved.

## Finding 1 (LOW) — Watchdog-failed dataflows are invisible to `dora list`

After teardown, the dataflow:
- Is **not** in `running_dataflows` → not in List's "Running" entries
- Is **not** in `dataflow_results` (watchdog never inserts there) → not in List's "Finished/Failed" entries
- Is in `archived_dataflows` (name-only, for `dora restart` and `resolve_name`)
- Is in the store as `Failed`, but **List doesn't query the store**

Net: `dora list` returns 0 entries for a uuid the user JUST got a "spawn timed out" error for. The user has to remember the uuid; they can't list to find it again.

Compare to the normal `DataflowFinishedOnDaemon` failure path (`lib.rs:493-604`): that path BOTH archives AND inserts into `dataflow_results` (line 505-508), so List shows the dataflow as Failed.

**Fix:** synthesize a `dataflow_results` entry from the watchdog teardown. For each daemon in the original `df.daemons` set, insert a per-daemon `DataflowDaemonResult` with `node_results` containing one synthetic per-node `Err(NodeError)` so the list-classification check (`results.values().all(|r| r.is_ok())` at lib.rs:1019) returns false and the dataflow is correctly labelled `Failed`.

`NodeErrorCause::FailedToSpawn` (libraries/message/src/common.rs:128, has a variant matching this case) is the natural choice — its `Display` impl reads `"failed to spawn node: <err>"`, which matches the watchdog scenario semantically.

~15 lines, mirrors the existing inserter at lib.rs:505-508. Worth doing for UX parity.

## Finding 2 (LOW/INFO) — Late `DataflowFinishedOnDaemon` for removed dataflow is discarded

If a wedged daemon eventually completes (sends `DataflowFinishedOnDaemon`) for a watchdog-removed dataflow:

```rust
match running_dataflows.entry(uuid) {
    Entry::Occupied(mut entry) => { /* normal teardown */ }
    Entry::Vacant(_) => {
        tracing::warn!("dataflow not running on DataflowFinishedOnDaemon");
        // result is silently dropped
    }
}
```

The `result: DataflowDaemonResult` carries the daemon's actual per-node completion info — which nodes ran, which errored, what their exit codes were. If the daemon DID eventually start the nodes and they DID eventually run, that's potentially useful post-mortem.

After my change, that info goes to a `tracing::warn!` and is lost.

**Fix (if Finding 1 is fixed)**: in the Vacant arm, check if the uuid is in `archived_dataflows` (i.e. was watchdog-failed). If so, MERGE the daemon's result into the synthetic `dataflow_results` entry so post-mortem queries show the late-completion details. This requires Finding 1's synthesis path to exist first.

If Finding 1 isn't addressed, this becomes moot (no synthetic entry to merge into).

**Severity LOW** because: the watchdog already determined the dataflow Failed and notified the user; late per-daemon completion details are post-mortem curiosity, not actionable. Not a blocker.

## Finding 3 (LOW/INFO) — `dora stop <uuid>` for watchdog-failed dataflow returns "no known running dataflow"

```rust
ControlRequest::Stop { dataflow_uuid, ... } => {
    if let Some(result) = dataflow_results.get(&dataflow_uuid) {
        // return cached DataflowStopped
        continue;
    }
    let dataflow = stop_dataflow(&mut running_dataflows, dataflow_uuid, ...).await;
    // stop_dataflow bails if not in running_dataflows
}
```

After watchdog removal: not in `dataflow_results`, not in `running_dataflows` → `stop_dataflow` errors with `"no known running dataflow found with UUID `{uuid}`"`. The user who got "spawn timed out" then tries to tidy up with `dora stop <uuid>` gets a confusing "doesn't exist" error.

**Fix (if Finding 1 is fixed)**: the synthetic `dataflow_results` entry would make the early-return at line 824 fire, returning a clean `DataflowStopped` reply.

**Severity LOW** because: the dataflow doesn't actually need stopping (it never started). The error message is awkward but doesn't cause harm.

## What's good

- **Round-5 fix is structurally correct**: archive+remove eliminates the entire "watchdog state vs. in-memory state" inconsistency class with a single helper. The pattern bank entry ("remove from in-memory collection, not mark+guard") is genuinely the right architectural choice here.
- **The teardown sequence mirrors the existing `DataflowFinishedOnDaemon` path exactly** — same archive, same subscriber close, same `MAX_ARCHIVED_DATAFLOWS` cap (now shared). Future readers see two callers of the same pattern and can verify symmetry.
- **`watchdog_teardown_drains_stop_reply_senders` is the most useful new test**: catches a silent regression (stop-callers hanging) that would only manifest in production-y scenarios with concurrent `dora stop` issued before the watchdog fired.
- **The `_idempotent_on_already_cached_result` assertion was correctly tightened** to "Cached(Ok) dataflow MUST stay present" — explicitly guards against over-eager teardown that would remove successfully-spawned dataflows. Round-5 could plausibly have introduced this regression; the test catches it.
- **Defense-in-depth from round 4 is preserved**: handle_spawn_result_ok/err guards and auto-recovery `is_terminal_error()` check still fire for the daemon-Err-first-then-daemon-Ok scenario.

## Recommendation

All three findings are **LOW severity** and **closely related** — fixing Finding 1 (synthesize `dataflow_results` entry) automatically resolves Findings 2 and 3.

If fixing: one ~30-line change. The synthesis loop iterates `df.daemons`, builds a per-daemon `DataflowDaemonResult` with synthetic `FailedToSpawn` node_results, inserts into `dataflow_results`. Plus update Stop's Vacant arm to merge if archived. Plus a test that asserts `dora list` shows the dataflow as Failed.

If not fixing in this PR: the watchdog-failed UX is mildly inconsistent with the normal-failure UX, but no correctness issue. The reviewer's two HIGH/MEDIUM findings are properly closed; the additional findings are polish.

PR head: `62f0d377`. Ready for either landing-as-is or addressing the LOW findings.

---

🤖 Adversarial self-review by Claude — round 6. Findings emerged from auditing what `dora list`, `dora stop`, and late `DataflowFinishedOnDaemon` look like post-watchdog. The user-visible behaviour after watchdog fire is now consistent (every consumer treats the dataflow as gone) but inconsistent with the *normal failure* UX (where the dataflow shows up as Failed in `dora list`).

heyong4725:
All three findings fixed in `92123988`. They share a single root cause (watchdog didn't populate `dataflow_results`), so the fix is one synthesis path + one merge.

## Finding 1 — Synthesize `dataflow_results` entry

Watchdog teardown now builds a `BTreeMap<DaemonId, DataflowDaemonResult>` for the dataflow before draining stop senders. Each per-daemon entry has one `Err(NodeError { cause: FailedToSpawn(watchdog_msg) })` per node assigned to that daemon:

```rust
let synth_results: BTreeMap<DaemonId, DataflowDaemonResult> = df
    .daemons
    .iter()
    .map(|daemon_id| {
        let nodes_for_daemon: BTreeMap<NodeId, Result<(), NodeError>> = df
            .node_to_daemon
            .iter()
            .filter(|(_, did)| *did == daemon_id)
            .map(|(node_id, _)| {
                (node_id.clone(), Err(NodeError {
                    timestamp: synth_timestamp,
                    cause: NodeErrorCause::FailedToSpawn(err_msg.clone()),
                    exit_status: NodeExitStatus::Unknown,
                }))
            })
            .collect();
        (daemon_id.clone(), DataflowDaemonResult { timestamp: synth_timestamp, node_results: nodes_for_daemon })
    })
    .collect();
dataflow_results.entry(uuid).or_default().extend(synth_results);
```

`results.values().all(is_ok)` returns false (every entry has at least one Err), so List correctly classifies the dataflow as `Failed`.

The `stop_reply` sent to drained `stop_reply_senders` now uses `dataflow_result(...)` over the synthesized entry, so in-flight `dora stop` callers see the FailedToSpawn details rather than `ok_empty`.

## Finding 2 — Merge late `DataflowFinishedOnDaemon` into synthetic entry

`DataflowFinishedOnDaemon` Vacant arm now checks whether the uuid is in `archived_dataflows` (i.e. was previously watchdog-failed). If yes, merge the daemon's `result` into `dataflow_results[uuid]`:

```rust
if archived_dataflows.contains_key(&uuid) {
    let entry = dataflow_results.entry(uuid).or_default();
    let existing = entry.entry(daemon_id.clone()).or_insert_with(|| {
        DataflowDaemonResult {
            timestamp: result.timestamp,
            node_results: BTreeMap::new(),
        }
    });
    existing.timestamp = result.timestamp;
    existing.node_results.extend(result.node_results);
} else {
    tracing::warn!("dataflow not running on DataflowFinishedOnDaemon");
}
```

`BTreeMap::extend` overwrites keys, so the daemon's real per-node results replace the synthetic `FailedToSpawn` placeholders where they overlap — best available post-mortem info instead of silently discarded.

The "true Vacant" case (unknown uuid) still gets the existing warning behaviour.

## Finding 3 — Auto-resolved by Finding 1

Once `dataflow_results.get(&uuid)` returns `Some` for a watchdog-failed uuid, the Stop handler's early-return at lib.rs:824 fires:

```rust
if let Some(result) = dataflow_results.get(&dataflow_uuid) {
    let reply = ControlRequestReply::DataflowStopped { uuid, result: dataflow_result(...) };
    let _ = reply_sender.send(Ok(reply));
    continue;
}
```

User gets a clean `DataflowStopped` reply with the FailedToSpawn details, instead of `stop_dataflow`'s "no known running dataflow" bail.

## Tests added (3 new, 81/81 total)

```
test tests::watchdog_synthesizes_dataflow_results_for_list_visibility ... ok
test tests::dora_stop_after_watchdog_finds_dataflow_results_entry ... ok
test tests::late_finished_on_daemon_merge_extends_node_results ... ok
```

- **`watchdog_synthesizes_dataflow_results_for_list_visibility`**: asserts the entry exists, classifies as Failed (mirrors List's check), and carries `FailedToSpawn` per-node errors naming the timeout reason. Validates Finding 1's contract end-to-end.
- **`dora_stop_after_watchdog_finds_dataflow_results_entry`**: asserts the Stop handler's early-return condition is met (`dataflow_results.contains_key(uuid)`). Locks in Finding 3.
- **`late_finished_on_daemon_merge_extends_node_results`**: unit test on the merge math directly. Verifies that extending a synthetic BTreeMap with a daemon's real results correctly overwrites the placeholder FailedToSpawn entries with the real per-node Ok/Err details.

Plus existing tests' plumbing updated to pass the new `dataflow_results` argument to `check_spawn_timeouts`.

## Round 1→6 progression

| Round | Findings | Fix shape |
|---|---|---|
| 1 | spawn timeout watchdog | rollback + persist Failed |
| 2 | rollback budget, cancellation | bounded wrapper |
| 3 | wrapper was cancellation-unsafe; late-Ok resurrection | fire-and-forget + `is_pending()` guard |
| 4 | Err arm overwrites, auto-recovery resurrects | `handle_spawn_result_err`, `is_terminal_error()` guard |
| 5 | reconcile, Check/List inconsistency | archive+remove (one helper > 9 guards) |
| 6 | UX parity (list/stop visibility) | synthesize `dataflow_results` entry |

Watchdog failure is now functionally indistinguishable from normal failure from every consumer's point of view — same archive path, same `dataflow_results` entry shape, same List/Check/Stop behaviour. The implementation mirrors `DataflowFinishedOnDaemon`'s teardown exactly.

## Validation

```
cargo test -p dora-coordinator                            81/81 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

PR head: `62f0d377` → `92123988`. Ready for re-review.

heyong4725:
Both findings fixed in `a990f618`. Both real and load-bearing.

## Finding 1 (HIGH) — Reconcile guard against archived dataflows

Added the guard at the top of the reconcile loop:

```rust
for entry in &reported_dataflows {
    let df_id = &entry.dataflow_id;
    if archived_dataflows.contains_key(df_id) {
        tracing::warn!(
            "daemon {daemon_id} reports archived (terminally-failed) \
             dataflow {df_id} as running; skipping reconcile so the \
             watchdog's Failed verdict is preserved"
        );
        continue;
    }
    match store.get_dataflow(df_id) { ... }
}
```

Archived (= watchdog-failed in current code, plus any future archive-on-failure path) dataflows are terminally failed and must never have their store status promoted to Running. The warning log surfaces "a daemon is reporting an archived dataflow" so operators investigating unexpected state see exactly what's happening.

**Crucial: the post-coordinator-restart case is unaffected.** On restart, `archived_dataflows` starts empty (it's purely in-memory), so the guard doesn't trip and the legitimate `Recovering -> Running` promotion fires as designed. The guard ONLY filters out dataflows the current coordinator process has archived.

## Finding 2 (MEDIUM) — Empty result map vacuously classifies as Finished

You're exactly right. The previous synthesis iterated `df.daemons`, and the disconnect cleanup at `lib.rs:1893-1899` clears that set. After:

```rust
for (_uuid, df) in running_dataflows.iter_mut() {
    for machine_id in &disconnected {
        df.daemons.remove(machine_id);
        df.pending_daemons.remove(machine_id);
        df.pending_spawn_results.remove(machine_id);
    }
}
```

`df.daemons` could be empty. Watchdog synthesis iterated it, produced `BTreeMap::new()`. List then checked `results.values().all(is_ok)` on the empty map — vacuously true — classified as `Finished`. The exact case the comment claims is handled would silently show as Finished in `dora list`.

Fix: synthesize from `df.node_to_daemon.values().collect()` instead:

```rust
let assigned_daemons: BTreeSet<DaemonId> =
    df.node_to_daemon.values().cloned().collect();
let synth_results: BTreeMap<DaemonId, DataflowDaemonResult> = assigned_daemons
    .iter()
    .map(|daemon_id| { ... })
    .collect();
```

`node_to_daemon` is set at construction (`handlers.rs:610`) and **never cleared** by the disconnect cleanup (which only touches `df.daemons`, `df.pending_daemons`, `df.pending_spawn_results`). It preserves the original assignment even when every daemon has disconnected — the right source of truth for "what daemons should have been running this".

**Defence in depth**: if `node_to_daemon` itself is somehow empty (defensive against future refactors), inject a sentinel entry under a synthetic `<watchdog>` daemon id with a single `FailedToSpawn` node result. The result map is never empty.

## Tests

```
test tests::archived_dataflows_predicate_distinguishes_watchdog_failed_from_unknown ... ok
test tests::watchdog_disconnect_mid_spawn_still_classifies_as_failed ... ok
```

The second test is the load-bearing one: simulates the post-disconnect-cleanup state (`df.daemons.clear()` while `node_to_daemon` retains the assignment), runs the watchdog, asserts (a) synthesis non-empty, (b) classifies as Failed (not vacuously Finished), (c) uses the original assignment from `node_to_daemon` not the empty `df.daemons`. This is the EXACT case the watchdog comment claims to handle — now actually backed by a test.

The first test is predicate-level — driving the full `DaemonStatusReport` event end-to-end would require synthesizing registered daemon connections + reported_dataflows entries (see the existing `restore_topic_debug_streams_re_issues_...` test pattern, ~100 lines of setup). Predicate-level test plus the archive-invariant test from round 5 is sufficient to lock the behaviour; integration test could be added as a follow-up if needed.

## Total round-7 stats

| Round | Tests added | Cumulative | Top severity |
|---|---|---|---|
| 1 | 4 | 70 | — |
| 2 | 1 | 71 | HIGH (cascade) |
| 3 | 2 | 73 | HIGH (late-Ok) |
| 4 | 3 | 76 | MEDIUM |
| 5 | 2 | 78 | HIGH (reconcile resurrection) |
| 6 | 3 | 81 | LOW × 3 |
| 7 | 2 | 83 | HIGH + MEDIUM |

```
cargo test -p dora-coordinator                            83/83 passing
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

## Pattern bank entry (final form)

When introducing a terminal-failure path, two invariants are independently load-bearing:

1. **In-memory state must be terminal** (round 5 — remove + archive)
2. **Persisted state must be respected by every reader** (round 7 — reconcile guard against archive)

Round 5 closed the first; round 7 closes the second. Both required because some readers go through in-memory state (Check, List) and some go through the persisted store (reconcile). A terminal state needs to be terminal in BOTH layers.

The empty-result-map gap (Finding 2) is the same root cause as round 5 just at a different level: the synthesis needs to use a source-of-truth that the disconnect cleanup doesn't mutate. `node_to_daemon` is that source; `df.daemons` is not.

PR head: `92123988` → `a990f618`. Ready for re-review.

heyong4725:
## Adversarial review — round 8 (head `a990f618`)

Round-7 fixes correctly close the two findings. Re-audited the full state-mutation surface and found **one HIGH** (related to round-7's Finding 1 but exposed after coordinator restart) and **one MEDIUM** (panic in dead-code defensive branch).

## Round-7 fixes verified

✓ **Reconcile guard via `archived_dataflows.contains_key(df_id)`**: correctly skips watchdog-failed dataflows in the same coordinator instance. Logging surfaces the skip for operator debugging.

✓ **Synthesis source switched to `df.node_to_daemon.values()`**: correctly survives the disconnect-mid-spawn case where `df.daemons` is empty. Test `watchdog_disconnect_mid_spawn_still_classifies_as_failed` directly exercises this — asserts (a) result non-empty, (b) classifies as Failed.

✓ **Defensive sentinel for empty `node_to_daemon`**: structurally correct in keeping the result-map non-empty. (But see Finding 2 below — the sentinel itself has a panic bug.)

## Finding 1 (HIGH) — Cross-restart reconcile resurrection

The round-7 reconcile guard checks `archived_dataflows.contains_key(df_id)`, but **`archived_dataflows` is in-memory only**. After a coordinator restart:

- Store has the dataflow as `Failed { error: "spawn timed out ..." }` (watchdog persisted)
- `archived_dataflows` is empty (initialized empty at lib.rs:303)
- Startup code at `lib.rs:357` correctly leaves `Failed | Succeeded` records as-is (no transition to Recovering)
- Wedged daemon eventually reconnects, reports the dataflow as running
- Reconcile loop:
  - `archived_dataflows.contains_key(df_id)` → **false** (wiped on restart)
  - Falls through to `store.get_dataflow()` → `Ok(Some(record))` with `Failed { .. }`
  - Match arm `Failed { .. } | Pending | Recovering` → **promotes to Running**
- Store now says Running, contradicting the watchdog's terminal Failed.

The user observed the spawn timeout via `wait_for_spawn` *before the restart*. After restart + daemon report, `dora list` would show the dataflow as Running. That's exactly the round-5 / round-7 invariant the in-memory guard was supposed to enforce — broken across restarts.

**Fix options:**

(a) **Tighten reconcile**: only promote `Pending | Recovering` (drop `Failed`). Failed records are treated as terminal regardless of daemon reports.

(b) **New store variant**: `StoreDataflowStatus::TerminallyFailed` distinct from `Failed`. Wire-protocol change.

(c) **Persist a magic marker in `Failed.error`**: ugly, brittle.

I'd go with **(a)**. The "daemon overrides coordinator-side Failed" path was permissive but rarely-justified:

- Multi-daemon partial failure case (daemon A reported Err, daemon B still running) is preserved correctly: the dataflow stays in `running_dataflows`, the reconcile path eventually re-promotes to Running... wait, no, under (a) it wouldn't promote anymore. Let me re-think.

Actually checking the multi-daemon scenario again: when daemon A reports `Err`, `handle_spawn_result_err` runs `set_result(Err)` + persists Failed but **does NOT remove from `running_dataflows`** (round-4 design). So daemon B's continued reports keep going through the reconcile path with `record.status == Failed`. Under option (a), reconcile would never promote, leaving the dataflow as Failed in the store even while daemon B is reporting it running. The user would see Failed in `dora list` but daemon B's nodes are actually running.

This is awkward — the multi-daemon partial-failure scenario currently relies on the reconcile path to "correct" the store back to Running. Under (a), the store would say Failed even though some daemons are still running.

Hmm. So option (a) isn't a clean win. The current behaviour (permissive promotion) was load-bearing for that scenario.

**Better fix**: extend the in-memory archive-marker idea across restarts. Could be done via the store itself:

(d) **Two-state Failed**: instead of changing the variant, change the startup behaviour. On startup, transition any `Failed { .. }` whose error message contains a watchdog-marker (e.g. prefix `[watchdog]`) into a separate in-memory set. The reconcile guard then checks that set too. Or: change `Failed` to carry a `terminal: bool` field — schema change but compatible.

(e) **Round-7 fix only protects same-coordinator; documented limitation**: leave cross-restart resurrection as an acknowledged limitation. Document in the watchdog code comment.

For a real fix, I'd recommend **(d) with a `terminal: bool` field** on `StoreDataflowStatus::Failed`. Wire-protocol change is small (additive). Watchdog persists `Failed { error, terminal: true }`. Reconcile checks the flag.

For this PR's scope, **(e) is the pragmatic choice** — acknowledge the cross-restart limitation in the watchdog code comment + file a follow-up issue. The user's primary problem (wait_for_spawn hangs) is fixed; cross-restart resurrection is a narrow second-order issue.

## Finding 2 (MEDIUM) — Defensive sentinel panics on NodeId::from invalid string

`binaries/coordinator/src/lib.rs:2840`:

```rust
node_results.insert(
    NodeId::from("<watchdog>".to_string()),
    Err(NodeError { ... }),
);
```

`validate_node_id` at `libraries/message/src/id.rs:10-23` rejects characters that aren't `[a-zA-Z0-9_.-]`. The `<` and `>` chars are invalid. `From<String> for NodeId` **panics** on validation failure (line 104 of id.rs):

```rust
impl From<String> for NodeId {
    fn from(id: String) -> Self {
        if let Err(e) = validate_node_id(&id) {
            panic!("invalid NodeId '{id}': {e}");
        }
        Self(id)
    }
}
```

The sentinel is dead code under normal operation (the comment says "shouldn't happen via the normal construction path, but be defensive against future refactors"). But if any future refactor makes `node_to_daemon` empty for a stuck dataflow, this branch fires and the coordinator panics. Same root cause as the `output_id.into()` panic from PR #1833 — `From<String>` impls on validated types are unsafe for caller-controlled inputs.

**Fix:** use a valid NodeId. `"watchdog"` (no angle brackets) parses cleanly:

```rust
node_results.insert(
    NodeId::from("watchdog".to_string()),
    Err(NodeError { ... }),
);
```

Or, more defensively, use `.parse()` and `unwrap_or_else` with a safe fallback in case validation rules tighten in the future. But `"watchdog"` is unambiguously valid.

**Test gap**: the sentinel branch isn't exercised by any test, so this panic-on-typo bug wasn't caught. A test that constructs a RunningDataflow with empty `node_to_daemon` (instead of just empty `df.daemons`) would have caught it.

## What's good

- Round-7's `archived_dataflows.contains_key(df_id)` guard correctly closes the same-coordinator reconcile resurrection. Cross-restart is a separate (harder) problem.
- The disconnect-mid-spawn synthesis via `node_to_daemon.values()` is the right source-of-truth choice. `node_to_daemon` is set at construction and untouched by the disconnect cleanup.
- `watchdog_disconnect_mid_spawn_still_classifies_as_failed` is a load-bearing test — exactly the scenario the original PR design comment claims to handle.
- The `archived_dataflows_predicate_distinguishes_watchdog_failed_from_unknown` test correctly locks in the guard's predicate behaviour without requiring full event-loop integration.

## Recommendation

- **Finding 1 (HIGH cross-restart)**: For this PR, document the limitation in the watchdog code comment + file a follow-up for the proper fix (add `terminal: bool` to `StoreDataflowStatus::Failed` or equivalent persistent terminal-marker). The fix is non-trivial — touches wire protocol — and the primary user-facing bug (wait_for_spawn hangs forever) is already fixed.

- **Finding 2 (MEDIUM panic)**: Quick fix — change `"<watchdog>"` to `"watchdog"` in the sentinel, plus add a test that exercises the empty-`node_to_daemon` branch. 5-10 lines.

If you'd rather close BOTH findings in this PR (full fix for cross-restart): adding a `terminal: bool` field to `StoreDataflowStatus::Failed` is the cleanest path. Backward compatible if we default the field on deserialize. ~30-50 lines including persistence + reconcile guard + tests.

PR head: `a990f618`. Both findings are tractable; recommend at least fixing Finding 2 (panic) since it's mechanical.

---

🤖 Adversarial self-review by Claude — round 8. Findings emerged from auditing what `archived_dataflows` looks like across a coordinator restart (Finding 1) and what `NodeId::from("<watchdog>")` actually does at runtime (Finding 2).

heyong4725:
Both fixed in `2972d716`.

## Finding 1 (HIGH) — Cross-restart `terminal` marker

Added `#[serde(default)] terminal: bool` to `DataflowStatus::Failed`:

```rust
Failed {
    error: String,
    /// Marks the failure as terminal -- subsequent daemon reports
    /// must NOT promote the record back to `Running` via the
    /// reconcile path. ... `#[serde(default)]` so persisted records
    /// written by older coordinators (which never set this field)
    /// deserialize with `terminal: false`.
    #[serde(default)]
    terminal: bool,
},
```

The reconcile match arm now splits Failed into two branches:

```rust
StoreDataflowStatus::Failed { terminal: false, .. }
| StoreDataflowStatus::Pending
| StoreDataflowStatus::Recovering => {
    // Promote to Running (preserves multi-daemon partial-failure
    // case + legacy semantics for older persisted records)
}
StoreDataflowStatus::Failed { terminal: true, .. } => {
    tracing::warn!("daemon reports terminally-failed dataflow as running; skipping reconcile");
}
```

The `terminal: true` marker is set by:

| Site | Rationale |
|---|---|
| Watchdog teardown | Terminal verdict already delivered to user via wait_for_spawn |
| `handle_spawn_result_err` | Daemon authoritatively reported spawn failed |
| Recovery timeout (Recovering → Failed) | Coordinator gave up; daemon comeback shouldn't undo |

The `terminal: false` marker is preserved for the **multi-daemon partial-failure** case: when one daemon reports Err but another is still running, `Failed { terminal: false }` lets the running daemon's reconcile re-promote to Running. This was the legitimate semantics that "drop Failed from the match arm" (one of the round-7 review's suggested fixes) would have broken.

**Backward compatibility**: `#[serde(default)]` makes the field default to `false` when deserializing old records. Test `dataflow_status_failed_terminal_field_defaults_to_false_on_legacy_records` locks this in by deserializing a legacy `{"Failed": {"error": "..."}}` JSON literal and asserting `terminal == false`. Older coordinators' persisted Failed records keep their pre-#1854 behaviour.

## Finding 2 (MEDIUM) — Sentinel panic

Changed sentinel `NodeId` from `"<watchdog>"` to `"watchdog"` — the angle brackets fail `validate_node_id`'s `[a-zA-Z0-9_.-]` allowlist and panic via `From<String> for NodeId`. The DaemonId machine_id was changed too for consistency (DaemonId doesn't validate, but keeping the names aligned avoids confusion).

```rust
node_results.insert(
    NodeId::from("watchdog".to_string()),  // was "<watchdog>" — panicked
    Err(NodeError { ... }),
);
sentinel.insert(
    DaemonId::new(Some("watchdog".to_string())),
    DataflowDaemonResult { ... },
);
```

## Tests added (3 new, 86/86 total)

```
test tests::watchdog_persists_terminal_marker_for_cross_restart_protection ... ok
test tests::dataflow_status_failed_terminal_field_defaults_to_false_on_legacy_records ... ok
test tests::watchdog_sentinel_branch_does_not_panic_on_empty_node_to_daemon ... ok
```

The sentinel-no-panic test specifically forces the branch by clearing both `df.daemons` AND `df.node_to_daemon` — if a future refactor reintroduces the angle-bracket NodeId, the test fails via panic during `check_spawn_timeouts.await`. The cross-restart test asserts the watchdog writes `terminal: true` into the store, which is the load-bearing invariant.

Plus existing test patterns updated:
- `DataflowStatus::Failed { error }` → `Failed { error, .. }`
- Test helpers' `make_record(StoreDataflowStatus::Failed { error })` constructions adjusted to set `terminal: true` (they simulate watchdog persistence)

## Round 1→8 final stats

| Round | Tests | Top severity | Theme |
|---|---|---|---|
| 1 | 70 | — | initial watchdog + rollback |
| 2 | 71 | HIGH (cascade) | bounded budget |
| 3 | 73 | HIGH (late-Ok) | fire-and-forget + guard |
| 4 | 76 | MEDIUM (Err arm) | symmetric guards |
| 5 | 78 | HIGH (reconcile resurrection same-coord) | archive+remove |
| 6 | 81 | LOW × 3 (List/stop UX) | synthesize dataflow_results |
| 7 | 83 | HIGH (reconcile + empty-map) | predicate + node_to_daemon |
| 8 | 86 | HIGH (cross-restart) + MEDIUM (panic) | terminal field |

Watchdog-failure path is now functionally identical to normal-failure across all consumers AND across coordinator restarts. The `terminal` marker is the persistent invariant that ties the in-memory archive to the store.

## Validation

```
cargo test -p dora-coordinator                            86/86 passing
cargo build --all --exclude python crates                 clean
cargo clippy -p dora-coordinator --all-targets -D warnings   clean
cargo fmt --all -- --check                                    clean
```

PR head: `a990f618` → `2972d716`. Ready for re-review.

PR #1865, opened 2026-05-19
fix(node): mark write_events_to recording poisoned on first record_event failure (closes #1857)

### LINKED ISSUE #1857: DORA_WRITE_EVENTS_TO recorder produces incomplete-but-valid JSON when record_event fails (follow-up to #1855)
## Context

Surfaced by Codex adversarial review on #1855 (rescue of #1608). Recording here so it doesn't get lost — out of scope for the rescue, which faithfully implemented PR 1608's panel-faithful panic→warn conversion.

## The shape of the regression

After #1855, `apis/rust/node/src/event_stream/mod.rs::EventStream::add_event` no longer panics when `record_event` returns `Err`:

```rust
fn add_event(&mut self, event: EventItem) {
    if let Err(err) = self.record_event(&event) {
        tracing::warn!(
            node = %self.node_id,
            "failed to record event to write_events_to log: {err:?}"
        );
    }
    self.scheduler.add_event(event);
}
```

When `DORA_WRITE_EVENTS_TO` is set and `record_event` fails — Arrow-to-JSON conversion error, file handle revoked, disk full, etc. — the failing event is silently dropped from the recording, but `scheduler.add_event(event)` still runs (correctly — the event still has to be delivered to the node). At drop, `write_out()` still produces a syntactically valid `inputs-{node_id}.json` that is **missing those events** with no indication anything went wrong.

For replay/audit users, the file LOOKS valid and replays cleanly, but the trace is incomplete. The pre-#1855 `.unwrap()` would have crashed the node and produced no file at all — a louder, more correct signal for that use case.

## Why we deferred fixing this in #1855

PR #1855 was scoped to faithfully rescue @LeonRust's #1608, which proposed a 1:1 `.unwrap()` → `warn!` conversion. A proper recorder fix requires new state and policy decisions beyond what the original PR proposed.

## Probability

Conditional on `DORA_WRITE_EVENTS_TO` being set (opt-in observability), which most users don't enable. But for users who DO enable it (replay/debug/audit workflows), this is exactly the corruption mode they need to detect.

## Proposed fix

Add a "poisoned" state to the recorder. On the first `record_event` failure when `write_events_to` is set, set the poisoned flag. At drop, either:

1. Suppress the JSON file entirely (replay tools see "no recording" rather than "incomplete recording"), or
2. Write the JSON with an explicit `{"recording_status": "poisoned", "first_failure": "..."}` marker so replay tools can refuse to load.

Option 2 is friendlier for partial-replay use cases; option 1 is simpler and impossible to misuse.

A minimal interim fix is to add a top-level `errors: [...]` array to the output JSON that downstream consumers can check.

## References

- PR #1855 (the rescue) — the regression-shaped change is in `apis/rust/node/src/event_stream/mod.rs::add_event`.
- `record_event` body at `apis/rust/node/src/event_stream/mod.rs:583+` shows the failure modes (Arrow conversion, file I/O, JSON serialization).
- Original PR #1608 (the source of the rescue intent).

heyong4725:
## Summary

Closes #1857. Surfaces partial `DORA_WRITE_EVENTS_TO` recordings to consumers so silent truncation can't masquerade as a clean trace.

## The problem

After #1855, `EventStream::add_event` no longer panics on `record_event` failure — it warns and continues. That's correct for the event loop (we don't want observability to kill the node), but it created a new failure mode: failing events get silently dropped from `events_buffer`, subsequent events keep landing, and at drop `write_out()` emits a syntactically-valid `inputs-{node_id}.json` missing events with **no signal that anything went wrong**.

For replay / audit pipelines, "appears valid but is incomplete" is worse than "loud crash" — the file looks loadable and decisions get made on bad data.

## The fix

Three additions, all in `apis/rust/node/src/event_stream/mod.rs`:

1. **`WriteEventsTo` gains a `poisoned: Option<PoisonInfo>` field.**
2. **`EventStream::add_event` calls `write_events_to.mark_poisoned(&err, time_offset_secs)`** on the first `record_event` Err. Subsequent failures bump `additional_failures` without overwriting the first.
3. **`write_out()` emits a top-level `recording_status` field unconditionally** — `{"state": "clean"}` for healthy recordings, otherwise:
   ```json
   {
     "state": "poisoned",
     "first_missed_index": <events_buffer.len() at first failure>,
     "first_failure_time_offset_secs": <secs since EventStream start>,
     "first_failure_error": "<format!(\"{:?}\")>",
     "additional_failures": <count after the first>
   }
   ```

Consumers (replay tools, audit pipelines) can now:
- Refuse to load when `state == "poisoned"`, or
- Opt-in load partial data with explicit awareness of where the gap begins

## Backwards compatibility

`IntegrationTestInput` (`libraries/message/src/integration_testing_format.rs:23`) is the canonical consumer of this JSON format. It does NOT use `#[serde(deny_unknown_fields)]`, so the new top-level field is **silently ignored** by existing deserializers. The `events` array shape is unchanged. Anyone not looking at `recording_status` keeps working.

## Tests

3 new unit tests under `event_stream::tests::`, no new dev-deps:

| Test | What it asserts |
|---|---|
| `write_events_clean_recording_emits_state_clean` | Healthy path → `recording_status.state == "clean"` |
| `write_events_poisoned_recording_emits_state_poisoned_with_first_failure` | Push 2 events, mark poisoned at time 1.5s, push 1 more successful event → state == "poisoned", first_missed_index == 2, error contains expected substring, additional_failures == 0, events.len() == 3 |
| `write_events_multiple_failures_keep_first_and_count_rest` | 3 sequential `mark_poisoned` calls → only the first error is captured; additional_failures == 2 |

```
test event_stream::tests::write_events_clean_recording_emits_state_clean ... ok
test event_stream::tests::write_events_multiple_failures_keep_first_and_count_rest ... ok
test event_stream::tests::write_events_poisoned_recording_emits_state_poisoned_with_first_failure ... ok

test result: ok. 82 passed; 0 failed; 0 ignored; 0 measured
```

(82 = previous 79 + my 3 new.)

## Test plan

- [x] `cargo clippy --all --exclude dora-{node-api,operator-api,ros2-bridge}-python -- -D warnings`
- [x] `cargo test -p dora-node-api` (82 pass)
- [x] `cargo check --examples`
- [ ] CI green
- [ ] Tonight's nightly stays green

## Related

- Closes #1857. Same "soft success masks failure" shape as #1864 (cli-tests soft-PASS, just merged); different subsystem, same lesson.
- Stacks behind #1855 (the rescue of #1608 that introduced the panic→warn that created this gap).

## Risk

Pure addition. Production code change is small (~50 LOC); the rest is tests. Only meaningful behavioral change is at write_out time and is gated on whether `record_event` ever failed — clean runs are identical to before except for the new `recording_status: { state: "clean" }` line in the JSON.

### DIFF
--- a/apis/rust/node/src/daemon_connection/node_integration_testing.rs
+++ b/apis/rust/node/src/daemon_connection/node_integration_testing.rs
@@ -16,7 +16,7 @@ use dora_message::{
     common::{DataMessage, Timestamped},
     daemon_to_node::{DaemonReply, NodeEvent},
     integration_testing_format::{
-        IncomingEvent, InputData, IntegrationTestInput, TimedIncomingEvent,
+        IncomingEvent, InputData, IntegrationTestInput, RecordingStatus, TimedIncomingEvent,
     },
     metadata::{ArrowTypeInfo, Metadata},
     node_to_daemon::DaemonRequest,
@@ -52,6 +52,30 @@ impl IntegrationTestingEvents {
             .with_context(|| format!("failed to deserialize {}", input_file_path.display()))?,
             TestingInput::Input(input) => input,
         };
+
+        // Refuse to replay poisoned recordings. The `events` array is
+        // known incomplete relative to the original run; loading it
+        // anyway would defeat the whole point of recording (#1857).
+        // Hand-authored fixtures and pre-#1857 recordings have
+        // `recording_status: None` and pass through cleanly.
+        if let Some(boxed) = &node_info.recording_status
+            && let RecordingStatus::Poisoned {
+                first_failure_event_index,
+                first_failure_time_offset_secs,
+                first_failure_error,
+                additional_failures,
+            } = boxed.as_ref()
+        {
+            eyre::bail!(
+                "refusing to replay poisoned recording for node `{node_id}`: \
+                 the original recorder failed at event index {first_failure_event_index} \
+                 (~{first_failure_time_offset_secs:.3}s into the run), then \
+                 {additional_failures} additional event(s) also failed to record. \
+                 The `events` array is incomplete and replay will not reproduce \
+                 the original behavior. First failure: {first_failure_error}",
+                node_id = node_info.id,
+            );
+        }
 
         let output_writer = match output {
             TestingOutput::ToFile(output_file_path) => {
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -213,6 +213,7 @@ impl EventStream {
                     node_id: node_id.clone(),
                     file,
                     events_buffer: Vec::new(),
+                    poisoned: None,
                 })
             }
             None => None,
@@ -576,6 +577,17 @@ impl EventStream {
                 node = %self.node_id,
                 "failed to record event to write_events_to log: {err:?}"
             );
+            // Mark the recording poisoned so consumers can detect events
+            // are missing from the final JSON. `write_out()` surfaces this
+            // as a top-level `recording_status` field (#1857).
+            if let Some(write_events_to) = self.write_events_to.as_mut() {
+                let time_offset_secs = self
+                    .clock
+                    .new_timestamp()
+                    .get_diff_duration(&self.start_timestamp)
+                    .as_secs_f64();
+                write_events_to.mark_poisoned(&err, time_offset_secs);
+            }
         }
         self.scheduler.add_event(event);
     }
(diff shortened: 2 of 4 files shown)


heyong4725:
## /pr-review caught an architectural gap — fixed in `928cf796`

The original commit (`a37fa381`) made the recorder emit `recording_status` in its JSON output. Codex adversarial review on the diff caught that this is necessary-but-not-sufficient: **the canonical consumer in this repo (`apis/rust/node/src/daemon_connection/node_integration_testing.rs:47`) deserializes recordings into `IntegrationTestInput`, which had no `recording_status` field, and `IntegrationTestInput` doesn't use `deny_unknown_fields`** — so serde silently drops the new marker. `DORA_TEST_WITH_INPUTS=poisoned.json` would replay the partial trace as if complete, preserving the exact silent partial-recording behavior #1857 was supposed to prevent.

Verbatim from Codex:

> A file marked `{"state":"poisoned"}` will be replayed by `DORA_TEST_WITH_INPUTS` as if it were complete, preserving the exact silent partial-recording failure this PR is meant to prevent.

Real finding. Fixed in the follow-on commit. End-to-end now:

1. **Producer** (`event_stream::write_out`) emits `recording_status` via the canonical `RecordingStatus` enum from `dora-message`, not a hand-built JSON map. Serializer and deserializer can't drift.
2. **Wire schema** lives in `libraries/message/src/integration_testing_format.rs::RecordingStatus` with `#[serde(tag = "state", rename_all = "lowercase")]`. Three serde round-trip tests pin the shape (legacy form, clean form, poisoned form).
3. **Consumer** (`IntegrationTestingEvents::new`) checks the deserialized `recording_status` field and `eyre::bail!`s with an actionable error if `Poisoned`. Hand-authored fixtures and pre-#1857 recordings have `None` and pass through cleanly.

Side notes:
- `IntegrationTestInput.recording_status` is `Option<Box<RecordingStatus>>` (niche-optimized to 8 bytes) so the rare `Poisoned` variant doesn't inflate the common case.
- The bigger `IntegrationTestInput` struct triggered clippy's `large_enum_variant` on `TestingInput::Input(IntegrationTestInput)`. Documented and `#[allow]`-ed in `integration_testing.rs` — boxing the variant would churn 32 call sites for no perf win on a non-hot-path setup API.

Updated test plan:
- `cargo test -p dora-message --lib` → 103 pass (was 100; +3 round-trip tests)
- `cargo test -p dora-node-api --lib` → 82 pass (unchanged; existing 3 write-side tests still valid)
- `cargo clippy --all --exclude *python -- -D warnings` ✓
- `cargo check --examples` ✓

PR is now end-to-end honest: poisoned recordings can't masquerade as clean anywhere in the codebase. Net diff: +313/-16 across 4 files.

PR #1870, opened 2026-05-19
fix(core): verify zenoh listener actually bound via info().locators() (closes #1858)

### LINKED ISSUE #1858: zenoh listener bind race: open Ok ≠ actually bound (follow-up to #1855 / #1856)
## Context

Surfaced by Codex adversarial review on #1855 (and independently flagged as a low-confidence appendix item in the same review). The bind race is pre-existing — it predates PR 1608 — but the source-of-truth refactor in #1855 (`open_zenoh_session_with_listen` returning `(Session, Option<String>)`, closing #1856) makes the residual gap structurally visible: the helper now *claims* to be the source of truth for whether the listener bound, but in fact it uses `zenoh::open(...) returned Ok` as a proxy, which is not the same thing.

## The shape of the race

1. `reserve_loopback_zenoh_endpoint()` binds port `N` on `127.0.0.1`, reads the port, drops the socket.
2. Tiny window: some other process binds port `N`.
3. `insert_json5("listen/endpoints", "tcp/127.0.0.1:N")` succeeds — it's just JSON5 parsing into the in-memory config, no bind attempt.
4. `zenoh::open(zenoh_config)` tries to bind `N` and fails. But we previously set `listen/exit_on_failure: false` (`libraries/core/src/topics.rs:134`), so zenoh logs the bind failure and returns `Ok(Session)` anyway.
5. Helper sees `Ok` → assigns `effective_listen_endpoint = Some("tcp/127.0.0.1:N")` at `libraries/core/src/topics.rs:151`.
6. Daemon stores it as `daemon.zenoh_listen_endpoint`.
7. Daemon's `Spawner.maybe_inject_zenoh_connect` at `binaries/daemon/src/spawn/spawner.rs:121` injects `DORA_ZENOH_CONNECT=tcp/127.0.0.1:N` into every spawned node.
8. Spawned nodes' `open_zenoh_session_with_listen` sees the env var, inserts it as their `connect/endpoints`, and disables multicast scouting (`libraries/core/src/topics.rs:103`).
9. Nodes try to connect to port `N`, which is either nothing or some unrelated process. Connection fails. Multicast was disabled, so there's no fallback. Large-payload data path partitioned.

The three fixes that landed in #1855 do NOT cover this case: the insert succeeds, the configured `zenoh::open` succeeds, the default-config fallback path is never reached, and the daemon's `requested.is_some() && effective.is_none()` warn at `binaries/daemon/src/lib.rs:807` never fires.

## Probability

Documented as "vanishingly rare" in the existing doc comment on `reserve_loopback_zenoh_endpoint` (`libraries/core/src/topics.rs:184`). The window between `drop(listener)` and zenoh's subsequent bind is microseconds in normal operation, and the OS allocates fresh ephemeral ports each call. Real-world hits would be:
- CI runners with aggressive port scanners
- Hosts running other services on the same ephemeral range
- Containers where the kernel port allocator is constrained

Low. But the consequence is silent partition, which is hard to debug when it does happen.

## Why this is pre-existing, not introduced by #1855

Pre-PR #1608 behavior: `.unwrap()` on each `insert_json5` would panic on insert failure. But `insert_json5` failing was never the bind-race scenario — that's `zenoh::open` reporting Ok with a silently-failed listen. Pre-#1608 had the SAME gap. PR #1855's source-of-truth refactor narrowed the failure modes the helper lies about (the helper used to lie whenever `insert_json5` failed; now it only lies in this much-narrower bind-race window), so this is a strict improvement.

## Fix strategies (in increasing order of effort)

1. **Doc honesty (10 min):** Update `reserve_loopback_zenoh_endpoint`'s doc + add a note on `open_zenoh_session_with_listen` stating that `effective_listen_endpoint = Some(ep)` is a best-effort signal — the configured open returning `Ok` does NOT guarantee the listener bound under `listen/exit_on_failure: false`. Defers the real fix but ensures future readers don't take the "source of truth" claim too literally.

2. **Hold the reservation socket until zenoh binds (30 min):** Refactor `reserve_loopback_zenoh_endpoint` to return a `(String, TcpListener)` guard. The caller drops the listener just before `zenoh::open`. The race window collapses from "from `drop(listener)` to `zenoh::open` returns" down to "OS hands the port to zenoh" — which is the OS's normal port-reuse-on-close behavior and is much smaller. Doesn't close the window absolutely but reduces it to OS-level scheduling.

3. **Verify bind from the session post-open (20-60 min, exploratory):** Investigate `zenoh::Session` for an API that exposes actually-bound endpoints. If zenoh-rust 1.x exposes this (e.g. via `session.config()` or a `runtime`/`transport` accessor), use it to compute `effective_listen_endpoint` from observed reality instead of inserted config. This is the architecturally correct fix but depends on what zenoh exposes.

The combination of (2) + (1) is probably the right amount of work for the actual risk level. (3) is the dream but might force (2) anyway if zenoh doesn't expose it.

## References

- PR #1855 (the rescue that introduced the source-of-truth contract).
- `libraries/core/src/topics.rs:140` — the `Ok(zenoh_session)` arm where `effective_listen_endpoint` gets promoted on (insufficient) evidence.
- `libraries/core/src/topics.rs:184` — the existing doc comment acknowledging the race.
- `binaries/daemon/src/spawn/spawner.rs:121` — downstream consumer of the (possibly-stale) endpoint.
- Closed sibling issue #1856 — the source-of-truth refactor that exposed this residual gap.

heyong4725:
## Summary

Closes #1858. Closes the residual bind-race gap that the #1855 source-of-truth refactor left behind.

## The race in one sentence

`zenoh::open(zenoh_config).await` returning `Ok` is **necessary but not sufficient** proof that the listener bound — with `listen/exit_on_failure: false` (which we deliberately set to tolerate ephemeral races), zenoh silently accepts a failed bind and the helper would have happily returned `Some(ep)` for an endpoint nothing is listening on.

## The fix

Zenoh exposes `session.info().locators()` which returns the **actually-bound** listener locators (zenoh-unstable API, gated behind the `unstable` feature flag that the workspace already enables at `Cargo.toml:130`). After `zenoh::open` succeeds, we query the session and only set `effective_listen_endpoint = Some(ep)` when our requested `addr:port` substring appears in the returned locator list.

The match is `contains` (substring), not exact equality, so locator metadata suffixes (`#iface=lo0` etc) don't break the check. The required `addr:port` still has to appear, so a port collision produces no match → `None` → daemon doesn't inject `DORA_ZENOH_CONNECT` → spawned nodes fall back to multicast scouting (correct behavior).

## What this preserves

| Path | Behavior |
|---|---|
| No listen requested | `effective = None` (unchanged) |
| Listen requested, insert accepted, bind succeeded | `effective = Some(ep)` (was wrongly Some before in race scenario, now correctly Some) |
| Listen requested, insert accepted, bind **silently failed** | `effective = None` + warn (was wrongly Some before, now correctly None) |
| Listen requested, insert failed | `effective = None` (unchanged) |
| Fallback to default config | `effective = None` (unchanged) |

The `Ok((session, effective))` shape is unchanged. The daemon at `binaries/daemon/src/lib.rs:803-806` already uses the returned `Option<String>` as source of truth and only injects `DORA_ZENOH_CONNECT` when it's `Some`, so the daemon-side path needs no change.

## Why I didn't add a unit test

Triggering the race in a unit test would need:
1. A concurrent process grabbing the reserved port between `drop(listener)` and `zenoh::open` — timing-flaky and platform-dependent, OR
2. A real zenoh runtime mock — heavy

The integration assertion is now implicit at `binaries/daemon/src/lib.rs:807` where `requested.is_some() && effective.is_none()` fires on real bind failures, not just on upstream insert failures.

## Doc update

`reserve_loopback_zenoh_endpoint`'s doc comment previously acknowledged the race as accepted risk. Updated to describe the two-layer defense:
1. `listen/exit_on_failure: false` keeps the daemon alive if the bind silently fails
2. `info().locators()` verifies what actually bound

## Test plan

- [x] `cargo clippy --all --exclude dora-{node-api,operator-api,ros2-bridge}-python -- -D warnings`
- [x] `cargo test -p dora-core -p dora-node-api -p dora-daemon --lib` (82+ tests pass)
- [x] `cargo check --examples`
- [ ] CI green
- [ ] Next nightly's `Examples` and `CLI Tests` lanes stay green (this PR changes happy-path behavior only when the race fires, which is vanishingly rare)

## Risk

Touches one function, ~50 lines. The only behavioral delta in the happy path is one additional async call (`session.info().locators().await`) that returns quickly from runtime state. No new dependencies, no caller-side changes, no API contract change.

### DIFF
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -149,9 +149,43 @@ pub async fn open_zenoh_session_with_listen(
                 warn!("failed to set zenoh connect/endpoints for coordinator {addr}: {err}");
             }
             if let Ok(zenoh_session) = zenoh::open(zenoh_config).await {
-                // The configured session opened — promote the (possibly None)
-                // listener now that we know it's actually live.
-                effective_listen_endpoint = listen_inserted_into_configured;
+                // Verify the listener actually bound. `zenoh::open` returning
+                // Ok is necessary but not sufficient — with
+                // `listen/exit_on_failure: false` (set above), zenoh tolerates
+                // a silently-failed listen bind. The most plausible cause is
+                // the race between `reserve_loopback_zenoh_endpoint` dropping
+                // its reservation socket and zenoh's own bind, during which
+                // some other process could grab the port. Trusting `Ok` here
+                // would advertise an endpoint nothing is listening on, and
+                // spawned nodes would fail their `DORA_ZENOH_CONNECT` connect
+                // attempts (#1858).
+                //
+                // `info().locators()` is the zenoh-canonical "what actually
+                // bound" query (unstable API gated behind the workspace
+                // `unstable` feature, already enabled in the root Cargo.toml
+                // zenoh dependency).
+                if let Some(requested) = listen_inserted_into_configured {
+                    let bound_locators: Vec<String> = zenoh_session
+                        .info()
+                        .locators()
+                        .await
+                        .into_iter()
+                        .map(|l| l.as_str().to_string())
+                        .collect();
+                    // Substring match tolerates zenoh's metadata suffixes
+                    // (e.g. `tcp/127.0.0.1:43217#iface=lo0`) while still
+                    // requiring the exact addr:port we asked for to appear.
+                    let bound = bound_locators.iter().any(|l| l.contains(&requested));
+                    if bound {
+                        effective_listen_endpoint = Some(requested);
+                    } else {
+                        warn!(
+                            "zenoh session opened but listener for `{requested}` \
+                             did not bind (actually bound: {bound_locators:?}); \
+                             spawned nodes will use multicast scouting only"
+                        );
+                    }
+                }
                 zenoh_session
             } else {
                 warn!("failed to open zenoh session, retrying with default config");
@@ -177,11 +211,21 @@ pub async fn open_zenoh_session_with_listen(
 /// endpoint. Returns a string suitable for the zenoh `listen/endpoints`
 /// config (e.g. `tcp/127.0.0.1:43217`).
 ///
-/// There is a small race window between dropping the reservation socket and
-/// zenoh's own bind. `open_zenoh_session_with_listen` sets
-/// `listen/exit_on_failure: false` so the daemon survives that race; in
-/// practice the OS keeps allocating new ports so collisions are vanishingly
-/// rare.
+/// There is a small race window between dropping the reservation socket
+/// and zenoh's own bind. `open_zenoh_session_with_listen` defends against
+/// it on two layers:
+///
+/// 1. `listen/exit_on_failure: false` keeps the daemon alive if zenoh's
+///    bind silently fails inside the race window.
+/// 2. After `zenoh::open`, the helper queries `session.info().locators()`
+///    and only advertises the returned endpoint when our requested
+///    `addr:port` is actually in the bound-locator list. If the port was
+///    grabbed by another process, the returned `effective_listen_endpoint`
+///    is `None` and callers fall back to multicast scouting instead of
+///    advertising a phantom endpoint (#1858).
+///
+/// In practice the OS keeps allocating fresh ephemeral ports each call,
+/// so collisions remain vanishingly rare.
 pub fn reserve_loopback_zenoh_endpoint() -> std::io::Result<String> {
     let listener = std::net::TcpListener::bind((Ipv4Addr::LOCALHOST, 0))?;
     let port = listener.local_addr()?.port();
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -172,10 +172,26 @@ pub async fn open_zenoh_session_with_listen(
                         .into_iter()
                         .map(|l| l.as_str().to_string())
                         .collect();
-                    // Substring match tolerates zenoh's metadata suffixes
-                    // (e.g. `tcp/127.0.0.1:43217#iface=lo0`) while still
-                    // requiring the exact addr:port we asked for to appear.
-                    let bound = bound_locators.iter().any(|l| l.contains(&requested));
+                    // Strip zenoh's endpoint-string separators before
+                    // comparing. Per `zenoh-protocol::core::endpoint`,
+                    // `?` separates metadata and `#` separates config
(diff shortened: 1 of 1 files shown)


heyong4725:
## /pr-review pass — two real fixes + one false positive

Two real findings from Codex adversarial review on the original commit. Fixed in `0c9336a3`.

### 1. Port-prefix false positive (caught by structured review)

`bound_locators.iter().any(|l| l.contains(&requested))` would silently confirm a bind for the wrong port. Example: requested `tcp/127.0.0.1:5000` would match a bound `tcp/127.0.0.1:50000` because the first string is a strict prefix of the second. Narrow but real — exactly the kind of mismatch this check exists to catch.

**Fix:** exact match on the prefix before zenoh's separators.

### 2. Wrong endpoint separator (caught by Codex adversarial)

The original strip used `#` (config separator). But per `zenoh-protocol-1.8.0/src/core/endpoint.rs:22-24`:
```rust
pub const PROTO_SEPARATOR: char = '/';
pub const METADATA_SEPARATOR: char = '?';
pub const CONFIG_SEPARATOR: char = '#';
```

And `From<EndPoint> for Locator` at `locator.rs:79-86` already truncates `#`:
```rust
impl From<EndPoint> for Locator {
    fn from(mut val: EndPoint) -> Self {
        if let Some(cidx) = val.inner.find(CONFIG_SEPARATOR) {
            val.inner.truncate(cidx);
        }
        Locator(val)
    }
}
```

So my `split('#')` was dead code on the wire. The actual surviving suffix would be `?metadata=...`. Switched to `split(['?', '#'])` — strips both, defensively.

### 3. Codex's `block_in_place` panic claim — false positive

Codex flagged `info().locators().await` as potentially panicking on Tokio CurrentThread runtime. Verified false:

```rust
// zenoh-1.8.0/src/net/runtime/mod.rs:205
fn get_locators(&self) -> Vec<Locator> {
    self.locators.read().unwrap().clone()
}
```

Synchronous `RwLock::read().clone()`. No `block_in_place`, no panic. Codex confused this with `get_zids()` two methods down at line 221, which IS gated on `ZRuntime::Application.block_in_place`. Different method, different call path.

### Diff summary

- Original commit `43001d9c`: +52/-8 in `topics.rs`
- Review fixup `0c9336a3`: +20/-4 in same file
- Total: +72/-12, single file

Issue #1882, opened 2026-05-20
nightly: CLI Tests (macos-latest) — python-dataflow sender exits 143 on --stop-after


heyong4725:
## Summary

The `CLI Tests (macos-latest)` nightly job has been failing on the `python-dataflow` `--stop-after 10s` step for several nights. The Linux equivalent passes. Filed as a tracked regression so it doesn't get lost while the cross-check failures (#1881) take priority.

## Symptom

In `python-dataflow` cli-test step (`dora run dataflow.yml --uv --stop-after 10s`):

```
sender: Sent message 90 (~10 seconds in, just before --stop-after fires)
dora_coordinator: Received destroy command
dora_daemon: received destroy command -> exiting
dora_coordinator: Coordinator took 206ms for handling event: Control
dora_coordinator: stopped

[ERROR]
Dataflow 019e40ae-d20f-755a-9796-993037f95238 failed:

Node `sender` failed: exited with code 143
```

Exit code **143** = 128 + SIGTERM. The sender was terminated by SIGTERM as part of the `--stop-after`-triggered destroy sequence, which is **expected shutdown behavior**, but dora reports the resulting exit code as a node failure.

## What's going wrong (two candidate root causes)

Either:

1. **dora's `--stop-after` shutdown path treats SIGTERM as a node failure** — the destroy command sends SIGTERM to nodes, then the dataflow result aggregator interprets the resulting 143 as `Node failed: exited with code 143`. This would mean the same pattern would surface on Linux too, but Linux passes... so probably not this.

2. **Python sender on macOS doesn't trap SIGTERM in time and gets force-killed** — the dora node API on Python likely listens for the stop signal and exits cleanly via a `KeyboardInterrupt`-style mechanism on Linux, but the equivalent on macOS isn't catching SIGTERM fast enough (or at all), so the sender gets force-killed. Linux's signal delivery + Python interrupt handling tends to be faster than macOS's. This is more likely.

Examples to inspect:
- `examples/python-dataflow/sender.py:17` — the only line that's emitting messages
- The shutdown handling in `apis/python/node/src/lib.rs` and `apis/python/node/dora/__init__.pyi`

## Reproduction

Local (macos-arm64):
```bash
dora build examples/python-dataflow/dataflow.yml --uv
dora run examples/python-dataflow/dataflow.yml --uv --stop-after 10s
```

Expected: clean exit, zero error.
Actual: `Node sender failed: exited with code 143`.

## Timeline

[...]

PR #1888, opened 2026-05-20
fix(ci): fetch nightly failure logs per job

phil-opp:
Fix nightly failure comments so they fetch completed failed job logs while the reporting workflow is still running.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1667,26 +1667,56 @@ jobs:
             logs_available=0
           fi
 
+          fetch_failed_job_log_once() {
+            local job_id="$1"
+            local job_name="$2"
+            local job_log_file="$3"
+
+            rm -f "$job_log_file"
+            if gh run view "$RUN_ID" --repo "$REPO" --job "$job_id" --log-failed \
+              > "$job_log_file" 2>> "$failed_logs_error" && [ -s "$job_log_file" ]; then
+              return 0
+            fi
+
+            echo "Failed-step log unavailable or empty for job '$job_name' ($job_id); falling back to the full job log." >> "$failed_logs_error"
+            if curl -fsSL \
+              -H "Authorization: Bearer $GH_TOKEN" \
+              -H "Accept: application/vnd.github+json" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              "$API_URL/repos/$REPO/actions/jobs/$job_id/logs" \
+              > "$job_log_file" 2>> "$failed_logs_error" && [ -s "$job_log_file" ]; then
+              return 0
+            fi
+
+            return 1
+          }
+
+          fetch_failed_job_log() {
+            local job_id="$1"
+            local job_name="$2"
+            local job_log_file="$3"
+
+            if fetch_failed_job_log_once "$job_id" "$job_name" "$job_log_file"; then
+              return 0
+            fi
+
+            echo "Log for failed job '$job_name' ($job_id) is not ready yet; retrying once after 5s." >> "$failed_logs_error"
+            sleep 5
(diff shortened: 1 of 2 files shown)


heyong4725 (approved):
Thanks for tackling this — the bot comment on #1882 yesterday hit the exact failure mode you're fixing (`Unable to fetch failed GitHub Actions log lines... run X is still in progress`). The per-job enumeration + curl fallback chain is the right design. Strict improvement over the broken single-call shape. Approving.

Three polish items worth a follow-up commit if you want them; none block merge.

## Real concerns

### 1. UTF-8-aware tail in `bounded_tail_file` (line 1722)

```bash
tail -c "$max_bytes" "$file"
```

`tail -c` is a byte slice. Runner logs are UTF-8. Slicing mid-multibyte-sequence produces a malformed UTF-8 prefix that GitHub's markdown renderer replaces with `?` or silently drops. With a 10 000-byte excerpt budget and arbitrary log content, this fires on a non-trivial fraction of runs — first few characters of the truncated excerpt may be garbage. Fix:

```bash
tail -c "$max_bytes" "$file" | iconv -c -f UTF-8 -t UTF-8
```

`iconv -c` strips invalid bytes. Pure cosmetic; doesn't change any execution path.

### 2. Optional `\r` strip after `read`

```bash
while IFS=$'\t' read -r job_id job_name; do
  job_name="${job_name%$'\r'}"   # defensive
```

I believe `jq @tsv` actually errors out on `\r` in input values (jq docs say "the script terminates with an error" for embedded tab/newline/CR/backslash), so this should never trigger. But it's a 1-line defensive measure if a future jq version softens that semantic, and it makes the loop a hair more robust to weird CI runner line-ending quirks.

### 3. Single retry on per-job log-fetch race (lines 1674–1687)

GitHub's per-job log storage is asynchronous — a job can be `completed.failure` per the API while its log isn't yet indexed for `gh run view --log-failed` or the `/jobs/$id/logs` endpoint. The fallback chain (`gh --log-failed` → `curl`) handles each-individually-empty case, but if BOTH are racing the same log-indexing lag, we lose the excerpt. A single `sleep 5 + retry` between (1) and (2) (or wrapping the whole thing) would close the practical window for most cases. Cheap insurance; can defer.

## Adversarial findings I'm rejecting

- **GH_TOKEN in curl argv** — single-tenant ephemeral runner, lifetime measured in ms. Real risk near zero. Could swap curl for `gh api` (which follows the 302-to-S3 the same way thanks to gh's `--log-failed` chain handling) but it's not worth the diff churn.
- **Pattern regex collisions** — no two `job_log_pattern` entries are subsets of each other today, verified manually. Naming-convention discipline going forward; current case-statement is fine.
- **Tempfile leak on SIGKILL** — runner is ephemeral, non-issue.

## What I verified

- `bash =~` and the prior `awk $1 ~ pattern` are equivalent for the actual `job_log_pattern` strings (all anchored ERE, escaped metacharacters, `(.+)` groups). ✓
- `gh api --paginate --jq @tsv` over the jobs endpoint produces well-formed concatenated TSV; `read -r IFS=$'\t'` handles it correctly. ✓
- The `trap ... EXIT` cleanup matches the four new tempfile/tempdir vars. ✓
- The fallback (no-pattern-match) path concatenates all failed jobs, then `bounded_tail_file` caps to `GLOBAL_FALLBACK_EXCERPT_BYTES` (4000) — explicit cap at line 1779-1781. ✓

## Verification

Couldn't end-to-end-test the bot comment locally (requires a real failing nightly + secrets.GITHUB_TOKEN), but the diff reads cleanly and the failure-mode-to-fix chain is sound.

Ready to merge. The two polish items above are nice-to-haves — happy to send a follow-up PR with them after this lands, or you can squash them in here.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

phil-opp:
Addressed in fb09b74

PR #1896, opened 2026-05-20
fix(daemon): emit --rt setup diagnostics before tracing init (#1701)

### LINKED ISSUE #1701: E2E coverage gap: dora daemon --rt (real-time profile)
## Gap

`dora daemon --rt` enables the real-time profile: `mlockall(MCL_CURRENT | MCL_FUTURE)` + `SCHED_FIFO` scheduling priority. Declared at `binaries/cli/src/command/daemon.rs:51-56`, applied at line 83+.

No E2E test exercises this path. We have no evidence that:

1. The flag parses and the daemon accepts it.
2. `mlockall` actually locks memory (requires `CAP_IPC_LOCK`).
3. `SCHED_FIFO` actually promotes the main thread (requires `CAP_SYS_NICE`).
4. A dataflow runs successfully under the RT profile.
5. Failures (missing capabilities) produce a useful error vs. a silent fallback.

The feature is documented at `docs/realtime-tuning.md` and promoted as a soft-realtime path — if any of the five items above is broken, the README is misleading users.

## Surfaced by

@haixuanTao in dora-rs/dora#1624, 2026-04-21.

## Expected test

A small-scope test that:

1. Sets the process's capabilities (or runs under a container that has them), OR checks the fallback error when capabilities are absent.
2. Invokes `dora daemon --rt` against a minimal coordinator.
3. Spawns a tiny dataflow (e.g., `examples/rust-dataflow` with 10-second timeout).
4. Asserts the dataflow completed successfully AND that the daemon logs contain `RT: mlockall enabled` / `RT: SCHED_FIFO` (or the documented fallback-warning string when capabilities missing).

Platform note: `mlockall` + `SCHED_FIFO` are Linux-only. The test should gate on `cfg(target_os = "linux")` and be marked as skipped on other OSes.

## Impact

Untested = no gate against regressions. Users following `docs/realtime-tuning.md` may hit a broken flag at worst, or an unpromoted main thread at worst-of-best.

## Affected files

- `binaries/cli/src/command/daemon.rs:51-105`
- `docs/realtime-tuning.md`

heyong4725:
## Summary

- Fixes the latent silent-fallback bug in `dora daemon --rt`: the RT setup at `binaries/cli/src/command/daemon.rs:84-122` ran **before** `init_tracing_subscriber` registered the global default subscriber, so the previous `tracing::info!`/`warn!` calls about `mlockall` and `SCHED_FIFO` success/failure were routed to `NoSubscriber` and silently dropped — exactly the "useful error vs. silent fallback" gap #1701 flagged as untested.
- Switches those six call sites to `eprintln!` so the diagnostics are visible unconditionally. Going to stderr also (correctly) means `--quiet` no longer suppresses them: a failed RT promotion is an operational warning the user must see.
- Adds a regression gate in `tests/example-smoke.rs::smoke_daemon_rt_emits_setup_messages` (`#[cfg(unix)]`) that asserts the messages actually appear. Linux accepts either the success or `failed:` variant of both `mlockall` and `SCHED_FIFO`, so it runs on unprivileged CI without needing `CAP_IPC_LOCK` / `CAP_SYS_NICE`. macOS asserts the documented "SCHED_FIFO not available" fallback path.

Closes #1701.

## Why eprintln! instead of reordering tracing init

Two clean options were on the table:

1. Move the `init_tracing_subscriber` block above the RT setup so `tracing::*` calls work.
2. Replace the six `tracing::*` calls in the RT block with `eprintln!`.

Option 2 was chosen because (a) it is more surgical (six lines vs. moving ~25 lines of subscriber/OtelGuard setup past the runtime construction, with new ordering subtleties), and (b) it aligns with the "RT failures must always be visible" semantic — the existing `--quiet` flag should not silence a mlockall/SCHED_FIFO warning that materially changes daemon behavior.

## Verified RED to GREEN

Empirical confirmation that the test guards the bug:

- With the daemon fix stashed (original `tracing::*` code restored):
  ```
  thread smoke_daemon_rt_emits_setup_messages panicked at tests/example-smoke.rs:1481:5:
  expected RT: mlockall line on stderr (success or failure variant).
  stderr was:
  ```
  Empty stderr — confirms the messages are dropped by `NoSubscriber`.

- With the fix applied:
  ```
  test result: ok. 1 passed; 0 failed
  ```

Manual stderr capture under the fix (macOS):
```
RT: mlockall failed: Function not implemented (os error 78). Ensure CAP_IPC_LOCK or ulimit -l unlimited.
RT: SCHED_FIFO not available on this platform (mlockall applied)
```

## Test plan

- [x] `cargo clippy -p dora-cli --tests -- -D warnings`
- [x] `cargo test -p dora-cli --lib` — 146 passed
- [x] `cargo test --test example-smoke smoke_daemon_rt_emits_setup_messages` — passes on macOS
- [ ] CI Linux: same test should pass on unprivileged GHA runner (asserts either success or `failed:` variant of mlockall + SCHED_FIFO)
- [ ] Manual: confirm `dora daemon --rt --quiet` still prints RT messages (operational correctness — must not be suppressed by `--quiet`)

### DIFF
--- a/binaries/cli/src/command/daemon.rs
+++ b/binaries/cli/src/command/daemon.rs
@@ -124,8 +124,11 @@ impl Executable for Daemon {
                         );
                     }
                 }
+                // Note: the previous "(mlockall applied)" parenthetical here
+                // could lie on macOS where `mlockall` returns ENOTSUP, so the
+                // message now stands on its own.
                 #[cfg(not(target_os = "linux"))]
-                eprintln!("RT: SCHED_FIFO not available on this platform (mlockall applied)");
+                eprintln!("RT: SCHED_FIFO not available on this platform");
             }
             #[cfg(not(unix))]
             eprintln!("RT: --rt flag is only supported on Unix systems");
--- a/binaries/cli/src/command/daemon.rs
+++ b/binaries/cli/src/command/daemon.rs
@@ -81,15 +81,24 @@ impl Executable for Daemon {
         let rt = builder.build().context("tokio runtime failed")?;
 
         // Apply real-time profile if requested.
+        //
+        // These diagnostics use `eprintln!` rather than `tracing::*` because
+        // the tracing subscriber is not yet installed at this point (see the
+        // `init_tracing_subscriber` call below) — `tracing::info!`/`warn!`
+        // calls before global subscriber registration route to
+        // `NoSubscriber` and are silently dropped (#1701). Going to stderr
+        // also ensures these are visible even with `--quiet`, which matters
+        // because a failed RT setup is an operational warning the user must
+        // see.
         if self.rt {
             #[cfg(unix)]
             {
                 // Lock all memory to prevent page faults.
                 let lock_result = unsafe { libc::mlockall(libc::MCL_CURRENT | libc::MCL_FUTURE) };
                 if lock_result == 0 {
-                    tracing::info!("RT: mlockall enabled (memory locked)");
+                    eprintln!("RT: mlockall enabled (memory locked)");
                 } else {
-                    tracing::warn!(
+                    eprintln!(
                         "RT: mlockall failed: {}. Ensure CAP_IPC_LOCK or ulimit -l unlimited.",
                         std::io::Error::last_os_error()
                     );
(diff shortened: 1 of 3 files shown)


heyong4725 (commented):
No findings from review. I checked the early daemon --rt diagnostics path and the new smoke test, including the no-coordinator retry behavior and a closed-stderr probe; the change keeps the RT messages visible before tracing subscriber initialization without introducing a blocking startup path.\n\nValidated locally:\n- cargo fmt --all -- --check\n- cargo test --test example-smoke smoke_daemon_rt_emits_setup_messages

PR #1898, opened 2026-05-21
test(cli): E2E coverage for --allow-shell-nodes security gate (#1702)

### LINKED ISSUE #1702: E2E coverage gap: dora run --allow-shell-nodes (shell escape gate)
## Gap

`dora run --allow-shell-nodes` disables the default security gate that blocks shell nodes from executing arbitrary commands. Declared at `binaries/cli/src/command/run.rs:83`, applied at line 142 by setting the `DORA_ALLOW_SHELL_NODES=true` environment variable before spawning nodes.

No E2E test exercises this flag. We have no evidence that:

1. The flag parses and the env var is set.
2. A dataflow with `shell:` directives actually executes them when the flag is on.
3. A dataflow with `shell:` directives is **blocked** when the flag is off (the whole point of the gate).
4. The error message when a shell node is rejected is useful.

Item 3 is the load-bearing one: this is a **security control**. If the gate silently allows shell execution without the flag, that's a real vulnerability, not just a missing test.

## Surfaced by

@haixuanTao in dora-rs/dora#1624, 2026-04-21.

## Expected test

Two paired tests in `tests/ws-cli-e2e.rs` or `tests/example-smoke.rs`:

1. **Positive path**: a tiny dataflow YAML with a `shell: echo hello > /tmp/dora-shell-test.out` node. Run with `--allow-shell-nodes`. Assert the file was created and contains `hello`.

2. **Negative path**: same YAML, run WITHOUT `--allow-shell-nodes`. Assert the dataflow exits with a clear error containing "shell" / "not allowed" / similar, and that `/tmp/dora-shell-test.out` was NOT created.

## Impact

A security-relevant flag with zero automated coverage. A regression that silently allowed shell execution without the flag would bypass the security model without anyone noticing.

## Affected files

- `binaries/cli/src/command/run.rs:78-83` (flag declaration)
- `binaries/cli/src/command/run.rs:142` (env var application)
- Wherever the shell-node check is enforced — likely `libraries/core/src/descriptor/` or `binaries/daemon/src/spawn/` (need to grep for `DORA_ALLOW_SHELL_NODES`).

heyong4725:
## Summary

Add paired E2E smoke tests for the `dora run --allow-shell-nodes` security gate (`binaries/daemon/src/spawn/command.rs:24`). Closes #1702.

- `smoke_shell_node_allowed_with_flag` — positive path: tiny `path: shell` dataflow + `--allow-shell-nodes` → asserts the shell command produced the marker file with the expected content.
- `smoke_shell_node_blocked_without_flag` — negative path (load-bearing for the security model): same dataflow WITHOUT the flag → asserts dora exits non-zero, stderr contains the documented `"Shell nodes are disabled"` string, and **no marker file is created**.

Both tests `#[cfg(unix)]` (Linux + macOS `sh -c` form). Each child is launched with `.env_remove("DORA_ALLOW_SHELL_NODES")` so the CLI flag is the sole source of truth — an inherited env var (CI runner, shell session, previous test contamination) cannot silently let the gate pass.

## Why the negative test matters

A regression that silently dropped the gate check would let shell nodes execute by default, bypassing the security model without anyone noticing. This test is the canary — `dora` exiting cleanly on a `path: shell` dataflow without the flag would be a real vulnerability, not just a missing test.

## Verified RED → GREEN

Temporarily replaced the gate condition with `if false && std::env::var(...)` and re-ran:

```
thread 'smoke_shell_node_blocked_without_flag' panicked at tests/example-smoke.rs:1542:5:
dora run unexpectedly SUCCEEDED without --allow-shell-nodes — the security gate failed.
```

Restoring the gate makes the test pass. The negative test is genuinely load-bearing against the security regression class.

## Test plan

- [x] `cargo clippy -p dora-cli --tests -- -D warnings`
- [x] `cargo test --test example-smoke smoke_shell_node -- --test-threads=1` — 2/2 pass on macOS
- [ ] CI Linux: same tests should pass under unprivileged GHA runner
- [x] Adversarial: replaced the gate with `if false && ...` and confirmed the negative test catches it

### DIFF
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -1498,6 +1498,17 @@ fn smoke_shell_node_allowed_with_flag() {
         .expect("failed to run dora run --allow-shell-nodes");
 
     let stderr = String::from_utf8_lossy(&output.stderr);
+    // Assert dora exited cleanly BEFORE checking the marker. Otherwise a
+    // regression where the shell node runs successfully but a subsequent
+    // lifecycle/result step fails (e.g. dataflow result handling returns
+    // non-zero) would still produce the marker file and slip past the
+    // test (review feedback on PR #1898).
+    assert!(
+        output.status.success(),
+        "dora run --allow-shell-nodes exited non-zero (status={:?}); \
+         marker file alone is not enough.\nstderr:\n{stderr}",
+        output.status.code()
+    );
     assert!(
         marker.exists(),
         "shell command did not produce marker file at {marker:?}; \
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -1437,6 +1437,128 @@ fn smoke_local_queue_size_latest_data_rust() {
 }
 
 // ---------------------------------------------------------------------------
+// Shell-node gate (issue #1702)
+// ---------------------------------------------------------------------------
+//
+// `dora run --allow-shell-nodes` toggles a security gate enforced at
+// `binaries/daemon/src/spawn/command.rs:24`. Both branches of that `if`
+// matter: without the flag, dora MUST refuse to spawn shell nodes.
+// The negative test is therefore load-bearing security coverage —
+// a regression that silently allowed shell execution would bypass the
+// security model. Linux + macOS only (`sh -c` form); Windows uses
+// `cmd /C` and would need a separate fixture.
+
+fn unique_temp_path(stem: &str) -> std::path::PathBuf {
+    let nanos = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
+        .as_nanos();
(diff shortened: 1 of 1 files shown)


heyong4725 (commented):
I found one test-coverage issue:

- tests/example-smoke.rs:1501: the positive path only asserts that the shell command wrote the marker, but never asserts that dora run --allow-shell-nodes exited successfully. The shell command can create the marker before later lifecycle/result handling fails, so a regression that returns a non-zero status after spawning the shell node would still pass this smoke test. Please add an output.status.success() assertion before checking the marker, including stderr in the failure message.

Validated locally:
- cargo fmt --all -- --check
- cargo clippy -p dora-cli --tests -- -D warnings
- cargo test --test example-smoke smoke_shell_node -- --test-threads=1

Issue #1703, opened 2026-04-21
E2E coverage gap: dora node stop/start/restart across Rust/Python/C++


heyong4725:
## Gap

`dora node stop`, `dora node restart`, `dora node add`, `dora node remove`, `dora node connect`, `dora node disconnect` — the full dynamic-topology CLI surface — is not exercised end-to-end against **all three supported node languages** (Rust, Python, C++).

Per `binaries/cli/src/command/node/`, the subcommands exist:

```
list        List all currently running nodes and their status.
info        Show detailed information about a specific node.
add         Add a node to a running dataflow
remove      Remove a node from a running dataflow
connect     Connect two nodes in a running dataflow (add a mapping)
disconnect  Disconnect two nodes in a running dataflow (remove a mapping)
restart     Restart a specific node without stopping the entire dataflow.
stop        Stop a specific node without stopping the entire dataflow.
```

`tests/ws-cli-e2e.rs` and `tests/example-smoke.rs` do not cover any of these per-node lifecycle operations. Related: dora-rs/dora#1682 filed a blocker on `dora node add` (daemon dispatch timeout), which means the fix for that is a prerequisite for meaningful coverage here.

## Surfaced by

@haixuanTao in dora-rs/dora#1624, 2026-04-21. Note: reported as "mentioned for 1 week" — whatever prior discussion exists should be linked here.

## Expected coverage matrix

For each of the 8 subcommands × 3 languages (Rust, Python, C++), at least one integration test:

| Subcommand | Rust | Python | C++ |
|---|---|---|---|
| `dora node list` | ? | ? | ? |
| `dora node info <node>` | ? | ? | ? |
| `dora node add` | blocked on #1682 | blocked on #1682 | blocked on #1682 |
| `dora node remove` | blocked on #1682 | blocked on #1682 | blocked on #1682 |
| `dora node connect` | blocked on #1682 | blocked on #1682 | blocked on #1682 |
| `dora node disconnect` | blocked on #1682 | blocked on #1682 | blocked on #1682 |
| `dora node restart <node>` | ? | ? | ? |
| `dora node stop <node>` | ? | ? | ? |

24 test cells. Realistically, a compact shape is: one long-running dataflow per language that exposes a node of each kind; the test spawns the dataflow with `--detach`, exercises all 8 subcommands against it, asserts via `dora node list` / observable side effects, then tears down.

## Impact

[...]

PR #1900, opened 2026-05-21
fix(coordinator,daemon): typed reply for AddMapping/RemoveMapping (unblocks dora node connect/disconnect)

heyong4725:
## Summary

`dora node connect` and `dora node disconnect` always timed out after 30s with `daemon dispatch failed: timeout waiting for daemon WS reply`. Reproducible against `examples/dynamic-add-remove` on a freshly-built binary, surfaced while trying to add E2E coverage for the `dora node *` matrix in #1703.

Root cause is the silent-reply class of bug previously fixed for `AddNode` (#1682 / #1873): the daemon's `AddMapping` / `RemoveMapping` handlers replied `None`, which the WS shim (`binaries/daemon/src/coordinator.rs:262`) drops without sending a wire response. The coordinator's `send_and_receive` then waited for a `daemon_command` reply that never came until `TCP_READ_TIMEOUT` (30s) elapsed — even though the mapping had already been applied locally on the daemon side.

## Changes

- `libraries/message/src/daemon_to_coordinator.rs` — add `AddMappingResult(Result<(), String>)` and `RemoveMappingResult(Result<(), String>)` variants to `DaemonCoordinatorReply`.
- `binaries/daemon/src/lib.rs` — daemon's `AddMapping` / `RemoveMapping` handlers now return the typed outcome via `Some(reply)`, including an explicit `Err` for unknown `dataflow_id` (previously silently ignored).
- `binaries/coordinator/src/lib.rs` — add `ensure_add_mapping_applied` / `ensure_remove_mapping_applied` helpers mirroring `ensure_add_node_applied`. Coordinator's CLI dispatch validates the reply variant before committing state, so a foreign reply or an explicit daemon `Err` surfaces to the CLI as `ControlRequestReply::Error` instead of committing on any successful TCP response.

## Tests

Six new unit tests for the validators (same shape as `add_node_reply_*` / `remove_node_reply_*`):

- `add_mapping_reply_accepts_daemon_success`
- `add_mapping_reply_reports_daemon_rejection`
- `add_mapping_reply_rejects_wrong_reply_variant`
- `remove_mapping_reply_accepts_daemon_success`
- `remove_mapping_reply_reports_daemon_rejection`
- `remove_mapping_reply_rejects_wrong_reply_variant`

All six pass. Existing coordinator/daemon tests unchanged.

## Verified end-to-end

Against `examples/dynamic-add-remove` (Python sender/receiver + filter add):

| Command | Before this PR | After |
|---|---|---|
| `dora node add --from-yaml filter-node.yml` | OK | OK |
| `dora node connect sender/value filter/input` | 30s timeout, mapping silently applied | OK |
| `dora node connect filter/output receiver/filtered` | 30s timeout | OK |
| `dora node disconnect filter/output receiver/filtered` | 30s timeout | OK |
| `dora node remove filter` | timeout (cascading) | OK |
| `dora node list` afterward | filter still visible | filter gone |

## Test plan

- [x] `cargo clippy --all --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python -- -D warnings`
- [x] `cargo test -p dora-coordinator --lib mapping` — 6/6 pass
- [x] Manual reproduction with `examples/dynamic-add-remove` (results above)
- [ ] CI Linux + macOS + Windows (full test matrix on the merge queue)

## Follow-ups

- Issue #1703 (E2E coverage for all 8 `dora node` subcommands across Rust/Python/C++) — this PR is the prerequisite that unblocks meaningful coverage on 4 of the 8 subcommands.
- The earlier manual probe also surfaced that `dora node stop <node>` returns success but the node remains in `Running` state on the next `dora node list`. That looks like a separate bug (not a wire-protocol timeout), so I have not bundled a fix here. Will investigate in a follow-up.

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -1726,13 +1726,29 @@ async fn start_inner(
                                         })?;
                                         match daemon_connections.get_mut(daemon_id) {
                                             Some(conn) => match conn.send_and_receive(&msg).await {
-                                                Ok(_) => Ok(ControlRequestReply::MappingAdded {
-                                                    dataflow_id,
-                                                    source_node,
-                                                    source_output,
-                                                    target_node,
-                                                    target_input,
-                                                }),
+                                                Ok(reply_raw) => {
+                                                    // Validate the daemon reply is specifically an
+                                                    // `AddMappingResult` before reporting success,
+                                                    // mirroring the #1682 / #1873 rescue for AddNode.
+                                                    let src =
+                                                        format!("{source_node}/{source_output}");
+                                                    let tgt =
+                                                        format!("{target_node}/{target_input}");
+                                                    match ensure_add_mapping_applied(
+                                                        &reply_raw, &src, &tgt,
+                                                    ) {
+                                                        Ok(()) => {
+                                                            Ok(ControlRequestReply::MappingAdded {
+                                                                dataflow_id,
+                                                                source_node,
+                                                                source_output,
+                                                                target_node,
+                                                                target_input,
+                                                            })
+                                                        }
+                                                        Err(e) => Err(e),
+                                                    }
+                                                }
                                                 Err(e) => Err(eyre!("daemon dispatch failed: {e}")),
                                             },
                                             None => {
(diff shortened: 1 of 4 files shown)


heyong4725 (commented):
No findings from review. I checked the AddMapping/RemoveMapping reply path end to end: daemon now always returns a typed reply, coordinator validates the exact reply variant before reporting success, and the stricter missing-mapping error is surfaced through the existing CLI error path.

Validated locally:
- cargo test -p dora-coordinator --lib mapping
- cargo test --test ws-cli-e2e real_dataflow -- --test-threads=1

PR #1904, opened 2026-05-21
feat(c++/operator): add on_input_parse_error callback (closes #1879)

### LINKED ISSUE #1879: feat(c++/operator): wire InputParseError to a C++ callback (residual phase-5 gap)
## Summary

Surfaced while closing #993 (C++ roadmap completeness audit). All seven of @PavelGuzenfeld's roadmap phases substantially landed, except a narrow event-type gap in the C++ operator FFI surface.

## The gap

The Rust operator API exposes **4** event variants at `apis/rust/operator/src/lib.rs:34-48`:

```rust
pub enum Event<'a> {
    Input { id, metadata, data },
    InputParseError { id, error },
    InputClosed { id },
    Stop,
}
```

The C++ operator FFI at `apis/c++/operator/src/lib.rs` only dispatches **3**:

- `on_input(...)`
- `on_input_closed(...)`
- `on_stop(...)`

There is **no** `on_input_parse_error(...)` callback. `InputParseError` is actively emitted at `apis/rust/operator/src/raw.rs:55` when input deserialization fails, so the Rust-side error path is real, not dead code. Today a C++ operator that receives a malformed input gets silence where its Rust counterpart would get an explicit error event it could log, surface as health, or reroute.

## Why this is bounded

- One file structurally affected on the C++ FFI side (`apis/c++/operator/src/lib.rs`).
- One file on the wrap-the-event-stream side that needs the new dispatch arm.
- One docs file to update (`docs/api-cxx.md` "Operator API" section already lists 3 callbacks; add the fourth).
- One header symmetry update for the C operator API (`apis/c/operator/operator_api.h`) since the C surface has the same gap.

## Proposed shape

Mirror the existing `on_input` callback declaration in `apis/c++/operator/src/lib.rs`:

```rust
fn on_input_parse_error(
    op: Pin<&mut Operator>,
    id: rust::Str,
    error: rust::Str,
    output_sender: &mut OutputSender,
) -> DoraOnInputResult;
```

And add a matching dispatch arm in the event loop that runs `ffi::on_input_parse_error(...)`. Consumers (C++ operators) opt in by implementing the function; backward compat is preserved by also accepting a weak symbol or a default-no-op shim — maintainer call on the exact compat shape.

For the C operator API, parallel surface in `apis/c/operator/operator_api.h`.

## Verification

Add a small smoke test under `tests/cmake-find-package-operator-cxx/` or a new `tests/operator-input-parse-error-cxx/` that:
1. Builds an operator that records every event type via the new callbacks.
2. Feeds it a deliberately malformed input.
3. Asserts the `on_input_parse_error` callback received the expected id + error.

The C/C++ smoke fixtures added in #1875 are a good shape to copy from.

## Credit chain

- @NellyWhads filed the original C++ roadmap discoverability issue (#993).
- @PavelGuzenfeld wrote the 7-phase roadmap (#993 thread) and shipped phases 1, 6, and 7-zero-copy directly (#1403, #1428, #1431).
- Phases 2–5 landed via panel-faithful rescues this session: #1846, #1847, #1848, #1849.
- This residual gap was caught during the #993 closure audit (today).

PavelGuzenfeld:
Wires `Event::InputParseError` to a C++ callback. Previously silently dropped by the catch-all arm in `OperatorWrapper::on_event`.

## what changed

- `apis/c++/operator/src/lib.rs`: `on_input_parse_error(op, id, error, output_sender)` added to the cxx bridge `unsafe extern "C++"` block and dispatched in `on_event`
- `apis/rust/operator/src/lib.rs`: 7 unit tests — basic dispatch, empty strings, multiple errors in order, interleaved events, `Send+Sync` compile-time proof, two concurrency cases; clean under ASAN and TSAN
- both `c++-dataflow` and `cmake-dataflow` operator-rust-api examples updated (header + impl)
- `tests/cmake-find-package-operator-cxx`: fixture updated so the find_package compile-link smoke still passes
- `docs/api-cxx.md`: five callbacks documented; breaking-change note updated; OutputSender availability note widened

## breaking change

Same shape as #1849. Existing C++ operators must add a stub:

```cpp
DoraOnInputResult on_input_parse_error(Operator &op, rust::Str id, rust::Str error, OutputSender &output_sender) {
    (void)op; (void)id; (void)error; (void)output_sender;
    return { rust::String(), false };
}
```

## C-API parallel surface

Intentionally omitted. C operators consume raw bytes via `dora_read_data` and never hit the arrow parse path that emits this event. `operator_types.h` is auto-generated by `safer_ffi` — manual edits would be clobbered.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -177,6 +177,9 @@ tokio-stream = "0.1.11"
 tempfile = { workspace = true }
 assert2 = "0.3.16"
 tokio-tungstenite = "0.28"
+libloading = "0.8"
+dora-operator-api-types = { workspace = true }
+arrow-array = { workspace = true }
 
 [[example]]
 name = "c-dataflow"
@@ -226,6 +229,10 @@ path = "examples/multiple-daemons/run.rs"
 [[example]]
 name = "cmake-dataflow"
 path = "examples/cmake-dataflow/run.rs"
+
+[[test]]
+name = "cpp-operator-cxx-bridge"
+path = "tests/cpp-operator-cxx-bridge.rs"
 
 [[test]]
 name = "ws-cli-e2e"
--- a/apis/c++/operator/src/lib.rs
+++ b/apis/c++/operator/src/lib.rs
@@ -60,6 +60,20 @@ mod ffi {
         /// operator can emit a final output (e.g. flush buffered
         /// state, send a "shutdown summary") before returning.
         fn on_stop(op: Pin<&mut Operator>, output_sender: &mut OutputSender) -> DoraOnInputResult;
+
+        /// Called when an input's Arrow data fails to deserialize
+        /// (`Event::InputParseError { id, error }`). The daemon emits
+        /// this when `arrow::ffi::from_ffi` returns `Err` for a
+        /// received input. Without this callback the error was silently
+        /// swallowed by the catch-all `_ => Continue` arm, leaving the
+        /// operator with no way to log, surface as health, or reroute
+        /// malformed inputs.
+        fn on_input_parse_error(
+            op: Pin<&mut Operator>,
+            id: &str,
+            error: &str,
+            output_sender: &mut OutputSender,
+        ) -> DoraOnInputResult;
     }
 }
 
@@ -133,6 +147,19 @@ impl DoraOperator for OperatorWrapper {
                 let operator = self.operator.as_mut().unwrap();
                 let mut output_sender = OutputSender(output_sender);
                 let result = ffi::on_stop(operator, &mut output_sender);
+                if result.error.is_empty() {
+                    Ok(match result.stop {
+                        false => DoraStatus::Continue,
+                        true => DoraStatus::Stop,
+                    })
+                } else {
+                    Err(result.error)
+                }
+            }
+            Event::InputParseError { id, error } => {
+                let operator = self.operator.as_mut().unwrap();
+                let mut output_sender = OutputSender(output_sender);
+                let result = ffi::on_input_parse_error(operator, id, error, &mut output_sender);
                 if result.error.is_empty() {
                     Ok(match result.stop {
                         false => DoraStatus::Continue,
--- a/apis/rust/operator/src/lib.rs
+++ b/apis/rust/operator/src/lib.rs
@@ -80,7 +80,7 @@ impl DoraOutputSender<'_> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::sync::{Arc, Mutex};
+    use std::sync::Arc;
 
     struct RecordingOperator {
         parse_errors: Vec<(String, String)>,
@@ -223,59 +223,4 @@ mod tests {
         assert_send_sync::<SendOutput>();
     }
 
-    #[test]
-    fn parse_errors_accumulated_from_concurrent_event_streams() {
-        // Simulates two independent operator instances processing events
-        // concurrently — exercises the no-shared-state invariant.
-        let results: Arc<Mutex<Vec<String>>> = Arc::new(Mutex::new(Vec::new()));
-        let results2 = Arc::clone(&results);
-
-        let h1 = std::thread::spawn(move || {
-            let sender = noop_send_output();
-            let mut output_sender = DoraOutputSender(&sender);
-            let mut op = RecordingOperator::default();
-            for i in 0..100u32 {
-                op.on_event(
-                    &Event::InputParseError {
-                        id: "thread-1",
-                        error: format!("err-{i}"),
-                    },
-                    &mut output_sender,
-                )
-                .unwrap();
-            }
-            results2
-                .lock()
-                .unwrap()
-                .push(format!("t1:{}", op.parse_errors.len()));
-        });
-
-        let results3 = Arc::clone(&results);
-        let h2 = std::thread::spawn(move || {
-            let sender = noop_send_output();
-            let mut output_sender = DoraOutputSender(&sender);
-            let mut op = RecordingOperator::default();
-            for i in 0..100u32 {
-                op.on_event(
-                    &Event::InputParseError {
-                        id: "thread-2",
-                        error: format!("err-{i}"),
-                    },
-                    &mut output_sender,
-                )
-                .unwrap();
-            }
-            results3
-                .lock()
-                .unwrap()
-                .push(format!("t2:{}", op.parse_errors.len()));
-        });
-
-        h1.join().unwrap();
-        h2.join().unwrap();
-
-        let r = results.lock().unwrap();
-        assert!(r.contains(&"t1:100".to_string()));
-        assert!(r.contains(&"t2:100".to_string()));
-    }
 }
--- a/apis/rust/operator/src/lib.rs
+++ b/apis/rust/operator/src/lib.rs
@@ -76,3 +76,206 @@ impl DoraOutputSender<'_> {
         result.into_result()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::{Arc, Mutex};
+
+    struct RecordingOperator {
+        parse_errors: Vec<(String, String)>,
+    }
+
+    impl Default for RecordingOperator {
+        fn default() -> Self {
+            RecordingOperator {
+                parse_errors: Vec::new(),
+            }
+        }
+    }
+
+    impl DoraOperator for RecordingOperator {
+        fn on_event(
+            &mut self,
+            event: &Event,
+            _output_sender: &mut DoraOutputSender,
+        ) -> Result<DoraStatus, String> {
+            if let Event::InputParseError { id, error } = event {
+                self.parse_errors.push((id.to_string(), error.clone()));
+            }
+            Ok(DoraStatus::Continue)
+        }
+    }
+
+    fn noop_send_output() -> SendOutput {
+        use types::safer_ffi::closure::ArcDynFn1;
+        SendOutput {
+            send_output: ArcDynFn1::new(Arc::new(|_output| types::DoraResult::SUCCESS)),
+        }
+    }
+
+    #[test]
+    fn input_parse_error_dispatched_to_on_event() {
+        let sender = noop_send_output();
+        let mut output_sender = DoraOutputSender(&sender);
+        let mut op = RecordingOperator::default();
+
+        let event = Event::InputParseError {
+            id: "my-input",
+            error: "bad arrow data".to_string(),
+        };
+        let status = op.on_event(&event, &mut output_sender).unwrap();
+
+        assert!(matches!(status, DoraStatus::Continue));
+        assert_eq!(op.parse_errors.len(), 1);
+        assert_eq!(op.parse_errors[0].0, "my-input");
+        assert_eq!(op.parse_errors[0].1, "bad arrow data");
+    }
+
+    #[test]
+    fn other_events_do_not_trigger_parse_error_handler() {
+        let sender = noop_send_output();
+        let mut output_sender = DoraOutputSender(&sender);
+        let mut op = RecordingOperator::default();
+
+        let status = op.on_event(&Event::Stop, &mut output_sender).unwrap();
+        assert!(matches!(status, DoraStatus::Continue));
+        assert!(op.parse_errors.is_empty());
+    }
+
+    #[test]
+    fn multiple_parse_errors_accumulated_in_order() {
+        let sender = noop_send_output();
+        let mut output_sender = DoraOutputSender(&sender);
+        let mut op = RecordingOperator::default();
+
+        let ids: Vec<String> = (0..5u32).map(|i| format!("input-{i}")).collect();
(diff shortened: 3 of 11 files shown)


heyong4725:
Did a pre-landing review pass. The PR is solid — the dispatch in `OperatorWrapper::on_event` matches the established #1849 pattern, all 3 in-tree C++ operator examples + the `cmake-find-package-operator-cxx` fixture are updated symmetrically, and the docs reflect the new "five callbacks" count with both #1849 and #1879 credited in the breaking-change note. The "C-API parallel surface intentionally omitted" note is also correct: C-API operators consume raw bytes via `dora_read_data` and the runtime doesn't Arrow-deserialize for them, so there's no event to dispatch on that side.

Three informational findings — none are merge blockers, all are follow-up candidates.

### 1. The unit tests don't exercise the cxx-bridge dispatch path

`apis/rust/operator/src/lib.rs:81-281` — the seven `RecordingOperator` tests verify the Rust trait's `on_event` behavior, which is downstream of the actual addition. The new cxx-bridge call site in `OperatorWrapper::on_event` (`apis/c++/operator/src/lib.rs:159-171`) is the dispatch you're actually adding, and it's not exercised by any of these tests. Signature drift between `apis/c++/operator/src/lib.rs:74-79` and the C++ `operator.h` declarations would only fail at C++ compile of the example, not in `cargo test`.

This gap pre-existed for `on_input_closed`/`on_stop` from #1849, so it's a consistent-with-convention gap, not introduced here. Worth a follow-up: a smoke test that injects a deliberately malformed Arrow payload at the daemon side and asserts the C++ callback fires (e.g. capture `std::cerr` from the example operator's `on_input_parse_error` line you already added at `examples/c++-dataflow/operator-rust-api/operator.cc:45`).

### 2. `DoraOnInputResult` error/stop semantics are ambiguous

`apis/c++/operator/src/lib.rs:163-170` — the dispatch arm treats `result.error.is_empty()` as the success/failure discriminator. A C++ operator that sets `{error: "too many parse errors, shutting down", stop: true}` gets `stop` ignored and `Err(error)` returned, which the runtime treats as fatal rather than a clean `DoraStatus::Stop`. Same shape across all four callbacks (carried from #1849).

Not introduced by this PR, but the doc could call it out. Suggestion for `docs/api-cxx.md` (under the `DoraOnInputResult` section): note that `error` and `stop` are mutually exclusive — non-empty `error` always wins and produces a fatal failure regardless of `stop`.

### 3. Test #7 (`parse_errors_accumulated_from_concurrent_event_streams`) doesn't test what it claims

`apis/rust/operator/src/lib.rs:227-280` — the test spins up two `RecordingOperator` instances on two threads, each with its own `noop_send_output`. They share no state — each thread appends to its own `Vec<(String, String)>`. The `Arc<Mutex<Vec<String>>>` is only used to aggregate final counts after both `join()`. So the test confirms "two independent Rust threads can each append to their own Vec", which isn't in question.

If the intent was to validate the `ArcDynFn1` + `Send + Sync` path under concurrent callers (which test #6 only checks at compile time), the test should share a single `SendOutput` instance across both threads (e.g. a `SendOutput` whose closure captures `Arc<Mutex<Vec<DataId>>>` and records each output call). Otherwise consider deleting it — it adds runtime cost (200 events, two thread spawns) without coverage signal.

---

Aside from those nits, this is a clean addition that closes the gap #1849 left for `Event::InputParseError`. Verified the dispatch logic, the example/fixture updates, the doc surface, and the Event variant origin in `apis/rust/operator/src/raw.rs:53`.

PavelGuzenfeld:
**Finding 2** — added a callout under `DoraOnInputResult` in `docs/api-cxx.md`: `error` and `stop` are mutually exclusive, non-empty `error` wins. Agreed the `stop` field silently losing was worth calling out.

**Finding 3** — dropped test #7. You're right, it was just two isolated instances appending to their own Vecs behind a thread barrier; no shared state was ever exercised. The `Send+Sync` proof is already in #6.

**Finding 1** — noted. The cxx-bridge dispatch gap is consistent with the #1849 baseline. I'll open a follow-up for a smoke test once this lands.

heyong4725:
Re-reviewed `699eb0ec` + `81d93097`. Two of the three follow-up changes landed cleanly; the third has a real problem.

### ✅ `699eb0ec` — both findings addressed

**Finding #2 (`DoraOnInputResult` error/stop semantics)**: `docs/api-cxx.md:480` now has a clear blockquote callout. Reads well.

**Finding #3 (drop redundant concurrent-operators test)**: removed cleanly. Test #6 (Send+Sync compile-time check) is the right level of coverage for that property.

Filing **#1910** as a separate trackable issue rather than treating the cxx-bridge gap as a quiet TODO was a good call.

### ⚠️ `81d93097` (smoke test) — two issues

#### 1. Test SIGABRTs instead of exercising the dispatch path

Verified locally with `examples/c++-dataflow/build/liboperator_rust_api.dylib` present:

```
thread panicked at arrow-array-58.3.0/src/array/primitive_array.rs:1594:9:
assertion `left == right` failed: PrimitiveArray data should contain a single buffer only (values buffer)
  left: 0
 right: 1
panic in a function that cannot unwind
stack backtrace:
  ...
  13:        0x10b3ba220 - _dora_on_event
  14:        0x102fc5f34 - cpp_operator_cxx_bridge::cxx_bridge_dispatches_input_parse_error
                               at tests/cpp-operator-cxx-bridge.rs:128:18
```

The test's design premise — "from_ffi returns Err on `release == None` without reading the array" — doesn't hold in arrow-array 58.3.0. The `PrimitiveArray` construction inside `from_ffi` asserts `n_buffers == 1` BEFORE looking at the release field, and our zeroed array has `n_buffers == 0`. The assertion fires inside the C-export's non-unwinding boundary → `panic_cannot_unwind` → SIGABRT in `_dora_on_event`.

The InputParseError dispatch path is never reached.

To actually exercise the cxx-bridge call site, the FFI_ArrowArray needs structural fields that survive arrow-array's internal asserts (e.g. `n_buffers = 1`, valid buffer pointers) but trigger an Err return some other way. Or use a different mechanism — e.g. a schema/data type mismatch, or a release callback that returns an error.

#### 2. Test isn't wired into any CI workflow

- `cargo test --all` in `.github/workflows/ci.yml:230` passes `--exclude dora-examples`, where this test lives. So PR CI never runs it.
- Nightly's Examples job builds the shared library but doesn't then `cargo test --test cpp-operator-cxx-bridge` to consume it.
- Result: the test produces zero signal in CI as currently structured — but a SIGABRT when manually invoked with conditions present (the worst combination).

If you wire it into nightly later (after Examples → before tear-down), a working version would catch signature drift on the cxx bridge. But the test has to work first.

### Suggested path

**Option A (smaller delta to land #1904 today):** drop the smoke test (`81d93097`) from this PR + reopen #1910. The doc + drop-test changes stand alone and are merge-ready. #1910 gets a focused follow-up.

**Option B (more ambitious):** fix the FFI array construction so the test actually runs, AND add explicit CI wire-up (probably in nightly's Examples job: `cargo test --test cpp-operator-cxx-bridge -- --test-threads=1` after the cmake build step).

I'd suggest Option A — there's no rush on #1910, and getting the rest of #1904 in unblocks downstream work on the C++ operator API.

Local repro for the SIGABRT: `cargo run --example cxx-dataflow` (produces the .dylib), then `cargo test --test cpp-operator-cxx-bridge -- --nocapture`. The skip-on-missing-library guard works correctly; the failure only fires when the lib exists.

heyong4725:
Follow-up after `5cab3a13` + `d5ed417f`. The fmt fixes will get Format CI green — appreciated. But the substantive concern from my [previous comment](https://github.com/dora-rs/dora/pull/1904#issuecomment-4519827895) is still open.

`git diff 81d93097..d5ed417f` confirms both new commits are pure whitespace/import-reordering. The FFI array construction at `tests/cpp-operator-cxx-bridge.rs:96-128` (the SIGABRT site) is byte-identical to v2. Same Rust path → same `arrow-array-58.3.0/src/array/primitive_array.rs:1594` assertion → same `panic_cannot_unwind` → same SIGABRT in `_dora_on_event` when the shared library is present.

The two open issues from v2 remain:

1. **Test SIGABRTs when the shared library is present** (manual invocation after `cargo run --example cxx-dataflow`). Reproduced locally last session with the full stack trace.
2. **Test isn't wired into any CI workflow** (`cargo test --all` in `ci.yml:230` excludes `dora-examples`; no explicit `cargo test --test cpp-operator-cxx-bridge` step anywhere).

The combination is a footgun: nightly CI doesn't catch it, but anyone running the test manually (e.g. while debugging the cxx bridge) hits the abort.

Two paths forward, same as my v2 comment:

- **Option A (recommended)**: drop `81d93097` from this PR, reopen #1910 for focused follow-up. The doc + drop-test-#7 changes in `699eb0ec` stand alone and are merge-ready right now. Lets the rest of #1904 ship today.
- **Option B**: fix the FFI array construction (structurally-valid `FFI_ArrowArray` with `n_buffers = 1` but corrupt internal data, OR a different InputParseError trigger like schema mismatch), AND wire into nightly after the Examples job builds the shared library.

Which would you prefer? Happy to draft a quick PR for Option A if that helps unblock the merge.

PavelGuzenfeld:
Thanks for the precise diagnosis. Going with **Option B** — the root cause is clear and the fix is small.

**What changed** (e3bba0ce — https://github.com/PavelGuzenfeld/dora/commit/e3bba0ce):

The original test used an all-zero `FFI_ArrowArray` assuming `from_ffi` would short-circuit on `release == None`. In arrow-array 58.3.0 it doesn't: `FFI_ArrowArray::buffer()` asserts `!self.buffers.is_null()` first, and with a zeroed struct `n_buffers == 0` causes `ArrayData` to end up with 0 buffers, which `PrimitiveArray::from` then asserts should be 1 → `panic_cannot_unwind` → SIGABRT.

Per the Arrow C Data Interface, Int32 requires `n_buffers == 2` (null bitmap + values buffer). The fix provides:
- `n_buffers = 2` (passes the structural check)
- `buffers` → a valid `[null_ptr, null_ptr]` array (not the field itself null, so `buffer(i)` doesn't assert)
- `length = 1` (keeps computed values-buffer byte-size > 0)

With this layout, `create_buffer` for index 1 sees a null value + non-zero length → `Err("null buffer at position 1")` returned cleanly from `from_ffi`, which `raw::dora_on_event` converts to `Event::InputParseError` → dispatches to `ffi::on_input_parse_error`. Also removes the `mem::forget` since schema and array are independent C Data Interface objects.

The CI wiring issue (point 2 from your review) is tracked in #1910 — I'll address that there as discussed.

PavelGuzenfeld:
@heyong4725 addressed in e3bba0ce — FFI_ArrowArray now has n_buffers=2 with null buffer pointers per the Arrow C Data Interface spec for Int32, so from_ffi returns Err cleanly instead of SIGABRTing. Can you re-review?

heyong4725:
Re-reviewed `e3bba0ce`. The FFI fix works end-to-end. Verified locally.

### What the test now does

Built the cxx-dataflow example (`cargo run --example cxx-dataflow`), then `cargo test --test cpp-operator-cxx-bridge -- --nocapture`:

```
running 1 test
Rust API operator: input parse error on `test-input`: C Data interface error: The external buffer at position 1 is null.
test cxx_bridge_dispatches_input_parse_error ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

The `std::cerr` log line is the proof the cxx-bridge dispatch actually fired — it comes from `examples/c++-dataflow/operator-rust-api/operator.cc`'s `on_input_parse_error` stub. The full chain works: structurally-valid FFI_ArrowArray → `from_ffi` returns `Err("...external buffer at position 1 is null")` → `raw::dora_on_event` converts to `Event::InputParseError` → `OperatorWrapper::on_event` → `ffi::on_input_parse_error` (the cxx-bridge call site). The smoke test is now actually testing what it claims.

### One remaining gap (non-blocking)

The test still isn't wired into any CI workflow. `cargo test --all` in `.github/workflows/ci.yml:230` passes `--exclude dora-examples` (where this test lives), and nightly's Examples job builds the shared library but doesn't subsequently run `cargo test --test cpp-operator-cxx-bridge`. So in current configuration the test would only fire if someone manually invokes it after building the example.

That's not a blocker — the test exists, the skip-on-missing-library guard works correctly, and a future PR can add a single step to nightly:

```yaml
- name: Run cxx-bridge dispatch smoke test
  run: cargo test --test cpp-operator-cxx-bridge -- --test-threads=1
```

(after the Examples job step that builds the shared library)

Worth filing a follow-up issue noting this so it doesn't get lost, but doesn't need to block #1904.

### Summary

This PR is now in good shape. All three findings from the original review are addressed:

- ✅ `699eb0ec` — docs callout for `error`/`stop` semantics, dropped redundant test #7
- ✅ `e3bba0ce` — smoke test FFI construction fixed; verified the full dispatch chain works
- ⚠️ CI wiring for the smoke test is the one remaining nit — non-blocking, follow-up worthy

Ready to merge from my side.

PR #1905, opened 2026-05-21
test(coordinator,daemon,cli): tighten ws-cli-e2e assertions + node lifecycle E2E (#1760, #1703)

### LINKED ISSUE #1760: Many integration/smoke tests assert "no error" rather than actual data, leaving regressions invisible
A large fraction of our test suite verifies only that a dataflow runs to completion without entering a `Failed` state. It does **not** verify that the dataflow produced the correct outputs, the correct number of messages, in the correct order, with the correct content. As a result, a regression that silently corrupts or drops data can pass CI.

This is not a hypothetical concern — we already have explicit "contract tests" that exist *because* the underlying smoke tests don't catch correctness bugs (see [tests/example-smoke.rs#L723-L740](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L723-L740)). The contract tests cover only a handful of paths; the rest of the surface relies on liveness checks.

### 1. The smoke-test harness itself ([tests/example-smoke.rs#L167-L257](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L167-L257))

`run_smoke_test()` only checks:
- `dora up` exits 0
- `dora build` exits 0
- `dora start` exits 0
- `dora list --json` does not contain `"Failed"` and eventually no longer contains `"Running"`

It **never** captures node stdout, never inspects message content, never counts messages. Every test that calls it (~20 tests, e.g. `smoke_rust_dataflow` at [example-smoke.rs#L319](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L319), `smoke_python_echo` at [example-smoke.rs#L422](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L422), `smoke_benchmark` at [example-smoke.rs#L353](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L353)) inherits this weakness.

### 2. Fault-tolerance tests admit ambiguity ([tests/fault-tolerance-e2e.rs](https://github.com/dora-rs/dora/blob/main/tests/fault-tolerance-e2e.rs))

- `restart_recovers_from_failure` ([fault-tolerance-e2e.rs#L37-L79](https://github.com/dora-rs/dora/blob/main/tests/fault-tolerance-e2e.rs#L37-L79)) asserts `result.is_ok()` and `node_results` contains the node — but does not assert the node was actually restarted, how many times, or that downstream nodes received post-restart data.
- `max_restarts_limit_reached` ([fault-tolerance-e2e.rs#L87-L126](https://github.com/dora-rs/dora/blob/main/tests/fault-tolerance-e2e.rs#L87-L126)) literally prints `"both Ok and Err are valid"` (line 125). The test passes regardless of whether the restart budget is enforced. This is acknowledged in the file itself — `max_restarts_exhaustion_marks_node_failed` was added later (#1631) to cover the gap.

### 3. Permissive error-string assertions ([tests/ws-cli-e2e.rs#L216-L243](https://github.com/dora-rs/dora/blob/main/tests/ws-cli-e2e.rs#L216-L243))

`cli_stop_nonexistent` checks the error reply, but the assertion is:

```rust
assert!(
    msg.contains(&fake_uuid.to_string())
        || msg.to_lowercase().contains("not found")
        || !msg.is_empty(),  // <- passes for ANY non-empty error
    "error should be descriptive: {msg}"
);
```

Any non-empty error string passes. A bug returning `"internal error"` for a missing UUID would not be caught.

## Counter-examples (good patterns to expand)

These tests *do* validate data — they're the model to follow:

- **Contract tests** ([example-smoke.rs#L742-L760](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L742-L760), [example-smoke.rs#L870-L909](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L870-L909), [example-smoke.rs#L1145-L1174](https://github.com/dora-rs/dora/blob/main/tests/example-smoke.rs#L1145-L1174)): capture stdout, assert exact message counts, assert specific success markers like `"sink: SUCCESS - validated 10 doubled values"`, parse correlated request/response pairs.
- **Restart-count validation** ([fault-tolerance-e2e.rs#L214-L271](https://github.com/dora-rs/dora/blob/main/tests/fault-tolerance-e2e.rs#L214-L271)): writes a marker file from each node incarnation, then `assert_eq!(incarnations, 3)`.
- **Input-closed sequencing** ([fault-tolerance-e2e.rs#L375-L432](https://github.com/dora-rs/dora/blob/main/tests/fault-tolerance-e2e.rs#L375-L432)): asserts both `Input:value` AND `InputClosed:value` appear in order.
- **Param round-trip** ([ws-cli-e2e.rs#L415-L516](https://github.com/dora-rs/dora/blob/main/tests/ws-cli-e2e.rs#L415-L516)): `assert_eq!(value, json!(100))`, `assert_eq!(params.len(), 2)`, etc.
- **Example node tests** ([examples/rust-dataflow/node/src/tests.rs#L19-L43](https://github.com/dora-rs/dora/blob/main/examples/rust-dataflow/node/src/tests.rs#L19-L43)): golden-file comparison of full output.

## Estimated ratio

Survey of tests in `tests/`, `binaries/coordinator/tests/`, and `binaries/daemon/tests/`:

| Tier | Total | "No-error" only | Validates data |
|---|---|---|---|
| Smoke tests (`example-smoke.rs`) | ~26 | ~20 | ~6 (the contract tests) |
| Fault-tolerance E2E | ~12 | ~5 | ~7 |
| WS/CLI E2E | ~25 | ~15 | ~10 |
| Coordinator/daemon protocol unit tests | many | few | most |
| `libraries/{core,message}` unit tests | many | few | most |

The pattern: **the lower in the stack, the better the assertions**. The integration/smoke layer — which is the only layer that exercises real cross-process and cross-language flows — is where the gap lives.

## Why this matters

A dataflow framework's primary correctness contract is "the right messages reach the right inputs in the right order with the right content." Tests that only check process exit codes can pass while:

- Messages are silently dropped (e.g. shared-memory fallback to heap on resource exhaustion)
- Message ordering is broken across restarts
- Cross-language type marshalling drops/coerces fields
- Service `request_id` correlation is mis-routed
- Duplicate delivery occurs after a daemon reconnect

## Proposed approach

1. Extend `run_smoke_test` (or add `run_smoke_test_validated`) to capture node stdout and accept an `assertions: &[&str]` of required substrings or a closure.
2. For each existing example dataflow, add a sink node (or stdout marker) that prints a final `SUCCESS - validated N <thing>` line, mirroring the contract-test pattern. Convert the existing smoke test to assert on the marker.
3. Tighten the loose assertions called out above (`!msg.is_empty()`, "both Ok and Err are valid") into specific expected outcomes; if both outcomes really are valid, split into two tests.
4. Add a TDD checklist item to `CLAUDE.md` / `docs/testing-guide.md`: any new dataflow feature requires at least one assertion on the *content* of a message, not just on its absence of error.

This is a multi-PR effort. Suggest starting with (1) + (2) on 3-5 high-traffic examples (`rust-dataflow`, `python-dataflow`, `cxx-dataflow`, `multiple-daemons`, one ROS2-bridge example) as a template, then expanding.

---

*Filed by Claude on behalf of @phil-opp.*

PavelGuzenfeld:
Closes #1760. Adds coverage for #1703 (node stop/start/restart lifecycle).

**#1760 — tighten error-path assertions (6 tests)**
- `cli_stop_nonexistent`: drops `|| !msg.is_empty()` fallback; error must name the unknown UUID
- `cli_restart_node_nonexistent`, `cli_stop_node_nonexistent`: were variant-only; now assert UUID is present in message
- `cli_param_set_rejects_unknown_target`, `cli_param_get_nonexistent`, `cli_param_delete_rejects_unknown_target`: same pattern

**#1703 — E2E coverage for `dora node list/info/stop/restart` (4 tests + fixture)**
- `e2e_node_list_shows_running_nodes`: verifies all three nodes appear in `dora node list` output
- `e2e_node_info_returns_descriptor`: checks `dora node info --format json` round-trips the descriptor
- `e2e_node_stop_exits_cleanly`: `dora node stop` exits 0 and node disappears from list
- `e2e_node_restart_increments_counter`: `dora node restart` on a `restart_policy: always` node; waits for re-appearance in list
- `tests/dataflows/node-lifecycle.yml`: 1 Hz tick dataflow (~100 s runtime), `rust-node` has `restart_policy: always` so restart actually re-spawns it

All 10 tests pass locally (docker, `cargo test -p dora-examples --test ws-cli-e2e`).

### DIFF
--- a/tests/dataflows/node-lifecycle.yml
+++ b/tests/dataflows/node-lifecycle.yml
@@ -6,7 +6,7 @@ nodes:
   - id: rust-node
     build: cargo build -p rust-dataflow-example-node
     path: ../../target/debug/rust-dataflow-example-node
-    restart_policy: always
+    restart_policy: never
     inputs:
       tick: dora/timer/secs/1
     outputs:
--- a/tests/dataflows/node-lifecycle.yml
+++ b/tests/dataflows/node-lifecycle.yml
@@ -1,0 +1,28 @@
+# Long-running dataflow for `dora node list/info/stop/restart` E2E tests.
+# Uses a 1-second tick so the rust-node runs for ~100 seconds, giving
+# tests enough time to issue node-level commands before the dataflow ends.
+
+nodes:
+  - id: rust-node
+    build: cargo build -p rust-dataflow-example-node
+    path: ../../target/debug/rust-dataflow-example-node
+    restart_policy: always
+    inputs:
+      tick: dora/timer/secs/1
+    outputs:
+      - random
+
+  - id: rust-status-node
+    build: cargo build -p rust-dataflow-example-status-node
+    path: ../../target/debug/rust-dataflow-example-status-node
+    inputs:
+      tick: dora/timer/millis/500
+      random: rust-node/random
+    outputs:
+      - status
+
+  - id: rust-sink
+    build: cargo build -p rust-dataflow-example-sink
+    path: ../../target/debug/rust-dataflow-example-sink
+    inputs:
+      message: rust-status-node/status
--- a/tests/ws-cli-e2e.rs
+++ b/tests/ws-cli-e2e.rs
@@ -454,10 +454,8 @@ fn cli_stop_nonexistent() {
     match reply {
         ControlRequestReply::Error(msg) => {
             assert!(
-                msg.contains(&fake_uuid.to_string())
-                    || msg.to_lowercase().contains("not found")
-                    || !msg.is_empty(),
-                "error should be descriptive: {msg}"
+                msg.contains(&fake_uuid.to_string()) || msg.to_lowercase().contains("not found"),
+                "error must name the unknown UUID {fake_uuid}: {msg}"
             );
         }
         other => panic!("expected Error, got {other:?}"),
@@ -522,20 +520,27 @@ fn cli_restart_node_nonexistent() {
     let addr: SocketAddr = format!("127.0.0.1:{port}").parse().unwrap();
     let session = WsSession::connect(addr).expect("failed to connect WsSession");
 
+    let fake_dataflow_id = uuid::Uuid::new_v4();
     let reply = send_request(
         &session,
         &ControlRequest::RestartNode {
-            dataflow_id: uuid::Uuid::new_v4(),
+            dataflow_id: fake_dataflow_id,
             node_id: "camera".to_string().into(),
             grace_duration: None,
         },
     )
     .unwrap();
 
-    assert!(
-        matches!(reply, ControlRequestReply::Error(_)),
-        "expected Error, got {reply:?}"
-    );
+    match reply {
+        ControlRequestReply::Error(msg) => {
+            assert!(
+                msg.contains(&fake_dataflow_id.to_string())
+                    || msg.to_lowercase().contains("not found"),
+                "error must name the unknown dataflow {fake_dataflow_id}: {msg}"
+            );
+        }
+        other => panic!("expected Error, got {other:?}"),
+    }
 }
 
 #[test]
@@ -544,20 +549,27 @@ fn cli_stop_node_nonexistent() {
     let addr: SocketAddr = format!("127.0.0.1:{port}").parse().unwrap();
     let session = WsSession::connect(addr).expect("failed to connect WsSession");
 
+    let fake_dataflow_id = uuid::Uuid::new_v4();
     let reply = send_request(
         &session,
         &ControlRequest::StopNode {
-            dataflow_id: uuid::Uuid::new_v4(),
+            dataflow_id: fake_dataflow_id,
             node_id: "sensor".to_string().into(),
             grace_duration: None,
         },
     )
     .unwrap();
 
-    assert!(
-        matches!(reply, ControlRequestReply::Error(_)),
-        "expected Error, got {reply:?}"
-    );
+    match reply {
+        ControlRequestReply::Error(msg) => {
+            assert!(
+                msg.contains(&fake_dataflow_id.to_string())
+                    || msg.to_lowercase().contains("not found"),
+                "error must name the unknown dataflow {fake_dataflow_id}: {msg}"
+            );
+        }
+        other => panic!("expected Error, got {other:?}"),
+    }
 }
 
 // -- Phase 3: Parameter operations via WsSession (E2E) --
@@ -581,10 +593,15 @@ fn cli_param_set_rejects_unknown_target() {
         },
     )
     .unwrap();
-    assert!(
-        matches!(reply, ControlRequestReply::Error(_)),
-        "expected Error for unknown param target, got {reply:?}"
-    );
+    match reply {
+        ControlRequestReply::Error(msg) => {
+            assert!(
+                msg.contains(&df_id.to_string()) || msg.to_lowercase().contains("not found"),
+                "error must name the unknown dataflow {df_id}: {msg}"
+            );
+        }
+        other => panic!("expected Error for unknown param target, got {other:?}"),
+    }
 }
 
 #[test]
@@ -593,20 +610,26 @@ fn cli_param_get_nonexistent() {
     let addr: SocketAddr = format!("127.0.0.1:{port}").parse().unwrap();
     let session = WsSession::connect(addr).expect("failed to connect WsSession");
 
+    let df_id = uuid::Uuid::new_v4();
     let reply = send_request(
         &session,
         &ControlRequest::GetParam {
-            dataflow_id: uuid::Uuid::new_v4(),
+            dataflow_id: df_id,
             node_id: "ghost".to_string().into(),
             key: "missing".into(),
         },
     )
     .unwrap();
 
-    assert!(
-        matches!(reply, ControlRequestReply::Error(_)),
-        "expected Error for missing param, got {reply:?}"
-    );
+    match reply {
+        ControlRequestReply::Error(msg) => {
+            assert!(
+                msg.contains(&df_id.to_string()) || msg.to_lowercase().contains("not found"),
+                "error must name the unknown dataflow {df_id}: {msg}"
+            );
+        }
+        other => panic!("expected Error for missing param, got {other:?}"),
+    }
 }
 
 #[test]
@@ -627,10 +650,15 @@ fn cli_param_delete_rejects_unknown_target() {
         },
     )
     .unwrap();
-    assert!(
-        matches!(reply, ControlRequestReply::Error(_)),
-        "expected Error for unknown param target, got {reply:?}"
-    );
+    match reply {
+        ControlRequestReply::Error(msg) => {
+            assert!(
+                msg.contains(&df_id.to_string()) || msg.to_lowercase().contains("not found"),
+                "error must name the unknown dataflow {df_id}: {msg}"
+            );
+        }
+        other => panic!("expected Error for unknown param target, got {other:?}"),
+    }
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


heyong4725:
Did a pre-landing review pass. Important context first, then five findings.

### Context worth knowing

PR #1903 merged ~4.5h after you opened this (so unavoidable parallelism, no fault). It also addresses #1703 and ended up a superset of your #1703 half:
- It added `tests/node-lifecycle-e2e.rs` covering **all 8** `dora node` subcommands (list/info/add/connect/disconnect/remove/restart/stop) across **3 language fixtures** (Python/Rust/C++).
- It also landed a daemon fix at `binaries/daemon/src/running_dataflow.rs:75-95` — a `force_restart_next: Arc<AtomicBool>` flag that makes `dora node restart` work for `restart_policy: never` nodes (pre-fix: silent no-op on never-policy nodes).

That second point is what makes finding #1 below interesting.

### Half 1 (#1760, the assertion tightening) is clean

Six assertions go from "matches Error variant" to "Error message names the unknown UUID OR contains `not found`". Tight, independent, valuable. No concerns on that half.

### Half 2 (#1703, the lifecycle E2E) has five findings, one critical

**1. [CRITICAL] `e2e_node_restart_increments_counter` cannot fail.** `tests/dataflows/node-lifecycle.yml:9` sets `rust-node` to `restart_policy: always`. The daemon's restart loop respawns on any exit when the policy is `always`. So:
- If `dora node restart` works → node killed, daemon respawns it → node in list. ✓
- If `dora node restart` is a complete no-op → node never exits (1Hz tick keeps it alive) → still in list. ✓
- Either way the test passes. The test cannot distinguish the two paths.

The fix that PR #1903 landed (`force_restart_next` flag) was specifically to make `dora node restart` actually fire on `restart_policy: never` nodes. To catch a regression of that fix, your test needs the same setup:

```yaml
- id: rust-node
  ...
  restart_policy: never   # was: always
```

With `never`, the only way `rust-node` reappears in `dora node list` after `dora node restart` is if the command actually killed-then-respawned it via the `force_restart_next` path. That's a falsifiable test.

**2. [HIGH] `start_lifecycle_dataflow_detached` picks `list.0.first()` — wrong UUID on stale state.** `tests/ws-cli-e2e.rs:1403-1413` — if a prior test's `cleanup()` leaves a stale dataflow registered (the 500ms sleeps in `cleanup` aren't always enough on slow CI runners), this returns the stale entry's UUID and every subsequent assertion silently operates on the wrong dataflow. Failures look like "node not found" with no hint that you grabbed the wrong dataflow.

Fix: filter by name (give the dataflow a known name via `--name X` on `dora start`, then `list.0.iter().find(|e| e.name == "X")`), or assert the list is empty before `dora start` and pick the new entry afterwards.

**3. [HIGH] `e2e_node_stop_exits_cleanly` stdout assertion is too loose.** `tests/ws-cli-e2e.rs:1532-1535`:
```rust
stdout.contains("stopped") || stdout.contains("rust-sink"),
```
The OR makes "rust-sink" alone pass. Any error message naming the node ("failed to stop node rust-sink: ...") satisfies the assertion. The exit-code check is what's actually validating success here; the stdout check adds nothing. Suggest either ANDing both terms or asserting the exact success line `dora node stop` prints.

**4. [HIGH] No RAII cleanup on test panic.** If any assertion between `start_cluster` and the final `cleanup` panics, coordinator + daemon leak. Since all `real_dataflow` tests share port 6013, the leak cascades into "address in use" on every subsequent test, hiding the original failure. PR #1903's `tests/node-lifecycle-e2e.rs` solved this with a `CleanupGuard` struct that calls `cleanup_stale(dora)` in `Drop` — worth copying that pattern. Code shape:

```rust
struct ClusterGuard<'a> { dora: &'a str }
impl<'a> Drop for ClusterGuard<'a> {
    fn drop(&mut self) { cleanup(self.dora); }
}
// usage: let _guard = ClusterGuard { dora: &dora };
```

**5. [INFO] 15s "all 3 nodes registered" deadline is tight on cold CI.** `tests/ws-cli-e2e.rs:1420-1442` — uses `dora node info` as the per-node readiness probe. On a cold CI cache, the daemon itself can take 5-6s to be ready to answer `info` calls, leaving ~30 iterations of 300ms polling. Works in practice but spuriously fails when runners are under load. Bump to 30s, or probe daemon readiness directly before dataflow start.

### Reasonable next steps

Given the #1903 overlap, three sensible paths depending on what value you want this PR to add:

- **Keep both halves, fix #1-#4**: the ws-cli-e2e layer coverage is lower-level than #1903's subprocess fixtures and genuinely complementary. But the restart test needs the `never` policy fix to be useful.
- **Drop the #1703 half, keep #1760**: cleanest given that #1903's coverage is broader. The assertion tightening is high-value on its own.
- **Drop the restart test specifically, keep list/info/stop**: list/info/stop add real ws-cli-e2e signal; restart is the broken one.

Happy to help on any of those paths if you want.

PavelGuzenfeld:
Good catches — addressed all four blocking findings in the latest push.

**#1 (restart test can't fail)** — changed `restart_policy: always` → `never` in `node-lifecycle.yml`. With `never`, the only way `rust-node` reappears in `dora node list` is if the `force_restart_next` path actually fired. The test is now falsifiable.

**#2 (wrong UUID on stale state)** — added a pre-start assertion that the coordinator list is empty before `dora start`. If a stale entry exists it fails loudly with the UUID(s) rather than silently operating on the wrong dataflow.

**#3 (stop assertion too loose)** — changed `||` to `&&`. An error message naming the node can no longer satisfy it.

**#4 (no RAII cleanup)** — added `ClusterGuard` with `Drop → cleanup`. All four tests get `let _guard = ClusterGuard(&dora)` immediately after `start_cluster`, so coordinator+daemon are torn down even on panic.

**#5 (15s deadline)** — bumped to 30s.

Re the `#1903` overlap: agreed this PR's ws-cli-e2e layer is complementary to the subprocess fixtures in `#1903`. Keeping both halves.

heyong4725:
Re-reviewed `16f7bd91` + `711d539a`. Verified each fix against the code:

| # | Original concern | Fix |
|---|---|---|
| 1 | restart test non-falsifiable (policy=always auto-respawned) | `tests/dataflows/node-lifecycle.yml`: `restart_policy: never`. Now if `force_restart_next` regresses, `rust-node` stays gone and the poll deadline fires. Falsifiable. |
| 2 | `list.0.first()` picks wrong UUID on stale state | `start_lifecycle_dataflow_detached` at `:1398-1412` asserts the coordinator list is empty before start; fails loudly with the stale UUID(s) if anything leaked through. |
| 3 | Stop assertion `\|\|` too loose | `:1568` switched to `&&` — both "stopped" AND "rust-sink" required. |
| 4 | No RAII cleanup on panic | `ClusterGuard<'a>(&'a str)` + `impl Drop → cleanup` at `:1083-1088`, applied via `let _guard = ClusterGuard(&dora)` in all 4 lifecycle tests. Coordinator/daemon tear down even on assertion panic. |
| 5 | 15s deadline tight on cold CI | Bumped to 30s at `:1445`. |

Local verification: `cargo test --test ws-cli-e2e real_dataflow::e2e_node -- --test-threads=1` → **4/4 pass in 84s**, including the now-falsifiable restart test.

Ship it from my side. Thanks for the clean iteration — the ws-cli-e2e layer coverage is a real complement to PR #1903's subprocess fixtures.

PR #1906, opened 2026-05-21
fix(cli): add root pyproject.toml workspace to Python dataflow template (closes #1844)

### LINKED ISSUE #1844: template: `dora new --lang python` projects fail `uv run pytest` with ModuleNotFoundError
## Summary

`dora new --lang python` generates a project where `uv run pytest` from the project root fails with `ModuleNotFoundError` for every node package. This is the same root cause as the cli-python CI failure being patched in [#1843](https://github.com/dora-rs/dora/pull/1843), but the CI fix is intentionally scoped to "unblock the merge queue" rather than fixing the underlying template. Filing this so the template-level fix doesn't get lost.

## Reproduction

On a clean checkout (after [#1843](https://github.com/dora-rs/dora/pull/1843) lands or with the CI fix patched locally is irrelevant — the issue is in the generated project):

```bash
dora new my_project --lang python
cd my_project
dora build dataflow.yml --uv
uv run pytest
```

Result:

```
FAILED listener-1/tests/test_listener_1.py::test_import_main
  > from listener_1.main import main
  E ModuleNotFoundError: No module named 'listener_1'
FAILED talker-1/tests/test_talker_1.py::test_import_main
  > from talker_1.main import main
  E ModuleNotFoundError: No module named 'talker_1'
FAILED talker-2/tests/test_talker_2.py::test_import_main
  > from talker_2.main import main
  E ModuleNotFoundError: No module named 'talker_2'
3 failed, 3 passed
```

## Root cause

PR #1820 ("feat(uv): per-node managed Python venvs at build + runtime") changed `dora build --uv` to install each node's package into an isolated venv at `.dora/python-envs/<node>/`. The test stubs generated by the template at `binaries/cli/src/template/python/__node-name__/tests/test___node_name__.py:11` do:

```python
from __node_name__.main import main
```

`uv run pytest` at the project root resolves against the ambient venv, which doesn't have the per-node packages (they're in the isolated venvs). The generated project has no root `pyproject.toml` to give `uv` a way to install them.

Pre-#1820, `dora build` installed nodes into whatever venv was active, so this Just Worked.

## How #1843 patches it for CI

`.circleci/config.yml` explicitly installs the three known template nodes into the active venv before `uv run pytest`:

```bash
uv pip install -e listener-1 -e talker-1 -e talker-2
uv run pytest
```

That fixes the CI smoke but doesn't help users who run `dora new` themselves. They hit the same error and have no obvious way to recover from the failure message.

## Proposed approaches (pick one)

1. **Root `pyproject.toml` in the generated project, with path-deps on each node.** Makes `uv sync` install everything; `uv run pytest` then resolves all node packages. Closest to "user expectations for a Python project." Requires teaching `dora new --lang python` to emit a root `pyproject.toml` listing the generated nodes as dependencies (`{ path = "./talker-1", editable = true }` etc.).

2. **Generated README that documents the test workflow.** Add a "Testing" section to `binaries/cli/src/template/python/README.md` that says: "Before running `uv run pytest`, install each node into the active venv with `uv pip install -e <node>/`." Lower-effort, but punishes users who skim docs.

3. **Run pytest inside each per-node venv.** Architecturally cleanest (tests run in the same env as the node) but requires a wrapper script and changes the pytest UX. Probably overkill for the template.

4. **A generated `Makefile` or shell script** that handles `dora build` + per-node install + pytest in one command. Avoids the bare `uv run pytest` UX entirely.

My pick would be (1) — a root `pyproject.toml` is the path of least surprise for anyone arriving from the broader Python ecosystem, and `uv` is already a hard dep for the template's build flow.

## Scope

This is a `dora new --lang python` template concern, in `binaries/cli/src/template/python/`. Not blocked by anything; #1843 unblocks CI separately. Could be picked up as a good-first-issue.

PavelGuzenfeld:
`dora new --lang python` now emits a root `pyproject.toml` that declares a uv workspace with all three generated nodes as members. `uv run pytest` syncs the workspace first so all node packages are importable without a manual `uv pip install` step.

Root cause: #1820 switched `dora build --uv` to per-node isolated venvs, but the template had no workspace root, so `uv run pytest` from the project directory couldn't find the node packages.

When `--internal-create-with-path-dependencies` is set (CI only), a `[tool.uv.sources]` block pins `dora-rs` to `../apis/python/node` so the local dev version is used. Removes the redundant `uv pip install -e talker-1 -e ...` workaround from CircleCI and GHA nightly that was added in #1843.


heyong4725:
Reviewed end-to-end. This is a clean targeted fix for #1844 — the root `pyproject.toml` with `[tool.uv.workspace] members = ["talker-1", "talker-2", "listener-1"]` is exactly the missing piece that lets `uv run pytest` discover the three generated node packages without a manual `uv pip install -e ...` step.

Verified:
- `___name___` placeholder follows the existing template convention (also used in `binaries/cli/src/template/rust/*Cargo-template.toml` and the per-node Python pyproject)
- Workspace member directory names match `create_custom_node`'s `root.join("talker-1")` / `"talker-2"` / `"listener-1"`
- Per-node `name = "talker-1"` (via `replace_space`) → workspace resolves the member correctly
- `[tool.uv.sources]` block is correctly gated on `use_path_deps` so the `../apis/python/node` pin only ships in CI's `--internal-create-with-path-dependencies` mode, not user-facing `dora new`
- CircleCI + GHA nightly workaround removals are symmetric and the comment explains the new mechanism

One minor uv-idiom nit (non-blocking): the workspace root has `[project]` without `[build-system]`. uv treats `[tool.uv.workspace]` roots without `[build-system]` as virtual roots, so this works today, but the strictly idiomatic options are either (a) drop `[project]` entirely and use only `[tool.uv.workspace]`, or (b) add `[build-system]`. Not worth holding the PR for.

Thanks for picking this one up — and for the clean commit message that explained the #1820 → #1843 → #1844 chain of cause + workaround + real fix.

PavelGuzenfeld:
done, went with (a) — dropped `[project]` entirely in 9081f69b

PR #1907, opened 2026-05-21
fix(core): replace expect with contextual errors on stdout/stderr pipe capture (closes #1499)

### LINKED ISSUE #1499: Return pipe capture failures as errors instead of panicking

PavelGuzenfeld:
Two `expect` calls in `prepare_managed_python_env` panic on missing stdout/stderr pipes instead of returning an error. Since the function already returns `eyre::Result<()>`, replace them with `ok_or_else(|| eyre!(...))` so a missing pipe propagates like any other spawn failure.

Only the two lines in `libraries/core/src/build/build_command.rs:198-199` are affected. The daemon spawn path (`binaries/daemon/src/spawn/prepared.rs`) has no equivalent issue.

### DIFF
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -195,8 +195,18 @@ pub async fn prepare_managed_python_env(
             .spawn()
             .wrap_err_with(|| format!("failed to spawn `uv venv {}`", python_env_dir.display()))?;
 
-        let child_stdout = BufReader::new(child.stdout.take().expect("failed to take stdout"));
-        let child_stderr = BufReader::new(child.stderr.take().expect("failed to take stderr"));
+        let child_stdout = BufReader::new(
+            child
+                .stdout
+                .take()
+                .ok_or_else(|| eyre!("failed to take stdout pipe from uv venv child"))?,
+        );
+        let child_stderr = BufReader::new(
+            child
+                .stderr
+                .take()
+                .ok_or_else(|| eyre!("failed to take stderr pipe from uv venv child"))?,
+        );
         let stdout_tx_clone = stdout_tx.clone();
 
         tokio::spawn(async move {


heyong4725:
Reviewed. Clean targeted fix — replacing two `.expect()` panics with `.ok_or_else(|| eyre!(...))?` is exactly the right shape for a function that already returns `eyre::Result<()>`.

Verified:
- Function signature at `libraries/core/src/build/build_command.rs:184` is `-> eyre::Result<()>`, so `?` propagation works
- `use eyre::{Context, eyre};` already in scope at the top of the file
- The PR's claim that the daemon spawn path has no equivalent issue is accurate: `binaries/daemon/src/spawn/prepared.rs:508` and `:571` already use `.ok_or_else(|| eyre::eyre!("failed to take stdout"))?` — daemon was already correct. Closes the full scope of #1499.

Small bonus: the new error messages (`"failed to take stdout pipe from uv venv child"`) carry more context than the existing daemon equivalents (`"failed to take stdout"`). Future debugging will benefit. If you ever want to tighten the daemon messages too, they're at `prepared.rs:508` and `:571`, but that's out of scope here.

Thanks for picking this up + for verifying the daemon path before claiming "no equivalent issue" in the PR description.

PR #1908, opened 2026-05-21
feat(python): async iteration support for MockNode + 3 async tests (closes #1557)

### LINKED ISSUE #1557: [bug] dora async Python nodes have no test coverage — async event loop interaction with GIL is historically deadlock-prone
## Problem

The new async Python API introduces a second dimension of threading complexity: asyncio 
event loops interacting with the Rust-backed event stream:

- There are no `pytest-asyncio` tests for the `async for event in node:` pattern.
- There is no test verifying that `await node.send_output(...)` does not deadlock when 
  called from a coroutine that is itself being driven by a daemon-connected event stream.
- There is no test for the interaction between `asyncio.sleep()`, `async node iteration`, 
  and the `STOP` event.

This is especially dangerous because:
1. Async/GIL deadlocks are notoriously hard to reproduce deterministically.
2. They often only manifest under load or with specific Python versions.
3. The fix for the previous GIL deadlock was non-trivial.

## Proposed Fix

Add a `tests/test_async_node.py` using `MockNode` (see Issue #XXXX) and `pytest-asyncio`:
```python
import pytest
import asyncio
import pyarrow as pa
from dora.testing import AsyncMockNode

@pytest.mark.asyncio
async def test_async_node_receives_input():
    async with AsyncMockNode([("tick", pa.array([0]))]) as node:
        async for event in node:
            if event["type"] == "INPUT":
                assert event["id"] == "tick"
                return
    pytest.fail("No input received")

@pytest.mark.asyncio
async def test_async_node_send_output_does_not_deadlock():
    async with AsyncMockNode([("tick", pa.array([0]))]) as node:
        async for event in node:
            if event["type"] == "INPUT":
                # This must complete within 1 second — no deadlock
                await asyncio.wait_for(
                    node.send_output("result", pa.array([42])), 
                    timeout=1.0
                )
                return
```

PavelGuzenfeld:
Adds `__aiter__`/`__anext__` to `MockNode` so it supports `async for event in node:` — the natural async iteration pattern used with the real `dora.Node` async API.

Three new tests added to `test_mock_node.py`:

- `test_mock_node_async_iteration`: `async for event in node:` collects INPUT events then STOP
- `test_mock_node_recv_async_loop`: `while True: event = await recv_async()` loop with `asyncio.wait_for` timeout (1 s) as deadlock guard — pins the canonical dora async pattern
- `test_mock_node_send_output_in_async_context`: `send_output` followed by `await asyncio.sleep(0)` verifies no lock is held across event-loop yields

All 11 tests pass. No new dependencies (uses stdlib `asyncio` only).

### DIFF
--- a/apis/python/node/dora/testing.py
+++ b/apis/python/node/dora/testing.py
@@ -79,15 +79,6 @@ class MockNode:
         """
         return self.next(timeout=timeout)
 
-    def __aiter__(self):
-        return self
-
-    async def __anext__(self) -> dict[str, Any]:
-        event = self.next()
-        if event is None:
-            raise StopAsyncIteration
-        return event
-
     def send_output(
         self,
         output_id: str,
--- a/apis/python/node/dora/testing.py
+++ b/apis/python/node/dora/testing.py
@@ -79,6 +79,15 @@ class MockNode:
         """
         return self.next(timeout=timeout)
 
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self) -> dict[str, Any]:
+        event = self.next()
+        if event is None:
+            raise StopAsyncIteration
+        return event
+
     def send_output(
         self,
         output_id: str,
--- a/apis/python/node/tests/test_mock_node.py
+++ b/apis/python/node/tests/test_mock_node.py
@@ -85,3 +85,52 @@ def test_mock_node_recv_async():
         assert event is None
 
     asyncio.run(run())
+
+
+def test_mock_node_async_iteration():
+    """async for event in node: mirrors the pattern used with the real dora.Node async API."""
+
+    async def run():
+        node = MockNode([("tick", pa.array([1])), ("tick", pa.array([2]))])
+        events = []
+        async for event in node:
+            events.append(event)
+        assert len(events) == 3
+        assert events[0]["type"] == "INPUT"
+        assert events[0]["value"].to_pylist() == [1]
+        assert events[1]["type"] == "INPUT"
+        assert events[1]["value"].to_pylist() == [2]
+        assert events[2]["type"] == "STOP"
+
+    asyncio.run(run())
+
+
+def test_mock_node_recv_async_loop():
+    """while/recv_async loop — the canonical dora async node pattern.  Timeout guards against deadlock."""
+
+    async def run():
+        node = MockNode([("a", pa.array([10])), ("b", pa.array([20]))])
+        seen_inputs = []
+        while True:
+            event = await asyncio.wait_for(node.recv_async(), timeout=1.0)
+            if event is None or event["type"] == "STOP":
+                break
+            seen_inputs.append(event["id"])
+        assert seen_inputs == ["a", "b"]
+
+    asyncio.run(run())
+
+
+def test_mock_node_send_output_in_async_context():
+    """send_output called from an async coroutine must complete without blocking."""
+
+    async def run():
+        node = MockNode([("tick", pa.array([0]))])
+        async for event in node:
+            if event["type"] == "INPUT":
+                # send_output is sync but must not block the event loop
+                node.send_output("result", pa.array([42]))
+                await asyncio.sleep(0)  # yield to event loop — if any lock is held, this deadlocks
+        assert node.outputs["result"][0].to_pylist() == [42]
+
+    asyncio.run(run())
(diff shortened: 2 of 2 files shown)


heyong4725:
Did a pre-landing review pass. One concern that I think should block merge as-is, and one nit on the test framing.

### The PR's central claim doesn't hold against the real `dora.Node`

The PR description says `async for event in node:` is "the natural async iteration pattern used with the real `dora.Node` async API." I verified the pyo3 binding and that's not the case:

- `apis/python/node/src/lib.rs:283` exposes `__iter__` + `__next__` (sync iteration).
- `apis/python/node/src/lib.rs:302` exposes `recv_async(timeout)` (async method).
- The real `dora.Node` does **not** expose `__aiter__` or `__anext__`. So `async for event in node:` raises `TypeError: 'Node' object is not an async iterator` against production code.
- The dora docs in that same file (`lib.rs:286-296`) show the canonical async pattern as a `while True: event = await node.recv_async()` loop — exactly what test #2 in this PR exercises.
- Grep across the whole repo confirms: the only `async for ... in node:` pattern in the codebase is the new tests added here. No production code uses this.

`MockNode`'s docstring at `apis/python/node/dora/testing.py:28` says: *"Drop-in replacement for dora.Node usable in unit tests."* Adding `__aiter__`/`__anext__` only to the mock breaks that — a test written against `MockNode` using `async for event in node:` cannot port to production code with the real `dora.Node`. The test gives false confidence.

Two reasonable resolutions:

1. **Drop `__aiter__`/`__anext__` from MockNode + drop test #1 and the `async for` body in test #3.** Keep test #2 (`recv_async` loop) — that's the documented real pattern. Smallest fix, restores the drop-in-replacement invariant.

2. **Land a separate PR adding `__aiter__`/`__anext__` to the real `dora.Node`** in the pyo3 binding first, then mirror it in MockNode. Larger scope, but legitimizes the `async for` pattern as a real API.

Option 1 is the surgical fix and probably the right call for this PR.

### Test #3's "deadlock guard" is theatrical

`apis/python/node/tests/test_mock_node.py:130-131`:

```python
node.send_output("result", pa.array([42]))
await asyncio.sleep(0)  # yield to event loop — if any lock is held, this deadlocks
```

`MockNode.send_output` is pure Python `list.append` — `self.outputs.setdefault(output_id, []).append(data)`. There are no locks anywhere in MockNode. The `if any lock is held, this deadlocks` framing is impossible to falsify against the mock, so it overpromises what the test proves.

The underlying issue (real Rust event stream + GIL deadlocks) is what #1557 cared about, but MockNode can't actually test it. Honest rephrase would be something like: "verify `send_output` is synchronous and the event loop can yield around it" — narrower claim, accurate.

### Issue #1557's deeper concern

The issue called out three risks:
1. No tests for `async for event in node:` (PR adds this against the mock)
2. No test that `send_output` doesn't deadlock from a coroutine (mock can't test deadlock)
3. No test for `asyncio.sleep + async iteration + STOP` interaction (PR exercises with the mock)

The issue's "Proposed Fix" used MockNode, so the contributor was working from the spec. But because MockNode is pure-Python, none of these tests actually exercise the Rust event stream / GIL interactions that originally motivated #1557. Worth a follow-up issue noting that mock-level coverage isn't sufficient to close the underlying safety concern — real-daemon integration tests would be needed for that.

### What I'd suggest

Land this PR after applying option 1 above (drop the `async for` pieces, keep the `recv_async` loop test). That gives you immediate value (the canonical pattern is now tested) without exposing API on the mock that doesn't exist on the real Node. The wider question of testing real Rust async paths can become a follow-up.

Happy to help refactor if useful.

PavelGuzenfeld:
applied option 1 in 137b8753 — dropped `__aiter__`/`__anext__` from MockNode, removed the `async for` test, rewrote test #3 to use the canonical `recv_async` loop, and softened the deadlock comment to just 'yield to let other coroutines run'

heyong4725:
Re-reviewed `137b8753`. This addresses all three concerns from my earlier pass cleanly:

1. `__aiter__`/`__anext__` removed from `MockNode` — `apis/python/node/dora/testing.py` is back to its original state, restoring the "drop-in replacement for `dora.Node`" docstring invariant.
2. Tests now use the canonical `while True: event = await node.recv_async()` pattern that matches dora.Node's documented async API (per `apis/python/node/src/lib.rs:286-296`). Tests written against MockNode using this pattern will port to production code with the real Node.
3. Test #3's comment is now accurate ("send_output called from an async coroutine completes synchronously") instead of overpromising what the mock can prove.

Verified locally — `uv run pytest tests/test_mock_node.py -v` reports **10 passed in 3.80s** (8 existing + 2 new async tests).

Thanks for the iteration. Ready to land from my side.

PR #1909, opened 2026-05-22
fix(daemon): treat SIGTERM-induced exit during planned stop as clean (closes #1882)

### LINKED ISSUE #1882: nightly: CLI Tests (macos-latest) — python-dataflow sender exits 143 on --stop-after

heyong4725:
Closes #1882.

## Summary

When `dora run --stop-after`, `dora stop`, `dora destroy`, or `dora node stop` triggers a planned shutdown, the daemon sends a cooperative `Event::Stop` over the event stream, then escalates to SIGTERM after the grace period if the node hasn't exited. The resulting non-zero exit was reaching `binaries/daemon/src/lib.rs:3664` (`SpawnedNodeResult` handler) and falling into `NodeErrorCause::Other` or `NodeErrorCause::GraceDuration`, producing an `Err(NodeError)` that bubbled up to `FormatDataflowError` as:

```
Dataflow X failed: Node `sender` failed: exited with code 143
```

This was the macOS nightly's `CLI Tests` failure for `python-dataflow --stop-after 10s`. The sender's 100 messages × 100ms loop runs ~12s on macOS (Python startup tax) vs ~10s on Linux. With `--stop-after 10s`, macOS gets SIGTERMed mid-loop; Linux just barely finishes first. Same underlying bug, masked on Linux by timing.

## The fix

In the `SpawnedNodeResult` arm, when `restarts_disabled` is true (set by `handle_dataflow_stop` / `stop_single_node`) AND `exit_status` is one of the SIGTERM-like variants AND there's no cascading error, report `Ok(())`. Cascading failures still win — we don't want a real upstream crash hidden by the shutdown that followed.

`grace_duration_kills` deliberately doesn't gate the new branch: `running_dataflow.rs::stop_all` populates it at SoftKill submission time, not at the hard-kill timeout, so it overlaps with every operator-requested stop. The branch comment documents this.

What `is_sigterm_like` catches:
- `Signal(15)` — SIGTERM, direct-spawned nodes
- `Signal(2)` — SIGINT, Ctrl-C path
- `ExitCode(143)` — 128+SIGTERM, wrapper-spawned (`uv run python` catches signal, exits with code)
- `ExitCode(130)` — 128+SIGINT, wrapper-spawned

## Test

`planned_stop_sigterm_reports_clean` in `tests/fault-tolerance-e2e.rs` uses a new `fire-and-forget-source-node` fixture: a producer that never calls `events.recv()`, so it physically CANNOT honor the cooperative `Event::Stop` — the daemon's only termination path is SIGTERM after grace. Pre-fix it fails with `Err(NodeError { cause: GraceDuration, exit_status: Signal(15) })`; post-fix it reports `Ok(())`.

I deliberately introduced the test before the second iteration of the fix and observed the failure mode (`GraceDuration` instead of the expected `Other`) — that informed the final exclusion-of-`!grace_duration_kill` decision documented in the daemon comment.

## Risk

A node that legitimately exits with `ExitCode(143)` during operator-requested stop (e.g. `sys.exit(143)` from Python) is silently marked clean. In practice that's a contrived case — code 143 is the shell convention for SIGTERM, so user code rarely produces it deliberately. The benefit (catching wrapper-translated planned-stop exits and Signal(15)/Signal(2) cases) outweighs the risk.

## Validation

| Check | Result |
|---|---|
| `cargo fmt --check` | ✓ |
| `cargo clippy --all -- -D warnings` | ✓ |
| `cargo test -p dora-daemon --lib` | ✓ 32/32 |
| `cargo test --test ws-cli-e2e -- --test-threads=1` | ✓ 21/21 in 89s |
| `cargo test --test fault-tolerance-e2e -- --test-threads=1` | ✓ **11/11** (10 existing + 1 new) in 100s |

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2589,6 +2589,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9844ddc3a6e533d62bba727eb6c28b5d360921d5175e9ff0f1e621a5c590a4d5"
 
 [[package]]
+name = "fire-and-forget-source-node"
+version = "0.2.1"
+dependencies = [
+ "dora-node-api",
+ "eyre",
+]
+
+[[package]]
 name = "fixedbitset"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,7 @@ members = [
     "tests/fault_tolerance/clean_exit_node",
     "tests/fault_tolerance/delayed_crash_node",
     "tests/fault_tolerance/event_log_observer_node",
+    "tests/fault_tolerance/fire_and_forget_source_node",
     "tests/fault_tolerance/hang_after_init_node",
     "tests/fault_tolerance/input_closed_observer_node",
     "tests/fault_tolerance/silent_source_node",
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -3777,6 +3777,18 @@ impl Daemon {
                     }
                 };
 
+                // Clear the per-incarnation kill marker so it doesn't
+                // leak into the next incarnation. `grace_duration_kills`
+                // is keyed only by `node_id`; if `restart=true` and we
+                // didn't clear here, a later external SIGTERM or
+                // unrelated 143 exit from the restarted process would
+                // still see `grace_duration_kill=true` and be
+                // misreported as `Ok(())`. The marker has done its job
+                // for this exit — drop it.
+                if let Some(dataflow) = self.running.get(&dataflow_id) {
+                    dataflow.grace_duration_kills.remove(&node_id);
+                }
+
                 logger
                     .log(
                         if node_result.is_ok() {
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -3673,44 +3673,94 @@ impl Daemon {
                         let grace_duration_kill = dataflow
                             .map(|d| d.grace_duration_kills.contains(&node_id))
                             .unwrap_or_default();
-
-                        let cause = match caused_by_node {
-                            Some(caused_by_node) => {
-                                logger
-                                    .log(
-                                        LogLevel::Info,
-                                        Some("daemon".into()),
-                                        format!("marking `{node_id}` as cascading error caused by `{caused_by_node}`")
-                                    )
-                                    .await;
-
-                                NodeErrorCause::Cascading { caused_by_node }
-                            }
-                            None if grace_duration_kill => NodeErrorCause::GraceDuration,
-                            None => {
-                                let cause = dataflow
-                                    .and_then(|d| d.node_stderr_most_recent.get(&node_id))
-                                    .map(|queue| {
-                                        let mut lines = Vec::new();
-                                        if queue.is_full() {
-                                            lines.push("[...]".into());
-                                        }
-                                        while let Some(line) = queue.pop() {
-                                            lines.push(line);
-                                        }
-                                        lines
-                                    })
-                                    .map(extract_err_from_stderr)
-                                    .unwrap_or_default();
-
-                                NodeErrorCause::Other { stderr: cause }
-                            }
-                        };
-                        Err(NodeError {
-                            timestamp: self.clock.new_timestamp(),
-                            cause,
+                        // The operator asked this node to stop (`dora stop`,
+                        // `dora destroy`, `dora node stop`, `dora run
+                        // --stop-after`) and the node responded to our
+                        // SIGTERM. On Unix the exit reports as
+                        // `Signal(15)`. Some node wrappers (notably `uv
+                        // run python`) catch SIGTERM and exit with code
+                        // 143 (= 128 + 15) instead of propagating the
+                        // signal, so `child.wait()` returns
+                        // `ExitCode(143)` not `Signal(15)`. Same shape
+                        // for SIGINT (2 / 130). Treat any of those as a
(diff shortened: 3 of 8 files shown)


heyong4725:
Both findings addressed in `90a24b38`:

**P1** — Confirmed independently by both Codex and a Claude adversarial subagent. `restarts_disabled` was the wrong discriminant: `disable_restart()` fires at subscribe time for source nodes (`lib.rs:3203`, when `open_inputs().is_empty()`), so the previous fix would silently swallow externally-sent SIGTERMs (`kill -TERM <pid>` from outside dora) on source nodes as clean. My regression test passed for the right observable outcome but for the wrong reason — `fire-and-forget-source-node` is itself a source node, so it hit the subscribe-time disable, not the operator-stop path.

Switched to `grace_duration_kill` as the discriminant — that flag is only populated when the daemon's own SoftKill/Kill submission path runs, which is the actual "daemon asked this node to stop" signal. Also extended `send_stop_and_schedule_kill` to populate `grace_duration_kills` at SoftKill and Kill submission time (mirroring what `stop_all` already does), so the `dora node stop` and `dora node restart` paths produce the signal too — otherwise the fix would only cover whole-dataflow stops.

Resulting behavior matrix:
- Operator-initiated stop + sigterm-like exit → `Ok(())` (the fix)
- Operator-initiated stop + `Signal(9)` SIGKILL → existing `NodeErrorCause::GraceDuration` branch (preserves the original "hard-killed during grace" semantic)
- External SIGTERM, no daemon kill → existing failure path (correctly reports as failure now)
- Source node natural completion → unaffected
- Cascading failure → still wins

**P2** — `cargo fmt --all` applied to the two affected files. My earlier local `cargo fmt --check | tail -2` truncated the fmt diff while showing only clippy's `Finished` line, so I thought it was clean. Verification habit fixed: `cargo fmt --all -- --check` is now run standalone, not piped.

Thanks for the catch.

heyong4725:
Addressed in `734dbe32`. Clearing `grace_duration_kills.remove(&node_id)` in the `SpawnedNodeResult` handler right after `node_result` is computed — the natural lifecycle boundary, since the marker has served its purpose for that exit classification and the next incarnation starts fresh.

Verified all four suites (fault-tolerance-e2e, ws-cli-e2e, node-lifecycle-e2e, daemon lib) still pass. `node-lifecycle-e2e` in particular now exercises the marker-clearing code path because PR #1903's lifecycle tests issue `dora node restart` + `dora node stop` which both go through `send_stop_and_schedule_kill` → marker insert → exit → marker clear.

A dedicated regression test for the restart-cycle leak (fixture that crashes on its Nth incarnation, asserts the second exit is reported as `Err`) would be a useful follow-up but needs new test infrastructure — left out of this PR's scope.

Thanks for the second pass.

phil-opp:
Note that we also merged https://github.com/dora-rs/dora/pull/1887, which might solve the issue more cleanly. I'm still not sold on the idea of treating force-killing of non-responsive nodes as a "clean exit".

PR #1919, opened 2026-05-24
fix(test): build rust-dynamic-add-remove-filter before lifecycle e2e (unblocks ci/circleci: e2e)

heyong4725:
## Summary

- **Unblocks `ci/circleci: e2e`** which has been red on `main` for ~30 consecutive commits since PR #1903 merged on 2026-05-21. Every PR opened since then (mine included) shows a red `e2e` check unrelated to its actual diff.
- Root cause: `lifecycle_rust_dynamic_add_remove` in `tests/node-lifecycle-e2e.rs` assumes `target/debug/rust-dynamic-add-remove-filter` exists, but nothing in the test setup actually builds it.
- The C++ variant has `ensure_cxx_built()` for exactly this reason. This PR adds the analogous `ensure_rust_filter_built()` helper, mirroring the existing pattern.

## Why `dora build` didn't catch it

The dynamic filter lives in a separate `filter-node.yml`, not in the main `dataflow.yml`:

```
examples/rust-dynamic-add-remove/
├── dataflow.yml         # sender + receiver (built by `dora build`)
├── filter-node.yml      # filter (added at runtime via `dora node add`)
└── filter/Cargo.toml    # the workspace crate
```

`dora build dataflow.yml` builds sender + receiver via their `build:` directives. The filter is never built by `dora build` because it isn't in `dataflow.yml`.

When the test later runs:

```
dora node add --dataflow rustlc --from-yaml filter-node.yml
```

`dora node add` reads the YAML and dispatches `ControlRequest::AddNode` to the coordinator (`binaries/cli/src/command/node/add.rs:33`). **It does not execute the `build:` field.** The daemon then tries to spawn `target/debug/rust-dynamic-add-remove-filter`, which doesn't exist on a cold CI runner.

## Why the PR that introduced this test (#1903) passed

The author almost certainly had a previously-compiled `rust-dynamic-add-remove-filter` binary sitting in their local `target/` tree from earlier development. Locally everything worked. CI's clean filesystem hits the cold-start path and exposes the gap.

## Reproduction

```bash
pkill -9 -f rust-dynamic-add-remove  # ensure cold start
rm -f target/debug/rust-dynamic-add-remove-filter

PYO3_NO_PYTHON=1 cargo test -p dora-examples \
  --test node-lifecycle-e2e -- --test-threads=1 \
  lifecycle_rust lifecycle_cxx
```

### Before this PR

```
[...]

PR #1918, opened 2026-05-24
ci: fix test-c-cpp-libraries.yml glob escape for c++ path

heyong4725:
## Summary

- **Closes #1917.** The `test-c-cpp-libraries.yml` workflow has been failing 100% of runs (100/100) since #1875 merged it on 2026-05-20. Zero jobs created per run; GitHub's UI is the only place the actual error surfaces ("This workflow file has issues") and even there it's silent on the why.
- Root cause: `apis/c++/**` in `paths:` filters. GH Actions glob treats `+` as a "one or more" quantifier; `c++` is malformed because the second `+` has no valid preceding character. The whole workflow gets rejected before matrix expansion.
- Fix: escape both `+` characters and switch the YAML quoting from double to single (double-quoted YAML rejects `\+` as an unknown escape sequence): `'apis/c\+\+/**'`.
- Diff is two lines (the `push:` and `pull_request:` `paths:` entries).

## Verification

`actionlint 1.7.7` reproduces the failure cleanly and confirms the fix:

```
# before
$ actionlint .github/workflows/test-c-cpp-libraries.yml
test-c-cpp-libraries.yml:15:17: invalid glob pattern. unexpected character '+'
  while checking special character + (one or more).  [glob]
test-c-cpp-libraries.yml:32:17: invalid glob pattern. unexpected character '+'
  while checking special character + (one or more).  [glob]

# after
$ actionlint .github/workflows/test-c-cpp-libraries.yml
# exit 0, no output
```

Swept `.github/workflows/*.yml` — no other file has `apis/c++` in a path glob. Sibling `publish-c-cpp-libraries.yml` uses `apis/c{,++}/**` in a comment only, not in an actual filter.

## Test plan

- [ ] Merge to main and observe the first run on a commit that touches `apis/c/` or `apis/c++/`. Expect 3 matrix jobs to start (ubuntu-22.04, macos-latest, windows-latest).
- [ ] Confirm subsequent `push` events to `main` no longer immediately fail with `jobs_count: 0`.
- [ ] Whether the actual smoke tests pass on first green run is a separate question — the immediate goal is to make the workflow *runnable* so we can find out.

## Follow-up (not in this PR)

Issue #1917 stretch goal: add `actionlint` as a CI step so future malformed workflows fail at PR time rather than burning runner budget for days. Worth a separate small PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1921, opened 2026-05-25
fix(cli): drop `editable = true` from python template's dora-rs source (unblocks cli-python)

heyong4725:
## Summary

- **Unblocks `ci/circleci: cli-python`** which has been red on both #1918 and #1919 at exactly 60 min runtime. Reproduces 100% deterministically locally — not a flake.
- Root cause: `dora new --lang python --internal-create-with-path-dependencies` generates a workspace pyproject declaring `dora-rs = { path = "../apis/python/node", editable = true }`. uv's editable install only writes a `.pth` file and **skips maturin's PEP 660 build hook**, so the compiled extension `dora/dora.abi3.so` never lands in the venv. Any `from dora import Node` crashes with `ModuleNotFoundError: No module named 'dora.dora'`.
- Fix: drop `editable = true`. With a plain path source, uv calls maturin to produce a real wheel and installs it normally — the `.so` ships into each per-node venv.

## Reproduction

```bash
cd <dora repo>
target/debug/dora new test_python_project --lang python --internal-create-with-path-dependencies
cd test_python_project
dora build dataflow.yml --uv
uv run pytest                 # before: ModuleNotFoundError: No module named 'dora.dora'
                              # after:  6 passed in 3.58s
dora run dataflow.yml --uv --stop-after 5s
                              # before: traceback in talker_1 + talker_2
                              # after:  listener_1: I heard Hello World from speech-2
```

## Why CI sometimes appeared green

`cli-python` on main commit `62f4bb80` reported SUCCESS at 04:01:14, but both PRs branched off it hit the failure 30 min and 90 min later. The most plausible explanation: `attach_workspace` carried a stale `target/dora-python-wheels/.../dora_rs-*.whl` from a prior pipeline that happened to still satisfy uv's resolver before the editable install kicked in. On a cold workspace (both my PRs), the lucky cache state isn't there and the bug bites every time.

## Diff scope

- `binaries/cli/src/template/python/mod.rs:139` — single line change (`, editable = true ` removed) + explanatory comment
- `Cargo.lock` — same 3-line dora-examples dev-deps regen as #1918 / #1919 so this PR's `msrv` passes independently. (Same minimal change in three PRs; whichever lands first, the others rebase cleanly.)

## Why a `[tool.uv.sources]` declaration without `editable` doesn't ignore the path

uv resolves `[tool.uv.sources]` before PyPI either way; the `editable` knob only switches between "build via the build backend then install the wheel" (off) and "drop a .pth file and skip the backend" (on). For a maturin-based package the second mode is structurally broken — there is no PEP 660 fast path for native extensions in uv's current implementation. Maturin's own docs recommend `maturin develop --uv` for editable workflows; `pip install -e` / `uv pip install -e` with PEP 660 isn't an equivalent substitute for compiled-extension cases.

## Why not switch to `maturin develop` here?

Out of scope. The template runs in CI; we just need a deterministic install path. The simplest correct option is "no editable" — uv compiles + installs the wheel, just like a normal user would on first `dora new`. If someone wants live-reload Rust development against the template (the only thing `editable = true` would actually buy you, *if* it worked), that's a separate workflow and merits a maturin-develop-aware setup.

## Test plan

- [ ] `ci/circleci: cli-python` passes on this PR (currently red on #1918 + #1919).
- [ ] After merge, the next PR opened against main shows a green `cli-python`.
- [ ] No regression in the existing `dora new --lang python` (no `--internal-create-with-path-dependencies`) path — that branch sets `dora_rs_source = ""`, untouched here.

## Related

- Introduced by: #1906 (`928e096c`) — added `editable = true` without explicit rationale
- Preserved by: #1913 (`f2ae047d`) — moved to placeholder substitution but kept the flag
- Concurrent unrelated CI fixes:
  - #1918 — workflow glob escape (`apis/c++/**` → `'apis/c\+\+/**'`)
  - #1919 — `lifecycle_rust_dynamic_add_remove` missing `ensure_rust_filter_built()`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -718,12 +718,13 @@ jobs:
           name: Python Queue Latency + Timeout Test
           command: dora run tests/queue_size_and_timeout_python/dataflow.yaml --uv
           no_output_timeout: 30m
-      - run:
-          name: Rust Queue Latency Test
-          command: |
-            dora build tests/queue_size_latest_data_rust/dataflow.yaml --uv
-            dora run tests/queue_size_latest_data_rust/dataflow.yaml --uv
-          no_output_timeout: 30m
+      # Rust Queue Latency Test relocated to `contract-tests` to keep
+      # cli-python under CircleCI's 60m job wall (the editable-fix in
+      # this PR makes the Python steps actually do real work, pushing
+      # totals to ~58m; the rust test added ~2m on top and was getting
+      # truncated). The rust test still needs uv (the dataflow has a
+      # Python sender feeding a Rust receiver), so contract-tests —
+      # which already runs setup-python-uv — is the natural home.
       - cargo-cache-save
 
   cli-templates:
@@ -960,6 +961,18 @@ jobs:
             --test-threads=1
             lifecycle_python
           no_output_timeout: 15m
+      # Rust Queue Latency Test moved from cli-python here to keep
+      # cli-python under CircleCI's 60m job wall. The dataflow has a
+      # Python sender (queue_size_latest_data_python/send_data.py)
+      # feeding a Rust receiver, so it needs the uv setup that this
+      # job already has via setup-python-uv. Adds ~2m to a job that
+      # otherwise runs ~22m.
+      - run:
+          name: Rust Queue Latency Test
+          command: |
+            dora build tests/queue_size_latest_data_rust/dataflow.yaml --uv
+            dora run tests/queue_size_latest_data_rust/dataflow.yaml --uv
+          no_output_timeout: 30m
       - cargo-cache-save
 
   # ===== Benchmark regression check =====
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1878,6 +1878,7 @@ dependencies = [
 name = "dora-examples"
 version = "0.0.0"
 dependencies = [
+ "arrow-array",
  "assert2",
  "dora-cli",
  "dora-coordinator",
@@ -1885,10 +1886,12 @@ dependencies = [
  "dora-daemon",
  "dora-download",
  "dora-message",
+ "dora-operator-api-types",
  "dora-tracing",
  "dunce",
  "eyre",
  "futures",
+ "libloading 0.8.9",
  "serde_json",
  "serde_yaml",
  "tempfile",
--- a/binaries/cli/src/template/python/mod.rs
+++ b/binaries/cli/src/template/python/mod.rs
@@ -135,8 +135,20 @@ fn create_dataflow(
     // a second `[tool.uv.sources]` header at the end of the file
     // (duplicate section = TOML parse error), so this stays an in-place
     // substitution rather than a `push_str`.
+    //
+    // NOTE: do NOT set `editable = true` here. uv's editable install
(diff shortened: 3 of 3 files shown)


heyong4725:
### Why cli-python failed at 60m on the first run

The CircleCI v1.1 step log on the first #1921 run (job 11039) shows ALL Python steps PASSED with this PR's `editable = true` fix:

| Step | green main (71e98088) | this PR first run | delta |
|---|---|---|---|
| Test Python template project | 5.0m | **9.7m** | +4.7m |
| Test Python Operator example | 9.6m | 10.7m | +1.1m |
| all other Python steps | ✅ | ✅ | — |
| Rust Queue Latency Test | 2.3m ✅ | (1.0m, **timed out**) | — |
| **Total** | **54.2m** | **60.0m wall — Build timed out** | — |

Per-step `no_output_timeout` config never fired. The failure was CircleCI's job-level **60-minute wall** — visible as a `"Build timed out"` marker step with 0.0m runtime right after `Rust Queue Latency Test` got cut off at 1.0m. Before the editable fix, the Python steps short-circuited fast (broken imports), keeping total runtime under 54m; with the fix they actually do real work, totaling ~58m, leaving no room for the rust test.

### What's new in `49e28025`

Moved `Rust Queue Latency Test` from `cli-python` (60m wall) → `contract-tests` (currently ~22m on green main, plenty of room).

Why contract-tests is the right home:
- The dataflow has a Python sender feeding a Rust receiver, so it needs `setup-python-uv` — already done in contract-tests, not in cli-rust.
- Sits alongside `Run dora node lifecycle E2E (Python)` which is also a Python-adjacent test that uses `setup-python-uv`.

Expected: cli-python at ~58m green (just under the wall); contract-tests at ~24m green.

heyong4725:
Closing — this PR's premise was wrong.

### What I missed

Main commit `014820ba` (the #1919 squash merge) ran CI overnight and **`ci/circleci: cli-python` passed in 56 min**, with the `editable = true` line still in place. So `editable = true` is NOT broken on Linux CI — the 60-min failures on #1918, #1919, and this PR's first run were transient CI infrastructure issues clustered around 04:00 UTC, not a real bug.

My local macOS repro DID show a real `ModuleNotFoundError: No module named 'dora.dora'` failure mode, but that's a macOS-specific maturin editable behavior; Linux CI's uv editable install of `apis/python/node` correctly invokes maturin's PEP 660 hook and produces the `.so`.

### Why this fix would be net-negative

The first run of this PR (before the Rust Queue Latency Test move) shows:

| Step | green main 71e98088 | this PR | delta |
|---|---|---|---|
| Test Python template | 5.0m | 9.7m | +4.7m |
| Test Python Operator | 9.6m | 10.7m | +1.1m |

Removing `editable = true` makes each per-node `uv pip install` actually fetch + install the maturin-built wheel instead of just creating a `.pth` file. That's strictly more work on Linux. Even with the Rust Queue Latency Test moved out, cli-python total would be ~58m vs main's 56m — eating into the budget for nothing on Linux.

### Closing

Going with rebase-#1918-onto-main as the real path forward. Main is healthy now; the open PRs just need to be on it.

Sorry for the noise.

PR #1953, opened 2026-05-27
fix(cli): defer distributed build session writes

### LINKED ISSUE #1472: fix(cli/coordinator): distributed build persists session git_sources before build success
## Problem

  In distributed build flow, the CLI updates and writes `dataflow_session.git_sources` **before** `wait_until_dataflow_built(...)` confirms build success.

  This means a failed remote build can still persist new source metadata in the local session file, causing session/build state drift.

  ## Why this matters

  This is a reproducibility and lifecycle-consistency bug at the build ↔ run boundary:

  - session metadata can claim sources from a build that never succeeded
  - subsequent runs may rely on incorrect source state
  - debugging failures becomes harder because persisted state is misleading

  This directly impacts Project #2 goals around reproducible dependency/source state.

  ## Expected behavior

  Session mutation (`git_sources`, `build_id`, etc.) should only be written after distributed build success is confirmed.

  ## Actual behavior

  Session state is written before build success is known, so failure paths can still persist new metadata.

  ## Scope proposal

  - Move distributed session write to success path only.
  - Ensure failure path leaves previous on-disk session unchanged.
  - Add a regression test for failed distributed build flow.

  ## References

  - `binaries/cli/src/command/build/mod.rs` (distributed build flow)

  ## Tests

  Added regression test:

  - `distributed_build_failure_does_not_persist_new_git_sources`

  What it verifies:

  - If distributed build result is `Err`, session finalization returns an error.
  - `session.git_sources` is not mutated in-memory.
  - On-disk session file remains unchanged (no accidental persistence of new git sources).

  Validation run locally:

  - `cargo fmt --all`
  - `cargo check -p dora-cli`
  - `cargo test -p dora-cli distributed_build_failure_does_not_persist_new_git_sources -- --nocapture`

iccccccccccccc:
Fixes #1472

## Changes
- Move distributed build session mutation and `write_out_for_dataflow` behind the `wait_until_dataflow_built` success result.
- Keep the failure path all-or-nothing: if the coordinator build is submitted but waiting for completion fails, the existing in-memory and on-disk session metadata is left alone.
- Add focused tests for both the failed wait path and the successful finalization path.

## Verification
- `cargo fmt --all --check`
- `cargo test -p dora-cli distributed_build_failure_does_not_persist_new_git_sources -- --nocapture`
- `cargo test -p dora-cli distributed_build_success_persists_git_sources_and_build_id -- --nocapture`
- `cargo test -p dora-cli command::build::tests -- --nocapture`
- `cargo check -p dora-cli`
- `git diff --check`

### DIFF
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -52,7 +52,7 @@ use dora_core::{
     topics::{DORA_COORDINATOR_PORT_WS_DEFAULT, LOCALHOST},
     types::TypeRegistry,
 };
-use dora_message::{BuildId, descriptor::NodeSource};
+use dora_message::{BuildId, common::GitSource, descriptor::NodeSource, id::NodeId};
 use eyre::Context;
 use std::{
     collections::BTreeMap,
@@ -437,24 +437,17 @@ pub fn build(cfg: BuildConfig) -> eyre::Result<()> {
                 uv,
             )?;
 
-            dataflow_session.git_sources = git_sources;
-            dataflow_session
-                .write_out_for_dataflow(&dataflow_path)
-                .context("failed to write out dataflow session file")?;
-
             // wait until dataflow build is finished
+            let build_result =
+                wait_until_dataflow_built(build_id, &coordinator_session, log::LevelFilter::Info);
 
-            wait_until_dataflow_built(build_id, &coordinator_session, log::LevelFilter::Info)?;
-
-            dataflow_session.build_id = Some(build_id);
-            dataflow_session.local_build = None;
-            // Same fingerprint-recording rationale as the local-build branch
-            // above (#1444). The session file is the source of truth for
-            // "what descriptor produced this `build_id`."
-            dataflow_session.build_fingerprint = Some(session_build_fingerprint.clone());
-            dataflow_session
-                .write_out_for_dataflow(&dataflow_path)
-                .context("failed to write out dataflow session file")?;
+            finalize_distributed_build_session(
+                &mut dataflow_session,
+                &dataflow_path,
+                git_sources,
+                build_result,
+                session_build_fingerprint,
+            )?;
         }
     };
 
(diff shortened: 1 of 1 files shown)


heyong4725:
## Pre-landing review (assisted by Claude Code, `/pr-review`)

**Verdict: clean fix, approve.** Correct, minimal, well-tested, consistent with the parallel local-build branch.

### Scope
One file, `binaries/cli/src/command/build/mod.rs` (+133/-17). Matches intent of #1472: make distributed `dora build` all-or-nothing. The branch is a few commits behind `main` (predates #1954), so a local `git diff` against `main` shows unrelated README/`architecture.svg` noise — a rebase clears that.

### Why it's correct
Old code wrote `git_sources` to disk *before* waiting for the distributed build, then wrote again after success — a failed wait left the session with new `git_sources` but a stale `build_id`. The fix captures the wait as a `Result` and routes it through `finalize_distributed_build_session`, which does `let build_id = build_result?;` **before any mutation**. Failure path mutates nothing, writes nothing. One write after success instead of two with a bad intermediate state.

### Tests (verified locally, 9/9 pass)
- `distributed_build_failure_does_not_persist_new_git_sources` — asserts both in-memory and on-disk session unchanged on failure ★★★
- `distributed_build_success_persists_git_sources_and_build_id` — asserts fields set + persisted on success ★★★

### Cross-model adversarial pass (Codex) — calibrated
Codex flagged surrounding hazards; none is a regression introduced by this PR:
- Non-atomic `write_out_for_dataflow` — pre-existing, unchanged; this PR *reduces* write count.
- Failed distributed build leaves a prior `--local` `build_id` intact — this is the intended "leave existing metadata alone" behavior of #1472. Worth a **separate follow-up issue** only if a failed coordinator build should invalidate a prior local build.
- `--write-lockfile` ordering and `wait` error-model collapsing — pre-existing, out of this diff.
- Success is logged inside `wait` before the session is persisted — minor/pre-existing; the command still exits non-zero with `failed to write out dataflow session file` if the write fails, so not silent.

### Non-blocking suggestions
1. Rebase on `main` to drop the unrelated README/SVG delta from the diff.
2. (optional, follow-up issue) Decide whether a failed distributed build should invalidate a prior local build's `build_id`.

PR #1956, opened 2026-05-27
fix(cli): reject a distributed start of a locally-built dataflow

### LINKED ISSUE #1955: dora build: failed distributed build leaves a prior local build_id reusable
### Summary

A **failed distributed build** currently leaves a prior **local** build's metadata (`build_id` + `local_build`) intact in the dataflow session. If the build-inputs fingerprint still matches, a later `dora start` can reuse that local-only `build_id` against the coordinator instead of forcing a fresh distributed rebuild.

This is a latent design question, not a regression. It was surfaced during the cross-model review of #1953 and is explicitly *out of scope* for that PR — #1953 (closing #1472) deliberately makes the distributed-build failure path all-or-nothing: on failure it leaves the existing on-disk session **untouched**. That is the correct fix for #1472. This issue is about whether "leave the prior state untouched" is the right behavior when the prior state is a *local* build and the user just attempted a *distributed* one.

### Repro shape

1. `dora build --local` on a dataflow with `deploy` sections → writes a local `build_id` + `local_build` and a build fingerprint to the session.
2. `dora build` through a coordinator, which **fails** (build error, or a transport/observe failure in `wait_until_dataflow_built`).
3. Because the failure path now leaves all prior metadata intact and the fingerprint can still match, a subsequent `dora start` may send the local-only `build_id` to the coordinator rather than rebuilding.

### Relevant code

- `binaries/cli/src/command/build/mod.rs` — `finalize_distributed_build_session()` returns early via `build_result?` before any mutation (intended).
- Local-build branch in `build()` sets `build_id` / `local_build` / `build_fingerprint`.

### Questions to resolve

- Should a *failed* distributed build invalidate a prior *local* build's `build_id` / `local_build` (e.g., clear `local_build` or mark the session as needing a distributed rebuild)?
- Or is fingerprint-based reuse already sufficient to force a rebuild here, making this a non-issue? (Needs verification of the fingerprint/`build_id` reuse path in `dora start`.)
- Related but separate: `wait_until_dataflow_built` collapses real build failures and transport/observe failures into the same `Err`. Should a post-success observation failure be distinguished so a genuinely-successful remote build isn't discarded?

### Notes

- Surfaced by Codex adversarial pass during `/pr-review` of #1953.
- No data corruption; worst case is a stale/local `build_id` being used or an unnecessary rebuild.

Follow-up to #1953 / #1472.

heyong4725:
## What

Fixes #1955. After a local `dora build` — or a failed distributed build that left a prior local build cached (surfaced reviewing #1953) — `dora start` would send a **local-only** `build_id` to the coordinator. Remote daemons resolve build metadata by id (`self.builds.get(id)` in `daemon/src/lib.rs`), and a local build's id is unknown there, so it behaves **exactly like no build**:

- nodes with a git source fail with the misleading *"node X has git source, but no `dora build` was run yet"* (the user *did* build, just locally);
- other nodes silently spawn with default working dirs / env, ignoring the build's metadata.

`dora start` never runs build commands (by design), so it cannot recover by sending `None` either — `None` and an unknown id produce the same `build_info = None` on the daemon.

## Fix

Detect the mismatch in the CLI and **fail fast, before connecting to the coordinator**, with an actionable message: when the dataflow has `deploy` sections **and** a local build is cached (`local_build.is_some()`), refuse and tell the user to run a distributed `dora build`.

- Replaces a confusing daemon-side error (git nodes) / silent wrong-metadata start (non-git nodes) with one clear CLI error.
- Local builds for **non-distributed** dataflows are unaffected — the common single-machine `dora build` + `dora start` flow still works.
- Does not touch #1953's failure path and adds no session writes.

Decision lives in a pure predicate `local_build_blocks_distributed_start(has_local_build, has_deploy_nodes)` with a truth-table unit test.

## Why fail instead of "drop the id"

An earlier revision dropped the local id to `None`, but the daemon treats `None` and an unknown id identically (`build_id.and_then(|id| self.builds.get(&id))` → `None`), so that changed nothing — the git error and the silent non-git start both remained. Since `dora start` can't build, failing early is the correct fix (thanks to review feedback for catching this).

## Test plan

- [x] `cargo test -p dora-cli` — 176 passed, 0 failed (incl. `command::start::tests`)

Closes #1955.

### DIFF
--- a/binaries/cli/src/command/start/mod.rs
+++ b/binaries/cli/src/command/start/mod.rs
@@ -143,6 +143,24 @@ fn start_dataflow(
     }
     drop(resolved_for_fingerprint);
 
+    // A local `dora build` produces artifacts that only exist on this machine.
+    // A distributed (`deploy`) dataflow is started through the coordinator and
+    // runs on remote daemons that cannot see that build, so the cached id is
+    // useless there and `dora start` cannot rebuild. Refuse early with an
+    // actionable message instead of letting the daemon fail confusingly (#1955).
+    let has_deploy_nodes = dataflow_descriptor.nodes.iter().any(|n| n.deploy.is_some());
+    if local_build_blocks_distributed_start(
+        dataflow_session.local_build.is_some(),
+        has_deploy_nodes,
+    ) {
+        eyre::bail!(
+            "this dataflow was built locally, but it has `deploy` sections and is started \
+             through the coordinator — remote daemons cannot use a local build.\n\n  \
+             run `dora build` against the running coordinator (without `--local`) to build on \
+             the target machines before `dora start`"
+        );
+    }
+
     if debug {
         dataflow_descriptor.debug.enable_debug_inspection = true;
     }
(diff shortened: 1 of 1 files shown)


heyong4725:
### Design note: why the guard keys on `deploy.is_some()`, not `deploy.machine`

A review pass suggested narrowing the guard from "any node has a `deploy` block" to "any node targets a remote `machine`". That would be incorrect — it would reintroduce #1955 for `deploy` blocks without a `machine`.

The reason is how `local_working_dir()` (`binaries/cli/src/common.rs:261`) decides whether the daemon can see the CLI's workspace:

```rust
dataflow_descriptor.nodes.iter()
    .all(|n| n.deploy.as_ref().map(|d| d.machine.as_ref()).is_none())
    && cli_and_daemon_on_same_machine(...)
```

`n.deploy.as_ref().map(|d| d.machine.as_ref())` is `Option<Option<&String>>`, so `.is_none()` is true **only when `n.deploy` is `None`**. A `deploy` block with `machine: None` yields `Some(None)`, which is *not* `is_none()`. So the CLI workspace is shared **only when no node has a `deploy` block at all**.

Therefore:
- **any** `deploy` block ⟺ `local_working_dir() == None` ⟺ the daemon uses its own working dirs ⟺ a local `dora build`'s artifacts are not reachable.
- That is exactly when reusing a local build is unsafe, so `has_deploy_nodes = any(n.deploy.is_some())` is the correct granularity.

Narrowing to `deploy.machine.is_some()` would let a `deploy: { machine: None, working_dir: ... }` dataflow through, but that dataflow also gets `local_working_dir() == None` — so its local build is unusable and we'd be back to the #1955 failure.

Leaving the condition as-is on purpose.

Issue #1972, opened 2026-05-30
C++ action client: latent response-stealing bug with concurrent goals (same class as #1970)


heyong4725:
## Summary

The generated C++ action client has the **same class of bug** as the service client fixed in #1970/#1971, but **latent** — it is not triggered by the current example, so it does not fail today.

## Where

`libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs`, in the generated action-client `request_result`: it spawns one `ActionClient::async_request_result(goal_id)` **per goal** on the node's shared executor.

`async_request_result` → `async_call_service` → `async_receive_response` on the client's single GetResult response reader, which **discards** any sample whose request id doesn't match (ros2-client `service/client.rs`: *"Received response for someone else"* → `continue`). Exactly the mechanism behind #1970.

## Why it doesn't fail today

With **multiple goals outstanding concurrently**, the per-goal result receivers would steal and drop each other's GetResult responses (same as #1970). But the only action-client example (`examples/ros2-bridge/c++/action-client/main.cc`) drives **one goal at a time** — it guards `send_goal`/`request_result` with `if (!goal_id)` and only clears `goal_id` after the goal succeeds. With at most one result request in flight, there is no concurrent stealing, so `cxx-ros2-dataflow-action-client` passes (verified on arm64 Humble: three sequential goals, all results received).

## Impact

A C++ action client that keeps **multiple goals in flight simultaneously** and requests results for each would likely receive results for only some/none of them on slower transports (the #1970 failure mode). Single-goal-at-a-time usage is unaffected.

## Fix sketch (harder than #1970)

The service fix (#1971) used a single per-client pump via `Client::receive_response()` (returns any response). The action client is trickier:

- `receive_result(req_id)` also discards non-matching samples **and** doesn't return the id, so it can't demux.
- The underlying result service client is reachable via `ActionClient::result_client(&mut self)` — but that needs `&mut`, while the generated client holds the `ActionClient` behind `Arc`. A proper fix needs interior mutability plus a single pump that calls `result_client().receive_response()` and demuxes `req_id → goal_id` to build the per-goal result events.

Deferred because there is **no multi-goal test** to verify against; a fix should land together with a concurrent-goal regression test.

Found while fixing #1970.

AxonOS-BCI:
@heyong4725  I should clarify my previous comment: I referenced the wrong runtime component.

[...]

PR #1986, opened 2026-06-01
feat(ros2-bridge): C++ ROS2 action server codegen + example (#1170)

heyong4725:
## Summary

Phase 5b — the **final feature** of the dora-ros2 bridge roadmap (#1170): a **C++ ROS2 action server** (cxx codegen + Fibonacci example). With this, every #1170 sub-item is covered across Rust, Python, and C++.

## What this adds

- **`cxx_action_server_creation_functions`** in `msg-gen/src/types/action.rs` — mirrors the daemon's `run_action_server` FSM:
  - a dedicated-thread goal pump receives goals, auto-accepts + starts executing them, stores each `ExecutingGoalHandle` keyed by goal-id uuid, and forwards `(goal_id, goal)` to the merged event stream;
  - it **rejects** goals once `MAX_CONCURRENT_GOALS` (16) are in flight, so the handle map can't grow unbounded;
  - C++ side: `server->matches(event)` → `server->downcast(event)` → a `GoalEvent` (`get_goal_id()` + `get_goal()`) → `server->publish_feedback(goal_id, feedback)` / `server->send_result(goal_id, status, result)`, which look the handle back up.
  - The shared `ActionGoalId` cxx type carries the goal id (no new ExternType). `ExecutingGoalHandle` is parameterized by the **goal-message** type (not the action marker) so it is `Clone`.
- **Example** `examples/ros2-bridge/c++/action-server/` (`Fibonacci` server) paired with the existing C++ action client in one **self-contained** dataflow. A real rmw client cannot discover a ros2-client-hosted server (ros2-client#4), so the peer is another dora node; `run.rs` builds both C++ binaries.
- **x86 nightly CI**: added the example to the C++ ROS2 Bridge job (self-contained, no extra package).

## Verification

| Step | arm64 harness | Notes |
|---|---|---|
| Codegen + C++ API compile (all action types) | ✅ | `dora-node-api-cxx` cxx-bridge + clang build of both `main.cc` |
| clippy `-D warnings` (msg-gen host + bridge in-container) | ✅ | |
| cxx client discovers cxx server, **goal received** | ✅ | `received goal (order 10)` |
| `get_result` round-trip delivered | ⛔ | does not complete on arm64/Docker — documented #1972-family limitation, independent of this code |

The macOS arm64 harness can't complete the deferred `get_result` service round-trip (same as the Rust action server, Phase 4), so the **x86 nightly C++ job is the oracle** (added here). Intentionally **not** added to `scripts/ros2dev.sh` (the stall would hang it on arm64).

Root-cause note: an initial pairing with the *Rust* action client failed to discover on arm64 (a Rust↔C++ DDS quirk); an isolation test (cxx↔cxx) confirmed the C++ server is discovered and receives goals, so the example pairs C++↔C++.

## Test plan

- [x] `cargo clippy -p dora-ros2-bridge-msg-gen -- -D warnings`
- [x] `cargo clippy -p dora-ros2-bridge --features ros2-examples -- -D warnings` (in-container)
- [x] `cargo build -p dora-node-api-cxx --features ros2-bridge` (in-container) — generated FFI compiles across all action types
- [x] `cargo run -p dora-ros2-bridge --example cxx-ros2-dataflow-action-server --features ros2-examples` → both C++ nodes build, server discovered, `received goal (order 10)` (arm64; get_result completes on x86 nightly)

Closes the last feature of #1170. (Remaining tracker item: examples aligned with ros2/examples — incremental.)


heyong4725 (commented):
Findings:

- P2 `libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs:1001`: `send_result` removes the executing goal from the local map before `send_result_response` has actually completed. If the 300s wait times out, or serialization/send fails, the method returns an error but the goal is already forgotten, so the caller cannot retry `send_result`, publish final feedback, cancel/abort it, or otherwise complete that accepted goal through this API. This is especially relevant because the PR notes a known `get_result` stall in some environments. The Python action-server API keeps the handle tracked until `send_result_response` succeeds (`libraries/extensions/ros2-bridge/python/src/lib.rs:1309`), which is the safer behavior for an API that reports a recoverable error to the caller. Please clone/get the handle first and only remove it after the send succeeds.

- P3 `scripts/qa/ci-nightly-jobs.sh:1126`: this local driver still says it mirrors the `nightly.yml` `ros2-bridge` job, but the new `C++ ROS2 Bridge action-server example` step added in `.github/workflows/nightly.yml:1542` is not represented here. The existing skip note at `scripts/qa/ci-nightly-jobs.sh:1167` only documents the Rust action examples as intentionally x86-nightly-only. Please either add the C++ action-server example under an appropriate x86/Linux guard or extend the explicit skip note/SKIPPED entry so the local mirror does not silently miss this newly added nightly coverage.

Validation performed:

- `git diff --check origin/main...HEAD`
- `cargo clippy -p dora-ros2-bridge-msg-gen -- -D warnings`
- `cargo check -p dora-node-api-cxx --features ros2-bridge`
- `cargo check -p dora-ros2-bridge --features ros2-examples --example cxx-ros2-dataflow-action-server`

The Rust checks passed. The environment did not have `AMENT_PREFIX_PATH` set, so these runs verified compile/check behavior but did not execute the ROS2 dataflow end-to-end.

Issue #1992, opened 2026-06-02
Multi-daemon: inter-daemon-event subscriber bincode-decodes node raw publishes (InterDaemonEvent deserialize flood, flaky multiple-daemons)


heyong4725:
## Summary

In distributed (multi-daemon) mode, a cross-daemon output is published to a Zenoh key in **node raw format**, but the consuming daemon's **inter-daemon-event subscriber** subscribes to that same key and unconditionally `bincode`-decodes every sample as a `Timestamped<InterDaemonEvent>`. The two payload formats are incompatible, so the consuming daemon floods deserialize errors:

```
ERROR dora_daemon: Daemon error: failed to deserialize InterDaemonEvent
    UUID parsing failed: invalid length: expected 16 bytes, found 63   (also 62 / 64)
```

The dataflow usually limps to completion via messages that do get through, but when the error volume/timing prevents finishing in time, the run fails. This is the cause of the intermittent **`multiple-daemons` nightly failure** (`Error: dataflow not finished after 101 retries`, ~2/12 recent nightly runs).

## Impact

This is a core distributed-mode bug, not an example artifact. Any multi-daemon / multi-machine deployment where a cross-daemon output takes the direct-Zenoh-publish path can hit it. The `multiple-daemons` example is simply the one CI job that exercises it.

## Reproduction

`cargo run --example multiple-daemons` triggers the deserialize errors reliably (36–200 per run observed locally on macOS/arm64). Whether the overall run then *fails* is timing-dependent (intermittent), but the underlying error flood is consistent.

> Note: the `v0.2.1` strings in build output are just crate versions (`fxhash v0.2.1`, etc.), not a version-mismatch confounder.

## Root cause (confirmed via instrumentation)

Two output-forwarding mechanisms collide on a single Zenoh key:

- **Node raw publish** — `DoraNode::zenoh_publish` (`apis/rust/node/src/node/mod.rs:1041`) puts the raw output bytes as the payload with `Metadata` in the Zenoh **attachment**, on key `dora/{net}/{dataflow}/output/{node}/{output}` (`zenoh_output_publish_topic`, `libraries/core/src/topics.rs:272`). This is consumed correctly by the node event-stream subscriber (`apis/rust/node/src/event_stream/mod.rs:305`) and the daemon debug-inspection subscriber, which read the raw payload + attachment.
[...]

PR #1993, opened 2026-06-02
fix(examples): make c-api consumer node tolerant of stream end (fixes flaky Windows nightly)

heyong4725:
## Problem

The nightly `Examples (windows-latest)` job intermittently fails (~2/10 runs; e.g. 2026-06-01 and 2026-06-02) on the `cxx-dataflow` example:

```
cxx-node-c-api: [c node] ERROR: unexpected end of event
ERROR node_failure: node exited with error: ExitCode(-1) node_id=cxx-node-c-api
Error: Dataflow failed: Node `cxx-node-c-api` failed: exited with code -1
error: process didn't exit successfully: `target\debug\examples\cxx-dataflow.exe` (exit code: 1)
```

It's a flake, not a hard regression — the same commit passes on the next run.

## Root cause

`examples/c++-dataflow/node-c-api/main.cc` hard-looped **exactly 20 events** and treated an early `dora_next_event() == NULL` as a fatal error (`return -1`). Its input is the upstream producer's output (`cxx-node-rust-api/counter`). The producer sends ~20 counters then exits. On Windows, Zenoh pub/sub discovery drops the first message(s) before the subscription is matched, so the consumer can't reach 20 inputs before the producer finishes and the stream closes → `NULL` → `-1`. Early-message loss before pub/sub discovery settles is inherent to Zenoh/DDS/ROS, so the real bug is the consumer assuming exact delivery.

## Fix

Process events until the stream is cleanly exhausted instead of demanding a fixed count:

```c
while (true) {
    void *event = dora_next_event(dora_context);
    if (event == NULL) { break; }  // all inputs closed = normal end
    ...
}
return 0;
```

`dora_next_event` returns `NULL` only when the event stream is cleanly exhausted (`events.recv()` → `None`, `apis/c/node/src/lib.rs:78`); genuine upstream failures arrive as a separate `Error` event, **not** `NULL` — so this does not mask real failures. The producer's bounded loop closes the consumer's input, so the unbounded loop terminates (no hang). This matches the canonical consumer idiom in `examples/c-dataflow/sink.c`.

### Scope: c++-dataflow only

This is intentionally **not** applied to `cmake-dataflow/node-c-api`. Although the C code is identical, the wiring differs: cmake-dataflow's `node-c-api` reads a **direct timer** (`dora/timer/millis/300`) that never closes, so it relies on the finite `for i < 20` bound to terminate — an unbounded loop there would hang until the CI timeout. It's also timer-driven (daemon-local), so it isn't subject to the cross-node Zenoh discovery-drop flake this fix addresses. It keeps its finite bound.

## Cross-platform impact

- **Windows:** fixes the recurring `cxx-dataflow` flake.
- **Linux/macOS:** removes the same latent fragility for `cxx-dataflow` (could break under slow/loaded discovery), gives cleaner failure reporting (no spurious `-1` masking the real cause), and aligns with the idiomatic event-loop pattern. No happy-path behavior change; no perf change.

## Validation

- Builds cleanly; ran a stripped two-node dataflow matching c++-dataflow's wiring (producer → this consumer) via the workspace CLI → exit 0, clean shutdown (`InputClosed` → `stop` → `GOODBYE`), no "unexpected end of event".
- `cargo check -p dora-daemon` green.

> Note: the underlying Zenoh startup message-loss is not addressed here (it's inherent to pub/sub; the correct layer to fix is the consumer being tolerant). A separate, unrelated Linux `multiple-daemons` flake was root-caused and filed as #1992.

### DIFF
--- a/examples/c++-dataflow/node-c-api/main.cc
+++ b/examples/c++-dataflow/node-c-api/main.cc
@@ -10,13 +10,15 @@ int run(void *dora_context)
 {
     unsigned char counter = 0;
 
-    for (int i = 0; i < 20; i++)
+    while (true)
     {
         void *event = dora_next_event(dora_context);
         if (event == NULL)
         {
-            printf("[c node] ERROR: unexpected end of event\n");
-            return -1;
+            // Stream exhausted (all inputs closed) -- the normal end of the
+            // dataflow, not an error. Don't assume a fixed input count: pub/sub
+            // discovery can drop early messages before subscriptions match.
+            break;
         }
 
         enum DoraEventType ty = read_dora_event_type(event);


heyong4725:
Addressed the P1: reverted `cmake-dataflow/node-c-api` (force-pushed). Good catch — its `node-c-api` reads a **direct timer** (`dora/timer/millis/300`) with no finite upstream, so `while(true)` would hang until the CI timeout. The PR is now scoped to `c++-dataflow` only, whose `node-c-api` consumes a finite upstream output (`cxx-node-rust-api/counter`) that closes when the producer exits — so the unbounded loop terminates. cmake-dataflow keeps its finite bound (and isn't subject to the cross-node discovery flake anyway).

Issue #1970, opened 2026-05-29
C++ ROS2 service client receives no responses on arm64 (async_receive_response never resolves on ThreadPool)


heyong4725:
## Summary

The C++ ROS2 bridge **service client** receives **zero responses** on arm64 (Apple Silicon, Docker), even though it successfully sends requests and the ROS2 service processes them. The `cxx-ros2-dataflow` example aborts at `examples/ros2-bridge/c++/turtle/main.cc:109` (`assert(responses_received > 0)`).

The Rust service client works on the same machine with the identical service, mapping, QoS, and domain — so this is specific to the **generated C++ client's response-forwarding path**, not the bridge's service mechanism in general.

> Found while building a macOS dev harness for #1170. The Python service client added in #1969 is unaffected (it uses the direct `block_on(async_receive_response)` path, which works on arm64).

## Environment

- Apple Silicon (arm64), ROS2 Humble in Docker (`osrf/ros:humble`), default RMW (`rmw_fastrtps_cpp`), `ROS_DISTRO=humble` → `ServiceMapping::Enhanced`.
- Deterministic: 3/3 runs fail with 0 responses.
- **x86 is unaffected**: the nightly `ROS2 Bridge Examples` job (ubuntu-22.04) was green on 2026-05-28.

## Reproduction

```bash
source /opt/ros/humble/setup.bash
QT_QPA_PLATFORM=offscreen cargo run -p dora-ros2-bridge --example cxx-ros2-dataflow --features ros2-examples
# -> "Received 0 service responses"
# -> main.cc:109: Assertion `responses_received > 0' failed.  (SIGABRT)
```

## Evidence (request path works, reply path does not)

- `ros2 service call /add_two_ints ...` against the same service node works in-container (returns `sum=9`).
- During the failing run, the `minimal_service` logs `request: 4 + 5` for **every** request — i.e. the C++ client's requests arrive and are answered by the service.
- Topic pub/sub works in the same dataflow (poses received), so RustDDS discovery/readers are fine on arm64.
- The **Rust** client example (`rust-ros2-dataflow`) round-trips correctly with the same service/mapping/QoS.
- Instrumented run (`RUST_LOG`): **no** `failed to receive response` / `failed to send service response` warning and **no panic** in the spawned task — it is simply parked, never resolving.

## Root cause (localized)

The generated C++ service client (`libraries/extensions/ros2-bridge/msg-gen/src/types/service.rs`, ~L142–201) handles responses by:

[...]

PR #1995, opened 2026-06-02
fix(ros2-bridge): demux C++ action-client GetResult responses by request id (#1972)

### LINKED ISSUE #1972: C++ action client: latent response-stealing bug with concurrent goals (same class as #1970)

heyong4725:
Fixes #1972.

## Problem

The generated C++ action client spawned **one `async_request_result` receiver per goal**. ros2-client's `async_request_result` / `receive_result` discard responses whose request id doesn't match (`"Received response for someone else" → continue`), so with multiple goals in flight the per-goal receivers **stole and dropped each other's `GetResult` responses** — concurrent goals received results for only some/none of them. Same class as the service-client bug #1970 (fixed in #1971); latent because the existing single-goal example drives one goal at a time.

## Fix

Replace the per-goal spawn with a **single result-response pump that demuxes by request id**, mirroring the merged #1971 service-client fix (`libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs`):

- `request_result(&self)` registers `(req_id, goal_id)` in `pending: Arc<Mutex<Vec<(RmwRequestId, ActionGoalId)>>>`, holding the lock across the synchronous `ActionClient::request_result(&self)` send so a result arriving before registration completes isn't dropped as "not ours".
- One pump thread calls `result_client().receive_response()` and routes each response to its goal; responses with no pending entry (another client on the topic) are dropped.
- Removed the now-unused `result_tx` / `executor` client fields (also drops the per-goal `ThreadPool` spawn — a net simplification).

The `&mut`-via-ptr access of `result_client()` mirrors the existing feedback-stream pattern in the same generated client (forced by ros2-client exposing the sub-client only via `&mut self`); the concurrent `&self` `send_request` + `receive_response` on the `Client` is data-race-free (disjoint request/response paths + atomic sequence ids) — the same concurrency #1971 relies on.

## Regression test

New `cxx-ros2-dataflow-action-client-concurrent` example: fires 3 Fibonacci goals concurrently, requires a result for **every** one, and fails fast (non-zero exit, `TIMEOUT: ... see #1972`) on shortfall.

## Verification (arm64 ROS2 Humble harness)

| | result | exit |
|---|---|---|
| existing single-goal example (regression) | 3/3 sequential | 0 ✅ |
| **concurrent test, post-fix** | **3/3** correct distinct sequences | 0 ✅ |
| **concurrent test, pre-fix** (codegen reverted) | **1/3** — response stealing | 1 ❌ |

So the bug reproduces (1/3) and the fix resolves it (3/3) — the new test genuinely catches #1972. `cargo clippy -p dora-ros2-bridge-msg-gen -- -D warnings` and `cargo fmt --all -- --check` pass.

## Note

ROS2 action examples don't run in nightly CI (they're exercised via the release `scripts/ros2dev.sh qa` harness), so this regression test guards against reintroduction during ROS2 QA, not on every PR.


heyong4725:
Addressed: added `cxx-ros2-dataflow-action-client-concurrent` to the `scripts/ros2dev.sh` `EXAMPLES` list (commit e5f6a0b1), right after `cxx-ros2-dataflow-action-client`. Good catch — since the ROS2 action examples don't run in nightly, the `scripts/ros2dev.sh qa` harness is the only documented path that exercises this regression test, so it needs to be in that list to actually run.

PR #1997, opened 2026-06-02
fix(examples): make multiple-daemons daemon lifecycle robust

### LINKED ISSUE #1996: Daemon does not exit when its coordinator is permanently gone (lingering daemon processes)
## Summary

When a daemon's coordinator connection is permanently lost (e.g. the coordinator has shut down after a dataflow finished), the daemon does **not exit** — it retry-reconnects with exponential backoff up to 50 times (minutes) and keeps running. Spawned daemons therefore **linger as orphan processes** after the work is done.

This is the root cause of the intermittent `multiple-daemons` example flake (`Error: dataflow not finished after 101 retries` / hangs) and shows up in the nightly `Examples` job.

## Evidence

From a local run of `cargo run --example multiple-daemons`, after the dataflow finished and the in-process coordinator shut down, the daemon kept trying to reconnect:

```
WARN dora_daemon::coordinator: WS coordinator connection error: WebSocket protocol error: Connection reset without closing handshake
WARN dora_daemon: daemon disconnected from coordinator: failed to send watchdog message ... Attempting reconnect...
WARN dora_daemon::coordinator: Could not connect to WS at ws://127.0.0.1:6013/api/daemon: IO error: Connection refused (os error 61). Retrying in 1.064s (1/50)..
... Retrying in 2.253s (2/50)..  4s (3/50)..  8.134s (4/50)..  16s (5/50)..  32.397s (6/50)..  30s (7/50)..
```

`pgrep` afterwards showed multiple `target/release/dora daemon ...` (and a `dora runtime`) processes still alive long after the example "finished".

## Impact

1. **Hang on shutdown.** The `multiple-daemons` example awaits its daemon subprocesses (`tasks.join_next()`); because the daemons never exit, the example hangs until the CI timeout (observed as the exit-124 / "not finished" failure).
2. **Cross-invocation contamination.** A lingering daemon from a previous run re-registers with the *next* run's coordinator (`daemon re-registering (replacing stale connection)`), interfering with it.
3. **Orphan process leak** for any embedded/test orchestration that starts daemons and expects them to stop when their coordinator does.

## Reproduction

`cargo run --example multiple-daemons` (intermittent; more likely on a cold target / under load). After it finishes or fails, `pgrep -f "dora daemon"` shows lingering daemons.

## Question / proposed fix

Should a daemon **self-terminate when its coordinator connection is permanently lost** (e.g. after the reconnect budget is exhausted, or immediately on a clean coordinator shutdown), rather than retry-reconnecting indefinitely? At minimum, exhausting the 50-retry budget should exit the process instead of leaving it running.

A separate PR makes the `multiple-daemons` example robust at the harness level (unique `--local-listen-port` per daemon + explicitly killing the daemon subprocesses on completion), which works around this but does not address the underlying daemon-lifecycle behavior tracked here.

### Secondary (lower confidence)

On a cold target, the example's daemon occasionally failed to spawn nodes with `No such file or directory` for `target/debug/multiple-daemons-example-node` / the operator `.dylib` — likely cargo target-lock contention between the debug node `build:` commands and the `--release` daemon builds. May be environment-specific; noting in case it recurs.

heyong4725:
## Problem

The `multiple-daemons` example intermittently hangs or stalls — the nightly `Examples (ubuntu-latest)` flake (`Error: dataflow not finished after 101 retries`). Root-caused by reproducing locally (12-run loop, ~5 failures), with **0 deserialize errors** (so distinct from #1992).

Two harness-level problems in `run.rs`:

1. **Both daemons used the same hardcoded `--local-listen-port 9843`** (the comment literally said `// random port`). On one host, daemon B's local listener bind fails with `AddrInUse` (logs *"Daemon listen port already in use"*).
2. **The daemon subprocesses were awaited via the `JoinSet`.** A daemon does **not** promptly exit when its coordinator goes away — it retry-reconnects with backoff (`Retrying ... (7/50)`; tracked in #1996). So `tasks.join_next()` could hang on a non-exiting daemon (the exit-124 mode), and a lingering daemon from a failed run re-registered with the next run's coordinator (`daemon re-registering (replacing stale connection)`), contaminating it (the 101-retry stall).

## Fix (harness-level)

- Distinct `--local-listen-port` per daemon (`9843` / `9844`) → no `AddrInUse`.
- Spawn the daemons as **killable child processes** (removed from the `JoinSet`) and **stop them after the dataflow finishes** (+ `kill_on_drop` safety net). The example no longer waits on daemons that may not exit; on a normal shutdown they exit cleanly.

## Verification

Local back-to-back runs (host):

| | result |
|---|---|
| before | ~5/12 fail; lingering `dora daemon` processes; cross-run contamination |
| after | **6/6 pass, 0 lingering daemons** after every run |

`cargo check --example multiple-daemons` + `cargo fmt --all -- --check` pass.

## Scope

This makes the **example** reliable. It does **not** close #1996 — that issue tracks the underlying daemon behavior (a daemon should self-terminate when its coordinator is permanently gone instead of retry-reconnecting for minutes), which this PR only works around at the harness level.

### DIFF
--- a/examples/multiple-daemons/run.rs
+++ b/examples/multiple-daemons/run.rs
@@ -57,14 +57,13 @@ async fn main() -> eyre::Result<()> {
     tracing::info!("coordinator running on {coordinator_port}");
 
     let coordinator_addr = Ipv4Addr::LOCALHOST;
-    let daemon_a = run_daemon(coordinator_addr.to_string(), "A");
-    let daemon_b = run_daemon(coordinator_addr.to_string(), "B");
+    // Distinct ports: two daemons on one host can't share a local listener.
+    let mut daemon_a = run_daemon(coordinator_addr.to_string(), "A", 9843)?;
+    let mut daemon_b = run_daemon(coordinator_addr.to_string(), "B", 9844)?;
 
     tracing::info!("Spawning coordinator and daemons");
     let mut tasks = JoinSet::new();
     tasks.spawn(coordinator);
-    tasks.spawn(daemon_b);
-    tasks.spawn(daemon_a);
 
     tracing::info!("waiting until daemons are connected to coordinator");
     let mut retries = 0;
@@ -111,6 +110,16 @@ async fn main() -> eyre::Result<()> {
     }
     tracing::info!("dataflow `{uuid}` finished, destroying coordinator");
     destroy(&coordinator_events_tx).await?;
+
+    // The daemons are intentionally NOT in the JoinSet above: a daemon does not
+    // promptly exit when its coordinator goes away (it retry-reconnects with
+    // backoff; see dora-rs/dora#1996), so awaiting them would hang. On a normal
+    // shutdown they exit on their own once the coordinator broadcasts Destroy.
+    // This kill is a best-effort net for the `cargo run` wrappers (it does not
+    // reap the daemon grandchild; that's covered by graceful exit / #1996).
+    tracing::info!("stopping daemons");
+    let _ = daemon_a.kill().await;
+    let _ = daemon_b.kill().await;
 
     tracing::info!("joining tasks");
     while let Some(res) = tasks.join_next().await {
@@ -246,7 +255,11 @@ async fn build_dataflow(dataflow: &Path) -> eyre::Result<()> {
     Ok(())
 }
 
-async fn run_daemon(coordinator: String, machine_id: &str) -> eyre::Result<()> {
+fn run_daemon(
+    coordinator: String,
+    machine_id: &str,
+    local_listen_port: u16,
+) -> eyre::Result<tokio::process::Child> {
     let cargo = std::env::var("CARGO").unwrap();
     let mut cmd = tokio::process::Command::new(&cargo);
     cmd.arg("run");
@@ -259,9 +272,8 @@ async fn run_daemon(coordinator: String, machine_id: &str) -> eyre::Result<()> {
         .arg("--coordinator-addr")
         .arg(coordinator)
         .arg("--local-listen-port")
-        .arg("9843"); // random port
-    if !cmd.status().await?.success() {
-        bail!("failed to run dataflow");
-    };
-    Ok(())
+        .arg(local_listen_port.to_string());
+    // Best-effort cleanup if we exit before the explicit kill below.
+    cmd.kill_on_drop(true);
+    cmd.spawn().wrap_err("failed to spawn daemon")
 }


heyong4725:
Ran `/review` + `/simplify`.

**Review:** approve. Verified the JoinSet change is correct (after `destroy()` the coordinator aborts its stream and its task resolves → no hang), imports/borrows clean. Found + fixed one P2: the cleanup comments overstated what `kill` does — `run_daemon` spawns `cargo run -- daemon`, so `kill`/`kill_on_drop` target the `cargo` wrapper, not the daemon grandchild (could orphan it on a failure path). Comments now state this accurately; the real daemon-self-exit fix stays tracked in #1996. (Process-group reaping would need a new root dev-dep for one example — deferred to #1996 per the reviewer's accepted alternative.)

**Simplify:** efficiency — strictly *better* (removes the blocking JoinSet daemon tasks); reuse/altitude — right depth, consistent with sibling root examples. Applied: tightened the comments and de-duplicated the grandchild/#1996 caveat (kept the canonical note at the kill site).

`cargo check --example multiple-daemons` + `cargo fmt --all -- --check` pass; verified 6/6 back-to-back runs, 0 lingering daemons.

heyong4725:
Thanks — I tried both of your suggestions.

**Direct binary spawn regressed reliability.** I changed `run_daemon` to spawn `target/release/dora daemon` directly (so `kill`/`kill_on_drop` reap the daemon, not a `cargo` wrapper). It reaped cleanly (0 lingering, no spawn errors) **but** the dataflow stopped finishing — nodes reach "all nodes ready", then the node↔daemon TCP times out (`unexpected I/O error while trying to receive DaemonRequest`) → the 101-retry stall. Measured on the same box: **direct-binary 1/6 vs cargo-run 3/4** (and ~6/6 for cargo-run on a clean box). Spawning without the `cargo` wrapper shifts startup timing and aggravates the node-comm race, so it's a net loss. I've reverted to cargo-run.

**On the leak itself:** the failure mode you cite (lingering daemon re-registering with *the next run's* coordinator) requires two runs sharing a coordinator — i.e. back-to-back invocations in one environment, which is how I reproduced it locally. In CI each example runs **once in a fresh job/container**, so there is no next run to contaminate, and the process is reaped when the job ends. On a `bail!` path the orphaned daemon also self-exits once it exhausts its 50× reconnect budget. The single-run nightly flake this PR targets is the **hang** (daemon awaited in the JoinSet → fixed by removing it) plus the **port collision** — both addressed here.

The clean cross-cutting fix is the daemon **self-terminating when its coordinator is permanently gone**, which removes the leak (and the reconnect-forever behavior) for *all* callers without per-example process plumbing — tracked in #1996.

If you'd still prefer belt-and-suspenders reaping in the example, I can wrap the daemons with `process_wrap` `ProcessGroup::leader()` + an error-path drop guard (keeps the cargo-run timing, kills the whole group on every path). Happy to add it, but it isn't needed for the CI-flake goal. Let me know your preference.

heyong4725:
Merged on the cargo-run version (admin) since it's the reliable approach here — to recap for anyone following the thread:

- The direct-binary spawn (to reap the daemon instead of the `cargo` wrapper) **regressed reliability** in testing (1/6 vs cargo-run 3/4 on the same box; node↔daemon TCP timeouts from the shifted startup timing), so it wasn't a viable swap.
- This PR fixes the **CI-relevant** failure modes: the hang (daemon awaited in the JoinSet → removed) and the duplicate `--local-listen-port` collision.
- The remaining concern — a `bail!`-path daemon orphan re-registering with a *later* run's coordinator — only manifests with back-to-back runs sharing a coordinator (local dev); CI runs each example once in a fresh env, and the orphan self-exits after its reconnect budget.

The proper cross-cutting fix (daemon **self-terminates when its coordinator is permanently gone**) is tracked in **#1996**; once that lands, this example's manual cleanup can be simplified or removed. Thanks for the careful review — the process-group/drop-guard option stays on the table if we want belt-and-suspenders reaping before #1996.

examples/multiple-daemons/run.rs
```diff
@@ -57,14 +57,13 @@ async fn main() -> eyre::Result<()> {
     tracing::info!("coordinator running on {coordinator_port}");
 
     let coordinator_addr = Ipv4Addr::LOCALHOST;
-    let daemon_a = run_daemon(coordinator_addr.to_string(), "A");
-    let daemon_b = run_daemon(coordinator_addr.to_string(), "B");
+    // Distinct ports: two daemons on one host can't share a local listener.
+    let mut daemon_a = run_daemon(coordinator_addr.to_string(), "A", 9843)?;
+    let mut daemon_b = run_daemon(coordinator_addr.to_string(), "B", 9844)?;
```
phil-opp, line 62: (If I recall correctly, we just skip the listen on the dynamic node port if it's already in use. So using the default port should just result in a warning, not in any errors.)

examples/multiple-daemons/run.rs
```diff
@@ -112,6 +111,16 @@ async fn main() -> eyre::Result<()> {
...
+    // The daemons are intentionally NOT in the JoinSet above: a daemon does not
+    // promptly exit when its coordinator goes away (it retry-reconnects with
+    // backoff; see dora-rs/dora#1996), so awaiting them would hang. On a normal
+    // shutdown they exit on their own once the coordinator broadcasts Destroy.
+    // This kill is a best-effort net for the `cargo run` wrappers (it does not
+    // reap the daemon grandchild; that's covered by graceful exit / #1996).
+    tracing::info!("stopping daemons");
+    let _ = daemon_a.kill().await;
+    let _ = daemon_b.kill().await;
+
```
phil-opp, line 123: Daemons should exit normally after a destroy command is sent. So the daemons should have stopped already after `destroy` returns.

Given that we use this example as an integration test, we should verify this stop behavior. With this change (i.e. kill always), we no longer have a regression test for the proper destroy behavior.

The only reason for killing should be if we exit early from this function, e.g. through `bail`. But then the control flow never reaches this part of the code. Instead, the `kill_on_drop` in `run_daemon` should take care of the killing already.

PR #2005, opened 2026-06-03
fix(core): drop invalid uv flags from local wheel build (#2003)

### LINKED ISSUE #2003: dora build broken with current uv (0.8.x): 'uv build --clear --no-create-gitignore' uses invalid flags, fails managed Python env nodes
## Summary

`dora build` fails for any node that uses a managed Python env, because dora invokes `uv build` with flags that current `uv` (0.8.x) does not accept.

## Reproduction

```
$ dora build examples/python-async/dataflow.yml --uv
receive_data_with_sleep: stdout   error: unexpected argument '--clear' found
receive_data_with_sleep: stdout     tip: a similar argument exists: '--color'
receive_data_with_sleep: stdout   Usage: uv build --wheel --color <COLOR_CHOICE> [SRC]
[ERROR] failed to build node `receive_data_with_sleep`
Caused by:
   0: managed Python env preparation failed
   1: local Dora Python wheel build `apis/python/node` returned exit status: 2
```

## Root cause

`libraries/core/src/build/build_command.rs` (the local Dora Python wheel build, added in #1820/#1829) builds the command:

```rust
let mut cmd = Command::new("uv");
cmd.arg("build");
cmd.arg("--wheel");
cmd.arg("--clear");                 // <- not a `uv build` flag
cmd.arg("--no-create-gitignore");  // <- not a `uv build` flag
cmd.arg("--out-dir");
cmd.arg(&wheel_dir);
cmd.arg(&source.package_dir);
```

In uv 0.8.3, `uv build --help` lists neither `--clear` nor `--no-create-gitignore`. `--clear` is a `uv venv` flag (used correctly elsewhere in the same file at the `uv venv --clear` call). They were presumably valid on `uv build` in the uv version dora was developed against and have since been dropped/renamed. clap stops at the first unknown arg, so the error only names `--clear`, but both are invalid.

## Verified fix

Dropping both flags makes the build succeed:

```
$ uv build --wheel --out-dir /tmp/wheeltest apis/python/node
Successfully built /tmp/wheeltest/dora_rs-1.0.0rc1-cp311-abi3-macosx_11_0_arm64.whl
```

So the fix is to remove `--clear` and `--no-create-gitignore` from the `uv build` invocation (keep `--clear` only on `uv venv`). If older uv versions must be supported, version-gate instead.

## Impact

Release-relevant for 1.0.0-rc1: `dora build` is broken with current uv for every node using a managed Python env. It's masked in CI because the nightly example harness installs `-e apis/python/node` itself (bypassing dora's own wheel-build path), so the broken path is never exercised there. Found while auditing `scripts/smoke-all.sh` coverage — it accounts for 7 Python examples (python-async, python-drain, python-log, python-multiple-arrays, python-concurrent-rw, python-zero-copy-send + contract) failing to build.

## Environment

uv 0.8.3, macOS arm64, workspace 1.0.0-rc1.

heyong4725:
## Problem

Closes #2003.

The local Dora Python wheel build (the managed-env feature from #1820/#1829) in `libraries/core/src/build/build_command.rs` invokes `uv build` with `--clear` and `--no-create-gitignore`. Neither is a valid `uv build` flag in current uv (0.8.x) — `--clear` is a `uv venv` flag (used correctly elsewhere in the same file). uv exits `2` (`unexpected argument '--clear'`), so `dora build` fails for every node that uses a managed Python env:

```
failed to build node `receive_data_with_sleep`
  managed Python env preparation failed
  local Dora Python wheel build `apis/python/node` returned exit status: 2
```

Masked in CI: the nightly example harness installs `-e apis/python/node` itself, bypassing dora's own wheel-build path, so the broken path is never exercised.

## Fix

Drop the two invalid flags from the `uv build` invocation (keep `--clear` on `uv venv`).

## Verification

```
$ uv build --wheel --out-dir /tmp/wheeltest apis/python/node
Successfully built /tmp/wheeltest/dora_rs-1.0.0rc1-cp311-abi3-macosx_11_0_arm64.whl
```

After this fix, `dora build --uv` succeeds for managed-env examples (python-async, python-drain, python-zero-copy-send, …) and they run via `dora run --uv`. `cargo fmt`, `cargo clippy --all -- -D warnings`, and `dora-core` tests pass.

Found while auditing `scripts/smoke-all.sh` coverage for 1.0.0-rc1.

### DIFF
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -419,8 +419,10 @@ async fn ensure_local_dora_python_wheel(
     let mut cmd = Command::new("uv");
     cmd.arg("build");
     cmd.arg("--wheel");
-    cmd.arg("--clear");
-    cmd.arg("--no-create-gitignore");
+    // NOTE: `--clear` and `--no-create-gitignore` are NOT valid for `uv build`
+    // (current uv, 0.8.x) -- `--clear` belongs to `uv venv`. Passing them makes
+    // `uv build` exit 2 ("unexpected argument '--clear'"), which broke `dora
+    // build` for every node using a managed Python env. See dora-rs/dora#2003.
     cmd.arg("--out-dir");
     cmd.arg(&wheel_dir);
     cmd.arg(&source.package_dir);


heyong4725:
Closing — this was a false alarm on my end. The `--clear` / `--no-create-gitignore` flags ARE valid on `uv build` in current uv (verified present in uv 0.11.18); they were added after my **local** uv (0.8.3, Jul 2025), which is why `dora build --uv` failed for me locally. CI uses the latest uv, so the original code is correct — and removing the flags changed `uv build`'s behavior on CI's uv and broke `ci/circleci: cli-python`. Verified: original code + uv 0.11.18 builds the managed-env wheel and runs cleanly (`run exit=0`, 0 unexpected-argument errors). The real fix is to upgrade local uv, not change dora. See #2003.

Issue #2004, opened 2026-06-03
Networked 'dora build --uv' + 'dora start --uv' fails: managed Python env not recorded for start (only 'dora run' works)


heyong4725:
## Summary

For a Python node using a managed Python env (`--uv`), the networked flow `dora build --uv` + `dora start --uv` fails at start with "no managed env was recorded for it during build", even though `dora build` succeeded and created the env on disk. The single-process `dora run --uv` works, so only the separate-invocation (build then start) path is broken.

## Reproduction (manual, no test harness)

```
$ dora up
$ dora build examples/python-async/dataflow.yaml --uv      # exit 0
   send_data: stdout   Using Python 3.13.5 environment at: .dora/python-envs/send_data
   send_data: INFO     running build command: `pip install numpy pyarrow` ...
$ dora start examples/python-async/dataflow.yaml --detach --uv   # exit 1
   [ERROR] node `receive_data_with_sleep` is a Python node that needs a managed env
   under `--uv`, but no managed env was recorded for it during build. Re-run
   `dora build --uv <dataflow>` against the current build session, or omit `--uv`
   to run against the ambient Python.
   Location: binaries/daemon/src/lib.rs:2671:25
```

`dora build` clearly DID create the per-node managed envs under `.dora/python-envs/<node>/`, but `dora start` (a separate invocation) cannot find them recorded for the build session — so the build→start handoff of managed-env state is lost.

Contrast: `dora run examples/python-async/dataflow.yaml --uv --stop-after 10s` (single process, build+run together) works fine.

## Impact

Release-relevant for 1.0.0-rc1: networked `dora up` + `dora build --uv` + `dora start --uv` is the primary multi-process deployment path, and it's broken for every Python node with a managed env. It's masked in CI because the nightly example harness runs these examples via `dora run --uv` (local single-process), e.g. `scripts/qa/ci-nightly-jobs.sh` runs `dora run examples/python-async/dataflow.yaml --uv` — never the networked build→start path.

Reproduces for: python-async, python-drain, python-log, python-multiple-arrays, python-concurrent-rw, python-zero-copy-send (+ contract) — every managed-env example. (Found alongside #2003; #2003's `uv build --clear` bug previously masked this by failing at the build step.)

## Pointers

- `binaries/daemon/src/lib.rs:2671` — the "no managed env was recorded" error.
- Managed-env feature: #1820 / #1829.
[...]

Issue #2002, opened 2026-06-03
Release QA gates under-check for 1.0.0-rc1: semver.sh never fails post-1.0; unwrap-budget.sh awk truncates production code


heyong4725:
Two release-QA gate scripts report "green" while silently under-checking. Both matter more now that the workspace is `1.0.0-rc1` (API stability is the headline promise). Found during a pre-rc1 audit of `scripts/`.

## 1. `scripts/qa/semver.sh` never fails, even on breaking changes (post-1.0)

The script's own header says the SemVer check "becomes a hard gate post-1.0", but the code unconditionally exits 0:

```bash
if [[ "$FAILED" == "1" ]]; then
  echo "SemVer check found breaking changes."
  echo "During 0.x this is a warning. After 1.0 this will be a hard fail."
  # Soft warning during 0.x — exit 0 even on breaks
  exit 0
fi
```

The workspace is now `version = "1.0.0-rc1"` (root `Cargo.toml`), so the documented post-1.0 hard gate was never switched on. `scripts/qa/all.sh` wires `semver` into `--deep` / `--release-gate`, so the release gate prints "found breaking changes" and then passes. For a 1.0 release this should hard-fail.

**Fix:** exit non-zero on breakage when the workspace version no longer starts with `0.` (read it from `Cargo.toml`), or at minimum flip to `exit 1` for the rc1 line and update the comment.

## 2. `scripts/qa/unwrap-budget.sh` Rule 3 awk truncates production code after the first `#[cfg(test)]`

The unwrap counter truncates each file at the **first** `#[cfg(test)]` line and counts only above it:

```awk
/^[[:space:]]*#\[cfg\(test\)\]/ { exit }
```

This assumes `#[cfg(test)]` appears only once, at end of file. That's already violated by `binaries/cli/src/output.rs`:

```
320:#[cfg(test)]        <- a test helper block
330:fn calculate_char_sum(...)   <- PRODUCTION code, after the first cfg(test)
346:#[cfg(test)]
347:mod tests {
```

The awk `exit`s at line 320, so production code at 330-345 is excluded from the unwrap count. Today that span happens to have zero `.unwrap()`/`.expect(`, so the count is still correct — but any unwrap added to that interspersed production code is **silently excluded** from the ratchet. Latent ratchet defeat.

[...]

Issue #2001, opened 2026-06-03
examples/ros2-comparison: dataflow.yml references non-existent node scripts (dora_sender.py vs adora_sender.py)


heyong4725:
## Summary

`examples/ros2-comparison/dataflow.yml` references node scripts that do not exist, so the dataflow fails at start.

## Details

The dataflow declares:

```yaml
    path: dora_sender.py
    ...
    path: dora_receiver.py
```

But the actual files in `examples/ros2-comparison/` are:

```
adora_receiver.py
adora_sender.py
analyze.py
ros2_receiver.py
ros2_sender.py
```

There is no `dora_sender.py` / `dora_receiver.py` — the dora-side nodes are `adora_sender.py` / `adora_receiver.py`. Running the dataflow as-is errors on a missing node path.

## Fix

Update the two `path:` lines in `examples/ros2-comparison/dataflow.yml` to `adora_sender.py` / `adora_receiver.py` (or rename the files to match the YAML, whichever matches the intended `run_comparison.sh` wiring).

## Context

Found during a pre-1.0.0-rc1 audit of `scripts/smoke-all.sh` coverage. `ros2-comparison` needs ROS2/rclpy so it stays out of the smoke suite (now explicitly skipped), but the broken YAML is a real correctness bug independent of that.

PR #2006, opened 2026-06-03
test(scripts): smoke-all.sh coverage + robustness for 1.0.0-rc1

heyong4725:
Audit-driven hardening of `scripts/smoke-all.sh` (the human smoke runner) ahead of 1.0.0-rc1, plus a coverage fix to `scripts/ros2dev.sh`. Found via a multi-agent audit + repeated local runs.

## Why

The current `smoke-all.sh` could **report a false green or abort mid-run with no summary**, and its Python half didn't reliably test the *workspace* bindings. Concretely, before this PR a single failing local example aborted the whole run (no FAIL line, every later example skipped), and dep-less Python nodes ran against whatever `dora-rs` happened to be in the ambient env, version-mismatching the rc1 workspace daemon.

## Robustness fixes

- **`run_local` `set -e` abort** — guard `wait` (`wait "$pid" || rc=$?`) so a failing example logs FAIL and the run continues. Previously one bad example killed the run with no summary.
- **`run_local --verbose`** — run dora directly in the background instead of `| tee` (so `$pid` is dora's, not tee's; `wait` was returning tee's always-0 status and masking node failures, and the hard-kill hit tee, orphaning dora).
- **`run_networked`** — parse `dora list --json` stdout only; it exits non-zero (2) even on success, so the exit code is not a failure signal (matches the authoritative `tests/example-smoke.rs`).
- **`cleanup_stale`** — reap orphan repo-local coordinator/daemon and the fixed-port MAVLink bridge/sim nodes between examples (a lingering daemon holds port 6013, the bridge holds `udp:14550` → "address already in use" in the next example). Drop the redundant `destroy` (it's an alias of `down`).
- **Build/setup failures** — abort with a clear line + the failing output instead of `| tail -1` (which hid the error and died with no summary).

## Coverage fixes

- **Workspace Python bindings** — build + activate a venv with `-e apis/python/node` (uv + maturin) and run Python examples against it. Under `--uv`, dora only builds a per-node managed env (with the workspace dora wheel) for nodes that declare pip/build deps; a **dep-less Python node runs in the ambient Python env** and resolves whatever `dora-rs` is installed there — a stale editable install, nothing, or the PyPI package — none of which matches a 1.0.0-rc1 daemon (e.g. `message format vX.Y.Z is not compatible with v1.0.0-rc1`, or `ModuleNotFoundError`). The venv makes that ambient env carry the workspace bindings. Mirrors `scripts/qa/ci-nightly-jobs.sh`. (Verified in a clean venv: nodes that declare deps get a managed env with the workspace wheel automatically; dep-less nodes are the ones that need this venv.)
- **python-zero-copy-send** + its self-checking contract added — the zero-copy `send_output_raw` API was entirely unsmoked.
- **Explicit skips** for every remaining example dir (cpu-affinity-probe, error-propagation, rust/cxx-dynamic-add-remove, python-parquet-recorder, python-yolo-detection, mavlink2-bridge-sitl-mission, ros2-comparison) → **0 silently-uncovered examples** (was 9).
- Header now documents that PASS means "ran without crashing", not "output verified" (semantic assertions live in `cargo test --test example-smoke`).

**`scripts/ros2dev.sh`**: add the 3 ros2-bridge `[[example]]` targets missing from the release QA gate (rust + cxx action-server, rust parameter) so it can't report all-green while shipped examples go unexercised.

## Local run result

After these script fixes the runner completes reliably with a trustworthy summary. With a current `uv` and a clean ambient env, the only genuine product bug it surfaces is **#2004** — networked `dora build --uv` + `dora start --uv` loses the managed-env recording ("no managed env was recorded during build"); `dora run --uv` (local) works. Verified against the latest uv (0.11.18), so it's a real handoff bug, not a tooling artifact.

It also surfaced #2001 (ros2-comparison broken YAML) and #2002 (semver.sh / unwrap-budget.sh QA-gate logic).

Note: an earlier draft of this PR referenced a `dora build` "invalid uv flag" bug — that was a false alarm from a stale local uv (0.8.3) and has been retracted (#2003 closed). `smoke-all.sh` does not pass any `uv build` flags and is unaffected.

`bash -n` clean.

### DIFF
--- a/scripts/ros2dev.sh
+++ b/scripts/ros2dev.sh
@@ -108,6 +108,10 @@ cargo test -p dora-ros2-bridge-python
 # Add new examples here as issue #1170 items land (action server, C++ service
 # server, parameter service, ...). `--features ros2-examples` is required by the
 # cxx/action targets and harmless for the rest.
+# Keep in sync with the [[example]] targets in
+# libraries/extensions/ros2-bridge/Cargo.toml -- every shipped ros2-bridge
+# example should be exercised by the release QA gate, or the gate reports
+# "all green" while real surfaces go untested.
 EXAMPLES=(
   rust-ros2-dataflow
   rust-ros2-dataflow-topic-pub
@@ -115,6 +119,8 @@ EXAMPLES=(
   rust-ros2-dataflow-service-client
   rust-ros2-dataflow-service-server
   rust-ros2-dataflow-action-client
+  rust-ros2-dataflow-action-server
+  rust-ros2-dataflow-parameter
   python-ros2-dataflow
   python-ros2-dataflow-service-client
   python-ros2-dataflow-service-server
@@ -125,6 +131,7 @@ EXAMPLES=(
   cxx-ros2-dataflow-action-client
   cxx-ros2-dataflow-action-client-concurrent
   cxx-ros2-dataflow-service-server
+  cxx-ros2-dataflow-action-server
 )
 
 passed=()
--- a/scripts/smoke-all.sh
+++ b/scripts/smoke-all.sh
@@ -5,7 +5,13 @@
 #
 # Networked lifecycle:
 #   dora up -> dora build -> dora start --detach -> poll dora list
-#   -> dora stop -> dora destroy -> dora down
+#   -> dora stop -> dora down
+#
+# What PASS means: the dataflow ran without crashing (no "Failed" state /
+# nonzero exit) for the smoke window. It does NOT assert correct OUTPUT --
+# the semantic contract assertions live in the Rust suite
+# (`cargo test --test example-smoke`, the contract_* tests). Treat a green
+# smoke-all.sh as "nothing blew up", not "output verified".
 #
 # Usage:
 #   ./scripts/smoke-all.sh               # Run all examples
@@ -13,10 +19,20 @@
 #   ./scripts/smoke-all.sh --python-only # Python examples only
 #   ./scripts/smoke-all.sh --verbose     # Stream dora stdout/stderr live
 #
-# Prerequisites: cargo, Python 3 with pyarrow + numpy installed.
+# Prerequisites: cargo. For Python examples also `uv` and `maturin` -- the
+# script builds the WORKSPACE Python bindings into target/smoke-venv so dep-less
+# Python nodes (which run in the ambient env, not a per-node managed env) use the
+# workspace dora instead of whatever `dora-rs` is otherwise installed. If uv or
+# maturin is missing, Python examples are still attempted but may fail.
 # Skips examples that need webcam, CUDA, ROS2, C/C++ toolchain, etc.
 
 set -euo pipefail
+
+# Surface the failing line if a build/setup step aborts the run (set -e).
+# Example pass/fail is handled explicitly via log_fail; this only fires on an
+# unguarded failure such as a broken cargo build, which would otherwise abort
+# with no summary and no obvious cause.
+trap 'rc=$?; echo; echo "smoke-all.sh: aborted (exit $rc) near line $LINENO -- likely a build/setup failure above"; exit $rc' ERR
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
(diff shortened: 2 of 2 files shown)


phil-opp:
## Reliability review of `smoke-all.sh`

The headline fixes are solid (the `tee`/`$pid` exit-status masking fix in `run_local`, the `dora list --json` exit-2 handling, and the guarded `wait`). Verified that `down` literally has `alias = "destroy"` (`binaries/cli/src/command/mod.rs:76`), so dropping the `destroy` call is correct. A few remaining sharp edges:

### 1. `cleanup_stale`'s `pkill` is repo-global, not run-scoped (highest concern)
`scripts/smoke-all.sh:132-140` kills **any** process matching `$ROOT/target/debug/dora-(coordinator|daemon|runtime)` etc. — not just processes this script spawned. So:
- A manual `dora up` from the same checkout (unrelated testing) gets SIGKILLed the moment `smoke-all.sh` runs.
- Two `smoke-all.sh` invocations (or smoke-all + the Rust `example-smoke` suite) on the same machine reap each other's daemons → spurious FAILs.

The old `stop --all` / `down` were polite (talk to the coordinator); `pkill` is blunt over the whole checkout. At minimum add a header caveat ("assumes exclusive use of this checkout's dora ports"); ideally scope the kill to PIDs this script started.

### 2. `run_local` hard-kill never calls `cleanup_stale`
The orphan-reaping that motivated `cleanup_stale` (fixed-port mavlink on `udp:14550`, lingering daemons on 6013) only runs in the **networked** path. The local hard-kill path (`scripts/smoke-all.sh:316-317`) does only `pkill -9 -P "$pid"` (direct children) + `kill -9 "$pid"`. A hard-killed **local** example that orphaned a fixed-port node leaves it for the *next networked* example to clean — back-to-back local runs (the local Python block, `local-mavlink2-bridge-python`) aren't protected, so the next local run can hit "address already in use." Call `cleanup_stale` after a local hard-kill to close the gap.

### 3. Top-level `cargo build … 2>&1 | tail -1` truncates the failure it's meant to surface
The PR says build/setup failures now print "the failing output instead of `| tail -1`", but that only applies to the per-example `run_dora_step` path. The CLI/node builds at `scripts/smoke-all.sh:352-410` still pipe through `| tail -1`. `pipefail` correctly propagates the failure to the ERR trap, but you're left with **one line** of cargo output + "likely a build/setup failure above" — and there's nothing useful above, because `tail -1` ate it. Consider tee-ing to a file and dumping the tail on failure, like the example path does.

### 4. `needs_uv` now silently depends on `python3`, which was dropped from prerequisites
`scripts/smoke-all.sh:144-152` shells out to `python3` for **every** example (Rust and Python) to detect `.py`. The new prereq comment lists only `cargo` + `uv` + `maturin` and removed the old "Python 3" line. If `python3` is absent, `needs_uv` exits 127 → the `if` guard reads it as "no uv" → a Python example runs **without `--uv`** and fails confusingly. Either keep `python3` in the prereqs, or make `needs_uv` a pure-bash `grep`.

### 5. Sub-2s false-PASS window in `run_networked` (minor, pre-existing)
The poll loop sleeps 2s before the first `dora list` (`scripts/smoke-all.sh:252-260`). A networked dataflow that crashes and is reaped within 2s shows no `Running`/`Failed` → loop breaks → **PASS**. Runtime node crashes in networked mode also aren't in `$logfile` (only `dora start --detach` output is), so `dump_tail` on a networked FAIL may not show the cause. Inherent smoke-runner limits, now honestly disclaimed by the new "PASS means nothing blew up" header — flagging for awareness, not as a regression.

### Also: leftover placeholder
`scripts/smoke-all.sh:576` — `... see #issue)` is a literal placeholder; should be `#2001` per the PR description.

### Checked and cleared
- `destroy` → `down` removal (verified alias).
- `wait`/`PIPESTATUS` exit-status capture in `run_local` and `run_dora_step`.
- `[ … ] && repin_python_bindings` under `set -e`/ERR trap — exempt, no spurious aborts.
- `pkill -f` ERE patterns — valid, won't match `dora list`/`dora stop`.
- Per-function `trap … RETURN` cleanup — not inherited by nested calls; benign regardless.

**Net:** #1 and #2 are the two I'd address before relying on this in a shared environment; #3/#4 are easy diagnostic wins. None block the PR's stated goal.


Created with Claude Code

PR #2014, opened 2026-06-03
fix(daemon): re-derive managed Python env at start instead of bailing (#2004)

### LINKED ISSUE #2004: Networked 'dora build --uv' + 'dora start --uv' fails: managed Python env not recorded for start (only 'dora run' works)

heyong4725:
## Summary

Closes #2004. Networked `dora build --uv` + `dora start --uv` failed at start with:

```
node `receive_data_with_sleep` is a Python node that needs a managed env under `--uv`,
but no managed env was recorded for it during build.
  binaries/daemon/src/lib.rs:2681
```

even though `dora build` succeeded and created the managed env on disk. `dora run --uv` (single-process) was unaffected.

## Root cause

Instrumented the daemon: at `dora start`, the daemon has the **correct `build_id`** but `self.builds` is **empty** — and no build-insert ever ran on it. A **networked build runs in a separate process** from the daemon that later serves `dora start` (the build log is the client's, with `target/dora-python-wheels/process-<pid>/…`), so the daemon's in-memory build record (`self.builds`, which carries `python_env_dirs`) is empty at start. The start path required that record to locate each node's managed env and bailed when it was missing. `dora run --uv` builds and starts in one process, so the record is present — which is why it worked.

## Fix

The managed env dir is **deterministic** (`managed_python_env_dir(node, working_dir)`), and the **node-restart path already re-derives it** without consulting `self.builds` (lib.rs:1715). So `dora start` was the inconsistent one. Now, when the build record lacks a node's env, start re-derives the path and **uses it if `dora build` actually created it on disk** (the venv interpreter is present). It fails closed — with a clear, path-named error — only when the env is genuinely missing (stale session, prior non-`--uv` build, or `start --uv` with no build), so `--uv` never silently falls back to the ambient Python.

This also generalizes correctly to multi-daemon deploys: each daemon builds its own nodes' envs locally and re-derives them at start on the same machine.

## Verification (foreground coordinator + daemon, workspace bindings)

| Scenario | Before | After |
|---|---|---|
| networked `build --uv` then `start --uv` | `start exit=1`, "no managed env recorded" | **`start exit=0`**, nodes spawn against the workspace env (no ModuleNotFound / version mismatch) |
| `start --uv` with **no** prior build (env absent) | n/a | **fails closed** — `… needs a managed env … but none was found at <path>. Run dora build --uv …` |

`cargo fmt`, `cargo clippy -p dora-daemon -- -D warnings`, and `dora-daemon` tests pass.

(Note: on this arm64 box the python-async benchmark itself is limited by a separate, pre-existing arm64 zenoh "misaligned … / event channel full" issue, unrelated to this fix — the nodes spawn and run with the correct managed env, which is what #2004 is about.)

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2466,6 +2466,10 @@ impl Daemon {
 
         let mut stopped = Vec::new();
 
+        // A present `build_id` means the session has a build that wasn't
+        // invalidated (the CLI clears it when build inputs change). We use this
+        // below to decide whether re-deriving an on-disk managed env is safe.
+        let have_build_id = build_id.is_some();
         let build_info = build_id.and_then(|build_id| self.builds.get(&build_id));
         let node_with_git_source = nodes.values().find(|n| n.has_git_source());
         if let Some(git_node) = node_with_git_source
@@ -2664,24 +2668,29 @@ impl Daemon {
                 // process from the daemon that later serves `dora start`, so the
                 // daemon's in-memory build record can be empty (dora-rs/dora#2004).
                 // The env dir is deterministic, so re-derive it (the restart path
-                // does the same), and use it IF `dora build` actually created it on
-                // disk. Fail closed only when the env is genuinely missing (stale
-                // session, prior non-`--uv` build, or `start --uv` without a build)
-                // so `--uv` never silently falls back to the ambient Python.
+                // does the same) and reuse the on-disk env -- but ONLY when a
+                // build id is present, i.e. a build occurred and this daemon just
+                // lacks the in-memory record. When the build id was cleared (the
+                // CLI invalidates it on build-input changes, a prior non-`--uv`
+                // build, or `start --uv` with no build), do NOT reuse a possibly
+                // stale env; require a rebuild. Either way, never silently fall
+                // back to the ambient Python.
                 if uv
                     && configured_python_env_dir.is_none()
                     && let Some(expected) =
                         dora_core::build::managed_python_env_dir(&node, &node_working_dir)
                 {
-                    if dora_core::build::managed_python_interpreter(&expected).is_file() {
+                    if have_build_id
+                        && dora_core::build::managed_python_interpreter(&expected).is_file()
+                    {
                         configured_python_env_dir = Some(expected);
                     } else {
                         eyre::bail!(
                             "node `{node_id}` is a Python node that needs a managed env under `--uv`, \
-                             but none was found at `{}`. \
+                             but no current build provides one (the build cache is absent or was \
+                             invalidated by changed build inputs). \
                              Run `dora build --uv <dataflow>` before `dora start --uv`, \
-                             or omit `--uv` to run against the ambient Python.",
-                            expected.display()
+                             or omit `--uv` to run against the ambient Python."
                         );
                     }
                 }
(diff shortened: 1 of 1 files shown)


heyong4725:
Good catch — fixed in `a4018f57`.

You're right: the re-derive fallback fired whenever the daemon lacked a recorded env, which also covers the invalidation case (`build_id` cleared because build inputs changed) — reusing the on-disk `.dora/python-envs/<node>` there would run stale artifacts and bypass the required rebuild.

Now the reuse is gated on a present `build_id` (`have_build_id`): re-derive + reuse the on-disk env **only** when a build occurred and this daemon merely lacks the in-memory `BuildInfo` (the #2004 networked build→start case). When `build_id` was cleared (invalidation / prior non-`--uv` build / `start --uv` with no build), it fails closed and requires a rebuild.

Verified both paths against the same on-disk env:

```
# build_id present + env on disk  -> start exit 0 (re-derive, no bail)   [#2004 fix]
# build_id cleared + env on disk  -> start exit 1, bails:                [your case]
#   \"... needs a managed env under --uv, but no current build provides one
#    (the build cache is absent or was invalidated by changed build inputs).
#    Run dora build --uv ...\"
```

PR #2026, opened 2026-06-04
test(mavlink2-bridge): cover write-path Arrow round-trips

### LINKED ISSUE #2023: mavlink2-bridge: write-path (dora→vehicle) Arrow conversions have no round-trip test coverage; silent same-type field-reorder regressions can ship undetected
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. It was found by randomly selecting a component (`mavlink2-bridge`) and reviewing it for correctness issues. Please treat the severity assessment below as a starting point for a human triage, not a confirmed defect.

## Summary

The `MavlinkArrow` conversions in `libraries/extensions/mavlink2-bridge/src/arrow_convert.rs` are otherwise well-built — every impl is complete (struct-literal exhaustiveness) and symmetric (`schema`/`to_record_batch`/`from_record_batch` share the same widths, enforced by the Rust type checker), and the read path is defensively guarded against malformed input (`read_primitive` checks dtype, zero rows, and null-at-row-0).

However, the **command / write path** (dora → MAVLink, the side that actually arms and moves a vehicle) has **zero round-trip test coverage**, while **every telemetry (read-path) message is round-trip tested**. The four untested impls are:

- `SET_MODE_DATA`
- `RC_CHANNELS_OVERRIDE_DATA`
- `SET_POSITION_TARGET_GLOBAL_INT_DATA`
- `SET_POSITION_TARGET_LOCAL_NED_DATA`

Verified:
- `tests/arrow_roundtrip.rs` imports/tests only the 13 telemetry types — none of the four above.
- `tests/arrow_malformed.rs` exercises negative decoding only for `COMMAND_LONG_DATA`; the four write-path types have neither round-trip nor malformed coverage.
- `grep -rn "SET_MODE\|RC_CHANNELS_OVERRIDE\|SET_POSITION_TARGET" libraries/extensions/mavlink2-bridge/tests binaries/mavlink2-bridge-node/tests` → **no matches**.

## Why this is a correctness risk, not just a coverage gap

Each impl maintains the schema field list and the column-builder list as **two parallel, manually-ordered lists**. `build()` hands them to `RecordBatch::try_new`, which validates **column count and dtype** but **not semantic field identity**. So if two fields of the **same Arrow type** are transposed between the schema list and the column list, the mismatch is:

- not a compile error (types match),
- not caught by `RecordBatch::try_new` (dtypes match),
- not caught by any test (none exist for these impls),

and the bridge would silently send the wrong value to the vehicle (e.g. a velocity component into an acceleration slot, or `x`↔`y`).

These two messages are the worst case for that failure mode because they carry long runs of adjacent same-type `Float32` fields:

`SET_POSITION_TARGET_LOCAL_NED_DATA` — 11 consecutive `Float32` fields (`arrow_convert.rs:1000`–`1048`):
```rust
Field::new("x",  DataType::Float32, false),
Field::new("y",  DataType::Float32, false),
Field::new("z",  DataType::Float32, false),
Field::new("vx", DataType::Float32, false),
Field::new("vy", DataType::Float32, false),
Field::new("vz", DataType::Float32, false),
Field::new("afx", DataType::Float32, false),
Field::new("afy", DataType::Float32, false),
Field::new("afz", DataType::Float32, false),
Field::new("yaw", DataType::Float32, false),
Field::new("yaw_rate", DataType::Float32, false),
```
`SET_POSITION_TARGET_GLOBAL_INT_DATA` has 9 adjacent `Float32` (`alt, vx, vy, vz, afx, afy, afz, yaw, yaw_rate`) at `arrow_convert.rs:916`–`992`.

The current code **is** consistent (I verified schema order == column order for all four impls), so there is **no live bug today** — this is about a latent footgun on a safety-relevant path that is wholly untested, where a future copy-paste/reorder edit would regress silently.

## Suggested remediation

1. Add round-trip tests (mirroring the existing telemetry tests in `tests/arrow_roundtrip.rs`) for all four write-path impls, asserting **every** field with **distinct sentinel values per field** (so a transposition actually fails the assert — note several existing telemetry tests assert only a subset of fields, e.g. `sys_status_arrow_roundtrip` skips `drop_rate_comm`/`errors_count*`, which would not catch a same-type swap either).
2. Optionally remove the structural fragility by deriving the column vector and the schema from a single source of truth, so the two lists cannot drift. For example, have each impl build `(Field, ArrayRef)` pairs once and split them in `build()`, instead of calling `Self::schema()` and listing columns separately.
3. Consider a `#[test]` that asserts, for every `MavlinkArrow` impl, `to_record_batch().schema() == schema()` and `num_rows() == 1`, to lock the encode side to its declared schema.

## Affected files
- `libraries/extensions/mavlink2-bridge/src/arrow_convert.rs` (impls at lines ~860, ~649, ~916, ~1000)
- `libraries/extensions/mavlink2-bridge/tests/arrow_roundtrip.rs` (telemetry-only)
- `libraries/extensions/mavlink2-bridge/tests/arrow_malformed.rs` (`COMMAND_LONG` only)

- **Field widths vs. MAVLink wire spec:** safe by construction — `arr_*(self.field)` only compiles if the Arrow width matches the generated struct field, which matches the wire type. `SERVO_OUTPUT_RAW.time_usec` is correctly `UInt32` (a known MAVLink quirk).
- **Outgoing sequence number always 0** (`MavHeader { sequence: 0 }` reused in `binaries/mavlink2-bridge-node/src/main.rs`): not a bug — rust-mavlink 0.13's connection `send()` overrides the header sequence with an internal wrapping counter.
- **Shutdown / reader-thread wake logic** (TCP read-timeout, UDP self-loopback heartbeat, serial grace+detach): reviewed, appears correct and well-tested.

imajij:
## Summary

The four command/write-path `MavlinkArrow` impls — the side that arms and moves a vehicle — had **zero** round-trip test coverage, while every telemetry message was tested:

- `SET_MODE_DATA`
- `RC_CHANNELS_OVERRIDE_DATA`
- `SET_POSITION_TARGET_GLOBAL_INT_DATA`
- `SET_POSITION_TARGET_LOCAL_NED_DATA`

This is a **latent footgun, not a live bug** (the impls are consistent today). Each impl maintains the schema field list and the column-builder list as two parallel, manually-ordered lists. Transposing two same-Arrow-type fields between them is **not** a compile error and is **not** caught by `RecordBatch::try_new` (the dtypes still match), so a future copy-paste/reorder edit could silently send e.g. a velocity into an acceleration slot, or swap x↔y. The `SET_POSITION_TARGET` messages are the worst case — 9 and 11 adjacent `Float32` fields.

## Changes (test-only, no production code change)

- Round-trip tests for all four write-path impls, asserting **every** field with a **distinct sentinel value per field** so a transposition actually fails the assert (reusing one value across the adjacent `Float32` runs would let a swap pass silently).
- `encode_matches_declared_schema`: locks **every** impl's `to_record_batch()` to its declared `schema()` (field names, order, dtypes, nullability) and to exactly one row.

## Verification

- Confirmed the tests are effective: a temporary `afx`↔`afy` swap in the source made the local-NED round-trip fail (`left: 8.0, right: 7.0`); reverting restored green.
- `cargo clippy -p dora-mavlink2-bridge --tests -- -D warnings` ✓
- `cargo test -p dora-mavlink2-bridge` ✓ (18 round-trip tests pass)

## Related issue

Closes #2023.

This implements the two test-based remediations from #2023: (1) round-trip tests for all four write-path impls with distinct per-field sentinels, and (3) a per-impl `to_record_batch().schema() == schema()` + single-row lock. The dangerous gap — a safety-relevant write path with zero coverage where a silent same-type transposition could ship undetected — is now closed and the tests are verified to catch it.

The issue's optional remediation (2) — eliminating the parallel schema-list / builder-list fragility structurally by deriving both from a single source of truth — is **deferred**. With the regression now caught by tests and no live bug today, the structural refactor is a separate, non-urgent follow-up.

Found by a scheduled automated code-review check on the `mavlink2-bridge` component.

### DIFF
--- a/libraries/extensions/mavlink2-bridge/tests/arrow_roundtrip.rs
+++ b/libraries/extensions/mavlink2-bridge/tests/arrow_roundtrip.rs
@@ -7,9 +7,11 @@ use dora_mavlink2_bridge::MavlinkArrow;
 use dora_mavlink2_bridge::mavlink::common::{
     ATTITUDE_DATA, ATTITUDE_QUATERNION_DATA, COMMAND_ACK_DATA, COMMAND_LONG_DATA,
     GLOBAL_POSITION_INT_DATA, GPS_RAW_INT_DATA, GpsFixType, HEARTBEAT_DATA,
-    LOCAL_POSITION_NED_DATA, MISSION_CURRENT_DATA, MavAutopilot, MavCmd, MavModeFlag, MavResult,
-    MavState, MavSysStatusSensor, MavType, RC_CHANNELS_DATA, SERVO_OUTPUT_RAW_DATA,
-    SYS_STATUS_DATA, SYSTEM_TIME_DATA,
+    LOCAL_POSITION_NED_DATA, MISSION_CURRENT_DATA, MavAutopilot, MavCmd, MavFrame, MavMode,
+    MavModeFlag, MavResult, MavState, MavSysStatusSensor, MavType, PositionTargetTypemask,
+    RC_CHANNELS_DATA, RC_CHANNELS_OVERRIDE_DATA, SERVO_OUTPUT_RAW_DATA, SET_MODE_DATA,
+    SET_POSITION_TARGET_GLOBAL_INT_DATA, SET_POSITION_TARGET_LOCAL_NED_DATA, SYS_STATUS_DATA,
+    SYSTEM_TIME_DATA,
 };
 
 #[test]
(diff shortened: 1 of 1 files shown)


heyong4725 (approved):
Reviewed via /review. Test-only, no production change. Verified locally: 18/18 mavlink2-bridge tests pass, clippy -D warnings clean, fmt clean. Adversarially mutation-tested — swapping afx/afy in the GLOBAL_INT builder failed the round-trip (left 5.0, right 4.0), confirming the distinct-sentinel design catches the transposition footgun. Closes #2023 (test remediations #1 + #3; structural refactor #2 deferred as optional).

PR #2038, opened 2026-06-05
fix(ros2-bridge): bound FFIWString::to_rust read by size

imajij:
## Summary

`FFIWString::to_rust` in `libraries/extensions/ros2-bridge/src/_core/string.rs` determined the wide-string length by scanning for a NUL terminator via `U16CStr::from_ptr_str`, **ignoring the `size` field the struct already carries**. Its 8-bit sibling `FFIString::to_str` was deliberately hardened to bound the read by `size` (with a comment explaining why), so the two conversion paths were inconsistent and the wide path carried two hazards:

- **Silent truncation of valid data.** A ROS2 wstring / `U16String` can legitimately contain an embedded `U+0000` — the rosidl C representation stores `data + size + capacity` precisely so interior NULs are representable. `from_ptr_str` stops at the first NUL, so any wstring with an embedded `U+0000` is silently truncated on the C++ → Rust conversion.
- **Unbounded read risk.** If a producer ever hands over a buffer that isn't NUL-terminated, `from_ptr_str` reads past the end of the buffer — exactly the hazard the `FFIString` hardening comment calls out, in `unsafe` code.

## Fix

Bound the read by `size`, building the `U16String` from a size-bounded slice, mirroring the hardened 8-bit path:

```rust
let slice = unsafe { std::slice::from_raw_parts(self.data, self.size) };
U16String(widestring::U16String::from_vec(slice.to_vec()))
```

This preserves embedded NULs and never reads beyond `size` elements. Also drops the now-unused `U16CStr` import.

## Test

Adds `ffi_wstring_preserves_embedded_nul`, which round-trips an `['A', NUL, 'B']` buffer constructed directly (`OwnedFFIWString::from_rust` can't produce this case — `U16CString` rejects interior NULs). The existing test only covered a NUL-free string.

## Verification

- New test **fails against the old `from_ptr_str` code** (`left: "A"` vs `right: "A\0B"` — the truncation) and passes with the fix, so it genuinely guards the regression.
- `cargo test -p dora-ros2-bridge --lib _core::string` — 3/3 pass.
- `cargo clippy -p dora-ros2-bridge --lib` ✓ (only environmental `AMENT_PREFIX_PATH` build-script warnings).

Found by a scheduled automated code-review check on the `ros2-bridge` component. The change strictly matches the already-accepted `FFIString` hardening; a reviewer may still want to confirm the rosidl wstring contract (that `size` counts code units excluding the terminator and interior NULs are valid).

### DIFF
--- a/libraries/extensions/ros2-bridge/src/_core/string.rs
+++ b/libraries/extensions/ros2-bridge/src/_core/string.rs
@@ -4,7 +4,7 @@ use std::{
     os::raw::c_char,
 };
 
-use widestring::{U16CStr, U16CString};
+use widestring::U16CString;
 
 use super::traits::{FFIFromRust, FFIToRust};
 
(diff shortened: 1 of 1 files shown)


phil-opp:
Sorry, I didn't see your PR in time and fixed the issue already in #2041. Thank you for submitting it!

PR #2048, opened 2026-06-05
fix(core): resolve uv-managed paths synchronously and return absolute path

### LINKED ISSUE #2016: Bug: `resolve_path` fire-and-forget `uv run which` spawn — exit code never checked, returns relative path
## Summary

`resolve_path` in `libraries/core/src/descriptor/mod.rs` has a broken fallback branch for environments where `uv` is installed. The spawned child process is never waited on, so its exit code (pass/fail) is never inspected, and the function unconditionally returns a **relative** path instead of an absolute one.

## Affected code

`libraries/core/src/descriptor/mod.rs`, lines 366–382:

```rust
} else if which::which("uv").is_ok() {
    // spawn: uv run which <path>
    let which = if cfg!(windows) { "where" } else { "which" };
    let _output = Command::new("uv")
        .arg("run")
        .arg(which)
        .arg(&path)
        .stdout(Stdio::null())
        .spawn()
        .context("Could not find binary within uv")?;
    Ok(path)        // ← returns the ORIGINAL (possibly relative) path
```

## Problems

1. **Exit code is never checked.** `spawn()` only verifies that the `uv` executable itself could be launched. If `uv run which <path>` exits with code 1 (binary not found), that failure is silently swallowed. The function returns `Ok(path)` even when the binary does not exist anywhere on the system.

2. **Returns a relative path.** Every other branch returns an absolute path via `canonicalize()` or `which::which()`. This branch returns the original relative `path` value, which is inconsistent and may break callers that expect an absolute path.

3. **`tokio::process::Command` used in a non-`async` function.** The import is `tokio::process::Command` but `resolve_path` is a synchronous `fn`. The process handle is immediately dropped (prefixed `_output`), so tokio never drives it to completion.

## Minimal reproducer

On any machine with `uv` installed:

```yaml
# dataflow.yaml
nodes:
  - id: node
    path: this_binary_does_not_exist
```

```
dora check dataflow.yaml
# Expected: error — binary not found
# Actual:   no error — resolve_path returns Ok("this_binary_does_not_exist")
```

## Expected fix

The branch should:
- Wait for the child and check its exit status, **or**
- Use `which::which` with the `uv`-managed environment (e.g. via `uv run -- which` and capture stdout), **or**
- Remove the `uv` special-case entirely if it cannot be reliably implemented in a sync context.

The returned path must be absolute (matching the contract of all other branches).

--- 

This issue was created by an scheduled automated Claude check.

phil-opp:
The `uv` fallback in `resolve_path` used `tokio::process::Command` inside a sync fn, spawned `uv run which <path>` fire-and-forget (exit status never checked, child immediately dropped), and returned the original relative path — so a missing binary was silently accepted and the path was inconsistent with the absolute-path contract of the sibling branches. It now runs `uv run which` synchronously via `std::process::Command`, checks the exit status so a missing binary errors out, and canonicalizes the captured location into an absolute path.

Closes #2016.

### DIFF
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -8,9 +8,8 @@ use std::{
     collections::{BTreeMap, HashMap},
     env::consts::EXE_EXTENSION,
     path::{Path, PathBuf},
-    process::Stdio,
+    process::Command,
 };
-use tokio::process::Command;
 
 // reexport for compatibility
 pub use dora_message::descriptor::{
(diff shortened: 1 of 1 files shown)


heyong4725 (approved):
Reviewed — the fix looks correct and is a real improvement. The old `uv` fallback used `tokio::process::Command` inside a sync fn, did `.spawn()` fire-and-forget (child dropped immediately, exit status never checked), and returned the original relative path — so a missing binary was silently accepted and the result broke the absolute-path contract of the sibling branches. The replacement runs `uv run which` synchronously, checks the exit status, and canonicalizes to an absolute path. Well-documented, with on-point regression tests.

I verified the runtime behavior locally: `uv run which <missing>` → exit 1 + empty stdout (→ `bail`), `uv run which python` → exit 0 + path (→ canonicalize); both fast, no hang risk.

**On the failing `ci/circleci: contract-tests` — it is not caused by this PR.** The modified branch only runs for a non-`.py`, non-file, uv-present source. Every contract-test dataflow uses either `.py` files (caught by the first `canonicalize` branch) or `build:` rust binaries (validation skips the path check for `build:` nodes, `validate.rs:88`), so none traverse the changed code. I confirmed empirically: `dora run examples/python-dataflow/dataflow.yml --uv` with this branch resolves the nodes via the `.py` branch (`spawning: uv run python -u /abs/.../sender.py`) and never invokes `resolve_path_via_uv`. The job also ran ~51 min (steps cap at 30m + 15m `no_output_timeout`), which is the signature of a hang/timeout rather than a clean assertion failure from a logically-isolated 1-file change.

Recommend just re-running the `contract-tests` job. LGTM otherwise.

Issue #2028, opened 2026-06-04
coordinator: daemon-disconnect cleanup only logs → three deadlocks (ready barrier, hung `dora stop`, orphaned dataflow)


heyong4725:
## Summary

`cleanup_disconnected_daemons_from_running_dataflows` (`binaries/coordinator/src/lib.rs:2899-2917`) removes a disconnected daemon from the in-memory sets and logs — but it never re-evaluates the readiness barrier, drains the stop waiters, or tears the dataflow down. Three failure modes share this single root cause:

- **Ready-barrier deadlock** — `binaries/coordinator/src/lib.rs:479-507` / `2899-2916`. If the last pending daemon disconnects after spawn success but before `ReadyOnDaemon`, `pending_daemons` empties via cleanup but `AllNodesReady` never broadcasts; surviving nodes block forever on the start barrier.
- **`dora stop` waiter hangs** — `binaries/coordinator/src/lib.rs:911` / `614` / `2906`. If all daemons disconnect after a stop is parked, `stop_reply_senders` is never drained, so the CLI `dora stop` hangs.
- **Orphaned running dataflow** — `binaries/coordinator/src/lib.rs:2899-2917`. A fully-running dataflow whose sole daemon dies lingers as "running" with zero daemons; `dataflow_results` / `spawn_result` never resolve.

## Suggested fix

Make this function the authoritative teardown path. After removing a disconnected daemon, for each affected dataflow:
- re-evaluate the ready barrier — broadcast `AllNodesReady` (or terminally fail) when `pending_daemons` just emptied but `daemons` is non-empty;
- drain `stop_reply_senders` with an error when `daemons` becomes empty;
- when a spawned dataflow loses its last daemon, mirror `check_spawn_timeouts` teardown: resolve results with `Err`, persist `Failed { terminal: true }`, archive, and remove.

One integration test per scenario (2-daemon ready-from-one + disconnect-other; stop-then-disconnect-all; running-then-kill-sole-daemon).

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3 + Deeper §6.1). Adversarially re-verified against source._

Issue #1799, opened 2026-05-12
daemon: implement Phase 1.4 graceful degradation — preserve running dataflows across coordinator reconnect


heyong4725:
Tracking issue for **Phase 1.4 (graceful degradation)** of the coordinator-HA plan. Most of the surrounding coordinator-HA work has shipped — this is the remaining piece that prevents seamless dataflow reclaim across coordinator restart.

Originally filed as a carry-over of [`dora-rs/adora#260`](https://github.com/dora-rs/adora/issues/260) (2026-04-15).

## Bug

When the daemon's WebSocket connection to the coordinator drops (or its heartbeat times out at 20s — `binaries/daemon/src/lib.rs:981`), `run_general` returns `Err`, the `Daemon` struct is dropped, and every `RunningDataflow`'s `ProcessHandle` runs `Drop`, which submits `ProcessOperation::Kill` (`binaries/daemon/src/running_dataflow.rs:157-163`). The reconnect loop in `run_with_builds` (`binaries/daemon/src/lib.rs:520-599`) then builds a **fresh** `Daemon` with `running: HashMap::new()`.

Net effect: every coordinator disconnect kills all node processes on every daemon, even though the coordinator can already reconcile state on reconnect.

## What's already shipped on `dora-rs/dora` main

Pulled from [`docs/plan-coordinator-ha.md`](../docs/plan-coordinator-ha.md):

| Plan phase | Status | Evidence |
|---|---|---|
| 1.1 Daemon reconnection loop | done | `binaries/daemon/src/lib.rs:520-599` |
| 1.2 Coordinator accepts re-registration | done | per plan-doc current-state table |
| 1.3 Persist daemon ID across reconnects | done | `DaemonRegisterRequest.previous_daemon_id` |
| **1.4 Graceful degradation** | **OPEN — this issue** | `Daemon`'s `running` HashMap is rebuilt on every reconnect |
| 2.1 Enrich `DataflowRecord` | done | `libraries/coordinator-store/src/redb_store.rs:20` schema v2 (`node_to_daemon`, `uv`, `Recovering`) |
| 2.2 StateCatchUp messaging | done | `libraries/message/src/coordinator_to_daemon.rs:154`, `StateCatchUpAck` |
| 2.3 `Recovering` status + 60s timeout | done | `binaries/coordinator/src/lib.rs:302-313, 1717-1735` |
| 3 ReDB by default | done | CLI default |

[...]

PR #2050, opened 2026-06-05
fix(daemon): correct misleading coordinator-reconnect comment and pin kill-on-disconnect behavior

### LINKED ISSUE #2029: daemon: coordinator-reconnect path kills all running nodes (contradicts #1996 "running dataflows continue")
## Summary

The daemon's coordinator-reconnect path kills every running node on a heartbeat-timeout, despite an inline comment claiming the opposite.

`binaries/daemon/src/lib.rs:1072-1076` does `bail!("coordinator heartbeat timeout (20s)")`, which drops the by-value `Daemon`, dropping each `RunningNode` → `ProcessHandle::drop` submits `Kill` to a still-alive wait task (`binaries/daemon/src/spawn/prepared.rs:650-666`), killing every node. The comment "Running dataflows continue" is false, and this silently defeats the transparent-reconnect design from #1996 / #1998.

## Suggested fix

Either:
- preserve the `ProcessHandle`s across reconnect (stash them so they are not dropped when the reconnect loop rebuilds `Daemon`), making reconnect actually transparent; **or**
- if kill-on-disconnect is the intended behavior, correct the comment and document it.

Test: assert a spawned node's PID survives a simulated >20s coordinator-heartbeat gap.

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3 + Deeper §6.2). Adversarially re-verified against source. Related: #1996, #1998._

phil-opp:
The daemon's heartbeat-timeout comment claimed running dataflows continue on coordinator disconnect, but dropping the by-value `Daemon` drops each `RunningNode` whose `ProcessHandle::drop` submits a `Kill`, so nodes are actually terminated — consistent with the coordinator tearing down a disconnected daemon's dataflows (#2028). This corrects the comment to document the real teardown behavior and adds a regression test pinning that `ProcessHandle::drop` submits a `Kill`. Transparent graceful degradation across reconnect remains future work tracked by #1799.

Closes #2029.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1070,8 +1070,16 @@ impl Daemon {
                             .wrap_err("failed to send watchdog message to dora-coordinator")?;
 
                         if self.last_coordinator_heartbeat.elapsed() > Duration::from_secs(20) {
-                            // Return error to trigger reconnection loop in run().
-                            // Running dataflows continue — nodes are separate processes.
+                            // Returning `Err` propagates up to the reconnect loop in
+                            // `run_with_builds`, which drops this by-value `Daemon` and
+                            // rebuilds a fresh one. Dropping the `Daemon` drops every
+                            // `RunningNode`, whose `ProcessHandle::drop` submits a
+                            // `ProcessOperation::Kill` — so running nodes ARE terminated
+                            // on a coordinator heartbeat timeout, not preserved. This is
+                            // consistent with the coordinator tearing down / releasing a
+                            // disconnected daemon's dataflows (#2028). Transparent graceful
+                            // degradation (preserving running dataflows across a coordinator
+                            // reconnect) is future work tracked by #1799 — not implemented here.
                             bail!("coordinator heartbeat timeout (20s)")
                         }
                     }
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -648,4 +648,28 @@ mod tests {
             "kill from the previous incarnation must not reach the replacement process"
         );
     }
+
+    // ---- #2029: a dropped ProcessHandle kills its node ----
+    //
+    // Pins the documented coordinator-reconnect behavior (lib.rs heartbeat
+    // timeout): when the by-value `Daemon` is dropped, each `RunningNode`'s
+    // `ProcessHandle::drop` submits a `Kill` to the still-alive wait task, so
+    // running nodes ARE terminated on coordinator disconnect. If anyone makes
+    // `Drop` stop killing (e.g. to "preserve nodes"), this test fails, forcing
+    // the lib.rs comment to be revisited rather than silently drifting.
+    #[test]
+    fn dropping_process_handle_submits_kill() {
+        let (tx, rx) = flume::bounded::<ProcessOperation>(2);
+        let handle = ProcessHandle::new(tx);
+        // Dropping the handle stands in for the `Daemon` (and thus the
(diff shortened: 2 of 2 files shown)


heyong4725:
I implemented the other option from the issue (preserve the `ProcessHandle`s across reconnect rather than document the kill) on a parallel branch, and comparing the two surfaced a subtlety worth recording here.

**The added test passes against the "preserve nodes" design too.** `dropping_process_handle_submits_kill` pins `ProcessHandle::drop`, but a node-preserving fix doesn't change `Drop` (it still submits `Kill`) — it just avoids *dropping the handles on reconnect* by borrowing the `Daemon` (`run_inner(&mut self)`) instead of consuming it. So this test pins behavior both designs keep; it doesn't distinguish kill-on-disconnect from preserve-on-disconnect. Worth a comment if the intent is to lock in the kill specifically.

**The real question is the intended semantics, and #2028 makes it concrete.** I verified end-to-end that a node *can* survive a coordinator drop+restore: with the by-value drop removed, on a coordinator restart the record loads as `Recovering` and the `DaemonStatusReport` path reconciles `Recovering -> Running` when the daemon reconnects with the dataflow still running. That's exactly the reclaim machinery #1996/#1998 added, and it matches the design intent laid out in the issue.

But it collides with #2028 (a2ca72b8) in the *other* disconnect path: a daemon disconnect marks the dataflow `Failed { terminal: true }` and archives it, and the reconcile path then refuses to re-adopt archived/terminal dataflows. So a node-preserving daemon that reconnects after the coordinator-side watchdog already fired would keep nodes the coordinator has terminally failed → orphaned/zombie nodes. For transparent reconnect to actually work end-to-end, #2028's transient-disconnect teardown would need to use a reclaimable `Recovering` state (or allow re-adopting still-running archived dataflows) rather than terminal `Failed`.

So I read it as a genuine design fork rather than a bug-vs-correct: **is a coordinator disconnect terminal (this PR + #2028) or a transient blip to ride out (preserve nodes + a #2028 follow-up)?** This PR is correct and internally consistent with #2028 as the tree stands today, and if kill-on-disconnect is the accepted semantics it's the right minimal fix. If transparent reconnect is the goal the issue author described, the preserve-nodes path needs the companion coordinator change and this would be superseded. Either way, since this `Closes #2029`, it'd be good to be explicit about which way we're deciding so #1799 inherits the right framing.

heyong4725:
@phil-opp heads up on direction: we'd like to go with the **survive-reconnect** approach in #2060 rather than documenting the kill here. Riding out a transient coordinator disconnect (within the existing reconnect window) without tearing down running nodes is important for fault tolerance — a coordinator blip shouldn't take down healthy dataflows, and the coordinator's reclaim machinery (#1996/#1998) was already built to re-adopt a reconnecting daemon's running dataflows.

Your PR is accurate as a description of today's behavior, and the `ProcessHandle::drop` test is still valid (it passes against #2060 too, since that approach keeps `Drop` killing and just avoids dropping the handles on reconnect). The one follow-up #2060 needs to make survival work end-to-end is on the #2028 side: a transient daemon disconnect should mark the dataflow reclaimable (`Recovering`) rather than terminal `Failed { terminal: true }` + archived, so the reconnecting daemon's nodes get re-adopted instead of orphaned.

Would you be up for closing this in favor of #2060 (it `Closes #2029`), or do you see a case where kill-on-disconnect is preferable? Happy to pair on the #2028 reclaim adjustment.

phil-opp:
I wasn't sure which way you prefer, that's why I asked for your review. Let's close this then.

Closing in favor of #2060.

PR #2053, opened 2026-06-05
fix(ros2-bridge): emit CDR length prefix for variable-length bool sequences

### LINKED ISSUE #2032: ros2-bridge: sequence<bool> serialized as a CDR tuple with no length prefix (silent data corruption)
## Summary

ROS2 variable-length `bool` sequences are serialized as a CDR *tuple* (no length prefix) instead of a CDR *sequence*, silently corrupting the field and everything after it in the message.

`libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs:298-318` routes the variable-length `bool` path through `serialize_tuple` (at :311), which omits the mandatory u32 CDR length prefix — whereas the correct variable-length path is `serialize_seq` (at :248). The deserializer (`deserialize/sequence.rs:34`) reads a length prefix that the serializer never wrote, so `bool[]` / `sequence<bool>` (occupancy grids, validity masks) and all subsequent fields decode garbage. Round-trips are broken.

## Suggested fix

Use `serialize_seq(Some(len))` for the variable-length `bool` path; keep `serialize_tuple` only for genuinely fixed-size `bool` arrays. Add a CDR-level round-trip test (encode with the bridge, decode with a real CDR reader — not `serde_assert`, which would mask the missing prefix).

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3 + Deeper §6.4). Adversarially re-verified against source._

phil-opp:
The variable-length `bool` sequence path in the Arrow→ROS2 serializer used `serialize_tuple`, which omits the mandatory u32 CDR length prefix that the deserializer reads, so `sequence<bool>` / `bool[]` values and every field after them decoded as garbage. The fix routes that path through `serialize_seq(Some(len))` (matching the deserializer's `deserialize_seq`), leaving genuinely fixed-size bool arrays in `array.rs` on `serialize_tuple`.

Closes #2032.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2199,6 +2199,8 @@ name = "dora-ros2-bridge-arrow"
 version = "1.0.0-rc1"
 dependencies = [
  "arrow",
+ "byteorder",
+ "cdr-encoding",
  "dora-ros2-bridge-msg-gen",
  "eyre",
  "ros2-client",
--- a/libraries/extensions/ros2-bridge/arrow/Cargo.toml
+++ b/libraries/extensions/ros2-bridge/arrow/Cargo.toml
@@ -13,3 +13,9 @@ arrow = { workspace = true }
 eyre = { workspace = true }
 ros2-client = "0.8"
 serde = { workspace = true }
+
+[dev-dependencies]
+# CDR codec used by ros2-client 0.8 on the wire; used in tests to verify the
+# real CDR byte layout (length prefixes, alignment) rather than just serde shape.
+cdr-encoding = "0.10"
+byteorder = "1"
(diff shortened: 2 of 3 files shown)


heyong4725 (approved):
Reviewed — the fix is correct and the test is excellent.

**Fix:** routing the variable-length `bool` sequence through `serialize_seq(Some(len))` instead of `serialize_tuple` is exactly right. CDR encodes `sequence<T>` with a leading u32 length prefix but fixed-size `T[N]` arrays without one, so the variable path belongs on `serialize_seq` (matching the deserializer's `deserialize_seq`) while genuinely fixed-size bool arrays correctly stay on `serialize_tuple` in `array.rs` (untouched here). Good call keeping those distinct.

**Test:** `bool_sequence_round_trips_through_real_cdr` round-trips through the actual `cdr-encoding` codec rather than just checking serde shape, asserts the u32 length prefix is present, and confirms the trailing `int32` field survives. Choosing `flags[0] = true` so the buggy (no-prefix) encoding would misread the leading `0x01` as a length of 1 is a nice touch — it makes the test fail loudly against the old code. Verified locally: `cargo test -p dora-ros2-bridge-arrow` passes.

**CI failure** was the `typos` gate, not the logic: the hyphenated `mis-parses` in the test doc comment splits into a standalone `mis` token that the dictionary flags. I pushed a one-word reword (`misparses`) to this branch to unblock it — comment-only, no behavior change. Repo-wide `typos` is now clean.

LGTM.

PR #2054, opened 2026-06-05
fix(ros2-bridge): deserialize empty sequence<struct> fields

### LINKED ISSUE #2033: ros2-bridge: empty sequence<struct> fails to deserialize and drops the whole ROS2 message
## Summary

An empty `sequence<struct>` field fails to deserialize and drops the entire ROS2 message.

`libraries/extensions/ros2-bridge/arrow/src/deserialize/sequence.rs:160-171` calls `arrow::compute::concat`, which returns an error on empty input. Because the whole message goes through a single top-level `StructDeserializer` (`lib.rs:147-149`), that error fails the entire message deserialize — so any message with an empty struct-sequence field is lost. This hits very common cases: `MarkerArray.markers`, `Path.poses`, `DiagnosticArray.status` are routinely empty.

## Suggested fix

Special-case `values.is_empty()` → build a length-0 array of the correct child type (e.g. `arrow::array::new_empty_array(&child_type)`) instead of calling `concat` on an empty slice. Add an empty-`sequence<struct>` deserialize test.

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3 + Deeper §6.4). Adversarially re-verified against source._

phil-opp:
`arrow::compute::concat` errors on an empty slice, so an empty `sequence<struct>` field (e.g. `MarkerArray.markers`, `Path.poses`, `DiagnosticArray.status`) failed the whole top-level message deserialize and dropped the message. The fix special-cases the empty case to build a length-0 array of the correct child struct type via `arrow::array::new_empty_array`, deriving that type from the message definition so it matches the non-empty path exactly.

Closes #2033.


heyong4725 (approved):
Reviewed — correct fix, and I verified the part that matters most.

The risk in this approach is that `struct_data_type` (the type computed for the empty case) must produce **exactly** the same Arrow `DataType` as the non-empty deserialize path, or empty vs non-empty `sequence<struct>` would yield divergent child schemas. I cross-checked every branch against the actual deserializers and it matches:
- primitives — incl. `char`/`byte`/`u8` → `UInt8` (matches `primitive.rs`'s `U8 | Char | Byte => deserialize_u8`);
- `GenericString` (both `String` and `WString`) → `Utf8` — correct, since `WStringDeserializer` converts to a UTF-8 `StringArray`;
- fixed-size `Array` → `List` — correct, since `ArrayDeserializer` delegates to the same `SequenceVisitor` that yields a `ListArray` (not `FixedSizeList`);
- `Sequence`/`BoundedSequence` → `List`;
- nested `NamedType`/`NamespacedType` recursion mirrors the visitor's package/name resolution;
- field nullability + the list item field name (`Field::new("item", _, true)`, struct fields `Field::new(name, _, true)`) match the non-empty path.

So a zero-row array built from `struct_data_type` is schema-identical to what an equivalent non-empty sequence produces. Verified locally: `cargo test -p dora-ros2-bridge-arrow` passes, and CI is green.

Non-blocking suggestion: `struct_data_type` necessarily duplicates the type mapping that's otherwise implicit in the deserializers, so the two could silently drift if a deserializer's output type ever changes. A cheap guard would be a test that deserializes a non-empty and an empty `sequence<struct>` for a struct with varied field types (a string, a nested struct, an array, a sub-sequence) and asserts the inner `data_type()` is identical — that pins the invariant rather than relying on the comment. Not required to land.

LGTM.

PR #2055, opened 2026-06-05
fix(mavlink2-bridge): exit reader on fatal disconnect instead of logging zombie

### LINKED ISSUE #2034: mavlink2-bridge: reader thread never exits/reconnects on fatal disconnect (logging zombie)
## Summary

On a fatal transport disconnect, the MAVLink bridge's reader thread never exits or reconnects — it becomes a logging zombie.

`binaries/mavlink2-bridge-node/src/main.rs:190-198` — when the underlying TCP connection fails fatally, the reader loop spins forever: it sleeps 50 ms and emits a `warn` (~20/s) with no reconnect and no fatal signal to the rest of the node. The same `Err` arm also conflates *transient* 100 ms read timeouts with *fatal* `Io`/EOF errors, so it can't tell "no data yet" from "connection dead."

## Suggested fix

Distinguish `WouldBlock` / `TimedOut` (transient — keep looping quietly) from fatal errors. On a fatal error, either reconnect via the existing `connect_with_retry` path or exit and surface a fatal error so the node terminates/restarts per policy. Rate-limit the warn so a dead link doesn't flood the logs.

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3). Adversarially re-verified against source._

phil-opp:
On a fatal MAVLink transport disconnect the reader loop spun forever, sleeping 50ms and emitting a warn ~20/s with no exit or reconnect, and it conflated transient 100ms read timeouts with fatal Io/EOF errors. A new `classify_recv_error` helper now treats `WouldBlock`/`TimedOut`/`Interrupted` and parse errors as transient (keep looping, rate-limited warn) and fatal Io errors (e.g. `UnexpectedEof`/`ConnectionReset`) as a hard error that the reader returns, so `main` tears the node down and the daemon restarts it per policy.

Closes #2034.

### DIFF
--- a/binaries/mavlink2-bridge-node/src/main.rs
+++ b/binaries/mavlink2-bridge-node/src/main.rs
@@ -61,7 +61,9 @@ use dora_node_api::{
     flume,
 };
 use eyre::{Context, Result, bail, eyre};
-use mavlink::{MavConnection, MavHeader, MavlinkVersion, Message, common::MavMessage};
+use mavlink::{
+    MavConnection, MavHeader, MavlinkVersion, Message, common::MavMessage, error::MessageReadError,
+};
 use serde::Deserialize;
 use std::{
     sync::{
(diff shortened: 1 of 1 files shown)


heyong4725 (approved):
Reviewed — clean, well-architected fix and excellent test coverage.

The split into `classify_recv_error` → `RecvDisposition::{Transient, Fatal}` is the right shape: `WouldBlock`/`TimedOut`/`Interrupted` and `Parse` errors keep the loop running (a read timeout or one garbled frame doesn't mean the link is dead), while other `Io` errors return a hard error so the node tears down and the daemon restarts it per policy. The catch-all `_ => Fatal` is the correct conservative default — erring toward restart-over-spin is exactly what you want for a dead link.

I verified the end-to-end propagation: when the reader returns `Err`, `main` breaks on `reader_handle.is_finished()`, and the post-loop join hits `Ok(Err(e)) => bail!(...)`, so `main` returns `Err` and the node exits non-zero. No race there — `is_finished()` is only true once the thread has already returned its `Err`, so the join observes it. The shutdown-first check inside the `Err` arm also correctly keeps an intentional shutdown from being reported as a fatal error.

Tests are thorough: every classification case is covered, plus `run_reader_exits_on_fatal_tcp_disconnect` which drives a real TCP peer that closes the socket and asserts the reader returns within 3s (pre-fix it never returns). Verified locally: `cargo test -p dora-mavlink2-bridge-node` is green (12 tests).

**On the failing `ci/circleci: contract-tests` — not caused by this PR.** This change only touches `binaries/mavlink2-bridge-node/src/main.rs`, which no contract test exercises; every compile/test gate (Check, Clippy, Format, test-linux, all 8 cross-checks, msrv, unwrap-budget) is green. The contract-tests job ran ~49 min and timed out (steps cap at 30m + 15m `no_output_timeout`), i.e. a hang — the same flaky signature seen on #2048. Recommend just re-running it.

LGTM.

PR #2060, opened 2026-06-05
fix(daemon): preserve running nodes across coordinator reconnect (#2029)

### LINKED ISSUE #2029: daemon: coordinator-reconnect path kills all running nodes (contradicts #1996 "running dataflows continue")

heyong4725:
## Summary

Fixes #2029 and #1799 (Phase 1.4 graceful degradation). A transient coordinator disconnect — a heartbeat blip, a WS drop, or a coordinator restart — used to kill every running node on every daemon and orphan the dataflow. This PR makes the disconnect **transparent within the reconnect window**: nodes keep running, and on reconnect the dataflow is fully re-adopted (visible + manageable + logging), or, if the coordinator has already given up, the orphans are cleaned up.

Supersedes #2050 (the "document the kill" alternative), which is closed in favor of this PR.

## How it works, end to end

**Daemon — preserve nodes across reconnect.** `run_inner` now borrows `&mut self` instead of consuming the `Daemon`, so a reconnect-triggering error (heartbeat timeout / coordinator-send failure) no longer drops the `ProcessHandle`s. The daemon is built once and only its per-connection fields are refreshed; it adopts the coordinator-assigned id each registration (the coordinator validates it). Permanent loss still exits + kills (orphan avoidance, #1996).

**Coordinator — reclaim instead of terminally fail.** A daemon disconnect now marks the dataflow `Recovering` (not `Failed{terminal}`+archived), so a reconnecting daemon's `DaemonStatusReport` can reconcile it back to `Running`; the existing 60s recovery-timeout still handles genuine permanent loss. (`DisconnectAction::ReclaimOrphaned`.)

**P1 — re-establish the live `running_dataflows` entry on reconnect.** The reconcile previously updated only the persisted store, but `dora list` / `stop` / `logs` / `node` / `param` all read the in-memory map. `reestablish_running_dataflow` now reconstructs the entry from the record + the daemon's report (or relinks it for a multi-daemon partial reconnect), so the survivor is manageable again, not just `Running` on disk.

**P2 — node log forwarding follows the reconnect.** Per-node log tasks captured the coordinator sender + daemon id at spawn, so after a reconnect their logs went to the dead sender / stale (validated) id. `LogDestination::Coordinator` now holds an `Arc<Mutex<CoordinatorLogTarget>>` swapped in place on reconnect; all clones share it.

**P3 — stop orphans that outlive the terminal verdict.** The daemon's reconnect window (~110s) outlives the coordinator's recovery timeout (~90s), so a partition healing in that gap lets a daemon reconnect with live nodes after the coordinator already failed the dataflow terminally. The reconcile preserves that verdict (won't resurrect it), so it now sends the daemon a `StopDataflow` to tear the orphans down rather than leave them unmanageable.

## Test plan

- [x] e2e `tests/daemon-reconnect-e2e.rs`, two scenarios, both asserting the node is the **same pid + alive** AND that **`dora list` shows the dataflow `Running`** after reconnect:
  - `node_survives_coordinator_reconnect` — SIGKILL + restart the coordinator (persistent redb).
  - `node_survives_daemon_watchdog_disconnect` — coordinator stays up; daemon SIGSTOPped past the 30s watchdog, then SIGCONTed.
  - Fails against the pre-fix code (node killed in the gap).
- [x] Unit tests: coordinator reclaim transition, `reestablish_running_dataflow` (reconstruct + relink), P3 orphan-stop (capturing daemon connection); daemon P2 (a spawn-time log clone follows the reconnect).
- [x] `fault-tolerance-e2e` 11/11 (the refactored `run_dataflow` path); daemon + coordinator lib suites green; fmt + clippy + `cargo check --examples` clean.


heyong4725:
Added the companion coordinator-side change so transparent reconnect works end-to-end (the gap flagged in the original description and on #2050).

**2nd commit — `fix(coordinator): reclaim (not terminally fail) dataflows on daemon disconnect`:**
- #2028's daemon-disconnect path persisted `Failed { terminal: true }` + archived the dataflow, and the reconcile path refuses to re-adopt archived/terminal records — so a daemon that reconnects with surviving nodes would be orphaned (zombies). 
- Replaced `tear_down_orphaned_dataflow` with `begin_orphaned_dataflow_reclaim`: persist `Recovering` (not terminal, not archived), release parked `dora stop` waiters (keeps #2028's deadlock-freedom), remove from `running_dataflows`. The existing `DaemonStatusReport` reconcile then promotes `Recovering -> Running` on reconnect, and the existing 60s `Recovering -> Failed` sweep still handles genuine permanent loss.
- Spawn-pending disconnects are untouched — `check_spawn_timeouts` still owns that case and still fails terminally.

**Tests:** updated the disconnect unit test to assert the reclaim transition; added a 2nd e2e `node_survives_daemon_watchdog_disconnect` (coordinator stays up, daemon SIGSTOPped past the 30s watchdog then SIGCONTed) that asserts the coordinator enters the reclaim window, reconciles back to Running, and the node survives as the same pid. Both e2e tests + all 85 coordinator unit tests green; fmt/clippy clean.

heyong4725:
Both findings confirmed and fixed — thanks, these were real gaps.

**P1 — coordinator now re-establishes the live `running_dataflows` entry on reconnect** (`2322cc9d`). Added `reestablish_running_dataflow`, called from the `DaemonStatusReport` reconcile: if the entry is absent (reclaimed away, or a coordinator restart) it reconstructs the `RunningDataflow` from the persisted record + the daemon's report (`RunningDataflow::recovered` — parse+resolve the descriptor, seed `daemons`/`node_to_daemon` from the report, cache `spawn_result` Ok so stop/wait don't hang); if present (multi-daemon partial reconnect) it relinks this daemon's share. So `dora list` / `stop` / `logs` / `node` / `param` see the survivor again, not just the store. Covers both the reclaim and coordinator-restart paths.

You were right that the e2e masked this: **both e2e scenarios now assert `dora list` shows the dataflow `Running` after reconnect** (replacing the log-only check), so the orphan is a gated regression. Added a coordinator unit test (`reconcile_reestablishes_running_dataflow_after_reconnect`) covering reconstruct-if-absent + relink-if-present.

**P2 — node log forwarding now follows the reconnect** (`a6a585cd`). `LogDestination::Coordinator` holds an `Arc<Mutex<CoordinatorLogTarget>>` (sender + daemon id) instead of a captured sender. `try_clone` clones the `Arc`, so every per-node log clone shares one target; on reconnect the daemon swaps the contents in place (`update_coordinator_target`), and `Logger::log` reads the current sender + validated wire id from it. Added a daemon unit test (`coordinator_log_target_update_reaches_spawn_time_clones`) proving a clone captured before the reconnect emits under the new id afterward.

Verification: P1 + P2 unit tests pass; both daemon-reconnect e2e scenarios pass with the new `dora list` assertions; `fault-tolerance-e2e` (the `run_dataflow` path through the refactored logger) 11/11; daemon (38) + coordinator (86) lib suites green; fmt + clippy + `cargo check --examples` clean.

heyong4725:
P3 fixed (`cb…` latest commit). The daemon's reconnect window (~110s) outlives the coordinator's recovery timeout (~90s), so a partition healing in that gap let a daemon reconnect with live nodes after the coordinator had already failed the dataflow terminally — and since the reconcile rightly preserves that terminal verdict, the survivors were left orphaned.

Fix: when a daemon reports a terminally-failed or archived dataflow as running, the coordinator sends it a one-way `StopDataflow` (method `daemon_event`, so the reconcile loop doesn't block on a round-trip and the daemon processes it without replying) to tear the orphans down. Wired into both reconcile skip arms (archived + `Failed{terminal:true}`); unit-tested with a capturing daemon connection.

This closes the last way the change could leave nodes running unmanageable. Verification: new P3 unit test + 87 coordinator lib tests green; both daemon-reconnect e2e scenarios still pass (no regression); fmt/clippy clean. Refreshed the PR description to cover the full daemon-survival → reclaim → P1 → P2 → P3 story.

PR #2063, opened 2026-06-05
test(hardening): cover verified security/framing test gaps (#2027)

heyong4725:
## Summary

Closes the four **verified security/framing test gaps** from the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md` §4 "Test gaps on security/framing"). Each guard was correct but untested, so a regression would ship silently. Test-only — the sole production change is one dev-dependency feature.

## Tests added (19)

- **Coordinator WebSocket auth** (`ws_server.rs`) — `validate_token`: disabled / matching / wrong / missing / empty token; `extract_token`: `Bearer` / missing header / non-`Bearer` scheme / case-sensitivity / empty-after-`Bearer`.
- **Daemon TCP framing** (`socket_stream_utils.rs`) — round-trip, empty payload, oversized-outgoing rejection, **oversized length-prefix rejection** (the DoS guard: rejects from the 8-byte header before allocating the body), and **read-timeout on a stalled peer**. The timeout test uses `#[tokio::test(start_paused = true)]` so the 30s `TCP_READ_TIMEOUT` is exercised instantly — this adds the `tokio` `test-util` dev-feature to the daemon.
- **Node TCP framing** (`daemon_connection/tcp.rs`) — round-trip, empty, oversized length-prefix rejection, truncated header, truncated body (via `Cursor`).
- **Raw-buffer Arrow round-trip** (`arrow_utils.rs`) — non-zero (sliced) `offset`, plus the empty-buffer early-return path for an empty primitive and an empty nested struct.

## Verification

- [x] `dora-coordinator` (93), `dora-daemon` (43), `dora-node-api` (93) lib suites green.
- [x] `cargo fmt --all -- --check` clean; CI-style `cargo clippy … -- -D warnings` clean.

### DIFF
--- a/apis/rust/node/src/daemon_connection/tcp.rs
+++ b/apis/rust/node/src/daemon_connection/tcp.rs
@@ -108,3 +108,63 @@ fn tcp_receive(connection: &mut (impl Read + Unpin)) -> std::io::Result<Vec<u8>>
     connection.read_exact(&mut reply)?;
     Ok(reply)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{tcp_receive, tcp_send};
+    use std::io::Cursor;
+
+    // Length-prefixed framing guards on the node side of the daemon<->node TCP
+    // transport (dora-rs/dora#2027 verified test gap). A malformed/hostile frame
+    // must error, not over-allocate or hang.
+
+    #[test]
+    fn framing_roundtrip() {
+        let payload = b"hello dora framing".to_vec();
+        let mut wire = Vec::new();
+        tcp_send(&mut wire, &payload).expect("send");
+        // 8-byte little-endian length prefix + payload.
+        assert_eq!(wire.len(), 8 + payload.len());
+        let got = tcp_receive(&mut Cursor::new(wire)).expect("receive");
+        assert_eq!(got, payload);
+    }
+
+    #[test]
+    fn empty_payload_roundtrips() {
+        let mut wire = Vec::new();
+        tcp_send(&mut wire, &[]).expect("send empty");
+        let got = tcp_receive(&mut Cursor::new(wire)).expect("receive empty");
+        assert!(got.is_empty());
+    }
+
+    #[test]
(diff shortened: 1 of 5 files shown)


heyong4725:
Self-review (adversarial, with a mutation check).

**These are real regression gates, not vacuous passes.** I confirmed the methodology by neutering the node-side oversized-length DoS guard and re-running: the test fails with `UnexpectedEof` instead of the asserted `InvalidData` (without the guard it allocates the body buffer, then hits EOF). With the guard it passes. The same reasoning holds for the others:
- roundtrip/empty tests fail if framing breaks;
- truncated-header/body tests fail if EOF handling regresses;
- the `start_paused` timeout test would hang (never resolve) if the `tokio::time::timeout` wrapper were removed — it correctly keeps the writer half alive (`_writer`) so the read pends rather than EOFs;
- the offset roundtrip fails if `copy_array_into_sample`/`buffer_into_arrow_array` stop preserving a non-zero `offset` (asserts full `ArrayData` equality).

**Two honest notes (both acceptable, no change needed):**
1. `raw_roundtrip_empty_primitive_uses_empty_buffer_path` *exercises* the `raw_buffer.is_empty()` early-return but isn't a strict mutation-gate for that exact branch — with the early-return removed, `try_new` on the zero-length buffers still yields an equivalent empty array, so the test would still pass. It's valuable coverage of the empty path (which the audit flagged as untested), just not a branch-killer.
2. `send_rejects_oversized_message` would *hang* (full duplex buffer, no reader) rather than cleanly fail if the outgoing-size guard were removed, and it transiently allocates `MAX_MESSAGE_BYTES + 1` (64 MiB). Low concern — a CI timeout is still a failure signal — but worth knowing.

Scope is clean: test-only except the `tokio` `test-util` dev-feature (needed for `start_paused` to exercise the 30s read-timeout instantly). Coordinator 93 / daemon 43 / node-api 93 lib suites green; fmt + CI-style clippy clean.

LGTM.

PR #2065, opened 2026-06-05
fix(ros2-bridge): propagate Name::new errors in generated create fns (#2027)

heyong4725:
## Summary

Quick Win #3 from the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md` §5). Closes 4 node-abort panics in the ROS2 codegen.

The generated service/action client+server `create` functions emitted:
```rust
&crate::ros2_client::Name::new(name_space, base_name).unwrap(),
```
so an invalid user-supplied name/namespace **panics the node**. The graceful path (`message.rs`) already uses `?`.

## Fix

Replace all four `.unwrap()`s (`service.rs:137,354`, `action.rs:402,882`) with:
```rust
&crate::ros2_client::Name::new(name_space, base_name)
    .map_err(|e| eyre::eyre!("invalid ROS2 name/namespace: {e:?}"))?,
```
The enclosing generated fns already return `eyre::Result<Box<…>>` and already wrap the `create_*` call with the identical `.map_err(…)?` — so this is mechanical and matches existing precedent on the same line.

## Tests

Added the **first tests to exercise the service/action codegen output**: each generates the four create fns and asserts (via `rust_format`/prettyplease, already a dependency) that the output (a) **parses as valid Rust** and (b) **routes `Name::new` through `map_err`** (the formatted source contains `invalid ROS2 name/namespace`), so the panic can't silently come back. Verified the gate fails when the fix is reverted to `.unwrap()`.

## Verification

- [x] `cargo test -p dora-ros2-bridge-msg-gen` green (66).
- [x] `cargo fmt --all -- --check` clean; clippy clean. No unrelated churn (diff is the 4 replacements + 2 test modules).

### DIFF
--- a/libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs
+++ b/libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs
@@ -399,7 +399,8 @@ impl Action {
 
                 let client = node.node.create_action_client::< action :: #self_name >(
                     crate::ros2_client::ServiceMapping:: #ros_service_mapping,
-                    &crate::ros2_client::Name::new(name_space, base_name).unwrap(),
+                    &crate::ros2_client::Name::new(name_space, base_name)
+                        .map_err(|e| eyre::eyre!("invalid ROS2 name/namespace: {e:?}"))?,
                     &crate::ros2_client::ActionTypeName::new(#package_name, #self_name_str),
                     qos.into(),
                 ).map_err(|e| eyre::eyre!("{e:?}"))?;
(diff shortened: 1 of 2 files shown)


heyong4725:
Self-review (adversarial + mutation check).

**Scope — complete.** All 4 audit-cited `Name::new(..).unwrap()` sites are fixed; I grepped the whole `msg-gen/types` tree and there are no other `Name::new`/`*TypeName::new(..).unwrap()` of the same class left (the only remaining matches are in my test comments). The adjacent `ServiceTypeName::new`/`ActionTypeName::new(..)` calls aren't `.unwrap()`ed (infallible/used directly), so no sibling gap.

**Correctness.** The replacement is type-valid in the generated context: the enclosing generated fn returns `eyre::Result<Box<…>>` and already wraps the `create_*` call with the identical `.map_err(|e| eyre::eyre!("{e:?}"))?` on the same line, so `eyre::eyre!`/`?`/`{e:?}` all resolve. The codegen tests confirm the output parses as valid Rust (via prettyplease).

**The test is a real gate — mutation-verified.** Reverting either file's `map_err` back to `.unwrap()` makes its codegen test fail (the generated source no longer contains `invalid ROS2 name/namespace`), e.g. the action test panics at `client create fn`. With the fix it passes. These are also the *first* tests to exercise the service/action codegen output at all (audit Deeper #4 noted the codegen was never test-driven).

**Honest limitation:** the tests verify the generated code *parses* as valid Rust + routes the error through `map_err`; they don't fully type-check/compile it (that needs a ROS distro consumer). Parse + the same-line precedent give high confidence; full compile happens in release QA (`scripts/ros2dev.sh`).

Diff is surgical: the 4 replacements + 2 test modules, no churn. fmt + clippy clean; `cargo test -p dora-ros2-bridge-msg-gen` green (66).

LGTM.

PR #2066, opened 2026-06-05
fix: bound coordinator persistence growth + tolerate corrupt redb rows (#2027)

heyong4725:
Closes the two **medium-severity coordinator-persistence findings** from `docs/audit-2026-06-04-soundness.md` (soundness audit, #2027).

## 1. redb `list_*`: skip corrupt rows instead of failing the whole listing

`list_daemons` / `list_dataflows` / `list_builds` pushed `decode(value)?` for every row, so a **single** undeserializable record made the entire call return `Err`. That would disable startup recovery, `dora list`, and `dora clean` for the whole store — one bad row bricks the coordinator's view of persisted state.

Each row that fails to decode is now logged (`tracing::warn!`) and skipped; healthy records are still returned. `tracing` is added as an optional dep gated behind the existing `redb-backend` feature.

## 2. `dataflow_results`: FIFO-bounded at 200

`dataflow_results` was a `HashMap` that only ever shrank via `dora clean`. Synthetic entries (coordinator reconnect, spawn-timeout watchdog) and results for dataflows that never reach archival accumulated without bound. It's now an insertion-ordered `IndexMap` capped at `MAX_DATAFLOW_RESULTS` (200) at every insert site, mirroring the existing `archived_dataflows` bound. Eviction is FIFO via `shift_remove_index(0)`; the `dora clean` removal switches to `shift_remove` to preserve order.

## Tests
- `list_dataflows_skips_corrupt_row` — injects a garbage row directly into the `DATAFLOWS` table, asserts the healthy record is still listed (gates the fix: old `?` code panics here).
- `watchdog_synthesis_keeps_dataflow_results_bounded` — pre-fills the map to the cap, drives the watchdog synthesis path, asserts the oldest entry is evicted FIFO and the new one survives.

## QA
- `cargo clippy -p dora-coordinator -p dora-coordinator-store --features dora-coordinator-store/redb-backend -- -D warnings` clean
- `cargo test -p dora-coordinator -p dora-coordinator-store --features redb-backend` — all green (coordinator 94, store 33)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1782,6 +1782,7 @@ dependencies = [
  "redb",
  "serde",
  "tempfile",
+ "tracing",
  "uuid",
 ]
 
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -513,9 +513,6 @@ async fn start_inner(
                                 .entry(uuid)
                                 .or_default()
                                 .insert(daemon_id, result);
-                            while dataflow_results.len() > MAX_DATAFLOW_RESULTS {
-                                dataflow_results.shift_remove_index(0);
-                            }
 
                             if dataflow.daemons.is_empty() {
                                 // Archive finished dataflow (cap at 200 to prevent unbounded growth)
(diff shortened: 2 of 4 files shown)


heyong4725:
Good catch on P2 — confirmed and fixed in f468713b.

You're right that `dataflow_results` is live state, not just history: a multi-daemon dataflow accumulates one entry per daemon while still running, and the final status (lib.rs ~564-586) reads the full set when the last daemon finishes. The blind `shift_remove_index(0)` cap could evict an earlier daemon's failed result while another daemon was still running, so the surviving entry alone would classify the dataflow as `Succeeded` — exactly the corruption the `dora clean` comments warn about.

**Fix:** replaced the three inline caps with a single `cap_dataflow_results` helper that only evicts the oldest entry whose dataflow is **not** in `running_dataflows`. Active (still-running) entries are never evicted; if every over-cap entry is still active, the map is left above the cap rather than corrupting live state — the bound targets finished-history growth, not concurrency.

Placement detail: the `DataflowFinishedOnDaemon` match scrutinee is `running_dataflows.entry(uuid)`, which borrows `running_dataflows` for the whole match, so the helper can't run inside either arm. It now runs once right after the match (borrow released) and in the spawn-timeout watchdog (where `running_dataflows` is a param).

New regression test `cap_dataflow_results_preserves_running_dataflow_entries` inserts an active dataflow's entry as the FIFO-oldest, fills past the cap, and asserts the active entry survives while a finished-history entry is evicted instead.

All green: `cargo test -p dora-coordinator -p dora-coordinator-store --features redb-backend` (coordinator 95, store 33), clippy `-D warnings`, fmt.

Issue #2081, opened 2026-06-07
arrow-convert: `into_vec` silently ignores the null bitmap, turning null slots into backing values (e.g. `0`)


phil-opp:
> **Automated report.** This issue was created by a scheduled automated Claude code-review check. A human should confirm before acting on it.

## Summary

`dora_arrow_convert::into_vec<T>()` iterates over `buffer.values()` and casts every element, without consulting the array's null/validity bitmap. As a result, a null slot is silently emitted as its (arbitrary) backing value — typically `0` — producing data that is silently wrong rather than an error.

This is inconsistent with the rest of the same crate: every `TryFrom<&ArrowData>` impl in `libraries/arrow-convert/src/from_impls.rs` explicitly rejects arrays with nulls (`if array.null_count() != 0 { eyre::bail!("array has nulls"); }`). `into_vec` is the one converter that doesn't, so the inconsistency strongly suggests this is an oversight rather than intended behavior.

`into_vec` is reachable from the public node/operator API (`pub use dora_arrow_convert::*;`), so a downstream node that receives a column containing nulls will get fabricated zeros instead of either an error or `None`.

## Where

`libraries/arrow-convert/src/lib.rs`, in the `register_array_handlers!` macro body (around lines 59-64):

```rust
let mut result = Vec::with_capacity(buffer.len());
for &v in buffer.values() {            // <-- iterates raw values, ignores validity
    let converted = NumCast::from(v)
        .context(format!("Failed to cast value from {} to target type", $type_name))?;
    result.push(converted);
}
Ok(result)
```

## Reproduction (verified)

```rust
use dora_arrow_convert::{into_vec, ArrowData};
use std::sync::Arc;
use arrow::array::Int32Array;

// [10, null, 30]
let data = ArrowData(Arc::new(Int32Array::from(vec![Some(10), None, Some(30)])));
let v: Vec<i32> = into_vec(&data).unwrap();
assert_eq!(v, vec![10, 0, 30]); // passes — the null became a real 0
```

Observed output:

```
into_vec over array with null = [10, 0, 30]
```

The missing `null` is indistinguishable from a genuine `0` value.

## Suggested fix

Pick one of these, ideally matching the crate's existing convention:

1. **Reject nulls** (consistent with the `TryFrom` impls): `if buffer.null_count() != 0 { eyre::bail!("array has nulls"); }` before the loop.
2. Document that `into_vec` requires a null-free array and is undefined otherwise.

[...]

PR #2085, opened 2026-06-08
fix(coordinator): two-phase restart to prevent cross-daemon Zenoh data-plane race after dataflow restart

LeonRust:
## Problem
Nightly CI (issue #2082) exhibits a flaky cluster-e2e failure: a restarted
dataflow gets stuck in Running and never reaches Finished. The root cause
is a race condition in the coordinator's restart flow:
1. The coordinator sends StopDataflow to each daemon and immediately calls
start_dataflow upon receiving StopResult(Ok).
2. Daemons reply with StopResult(Ok) before their node processes have
actually exited, which means old nodes' Zenoh subscribers and declarations
are still active.
3. New nodes spawn while old nodes' Zenoh subscribers are still alive,
causing the cross-daemon Zenoh data plane to drop or misroute messages.
This race was widened by the post-Stop drain added in commit b95823d5.

### Fix
Replace the single-phase restart_dataflow with a two-phase, event-driven
restart that defers starting the replacement dataflow until all daemons
have confirmed node exit:
- Phase 1 — initiate_restart: sends StopDataflow to all daemons,
persists the dataflow as Stopping, and registers a PendingRestart
entry (storing the descriptor, name, uv, and the caller's oneshot reply
sender).
- Phase 2 — deferred completion in DataflowFinishedOnDaemon handler:
when the last daemon reports finish (meaning all nodes have exited and
Zenoh subscribers are torn down), the handler removes the PendingRestart,
calls start_dataflow, persists the new dataflow, and sends the reply.
This guarantees the old dataflow is fully stopped before new nodes spawn.

### Safeguards included
- Double-restart guard: rejecting duplicate Restart requests for a
dataflow that is already being restarted (prevents reply_sender leak).
- Deletion of the old restart_dataflow function: replaced by the
two-phase pattern; git history preserves the original.
- TODO for stale PendingRestart cleanup: daemon crash before
DataflowFinishedOnDaemon is a pre-existing issue not fixed here.

### dora-rs/dora #2082 macOS build timeout mitigation
The macos-latest CLI Tests job also fails sporadically with a Python
operator build timeout. Mitigated by:
- Pre-building dora-operator-api outside of maturin to warm the Cargo
cache and avoid build-server deadlocks (where concurrent pip install /
maturin invocations fight over the workspace lock).
- Increasing the step timeout from 20 → 30 minutes.

### Files Changed

|File | Changes|
|---|---|
|binaries/coordinator/src/state.rs |	+11 lines — PendingRestart struct|
|binaries/coordinator/src/lib.rs |	+128 / -60 lines — two-phase restart, double-restart guard, dead code removal, clippy fix|
|.github/workflows/nightly.yml |	+6 / -1 lines — macOS pre-build + timeout|

### Testing
- cargo test -p dora-coordinator -p dora-daemon -p dora-node-api -p dora-operator-api -p dora-message -p dora-core — 572+ tests passed, 0 failed
- cargo clippy -p dora-coordinator -- -D warnings — clean
- cargo fmt --all -- --check — clean

### Verification needed
- Cluster SSH e2e — requires Linux; should be validated on CI before merge
- macOS CLI Tests — verify the pre-build step resolves the timeout on the next nightly run
Refs #2082 (or references #2082 — the cluster-e2e and macOS timeout failures)


heyong4725:
## Review — two-phase restart (#2082)

**Verdict: Request changes.** The core restart fix is sound in design and compiles/lints clean (build ✅, `clippy -D warnings` ✅, tests ✅), but the PR has (1) significant undisclosed scope drift, (2) zero tests for the actual behavior change it ships, and (3) a new hang regression mitigated only by a TODO.

### Scope check
Description lists **3** changed files; the diff has **5**. Two are unrelated to #2082:
- `libraries/arrow-convert/src/lib.rs` (+ test) — commit `09012b1c`, a public-API behavior change.
- `lib.rs:4542` `sort_by_key`/`Reverse` clippy tidy.

---

### 🔴 C1 — Unrelated behavior change bundled in (arrow-convert null rejection)
`09012b1c` makes `dora_arrow_convert::into_vec` **reject arrays with nulls** (previously read from the raw buffer as 0/garbage). Genuine improvement, but it's a **public-API behavior change in a published crate** (`dora-arrow-convert` rc1) — semver/QA-policy relevant — buried in a coordinator-restart PR and absent from the description's file table. Violates *"Don't fix unrelated warnings in PRs."* Please split into its own PR (with the behavior-change-class treatment it warrants). Same for the `sort_by_key` tidy.

### 🔴 C2 — The actual fix ships with no test coverage
The two-phase restart — the point of this PR — has **zero** unit/integration tests; the only tests added are for the unrelated arrow-convert change. Per CLAUDE.md, *"New coordinator/daemon behavior → integration test in `binaries/coordinator/tests/`."* Currently it relies solely on Linux-only `cluster-e2e`. At minimum: assert the `Restart` reply is **not** sent until a `DataflowFinishedOnDaemon` arrives, and that a second concurrent `Restart` is rejected.

### 🟠 H1 — New indefinite-hang regression on daemon loss
The caller's `reply_sender` is now held in `pending_restarts` until the last `DataflowFinishedOnDaemon`. If a daemon crashes/disconnects after `StopDataflow` but before reporting finish, the entry never drains and **the caller hangs forever**. The old synchronous `restart_dataflow` could not hang this way — this is a regression, acknowledged only by the TODO at `lib.rs:4665`. The cleanup hook already exists: `cleanup_disconnected_daemons_from_running_dataflows` (`lib.rs:3019`) runs on disconnect — draining matching `pending_restarts` there (sending `Err` to the caller) is a small in-pattern fix that closes the regression rather than deferring it.

### 🟡 M1 — Double-restart guard runs *after* `stop_dataflow`
In `initiate_restart`, the `pending_restarts.contains_key` guard (`lib.rs:4661`) is checked **after** the dataflow is already re-stopped, so a duplicate `Restart` re-sends `StopDataflow` to every daemon before being rejected. Move the guard to the top of the function (before extracting the descriptor / stopping) so duplicates are rejected cheaply with no side effects.

### 🟢 L1 — nightly.yml prebuild swallows diagnostics
`cargo build -p dora-operator-api 2>/dev/null || true` is fine as best-effort warming, but `2>/dev/null` hides the error if the prebuild itself is what's timing out. Consider dropping `2>/dev/null` (keep `|| true`).

---

### What's good
- **Design is correct.** Confirmed `stop_dataflow` (`handlers.rs:213`) does *not* remove from `running_dataflows`, so deferred completion firing on `dataflow.daemons.is_empty()` genuinely means all nodes across all daemons exited and Zenoh state is torn down — exactly the invariant the old single-phase flow lacked.
- Deferred block handles the `start_dataflow` failure path correctly (old df already archived/removed, `Err` returned).
- `entry.remove()` → restart → re-`insert` ordering is borrow-safe; the new df re-enters the normal `ReadyOnDaemon` flow cleanly.

### Suggested path to merge
1. Drop `09012b1c` + the `sort_by_key` tidy → separate PR(s). (C1)
2. Add a coordinator integration test for the deferred-reply + double-restart-reject contract. (C2)
3. Drain `pending_restarts` in the daemon-disconnect cleanup path. (H1)
4. Hoist the double-restart guard above `stop_dataflow`. (M1)

heyong4725:
## Follow-up: #2082 is a bucket of 4 failures — this PR closes 2 of them

I dug into the nightly runs behind #2082. The auto-filed excerpts are truncated to the *last* 10 KB (just git/process teardown), so the real errors are buried mid-log. Pulling them out, #2082 is **four distinct failures** across two nights:

| Night | Job | Platform | Real error | This PR? |
|---|---|---|---|---|
| 06-07 | `cluster-e2e` | Linux | `dataflow … did not reach Finished within 30s (status: Running)` | ✅ two-phase restart |
| 06-07 | `cli-tests` | macOS | `'Test Python Operator example' has timed out after 20 minutes` | ✅ prebuild + 30-min bump |
| 06-08 | `cli-tests` | macOS | `object_detection exited before connecting to dora` → cascading SIGTERM | ❌ |
| 06-08 | `test-cross-platform` | Windows | 3 `check_build_timeouts_*` tests panic at `std\time.rs:446` | ❌ |

### So `Closes #2082` is too broad
This PR legitimately fixes the cluster-e2e Zenoh race and the night-1 macOS build timeout, but two failures bucketed under #2082 are untouched. Please change `Closes #2082` → `Refs #2082` so the issue stays open for the remaining two.

### The Windows panic is a real, separate bug — fixed in #2088
`check_build_timeouts_*` backdate `build_started_at` via `Instant::now() - build_result_timeout()`, and `build_result_timeout()` defaults to **20 min**. On Windows `Instant` is QPC-since-boot, so `now - 20min` underflows and panics (`std\time.rs:446`) on any runner with uptime < ~20 min — intermittent by nature. The spawn-watchdog tests use the same idiom but backdate by only 61s, which is why *only* the build tests fail. I opened **#2088** (dependency-inject the timeout; tests backdate ~2s instead of 20 min) to fix it independently of the restart work here.

### The night-2 macOS `cli-tests` flake is also separate
Night 1 timed out in the *build* step (your prebuild + 30-min bump targets that). Night 2 *built fine* but failed at *run* time: `object_detection` (YOLO/Ultralytics first-run model download + matplotlib font-cache) didn't connect before the `--stop-after 20s` window, so the daemon SIGTERM'd the graph. That's a slow-cold-start flake in `python-operator-dataflow`, not a build-server lock — the prebuild mitigation here won't touch it. Worth a separate tracking item.

heyong4725:
Heads up: the arrow-convert null-rejection change that's bundled here (commit `09012b1c`) has now landed independently via #2091 (merged to `main`, closes #2081) — thanks for splitting it out. On your next rebase onto `main`, please drop `09012b1c` (and the unrelated `sort_by_key`/`Reverse` clippy tidy) so this PR is scoped to just the two-phase restart fix. After rebase that commit should be an empty/duplicate change anyway.

That leaves the substantive review items still open here: change `Closes #2082` → `Refs #2082` (this PR covers the cluster-e2e race + macOS build timeout, not the Windows panic or the macOS cold-start flake, both since fixed in #2088/#2090), and add coordinator test coverage for the deferred-reply + double-restart-reject contract. See my earlier review for the full list.

LeonRust:
Closes has been changed to Refs。I'll take a closer look at #2088 and #2090.

heyong4725:
Thanks for the thorough turnaround, @LeonRust — every review item is addressed and verified:

- Scope trimmed to the restart fix (arrow-convert split out to #2091, `sort_by_key` tidy reverted via `#[allow]`).
- Double-restart guard moved ahead of `stop_dataflow` so duplicates are rejected with no side effects.
- Hang regression fixed: `pending_restarts` is drained in the daemon-disconnect cleanup path, releasing the caller with a clear error instead of hanging forever.
- Two focused integration tests (`cleanup_disconnected_daemons_drains_pending_restarts`, `initiate_restart_rejects_duplicate_request`) — the duplicate test cleverly uses an empty `DaemonConnections`, so the "already being restarted" assertion also proves the guard runs before `stop_dataflow`.
- `Closes` → `Refs #2082`; nightly prebuild no longer swallows stderr.

Verified locally: `cargo test -p dora-coordinator --lib` 97 passed, clippy `-D warnings` clean, fmt clean. The unit tests cover the restart contract; the end-to-end Zenoh race itself is confirmed by the Linux `cluster-e2e` nightly. Merging.

Issue #2089, opened 2026-06-08
Nightly macOS CLI Tests: python-operator-dataflow flakes — object_detection cold init exceeds --stop-after window


heyong4725:
### Summary

The nightly **CLI Tests (macOS)** job intermittently fails on the `python-operator-dataflow` example: `object_detection` is killed before it finishes initializing, producing `node exited with error: Signal(15)` → `object_detection exited before connecting to dora` → cascading SIGTERM of `plot`. Split out from #2082 (which buckets four unrelated failures).

### Evidence

From the 2026-06-08 nightly run (https://github.com/dora-rs/dora/actions/runs/27123358401), `cli-tests` job:

```
09:06:00 stdout  object_detection:  Creating new Ultralytics Settings v0.0.6 file ✅
09:06:06 stdout  object_detection:  Matplotlib is building the font cache; this may take a minute
09:06:12 INFO dora_daemon: stop-after timeout reached (20s) -> stopping dataflow
09:06:22 ERROR dora_daemon::spawn::prepared: node exited with error: Signal(15) node_id=object_detection
[ERROR] Dataflow failed:
Node `plot` failed: exited because of signal SIGTERM. This error occurred because node `object_detection` exited before connecting to dora.
```

### Root cause

The step runs:

```
dora run examples/python-operator-dataflow/dataflow.yml --uv --stop-after 20s
```

`object_detection.py` imports `torch` + `ultralytics` and constructs `YOLO("yolov8n.pt")` at module load. On a cold runner this triggers two one-time setup costs **before the operator connects to dora**:

- `Creating new Ultralytics Settings` file
- `Matplotlib is building the font cache; this may take a minute`

When that cold init exceeds the **20 s** `--stop-after` window, the daemon hits stop-after and SIGTERMs the graph mid-initialization, so `object_detection` never connects. It's timing-dependent (cache warmth), hence flaky rather than a hard failure. `yolov8n.pt` is already pre-fetched on `main` (`.github/workflows/nightly.yml`), so the model download is **not** the cause — it's the import + first-run setup time.

This is **not** the same as:
- #2088 — Windows `check_build_timeouts_*` unit-test panic (fixed).
- The #2085 macOS *build-step* timeout mitigation — that targets `dora build`, not the timed `dora run`.

### Possible fixes (pick one)

1. **Pre-warm in the setup step** (preferred): before the timed run, pay the one-time costs once so the timed run starts warm, e.g.
   ```bash
[...]

PR #2090, opened 2026-06-08
ci(nightly): pre-warm python-operator example to fix macOS cold-start flake (#2089)

heyong4725:
## Problem

Fixes the macOS cold-start flake tracked in #2089 (split out from #2082).

The nightly **CLI Tests (macOS)** job intermittently fails on `python-operator-dataflow`. From the 2026-06-08 run (https://github.com/dora-rs/dora/actions/runs/27123358401):

```
object_detection:  Creating new Ultralytics Settings v0.0.6 file ✅
object_detection:  Matplotlib is building the font cache; this may take a minute
dora_daemon: stop-after timeout reached (20s) -> stopping dataflow
dora_daemon::spawn::prepared: node exited with error: Signal(15) node_id=object_detection
Node `plot` failed: ... node `object_detection` exited before connecting to dora.
```

`object_detection.py` imports `torch` + `ultralytics` and constructs `YOLO("yolov8n.pt")` at module load. On a cold runner this triggers two one-time setup costs **before the operator connects to dora**: creating the Ultralytics settings file and building the matplotlib font cache. When that cold init exceeds the **20 s** `--stop-after` window, the daemon SIGTERMs the graph mid-init, so `object_detection` never connects and `plot` dies with a cascading SIGTERM. (`yolov8n.pt` is already pre-fetched on `main`, so the model download is not the cause.)

## Fix

Pre-warm `object_detection`'s built venv before the timed run:

```bash
if [ "$RUNNER_OS" != "Windows" ]; then
  examples/python-operator-dataflow/.dora/python-envs/object_detection/bin/python -c \
    "import matplotlib.pyplot; from ultralytics import YOLO; YOLO('examples/python-operator-dataflow/yolov8n.pt')" || true
fi
```

Both the matplotlib font cache and the Ultralytics settings live under `$HOME`, so paying those one-time costs once ahead of `dora run` leaves the timed run warm. Properties:

- **No extra install** — reuses the venv `dora build --uv` already created (`.dora/python-envs/object_detection`, path per `libraries/core/src/build/mod.rs:270`).
- **Best-effort** — `|| true`, never fails the job.
- **Non-Windows only** — Windows uses `Scripts/` not `bin/`, and the flake isn't observed there.

This is option 1 (the preferred fix) from #2089.

## Testing

CI-only change for a cold-runner flake; can't be reproduced locally once matplotlib/ultralytics caches are warm. YAML validated (`yaml.safe_load`). Real confirmation is a green nightly **CLI Tests (macOS)** run.

`Refs #2089` rather than `Closes` — leave the issue open until a nightly confirms the flake is gone, since the fix can't be verified locally.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1351,6 +1351,17 @@ jobs:
           curl -fL --retry 10 --retry-delay 10 --retry-all-errors \
             -o examples/python-operator-dataflow/yolov8n.pt \
             https://github.com/ultralytics/assets/releases/download/v8.4.0/yolov8n.pt
+          # Pre-warm object_detection's env so the one-time matplotlib font-cache
+          # build and Ultralytics settings creation happen BEFORE the timed run.
+          # On a cold runner those can exceed the 20s --stop-after window, so the
+          # node gets SIGTERM'd mid-init and never connects to dora (#2089). Both
+          # caches live under $HOME, so warming the node's built venv (reusing the
+          # deps `dora build` already installed) warms the subsequent run. Best
+          # effort: non-Windows only (Scripts/ vs bin/), never fail the job.
+          if [ "$RUNNER_OS" != "Windows" ]; then
+            examples/python-operator-dataflow/.dora/python-envs/object_detection/bin/python -c \
+              "import matplotlib.pyplot; from ultralytics import YOLO; YOLO('examples/python-operator-dataflow/yolov8n.pt')" || true
+          fi
           dora run examples/python-operator-dataflow/dataflow.yml --uv --stop-after 20s
 
       - name: Test Python Multiple Arrays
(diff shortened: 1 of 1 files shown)


heyong4725:
Reviewed the nightly workflow change. I did not find any blocking issues. The prewarm runs after dora build creates the managed object_detection env, the interpreter path matches libraries/core/src/build/mod.rs for non-Windows managed envs, and the step remains best-effort while logging path drift or warm-up failure.\n\nValidation run during review:\n- python3 -c "import yaml, pathlib; yaml.safe_load(pathlib.Path('.github/workflows/nightly.yml').read_text()); print('yaml ok')"\n- bash -n on the modified Test Python Operator example shell block\n- git diff --check origin/main...HEAD\n\nNote: GitHub would not let me submit this as an approval because the authenticated account owns the PR.

Issue #2098, opened 2026-06-10
Bug: `check_dataflow` skips source-path validation for local nodes whenever `remote_daemon_id` is `Some` (missing `else`)


phil-opp:
## Summary

In `check_dataflow` (`libraries/core/src/descriptor/validate.rs`), the branch that handles a `Custom`/`Local` node whose `source` is a plain path has a control-flow gap. When `remote_daemon_id` is `Some(...)`, the code enters the `else if let Some(remote_daemon_id) = remote_daemon_id { ... }` arm, but the inner `if` only decides whether to **log** a "skipping path check" message — **neither branch performs (nor skips, with intent) the actual path validation**. As a result, for a node that is *not* remote (its machine is not in `remote_daemon_id` and the coordinator is not remote), the source-path existence check and the `build.is_some()` skip are **both silently bypassed**.

## Affected code

`libraries/core/src/descriptor/validate.rs`, lines 74-96:

```rust
dora_message::descriptor::NodeSource::Local => match custom.path.as_str() {
    SHELL_SOURCE => (),
    DYNAMIC_SOURCE => (),
    source => {
        if source_is_url(source) {
            info!("{source} is a URL."); // TODO: Implement url check.
        } else if let Some(remote_daemon_id) = remote_daemon_id {
            if let Some(deploy) = &node.deploy
                && let Some(machine) = &deploy.machine
                && (remote_daemon_id.contains(&machine.as_str())
                    || coordinator_is_remote)
            {
                info!("skipping path check for remote node `{}`", node.id);
            }
            // <-- BUG: no `else` here. A *local* node (machine not in
            //     remote_daemon_id, coordinator not remote, or no deploy/machine)
            //     falls through with NO path validation and NO build-command check.
        } else if custom.build.is_some() {
            info!("skipping path check for node with build command");
        } else {
            resolve_path(source, working_dir).wrap_err_with(|| {
                format!("Could not find source path `{source}`")
            })?;
        };
    }
},
```

## Why this is a bug

The inner `if`'s only effect is to emit a log line; both its taken and not-taken paths perform no validation. That is the tell-tale sign of a missing `else`: the author appears to have intended "skip the check for *remote* nodes, but still validate *local* nodes," yet the local-node validation (`resolve_path`) and the `build.is_some()` skip live in the *subsequent* `else if` / `else` arms, which are unreachable once `remote_daemon_id` is `Some`.

Consequences when `remote_daemon_id` is `Some`:
[...]

PR #2099, opened 2026-06-10
Fix/issue 2098 check dataflow missing else

LeonRust:
## Summary

Refs #2098

Fix a control-flow gap in `check_dataflow` (`libraries/core/src/descriptor/validate.rs`): when `remote_daemon_id` is `Some(...)`, the inner `if` block only handled remote nodes (skipping path checks), but lacked an `else` clause for local nodes. This caused local-node path validation (`resolve_path`) and build-command checks to be silently bypassed.

Currently latent — all call sites pass `None` for `remote_daemon_id`. The bug will manifest as soon as a caller populates this parameter, at which point local nodes in distributed deployments would lose all path validation.

## Root Cause

```rust
// libraries/core/src/descriptor/validate.rs:80-87 (before fix)
} else if let Some(remote_daemon_id) = remote_daemon_id {
    if let Some(deploy) = &node.deploy
        && let Some(machine) = &deploy.machine
        && (remote_daemon_id.contains(&machine.as_str())
            || coordinator_is_remote)
    {
        info!("skipping path check for remote node `{}`", node.id);
    }
    // ❌ no else — local nodes fall through silently
} else if custom.build.is_some() {
    // unreachable when remote_daemon_id is Some
} else {
    // unreachable when remote_daemon_id is Some
};
```

## Fix

Extract `is_remote` to a boolean and add the missing branches inside the `remote_daemon_id` block:

```rust
} else if let Some(remote_daemon_id) = remote_daemon_id {
    let is_remote = node
        .deploy
        .as_ref()
        .and_then(|d| d.machine.as_ref())
        .is_some_and(|m| remote_daemon_id.contains(&m.as_str()))
        || coordinator_is_remote;
    if is_remote {
        info!("skipping path check for remote node `{}`", node.id);
    } else if custom.build.is_some() {
        info!("skipping path check for node with build command");
    } else {
        resolve_path(source, working_dir)
            .wrap_err_with(|| format!("Could not find source path `{source}`"))?;
    }
}
```

## Tests

3 regression tests using `serde_yaml::from_str` to construct `Descriptor` instances:

| Test | remote_daemon_id | Node config | Expected |
|------|------------------|-------------|----------|
| `test_local_node_validated_with_remote_daemon_id` | `Some(&["remote"])` | `path: nonexistent_binary_2098_regression_test` | `Err` |
| `test_remote_node_skipped_with_remote_daemon_id` | `Some(&["remote"])` | `_unstable_deploy: { machine: remote }` | `Ok` |
| `test_local_node_with_build_skipped_with_remote_daemon_id` | `Some(&["remote"])` | `build: cargo build --release` | `Ok` |

## Verification

```
cargo test -p dora-core -- test_local_node test_remote_node test_local_node_with_build
  → 3/3 passed (T1 confirms the fix: before fix → Ok, after fix → Err)

cargo test -p dora-core
  → 168/168 passed, zero regressions

cargo fmt --all -- --check
  → clean

cargo clippy -p dora-core -- -D warnings
  → zero warnings


phil-opp:
Thanks for the PR! It turns out that we don't really use the `remote_daemon_id` anymore, so I opened #2100 to remove the dead code instead.

PR #2110, opened 2026-06-10
perf(node-api): avoid per-call allocations in event scheduler hot path

phil-opp:
## Issue

`Scheduler::next()` runs once per yielded event — the hottest path of the node event stream — and performed two avoidable allocations on every call:

1. `self.last_used.clone()` copied the entire `VecDeque<DataId>` (one `String` allocation per input ID) just to iterate it, even though the loop returns immediately after the first mutation.
2. `DataId::from(NON_INPUT_EVENT.to_string())` allocated a fresh `String` for the non-input-event sentinel on every `next()` call, and again on every `add_event()` call for non-input events.

## Fix

- Iterate `last_used` by index instead of cloning it. The loop `return`s immediately after `remove(index)` + `push_back(id)`, so no iteration ever continues over the mutated deque, and the removed element is moved (not cloned) to the back.
- Introduce a `LazyLock<DataId>` for the `NON_INPUT_EVENT` sentinel and use it in `add_event`, `next`, and `with_policies`, removing the per-call `String` allocations.

No behavior change; scheduling order is identical. Found during a codebase audit (Optimization category).

## Validation

- `cargo test -p dora-node-api` — 96 unit + 5 integration tests pass (includes the scheduler fairness/eviction/flush test suite)
- `cargo clippy -p dora-node-api -- -D warnings`, `cargo fmt --check`

https://claude.ai/code/session_01L1wL2zpWyF7xvURnsUBSNw

---
_Generated by [Claude Code](https://claude.ai/code/session_01L1wL2zpWyF7xvURnsUBSNw)_

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -1,4 +1,7 @@
-use std::collections::{HashMap, VecDeque};
+use std::{
+    collections::{HashMap, VecDeque},
+    sync::LazyLock,
+};
 
 use dora_message::{
     config::{DEFAULT_QUEUE_SIZE, QueuePolicy},
@@ -84,6 +87,11 @@ fn log_correlation_drop(event_id: &DataId, dropped: &EventItem) {
     );
 }
 pub(crate) const NON_INPUT_EVENT: &str = "dora.non_input_event";
+
+/// Shared [`DataId`] for [`NON_INPUT_EVENT`], so the hot `add_event`/`next`
+/// paths don't have to allocate a fresh `String` on every call.
+static NON_INPUT_EVENT_ID: LazyLock<DataId> =
+    LazyLock::new(|| DataId::from(NON_INPUT_EVENT.to_string()));
 
 /// This scheduler will make sure that there is fairness between inputs.
 ///
@@ -176,7 +184,7 @@ impl Scheduler {
         let topic = VecDeque::from_iter(
             event_queues
                 .keys()
-                .filter(|t| **t != DataId::from(NON_INPUT_EVENT.to_string()))
+                .filter(|t| **t != *NON_INPUT_EVENT_ID)
                 .cloned(),
         );
         Self {
@@ -211,7 +219,7 @@ impl Scheduler {
                 ) == Some(true);
                 (id, flush)
             }
-            _ => (&DataId::from(NON_INPUT_EVENT.to_string()), false),
+            _ => (&*NON_INPUT_EVENT_ID, false),
         };
 
         // Flush older queued messages when flush=true is present.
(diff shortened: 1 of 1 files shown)


phil-opp:
## Codebase audit summary

An in-depth review of the repository was performed (daemon/coordinator, node/operator APIs, core libraries, CLI/runtime), scanning for correctness issues, optimization low-hanging fruit, documentation gaps, and simplification opportunities. Each confirmed finding was fixed in its own PR:

### Correctness
- #2112 — `fix(cli)`: two avoidable panic paths — `unwrap()` on non-UTF-8 workspace paths in the C/C++ `dora new` templates, and `unwrap()` on control-request serialization in `send_control_request` — now propagate proper errors.

### Optimization
- #2110 — `perf(node-api)`: `Scheduler::next()` cloned the whole `last_used` `VecDeque<DataId>` and allocated a fresh sentinel `DataId` `String` on every event yield; both per-call allocations removed from the hot path.
- #2111 — `perf(daemon)`: three broadcast loops (`Timer`, `Logs`, `LogBroadcast`) allocated `Arc::new(metadata.clone())` / cloned the full message payload **per subscriber**; the `Arc`s are now created once and shared, matching the pattern already used in the output-dispatch loop.

### Documentation
- #2114 — `docs(arrow-convert)`: the public `IntoArrow` trait now documents which implementations exist, the `()` → `NullArray` convention, the relation to the receive-side `TryFrom<&ArrowData>` conversions, and includes a compile-checked doctest.

### Simplification
- #2113 — `refactor(node-api)`: `EventStream::is_empty()` used bitwise `&` on `bool`s instead of logical `&&`.
- #2112 (also) — the manifest→workspace path-resolution block was duplicated 4× across the C and C++ templates; extracted into a single `template::workspace_dir()` helper.

### Review notes
Findings that were investigated and **rejected** as false positives during verification: a claimed out-of-bounds in the scheduler loop rewrite (the loop returns immediately after mutation), claimed `Arc`-uniqueness hazards in the daemon hoist (no `get_mut`/`try_unwrap`/`into_inner` usage exists on these values), and several cold-path micro-optimizations not worth the churn. All PRs pass `cargo fmt`, `clippy -D warnings`, and their crate test suites locally.

---

phil-opp (commented):
Reviewed the scheduler hot-path changes. The index-based iteration over `last_used` is equivalent to the previous `clone()`-then-iterate: the loop `return`s immediately after the single `remove`/`push_back`, so it never iterates over the mutated deque. The `LazyLock<DataId>` sentinel removes the per-call `String` allocations without changing scheduling order.

No issues found.

---

phil-opp:

Checked the `Scheduler::next()` rewrite (index-based iteration with `remove(index)` + `push_back`) and the `LazyLock` sentinel. The loop `return`s immediately after the mutation, so the mutated `last_used` is never iterated further — no skipped or duplicated entries, and LRU/eviction order is identical to the previous `clone()`-then-iterate version, just without the per-call `VecDeque`/`String` allocations. The `id` borrow ends before `remove` (the `get_mut` reference points into the separate `event_queues` field), so it compiles cleanly. `DataId` equality is value-based, so the shared `NON_INPUT_EVENT` sentinel compares and hashes identically. Behavior is preserved; scheduler tests pass.

---

---

phil-opp:

I verified the index-based LRU iteration is correct: the loop `return`s immediately after the single `remove(index)` + `push_back`, so no element is skipped or double-visited and the eviction order is identical to the old `clone()`-based loop. The `LazyLock` sentinel is behaviorally equivalent since `DataId` equality/hashing is value-based, so `&*NON_INPUT_EVENT_ID` compares and hashes the same as a freshly-allocated `DataId` in all three call sites.

No issues found.

---

PR #2122, opened 2026-06-10
fix(mavlink2-bridge): reject multi-row command batches instead of silently truncating to row 0

### LINKED ISSUE #2075: mavlink2-bridge: multi-row command batches are silently truncated to row 0 — extra commands dropped on the dora→vehicle path
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. It was found by randomly selecting a component (`mavlink2-bridge`) and reviewing it for correctness issues. Please treat the severity assessment below as a starting point for a human triage, not a confirmed defect. This is **distinct from** the earlier automated review #2023 (which covered the *missing round-trip tests* for the write-path impls and explicitly ruled out the sequence/field-width concerns); this issue is about a concrete decode-path behavior on **multi-row** input batches.

## Summary

The Arrow → MAVLink decode path in `libraries/extensions/mavlink2-bridge/src/arrow_convert.rs` reads **only row 0** of each incoming `RecordBatch` and never checks the row count. A dora producer that publishes an **N-row** batch (N > 1) on a command input (e.g. `command_long_cmd`, `set_position_target_local_ned_cmd`) will have **rows 1..N-1 silently discarded** — only the first command is encoded and sent to the vehicle. There is no error, warning, or trace.

Because this is the **dora → vehicle** path (the side that arms motors and commands position/velocity), silently dropping all-but-the-first command is a safety-relevant silent data loss, not just a cosmetic gap.

## Affected code

`read_primitive` (the single chokepoint every `read_*` helper funnels through) guards against zero rows but then unconditionally returns `value(0)`:

```rust
// libraries/extensions/mavlink2-bridge/src/arrow_convert.rs:86-96
if arr.is_empty() {
    return Err(BridgeError::Mavlink(format!(
        "column '{name}' has zero rows; expected ≥1 row"
    )));
}
if arr.is_null(0) {
    return Err(BridgeError::Mavlink(format!(
        "column '{name}' is null at row 0"
    )));
}
Ok(arr.value(0))   // ← rows 1..N-1 are never read
```

Note the error message even says *"expected ≥1 row"* — i.e. the contract was written as "at least one row", but every `from_record_batch` impl only ever consumes row 0, so the effective contract is "exactly one row" and the surplus is dropped instead of rejected.

The dispatch path has no row-count guard either:

```rust
// binaries/mavlink2-bridge-node/src/main.rs:322-331
fn decode_input<T: MavlinkArrow>(data: &ArrayRef, input: &str) -> Result<T> {
    let struct_array = data.as_struct_opt().ok_or_else(|| { /* ... */ })?;
    let batch = RecordBatch::from(struct_array);   // N rows preserved here…
    T::from_record_batch(&batch)                   // …but only row 0 is decoded
        .with_context(|| format!("decoding {input} from incoming Arrow"))
}
```

A `StructArray` of length N round-trips into an N-row `RecordBatch`, so the row count *is* available at decode time — it is simply ignored.

## Why this is a correctness/safety risk

The module's own doc header states the convention is *"One row per call."* (`arrow_convert.rs:4-5`), and the negative-test file notes these inputs are public — *"any node in the dataflow can publish to them"* (`tests/arrow_malformed.rs`). But "one row per call" is enforced **nowhere**:

- A Rust/Python node that batches several setpoints/commands into a single Arrow message (a natural thing to do — e.g. a planner emitting a short trajectory of `SET_POSITION_TARGET_LOCAL_NED` rows) will have **only the first row flown**, with the rest silently dropped.
- The failure is invisible: no `Err`, no log. The operator believes N commands were sent; one was.

This is the opposite of the existing, deliberate hardening on this path: `read_primitive` was specifically written to turn malformed input into `BridgeError` rather than a panic (see the comment block at `arrow_convert.rs:57-70`). Multi-row input is just another malformed-shape case that should be handled explicitly rather than truncated.

## Verification

- `grep -rn "num_rows\|len() != 1" libraries/extensions/mavlink2-bridge/src binaries/mavlink2-bridge-node/src` → **no row-count guard anywhere**.
- `tests/arrow_malformed.rs` covers wrong-column-type, null-at-row-0, and zero-row cases for `COMMAND_LONG_DATA`, but has **no multi-row case** — so this behavior is untested.
- The encode side (`to_record_batch`) always emits exactly one row, so telemetry (vehicle → dora) is unaffected; the bug is purely on inbound commands.

## Reproducer sketch

```rust
// Build a 2-row COMMAND_LONG batch (two distinct commands) and decode it.
let batch = /* RecordBatch with the 11 COMMAND_LONG columns, each length 2,
               row0 = MAV_CMD_NAV_TAKEOFF, row1 = MAV_CMD_NAV_LAND */;
let decoded = COMMAND_LONG_DATA::from_record_batch(&batch).unwrap();
assert_eq!(decoded.command as u32, 22); // row 0 (TAKEOFF) — row 1 (LAND) silently lost
```

## Suggested fix

Reject (or, if intentional, explicitly iterate) multi-row batches at the decode boundary. The minimal, lowest-risk fix is a single guard in `decode_input` (one place, covers all six command inputs):

```rust
let batch = RecordBatch::from(struct_array);
if batch.num_rows() != 1 {
    bail!(
        "input '{input}' expects exactly one row per message, got {} \
         (multi-row command batches are not supported; send one command per message)",
        batch.num_rows()
    );
}
```

Alternatively, enforce it inside `read_primitive` (change the `is_empty()` check to `arr.len() != 1`) so the library contract is self-enforcing regardless of caller, and add a `rejects_multi_row` case to `tests/arrow_malformed.rs`. If batched commands are a desired feature, the proper fix is to loop over all rows in `handle_input` and send one MAVLink frame per row — but that is a larger change and should be a deliberate decision.

- **Outgoing sequence number always 0** (`MavHeader { sequence: 0 }` in `main.rs`): not a bug — independently confirmed against published mavlink-core 0.13.1 source that the concrete TCP `send()` overrides the header sequence with an internal `wrapping_add(1)` counter (consistent with the note in #2023).
- **Field widths vs. MAVLink wire spec** (incl. `SERVO_OUTPUT_RAW.time_usec` = `UInt32`): safe by construction, as already noted in #2023.
- **Zero-row and wrong-type batches**: already correctly rejected and tested.

## Other observation (lower priority, not the focus of this issue)
Inbound telemetry forwarded via the reader thread can sit in the `flume` channel for up to `POLL_INTERVAL` (100 ms) before `main` drains it, because `main` only drains the channel at the top of each loop iteration and then blocks in `events.recv_timeout(POLL_INTERVAL)` with no wake from the reader (`main.rs:453-481`). This adds up to ~100 ms of latency/jitter to telemetry outputs. Flagging for awareness; a human may judge this acceptable for v1 or worth a separate tracking issue.

phil-opp:
`from_record_batch` read only row 0 of an incoming RecordBatch, so a producer publishing a multi-row command batch (e.g. `[TAKEOFF, LAND]` on `command_long_cmd`) had every row after the first silently dropped on the dora-to-vehicle path — safety-relevant silent data loss. `read_primitive`, the single decode chokepoint all `MavlinkArrow` impls funnel through, now requires exactly one row and returns a `BridgeError` otherwise; the bridge node logs the decode error and keeps running.

Behavior change: out-of-tree producers that were sending multi-row batches (and silently losing rows 1..N) now get the whole message rejected with an error log — fail-closed instead of partial delivery. The encode side already emits exactly one row per batch (locked by `encode_matches_declared_schema`), so telemetry output is unaffected.

Closes #2075

### DIFF
--- a/libraries/extensions/mavlink2-bridge/src/arrow_convert.rs
+++ b/libraries/extensions/mavlink2-bridge/src/arrow_convert.rs
@@ -3,7 +3,7 @@
 //! Arrow-side conventions used by every impl:
 //!
 //! * One row per call. `to_record_batch` returns a 1-row `RecordBatch`;
-//!   `from_record_batch` reads row 0.
+//!   `from_record_batch` requires exactly one row and rejects anything else.
 //! * All MAVLink enums (regardless of wire width) are stored as `UInt32`
 //!   to match the Rust `repr(u32)` of the generated enum types.
 //! * Bitflag fields are stored at their bitflag's natural width
@@ -28,7 +28,8 @@ use std::sync::Arc;
 /// Bidirectional conversion between a MAVLink 2 message struct and a 1-row
 /// Apache Arrow `RecordBatch`. Each impl is symmetric: the schema returned
 /// by [`schema`] is what [`to_record_batch`] emits and what
-/// [`from_record_batch`] expects.
+/// [`from_record_batch`] expects; both sides carry exactly one row, and
+/// [`from_record_batch`] rejects any other row count.
 ///
 /// [`schema`]: MavlinkArrow::schema
 /// [`to_record_batch`]: MavlinkArrow::to_record_batch
@@ -62,6 +63,9 @@ fn build(fields: Vec<Field>, columns: Vec<ArrayRef>) -> BridgeResult<RecordBatch
 //   * wrong column dtype  → `as_primitive` would panic; downcast_ref
 //     returns None and we return a `BridgeError::Mavlink` instead.
 //   * zero-row batch      → `value(0)` would panic on out-of-bounds.
+//   * multi-row batch     → `value(0)` would silently drop every row
+//     after the first (e.g. the LAND in a [TAKEOFF, LAND] command
+//     batch); reject the whole batch instead.
 //   * null at row 0       → `value(0)` returns the slot but ignores the
 //     null bit, which silently invents a value; reject it.
 //
@@ -83,9 +87,11 @@ where
                 col.data_type()
             ))
         })?;
-    if arr.is_empty() {
+    let len = arr.len();
+    if len != 1 {
         return Err(BridgeError::Mavlink(format!(
-            "column '{name}' has zero rows; expected ≥1 row"
+            "column '{name}' has {len} rows; expected exactly 1 (one MAVLink message \
+             per batch — multi-row command batches are not supported)"
         )));
     }
     if arr.is_null(0) {
--- a/libraries/extensions/mavlink2-bridge/tests/arrow_malformed.rs
+++ b/libraries/extensions/mavlink2-bridge/tests/arrow_malformed.rs
@@ -81,7 +81,40 @@ fn rejects_zero_row_batch() {
 
     let err = COMMAND_LONG_DATA::from_record_batch(&batch).expect_err("expected decode error");
     let msg = err.to_string();
-    assert!(msg.contains("zero rows"), "unexpected error: {msg}");
+    assert!(
+        msg.contains("has 0 rows") && msg.contains("exactly 1"),
+        "unexpected error: {msg}"
+    );
+}
+
+#[test]
+fn rejects_multi_row_batch() {
+    // Two commands in one batch: row 0 is TAKEOFF, row 1 is LAND. Reading
+    // only row 0 would silently drop the LAND command — safety-relevant
+    // data loss — so the whole batch must be rejected.
+    let schema = COMMAND_LONG_DATA::schema();
+    let cols: Vec<arrow::array::ArrayRef> = vec![
+        Arc::new(Float32Array::from(vec![1.0, 101.0])),
+        Arc::new(Float32Array::from(vec![2.0, 102.0])),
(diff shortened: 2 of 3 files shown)


phil-opp (commented):
Reviewed the multi-row rejection. Funnelling the check through `read_primitive` (the single decode chokepoint every `MavlinkArrow` impl uses) is the right place to enforce it, the fail-closed behavior is correct for a safety-relevant path, and the encode side is unaffected (one row per batch, locked by `encode_matches_declared_schema`). The behavior change is clearly documented and covered by the new `rejects_multi_row_batch` test.

No issues found.

---

phil-opp:

Verified `read_primitive` is the single decode chokepoint every `MavlinkArrow::from_record_batch` impl funnels through (each reads ≥1 column), so the new exactly-one-row gate cannot be bypassed. `from_record_batch` is only reached on the incoming `*_cmd` write inputs — a MAVLink command frame is inherently one message, and all shipped example producers emit single-element arrays, so no legitimate multi-row use is broken; the old code silently read row 0 and dropped the rest. The encode/telemetry direction is unaffected. Test updates match the new error text, including the zero-row case (`has 0 rows` / `expected exactly 1`).

---

---

PR #2123, opened 2026-06-10
test(smoke): bound contract-test child processes with hard deadlines (fix 50-min CI hangs)

### LINKED ISSUE #2064: ci: contract-tests CircleCI job flakes/hangs (~50 min no_output_timeout) on unrelated PRs
## Summary

The `ci/circleci: contract-tests` job intermittently **hangs and fails** (~49–51 min, hitting a `no_output_timeout`) on PRs that touch nothing in the Python / ROS2 / example-dataflow surface it exercises. It has blocked the audit/merge of several unrelated PRs this week and is forcing admin-merges.

## Evidence

Observed failing on (none of which touch contract-test code):
- #2048 — `resolve_path` (1 file, `libraries/core`); job `17368`, ran ~51 min → failed.
- #2055 — mavlink reader (`binaries/mavlink2-bridge-node` only); job `17175`, ran ~49 min → failed.
- #2053 / #2054 — ros2-bridge **arrow** (not Python); #2060, #2063 — daemon/coordinator + test-only.

The job runs two `--uv` Python steps (`.circleci/config.yml`):
```
- cargo test -p dora-examples --test example-smoke contract_   # no_output_timeout: 30m
- cargo test -p dora-examples --test node-lifecycle-e2e -- lifecycle_python   # no_output_timeout: 15m
```
A ~50-min runtime is the signature of a `no_output_timeout` (a hung child process), not a clean assertion failure — a `--uv` dataflow is almost certainly not terminating.

## Suspected cause

Likely a hung `--uv` Python dataflow. A strong candidate is **PyPI `dora-rs` message-format drift (#1710)**: the CI venv may pull a `dora-rs` whose wire format has drifted from the workspace, so a node never produces the expected output and the dataflow runs until `--stop-after` never fires / the test polls forever. The local nightly harness works around this by installing `-e apis/python/node` into the scratch venv; the CircleCI `setup-python-uv` path should be checked for the same.

## Suggested next steps

1. Pin **which** test hangs: add `--nocapture` (already present) + a per-test wall-clock guard, or split `contract_` / `lifecycle_python` into separate steps with tighter `no_output_timeout` so the log shows the culprit before the 30-min cutoff.
2. Verify the CircleCI venv installs the **workspace** `dora-node-api-python` (`-e apis/python/node`), not PyPI `dora-rs` (cf. #1710).
3. If a single example is the offender, gate/skip it pending a fix and track separately, so the rest of the gate stays fast and reliable.

Until fixed, this job is a recurring false-negative on the merge gate.

phil-opp:
CircleCI's contract-tests job occasionally hangs ~50 minutes until no_output_timeout kills it with zero diagnostics (#2064): a dora child in the local dataflow path rarely never exits (a real-but-rare daemon-side shutdown wedge), and the harness waited on children with no wall-clock bound. This adds a `wait_with_deadline` helper (piped stdout/stderr drained on reader threads, `try_wait` polling; on expiry kill + panic with the full captured output) and routes every unbounded child wait through it — `run_dora_capture`, `run_smoke_test_local` (including its previously-invisible `dora build` step), and both `dora record`/`dora replay` steps. Service/action nodes are now pre-built outside the silent piped child, and the contract-tests `no_output_timeout` drops 30m/15m → 10m, so a wedge now fails in minutes with the dora logs attached instead of eating a 50-minute runner slot.

The daemon shutdown wedge itself is real but rare; it will be root-caused from the first captured log (follow-up issue to be filed).

Closes #2064

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -859,7 +859,7 @@ jobs:
             --
             --test-threads=1
             --nocapture
-          no_output_timeout: 30m
+          no_output_timeout: 10m
       # Python variant of the lifecycle test piggybacks on this
       # job because it needs uv (already set up via setup-python-uv
       # above). Rust + C++ variants run in the `e2e` job.
@@ -870,7 +870,7 @@ jobs:
             --
             --test-threads=1
             lifecycle_python
-          no_output_timeout: 15m
+          no_output_timeout: 10m
       - cargo-cache-save
 
   # ===== Benchmark regression check =====
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -184,6 +184,104 @@ fn needs_uv(yaml_path: &Path) -> bool {
         .unwrap_or(false)
 }
 
+/// Spawn `cmd` with piped stdout/stderr and wait for it with a hard
+/// wall-clock bound (#2064). A rare daemon shutdown wedge can leave a
+/// `dora` child running forever; an unbounded `.output()` then hangs the
+/// test until CI's `no_output_timeout` kills the whole job ~50 minutes
+/// later with no diagnostics. This helper converts that into a fast,
+/// loud failure: on deadline expiry the child is killed and the panic
+/// message carries the full captured stdout/stderr for triage.
+fn wait_with_deadline(
+    label: &str,
+    cmd: &mut Command,
+    deadline: Duration,
+) -> (std::process::ExitStatus, String, String) {
+    let mut child = cmd
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .unwrap_or_else(|e| panic!("{label}: failed to spawn: {e}"));
+
+    // Drain both pipes on dedicated threads so a chatty child can't
+    // deadlock on a full pipe buffer while we poll for exit.
+    let mut child_stdout = child.stdout.take().expect("child stdout not piped");
+    let mut child_stderr = child.stderr.take().expect("child stderr not piped");
+    let stdout_reader = std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        let _ = std::io::Read::read_to_end(&mut child_stdout, &mut buf);
+        buf
+    });
+    let stderr_reader = std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        let _ = std::io::Read::read_to_end(&mut child_stderr, &mut buf);
+        buf
+    });
+    let join_readers = |stdout_reader: std::thread::JoinHandle<Vec<u8>>,
+                        stderr_reader: std::thread::JoinHandle<Vec<u8>>| {
+        let stdout = stdout_reader.join().expect("stdout reader panicked");
+        let stderr = stderr_reader.join().expect("stderr reader panicked");
+        (
+            String::from_utf8_lossy(&stdout).into_owned(),
+            String::from_utf8_lossy(&stderr).into_owned(),
+        )
+    };
+
+    let start = std::time::Instant::now();
+    loop {
+        match child.try_wait() {
+            Ok(Some(status)) => {
+                let (stdout, stderr) = join_readers(stdout_reader, stderr_reader);
+                return (status, stdout, stderr);
+            }
+            Ok(None) => {}
+            Err(e) => panic!("{label}: failed to wait on child: {e}"),
+        }
+        if start.elapsed() >= deadline {
+            let _ = child.kill();
+            let _ = child.wait();
+            let (stdout, stderr) = join_readers(stdout_reader, stderr_reader);
+            panic!(
+                "{label}: child timed out after {:.1}s (deadline {deadline:?}); killed.\n\
+                 ---- stdout ----\n{stdout}\n---- stderr ----\n{stderr}",
+                start.elapsed().as_secs_f64()
+            );
+        }
+        std::thread::sleep(Duration::from_millis(200));
+    }
+}
+
+/// Harness self-test (#2064): a child that outlives its deadline must be
+/// killed and reported via panic, not waited on forever.
+#[test]
+#[cfg(unix)]
+#[should_panic(expected = "timed out")]
+fn contract_harness_wait_with_deadline_kills_hung_child() {
+    let mut cmd = Command::new("sleep");
+    cmd.arg("600");
+    wait_with_deadline(
+        "contract-harness-hung-child",
+        &mut cmd,
+        Duration::from_secs(2),
+    );
+}
+
+/// Harness self-test (#2064): a fast child's exit status and captured
+/// stdout pass through unchanged.
+#[test]
+fn contract_harness_wait_with_deadline_returns_fast_child_output() {
+    ensure_cli_built();
+    let dora = dora_bin();
+    let mut cmd = Command::new(&dora);
+    cmd.arg("--help");
+    let (status, stdout, _stderr) = wait_with_deadline(
+        "contract-harness-fast-child",
+        &mut cmd,
+        Duration::from_secs(60),
+    );
+    assert!(status.success(), "dora --help exited non-zero: {status:?}");
+    assert!(!stdout.is_empty(), "dora --help produced no stdout");
+}
+
 /// Run an example dataflow through the full WS control plane lifecycle.
 ///
 /// 1. `dora up` -- start coordinator + daemon
@@ -301,17 +399,21 @@ fn run_smoke_test_local(name: &str, yaml_path: &str, stop_after_secs: u64) {
 
     // Build first so local runs also have resolved artifacts (required for
     // Python/git/etc. and examples that don't prebuild their nodes in the test).
+    // Generous deadline: a cold cargo/uv cache can legitimately take minutes.
     let mut build_cmd = Command::new(&dora);
     build_cmd.args(["build", full_yaml.to_str().unwrap()]);
     if uv {
         build_cmd.arg("--uv");
     }
-    let build_status = build_cmd
-        .stdout(Stdio::null())
-        .stderr(Stdio::null())
-        .status()
-        .unwrap_or_else(|e| panic!("{name}: failed to run dora build: {e}"));
-    assert!(build_status.success(), "{name}: dora build failed");
+    let (build_status, build_stdout, build_stderr) = wait_with_deadline(
+        &format!("{name}: dora build"),
+        &mut build_cmd,
+        Duration::from_secs(600),
+    );
+    assert!(
+        build_status.success(),
+        "{name}: dora build failed\n---- stdout ----\n{build_stdout}\n---- stderr ----\n{build_stderr}"
+    );
 
     let stop_after = format!("{stop_after_secs}s");
     let mut cmd = Command::new(&dora);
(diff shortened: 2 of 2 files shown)


phil-opp (commented):
The `wait_with_deadline` helper itself is solid — reader threads drain both pipes so a chatty child can't deadlock the poll loop, the kill-and-capture-on-expiry path attaches the full output, and the harness self-tests cover both the hung-child and fast-child cases.

One finding worth a second look before merge:

**The build deadline and the lowered `no_output_timeout` may be too tight for cold-cache CI, risking false-positive timeouts.** The `dora build` step in `run_smoke_test_local` is bounded at 600s, and the same comment notes "a cold cargo/uv cache can legitimately take minutes." In parallel, the contract-tests `no_output_timeout` drops to 10m (= 600s, the same value). Two interacting concerns:

- Child stdout/stderr is now *captured* and only surfaced on failure, so during a long build/run CircleCI sees no console output. A legitimately slow-but-progressing phase (full workspace build + uv venv provisioning on a cold runner) that stays quiet past 10m would be killed by `no_output_timeout` before completing — and a cold build that genuinely needs >600s would also trip the in-test deadline and panic with "timed out", i.e. a false failure rather than the real wedge this PR targets.
- Because the two bounds are equal (600s build deadline vs 600s no_output), there's no headroom between "the harness reports the wedge with logs" and "CircleCI kills the job blind" — the very outcome the PR is trying to replace.

Consider giving the cold-build deadline more headroom than the wedge case needs (or deriving it from observed cold-cache build times), and/or keeping `no_output_timeout` comfortably above the largest per-child deadline so the in-test deadline always fires first and you get the captured logs.

— 🤖 This is a fully automated review by Claude (Claude Code), with no human in the loop. Treat the finding as advisory.

---

phil-opp:
**Automated review — no blocking issues found.**

I checked `wait_with_deadline` for the failure modes that matter: pipes are drained on reader threads *before* the `try_wait` loop (no wait-before-drain deadlock); the expiry path does `kill()` + `wait()` (no zombie) and joins readers only after the child is gone; normal exit and output-capture-before-panic both work. For `dora run`/`dora build`, only the `dora` process holds the test's pipe write-ends, so SIGKILL closes them and the reader threads reach EOF — the #2064 in-process-daemon shutdown wedge is genuinely unblocked.

One minor, non-blocking observation: `wait_with_deadline` buffers child stdout/stderr and only emits on panic, so a long `dora build` prints nothing to the CI console in the meantime. The CircleCI `no_output_timeout` is now 10m and the in-test build deadline is 600s (also 10m); a cold-cache build that legitimately nears its budget could trip the CI watchdog before the helper's own deadline/panic fires, defeating the "fast, loud failure with diagnostics" goal for that one case. Consider keeping the CI watchdog comfortably above the largest in-test deadline (or emitting a periodic heartbeat line).

---

---

phil-opp:
**🤖 Automated review by Claude (Claude Code)** — *This is a fully automated review with no human in the loop. Findings are advisory; please verify before acting.*

I reviewed the current diff and have one issue I don't think the previous review caught.

**`join_readers()` after a single-PID `kill()` can block forever — re-introducing the exact silent hang this PR is meant to fix.**

In the deadline-expiry branch of `wait_with_deadline` (`tests/example-smoke.rs`):
```rust
if start.elapsed() >= deadline {
    let _ = child.kill();      // SIGKILL to the `dora` CLI PID only
    let _ = child.wait();
    let (stdout, stderr) = join_readers(stdout_reader, stderr_reader); // <-- can block forever
    panic!(...);
}
```
`join_readers` joins the two reader threads, each of which is parked in `read_to_end`. `read_to_end` only returns at pipe EOF, i.e. once **every** write-end of the pipe is closed.

In the local `dora run` path the daemon spawns each node as its own OS subprocess in a **separate process group** (`binaries/daemon/src/spawn/prepared.rs`, `ProcessGroup::leader()`), and those nodes **inherit** the daemon's stdout/stderr — which in this harness are the captured pipe FDs (the daemon doesn't redirect node stdio). `child.kill()` targets only the `dora` CLI PID, so a node in its own process group can survive, keep the pipe write-end open, and `read_to_end` never returns. `join_readers()` then blocks indefinitely and the `panic!` below it is never reached — the test hangs with no diagnostics until CI's `no_output_timeout` kills the whole job blind, which is the #2064 behavior this PR sets out to eliminate.

This matters precisely because #2064 is a *daemon-shutdown wedge*: that's exactly the scenario where a node grandchild outlives the CLI kill. The two harness self-tests don't catch it — `sleep 600` and `dora --help` have no surviving grandchildren holding the pipe open.

Suggested fix: kill the whole process group rather than the single PID (e.g. `killpg` / negative-PID signal on Unix), and/or bound the reader join itself (timed join that drops the read on expiry) so the `panic!` always fires even if a write-end is still open.

(I'm deliberately not repeating the earlier point about the 600s build deadline equaling the 10m `no_output_timeout` — that one still stands but was already raised.)

---

phil-opp:
**Automated review (Claude)** — re-reviewed after the merge from main; reviewed the diff directly.

The helper itself looks correct: reader threads are spawned before the `try_wait` poll loop (no wait-before-drain pipe deadlock), the expiry path does `kill()` + `wait()` (no zombie), and readers are joined only after the child is gone.

One outstanding gap (raised in the earlier automated review and still unaddressed after the latest merge): `run_smoke_test_local`'s build deadline is `Duration::from_secs(600)` while the contract-tests `no_output_timeout` was lowered to `10m`. Because `wait_with_deadline` buffers all output and prints nothing until it panics, a cold-cache `dora build` that approaches its 600s budget can trip the 10m CI watchdog *before* the helper's own panic fires — defeating the "fail fast and loud with captured logs" goal for exactly that case. Keeping the CI `no_output_timeout` strictly above the largest in-test deadline (or emitting a periodic heartbeat from the reader threads) closes it.

Minor (pre-existing, not introduced here): on the timeout path `child.kill()` only kills the in-process-daemon CLI; node subprocesses are spawned in their own process groups and are orphaned. They don't hold the test's pipes, so there's no hang — just a leak until they self-exit. A process-group kill would be cleaner.

---

phil-opp:
Re-reviewed after the recent commits. The `wait_with_deadline` helper is still solid — reader threads drain both pipes so a chatty child can't deadlock the poll loop, the kill-and-capture-on-expiry path attaches the full output, and the self-tests cover the hung-child and fast-child cases. The new `ensure_service_nodes_built()` / `ensure_action_nodes_built()` pre-builds outside the piped child are a good fix for the previously-invisible build step.

One finding from the earlier review is **still outstanding** in the current diff (flagging again only because there are new commits and it wasn't addressed):

**The cold-build deadline and the lowered `no_output_timeout` have no headroom between them.** `run_smoke_test_local`'s `dora build` is bounded at 600s, and the contract-tests `no_output_timeout` is now 10m (= 600s) — the same value. Because child stdout/stderr is captured and only surfaced on failure, a legitimately slow-but-progressing cold-cache build that stays quiet can hit whichever 600s bound fires first. If `no_output_timeout` wins, CircleCI kills the job blind — exactly the diagnostic-less outcome this PR is trying to eliminate. Giving the in-test deadline meaningful headroom *below* `no_output_timeout` (so the harness always fires first and attaches logs), and/or sizing the cold-build deadline from observed cold-cache times, would make the guarantee actually hold.

Nothing else new stands out.

---

---

heyong4725 (changes requested):
Reviewed in depth (structured pass + two specialist passes + independent Claude and Codex adversarial passes — all four converged on the same top findings). The direction is exactly right and the helper itself is well built — piped reader threads, kill-then-reap, panic with captured output. Requesting changes on three points that all reviewers independently flagged, each with a small fix:

### 1. The deadline-kill path can recreate the hang it fixes, and poisons the serialized suite (`wait_with_deadline`)

`child.kill()` SIGKILLs only the direct `dora` process:

- **Pipe-EOF invariant is real but unenforced.** Today every dora spawn hop uses piped/null stdio (verified: `spawner.rs:260-262` nodes, `up.rs:149-151` coordinator/daemon, `build_command.rs` builds), so `read_to_end` does see EOF after the kill. But one future `Stdio::inherit()` anywhere in the CLI/daemon makes `join_readers` block forever *after* the kill — an unbounded hang with **zero** output (the panic never fires), strictly worse than today's failure mode.
- **Orphaned grandchildren poison the run.** Nodes are spawned as process-group leaders (`prepared.rs:492`), so they survive the kill outright. Worse: a killed `dora build`'s orphaned `cargo`/`uv` child keeps holding the target-dir/uv-cache file lock — the next test in this `--test-threads=1` suite then blocks silently on the lock, and the 10m `no_output_timeout` kills the job with the real culprit buried pages above.

**Fix (~15 lines):** spawn the harness child with `CommandExt::process_group(0)` and `libc::killpg` on deadline (kills the whole tree), and join the readers with a bounded wait that falls back to partial buffers. That also makes the helper future-proof against stdio-inheritance drift.

### 2. The most wedge-prone commands remain unbounded — and one lost CI headroom for nothing

The PR text says every unbounded child wait is routed through the helper, but the conversion covers `run_smoke_test_local`, `run_dora_capture`, and record/replay only. Still unbounded: `cleanup_stale` (`dora stop --all` / `destroy` / `down` — *shutdown* commands, which is exactly where the #2064 wedge lives), the entire networked `run_smoke_test` path (`dora up`/`build`/`start` + the `dora list` poll), the shell-gate tests (`example-smoke.rs:1788, 1851`), and `tests/node-lifecycle-e2e.rs:161` — whose `no_output_timeout` this PR cuts 15m→10m without giving it any harness deadline. A wedge there now dies at 10m instead of 15m, still with zero diagnostics. Suggest converting those paths (a `.status()`-shaped variant of the helper covers the `Stdio::null` call sites), or at minimum reverting the `lifecycle_python` timeout cut until that harness adopts the helper.

### 3. 600s build deadline vs 10m `no_output_timeout`: CircleCI wins the photo finish

`wait_with_deadline` buffers all child output, so the CI console is **silent for the entire wait**. A cold-cache `dora build` can legitimately use most of its 600s window; with adjacent quiet phases, total silence reaches 10m before the helper's diagnostic panic fires — CircleCI kills the job diagnostics-free, the exact outcome this PR eliminates. **Fix (1 line):** a keepalive `eprintln!` every ~60s in the `try_wait` poll loop (resets the silence clock *and* timestamps progress), after which the 10m reduction is safe.

### Informational (non-blocking)

- Contract tests that call `run_dora_capture(…, uv=true)` without a prior `dora build` put cold-cache venv provisioning inside the 180s grace — a new flake vector (the service/action tests got `ensure_*_built`, but builds themselves are unbounded per point 2).
- `let grace_secs = if uv { 180 } else { 120 };` is duplicated at two sites with the rationale comment at only one — worth a tiny named helper.
- The kill path has no Windows coverage (`cfg(unix)` self-test) while the helper runs unconditionally on Windows nightly; a portable hung-child (`ping -n 600 127.0.0.1` behind `cfg(windows)`) would close that.
- A `catch_unwind` self-test asserting the panic payload contains a child-emitted MARKER would pin the "panic carries captured output" contract that motivates the whole change.

Supporting datapoint for the premise: this PR's own `cli-python` CI failure is a live specimen — the template-project dataflow **succeeded**, then `dora run` wedged in post-completion zenoh teardown ("Unable to publish transport event: session closed" → silence → job killed at the deadline). Exactly the class this harness work targets, in a job it doesn't yet cover.

Happy to re-review quickly — points 1 and 3 are each small diffs and the rest can reasonably be follow-ups if you'd rather scope this PR down to match.

PR #2124, opened 2026-06-10
fix(node-api): bound zenoh teardown in DoraNode::drop so a wedged session cannot hang node shutdown

phil-opp:
The 2026-06-09 macOS nightly hung for 30 minutes (run 27190356408, auto-filed on #2089) because a node wedged inside the zenoh teardown block of `DoraNode::drop`: when zenoh's net runtime stalls, the `Session` close's internal 10s timeout is unenforceable and `Publisher::drop` -> `undeclare_impl()` has no timeout at all, so `report_closed_outputs` never runs and the daemon waits forever for `InputClosed`. This moves the zenoh teardown (publishers -> SHM provider -> session -> owned runtime, preserving the documented drop order and the teardown-before-daemon-notify ordering) onto a dedicated thread bounded by a 15s deadline (> zenoh's internal 10s); on timeout we warn and continue shutdown, leaking the wedged thread and its zenoh state until process exit — acceptable for exiting node processes, and long-lived hosts (e.g. Python GC) inherit only the bounded delay. Nodes without zenoh state keep dropping inline without spawning a thread.

Follow-up: bump the pinned `dora-node-api` rev in `examples/rust-dataflow-git` once this lands on main.

Refs #2089

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -102,6 +102,10 @@ pub const ZERO_COPY_THRESHOLD: usize = 4096;
 const ZENOH_FIRST_PUBLISH_MATCH_TIMEOUT: Duration = Duration::from_millis(200);
 const ZENOH_FIRST_PUBLISH_MATCH_POLL_INTERVAL: Duration = Duration::from_millis(5);
 
+/// Must exceed zenoh's internal 10s session close timeout, which is not
+/// enforceable when zenoh's net runtime is wedged.
+const ZENOH_TEARDOWN_TIMEOUT: Duration = Duration::from_secs(15);
+
 /// Allows sending outputs and retrieving node information.
 ///
 /// The main purpose of this struct is to send outputs via Dora. There are also functions available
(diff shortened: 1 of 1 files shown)


phil-opp (commented):
Reviewed the bounded zenoh teardown. The dedicated-thread + `recv_timeout` bound is sound, panics in the closure are contained via `catch_unwind`, the documented drop order (publishers → SHM provider → session → owned runtime) is preserved, and the leak-on-timeout trade-off is acceptable and well documented for exiting node processes.

One thing I checked specifically, since it's a behavior change: `_owned_runtime` is now dropped *before* `report_closed_outputs`/`report_outputs_done` (previously it was dropped at the end of `Drop`, after the daemon notification). That turns out to be safe — the control channel uses a synchronous `std::net::TcpStream` (`daemon_connection/tcp.rs::request`), which does not depend on the tokio runtime, so the daemon notification still works after the runtime is gone.

No issues found.

---

phil-opp:
**Automated review — no blocking issues found.**

The bounded teardown is implemented correctly: fields are moved out of `&mut self` via `Option::take()`/`mem::take` (no double-drop or use-after-move), the captured zenoh objects are `Send + 'static`, the `JoinHandle` is discarded on timeout so nothing blocks on join, and the happy path blocks on `recv_timeout` until teardown completes *before* notifying the daemon — preserving the teardown-before-`report_closed_outputs` ordering.

One non-blocking observation: on the timeout path the entire owned multi-thread tokio `Runtime` (plus its worker pool and the wedged teardown thread) is leaked, not merely delayed. The doc comment frames a long-lived host (e.g. Python GC) as inheriting "only the bounded delay", but a host that repeatedly creates/drops nodes hitting the wedged path would see unbounded thread/FD growth. That is acceptable as the lesser evil over a permanent hang, but the comment undersells the cost — worth tightening the wording so it is not read as a bounded one-time delay.

---

---

phil-opp:

I confirmed the teardown ordering (publishers → SHM provider → session → owned runtime) and the teardown-before-`report_closed_outputs` ordering are preserved on both the success and timeout paths; every field is `take()`/`mem::take`-d exactly once (no use-after-move); nodes without zenoh state still drop inline without spawning a thread; and the leak-on-timeout is sound (the `JoinHandle` is dropped/detached, the moved-in objects leak but there's no double-free or dangling pointer).

No correctness issues found. Minor (already raised, still applies): for a long-lived host that repeatedly creates/drops nodes hitting the wedged path, the leak is per-occurrence, so the doc comment's "only the bounded delay" slightly undersells the cost — worth tightening the wording.

---

phil-opp:
Re-reviewed the current diff after the recent commits. The bounded zenoh teardown still holds up: the closure is handed to a dedicated thread via a channel and bounded with `recv_timeout`, panics inside it are contained with `catch_unwind`, and the documented drop order (publishers → SHM provider → session → owned runtime, runtime last) is preserved so async cleanup can still run during the session close. Dropping the owned tokio runtime on a fresh `std::thread` rather than inline is the right call — it avoids the "cannot drop a runtime in an async context" panic. The leak-on-timeout trade-off is well documented and acceptable for exiting node processes, and the no-zenoh-state fast path still drops inline without spawning a thread. The three helper tests cover the fast, wedged, and panicking cases.

One minor edge (not blocking): the spawn-failure fallback runs the teardown — including `drop(runtime)` — inline, which could hit the runtime-drop-in-async-context panic if `Drop` runs on a tokio worker thread. It's a rare path (thread spawn failure) and no worse than the pre-PR behavior, so just noting it.

No issues found.

---

---

PR #2125, opened 2026-06-10
ci(examples): bound rust-dataflow-git run with a 120s stop-after safety net

phil-opp:
The rust-dataflow-git example pins its node binaries to rev f40b5e7e, so node-API fixes don't reach it until a rev bump — when a pinned node wedges, the runner's unbounded `run(dataflow, false)` call hangs until the GHA step timeout, as in the 2026-06-09 nightly (run 27190356408, 30 minutes of warn spam, auto-filed on #2089). This switches the runner to `RunCommand` with `stop_after = 120s` so the daemon's stop escalation (Stop -> 10s grace -> SoftKill -> +5s -> Kill) kills wedged nodes and the step fails in ~2.5 minutes with a clear error; a healthy run self-terminates in ~2s, so the bound only fires on a wedge.

Refs #2089

### DIFF
--- a/examples/rust-dataflow-git/run.rs
+++ b/examples/rust-dataflow-git/run.rs
@@ -1,6 +1,6 @@
-use dora_cli::{BuildConfig, build, run};
+use dora_cli::{BuildConfig, Executable, RunCommand, build};
 use eyre::Context;
-use std::path::Path;
+use std::{path::Path, time::Duration};
 
 fn main() -> eyre::Result<()> {
     let root = Path::new(env!("CARGO_MANIFEST_DIR"));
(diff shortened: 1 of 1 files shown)


phil-opp (commented):
Reviewed the stop-after bound. `RunCommand::execute()` routes the result through `handle_dataflow_result`, which returns `Err` whenever the dataflow result isn't ok — so a wedged node force-killed by the stop-after escalation (Stop → grace → SoftKill → Kill) surfaces as a non-zero exit and fails the CI step, while a healthy run that self-terminates before 120s shuts down cleanly with an ok result. Behaves as intended.

No issues found.

---

phil-opp:
@claude Doesn't this silence failing tests without fixing the root cause (updating the commit hash instead)?

phil-opp:
**Automated review — no correctness issues found.**

The `RunCommand { stop_after: 120s }` swap is wired correctly: a healthy run still self-terminates in ~2s (the `stop_after` timer is an independent spawn that only fires if the dataflow is still running at T+120s), and a node that ignores SIGTERM is SIGKILL'd → `Signal(9)`, which is *not* in `is_sigterm_like`, so it is classified as an error and the example exits non-zero. A genuinely wedged dataflow therefore does surface as a CI failure rather than being silently masked.

One blind spot worth flagging (sharpening the concern already raised in the thread): a node that is wedged but *still responds to SIGTERM* exits `Signal(15)`/143, which `is_sigterm_like` classifies as a **clean stop** (`Ok`). That specific hang would now pass green at T+120s instead of failing. If that class of wedge is one you care about catching, the 120s bound alone will not surface it.

---

---

PR #2126, opened 2026-06-10
fix(daemon): treat idle node connections as normal instead of warn-spamming TCP header timeouts

phil-opp:
The daemon applied `TCP_READ_TIMEOUT` (30s) to the frame-header read on node connections, so any healthy connection that simply stayed quiet produced a misleading "TCP read header timed out" warning — 184 of them in the 2026-06-09 nightly hang (run 27190356408), which derailed diagnosis, and they show up in green runs too.

`socket_stream_receive` now has a variant with an optional header timeout; the daemon's node-communication TCP path passes `None` because idling between requests is legitimate and the daemon must never give up on a live connection just because it is quiet. The body read keeps `TCP_READ_TIMEOUT`, so a mid-frame stall still surfaces at warn through the existing path. All other callers keep the previous behavior.

Refs #2089

### DIFF
--- a/binaries/daemon/src/node_communication/tcp.rs
+++ b/binaries/daemon/src/node_communication/tcp.rs
@@ -6,7 +6,7 @@ use std::{
 use super::{Connection, Listener};
 use crate::{
     Event,
-    socket_stream_utils::{socket_stream_receive, socket_stream_send},
+    socket_stream_utils::{socket_stream_receive_with_header_timeout, socket_stream_send},
 };
 use dora_core::uhlc::HLC;
 use dora_message::{
@@ -67,7 +67,9 @@ struct TcpConnection(TcpStream);
 
 impl Connection for TcpConnection {
     async fn receive_message(&mut self) -> eyre::Result<Option<Timestamped<DaemonRequest>>> {
-        let raw = match socket_stream_receive(&mut self.0).await {
+        // No header timeout: a node connection may legitimately stay idle
+        // between requests, so only mid-frame (body) stalls are faults.
+        let raw = match socket_stream_receive_with_header_timeout(&mut self.0, None).await {
             Ok(raw) => raw,
             Err(err) => match err.kind() {
                 ErrorKind::UnexpectedEof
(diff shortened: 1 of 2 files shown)


phil-opp (commented):
Reviewed the optional-header-timeout split. Since the daemon↔node TCP path is always local (same machine), a dead node delivers FIN/RST and is caught by the existing `UnexpectedEof`/`ConnectionAborted` arm — so dropping the header timeout for that path doesn't risk a permanently-stuck receive, it just stops penalizing legitimately idle connections. The body read still keeps `TCP_READ_TIMEOUT`, so a mid-frame stall surfaces as before, and both new tests (idle-then-deliver, stalled-body) pin the intended behavior.

No issues found.

---

phil-opp:

Confirmed only the daemon node-communication path (`tcp.rs`) passes `None` for the header timeout; `local_listener.rs` still goes through the `Some(TCP_READ_TIMEOUT)` delegate, so other callers are byte-for-byte unchanged. `None` gates only the header read — the body read is unconditionally bounded by `TCP_READ_TIMEOUT`, so a mid-frame stall still faults at warn. The key liveness concern does not materialize: a dead local node closes its socket → the untimed `read_exact` returns `UnexpectedEof` → the task exits cleanly, and a hung-but-alive node is reaped by the independent `check_node_health` wall-clock watchdog, not by this TCP timeout. The new tests cover both the idle-stays-pending and stalled-body cases.

---

---

phil-opp:

I confirmed: the only other non-test caller (`local_listener.rs`) still goes through the wrapper with `Some(TCP_READ_TIMEOUT)`, so its behavior is unchanged; passing `None` gates only the header read while the body `read_exact` stays unconditionally bounded by `TCP_READ_TIMEOUT`; and removing the header timeout doesn't create a liveness/leak regression — a hung-but-alive node is reaped by the independent `NodeHealthCheck` heartbeat, and a node that closes its socket makes the untimed `read_exact` return `UnexpectedEof` so the task exits.

No issues found.

---

phil-opp:
Re-reviewed the current diff after the recent commits. The optional-header-timeout split is unchanged in substance and still looks correct: `socket_stream_receive_with_header_timeout(.., None)` is used only on the daemon↔node TCP path, where the connection is local and a dead node delivers FIN/RST that's caught by the existing `UnexpectedEof`/`ConnectionAborted` arm — so dropping the header timeout there stops penalizing legitimately idle connections without risking a permanently-stuck receive. The body read keeps `TCP_READ_TIMEOUT`, so a mid-frame stall still surfaces as a warn. All other callers go through the wrapper that preserves the previous 30s header timeout, and the two new tests pin both the idle-then-deliver and stalled-body behaviors.

No issues found.

---

---

PR #2127, opened 2026-06-10
fix(qa): replay cluster recording at real-time speed to stop DropOldest drops

phil-opp:
The nightly cluster-record-replay job replayed the recording with `--speed 0` (as fast as possible), blasting the ~100-message burst into rust-sink's input within ~20ms. That input uses the default size-10 DropOldest queue, so ~20 messages were silently dropped and the exact seed(42) baseline validation failed — the 2026-06-10 nightly failure auto-filed on #2089 (80 of 100 values, first divergence at index 48; the "after 8 ticks" group missing entirely from the sink output while both replay nodes reported 100/100 replayed).

Replay at default real-time speed instead: the recording's messages are ~10ms apart (rust-node ticks at 10ms) and the live recording phase delivered 100/100 at that pace on the same runner class, so a size-10 queue trivially keeps up. Adds ~1s to the job against its 120s timeout.

Refs #2089

### DIFF
--- a/scripts/qa/ci-nightly-jobs.sh
+++ b/scripts/qa/ci-nightly-jobs.sh
@@ -1159,13 +1159,22 @@ EOF
   # `dora replay` runs a local single-daemon `dora run` from the deploy-free
   # descriptor stored in the .drec. The recorded source nodes (rust-node and
   # rust-status-node) are swapped for replay nodes; rust-sink reprocesses the
-  # replayed status strings. `--speed 0` replays as fast as possible. Invoked
-  # from the repo root so the replay-node binary resolves under target/debug;
-  # the run's working_dir is the .drec's parent ($WORK), so node logs land in
-  # $WORK/out/.
-  echo "=== dora replay $DREC (local single-daemon run) ==="
+  # replayed status strings. Invoked from the repo root so the replay-node
+  # binary resolves under target/debug; the run's working_dir is the .drec's
+  # parent ($WORK), so node logs land in $WORK/out/.
+  #
+  # Replay at the recorded (real-time) speed -- NOT `--speed 0`. The exact
+  # baseline match below requires lossless delivery into rust-sink, but
+  # dora inputs default to `queue_size: 10` with the `DropOldest` overflow
+  # policy. `--speed 0` blasts all ~100 status outputs near-instantly, so the
+  # sink's size-10 queue overflows and silently drops the events it can't keep
+  # up with -- a flaky, scheduling-dependent ~80-of-100 sequence that diverges
+  # mid-stream (#2089). Real-time pacing (~10ms between outputs) lets the sink
+  # drain each message before the next arrives, which is exactly why the
+  # non-cluster `record-replay` job (also default speed) is stable.
+  echo "=== dora replay $DREC (local single-daemon run, real-time speed) ==="
   rm -rf "$WORK/out"
-  if ! timeout 120s dora replay "$DREC" --speed 0; then
+  if ! timeout 120s dora replay "$DREC"; then
     echo "ERROR: dora replay failed or exceeded 120s"
     rm -rf "$WORK"
     return 1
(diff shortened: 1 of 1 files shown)


phil-opp (commented):
The diagnosis is correct — `--speed 0` bursts ~100 outputs into a size-10 `DropOldest` queue and silently drops them — and real-time pacing does make the burst fit. The doc-comment rewrite explaining why is good.

One finding: **this trades one timing assumption for another rather than removing the dependence on timing.** The test is a *fidelity* check (replay must exactly reproduce the seed(42) baseline), but with real-time pacing it now passes only because the sink happens to drain each message within the ~10ms inter-output gap. On a loaded or slow CI runner a transient sink stall >10ms could still overflow the size-10 `DropOldest` queue and produce a flaky failure — a timing artifact, not a real record/replay regression.

A more robust fix would make the sink input *lossless* (a queue large enough to hold the full ~100-message sequence, or a blocking/`DropNone`-style policy on that input) so delivery is guaranteed regardless of replay speed or runner load. Then the test validates record/replay fidelity independent of pacing — and you could even keep `--speed 0`. Real-time pacing is a reasonable stopgap, but as written the fidelity test remains scheduling-sensitive.

— 🤖 This is a fully automated review by Claude (Claude Code), with no human in the loop. Treat the finding as advisory.

---

phil-opp:

Verified the fix against the code: replay default speed is `1.0` (`binaries/cli/src/command/replay.rs`), and the replay node only `thread::sleep`s the recorded inter-output delta when `speed > 0.0` (`binaries/replay-node/src/main.rs`), so `--speed 0` was indeed blasting the burst back-to-back into the size-10 `DropOldest` queue. Dropping the flag restores ~10ms pacing, which the size-10 queue trivially keeps up with; the ~1s recording stays far under the 120s timeout, and the exact seed(42) baseline comparison remains meaningful. This also matches the already-stable non-cluster `record-replay` job, which runs `dora replay` with no `--speed`.

(One non-blocking note: correctness still rests on a timing margin — the sink draining each message within the ~10ms gap — rather than hard backpressure, since the queue size lives inside the generated `.drec` descriptor and is not controllable from the script. That is the same property the existing stable job relies on, so it is an acceptable pragmatic fix; only worth revisiting if the cluster job re-flakes.)

---

---

heyong4725:
Heads-up: this looks superseded by #2146 (merged 2026-06-10 as 563b5d38). That fix addressed the same DropOldest losses at the root — `dora replay` now sizes receiver queues from the recorded message counts (+ `queue_policy: backpressure`), so the cluster-record-replay validation passes at `--speed 0` losslessly (the 06-11 nightly's cluster-record-replay job was green). Replaying at real-time speed would also work but makes the job ~10× slower and leaves the underlying lossiness for any other fast-replay consumer. Suggest closing this one unless there's a case the queue-sizing fix doesn't cover.

phil-opp:
**Automated review (Claude)** — reviewed the diff directly, not the description.

The change itself does what it claims: `dora replay` defaults to real-time speed (the replay node only sleeps when `speed > 0`), so dropping `--speed 0` restores ~10ms pacing and the size-10 `DropOldest` queue keeps up.

That said, I'm not sure this is the right fix. #2146 (merged to main as `563b5d3`) appears to address the same `DropOldest` losses at the root — sizing replay receiver queues from the recorded message counts and switching the policy to backpressure — which would make the cluster-record-replay validation lossless *even at* `--speed 0`. If that's accurate, this PR masks the lossiness with timing rather than fixing it, makes the job ~10x slower, and leaves fast replay lossy for any other consumer. The base of this branch also looks like it predates #2146, so the two touch the same script region.

Could you confirm whether #2146 already covers the cluster-record-replay job? If so, this is likely better closed in favor of it.

---

phil-opp:

No issues found. Replacing `--speed 0` with default real-time pacing is a sound fix for the size-10 `DropOldest` overflow on rust-sink's input: at ~10ms spacing the queue drains between messages, matching the stable non-cluster `record-replay` job. The change is confined to the nightly QA script and stays well within the 120s timeout.

---

phil-opp:
@claude

PR #2132, opened 2026-06-10
Add formal verification (Kani proofs) for shared-memory containment and auth comparison

phil-opp:
## Summary

Adds a formal-verification tier to dora's QA ladder: **Kani (CBMC) model-checking proofs** over the two most safety-critical pure functions, plus proptest properties for the untrusted-YAML parsers, a `make qa-kani` gate, and a nightly CI job so the proofs cannot silently rot. Writing the proofs surfaced and fixed two latent issues in the shared-memory path.

### `dora-core` — shared-memory buffer containment (`libraries/core/src/metadata.rs`)

The containment check behind `ArrowTypeInfo::from_array` (which converts Arrow buffer pointers into offsets handed across processes, and has had two historical off-by-one bugs) is extracted into a pure `buffer_offset_in_region` helper with proofs **exhaustive over the full `usize` input space**:

1. **Soundness + panic-freedom** — accepted ⇒ the buffer really is contained and the offset is exact
2. **Completeness** — contained ⇒ accepted (the property the historical off-by-one bugs violated)
3. **Accurate rejection** — a buffer before the region is always reported as such

### `dora-message` — auth token comparison (`libraries/message/src/auth.rs`)

`constant_time_eq(a, b) == (a == b)` proven for all slice pairs up to length 8 (the loop is length-uniform, so the bound covers every control-flow path). Kani proves functional properties, not timing; the doc records `subtle` as the hardening option if stronger timing guarantees are ever needed.

## Hardening found by writing the proofs

- **Release-mode wraparound mis-acceptance**: the old end-of-region comparison used unchecked additions (`region_start + region_len`, `ptr + len`) that wrap in release builds — a wrapped `ptr + len` could *accept* an out-of-region buffer with a bogus offset. The predicate is now stated with guarded subtractions that cannot wrap (and correctly handles regions ending at the top of the address space).
- **Latent provenance UB removed**: `ptr.offset_from(region_start)` requires both pointers to share an allocation — a precondition `from_array` cannot locally guarantee. Replaced with proven-safe integer subtraction; a new test covers the cross-allocation rejection case that was UB to even write before.

## Property-based tests (proptest)

`libraries/core/src/descriptor/validate.rs` gains properties for the user-YAML parsers, where random sampling beats symbolic strings:

- `validate_ros2_name`: total (never panics), complete (structurally valid names accepted), sound (accepted ⇒ all documented ROS2 naming rules hold)
- `parse_byte_size`: total; exact multiplication for all 7 integer unit spellings (upper- and lowercase) with overflow reported, never wrapped; bare-integer identity; float-path totality

## Infrastructure

- `scripts/qa/kani.sh` + `make qa-kani` / `make qa-kani-install` — **auto-discovers** proof crates by scanning for `#[kani::proof]`, so a stale crate list can't silently skip new proofs (~5-10 min cold, <1 min warm)
- **`kani-proofs` nightly job** in `.github/workflows/nightly.yml` (+ parity job in `scripts/qa/ci-nightly-jobs.sh`): `cfg(kani)` code is never compiled by the normal fmt/clippy/test gates, so without this a refactor could break a harness unnoticed. Deliberately **not** added to PR CI per the lean-CI policy.
- `cfg(kani)` registered via `[workspace.lints.rust]` (crates opt in with `[lints] workspace = true`); `proptest` moved to `[workspace.dependencies]`
- `docs/formal-verification.md` — the verification ladder, proof inventory, how to add a new proof, and a prioritized backlog (loom for the daemon restart-loop atomics, coordinator `pending_replies`, ROS2 bridge FFI invariants)

## Validation

- `scripts/qa/kani.sh`: 4/4 harnesses verified
- Full workspace test suite green (110 suites); `dora-core` 173 tests incl. the new property tests
- fmt, clippy `-D warnings` (workspace), typos, unwrap-budget, `cargo check --examples` all green
- nightly.yml parses; shell scripts pass `bash -n`

Caveat: the nightly job's first real execution will be the next scheduled nightly run (the workflow auto-files a `nightly-regression` issue if anything is off; it also re-runs automatically on pushes to main that touch nightly.yml).

https://claude.ai/code/session_01DfM7EJhQFGgEo4FZRYZHJp

---
_Generated by [Claude Code](https://claude.ai/code/session_01DfM7EJhQFGgEo4FZRYZHJp)_


phil-opp:
**Automated review by Claude** — _this is a fully automated review; a human has not vetted these findings._

I reviewed the diff directly (not the PR description), focusing on the load-bearing claims in the shared-memory containment rewrite and the auth comparison.

**No issues found.** Specifically:

- **`buffer_offset_in_region` (`libraries/core/src/metadata.rs`)** — the new predicate is correct and a genuine improvement. `ptr - region_start` is guarded by the prior `ptr < region_start` check (no underflow); `region_len - offset` is protected by the `||` short-circuit on `offset > region_len` (no underflow). The accept condition is equivalent to `ptr + len <= region_start + region_len` computed entirely with non-wrapping subtraction, so the release-mode wraparound mis-acceptance the old `ptr + len > region_end` form allowed is genuinely closed. Boundary cases (`ptr == region_start`, empty buffer, buffer ending exactly at the region end / top of address space) are accepted correctly — no off-by-one. Replacing `unsafe ptr.offset_from(region_start)` (which carries a same-allocation provenance precondition `from_array` cannot guarantee) with checked integer subtraction is a legitimate soundness fix.
- **Kani harnesses** — soundness (accepted ⇒ contained, offset exact), completeness (contained ⇒ accepted), and the start-before-region rejection each match the implementation. Together soundness + completeness already pin not-contained ⇒ rejected, so behavior is fully constrained even though there's no harness asserting the `EndsPastRegion` direction specifically (a coverage nit, not a bug).
- **`constant_time_eq` (`libraries/message/src/auth.rs`)** — length check + XOR-accumulate loop with no early exit is modeled faithfully; `constant_time_eq(a,b) == (a == b)` over lengths ≤ 8 with `#[kani::unwind(9)]` is sound (functional, not timing — correctly documented as such).
- **proptest properties (`validate.rs`)** — checked against the real `parse_byte_size` (multiplier table `1<<10 / 1<<20 / 1<<30` matches `1024 / 1024² / 1024³`, overflow reported not wrapped) and `validate_ros2_name` (token/digit/slash rules). The asserted properties mirror the implementations; none are vacuous or wrong.

The infrastructure (auto-discovery in `kani.sh`, the nightly `kani-proofs` job, `cfg(kani)` lint registration) is consistent with the diff. Note CI shows `unstable`, but no failing check is attributable to this diff's logic — worth a re-run.

---

phil-opp:

I re-checked the load-bearing production change — `buffer_offset_in_region` in `libraries/core/src/metadata.rs`. The `ptr - region_start` subtraction is guarded by the `ptr < region_start` early return, and `region_len - offset` by the `offset > region_len` short-circuit, so the accept condition is exactly `offset + len <= region_len` computed with no addition. The release-mode wraparound mis-acceptance is genuinely closed and a region ending at the top of the address space is handled. Replacing `unsafe ptr.offset_from(region_start)` (which carries a same-allocation provenance precondition `from_array` can't satisfy) with integer subtraction is a real soundness improvement, and the Kani harnesses (soundness + completeness + before-region rejection) are non-vacuous.

No issues found.

---

phil-opp:
Reviewed the diff directly, focusing on the two source-logic changes (the doc/script/CI scaffolding around them is straightforward).

`libraries/core/src/metadata.rs`: extracting the containment check into `buffer_offset_in_region` is correct. The predicate uses guarded subtractions (`ptr < region_start` early-out, then `offset > region_len || len > region_len - offset`) so nothing can wrap even for a region ending at the top of the address space, and the empty-buffer-at-region-end case is accepted (`offset == region_len, len == 0` passes) — which is the case the historical off-by-one rejected. Replacing `ptr.offset_from(region_start)` with integer subtraction also removes a genuine provenance UB (the two pointers need not share an allocation). The three Kani harnesses (soundness+panic-freedom, completeness, accurate rejection) match the helper's contract.

`libraries/message/src/auth.rs`: the `constant_time_eq == ==` proof bounded at `MAX_LEN = 8` is justified — the loop body is length-uniform so the bound covers every control-flow path — and the doc correctly scopes the proof to functional equality (not timing).

No issues found.

---

---

PR #2145, opened 2026-06-10
fix(core): share local Python bindings wheel across dora processes (#2143)

### LINKED ISSUE #2143: Nightly: Smoke suite & CLI Tests chronically killed at timeout — per-env maturin builds of Python bindings dominate job time
### Summary

Three long nightly jobs are being killed at their `timeout-minutes` limit on (nearly) every run, reported as `cancelled` rather than `failure` — so they never produce a failure excerpt and have been hiding behind the visible flakes in #2082 / #2089:

| Job | Budget | History |
|---|---|---|
| Smoke suite (`tests/example-smoke.rs`, ubuntu) | 60 min | killed at 60 min on **every scheduled run since at least 2026-06-01** (10+ consecutive; the only non-timeouts in that window were two real failures on 06-01) |
| CLI Tests (windows-latest) | 90 min | killed at 90 min every run since at least 06-05 |
| CLI Tests (macos-latest) | 90 min | killed 06-06, 06-09, 06-10; passed 06-05 at 75 min — already at the edge |

CLI Tests (ubuntu) still passes but has crept to 52–75 min. Note the smoke-suite budget was already raised once for exactly this symptom (see the comment above `timeout-minutes: 60` in `nightly.yml` referencing the 04-20/04-21 runs) — the suite has outgrown it again.

Recent runs: [06-10](https://github.com/dora-rs/dora/actions/runs/27260677135), [06-09](https://github.com/dora-rs/dora/actions/runs/27190356408), [06-08](https://github.com/dora-rs/dora/actions/runs/27123358401), [06-07](https://github.com/dora-rs/dora/actions/runs/27086112725), [06-06](https://github.com/dora-rs/dora/actions/runs/27055800585), [06-05](https://github.com/dora-rs/dora/actions/runs/27007514297).

### Root cause: per-env from-source maturin builds of the workspace Python bindings

These jobs are build-bound, not test-bound or hung. Every Python example/test builds its own uv env, and each env compiles `dora-node-api-python` from workspace source via `maturin pep517 build-wheel`, with no wheel reuse across envs.

**Smoke suite, 06-10 run** — the gaps between consecutive `ok` lines are env builds, ~5–6 min per Python test:

```
08:02:57 test smoke_local_module_dataflow ... ok
08:08:58 test smoke_local_python_async ... ok          (+6 min)
08:14:05 test smoke_local_python_concurrent_rw ... ok  (+5 min)
08:19:06 test smoke_local_python_drain ... ok          (+5 min)
08:24:04 test smoke_local_python_log ... ok            (+5 min)
08:29:16 test smoke_local_python_multiple_arrays ... ok (+5 min)
```

At the 60-min kill, the runner's orphan-process cleanup shows the suite mid-build: `cargo`, `example_smoke`, `dora`, `uv`, `python`, `maturin`, `rustc`.

**CLI Tests (macos), 06-10 run** — step-time breakdown of the 90 min:

```
07:33:40  cargo install --path binaries/cli           ~17 min
07:57:18  uv venv --seed step (incl. bindings build)  ~16 min
08:31:56  operator-API prebuild (#2085 mitigation)    ~21 min
08:53:14  python-multiple-arrays: maturin pep517      ~10 min
09:03:32  job killed at 90 min
```

The `python-multiple-arrays` dataflow itself had been running normally for 18 s (inside its `--stop-after 30s` window) when the job timer fired — the *tests* are fine; the budget is consumed by repeated identical Rust compilation. Note the 10-min maturin build at 08:53 happened **after** an earlier env had already built the same bindings in the same job — there is no cache hit even within one job.

### Why it surfaced now

Test count grew (the #1716 nightly rebalance, service-action, log-sinks, #2013 cluster-record-replay, drain/concurrent-rw examples), each addition paying the full bindings build. Totals crossed the budgets in early June.

### Proposed fix

Build the bindings wheel **once per job** and share it:

```bash
# once, after toolchain setup
maturin build --release -m apis/python/node/Cargo.toml -o /tmp/dora-wheels
export UV_FIND_LINKS=/tmp/dora-wheels   # or pass --find-links to each env install
```

Each per-example env then installs the prebuilt wheel instead of recompiling, collapsing N×5-min builds into one. This keeps the #1710 invariant (envs use workspace bindings, never PyPI `dora-rs`) because the wheel is built from the workspace tree. The same pattern applies to `tests/example-smoke.rs` (the helper that provisions envs) and to the CLI Tests workflow steps.

Raising `timeout-minutes` again would only postpone the next crossing, as it did after 04-21.

### Impact

Until fixed, `CLI Tests (macos)` cannot produce the green run needed to formally confirm the #2090 prewarm fix (#2089), and the nightly bot keeps appending unrelated timeout noise to open nightly-regression issues.

heyong4725:
## Summary

Fixes #2143 — nightly Smoke suite (60 min) and CLI Tests (90 min) jobs have been killed at their `timeout-minutes` budgets on (nearly) every run since early June because **every dora process rebuilt the workspace Python bindings from source via maturin**: the built wheel was cached in-memory only and written to a per-PID dir, so each `dora build`/`dora run` invocation and each smoke test paid a full 5–10 min compile.

This keys the on-disk wheel cache by **source fingerprint + platform** instead, making it cross-process:

- The wheel builds into a per-PID staging dir (`build-<pid>`), then is promoted via `fs::rename` to a shared `fp-<fingerprint>-<os>-<arch>` dir. Losing the promotion race means another process just built the same wheel — its copy wins and staging is discarded.
- A wheel-less shared dir (e.g. left by an interrupted cleanup) is replaced on the next promotion, so the cache self-heals instead of failing permanently (matters on Windows, where renaming onto any existing dir fails).
- Cleanup age-evicts all managed dirs uniformly at 24 h, **including the current fingerprint's**: the fingerprint hashes workspace `Cargo.lock` + root `Cargo.toml` + `apis/python/node` sources but not workspace dependency crates (`libraries/message`, `apis/rust/node`, …), so eviction bounds that drift to one day. For comparison, the pre-existing env reinstall marker accepts the same staleness class indefinitely for existing envs.
- The platform suffix prevents checkouts visible to several platforms (WSL + native Windows, NFS-shared workspaces) from serving each other incompatible wheels.

## Measured

Local macOS (arm64), `examples/python-multiple-arrays` with fresh envs each time:

| Run | Wall time | maturin invocations |
|---|---|---|
| 1st process (cold cache) | 2m03s | 1 (built + promoted) |
| 2nd process (warm cache) | **0.25s** | **0** |

Both envs in run 2 install from `target/dora-python-wheels/fp-<hash>-macos-aarch64/dora_rs-*.whl`.

In CI this collapses N×(5–10 min) bindings builds per job into one: the smoke suite's per-test gaps (5–6 min each on the 06-10 nightly) and the CLI Tests' per-step rebuilds all become cache hits after the first build. No workflow changes needed — the daemon/CLI pick this up transparently.

## Not in scope (noted for follow-up)

- Concurrent cold-start processes still build redundantly (no cross-process build lock); correctness is preserved via promotion races, and CI paths are sequential. Same behavior as before this PR — previously *every* process built.
- The fingerprint's blind spot for workspace dependency crates is pre-existing (shared with the env reinstall marker); this PR bounds its effect to 24 h rather than extending it indefinitely.

## Test plan

- [x] New unit tests: fingerprint+platform key composition, staging→shared promotion, lost-race reuse, wheel-less-corpse replacement, missing-wheel error, uniform age eviction, fresh-dir survival (`cargo test -p dora-core --features build` — 188 pass)
- [x] `cargo clippy --all-features -D warnings`, `cargo fmt --check`, `make qa-fast`, `cargo check --examples` all green
- [x] End-to-end two-process verification on macOS (table above)
- [ ] Nightly Smoke suite / CLI Tests go green after merge (the real confirmation, tracked in #2143)

### DIFF
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -391,11 +391,17 @@ async fn dora_runtime_install_args(
     }
 }
 
-/// Builds the in-tree Python API once per Dora process and returns its wheel path.
+/// Builds the in-tree Python API at most once per source fingerprint and returns its wheel path.
 ///
 /// Installing the local source tree directly asks maturin to build once per managed
-/// node env. A wheel install keeps local checkout behavior while sharing that build
-/// across all node envs created by the current CLI invocation.
+/// node env. The wheel is built into a per-process staging dir and then promoted to a
+/// fingerprint-and-platform-keyed shared dir under `target/dora-python-wheels/`, so
+/// subsequent Dora processes (separate `dora build`/`dora run` invocations, each smoke
+/// test) reuse one build instead of recompiling per process (#2143).
+///
+/// The fingerprint does not cover workspace dependency crates outside
+/// `apis/python/node` (same scope the env reinstall marker already accepts), so shared
+/// wheels are age-evicted after [`LOCAL_DORA_WHEEL_CACHE_MAX_AGE`] to bound that drift.
 async fn ensure_local_dora_python_wheel(
     source: &LocalDoraPythonSource,
     envs: &Option<BTreeMap<String, EnvValue>>,
@@ -408,10 +414,19 @@ async fn ensure_local_dora_python_wheel(
         return Ok(wheel.clone());
     }
 
-    let wheel_dir = local_dora_python_wheel_dir(&source.package_dir)?;
+    let wheel_root = local_dora_python_wheel_root(&source.package_dir)?;
+    cleanup_stale_local_dora_python_wheel_dirs(&wheel_root, LOCAL_DORA_WHEEL_CACHE_MAX_AGE);
+
+    let shared_dir = local_dora_python_shared_wheel_dir(&wheel_root, &source.fingerprint);
+    if let Ok(wheel) = find_local_dora_python_wheel(&shared_dir) {
+        cache.insert(source.clone(), wheel.clone());
+        return Ok(wheel);
+    }
+
+    let wheel_dir = wheel_root.join(format!("build-{}", std::process::id()));
     std::fs::create_dir_all(&wheel_dir).wrap_err_with(|| {
         format!(
-            "failed to create local Dora Python wheel cache `{}`",
+            "failed to create local Dora Python wheel staging dir `{}`",
             wheel_dir.display()
         )
     })?;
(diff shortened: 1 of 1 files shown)


heyong4725:
Found one issue that should be fixed before relying on the 24h staleness bound.

- libraries/core/src/build/build_command.rs:410: ensure_local_dora_python_wheel checks LOCAL_DORA_WHEEL_CACHE before running cleanup_stale_local_dora_python_wheel_dirs. Once a long-lived process, notably dora-daemon, has cached a wheel path, later builds with the same LocalDoraPythonSource return that path indefinitely as long as the file still exists, so the new 24h eviction never gets a chance to bound the fingerprint blind spot for workspace dependency crates. Distributed builds are handled inside the daemon process, which can stay up across many dora build requests, so this is not just a short-lived CLI concern. Move the cleanup/staleness check before accepting the in-memory entry, or validate the cached file age and evict the map entry before returning.

Validation run during review:
- cargo test -p dora-core --features build

heyong4725:
Confirmed and fixed in 205e8117. The in-memory hit at the old line 410 returned before cleanup ran, so a long-lived daemon kept serving its cached path indefinitely — the 24h bound only ever applied to fresh processes, which undercuts exactly the property the PR claims for the fingerprint blind spot.

Rather than re-validating the map entry's age (a second staleness rule to keep coherent with the disk one), the fix removes the in-memory layer entirely: the fingerprint-keyed shared dir is now the only cache, and every call runs cleanup before consulting it. The process-global mutex stays as a plain build lock (`Mutex<()>`), since concurrent env preparations within one process share the per-PID staging dir.

Cost of losing the memoization is one `read_dir` of `target/dora-python-wheels` plus one of the `fp-` dir per env preparation — noise next to the `uv` invocations on the same path. Re-verified end-to-end on macOS: warm-path `dora build` is still 0.25s with zero maturin invocations; `cargo test -p dora-core --all-features` (193), clippy `-D warnings`, fmt, and `make qa-fast` all pass.

heyong4725:
Re-reviewed the updated PR. The follow-up commit resolves my previous finding: the wheel path is no longer memoized in memory, cleanup runs before disk-cache reuse, and the process-global mutex now only serializes same-process builds that share the per-PID staging dir.

I did not find any remaining blocking issues.

Validation run during re-review:
- cargo test -p dora-core --features build
- cargo clippy -p dora-core --features build -- -D warnings

CI still had several pending jobs at the time of review.

PR #2146, opened 2026-06-10
fix(cli): size receiver queues from recorded counts so fast replay is lossless (#2144)

### LINKED ISSUE #2144: Nightly cluster-record-replay: cross-daemon replay drops messages (80/100 delivered) — Zenoh data-plane race
### Summary

The nightly `Cluster record/replay end-to-end` job failed on 2026-06-10 with **cross-daemon message loss during replay**: the sink received only 80 of 100 replayed values, diverging from the seed(42) baseline mid-stream. Split out of #2089, where the bot appended it (unrelated to the macOS cold-start flake tracked there).

Run: https://github.com/dora-rs/dora/actions/runs/27260677135 (commit a7b89691)

### Evidence

```
07:51:51 stdout  rust-status-node:  dora-replay-node[rust-status-node]: replayed 100 messages
07:51:51 stdout  rust-node:         dora-replay-node[rust-node]: replayed 100 messages
07:51:51 WARN zenoh::net::runtime::orchestrator: Unable to connect to any locator of scouted peer e962f64a...: [tcp/[fe80::...]:41319, tcp/10.1.0.191:41319]
07:51:51 WARN zenoh::net::runtime::orchestrator: Unable to connect to any locator of scouted peer 25447322...: [tcp/[fe80::...]:38635, tcp/10.1.0.191:38635]
=== validate replayed state against seed(42) baseline ===
ERROR: replayed sequence (80 values) does not match the seed(42) baseline (100 values)
  first divergence at index 48: replayed=0x1ecd87584fa9c9a8 baseline=0xa945b539f0fa8a3a
  FAIL: cluster-record-replay
```

Both replay publishers report a full 100 messages sent; the sink received 80. The first 48 values match the baseline, then diverge — consistent with ~20 messages dropped mid-stream on the cross-daemon path (a drop window shifts every subsequent index). The Zenoh scouted-peer connect warnings land right in the replay window, pointing at data-plane connectivity churn rather than recorder corruption.

### Why this looks like a Zenoh data-plane race, not a recorder bug

- The single-machine `Record/replay round-trip` job **passed** in the same run — recording and replay logic are fine in-process; only the cross-daemon transport loses data.
- It is flaky: the same job passed on the [06-09 scheduled run](https://github.com/dora-rs/dora/actions/runs/27190356408).
- The failure family matches #2085 (two-phase restart fix, merged as bf2adb85, included in this run): there, dataflow *restart* republished into a Zenoh data plane whose remote subscribers weren't re-established yet. Replay across daemons appears to have its own version of the same establishment/els connectivity window — replay publishers start emitting as soon as nodes spawn, with no barrier confirming remote subscriber connectivity.

### Suggested direction

Apply the #2085 pattern to the replay path: gate replay publishing on cross-daemon subscriber readiness (or require data-plane establishment acknowledgment before the replay nodes begin emitting). Alternatively/additionally, investigate why an established Zenoh session dropped mid-stream (the scouted-peer warnings suggest a peer connection was lost and re-scouted during replay).

### Repro

```
scripts/qa/ci-nightly-jobs.sh cluster-record-replay
```

(Linux; needs `openssh-server`. Flaky — expect multiple runs to hit it.)

heyong4725:
## Summary

Fixes #2144 — the nightly `cluster-record-replay` job flaked with the replayed sink receiving 80/100 values, diverging mid-stream.

**Root cause** (revised from the issue's initial hypothesis): the failing phase is the *local* replay (`dora replay --speed 0`), not the cross-daemon recording. At full speed, the replay node outpaces the sink, and dora's real-time defaults are lossy under pressure — each input queue holds `DEFAULT_QUEUE_SIZE = 10` messages with drop-oldest, and the node event channel is sized from the queue sizes. On a slow CI runner the sink falls behind and messages are silently dropped. The Zenoh "unable to connect to scouted peer" warnings in the failure log were red herrings — they appear identically in healthy local runs (stale multicast scouting advertisements). The issue's proposed "gate first publish on subscriber readiness" already exists in the node API (`ZENOH_FIRST_PUBLISH_MATCH_TIMEOUT`), which is consistent with the first 48 values arriving intact.

**Reproduced deterministically on macOS** under CPU pressure (12 `yes` burners): 5/6 replay runs diverged (41–94 of 100 values) — the exact nightly signature.

## Fix

`dora replay` already scans the recording to discover replayable nodes; it now also counts messages per `(node, output)`. For every input fed by a replayed node, the descriptor rewrite sets:

- `queue_size` = recorded message count (floored at the default), so one full pass of the recording fits in the receiver queue and the event channel sized from it, and
- `queue_policy: backpressure`, so any residual overflow is logged loudly at the backpressure hard cap instead of dropped silently.

All four input shapes are covered: node-level `inputs:`, legacy `custom.inputs`, single `operator.inputs`, and per-entry `operators[].inputs`.

## Scope decisions (documented on the function)

- **Explicit user `queue_size` is never overridden** — an explicit size encodes freshness semantics (e.g. `queue_size: 1` latest-only control consumers) that replay should reproduce, not defeat.
- **Direct inputs only** — with a partial `--replace`, live intermediate nodes can re-emit at full speed into their own consumers; `dora replay` now prints a warning for partial `--replace` at `--speed 0`.
- **One-pass bound** — `--loop` can still overflow; with the backpressure policy that now fails loudly (ERROR log) instead of silently.
- **Zenoh-level congestion drop** (`CongestionControl::Drop` on publishers) for *multi-daemon* replay is out of scope here; single-machine replay — what the failing job runs — is fully covered.

## Verification

A/B on macOS (`examples/rust-dataflow`, fresh recording, `--speed 0`, validated against the committed seed(42) baseline):

| Condition | Result |
|---|---|
| Pre-fix, 12 CPU burners | **5/6 runs DIVERGED** (41, 83, 88, 94, 68 of 100 values) |
| Post-fix, 12 CPU burners | **6/6 MATCH** (×2 rounds) |
| Pre/post-fix, unloaded | 10/10 MATCH (no regression) |

Rewritten descriptor (via `--output-yaml`):

```yaml
inputs:
  message:
    source: rust-status-node/status
    queue_size: 100
    queue_policy: backpressure
```

## Test plan

- [x] 8 new unit tests on the rewrite: string/map input forms, explicit-queue_size preservation, explicit-policy preservation, default floor, timer/live inputs untouched, replaced nodes skipped, operator/custom/operators[] coverage (`cargo test -p dora-cli` — 193 pass)
- [x] clippy `-D warnings`, fmt, `make qa-fast`, `cargo check --examples` green
- [x] Local record→replay round-trip + stress A/B above
- [ ] Nightly `cluster-record-replay` green after merge (the real confirmation, tracked in #2144)

### DIFF
--- a/binaries/cli/src/command/replay.rs
+++ b/binaries/cli/src/command/replay.rs
@@ -81,16 +81,18 @@ fn run_replay(args: Replay) -> eyre::Result<()> {
     let mut descriptor: serde_yaml::Value =
         serde_yaml::from_str(descriptor_yaml).wrap_err("failed to parse descriptor YAML")?;
 
-    // Discover which nodes produced recorded data
-    let mut recorded_nodes: BTreeMap<String, BTreeSet<String>> = BTreeMap::new();
+    // Discover which nodes produced recorded data and how many messages each
+    // output carries (used to size receiver queues below).
+    let mut recorded_counts: BTreeMap<String, BTreeMap<String, u64>> = BTreeMap::new();
     while let Some(entry) = reader.next_entry()? {
-        recorded_nodes
+        *recorded_counts
             .entry(entry.node_id)
             .or_default()
-            .insert(entry.output_id);
+            .entry(entry.output_id)
+            .or_default() += 1;
     }
 
-    if recorded_nodes.is_empty() {
+    if recorded_counts.is_empty() {
         bail!(
             "recording `{}` contains no messages\n\n  \
              hint: the recording may be empty or corrupted. \
@@ -101,14 +103,14 @@ fn run_replay(args: Replay) -> eyre::Result<()> {
 
     // Determine which nodes to replace
     let nodes_to_replace: BTreeSet<String> = if args.replace.is_empty() {
-        recorded_nodes.keys().cloned().collect()
+        recorded_counts.keys().cloned().collect()
     } else {
         let requested: BTreeSet<String> = args.replace.into_iter().collect();
         for name in &requested {
-            if !recorded_nodes.contains_key(name) {
+            if !recorded_counts.contains_key(name) {
                 bail!(
                     "node `{name}` not found in recording. Recorded nodes: {}",
-                    recorded_nodes
+                    recorded_counts
                         .keys()
                         .cloned()
                         .collect::<Vec<_>>()
@@ -118,6 +120,13 @@ fn run_replay(args: Replay) -> eyre::Result<()> {
         }
         requested
     };
+
+    if args.speed == 0.0 && nodes_to_replace.len() < recorded_counts.len() {
+        eprintln!(
+            "warning: partial --replace with --speed 0: live intermediate nodes may re-emit \
+             faster than their downstream consumers' default input queues can absorb"
+        );
+    }
 
     // Find replay node binary
     let replay_node_bin = find_replay_node_binary()?;
@@ -195,6 +204,8 @@ fn run_replay(args: Replay) -> eyre::Result<()> {
         }
     }
 
+    raise_replayed_input_queue_sizes(nodes, &nodes_to_replace, &recorded_counts);
+
     let modified_yaml =
         serde_yaml::to_string(&descriptor).wrap_err("failed to serialize modified descriptor")?;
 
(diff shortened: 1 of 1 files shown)


heyong4725:
Reviewed PR 2146. I did not find any blocking issues.

The rewrite covers direct inputs from replayed nodes across the supported descriptor shapes, preserves explicit queue sizing/policy, and the runtime-operator `node/operator/output` form should still match because the recorded output id uses the `operator/output` suffix.

Validation run during review:
- cargo test -p dora-cli replay
- cargo test -p dora-cli
- cargo clippy -p dora-cli -- -D warnings

CI still had several pending jobs at the time of review.

PR #2112, opened 2026-06-10
fix(cli): propagate errors instead of panicking; dedup workspace-path resolution in templates

phil-opp:
## Issue

Two avoidable panic paths in user-facing CLI code, plus a 4× code duplication:

1. **`dora new` C/C++ templates**: `workspace_dir.to_str().unwrap()` panicked if the dora checkout path contains non-UTF-8 bytes. The unwrap appeared in four identical copies of the manifest→workspace path-resolution block (`create_cmakefile` and `create_node_cmakefile` in both `template/c/mod.rs` and `template/cxx/mod.rs`).
2. **`send_control_request`**: `serde_json::to_vec(request).unwrap()` panicked instead of returning `Err` from a function that already returns `eyre::Result` and is the funnel for every CLI→coordinator request.

## Fix

- Extracted the duplicated path-resolution block into a single `template::workspace_dir()` helper that returns a proper error (`"dora workspace path is not valid UTF-8"`) instead of panicking. Net −6 lines despite the added doc comment.
- Serialization errors in `send_control_request` are now propagated with context instead of panicking.

Both changes also reduce the production unwrap count tracked by the CI unwrap-budget ratchet. Found during a codebase audit (Correctness + Simplification categories).

## Validation

- `cargo test -p dora-cli` passes
- `cargo clippy -p dora-cli -- -D warnings`, `cargo fmt --check`
- `cargo check --examples` (gates `dora_cli` API surface, #1680)

https://claude.ai/code/session_01L1wL2zpWyF7xvURnsUBSNw

---
_Generated by [Claude Code](https://claude.ai/code/session_01L1wL2zpWyF7xvURnsUBSNw)_

PR #2111, opened 2026-06-10
perf(daemon): hoist Arc allocations out of event broadcast loops

phil-opp:
## Issue

Three broadcast loops in `Daemon::handle_dora_event` re-allocated identical data once per subscriber instead of sharing it:

- **`DoraEvent::Timer`**: `Arc::new(metadata.clone())` inside the subscriber loop — a full `Metadata` clone + `Arc` allocation per timer subscriber, on every tick.
- **`DoraEvent::Logs`**: same per-subscriber `Arc::new(metadata.clone())`, plus `Arc::new(message.clone())` which copies the whole message payload per subscriber.
- **`DoraEvent::LogBroadcast`**: `Arc::new(DataMessage::Vec(sample.clone()))` copied the entire serialized log sample per subscriber — despite the comment directly above saying "Convert to Arrow once, share the sample across subscribers".

The codebase already uses the correct pattern elsewhere (e.g. the output-dispatch loop around line 4730): wrap in `Arc` once before the loop, bump the refcount inside.

## Fix

Hoist the `Arc::new(...)` out of all three loops. In `Timer`/`Logs` the event owns `metadata`/`message`, so the values are moved into the `Arc` — the original clone is eliminated entirely, not just amortized. In `LogBroadcast`, the per-subscriber `Metadata` is left as-is (it carries a fresh timestamp per subscriber by design); only the payload sample is shared.

No consumer relies on `Arc` uniqueness (`get_mut`/`try_unwrap`/`into_inner` are not used on these in the node API, daemon, or runtime), so sharing one allocation is observationally equivalent. Found during a codebase audit (Optimization category).

## Validation

- `cargo test -p dora-daemon` — 56 tests pass
- `cargo clippy -p dora-daemon -- -D warnings`, `cargo fmt --check`

https://claude.ai/code/session_01L1wL2zpWyF7xvURnsUBSNw

---
_Generated by [Claude Code](https://claude.ai/code/session_01L1wL2zpWyF7xvURnsUBSNw)_

PR #2147, opened 2026-06-10
fix(message): reject negative/overflowing byte sizes instead of silently saturating

phil-opp:

## Issue

`ByteSize` parses the user-facing `shared_memory_pool_size` dataflow YAML field. Its `FromStr` cast the parsed `f64` straight to `usize`:

```rust
Ok(ByteSize((num * multiplier as f64) as usize))
```

That cast **saturates** instead of erroring, so:

- `shared_memory_pool_size: "-1KB"` was silently accepted as a **zero-byte pool** (negative f64 → usize saturates to 0) — inconsistent with the integer deserialize path, whose `visit_i64` already rejects `-1` with "byte size cannot be negative";
- absurdly large values silently saturated to `usize::MAX`;
- integer values above 2^53 (e.g. `"9007199254740993B"`) lost precision through the f64 round-trip;
- `visit_u64`/`visit_i64` used `as usize` casts, which truncate on 32-bit targets.

dora-core's `parse_byte_size` (`libraries/core/src/descriptor/validate.rs`) was already hardened against exactly these pitfalls — and even carries a comment documenting the saturation hazard — but the dora-message twin never got the same treatment.

## Fix

Mirrors the dora-core algorithm:

- try an exact integer parse first (no f64 rounding above 2^53), with `checked_mul` overflow detection;
- reject negative and non-finite floats up front;
- reject float results `>= usize::MAX as f64` — `>=` rather than `>` because `usize::MAX as f64` rounds **up** to 2^64 and no f64 values exist between `usize::MAX` and 2^64, so `>=` rejects exactly the results that exceed `usize::MAX` (a `>` check lets a result of exactly 2^64 saturate silently; there's a regression test for that 1-ULP edge);
- `visit_u64`/`visit_i64` now use `usize::try_from` instead of truncating casts;
- the `ByteSize` docs gain a compile-checked doctest.

Note: the two parsers can't trivially share code — the dependency direction is dora-core → dora-message, so consolidating would mean moving `parse_byte_size` into dora-message (reasonable follow-up, out of scope here).

## Validation

- TDD: new tests (`byte_size_rejects_negative`, `byte_size_rejects_overflow`, `byte_size_integer_values_parse_exactly`, `byte_size_float_path_still_works`) were written first and confirmed failing against the old implementation, then made green.
- `cargo test -p dora-message` — 114 unit + 5 integration + 1 doctest, all green.
- `cargo clippy -p dora-message -- -D warnings`, `cargo fmt --all -- --check`.
- Grepped examples/, tests/, docs/ — no existing config relies on the previously-lenient inputs.

---

### DIFF
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -470,6 +470,17 @@ pub enum RemoteCommunicationConfig {
 /// string with a unit suffix (`KB`, `MB`, `GB`, case-insensitive).
 ///
 /// Examples: `67108864`, `"64MB"`, `"1 GB"`, `"512kb"`.
+///
+/// Negative, non-finite, and overflowing values are rejected:
+///
+/// ```
+/// use dora_message::config::ByteSize;
+///
+/// assert_eq!("64MB".parse::<ByteSize>().unwrap().as_bytes(), 64 * 1024 * 1024);
+/// assert_eq!("1.5 KB".parse::<ByteSize>().unwrap().as_bytes(), 1536);
+/// assert!("-1KB".parse::<ByteSize>().is_err());
+/// assert!("1TB".parse::<ByteSize>().is_err());
+/// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct ByteSize(pub usize);
 
@@ -493,10 +504,6 @@ impl FromStr for ByteSize {
             }
         };
 
-        let num: f64 = num_part
-            .parse()
-            .map_err(|_| format!("invalid number in byte size: `{num_part}`"))?;
-
         let multiplier: usize = match unit_part.to_uppercase().as_str() {
             "B" => 1,
             "KB" | "K" => 1024,
@@ -505,7 +512,34 @@ impl FromStr for ByteSize {
             other => return Err(format!("unknown byte size unit: `{other}`")),
         };
 
-        Ok(ByteSize((num * multiplier as f64) as usize))
+        // Use integer parse when possible to avoid f64 rounding above 2^53.
+        if let Ok(num) = num_part.parse::<usize>() {
+            return num
+                .checked_mul(multiplier)
+                .map(ByteSize)
+                .ok_or_else(|| format!("byte size `{s}` is too large"));
+        }
+
+        let num: f64 = num_part
+            .parse()
+            .map_err(|_| format!("invalid number in byte size: `{num_part}`"))?;
+
+        // Casting a negative or non-finite f64 to usize saturates (negatives
+        // and NaN to 0, +inf to usize::MAX) instead of erroring, so reject
+        // them up front.
+        if !num.is_finite() || num < 0.0 {
+            return Err(format!(
+                "byte size must be a non-negative, finite number: `{s}`"
+            ));
+        }
+        let bytes = num * multiplier as f64;
+        // `usize::MAX as f64` rounds up to 2^64, and no f64 values exist
+        // between usize::MAX and 2^64, so `>=` rejects exactly the results
+        // that exceed usize::MAX.
+        if bytes >= usize::MAX as f64 {
+            return Err(format!("byte size `{s}` is too large"));
+        }
+        Ok(ByteSize(bytes as usize))
     }
 }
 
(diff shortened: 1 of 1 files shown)


phil-opp:
## Automated review session summary

🤖 Machine-generated summary of this Claude Code review run (in-depth codebase scan of `origin/main` @ 8352ae8, rebased onto 563b5d3).

### PRs opened (3)

**Correctness**
- #2147 — `ByteSize::from_str` silently saturated negative (`"-1KB"` → 0 bytes) and overflowing byte sizes in `shared_memory_pool_size`; hardened to match dora-core's `parse_byte_size`, plus 32-bit-safe integer visitors and a compile-checked doctest (also covers the Documentation category).

**Optimization**
- #2148 — daemon: `send_output_to_local_receivers` memcpy'd the full payload for remote forwarding even when no local receiver shared the `Arc`; `Arc::try_unwrap` moves it out in pure remote topologies.
- #2149 — C++ node API: redundant `ArrayData` clone on every typed input event crossing the FFI boundary.

### Findings rejected (verified, then dropped)

- Daemon `LogBroadcast` per-subscriber payload clone — duplicate of open #2111.
- CLI template `to_str().unwrap()` ×4 and `send_control_request` serialize unwrap — duplicates of open #2112.
- Suspected `data: None` delivered to debug topic watchers when `remote_receivers` is false — not reachable: the coordinator refuses `start_topic_debug_stream` unless `enable_debug_inspection` is set, and that flag forces `remote_receivers = true` in `send_out`.
- `event_as_input` `values().to_vec()` copy (C++ API) — required; the cxx FFI struct owns a `Vec<u8>`.
- "Misleading" `expect()` messages on `NonNull::new(slice.as_ptr())` in the node-api event stream — the panics are unreachable; renaming messages is cosmetic churn.
- Coordinator `spawn_dataflow` cloning `nodes`/descriptor per daemon — cold setup path, and each daemon needs an owned message.
- `close_input` tuple-key clones in the daemon — required by `BTreeMap` `Borrow` semantics for tuple keys.
- `#[allow(dead_code)]` on `OperatorEvent` — cosmetic.
- Session-fingerprint `Some(empty env)` vs `None` asymmetry — speculative; YAML deserializes a missing `env` to `None`.
- `Metadata::open_telemetry_context` empty-string allocation — trivial, not on a hot path.
- `dora-message` compile-time version-parse unwrap — constant input, exercised by every test run.
- ByteSize/`parse_byte_size` duplication across dora-message/dora-core — real, but consolidation requires moving the core parser across the crate boundary (dependency direction is core → message); noted as a possible follow-up instead of a PR.

### Validation

Each branch: affected-crate tests + clippy `-D warnings` + fmt. Combined diff: full workspace suite (110 test binaries green), workspace clippy `-D warnings`, `cargo check --examples`, and review/simplify passes (which caught and fixed a 1-ULP boundary bug in the original overflow check of #2147 — see its PR body).

---

phil-opp (commented):
Reviewed the `ByteSize` parsing hardening against the diff.

- The integer fast path (`num_part.parse::<usize>()` + `checked_mul`) preserves values above 2^53 exactly (e.g. `9007199254740993B`) and reports overflow instead of saturating to `usize::MAX`.
- The float path now rejects negative/non-finite inputs up front (fixing the `-1KB` → 0-byte silent acceptance) and uses `>= usize::MAX as f64`, which correctly catches the 2^64 round-up edge that a `>` check would let saturate.
- `visit_u64`/`visit_i64` switch to `usize::try_from`, removing the 32-bit truncation.

The behavior now matches dora-core's `parse_byte_size`, and the new tests cover negatives, overflow, the exact-integer path, and the float path. The note that the two parsers can't trivially share code (dora-core → dora-message dependency direction) is accurate.

No issues found.

---

---

phil-opp:

Reviewed the `ByteSize` parsing rewrite against the diff:

- The integer fast path (`num_part.parse::<usize>()` + `checked_mul`) preserves exact values above 2^53 and reports overflow instead of saturating — the regression test for `9007199254740993B` confirms it.
- The f64 path rejects negative and non-finite inputs *before* the `as usize` cast (which would otherwise saturate to 0 / `usize::MAX`), matching the integer `visit_i64` behaviour.
- The `bytes >= usize::MAX as f64` boundary is the correct comparison: `usize::MAX as f64` rounds up to 2^64 and no f64 values exist in `(usize::MAX, 2^64)`, so `>=` rejects exactly the results that would overflow — and the `18014398509481984.0KB` (= 2^64) test pins the 1-ULP edge a `>` check would have let through.
- `visit_u64`/`visit_i64` switching to `usize::try_from` removes the 32-bit truncation.

Logic mirrors dora-core's hardened `parse_byte_size`. No correctness issues.

---

---

PR #2148, opened 2026-06-10
perf(daemon): move output payload out of Arc instead of copying when uniquely owned

phil-opp:

## Issue

`send_output_to_local_receivers` always memcpy'd the full payload to produce the `data_bytes` returned for remote forwarding (inter-daemon Zenoh publish / debug inspection):

```rust
data.as_ref().map(|arc| {
    let DataMessage::Vec(v) = arc.as_ref();
    v.clone()   // O(payload) copy on every remotely-forwarded output
})
```

But after the local fan-out loop, the `Arc` is only shared if a local receiver's channel actually took a clone. In a **pure remote topology** — every cross-daemon output, i.e. the common case in multi-machine deployments — the refcount is 1 and the copy is avoidable. For robotics payloads (images, point clouds) that's a multi-MB memcpy per message.

## Fix

`Arc::try_unwrap` moves the payload out when uniquely owned and falls back to the previous clone only when a local receiver shares the allocation. Behavior is unchanged; an in-code comment documents why the uniqueness condition holds relative to the fan-out loop above.

## Validation

- New test `data_bytes_returned_with_and_without_local_receivers` locks the `data_bytes` contract for both paths (unique ownership → moved; shared with a local receiver → cloned, and the receiver still gets its event).
- `cargo test -p dora-daemon` green (57 unit tests incl. the new one), `cargo clippy -p dora-daemon -- -D warnings`, `cargo fmt --all -- --check`.
- All four callers checked: the two `need_data_bytes: false` callers and the bench fixture are unaffected; `send_out` (the only `true` caller) consumes the return value exactly as before.

Related: complements #2111, which hoists `Arc` allocations out of the daemon's broadcast loops — this PR addresses the remaining copy on the remote-forwarding side of the same hot path (no code overlap).

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4828,9 +4828,15 @@ async fn send_output_to_local_receivers(
         dataflow.subscribe_channels.remove(id);
     }
     let data_bytes = if need_data_bytes {
-        data.as_ref().map(|arc| {
-            let DataMessage::Vec(v) = arc.as_ref();
-            v.clone()
+        // If no local receiver kept a reference (pure remote topology), the
+        // Arc is uniquely owned and the payload can be moved out instead of
+        // copied.
+        data.map(|arc| match Arc::try_unwrap(arc) {
+            Ok(DataMessage::Vec(v)) => v,
+            Err(arc) => {
+                let DataMessage::Vec(v) = arc.as_ref();
+                v.clone()
+            }
         })
     } else {
         None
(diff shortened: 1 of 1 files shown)


phil-opp (commented):
Reviewed the `Arc::try_unwrap` change in `send_output_to_local_receivers` against the surrounding function.

`data` is wrapped in a single `Arc` before the fan-out loop (`let data = data.map(Arc::new)`) and cloned into local receivers' channels via `try_send`. After the loop the refcount is >1 only when a local receiver actually kept a clone; a `Full`/`Closed` send drops its clone, so a pure-remote topology (or an all-dropped fan-out) leaves the `Arc` uniquely owned. `try_unwrap` therefore moves the payload out exactly in the case the old code memcpy'd it, and falls back to the previous clone when a receiver shares the allocation — the returned bytes are identical either way. Consuming `data` here is fine since it has no later use. The new `data_bytes_returned_with_and_without_local_receivers` test pins both paths.

No issues found.

---

---

phil-opp:

The `Arc::try_unwrap` change is sound: `data_bytes` is the same byte content on both branches. When the `Arc` is uniquely owned (pure remote topology, no local receiver took a clone) the payload is moved out instead of memcpy'd; when a local receiver holds a clone, `try_unwrap` returns `Err` and the previous clone behaviour is preserved. `data` is consumed by `.map(...)`, so the compiler guarantees no use-after-move, and `try_unwrap` is always safe regardless of how many holders exist — the optimization can never produce wrong bytes. The new test locks both the moved and cloned paths (and confirms the local receiver still gets its event).

---

---

PR #2149, opened 2026-06-10
perf(cxx-node): drop redundant ArrayData clone in arrow input export

phil-opp:

## Issue

`event_as_arrow_input` in the C++ node API cloned the `ArrayData` just to pass a reference to `arrow::ffi::to_ffi`, which only borrows its argument:

```rust
let array_data = data.to_data();
match arrow::ffi::to_ffi(&array_data.clone()) {
```

The clone duplicates the buffer/child-data vecs and bumps every buffer `Arc` on **every typed input event** crossing the C++ FFI boundary — pure waste, since `array_data` is dropped right after the call either way.

## Fix

One-line change: pass `&array_data` directly.

## Validation

- `cargo test -p dora-node-api-cxx`, `cargo clippy -p dora-node-api-cxx -- -D warnings`, `cargo fmt --all -- --check` — green.
- Verified `array_data` has no use after the `to_ffi` call, and that the sibling `event_as_arrow_input_with_info` path doesn't have the same redundancy.

---

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -536,7 +536,7 @@ unsafe fn event_as_arrow_input(
 
     let array_data = data.to_data();
 
-    match arrow::ffi::to_ffi(&array_data.clone()) {
+    match arrow::ffi::to_ffi(&array_data) {
         Ok((ffi_array, ffi_schema)) => {
             unsafe {
                 std::ptr::write(out_array, ffi_array);


phil-opp (commented):
Reviewed the one-line change against the diff directly.

`arrow::ffi::to_ffi` takes `&ArrayData` and only borrows its argument, and `array_data` is dropped immediately after the `match` with no later use, so passing `&array_data` instead of `&array_data.clone()` is behavior-preserving. It removes a per-event duplication of the buffer/child-data vecs plus an `Arc` bump on every typed input event crossing the C++ FFI boundary. The sibling `event_as_arrow_input_with_info` path doesn't share this redundancy.

No issues found.

---

---

phil-opp:

`arrow::ffi::to_ffi` takes `&ArrayData` and only borrows it, so the `.clone()` produced a throwaway deep copy of the buffer/child-data vectors (bumping every buffer `Arc`) on each typed input event for no reason. `array_data` is not used after the `to_ffi` call, so passing `&array_data` directly is behaviorally identical and removes the per-event copy. The sibling `event_as_arrow_input_with_info` path doesn't have the same pattern. Correct change.

---

---

Issue #1872, opened 2026-05-19
Opt-in pinned-host memory pool for high-throughput CPU→GPU transfers


heyong4725:
## Context

dora today supports two zero-copy transports between nodes:

| Path | Memory | Pinned? | GPU-direct? |
|---|---|---|---|
| Zenoh SHM (≥4 KiB messages) | Host shared memory, page-aligned | **No** | No — requires `cudaMemcpyHostToDevice` from non-pinned memory |
| cudaIpc handles | GPU memory | N/A | Yes, but only useful when both producer and consumer are on GPU |

There's a gap: **CPU-produced data destined for GPU consumption**. Camera frames, sensor batches, real-time perception inputs — common dora workloads where the producer is a CPU node (driver, decoder, preprocessor) and the consumer is a CUDA-using node.

Today the consumer has to either:
- Allocate its own pinned host buffer, copy from zenoh-SHM into it, then `cudaMemcpyHostToDevice` (two copies)
- Skip pinning and accept the slower non-pinned `cudaMemcpyHostToDevice` from zenoh-SHM (one copy, slower)
- Implement custom pinned-memory infrastructure outside dora (complex, not portable across users)

None of these are great. A daemon-brokered opt-in pinned-host-buffer pool would let producers write directly into pinned memory that the consumer can DMA from. The expected throughput improvement on this specific CPU→GPU path is 3-5× over the non-pinned baseline (claim from PR #1623; needs independent measurement).

## What this issue is, and isn't

**This issue documents the gap and invites proposals.** It does NOT commit to an architecture. dora has historically moved AWAY from custom shared-memory infrastructure (see #1745 which removed the previous SHM cache in favor of zenoh-SHM). Any new transport has to clear a high bar.

PR #1623 attempted a 2,717-line implementation of one possible architecture (daemon-brokered `/dev/shm/dora_pool_*` ring buffers with custom protocol). It was closed because the implementation was 7× larger than the feature needed and the design framing was a fork-for-paper not a focused proposal. The underlying gap it tried to address is real; this issue keeps that gap visible for a future, leaner attempt.

## What a reviewable proposal would address

Before writing code, an issue/RFC should answer:

1. **Why a new transport, not extending an existing one?**
   - Can zenoh-shm be extended to support optionally-pinned host memory? Zenoh has a SHM provider interface; could a `PinnedHostShmProviderBackend` plug in cleanly?
[...]

PR #2151, opened 2026-06-11
add pinned memory-pool transport for repeated CPU/CUDA tensor transfer

tang-canran:
add an opt-in memory-pool transport path for large, fixed-shape tensor transfer between CPU and CUDA nodes(updates #1872 )

Architecture introduced relative to upstream dora:
- add a dedicated pinned-memory transport path for repeated high-throughput
  transfer of large, fixed-shape tensors between CPU and CUDA nodes using
  page-locked host memory plus DMA-based heterogeneous transfer
- add a dedicated `libraries/extensions/memory-pool` crate for daemon-side pool
  metadata tracking, lifecycle management, and shutdown cleanup
- add three additive daemon request/reply operations for the memory-pool side
  channel: `RegisterPinnedMemory`, `ReadPinnedMemory`, and `FreePinnedMemory`
- add a reusable shared-memory pool identified by `memory_pool_id`, so the
  sender can register once and overwrite many times instead of re-sending the
  full tensor through the normal message path on every iteration
- add a node-side fast path that reads pool metadata directly from the
  shared-memory header and only falls back to daemon metadata lookup when needed,
  reducing repeated daemon query overhead
- add four Python APIs for the feature:
  `register_memory_pool`, `write_memory_pool`, `read_memory_pool`, and
  `free_memory_pool`
- add an `examples/memory-pool` example set to validate pool transfer throughput
  and exercise the fault-tolerance paths

Performance notes:
- verified CPU -> CUDA repeated transfer throughput remains around 4.5 GB/s in
  the current local environment
- verified CUDA -> CPU repeated transfer throughput remains around 7.6 GB/s in
  the current local environment


phil-opp:
**Automated review by Claude — I found several issues.** I reviewed the diff directly (not the PR description), focusing on the unsafe shared-memory / CUDA paths in `apis/python/node/src/lib.rs`, the daemon side, and the new `dora-memory-pool` crate.

### 1. (High) `write_memory_pool` does an unbounded `copy_nonoverlapping` into the pool → out-of-bounds write

The pool's shared memory is allocated **once** in `register_memory_pool` with `total_size = data_offset + size`, where `size` is the *registration-time* tensor size. `write_memory_pool` then copies `size` bytes — taken from the **caller-supplied `tensor_info` on the write call** — to `shmem_ptr.add(data_offset)`:

```rust
std::ptr::copy_nonoverlapping(ptr_val as *const u8, shmem_ptr.add(data_offset), size);
```

Nothing checks that `data_offset + size <= pool capacity` (neither `slot_data.size` on the cache-hit path nor `shmem.len()` on the open path). And unlike `register_memory_pool` (which rejects `size == 0 || size > 1 GiB`), the write path has **no size validation at all**. So a write whose `size` exceeds the originally-registered size silently writes past the end of the shared-memory mapping — memory corruption / SIGSEGV. The CUDA `_cuda_memcpy` and `dma_copy` branches have the same unbounded-size property. This should validate `size` against the pool's real capacity (and reject 0 / oversize) on every write before copying.

### 2. (Medium) Slow-path `read_memory_pool` returns the *sender's* process-local pointer

The daemon stores `ptr` = the registering process's source tensor pointer. The slow path (`read_pinned_memory` fallback) builds a `tensor_info` with that `ptr` and `device="cpu"` and no shmem mapping. In a different receiver process that pointer is meaningless — dereferencing it reads a bogus address. Only the fast (`try_doradma_read`) path, which maps the shmem and returns a shmem-relative pointer, is actually usable cross-process. The daemon fallback is effectively unsafe/broken for the cross-process case it exists to serve.

### 3. (Medium) Pool resources are never reclaimed until daemon exit; several caches grow unboundedly

- `free_shared_memory` returns `Ok(())` without unlinking for any `dora_pool_*` name, deferring all cleanup to `cleanup_all` on daemon shutdown. So `free_memory_pool` during a running dataflow frees nothing on disk — the `/dev/shm` file (and the exported GPU IPC buffer) persist for the daemon's lifetime. For the advertised "register once, write many, free, re-register" pattern this leaks one shmem file + GPU buffer per cycle.
- The daemon `memory_pool` table is daemon-global and is only emptied on daemon exit — entries from a stopped dataflow are never removed, and there's no per-dataflow isolation (any node can `read`/`free` any pool id; `registered_by` is recorded but never checked).
- Receiver caches `RECV_GPU_VA` / `RECV_CPU_SHMEM` and `FREED_POOL_IDS` are keyed by the monotonic counter and never pruned, so they grow by one entry per pool for the life of the process. The comment on `FREED_POOL_IDS` ("naturally bounded") is inaccurate — it grows without bound.

### 4. (Low / structural)
- In `register_memory_pool`, the `PINNED_POOL.remove(&pool_counter)` reuse branch is dead code: `pool_counter` is freshly incremented on every call, so it can never match an existing slot.
- `#[repr(u8)]` was added to `DaemonRequest`. The enum is serialized via serde (repr doesn't affect the wire format) and isn't FFI'd, so the purpose is unclear — worth removing or explaining.
- The only added Rust test (`test_shmem_creation`) exercises `ShmemConf`, not any of the register/write/read/free transfer logic, and CUDA paths aren't covered in CI. Given the volume of new `unsafe` + `unsafe impl Send/Sync` on raw-pointer structs, this is a large untested surface.

The core idea (a reusable pinned-memory pool with a self-describing shmem header) is reasonable, but issue #1 is a memory-safety bug that should be fixed before this lands, and #2–#3 affect correctness/robustness of the documented fallback and lifecycle.

---

---

phil-opp:
**🤖 Automated review by Claude (Claude Code)** — *This is a fully automated review with no human in the loop. Findings are advisory; please verify before acting.*

I re-reviewed against the current diff. The latest commits do fix the earlier out-of-bounds `write_memory_pool` write and the slow-path `read_memory_pool` pointer bug — those look resolved. A few things still stand out:

**1. Leftover debug cruft — will fail CI clippy `-D warnings` (blocking).**
`apis/python/operator/src/lib.rs`, in `pydict_to_metadata`:
```rust
// 写入紧急调试文件
...
if key_str == "shared_memory_name" {
    // 尝试提取字符串
    match value.extract::<String>() {
        Ok(s) => (),
        Err(e) => (),
    }
}
```
This is a no-op block left over from debugging. `Ok(s)` / `Err(e)` bind unused variables, which trips `unused_variables`, so `cargo clippy --all -- -D warnings` (a hard PR gate) will reject it. The leftover comments should be removed too.

**2. Unrelated operator behavior change.**
In the same file, `PyEvent::metadata` now returns an empty `PyDict` for non-INPUT events instead of `None`. That changes the shared operator API surface and is unrelated to the memory-pool feature — worth justifying or splitting into its own PR so it gets reviewed on its own merits.

**3. Sender-side pool leak on `free_memory_pool`.**
In `apis/python/node/src/lib.rs`, `free_memory_pool` clears the receiver caches (`RECV_GPU_VA` / `RECV_CPU_SHMEM`) and the daemon entry, but never removes the entry from `PINNED_POOL`. The registering process therefore keeps the `Shmem` mapping and the `cudaHostRegister` pinned host region alive for the process lifetime, even though the daemon unlinks the `/dev/shm` file. A register → write → free → re-register loop leaks one mapping + one pinned host region per cycle.

**4. Slow-path `read_memory_pool` reports `device="cpu"` for CUDA pools.**
The daemon-fallback branch hardcodes `device="cpu"` regardless of `ipc_present` / pinned type. If the fast path ever returns `None` for a CUDA pool, the consumer wraps host memory as a CPU tensor — a silent device mismatch (not UB, but wrong).

**5. No ownership/isolation on the daemon pool table.**
`MemoryPoolManager` is daemon-global and keyed by `pool_{counter}` from a per-process counter; `registered_by` is recorded but never enforced, and entries are only reclaimed on explicit free or daemon shutdown (not on dataflow stop). Two nodes/dataflows can collide on the same id, and any node can `read`/`free` another node's pool.

I'd treat **#1 as a hard blocker** (CI), **#2–#3 as worth fixing before merge**, and **#4–#5 as correctness/robustness concerns** on the fallback and lifecycle paths.

---

heyong4725:
Thanks for the contribution! This PR now has merge conflicts with `main` (a few daemon/CLI changes landed this week). Could you rebase your branch onto the latest `main`? Tip: since this PR is from your fork's `main` branch, consider moving it to a feature branch (e.g. `feat/pinned-memory-pool`) — that makes future updates easier. Once it's conflict-free, CI can run and we can give it a proper review.

phil-opp:
Reviewed the diff directly (focusing on the daemon-side lifecycle and the new `dora-memory-pool` crate). The core idea — register-once / overwrite-many to skip re-sending fixed-shape tensors — is reasonable, but I found several issues I think are important before this can land.

**1. `cleanup_all` deletes shared-memory files host-wide, not scoped to this daemon/dataflow.** In `libraries/extensions/memory-pool/src/lib.rs`, after removing tracked entries `cleanup_all` scans `/dev/shm` and removes *every* file whose name starts with `dora_pool_`:

```rust
if name.starts_with("dora_pool_") {
    if let Err(err) = std::fs::remove_file(entry.path()) { ... }
}
```

dora explicitly supports multiple daemons per host and runs many dataflows over a machine's lifetime. Since the prefix is shared across all of them, one daemon shutting down will unlink the pool segments belonging to another *live* daemon/dataflow, corrupting their in-flight pools. The cleanup needs to be scoped (e.g. only segments this manager actually registered, or a per-daemon/per-dataflow prefix), not a global glob.

**2. Path traversal in `free_shared_memory`.** The name is interpolated straight into a filesystem path and unlinked:

```rust
let shm_path = format!("/dev/shm/{}", shm_name);
std::fs::remove_file(&shm_path)
```

`shm_name` comes from the node-supplied `shared_memory_name` field in `RegisterPinnedMemory`. A node can register a pool with `shared_memory_name = "../../<something>"` and then free it, causing the daemon to `remove_file` an arbitrary path the daemon user can write. The name should be validated (reject `/`, `..`, restrict to an expected charset) before it's used in a path.

**3. No access control on `ReadPinnedMemory` / `FreePinnedMemory`.** Both look up purely by string id; `registered_by` is recorded but never checked. Any node in the dataflow can free (or read, or read-with-`free`) a pool another node owns, pulling the segment out from under the owner. Given the example even exercises `read_after_free` / `duplicate_free`, the ownership/lifecycle model here seems under-specified — at minimum a free from a non-owner should be rejected or clearly defined.

**4. No Rust tests for the new crate or the three new daemon protocol ops.** `dora-memory-pool` ships with zero `#[cfg(test)]` coverage, and the `Register/Read/Free` handlers in `binaries/daemon/src/lib.rs` are untested. Per the repo's TDD policy this is a notable gap, and points 1–3 above are exactly the kind of thing a unit test on `cleanup_all`/`free_shared_memory` would pin.

Smaller notes: `MemoryPoolEntry.in_use` is written (`true`) but never read; and in the example the receiver maps the tensor once at `i == 0` and then only reads `torch_tensor.nbytes` per iteration, so the reported throughput isn't actually measuring per-message data transfer — worth tightening if those numbers are meant to back the perf claims.

Also, this is a large, additive architectural change (new transport, new daemon ops, new Python API surface) — it would benefit from a design discussion on the linked issue (#1872) before the protocol is committed to.

---

---

PR #2153, opened 2026-06-11
ci(nightly): raise CLI Tests timeout to 100 min for Windows cache-save overhang

heyong4725:
## Summary

Closes out the Windows residual from #2143. On the [2026-06-11 nightly](https://github.com/dora-rs/dora/actions/runs/27331903367) — the first run with the #2145 wheel cache — `CLI Tests (windows-latest)` **passed every test step** at ~88 min but was killed at the 90-min limit during the post-job rust-cache save. macOS and Linux finished in 72/57 min.

The structural per-step rebuild creep is fixed (the wheel cache works on Windows too: `Test Python Multiple Arrays` took 37 s vs ~10 min pre-fix). The remaining Windows cost is fixed overhead — 23 min CLI build + 19 min venv setup — so a 10-minute bump covers the cache save without masking future creep. The workflow comment documents this and says not to raise further; per-step times are the next lever if it creeps again.

YAML-only change; `timeout-minutes: 90 → 100` on the `cli-tests` job plus an explanatory comment.

## Test plan
- [x] YAML parses
- [ ] Next nightly: `CLI Tests (windows-latest)` green end-to-end

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1206,7 +1206,13 @@ jobs:
       fail-fast: false
     # Windows builds + Python Dynamic Node tests are slower than other
     # platforms; 60 min is not enough when the runner is under load.
-    timeout-minutes: 90
+    # 100 min: on 2026-06-11 (first run with the #2145 wheel cache) Windows
+    # finished all test steps at ~88 min but was killed at 90 during the
+    # post-job rust-cache save. The structural per-step rebuild creep is
+    # fixed (#2143); this covers the remaining fixed cost (23 min CLI build
+    # + 19 min venv setup) plus the cache save. Do not raise further —
+    # investigate per-step times instead (see #2143).
+    timeout-minutes: 100
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff directly. This is a YAML-only change: `timeout-minutes: 90 → 100` on the `cli-tests` job in `.github/workflows/nightly.yml`, plus an explanatory comment. The bump is scoped to the single job whose [2026-06-11 nightly](https://github.com/dora-rs/dora/actions/runs/27331903367) Windows run passed all test steps (~88 min) but was killed during the post-job rust-cache save at the 90-min limit; macOS/Linux finish well under it, so a per-job bump rather than a matrix-wide one is the right scope. No semantic change to any build/test step, and the inline comment documents the reasoning and the "don't raise further — investigate per-step times" guidance, which matches the project's stance on keeping nightly cost bounded.

One thing only the next nightly can confirm (already noted as unchecked in the PR's own test plan): that the cache-save tail actually fits inside the extra 10 minutes rather than the job having a deeper creep. The change itself is low-risk and self-documenting.

---

---

Issue #2152, opened 2026-06-11
Nightly Examples (macOS): node fails to exit after dataflow stop — daemon spins on TCP read timeouts until step timeout


heyong4725:
### Summary

The nightly **Examples (macOS)** job has flaked twice in three days with the same signature in *different* examples: the dataflow completes its work, one node fails to exit after stop, and the daemon logs `TCP read header timed out` every ~30 s until the CI step's 15-minute timeout kills the job. Split out of #2082 (where the bot has been appending these).

| Date | Run | Failed step | Node that hung |
|---|---|---|---|
| 2026-06-09 | [27190356408](https://github.com/dora-rs/dora/actions/runs/27190356408) | Rust Git Dataflow example | (step timeout, same TCP spam) |
| 2026-06-11 | [27331903367](https://github.com/dora-rs/dora/actions/runs/27331903367) | C Dataflow example | `runtime-node` (operator runtime) |

### Evidence (06-11, C Dataflow)

The dataflow finished its work completely — this is a shutdown hang, not a data-path failure:

```
08:20:27 c_node:  [c node] received 10 events / finished successfully
08:20:27 c_sink:  [c sink] received input `counter` ... value is 100
08:20:27 c_sink:  [c sink] received InputClosed event / received stop event / finished successfully
08:20:27 daemon:  c_node finished successfully
08:20:27 daemon:  c_sink finished successfully
-- runtime-node never reports stop; no "dataflow finished" --
08:20:57 WARN dora_daemon::node_communication: failed to receive DaemonRequest
         Caused by: unexpected I/O error ... TCP read header timed out
         Location: binaries/daemon/src/node_communication/tcp.rs:78:26
(repeats every ~30 s until the 15-min step timeout)
```

The only output from `runtime-node` during the run was a startup warning that may or may not be relevant:

```
08:20:22 runtime-node: WARN dora_node_api::event_stream: received misaligned Zenoh payload,
         copying before Arrow decode source="zenoh SHM" address=0x138008069 alignment=64
```

### Reading

- The hanging process differs per occurrence (operator runtime on 06-11; the rust-git example on 06-09), so this looks like a shared shutdown-path race — a node/runtime that received (or should have received) stop fails to exit, the daemon's per-node TCP receive loop times out repeatedly, and nothing escalates to a kill.
- Both occurrences are macOS; Linux/Windows examples jobs were green on the same commits.
- Two independent questions worth separating:
[...]

PR #2159, opened 2026-06-12
fix(daemon): bound and diagnose nodes that block an otherwise-finished dataflow (#2152)

heyong4725:
## Summary

Implements both fixes proposed in #2152 (analysis [here](https://github.com/dora-rs/dora/issues/2152#issuecomment-4686174951)): the nightly macOS `examples` job has flaked twice with a node failing to exit after the dataflow finished its work, hanging the daemon — and CI — until the 15-minute step timeout, with nothing in the logs but TCP read-timeout spam.

**Root cause**: the natural-finish path sends `AllInputsClosed` and waits forever; only explicit stops had a Stop → SIGTERM → SIGKILL escalation ladder. Startup has a watchdog; shutdown had none.

## What this adds

**1. Finish-straggler watchdog** (bounded damage): the daemon records when each node's drain phase starts (`AllInputsClosed` sent). On the existing health-check tick, once *every* running non-dynamic node of a dataflow is draining and a node has been draining longer than the grace period (default 120 s, `DORA_FINISH_DRAIN_GRACE_SECS` to override), it is escalated through the **same `stop_single_node` ladder** explicit stops use (Stop event → 10 s → SIGTERM → +5 s → SIGKILL).

**2. Stack capture** (diagnosability): just before escalation, the daemon captures a stack sample of the stuck process into the logs (macOS `sample` — the platform where #2152 manifests), so the *why didn't it exit* question is answerable from the next CI occurrence instead of requiring a live repro. This subsumes the originally-proposed CI `timeout` wrapper: with the watchdog bounding the hang, a step-level wrapper would never fire — the daemon now produces the diagnostic itself, in CI and locally alike.

## Semantics (deliberate choices)

- **An escalated straggler that has to be killed is a node FAILURE** (`GraceDuration`: "killed by dora because it didn't react to a stop message in time"). Bounded *and* loud — a green run would hide every recurrence of the shutdown bug. A straggler that exits 0 on the escalation's Stop event stays clean (work done, responded to stop, just late).
- **A running source vetoes escalation**: sources never receive `AllInputsClosed`, so a dataflow with a live source is never "otherwise finished". A stuck *source* therefore still hangs (documented limitation; different bug class).
- **Cross-daemon conservatism**: dataflows with open external output mappings skip escalation — a local "straggler" may be flushing to consumers on other daemons this daemon can't see.
- Drain state is cleared on node exit, `RemoveNode`, and `restart_single_node`, so restarted/re-added nodes under the same id get a fresh clock.

## Verification

- 6 new unit tests on the selection logic (gating on non-draining nodes, grace boundary, dynamic nodes, one-shot escalation, multi-straggler); 62 daemon tests pass. Test backdates kept ≤2 s (the `Instant` underflow class from #2088).
- **E2E with a deliberately hanging Python node** (`DORA_FINISH_DRAIN_GRACE_SECS=5`): warn at +7 s naming the straggler, stack sample captured (full `sample` process analysis in the log), SIGTERM at +16 s, run exits **1** with `Node 'hang' failed: node was killed by dora because it didn't react to a stop message in time` — ~20 s bounded instead of forever.
- Control runs (`examples/rust-dataflow`, with and without the change): exit 0, zero escalations.
- `make qa-fast`, clippy `-D warnings`, fmt all green. Adversarially reviewed (3 finder angles, 12 findings); review revisions are the second commit.

## Effect on #2152

Fixes the hang-forever half: the nightly flake becomes a ~2-minute bounded failure with the culprit named and its stack in the log. The underlying node-exit race stays open in #2152, now self-diagnosing — the next occurrence hands us the stuck stack directly.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1203,6 +1203,7 @@ impl Daemon {
                 Event::Dora(event) => self.handle_dora_event(event).await?,
                 Event::DynamicNode(event) => self.handle_dynamic_node_event(event).await?,
                 Event::HeartbeatInterval => {
+                    self.check_finish_stragglers();
                     if let Some(sender) = &self.coordinator_sender {
                         let msg = serde_json::to_vec(&Timestamped {
                             inner: CoordinatorRequest::Event {
@@ -2236,6 +2237,46 @@ impl Daemon {
             }
         };
         Ok(status)
+    }
+
+    /// Watchdog for nodes that block an otherwise-finished dataflow
+    /// (dora-rs/dora#2152).
+    ///
+    /// The natural-finish path sends `AllInputsClosed` and waits for nodes
+    /// to exit voluntarily; unlike explicit stops it had no deadline, so a
+    /// stuck node hung `dora run` (and CI) until an external timeout. Once
+    /// every running node of a dataflow has been draining for longer than
+    /// the grace period, escalate through the same Stop → SIGTERM → SIGKILL
+    /// ladder used by explicit stops — after capturing a stack sample of
+    /// the stuck process so the hang itself stays diagnosable.
+    fn check_finish_stragglers(&mut self) {
+        let grace = finish_drain_grace();
+        for (dataflow_id, dataflow) in self.running.iter_mut() {
+            for node_id in dataflow.finish_stragglers(grace) {
+                let drained_for_secs = dataflow
+                    .all_inputs_closed_at
+                    .get(&node_id)
+                    .map(|since| since.elapsed().as_secs())
+                    .unwrap_or_default();
+                let pid = dataflow
+                    .running_nodes
+                    .get(&node_id)
+                    .and_then(|node| node.pid.as_ref())
+                    .map(|pid| pid.load(atomic::Ordering::Relaxed));
+                tracing::warn!(
+                    "dataflow {dataflow_id} is finished except node `{node_id}` \
+                     (AllInputsClosed sent {drained_for_secs}s ago) — escalating stop \
+                     (set DORA_FINISH_DRAIN_GRACE_SECS to adjust the grace period)"
+                );
+                spawn_stack_sample_capture(node_id.clone(), pid);
+                dataflow.finish_escalated.insert(node_id.clone());
+                if let Err(err) = dataflow.stop_single_node(&node_id, &self.clock, None) {
+                    tracing::warn!(
+                        "failed to escalate stop for finish straggler `{node_id}`: {err:?}"
+                    );
+                }
+            }
+        }
     }
 
     fn check_node_health(&self) {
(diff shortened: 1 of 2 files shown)


heyong4725:
Found one issue that should be fixed before merging.

- binaries/daemon/src/lib.rs:3823: finish-straggler exits are still reported to the coordinator as clean stopped nodes. The watchdog path calls stop_single_node, which disables restart on the node. Later, handle_node_stop_inner computes clean_stop as exit_clean || restarts_disabled and sends that in DaemonEvent::NodeStopped. For a straggler killed by the watchdog, node_result is correctly Err(GraceDuration) above, but restarts_disabled is also true, so the coordinator marks the node as NodeStatus::Stopped instead of Failed and arms the auto-expire timer. That contradicts the PR semantics that a force-killed natural-finish straggler is a loud node failure and hides it from dora doctor / node list after the stopped-row grace. The clean_stop computation needs to account for finish_escalated, e.g. clean only if exit_clean || (restarts_disabled && !finish_escalated).

Validation run during review:
- cargo test -p dora-daemon --lib finish_straggler (0 matched; no-op filter)
- cargo test -p dora-daemon --lib

PR CI was green at the time of review.

heyong4725:
Confirmed and fixed in ecff5a5c. The watchdog path does set `disable_restart` via `stop_single_node`, and `clean_stop = exit_clean || restarts_disabled` then reported the force-killed straggler to the coordinator as a clean `Stopped` (auto-expiring) — contradicting the failure classification one layer up.

Fix: `handle_node_stop_inner` now reads **and consumes** `finish_escalated` (the SpawnedNodeResult-site cleanup no longer clears it, so the flag survives until this read — an escalated node never restarts, so it can't leak into a next incarnation) and computes `clean_stop = exit_clean || (restarts_disabled && !finish_escalated)`. Escalated kills now surface as `NodeStatus::Failed` and stay visible to `dora doctor`; an escalated node that still exits 0 remains a clean stop, consistent with the exit-classification semantics.

Verified with the networked hang fixture (`dora up/build/start`, grace 5s): `dora list` reports the dataflow `Failed` after escalation. One note from that run: per-node rows vanish for *both* statuses once the dataflow itself finishes — `node_metrics` lives under `running_dataflows` — so the Stopped/Failed distinction is observable while the dataflow is alive (multi-node case) and in doctor's bucketing, which is exactly where it matters. 62 daemon tests, clippy `-D warnings`, fmt all green.

heyong4725:
Re-reviewed the updated PR. The new commit resolves my previous finding: finish_escalated is now consumed in handle_node_stop_inner before the node is removed, and clean_stop is computed as exit_clean || (restarts_disabled && !finish_escalated), so watchdog-killed stragglers report NodeStatus::Failed to the coordinator while an escalated node that still exits 0 remains clean.

I did not find any remaining blocking issues.

Validation run during re-review:
- cargo test -p dora-daemon --lib
- cargo clippy -p dora-daemon -- -D warnings

Note: the worktree has an untracked _work/ directory that is not part of the PR and was ignored for this review. CI still had several pending CircleCI jobs at the time of review.

PR #2163, opened 2026-06-12
fix(arrow-convert): saturate out-of-range NaiveDateTime instead of silently returning epoch

### LINKED ISSUE #2160: arrow-convert: `IntoArrow for NaiveDateTime` silently corrupts out-of-range timestamps to the Unix epoch
> **Note:** This issue was created by a scheduled automated Claude code-review check. It is a single, low-severity correctness finding; please triage/close as appropriate.

## Summary

`IntoArrow for NaiveDateTime` in `libraries/arrow-convert/src/into_impls.rs` silently maps any `NaiveDateTime` outside the nanosecond-representable range to **`1970-01-01T00:00:00` (the Unix epoch)** instead of surfacing the conversion failure. A node that sends a far-future or far-past timestamp as an output gets silent data corruption with no error and no log warning.

## Location

`libraries/arrow-convert/src/into_impls.rs:98-108`

```rust
impl IntoArrow for NaiveDateTime {
    type A = arrow::array::TimestampNanosecondArray;
    fn into_arrow(self) -> Self::A {
        let timestamp =
            match arrow::datatypes::TimestampNanosecondType::from_naive_datetime(self, None) {
                Some(timestamp) => timestamp,
                None => arrow::datatypes::TimestampNanosecondType::default_value(), // == 0i64
            };
        TimestampNanosecondArray::from(vec[timestamp])
    }
}
```

## The problem

A nanosecond-resolution `i64` timestamp can only represent roughly **1677-09-21 .. 2262-04-11**. For any `NaiveDateTime` outside that window, `from_naive_datetime(..)` returns `None`, and the fallback arm substitutes `TimestampNanosecondType::default_value()`, which is `0i64` — i.e. `1970-01-01T00:00:00 UTC`.

Consequences:
- A timestamp like `3000-01-01T00:00:00` is transmitted as `1970-01-01T00:00:00`.
- The corruption is **silent**: no `Err`, no `tracing::warn!`. Downstream nodes (and `TryFrom<&ArrowData> for NaiveDateTime`) happily decode the epoch value as if it were the real data.
- The failure is value-dependent, so it passes all in-range round-trip tests and only bites in production on edge-case dates.

By contrast, the sibling impls don't have this footgun: `NaiveTime` (time-of-day) is always representable in `i64` ns, and `NaiveDate -> Date64` uses an infallible conversion.

## Why it's structured this way

The `IntoArrow` trait signature is infallible (`fn into_arrow(self) -> Self::A`), so the impl genuinely cannot return a `Result`. That constraint is the root cause, but "silently rewrite the value to epoch" is the most surprising of the available options.

## Suggested fixes (pick one)

1. **Saturate** to the min/max representable nanosecond timestamp instead of collapsing to epoch — still lossy, but preserves ordering and is far less misleading than jumping to 1970.
2. **Log on truncation** (`tracing::warn!`) so the corruption is at least observable.
3. **Document the limitation** on the impl and/or add a fallible `try_into_arrow` path for datetime types so callers handling wide date ranges can detect the failure.

Option 1 or 2 (or both) is the minimal change; the round-trip test in `from_impls.rs` could be extended with an out-of-range case to lock in whichever behavior is chosen.

## Severity

Low — only affects timestamps outside ~1677–2262, which is uncommon for robotics/telemetry workloads. Filing for visibility because the corruption is silent rather than a hard error.

Supull:
Hey! This fixes the bug reported in #2160.

Basically if you passed a NaiveDateTime outside the ~1677-2262 range, 
it would silently get converted to the Unix epoch (1970-01-01) with no 
error or warning, which is pretty nasty since you'd have no idea your 
data got corrupted.

The fix saturates to i64::MAX for far-future dates and i64::MIN for 
far-past dates instead, and emits a tracing::warn! so at least it shows 
up in logs if it ever happens.

Fixes #2160


phil-opp:
Reviewed the diff directly. The saturation logic in `into_impls.rs` is correct — `if self >= epoch { i64::MAX } else { i64::MIN }` maps far-future dates to the max representable timestamp and far-past to the min, and the `warn!` makes the clamp visible. That's a clear improvement over silently returning the Unix epoch.

Two things worth addressing:

1. **No test.** This is a behavior change to a conversion that previously returned epoch and now returns a saturated boundary value, but no test pins the new behavior (the repo's policy expects a regression test for a bug fix). A small unit test on `NaiveDateTime::MAX.into_arrow()` / `MIN.into_arrow()` asserting the `i64::MAX` / `i64::MIN` timestamps would lock this in and guard against regressions — and would have to live here since `into_arrow` can't return a `Result`.

2. **The doc comment is garbled.** The added comment reads: *"The nanosecond-resolution `i64` timestamp can only represent dates in roughly representable boundary and a warning is emitted, rather than silently mapping to the Unix epoch."* That sentence is incomplete/ungrammatical — it'd be worth rewriting to state the actual representable range (~1677-09-21 .. 2262-04-11) and the saturate-to-boundary behavior.

No correctness issues with the saturation itself.

---

---

Supull:
> Reviewed the diff directly. The saturation logic in `into_impls.rs` is correct — `if self >= epoch { i64::MAX } else { i64::MIN }` maps far-future dates to the max representable timestamp and far-past to the min, and the `warn!` makes the clamp visible. That's a clear improvement over silently returning the Unix epoch.
> 
> Two things worth addressing:
> 
> 1. **No test.** This is a behavior change to a conversion that previously returned epoch and now returns a saturated boundary value, but no test pins the new behavior (the repo's policy expects a regression test for a bug fix). A small unit test on `NaiveDateTime::MAX.into_arrow()` / `MIN.into_arrow()` asserting the `i64::MAX` / `i64::MIN` timestamps would lock this in and guard against regressions — and would have to live here since `into_arrow` can't return a `Result`.
> 2. **The doc comment is garbled.** The added comment reads: _"The nanosecond-resolution `i64` timestamp can only represent dates in roughly representable boundary and a warning is emitted, rather than silently mapping to the Unix epoch."_ That sentence is incomplete/ungrammatical — it'd be worth rewriting to state the actual representable range (~1677-09-21 .. 2262-04-11) and the saturate-to-boundary behavior.
> 
> No correctness issues with the saturation itself.
> 
> 🤖 This is a fully automated review by Claude (Claude Code) with no human in the loop. Treat the findings as advisory.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

Hey, addressed both points! Fixed the garbled doc comment to clearly state the representable range (1677-09-21 to 2262-04-11) and the saturate-to-boundary behavior, and added a unit test covering both the far-future (year 3000 to i64::MAX) and far-past (year 1000 to i64::MIN) cases.

phil-opp:
Re-reviewed after the latest commit. Both points from the previous automated review are addressed:

- The doc comment now states the actual representable range (~1677-09-21 .. 2262-04-11) and the saturate-to-boundary behaviour, and reads cleanly.
- A unit test (`naive_datetime_out_of_range_saturates`) pins both directions — year 3000 → `i64::MAX`, year 1000 → `i64::MIN`.

The saturation logic (`if self >= epoch { i64::MAX } else { i64::MIN }` + `warn!`) is correct. No further issues found.

---

---

PR #2168, opened 2026-06-12
feat: add pinned memory-pool transport for CPU/CUDA tensor transfer

tang-canran:
add an opt-in memory-pool transport path for large, fixed-shape tensor transfer between CPU and CUDA nodes(updates #1872 )

Architecture introduced relative to upstream dora:
- add a dedicated pinned-memory transport path for repeated high-throughput
  transfer of large, fixed-shape tensors between CPU and CUDA nodes using
  page-locked host memory plus DMA-based heterogeneous transfer
- add a dedicated `libraries/extensions/memory-pool` crate for daemon-side pool
  metadata tracking, lifecycle management, and shutdown cleanup
- add three additive daemon request/reply operations for the memory-pool side
  channel: `RegisterPinnedMemory`, `ReadPinnedMemory`, and `FreePinnedMemory`
- add a reusable shared-memory pool identified by `memory_pool_id`, so the
  sender can register once and overwrite many times instead of re-sending the
  full tensor through the normal message path on every iteration
- add a node-side fast path that reads pool metadata directly from the
  shared-memory header and only falls back to daemon metadata lookup when needed,
  reducing repeated daemon query overhead
- add four Python APIs for the feature:
  `register_memory_pool`, `write_memory_pool`, `read_memory_pool`, and
  `free_memory_pool`
- add an `examples/memory-pool` example set to validate pool transfer throughput
  and exercise the fault-tolerance paths

Performance notes:
- verified CPU -> CUDA repeated transfer throughput remains around 4.5 GB/s in
  the current local environment
- verified CUDA -> CPU repeated transfer throughput remains around 7.6 GB/s in
  the current local environment


heyong4725:
Thanks for moving this to a feature branch! 🙌 One thing before CI can run: the branch still shows merge conflicts with current `main` (several daemon/CLI changes landed this week — `git fetch upstream && git rebase upstream/main` should be mechanical). Once it's conflict-free we'll get CI running and give it a proper review. If this fully replaces #2151, feel free to close that one to keep the queue tidy.

phil-opp:
Reviewed the new `libraries/extensions/memory-pool` crate directly (the self-contained, genuinely-new core of this PR). I scoped the review to that crate: the overall diff currently includes large unrelated deletions against `main` (e.g. `event_stream/mod.rs` -1673, `node/mod.rs` -1203, `README.md` -530, `Cargo.lock` -2209), which is the stale-branch artifact already noted — once it's rebased the diff will be reviewable end-to-end.

One issue I think is important, plus two smaller notes:

**1. `cleanup_all` blanket-deletes every `/dev/shm/dora_pool_*` file, including pools owned by other dataflows/daemons.**

In `src/lib.rs`, `cleanup_all` first removes the shared memory for every entry *in this manager's own table* (the `for id in &ids` loop, lines ~211-221). It then does a second, unconditional sweep of `/dev/shm` that unlinks **any** file whose name starts with `dora_pool_` (lines ~223-241):

```rust
if name.starts_with("dora_pool_") {
    if let Err(err) = std::fs::remove_file(entry.path()) { ... }
}
```

By construction, this second sweep can only touch files that were *not* in this manager's table (those were already handled above) — i.e. pool buffers belonging to a different dataflow, a different daemon, or another process on the same host. On any machine running more than one dataflow/daemon concurrently (multi-dataflow daemon, distributed/cluster tests, two daemons on one box), tearing one down would unlink another's live pinned-memory buffers out from under it. The prefix `dora_pool_` is not scoped to the owning dataflow/daemon. I'd drop the global sweep entirely and rely on the table-driven removal, or scope the swept filenames to this dataflow/daemon id.

**2. `Mutex` poisoning will panic the daemon.** Every accessor uses `self.memory_pool_table.lock().unwrap()`. If any thread panics while holding the lock, all subsequent pool operations (including `cleanup_all` on shutdown) panic instead of degrading gracefully — worth handling the poison case given this runs daemon-side and the repo tracks an unwrap budget.

**3. Dead-ish `in_use` field / process-local `ptr`.** `MemoryPoolEntry::in_use` is always set to `true` on register and never read by `read`/`free`/`cleanup`, so it currently carries no behaviour. Separately, `MemoryPoolMetadata::ptr: u64` is a raw pointer valid only in the registering process but is cloned out to cross-node readers via `read_memory_pool`; worth a doc note that consumers must transfer via `shared_memory_name`, not `ptr`, to avoid misuse.

I did not review the daemon request/reply wiring, the node-side fast path, or the Python APIs in depth — recommend a re-review of those once the branch is rebased and CI can run.

---

---

heyong4725:
Pushed a small maintainer commit (c7a448de) to unblock the required CI gates: `cargo fmt` across the PR-touched files plus mechanical clippy fixes (4× collapsible-if → let-chains in `dora-memory-pool` and the daemon, 3× `node_id.to_string()` → `node_id.as_ref()` at the memory-pool call sites). **No behavior changes** — full workspace clippy `-D warnings` and fmt now pass locally. The substantive review of the memory-pool design will come separately once CI is green.

phil-opp:
**Automated review by Claude** — _this is a fully automated review; no human has vetted these findings, so please apply the usual scrutiny._

I reviewed the current head (commit `c7a448de`). The three issues from the earlier partial review of the `memory-pool` crate (global `/dev/shm/dora_pool_*` sweep in `cleanup_all`, `.lock().unwrap()` daemon-panic, dead `in_use` field) are all resolved in `1280b6ff` — confirmed. That earlier review only covered the `memory-pool` crate; the Python/daemon/node-side fast-path code below was not reviewed before. Looking at it, I found the following issues:

### 1. Pool IDs are not unique across processes → cross-process buffer corruption (high confidence)

`PINNED_COUNTER` is a **per-process** static (`apis/python/node/src/lib.rs:129`), and the pool identity is derived solely from it:

```rust
let shmem_name = format!("dora_pool_{}", pool_counter);   // lib.rs:1076
let buffer_id  = format!("pool_{}", pool_counter);        // lib.rs:1202
```

There is no node/dataflow namespacing. Two different sender nodes are separate OS processes, each starting its counter at 0, so both produce `dora_pool_1` / `pool_1` for their first registration. The second `register_memory_pool` then hits the `create()`-fails fallback and **opens the other node's live buffer**:

```rust
Ok(s) => s,
Err(_) => ShmemConf::new().os_id(&shmem_name).writable(true).open()  // lib.rs:1110-1114
    .wrap_err("failed to open pool shared memory")?,
```

…and proceeds to overwrite its DORADMA header and tensor data (lib.rs:1126-1156). The daemon registration collides too, but the Python side only logs and continues with the now-untracked pool:

```rust
if let Err(e) = self.node.get_mut().register_pinned_memory(buffer_id.clone(), meta) {
    tracing::warn!("[{}] failed to register memory pool: {:#}", self.node_id, e);  // lib.rs:1261
}
```

Any dataflow with more than one node using this API will silently corrupt each other's pools. The pool name/ID needs to be globally unique (e.g. namespaced by node id + dataflow id, or a daemon-allocated id), and a failed `create()` for a name we intend to *own* should be an error, not a silent `open()`.

### 2. Pool free is tracked per-process, so other nodes can read freed/recreated memory (high confidence)

`FREED_POOL_IDS` is also a per-process static (`lib.rs:134`). `free_memory_pool` inserts into it (`lib.rs:1658`) and `try_doradma_read` consults it (`lib.rs:1830-1838`) to fall back to the daemon. But a **receiver node in a different process** never sees that set, and it caches the mapping in `RECV_GPU_VA` / `RECV_CPU_SHMEM` (`lib.rs:1940-1988`). After the owner frees the pool, that receiver keeps handing out `read_ptr` into the stale mapping with no staleness check. Combined with issue #1, if `dora_pool_N` is later recreated by an unrelated registration, the receiver's counter-keyed cache (`RECV_GPU_VA`, keyed only by `counter` at `lib.rs:1925`) silently serves a *different* pool's bytes. The only protection today is the "must not hold tensors after free" doc comment; cross-process lifecycle needs a real invalidation signal.

### 3. (Lower priority) Secondary concerns worth a look

- **GPU cache slot shapes are incompatible across paths.** The `ipc_present == 1` path stores `RecvGpuSlot { gpu_va: 0, gpu_buf: gpu_ptr }` (`lib.rs:1941-1948`); the `effective_as_cuda` path reads `slot_data.gpu_va + data_offset` on a cache hit (`lib.rs:1957`). These are keyed by the same `counter` map, so once IDs collide (issue #1) a `gpu_va: 0` slot yields a near-null device pointer. Distinguish the slot kinds, or this becomes a live fault the moment a counter is reused.
- **`ReadPinnedMemory { free: true }`** (`binaries/daemon/src/lib.rs`) frees/unlinks the pool while still returning `shared_memory_name` for the caller to `open()`. Not currently exercised (Python always passes `free=false`), but the API is unsound as written.

The new protocol additions themselves look sound — the `RegisterPinnedMemory` / `ReadPinnedMemory` / `FreePinnedMemory` request variants are correctly classified for the bincode reply path and handled by all channel impls; I found no serialization/compat break there.

Issues #1 and #2 look like blockers for any multi-node use of this feature.

---

heyong4725 (changes requested):
Thanks for this — it's a serious, substantial piece of work, and a pooled zero-copy transport for repeated large-tensor transfer is genuinely valuable (the right target: register once, overwrite many, skip the per-message Arrow path). I reviewed it in depth (structured pass + security, testing, and maintainability specialists + an independent adversarial pass; the passes converged). The architecture is sound and the cross-process pointer handling is mostly right — the daemon never dereferences the foreign `ptr`, and the slow read path correctly re-resolves a local pointer from the shmem + DORADMA header and caches the mapping to prevent munmap. Requesting changes on one design-level correctness issue and three safety/coverage gaps; everything else is follow-up.

## Blocking

### B1 — No synchronization between in-place writes and zero-copy reads; the benchmark hides it

This is the core one. `write_memory_pool` does `copy_nonoverlapping(src, shmem_ptr + data_offset, size)` straight over the bytes a receiver is mapping zero-copy, and the DORADMA header has no sequence/generation field or ready-flag a reader could check. A receiver iterating its tensor while the sender overwrites observes a torn buffer (front = new, back = old) — unsynchronized concurrent cross-process access, i.e. silent corruption / UB, on every real repeated transfer.

The example masks this completely: `receiver.py` builds `torch_tensor` once at `i==0` (line 33) and reuses that same object for all 100 iterations — it never re-reads. So "100 messages received" and the throughput number are the receiver re-timing one cached buffer (`data_bytes` is constant; only `t_send`→`t_received` skew varies). The 4.5 GB/s figure in the PR description isn't measuring repeated pooled transfer — no second read happens. Meanwhile the one buffer it does hold is being mutated under it.

**Needs a real memory model before merge:** a seqlock/generation counter in the header (reader retries on odd/changed generation) or double-buffering, plus a receiver example that actually re-reads each iteration so the benchmark measures what it claims. Happy to talk through the seqlock approach — it's ~20 lines in the header read/write helpers once those exist (see S4).

### B2 — DORADMA read paths trust `data_offset`/`size` with no bounds check → OOB read → segfault

The write paths got capacity checks, but the read paths didn't. `try_doradma_read` (apis/python/node/src/lib.rs ~1871) and the slow path in `read_memory_pool` (~1567) read `data_offset`/`json_len` from the shmem header and return `shmem_ptr + data_offset` with a `size`-byte tensor **without verifying `data_offset + size <= shmem_size`** (`shmem_size` is read but unused for bounds). Both also read offsets 8/16/24 after validating only the 8-byte magic, so a segment smaller than `HEADER_SIZE` causes an OOB read in the header parse itself. A corrupted or version-mismatched header yields an out-of-bounds tensor that segfaults the node on consume. **Fix:** mirror the write-path capacity check on all three header-parse sites; reject `shmem_size < HEADER_SIZE` up front; return `None` (fall back to daemon) on violation.

### B3 — `read_ptr == 0` returns `{"ptr": 0}`, then `from_address(0)` segfaults

The slow path and error path return `{"ptr": 0, ...}` (lib.rs ~1594, ~1606); `tensor_from_info`'s CPU path does `(c_byte*size).from_address(0)` then `np.frombuffer(...).reshape(...)` → dereferences address 0 → hard segfault, no Python exception. The example only dodges this because its `read_after_free` branch discards the result without calling `tensor_from_info`; any real consumer that reads-then-builds after a miss/free crashes the node. **Fix:** raise from `read_memory_pool` on `ptr == 0` (and/or reject `ptr == 0` in `tensor_from_info`) instead of handing back a NULL tensor_info.

### B4 — The feature ships with no tests and no CI coverage

- `MemoryPoolManager` (libraries/extensions/memory-pool/src/lib.rs) is a pure, GPU-free `Arc<Mutex<HashMap>>` module with register/read/free/cleanup_all, ownership rejection, double-free, and poison recovery — and **zero** unit tests. Every error branch is trivially testable on any CPU runner with `shared_memory_name: None`. This one's a quick, high-value add.
- None of the six example YAMLs are referenced in `tests/example-smoke.rs`, `scripts/smoke-all.sh`, or `nightly.yml` (grep: zero hits), and every YAML forces a GPU (`receiver_device: cuda`, and `receiver.py` hard-raises without CUDA), so there's no CPU-only scenario a GPU-less runner could exercise. Add a `cpu2cpu.yml` and wire it into the smoke suite so the daemon pool path runs in CI.
- The fault-tolerance scenarios (`duplicate_free`, `read_after_free`, `write_after_free`) assert nothing — they run the illegal op and discard the result. A scenario that passes whether or not the daemon correctly rejects the op is a demo, not a test; wrap the second call and assert it raises / is rejected.

## Should-fix (follow-ups OK)

- **S1 — /dev/shm leak on unclean exit.** Cleanup runs only in the daemon's graceful `cleanup_all`; a node crash without `free`, or a daemon SIGKILL (the #2152 class), orphans up-to-61MB `dora_pool_*` segments in tmpfs with no startup sweep to reclaim them — unbounded across restart cycles. Add a best-effort boot-time sweep, or have nodes `set_owner(true)` so the OS reclaims on exit.
- **S2 — CUDA resource leaks on the GPU path** (untested in CI). There's no `cudaIpcCloseMemHandle` call anywhere — `RECV_GPU_VA.remove()` drops the slot without closing the IPC mapping; `cudaHostRegister` and the sender's `cudaMalloc` aren't paired with unregister/free on error paths. These leak GPU memory across pool churn.
- **S3 — stringly-typed metadata marshalling.** Every field is hand-packed/unpacked via `.get(key).and_then(|p| if let Parameter::X ...)` — ~10 blocks in the daemon, mirrored 3× in Python, with bare string keys on both sides and `unwrap_or(default)` swallowing renames. `MemoryPoolMetadata` already exists; give it serde derives and ship it as a typed field on the request/reply variants, deleting the hand-rolled blocks.
- **S4 — DORADMA layout duplicated.** `HEADER_MAGIC` is redeclared 5×, the offset arithmetic (8/16/24/32) hand-coded at ~14 sites with no single source of truth — writer and readers can drift into silent corruption. Define the layout once (`#[repr(C)]` header struct or named offset consts + read/write helpers). This is also the natural home for the B1 generation counter and the B2 bounds checks.
- **S5 — `free=true` on ReadPinnedMemory swallows the free error** at debug level and still returns success, so a non-owner read-and-free silently fails to reclaim → contributes to S1's leak. Surface it or warn-level it.

## Informational

- The env var is misspelled `massage_num` across `sender.py`, `receiver.py`, and all six YAMLs — it's self-consistent so it works, but it's the public knob users will copy. Rename to `message_num` while it's free.
- `FREED_POOL_IDS` only grows (one `String` per registration, never pruned); the "naturally bounded" comment is backwards. Bound it or fix the comment.
- Dead fields `allocator_pid` / `cuda_registered` are extracted in the daemon but never produced or read; `host_ptr` is set by the fast read path but read by no one and absent from the slow path (the two read dicts differ). Drop or wire them.

Net: the bones are good and most of this is mechanical or additive — B1 is the one that needs a design decision before merge. I'm happy to re-review quickly once the synchronization model and the read-path bounds checks land, and the rest can reasonably be follow-up PRs if you'd rather land this incrementally behind the (already present) opt-in.

phil-opp:
**Automated review by Claude** — _this is a fully automated review with no human in the loop; please apply the usual scrutiny._

Re-reviewed at head `40606929` (force-rebased since my last review of `c7a448de`, adding the seqlock/namespacing commits `ea342235` + `40606929`).

Good news first — several blockers from the earlier reviews are addressed by the rebase:
- Pool IDs are now namespaced `dora_pool_{node_id}_{counter}` and a failed `create()` errors instead of silently opening another node's buffer (prior automated review issue #1).
- The read paths now bounds-check `data_offset + size <= shmem_size` and reject `shmem_size < HEADER_SIZE` in `try_doradma_read`, and the slow path now errors on `ptr == 0` instead of returning a NULL tensor (heyong's B2/B3).
- A seqlock generation field was added (heyong's B1).

The new commits, however, introduce/leave the following important issues:

### 1. Use-after-free: the CPU slow path returns a dangling pointer (high confidence)

`read_memory_pool` slow path, `apis/python/node/src/lib.rs:1706-1718`:

```rust
read_ptr = (shmem_ptr as u64 + data_offset as u64) as i64;
// Cache shmem to prevent munmap of the returned pointer.
// Extract counter from "dora_pool_{counter}".
if let Some(counter_str) = name.strip_prefix("dora_pool_") {
    if let Ok(counter) = counter_str.parse::<u64>() {
        ... cpu_cache.entry(counter).or_insert(RecvCpuSlot { _shmem: shmem, ... });
    }
}
```

The pool name is now `dora_pool_{node_id}_{counter}` (set at lib.rs:1103 / 1373). After `strip_prefix("dora_pool_")` the remainder is `{node_id}_{counter}`, which never parses as `u64`, so the cache insert is skipped. `shmem` is then dropped at the end of the `if let Ok(shmem)` block (munmap), and the `read_ptr` handed back to Python (and into `tensor_from_info`) points into the unmapped region — a use-after-free / segfault. The stale comment (`"dora_pool_{counter}"`) confirms this parser wasn't updated when the namespacing was added. `try_doradma_read` parses correctly via `splitn(3, '_')`, so this only bites when the daemon fallback path is taken; `free_memory_pool` at lib.rs:1751 already uses the correct `pool_{node_id}_{counter}` parsing pattern.

### 2. The seqlock doesn't guard the consumer's actual read — B1's torn-buffer window remains for CPU zero-copy (high confidence)

The seqlock retry only wraps the in-function header/pointer fetch: `try_doradma_read` reads `read_gen`, computes `read_ptr = shmem_ptr + data_offset` (CPU path, lib.rs:2178), re-checks `read_gen2 == read_gen` (lib.rs:2192) and returns the **raw pointer** in the dict. The consumer (numpy/torch) reads the tensor bytes *after* the function returns, with no generation guard — so a sender doing `copy_nonoverlapping` into the same buffer on the next `write_memory_pool` still tears the buffer a receiver is iterating. The seqlock protects the metadata window, not the data consumption it was meant (per B1) to protect. This is only safe under a strict turn-based single-writer/single-reader discipline; that should be enforced (double-buffer) or at minimum documented. As written, the code looks like B1 is fixed while the torn-read window is still open for any receiver that re-reads.

### 3. The GPU zero-copy fast path is defeated after the first overwrite (medium confidence)

The writer bumps the generation by 2 on every write (begin+end, lib.rs:1417/1430/...). The GPU read paths cache `generation: read_gen` at first import (lib.rs:2131 / 2170) and bail with `return Ok(None)` whenever `current_gen != slot_data.generation` (lib.rs:2106 / 2145), never refreshing the stored value. So after the first `write_memory_pool`, every subsequent read sees a changed generation and falls back to the daemon — defeating the "register once, overwrite many" headline for the GPU path. The cached IPC handle / device VA is actually stable across data overwrites (only the bytes change), so invalidating on a data-generation bump is unnecessary and a per-iteration perf cliff. Worth also verifying the daemon fallback doesn't produce wrong results here, since the slow read path returns a host pointer (`shmem_ptr + data_offset`) for what is a GPU pool.

Still outstanding from the prior review (unchanged across these commits): `ReadPinnedMemory { free: true }` in `binaries/daemon/src/lib.rs` unlinks the segment yet still returns `shared_memory_name` for the caller to `open()`; unexercised today (Python always passes `free=false`) but unsound as written.

---

phil-opp:
**Automated review by Claude** — _fully automated, no human has vetted these findings; treat as suggestions._

Re-reviewed at head `6e028bd3` (the commit pushed after the previous review). The two code fixes in that commit look correct as far as they go: the CPU slow-path counter parse now uses `rsplit_once('_')` so the `RECV_CPU_SHMEM` insert actually runs (no more dropped-handle dangling pointer), and dropping the per-write generation check on the GPU cache-hit path does restore "register once, overwrite many" since the writer bumps the gen on every write. The seqlock torn-data limitation is now documented rather than fixed, which is a reasonable call given the turn-based discipline.

One issue I think is important, and it's the same class the commit message claims is already closed:

### Receiver caches are keyed by `counter` alone, so the `node_id` namespacing doesn't prevent cross-pool aliasing

`6e028bd3` removes the GPU generation check and justifies it with *"pool ID uniqueness (node_id namespacing) prevents cross-pool confusion."* But the namespacing is only on the **shmem name** (`dora_pool_{node_id}_{counter}`, line 1103) and the **buffer_id** (line 1248). The receiver-side caches are keyed by the bare `counter`:

- `RECV_GPU_VA: HashMap<u64, RecvGpuSlot>` (line 175)
- `RECV_CPU_SHMEM: HashMap<u64, RecvCpuSlot>` (line 191)

In `try_doradma_read`, `pool_node_id` is parsed (line 1994) and used to build `shmem_name` (line 2009), but the cache lookup/insert uses only `counter` (`cache.get(&counter)` at lines 2124/2161, `cpu_cache.insert(counter, …)` at 2196). And `counter` comes from `PINNED_COUNTER` (line 129), which is a **per-process** static.

So two different sender nodes each produce `counter = 1` for their first pool. A receiver that consumes pools from **both** senders (a normal fan-in topology) keys both under `counter = 1`. On the steady-state cache-hit path, `cache.get(&counter)` returns the **first** pool's cached slot — its `_shmem` / `gpu_buf` / `gpu_va` — with no `node_id` re-check, so reads of the second pool silently return the first pool's tensor bytes (or device pointer). Removing the generation check makes this worse, since there's now no invalidation signal at all on the GPU hit path — but note the gen check never protected against this anyway, because the cached `_shmem` itself is the wrong pool's.

The fix is to key the receiver caches by the full `buffer_id` (or a `(node_id, counter)` tuple), not the bare counter.

### Related: three different parse strategies for the same `pool_{node_id}_{counter}` format

The same id format is parsed three different ways — `rsplit_once('_')` in the CPU slow path (line 1717), `splitn(3, '_')` + `parts[2]` in `try_doradma_read` (line 1986), and `splitn(2, '_')` + `parts[1]` in `free_memory_pool` (line 1761). The two `splitn` variants misparse any `node_id` that itself contains `_` (legal in dora node ids): `free_memory_pool` then computes `counter = None` and silently skips the receiver-cache cleanup, leaking/staling the cache entry. Keying the caches by the full `buffer_id` would remove the need to re-extract the counter at all and resolves both this and the aliasing above in one move.

---

phil-opp:
**Automated review by Claude** — _this is a fully automated review; please verify before acting on it._

I re-reviewed against the latest head (`48d8f278`).

✅ The headline blocker from the previous review is resolved: `RECV_GPU_VA` and `RECV_CPU_SHMEM` are now `HashMap<String, …>` keyed by the full namespaced `buffer_id`, so the cross-pool aliasing in fan-in topologies (two senders both using `counter=1`) is fixed. The read-path bounds checks are also in place.

Two previously-raised items still appear unaddressed in this revision:

1. **Underscore in node id silently disables the fast path.** `try_doradma_read` does `buffer_id.splitn(3, '_')` then `parts[2].parse::<u64>()` on `buffer_id = "pool_{node_id}_{counter}"`. dora node ids may legally contain `_`; for those, `parts[2]` carries the rest of the node id plus the counter, the `parse::<u64>()` fails, and the function returns `Ok(None)` → it always falls back to the daemon slow path. Not a correctness bug (the String-keyed caches make the fallback safe), but the "register once, overwrite many" fast path never engages for such nodes. Consider keying on a delimiter that can't collide with node-id characters, or rsplit on the counter.

2. **`ReadPinnedMemory { free: true }` is unsound as written.** The daemon handler frees/unlinks the pool when `free` is true, then still returns `shared_memory_name` in the reply for the caller to `open()` — a read-after-free. It's dead today since the Python side always passes `free=false`, but I'd either remove the `free` flag from the read op or stop returning `shared_memory_name` when freeing.

No new issues beyond these.

---

phil-opp:

I focused on the latest commit (`5153cbf`, "restore single-read pattern in receiver"), since it's the only change since the previous review.

**Correctness: the single-read pattern is safe here — it is *not* a torn read.** The example dataflow is strictly turn-based. `sender_node` has `inputs: { next_require: receiver_node/next_require }` and blocks on `node.next()` before each `write_memory_pool`; the receiver only emits `next_require` *after* it has accessed the tensor for that iteration. So the sender's overwrite for iteration `i+1` always happens-after the receiver's access for iteration `i`. The "the existing zero-copy tensor reflects the new data in place" reasoning in the commit message holds given this protocol — no data race.

**But the change makes the throughput benchmark measure the wrong thing.** In steady state (`i >= 1`) the receiver loop never reads the tensor data — it only touches `torch_tensor.nbytes`, a constant. As a result:

- `Average transfer throughput = data_bytes / (t_received - t_send)` is effectively the round-trip latency of the empty `data` signal expressed as a bandwidth number. It does not reflect any bytes being moved or read on the receive side (in a zero-copy pool nothing is copied on receive — the "transfer" is the sender's in-place write).
- The example no longer validates that the received contents are correct for `i >= 1` (only the `i == 0` preview actually reads data).

The previous per-iteration `read_memory_pool` + `tensor_from_info` at least exercised the read/seqlock path on every iteration. As written now, the headline GB/s figures don't demonstrate pooled-tensor transfer throughput. If the intent is a throughput (and correctness) benchmark, the receiver should actually consume — and ideally assert on — the data each iteration.

---

phil-opp:

One follow-up on the B2 read-path bounds checks (the rest of the earlier review feedback looks addressed): the *write* paths were hardened with overflow-safe `saturating_sub`, but the two *read* checks still use plain `usize` addition —

- `try_doradma_read`: `if data_offset + size > shmem_size` (`apis/python/node/src/lib.rs:1589`)
- slow path: `if data_offset + (size as usize) > shmem.len()` (`apis/python/node/src/lib.rs:1339`)

`data_offset` is read straight from the shmem header (line 1546) and `size` from the in-segment JSON / daemon metadata. A corrupted or version-mismatched header with a large `data_offset` (near `usize::MAX`) overflows the sum, passes the check, and then `read_ptr = shmem_ptr + data_offset` (1696 / 1349) is handed back as the tensor base — an out-of-bounds pointer the consumer dereferences (segfault / OOB read). This is the same class of bug B2 closed on the write side; the read side still has the hole.

Suggest mirroring the write paths: `data_offset.checked_add(size)` (or `size > shmem_size.saturating_sub(data_offset)`) at both sites, and rejecting `data_offset > shmem_size` up front — `data_offset` on its own is currently never bounds-checked (only `json_len` is, at line 1549). In normal same-dataflow operation the sender always writes a small `data_offset`, so this only bites on a corrupted/forged/mismatched header — precisely the threat model B2 targeted.

---

phil-opp:
**Use-after-free on the CPU read fast path (`try_doradma_read`)**

Reviewed at head `5153cbf`. I found one issue I don't see raised in the prior reviews — a use-after-free on the **CPU cache-hit path** in `try_doradma_read` (`apis/python/node/src/lib.rs`).

Each call opens a *fresh* mapping:

```rust
let shmem = match ShmemConf::new().os_id(&shmem_name).open() { ... };
let shmem_ptr = shmem.as_ptr();
```

and the CPU branch derives the returned pointer from that fresh mapping, then only keeps it alive on a cache miss:

```rust
} else {
    read_ptr = shmem_ptr as u64 + data_offset as u64;   // fresh mapping's base
    let mut cpu_cache = RECV_CPU_SHMEM.lock()...;
    if !cpu_cache.contains_key(buffer_id) {
        cpu_cache.insert(buffer_id.to_string(),
            RecvCpuSlot { _shmem: shmem, generation: read_gen });
    }
}
```

On the **first** read for a `buffer_id`, `contains_key` is false, so the fresh `shmem` is moved into `RECV_CPU_SHMEM` and the mapping stays alive — `read_ptr` is valid. But on **every subsequent** read (cache hit), the insert is skipped, so the freshly-opened `shmem` is **dropped (munmap) at the end of the function**, while the returned `read_ptr` points into *that* now-unmapped mapping — not into the cached one (the two `mmap`s are at different addresses). The consumer then builds a zero-copy tensor over the dangling host pointer in `tensor_from_info` → use-after-free / SIGSEGV.

This is the steady-state path (every read after the first) on the primary CPU scenario (`cpu2cpu`), so it isn't an edge case. It's distinct from the previously-fixed slow-path parse and counter-aliasing bugs — those were about *which* pool/parse; this is the fresh-vs-cached mapping mixup on a correctly-identified pool. The GPU branches are unaffected because they return the cached GPU VA / IPC pointer (stable), not a pointer into the fresh host mapping.

`RecvCpuSlot` currently stores no base, so the cache hit has nothing to recompute `read_ptr` from. Suggested fix: store the cached mapping's base (e.g. `base: u64`) in `RecvCpuSlot`, and on a cache hit set `read_ptr = cached.base + data_offset` and let the fresh `shmem` drop — only the first call's mapping should ever back the returned pointer.

(Separately, the plain-`usize` read-path bounds check `data_offset + size > shmem_size` at line ~1591 can still overflow on hostile metadata, unlike the `saturating`/checked write paths — already noted in an earlier review and still standing.)

---

---

phil-opp:
Took a focused pass at the write path at head `5153cbf` and think there's a use-after-free that hasn't been called out yet — distinct from the read-path UAFs already discussed.

**`write_memory_pool` fast path, `PINNED_POOL` cache-miss branch (`apis/python/node/src/lib.rs:1404-1414`):**

```rust
} else {
    // Cache miss: open via ShmemConf (mmap, slower)
    let shmem_name = format!("dora_pool_{}_{}", self.node_id, counter);
    match ShmemConf::new().os_id(&shmem_name).open() {
        Ok(shmem) => {
            let cap = shmem.len();
            (shmem.as_ptr(), None, cap)   // <- shmem dropped (munmap) when this `let` stmt ends
        }
        Err(_) => (std::ptr::null_mut(), None, 0),
    }
};
```

`shmem` is a local in the match arm, so it's dropped (munmap) at the end of the `let (shmem_ptr, store_back, shmem_capacity) = …` statement — only the raw pointer and `None` survive. `store_back` is `None`, so nothing keeps the mapping alive. `shmem_ptr` is then used for the magic/header reads (line 1417 onward) and, critically, the seqlock writes + `copy_nonoverlapping(ptr_val, shmem_ptr.add(data_offset), size)` at lines 1501-1507 — all through a pointer into a now-unmapped region. That's a write through a dangling pointer (memory corruption or SIGSEGV), not just a read.

The cache-hit branch is fine because it returns `Some(slot_data)` and re-inserts it into `PINNED_POOL` after the write (line 1518); the miss branch has no equivalent keep-alive. Since this branch is the explicit fallback for when the writing context hasn't populated `PINNED_POOL`, it's reachable in real use. Suggest binding the opened `Shmem` in an outer scope for the duration of the write (or inserting it into `PINNED_POOL` before writing, mirroring the hit path).

Two read-side items still appear to stand from earlier reviews, for tracking — I'm re-mentioning since they're unaddressed at this head:
- The CPU cache-*hit* path in `try_doradma_read` (lib.rs:~2217-2227) keeps the fresh mapping alive only on a cache miss (`if !contains_key`), so steady-state hits return a pointer into a dropped mapping; the slow path's `or_insert` (lib.rs:~1742-1748) has the same shape.
- The read-path bounds checks (lib.rs:~1727 / ~2110) still use plain `usize` addition (`data_offset + size > len`), which can overflow on a hostile/corrupted `data_offset` and pass the check, where the write paths correctly use `saturating_sub`.

---

---

phil-opp:
Re-reviewed at head `6e0b6ca7`. The only change since the last review (`5153cbf`) is the `write_memory_pool` counter-parse `rsplit` perf fix, so the memory-safety items from the last two reviews are still open — and I found one additional issue that I don't think was raised before.

**New finding — daemon unlinks an arbitrary file from a node-controlled name (`libraries/extensions/memory-pool/src/lib.rs`).** `free_shared_memory` does `remove_file(format!("/dev/shm/{}", shm_name))` with no validation of `shm_name`, and `register_memory_pool` stores `shared_memory_name` verbatim from the `RegisterPinnedMemory` request metadata (only `ptr`/`size`/`dtype`/`shape` are validated). So a node can register a pool with `shared_memory_name = "../../<path>"` and have the daemon unlink an arbitrary file (with daemon privileges) on `FreePinnedMemory` / shutdown `cleanup_all`. Note the asymmetry: the startup `cleanup_orphans` sweep *does* enforce the `dora_pool_` prefix, but the per-entry free path does not. Fix: validate the name (require the `dora_pool_` prefix, reject any path separator and `..`) in the register handler and again in `free_shared_memory`.

**Still open from prior reviews (unaddressed by the latest commit):**

1. **CPU read fast-path use-after-free (`apis/python/node/src/lib.rs`, `try_doradma_read`).** Each call opens a fresh `Shmem`, but the CPU branch only keeps that mapping alive on a cache *miss*. On every steady-state cache hit the fresh mapping is munmapped at function exit while the returned `read_ptr` still points into it, so the consumer builds a zero-copy tensor over a dangling host pointer. `RecvCpuSlot` needs a stored base so the hit path can recompute `read_ptr = base + data_offset` from the cached mapping.

2. **`write_memory_pool` cache-miss use-after-free.** In the cache-miss arm `Ok(shmem) => (shmem.as_ptr(), None, cap)`, `shmem` is dropped (munmap) at the end of the `let` statement with nothing keeping it mapped, and the seqlock writes + `copy_nonoverlapping` then write through the dangling `shmem_ptr`. Bind the opened `Shmem` for the duration of the write (the cache-hit arm is fine because it re-inserts into `PINNED_POOL`).

3. **Read-path bounds checks can integer-overflow (MEDIUM).** `data_offset + size > shmem_size` in both read sites uses plain `usize` addition (the write paths already use `saturating_sub`); a near-`usize::MAX` `data_offset` from a corrupted/hostile header overflows the sum, passes the check, and yields an out-of-bounds tensor base. Use `checked_add` and reject `data_offset > shmem_size` up front.

Items 1 and 2 fire on the normal CPU read/write path, so I'd treat them as blocking. I didn't re-verify the CUDA paths in this round.

---
*This is a fully automated review by Claude (Claude Code). No human has vetted these findings — please verify before acting on them.*

---

phil-opp:
**Automated review by Claude** — _this is a fully automated review; please verify before relying on it._

Re-reviewed at head `292c400`. This commit resolves all five blockers from the previous review (`6e0b6ca7`) — confirmed in code:

- ✅ **CPU read fast-path UAF** — `RecvCpuSlot` now stores `base`; the cache-hit branch returns `cached.base + data_offset` and lets the fresh mapping drop (`apis/python/node/src/lib.rs:2238-2252`).
- ✅ **`write_memory_pool` cache-miss UAF** — the cache-miss arm now wraps the opened `Shmem` in a `PoolSlot` and re-inserts it into `PINNED_POOL`, keeping the mapping alive through the copy (`lib.rs:1408-1425`).
- ✅ **Read-path bounds overflow** — both read sites use overflow-safe checks (`data_offset > shmem_size || size > shmem_size.saturating_sub(data_offset)`) at `lib.rs:2127` and `lib.rs:1738`.
- ✅ **Path traversal** — `free_shared_memory` validates the `dora_pool_` prefix and rejects `/` and `..`, enforced on both `free_memory_pool` and `cleanup_all`.
- ✅ **Benchmark** — the receiver now reads the tensor each iteration (checksum + change assertion).

**One residual (low–medium), same shape as the fast-path fix that just landed:**

The *slow*-path read still derives the returned pointer from the **fresh** mapping — `read_ptr = shmem_ptr + data_offset` (`lib.rs:1750`) — then `cpu_cache.entry(buffer_id).or_insert(...)` (`lib.rs:1756`). On a cache **hit**, `or_insert` is a no-op, so the freshly-opened `shmem` is munmapped at function exit while `read_ptr` still points into it — a use-after-free identical to the one just fixed in `try_doradma_read`. The slow path only fires on a fast-path miss (freed pool, odd seqlock generation mid-write, JSON/parse failure), so it isn't the steady-state CPU loop, but a receiver that repeatedly falls back can reach it. Suggest mirroring the fast-path fix: on a cache hit, set `read_ptr = cached.base + data_offset` and let the fresh `shmem` drop.

Nothing else new. The protocol additions remain additive/compat-safe, and the `dora-memory-pool` crate is poison-safe, name-validated, and unit-tested. The seqlock-guards-metadata-only / turn-based single-writer-single-reader discipline is documented and remains a design constraint by intent.

---

heyong4725:
Review findings from the latest head (`292c400`):

1. **High: daemon fallback CPU read can still return a dangling pointer.** In `apis/python/node/src/lib.rs`, `read_memory_pool` computes `read_ptr` from the freshly opened `shmem` (`read_ptr = shmem_ptr + data_offset`), then uses `RECV_CPU_SHMEM.entry(buffer_id).or_insert(...)`. On a cache hit, `or_insert` is a no-op, so the fresh mmap is dropped at function exit while the returned pointer still points into that fresh mapping. This is the same fresh-vs-cached mapping issue fixed in `try_doradma_read`; the fallback should mirror that fix and return `cached.base + data_offset` on cache hits.

2. **High: the daemon fallback can label a host mmap pointer as CUDA memory.** The fallback always derives `read_ptr` from `shmem_ptr + data_offset`, but returns `device = pinned_type`. For a CUDA receiver, `tensor_from_info` treats that value as a CUDA pointer, even though it is a host mmap address. If the fast path falls back for a CUDA pool, this can produce an invalid device pointer. Either avoid the daemon fallback for CUDA zero-copy pools or reconstruct/import the correct CUDA pointer before returning `device: cuda`.

3. **Medium: daemon startup globally unlinks every `/dev/shm/dora_pool_*` segment.** `MemoryPoolManager::cleanup_orphans()` sweeps all matching names and `Daemon::new` calls it unconditionally. Starting another daemon on the same host can unlink live pools owned by an existing daemon/dataflow, causing future receiver opens to fail. The sweep should be scoped to this daemon/dataflow or use a staleness/ownership check.

Validation I ran locally:
- `git diff --check refs/remotes/origin/main...HEAD`
- `cargo test -p dora-memory-pool`
- `cargo test -p dora-daemon memory_pool` (built successfully, 0 matching tests)
- `cargo check -p dora-node-api-python` (completed with warnings)

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Please verify independently before acting.

Re-reviewed at head `2509a184` (commits `61f4808` and `2509a184` landed since the last review). The two new commits resolve the issues flagged previously:

- **slow-path (`read_memory_pool`) use-after-free** — the CPU slow path now checks `RECV_CPU_SHMEM` first and returns `cached.base + data_offset` on a hit, letting the freshly-created mapping drop instead of returning a pointer into it (`apis/python/node/src/lib.rs:1753-1765`);
- **CUDA device mislabel** — CUDA pools in the slow path now `bail!` rather than returning a host pointer labeled `device=cuda` (`apis/python/node/src/lib.rs:1781-1791`);
- **global orphan sweep** — `cleanup_orphans` is no longer invoked by the daemon (`libraries/extensions/memory-pool/src/lib.rs`), removing the multi-daemon `/dev/shm/dora_pool_*` unlink hazard.

No new blocking issues at this head. The protocol additions remain cleanly additive (new `DaemonRequest` variants appended at the end of the enum, new `DaemonReply::PinnedMemoryMetadata`, both correctly classified in the reply-path dispatch; `NodeEvent` untouched), so this stays backward-compatible at the wire level.

One minor, non-blocking observation: `FREED_POOL_IDS` is a permanent per-process tombstone. If a *sender* process restarts, its `PINNED_COUNTER` resets and it re-emits e.g. `pool_{node}_1`; a long-lived receiver that already freed that id will permanently fall back to the daemon slow path for the recreated pool. Correctness is preserved (the String-keyed caches still resolve it), but the register-once/overwrite-many fast path silently never re-engages for that id. Not a blocker for the turn-based example.

---

phil-opp:

Re-reviewed at head `9320478`. The only change since the last review (`2509a184`) is the 8-line `fix: clear FREED_POOL_IDS tombstone on re-register`, targeting the non-blocking `FREED_POOL_IDS` tombstone note from the previous review. No new blocking issues and no regressions — the protocol additions remain additive and the previously-fixed UAF/bounds/path-traversal/CUDA-mislabel/orphan-sweep items are unchanged.

One observation on the new commit itself (non-blocking, same class as the note it addresses):

**The tombstone clear is placed in the wrong process to fix the scenario in its own commit message.** `FREED_POOL_IDS` is a per-process static. In the normal lifecycle (sender registers, receiver frees — per `22ba1fb`), the tombstone is *inserted* by `free_memory_pool` and *consulted* by `try_doradma_read`, both of which run in the **receiver** process. The new `FREED_POOL_IDS.remove(&buffer_id)` is in `register_memory_pool`, which runs only in the **sender** process. So when a sender restarts and re-emits `pool_{node}_1`, it clears its own (irrelevant) tombstone, while the long-lived receiver — whose fast path is actually gated by the tombstone — never calls `register_memory_pool` for that id and keeps its stale entry. The fix therefore only helps the degenerate same-process register+free case, not the cross-process case described.

This is harmless (correctness is preserved via the daemon slow path; it just doesn't deliver the promised fast-path re-engagement cross-process). To actually re-engage the receiver's fast path on pool recreation, the receiver should drop the tombstone when it observes a fresh registration for that `buffer_id` (e.g. invalidate its `FREED_POOL_IDS` entry on the next successful `try_doradma_read` that sees a live seqlock header for a re-created segment), rather than relying on the sender to clear its own set.

---
_Generated by Claude Code_

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Please verify before acting.

Re-reviewed at head `f397b74`. The only change since the last review (`9320478`) is the 8-line follow-up moving the `FREED_POOL_IDS` tombstone clear from sender-side `register_memory_pool` to receiver-side `try_doradma_read` — exactly the fix the previous review recommended. Verified: `try_doradma_read` now drops the tombstone on a successful read (`apis/python/node/src/lib.rs:1789-1794`), so the receiver fast path correctly re-engages after a sender restart recreates the same `buffer_id`. Correctness-neutral; no regression.

All memory-safety items from earlier rounds remain fixed at this head: the CPU / slow-path / write-cache-miss use-after-free fixes, read-path integer-overflow bounds checks (`saturating_sub`), path-traversal validation in `free_shared_memory` (`dora_pool_` prefix, rejects `/` and `..`), CUDA device-mislabel rejection in the daemon fallback, table-driven (non-global) shutdown cleanup, the `ReadPinnedMemory{free:true}` dangling-name fix, and poison-safe locks. The seqlock-guards-metadata-only / turn-based single-writer-single-reader model remains a documented design constraint, not a bug.

No new blocking issues found.

_Generated by Claude Code_

---

phil-opp:

I focused on the untrusted-header consume path (a node reading pool metadata written by a peer) and found two issues that I don't think earlier reviews covered. The previously-flagged items (the UAFs, `data_offset + size` bounds, path traversal, device-mislabel, `cudaIpcCloseMemHandle`, etc.) all look addressed at the current head.

**1. `shape`/`dtype` are never validated against `size`, and the CUDA consume path ignores `size` entirely → OOB GPU tensor.**
`try_doradma_read` (`apis/python/node/src/lib.rs` ~2138-2159) bounds-checks only the JSON `size` against the shmem segment, then passes the peer-controlled `shape`/`dtype` through into the returned dict (~2291-2303). `tensor_from_info` (`apis/python/node/dora/cuda.py` ~391-398) builds the CUDA tensor's `__cuda_array_interface__` from `ptr` + `shape` + `typestr` and **never references `size`**. A peer that writes a small allocation but a large `shape` (or a wider `dtype`) produces a GPU tensor whose `shape × itemsize` exceeds the real buffer → out-of-bounds device reads on consume. The daemon-side check (`binaries/daemon/src/lib.rs` ~3473-3484) only verifies the fields are non-empty, not consistent. The CPU branch is incidentally saved by numpy's `.reshape()` raising on mismatch (`cuda.py` ~411), but the GPU path has no equivalent backstop. Suggested fix: in `try_doradma_read` (and the daemon register handler) require `product(shape) * itemsize(dtype) == size` and reject otherwise. Same class as the earlier `data_offset+size` bounds work, but on the shape axis.

**2. Receiver-side `cudaHostRegister` is never paired with `cudaHostUnregister` before munmap → leaked pinned registration / UB.**
In the `effective_as_cuda` non-IPC branch the receiver calls `_register_host(shmem_ptr, shmem_size)` on the peer's mapping (`apis/python/node/src/lib.rs` ~2235) and caches it in `RECV_GPU_VA` with `gpu_va != 0, gpu_buf == 0`. But `free_memory_pool` (~1850-1862) only calls `_ipc_close` when `gpu_buf != 0`; for the host-registered slot it just drops `_shmem` (munmap) without `cudaHostUnregister`. munmap'ing a region still CUDA-host-registered leaks the pinned registration and is undefined. The earlier fix added the sender-side `_unregister_host` (~1838) but not this receiver path. Suggested fix: call `_unregister_host(...)` before dropping a host-registered `RECV_GPU_VA` slot.

**3. (lower severity, already-known) tmpfs leak on unclean exit.** `cleanup_orphans()` is `#[allow(dead_code)]` and `set_owner(false)` means the OS doesn't reclaim segments on process exit either, so a node/daemon killed without a graceful free orphans `dora_pool_*` segments with no reclaim path. This is the previously-discussed S1 — flagging that it currently has no mitigation after the startup-sweep was reverted; a dataflow-id-namespaced sweep would close it.

Both #1 and #2 are on the GPU path, which isn't exercised in CI, so they wouldn't surface in the test suite.

---

phil-opp:
Re-reviewed at head `06c72cb`. The two new commits since the last review address the three items flagged previously: the GPU `product(shape) * itemsize <= size` validation in `tensor_from_info` and the dataflow-id-namespaced orphan sweep both look correct. One problem in the receiver-side unregister fix, though:

**`f5ecc75` unregisters the wrong pointer in the receiver host-register path.**

In `try_doradma_read` (`apis/python/node/src/lib.rs` ~2253) the receiver registers the *host* mapping:

```rust
bound.call_method1("_register_host", (shmem_ptr as u64, shmem_size))?;   // cudaHostRegister(host_ptr)
let va = bound.call_method1("_get_device_ptr", (shmem_ptr as u64,))?...; // cudaHostGetDevicePointer -> DEVICE VA
// slot.gpu_va = va
```

So `slot.gpu_va` holds the **device** VA returned by `cudaHostGetDevicePointer`, not the host pointer. But the new cleanup (~1873) does:

```rust
} else if slot.gpu_va != 0 {
    let _ = bound.call_method1("_unregister_host", (slot.gpu_va,)); // cudaHostUnregister(device_va)
}
```

`cudaHostUnregister` must be called with the **host pointer originally passed to `cudaHostRegister`** (i.e. `shmem_ptr` / the shmem base), not the device VA. On any system where host and device addresses differ (the general non-UVA case `cudaHostGetDevicePointer` exists for), this call returns `cudaErrorInvalidValue`, the error is swallowed by `let _ =`, and the pinned registration is still leaked while `_shmem` is munmap'd underneath it — the exact UB this commit set out to fix. On UVA systems host==device so it happens to work, which would mask the bug on common dev setups.

Compare the sender-side cleanup at ~1844, which correctly passes the host base (`slot.base`). `RecvGpuSlot` currently doesn't store the host base, so the fix likely needs to either add a host-base field or unregister via `slot._shmem.as_ptr()`.

The shape/size validation and the namespaced orphan sweep both look sound.

---

---

phil-opp:
**Automated review by Claude** — this is a fully automated review; no human has vetted it. Treat findings as suggestions to verify.

I re-reviewed the new commit `ccaa961` (the only change since the last review). The fix looks correct: `RecvGpuSlot` now carries `host_base` (the original host pointer passed to `cudaHostRegister`, set at construction), and `free_memory_pool` passes `slot.host_base` to `_unregister_host` instead of the device VA from `cudaHostGetDevicePointer`. That resolves the host-vs-device pointer issue flagged previously. No new memory-safety problems in this commit.

One pre-existing issue I'll note as a non-blocking follow-up (not introduced by this commit):

- In `write_memory_pool`, the **slow path** (`apis/python/node/src/lib.rs`, around line 1587) branches only on `is_cuda` (the *sender* device). It's missing the `ipc_present == 1 && !is_cuda` → `dma_copy` arm that the fast path has (line ~1463). So for a CPU-sender → CUDA-receiver pool that falls back to the slow path, it does a plain `copy_nonoverlapping` into the host shmem data region instead of DMA-ing into the GPU IPC buffer the receiver reads from. This isn't a memory-safety bug — it only produces stale GPU bytes on the rare fallback path — but the two write paths handling the `ipc_present` case differently is a latent correctness gap worth closing.

Otherwise the commit looks safe to merge from a correctness standpoint.

---

phil-opp:
**Re: new commit `7af5691` ("fix: add missing dma_copy arm to write_memory_pool slow path")**

This commit resolves the issue flagged in the previous automated review — the `write_memory_pool` *slow* path branched only on `is_cuda` and was missing the `ipc_present == 1 && !is_cuda → dma_copy` arm that the fast path has, so a CPU-sender → CUDA-receiver pool that fell back to the slow path would `copy_nonoverlapping` into host shmem instead of DMA-ing into the GPU IPC buffer the receiver reads (stale GPU bytes on the fallback path).

Verified at the current head: `apis/python/node/src/lib.rs` (~lines 1591–1614) now adds that arm, deriving the counter via `buffer_id.rsplit_once('_')` and calling `dma_copy(...)` inside the same seqlock begin/end as the fast path. The slow path now mirrors the fast path's three-way branch — the fix looks correct.

One residual hardening note (low severity, **pre-existing** — not introduced by this commit, so not blocking): the slow paths in `write_memory_pool` (~line 1569) and `read_memory_pool` (~line 1769) read the magic and then `read_header_u64(ptr.add(16))` / `read(ptr.add(24))` from a freshly-opened segment **without** first checking `shmem.len() >= DORADMA_HEADER_SIZE`, whereas `try_doradma_read` (~line 2130) does guard this. It's only reachable for a corrupted or size-mismatched same-named segment on the daemon-tracked path, so exposure is limited, but mirroring the fast-path length check at both sites would close it.

No other new findings since the last review.

---
🤖 _This is an automated review by Claude (fully automated, no human in the loop). It may contain mistakes — please verify independently._

---

phil-opp:
🤖 **Automated review by Claude** — _this is a fully automated review with no human in the loop; please sanity-check before acting on it._

Re-checked the new commit `362187d` ("add HEADER_SIZE guard to slow-path read/write before pointer ops") since the previous review. It correctly closes the residual note from last time: both slow paths now bail out when `shmem.len() < DORADMA_HEADER_SIZE` before doing any pointer arithmetic — write path at `apis/python/node/src/lib.rs` (~L1196) and read path (~L1402) — and the guard is placed ahead of the magic / `data_offset` / `ipc_present` reads, so the bounds check is complete. No new issues found in this commit; the latest changes look safe to merge.

(Note for the record, not blockers — all by-design or latent: the long-lived receiver tensor is only protected from torn data by the application-level turn-based `next_require` discipline, not by the seqlock itself; and `effective_as_cuda` defaults a missing/empty `pinned_type` to the CUDA branch. Both are already constrained by the current code paths.)

---

tang-canran:
Hi @heyong4725 ,
  Hope you're doing well. I wanted to give you a quick summary of what's changed since your last review (head 292c400), so you don't have to piece it together from the commit log.

  Your three issues resolved:
  - H1 slow-path CPU read UAF — mirrored the fast-path fix, returning cached.base + data_offset on a cache hit
  - H2 CUDA slow-path host pointer labeled device="cuda" — slow path now bails for CUDA pools instead of returning a host pointer as GPU
  - H3 global orphan sweep — removed the unconditional startup sweep; added a dataflow-namespaced sweep in spawn_dataflow that only cleans up segments belonging to the same dataflow UUID
  
  Issues found and fixed by automated review since then:
  - GPU shape/size validation — tensor_from_info CUDA path now checks product(shape) × itemsize ≤ registered_size
  - Receiver-side cudaHostUnregister — added paired cleanup for cudaHostRegister on the GPU VA path, including the host_base vs device VA pointer fix
  - tmpfs leak — closed via dataflow-namespaced orphan sweep (addresses your S1)
  - Read-path integer overflow — bounds checks now use saturating_sub consistently
  - Missing dma_copy arm in write_memory_pool slow path — the slow path now mirrors the fast path's three-way branch
  - HEADER_SIZE guard on slow paths — both slow paths now reject segments smaller than the header before pointer arithmetic

  The automated review (phil-opp) ran again after the latest commit (362187d) and found no new issues. All six example scenarios and 7 unit tests pass, clippy/fmt clean.

  Let me know if there's anything else you'd like me to address. Thanks for your time on this.

  Best,
  Canran

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop; please weigh it accordingly.

I went through the current head (`362187d`) end-to-end (the `dora-memory-pool` crate, the daemon protocol wiring, and the full Python memory-pool surface). The heavy items from earlier rounds all check out as fixed in the code: the use-after-frees, the integer-overflow bounds checks, path traversal, the host-vs-device unregister pointer, and the slow-path `DORADMA_HEADER_SIZE` guards. The new `DaemonRequest`/`DaemonReply` variants are appended at the enum tail and classified correctly, so there's no serialization/compat break.

One issue I haven't seen raised on the thread yet:

**Sender-side GPU buffer leak on the `receiver_is_cuda` path.** On registration, `dma_copy` allocates a pooled GPU buffer via `_get_gpu_buf`/`cudaMalloc`, keyed by `pool_counter`, into `_gpu_bufs` (`dora/cuda.py`, ~L301-312). But `free_memory_pool` (`apis/python/node/src/lib.rs`, ~L1875-1893) only unmaps the host shmem and calls `_unregister_host` — it never `cudaFree`s that sender-side GPU buffer. So across repeated register/free cycles the sender leaks GPU DRAM. It isn't memory-unsafe, and the CUDA path isn't exercised in CI, so it's untested — but it's a real leak on exactly the high-throughput register/overwrite churn this feature is built for. Worth either releasing the `_gpu_bufs[pool_counter]` entry inside `free_memory_pool`, or documenting why it's intentionally retained.

Nothing else new from me. The remaining low-severity notes already on the thread — `effective_as_cuda` defaulting a missing/empty `pinned_type` to the CUDA branch, and pool entries lingering in the daemon-wide table until daemon shutdown — look acceptable as-is given the current call paths.

---

heyong4725:
## Pre-Landing Review — pinned memory-pool transport

Reviewed the full diff (base `main`, merge-base `7a35064e3`). Scope is clean: the change does what the title says (opt-in pinned-memory transport for large fixed-shape CPU/CUDA tensors) with no scope creep. `dora-memory-pool` tests pass (7/7) and clippy on the daemon + crate is clean. The many prior review rounds handled the easy stuff well (poison-safe locks, header bounds checks gated on `shmem_size >= HEADER_SIZE`, null-ptr checks, path-traversal guard on unlink, CUDA shape/size validation in `tensor_from_info`). What's left are a few real issues.

### P1 — Cross-dataflow collision: a free in one dataflow unlinks another dataflow's shared memory

The daemon's `MemoryPoolManager` is a single table on `Daemon`, shared across **all** dataflows that daemon runs. Its key is the node's `buffer_id = format!("pool_{}_{}", node_id, pool_counter)` (`apis/python/node/src/lib.rs:1282`), where `pool_counter` is a per-process static starting at 1. That key carries **no dataflow_id** — even though the on-disk segment name does (`dora_pool_{dataflow_id}_{node_id}_{counter}`).

Two dataflows on the same daemon, each with a node named `sender_node` (e.g. running the example twice, or any two flows sharing a node name like `camera`), both produce `pool_sender_node_1`. Consequences:

- **Second register** returns `Err("already registered")`; the node only logs a warning (`apis/python/node/src/lib.rs:1341`) and continues, so dataflow #2 loses daemon-side tracking.
- **Free is worse**: dataflow #2's receiver calling `free_memory_pool("pool_sender_node_1")` removes dataflow #1's entry and unlinks **#1's** segment via `entry.metadata.shared_memory_name` (`libraries/extensions/memory-pool/src/lib.rs:164-167`). The prefix guard at line 176 checks `dora_pool_` but not the dataflow_id, so it unlinks the wrong dataflow's `/dev/shm` file.

The dataflow-scoped fast path masks this for reads, but the free path actively corrupts an unrelated dataflow.

**Fix is clean and local:** `dataflow_id: DataflowId` is already in scope at the handler (`binaries/daemon/src/lib.rs:3337`). Build the `MemoryPoolId` from `(dataflow_id, shared_memory_id)` in all three handlers (Register/Read/Free). *(confidence 8/10)*

### P2 — No cap on pool registrations → unbounded daemon table + /dev/shm growth

`register_memory_pool` (`libraries/extensions/memory-pool/src/lib.rs:83-104`) inserts unconditionally; nothing bounds `memory_pool_table`. A node that registers without freeing grows the daemon's table and tmpfs usage until shutdown's `cleanup_all`. `table_size()` exists but is never checked. Add a per-daemon (and ideally per-node) cap in the Register handler. *(confidence 8/10)*

### P2 — Daemon-side validation gaps for non-Python / direct clients

`pool_metadata_from_params` defaults every field; the Register handler re-checks `ptr`/`size`/`dtype`/`shape` but (a) does not require a non-empty `shared_memory_name` — a `None`/empty name silently skips cleanup (`libraries/extensions/memory-pool/src/lib.rs:165`) and leaks the segment, and (b) does not bound `size`. The Python helper caps size at 1 GiB (`apis/python/node/src/lib.rs:1128`), but a node speaking the daemon protocol directly has no such cap. Mirror the size cap and require `shared_memory_name` daemon-side. *(confidence 6/10)*

### P3 — GPU buffer leak on free

The pooled `cudaMalloc` buffer in the embedded helper's `_gpu_bufs` (keyed by counter, `apis/python/node/src/lib.rs:311`) is never freed. `free_memory_pool` calls `_unregister_host` / `_ipc_close` but never frees the device allocation, so register→free churn leaks GPU memory for the process lifetime. Add a `_free_gpu_buf(slot)` and call it from `free_memory_pool`. *(confidence 6/10)*

### P3 — Seqlock placement doesn't match its doc claim (not currently reachable)

In `try_doradma_read` the begin-generation is read at `apis/python/node/src/lib.rs:2235`, *after* `json_len`/`data_offset`/JSON are parsed and bounds-checked (2165-2221); in `register_memory_pool` the header is written while gen is still even (1186-1200), going odd only at 1207. The doc says the seqlock "protects metadata integrity" — structurally it doesn't.

**However, this is not exploitable under the actual protocol:** header `json_len`/`data_offset` are written once at registration *before* the `buffer_id` is ever published to a reader, and `write_memory_pool` only rewrites the data region (never offsets 8/16), so no reader can observe a torn header. The end-of-read gen re-check (line 2343) does correctly guard steady-state data overwrites. Worth a comment / doc correction rather than a code change. *(confidence 7/10 it's an ordering smell; high confidence it is not a live bug)*

### Checked and cleared (not issues)

- **dtype/shape OOB on the consumer**: already mitigated — the CUDA path validates `product(shape) * itemsize <= size` (`apis/python/node/dora/cuda.py:402-409`) and the CPU path is constrained by numpy `reshape` requiring an exact element count.
- **Torn metadata reads via the seqlock**: not reachable under the documented + example-enforced turn-based protocol (see P3 above).

---

**Recommendation:** The P1 cross-dataflow key is the one I'd treat as blocking — it's a real isolation bug with a trivial, local fix (dataflow-scope the table key) that otherwise silently corrupts a co-tenant dataflow. The P2/P3 items are reasonable hardening follow-ups; since the feature is opt-in, experimental, and Linux/CUDA-only, they could land separately if you'd rather not grow this PR further.

phil-opp:
> 🤖 _Fully automated review by Claude — no human in the loop. Findings are based only on the code and the diff, not the PR description or prior comments. Treat accordingly._

Re-reviewed after the latest commits (through `362187d`). The earlier blocking/should-fix items all look addressed (seqlock, read-path bounds checks, `ptr==0` guards, unit tests + `cpu2cpu.yml`, orphan sweep, IPC close / host-unregister, typed metadata helpers, header-layout constants), along with the follow-on regression fixes (counter-parse UAFs, GPU shape validation, host-pointer-vs-VA fix, slow-path `dma_copy` arm). Two things still look worth a pass before merge:

**1. `write_memory_pool` fast path is missing the `DORADMA_HEADER_SIZE` guard (`apis/python/node/src/lib.rs`, ~L1061).** The last commit (`362187d`) added `len < DORADMA_HEADER_SIZE` guards to the two *slow* paths, and `try_doradma_read` has the same guard — but the **fast write path** still validates only the 8-byte magic (`from_raw_parts(shmem_ptr, 8)`) before reading header fields at offsets 16/24 and writing the seqlock at offset 96. The cache-miss arm opens an arbitrary `dora_pool_{dataflow}_{node}_{counter}` segment by name via `ShmemConf::open()`; a leftover/truncated same-named segment of 8–255 bytes that begins with `DORADMA\0` produces an OOB read at `+16/+24` and an OOB **write** at `+96` (`write_volatile`). Lower likelihood than the slow paths (the writer normally created the segment at full size), but it's the same bug class and deserves the same guard for parity.

**2. Non-Linux resource leak.** `free_shared_memory` returns `Err` on non-Linux and `cleanup_orphans` is a no-op there, but Python `register_memory_pool` creates the segment via `ShmemConf` (which works on macOS/Windows) with `set_owner(false)`. So on a non-Linux host every pool segment is created and handed out but never reclaimed — `free_memory_pool`/`cleanup_all` only log the platform error. Failing fast at `register_memory_pool` on unsupported platforms would be safer than silently leaking every segment.

Non-blocking: data-byte safety still rests entirely on callers honoring the turn-based `next_require` round-trip — the seqlock guards only the metadata, and the slow-path CPU cache stores `generation: 0` and never re-checks. That's documented, but it's a sharp edge for any consumer written outside the bundled example.

---

phil-opp:
Re-reviewed at head `5b0dd63` ("cross-dataflow isolation, GPU buffer leak, HEADER_SIZE guard, caps") — the only change since the previous review (`362187d`).

Both items from the last review are resolved in code:
- ✅ **`write_memory_pool` fast-path `DORADMA_HEADER_SIZE` guard** — the cache-hit and cache-miss arms now bail (`if shmem_capacity < DORADMA_HEADER_SIZE`) before any header read at offset 16/24 or the seqlock write at offset 96, matching the slow paths and `try_doradma_read`.
- ✅ **Non-Linux resource leak** — `register_memory_pool` now fails fast on `cfg!(not(target_os = "linux"))` before creating any segment, so unsupported platforms no longer create-and-leak shmem.

The other changes in this commit also check out: `MemoryPoolId` is now `dataflow_id`-scoped in all three daemon handlers (no cross-dataflow collision on same node name); the sender-side `cudaMalloc` buffer is released via `_free_gpu_buf` in `free_memory_pool` (closes the register/free churn leak); and the daemon register handler now caps size at 1 GiB, requires a prefix-validated `shared_memory_name`, and caps the table at 512 pools.

I re-checked the unsafe/pointer paths end-to-end (shmem-name construction consistency across register/write/read, receiver host-unregister using `host_base`, saturating bounds checks, seqlock re-read) and found no new issues. The only remaining sharp edge is the documented one: data-byte consistency relies entirely on callers honoring the turn-based `next_require` round-trip — fine for the bundled example, but worth a prominent caveat for any consumer written directly against this API.

---
*🤖 This is a fully automated review by Claude (Claude Code). No human has reviewed these comments. Please treat findings as suggestions to verify, not authoritative conclusions.*

---

phil-opp:
Short follow-up: the only change since the last review (`5b0dd63`) is `ba4f8580`, which is docstring-only — it adds prominent `# Concurrency / safety` sections to `register_memory_pool`, `write_memory_pool`, and `read_memory_pool` documenting the turn-based-discipline requirement (the seqlock detects in-flight overwrites but does not block a writer from overwriting a tensor a consumer still holds). No code paths change (57 insertions / 5 deletions, all comments in `apis/python/node/src/lib.rs`).

This directly addresses the one remaining sharp edge flagged previously, and I found no new issues — the latest commit is safe.

---

---

tang-canran:
## Summary of heyong4725's Pre-Landing Review

  @heyong4725 raised 5 items in the pre-landing review (P1–P3). All are resolved in 5b0dd63 and confirmed by @phil-opp in subsequent reviews.

  ### P1 — Cross-dataflow collision ✅

  **Problem**: `MemoryPoolId` used `pool_{node_id}_{counter}` as the daemon table key, which carries no `dataflow_id`. Two dataflows on the same daemon with a node named `sender_node`
  would both produce `pool_sender_node_1`, causing the second register to fail silently and `free` to unlink the wrong dataflow's `/dev/shm` segment.

  **Fix** (5b0dd63): Scoped `MemoryPoolId` by `(dataflow_id, shared_memory_id)` in all three daemon handlers (Register/Read/Free). The shmem name already included the dataflow_id 
  (`dora_pool_{dataflow_id}_{node_id}_{counter}`); the table key now matches.

  ### P2 — No cap on pool registrations ✅

  **Problem**: `register_memory_pool` inserted unconditionally — no bound on `memory_pool_table` or `/dev/shm` usage until shutdown's `cleanup_all`.

  **Fix** (5b0dd63): The daemon Register handler now caps the table at 512 pools.

  ### P2 — Daemon-side validation gaps for non-Python clients ✅

  **Problem**: `pool_metadata_from_params` defaults every field. The Python helper caps size at 1 GiB and ensures a valid `shared_memory_name`, but a client speaking the daemon
  protocol directly had no such guard.

  **Fix** (5b0dd63): The daemon Register handler now enforces a 1 GiB size cap and requires a prefix-validated `shared_memory_name`.

  ### P3 — GPU buffer leak on free ✅

  **Problem**: The pooled `cudaMalloc` buffer in `_gpu_bufs` was never freed. `free_memory_pool` called `_unregister_host` / `_ipc_close` but skipped the device allocation, leaking GPU
  memory on every register→free cycle.

  **Fix** (5b0dd63): Added `_free_gpu_buf(slot)`, called from `free_memory_pool`.

  ### P3 — Seqlock placement vs. doc claim ✅

  **Problem**: `try_doradma_read` reads the begin-generation after parsing `json_len`/`data_offset`/JSON, not before — the doc claimed the seqlock "protects metadata integrity" but
  structurally it doesn't guard header reads.

  **Assessment**: Not exploitable under the actual protocol. Header fields (`json_len`, `data_offset`) are written once at registration before the `buffer_id` is ever published to a
  reader. `write_memory_pool` only rewrites the data region (never offsets 8/16). The end-of-read generation re-check correctly guards steady-state data overwrites.

  **Fix**: Corrected the doc comment to accurately describe what the seqlock protects (data-byte consistency across writes, not header metadata reads). No code change needed — this is
  a comment accuracy fix, not a live bug.

  ---

  
  @phil-opp conducted 11 rounds of review across the PR's lifetime. The final two rounds are the most relevant:
  
  ### Round 10 (5b0dd63) — Full re-review
  
  All P1/P2/P3 items from heyong4725 confirmed resolved. Unsafe/pointer paths re-checked end-to-end. The only remaining sharp edge: the turn-based discipline for data-byte consistency
  (seqlock detects in-flight overwrites but does not block the writer). 
  
  ### Round 11 (ba4f8580) — Follow-up approval
  
  > "This directly addresses the one remaining sharp edge flagged previously, and I found no new issues — the latest commit is safe."
  
  The docstring-only commit (ba4f8580) added prominent `# Concurrency / safety` sections to `register_memory_pool`, `write_memory_pool`, and `read_memory_pool`, documenting the
  turn-based contract. A dedicated §8 "Concurrency Model" section was added to the design document.
  
  ---

  ## Current State
  
  | Gate | Status |
  |------|:---:|
  | @heyong4725 P1–P3 | ✅ All resolved (5b0dd63) |
  | @phil-opp 11 review rounds | ✅ All resolved |
  | `cargo fmt --check` | ✅ |
  | `cargo clippy -D warnings` | ✅ |
  | `cargo test -p dora-memory-pool` | ✅ 7/7 |
  | `unwrap-budget` | ✅ 164/164 |
  | All 7 example scenarios | ✅ |
  | Latest commit | `ba4f8580` |

  All review feedback has been addressed. Happy to make any further adjustments.

Issue #2097, opened 2026-06-10
Feature —— Dora Hub Design


LeonRust:
# Dora Hub Design

> **Status:** Draft  
> **Author:** @LeonRust  
> **Date:** 2026-06-09  
> **Discussion:** https://github.com/orgs/dora-rs/discussions/2095

Provides publishing, discovery, distribution, and verification infrastructure for the Dora node ecosystem.

---

## I. Core Philosophy

Dora Hub is a **multi-registry node registry** responsible for managing Dora node metadata, versions, and distribution packages. Hub is not involved in runtime orchestration — it does not manage how dataflows run, how venvs are built, or how connections are validated. Those are the responsibility of Dora Runtime.

---

## II. Architecture Overview

```
┌─────────────────────────────────────────────────────┐
│                   Web Platform                      │
│  (Node detail pages / Dashboard / Token mgmt / CI)  │
├─────────────────────────────────────────────────────┤
│                  Registry API                       │
│  (Search / Metadata / Versions / Publish / Yank)    │
├─────────────────────────────────────────────────────┤
│                  Storage Layer                      │
│  PostgreSQL (metadata)  +  S3 (artifacts)           │
├─────────────────────────────────────────────────────┤
│                    CI System                        │
│  (GitHub Actions shared workflow / Docker buildx)   │
└─────────────────────────────────────────────────────┘
```

---

## III. CLI Surface

11 subcommands:

### Authentication

```
dora hub login          # Paste API token → write to ~/.dora/auth.json
dora hub logout         # Clear ~/.dora/auth.json
```

### Consumption

```
dora hub search <query>               # Search for nodes
dora hub install <node>@<version>     # Install node into .dora/
dora hub uninstall <node>             # Uninstall node
dora hub info <node>                  # View node details (schema/inputs/outputs/config)
```

### Publishing

```
dora hub publish        # Read node.json from current directory → push to registry
dora hub yank <node>@<version>   # Yank version (block new installs, existing continue)
```

### Maintenance

```
dora hub outdated       # View updatable nodes (based on lock.json)
dora hub update         # Update to latest compatible versions
dora hub list           # List installed nodes
```

---

### 4.1 Multi-registry + default official

```
No prefix = official registry (hub.dora-rs.ai):
  hub: camera@^1.2

With prefix = FQDN:
  hub: hub.my-company.com/lidar@1.0
  hub: hub.other-org.com/filter@2.0
```

### 4.2 Resolution rules

```
[...]

PR #2170, opened 2026-06-12
docs(node-hub): source-by-git distribution + in-repo index (rewrite §8/§10)

heyong4725:
## Summary

Reframes the Dora Hub spec (`docs/plan-node-hub.md`) around **source-by-git distribution** and an **in-repo index**, per the maintainer decision discussed on #2097. The core shift: `hub: name@version` **desugars to the `git:` source machinery dora already ships** (`GitManager` clone at a pinned commit + the node's `build:`), rather than a multi-backend fetch stack. Hub becomes a discovery + typed-contract layer over mechanisms that exist, not a distribution system to build.

This follows from two design calls:

1. **Source, not binary, regardless of language.** dora runs on many heterogeneous/embedded targets (aarch64 Jetsons/Pis, musl/glibc, armv7, CUDA-version coupling), where cross-platform binary distribution is hard and source + compile-per-target is portable and already works. dora itself already does this for `git:` nodes.
2. **One repo, not two.** The node index lives **inside `dora-rs/dora-hub`** (`node-index/`), next to `node-hub/` source — so a node's source and its index entry can land in one PR. To be discoverable, a developer publishes an index entry that *points at* their source wherever it's hosted (in dora-hub, in their own git repo, or — opt-in — a binary URL); the index never hosts bytes it doesn't already have.

## What changed

**§8 — collapsed the 5-backend table to one `source` pointer:**
- **git (default)**: `git: <repo>` + `rev: <commit>` + `subdir:` → reuses `NodeSource::GitBranch` → `GitManager`. One mechanism for every language (Python `build: pip install .` with heavy deps still from PyPI; Rust `cargo build`; C/C++ cmake).
- **binary (opt-in, deferred to a heavy-node need)**: per-platform `url` + `sha256` via the existing `path: <url>` + `dora-download`, with a `fallback-git` so an uncovered arch degrades to a source build instead of a hard wall (the cargo-binstall model).
- Dropped: the `HubBackend` trait, OCI artifact layout, pypi/crates.io/oci as distinct integrations.
- Added **§8.3 build-location** — compile-on-target vs build-on-a-host-and-ship, the one real embedded decision, kept separate from the distribution format.

**§10 — `hub:` desugars to `ResolvedNodeSource::GitCommit`:**
- Multi-daemon *is* the existing `git_sources` flow (only addition: a daemon capability flag for clear "too old" errors).
- Lockfile extends `git_sources` with hub provenance over the pinned commit — no new dependency-closure machinery; same reproducibility model `git:` nodes have today.

**Index location + propagation:** `node-index/` in `dora-hub`, fetched via blobless **sparse** clone so it stays cheap despite the larger repo. Updated §1, §4 (diagram + layers), §7.1/§7.3, §9.1 (publish = "add an entry pointing at your source," no upload step), §11 (security), §14 (roadmap: P2.7 collapses to reusing `GitManager`; P2.8/2.13 binary merged + deferred; P2.3 gains a **path-scoped-CI** requirement so the bot-merged index doesn't gate human-reviewed node source), and Appendix A (D18).

Net **+309 / −234**, but the conceptual surface shrank substantially.

## The one operational note worth reviewing

Co-locating a bot-merged append-only index with human-reviewed node source in one repo requires **path-scoped CI**: PRs touching only `node-index/**` (append-only + owner check) auto-merge; PRs touching `node-hub/**` get human review. Trivial GitHub config, and the model is reversible (the index is a directory that can be split into its own repo later if churn ever demands).

## Open fork left for reviewers

The `binary` form's `fallback-git`: should a missing per-arch binary fall back to a source build automatically (best UX), or fail with "no binary for your platform"? Defaulted to fallback in the spec; flagged as open in §8.2 / open-questions.

## Relationship

Amends the spec landed in #2154; tracked under umbrella #2097. Docs-only.

### DIFF
--- a/docs/plan-node-hub.md
+++ b/docs/plan-node-hub.md
@@ -283,9 +283,14 @@ categories: [ml-inference]    # from the fixed category list, §7.6
 keywords: [vision, detection, yolo]
 
 runtime: python               # python | rust | c | cpp
-entrypoint: dora-yolo         # bare command name — no path separators (§11);
-                              # resolved only inside the node's managed env /
-                              # hub cache, then spawned like `path:` today
+entrypoint: dora-yolo         # what to run, relative to the node's working
+                              # dir (= <clone>/<subdir>). Python: a console
+                              # script on the managed-env PATH (`dora-yolo`).
+                              # Rust/C/C++: the build output, e.g.
+                              # `target/release/dora-yolo` or `build/lidar`.
+                              # Validated relative (no absolute, no `..`); maps
+                              # to the synthesized git node's `path:` (§10.1),
+                              # resolved exactly as a `git:` node's path today.
 platforms: []                 # optional allowlist, e.g. [linux-x86_64,
                               # linux-aarch64, macos-aarch64]; empty = all.
                               # Surfaced in search/info; checked pre-dispatch.
@@ -333,7 +338,7 @@ Field rules:
 |---|---|---|
 | `apiVersion` | yes | manifest schema version, starts at 1 |
 | `name`, `namespace` | yes (name defaultable) | index key is `namespace/name`; pypi names PEP 503-normalized |
-| `runtime`, `entrypoint` | yes | entrypoint is a bare command name (validated) |
+| `runtime`, `entrypoint` | yes | entrypoint is a validated relative path within the working dir — a console script (Python) or build output `target/release/<bin>` (Rust/C/C++); no absolute, no `..` |
 | `inputs`, `outputs` | yes, **may be empty** | sinks have no outputs; sources no inputs (fixes the original proposal's ≥1 requirement) |
 | `inputs.*.type` etc. | recommended | type URNs; omitted = untyped (validation skips) |
 | `platforms`, `dora` | recommended | platform facet for search + pre-dispatch check; dora compat range |
@@ -747,7 +752,7 @@ nodes:
   |---|---|
   | `source.git` + resolved `commit` | `git:` + pinned rev (→ `GitManager` clone) |
   | `source.subdir` | the node's working dir = **`<clone-root>/<subdir>`** (see below) |
-  | manifest `entrypoint` | `path:` (relative to that working dir) |
+  | manifest `entrypoint` | `path:` (relative to that working dir) — a build output like `target/release/<bin>` (Rust/C/C++) resolves via `resolve_path`'s `working_dir.join(path)` branch, a console script via its uv-env/PATH branch, exactly as a `git:` node's `path:` does today |
   | manifest `build` | `build:` (runs in that working dir) |
   | manifest `inputs/outputs` types | `input_types`/`output_types` (§6.2) |
 
@@ -847,7 +852,7 @@ hub_sources:                         # provenance layer over git_sources
 | Version immutability | **machine-enforced** append-only index CI (§7.5): existing version files accept only `yanked` flips and artifact *additions*; anything else needs an index admin. Git history audits everything |
 | Source/artifact integrity | the source is pinned to a **commit hash** in the index entry + lockfile (git's own content-addressing); the opt-in binary form pins `sha256`, re-verified on every run (§8.4). Only the pinned commit/binary is used |
 | Build-time code execution | installing a node runs its `build:` (cargo build scripts, pip install) — arbitrary code at build time, identical to a `git:` node today. `--locked` binds it to the reviewed, commit-pinned source. Heavy binary deps (torch) come as wheels from PyPI, the language ecosystem's trust boundary, unchanged |
-| Malicious manifest fields | `entrypoint` restricted to bare command names resolved inside the managed env/cache (no absolute paths, no traversal — the executed binary cannot escape the pinned artifact); `env:` deny-list for loader-hijack variables (LD_PRELOAD, DYLD_*, PYTHONPATH, PATH) |
+| Malicious manifest fields | `entrypoint` is a validated **relative path resolved within the node's working dir / managed env** — no absolute paths, no `..` traversal, so the executed binary cannot escape the cloned source or downloaded artifact (a bare console-script name resolves in the managed env; a relative build-output path like `target/release/<bin>` resolves under the clone+subdir). `env:` deny-list for loader-hijack variables (LD_PRELOAD, DYLD_*, PYTHONPATH, PATH) |
 | Index integrity | git over HTTPS/SSH; protected branch; bot merge only on rule-conformant diffs; fast-forward-only refresh with rollback warning (§7.3) |
 | Namespace integrity | CI-verified GitHub identity match, reserved list, confusable-name screening, human review for new namespaces (§7.4) |
 | Publisher auth | delegated to git: the source lives in a git repo the author controls (or `dora-hub`), index PRs are GitHub-authenticated — dora stores no credentials. The dora-rs org **requires 2FA** for `dora-hub` write access (= index/source committers) |
@@ -922,10 +927,10 @@ its own issue when its phase opens. Items are unowned until claimed —
 - **P1.1** `dora-node-manifest` parsing + JSON Schema in `dora-core`
   (serde types, schema generation alongside `dora-schema.json`).
 - **P1.2** `dora hub init` scaffolding + `dora new` integration. *Depends: P1.1.*
-- **P1.3** Manifest validation: entrypoint bare-name rule, env deny-list,
-  type-URN resolution against `TypeRegistry`, port sanity. Core of
-  `dora hub publish --dry-run` and `dora validate --node-manifest <file>`.
-  *Depends: P1.1.*
+- **P1.3** Manifest validation: entrypoint is a relative path within the
+  working dir (no absolute, no `..`), env deny-list, type-URN resolution
+  against `TypeRegistry`, port sanity. Core of `dora hub publish --dry-run`
+  and `dora validate --node-manifest <file>`. *Depends: P1.1.*
 - **P1.4** Manifest→descriptor contract injection, including for `path:`
   nodes with an adjacent `dora-node.yml` (§6.2; testable without any index).
   *Depends: P1.1.*
--- a/docs/plan-node-hub.md
+++ b/docs/plan-node-hub.md
@@ -88,9 +88,10 @@ parallel machinery:
   index is a git repo; everything a hosted service would add (search API,
   download stats, web UI) is Phase 4 and optional — built *on top of* the
   same index data, never replacing it.
-- **A dora-specific archive format.** Wheels, crates, OCI artifacts, and
-  checksummed release assets already exist; dora pins them, it does not host
-  them.
+- **A dora-specific archive format, or hosting node bytes.** The source
+  already lives in git; dora pins a commit and clones it (or, for the opt-in
+  binary form, downloads a checksummed artifact). The index never hosts the
+  bytes.
 - **Runtime QoS / shape enforcement.** The manifest reuses the existing type
   system; deepening *enforcement* (per-message validation, tensor shapes,
   metadata contracts at runtime) is a parallel type-system workstream
@@ -189,8 +190,9 @@ company's existing credentials) gates both repos.
 **UC9 — Migrating the existing node-hub** *(Phase 3)*
 A script generates draft `dora-node.yml` manifests for the ~60 active nodes
 in `dora-rs/dora-hub`; typing the contracts is farmed out as one
-good-first-issue per node (§12). Index entries pin the wheels already on
-PyPI. Nothing is re-published; `build: pip install dora-yolo` keeps working
+good-first-issue per node (§12). Each index entry's `source` points at the
+`dora-hub` repo + the node's release commit + `subdir: node-hub/<name>`.
+Nothing is re-published; `build: pip install dora-yolo` keeps working
 forever, `hub: dora-yolo@^0.5` becomes the better way.
 
 **UC10 — Yank a broken release** *(Phase 3)*
@@ -528,8 +530,9 @@ Publish-via-PR only scales with automation. Index CI enforces, structurally
 
 1. **Append-only**: a PR may *add* new version files. The only permitted
    modifications to an existing version file are (a) flipping `yanked` (plus
-   a reason), and (b) *adding* entries to `dist.artifacts` (late platform
-   wheels). Any other diff to an existing version file fails CI and requires
+   a reason), and (b) for the binary form, *adding* late per-platform entries
+   to `source.binary` (never changing an existing `url`/`sha256`). Any other
+   diff to an existing version file fails CI and requires
    an index admin.
 2. **Owner check**: the PR author must be in the target namespace's
    `package.yml` OWNERS list.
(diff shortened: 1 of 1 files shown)


heyong4725:
Review findings:

1. docs/plan-node-hub.md:65 and docs/plan-node-hub.md:119 still describe the old native-artifact design, which conflicts with the source-by-git rewrite in §8/§10. The goals say publish reuses wheel/crate/OCI/release assets, UC1 installs pinned wheels, UC3 publishes wheel hashes, UC5/UC7 promise hash-verified dependency closures, UC8 talks about Artifactory/Harbor artifacts, and §12 still says migration pins already-published PyPI wheels. That leaves the spec with two incompatible implementation contracts: source-by-git where `build:` owns deps, and registry-artifact distribution with full closure hashes. I would update these acceptance tests/goals to commit-pinned git source plus optional binary, or explicitly keep native registry backends in the roadmap.

2. docs/plan-node-hub.md:717-733 and docs/plan-node-hub.md:748-760 under-specify how `subdir` and the manifest entrypoint reach the daemon while claiming this is byte-for-byte the existing git-node path. Today the existing `git_sources` wire state is only `{ repo, commit_hash }`, and the builder uses the clone root as the git node working directory; custom nodes also require a concrete `path`. For `dora-hub/node-hub/<name>` entries, the daemon must know the node subdir and the entrypoint path or command. The spec should either say the CLI synthesizes a normal git descriptor with `path`/`build`/`deploy.working_dir` before dispatch, or add the needed hub/subdir state to the protocol/build info. As written, official monorepo hub nodes would be cloned at the right commit but built/spawned from the wrong directory or without a path.

3. docs/plan-node-hub.md:557, docs/plan-node-hub.md:638, and docs/plan-node-hub.md:732 treat binary hub entries as an existing `path: <url>` desugaring while also promising sha256 verification on every run. `dora-download` has an `expected_sha256` parameter, but the current URL `path` descriptor has no checksum field and the runtime/daemon call sites pass `None`. If binary hub nodes literally desugar to `path: <url>`, the lockfile sha256 is lost at spawn time and the stated integrity guarantee does not hold. The plan needs to call out new checksum plumbing for binary hub sources, or avoid describing it as the existing `path:` path.

Validation: reviewed the PR diff against current git/build/download code and ran `git diff --check origin/main...HEAD` successfully.

heyong4725:
Thanks — all three findings were valid and self-introduced by the rewrite (I propagated §8/§10 but left the goals/use-cases/migration and over-claimed "zero new code"). Fixed in `bb17d873`:

**1 — internal inconsistency (goals/UC/migration).** Rewrote the stragglers to the source-by-git model:
- Goal 1: publish = pin a commit (no upload). Goal 5: reproduce = rebuild from the pinned commit; dependency-closure pinning now explicitly stated as *the same gap `git:` nodes have today*, not a Hub guarantee.
- UC1: clone the pinned commit + run `build:` (Python deps still from PyPI). UC3: pin the author's repo commit, no PyPI publish required. UC5/UC7: commit-pinned source / pre-clone; closure caveat called out. UC8: private source repo + private index repo (dropped Artifactory/Harbor).
- §12: migration pins each node's commit in `dora-hub/node-hub`, not PyPI wheel hashes.

**2 — `subdir`/entrypoint reaching the daemon.** You're right that `git_sources` is `{repo, commit_hash}` only and the builder uses the clone root as working dir, so the manifest's subdir/entrypoint can't ride the existing protocol. Specced the honest mechanism: **the CLI desugars `hub:` into a concrete git `Node` before dispatch** — `path: <subdir>/<entrypoint>` (relative, so `resolve_path` joins it with no new field), `build:` from the manifest, `git:`+rev pinned. The daemon then sees a normal git node; wire state and `resolve_path` are unchanged. The new code is one CLI-side desugar step (where rev→commit resolution already happens), not a protocol change. Added a desugar mapping table to §10.1 and rewrote §10.2 step 1 around it.

**3 — binary `sha256` lost at spawn.** Correct — the URL `path:` carries no checksum and the call sites pass `None`, so a literal `path: <url>` desugar drops the integrity guarantee. Now called out as **one small addition**: thread the pinned `sha256` into `download_file`'s already-present `expected_sha256`. Scoped to the binary form (P2.8), which is itself deferred until a heavy node needs it. Softened the "reuses the existing path" claim in §8.1/§8.4 accordingly.

`typos` clean; docs-only.

heyong4725:
Re-review findings:

1. docs/plan-node-hub.md:743-747 still loses `source.subdir` for the build step. The new table correctly synthesizes `path: <subdir>/<entrypoint>`, but it maps manifest `build` straight to `build:`. Existing git nodes run the build command from the clone root (`Builder` uses the prepared git clone dir as `node_working_dir`), while the hub examples put official nodes under `node-hub/<name>` and describe Python nodes as `build: pip install .`. For `dora-rs/dora-hub` that would run `pip install .` at the repo root, not in `node-hub/dora-yolo`, so official monorepo hub nodes will build the wrong directory. The spec should make the desugar subdir-aware for build too, e.g. synthesize a build command that runs in `source.subdir`, add an explicit working-dir field, or restrict manifest build commands to be written relative to the clone root.

2. docs/plan-node-hub.md:91, docs/plan-node-hub.md:189-194, and docs/plan-node-hub.md:529-538 still contain stale native-artifact language from the older backend design. The non-goal says Dora pins wheels/crates/OCI artifacts, UC9 says migration index entries pin PyPI wheels, and the index rules still allow additions to `dist.artifacts` with late platform wheels. Those conflict with the updated §8/§10 model where the normal source is git commit + subdir and binary entries use `source.binary` + sha256. I would update these remaining references so the machine-enforced rules and migration acceptance test match the new source-by-git design.

Validation: re-read the updated PR at commit bb17d873, checked the relevant git build path in `libraries/core/src/build/mod.rs`, and ran `git diff --check origin/main...HEAD` successfully.

heyong4725:
Both valid again — fixed in `d9b263a9`.

**1 — `subdir` lost for the build step.** Correct, and this is the spot where the rewrite is genuinely *not* pure reuse — I'd glossed it. Verified in `build/mod.rs`: `build_node_inner` sets `node_working_dir = clone_dir` (the clone root) and runs `build:` there, so a monorepo node's `build: pip install .` would install the repo root, not `node-hub/dora-yolo`. Fixed by making the git source **subdir-aware**: it gains a `subdir` that joins onto the clone dir (`clone_dir.join(subdir)`), and build, env-prep, and spawn all root at `<clone>/<subdir>`; the synthesized `path` becomes the bare `entrypoint`. Specced as the one real addition (a few lines in `build/mod.rs`), and noted it's independently useful — it lets any plain `git:` node target a subdir of a monorepo, which it can't today. Updated the §10.1 desugar table, §10.2 steps 2–3 (the wire change is now "one optional `subdir` field," not "no new wire type"), and P2.7.

**2 — remaining native-artifact stragglers.** Cleared all three:
- §2 non-goal: "Wheels, crates, OCI artifacts… dora pins them" → "the index never hosts node bytes; dora pins a git commit and clones it (or downloads a checksummed binary)."
- UC9: migration entries now pin the `dora-hub` commit + `subdir: node-hub/<name>`, not PyPI wheels.
- §7.5 append-only rule (b): "adding entries to `dist.artifacts` (late platform wheels)" → "adding late per-platform entries to `source.binary` (never changing an existing url/sha256)."

`typos` clean; a full sweep for `wheel`/`dist.artifacts`/`Artifactory`/`OCI`/`closure` stragglers now returns empty. Docs-only.

(Note: an earlier draft of these edits briefly landed on my local `main` by mistake — discarded and reapplied on the branch; nothing wrong reached the PR.)

heyong4725:
Re-review finding:

1. docs/plan-node-hub.md:285-288 / docs/plan-node-hub.md:336 / docs/plan-node-hub.md:604-610 still make non-Python source nodes underspecified. The manifest requires `entrypoint` to be a bare command name with no path separators, and §10.1 maps that directly to the synthesized git node `path:` relative to `<clone>/<subdir>`. That works for Python console scripts installed into the uv env, but the same section says Rust is just `build: cargo build --release` and C/C++ is a cmake build. Those build outputs normally live at paths such as `target/release/<bin>` or `build/<bin>`, which the manifest cannot express under the bare-name rule and which `resolve_path` will not find unless the build also installs/copies the binary into PATH. Since source-by-git for Rust/C/C++ is a core motivation of this rewrite, the spec needs an explicit output contract here: either allow a validated relative `entrypoint_path`, require build commands to install/copy a bare command into the working dir or managed bin dir, or add per-runtime conventions that make the synthesized `path:` resolvable.

Validation: reviewed the latest PR head `d9b263a9`, rechecked the updated subdir/build text, and ran `git diff --check origin/main...HEAD` successfully.

heyong4725:
Valid — fixed in `5b22369e`. The bare-name rule was right for Python console scripts but wrong for compiled languages, and I confirmed it against the existing convention: `examples/rust-dataflow-git/dataflow.yml` already uses `path: target/debug/<bin>`, resolved by `resolve_path`'s `working_dir.join(path)` branch. So a manifest `entrypoint` for a Rust/C/C++ node genuinely needs to be a relative path to the build output, not a bare name.

Fix: **`entrypoint` is a validated relative path within the node's working dir** (= `<clone>/<subdir>`):
- Python → a console script on the managed-env PATH (`dora-yolo`), resolved via `resolve_path`'s uv-env branch.
- Rust/C/C++ → the build output (`target/release/<bin>`, `build/<bin>`), resolved via the `working_dir.join(path)` branch — exactly how a `git:` Rust node's `path:` resolves today.
- Validation: relative only, no absolute, no `..` traversal — so it still cannot escape the cloned source / managed env (the §11 safety property is preserved, just expressed as relative-within-workdir instead of bare-name).

Updated §5 (field + inline comment), the §5 field-rules table, the §10.1 desugar table, the §11 "malicious manifest fields" row, and P1.3. No new resolution code — the synthesized `path:` carries the relative entrypoint and `resolve_path` already handles both shapes. `typos` clean; docs-only.

That closes the source-by-git output contract for all four runtimes — thanks for the three rounds, each caught a real spot where the "pure reuse" framing outran the code.

heyong4725:
Re-review finding:

1. docs/plan-node-hub.md:755 and docs/plan-node-hub.md:855 now say hub entrypoints are resolved exactly like existing git-node `path:` values and also claim the executable cannot escape the node working dir or managed env. Those two statements do not both hold today. `resolve_path` first checks `working_dir.join(path)`, then `uv run which`, then falls back to the ambient system `$PATH` (`libraries/core/src/descriptor/mod.rs:347-365`). So a bare entrypoint such as `dora-yolo` can resolve to a system binary if the managed env is missing, `uv` is unavailable, or the console script was not installed. That violates the security/reproducibility guarantee in §11 and can make a typo run host state instead of the pinned source. The spec should either require hub spawn resolution to disable system-PATH fallback, require Python hub nodes to use uv/managed env before allowing bare entrypoints, or weaken the security row to state the inherited `git:` behavior explicitly.

Validation: reviewed latest head `5b22369e`, checked the local `resolve_path` implementation, and ran `git diff --check origin/main...HEAD` successfully.

heyong4725:
Valid — and a genuine security hole in my framing. Fixed in `6427373f`. Confirmed `resolve_path` (`descriptor/mod.rs:347-365`) resolves `working_dir.join(path)` → `uv run which` → **ambient `$PATH`**, so a bare entrypoint like `dora-yolo` could resolve to a host binary if the managed env lacks it (uv missing, console script not installed, or a typo). That breaks the §11 "cannot escape the pinned source" guarantee.

Fix: **hub spawn resolution disables the ambient-`$PATH` fallback.** A bare entrypoint resolves only within the node's working dir (`<clone>/<subdir>`) or its managed uv env, and **fails loudly otherwise** — it never silently runs host state. Corrected the over-claim that hub entrypoints resolve "exactly like a `git:` node's `path:`" (they do, *minus* the PATH fallback — that's the deliberate difference).

Updated:
- §10.1 desugar table — entrypoint row now states the disabled-`$PATH` confinement.
- §10.2 step 3 — "existing git-node path plus the subdir join, with the ambient-`$PATH` fallback disabled."
- §11 malicious-manifest row — "resolved only within the working dir or managed env; bare entrypoint fails rather than falling through to a host binary."
- P2.7 — now lists **two** real additions: the `subdir` join and the confined resolution mode (drop `which::which($PATH)` for hub spawn).

`typos` clean; docs-only. That's the confinement guarantee actually holding now rather than asserted — thanks for catching it, the PATH fallback would have been a real foot-gun.

heyong4725:
Re-review finding:

1. docs/plan-node-hub.md:769-771 / docs/plan-node-hub.md:797-810 / docs/plan-node-hub.md:964-980 still do not specify how the daemon knows to use the new confined resolver. The spec says the CLI rewrites `hub:` into an ordinary git-sourced `Node`, the coordinator sees it as a normal git node, and the only wire change is optional `subdir`. But the new security rule is hub-specific: hub spawn disables the ambient `$PATH` fallback while plain `git:` nodes keep it. Once `hub:` has been erased before dispatch and the wire only carries `{repo, commit_hash, subdir}`, the daemon has no marker telling it this git node came from Hub and should use confined resolution. Either the synthesized descriptor/wire state needs an explicit `confined_path_resolution`/hub-origin bit, or the confined resolver must apply to all git nodes with the compatibility impact stated. As written, the security guarantee in §11 cannot be implemented from the state that §10.2 sends.

Validation: reviewed latest head `6427373f`, checked the updated confined-resolution text, and ran `git diff --check origin/main...HEAD` successfully.

heyong4725:
Valid — and it exposes an over-claim I made in the previous round. Fixed in `1d85f629`.

The contradiction was real: §10.2 said the daemon sees an "indistinguishable" git node, but §11 needs hub-specific **confined** resolution (no ambient-`$PATH` fallback). If the wire only carries `{repo, commit_hash, subdir}`, the daemon has no signal to disable that fallback — so the §11 guarantee couldn't be implemented from the state §10.2 actually dispatches. "Indistinguishable git node" and "hub-only confined resolution" can't both be true.

Fix — option (a) from your finding: **carry a hub-origin marker on the wire.** The synthesized git source gains an optional `hub: {name, version}` provenance bit alongside `subdir`. The honest wire delta is now:

```
{repo, commit_hash}  ->  {repo, commit_hash, subdir?, hub?}
```

and `resolve_path` gains a **confined mode gated on `hub?`** — the daemon disables the ambient-`$PATH` fallback precisely when the marker is set. Confined resolution is now *derivable from wire state*, not asserted.

Updated:
- **§10.1** — wire-state paragraph now describes **two** new optional bits (`subdir?`, `hub?`), not one; `resolve_path` gains the confined mode. Added a note on the alternative you raised (confine *all* git nodes — dropped because it'd silently break existing `git:` nodes that rely on a PATH-resolved entrypoint).
- **§10.2** — step 2: git source carries `subdir?` + `hub?`; step 3: daemon applies confined resolution **when `hub?` is set**.
- **§11** — malicious-manifest row now names the `hub?` marker as the mechanism the daemon keys on.
- **P2.5 / P2.7** — marker rides the `BuildDataflowNodes` git source so confined resolution is derivable, not asserted.

`typos` clean; docs-only. I also corrected the "indistinguishable git node" framing from round 4 — the wire node is *deliberately* distinguishable now, by exactly one optional marker. Thanks for catching that the security property had no carrier on the wire.

heyong4725:
Re-review finding:

1. docs/plan-node-hub.md:254-261 and docs/plan-node-hub.md:291-293 still contain stale reuse wording that contradicts the updated §10 contract. §10 now correctly says hub git sources need two optional wire bits (`subdir?`, `hub?`) and a confined resolver that disables ambient `$PATH` fallback when `hub?` is set. But the architecture summary still says hub nodes flow through the existing build/spawn/multi-daemon pipeline “unchanged,” and the manifest example says `entrypoint` is resolved “exactly as a `git:` node’s path today.” That reintroduces the same ambiguity the later sections fixed: implementers reading the summary/example could skip the `hub?` marker or use plain `resolve_path`, losing the §11 confinement guarantee. Please update those earlier lines to say “mostly reuses the existing git path, with the subdir/hub marker and confined path resolution described in §10.”

Validation: reviewed latest head `1d85f629`, checked the updated §10/§11 text, and ran `git diff --check origin/main...HEAD` successfully.

heyong4725:
Valid — fixed in `4ed5f87c`. These were exactly the earlier sections the §10/§11 rounds left behind: the summary still said the pipeline runs "unchanged," and the manifest example still said the entrypoint resolves "exactly as a `git:` node's path today" — both of which an implementer could read *instead of* §10 and skip the `hub?` marker / confined resolver, losing the §11 guarantee. So this is a real internal-consistency hole, not just wording.

- **§4, layer 4** — "flows through the existing build/spawn/multi-daemon pipeline **unchanged**" → "**mostly reuses** the existing pipeline — with the two optional wire bits (`subdir?`, `hub?`) and the confined path resolver described in §10."
- **§5, entrypoint comment** — "resolved **exactly as a `git:` node's path today**" → "resolved **like** a `git:` node's path but under **confined resolution** (no ambient-`$PATH` fallback, §10/§11)."

I also swept the rest of the doc for stragglers — the other `unchanged` / `exactly as a git:` hits are all about things that *are* genuinely identical (manifest validation pipeline, build-time code execution, PyPI dep sourcing), not path resolution, so they stay. `typos` clean; docs-only.

That's the summary/example now pointing at §10 rather than contradicting it — thanks; this is the kind of drift that bites an implementer who reads the overview and stops.

heyong4725:
Re-review: no findings.

The latest head `4ed5f87c` resolves the prior stale wording: the architecture summary now says hub mostly reuses the git path while explicitly naming `subdir?`, `hub?`, and confined path resolution, and the manifest example no longer claims exact plain-git `resolve_path` behavior. I re-read §4/§5/§10/§11/roadmap for consistency and did not find another blocking issue.

Validation: ran `git diff --check origin/main...HEAD` successfully.

PR #2174, opened 2026-06-12
feat(hub): node manifest parsing, JSON schema, and validation (P1.1+P1.3)

heyong4725:
Implements **P1.1 + P1.3** of the Dora Hub spec ([docs/plan-node-hub.md](https://github.com/dora-rs/dora/blob/main/docs/plan-node-hub.md), umbrella #2097): the node manifest (`dora-node.yml`) parsing/validation layer that `dora hub publish --dry-run` (P3.1) and manifest→descriptor contract injection (P1.4) will build on. No network, no index — immediately useful standalone.

## What's included

- **`dora_core::manifest`** — `NodeManifest` serde types matching spec §5 (contracts, entrypoint, env surface, shipped custom types, requirements, example snippet), with a 1 MiB input cap for untrusted index content.
- **Validation (spec §5/§11)** — returns all problems with field paths:
  - entrypoint: relative-path confinement (no absolute, no `..`, no `~`/drive letters) via a conservative charset that also excludes whitespace/shell metacharacters/control chars;
  - env deny-list: `PATH`, `PYTHONPATH`/`PYTHONHOME`/`PYTHONSTARTUP`/`PYTHONEXECUTABLE`/`PYTHONBREAKPOINT`, `VIRTUAL_ENV`, `BASH_ENV`, `ENV`, `IFS`, and `LD_*`/`DYLD_*` prefixes, case-insensitive; names must be valid env identifiers (closes whitespace-padding bypasses);
  - type URNs resolve against the `TypeRegistry` or the manifest's own `types:` block, with typo suggestions; shipped types must live under the package namespace, `std/` rejected;
  - port names validated with the runtime's own `DataId` rules; `required` rejected on outputs; platform/semver/example checks;
  - issue rendering strips control characters (hostile manifests can't inject terminal escapes).
- **JSON Schema** — `dora-node-schema.json` generated alongside `dora-schema.json`; a unit test fails if the checked-in schema drifts from the types, and the regenerate-schemas workflow now covers both files.
- **CLI** — `dora validate --node-manifest <file>` (mutually exclusive with the dataflow argument).
- `semver` promoted to a workspace dependency (spec §7.2 note).
- `.gitignore`: dropped the distutils `MANIFEST` pattern — on case-insensitive filesystems it silently hides Rust `manifest/` module dirs from git, and nothing in this repo produces a distutils `MANIFEST`.

## Verification

- `cargo test -p dora-core manifest` — 26 tests (parsing, validation rules, bypass attempts, schema drift).
- `dora validate --node-manifest` live-tested against a clean manifest (OK, exit 0) and a hostile one (4 errors incl. typo suggestion, exit 1).
- fmt + clippy `-D warnings` + full workspace test suite + `cargo check --examples`.

## Notes for reviewers

- Lexical entrypoint validation deliberately does **not** claim to stop symlink escapes — that requires canonicalization at spawn time, which is the confined-resolution work in P2.7 (tracked in the spec §10.1/§11).
- The validate-once-run-anywhere stance means `..` is rejected with either separator regardless of host platform; a Unix file literally named `a\..\b` is conservatively rejected.

Part of #2097.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,11 +115,16 @@ commands:
           command: |
             uv venv --seed -p 3.12
             echo "export VIRTUAL_ENV=$PWD/.venv" >> "$BASH_ENV"
+            # Dora's example nodes spawn `uv run python` from nested
+            # directories; pin uv to the CI venv that contains the editable
+            # dora package instead of relying on cwd-based discovery.
+            echo "export UV_PYTHON=$PWD/.venv/bin/python" >> "$BASH_ENV"
             echo "export PATH=$PWD/.venv/bin:\$PATH" >> "$BASH_ENV"
             source "$BASH_ENV"
             uv pip install pyarrow
             uv pip install -e apis/python/node
             uv pip install ruff pytest
+            uv run python -c "import dora; print(dora.__file__)"
 
 # ────────────────────────────────────────────────────────────────────
 # Executors
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,10 +115,6 @@ commands:
           command: |
             uv venv --seed -p 3.12
             echo "export VIRTUAL_ENV=$PWD/.venv" >> "$BASH_ENV"
-            # Dora's example nodes spawn `uv run python` from nested
-            # directories; pin uv to the CI venv that contains the editable
-            # dora package instead of relying on cwd-based discovery.
-            echo "export UV_PYTHON=$PWD/.venv/bin/python" >> "$BASH_ENV"
             echo "export PATH=$PWD/.venv/bin:\$PATH" >> "$BASH_ENV"
             source "$BASH_ENV"
             uv pip install pyarrow
@@ -857,6 +853,16 @@ jobs:
             cp target/debug/dora ~/.cargo/bin/dora
             chmod +x ~/.cargo/bin/dora
       - setup-python-uv
+      - run:
+          name: Pin uv to the CI venv for the contract examples
+          # The contract examples spawn `uv run python` from nested example
+          # dirs; pin uv to the venv that holds the editable dora package
+          # rather than cwd-based discovery. Scoped to this job on purpose —
+          # cli-python's template-project nodes need their own per-project env
+          # resolution and break under a global UV_PYTHON pin.
+          command: |
+            echo "export UV_PYTHON=$PWD/.venv/bin/python" >> "$BASH_ENV"
+            source "$BASH_ENV"
       - run:
           name: Run semantic contract tests
           command: >
--- a/.github/workflows/regenerate-schemas.yml
+++ b/.github/workflows/regenerate-schemas.yml
@@ -25,11 +25,11 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          if git diff --exit-code -- libraries/core/dora-schema.json; then
+          if git diff --exit-code -- libraries/core/dora-schema.json libraries/core/dora-node-schema.json; then
               echo "Schema file was not changed"
           else
               git switch -c schema-update-for-${{ github.sha }}
-              git add libraries/core/dora-schema.json
+              git add libraries/core/dora-schema.json libraries/core/dora-node-schema.json
               git config user.email "dora-bot@phil-opp.com"
               git config user.name "Dora Bot"
               git commit -m "Update JSON schema for ${{ github.sha }}"
--- a/.gitignore
+++ b/.gitignore
@@ -68,7 +68,9 @@ share/python-wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-MANIFEST
+# (the distutils `MANIFEST` pattern was removed here: nothing in this repo
+# produces one, and on case-insensitive filesystems it silently hides Rust
+# `manifest/` module directories from git)
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1822,6 +1822,7 @@ dependencies = [
  "itertools 0.14.0",
  "log",
  "schemars 1.2.1",
+ "semver",
  "serde",
  "serde-with-expand-env",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,6 +135,7 @@ git2 = { version = "0.20.4", features = ["vendored-openssl"] }
 serde_yaml = "0.9.33"
 zenoh = { version = "~1.8", default-features = false, features = ["shared-memory", "unstable", "transport_tcp", "transport_udp", "transport_unixsock-stream"] }
 serde_json = "1.0.145"
+semver = { version = "1.0.23", features = ["serde"] }
 thiserror = "2.0"
 colored = "3.0.0"
 inquire = { version = "0.7.5", default-features = false }
--- a/binaries/cli/src/command/mod.rs
+++ b/binaries/cli/src/command/mod.rs
@@ -388,6 +388,21 @@ mod tests {
     }
 
     #[test]
+    fn parse_validate_node_manifest() {
+        parse_ok(&["dora", "validate", "--node-manifest", "dora-node.yml"]);
+        // a dataflow and --node-manifest are mutually exclusive
+        parse_err(&[
+            "dora",
+            "validate",
+            "dataflow.yml",
+            "--node-manifest",
+            "dora-node.yml",
+        ]);
+        // one of the two is required
+        parse_err(&["dora", "validate"]);
+    }
+
+    #[test]
     fn parse_new() {
         parse_ok(&["dora", "new", "test"]);
     }
--- a/binaries/cli/src/command/mod.rs
+++ b/binaries/cli/src/command/mod.rs
@@ -400,6 +400,14 @@ mod tests {
         ]);
         // one of the two is required
         parse_err(&["dora", "validate"]);
+        // --strict-types is dataflow-only
+        parse_err(&[
+            "dora",
+            "validate",
+            "--node-manifest",
+            "dora-node.yml",
+            "--strict-types",
+        ]);
     }
 
     #[test]
--- a/binaries/cli/src/command/validate.rs
+++ b/binaries/cli/src/command/validate.rs
@@ -11,7 +11,7 @@ use eyre::{Context, bail};
 use std::path::{Path, PathBuf};
 
 #[derive(Debug, clap::Args)]
-/// Validate a dataflow YAML file and check type annotations
+/// Validate a dataflow YAML file (or a node manifest) and check type annotations
 pub struct Validate {
     /// Path to the dataflow descriptor file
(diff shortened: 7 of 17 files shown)


heyong4725:
**Self-review round 2 (post-open):** three fresh correctness-angle reviews (line-by-line, removed-behavior audit, cross-file trace) surfaced 6 real issues, fixed in the follow-up commit:

1. `EnvVarDef.default` reused the descriptor's `EnvValue`, whose deserializer expands `$VAR` against the local environment at parse time — validation was nondeterministic across machines and could bake the publisher's local env into a published manifest. Replaced with a literal `EnvDefault` enum.
2. Shipped `types:` keys were not format-checked, and a parameterized key could never be resolved by the port checker (it strips params before lookup). Keys are now charset-checked and parameterized declarations rejected; parameterized *references* to a declared base still resolve.
3. The output sanitizer kept `\n`, letting a hostile manifest spoof extra output lines (e.g. a fake "Node manifest OK."). Newlines are now stripped too.
4. The `read()` size cap used `fs::metadata`, which reports 0 for FIFOs/devices — now bounds the read itself with `take(MAX+1)`.
5. Declared env `type:` is now cross-checked against `default:`.
6. `--strict-types` was silently ignored with `--node-manifest`; now a clap conflict. Help text and `docs/cli.md` updated.

The removed-behavior audit verified zero lost invariants in the CLI restructure, the `.gitignore` `MANIFEST` removal (nothing in-repo produces a distutils MANIFEST; the pattern was actively hiding `src/manifest/` on case-insensitive filesystems), the semver workspace move (features preserved), and the `types.rs` derive additions (zero schema drift).

phil-opp:

No significant correctness or security issues found. The 1 MiB cap is enforced on the read itself (`take(MAX+1)`, so FIFOs / growing files are bounded rather than relying on `metadata`), the env deny-list runs after a strict `[A-Za-z_][A-Za-z0-9_]*` identifier check (closing whitespace-padding bypasses), URN/namespace confinement and `std/` rejection are correct, and the issue renderer strips control chars *and* newlines.

One minor sharp-edge worth considering: because `EnvDefault` is an untagged enum (variant order Bool → Integer → Float → String), an env var declared `type: string` with an unquoted scalar default such as `default: true` or `default: 123` deserializes to `Bool`/`Integer`, so `env_default_matches` reports "default value does not match declared type `string`". A publisher who actually wants the literal string then has to quote it (`default: "true"`). It's technically correct but surprising — accepting scalar defaults for `type: string` (or documenting the quoting requirement) would avoid the false rejection.

---

heyong4725:
Review finding:

- `libraries/core/src/manifest/validate.rs:189`: manifest-local type resolution only checks that `self.types` contains the referenced base URN, but the corresponding `TypeDef` bodies are never validated. A manifest can declare `types: { dora-rs/lidar/v1/Bad: { arrow: Struct, fields: [{ name: x, type: DefinitelyNotAType }] } }` and then use `outputs.cloud.type: dora-rs/lidar/v1/Bad`; `validate()` accepts the port because the key exists, even though the shipped type cannot be materialized into a usable Arrow schema later (`TypeDef::to_arrow_schema_with_registry` fails closed when fields cannot resolve). That weakens the P1.3 dry-run gate described in `docs/plan-node-hub.md:713` and can let unusable contracts into the hub index. Please validate every shipped type definition during manifest validation, preferably against a registry that includes the manifest-local types, and report invalid field/type definitions before accepting references to them.

Validation run: `cargo test -p dora-core manifest`.

heyong4725:
Thanks — both findings fixed in 6e33dcae2:

1. **Shipped `types:` bodies not validated (validate.rs:189).** Confirmed and fixed: `validate()` now builds a registry that includes the manifest's own `types:`, and checks every shipped type's field types resolve against it (primitive Arrow types, `List<T>`, std types, and sibling shipped types all resolve; a genuinely-unknown field type is rejected). A reference to a type whose body can't be materialized now fails the dry-run gate with `types.<urn>.fields.<name>: unknown field type ...`, rather than passing just because the URN key exists. Added `TypeRegistry::{field_type_resolves, insert_type}` (registry is now `Clone`) and a regression test (`shipped_type_bodies_are_validated`).

2. **`type: string` with an unquoted scalar default (the untagged-enum sharp edge).** Good catch — `default: true` / `default: 8080` deserialized to `Bool`/`Integer` and `env_default_matches` falsely rejected them. Any scalar is now accepted as a string default (the publisher means the literal string), so no quoting is required. Test extended to cover `true`/`8080`/`1.5`.

Validation: `cargo test -p dora-core manifest` (30 passing) + the live `dora validate --node-manifest` cases. The rest of the stack (#2176–#2180) inherits these on rebase.

heyong4725:
**Pre-landing review** (two independent adversarial passes + verification). Fixed in `7002e104`:

- **Shipped-type URN path traversal** (Medium): `check_shipped_type_urn` allowed `.`/`/` and only checked the namespace prefix, so `acme/../../std/core/v1/Foo` validated — keeps the `acme/` prefix yet escapes the namespace. Since shipped types get materialized into cache paths (P2.12), now rejects empty/`.`/`..` segments, the same confinement class the entrypoint check already enforces.
- **Reserved `DORA_*` env** (Low): the deny-list covered loader/interpreter/shell vars but not the framework's `DORA_*` control variables. Added the `DORA_` prefix so a manifest can't advertise `DORA_NODE_CONFIG` etc. as its config surface. (The runtime override is separately gated by the daemon's `ENV_DENYLIST`, which already denies the security-critical `DORA_AUTH_TOKEN`/`DORA_ALLOW_SHELL_NODES`.)

Both have regression tests. Verified non-issues: the **YAML alias-bomb** concern doesn't hold against the typed `NodeManifest` path (exponential bomb type-errors in 0.01s, linear in 0.06s/20MB — it only blows up against untyped `serde_yaml::Value`); the env deny-list resists case/whitespace/Unicode evasion (charset gate first); entrypoint confinement and the `take()`-bounded read path are solid.

Noted follow-ups (out of this PR's scope): daemon `ENV_DENYLIST` could add `DORA_NODE_CONFIG`/`DORA_RUNTIME_CONFIG` (pre-existing, limited impact); type-URN param charset + display-field control chars are sanitized at the render layer in the discovery PR.

The downstream stack (#2176-#2180) was rebased onto this so it inherits the hardening.

phil-opp:

Re-reviewed after the latest commits. One new issue, introduced by the shipped-type-body validation added in `6e33dcae`:

**Unbounded recursion → stack-overflow DoS in `TypeRegistry::field_type_resolves`** (`libraries/core/src/types.rs`). The new helper recurses on `List<…>` with no depth limit:

```rust
if let Some(inner) = t.strip_prefix("List<").and_then(|s| s.strip_suffix('>')) {
    return self.field_type_resolves(inner);
}
```

It runs (via `validate.rs`) on the `type` string of every shipped `types:` field, which is untrusted manifest content and is *not* length/charset-checked first — `check_shipped_type_urn` validates only the URN key, not the field-type strings. A manifest within the 1 MiB cap can ship a field typed `List<List<…>>` nested ~175k deep, overflowing the stack and crashing the validator. That's a DoS in exactly the "hub index validates a published manifest" path this PR is meant to harden.

Note the pre-existing sibling `resolve_field_type` (same file) is deliberately capped at `MAX_TYPE_DEPTH = 8`; the new helper dropped that guard. Suggest applying the same depth bound (or an iterative `List<>` peel with a limit) to `field_type_resolves`.

Everything else in the validation boundary holds up on this pass: entrypoint path confinement, the env identifier-gate + deny-list (including the new `DORA_*` prefix), shipped-type URN traversal rejection, the `take()`-bounded read, and control-char/newline output sanitization. The struct-reference branch of `field_type_resolves` doesn't recurse, so nested `List<>` is the only recursion vector.

---

phil-opp:

The latest push (`a3861e9`) is a CI-only change: it only edits `.circleci/config.yml` (scoping the `UV_PYTHON` pin to contract-tests) and doesn't touch the manifest parsing/validation code. So the previously-reported issue is still present and unaddressed:

`TypeRegistry::field_type_resolves` (`libraries/core/src/types.rs`) recurses through `List<...>` with no depth guard:

```rust
if let Some(inner) = t.strip_prefix("List<").and_then(|s| s.strip_suffix('>')) {
    return self.field_type_resolves(inner);
}
```

unlike the sibling `resolve_field_type`, which caps recursion at `MAX_TYPE_DEPTH = 8`. This runs on untrusted shipped-type strings (only the URN *key* is charset-checked, not the field-type body), and the manifest read is capped at 1 MiB — large enough that a deeply-nested `List<List<…>>` field type can overflow the stack and crash the validator. Applying the same depth bound here would close it.

---

heyong4725:
Good catch — fixed in `7726c388`. `field_type_resolves` recursed through `List<…>` with no depth guard while its sibling `resolve_field_type` caps at `MAX_TYPE_DEPTH`. Since it runs on untrusted shipped-type field bodies and the manifest read is capped at 1 MiB (room for ~85k `List<` nestings), that was a stack-overflow DoS on the validator. Now delegates to a depth-tracked helper and rejects anything past `MAX_TYPE_DEPTH`, matching `resolve_field_type`. Added a regression test that a 100k-deep `List<…>` returns false instead of overflowing. Cascaded to #2176-#2180.

phil-opp:

`7726c388` correctly resolves the `field_type_resolves` stack-overflow DoS I flagged earlier: the `List<…>` recursion is now depth-tracked and capped at `MAX_TYPE_DEPTH`, matching the sibling `resolve_field_type`, with a 100k-deep regression test. No new findings on this commit.

---

heyong4725:
Thanks — triaged all three:

- **`arrow:` not validated** (real, fixed in `916c9df6`): manifest validation checked the URN and field types but never the top-level `arrow:` discriminant, so `arrow: NotAType` passed (the URN exists) and only failed at build/materialization. Now rejects any `arrow:` that isn't a known primitive or `Struct` (the exact set the std library uses and the runtime can build); regression test added.
- **spec `types: []`** (real, fixed): the spec example showed an empty *sequence* while `types` is a *map* — corrected to `types: {}` in `docs/plan-node-hub.md`.
- **inputs/outputs not required in the JSON schema**: this is intentional. The Rust fields default to empty maps, which satisfies the spec's "may be empty"; the impl additionally allows omission (a source/sink simply declares no ports), which is more ergonomic than forcing `inputs: {}` everywhere. Marking them schema-required would only force boilerplate without catching a real error — the wiring check already flags any port a dataflow uses that the manifest doesn't declare. Left as-is by design.

(CI note: the `examples-linux` red is the pre-existing intermittent C++/CMake-example teardown-hang flake, unrelated to this PR — it needs a CircleCI job re-run.) Cascaded to #2176-#2180.

phil-opp:
Re-reviewed after the new commit `916c9df6` (validate the shipped type's top-level `arrow:` discriminant). `is_known_arrow_type` matches exactly the set `resolve_field_type`/`build_schema` can actually materialize (the `arrow_type_from_name` primitives plus `Struct`), so it closes the validate-pass/build-fail gap without opening a new one, and it has a regression test. No new findings — the latest commit looks safe.

---

---

PR #2176, opened 2026-06-12
feat(hub): inject node manifest contracts into path: nodes (P1.4)

heyong4725:
Implements **P1.4** of the Dora Hub spec ([docs/plan-node-hub.md](https://github.com/dora-rs/dora/blob/main/docs/plan-node-hub.md) §6.2, umbrella #2097): a `dora-node.yml` sitting next to a plain `path:` node is picked up by `dora validate` and `dora build`, so local development gets the same compose-time contract checking that hub consumption will get (UC11, UC2).

**Stacked on #2174** (needs the manifest module) — base is `feat/hub-node-manifest`; will retarget to `main` once that merges.

## How it works

- For each plain local-`path:` node, the injector walks up from the executable's path looking for a `dora-node.yml`. A manifest applies **iff** `<manifest dir>/<entrypoint> == <working dir>/<path>` — exact association even when the manifest sits at the node's source root and the binary is at `target/release/<bin>`. The walk never leaves the dataflow's working directory; absolute and `..`-escaping paths are skipped entirely.
- A matched manifest's typed ports are injected as the node's `input_types`/`output_types` annotations (the dataflow author's own annotations win); from there the existing validation pipeline applies unchanged — edge compatibility, field-level struct diffs, inference, `strict_types`.
- The dataflow's wiring is checked against the declared ports (§10.1): undeclared inputs/outputs and unwired required inputs are warnings, fatal under `--strict-types`.
- Custom types shipped in manifests (`types:`) are registered into the `TypeRegistry` in a first pass over all matched manifests, so cross-manifest type references resolve regardless of node order.

Live-tested (UC2): a manifest-typed `bbox` output wired into a consumer expecting `std/media/v1/Image` fails at compose time with the producer/consumer URNs and the field-level diff, before anything runs.

## Hardening (from self-review)

- Walk containment: a hostile or accidental `dora-node.yml` outside the working dir can never be associated with a node.
- All injection notes/warnings are control-character-sanitized (they embed manifest/path content and are printed by the CLI).
- Manifest entrypoints using `\` separators match on every platform.
- Entrypoint mismatch warns rather than silently skipping, so a typo'd `entrypoint` is visible.

## Verification

- 11 new unit tests in `manifest/inject.rs` (injection, author-override, mismatch, containment, cross-manifest types, backslash entrypoints, invalid-manifest skip).
- fmt + clippy `-D warnings` + full workspace tests + `cargo check --examples`.
- Live: `dora validate` and `dora build` on a fixture dataflow — injection note printed, downstream mismatch caught, `--strict-types` exits 1.

Part of #2097.

### DIFF
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -221,7 +221,7 @@ pub fn build(cfg: BuildConfig) -> eyre::Result<()> {
         eyre::bail!("`--lockfile` requires either `--locked` or `--write-lockfile`");
     }
     let working_dir = working_dir_or_parent(working_dir_override.as_deref(), &dataflow_path);
-    let dataflow_descriptor = Descriptor::blocking_read(&dataflow_path)
+    let mut dataflow_descriptor = Descriptor::blocking_read(&dataflow_path)
         .wrap_err_with(|| {
             format!(
                 "failed to read dataflow at `{}`\n\n  \
@@ -247,6 +247,15 @@ pub fn build(cfg: BuildConfig) -> eyre::Result<()> {
             _ => {}
         }
     }
+    // Inject contracts from node manifests adjacent to path: nodes (§6.2)
+    let injection = dora_core::manifest::inject::inject_adjacent_manifests(
+        &mut dataflow_descriptor,
+        working_dir,
+        &mut registry,
+    );
+    for note in &injection.notes {
+        println!("  {note}");
+    }
     let type_result = dora_core::descriptor::validate::check_type_annotations_full(
         &dataflow_descriptor,
         &registry,
@@ -255,16 +264,19 @@ pub fn build(cfg: BuildConfig) -> eyre::Result<()> {
     for inf in &type_result.inferences {
         println!("  {inf}");
     }
-    if !type_result.warnings.is_empty() {
+    let warning_count = injection.warnings.len() + type_result.warnings.len();
+    if warning_count > 0 {
+        for w in &injection.warnings {
+            eprintln!("  warning: {w}");
+        }
         for w in &type_result.warnings {
             eprintln!("  warning: {w}");
         }
-        let count = type_result.warnings.len();
         if strict {
-            eyre::bail!("{count} type error(s) found (strict mode)");
+            eyre::bail!("{warning_count} type error(s) found (strict mode)");
         } else {
             eprintln!(
-                "{count} type warning(s) found.\n  \
+                "{warning_count} type warning(s) found.\n  \
                  hint: use --strict-types to fail on type warnings"
             );
         }
--- a/binaries/cli/src/command/validate.rs
+++ b/binaries/cli/src/command/validate.rs
@@ -4,7 +4,7 @@ use dora_core::{
         Descriptor, DescriptorExt,
         validate::{check_type_annotations_full, check_wiring},
     },
-    manifest::NodeManifest,
+    manifest::{NodeManifest, inject::inject_adjacent_manifests},
     types::TypeRegistry,
 };
 use eyre::{Context, bail};
@@ -63,7 +63,7 @@ fn validate_dataflow(dataflow: &Path, strict_types: bool) -> eyre::Result<()> {
     println!("Validating {}...", dataflow.display());
 
     // Parse and expand modules (no runtime needed)
-    let descriptor = Descriptor::blocking_read(dataflow)
+    let mut descriptor = Descriptor::blocking_read(dataflow)
         .with_context(|| {
             format!(
                 "failed to read dataflow at `{}`\n\n  \
@@ -95,6 +95,12 @@ fn validate_dataflow(dataflow: &Path, strict_types: bool) -> eyre::Result<()> {
         }
     }
 
+    // Inject contracts from node manifests adjacent to path: nodes (§6.2)
+    let injection = inject_adjacent_manifests(&mut descriptor, working_dir, &mut registry);
+    for note in &injection.notes {
+        println!("  {note}");
+    }
+
     // Run type annotation checks
     let result = check_type_annotations_full(&descriptor, &registry, strict);
 
@@ -103,14 +109,17 @@ fn validate_dataflow(dataflow: &Path, strict_types: bool) -> eyre::Result<()> {
         println!("  {inf}");
     }
 
-    if result.warnings.is_empty() {
+    let count = injection.warnings.len() + result.warnings.len();
+    if count == 0 {
         println!("All type annotations OK.");
     } else {
         println!("Type warnings:");
+        for w in &injection.warnings {
+            println!("  - {w}");
+        }
         for w in &result.warnings {
             println!("  - {w}");
         }
-        let count = result.warnings.len();
         println!("{count} type warning(s) found.");
         if strict {
             bail!("{count} type warning(s) found (--strict mode)");
--- a/libraries/core/src/manifest/inject.rs
+++ b/libraries/core/src/manifest/inject.rs
@@ -6,6 +6,11 @@
 //! own annotations win), its custom types are registered, and the dataflow's
 //! wiring is checked against the declared ports. Local development thereby
 //! gets the same contract validation as hub consumption (UC11).
+//!
+//! Scope (P1.4): injection runs in `dora build` and `dora validate` only —
+//! compose-time checking. The injected annotations do not reach the runtime
+//! first-message type check or `dora graph`, which read the descriptor
+//! through paths that don't inject (deliberate for now; see plan §6.2).
 
 use std::path::{Component, Path, PathBuf};
 
@@ -62,10 +67,14 @@ pub fn inject_adjacent_manifests(
     // manifest referencing a type shipped by another node resolves regardless
     // of node order. URN policy violations (namespace, charset) are reported
     // by each manifest's validation in pass 2; only the `std/` override guard
-    // is load-bearing here.
+    // is load-bearing here. Already-registered URNs are left untouched: types
+    // the dataflow author defined in `types/` win over manifest-shipped ones,
+    // and two nodes sharing one manifest re-ship identical definitions.
     for (_, (_, manifest)) in &matched {
         for (urn, def) in &manifest.types {
-            let _ = registry.add_user_type(urn, def.clone());
+            if registry.resolve(urn).is_none() {
+                let _ = registry.add_user_type(urn, def.clone());
+            }
         }
     }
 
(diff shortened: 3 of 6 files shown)


heyong4725:
**Self-review round 2 (post-open):** a fresh-eyes pass found one fix-first bug, fixed in the follow-up commit: with the default working dir `.` (running `dora build dataflow.yml` from the dataflow's own directory), the lexical containment guard normalized to an empty prefix — absolute node paths walked to the filesystem root looking for manifests, and `../` paths were remapped instead of skipped. The working dir is now absolutized first, with regression tests for both cases.

Also from that pass: author-defined `types/` definitions now take precedence over manifest-shipped type URNs (last-write-wins collisions eliminated), and the compose-time-only scope of P1.4 is documented (injected annotations deliberately don't reach the runtime first-message check or `dora graph` — that's descriptor-propagation work for Phase 2). The reviewer also verified module expansion rebasing, lockfile/session fingerprints, and `dora start` cache matching are unaffected by injection.

phil-opp:

The path-walk containment (working dir absolutized, component-wise `starts_with`, `..`/absolute rejection), control-char sanitization, cross-platform entrypoint matching, and author-override precedence all look correct and well-tested. One issue I think is worth addressing:

**Cross-manifest type contamination: shipped `types:` are registered before they're validated.** In `inject.rs`, pass 1 registers every matched manifest's shipped `types:` URNs into the shared `TypeRegistry` via `add_user_type`, which only rejects the `std/` prefix. The namespace-confinement check (`check_shipped_type_urn`) runs later, in pass 2's `manifest.validate(registry)`, and on failure merely skips injecting *that* manifest's ports — it never un-registers the already-inserted type. So a manifest adjacent to node A can register an out-of-namespace URN (e.g. an `acme` node shipping `beta/foo/v1/X`), and because `check_port_type` treats a URN as valid as soon as `registry.resolve(urn).is_some()`, a *different* node B that references that URN then passes its type check on the strength of a definition its sibling was never entitled to ship. A valid node's validation result is silently changed by an invalid sibling manifest. The "only the `std/` override guard is load-bearing here" comment understates this — once registration happens before validation, the namespace guard is load-bearing *across* manifests. Suggest validating each manifest's shipped-type URNs before registering them (or only registering after the manifest passes validation).

Supporting note: the new `pub fn add_user_type` inserts the URN as a key with no charset/parameterization check (unlike `check_shipped_type_urn`), which is what lets the above through and is a loose contract now that the method is public.

---

heyong4725:
Review finding:

- `libraries/core/src/manifest/inject.rs:73`: pass 1 registers `manifest.types` from every matched manifest before that manifest has passed `manifest.validate()` in pass 2. That means a manifest that is later rejected, for example because it ships `other-ns/...` under the wrong namespace or has any other validation error, still mutates the shared `TypeRegistry`. A second valid manifest can then reference that rejected type and pass validation/type checking because `registry.resolve()` succeeds. This bypasses the namespace/validity gate that is supposed to protect shipped custom types and can let an invalid local manifest influence unrelated nodes in the same dataflow. Please only add types from manifests that have passed validation, or split validation so URN ownership/manifest sanity is checked before global registration, then re-run cross-manifest resolution.

Validation run: `cargo test -p dora-core manifest::inject`; `git diff --check origin/feat/hub-node-manifest...HEAD`.

heyong4725:
Both comments flag the same real bug — fixed in f16e3037c.

**Cross-manifest registry contamination.** `inject.rs` pass 1 registered every matched manifest's shipped `types:` into the shared `TypeRegistry` before the manifest was validated, so an invalid sibling (an `acme` node shipping a `beta/…` URN) polluted the registry, and because `check_port_type` accepts any URN that `registry.resolve()` finds, an unrelated node referencing that URN passed its check on a definition the sibling was never entitled to ship.

Fix: registration is exactly what makes a type visible to *sibling* manifests, so the namespace-ownership gate (`check_shipped_type_urn`) now runs **before** registration. Ownership is intrinsic (needs no sibling types), so this is safe and preserves legitimate cross-manifest references; other per-manifest errors still surface in pass 2 and only suppress that manifest's own injection. The supporting note is addressed too — `add_user_type` now rejects control characters and non-URN bytes as defense-in-depth for its public contract.

Regression test `invalid_sibling_cannot_contaminate_the_registry` pins it; live-verified that a `gamma` node referencing the smuggled `beta/foo/v1/X` now fails with `unknown type` instead of silently resolving. `cargo test -p dora-core manifest::inject` (13 passing).

**CI:** the red `ci/circleci: contract-tests` was a flaky networked smoke job (the same job passes on the #2178/#2180 supersets, which contain this branch's code). The whole stack was just rebased onto this fix, so a fresh run is in progress.

phil-opp:

The namespace-ownership gate added in `9bf4e5b7` closes the entitlement angle of the cross-manifest contamination (a sibling can no longer register a URN under a namespace it doesn't own). One residual of the same class remains, though: pass-1 registration in `inject.rs` is gated only on `check_shipped_type_urn`, not on the shipped type's **body** resolving.

So a manifest that owns its namespace but is otherwise invalid still pollutes the shared `TypeRegistry`. Concretely — manifest A (namespace `acme`) ships `acme/foo/v1/X` with a field `type: NotARealType`: `check_shipped_type_urn` passes (A owns `acme/`), so `add_user_type` registers the unresolvable definition. In pass 2, A fails `validate()` and its own injection is suppressed, but `acme/foo/v1/X` is already registered. A sibling B that references `acme/foo/v1/X` then passes its port check, since `check_port_type` only tests `registry.resolve(urn).is_some()` — validating B clean against a type that can never be built into an Arrow schema. A valid node's result is again silently changed by an invalid sibling.

Suggest registering a manifest's shipped types only after it passes full `validate()`, or validating the type body before the pass-1 registration. (Since this is a stack, the same code is present in #2178's base — fixing it here propagates up.)

---

phil-opp:

The latest change is the rebase onto the inherited `field_type_resolves` DoS fix (`7726c388`) — that part is safe. My earlier finding is still open, though, and unaddressed by these commits:

In `inject.rs` pass-1 registration gates only on `check_shipped_type_urn` (namespace ownership), not on whether the shipped type's **body** resolves. A manifest that *owns* its namespace but ships `acme/foo/v1/X` with an unresolvable field (`type: NotARealType`) still gets `add_user_type`'d before pass-2 `validate()` rejects it. Its own injection is then suppressed, but `acme/foo/v1/X` stays in the shared registry, so a sibling node B referencing it passes its port check (`check_port_type` only tests `registry.resolve(urn).is_some()`) against a type that can never be materialized into an Arrow schema — a valid node's validation result is silently changed by an invalid sibling. Registering shipped types only after the manifest passes full `validate()` (or validating the body before pass-1 registration) would close it.

---

heyong4725:
Fixed in `88fc494a`. Pass-1 now gates shipped-type registration on **body-resolvability** as well as namespace ownership: it builds a scratch registry of all namespace-owned candidate types (field bodies may reference sibling types), then registers into the shared registry only types whose fields all resolve. So a sibling shipping `acme/foo/v1/X` with an unresolvable field no longer contaminates the registry, and a referencing node's port check correctly fails to resolve `X`. Regression test added.

heyong4725:
Both fixed in `21332d76` (cascaded to #2177-#2180):

- **arrow not gated in pass-1 registration**: registration gated on field-resolvability but not the `arrow:` discriminant, so a sibling shipping `acme/foo/v1/X` with `arrow: NotAnArrowType` still reached the shared registry and a referencing node resolved it (yet it could never materialize). Now also gates on `is_known_arrow_type`, matching `NodeManifest::validate`. Regression test added.
- **bare `path:` skipped**: `find_manifest_for` returned early for any bare path, so a console-script / local-exe node next to its `dora-node.yml` (`path: dora-yolo`) missed injection. Bare paths now go through `containment_roots` (rooted at the working dir) and match an adjacent manifest whose entrypoint names them. A bare path that merely shares a dir with an *unrelated* manifest is skipped **silently** (no spurious "doesn't match" warning) — only separator-bearing paths warn on mismatch. Regression test added.

Thanks for the thorough pass. (CI `examples-linux` red remains the pre-existing C++/CMake-example flake — needs a job re-run.)

phil-opp:
Re-reviewed the two new commits (`747943e8` body-resolvability gate, `21332d76` arrow gate + bare-path injection). The direct registry-contamination finding from the last review is now correctly closed, and the bare-path injection logic (inject only when `normalize(dir.join(entrypoint)) == full`, silent skip for non-matching bare paths) is sound and well-tested.

One residual of the same class, low severity: pass-1 builds the `scratch` registry from all namespace-owned candidates gated only on `check_shipped_type_urn`, **not** on body-resolvability. Since `field_type_resolves` resolves a struct reference by key presence (`resolve(urn).is_some()`) without recursing into the referenced type's body, a type `X` whose field references a sibling `Y` that is itself body-invalid is judged `body_valid` (because `Y` is present in `scratch`) and registered into the shared registry — even though `Y` itself is correctly rejected. A sibling referencing `X` would then resolve it.

The blast radius is small (it requires a chained `X → broken Y` reference, and materialization still fails closed at build time), so this is a hardening note rather than a blocker. Building `scratch` only from candidates whose own bodies resolve (or iterating to a fixpoint) would make the gate transitively tight.

---

---

heyong4725:
Fixed in `cd5f4b14a` (cascaded to #2177-#2180). You're right — the residual was real: pass-1 built `scratch` by inserting **every** namespace-owned candidate unconditionally, so `field_type_resolves` saw a body-invalid sibling `Y` by key presence and wrongly admitted a type `X` that references it (and then a third type referencing `X`).

Admission is now a **fixpoint**: a candidate enters `scratch` only once its `arrow:` discriminant is known *and* every field type resolves against what is already admitted. A referent of a body-invalid type is therefore never admitted, and nothing downstream of it can be either — the gate is transitively tight. Registration draws only from the survivors.

Regression test `type_referencing_a_body_invalid_sibling_is_not_registered`: `bad` ships body-invalid `acme/foo/v1/Y` and `acme/foo/v1/X` (field → `Y`); asserts neither reaches the registry and the referencing node's port stays unresolved. Fails under the old unconditional-insert pass (X was admitted because Y sat in scratch), passes now.

phil-opp:
Re-reviewed the latest commit (`cd5f4b14`, the fixpoint admission for cross-manifest type registration). It looks like it closes the transitivity residual from the last round: a candidate now enters `scratch` only when its `arrow:` discriminant is known **and** every one of its fields resolves against already-admitted types, so a type whose field references a body-invalid sibling is no longer admitted via mere key presence, and only survivors reach the shared registry.

I also re-checked the other focus areas on this pass and found no issues: the walk containment (working dir absolutized, then lexical `..`/absolute rejection plus the ancestor `starts_with` break), the `normalize(dir.join(entrypoint)) == full` association, and the author-annotation precedence (only inserting into a `Vacant` entry) all look correct.

One minor, non-blocking observation: `normalize` is purely lexical and doesn't resolve symlinks, so a node `path:` traversing a symlink that resolves outside the working dir would still pass the lexical containment check. Since injection only adds compose-time type annotations and never executes the manifest, the worst case is a spurious type warning rather than an escape — so this is fine as-is, perhaps worth a one-line comment noting the lexical-only guarantee.

No outstanding issues on my end.

---

---

PR #2177, opened 2026-06-12
feat(hub): dora hub subcommand with init scaffolding (P1.2)

heyong4725:
Implements **P1.2** of the Dora Hub spec ([docs/plan-node-hub.md](https://github.com/dora-rs/dora/blob/main/docs/plan-node-hub.md) §9, umbrella #2097): the `dora hub` subcommand skeleton and manifest scaffolding.

**Stacked on #2176** — will retarget as the stack merges.

## What's included

- **`dora hub init [PATH]`** — scaffolds `dora-node.yml`: name/runtime/entrypoint pre-filled from `pyproject.toml` (PEP 621 `[project]`, with a `[tool.poetry]` fallback; prefers the console script matching the package name) or `Cargo.toml` (`target/release/<name>`), namespace from the `origin` remote (forge-agnostic first-path-segment parse). Values are YAML-quoted and PEP 503-normalized; the scaffold is parse-checked against the manifest schema before writing, so the template can never drift. Refuses to overwrite.
- **`dora hub install`** — deliberate error stub explaining the per-dataflow model (spec §9: no global store; `hub:` + `dora build`).
- `search`/`info`/`fetch`/`publish`/`yank` arrive with their phases (P2.4/P2.11/P3.1).

## Verification

- 7 unit tests (detection paths, poetry fallback, script-name preference, name normalization edge cases, scaffold-parses-for-all-paths) + CLI parse tests.
- Live: `dora hub init` on a pyproject fixture → generated manifest passes `dora validate --node-manifest` as-is; re-run refuses to overwrite; `hub install` prints the guidance error.
- fmt + clippy `-D warnings` + full dora-cli tests + qa-fast.

Part of #2097.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1739,6 +1739,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-tungstenite 0.28.0",
+ "toml",
  "tracing",
  "tracing-log",
  "uuid",
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -40,6 +40,7 @@ dora-operator-api-c = { workspace = true }
 dora-download = { workspace = true }
 serde = { workspace = true }
 serde_yaml = { workspace = true }
+toml = "0.9"
 webbrowser = "0.8.3"
 serde_json = { workspace = true }
 termcolor = "1.1.3"
--- a/binaries/cli/src/command/hub/init.rs
+++ b/binaries/cli/src/command/hub/init.rs
@@ -6,9 +6,71 @@
 use std::path::{Path, PathBuf};
 
 use dora_core::manifest::{MANIFEST_FILENAME, NodeManifest};
+use dora_core::types::TypeRegistry;
 use eyre::{Context, bail};
 
 use crate::command::Executable;
+
+/// Placeholders used when a detected value can't be made to satisfy the
+/// manifest validator (both are valid per the validator's rules).
+const FALLBACK_NAME: &str = "my-node";
+const FALLBACK_NAMESPACE: &str = "your-namespace";
+
+/// A valid entrypoint derived from the (already-valid) node name.
+fn fallback_entrypoint(node: &DetectedNode) -> String {
+    match node.runtime {
+        "rust" | "c" | "cpp" => format!("target/release/{}", node.name),
+        _ => node.name.clone(),
+    }
+}
+
+/// Render a scaffold and guarantee it satisfies the manifest validator.
+///
+/// Detected values come from native manifests / the git remote and may violate
+/// the stricter manifest rules (a `std` org, `foo--bar`, an odd console-script
+/// name, …). Any field the validator rejects is replaced with a safe
+/// placeholder, then the result is self-checked with `validate()` (not just
+/// `parse()`), so `dora validate --node-manifest` passes on it immediately.
+fn render_valid_manifest(mut node: DetectedNode, mut namespace: String) -> eyre::Result<String> {
+    let registry = TypeRegistry::new();
+    loop {
+        let content = render_manifest(&node, &namespace);
+        let manifest =
+            NodeManifest::parse(&content).context("internal error: scaffold is not valid YAML")?;
+        let issues = manifest.validate(&registry);
+        let mut changed = false;
+        for issue in &issues {
+            match issue.field.as_str() {
+                "name" if node.name != FALLBACK_NAME => {
+                    node.name = FALLBACK_NAME.into();
+                    changed = true;
+                }
+                "namespace" if namespace != FALLBACK_NAMESPACE => {
+                    namespace = FALLBACK_NAMESPACE.into();
+                    changed = true;
+                }
+                "entrypoint" if node.entrypoint != fallback_entrypoint(&node) => {
+                    node.entrypoint = fallback_entrypoint(&node);
+                    changed = true;
+                }
+                _ => {}
+            }
+        }
+        if !changed {
+            if !issues.is_empty() {
+                bail!(
+                    "internal error: generated manifest is invalid: {}",
+                    issues
+                        .iter()
+                        .map(|i| i.to_string())
+                        .collect::<Vec<_>>()
+                        .join("; ")
+                );
+            }
+            return Ok(content);
+        }
+    }
+}
 
 /// Scaffold a dora-node.yml manifest for a node
 #[derive(Debug, clap::Args)]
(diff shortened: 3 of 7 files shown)


heyong4725:
Self-review round 2 (post-open) verdict was ship, with one consistency item applied in the follow-up commit: entrypoint now goes through the same YAML escaping as description and the git-remote-derived namespace is charset-filtered, closing the template-interpolation class for every value. The install stub's `dora hub fetch` hint is marked as planned (lands with P2.11). Verified clean: description escape ordering (backslash-then-quote), ssh-URL port parsing, file-vs-dir check ordering, clap hyphen-value capture.

phil-opp:
**Automated review by Claude** — this is a fully automated review, not yet vetted by a human, so please treat the findings as suggestions.

I found the following issues:

**1. The `init` self-check only verifies YAML deserialization, not manifest validity.** The scaffold is checked with `NodeManifest::parse()` before writing, but `parse()` is just `serde_yaml::from_str` + the size cap — it does **not** run `validate()` (the charset/semantic rules live in `manifest/validate.rs`). So the "the template can never drift from the schema" comment, and the PR description's "parse-checked against the manifest schema before writing", overstate the guarantee: any detected value that is well-formed YAML but violates the stricter validator rules gets written to disk and the self-check still passes — the user only discovers it later at `dora validate`. Running `validate()` in the self-check would make the claim true.

**2. `detect_namespace` doesn't enforce the namespace rules the validator requires.** It lowercases and filters to `[a-z0-9-]`, but `check_namespace` additionally rejects leading/trailing hyphens, `--`, names longer than 39 chars, and the reserved `std`. So a git `origin` org like `-foo`, `foo--bar`, an over-long org, or literally `std` produces a manifest that the manifest's own validator rejects — written silently because of #1. Applying the same normalization (or validating before writing) closes this.

**3. (minor, same shape)** Console-script entrypoints are written verbatim with no charset check, while `check_entrypoint` only permits `[A-Za-z0-9._\-/\\]`. A legitimate PEP 503 script name containing other characters yields the same write-now-fail-later manifest as #2.

Everything else I checked (YAML quoting/escaping of untrusted toml/remote values, the git remote URL parse, overwrite refusal, the `install` stub) looked correct.

---

heyong4725:
Review finding:

- `binaries/cli/src/command/hub/init.rs:126`: `detect_rust` runs Cargo package names through `normalize_name`, which replaces `_` with `-`, and then uses that normalized value for `entrypoint` at line 135. Cargo binary target names keep underscores by default, so a package named `lidar_driver` builds `target/release/lidar_driver`, but `dora hub init` scaffolds `entrypoint: "target/release/lidar-driver"` (the new test at lines 319-330 currently asserts this wrong path). The generated manifest then points at a binary that does not exist, and the adjacent-manifest injection in the stacked PR will not match the real `path: target/release/lidar_driver`. Please keep a separate Cargo target/binary name for the entrypoint, or read the actual bin target/default-run from Cargo metadata, while only normalizing the manifest package `name` if needed.

Validation run: `cargo test -p dora-cli detects_rust_project`; `cargo test -p dora-cli parse_hub`; `git diff --check origin/feat/hub-contract-injection...HEAD`.

heyong4725:
Fixed in `59222cd5`. `detect_rust` now derives the entrypoint from the **binary target name** (Cargo keeps it verbatim — underscores and all), preferring `default-run` / the first explicit `[[bin]]`, else the package name verbatim — while the manifest `name` field is still hyphen-normalized. So a `lidar_driver` package now scaffolds `entrypoint: "target/release/lidar_driver"`, matching the artifact Cargo actually builds (and the adjacent-manifest injection). Updated the existing test and added coverage for an explicit `[[bin]]` target.

(Re: the automated-review findings on the self-check — those are already addressed: `render_valid_manifest` runs the full `validate()`, not just `parse()`, and falls back to safe placeholders for any name/namespace/entrypoint the validator rejects.)

phil-opp:

Re-reviewed after `5eef893f` and `fafb9e47`. The prior automated findings are addressed: the scaffold now self-checks with `validate()` (not just `parse()`) and replaces any name/namespace/entrypoint the validator rejects with a safe placeholder, so a `std` / `foo--bar` / over-long git org or an odd console-script name no longer produces a write-now-fail-later manifest; and the Rust entrypoint is now derived from the verbatim Cargo binary target (`default-run` / `[[bin]]` / package name) rather than the hyphen-normalized name, so it matches the artifact Cargo actually builds and the adjacent-manifest injection. No new issues found.

---

phil-opp:

Only change since the last review is the rebase onto the inherited `field_type_resolves` DoS fix (`7726c388`); no `dora hub init` code changed. No new findings — the latest commits look safe to merge.

---

heyong4725:
**Pre-landing review** (Claude + Codex adversarial passes, both converged). Fixed in `78872ac5`:

- **Non-atomic + symlink-following scaffold write**: `dora hub init` checked `manifest_path.exists()` then `std::fs::write`. That's TOCTOU and follows symlinks — a planted `dora-node.yml -> /elsewhere` symlink isn't caught by `exists()` and `fs::write` (O_TRUNC) follows it, redirecting the write out of the target dir. Now writes with `OpenOptions::create_new` (O_EXCL): atomic and refuses to follow a symlink (live-verified the victim file is untouched). The early `exists()` check stays as a friendly fast-fail.
- **`yaml_quote` control-char hole**: it escaped only `\`/`"`/`\n`, so a valid `pyproject.toml`/`Cargo.toml` whose description or script name carries a `\u`-escaped control char (`\x1b`, `\r`, …) made serde_yaml reject the scaffold and abort `init` with a misleading "internal error" at `parse()`, before the validate-fixpoint could recover. Now strips control chars; regression test added.

Verified solid (no change): the `git remote get-url origin` namespace detection neutralizes injection (static args, no shell, charset-filtered output), name/namespace interpolation is charset-confined, and the TOML parsing degrades panic-free. Codex's "malformed TOML silently falls back" is intended best-effort scaffolding (the user reviews + runs `dora validate`). Cascaded to #2178-#2180.

heyong4725:
Both fixed in `9a31ce54` (cascaded to #2178-#2180):

- **`normalize_name` not actually PEP 503**: it replaced `_`→`-` but didn't collapse runs, so `My.Package__Name` → `my.package--name` — and `check_package_name` *accepts* that (it allows `.`/`_` and doesn't reject `--`), so a non-canonical name shipped. Now collapses every run of `-`/`_`/`.` to a single `-`, mapping confusable spellings to one canonical index key (`my-package-name`). (Verified the existing `normalizes_names` cases still hold.)
- **console-script matched against the normalized name**: `detect_python` compared `[project.scripts]` keys (raw, `my_pkg`) against the normalized package name (`my-pkg`), so it missed the real match and fell back to the first script. Now compares normalized forms and keeps the raw key as the entrypoint.

Regression tests added for both. Thanks again.

phil-opp:
Re-reviewed the two new commits (`eb21235e` atomic/symlink-safe write + control-char-safe `yaml_quote`; `9a31ce54` PEP 503 separator-run collapsing + raw-key console-script matching). Both look correct: `OpenOptions::create_new` (`O_EXCL`) closes the prior TOCTOU/symlink-follow hole, `normalize_name` now collapses runs of `-`/`_`/`.` to a single `-` as claimed (covered by tests), and console-script matching compares normalized forms while preserving the raw key as the entrypoint. No new findings.

---

---

PR #2178, opened 2026-06-12
feat(hub): dora-hub-client crate — index format, resolution, transport (P2.1+P2.2)

heyong4725:
Implements **P2.1 + P2.2** of the Dora Hub spec ([docs/plan-node-hub.md](https://github.com/dora-rs/dora/blob/main/docs/plan-node-hub.md) §7/§8.1, umbrella #2097): the `dora-hub-client` crate — everything needed to resolve `namespace/name@req` against a catalog, before any descriptor integration.

**Stacked on #2177** — will retarget as the stack merges.

## What's included

- **`reference`** — `[<namespace>/]<name>@<semver-req>` parsing; bare names are deterministic `dora-rs/` shorthand (§7.2); full `semver` requirement syntax.
- **`index`** — the catalog format (`<ns>/<name>/<version>.yml` = manifest snapshot + source pointer, `package.yml` owners): semver resolution picking the highest non-yanked match (cargo prerelease convention, tested), yank skipping (UC10), typo suggestions, and the **binary source form reserved in the schema** but erroring clearly on use (P2.8 deferral per §15 Q3).
- **`config`** — `hub.toml` `[[index]]` entries with **namespace↔index binding**: explicit bindings win, official carries the rest, double-binding is a hard error (no search order to race, §7.3); missing config = official only; `official` alias overridable for enterprise mirrors; local `path =` indexes for fixtures/tests (UC11).
- **`transport`** — blobless **sparse** clone of remote indexes into `~/.cache/dora/index/<alias>` (graceful fallback to plain clone when the server refuses filters), refresh at most once per invocation, **fast-forward check** that warns loudly on history rewrite/rollback (replayed indexes can hide yanks), `--offline` fails loudly on cache miss. Shells out to git deliberately — libgit2 (GitManager) has no partial-clone/sparse support.

## Security posture (self-review hardened)

This crate consumes untrusted input (public index content, user hub.toml), so:
- central `validate_git_url` rejects argument-injection (`--upload-pack=…`) and command transports (`ext::`), applied at config load, at `SourceSpec::git_pin()`, and at the transport boundary, plus `--` separators on git positionals;
- index aliases (cache dir names) and git-index `path` values are charset/confinement validated;
- package keys are charset-checked before filesystem joins; catalog listing skips out-of-charset directory names (terminal-output safety); commit hashes validated alphanumeric;
- index files are read with a bounded `take()` read.

## Verification

- 27 unit tests, including end-to-end git transport tests against local repos (clone → resolve → publish upstream → refresh-once semantics → new invocation sees the new version), offline cache-hit/miss, non-fast-forward warning, hostile config/URL rejection, prerelease semantics.
- fmt + clippy `-D warnings` + full workspace suite + qa-fast.

Part of #2097. Next: P2.5–P2.7 wire `hub:` descriptor resolution through this crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1928,6 +1928,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "dora-hub-client"
+version = "1.0.0-rc1"
+dependencies = [
+ "dirs 6.0.0",
+ "dora-core",
+ "eyre",
+ "semver",
+ "serde",
+ "serde_yaml",
+ "tempfile",
+ "toml",
+]
+
+[[package]]
 name = "dora-log-utils"
 version = "1.0.0-rc1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ members = [
     "libraries/log-utils",
     "libraries/coordinator-store",
     "libraries/core",
+    "libraries/hub-client",
     "libraries/message",
     "libraries/recording",
     "binaries/replay-node",
@@ -102,6 +103,7 @@ dora-operator-api-python = { version = "1.0.0-rc1", path = "apis/python/operator
 dora-operator-api-c = { version = "1.0.0-rc1", path = "apis/c/operator" }
 dora-node-api-c = { version = "1.0.0-rc1", path = "apis/c/node" }
 dora-core = { version = "1.0.0-rc1", path = "libraries/core" }
+dora-hub-client = { version = "1.0.0-rc1", path = "libraries/hub-client" }
 dora-recording = { version = "1.0.0-rc1", path = "libraries/recording" }
 dora-arrow-convert = { version = "1.0.0-rc1", path = "libraries/arrow-convert" }
 dora-tracing = { version = "1.0.0-rc1", path = "libraries/extensions/telemetry/tracing" }
--- a/libraries/core/src/types.rs
+++ b/libraries/core/src/types.rs
@@ -692,7 +692,7 @@ impl Default for TypeRegistry {
 
 /// Simple edit distance (Levenshtein) for typo suggestions.
 /// Returns `usize::MAX` for inputs longer than 256 characters to prevent DoS.
-fn edit_distance(a: &str, b: &str) -> usize {
+pub fn edit_distance(a: &str, b: &str) -> usize {
     if a.len() > 256 || b.len() > 256 {
         return usize::MAX;
     }
--- a/libraries/hub-client/Cargo.toml
+++ b/libraries/hub-client/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "dora-hub-client"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+documentation.workspace = true
+description = "Dora Hub index client: package references, index resolution, and transport"
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+dora-core = { workspace = true }
+eyre = { workspace = true }
+semver = { workspace = true }
+serde = { workspace = true }
+serde_yaml = { workspace = true }
+toml = "0.9"
+dirs = "6"
+
+[dev-dependencies]
+tempfile = { workspace = true }
--- a/libraries/hub-client/src/config.rs
+++ b/libraries/hub-client/src/config.rs
@@ -100,7 +100,9 @@ impl ResolvedConfig {
         let mut seen_aliases = BTreeMap::new();
         for (idx, index) in indexes.iter().enumerate() {
             validate_index_entry(index)?;
-            if let Some(previous) = seen_aliases.insert(index.alias.clone(), idx) {
+            // case-insensitive: aliases become cache directory names, and
+            // `acme`/`ACME` would silently share one clone on macOS/Windows
+            if let Some(previous) = seen_aliases.insert(index.alias.to_ascii_lowercase(), idx) {
                 eyre::bail!(
                     "duplicate index alias `{}` (entries {} and {idx})",
                     index.alias,
@@ -166,7 +168,7 @@ impl ResolvedConfig {
 /// Validate one `[[index]]` entry. The alias becomes a cache directory name
 /// and the git URL/path become `git` subprocess arguments — all of it is
 /// user- or repo-supplied, so each field is constrained before use.
-fn validate_index_entry(index: &IndexConfig) -> eyre::Result<()> {
+pub(crate) fn validate_index_entry(index: &IndexConfig) -> eyre::Result<()> {
     let alias_valid = index.alias.len() <= 64
         && index
             .alias
--- a/libraries/hub-client/src/config.rs
+++ b/libraries/hub-client/src/config.rs
@@ -1,0 +1,356 @@
+//! Multi-index configuration and namespace binding (spec §7.3).
+//!
+//! `~/.config/dora/hub.toml` declares `[[index]]` entries. **A namespace
+//! resolves against exactly one index**: explicit `namespaces` bindings win,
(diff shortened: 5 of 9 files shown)


heyong4725:
**Self-review round 2 (post-open):** a fresh-eyes pass reproduced two transport bugs, fixed in the follow-up commits:

1. `git fetch` never updates an existing `refs/remotes/origin/HEAD`, so an index repo renaming its default branch (master→main style) would silently freeze every cache on the old tip — never seeing new yanks. Refresh now runs `git remote set-head origin --auto` after fetching.
2. An interrupted checkout (the heavy network step of a blobless clone) left a valid `.git` with no worktree, wedging the cache permanently behind a misleading "no catalog directory" error. The cache now self-heals by re-cloning (with an `--offline` repair hint); corrupt-`.git` refresh failures self-heal too, while genuine network errors still propagate so a flaky connection can't wipe a cache offline runs depend on.

Also hardened from that pass: `rev` pins must be hex commit hashes (a branch name would make the pin mutable, defeating the audit trail); cleartext `http://`/`git://` dropped from the URL allowlist (the index is a supply-chain trust root — fast-forward protection is meaningless against a MITM); the transport boundary validates the full index entry; case-insensitive duplicate-alias detection (cache-dir collisions on macOS/Windows); `GIT_TERMINAL_PROMPT=0`; IO errors no longer masquerade as package-not-found. The reviewer verified the probes for semver overflow, `config_dir` fallback, and hostile local-path sources came back clean. Carried forward to the P2.7 PR: `SourceSpec.subdir` confinement validation at the consumer.

phil-opp:

This crate's untrusted-input hardening looks solid — per-segment path validation blocks `..`/`/`/leading-`.`, git positionals get `--` separators, `read_capped` uses a `take(MAX+1)` bounded read, the fast-forward/rollback detection (`merge-base --is-ancestor`) is in the right direction, and double-binding / duplicate-alias are hard errors. One thing worth a look:

**`remote_tip` can silently pick the wrong branch.** In `transport.rs`, `remote_tip` tries `refs/remotes/origin/HEAD`, then falls back to `origin/main`, then `origin/master`. The `git remote set-head origin --auto` that populates `origin/HEAD` is best-effort (`let _ = ...`). For an index whose default branch is neither `main` nor `master` (e.g. `develop`), if `set-head` fails (older git, transient ls-remote failure), `remote_tip` either bails or — if a stale `origin/main` happens to exist — compares/resets against the wrong branch, which can produce spurious "history was rolled back" warnings or resolve against a stale tip. Consider failing loudly when `origin/HEAD` can't be determined, rather than the silent `main`/`master` fallback chain.

Two cosmetic notes (non-blocking): the unsatisfiable-requirement error lists yanked versions in its "available" set, and `validate_git_url`'s `git@` branch validates nothing past the prefix (argument injection is still neutralized downstream by the `--` separator, so not exploitable as far as I can tell).

---

heyong4725:
Review finding:

- `libraries/hub-client/src/index.rs:89`: `SourceSpec::git_pin()` accepts any 7- to 64-character hex string as a valid `rev`, so index entries can publish abbreviated commit IDs. For a hub index, this is not a stable source pin: an abbreviated object name can become ambiguous as the source repo grows, and it does not provide the immutable audit pointer/yank trail promised by the index format (`docs/plan-node-hub.md:579` calls this a commit hash/pinned commit). Please require full object IDs, e.g. 40 hex chars for SHA-1 and 64 for SHA-256, and update the tests/fixtures to use full hashes.

Validation run: `cargo test -p dora-hub-client`; `git diff --check origin/feat/hub-cli-init...HEAD`.

heyong4725:
Addressed in `0e905b86` (rebased onto the current #2177 tip):

- **`remote_tip` resolution** — now resolves strictly from `refs/remotes/origin/HEAD` instead of guessing `origin/main` → `origin/master`. The fallback chain could latch onto a stale or wrong default branch on repos whose HEAD points elsewhere.
- **Commit pins** — `git_pin` now requires a full 40-char (SHA-1) or 64-char (SHA-256) hex object id. Abbreviated hashes grow ambiguous as the source repo gains commits, and a branch/tag name would make the "pin" mutable, defeating yanks and the audit trail. New test `git_pin_requires_a_full_object_id`.
- **Yank reporting** — resolution errors now list only non-yanked available versions and say "every version matching X has been yanked" when all matches are yanked, instead of suggesting versions the resolver will never pick.

The downstream branches (#2179, #2180) were rebased onto this so they inherit the fix; the full stack is properly chained again.

phil-opp:

Re-reviewed after the latest commits (`24f7deb5` et al.). The two correctness items and the cosmetics from the prior automated review are addressed: `remote_tip` now resolves strictly through `refs/remotes/origin/HEAD` (no silent `main`/`master` fallback) and fails loudly when it can't, `git_pin` requires a full 40/64-char hex object id, and resolution errors list only non-yanked versions / say so when every match is yanked. The transport hardening (clone/refresh/fast-forward warning, corrupt-cache self-heal, `--offline` cache-miss, `--` separators, alias/path/key confinement, capped reads) holds up. No new issues found.

---

phil-opp:

Only change since the last review is the rebase onto the inherited `field_type_resolves` DoS fix (`7726c388`); no hub-client transport/index code changed. No new findings — the latest commits look safe to merge.

---

heyong4725:
**Pre-landing review** (Claude + Codex adversarial passes). Both converged; fixed in `83ab69ad` (cascaded to #2179/#2180):

- **Symlink catalog escape** (Codex Med, both flagged): the catalog read followed in-repo symlinks, so a malicious index repo could ship a symlinked namespace/package/version entry (or the `node-index` subpath) pointing outside the clone. `IndexCatalog` now confines every read under the canonicalized root; the transport confines the catalog subpath under the clone. Unix symlink-escape regression test added.
- **Alias-only cache reuse** (Codex High): the cache is keyed by alias, so re-pointing an alias to a different URL/subpath in `hub.toml` kept fetching the old clone's origin. Now records the source in a `.dora-index-source` marker and re-clones on mismatch.
- **Official-namespace shadowing** (Claude F1): a non-official `[[index]]` binding `namespaces = ["dora-rs"]` silently shadowed the official source. Now rejected (test added) — makes the "structurally impossible" doc claim hold.
- **Local sources in untrusted entries** (Codex High / architect F5): `git_pin` accepted `file://`/absolute `source.git`, which would clone victim-local content. New `validate_git_url_untrusted` rejects local sources for index entries unless `DORA_HUB_ALLOW_LOCAL_SOURCES` opts in (the hub-smoke fixture sets it); `hub.toml` index *locations* stay permissive (trusted user config).

Confirmed solid (no change): `--` git separators + scheme allowlist close argument injection; `is_valid_key_part` blocks catalog key traversal; `git_pin`'s full-hash requirement. The non-FF-rollback warn-and-continue is a deliberate documented tradeoff (a legitimate index force-push during bootstrap shouldn't hard-fail). hub-smoke verified green end-to-end (47s) with the new validation.

heyong4725:
Good catch — fixed in `00f47d65`. `package_meta()` read `package.yml` after only an `is_file()` check, bypassing the `self.confine()` symlink-containment that `versions()` and `entry()` use. Added the same gate, so a symlinked `package.yml` (or package dir) pointing outside the catalog is refused. Extended the symlink-escape regression test to assert `package_meta` also rejects it. Cascaded to #2179/#2180.

phil-opp:
Re-reviewed after the post-review hardening commits (`3b4e7fdd`, `00f47d65`). The symlink-escape confinement (`IndexCatalog::confine` canonicalize + prefix-check, now also covering `package_meta`), the `.dora-index-source` cache-identity marker, the official-namespace shadowing rejection, and `validate_git_url_untrusted` for index-entry local sources all hold up. No new findings — the latest commits look safe.

---

---

Issue #1444, opened 2026-03-18
session: invalidate `DataflowSession.build_id` on build-process delta


Bhanudahiyaa:
## Problem
  `DataflowSession` currently reuses `build_id` / `local_build` across commands, but session reuse is not tied to whether the current dataflow descriptor still matches the one that produced that
  build state.

  When users modify topology or node sources and then run `dora start`/`dora run`, stale build metadata may still be consumed.

  ## Why this matters
  - Can cause subtle runtime failures when old build artifacts are reused after descriptor changes.
  - Can cause cross-machine mismatch in distributed scenarios (`dora start` path).
  - Hard to debug because the stale state is valid structurally but semantically outdated.

  ## Proposal
  Persist a descriptor fingerprint in `DataflowSession` and synchronize on session read:
  - Compute fingerprint from resolved descriptor graph.
  - If fingerprint changed:
    - clear `build_id`
    - clear `local_build`
    - clear cached `git_sources`
    - store new fingerprint

  Then use this synchronized session read path in:
  - `dora build`
  - `dora start`
  - `dora run`
  - hidden daemon local dataflow path (`dora daemon --run-dataflow`)

Bhanudahiyaa:
Thanks, this policy makes sense to me. I’ll pivot to this narrower scope and update PR #1445 afterwards after confirmation.

  I agree with:
  - invalidate only on build-process deltas
  - ignore `path`-only deltas
  - ignore `deploy.machine`-only deltas
  - keep it CLI-side by sending `build_id: None` on `start` when changed

One thing I think is missing: env changes (global/node env), because build commands inherit node env during build. So changing env can change build outputs even if `build` text is unchanged.

  Also for “build key changed”, I’ll include both:
  - custom node `build`
  - runtime operator `build` fields

  Implementation plan:
  1. Compute a build-process fingerprint from build-relevant inputs only:
     - build command fields
     - source identity (repo/rev config and local↔git switch)
     - effective build working directory
     - build env
  2. On `dora start`, compare current fingerprint to the one from last build.
  3. If changed, send `build_id: None` in `StartRequest`.
  4. Do not change coordinator/daemon behavior, and do not clear session state on start.

One clarification: with current session metadata, reliably detecting build-process deltas requires storing one build fingerprint at build time.
If you prefer zero session schema changes, I can restrict detection to source/working-dir deltas only.
Would storing this minimal build fingerprint in session be acceptable?

heyong4725:
[...]

PR #2179, opened 2026-06-12
feat(hub): hub: descriptor field with end-to-end resolution (P2.5-P2.7)

heyong4725:
Implements **P2.5 + P2.6 + P2.7** of the Dora Hub spec ([docs/plan-node-hub.md](https://github.com/dora-rs/dora/blob/main/docs/plan-node-hub.md) §10/§11, umbrella #2097): the `hub:` field works end-to-end — `dora build`, `dora run`, `dora validate`, `--locked`, `--offline`.

**Stacked on #2178** — will retarget as the stack merges.

## The mechanism (spec §10.1)

The CLI desugars `hub:` into a concrete git node **before dispatch**: the index entry supplies the pinned commit + `subdir`, the manifest supplies the entrypoint, `build:` command, and typed contracts (injected through the shared `apply_manifest_contracts`; contract violations are hard errors for hub nodes — the manifest is the contract). Downstream, coordinator/daemons/lockfile see an ordinary git node whose `GitSource` carries two new optional wire bits: `subdir` and the `hub` provenance marker.

- **`subdir` (P2.7a)**: build, env prep, and spawn root at `<clone>/<subdir>` — independently useful for plain `git:` monorepo nodes. Validated (relative, no `..`, charset) at desugar *and* at build consumption.
- **Confined resolution (P2.7b)**: daemons recognize hub nodes by the provenance marker (recorded as `BuildInfo.confined_nodes`) and spawn through `resolve_path_confined`: working dir or managed env only, **no ambient-`$PATH` fallback**, and the resolved path is canonicalized + prefix-checked, so a package shipping `bin/run -> /bin/sh` fails loudly instead of escaping (closes the symlink-escape residual flagged in the P1 reviews).
- **Lockfile (P2.6)**: hub provenance rides the existing `git_sources` map (the pinned commit *is* the lock). Files carrying hub entries are written as **v3** so older dora fails loudly instead of silently ignoring `subdir` and mis-rooting builds; plain-git lockfiles stay byte-compatible v2, and the fingerprint schema tag is deliberately decoupled so existing lockfiles aren't invalidated. `--locked` uses pins verbatim (no resolution), verifies name/range still match, and warns on yanked pins (UC10).
- **`dora start`/`dora run`**: both re-read YAML from disk, which still contains unresolved `hub:` — the desugared descriptor is stored in the dataflow session at build time; `start` verifies the on-disk references still match the built provenance before substituting, `run` passes it to the daemon as a descriptor override. An unresolved `hub:` reaching `kind()` errors with "run `dora build` first".
- Plus: manifest `build:` field (spec §10.1 requires it; §5 omitted it), `--offline` on build/validate, `DORA_HUB_CONFIG` env override for hermetic tests, and `file://` git URLs in GitManager (local mirrors/air-gapped, host falls back to `localhost`).

## Live-tested end-to-end

Hermetic fixture (local catalog + `file://` monorepo source repo, `namespaces = ["test"]` binding):
- `dora build --write-lockfile` → resolution note, clone at the pinned commit, `cargo build` running **in `<clone>/node-hub/hello`**, v3 lockfile with `subdir` + `hub: {name, version}`.
- `dora run` → `hello from the hub node` through confined spawn.
- `dora build --locked` → pin used verbatim, no resolution.
- `dora validate` → resolution + contracts + wiring, all OK.
- Negative: a version whose entrypoint is `sh` **fails** with "hub nodes resolve only within their own package (no $PATH fallback)" instead of running the host shell.

## Self-review round (pre-open)

Fixed from review: session invalidation now clears `resolved_dataflow` (stale-error trap after non-hub edits); clone-reuse compares repo+commit identity only (provenance fields would have broken `still_needed_for_this_build` and forced re-clones); lockfile version assertions; script-only Python interpreter caveat documented (hub Python nodes are expected to ship `build:` for a managed env, spec §15 Q4).

## Verification

- New unit tests: subdir confinement, confined resolution (no-PATH + symlink-escape rejection), unresolved-hub kind() error, lockfile v2/v3 selection + hub roundtrip, `--offline` parse tests.
- fmt + clippy `-D warnings` (workspace) + full test suite + `cargo check --examples` + qa-fast.

Part of #2097. Next: P2.4/P2.11 (`dora hub search/info/fetch`, `--hub-override`, fixture-index smoke tests, `examples/hub-dataflow`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1708,6 +1708,7 @@ dependencies = [
  "dora-core",
  "dora-daemon",
  "dora-download",
+ "dora-hub-client",
  "dora-message",
  "dora-node-api-c",
  "dora-operator-api-c",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1733,6 +1733,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2",
  "sysinfo 0.36.1",
  "tabwriter",
  "tempfile",
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -44,6 +44,7 @@ serde_yaml = { workspace = true }
 toml = "0.9"
 webbrowser = "0.8.3"
 serde_json = { workspace = true }
+sha2 = "0.10"
 termcolor = "1.1.3"
 uuid = { workspace = true }
 inquire = { workspace = true }
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -34,6 +34,7 @@ clap_complete = "4.5.61"
 eyre = { workspace = true }
 libc = { version = "0.2", optional = true }
 dora-core = { workspace = true, features = ["zenoh"] }
+dora-hub-client = { workspace = true }
 dora-message = { workspace = true }
 dora-node-api-c = { workspace = true }
 dora-operator-api-c = { workspace = true }
--- a/binaries/cli/src/command/build/git.rs
+++ b/binaries/cli/src/command/build/git.rs
@@ -39,6 +39,8 @@ pub fn fetch_commit_hash(repo_url: String, rev: Option<GitRepoRev>) -> eyre::Res
         Some(commit_hash) => Ok(GitSource {
             repo: repo_url,
             commit_hash,
+            subdir: None,
+            hub: None,
         }),
         None => eyre::bail!("no matching commit for `{rev:?}`"),
     }
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -96,9 +96,26 @@ pub fn resolve_hub_nodes(
         let catalog = IndexCatalog::open(&catalog_dir)?;
 
         let pin = pins.and_then(|p| p.get(&node.id));
+        // under `--locked` (pins present), a node with no lockfile entry must
+        // fail fast rather than silently resolving from the network
+        if pins.is_some() && pin.is_none() {
+            eyre::bail!(
+                "node `{}`: `hub:` reference is not in the lockfile — \
+                 regenerate with `dora build --write-lockfile`",
+                node.id
+            );
+        }
         let (version, entry, source) = match pin {
             Some(pin) => {
                 // --locked: the pinned commit is used verbatim, no resolution
+                let valid_hash = (7..=64).contains(&pin.commit_hash.len())
+                    && pin.commit_hash.chars().all(|c| c.is_ascii_hexdigit());
+                if !valid_hash {
+                    eyre::bail!(
+                        "node `{}`: lockfile commit hash is not a valid hex hash",
+                        node.id
+                    );
+                }
                 let provenance = pin.hub.as_ref().with_context(|| {
                     format!(
                         "node `{}`: lockfile entry has no hub provenance — \
@@ -148,9 +165,28 @@ pub fn resolve_hub_nodes(
                             .as_deref()
                             .map(|r| format!(
                                 " ({})",
-                                r.chars().filter(|c| !c.is_control()).collect::<String>()
+                                r.chars()
+                                    .filter(|c| !c.is_control())
+                                    .take(200)
+                                    .collect::<String>()
                             ))
                             .unwrap_or_default()
+                    ));
+                }
+                // the lockfile is authoritative for the pin, but the index
+                // entry of the pinned version should still agree — a mismatch
+                // means the append-only index was tampered with or rewritten
+                if let Ok((git, rev)) = entry.source.git_pin()
+                    && (git != pin.repo
+                        || rev != pin.commit_hash
+                        || entry.source.subdir != pin.subdir)
+                {
+                    resolution.warnings.push(format!(
+                        "node `{}`: the index entry for `{}@{version}` no longer matches \
+                         the lockfile pin — the index may have been rewritten; \
+                         the lockfile pin is used",
+                        node.id,
+                        reference.key()
                     ));
                 }
                 (version, entry, pin.clone())
@@ -214,10 +250,10 @@ pub fn resolve_hub_nodes(
         node.git = Some(source.repo.clone());
         node.rev = Some(source.commit_hash.clone());
         node.hub = None;
+        let short_hash: String = source.commit_hash.chars().take(12).collect();
         resolution.notes.push(format!(
-            "node `{}`: resolved hub package {label} -> {}",
-            node.id,
-            &source.commit_hash[..source.commit_hash.len().min(12)]
+            "node `{}`: resolved hub package {label} -> {short_hash}",
+            node.id
         ));
         resolution.sources.insert(node.id.clone(), source);
     }
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -1,0 +1,227 @@
+//! `hub:` reference resolution — the desugar step (spec §10.1, P2.5/P2.7).
+//!
+//! The CLI rewrites every `hub:` node into a concrete git-sourced node
+//! *before dispatch*: the index entry supplies the pinned commit (and
+//! optional `subdir`), the package manifest supplies the entrypoint, build
+//! command, and typed contracts. Downstream — coordinator, daemons, the
+//! lockfile — sees an ordinary git node carrying the optional `subdir` and
+//! hub provenance marker on its [`GitSource`].
+
+use std::collections::BTreeMap;
+
+use dora_core::{
+    build::validate_subdir,
+    manifest::inject::{InjectionResult, apply_manifest_contracts},
+    types::TypeRegistry,
+};
+use dora_hub_client::{
+    config::ResolvedConfig, index::IndexCatalog, reference::PackageRef, transport::IndexFetcher,
+};
+use dora_message::{
+    common::{GitSource, HubProvenance},
+    descriptor::Descriptor,
+    id::NodeId,
+};
+use eyre::{Context, ContextCompat};
+
+/// Outcome of resolving the `hub:` nodes of a dataflow.
+#[derive(Debug, Default)]
+pub struct HubResolution {
+    /// Progress/info lines for the user.
+    pub notes: Vec<String>,
+    /// Non-fatal problems (yanked pins, index rollback warnings).
+    pub warnings: Vec<String>,
+    /// Per-node resolved git source (with subdir + hub provenance) — merged
+    /// into the build's `git_sources` map and the lockfile.
+    pub sources: BTreeMap<NodeId, GitSource>,
+}
+
+impl HubResolution {
+    pub fn is_empty(&self) -> bool {
+        self.sources.is_empty()
+    }
+}
+
+/// Resolve and desugar every `hub:` node in `dataflow`.
+///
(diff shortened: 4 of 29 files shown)


heyong4725:
**Final-pass self-review (post-open):** a fresh read found two functional breaks plus hardening, fixed in the follow-up commit:

- **F1 (fix-first):** `dora daemon --run-dataflow` fingerprinted the raw on-disk YAML *before* substituting the desugared descriptor — so the new `kind()` guard rejected the unresolved `hub:` reference and the command errored even after a successful `dora build`. It now substitutes first (mirroring `dora start`).
- **F2 (fix-first):** the `dora start` staleness check only compared the per-hub-node reference key, so a non-hub edit to a hub dataflow (or a structural change) was silently discarded — defeating the #1444 guarantee. Both `start` and `daemon --run-dataflow` now compare a digest of the expanded descriptor captured at build time, catching *any* on-disk edit.
- **F3/F5:** `--locked` cross-checks the index entry against the lockfile pin and warns on divergence (index-rewrite signal), and a hub node missing from the lockfile fails fast instead of resolving from the network.
- **F4:** the lockfile commit hash is validated as hex and the short-hash note uses char-safe truncation (no multi-byte slice panic); yank reasons are length-capped.
- **F6:** a manifest entrypoint may not be the reserved `dynamic`/`shell` spawn sentinels, which would otherwise bypass confined resolution.

New unit test pins F2 (source digest changes on any edit). Re-verified end-to-end (build/run/`--locked`/`--offline`/confined rejection) on the hermetic local-index fixture.

heyong4725:
CI note: the `ci/circleci: e2e` red is flaky test-parallelism, not a regression. The 5 failing `real_dataflow` cases contend for coordinator port 6013 when run concurrently; each passes in isolation (`--test-threads=1`), and this PR's runtime changes (daemon confined-spawn flag, start/run descriptor substitution) are all gated behind `nodes.any(|n| n.hub.is_some())` — inert for the non-hub dataflows e2e exercises. A re-run should go green.

heyong4725:
Review finding:

- `libraries/hub-client/src/lib.rs:35` / `binaries/cli/src/command/build/hub.rs:206`: hub index sources accept `git@...` scp-style URLs (and absolute paths) via `SourceSpec::git_pin()`, and the desugar step stores that string directly into `GitSource.repo`. The actual build path then calls `GitManager::choose_clone_dir`, which does `url::Url::parse(&repo)` at `libraries/core/src/build/git.rs:38`; that rejects scp-style `git@github.com:org/repo.git` (and plain absolute paths). So a catalog entry that passes hub-client validation and resolution fails later during build with "failed to parse git repository URL". Please either narrow `validate_git_url`/`git_pin` to the URL forms `GitManager` can consume, or teach `GitManager` to normalize/support the same accepted forms before deriving the clone directory.

Validation run: `cargo test -p dora-cli hub`; `cargo test -p dora-core resolve_path_confined`; `git diff --check origin/feat/hub-client...HEAD`.

heyong4725:
Fixed in `df1991d0`. The desugar now normalizes the index `source.git` into a form `GitManager::choose_clone_dir` (`url::Url::parse`) accepts: scp-style `git@host:path` → `ssh://git@host/path`, absolute path → `file://…`. Applied both where the resolved `GitSource` is built and in the lockfile-vs-index agreement check, so the stored repo and the comparison stay consistent. `validate_git_url` stays permissive on purpose — ssh `hub.toml` indexes are fetched by `transport` (raw git), which handles scp-style fine, so narrowing it would have broken that path. Added a unit test covering scp/path/https/ssh/file forms.

heyong4725:
CI fix (`48c0fb78`): the `e2e` job was failing to **compile**, not flaking — `Daemon::run_dataflow` gained the `descriptor_override` parameter in this PR, but the `tests/fault-tolerance-e2e.rs` callers still passed 10 args (E0061 ×11). That target isn't built by `cargo check --all` or the standard `cargo test` run, so it slipped past local gates. Updated all 11 call sites to pass `None` and dropped a duplicated `#[allow(clippy::too_many_arguments)]`. Both e2e targets now compile; the change is test-only + an attribute dedupe.

phil-opp:

One hardening gap worth a look on this security-sensitive resolution path:

**Subdir confinement doesn't account for in-repo symlinks** — `libraries/core/src/build/mod.rs`, the `Some(subdir)` arm (`let dir = clone_dir.join(&subdir); if !dir.is_dir()`). `validate_subdir` rejects `..` / absolute / leading-`-` / control chars, but a hub package can still ship a *symlink inside the clone* — e.g. `node-hub/hello -> /etc`. `subdir = "node-hub/hello"` passes validation, `is_dir()` follows the link, and build / env-prep / spawn all get rooted at the symlink target; the entrypoint `confine()` check then anchors to that escaped dir, so it can't catch the escape. This doesn't grant net-new code execution (the manifest `build:` step is already trusted to run arbitrary code), but it defeats the "confined to the clone" property the subdir validation is meant to provide. Suggest canonicalizing `clone_dir.join(subdir)` and requiring it to stay under the canonicalized `clone_dir`, mirroring the prefix check `confine()` already does for entrypoints.

Minor: in `resolve_path_confined`, the working-dir branch tests `candidate.exists()` while the managed-env branch tests `candidate.is_file()`; using `is_file()` in both avoids accepting a directory that happens to match the entrypoint name.

(For the record, `dora run` consuming `resolved_dataflow` without a `source_fingerprint` staleness check is fine — it always builds immediately beforehand, unlike `start` / `daemon --run-dataflow`. The F1–F6 / scp-URL self-review items are all present in the diff.)

---

phil-opp:

Rebased onto the inherited `field_type_resolves` DoS fix (`7726c388`) — that part is safe. Still open from the last review and unaddressed by these commits:

The `Some(subdir)` arm in `build/mod.rs` validates `subdir` only **lexically** (`validate_subdir` rejects `..`/absolute/etc.) and then does `clone_dir.join(subdir)` guarded by `is_dir()`, which **follows an in-repo symlink**. A hub package can ship `node-hub/x -> /etc` inside its own clone: `subdir = "node-hub/x"` passes validation, `is_dir()` follows the link, and build / env-prep / spawn all root at the symlink target — outside the clone. The entrypoint `confine()` check then anchors to the escaped dir, so it can't catch it. Canonicalizing `clone_dir.join(subdir)` and requiring it to stay under the canonicalized `clone_dir` (mirroring `confine()`) would restore the confinement `validate_subdir` is meant to provide.

---

heyong4725:
Good catch — fixed in `b7fa214e`. Extracted a `confine_subdir` helper that canonicalizes `<clone>/<subdir>` and requires the real path to stay under the canonical clone, mirroring the spawn-time `confine` check. So a package shipping an in-repo symlink (`node-hub/x -> /etc`) that passes lexical `validate_subdir` and `is_dir()` is now rejected before build/env-prep/spawn root anywhere outside the clone. Added a unix symlink-escape regression test. Cascaded to #2180.

heyong4725:
**Pre-landing review** (Claude adversarial pass; Codex run errored on a flag-quoting bug, re-verified the key finding by hand). The confinement core is **solid** — canonicalize+containment defeats symlink/`..`/absolute escapes, no ambient `$PATH`, `dynamic`/`shell` sentinels rejected, and the "git source ⇒ build_info required" gate makes a dropped `confined` flag structurally impossible. Two residual findings, fixed in `eae3a0ec` (cascaded to #2180):

- **`--locked` didn't pin the entrypoint/build command**: the lockfile stored only `HubProvenance {name, version}`, and the desugar pulled `entrypoint`/`build` from the *mutable* index entry — the lockfile-vs-index check compared only the git pin and merely warned. So a rewritten index entry could inject an arbitrary `build:` command at an already-pinned version with no error, contradicting the documented guarantee. Now records `entrypoint`+`build` in `HubProvenance` and **hard-errors** under `--locked` if they changed; the `entrypoint` field doubles as the new-format sentinel so old lockfiles stay back-compat. Added a hub-smoke tamper-detection test (rewrite `build:` → `--locked` build fails).
- **daemon `--run-dataflow` staleness gate**: `resolved_dataflow` was passed unconditionally while the source-fingerprint gate only ran for hub dataflows, so a non-hub on-disk file could run against a stale resolved descriptor. Now binds a `descriptor_override` that is `Some` only when the hub gate is taken (mirrors `dora start`).

Noted (not changed): `DORA_HUB_CONFIG` env override is intentionally build-time-only (trusted-position config).

heyong4725:
Both fixed in `a2680bfe` (cascaded to #2180):

- **`--locked` only pinned entrypoint+build, not the contract**: you're right — `apply_manifest_contracts` reads the mutable index manifest, so a rewritten entry could change `inputs/outputs/types` (and thus validation/annotations) for a pinned version. Replaced the `entrypoint`/`build` fields in `HubProvenance` with a single SHA-256 **`manifest_digest`** over the whole manifest; `--locked` hard-errors if it no longer matches. The hub-smoke tamper test now exercises that path (rewrite `build:` → digest mismatch → hard error).
- **index fetched before the missing-pin bail**: moved the pin lookup + "not in lockfile" bail ahead of `catalog_dir()`, so a `--locked` node absent from the lockfile fails fast without any network / offline-cache-miss.

Thanks — the digest is strictly more complete than the entrypoint/build pinning it replaces.

phil-opp:
Re-reviewed after `a8fdface`, `33592dcb`, `a2680bfe`. The subdir in-repo-symlink escape flagged last time is fixed (`confine_subdir` canonicalizes `<clone>/<subdir>` and requires it under the canonical clone, with a regression test). The `--locked` path now pins the whole manifest via a SHA-256 `manifest_digest` and hard-errors on mismatch, closing the "rewritten index entry changes entrypoint/build/contract at a pinned version" gap, and the missing-pin bail now precedes any index fetch. Staleness gating and the confined-spawn resolution hold up. No new findings.

---

---

phil-opp:
Took a focused pass over the security-sensitive paths (confined resolution, subdir containment, `--locked` pinning, lockfile v2/v3). The hardening holds up well — `confine_subdir`/`confine` canonicalize-then-contain defeats the symlink/`..`/absolute escapes, the spawned path is the canonicalized one (so there's no check-vs-spawn re-resolution gap), the `manifest_digest` hard-error closes the rewritten-index-entry hole, and v3 lockfiles fail loudly on older dora via the version-range check. Nothing blocking.

One thing I'd flag for a comment or a defensive guard (not necessarily a fix):

In `binaries/daemon/src/lib.rs`, the two `spawn_node` call sites disagree on confinement. The main build/spawn path derives it from the build info:

```rust
// ~line 2936
.spawn_node(
    node,
    node_working_dir,
    configured_python_env_dir,
    confined_nodes.contains(&node_id),   // <- honors §11 confinement
    ...
)
```

but the `DaemonCoordinatorEvent::AddNode` path (runtime add-node-to-running-dataflow) hardcodes it and uses `base_working_dir` rather than `<clone>/<subdir>`:

```rust
// ~line 1886
.spawn_node(
    node.clone(),
    base_working_dir,
    python_env_dir,
    false,                                // <- always unconfined
    ...
)
```

I believe this is fine today, since dynamically-added nodes don't go through the hub clone/build pipeline and have no `confined_nodes` entry to honor — but the hardcoded `false` means that *if* a hub-provenanced node ever reached this path, it would resolve its entrypoint against the ambient `$PATH`, silently bypassing the §11 confinement guarantee. A short comment documenting why this path is unconditionally unconfined (or a defensive `node.git`/provenance check that refuses a hub node here) would make the invariant explicit and harder to regress later.

Nothing else stood out.

---

---

PR #2180, opened 2026-06-12
feat(hub): discovery commands — search, info, list, fetch (P2.4+P2.11)

heyong4725:
Implements **P2.4 + P2.11** of the Dora Hub spec ([docs/plan-node-hub.md](https://github.com/dora-rs/dora/blob/main/docs/plan-node-hub.md) §9, umbrella #2097): the discovery-side `dora hub` commands and a hermetic end-to-end test. This completes the Phase-2 consumption surface.

**Stacked on #2179** — will retarget as the stack merges.

## What's included

- **`dora hub search <query> [--category C] [--platform P]`** — searches every configured index's catalog (latest version of each package), matching name/description/keywords, with category and platform facets.
- **`dora hub info <pkg>[@<ver>]`** — resolves and prints the typed inputs/outputs, env surface, and example snippet.
- **`dora hub list <dataflow.yml>`** — the hub packages pinned in a dataflow's lockfile (name, version, commit).
- **`dora hub fetch <dataflow.yml | pkg@ver> [--target-dir DIR]`** — warms the index cache (so `--offline` *resolution* works) and clones each pinned source into a portable directory.
- `examples/hub-dataflow/` (README + dataflow) and a docs/cli.md section.

All index-supplied strings are control-character-sanitized before printing; categories round-trip through their canonical kebab-case names; commit hashes are re-validated as hex at the clone boundary.

## Transport fix

The index self-heal heuristic switched from "empty worktree" to a `.dora-clone-complete` marker: a *sound* clone whose repository simply lacks the catalog subpath (e.g. the official `dora-rs/dora-hub` before its `node-index/` is bootstrapped) was being re-cloned on **every** invocation. It now errors clearly instead, and only a genuinely interrupted clone self-heals.

## Verification

- **`tests/hub-smoke.rs`** — a fully hermetic e2e test: builds its own git source repo + local `node-index` catalog + `hub.toml` in a tempdir, then drives `hub search`/`info`/`list`/`fetch`, `dora build --write-lockfile`, `dora run` (asserts node output), and `dora build --locked --offline`. No network, no live catalog, no Python bindings (a plain Rust node). Registered as a workspace test.
- Live-tested all four commands against a local fixture index.
- fmt + clippy `-D warnings` + full workspace tests + `cargo check --examples` + qa-fast.

## Honest scope

`dora hub fetch` guarantees offline *resolution* + a verifiable source mirror today. Full air-gapped *source* reuse by `dora build` (which clones through its own libgit2 build cache, a different layout) is a tracked follow-up — the help text and README say so rather than over-claiming. `--hub-override` (UC11) and custom-type shipping (P2.12) are also deferred to a focused follow-up.

Part of #2097.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -250,6 +250,10 @@ path = "tests/ws-cli-e2e.rs"
 name = "example-smoke"
 path = "tests/example-smoke.rs"
 
+[[test]]
+name = "hub-smoke"
+path = "tests/hub-smoke.rs"
+
 [profile.release]
 lto = "thin"
 
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -79,7 +79,7 @@ mod distributed;
 mod git;
 pub mod hub;
 mod local;
-mod lockfile;
+pub(crate) mod lockfile;
 
 #[derive(Debug, clap::Args)]
 /// Run build commands provided in the given dataflow.
--- a/binaries/cli/src/command/hub/common.rs
+++ b/binaries/cli/src/command/hub/common.rs
@@ -1,0 +1,68 @@
+//! Shared helpers for the discovery-side `dora hub` commands (P2.4/P2.11).
+
+use dora_hub_client::{
+    config::{IndexConfig, ResolvedConfig},
+    index::IndexCatalog,
+    transport::IndexFetcher,
+};
+use eyre::Context;
+
+/// Strip control characters from index-supplied strings before printing them
+/// (a hostile index entry could otherwise inject terminal escapes).
+pub fn sanitize(s: &str) -> String {
+    s.chars().filter(|c| !c.is_control()).collect()
+}
+
+/// Loaded hub configuration plus a fetcher, shared by the discovery commands.
+pub struct HubContext {
+    pub config: ResolvedConfig,
+    pub fetcher: IndexFetcher,
+}
+
+impl HubContext {
+    pub fn load(offline: bool) -> eyre::Result<Self> {
+        Ok(Self {
+            config: ResolvedConfig::load_default().context("failed to load hub configuration")?,
+            fetcher: IndexFetcher::new(offline)?,
+        })
+    }
+
+    /// Open the catalog the given namespace resolves against.
+    pub fn catalog_for_namespace(&mut self, namespace: &str) -> eyre::Result<IndexCatalog> {
+        let index = self.config.index_for_namespace(namespace).clone();
+        self.open(&index)
+    }
+
+    /// Open every configured index's catalog (for cross-namespace search).
+    pub fn all_catalogs(&mut self) -> Vec<(String, IndexCatalog)> {
+        let indexes = self.config.indexes.clone();
+        indexes
+            .into_iter()
+            .filter_map(|index| {
+                let alias = index.alias.clone();
+                match self.open(&index) {
+                    Ok(catalog) => Some((alias, catalog)),
+                    Err(err) => {
+                        eprintln!("  warning: skipping index `{alias}`: {err:#}");
+                        None
+                    }
+                }
+            })
+            .collect()
+    }
+
+    fn open(&mut self, index: &IndexConfig) -> eyre::Result<IndexCatalog> {
+        let dir = self
+            .fetcher
+            .catalog_dir(index, &self.config.config_dir)
+            .with_context(|| format!("failed to fetch index `{}`", index.alias))?;
+        IndexCatalog::open(&dir)
+    }
+
+    /// Print any accumulated transport warnings (e.g. index rollback).
+    pub fn drain_warnings(&mut self) {
+        for warning in self.fetcher.warnings.drain(..) {
+            eprintln!("  warning: {warning}");
+        }
+    }
+}
--- a/binaries/cli/src/command/hub/fetch.rs
+++ b/binaries/cli/src/command/hub/fetch.rs
@@ -1,0 +1,161 @@
+//! `dora hub fetch` — warm the index cache and mirror pinned node sources
+//! (UC7). Resolving every reference populates the on-disk index cache, so a
+//! later `dora build --offline` can resolve `hub:` references without the
+//! network; the sources are additionally cloned into a portable directory.
+//!
+//! Note: `dora build` clones git sources through its own (libgit2) build
+//! cache, so it does not yet re-use this mirror for the *source* fetch — full
+//! air-gapped source reuse is a follow-up (tracked under #2097). What this
+//! command guarantees today is offline *resolution* plus a verifiable local
+//! copy of each pinned source.
+
+use std::path::PathBuf;
+
+use super::common::HubContext;
+use crate::command::Executable;
+use dora_hub_client::reference::PackageRef;
+use dora_message::common::GitSource;
+use eyre::Context;
+
+/// Warm the index cache and mirror hub package sources
+#[derive(Debug, clap::Args)]
+pub struct Fetch {
+    /// A dataflow YAML, or a single `name@version` package reference
+    #[clap(value_name = "TARGET")]
+    target: String,
+    /// Directory to clone sources into (default: ./hub-cache)
+    #[clap(long, value_name = "DIR", default_value = "hub-cache")]
+    target_dir: PathBuf,
+}
+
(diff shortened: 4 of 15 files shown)


heyong4725:
Review finding:

- `binaries/cli/src/command/hub/fetch.rs:59`: despite the function name/comment saying it resolves dataflow pins, the dataflow path reads the YAML and calls `catalog.resolve(&reference)` for each `hub:` node (lines 67-80). It never reads the existing `.dora-lock.yaml`. If a lockfile pins `test/pkg` to 0.1.0 and the index later publishes 0.2.0 for the same `^0.1` range, `dora hub fetch dataflow.yml` mirrors 0.2.0, while `dora build --locked` uses the 0.1.0 commit from the lockfile (`binaries/cli/src/command/build/hub.rs:109-148`). That breaks the command contract of mirroring the pinned sources and can leave an air-gapped/CI bundle missing the exact commit the locked build needs. Please read the dataflow lockfile for dataflow targets, reusing the hub provenance entries like `hub list`, and only resolve live requirements for the single-package form.

Validation run: `cargo test -p dora-cli parse_hub`; `cargo test --test hub-smoke`; `git diff --check origin/feat/hub-descriptor-resolution...HEAD`.

heyong4725:
Fixed in `37e2f33d`. `dora hub fetch <dataflow.yml>` now reads the pinned `hub:` sources straight from the dataflow's lockfile (same contract as `dora hub list`) instead of re-resolving the index, so the mirror always matches what `dora build --locked` rebuilds — no drift if the index publishes a newer in-range version after the lock was written. Only the single-package `name@version` form (which has no lockfile) resolves live; the dataflow path no longer loads the index at all. Verified end-to-end by the hub-smoke test (build → lockfile → fetch).

phil-opp:
**Automated review by Claude** — fully automated, not vetted by a human; treat as suggestions. Note: the `dora-hub-client` crate sits on the unmerged base of this stack and isn't in this branch's tree, so the second item below is "worth confirming" rather than confirmed.

**Misleading error after an online reclone** — `libraries/hub-client/src/transport.rs`, the reworked `if !catalog.is_dir()` block. `incomplete` is captured once *before* the reclone:

```rust
let incomplete = !clone_dir.join(CLONE_COMPLETE_MARKER).exists();
if incomplete && !self.offline { self.reclone(...)?; }   // writes the marker
if !catalog.is_dir() {
    if incomplete { bail!("...incomplete (interrupted clone) and `--offline` is set — re-run with network access"); }
    bail!("...no `{catalog_subpath}` directory — check the index's `path` setting");
}
```

In online mode, when the clone was incomplete you reclone (which writes the marker and produces a sound clone), but if the repo genuinely lacks the catalog subpath, `catalog.is_dir()` is still false and `incomplete` is *still* true — so it bails with "interrupted clone … `--offline` is set" even though the run is online and just recloned. The intended "no `<catalog_subpath>` directory — check the `path` setting" message is then unreachable for a freshly-recloned sound clone. Suggest recomputing the marker state after `reclone`, or branching on `self.offline` instead of the stale `incomplete`. (Wrong diagnostic, not wrong behavior.)

Worth confirming: `search.rs` uses `catalog.versions(...).next_back()` to pick the latest version, which assumes `versions()` returns semver-ascending order — lexicographic ordering would mis-rank e.g. `0.9.0` vs `0.10.0`.

---

phil-opp:

Rebased onto the inherited `field_type_resolves` DoS fix (`7726c388`). Re the version-ordering item I flagged as "worth confirming" last time: confirmed a **non-issue** — `IndexCatalog::versions()` sorts `semver::Version` semantically and `search.rs` uses `next_back()`, so `0.10.0` correctly ranks above `0.9.0`. No new findings on these commits.

---

phil-opp:

`dora hub info` prints `platforms` without `sanitize()` — index-supplied terminal-escape injection (`binaries/cli/src/command/hub/info.rs:42`):

```rust
if !m.platforms.is_empty() {
    println!("  platforms: {}", m.platforms.join(", "));
}
```

Every other index-supplied field in this function is run through `sanitize()` (description, env names, ports, example snippet), but `platforms` is printed raw. The discovery path reads catalog entries via `IndexCatalog::entry`/`resolve`, which only call `NodeManifest::parse()` (YAML + size cap) and never `validate()` — so `check_platform()`'s charset check never runs on untrusted catalog content. A hostile/compromised index entry with `platforms: ["[2J…"]` reaches the terminal unescaped via `dora hub info <pkg>` — the exact class this PR hardens against everywhere else.

Suggest `m.platforms.iter().map(|p| sanitize(p)).collect::<Vec<_>>().join(", ")`.

---

heyong4725:
Fixed in `322cdeab`. `dora hub info` now sanitizes `platforms` like every other index-supplied field — the discovery path only `parse()`s catalog entries (never `validate()`), so `check_platform()` never runs on untrusted content and a terminal-escape in `platforms` would otherwise reach the terminal raw. (categories on the line above are a fixed `Category` enum, so not free-text.)

heyong4725:
**Pre-landing review** (Claude + Codex adversarial passes). Sanitization coverage in search/info/list, the fetch hash validation, and the transport self-heal are otherwise solid. Three findings fixed in `6e582927`:

- **Lockfile clone URLs not validated** (Codex High): `clone_pinned` passed the dataflow-path `repo` (read straight from the lockfile) to `git clone` without the scheme/transport allowlist the single-package path gets via `git_pin` — a crafted lockfile could drive an arbitrary git transport/local path. Now re-validates with `validate_git_url_untrusted` before cloning.
- **Fetch completion marker inside the clone** (both, High): `.dora-clone-complete` lived in the clone working tree, so a source repo shipping a file of that name could mask an interrupted checkout. Moved to `<target_dir>/.<commit>.complete`, outside the worktree.
- **Listed commit hash unsanitized** (Codex Low): `dora hub list` printed the lockfile commit prefix raw; `read_from` doesn't charset-validate `GitSource`, so a crafted lockfile could inject terminal escapes. Now sanitized.

Verified hub-smoke green (15s). Noted as follow-ups (Medium, pre-exploit — index doesn't exist yet): the `CLONE_COMPLETE_MARKER` self-heal edge cases (a partial `node-index/` with `.git` + source marker but no clone-complete marker accepted offline) and the unbounded package/version directory walk in `search`/`list` — both warrant a count bound + tighter marker checks under #2097.

heyong4725:
Both fixed in `16ec2471`:

- **`dora hub fetch <dataflow.yml>` never validated the YAML**: it now expands the descriptor and validates its `hub:` nodes against the lockfile before cloning — every hub node must be pinned, each reference's requirement must still be satisfied by the pinned version, and the lockfile must not pin a hub node the dataflow no longer has. This honors the "mirrors what `--locked` builds" contract (a stale lockfile errors instead of mirroring orphaned pins). The check is offline (no index fetch), matching how fetch already works.
- **`clone_pinned` trusted the completion marker alone**: now verifies the clone dir exists and actually holds the pinned commit (`git cat-file -e <commit>`); a deleted/corrupted clone with a leftover `.<commit>.complete` marker drops the stale marker and re-clones instead of reporting a phantom success.

hub-smoke now covers both (stale-marker re-clone + stale-lockfile rejection). Thanks for the complete sweep — all six PRs reviewed.

phil-opp:
Re-reviewed after `55404e8`, `a75766f0`, `16ec2471`. The `platforms` sanitization, the lockfile-URL re-validation + marker-outside-clone in `fetch`, the dataflow-vs-lockfile validation, and the "verify the pinned commit is actually present before trusting the `.complete` marker" check all hold up. No new issues there.

One earlier item appears still unaddressed (low priority, diagnostic-only — re-flagging only because the later commits didn't touch this block): in `libraries/hub-client/src/transport.rs`, the reworked `if !catalog.is_dir()` block captures `incomplete` *before* `reclone()` (which writes `CLONE_COMPLETE_MARKER`). In online mode, when an incomplete clone is recloned but the repo genuinely lacks the catalog subpath, `catalog.is_dir()` is still false and `incomplete` is still `true`, so it bails with "incomplete (interrupted clone) and `--offline` is set" — even though the run is online and just recloned. The intended "no `<catalog_subpath>` directory — check the `path` setting" message is unreachable for a freshly-recloned sound clone. Recomputing the marker state after `reclone` (or branching on `self.offline`) would give the right diagnostic. Behavior is correct; only the error message is misleading.

---

---

Issue #2083, opened 2026-06-08
Raw zero-copy path silently truncates length of zero-footprint arrays (e.g. NullArray) to 0


phil-opp:
> ⚠️ This issue was created by an automated, scheduled Claude code-review check. A maintainer should confirm before any fix is merged.

## Summary

When an Arrow array's *entire* serialized footprint is zero bytes but its logical length is **non-zero**, the raw (non-IPC) reconstruction path silently rebuilds it with **length 0**. This is silent data corruption (wrong length), not an error or panic.

The most common trigger is `NullArray::new(n)` with `n > 0`, but any array whose buffers and validity are all empty while `len > 0` is affected (e.g. a struct whose only fields are `Null`-typed children, or a zero-field struct without a validity buffer).

## Root cause

On the send side, `copy_array_into_sample` records the true length in `ArrowTypeInfo.len`, but a `NullArray` (and similar) has no data buffers and no validity, so `required_data_size` is `0` and the payload is empty.

On the receive side, `buffer_into_arrow_array_inner` early-returns for an empty buffer and rebuilds the array with `ArrayData::new_empty(&type_info.data_type)`, which **discards `type_info.len`** (and `offset`, `null_count`, `validity`, `child_data`):

```rust
// apis/rust/node/src/node/arrow_utils.rs:148-150
if raw_buffer.is_empty() {
    return Ok(arrow::array::ArrayData::new_empty(&type_info.data_type));
}
```

`new_empty` produces a length-0 array, so `NullArray::new(5)` decodes back as `NullArray` of length **0**.

The earlier #2027 work that introduced this branch was specifically aimed at preserving the declared **type** (so a downstream `as_primitive` wouldn't panic on a zero-length typed array). Preserving the **length** for zero-footprint arrays was overlooked.

## Affected sites (same defect, three copies)

- `apis/rust/node/src/node/arrow_utils.rs:148-150` — `buffer_into_arrow_array_inner` (daemon / local raw path, also used by the Python operator API at `apis/python/operator/src/lib.rs`).
- `apis/rust/node/src/event_stream/mod.rs:1087-1108` — `zenoh_payload_to_arrow_array` (distributed/zenoh path) uses the same `new_empty(data_type)` mirror.
- `binaries/cli/src/command/topic/echo.rs:275-277` — a third hand-rolled copy with the same `raw_buffer.is_empty()` → `new_empty` branch.

[...]

PR #2181, opened 2026-06-12
fix(cli): make `dora topic echo` render errors instead of panicking on undecodable data

phil-opp:

## Issue

The data-rendering block of `dora topic echo` (`binaries/cli/src/command/topic/echo.rs`) had three panic paths on peer-controlled input — in a file whose *decode* half was already hardened against exactly this threat model (#2083):

1. **`writer.write(&batch).unwrap()` / `writer.finish().unwrap()`** — the Arrow JSON writer does not support every Arrow type a node can legitimately send. A union array, for example, fails with `Unsupported data type for JSON encoding` (verified against arrow-json 58's encoder), so echoing such a topic killed the entire echo stream with a panic.
2. **`array.len() as _` (to `i32`)** — `type_info.len` is peer-controlled, and a `NullArray` carries an arbitrary `len` with **no** backing buffers to bound it. A value above `i32::MAX` wrapped negative and panicked inside `OffsetBuffer::new` ("offsets must be monotonically increasing").
3. **`buf[4..buf.len() - 2]`** — assumed the writer's `{"":[...]}\n` output shape without checking it.

## Fix

Rendering is extracted into `decode_and_render` / `render_array_json`, which surface all of these as errors. The stream now prints a per-message diagnostic and continues — the same behavior the existing `invalid data on ...` decode path already has. On success, `render_array_json` guarantees the `{"":` prefix and `}\n` suffix, making the call-site slice provably in-bounds.

## Validation

- New unit tests: happy path (`[1,2,3]` round-trip), unsupported-JSON-type (dense union → error, previously panic), oversized `NullArray` length (`i32::MAX + 1` → error, previously panic).
- `cargo test -p dora-cli --lib` (7 echo/topic tests), `cargo clippy -p dora-cli -- -D warnings`, `cargo fmt --check`, typos, unwrap-budget all green; full workspace fmt + clippy + test suite run on the combined review diff.

---

### DIFF
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -171,32 +171,12 @@ fn inspect(
                     .as_millis();
 
                 let data_str = if let Some(data) = data {
-                    let array = match NonNull::new(data.as_ptr() as *mut u8)
-                        .ok_or_else(|| eyre!("payload pointer is null"))
-                        .and_then(|ptr| {
-                            let len = data.len();
-                            let buffer = unsafe {
-                                arrow::buffer::Buffer::from_custom_allocation(
-                                    ptr,
-                                    len,
-                                    Arc::new(data),
-                                )
-                            };
-                            buffer_into_arrow_array(&buffer, &metadata.type_info)
-                        }) {
-                        Ok(array) => array,
-                        Err(e) => {
-                            eprintln!("invalid data on {output_name}: {e}");
-                            continue;
-                        }
-                    };
-
-                    match render_array_json(array, &mut buf) {
+                    match decode_and_render(data, &metadata.type_info, &mut buf) {
                         // `render_array_json` guarantees the `{"":` prefix and
                         // `}\n` suffix, so this slice cannot go out of bounds.
                         Ok(()) => std::str::from_utf8(&buf[4..buf.len() - 2]).ok(),
                         Err(e) => {
-                            eprintln!("cannot render data on {output_name}: {e}");
+                            eprintln!("invalid data on {output_name}: {e}");
                             continue;
                         }
                     }
@@ -266,6 +246,20 @@ fn inspect(
     }
 
     Ok(())
+}
+
+/// Reconstruct the Arrow array from a raw payload and render it into `buf`.
+fn decode_and_render(
+    data: dora_message::aligned_vec::AVec<u8, dora_message::aligned_vec::ConstAlign<128>>,
+    type_info: &ArrowTypeInfo,
+    buf: &mut Vec<u8>,
+) -> eyre::Result<()> {
+    let ptr =
+        NonNull::new(data.as_ptr() as *mut u8).ok_or_else(|| eyre!("payload pointer is null"))?;
+    let len = data.len();
+    let buffer = unsafe { arrow::buffer::Buffer::from_custom_allocation(ptr, len, Arc::new(data)) };
+    let array = buffer_into_arrow_array(&buffer, type_info)?;
+    render_array_json(array, buf)
 }
 
 /// Render a decoded Arrow array into `buf` as `{"":[...]}\n`.
--- a/binaries/cli/src/command/topic/echo.rs
+++ b/binaries/cli/src/command/topic/echo.rs
@@ -171,12 +171,19 @@ fn inspect(
                     .as_millis();
 
                 let data_str = if let Some(data) = data {
-                    let ptr = NonNull::new(data.as_ptr() as *mut u8).unwrap();
-                    let len = data.len();
-                    let buffer = unsafe {
-                        arrow::buffer::Buffer::from_custom_allocation(ptr, len, Arc::new(data))
-                    };
-                    let array = match buffer_into_arrow_array(&buffer, &metadata.type_info) {
+                    let array = match NonNull::new(data.as_ptr() as *mut u8)
+                        .ok_or_else(|| eyre!("payload pointer is null"))
+                        .and_then(|ptr| {
+                            let len = data.len();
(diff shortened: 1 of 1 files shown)


phil-opp:
## 🤖 Automated repository review — run summary (2026-06-12)

This run scanned origin/main (`1312143`) across daemon, coordinator, CLI, node/operator APIs, message/core/arrow libraries, and the Python/C/C++ bindings. Open PRs (#2174–#2180 hub stack, #2168, #2132, #2127, #1796) were checked first to avoid duplicates. 4 findings met the bar; PRs below.

### Opened PRs

**Correctness**
- #2181 — `dora topic echo`: three panic paths on peer-controlled data (JSON-writer `unwrap` on unsupported Arrow types, `as i32` wrap of a peer-controlled length, unchecked slice). *(this PR)*
- #2182 — Python `setup_logging` re-wraps `logging.basicConfig` on every `Node()`; second node in one interpreter ⇒ `RecursionError` on any `basicConfig` call.

**Diagnostics**
- #2184 — `check_input` reported "operator `` does not exist" when a runtime-node mapping lacks the `<operator_id>/` segment; now states the expected format.

**Documentation**
- #2183 — 16 broken/redundant intra-doc links across dora-node-api / dora-message / dora-core (`cargo doc` now warning-free).

### Validation
Per branch: affected-crate tests (TDD where applicable — regression tests written first and confirmed failing), clippy `-D warnings`, fmt; #2182 validated end-to-end against a locally built wheel. Combined diff: workspace fmt + clippy + typos + unwrap-budget, `cargo test` for all crates with test targets (87 suites green; the 35 test-less example-binary crates were compile-covered via workspace clippy — the review container's disk cannot hold their ~430 MB-each debug link artifacts), doctests for the four touched crates, `cargo check --examples`, and `/review`+`/simplify` passes (one readability refactor applied to #2181).

### Considered and rejected
- `DelayedCleanup::get_mut` `try_lock().expect` (py operator API): switching to a blocking lock risks a GIL-vs-mutex deadlock; the panic surfaces as a catchable `PanicException` — net-negative change.
- `logging.index()` "wrong API" (scanner claim): false positive — PyO3's `PyModule::index()` *is* the `__all__` accessor.
- `topic info` `.next().unwrap()`: false positive — guarded by the `is_empty()` bail just above.
- Hot-reload watcher `unwrap` (`start/attach.rs`): panic confined to notify's callback thread during a shutdown race; cosmetic.
- Recording header length casts (`len() as u16/u32`): truncate only at absurd sizes (>64 KiB node ids, >4 GiB events).
- Telemetry context parser keeps empty-string keys: phantom entries are never queried; no observable effect.
- Per-event `input_type_checks.get(id).cloned()`: check is consumed at the first non-pattern message per input; negligible.
- Defensive `NonNull`/serialization `unwrap` rewrites (node-api zenoh paths, `topic pub`, ws_client): values provably non-null / serialization cannot fail.
- Pre-epoch timestamp `unwrap_or(UNIX_EPOCH)` in py metadata: data outside supported range; silent-clamp behavior matches the arrow-convert convention.
- `echo.rs`'s local `buffer_into_arrow_array` duplicates `dora-node-api::arrow_utils`: pre-existing duplication; consolidating couples dora-cli to node-API internals — left as a possible follow-up.
- C++ operator `as_mut().unwrap()` on `UniquePtr`: `new_operator` contract guarantees non-null today.
- Daemon and coordinator deep scans surfaced no concrete, verifiable defects.

---

phil-opp:

No issues found. The fix correctly converts all three panic paths (the JSON-writer `unwrap`s, the `array.len() as i32` wrap, and the unchecked output slice) into recoverable per-message errors that print a diagnostic and `continue`, matching the existing decode-side behavior. The slice safety is sound: `buf` is cleared at the top of each loop iteration, so the `{"":`-prefix / `}\n`-suffix invariant enforced in `render_array_json` keeps the `buf[4..len-2]` slice provably in-bounds. The three new tests exercise the paths they claim (the `NullArray::new(i32::MAX+1)` test is cheap since null arrays carry no backing buffers, so it exercises the `try_from` guard without OOM).

---

phil-opp:
placeholder

---

phil-opp:

No issues found. The render path is correctly hardened: `buf` is cleared each iteration (`echo.rs:139`) so it never accumulates, and `render_array_json` guarantees the `{"":` prefix / `}\n` suffix (with a `buf.len() < 6` guard) before the call site slices `buf[4..buf.len()-2]`, so that slice is provably in-bounds. The `i32::try_from(len)`, unsupported-JSON-type, and output-shape checks all surface errors and `continue` instead of panicking, matching the existing `invalid data on ...` decode behavior. Tests cover all three former panic paths.

---

phil-opp:
**Automated review by Claude** — fully automated review, not human-verified.

No new issues in the code — the `decode_and_render` / `render_array_json` refactor is sound, the three former panic paths (writer `unwrap`, `len as i32` wrap, unchecked `buf[4..len-2]` slice) are correctly converted to recoverable per-message errors, and the prefix/suffix guard makes the call-site slice provably in-bounds.

The current red merge-queue check is **unrelated to this PR's changes**. The Audit (cargo-audit) job now fails on **RUSTSEC-2026-0185** — quinn-proto 0.11.14, "Remote memory exhaustion … from unbounded out-of-order stream reassembly", 7.5 (high), advisory published 2026-06-22 (after this PR's own per-commit CI last ran green). It's a transitive-dependency issue; the fix is bumping `quinn-proto` to `>= 0.11.15` in `Cargo.lock` (sibling PRs #2303/#2304 already carry that bump). Once that bump is on `main` and this branch is rebased, the audit gate should clear — nothing in `echo.rs` needs to change for it.

---

PR #2182, opened 2026-06-12
fix(python): install the logging `basicConfig` wrapper only once per process

phil-opp:

## Issue

`setup_logging` (`apis/python/node/src/lib.rs`) runs on **every** `dora.Node` construction. The wrapper it installs resolves `oldBasicConfig` from the logging module dict *at call time*, so a second `Node()` in the same interpreter — notebooks, pytest suites, sequential dynamic-node sessions — rebinds `oldBasicConfig` to wrapper #1. Wrapper #1 then calls itself through that same lookup:

- every subsequent `logging.basicConfig()` — **including the implicit one inside `logging.info(...)`** — dies with `RecursionError`;
- `logging.__all__` accumulates duplicate `"HostHandler"` entries.

## Fix

Guard the install with the `oldBasicConfig` attribute, which exists exactly when the wrapper has already been installed (no new sentinel needed — the marker *is* the invariant being protected).

## Validation

Validated against a locally built debug wheel (`maturin develop`) using the daemon-free `DORA_TEST_WITH_INPUTS` integration-test mode, constructing two `Node`s in one interpreter and then calling `logging.basicConfig(level=INFO)`:

- **before** this change: `RecursionError` (reproduced);
- **after**: succeeds, with exactly one `HostHandler` attached.

Also: `cargo check -p dora-node-api-python`, `cargo fmt --check`, typos, and the full workspace fmt + clippy + test gates on the combined review diff. (The crate is a cdylib excluded from workspace tests, hence the wheel-based runtime validation.)

---

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -83,6 +83,15 @@ fn host_log<'py>(record: Bound<'py, PyAny>) -> PyResult<()> {
 /// delivered to `crate::host_log`, which will send them to `tracing::event!`.
 pub fn setup_logging(py: Python) -> PyResult<()> {
     let logging = py.import("logging")?;
+    // `setup_logging` runs on every `Node` construction. Installing the
+    // `basicConfig` wrapper a second time would wrap the wrapper (the new
+    // `oldBasicConfig` would be the first wrapper), so every `basicConfig`
+    // call would attach one `HostHandler` per `Node` ever created in this
+    // process and emit duplicate log records. `oldBasicConfig` only exists
+    // once the wrapper is installed, so use it as the install marker.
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The guard is correct: `oldBasicConfig` is assigned into the logging module dict (the `py.run` globals are `logging.__dict__`), so `logging.hasattr("oldBasicConfig")` reliably signals the wrapper is already installed, and the early return prevents the re-wrap that caused the `RecursionError` and duplicate `HostHandler` entries. The check-then-set runs under `Python::attach` (GIL held), so it's atomic against other Python threads — no TOCTOU race. Using the existing `oldBasicConfig` as the marker (rather than a separate sentinel) keeps the install flag in sync with the actual invariant.

---

PR #2184, opened 2026-06-12
fix(core): report expected format when a runtime-node input mapping lacks the operator segment

phil-opp:

## Issue

`check_input` (`libraries/core/src/descriptor/validate.rs`) parsed runtime-node output references with `split_once('/').unwrap_or_default()`. A mapping like `runtime-node/out` — missing the operator id — silently fell back to an **empty** operator id and failed with:

```
source operator `runtime-node/` used for input `sink/in` does not exist
```

which points the user at a nonexistent empty operator instead of the actual mistake (the missing `<operator_id>/` segment).

## Fix

Bail out explicitly when the reference has no `/`, telling the user the expected `<node>/<operator_id>/<output_id>` format.

## Validation

- TDD: regression test asserting the new message was written first and failed against the old code; a positive test covers the well-formed three-segment mapping.
- `cargo test -p dora-core` (167 lib tests + doctests), `cargo clippy -p dora-core -- -D warnings`, `cargo fmt --check` green; grep confirms no other code or test matches the old message text; full workspace fmt + clippy + test gates run on the combined review diff.

---

### DIFF
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -370,7 +370,18 @@ fn check_input(
                     }
                 }
                 CoreNodeKind::Runtime(runtime) => {
-                    let (operator_id, output) = output.split_once('/').unwrap_or_default();
+                    // Outputs of runtime nodes are referenced as
+                    // `<node>/<operator>/<output>`. Without the operator
+                    // segment, falling back to an empty operator id would
+                    // produce a misleading "operator `` does not exist" error.
+                    let Some((operator_id, output)) = output.split_once('/') else {
+                        bail!(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The change correctly replaces the `split_once('/').unwrap_or_default()` fallback (which produced a misleading "operator `` does not exist" message) with an explicit `bail!` naming the expected `<node>/<operator_id>/<output_id>` format. The `{output}` in the new message refers to the original outer `DataId` (the user's full reference), which is the right value to surface — the `Some`-path binding only shadows it on the success branch. The regression and positive tests cover both the malformed and well-formed cases.

---

PR #2191, opened 2026-06-13
fix(inference): read arg index 2 for send_output data array

### LINKED ISSUE #2150: Type inference reads wrong `send_output` argument (metadata instead of data), so it never fires on real Rust nodes
## Summary

The type-inference scanner in `libraries/core/src/inference.rs` (feature `type-inference`) inspects the **wrong positional argument** of `send_output`. It treats argument index `1` as the Arrow data array, but in the actual Rust node API the data is argument index `2`. Index `1` is the `MetadataParameters`. As a result the inference inspects the metadata-parameters expression instead of the data array, and produces no/incorrect suggestions for every real dora Rust node.

The unit tests in the same file mask this because they exercise a **two-argument `send_output` signature that does not exist** in the dora Rust API.

> _This issue was created by an automated, scheduled Claude code-review check._

## The mismatch

Real Rust API (`apis/rust/node/src/node/mod.rs:798`):

```rust
pub fn send_output(
    &mut self,
    output_id: DataId,
    parameters: MetadataParameters,   // <-- arg index 1
    data: impl Array,                 // <-- arg index 2 (the Arrow array)
) -> NodeResult<()>
```

Every real call site uses the `(output_id, parameters, data)` order, e.g.:

- `examples/error-propagation/producer/src/main.rs:10`
  `node.send_output("data".into(), MetadataParameters::default(), i.into_arrow())?;`
- `examples/multiple-daemons/node/src/main.rs:31`
  `node.send_output(output.clone(), metadata.parameters, random.into_arrow())?;`
- `examples/mavlink2-bridge/heartbeat-emitter/src/main.rs:43`
  `node.send_output(output.clone(), MetadataParameters::default(), arr)`

## The bug

`libraries/core/src/inference.rs:82-86`:

```rust
if method_name == "send_output" && node.args.len() >= 2 {
    // Try to extract output_id from first argument (string literal)
    if let Some(output_id) = extract_string_lit(&node.args[0]) {
        // Try to infer type from second argument
        if let Some(urn) = infer_type_from_expr(&node.args[1]) {   // <-- WRONG INDEX
```

`node.args[1]` is the `MetadataParameters` argument (`MetadataParameters::default()`, `metadata.parameters`, `params`, …), never the Arrow array. `infer_type_from_expr` therefore returns `None` for all idiomatic code, so the scanner emits no suggestions. The data array it is supposed to read lives at `node.args[2]`.

## Why the tests don't catch it

`libraries/core/src/inference.rs:174-186` asserts against a fictional 2-arg form:

```rust
node.send_output("reading", Float64Array::from(vec![42.0]));
```

This signature does not exist in `dora-node-api` (data is never the 2nd arg), so the test passes while the real-world behavior is broken.

## Suggested fix

Read the data argument at index `2`, and update the tests to use the real 3-argument signature, e.g.:

```rust
if method_name == "send_output" && node.args.len() >= 3 {
    if let Some(output_id) = extract_string_lit(&node.args[0]) {
        if let Some(urn) = infer_type_from_expr(&node.args[2]) {
            ...
```

## Secondary note (same code path)

Even after fixing the index, the most common idiomatic data expression — `value.into_arrow()` — would still not be recognized. `infer_type_from_expr` only follows method calls named `into`/`from` (`inference.rs:131-139`), so `.into_arrow()` returns `None`. Recognizing `into_arrow()` (and adding the missing `Int8Array`/`Int16Array`/`UInt16Array` mappings in `map_constructor_to_urn`) would make the inference actually useful on real nodes. These can be folded into the same fix or tracked separately.

## References

- `libraries/core/src/inference.rs:82` (https://github.com/dora-rs/dora/blob/563b5d38b1c194d234f57c3ba0c5d3778e9f7e59/libraries/core/src/inference.rs#L80-L96)
- `apis/rust/node/src/node/mod.rs:798` (https://github.com/dora-rs/dora/blob/563b5d38b1c194d234f57c3ba0c5d3778e9f7e59/apis/rust/node/src/node/mod.rs#L798-L803)

Supull:
hey addressing issue #2150 the type inference scanner was reading argument index 1 of send_output which is MetadataParameters instead of index 2 the actual Arrow data array so type inference was silently doing nothing on every real Rust node. also added into_arrow() support and the missing Int8/Int16/UInt16 mappings while i was in there

fixes #2150

### DIFF
--- a/libraries/core/src/inference.rs
+++ b/libraries/core/src/inference.rs
@@ -7,6 +7,7 @@
 //! Feature-gated behind `type-inference`.
 
 use std::path::Path;
+
 use syn::{Expr, ExprMethodCall, visit::Visit};
 
 /// A suggested type annotation inferred from source code.
@@ -60,7 +61,6 @@ pub fn infer_types_from_source(
 ) -> Result<Vec<TypeSuggestion>, String> {
     let syntax =
         syn::parse_file(source).map_err(|e| format!("failed to parse {file_name}: {e}"))?;
-
     let mut visitor = TypeInferenceVisitor {
         suggestions: Vec::new(),
         node_id: node_id.to_string(),
@@ -79,11 +79,14 @@ struct TypeInferenceVisitor {
 impl<'ast> Visit<'ast> for TypeInferenceVisitor {
     fn visit_expr_method_call(&mut self, node: &'ast ExprMethodCall) {
         let method_name = node.method.to_string();
-        if method_name == "send_output" && node.args.len() >= 2 {
+        // The real Rust node API signature is:
+        //   send_output(output_id, parameters: MetadataParameters, data: impl Array)
+        // so the data (Arrow array) is at argument index 2, not 1.
+        if method_name == "send_output" && node.args.len() >= 3 {
             // Try to extract output_id from first argument (string literal)
             if let Some(output_id) = extract_string_lit(&node.args[0]) {
-                // Try to infer type from second argument
-                if let Some(urn) = infer_type_from_expr(&node.args[1]) {
+                // Try to infer type from third argument (the Arrow data array)
+                if let Some(urn) = infer_type_from_expr(&node.args[2]) {
                     self.suggestions.push(TypeSuggestion {
                         node_id: self.node_id.clone(),
                         output_id,
@@ -94,6 +97,7 @@ impl<'ast> Visit<'ast> for TypeInferenceVisitor {
                 }
             }
         }
+
         // Continue visiting nested expressions
         syn::visit::visit_expr_method_call(self, node);
     }
@@ -111,7 +115,8 @@ fn extract_string_lit(expr: &Expr) -> Option<String> {
     }
 }
 
-/// Try to infer a type URN from an expression (e.g. `Float64Array::from(...)`).
+/// Try to infer a type URN from an expression (e.g. `Float64Array::from(...)`
+/// or `value.into_arrow()`).
 fn infer_type_from_expr(expr: &Expr) -> Option<String> {
     match expr {
         Expr::Call(call) => {
@@ -129,10 +134,16 @@ fn infer_type_from_expr(expr: &Expr) -> Option<String> {
             None
         }
         Expr::MethodCall(mc) => {
-            // Look for .into() on typed values, or ArrowArray constructors
             let method = mc.method.to_string();
             if method == "into" || method == "from" {
                 // Try to get type from receiver
+                return infer_type_from_expr(&mc.receiver);
+            }
+            // Recognise the idiomatic `.into_arrow()` call: the type information
+            // lives in the receiver (e.g. `Float64Array::from(...).into_arrow()`
+            // or a named variable — the latter we cannot resolve statically, but
+            // we can still handle the constructor chain form).
+            if method == "into_arrow" {
                 return infer_type_from_expr(&mc.receiver);
             }
             None
@@ -156,9 +167,12 @@ fn map_constructor_to_urn(name: &str) -> Option<String> {
     match name {
         "Float32Array" | "Float32Array::from" => Some("std/core/v1/Float32".into()),
         "Float64Array" | "Float64Array::from" => Some("std/core/v1/Float64".into()),
+        "Int8Array" | "Int8Array::from" => Some("std/core/v1/Int8".into()),
+        "Int16Array" | "Int16Array::from" => Some("std/core/v1/Int16".into()),
         "Int32Array" | "Int32Array::from" => Some("std/core/v1/Int32".into()),
         "Int64Array" | "Int64Array::from" => Some("std/core/v1/Int64".into()),
         "UInt8Array" | "UInt8Array::from" => Some("std/core/v1/UInt8".into()),
+        "UInt16Array" | "UInt16Array::from" => Some("std/core/v1/UInt16".into()),
         "UInt32Array" | "UInt32Array::from" => Some("std/core/v1/UInt32".into()),
         "UInt64Array" | "UInt64Array::from" => Some("std/core/v1/UInt64".into()),
         "StringArray" | "StringArray::from" => Some("std/core/v1/String".into()),
@@ -171,13 +185,21 @@ fn map_constructor_to_urn(name: &str) -> Option<String> {
 mod tests {
     use super::*;
 
+    // ── primary fix: 3-argument real API ────────────────────────────────────
+
     #[test]
-    fn infer_float64_from_send_output() {
+    fn infer_float64_from_send_output_real_api() {
+        // Uses the actual 3-arg dora Rust API signature:
+        //   send_output(output_id, MetadataParameters::default(), data)
         let source = r#"
-            fn main() {
-                node.send_output("reading", Float64Array::from(vec![42.0]));
-            }
-        "#;
+fn main() {
+    node.send_output(
+        "reading",
+        MetadataParameters::default(),
+        Float64Array::from(vec![42.0]),
+    );
+}
+"#;
         let suggestions = infer_types_from_source(source, "test.rs", "sensor").unwrap();
         assert_eq!(suggestions.len(), 1);
         assert_eq!(suggestions[0].output_id, "reading");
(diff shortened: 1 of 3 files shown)


phil-opp:

No issues found. The core fix is correct: the real Rust node API is `send_output(output_id, parameters, data)`, so the Arrow data array is at argument index 2 — the previous `args.len() >= 2` / `args[1]` matched the parameters slot (or a fictional 2-arg form) and would have inferred a type from the wrong argument. Requiring `>= 3` args and reading `args[2]` aligns the visitor with the actual signature, and `two_arg_send_output_does_not_fire` pins that the old form no longer produces a false positive.

The `into_arrow()` handling recurses into the receiver, which resolves the constructor-chain form (`Float32Array::from(...).into_arrow()` → `Float32Array::from` → `map_constructor_to_urn`) while gracefully yielding no suggestion for an opaque receiver (`random.into_arrow()`) rather than panicking — both covered by tests. The added Int8/Int16/UInt16 mappings match the existing URN scheme. Change is confined to the feature-gated `type-inference` helper, so no runtime impact.

---

phil-opp:

The core index fix is correct: `send_output(output_id, parameters: MetadataParameters, data: impl Array)` (`apis/rust/node/src/node/mod.rs:802`) genuinely carries the Arrow data at argument index 2, and every real example call uses the 3-arg form — so the change to `args.len() >= 3` / `args[2]` introduces no regression, and the `into_arrow()` handling is sound.

One issue with the new type mappings in `map_constructor_to_urn`: the URNs `std/core/v1/Int8`, `std/core/v1/Int16`, and `std/core/v1/UInt16` are not defined in the type registry — `types/std/core/v1.yml` only defines `Float32/Float64/Int32/Int64/UInt8/UInt32/UInt64/String/Bytes/Bool`. So a suggestion emitting e.g. `std/core/v1/Int16` would not resolve if a user adopted it. Either add `Int8`/`Int16`/`UInt16` (all valid Arrow types) to `types/std/core/v1.yml`, or drop those three mappings until the registry supports them.

Impact is advisory-only today, since the `inference` module is feature-gated (`type-inference`) and currently has no production caller — but it's worth fixing before that path is wired up, as the existing "No issues found" note above missed it.

---

Supull:
> 🤖 **Automated review by Claude** — this is a fully automated code review; no human has vetted these comments. Treat accordingly.
> 
> The core index fix is correct: `send_output(output_id, parameters: MetadataParameters, data: impl Array)` (`apis/rust/node/src/node/mod.rs:802`) genuinely carries the Arrow data at argument index 2, and every real example call uses the 3-arg form — so the change to `args.len() >= 3` / `args[2]` introduces no regression, and the `into_arrow()` handling is sound.
> 
> One issue with the new type mappings in `map_constructor_to_urn`: the URNs `std/core/v1/Int8`, `std/core/v1/Int16`, and `std/core/v1/UInt16` are not defined in the type registry — `types/std/core/v1.yml` only defines `Float32/Float64/Int32/Int64/UInt8/UInt32/UInt64/String/Bytes/Bool`. So a suggestion emitting e.g. `std/core/v1/Int16` would not resolve if a user adopted it. Either add `Int8`/`Int16`/`UInt16` (all valid Arrow types) to `types/std/core/v1.yml`, or drop those three mappings until the registry supports them.
> 
> Impact is advisory-only today, since the `inference` module is feature-gated (`type-inference`) and currently has no production caller — but it's worth fixing before that path is wired up, as the existing "No issues found" note above missed it.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

added Int8/Int16/UInt16 to types/std/core/v1.yml so the URNs actually resolve in the registry

phil-opp:
🤖 _Automated review by Claude — this is a fully automated review with no human in the loop._

Re-reviewed after the latest push. The registry gap flagged in the earlier review is now closed: `d5dd755` adds `Int8`/`Int16`/`UInt16` to `types/std/core/v1.yml`, so the URNs emitted by `map_constructor_to_urn` now resolve. The keys match the URN suffixes (`std/core/v1/Int8` etc.) and the Arrow type names are valid.

No remaining issues from my side. The core index fix is correct against the real signature `send_output(output_id, parameters: MetadataParameters, data: impl Array)` (`apis/rust/node/src/node/mod.rs:802`) — the data array really is at argument index 2, so `args.len() >= 3` / `args[2]` is right, and the `two_arg_send_output_does_not_fire` test pins the old false-positive form. The `into_arrow()` handling recurses into the receiver and degrades gracefully (no suggestion, no panic) for opaque receivers, both covered by tests. Latest commits look safe to me.

---

phil-opp:
I took a closer look at the new `Int8`/`Int16`/`UInt16` types and found one issue that the earlier registry-gap discussion didn't fully close.

Adding the entries to `types/std/core/v1.yml` makes them exist as registry *entries*, but they still don't resolve to an Arrow `DataType` for runtime validation. `arrow_type_from_name` in `libraries/core/src/types.rs:97-111` only matches `Float32/Float64/Int32/Int64/UInt8/UInt32/UInt64/Utf8/LargeBinary/Boolean` — `"Int8"`, `"Int16"`, and `"UInt16"` fall through to `_ => None`. So `TypeDef::arrow_data_type()` / `resolve_field_type()` return `None` for these, which the code treats as "skip runtime validation."

Net effect: an inferred `std/core/v1/Int8` annotation gets no runtime type checking, unlike the `Int32`/`Float64` cases this PR is meant to mirror. To wire them through fully, add the missing arms:

```rust
"Int8"   => Some(DataType::Int8),
"Int16"  => Some(DataType::Int16),
"UInt16" => Some(DataType::UInt16),
```

The core fix itself looks correct: reading `args[2]` for the data array matches the real `send_output(output_id, parameters: MetadataParameters, data: impl Array)` signature (`apis/rust/node/src/node/mod.rs`), and the `into_arrow()` receiver handling is sound. No issues there.

---

---

phil-opp:
Re-reviewed at head `bb9899b5`. The new commit since the last review adds the `Int8`/`Int16`/`UInt16` arms to `arrow_type_from_name` (`types.rs`), which is exactly the gap flagged previously — the three layers (the `map_constructor_to_urn` constructor table in `inference.rs`, the `types/std/core/v1.yml` registry, and `arrow_type_from_name`) are now consistent for those types. I re-checked the core fix as well: in a `node.send_output(a, b, c)` method call, syn's `ExprMethodCall.args` holds `[output_id, parameters, data]` (the receiver is separate), so reading index 2 is genuinely the Arrow data array, and the `>= 3` guard correctly stops the old 2-arg false-positive form from firing. No new issues found; looks safe to merge from a correctness standpoint.

---
*This is a fully automated review by Claude (Claude Code). No human has vetted these findings — please verify before acting on them.*

---

phil-opp:

The `args[2]` index fix is correct (I confirmed `send_output(output_id, parameters: MetadataParameters, data: impl Array)` at `apis/rust/node/src/node/mod.rs`, and in a syn `ExprMethodCall` the receiver is separate, so the data array is `args[2]`), and the `Int8`/`Int16`/`UInt16` additions are consistent across the three tables. But I don't think this PR actually fixes #2150 — inference still produces zero suggestions for real Rust nodes, for a reason upstream of the index that was changed.

Output IDs are extracted by `extract_string_lit` (`libraries/core/src/inference.rs:103`), which only matches `Expr::Lit(Str)` and `Expr::Reference`. However, `send_output`'s first parameter is a `DataId`, so no real call site passes a bare string literal — they all use `"x".into()` (an `Expr::MethodCall`) or a variable:

- `examples/error-propagation/producer/src/main.rs:10` — `send_output("data".into(), ...)`
- `examples/action-example/client/src/main.rs:28` — `send_output("goal".into(), ...)`
- `examples/rust-dataflow/node/src/main.rs:47` — `send_output(output.clone(), ...)`

For each of these, `extract_string_lit` returns `None`, so the visitor never reaches the (now-correct) `args[2]` inference. End to end, inference remains the no-op that #2150 describes — the bug is just relocated from the data argument to the output-id argument.

This is also why the new `infer_from_metadata_variable` test passes: its first arg is `output.clone()`, so it short-circuits at output-id extraction and never exercises the `into_arrow()` opaque-receiver path the comment claims it tests. A test whose first arg is a string-literal-via-`.into()` (the real idiom) would currently fail.

To make inference fire on real code, `extract_string_lit` needs to recognize `"x".into()` (a string-literal receiver of `.into()`). Worth resolving before this path is wired to a caller, otherwise #2150 isn't closed in practice.

---
_Generated by Claude Code_

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Please verify before acting.

Re-reviewed at head `7a062be`, which landed after my previous review. The output-id gap I flagged last time is now closed: `extract_string_lit` recognizes the `"x".into()` `DataId` idiom (`Expr::MethodCall` with `method == "into"`, recursing into the receiver), so inference now fires on the real call-site form used throughout `examples/` (e.g. `send_output("data".into(), MetadataParameters::default(), x.into_arrow())`) rather than silently no-opping.

End to end the fix is correct and consistent across all three layers:
- `inference.rs` reads `args[2]` (the data array — confirmed against `send_output(output_id, parameters, data)` at `apis/rust/node/src/node/mod.rs:802`), with the `>= 3` guard ruling out the old 2-arg false positive; `.into()` output-id and `.into_arrow()` data handling both recurse correctly and degrade to "no suggestion" (not a panic) for opaque receivers.
- `map_constructor_to_urn`, `types/std/core/v1.yml`, and `arrow_type_from_name` are all consistent for the new `Int8`/`Int16`/`UInt16` types, so suggestions both resolve in the registry and get runtime validation.

No remaining issues found — #2150 is now actually addressed (inference fires on real Rust nodes).

_Generated by Claude Code_

---

PR #2185, opened 2026-06-13
refactor(tracing): remove unused add_layer/with_layers builder methods

phil-opp:
> 🤖 **This PR is auto-generated by Claude** (automated dead-code sweep). The commits are authored by Claude, not by a human contributor. Please review before merging.

## What

Removes two dead `TracingBuilder` methods from `dora-tracing`:

- `add_layer<L>(self, layer: L) -> Self`
- `with_layers<I, L>(self, layers: I) -> Self`

## Why

These generic escape-hatch methods are never called anywhere in the workspace (`git grep` finds only the definitions). Layers are always added through the specific builder methods that *are* used — `with_telemetry`, `with_span_capture`, etc. (e.g. `binaries/cli/src/command/coordinator.rs` uses `with_span_capture`). The two generic helpers are speculative API that never got a consumer.

## Validation

- `cargo fmt -p dora-tracing -- --check` ✓
- `cargo clippy -p dora-tracing -- -D warnings` ✓
- `cargo test -p dora-tracing` ✓
- `git grep -w add_layer` / `with_layers` confirm no callers remain.

_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

---
_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

PR #2193, opened 2026-06-14
refactor(ros2-bridge): remove unused Package::message_structs codegen helper

phil-opp:

## What

Removes the unused `Package::message_structs` method from `dora-ros2-bridge-msg-gen`:

```rust
pub fn message_structs(&self, gen_cxx_bridge: bool) -> (impl ToTokens, impl ToTokens)
```

## Why

This method builds the message struct definitions/impls token streams for a package, but **nothing in the workspace calls it**. The actual code generator in `msg-gen/src/lib.rs` produces message structs directly by iterating `package.messages` and calling `Message::struct_token_stream(...)` inline:

```rust
for message in &package.messages {
    let (def, imp) = message.struct_token_stream(&package.name, create_cxx_bridge);
    shared_type_defs.push(def);
    message_struct_impls.push(imp);
    // ...
}
```

So `message_structs` is dead, duplicated logic. (The sibling alias/service/action helpers on `Package` are all still used and untouched.)

## Verification

- `cargo fmt -p dora-ros2-bridge-msg-gen -- --check` — clean
- `cargo test -p dora-ros2-bridge-msg-gen` — 66 tests pass

Pure deletion; generated ROS 2 bindings are unaffected (no call site existed).

---
_Generated by [Claude Code](https://claude.ai/code/session_01SjmqFXAn8idhNcv6Yyavdo)_

### DIFF
--- a/libraries/extensions/ros2-bridge/msg-gen/src/types/package.rs
+++ b/libraries/extensions/ros2-bridge/msg-gen/src/types/package.rs
@@ -32,27 +32,6 @@ impl Package {
         self.messages.is_empty() && self.services.is_empty() && self.actions.is_empty()
     }
 
-    pub fn message_structs(&self, gen_cxx_bridge: bool) -> (impl ToTokens, impl ToTokens) {
-        if self.messages.is_empty() {
-            // empty msg
-            (quote! {}, quote! {})
-        } else {
-            let items = self
-                .messages
-                .iter()
-                .map(|v| v.struct_token_stream(&self.name, gen_cxx_bridge));
-            let defs = items.clone().map(|(def, _)| def);
-            let impls = items.clone().map(|(_, im)| im);
-            let def_tokens = quote! {
-                #(#defs)*
-            };
-            let impl_tokens = quote! {
-                #(#impls)*
-            };
-            (def_tokens, impl_tokens)
-        }
-    }
-
     fn message_aliases(&self, package_name: &Ident) -> impl ToTokens {
         if self.messages.is_empty() {
             quote! {


phil-opp:
Context for reviewers, given the discussion on #2185 (closed because `TracingBuilder::add_layer`/`with_layers` are a deliberate public extension point, not dead code): I want to be explicit about why I believe this one is different and is a genuine refactor leftover rather than an external-API escape hatch.

- **It's an orphaned duplicate, not an extension point.** `msg-gen/src/lib.rs` already emits message structs inline by iterating `package.messages` and calling `Message::struct_token_stream(...)` directly. `Package::message_structs` reimplements that same loop but is called from nowhere — it was bypassed by that inline generation.
- **Its siblings are private.** The directly-analogous per-package generators on `Package` — `message_aliases`, `service_aliases`, `services_block`, `action_aliases`, `actions_block` — are all `fn` (private). `message_structs` is the only one of that family left `pub`, which reads as a leftover rather than an intentional public surface. The `pub` helpers that genuinely are part of the codegen flow (`aliases_token_stream`, `token_stream`, `reuse_bindings_token_stream`, `dependencies_import_token_stream`) all have call sites in `lib.rs`; this one does not.
- **No "downstream needs it" story.** Unlike the tracing builder, there's no capability lost: external codegen consumers drive generation through the crate's top-level entry point, not per-package struct emitters, and the private siblings show this family was never meant as a public composition API.

That said, `dora-ros2-bridge-msg-gen` is publishable, so if you do consider `message_structs` an intentional public building block I'm missing, this is semver-relevant — happy to close this too. Deferring to your call.

---

PR #2196, opened 2026-06-14
docs(message): fix DataId try_from guidance

### LINKED ISSUE #2195: `DataId` doc recommends `DataId::try_from(s)` for untrusted input, but it panics

## Summary

In `libraries/message/src/id.rs`, the doc comment on `impl From<String> for DataId` tells callers to prefer `DataId::try_from(s)` when handling **untrusted input**:

```rust
/// # Panics
///
/// Panics if `id` contains invalid characters. Prefer `id.parse::<DataId>()`
/// or `DataId::try_from(s)` when handling untrusted input.
impl From<String> for DataId {
    fn from(id: String) -> Self {
        if let Err(e) = validate_data_id(&id) {
            panic!("invalid DataId '{id}': {e}");
        }
        Self(id)
    }
}
```
(`libraries/message/src/id.rs:182-193`)

This guidance is incorrect and dangerous. `DataId` has **no** explicit `TryFrom<String>` impl, so `DataId::try_from(s)` resolves to the standard-library blanket impl:

```rust
impl<T, U: Into<T>> TryFrom<U> for T { type Error = Infallible; /* delegates to From */ }
```

That blanket impl delegates straight to the panicking `From<String>`. So `DataId::try_from(malformed_string)` **panics** on invalid input rather than returning an `Err` — exactly the opposite of what the doc promises for untrusted input.

## This is the same trap the `NodeId` doc warns about

Just above, the `NodeId` doc in the **same file** explicitly and correctly warns against this:

```rust
/// Do NOT use `NodeId::try_from(s)`: `TryFrom<String>` is the
/// auto-derived blanket impl that delegates to this `From` impl, so it
/// panics exactly like `.into()`. Only `parse::<NodeId>()` / `from_str`
/// is fallible.
```
(`libraries/message/src/id.rs:97-100`)

The `DataId` doc contradicts the `NodeId` doc and steers callers toward a panic.

## Impact

- Any current/future code that follows the documented guidance — `DataId::try_from(untrusted_input)` — will **panic the process** on malformed input instead of getting a recoverable `Err`. For a `DataId` derived from network/config/user-supplied data, that is a remotely-triggerable crash (DoS) on what looks like a safe, fallible API.
- Currently latent: a repo-wide search found no caller using `DataId::try_from`, so no live bug today — but the misleading doc invites one.

## Reproduction

Minimal standalone program mirroring the pattern (a panicking `From<String>` + the std blanket `TryFrom`):

```rust
struct DataId(String);
fn validate(id: &str) -> Result<(), String> {
    if id.contains(';') { return Err(format!("invalid char in '{id}'")); }
    Ok(())
}
impl From<String> for DataId {
    fn from(id: String) -> Self {
        if let Err(e) = validate(&id) { panic!("invalid DataId '{id}': {e}"); }
        Self(id)
    }
}
fn main() {
    let untrusted = "bad;id".to_string();
    let r = std::panic::catch_unwind(|| { let _ = DataId::try_from(untrusted); });
    println!("try_from on invalid input panicked: {}", r.is_err()); // prints: true
}
```
Output: `try_from on invalid input panicked: true`

## Suggested fix

You can't add an explicit fallible `TryFrom<String> for DataId` while the `From<String>` impl exists (the std blanket impl conflicts), so the cleanest fix is to **correct the documentation** to match `NodeId`'s: recommend only `id.parse::<DataId>()` / `FromStr` (the genuinely fallible path) for untrusted input, and add the same "do NOT use `DataId::try_from(s)` — it panics" warning. `FromStr` is already implemented and validates correctly (`libraries/message/src/id.rs:167-174`).

Optionally, add a regression test asserting the safe path:
```rust
assert!("bad;id".parse::<DataId>().is_err());
```
(One already exists at `id.rs:310-313`; the doc just needs to point at it instead of `try_from`.)

sjh9714:
## Summary

Fixes #2195.

The `DataId` docs suggested `DataId::try_from(s)` for untrusted input, but that resolves to the standard blanket `TryFrom<String>` impl and delegates to the panicking `From<String>` implementation. This updates the guidance to recommend the fallible `parse::<DataId>()` / `FromStr` path, matching the existing `NodeId` warning.

## Changes

- Clarify the `From<String> for DataId` panic docs.
- Explicitly warn that `DataId::try_from(s)` is not the fallible path here.

## Testing

- `cargo test -p dora-message data_id_parse_rejects_invalid`
- `cargo test -p dora-message id::tests`
- `cargo clippy -p dora-message -- -D warnings`
- `git diff --check`

Note: Codex helped with issue scouting and local review; I reviewed the final diff and ran the listed checks locally.

### DIFF
--- a/libraries/message/src/id.rs
+++ b/libraries/message/src/id.rs
@@ -181,8 +181,15 @@ impl From<DataId> for String {
 
 /// # Panics
 ///
-/// Panics if `id` contains invalid characters. Prefer `id.parse::<DataId>()`
-/// or `DataId::try_from(s)` when handling untrusted input.
+/// Panics if `id` contains invalid characters (not in `[a-zA-Z0-9_./-]`).
+///
+/// **For untrusted input, use `id.parse::<DataId>()`** which calls
+/// `FromStr::from_str` and returns `Result<Self, InvalidId>`.
+///
+/// Do NOT use `DataId::try_from(s)`: `TryFrom<String>` is the
+/// auto-derived blanket impl that delegates to this `From` impl, so it
+/// panics exactly like `.into()`. Only `parse::<DataId>()` / `from_str`
+/// is fallible.
 impl From<String> for DataId {
     fn from(id: String) -> Self {
         if let Err(e) = validate_data_id(&id) {


phil-opp:
Thanks!

PR #2199, opened 2026-06-14
test(smoke): retry record-replay contract test on daemon TCP-read timeout

heyong4725:
## Problem

`contract_record_replay_reproduces_validated_pipeline` is an intermittent
red in the CircleCI `contract-tests` job. It fails with **"TCP read header
timed out"**, which originates in the daemon<->node transport
(`binaries/daemon/src/socket_stream_utils.rs`): a first-message read is
aborted after `dora_message::TCP_READ_TIMEOUT` (30s). On a loaded CI runner
the record- or replay-node spawn occasionally overshoots that on startup —
an **infrastructure stall, not a semantic break** in the recording.

It was caught live on the dora-hub PR stack: in a single CI workflow the
test **passed on #2177/#2178/#2179 and failed on #2176/#2180**, against
byte-identical base test code — the textbook signature of a non-deterministic
flake rather than a regression.

## Fix

Wrap the record+replay cycle in a bounded retry (3 attempts) that re-runs
**only** when the failure carries the TCP-timeout signature. Every other
failure mode fails immediately:

- non-zero exit *without* the timeout signature → fail now
- missing SUCCESS marker on a clean run → fail now (real semantic break)
- missing/empty `.drec` recording → fail now

So a genuine record/replay regression is never masked by a retry. All
original assertion messages are preserved verbatim. The change is **confined
to `tests/example-smoke.rs`** — no runtime/transport behavior changes.

## Test plan

- [x] `cargo test -p dora-examples --test example-smoke contract_record_replay_reproduces_validated_pipeline` passes locally (happy path, no retry triggered)
- [x] `cargo fmt --check` + `cargo clippy --tests -D warnings` clean
- [ ] CI `contract-tests` green

### DIFF
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -1097,6 +1097,68 @@ fn contract_record_replay_reproduces_validated_pipeline() {
     );
 
     let dora = dora_bin();
+
+    // The daemon<->node TCP transport aborts a first-message read after
+    // `TCP_READ_TIMEOUT` (30s, daemon `socket_stream_utils`). On a loaded CI
+    // runner the record- or replay-node spawn occasionally overshoots that on
+    // startup, surfacing as "TCP read header timed out" — an infrastructure
+    // stall, not a semantic break in the recording. Retry the whole
+    // record+replay cycle a bounded number of times on *that signature only*;
+    // a missing SUCCESS marker on an otherwise-clean run fails immediately, so
+    // a real record/replay regression is never masked by a retry.
+    const MAX_ATTEMPTS: u32 = 3;
+    for attempt in 1..=MAX_ATTEMPTS {
+        match record_replay_once(&dora) {
+            Ok(()) => return,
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** — _this is a fully automated review; please verify before relying on it._

No issues found. The change is confined to `tests/example-smoke.rs` and the retry logic checks out:

- A final-attempt `InfraTimeout` falls through to the catch-all `panic!` arm, so a *persistent* TCP-read timeout still fails the test rather than passing silently — the retry only masks transient stalls.
- `record_replay_once` re-writes the fixture and clears any stale `.drec` at the start of each attempt, and removes the `.drec` on every early-return failure path, so no state leaks across retries.
- Only the two TCP-timeout signatures are classified as retryable; a missing SUCCESS marker, a non-zero exit without the timeout signature, and a missing/empty `.drec` all fail immediately — so a genuine record/replay regression can't be hidden behind a retry. Original assertion messages are preserved.

Test-only change, no runtime/transport behavior affected.

---

PR #2206, opened 2026-06-14
fix(hub): enforce shipped-type namespace rule on untrusted index entries (P2.12)

### LINKED ISSUE #2205: Hub P2.12: load hub-shipped custom types from cache into TypeRegistry
**Phase 2 · P2.12** — part of the Dora Hub roadmap (umbrella #2097, spec `docs/plan-node-hub.md`).

Manifest `types:` are registered for **adjacent `path:` nodes** via contract injection (#2176), but the dedicated path that loads a **resolved `hub:` node's shipped types from the hub cache** into the `TypeRegistry` for the dataflow is not wired.

### Scope (§6.3)
- at resolution time, materialize a hub node's `types:` into the hub cache and add that directory to the `TypeRegistry` load path for the dataflow (today the registry loads `<working_dir>/types`; add the hub-cache search path)
- enforce the namespace rule: shipped types must live under the package's index namespace (`acme/lidar/v1/PointCloud`); `std/` is rejected

### Acceptance
- A hub node shipping a custom type resolves that type in a consumer's port check at `dora build`
- A hub node shipping a `std/...` type is rejected

Depends: P2.7 (landed). Parallel-safe. Spec: §6.3. Roadmap P2.12.

heyong4725:
## P2.12 — enforce shipped-type namespace rule at hub resolve

Closes #2205.

The hub resolver loads a resolved package's shipped `types:` into the dataflow's `TypeRegistry` so a consumer's port check can resolve them. The index entry is an **untrusted trust boundary** (spec §11), but resolution called `add_user_type` directly — skipping the namespace rule that `NodeManifest::validate` already enforces. A rewritten/malformed entry could register a `std/` override or a cross-namespace type a consumer then resolves.

### Change
- Factor the shipped-type checks out of `NodeManifest::validate` into a public `NodeManifest::shipped_type_issues` (namespace ownership, no `std/`, no traversal, materializable `arrow:`, resolvable fields).
- Run it in `resolve_hub_nodes` before loading any types; bail with the offending URNs.

The in-memory `add_user_type` registration already gives the compose-time resolution the spec's cache-dir load path (§6.3) was one way to achieve, so no separate cache materialization is needed.

### Tests
- [x] unit: `shipped_type_issues` rejects `std/` + cross-namespace, accepts own-namespace
- [x] hub-smoke: `dora build` rejects a package whose entry ships a `std/` type
- [x] `cargo fmt --check` + clippy clean; full `dora-core` manifest suite green

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -297,7 +297,10 @@ pub fn resolve_hub_nodes(
                 node.id,
                 type_issues
                     .iter()
-                    .map(|i| format!("{}: {}", i.field, i.message))
+                    // route through `ManifestIssue`'s Display, which strips
+                    // control chars — `i.message` can embed an unvalidated
+                    // field `type:` string from a hostile manifest
+                    .map(|i| i.to_string())
                     .collect::<Vec<_>>()
                     .join("\n  - ")
             );
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -284,6 +284,24 @@ pub fn resolve_hub_nodes(
         let manifest = &entry.manifest;
         let label = format!("{}@{version}", reference.key());
 
+        // The index entry is an untrusted input: validate its shipped types
+        // (namespace ownership, materializable bodies) before loading them, so
+        // a rewritten entry can't register a `std/` override or a
+        // cross-namespace type that a consumer's port check would resolve
+        // (spec §6.3, P2.12).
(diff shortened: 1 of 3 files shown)


heyong4725:
Review at head `ea01fe231`: no blocking findings.

I checked the new `NodeManifest::shipped_type_issues` helper and the hub resolver call site. The resolver now validates untrusted index-entry `types:` before registering them, and the new regression coverage exercises a rewritten `std/` shipped type.

Local validation run: `cargo test -p dora-core shipped_type_issues_gates_untrusted_entries` passed. GitHub checks were green except for pending CircleCI jobs at the time of review.

PR #2208, opened 2026-06-14
feat(hub): dora build --hub-override for local package substitution (P2.11, UC11)

### LINKED ISSUE #2204: Hub P2.11: dora build --hub-override for local package substitution (UC11)
**Phase 2 · P2.11 residual** — part of the Dora Hub roadmap (umbrella #2097, spec `docs/plan-node-hub.md`).

`dora hub fetch`, the `examples/hub-dataflow`, and `hub-smoke.rs` landed (#2180), but the **`--hub-override` flag is not implemented** (UC11 — node author's inner loop).

### Scope (UC11, §9, §10.1)
- `dora build dataflow.yml --hub-override <ns>/<name>=<local-path>` substitutes a local checkout for a hub package: manifest read from the checkout, contracts still validated, env built from local source — no index resolution for that node
- works for `dora run`/`dora validate` too (they build internally)

### Acceptance
- A consumer dataflow builds against a local node checkout via `--hub-override`, with contract validation still applied
- An override path with a missing/invalid `dora-node.yml` fails loudly

Depends: P2.7 (landed). Spec: UC11, §9. Roadmap P2.11.

heyong4725:
## P2.11 — `dora build --hub-override` (UC11 inner loop)

Closes #2204.

Lets a node author test a local checkout against a consumer dataflow **without publishing**:

```bash
dora build consumer.yml --hub-override dora-rs/dora-yolo=../dora-yolo
```

The manifest is read from `../dora-yolo`, contracts are still validated against the consumer wiring, and the node **builds + runs from local source** — no index resolution, no clone, no lockfile pin. Repeatable.

### Design
- The resolver desugars an overridden `hub:` node into a `path:` node and records its checkout dir.
- The **local** builder roots that node build at the checkout, so its recorded working dir also drives spawn.
- `--hub-override` is **local-only** (the checkout exists only on this machine): it forces a local build and errors if combined with a remote coordinator.
- An override that matches no hub node is **warned**, not silently ignored.

Deliberately avoids the `_unstable_deploy.working_dir` field (setting `deploy` flips the build to distributed dispatch) and needs **no daemon protocol change**.

### Scope
Wired for `dora build` (the UC11 command). `dora run`/`dora validate` pass an empty override map for now — a follow-up can thread the same flag.

### Tests
- [x] `hub_override_builds_from_local_checkout` — asserts the build lands in the checkout `target/`
- [x] `hub_override_unused_is_warned`
- [x] fmt + clippy clean; `cargo check --examples` green

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -7,12 +7,15 @@
 //! lockfile — sees an ordinary git node carrying the optional `subdir` and
 //! hub provenance marker on its [`GitSource`].
 
-use std::collections::BTreeMap;
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    path::PathBuf,
+};
 
 use dora_core::{
     build::validate_subdir,
     manifest::{
-        NodeManifest,
+        MANIFEST_FILENAME, NodeManifest,
         inject::{InjectionResult, apply_manifest_contracts},
     },
     types::TypeRegistry,
@@ -37,11 +40,15 @@ pub struct HubResolution {
     /// Per-node resolved git source (with subdir + hub provenance) — merged
     /// into the build's `git_sources` map and the lockfile.
     pub sources: BTreeMap<NodeId, GitSource>,
+    /// Per-node local working dir for `--hub-override` substitutions (UC11):
+    /// the node builds and runs from this checkout instead of a cloned commit.
+    /// Threaded into the local builder so build + spawn root there.
+    pub override_dirs: BTreeMap<NodeId, PathBuf>,
 }
 
 impl HubResolution {
     pub fn is_empty(&self) -> bool {
-        self.sources.is_empty()
+        self.sources.is_empty() && self.override_dirs.is_empty()
     }
 }
 
@@ -90,11 +97,13 @@ pub fn resolve_hub_nodes(
     registry: &mut TypeRegistry,
     offline: bool,
     pins: Option<&BTreeMap<NodeId, GitSource>>,
+    overrides: &BTreeMap<String, PathBuf>,
 ) -> eyre::Result<HubResolution> {
     let mut resolution = HubResolution::default();
     if !dataflow.nodes.iter().any(|n| n.hub.is_some()) {
         return Ok(resolution);
     }
+    let mut used_overrides = BTreeSet::new();
     resolution
         .notes
         .push("`hub:` is an unstable feature — its behavior may change in future releases".into());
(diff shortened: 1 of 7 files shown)


heyong4725:
Review findings at head `7d63a26c9`:

1. **High: `--hub-override --write-lockfile` can clobber a valid hub lockfile.**

The override path converts the hub node into a local path node and intentionally does not add a `GitSource` for that node, but `BuildLockfile::write_git_sources` still runs whenever `--write-lockfile` is set. A normal hub build can write a pinned lockfile, then a later local debug build with `--hub-override pkg=../checkout --write-lockfile` rewrites it without the hub pin. The next `--locked` build without the override no longer has the reference it needs, or worse, no longer protects the original hub dependency. Please reject lockfile writes while overrides are active, or make the lockfile logic explicitly preserve the original hub pins. Relevant code: `binaries/cli/src/command/build/hub.rs` around the override-to-local rewrite, and `binaries/cli/src/command/build/mod.rs` where lockfile writing is unconditional.

2. **Medium: an unmatched `--hub-override` can still build through a remote coordinator.**

The remote-coordinator guard only checks `hub_override_node_dirs.is_empty()` after resolution. If the override name is typoed or currently unused, the flag is merely warned about and the command can continue with the remote coordinator even though the PR documents `--hub-override` as local-only. Please reject any explicit remote coordinator whenever `--hub-override` is non-empty, independent of whether it matched a node.

Local validation run: `cargo test --test hub-smoke hub_override` passed.

heyong4725:
Both real — fixed in \`e744d6506\`.

**1 (High) — lockfile clobber.** Right: the override desugars its node to a local path node with no \`GitSource\` (only \`override_dirs\`), so \`--write-lockfile\` regenerated the lockfile from \`git_sources\` and dropped that hub node's pin. Now \`--hub-override\` + \`--write-lockfile\` is rejected with a clear error (the override is a transient local substitution; you shouldn't regenerate the committed pin from it). Test: \`hub_override_rejects_write_lockfile\`.

**2 (Medium) — guard only fired on matched overrides.** Right: it keyed on \`hub_override_node_dirs\` (matched), so a typo'd package fell through to a remote coordinator / distributed build. Re-keyed the guard on the *requested* overrides (parsed flag presence) and moved it **before** any index resolution or lockfile write, so the remote-coordinator / \`deploy:\` conflict is rejected regardless of whether the package matched a node. Test: \`hub_override_rejects_remote_coordinator_even_if_unmatched\` (uses a non-matching package). \`build_kind\` force-local likewise now keys on the requested overrides.

All 8 hub-smoke tests pass; fmt + clippy clean.

PR #2209, opened 2026-06-14
feat(coordinator): explicit daemon hub-capability flag + clear errors (P2.10)

### LINKED ISSUE #2203: Hub P2.10: multi-daemon hub-capability flag + too-old-daemon error
**Phase 2 · P2.10** — part of the Dora Hub roadmap (umbrella #2097, spec `docs/plan-node-hub.md`).

A resolved `hub:` node *is* a git node, so it already rides the existing multi-daemon `git_sources` flow (#2179). The remaining gap is the **mixed-version safety net**.

### Scope (§10.2)
- a daemon capability flag advertising `hub:`/`subdir` support
- the coordinator gives a clear "upgrade this daemon" error if a target daemon predates that support, instead of an opaque failure

(Belt-and-suspenders: desugaring is central, so an old daemon would never normally receive a hub node — but the flag makes the failure mode legible for mixed-version clusters.)

### Acceptance
- A simulated old daemon (no hub capability) assigned a hub node yields a clear per-node, per-daemon error naming the daemon to upgrade

Depends: P2.7 (landed). Spec: §10.2. Roadmap P2.10.

heyong4725:
## P2.10 — clear daemon version-mismatch error (capability flag analysis)

Closes #2203.

P2.10 asks for a daemon capability flag so the coordinator can give a clear *"upgrade this daemon"* error when a daemon predates hub `subdir`/`hub:` support.

### Why a separate capability flag would be dead code

Hub support landed in the **1.0 major line**. The existing registration gate (`DaemonRegisterRequest::check_version` → `versions_compatible`) already rejects a pre-hub **0.x** daemon against a hub-aware **1.x** coordinator: a `^0.x` requirement can't match `1.x` and a `^1.x` requirement can't match `0.x`, so they're incompatible and the daemon never registers. There is **no registrable 1.x daemon without hub support** (hub is in the very first 1.x). A version-derived `supports_hub` flag would therefore never reject anything — it can't fire. The spec itself flags this as belt-and-suspenders ("an old daemon would never normally receive such a node, since desugaring is central").

So the `dora_version` registration gate **is** the capability gate. What was missing is the *clear* part.

### Change

The daemon registration rejection message now:
- names both versions (daemon vs coordinator),
- tells the operator to upgrade the daemon (with the exact `cargo install` version),
- notes that an older daemon lacks newer wire features such as hub `subdir`/`hub:` node sources.

Tests cover the compatible path and the incompatible (actionable-error) path.

### Tests
- [x] `current_version_is_compatible`
- [x] `incompatible_daemon_gets_an_actionable_upgrade_error`
- [x] fmt + clippy clean

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -495,6 +495,22 @@ pub(crate) async fn build_dataflow(
 
     for (node_id, node) in &nodes {
         let daemon_id = resolve_daemon(daemon_connections, node.deploy.as_ref())?;
+        // A hub-sourced node carries `subdir` / `hub` provenance on its git
+        // source. A daemon too old to understand those fields can't build it,
+        // and the central desugar means it would otherwise fail with an opaque
+        // error on the daemon. Refuse up front with a clear, per-node message
+        // (P2.10) — the version gate alone can't catch this within an rc line
+        // where a pre-hub daemon and a hub-aware coordinator share a version.
+        if let Some(gs) = git_sources.get(node_id)
+            && (gs.subdir.is_some() || gs.hub.is_some())
+            && !daemon_connections.supports_hub_sources(&daemon_id)
+        {
+            eyre::bail!(
+                "node `{node_id}` is a hub package (or a monorepo `subdir` git node), but \
+                 daemon `{daemon_id}` is too old to build it — it does not support hub \
+                 `subdir`/`hub:` git sources. Upgrade that daemon to a dora build with hub support."
+            );
+        }
         nodes_by_daemon
             .entry(daemon_id.clone())
             .or_default()
--- a/binaries/coordinator/src/state.rs
+++ b/binaries/coordinator/src/state.rs
@@ -49,6 +49,12 @@ impl DaemonConnections {
             .find(|id| id.matches_machine_id(machine_id))
     }
 
+    /// Whether the given daemon advertised support for hub-sourced git nodes
+    /// (`subdir` / `hub` provenance, P2.10). Unknown daemon → `false`.
+    pub(crate) fn supports_hub_sources(&self, id: &DaemonId) -> bool {
+        self.daemons.get(id).is_some_and(|c| c.supports_hub_sources)
+    }
+
     pub(crate) fn drain(&mut self) -> impl Iterator<Item = (DaemonId, DaemonConnection)> {
         std::mem::take(&mut self.daemons).into_iter()
     }
@@ -90,6 +96,11 @@ pub(crate) struct DaemonConnection {
     pub(crate) labels: BTreeMap<String, String>,
     /// Latest fault tolerance stats from this daemon (updated on each heartbeat).
     pub(crate) ft_stats: Option<FaultToleranceSnapshot>,
+    /// Whether this daemon can build/spawn hub-sourced git nodes (`subdir` /
+    /// `hub` provenance, P2.10). Reported at registration; defaults `false`
+    /// for a daemon too old to advertise it, so the coordinator can refuse to
+    /// route a hub node to it with a clear error.
+    pub(crate) supports_hub_sources: bool,
 }
 
 impl DaemonConnection {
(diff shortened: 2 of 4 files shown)


phil-opp:

The error-message improvement itself looks correct, and both tests are sound (the incompatible-path test genuinely reaches the `Err` branch — it's not tautological). My concern is with the **rationale**, not the code:

The PR argues a `supports_hub` capability flag would be dead code because "there is no registrable 1.x daemon without hub support," so the existing `dora_version` gate *is* the capability gate. That doesn't hold for the version the project is currently on. The workspace is pinned at `1.0.0-rc1` (`Cargo.toml`), and that pin predates the hub commits — so a **pre-hub** daemon and a **hub-aware** coordinator both report `dora_version = 1.0.0-rc1`. `versions_compatible(1.0.0-rc1, 1.0.0-rc1)` returns `true` (`libraries/message/src/lib.rs`), so a pre-hub daemon registers successfully; the gate does **not** reject it. Within the rc window the version gate cannot distinguish a hub-capable daemon from a pre-hub one — which is exactly the gap #2203 asked the capability flag to cover.

The diff is still a fine improvement and makes nothing worse. But the description's "the gate IS the capability gate / a flag can't fire" claim (and the test comment about keeping a "pre-hub `0.x` daemon out of a hub-aware `1.x` cluster") describes a state the project isn't in yet. I'd suggest softening the rationale and not closing #2203 as "capability gating done" on this basis: this PR improves the error for *cross-major* mismatches, but same-version (rc) hub-capability gating remains genuinely unsolved. Tracking that gap separately would be reasonable too — the main thing is the inaccurate justification shouldn't stand unchallenged.

---

heyong4725:
Review finding at head `f0e8104d`:

**Medium: the new version-mismatch advice is wrong when the daemon is newer than the coordinator.**

`Register::parse_version` now always says to upgrade the daemon, but `versions_compatible` rejects both older and newer incompatible majors. The new test constructs a daemon version at `current.major + 1`, which is the newer-daemon case, and codifies the wrong action. In that case users need to upgrade the coordinator or downgrade the daemon, not upgrade the daemon further. Please branch the message by comparing daemon and coordinator versions, and cover both older-daemon and newer-daemon cases.

Local validation run: `cargo test -p dora-message register_version_tests` passed.

heyong4725:
You're right, and thank you — the rc-window point is the crux I got wrong. The workspace is `1.0.0-rc1`, that pin predates the hub commits, so a pre-hub daemon and a hub-aware coordinator report the **same** version and `versions_compatible` returns true. The version gate genuinely cannot distinguish them, and (as you note) a *version-derived* flag cant either. So the capability needs an **explicit** signal.

Implemented that in `9ae7e02f7`:

- `DaemonRegisterRequest.supports_hub_sources` — `#[serde(default)]` ⇒ **false** for a daemon too old to send it; current daemons advertise `true` via `new()`.
- The coordinator records it on the in-memory `DaemonConnection` at registration and, when distributing a build, **refuses to route a node whose git source carries `subdir`/`hub` provenance to a daemon that lacks the capability** — clear per-node "upgrade that daemon" error instead of an opaque downstream failure. Fail-closed (unknown daemon ⇒ incapable).
- No behavior change for same-version clusters: every current daemon reports `true`, so the routing check is a no-op there; it only fires for a genuinely pre-hub daemon in a mixed cluster.

Also corrected the inaccurate justification and the misleading test comment (the gap is same-version, not cross-major `0.x`). The cross-version error-message improvement is kept. Tests: serde default (current=true / legacy-json=false) + the coordinator capability accessor (fail-closed). Now genuinely closes #2203 rather than over-claiming on the version gate.

heyong4725:
Good catch — fixed in \`f17b2fc5e\`. \`versions_compatible\` rejects both older *and* newer daemons, so the remediation is now direction-aware:

- daemon **older** than the coordinator → "upgrade the daemon" (and the note that an older daemon also lacks newer wire features like hub \`subdir\`/\`hub:\` sources, which only applies in that direction);
- daemon **newer** than the coordinator → "upgrade the coordinator (or run an older daemon)".

The test now asserts both directions — the previous one used \`major + 1\` (a *newer* daemon) yet asserted the daemon-upgrade text, codifying the wrong advice as you noted. The explicit \`supports_hub_sources\` path is unchanged.

PR #2211, opened 2026-06-15
refactor(node-cxx): remove unreachable MergedEvents::merge

phil-opp:

## What

Removes `MergedEvents::merge` from `dora-node-api-cxx`, plus the now-dead private `next_id` field (and its two initializers) and the now-unused `MergeExternal` import.

## Why it's unreachable

`merge` is `pub`, but nothing can call it:

- It is **not** exported in the `#[cxx::bridge]` `extern "Rust"` block (lines 108–249), so it is not part of the C++ API surface. Its signature `impl Stream<Item = Box<dyn Any>>` could not cross the cxx FFI boundary in any case, so no C++ node can reach it.
- A whole-workspace `rg "\.merge\(" / "MergedEvents::merge"` finds no Rust caller — only the definition.

It's leftover plumbing from an external-stream-merging path that was never wired into the C++ node API. The live event path — `dora_events_into_combined` → `CombinedEvents::next` → `downcast_dora` — never goes through it; every item the stream produces is a `MergedEvent::Dora`.

Once `merge` is gone, `next_id` is only ever written (never read) and would trip `-D warnings`, so it is removed along with `MergeExternal`.

## Scope note (deliberately minimal)

The surrounding `MergedEvent` / `ExternalEvent` stream scaffolding is **left in place** — it is the event stream's type structure and is a plausible host for a future *bridged* merge API. Fully unwinding it (re-typing the stream to drop the `MergedEvent` wrapper) would be a larger refactor and a feature-level decision; this PR is a pure, low-risk deletion of the unreachable method only.

## Caveat for reviewers

`dora-node-api-cxx` does not set `publish = false`, so `merge` is technically part of the crate's published Rust surface. In practice the crate exists only to generate the C++ bridge and is not consumed as a Rust library, and this method is unreachable through the bridge — but flagging it so the API call is yours to make.

## Verification

- `cargo fmt -p dora-node-api-cxx -- --check` — clean
- `cargo clippy -p dora-node-api-cxx -- -D warnings` — clean
- `cargo test -p dora-node-api-cxx` — passes

---
_Generated by [Claude Code](https://claude.ai/code/session_01WRvTzEb58Xi7bm9DE5twsB)_

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -12,7 +12,7 @@ use dora_node_api::{
     self, Event, EventStream, Metadata as DoraMetadata,
     MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter, TryRecvError,
     arrow::array::{AsArray, UInt8Array},
-    merged::{MergeExternal, MergedEvent},
+    merged::MergedEvent,
 };
 use eyre::{Result as EyreResult, bail, eyre};
 use serde::Serialize;
@@ -365,7 +365,6 @@ fn dora_events_into_combined(events: Box<Events>) -> ffi::CombinedEvents {
     ffi::CombinedEvents {
         events: Box::new(MergedEvents {
             events: Some(Box::new(events)),
-            next_id: 1,
         }),
     }
 }
@@ -374,7 +373,6 @@ fn empty_combined_events() -> ffi::CombinedEvents {
     ffi::CombinedEvents {
         events: Box::new(MergedEvents {
             events: Some(Box::new(stream::empty())),
-            next_id: 1,
         }),
     }
 }
@@ -904,7 +902,6 @@ fn send_output_internal(
 
 pub struct MergedEvents {
     events: Option<Box<dyn Stream<Item = MergedEvent<ExternalEvent>> + Unpin>>,
-    next_id: u32,
 }
 
 fn new_metadata() -> Box<Metadata> {
@@ -980,22 +977,6 @@ impl MergedEvents {
     fn next(&mut self) -> MergedDoraEvent {
         let event = futures_lite::future::block_on(self.events.as_mut().unwrap().next());
         MergedDoraEvent(event)
-    }
-
-    pub fn merge(&mut self, events: impl Stream<Item = Box<dyn Any>> + Unpin + 'static) -> u32 {
-        let id = self.next_id;
-        self.next_id += 1;
-        let events = Box::pin(events.map(move |event| ExternalEvent { event, id }));
-
-        let inner = self.events.take().unwrap();
-        let merged: Box<dyn Stream<Item = _> + Unpin + 'static> =
-            Box::new(inner.merge_external(events).map(|event| match event {
-                MergedEvent::Dora(event) => MergedEvent::Dora(event),
-                MergedEvent::External(event) => MergedEvent::External(event.flatten()),
-            }));
-        self.events = Some(merged);
-
-        id
     }
 }
 


phil-opp:

No issues found. Confirmed `merge` is unreachable: the `extern "Rust"` bridge block exposes `MergedEvents` only as an opaque type and bridges `next` — not `merge`, whose `impl Stream<Item = Box<dyn Any>>` signature couldn't cross the cxx boundary anyway — and a whole-workspace search finds no Rust caller. Removing the now-dead `next_id` field and the unused `MergeExternal` import alongside it is correct. The caveat you raise (the crate isn't `publish = false`, so `merge` is technically part of its published Rust surface) is the only judgement call here, and a fair one to leave to the maintainers.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Please verify independently before acting.

I found an issue: I don't think this is safe to merge as-is. `MergedEvents::merge` is **not** dead code — it's the call target of the ROS2-bridge code generator.

`dora-ros2-bridge-msg-gen` emits `events.events.merge(Box::pin(stream))` against `&mut crate::ffi::CombinedEvents` (whose `events` field is the `Box<MergedEvents>` this PR edits) at six sites:

- `libraries/extensions/ros2-bridge/msg-gen/src/types/service.rs:152` and `:374`
- `libraries/extensions/ros2-bridge/msg-gen/src/types/message.rs:435`
- `libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs:412`, `:507`, `:527`, `:904`

That generated code is written to `messages.rs` by `apis/c++/node/build.rs` (`ros2::generate()`) and `include!`-ed back into this same crate at `apis/c++/node/src/lib.rs:255`, so the deleted method is exactly what those call sites bind to (`crate::ffi::CombinedEvents.events` is the `Box<MergedEvents>` defined at `apis/c++/node/src/lib.rs:101`/`:905`). Removing `merge` — and `next_id`/`MergeExternal`, which exist only to support it — makes `dora-node-api-cxx` fail to compile with `error[E0599]: no method named 'merge'` at all six sites when built with the `ros2-bridge` feature against a real ROS2 distro.

Why the whole-workspace grep (and PR/nightly CI) miss it: the `.merge()` calls only materialize when `AMENT_PREFIX_PATH` points at actual ROS2 message packages. With no ROS distro installed — the case for PR and nightly CI, which only do `cargo check` with no distro — `msg-gen` emits no per-message client/subscription/action functions, so the crate compiles without ever referencing `merge`. I confirmed empirically that the deletion builds cleanly in a no-distro environment, which is precisely why this slips past automated checks.

The "it can't cross the cxx FFI boundary anyway" reasoning doesn't apply here: `merge` is a plain Rust method called by Rust-side generated glue; it never crosses FFI itself. Recommend keeping it — it is live for ROS2 builds.

---

phil-opp:
🤖 _This is a fully automated review by Claude — no human has vetted this comment before posting. Please verify before acting._

I found the following issue.

**`MergedEvents::merge` is not dead code — removing it breaks the C++ ROS2 bridge build.**

The PR's premise is that a whole-workspace search finds no Rust caller of `merge`. There are seven, all in the ros2-bridge code generator, which emits Rust that is compiled *into this very crate* under the `ros2-bridge` feature:

- `libraries/extensions/ros2-bridge/msg-gen/src/types/service.rs:152` (create_client), `:374` (create_server)
- `libraries/extensions/ros2-bridge/msg-gen/src/types/message.rs:435` (create_subscription)
- `libraries/extensions/ros2-bridge/msg-gen/src/types/action.rs:412`, `:507`, `:527`, `:904`

Each generated fn takes `events: &mut crate::ffi::CombinedEvents` (e.g. `message.rs:428`) and calls `events.events.merge(Box::pin(stream))`. `CombinedEvents.events` is `Box<MergedEvents>` (`apis/c++/node/src/lib.rs:100-101`), so this resolves to exactly the `MergedEvents::merge` (`lib.rs:985`) being deleted — and `next_id`, which the PR also removes, is read/written only by `merge` and by the `MergedEvents { next_id: 1 }` initializers that construct these values. The codegen is wired in via `apis/c++/node/build.rs:167` (`dora_ros2_bridge_msg_gen::generate(..., /*gen_cxx_bridge=*/ true)`), gated on the `ros2-bridge` feature.

The reason a text grep "missed" these is that they live inside `quote!{}` token streams as `events.events.merge(...)`, and the default crate build (no `ros2-bridge` feature) never compiles them. This also won't be caught by CI: `test-c-cpp-libraries.yml` builds the cxx crate without `--features ros2-bridge`, and the nightly `ros2-bridge` job only runs `cargo check -p dora-ros2-bridge --no-default-features` plus msg-gen token tests — nothing compiles the generated callers. The break would surface only at release time or for users building the C++ ROS2 bridge (`cargo build --package dora-node-api-cxx --features ros2-bridge`).

Suggest verifying with `cargo build --package dora-node-api-cxx --features ros2-bridge` — I expect it to fail with E0599 (no method `merge`). The method (and `next_id`) should be kept.

---

PR #2214, opened 2026-06-15
fix(scheduler): log correlation drop for ZenohInput

### LINKED ISSUE #2212: Scheduler: correlation-breaking drop is silent for Zenoh (distributed) inputs — `log_correlation_drop` ignores `EventItem::ZenohInput`
_This issue was filed by an automated, scheduled Claude code-review check._

## Summary

In the event-stream scheduler, when a full input queue is forced to evict a **correlated** message (a service response / action result carrying `request_id`, `goal_id`, or `goal_status`), it is supposed to emit a *loud* error so operators can trace the broken request/response contract. That loud log is **silently skipped when the dropped message is a `ZenohInput`** — i.e. exactly for **distributed / cross-machine dataflows**, which are the case where Zenoh is the transport and where correlation breakage is hardest to debug.

`apis/rust/node/src/event_stream/scheduler.rs`

## Root cause

`is_correlated` correctly treats both `NodeEvent::Input` **and** `ZenohInput` as correlation-bearing (scheduler.rs:22-34):

```rust
fn is_correlated(event: &EventItem) -> bool {
    let params = match event {
        EventItem::NodeEvent { event: NodeEvent::Input { metadata, .. }, .. } => &metadata.parameters,
        EventItem::ZenohInput { metadata, .. } => &metadata.parameters,   // <-- Zenoh counts as correlated
        _ => return false,
    };
    params.contains_key(REQUEST_ID) || params.contains_key(GOAL_ID) || params.contains_key(GOAL_STATUS)
}
```

So a `ZenohInput` can legitimately be the front-of-queue correlated event that gets evicted via `Eviction::DropFrontLoud`.

But `log_correlation_drop` — the function whose entire purpose is to be loud about that eviction — only destructures `NodeEvent::Input` and `return`s early for anything else, including `ZenohInput` (scheduler.rs:66-88):

```rust
fn log_correlation_drop(event_id: &DataId, dropped: &EventItem) {
    let EventItem::NodeEvent {
        event: NodeEvent::Input { metadata, .. },
        ..
    } = dropped
    else {
        return;   // <-- ZenohInput falls through here: NO error is logged
    };
    // ... tracing::error!(...) only reached for NodeEvent::Input
}
```

The caller treats the eviction as "loud", but the log never fires (scheduler.rs:289-293):

```rust
Eviction::DropFrontLoud => {
    if let Some(front) = queue.pop_front() {
        log_correlation_drop(event_id, &front);   // no-op for ZenohInput
    }
}
```

## Impact

- A remote service/action correlation can be dropped under queue pressure **with zero diagnostic output**. The client waits forever for a response/result that was silently discarded — the exact failure mode the loud log (and dora-rs/adora#145) was meant to surface.
- The miss is biased toward distributed deployments (Zenoh transport), where this class of bug is most painful to diagnose.
- The drop counter (`self.dropped`) still increments, but the per-correlation context (`request_id` / `goal_id` / `goal_status`) — the part that lets an operator trace *which* request was lost — is never emitted.

## Suggested fix

Extend `log_correlation_drop` to also pull `&metadata.parameters` from `EventItem::ZenohInput` (mirroring `is_correlated`), so the same `tracing::error!` with the correlation keys fires regardless of transport. A small unit test asserting the loud path is exercised for a correlated `ZenohInput` front would lock this in (the existing `drop_oldest_drops_front_loudly_when_both_queue_and_incoming_are_correlated` test only covers `NodeEvent::Input`).

## Notes

Found by reviewing `apis/rust/node/src/event_stream/scheduler.rs`. Severity is "lost observability on an already-bad path" rather than data corruption, but it defeats a guardrail that was added deliberately. Please verify before acting — flagged by an automated check, not a human reviewer.

phil-opp:
Closes #2212

## Summary

`log_correlation_drop` in `apis/rust/node/src/event_stream/scheduler.rs` silently returned early for `EventItem::ZenohInput`, even though `is_correlated` correctly treats `ZenohInput` as correlation-bearing. This meant that when a correlated Zenoh message (service response / action result) was evicted via `DropFrontLoud`, the error log — which exists to surface broken request/response contracts — never fired.

This was especially harmful for distributed dataflows where Zenoh is the transport and correlation breakage is hardest to diagnose.

## Root Cause

The original `log_correlation_drop` used a `let-else` guard that only destructured `NodeEvent::Input`:

```rust
fn log_correlation_drop(event_id: &DataId, dropped: &EventItem) {
    let EventItem::NodeEvent {
        event: NodeEvent::Input { metadata, .. },
        ..
    } = dropped
    else {
        return;  // ZenohInput falls through here — no error logged
    };
    ...
}
```

## Fix

Replace the guard with a `match` that extracts `&metadata.parameters` from both variants, mirroring the pattern already in `is_correlated`:

```rust
fn log_correlation_drop(event_id: &DataId, dropped: &EventItem) {
    let params = match dropped {
        EventItem::NodeEvent {
            event: NodeEvent::Input { metadata, .. },
            ..
        } => &metadata.parameters,
        EventItem::ZenohInput { metadata, .. } => &metadata.parameters,
        _ => return,
    };
    ...
}
```

## Tests Added

- `zenoh_drop_oldest_drops_front_loudly_when_both_queue_and_incoming_are_correlated` — when a full queue of correlated `ZenohInput` items receives another correlated one, the oldest is evicted and the two newest survive
- `zenoh_drop_oldest_preserves_correlated_when_non_correlated_present` — a correlated `ZenohInput` is not evicted when non-correlated items are available to drop instead

## Verification

- `cargo fmt -p dora-node-api -- --check` ✓
- `cargo clippy -p dora-node-api -- -D warnings` ✓
- `cargo test -p dora-node-api` — 103 tests pass ✓

---
_Generated by [Claude Code](https://claude.ai/code/session_01SjmbDSV9GfWdiUovySqHTB)_

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -64,14 +64,14 @@ fn select_eviction(queue: &VecDeque<EventItem>, incoming: &EventItem) -> Evictio
 /// everything in the queue is also correlated. Identifies the correlation
 /// keys so operators can trace the affected request/goal.
 fn log_correlation_drop(event_id: &DataId, dropped: &EventItem) {
-    let EventItem::NodeEvent {
-        event: NodeEvent::Input { metadata, .. },
-        ..
-    } = dropped
-    else {
-        return;
+    let params = match dropped {
+        EventItem::NodeEvent {
+            event: NodeEvent::Input { metadata, .. },
+            ..
+        } => &metadata.parameters,
+        EventItem::ZenohInput { metadata, .. } => &metadata.parameters,
+        _ => return,
     };
-    let params = &metadata.parameters;
     let request_id = get_string_param(params, REQUEST_ID);
     let goal_id = get_string_param(params, GOAL_ID);
     let goal_status = get_string_param(params, GOAL_STATUS);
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Verified that `is_correlated` (scheduler.rs:28) treats `ZenohInput` as correlation-bearing, while the old `let-else` in `log_correlation_drop` (scheduler.rs:67) only matched `NodeEvent::Input` and silently returned for `ZenohInput` — so a dropped correlated `ZenohInput` skipped the loud error log. The `match` rewrite extracting `&metadata.parameters` from both arms is correct and mirrors `is_correlated`. The two new tests route `ZenohInput` through the same `id`-keyed queue and exercise the `DropFrontLoud` eviction path as claimed.

---

phil-opp:

The production change is correct: the new `match` extracts `&metadata.parameters` from both `NodeEvent::Input` and `ZenohInput`, mirroring `is_correlated`, so a correlated `ZenohInput` evicted via `DropFrontLoud` now logs instead of silently returning. That closes the gap in #2212.

I found one issue, though — the added tests don't actually guard the fix. `log_correlation_drop` only emits a `tracing::error!`; it has no effect on queue state. Both new tests assert solely on queue contents (`queue.len()` / which `request_id`s survive), and those are determined by `select_eviction`, which already handled `ZenohInput` correctly on `main` (via the unchanged `is_correlated`). As a result:

- `zenoh_drop_oldest_drops_front_loudly_when_both_queue_and_incoming_are_correlated` passes identically against the *unpatched* code — the old `log_correlation_drop` silently returns and the queue ends in the same state.
- `zenoh_drop_oldest_preserves_correlated_when_non_correlated_present` takes the `RemoveAt` eviction path and never reaches `log_correlation_drop` at all.

So neither test would catch a revert of the actual change. To genuinely exercise it, the test needs to capture the tracing output (e.g. `tracing-test` / a capturing subscriber) and assert the error was emitted for the correlated `ZenohInput`. The fix itself is fine; the regression coverage is the concern.

---

Issue #2201, opened 2026-06-14
Hub P2.3: bootstrap node-index catalog + CI + auto-merge bot (in dora-rs/dora-hub)


heyong4725:
**Phase 2 · P2.3** — part of the Dora Hub roadmap (umbrella #2097, spec `docs/plan-node-hub.md`).

The consumption side (`dora-hub-client`, resolver, transport — #2178) is landed, but the **index catalog itself does not exist yet**. This work lives in the **`dora-rs/dora-hub` repo**, not this one.

### Scope (§7.1, §7.4, §7.5)
- `node-index/` directory layout in `dora-rs/dora-hub`
- path-scoped index CI: append-only enforcement, schema + pin validation, namespace identity (GitHub API) + confusable-name screening — scoped to `node-index/**`, not gating the human-reviewed `node-hub/` source path
- auto-merge bot for rule-conformant diffs (§7.5)
- namespace policy page + PR templates

### Acceptance
- A fixture/real `node-index/` tree resolves via `dora-hub-client`
- CI rejects a non-append-only diff and a namespace-identity mismatch
- A conformant version-add PR auto-merges

Larger than one PR (2–3 PRs for policy/bot). Cross-repo. Spec: §7. Roadmap P2.3.

heyong4725:
**Scope note — this is cross-repo work.**

P2.3 bootstraps the `node-index/` catalog, its append-only CI, the auto-merge bot, and the namespace policy. Per the spec (§7.1) the catalog lives **in the `dora-rs/dora-hub` repo**, not in `dora-rs/dora`. None of those artifacts (index tree, GitHub Actions workflow, bot, PR templates) can be created from this repo — they need PRs against `dora-rs/dora-hub`.

What this repo already provides, so the format is pinned before that work starts:
- the **consumer contract** is implemented and merged: `dora-hub-client` parses the index tree (`<ns>/<name>/<version>.yml` = `manifest:` + `source:`), resolves semver, handles yank + namespace binding, and enforces append-only-relevant invariants client-side (#2178).
- a **reference index fixture** exists at `tests/hub-smoke.rs`'s `build_fixture()` — a minimal valid `node-index` tree the CI in dora-hub can mirror for its own tests.

Leaving this issue **open** as the tracking item for the dora-hub-side work; it is not actionable as a PR in this repository. Suggest transferring it to `dora-rs/dora-hub` (or opening the sibling issue there) when that phase starts. Umbrella: #2097.

[...]

PR #2215, opened 2026-06-15
feat(hub): dora hub publish — validate + add a pinned index entry (P3.1)

heyong4725:
## P3.1 — `dora hub publish`

First Phase 3 (publishing) deliverable. Publishing is just *adding an index entry that points at your source* — no upload, since the bytes already live in git (spec §9.1).

`dora hub publish [PATH]`:
1. reads + validates the node manifest (same gate as `dora validate --node-manifest`);
2. reads the version from `[package].version` (Cargo) / `[project].version` (pyproject), or `--version`;
3. resolves the source pin: repo (`origin` remote or `--repo`), the `--rev`/`HEAD` commit, and the directory's subdir within the repo;
4. constructs the index entry `<namespace>/<name>/<version>.yml`.

- `--dry-run` validates + prints the entry (no write).
- Writes into a **local** (`path =`) index — the enterprise/private and fixture form (UC8/UC11) — and refuses to overwrite an existing version (**append-only**, §7.5).
- For a git-backed index, prints the entry + where to add it. Automated PR-opening against the official `node-index` lands with the index bootstrap (P2.3, which lives in the `dora-rs/dora-hub` repo, tracked by #2201).

### Tests
- [x] `hub_publish_dry_run_previews_entry` — validates + previews, writes nothing
- [x] `hub_publish_then_build_resolves` — **full publish→build loop**: `dora hub publish` writes a pinned entry, `dora build` resolves and builds it; plus the append-only re-publish rejection
- [x] fmt + clippy clean, `cargo check --examples` green

### Scope note
P3.2 (`yank`/`outdated`/`update`) follows next. P3.3 (62-node migration) and P3.4 (index CI) are `dora-rs/dora-hub` repo work. Part of umbrella #2097.

### DIFF
--- a/binaries/cli/src/command/hub/mod.rs
+++ b/binaries/cli/src/command/hub/mod.rs
@@ -11,6 +11,7 @@ mod info;
 mod init;
 mod install;
 mod list;
+mod publish;
 mod search;
 
 /// Package, discover, and use dora nodes (unstable)
@@ -22,6 +23,7 @@ pub enum Hub {
     List(list::List),
     Fetch(fetch::Fetch),
     Install(install::Install),
+    Publish(publish::Publish),
 }
 
 impl Executable for Hub {
@@ -33,6 +35,7 @@ impl Executable for Hub {
             Hub::List(cmd) => cmd.execute(),
             Hub::Fetch(cmd) => cmd.execute(),
             Hub::Install(cmd) => cmd.execute(),
+            Hub::Publish(cmd) => cmd.execute(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


phil-opp:

No blocking issues found. Reviewed `publish.rs`, the docs, and the smoke tests:

- Append-only is correctly enforced via `OpenOptions::create_new(true)` (O_EXCL), so an existing version is rejected even under concurrent publish.
- The entry write path `{namespace}/{name}/{version}.yml` is path-safe: `validate_strict` runs first and the namespace/name charsets forbid `/` and `..`, while `version` is a parsed semver.
- The source repo URL is validated with `validate_git_url_untrusted` and the commit must be a full 40-char hash.
- `--index` correctly refuses to seed a namespace into an index it isn't bound to.

One minor footgun (non-blocking, user-error territory rather than a defect): `--repo` overrides the embedded URL while the commit and subdir are still derived from the local directory's git, so a mismatched `--repo` could pin a commit/subdir that doesn't exist in that repo. A sanity note in the `--repo` help text would prevent surprises.

---

PR #2216, opened 2026-06-15
feat(hub): dora hub yank — yank/restore a published version (P3.2, UC10)

heyong4725:
## P3.2 — `dora hub yank` (UC10)

`dora hub yank <pkg>@<version> [--reason R] [--undo]` flips the `yanked` flag — the one mutation an existing index version file is allowed (spec §7.5):

- new resolutions **skip** a yanked version;
- an existing `--locked` pin keeps working but `dora build` **warns**;
- `--undo` restores it.

(The resolver already implements both behaviors — this command just sets the flag.)

For a **local** (`path =`) index the flag is flipped in place (read entry → set `yanked`/`yank_reason` → atomic rename). For a **git-backed** index it prints the flag-flip change to open as a PR (the official index's flag-flip lands as a bot-merged PR, §7.5).

### Tests
- [x] `hub_yank_skips_version_then_undo_restores` — yank → a fresh resolve fails (only matching version yanked) → `--undo` → builds again
- [x] `hub_yank_git_index_prints_pr_instructions`
- [x] fmt + clippy clean

### Scope
First of the P3.2 trio. `outdated` (lockfile pins vs index) and `update` (re-resolve + rewrite lockfile) follow in a separate PR. Part of umbrella #2097.

### DIFF
--- a/binaries/cli/src/command/hub/mod.rs
+++ b/binaries/cli/src/command/hub/mod.rs
@@ -13,6 +13,7 @@ mod install;
 mod list;
 mod publish;
 mod search;
+mod yank;
 
 /// Package, discover, and use dora nodes (unstable)
 #[derive(Debug, clap::Subcommand)]
@@ -24,6 +25,7 @@ pub enum Hub {
     Fetch(fetch::Fetch),
     Install(install::Install),
     Publish(publish::Publish),
+    Yank(yank::Yank),
 }
 
 impl Executable for Hub {
@@ -36,6 +38,7 @@ impl Executable for Hub {
             Hub::Fetch(cmd) => cmd.execute(),
             Hub::Install(cmd) => cmd.execute(),
             Hub::Publish(cmd) => cmd.execute(),
+            Hub::Yank(cmd) => cmd.execute(),
         }
     }
 }
--- a/binaries/cli/src/command/hub/yank.rs
+++ b/binaries/cli/src/command/hub/yank.rs
@@ -60,22 +60,34 @@ impl Executable for Yank {
         let action = if self.undo { "restore" } else { "yank" };
 
         if !index.is_local() {
-            println!(
-                "the `{}` index is git-backed ({}). {action} is a flag-flip PR: in `{rel_path}` set \
-                 `yanked: {}`{} and open a PR against the index repo.",
-                index.alias,
-                index.git.as_deref().unwrap_or("?"),
-                !self.undo,
+            // Mirror the local path: yanking records the reason, restoring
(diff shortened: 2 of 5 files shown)


phil-opp:

No issues found. Reviewed `yank.rs`, the `is_valid_key_part` visibility change, and the smoke tests. The key-part guard correctly runs *before* any write path is built — closing the `..` traversal that `PackageRef::parse` alone does not reject — with a dedicated regression test. `write_atomic` writes the temp file in the same directory (`with_extension("yml.tmp")` on `<version>.yml`) and renames, so the flag flip is atomic. The read→flip→reserialize round-trip is safe because `IndexEntry` is `deny_unknown_fields`, and the git-backed path correctly prints PR instructions instead of mutating in place.

---

phil-opp:
🤖 _This is a fully automated review by Claude — no human has vetted this comment before posting. Please verify before acting._

The core of this looks correct — the path-traversal guard runs before any path is built, the version comes from `semver::Version::parse` (can't contain `/` or `..`), `yank_reason` is written through `serde_yaml` (no YAML injection), and the temp file is on the same directory so the rename is atomic. Two issues worth raising:

**1. The yank write path skips the `confine()` symlink check that every read path enforces.**
In `binaries/cli/src/command/hub/yank.rs`, the local write resolves `catalog.join(rel_path)` and writes through `write_atomic`, guarded only by `is_valid_key_part` on the namespace/name. But every read in `index.rs` (`versions`, `entry`, `package_meta`, `all_packages`) additionally calls `self.confine(path)` to reject entries that, after symlink resolution, escape `canonical_root`. The write path does not. A local index whose `<version>.yml` is a symlink pointing outside the catalog (e.g. `0.1.0.yml -> /etc/something`) would be followed and overwritten here — the key-part guard blocks `..` in the namespace/name but not a symlinked entry file. The new doc comment on `is_valid_key_part` ("a write must never escape the catalog root") slightly overstates what's actually enforced. Consider routing the write through `IndexCatalog` (which already holds `canonical_root`) or applying the same `confine` check to the entry path. Local indexes are operator-controlled so practical risk is low, but this is exactly the asymmetry the read-side `confine()` exists to close.

**2. The temp file isn't cleaned up if the rename fails.**
In `write_atomic`, on a `std::fs::rename` failure the `*.yml.tmp` file is left behind, and the fixed `.tmp` suffix (no random/PID component) isn't safe against two concurrent yanks sharing the name. Atomicity of the visible entry is preserved, but a best-effort `remove_file(&tmp)` on the error branch (and/or a unique suffix) would clean this up.

Neither blocks merge; (1) is the more substantive one.

---

PR #2217, opened 2026-06-15
feat(hub): dora hub outdated — locked hub pins behind the index (P3.2)

heyong4725:
## P3.2 — `dora hub outdated`

`dora hub outdated <dataflow>` reads the dataflow's lockfile and, for each pinned hub package, compares the pin against the **latest non-yanked version** in the index (`IndexCatalog::resolve` with a `*` requirement, so a major/minor bump still surfaces — not just in-range patches). Reports each outdated pin as `pinned -> latest`, or "all up to date".

- `--offline` uses only the cached index.
- Index/lockfile strings are sanitized before printing (an untrusted entry can't inject terminal escapes).
- A malformed pin or unreadable index degrades to a per-node note, not a whole-command failure.

### Test
- [x] `hub_outdated_reports_newer_version` — build pins `0.1.0` (reports up to date) → add `0.2.0` to the index → reports `0.1.0 -> 0.2.0 (newer available)`
- [x] fmt + clippy clean

### Scope
Completes the read-side of P3.2. `dora hub update` (re-resolve + rewrite the lockfile without building) is the remaining piece — it overlaps `dora build --write-lockfile` and wants a "resolve-and-lock, no build" mode, so it's a focused follow-up. With `publish` (#2215) + `yank` (#2216) + this, the hub publishing CLI is nearly complete. Part of umbrella #2097.

### DIFF
--- a/binaries/cli/src/command/hub/mod.rs
+++ b/binaries/cli/src/command/hub/mod.rs
@@ -11,6 +11,7 @@ mod info;
 mod init;
 mod install;
 mod list;
+mod outdated;
 mod publish;
 mod search;
 mod yank;
@@ -26,6 +27,7 @@ pub enum Hub {
     Install(install::Install),
     Publish(publish::Publish),
     Yank(yank::Yank),
+    Outdated(outdated::Outdated),
 }
 
 impl Executable for Hub {
@@ -39,6 +41,7 @@ impl Executable for Hub {
             Hub::Install(cmd) => cmd.execute(),
             Hub::Publish(cmd) => cmd.execute(),
             Hub::Yank(cmd) => cmd.execute(),
+            Hub::Outdated(cmd) => cmd.execute(),
         }
     }
 }
--- a/binaries/cli/src/command/hub/outdated.rs
+++ b/binaries/cli/src/command/hub/outdated.rs
@@ -34,6 +34,7 @@ impl Executable for Outdated {
 
         let mut checked = 0usize;
         let mut outdated = 0usize;
+        let mut errored = 0usize;
         for (node_id, source) in &lockfile.git_sources {
             let Some(hub) = &source.hub else {
                 continue;
@@ -44,6 +45,7 @@ impl Executable for Outdated {
             // before echoing, and treat a malformed pin as "can't check".
             let Some((namespace, name)) = hub.name.split_once('/') else {
                 println!("{node_id}: malformed hub name `{}`", sanitize(&hub.name));
+                errored += 1;
                 continue;
             };
             let pinned = match hub.version.parse::<dora_hub_client::semver::Version>() {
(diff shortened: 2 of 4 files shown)


phil-opp:

No issues found. Reviewed `binaries/cli/src/command/hub/outdated.rs`, the docs, and the smoke test. The resolve-with-`*` / compare-`>`-against-pin logic is correct, yanked versions are excluded by `IndexCatalog::resolve`, and all lockfile-sourced strings are passed through `sanitize()` before printing (`node_id` is additionally charset-validated by its deserializer, so printing it raw is safe). Malformed pins degrade to a per-node note rather than failing the whole command, as intended.

---

phil-opp:

One gap: the threat model this PR sets up (sanitize untrusted lockfile/index strings before printing) isn't actually closed on the error path. In the `Err(err)` arm of `catalog.resolve(...)`, `{err:#}` is printed **un-sanitized**, and it re-embeds the untrusted `name`:

```rust
Err(err) => println!(
    "{node_id}: {}/{} {pinned} — {err:#}",
    sanitize(namespace),
    sanitize(name)          // sanitized here…
),                          // …but {err:#} contains the raw name
```

`IndexCatalog::resolve` builds its errors via `reference.key()` (`format!("{namespace}/{name}")`, `libraries/hub-client/src/index.rs` / `reference.rs:59`), and the "not found" branch additionally appends `suggest()` (an index-supplied package name). `hub.name` comes from the lockfile and isn't charset-validated on read (`HubProvenance.name` is a plain `String`; `BuildLockfile::read_from` is a bare `serde_yaml::from_str`), so a crafted pin can inject terminal escapes here. This path fires on the **common** case of a renamed/removed pin, not a rare corner — so it's the easiest path to hit, and it defeats the exact protection the PR is adding.

Suggest sanitizing the rendered error too, e.g. `sanitize(&format!("{err:#}"))`, in both the catalog-open and resolve `Err` arms.

The rest looks right: `*`-resolve vs `>`-pin correctly surfaces major/minor bumps (not just in-range patches), `resolve` excludes yanked versions, per-node degradation is sound, and `node_id` is safe to print raw since its deserializer enforces `[a-zA-Z0-9_.-]`.

---
_Generated by Claude Code_

---

phil-opp:

Re-reviewed at the latest commit `c6ca88e` (adds the `errored` counter so failed per-node checks no longer get summarized as "all up to date" and the command now exits non-zero). That change looks correct, and I found no new issues in it.

On the earlier escape-injection concern (printing `{err:#}` without `sanitize()` in the resolve/catalog `Err` arms): re-examining the code, I don't think it's actually reachable. Before `IndexCatalog::resolve` constructs any error that embeds the raw package key, `package_dir()` runs `is_valid_key_part()`, whose charset is `[A-Za-z0-9._-]` — it rejects every control/escape character and bails with an already-sanitized message (`libraries/hub-client/src/index.rs`). So a name carrying terminal escapes never reaches the `{err:#}` formatting path; the strings it can carry there are already control-char-free. Wrapping it in `sanitize(&format!("{err:#}"))` would be harmless defense-in-depth, but it isn't closing a live vulnerability.

No issues found in the current diff.

---

PR #2225, opened 2026-06-16
security: remove committed .adora-token and add to .gitignore

### LINKED ISSUE #2194: Security: coordinator auth token committed to repo (`.adora-token`), tracked on main and not gitignored
## Summary

A file named **`.adora-token`** is tracked on `main` and contains a 64-hex-character value:

```
46832365ba2afc9989120bcab1b07a6a5e8afb629c27107ab116a3179b2244a0
```

This is **exactly the format of the coordinator's shared bearer-auth token** (a cryptographically random 32-byte value rendered as 64 hex chars), i.e. a leaked secret committed to a public repository. It is the fork-era (`adora`) filename of the same auth mechanism that is now `.dora-token`.

Verified present on `main` at the current tip (`419ce80`) via the GitHub API:
- Path: `.adora-token`
- Blob SHA: `56bc85f433691e33a02b9cec6693374d16b3006c`

## Why this is a real token (not a checksum/marker)

`libraries/message/src/auth.rs` defines the coordinator's shared-token auth:

- `const TOKEN_BYTES: usize = 32;` with the doc comment *"32 bytes = 64 hex chars"* (`auth.rs:13-14`)
- `generate_token()` fills 32 random bytes and renders them as lowercase hex (`auth.rs:56-62`)
- The token is written to `const TOKEN_FILENAME: &str = ".dora-token";` with mode `0o600` and is explicitly treated as a runtime secret (`auth.rs:17`, `write_token`)

The committed `.adora-token` value is a 64-char lowercase-hex string — byte-for-byte the shape `generate_token()` produces. The `.gitignore` already treats the current name as a secret-not-to-commit:

```
# Dora auth token (generated at runtime)
.dora-token
.daemon-id
```

…but there is **no entry for the fork-era `.adora-token`**, so it stayed in the tree. `git check-ignore .adora-token` returns non-zero (not ignored).

## How it got in

The file was introduced by the v1.0 consolidation squash-merge, commit
[`145ccce04fe5bd03fb392d3ec5224435ea6700e9`](https://github.com/dora-rs/dora/commit/145ccce04fe5bd03fb392d3ec5224435ea6700e9)
("dora 1.0: consolidation of the Q1 2026 Rust-first rewrite", 2026-04-17), which brought the `adora` fork tree into `dora-rs/dora`. The fork generated this auth token at runtime, it was committed into the fork working tree, and the consolidation swept it into `main`. The later rename `adora → dora` (and `.adorec → .drec`) added the `.dora-token` gitignore entry but left the already-committed `.adora-token` blob behind. (Local attribution to PRs #2068/#1948 is a shallow-clone artifact — those are just the shallow boundary commits; the real introduction is `145ccce`.)

## Risk / affected paths

- **Leaked secret in a public repo + git history.** Anyone who clones or browses the repo obtains a valid-format coordinator bearer token. It is also permanently in history under `145ccce`, so removing it from the tree alone does not un-leak it.
- **Potential auth bypass if ever reused.** The coordinator/CLI/daemon auth flow (`binaries/coordinator/src/ws_server.rs`, `binaries/daemon/src/coordinator.rs`, `binaries/cli/src/ws_client.rs`) authenticates connections with `?token=<hex>` compared via `constant_time_eq`. If this value was ever a real token for any live or reachable deployment, that deployment is compromised. Current code reads `.dora-token` (not `.adora-token`), so the committed file is not loaded by today's binaries — which limits *active* exploitability — but it remains a real secret disclosure.
- **Hygiene/consistency defect.** The repo's own policy (gitignore comment + `auth.rs` 0600 handling) says this class of file must never be committed.

## Suggested fix

1. **Treat the value as compromised and rotate** any credential that ever used this token (regenerate via `generate_token()` on affected coordinators).
2. **Remove the file from the tree:** `git rm .adora-token`.
3. **Add `.adora-token` to `.gitignore`** next to `.dora-token` (cover the legacy name so it cannot be re-committed):
   ```gitignore
   # Dora auth token (generated at runtime)
   .dora-token
   .adora-token
   .daemon-id
   ```
4. **Purge it from history** (e.g. `git filter-repo --path .adora-token --invert-paths`, then force-push), since the secret remains recoverable from `145ccce` otherwise. (This is the disruptive step; the maintainers should decide given it's a squash-merge base.)
5. Optionally add a CI secret-scan (the repo mentions `run_secret_scanning`/typos gates) to catch `.{,a}dora-token`-style blobs going forward.

---
*This issue was created by a scheduled automated Claude check that reviews recent commits/PRs for bugs and regressions introduced during AI-assisted development. Please verify the findings before acting; flagged as a security item for maintainer triage.*

phil-opp:
Closes #2194

## Summary

- Removes `.adora-token` from the repo tree — this file contains a 64-hex-char coordinator bearer token that was accidentally committed during the v1.0 consolidation squash-merge.
- Adds `.adora-token` to `.gitignore` alongside `.dora-token` so the legacy name cannot be re-committed in the future.

## Notes

- Today's binaries read `.dora-token` (not `.adora-token`), so there is no active exploitability — but the value is a real-format secret that should never be in a public repository.
- As recommended in the issue, anyone who used this token on a live coordinator should rotate it by regenerating via `generate_token()`.
- Purging the value from git history (via `git filter-repo`) is outside the scope of this PR and is a maintainer decision given the squash-merge base.

## Test plan

- [ ] Verify `.adora-token` is no longer present in the tree after merge
- [ ] Verify `.gitignore` now contains `.adora-token`
- [ ] `git check-ignore .adora-token` returns zero (file is now ignored)

---


phil-opp:
Thanks for catching this — pulling the committed token out of the tree and gitignoring the legacy name is the right move. Two things worth flagging before this is considered fully resolved:

1. **The secret is still in git history.** This commit only deletes the file from the current tree. The bearer token remains fully retrievable from the earlier commit that introduced it (e.g. `git show <old-sha>:.adora-token`), so cloning the repo or browsing history still exposes it. A tree-only deletion doesn't remediate a leaked credential.

2. **The token must be treated as compromised and rotated.** Since this value has been publicly committed, scrubbing history (e.g. `git filter-repo`) alone is insufficient — anyone could already have copied it. The actual fix is to revoke/rotate the coordinator bearer token so the leaked value is no longer valid; the file removal is cleanup on top of that.

The `.gitignore` addition itself looks correct, and I confirmed nothing in the codebase reads `.adora-token` (only `.dora-token` is the live name), so there's no functional regression here.

---

---

PR #2226, opened 2026-06-16
fix(daemon): apply no-header-timeout to local_listener receive path

### LINKED ISSUE #2218: Record/replay smoke retry (#2199) retries on the #2152 shutdown-bug fingerprint; #2126 left local_listener.rs on the old 30s header timeout
_This issue was filed by an automated, scheduled Claude code-review check. The mechanism analysis below is from reading the code, not the commit messages — please sanity-check before acting._

## Summary

Two related changes around the daemon TCP read-timeout handling interact in a way that can hide a real regression:

- **#2126** (`fix(daemon): treat idle node connections as normal…`) removed the 30s header timeout from `binaries/daemon/src/node_communication/tcp.rs` (the path that **spawned `path:` nodes** use) but left the **parallel `local_listener.rs` path** on the old 30s header timeout.
- **#2199** (`test(smoke): retry record-replay contract test on daemon TCP-read timeout`) then wrapped the record+replay cycle in a retry that re-runs on the **`"TCP read header timed out"` / `"TCP read body timed out"`** signatures, classifying them as retryable "infra stalls."

The problem: those exact strings are, per the still-open **#2152**, the documented fingerprint of a *real* unresolved shutdown-path race (a node fails to exit after stop; the daemon spins on TCP read timeouts with no SIGKILL escalation). A test that retries up to 3× on its own project's known-bug fingerprint can turn a genuine regression green.

### 1. #2126 is incomplete — `local_listener.rs` still has the 30s header timeout

After #2126, `node_communication/tcp.rs:69-72` passes `None`:

```rust
// No header timeout: a node connection may legitimately stay idle
let raw = match socket_stream_receive_with_header_timeout(&mut self.0, None).await {
```

But the dynamic-node path was not converted — `binaries/daemon/src/local_listener.rs:155` still calls the 30s variant:

```rust
let raw = match socket_stream_receive(connection).await { // 30s header timeout
```

and on timeout it logs `warn!("{err:?}")` and `break`s the connection loop (`local_listener.rs:141-144`). So the same "idle/slow first read is treated as a hard fault" behavior that #2126 set out to remove still exists for the dynamic-node config-fetch path. (In the common dynamic-node flow the client sends `NodeConfig` immediately and drops the connection, so this is unlikely to fire in practice — but it's the one remaining source of the `"TCP read header timed out"` string after #2126, which is exactly what #2199 retries on.)

### 2. #2199's retry rationale no longer matches the code, and conflates infra noise with a real bug

The retry's comment (`tests/example-smoke.rs:1101-1108`) says the daemon "aborts a first-message read after `TCP_READ_TIMEOUT` (30s)". That was true *before* #2126, but record/replay nodes are injected as ordinary `path:` nodes (`binaries/cli/src/command/record.rs:189`) and therefore connect via `node_communication/tcp.rs`, where #2126 **removed** the header timeout. So the "slow first-message read → header timeout" failure mode the retry is built around is no longer reachable on that path.

What the retry *does* still match (`tests/example-smoke.rs:1146-1154`):

```rust
if combined.contains("TCP read header timed out")
    || combined.contains("TCP read body timed out")
{
    RecordReplayFailure::InfraTimeout(detail) // retried up to 3x
```

Per **#2152**, both of those strings are the signature of a genuine shutdown-path race, not infra noise. The retry only fires when a step actually failed (non-zero exit or missing `SUCCESS` marker — `example-smoke.rs:1109-1121`), i.e. exactly the situation where a real regression would surface. The `dora record`/`dora replay` calls use blocking `.output()` with no per-command timeout, so a *silent* hang (the post-#2126 #2152 shape, where no header timeout fires anymore) would hang the test rather than retry — but any failure that *does* emit one of these strings gets retried past.

Net effect: the retry gives false confidence (its stated trigger is unreachable on the exercised path) and, where it is reachable, risks masking the #2152 regression in the record/replay dataflow.

## Affected code

- `binaries/daemon/src/node_communication/tcp.rs:69-72` (#2126 — converted)
- `binaries/daemon/src/local_listener.rs:155` (#2126 — **not** converted)
- `tests/example-smoke.rs:1101-1154` (#2199 — retry + classifier)
- Related: #2152 (open), #2089

## Suggested fix

1. Complete #2126: give `local_listener.rs::receive_message` the same `socket_stream_receive_with_header_timeout(connection, None)` treatment (the body read keeps `TCP_READ_TIMEOUT`), so the two daemon-side node-receive paths are consistent.
2. Reconsider #2199's retry: at minimum, fix the now-inaccurate comment; better, do **not** classify the #2152 timeout fingerprints as retryable. If the record/replay test is flaky for a genuine infra reason, bound each `dora record`/`dora replay` with an explicit per-command timeout and retry on *that* (a hang), rather than on a string the project elsewhere treats as a real bug.

phil-opp:
Closes #2218

## Summary

PR #2126 removed the 30s TCP read-header timeout from the `path:` node communication path (`node_communication/tcp.rs`), so idle node connections aren't killed between messages. However, the parallel **dynamic-node config-fetch path** (`local_listener.rs`) was left on the old `socket_stream_receive` wrapper, which still applies `TCP_READ_TIMEOUT` to the header read.

This makes `local_listener.rs` the last remaining source of the `"TCP read header timed out"` string after #2126 — which is the exact string that PR #2199's record/replay retry classifier treats as retryable "infra noise."

## Changes

- `local_listener.rs`: switch `receive_message` to `socket_stream_receive_with_header_timeout(connection, None)` so the two daemon-side receive paths are consistent.
- `socket_stream_utils.rs`: add `#[allow(dead_code)]` to `socket_stream_receive`, which is now only called from tests (both production call sites use the explicit header-timeout form).

## Test plan

- [ ] `cargo clippy -p dora-daemon -- -D warnings` passes
- [ ] `cargo test -p dora-daemon` passes

---

### DIFF
--- a/binaries/daemon/src/local_listener.rs
+++ b/binaries/daemon/src/local_listener.rs
@@ -1,4 +1,4 @@
-use crate::socket_stream_utils::{socket_stream_receive, socket_stream_send};
+use crate::socket_stream_utils::{socket_stream_receive_with_header_timeout, socket_stream_send};
 use dora_message::{
     daemon_to_node::DaemonReply,
     node_to_daemon::{DaemonRequest, DynamicNodeEvent, Timestamped},
@@ -152,7 +152,7 @@ async fn handle_connection_loop(
 async fn receive_message(
     connection: &mut TcpStream,
 ) -> eyre::Result<Option<Timestamped<DaemonRequest>>> {
-    let raw = match socket_stream_receive(connection).await {
+    let raw = match socket_stream_receive_with_header_timeout(connection, None).await {
         Ok(raw) => raw,
         Err(err) => match err.kind() {
             ErrorKind::UnexpectedEof
--- a/binaries/daemon/src/socket_stream_utils.rs
+++ b/binaries/daemon/src/socket_stream_utils.rs
@@ -25,6 +25,7 @@ pub async fn socket_stream_send(
     Ok(())
 }
 
+#[allow(dead_code)]
 pub async fn socket_stream_receive(
     connection: &mut (impl AsyncRead + Unpin),
 ) -> std::io::Result<Vec<u8>> {


phil-opp:
This looks correct. I confirmed that `binaries/daemon/src/node_communication/tcp.rs` already calls `socket_stream_receive_with_header_timeout(..., None)`, so switching `local_listener::receive_message` to the same form makes the two daemon-side node-receive paths consistent and finishes the #2126 work for the dynamic-node config-fetch path. Idle nodes will no longer hit the 30s header timeout between requests, while a mid-frame stall still trips `TCP_READ_TIMEOUT` on the body read.

After this change `socket_stream_receive` has no remaining production callers (only the tests in `socket_stream_utils.rs`), so the added `#[allow(dead_code)]` is accurate rather than masking a real dead-code problem. No issues found.

---

---

PR #2228, opened 2026-06-16
fix(hub): harden manifest_digest, git_pin tamper warning, and URL host validation

### LINKED ISSUE #2220: Hub --locked tamper check: manifest_digest() hashes empty bytes on serialization failure (digest collision)
_This issue was filed by an automated, scheduled Claude code-review check. Low severity / hardening — the failure paths are practically unreachable today, but they sit in trust-boundary code (untrusted index entries), so they're worth tightening. All from #2178/#2179 (`d195e87`, `087981f`)._

## Primary: `manifest_digest()` defaults to empty bytes on serialize failure

`binaries/cli/src/command/build/hub.rs:80-87`:

```rust
fn manifest_digest(manifest: &NodeManifest) -> String {
    use sha2::{Digest, Sha256};
    let bytes = serde_json::to_vec(manifest).unwrap_or_default();   // <-- empty on failure
    Sha256::digest(&bytes) ...
}
```

This digest is recorded in the lockfile so `--locked` can detect a rewritten/tampered index entry. If serialization ever fails, `unwrap_or_default()` yields empty bytes, so **any** manifest that fails to serialize hashes to the same constant digest — i.e. the tamper gate would treat two different non-serializable manifests as identical. `NodeManifest` serialization effectively never fails in practice, so this is theoretical, but a tamper-detection digest should never silently collapse to a constant.

**Suggested fix:** propagate the error (`?`) instead of digesting empty bytes — a manifest that can't be serialized should be a hard error, not a constant digest.

## Related minor observations (low confidence)

- **Swallowed `git_pin()` error in the `--locked` agreement check** — `binaries/cli/src/command/build/hub.rs:298`: `if let Ok((git, rev)) = entry.source.git_pin()` silently skips the "index repo/rev/subdir diverged" warning when the entry's source is malformed. Not a security hole (the build still proceeds from the hex-validated lockfile pin, and the separate `manifest_digest` gate still runs), but a tamper signal can be dropped silently. Consider pushing a warning on `Err` instead of continuing.

- **`validate_git_url` accepts leading-dash hosts** — `libraries/hub-client/src/lib.rs:31-45`: `ssh://-oProxyCommand=…` / `git@-…` pass validation (scheme + no-control-char checks only). Not currently exploitable — every in-scope caller that shells out uses a `--` argv separator (`transport.rs`, `fetch.rs`) and node-source clones go through `git2`/libgit2 with a parsed `url::Url`, not argv. But the validator is documented as the single chokepoint "before it is ever passed to a `git` subprocess," so it should also reject hosts beginning with `-` (classic git ssh-URL option-injection shape, cf. CVE-2017-1000117) as defense-in-depth against a future bare-`git` caller.

## Affected code

- `binaries/cli/src/command/build/hub.rs:80-87`, `:298` (#2179 / `087981f`)
- `libraries/hub-client/src/lib.rs:31-45` (#2178 / `d195e87`)

phil-opp:
Closes #2220

## Summary

Three hardening fixes in the hub trust-boundary code:

### 1. `manifest_digest()` — propagate serialization error (primary fix)

`binaries/cli/src/command/build/hub.rs:82` used `serde_json::to_vec(manifest).unwrap_or_default()`. If serialization ever fails, the previous code digested an empty byte slice — every non-serializable manifest would hash to the same constant, making the `--locked` tamper check ineffective. The function now returns `eyre::Result<String>` and propagates the error with `?`.

### 2. Swallowed `git_pin()` error in the `--locked` index-agreement check

`build/hub.rs:298` used `if let Ok((git, rev)) = entry.source.git_pin()`, silently dropping the tamper warning when the index entry's source was malformed. A malformed source is itself a suspicious signal. The check is now a `match` that emits a warning in the `Err` arm too.

### 3. `validate_git_url()` — reject leading-dash hosts

`ssh://-oProxyCommand=...`, `https://-evil.com/r`, and `git@-evil.com:...` all pass the scheme check but have a leading `-` in the host, making them look like option flags to a bare `git` subprocess (CVE-2017-1000117 shape). All current callers use libgit2 or a `--` separator, but `validate_git_url` is documented as the single pre-subprocess chokepoint. Three new tests cover all leading-dash variants.

## Test plan

- [ ] `cargo test -p dora-hub-client --lib` — 35 tests pass, including new `rejects_leading_dash_in_host`
- [ ] `cargo test -p dora-cli --lib` — 214 tests pass
- [ ] `cargo clippy -p dora-cli -p dora-hub-client -- -D warnings` passes

---

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -77,13 +77,13 @@ fn normalize_git_source_url(url: &str) -> String {
 /// lockfile so `--locked` can detect a rewritten index entry (entrypoint,
 /// build command, or typed contract). `serde_json` sorts the manifest's
 /// `BTreeMap` fields, so the serialization — and the digest — is canonical.
-fn manifest_digest(manifest: &NodeManifest) -> String {
+fn manifest_digest(manifest: &NodeManifest) -> eyre::Result<String> {
     use sha2::{Digest, Sha256};
-    let bytes = serde_json::to_vec(manifest).unwrap_or_default();
-    Sha256::digest(&bytes)
+    let bytes = serde_json::to_vec(manifest).context("failed to serialize node manifest")?;
+    Ok(Sha256::digest(&bytes)
         .iter()
         .map(|b| format!("{b:02x}"))
-        .collect()
+        .collect())
 }
 
 /// Resolve and desugar every `hub:` node in `dataflow`.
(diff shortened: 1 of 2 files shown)


phil-opp:
I reviewed this diff against the surrounding code and it looks solid. A few things I verified:

- `manifest_digest` now returns `eyre::Result<String>` and both call sites propagate with `?`; no other production callers exist, so the empty-bytes-on-failure footgun is genuinely closed.
- The `git_pin()` match correctly adds a warning on the `Err` arm while preserving the existing mismatch warning, and the catch-all keeps the happy path silent.
- The new leading-dash host check parses the authority correctly for each accepted scheme. I spot-checked the no-false-positive cases: absolute paths (`/srv/repo`) and `file:///tmp/repo` both resolve to an empty host and pass, while `ssh://-o...`, `https://-evil...`, and `git@-evil...:...` are all rejected. The userinfo-stripping via `rsplit('@')` also handles `ssh://git@host/...` correctly.

No changes requested from me — no issues found.

---

---

PR #2229, opened 2026-06-16
fix(cli): make dora validate use lockfile pins for hub: resolution

### LINKED ISSUE #2219: dora validate always resolves hub: nodes against the live index, ignoring an existing lockfile
_This issue was filed by an automated, scheduled Claude code-review check. Verified against the code, not the commit message._

## Summary

`dora validate` on a dataflow that uses `hub:` nodes always resolves against the live hub index, even when a `.dora-lock.yaml` exists (including right after a `--locked` build). Introduced by #2179 (`087981f`, hub `descriptor:` field with end-to-end resolution).

## Details

`binaries/cli/src/command/validate.rs:99-105`:

```rust
let hub_resolution = super::build::hub::resolve_hub_nodes(
    &mut descriptor,
    &mut registry,
    offline,
    None,                                                   // <-- pins = None
    &std::collections::BTreeMap::<String, std::path::PathBuf>::new(),
)?;
```

`pins` is passed unconditionally as `None`, unlike the `build` path which reads the lockfile and passes its `git_sources` as pins. Consequences:

- **Offline failure:** `dora validate` on a hub dataflow fails without network access even when a lockfile + cache fully pin the versions. (`--offline` exists as a manual workaround.)
- **Version skew:** validate may resolve a *different* index version than the one a prior `--locked` build pinned, so validation can pass/fail against versions that won't actually be built.

This is a behavior/availability issue rather than a correctness/security bug — validation output isn't persisted, and the resolver itself is the same trust-checked path used by build. But it makes `dora validate` inconsistent with `dora build` and surprising in a pinned/offline workflow.

## Affected code

- `binaries/cli/src/command/validate.rs:99-105` (#2179 / `087981f`)

## Suggested fix

Mirror `build`: if a `.dora-lock.yaml` is present, read it and pass its `git_sources` as `pins` to `resolve_hub_nodes`, so validate reflects the pinned versions and works offline against the cache.

phil-opp:
Closes #2219

## Summary

`dora validate` on a dataflow with `hub:` nodes always resolved against the live hub index, even when a `.dora-lock.yaml` existed. This caused two problems:

1. **Offline failure**: `dora validate` failed without network access even when a lockfile + cached index fully pinned the versions (the `--offline` flag was the only workaround).
2. **Version skew**: validate could resolve a *different* index version than what a prior `--locked` build pinned, making validation results inconsistent with the actual build.

## Fix

Mirror the build command's pattern: if a `.dora-lock.yaml` exists next to the dataflow, read its `git_sources` and pass them as `pins` to `resolve_hub_nodes`. If the lockfile is missing or unreadable, fall through to live resolution (previous behavior) with a warning.

The change is limited to `binaries/cli/src/command/validate.rs` — no new flags or options needed.

## Test plan

- [ ] `cargo clippy -p dora-cli -- -D warnings` passes
- [ ] `cargo test -p dora-cli --lib` passes

---

### DIFF
--- a/binaries/cli/src/command/validate.rs
+++ b/binaries/cli/src/command/validate.rs
@@ -94,13 +94,39 @@ fn validate_dataflow(dataflow: &Path, strict_types: bool, offline: bool) -> eyre
         }
     }
 
-    // Resolve hub: references against the (cached) index so their contracts
-    // take part in validation (spec §6.2 ordering note)
+    // If a lockfile exists, use its pins for hub: resolution so validate
+    // reflects the same versions a --locked build would use and works offline
+    // against the index cache. Without a lockfile, fall through to live
+    // resolution (same as before).
+    let lockfile_path = super::build::lockfile::BuildLockfile::path_for_dataflow(dataflow, None);
+    let hub_pins = if lockfile_path.exists() {
+        match super::build::lockfile::BuildLockfile::read_from(&lockfile_path) {
+            Ok(lf) => {
+                println!(
+                    "  Using hub pins from lockfile `{}`",
+                    lockfile_path.display()
+                );
+                Some(lf.git_sources)
+            }
+            Err(e) => {
+                eprintln!(
+                    "  warning: could not read lockfile `{}`: {e} — resolving hub: nodes live",
+                    lockfile_path.display()
+                );
+                None
+            }
+        }
+    } else {
+        None
+    };
+
+    // Resolve hub: references so their contracts take part in validation
+    // (spec §6.2 ordering note).
     let hub_resolution = super::build::hub::resolve_hub_nodes(
         &mut descriptor,
         &mut registry,
         offline,
-        None,
+        hub_pins.as_ref(),
         &std::collections::BTreeMap::<String, std::path::PathBuf>::new(),
     )?;
     for note in &hub_resolution.notes {


phil-opp:
I reviewed this against the diff and the build path it references. The motivation (offline validation + consistency with `--locked` builds) makes sense, but I think the gating differs from `build` in a way that could surprise users:

1. **Implicit pinning vs. `build`'s explicit `--locked`.** `build` only consults lockfile pins when `--locked` is passed (`build/mod.rs`); without it, build does live resolution even if a lockfile exists. This PR makes `validate` use the pins *whenever a lockfile is present*, with no flag to opt out. So "mirror the build command" isn't quite accurate — validate auto-pins where build does not.

2. **Stale/partial lockfile can turn a passing validate into a hard failure.** `resolve_hub_nodes` bails when `pins.is_some()` and a hub node has no pin entry (`build/hub.rs`: "`hub:` reference is not in the lockfile"). Today, with `pins == None`, validate resolves that node live and succeeds. After this change, if a lockfile exists but is out of date (e.g. a hub node was added to the dataflow since the lockfile was written), validate will fail hard instead of validating. That's a behavior regression for the common "lockfile exists but is stale" case.

Two options worth considering:
- Gate the lockfile-pin behavior behind an explicit flag so the default stays live resolution, matching `build`; or
- If implicit pinning is intended, handle the missing-pin case gracefully (fall back to live resolution for unpinned nodes) rather than letting `resolve_hub_nodes` bail.

Either way it'd be worth a test covering "lockfile exists but is missing a hub node the dataflow references."

---

---

PR #2230, opened 2026-06-16
test(smoke): scope record/replay TCP-timeout retry to record step only

### LINKED ISSUE #2223: Record/replay contract-test retry (#2199) can misclassify a dropped-message replay regression as a retryable TCP timeout
> ⚠️ This issue was filed by a **scheduled automated Claude review** of the last two days of commits. Please verify before acting.

## Summary

PR **#2199** (`test(smoke): retry record-replay contract test on daemon TCP-read timeout`, commit bbbc727) wraps `contract_record_replay_reproduces_validated_pipeline` in a 3-attempt retry that re-runs only on the daemon TCP-read-timeout signature. The intent — absorb CI-runner startup stalls without masking real regressions — is reasonable and mostly well implemented. But the **replay-equivalence** check routes its failure through the same retry classifier, which creates a narrow but real overlap where a genuine replay regression looks identical to infra noise.

## Where

`tests/example-smoke.rs` (added in #2199):

- `classify_record_replay` treats any output containing `"TCP read header timed out"` / `"TCP read body timed out"` as `InfraTimeout` (retryable).
- The replay semantic-equivalence failure (`replay did not reproduce the SUCCESS marker — semantic equivalence broken`, ~line 1249) is passed through `classify_record_replay` rather than being unconditionally `Semantic`.

## Why it's a risk

The test exists specifically to catch a replay that **drops / reorders / mutates** values. The sink uses `expected = received * 2` and bails on mismatch — that produces a clean non-timeout error and is correctly classified `Semantic`. **However**, a regression where a recorded value simply **never arrives** leaves a downstream node blocked on its first read until the daemon's `TCP_READ_TIMEOUT` (30s) fires, emitting exactly `"TCP read header timed out"`. That is the precise string the classifier treats as retryable infra noise.

So a *missing-message* replay regression on the replay step is misclassified as `InfraTimeout` and retried up to 3×. Severity is bounded — a 100%-reproducible regression still falls through to `panic!` on attempt 3 (`attempt < MAX_ATTEMPTS` guard), so it cannot silently go green — but an **intermittent** dropped-message regression could pass on a lucky retry and/or be reported only as "infra stall," defeating the contract test's purpose for exactly the failure class it was written to guard.

## Suggested fix

Scope the retryable-timeout classification to the **record** step only, where node *spawn* genuinely races the 30s timeout. For the **replay** step, the recording is already known-good (record reached SUCCESS), so a timeout there is far more likely a real replay defect — classify it as `Semantic` (fail immediately, no retry). Concretely, pass a `step` flag into `classify_record_replay` and only return `InfraTimeout` for `Step::Record`.

## Affected PR / commit
- **#2199** (bbbc727)

phil-opp:
Closes #2223

## Summary

PR #2199 added a retry to `contract_record_replay_reproduces_validated_pipeline` that retried on `"TCP read header timed out"` / `"TCP read body timed out"`. The intent was to absorb CI-runner startup stalls, but `classify_record_replay` was called for **both** the record and replay step failures with the same retryability logic.

The problem: during the replay step the recording is already known-good (record reached SUCCESS), so a TCP timeout means a message **never arrived** — that's the exact failure a dropped-message replay regression would produce (#2152 symptom). Classifying that as retryable infra noise could hide a genuine regression for up to 3 attempts.

## Fix

Pass `in_record_step: bool` to `classify_record_replay`:

- **Record step** (`true`): TCP timeout may be a slow node spawn on a loaded CI runner — genuinely infra noise, safe to retry.  
- **Replay step** (`false`): the recording is known-good; any TCP timeout means a message never arrived, which is a real semantic regression — always classified `Semantic`, never retried.

Also updates the doc comment to drop the now-outdated reference to the 30s header timeout on the `path:` node transport (removed by #2126 for that path), and tightens the `InfraTimeout` variant's doc to reflect its narrower scope.

## Test plan

- [ ] `cargo check --test example-smoke` passes
- [ ] `cargo fmt -- --check tests/example-smoke.rs` passes

---

### DIFF
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -1098,14 +1098,14 @@ fn contract_record_replay_reproduces_validated_pipeline() {
 
     let dora = dora_bin();
 
-    // The daemon<->node TCP transport aborts a first-message read after
-    // `TCP_READ_TIMEOUT` (30s, daemon `socket_stream_utils`). On a loaded CI
-    // runner the record- or replay-node spawn occasionally overshoots that on
-    // startup, surfacing as "TCP read header timed out" — an infrastructure
-    // stall, not a semantic break in the recording. Retry the whole
-    // record+replay cycle a bounded number of times on *that signature only*;
-    // a missing SUCCESS marker on an otherwise-clean run fails immediately, so
-    // a real record/replay regression is never masked by a retry.
+    // On a loaded CI runner the record-node spawn can be slow enough that
+    // the body read times out before the node sends its first message,
+    // surfacing as "TCP read body timed out" — an infrastructure stall, not a
+    // semantic break in the recording. Retry the whole record+replay cycle a
+    // bounded number of times on *that signature in the record step only*; the
+    // replay step treats TCP timeouts as semantic failures (the recording is
+    // already known-good, so a missing message is a real regression). A missing
+    // SUCCESS marker on an otherwise-clean run fails immediately.
     const MAX_ATTEMPTS: u32 = 3;
     for attempt in 1..=MAX_ATTEMPTS {
         match record_replay_once(&dora) {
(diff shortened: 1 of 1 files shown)


phil-opp:
I reviewed the diff. This is a clean, well-scoped fix to the retry classifier. Previously a TCP body-read timeout during the **replay** step — e.g. a node blocking on a message that never arrives — would be classified as `InfraTimeout` and retried, which could mask a genuine dropped/mutated-message regression since the recording is already known-good.

Threading `in_record_step` through `classify_record_replay` is the right fix, and all four call sites look correctly tagged: the two record-step failures pass `true` and the two replay-step failures pass `false`. The updated doc comments are consistent with the code.

No issues found.

---

---

PR #2231, opened 2026-06-16
fix(hub): bind shipped-type namespace gate to requested reference.namespace

### LINKED ISSUE #2222: Hub P2.12 namespace gate trusts the entry's self-declared `namespace`, allowing a tampered index entry to register cross-namespace shipped types into the shared TypeRegistry
> ⚠️ This issue was filed by a **scheduled automated Claude review** of the last two days of commits. Please verify before acting; the analysis is from reading the code, not from a reproduced exploit.

## Summary

The shipped-type namespace gate added in **#2206** (`fix(hub): enforce shipped-type namespace rule on untrusted index entries (P2.12)`, commit 95f1261) does **not** enforce what its own doc comment and PR title claim. It validates each shipped type's URN against `entry.manifest.namespace` — a value that comes from the **same untrusted index entry** — and nothing ever binds that namespace to the trusted coordinate the user actually requested (`reference.namespace`, i.e. the namespace/index the entry was fetched under). A tampered or malicious index entry can therefore self-declare any namespace and ship matching-prefix types straight into the shared `TypeRegistry`.

## Where

`binaries/cli/src/command/build/hub.rs:363-393` (introduced by #2206):

```rust
let manifest = &entry.manifest;
...
// "a rewritten entry can't register a `std/` override or a
//  cross-namespace type that a consumer's port check would resolve"
let type_issues = manifest.shipped_type_issues(registry);   // <-- gate
if !type_issues.is_empty() { eyre::bail!(...) }
// register the package's shipped types
for (urn, def) in &manifest.types {
    if registry.resolve(urn).is_none() {
        let _ = registry.add_user_type(urn, def.clone());   // <-- into SHARED registry
    }
}
```

`shipped_type_issues` (`libraries/core/src/manifest/validate.rs:174-210`, also #2206) calls `check_shipped_type_urn(urn, &self.namespace)` — i.e. it checks the URN prefix against **`self.namespace`**, the manifest's own self-declared field. `has_namespace_prefix(urn, namespace)` (validate.rs:406) is a plain `strip_prefix` check.

There is **no** check anywhere in the resolve path that `entry.manifest.namespace == reference.namespace`. `IndexCatalog::entry()` (`libraries/hub-client/src/index.rs:229-239`) deserializes the embedded `manifest: NodeManifest` verbatim from the entry YAML with no validation that its `namespace` matches the `<namespace>/<name>/<version>.yml` directory it was loaded from. (Confirmed by grep: the only `manifest.namespace` reads are in `publish.rs`, the inject Pass-1, and this gate — none compare it to the requested reference.)

## Why it's a bug

The gate is supposed to defend against an **untrusted/rewritten index entry** (that is the entire premise of P2.12). But the namespace it validates against is part of that same untrusted entry, so the check is circular:

- To resolve `evil/foo@^1`, an attacker who controls the `evil` entry sets `namespace: victim` in the embedded manifest and ships `victim/sensor/v1/Image` with a bogus Arrow body.
- `shipped_type_issues` checks `has_namespace_prefix("victim/sensor/v1/Image", "victim")` → passes. `arrow:` discriminant and fields resolve → no issues.
- The bogus type is `add_user_type`'d into the **shared** registry.
- A sibling node in the same dataflow that legitimately uses `victim/bar`'s real `victim/sensor/v1/Image` is **skipped** (`registry.resolve(urn).is_none()` is now false), so the attacker's definition wins → cross-package type confusion / mismatched Arrow schema.

The `std/` prefix is separately hard-blocked (validate.rs:379) and builtins are already-present, so those are safe — but arbitrary community-namespace shadowing is not.

A secondary facet: because this untrusted path calls only `shipped_type_issues` (not the full `validate()`), `check_namespace` is **never** run on the entry's namespace, so malformed namespaces (`Acme`, `a/b`, …) are also not rejected here. Note the trusted `--hub-override` path (build/hub.rs:163) *does* call the full `manifest.validate(registry)` — the stricter check is applied to the trusted input and the weaker check to the untrusted one, which is backwards.

## Why existing tests don't catch it

`shipped_type_issues_gates_untrusted_entries` (validate.rs:680) gives false confidence: its `minimal()` helper hardcodes `namespace: dora-rs`. Its "cross-namespace must be rejected" case ships `acme/...` only because the namespace is pinned to the trusted `dora-rs`. It never exercises an entry that sets `namespace: acme` **and** ships `acme/...` — which passes the gate cleanly.

## Suggested fix

In the hub resolver, before running the gate / registering types, assert that the entry's manifest namespace matches the requested coordinate:

```rust
if manifest.namespace != reference.namespace {
    eyre::bail!("node `{}`: index entry for `{}` declares namespace `{}` \
                 but was served under `{}` — the index may be tampered with",
        node.id, reference.key(), manifest.namespace, reference.namespace);
}
```

and/or have `shipped_type_issues` take the authoritative namespace as a parameter rather than trusting `self.namespace`, and run `check_namespace` on untrusted entries. A regression test should set the manifest's `namespace` to the attacker-chosen value (not the `minimal()` default) and assert the cross-namespace type is still rejected.

## Affected PRs / commits
- **#2206** (95f1261) — introduced the gate and the `shipped_type_issues` call (primary).
- **#2179** (087179f), **#2178** — the resolution/transport path that feeds `entry.manifest` in unvalidated.

phil-opp:
Closes #2222

## Summary

The shipped-type namespace gate (spec §6.3, P2.12, introduced in #2206) validated type URN prefixes against `manifest.namespace` — a field that comes from the **same untrusted index entry** being validated. This is circular: an attacker controlling an index entry could:

1. Set `namespace: victim` in the manifest
2. Ship `victim/sensor/v1/Image` (a valid `victim/…` type)
3. `shipped_type_issues` checks the URN against `self.namespace` = `"victim"` → **passes**
4. The bogus type is registered into the shared `TypeRegistry`
5. Legitimate `victim/…` types from other packages get skipped (already resolved) → **type confusion**

## Fix

Before calling `shipped_type_issues`, assert that `manifest.namespace == reference.namespace` — the namespace the **user actually requested** (from the `hub:` field in the dataflow YAML), not the one the entry self-declares. A mismatch is treated as a tamper signal and the build fails with a clear error.

The check is in `binaries/cli/src/command/build/hub.rs` right before the existing type gate:

```rust
if manifest.namespace != reference.namespace {
    eyre::bail!(
        "node `{}`: the index entry for `{}` declares namespace `{}` \
         but was fetched under namespace `{}` — the index may have been tampered with",
        ...
    );
}
```

## Documentation of the gap

A comment-test was added to `libraries/core/src/manifest/validate.rs` documenting that `shipped_type_issues` alone cannot catch this class of attack (it's self-referential), and that the namespace binding check in `hub.rs` is the necessary guard. The test verifies the gap exists so it cannot be accidentally closed at the wrong level.

## Test plan

- [ ] `cargo test -p dora-core -p dora-cli --lib` passes (247 tests)
- [ ] `cargo clippy -p dora-cli -p dora-core -- -D warnings` passes

---

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -363,11 +363,29 @@ pub fn resolve_hub_nodes(
         let manifest = &entry.manifest;
         let label = format!("{}@{version}", reference.key());
 
-        // The index entry is an untrusted input: validate its shipped types
-        // (namespace ownership, materializable bodies) before loading them, so
-        // a rewritten entry can't register a `std/` override or a
-        // cross-namespace type that a consumer's port check would resolve
-        // (spec §6.3, P2.12).
+        // The index entry is untrusted. Its self-declared `namespace` must
+        // match the namespace the entry was actually fetched under
+        // (`reference.namespace`): `shipped_type_issues` gates URNs against
+        // `manifest.namespace`, so an entry that sets `namespace: victim` and
+        // ships `victim/...` types would bypass the cross-namespace guard if we
+        // didn't bind it to the requested namespace first (spec §6.3, P2.12).
+        if manifest.namespace != reference.namespace {
+            eyre::bail!(
+                "node `{}`: the index entry for `{}` declares namespace `{}` \
+                 but was fetched under namespace `{}` — the index may have been tampered with",
+                node.id,
+                reference.key(),
+                manifest
+                    .namespace
+                    .chars()
+                    .filter(|c| !c.is_control())
+                    .collect::<String>(),
+                reference.namespace
+            );
+        }
+
+        // With the namespace now verified, validate shipped types
+        // (materializable bodies, no `std/` override, no cross-namespace URN).
         let type_issues = manifest.shipped_type_issues(registry);
         if !type_issues.is_empty() {
             eyre::bail!(
--- a/libraries/core/src/manifest/validate.rs
+++ b/libraries/core/src/manifest/validate.rs
@@ -710,6 +710,28 @@ outputs:
              fields:\n      - name: x\n        type: Float32\n",
         );
         assert_eq!(ok.shipped_type_issues(&reg), vec![]);
+
+        // Document the gap that the namespace-consistency check in hub.rs closes:
+        // an entry can self-declare `namespace: victim` and ship `victim/…` types
+        // that pass `shipped_type_issues` (which validates URNs against self.namespace).
+        // The hub resolver must compare manifest.namespace against the *requested*
+        // reference.namespace *before* calling this function, so this circular
+        // self-attestation cannot be exploited.
+        let attacker_namespace = "victim";
+        let attacker = NodeManifest {
+            namespace: attacker_namespace.to_string(),
+            ..minimal(
+                "types:\n  victim/sensor/v1/Image:\n    arrow: Struct\n    \
+                 fields:\n      - name: x\n        type: Float32\n",
+            )
+        };
+        // shipped_type_issues alone is NOT sufficient to catch this; it passes:
+        assert_eq!(
+            attacker.shipped_type_issues(&reg),
+            vec![],
+            "shipped_type_issues alone cannot reject a self-consistent attacker manifest \
+             (namespace binding in hub.rs is the guard)"
+        );
     }
 
     #[test]


phil-opp:
I reviewed this change against the diff. The fix looks correct: `manifest.namespace` is attacker-controlled (it ships inside the untrusted index entry), and `shipped_type_issues` validates URN prefixes against that same self-declared field — so an entry could set `namespace: victim`, ship self-consistent `victim/...` types, and pass the gate. Binding the check to `reference.namespace` (the namespace the user actually requested in the dataflow, used for the catalog lookup) before calling `shipped_type_issues` is the right place to break the circularity.

A couple of things I confirmed rather than concerns:
- The error message strips control chars from `manifest.namespace` (good, since it's hostile input) while interpolating `reference.namespace` raw — that's fine because the reference comes from the trusted local dataflow file.
- The new comment-test in `validate.rs` documents the gap clearly and asserts that `shipped_type_issues` alone cannot reject the self-consistent attacker manifest, which matches the implementation.

No issues found from my side.

---

---

PR #2236, opened 2026-06-17
fix(recording): replace silent as u16 truncation with try_from in write_entry

### LINKED ISSUE #2235: recording: RecordingWriter silently truncates node_id/output_id length to u16, producing an unreadable .drec for identifiers >64 KiB (writer caps records at 64 MiB)
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. It was found by randomly selecting a component (`libraries/recording`) and reviewing it for correctness issues. Please treat the severity assessment below as a starting point for human triage, not a confirmed defect. Severity is **low** (requires a pathologically long identifier), but it is a genuine write-time silent-corruption inconsistency, in the same family as previously-accepted findings #2083 / #2081 / #2018.

## Summary

`RecordingWriter::write_entry` (`libraries/recording/src/lib.rs:56`) encodes the `node_id` and `output_id` byte lengths as **`u16`** length prefixes, but the only size guard on the record is the **64 MiB** `MAX_RECORD_BYTES` cap on the *total* record length. Because `64 MiB` (`0x0400_0000`) is far larger than `u16::MAX` (`65 535`), the writer will happily accept a `node_id` (or `output_id`) between ~64 KiB and ~64 MiB: the total-length cap passes, but the per-field `as u16` cast **silently truncates** the length prefix.

The full identifier bytes are still written (`write_all(node_id_bytes)`), so the on-disk record is internally inconsistent: the length prefix says N bytes, but N+65536·k bytes follow. `write_entry` returns `Ok(())` — no error — yet the file can never be read back correctly.

## Affected code

```rust
// libraries/recording/src/lib.rs:56-86  (write_entry)
let record_len_usize =
    2 + node_id_bytes.len() + 2 + output_id_bytes.len() + 8 + 4 + entry.event_bytes.len();
if record_len_usize > MAX_RECORD_BYTES {           // 64 MiB cap — the ONLY length guard
    eyre::bail!("record too large to write: ...");
}
// ...
self.writer
    .write_all(&(node_id_bytes.len() as u16).to_le_bytes())?;   // ← silent truncation
self.writer.write_all(node_id_bytes)?;                          // ← full bytes still written
self.writer
    .write_all(&(output_id_bytes.len() as u16).to_le_bytes())?; // ← same problem
self.writer.write_all(output_id_bytes)?;
```

Note the contrast with `event_bytes`, which uses a `u32` length prefix (`as u32`, line 80) — and `u32::MAX` > 64 MiB, so the cap protects it. The `u16` fields have no such protection.

## Why this is a (latent) correctness bug, not just a coverage gap

- `NodeId`/`DataId` validation (`libraries/message/src/id.rs`) enforces a character set but **no maximum length**, so a >64 KiB identifier is a structurally valid `NodeId`. The recorder is the component that breaks.
- The defect is a **write succeeds → read fails** asymmetry: a producer that recorded a dataflow believes the `.drec` is intact; on replay, `next_entry()` mis-parses (reads the truncated length, then interprets the middle of the identifier bytes as the next `output_id_len`/`timestamp`/`event_bytes_len`), yielding either a decode `Err` or a wrong entry — data loss with no signal at record time.
- A writer should never emit a record it cannot itself read back. The 64 MiB record cap and the 64 KiB field width are simply inconsistent.

## Reproducer sketch

```rust
let header = sample_header();
let huge_node = "a".repeat(70_000);          // valid NodeId chars; > u16::MAX, < 64 MiB
let entry = sample_entry(&huge_node, "out", 0, b"x");

let mut buf = Vec::new();
let mut w = RecordingWriter::new(&mut buf, &header).unwrap();
w.write_entry(&entry).unwrap();              // ← succeeds today (BUG: should error)
w.finish().unwrap();

let mut r = RecordingReader::open(std::io::Cursor::new(&buf)).unwrap();
let read_back = r.next_entry();              // ← Err or wrong entry, not the original
// 70_000 as u16 == 4_464, so node_id_len on disk is 4_464, not 70_000.
```

## Suggested fix

Reject over-long identifiers explicitly in `write_entry`, mirroring the existing `MAX_RECORD_BYTES` guard, before the `as u16` casts:

```rust
let node_len = u16::try_from(node_id_bytes.len())
    .map_err(|_| eyre::eyre!("node_id too long: {} bytes (max {})", node_id_bytes.len(), u16::MAX))?;
let output_len = u16::try_from(output_id_bytes.len())
    .map_err(|_| eyre::eyre!("output_id too long: {} bytes (max {})", output_id_bytes.len(), u16::MAX))?;
// ...write node_len / output_len instead of the `as u16` casts
```

Then add a regression test (e.g. `writer_rejects_oversized_node_id`) alongside the existing `writer_rejects_oversized_record` / `writer_rejects_oversized_descriptor_yaml` tests. Alternatively, if arbitrarily long identifiers are a desired feature, widen the on-disk length fields to `u32` and bump `FORMAT_VERSION` — but the bounded-guard fix is the minimal, lowest-risk option.

- **Footer-magic false positive in `next_entry`** (`len_buf == FOOTER_MAGIC[..4]`, i.e. bytes `"DORA"`): not reachable from a valid record. A `record_len` equal to `"DORA"` LE is `0x4152_4F44` ≈ 1.09 GiB, which exceeds the 64 MiB cap, so no valid record length can collide with the footer marker.
- **`event_bytes_len` (u32) truncation**: safe — `u32::MAX` > 64 MiB, and the 64 MiB record cap is checked first.
- **`read_slice` / `read_array` bounds on corrupt input**: correctly return `Err` (not panic); `checked_add` guards the `pos + len` overflow. Already covered by `corrupt_node_id_len_returns_err_not_panic` and `corrupt_event_bytes_len_returns_err_not_panic`.
- **`total_bytes` accounting** (`record_len + 4`): consistent (counts body + length prefix per record; excludes header/footer by design).

## Affected file
- `libraries/recording/src/lib.rs` (`write_entry`, lines ~56–86)

phil-opp:
Closes #2235

## Summary

`RecordingWriter::write_entry` (`libraries/recording/src/lib.rs`) was encoding `node_id` and `output_id` byte lengths as `u16` via silent `as u16` casts. For identifiers longer than `u16::MAX` (65 535 bytes) the total-length guard (`MAX_RECORD_BYTES` = 64 MiB) would pass, but the per-field cast would silently truncate the length prefix. The result: `write_entry` returned `Ok(())` while writing a file that cannot be read back — the length prefix said N bytes but N+65536·k bytes followed.

## Fix

- Replace both `as u16` casts with `u16::try_from()` calls that return a descriptive error naming the offending field and its byte length.
- Add two regression tests alongside the existing `writer_rejects_oversized_record`:
  - `writer_rejects_oversized_node_id` — verifies an error is returned for a node_id longer than `u16::MAX`
  - `writer_rejects_oversized_output_id` — same for output_id

## Test plan

- [x] `cargo fmt -p dora-recording -- --check`
- [x] `cargo clippy -p dora-recording -- -D warnings`
- [x] `cargo test -p dora-recording` — all 14 tests pass

🤖 Generated with [Claude Code](https://claude.ai/code/session_011W4GgG3pYa7AhaGiy5m54b)

---
_Generated by [Claude Code](https://claude.ai/code/session_011W4GgG3pYa7AhaGiy5m54b)_

### DIFF
--- a/libraries/recording/src/lib.rs
+++ b/libraries/recording/src/lib.rs
@@ -67,12 +67,25 @@ impl<W: Write> RecordingWriter<W> {
         let record_len = u32::try_from(record_len_usize)
             .map_err(|_| eyre::eyre!("record length {record_len_usize} exceeds u32::MAX"))?;
 
+        let node_len = u16::try_from(node_id_bytes.len()).map_err(|_| {
+            eyre::eyre!(
+                "node_id too long: {} bytes (max {})",
+                node_id_bytes.len(),
+                u16::MAX
+            )
+        })?;
+        let output_len = u16::try_from(output_id_bytes.len()).map_err(|_| {
+            eyre::eyre!(
+                "output_id too long: {} bytes (max {})",
+                output_id_bytes.len(),
+                u16::MAX
+            )
+        })?;
+
         self.writer.write_all(&record_len.to_le_bytes())?;
-        self.writer
-            .write_all(&(node_id_bytes.len() as u16).to_le_bytes())?;
+        self.writer.write_all(&node_len.to_le_bytes())?;
         self.writer.write_all(node_id_bytes)?;
-        self.writer
-            .write_all(&(output_id_bytes.len() as u16).to_le_bytes())?;
+        self.writer.write_all(&output_len.to_le_bytes())?;
         self.writer.write_all(output_id_bytes)?;
         self.writer
             .write_all(&entry.timestamp_offset_nanos.to_le_bytes())?;
@@ -486,6 +499,38 @@ mod tests {
         assert!(
             err.to_string().contains("too large"),
             "expected 'too large' error, got: {err}"
+        );
+    }
+
+    /// `node_id` longer than `u16::MAX` bytes must return `Err`, not silently
+    /// truncate the length field and write a corrupt entry.
+    #[test]
+    fn writer_rejects_oversized_node_id() {
+        let header = sample_header();
+        let long_id = "x".repeat(usize::from(u16::MAX) + 1);
+        let entry = sample_entry(&long_id, "out", 0, b"data");
+        let mut buf = Vec::new();
+        let mut writer = RecordingWriter::new(&mut buf, &header).unwrap();
+        let err = writer.write_entry(&entry).unwrap_err();
+        assert!(
+            err.to_string().contains("node_id too long"),
+            "expected 'node_id too long' error, got: {err}"
+        );
+    }
+
+    /// `output_id` longer than `u16::MAX` bytes must return `Err`, not silently
+    /// truncate the length field and write a corrupt entry.
+    #[test]
+    fn writer_rejects_oversized_output_id() {
+        let header = sample_header();
+        let long_id = "y".repeat(usize::from(u16::MAX) + 1);
+        let entry = sample_entry("node", &long_id, 0, b"data");
+        let mut buf = Vec::new();
+        let mut writer = RecordingWriter::new(&mut buf, &header).unwrap();
+        let err = writer.write_entry(&entry).unwrap_err();
+        assert!(
+            err.to_string().contains("output_id too long"),
+            "expected 'output_id too long' error, got: {err}"
         );
     }
 


phil-opp:

The fix replaces the two silent `len() as u16` truncations of the `node_id` / `output_id` length fields with checked `u16::try_from`, returning a descriptive error instead of writing a corrupt length prefix — consistent with the `record_len` `u32::try_from` guard already present just above. The reader side already rejects oversized records, and the two new tests (`writer_rejects_oversized_node_id` / `writer_rejects_oversized_output_id`) drive the `> u16::MAX` path and assert the error rather than silent truncation. The values are written little-endian both before and after, so the on-disk format is unchanged for all valid (sub-64 KiB) IDs.

---

---

PR #2237, opened 2026-06-17
fix(validate): make hub: pin resolution best-effort, not strict-locked

### LINKED ISSUE #2234: `dora validate` regresses: implicitly enforces strict `--locked` hub resolution whenever a lockfile exists
> ⚠️ **This issue was created by a scheduled, automated Claude code-review check.** It reviews recent merges for likely bugs/regressions. Please verify before acting; the analysis may be wrong.

## Summary

PR #2229 (commit `5e30314`, "make `dora validate` use lockfile pins for hub: resolution") made `dora validate` pass the lockfile's `git_sources` to `resolve_hub_nodes` **whenever a `.dora-lock.yaml` exists** — with no `--locked` flag and no per-node fallback. Because `resolve_hub_nodes` treats `pins.is_some()` as an **all-or-nothing strict locked mode**, this silently turns `dora validate` into a strict-locked operation, and it now hard-fails on dataflows that the *old* `dora validate` and a plain `dora build` (without `--locked`) both accept.

## Root cause

`binaries/cli/src/command/validate.rs:101-131` (after #2229):

```rust
let hub_pins = if lockfile_path.exists() {
    match BuildLockfile::read_from(&lockfile_path) {
        Ok(lf) => { /* print "Using hub pins" */ Some(lf.git_sources) }
        Err(e) => { /* warn */ None }
    }
} else { None };

let hub_resolution = resolve_hub_nodes(&mut descriptor, &mut registry, offline, hub_pins.as_ref(), &BTreeMap::new())?;
```

`resolve_hub_nodes` (`binaries/cli/src/command/build/hub.rs:212-219`) interprets *any* non-`None` `pins` as "fully locked", and **bails** for any hub node not present in the lockfile:

```rust
let pin = pins.and_then(|p| p.get(&node.id));
if pins.is_some() && pin.is_none() {
    eyre::bail!(
        "node `{}`: `hub:` reference is not in the lockfile — \
         regenerate with `dora build --write-lockfile`", node.id);
}
```

Compare with `build` (`binaries/cli/src/command/build/mod.rs:334-346`), which only reads the lockfile and passes pins **when `locked == true`**:

```rust
let build_lockfile = if locked { Some(BuildLockfile::read_from(...)?) } else { None };
let hub_pins = build_lockfile.as_ref().map(|l| l.git_sources.clone());
```

So `validate` does **not** actually "mirror the build command" as the PR claims — `build` gates pin usage on the explicit `--locked` flag, whereas `validate` applies it implicitly based on mere file existence. `validate` has no `--locked` flag and no way to opt out (short of deleting the lockfile).

## Impact

Realistic dev-loop repro:

1. `dora build --write-lockfile dataflow.yml` → lockfile pins hub nodes `A`, `B`.
2. Edit `dataflow.yml`, add a new hub node `C` (or rename a node).
3. `dora validate dataflow.yml`

- **Before #2229:** `validate` passed `None`, resolved all hub nodes live → **succeeds**.
- **After #2229:** lockfile exists, so pins `{A, B}` are passed; node `C` hits `pins.is_some() && pin.is_none()` → **hard error**: *"node `C`: `hub:` reference is not in the lockfile — regenerate with `dora build --write-lockfile`"*.

Meanwhile `dora build dataflow.yml` (no `--locked`) and `dora run dataflow.yml` on the *same* file still resolve `C` live and succeed — so `validate` is now stricter than the commands it is supposed to pre-check.

The same root cause produces a second class of spurious failures even when every node is pinned: the locked path also enforces the `manifest_digest` tamper check (`hub.rs:330-341`), yank warnings, and the namespace-binding check (#2231). If the index legitimately changed a pinned entry since the lockfile was written, `validate` now hard-errors with *"the locked index entry … changed its manifest … regenerate"* — even though the user only asked to validate, not to perform a reproducible locked build.

## Suggested fix

Make pin usage **per-node and best-effort** for `validate` (validation is advisory, not a reproducibility gate):

- Preferred: teach `resolve_hub_nodes` to fall back to live resolution for a hub node that has no pin, instead of the global `pins.is_some()` bail — i.e. use a pin when present, resolve live otherwise. This keeps the strict all-or-nothing semantics only on the explicit `dora build --locked` path (which could keep enforcing "every node must be pinned" via a separate flag/param).
- Alternative: only pass pins from `validate` when the lockfile's descriptor fingerprint matches the current dataflow (mirroring `ensure_descriptor_fingerprint_matches` in `build/mod.rs:428-437`); on mismatch, warn and resolve live rather than erroring.

Either way, a stale/partial lockfile should make `dora validate` *warn and resolve live*, not fail.

## References
- Introduced by: PR #2229, commit `5e30314`
- Affected code: `binaries/cli/src/command/validate.rs:97-131`, `binaries/cli/src/command/build/hub.rs:212-219`
- Contrast: `binaries/cli/src/command/build/mod.rs:334-346`

phil-opp:
Closes #2234

## Summary

PR #2229 made `dora validate` pass lockfile pins to `resolve_hub_nodes` whenever a `.dora-lock.yaml` exists — but `resolve_hub_nodes` treats any non-`None` `pins` as all-or-nothing strict-locked mode, hard-failing for any hub node not present in the lockfile. This made `dora validate` stricter than `dora build` (without `--locked`) on the same dataflow.

**Concrete regression:** add a new hub node `C` to a dataflow that already has a lockfile for nodes `A` and `B` → `dora validate` fails with *"node `C`: `hub:` reference is not in the lockfile"* even though `dora build` and `dora run` both resolve `C` live and succeed.

## Fix

Add a `strict_pins: bool` parameter to `resolve_hub_nodes`:

- **`strict_pins: true`** (used by `dora build --locked`) — existing behaviour: all hub nodes must be pinned, hard-fail otherwise.
- **`strict_pins: false`** (used by `dora validate`) — best-effort: use a pin when present, fall through to live resolution otherwise. A stale or partial lockfile warns but doesn't block validation.

Changed files:
- `binaries/cli/src/command/build/hub.rs` — adds `strict_pins` param; the bail is now guarded by `strict_pins`.
- `binaries/cli/src/command/build/mod.rs` — passes `strict_pins: build_lockfile.is_some()` (preserves existing locked build behaviour).
- `binaries/cli/src/command/validate.rs` — passes `strict_pins: false`.

## Test plan

- [x] `cargo fmt -p dora-cli -- --check`
- [x] `cargo clippy -p dora-cli -- -D warnings`
- [x] `cargo test -p dora-cli` — all 214 tests pass

🤖 Generated with [Claude Code](https://claude.ai/code/session_011W4GgG3pYa7AhaGiy5m54b)

---
_Generated by [Claude Code](https://claude.ai/code/session_011W4GgG3pYa7AhaGiy5m54b)_

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -92,11 +92,20 @@ fn manifest_digest(manifest: &NodeManifest) -> eyre::Result<String> {
 /// the pinned commit is used verbatim and the index entry of the pinned
 /// version supplies the manifest. Without pins, each reference resolves to
 /// the highest non-yanked matching version.
+///
+/// `strict_pins` controls what happens when `pins` is `Some` but a hub node
+/// has no entry in the pins map:
+/// - `true` (used by `dora build --locked`): hard-fail immediately — the
+///   caller demanded every hub node be pinned.
+/// - `false` (used by `dora validate`): fall back to live resolution for
+///   the unpinned node rather than erroring; pinned nodes still use their
+///   lockfile entry.
 pub fn resolve_hub_nodes(
     dataflow: &mut Descriptor,
     registry: &mut TypeRegistry,
     offline: bool,
     pins: Option<&BTreeMap<NodeId, GitSource>>,
+    strict_pins: bool,
     overrides: &BTreeMap<String, PathBuf>,
 ) -> eyre::Result<HubResolution> {
     let mut resolution = HubResolution::default();
@@ -206,11 +215,17 @@ pub fn resolve_hub_nodes(
             continue;
         }
 
-        // under `--locked` (pins present), a node with no lockfile entry must
-        // fail fast — *before* any index fetch — rather than hitting the
-        // network (or an offline cache miss) only to bail afterwards.
+        // under `--locked` (pins present, strict_pins=true), a node with no
+        // lockfile entry must fail fast — *before* any index fetch — rather
+        // than hitting the network (or an offline cache miss) only to bail
+        // afterwards.
+        //
+        // When strict_pins=false (e.g. `dora validate` with a lockfile), a
+        // missing pin is not an error: the node is resolved live so that
+        // `dora validate` never fails harder than `dora build` (without
+        // `--locked`) would on the same dataflow.
         let pin = pins.and_then(|p| p.get(&node.id));
-        if pins.is_some() && pin.is_none() {
+        if strict_pins && pins.is_some() && pin.is_none() {
             eyre::bail!(
                 "node `{}`: `hub:` reference is not in the lockfile — \
                  regenerate with `dora build --write-lockfile`",
--- a/binaries/cli/src/command/build/mod.rs
+++ b/binaries/cli/src/command/build/mod.rs
@@ -349,6 +349,9 @@ pub fn build(cfg: BuildConfig) -> eyre::Result<()> {
         &mut registry,
         offline,
         hub_pins.as_ref(),
+        // `--locked` (build_lockfile.is_some()) means every hub node must have
+        // a lockfile pin; no fallback to live resolution.
+        build_lockfile.is_some(),
         &hub_override_dirs,
     )?;
     let hub_override_node_dirs = hub_resolution.override_dirs.clone();
--- a/binaries/cli/src/command/validate.rs
+++ b/binaries/cli/src/command/validate.rs
@@ -122,11 +122,17 @@ fn validate_dataflow(dataflow: &Path, strict_types: bool, offline: bool) -> eyre
 
     // Resolve hub: references so their contracts take part in validation
     // (spec §6.2 ordering note).
+    //
+    // Use best-effort pins (strict_pins=false): nodes present in the lockfile
+    // use their pinned version; nodes absent from the lockfile fall back to
+    // live resolution rather than hard-failing. This keeps `dora validate` no
+    // stricter than `dora build` (without `--locked`) on the same dataflow.
     let hub_resolution = super::build::hub::resolve_hub_nodes(
         &mut descriptor,
         &mut registry,
         offline,
         hub_pins.as_ref(),
+        false,
         &std::collections::BTreeMap::<String, std::path::PathBuf>::new(),
     )?;
     for note in &hub_resolution.notes {


phil-opp:
Two concerns with this approach, both relative to the scope of #2234:

1. **The fix only covers half the regression.** `strict_pins` here gates only the early "not in the lockfile" bail. A hub node that *is* in the lockfile but whose pin has gone stale — version no longer satisfies the requirement, missing from the index, or the manifest digest changed since the lockfile was written — still hits the unconditional `eyre::bail!` calls in the `Some(pin) => { … }` arm of `resolve_hub_nodes`, which this change does not touch. So `dora validate` will still fail harder than a plain `dora build` on the "index entry changed since the lockfile was written" case that #2234 also calls out. To fully match `build` (no `--locked`), an unusable *existing* pin needs to fall back to live resolution (with a warning), not just a missing one.

2. **This appears branched off an older `main`.** The `resolve_hub_nodes` signature in this diff has no `binary_pins` parameter, but `main` already has it (from #2233), and the PR currently shows as not mergeable. After rebasing, the binary-pin path would also need the same best-effort treatment, otherwise it stays strict.

#2238 looks like it addresses both of these (extracts `resolve_locked_git`/`resolve_live` and degrades to live resolution for any unusable git *or* binary pin, on top of current `main`). Worth deciding between the two rather than landing both.

---
*🤖 This is a fully automated review by Claude (Claude Code). No human has reviewed these comments. Please treat findings as suggestions to verify, not authoritative conclusions.*

---

heyong4725:
Thanks @phil-opp — your root-cause analysis here is spot-on: #2229 made `dora validate` pass lockfile pins to `resolve_hub_nodes`, which treats any non-`None` `pins` as all-or-nothing strict-locked mode.

Closing as **superseded by #2238** (now merged, closing #2234). #2238 independently landed the same diagnosis with slightly wider coverage — a strict superset of this fix:

- **Missing pin → live** — same as here.
- **Stale / rewritten pin → live** — this branch still hard-`bail!`s on a manifest-digest change (`hub.rs:345-356`) regardless of `strict_pins`, so `dora validate` would still fail harder than `dora build` on a *stale* (not just *missing*) pin. #2238 falls back to live resolution with a note in that case too.
- **Binary pins** — #2238 is on top of P2.8 (#2233), so it also threads `binary_pins` through the resolver. This branch predates that merge and would need a rebase onto the `NodeBytes` source architecture.
- **`--strict-types`** — #2238 emits the fallback diagnostics as *notes* (not warnings), so `dora validate --strict-types` doesn't fail on a best-effort fallback.

Sorry for the parallel work, and thanks for catching this and writing it up so clearly.

PR #2238, opened 2026-06-17
fix(cli): dora validate uses lockfile pins best-effort, not strict-locked (#2234)

### LINKED ISSUE #2234: `dora validate` regresses: implicitly enforces strict `--locked` hub resolution whenever a lockfile exists

heyong4725:
Fixes #2234.

## Problem

PR #2229 made `dora validate` pass the lockfile's pins to `resolve_hub_nodes` **whenever a `.dora-lock.yaml` exists** — with no `--locked` flag and no per-node fallback. Since `resolve_hub_nodes` treated any non-`None` `pins` as **all-or-nothing strict locked mode**, this silently turned `dora validate` into a strict-locked operation that hard-fails on dataflows the *old* `validate` and a plain `dora build` both accept:

- add a hub node not yet in the lockfile → `validate` errors "not in the lockfile" (while `dora build` / `dora run` resolve it live and succeed);
- an index entry that legitimately changed since the lockfile was written → `validate` errors on the manifest-digest tamper check.

## Fix

Add an explicit `locked: bool` to `resolve_hub_nodes`:

- **strict** (`dora build --locked`, `locked = true`): every hub node must be pinned and each pin must still be valid, else hard-fail — **unchanged**.
- **best-effort** (`dora validate`, `locked = false`): use a pin when valid; a **missing or stale pin falls back to live resolution with a warning** instead of failing.

The locked vs. live resolution is extracted into `resolve_locked_git` / `resolve_live`, so the dispatch can try the pin and degrade gracefully. The fingerprint alternative from the issue wouldn't work — hub nodes don't appear in the descriptor git-source fingerprint, so adding a hub node wouldn't be caught.

## Test

`tests/hub-smoke.rs::validate_resolves_unpinned_hub_node_live` (real local-index fixture): after `dora build --write-lockfile`, adding an unpinned hub node makes `dora validate` **succeed** (resolves live), while `dora build --locked` still **rejects** it. The existing `hub_end_to_end` (strict `--locked` rewrite rejection) still passes.

> Note: branched off `main`; it touches `resolve_hub_nodes`, which #2233 (P2.8, pending) also refactors — I'll rebase whichever lands second.

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -548,14 +548,20 @@ pub fn resolve_hub_nodes(
         // live, so those warnings describe an entry we throw away. Snapshot the
         // warning count and truncate back on the fallback so they don't leak
         // (and don't inflate `validate --strict`'s warning count).
+        // the best-effort fallback is advisory: its diagnostics are *notes*, not
+        // *warnings*, so they don't fail `validate --strict-types` (a lockfile
+        // that's merely stale is not a type error).
         let warns_before = resolution.warnings.len();
+        let lockfile_present = pins.is_some() || binary_pins.is_some();
         let (version, entry, bytes) = if let Some(bpin) = binary_pin {
             match resolve_locked_binary(bpin, &catalog, &reference, &node.id, &mut resolution) {
                 Ok(resolved) => resolved,
+                // strict `--locked`: an unusable pin is fatal.
                 Err(e) if locked => return Err(e),
+                // best-effort: a stale/rewritten pin notes and resolves live.
                 Err(e) => {
                     resolution.warnings.truncate(warns_before);
-                    resolution.warnings.push(format!(
+                    resolution.notes.push(format!(
                         "node `{}`: lockfile binary pin unusable ({e:#}) — resolving live instead",
                         node.id
                     ));
@@ -565,13 +571,10 @@ pub fn resolve_hub_nodes(
         } else if let Some(pin) = pin {
             match resolve_locked_git(pin, &catalog, &reference, &node.id, &mut resolution) {
                 Ok(resolved) => resolved,
-                // strict `--locked`: an unusable pin is fatal.
                 Err(e) if locked => return Err(e),
-                // best-effort (validate): a stale/rewritten pin warns and
-                // resolves live instead of failing the whole command.
                 Err(e) => {
                     resolution.warnings.truncate(warns_before);
-                    resolution.warnings.push(format!(
+                    resolution.notes.push(format!(
                         "node `{}`: lockfile pin unusable ({e:#}) — resolving live instead",
                         node.id
                     ));
@@ -579,6 +582,16 @@ pub fn resolve_hub_nodes(
                 }
             }
         } else {
+            // no pin for this node. If a lockfile was provided it just doesn't
+            // cover this node (added since it was written) — note it so a stale
+            // lockfile isn't silently hidden until `build --locked` fails.
+            if lockfile_present {
+                resolution.notes.push(format!(
+                    "node `{}`: not in the lockfile — resolving live; \
+                     regenerate with `dora build --write-lockfile` to pin it",
+                    node.id
+                ));
+            }
             resolve_live(&catalog, &reference, &node.id, &raw_reference)?
         };
 
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -267,12 +267,154 @@ fn resolve_locked_binary(
     Ok((version, entry, bytes))
 }
 
+type Resolved = (
+    dora_hub_client::semver::Version,
+    dora_hub_client::index::IndexEntry,
+    NodeBytes,
+);
+
+/// Resolve a node from its git lockfile pin (`--locked`): the pinned commit is
+/// used verbatim, but the manifest comes from the (mutable) index entry, so the
+/// entry is cross-checked (warn on rewrite) and its manifest digest hard-checked.
+/// Returns an error when the pin is unusable (stale version, rewritten manifest,
+/// missing from the index); the caller decides whether that is fatal (strict
+/// `--locked`) or a cue to resolve live (best-effort, e.g. `validate`).
+fn resolve_locked_git(
+    pin: &GitSource,
+    catalog: &IndexCatalog,
+    reference: &PackageRef,
+    node_id: &NodeId,
+    resolution: &mut HubResolution,
+) -> eyre::Result<Resolved> {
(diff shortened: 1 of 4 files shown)


phil-opp:
I reviewed the diff directly. This is the more complete of the two competing fixes for #2234 (vs #2237) and it addresses both concerns raised on that PR:

- It is rebased on current `main`, so the `binary_pins` parameter is present and binary pins get the same best-effort treatment as git pins (the `resolve_locked_binary` arm degrades to `resolve_live` on error when not locked).
- It covers the "existing-but-stale pin" case, not just the "missing pin" case: extracting `resolve_locked_git`/`resolve_live` and matching on `Err(e) if locked => return Err(e)` / else warn-and-resolve-live means a pin that is present but unusable (stale version, rewritten manifest, missing from the index, digest changed) now falls back to live resolution under `validate` instead of hard-failing.

The guard change from `pins.is_some()` to `locked` is behavior-preserving for `build`: `build_lockfile` is `Some` only when `locked` (build/mod.rs:334) and `hub_pins` derives from it, so `pins.is_some() ⟺ locked` on the build path. `validate` passes `false`, which is the intended relaxation. The regression test exercises both directions (validate resolves the unpinned node live; `--locked --offline` still rejects it).

No significant issues found. One cosmetic note (non-blocking): when `resolve_locked_git` bails on a digest mismatch *after* having already pushed the "index entry no longer matches the lockfile pin" warning, the caller then pushes a second "lockfile pin unusable … resolving live" warning, so that narrow case emits two overlapping warnings for the same pin. Purely a log-noise nit.

---

---

phil-opp:
**Automated review by Claude** (fully automated — please verify).

Re-reviewed after the two follow-up commits (`91ad38c`, `5a7b9240`). They resolve the double-warning nit from my earlier review: on the best-effort `Err` fallback the code now snapshots `warns_before` and `resolution.warnings.truncate(warns_before)`, so a discarded pin's yank/index-rewrite warnings no longer leak alongside the fallback message. The fallback diagnostics are also now pushed to `resolution.notes` instead of `warnings`, so a merely-stale pin no longer trips `validate --strict-types`, and a missing pin emits a "not in the lockfile — resolving live" note. The two added tests cover the stale-pin and strict-types paths.

No new issues found in the latest diff.

---

PR #2240, opened 2026-06-17
chore(deps): bump git2 0.20.4 -> 0.21 (fixes RUSTSEC-2026-0183)

heyong4725:
## What

Bump the workspace `git2` dependency `0.20.4 → 0.21` to fix **RUSTSEC-2026-0183**.

## Why

git2 `< 0.21.0` passes a null pointer to `slice::from_raw_parts()` in `Remote::list()` when a remote advertises no references — undefined behaviour (the advisory is **INFO / unsound**). The only patched release is **`>= 0.21.0`**, so a bump is the only fix (no 0.20.x patch exists).

The advisory blocks the **supply-chain audit gate (`cargo-audit` + `cargo-deny`)** on every open PR, so this unblocks them all.

## Verification

`git2` is a direct workspace dependency (used by `dora-core`, `dora-cli`, `dora-daemon`). 0.21.0 is a clean upgrade:

- `cargo check --workspace` (excl. Python) builds; `dora-core`/`dora-cli`/`dora-daemon` compile **unchanged** (no API breaks).
- `cargo deny check advisories` → **advisories ok**; `cargo audit` → **no vulnerabilities** (only pre-existing allowed warnings).
- The git2 clone path still works at runtime: `tests/hub-smoke.rs::hub_end_to_end` passes.
- Single `git2 0.21.0` in `Cargo.lock` (no duplicate versions).

Diff is `Cargo.toml` (one line) + `Cargo.lock`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2991,9 +2991,9 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.20.4"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7b88256088d75a56f8ecfa070513a775dd9107f6530ef14919dac831af9cfe2b"
+checksum = "ddddbf932745a6be37109b6112d3ee09696106f848449069d3a57bba937ab82e"
 dependencies = [
  "bitflags 2.13.0",
  "libc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,7 +133,10 @@ pyo3 = { version = "0.28", features = [
 ] }
 pythonize = "0.28"
 pyo3-build-config = { version = "0.28", features = ["resolve-config", "abi3-py311"] }
-git2 = { version = "0.20.4", features = ["vendored-openssl"] }
+# git2 0.21 dropped `https`/`ssh` from its default features (0.20 had
+# `default = ["ssh", "https"]`), so request them explicitly — otherwise
+# libgit2 ships without a TLS transport ("no TLS stream available").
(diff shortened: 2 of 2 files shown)


phil-opp:
**Automated review by Claude** — this is a fully automated review; a human maintainer should confirm before merging.

I found an issue: this bump silently disables both HTTPS and SSH git transport.

In git2 0.21.0 the default feature set changed from `default = ["ssh", "https"]` to `default = []` (the `ssh`, `https`, and `cred` features are no longer default). This PR keeps only `features = ["vendored-openssl"]`, and `vendored-openssl` does **not** transitively enable `https` or `ssh` — it only flips openssl-sys/libgit2-sys to vendored mode. So after this bump, libgit2 is compiled without HTTPS or SSH transport.

The lockfile diff confirms this rather than the "clone path still works" claim:
- `openssl-probe` is removed — it's a dependency of the `https` feature, so its absence means `https` is off.
- `libssh2-sys` is removed — it's the `ssh` feature's transport, so SSH is off.

The codebase clones git sources over both protocols (`clone_into`/`fetch_changes` in `libraries/core/src/build/git.rs`, and `libraries/hub-client` explicitly validates and accepts `https://`, `ssh://`, and `git@` URLs). With this bump as written, `dora build` against any git source would fail at runtime with an "unsupported URL protocol" error.

Suggested fix: restore the transports explicitly, e.g.

```toml
git2 = { version = "0.21", features = ["vendored-openssl", "https", "ssh"] }
```

(`https`/`ssh` both pull in `cred` transitively.) It's also worth adding a smoke/integration test that actually performs an `https://` clone so this regression can't recur silently — the existing checks in the PR description don't appear to catch it.

---

PR #2266, opened 2026-06-18
fix(tracing): truncate span field values at UTF-8 char boundary

### LINKED ISSUE #2263: Panic risk: `String::truncate` at fixed byte offset in span_store `FieldVisitor` can split a UTF-8 char
## Summary

`FieldVisitor::push_field` in the telemetry tracing layer truncates an
over-long field value at a fixed **byte** offset using `String::truncate`.
`String::truncate(new_len)` panics if `new_len` does not fall on a UTF-8
char boundary, so a field value longer than `MAX_FIELD_VALUE_BYTES` (4096)
whose 4096th byte lands in the middle of a multi-byte character will panic
inside the `tracing` layer.

## Location

`libraries/extensions/telemetry/tracing/src/span_store.rs:84-93`

```rust
fn push_field(&mut self, name: &str, mut value: String) {
    if self.0.len() >= MAX_FIELDS_PER_SPAN {
        return;
    }
    if value.len() > MAX_FIELD_VALUE_BYTES {
        value.truncate(MAX_FIELD_VALUE_BYTES); // <-- panics if 4096 is not a char boundary
        value.push_str("...");
    }
    self.0.push((name.to_string(), value));
}
```

`value.len()` is the byte length, and `MAX_FIELD_VALUE_BYTES` is a fixed
byte count, so the cut point is not guaranteed to be a char boundary.

## Why it's reachable

`push_field` is fed by `record_str` (pushes `value.to_string()` verbatim)
and `record_debug` (pushes `format!("{value:?}")`). Both can produce
arbitrary-length strings containing multi-byte UTF-8 characters — e.g. a
span field carrying a path, a JSON blob, a log line, or any user/robot
data with non-ASCII content. If such a value exceeds 4096 bytes and a
character straddles byte 4096, `truncate` panics.

Because this runs inside a `tracing_subscriber::Layer` callback
(`on_new_span` / `on_record`), the panic propagates out of whatever code
emitted the span/event — potentially aborting a node or the daemon — and
it is data-dependent, so it would slip through normal testing.

## Reproduction

```rust
let mut s = "a".repeat(4095);
s.push('😀');          // 4-byte char occupying bytes 4095..4099
assert!(!s.is_char_boundary(4096));
s.truncate(4096);      // panics: assertion failed: self.is_char_boundary(new_len)
```

Verified standalone: this panics with
`assertion failed: self.is_char_boundary(new_len)`.

## Suggested fix

Truncate on a char boundary instead of a raw byte index, e.g.:

```rust
if value.len() > MAX_FIELD_VALUE_BYTES {
    let mut end = MAX_FIELD_VALUE_BYTES;
    while !value.is_char_boundary(end) {
        end -= 1;
    }
    value.truncate(end);
    value.push_str("...");
}
```

(or use `value.floor_char_boundary(MAX_FIELD_VALUE_BYTES)` once it is
stabilized). A regression test with a multi-byte char straddling the
boundary should accompany the fix.

---
*This issue was created by an automated, scheduled Claude code-review check.*

phil-opp:
## Summary

Fixes the panic reported in #2263.

`FieldVisitor::push_field` in `libraries/extensions/telemetry/tracing/src/span_store.rs` called `String::truncate(MAX_FIELD_VALUE_BYTES)` where `MAX_FIELD_VALUE_BYTES = 4096`. `String::truncate` panics if the byte index does not fall on a UTF-8 character boundary. Any span field carrying non-ASCII content (paths, JSON blobs, emoji) that exceeded 4096 bytes and had a multi-byte character straddling byte 4096 would panic inside the `tracing` subscriber layer — propagating out of the code that emitted the span and potentially aborting a node or the daemon.

## Fix

Walk backwards from `MAX_FIELD_VALUE_BYTES` to find the nearest valid char boundary before truncating:

```rust
let mut end = MAX_FIELD_VALUE_BYTES;
while !value.is_char_boundary(end) {
    end -= 1;
}
value.truncate(end);
value.push_str("...");
```

## Tests

Added a regression test (`field_visitor_truncates_at_char_boundary`) that constructs a string where a 4-byte emoji (`😀`) straddles byte 4096 — the exact case that previously panicked — and asserts the stored value is valid UTF-8 ending with `"..."`.

All 3 `dora-tracing` unit tests pass; clippy and fmt clean.

Closes #2263

---
_Generated by [Claude Code](https://claude.ai/code/session_01QheETdNcNiZ6JZa8oz5sRK)_

### DIFF
--- a/libraries/extensions/telemetry/tracing/src/span_store.rs
+++ b/libraries/extensions/telemetry/tracing/src/span_store.rs
@@ -86,7 +86,11 @@ impl FieldVisitor {
             return;
         }
         if value.len() > MAX_FIELD_VALUE_BYTES {
-            value.truncate(MAX_FIELD_VALUE_BYTES);
+            let mut end = MAX_FIELD_VALUE_BYTES;
+            while !value.is_char_boundary(end) {
+                end -= 1;
+            }
+            value.truncate(end);
             value.push_str("...");
         }
         self.0.push((name.to_string(), value));
@@ -217,6 +221,23 @@ where
 mod tests {
     use super::*;
     use tracing_subscriber::{Registry, layer::SubscriberExt};
+
+    #[test]
+    fn field_visitor_truncates_at_char_boundary() {
+        // A 4-byte emoji straddles byte 4096 — raw truncate would panic.
+        let mut value = "a".repeat(4095);
+        value.push('😀'); // occupies bytes 4095..4099
+        assert!(!value.is_char_boundary(4096));
+
+        let mut visitor = FieldVisitor(Vec::new());
+        visitor.push_field("key", value);
+
+        let (_, stored) = &visitor.0[0];
+        assert!(stored.ends_with("..."));
+        assert!(stored.is_char_boundary(stored.len()));
+        // Stored value must be valid UTF-8 (would panic on display otherwise)
+        let _ = stored.len();
+    }
 
     #[test]
     fn span_store_evicts_oldest() {


phil-opp:
**Automated review by Claude** — _fully automated, no human has vetted this; treat as a suggestion._

No issues found. Reviewed the diff directly:

- The char-boundary walk-back is correct and cannot underflow: the surrounding `value.len() > MAX_FIELD_VALUE_BYTES` guard means `end = 4096` is always a valid in-range index, and `is_char_boundary(0)` is always `true`, so the loop terminates after at most 3 decrements (max UTF-8 char width). `truncate(end)` then lands on a valid boundary, fixing the panic from #2263.
- The regression test reproduces the exact reported case (a 4-byte `😀` straddling byte 4096) and would have panicked inside `push_field` before this fix, so it genuinely guards the path.

One minor (non-blocking) note: a couple of the test assertions are tautological — `stored.is_char_boundary(stored.len())` is always true for any `String`, and `let _ = stored.len();` is a no-op. The real coverage comes from `push_field` not panicking plus `ends_with("...")`, so the test is still meaningful; the extra lines just don't add anything.

---

phil-opp:

No issues found. The backward `is_char_boundary` walk is safe — `end` starts at `MAX_FIELD_VALUE_BYTES` (≤ `value.len()` in this branch) and the loop always terminates since byte 0 is a valid boundary. The regression test correctly reproduces the original panic case (4-byte emoji straddling byte 4096) and asserts valid UTF-8 output.

---

phil-opp:
/trunk merge

(Re-submitting: the only failing required check was "Semantic contract tests: Cancelled" — the job was interrupted mid-run, not a real test failure. All 35 other checks passed including Test, Clippy, E2E, Benchmark, Format, etc. The `span_store.rs` change has no interaction with the contract tests.)

---

PR #2267, opened 2026-06-18
fix(memory-pool): replace flaky sum-based assertion with monotonic counter check

### LINKED ISSUE #2264: memory-pool transport (#2168): 2,977-line unsafe/FFI feature merged with no integration test coverage + a flaky example assertion

## Summary

PR #2168 (`feat: add pinned memory-pool transport for CPU/CUDA tensor transfer`, commit `84da59a`) merged a large new zero-copy transport: **30 files changed, 2,977 insertions**, including ~**44 `unsafe` blocks** of raw pointer arithmetic, a hand-rolled seqlock, `/dev/shm` IPC, and `ctypes`/embedded-CUDA FFI in `apis/python/node/src/lib.rs` and `apis/python/node/dora/cuda.py`.

Despite the size and the amount of `unsafe`/FFI, **none of the actual transport code paths are exercised by any automated test in CI**, and one of the shipped example nodes contains a probabilistically-flaky assertion. These are the two concrete, isolated findings below.

The rest of the recent (last ~2 days) merges I reviewed — `bc1a02f` (recording u16 try_from), `b04858e` (scheduler ZenohInput drop log), `003e269` (local_listener header timeout), `4ac999c` (Python basicConfig once), `2d6c1db` (inference arg index), `5e30314`+`a76c0c7` (`dora validate` lockfile pins), `df1e60e` (hub binary source / `path_sha256` verification), `d43c3e5` (record/replay retry scoping), `cd6aeb8`/`8fef3d0` (CI/unwrap-budget) — all look correct and are well-covered by tests. This issue is specifically about the memory-pool feature.

---

## Finding 1 — no integration/smoke test coverage for the unsafe transport paths

The only automated tests added are 7 unit tests in `libraries/extensions/memory-pool/src/lib.rs`. Those exercise the pure-`HashMap` `MemoryPoolManager` (register / double-register / read / cross-owner-free / double-free / cleanup / poison-recovery) and **none of the `unsafe` shmem mapping, seqlock, bounds-check, DMA, or CUDA-IPC code** that constitutes the actual risk surface.

The feature ships **six example dataflows** that *do* exercise the real paths:

```
examples/memory-pool/{cpu2cpu,cpu2cuda,cuda2cpu,auto_cleanup,duplicate_free,read_after_free,write_after_free}.yml
```

but none of them is wired into the test harness. Searching the whole CI/test surface returns nothing:

```
$ grep -rn "memory-pool\|memory_pool\|cpu2cpu" .github/ scripts/ tests/ Makefile
(no matches)
```

This is notable because `cpu2cpu.yml` was added specifically as the CPU-only path (its own header comment: *"This scenario can run on GPU-less CI runners and exercises the daemon pool path"*), yet it is not referenced from `tests/example-smoke.rs`, `scripts/smoke-all.sh`, or any workflow — so even the GPU-free path that *could* run on the standard Linux CI runner never does.

This directly contradicts the repository's documented contribution policy in `CLAUDE.md`:

> | New example dataflow | Smoke test entry | Add to `tests/example-smoke.rs` and `scripts/smoke-all.sh` |
>
> New example dataflows should have tests in both modes.

**Why this is risky here specifically:** issue #1872 (the design issue for exactly this feature) explicitly set a high bar and asked for an RFC before implementation, and notes that a prior near-identical attempt (PR #1623, ~2,717 LOC, same `/dev/shm/dora_pool_*` architecture) was **closed** for being far larger than the feature needed. The merged implementation is *larger still* (2,977 insertions) and lands the unsafe machinery with no end-to-end guard. A regression in any of the pointer-arithmetic / seqlock / bounds paths would be invisible to CI.

**Suggested fix:** wire `cpu2cpu.yml` (the GPU-free scenario) into `tests/example-smoke.rs` (both `run_smoke_test` and `run_smoke_test_local`) and `scripts/smoke-all.sh`, so the daemon pool path, register/write/read/free lifecycle, and shutdown cleanup get at least one automated end-to-end run on standard Linux CI. The negative-lifecycle YAMLs (`duplicate_free`, `read_after_free`, `write_after_free`, `auto_cleanup`) are also good smoke candidates since they assert the "warn, don't crash" contract the PR claims.

---

## Finding 2 — `examples/memory-pool/receiver.py` has a probabilistically-flaky assertion

In the throughput scenario (used by `cpu2cpu.yml`, `cpu2cuda.yml`, `cuda2cpu.yml`), the receiver validates that data changed between iterations:

```python
curr_sum = int(torch_tensor[:8].sum().item())
if prev_sum is not None and SCENARIO != "write_after_free":
    assert curr_sum != prev_sum, (
        f"iteration {i}: tensor data did not change — pool write may not have propagated"
    )
prev_sum = curr_sum
```

The sender fills each message with fresh random data:

```python
random_data = data_generation.integers(1000, size=SIZE, dtype=np.int64)
```

The checksum is the sum of only the **first 8** `int64` values, each in `[0, 1000)`. Two independent draws can produce the same 8-element sum purely by chance (sum range `[0, 7992]`, roughly bell-shaped; collision probability per adjacent pair ≈ 1/3000). With `message_num=100` (~99 comparisons) that is on the order of a **few-percent spurious-failure rate per run** — even when the transport works perfectly.

This makes the example unreliable for any user who runs it to evaluate the feature, and it would make `cpu2cpu.yml` flaky if it were added to CI as suggested in Finding 1.

**Suggested fix:** make the change-detection deterministic instead of statistical — e.g., have the sender stamp a monotonically increasing counter into a known tensor element (`random_data[0] = i`) and assert on that, or checksum the full tensor rather than the first 8 elements (still not airtight, but vastly less collision-prone). The deterministic-counter approach is the robust one.

---

### Affected code
- `apis/python/node/src/lib.rs` (memory-pool register/write/read/free, seqlock, bounds checks — `unsafe`)
- `apis/python/node/dora/cuda.py` (`tensor_from_info`, CUDA FFI)
- `binaries/daemon/src/lib.rs` (`Register/Read/FreePinnedMemory` handlers, shutdown cleanup, orphan sweep)
- `libraries/extensions/memory-pool/src/lib.rs` (manager; only this has tests)
- `examples/memory-pool/*.yml`, `examples/memory-pool/receiver.py`

### Introduced by
- PR #2168 / commit `84da59a807aebbe0b5cb53ef7f94c09973499271`

### Context
- #1872 (design issue / RFC request for this feature)
- #1623 (prior, closed implementation attempt)

phil-opp:
## Summary

Addresses the two concrete findings in #2264.

### Finding 2 fix: eliminate the ~3% spurious failure rate in `receiver.py`

The old assertion summed the first 8 `int64` elements (each in `[0, 1000)`) of two consecutive random tensors and asserted they differed. With a sum range of `[0, 7992]` and ~99 comparisons per run, two adjacent draws had a ~1/3000 collision probability, giving roughly **3% spurious test failure per run** even when the transport worked perfectly.

Fix: stamp element `[0]` with the iteration counter `i` in the sender, and assert `tensor[0] == i` in the receiver. Zero collision probability.

**sender.py:**
```python
random_data[0] = i  # monotonic counter lets receiver detect change without collision risk
```

**receiver.py:**
```python
if SCENARIO != "write_after_free":
    actual = int(torch_tensor[0].item())
    assert actual == i, f"iteration {i}: tensor[0] expected {i}, got {actual}"
```

### Finding 1 (partial): wire `cpu2cpu.yml` into the test harness

`cpu2cpu.yml` is the only GPU-free scenario in the memory-pool suite. Added:
- `tests/example-smoke.rs`: two `#[ignore]` smoke tests (torch is not installed in standard PR CI; run explicitly with `cargo test --test example-smoke -- --ignored smoke_memory_pool`)
- `scripts/smoke-all.sh`: conditional block that runs `cpu2cpu.yml` when `python3 -c "import torch, tqdm"` succeeds, skips otherwise
- Coverage audit table updated

The CUDA-dependent scenarios (`cpu2cuda`, `cuda2cpu`) remain blocked by the need for NVIDIA hardware.

Closes #2264

---
_Generated by [Claude Code](https://claude.ai/code/session_01QheETdNcNiZ6JZa8oz5sRK)_

### DIFF
--- a/examples/memory-pool/receiver.py
+++ b/examples/memory-pool/receiver.py
@@ -22,7 +22,6 @@ pbar = tqdm(total=MESSAGE_COUNT)
 velocities = []
 memory_pool_id = None
 torch_tensor = None
-prev_sum = None
 
 for i in range(MESSAGE_COUNT):
     event = node.next()
@@ -38,14 +37,14 @@ for i in range(MESSAGE_COUNT):
     # the shmem bytes in place, so the receiver's existing tensor object
     # automatically reflects new data.  Turn-based signaling ensures the
     # sender has finished writing before the receiver accesses the tensor.
-    # Compute a cheap checksum to validate that the tensor data changed
-    # (and therefore the pooled transfer actually delivered new bytes).
-    curr_sum = int(torch_tensor[:8].sum().item())
-    if prev_sum is not None and SCENARIO != "write_after_free":
-        assert curr_sum != prev_sum, (
-            f"iteration {i}: tensor data did not change — pool write may not have propagated"
+    # The sender stamps element[0] with the iteration counter so we can
+    # verify propagation deterministically (sum-of-8 had ~3% collision rate).
+    if SCENARIO != "write_after_free":
+        actual = int(torch_tensor[0].item())
+        assert actual == i, (
+            f"iteration {i}: tensor[0] expected {i}, got {actual}"
+            " — pool write may not have propagated"
         )
-    prev_sum = curr_sum
 
     t_received = time.perf_counter_ns()
     delta_t = t_received - t_send
--- a/examples/memory-pool/sender.py
+++ b/examples/memory-pool/sender.py
@@ -22,6 +22,7 @@ data_generation = np.random.default_rng()
 memory_pool_id = None
 for i in range(MESSAGE_COUNT):
     random_data = data_generation.integers(1000, size=SIZE, dtype=np.int64)
+    random_data[0] = i  # monotonic counter lets receiver detect change without collision risk
     torch_tensor = torch.tensor(random_data, dtype=torch.int64, device=SENDER_DEVICE)
     t_send = time.perf_counter_ns()
     metadata = {"t_send": t_send, "scenario": SCENARIO}
--- a/scripts/smoke-all.sh
+++ b/scripts/smoke-all.sh
@@ -501,6 +501,15 @@ if [ "$RUN_PYTHON" = true ]; then
     echo "=== Queue/timeout regression tests (local, timing-sensitive) ==="
     run_local "local-queue-size-and-timeout"         "tests/queue_size_and_timeout_python/dataflow.yaml" 20
     run_local "local-queue-size-latest-data-python"  "tests/queue_size_latest_data_python/dataflow.yaml" 20
+
+    echo ""
+    echo "=== Memory-pool CPU transport (requires torch) ==="
+    if python3 -c "import torch, tqdm" 2>/dev/null; then
+        run_networked "memory-pool-cpu2cpu"       "examples/memory-pool/cpu2cpu.yml" 60
+        run_local     "local-memory-pool-cpu2cpu" "examples/memory-pool/cpu2cpu.yml" 60
+    else
+        log_skip "memory-pool-cpu2cpu" "requires torch + tqdm (pip install torch tqdm)"
+    fi
 fi
 
 # ---------------------------------------------------------------------------
--- a/tests/example-smoke.rs
+++ b/tests/example-smoke.rs
@@ -1884,6 +1884,35 @@ fn smoke_shell_node_blocked_without_flag() {
 }
 
 // ---------------------------------------------------------------------------
+// Memory-pool CPU transport (#2168)
+//
+// Requires `torch` and `tqdm` — not installed in standard PR CI. Run
+// explicitly on machines with torch available:
+//   cargo test --test example-smoke -- --ignored smoke_memory_pool
+// or via `scripts/smoke-all.sh` which gates on `python3 -c "import torch"`.
+// ---------------------------------------------------------------------------
+
+#[test]
+#[ignore = "requires `torch` and `tqdm` (not in standard CI)"]
+fn smoke_memory_pool_cpu2cpu() {
+    run_smoke_test(
+        "memory-pool-cpu2cpu",
+        "examples/memory-pool/cpu2cpu.yml",
+        Duration::from_secs(60),
+    );
+}
+
+#[test]
+#[ignore = "requires `torch` and `tqdm` (not in standard CI)"]
+fn smoke_local_memory_pool_cpu2cpu() {
+    run_smoke_test_local(
+        "local-memory-pool-cpu2cpu",
+        "examples/memory-pool/cpu2cpu.yml",
+        60,
+    );
+}
+
+// ---------------------------------------------------------------------------
 // Examples under `examples/` that do NOT have a corresponding `smoke_*` or
 // `contract_*` test in this file. Some are blocked (filed issue or external
 // dep); others are intentionally covered by a DIFFERENT CI job. Keep this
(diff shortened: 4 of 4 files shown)


phil-opp:
**Automated review by Claude** — _fully automated, no human has vetted this; treat as a suggestion._

No issues found. Reviewed the diff directly:

- The sender stamps `random_data[0] = i` (int64, no overflow for the message range) and the receiver asserts `tensor[0] == i`, which is deterministic and eliminates the collision risk of the old sum-of-8 check. The assertion is correctly skipped for the `write_after_free` scenario where reading the tensor is intentionally unsafe.
- Checking on every iteration (including `i == 0`) is strictly better than the old `prev_sum is not None` skip of the first comparison.
- The smoke-test wiring is gated correctly: `#[ignore]` on the two Rust tests and a `python3 -c "import torch, tqdm"` guard in `smoke-all.sh`, so standard PR CI (no torch) is unaffected.

Test-only / example-only change with no production-code impact.

---

phil-opp:

No issues found. The counter fix is correct, the two smoke tests are `#[ignore]`-gated, and `scripts/smoke-all.sh` gates the run on `python3 -c "import torch, tqdm"` while only invoking the CPU-only `cpu2cpu.yml` (both nodes `*_device: cpu`), so CPU CI is safe.

One note: this overlaps with #2268 and #2279, which address the same #2264 findings. #2268 additionally wires the four `*_device: cuda` negative scenarios under the torch-only gate (a problem already flagged there); this PR avoids that by staying CPU-only. Worth consolidating to a single PR.

---

phil-opp:
🤖 _Automated review by Claude (Claude Code). This is a fully automated review with no human in the loop — please treat it as a suggestion._

I found one issue I don't think the earlier reviews caught:

**The torch gate in `scripts/smoke-all.sh` checks the wrong interpreter, so the new memory-pool coverage never executes in a normal run.** The block is gated on `python3 -c "import torch, tqdm"`, but by the time it runs an earlier `--uv` Python example has already activated `target/smoke-venv`, which only installs `pyarrow numpy` + `-e apis/python/node` (no torch/tqdm). So `python3` resolves to that venv, the gate is always false, and the entries are logged as SKIP. Combined with the Rust smoke tests being `#[ignore]`-gated, this PR adds no coverage that runs in CI or a plain `smoke-all.sh` invocation — it only runs if a developer happens to have torch on the ambient `python3` *and* passes `--ignored` manually. Either drop the gate and provision deps via per-node `build:` steps (as #2279 attempts), or point the check at the interpreter the dataflow nodes actually use.

The counter assertion itself (`tensor[0] == i`) is correct and robust — checking against absolute ground truth every iteration is better than the consecutive-pair approach. Worth noting that #2267 / #2268 / #2279 are three competing fixes for #2264; consolidating to one would reduce the churn.

---

phil-opp:
**Automated review by Claude** — fully automated review, not human-verified.

The `receiver.py` counter assertion (`tensor[0] == i`, skipped only for the `write_after_free` scenario) is correct.

But this PR now appears **superseded by #2268**, which merged the same #2264 fix on 2026-06-22; this branch is consequently in merge conflict (`mergeable_state: dirty`). The previously-noted `scripts/smoke-all.sh` gate issue also remains unaddressed: the `python3 -c "import torch, tqdm"` check runs after earlier examples have activated `target/smoke-venv` (which installs only `pyarrow`/`numpy` + the node bindings, no torch/tqdm), so the gate is effectively always false and the new memory-pool entries always log SKIP — meaning the added coverage never actually runs in a normal `smoke-all.sh` invocation.

Suggest closing as superseded by #2268.

---

PR #2269, opened 2026-06-18
fix(hub-client): accept Windows absolute paths in git URL validator

heyong4725:
## Problem

The nightly **Test (windows-latest)** job has been red because 5 `dora-hub-client` `transport::tests` panic on Windows:

```
invalid git URL `C:\Users\RUNNER~1\AppData\Local\Temp\.tmpOmd9fE`:
  expected a https/ssh/file URL or absolute path
```

## Root cause

`validate_git_url` (`lib.rs`) documents that **absolute paths** are accepted for local repos/mirrors, but its check only matched `/`-rooted Unix paths:

```rust
let valid = SCHEMES.iter().any(|s| url.starts_with(s)) || url.starts_with('/');
```

So a Windows absolute path (`C:\repo`) is rejected — both the transport tests' temp-dir fixtures **and** a real-world `git = 'C:\repo'` local mirror in `hub.toml`. Green on Linux/macOS, red only on the Windows nightly runner. This is a **product** gap, not just a test bug.

## Fix

Accept Windows drive-letter absolute paths (`C:\…` / `C:/…`) via a shared `is_local_path` helper, and route `validate_git_url_untrusted`'s local-source detection through the **same** helper.

The second part is load-bearing security: without it, accepting `C:\…` in the trusted validator while the untrusted gate still only looked for `file://`/`/` would let a hostile public index entry point `source.git` at a victim-local Windows path and bypass the `DORA_HUB_ALLOW_LOCAL_SOURCES` opt-in. UNC paths (`\\host\share`), drive-relative `C:repo`, and relative paths stay rejected.

The test fixtures stay as plain temp-dir paths (now valid on every platform) — no `file://` workaround needed, and the transport tests now exercise the Windows-absolute-path branch end-to-end on the nightly Windows runner.

## Verification

- `cargo test -p dora-hub-client --lib` → 39 passed (incl. new `accepts_windows_absolute_paths` + extended untrusted-gate test, and the 5 transport tests)
- `cargo fmt -p dora-hub-client -- --check` → clean
- `cargo clippy -p dora-hub-client --all-targets -- -D warnings` → clean
- Windows runtime path traced by hand (validator now accepts `C:\…`; `git clone` handles local Windows paths); confirmation lands with the next nightly Windows job.

## Notes

Distinct from #2152 (the macOS post-stop shutdown hang) — a recent automated comment on #2152 misfiled this Windows failure there.

### DIFF
--- a/binaries/cli/src/command/build/hub.rs
+++ b/binaries/cli/src/command/build/hub.rs
@@ -61,10 +61,13 @@ impl HubResolution {
 
 /// Normalize an index `source.git` into a form the build's `GitManager` can
 /// consume. `validate_git_url` accepts two scheme-less forms — scp-style
-/// `git@host:path` and absolute local paths — but `GitManager::choose_clone_dir`
-/// parses the URL with `url::Url`, which rejects both. Rewrite them into the
-/// equivalent `ssh://` / `file://` URLs git treats identically, so a source
-/// that resolves also clones instead of failing late during the build.
+/// `git@host:path` and absolute local paths (Unix `/repo` and Windows
+/// `C:\repo`) — but `GitManager::choose_clone_dir` parses the URL with
+/// `url::Url`, which rejects `/repo` and misreads `C:\repo` as a `c:` URL.
+/// Rewrite them into the equivalent `ssh://` / `file://` URLs git treats
+/// identically, so a source that resolves also clones instead of failing late
+/// during the build. The local-path rewrite is shared with the validators so
+/// every path they accept is one the build can normalize.
 fn normalize_git_source_url(url: &str) -> String {
     if url.contains("://") {
         return url.to_string();
@@ -74,8 +77,8 @@ fn normalize_git_source_url(url: &str) -> String {
     {
         return format!("ssh://git@{host}/{}", path.trim_start_matches('/'));
     }
-    if url.starts_with('/') {
-        return format!("file://{url}");
+    if let Some(file_url) = dora_hub_client::local_path_to_file_url(url) {
+        return file_url;
     }
     url.to_string()
 }
@@ -740,6 +743,10 @@ mod tests {
                 "ssh://git@github.com/org/repo.git",
             ),
             ("/srv/mirrors/repo", "file:///srv/mirrors/repo"),
+            // Windows absolute paths → file:///C:/... (url::Url would read
+            // `C:\repo` as a `c:` scheme and drop the drive from `C:/repo`)
+            ("C:\\mirrors\\repo", "file:///C:/mirrors/repo"),
+            ("C:/mirrors/repo", "file:///C:/mirrors/repo"),
             // already-parseable forms are left untouched
             (
                 "https://github.com/org/repo.git",
--- a/libraries/hub-client/src/lib.rs
+++ b/libraries/hub-client/src/lib.rs
@@ -37,6 +37,26 @@ fn is_local_path(url: &str) -> bool {
     b.len() >= 3 && b[0].is_ascii_alphabetic() && b[1] == b':' && (b[2] == b'/' || b[2] == b'\\')
 }
 
+/// Rewrite a local-filesystem repo path into its `file://` URL, or `None` for
+/// anything that is not a local path (a remote/scp URL, a relative path).
+///
+/// `git` accepts a bare path directly, but consumers that parse the source with
+/// `url::Url` (the CLI build's `GitManager`) read `C:\repo` as a `c:` URL and
+/// reject `/repo` outright. Sharing this with [`is_local_path`] keeps the set of
+/// paths the validators accept and the set the build can normalize identical.
+pub fn local_path_to_file_url(url: &str) -> Option<String> {
+    if !is_local_path(url) {
+        return None;
+    }
+    let slashed = url.replace('\\', "/");
+    // Unix `/srv/repo` -> `file:///srv/repo`; Windows `C:/repo` -> `file:///C:/repo`.
+    Some(if slashed.starts_with('/') {
+        format!("file://{slashed}")
+    } else {
+        format!("file:///{slashed}")
+    })
+}
+
 /// Validate a git URL before it is ever passed to a `git` subprocess.
 ///
 /// URLs come from untrusted places (a public index entry's `source.git`, a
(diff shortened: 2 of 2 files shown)


phil-opp:

No issues found. The diagnosis matches the validator: `validate_git_url` (`libraries/hub-client/src/lib.rs`) accepts `file://` unconditionally but its bare-path branch only matches a leading `/`, so a Windows `C:\…` fixture path was the sole reason the five `transport::tests` failed on the nightly Windows runner. Routing `remote_index` through the new `file_url` helper fixes it on every platform.

The helper itself is correct: backslash→slash normalization plus the leading-`/` discriminator yields `file:///tmp/…` on Unix and `file:///C:/…` on Windows, both valid for `validate_git_url` and `git clone`. I also checked the validator's leading-`-` host defense doesn't trip on these — for `file:///…` the authority component is empty, so the host check is a no-op. Test-only change with no production impact.

Minor (non-blocking): `file_url` does no percent-encoding, so a temp path containing a space or `#`/`?` would produce a malformed URL — but `tempdir()` paths don't contain those, so it can't bite the current fixtures.

---

heyong4725:
Reworked per the review finding. Rather than route the test fixtures through `file://` (which would have masked the real gap), this now **fixes the validator**: `validate_git_url` accepts Windows drive-letter absolute paths, so a real `git = 'C:\repo'` local mirror in `hub.toml` works — not just the tests.

Critically, `validate_git_url_untrusted`'s `is_local` check now goes through the same `is_local_path` helper. Without that, accepting `C:\…` in the trusted validator would have opened a Windows-only supply-chain bypass (a hostile public index entry pointing `source.git` at a victim-local path, dodging the `DORA_HUB_ALLOW_LOCAL_SOURCES` gate). UNC / drive-relative / relative paths stay rejected.

Test fixtures reverted to plain temp-dir paths (now valid cross-platform), so the transport tests exercise the new branch directly. Added unit coverage for accepted/rejected forms + the untrusted gate. Net change is `lib.rs` only.

phil-opp:

Re-reviewed the reworked diff directly (the earlier review was against the now-discarded `file_url` test-fixture approach). No issues found.

The new `is_local_path` helper is correct — it matches Unix `/…` and Windows drive-letter `C:\…` / `C:/…` while rejecting drive-relative `C:repo` (no separator), UNC `\\host\share` (first byte isn't alphabetic), and relative paths. I confirmed there are no false positives against the remote schemes: `https://`, `ssh://`, `git@…` all fail the `b[1] == b':'` test.

The load-bearing security point holds: routing `validate_git_url_untrusted`'s `is_local` check through the same helper means a Windows path the trusted validator now accepts also trips the untrusted local-source gate — so accepting `C:\…` can't open a Windows-only bypass of `DORA_HUB_ALLOW_LOCAL_SOURCES`. Windows paths also pass cleanly through the later leading-dash host check (they hit the `else` branch with an empty host), which `accepts_windows_absolute_paths` asserts end-to-end. Unit coverage for the accepted/rejected forms and the untrusted gate looks complete.

---

heyong4725:
Good catch — fixed in `b4b771ec3`.

Accepting `C:\repo` in the validator was only half the path: the CLI desugar step (`normalize_git_source_url`) still handed scheme-less paths to `GitManager`, which parses with `url::Url` — `C:\repo` reads as a `c:` URL (no path segments → clone-dir selection fails) and `C:/repo` drops the drive. So a resolved Windows local source failed late during the build.

As suggested, the path→`file://` rewrite is now a **shared** helper, `dora_hub_client::local_path_to_file_url`, built on the same `is_local_path` predicate the validators use — so every path the validators accept is one the build can normalize (they can't drift). The CLI's `normalize_git_source_url` calls it in place of the old Unix-only `/`-prefix branch:

- `C:\repo` → `file:///C:/repo`
- `C:/repo` → `file:///C:/repo`
- `/srv/repo` → `file:///srv/repo` (unchanged)

Coverage added: `local_paths_rewrite_to_file_urls` (hub-client) and the two Windows cases in the CLI's `normalizes_scp_and_path_sources_to_parseable_urls`.

Verification: `cargo test -p dora-hub-client --lib` (40 passed), `cargo test -p dora-cli normalizes_scp_and_path_sources_to_parseable_urls` (passed), fmt + clippy (`-D warnings`) clean on both crates.

heyong4725:
Fixed in `a8832349b`.

`local_path_to_file_url` now percent-encodes the path instead of interpolating it raw. A legitimate `/tmp/a#b` or `C:\a?b` was being dropped into the `file://` URL verbatim, so a later `url::Url` re-parse in the build read `#`/`?` as a fragment/query — truncating the clone-dir identity and risking a collision between two distinct repos. (As noted, this predated the shared helper — it was in the old Unix-only `/`-branch too — but the shared helper is the right place to fix it.)

The encoder keeps RFC 3986 `pchar` + the `/` separator and percent-encodes everything else (including an existing `%`, so a literal path round-trips rather than mis-decoding):

- `/tmp/a#b` → `file:///tmp/a%23b`
- `C:\a?b` → `file:///C:/a%3Fb`
- `/tmp/a b` → `file:///tmp/a%20b`
- `/tmp/a%20b` → `file:///tmp/a%2520b`

Hand-rolled (~15 lines) rather than pulling in `percent-encoding`/`url`, to keep hub-client's dependency surface minimal and the helper platform-independent (so the Windows cases stay unit-testable on Linux CI).

Coverage added to `local_paths_rewrite_to_file_urls`. Verification: `cargo test -p dora-hub-client --lib` (40 passed), CLI normalizer test still green, fmt + clippy (`-D warnings`) clean.

heyong4725:
Both fixed (folded into the encoding commit, now `2c4d13548`).

**Unwrap budget:** the two `char::from_digit(…).unwrap()` calls in `percent_encode_path` were the only new production unwraps. Replaced with a static hex table indexed by a nibble (`0..=15`), which can't panic — no unwrap, no `expect`:

```rust
const HEX: &[u8; 16] = b"0123456789ABCDEF";
out.push('%');
out.push(HEX[(b >> 4) as usize] as char);
out.push(HEX[(b & 0x0f) as usize] as char);
```

Count is back to 163/163.

**Format:** ran `cargo fmt`; the helper is now rustfmt-clean.

My fault on both — when I "verified" fmt earlier I'd piped through `tail` (`cargo fmt … --check | tail -3 && echo OK`), so the `&&` checked `tail`'s exit code, not rustfmt's, and the failure slipped through. Re-verified with exit codes this time:

- `cargo fmt -p dora-hub-client -p dora-cli -- --check` → exit 0
- `scripts/qa/unwrap-budget.sh` → `163 (budget: 163)`, exit 0
- `cargo test -p dora-hub-client --lib` → 40 passed; clippy `-D warnings` clean

Issue #2270, opened 2026-06-18
daemon: bound the post-finish shutdown — escalate to SIGKILL for a node stuck outside the AllInputsClosed drain (#2152)


heyong4725:
## Summary

The daemon has a post-finish escalation ladder (`check_finish_stragglers` → Stop → SIGTERM → SIGKILL, gated on `finish_drain_grace()` / `DORA_FINISH_DRAIN_GRACE_SECS`), added so the natural-finish path no longer hangs `dora run` on a stuck node. But the macOS nightly hang in #2152 still occurs, which means **the escalation does not cover this shutdown-hang variant**. This issue tracks closing that gap so a single stuck node always converts into a *bounded* shutdown instead of an indefinite hang — which also bounds the damage from any future variant of "a node fails to exit after stop."

Split out of #2152 (the flake report) as the actionable daemon-side fix; #2152 keeps tracking the separate question of *why* the node fails to exit.

## Current behavior

`check_finish_stragglers` (`binaries/daemon/src/lib.rs:2347`) escalates a node only once it appears in `dataflow.finish_stragglers(grace)`, which is keyed on `all_inputs_closed_at` — i.e. the node must have entered the **AllInputsClosed drain** state and stayed there longer than the grace period.

In the #2152 repro the dataflow finishes its work, but one node (e.g. the operator `runtime-node`) never reports stop. The daemon's per-node TCP receive loop then spins on:

```
WARN dora_daemon::node_communication: failed to receive DaemonRequest
  Caused by: unexpected I/O error ... TCP read header timed out
  Location: binaries/daemon/src/node_communication/tcp.rs:78:26
(repeats every ~30s until the CI step's 15-min timeout)
```

No escalation fires. The likely gap: the stuck node never registered an `all_inputs_closed_at` timestamp (it's wedged in/around the TCP receive path, or the operator-runtime shutdown path differs from a plain node's), so `finish_stragglers` never selects it and the deadline ladder is never armed.

## Proposed fix

Extend the post-finish deadline so it covers **any** node that has not exited within a grace period after the dataflow is otherwise complete — not just nodes observed in the AllInputsClosed drain state. Concretely:

- When all other nodes of a dataflow have finished, arm a deadline against the remaining node(s) regardless of whether an `all_inputs_closed_at` timestamp exists for them, and escalate Stop → SIGTERM → SIGKILL through the existing ladder on expiry.
[...]

PR #2271, opened 2026-06-18
fix(daemon): escalate a node that wedges before draining (#2270)

heyong4725:
Towards #2270 — **does not auto-close it** (the watchdog ships disabled by default; see Rollout). Supersedes the drain-clock alternative in #2272.

## Why the liveness approach (not drain-clock)

#2270's acceptance case is "all nodes finished except one **wedged node that never entered the AllInputsClosed drain**." The drain-clock alternative (#2272) only arms when a node's `open_inputs` actually empties, so it can't cover a node that never drains — structurally partial. The liveness approach escalates on **silence past grace** regardless of drain state, so it covers the outside-drain case the issue requires.

## What it does

The finish-straggler watchdog escalates (Stop → SIGTERM → SIGKILL, with a pre-kill stack sample) a node blocking an otherwise-finished dataflow. `select_finish_stragglers`:

- **dynamic** → skip
- **never-finishes** (source, or fed by a `Timer`/`Logs` input) → veto
- **drained** (`AllInputsClosed`) past grace → escalate
- **never drained but silent** past grace **and connected** → escalate (the #2270 wedge)
- never drained, silent < grace, or not connected → veto

## Edge cases handled (surfaced across both approaches' reviews)

- **sources / timer-log nodes** never escalate.
- **slow-start nodes**: gated on a sticky **`connected_nodes`** signal (set on subscribe).
- **dropped-stream nodes**: `connected_nodes` is not tied to current channel presence, so a node that dropped its stream but stayed alive is still a candidate (SIGKILL signals the process).
- **node-ID reuse / restart**: `connected_nodes` is cleared on `RemoveNode`, on `restart_single_node`, and on node exit, so a re-added/respawned ID must re-subscribe before counting as connected.

## Rollout (why this leaves #2270 open)

This escalates to SIGKILL on a shutdown-only path that **isn't reproducible locally**, so it ships **dark**: gated on `DORA_FINISH_DRAIN_GRACE_SECS` (unset → disabled, current default). Plan:

1. **This PR**: land the mechanism, disabled by default. No behavior change for anyone.
2. **Validate**: enable it on nightly (`DORA_FINISH_DRAIN_GRACE_SECS=120` in the nightly jobs that exercise the #2152/#2270 shutdown path) and confirm it bounds the hang without false-positive kills over several runs.
3. **Flip the default**: once proven, make unset → enabled (one-line change), after which #2270 can be closed.

So #2270 stays open until step 3. Step 2 is a separate, reviewable change (it turns a force-kill heuristic on in CI), not bundled here.

## Tests

Pure-core selector cases + daemon-level: timer-fed spared, wedged-user escalated, unconnected slow-start spared, dropped-stream escalated, reused-ID spared pre-subscribe, restart clears the connected marker, opt-in parse.

`cargo test -p dora-daemon --lib` → 77 passed; fmt + clippy (`-D warnings`) clean.

Companion to #2152 (root cause of *why* a node fails to exit stays separately tracked).

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2156,6 +2156,10 @@ impl Daemon {
                     dataflow.subscribe_channels.remove(&node_id);
                     dataflow.pending_messages.remove(&node_id);
                     dataflow.all_inputs_closed_at.remove(&node_id);
+                    // clear the connected marker too, else a re-added node ID
+                    // would look already-connected before its new incarnation
+                    // subscribes and could be selected mid-startup (dora#2270).
+                    dataflow.connected_nodes.remove(&node_id);
                     dataflow.finish_escalated.remove(&node_id);
 
                     // Remove from stored descriptor (inverse of AddNode
@@ -2345,7 +2349,11 @@ impl Daemon {
     /// ladder used by explicit stops — after capturing a stack sample of
     /// the stuck process so the hang itself stays diagnosable.
     fn check_finish_stragglers(&mut self) {
-        let grace = finish_drain_grace();
+        // Opt-in: with DORA_FINISH_DRAIN_GRACE_SECS unset the watchdog is
+        // disabled and never escalates (ships dark; see `finish_drain_grace`).
+        let Some(grace) = finish_drain_grace() else {
+            return;
+        };
         let now_millis = node_communication::current_millis();
         for (dataflow_id, dataflow) in self.running.iter_mut() {
             for node_id in dataflow.finish_stragglers(grace, now_millis) {
@@ -3882,6 +3890,10 @@ impl Daemon {
         event_sender: mpsc::Sender<Timestamped<NodeEvent>>,
         clock: &HLC,
     ) {
+        // record that this node has connected — it stays a finish-straggler
+        // candidate even if it later drops its event stream (dora#2270).
+        dataflow.connected_nodes.insert(node_id.clone());
+
         // some inputs might have been closed already -> report those events
         let closed_inputs = dataflow
             .mappings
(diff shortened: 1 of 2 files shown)


heyong4725:
Good catch — fixed in `e17918ac8`. This was a real way to kill a **live** dataflow, not just a missed escalation.

## Root cause

My `never_finishes` veto was keyed on the descriptor source check alone (`inputs.is_empty()`). But a node fed by a `Timer` or `Logs` input is **not** a source and **also never drains**:
- every input — including `Timer`/`Logs` — is inserted into `open_inputs` (`lib.rs:1993`), and virtual inputs never close, so `open_inputs` never empties → `AllInputsClosed` is never sent → `drained_for = None` forever;
- timer/log delivery is daemon→node, so it never refreshes `last_activity`.

So a long-running timer-only side-effect node (your `examples/cpu-affinity-probe` fixture — `tick: dora/timer/millis/100`, stays alive until explicit stop, emits no outputs) looks "silent + never drained" exactly like a wedge, and the new silent-before-drain path would force-kill it after the 120s grace. The watchdog runs every health interval over all running dataflows, so yes — it could stop a live dataflow.

## Fix

Widen the veto from "is a source" to "never naturally finishes":

```
never_finishes = descriptor source (no inputs)
              || fed by a Timer input   (self.timers)
              || fed by a Logs input    (self.log_subscribers)
```

I read this from the daemon's own runtime state (`timers` / `log_subscribers`) rather than re-walking descriptor mappings — it's authoritative for the running dataflow and avoids an `InputMapping` traversal. Such nodes are stopped only via the explicit-stop path; a wedged node with real (User) inputs is still escalated.

## Tests

- `timer_fed_node_is_not_a_finish_straggler` (daemon-level): a silent timer-fed node is **not** selected.
- `wedged_user_input_node_is_a_finish_straggler` (daemon-level): a silent user-input node **is** selected.
- renamed the pure-core flag `is_source` → `never_finishes`; the existing veto test now documents that it guards the timer/log case.

`cargo test -p dora-daemon --lib` → 71 passed; fmt + clippy (`-D warnings`) clean.

heyong4725:
Fixed in `a50763de3` — real regression, good catch.

## Root cause

My silent arm assumed `last_activity == 0` meant "never connected", but the spawner seeds it to `current_millis()` at spawn (`spawn/spawner.rs:149`, carried via `prepared.rs:133`). So the `last == 0` guard was dead in production, and a node's `silent_for` is measured **from spawn time**. A slow-starting node (model load) whose source exits before it subscribes records no `AllInputsClosed` timestamp (`drained_for = None`), so the silent arm could force-kill it after the grace period.

## Fix

Gate the silent arm on a genuine connection signal instead of `last_activity`: the node must be in `subscribe_channels`. That map is populated in the **same subscribe handler** (`lib.rs:3942`) that sends `AllInputsClosed` (`lib.rs:3913`) and is the channel finish events travel — so membership is exactly "has connected and can receive finish events." An unconnected node is treated as not-yet-finished (veto), so it's never escalated while starting up. Dropped the now-dead `last_activity == 0` special case.

This preserves the #2152 catch: the wedged runtime node *had* subscribed (it was running and receiving events), so it's still eligible.

## Tests

- `unconnected_node_silent_past_grace_is_not_escalated` (pure core) and `unconnected_slow_starting_node_is_not_a_finish_straggler` (daemon-level): a silent-but-unsubscribed node is **not** escalated.
- `wedged_user_input_node_is_a_finish_straggler` now subscribes its node (connected) and is still escalated; the `insert_silent_node` helper marks nodes connected.

`cargo test -p dora-daemon --lib` → 73 passed; fmt + clippy (`-D warnings`) clean.

---

Reflection: this is the third review round to find a correctness gap in this change (escalation gap → timer/log false-positive → slow-start false-positive). The common thread is that "node is silent/idle" is a deceptively weak signal in this subsystem — sources, timer/log nodes, and not-yet-connected nodes all look idle for legitimate reasons. The gate is now: escalate only a **connected, finishing (non-source, no timer/log), non-dynamic** node that is **drained past grace** or **silent past grace**. If reviewers see a fourth "looks idle but is fine" class I've missed, I'd rather hear it now than ship it.

heyong4725:
Superseded by #2272. After prototyping the drain-clock alternative for comparison, we're going with it: it's a ~5x smaller diff (+65/−11 vs +323/−65), leaves the finish-straggler selector untouched, and is structurally immune to the idle-node false-positives this approach kept surfacing (sources, timer/log nodes, slow-start nodes — three review rounds). It arms the existing drain clock when the AllInputsClosed delivery fails (the plausible #2152 channel-stalled trigger) instead of escalating on raw silence. Thanks for the three rounds of review here — each one directly informed the decision to switch. Closing in favor of #2272.

heyong4725:
Reopened and updated. Two reviews on #2272 (the drain-clock alternative) established that it can't satisfy #2270's acceptance case — a node that never enters the AllInputsClosed drain is never armed, so it's a structurally partial fix. The liveness approach covers that case (escalates on silence). Carried the lifecycle hardening from the drain-clock rounds back here — crucially `connected_nodes` (sticky, cleared on removal) instead of `subscribe_channels` presence, which closes the dropped-stream and ID-reuse holes that *both* approaches shared — plus the opt-in safety valve so it ships dark. `41da60d18`.

heyong4725:
Both High findings addressed.

**Restart path (High #2):** `connected_nodes` is now cleared on the restart/respawn reset paths too — `restart_single_node` and the node-exit handler — alongside the existing per-incarnation drain-state reset. So a respawned node must re-subscribe before counting as connected; a slow restart/model-reload can't be silence-escalated mid-startup. Added `restart_clears_connected_marker` (drives `restart_single_node` directly). `7009a1d4c`.

**Fixes/ship-dark (High #1):** changed the PR so it no longer auto-closes #2270 (verified: closingIssuesReferences = 0), and documented the 3-step rollout (land dark → validate on nightly → flip default, then resolve #2270) both in the PR body and on #2270. The issue stays open until the default is flipped.

`cargo test -p dora-daemon --lib` → 77 passed; fmt + clippy clean.

heyong4725:
Both Mediums fixed in `c2a92123b`.

**Timer/log state on removal:** `RemoveNode` now drops the node's `timers` and `log_subscribers` entries (retain-out, like `mappings` already are). Two benefits: a removed node stops receiving ticks/logs, and `node_never_finishes` — which reads that state — no longer marks a re-added user-input node under a recycled ID as never-finishing. I went with clearing the runtime state (rather than deriving from the descriptor) because not clearing it was an independent leak anyway, and it keeps `never_finishes` consistent with how the inputs are actually registered.

**AddMapping reopen:** clears `all_inputs_closed_at` when re-adding an input, so the selector won't treat a node as drained-and-eligible on a timestamp from before the mapping came back. This was the one fix from the closed drain-clock branch that hadn't been carried over to here — good catch.

Tests: `cleared_timer_state_makes_reused_id_escalatable` (timer node vetoed → escalatable once its timer state is cleared) and `reopened_input_clears_stale_drain_timestamp` (active node with reopened input not selected on a stale drain).

`cargo test -p dora-daemon --lib` → 79 passed; fmt + clippy (`-D warnings`) clean.

---

These were both the same class as the last round — stale per-node state surviving an ID-reuse. The per-node state touched by the watchdog is now cleared at every teardown point I can enumerate: RemoveNode (running_nodes, open_inputs, subscribe_channels, pending_messages, all_inputs_closed_at, connected_nodes, finish_escalated, **timers, log_subscribers**, descriptor), restart_single_node, and the node-exit handler — plus the reopen-clear on AddMapping. If a reviewer can still point at a per-node map the watchdog reads that isn't torn down on one of those transitions, that's the remaining gap; I'm not aware of one now, but this PR's history argues for verifying rather than trusting that.

PR #2272, opened 2026-06-19
fix(daemon): arm the drain clock on input-close, even if delivery fails (#2270)

### LINKED ISSUE #2270: daemon: bound the post-finish shutdown — escalate to SIGKILL for a node stuck outside the AllInputsClosed drain (#2152)

heyong4725:
Fixes #2270. Supersedes #2271 (the liveness approach) after a side-by-side comparison — see the note on #2271.

## Problem

The finish-straggler watchdog escalates a node (Stop → SIGTERM → SIGKILL) once its drain clock `all_inputs_closed_at` exceeds the grace period. But the clock was only armed when the `AllInputsClosed` **delivery** returned `Some(true)`:

```rust
if send_with_timestamp(channel, NodeEvent::AllInputsClosed, clock).ok() == Some(true) {
    dataflow.inc_pending(receiver_id);
    dataflow.all_inputs_closed_at.insert(receiver_id.clone(), Instant::now());
}
```

A node that wedges and stops draining its event channel makes that send fail (full/closed channel) → no timestamp → the watchdog never arms → the dataflow hangs until an external timeout (#2152).

## Fix

Arm the clock whenever the daemon determines a node's inputs are closed, **independent of whether the AllInputsClosed send succeeded** — at all three sites (subscribe handler, `close_input`, `break_input`), via `entry(...).or_insert_with(Instant::now)` (monotonic, idempotent).

The straggler **selector is unchanged from main** (drained-only). This is the key advantage over the liveness alternative: because arming stays gated on the node being subscribed and on `open_inputs` actually emptying, it is correct by construction for the cases a silence heuristic had to special-case:

- **sources** and **timer/log-fed nodes** never empty `open_inputs` → never armed → never escalated;
- a **not-yet-subscribed slow-starting node** (e.g. model load) has no channel, so nothing is recorded until it connects.

## Why this approach over #2271

| | This (drain-clock) | #2271 (liveness) |
|---|---|---|
| Diff vs main | +65/−11, 1 file | +323/−65, 2 files |
| Straggler selector | unchanged | rewritten |
| Idle-node edge cases | none (structural) | sources / timer-log / slow-start — 3 review rounds |
| Signal | "daemon knows inputs closed" | "node looks idle" |

Trade-off: this catches a wedge whose `open_inputs` emptied but didn't exit (incl. the channel-stalled case above). It does **not** catch a node whose `open_inputs` never empties (a phantom open input) — if that turns out to be a real #2152 variant, the right fix is in input-close propagation, not a silence heuristic.

## Tests

`drain_clock_arms_when_all_inputs_closed_send_fails` — with the receiver dropped (every send fails), `close_input` still arms `all_inputs_closed_at`. Plus an added assertion on the existing happy-path test.

`cargo test -p dora-daemon --lib` → 66 passed; fmt + clippy (`-D warnings`) clean.

Companion to #2152 (root-cause of *why* a node fails to exit remains separately tracked).

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -3916,13 +3916,22 @@ impl Daemon {
             if let Some(node) = dataflow.descriptor.nodes.iter().find(|n| n.id == node_id) {
                 if node.inputs.is_empty() {
                     // do not send AllInputsClosed for source nodes
-                } else if send_with_timestamp(&event_sender, NodeEvent::AllInputsClosed, clock).ok()
-                    == Some(true)
-                {
-                    dataflow.inc_pending(&node_id);
+                } else {
+                    if send_with_timestamp(&event_sender, NodeEvent::AllInputsClosed, clock).ok()
+                        == Some(true)
+                    {
+                        dataflow.inc_pending(&node_id);
+                    }
+                    // Start the drain clock once the daemon knows the inputs are
+                    // closed, even if the AllInputsClosed send was dropped (e.g.
+                    // a full event channel on a node that has stopped draining).
+                    // Without this the finish-straggler watchdog never arms and
+                    // the node hangs the dataflow (dora#2152). `or_insert` keeps
+                    // the clock monotonic across repeated calls.
                     dataflow
                         .all_inputs_closed_at
-                        .insert(node_id.clone(), Instant::now());
+                        .entry(node_id.clone())
+                        .or_insert_with(Instant::now);
                 }
             }
         }
@@ -5244,10 +5253,16 @@ fn close_input(
             }
             if send_with_timestamp(channel, NodeEvent::AllInputsClosed, clock).ok() == Some(true) {
                 dataflow.inc_pending(receiver_id);
-                dataflow
-                    .all_inputs_closed_at
-                    .insert(receiver_id.clone(), Instant::now());
             }
+            // Start the drain clock once the daemon knows the inputs are closed,
+            // even if the AllInputsClosed send was dropped (e.g. a full event
+            // channel on a node that stopped draining) — otherwise the finish-
+            // straggler watchdog never arms and the node hangs the dataflow
+            // (dora#2152). `or_insert` keeps the clock monotonic.
+            dataflow
+                .all_inputs_closed_at
+                .entry(receiver_id.clone())
+                .or_insert_with(Instant::now);
         }
     }
 }
@@ -5289,10 +5304,16 @@ fn break_input(
             }
             if send_with_timestamp(channel, NodeEvent::AllInputsClosed, clock).ok() == Some(true) {
                 dataflow.inc_pending(receiver_id);
-                dataflow
-                    .all_inputs_closed_at
-                    .insert(receiver_id.clone(), Instant::now());
             }
+            // Start the drain clock once the daemon knows the inputs are closed,
+            // even if the AllInputsClosed send was dropped (e.g. a full event
+            // channel on a node that stopped draining) — otherwise the finish-
+            // straggler watchdog never arms and the node hangs the dataflow
+            // (dora#2152). `or_insert` keeps the clock monotonic.
+            dataflow
+                .all_inputs_closed_at
+                .entry(receiver_id.clone())
+                .or_insert_with(Instant::now);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


heyong4725:
Both fixed in `812b2997d`.

## High — dropped-stream nodes now arm

The all-inputs-closed handling (including arming the drain clock) lived inside `if let Some(channel) = subscribe_channels.get(..)`. So a node that connected and then **dropped its event stream** (`EventStreamDropped` removes the channel) but kept its process alive would close its last input with no timestamp recorded → the watchdog vetoes forever (`running_dataflow.rs` `None => return`). That left the exact #2270 class — "stuck outside AllInputsClosed drain" — unbounded.

Worth noting: the **closed liveness PR had the identical hole** via `connected = subscribe_channels.contains(..)`. The real fix, for either approach, is a true *has-ever-connected* signal rather than *currently-has-a-channel*.

Fix:
- New `connected_nodes: BTreeSet<NodeId>` on `RunningDataflow`, set in `subscribe()`, never tied to the current channel.
- Pulled the duplicated `close_input`/`break_input` tail into `arm_all_inputs_closed`, which arms the clock for a **connected** node **regardless of channel** (escalation SIGKILLs the process, not the channel). InputClosed/AllInputsClosed delivery stays gated on the channel; only the clock is decoupled.
- A node that has **never** connected (slow startup, inputs close before it subscribes) is still not armed — it arms later, at subscribe time. So the slow-start protection holds.

## Medium — reopen clears the clock

`or_insert_with` preserved a stale timestamp when `AddMapping` re-adds an input to `open_inputs`. Now `all_inputs_closed_at` is cleared on reopen, so a later re-close arms a fresh grace period instead of reusing the old deadline (and a node with reopened inputs isn't seen as still-draining in the interim).

## Tests
- `drain_clock_arms_for_connected_node_after_stream_drop` — connected node, no channel → arms.
- `drain_clock_not_armed_for_node_that_never_connected` — never connected → does not arm.
- `reopened_input_clears_then_rearms_drain_clock` — reopen clears, re-close re-arms.
- updated the happy-path/send-fails tests to mark nodes connected.

`cargo test -p dora-daemon --lib` → 69 passed; fmt + clippy (`-D warnings`) clean.

This is the round that, in hindsight, validates the drain-clock pivot: it's the same `connected_nodes` signal the liveness approach would have needed too, but here it gates a single arming site instead of threading through the selector — and the selector stays exactly as on main.

heyong4725:
Fixed in `2e91b4a37`.

`connected_nodes` is now cleared in the `RemoveNode` cleanup (`lib.rs:2153`) alongside the other per-node state (`running_nodes`, `open_inputs`, `subscribe_channels`, `all_inputs_closed_at`, `finish_escalated`). A re-added node ID starts a fresh incarnation, so its slow-starting new process isn't treated as already-connected before it subscribes — closing the reused-ID slow-start regression. Corrected the field doc ("cleared when the node is removed") and added `removed_node_id_is_not_connected_on_reuse`.

Considered making the marker incarnation-specific (your alternative), but clearing on removal is the smaller change and matches how every other per-node map is already torn down at that exact site.

`cargo test -p dora-daemon --lib` → 70 passed; fmt + clippy (`-D warnings`) clean.

heyong4725:
## Safety valve: watchdog is now opt-in (`25ef05449`)

Given this escalates to SIGKILL on a shutdown-only path that can't be reproduced locally and took several rounds to classify correctly, the whole finish-straggler watchdog now ships **dark**:

- `DORA_FINISH_DRAIN_GRACE_SECS` **unset** → watchdog disabled, never escalates (current default).
- **set** to N → enabled at an N-second grace.
- set-but-garbage → enabled at the 120s default (the user clearly meant it on).

`finish_drain_grace()` now returns `Option<Duration>` (None = disabled) and `check_finish_stragglers` early-returns when disabled. Parsing is split into a pure `parse_finish_drain_grace` with unit tests (`finish_drain_grace_is_opt_in`).

### Why
This lets the SIGKILL-on-stuck-node behaviour be turned on per-deployment and validated on real nightly workloads before it becomes default — rather than shipping a force-kill heuristic enabled-by-default on a path we can't exercise in CI. If it proves out on nightlies, flipping the default later is a one-line change.

The drain-clock arming itself (recording `all_inputs_closed_at`) stays unconditional and harmless — only the escalation/kill is gated. So if we want telemetry on how often the watchdog *would* fire, that's a cheap follow-up.

`cargo test -p dora-daemon --lib` → 71 passed; fmt + clippy (`-D warnings`) clean.

heyong4725:
Closing in favor of #2271 (reopened). Two review rounds here established that the drain-clock can't satisfy #2270's acceptance case: it only arms `all_inputs_closed_at` when a node's open_inputs actually empties, so a node stuck *outside* the AllInputsClosed drain — exactly what the issue targets — is never escalated. The liveness approach covers it. The hardening developed on this branch (connected_nodes lifecycle, opt-in safety valve) has been carried over to #2271. Thanks — prototyping this was what made the coverage gap concrete.

PR #2277, opened 2026-06-19
fix(hub): surface yanked versions in `dora hub info` with (yanked) marker

### LINKED ISSUE #2275: hub: `dora hub info`'s `(yanked)` marker is unreachable dead code
> **Created by a scheduled automated Claude code-review check.** Please verify before acting — this is an automated finding, not a human report.

## Summary

`dora hub info` has a branch that prints `(yanked)`, but it can never execute: it inspects the entry returned by `IndexCatalog::resolve()`, and `resolve()` is defined to **skip yanked versions** and only ever return a non-yanked entry (or an error). So `resolved.entry.yanked` is always `false`, and `dora hub info` can never tell the user that a package's latest version was yanked.

## Where

`binaries/cli/src/command/hub/info.rs:24-33`

```rust
let resolved = catalog
    .resolve(&reference)
    .with_context(|| format!("failed to resolve `{}`", self.package))?;
ctx.drain_warnings();

let m = &resolved.entry.manifest;
println!("{} {}", reference.key(), resolved.version);
if resolved.entry.yanked {
    println!("  (yanked)");   // <-- unreachable: resolve() never returns a yanked entry
}
```

`IndexCatalog::resolve()` (`libraries/hub-client/src/index.rs:307`) walks matching versions highest-first and `continue`s past any entry with `yanked == true`, returning the first **non-yanked** match. Yanked entries are only ever returned by the direct `entry()` lookup (used by the locked-build path, which warns separately). Therefore the `if resolved.entry.yanked` condition in `info.rs` is dead code.

## Impact

Low severity (no incorrect output today), but:

- The dead branch signals an intended capability — telling users a version is yanked — that `info` silently does not provide.
- If a user passes an explicit yanked version (`dora hub info pkg@0.6.0` where `0.6.0` is yanked), `resolve()` reports it as unsatisfiable rather than showing it marked `(yanked)`, which is likely the opposite of what the author intended given this branch exists.

## Suggested fix

Either remove the unreachable branch, or — if the intent was for `info` to surface yanked status — look up the entry in a way that can return a yanked version (e.g. when the reference pins an exact version, fall back to `catalog.entry(...)` and mark it `(yanked)`), so the marker can actually fire. The `yank_reason` field is also available to display, mirroring the yank warning in `binaries/cli/src/command/build/hub.rs`.

phil-opp:
## Summary

Fixes #2275.

`resolve()` only returns non-yanked index entries, making the existing `if resolved.entry.yanked` branch in `info.rs` dead code. When a user explicitly requests a yanked version (`dora hub info pkg@0.6.0` where `0.6.0` is yanked), `resolve()` fails with an opaque "every version … has been yanked" error rather than showing the package info marked `(yanked)`.

**Fix:** when `resolve()` fails, fall back to the highest yanked version matching the requirement (via `catalog.versions()` + `catalog.entry()`). If a yanked match is found, display it with the `(yanked)` marker as originally intended. If no matching version exists at all, re-raise the original resolve error unchanged.

This makes the `(yanked)` branch reachable and aligns `dora hub info` with the user expectation implied by the branch: explicit version lookups surface yanked status instead of returning an error.

## Test plan

- [ ] `dora hub info pkg@<non-yanked>` still resolves and displays normally
- [ ] `dora hub info pkg@<yanked>` now shows the entry with `(yanked)` instead of an error
- [ ] `dora hub info pkg` (bare, all versions yanked) surfaces the yanked entry with `(yanked)`
- [ ] `dora hub info nonexistent` still returns an error (no yanked fallback found)
- [ ] `cargo clippy` and `cargo fmt` pass

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

https://claude.ai/code/session_01Gc578USnixf6Z8mycBu3wR

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gc578USnixf6Z8mycBu3wR)_

### DIFF
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -104,3 +104,6 @@ crate-type = ["lib", "cdylib"]
 [package.metadata.dist]
 # Enable dora-cli binary to be distributed
 dist = true
+
+[dev-dependencies]
+tempfile = { workspace = true }
--- a/binaries/cli/src/command/hub/info.rs
+++ b/binaries/cli/src/command/hub/info.rs
@@ -236,4 +236,26 @@ mod tests {
             .is_err()
         );
     }
+
+    // A corrupt/unreadable matching entry must surface its parse error rather
+    // than being skipped so a lower yanked entry is shown in its place
+    // (#2275 review): the highest matching version (0.7.0) is malformed.
+    #[test]
+    fn corrupt_matching_entry_is_surfaced_not_masked_by_yanked() {
+        let tmp = tempfile::tempdir().unwrap();
+        let yolo = tmp.path().join("dora-rs/dora-yolo");
+        std::fs::create_dir_all(&yolo).unwrap();
+        write_entry(&yolo, "0.6.0", true); // readable yanked
+        // higher version, but its entry fails to parse (manifest must be a map)
+        std::fs::write(yolo.join("0.7.0.yml"), "manifest: not-a-map\n").unwrap();
+        let catalog = IndexCatalog::open(tmp.path()).unwrap();
+
+        let err = resolve_or_yanked(&catalog, &parse("dora-rs/dora-yolo"), "dora-rs/dora-yolo")
+            .expect_err("a corrupt matching entry must surface, not fall back to yanked");
+        let msg = format!("{err:#}").to_lowercase();
+        assert!(
+            msg.contains("0.7.0") || msg.contains("invalid index entry"),
+            "expected the corrupt-entry error, got: {msg}"
+        );
+    }
 }
--- a/binaries/cli/src/command/hub/info.rs
+++ b/binaries/cli/src/command/hub/info.rs
@@ -96,21 +96,25 @@ fn resolve_or_yanked(
 )> {
     match catalog.resolve(reference) {
         Ok(resolved) => Ok((resolved.version, resolved.entry)),
-        Err(resolve_err) => catalog
+        Err(resolve_err) => {
             // `versions()` is sorted ascending, so `.rev()` yields highest-first.
-            .versions(&reference.namespace, &reference.name)
-            .unwrap_or_default()
-            .into_iter()
-            .rev()
-            .filter(|v| reference.requirement.matches(v))
-            .find_map(|v| {
-                catalog
-                    .entry(&reference.namespace, &reference.name, &v)
-                    .ok()
-                    .filter(|e| e.yanked)
-                    .map(|e| (v, e))
-            })
-            .ok_or_else(|| resolve_err.wrap_err(format!("failed to resolve `{package_label}`"))),
+            // Use `?` rather than `.ok()` so that corrupt entry files surface their
+            // parse error instead of being silently skipped — otherwise a malformed
+            // non-yanked entry could be masked by a lower valid yanked result.
+            for v in catalog
+                .versions(&reference.namespace, &reference.name)
+                .unwrap_or_default()
+                .into_iter()
+                .rev()
+                .filter(|v| reference.requirement.matches(v))
+            {
+                let entry = catalog.entry(&reference.namespace, &reference.name, &v)?;
+                if entry.yanked {
+                    return Ok((v, entry));
+                }
+            }
+            Err(resolve_err.wrap_err(format!("failed to resolve `{package_label}`")))
+        }
     }
 }
 
--- a/binaries/cli/src/command/hub/info.rs
+++ b/binaries/cli/src/command/hub/info.rs
@@ -3,7 +3,6 @@
 use super::common::{HubContext, sanitize};
 use crate::command::Executable;
 use dora_hub_client::reference::PackageRef;
-use eyre::Context;
 
 /// Show details of a hub node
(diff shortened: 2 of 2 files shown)


phil-opp:

Reviewed the yanked-fallback in `dora hub info`. No issues found. The fallback only triggers when `resolve()` fails, walks versions highest-first filtering on `requirement.matches(v)` + `e.yanked`, and re-raises the original resolve error (wrapped) when no matching version exists — so non-yanked lookups and genuine not-found cases are unchanged, while an explicitly-requested yanked version now surfaces with the `(yanked)` marker instead of an opaque error.

---

phil-opp:
> 🤖 This is a fully automated review by Claude. No human reviewed it before posting — please verify before acting on it.

I took an independent look at the yanked fallback. The happy path and the stated design (highest-first, re-raise the original error when no match) are correct. One substantive concern:

**The fallback can mask a genuine parse/corruption error by displaying a *different* yanked entry.**

`resolve()` walks matching versions highest-first and calls `self.entry(...)?` on each. So if the *highest* matching `<version>.yml` is malformed, `resolve()` fails with an "invalid index entry" parse error — not a yank error. The new fallback then does `catalog.entry(...).ok().filter(|e| e.yanked)`, and the `.ok()` silently swallows that parse error, skips the corrupt file, and returns a *lower* yanked version that happens to parse.

Concretely, with matching versions `[0.5.0 (yanked, valid), 0.6.0 (corrupt yaml)]`: `resolve()` errors on the `0.6.0` parse, then the fallback displays `pkg 0.5.0 (yanked)` and the real corruption of `0.6.0` is hidden from the user.

It's a narrow case (requires a malformed entry file alongside a yanked one) and `info` is read-only, so this isn't necessarily a blocker — but it's exactly the "mask a non-yank resolve error" footgun the fallback should avoid. A tighter version would only take the fallback when the resolve error is the "has been yanked" variant, or would not swallow `entry()` errors in the fallback. Flagging for your call.

---

heyong4725:
Added regression coverage in `e2d17a805` (the one gap from my review).

To make it testable without stdout capture, I extracted the resolve-or-yanked-fallback from `Info::execute` into a pure `resolve_or_yanked(catalog, reference, label) -> Result<(Version, IndexEntry)>` helper — **behavior-identical** (exact logic, same error message/wrap). Then 4 tests against an on-disk fixture catalog (`IndexCatalog::open` on a tempdir, mirroring the hub-client fixture):

- `surfaces_explicitly_requested_yanked_version` — `dora-yolo@0.6.0` (yanked) → returns it with `yanked=true`, **proving the `(yanked)` branch is now reachable** (the core #2275 regression).
- `bare_reference_resolves_highest_non_yanked` — bare `dora-yolo` still resolves to 0.5.2, skipping the yanked 0.6.0 (fallback doesn't disturb the happy path).
- `bare_reference_surfaces_yanked_when_all_yanked` — a package whose only version is yanked surfaces it.
- `errors_when_no_version_matches` — non-existent package still re-raises the resolve error.

Also added `tempfile` as a dora-cli dev-dependency. `cargo test -p dora-cli --lib command::hub::info` → 4 passed; fmt + clippy clean.

The fix itself reviewed clean (verified `versions()` sorts ascending so `.rev()` = highest, and `Report::wrap_err` is inherent so the `eyre::Context` import removal is correct).

phil-opp:
Good catch on the masked parse error — addressed in `ec283bc`.

The fix switches the fallback from the `find_map` + `.ok()` chain to an explicit `for` loop with `?`. When `catalog.entry()` returns a parse error, it now propagates immediately rather than being swallowed by `.ok()`. This means a corrupt non-yanked entry file surfaces its real error instead of being silently skipped in favour of a lower valid yanked result.

Behaviour unchanged for the cases the tests cover (all-yanked, highest-non-yanked, explicit yanked, not-found); the only difference is the corrupt-entry edge case now propagates the parse error, which matches what `resolve()` itself does (it also uses `?` at line 328).

---

heyong4725:
Re: the Medium (fallback masking a corrupt entry) — looks like we converged on it in parallel. Your `ec283bc3e` already fixed it (switched the fallback's `.ok()` to `?` so `entry()` parse errors propagate). 👍

I'd independently written the same fix locally; rather than push a competing implementation I dropped mine and instead added the **regression test it was missing** (`f58a5c187`): `corrupt_matching_entry_is_surfaced_not_masked_by_yanked` — the highest matching version (0.7.0) is malformed, and the test asserts the parse error surfaces instead of falling back to the readable yanked 0.6.0. All 5 `command::hub::info` tests pass; fmt + clippy (`-D warnings`) clean.

One non-blocking observation on the chosen shape: the loop `return`s on the first (highest) yanked, so a *corrupt entry that sits below a yanked match* (e.g. `[0.6.0 yanked, 0.5.0 corrupt]`) won't be read, and its error won't surface. For `info` that's arguably fine — it displays the highest version and there's no reason to read/validate lower ones it won't show — so I'd leave it as-is. Flagging only so it's a deliberate choice rather than an accident.

Net: with `ec283bc3e` + the test, the Medium is addressed and guarded. LGTM.

PR #2280, opened 2026-06-19
test(daemon): deterministic e2e for the finish-straggler watchdog (#2270)

heyong4725:
Towards #2270 (does not close it). Follow-up to #2271 (watchdog) and #2276 (nightly enablement).

## What

A deterministic end-to-end test for the finish-straggler watchdog, so we no longer depend on the flaky #2152 macOS hang actually occurring to validate it.

- New fixture `emit-then-exit-source-node`: emits one value, then exits.
- Reuses the existing `hang-after-init-node` (subscribes → drops its event stream → sleeps forever).
- `tests/dataflows/finish-straggler-escalation.yml` wires them: once `source` exits, the dataflow is "otherwise finished" except `wedge`, which never exits on its own.
- `tests/finish-straggler-e2e.rs` runs it in-process with `DORA_FINISH_DRAIN_GRACE_SECS=2` and asserts the watchdog bounds the hang.

## Why this is the deterministic version

The hang in #2152 is a ~2/3-of-runs macOS flake, so nightly only *sometimes* exercises the escalation. This forces the exact wedge condition every run: `source` exits, `wedge` stays alive and silent, and the watchdog must escalate it.

## Observed result

```
dataflow completed in 14.43s with 2 node results
  source: Ok(())
  wedge:  Err(NodeError { cause: GraceDuration, exit_status: Signal(15) })
```

The wedge is escalated and killed (~2s grace + ~10s kill ladder ≈ 14s) instead of hanging until the 40s `stop_after` net. The test asserts:
1. completion well under `stop_after` (a disabled/broken watchdog runs to 40s → fails the `< 30s` bound), and
2. the wedge exited via signal (SIGTERM/SIGKILL from the escalation ladder).

The `< 30s` bound is the load-bearing assertion: both the watchdog and `stop_after` produce a signal exit, so timing is what distinguishes "watchdog fired at ~14s" from "hung to the 40s net."

## Relationship to the #2270 rollout

This strengthens step 2→3: we now have a permanent, deterministic regression test for the escalation, rather than waiting to catch the nightly flake. (Default stays off; flipping it remains a separate step.)

Notes:
- Lives in its own test binary so its `DORA_FINISH_DRAIN_GRACE_SECS` set doesn't affect other tests' daemons.
- `cargo test --test finish-straggler-e2e` → 1 passed (~15s); fmt + clippy (`-D warnings`) clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -280,6 +280,15 @@ jobs:
           --test-threads=1
           lifecycle_rust lifecycle_cxx
         timeout-minutes: 15
+      # Finish-straggler watchdog regression (#2270): a deterministic wedge
+      # (source exits, downstream node hangs) must be escalated/killed by the
+      # watchdog rather than hanging until stop_after. Lives in `dora-examples`
+      # (excluded from the bulk `cargo test --all` step), so it needs an
+      # explicit invocation; the test sets DORA_FINISH_DRAIN_GRACE_SECS itself.
+      - name: Run finish-straggler watchdog E2E
+        run: >
+          cargo test -p dora-examples --test finish-straggler-e2e -- --test-threads=1
+        timeout-minutes: 10
 
   # Semantic contract tests — promoted from nightly per #1632.
   #
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2422,6 +2422,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "91622ff5e7162018101f2fea40d6ebf4a78bbe5a49736a2020649edf9693679e"
 
 [[package]]
+name = "emit-then-exit-source-node"
+version = "1.0.0-rc1"
+dependencies = [
+ "dora-node-api",
+ "eyre",
+]
+
+[[package]]
 name = "encode_unicode"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


heyong4725:
Both fixed in `2372d54d3`.

## High — the test wasn't actually run by CI

You're right, and it's worse than just the command: the bulk `cargo test --all` step **excludes `dora-examples`** (ci.yml line 147), and the e2e job invokes each e2e test **by name** (`cargo test -p dora-examples --test <name>`). My test was neither — so it would never have run in CI. (The literal `--test finish-straggler-e2e` resolves fine for me via auto-discovery, but that's beside the point — CI wouldn't have invoked it.)

Fixed by matching the existing pattern used for `daemon-reconnect-e2e` / `node-lifecycle-e2e`:
- added an explicit `[[test]]` entry (unambiguous target resolution), and
- added an explicit step to the ci.yml `e2e` job: `cargo test -p dora-examples --test finish-straggler-e2e -- --test-threads=1`.

## Medium — unsound `set_var` under a multi-thread runtime

Correct — `#[tokio::test(flavor = "multi_thread")]` means the runtime threads already exist when the body runs, so the `unsafe set_var` SAFETY claim was invalid. Switched to a plain `#[test]` that sets the env var **first, while the process is still single-threaded** (only `cargo build` child processes are spawned before it, which don't share our address space), then builds the runtime and `block_on`s the body.

## Verified
- `cargo test -p dora-examples --test finish-straggler-e2e -- --test-threads=1` (the exact CI invocation) → 1 passed (~17s)
- fmt + clippy (`-D warnings`) clean; ci.yml YAML validated

Now it's real, CI-enforced coverage for the watchdog rather than a test that silently never ran.

PR #2285, opened 2026-06-20
feat(daemon): add per-node finish_grace_secs to prevent false-positive SIGKILL of long-running terminal nodes

### LINKED ISSUE #2284: daemon: default-on finish-straggler watchdog can SIGKILL a node doing legitimate long (>120s) post-input compute (e.g. terminal ML/training stage)
> **This issue was created by a scheduled automated Claude review** of the last ~2 days of merged commits/PRs. It is a design/correctness concern, not a confirmed crash — please sanity-check the reasoning before acting.

## Summary

PR #2281 (`86bb63b`, *"feat(daemon): enable finish-straggler watchdog by default"*) flipped the finish-straggler watchdog from **opt-in / off** (#2271, `9e5c18f`) to **on by default** at a 120s grace. The watchdog's only liveness signal is *node→daemon traffic*. A node that is **CPU/IO-bound and silent** after its inputs close is indistinguishable from a wedged node, so when the rest of the dataflow has gone quiescent, the watchdog will escalate it `Stop → SIGTERM → SIGKILL` — killing a node that is legitimately still working, and losing its final output.

Before #2281 this affected nobody (watchdog was dark unless `DORA_FINISH_DRAIN_GRACE_SECS` was explicitly set). After #2281 it is the default for every `dora run` / deployment.

## Why this is a false positive

The watchdog never observes whether a node is busy — it only observes silence:

- `last_activity` is updated **only** on a `DaemonRequest` from the node (`binaries/daemon/src/node_communication/mod.rs:221`). A node busy in a compute loop (no `recv`, no `send_output`) sends nothing, so `silent_for` grows unbounded. There is no node→daemon heartbeat (that's why `check_node_health` exists and is opt-in).
- `all_inputs_closed_at` is stamped when the daemon **sends** `AllInputsClosed` (the moment the node's last input closes), *not* when the node acknowledges or finishes — `binaries/daemon/src/lib.rs:3951` and `:5279`. So `drained_for` is pure wall-clock since inputs closed, independent of node progress.

In `select_finish_stragglers` (`binaries/daemon/src/running_dataflow.rs:689`) a node becomes **eligible for SIGKILL** purely on elapsed time:

```rust
// drained arm — running_dataflow.rs:711
if drained_for >= grace { eligible.push(node.id.clone()); }
// never-drained arm — running_dataflow.rs:720
if node.connected && node.silent_for >= grace { eligible.push(node.id.clone()); }
```

No CPU/IO liveness is consulted. The tracking issue #2270 actually *codifies* this: it asks to "treat a node whose receive loop has only been emitting read-timeouts (no liveness) past the grace period as a straggler eligible for escalation" — which is exactly a busy-but-silent node.

## Realistic trigger (common in an AI/robotics framework)

A terminal batch stage, e.g. `source → preprocessor → trainer`:

1. `source` emits a dataset and exits; `preprocessor` forwards and exits.
2. `trainer`'s last input closes → daemon stamps `all_inputs_closed_at[trainer]` and sends `AllInputsClosed`.
3. `trainer` then runs a 10-minute training / large-batch inference / checkpoint write, silently, before emitting a result and exiting.

At that point `source`/`preprocessor` are gone from `running_nodes`, no live source/timer/log node remains to veto (`node_never_finishes` is false for `trainer`), and `trainer` is the only running node with `drained_for >= 120s` → it is escalated and **SIGKILLed at 120s, mid-training**. The final output is lost and `node_results[trainer]` reports `Signal(15)/Signal(9)`.

Gate recap (`select_finish_stragglers`): escalation requires every non-dynamic node to be drained-past-grace or connected-silent-past-grace, with no live source/timer/log node. That is precisely the "terminal compute stage" topology — not an exotic one for ML/robotics pipelines.

## Risk

- **Data loss / silent truncation**: a legitimate long final compute/flush is killed and its output never produced.
- **Default-on**: every deployment is now exposed; the only escape hatch is `DORA_FINISH_DRAIN_GRACE_SECS=off`, which is all-or-nothing and undiscoverable for most users, and disabling it reinstates the original #2152 hang it was meant to fix.
- The killed node surfaces as a signal-exit failure, which downstream tests/tooling may misclassify.

## Affected code / commits

- Default flip: PR #2281 (`86bb63b`) — `parse_finish_drain_grace` unset→enabled (`binaries/daemon/src/lib.rs:5355`), `DEFAULT_FINISH_DRAIN_GRACE = 120s` (`:5335`).
- Selection logic: `binaries/daemon/src/running_dataflow.rs:689` (`select_finish_stragglers`), invoked from `check_finish_stragglers` (`binaries/daemon/src/lib.rs:2365`).
- Mechanism introduced in PR #2271 (`9e5c18f`); tracking issue #2270.

## Suggested mitigations (pick one or combine)

1. **Liveness, not just silence**: before escalating, sample the process (CPU time / `/proc/<pid>/stat` delta, or reuse the existing `spawn_stack_sample_capture`) and skip nodes that are demonstrably making progress; only kill nodes that are genuinely idle/blocked.
2. **Per-node opt-out / per-grace config**: allow a node to declare a longer (or infinite) finish grace in the descriptor, so a known long-running terminal stage isn't governed by the global 120s. Today the grace is global-only.
3. **Raise / document the default conservatively**, and/or require the dataflow to be explicitly "finished" (e.g. only escalate nodes whose event stream has actually dropped) rather than merely silent-past-grace, so a connected, still-polling node is never killed for being CPU-bound.

At minimum, the default-on behavior and its data-loss failure mode for long post-input compute should be documented prominently next to `DORA_FINISH_DRAIN_GRACE_SECS`.

---
*Filed automatically by a scheduled Claude code-review pass; no human has triaged this yet.*

phil-opp:
## Summary

Fixes #2284.

The finish-straggler watchdog (default-on since #2281 at 120 s grace) kills any node that is silent after its inputs close, without being able to distinguish a legitimately busy terminal stage (ML training, large-batch inference, checkpoint write) from a wedged one. A trainer that silently runs for 10 minutes after its inputs close is SIGKILLed at 120 s and its output is lost.

This PR adds a `finish_grace_secs` descriptor field that lets a node declare its own finish grace period, overriding the global `DORA_FINISH_DRAIN_GRACE_SECS`:

```yaml
nodes:
  - id: trainer
    path: ./trainer.py
    finish_grace_secs: 3600.0   # allow up to 1 h of post-input compute
    inputs:
      data: preprocessor/output
    outputs:
      - model
```

Nodes without the field continue to use the global grace (default 120 s), so there is **no change in default behaviour**.

## Changes

| File | Change |
|------|--------|
| `libraries/message/src/descriptor.rs` | `finish_grace_secs: Option<f64>` on `Node` and `CustomNode` |
| `libraries/core/src/descriptor/mod.rs` | Propagate field when resolving `Node` → `CustomNode` |
| `libraries/core/dora-schema.json` | Regenerated to include the new field |
| `binaries/daemon/src/spawn/prepared.rs` | `finish_grace_secs()` extractor; wire into `RunningNode` |
| `binaries/daemon/src/running_dataflow.rs` | `finish_grace_secs` on `RunningNode`; `node_grace` on `StragglerNode`; use `node_grace.unwrap_or(global_grace)` in `select_finish_stragglers` |
| `binaries/daemon/src/lib.rs` | Fill `finish_grace_secs: None` in dynamic-node and test constructors |

## Tests

Three new unit tests in `running_dataflow::tests`:

- `node_with_longer_per_node_grace_is_not_escalated_when_global_grace_expired` — trainer with 600 s per-node grace is not killed after global grace (100 ms) expires
- `node_with_longer_per_node_grace_is_escalated_after_its_own_grace` — trainer is eventually killed after its own grace
- `per_node_grace_does_not_block_other_nodes_already_past_global_grace` — a sibling without a per-node override still escalates on the global grace

All 82 existing daemon tests continue to pass.

## Test plan

- [x] `cargo test -p dora-daemon` — 82 passed, 0 failed
- [x] `cargo clippy -p dora-daemon -p dora-core -p dora-message -p dora-cli -- -D warnings` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_015aeX5FYqLW7YvfQQRHWEkd)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5514,9 +5514,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.14"
+version = "0.11.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "434b42fec591c96ef50e21e886936e66d3cc3f737104fdb9b737c40ffb94c098"
+checksum = "4fcb935c5bec503c2f0e306bdd3e58bb9029dcb14fa8d9ac76e3a5256ac0763e"
 dependencies = [
  "bytes",
  "getrandom 0.3.4",
--- a/binaries/cli/src/command/node/info.rs
+++ b/binaries/cli/src/command/node/info.rs
@@ -81,6 +81,8 @@ struct NodeInfoOutput {
     #[serde(skip_serializing_if = "Option::is_none")]
     health_check_timeout: Option<f64>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    finish_grace_secs: Option<f64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     metrics: Option<MetricsOutput>,
 }
 
@@ -162,6 +164,7 @@ fn info(
         max_restarts: node_desc.max_restarts,
         restart_delay: node_desc.restart_delay,
         health_check_timeout: node_desc.health_check_timeout,
+        finish_grace_secs: node_desc.finish_grace_secs,
         metrics: metrics.and_then(|m| m.metrics).map(|m| MetricsOutput {
             status: m.status.to_string(),
             pid: m.pid,
@@ -266,6 +269,9 @@ fn print_table(output: &NodeInfoOutput) {
     }
     if let Some(timeout) = output.health_check_timeout {
         println!("  Health check timeout: {timeout}s");
+    }
+    if let Some(grace) = output.finish_grace_secs {
+        println!("  Finish grace: {grace}s");
     }
     println!();
 
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2088,6 +2088,7 @@ impl Daemon {
                             max_restart_delay: None,
                             restart_window: None,
                             health_check_timeout: None,
+                            finish_grace_secs: None,
                             module: None,
                             params: Default::default(),
                             cpu_affinity: None,
@@ -5596,6 +5597,7 @@ mod fault_tolerance_tests {
             force_restart_next: Arc::new(AtomicBool::new(false)),
             last_activity: Arc::new(AtomicU64::new(0)),
             health_check_timeout: None,
+            finish_grace_secs: None,
         }
     }
 
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -1053,8 +1053,8 @@ mod tests {
             dynamic: false,
             never_finishes: false,
             connected: true,
-            drained_for: None,       // has not received AllInputsClosed yet
-            silent_for: PAST_GRACE,  // silent past global grace, within per-node grace
+            drained_for: None,      // has not received AllInputsClosed yet
+            silent_for: PAST_GRACE, // silent past global grace, within per-node grace
             node_grace: Some(long_grace),
         };
         let selected = select_finish_stragglers(
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -89,6 +89,10 @@ pub struct RunningNode {
     pub(crate) force_restart_next: Arc<AtomicBool>,
     pub(crate) last_activity: Arc<AtomicU64>,
     pub(crate) health_check_timeout: Option<Duration>,
+    /// Per-node finish-drain grace override (from `finish_grace_secs` in the
+    /// descriptor). When `Some`, overrides the global `DORA_FINISH_DRAIN_GRACE_SECS`
+    /// for this node in the finish-straggler watchdog.
+    pub(crate) finish_grace_secs: Option<Duration>,
 }
 
 impl RunningNode {
@@ -634,6 +638,7 @@ impl RunningDataflow {
                     connected: self.connected_nodes.contains(id),
                     drained_for: self.all_inputs_closed_at.get(id).map(Instant::elapsed),
                     silent_for: Duration::from_millis(now_millis.saturating_sub(last)),
+                    node_grace: node.finish_grace_secs,
                 }
             }),
             &self.finish_escalated,
@@ -683,6 +688,10 @@ struct StragglerNode<'a> {
     /// Time since the last daemon-bound message from this node. Only meaningful
     /// once `connected` — `last_activity` is seeded at spawn time.
     silent_for: Duration,
+    /// Per-node grace override from `finish_grace_secs` in the descriptor.
+    /// When `Some`, takes precedence over the global grace passed to
+    /// [`select_finish_stragglers`] for this node only.
+    node_grace: Option<Duration>,
 }
 
 /// Pure core of [`RunningDataflow::finish_stragglers`].
@@ -704,11 +713,15 @@ fn select_finish_stragglers<'a>(
             // it is stopped only via the explicit-stop path, never escalated here
             return Vec::new();
         }
+        // Per-node `finish_grace_secs` overrides the global grace so that nodes
+        // with long post-input compute (ML training, large-batch inference) are
+        // not SIGKILLed while legitimately busy (dora-rs/dora#2284).
+        let effective_grace = node.node_grace.unwrap_or(grace);
         match node.drained_for {
             // draining: ready past grace; still within grace it is progressing
             // toward exit and does not veto, but is not escalated yet
             Some(drained_for) => {
-                if drained_for >= grace {
+                if drained_for >= effective_grace {
                     eligible.push(node.id.clone());
                 }
             }
@@ -717,7 +730,7 @@ fn select_finish_stragglers<'a>(
             // the dataflow is not otherwise finished); a connected but active
             // node still has work in progress — both veto.
             None => {
-                if node.connected && node.silent_for >= grace {
+                if node.connected && node.silent_for >= effective_grace {
                     eligible.push(node.id.clone());
                 } else {
                     return Vec::new();
@@ -755,6 +768,7 @@ mod tests {
             connected: true,
             drained_for: Some(age),
             silent_for: Duration::ZERO,
+            node_grace: None,
         }
     }
 
@@ -767,6 +781,7 @@ mod tests {
             connected: true,
             drained_for: None,
             silent_for,
+            node_grace: None,
         }
     }
 
@@ -805,6 +820,7 @@ mod tests {
             connected: true,
             drained_for: None,
             silent_for: PAST_GRACE,
+            node_grace: None,
         };
         let selected = select_finish_stragglers(
             [source_node, drained(&sink, PAST_GRACE)].into_iter(),
@@ -825,6 +841,7 @@ mod tests {
             connected: true,
             drained_for: None,
             silent_for: WITHIN_GRACE,
+            node_grace: None,
         };
         let selected = select_finish_stragglers(
             [dynamic_node, drained(&sink, PAST_GRACE)].into_iter(),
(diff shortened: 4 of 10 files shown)


phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop; please verify before relying on it.

No blocking issues found. I traced the new `finish_grace_secs` field end-to-end against the existing `health_check_timeout` plumbing and the threading is complete and consistent:

- Descriptor field added on both `Node` and `CustomNode` (`libraries/message/src/descriptor.rs`), propagated in the `Node → CustomNode` resolve and the Ros2Bridge path (`libraries/core/src/descriptor/mod.rs`), and surfaced through `prepared.rs` → `RunningNode` → `StragglerNode.node_grace`.
- `select_finish_stragglers` applies `node_grace.unwrap_or(global_grace)` in **both** the drained arm and the silent-wedge arm, so a node that doesn't set the field keeps the global default (120s) — i.e. no change in default behavior, which matches the PR's claim.
- Every `RunningNode`/`StragglerNode` constructor (including the dynamic-node and test paths in `lib.rs`) is updated, so it compiles, and the three new unit tests correctly exercise within-grace (no escalation), past-per-node-grace (escalation), and the sibling-not-blocked case.

One minor, non-blocking note: `dora node info` (`binaries/cli/src/command/node/info.rs`) displays `health_check_timeout` but wasn't extended to surface the new `finish_grace_secs`. Purely a display-completeness gap, no correctness impact.

---

phil-opp:
**Automated review by Claude** — _fully automated review; no human has vetted these findings, please verify before acting._

One finding (non-blocking):

`finish_grace_secs` is read via `n.finish_grace_secs.map(Duration::from_secs_f64)` in `binaries/daemon/src/spawn/prepared.rs:169`. `Duration::from_secs_f64` **panics** if the value is negative, NaN, or non-finite, so a descriptor like `finish_grace_secs: -1.0` (or `.nan`) would crash the daemon during node spawn instead of surfacing a clean descriptor error.

This isn't a regression introduced by this PR — the sibling timeout fields right next to it use the exact same unguarded pattern (`health_check_timeout` at :160, and `restart_delay`/`max_restart_delay`/`restart_window` in `restart_config()`), so they'd all panic the same way on a malformed value. The cleanest fix is probably a shared validation step that rejects negative/non-finite values at descriptor-parse time for all of these, rather than special-casing the new field — flagging it here since this PR adds another instance on user-controlled config input.

Otherwise the change looks correct and complete: the field is plumbed through `Node`/`CustomNode` and both resolve arms, `node_grace.unwrap_or(global_grace)` is applied in both the drained and silent-wedge arms of `select_finish_stragglers`, a per-node grace correctly does not veto sibling escalation, and the three new tests exercise the right cases (verified the tests pass on the current head). No other issues found.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review; no human has verified it.

One finding not raised in the previous reviews (the `from_secs_f64` panic and the `dora node info` display gap are already noted, and there are no new commits since, so I won't repeat those).

**A node with a large `finish_grace_secs` that is connected but has *not* drained vetoes escalation for the entire dataflow.** `select_finish_stragglers` (`binaries/daemon/src/running_dataflow.rs`) applies `effective_grace = node.node_grace.unwrap_or(grace)` in both arms, but the two arms behave very differently for a sibling-blocking scenario:

- **Drained arm** (`Some(drained_for)`): a node within its (large) per-node grace simply doesn't push itself and falls through — it does *not* veto. This is the case the new test `per_node_grace_does_not_block_other_nodes_already_past_global_grace` exercises, and it's fine.
- **Never-drained arm** (`None`): `if node.connected && node.silent_for >= effective_grace { push } else { return Vec::new() }`. A connected node that hasn't received `AllInputsClosed` yet and has a large `finish_grace_secs` has `silent_for < effective_grace`, so it hits the `else` and **`return Vec::new()`**, discarding every already-collected eligible sibling and aborting selection for the whole dataflow.

Net effect: a single node with `finish_grace_secs: 3600.0` that wedges (or just hasn't drained yet) now holds the *entire* dataflow's finish-straggler watchdog open for up to an hour — even for sibling nodes that drained and are long past the global 120s grace. Before this PR that window was bounded by the global grace. The previous review's claim that "per-node grace does not veto sibling escalation" only holds for the drained arm; the never-drained arm is not covered by any test.

If the intended semantics are "a busy long-grace node legitimately keeps the dataflow open," this may be acceptable — but it's a behavioral expansion worth making explicit, and a test for the `None`-arm veto case (large per-node grace, not yet drained, alongside a drained-past-global-grace sibling) would pin it down either way.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop; please treat it as a suggestion and verify before acting.

I found one issue not raised in the earlier review rounds:

**The user-facing root `dora-schema.json` was not updated, and it will reject valid YAML that uses the new field.**

The PR regenerates `libraries/core/dora-schema.json` but not the separately-maintained schema at the repository root, `/dora-schema.json`. That root file is the one the docs tell users to wire into their editor:

- `docs/yaml-spec.md:29` and `guide/src/concepts/dataflow-yaml.md:29`:
  ```yaml
  # yaml-language-server: $schema=https://raw.githubusercontent.com/dora-rs/dora/main/dora-schema.json
  ```

Its `Node` definition has `"additionalProperties": false` and enumerates every sibling timing field (`restart_delay`, `max_restart_delay`, `restart_window`, `health_check_timeout`) — but not `finish_grace_secs`. So any user who follows the documented editor setup and adds `finish_grace_secs:` to a node will get a hard `property 'finish_grace_secs' is not allowed` validation error on otherwise-valid YAML — which is exactly the field this PR is introducing.

The schema-regeneration workflow (`.github/workflows/regenerate-schemas.yml`) only touches the two `libraries/core/*.json` files, so this won't self-heal — the root file needs the field added by hand (matching the style of the existing `health_check_timeout` entry).

Otherwise the core daemon logic (`node_grace.unwrap_or(global_grace)`, the `Node → CustomNode` propagation, and the new tests for the common path) looks correct to me.

---

phil-opp:
@claude address the review comments please

phil-opp:

Re-reviewed after `25d8614`/`729aaae`/`5843559`. The findings from the earlier review rounds are all addressed:

- **`from_secs_f64` panic on negative/NaN/non-finite** → `check_timing_fields` in `libraries/core/src/descriptor/validate.rs` now rejects negative/non-finite `finish_grace_secs`, `health_check_timeout`, `restart_delay`, `max_restart_delay`, `restart_window` at descriptor-validation time (called from `check_dataflow`), with tests. Covers the whole field family.
- **Root `dora-schema.json` missing the field** → added to the `Node` def (which has `additionalProperties: false`), so editors with the documented `$schema` wiring no longer reject valid YAML.
- **`dora node info` display gap** → `finish_grace_secs` now in `NodeInfoOutput` (JSON, skipped when `None`) and the table renderer.
- **Never-drained-arm veto** → now pinned by a test and documented as a deliberate behavioral expansion. Worth a maintainer eyeball that holding the entire finish-watchdog open for one node's large grace (up to its `finish_grace_secs`) is the intended semantics — but it's explicit now. The explicit-stop path (`finish_stragglers` early-returns on `stop_sent`) is unaffected, so `dora stop` shutdown is not delayed by this.

One minor (non-blocking) gap: `check_timing_fields` only runs on `CoreNodeKind::Custom`, and `prepared.rs::finish_grace_secs()` returns `None` for `Runtime` nodes — so a `finish_grace_secs` set on an operator/runtime-resolved node is silently ignored rather than applied. Harmless (no panic, since the value is never converted for Runtime nodes), but the field is a no-op there; worth a doc note or a validation warning if operator nodes are expected to support it.

No blocking issues.

---

Issue #1915, opened 2026-05-22
DORA Event Response Framework: SerialEventLoop


RuPingCen:
We propose the DORA SerialEventLoop framework. It provides thread-isolated parallel processing, low-latency event response, built-in timers and thread-safe output, thereby enhancing **QoS for multi-message reception** while improving the real-time performance, scalability and development efficiency of DORA nodes.

### 1 Problem Description
Traditional Dora nodes use native while/for loops to process events synchronously and serially. In scenarios with high-frequency sensor input, multi-topic input, or long-running algorithms, this low-level underlying interaction mechanism suffers from critical technical flaws: poor real-time performance, difficult scalability, significant thread-safety risks, and chaotic lifecycle management. The specific core issues are as follows:

1. **Synchronous blocking & high latency:** All event handlers run serially. A single long-running core algorithm (e.g., image processing) blocks the entire main loop. High-frequency events (e.g., Tick, IMU data) arriving during this period cannot be fetched in time and pile up in the underlying queue, driving system latency from microseconds up to the execution time of the long task.
2. **Inefficient & redundant event routing:** The while/for pattern relies on if/else checks on event IDs for routing. As input events grow, code becomes bloated, lookup efficiency drops, and event-parsing logic is duplicated.
3. **Thread-restricted output:** The native node.send_output() only works on the main thread. Directly sending output from a child thread causes errors; no automatic thread-routing adaptation is available.
4. **No native timer support**: Timed tasks require manual implementation of daemon threads and polling logic. Thread safety and lifecycle management depend entirely on developer-written code.

Table 1: Traditional for/while Event Loop Code in the DORA Framework
```
#include <iostream>
#include <string>
#include <vector>
#include <chrono>
#include <thread>
#include <cstring>

// Include your template's specific API header
#include "node_api.h"

// Simulate a time-consuming algorithm processing function
void process_image(const char* data, size_t data_len) {
    std::cout << "[RawNode] Start processing image, data size: " << data_len << " bytes..." << std::endl;
    // Simulate a 200ms processing delay
    std::this_thread::sleep_for(std::chrono::milliseconds(200));
    std::cout << "[RawNode] Image processing completed." << std::endl;
}

int main() {
    // 1. Initialize the Dora context using your environment-based API
    void *dora_context = init_dora_context_from_env();
    if (dora_context == nullptr) {
        std::cerr << "[RawNode] init dora context failed" << std::endl;
        return -1;
    }

[...]

PR #2288, opened 2026-06-20
feat(c++): thread-safe send_output from worker threads (#1915)

Duckaet:
## TLDR

Implements the suggestions in #1915:

- **A. Thread-safe `send_output`** — addition to `dora-node-api-cxx`.
- **B. The pattern as an opt-in example**
- **C. benchmark**

`init_dora_node` keeps the node in an `Arc<Mutex<DoraNode>>`. `clone_output_sender` *borrows* the node's `OutputSender` and returns a `SafeOutputSender` sharing the same node, so the main thread keeps the full single-threaded API while worker threads get a `Send + Sync` sending handle they can move or share by reference. Every existing bridge signature stays byte-identical.

Every node operation takes that one lock, so the FFI keeps the `&mut DoraNode` exclusivity it requires and a worker's send can never overlap a main-thread send. A poisoned lock fail-stops on *every* handle rather than handing the next caller a daemon connection that may be mid-frame — there is no recovery, by design.

## Benchmark results

12 s per config, 200 ms image work, 10 ms ticks:

| Approach | p50 | **p99** | max |
|---|---|---|---|
| inline blocking | 1.6 ms | **91.7 ms** | 101 ms |
| worker thread + `SafeOutputSender` | 1.6 ms | **3.0 ms** | 5.2 ms |
| split into two nodes (idiomatic) | 1.6 ms | **1.8 ms** | 4.3 ms |

The honest takeaway, written down in the example README: the worker-thread pattern is not faster than Dora's existing multi-node design. Reach for it only when the slow and fast work must share in-process state.

## Testing

- `concurrent_safe_send_output_delivers_every_message_exactly_once` — 8 threads released from a `std::sync::Barrier` into `safe_send_output` at once, 16 sends each; asserts all 128 payloads are delivered exactly once. Daemon-less via `DoraNode::init_testing`, so it runs in the normal test job.
- `a_poisoned_node_fails_stop_on_every_handle` — a panic under the lock must wedge the worker handle, the main-thread sender, and `log_message` alike.
- `safe_output_sender_is_send_and_sync` — compile-time backstop.
- `cargo fmt --all --check`, `cargo clippy -p dora-node-api-cxx --all-targets`, `cargo test -p dora-node-api-cxx`, `cargo check --examples` — all clean.
- `cargo run --example cxx-thread-safe-output` — end-to-end against a real daemon; the example is now wired into the nightly examples job.

## Docs

`docs/api-cxx.md` gains a threading contract under `OutputSender` and a `SafeOutputSender` section (`clone_output_sender`, `safe_send_output`, serialization, fail-stop on poison, handle lifetime).


phil-opp:

**No blocking issues found.** I traced the thread-safety story end-to-end and it holds up:

- `SafeOutputSender { inner: Mutex<OutputSender> }` is the right shape. `create_safe_output_sender` consumes the `Box<OutputSender>` and `safe_send_output` takes `&SafeOutputSender`, locking the `Mutex` before each send — so the `&mut DoraNode` exclusivity the FFI requires is preserved even when many C++ worker threads call concurrently. Poison recovery (`unwrap_or_else(|e| e.into_inner())`) is the correct choice so one panicking send doesn't wedge the others.
- Moving `send_output` out of the node and continuing to drain `dora_node.events` on the main thread is sound: `init_from_env()` returns `events` (the `EventStream`) and the `DoraNode` as independent objects, so there's no shared mutable state between the two paths once the sender is behind the `Mutex`.
- `send_output_internal`'s signature change from `&mut Box<OutputSender>` to `&mut OutputSender` is fine — `&mut Box<_>` and `&mut MutexGuard<_>` both deref-coerce to it, so the existing `send_output` / `send_output_with_metadata` callers are unaffected.
- `safe_output_sender_is_send_and_sync` is a genuine compile-time backstop: `Mutex<OutputSender>: Sync` requires `OutputSender: Send` (i.e. `DoraNode: Send`), so the load-bearing guarantee is enforced by CI rather than assumed.
- The FFI functions the new C++ examples call (`new_metadata`, `event_as_input`, `event_type`, `init_dora_node`, `send_output`) are all exported, and the example/bench dataflows are wired consistently (`worker` mode declares its `result` output, the source emits `tick`/`image`, etc.). Workers borrow the sender and are `join()`ed before it drops.

A couple of non-blocking observations, no action needed: CI hadn't started at review time, so the `DoraNode: Send` assumption and `cargo check --examples` (the only thing that catches a `dora_cli::RunCommand` signature break, since the C++ is compiled at runtime by clang) are worth confirming green before merge. The README's own caveat already flags that the benchmark's headline numbers are illustrative.

---

phil-opp:

One concern on `safe_send_output` (`apis/c++/node/src/lib.rs`):

```rust
let mut guard = sender.inner.lock().unwrap_or_else(|e| e.into_inner());
```

Recovering from a poisoned lock with `into_inner()` is risky here because the lock guards a **stateful** connection. The entire send runs while the guard is held: `send_output_internal` → `DoraNode::send_output_sample` → the daemon control-channel request, which is a non-atomic write-then-read over a single TCP stream. If a worker's send ever panics *after* request bytes are on the wire but *before* the reply is consumed, the mutex is poisoned with the stream's read cursor desynchronized from the daemon framing. `into_inner()` discards that poison and hands the next worker an `OutputSender` whose control channel is mid-frame — its next request reads leftover body bytes as a length prefix and can desync the connection for *every* subsequent worker send (silent data-plane corruption) instead of failing cleanly. Suppressing poison defeats the exact protection poisoning exists to provide.

In fairness, the *concrete* panic surface on this specific path is small today: the fixed copy closure can't mismatch (`data_len == data.len()`), and `send_output_sample`/`zenoh_publish` propagate errors via `?` rather than panicking — so this is a latent/forward-looking risk rather than a guaranteed failure right now. But the prior automated review's claim that the poison recovery is "the correct choice" is too strong. The safer default is to let poison propagate (fail-stop), or — if non-wedging recovery is genuinely wanted — reset/reconnect the `ControlChannel` on recovery rather than reuse a possibly-torn one.

Everything else checks out: `Mutex<OutputSender>` serialization gives the FFI the `&mut DoraNode` exclusivity it needs, the send path and the event-stream path use separate channels (no cross-thread sharing), and the compile-time `Send + Sync` assertion is legitimate.

---

Duckaet:
Thanks for the catch
>
> I've switched `safe_send_output` to fail-stop: if the lock is poisoned, it now returns a `DoraResult` error explaining that a previous send panicked and the daemon connection may be inconsistent, instead of reusing the sender. 
>
> I looked at the reconnect-on-recovery alternative, but it needs `DoraNode` to retain the `DaemonCommunication` config for re-dialing plus daemon-side support for mid-run re-registration — too big change for this PR, if thats required i can raise another issue on this, and push another pr.
>
> Also rebased, to main

phil-opp:
🤖 **Automated review by Claude** — _fully automated (Claude Code); no human has vetted it. Treat as a suggestion, not an approval._

Re-reviewed the diff after `b33485f8` ("fix(c++): fail-stop on poisoned lock").

The fail-stop switch correctly resolves the poison concern raised in my earlier review:

- On a poisoned lock, `safe_send_output` now returns a `DoraResult` error and **never** calls `into_inner()` / `get_mut()` / `clear_poison()` — so a subsequent worker is never handed the possibly-torn `OutputSender`. That's exactly the fail-stop behavior that was requested.
- Because the poison branch never acquires the guard, the mutex stays poisoned, so every later `safe_send_output` re-observes the error and fail-stops identically. No accidental recovery.
- No new panic/FFI hazard in the added code: it's a `match` + `format!` (neither panics), and the metadata is dropped cleanly on the error path.

The rest of the thread-safety design is unchanged and still holds (`Mutex<OutputSender>` gives the FFI its `&mut DoraNode` exclusivity, the event-stream and send paths use separate channels, and the `Send + Sync` assertion is a real compile-time backstop). **No new issues found.**

Non-blocking note: `safe_send_output` returns a plain `DoraResult` struct, not a `Result`, so cxx's boundary `catch_unwind` will **abort** the process if a send ever panics while the guard is held, rather than unwinding to leave the mutex observably poisoned. In practice that makes the poison branch defense-in-depth here — which is fine; fail-stop is strictly safer than the previous recovery either way. Deferring true reconnect-on-recovery to a follow-up seems a reasonable scope call.

Worth confirming CI is green before merge — in particular `cargo check --examples`, since it's the only gate that catches a `dora_cli::RunCommand` signature break (the C++ is compiled at runtime by clang).

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no human in the loop._

---

Duckaet:
@phil-opp can i get a green flag?

phil-opp:
**Automated review by Claude** — fully automated, no human has vetted this. Starting point, not a verdict.

I reviewed the diff directly (not the description). The thread-safety design is sound: `SafeOutputSender` moves the `DoraNode` behind a `Mutex`, `safe_send_output` serializes on it, and `DoraNode` shares no unsynchronized mutable state with the event loop, so concurrent worker sends are safe. The fail-stop on a poisoned lock is the right call. No correctness issue found there.

One thing worth flagging, since the whole point of the PR is concurrent `send_output`: **nothing added actually exercises concurrent sends.**

- `apis/c++/node/src/lib.rs:1141` `safe_output_sender_is_send_and_sync` is a compile-time `Send + Sync` assertion only — it never spawns a thread.
- The examples that spawn threads are structured so the sends never overlap:
  - `examples/cxx-thread-safe-output/node-rust-api/main.cc`: `tick` every 300 ms, each worker sleeps 200 ms before sending — so worker N finishes (~t+200) before worker N+1 is even spawned (~t+300). At most one thread is ever inside `safe_send_output`.
  - `examples/cxx-thread-safe-output-bench` (`worker` mode): `image` every 1000 ms, 200 ms work — same non-overlap.
- Both examples are `log_skip`'d in `scripts/smoke-all.sh`, so they only run in the nightly cxx examples job, and even there they don't create mutex contention.

So the "thread-safe" guarantee is currently validated by type-system + `Mutex` reasoning, not by any executed concurrent path. That reasoning is correct, so I wouldn't block on this — but a test that actually puts two threads inside `safe_send_output` simultaneously (e.g. spawn N workers that block on a shared barrier, then all send at once, and assert every send succeeds), ideally under ThreadSanitizer, would make the added coverage meaningfully test the property the PR is about. As written, the concurrency tests would pass even if the send path weren't actually concurrency-safe.

---

phil-opp:

Re-checked after the 2026-08-18 rebase onto `main` (`fe89982`). That commit only resolves merge conflicts in `Cargo.toml`, `apis/c++/node/src/lib.rs`, and `scripts/smoke-all.sh` — the PR's actual content (fail-stop `safe_send_output`, the `&mut OutputSender` signature change, the two examples, the compile-time `Send + Sync` assertion) is unchanged. The thread-safety design still holds: `Mutex<OutputSender>` serializes access so the FFI keeps its `&mut DoraNode` exclusivity, and fail-stop on a poisoned lock is the right call. No new issues from the rebase.

One previously-raised item is still outstanding and I'd flag again: **nothing in the PR actually executes two concurrent sends.**
- `safe_output_sender_is_send_and_sync` is a compile-time `Send + Sync` assertion; it never spawns a thread.
- `examples/cxx-thread-safe-output` — `tick` every 300 ms, each worker sleeps 200 ms before sending, so worker *N* finishes before worker *N+1* spawns; at most one thread is ever inside `safe_send_output`.
- `examples/cxx-thread-safe-output-bench` (`worker` mode) — `image` every 1000 ms, 200 ms work — same non-overlap. Both examples are `log_skip`'d in `scripts/smoke-all.sh`.

So the "thread-safe" guarantee is currently proven by the type system + `Mutex` reasoning (which is sound), not by any executed concurrent path — the tests as written would pass even if the send path weren't concurrency-safe. Not a blocker given the design, but a test that puts N threads inside `safe_send_output` simultaneously (shared barrier → all send at once → assert every send succeeds), ideally under ThreadSanitizer, would make the coverage actually exercise the property this PR is about.

---

phil-opp:

Thanks for sticking with this, and sorry about the wait. The benchmark comparing the worker-thread pattern against the idiomatic two-node split is exactly what #1915 asked for, and the honest conclusion in the README ("not faster than multiple nodes; reach for it when the work must share in-process state") is the right thing to write down. The thread-safety reasoning is correct too: `Mutex<OutputSender>` gives the FFI the `&mut DoraNode` exclusivity it needs, the event stream and the control channel are separate daemon connections so there is no aliasing with the main loop, and fail-stop on a poisoned lock is the right call.

The one thing that needs to change is the ownership shape. `create_safe_output_sender` takes the `Box<OutputSender>` by value, and cxx's move constructor nulls out the source (`Box(Box &&other) : ptr(other.ptr) { other.ptr = nullptr; }`). So after `create_safe_output_sender(std::move(dora_node.send_output))`, `dora_node.send_output` holds null — and the eleven other bridge functions that take `&mut Box<OutputSender>` or `&Box<OutputSender>` still compile against it: `send_output`, `send_output_with_metadata`, `send_arrow_output`, `send_arrow_output_with_metadata`, `close_outputs`, `log_message`, `node_config_json`, `dataflow_descriptor_json`, `send_service_request`, `send_arrow_service_request`, `send_service_response`. That is a segfault rather than a compile error, and since 1.0 freezes the cxx bridge we would be stuck with it. It also diverges from what #1915 asked for, which was a shared handle.

Could you make it a shared handle instead?

```rust
pub struct OutputSender(Arc<Mutex<dora_node_api::DoraNode>>);   // private field
pub struct SafeOutputSender(Arc<Mutex<dora_node_api::DoraNode>>);

fn clone_output_sender(output_sender: &Box<OutputSender>) -> Box<SafeOutputSender>;
```

Every existing bridge signature stays byte-identical (so the breaking-change gate does not fire), the main thread keeps the full API, and workers get a handle they can move or share. An uncontended `Mutex` is ~20 ns against a shm publish, so the cost is noise.

Three smaller things alongside that:

- The examples are not actually run by CI. `nightly.yml`'s examples job lists each `cargo run --example` explicitly and neither of these is in it, so the `scripts/smoke-all.sh` skip reason "covered by cxx examples job" is not true yet. Adding a step next to `cxx-service-action` would fix it. The bench does not need to run in CI — mark it as a manual measurement tool.
- `docs/api-cxx.md` needs an entry for the new function, and more importantly a statement of the threading contract. That file currently contains zero occurrences of the word "thread", so the rule this PR depends on is written down only inside the new example's README.
- The concurrency test raised earlier is still open, and it is easier than it looks now: `DoraNode::init_testing` (`apis/rust/node/src/node/mod.rs:1112`) is `pub` and needs no daemon, so a `#[cfg(test)]` test in `apis/c++/node/src/lib.rs` can put 8 threads on a `std::sync::Barrier`, have them all send at once, and assert the `TestingOutput` received every message with no loss or duplication. Right now both examples are structured so sends never overlap (300 ms tick vs 200 ms work), so nothing exercises the property the PR is about.

Happy to look again quickly once those are in.

phil-opp:

Thanks for sticking with this, and sorry about the wait. The benchmark comparing the worker-thread pattern against the idiomatic two-node split is exactly what #1915 asked for, and the honest conclusion in the README ("not faster than multiple nodes; reach for it when the work must share in-process state") is the right thing to write down. The thread-safety reasoning is correct too: `Mutex<OutputSender>` gives the FFI the `&mut DoraNode` exclusivity it needs, the event stream and the control channel are separate daemon connections so there is no aliasing with the main loop, and fail-stop on a poisoned lock is the right call.

The one thing that needs to change is the ownership shape. `create_safe_output_sender` takes the `Box<OutputSender>` by value, and cxx's move constructor nulls out the source (`Box(Box &&other) : ptr(other.ptr) { other.ptr = nullptr; }`). So after `create_safe_output_sender(std::move(dora_node.send_output))`, `dora_node.send_output` holds null — and the eleven other bridge functions that take `&mut Box<OutputSender>` or `&Box<OutputSender>` still compile against it: `send_output`, `send_output_with_metadata`, `send_arrow_output`, `send_arrow_output_with_metadata`, `close_outputs`, `log_message`, `node_config_json`, `dataflow_descriptor_json`, `send_service_request`, `send_arrow_service_request`, `send_service_response`. That is a segfault rather than a compile error, and since 1.0 freezes the cxx bridge we would be stuck with it. It also diverges from what #1915 asked for, which was a shared handle.

Could you make it a shared handle instead?

```rust
pub struct OutputSender(Arc<Mutex<dora_node_api::DoraNode>>);   // private field
pub struct SafeOutputSender(Arc<Mutex<dora_node_api::DoraNode>>);

fn clone_output_sender(output_sender: &Box<OutputSender>) -> Box<SafeOutputSender>;
```

Every existing bridge signature stays byte-identical (so the breaking-change gate does not fire), the main thread keeps the full API, and workers get a handle they can move or share. An uncontended `Mutex` is ~20 ns against a shm publish, so the cost is noise.

Three smaller things alongside that:

- The examples are not actually run by CI. `nightly.yml`'s examples job lists each `cargo run --example` explicitly and neither of these is in it, so the `scripts/smoke-all.sh` skip reason "covered by cxx examples job" is not true yet. Adding a step next to `cxx-service-action` would fix it. The bench does not need to run in CI — mark it as a manual measurement tool.
- `docs/api-cxx.md` needs an entry for the new function, and more importantly a statement of the threading contract. That file currently contains zero occurrences of the word "thread", so the rule this PR depends on is written down only inside the new example's README.
- The concurrency test raised earlier is still open, and it is easier than it looks now: `DoraNode::init_testing` (`apis/rust/node/src/node/mod.rs:1112`) is `pub` and needs no daemon, so a `#[cfg(test)]` test in `apis/c++/node/src/lib.rs` can put 8 threads on a `std::sync::Barrier`, have them all send at once, and assert the `TestingOutput` received every message with no loss or duplication. Right now both examples are structured so sends never overlap (300 ms tick vs 200 ms work), so nothing exercises the property the PR is about.

Happy to look again quickly once those are in.

Duckaet:
Thanks — the null-`Box` problem is real and I've reshaped it the way you suggested. All four points are addressed.

**1. Ownership: shared handle, not a consumed one** (`ef669b94`)

```rust
type SharedNode = Arc<Mutex<dora_node_api::DoraNode>>;

pub struct OutputSender(SharedNode);
pub struct SafeOutputSender(SharedNode);

fn clone_output_sender(output_sender: &Box<OutputSender>) -> Box<SafeOutputSender>;
```

`clone_output_sender` borrows, so `dora_node.send_output` stays valid and keeps the full API. Every existing bridge signature is byte-identical — the generated header confirms it, and the only new entries are `clone_output_sender` and the unchanged `safe_send_output`.

One consequence worth calling out: because *every* node operation now takes the same lock, fail-stop on poison applies to the whole API, not just `safe_send_output`. `send_output`, `log_message`, `close_outputs`, `node_config_json`, `dataflow_descriptor_json` and the service helpers all return the poison error once the node is wedged, which is what the `a_poisoned_node_fails_stop_on_every_handle` test asserts. The other consequence is lifetime: the node is dropped when the last handle goes, so callers must join their workers — documented in both the header and `docs/api-cxx.md`.

**2. Concurrency test** (`ef669b94`)

`concurrent_safe_send_output_delivers_every_message_exactly_once` puts 8 threads on a `std::sync::Barrier`, releases them into `safe_send_output` at once (16 sends each), and asserts all 128 payloads arrive through `TestingOutput::ToChannel` with no loss and no duplication — plus that the main-thread sender still works afterwards. `DoraNode::init_testing` means no daemon, so it runs in the normal `cargo test` job. I didn't run it under ThreadSanitizer (needs a nightly toolchain, which the CI matrix doesn't have); happy to add a dedicated job if you'd like that gate.

I also fixed the examples so they actually overlap: the tick was 300 ms against 200 ms of work, so worker N finished before worker N+1 was spawned. It's now 100 ms against 300 ms, and the run shows workers 0–3 dispatched before worker 0 returns.

**3. CI** (`b47b4df2`)

Added a `C++ Thread-Safe Output example` step next to `cxx-service-action` in `nightly.yml`, and corrected the `smoke-all.sh` skip reasons — the old "covered by cxx examples job" was not true. The bench is marked a manual measurement tool and stays out of CI: percentiles from a shared runner would be noise.

**4. Docs** (`3439f9bb`)

`docs/api-cxx.md` gains a `Threading contract` section under `OutputSender` and a `SafeOutputSender` section covering `clone_output_sender`, `safe_send_output`, the serialization guarantee, the fail-stop-on-poison behavior, and the handle lifetime.

**Verification**

- `cargo test -p dora-node-api-cxx` — 19 passed, including the two new tests
- `cargo clippy -p dora-node-api-cxx --all-targets` — no findings in the crate
- `cargo fmt --all --check`, `cargo check --examples` — clean
- `cargo run --example cxx-thread-safe-output` — ran end-to-end against a real daemon; `log_message` and `close_outputs` on the main-thread sender both work after the clone, and the workers overlap

phil-opp:

Re-reviewed after the three new commits (`ef669b94`, `b47b4df2`, `3439f9bb`). The blocking issue from the previous review — `create_safe_output_sender` consuming the `Box<OutputSender>`, which cxx's move constructor would null out into a segfault against the eleven other bridge functions — is fixed: `clone_output_sender(&Box<OutputSender>)` now borrows and hands out an `Arc<Mutex<DoraNode>>` clone, and every existing bridge signature is byte-identical (only `clone_output_sender` and `safe_send_output` are added, both additive to the frozen cxx surface).

I verified the reshaping is internally consistent: every node-access path now funnels through `lock_node` / `send_output_locked` (including `send_service_response` → `send_output_with_metadata`, and all arrow paths → `send_arrow_output_impl`), so nothing accesses the node off-lock, and there is no nested/reentrant locking that would deadlock the non-reentrant `std::Mutex`. The event stream stays outside the lock, so the main loop's `next()` is never blocked by a worker send. The new `concurrent_safe_send_output_delivers_every_message_exactly_once` test genuinely races 8 barrier-released threads and asserts exactly-once delivery, and `a_poisoned_node_fails_stop_on_every_handle` covers the fail-stop-on-poison path across handles; the example joins all workers before the node leaves scope, matching the documented handle-lifetime contract. No new issues found.



---

phil-opp:
CI clippy check fails

Duckaet:
CI is green on `ca2a019e`.

The first push went red on Clippy for a semantic conflict with `main` rather than anything in the change: #2972 replaced the `dora_node_api::integration_testing` re-exports of tokio's channel types with an opaque `output_channel()` / `OutputReceiver` pair, and CI compiles the merge ref, so the new concurrency test's imports resolved locally but not against current `main`.

Merged `main` in and resolved that inside the merge commit — the test helper builds its channel with `output_channel()` now, path-qualified because `integration_testing::OutputSender` collides with this crate's own `OutputSender`. I merged rather than rebased to avoid force-pushing; happy to rebase if you'd prefer a linear history before merge.

Re-verified on the merged tree: `cargo fmt --all --check`, `cargo clippy -p dora-node-api-cxx --all-targets -- -D warnings` (exit 0, zero diagnostics), `cargo test -p dora-node-api-cxx` (19 passed), `cargo check --examples`.

Worth noting for the CI point: `Test (ubuntu-latest)` shows as skipped here because `.github/workflows/ci.yml` gates it on `github.event_name != 'pull_request'`, so the suite runs in the merge-queue batch. Its step is `cargo test --all` with no `dora-node-api-cxx` exclusion, so the two new tests do get exercised — just at merge time rather than on the PR.

phil-opp:
Thanks!

phil-opp:
🤖 **Automated review** — fully automated Claude Code review; no human vetted this.

Reviewed the merge commit `ca2a019e` that landed since the last review. It merges `main` and resolves the semantic conflicts in this PR's own code against two upstream changes: the `arrow` → `arrow_v59` / `ArrowData` → `DoraArray` migration, and #2972 replacing `integration_testing::unbounded_channel` with `output_channel()` / `OutputReceiver`. The resolutions are mechanical API adaptations — the concurrency test helper still drives `TestingOutput::ToChannel` via `DoraNode::init_testing`, and `input_bytes` / the arrow-FFI paths route correctly through `data.as_array()`. No new issues found; the shared-handle design reviewed previously is unchanged.



---

PR #2289, opened 2026-06-20
fix(cli): require dot boundary in find_node_log_files prefix match

### LINKED ISSUE #2283: `dora logs --node <NAME>` matches other nodes whose names share a prefix (local log lookup)
## Summary

When reading local logs for a single node, `find_node_log_files` in `binaries/cli/src/command/logs.rs` selects log files using a bare `starts_with(prefix)` test **without** requiring a `.` boundary after the node name. As a result, `dora logs --local --node <NAME>` also picks up log files belonging to *other* nodes whose names start with `<NAME>`, silently interleaving unrelated logs into the output.

## Location

`binaries/cli/src/command/logs.rs`, `find_node_log_files` (lines ~425–439):

```rust
let prefix = format!("log_{node_str}");
if name.starts_with(&prefix) && (name.ends_with(".jsonl") || name.ends_with(".txt")) {
    files.push(entry.path());
}
```

## Why it's wrong

Daemon log files are named `log_<node_id>.jsonl` and `log_<node_id>.<index>.jsonl` (see `binaries/daemon/src/log.rs::log_path` / `log_path_rotated`). The intended match — per the inline comment `// Match: log_<node>.jsonl, log_<node>.1.jsonl, log_<node>.txt` — requires the character immediately following the node name to be a `.`. The current check does not enforce that boundary.

### Concrete reproduction

Given a dataflow with two nodes `cam` and `cam_left`, the `out/<dataflow>/` directory contains:

```
log_cam.jsonl
log_cam_left.jsonl
```

Running:

```
dora logs --local --node cam
```

produces `prefix = "log_cam"`, and `log_cam_left.jsonl`:
- `starts_with("log_cam")` → **true**
- `ends_with(".jsonl")`    → **true**

so it is incorrectly included. The output for node `cam` then contains `cam_left`'s log lines as well.

A second-order effect: `rotation_index("log_cam_left.jsonl")` strips `log_` → `cam_left`, strips `.jsonl` → `cam_left`, finds no `.`, and returns `0`, so the foreign file is also treated as a "current" file rather than being rejected.

Note that the all-nodes path (`find_log_files`) is unaffected because it intentionally matches any `log_*` file; the bug is specific to the per-node filter.

## Suggested fix

Require a `.` immediately after the node name before checking the extension, e.g. match against candidate prefixes `log_<node>.` (current/rotated `.jsonl`) and `log_<node>.txt`, or strip the `log_<node>` prefix and verify the remainder starts with `.`:

```rust
let stem = format!("log_{node_str}");
let rest = match name.strip_prefix(&stem) {
    Some(rest) => rest,
    None => continue,
};
// require a boundary: ".jsonl", ".<idx>.jsonl", or ".txt"
if rest.starts_with('.') && (rest.ends_with(".jsonl") || rest.ends_with(".txt")) {
    files.push(entry.path());
}
```

A regression unit test with two prefix-sharing node names (`cam` / `cam_left`) would lock this down.

---
*This issue was created by an automated scheduled Claude code-review check. Please verify before acting on it.*

Supull:
Fixes #2283

When filtering log files for a specific node, the previous `starts_with(prefix)` 
check would also match other nodes whose names share a prefix (e.g. `cam` 
matching `cam_left`).

The fix strips the node stem first, then requires the remainder to start with `.` 
before checking the extension — ensuring only `log_<node>.jsonl`, 
`log_<node>.<idx>.jsonl`, and `log_<node>.txt` are matched.

Added a regression test with two prefix-sharing node names to prevent recurrence.

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -432,8 +432,12 @@ fn find_node_log_files(dataflow_dir: &Path, node: &NodeId) -> Result<Vec<PathBuf
         let entry = entry?;
         let name = entry.file_name().to_str().unwrap_or_default().to_string();
         // Match: log_<node>.jsonl, log_<node>.1.jsonl, log_<node>.txt
-        let prefix = format!("log_{node_str}");
-        if name.starts_with(&prefix) && (name.ends_with(".jsonl") || name.ends_with(".txt")) {
+        let stem = format!("log_{node_str}");
+        let rest = match name.strip_prefix(&stem) {
+            Some(rest) => rest,
+            None => continue,
+        };       
+        if rest.starts_with('.') && (rest.ends_with(".jsonl") || rest.ends_with(".txt")) {
             files.push(entry.path());
         }
     }
@@ -898,5 +902,21 @@ mod tests {
         let now = Utc::now();
         let msg = make_msg("hello", Some("sensor"), Some("target"), now);
         assert!(!matches_grep(&msg, Some("zzz_missing")));
+    }
+
+    #[test]
+    fn find_node_log_files_does_not_match_prefix_sharing_nodes() {
+        use std::fs::File;
+        use tempfile::tempdir;
+
+        let dir = tempdir().unwrap();
+        File::create(dir.path().join("log_cam.jsonl")).unwrap();
+        File::create(dir.path().join("log_cam_left.jsonl")).unwrap();
+
+        let node_id = NodeId::from("cam".to_string());
+        let files = find_node_log_files(dir.path(), &node_id).unwrap();
+
+        assert_eq!(files.len(), 1);
+        assert!(files[0].file_name().unwrap() == "log_cam.jsonl");
     }
 }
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** — _this is a fully automated review; no human has vetted it. Treat it as advisory._

I reviewed the diff against the actual on-disk log naming and call sites (not the PR description).

**No issues found.**

- The fix correctly scopes to `find_node_log_files`: it strips the `log_<node>` stem and then requires the remainder to start with `.` before checking the extension. This properly rejects prefix-sharing siblings (`log_cam_left.jsonl` no longer matches node `cam`) while still accepting `log_<node>.jsonl`, `log_<node>.<index>.jsonl`, and `log_<node>.txt`, matching the format emitted in `binaries/daemon/src/log.rs`.
- Node names containing a `.` are handled correctly because the full node name is part of the stem checked before the `.starts_with('.')` boundary test.
- The sibling all-files path (`find_log_files`, the no-filter case) intentionally keeps `starts_with("log_")` and is correctly left untouched — it isn't affected by this bug.
- The regression test reproduces the exact reported case (`log_cam.jsonl` + `log_cam_left.jsonl`) and would have failed against the old `starts_with` logic.

Minor, non-blocking: the final commit is an empty "ci: trigger workflows" commit — harmless history noise, no action needed.

---

PR #2294, opened 2026-06-21
fix(message): reject node ids starting with '.' to prevent path traversal

### LINKED ISSUE #2293: Node id of `.` or `..` is accepted and resolves the managed Python env dir into `.dora/`, which `uv venv --clear` then wipes

## Summary

A node `id` consisting only of dot characters (`.` or `..`) passes all validation and then flows **unsanitized** into a filesystem path that is destroyed with `uv venv --clear`. Because `..` and `.` are path-relative segments, the env directory collapses out of its intended `python-envs/<node-id>` subdirectory and lands on a parent directory, which is then cleared.

## Root cause

`validate_node_id` permits `.` and `-` in any position and only rejects the empty string and disallowed characters. A string of only dots is therefore valid:

`libraries/message/src/id.rs:10`
```rust
fn validate_node_id(id: &str) -> Result<(), InvalidId> {
    if id.is_empty() {
        return Err(InvalidId("identifier must not be empty".into()));
    }
    if let Some(ch) = id
        .chars()
        .find(|c| !c.is_ascii_alphanumeric() && *c != '_' && *c != '-' && *c != '.')
    {
        // ...
    }
    Ok(())
}
```

So `validate_node_id("..")` and `validate_node_id(".")` both return `Ok(())`. Nothing in `descriptor/validate.rs` adds a `.`/`..` guard either.

## The sink

The node id is joined directly into a path with no normalization or containment check:

`libraries/core/src/build/mod.rs:338`
```rust
pub fn managed_python_env_dir(node: &ResolvedNode, node_working_dir: &Path) -> Option<PathBuf> {
    node_requires_managed_python_env(node).then(|| {
        node_working_dir
            .join(".dora")
            .join("python-envs")
            .join(node.id.as_ref())   // <-- raw node id, no sanitization
    })
}
```

That path is then handed to `uv venv --clear`:

`libraries/core/src/build/build_command.rs:176`
```rust
if !managed_python_interpreter(python_env_dir).is_file() {
    let mut cmd = Command::new("uv");
    cmd.arg("venv");
    cmd.arg("--clear");
    cmd.arg(python_env_dir);
    // ...
```

## Impact

For a Python node that needs a managed env (a `.py` custom node with a `build:` block, or a runtime node with a non-conda Python operator), building a dataflow that contains:

```yaml
nodes:
  - id: ..
    path: foo.py
    build: pip install ...
```

resolves `python_env_dir` to `<working_dir>/.dora/python-envs/..`, which the OS resolves to `<working_dir>/.dora`. `uv venv --clear` then wipes the entire `.dora/` directory — destroying **all** sibling nodes' managed environments and other dora build state, not just this node's env. A node id of `.` similarly targets `<working_dir>/.dora/python-envs` (clearing every node env).

Because node ids cannot contain `/` (rejected by `validate_node_id`), the traversal is bounded to one level up, so this is a data-loss/correctness bug rather than arbitrary-path write. It still fails **silently and destructively** instead of producing a clean validation error.

The maintainers already treat leading-dot segments as a traversal risk elsewhere — the hub catalog validator explicitly rejects them:

`libraries/hub-client/src/index.rs:427` (`is_valid_key_part`)
```rust
pub fn is_valid_key_part(part: &str) -> bool {
    !part.is_empty()
        && part.len() <= 64
        && part.chars().all(|c| c.is_ascii_alphanumeric() || matches!(c, '-' | '_' | '.'))
        && !part.starts_with('.')   // <-- so `..` and dotfiles can't become a directory
}
```

`validate_node_id` lacks the equivalent `!starts_with('.')` / `.`/`..` rejection.

## Suggested fix

Reject node ids that are `.` or `..` (and arguably any id beginning with `.`, mirroring `is_valid_key_part`) in `validate_node_id`. The same consideration likely applies to `validate_data_id` (`.`, `..`, and `../x` are all currently accepted) if DataIds ever reach a filesystem sink. A defense-in-depth containment check in `managed_python_env_dir` (verify the joined path still starts with `python-envs/`) would also harden the sink.

## Suggested regression test

```rust
#[test]
fn node_id_rejects_dot_segments() {
    assert!(validate_node_id(".").is_err());
    assert!(validate_node_id("..").is_err());
}
```

phil-opp:
Fixes #2293

## Summary

A node `id` of `.` or `..` passed `validate_node_id` because only the empty string and disallowed characters were checked. The id was then joined **unsanitized** into `.dora/python-envs/<node-id>`, so `uv venv --clear` silently destroyed sibling python envs or the entire `.dora/` directory.

### `libraries/message/src/id.rs`

Added a `starts_with('.')` guard to `validate_node_id`, mirroring the same check already present in `hub-client`'s `is_valid_key_part`:

```rust
if id.starts_with('.') {
    return Err(InvalidId(format!(
        "identifier '{id}' must not start with '.' \
         (dot-segments such as '.' and '..' can traverse parent directories)"
    )));
}
```

This rejects `.`, `..`, and any dotfile-style id (e.g. `.hidden`). Embedded dots (`node.v2`, `a.b.c`) are still allowed.

Added regression tests:
- `.` and `..` are rejected
- `.hidden`, `.config` are rejected  
- `node.v2`, `a.b.c` still pass

### `libraries/core/src/build/mod.rs`

Added a `debug_assert` containment check in `managed_python_env_dir` as defense-in-depth: the joined path must have more components than the `python-envs/` base. This assertion guards against future callers that bypass `validate_node_id`.

## Verification

- `cargo test -p dora-message --lib -- id::tests` — 12 passed ✅
- `cargo clippy -p dora-message -p dora-core -- -D warnings` — clean ✅
- `cargo fmt --all -- --check` — clean ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_01GvxT3tKqQE9kW2PAJDLpkx)_

### DIFF
--- a/libraries/core/src/build/mod.rs
+++ b/libraries/core/src/build/mod.rs
@@ -337,10 +337,18 @@ fn confine_subdir(clone_dir: &Path, subdir: &str) -> eyre::Result<PathBuf> {
 /// directory and silently clobber each other's envs.
 pub fn managed_python_env_dir(node: &ResolvedNode, node_working_dir: &Path) -> Option<PathBuf> {
     node_requires_managed_python_env(node).then(|| {
-        node_working_dir
-            .join(".dora")
-            .join("python-envs")
-            .join(node.id.as_ref())
+        let envs_base = node_working_dir.join(".dora").join("python-envs");
+        let env_dir = envs_base.join(node.id.as_ref());
+        // Defense-in-depth: the node id must not escape the python-envs/ directory.
+        // validate_node_id already rejects leading-dot ids (`.`, `..`, `.hidden`),
+        // so this assertion should never fire in practice — it guards against future
+        // callers that bypass validation.
+        debug_assert!(
+            env_dir.components().count() > envs_base.components().count(),
+            "node id '{id}' escaped python-envs directory (env_dir={env_dir:?}, base={envs_base:?})",
+            id = node.id,
+        );
+        env_dir
     })
 }
 
--- a/libraries/message/src/id.rs
+++ b/libraries/message/src/id.rs
@@ -3,13 +3,26 @@ use std::{borrow::Borrow, convert::Infallible, str::FromStr};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-/// Validate that a node identifier contains only safe characters: `[a-zA-Z0-9_.-]`.
+/// Validate that a node identifier contains only safe characters: `[a-zA-Z0-9_.-]`,
+/// does not start with `.`, and is not empty.
 ///
 /// NodeIds must NOT contain `/` because `/` is the separator between
 /// `<node_id>/<output_id>` in input mapping syntax.
+///
+/// Leading dots are rejected because the node id is joined into filesystem paths
+/// (e.g. `managed_python_env_dir` appends `node.id` under `.dora/python-envs/`).
+/// A node id of `..` or `.` would resolve to a parent directory, and
+/// `uv venv --clear` would then wipe that directory — destroying sibling envs.
+/// This mirrors the `is_valid_key_part` guard in the hub-client index.
 fn validate_node_id(id: &str) -> Result<(), InvalidId> {
     if id.is_empty() {
         return Err(InvalidId("identifier must not be empty".into()));
+    }
+    if id.starts_with('.') {
+        return Err(InvalidId(format!(
+            "identifier '{id}' must not start with '.' \
+             (dot-segments such as '.' and '..' can traverse parent directories)"
+        )));
     }
     if let Some(ch) = id
         .chars()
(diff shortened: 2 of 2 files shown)


phil-opp:
**Automated review by Claude** — _fully automated review; no human has vetted these findings, please verify before acting._

Reviewed the diff directly (not the description).

The core fix is correct: `validate_node_id` now rejects empty, leading-dot, and disallowed-character ids, so `.` / `..` / `.hidden` can no longer be joined into `.dora/python-envs/<id>` and let `uv venv --clear` escape the intended directory. Embedded dots (`node.v2`, `a.b.c`) still pass, and the new tests cover both directions.

One minor, non-blocking note on the defense-in-depth `debug_assert!` in `managed_python_env_dir`: as written it does **not** catch the `..` case it claims to guard. `PathBuf::join("..")` *appends* a `ParentDir` component rather than resolving it, so `env_dir.components().count()` is `base + 1` and the `> base` check passes (no panic) for id `..`. Conversely, `join(".")` is normalized away by `components()`, so the count is equal and the assert *does* fire for `.`. Net: the redundant guard trips on the harmless `.` and misses the dangerous `..`. Since `validate_node_id` is the real guard and already rejects both, this is cosmetic — but if the assertion is meant to be a genuine backstop against a future caller that bypasses validation, it should compare a normalized/canonicalized path rather than raw component counts.

---

PR #2303, opened 2026-06-23
fix(cli): show orphaned spans in `dora trace view` instead of silently dropping them

### LINKED ISSUE #2301: `dora trace view` silently drops orphaned spans (and prints nothing for a root-less trace)
## Summary

`print_span_tree` in `binaries/cli/src/command/trace/view.rs` only emits spans that are reachable from a span whose `parent_span_id == None`. Any span whose `parent_span_id` points to a `span_id` that is **not present in the returned set** (an "orphan") — together with its entire subtree — is silently omitted from the output. There is no warning and no fallback.

A direct consequence: if the returned span set is non-empty but contains **no** span with `parent_span_id == None` (e.g. the root span is not in the set), `dora trace view <id>` prints **nothing at all**, even though the early `spans.is_empty()` guard already passed.

## Code reference

`binaries/cli/src/command/trace/view.rs:75-90`

```rust
fn print_span_tree(spans: &[TraceSpan]) {
    let mut children: HashMap<Option<u64>, Vec<&TraceSpan>> = HashMap::new();
    for span in spans {
        children.entry(span.parent_span_id).or_default().push(span);
    }

    for list in children.values_mut() {
        list.sort_by_key(|s| s.start_time);
    }

    if let Some(roots) = children.get(&None) {   // <-- only `None`-parent spans are roots
        for root in roots {
            print_span(root, &children, 0);
        }
    }
    // spans with a Some(parent) that isn't in `spans` are never visited
}
```

Only `children.get(&None)` is used as the set of roots. A span keyed under `Some(parent_id)` where `parent_id` is absent from the set is never reached by the recursion in `print_span`.

## Why this is reachable (not theoretical)

The sibling code path on the coordinator side already defends against exactly the "no `None`-parent root" case, which is strong evidence the maintainers consider it reachable:

`binaries/coordinator/src/lib.rs:4536-4539`

```rust
let root = spans
    .iter()
    .find(|s| s.parent_span_id.is_none())
    .unwrap_or(&spans[0]);   // <-- falls back when no root span is present
```

`handle_get_traces` (the `dora trace list` path) handles the root-less group; `handle_get_trace_spans` (the `dora trace view` path) returns the spans verbatim and the CLI then drops them. The two paths are inconsistent.

Realistic ways the returned set can lack the root / contain orphans:

- **Distributed traces.** dora is a coordinator ↔ multiple-daemon ↔ node system with shared `trace_id`s. A child span recorded by one component can reference a parent `span_id` created by another component, so the parent need not be in the coordinator's snapshot.
- **Bounded ring buffer eviction.** The span store is a fixed `VecDeque` of `MAX_SPANS = 4096` that evicts oldest-first (`libraries/extensions/telemetry/tracing/src/span_store.rs:31,53-58`). A trace whose span count approaches/exceeds the buffer, or interleaving with other traces, can leave a subset of a trace's spans in the store with dangling parent references.

In all of these, the user is shown an incomplete tree — or nothing — with no indication that spans were dropped.

## Suggested fix

Treat a span as a root when its `parent_span_id` is `None` **or** when its `parent_span_id` is not present among the returned spans, and surface a note that the tree is partial. Sketch:

```rust
let present: HashSet<u64> = spans.iter().map(|s| s.span_id).collect();
// roots: no parent, or a parent that isn't in this set (orphan)
let mut roots: Vec<&TraceSpan> = spans
    .iter()
    .filter(|s| s.parent_span_id.map_or(true, |p| !present.contains(&p)))
    .collect();
roots.sort_by_key(|s| s.start_time);
for root in roots {
    print_span(root, &children, 0);
}
```

(Optionally print a one-line warning when any orphan roots were promoted, so partial output is visible to the user.)

A secondary hardening note: `print_span` recurses on `children.get(&Some(span.span_id))` with no visited-set / depth guard, so a malformed span set with a parent-pointer cycle (e.g. two spans referencing each other, or a `span_id == parent_span_id`) would recurse until the stack overflows. Worth a cycle guard while in this area.

---
*This issue was created by an automated, scheduled Claude code-review check. The analysis is based on reading the source at the current `main` tip; please verify against your tracing setup before acting.*

phil-opp:
## Summary

Fixes #2301.

`print_span_tree` in `binaries/cli/src/command/trace/view.rs` only printed spans reachable from a span with `parent_span_id == None`. Any span whose parent was absent from the snapshot — due to ring-buffer eviction, a distributed trace spanning multiple components, or the root span simply not being returned — was silently dropped. If the snapshot contained no `None`-parent span at all, `dora trace view` printed nothing despite the early `spans.is_empty()` guard having passed.

## Changes

- **Orphan promotion**: collect all present `span_id`s into a `HashSet`; treat any span whose `parent_span_id` is not in that set (or is `None`) as a root so it appears in the output rather than being silently discarded.
- **Partial-trace note**: when orphan roots are promoted, emit a one-line `note:` to stderr so users know the tree is incomplete.
- **Cycle guard**: pass a `visited: HashSet<u64>` into `print_span` so a malformed span set with a parent-pointer cycle cannot overflow the stack.

## Test plan

- `cargo check -p dora-cli` ✅
- `cargo clippy -p dora-cli -- -D warnings` ✅

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_0168S4nQPLo1Nz5tEDhKkYTB

---
_Generated by [Claude Code](https://claude.ai/code/session_0168S4nQPLo1Nz5tEDhKkYTB)_

### DIFF
--- a/binaries/cli/src/command/trace/view.rs
+++ b/binaries/cli/src/command/trace/view.rs
@@ -78,7 +78,7 @@ fn print_span_tree(spans: &[TraceSpan]) {
     if partial {
         eprintln!(
             "note: trace is partial — some spans reference a parent that is absent from this \
-             snapshot or forms a cycle (shown as top-level roots)"
+             snapshot or form a cycle (shown as top-level roots)"
         );
     }
 
--- a/binaries/cli/src/command/trace/view.rs
+++ b/binaries/cli/src/command/trace/view.rs
@@ -89,7 +89,7 @@ fn print_span_tree(spans: &[TraceSpan]) {
     let present: HashSet<u64> = spans.iter().map(|s| s.span_id).collect();
     let mut roots: Vec<&TraceSpan> = spans
         .iter()
-        .filter(|s| s.parent_span_id.map_or(true, |p| !present.contains(&p)))
+        .filter(|s| s.parent_span_id.is_none_or(|p| !present.contains(&p)))
         .collect();
     roots.sort_by_key(|s| s.start_time);
 
--- a/binaries/cli/src/command/trace/view.rs
+++ b/binaries/cli/src/command/trace/view.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use clap::Args;
 use dora_message::{cli_to_coordinator::ControlRequest, coordinator_to_cli::TraceSpan};
@@ -82,14 +82,39 @@ fn print_span_tree(spans: &[TraceSpan]) {
         list.sort_by_key(|s| s.start_time);
     }
 
-    if let Some(roots) = children.get(&None) {
-        for root in roots {
-            print_span(root, &children, 0);
-        }
+    // Spans whose parent is absent from this snapshot (e.g. evicted from the
+    // ring buffer, or recorded by a different component) are "orphans".  Treat
+    // them as roots so they appear in the output rather than being silently
+    // dropped.
+    let present: HashSet<u64> = spans.iter().map(|s| s.span_id).collect();
+    let mut roots: Vec<&TraceSpan> = spans
+        .iter()
+        .filter(|s| s.parent_span_id.map_or(true, |p| !present.contains(&p)))
+        .collect();
+    roots.sort_by_key(|s| s.start_time);
+
+    if roots.iter().any(|r| r.parent_span_id.is_some()) {
+        eprintln!(
+            "note: trace is partial — some spans reference a parent not present in this \
+             snapshot (orphaned spans shown as top-level roots)"
+        );
+    }
+
+    let mut visited = HashSet::new();
+    for root in roots {
+        print_span(root, &children, 0, &mut visited);
     }
 }
 
-fn print_span(span: &TraceSpan, children: &HashMap<Option<u64>, Vec<&TraceSpan>>, depth: usize) {
+fn print_span(
+    span: &TraceSpan,
+    children: &HashMap<Option<u64>, Vec<&TraceSpan>>,
+    depth: usize,
+    visited: &mut HashSet<u64>,
+) {
+    if !visited.insert(span.span_id) {
+        return; // cycle guard: skip already-visited spans
+    }
     let indent = "  ".repeat(depth);
     let dur = format_duration_us(span.duration_us);
     let fields = if span.fields.is_empty() {
@@ -106,7 +131,7 @@ fn print_span(span: &TraceSpan, children: &HashMap<Option<u64>, Vec<&TraceSpan>>
 
     if let Some(kids) = children.get(&Some(span.span_id)) {
         for child in kids {
-            print_span(child, children, depth + 1);
+            print_span(child, children, depth + 1, visited);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** 🤖 _(fully automated; no human has vetted this)_

I reviewed the latest commits (through `15696d0`). The orphan-as-root handling and the `visited`-set cycle guard in `print_span` look correct: orphans (parent absent from the snapshot) are surfaced as top-level roots, and the shared `visited` set prevents stack overflow on parent-pointer cycles.

One edge case that this change does not cover: a pure cycle among spans that are *all present* in the snapshot (e.g. A.parent = B, B.parent = A, with both A and B in the set). Such spans never qualify as roots under the `is_none_or(|p| !present.contains(&p))` filter — neither parent is absent — so the entire cycle is still silently dropped, which is the same failure mode this PR sets out to fix. Probably acceptable since it only arises from malformed/corrupt span data, but worth noting if robustness against bad input is a goal here.

Separately: this PR also carries the unrelated `quinn-proto` 0.11.15 Cargo.lock bump, which is duplicated in #2304 and #2305. Whichever lands first will likely make the others' lockfile hunk a no-op/conflict — may be worth consolidating to a single PR.

---

phil-opp:
Good catch on the all-present cycle — addressed in e8157981d.

The orphan filter only promotes a span to a root when its parent is absent from the snapshot, so a cycle whose members are all present (A→B→A) had no qualifying root and was still dropped. The fix walks the real roots first, then surfaces any span left unvisited (only reachable through an all-present cycle) from its earliest member; the existing visited guard stops the recursion and prevents double-printing. The partial-trace note now fires for the cyclic case too.

Pulled the ordering into a pure plan_span_tree and added unit tests for the normal, orphan, all-present-cycle, and mixed (real tree + isolated cycle) cases.

Also rebased onto main and dropped the quinn-proto bump — 0.11.15 already landed there separately, so the lockfile hunk is gone and this PR is now scoped to the trace-view fix only.

Generated by claude

phil-opp:
@claude Please review again

phil-opp:
Re-reviewed at head `7665367`. The only change since my previous review is the Copilot autofix commit, which applies the `forms`→`form` grammar fix to the partial-trace note — i.e. exactly the one actionable item I had flagged. No behavior change, and no new findings on my end.

For the record, my earlier analysis still holds against the current code: orphan promotion (`is_none_or(|p| !present.contains(&p))`), the `visited`-set cycle guard (prevents both stack overflow and double-printing), and the `leftover` pass for all-present cycles (which sets `partial = true`) are all correct. Output ordering is deterministic via `start_time` sorts rather than HashSet iteration order. The remaining open Copilot thread (root-reachable cycle leaving `partial == false`, e.g. `1→2→3→2`) is not reachable under the single-`parent_span_id` data model, so it isn't a correctness gap. Tests cover the normal/orphan/cycle/mixed cases and would fail against the old `None`-parent-only logic.

---
🤖 *Automated review by Claude (Claude Code) — fully automated, no human has verified these findings. Treat as suggestions and verify before acting.*

---

phil-opp:
Thanks for catching the all-present cycle case — that was a real gap in the original implementation. The `plan_span_tree` / `collect_rows` refactor is much cleaner, and the unit tests for each failure mode (normal, orphan, all-present cycle, mixed) are a nice addition.

---

PR #2305, opened 2026-06-23
fix(deps): upgrade quinn-proto to 0.11.15 (RUSTSEC-2026-0185)

phil-opp:
## Summary

Upgrades `quinn-proto` from 0.11.14 to 0.11.15 to fix RUSTSEC-2026-0185.

**Vulnerability:** Remote memory exhaustion in quinn-proto from unbounded out-of-order stream reassembly (published 2026-06-22, CVSS 7.5 / high). A remote peer can send many out-of-order stream frames to exhaust the daemon's or coordinator's memory.

**Dependency chain:** `reqwest` → `quinn` → `quinn-proto`

**Fix:** `cargo update quinn-proto` — `quinn` 0.11 is semver-compatible with `quinn-proto` 0.11.15.

This is a Cargo.lock-only change; no code changes are required.

## Test plan

- `cargo audit` passes with quinn-proto 0.11.15 ✅
- No API changes — quinn-proto 0.11.14→0.11.15 is a patch release

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_0168S4nQPLo1Nz5tEDhKkYTB

---
_Generated by [Claude Code](https://claude.ai/code/session_0168S4nQPLo1Nz5tEDhKkYTB)_

PR #2308, opened 2026-06-23
fix(node-api): make close_outputs atomic on unknown output id

phil-opp:
## Issue

`DoraNode::close_outputs` ([`apis/rust/node/src/node/mod.rs`](https://github.com/dora-rs/dora/blob/main/apis/rust/node/src/node/mod.rs#L1023)) iterated over the requested ids and **removed each one from the local output set as it went**, bailing out on the first unknown id:

```rust
for output_id in &outputs_ids {
    if !self.node_config.outputs.remove(output_id) {   // mutates, then maybe errors
        return Err(NodeError::Output(format!("unknown output {output_id}")));
    }
}
self.control_channel.report_closed_outputs(outputs_ids)?;  // only on success
```

The daemon is notified **only on the success path** (`report_closed_outputs`). So for a batch like `[valid, unknown]`:

1. `valid` is removed from the node's local `outputs` set.
2. `unknown` triggers the early `return Err(...)`.
3. `report_closed_outputs` is **never called**.

The node now believes `valid` is closed while the daemon still believes it is open — and because the local set drives `validate_output`, any later `send_output(valid, ...)` is **silently dropped** instead of delivered.

## Fix

Validate the entire batch before mutating any state, making the operation atomic — it either closes every id and notifies the daemon, or fails leaving local state untouched:

```rust
for output_id in &outputs_ids {
    if !self.node_config.outputs.contains(output_id) {
        return Err(NodeError::Output(format!("unknown output {output_id}")));
    }
}
for output_id in &outputs_ids {
    self.node_config.outputs.remove(output_id);
}
self.control_channel.report_closed_outputs(outputs_ids)?;
```

## Validation

- Added regression test `close_outputs_is_atomic_on_unknown_id` exercising the partial-failure path (fails on the old code, passes on the new code).
- `cargo test -p dora-node-api` ✅

---

🤖 This PR was generated by Claude, an AI agent. The diff is machine-generated and was verified locally as described above; a human should still review before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RrWzN9zDS9okB4FBH49dF9)_

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -1031,10 +1031,18 @@ impl DoraNode {
     ///
     /// Closing outputs early can be helpful to receivers.
     pub fn close_outputs(&mut self, outputs_ids: Vec<DataId>) -> NodeResult<()> {
+        // Validate the whole batch before mutating any local state. Removing
+        // outputs eagerly would leave the node's local output set out of sync
+        // with the daemon if a later id is unknown: the early ones would be
+        // gone locally, yet `report_closed_outputs` is skipped on error so the
+        // daemon never learns about them.
         for output_id in &outputs_ids {
-            if !self.node_config.outputs.remove(output_id) {
+            if !self.node_config.outputs.contains(output_id) {
                 return Err(NodeError::Output(format!("unknown output {output_id}")));
             }
+        }
+        for output_id in &outputs_ids {
+            self.node_config.outputs.remove(output_id);
         }
 
         self.control_channel
@@ -1984,6 +1992,32 @@ mod tests {
         assert!(
             err.to_string().contains("does not match"),
             "unexpected error message: {err}"
+        );
+
+        drop(node);
+        drop(events);
+    }
+
+    /// `close_outputs` must be atomic: if any id in the batch is unknown, the
+    /// call fails *without* removing the valid ids from the local output set.
+    /// Otherwise the daemon (never notified, because `report_closed_outputs` is
+    /// skipped on error) and the node disagree about which outputs are open, and
+    /// the node would silently drop subsequent sends to a still-open output.
+    #[test]
+    fn close_outputs_is_atomic_on_unknown_id() {
+        let (mut node, events, _rx) = test_node();
+        let valid: DataId = "valid".into();
+        node.node_config.outputs.insert(valid.clone());
+
+        let result = node.close_outputs(vec![valid.clone(), "unknown".into()]);
+
+        assert!(
+            result.is_err(),
+            "closing a batch containing an unknown output must fail"
+        );
+        assert!(
+            node.node_config.outputs.contains(&valid),
+            "a failed close_outputs must not remove the valid output from local state"
         );
 
         drop(node);


phil-opp:
## 🤖 Automated review run summary

This is the summary for an automated codebase-review run by **Claude** (an AI agent). Each PR below is machine-generated; diffs were verified locally (fmt + clippy + affected-crate tests), and the combined diff additionally passed the full workspace test suite (`cargo test --all`, excluding Python/examples) plus the `/review` and `/simplify` gates. Please review before merging.

### PRs opened (by category)

**Correctness**
- **#2308** — `fix(node-api): make close_outputs atomic on unknown output id` (this PR). `DoraNode::close_outputs` removed outputs from the local set as it iterated, so a batch like `[valid, unknown]` removed `valid` locally but errored before notifying the daemon, leaving node/daemon state divergent and silently dropping later sends to `valid`. Now validates the whole batch before mutating any state.
- **#2309** — `fix(node-api): return error on send_output_bytes data_len mismatch`. A `data_len` that disagrees with `data.len()` panicked deep inside `copy_from_slice`; now returns a clear `NodeError::Output` like the rest of the node API.

### Findings investigated and rejected (with reasons)

- **serde_json `.unwrap()` in CLI** (`topic/pub_.rs`, `topic/echo.rs`, `inspect/top.rs`) — serializing `Parameter`/plain JSON values is infallible in practice; no realistic panic, and the unwrap-budget gate already governs this. Not a real bug.
- **Float precision in `parse_byte_size`** (`core/.../validate.rs:325`) — already guards non-finite/negative and uses an integer fast-path; precision loss only at petabyte-scale *fractional* sizes. Not worth a change.
- **`restart_count: u32` overflow** (`daemon/.../prepared.rs`) — would require ~4 billion restarts to wrap. Not realistic.
- **Memory-pool `ptr`/`size`/`shape` "truncation" via `as i64`** (`daemon/src/lib.rs:530–586`) — **false**: same-width `as` casts are bit-preserving, so `x as i64 as u64 == x` for all `u64`; the round-trip is lossless.
- **`MemoryPoolManager::cleanup_all` mis-counts releases** (`memory-pool/src/lib.rs:285`) — **false**: `released_count` is only reported in the `errors.is_empty()` branch where it's accurate; failures return `Err(errors)`.
- **`span_store.rs` `as_micros() as u64` truncation** — would need a single span lasting ~584,000 years. Theoretical only.
- **Coordinator topic-debug payload / daemon label-map size DoS** — speculative, low confidence; the payload behavior is intentional. Skipped.

### Scope notes

The `Audit (cargo-audit + cargo-deny)` CI job fails on **both** PRs for the same **pre-existing, repo-wide** RUSTSEC advisories (e.g. unmaintained `paste`/`proc-macro-error2`, plus the advisory tracked by #2305). Both PRs are dependency-free, so this failure is unrelated to them; the relevant gates (fmt, clippy, tests) pass.

---
*Generated by Claude. Branches are suffixed off a freshly-fetched `origin/main`.*

---

phil-opp:

I reviewed the diff and found no issues. The fix makes `close_outputs` atomic: it validates the entire batch with `contains` before removing anything, so a batch like `[valid, unknown]` now fails without mutating local state. This correctly prevents the prior bug where `valid` was removed locally but the daemon was never notified (because `report_closed_outputs` is skipped on the error path), leaving node and daemon disagreeing about open outputs. The regression test covers the partial-failure path. Looks good.

---

phil-opp:
Rebased onto the latest `main` (now includes #2309 and #2310) to clear the merge conflict that bumped this from the queue. Only the `close_outputs` change remains in the diff; the `send_output_bytes` test that previously overlapped is now in `main` via #2309. Verified locally on the rebased branch: `cargo fmt --all -- --check`, `cargo clippy -p dora-node-api -- -D warnings`, and `cargo test -p dora-node-api` all pass. Ready to re-submit to the merge queue.

*(automated note from Claude)*

---

PR #2310, opened 2026-06-24
refactor(message): remove unused DaemonId::load_or_create persistence helper

phil-opp:
> ⚠️ **Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code-removal routine, not by a human contributor. Please review accordingly and close if the analysis is wrong.

## What

Removes `DaemonId::load_or_create` from `libraries/message/src/common.rs`, along with its two unit tests (`daemon_id_load_or_create_persists`, `daemon_id_load_or_create_different_dirs`).

```rust
/// Load a persisted daemon ID from `<working_dir>/.daemon-id`, or create and persist a new one.
pub fn load_or_create(working_dir: &Path, machine_id: Option<String>) -> std::io::Result<Self>
```

## Why it's dead

- **No production caller.** A workspace-wide search (`rg 'load_or_create'`, `rg '\.daemon-id'`) finds the method only at its definition site and in its own two unit tests. Nothing reads or writes a `.daemon-id` file anywhere else.
- **The live path uses `DaemonId::new`.** The actual daemon-id is constructed via `DaemonId::new(machine_id)` (e.g. `binaries/coordinator/src/lib.rs:419`), which mints a fresh v7 UUID. The persist-to-disk variant was never wired into that path.
- This is an internal protocol type. `DaemonId` identifies daemons/coordinators on the wire; downstream node authors never construct it, so `load_or_create` is **not** a public extension point — it's an unwired internal helper.

## Verification

- `cargo clippy -p dora-message --all-targets -- -D warnings` — clean
- `cargo test -p dora-message` — passes
- `cargo fmt -p dora-message -- --check` — clean

## Note for maintainers

If persistent daemon identity across restarts is **planned** (this looks related to the daemon-reconnect work), feel free to close this — the helper would be the natural building block for that feature. As of today it has no callers, so the dead-code sweep flagged it.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01UjweRVzqRE2krY2a9mKJfJ)_

### DIFF
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -281,22 +281,6 @@ impl DaemonId {
         }
     }
 
-    /// Load a persisted daemon ID from `<working_dir>/.daemon-id`, or create and persist a new one.
-    pub fn load_or_create(
-        working_dir: &std::path::Path,
-        machine_id: Option<String>,
-    ) -> std::io::Result<Self> {
-        let path = working_dir.join(".daemon-id");
-        if let Ok(contents) = std::fs::read_to_string(&path)
-            && let Ok(uuid) = contents.trim().parse::<Uuid>()
-        {
-            return Ok(DaemonId { machine_id, uuid });
-        }
-        let id = Self::new(machine_id);
-        std::fs::write(&path, id.uuid.to_string())?;
-        Ok(id)
-    }
-
     pub fn matches_machine_id(&self, machine_id: &str) -> bool {
         self.machine_id
             .as_ref()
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed the diff and found no issues. I confirmed `DaemonId::load_or_create` has no production callers anywhere in the workspace — the live daemon-id path uses `DaemonId::new`, and the only references to `load_or_create`/`.daemon-id` outside this method are its own two unit tests, docs (`docs/architecture.md`, `guide/`), and `.gitignore`. Since `DaemonId` is internal daemon/coordinator protocol infrastructure (not a downstream extension point), removing the dead method and its tests is safe.

(Minor, non-blocking: the `.daemon-id` references in `docs/architecture.md` / `guide/` and `.gitignore` are now stale, but they predate this PR.)

---

PR #2334, opened 2026-06-25
fix(core): normalize inner-node path before containment check to block `..` traversal

### LINKED ISSUE #2333: Module expansion: inner-node path containment check bypassable via `..` (lexical `strip_prefix`)
> **Note:** This issue was created by an automated scheduled Claude code-review check. It reviews a randomly selected component each run. Please verify the analysis before acting on it.

## Summary

In `libraries/core/src/descriptor/expand.rs`, the path-resolution step inside `expand_module_node` tries to keep an inner module node's `path:` inside the project directory, but the containment check is purely **lexical** and does not normalize `..` components. A relative inner-node path containing `..` (e.g. `../../etc/evil-binary`) passes the check and resolves to a location **outside** the project root — the exact condition the check claims to reject.

This is inconsistent with how the *module file* path itself (lines 285–295) and *nested module* paths in `check_module_file` (lines 215–228) are validated — both `canonicalize()` before the `starts_with` check, which correctly blocks traversal. Only the inner-node path resolution skips canonicalization.

## Location

`libraries/core/src/descriptor/expand.rs`, lines ~380–395:

```rust
// Resolve relative paths: make inner node paths relative to base_dir
if let Some(ref path) = inner_node.path
    && !super::source_is_url(path)
    && !Path::new(path).is_absolute()
{
    let resolved = module_dir.join(path);
    let relative = resolved.strip_prefix(canonical_base).map_err(|_| {
        eyre::eyre!(
            "module node `{}` path `{}` resolves outside the project \
                 directory (resolved to `{}`)",
            inner_node.id,
            path,
            resolved.display()
        )
    })?;
    inner_node.path = Some(relative.to_string_lossy().into_owned());
}
```

`resolved = module_dir.join(path)` is a lexical join (no canonicalization — the node binary may not be built yet, so the file need not exist). `Path::strip_prefix` is **component-based and lexical**: it succeeds whenever the path literally begins with the prefix components, regardless of any `..` that follow. So `..`-escaping paths slip through, and the stored node path still contains `..`, which the daemon later resolves against the working dir — landing outside the project.

## Reproduction

Standalone demonstration of the `strip_prefix` behavior the code relies on:

```rust
use std::path::Path;
fn main() {
    let canonical_base = Path::new("/project");
    let module_dir = Path::new("/project/sub");   // module file lives here
    let inner_path = "../../etc/evil-binary";       // inner node's `path:`
    let resolved = module_dir.join(inner_path);     // lexical join, no canonicalize
    println!("{:?}", resolved.strip_prefix(canonical_base));
    // => Ok("sub/../../etc/evil-binary")  -- check SUCCEEDS, no error raised
}
```

Output:

```
strip_prefix SUCCEEDED -> stored node path = "sub/../../etc/evil-binary"
daemon resolves to (lexically): "/project/sub/../../etc/evil-binary"   (= /etc/evil-binary)
```

The guard's error message — *"resolves outside the project directory"* — never fires, even though the path provably resolves outside.

## Impact

- The intended containment boundary for inner-node executable paths is not enforced. A module file (potentially third-party / hub-shipped) can point an inner node at an executable outside the project tree by using a `..`-relative path, while the surrounding code reads as though this is prevented.
- Severity is moderate rather than critical because node `path:` fields are user-authored in ordinary dataflows; the concern is specifically the *broken-but-claimed* security boundary for module-supplied node paths, and the inconsistency with the sibling checks that do canonicalize.

## Suggested fix

Lexically normalize `resolved` (collapse `.` / resolve `..` without touching the filesystem, as `manifest/inject.rs::normalize` already does) **before** the `strip_prefix`/containment check, or explicitly reject any `resolved` whose normalized form is not still prefixed by `canonical_base`. Filesystem `canonicalize()` is not appropriate here because the binary may not exist yet, so a lexical normalizer is the right tool — mirroring the approach in `libraries/core/src/manifest/inject.rs`.

A regression test should cover an inner node with `path: ../../something` and assert that expansion errors with the "resolves outside the project directory" message.

phil-opp:
Closes #2333

## Summary

The `strip_prefix` containment guard in `expand_module_node` was purely lexical and did not collapse `..` path components. A module inner node with `path: ../../etc/evil-binary` would pass the check even though it resolves outside the project root — the exact condition the check claims to prevent.

**Root cause:** `module_dir.join(path)` is a lexical join. `Path::strip_prefix` is also component-based and lexical, so it succeeds whenever the path *starts* with the prefix components regardless of any `..` that follow. For example, `/project/sub/../../etc/evil` does strip `/project` prefix successfully, returning `sub/../../etc/evil`.

**Fix:** Normalize the joined path lexically (collapse `.` and resolve `..` without touching the filesystem) before `strip_prefix`. Uses the same approach as `libraries/core/src/manifest/inject.rs::normalize`. Filesystem `canonicalize()` is intentionally avoided because the binary may not exist yet at expansion time.

## Changes

- `libraries/core/src/descriptor/expand.rs`: add `normalize_path` helper; apply it to `resolved` before the `strip_prefix` containment check
- Add regression test `reject_inner_node_path_traversal_via_dotdot`

## Test

```
cargo test -p dora-core -- reject_inner_node_path_traversal_via_dotdot
```

---

### DIFF
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -7,8 +7,24 @@ use eyre::{Context, bail};
 use serde::Deserialize;
 use std::{
     collections::{BTreeMap, BTreeSet, HashSet},
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
 };
+
+/// Lexically normalize a path (collapse `.` and resolve `..`) without touching
+/// the filesystem — the executable may not be built yet when this is called.
+fn normalize_path(path: &Path) -> PathBuf {
+    let mut out = PathBuf::new();
+    for component in path.components() {
+        match component {
+            Component::CurDir => {}
+            Component::ParentDir => {
+                out.pop();
+            }
+            other => out.push(other),
+        }
+    }
+    out
+}
 
 /// Check if a path string is absolute on any platform.
 /// On Windows, `Path::is_absolute()` returns false for Unix-style `/foo` paths,
@@ -376,12 +392,15 @@ fn expand_module_node(
         }
         inner_node.inputs = new_inputs;
 
-        // Resolve relative paths: make inner node paths relative to base_dir
+        // Resolve relative paths: make inner node paths relative to base_dir.
+        // Normalize lexically (collapse `..`) before the containment check so
+        // that paths like `../../evil` are detected even though the binary may
+        // not exist yet (ruling out filesystem canonicalize).
         if let Some(ref path) = inner_node.path
             && !super::source_is_url(path)
             && !Path::new(path).is_absolute()
         {
-            let resolved = module_dir.join(path);
+            let resolved = normalize_path(&module_dir.join(path));
             let relative = resolved.strip_prefix(canonical_base).map_err(|_| {
                 eyre::eyre!(
                     "module node `{}` path `{}` resolves outside the project \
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the path-normalization fix — no issues found.

- `normalize_path` is applied to `module_dir.join(path)` before `strip_prefix(canonical_base)`, which correctly closes the lexical `strip_prefix` bypass (trailing `..` after a matching prefix used to slip through).
- Symlink-safe: both `canonical_base` and `module_dir` are already canonicalized, so lexical `..` resolution over the base portion matches physical resolution. Skipping `canonicalize()` on the resolved path is the right call since the binary may not exist at expansion time.
- Root / excess-`..` edge case holds: `PathBuf::pop()` won't ascend above `/`, so over-long `../` chains normalize to an absolute path that still fails the containment check rather than wrapping around.
- The helper matches the existing `manifest/inject.rs::normalize`, and the regression test covers the documented attack.

---

---

PR #2335, opened 2026-06-25
fix(core): propagate outer module build through nested module expansion

### LINKED ISSUE #2311: Module-level `build` is silently dropped for nodes coming from nested modules

## Summary

In `libraries/core/src/descriptor/expand.rs`, a module file's top-level `build` command is silently lost for any node that is produced through a **nested** module reference. The build step is correctly applied to direct (flat) inner nodes, but not to leaf nodes reached via a nested module. `deploy` does *not* have this problem, which makes the asymmetry clear.

## Where

`expand_module_node` does its work in phases. Phase 1 prepends the module-level build to **every** inner node, including nodes that are themselves module references:

```rust
// expand.rs:407-414
// Prepend module-level build command to inner node builds
if let Some(ref module_build) = module_file.build {
    let inner_build = inner_node.build.take();
    inner_node.build = Some(match inner_build {
        Some(existing) => format!("{module_build}\n{existing}"),
        None => module_build.clone(),
    });
}
```

Phase 2 then re-expands nodes that are module references via a recursive call:

```rust
// expand.rs:424-434
for inner_node in prefixed_nodes {
    if inner_node.module.is_some() {
        let nested_id = inner_node.id.to_string();
        let (nested, nested_omap) =
            expand_module_node(&inner_node, module_dir, canonical_base, depth + 1, seen)?;
        nested_output_maps.insert(nested_id, nested_omap);
        final_nodes.extend(nested);   // <-- inner_node.build is never read
    } else {
        final_nodes.push(inner_node);
    }
}
```

The recursive `expand_module_node` consumes `node.module`, `node.inputs`, `node.params`, and `node.deploy`, but **never reads `node.build`**. The returned `nested` nodes are constructed fresh from the nested module file. So the outer module's build command — just attached to the nested-module node at lines 408–414 — is thrown away.

## Why `deploy` is fine but `build` is not

`deploy` is propagated at lines 398–400 and is **re-applied inside the recursion** (the same lines run again in the nested call, propagating the nested-module node's `deploy` — which was set from the outer node — down to the nested leaf nodes). `build` has no equivalent re-application in the recursion, so it does not survive nesting.

## Reproduction (conceptual)

Outer module `A.yml`:
```yaml
module:
  name: outer
build: pip install -r requirements.txt   # should apply to all of A's nodes
nodes:
  - id: inner
    module: B.yml                          # nested module reference
```

After expansion, the leaf nodes produced by `B.yml` do **not** get `pip install -r requirements.txt` prepended to their build. The dependency-install step silently vanishes for exactly those nodes.

## Impact

Severity: **Medium**. No panic and no mis-wiring, but a declared build step is silently omitted for nested-module nodes. This typically surfaces much later as a confusing build/runtime failure (missing dependencies) that is hard to trace back to the descriptor, because the YAML clearly declares the build.

## Suggested fix direction

In phase 2, after obtaining the `nested` nodes, prepend `inner_node.build` (the build accumulated on the module-reference node) to each returned node's build — mirroring how `deploy` is re-applied through the recursion. A regression test combining `expand_nested_modules` with a module-level `build` (i.e. an outer module that both declares `build:` and references a nested module) would lock this in; the two features are currently tested only in isolation (`expand_nested_modules`, `expand_module_build_prepended`).

phil-opp:
Closes #2311

## Summary

When a module file references a nested module, the outer module's `build:` command was silently dropped for the resulting leaf nodes.

**Root cause:** In `expand_module_node`, phase 1 correctly prepends the outer module's `build` to every inner node (including nested-module reference nodes). But in phase 2, when a nested-module node is recursively expanded, the function only passes `inner_node` to the recursive call. The recursive call constructs fresh leaf nodes from the nested module file and never reads `inner_node.build`. The accumulated outer build is thrown away.

`deploy` did not have this bug because it is re-applied inside the recursion (propagated down at lines 398–400, which run again in the nested call).

**Fix:** After the recursive `expand_module_node` returns `nested` nodes, prepend `inner_node.build` (the accumulated outer build) to each returned node's `build` — mirroring how `deploy` flows through recursion.

## Changes

- `libraries/core/src/descriptor/expand.rs`: propagate `accumulated_build` from the outer module-reference node to each leaf node returned by the recursive expansion
- Add regression test `expand_nested_module_build_propagated`

## Test

```
cargo test -p dora-core -- expand_nested_module_build_propagated
```

---

### DIFF
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -424,8 +424,20 @@ fn expand_module_node(
     for inner_node in prefixed_nodes {
         if inner_node.module.is_some() {
             let nested_id = inner_node.id.to_string();
-            let (nested, nested_omap) =
+            let accumulated_build = inner_node.build.clone();
+            let (mut nested, nested_omap) =
                 expand_module_node(&inner_node, module_dir, canonical_base, depth + 1, seen)?;
+            // Propagate the outer module's accumulated build to each nested leaf node,
+            // mirroring how `deploy` is propagated through recursion.
+            if let Some(ref outer_build) = accumulated_build {
+                for nested_node in &mut nested {
+                    let inner_build = nested_node.build.take();
+                    nested_node.build = Some(match inner_build {
+                        Some(existing) => format!("{outer_build}\n{existing}"),
+                        None => outer_build.clone(),
+                    });
+                }
+            }
             nested_output_maps.insert(nested_id, nested_omap);
             final_nodes.extend(nested);
         } else {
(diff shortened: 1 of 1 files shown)


phil-opp:

I traced `expand_module_node`'s recursion for 1-, 2-, and 3-level nesting. The recursive call applies only the *nested* module file's own `build` in phase 1, so the outer module's accumulated build is never applied inside the recursion — the new post-recursion prepend applies it exactly once per level (accumulating as `outer\ninner` for deeper nests). None/Some handling and prepend ordering match the existing phase-1 convention.

The regression test `expand_nested_module_build_propagated` exercises the bug correctly. No correctness, edge-case, or concurrency issues found.

---

---

PR #2336, opened 2026-06-25
fix(daemon): restore newlines in Indent log formatter for multi-line messages

### LINKED ISSUE #2312: `Indent` log formatter collapses multi-line messages onto one line (drops newlines)

## Summary

The `Indent` `Display` adapter in `binaries/daemon/src/log.rs` strips the line terminators out of multi-line log messages and writes the lines back-to-back with no separator. Any forwarded node log that contains newlines (notably Python tracebacks) is mangled into a single line when emitted through the `Tracing` log destination.

## Where

```rust
// binaries/daemon/src/log.rs:531-540
struct Indent<'a>(&'a str);

impl std::fmt::Display for Indent<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        for line in self.0.lines() {
            write!(f, "   {line}")?;
        }
        Ok(())
    }
}
```

`str::lines()` yields each line **without** its trailing `\n` (or `\r\n`), and the loop writes no separator between iterations. So the line structure is destroyed.

This adapter wraps every message routed to the `Tracing` destination — see the `tracing::{error,warn,info,debug}!(..., "{}", Indent(&message.message))` calls around `log.rs:372`, `386`, `399`, `412`, `425`.

## Behavior vs. expected

For input `"Traceback (most recent call last):\n  File \"x.py\", line 3\nValueError: boom"`:

- **Current output:** `   Traceback (most recent call last):  File "x.py", line 3ValueError: boom` — one line, internal newlines erased, and the lines run together with no gap.
- **Expected:** each line indented and preserved on its own line, e.g.
  ```
     Traceback (most recent call last):
       File "x.py", line 3
     ValueError: boom
  ```

A trailing newline is also silently dropped (`"line1\n"` → `"   line1"`), since `lines()` does not yield a final empty element.

## Impact

Severity: **Low** (cosmetic — no panic, no data loss to disk, no mis-routing). But it meaningfully degrades debuggability: Python nodes routinely emit multi-line tracebacks, and those are exactly the messages this formatter mangles, on the fallback `Tracing` path.

## Suggested fix direction

Re-emit a newline after each indented line, e.g. `writeln!(f, "   {line}")?;` (optionally trimming a single trailing newline if the doubled blank line is undesirable), so the original line structure is preserved while still indenting each line. A small unit test on `Indent("a\nb").to_string()` would guard the behavior.

phil-opp:
Closes #2312

## Summary

The `Indent` `Display` adapter in `binaries/daemon/src/log.rs` used `write!` (no trailing newline) for each line from `str::lines()`. Since `lines()` strips line terminators, the loop wrote all lines back-to-back with no separator, collapsing multi-line messages into a single unreadable line.

This affected all node log messages routed to the `Tracing` destination, which wraps every message in `Indent(...)`. Python tracebacks (the most common multi-line messages) were the most visibly impacted.

**Fix:** Change `write!(f, "   {line}")` to `writeln!(f, "   {line}")` so each line is indented and properly terminated.

## Changes

- `binaries/daemon/src/log.rs`: `Indent::fmt` — use `writeln!` instead of `write!`
- Add unit tests: multi-line, single-line, and empty-string cases

## Test

```
cargo test -p dora-daemon indent
```

---

### DIFF
--- a/binaries/daemon/src/log.rs
+++ b/binaries/daemon/src/log.rs
@@ -533,7 +533,7 @@ struct Indent<'a>(&'a str);
 impl std::fmt::Display for Indent<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         for line in self.0.lines() {
-            write!(f, "   {line}")?;
+            writeln!(f, "   {line}")?;
         }
         Ok(())
     }
@@ -695,5 +695,23 @@ mod tests {
 
         // Should not error even with no files
         rotate_log_files(tmp.path(), &uuid, &node, 5).unwrap();
+    }
+
+    #[test]
+    fn indent_preserves_newlines_between_lines() {
+        let out = Indent("line1\nline2\nline3").to_string();
+        assert_eq!(out, "   line1\n   line2\n   line3\n");
+    }
+
+    #[test]
+    fn indent_single_line_gets_newline() {
+        let out = Indent("hello").to_string();
+        assert_eq!(out, "   hello\n");
+    }
+
+    #[test]
+    fn indent_empty_string_produces_empty() {
+        let out = Indent("").to_string();
+        assert_eq!(out, "");
     }
 }


phil-opp:

The fix is correct: `str::lines()` strips line terminators, so the old `write!` collapsed multi-line node logs (e.g. Python tracebacks) into a single line; switching to `writeln!` restores per-line separation. The three unit tests are accurate, including the empty-string case (`"".lines()` yields no iterations → empty output).

One minor, non-blocking note: `writeln!` also appends a trailing newline to single-line messages. Since `Indent(...)` is the message body of the `tracing::{info,error,…}!` calls and the tracing formatter adds its own record-terminating newline, single-line logs in the `Tracing` destination will now render with a trailing blank line. That's a reasonable trade-off for fixing the multi-line collapse and is purely cosmetic.

---

---

PR #2350, opened 2026-06-25
fix: correct debug_assert in managed_python_env_dir to detect .. path traversal

### LINKED ISSUE #2332: debug_assert in managed_python_env_dir fails to detect `..` path traversal (defense-in-depth is ineffective)
## Summary

The defense-in-depth `debug_assert!` added to `managed_python_env_dir` in **PR #2294** (commit `2b0aee9`, `fix(message): reject node ids starting with '.' to prevent path traversal`) does **not** detect the parent-directory traversal it was written to catch. It fires only for the harmless `.` case and silently passes for `..` and `../../etc`.

## Affected code

`libraries/core/src/build/mod.rs`, `managed_python_env_dir`:

```rust
let envs_base = node_working_dir.join(".dora").join("python-envs");
let env_dir = envs_base.join(node.id.as_ref());
// Defense-in-depth: the node id must not escape the python-envs/ directory.
debug_assert!(
    env_dir.components().count() > envs_base.components().count(),
    "node id '{id}' escaped python-envs directory (env_dir={env_dir:?}, base={envs_base:?})",
    id = node.id,
);
env_dir
```

## The bug

`Path::components()` does **not** normalize `..` — a `..` segment is yielded as a `ParentDir` component, which *increases* the component count rather than logically removing the preceding one. So a traversing id makes `env_dir.components().count() > envs_base.components().count()` **true**, and the assertion passes — exactly when an escape occurs.

Empirical check (base = `/work/.dora/python-envs`, count = 4):

| node id | env_dir | env count | `env > base`? | assertion |
|---|---|---|---|---|
| `..` | `/work/.dora/python-envs/..` | 5 | true | **passes (miss)** |
| `../../etc` | `/work/.dora/python-envs/../../etc` | 7 | true | **passes (miss)** |
| `.` | `/work/.dora/python-envs/.` | 4 | false | fires (but `.` doesn't escape) |
| `ok` | `/work/.dora/python-envs/ok` | 5 | true | passes (correct) |

The only id the assertion catches (`.`) is the one that does **not** escape (`CurDir` components are dropped by `components()`, so the path stays in place). Every id that actually escapes via `..` slips through. The check is effectively inverted relative to its stated intent.

## Risk

Low in practice. The real guard is `validate_node_id`, which #2294 correctly extended to reject any leading-dot id (`.`, `..`, `.hidden`), and that guard is enforced on every `NodeId` construction path (`Deserialize`, `FromStr`, `From<String>`). The `debug_assert` is only a debug-build, defense-in-depth layer.

The concern is that both the commit message and the in-code comment claim this assertion "guards against future callers that bypass validation" — it does not, for the dangerous `..` case. It provides false confidence: a future code path that constructs an env dir from an unvalidated id containing `..` would pass this check and then `uv venv --clear` could wipe a parent directory (the original #2293 failure mode).

## Suggested fix

Make the containment check actually reject `..`. Options:

1. Explicitly reject any `ParentDir` component in the joined path:
   ```rust
   debug_assert!(
       !env_dir.components().any(|c| matches!(c, std::path::Component::ParentDir)),
       "node id '{id}' contains a parent-dir segment (env_dir={env_dir:?})",
       id = node.id,
   );
   ```
2. Or verify the (lexically normalized) `env_dir` still starts with `envs_base`.

Option 1 is the minimal change that matches the stated intent.

---
*This issue was created by a scheduled automated Claude code-review check (review of merged PRs/commits from the last 2 days). The finding was verified empirically but please confirm before acting.*

Supull:
Fixes #2332

The `debug_assert!` in `managed_python_env_dir` used `components().count()` to detect `..` path traversal, but this is ineffective — `..` adds a `ParentDir` component, increasing the count and causing the assertion to pass silently for the dangerous case while firing for the harmless `.` case.

Fix: replace the count check with an explicit `ParentDir` check using `.any()`.

### DIFF
--- a/libraries/core/src/build/mod.rs
+++ b/libraries/core/src/build/mod.rs
@@ -344,8 +344,8 @@ pub fn managed_python_env_dir(node: &ResolvedNode, node_working_dir: &Path) -> O
         // so this assertion should never fire in practice — it guards against future
         // callers that bypass validation.
         debug_assert!(
-            env_dir.components().count() > envs_base.components().count(),
-            "node id '{id}' escaped python-envs directory (env_dir={env_dir:?}, base={envs_base:?})",
+            !env_dir.components().any(|c| matches!(c, std::path::Component::ParentDir)),
+            "node id '{id}' contains a parent-dir segment (env_dir={env_dir:?})",
             id = node.id,
         );
         env_dir
--- a/libraries/core/src/build/mod.rs
+++ b/libraries/core/src/build/mod.rs
@@ -344,7 +344,9 @@ pub fn managed_python_env_dir(node: &ResolvedNode, node_working_dir: &Path) -> O
         // so this assertion should never fire in practice — it guards against future
         // callers that bypass validation.
         debug_assert!(
-            !env_dir.components().any(|c| matches!(c, std::path::Component::ParentDir)),
+            !env_dir
+                .components()
+                .any(|c| matches!(c, std::path::Component::ParentDir)),
             "node id '{id}' contains a parent-dir segment (env_dir={env_dir:?})",
             id = node.id,
         );
--- a/libraries/core/src/build/mod.rs
+++ b/libraries/core/src/build/mod.rs
@@ -344,10 +344,10 @@ pub fn managed_python_env_dir(node: &ResolvedNode, node_working_dir: &Path) -> O
         // so this assertion should never fire in practice — it guards against future
         // callers that bypass validation.
         debug_assert!(
-            !env_dir
+            !Path::new(node.id.as_ref())
                 .components()
                 .any(|c| matches!(c, std::path::Component::ParentDir)),
-            "node id '{id}' contains a parent-dir segment (env_dir={env_dir:?})",
+            "node id '{id}' contains a parent-dir segment",
             id = node.id,
         );
         env_dir


phil-opp:
🤖 _This is a fully automated review by Claude. No human has reviewed these comments — please verify before acting on them._

I reviewed the diff against the surrounding code. The fix correctly addresses the real bug: `PathBuf::join` doesn't normalize, so a `..` node id produces `env_dir = .../python-envs/..`, which has *more* components than `envs_base` — meaning the old `count >` check passed silently for exactly the dangerous case. Replacing it with a `ParentDir` scan is the right idea.

One issue with the new check, though: it scans the **entire `env_dir`**, but `env_dir = node_working_dir.join(".dora/python-envs").join(node.id)`. The `node_working_dir` / `base_working_dir` prefix is caller-supplied and not canonicalized — it routinely derives from the dataflow descriptor's location, which can legitimately be a relative path containing `..` (e.g. `dora run ../foo/dataflow.yml`). In that case `env_dir.components()` contains a `ParentDir` that has nothing to do with the node id, so in debug builds the assertion fires with a message that misattributes the cause:

```rust
"node id '{id}' contains a parent-dir segment (env_dir={env_dir:?})"
```

Since `node.id` is the only attacker-influenced, validation-relevant segment here (and the comment + message both describe checking the *node id*), the check would be more precise applied to `node.id` alone — e.g. assert that `Path::new(node.id.as_ref())` contains no `ParentDir` component, or that `env_dir` starts with `envs_base` and the remainder has no `..`. As written, the check is over-scoped relative to what it claims to guard.

Note this is debug-only defense-in-depth (`debug_assert!` compiles out in release), and `validate_node_id` already rejects `..`/`.` ids today, so it can't fire from a real node id — the concern is purely the spurious/misleading debug panic from a relative working-dir prefix. So: not a merge blocker, but worth tightening since the whole point of the PR is to make this assertion correct.

No other issues found.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; please verify independently before relying on it.

Re-reviewing after the latest commit (`737eb35`), which scopes the `ParentDir` check to `node.id` instead of the full `env_dir` path. This resolves the concern from the earlier review: scanning the whole `env_dir` would also see the caller-supplied working-dir prefix, so a legitimate invocation like `dora run ../foo/dataflow.yml` could embed a `..` component and trip the debug assert spuriously.

I checked the current diff against `validate_node_id`: node ids are restricted to `[a-zA-Z0-9_.-]` with no path separators and no leading `.`, so the only value that yields a `ParentDir` component is the literal `..` — exactly the case the assert is meant to catch (`.` maps to `CurDir` and is correctly ignored). The logic is sound.

No issues found in the latest revision; the change looks safe to merge.

---

phil-opp:
Thanks!

Supull:
> Thanks!

No problem!

PR #2352, opened 2026-06-25
fix(core): reject byte sizes that overflow u64 in parse_byte_size

phil-opp:

## Issue

`parse_byte_size` in `libraries/core/src/descriptor/validate.rs` has an integer fast-path (which uses `checked_mul`) and an f64 fallback for numeric parts that don't fit in `u64`. The f64 branch guarded negative and non-finite inputs but **not** finite-but-overflowing positive values:

```rust
if !num.is_finite() || num < 0.0 { bail!(...); }
Ok((num * multiplier as f64) as u64)   // <- saturates on overflow
```

Casting an out-of-range positive `f64` to `u64` **saturates to `u64::MAX`** instead of erroring. So an absurd value like `max_log_size: "99999999999999999999GB"` (20 digits — too large for `u64`, so it takes the f64 path) was silently clamped to the maximum rather than rejected.

The sibling `ByteSize::from_str` in `dora-message` (`libraries/message/src/config.rs`) already has this exact guard (`bytes >= usize::MAX as f64`) with a regression test (`byte_size_rejects_overflow`); `validate.rs` was simply missing the matching check. The existing `parse_byte_size_rejects_integer_overflow` test only exercises the integer path.

## Fix

Add the `bytes >= u64::MAX as f64` guard before the cast. `u64::MAX as f64` rounds up to 2^64 and no f64 values exist between `u64::MAX` and 2^64, so `>=` rejects exactly the products that overflow `u64` without falsely rejecting any in-range value. A new `parse_byte_size_rejects_float_overflow` test covers the GB-multiplier and bare-`B` overflow cases.

## Validation

- `cargo test -p dora-core --lib parse_byte_size` — 14 passed (incl. the new test)
- `cargo clippy -p dora-core -- -D warnings` — clean

---

### DIFF
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -322,7 +322,17 @@ fn parse_byte_size(s: &str) -> eyre::Result<u64> {
     if !num.is_finite() || num < 0.0 {
         bail!("byte size must be a non-negative, finite number: '{s}'");
     }
-    Ok((num * multiplier as f64) as u64)
+    let bytes = num * multiplier as f64;
+    // A finite product can still exceed u64::MAX (e.g. "99999999999999999999GB"),
+    // and casting an out-of-range f64 to u64 saturates to u64::MAX instead of
+    // erroring, silently turning an absurd limit into the maximum. `u64::MAX as
+    // f64` rounds up to 2^64 and no f64 values exist between u64::MAX and 2^64,
+    // so `>=` rejects exactly the products that overflow u64. Mirrors the guard
+    // in `ByteSize::from_str` (dora-message).
+    if bytes >= u64::MAX as f64 {
+        bail!("byte size '{s}' overflows u64");
+    }
+    Ok(bytes as u64)
 }
 
 fn parse_log_level(s: &str) -> eyre::Result<dora_message::common::LogLevelOrStdout> {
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed the overflow fix in `parse_byte_size`. No issues found.**

The guard is correct and complete:
- The product `num * multiplier as f64` is computed after `num` is already validated as finite and non-negative; f64 multiplication can only saturate to `+inf`, and `+inf >= u64::MAX as f64` is `true`, so even an infinite product is rejected — no pre-check overflow escape.
- `u64::MAX as f64` rounds to exactly 2^64 and no f64 is representable strictly between `u64::MAX` (2^64−1) and 2^64, so `>=` rejects exactly the products that would saturate on the `as u64` cast, with no false rejection of in-range values.
- All multipliers defined in this function (B, KB/K, MB/M, GB/G) are covered, including the bare-`B` (multiplier 1) case.
- The new `parse_byte_size_rejects_float_overflow` test genuinely exercises the f64 path (20-digit integer, explicit `.0` decimal, and a value just above 2^64), distinct from the existing integer-path test.

Minor (description only, non-blocking): the PR body mentions "KB/MB/GB/TB/PB…" but the unit match only defines up to GB — TB/PB hit the `unknown byte size unit` bail arm. The code is fine; the description just overstates the unit set.

---

PR #2353, opened 2026-06-25
perf(node): avoid per-send String allocation in compute_schema_hash

phil-opp:

## Issue

`compute_schema_hash` (`apis/rust/node/src/node/arrow_utils.rs`) computed `format!("{data_type:?}")` and then ran FNV-1a over the resulting bytes:

```rust
let s = format!("{data_type:?}");
let mut hash: u64 = 0xcbf29ce484222325;
for byte in s.bytes() { hash ^= byte as u64; hash = hash.wrapping_mul(0x100000001b3); }
```

It is called once **per array node — recursively over every child** — inside `copy_array_into_sample_inner`, which runs on every `send_output` (the default Raw framing path, and also the IPC path in `node/mod.rs`). So a nested struct/list array incurs one heap allocation per buffer-tree node on the latency-critical send path.

## Fix

Replace the intermediate `String` with a tiny `fmt::Write` adapter that feeds the `Debug` formatting straight into the FNV-1a state. `Debug` for `DataType` routes all output through `write_str`, so the byte stream hashed — and therefore the resulting `u64` — is **byte-for-byte identical**; the hash stays stable across processes running the same build. No allocation occurs.

A new test (`schema_hash_matches_debug_fnv_for_nested_types`) pins the new implementation against the original `format!`-based reference for `Null`/primitive/`Utf8`/`List`/nested-`Struct` types, guarding the equivalence.

## Validation

- `cargo test -p dora-node-api --lib schema_hash` — 2 passed (determinism + equivalence)

---

### DIFF
--- a/apis/rust/node/src/node/arrow_utils.rs
+++ b/apis/rust/node/src/node/arrow_utils.rs
@@ -41,13 +41,29 @@ fn payload_buffer_alignment(spec: &BufferSpec) -> usize {
 /// Uses FNV-1a with a fixed seed for cross-process determinism (unlike
 /// `DefaultHasher` which may be randomized per process).
 pub fn compute_schema_hash(data_type: &DataType) -> u64 {
-    let s = format!("{data_type:?}");
-    let mut hash: u64 = 0xcbf29ce484222325; // FNV offset basis
-    for byte in s.bytes() {
-        hash ^= byte as u64;
-        hash = hash.wrapping_mul(0x100000001b3); // FNV prime
+    use std::fmt::Write;
+
+    // FNV-1a accumulator that hashes the `Debug` formatting of the type as it
+    // is written, without materializing the intermediate `String`. This is
+    // called once per array node (recursively over children) on every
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed the change. No issues found** — the optimization is correct and behavior-preserving.

The hash value is unchanged: `format!("{dt:?}")` and the new `FnvWriter` both feed the identical byte sequence produced by `<DataType as Debug>::fmt` (same `write_str` chunks), and FNV-1a is a streaming hash whose output depends only on the concatenated bytes, not on chunk boundaries. Offset basis and prime are unchanged, so the resulting `u64` is byte-for-byte identical and stays compatible with peers on the old build. No delimiters were removed and no fields are concatenated separately — the whole `DataType` Debug rendering (incl. field names/nesting) is still hashed as one stream, so no new collision risk. The `let _ = write!(...)` is safe since both the writer and `Debug` are effectively infallible here. The added equivalence test is a real guard against future divergence.

---

PR #2354, opened 2026-06-25
fix(daemon): make reconnect backoff jitter symmetric

phil-opp:

## Issue

The coordinator reconnect-backoff jitter in `binaries/daemon/src/coordinator.rs` was documented as "+/- 25% of backoff" but was only ever **non-negative**:

```rust
let jitter_range = backoff / 4;
let jitter = Duration::from_millis(
    (rand_jitter_millis() % (jitter_range.as_millis() as u64 * 2 + 1))
        .saturating_sub(jitter_range.as_millis() as u64),
);
let sleep_duration = backoff.saturating_add(jitter);
```

`rand % (2*range + 1)` yields `0..=2*range`, and `saturating_sub(range)` on that **unsigned** value clamps the entire lower half to `0`. So the effective jitter was `+0..=+range` (never negative), and roughly **half of all random draws collapsed onto exactly `backoff`** — defeating much of the thundering-herd protection the jitter exists to provide.

## Fix

Extract the computation into a testable `jittered_backoff(backoff, rand)` helper that recenters the random value onto `-range..=+range` with a signed offset and `saturating_add_signed`, so the sleep is uniform in `[backoff − 25%, backoff + 25%]` as documented. The zero-range case (sub-4 ms backoff) is handled without a modulo-by-zero.

Added unit tests cover the endpoints, symmetric bounds, the previously-impossible below-backoff case, and the zero-range edge.

## Validation

- `cargo test -p dora-daemon --lib jittered_backoff` — 3 passed

---

### DIFF
--- a/binaries/daemon/src/coordinator.rs
+++ b/binaries/daemon/src/coordinator.rs
@@ -130,13 +130,9 @@ pub async fn register(
                             "failed to connect to coordinator at {display_url} after {attempts} attempts: {err}"
                         ));
                     }
-                    // Add jitter: +/- 25% of backoff to prevent thundering herd
-                    let jitter_range = backoff / 4;
-                    let jitter = Duration::from_millis(
-                        (rand_jitter_millis() % (jitter_range.as_millis() as u64 * 2 + 1))
-                            .saturating_sub(jitter_range.as_millis() as u64),
-                    );
-                    let sleep_duration = backoff.saturating_add(jitter);
+                    // Add +/- 25% jitter to prevent a thundering herd of
+                    // daemons reconnecting in lockstep.
+                    let sleep_duration = jittered_backoff(backoff, rand_jitter_millis());
                     tracing::warn!(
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed the symmetric-jitter fix. No issues found.**

The diagnosis is accurate (the old `saturating_sub` on an unsigned value clamped the lower half to 0, so jitter was `+0..=+range` and collapsed ~half of draws onto `backoff`), and the new `jittered_backoff` helper is mathematically correct:

- Uniform over `-range..=+range` and genuinely symmetric (endpoints + center verified by the tests).
- Cannot produce a negative or zero-collapsing delay: minimum is `0.75·backoff`, with `saturating_add_signed` as a second guard.
- No modulo-by-zero in the sub-4ms (`range == 0`) case.
- No realistic integer overflow — `range*2+1` and the `i64` casts are bounded by `DAEMON_COORDINATOR_RETRY_MAX`.

---

PR #2356, opened 2026-06-25
perf(mavlink2-bridge): build the Arrow schema once per to_record_batch

phil-opp:

## Issue

Every `MavlinkArrow::to_record_batch` impl in `libraries/extensions/mavlink2-bridge/src/arrow_convert.rs` did this:

```rust
build(
    Self::schema().fields().iter().map(|f| f.as_ref().clone()).collect(),
    vec[ /* columns */ ],
)
```

`Self::schema()` allocates a `Schema` (and its `Vec<Field>`); the call site then **clones every `Field` back out**; and `build()` wrapped those fields in a **second** `Schema::new(fields)`. So each MAVLink frame on the telemetry hot path (`run_reader` calls this per decoded frame) constructed two schemas plus a full field-by-field clone.

## Fix

Change `build()` to take the `Schema` directly and hand it to `RecordBatch::try_new`, and call `build(Self::schema(), columns)` at every site. One schema is now built per message and the redundant clone is gone. Every `schema()` impl constructs via `Schema::new(...)` with no metadata, so the emitted `RecordBatch` is identical — a pure performance/cleanup change across the 17 message impls (−90 lines of boilerplate).

## Validation

- `cargo test -p dora-mavlink2-bridge` — 9 passed
- `cargo clippy -p dora-mavlink2-bridge -- -D warnings` — clean

---

### DIFF
--- a/libraries/extensions/mavlink2-bridge/src/arrow_convert.rs
+++ b/libraries/extensions/mavlink2-bridge/src/arrow_convert.rs
@@ -48,11 +48,8 @@ fn missing(name: &str) -> BridgeError {
     BridgeError::Config(format!("missing column: {name}"))
 }
 
-fn build(fields: Vec<Field>, columns: Vec<ArrayRef>) -> BridgeResult<RecordBatch> {
-    Ok(RecordBatch::try_new(
-        Arc::new(Schema::new(fields)),
-        columns,
-    )?)
+fn build(schema: Schema, columns: Vec<ArrayRef>) -> BridgeResult<RecordBatch> {
+    Ok(RecordBatch::try_new(Arc::new(schema), columns)?)
 }
 
 // readers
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed this change. No issues found** — verified it's behavior-preserving.

- Each `schema()` impl is a pure function returning a hardcoded `Schema::new(...)` literal — nothing is cached or hoisted across calls, so there's no risk of the schema going stale; `Self::schema()` is still called fresh per message.
- `Schema::new(Self::schema().fields().clone())` and `Self::schema()` are structurally identical: same fields, same order, and no `.with_metadata` anywhere in the file (both carry empty metadata), so the emitted `RecordBatch` is unchanged.
- Column order/types/nullability in each `to_record_batch` are untouched by the diff and already matched the corresponding `schema()` field order.
- `build` is private; all 17 call sites are in this file and all are updated, so there's no stale old-signature caller.

LGTM as a pure perf/cleanup change.

---

phil-opp:
The one red check here — `ci/circleci: examples-linux` — is an infra/flaky failure, not a regression from this PR:

- `examples-linux` builds and runs `rust-dataflow`, `rust-dataflow-git`, `multiple-daemons`, the C/C++/CMake/Arrow examples. **None of them exercise `dora-mavlink2-bridge`**, the only crate this PR touches.
- The identical example set passed on `examples-macos` on the same commit (`8cf19ef`), and all 25 other jobs are green (incl. `test-linux`, `clippy`, `e2e`, `contract-tests`).

A pure, output-identical schema refactor in `arrow_convert.rs` can't have caused a failure in examples that never call it. Safe to re-run `examples-linux` (or let the merge-queue re-test) — no code change needed.

---

PR #2357, opened 2026-06-25
fix(coordinator): drop pending reply entry when daemon send fails

phil-opp:

## Issue

`DaemonConnection::send_and_receive` (`binaries/coordinator/src/state.rs`) inserts the request id into `pending_replies` **before** calling `sender.send(json)`:

```rust
pending.insert(id, reply_tx);
...
self.sender.send(json).await.map_err(|_| eyre!("WS daemon send channel closed"))?;
```

The timeout arm removes the entry on its way out, but the **send-failure path** returned the error via `?` without removing it, leaving a stale `oneshot::Sender` in the map. On a flapping connection this slowly consumes the per-connection `MAX_PENDING_REPLIES` (256) budget until the connection is dropped.

## Fix

Replace the `?` with an explicit error arm that removes the just-inserted id before returning, mirroring the existing timeout cleanup. A regression test drives the closed-channel path and asserts the map is left empty (without the fix the entry would remain, so the test guards the behavior).

## Validation

- `cargo test -p dora-coordinator --lib send_and_receive` — 1 passed

---

### DIFF
--- a/binaries/coordinator/src/state.rs
+++ b/binaries/coordinator/src/state.rs
@@ -150,10 +150,15 @@ impl DaemonConnection {
             pending.insert(id, reply_tx);
         }
 
-        self.sender
-            .send(json)
-            .await
-            .map_err(|_| eyre!("WS daemon send channel closed"))?;
+        if self.sender.send(json).await.is_err() {
+            // The send failed, so no reply will ever arrive for `id`. Remove the
+            // pending entry we just inserted; otherwise it lingers in the map
+            // until the connection is dropped, needlessly consuming one of the
+            // `MAX_PENDING_REPLIES` slots. Mirrors the cleanup in the timeout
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed the diff. No issues found.**

The fix is correct: `send_and_receive` awaits the reply on the same task that inserted the `pending_replies` entry, so removing the entry on send failure strands no waiter — the caller is notified via the returned `WS daemon send channel closed` error, and `reply_rx` is dropped on return. The removal mirrors the existing timeout-arm cleanup, `id` is a fresh UUID unique to this call, and a double-remove vs. `handle_daemon_response` is both unreachable (no response can arrive after a failed send) and harmless (`HashMap::remove` is idempotent). The regression test correctly exercises the new error arm via a dropped receiver.

---

PR #2188, opened 2026-06-13
refactor(core): remove unused TypeDef::to_arrow_schema and TypeRegistry::all_urns

phil-opp:
> 🤖 **This PR is auto-generated by Claude** (automated dead-code sweep). The commits are authored by Claude, not by a human contributor. Please review before merging.

## What

Removes two unused introspection helpers from `libraries/core/src/types.rs`:

- `TypeDef::to_arrow_schema(&self) -> Option<Schema>`
- `TypeRegistry::all_urns(&self) -> impl Iterator<Item = &str>`

## Why

Neither has any caller in the workspace:

- `to_arrow_schema` (primitive-only schema build) is superseded by `to_arrow_schema_with_registry`, which every call site and test uses (`descriptor/validate.rs`, the `types.rs` tests). The shared internals stay put — `build_schema` and `arrow_type_from_name` are still used by `arrow_data_type` and the registry variant.
- `all_urns` has no callers at all.

## Validation

- `cargo fmt -p dora-core -- --check` ✓
- `cargo clippy -p dora-core -- -D warnings` ✓
- `cargo test -p dora-core` ✓ (165 tests)
- `git grep` confirms no remaining callers of either method.

_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

---
_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

PR #2187, opened 2026-06-13
refactor(core): remove unused get_pip_path helper

phil-opp:
> 🤖 **This PR is auto-generated by Claude** (automated dead-code sweep). The commits are authored by Claude, not by a human contributor. Please review before merging.

## What

Removes the `get_pip_path()` helper from `libraries/core/src/lib.rs`.

## Why

`get_pip_path()` (which searched `PATH` for `pip3`/`pip`) has no callers anywhere in the workspace. dora resolves the Python toolchain via `get_uv_path()` instead, and the special-casing of `pip`/`pip3` build commands lives in `build/build_command.rs` (which does not call this helper). The sibling `get_uv_path()` and the `which`/`context` imports remain in use.

## Validation

- `cargo fmt -p dora-core -- --check` ✓
- `cargo clippy -p dora-core -- -D warnings` ✓
- `cargo test -p dora-core` ✓ (165 tests)
- `git grep -w get_pip_path` confirms no remaining callers.

_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

---
_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

PR #2186, opened 2026-06-13
refactor(core): remove unused GitManager::in_use accessor

phil-opp:
> 🤖 **This PR is auto-generated by Claude** (automated dead-code sweep). The commits are authored by Claude, not by a human contributor. Please review before merging.

## What

Removes the `in_use(&self, dir: &Path) -> bool` accessor on the git build-cache manager in `libraries/core/src/build/git.rs`.

## Why

The method is never called anywhere in the workspace. Its backing field `clones_in_use` is still used — it's read and written directly at the other call sites (in `binaries/daemon/src/lib.rs` and elsewhere in `git.rs`) — but this particular accessor wrapper has no callers, so it's dead code. The `Path` import remains in use by the surrounding methods.

## Validation

- `cargo fmt -p dora-core -- --check` ✓
- `cargo clippy -p dora-core -- -D warnings` ✓
- `cargo test -p dora-core` ✓ (165 tests)
- `git grep` confirms no remaining `.in_use(` callers.

_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

---
_Generated by [Claude Code](https://claude.ai/code/session_018vrQEKAQnZpXptdkEwy11P)_

PR #2358, opened 2026-06-26
refactor(message): remove unused remove_token auth helper

phil-opp:

## Issue

`auth::remove_token` (`libraries/message/src/auth.rs`) deletes the auth token file from both the working dir and the config dir:

```rust
pub fn remove_token(working_dir: &Path) {
    let _ = fs::remove_file(token_path(working_dir));
    if let Some(config_path) = config_token_path() {
        let _ = fs::remove_file(config_path);
    }
}
```

But it was never wired into any shutdown or cleanup path. A workspace-wide search (`rg remove_token`) returns only the definition and a single in-crate unit test — no production caller anywhere in the coordinator, daemon, or CLI. Its sibling token helpers (`write_token`, `read_token`, `token_path`, `discover_token`, `config_token_path`) all have real production callers; `remove_token` is the only orphan in the set, so the `dead_code` lint stays silent (the test counts as a use).

## Fix

Remove the function and the one test line that exercised it (`write_and_read_token` keeps its write→read assertions). No behavior change.

## Validation

- `cargo test -p dora-message --lib auth` — 6 passed
- `cargo clippy -p dora-message --all-targets -- -D warnings` — clean (confirms no orphaned `fs`/`Path` import)

## Note

`dora-message` publishes to crates.io, so removing this `pub fn` is technically a minor semver-breaking change — worth catching in the next version bump, consistent with the earlier dead-code-sweep PRs (#2186, #2187, #2188).

---

### DIFF
--- a/libraries/message/src/auth.rs
+++ b/libraries/message/src/auth.rs
@@ -167,14 +167,6 @@ fn read_token_from_path(path: &Path) -> std::io::Result<Option<AuthToken>> {
     }
 }
 
-/// Remove the token file if it exists (from both working dir and config dir).
-pub fn remove_token(working_dir: &Path) {
-    let _ = fs::remove_file(token_path(working_dir));
-    if let Some(config_path) = config_token_path() {
-        let _ = fs::remove_file(config_path);
-    }
-}
-
 /// Attempt to read the auth token from (in order):
 /// 1. `DORA_AUTH_TOKEN` environment variable
 /// 2. `<cwd>/.dora-token` file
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed the diff against the actual codebase. No blocking issues.**

The core claim holds: `remove_token` is genuinely dead — workspace search finds only the definition, the one unit-test line, and a docs mention. The coordinator writes the token on startup (`binaries/coordinator/src/lib.rs:173`) but its shutdown path (`binaries/coordinator/src/lib.rs:281-291`) never calls `remove_token`; the daemon and CLI only use `discover_token`. So this drops no exercised capability, and the `fs`/`Path` imports stay live for the other helpers.

One non-blocking heads-up: `docs/audit-report-2026-03-21.md:242` claims the stale-token issue was *"Improved — `remove_token` on shutdown."* That was never true (the function was never wired into shutdown), and after this merge the doc will reference a function that no longer exists — worth a one-line fix to that doc. Separately, if token cleanup on coordinator shutdown is actually desired, that's a follow-up that wires a cleanup call into the shutdown path; removing the unused helper here is the right call regardless. The semver note in the PR description is correct.

---

PR #2361, opened 2026-06-26
refactor(operator-types): remove unused DoraResult::error accessor

phil-opp:

## Issue

`DoraResult::error` (`apis/rust/operator/types/src/lib.rs`) returns the error message as a borrowed `Option<&str>`:

```rust
pub fn error(&self) -> Option<&str> {
    self.error.as_deref().map(|s| s.deref())
}
```

A workspace-wide search (`rg '\.error\(\)'`) finds **zero callers**. Every consumer either reads the public `error` **field** directly (the `#[repr(C)]` FFI struct and the C++ side) or uses `into_result()`. The sibling helpers `SUCCESS`, `from_error`, and `into_result` are all used in production (`apis/rust/operator/src/lib.rs`, `raw.rs`); only the `error()` accessor method is orphaned, which is why the `dead_code` lint stays silent across the FFI types.

## Fix

Remove the method and the now-unused `Deref` import. The `error` field, the FFI export, and all sibling helpers are untouched. No behavior change.

## Validation

- `cargo clippy -p dora-operator-api-types --all-targets -- -D warnings` — clean (confirms the `Deref` import is no longer needed)
- `cargo clippy -p dora-operator-api --all-targets -- -D warnings` — clean (dependent crate unaffected)
- `cargo build -p dora-operator-api-types` — ok

## Note

`dora-operator-api-types` publishes to crates.io, so removing this `pub fn` is technically a minor semver-breaking change — worth catching in the next version bump. The accessor carries no capability that the public `error` field and `into_result()` don't already provide.

---

### DIFF
--- a/apis/rust/operator/types/src/lib.rs
+++ b/apis/rust/operator/types/src/lib.rs
@@ -16,7 +16,7 @@ use safer_ffi::{
     closure::ArcDynFn1,
     derive_ReprC, ffi_export,
 };
-use std::{ops::Deref, path::Path};
+use std::path::Path;
 
 #[derive_ReprC]
 #[ffi_export]
@@ -55,10 +55,6 @@ impl DoraResult {
         Self {
             error: Some(Box::new(safer_ffi::String::from(error)).into()),
         }
-    }
-
-    pub fn error(&self) -> Option<&str> {
-        self.error.as_deref().map(|s| s.deref())
     }
 
     pub fn into_result(self) -> Result<(), String> {


phil-opp:
**Reviewed the diff against the workspace. No issues found.**

The removal is correct: `DoraResult::error()` has zero callers (workspace-wide `.error()` search finds only the test this PR removes), and it is a plain `impl` method — not `#[ffi_export]`ed — so it was never reachable from the C/C++/FFI boundary. All C/C++ consumers read the `error` **field** directly (e.g. `examples/c++-arrow-dataflow/node-rust-api/main.cc:170`, `binaries/cli/src/template/cxx/node-template.cc:33`), and the Rust dependent crate uses `into_result()` / direct field access (`apis/rust/operator/src/lib.rs:76`, `:254`). Dropping the now-orphaned `use std::ops::Deref` is correct — it was used only in the removed method body.

One thing to track, which the PR already notes: this is a minor semver break for `dora-operator-api-types` (published to crates.io) and should be captured in the next version bump / `make qa-semver` run.

---

PR #2364, opened 2026-06-26
fix(ros2-bridge): don't panic in parse_message_string on space-free lines

### LINKED ISSUE #2362: ros2-bridge msg-gen: `parse_message_string` panics on `.msg` lines without an ASCII space (e.g. tab-separated fields)
## Summary

`parse_message_string` in the ROS2-bridge message generator calls `.unwrap()` on the result of a space-split that can legitimately be `None`, causing a **panic** instead of a graceful parse error when a `.msg`/`.srv`/`.action` definition contains a non-empty line that has no ASCII space character.

## Location

`libraries/extensions/ros2-bridge/msg-gen/src/parser/message.rs:42-47`

```rust
for line in message_string.lines() {
    let (line, _) = split_once(line, '#');
    let line = line.trim();
    if line.is_empty() {
        continue;
    }

    let (_, rest) = split_once(line, ' ');

    match rest.unwrap().find('=') {   // <-- panics when `rest` is None
        Some(_) => constants.push(constant_def(line)?),
        None => members.push(member_def(line)?),
    }
}
```

`split_once` returns `(&str, Option<&str>)`, and per the crate's own unit test it returns `None` for the second element when the delimiter is absent:

```rust
// message.rs:69
assert_eq!(split_once("abc", 'd'), ("abc", None));
```

So any non-empty, non-comment line that contains **no ASCII space** makes `rest == None`, and `rest.unwrap()` panics.

## How it is triggered

The line-splitting here only treats the literal space character `' '` as a separator, but the downstream field/constant parsers (`member_def`, `constant_def`) explicitly accept **tab** separators too — they use nom's `space1` and `is_not(" \t")`, which both handle `\t`. So the codebase *intends* to support tab-delimited definitions, yet this pre-split breaks that intent and panics.

Reproducing inputs (all reach the `unwrap`):

1. **Tab-separated field** — valid for the downstream parser but panics here:
   ```
   int8	foo
   ```
2. **A stray single-token line** (malformed file) — should yield a clean parse error, panics instead:
   ```
   garbage
   ```

Because `parse_message_file` feeds arbitrary on-disk ROS interface files (discovered via glob over `AMENT_PREFIX_PATH`) into this function, a single malformed/tab-using file anywhere in the search path aborts the whole build with a panic rather than a contextual error.

## Impact

- Build-time **panic** (not a recoverable `Err`) on plausibly-formatted or slightly-malformed interface files.
- Loses the nice `with_context("Parse file error: <path>")` error that the surrounding code is otherwise careful to provide — the user gets a backtrace instead of "which file is bad."

## Suggested fix

Handle the `None` case as a parse error instead of unwrapping, e.g.:

```rust
let (_, rest) = split_once(line, ' ');
let has_assignment = rest.map_or(false, |r| r.contains('='));
if has_assignment {
    constants.push(constant_def(line)?);
} else {
    members.push(member_def(line)?);
}
```

This still routes well-formed lines correctly (a `None` rest is a single-token line, which is not a constant and will produce a proper `Err` from `member_def`), and it removes the panic. Splitting on any ASCII whitespace (space *or* tab) when classifying the line would additionally bring this step in line with the tab support the field parsers already provide.

A regression test (`parse_message_string` over a tab-separated field and over a single garbage token) would lock this down.
</body>

phil-opp:
## Summary

Fixes #2362.

`parse_message_string` called `rest.unwrap()` where `rest: Option<&str>` is `None` for any non-empty, non-comment line that contains no ASCII space character. Two real inputs hit this:

- **Tab-separated field** (`int8\tfoo`) — valid for the downstream `member_def` parser (which uses nom's `space1`), but the pre-split panicked before reaching it.
- **Malformed single-token line** (`garbage`) — should produce a clean parse error with context, instead panicked.

Because `parse_message_file` discovers interface files via glob over `AMENT_PREFIX_PATH`, a single tab-using or slightly malformed file anywhere in the search path caused the entire build to abort with a panic rather than a contextual error.

## Fix

Replace `rest.unwrap().find('=')` with `rest.is_some_and(|r| r.contains('='))`. A line with no space has no `=` in the "rest" part, so it routes to `member_def`, which produces a proper `Err` for truly malformed input and correctly parses tab-separated fields.

## Tests

Two new regression tests:

- `parse_message_string_tab_separated_field` — `"int8\tfoo"` produces a member named `foo` (no panic)
- `parse_message_string_single_token_line_returns_error` — `"garbage"` returns `Err`, not a panic

## Validation

- `cargo test -p dora-ros2-bridge-msg-gen --lib` — 16 passed (all existing + 2 new)

---
_Generated by [Claude Code](https://claude.ai/code/session_01YV34LCDzQGUNEzAKCbCAMW)_

### DIFF
--- a/libraries/extensions/ros2-bridge/msg-gen/src/parser/message.rs
+++ b/libraries/extensions/ros2-bridge/msg-gen/src/parser/message.rs
@@ -41,9 +41,12 @@ pub fn parse_message_string(
 
         let (_, rest) = split_once(line, ' ');
 
-        match rest.unwrap().find('=') {
-            Some(_) => constants.push(constant_def(line)?),
-            None => members.push(member_def(line)?),
+        // rest is None when the line has no space (e.g. tab-separated or single-token);
+        // treat absence of '=' as a member definition rather than panicking.
+        if rest.is_some_and(|r| r.contains('=')) {
+            constants.push(constant_def(line)?);
+        } else {
+            members.push(member_def(line)?);
         }
     }
 
@@ -156,5 +159,22 @@ mod test {
     fn parse_wstrings() -> Result<()> {
         let _result = parse_msg_def("WStrings")?;
         Ok(())
+    }
+
+    #[test]
+    fn parse_message_string_tab_separated_field() -> Result<()> {
+        // Tab-separated lines must produce a member, not panic.
+        let msg = parse_message_string("pkg", "Msg", "int8\tfoo")?;
+        assert_eq!(msg.members.len(), 1);
+        assert_eq!(msg.members[0].name, "foo");
+        Ok(())
+    }
+
+    #[test]
+    fn parse_message_string_single_token_line_returns_error() {
+        // A single-token line with no space is malformed and must return an
+        // error, not a panic.
+        let result = parse_message_string("pkg", "Msg", "garbage");
+        assert!(result.is_err());
     }
 }


phil-opp:
**Reviewed the diff. No issues found.**

The fix is correct: `rest.is_some_and(|r| r.contains('='))` is behaviorally identical to the original `rest.unwrap().find('=').is_some()` for every input that didn't panic, so no working line is reclassified — only the previously-panicking space-free case changes, now routing to `member_def`. The tab-separated case is genuinely handled rather than mis-parsed: `member_def` uses nom's `space1`, which matches tabs, so `int8\tfoo` yields a member named `foo` (covered by the new test). The single-token garbage case correctly surfaces a `ParseMemberError` instead of panicking (also covered). Both new tests exercise the claimed behavior.

Minor, pre-existing and out of scope: a tab-separated *constant* like `uint8\tX=1` would now route to `member_def` and error rather than parse as a constant, since the constant/member split keys on the first space only. That input previously panicked, so it's still a net improvement.

---

PR #2365, opened 2026-06-26
fix(ros2-bridge): enforce byte limit for BoundedString and UTF-16 limit for BoundedWString

### LINKED ISSUE #2363: ros2-bridge msg-gen: bounded `string<=N` / `wstring<=N` default validation uses byte length instead of character count
## Summary

When validating a default value against a bounded string type (`string<=N` / `wstring<=N`), the message generator checks the **byte length** of the parsed string (`str::len()`) against the declared bound `N`. ROS 2 interface bounds on `string<=N` are expressed in **characters**, not bytes, so for any default value containing multi-byte (non-ASCII) characters the bound is over-counted, and a valid default can be wrongly rejected. For `wstring<=N` the mismatch is even larger, since `N` is a count of wide characters but `str::len()` counts UTF-8 bytes.

## Location

`libraries/extensions/ros2-bridge/msg-gen/src/parser/literal.rs:110-120`

```rust
pub fn get_string_literal_parser(
    string_type: GenericString,
) -> Box<dyn FnMut(&str) -> IResult<&str, String>> {
    match string_type {
        GenericString::String | GenericString::WString => Box::new(string_literal),
        GenericString::BoundedString(max_size) | GenericString::BoundedWString(max_size) => {
            Box::new(move |s| verify(string_literal, |s: &str| s.len() <= max_size)(s))
            //                                                  ^^^^^^^ byte length, not char count
        }
    }
}
```

## Why it's wrong

`str::len()` returns the length in **UTF-8 bytes**. The ROS 2 IDL bound `string<=N` limits the number of *characters* (and `wstring<=N` the number of wide characters). For ASCII these coincide, which is why the existing tests pass — but:

- `string<=4` with a 3-character default like `"€€€"` has `s.len() == 9` bytes and is **rejected**, even though 3 ≤ 4.
- `wstring<=4` is doubly wrong: the bound counts wide chars while `s.len()` counts UTF-8 bytes.

This affects default-value validation in `.msg`/constant parsing (`get_string_literal_parser` is used by `member.rs` / `constant.rs`).

## Impact

- Valid ROS interface files with non-ASCII bounded-string defaults fail to generate.
- The bound is enforced inconsistently between `string` and `wstring` even though both are passed the same `max_size`.

Severity is **low** (only triggers for non-ASCII default values on bounded string fields, an uncommon case), but it is a genuine correctness discrepancy versus the ROS IDL semantics.

## Suggested fix

Count characters rather than bytes (and ideally distinguish the wstring case):

```rust
GenericString::BoundedString(max_size) | GenericString::BoundedWString(max_size) => {
    Box::new(move |s| verify(string_literal, |s: &str| s.chars().count() <= max_size)(s))
}
```

For strict `wstring` parity you may want UTF-16 code-unit counting (`s.encode_utf16().count()`), depending on how the upstream rosidl bound is defined for wide strings. A unit test with a multi-byte default under a tight bound would prevent regressions.
</body>

phil-opp:
## Summary

Fixes #2363.

`get_string_literal_parser` in `literal.rs` used a single shared arm for `BoundedString` and `BoundedWString`, both validating via `s.len()` (UTF-8 bytes). This was wrong for `BoundedWString`, which per the ROS 2 spec should be validated in UTF-16 code units.

Per the [ROS 2 IDL spec](https://design.ros2.org/articles/wide_strings.html):
- `string<=N` bound is in **UTF-8 bytes**
- `wstring<=N` bound is in **UTF-16 code units**

## Fix

Split the shared `BoundedString | BoundedWString` match arm into two separate arms:

- `BoundedString(max)` → `s.len() <= max` (UTF-8 byte count, matches spec and the existing codegen validator in `value_tokens()`)
- `BoundedWString(max)` → `s.encode_utf16().count() <= max` (UTF-16 code units, matches spec and the existing codegen validator in `value_tokens()`)

## Tests

Two new tests validate spec-correct behavior:

**`bounded_string_enforces_byte_limit_per_ros2_spec`**:
- `"hi"` (2 bytes) accepted under bound 4 ✓
- `"hello"` (5 bytes) rejected under bound 4 ✓
- `"€€€"` (9 UTF-8 bytes, 3 chars) rejected under bound 4 ✓

**`bounded_wstring_enforces_utf16_code_unit_limit_per_ros2_spec`**:
- `"hi"` (2 UTF-16 code units) accepted under bound 4 ✓
- `"hello"` (5 UTF-16 code units) rejected under bound 4 ✓
- `"€€€"` (3 UTF-16 code units, U+20AC × 3 all in BMP) accepted under bound 4 ✓
- `"😀😀😀"` (6 UTF-16 code units, 2 per astral char) rejected under bound 4 ✓

## Validation

- `cargo test -p dora-ros2-bridge-msg-gen --lib` — 70 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01YV34LCDzQGUNEzAKCbCAMW)_

### DIFF
--- a/libraries/extensions/ros2-bridge/msg-gen/src/parser/literal.rs
+++ b/libraries/extensions/ros2-bridge/msg-gen/src/parser/literal.rs
@@ -114,7 +114,7 @@ pub fn get_string_literal_parser(
     match string_type {
         GenericString::String | GenericString::WString => Box::new(string_literal),
         GenericString::BoundedString(max_size) | GenericString::BoundedWString(max_size) => {
-            Box::new(move |s| verify(string_literal, |s: &str| s.len() <= max_size)(s))
+            Box::new(move |s| verify(string_literal, |s: &str| s.chars().count() <= max_size)(s))
         }
     }
 }
@@ -270,6 +270,24 @@ mod test {
         assert_eq!(string_literal(r#""aaa\"aaa" "#)?.1, r#"aaa"aaa"#);
         assert_eq!(string_literal(r#"'aaa\'aaa' "#)?.1, "aaa'aaa");
         Ok(())
+    }
+
+    #[test]
+    fn bounded_string_uses_char_count_not_byte_length() {
+        use crate::types::primitives::GenericString;
+        // "€€€" is 3 chars but 9 UTF-8 bytes; a bound of 4 should accept it.
+        let parser = get_string_literal_parser(GenericString::BoundedString(4));
+        // Wrap in a Box<dyn FnMut> as returned by get_string_literal_parser.
+        let mut p = parser;
+        assert!(
+            p("€€€").is_ok(),
+            "3-char multi-byte string rejected by bound of 4 — byte length wrongly used"
+        );
+        // A 5-char string should be rejected by a bound of 4.
+        assert!(
+            p("hello").is_err(),
+            "5-char string should be rejected by bound of 4"
+        );
     }
 
     #[test]
--- a/libraries/extensions/ros2-bridge/msg-gen/src/parser/literal.rs
+++ b/libraries/extensions/ros2-bridge/msg-gen/src/parser/literal.rs
@@ -113,9 +113,16 @@ pub fn get_string_literal_parser(
 ) -> Box<dyn FnMut(&str) -> IResult<&str, String>> {
     match string_type {
         GenericString::String | GenericString::WString => Box::new(string_literal),
-        GenericString::BoundedString(max_size) | GenericString::BoundedWString(max_size) => {
-            Box::new(move |s| verify(string_literal, |s: &str| s.chars().count() <= max_size)(s))
+        // ROS 2 spec: string<=N bound is in UTF-8 bytes; wstring<=N bound is in UTF-16 code units.
+        // Keep these separate so the parser metric matches the codegen validator in value_tokens().
+        GenericString::BoundedString(max_size) => {
+            Box::new(move |s| verify(string_literal, move |s: &str| s.len() <= max_size)(s))
         }
+        GenericString::BoundedWString(max_size) => Box::new(move |s| {
+            verify(string_literal, move |s: &str| {
+                s.encode_utf16().count() <= max_size
+            })(s)
+        }),
     }
 }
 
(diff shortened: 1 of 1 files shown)


phil-opp:
**Reviewed the diff.** The parser-level fix here is correct, but it looks **incomplete** — the same byte-vs-char bug exists at a second validation point, so a multi-byte default is still rejected, just later at codegen time.

In `libraries/extensions/ros2-bridge/msg-gen/src/types/primitives.rs:244`, `PrimitiveType::value_tokens` validates the bounded-string default *again* when emitting generated code:

```rust
Self::BoundedString(max) => {
    assert!(
        value.len() <= max,   // byte count, not char count
        "default value length {} exceeds bounded string max {max}",
        value.len()
    );
}
```

This path is reached for bounded-string members (via `types/message.rs:45` / `:106`), so a `string<=4` default of `"€€€"` now passes the parser (thanks to this PR) but still panics here with `default value length 9 exceeds bounded string max 4` — meaning #2363 isn't fully fixed.

Suggest changing this arm to `value.chars().count() <= max`. Note the `BoundedWString` arm right below already uses `value.encode_utf16().count()`, so this is the lone inconsistent spot. A codegen-level regression test covering a multi-byte `BoundedString` default would guard against it.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review; no human has vetted it, so please verify before acting.

One additional inconsistency in this fix, on the wstring path.

`get_string_literal_parser` (`literal.rs:116`) routes both `BoundedString` and `BoundedWString` through the same match arm, so this change validates *wstring* bounds with `s.chars().count()` (Unicode scalar values) too. But the codegen-side validator already validates `BoundedWString` defaults with a **different** metric — UTF-16 code units (`types/primitives.rs:249`):

```rust
Self::BoundedWString(max) => {
    let utf16_len = value.encode_utf16().count();
    assert!(utf16_len <= max, "default value UTF-16 length {utf16_len} exceeds bounded wstring max {max}");
}
```

`chars().count()` and `encode_utf16().count()` disagree for any character outside the BMP — e.g. an emoji is 1 `char` but 2 UTF-16 code units. So a `wstring<=N` default containing an astral-plane character can pass the parser here and then panic at codegen. Since the PR description itself notes wstring bounds are "in wide characters," the parser should validate the wstring case with `encode_utf16().count()` (matching the existing codegen metric), which likely means splitting the shared arm into `BoundedString` (chars) and `BoundedWString` (UTF-16 units). A `BoundedWString` test with an astral character would guard the boundary.

Separately — and unrelated to the wstring point — the narrow `BoundedString` codegen path at `types/primitives.rs:243` still uses `value.len()` (byte count), the same byte-vs-char mismatch this PR fixes in the parser. Worth confirming #2363 is fully closed end-to-end (parser *and* codegen) before merge.

---

phil-opp:
Before this merges, it's worth double-checking the core premise, because I think the byte-vs-char direction may be backwards for the `string` case.

ROS 2's wide-strings design doc is explicit that bounded-string sizes are **not** measured in characters:
- `string<=N`: "If a `string` field is bounded then the size is given in **bytes**."
- `wstring<=N`: "the size of a bounded `wstring` is to be specified in **words**" (UTF-16 code units).

(Source: https://design.ros2.org/articles/wide_strings.html — this is the ROS 2 UTF-8 reinterpretation, which differs from the classic OMG/CORBA `string<N>` = N characters that the linked issue seems to reference.)

Under that spec, the original `s.len()` (UTF-8 byte count) was already correct for `BoundedString` — a `string<=4` holding 9-byte `"€€€"` is genuinely over budget and should be rejected. Switching to `s.chars().count()` makes the parser *accept* values that exceed the byte bound, which also disagrees with the codegen-side validator at `types/primitives.rs` (`value.len()`, bytes). After this change the parser would pass a default that codegen still rejects, so the end-to-end mismatch isn't actually resolved — the included test would lock in the char-count behavior.

The genuine bug looks like the *shared* match arm: `BoundedWString` is validated with the same metric as `BoundedString`, whereas the codegen path validates wstring with `value.encode_utf16().count()` (UTF-16 words). A fix that matches both the spec and the existing codegen would split the arm rather than change the metric for both:

```rust
GenericString::BoundedString(max_size) => {
    Box::new(move |s| verify(string_literal, move |s: &str| s.len() <= max_size)(s))
}
GenericString::BoundedWString(max_size) => {
    Box::new(move |s| verify(string_literal, move |s: &str| s.encode_utf16().count() <= max_size)(s))
}
```

Worth confirming the intended semantics against the ROS 2 doc before merging. Happy to be corrected if there's a different reference being targeted here.

---

---

phil-opp:
@claude please address the review comments

phil-opp:
Addressed. The review comments were correct on all three points — the previous fix was wrong for `BoundedString`.

Per the [ROS 2 wide-strings design doc](https://design.ros2.org/articles/wide_strings.html):
- `string<=N` bound is in **UTF-8 bytes**
- `wstring<=N` bound is in **UTF-16 code units**

The codegen validator in `value_tokens()` (`primitives.rs`) already implements this correctly. The fix splits the shared parser arm to match:

```rust
GenericString::BoundedString(max_size) => {
    Box::new(move |s| verify(string_literal, move |s: &str| s.len() <= max_size)(s))
}
GenericString::BoundedWString(max_size) => Box::new(move |s| {
    verify(string_literal, move |s: &str| {
        s.encode_utf16().count() <= max_size
    })(s)
}),
```

The original `BoundedString` byte check was already correct — only `BoundedWString` was wrong (it used `s.len()` instead of `s.encode_utf16().count()`). Tests now cover both byte semantics for `BoundedString` and UTF-16 semantics for `BoundedWString`, including astral-plane characters (U+1F600, 2 UTF-16 code units).

---

phil-opp:

The latest commit resolves the concern raised in the earlier automated reviews. The shared `BoundedString | BoundedWString` parser arm is now split so each kind uses the correct metric, matching both the [ROS 2 wide-strings spec](https://design.ros2.org/articles/wide_strings.html) and the codegen validator in `types/primitives.rs`:

- `BoundedString`: `s.len()` (UTF-8 bytes) — matches `value.len()` in codegen
- `BoundedWString`: `s.encode_utf16().count()` (UTF-16 code units) — matches `value.encode_utf16().count()` in codegen

Parser and codegen now agree end-to-end (both `<=`, both per-kind metrics), and the tests cover the boundary where chars and UTF-16 units diverge (astral `😀😀😀` = 6 units). No remaining issues from my side.

---

PR #2366, opened 2026-06-26
Arrow IPC data plane: official wire format, ≤1-copy send, schema-once

phil-opp:
## Motivation

dora's data plane used a hand-rolled Arrow serialization: raw concatenated buffers plus an `ArrowTypeInfo` sidecar (`buffer_offsets`, `validity`, `child_data`, `schema_hash`) carried in every message's `Metadata`. It was not self-describing — nodes had to pre-agree on types — and it was a bespoke protocol to maintain.

This PR switches the data plane to the official Apache Arrow IPC wire format. Messages are now self-describing, decode through arrow's official zero-copy `StreamDecoder`, and the custom sidecar protocol is gone entirely (less code, better interop).

Two hard constraints shaped the design:

1. **≤1 copy on send** (0 when the producer constructs the sample in place). Arrow's official `StreamWriter` always stages the record-batch body in an internal `Vec` (≥2 copies), so the encode side is hand-rolled to write directly into the shared-memory sample, with a fallback to the official writer for rare/complex types.
2. **A latent 2-copy regression** (introduced when the old custom shmem impl was removed): `allocate_data_sample` returned a heap buffer, so publishing did a second `copy_from_slice` into SHM. Large sends were 2 copies regardless of format. This is fixed — producers now allocate and write straight into the zenoh SHM sample, which is moved into `put()`.

## What changed

- **Hand-rolled 1-copy IPC encoder (`arrow_utils::ipc_encode`):** writes the IPC flatbuffer headers and copies each array buffer directly into a caller-provided `&mut [u8]` (the SHM sample). Body buffers are 64-byte aligned so the official `StreamDecoder` reads them zero-copy. Falls back to the official writer for `*View`/`Dictionary`/`Union`/`Map`/`RunEnd`/sliced arrays. Verified by bidirectional round-trip against arrow's official `StreamReader`, a `require_alignment(true)` + pointer-alias zero-copy proof, and a randomized fuzz.
- **Schema-once zenoh path (periodic refresh):** small high-rate messages would otherwise pay ~400 bytes of IPC schema framing each. The publisher re-sends the full stream every N messages (and on schema change) and ships only the schema-less record batch in between; the subscriber decodes in zenoh receipt order via a per-input persistent `StreamDecoder`. This recovers small-message overhead without a new dependency. (Implemented with in-band periodic refresh rather than zenoh-ext `AdvancedPublisher`/`Subscriber` — a persistent `StreamDecoder` rejects EOS-terminated streams, which would force an async schema fetch that collides with the sync receive path.)
- **0-copy `send_output_raw` / Python buffer protocol:** the UInt8 IPC header is pre-written into the (SHM) sample and the caller writes the payload directly into the data region — restoring zero copies for the large-image/tensor path under IPC framing.

## Breaking changes

This targets the 1.0 release with no transition period — the wire format is incompatible with prior versions.

- `Metadata` no longer carries `type_info`; `Metadata::new`/`from_parameters` drop the parameter; `metadata_version` bumped 0 → 1.
- `ArrowTypeInfo` and `BufferOffset` are deleted from dora-message.
- dora-core's metadata module (`ArrowTypeInfoExt` + the SHM buffer-containment helper and its Kani proofs) is removed — the daemon no longer inspects payloads (opaque IPC passthrough).
- `send_typed_output` is removed; `send_output_sample` drops its `type_info` parameter.
- Receivers get a self-describing IPC stream; the raw-buffer decode path and its alignment helpers are gone.
- Net −1291 lines across 9 crates (node API, daemon, coordinator, runtime, CLI, record/replay, C/C++, Python).

## Copies on the send path

| Path | Copies |
|------|-------:|
| `send_output_raw` / Python, large (≥4 KiB → SHM, construct in place) | 0 |
| `send_output_raw` / Python, small (<4 KiB → heap) | 1 |
| `send_output(array)` fast-path, large (SHM) | 1 |
| `send_output(array)` fast-path, small (heap) | 2 |
| Fallback types (View/Dict/Union/Map/RunEnd), large | 2 |
| Fallback types, small | 3 |
| Daemon/TCP fallback (no zenoh subscriber) | +1–2 |

A counting-allocator test harness asserts these (no payload-sized intermediate for the construct-in-place / fast paths; one for the fallback).

## Performance (end-to-end over real zenoh)

| Size | p50 | min | avg | throughput |
|------|----:|----:|----:|-----------:|
| 8 B | 102 µs | 66 µs | 143 µs | 58,071 msg/s |
| 512 B | 73 µs | 38 µs | 79 µs | 59,127 msg/s |
| 2 KB | 67 µs | 34 µs | 65 µs | 42,101 msg/s |
| 16 KB | 71 µs | 37 µs | 102 µs | 61,377 msg/s |
| 400 KB | 73 µs | 41 µs | 83 µs | 22,966 msg/s |
| 3 MB | 82 µs | 46 µs | 123 µs | 849 msg/s (~2.5 GB/s) |

The flat latency curve and the 3 MB round-trip at 82 µs p50 confirm the zero-copy large path: a single 3 MB memcpy alone is ~50–100 µs, so a multi-copy path could not hit this. Small high-rate messages stay at ~60k msg/s thanks to schema-once.

## Validation

Full workspace `cargo test --all` (excl. Python/maturin crates): 1250 tests, 0 failures. `clippy --all -D warnings`, `fmt --all --check`, `check --all`/`--examples` all green; the PyO3 crate compiles. The `rust-dataflow` and `benchmark` examples run end-to-end over real zenoh. An adversarial review caught and fixed a data-corruption bug where a truncated/corrupt batch poisoned the per-input decoder (it now resets, dropping until the next periodic full stream re-primes).


phil-opp:
**Automated review by Claude** — _This is a fully automated review with no human oversight. Please verify any findings before acting on them._

I reviewed the full diff. The Arrow IPC migration looks consistent across all crates — no leftover `type_info` / `required_data_size` / `copy_array_into_sample` / `send_typed_output` call sites — and the encoder/decoder, the schema-once decoder reset-on-error, and the zenoh-SHM-backed buffer lifetimes are well covered by tests (official-reader oracle, alignment/aliasing zero-copy checks, deterministic fuzz, decoder-reset regression).

Two non-blocking notes, both on the lower-severity edges:

**1. Record/replay collapses zero-length *typed* arrays to `Null`.**
On the record side (`binaries/record-node/src/main.rs`), payload presence is decided with `if arrow_data.is_empty()` (i.e. `len() == 0`) → records `None`; on replay (`binaries/replay-node/src/main.rs`) an absent payload is reconstructed as `NullArray::new(0)`. So a recorded empty-but-typed array (e.g. a length-0 `Float32Array`) replays as a `Null` array, dropping the declared type — a case the deleted `type_info` path used to preserve (#2027/#2083). Everywhere else the new "empty payload ⇒ unit array" rule is safe, because the live send paths IPC-encode even zero-length typed arrays into a non-empty, self-describing stream; only the record path short-circuits on `is_empty()`. Worth either encoding an IPC stream whenever the array is non-`Null`-typed (regardless of `len`), or documenting that recordings collapse empty typed arrays to the unit type.

**2. SHM `put` failure is now a silent drop with no daemon fallback.**
In `zenoh_publish`'s `FinalizedSample::Shm` arm (`apis/rust/node/src/node/mod.rs`), if `put` fails the SHM buffer has already been consumed, so the message is lost with only a `warn!` — where the non-SHM path can fall back to the daemon. This is documented as a deliberate zero-copy trade-off and only affects large/SHM payloads, but it does make large-output delivery best-effort on an SHM put error. Flagging only so a maintainer can sign off that this is the intended semantics.

Neither of these is a blocker; the core migration reads as sound. No other issues found.

---

phil-opp:
**Automated review by Claude** — _This is a fully automated review with no human oversight. Please verify any findings before acting on them._

I re-reviewed the full diff at `15b81a5b`. The two notes from my previous review are addressed: record/replay now IPC-encodes every non-unit array so empty *typed* arrays keep their type (`92a7b4f9`), and the SHM `put` best-effort drop is now documented as intended. The bigger change since then is the **schema-once redesign** (`fee74f9e`) — the periodic in-band refresh was replaced wholesale by a zenoh-ext `AdvancedPublisher`/`AdvancedSubscriber` `@schema` subtopic with a depth-1 cache. The prior review predates this, so I focused on it.

Overall the new path looks sound: the schema-only priming message (no record batch, no EOS) avoids the persistent-`StreamDecoder` EOS rejection, the dedicated `@schema` subscriber keeps schema fetching off the sync data callback, and `set_schema` now *retaining* the schema gives instant local re-prime after a corrupt-batch soft-reset (strictly better than the old up-to-64-message poison window). The `SCHEMA_HASH` batch-vs-full-stream disambiguation, the daemon fallback always carrying a full stream (never a batch slice), and the SHM buffer lifetimes still check out.

One issue worth a maintainer's eyes, plus one minor hardening note:

**1. Schema oscillation can silently drop inputs.** `InputDecoder` retains a single schema (`set_schema` replaces), the `@schema` `AdvancedPublisher` cache is depth-1, and the producer publishes the schema only on change. If an output's schema alternates (A→B→A…), a data batch tagged `hash(A)` reaching a subscriber that is currently primed for B is dropped (`decode_batch` → `Ok(None)`) until schema A is re-delivered over the separate `@schema` subscriber — and because the data and schema topics are delivered independently, that re-priming can lag the batch. On the `CongestionControl::Drop` data plane this is "acceptable loss," but it's a silent, intermittent input-loss mode distinct from the documented one-time late-join window. If alternating schemas on a single output are a supported case, retaining a small set of recent schemas (keyed by hash) rather than just the latest would avoid it; otherwise it's worth a comment that per-output schema must be stable.

**2. `StreamDecoder::decode` progress assumption (minor/defensive).** `decode_arrow_ipc_zero_copy`, `decode_one_batch`, the priming path, and `cli/src/command/topic/echo.rs` all loop `while !buffer.is_empty()` and break only on a decoded batch. This relies on `decode` always consuming bytes; if it ever returned `Ok(None)` without advancing on a crafted/truncated payload, the loop would spin on the zenoh IO worker. Arrow buffers partial input internally so it terminates in practice (and `MAX_IPC_BYTES` bounds the size), but an explicit "bail if the buffer length didn't shrink" check would harden the untrusted-peer path.

Neither blocks; the migration reads as well-engineered and is unusually well tested. No other issues found.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human oversight. Please verify independently before acting on it.

The diff is unchanged since the last automated review, so I won't repeat that analysis. One point can now be **escalated from the earlier "schema oscillation" note to a concrete regression**, because the open question it left ("*if* alternating schemas on a single output are a supported case") is answered by the code itself.

### Schema-once silently drops messages on outputs that multiplex multiple Arrow schemas (service/action/streaming)

`send_output`'s own comment documents this as a supported pattern:

> Service, action, and streaming patterns legitimately multiplex multiple Arrow schemas through a single output — a service server may reply with different response shapes for different request types (`apis/rust/node/src/node/mod.rs:831-836`).

Those small responses go through schema-once **with no exclusion**: `publish_schema_once` is gated only by `avec.len() < zenoh_zero_copy_threshold` (`node/mod.rs:1245`), not by `carries_pattern_correlation` (which only gates the runtime type check).

The receive side retains exactly **one** schema — `InputDecoder { schema_hash: Option<u64>, primed_schema: Option<(u64, ArrowBuffer)> }` (`node/arrow_utils/ipc_encode.rs:568-576`) — and a batch whose `_schema_hash` it isn't currently primed for returns `Ok(None)` and is dropped (`decode_batch`, `ipc_encode.rs:633-651`). The producer is symmetric (single `last_schema_hash` per output, depth-1 `@schema` `AdvancedPublisher` cache).

The drop is triggered by the **QoS asymmetry between the two topics**: the data plane is `express(true)` + `Priority::RealTime` + `CongestionControl::Drop` (`node/mod.rs:1122-1124`), while `@schema` is plain `CongestionControl::Block` (`node/mod.rs:1975`). On every schema switch (each message in an alternating service workload), the switched batch can reach the consumer before its schema sample and is dropped. The retained-schema local re-prime doesn't rescue it, because alternation constantly overwrites the single retained schema. The previous self-contained-stream format carried the schema in every message and had no such race — so for schema-switching outputs this is message **loss** that didn't exist before, not merely a late-join window.

Suggested fix: keep a small hash-keyed LRU of recent schemas in `InputDecoder` (and track multiple hashes / deepen the `@schema` cache producer-side), or exclude `carries_pattern_correlation` outputs from schema-once and send them as standalone full streams.

### Secondary (lower severity, not previously raised)

`publish_schema_once` does `publisher.put(schema_bytes).wait()` with `CongestionControl::Block` **inline on the user's `send_output` call** (`node/mod.rs:1932` + the `Block` publisher at `:1975`). For a frequently-schema-switching output this places a blocking publish on the hot send path; if `@schema` is congested, the producer can stall. Consider a non-blocking/bounded policy for the schema put, or moving it off the synchronous send path.

Everything else I traced looks correct: hash consistency producer↔consumer, batch framing / `batch_slice`, the `Mutex`-shared decoder's memory safety, SHM/`ZBytes` buffer lifetimes, 64-byte alignment with realign fallback, decode-loop termination, and the intentional, version-checked wire-format break.

---

phil-opp:
**Automated review by Claude** — _this is a fully automated review with no human oversight; please verify independently before acting on it._

Re-reviewed at `0e449067`. The schema-once regression escalated in the previous review is **resolved**:

- `schema_once_eligible` (`apis/rust/node/src/node/mod.rs`) now excludes outputs carrying `request_id`/`goal_id`/`goal_status`, so service/action outputs that multiplex multiple response schemas through one output send **full self-describing streams** (each batch decodes standalone regardless of schema arrival order) instead of schema-less batches that could reach a consumer primed for a different schema and be silently dropped. Streaming stays eligible — every chunk of a stream shares one schema — so it remains the intended high-rate beneficiary, and the inline `@schema` blocking `put().wait()` now fires only on schema change (once per stream), which also addresses the secondary "blocking publish on the hot send path" note. The gate is unit-tested.
- The `StreamDecoder::decode` no-progress assumption is hardened: the four decode loops (`decode_arrow_ipc_zero_copy`, `prime_with_schema`, `decode_one_batch`, and CLI `topic echo`) now bail if a pass consumes no bytes, so a crafted/truncated peer payload can't spin the zenoh IO worker.

I re-traced producer↔consumer schema-hash consistency (FNV over identical schema-block bytes), the `@schema`-vs-data key separation (concrete data key, no wildcard cross-delivery), `decode_batch`'s soft-reset + retained-schema re-prime, the `ZBytes`/SHM buffer lifetimes behind `Buffer::from_custom_allocation`, the 0-copy UInt8 header arithmetic, the daemon-fallback/record-replay/CLI full-stream paths, and the `metadata_version` receive check. **No new issues found.**

---

phil-opp:
## Benchmark: this PR vs `main`

Ran the `examples/benchmark` dataflow (latency + throughput across payload sizes) and the `arrow_framing` criterion micro-bench, comparing this branch against the baseline.

**Setup.** AMD Ryzen 9 5900X. Baseline = `origin/main` (`8556977`) in a separate worktree (cleanest counterfactual, since this branch already merged `main` on top of its perf work). Both trees release-built; the benchmark node/sink/dataflow are **byte-identical** between them, so the only variable is the data plane. End-to-end numbers are the **median of 4 interleaved runs** (alternating branches, warmup discarded).

> ⚠️ **Caveats:** CPU governor was `powersave`, and N=4. Small-payload numbers are at the ~50–70 µs scale and **noise-dominated** — treat ±10% there as no signal. The unambiguous movement is all at ≥400 KB, which is what this PR targets.

### Latency (median, µs — lower is better; Δ<0 = PR faster)

| size | main p50 | PR p50 | Δ p50 | main p99 | PR p99 | Δ p99 |
|---|---:|---:|---:|---:|---:|---:|
| 8B | 66.6 | 60.4 | −9% | 127.3 | 90.0 | −29% |
| 64B | 55.4 | 57.1 | +3% | 89.5 | 99.7 | +11% |
| 512B | 51.6 | 54.3 | +5% | 98.8 | 94.1 | −5% |
| 2KB | 62.9 | 60.7 | −4% | 105.2 | 86.7 | −18% |
| 4KB | 62.5 | 69.2 | +11% | 129.0 | 102.5 | −21% |
| 16KB | 69.4 | 65.4 | −6% | 103.8 | 92.1 | −11% |
| 40KB | 67.8 | 66.4 | −2% | 109.3 | 103.7 | −5% |
| **400KB** | 91.2 | 62.9 | **−31%** | 153.6 | 102.4 | **−33%** |
| **3MB** | 259.5 | 64.2 | **−75%** | 582.2 | 95.2 | **−84%** |

3MB p50 per run: main `388 / 224 / 220 / 295 µs` vs PR `64 / 64 / 63 / 66 µs` — PR is flat and ~4× faster.

### Throughput (median msg/s — higher is better; Δ>0 = PR faster)

| size | main | PR | Δ |
|---|---:|---:|---:|
| 8B | 37,730 | 32,248 | −14% |
| 64B | 36,922 | 39,689 | +8% |
| 512B | 36,809 | 38,608 | +5% |
| 2KB | 40,705 | 39,597 | −3% |
| 4KB | 41,556 | 38,526 | −7% |
| 16KB | 38,214 | 37,613 | −2% |
| 40KB | 35,668 | 33,087 | −7% |
| **400KB** | 17,258 | 22,366 | **+30%** |
| **3MB** | 1,550 | 872 | **−44%** ⚠️ |

3MB per run: main `1450 / 1573 / 1578 / 1528` vs PR `901 / 893 / 852 / 840` — tight clusters, so the regression is real, not variance.

### Micro-bench (`arrow_framing`, criterion) — the mechanism

`ipc_official` / `streamreader_copy` ≈ the path `main` uses; `ipc_fast` / `streamdecoder_zerocopy` ≈ this PR's path. (Absolute numbers from a single run; criterion's own "regressed/improved" change-lines compare to a stale local baseline and are *not* the PR-vs-main signal.)

| op | size | main path | PR path | speedup |
|---|---:|---:|---:|---:|
| encode | 64 KiB | 3.78 µs | 1.83 µs | **2.1×** |
| encode | 4 MiB | 3.33 ms | 97.3 µs | **34×** |
| decode | 64 KiB | 2.52 µs | 855 ns | **2.9×** |
| decode | 4 MiB | 562 µs | 857 ns | **~650×** |

The zero-copy decode is **O(1)** — ~855 ns at 1 KiB, 64 KiB *and* 4 MiB — because it aliases the buffer instead of copying. That's what drives the large-payload latency win.

---

### Two things worth a look before merge

**A. Latency-vs-throughput trade-off at 3 MB.** Single-message latency is ~4× better, but sustained throughput (100×3 MB burst) is ~44% *worse*, reproducibly. The micro-bench shows it's **not** CPU framing cost (encode and decode are both faster), so the leading hypothesis is **SHM buffer lifetime**: the receiver's zero-copy alias pins each shared-memory segment until the decoded array is dropped, so the pool recycles slower under burst and back-pressures the sender — whereas `main` copies out and releases the segment immediately. This is a hypothesis from the measurements, not yet code-confirmed; might be worth checking SHM pool sizing / segment-return timing.

**B. Zero-length outputs appear to be dropped on receive.** Deterministic across all 4 runs: the node sends 100 latency + 100 throughput messages at `data_len = 0` (the `0B` bracket); `main` delivers them (sink records `0B`, n=100), but on this branch **nothing for `0B` arrives** (the sink's first delivered bracket is `8B` with n=100, not 200, so they aren't merged either) — and there's no error in the logs.

It looks like a 0-row `send_output_raw` builds a *non-empty* IPC stream, which `decode_arrow_ipc_zero_copy` then rejects as `"Arrow IPC stream contained no record batches"` (`arrow_utils.rs:141`), and that error path drops the event. Meanwhile a *truly* empty payload is separately remapped to a unit array (`decode_zenoh_sample` `mod.rs:1264`, `RawData::Empty` in `data_conversion.rs:18`) — so the empty/0-row-array case falls between the two handlers. Emitting an empty array is a legitimate pattern (e.g. "0 detections this frame"), so this seems like a real gap — an empty-array send→receive round-trip regression test would pin it down.

<sub>Numbers are median-of-4 on a `powersave`-governor desktop; absolute small-payload values are noisy, but the ≥400 KB deltas and the 0 B behavior reproduced on every run.</sub>


Created by Claude Code

phil-opp:
Re-reviewed the three commits added since the last review (`0e449067` → `0f96b97c`): the empty-array fix (`b1d06b27`), the non-blocking SHM alloc (`41eb9e42`), and the large-output daemon routing (`0f96b97c`). Both issues raised in the earlier benchmark comment look resolved, and I traced the fixes rather than taking the commit messages at face value. No new correctness, data-corruption, or memory-safety issues found — the new commits look safe to me.

**Empty (0-row) batch fix is correct.** Checked against arrow-ipc 58.3.0's `StreamDecoder::decode`. Without a trailing marker, a schema-less batch leaves the decoder in `Body{len=0}` and the `while !buffer.is_empty()` loop exits with `Ok(None)` — the empty array is dropped, exactly the reported symptom. Appending the 8-byte EOS marker makes the `Message` arm's strict `buffer.len() > len` and the `Body` arm's `buffer.len() >= body_length` both hold, so the empty batch is flushed and returned *before* the EOS bytes are consumed. The decoder therefore never reaches `DecoderState::Finished` (which would make the next `decode()` error with "Unexpected EOS"), so the persistent per-input decoder stays reusable. The non-empty path is also safe. The interleaved empty/non-empty reuse test and the `batch_slice` production-path test cover this.

**Large-output daemon routing is correct.** A large heap-fallback payload (SHM unavailable / pool momentarily full) now returns `NotPublished(Vec)` → `Delivery::Daemon`, decoded standalone via `decode_arrow_ipc_zero_copy`. It never touches the schema-once `InputDecoder`, so it can't desync the persistent decoder, and large messages are always full self-describing streams (schema-once is sub-threshold only). No double-delivery — the Zenoh and Daemon arms are mutually exclusive.

Two non-blocking notes:
- The `_framing` metadata key (`FRAMING`/`FRAMING_ARROW_IPC`) is still stamped by every producer (node/daemon/coordinator/python) but is no longer read on decode now that all payloads are IPC — it's effectively dead and could be dropped.
- On the current head, the `Test`, `E2E Tests`, `Semantic contract tests`, and `Benchmark regression check` CI jobs show as **skipped** (merge-queue gating), so the new empty-array and daemon-routing regression tests haven't actually run in CI for this commit. Worth confirming the merge queue runs the full suite before merge.

---
🤖 This is a fully automated review by Claude. No human has reviewed this comment; please double-check before acting on it.

---

phil-opp:
**Automated review by Claude** — ⚠️ _This is a fully automated review, no human in the loop. Please verify before acting on it._

I re-reviewed the current head (`0f96b97c`) against the diff and the surrounding code. One issue that I don't think earlier rounds covered:

### `dora topic echo` / `info` can't decode schema-once outputs

When `_unstable_debug.enable_debug_inspection` is on (the prerequisite for `dora topic` to work at all), the daemon declares its **own** zenoh subscriber on each output's *data* topic to feed `topic echo`/`info`/`hz` (`binaries/daemon/src/lib.rs` ~2892):

```rust
let topic = zenoh_output_publish_topic(dataflow_id, &node.id, &output_id);
let subscriber = self.zenoh_session.declare_subscriber(topic.clone())...
// callback forwards sample.payload() verbatim as DebugTopicData { data, .. }
```

This subscriber listens **only** on `zenoh_output_publish_topic` and holds **no** `InputDecoder` — it never subscribes to the `@schema` subtopic. But for schema-once-eligible outputs (`schema_once_eligible`, `apis/rust/node/src/node/mod.rs` ~2078: payload `< zenoh_zero_copy_threshold` (4 KiB) and no `request_id`/`goal_id`/`goal_status`), `zenoh_publish` puts only the **schema-less** record batch on the data topic (`batch_slice(&avec)`, ~1324) and publishes the schema separately on `@schema`. Those schema-less bytes are forwarded verbatim to the CLI, which decodes each buffer with a fresh `StreamDecoder` via `decode_arrow_ipc_zero_copy` (`binaries/cli/src/command/topic/echo.rs` ~177) — and that fails, because the batch has no leading Schema message.

Net effect (traced end-to-end):
- `dora topic echo` prints `invalid data on <output>` and shows no content
- `dora topic info` reports the data type as `unknown`
- `dora topic hz` still works (it only counts frames)

This hits ordinary small outputs (scalars, poses, small arrays < 4 KiB with no service/action metadata) — e.g. `ticker/count` in `tests/fixtures/zenoh-debug-dataflow.yml`. Large outputs (≥ 4 KiB → SHM → full self-describing stream) and service/action outputs (excluded from schema-once) are unaffected. It's a regression vs. the pre-PR behavior, where every data-topic message was a self-contained stream.

The node→node receive path is fine: `event_stream` declares an `@schema` subscriber and primes a per-input `InputDecoder`. The gap is that the daemon's *separate* debug subscriber (and the `topic echo` CLI decoder) were never given the same schema priming.

Possible directions: have the daemon debug path also subscribe to `@schema` and keep a per-output `InputDecoder` (mirroring `event_stream::declare_schema_subscriber`) and decode before forwarding; or exclude an output from schema-once while it has debug watchers; or re-emit a full self-contained stream on the debug path. The extra decode/copy is acceptable on an inspection-only path.

Everything else I checked held up — the hand-rolled IPC encoder (offsets, 64-byte alignment, overflow guards), the fast-path/fallback type classifier (`*View`/Dictionary/Union/Map/RunEnd/sliced → official writer), the decoder reset / no-progress guards, the empty-batch EOS fix, the metadata-version bump, and the 0-copy `send_output_raw`/Python buffer-protocol state machine all look correct and are well covered by tests. So this debug-path decode is the one outstanding item I found.

---

phil-opp:

No new data-corruption or memory-safety issue: the high-risk areas all check out — 64-byte buffer alignment and overflow guards in the encoder, the decoder soft-reset on corrupt batches, producer↔consumer schema-hash consistency, malformed-input bounds checks (no unwraps on network data), and the fast-path-vs-fallback type classifier (including nested `View`/`Dictionary` children caught by recursion).

One robustness gap I didn't see raised, on the schema-once receive path:

If the `@schema` `AdvancedSubscriber` fails to declare in `declare_schema_subscriber` (`apis/rust/node/src/event_stream/mod.rs`, ~L1243), the code only logs a `warn!` and returns — but the data subscriber is still declared. For every schema-once-eligible input (any plain output < 4 KiB), the per-input decoder is then never primed, so `decode_batch` returns `Ok(None)` for every batch and the callback drops it silently. That input receives **nothing** for the lifetime of the dataflow, with only a one-time startup warning and no `FatalError`. Unlike the documented late-join window, there's no recovery path, and the producer has no way to learn that the consumer's schema plane is dead.

The trigger — a `@schema` subscriber declare failure while the data subscriber succeeds — is rare, but the failure mode (permanent, silent loss of an input that otherwise looks healthy, on a breaking change targeting 1.0) is severe enough to handle explicitly: surface a `FatalError` on declare failure, or have the consumer fall back to requesting full streams for any input that stays unprimed.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review (no human in the loop); please verify before relying on it.

Re-reviewed after the commits pushed since the last review round. The two newest non-merge commits resolve the two outstanding findings:

- `c4a5ba6c` ("fail loudly when the @schema subscriber dies") addresses the previously-flagged issue where a failed `@schema` subscriber declaration would silently drop every schema-once input for the dataflow's lifetime — it now latches a fatal error rather than dropping silently.
- `e6f78c2c` ("decode schema-once outputs in dora topic via an @schema subscriber") addresses the earlier finding that `dora topic echo`/`info` couldn't decode schema-once outputs.

I independently re-traced the highest-risk areas against the actual diff: the hand-rolled IPC encoder (buffer offset/alignment, `checked_add` + `MAX_IPC_BYTES` overflow guards, struct/fixed-size-list child slicing, the `offset()==0` fast-path gate), the schema-once decode state machine (per-input persistent `StreamDecoder` reuse, soft-reset-and-re-prime on a corrupt batch, empty-batch/EOS handling), and schema-hash agreement across producer/node/daemon (all three hash the identical schema-block bytes via the shared `fnv1a`). The two new fixes look correct.

No new issues found. The only un-actioned item is the cosmetic dead `_framing` metadata key, which is non-blocking.

---

phil-opp:
@claude please resolve the merge conflict

phil-opp:

Re-reviewed at head `3503688`. The `2026-07-01` change is only a merge of `origin/main` resolving the `topic/info.rs` conflict; the resolution is clean (decodes the daemon-rebuilt full IPC stream via `StreamReader`), and I re-confirmed the previously-cleared items stay fixed. I won't repeat prior rounds. I did find **two new schema-once loss modes** not raised before, both sharing the same root: the batch (data plane) can arrive at a consumer whose decoder isn't primed yet, and `decode_batch` → `Ok(None)` then drops it silently. The FatalError safety net added in `c4a5ba6c` does **not** cover either, because it only fires when `schema_plane_failed` is set, and that flag is set *only* on `@schema` **declare** failure (`event_stream/mod.rs:1263`, `:1297`; gated at `:447-466`).

**1. Silent, permanent input loss when `@schema` declares successfully but is never primed (medium–high).**
The producer publishes the schema only on hash change (`node/mod.rs:1989`), so a stable-schema output emits its schema essentially once. A consumer whose zenoh-ext history query / depth-1 cache misses that single emission (timing / cache handshake) gets `Ok(None)` forever — every batch dropped, no FatalError, only a warn. Prior review #8 suggested a second option — "fall back to full streams for any input that stays unprimed" — which wasn't implemented, so the net is narrower than the failure mode it was meant to close. (Confidence moderate that it manifests in practice, since it depends on zenoh-ext cache reliability, but the gap in the net is concrete.)

**2. First (and only) message of a low-rate/one-shot eligible output can be dropped by a QoS race (medium).**
Within one `send_output`, the producer publishes the schema on `@schema` with `CongestionControl::Block` (`node/mod.rs:2039`), then immediately puts the schema-less batch on the data topic with `express(true)` + `Priority::RealTime` + `CongestionControl::Drop` (`node/mod.rs:1134-1136`, put at `:1330`) — **with no delivery barrier between them**. The express/RealTime batch can overtake the non-express Block schema on the very first message, so the consumer's decoder isn't primed yet and the batch is dropped. For a high-rate stream this is one lost message out of many (negligible). For a **one-shot or low-rate** small output (a startup/config broadcast, a single "0 detections" result), that lost message is the *only* message. The design comment at `node/mod.rs:1284-1287` acknowledges the Drop-vs-Block race only for the excluded alternating-schema case — it doesn't recognize that the same race costs the first message of *every* eligible output, including streaming (whose first chunk is exactly the priming window). This is a regression vs. the pre-PR self-contained-stream format, where the first message decoded standalone; single-message loss wouldn't fail the `rust-dataflow`/benchmark smoke runs, which likely explains why it went unnoticed.

Everything else I traced (hand-rolled encoder alignment/overflow guards, zero-copy decode + no-progress bail, schema hashing consistency, `type_info` removal, Python `SampleBuffer` UAF protection, daemon debug path) looked correct. The remaining `_framing` dead metadata key is cosmetic, as already noted.

---

phil-opp:

The diff is unchanged since the last review (head `3503688` is just a `main` merge resolving `topic/info.rs`), so the two schema-once loss modes raised on 2026-07-02 still stand. The encoder, decoder state machine, empty-batch EOS handling, and zero-copy buffer backing all hold up under a fresh read. One **new, higher-severity** issue on the same path that earlier rounds didn't cover:

**The internal `_schema_hash` key leaks into user-visible input metadata and silently drops re-forwarded large / service-action outputs.**

`EventItem::ZenohInput` → `Event::Input` (`apis/rust/node/src/event_stream/mod.rs:1159`) hands the raw zenoh attachment metadata to the user without stripping `_schema_hash`, which `publish_schema_once` inserted for schema-once (small) inputs. A node that forwards `metadata.parameters` — a standard dora pattern; the replay node does exactly `node.send_output(id, metadata.parameters, …)` — while emitting a **large** payload (≥ `zero_copy_threshold`) or a **service/action** output takes the non-schema-once branch in `zenoh_publish`, where the stale `_schema_hash` is *not* overwritten and rides on the attachment (`node/mod.rs:1328`). The receiver routes purely on `SCHEMA_HASH` presence (`decode_zenoh_sample`, `event_stream/mod.rs:1319`) → `decode_batch(stale_hash)` against a full self-describing stream → `schema_hash` / `primed_schema` mismatch → `Ok(None)` → the callback silently drops a valid message (`event_stream/mod.rs:447`). Net: permanent, error-free loss of a large output whenever its parameters were inherited from a schema-once input.

Repro: a transform node receives a small high-rate input `A` (schema-once, `_schema_hash=H`), then emits `send_output(out, A.metadata.parameters, big_tensor)` with `big_tensor` ≥ 4 KiB. The output goes out as a full stream over SHM but with `_schema_hash=H` in the attachment; every zenoh subscriber routes it to `decode_batch(H)`, gets `Ok(None)`, and drops it. (Record/replay happens to be safe today only because it preserves the size class, so the overwrite branch is hit — the general forwarding case is not.)

Suggested fix: strip `_schema_hash` (and the already-dead `_framing`) from parameters on receive, or drop internal `_`-prefixed protocol keys in `send_output`, or clear `_schema_hash` from the outgoing attachment in the large / pattern-correlated branches of `zenoh_publish`.

Related, lower severity: `publish_schema_once` returns `Some(attachment_with_hash)` even when the `@schema` publisher fails to declare or its `put` errs (`node/mod.rs:1975-2015`) — the batch then ships tagged with a hash whose schema was never published, and the consumer-side `schema_plane_failed` FatalError latch can't catch it (that flag is set only on the *consumer's* own `@schema` declare failure). So this is a producer-side variant of the same silent-permanent-loss family, distinct from the consumer-side case the `c4a5ba6c` net was built for.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; treat it accordingly. I spent this pass on the hand-rolled encoder/decoder and the schema-once dual-plane protocol. The IPC encoder core (fallback detection recursing into nested children, buffer offset/length bookkeeping, 64-byte body alignment, empty/0-row + EOS handling, overflow guards, the zero-copy decode proof) looks correct and unusually well-tested — no new correctness or memory-safety issue there.

One new finding, in `publish_schema_once` (`apis/rust/node/src/node/mod.rs:1996-2014`):

**When the `@schema` publish fails, the batch is still shipped tagged with a schema hash that was never delivered → silent, undecodeable drop.**

```rust
match publisher.put(schema_bytes).wait() {
    Ok(()) => { last_schema_hash.insert(output_id.clone(), hash); }
    Err(e) => { tracing::warn!(...); }   // logs, then falls through
}
// ... unconditionally:
metadata.parameters.insert(SCHEMA_HASH.to_string(), Parameter::Integer(hash as i64));
bincode::serialize(&metadata).ok()       // returns Some(hash-tagged attachment)
```

On the `Err` path the schema was never put on the `@schema` subtopic (and, since the put failed, it isn't in the `AdvancedPublisher` cache either), yet the function still returns `Some(...)` with `SCHEMA_HASH` set and the caller then sends the *schema-less* batch on the data topic. The consumer routes on `SCHEMA_HASH` presence to `decoder.decode_batch(hash)`, but its decoder for that hash was never primed → `Ok(None)` → the callback silently drops the message. Not updating `last_schema_hash` correctly makes the *next* message retry the schema put, so a stable-schema output self-heals — but the current message (and any until a put succeeds) is lost with no error surfaced. The existing `schema_plane_failed` FatalError can't catch this; it only trips on the *consumer's own* `@schema` declare failure, not a producer-side put failure.

Fix: on `Err`, fall through to sending the full self-describing stream (return `None`) instead of a hash-tagged schema-less batch — degrade to "unframed but decodable" rather than "framed but undecodable."

Minor / FYI: the `mergeable_state` is currently `blocked` because the **Typos** gate is red on a trivial typo in the daemon debug path (`mis-decode`) — worth fixing so the merge-queue-gated Test/E2E/contract jobs actually run at this head; they're currently skipped, so the new empty-array / routing / schema-once regression tests haven't executed in CI on this commit.

---

phil-opp:
Re-reviewed the two non-merge commits pushed since the last review — `faaba0d` (schema-once loss-mode fixes) and `b799f5a` (audit/lockfile only). I traced `faaba0d` against the code rather than the commit message, and it closes the silent-loss modes raised earlier. **No new findings.**

- **First-message / QoS race** (express batch overtaking the non-express `@schema` publish): the producer now ships a full self-describing stream for the first message of each schema, on schema change, after a failed `@schema` put, and on a 5s refresh; the receiver primes its per-input decoder **in-band, in data-plane order**, so priming lands before the schema-less batches that reference it — the race is closed without depending on the separate `@schema` plane.
- **Stale `_schema_hash` leak**: `strip_internal_parameters` is now applied on both the zenoh and daemon receive paths, and `SCHEMA_HASH` is removed from outgoing parameters, so a forwarded (e.g. replay) hash can no longer ride onto a large/service output and get hash-mismatched and dropped.
- **Producer-side failed put**: a failed data-plane put now forgets the output's schema-once state so the next message re-primes; a failed `@schema` put degrades to "full stream every message" (decodable) rather than "hash-tagged but undecodable".
- **Schema retention**: `InputDecoder` retains up to 8 schemas (LRU), so in-band priming from one schema's full stream can't clobber a schema that later batches still reference.
- **`@schema`-declare failure** now surfaces a `FatalError` only after a 3×-refresh grace window, avoiding both the old permanent silent-drop and a spurious kill.

Sub-5s low-rate eligible outputs now effectively send a full stream every message, which is correct (they decode standalone) and just forgoes the schema-once saving where it doesn't matter. From my read these commits look safe to merge with respect to the earlier concerns.

---
🤖 _This is a fully automated review by Claude. It may be incorrect — please verify independently._

---

PR #2373, opened 2026-06-26
fix(message): validate node filter in dora/logs input parsing

phil-opp:

## Issue

`InputMapping::from_str` parses `dora/logs/{level}/{node_id}` filters in `libraries/message/src/config.rs`, but constructed the node id directly:

```rust
Some((level, node)) => (Some(level), Some(NodeId(node.to_owned()))),
```

This bypasses `validate_node_id` (`libraries/message/src/id.rs`), which every other `NodeId` construction path enforces. Because `rest.split_once('/')` keeps **every slash after the first** inside `node`, an input such as `dora/logs/info/a/b` produced a `NodeId` containing `/` — a character the NodeId invariant explicitly forbids ("NodeIds must NOT contain `/`"). Spaces and other invalid characters were likewise accepted.

Such a filter can never equal a real, validated node id, so instead of erroring it silently matched **nothing** (the daemon compares it by equality against each `log_message.node_id` at `binaries/daemon/src/lib.rs:4355`). A user typo became a silent no-op rather than a clear error.

## Fix

Parse the node segment through `NodeId::from_str` so invalid filters are rejected at parse time:

```rust
let node_id = node.parse::<NodeId>().map_err(|e| e.to_string())?;
(Some(level), Some(node_id))
```

## Validation

- New regression test `parse_logs_rejects_invalid_node_filter` (rejects `a/b` and `bad node`).
- `cargo test -p dora-message --lib config::` — 43 passed.
- `cargo clippy -p dora-message -- -D warnings` — clean.

---

### DIFF
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -295,7 +295,17 @@ impl FromStr for InputMapping {
                 Some(("logs", rest)) => {
                     // dora/logs/{level} or dora/logs/{level}/{node_id}
                     let (level_str, node_filter) = match rest.split_once('/') {
-                        Some((level, node)) => (Some(level), Some(NodeId(node.to_owned()))),
+                        Some((level, node)) => {
+                            // Validate the node segment instead of constructing a
+                            // `NodeId` directly: `rest.split_once('/')` keeps every
+                            // slash after the first inside `node`, so an input such
+                            // as `dora/logs/info/a/b` would otherwise build a NodeId
+                            // containing `/` -- a value `validate_node_id` forbids and
+                            // that no real node id can ever equal, silently producing
+                            // a filter that never matches.
+                            let node_id = node.parse::<NodeId>().map_err(|e| e.to_string())?;
+                            (Some(level), Some(node_id))
+                        }
                         None => {
                             if rest.is_empty() {
                                 (None, None)
@@ -864,6 +874,19 @@ mod tests {
     fn parse_logs_invalid_level() {
         let result: Result<InputMapping, _> = "dora/logs/banana".parse();
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn parse_logs_rejects_invalid_node_filter() {
+        // A node segment with an extra `/` (kept by `split_once`) is not a valid
+        // NodeId and must be rejected rather than silently yielding a filter that
+        // can never match a real (validated) node id.
+        let result: Result<InputMapping, _> = "dora/logs/info/a/b".parse();
+        assert!(result.is_err(), "node filter `a/b` must be rejected");
+
+        // A node segment containing a space is likewise invalid.
+        let result: Result<InputMapping, _> = "dora/logs/info/bad node".parse();
+        assert!(result.is_err(), "node filter `bad node` must be rejected");
     }
 
     #[test]


phil-opp:

This PR is one of a batch opened by an automated Claude code-review run over `origin/main` (`89c2c8d`). All are machine-generated; please review each independently before merging.

### PRs by category

**Correctness**
- #2373 — `fix(message)`: validate node filter in `dora/logs` input parsing (bypassed `validate_node_id`, allowed `/`/spaces → silent no-op filter)
- #2374 — `fix(node)`: `checked_add` guard against offset overflow in the Arrow alignment check (peer-controlled `buffer.offset` → debug panic / release misdecision)
- #2375 — `fix(memory-pool)`: don't log cleanup success when frees failed; report accurate `released_count`
- #2378 — `fix(cli)`: compute `dora topic info` bandwidth over the actual collection window (under-reported when a topic closes mid-window); also drops a dead `else` branch

**Security / robustness**
- #2376 — `fix(coordinator)`: prune expired entries from the per-IP rate limiter (unbounded `HashMap` growth → slow leak / DoS surface); the now-unreachable reset branch is removed (simplification)

**Documentation**
- #2377 — `docs(mavlink2-bridge)`: add a compile-checked doctest for the `MavlinkArrow` round-trip contract

Each PR adds a regression test and was verified per-branch with `cargo fmt`, `clippy -D warnings`, and the affected crate's tests. The combined diff also passed the `/review` and `/simplify` gates (the only finding — a dead rate-limiter branch — was folded into #2376).

### Rejected / not pursued (with reasons)

- **Arrow-IPC send path allocates+copies an already-owned `Vec`** (`node/mod.rs`) — real extra copy, but it's an opt-in warm path that overlaps the in-flight Arrow-IPC rewrite (#2366); skipped to avoid churn/conflict.
- **`DataSample::len` is a redundant field** (`node/mod.rs`) — removing it is invasive for marginal benefit and it is still read by `Deref`; not clearly dead.
- **record-node serializes a re-packed buffer but keeps the original `type_info`** (`record-node/src/main.rs`) — could not prove a current defect: the original offsets were themselves produced by the same packer, so the common case round-trips; latent at most.
- **`dora topic info` issues a duplicate coordinator `resolve`** (`topic/info.rs`) — minor redundant network request; not worth a standalone PR.
- **`LogSubscriptionFilter` `Display` drops `node_filter` when `min_level` is `None`** — not reachable through the parser today; latent only.

### Verification caveat

Per-branch verification (fmt + clippy + affected-crate tests) is green for all six. The single combined-diff `cargo test --all` run could **not** complete in this environment: the sandbox enforces a ~38 GB disk quota and the full workspace test build exhausts it (ENOSPC during test-binary linking), independent of the changes. The six changes are disjoint single-file edits in six different crates with no shared public-API/type changes, so the combined behavior is the union of the per-crate suites that did pass.

---

phil-opp:
Reviewed the `dora/logs` node-filter validation. Routing the node segment through `parse::<NodeId>()` applies the same `validate_node_id` used on every other untrusted node-id path, so inputs containing embedded slashes (e.g. `dora/logs/info/a/b`) or spaces are now rejected at parse time instead of producing a filter that silently matches nothing. Valid single-segment ids still parse. No issues found.

---

---

PR #2374, opened 2026-06-26
fix(node): guard against offset overflow in Arrow alignment check

phil-opp:

## Issue

`raw_payload_buffer_offsets_are_arrow_aligned` (`apis/rust/node/src/event_stream/mod.rs`) decides whether a zenoh payload can be wrapped zero-copy by checking each buffer's alignment:

```rust
&& !(base_ptr as usize + buffer.offset).is_multiple_of(alignment)
```

`buffer.offset` is **peer-controlled** — it is bincode-decoded from the zenoh attachment (`ArrowTypeInfo::buffer_offsets`). A hostile or buggy peer sending an offset near `usize::MAX` overflows this addition: a panic in debug builds, a wrapping add (and thus a wrong align/copy decision) in release. This runs on the zenoh receive path **outside** any `catch_unwind`.

Memory safety is ultimately preserved because the later `buffer_into_arrow_array` re-validates with `checked_add`, but that same discipline should apply to this earlier check rather than letting it overflow.

## Fix

Use `checked_add` and treat an overflowing offset as "not aligned", so the caller falls back to the copy path — mirroring the existing `checked_add` guard in `buffer_into_arrow_array`:

```rust
&& (base_ptr as usize)
    .checked_add(buffer.offset)
    .is_none_or(|addr| !addr.is_multiple_of(alignment))
```

## Validation

- New regression test `alignment_check_does_not_overflow_on_hostile_offset` (offset = `usize::MAX`; panics in debug without the fix).
- `cargo test -p dora-node-api --lib` — passes.
- `cargo clippy -p dora-node-api -- -D warnings` — clean.

---

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -1187,7 +1187,14 @@ fn raw_payload_buffer_offsets_are_arrow_aligned(
     let layout = arrow::array::layout(&type_info.data_type);
     for (buffer, spec) in type_info.buffer_offsets.iter().zip(&layout.buffers) {
         if let arrow::array::BufferSpec::FixedWidth { alignment, .. } = *spec
-            && !(base_ptr as usize + buffer.offset).is_multiple_of(alignment)
+            // `buffer.offset` is peer-controlled (bincode-decoded from the zenoh
+            // attachment), so a hostile/buggy value could overflow this add and
+            // panic in debug builds. Treat an overflowing offset as "not aligned"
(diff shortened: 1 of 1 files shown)


phil-opp:
Reviewed the Arrow alignment-overflow guard. `checked_add(...).is_none_or(...)` preserves the original alignment predicate when the address is in range and treats overflow as "not aligned" (falling back to the copy path), which is the correct safe direction for the peer-controlled `buffer.offset`. All paths (loop + child recursion) funnel through this single guarded expression, so no unchecked arithmetic remains. No issues found.

---

---

PR #2376, opened 2026-06-26
fix(coordinator): prune expired entries from per-IP rate limiter

phil-opp:

## Issue

`IpRateLimiter::check` (`binaries/coordinator/src/ws_server.rs`) inserts an entry for every distinct non-loopback IP that connects and **never removes any**:

```rust
let entry = map.entry(ip).or_insert((0, now));
```

The fixed-window reset only overwrites an existing counter; it never evicts. A peer connecting from many — or rotating — source addresses therefore grows the `HashMap<IpAddr, ..>` without bound for the coordinator's entire lifetime: a slow memory leak and a DoS amplification surface (the limiter meant to bound abuse itself grows under abuse).

## Fix

Prune entries whose window has already elapsed at the start of `check`, bounding the map to IPs seen within the last `RATE_WINDOW_SECS`:

```rust
map.retain(|_, (_, started)| now.duration_since(*started) < window);
```

Because pruning leaves only in-window entries (and a freshly inserted entry starts at `now`), the previous `now.duration_since(entry.1) >= window` "reset window" branch became unreachable and is removed, simplifying the counter logic. Fixed-window semantics are unchanged.

## Validation

- New regression test `rate_limiter_prunes_expired_entries` (seeds a stale entry, asserts a later unrelated check evicts it). Existing limiter tests still pass.
- `cargo test -p dora-coordinator --lib rate_limiter` — 5 passed.
- `cargo clippy -p dora-coordinator -- -D warnings` — clean.

---

### DIFF
--- a/binaries/coordinator/src/ws_server.rs
+++ b/binaries/coordinator/src/ws_server.rs
@@ -55,12 +55,18 @@ impl IpRateLimiter {
         let now = Instant::now();
         let window = std::time::Duration::from_secs(RATE_WINDOW_SECS);
 
+        // Drop entries whose window has already elapsed. Without this the map
+        // grows without bound: `check` inserts an entry for every distinct IP
+        // that ever connects and nothing ever removes them, so a peer rotating
+        // source addresses leaks memory for the coordinator's whole lifetime.
+        // Pruning here bounds the map to IPs seen within the last window, and
+        // means every surviving (or freshly inserted) entry is within its
+        // window — so the count below is always compared within the current
+        // fixed window.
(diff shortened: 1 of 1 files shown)


phil-opp:
Reviewed the per-IP rate-limiter pruning change. The `retain` runs under the same lock as the counter increment (no TOCTOU window), and it only evicts entries whose fixed window has fully elapsed — exactly the entries the old code would have reset to `(1, now)` anyway — so there is no rate-limit bypass and no eviction of still-active entries. The unbounded-map growth is genuinely fixed. No issues found.

---

---

PR #2381, opened 2026-06-27
refactor(node-python): remove never-read generation field from recv caches

phil-opp:
> [!NOTE]
> **This PR was generated automatically by Claude (Anthropic's Claude Code agent).** It was opened via an automated integration, so it appears under the account owner's name — the actual author is Claude. Please review accordingly.

## What

Removes the `generation: u64` field from the receiver-side DORADMA caches `RecvGpuSlot` and `RecvCpuSlot` in `apis/python/node/src/lib.rs`, along with its four construction-site assignments.

## Why

The field is **written at every insertion site but never read** (`cargo check` reports `field 'generation' is never read` for both structs).

I checked that this is genuinely dead and not a latent seqlock bug:
- The DORADMA seqlock that protects against torn reads uses **fresh volatile reads** of the shared-memory header generation word — `read_gen` (pre-read) and `read_gen2` (post-read), compared directly. It does not consult the cached struct field.
- On cache hits the code returns the cached `gpu_va` / `gpu_buf` / `base` pointers, which the surrounding comments note are *stable across data overwrites*; the per-call seqlock still validates each read.

So the cached `generation` value plays no role in correctness. The seqlock logic and its explanatory comments are left untouched.

This is not flagged by CI because `dora-node-api-python` is excluded from the `-D warnings` clippy gate (it builds with maturin/PyO3).

## Verification

- `cargo check -p dora-node-api-python` passes; the two `field 'generation' is never read` warnings are gone.
- `cargo fmt -p dora-node-api-python -- --check` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NBPHjbS271WCrQrL3dJ97A)_

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -174,7 +174,6 @@ struct RecvGpuSlot {
     gpu_va: u64,    // device VA from cudaHostGetDevicePointer, 0 if IPC path
     gpu_buf: u64,   // IPC-opened GPU DRAM pointer, 0 if GPU VA path
     host_base: u64, // original host ptr passed to cudaHostRegister
-    generation: u64,
 }
 unsafe impl Send for RecvGpuSlot {}
 unsafe impl Sync for RecvGpuSlot {}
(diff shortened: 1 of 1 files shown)


phil-opp:
Reviewed the removal of the `generation` field from `RecvGpuSlot`/`RecvCpuSlot`. Confirmed the field is never read anywhere — the seqlock torn-read guard uses a fresh volatile read of the shmem header word, not this cached field — all four construction sites are updated, and neither struct derives anything that would depend on it. No issues found.

---

---

PR #2384, opened 2026-06-27
fix(daemon): remove trailing newline from Indent log formatter

### LINKED ISSUE #2382: `Indent` log formatter now appends a trailing newline to every message, adding a blank line after each log line (regression from #2336)

## Summary

PR #2336 (commit e04a43a, "fix(daemon): restore newlines in Indent log formatter for multi-line messages") correctly fixed the multi-line collapse reported in #2312 by switching the per-line `write!` to `writeln!`. However, it terminates **every** line — including the last — with a newline and does **not** trim the trailing one. Because the `Indent` adapter's output is substituted into a `tracing` event message, and the active `fmt` layer adds its *own* per-event newline, this produces a spurious blank line after **every** message routed to the `Tracing` destination — including the common single-line case, which previously rendered cleanly.

This exact risk was explicitly anticipated in the suggested-fix note of the original issue #2312:

> Re-emit a newline after each indented line, e.g. `writeln!(f, "   {line}")?;` (**optionally trimming a single trailing newline if the doubled blank line is undesirable**) ...

The trim was not applied, and a test (`indent_single_line_gets_newline`) was added pinning the trailing newline, so the doubled-blank-line behavior is now locked in.

## Where

```rust
// binaries/daemon/src/log.rs:531-540
struct Indent<'a>(&'a str);

impl std::fmt::Display for Indent<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        for line in self.0.lines() {
            writeln!(f, "   {line}")?;   // <-- terminates the LAST line too
        }
        Ok(())
    }
}
```

`Indent(&message.message)` is used as the message body of `tracing::{info,warn,error,debug}!` at `binaries/daemon/src/log.rs:372, 386, 399, 412, 425` (the `LogDestination::Tracing` fallback path).

## Why this adds a blank line

The daemon's human-readable subscriber is a `tracing_subscriber::fmt::layer().compact()` (see `libraries/extensions/telemetry/tracing/src/lib.rs:132-134`, the `json == false` branch). The compact/full formatter renders the message inline and then terminates the **event** with its own `\n`.

For a single-line message `"hello"`:

- **Before #2336:** `Indent` produced `   hello` (no trailing newline) → formatter appends `\n` → `   hello\n` (one clean line).
- **After #2336:** `Indent` produces `   hello\n` → formatter appends `\n` → `   hello\n\n` (a blank line after the message).

The added unit test confirms the new `Indent` output for a single line is `"   hello\n"`.

For the JSON layers (`lib.rs:107`, `137`, `163`), the trailing `\n` is instead embedded into the `message` string field of each JSON record (e.g. `"message":"   hello\n"`), which is also spurious.

## Impact

Severity: **Low** (cosmetic — no panic, no data loss, no mis-routing). But it regresses the common single-line log case for a multi-line fix: every forwarded node log line emitted through the `Tracing` destination now has a trailing blank line (human path) or an embedded trailing `\n` (JSON path). The multi-line fix from #2336 itself remains correct and should be kept.

## Suggested fix

Separate lines without terminating the last one, so multi-line messages stay readable *and* no trailing newline is emitted. For example:

```rust
impl std::fmt::Display for Indent<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let mut first = true;
        for line in self.0.lines() {
            if !first {
                writeln!(f)?;
            }
            write!(f, "   {line}")?;
            first = false;
        }
        Ok(())
    }
}
```

The existing `indent_single_line_gets_newline` test should then be updated to expect `"   hello"` (no trailing newline), and `indent_preserves_newlines_between_lines` to expect `"   line1\n   line2\n   line3"` (no trailing newline). The `indent_empty_string_produces_empty` case stays `""`.

phil-opp:
## Summary

Fixes #2382.

PR #2336 fixed multi-line log message rendering by switching from `write!` to `writeln!` inside `Indent::fmt`. However `writeln!` also terminates the **last** line, and because the tracing formatter appends its own per-event `\n`, every log message ended with a spurious blank line.

### Root cause

```rust
// binaries/daemon/src/log.rs — before
impl std::fmt::Display for Indent<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        for line in self.0.lines() {
            writeln!(f, "   {line}")?;  // trailing \n on every line, incl. last
        }
        Ok(())
    }
}
```

The tracing compact/full formatter terminates each event with its own `\n`, so a single-line message `"hello"` produced `"   hello\n\n"` — one blank line per log message.

### Fix

Use a `Peekable` iterator to distinguish the last line and call `write!` (no newline) instead of `writeln!` for it:

```rust
impl std::fmt::Display for Indent<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let mut lines = self.0.lines().peekable();
        while let Some(line) = lines.next() {
            if lines.peek().is_some() {
                writeln!(f, "   {line}")?;
            } else {
                write!(f, "   {line}")?;
            }
        }
        Ok(())
    }
}
```

Two unit tests are updated to assert the corrected (no-trailing-newline) output; `indent_single_line_gets_newline` is renamed to `indent_single_line_no_trailing_newline` to reflect the fix.

🤖 Generated with [Claude Code](https://claude.ai/code/session_011M2xjArXh256K9iuc3pPD9)

---
_Generated by [Claude Code](https://claude.ai/code/session_011M2xjArXh256K9iuc3pPD9)_

### DIFF
--- a/binaries/daemon/src/log.rs
+++ b/binaries/daemon/src/log.rs
@@ -532,8 +532,13 @@ struct Indent<'a>(&'a str);
 
 impl std::fmt::Display for Indent<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        for line in self.0.lines() {
-            writeln!(f, "   {line}")?;
+        let mut lines = self.0.lines().peekable();
+        while let Some(line) = lines.next() {
+            if lines.peek().is_some() {
+                writeln!(f, "   {line}")?;
+            } else {
+                write!(f, "   {line}")?;
+            }
         }
         Ok(())
     }
@@ -700,13 +705,15 @@ mod tests {
     #[test]
     fn indent_preserves_newlines_between_lines() {
         let out = Indent("line1\nline2\nline3").to_string();
-        assert_eq!(out, "   line1\n   line2\n   line3\n");
+        // No trailing newline: tracing adds its own per-event newline.
+        assert_eq!(out, "   line1\n   line2\n   line3");
     }
 
     #[test]
-    fn indent_single_line_gets_newline() {
+    fn indent_single_line_no_trailing_newline() {
         let out = Indent("hello").to_string();
-        assert_eq!(out, "   hello\n");
+        // No trailing newline: tracing adds its own per-event newline.
+        assert_eq!(out, "   hello");
     }
 
     #[test]


phil-opp:
Reviewed the `Indent` formatter change. The `Peekable` rewrite correctly drops only the trailing newline (the tracing layer appends its own per-event `\n`), and `str::lines()` semantics mean empty and trailing-newline inputs are handled without a spurious indent line. No issues found.

---

---

PR #2385, opened 2026-06-27
fix(cli): match running dataflow by descriptor name in dora record --proxy

### LINKED ISSUE #2383: `dora record --proxy` records the wrong (first) running dataflow and stores a mismatched descriptor
> 🤖 **This issue was filed automatically by a scheduled Claude code-review check.** It is the result of an automated review of a randomly selected component (`binaries/cli/src/command/record.rs`). Please verify the analysis before acting on it.

## Summary

In proxy recording mode (`run_record_proxy` in `binaries/cli/src/command/record.rs`), the CLI never matches the running dataflow against the descriptor the user passed on the command line. It simply records the **first** active dataflow (or, with multiple, whichever the user picks from a list), while writing the recording header with the descriptor read from `args.file`. When the running dataflow is not the one in `args.file`, the recorded data and the stored descriptor disagree, and the topics subscribed to are derived from the wrong descriptor.

The code's own comment states the intended behavior, which is not implemented:

https://github.com/dora-rs/dora/blob/main/binaries/cli/src/command/record.rs#L301

```rust
// List running dataflows, find the one that matches our descriptor name
// For proxy mode, the user should have already started the dataflow
```

## Details

`run_record_proxy` does the following (`binaries/cli/src/command/record.rs`):

1. Reads and parses the descriptor from `args.file` and derives the topic list from it (lines 247–277).
2. Lists active dataflows from the coordinator and selects one **without comparing against the supplied descriptor** (lines 312–339):

```rust
// Use the first active dataflow (or let user pick if multiple)
let dataflow_id = if active_ids.len() == 1 {
    active_ids[0].uuid
} else {
    let choices: Vec<String> = active_ids.iter().map(|d| d.to_string()).collect();
    let selected = inquire::Select::new("Select dataflow to record:", choices)
        .prompt()
        .wrap_err("dataflow selection cancelled")?;
    active_ids
        .iter()
        .find(|d| d.to_string() == selected)
        .unwrap()
        .uuid
};
```

3. Subscribes to topics derived from `args.file` against that `dataflow_id` (lines 342–352).
4. Writes the recording header with `descriptor_yaml: yaml_bytes` taken from `args.file` (lines 360–365):

```rust
let header = RecordingHeader {
    version: 1,
    start_nanos,
    dataflow_id,
    descriptor_yaml: yaml_bytes,
};
```

## Why this is a bug

- **Single-dataflow case:** If exactly one dataflow is running and it is **not** the one described by `args.file`, the CLI silently records that unrelated dataflow. The header stores `args.file`'s descriptor (`descriptor_yaml`) and its topic set, so the saved descriptor does not correspond to the data actually captured. Topic subscriptions derived from the wrong descriptor may also match nothing (or coincidentally match), producing empty or partial recordings with no warning.
- **Multi-dataflow case:** The user is asked to pick from *all* active dataflows by UUID/name, with nothing tying the choice back to the descriptor they passed, making a wrong selection easy.

The matching the comment describes is implementable: `ControlRequestReply::DataflowList::get_active()` returns `Vec<DataflowIdAndName>`, and `DataflowIdAndName` carries `name: Option<String>` (see `libraries/message/src/coordinator_to_cli.rs:255-258`). The descriptor's name is therefore available on both sides to filter `active_ids` before selecting.

## Suggested fix

Filter `active_ids` by the name of the supplied descriptor before the single-vs-multiple selection:

- If exactly one active dataflow matches the descriptor name → record it.
- If none match → `bail!` with a clear message (instead of recording an unrelated dataflow).
- If multiple match (e.g. same descriptor started several times) → keep the interactive `inquire::Select`, but only among matching dataflows.

This makes the behavior match the stated intent and prevents recording data that does not correspond to the stored descriptor.

## Affected file

- `binaries/cli/src/command/record.rs` — `run_record_proxy`, lines ~301–365

phil-opp:
## Summary

Fixes #2383.

`run_record_proxy` previously picked the **first** active dataflow (or prompted the user to pick from all of them) without ever comparing the selection against the descriptor file the user supplied on the command line. When the running dataflow was unrelated to `args.file`, the CLI silently recorded the wrong one — the recording header stored the descriptor from `args.file` while the actual data came from a different dataflow.

### Root cause

```rust
// binaries/cli/src/command/record.rs — before
// Use the first active dataflow (or let user pick if multiple)
let dataflow_id = if active_ids.len() == 1 {
    active_ids[0].uuid  // no check against the descriptor
} else {
    // user picks from ALL running dataflows, not filtered by descriptor
    ...
};
```

### Fix

Derive an expected name from the descriptor file stem (the same token a user would supply to `dora start --name <name>`). Filter `active_ids` to only the dataflows whose coordinator-assigned name matches that stem:

- **One matching dataflow** → auto-select (no prompt, clearly the right one)
- **Multiple matching** → prompt from the filtered set
- **No match** → emit a warning and fall back to all running dataflows so the user can still pick, rather than silently picking the wrong one

```
warning: no running dataflow named `dataflow` matches the descriptor file.
Showing all running dataflows — select the one started from `examples/foo/dataflow.yml`.
```

This preserves backwards compatibility for un-named dataflows (the warning + fallback path) while making the common case automatic when the user did pass `--name`.

🤖 Generated with [Claude Code](https://claude.ai/code/session_011M2xjArXh256K9iuc3pPD9)

---
_Generated by [Claude Code](https://claude.ai/code/session_011M2xjArXh256K9iuc3pPD9)_

### DIFF
--- a/binaries/cli/src/command/record.rs
+++ b/binaries/cli/src/command/record.rs
@@ -323,43 +323,26 @@ fn run_record_proxy(args: Record) -> eyre::Result<()> {
         );
     }
 
-    // Try to narrow to dataflows whose coordinator-assigned name matches the
-    // descriptor's file stem (e.g. "dataflow" for "examples/foo/dataflow.yml").
-    // Users can set this name via `dora start --name <name>`. If no match is
-    // found fall back to all running dataflows with a warning so the operator
-    // can pick the right one rather than silently recording an unrelated flow.
-    let expected_name = PathBuf::from(&args.file)
-        .file_stem()
-        .and_then(|s| s.to_str())
-        .map(str::to_string);
-    let candidates: Vec<_> = if let Some(ref name) = expected_name {
-        let matched: Vec<_> = active_ids
-            .iter()
-            .filter(|d| d.name.as_deref() == Some(name.as_str()))
-            .cloned()
-            .collect();
-        if matched.is_empty() {
-            eprintln!(
-                "warning: no running dataflow named `{name}` matches the descriptor file. \
-                 Showing all running dataflows — select the one started from `{}`.",
-                args.file
-            );
-            active_ids.clone()
-        } else {
-            matched
-        }
+    // Select which running dataflow to record.  There is no reliable way to
+    // match a descriptor file against a running dataflow from the CLI side
+    // (the coordinator assigns random petnames to un-named dataflows), so we
+    // make the selection *visible* instead of silent so the user can abort
+    // (Ctrl-C) if the wrong dataflow is shown.
+    let dataflow_id = if active_ids.len() == 1 {
+        let selected = &active_ids[0];
+        eprintln!("Auto-selected dataflow: {selected}");
+        eprintln!(
+            "hint: if this is not the dataflow for `{}`, press Ctrl-C and use \
+             `dora list` to find the correct dataflow UUID.",
+            args.file
+        );
+        selected.uuid
     } else {
-        active_ids.clone()
-    };
-
-    let dataflow_id = if candidates.len() == 1 {
-        candidates[0].uuid
-    } else {
-        let choices: Vec<String> = candidates.iter().map(|d| d.to_string()).collect();
+        let choices: Vec<String> = active_ids.iter().map(|d| d.to_string()).collect();
         let selected = inquire::Select::new("Select dataflow to record:", choices)
             .prompt()
             .wrap_err("dataflow selection cancelled")?;
-        candidates
+        active_ids
             .iter()
             .find(|d| d.to_string() == selected)
             .unwrap()
--- a/binaries/cli/src/command/record.rs
+++ b/binaries/cli/src/command/record.rs
@@ -1,7 +1,10 @@
 use std::{collections::BTreeMap, io::Write, path::PathBuf, time::SystemTime};
 
 use clap::Args;
-use dora_message::{common::Timestamped, daemon_to_daemon::InterDaemonEvent, id::NodeId};
+use dora_message::{
+    common::Timestamped, coordinator_to_cli::DataflowIdAndName, daemon_to_daemon::InterDaemonEvent,
+    id::NodeId,
+};
 use dora_recording::{RecordEntry, RecordingHeader, RecordingWriter};
 use eyre::{Context, bail};
 
@@ -51,6 +54,13 @@ pub struct Record {
     /// Useful when the target machine has no local disk.
     #[clap(long)]
     proxy: bool,
+
+    /// Name of the running dataflow to record (matches `dora start --name`).
+    ///
+    /// Only used in `--proxy` mode. When omitted, dora auto-selects if a
+    /// single dataflow is running and prompts otherwise.
+    #[clap(long, value_name = "NAME")]
+    name: Option<String>,
 
     #[clap(flatten)]
     coordinator: crate::common::CoordinatorOptions,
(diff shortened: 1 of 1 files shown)


phil-opp:
I took a look at this change and have one concern about the matching premise.

The fix derives an expected name from the descriptor's file stem and matches it against each running dataflow's coordinator-assigned `name`. But `dora start` only populates that name from an explicit `--name`/`-n` flag and otherwise sends `None` (the value flows from `Start.name` straight through `ControlRequest::Start`, and the coordinator stores it verbatim — there's no file-stem defaulting anywhere on the start/coordinator path).

As a result, for the common case where a user runs `dora start dataflow.yml` without `--name`, `d.name` is `None`, the filter never matches, and every `dora record --proxy` invocation falls into the "no running dataflow named ..." warning + full-list fallback. So the new auto-select effectively only fires when the user happened to pass `--name <file-stem>`, and a user who previously got a clean single-dataflow auto-pick now sees a spurious warning first.

The fallback preserves the old behavior, so this isn't a correctness regression — but the matching logic (and the comment claiming the stem is "the same token a user would supply to `dora start --name`") won't help in the default no-`--name` case. It may be worth matching on something that's actually populated by default, or scoping the warning so it doesn't fire for unnamed dataflows.

---

---

phil-opp:

One correction to the earlier automated note about the matching premise: for a dataflow started without `--name`, the coordinator does **not** leave the name as `None`. In `binaries/coordinator/src/lib.rs:750` it does `let name = name.or_else(|| petname(2, "-"));`, so unnamed dataflows are assigned a random petname (e.g. `happy-tree`), which `List` returns verbatim.

So the file-stem filter in `record.rs` never matches in the default case — not because the name is `None`, but because it's a random petname that can't equal the descriptor's file stem. It also won't match a *named* dataflow unless the user passed `--name <file-stem>` exactly (e.g. `dora start --name foo dataflow.yml` yields name `foo`, while the proxy derives stem `dataflow`).

The fallback keeps this from being a correctness regression, but the auto-select path effectively only fires in the narrow case where `--name` equals the descriptor's file stem. Matching on a value that's actually populated by default (or scoping the "no match" warning to skip the unnamed/petname case) would make the feature useful in the common path.

---

phil-opp:
@claude Address the review feedback

phil-opp:
Thanks for the analysis — both automated notes are correct. The key insight from the second comment is that the coordinator **always** populates `name` (either from `--name` or via `petname(2, "-")`), so the file-stem filter was effectively dead code that triggered a spurious warning on every normal invocation.

I've pushed an updated commit that takes the simpler and more correct approach:

- **Removes** the broken name-matching / petname-aware logic entirely
- **Prints** the auto-selected dataflow to stderr in the single-dataflow case (making the selection visible and abortable via Ctrl-C), rather than silent
- The multi-dataflow case already prompts the user interactively — no change needed there

This directly addresses the root complaint from #2383 (silent wrong pick) without false positives or requiring users to pass `--name`.

---

phil-opp:
Pushed an updated commit implementing the approach from the automated review. Changes in `binaries/cli/src/command/record.rs`:

- **New `--name` flag** on `dora record --proxy`: matches the exact coordinator-assigned name (what the user passed to `dora start --name`). Hard error if no match — prevents silently recording the wrong dataflow.
- **Silent file-stem convenience**: if no `--name` is given but the descriptor's file stem exactly matches one running dataflow's name, auto-select it without any warning.
- **Clean fallback**: without `--name` or a stem match (the common petname case), reverts to original behavior — auto-select if one dataflow is running, prompt otherwise. No spurious warning.
- **`select_dataflow()` helper**: pure function extracted for testability, with 5 unit tests covering all branches (explicit match, explicit no-match error, stem auto-select, petname fallback, multi-running prompt). All tests pass locally (`cargo test -p dora-cli --lib -- record::tests`). Clippy and fmt also pass.

---

phil-opp:

The latest commits (`393e12b`, `1a7bcdb`) address the earlier concern. The file-stem match is now **silent** (returns a selection only on an exact single match, otherwise falls through), so the spurious `no running dataflow named ...` warning on every normal petname-default invocation is gone, and the clean single-dataflow auto-pick / multi-dataflow prompt behavior is preserved. The hard error now only fires when an explicit `--name` is passed and matches nothing, which is the intended guard against silently recording the wrong dataflow.

The extracted `select_dataflow()` is pure and well-covered by the unit tests. The remaining `.unwrap()` in the `Many` prompt branch is safe (the selected string is drawn from the candidate list). No new substantive issues found.

---

PR #2386, opened 2026-06-27
feat(memory-pool): auto-select pinned vs pageable DMA based on tensor size

tang-canran:
register_memory_pool now sets is_pinned based on a 25 MiB threshold:
- >25 MiB: pinned DMA (bandwidth advantage outweighs pin overhead)
- <25 MiB: pageable DMA (pin/unpin fixed cost dominates)

PoolSlot stores is_pinned so write_memory_pool fast path reads it directly from the local cache (no env var, no daemon query).

Inline Python dma_copy accepts no_dma parameter: when true, skips cudaHostRegister/cudaHostUnregister — cudaMemcpy uses pageable source.

Ablation finding: crossover at ~25 MiB where pinned DMA bandwidth overtakes pageable + pin overhead.

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -230,6 +230,12 @@ const OFF_IPC_FLAG: usize = 24;
 const OFF_IPC_HANDLE: usize = 32;
 const OFF_WRITE_GEN: usize = 96;
 
+/// Crossover where pinned-DMA bandwidth overtakes pageable copy +
+/// cudaHostRegister/unregister fixed cost (~100 µs).  Determined by
+/// ablation study (2026-06-27): pageable faster below, pinned faster
+/// above.  Shared by `register_memory_pool` and `write_memory_pool`.
+const DMA_PIN_THRESHOLD_BYTES: usize = 25 * 1024 * 1024;
+
 /// Get (or compile) the persistent CUDA DMA helper module.
 ///
 /// Compiled once at first use and reused across all subsequent iterations.
@@ -364,12 +370,15 @@ def _ipc_close(d_ptr):
 def dma_copy(ptr, size, slot, no_dma):
     """DMA transfer from host to pre-allocated GPU buffer.
 
-    Pins source memory, copies via cudaMemcpyHtoD (DMA engine),
-    then unpins. Returns the device pointer of the pooled GPU buffer.
-
-    When no_dma=True, the pin/unpin calls are skipped so the copy uses
+    Copies via cudaMemcpyHtoD (DMA engine).  When *no_dma* is false
+    (the default), the source memory is pinned (cudaHostRegister)
+    before the copy and unpinned after — this is the fast path for
+    large tensors where pinned-DMA bandwidth outweighs the pin/unpin
+    fixed cost.  When *no_dma* is true, pin/unpin is skipped, using
     pageable memory (faster for small tensors where pin overhead
-    dominates the DMA bandwidth advantage).
+    dominates).
+
+    Returns the device pointer of the pooled GPU buffer.
     """
     if not no_dma:
         _register_host(ptr, size)
@@ -1166,12 +1175,10 @@ impl Node {
         let is_cuda = tensor_device.starts_with("cuda");
         let receiver_is_cuda = device.starts_with("cuda");
         let cpu_mode = !receiver_is_cuda;
-        // Auto-select pinning strategy: for CPU→CUDA transfers, pin only
-        // when tensor > 25 MiB (where pinned DMA bandwidth outweighs the
-        // per-frame cudaHostRegister overhead).  Below 25 MiB, pageable
-        // cudaMemcpy is faster because pin/unpin fixed cost (~100 µs)
-        // dominates the transfer time.
-        let is_pinned = !cpu_mode && size > 25 * 1024 * 1024;
+        // Auto-select pinning: key off the source device — pinning only
+        // matters when the source is CPU (cudaHostRegister on device memory
+        // is a no-op).  Use the 25 MiB crossover from the ablation study.
+        let is_pinned = !is_cuda && size > DMA_PIN_THRESHOLD_BYTES;
         let pinned_type = if cpu_mode { "cpu" } else { "cuda" };
 
         if ptr_val == 0 {
@@ -1444,12 +1451,9 @@ impl Node {
             }
         }
 
-        // Auto-select pinning: for CPU→CUDA transfers, pin only when
-        // tensor > 25 MiB (matching the threshold in register_memory_pool).
-        // The sender-side is_cuda guards this: pinning only matters when the
-        // source is CPU (cudaHostRegister on CUDA memory is a no-op).
-        // Used by both fast-path (cache-miss) and slow-path below.
-        let auto_pin = !is_cuda && size > 25 * 1024 * 1024;
+        // Pin-decision guard shared by cache-miss PoolSlot construction and
+        // slow-path dma_copy (cache-hit reuses the slot's stored is_pinned).
+        let auto_pin = !is_cuda && size > DMA_PIN_THRESHOLD_BYTES;
 
         // Fast path: pool_ format -> DORADMA
         if buffer_id.starts_with("pool_") {
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -236,6 +236,60 @@ const OFF_WRITE_GEN: usize = 96;
 /// above.  Shared by `register_memory_pool` and `write_memory_pool`.
 const DMA_PIN_THRESHOLD_BYTES: usize = 25 * 1024 * 1024;
 
+/// Returns `true` when the source tensor should be pinned before DMA.
+///
+/// Pinning is a property of the *source* pointer: `cudaHostRegister` only
+/// makes sense for host (CPU) memory, and the pin/unpin fixed cost
+/// (~100 µs) is only worth paying when the tensor is large enough that
+/// the DMA bandwidth gain outweighs it.
+///
+/// # Unit-testable
+///
+/// The decision is pure integer logic — no CUDA runtime calls — so the
+/// boundary (25 MiB ± 1 byte) can be exercised in CI even without a GPU.
+#[inline]
+const fn should_pin(is_cuda: bool, size: usize) -> bool {
+    !is_cuda && size > DMA_PIN_THRESHOLD_BYTES
+}
+
+#[cfg(test)]
+mod pin_tests {
+    use super::*;
+
+    #[test]
+    fn pin_cpu_source_above_threshold() {
+        // CPU source, 25 MiB + 1 byte → should pin
+        assert!(should_pin(false, 25 * 1024 * 1024 + 1));
+        // CPU source, 100 MiB → should pin
+        assert!(should_pin(false, 100 * 1024 * 1024));
+    }
+
+    #[test]
+    fn pin_cpu_source_below_threshold() {
+        // CPU source, exactly at threshold → should NOT pin (> not >=)
+        assert!(!should_pin(false, 25 * 1024 * 1024));
+        // CPU source, 1 byte below → should NOT pin
+        assert!(!should_pin(false, 25 * 1024 * 1024 - 1));
+        // CPU source, tiny → should NOT pin
+        assert!(!should_pin(false, 1));
+    }
+
+    #[test]
+    fn pin_cuda_source_never_pins() {
+        // CUDA source regardless of size → never pin
+        assert!(!should_pin(true, 0));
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I reviewed the 25 MiB pinned/pageable DMA auto-selection.

- The new `no_dma=true` (pageable) path is correct: `cudaMemcpy` from unpinned host memory is valid CUDA and differs only in bandwidth, so small-tensor transfers stay correct.
- The receiver read-path selection (`apis/python/node/src/lib.rs`) keys off `ipc_present` (header byte 24) and `pinned_type`, **not** the `is_pinned` parameter, so making `is_pinned` size-dependent doesn't affect consumers. `is_pinned` is metadata-only across the daemon and the memory-pool extension — nothing branches on its value.
- All three `dma_copy` call sites were updated to the new 4-arg signature, so there's no stale 3-arg call.

One minor, non-blocking note: `PoolSlot.pinned_type` is stored but never read anywhere — it looks like a dead field.

---

phil-opp:
Re-reviewed after the latest commit (`5edee47`, "remove dead PoolSlot.pinned_type field").

No issues found in the new commit. It cleanly removes the dead `pinned_type` field flagged in the previous review — the field was written but never read, so the removal is inert and the size-based pinned/pageable selection logic is unaffected. The `is_pinned` daemon metadata remains purely descriptive (nothing branches on it to perform pin/unpin), so flipping it from always-`true` to size-dependent stays safe for daemon/receiver consumers, and all three `dma_copy` call sites use the updated 4-arg signature.

Two pre-existing, perf-only (non-blocking) observations carried over, neither a correctness bug:
- The fast-path **cache-miss** constructs its `PoolSlot` with `is_pinned: false`, so a >25 MiB tensor that should be pinned falls back to a pageable copy on a cache miss.
- The slow path always pins regardless of size, ignoring the 25 MiB threshold (acknowledged in a code comment as a rare fallback).

The latest commit looks safe to merge from a correctness standpoint.

---
*This is a fully automated review by Claude. No human has verified these comments.*

---

phil-opp:
Re-reviewed after the latest commit (`1471487`, "compute auto-pin in cache-miss and slow path").

No issues found in the new commit. It directly addresses the two perf-only observations carried in the previous reviews:

- The fast-path **cache-miss** now builds its `PoolSlot` with `is_pinned: auto_pin` (and threads that `is_pinned` into the `dma_copy(..., !is_pinned)` call) instead of hardcoding `false`, so a >25 MiB tensor is now correctly pinned on a cache miss.
- The **slow path** now passes `slow_no_dma = !auto_pin` instead of always pinning, so it respects the same 25 MiB threshold.

The shared `auto_pin = !is_cuda && size > 25 * 1024 * 1024` is guarded on the source being CPU, which is correct — `cudaHostRegister` on device memory wouldn't apply. As established earlier, `is_pinned`/`pinned_type` remain descriptive daemon metadata; the only behavioral switch is the `no_dma` argument to `dma_copy`, and all three call sites now derive it from the same size-based threshold. The cache-hit branch correctly reuses the slot's stored `is_pinned`.

The latest commit looks safe to merge from a correctness standpoint and resolves the earlier non-blocking notes.

---
🤖 Automated review by Claude — this review is fully automated and may contain mistakes; please verify findings independently.

---

tang-canran:
@heyong4725 gentle ping — phil-opp has reviewed and approved the latest commits. Would appreciate your eyes when you have time. No rush.

apis/python/node/src/lib.rs
```diff
@@ -1155,6 +1166,13 @@ impl Node {
         let is_cuda = tensor_device.starts_with("cuda");
         let receiver_is_cuda = device.starts_with("cuda");
         let cpu_mode = !receiver_is_cuda;
+        // Auto-select pinning strategy: for CPU→CUDA transfers, pin only
+        // when tensor > 25 MiB (where pinned DMA bandwidth outweighs the
+        // per-frame cudaHostRegister overhead).  Below 25 MiB, pageable
+        // cudaMemcpy is faster because pin/unpin fixed cost (~100 µs)
+        // dominates the transfer time.
+        let is_pinned = !cpu_mode && size > 25 * 1024 * 1024;
```
heyong4725: **[correctness] `register_memory_pool` and `write_memory_pool` key the pin decision off different variables.**

Here: `is_pinned = !cpu_mode && size > 25 MiB`, where `!cpu_mode` is `receiver_is_cuda` (the **destination**). But `write_memory_pool` (line 1452) uses `auto_pin = !is_cuda && size > 25 MiB` — the **source** device. The write-path comment has the correct rationale: pinning only matters when the *source* is CPU, since `cudaHostRegister` on device memory is a no-op. By that logic this line should key off the source too.

For the documented CPU→CUDA path they're equal, so the common case is fine. They diverge when the **source is CUDA** (>25 MiB): this stores `is_pinned=true` into the `PoolSlot`, but a later cache-**miss** write recomputes `auto_pin=false` — same pool, opposite pinning depending on cache state. On cache-hit it then calls `cudaHostRegister` on a device pointer, which `_register_host` raises on (error != 712), and `dma_copy`'s `if let Ok(...)` silently swallows it, dropping the GPU IPC path.

```rust
let is_pinned = !is_cuda && size > 25 * 1024 * 1024;
```
The existing `if receiver_is_cuda` block already gates whether `dma_copy` runs, so the receiver guard isn't lost.

apis/python/node/src/lib.rs
```diff
@@ -1429,6 +1444,13 @@ impl Node {
             }
         }
 
+        // Auto-select pinning: for CPU→CUDA transfers, pin only when
+        // tensor > 25 MiB (matching the threshold in register_memory_pool).
+        // The sender-side is_cuda guards this: pinning only matters when the
+        // source is CPU (cudaHostRegister on CUDA memory is a no-op).
+        // Used by both fast-path (cache-miss) and slow-path below.
+        let auto_pin = !is_cuda && size > 25 * 1024 * 1024;
```
heyong4725: **[maintainability] Extract the 25 MiB threshold into a named constant.**

`25 * 1024 * 1024` appears in two predicates (here and line 1174) plus three prose comments. It's a tuning constant from the ablation study — exactly the value that should live in one place so the two sites can't drift apart on the next tuning pass:

```rust
/// Crossover where pinned-DMA bandwidth overtakes pageable + pin/unpin cost.
const DMA_PIN_THRESHOLD_BYTES: usize = 25 * 1024 * 1024;
```

Extracting `should_pin(is_cuda, size)` into a tiny pure helper would also make the boundary (25 MiB ± 1) unit-testable and lock in the register/write consistency from the other comment — the CUDA copy can't run in CI, but the *decision* is pure integer logic and currently has no test.

apis/python/node/src/lib.rs
```diff
@@ -356,13 +361,18 @@ def _ipc_close(d_ptr):
     if err != 0:
         raise RuntimeError(f'cudaIpcCloseMemHandle(0x{d_ptr:x}) failed: {err}')
 
-def dma_copy(ptr, size, slot):
+def dma_copy(ptr, size, slot, no_dma):
     """DMA transfer from host to pre-allocated GPU buffer.
 
     Pins source memory, copies via cudaMemcpyHtoD (DMA engine),
```
heyong4725: **[docs] Docstring now overstates behavior.** The opening line still says "Pins source memory ... then unpins" unconditionally, but pinning is now gated on `no_dma`. The added `no_dma` paragraph is good; just soften the opening sentence to say pinning is conditional so it doesn't contradict the body.

heyong4725 (commented):
Reviewed the net diff. The optimization is correct on the primary CPU→CUDA path and downstream `is_pinned` consumers (daemon + memory-pool extension) only record it as metadata, so flipping it from always-true to variable is safe. Four findings below — one I'd want addressed before merge (the register/write predicate mismatch), three cleanups.

heyong4725:
Thanks for this @tang-canran — the threshold approach is reasonable and the net diff is clean (no conflicts, CI green). One thing I'd like to sort out before merge, plus two optional cleanups.

**Blocking — register/write pin predicate mismatch (`lib.rs:1174`):**
`register_memory_pool` computes `is_pinned = !cpu_mode && size > 25 MiB`, where `!cpu_mode` is the *receiver* (`receiver_is_cuda`). But `write_memory_pool` (line 1452) uses `auto_pin = !is_cuda && size > 25 MiB` — the *source* device. Pinning is a property of the source pointer (`cudaHostRegister` is a no-op on device memory, as your write-path comment notes), so the two sites should agree.

They match on the normal CPU→CUDA path, so this isn't a problem there. The divergence shows up for a CUDA-source pool >25 MiB: register caches `is_pinned=true` in the `PoolSlot`, but a later cache-*miss* write recomputes `auto_pin=false` — same pool, opposite behavior depending on cache state, and the cache-hit case ends up calling `cudaHostRegister` on a device pointer (swallowed by `dma_copy`'s `if let Ok(...)`). One-liner to align them:
```rust
let is_pinned = !is_cuda && size > 25 * 1024 * 1024;
```
The `if receiver_is_cuda` block already gates whether `dma_copy` runs, so the receiver guard isn't lost.

**Optional, non-blocking:**
- Pull the `25 * 1024 * 1024` threshold into a named const (it's duplicated across both predicates + three comments). A small `should_pin(is_cuda, size)` helper would also make the boundary unit-testable — worth it since CI can't exercise the CUDA path.
- `dma_copy` docstring still says it pins unconditionally; pinning is now gated on `no_dma`.

Happy to re-review once the first item's in. Nice work on the ablation.

phil-opp:
**Automated review by Claude** — _this is a fully automated review; no human has vetted these findings, so please verify before relying on them._

Re-reviewed after the new commit `373551f`. No correctness, safety, or leak issues found.

The commit aligns the register-path pin predicate with the write path — both now key off the **source** device (`!is_cuda && size > DMA_PIN_THRESHOLD_BYTES`) — which resolves the cache-state-dependent divergence raised in the earlier review. Verified across all four DMA sites:
- register: `is_pinned = !is_cuda && size > THRESHOLD`
- write cache-miss: `auto_pin = !is_cuda && size > THRESHOLD`
- write cache-hit: reuses `slot.is_pinned`
- slow path: `slow_no_dma = !auto_pin`

`is_pinned`/`pinned_type` remain pass-through daemon/receiver metadata — the receiver branches only on the `ipc_present` header byte and `pinned_type`, so decoupling `is_pinned` from the size threshold is safe for consumers. Pageable transfers still export the GPU IPC handle, so the normal CPU→CUDA path is unaffected.

Minor (non-blocking, wording only): the comments describe `cudaHostRegister` on device memory as a "no-op", but `_register_host` would actually raise on a device pointer — the code is still correct because the `!is_cuda` guard prevents that call from ever happening on CUDA source memory.

---

phil-opp:

Re-checked after the latest commit (`f911250`): it's docs-only (a comment reword), no code changed since the previous review.

The earlier blocking concern — `register_memory_pool` keying `is_pinned` on the *receiver* device while the write path keyed it on the *source* — was resolved in `373551f`: register now uses `!is_cuda && size > DMA_PIN_THRESHOLD_BYTES`, matching the write path, with the threshold extracted to a shared constant. I re-verified that the pin decision only affects bandwidth (never copy correctness), that the fast-path DMA branch is gated on `!is_cuda` so it can't `cudaHostRegister` a device pointer, and that `cudaHostRegister`/`cudaHostUnregister` stay balanced on the error path (no host-registration leak). A stale cached `is_pinned` on a cache-hit with a different tensor size is at most a suboptimal pin choice, not a data bug.

No new findings. One optional, non-blocking suggestion still stands: extracting a pure `should_pin(is_cuda, size)` helper would make the 25 MiB boundary unit-testable (the CUDA copy can't run in CI, but the decision is pure integer logic with no coverage today).

---

phil-opp:
**Automated review by Claude** — this is a fully automated review (no human in the loop); please verify before relying on it.

Re-reviewed the commit pushed after the last review round (`ef6bbdde`, "extract should_pin helper, add unit tests"). It's a pure refactor of the existing pin predicate plus tests — no behavior change:

- `should_pin(is_cuda, size)` is identical to the inline predicate it replaces (`!is_cuda && size > DMA_PIN_THRESHOLD_BYTES`), `const fn`, no overflow concern.
- Both call sites (`register_memory_pool` and `write_memory_pool`) now key off the same source-device predicate, so the register/write mismatch flagged earlier stays resolved.
- The boundary test correctly asserts strict `>` (exactly 25 MiB → not pinned) and the CUDA-source-never-pins / zero-size cases.
- The conditional `cudaHostRegister`/`cudaHostUnregister` skip remains balanced inside the `try/finally`, so no host-registration leak; `cudaMemcpy` HtoD is valid from pageable source either way.

No new issues found. (The previously-noted cache-hit stale-`is_pinned` on a differently-sized reuse is at most a suboptimal pin choice, not a correctness bug — unchanged.)

---

PR #2387, opened 2026-06-27
fix(message): reject invalid ids in user input mappings instead of panicking

phil-opp:
## Issue

`InputMapping::from_str` (`libraries/message/src/config.rs`) builds the `User` variant's source `NodeId` and output `DataId` with `.into()`:

```rust
_ => Self::User(UserInputMapping {
    source: source.to_owned().into(),
    output: output.to_owned().into(),
}),
```

`From<String> for NodeId` / `From<String> for DataId` are documented as **panicking** on invalid identifiers (they call `validate_*_id` and `panic!` on failure). Input mappings are parsed straight from user-authored descriptor YAML — `InputMapping::deserialize` calls `string.parse()` — so a descriptor input like:

```yaml
inputs:
  x: "bad node/output"   # space is not allowed in a NodeId
```

aborts descriptor parsing with a **panic** instead of returning a clean deserialization error.

## Fix

Validate the segments with the fallible `str::parse::<NodeId>()` / `parse::<DataId>()` and map the error into `FromStr`'s existing `String` error type:

```rust
let source = source.parse::<NodeId>().map_err(|e| e.to_string())?;
let output = output.parse::<DataId>().map_err(|e| e.to_string())?;
Self::User(UserInputMapping { source, output })
```

This mirrors how the open PR #2373 hardens the sibling `dora/logs` node-filter branch; that PR fixes line ~298 (the `Logs` branch), this one fixes the distinct `User` branch (line ~323) which previously *panicked* rather than silently mis-parsing.

## Validation

- Added regression test `parse_user_mapping_rejects_invalid_ids`. Verified it **panics** against the pre-fix code and **passes** with the fix.
- `cargo fmt --all -- --check`, `cargo clippy -p dora-message -- -D warnings`, `cargo test -p dora-message` (116 + 5 + 1 doctest, all green).

---

🤖 This is a machine-generated PR authored by Claude (Claude Code). Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01A43PRFZVJLmyW77ZMdYcXd)_

### DIFF
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -320,10 +320,16 @@ impl FromStr for InputMapping {
                 }),
                 None => return Err("dora input has invalid format".into()),
             },
-            _ => Self::User(UserInputMapping {
-                source: source.to_owned().into(),
-                output: output.to_owned().into(),
-            }),
+            _ => {
+                // Validate the source/output segments instead of using the
+                // panicking `From<String>` impls (`.into()`): an input mapping
+                // string comes straight from user-authored descriptor YAML, so an
+                // invalid identifier (e.g. containing a space) must surface as a
+                // clean deserialization error rather than panicking the parser.
+                let source = source.parse::<NodeId>().map_err(|e| e.to_string())?;
+                let output = output.parse::<DataId>().map_err(|e| e.to_string())?;
+                Self::User(UserInputMapping { source, output })
+            }
         };
 
         Ok(mapping)
@@ -621,6 +627,23 @@ mod tests {
         let input: Input = serde_yaml::from_str(yaml).unwrap();
         assert_eq!(input.queue_size, Some(5));
         assert_eq!(input.queue_policy, None);
+    }
+
+    #[test]
+    fn parse_user_mapping_rejects_invalid_ids() {
+        // A source node id with a space is not a valid `NodeId` and must be
+        // rejected with a clean error rather than panicking the parser via the
+        // `From<String>` impl.
+        let result: Result<InputMapping, _> = "bad node/output".parse();
+        assert!(result.is_err(), "invalid source node id must be rejected");
+
+        // An output data id with an invalid character must likewise be rejected.
+        let result: Result<InputMapping, _> = "node_a/bad output".parse();
+        assert!(result.is_err(), "invalid output data id must be rejected");
+
+        // A valid mapping still parses successfully.
+        let mapping: InputMapping = "node_a/output_1".parse().unwrap();
+        assert!(matches!(mapping, InputMapping::User(_)));
     }
 
     #[test]


phil-opp:
## Codebase review summary (machine-generated)

This is the summary for an in-depth review run by **Claude (Claude Code)**. Findings were verified against fresh `origin/main`, cross-checked against currently-open PRs to avoid duplicates, and each opened PR was validated with fmt + clippy + the affected crate's tests. The combined diff additionally passed `cargo check --all` and `cargo check --examples`.

### PRs opened this run

**Correctness**
- #2387 — `fix(message)`: reject invalid ids in user input mappings instead of panicking. `InputMapping::from_str`'s `User` branch built `NodeId`/`DataId` via the panicking `From<String>` (`.into()`), so an invalid identifier in user-authored descriptor YAML (e.g. `bad node/output`) panicked the parser instead of returning a clean deserialization error. (Distinct from the open #2373, which hardens the sibling `dora/logs` branch.)

**Optimization**
- #2388 — `perf(daemon)`: skip the per-message `(NodeId, DataId)` key clone + hash for `input_deadlines`/`broken_inputs` lookups in `send_output_to_local_receivers` when those maps are empty (the common case — both are only populated when input deadlines or circuit breakers are configured). Behavior-preserving.

### Findings examined and rejected

- **replay-node** float division by zero (`delta_nanos / speed`) — already guarded by `if speed > 0.0` upstream.
- **recording** `total_bytes` u64 overflow — would require ~16 EB written; not realistic.
- **coordinator-store** `list_node_params` `param_key.to_string()` — the owned `String` is required by the return type; cold query path, not an allocation that can be avoided.
- **daemon** assorted clone-reduction micro-opts (`OutputId`, `git_sources`, `open_inputs`) — speculative, no clean borrow-friendly rewrite without restructuring map key types; no benchmark to justify.
- **node API / arrow-convert** `NonNull::new(...).unwrap()` and `SystemTime::duration_since(...).unwrap()` — defensive-only; the pointers come from non-null `AVec`/Zenoh payloads, and the clock unwraps are on cold setup paths.
- **cli** `output.rs` `NodeId::from(<log "node" field>)` — reconstructed from system-emitted logs whose node ids were already validated at descriptor parse time; not reachable with an invalid id. `logs.rs:700` is test-only code.
- **cli** `dora topic echo` slice/panic on malformed Arrow JSON — already addressed by the open #2181.

🤖 Generated by Claude (Claude Code). Machine-generated — please review before merging.

---

phil-opp:
No issues found.

The change correctly swaps the panicking `From<String>` conversions (`.into()`) for the fallible `parse::<NodeId>()` / `parse::<DataId>()` in the `User` branch of `InputMapping::from_str`, mapping `InvalidId` into the function's existing `String` error type. Since input mappings are parsed straight from user-authored descriptor YAML, this turns a parser panic on something like `bad node/output` into a clean deserialization error, which is the right behavior.

Verified against the source:
- `From<String>` for both `NodeId` and `DataId` panic on invalid characters; `FromStr`/`parse` return `Result<_, InvalidId>`. The substitution is correct and error propagation via `?` is sound.
- Both `.into()` calls in the `User` branch were the only panic sources in that arm, and both are now fallible.
- The regression test covers an invalid source id, an invalid output id, and a valid mapping.

One observation, not blocking and outside this PR's scope: the `dora/logs/.../{node}` branch still builds the filter `NodeId` via direct tuple construction (`NodeId(node.to_owned())`), which bypasses validation entirely rather than panicking — consistent with the note that #2373 handles that sibling path separately.

---

---

PR #2388, opened 2026-06-27
perf(daemon): skip per-message key clone for empty deadline/circuit-breaker maps

phil-opp:
## Issue

`send_output_to_local_receivers` (`binaries/daemon/src/lib.rs`) runs on the per-message local delivery path. On every successfully delivered message it performs two map lookups:

```rust
if let Some(deadline) = dataflow
    .input_deadlines
    .get_mut(&(receiver_id.clone(), input_id.clone()))
{ ... }

if let Some(timeout) = dataflow
    .broken_inputs
    .remove(&(receiver_id.clone(), input_id.clone()))
{ ... }
```

Both maps are keyed by `(NodeId, DataId)` — a tuple key that cannot be borrowed for lookup — so each probe **clones both `String` ids** to build the key, hashes it, and (in the common case) finds nothing.

`input_deadlines` is only populated when an input declares a `deadline`, and `broken_inputs` is only populated after a circuit breaker trips. For the common dataflow that configures neither, these clones+hashes are pure per-message overhead.

## Fix

Guard each lookup with `is_empty()` so the key clone and hash are skipped when the map is empty:

```rust
if !dataflow.input_deadlines.is_empty()
    && let Some(deadline) = dataflow.input_deadlines.get_mut(&(receiver_id.clone(), input_id.clone()))
{ ... }
```

This is behavior-preserving — a lookup into an empty map always returns `None` — and adds only a cheap length check when the maps *are* in use.

## Validation

- `cargo clippy -p dora-daemon -- -D warnings` (clean)
- `cargo test -p dora-daemon` (85 passed, 0 failed)

---

🤖 This is a machine-generated PR authored by Claude (Claude Code). Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01A43PRFZVJLmyW77ZMdYcXd)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -5136,16 +5136,22 @@ async fn send_output_to_local_receivers(
             }) {
                 Ok(()) => {
                     dataflow.inc_pending(receiver_id);
-                    if let Some(deadline) = dataflow
-                        .input_deadlines
-                        .get_mut(&(receiver_id.clone(), input_id.clone()))
+                    // Looking up these maps requires cloning the `(NodeId, DataId)`
+                    // key (the tuple key type can't borrow). Both maps are empty
+                    // unless input deadlines or circuit breakers are configured, so
+                    // skip the per-message key clone + hash in the common case.
+                    if !dataflow.input_deadlines.is_empty()
+                        && let Some(deadline) = dataflow
+                            .input_deadlines
+                            .get_mut(&(receiver_id.clone(), input_id.clone()))
                     {
                         deadline.last_received = Some(Instant::now());
                     }
                     // Circuit breaker recovery: re-open broken input
-                    if let Some(timeout) = dataflow
-                        .broken_inputs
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found.

The change guards each `input_deadlines.get_mut(...)` / `broken_inputs.remove(...)` lookup with an `is_empty()` check before building the `(NodeId, DataId)` probe key.

Verified against the code:
- Both fields are `HashMap<(NodeId, DataId), _>` (`running_dataflow.rs:192-193`). `HashMap::is_empty()` is O(1).
- A `get_mut`/`remove` into an empty map always returns `None`, so the added guard produces identical control flow — the optimization is behavior-preserving.
- The `&&`-`let` chains short-circuit on the `is_empty()` check first, so the two `String` clones + tuple build + hash are skipped on the empty-map hot path.
- The insert paths that populate the maps (the circuit-breaker recovery block at the same site) are untouched, so the maps still get populated correctly when deadlines/breakers are configured; only the per-message lookups are gated.

No correctness, panic, or behavior concerns found.

---

---

PR #2394, opened 2026-06-28
fix(cli): avoid Instant underflow panic in `dora topic info` hz calc

phil-opp:

## Issue

`TopicStats::calculate_hz` in `binaries/cli/src/command/topic/info.rs` computed the window cutoff as:

```rust
let now = Instant::now();
let cutoff = now - window;
```

`Instant - Duration` panics (`"overflow when subtracting duration from instant"`) when `window` exceeds the monotonic-clock value. This is reachable from user input:

- The `--duration` flag is `#[clap(... value_parser = clap::value_parser!(u64).range(1..))]`, so it accepts any `u64` up to `u64::MAX` seconds.
- The collection thread breaks early on `OutputClosed`/`Disconnected` (a node that emits a few outputs then stops), so the full window need **not** elapse before `calculate_hz` runs.

So `dora topic info <topic> --duration <very-large>` against a topic that closes early reaches `now - Duration::from_secs(very_large)` and crashes the command instead of printing stats. (A sufficiently large value such as near `u64::MAX` overflows the `Instant` arithmetic on any machine; smaller-but-large values reach it on a freshly-booted machine where uptime is below the window.)

## Fix

Guard the subtraction with `Instant::checked_sub`. When the cutoff would predate the monotonic epoch, every recorded timestamp is necessarily within the window, so all are kept:

```rust
let recent: Vec<_> = match now.checked_sub(window) {
    Some(cutoff) => timestamps.iter().filter(|&&t| t >= cutoff).copied().collect(),
    None => timestamps.iter().copied().collect(),
};
```

## Validation

- New regression test `calculate_hz_does_not_panic_on_oversized_window` (records two timestamps, calls `calculate_hz(Duration::from_secs(u64::MAX))`, asserts a finite positive rate instead of a panic). It panics on `main`, passes here.
- `cargo test -p dora-cli --lib` (passes), `cargo clippy -p dora-cli -- -D warnings`, `cargo fmt --all -- --check` all green. Full workspace fmt/clippy/test run on the combined review diff.

---
_Generated by Claude Code. Machine-generated; please review before merging._

---
_Generated by [Claude Code](https://claude.ai/code/session_013E9DFdCzXco92HGKE7UcBr)_

### DIFF
--- a/binaries/cli/src/command/topic/info.rs
+++ b/binaries/cli/src/command/topic/info.rs
@@ -87,12 +87,20 @@ impl TopicStats {
         }
 
         let now = Instant::now();
-        let cutoff = now - window;
-        let recent: Vec<_> = timestamps
-            .iter()
-            .filter(|&&t| t >= cutoff)
-            .copied()
-            .collect();
+        // `now - window` would panic ("overflow when subtracting duration from
+        // instant") when `window` exceeds the monotonic clock value -- reachable
+        // via a large `--duration` (the flag accepts any `u64`) when the topic
+        // closes early, so the full window never elapses before this runs. When
+        // the cutoff predates the monotonic epoch, every recorded timestamp is
+        // within the window, so keep them all.
+        let recent: Vec<_> = match now.checked_sub(window) {
+            Some(cutoff) => timestamps
+                .iter()
+                .filter(|&&t| t >= cutoff)
+                .copied()
+                .collect(),
+            None => timestamps.iter().copied().collect(),
+        };
 
         if recent.len() < 2 {
             return None;
@@ -324,5 +332,31 @@ fn format_arrow_type(data_type: &DataType) -> String {
             )
         }
         _ => format!("{:?}", data_type),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn calculate_hz_does_not_panic_on_oversized_window() {
+        // Regression: a large `--duration` (the flag accepts any u64) made
+        // `now - window` underflow the monotonic clock and panic with
+        // "overflow when subtracting duration from instant". This is reachable
+        // when the topic closes early so the full window never elapses.
+        let stats = TopicStats::default();
+        let now = Instant::now();
+        {
+            let mut ts = stats.timestamps.lock().unwrap();
+            ts.push(now);
+            ts.push(now + Duration::from_millis(100));
+        }
+        // A window far larger than any plausible monotonic-clock value.
+        let hz = stats.calculate_hz(Duration::from_secs(u64::MAX));
+        // With the cutoff before the monotonic epoch, every timestamp counts,
+        // so a finite positive rate is returned instead of a panic.
+        let hz = hz.expect("expected a frequency from two timestamps");
+        assert!(hz.is_finite() && hz > 0.0, "unexpected hz: {hz}");
     }
 }


phil-opp:

This is the summary for an automated repository review run (machine-generated by Claude). It scanned for correctness, optimization, documentation, and simplification issues against current `origin/main` (`89c2c8d`). Each finding was verified by hand; weak/unreachable candidates were rejected rather than turned into PRs.

### PRs opened

**Correctness**
- **#2394** — `fix(cli)`: `dora topic info` panicked on `Instant - Duration` underflow. A large `--duration` (the flag accepts any `u64`) combined with an early `OutputClosed`/`Disconnected` break reaches `now - window` with an oversized window and crashes the command. Fixed with `Instant::checked_sub` + regression test.

**Documentation**
- **#2395** — `docs(coordinator)`: the doc block on `MAX_PENDING_REPLIES` mis-attached `send_and_receive`'s doc to the const **and** falsely claimed "each clone keeps an independent pending-reply map" — the map is an `Arc<Mutex<…>>` shared by all clones, so the cap is global per connection. Moved/corrected the docs.
- **#2396** — `docs(core)`: `source_is_url` (a `pub` security-relevant gate used across daemon/runtime/cli) had no docs. Added a compile-checked doctest pinning the accepted schemes and the case-sensitivity of the prefix match.

### Rejected findings (verified not actionable)

| Candidate | Why rejected |
|---|---|
| `arrow_utils::required_data_size` accumulation "overflow" | Send path; buffer lengths come from already-allocated in-memory `ArrayData`, so their sum cannot exceed `usize::MAX`. Not reachable. |
| `recording::write_entry` `event_bytes.len() as u32` truncation / size-add overflow | `record_len_usize` (incl. `event_bytes.len()`) is checked against `MAX_RECORD_BYTES` (64 MB) **before** any `u32` cast; `Vec` len ≤ `isize::MAX`. Provably bounded. |
| `arrow-convert::check_single_datetime` "inverted logic" | Returns `true` when invalid; callers do `if check_single_datetime(..) { bail }`. Behavior is correct (only the name reads oddly). |
| daemon `send_with_timestamp` / `send_output_to_local_receivers` capacity-vs-`try_send` "race" | Deliberate best-effort headroom heuristic with a critical-log fallback on `Full`. By design. |
| `trace/list.rs` `trace_id[..12]` UTF-8 panic | `trace_id` is always a `uuid::Uuid::new_v4()` (ASCII, 36 chars); the non-ASCII path is unreachable. |
| `topic/echo.rs` `buf[4..buf.len()-2]` slice panic | Already addressed by open PR #2181. |
| `payload_buffer_alignment` div-by-zero | Alignment is `.max(ARROW_BUFFER_ALIGNMENT)`, never 0. |
| telemetry span-field silent truncation; ros2 msg-gen `file_stem().unwrap()` | Build-time/by-design; not runtime-reachable defects. |
| hub-client / log-utils / coordinator-store sweep | Clean — inputs size-checked, paths confined, no unwrap on untrusted data. |

### Validation

- Per branch: `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, and tests for the affected crate — all green.
- Combined diff (all 3 PRs): full workspace `cargo test` suite green (exit 0); full `cargo clippy --all -- -D warnings`; `/review` and `/simplify` gates run — no issues found.
- The three changes touch disjoint crates (`dora-cli`, `dora-coordinator`, `dora-core`) with no overlap.

_Generated by Claude Code. Machine-generated; please review before merging._

---

phil-opp:
No issues found.

Guarding the `Instant - Duration` subtraction with `now.checked_sub(window)` correctly prevents the panic reachable via a large `--duration` when a topic closes early. The `None` fallback (keep all timestamps) is mathematically equivalent to the filter when the cutoff would predate the monotonic epoch, so the hz result is unchanged in that path. It is the only `Instant - Duration` in `calculate_hz` — nearby time arithmetic already uses `saturating_sub` / `duration_since` on monotonic values, and the bandwidth path uses `duration_secs`, so no other underflow remains.

---

---

PR #2295, opened 2026-06-22
refactor(message): remove unused WS_PING_INTERVAL constant

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and opened via a maintainer's account because the bot cannot author PRs directly. It has **not** been hand-reviewed by a human yet — please treat it as a suggestion and review before merging.

## What

Removes the unused `WS_PING_INTERVAL` constant from `dora-message`:

```rust
/// WebSocket ping interval for keepalive (10 seconds).
pub const WS_PING_INTERVAL: std::time::Duration = std::time::Duration::from_secs(10);
```

## Why

`WS_PING_INTERVAL` has **zero references anywhere in the workspace** — a repo-wide search for the identifier returns only the definition site.

The coordinator's WebSocket handlers (`ws_control.rs`, `ws_daemon.rs`) only *respond* to inbound `Message::Ping(..)` frames with a pong; nothing sends interval-based keepalive pings, so this constant is never read. Its siblings in the same module — `MAX_MESSAGE_BYTES` and `TCP_READ_TIMEOUT` — are both actively used; `WS_PING_INTERVAL` is the odd one out.

This is a plain magic-number constant, not an extension point: anyone who needs a 10s interval can write `Duration::from_secs(10)` directly, so removing it is not a capability regression.

## Verification

- `cargo clippy -p dora-message -- -D warnings` ✅
- `cargo test -p dora-message` ✅
- `cargo check --all` (excluding the Python packages) ✅ — nothing else references it
- `cargo fmt --all -- --check` ✅

## Note

`dora-message` is a published library, so this is technically a public-API change. If interval-based WS keepalive is planned and this constant was reserved for it, feel free to close — but as it stands it is unreferenced.

---
_Generated by [Claude Code](https://claude.ai/code/session_018wPn3gy8cweX8VPbgorn55)_

PR #2397, opened 2026-06-29
refactor(operator-c): remove unused HEADER_OPERATOR_API/HEADER_OPERATOR_TYPES consts

phil-opp:

## What

Removes two unused `pub const`s from `dora-operator-api-c` (`apis/c/operator/src/lib.rs`):

- `HEADER_OPERATOR_API` — `include_str!("../operator_api.h")`
- `HEADER_OPERATOR_TYPES` — `include_str!("../operator_types.h")`

## Why

Both embed the crate's C header files as Rust string constants, but **nothing in the workspace reads either one** (`grep` for both identifiers across the repo returns only their definitions).

`dora-operator-api-c` is consumed only as a **staticlib** (for linking C operators) and via its **on-disk header files** — `build.rs` copies `operator_api.h` / `operator_types.h`, and the CLI's C/C++ templates `#include` them by path. No Rust code imports `dora_operator_api_c::HEADER_*`.

Contrast with the node-side analog `dora_node_api_c::HEADER_NODE_API`, which **is** wired into the CLI's C-node template generator (`binaries/cli/src/template/c/mod.rs:1,118`). The operator equivalents were never wired into any generator, so they are leftover convenience constants with zero consumers.

This follows the recent cleanup of unused public helpers/constants (e.g. #2295, #2360, #2361).

## Verification

- `cargo clippy -p dora-operator-api-c -- -D warnings` — clean
- `cargo test -p dora-operator-api-c` — passes

Pure deletion; the `.h` files and `build.rs` behavior are untouched, so the C/C++ operator build path is unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TPhxJigtSTpZiDEr9VmahR)_

### DIFF
--- a/apis/c/operator/src/lib.rs
+++ b/apis/c/operator/src/lib.rs
@@ -1,4 +1,1 @@
-pub const HEADER_OPERATOR_API: &str = include_str!("../operator_api.h");
-pub const HEADER_OPERATOR_TYPES: &str = include_str!("../operator_types.h");
-
 pub use dora_operator_api_types;


phil-opp:
No issues found.

Verified via repo-wide grep that `HEADER_OPERATOR_API` and `HEADER_OPERATOR_TYPES` have zero consumers — the only matches were the two removed definitions. The on-disk `operator_api.h` / `operator_types.h` files and `build.rs` are untouched, and the node-side analog `HEADER_NODE_API` (still used by the CLI template generator) confirms the operator-side consts were genuinely dead code.

---

---

phil-opp:
I think the motivation behind these constants was that users of the crate can write the header files via a cargo build script. Not sure about that though. Or do you see any other potential use cases for these constants?

phil-opp:
You're right, and that use case holds up — I'll close this.

These are the operator-side analog of `dora_node_api_c::HEADER_NODE_API`, which the dora CLI itself uses exactly the way you describe: it writes the embedded header to disk in the C-node template generator (`binaries/cli/src/template/c/mod.rs:118` → `fs::write(header_path, HEADER_NODE_API)`). A downstream crate building a C operator can do the same from its own `build.rs` — e.g. `fs::write("operator_api.h", dora_operator_api_c::HEADER_OPERATOR_API)` — to obtain the headers without vendoring the files.

So these are a public-API convenience for downstream consumers, not dead code: "unused in-repo" doesn't make them safe to remove here. Closing. Thanks for the catch.

---

PR #2331, opened 2026-06-25
refactor(daemon): remove unused socket_stream_receive production helper

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep, not by a human contributor, and is opened under a maintainer's account only because the bot cannot author PRs directly. The commit itself is authored as `Claude <noreply@anthropic.com>`. Please review before merging and close if the analysis is wrong.

## What

Removes the unused `socket_stream_receive` function from `binaries/daemon/src/socket_stream_utils.rs`:

```rust
#[allow(dead_code)]
pub async fn socket_stream_receive(
    connection: &mut (impl AsyncRead + Unpin),
) -> std::io::Result<Vec<u8>> {
    socket_stream_receive_with_header_timeout(connection, Some(dora_message::TCP_READ_TIMEOUT)).await
}
```

## Why it's dead

- **No production caller.** A whole-workspace search (`rg 'socket_stream_receive\b'`) finds the function only at its definition site and in the framing unit tests in the same file — the four tests that exercise the wrapper itself. Nothing in the daemon, node, or any other crate calls it.
- **The live paths use the timeout-aware variant directly.** Every request/response call site constructs the receiver via `socket_stream_receive_with_header_timeout(.., Some(..))` (or `None` for idle-tolerant connections). The thin default-timeout wrapper was never wired into them.
- It is already annotated `#[allow(dead_code)]`. `dora-daemon` is a binary crate, so a `pub` item here is **not** a public API surface — an unreferenced `pub fn` is genuinely dead, not an external extension point.

## Scope note (coverage preserved)

The four framing tests (`framing_roundtrip`, `empty_payload_roundtrips`, `receive_rejects_oversized_length_prefix`, `receive_times_out_on_stalled_peer`) verify real round-trip / DoS-guard / stalled-peer behavior, so rather than delete them, the default-timeout wrapper is moved into the test module as a private helper. `socket_stream_receive(conn)` is exactly `socket_stream_receive_with_header_timeout(conn, Some(TCP_READ_TIMEOUT))`, so the tests keep identical coverage while the production surface drops the dead `pub` function.

## Verification

- `cargo fmt -p dora-daemon -- --check` — clean
- `cargo clippy -p dora-daemon -- -D warnings` — clean
- `cargo test -p dora-daemon --lib socket_stream` — 7 tests pass

Pure deletion of an unreferenced helper; no behavior change.

---
[...]

PR #2291, opened 2026-06-21
refactor(core): remove unused infer_types_from_file helper

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and opened via a maintainer's account because the bot cannot author PRs directly. It has **not** been hand-reviewed by a human yet — please treat it as a suggestion and review before merging.

## What

Removes the unused `infer_types_from_file` helper from the `type-inference`-gated `inference` module (and its now-unused `std::path::Path` import):

```rust
pub fn infer_types_from_file(source_path: &Path, node_id: &str) -> Result<Vec<TypeSuggestion>, String> {
    let content = std::fs::read_to_string(source_path)...?;
    infer_types_from_source(&content, source_path.to_string_lossy().as_ref(), node_id)
}
```

## Why

`infer_types_from_file` was a thin wrapper that read a file and delegated to `infer_types_from_source`. It has **zero call sites anywhere in the workspace** — not in production, and not even in the module's own unit tests, which all drive `infer_types_from_source` directly with in-memory source strings. The `type-inference` feature is not enabled by any crate, so nothing reaches this wrapper.

The actual entry point, `infer_types_from_source`, is **kept** — so any caller that needs file-based inference can still read the file and call it. This is not an irreplaceable extension point; it's a convenience wrapper with no users.

## Verification

- `cargo clippy -p dora-core --features type-inference -- -D warnings` ✅
- `cargo test -p dora-core --features type-inference inference` ✅ (14 passed)
- `cargo check -p dora-core` (feature off) ✅
- `cargo fmt -p dora-core -- --check` ✅

## Note

`dora-core` is a published library. If `infer_types_from_file` was intended as the future public entry point for CLI/build wiring, feel free to close this — but as it stands it is unreferenced including by tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Feut1TvNKPGU439BfP4ymS)_

PR #2398, opened 2026-06-29
refactor(core): remove unused `run` process-spawn helper

phil-opp:

## What

Removes the unused generic helper `dora_core::run` from `libraries/core/src/lib.rs`:

```rust
pub async fn run<S>(program: S, args: &[&str], pwd: Option<&Path>) -> eyre::Result<()>
where S: AsRef<OsStr> { ... }
```

and the `std::ffi::OsStr` import that only this function needed.

## Why

The helper spawned a process via `tokio::process::Command` and bailed on a non-zero exit, but **nothing in the workspace calls it**: a grep for `dora_core::run` / `core::run` returns no hits, and every component that spawns subprocesses (daemon spawner, CLI, runtime) builds its own `Command` directly.

This follows the recent cleanup of unused public helpers (e.g. #2187 `get_pip_path`, #2291 `infer_types_from_file`, #2331 `socket_stream_receive`).

## Verification

- `cargo clippy -p dora-core --all-features -- -D warnings` — clean
- `cargo test -p dora-core --all-features` — 269 tests + doctests pass

---
_Generated by [Claude Code](https://claude.ai/code/session_01TPhxJigtSTpZiDEr9VmahR)_

### DIFF
--- a/libraries/core/src/lib.rs
+++ b/libraries/core/src/lib.rs
@@ -1,7 +1,6 @@
 use eyre::{Context, bail, eyre};
 use std::{
     env::consts::{DLL_PREFIX, DLL_SUFFIX},
-    ffi::OsStr,
     path::Path,
 };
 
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found.

Verified that `dora_core::run` has zero references across the workspace (code, examples, tests, and cfg-gated paths); the only re-export (`pub use dora_core::{self, uhlc}`) does not surface it, and every other `run` match is an unrelated symbol. The removed `std::ffi::OsStr` import was used only inside the deleted function, while `Path` (still used by `adjust_shared_library_path`) is correctly retained. Safe removal.

---

---

PR #2402, opened 2026-06-29
fix(core): reject duplicate node IDs in plain dataflows

### LINKED ISSUE #2393: core/descriptor: duplicate node IDs in a plain (non-module) dataflow are silently collapsed instead of rejected

## Summary

A dataflow YAML with two `nodes:` entries that share the same `id` is accepted, and one of the duplicates is **silently dropped**. The duplicate-node-ID guard exists *only* in the module-expansion path; a plain dataflow (no `module:` nodes) never reaches it, and the only normalization it goes through (`resolve_aliases_and_set_defaults`) collapses duplicates via a `BTreeMap::insert`.

## Locations

Silent overwrite — `libraries/core/src/descriptor/mod.rs:72` and `:191-206`:

```rust
let mut resolved = BTreeMap::new();      // BTreeMap<NodeId, ResolvedNode>
for mut node in self.nodes.clone() {
    ...
    resolved.insert(node.id.clone(), ResolvedNode { id: node.id, ... });  // duplicate id -> earlier node discarded
}
```

The only duplicate check is in the module-expansion path, which **returns early before the check** when there are no module nodes — `libraries/core/src/descriptor/expand.rs:102-108` and `:134-144`:

```rust
let has_modules = descriptor.nodes.iter().any(|n| n.module.is_some());
if !has_modules {
    return Ok(ExpandedDescriptor { descriptor: descriptor.clone(), boundaries: Default::default() });  // <-- skips the check below
}
...
// Verify expanded node IDs are unique  (only runs when modules are present)
let mut id_set = HashSet::new();
for node in &flat_nodes {
    if !id_set.insert(node.id.to_string()) {
        bail!("duplicate node ID `{}` after module expansion ...");
    }
}
```

The descriptor field is `nodes: Vec<Node>` (`libraries/message/src/descriptor.rs`), so duplicate ids parse fine. A repo-wide search for a duplicate-node-ID check finds it only in `expand.rs`.

## Problem

For a plain dataflow, one of the duplicately-named nodes silently disappears from the running dataflow:

- If the surviving node happens to declare the outputs downstream nodes subscribe to, validation passes and the dataflow runs with a node (and its process/path) missing — a silent wrong result.
- Otherwise the user gets a misleading `output \`a/x\` does not exist` error that points at the wrong cause.

The module path explicitly guards against exactly this; the plain path does not, even though the duplicate-ID condition is just as invalid there.

## Trigger

Any plain dataflow YAML (no `module:` nodes) with two `nodes:` entries having the same `id`:

```yaml
nodes:
  - id: a
    path: a1.py
    outputs: [x]
  - id: a            # duplicate id — silently dropped
    path: a2.py
    outputs: [x]
  - id: b
    path: b.py
    inputs:
      x: a/x
```

Only `a2.py` runs; the first `a` (path `a1.py`) is silently discarded.

## Suggested fix

Add a duplicate-node-ID check to the plain validation path (e.g. in `check_dataflow` / before/within `resolve_aliases_and_set_defaults`), mirroring the message already used in `expand.rs`, so duplicates `bail!` instead of being collapsed by the `BTreeMap`.

## Confidence

Medium — the silent overwrite and the absence of any duplicate check on the plain path are certain from the code; rated medium (not high) only because the user-visible outcome ranges from "silent wrong result" to "misleading error" rather than a hard panic.

phil-opp:
Fixes #2393

## Summary

A plain dataflow YAML (no `module:` nodes) with two `nodes:` entries sharing the same `id` was silently accepted: the second entry would overwrite the first in the `BTreeMap` inside `resolve_aliases_and_set_defaults`, causing one node to disappear without any error.

The duplicate-check in `expand.rs` only runs when the dataflow contains module nodes, and returns early otherwise, so plain dataflows were unprotected.

## Changes

**`libraries/core/src/descriptor/mod.rs`**
- Added a `contains_key` guard before `resolved.insert(...)` in `resolve_aliases_and_set_defaults`
- On duplicate, returns `eyre::bail!("duplicate node ID \`{}\` — each node must have a unique \`id\`", node.id)`, consistent with the message style in `expand.rs`

**Test added**: `duplicate_node_id_is_rejected` — parses a YAML with two nodes both named `my-node` and asserts the error is returned (233 tests pass including the new one).

---

### DIFF
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -188,6 +188,12 @@ impl DescriptorExt for Descriptor {
                 }
             };
 
+            if resolved.contains_key(&node.id) {
+                eyre::bail!(
+                    "duplicate node ID `{}` — each node must have a unique `id`",
+                    node.id
+                );
+            }
             resolved.insert(
                 node.id.clone(),
                 ResolvedNode {
@@ -733,5 +739,27 @@ nodes:
         );
         let b = resolved.get(&NodeId::from("b".to_string())).unwrap();
         assert!(b.env.is_none(), "node b has no env anywhere");
+    }
+
+    #[test]
+    fn duplicate_node_id_is_rejected() {
+        // Regression for #2393: a plain dataflow with two nodes sharing the
+        // same `id` must return an error instead of silently discarding one.
+        let yaml = r#"
+nodes:
+  - id: my-node
+    path: ./a
+  - id: my-node
+    path: ./b
+"#;
+        let desc: Descriptor = serde_yaml::from_str(yaml).expect("parse");
+        let err = desc
+            .resolve_aliases_and_set_defaults()
+            .expect_err("duplicate node ID must be rejected");
+        let msg = format!("{err:#}");
+        assert!(
+            msg.contains("duplicate node ID") && msg.contains("my-node"),
+            "unexpected error message: {msg}"
+        );
     }
 }


phil-opp:
No issues found.

The `contains_key` guard in `Descriptor::resolve_aliases_and_set_defaults` correctly bails on a duplicate node `id` in plain dataflows instead of silently overwriting via `BTreeMap::insert`. The guard reads `&node.id` before the id is moved into the map; the module-expansion path already has its own duplicate-id check producing a flat, unique-id descriptor (so this does not false-positive on expanded descriptors); and node ids in a single `nodes:` list are not namespaced, so rejecting the collision is the correct behavior rather than dropping a node silently.

---

---

PR #2403, opened 2026-06-29
fix(ros2-bridge): parse byte preset default as u8 instead of ASCII bytes

### LINKED ISSUE #2400: ros2-bridge: `byte` field preset defaults are decoded as ASCII bytes instead of the numeric value

## Summary

In the ROS2-bridge Arrow serialization layer, the default-value generator decodes a `byte`-typed field's preset default by taking the **raw UTF-8 bytes of the literal string** instead of **parsing it as a `u8`**. Every other integer type (including `char`, which is also a `UInt8` on the wire) parses the literal correctly. As a result, a `.msg` field such as `byte status 5` produces the wrong value, and many `byte` defaults make the whole message fail to serialize.

## Location

`libraries/extensions/ros2-bridge/arrow/src/serialize/defaults.rs:203`

```rust
fn preset_default_for_basic_type(t: &NestableType, preset: &str) -> Result<ArrayData> {
    Ok(match t {
        NestableType::BasicType(t) => match t {
            // ...
            BasicType::U8 => UInt8Array::from(vec[preset.parse::<u8>()
                .context("Could not parse preset default value")?]).into(),
            // ...
            // `char` is a `UInt8` on the wire (see `primitive.rs`); its preset
            // default is the numeric byte value, not a string.
            BasicType::Char => UInt8Array::from(vec[preset.parse::<u8>()
                .context("Could not parse preset default value")?]).into(),
            BasicType::Byte => UInt8Array::from(preset.as_bytes().to_owned()).into(), // <-- BUG
            // ...
        },
        // ...
    })
}
```

`BasicType::Byte` uses `preset.as_bytes().to_owned()`, while `U8`/`Char` use `preset.parse::<u8>()`.

## How the value reaches this function

When a `.msg` line like `byte status 5` is parsed (`msg-gen/src/parser/member.rs` → `literal::get_basic_type_literal_parser(BasicType::Byte)` = `validate_integer_literal::<u8>`), the literal is validated and **normalized to a decimal `u8` string**, so `Member::default` becomes `Some(vec["5"])`. A hex default `byte status 0x42` is normalized to `Some(vec["66"])`.

`default_for_member` calls `preset_default_for_basic_type(Byte, "5")` whenever the field is omitted from an incoming Arrow message, so its default has to be materialized.

## Why it's wrong

`"5".as_bytes()` is `[0x35]` (the ASCII code of the character `'5'`), not `[5]`. So the generated default is the byte `53`, not `5`.

This manifests two ways, both confirmed against the serialize path (`serialize/primitive.rs::as_single_primitive`, which requires a length-1 array for a scalar field):

1. **Single-digit defaults (0–9): silent value corruption.** `byte b 5` → default `[53]` (length 1) → serializes as `53` instead of `5`.
2. **Multi-digit defaults (10–255): hard serialize failure.** `byte b 66` (or `byte b 0x42`, normalized to `"66"`) → `"66".as_bytes()` = `[54, 54]` (length 2) → `as_single_primitive` rejects it with `expected single-element u8 array, got length 2`, failing the whole message even though the `.msg` default was valid.

The same root cause also affects `byte[]` / `sequence<byte>` defaults: `list_default_values` calls `preset_default_for_basic_type(value_type, default)` per element, so each element string is turned into its ASCII bytes — corrupting values and (for any element ≥ 10) producing the wrong element count.

## Impact

- ROS2 messages with `byte` fields that carry preset defaults are silently corrupted or fail to serialize whenever the field is omitted by the producing dora node.
- Severity is **moderate**: `byte` fields are less common than the integer types, and the bug only triggers on the omitted-field default path — but when it does trigger it produces wrong wire data with no warning (single-digit case) or a confusing length error (multi-digit case).

## Suggested fix

Decode `byte` exactly like `u8`/`char` (the literal parser has already validated it as a `u8`):

```rust
BasicType::Byte => UInt8Array::from(vec[
    preset
        .parse::<u8>()
        .context("Could not parse preset default value")?,
])
.into(),
```

A regression test mirroring the existing `preset_char_default_is_uint8` / `preset_char_array_default_serializes` tests (asserting `byte b 5` → value `5`, and that a `byte b 66` default round-trips to a single byte `66`) would prevent this from recurring.

## Notes

- Found by reviewing `libraries/extensions/ros2-bridge/arrow/src/serialize/defaults.rs`. The surrounding default-handling code (fixed-array length wrapping, `char`-as-`UInt8` handling) is otherwise well covered by tests from #2027; the `byte` preset path is the one untested branch.
- This is distinct from #2363 (bounded-string default length counting in `msg-gen/.../literal.rs`).

phil-opp:
Fixes #2400

## Summary

In `libraries/extensions/ros2-bridge/arrow/src/serialize/defaults.rs`, the `BasicType::Byte` arm in `preset_default_for_basic_type` used `preset.as_bytes().to_owned()` instead of parsing the string as a `u8`.

This caused two classes of bugs:
- **Single-digit defaults (0–9):** silent value corruption — `byte b 5` produced `[0x35]` (decimal 53) instead of `[5]`
- **Multi-digit defaults (10–255):** hard serialization failure — `byte b 66` produced a 2-element array `[0x36, 0x36]` which fails the scalar-field length check

The `U8` and `Char` arms already use `preset.parse::<u8>()` correctly; this fixes `Byte` to match.

## Change

`libraries/extensions/ros2-bridge/arrow/src/serialize/defaults.rs` — change `BasicType::Byte` arm from `UInt8Array::from(preset.as_bytes().to_owned()).into()` to the same `parse::<u8>()` pattern used by `Char`/`U8`.

All 10 existing `dora-ros2-bridge-arrow` tests pass.

---

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/serialize/defaults.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/serialize/defaults.rs
@@ -200,7 +200,12 @@ fn preset_default_for_basic_type(t: &NestableType, preset: &str) -> Result<Array
                     .context("Could not parse preset default value")?,
             ])
             .into(),
-            BasicType::Byte => UInt8Array::from(preset.as_bytes().to_owned()).into(),
+            BasicType::Byte => UInt8Array::from(vec![
+                preset
+                    .parse::<u8>()
+                    .context("Could not parse preset default value")?,
+            ])
+            .into(),
             BasicType::Bool => BooleanArray::from(vec![
                 preset
                     .parse::<bool>()


phil-opp:
No issues found.

Parsing a `byte` preset default via `preset.parse::<u8>()` is correct: ROS 2 `byte`/`octet` defaults are numeric literals, whereas the old `UInt8Array::from(preset.as_bytes().to_owned())` treated the default string as raw ASCII bytes — producing wrong values and multi-element arrays for multi-digit defaults (e.g. `"66"` became `[0x36, 0x36]` instead of `[66]`). The fix is consistent with the sibling `U8`/`Char` arms.

---

---

PR #2405, opened 2026-06-29
fix(coordinator): continue destroy_daemons even when a dataflow stop fails

### LINKED ISSUE #2391: coordinator: handle_destroy bails before destroy_daemons when any stop_dataflow errors, orphaning live daemons

## Summary

In `handle_destroy`, each running dataflow is stopped with `stop_dataflow(...).await?`. The `?` propagates the first error and returns early, so `destroy_daemons(...)` — which is what actually tells the daemons to shut down and drains `daemon_connections` — is **never reached** if any single `stop_dataflow` call fails. The result is a partial/failed teardown: live daemons are left running and the `Destroy` request returns an error to the CLI.

## Location

`binaries/coordinator/src/handlers.rs:176-188`

```rust
for dataflow_uuid in running_dataflows.keys().cloned().collect::<Vec<_>>() {
    let _ = stop_dataflow(
        running_dataflows,
        dataflow_uuid,
        daemon_connections,
        clock.new_timestamp(),
        None,
        false,
    )
    .await?;            // <-- bails out of handle_destroy on the first error
}

destroy_daemons(daemon_connections, clock.new_timestamp()).await?;   // never runs on error
```

Note that `let _ = ...?` only discards the `Ok` value; the `?` still propagates `Err`.

## Problem

`stop_dataflow` returns `Err` in realistic teardown conditions, e.g.:

- a dataflow's `daemons` set references a daemon that is no longer in `daemon_connections`:
  `binaries/coordinator/src/handlers.rs:235-237` → `daemon_connections.get_mut(daemon_id).wrap_err("no daemon connection")?`;
- a daemon returns a non-`Ok` `StopResult` (`handlers.rs:246-248`).

When that happens during `dora destroy` / Ctrl-C:

- `destroy_daemons` (line 188) never runs, so the remaining **live** daemons never receive `DaemonCoordinatorEvent::Destroy` and `daemon_connections` is never drained;
- the CLI `Destroy` request returns an error (partial teardown);
- `abortable_events.abort()` (line 166) has *already* fired, leaving the coordinator itself half torn down.

## Trigger

Run `dora destroy` (or Ctrl-C the coordinator) while at least one running dataflow still lists a daemon whose WS connection has already dropped — a common race where a daemon disconnects before its membership was cleaned up — or while any daemon returns a stop error.

## Suggested fix

During destroy, do not let a single dataflow's stop failure abort the whole teardown. Log and continue so `destroy_daemons` always runs:

```rust
for dataflow_uuid in running_dataflows.keys().cloned().collect::<Vec<_>>() {
    if let Err(e) = stop_dataflow(running_dataflows, dataflow_uuid, daemon_connections,
                                  clock.new_timestamp(), None, false).await {
        tracing::warn!("failed to stop dataflow {dataflow_uuid} during destroy: {e:#}");
    }
}
destroy_daemons(daemon_connections, clock.new_timestamp()).await?;
```

## Confidence

High that the control flow behaves as described (the `?` short-circuits before `destroy_daemons`).

phil-opp:
Fixes #2391

## Summary

In `handle_destroy` (`binaries/coordinator/src/handlers.rs`), each running dataflow was stopped with `.await?`. The `?` caused early return on the first stop failure, so `destroy_daemons` — which tells daemons to shut down and drains `daemon_connections` — was never reached when any dataflow stop errored. The result: live daemons were left running after `dora destroy` if any dataflow's stop failed (e.g. because its daemon had already disconnected).

## Change

Replace `let _ = stop_dataflow(...).await?` with `if let Err(e) = ... { tracing::warn!(...) }` so failures are logged and the loop continues. `destroy_daemons` is now always called regardless of individual stop results.

All 124 coordinator tests pass.

---

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -174,7 +174,7 @@ pub(crate) async fn handle_destroy(
         }
     }
     for dataflow_uuid in running_dataflows.keys().cloned().collect::<Vec<_>>() {
-        let _ = stop_dataflow(
+        if let Err(e) = stop_dataflow(
             running_dataflows,
             dataflow_uuid,
             daemon_connections,
@@ -182,7 +182,10 @@ pub(crate) async fn handle_destroy(
             None,
             false,
         )
-        .await?;
+        .await
+        {
+            tracing::warn!("failed to stop dataflow {dataflow_uuid} during destroy: {e:#}");
+        }
     }
 
     destroy_daemons(daemon_connections, clock.new_timestamp()).await?;


phil-opp:
No issues found.

Replacing `stop_dataflow(...).await?` with a log-and-continue (`if let Err(e) = ... { warn!(...) }`) correctly ensures `destroy_daemons` and the subsequent grace-period sleep are always reached even when an individual dataflow stop fails (e.g. its daemon already disconnected). `stop_dataflow`'s only side effects are the network sends to daemons, and its `Ok(&mut RunningDataflow)` return value was already discarded by the original `let _`, so dropping it in the new form introduces no borrow or state issue.

---

---

PR #2406, opened 2026-06-29
fix(coordinator): reject daemon events sent before Register

### LINKED ISSUE #2390: coordinator: unregistered /api/daemon WS connection can forge events for any daemon (impersonation guard only applies after Register)

## Summary

The daemon WebSocket handler validates `daemon_id` against the registered identity **only when the connection has already registered**. A connection that has not yet sent `CoordinatorRequest::Register` has `tracked_daemon_id == None`, so the impersonation guard is skipped entirely and any `CoordinatorRequest::Event { daemon_id, event }` it sends is accepted and forwarded to the coordinator event loop with the *attacker-supplied* `daemon_id`.

## Location

`binaries/coordinator/src/ws_daemon.rs:173-188`

```rust
CoordinatorRequest::Event { daemon_id, event } => {
    // Verify daemon_id matches the registered one (prevent impersonation)
    if let Some(tracked) = tracked_daemon_id
        && *tracked != daemon_id
    {
        tracing::warn!(
            "daemon sent event with mismatched id: expected `{tracked}`, got `{daemon_id}` — closing connection"
        );
        return false;
    }
    if let Some(coordinator_event) = translate_daemon_event(daemon_id, event) {
        event_tx.send(coordinator_event).await.is_ok()
    } else {
        true
    }
}
```

The guard is `if let Some(tracked) = tracked_daemon_id && *tracked != daemon_id`. When `tracked_daemon_id` is `None` (no `Register` yet on this connection), the `if let Some(..)` is false, the branch is skipped, and the event is translated + forwarded unconditionally.

## Problem

The comment states the intent is to "Verify daemon_id matches the registered one (prevent impersonation)". Before registration there is no registered identity, so the safe behavior is to **reject** events from an unregistered connection. Instead they are accepted with whatever `daemon_id` the sender chose.

Forwarded events are then acted on by the coordinator's main loop. Examples of impact via `translate_daemon_event` (`ws_daemon.rs:192-271`):

- `DaemonEvent::Exit` → `Event::DaemonExit { daemon_id }`, which at `binaries/coordinator/src/lib.rs:2199-2225` runs `daemon_connections.remove(&daemon_id)` plus `cleanup_disconnected_daemons_from_running_dataflows` — i.e. it tears down a *different, legitimately-registered* daemon and its dataflows.
- `DaemonEvent::AllNodesFinished { .. }` → `DataflowEvent::DataflowFinishedOnDaemon`, prematurely advancing/finishing a running dataflow.
- `StatusReport`, `NodeMetrics`, `Heartbeat`, etc. — forged state/metrics for arbitrary daemons.

## Trigger

Any client that can open the `/api/daemon` WebSocket and sends, *before* a `Register`, a request such as:

```jsonc
{ "method": "...", "params": { "inner": { "Event": { "daemon_id": "<victim>", "event": "Exit" } }, "timestamp": ... } }
```

No registration (and, depending on deployment, no authentication) is required to reach this arm.

## Suggested fix

Reject `CoordinatorRequest::Event` when `tracked_daemon_id` is `None` (an event before `Register` is always invalid), and otherwise keep the existing equality check. For example:

```rust
CoordinatorRequest::Event { daemon_id, event } => {
    match tracked_daemon_id {
        None => {
            tracing::warn!("daemon sent event before registering — closing connection");
            return false;
        }
        Some(tracked) if *tracked != daemon_id => { /* existing mismatch path */ return false; }
        Some(_) => { /* forward */ }
    }
    ...
}
```

## Confidence

High that the code path behaves as described; the practical impact depends on whether the `/api/daemon` endpoint is network-reachable by untrusted peers in a given deployment.

phil-opp:
Fixes #2390

## Summary

The `/api/daemon` WebSocket handler verified that `daemon_id` in an incoming `CoordinatorRequest::Event` matched the registered identity — but only when `tracked_daemon_id` was `Some`. Before a `Register` message (`tracked_daemon_id == None`), the guard was an `if let Some(..)` that evaluated to false, so the entire identity check was skipped. Any client that could open the `/api/daemon` socket could then inject arbitrary events (e.g. `DaemonExit` for a legitimate daemon, `AllNodesFinished` for a running dataflow) without registering first.

## Change

Replace the `if let Some(tracked) = tracked_daemon_id && *tracked != daemon_id` guard with a `match` on `tracked_daemon_id` that explicitly handles:
- `None` → log a warning and close the connection ("daemon sent event before registering")
- `Some(tracked) if *tracked != daemon_id` → log a warning and close (existing mismatch path)
- `Some(_)` → forward the event normally

All 124 coordinator tests pass.

---

### DIFF
--- a/binaries/coordinator/src/ws_daemon.rs
+++ b/binaries/coordinator/src/ws_daemon.rs
@@ -171,14 +171,18 @@ async fn handle_daemon_request(
             true
         }
         CoordinatorRequest::Event { daemon_id, event } => {
-            // Verify daemon_id matches the registered one (prevent impersonation)
-            if let Some(tracked) = tracked_daemon_id
-                && *tracked != daemon_id
-            {
-                tracing::warn!(
-                    "daemon sent event with mismatched id: expected `{tracked}`, got `{daemon_id}` — closing connection"
-                );
-                return false;
+            match tracked_daemon_id {
+                None => {
+                    tracing::warn!("daemon sent event before registering — closing connection");
+                    return false;
+                }
+                Some(tracked) if *tracked != daemon_id => {
+                    tracing::warn!(
+                        "daemon sent event with mismatched id: expected `{tracked}`, got `{daemon_id}` — closing connection"
+                    );
+                    return false;
+                }
+                Some(_) => {}
             }
             if let Some(coordinator_event) = translate_daemon_event(daemon_id, event) {
                 event_tx.send(coordinator_event).await.is_ok()


phil-opp:
No issues found.

The new `match` correctly closes an impersonation path. Before this change, a daemon WebSocket connection that had not yet registered (`tracked_daemon_id == None`) bypassed the `if let Some(tracked)` guard entirely, so a forged `CoordinatorRequest::Event` carrying an arbitrary `daemon_id` (e.g. `DaemonEvent::Exit` → `Event::DaemonExit { daemon_id }`) was translated and dispatched, forcing cleanup of a legitimate daemon. The added `None` arm rejects pre-registration events while preserving the existing mismatch check and the registered pass-through unchanged.

---

---

PR #2407, opened 2026-06-29
fix(coordinator): guard DaemonExit against stale connection from daemon reconnect race

### LINKED ISSUE #2392: coordinator: daemon reconnect race — stale connection's DaemonExit can evict the freshly re-registered connection

## Summary

When a named daemon reconnects, the coordinator **reuses the same `DaemonId`** and `DaemonConnections::add` replaces the old connection with the new one. The old connection's handler task, when it finally observes its socket close, emits `Event::DaemonExit { daemon_id }` with that same id. `Event::DaemonExit` removes the connection by `daemon_id` **unconditionally**, with no connection-identity / generation token. If the stale `DaemonExit` is processed *after* the new connection has re-registered, the coordinator deletes the brand-new, healthy connection and runs disconnect cleanup against it (tearing down its dataflows).

## Locations

Reused `DaemonId` on re-register — `binaries/coordinator/src/lib.rs:406-410`:

```rust
// Reuse existing DaemonId if daemon is re-registering (#5 fix)
let daemon_id = match existing {
    Some(existing_id) => existing_id.clone(),
    None => DaemonId::new(machine_id),
};
// ... later:
daemon_connections.add(daemon_id.clone(), connection);   // lib.rs:429 — replaces old entry
```

Old connection task emits `DaemonExit` on socket close — `binaries/coordinator/src/ws_daemon.rs:92-96`:

```rust
// Emit DaemonExit on WS close for immediate cleanup
if let Some(daemon_id) = tracked_daemon_id {
    tracing::info!("daemon WS connection closed for `{daemon_id}`");
    let _ = event_tx.send(Event::DaemonExit { daemon_id }).await;
}
```

Unconditional removal — `binaries/coordinator/src/lib.rs:2199-2225`:

```rust
Event::DaemonExit { daemon_id } => {
    tracing::info!("Daemon `{daemon_id}` exited");
    daemon_connections.remove(&daemon_id);          // removes whatever is currently registered
    ...
    let disconnected = BTreeSet::from([daemon_id]);
    let disconnect_actions = cleanup_disconnected_daemons_from_running_dataflows(...);
    ...
}
```

## Problem

There is no way to distinguish a `DaemonExit` originating from the *old* (already-replaced) connection from the live one. The `remove`/cleanup operate purely on `daemon_id`, which is identical across the reconnect. So a late `DaemonExit` from the stale task evicts the new connection.

## Trigger

A named daemon drops and reconnects (transient network blip, coordinator-side reconnect, etc.) such that the new `Register` (and its `daemon_connections.add`) is processed **before** the old connection task's `DaemonExit` reaches the event loop. Because the old task only emits `DaemonExit` after it detects the socket close — which can be arbitrarily delayed — this ordering is realistic.

## Suggested fix

Attach a per-connection generation/epoch (or a unique connection token) to each accepted connection and ignore a `DaemonExit` whose token does not match the currently-registered connection for that `daemon_id`. Alternatively, have `handle_daemon_ws` only emit `DaemonExit` if it is still the active connection for that id at close time.

## Confidence

Medium — this is a timing-dependent race; the code path and the absence of any identity/generation guard are confirmed, but triggering it requires the specific reconnect ordering above.

phil-opp:
Fixes #2392

## Summary

When a named daemon reconnects, the coordinator reuses its `DaemonId` and `DaemonConnections::add` replaces the old connection entry with the new one. The old connection's handler task still emits `Event::DaemonExit { daemon_id }` when it detects the socket close. If this stale `DaemonExit` is processed *after* the new connection has registered, the handler calls `daemon_connections.remove(&daemon_id)` — removing the freshly registered healthy connection — and then runs disconnect cleanup against the new connection's dataflows.

## Fix

Assign a unique `connection_id: Uuid` to each `DaemonConnection` at creation time. Include `connection_id` in `Event::DaemonExit`. In the `DaemonExit` handler, only proceed with removal and cleanup when the stored `connection_id` matches the current registration — a stale exit from a replaced connection is silently dropped with a `debug!` log.

## Changes

- **`state.rs`**: Add `connection_id: Uuid` to `DaemonConnection` (generated via `Uuid::new_v4()` in `new()`); add `DaemonConnections::connection_id_of()` lookup
- **`events.rs`**: Add `connection_id: Uuid` field to `Event::DaemonExit`
- **`ws_daemon.rs`**: Track `tracked_connection_id` alongside `tracked_daemon_id`; capture `connection.connection_id` before moving `connection` into the `Register` event; pass `connection_id` through both the WS-close and explicit `DaemonEvent::Exit` paths
- **`lib.rs`**: Check `connection_id_of(&daemon_id) == Some(connection_id)` before removing; skip cleanup with a `debug!` log on mismatch

All coordinator tests pass. Also includes `.gitignore` update to ignore `.claude/worktrees/` (temporary agent working directories).

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,8 +200,9 @@ out/
 # MacOS DS_Store
 .DS_Store
 
-# Claude Code local settings (machine-specific)
+# Claude Code local settings and temporary agent worktrees (machine-specific)
 .claude/settings.local.json
+.claude/worktrees/
 
 # macOS debug symbol bundles
 *.dSYM/
--- a/binaries/coordinator/src/events.rs
+++ b/binaries/coordinator/src/events.rs
@@ -33,6 +33,10 @@ pub enum Event {
     Log(LogMessage),
     DaemonExit {
         daemon_id: DaemonId,
+        /// Connection-instance ID stamped at registration time. Used to guard
+        /// against a stale connection's exit evicting a freshly re-registered
+        /// connection that reused the same `DaemonId` (#2392).
+        connection_id: Uuid,
     },
     DataflowBuildResult {
         build_id: BuildId,
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2196,32 +2196,47 @@ async fn start_inner(
                         .await;
                 }
             }
-            Event::DaemonExit { daemon_id } => {
-                tracing::info!("Daemon `{daemon_id}` exited");
-                daemon_connections.remove(&daemon_id);
-                if let Err(e) = store.unregister_daemon(&daemon_id) {
-                    tracing::warn!("failed to persist daemon unregistration: {e}");
+            Event::DaemonExit {
+                daemon_id,
+                connection_id,
+            } => {
+                // Only act on the exit if the connection_id matches the
+                // currently-registered connection. A named daemon that
+                // reconnects before the old connection's handler task
+                // delivers its DaemonExit must NOT evict the new connection
+                // or trigger spurious dataflow cleanup (#2392).
+                if daemon_connections.connection_id_of(&daemon_id) == Some(connection_id) {
+                    tracing::info!("Daemon `{daemon_id}` exited");
+                    daemon_connections.remove(&daemon_id);
+                    if let Err(e) = store.unregister_daemon(&daemon_id) {
+                        tracing::warn!("failed to persist daemon unregistration: {e}");
+                    }
+                    let disconnected = BTreeSet::from([daemon_id]);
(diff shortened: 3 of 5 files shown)


phil-opp:
No issues found.

The connection-id approach to the reconnect race (#2392) is sound: stamping each `DaemonConnection` with a unique `connection_id` and gating the `DaemonExit` handler on `connection_id_of(&daemon_id) == Some(connection_id)` correctly distinguishes a stale connection's exit from a freshly re-registered one. I traced both event orderings in the serial coordinator loop and both resolve correctly — the captured `connection_id` is the same value moved into `daemon_connections.add()`, and the tracked id is only committed on successful registration, so a stale `DaemonExit` cannot evict the new connection.

---

---

PR #2424, opened 2026-06-30
fix(daemon): include debug watchers in need_data_bytes gate in send_out

### LINKED ISSUE #2423: daemon: `dora topic echo` shows empty data for nodes without a Zenoh session (send_out drops payload for debug watchers)
> **Note:** This issue was filed by an automated, scheduled Claude code-review check. The finding was manually traced through the code paths below, but please validate before acting on it.

## Summary

In `Daemon::send_out`, the data payload requested from `send_output_to_local_receivers` is gated **only** on `remote_receivers`, not on whether a topic-debug watcher is active. As a result, when a node produces output through the **daemon control-plane fallback path** (no Zenoh session) and a `dora topic echo` / debug stream is watching that output, the frame forwarded to the watcher is serialized with `data: None` — the payload is silently dropped and `dora topic echo` displays empty data even though the node sent a real payload.

## Location

`binaries/daemon/src/lib.rs`, `send_out` (around lines 3668–3711):

```rust
let remote_receivers = dataflow.open_external_mappings.contains(&output_id_key)
    || dataflow.enable_debug_inspection;
let has_debug_watchers = dataflow.debug_topic_watchers.contains_key(&output_id_key);
let data_bytes = send_output_to_local_receivers(
    node_id.clone(),
    output_id.clone(),
    dataflow,
    &metadata,
    data,
    &self.clock,
    Some(&self.ft_stats),
    remote_receivers,   // <-- need_data_bytes: ignores has_debug_watchers
)
.await?;

if !remote_receivers && !has_debug_watchers {
    return Ok(());
}
// ...
if has_debug_watchers {
    if remote_receivers {
        self.send_topic_debug_frames(dataflow_id, &output_id, serialized_event.clone()).await?;
    } else {
        // dedicated branch for "debug watcher, no remote receivers"
        self.send_topic_debug_frames(dataflow_id, &output_id, serialized_event).await?;
        return Ok(());   // <-- frame here carries data: None
    }
}
```

`send_output_to_local_receivers` returns `None` for the payload whenever `need_data_bytes` is `false` (`binaries/daemon/src/lib.rs:5214–5227`):

```rust
let data_bytes = if need_data_bytes {
    data.map(|arc| /* move/clone out the Vec */)
} else {
    None
};
```

So in the `else` branch above (`has_debug_watchers && !remote_receivers`), `data_bytes` is always `None`, and `send_topic_debug_frames` forwards an `InterDaemonEvent::Output { data: None, .. }` to the watching CLI.

## Why this is a bug (internal inconsistency)

The **Zenoh data path** serves the *same* debug watchers correctly. `handle_debug_topic_data` (`binaries/daemon/src/lib.rs:2545–2578`) builds the identical `InterDaemonEvent::Output` frame and includes the payload, gated only on `has_watchers` (i.e. `debug_topic_watchers`), **independently of `enable_debug_inspection`**:

```rust
let event = InterDaemonEvent::Output {
    // ...
    data: data.map(|d| AVec::from_slice(128, &d)),   // payload included
};
```

The dedicated `else` branch in `send_out` exists specifically to serve debug watchers when there are no remote receivers — but because `need_data_bytes` was computed without `has_debug_watchers`, it ships an empty-payload frame. The two paths therefore disagree: a `dora topic echo` watcher sees full data for a Zenoh-published output but empty data for a daemon-routed (fallback) output of the same dataflow.

Note the bug is **masked when `enable_debug_inspection: true`** is set on the dataflow, because that forces `remote_receivers = true` (line 3669), which in turn makes `data_bytes` populated. It only manifests when `enable_debug_inspection` is not set — yet the Zenoh path shows debug inspection is meant to work without it.

## Concrete failure scenario

1. A node runs **without a Zenoh session** (e.g. interactive/testing mode, or Zenoh failed to open), so output goes through the daemon control-plane fallback: `DoraNode::send_output` calls `control_channel.send_message(output_id, metadata, Some(data))` (`apis/rust/node/src/node/mod.rs:1018–1022`), and the daemon's `send_out` receives `Some(data)`.
2. A user runs `dora topic echo <node>/<output>`; the coordinator issues `StartTopicDebugStream`, which populates `debug_topic_watchers` (`lib.rs:2264–2285`). This does **not** set `enable_debug_inspection`.
3. The output has no external/cross-daemon receivers, and `enable_debug_inspection` is not set, so `remote_receivers == false`.
4. → `need_data_bytes == false` → `data_bytes == None` → the watcher receives a payload-less frame; `dora topic echo` shows the output name and metadata but no `data` (the CLI renders `data` only `if let Some(data) = data`, `binaries/cli/src/command/topic/echo.rs:173`).

## Suggested fix

Include debug watchers in the payload gate so the fallback path mirrors the Zenoh path:

```rust
let need_data_bytes = remote_receivers || has_debug_watchers;
let data_bytes = send_output_to_local_receivers(
    // ...
    need_data_bytes,
)
.await?;
```

A regression test exercising `send_out` with `has_debug_watchers = true`, `remote_receivers = false`, and asserting the forwarded `InterDaemonEvent::Output` carries `Some(data)` would lock this in.

phil-opp:
## Summary

Fixes #2423.

`Daemon::send_out` computed `need_data_bytes` from `remote_receivers` alone, ignoring `has_debug_watchers`. When a node sends output through the daemon control-plane fallback path (no Zenoh session) and only a `dora topic echo` / debug stream is watching, `remote_receivers` is `false` while `has_debug_watchers` is `true`. This caused `send_output_to_local_receivers` to return `None` for the payload, so the frame forwarded to the watcher carried `data: None` — `dora topic echo` showed empty output even though the node sent real data.

## Root cause

```rust
// Before — data_bytes is always None when !remote_receivers
let data_bytes = send_output_to_local_receivers(
    ...
    remote_receivers,   // need_data_bytes: ignores has_debug_watchers
)
.await?;
```

## Fix

```rust
// After — payload is retained whenever anyone will consume it
let data_bytes = send_output_to_local_receivers(
    ...
    remote_receivers || has_debug_watchers,
)
.await?;
```

This matches the behaviour of the Zenoh data path (`handle_debug_topic_data`), which always includes the payload when debug watchers are present, independently of `enable_debug_inspection`.

## Verification

- `cargo test -p dora-daemon` — 89 passed, 0 failed

🤖 Generated with [Claude Code](https://claude.ai/code/session_01QwbPZwQpqSXdScUcpH3VM1)

---
_Generated by [Claude Code](https://claude.ai/code/session_01QwbPZwQpqSXdScUcpH3VM1)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -3676,7 +3676,7 @@ impl Daemon {
             data,
             &self.clock,
             Some(&self.ft_stats),
-            remote_receivers,
+            remote_receivers || has_debug_watchers,
         )
         .await?;
 


phil-opp:

No issues found. I traced the full data flow and the one-line fix is correct and complete:

- `send_output_to_local_receivers` only retains the payload as `data_bytes` when `need_data_bytes == true`; otherwise it returns `None`, which then becomes the `data: None` forwarded to debug watchers via `send_topic_debug_frames`. So gating on `remote_receivers || has_debug_watchers` is the right condition.
- This matches the Zenoh debug path (`handle_debug_topic_data`), which always forwards the payload regardless of `enable_debug_inspection`.
- The other call sites of `send_output_to_local_receivers` (notably the inter-daemon receive path) correctly pass `false` and don't participate in this gate, so `send_out` was the only site needing the change — no missed call sites.

One minor note (not blocking): no regression test was added for the `has_debug_watchers && !remote_receivers` path, so nothing guards against this regressing. A unit test on `send_out`, or a `dora topic echo` smoke test over the control-plane fallback, would close that gap.

---

PR #2439, opened 2026-07-01
fix(node): bound zenoh subscriber teardown to prevent node shutdown hang (#2425)

phil-opp:
A node with user-mapped inputs declares a per-input zenoh callback
subscriber on the shared session. `EventStream`'s `_zenoh_subscribers`
field dropped these inline, and `Subscriber::drop` (undeclare) blocks
indefinitely when zenoh's net runtime is wedged (e.g. stuck retrying an
unreachable scouted peer). Because that field drops before
`DoraNode::Drop` — which already bounds its own session teardown — a
wedged runtime hangs the whole node in the field-drop sequence, so the
daemon never sees the node finish and the dataflow stalls until an outer
timeout. This surfaced as the intermittent nightly hangs in #2425
(record-replay: `dora record` never returns; multiple-daemons: "dataflow
not finished after 101 retries").

Tear the subscribers down inside `EventStream::Drop` under the existing
`teardown_with_timeout` / `ZENOH_TEARDOWN_TIMEOUT` guard, mirroring the
`DoraNode` session teardown. The callbacks use `try_send`, so undeclaring
before `receiver` drops cannot deadlock on a blocked callback.

Reproduced locally (~10-20% of `dora record` runs hung before the fix;
0/60 after). Nodes without user-mapped inputs (timer-only) were never
affected because they declare no subscribers.

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -28,6 +28,7 @@ use crate::{
     DaemonCommunicationWrapper, PatternError,
     daemon_connection::{DaemonChannel, node_integration_testing::convert_output_to_json},
     event_stream::data_conversion::RawData,
+    node::{ZENOH_TEARDOWN_TIMEOUT, teardown_with_timeout},
 };
 use dora_core::{
     config::{Input, NodeId},
@@ -75,9 +76,11 @@ pub struct EventStream {
     receiver: tokio::sync::mpsc::Receiver<EventItem>,
     _thread_handle: EventStreamThreadHandle,
     /// Callback subscribers — kept alive for the lifetime of the
-    /// EventStream. Dropping a subscriber stops further callbacks; the
-    /// callback itself will see `blocking_send` return Err once the
-    /// `receiver` (declared above) is dropped.
+    /// EventStream. Dropping a subscriber undeclares it and stops further
+    /// callbacks. The callbacks use `try_send` (never block), so undeclaring
+    /// does not depend on `receiver` being dropped first. `EventStream::drop`
+    /// explicitly tears these down under a deadline (see there); by the time
+    /// this field drops it is already empty.
     _zenoh_subscribers: Vec<zenoh::pubsub::Subscriber<()>>,
     close_channel: DaemonChannel,
     clock: Arc<uhlc::HLC>,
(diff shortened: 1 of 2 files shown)


phil-opp:
**Automated review by Claude** (fully automated — no human in the loop)

Reviewed the diff (moving `_zenoh_subscribers` teardown into `EventStream::Drop` under the existing `teardown_with_timeout` / `ZENOH_TEARDOWN_TIMEOUT` guard). No issues found.

Things I specifically checked and cleared:

- **No undeclare-vs-callback deadlock:** the subscriber callbacks use `try_send` (not `blocking_send`), so `Subscriber::drop` (undeclare) can't block on an in-flight callback. The stale base comment referencing `blocking_send` is corrected in the diff.
- **No use-after-free of the session:** zenoh subscribers hold an owning (Arc-cloned) session handle, so a subscriber leaked by a timed-out teardown thread keeps the runtime alive independently of `DoraNode::Drop` dropping its own `zenoh_session` — safe even if `DoraNode` drops first/concurrently.
- **No double-drop:** `std::mem::take` empties the field before the implicit field-drop runs on an empty `Vec`.
- **Signal invariant preserved:** dropping the subscribers earlier in `Drop::drop` only closes their `tx_cb` senders sooner, which is strictly safe for unblocking `receiver`.
- **Leak-on-timeout** matches the already-accepted tradeoff documented for the `DoraNode` session teardown; the `!is_empty()` guard avoids spawning a thread for timer-only nodes.

The change is small, targeted, and correctly reuses the existing bounded-teardown mechanism.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; treat it accordingly.

The core fix looks correct: the callbacks use `try_send` (not `blocking_send`), so undeclaring the subscribers at the top of `EventStream::Drop` before `receiver` drops can't deadlock on an in-flight callback; `mem::take` avoids a double-drop; and the subscribers hold an `Arc` clone of the session, so a subscriber leaked on timeout can't cause a use-after-free relative to `DoraNode::Drop`. Nice.

One thing worth a second look before merge — **cumulative teardown budget**. `EventStream` and `DoraNode` hold handles to the *same* zenoh session, and each now runs its own independent `teardown_with_timeout` (15s `ZENOH_TEARDOWN_TIMEOUT` each) in its own `Drop`. In the wedged-runtime case they run sequentially (the usual `let (node, events) = ...` drops them one after the other), so a node can now block up to ~2×15s ≈ 30s before exiting, rather than 15s. That's still a strict improvement over the previous unbounded hang, but if the outer daemon/dataflow timeout that #2425 relies on assumes a single ~15s bound, a wedged node could still overshoot it. Consider sharing a single deadline across the two teardowns, or documenting the 2× worst case.

Separately: the failing **Audit** check appears unrelated to this diff — it's tripping on newly-published RUSTSEC advisories in transitive deps (e.g. `quick-xml` RUSTSEC-2026-0194/0195, published 2026-06-29), and this PR touches only two `.rs` files with no `Cargo.toml`/`Cargo.lock` changes, so it can't have introduced them. All CircleCI statuses are green.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; treat it accordingly.

I traced the `Drop` path again and I think this fix is **incomplete**. It bounds teardown of `_zenoh_subscribers` (the data callback subscribers) under `teardown_with_timeout`, but `EventStream` holds a *second* subscriber field that the fix doesn't touch:

```rust
// apis/rust/node/src/event_stream/mod.rs
_zenoh_subscribers: Vec<zenoh::pubsub::Subscriber<()>>,              // line 79 — torn down under the deadline
/// Per-input `@schema` subscribers (zenoh-ext AdvancedSubscriber) ...
_zenoh_schema_subscribers: Vec<zenoh_ext::AdvancedSubscriber<()>>,   // line 82 — NOT torn down
```

`_zenoh_schema_subscribers` is never `mem::take`-n or torn down inside the explicit `Drop::drop`, so it drops **inline as a struct field** after `drop()` returns, in declaration order — and unbounded. `AdvancedSubscriber::drop` undeclares its subscription(s) on the same shared zenoh session, which is exactly the operation this PR describes as blocking indefinitely when zenoh's net runtime is wedged.

The two subscriber vectors are populated *together*, for the same inputs: in the input loop, each `InputMapping::User` input calls `declare_schema_subscriber(...)` (pushes to `zenoh_schema_subscribers`) and then `declare_subscriber(...)` (pushes to `zenoh_subscribers`). So a node with user-mapped inputs — the #2425 scenario this PR targets — has both vectors non-empty. On a wedged runtime the data-subscriber teardown now completes/bounds correctly, then the inline field-drop of `_zenoh_schema_subscribers` blocks forever, and the node still hangs in `EventStream`'s field-drop sequence. That's the same failure mode #2425 describes.

Suggest tearing the schema subscribers down under the same deadline too — e.g. `std::mem::take(&mut self._zenoh_schema_subscribers)` and drop them inside the same bounded closure (or a second `teardown_with_timeout`). Otherwise the data-subscriber path is fixed but a parallel unbounded hang path remains for exactly the inputs that trigger the bug.

(For reference: none of this changes the earlier-noted cumulative-budget point — if the schema subscribers get their own separate `teardown_with_timeout`, that's a third independent 15s bound to keep in mind.)

---

Issue #1625, opened 2026-04-17
[NOTICE] main frozen for dora 1.0 consolidation


heyong4725:
**dora-rs/dora is frozen for 1.0 consolidation (2026-04-17 → ~2026-07-31)**

The `main` branch of this repo is under branch protection while the [`dora-rs/adora` → `dora-rs/dora` consolidation](https://github.com/dora-rs/adora/issues/305) lands as dora 1.0. Non-critical PRs will hold until the freeze lifts at Phase 5c (GA tag).

Critical security patches: ping @heyong4725 directly. We can land CVE-level fixes during the freeze with admin bypass.

Tracking: [Consolidation execution epic](https://github.com/dora-rs/adora/issues/305) · [Plan document](https://github.com/dora-rs/adora/blob/main/docs/plan-dora-1.0-consolidation.md)

**Expected lift:** Phase 5c tag of `v1.0.0` (revised to ~2026-07-31 from the prior ~2026-06-01 estimate — consolidation is taking longer than planned; see #1624 draft PR for current status). Updated 2026-06-22.

EthanFrostpro:
really glad to see this heading toward 1.0. been watching the adora → dora consolidation from the sidelines and the phased approach makes a lot of sense for something this complex. the dataflow pipeline model is genuinely one of the cleaner ways i've seen to handle multi-modal robotics data compared to ROS2's more traditional pub/sub. quick question - after the freeze lifts will there be a migration guide for folks who built on the pre-1.0 API? or is the plan to keep backward compat through the transition?

heyong4725:
Thanks @EthanFrostpro — short answer: **yes to both, and for app code it's mostly a no-op.**

## For app developers (node/operator code + YAML + CLI):

Your existing pre-1.0 code should compile and run against 1.0 with little-to-no change:

- **Rust node/operator APIs** — `dora-node-api`, `dora-operator-api` trait signatures and imports unchanged. Crate names unchanged (no rename, same `Cargo.toml` entries).
- **Python node API** — imports stay the same (`from dora import Node`). PyO3 surface is additive.
- **C / C++ node APIs** — `apis/c/node` and `apis/c++/node` headers and symbols preserved.
- **YAML dataflow descriptor** — most 0.x descriptors work unchanged. A few fields gained new defaults or small renames, documented in [`yaml-spec.md`](https://github.com/dora-rs/dora/blob/main/docs/yaml-spec.md).
[...]

PR #2442, opened 2026-07-02
fix(core): rewrite operator/custom module inputs, not just node-level inputs

### LINKED ISSUE #2441: Module expansion does not rewrite operator/custom inputs, leaving `_mod/` and sibling references dangling

## Summary

In `libraries/core/src/descriptor/expand.rs`, module expansion only rewrites a node's **top-level `inputs`** map. For **runtime nodes** (nodes that use `operators:` / `operator:`) and legacy `custom:` nodes, the actual input wiring lives in the operator/custom `config.inputs` map, not the node-level `inputs`. Those inputs are never passed through `rewrite_module_input`, so:

1. `_mod/<port>` references inside an operator/custom input are **not** resolved to the module node's binding, and
2. internal sibling references (e.g. `stage_a/out`) inside an operator/custom input are **not** prefixed with the module id.

The result is a flat descriptor with dangling inputs (`_mod` is not a real node; the sibling id no longer exists because siblings were prefixed to `<module_id>.<sibling>`).

## Where

Phase 1 of `expand_module_node` only iterates `inner_node.inputs`:

https://github.com/dora-rs/dora/blob/main/libraries/core/src/descriptor/expand.rs#L377-L393

```rust
// Rewrite inputs (None = optional input not provided, skip it)
let mut new_inputs = BTreeMap::new();
for (input_id, input) in &inner_node.inputs {
    match rewrite_module_input(
        input,
        &module_id,
        &node.inputs,
        &inner_node_ids,
        &optional_inputs,
    )? {
        Some(new_input) => { new_inputs.insert(input_id.clone(), new_input); }
        None => continue, // optional input not provided
    }
}
inner_node.inputs = new_inputs;
```

`rewrite_module_input` — the only function that resolves `_mod/` and prefixes sibling refs — is called *only* here (grep confirms a single call site at `expand.rs:380`).

`rewrite_external_refs` *does* iterate operator and custom inputs:

https://github.com/dora-rs/dora/blob/main/libraries/core/src/descriptor/expand.rs#L593-L607

…but it only rewrites references that match a **module output map** (`output_maps`), via `rewrite_inputs_map`. It never handles `_mod/` (`MODULE_INPUT_SOURCE`) or internal sibling cross-references. So operator/custom inputs get the "external module output" treatment but miss the two rewrites that Phase 1 applies to node-level inputs.

The isolated validator `check_module_file` has the same blind spot — it only checks `node.inputs` for `_mod/` references (`expand.rs:188-208`), so a malformed operator input isn't caught at validation time either.

## Reproduction

Module file `runtime_module.yml`:

```yaml
module:
  name: rt
  inputs: [data]
  outputs: [result]

nodes:
  - id: runtime
    operators:
      - id: proc
        python: proc.py
        inputs:
          x: _mod/data        # references a declared module input
        outputs: [result]
```

Dataflow:

```yaml
nodes:
  - id: src
    path: src.py
    outputs: [val]
  - id: m
    module: runtime_module.yml
    inputs:
      data: src/val
```

**Expected** after expansion: operator `proc` input `x` → `src/val`.
**Actual:** operator `proc` input `x` stays `_mod/data` (source `_mod`, which is not a node), producing a dangling input in the flattened descriptor.

An analogous failure happens for an internal sibling reference inside an operator input: it stays unprefixed (`stage_a/out`) even though the sibling node was renamed to `m.stage_a`, so it points at a non-existent node.

## Impact

Any module whose inner nodes are runtime (operator) nodes — or legacy `custom` nodes — that consume module inputs or sibling outputs will silently produce a broken/mis-wired dataflow at expansion time. Node-only modules (the case every existing test covers) are unaffected, which is why this slipped through: `expand.rs`'s test module has no coverage for operators/custom inner nodes.

## Suggested fix

Run `rewrite_module_input` over the operator/custom `config.inputs` maps in Phase 1, mirroring the node-level `inputs` handling (and extend `check_module_file` to validate operator/custom `_mod/` references). Add regression tests for a runtime/operator inner node that references both `_mod/<port>` and a sibling node.

phil-opp:
Fixes #2441

## Summary

Module expansion (`libraries/core/src/descriptor/expand.rs`) only rewrote a node's top-level `inputs` map. Runtime (operator) nodes and legacy `custom` nodes wire their inputs through `config.inputs` / `run_config.inputs` instead, so `_mod/` references and sibling cross-references inside those inputs were never resolved — producing a flat descriptor with dangling inputs (`_mod` is not a real node; unprefixed sibling ids no longer exist after expansion). `check_module_file`'s isolated validator had the same blind spot for `_mod/` reference checking.

## Change

- Added `node_input_maps` / `node_input_maps_mut`, which centralize "where do a node's inputs live" (node-level `inputs`, plus `operators[].config.inputs`, `operator.config.inputs`, `custom.run_config.inputs`) so both the `_mod/` validator and the Phase 1 rewrite pass in `expand_module_node` stay in sync.
- `check_module_file` now validates `_mod/` references inside operator/custom inputs, not just node-level inputs.
- `expand_module_node`'s Phase 1 now rewrites `_mod/` and sibling references for operator/custom inner-node inputs the same way it already did for node-level inputs.
- Added regression tests: an operator inner node and a legacy custom inner node, each referencing both `_mod/` and a sibling node's output, plus a `check_module_file` test for an invalid `_mod/` reference inside an operator input.

## Validation

- `cargo test -p dora-core descriptor::expand::` — 30 passed (27 pre-existing + 3 new)
- `cargo test --all --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python --exclude dora-cli-api-python --exclude dora-examples` — full workspace suite green
- `cargo check --examples`
- Reviewed via `/code-review` (no findings) and `/simplify` (centralized the repeated operator/operator/custom enumeration into shared helpers per feedback from parallel review agents)

Note: this repo is currently under the 1.0 consolidation freeze noted in #1625 — opening this now per the issue-triage request; happy to hold for merge until the freeze lifts.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QNy843sppSMqVmoUmt6ooV)_


phil-opp:

I reviewed the diff against the code. The input-side fix looks correct and complete: `node_input_maps` / `node_input_maps_mut` cover all five input-bearing locations (`node.inputs`, `operators[].config.inputs`, `operator.config.inputs`, `custom.run_config.inputs`), the immutable validator and the Phase-1 rewrite share the same enumeration so they stay in sync, and the mutable borrows are of disjoint fields so there's no aliasing issue. Sibling/`_mod/` resolution for operator and custom inner nodes is exercised by the new tests. No issues there.

**I found one issue: the output side has the exact blind spot this PR fixes on the input side, and it's left unfixed.**

Module *output* resolution still only looks at node-level `outputs`:

- Phase 3 in `expand_module_node` finds the producer via `final_nodes.iter().find(|n| n.outputs.contains(declared_output))` (`libraries/core/src/descriptor/expand.rs:475-477`).
- `check_module_file` collects `inner_outputs` the same way — `.filter(|n| n.module.is_none()).flat_map(|n| n.outputs.iter()...)` (`expand.rs:212-217`), then rejects at `:255-260`.

An operator's outputs live in `op.config.outputs` and a custom node's in `custom.run_config.outputs`; both leave the top-level `node.outputs` empty. So a module that exposes an output produced by an operator/custom inner node is rejected with "declares output `X` but no inner node produces it", even though it clearly does. Example:

```yaml
module: { name: rt, inputs: [data], outputs: [result] }
nodes:
  - id: runtime
    operators:
      - id: proc
        shared-library: proc.so
        inputs: { x: _mod/data }
        outputs: [result]
```

With this PR the operator can now *consume* `_mod/data`, but the module still can't *expose* `result`. That leaves the operator/custom module feature asymmetric. Suggest an analogous "all output locations" helper (`node.outputs` + `op.config.outputs` + `operator.config.outputs` + `custom.run_config.outputs`) used by both Phase 3 and `check_module_file`, plus a test where a module output is produced by an operator.

No other correctness issues found in the diff.

---

phil-opp:
The only change since my earlier review is commit `02a6922` (`ci: retrigger CircleCI`), an empty commit with no code changes, so my earlier feedback still stands. In particular, the output-side asymmetry I flagged remains unaddressed: module outputs produced by an operator (`op.config.outputs`) or legacy custom (`run_config.outputs`) inner node are still not recognized by `check_module_file` (`expand.rs:212-217`, `:253-262`) or the Phase 3 producer lookup, so a module that exposes an operator/custom-produced output is still rejected even though the input side now resolves correctly.

---

---

phil-opp:
The `contract-tests` failure was the known flake tracked in #2064 (this PR only touches `libraries/core/src/descriptor/expand.rs`, which no `contract_*` test exercises — module expansion is only covered by `smoke_module_dataflow`, a different filter that already passed as part of `test-linux`). Pushed an empty commit to retrigger.

The `audit` failure (6 vulnerabilities, incl. RUSTSEC-2026-0194/0195 in `quick-xml`) is unrelated to this diff too — this PR doesn't touch `Cargo.lock`/`Cargo.toml`, and the advisory was only published 2026-06-29, so it started failing between my first push and the retrigger purely from the passage of time. It's already tracked in #2451 and being fixed by the in-flight dependency-upgrade PR #2445, so no action needed here.

---

PR #2443, opened 2026-07-02
fix(python-node): don't publish a memory-pool frame when the GPU copy fails

### LINKED ISSUE #2436: memory-pool (Python): GPU copy failures are discarded but the seqlock is still closed as "complete", publishing stale/garbage data to consumers
> **Note:** This issue was created by an automated, scheduled Claude code-review check. It reflects a static reading of the code (no GPU was available to reproduce at runtime), so please treat the impact assessment as "needs confirmation." The code-level facts (discarded results + unconditional seqlock close) are directly verifiable in the source.

## Summary

In `apis/python/node/src/lib.rs`, the memory-pool write paths guard each write with a seqlock at DORADMA header offset 96 (documented at `lib.rs:216` as *"u64 LE, seqlock: even = complete, odd = writing"*). Consumers use this generation counter to know when a frame is fully written and safe to read.

The **GPU copy** calls that actually move tensor data, however, discard their result with `let _ = ...` (or otherwise ignore failure), while the seqlock is **unconditionally** advanced from odd → even ("complete") immediately afterward. So if a GPU copy fails or is skipped, the consumer sees a valid "complete" generation over a buffer that was never written this cycle, and reads **stale data from the previous frame** (or uninitialized memory on the first frame). The failure is entirely silent — no warning, no error propagation.

For the CPU path this cannot happen (`std::ptr::copy_nonoverlapping` is infallible), so the bug is specific to the CUDA / DMA paths.

## Affected sites (line numbers as of current `main`)

1. **`write_memory_pool` fast-path DMA** — `dma_copy` result discarded, seqlock closed regardless:
   ```rust
       let bound = helpers.bind(py);
       let _ = bound
           .call_method1("dma_copy", (ptr_val, size, counter, !is_pinned)); // ~L1605
   }
   // Seqlock: end write  (runs even if the copy above failed or was skipped)
   unsafe { /* gen odd -> even */ }                                         // ~L1608
   ```
   If `get_cuda_helpers` returns `Err`, the copy is skipped entirely; if `dma_copy` raises (e.g. `cudaMemcpy` / `cudaHostRegister` error), the exception is swallowed. Either way the seqlock closes as "complete".

2. **`write_memory_pool` fast-path CUDA memcpy** (`is_cuda` branch) — `_cuda_memcpy` result discarded at ~L1625, seqlock closed at ~L1630.

3. **`write_memory_pool` slow-path DMA** — `dma_copy` result discarded at ~L1741, seqlock closed at ~L1744.

4. **`write_memory_pool` slow-path CUDA memcpy** — `_cuda_memcpy` discarded at ~L1761, seqlock closed at ~L1766.

5. **`register_memory_pool` initial write** — `_cuda_memcpy` result discarded at ~L1324; the `dma_copy` at ~L1344 only gates whether the IPC handle is written (`ipc_present`), and the seqlock is still closed at ~L1362 regardless of copy success.

## Why this matters

The whole point of the seqlock (see the doc comments at `lib.rs:1187`, `lib.rs:1460`) is to give consumers a consistency guarantee: an even generation ⇒ the buffer holds a fully-written frame. Discarding the copy result while still advancing the generation breaks that invariant precisely in the failure case it should protect against. A GPU copy can realistically fail (out-of-memory, an already-registered host range causing `cudaHostRegister` to error, a driver hiccup), and today that turns into silent data corruption for the downstream node rather than a surfaced error or a skipped frame.

Note this contrasts with the rest of the codebase's care around surfacing write failures (e.g. the mavlink bridge logs writer errors at `error` level; `handle_input` in `binaries/mavlink2-bridge-node/src/main.rs`).

## Suggested fix (for maintainers to weigh)

On a failed/skipped GPU copy, do **not** advance the seqlock to "complete" — leave the generation odd (or roll it back) so consumers reject the frame — and log the error at `warn`/`error` level instead of `let _ =`. At minimum, capture the `Result`, and on `Err` emit a diagnostic and skip the closing `gen += 1`. The same treatment should apply to the `get_cuda_helpers` `Err` branch (copy skipped ⇒ frame must not be marked complete).

## Repro sketch (needs a CUDA host to confirm)

1. Set up a GPU memory pool (`ipc_present == 1`), producer + consumer.
2. Inject a `cudaMemcpy`/`cudaHostRegister` failure (e.g. force an error from `dma_copy`).
3. Observe: producer's `write_memory_pool` returns `Ok(())`, seqlock generation is even, and the consumer reads the previous frame's contents as if they were new — no error surfaced anywhere.

phil-opp:
Fixes #2436

## Summary

In `apis/python/node/src/lib.rs`, the memory-pool write paths guard each write with a seqlock at DORADMA header offset 96 (`u64 LE, seqlock: even = complete, odd = writing`). The GPU copy calls that actually move tensor data (`dma_copy` / `_cuda_memcpy`) discarded their result with `let _ = ...`, while the seqlock was unconditionally advanced to "complete" immediately afterward. A failed or skipped GPU copy therefore let a consumer read stale data from the previous frame (or uninitialized memory on the pool's first frame) as if it were fresh — silently, with no error surfaced anywhere.

## Change

- Added `seqlock_begin_write` / `seqlock_end_write`: `seqlock_end_write` only advances the generation (`pre_write_gen + 2`) when the copy succeeded. On failure it rolls back to `pre_write_gen` instead — from a consumer's point of view the write never happened, so readers keep observing the last known-good, still-valid frame rather than a falsely "complete" one, and the failure is logged at `error` level instead of swallowed.
- Applied this to all 4 `write_memory_pool` call sites (fast/slow path × `dma_copy`/`_cuda_memcpy`). The CPU-only copy path is untouched — `copy_nonoverlapping` is infallible, so there's nothing to roll back.
- `register_memory_pool`'s initial GPU copy has no previous valid frame to fall back to, so on failure it now aborts registration entirely (`eyre::bail!`) rather than publishing a pool whose backing memory was never actually written. Since the shmem segment was already `cudaHostRegister`-pinned before the copy, the abort path also calls `_unregister_host` before unmapping (mirroring `free_memory_pool`'s existing cleanup order) so the CUDA driver doesn't retain a pinned registration on memory about to be unmapped and reused — this was caught by an adversarial review pass and fixed in a follow-up commit.
- Added unit tests for the pure seqlock generation arithmetic (success/failure/first-write cases) — no GPU required.

## Validation

No CUDA hardware is available in this environment, so the GPU code paths themselves couldn't be exercised at runtime (same caveat the originating issue notes: "reflects a static reading of the code"). What was validated:

- `cargo test -p dora-node-api-python` — 11 passed (8 pre-existing + 3 new)
- `cargo clippy -p dora-node-api-python -- -D warnings` — clean except one pre-existing, unrelated `downcast_into` deprecation warning (present before this change; this is why the crate is excluded from the repo's mandated clippy gate)
- `cargo clippy --all --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python -- -D warnings` — clean
- `cargo test --workspace --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python --exclude dora-cli-api-python --exclude dora-examples --lib --bins` — clean (the full `cargo test --all` example-linking step repeatedly hit this sandbox's disk quota on unrelated ros2-bridge/CLI example binaries — an environment constraint, not caused by this diff, which only touches `apis/python/node/src/lib.rs`)
- Verified the default bundled `examples/memory-pool/` dataflow defaults both sender and receiver to `device=cpu`, so the new GPU-only failure path is inert for the smoke-tested example
- Reviewed via `/code-review` at medium effort (8 independent finder angles + verification) — 2 agents independently confirmed a `_unregister_host` leak in the registration abort path, fixed in the second commit; a `seqlock_begin_write` helper was also factored out to deduplicate the seqlock-begin boilerplate across all 6 call sites

Note: this repo is currently under the 1.0 consolidation freeze noted in #1625 — opening this now per the issue-triage request; happy to hold for merge until the freeze lifts.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QNy843sppSMqVmoUmt6ooV)_

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -504,6 +504,95 @@ fn warn_missing_memory_pool(node_id: &NodeId, action: &str, buffer_id: &str) {
     );
 }
 
+/// Closes a memory-pool seqlock write (header offset 96) started at
+/// `pre_write_gen` — the generation value read just before the "begin
+/// write" increment, i.e. the last known-good, even generation.
+///
+/// On success this publishes the new frame by advancing to
+/// `pre_write_gen + 2` (restoring even parity). On failure the generation
+/// is rolled back to `pre_write_gen` instead of being advanced: from a
+/// consumer's point of view the write never happened, so readers keep
+/// observing the previous, still-valid frame rather than an falsely
+/// "complete" generation over data that was never actually written (see
+/// dora-rs/dora#2436).
+unsafe fn seqlock_end_write(
+    node_id: &NodeId,
+    context: &str,
+    gen_ptr: *mut u64,
+    pre_write_gen: u64,
+    copy_ok: bool,
+) {
+    unsafe {
+        if copy_ok {
+            std::ptr::write_volatile(gen_ptr, pre_write_gen.wrapping_add(2));
+        } else {
+            tracing::error!(
+                "[{}] memory-pool {} copy failed; keeping the previous frame visible instead of publishing a corrupt one",
+                node_id,
+                context
+            );
+            std::ptr::write_volatile(gen_ptr, pre_write_gen);
+        }
+        std::sync::atomic::fence(std::sync::atomic::Ordering::Release);
+    }
+}
+
+#[cfg(test)]
+mod seqlock_end_write_tests {
+    use super::*;
+
+    fn node_id() -> NodeId {
+        "test-node".parse().unwrap()
+    }
+
+    /// Regression test for #2436: a successful copy must publish the new
+    /// frame by advancing the generation by 2 (odd -> even, next cycle).
+    #[test]
+    fn success_advances_generation_and_stays_even() {
+        let mut generation: u64 = 5; // pre_write_gen(4) + 1 (the "begin write" step)
+        let pre_write_gen = 4;
+        unsafe {
+            seqlock_end_write(&node_id(), "test", &mut generation, pre_write_gen, true);
+        }
+        assert_eq!(generation, pre_write_gen + 2);
+        assert_eq!(
+            generation % 2,
+            0,
+            "generation must be even (complete) on success"
+        );
+    }
+
+    /// Regression test for #2436: on a failed copy, the generation must be
+    /// rolled back to `pre_write_gen` (not left odd, not advanced) so a
+    /// consumer sees the same, still-valid previous frame instead of a
+    /// "complete" generation over data that was never written.
+    #[test]
+    fn failure_rolls_back_to_pre_write_generation() {
+        let mut generation: u64 = 9; // pre_write_gen(8) + 1
+        let pre_write_gen = 8;
+        unsafe {
+            seqlock_end_write(&node_id(), "test", &mut generation, pre_write_gen, false);
+        }
+        assert_eq!(generation, pre_write_gen);
+        assert_eq!(
+            generation % 2,
+            0,
+            "rolled-back generation must be even (complete)"
+        );
+    }
+
+    /// The very first write of a pool starts `pre_write_gen == 0`; a
+    /// failure must roll back to 0, not underflow.
+    #[test]
+    fn failure_on_first_write_rolls_back_to_zero() {
+        let mut generation: u64 = 1;
+        unsafe {
+            seqlock_end_write(&node_id(), "test", &mut generation, 0, false);
+        }
+        assert_eq!(generation, 0);
+    }
+}
+
 /// The custom node API lets you integrate `dora` into your application.
 /// It allows you to retrieve input and send output in any fashion you want.
 ///
@@ -1319,11 +1408,31 @@ impl Node {
 
         // Copy tensor data to shmem
         if is_cuda {
-            if let Ok(helpers) = get_cuda_helpers(py) {
-                let bound = helpers.bind(py);
-                let _ = bound.call_method1(
-                    "_cuda_memcpy",
-                    (shmem_ptr as u64 + data_offset as u64, ptr_val, size, 2u32),
+            let copy_ok = match get_cuda_helpers(py) {
+                Ok(helpers) => {
+                    let bound = helpers.bind(py);
+                    bound
+                        .call_method1(
+                            "_cuda_memcpy",
+                            (shmem_ptr as u64 + data_offset as u64, ptr_val, size, 2u32),
+                        )
+                        .is_ok()
+                }
+                Err(_) => false,
+            };
+            if !copy_ok {
+                // This is the pool's first (registering) write — there is no
+                // previous valid frame to fall back to, so unlike
+                // `write_memory_pool` the only safe option is to abort the
+                // registration entirely rather than publish a pool whose
+                // backing memory was never actually written (see #2436).
+                // Restore ownership so `shmem`'s `Drop` unlinks the segment
+                // instead of leaking it.
+                shmem.set_owner(true);
+                eyre::bail!(
+                    "[{}] memory-pool registration: initial GPU copy failed; \
+                     aborting before the pool is published to any consumer",
+                    self.node_id
                 );
             }
         } else {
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed the diff against the code at head (`3214670`). No issues found — the core fix is sound:

- Seqlock arithmetic is correct: `seqlock_begin_write` reads the even `pre_write_gen` and writes `pre+1` (odd = writing); `seqlock_end_write` writes `pre+2` on success and rolls back to `pre` on failure. This matches the old advance-by-2 on success and correctly leaves the previous even generation on failure. The first-write (`pre=0`) no-underflow case is covered.
- All 4 fallible copy sites (fast/slow × `dma_copy`/`_cuda_memcpy`) thread `pre_write_gen` through and pass a real `copy_ok`; the infallible CPU-copy sites correctly keep the unconditional advance. No fallible write site was missed.
- The `register_memory_pool` abort path calls `_unregister_host` before restoring `set_owner(true)`, mirroring `free_memory_pool`'s ordering, and aborts before insertion into `PINNED_POOL` / daemon registration — no partially-published pool, no pinned-registration leak.

One non-blocking design note (not a defect): on a failed GPU copy, `write_memory_pool` rolls the seqlock back but still returns `Ok(())` and only logs at `error` level. The consumer then reads the last-known-good frame believing it is fresh, and the producer gets no signal that a frame was dropped. This is strictly safer than the old "publish torn/uninitialized as complete" behavior, so it's a reasonable choice — but if silent frame drops matter for a given pipeline, you may eventually want to surface the failure to the caller rather than only logging it.

---

phil-opp:
@haixuanTao I'm not familiar with this code, maybe you could take a look? Thanks!

haixuanTao:
> @haixuanTao I'm not familiar with this code, maybe you could take a look? Thanks!

I'm not familiar either, and it feels like the wrong place to put unsafe logic in, no?

phil-opp:

The seqlock generation arithmetic is correct, and the register-abort cleanup order (`_unregister_host` → `set_owner(true)` → `bail!`, before the `PINNED_POOL` insert and daemon registration) is right — no partially-published pool and no leaked pinned registration on that path.

One correctness caveat that the "on failure, readers keep observing the previous, still-valid frame" framing glosses over: on failure `seqlock_end_write` restores the generation to its **exact** pre-write value, which only preserves the previous frame if the failed copy left the destination bytes untouched. Both fallible copies write **in place** into the previous frame's storage (`_cuda_memcpy` → `shmem_ptr + data_offset`; `dma_copy` → the reused pooled GPU slot). If such a copy fails *after* partially writing (GPU fault / ECC / mid-transfer page fault), the previous frame is already clobbered, yet the generation is rolled back to look unchanged — so a consumer (including one whose read straddles the write window) observes the same even generation before and after and accepts the corrupted bytes as a valid, unchanged frame, with no torn-read retry.

This is still a clear improvement over the old unconditional advance for the common case (copy skipped, or failed before touching the buffer). But two things would make the guarantee actually hold: (1) scope the doc-comment wording to "failed before writing," and (2) for a genuine mid-transfer failure, surface the error to the caller rather than only logging at `error!` — a partial in-place write can't be hidden behind a generation rollback. Also worth resolving @haixuanTao's open question about whether this seqlock/unsafe logic belongs inline here vs. a dedicated abstraction.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; treat it accordingly. (The two items raised in the earlier review — the partial in-place-write masking on GPU-copy failure, and `write_memory_pool` still returning `Ok(())` on a failed copy — are not repeated here; they still stand.)

One new finding on the seqlock itself, which is worth addressing since seqlock correctness is the whole point of this PR:

**The release fence in `seqlock_end_write` is on the wrong side of the generation store** (`apis/python/node/src/lib.rs:538-550`):

```rust
if copy_ok {
    std::ptr::write_volatile(gen_ptr, pre_write_gen.wrapping_add(2)); // publish even
} else {
    ...
    std::ptr::write_volatile(gen_ptr, pre_write_gen);
}
std::sync::atomic::fence(std::sync::atomic::Ordering::Release);       // <-- after the store
```

A `Release` fence orders preceding memory operations before *subsequent stores*. Here there is no store after the fence, so it does nothing to guarantee that the frame data writes (the `copy_nonoverlapping` / CUDA copies that ran before this call) are visible **before** the even generation is published. The correct unlock is the mirror of `seqlock_begin_write` (which correctly does store-odd → `fence(Release)`): the end side should be `fence(Release)` **then** `write_volatile(even)`.

Failure scenario: on a weakly-ordered architecture (aarch64 — Jetson/ARM, a first-class dora target), the store publishing the even generation can be reordered ahead of the frame-data stores. A consumer polling the seqlock then reads an even, "complete" generation while the payload bytes haven't landed yet, and accepts a torn/stale frame as valid — the exact class of bug this PR is trying to close, just on the publish edge instead of the copy-failure edge. This is masked on x86 (TSO preserves store-store order) but real on ARM. The CPU-only paths (lines ~1791, ~1936) advance the generation the same store-then-fence way and have the same issue.

(Also note the added unit tests can't catch this — they're single-threaded with no reader and `seqlock_end_write` never reads `gen_ptr`, so they only assert the `pre+2` / `pre` arithmetic, not ordering.)

---

phil-opp:
> I'm not familiar either, and it feels like the wrong place to put unsafe logic in, no?

Sorry for the wrong ping! Yeah, it feels a bit weird to have this Python-only API that does a lot of unsafe CUDA operations.

The code was actually added by @tang-canran in https://github.com/dora-rs/dora/pull/2168 . Maybe @tang-canran or @heyong4725 could do a review of this PR? Thanks!

tang-canran (commented):
**Overall: Approved.** Both fixes are correct and well-structured. Thanks for extracting the seqlock helpers — this makes all the memory-pool write paths auditable in one place instead of scattered inline unsafe blocks.

**Fix 1 (seqlock rollback):** Correctly addresses #2436. If `cudaMemcpy` / `dma_copy` fails, rolling back the generation to the previous good frame is exactly the right behavior — consumers see a stale-but-valid frame instead of a "complete" generation over unwritten garbage. The three unit tests cover the critical cases (success advance, failure rollback, first-write-to-zero).

**Fix 2 (register cleanup):** Correct. `cudaHostRegister` must be unregistered before unmapping, and `set_owner(true)` prevents shmem segment leaks on registration abort.

**One non-blocking suggestion:** The CPU copy paths (fast-path `else` branch and slow-path `else` branch) now call `seqlock_begin_write()` but still use the inline `old_gen + 1` pattern for the end-write, instead of `seqlock_end_write(..., copy_ok=true)`. The CPU `copy_nonoverlapping` is indeed infallible, so this is not a correctness issue — but using the same helper pair on all four paths would make future seqlock audits simpler. Purely a style nit.

**Broader observation (cc @phil-opp):** This PR actually demonstrates *why* the unsafe CUDA operations in the Python node API need careful review — the original `let _ =` error-swallowing in cudaMemcpy calls masked failures that then caused silent data corruption via seqlock. Extracting `seqlock_begin_write` / `seqlock_end_write` was the right direction. A follow-up could apply the same `.is_ok()` error propagation to the inline Python helpers (`dma_copy`, `_cuda_memcpy`), but that's beyond the scope of this PR.

LGTM.

Issue #2157, opened 2026-06-12
chore: upgrade PyO3 to 0.29 (RUSTSEC-2026-0176)


heyong4725:
RustSec published [RUSTSEC-2026-0176](https://rustsec.org/advisories/RUSTSEC-2026-0176) on 2026-06-11: out-of-bounds read in `nth`/`nth_back` on `PyList`/`PyTuple` iterators in pyo3 <= 0.28.x. Fix: upgrade to pyo3 >= 0.29.

**Exposure in dora: none found.** The affected iterator fast-path methods are not called anywhere in the workspace (the only `nth` is on a str-split iterator in `libraries/extensions/download`). pyo3 reaches us via dora-node-api-python, dora-operator-api-python, dora-ros2-bridge-python, pythonize, and dora-runtime.

A temporary audit waiver is being added (deny.toml + .cargo/audit.toml) so PR CI stays green; this issue tracks the real fix:

- [ ] Bump workspace pyo3 0.28 → 0.29 (check pythonize 0.29 availability and PyO3 migration notes — minor versions carry API breaks)
- [ ] Update CLAUDE.md ("Python packages use PyO3 0.28")
- [ ] Remove the RUSTSEC-2026-0176 entries from deny.toml and .cargo/audit.toml

Waiver review date: 2026-07 (or earlier when pythonize catches up).

phil-opp:
Attempted the upgrade locally. Two blockers found:

**1. `pyo3-build-config` 0.29 dropped the `abi3-py311` feature.**
`pyo3-build-config` 0.29.0 no longer exposes `abi3-py311` as a feature (available features: `default`, `extension-module`, `generate-import-lib`, `resolve-config`). The workspace `Cargo.toml` currently specifies `features = ["resolve-config", "abi3-py311"]` for that crate, which would need to be updated to just `["resolve-config"]`.

**2. `cargo update` hits a dependency resolution conflict in the full workspace.**
After fixing blocker 1, cargo fails to resolve `arrow = "^58"` — it reports a `links = "python"` conflict between two packages in the resolution graph. This appears to be triggered by pyo3 0.29's changed native-library linking, interacting with how the workspace's Python packages and the C++ node API are wired together.

[...]

PR #2435, opened 2026-06-30
docs(operator): document Event enum and DoraOperator trait with a compile-checked doctest

phil-opp:
## Summary

The two central public items of the operator API — the `Event` enum that operators pattern-match on and the `DoraOperator` trait they implement — had **no doc comments**. New users implementing an operator had to infer:

- the meaning of each event variant (e.g. `InputParseError` vs `InputClosed`),
- when `Stop` fires,
- the contract of `on_event`'s return value (`Continue`/`Stop`/`Err`),
- that the enum is `#[non_exhaustive]` and therefore needs a catch-all arm.

This PR documents the enum, each variant, the trait, and its method, and adds a **compile-checked doctest** showing a minimal operator implementation (with a catch-all arm and an `output_sender.send(...)` call via `IntoArrow`).

## Issue

`apis/rust/operator/src/lib.rs` is the recommended entry point for writing dora operators (the crate-level docs even say "It is the recommended way of using `dora`"), yet its two key public types carried zero item-level documentation.

## Fix

- Doc comments on `Event` and each of its fields/variants.
- Doc comments on `DoraOperator` and `on_event`, including the return-value contract.
- A `///` doctest demonstrating a complete `impl DoraOperator`.

No behavior change — documentation only.

## Validation

Run against the affected crate:

- `cargo test -p dora-operator-api --doc` → the new doctest (`DoraOperator (line 82)`) passes.
- `cargo clippy -p dora-operator-api -- -D warnings` → clean.
- `cargo fmt --all -- --check` → clean.

---

⚠️ **Disclaimer:** This pull request was generated by Claude (an AI agent) and is **machine-generated**. It has been verified locally (fmt, clippy, doctest) but should be reviewed by a human before merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01MykPCxNT1WK3ekBaVyM6ig)_

PR #2445, opened 2026-07-02
chore(deps): upgrade workspace dependencies to latest versions

phil-opp:
Upgrades all workspace dependencies to their latest versions and migrates the code to the new APIs: arrow 58→59, zenoh 1.8→1.9, mavlink 0.13→0.18 (dialects moved to `mavlink::dialects::common`; feature names renamed), nom 7→8 (msg-gen parser rewritten to `.parse(input)` style, with a scoped workaround for the upstream `recognize`-over-`*0`-parsers bug in nom 8.0.0, rust-bakery/nom#1808), syn 1→2, process-wrap 8→9, notify 5→8, duration-str 0.5→0.21, self_update 0.42→0.44 (HTTP backend is now an explicit `reqwest` feature), reqwest 0.12→0.13, sha2 0.10→0.11, sysinfo 0.36→0.38, criterion 0.5→0.8, rand 0.8→0.10, aligned-vec 0.5→0.6 (bincode wire format verified byte-identical), inquire, mavlink, toml 1.0, which 8, webbrowser 1.x, petname 3, tower-http 0.7, heck 0.5, quick-xml 0.41, getrandom 0.4, libloading 0.9, shlex 2, windows 0.62, and assorted minor bumps.

Deliberately **not** upgraded (each blocked; tracking issues linked):

- **uhlc 0.5 → 0.9** (#2446): verified empirically that 0.9 serializes the HLC `ID` byte-reversed vs 0.5 under bincode — `Timestamp` rides in every daemon↔node message, so this is a wire-format break; needs a coordinated protocol bump.
- **bincode 1.3/2.0 → 3.0** (#2447): bincode 3.0.0 is a `compile_error!` protest stub on crates.io, not a real release. 1.3.3/2.0.1 remain the latest usable versions.
- **pyo3/pythonize/pyo3-build-config 0.28 → 0.29** (#2157): `arrow-pyarrow` 59 requires pyo3 ^0.28 and only one `pyo3-ffi` may link libpython. pyo3 advisories RUSTSEC-2026-0176/0177 stay waived (APIs unused).
- **opentelemetry 0.31 → 0.32** (+ tracing-opentelemetry 0.33) (#2448): blocked on an `opentelemetry-system-metrics` 0.32 release (upstream bump prepared separately).
- **redb 2 → 4** (#2449): all redb 3.x/4.x require Rust 1.89 > MSRV 1.88.
- **sysinfo 0.39** (#2449): requires Rust 1.95 > MSRV; 0.38.4 is the newest MSRV-compatible.
- **ros2-client 0.8.2/0.9 + rustdds 0.11.8** (#2450): still blocked by the `rustdds = 0.11.4` pin (Atostek/RustDDS#375 remains open).
- **prettyplease (c++ node API)** (#2452): intentionally kept at 0.1 — it exists only to switch on the `verbatim` feature of rust-format's internal prettyplease 0.1 via version unification; 0.2 would be inert.

Fixes found while migrating: mavlink 0.18's UDP `recv()` surfaces I/O errors that 0.13 swallowed internally, so `classify_recv_error` gained a datagram mode — otherwise an ICMP port-unreachable from a restarting peer, or even a single zero-length datagram (`UnexpectedEof`), would kill the bridge node. New tests pin the datagram policy and the CLI duration grammar.

Behavior changes to be aware of: reqwest 0.13 validates TLS against the OS trust store (platform verifier) instead of bundled Mozilla roots — `dora build` downloads and `dora self-update` now need `ca-certificates` on minimal images. Supply-chain audit updated: quick-xml advisories fixed for msg-gen; new waivers (with review dates) for lz4_flex-via-zenoh, quick-xml-via-self_update/mavlink-bindgen, and rustls-pemfile-unmaintained-via-zenoh — removal tracked in #2451. Known cost: zenoh 1.9 still pins flume 0.11/toml 0.9/reqwest 0.12 internally, so the tree temporarily builds two majors of those until zenoh catches up; hoisting the remaining per-crate dep versions into `[workspace.dependencies]` is left as a follow-up.

Verified: `cargo fmt --check`, `clippy --all -D warnings`, full workspace test suite (115 suites / 1,287 tests), `cargo check --examples`, cargo-audit + cargo-deny green.


phil-opp:

The dependency migrations look carefully done and well-tested, and the deliberately-held-back crates (uhlc / bincode / pyo3) have sound wire-format / linkage justifications. I didn't spot a correctness bug in the call-site migrations (rand 0.10 `Rng::fill_bytes`, syn 2, process-wrap 9, getrandom `fill()`, sha2 0.11 hex are all migrated correctly). A few things worth confirming before merge:

1. **reqwest `rustls-tls` → `rustls` in the download extension changes the TLS trust roots.** reqwest 0.13's `rustls` feature uses `rustls-platform-verifier` (the system/native trust store) instead of the bundled webpki-roots that 0.12's `rustls-tls` provided — the `Cargo.lock` corroborates this (`webpki-roots` dropped from `hyper-rustls`, `aws-lc-rs` pulled in). `dora-download` fetches over HTTPS during `dora build`, so on minimal images without a system CA store (scratch/distroless, some embedded targets) downloads that previously succeeded may now fail cert verification. The PR body notes the `ca-certificates` requirement, so this is intentional — just flagging that it's a real regression surface for trimmed deployment images. Is relying on the platform store acceptable for all supported build environments, or is it worth keeping bundled roots (`rustls` + explicit `webpki-roots`)?

2. **mavlink UDP `datagram` classification** makes every UDP recv IO error `Transient`, so a UDP bridge node has no remaining path to a fatal error / daemon restart — a genuinely wedged socket would warn-loop indefinitely rather than surface. This matches mavlink 0.13's old internal-swallow behavior and defends against the single-packet (ICMP unreachable / zero-length) kill switch, so it's defensible; noting only that "all IO errors transient" removes the restart path entirely.

3. **nom 8 `message_name` rewrite** (msg-gen `ident.rs`): the cited `recognize` / `split_at_position_complete` EOF quirk, if it's general rather than specific to that combinator shape, would also affect the *unchanged* `constant_def` value parser (`recognize(separated_list1(space1, is_not(" \t")))` in `constant.rs`), which could silently parse constant values as empty. Presumably the msg-gen constant/value tests are green under nom 8 — worth a quick confirm that they actually exercise a constant with a value.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; treat it accordingly. I focused on the actual code migrations rather than the version numbers, and independently built/ran the two highest-risk crates (the nom 7→8 msg-gen parser: 70 tests pass, and the `constant_def`/`member_name` parsers are unaffected by the nom 8 `recognize` quirk; the mavlink 0.18 `classify_recv_error(datagram)` node tests + TCP/UDP loopback pass). The rand 0.10 / getrandom 0.4 / sha2 0.11 / aligned-vec 0.6 / process-wrap 9 / duration-str 0.21 migrations all look semantically equivalent — no incorrect API migration found. (The reqwest OS-trust-store and UDP-no-fatal-path items already raised still stand; not repeating them.)

One new finding — **unrelated public-API documentation regression bundled into this deps PR**:

`apis/rust/operator/src/lib.rs` drops ~68 lines that remove **all** the rustdoc on the public `Event` enum (and its variants) and the `DoraOperator` trait, **including the compile-checked `# Example` doctest**. This is exactly the doc block just added by `#2435` (commit `25bac6b`, `docs(operator): document Event enum and DoraOperator trait with a compile-checked doctest`) — which is this PR's own base — so the change effectively reverts a freshly-merged documentation commit. I verified at the PR head (`b0c2d80`): the `Event` enum (line 34) and `DoraOperator` trait (line 50) now carry no rustdoc and the `ExampleOperator` doctest is gone.

This is out of scope for a `chore(deps)` PR and I don't see why it's necessary. If the doctest broke under arrow 59 / the `IntoArrow` surface, the fix is to update the example, not delete the docs. Recommend restoring the `#2435` doc block (adapted to the new arrow API if needed) rather than dropping it here.

Process note: the `Test`, `E2E`, `Semantic contract`, and `Benchmark regression` CI jobs show `skipped` (they're merge-queue-gated), so the workspace test suite hasn't actually run in this PR's CI. Given this touches wire serialization (aligned-vec in `DataMessage::Vec`) and the daemon spawn path (process-wrap), worth confirming they go green in the merge queue before merging.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop; please verify before acting.

Re-reviewing the current diff (`b0c2d80`, unchanged since the earlier automated reviews) — two corrections/confirmations to the previous automated comments, no new issues in the code itself:

- **Retracting the earlier "doc regression in `apis/rust/operator/src/lib.rs`" finding — it's a false positive.** That file is not part of this PR. The only operator-crate files touched are `Cargo.toml` bumps (`apis/rust/operator/macros`, `apis/rust/operator/types`, plus the c++/python operator `Cargo.toml`s); the `syn 1→2` change only touches `macros/Cargo.toml`. The `#2435` `Event`/`DoraOperator` doctest in `operator/src/lib.rs` is untouched, so this should not gate the merge.
- **The suspected `nom 8` `recognize` breakage in the (unchanged) msg-gen constant/member value parser does not reproduce.** The `is_not`-based value parser reaches EOF correctly (`constant_def("int32 AAA=30")` → `value == ["30"]`); the `nom 8` `split_at_position` quirk only affects the `alphanumeric0` path in `message_name`, which is exactly and only where the workaround was applied. msg-gen tests pass.

Otherwise no important issues found. The two intentional behavior changes remain worth a maintainer's awareness (both test-covered): reqwest 0.13's rustls feature moves TLS to the OS trust store (minimal images now need `ca-certificates` for `dora build`/self-update), and mavlink 0.18 UDP recv errors are now all classified `Transient` (removes the fatal/restart path for a wedged UDP socket).

---

phil-opp:
**Resolving the open items from the automated reviews above:**

**1. reqwest TLS trust roots (platform verifier vs bundled webpki-roots)** — keeping the platform verifier; the `ca-certificates` requirement is now documented in-repo (`docs/cli.md`, under `dora build` → URL sources and `dora self update`), not just in the PR body. Bundled roots stopped being a supported configuration in reqwest 0.13 — the 0.13.0 changelog: *"rustls roots features removed, `rustls-platform-verifier` is used by default. To use different roots, call `tls_certs_only(your_roots)`"* — so keeping webpki-roots would mean hand-rolling the rustls config in `dora-download`, and it still wouldn't cover `dora self-update`: `self_update` 0.44 resolves to reqwest 0.13.4 and builds its own client through the same `rustls` feature (verified via `cargo tree`; its optional `ureq` backend is not activated, and the only reqwest 0.12 left in the graph is opentelemetry's plain-HTTP OTLP exporter). The OS store is the one configuration that covers both HTTPS surfaces consistently, and it honors enterprise root CAs, which bundled Mozilla roots never did.

**2. mavlink UDP recv errors all `Transient`** — intended, keeping as is. A datagram socket has no connection that can die, so no `ErrorKind` reliably means "socket wedged": `ECONNREFUSED` (ICMP unreachable from a restarting peer) and `UnexpectedEof` (zero-length datagram) are routine recoverable events, and promoting any of them to fatal reintroduces the single-packet kill switch this arm exists to prevent. It is exact behavior parity with mavlink 0.13's internal retry loop, pinned by the `datagram_io_errors_are_transient` test, and a persistently erroring socket still surfaces through the rate-limited warn (one per 5 s). If a real wedged-socket mode shows up in practice, an escalation policy (N consecutive errors with zero successful recvs → fatal) can be added then — adding one speculatively risks killing bridges during normal GCS/SITL restarts.

**3. nom 8 `constant_def` quirk / operator-docs regression** — both already closed by the third review pass (value parser verified unaffected; the doc-regression finding was retracted as a false positive). No action.

Found while re-running the audit gate: **RUSTSEC-2026-0204** (crossbeam-epoch 0.9.18, invalid pointer deref in `fmt::Pointer`, published 2026-07-06 — the day after the last review) turns `cargo audit` red, which would fail this PR's supply-chain CI job. Bumped to 0.9.20 in the lockfile on this branch.

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; treat accordingly.

Re-reviewed the three commits added since the previous automated pass (07-03). All three are low-risk and don't change the earlier assessment — I confirmed each commit's file scope:

- `c36f6dc` — crossbeam-epoch 0.9.18→0.9.20 (RUSTSEC-2026-0204), **`Cargo.lock` only**; clears the cargo-audit gate.
- `617fdc0` — **`docs/cli.md` only**, documenting the `ca-certificates` requirement for the reqwest 0.13 OS-trust-store change.
- `a237e41` — **`_typos.toml` only**, excluding local `worktrees/` from the typos gate.

No new correctness issues. The two intentional behavior changes flagged in the earlier passes (reqwest moving to the platform trust store; mavlink UDP recv errors now all classified `Transient`) remain the only maintainer-awareness items, and both are documented and test-covered. Worth confirming the merge-queue Test / E2E / contract jobs go green before merging, since the trunk status currently shows a failing semantic-contract check.

---

phil-opp:

Reviewed the one new non-merge commit since the last automated pass (`23935c6`, "drop stale aligned-vec 0.5 dev-dependency"). It looks correct and safe: `apis/rust/node`'s `[dependencies]` was bumped to aligned-vec 0.6.4 but the `[dev-dependencies]` duplicate was left at 0.5.0, so both semver-incompatible versions resolved into `Cargo.lock` and the doctest rustdoc invocation got `--extern aligned_vec=` twice (E0464). Removing the redundant dev-dep entry dedupes the lockfile (0.5.0 gone; all crates now reference a single `aligned-vec`), and bench targets still link the crate's `[dependencies]`, so `benches/arrow_framing.rs` keeps compiling. No behavior change to shipped code.

The earlier maintainer-awareness items (reqwest 0.13 moving to the OS trust store; mavlink 0.18 UDP recv errors all classified `Transient`) were addressed and accepted in the PR discussion, and the earlier doc-regression / nom-8 findings were already retracted as false positives. No outstanding correctness issues.

---
🤖 *This is a fully automated review by Claude. It may contain mistakes — please verify before acting.*

---

phil-opp:
@claude What is this CI error about: https://github.com/dora-rs/dora/actions/runs/28983542760/job/86013900400?pr=2445

phil-opp:
> 🤖 **Automated review by Claude** — this is a fully automated review (no human in the loop). Findings are advisory; please verify before acting.

Re-reviewed the three non-merge commits added since the last automated pass (`23935c6`):

- **`072b5f0` (reset `StreamDecoder` after each batch decode)** — the fix is correct and directly targets the schema-once/Arrow-59 regression the merge queue was hitting. On both the success and error paths, `decode_batch` now discards the primed `StreamDecoder` and clears `schema_hash`, so the next schema-less batch re-primes from the retained schema bytes via `prime()` (a fresh decoder each time) rather than feeding a decoder that Arrow 59 has already driven to its terminal `Finished` state. Re-priming is a local, in-memory operation (retained-schema lookup, no network round-trip), and the schema set is still bounded — so the only cost is losing per-batch decoder reuse, which is an acceptable trade for correctness and is documented in the comments. The added `input_decoder_handles_sequential_batches_arrow_59_terminal_state` test covers the multi-batch reuse case. The `node/mod.rs` and `event_stream/mod.rs` changes are `tracing::debug!` logging plus a `#[derive(Debug)]`, no behavior change.
- **`233d8f7` / `5fe013e` (fault-tolerance test timeout/restart-window bounds)** — test-only reliability tweaks, low risk.

The `prepared.rs` clone-then-`.env()`-then-reassign in `072b5f0` is just the process-wrap builder-API adaptation of the old `set_env` call; semantically equivalent.

No new correctness issues. The two intentional behavior changes flagged earlier (reqwest 0.13 OS trust store; mavlink 0.18 UDP recv errors all `Transient`) were already accepted in the discussion.

---

phil-opp:

Re-reviewed at current head `f60cce81`. Changes since the last automated pass (`072b5f0`): a forward-merge of `main`, a CI-only commit (`55a6824`, adds the "free disk space" step to the parallel merge-queue e2e/contract/bench jobs — workflow YAML, no code), and a re-squash of the upgrade commit onto the new base. Checking the PR-authored delta against current `main`, the merged production files (`doctor.rs`, `validate.rs`, `recording`, `hub-client`) are identical to `main` (inherited, not introduced here); the only genuine new code vs `main` is two `tracing::debug!` lines in `decode_zenoh_sample` and the process-wrap 9 `TokioChildWrapper`→`ChildWrapper` trait rename in `running_dataflow.rs`.

No new correctness issues. The earlier schema-once/Arrow-59 regression remains fixed by `072b5f0`, and the two previously-accepted behavior changes (reqwest 0.13 OS trust store; mavlink 0.18 UDP recv errors all `Transient`) still stand. Latest commits look safe to merge.

---

PR #2454, opened 2026-07-02
fix(download): derive fallback filename from URL path, not raw URL string

phil-opp:
## Summary

When a download has no `Content-Disposition` header, `dora-download` derived the filename with `Path::new(response.url().as_str()).file_name()` — a filesystem-path parser run over the **entire URL string**. A query string is not a path separator, so for presigned / CDN URLs (very common for model and operator downloads) like:

```
https://host/model.bin?X-Amz-Signature=abc123&x=1
```

the saved file was named `model.bin?X-Amz-Signature=abc123&x=1`.

## Fix

Derive the name from the parsed URL's last path segment (`filename_from_url`), which drops the query string and fragment. Added unit tests for plain, query-string, fragment, and trailing-slash URLs.

### Behavior change to be aware of

For a URL whose path ends in `/` (e.g. `https://host/path/`) or a bare host (`https://host/`), the old code guessed a name from the trailing/authority component (`path` / `host`); the new code returns `None`, so `download_file` errors with `Could not find a filename` instead of saving to a misleadingly-named, extension-less file. This is intentional and covered by a test — a directory/host URL has no real filename, and a clear error is better than a silent wrong name. (When `Content-Disposition` is present, which is typical for artifact downloads, this fallback isn't used at all.)

## Validation

- `cargo test -p dora-download` (4 new tests) ✅
- `cargo clippy -p dora-download -- -D warnings` ✅

---

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -5,6 +5,21 @@ use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
 use tokio::io::AsyncWriteExt;
 
+/// Derive a fallback filename from a URL's last path segment.
+///
+/// Uses the parsed path segments rather than the raw URL string: a query
+/// string is not a path separator, so running `Path::file_name` over the full
+/// URL would fold it into the name (e.g. a presigned/CDN URL like
+/// `https://host/model.bin?X-Amz-Signature=...` would yield
+/// `model.bin?X-Amz-Signature=...`). Returns `None` for URLs whose last
+/// segment is empty (e.g. a trailing slash).
+fn filename_from_url(url: &reqwest::Url) -> Option<String> {
+    url.path_segments()
+        .and_then(|mut segments| segments.next_back())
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** — verified the fix at `2f3e04d` against the URL-parsing edge cases.

No issues found. Query strings and fragments are correctly excluded (they aren't part of `path_segments()`), and path traversal isn't reachable: the `url` crate normalizes `.`/`..` during parse and percent-encoded separators stay within a single segment, so `next_back()` can't return a real separator; the downstream `Path::file_name()` step flattens anyway. Trailing-slash / bare-host / empty-path URLs correctly yield `None`, surfacing a clear "Could not find a filename" error instead of a misleadingly-named file. The 4 tests cover the meaningful cases.

---

phil-opp:
Superseded — the bug is real and still live on main, but this branch predates the #2485 `get_filename` rewrite and no longer rebases. Re-applied against current main (with the fix extracted into a tested `filename_from_url` helper) together with #2465, since both touch the same function and would otherwise conflict with each other and with #2610.

PR #2455, opened 2026-07-02
refactor(daemon): dedup all-inputs-closed tail of close_input/break_input

phil-opp:
## Summary

`close_input` and `break_input` in `binaries/daemon/src/lib.rs` each contained a byte-for-byte identical ~15-line block: recompute whether any input is still circuit-broken and, if the node has no remaining open inputs, disable its restart policy, send `AllInputsClosed`, bump the pending counter, and record `all_inputs_closed_at`. Two copies had to be kept in sync manually.

## Fix

Extract the shared block into `signal_all_inputs_closed_if_drained(dataflow, receiver_id, clock)`.

Behavior is preserved exactly:
- The helper opens with `let Some(channel) = dataflow.subscribe_channels.get(...) else { return; }`, so `disable_restart` + `AllInputsClosed` still only run when the node has a subscribe channel (the drain logic previously lived inside the `if let Some(channel)` guard in both callers).
- `close_input` keeps its `was_open` guard on the `InputClosed` send; `break_input` still has none — unchanged.
- `has_broken` timing is unchanged (`break_input`'s input is moved into `broken_inputs` before the call, so the helper correctly suppresses `AllInputsClosed`).

## Validation

- `cargo test -p dora-daemon` — 89 tests pass, incl. the circuit-breaker / close-input cases ✅
- `cargo clippy -p dora-daemon -- -D warnings` ✅

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -5516,35 +5516,50 @@ fn close_input(
         return;
     }
 
-    if let Some(channel) = dataflow.subscribe_channels.get(receiver_id) {
-        if was_open
-            && send_with_timestamp(
-                channel,
-                NodeEvent::InputClosed {
-                    id: input_id.clone(),
-                },
-                clock,
-            )
-            .ok()
-                == Some(true)
-        {
-            dataflow.inc_pending(receiver_id);
-        }
+    if let Some(channel) = dataflow.subscribe_channels.get(receiver_id)
+        && was_open
+        && send_with_timestamp(
+            channel,
+            NodeEvent::InputClosed {
+                id: input_id.clone(),
+            },
+            clock,
+        )
+        .ok()
+            == Some(true)
+    {
+        dataflow.inc_pending(receiver_id);
+    }
 
-        let has_broken = dataflow
-            .broken_inputs
-            .keys()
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** — verified this refactor at `1d4644d` preserves the behavior of both original branches.

No issues found.

- `close_input` keeps its `was_open` guard on the `InputClosed` send; `break_input` remains unguarded — both correctly retained.
- The extracted `signal_all_inputs_closed_if_drained` runs unconditionally in both callers, matching the original tails, which ran independent of `was_open` (relevant on `close_input`'s `was_broken`-only path).
- `has_broken` is evaluated at the same logical point: `close_input` removes the input from `broken_inputs` before the call, while `break_input`'s input is already in `broken_inputs`, so `AllInputsClosed` stays suppressed as before.
- Re-fetching the channel inside the helper is safe — `subscribe_channels` isn't mutated between the caller's send and the helper, and `None` short-circuits identically to the old `if let Some(channel)` wrapper.
- Operation order and success-gating (`InputClosed` → `disable_restart` → `AllInputsClosed` → `inc_pending`/`all_inputs_closed_at`) are unchanged.

---

PR #2456, opened 2026-07-02
fix(coordinator-store): reject null-byte param keys in shared validator

phil-opp:
## Summary

The two `CoordinatorStore` backends disagreed on which param keys they accept:

- `RedbStore::put_node_param` builds a composite string key using `\0` as a separator (`KEY_SEPARATOR`) and rejects any key containing `\0` via `make_param_key`/`check_no_separator`.
- `InMemoryStore::put_node_param` stores the key in a tuple `BTreeMap` and only checked length.

So the same `put_node_param(..., "a\0b", ...)` call returned `Err` on the redb backend but `Ok` on the in-memory backend — a latent inconsistency that would surface if the coordinator switched backends.

## Fix

Move the null-byte rejection into the shared `validate_param_limits`, which both backends already call, so they now accept/reject identical input. The redb-side `check_no_separator` stays as defense in depth. Added an in-memory regression test mirroring the existing redb one (`param_key_with_null_byte_is_rejected`).

## Validation

- `cargo test -p dora-coordinator-store` and `... --features redb-backend` — both new tests pass, all 34 redb tests pass ✅
- `cargo clippy -p dora-coordinator-store --features redb-backend -- -D warnings` ✅

---

---

### DIFF
--- a/libraries/coordinator-store/src/in_memory.rs
+++ b/libraries/coordinator-store/src/in_memory.rs
@@ -340,6 +340,17 @@ mod tests {
         assert_eq!(store.list_node_params(&df, &node).unwrap().len(), 1);
     }
 
+    /// A `\0`-containing param key must be rejected here just as the redb
+    /// backend rejects it (it reserves `\0` as a composite-key separator), so
+    /// the two backends agree on which input is valid.
+    #[test]
+    fn param_key_with_null_byte_is_rejected() {
+        let store = InMemoryStore::new();
+        let df = Uuid::new_v4();
+        let node: NodeId = "sensor".to_string().into();
+        assert!(store.put_node_param(&df, &node, "a\0b", b"v").is_err());
+    }
+
     /// `delete_dataflow` must remove every node param tied to that
     /// dataflow so `dora clean` doesn't leak orphan rows. Params for
     /// other dataflows must survive untouched.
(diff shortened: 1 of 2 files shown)


phil-opp:
**Automated review by Claude** — verified the diff at `f13ee1e` against both store backends.

No blocking issues. `validate_param_limits` is the first call in both backends' `put_node_param`, so the new `\0` rejection covers every write path; keeping redb's `check_no_separator` as defense-in-depth is right, and limiting the rejection to `\0` (the redb `KEY_SEPARATOR`) rather than all control chars is correct.

One minor, non-blocking note: the read/delete paths stay asymmetric — `get_node_param`/`delete_node_param`/`list_node_params` don't call the validator, so a `\0`-containing key returns `Err` on redb (via `make_param_key`) but `Ok(None)`/`Ok(())` in-memory. Harmless now that `put` rejects such keys on both backends (one can never be stored), so the "identical accept/reject" invariant holds for the write path but not strictly for reads/deletes. Fine to leave as-is.

---

phil-opp:

The branch was rebased since my last review, but the diff is unchanged — the shared `validate_param_limits` null-byte rejection plus the in-memory regression test. The commit since my last review doesn't introduce new issues — looks safe to merge from my side.

---

PR #2457, opened 2026-07-02
perf(record-node): serialize arrow payload directly into aligned buffer

phil-opp:
## Summary

The per-message record path in `binaries/record-node/src/main.rs` allocated and copied the arrow payload **twice**:

```rust
let mut buf = vec[0u8; data_size];                 // alloc #1
arrow_utils::copy_array_into_sample(&mut buf, ...); // copy #1
Some(AVec::from_slice(128, &buf))                   // alloc #2 + copy #2
```

`AVec::from_slice` allocates a new 128-byte-aligned buffer and memcpys `buf` into it, so every recorded message paid for two allocations and two payload-sized copies where one of each suffices.

## Fix

Allocate the aligned `AVec` once (zeroed) and serialize the arrow buffers straight into it, matching the idiom already used in the daemon (`lib.rs`) and runtime:

```rust
let mut buf: AVec<u8, ConstAlign<128>> = AVec::__from_elem(128, 0, data_size);
arrow_utils::copy_array_into_sample(&mut buf, &arrow_data);
Some(buf)
```

Same result type as before. The zeroing is retained because `copy_array_into_sample` leaves alignment-padding gaps between buffers unwritten.

## Validation

- `cargo test -p dora-record-node` ✅
- `cargo clippy -p dora-record-node -- -D warnings` ✅

---

---

### DIFF
--- a/binaries/record-node/src/main.rs
+++ b/binaries/record-node/src/main.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, fs::File, time::SystemTime};
 
-use aligned_vec::AVec;
+use aligned_vec::{AVec, ConstAlign};
 use dora_message::{
     common::Timestamped,
     daemon_to_daemon::InterDaemonEvent,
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** — verified the buffer math at `303114a` end-to-end.

No issues found. `data_size = required_data_size(&arrow_data)` yields the exact footprint including inter-buffer alignment padding, and `AVec::<u8, ConstAlign<128>>::__from_elem(128, 0, data_size)` allocates a length-`data_size`, 128-byte-aligned, zeroed buffer that `copy_array_into_sample` fills without overrun or truncation. Alignment and the resulting `Option<AVec<…>>` type are unchanged from the previous `AVec::from_slice(128, &buf)`, and the retained zero-fill still covers the padding gaps the copy leaves unwritten. Removes one alloc + one payload-sized memcpy per recorded message, matching the idiom already used in the daemon/runtime/node paths.

---

phil-opp:
Obsoleted in its current form by #2366 (Arrow IPC data plane). The double-copy this targets does still exist on main — `encode_arrow_ipc` produces a `Vec<u8>` that `AVec::from_slice(128, &ipc_bytes)` then copies again (`binaries/record-node/src/main.rs:81-83`) — but the mechanism here (`copy_array_into_sample` into a pre-aligned `AVec`) no longer applies, since the raw-buffer path it built on is gone. Closing; worth reopening as a fresh PR against `encode_arrow_ipc` if the extra copy still shows up in record-node profiles.

PR #2458, opened 2026-07-02
docs(node): make InputTracker example a compile-checked doctest

phil-opp:
## Summary

`InputTracker` (`apis/rust/node/src/event_stream/input_tracker.rs`) is a public API, but its usage example was fenced ` ```ignore `, so it was never compiled and had drifted from the real API. As written it would not compile: it destructured `Event::Input { id, data, .. }` and then also called `tracker.last_value(&id)` on the moved `id`, and used an undefined `events.recv().await`.

## Fix

Replace it with a compile-checked example (runs under `cargo test --doc`) that exercises the real constructor and query methods (`state`, `any_closed`, `closed_inputs`, `last_value`) and shows the graceful-degradation pattern via a `process_event` helper.

The example references `Event` by type rather than constructing it, because `Event` is `#[non_exhaustive]` and cannot be built outside the crate — the doctest documents the real, compilable usage shape.

## Validation

- `cargo test -p dora-node-api --doc input_tracker` — the doctest compiles and passes ✅

---

---


phil-opp:
**Automated review by Claude** — verified the doctest at `c5820d6` against the head-SHA source.

No issues found. `Event`, `InputState`, and `InputTracker` are all re-exported from `dora_node_api`, and every method used (`new`, `state`, `any_closed`, `process_event`, `closed_inputs`, `last_value`) matches the real signatures, so the doctest compiles. Leaving it un-`no_run` is appropriate — the body is pure and the assertions hold on a fresh tracker, so it runs under `cargo test --doc`. Good catch replacing the old `ignore` block, which had drifted (moved `id` then reused it, referenced an undefined `events.recv()`).

---

PR #2462, opened 2026-07-03
fix(core): apply operator-id prefix when checking single-operator edge types

### LINKED ISSUE #2461: Cross-edge type checking silently skipped for edges sourced from single-operator (`operator:`) nodes

## Summary

`check_type_annotations_full` (the `dora build` / `dora graph --validate` type-annotation checker) never performs cross-edge type-mismatch detection or type inference for any edge whose **source is a single-operator node** (a node declared with the `operator:` block). Multi-operator (`operators:`) source nodes work correctly. The result is that a real type mismatch on such an edge produces no warning, and in `--strict` mode it produces a *misleading* warning instead.

## Root cause: a key-shape mismatch caused by running the check pre-resolution

In `libraries/core/src/descriptor/validate.rs`, operator outputs are registered into `output_type_map` with an **operator-id-prefixed** key:

```rust
// single-operator node (line ~890)
let op_id = op.id.as_ref().map(|id| id.to_string())
    .unwrap_or_else(|| super::SINGLE_OPERATOR_DEFAULT_ID.to_string()); // "op"
...
output_type_map.insert((nid.clone(), format!("{op_id}/{output_id}")), urn.clone());
```

The consumer-side lookup uses the mapping exactly as written in the descriptor (line ~1108):

```rust
let key = (mapping.source.to_string(), mapping.output.to_string());
let upstream_urn = output_type_map.get(&key);
```

`check_type_annotations_full` runs on the **unresolved** descriptor — both callers (`binaries/cli/src/command/validate.rs:153` and `binaries/cli/src/command/build/mod.rs:376`) invoke it *before* `resolve_aliases_and_set_defaults()`. In the unresolved descriptor, a consumer of a single-operator node writes the **bare** output name (`source: A, output: out`). The op-id prefix (`A/op/out`) is only added **later**, during resolution:

```rust
// libraries/core/src/descriptor/mod.rs:105-107
if let Some(op_name) = single_operator_nodes.get(&mapping.source).copied() {
    mapping.output = DataId::from(format!("{op_name}/{}", mapping.output));
}
```

So at check time the producer key is `(A, "op/out")` but the consumer lookup key is `(A, "out")` — they never match, and `upstream_urn` is always `None`. This drops the edge into the "no upstream type" branches:

- **non-strict:** `(None, Some(in_urn))` → `_ => {}` — the mismatch check and the `(Some, None)` inference branch are both skipped.
- **strict:** `(None, Some(in_urn))` → emits *"upstream A/out has no type annotation"* even when the operator **does** declare `output_types` for that port — a false/misleading warning.

The multi-operator (`operators:`) form is unaffected because there the consumer writes the operator id verbatim (`A/opid/out`), so the lookup key already matches the registered key. I confirmed the default op id used on both sides is the same constant (`SINGLE_OPERATOR_DEFAULT_ID = "op"`), so the divergence is purely bare-vs-prefixed, i.e. an ordering bug — not intentional.

## Reproduction (conceptual)

A single-operator node `A` declaring `output_types: { out: std/core/v1/Float32 }`, feeding a consumer with `input_types: { x: std/media/v1/Image }` and `inputs: { x: A/out }`, produces **no** type-mismatch warning from `dora build` / `dora graph --validate`. The equivalent dataflow written with the `operators:` (multi-operator) form does warn.

## Impact

Advisory subsystem only — it does not accept an invalid dataflow or reject a valid one; the dataflow still runs. But the type-annotation feature (mismatch warnings + inference) silently covers only part of the graph, and `--strict` emits an incorrect "no type annotation" message for these edges.

## Suggested fix

Either run the annotation check after `resolve_aliases_and_set_defaults()` (so consumer outputs are already op-prefixed), or normalize the single-operator consumer key at lookup time in `check_edge_mismatches_with_compat` — apply the same `{op_id}/{output}` prefixing the resolver uses when `mapping.source` is a single-operator node before probing `output_type_map`. A regression test covering an `operator:`-sourced edge (both mismatch and inference cases) would lock this in; the existing tests appear to exercise only the `operators:` form.
</body>

phil-opp:
Fixes #2461

## Summary

`check_type_annotations_full` (the `dora build` / `dora graph --validate` type-annotation checker) never performed cross-edge type-mismatch detection or type inference for any edge whose **source is a single-operator node** (a node declared with the `operator:` block). Multi-operator (`operators:`) source nodes worked correctly.

## Root cause

`check_type_annotations_full` runs on the **unresolved** descriptor. Operator outputs are registered into `output_type_map` with an operator-id-prefixed key (`{op_id}/{output}`), but a consumer of a single-operator node still addresses it by the **bare** output name in the unresolved descriptor (`source: A, output: out`) — the `{op_id}/` prefix is only added later by `resolve_aliases_and_set_defaults()`. The producer key (`A`, `"op/out"`) and consumer lookup key (`A`, `"out"`) never matched, so:

- **non-strict mode:** the mismatch check and type-inference branch were both silently skipped.
- **strict mode:** a misleading *"upstream A/out has no type annotation"* warning was emitted even when the operator declared `output_types` for that port.

The multi-operator (`operators:`) form was unaffected because there the consumer already writes the operator id verbatim.

## Fix

Normalize the single-operator consumer key at lookup time in `check_edge_mismatches_with_compat`: apply the same `{op_id}/{output}` prefixing the resolver uses later, before probing `output_type_map`. This mirrors the smaller of the two approaches suggested in the issue.

Added regression tests covering:
- Type mismatch detection on an `operator:`-sourced edge
- Type inference from an annotated `operator:`-sourced upstream
- An explicit (non-default) operator id, to ensure the fix prefixes with the actual op id rather than the `"op"` default
- The strict-mode false positive no longer occurring when `output_types` is declared

## Validation

- `cargo test -p dora-core descriptor::validate::` — 86 passed (82 pre-existing + 4 new)
- `cargo clippy -p dora-core -- -D warnings`
- `cargo check --examples`

---
⚠️ **Machine-generated PR.** This change was authored by Claude (an AI agent) reviewing an automated code-review issue. It has been verified as described above but should receive human review before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HzhJpuL3Y3QMykwvx31Krt)_


phil-opp:
**Automated review by Claude** — verified the fix at `033e750` against the actual code.

No issues found. The root cause holds: for `operator:` source nodes the producer keys `output_type_map` as `(node_id, "{op_id}/{output}")` (`validate.rs:898-903`), while the unresolved consumer mapping uses the bare output name, so lookups never matched and mismatch/inference were silently skipped. The fix's consumer-side prefix (`validate.rs:1127-1131`) mirrors the producer construction exactly — including the explicit-op-id case via `op.id.unwrap_or(SINGLE_OPERATOR_DEFAULT_ID)` — so keys now match. `Timer`/`Logs` mappings are untouched and multi-operator nodes were already correct. The 4 regression tests each exercise a distinct path and would fail pre-fix.

---

phil-opp:
Closing in favor of #2593

PR #2463, opened 2026-07-03
fix(cli): scope `dora logs --node --follow` to the requested node

### LINKED ISSUE #2460: `dora logs --node <N> --follow` streams logs from ALL nodes once it starts following

## Summary

`dora logs <dataflow> --node <N> --follow` (via the coordinator) correctly restricts the **historical** portion of the output to node `N`, but the **follow** portion streams live logs from **every** node in the dataflow. The `--node` filter is silently dropped once streaming begins.

## Location

`binaries/cli/src/command/logs.rs`

The `--node` path dispatches to `logs()` (line ~116). Inside `logs()`:

```rust
// historical fetch — correctly node-scoped server-side:
ControlRequest::Logs { uuid: Some(uuid), name: None, node: node.to_string(), tail: ... }
...
if !follow {
    return Ok(());
}
stream_logs_from_coordinator(session, uuid, level, since, until, grep, config)  // <-- `node` not passed
```

`stream_logs_from_coordinator` (line ~573) subscribes with:

```rust
ControlRequest::LogSubscribe { dataflow_id: uuid, level: log_level }
```

There is no node field in the subscription, and the per-message filtering in the loop only applies `since`/`until`/`grep` — never a node filter. On the display side, `print_log_message` → `should_display` in `binaries/cli/src/output.rs:38-52` uses `config.node_filters` **only to pick a per-node level**, never to drop messages from unlisted nodes. `--node` is stored in `self.node`, which never populates `node_filters` (only `--log-filter` does). So nothing filters out the other nodes.

## Reproduction

```
dora logs mydf --node sensor --follow
```

Expected: history for `sensor`, then a live tail of **only** `sensor`.
Actual: history for `sensor`, then a live tail of **all** nodes.

## Why this is a bug

- It contradicts the `--node` flag's documented purpose ("Show logs for the given node").
- It is inconsistent with the `--local` follow path (`follow_local_logs`), which reads only the requested node's log files and stays node-scoped throughout.

## Suggested fix

Thread the requested node through the follow path and filter on it — either server-side (add a node field to `LogSubscribe`, mirroring `Logs`), or client-side by dropping messages whose `node_id` != the requested node inside the `stream_logs_from_coordinator` loop. A client-side guard is the smaller change:

```rust
// in stream_logs_from_coordinator, given an Option<&NodeId> `node`:
if let Some(want) = node
    && log_message.node_id.as_ref().map(|n| n.as_ref()) != Some(want.as_ref())
{
    continue;
}
```
</body>

phil-opp:
Fixes #2460

## Summary

`dora logs <dataflow> --node <N> --follow` correctly restricted the **historical** portion of the output to node `N`, but the **follow** portion streamed live logs from **every** node in the dataflow — the `--node` filter was silently dropped once streaming began.

## Root cause

In `binaries/cli/src/command/logs.rs`, the `--node` path's `logs()` function fetches history correctly scoped via `ControlRequest::Logs { node: node.to_string(), .. }`, but then handed off to `stream_logs_from_coordinator(session, uuid, level, since, until, grep, config)` with no `node` parameter at all. That function subscribes via `ControlRequest::LogSubscribe { dataflow_id, level }` (no node field), and its receive loop only filtered on `since`/`until`/`grep`, never on node — so messages from every node passed straight through. This was also inconsistent with the `--local` follow path (`follow_local_logs`), which stays node-scoped throughout.

## Fix

Client-side guard, as suggested in the issue (smaller change, no wire-protocol update needed):

- Added `matches_node_filter(msg_node: Option<&str>, want: Option<&str>) -> bool`, a small testable predicate mirroring the existing `matches_grep` pattern in the same file.
- `stream_logs_from_coordinator` now takes a `node: Option<&NodeId>` parameter and applies `matches_node_filter` inside the receive loop before the time/grep filters, dropping messages from other nodes.
- Updated both call sites: `logs()` now passes `Some(&node)`; the no-node-specified path passes `None`, preserving unfiltered streaming there.

Added unit tests for `matches_node_filter` covering: no filter passes everything, matching node passes, non-matching node is dropped, and system messages (`node_id: None`) are dropped when a node filter is active.

## Validation

- `cargo test -p dora-cli` — 244 passed (26 in the `logs` module, incl. 4 new)
- `cargo check --examples`

---
⚠️ **Machine-generated PR.** This change was authored by Claude (an AI agent) reviewing an automated code-review issue. It has been verified as described above but should receive human review before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HzhJpuL3Y3QMykwvx31Krt)_

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -140,6 +140,7 @@ impl Executable for LogsArgs {
                 stream_logs_from_coordinator(
                     &session,
                     uuid,
+                    None,
                     &self.level,
                     self.since,
                     self.until,
@@ -569,10 +570,27 @@ fn matches_grep(msg: &LogMessage, pattern: Option<&str>) -> bool {
     false
 }
 
-/// Subscribe to coordinator log stream with time/grep filtering.
+/// Returns whether a log message should be kept given an optional node filter.
+/// `None` means no filtering (messages from every node pass); `Some(want)`
+/// keeps only messages whose node id equals `want`.
+fn matches_node_filter(msg_node: Option<&str>, want: Option<&str>) -> bool {
+    match want {
+        None => true,
+        Some(want) => msg_node == Some(want),
+    }
+}
+
+/// Subscribe to coordinator log stream with time/grep/node filtering.
+///
+/// `node`, when set, restricts the stream to messages from that single node —
+/// mirroring the node scoping already applied to the historical fetch in
+/// [`logs`]. Without this, `--node <N> --follow` would show history for `N`
+/// but then stream live logs from every node once following began.
+#[allow(clippy::too_many_arguments)]
 fn stream_logs_from_coordinator(
     session: &WsSession,
     uuid: Uuid,
+    node: Option<&NodeId>,
     level: &dora_core::build::LogLevelOrStdout,
     since: Option<std::time::Duration>,
     until: Option<std::time::Duration>,
@@ -589,6 +607,7 @@ fn stream_logs_from_coordinator(
         since.and_then(|d| chrono::TimeDelta::from_std(d).ok().map(|td| now - td));
     let until_threshold =
         until.and_then(|d| chrono::TimeDelta::from_std(d).ok().map(|td| now - td));
+    let want_node = node.map(|n| n.as_ref());
 
     let log_rx = session.subscribe_logs(
         &serde_json::to_vec(&ControlRequest::LogSubscribe {
@@ -610,6 +629,10 @@ fn stream_logs_from_coordinator(
             serde_json::from_slice(&raw).context("failed to parse log message");
         match parsed {
             Ok(log_message) => {
+                if !matches_node_filter(log_message.node_id.as_ref().map(|n| n.as_ref()), want_node)
+                {
+                    continue;
+                }
                 if let Some(threshold) = since_threshold
                     && log_message.timestamp < threshold
                 {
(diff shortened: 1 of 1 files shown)


phil-opp:
**Automated review by Claude** — verified the diff at `80357c6` against the actual code, not the PR description.

No issues found. The `matches_node_filter` predicate is correct (`None` passes all; `Some(want)` keeps only matching node ids), it's threaded into `stream_logs_from_coordinator` and applied before the time/grep filters, and the two call sites pass the right argument (`Some(&node)` for `--node`, `None` for the all-nodes path). Consistent with how the historical fetch scopes via `ControlRequest::Logs`.

One thing worth being deliberate about: system messages (`node_id: None`) are now dropped under an active `--node <N> --follow`. That matches the `--local` follow path and is covered by a test, but it does mean coordinator/daemon-level messages won't surface in a node-scoped follow.

---

phil-opp:
Re-verified this branch locally after the CI `test-linux` failure on `5004f30`:

- `cargo test -p dora-cli` — 253 passed, 0 failed (includes the `logs::tests::*` module with the new `matches_node_filter` cases).
- The diff is scoped to a single file (`binaries/cli/src/command/logs.rs`), touches no public API used outside that file (`stream_logs_from_coordinator` is a private fn called only within `logs.rs`), and doesn't change wire types.
- `cargo fmt --all -- --check` and `cargo clippy -p dora-cli -- -D warnings` are clean.

I wasn't able to pull the actual CircleCI `test-linux` job log from this session (outbound network policy here blocks `circleci.com`), so I can't confirm which specific test failed or whether it's related to this change. Given the change's narrow blast radius and a clean local run of the full `dora-cli` suite, this looks more likely to be a flake or an unrelated pre-existing failure elsewhere in the full-workspace `cargo test --all` run than something caused by this diff — but flagging rather than asserting, since I couldn't inspect the log directly.

Could someone with CircleCI access check the `test-linux` log for `5004f30` and either re-run the job or point out the actual failing test?

---

PR #2464, opened 2026-07-03
fix(runtime): spawn the metrics monitor instead of dropping its future

phil-opp:

## Issue

`run_metrics_monitor` is an `async fn`, so calling it only builds a future — nothing runs until that future is polled. In `dora-runtime`'s `run()` the future was bound to a `_meter_provider` local and immediately dropped, never awaited or spawned:

```rust
#[cfg(feature = "metrics")]
let _meter_provider = run_metrics_monitor(config.node_id.to_string());
```

As a result the **default-enabled** `metrics` feature silently exported nothing: `init_metrics()`, `global::set_meter_provider(...)`, and `init_process_observer(...)` never executed. No CPU/memory/process metrics were emitted from the runtime even though the feature appeared active.

## Fix

Spawn the monitor as a background task, gated on `DORA_OTLP_ENDPOINT` — mirroring exactly what the node API already does (`apis/rust/node/src/node/mod.rs`). The future never resolves (the process observer runs for the node's lifetime), so it must be spawned rather than awaited inline; and gating on the endpoint avoids connecting to `localhost:4317` on every runtime start when no collector is configured. The `use dora_metrics::run_metrics_monitor;` import is moved into the `#[cfg(feature = "metrics")]` block alongside its only use site.

## Validation

- `cargo clippy -p dora-runtime -- -D warnings` (default features, so the metrics path is compiled)
- Full workspace test suite for the affected crates (green)

---


phil-opp:
## 🤖 Automated codebase review — run summary

This PR is part of a batch of **6 machine-generated PRs** (authored by Claude / Claude Code) from one in-depth review pass over the repository. Each finding was verified against `origin/main`, and each fix ships with a regression test that was confirmed to fail before the fix (except the pure perf refactor, which is covered by existing tests). The combined diff passed `cargo fmt --all --check`, per-crate `clippy -D warnings`, the affected-crate test suites, and independent adversarial review (no correctness issues found).

### Correctness
- **#2464** `fix(runtime)` — metrics monitor future was dropped, never spawned; the default-enabled `metrics` feature silently exported nothing.
- **#2465** `fix(download)` — `download_file` never checked HTTP status; a 404/500 error page was hashed, written as the artifact, and marked executable.
- **#2467** `fix(core)` — operator-id prefixing used panicking `DataId::from`; an operator id with an out-of-charset character aborted `dora check`/`graph`/`build` instead of returning a clean error.
- **#2468** `fix(daemon)` — `read_last_n_lines` overflowed `usize` on an unbounded `--tail N` (`saturating_mul`).
- **#2469** `fix(cli)` — `dora topic hz` mixed poison-tolerant and poison-panicking mutex locking; made all lock sites recover consistently.

### Optimization
- **#2466** `perf(node)` — `recv_async` re-scanned every input queue (`Scheduler::is_empty()`, O(#inputs)) on every drain iteration; check once, then drain non-blocking.

### Findings considered but **not** turned into PRs
- **MAVLink outbound `sequence` always 0** (`mavlink2-bridge-node`) — real, but a correct fix must increment only on actual transmission (UDP server-mode sends 0 bytes with no peer); modest value, deferred.
- **Dynamic-node `Subscribe` hang in multi-daemon dataflows** (`daemon/pending.rs`) — plausible, but changes the distributed readiness handshake; too risky without deeper design review.
- **`Stream::poll_next` bypasses the scheduler** (`node/event_stream`) — real divergence from `recv()` (queue limits/drop policy not applied), but the fix is a design choice (route through scheduler vs. document); needs maintainer input.
- **`LogLevelOrStdout::passes` drops stdout under level filters** (`message/common.rs`) — doc/code contradiction, but the asymmetry may be intended (stdout shown only when explicitly requested); ambiguous.
- **`operator_output_id` / daemon still build qualified `DataId` via panicking `DataId::from`** — mostly unreachable once #2467 errors earlier; the deep fix (validate `OperatorId` charset at parse time) is a larger, potentially-breaking change, out of scope.
- **`download::get_filename` mishandles query-strings / multi-param `Content-Disposition`** — overlaps existing open PR #2454; skipped.
- **`extract_err_from_stderr` over-captures on a stray `Warning:`** — minor, snippet-verbosity only.
- **`NodeId` reconstructed from log `fields` bypasses validation** (`message/common.rs`) — low severity, no reachable exploit path found.
- **Module param substitution is order-dependent / case-colliding** (`core/expand.rs`) — edge case, not a panic.
- **Runtime metrics meter label uses bare `node_id`** vs the node API's `{dataflow_id}/{node_id}` — cosmetic only.

### ⚠️ Note on CI
The failing **`Audit (cargo-audit + cargo-deny)`** check is **pre-existing and unrelated** to these PRs — none of them change dependencies. It flags two *unmaintained* transitive crates: `proc-macro-error2` 2.0.1 (RUSTSEC-2026-0173, published 2026-06-07) and `serial` 0.4.0 (RUSTSEC-2017-0008). It fails identically on `main` and on every PR here; addressing it (a `deny.toml` ignore entry or dependency bump) is out of scope per the "don't fix unrelated warnings" convention.

*These are machine-generated changes — please review before merging.*

---

phil-opp:

No issues found. The metrics monitor is now spawned as a detached `tokio::task` (gated on `DORA_OTLP_ENDPOINT`) instead of building the future and immediately dropping it. It runs inside the `block_on` multi-threaded runtime with `enable_all()` (so the exporter's `tokio::time::interval` has a time driver), there is exactly one spawn (no double-spawn), the task is intentionally detached for the node's lifetime, and errors are surfaced via `tracing::warn!`. It mirrors the existing node-API metrics pattern.

---

PR #2465, opened 2026-07-03
fix(download): reject HTTP error statuses before writing the artifact

phil-opp:

## Issue

`reqwest::get` resolves to `Ok` for any completed HTTP exchange, including 4xx/5xx responses, and `download_file` never called `error_for_status()`. A URL that returned a 404/500 (typo, moved artifact, captive-portal or proxy HTML) therefore had its **error-page body** hashed, written to `build/<name>`, marked executable on Unix, and handed to `libloading`/Python. The real failure was masked behind a confusing downstream *"failed to load shared library"*, and a bogus executable was left on disk. Callers pass `expected_sha256 = None`, so the integrity check does not catch this either.

## Fix

Add `response.error_for_status()?` immediately after the request, so an error status fails the download with a clear message before anything is read or written:

```rust
let response = response
    .error_for_status()
    .wrap_err_with(|| format!("server returned an error status for `{url}`"))?;
```

`error_for_status()` consumes and returns the same `Response` on success, so `get_filename(&response)` and `response.bytes()` still operate on an intact body/headers.

## Validation

- Added a regression test (`rejects_http_error_status`) that serves a local `404` from a one-shot `TcpListener` and asserts the download errors **and** writes no artifact. Confirmed it fails without the fix (the error page is otherwise written to disk) and passes with it.
- New `[dev-dependencies]` on `tokio` (`macros`, `rt`, `net`, `io-util`) for the test only.
- `cargo fmt`/`clippy -p dora-download --all-targets -- -D warnings`/`cargo test -p dora-download` all green.

---

### DIFF
--- a/libraries/extensions/download/Cargo.toml
+++ b/libraries/extensions/download/Cargo.toml
@@ -19,3 +19,6 @@ reqwest = { version = "0.12.4", default-features = false, features = [
 sha2 = "0.10"
 tokio = { workspace = true, features = ["fs"] }
 tracing = { workspace = true }
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt", "net", "io-util"] }
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -63,6 +63,15 @@ where
         .await
         .wrap_err_with(|| format!("failed to request operator from `{url}`"))?;
 
+    // `reqwest::get` resolves to `Ok` for any completed HTTP exchange, including
+    // 4xx/5xx responses. Without this check a 404/500 error page (HTML/JSON)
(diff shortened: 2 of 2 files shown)


phil-opp:

No issues found. The `error_for_status()` placement is correct: it rejects all 4xx/5xx statuses, runs *after* reqwest's default redirect following (so the final status is what gets checked), and sits before any file is created — so a failed download can't leave a partial artifact on disk. The regression test covers both the error result and the no-file-written invariant.

---

phil-opp:
Superseded — same story as #2454: the missing status check is real and still live on main (`error_for_status` appears nowhere in `download/src/lib.rs`), but this branch predates the #2485 rewrite. Re-applied against current main in the combined PR above.

PR #2466, opened 2026-07-03
perf(node): stop re-scanning all input queues in the recv_async drain loop

phil-opp:

## Issue

The scheduler-path drain loop in `EventStream::recv_async` called `Scheduler::is_empty()` on **every** iteration:

```rust
loop {
    if self.scheduler.is_empty() {            // O(#inputs): scans every input queue
        if let Some(event) = self.receiver.recv().await { self.add_event(event); } else { break; }
    } else {
        match self.receiver.try_recv() { Ok(e) => self.add_event(e), Err(_) => break };
    }
}
```

`Scheduler::is_empty()` scans every input queue (O(#inputs)). Inside this loop the scheduler only ever grows — it exclusively `add_event`s — so once it becomes non-empty it stays non-empty for the rest of the drain. Draining `K` events with `M` inputs therefore cost **O(K·M)** scans on the per-event hot path.

## Fix

Block for events while the scheduler is still empty, then drain the remaining ready events with `try_recv` unconditionally:

```rust
while self.scheduler.is_empty() {
    match self.receiver.recv().await {
        Some(event) => self.add_event(event),
        None => break,
    }
}
while let Ok(event) = self.receiver.try_recv() {
    self.add_event(event);
}
```

The blocking part is a `while`, **not** an `if`: `add_event` usually pushes, but it can also *drop* the event without retaining it (e.g. `queue_size: 0` → `DropIncoming`). A single `recv` therefore does not guarantee the scheduler becomes non-empty, so we must keep blocking while it is still empty. An `if` would let a dropped-only first event fall through to `scheduler.next() == None` and be misread as a closed stream.

Behavior is unchanged: `add_event` is invoked on exactly the same events in the same order, and channel-close still yields `scheduler.next() == None` (stream ends). So queue-size limits, drop policy, and drop-count accounting are all preserved. The common case now checks `is_empty()` only twice (once true → block+add, once false → exit) instead of once per drained event.

## Validation

- Existing 107 `dora-node-api` unit tests pass (scheduler drop/flush/correlation coverage pins the behavior).
- `cargo fmt`/`clippy -p dora-node-api -- -D warnings` green.

---


phil-opp:
The optimization itself looks correct for all realistic configurations: `add_event` runs on the same events in the same order, and hoisting the `is_empty()` check out of the loop is sound because the scheduler only ever grows during the drain (you only `add_event`), so once it's non-empty it stays non-empty.

One caveat on the "behavior is unchanged" claim, in case it matters to you: the equivalence relies on `recv().await` returning `Some` ⇒ the scheduler becomes non-empty. That can be false for an input with `queue_size: 0` (a `drop_oldest` cap of 0, which nothing currently rejects): `add_event` drops the event immediately, leaving the scheduler still empty. In that case the old `loop { if is_empty() { recv().await } }` would block again waiting for a surviving event, whereas the new code falls through to `scheduler.next() == None` and **ends the event stream** (node shuts down) even though the dataflow is still live.

It's a degenerate config — `queue_size: 0` means "drop everything on this input", and the old behavior (block forever) was also broken — so I don't think it blocks the change. Just flagging that the strict-equivalence framing has this one hole; a one-line comment noting the fast path assumes "non-empty scheduler ⇒ a retained event exists" would capture the real invariant.

---
*🤖 This is a fully automated review by Claude. No human has reviewed these comments — please treat them as suggestions from an automated tool and verify before acting.*

---

phil-opp:
## Blocking: the stated loop invariant does not hold

> inside this drain the scheduler only ever grows — we exclusively `add_event`

`Scheduler::add_event` has an early `return` that does not push: the `Eviction::DropIncoming` arm at `apis/rust/node/src/event_stream/scheduler.rs:293-297`.

Reachability: that arm needs `queue.len() >= cap`; with an empty queue that means `cap == 0`. `cap = policy.effective_cap(*size)`, and the default `DropOldest` returns `queue_size` verbatim (`libraries/message/src/config.rs:45`). `queue_size` is a plain `Option<usize>` (`config.rs:101`) fed through `config.queue_size.unwrap_or(DEFAULT_QUEUE_SIZE)` (`event_stream/mod.rs:174-177`) — and nothing under `libraries/core/src/descriptor/` validates or clamps it, so `queue_size: 0` is reachable from user YAML.

With `queue_size: 0`, `select_eviction` on an empty queue (`scheduler.rs:53-61`) returns `DropIncoming` for any plain input. Resulting behaviour change:

* **before:** block in `recv().await` forever — a hang, stream stays open
* **after:** one dropped `add_event` -> `try_recv` drains -> `scheduler.next()` -> `None` -> `recv_async` returns `None`, which is the *stream-closed* signal — **the node silently exits as if the dataflow finished**

Hang -> silent early termination is the worse of the two failure modes.

To be clear about what is right here: for every `cap >= 1` the change **is** exactly equivalent — `DropIncoming` then requires a non-empty queue, so the new code never blocks in that case and the `add_event` sequence and order are identical. No starvation, no lost messages, no fairness change (`Scheduler::next` is untouched). The O(K·M) -> O(K) win is real and worth having.

Minimal fix that keeps the win:

```rust
while self.scheduler.is_empty() {
    match self.receiver.recv().await {
        Some(event) => self.add_event(event),
        None => break,
    }
}
while let Ok(event) = self.receiver.try_recv() { self.add_event(event); }
```

Costs 2 `is_empty()` calls in the normal case instead of K+1. Separately, `queue_size: 0` is a latent hang on main today — worth clamping/validating to `>= 1` regardless of this PR.

phil-opp:
Good catch — you're right, and thanks for the precise reachability analysis. My "only ever grows" claim was wrong: `add_event`'s `Eviction::DropIncoming` arm returns without pushing, and `queue_size: 0` makes it reachable on an empty queue, so a single blocking `recv` can leave the scheduler empty and fall through to `scheduler.next() == None` — turning the old hang into a silent stream-close. That's the worse failure mode.

Applied your suggested fix (pushed as `64d0d6c`, rebased onto the branch's current head after your merge of `main`):

```rust
while self.scheduler.is_empty() {
    match self.receiver.recv().await {
        Some(event) => self.add_event(event),
        None => break,
    }
}
while let Ok(event) = self.receiver.try_recv() {
    self.add_event(event);
}
```

This restores the exact previous "block until a *retained* event arrives" semantics while keeping the O(K·#inputs) → O(K) win (two `is_empty()` calls in the common case). The comment is updated to describe the drop case instead of overstating the invariant. `clippy -D warnings` and the node-api unit tests (140) pass.

I left the separate `queue_size: 0` latent-hang on `main` alone here, as you suggested — happy to open a follow-up that clamps/validates `queue_size >= 1` at descriptor parse time if you'd like.

---

phil-opp:

The `64d0d6c` fix (switching the fast path to `while self.scheduler.is_empty()`) resolves the earlier `queue_size: 0` / `DropIncoming` concern: the drain now keeps blocking until a *retained* event arrives, so a dropped-only event no longer falls through to `scheduler.next() == None` and silently closes the stream (`scheduler.rs:294` is the reachable non-pushing `return`). For every `cap >= 1` it stays exactly equivalent to the old loop while keeping the O(K·#inputs) → O(K) win. No new issues from my side.

---

PR #2467, opened 2026-07-03
fix(core): return an error instead of panicking on invalid operator-id prefix

phil-opp:

## Issue

When a single-operator node is referenced by another node's input, `resolve_aliases_and_set_defaults` rewrites the referencing output to `<operator-id>/<output>`. It built that qualified id with `DataId::from(String)`, which **panics** on any character outside `[a-zA-Z0-9_./-]`.

Operator ids are *not* validated against the `DataId` charset at parse time (`OperatorId` stores its string verbatim and its `Deserialize` is the plain derived impl), so a descriptor like:

```yaml
nodes:
  - id: producer
    operator:
      id: "bad id"          # space accepted by OperatorId
      python: op.py
      outputs: [result]
  - id: consumer
    path: ./consumer
    inputs:
      x: producer/result
```

parses fine but aborts `dora check` / `graph` / `build` with a panic (`invalid DataId 'bad id/result'`) instead of a clean, actionable descriptor error.

## Fix

Build the qualified id via `parse::<DataId>()` in a small helper shared by both prefix sites (the ROS2 branch and the general input-mapping branch), mapping a parse failure to an `eyre` error that is propagated with `?`. For valid ids behavior is identical.

## Validation

- Added a regression test (`invalid_operator_id_prefix_errors_instead_of_panicking`) that resolves the descriptor above and asserts `Err`. Confirmed it **panics** at `id.rs:209` without the fix and returns a clean error with it.
- `cargo fmt`/`clippy -p dora-core -- -D warnings`/`cargo test -p dora-core descriptor::tests` all green.

---


phil-opp:

No issues found. Both `DataId::from(format!(...))` panic sites are replaced by the fallible `prefix_output_with_operator_id` helper (which uses `parse::<DataId>()`), and both call sites live inside `resolve_aliases_and_set_defaults` (which returns `eyre::Result`) and propagate via `?` — no `unwrap`/`expect` reintroduces a panic. `From<String>` for `DataId` panics while `FromStr` is fallible, so the swap is semantically sound, and the error message includes the offending operator/output ids. The regression test covers the invalid-prefix path.

---

PR #2468, opened 2026-07-03
fix(daemon): saturate the log-tail buffer estimate to avoid usize overflow

phil-opp:

## Issue

`read_last_n_lines` sizes its read buffer from `estimated_line_length * tail`, where `tail` is the client-supplied `dora logs --tail N` line count **with no upper bound**. Once the first chunk is read `estimated_line_length >= 1`, so a large `N` (e.g. near `usize::MAX`) overflows the multiply — a panic in the spawned log task in debug builds, a silent wrap in release.

## Fix

The product only feeds a buffer-growth heuristic, so use `saturating_mul`:

```rust
let estimated_buffer_length = estimated_line_length.saturating_mul(tail);
```

For all non-overflowing inputs this is identical to `*`, so normal behavior is unchanged.

## Validation

- Added a regression test (`read_last_n_lines_handles_huge_tail_without_overflow`) that requests `usize::MAX` lines from a small file and asserts the whole file is returned. Confirmed it **panics with "attempt to multiply with overflow"** at `lib.rs:4942` without the fix and passes with it.
- `cargo fmt`/`clippy -p dora-daemon -- -D warnings`/`cargo test -p dora-daemon` all green.

---


phil-opp:

No issues found. The `saturating_mul(tail)` change at `binaries/daemon/src/lib.rs:4938` correctly avoids the `usize` overflow: at that point the loop guard guarantees `tail > 0` (the `tail == 0` case breaks earlier) and `estimated_line_length >= 1` after the first read, so even a saturated `usize::MAX` still satisfies the `>= buffer.len() * 2` growth check — the same outcome as the un-overflowed value, with no behavior change for normal inputs and no off-by-one. The regression test exercises the previously-panicking path.

---

PR #2469, opened 2026-07-03
fix(cli): recover from lock poisoning consistently in `dora topic hz`

phil-opp:

## Issue

`HzStats::timestamps` is shared between the receiver thread (which writes) and the UI/summary readers. Two of the four lock sites already tolerate a poisoned mutex with `unwrap_or_else(|e| e.into_inner())`, but the oneshot summary (`run_hz_oneshot`) and `intervals_ms` used a bare `.unwrap()`. If any holder ever panicked while holding the lock, the writer kept working while the next reader **panicked on the poisoned lock**, taking down the TUI — an inconsistency relative to the recovery the same file already uses elsewhere.

## Fix

Make all four lock sites recover consistently, matching the pattern the `record` path already uses (`unwrap_or_else(|e| e.into_inner())`). This is the established idiom in the sibling `topic/info.rs` as well.

## Validation

- Added a regression test (`lock_reads_recover_from_poisoning`) that poisons the mutex from a panicking thread, then exercises `record` and `intervals_ms`. Confirmed it **panics at the reader** without the fix and passes with it.
- `cargo fmt`/`clippy -p dora-cli -- -D warnings`/`cargo test -p dora-cli` all green.

---


phil-opp:

No issues found. The fix makes all four `HzStats::timestamps` lock sites recover from poisoning consistently (previously `record` and the TUI loop already recovered via `unwrap_or_else(|e| e.into_inner())`, while `intervals_ms` and the oneshot sampler still used bare `.unwrap()`). Recovering rather than propagating can't hide dangerous state here: the guarded data is a single `VecDeque<Instant>` used only for display statistics with no cross-field invariant, so a poisoned lock leaves a structurally valid deque — at worst one timestamp sits briefly outside the sliding window and self-heals on the next `record`. The added test poisons the mutex and confirms both the writer (`record`) and reader (`intervals_ms`) paths still work.

---

Issue #2452, opened 2026-07-02
chore(deps): unblock prettyplease 0.2 in dora-node-api-cxx (rust-format verbatim coupling)


phil-opp:
`apis/c++/node/Cargo.toml` deliberately keeps `prettyplease = "0.1"` (now with an explanatory comment): the dependency exists only to enable the `verbatim` feature on the prettyplease **0.1** copy that `rust-format` 0.3 uses internally, via cargo version unification. Bumping it to 0.2 makes it inert — rust-format's internal 0.1.25 then builds *without* `verbatim`, and codegen containing verbatim items would panic at build time.

Options to actually modernize:

1. Migrate `apis/c++/node/build.rs` off `rust-format` to `syn 2` + `prettyplease 0.2::unparse` directly (the sibling `libraries/extensions/ros2-bridge/build.rs` already formats the same generated code — align the two while at it), or
2. Wait for a rust-format release that uses prettyplease 0.2 / syn 2.

Until one of those happens, the 0.1 pin must stay.

Deferred from the dependency-upgrade PR #2445.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2451, opened 2026-07-02
chore(deps): remove RUSTSEC waivers for upstream-pinned transitives (lz4_flex, quick-xml, rustls-pemfile)


phil-opp:
The dependency-upgrade PR #2445 added RUSTSEC waivers to `deny.toml` + `.cargo/audit.toml` for advisories that can only be fixed by upstream releases. Tracking them here so the waivers don't outlive their justification (all carry `Review: 2026-09`):

- **RUSTSEC-2026-0041** — lz4_flex 0.10 uninitialized-memory leak (high, 8.2). Transitive via `zenoh-transport` 1.9 (pins ^0.10). dora doesn't enable zenoh compression, so the decompression path isn't reachable. Fix: zenoh release on lz4_flex ≥0.11.6.
- **RUSTSEC-2026-0194 / RUSTSEC-2026-0195** — quick-xml <0.41 quadratic attribute check + NsReader memory-exhaustion DoS (high, 7.5). Our own parser (ros2 msg-gen) is on the fixed 0.41; remaining copies are transitive via `self_update` 0.44 (quick-xml 0.38) and `mavlink-bindgen` 0.18 (quick-xml 0.39, build-time codegen only). Fix: upstream bumps in self_update / rust-mavlink.
- **RUSTSEC-2025-0134** — rustls-pemfile unmaintained. Newly introduced by zenoh 1.9 via `zenoh-link-commons`. Fix: zenoh migrating to the `PemObject` API in rustls-pki-types.

When a parent releases, `cargo update` + removing the waiver line should be enough; cargo-deny warns `advisory-not-detected` when a waiver goes stale.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

phil-opp:
Scheduled-review checkpoint (the waivers carry `Review: 2026-09`). Verified against the current `Cargo.lock` today — all three transitives are still on the vulnerable/unmaintained versions, and each is pulled by a semver-pinned parent, so `cargo update` alone cannot advance them and removing any waiver now would just re-break the audit gate. No action yet; the waivers remain justified.

| Advisory | Crate (current) | Parent pinning it | Fix still pending on |
| --- | --- | --- | --- |
| RUSTSEC-2026-0041 | `lz4_flex 0.10.0` | `zenoh-transport 1.9.0` | zenoh bump to `lz4_flex ≥0.11.6` |
[...]

Issue #2450, opened 2026-07-02
chore(deps): lift rustdds =0.11.4 pin + bump ros2-client once RustDDS#375 is fixed


phil-opp:
`dora-ros2-bridge` pins `rustdds = "=0.11.4"` because rustdds 0.11.5+ pulls in `pnet`, which breaks Windows builds (Atostek/RustDDS#375, still open — requires Packet.lib/WinPcap SDK). This pin transitively holds `ros2-client` at 0.8.1 (0.8.2 needs rustdds ~0.11.6, 0.9.0 needs rustdds ^0.12).

It also keeps the RUSTSEC advisory surface larger than necessary: rustdds 0.11.6+ is where the fixed lz4_flex range lands for that stack.

When Atostek/RustDDS#375 is resolved (or pnet becomes optional upstream):

1. Drop the `=0.11.4` pin in `libraries/extensions/ros2-bridge/Cargo.toml`.
2. Bump `ros2-client` to latest (0.9+ at time of writing; check API changes).
3. Re-run the nightly Windows cross-check before merging.

Deferred from the dependency-upgrade PR #2445.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

phil-opp:
Fixed upstream, but not released yet: https://github.com/Atostek/RustDDS/commit/92c9b117c

PR #2472, opened 2026-07-04
fix(core): reject overflowing timing fields in check_timing_fields

### LINKED ISSUE #2470: daemon: check_timing_fields misses Duration::from_secs_f64 overflow — a large finite finish_grace_secs/restart_delay/etc. still panics the daemon on spawn
## Summary

PR #2285 (`cc2b9b5`) added `check_timing_fields` in `libraries/core/src/descriptor/validate.rs` whose stated purpose is to *"surface a clean error instead of crashing the daemon in `Duration::from_secs_f64` on spawn."* The guard is incomplete: it rejects negative / NaN / infinite values but **not finite-but-overflowing** ones, so the exact panic it was written to prevent is still reachable.

## Affected code

Guard (`libraries/core/src/descriptor/validate.rs:157-183`):

```rust
/// The daemon converts these to `Duration` via `Duration::from_secs_f64`, which
/// panics on negative, NaN, or infinite input. Catching them here surfaces a
/// clean descriptor error instead of crashing the daemon on node spawn.
fn check_timing_fields(...) -> eyre::Result<()> {
    for (field, value) in [
        ("finish_grace_secs", custom.finish_grace_secs),
        ("health_check_timeout", custom.health_check_timeout),
        ("restart_delay", custom.restart_delay),
        ("max_restart_delay", custom.max_restart_delay),
        ("restart_window", custom.restart_window),
    ] {
        if let Some(value) = value
            && (!value.is_finite() || value < 0.0)   // <-- misses overflow
        {
            bail!("... must be a finite, non-negative number of seconds");
        }
    }
    Ok(())
}
```

`Duration::from_secs_f64` has **four** panic conditions, not three: negative, NaN, infinite, **and overflow** (any finite value greater than `Duration::MAX.as_secs_f64()` ≈ `1.8e19`). The docstring itself lists only the first three, so the overflow case was simply overlooked.

The five daemon conversion sites this guard is meant to protect all use the panicking API with no fallback (`binaries/daemon/src/spawn/prepared.rs`):

```
160:  n.health_check_timeout.map(Duration::from_secs_f64)
169:  n.finish_grace_secs.map(Duration::from_secs_f64)
179:  restart_delay:      n.restart_delay.map(Duration::from_secs_f64),
180:  max_restart_delay:  n.max_restart_delay.map(Duration::from_secs_f64),
181:  restart_window:     n.restart_window.map(Duration::from_secs_f64),
```

## Repro / failure scenario

A descriptor value like the following is finite and non-negative, so it passes `check_timing_fields`, then panics when `PreparedNode` builds the node:

```yaml
nodes:
  - id: trainer
    path: ./trainer
    finish_grace_secs: 1e20   # or restart_delay: 1e19, etc.
```

Verified with the stdlib directly:

```
is_finite = true, >= 0 = true
Duration::try_from_secs_f64(1e20) = Err(TryFromFloatSecsError { kind: OverflowOrNan })
Duration::from_secs_f64(1e20)     -> panics
```

The panic aborts the daemon task on node spawn, taking down dataflows on that daemon — precisely the outcome the new validation was added to avoid. It is an edge case (requires an absurdly large number, e.g. someone writing `1e30` to mean "effectively infinite", or a config-generation bug), but it is a real robustness gap that defeats the stated purpose of the newly added code.

## Suggested fix

The codebase already establishes the correct pattern for exactly this hazard — use the non-panicking API:

- `libraries/extensions/ros2-bridge/python/src/lib.rs:1389` — `Duration::try_from_secs_f64(secs).unwrap_or_else(...)`, with a comment noting it "guards against `Duration::from_secs_f64` panicking".
- `libraries/message/src/config.rs:283` — `Duration::try_from_secs_f64(...).map_err(...)`, with a comment that the previous `from_secs_f64` "would panic".

Two straightforward options:

1. Extend the guard's condition to also reject overflow, e.g. `!value.is_finite() || value < 0.0 || value > Duration::MAX.as_secs_f64()`, and update the docstring to mention overflow; **or**
2. Replace the five `map(Duration::from_secs_f64)` call sites in `prepared.rs` with `Duration::try_from_secs_f64(...)` and handle the `Err` (defense-in-depth, since validation is not guaranteed to run on every path that reaches the daemon).

Option 1 is the minimal fix aligned with the PR's intent; option 2 additionally protects paths that might bypass `check_dataflow`.

---
*This issue was created by a scheduled automated Claude review of recent commits/PRs. The finding was verified against the code (the overflow panic and the guard gap were reproduced with the stdlib), but please double-check before acting on it.*

phil-opp:
## Summary

`check_timing_fields` (added in #2285) was meant to surface a clean descriptor
error instead of crashing the daemon in `Duration::from_secs_f64` on node
spawn. It rejected negative, NaN, and infinite values, but `Duration::from_secs_f64`
has a fourth panic condition: finite values larger than `Duration::MAX.as_secs_f64()`
(≈ 1.8e19). A descriptor field like `finish_grace_secs: 1e20` passes the
existing guard (finite, non-negative) and then panics the daemon on spawn —
exactly the failure mode the guard was added to prevent.

## Fix

Extend the guard's condition to also reject values greater than
`Duration::MAX.as_secs_f64()`, and update the docstring/error message
accordingly.

## Test plan

- Added `timing_fields_reject_finite_value_that_overflows_duration`, which
  asserts `restart_delay: 1e20` is rejected by `check_timing_fields` (and
  independently confirms `Duration::try_from_secs_f64(1e20)` returns `Err`).
- `cargo test -p dora-core` — all 237 tests + 2 doctests pass.
- `cargo fmt --all -- --check` and `cargo clippy -p dora-core -- -D warnings` — clean.
- `cargo check --examples` — passes.

Fixes #2470.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VD5Utmx4fE9YFB64fduR8x)_

### DIFF
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -126,8 +126,8 @@ pub fn check_dataflow(dataflow: &Descriptor, working_dir: &Path) -> eyre::Result
         }
     }
 
-    // reject negative / non-finite timing values before they reach the daemon,
-    // where `Duration::from_secs_f64` would panic on spawn.
+    // reject negative / non-finite / overflowing timing values before they
+    // reach the daemon, where `Duration::from_secs_f64` would panic on spawn.
     for node in nodes.values() {
         if let descriptor::CoreNodeKind::Custom(custom) = &node.kind {
             check_timing_fields(&node.id, custom)?;
@@ -175,11 +175,13 @@ pub fn check_dataflow(dataflow: &Descriptor, working_dir: &Path) -> eyre::Result
     Ok(())
 }
 
-/// Reject negative / non-finite second-valued timing fields on a custom node.
+/// Reject negative / non-finite / overflowing second-valued timing fields on a
+/// custom node.
 ///
 /// The daemon converts these to `Duration` via `Duration::from_secs_f64`, which
-/// panics on negative, NaN, or infinite input. Catching them here surfaces a
-/// clean descriptor error instead of crashing the daemon on node spawn.
+/// panics on negative, NaN, infinite, or overflowing (> `Duration::MAX`) input.
+/// Catching them here surfaces a clean descriptor error instead of crashing the
+/// daemon on node spawn.
 fn check_timing_fields(
     node_id: &NodeId,
     custom: &dora_message::descriptor::CustomNode,
@@ -197,15 +199,22 @@ fn check_timing_fields(
     Ok(())
 }
 
-/// Reject a negative / non-finite second-valued field before it reaches the
-/// daemon, where `Duration::from_secs_f64` would panic on such input.
+/// Reject a negative / non-finite / overflowing second-valued field before it
+/// reaches the daemon, where `Duration::from_secs_f64` would panic on such
+/// input.
+///
+/// `Duration::from_secs_f64` panics on negative, NaN, infinite, *and* finite
+/// values too large to fit in a `Duration` (> `Duration::MAX`, ≈ 1.8e19 s). We
+/// probe the exact same boundary with its non-panicking twin
+/// `try_from_secs_f64`, so a value accepted here can never panic the daemon.
 fn check_seconds_field(owner: &str, field: &str, value: Option<f64>) -> eyre::Result<()> {
     if let Some(value) = value
-        && (!value.is_finite() || value < 0.0)
+        && std::time::Duration::try_from_secs_f64(value).is_err()
     {
         bail!(
             "{owner} has invalid `{field}`: {value} \
-             (must be a finite, non-negative number of seconds)"
+             (must be a finite, non-negative number of seconds smaller than {})",
+            std::time::Duration::MAX.as_secs_f64()
         );
     }
     Ok(())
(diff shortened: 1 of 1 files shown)


phil-opp:
The `ci/circleci: audit` failure appears unrelated to this change — this PR doesn't touch `Cargo.lock`, `Cargo.toml`, `deny.toml`, or `.cargo/audit.toml` at all (only `libraries/core/src/descriptor/validate.rs`), so it can't have introduced a new advisory or policy violation. This matches the known, already-tracked RUSTSEC advisories in #2451, #2447, #2446, #2450, #2448, #2449, and #2452, which document dependency advisories that are pending upstream fixes. Leaving this as-is; let me know if the audit gate should block merge here regardless.

---

phil-opp:
I found one issue: the overflow guard has a single-value boundary hole that leaves exactly the panic this PR is meant to close.

The new condition is `value > std::time::Duration::MAX.as_secs_f64()`. But `Duration::MAX.as_secs_f64()` rounds **up** to `1.8446744073709552e19`, which is strictly larger than the true `Duration::MAX`. So `Duration::from_secs_f64` still panics on that exact boundary value, while `value > max` is `false` when `value == max` — meaning the value slips through validation and then panics the daemon on node spawn.

I verified this locally:

```rust
let max = std::time::Duration::MAX.as_secs_f64(); // 1.8446744073709552e19
assert!(std::time::Duration::try_from_secs_f64(max).is_err()); // from_secs_f64 would panic
assert!(!(max > max));                                         // guard does NOT reject it
```

So a descriptor like:

```yaml
restart_delay: 1.8446744073709552e19
```

passes `check_timing_fields` and then crashes the daemon — the failure mode #2470 is about.

Two ways to close it:
- change `>` to `>=`, or
- drop the hand-rolled float checks and gate on the same function the daemon actually calls:

```rust
if let Some(value) = value
    && std::time::Duration::try_from_secs_f64(value).is_err()
{
    bail!(...);
}
```

The `try_from_secs_f64` form is the tightest match to the real panic condition (it also covers negative / NaN / infinite for free, so it can replace the whole `!value.is_finite() || value < 0.0 || value > ...` expression) and it can't drift from `from_secs_f64`'s actual boundary. Worth adding a test at exactly `Duration::MAX.as_secs_f64()` alongside the existing `1e20` case — the current test passes despite the gap because `1e20 > max`.

---
*🤖 This is a fully automated review by Claude. No human has reviewed these comments — please treat them as suggestions from an automated tool and verify before acting.*

---

Issue #2449, opened 2026-07-02
chore(deps): redb 4 + sysinfo 0.39 blocked on MSRV (need Rust 1.89 / 1.95)


phil-opp:
Two dependencies are held back solely by the workspace MSRV (`rust-version = "1.88.0"`):

- **redb 2.6.3 → 4.1**: every redb 3.x and 4.x release requires Rust 1.89. When bumping, also check the redb file-format implications for `coordinator-store` persistence (redb 3 introduced a new file format; existing store files may need a migration path).
- **sysinfo 0.38.4 → 0.39+**: 0.39.5 requires Rust 1.95.

Revisit at the next MSRV bump; cargo's MSRV-aware resolver will pick these up once `rust-version` allows (they currently show as "unchanged behind latest" in `cargo update --verbose`).

Deferred from the dependency-upgrade PR #2445.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2448, opened 2026-07-02
chore(deps): upgrade OpenTelemetry stack to 0.32 once opentelemetry-system-metrics 0.32 ships


phil-opp:
`dora-tracing` and `dora-metrics` are held at opentelemetry/opentelemetry_sdk/opentelemetry-otlp 0.31 (and tracing-opentelemetry 0.32) because `opentelemetry-system-metrics` — a direct dependency of `dora-metrics` — has no release supporting otel 0.32 yet (latest on crates.io is 0.31.0, which pins otel 0.31; two otel majors in one metrics pipeline don't type-check).

The upstream bump is already prepared (see the `opentelemetry-system-metrics` 0.32 patch). Once 0.32.0 is published:

- opentelemetry, opentelemetry_sdk, opentelemetry-otlp → 0.32
- tracing-opentelemetry → 0.33
- opentelemetry-system-metrics → 0.32

This also removes one of the two reqwest majors in the CLI build (opentelemetry-http still pulls reqwest 0.12 while self_update/dora-download are on 0.13).

Deferred from the dependency-upgrade PR #2445.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

phil-opp:
@claude status?

Issue #2447, opened 2026-07-02
chore(deps): migrate off unmaintained bincode 1.x/2.x (3.0.0 is a protest stub)


phil-opp:
The workspace uses bincode 1.3.3 (dora-message wire codec: daemon↔node TCP, inter-daemon events, recordings) and bincode 2.0.1 (coordinator-store persisted values). Both are flagged unmaintained (RUSTSEC-2025-0141, waived in `deny.toml`), and upstream's "latest" release cannot be adopted:

- bincode 3.0.0 on crates.io is a protest stub — its entire source is `compile_error!("https://xkcd.com/2347/")`. Development moved to sourcehut (https://sr.ht/~stygianentity/bincode/).

Migration options to evaluate post-1.0:

1. bincode 2's serde compat layer with `config::legacy()`, which is documented to be byte-compatible with bincode 1's default format — would unify on 2.0.1 without a wire break (verify with golden-bytes tests before switching).
2. A maintained alternative (e.g. postcard), which *is* a wire break and needs a protocol-version bump — could be batched with the uhlc 0.9 migration (see the uhlc issue) so we pay for one coordinated break, not two.

Deferred from the dependency-upgrade PR #2445.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

phil-opp:
Fixed by #3153

PR #2473, opened 2026-07-04
fix(coordinator-store): bump SCHEMA_VERSION for terminal field so old records fail loudly, not silently

### LINKED ISSUE #2471: `#[serde(default)] terminal` is ineffective under bincode — pre-#1854 `Failed` dataflow records silently vanish on coordinator restart

## Summary

The `terminal` field on `DataflowStatus::Failed` carries a doc comment promising that `#[serde(default)]` makes older persisted records (written before the field existed) deserialize with `terminal: false`. **That promise does not hold**, because the redb backend encodes/decodes with **bincode**, which is not self-describing. As a result, a `Failed { error }` record persisted by a pre-#1854 coordinator becomes **undecodable** after an upgrade, and the row is then **silently dropped** from `list_dataflows()` (and errors in `get_dataflow()`). This defeats the exact invariant the `terminal` flag was introduced to protect.

## Location & code

`libraries/coordinator-store/src/lib.rs`, `DataflowStatus::Failed` (lines ~80–97):

```rust
Failed {
    error: String,
    /// ...
    /// `#[serde(default)]` so persisted records written by older
    /// coordinators (which never set this field) deserialize with
    /// `terminal: false` -- preserving the pre-#1854 behaviour
    /// where a daemon report could promote Failed -> Running.
    #[serde(default)]
    terminal: bool,
},
```

`libraries/coordinator-store/src/redb_store.rs` — the encoding is **bincode 2.x standard config**, not JSON/self-describing:

```rust
fn encode<T: serde::Serialize>(value: &T) -> Result<Vec<u8>> {
    bincode::serde::encode_to_vec(value, bincode::config::standard())
        .map_err(|e| eyre!("encode error: {e}"))
}

fn decode<T: serde::de::DeserializeOwned>(bytes: &[u8]) -> Result<T> {
    let config = bincode::config::standard().with_limit::<{ 64 * 1024 * 1024 }>();
    let (val, _) =
        bincode::serde::decode_from_slice(bytes, config).map_err(|e| eyre!("decode error: {e}"))?;
    Ok(val)
}
```

## Why the `#[serde(default)]` promise is false

`#[serde(default)]` only fills in a missing field for **self-describing** formats (JSON, YAML, etc.), where the deserializer can see that a field is absent. bincode is positional/non-self-describing: an enum variant is encoded as `variant_index` followed by its fields **in order, with no field names or count**. Decoding old bytes for `Failed { error }` into the new `Failed { error, terminal }` shape reads the variant index and the `error` string, then tries to read one more byte for the `bool` — and there is nothing left, so it fails with `UnexpectedEnd`.

Empirically (bincode 2.0):

```
encoded old bytes: [1, 4, 98, 111, 111, 109]   // Failed { error: "boom" }, no terminal
decode into { error, terminal } => Err(UnexpectedEnd { additional: 1 })
```

## Why the schema-version guard does not catch it

`redb_store.rs` has a schema-version mechanism intended for exactly this situation:

```rust
const SCHEMA_VERSION: u32 = 2; // Sprint 9: added node_to_daemon, uv, Recovering status
```

On open, a stored version `!= SCHEMA_VERSION` refuses to open the DB and tells the user to delete it. But `terminal` was added (per the field's own `pre-#1854` provenance) **without bumping `SCHEMA_VERSION`** — the constant is still `2`, and its comment lists only the Sprint 9 additions (`node_to_daemon`, `uv`, `Recovering`), **not** `terminal`. So an old v2 database opens cleanly (version matches), and the format skew is only discovered — fatally — at row-decode time.

## Concrete failing scenario

1. A pre-#1854 coordinator persists a terminally-failed dataflow as `Failed { error }` (schema v2).
2. The coordinator binary is upgraded to current `main`. `RedbStore::open` sees stored version `2 == SCHEMA_VERSION`, so it does **not** reject or migrate.
3. On recovery / `dora list` / `clean`, `list_dataflows()` calls `decode()` on that row, which returns `Err`, and the row is skipped:

```rust
match decode(value.value()) {
    Ok(record) => result.push(record),
    Err(e) => tracing::warn!("skipping corrupt dataflow record: {e}"),
}
```

`get_dataflow()` returns the decode `Err` for the same row.

**Net effect:** the terminally-failed dataflow record disappears from the store after restart — precisely the "resurrection / inconsistent store-vs-CLI view across coordinator restarts" that `terminal` was added to prevent. The field meant to *harden* the failure invariant instead makes the record carrying that invariant unreadable.

## Generalization

This is a latent trap for the whole persisted schema: **any** future field added to an already-persisted bincode struct/enum-variant without bumping `SCHEMA_VERSION` will break decoding of pre-existing rows the same way, and the corrupt-row-skipping in `list_daemons` / `list_dataflows` / `list_builds` will hide the loss (logged only at `warn`). Note `uv` (lib.rs:61, also `#[serde(default)]`) got away with it only because Sprint 9 *did* bump the schema to 2.

## Suggested fix

Either of these makes the compatibility real; the `#[serde(default)]` attribute alone does not:

- **(a) Treat any persisted-shape change as a schema change:** bump `SCHEMA_VERSION` (to 3 for `terminal`) whenever a stored record's byte layout changes, and provide an explicit migration (or accept the "delete and restart" path the guard already offers). This restores the guarantee that a version match implies a decodable layout.
- **(b) Use a tolerant/self-describing encoding** for these records (or a manual `Deserialize` that treats a truncated buffer as `terminal = false`), so trailing-field additions really are backward-compatible.

A regression test that encodes the old `Failed { error }` byte layout and asserts it decodes to `Failed { error, terminal: false }` (or is handled by migration) would lock this in and would currently fail.

phil-opp:
`DataflowStatus::Failed` gained a `terminal` field (#1854) without bumping the
persisted-store `SCHEMA_VERSION`, but the redb backend encodes with **bincode**,
which is not self-describing. `#[serde(default)]` cannot backfill a missing
trailing field under bincode the way it can for JSON/YAML — and because `status`
is a *middle* field of `DataflowRecord`, an old record doesn't fail cleanly: the
decoder reads the next field's byte as `terminal` and shifts by one into every
field after it. So `list_dataflows`/`get_dataflow` either error or silently
mis-decode, and a terminally-failed dataflow can vanish from `dora list` after a
coordinator restart — the exact invariant `terminal` was added to protect.

This bumps `SCHEMA_VERSION` `2` → `3` so a pre-`terminal` database is rejected
loudly at `open()` (the existing "delete the file and restart" error) instead of
opening cleanly and dropping rows at decode time. Also corrects the `terminal`
and `SCHEMA_VERSION` doc comments that wrongly described `#[serde(default)]` as
sufficient for bincode backward-compatibility.

Note: schema `2` is ambiguous — `v1.0.0-rc.1` stamped pre-`terminal` records as
v2, and `v1.0.0-rc.2` stamped post-`terminal` records as v2 too — so **all** v2
stores are rejected at `open()`, including rc.2 stores whose rows would still
decode correctly, because the two shapes can't be told apart from the version
alone. Existing coordinator stores (`~/.dora/`, `redb` is the default `--store`)
must be deleted (the path is named in the error) or bypassed with `--store
memory`. Documented in `Changelog.md`.

Fixes #2471.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VD5Utmx4fE9YFB64fduR8x)_


phil-opp:
Same as noted on #2472: the `ci/circleci: audit` failure is unrelated to this change — this PR only touches `libraries/coordinator-store/src/lib.rs` and `libraries/coordinator-store/src/redb_store.rs`, no dependency manifests, so it can't have introduced a new advisory. This matches the already-tracked RUSTSEC advisories in #2451, #2447, #2446, #2450, #2448, #2449, and #2452.

---

phil-opp:

The core fix — bumping `SCHEMA_VERSION` to 3 — is the right mechanism and is correctly implemented: the `terminal` field was added to the bincode-persisted `DataflowStatus::Failed` without a version bump, so old databases would open successfully and then hit a decode mismatch. Rejecting on version mismatch matches the existing "reject and restart" policy for coordinator scratch state, so no migration path is warranted.

One note on the new test `old_failed_record_without_terminal_field_fails_to_decode`: it encodes a *standalone* `OldDataflowStatus::Failed { error }` and asserts it fails to decode as `DataflowStatus`, with a comment calling this "the exact byte layout written by a pre-#1854 coordinator." In a real stored record `status` is not the trailing field — `DataflowRecord` continues with `daemon_ids`, `generation`, etc. Standalone, decode reliably fails with `UnexpectedEnd` (no bytes left for `terminal: bool`); but embedded in a full record it would instead read the *next* byte (the `daemon_ids` length varint) as `terminal`, and when `daemon_ids` is empty that byte is `0x00` → decodes as `terminal: false` and silently shifts every subsequent field by one byte.

So the real-world premise isn't "old bytes reliably fail to decode" — depending on trailing bytes they can silently mis-decode into a wrong-but-structurally-valid record, which actually *strengthens* the case for the version bump (it's needed precisely because decode is not guaranteed to fail loudly). Consider either tightening the comment, or encoding a full pre-terminal `DataflowRecord` (with empty `daemon_ids`) so the test demonstrates the silent mis-decode rather than a trailing `UnexpectedEnd`. Not blocking.

---

phil-opp:
## Reasoning is sound — blocking only on the missing upgrade note

The diagnosis holds, and the evidence is stronger than the PR states. `SCHEMA_VERSION=2` shipped in **v1.0.0-rc.1** (`145ccce04`, Apr 17) *without* `terminal`; `terminal` landed May 18 (`3d33cf1e6`, #1854) with no bump; **v1.0.0-rc.2** contains it. So two mutually incompatible on-disk formats both label themselves v2, and **both shipped in released artifacts**. bincode is positional, so `#[serde(default)]` cannot fill a missing trailing field. The silent-loss claim checks out: `list_dataflows` (`redb_store.rs:219`) skips undecodable rows with only a `warn!`. Bumping turns rc.1 row-vanishing into a loud `open()` failure — strictly better.

**The blocker is blast radius + no documentation.** `--store` defaults to `redb` (`binaries/cli/src/command/coordinator.rs:31`), `redb-backend` is a default feature (`binaries/cli/Cargo.toml:19`), default path `~/.dora/`, and `create_store` propagates `RedbStore::open(&path)?` — so **`dora coordinator` hard-fails to start for every rc.1/rc.2 user with an existing store**. The error text is actionable, but the PR touches only 2 source files, there is no `CHANGELOG.md`, and `"schema version mismatch"` has zero hits under `docs/`. With rc.2 tagged two days ago, this needs a v1.0.0 release/upgrade note before it lands.

Worth deciding explicitly: rc.2 stores are *fully decodable* by current code and this rejects them anyway. Because v2 is ambiguous you cannot distinguish rc.1-v2 from rc.2-v2 by version alone, so blanket rejection is the only simple correct option (the alternative is per-row try-new-then-old fallback + rewrite). Defensible pre-1.0 — just make it a documented decision rather than an implicit one.

**Test-coverage caveat:** `older_schema_version_is_rejected:1007` is parameterized on `SCHEMA_VERSION - 1`, so it passes identically before and after the bump — it auto-follows the constant and guards nothing new; the description oversells it. The new `old_failed_record_without_terminal_field_fails_to_decode` is a sound characterization test, but nothing pins the deliberate rejection of a decodable rc.2 store — i.e. the actual data-loss decision is untested.

phil-opp:
Thanks for the thorough review — both points addressed:

1. **Upgrade note**: added a `### Breaking` entry to `Changelog.md` (this repo's convention, e.g. the `on_input_closed`/`on_stop` entry right above it) spelling out the rc.1-vs-rc.2 schema-2 collision, the blast radius (`dora coordinator` refuses to start against an existing `~/.dora/` store from either rc), and the workaround (delete the store file or pass `--store memory`).
2. **Test-coverage gap**: added `schema_v2_is_rejected_even_though_this_particular_row_would_decode`, which hardcodes `2u32` (not `SCHEMA_VERSION - 1`) and constructs an actual rc.2-shaped row — independently verified to decode correctly — to pin down that the blanket v2 rejection is deliberate, not just "rejects the immediately-prior version." `older_schema_version_is_rejected` is left as-is; it's still a useful boundary test, just not the one that covers this specific decision.

Also reworked the `#2471` regression test per your earlier automated-review comment: it now encodes a full pre-`terminal` `DataflowRecord` (not a standalone `DataflowStatus`), since `status` isn't the trailing field — empirically this surfaces as an unrelated `InvalidBooleanValue` error on the shifted `uv` field, not anything mentioning `terminal`, which is an even stronger argument for the version bump.

Didn't see anything actionable to decide beyond documenting it — the blanket-rejection choice you outlined (simple + correct, given v2 is irreducibly ambiguous) matches what's implemented, so I went with writing it up rather than changing the approach. Let me know if you'd rather see the alternative (per-row try-new-then-old-shape fallback) instead.

---

phil-opp:

The commits added since my last review — encoding a full pre-`terminal` `DataflowRecord` in the regression test (instead of a standalone `DataflowStatus`), the `### Breaking` `Changelog.md` entry, and `schema_v2_is_rejected_even_though_this_particular_row_would_decode` (hardcoded `2u32`) — don't introduce new issues. The `SCHEMA_VERSION` 2 → 3 bump and blanket rejection of every v2-stamped store (the only safe option given schema 2 is ambiguous between the rc.1 and rc.2 layouts) look correct from my side. No issues found.

---

phil-opp:

The only commit since my previous review (`be4d7f0`) is a typos-check fix (`mis_decodes`/`mis-decodes` → `misdecodes`) with no functional change. It introduces no new issues, so the latest commit looks safe to merge.

---

PR #2474, opened 2026-07-04
fix(daemon): spawn timer tasks for nodes added to a running dataflow

phil-opp:
## Issue

Timer tick tasks are created only in `RunningDataflow::start()`, which the daemon calls on `AllNodesReady` and on the first `Subscribe`. The `AddNode` coordinator handler registers a newly added node's timer inputs into `dataflow.timers` (`binaries/daemon/src/lib.rs`) but never (re-)invokes `start()`.

As a result, a node added to an **already-running** dataflow whose timer input uses an interval that no existing node uses gets no tick-emitting task and is **silently starved of timer events forever**. The failure is intermittent and confusing: if the new node's interval happens to match one already in use, ticks arrive normally.

## Fix

After registering the new node's inputs in the `AddNode` handler, call `dataflow.start()` when the dataflow has already started:

```rust
if dataflow.dataflow_started {
    dataflow.start(&self.events_tx, &self.clock).await?;
}
```

`start()` is idempotent — it skips intervals that already have a handle — so it only spawns tasks for genuinely new intervals. The `dataflow_started` guard keeps `AllNodesReady` as the sole trigger during initial bring-up (when `AddNode` may be used before the dataflow has started).

## Validation

- Added a regression test (`start_spawns_task_for_interval_added_after_first_start`) asserting that a second `start()` spawns a task for an interval added after the first `start()` while leaving the existing handle intact.
- `cargo test -p dora-daemon` — 90 lib tests pass (incl. the new test).
- `cargo fmt --all -- --check` and `cargo clippy -p dora-daemon -- -D warnings` clean.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2265,6 +2265,23 @@ impl Daemon {
                             deploy: None,
                         });
 
+                    // Spawn timer tasks for any interval this node just
+                    // registered. Timer tasks are only ever created in
+                    // `start()`, so a node added to an already-running dataflow
+                    // whose timer input uses an interval no existing node uses
+                    // would otherwise register into `dataflow.timers` but never
+                    // get a tick-emitting task — silently starving that input.
+                    // `start()` is idempotent (it skips intervals that already
+                    // have a handle), so this only spawns tasks for genuinely
+                    // new intervals. Guard on `dataflow_started` so that during
+                    // initial bring-up the readiness path stays the sole
+                    // trigger. Done last, after all state mutations, so a
+                    // spawn/registration failure above never leaves a half-added
+                    // node with live timer tasks.
+                    if dataflow.dataflow_started {
+                        dataflow.start(&self.events_tx, &self.clock).await?;
+                    }
+
                     tracing::info!(
                         %dataflow_id,
                         %node_id,
@@ -3676,8 +3693,9 @@ impl Daemon {
                                         "all nodes are ready, starting dataflow",
                                     )
                                     .await;
+                                // `start()` sets `dataflow_started`; the guard
+                                // above keeps this to a single spawn.
                                 dataflow.start(&self.events_tx, &self.clock).await?;
-                                dataflow.dataflow_started = true;
                             }
                             _ => {}
                         }
@@ -6049,6 +6067,57 @@ mod fault_tolerance_tests {
             env: None,
         };
         RunningDataflow::new(Uuid::nil(), DaemonId::new(None), descriptor)
+    }
+
+    // dora-rs/dora: a node added to an already-running dataflow (via
+    // `AddNode`) may register a timer input on an interval no existing node
+    // uses. `start()` is the only place timer tasks are spawned, so the
+    // `AddNode` handler re-invokes it. This test locks in the property that
+    // makes that safe: `start()` is idempotent for existing intervals yet
+    // still spawns a task for a newly-added one.
+    #[tokio::test]
+    async fn start_spawns_task_for_interval_added_after_first_start() {
+        let mut df = test_dataflow();
+        let clock = Arc::new(HLC::default());
+        let (events_tx, _events_rx) = mpsc::channel(8);
+
+        let first = Duration::from_millis(100);
+        df.timers
+            .entry(first)
+            .or_default()
+            .insert((NodeId::from("a".to_string()), DataId::from("t".to_string())));
+        df.start(&events_tx, &clock).await.unwrap();
+        assert!(df._timer_handles.contains_key(&first));
+
+        // Simulate a node added later that registers a brand-new interval.
+        let second = Duration::from_millis(250);
+        df.timers
+            .entry(second)
+            .or_default()
+            .insert((NodeId::from("b".to_string()), DataId::from("t".to_string())));
+        df.start(&events_tx, &clock).await.unwrap();
+
+        assert!(df._timer_handles.contains_key(&first));
+        assert!(df._timer_handles.contains_key(&second));
+        assert_eq!(df._timer_handles.len(), 2);
+    }
+
+    // dora-rs/dora: the `AddNode` handler guards its re-invocation of
+    // `start()` on `dataflow_started`, so that flag must be set on *every*
+    // start path. It used to be set only at the single-daemon `Subscribe`
+    // readiness call site, leaving it `false` for the whole life of a
+    // distributed dataflow (which starts via the coordinator `AllNodesReady`
+    // path instead) — so a node added later was silently starved of ticks.
+    // `start()` now sets the flag itself; this pins that invariant.
+    #[tokio::test]
+    async fn start_marks_dataflow_started() {
+        let mut df = test_dataflow();
+        let clock = Arc::new(HLC::default());
+        let (events_tx, _events_rx) = mpsc::channel(8);
+
+        assert!(!df.dataflow_started);
+        df.start(&events_tx, &clock).await.unwrap();
+        assert!(df.dataflow_started);
     }
 
     fn test_running_node() -> RunningNode {
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -366,6 +366,17 @@ impl RunningDataflow {
             self._timer_handles.insert(interval, handle);
         }
 
+        // Record that the dataflow has been started. This must hold on *every*
+        // start path — the single-daemon `Subscribe`-readiness path and the
+        // distributed coordinator `AllNodesReady` path both funnel through
+        // here — so the flag lives in `start()` rather than at the call sites.
+        // The `AddNode` handler relies on it to tell "already running, spawn a
+        // task for a freshly-added interval" apart from "still bringing up, let
+        // the readiness path spawn the tasks". Setting it at only one call site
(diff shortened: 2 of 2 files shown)


phil-opp:
## Automated code-review session — summary

This PR is part of a batch of 6 opened by an automated (machine-generated) code-review pass over the repository at `origin/main` (`25bac6b`). Each was verified individually against current `main`, checked for duplication against existing open PRs, and gated with `cargo fmt`/`cargo clippy -D warnings`/tests for the affected crates plus a combined-diff review pass. **All are machine-generated and warrant human review.**

### PRs by category

**Correctness**
- #2474 — `fix(daemon)`: spawn timer tasks for nodes added to a running dataflow (`AddNode` never re-invoked `start()`, so a dynamically-added node with a fresh timer interval got no ticks).
- #2475 — `fix(cli)`: resolve the dataflow once in `dora topic info` (was resolving twice → double coordinator round-trip and a double interactive dataflow prompt that could pick two different dataflows).
- #2476 — `fix(cli)`: stop reporting an all-stopped dataflow as "all healthy" in `dora doctor`.
- #2477 — `fix(daemon)`: drop per-dataflow node results on finish to bound memory (`dataflow_node_results` grew unbounded on a long-lived daemon).

**Optimization**
- #2478 — `perf(daemon)`: drop redundant per-message `NodeId`/`DataId` clones in `send_out` (hot path).

**Documentation**
- #2479 — `docs(mavlink2-bridge)`: replace the stale "scaffold" crate doc with the implemented surface.

### Rejected / skipped findings (one-line reasons)
- **operator `raw.rs` swallows `Event::Error`** — false positive: the runtime `bail!`s on `Event::Error` (`binaries/runtime/src/lib.rs:329`) and never routes it into the operator channel (only `Input`/`InputClosed`/`Stop`/`Reload`), so `RawEvent.error` is never populated.
- **`dora top` clones the node-info Vec each refresh** — real but cold-path (TUI refresh); the hot-path clone reduction in #2478 was prioritized instead.
- **MAVLink outbound frames use `sequence: 0`** — unverifiable from the checkout; depends on rust-mavlink 0.13 `send()` auto-sequencing semantics.
- **`DoraStatus::StopAll` is a public no-op** — a design decision for maintainers, not a clear-cut bug.
- **`FormatDataflowError` stale comment + `hidden > 1` threshold** — trivial comment nit; the threshold is plausibly intentional noise reduction.
- **single-operator edge type-check gap (core `validate.rs`)** — already covered by existing open PR #2462.

### Note on CI
The `audit` (cargo-audit + cargo-deny) check fails on all six PRs due to **pre-existing** advisories on dependencies that none of these PRs touch — `RUSTSEC-2026-0173` (`proc-macro-error2` unmaintained, 2026-06-07) and `RUSTSEC-2017-0008` (`serial` unmaintained). This is a `main`-branch issue, not introduced here, and left for maintainers to address (e.g. `deny.toml` ignore or dep bump). Per repo conventions these PRs don't touch unrelated dependency config.

The full-workspace `cargo test --all` could not complete in the review environment due to a disk-capacity limit (unrelated crates failed to *link* with `No space left on device`, not a test failure); the affected crates (`dora-daemon`, `dora-cli`, `dora-mavlink2-bridge`) were fully tested and pass.

---

phil-opp:
**Correction to the batch summary:** #2477 (`fix(daemon): drop per-dataflow node results on finish`) has been **closed/withdrawn**. Its `dora run`-based e2e jobs (`examples-linux`, `cli-rust`) revealed that the one-line removal in `finish_dataflow` regressed the single-shot `dora run` result-reporting path — `run_inner`'s `mem::take` of the map (`lib.rs:1494`) is consumed at `lib.rs:999`, which errors with *"no node results for dataflow_id"* when the entry was already removed at finish. Details in #2477. The underlying unbounded-growth finding is valid but needs a more careful fix (single-shot vs. reconnecting daemon) and is left for future work.

The remaining 5 PRs — **#2474, #2475, #2476, #2478, #2479** — are independent and unaffected. (The `audit` failure on all of them remains the pre-existing dependency-advisory issue described above.)

---

phil-opp:
No issues found.

I traced the change: after the `AddNode` handler registers the new node's timer intervals into `dataflow.timers` (`binaries/daemon/src/lib.rs`), it now re-invokes `dataflow.start()`. `start()` (`binaries/daemon/src/running_dataflow.rs:302-305`) skips intervals that already have a handle in `_timer_handles`, so the re-invocation is idempotent for existing intervals and spawns tasks only for genuinely new ones. The call is placed after timer registration but before the running node is inserted; since `start()` only reads `self.timers`, the ordering is safe. The `dataflow_started` guard correctly defers to the `AllNodesReady` path during initial bring-up. The regression test exercises the idempotency-plus-new-interval property.

---

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review; no human has verified it. Advisory only.

Following up on the earlier automated pass that reported no issues — I think there's a gap worth a second look.

The new guard relies on `dataflow.dataflow_started` as a proxy for "the dataflow has started", but that flag is only ever set in the local node-subscribe path (`lib.rs:3416-3426`, and grep confirms line 3426 is the *only* assignment of `dataflow_started = true`). The coordinator-driven start path (`DaemonCoordinatorEvent::AllNodesReady` at `lib.rs:1659` → `dataflow.start(...)` at `lib.rs:1683`) calls `start()` but never sets `dataflow_started = true`.

For any distributed dataflow (a node placed on another daemon), `external_nodes` is `true`, so `update_dataflow_status` returns `DataflowStatus::Pending` rather than `AllNodesReady` (`pending.rs:152-166`). The local subscribe branch at `lib.rs:3416` therefore never runs, and the dataflow instead starts via the coordinator `AllNodesReady` handler that doesn't set the flag. Result: in a running multi-daemon dataflow, `dataflow_started` stays `false`, the new `if dataflow.dataflow_started { dataflow.start(...) }` guard is skipped, and a dynamically-added node with a fresh timer interval is still silently starved — i.e. the fix doesn't cover the distributed case, which is the case dora primarily targets.

The unit test doesn't catch this because it sets `df.dataflow_started = true` by hand, bypassing the path asymmetry.

Two possible directions:
- Also set `dataflow_started = true` in the coordinator `AllNodesReady` handler (`lib.rs:1683`), so the flag consistently means "started" across both paths; or
- Gate the `AddNode` re-invocation on a condition that holds in both start paths (e.g. whether timer handles / running nodes already exist) rather than on `dataflow_started`.

Since `start()` is already idempotent, the risk of either approach is low. Worth confirming the intended behavior for dynamically-added timer nodes in multi-daemon deployments before merging.

---

phil-opp:
The `ci/circleci: test-linux` failure on `5225be2` does not appear to be caused by this change. I checked out that exact commit (this branch rebased on current `main`) and ran the daemon suite:

```
cargo test -p dora-daemon
test fault_tolerance_tests::start_spawns_task_for_interval_added_after_first_start ... ok
test result: ok. 101 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

All 101 daemon tests pass, including the new regression test. This PR touches only `binaries/daemon/src/lib.rs`, and every other required check on this commit is green (`clippy`, `fmt`, all `cross-check-*`, `audit`, `typos`, `license`, `unwrap-budget`) — so the full-workspace `test-linux` failure is in an unrelated/flaky test in another crate, not this diff. A re-run (or the merge-queue's retry/bisection) should clear it.

🤖 Verification performed by the agent (machine-generated PR).

---

phil-opp:

The branch was rebased since the last review, but the previously-raised issue is still unfixed, so restating it: the new `AddNode` guard `if dataflow.dataflow_started { dataflow.start(...) }` relies on `dataflow_started`, which is only set in the local node-subscribe path (`binaries/daemon/src/lib.rs:3680`, inside the `DaemonNodeEvent::Subscribe` → `AllNodesReady` branch). The coordinator-driven start path, `DaemonCoordinatorEvent::AllNodesReady` at `binaries/daemon/src/lib.rs:1852`, calls `dataflow.start(...)` but never sets `dataflow_started = true`.

For a distributed dataflow (a node placed on another daemon), `external_nodes` is true, so `update_dataflow_status` returns `Pending` rather than `AllNodesReady`, the local subscribe branch never fires, and the dataflow starts via the coordinator handler that leaves `dataflow_started == false`. Result: in a running multi-daemon dataflow, the new guard is skipped and a dynamically-added node with a fresh timer interval is still silently starved — i.e. the fix doesn't cover the distributed case, which is dora's primary target. The unit test masks this by setting `df.dataflow_started = true` by hand.

Fix direction: either also set `dataflow_started = true` in the coordinator `AllNodesReady` handler (lib.rs:1852) so the flag means "started" in both paths, or gate the `AddNode` re-invocation on a condition that holds in both start paths. Since `start()` is idempotent, either is low-risk.

---

phil-opp:

The current head addresses the distributed-dataflow concern from the earlier reviews: `start()` (`binaries/daemon/src/running_dataflow.rs`) now sets `self.dataflow_started = true` itself, and the redundant assignment at the single-daemon `Subscribe` call site was removed. Since both the `Subscribe`-readiness path and the coordinator `AllNodesReady` path funnel through `start()`, `dataflow_started` is now set on every start path, so the `AddNode` timer-respawn guard fires for dynamically-added nodes in multi-daemon dataflows too. `start()` remains idempotent for existing intervals. No issues found in the latest revision.

---

PR #2475, opened 2026-07-04
fix(cli): resolve the dataflow once in `dora topic info`

phil-opp:
## Issue

`info()` in `binaries/cli/src/command/topic/info.rs` resolved the dataflow **twice**:

- `TopicSelector::resolve` already fetches the dataflow descriptor (via a coordinator `Info` request, and an interactive dataflow picker when the target is ambiguous).
- `info()` then called `selector.dataflow.resolve()` a second time purely to obtain the descriptor for the subscriber scan — discarding the one the first resolve had already fetched.

Consequences when running `dora topic info <topic>` without `-d` while multiple dataflows are running:

- The user is prompted to pick a dataflow **twice**.
- If they pick a **different** dataflow the second time, the WS subscription (keyed on the first resolve's id) and the subscriber list (from the second descriptor) refer to different dataflows, producing an inconsistent report.
- Even with `-d`, it doubles the coordinator round-trips.

## Fix

Add `TopicSelector::resolve_with_descriptor`, which returns the descriptor alongside the topics; `resolve` now delegates to it (discarding the descriptor). `info()` uses `resolve_with_descriptor` to resolve exactly once and reuse the descriptor for the subscriber scan. The other `resolve` callers (`echo`, `hz`, `pub`, `list`) are unchanged.

## Validation

- `cargo build -p dora-cli` and `cargo clippy -p dora-cli -- -D warnings` clean.
- `cargo test -p dora-cli` — all 246 tests pass.
- `cargo fmt --all -- --check` clean.

---

---


phil-opp:
No issues found. I verified the refactor against the base code: `resolve_with_descriptor` returns exactly the same `dataflow_descriptor` that was previously re-fetched by the second `selector.dataflow.resolve()` call, so `info()`'s subscriber scan is unchanged in behavior while the duplicate coordinator round-trip and second interactive dataflow-picker prompt are correctly eliminated. `resolve`'s delegation preserves its existing signature and behavior for the other callers.

---

---

PR #2476, opened 2026-07-04
fix(cli): stop reporting an all-stopped dataflow as "all healthy" in doctor

phil-opp:
## Issue

`check_node_health` in `binaries/cli/src/command/doctor.rs` skips cleanly-stopped nodes when counting, so a dataflow whose nodes have all stopped yields `healthy = degraded = failed = 0` while `total > 0`. The reporting block's final `else if total > 0` arm then printed:

```
Nodes: N total, all healthy
```

— claiming health for nodes that are all stopped. The same "all healthy" wording was also wrong for the mixed case (e.g. 2 healthy + 3 stopped reported `5 total, all healthy`).

## Fix

Extract the message selection into a **pure** `node_health_report` function over a `NodeHealthCounts` struct (now including a `stopped` count), and fold the stopped count into every message so the numbers reconcile with `total`:

- All-stopped dataflow → `Nodes: N total, all stopped`
- Partially-stopped → healthy count plus a `, M stopped` suffix
- Fail/Warn branches → previous wording plus the reconciling stopped suffix

## Validation

- Added unit tests covering all-stopped, mixed healthy+stopped, all-healthy, degraded, failed, and empty cases.
- `cargo test -p dora-cli` — all pass (incl. the 6 new `node_health_tests`).
- `cargo clippy -p dora-cli -- -D warnings` and `cargo fmt --all -- --check` clean.

---
🤖 This PR was generated by Claude (an AI agent) in an automated code-review session. **Machine-generated change — please review carefully before merging.** It has been reviewed by the agent before submission but warrants human review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PxqZqpep1HtBrYurEGRpJw)_

PR #2477, opened 2026-07-04
fix(daemon): drop per-dataflow node results on finish to bound memory

phil-opp:
## Issue

`Daemon::dataflow_node_results` (`binaries/daemon/src/lib.rs`) accumulates one entry per node for every dataflow the daemon runs, but `finish_dataflow` only removed the dataflow from `self.running` — never from this map. The entries were reclaimed solely by the wholesale `mem::take` on daemon exit/reconnect, so a **long-lived daemon** that services many short dataflows (e.g. repeated `dora run` against a persistent daemon) grows this map without bound.

## Fix

`finish_dataflow` already clones the entry into the `DataflowDaemonResult` it reports to the coordinator **before** any cleanup, and nothing reads the map for a dataflow once it has finished. Remove the entry alongside `self.running`:

```rust
self.dataflow_node_results.remove(&dataflow_id);
```

UUID keys are unique, so there is no stale-lookup hazard.

## Validation

- `cargo test -p dora-daemon` — all lib tests pass.
- `cargo clippy -p dora-daemon -- -D warnings` and `cargo fmt --all -- --check` clean.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4210,6 +4210,11 @@ impl Daemon {
             let _ = df.listener_shutdown_tx.send(true);
         }
         self.running.remove(&dataflow_id);
+        // Reclaim the per-node result entries for this dataflow. They were
+        // already consumed into `result` above and have no reader once the
+        // dataflow is no longer running, so keeping them would leak one entry
+        // per node for every dataflow a long-lived daemon ever runs.
+        self.dataflow_node_results.remove(&dataflow_id);
 
         Ok(())
     }


phil-opp:
Closing this PR — CI (`examples-linux`, `cli-rust`) surfaced a **regression that this change introduces**, and my verification confirms it.

**Root cause.** `dataflow_node_results` has a second reader I missed: `run_inner` returns `std::mem::take(&mut self.dataflow_node_results)` (`binaries/daemon/src/lib.rs:1494`), and the single-shot `dora run` path consumes that map at `lib.rs:999`:

```rust
let node_results = match dataflow_results.remove(&dataflow_id) {
    Some(results) => results,
    None if all_nodes_dynamic => BTreeMap::new(),
    None => return Err(eyre::eyre!("no node results for dataflow_id {dataflow_id}")),
};
```

`finish_dataflow` runs **before** `run_inner` returns, so removing the entry there makes this lookup `None` for every non-all-dynamic dataflow → `dora run` fails with *"no node results for dataflow_id"*. Unit tests didn't catch it because they don't exercise the single-shot result-reporting path; the `dora run`-based e2e jobs did.

The underlying finding (this map grows unbounded on a **long-lived reconnecting** daemon) is still valid, but a correct fix has to distinguish the single-shot `dora run` daemon — which needs the entry to survive until `run_inner` returns — from the reconnecting daemon where the growth actually matters. That's more involved than this one-line removal and warrants careful e2e verification, so I'm withdrawing this rather than stacking a speculative fix.

Apologies for the noise. The other PRs in the batch (#2474, #2475, #2476, #2478, #2479) are independent and unaffected.

🤖 Machine-generated review session — this withdrawal was decided by the agent after diagnosing the CI failure.

---

PR #2482, opened 2026-07-05
fix(core): stop reusing a wrong-commit git clone after a failed build

### LINKED ISSUE #2480: Git build: failed checkout in CopyAndFetch/RenameAndFetch leaves a wrong-commit clone that is silently reused

## Summary

In `libraries/core/src/build/git.rs`, `GitFolder::prepare()` cleans up the clone directory when a **`NewClone`** checkout fails, but the **`CopyAndFetch`** and **`RenameAndFetch`** paths do **not**. Because `clone_dir_ready()` treats *any existing directory* as "ready", a half-prepared directory left behind by a failed checkout is later silently reused — producing a build at the **wrong commit** with no error.

## Details

The `NewClone` arm removes the directory on checkout failure:

```rust
// libraries/core/src/build/git.rs (NewClone arm)
match checkout_result {
    Ok(()) => target_dir,
    Err(err) => {
        logger.log_message(LogLevel::Error, format!("{err:?}")).await;
        // remove erroneous clone again
        if let Err(err) = std::fs::remove_dir_all(target_dir) { /* ... */ }
        bail!(err)
    }
}
```

But `CopyAndFetch` and `RenameAndFetch` propagate a checkout failure via `?` with **no cleanup**:

```rust
// CopyAndFetch arm
let repository = fetch_changes(&target_dir, None).await?;
checkout_tree(&repository, &commit_hash)?;   // <-- on failure, target_dir is left behind
target_dir

// RenameAndFetch arm
tokio::fs::rename(&from, &target_dir).await.context("failed to rename repo clone")?;
// ...
let repository = fetch_changes(&target_dir, None).await?;
checkout_tree(&repository, &commit_hash)?;   // <-- on failure, target_dir is left behind
target_dir
```

This matters because reuse is decided purely by directory existence:

```rust
pub fn clone_dir_ready(&self, session_id: SessionId, dir: &Path) -> bool {
    self.prepared_builds
        .get(&session_id)
        .map(|p| p.planned_clone_dirs.contains(dir))
        .unwrap_or(false)
        || dir.exists()          // <-- any leftover directory counts as "ready"
}
```

And the `Reuse` arm performs **no verification** that `HEAD` actually points at the requested `commit_hash`:

```rust
ReuseOptions::Reuse { dir } => {
    logger.log_message(LogLevel::Info, format!("reusing up-to-date {}", dir.display())).await;
    dir   // returned as-is; never checks the checked-out commit
}
```

Note that `GitManager::clear_planned_builds()` (called in `binaries/daemon/src/lib.rs`) only clears the in-memory `prepared_builds` map — it does **not** delete on-disk directories — so the leftover survives across builds.

## Reproduction (RenameAndFetch)

1. Build node at commit **A** with repo `R` → `clone_dir_A` created and correctly checked out to A.
2. Build the same node at commit **B** with `prev_git` pointing at A, where `clone_dir_A` is unused and no longer needed → `RenameAndFetch`: `clone_dir_A` is renamed to `clone_dir_B`, changes are fetched, then `checkout_tree(B)` **fails** (e.g. `B` is a ref the default-branch fetch didn't retrieve, a transient git error, or a ref `revparse_ext` can't resolve). The build aborts, but `clone_dir_B` now exists on disk as a valid repo checked out at **commit A**.
3. Build the node at commit **B** again → `clone_dir_ready()` returns `true` because `clone_dir_B` exists → `ReuseOptions::Reuse` → `prepare()` logs "reusing up-to-date …" and returns the directory. **The node is built from commit A's source while the system believes it is B.**

## Impact

Silent wrong-commit builds after a transient/partial failure. This defeats the reproducibility guarantees of pinned git sources and is hard to diagnose because no error is surfaced.

## Suggested fix

Mirror the `NewClone` cleanup in the `CopyAndFetch` and `RenameAndFetch` arms: on `checkout_tree` (or fetch) failure, `remove_dir_all(&target_dir)` before bailing. Optionally, have the `Reuse` path assert that the repository's `HEAD` resolves to the requested `commit_hash` (when it is a full hash) before reusing, as defense-in-depth against any other source of a stale/partial clone.

## Location

- `libraries/core/src/build/git.rs` — `GitFolder::prepare()` (`CopyAndFetch` and `RenameAndFetch` arms) and `GitManager::clone_dir_ready()`.

keirsalterego:
Fixes #2480.

When a build clones a pinned git source, a failed checkout could leave a clone dir behind that was still sitting on the *old* commit. Nothing cleaned it up, and since `clone_dir_ready()` treats any existing directory as ready, the next build would reuse that leftover and quietly build the wrong commit. No error, no warning, which is the nasty part.

Digging in, I found two things behind it:

1. Only the `NewClone` path removed its dir when a checkout failed. `CopyAndFetch` and `RenameAndFetch` just bubbled the error up with `?` and left the half-prepared dir on disk. I pulled the cleanup into one small helper and ran all three arms through it, so a failed prepare no longer leaves a landmine behind.

2. `Reuse` trusted mere existence and never checked what was actually checked out. As a backstop, before reusing a clone that's pinned to a full commit hash I now verify HEAD actually points at that commit. If it doesn't, I delete the dir and bail loudly so the next build re-clones from scratch. Branch and tag pins get skipped (I can't verify those offline), and the check only runs when the dir is really on disk, so the parallel-build case where a sibling node has only *planned* a dir is left alone.

One thing worth pointing out in the reuse check: if opening the leftover repo fails (a corrupt half-clone), I treat that as a mismatch and remove it rather than propagating a raw error, so we self-heal instead of tripping on the same bad dir every build.

### Testing

- Two regression tests for the copy/rename cleanup: they drive the fetch-failure path and assert the dir is actually gone. On the old code the dir survives, so these fail without the fix.
- Three for the reuse check: a mismatch removes the dir and errors, a matching commit reuses it, and a branch-name pin skips verification.
- Full workspace `cargo test`, `clippy -D warnings`, `fmt --check`, and the unwrap-budget ratchet all pass.

Kept it as two commits so the cleanup (the real fix) and the HEAD-check backstop can be reviewed on their own.


phil-opp:
I reviewed the diff against the surrounding build code and found one issue worth flagging.

**The new Reuse HEAD-check can delete a clone dir that a sibling node is concurrently creating (`--parallel` builds).**

The comment on the reuse check argues the `dir.exists()` guard leaves the parallel case alone: *"parallel builds that have merely planned this dir (not created it yet) are left alone."* But `dir.exists()` does not distinguish "planned but not yet created" from "currently being created by a concurrent sibling."

Concrete scenario for two nodes sharing the same repo + full commit hash:

1. `choose_clone_dir` runs sequentially (serialized by `&mut GitManager`). Node A gets `NewClone` and registers the planned dir; node B then sees it in `planned_clone_dirs` and gets `Reuse { dir, commit_hash }` with the full hash.
2. The returned prepare tasks then run **concurrently** — `binaries/cli/src/command/build/local.rs:104` spawns them into a `JoinSet` with no per-dir locking.
3. Node A's `NewClone` runs `clone_into`, which creates the directory and populates it over time. During that window `dir.exists()` is true but HEAD is not yet at the target commit (or the repo is a partial half-clone).
4. Node B's `Reuse` check fires: `is_full_commit_hash` is true, `dir.exists()` is true, so it opens the repo / reads HEAD. Either the open/peel fails (a partial clone → treated as a mismatch by design) or HEAD isn't at the commit yet → mismatch. B then calls `cleanup_failed_clone(&dir)` → `remove_dir_all` on the directory A is actively cloning into.

The result is a corrupted/destroyed in-progress clone and a spurious build failure, triggered precisely by the common case (multiple nodes pinned to the same full commit hash, parallel build). Before this PR the `Reuse` arm was a read-only no-op, so this active deletion is a new hazard. The sequential daemon path (`binaries/daemon/src/lib.rs:2751`) is unaffected, but the CLI parallel path is.

A safe fix would be to only run the HEAD verification when the dir was reused from a *prior completed build* (i.e. not one this session merely planned), or to synchronize on the dir so the reuse check can't race a concurrent `NewClone`.

Two smaller points: `spawn_blocking(...).await.unwrap()` at the HEAD check turns a `JoinError` into a panic rather than a propagated error, and `cleanup_failed_clone` uses the blocking `std::fs::remove_dir_all` from an async context. Both mirror existing patterns in this file, so they're lower priority than the race above.

---
🤖 _This review was generated by Claude and posted fully automatically, with no human in the loop. Apologies in advance if I've misread any part of the diff — please treat this as one more input, not a verdict._

---

keirsalterego:
the audit job failure is unrelated to this change no dependency files touched here and it's failing on newly-published advisories (incl. the cxx one dated today) that also affect main

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review; no human has verified it. Advisory only.

Thanks for the follow-up commit (`5868b1aa`) — this resolves the parallel-build race I raised earlier. Deciding `verify_commit` inside `choose_clone_dir` (serialized by `&mut GitManager`) means a sibling-owned dir is always in this session's `planned_clone_dirs`, so `planned_this_session` is true and `verify_commit` is `None`; the Reuse arm then never opens or deletes a dir another node may still be cloning into. The HEAD check now only runs against a prior-build leftover on disk, which is exactly the case that needs it. No new issues found.

Two minor points remain (both pre-existing patterns in this file, so low priority): `spawn_blocking(...).await.unwrap()` on the HEAD read turns a `JoinError` into a panic rather than a propagated error, and `cleanup_failed_clone` calls the blocking `std::fs::remove_dir_all` from an async context. Neither is a correctness problem, but they'd be easy hardening if you're already touching this arm.

---

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human in the loop. Advisory only.

Re-reviewed the current head (`dc076b4`) after the force-push. The two minor points from the prior automated review both appear addressed now:

- The HEAD read no longer panics on a `JoinError` — it uses `.await.context("HEAD read task panicked")?`, so the task failure propagates.
- `cleanup_failed_clone` now uses the async `tokio::fs::remove_dir_all` instead of the blocking `std::fs::remove_dir_all`.

I also re-walked the reuse-vs-reclone decision against the surrounding `GitManager` code and it holds up: `clone_dir_ready` returns true for both a sibling-planned dir (in `planned_clone_dirs`) and a prior-build leftover (`dir.exists()` only), and `planned_this_session` cleanly separates them — leftover → `verify_commit: Some(hash)` (HEAD-checked), sibling-owned → `None` (never opened or deleted). Since `choose_clone_dir` is serialized by `&mut GitManager`, a concurrently-owned dir is always already in `planned_clone_dirs` by the time the second node evaluates it, so the earlier parallel-build deletion race can't recur. The HEAD comparison is case-insensitive so a good clone isn't spuriously discarded, and cleanup on the `CopyAndFetch`/`RenameAndFetch` failure paths only ever targets this node's own `target_dir`.

No new issues found.

---
🤖 This is a fully automated review by Claude (no human has reviewed it). It may be wrong — treat it as a starting point, not a verdict.

---

keirsalterego:
Thanks @phil-opp for the review!
All checks are green, clippy and fmt are clean.
It's ready to merge.

/trunk merge

PR #2483, opened 2026-07-05
fix(runtime): compact dropped operator inputs to bound queue memory

phil-opp:
## Issue

`InputBuffer::drop_oldest_inputs` (`binaries/runtime/src/operator/channel.rs`) enforced the per-input queue cap by overwriting evicted events with `None` tombstones (`*event = None`) instead of removing them. Those tombstones are only ever cleared by `pop_front` in `send_next_queued`, which does **not** run while the outgoing rendezvous send (`outgoing = flume::bounded(0)`) is still pending.

When a downstream operator stalls (a slow `on_event`), the input drain loop in `run` keeps calling `add_event` for each new input. Each call pushes one `Some` and converts at most one older entry to `None`. The number of live `Some` events stays capped, but the deque grows by one `Option::None` slot per received input **without bound**.

**Failure scenario:** an operator with a heavy `on_event` (e.g. a vision model) fed by a high-rate camera input under the default `drop_oldest` policy. The live-event count stays at the cap, but tombstones accumulate one per frame → unbounded memory growth → OOM. This defeats the bounded-memory guarantee the drop-oldest policy is supposed to provide. The node-side `Scheduler` (`apis/rust/node/src/event_stream/scheduler.rs`) already removes dropped entries rather than tombstoning.

## Fix

Compact the deque with `self.queue.retain(Option::is_some)` after the drop loop (only when `dropped > 0`), so the total length stays bounded by the effective cap. Compaction is safe: the in-flight send owns a moved-out `Event` (via `send_async`), never an index into the queue, and `VecDeque::retain` preserves FIFO order of the retained `Some` events.

## Validation

- Added a regression test (`drop_oldest_keeps_queue_bounded`) that pushes 1000 inputs at cap 3 and asserts the queue stays at length 3 with no tombstones. It fails without the `retain` line (queue would be ~1000) and passes with it.
- `cargo test -p dora-runtime`, `cargo clippy -p dora-runtime -- -D warnings`, and `cargo fmt --all -- --check` all pass.
- Full workspace `cargo test` / `cargo clippy --all` pass on the combined branch.

---

---

### DIFF
--- a/binaries/runtime/src/operator/channel.rs
+++ b/binaries/runtime/src/operator/channel.rs
@@ -138,7 +138,74 @@ impl InputBuffer {
         }
 
         if dropped > 0 {
+            self.compact();
             tracing::warn!("dropped {dropped} operator inputs because event queue was too full");
         }
+    }
+
+    /// Physically remove the `None` tombstones left by [`Self::drop_oldest_inputs`].
+    ///
+    /// Without this, a stalled operator (slow `on_event`) makes the deque grow
+    /// by one `Option::None` slot per received input forever — the number of
+    /// live `Some` events stays capped, but the tombstones are only ever cleared
+    /// by `pop_front` in `send_next_queued`, which never runs while the outgoing
+    /// send is pending. That defeats the bounded-memory guarantee of the
+    /// drop-oldest policy and leaks until OOM. `retain` preserves the FIFO order
+    /// of the surviving events.
+    fn compact(&mut self) {
+        self.queue.retain(Option::is_some);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn closed_event(id: &str) -> Event {
+        Event::InputClosed {
+            id: DataId::from(id.to_string()),
+        }
+    }
+
+    // The compaction that `drop_oldest_inputs` runs after evicting over-cap
+    // inputs must *physically* remove the `None` tombstones, not just leave them
+    // in place. Otherwise a stalled consumer accumulates one tombstone per
+    // dropped input without bound (they are only ever cleared by `pop_front` in
+    // `send_next_queued`, which does not run while the outgoing send is pending)
+    // — an unbounded memory leak that defeats the drop-oldest policy. This
+    // exercises the compaction primitive directly with the metadata-free
+    // `Event::InputClosed` stand-in (`compact` only distinguishes `Some`/`None`).
+    #[test]
+    fn compact_removes_tombstones_preserving_order() {
+        let mut buffer = InputBuffer::new(BTreeMap::new());
+        buffer.queue = VecDeque::from([
+            None,
+            Some(closed_event("a")),
+            None,
+            None,
+            Some(closed_event("b")),
+            None,
+        ]);
+
+        buffer.compact();
+
+        assert_eq!(
+            buffer.queue.len(),
+            2,
+            "every `None` tombstone must be removed, not merely nulled out"
+        );
+        let ids: Vec<String> = buffer
+            .queue
+            .iter()
+            .map(|e| match e {
+                Some(Event::InputClosed { id }) => id.to_string(),
+                _ => panic!("unexpected queue entry after compaction"),
+            })
+            .collect();
+        assert_eq!(
+            ids,
+            ["a", "b"],
+            "surviving events must keep their FIFO order"
+        );
     }
 }


phil-opp:
## 🤖 Automated code-review pass — summary

This PR is part of a batch opened by an automated Claude review of `origin/main`. All branches were cut from a freshly-fetched `origin/main`, verified individually (fmt + clippy + tests for the affected crate) and once more on the **combined diff** (full workspace `cargo test`, `cargo clippy --all -D warnings`, `cargo fmt --check`, `cargo check --examples`, plus the `/code-review` and `/simplify` gates — both came back clean). Each PR is **machine-generated** and should be reviewed carefully.

### PRs by category

**Correctness**
- #2483 — `fix(runtime)`: compact dropped operator inputs so a stalled operator under `drop_oldest` doesn't leak `None` tombstones without bound (OOM).
- #2484 — `fix(core)`: honor an operator `build:` command in the Python/WASM source-existence checks, matching custom-node / shared-library behavior (fixes false `dora doctor` / `dora system status` failures on unbuilt dataflows).
- #2485 — `fix(download)`: parse `Content-Disposition` filenames that carry trailing parameters (e.g. `filename="model.bin"; size=1000`) instead of writing the artifact under a junk name.

**Simplification**
- #2486 — `refactor(daemon)`: use the existing `DEFAULT_STOP_GRACE` constant in `stop_all` instead of a duplicated magic `10000`.

**Documentation**
- #2487 — `docs(recording)`: document the `dora-recording` public API with a compile-checked round-trip doctest.

### Rejected / not-filed findings (with reasons)

- **`dora logs --local --follow` ignores log rotation** — real bug (follow stops after a rotation), but a correct fix needs inode-level file-identity tracking across renames; a naive re-discovery causes duplicate or missed output. Too risky for a mechanical pass.
- **Timer tick mints two HLC timestamps** (`running_dataflow.rs`) — not anomalous: `spawn/prepared.rs` uses the same double-mint pattern, so it's an intentional convention (metadata vs event-wrapper timestamp); perf gain is negligible.
- **Duplicated byte-size parser** (`core::parse_byte_size` vs `message::ByteSize::from_str`) — genuine duplication, but consolidating narrows the accepted range on 32-bit targets (`usize` vs `u64`) and disturbs large mirrored test suites; low confidence it's wanted.
- **`download` buffers the whole body with no size cap** — a robustness/feature improvement rather than a clear defect, and likely overlaps in-flight download work.
- **Dead commented-out field** in `build/git.rs` — trivial one-liner, not worth a standalone PR.
- **`open_external_mappings` watchdog no-op for distributed dataflows** — verified intentional and documented, not a defect.

### Note on CI

The `Audit (cargo-audit + cargo-deny)` check is red on all of these PRs (and on `main`): it reports RUSTSEC advisories on pre-existing transitive dependencies — most notably `cxx` (RUSTSEC-2026-0202, **published 2026-07-05**), plus `paste`, `number_prefix`, `proc-macro-error2`, and `serial`. None of these PRs add or change any dependency, so this failure is pre-existing/environmental and can only be resolved by a maintainer dependency bump or advisory-allowlist update — out of scope here.

---

---

phil-opp:

The leak is real: `drop_oldest_inputs` evicts over-cap entries by setting `*event = None`, and those `None` tombstones are only ever removed by `pop_front` in `send_next_queued`. That path doesn't run while the outgoing rendezvous send future is still pending, so under a stalled operator (slow `on_event`) the deque grows by one tombstone per received input without bound — the live `Some` count stays capped, but total memory leaks until OOM. Adding `self.queue.retain(Option::is_some)` after the drop loop fixes this.

The compaction is safe: `retain` preserves FIFO order, nothing indexes into `queue` (both `send_next_queued` and `drop_oldest_inputs` only `pop_front`/`iter_mut`), and `send_next_queued` moves the event out via `pop_front` into `send_async(next)`, so the in-flight future owns it and holds no borrow into the deque being compacted. After compaction the total length is bounded by the sum of per-id caps, which also keeps the per-event `retain` cost at O(cap). The regression test (1000 inputs at cap 3) exercises exactly this scenario.

---

---

phil-opp:
## Merge-queue failure is a cross-PR batch conflict, not a defect in this PR

The `Test (ubuntu-latest)` failure that removed this PR from the queue is a compile error in the **batched** test PR, caused by an interaction with another in-flight PR — not a problem with this change against `main`.

The batch build failed with:

```
error[E0432]: no `ArrowTypeInfo` in `dora_message::metadata`
error[E0061]: Metadata::new takes 1 argument but 2 were supplied
   note: pub fn new(timestamp: uhlc::Timestamp) -> Self
```

That signature (`Metadata::new` with a single argument, and no `ArrowTypeInfo` in `dora_message::metadata`) does **not** exist on `main` — current `main` still has `Metadata::new(timestamp, type_info)` and `ArrowTypeInfo` in `metadata`. It only appears because the merge queue co-batched this PR with an in-flight refactor that drops the `type_info` parameter from `Metadata`. That refactor updated all existing call sites but naturally couldn't update the **new** unit test added here, so the combined tree fails to compile.

This PR compiles and tests cleanly against current `main` (`cargo test -p dora-runtime`, `clippy`, `fmt` all green), and GitHub reports it as mergeable (no conflict). Two clean resolutions:

1. Land this PR **before** the `Metadata` refactor — no changes needed.
2. If the refactor lands first, the new `drop_oldest_keeps_queue_bounded` test needs a one-line update to the new metadata constructor. I'm happy to push that update once the refactor is on `main` — just let me know (or it'll come through on the next re-queue that surfaces the new signature).

The production change itself (`self.queue.retain(Option::is_some)`) is independent of the `Metadata` API and is unaffected either way.

---

---

phil-opp:
/trunk merge

Re-queuing: this PR is still correct against current `main` (`caf62a0`) — `Metadata::new` there still takes `(timestamp, type_info)` and `ArrowTypeInfo` is still in `dora_message::metadata`, so the branch compiles and tests clean. The earlier removal was a co-batch compile conflict with the not-yet-merged `Metadata` refactor, not a defect here. The other four PRs from this review pass have now merged and the queue has drained, so this should get a clean batch. If it's co-batched with the `Metadata` refactor again and fails on that same `Metadata::new` arity error, I'll update the new test to the refactored constructor once that lands on `main`.

---

phil-opp:

No issues found in this diff.

Traced the fix against the current source:
- The leak is real. `drop_oldest_inputs` evicts over-cap entries by setting `*event = None`, and those tombstones are only cleared by `pop_front` in `send_next_queued`. That path is starved while the outgoing rendezvous send (`flume::bounded(0)`) is pending, so under a stalled operator the deque grows by one `None` per received input without bound — live `Some` count stays capped, but memory leaks until OOM.
- `self.queue.retain(Option::is_some)` runs only when `dropped > 0`, preserves FIFO order, and touches nothing in-flight (the event being sent was already moved out via `pop_front` into `send_async`, so no borrow into the deque). After compaction the total length is bounded by the sum of per-id effective caps, keeping the `retain` cost O(cap). This matches the node-side `Scheduler`, which already removes dropped entries rather than tombstoning.
- The regression test is valid: `DropOldest`'s `effective_cap(3) == 3`, and pushing 1000 inputs at cap 3 keeps `queue.len() == 3` only with the compaction — without it the deque grows ~1 slot per push. The `Metadata::new(timestamp, type_info)` call matches the current `main` signature, so it compiles.

Note (not a defect in this diff): the prior merge-queue failures were a co-batch compile conflict with an in-flight `Metadata` refactor, already documented above — the production change here is independent of that API.

---

phil-opp:
The re-queue keeps failing on the same root cause: this PR's new `drop_oldest_keeps_queue_bounded` test constructs an `Event::Input`, which needs a `Metadata` built via `Metadata::new(timestamp, type_info)` / `ArrowTypeInfo`. A queued-ahead PR refactors that API (drops `type_info`), so any batch that puts the refactor ahead of this one fails to compile the test. This PR is correct against current `main` (`c244bb5` still has the 2-arg signature) — it just can't co-merge behind that refactor.

Three ways forward, your call:

1. **Land this ahead of the `Metadata` refactor** — zero changes needed.
2. **I decouple the test from `Metadata`** — I can rework it to exercise the tombstone-compaction primitive directly (no `Event::Input`/`Metadata` construction), so it compiles against both the current and refactored API and merges regardless of queue order. Slightly narrower coverage, but unblocks immediately. Say the word and I'll push it.
3. **I update the one test line** to the new constructor once the refactor lands on `main`.

The production fix itself (`self.queue.retain(Option::is_some)`) is independent of the `Metadata` API and unaffected either way.

---

phil-opp:
/trunk merge

Pushed option 2 (force-pushed a rebase onto current `main`, `c244bb5`). The branch no longer depends on the `Metadata`/`ArrowData` construction API, so it now compiles regardless of where the `Metadata` refactor sits in the queue:

- The production fix is unchanged in behavior — I just extracted the compaction into a small `compact()` method (`retain(Option::is_some)`, order-preserving) called from `drop_oldest_inputs`.
- The regression test now exercises `compact()` directly using the metadata-free `Event::InputClosed` as stand-in payload (compaction only distinguishes `Some`/`None`), instead of constructing an `Event::Input` with `Metadata::new(..)`. That was the only thing coupling this PR to the refactored API.

Trade-off: the test now covers the compaction primitive rather than the full 1000-input integration path, but it directly guards the leak (tombstones must be physically removed, not just nulled). Verified locally: `cargo test -p dora-runtime`, `clippy -D warnings`, and `fmt --check` all green.

The earlier force-push replaced only a main-merge commit (no unique changes) with this rebase, so no work was lost.

---

PR #2484, opened 2026-07-05
fix(core): honor operator build command in Python/WASM source checks

phil-opp:
## Issue

`check_dataflow` (`libraries/core/src/descriptor/validate.rs`) skips the on-disk source existence check when a node/operator declares a `build:` command, because the artifact is produced by that build. This guard is present for **custom nodes** (`custom.build.is_some()`) and **shared-library operators** (`operator_definition.config.build.is_some()`), but was **missing** from the Python and WASM operator arms:

```rust
OperatorSource::Python(python_source) => {
    ...
    } else if !working_dir.join(path).exists() {   // no build.is_some() skip
        bail!("no Python library at `{path}`");
    }
}
OperatorSource::Wasm(path) => {
    ...
    } else if !working_dir.join(path).exists() {   // same gap
        bail!("no WASM library at `{path}`");
    }
}
```

**Failure scenario:** a runtime node with a Python (or WASM) operator whose source artifact is generated by its `build:` command. Validation that runs *before* the build — e.g. `dora doctor` and `dora system status`, which call `Descriptor::check()` on an unbuilt dataflow — bails with `no Python library at ...` / `no WASM library at ...`, while the equivalent custom-node and shared-library cases pass. The four source kinds behaved inconsistently.

## Fix

Add the same `else if operator_definition.config.build.is_some()` skip (with the existing `info!("skipping path check for operator with build command")` log) to both the Python and WASM arms, so all four source kinds behave consistently.

## Validation

- Added two regression tests via the WASM arm: one asserts a missing source with a `build:` command passes `check_dataflow`, the other asserts a missing source **without** a build command is still rejected with `no WASM library`. The tests use the WASM arm deliberately so they don't require an installed Python `dora-rs` runtime (`check_python_runtime` spawns a Python subprocess); the Python arm now carries the identical guard.
- `cargo test -p dora-core`, `cargo clippy -p dora-core -- -D warnings`, and `cargo fmt --all -- --check` all pass.

---

---


phil-opp:

The Python and WASM operator-source arms were missing the `build.is_some()` guard that the `SharedLibrary` and custom-node arms already carry; adding it to both makes all four source kinds consistent, so a runtime operator whose artifact is produced by a `build:` command is no longer rejected by the on-disk existence check before that build runs. The regression tests exercise the WASM arm.

---

---

PR #2485, opened 2026-07-05
fix(download): parse Content-Disposition filename with trailing params

phil-opp:
## Issue

`get_filename` (`libraries/extensions/download/src/lib.rs`) took everything after the first `filename=` to the end of the header and only stripped surrounding quotes:

```rust
filename.split("filename=").nth(1).map(|n| n.trim_matches('"').to_string())
```

Real `Content-Disposition` headers routinely append more parameters. For `attachment; filename="model.bin"; size=1000`, `nth(1)` yields `"model.bin"; size=1000`, and `trim_matches('"')` strips only the leading quote (the string no longer ends in one), producing the literal filename `model.bin"; size=1000`. The artifact is then written under that junk name (`target_dir.join(filename)`), so any later lookup by the expected name fails.

## Fix

Extract a testable helper `parse_content_disposition_filename(header: &str) -> Option<String>` that:
- cuts the value at the closing quote (quoted form) or the next `;` (token form),
- trims whitespace/quotes,
- returns `None` for an empty value so the existing URL-path fallback still applies.

A `;` inside a quoted filename is preserved. The RFC 5987 `filename*=` extended form is left undecoded and falls through (unchanged from before). The downstream sanitize block (path-component stripping, null-byte / >255-length rejection) is untouched.

## Validation

- Added unit tests covering: quoted with/without trailing params, unquoted with trailing params, semicolon inside quotes, and empty/missing/`filename*=` → `None`.
- Verified against the old behavior: every input the old code handled correctly produces identical output; the previously-broken multi-parameter case is now correct.
- `cargo test -p dora-download`, `cargo clippy -p dora-download -- -D warnings`, and `cargo fmt --all -- --check` all pass.

---

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -5,20 +5,31 @@ use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
 use tokio::io::AsyncWriteExt;
 
-fn get_filename(response: &reqwest::Response) -> Option<String> {
-    let raw_name = if let Some(content_disposition) = response.headers().get("content-disposition")
-    {
-        if let Ok(filename) = content_disposition.to_str() {
-            filename
-                .split("filename=")
-                .nth(1)
-                .map(|n| n.trim_matches('"').to_string())
-        } else {
-            None
-        }
+/// Extract the `filename` parameter from a `Content-Disposition` header value.
+///
+/// Handles headers that carry additional parameters after the filename, e.g.
+/// `attachment; filename="model.bin"; size=1000`, and both quoted and unquoted
+/// forms. Returns `None` when no non-empty `filename=` value is present (the
+/// RFC 5987 extended `filename*=` form is not decoded and falls through).
+fn parse_content_disposition_filename(header: &str) -> Option<String> {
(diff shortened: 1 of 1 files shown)


phil-opp:

The new helper correctly cuts quoted values at the closing quote and token values at the next `;`, so trailing parameters like `; size=1000` no longer leak into the saved filename. The edge cases hold up too: the `filename*=` extended form does not contain a literal `filename=` substring, so it falls through to `None` and preserves the URL-path fallback; unterminated quotes degrade gracefully; and an empty value returns `None`.

---

---

phil-opp:
/trunk merge

Re-queuing: the removal is collateral from a shared batch, not a fault in this PR. Batch PR #2569 failed with 9 of 11 `fault-tolerance-e2e` **restart** tests failing (`max_restarts_*`, `restart_recovers_from_failure`, `node_restarted_is_delivered_to_downstream`, etc.) — the process-restart subsystem, which this change does not touch. This PR only modifies `Content-Disposition` filename parsing in the `download` crate (operator/artifact fetch), and it passed Trunk's isolated batch bisection earlier. The mass fault-tolerance failure pattern indicates a poison PR elsewhere in the batch or runner resource/timing overload (the queue has also been hitting `No space left on device` infra failures).

---

PR #2490, opened 2026-07-06
fix(core): extend checkout_tree rev-spec denylist to ~, @, {, }

### LINKED ISSUE #2481: Git build: checkout_tree rev-spec denylist is incomplete (allows HEAD~N, @{...}, etc.)

## Summary

`checkout_tree()` in `libraries/core/src/build/git.rs` guards against arbitrary git rev-spec expressions with a small denylist, but the denylist is incomplete. Its own comment states the intent is to "only allow hex commit hashes and branch/tag names," yet several rev-spec forms slip through and are resolved by `revparse_ext`.

## Details

```rust
fn checkout_tree(repository: &git2::Repository, commit_hash: &str) -> eyre::Result<()> {
    // Reject arbitrary rev-spec expressions; only allow hex commit hashes and branch/tag names.
    if commit_hash.contains("..") || commit_hash.contains(':') || commit_hash.contains('^') {
        eyre::bail!(
            "invalid commit reference '{commit_hash}': rev-spec expressions are not allowed"
        );
    }
    let (object, reference) = repository
        .revparse_ext(commit_hash)
        .context("failed to parse ref")?;
    // ...
}
```

The denylist blocks `..`, `:`, and `^`, but does **not** block other rev-spec navigation/selection operators that `revparse_ext` happily resolves, including:

- `~` — e.g. `HEAD~3`, `main~10` (ancestor navigation)
- `@{...}` — e.g. `main@{upstream}`, `HEAD@{yesterday}`, `HEAD@{2}` (reflog / date / upstream selectors)
- bare `@` as an alias for `HEAD`

So a value such as `rev: HEAD~1` or `rev: main@{yesterday}` in a dataflow descriptor passes the guard and resolves to a commit that is *not* the intended pinned hash/branch/tag — exactly the class of input the comment says should be rejected.

## Impact

- The validation does not enforce its stated contract, so a build can resolve to an unintended commit (reproducibility / correctness).
- A denylist approach here is inherently fragile: any rev-spec operator not explicitly enumerated bypasses it.

## Suggested fix

Prefer an **allowlist** over a denylist. For example:
- Accept a value only if it matches a hex object id (`^[0-9a-fA-F]{7,40}$`), **or**
- It is a valid branch/tag name that resolves via `revparse_single`/`revparse_ext` **and** the resulting reference is confirmed to be a local/remote branch or tag (not an arbitrary rev-spec).

At minimum, extend the denylist to also reject `~`, `@`, `{`, and `}` so the current guard matches its documented intent.

## Location

- `libraries/core/src/build/git.rs` — `checkout_tree()`.

phil-opp:
Fixes #2481.

## Summary

`checkout_tree()` in `libraries/core/src/build/git.rs` guards against arbitrary git rev-spec expressions with a denylist, whose comment states the intent is to "only allow hex commit hashes and branch/tag names." The denylist blocked `..`, `:`, and `^`, but not other rev-spec navigation/selection operators that `revparse_ext` happily resolves:

- `~` — e.g. `HEAD~3`, `main~10` (ancestor navigation)
- `@{...}` — e.g. `main@{upstream}`, `HEAD@{yesterday}`, `HEAD@{2}` (reflog / date / upstream selectors)
- bare `@` as an alias for `HEAD`

So a value such as `rev: HEAD~1` or `rev: main@{yesterday}` in a dataflow descriptor passed the guard and resolved to a commit that is *not* the intended pinned hash/branch/tag — exactly the class of input the comment says should be rejected.

## Fix

Extend the denylist to also reject `~`, `@`, `{`, and `}`, per the issue's suggested minimal fix.

## Test plan

- [x] Added `rejects_rev_spec_navigation_operators`, exercising `checkout_tree` against a real two-commit repo with `HEAD~1`, `main~1`, `HEAD@{1}`, `main@{upstream}`, `@`, plus the previously-covered `HEAD^`, `main..HEAD`, `HEAD:foo` — all rejected.
- [x] Added `accepts_branch_name_and_commit_hash` to confirm legitimate branch names and hex commit hashes still resolve correctly.
- [x] Verified both new tests fail on the pre-fix code and pass after the fix.
- [x] `cargo test -p dora-core --features build` — all 263 tests pass
- [x] `cargo clippy -p dora-core --features build --all-targets -- -D warnings`

---
_Generated by [Claude Code](https://claude.ai/code/session_01FRbGP7i2A7p3vW8A6ULFgq)_

### DIFF
--- a/libraries/core/src/build/git.rs
+++ b/libraries/core/src/build/git.rs
@@ -465,7 +465,19 @@ async fn fetch_changes(
 
 fn checkout_tree(repository: &git2::Repository, commit_hash: &str) -> eyre::Result<()> {
     // Reject arbitrary rev-spec expressions; only allow hex commit hashes and branch/tag names.
-    if commit_hash.contains("..") || commit_hash.contains(':') || commit_hash.contains('^') {
+    // This must stay a denylist of every rev-spec operator `revparse_ext` understands, not just
+    // the ones above: `~`/`@` enable ancestor/reflog/upstream navigation (e.g. `HEAD~3`,
+    // `main@{upstream}`) just as much as `..`, `:`, and `^` do. We reject `@`/`{`/`}` wholesale
+    // rather than only the `@{…}` sequence; that also turns away the rare valid ref name that
+    // embeds one of these characters, but a hard error is the safe failure mode for this guard.
+    if commit_hash.contains("..")
+        || commit_hash.contains(':')
+        || commit_hash.contains('^')
+        || commit_hash.contains('~')
+        || commit_hash.contains('@')
+        || commit_hash.contains('{')
+        || commit_hash.contains('}')
+    {
         eyre::bail!(
             "invalid commit reference '{commit_hash}': rev-spec expressions are not allowed"
         );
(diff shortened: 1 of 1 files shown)


phil-opp:
The `Audit (cargo-audit + cargo-deny)` failure is pre-existing and unrelated to this change — this PR only touches `libraries/core/src/build/git.rs` (no `Cargo.toml`/`Cargo.lock` changes). The failure is newly-published RUSTSEC advisories in transitive dependencies (`quick-xml` RUSTSEC-2026-0194/0195, `cxx` RUSTSEC-2026-0202, `bincode` unmaintained, etc.), already tracked separately in #2447/#2451/#2452 as blocked dependency work. No action needed here.

---

phil-opp:

The extended denylist covers the remaining rev-spec navigation operators that `revparse_ext` would otherwise interpret (`~` ancestor, `@` reflog/`@{...}` upstream, and the `{`/`}` used by those specs), while plain hex hashes and branch/tag names still pass through unchanged. Conservatively rejecting the rare ref that legitimately contains `@` is reasonable hardening rather than a regression, and the added accept/reject unit tests match the intended behavior.

---

---

PR #2491, opened 2026-07-06
fix(coordinator): actually stop partially-finished multi-daemon dataflows

phil-opp:
## Summary

`ControlRequest::Stop` and `StopByName` short-circuit and return a cached `DataflowStopped` reply whenever `dataflow_results` contains an entry for the dataflow:

```rust
ControlRequest::Stop { dataflow_uuid, .. } => {
    if let Some(result) = dataflow_results.get(&dataflow_uuid) {
        // reply DataflowStopped and `continue` — never calls stop_dataflow
    }
    let dataflow = stop_dataflow(...).await;
```

But `dataflow_results` is **not** finished-history only. The `DataflowFinishedOnDaemon` handler inserts one entry **per daemon** as each finishes, while the dataflow stays in `running_dataflows` until the *last* daemon finishes. This is documented explicitly on `cap_dataflow_results`:

> `dataflow_results` is not pure history: a partially-finished multi-daemon dataflow accumulates one entry per daemon while it is still running […]

## The bug

In the window after one daemon's nodes exit (e.g. a finite/source node on machine A) but while another daemon still runs (e.g. an infinite node on machine B), `dataflow_results.get(uuid)` is `Some`. So `dora stop <uuid>` returns `DataflowStopped` with a partial result and **never sends `StopDataflow` to the still-running daemon B** — its nodes run forever.

It is also inconsistent with `ControlRequest::Check`, which keys off `running_dataflows` and still reports the dataflow as `DataflowSpawned` (running) in the same window.

## Fix

Guard the cached-result short-circuit with `!running_dataflows.contains_key(&dataflow_uuid)` (applied to both `Stop` and `StopByName`), so it only fires once the dataflow has *fully* finished — matching `Check` and `cap_dataflow_results`. A genuinely finished dataflow is removed from `running_dataflows`, so its cached result is still returned exactly as before.

## Validation

- `cargo test -p dora-coordinator` — all pass (100 + 22 + 4 tests)

Verified against current `origin/main`.

---
🤖 This PR was generated automatically by Claude (Anthropic) as part of an automated code-review pass. A human should review before merging.

https://claude.ai/code/session_01J8d32tkjdFsaDxnse5By9t

---
_Generated by [Claude Code](https://claude.ai/code/session_01J8d32tkjdFsaDxnse5By9t)_


phil-opp:

This is the summary for a machine-generated review pass by Claude (Anthropic). Each PR below was verified against current `origin/main` (fmt + clippy + affected-crate tests per branch). **All are machine-generated — please review before merging.**

### PRs opened this run

**Correctness**
- #2491 — `fix(coordinator)`: actually stop partially-finished multi-daemon dataflows. `Stop`/`StopByName` short-circuited on any `dataflow_results` entry, but that map accumulates one entry *per daemon* while the dataflow is still running, so `dora stop` became a silent no-op (leaving other daemons' nodes alive) in the staggered-exit window.
- #2492 — `fix(tracing)`: send OTLP metrics to the configured `DORA_OTLP_ENDPOINT`. The metric exporter ignored the endpoint and always defaulted to `localhost:4317`, so metrics silently diverged from traces when a remote collector was configured.

**Optimization**
- #2494 — `perf(ros2-bridge)`: resolve the struct message once per deserialize. Eliminates a duplicate `HashMap` lookup + throwaway allocation that ran once per element of every `sequence<struct>`/`struct[N]` on the ROS2→Arrow hot path.

**Simplification / dead code**
- #2493 — `refactor(coordinator)`: remove the dead `NodeMetricsExpire` event (no producer anywhere; behavior already covered by the `expire_stopped_nodes()` sweep).

- **operator `Event::Error` silently dropped** (`apis/rust/operator/src/raw.rs`): `RawEvent.error` is populated by the runtime but never read by `dora_on_event`, so runtime error events are dropped for Rust operators. *Rejected:* a proper fix requires adding a public `Event` variant and deciding error-delivery semantics — an API/design change too significant for an autonomous mechanical PR. Better raised as an issue.
- **memory-pool `cleanup_all` over-reports "Successfully released"** (`libraries/extensions/memory-pool/src/lib.rs`): a non-`NotFound` unlink failure is only `warn!`-logged and counted as released. *Rejected:* low severity (segment leaks regardless); an honest fix means changing `free_shared_memory`'s contract, which ripples into the normal `free_memory_pool` path — risk outweighs the reporting nicety.
- **mavlink `?proto=` handling** (`mavlink2-bridge/src/transport.rs`): an unrecognized `proto` value silently falls back to V2, and serial ignores `?proto=` entirely. *Rejected:* behavioral/UX choice, low severity.
- **ros2-bridge `WStringVisitor` lacks the `MAX_SEQUENCE_ELEMENTS` cap** used by other sequence paths. *Rejected:* allocation is already bounded by remaining input; defense-in-depth only.
- **ros2-bridge dead `visit_unit`/`visit_none`** returning length-0 arrays (latent panic if ever reached). *Rejected:* unreachable dead code, low value.
- **telemetry `span_store` truncation exceeds `MAX_FIELD_VALUE_BYTES` by 3** (the `"..."` marker). *Rejected:* cosmetic.
- **scheduler.rs misleading comment** ("BTreeMap Ordering" for a `VecDeque` LRU). *Rejected:* trivial comment-only.
- Minor duplication in `arrow-convert` string extraction and duplicate OTLP meter-provider constructors. *Rejected:* low-value churn.

### ⚠️ Note on the failing `Audit` CI check

The `Audit (cargo-audit + cargo-deny)` job is **red on all four PRs and on `main`** because of **`RUSTSEC-2026-0202` (`cxx 1.0.194`, published 2026-07-05)** plus the long-standing `serial` unmaintained advisory (`RUSTSEC-2017-0008`). None of these PRs touch `Cargo.toml`/`Cargo.lock`, `cxx`, or `serial` — this is a pre-existing, repo-wide advisory failure, **not** introduced by any of these changes. The remedy (bump `cxx` or add a `deny.toml` ignore) is a separate maintenance change that belongs on `main` once, so it's intentionally left out of these review PRs.

---

https://claude.ai/code/session_01J8d32tkjdFsaDxnse5By9t

---

phil-opp:
**Automated review by Claude** — _this is a fully automated review with no human in the loop; please verify independently before acting on it._

I reviewed the fix and it looks correct to me.

- Guarding the cached-result short-circuit with `!running_dataflows.contains_key(&dataflow_uuid)` matches the documented semantics of `dataflow_results` (accumulates one entry per daemon while the dataflow is still running) and aligns `Stop`/`StopByName` with `Check`.
- No regression on the genuinely-finished path: a fully-finished dataflow has been removed from `running_dataflows`, so the guard passes and the cached `DataflowStopped` reply is still returned exactly as before.
- The partially-finished case now falls through to `stop_dataflow(...)`, which is what actually sends `StopDataflow` to the still-running daemons.
- Same guard applied consistently to both handlers; the let-chain (`&& let Some(...)`) is valid on the pinned toolchain.

---

phil-opp:
Superseded by #2614. The Stop/StopByName guard here is identical to #2614's (only the comment prose differs), but #2614 also fixes the `List` handler so a partially-finished dataflow no longer double-lists as both Running and Finished, and it ships a test. Merging both would conflict for no gain.

PR #2492, opened 2026-07-06
fix(tracing): send OTLP metrics to the configured DORA_OTLP_ENDPOINT

phil-opp:
## Summary

`TracingBuilder::with_otlp_tracing` reads the OTLP endpoint from `DORA_OTLP_ENDPOINT` (falling back to `DORA_JAEGER_TRACING`) and passes it to the **span** exporter via `.with_endpoint(endpoint)` (see `telemetry.rs::init_tracing`).

The **metric** exporter built by `init_meter_provider`, however, took no endpoint:

```rust
let exporter = MetricExporter::builder()
    .with_tonic()
    .build()                       // <-- no .with_endpoint(...)
    .expect("Failed to create metric exporter");
```

so it always used the OTLP gRPC default (`http://localhost:4317`).

## The bug

Pointing `DORA_OTLP_ENDPOINT` at a remote collector (e.g. `http://collector:4317`) sent **traces** there while **metrics** silently went to `localhost` and were dropped when nothing listened locally — with no indication that the two had diverged.

## Fix

Thread the endpoint into `init_meter_provider(endpoint: &str)` and apply `.with_endpoint(endpoint)` to the metric exporter so metrics and traces target the same collector. The only caller (`with_otlp_tracing`) already has `endpoint` in scope. Also imports `WithExportConfig` (same trait `telemetry.rs` already uses for the span exporter).

## Validation

- `cargo fmt -p dora-tracing -- --check` — clean
- `cargo clippy -p dora-tracing -- -D warnings` — clean
- `cargo test -p dora-tracing` — all pass

The endpoint plumbing is verified by inspection (exercising a live OTLP collector isn't unit-testable here); the change mirrors the existing, working span-exporter path.

Verified against current `origin/main`.

---
🤖 This PR was generated automatically by Claude (Anthropic) as part of an automated code-review pass. A human should review before merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01J8d32tkjdFsaDxnse5By9t

---
_Generated by [Claude Code](https://claude.ai/code/session_01J8d32tkjdFsaDxnse5By9t)_

PR #2487, opened 2026-07-05
docs(recording): document the public API with a compile-checked doctest

phil-opp:
## Issue

The `dora-recording` crate (`libraries/recording/src/lib.rs`) exposed `RecordingWriter`, `RecordingReader`, and their public methods (`new`, `write_entry`, `finish`, `flush`, `open`, `header`, `next_entry`) with no documentation and no usage example, unlike the sibling crates in the workspace. A reader had to reverse-engineer the write → finish → open → next_entry lifecycle from the source.

## Fix

- Added a crate-level overview and a **runnable, compile-checked doctest** that performs a full round trip (write an entry into an in-memory buffer, then read the header and entry back).
- Added short doc comments to the public `RecordingWriter` / `RecordingReader` methods.

Docs only — no behavior change.

## Validation

- `cargo test -p dora-recording --doc` passes (the doctest compiles and runs).
- `cargo clippy -p dora-recording -- -D warnings` and `cargo fmt --all -- --check` pass.

---
🤖 This pull request was generated by Claude (an AI assistant) as part of an automated code-review pass. It is **machine-generated** and should be reviewed carefully before merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01GkeKgBFLZeS6iM6cJT745E)_

PR #2453, opened 2026-07-02
fix(message): correct mislabeled signal names in node-exit diagnostics

phil-opp:
## Summary

`NodeError`'s `Display` impl (`libraries/message/src/common.rs`) maps a Unix termination-signal number to a name for user-facing exit diagnostics. Two entries were wrong:

- `22 => "SIGABRT"` — a duplicate of `6` (which is already correctly SIGABRT). On Linux `signal(7)`, signal 22 is **SIGTTOU**.
- `23 => "NSIG"` — `NSIG` is a signal-count sentinel, not a deliverable signal. On Linux, signal 23 is **SIGURG**.

A node actually killed by SIGTTOU or SIGURG was therefore reported with a wrong or nonsensical cause (`exited because of signal SIGABRT` / `... signal NSIG`).

## Fix

- Correct `22 => SIGTTOU` and `23 => SIGURG`.
- Extract the lookup into a small, testable `signal_name(i32) -> Cow<'static, str>` helper (the `Display` body is otherwise unchanged).
- Add a regression unit test pinning the numbering (`signal_names_match_linux_numbering`).

## Validation

- `cargo test -p dora-message` (incl. new test) ✅
- `cargo clippy -p dora-message -- -D warnings` ✅
- `cargo fmt --all -- --check` ✅

---

⚠️ **Machine-generated PR.** This change was authored by Claude (an AI agent) during an automated codebase review. It has been verified as described above but should receive human review before merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01N42W9zv3r7PXcEuox4Wqx1)_

PR #2500, opened 2026-07-07
fix(message): validate node_id recovered from log fields

phil-opp:
## Issue

When a `LogMessage` lacks a typed `node_id` but carries one in its untrusted `fields` map, `LogMessage::from(LogMessageHelper)` (`libraries/message/src/common.rs`) promoted it to the typed field via `.map(NodeId)` — the raw tuple constructor, which **bypasses `validate_node_id`**.

Every sibling field recovered in the same function parses defensively:
- `build_id` / `dataflow_id` via `Uuid::parse_str`
- `daemon_id` via `DaemonId::from_display_str`

`node_id` was the odd one out. The `NodeId` newtype invariant exists specifically to forbid `/` and dot-segments, because node ids are joined into filesystem paths (e.g. `.dora/python-envs/<node.id>`) where `.`/`..` could traverse parent directories (see the doc comment on `validate_node_id` in `id.rs`). Wrapping an unvalidated string produced a `NodeId` that:
- no real (validated) node can ever equal — so a `node_filter` built from it silently never matches, and
- re-introduced the path-traversal characters the newtype was designed to reject.

## Fix

Parse through `FromStr` (`id.parse::<NodeId>().ok()`) so an invalid id in `fields` yields `None` instead of a malformed `NodeId`. This matches the defensive-parse pattern already used by the other recovered fields. One line changed; the only production call site of `.map(NodeId)` in the crate.

## Validation

- New regression test `node_id_recovered_from_fields_is_validated` (valid id promoted; `a/b` and `../evil` rejected to `None`).
- `cargo fmt`, `cargo clippy -D warnings`, and the full `dora-message` test suite pass.

---

---


phil-opp:
## 🤖 Automated codebase review — run summary

This PR is part of a batch of machine-generated findings (Claude, automated code review). Each finding is a separate branch/PR off a freshly-fetched `main`. Every PR was verified individually (fmt + clippy `-D warnings` + affected-crate tests); the combined diff additionally passed the full workspace test suite and an adversarial correctness re-review (no issues found).

### PRs opened, by category

**Correctness**
- **#2500** `fix(message)` — validate `node_id` recovered from log `fields` (was bypassing `NodeId` path-traversal validation via the raw tuple constructor).
- **#2501** `fix(cli)` — guard `Instant - Duration` overflow panic in `dora topic hz` (large `--window` aborts the command / freezes the TUI sampler thread).
- **#2502** `fix(mavlink2-bridge)` — increment the MAVLink sequence number per frame (every frame was sent with `sequence: 0`, breaking receiver drop-rate detection).

**Optimization**
- **#2503** `perf(memory-pool)` — release the pool-table lock before the `/dev/shm` unlink syscall (unrelated pool ops were serialized behind filesystem work on the daemon event loop).

**Simplification**
- **#2504** `refactor(arrow-convert)` — `String`↔`ArrowData` conversion delegates to the `&str` impl (removes duplicated validation).

**Documentation**
- **#2505** `docs(node)` — fix two comments / a test name that describe code which doesn't exist (a nonexistent `is_empty()` early-return whose misleading name invites regressing #2083, and a scheduler comment claiming "BTreeMap Ordering" for LRU-over-`VecDeque`).

### Findings considered and rejected (with reasons)

- **Signal names 22/23 mislabeled** (`common.rs`) — duplicate of open **#2453**.
- **Recording footer validation / writer-version trust / doctest** (`recording/src/lib.rs`) — docs already covered by open **#2487**; the version-trust fix is low value and would contend on the same file.
- **Telemetry `init_meter_provider` panics on misconfig; `OtelGuard` dropped in `build`** — the panic site is edited by open **#2492** (would conflict); the guard case is latent (in-tree caller `take()`s it first).
- **`topics.rs` overwrites `connect/endpoints` when `coordinator_addr` is `Some`** — genuine logic bug but latent/unreachable: all in-tree callers pass `coordinator_addr = None`, so fixing it is speculative.
- **`dora list -f json` emits NDJSON, not a JSON array** — ambiguous; changing a public CLI output contract could break streaming consumers and may be intentional.
- **Per-message `DataId` clone in `send_output`** — cannot be removed behavior-preservingly (would drop the output id from the error context); borderline, not worth the diagnostic regression.
- **`to_record_batch` rebuilds the Arrow schema per message** (mavlink) — real per-message allocation, but eliminating it needs a broad refactor across ~20 `MavlinkArrow` impls / the trait signature; deferred as too large for this batch.
- **Temporal-extraction boilerplate macro** (`arrow-convert`) — cosmetic; a macro would obscure the per-type guards.

### Note on CI

The **Audit (cargo-audit + cargo-deny)** check is failing on this PR — and on `main` and every branch — due to newly-published RUSTSEC advisories against transitive dependencies (`cxx` RUSTSEC-2026-0202, `proc-macro-error2` RUSTSEC-2026-0173, `serial` RUSTSEC-2017-0008). None of these PRs modify `Cargo.toml`/`Cargo.lock` or add dependencies, so this failure is pre-existing and unrelated to the changes; it needs a repo-level `deny.toml`/dependency triage rather than a change here.

---

---

phil-opp:

No issues found.

What I checked:
- Confirmed `FromStr for NodeId` (libraries/message/src/id.rs:97-100) runs `validate_node_id`, which rejects empty ids, leading `.`, and characters outside `[a-zA-Z0-9_.-]`. The prior `.map(NodeId)` used the crate-visible tuple constructor and bypassed this, so the fix genuinely closes the path-traversal / never-matching-`NodeId` gap.
- The `helper.node_id.or(fields...)` precedence is preserved: the typed field still wins, and only the untrusted `fields` fallback is now validated.
- The added test exercises both the valid-promotion and the `a/b` / `../evil` rejection cases correctly.

---

PR #2501, opened 2026-07-07
fix(cli): guard against Instant subtraction overflow in `dora topic hz`

phil-opp:
## Issue

`HzStats::record` (`binaries/cli/src/command/topic/hz.rs`) computed the sliding-window cutoff with `now - self.window_duration`, which panics with *"overflow when subtracting duration from instant"* whenever the window exceeds the current monotonic-clock value.

`--window` accepts any `usize` (only `0` is rejected by `parse_window`), so:

```
dora topic hz --duration 5 --window 100000000000 <topic>
```

aborts on the first sampled message. In one-shot (`--duration`) mode `record` runs on the main thread, so the whole command crashes with a backtrace; in the interactive TUI it panics the spawned sampler thread silently, leaving the stats table frozen — a confusing partial failure.

The sibling `topic/info.rs` already guards this exact class with `Instant::checked_sub`; `hz.rs` was not given the same treatment.

## Fix

- `record`: use `now.checked_sub(self.window_duration)` and treat a cutoff that predates the monotonic epoch as "keep everything" (nothing to evict), mirroring `info.rs`.
- The TUI frame loop's `now - sub_window` is guarded the same way and hoisted out of the per-topic loop (it is loop-invariant). `sub_window` is a fixed 1 s so it is only reachable within the first second of uptime, but it is the same class of bug.

## Validation

- New regression test `record_does_not_overflow_on_huge_window` (records with a `usize::MAX` window without panic).
- `cargo fmt`, `cargo clippy -D warnings`, and the full `dora-cli` test suite (241 tests) pass.

Note: PR #2469 also touches `hz.rs` but only hardens lock-poisoning in `run_hz_oneshot`/`intervals_ms`; it does not touch this overflow. The changes are in disjoint functions.

---

---


phil-opp:

No issues found.

- ✅ `HzStats::record` uses `checked_sub` and returns early when the cutoff predates the monotonic epoch, so no timestamps are evicted — consistent with keeping everything in the window.
- ✅ The TUI loop in `run_hz` hoists `now.checked_sub(sub_window)` out of the per-topic loop; `cutoff.is_some_and(|c| t < c)` correctly never breaks when the cutoff is `None`, counting all timestamps.
- ✅ Matches the existing guard in `topic/info.rs`; regression test exercises `usize::MAX` window without panic.

---

phil-opp:
Duplicate of #2613

PR #2502, opened 2026-07-07
fix(mavlink2-bridge): increment the MAVLink sequence number per frame

phil-opp:
## Issue

Every outbound frame from the MAVLink bridge node (`binaries/mavlink2-bridge-node/src/main.rs`) was transmitted with `sequence: 0`. The `MavHeader` was built once with `sequence: 0` and reused unchanged for both the `REQUEST_DATA_STREAM` setup message and every command sent by `handle_input`. `MavConnection::send` serializes the header verbatim, so the sequence field never advanced.

The MAVLink sequence field is meant to increment per transmitted frame (wrapping at 255) so receivers can compute link drop-rate / detect lost packets (e.g. `SYS_STATUS.drop_rate_comm`, radio-link statistics). A constant `0` makes loss detection impossible and is non-conformant with the protocol.

## Fix

- Add a shared `AtomicU8` sequence counter and a `next_header` helper that stamps each frame with the next value (`fetch_add`, wraps 255→0) before it is sent, preserving the configured `system_id`/`component_id`.
- Thread the counter through `handle_input` and `attempt_data_stream_request`.

`Ordering::Relaxed` is sufficient: the counter is only touched from the single-threaded event loop (the reader thread never sends).

## Validation

- New regression test `next_header_increments_and_wraps_sequence` (increment `0,1,2`; id fields preserved; wraps `255→0`).
- `cargo fmt`, `cargo clippy -D warnings`, and the full `dora-mavlink2-bridge-node` test suite pass.

---

---

### DIFF
--- a/binaries/mavlink2-bridge-node/src/main.rs
+++ b/binaries/mavlink2-bridge-node/src/main.rs
@@ -73,7 +73,7 @@ use serde::Deserialize;
 use std::{
     sync::{
         Arc,
-        atomic::{AtomicBool, Ordering},
+        atomic::{AtomicBool, AtomicU8, Ordering},
     },
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

- ✅ `next_header` uses `fetch_add(1, Relaxed)`, which returns the prior value, so sequences run 0, 1, 2, … as the test asserts.
- ✅ `AtomicU8::fetch_add` wraps 255 → 0 without panicking, matching the wraparound test.
- ✅ Both MAVLink `conn.send` call sites (`handle_input` and `attempt_data_stream_request`) are switched to `next_header`; no other send path keeps the constant-0 header. `Relaxed` ordering is adequate for a monotonic frame counter.

---

PR #2503, opened 2026-07-07
perf(memory-pool): release table lock before unlinking shared memory

phil-opp:
## Issue

`free_memory_pool` (`libraries/extensions/memory-pool/src/lib.rs`) held the pool-table mutex for the whole function, including the `free_shared_memory` call that performs a blocking `std::fs::remove_file` on the `/dev/shm` segment. Every other pool operation (`register`, `read`, `table_size`, and other frees) across all pools and nodes therefore blocked behind that unlink syscall.

Because the daemon frees pools on its main event loop after each read (`binaries/daemon/src/lib.rs`), this needlessly serialized unrelated pool traffic behind filesystem work.

## Fix

Scope the lock to just the `HashMap::remove` and drop the guard before the unlink. Semantics are unchanged — the entry is still removed first, then the segment freed, and a free error still propagates — only the critical section shrinks.

## Validation

- New `concurrent_frees_across_pools_all_succeed` regression test (16 pools freed concurrently, table left empty).
- Existing `cross_owner_free_succeeds` / `double_free_second_fails` cover the unchanged error/removal semantics.
- `cargo fmt`, `cargo clippy -D warnings`, and the full `dora-memory-pool` test suite pass.

The lock only ever guarded the in-memory table, never the filesystem, so moving the unlink out introduces no new free-vs-register race.

---

---


phil-opp:

No issues found.

- After the guard is dropped, the code only touches `entry` (now owned) and calls `free_shared_memory`, which accesses no shared `self` state beyond the filesystem — so moving the unlink out of the critical section introduces no data race.
- The removal is still atomic under the lock, so two concurrent frees of the same id still resolve to exactly one success and one "memory pool not found" (same as before); no double-unlink.
- The `registered_by != requested_by` debug log and the `free_shared_memory` error propagation are unchanged; only lock scope shrinks. The rationale (keeping `remove_file` off the shared critical section that the daemon hits per-read) is sound.

---

PR #2504, opened 2026-07-07
refactor(arrow-convert): delegate String conversion to the &str impl

phil-opp:
## Issue

`TryFrom<&ArrowData>` for `String` (`libraries/arrow-convert/src/from_impls.rs`) duplicated the `&str` impl byte for byte, differing only in a trailing `.to_string()`. The two single-element validation blocks (empty / length-1 / null checks, including the error strings) had to be kept in sync by hand.

## Fix

Have the `String` impl obtain a `&str` via the existing conversion and own only the `.to_string()`, so the validation lives in exactly one place and the two conversions can never drift apart. Removes ~13 duplicated lines.

## Validation

- Behavior is unchanged; `test_string_round_trip` and the rest of the `dora-arrow-convert` suite still pass.
- `cargo fmt` and `cargo clippy -D warnings` clean.

---

---


phil-opp:

No issues found.

- The `&str` impl (from_impls.rs:99-113) performs the exact same validation that was removed from the `String` impl (empty / `len != 1` / `null_count != 0`) and returns `array.value(0)`. Delegating and appending `.to_string()` is behavior-preserving, including identical error messages.
- `value.try_into()?` with the `let s: &str` annotation resolves unambiguously to the `&str` `TryFrom` impl; the borrow is immediately materialized into an owned `String`, so no lifetime problem.

---

PR #2505, opened 2026-07-07
docs(node): fix comments/test name that contradict the implementation

phil-opp:
## Issue

Two comments in `dora-node-api` described code that does not exist, both of which could mislead a future maintainer:

1. **`apis/rust/node/src/node/arrow_utils.rs`** — the test `raw_roundtrip_empty_primitive_uses_empty_buffer_path` and its comment claimed to exercise a `raw_buffer.is_empty()` early-return in `buffer_into_arrow_array`. No such early-return exists: the `NOTE` in that function explicitly documents that the empty-buffer case is deliberately *not* special-cased, because doing so discarded the array's declared length and reintroduced dora-rs/dora#2083. The misleading name actively invites "restoring" exactly the bug the `NOTE` warns against.

2. **`apis/rust/node/src/event_stream/scheduler.rs`** — the comment *"Process the ID with the oldest timestamp using BTreeMap Ordering"* describes an algorithm that isn't there. `next()` scans `last_used` (a `VecDeque` in least-recently-used order); there is no timestamp comparison and no `BTreeMap`. The struct field is already documented correctly (*"Tracks the last-used event ID"*).

## Fix

- Rename the test to `raw_roundtrip_empty_primitive_via_general_path` and correct its comment to state that the empty case goes through the general `ArrayData::try_new` path (no early-return), per the `NOTE`.
- Rewrite the scheduler comment to describe the actual LRU-over-`VecDeque` behavior.

Comments / test name only — no behavior change.

## Validation

`cargo fmt`, `cargo clippy -D warnings`, and the full `dora-node-api` test suite pass.

---

---

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -311,7 +311,10 @@ impl Scheduler {
             return Some(event);
         }
 
-        // Process the ID with the oldest timestamp using BTreeMap Ordering
+        // Return the next event from the least-recently-used input queue:
+        // scan `last_used` (a `VecDeque` in LRU order), pop from the first
+        // non-empty queue, and move that id to the back so other inputs get a
+        // turn. There is no timestamp comparison or `BTreeMap` involved.
         for index in 0..self.last_used.len() {
             let id = &self.last_used[index];
             if let Some((_size, queue)) = self.event_queues.get_mut(id)
--- a/apis/rust/node/src/node/arrow_utils.rs
+++ b/apis/rust/node/src/node/arrow_utils.rs
@@ -529,12 +529,15 @@ mod tests {
     }
 
     #[test]
-    fn raw_roundtrip_empty_primitive_uses_empty_buffer_path() {
+    fn raw_roundtrip_empty_primitive_via_general_path() {
         let data = UInt64Array::from(Vec::<u64>::new()).into_data();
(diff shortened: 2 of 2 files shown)


phil-opp:

No issues found. This is a comments/test-name-only change with no behavior impact, and both corrections match the code:

- `scheduler.rs::next()` (lines 306-329) scans `last_used` (a `VecDeque` in LRU order) and rotates the popped id to the back — there is no timestamp comparison and no `BTreeMap`, so the old comment was inaccurate and the new one is correct.
- `arrow_utils.rs`: the `NOTE` in `buffer_into_arrow_array_inner` (lines 173-180) explicitly documents that the empty-`raw_buffer` case is deliberately *not* special-cased (to avoid regressing #2083), so the old test name/comment referencing a `raw_buffer.is_empty()` early-return described nonexistent code. The rename and corrected comment are accurate; the assertions are unchanged.

---

phil-opp:
Obsoleted by #2366 (Arrow IPC data plane). This corrects a comment and test name describing the `raw_buffer.is_empty()` early-return in `buffer_into_arrow_array` — but #2366 removed that function entirely (`grep -rn 'fn buffer_into_arrow_array' apis/rust/node/src/` returns no hits), along with the `raw_roundtrip_empty_primitive_uses_empty_buffer_path` test. Nothing left to fix; closing.

PR #2509, opened 2026-07-08
Found during testing-fix(cxx): keep merged external event ids stable

DGHX12345:
# Issue

Generated C++ ROS2 action clients attach their result/status/feedback streams to Dora's event stream through `MergedEvents`. The old C++ node API had no stable per-external-stream identifier: external events were passed across the FFI boundary as erased `Any` payloads without an ID assigned by the combined stream.

That makes generated action-client matching fragile. A client can only distinguish “this event belongs to my result/status/feedback stream” if the external stream identity survives merging and crossing the C++/Rust boundary. Without a stable ID, generated `matches_*` helpers can fail to associate incoming external events with the stream that created them, which makes action examples miss result/status/feedback events or hang while waiting for events that were actually delivered.

There was also a small generated-header reliability issue: wrapper headers were written with `Write::write`, which is allowed to perform a partial write. For generated C++ headers this should use `write_all` so the wrapper include is either fully written or errors.

## Fix

- Add a `next_id` counter to `MergedEvents` when the ROS2 bridge feature is enabled.
- Add `MergedEvents::merge`, which assigns each external stream a monotonically increasing ID before merging it into the Dora event stream.
- Store that ID on `ExternalEvent` so generated C++ ROS2 clients can match result/status/feedback events against the stream they registered.
- Use `write_all` when generating C++ wrapper headers.

## Validation

- Added regression coverage: `merged_events_assigns_ids_to_external_streams`.
  - Verifies the first merged stream receives ID `1`.
  - Verifies the second merged stream receives ID `2`.
  - Verifies delivered external events preserve those IDs after merging.
- Ran:
  - `cargo clippy -p dora-node-api-cxx --features ros2-bridge -- -D warnings`
  - `cargo test -p dora-node-api-cxx --features ros2-bridge`


phil-opp:

No issues found.

This restores `MergedEvents::merge` / the `next_id` field that #2211 removed as "unreachable". I verified the removal rationale was mistaken: `libraries/extensions/ros2-bridge/msg-gen` generates `events.events.merge(Box::pin(stream))` against `crate::ffi::CombinedEvents` (e.g. `message.rs:388`, `service.rs:151/373`, `action.rs:298/393/790`), and `ffi::CombinedEvents` is the cxx-only struct defined in this C++ node crate. That generated code is `include!`d into `pub mod ros2` (lib.rs:255), so `merge` is genuinely reachable under the `ros2-bridge` feature — a plain workspace `.merge(` search misses it because it lives inside `quote!` templates.

Things I checked and cleared:
- The `#[cfg(any(feature = "ros2-bridge", test))]` gating is applied consistently to the field, both constructors, the method, and the `MergeExternal` import, so default builds carry no dead field/import and the new unit test still exercises `merge` under plain `cargo test`.
- Making `merge` private (`fn` vs the old `pub fn`) is fine: the generated caller lives in `mod ros2`, a descendant of the crate-root module where the `impl` is defined, so it can still reach the private method and the `MergedEvents.events` field.
- The `merge` body is a verbatim restore of previously-compiling code; the `Either<ExternalEvent, ExternalEvent>::flatten()` and `merge_external` types line up with `MergedEvent<ExternalEvent>` (merged.rs:104-123).
- The `build.rs` `write` → `write_all` change is a correct fix (guards against short writes).

---

phil-opp:
This is a corrected revert of #2211. That PR removed MergedEvents::merge as "unreachable" because a whole-workspace search found no caller — but the callers are generated: the ROS2 msg-gen templates emit events.events.merge(Box::pin(stream)) at 8+ sites (message.rs, service.rs, action.rs), and that code is include!'d into this crate only under --features ros2-bridge. A text search can't see it. The practical consequence is that main currently fails to compile dora-node-api-cxx --features ros2-bridge, and this PR is the fix. Worth stating that explicitly in the description — it reframes this from "new feature" to "regression fix," and flags for the author of #2211 that this re-adds code they deliberately removed, with the reason the removal reasoning was wrong.

Thanks for the PR!

PR #2511, opened 2026-07-08
fix(ros2-bridge): encode Binary-backed byte sequence as bytes, not rows

### LINKED ISSUE #2507: ros2-bridge: `sequence<u8>`/`char`/`byte` field given as Arrow `Binary` is double-nested on the CDR wire (extra length prefix corrupts the message)
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. It was found by randomly selecting a component (the `dora-ros2-bridge-arrow` serialize path) and reviewing it for correctness issues. The core finding was then **empirically confirmed** with a temporary unit test (byte output quoted below). Please still treat the severity assessment as a starting point for human triage.

## Summary

On the **dora → ROS2** serialize path, a variable-length byte sequence field (`sequence<u8>`, `sequence<char>`, `sequence<byte>`, i.e. ROS2 `uint8[]` / `char[]` / `byte[]`) is CDR-encoded **incorrectly when the Arrow column is a `Binary`/`LargeBinary` array**. The encoder emits an extra outer `u32` length prefix and nests the bytes one level too deep, so a real ROS2 peer decodes the sequence length as `1`, consumes a single byte, and every subsequent field in the message is misaligned → the whole message is corrupt.

The same field supplied as an Arrow `List<UInt8>` encodes correctly. So the two accepted Arrow representations of a byte vector diverge, and only the `List` one is right.

## Affected code

`libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs`

The top-level dispatch explicitly accepts a `Binary`/`LargeBinary` column and slices it to a 1-row array (`entry`):

```rust
// sequence.rs:44-59
} else if let Some(list) = self.column.as_binary_opt::<i32>() {
    Arc::new(list.slice(0, 1)) as ArrayRef        // entry = 1-ROW binary array
} else if let Some(list) = self.column.as_binary_opt::<i64>() {
    Arc::new(list.slice(0, 1)) as ArrayRef
}
```

For `BasicType::U8 | Char | Byte` that `entry` is handed to `ByteSequence`:

```rust
// sequence.rs:98-100
BasicType::U8 | BasicType::Char | BasicType::Byte => {
    ByteSequence { value: &entry }.serialize(serializer)
}
```

```rust
// sequence.rs:262-296
impl serde::Serialize for ByteSequence<'_> {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        if let Some(binary) = self.value.as_binary_opt::<i32>() {
            serialize_binary(serializer, binary)      // <-- 1-row binary array
        } else if let Some(binary) = self.value.as_binary_opt::<i64>() {
            serialize_binary(serializer, binary)
        } else {
            BasicSequence { value: self.value, ty: PhantomData::<UInt8Type> }
                .serialize(serializer)                // <-- correct List<UInt8> path
        }
    }
}

fn serialize_binary<S, O>(serializer: S, binary: &GenericByteArray<GenericBinaryType<O>>) -> ... {
    let mut seq = serializer.serialize_seq(Some(binary.len()))?;   // binary.len() == 1 (ROWS, not bytes)
    for value in binary.iter() {                                   // 1 iteration
        seq.serialize_element(value.unwrap_or_default())?;         // value: &[u8] -> serialized as its OWN seq
    }
    seq.end()
}
```

`serialize_binary` iterates **rows** (`binary.len()` == 1) rather than the bytes of the single row, and `serialize_element(&[u8])` routes through serde's `impl Serialize for [u8]`, which emits another `serialize_seq`. Net result for a byte field of length N:

- **Correct** CDR: `[u32 N][N bytes]`
- **Actual** CDR: `[u32 1][u32 N][N bytes]`  ← extra prefix + one nesting level

The dispatch that feeds this is `serialize/mod.rs:201-216`, which passes the field column verbatim into `SequenceSerializeWrapper`, so nothing upstream normalizes `Binary` → `List<UInt8>`.

## Empirical confirmation

Temporary unit test (built a `test_msgs/ByteSeqMsg { uint8[] data; int32 tail; }`, serialized the same value both as `List<UInt8>` and as `Binary`, via the real `cdr_encoding` codec used elsewhere in these tests). Value: `data = [10, 20, 30]`, `tail = 0x12345678`.

```
LIST   cdr bytes = [03, 00, 00, 00, 0a, 14, 1e, 00, 78, 56, 34, 12]
BINARY cdr bytes = [01, 00, 00, 00, 03, 00, 00, 00, 0a, 14, 1e, 00, 78, 56, 34, 12]
```

- `LIST` (correct): length `3`, bytes `0a 14 1e`, 1 byte CDR padding, then `int32 = 0x12345678`.
- `BINARY` (buggy): length `1`, then a nested length `3`, then the bytes — a ROS2 subscriber reads `len = 1`, consumes `0x03` as the single element, and misparses `tail` and everything after.

## Why this is reachable / a real risk

- The `Binary` branches are **explicitly implemented** (lines 44-59, 267-270), not a fallback error, so `Binary`/`LargeBinary` is a supported input representation — and it is the idiomatic Arrow encoding for a bytes field (what a `pyarrow` producer, or a `Vec<u8>` → Arrow conversion, naturally yields for e.g. `sensor_msgs/Image.data`, `CompressedImage.data`).
- The deserializer only ever emits `List<UInt8>` for byte sequences, so no existing round-trip test exercises the `Binary` input path — the bug is untested and silent.
- It only affects the **outbound (dora → ROS2)** path; inbound telemetry is unaffected.

## Suggested fix

`serialize_binary` should serialize the bytes of the **single row** as one CDR sequence, not iterate rows. Roughly:

```rust
fn serialize_binary<S, O>(serializer: S, binary: &GenericByteArray<GenericBinaryType<O>>) -> ... {
    // entry is always a 1-row binary array (sliced in SequenceSerializeWrapper);
    // the sequence is the bytes of that row.
    let bytes: &[u8] = if binary.is_empty() { &[] } else { binary.value(0) };
    let mut seq = serializer.serialize_seq(Some(bytes.len()))?;
    for b in bytes {
        seq.serialize_element(b)?;
    }
    seq.end()
}
```

and add a round-trip test that encodes a `sequence<u8>` field both as `Binary` and as `List<UInt8>` and asserts the two produce **identical** CDR bytes (plus a real-CDR decode of `(Vec<u8>, <trailing field>)` to prove alignment), mirroring the existing `bool_sequence_round_trips_through_real_cdr` test.

## Affected files
- `libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs` (`ByteSequence` / `serialize_binary`, lines ~44-59, ~98-100, ~262-296)
- `libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs` (dispatch, lines ~201-216)

phil-opp:
Fixes #2507

## Problem

On the **dora → ROS2** serialize path, a variable-length byte field (`uint8[]` / `char[]` / `byte[]`) supplied as an Arrow **`Binary`/`LargeBinary`** column was CDR-encoded incorrectly.

`serialize_binary` iterated the array's **rows** (always 1, since `SequenceSerializeWrapper` slices the column to a single row) and handed each row `&[u8]` to `serialize_element`, which routes through serde's `impl Serialize for [u8]` and emits *another* `serialize_seq`. The wire output for a byte field of length N was:

- **Correct** CDR: `[u32 N][N bytes]`
- **Actual** CDR: `[u32 1][u32 N][N bytes]` ← extra outer length prefix + one nesting level too deep

A real ROS2 peer reads the sequence length as `1`, consumes a single byte, and misaligns every subsequent field — corrupting the whole message.

The same field supplied as an Arrow `List` encoded correctly, so the two accepted Arrow representations of a byte vector diverged and only `List` was right. Because the deserializer only ever emits `List` for byte sequences, no existing round-trip test exercised the `Binary` input path, so the bug was silent. `Binary` is the idiomatic Arrow encoding for a bytes field (what a `pyarrow` producer or a `Vec<u8>` → Arrow conversion naturally yields, e.g. `sensor_msgs/Image.data`). Only the outbound path is affected; inbound telemetry is fine.

## Fix

Serialize the bytes of the single row directly instead of iterating rows:

```rust
let bytes: &[u8] = if binary.is_empty() { &[] } else { binary.value(0) };
let mut seq = serializer.serialize_seq(Some(bytes.len()))?;
for b in bytes {
    seq.serialize_element(b)?;
}
seq.end()
```

## Test

Added `byte_sequence_binary_matches_list_through_real_cdr` (mirroring the existing `bool_sequence_round_trips_through_real_cdr`): it builds `test_msgs/ByteSeqMsg { uint8[] data; int32 tail; }`, serializes `data = [10, 20, 30]`, `tail = 0x12345678` both as `Binary` and as `List<UInt8>` through the real `cdr_encoding` codec, and asserts:

- the two encodings produce **identical** CDR bytes,
- the length prefix is the byte count (`3`), not the row count (`1`),
- a real CDR decode of `(Vec<u8>, i32)` recovers both the bytes and the aligned trailing field.

The test fails on the old encoder and passes with the fix.

## Verification

- `cargo test -p dora-ros2-bridge-arrow` ✓ (11 passed)
- `cargo clippy -p dora-ros2-bridge-arrow` ✓ (no warnings)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TSEjAKQG3FCjZibXJDAkJW)_


phil-opp:

No issues found.

I verified the fix against the actual code:
- `ByteSequence`/`serialize_binary` is only reached from `SequenceSerializeWrapper` (sequence.rs:99), which slices the `Binary`/`LargeBinary` column to a single row (`list.slice(0, 1)`, sequence.rs:51/59). So `binary` is always a 1-row array and `binary.value(0)` correctly returns that row's bytes — the diff's premise holds.
- The old `serialize_seq(Some(binary.len()))` + `binary.iter()` path did produce `[u32 1][u32 N][N bytes]` because `serialize_element(&[u8])` routes through serde's `impl Serialize for [u8]` and emits a nested seq. The new code emits `[u32 N][N bytes]`, matching the `List<UInt8>` path.
- The `is_empty()` guard is harmless/defensive (after the slice the array has len 1). The generic `O: OffsetSizeTrait` correctly covers both `Binary` and `LargeBinary`.
- The regression test asserts `Binary` and `List<UInt8>` produce identical CDR bytes and that a real CDR decode of `(Vec<u8>, i32)` stays aligned, which is exactly the invariant at stake.

---

PR #2512, opened 2026-07-08
fix: full-repo audit fixes — install scripts, managed Python env, CLI correctness (F1–F10)

heyong4725:
## Summary

This branch lands the 10 defect fixes from a full-repository feature audit (340 user stories cataloged from the code and individually tested on macOS: unit suites, all e2e suites, live daemon sessions, example dataflows, and network installs). Final tracker: 281 Pass / 19 Retested-Pass / 37 N-A-Env / 3 Blocked / **0 Fail**.

**Install & release pipeline**
- `install.sh` / `install.ps1` requested `dora-<target>` assets while releases ship `dora-cli-<target>`, so every scripted install 404'd. Scripts now match the published assets (unix nested tar, windows flat zip), `release.yml` is aligned to the same convention, and Intel Macs get a clear unsupported-platform error instead of a 404. Verified end-to-end against v0.5.0. (b070b984)

**Python operators on pre-release builds**
- The managed Python env pins `dora-rs==<workspace version>`, which is unpublished on PyPI for pre-release builds — all Python operators failed to build, and no CI caught it (the only operator example needs a webcam). Added a `DORA_PYTHON_RUNTIME_SOURCE` override plus an actionable error naming the cause and fix. Operator dataflows verified green outside the checkout. (b0dfbda9)

**CLI correctness**
- `dora logs --all-nodes --tail N` hung forever without `--follow`; it now fetches per-node history, merges by timestamp, applies filters/tail, and exits — and it survives per-node fetch failures and falls back to local logs (with a hint) for finished dataflows. (1ac3959d)
- `dora validate` missed whole-descriptor checks (negative timing fields, bad log sizes, ros2 conflicts) that only fired at `dora run`; `check_dataflow_static` is now wired into validate, and descriptor schema errors now point at the detailed cause instead of "check the file exists and is valid YAML". (04dddbd6)
- `node remove --grace` took bare floats while `node restart/stop --grace` took duration strings; all three now share a lenient parser accepting both, and a `Duration::from_secs_f64` overflow panic on huge input is now a clean clap error. (53c4f5a7)
- `self update --check-only` compared versions with string `!=` and offered a downgrade (1.0.0-rc1 → 0.5.0); it now uses semver comparison. (0af39d80)

**Housekeeping**
- Resolved the red `cargo audit` gate (7 advisories incl. quick-xml DoS and a fresh cxx advisory): crossbeam-epoch and cxx updated, quick-xml 0.41 in the msg-gen crate (70/70 tests green), documented time-boxed ignores for two transitive copies. (942a9342)
- Widened the `input_recovered` fault-tolerance e2e window 5s → 15s; the macOS failure root-caused to test timing (startup + zenoh latency), not a daemon bug — recovery logic verified correct. (853188fd)
- Synced the drifted repo-root `dora-schema.json` with the generated schema. (ac9640fb)

## Review

The diff went through a 26-agent adversarial code review (29 verified findings); the material ones are fixed in-place on this branch — including `release.yml` asset alignment, install.ps1 flat-zip handling, the Intel-Mac error, the `--grace` overflow panic, and `logs --all-nodes` per-node failure tolerance with archived fallback.

## Test plan

- [x] `make qa-fast` green at HEAD (fmt, clippy `-D warnings`, cargo-audit + cargo-deny, unwrap budget 159/163, typos)
- [x] Full workspace test suite green at HEAD (`cargo test --all` excluding Python/examples packages)
- [x] `cargo check --examples` green at HEAD
- [x] fault-tolerance e2e 11/11, ws-cli e2e 25/25, example-smoke spot checks green (audit phase 4 retest)
- [x] Every fix re-verified individually against its original failing repro

## Follow-ups proposed (not in this PR)

- Release-pipeline conformance test: CI job that builds the release archives and runs the install scripts against them (this class of breakage was invisible to every existing gate)
- Headless Python-operator smoke in CI (the managed-env blind spot)
- Schema drift check in CI (`generate_schema && git diff --exit-code`)
- Archived-dataflow introspection (persist the descriptor so post-mortem `logs --all-nodes` can enumerate nodes)


phil-opp:

No issues found.

I checked the individual fixes bundled here:
- `parse_grace_period` (`common.rs`) uses `Duration::try_from_secs_f64`, so negative/non-finite bare-seconds values now error out instead of panicking — the three node subcommands are consistent.
- `self update --check-only`: `bump_is_greater(current_version, &release.version)` returns `release > current`, so a dev build ahead of the latest release is correctly reported as up to date; the `unwrap_or(!=)` fallback is safe.
- `check_dataflow_static` extraction: `check_dataflow` still runs it first, so no static check is dropped from the `dora run` path. `validate` ends up running `check_wiring` twice (once directly, once inside the static check) — harmless duplication, not a correctness problem.
- `logs --all-nodes`: the non-`--follow` path now fetches per-node history, merges by timestamp, applies tail after the merge, and returns instead of hanging; with `--follow` it fetches history then streams live, which shouldn't double-print.
- install scripts / `release.yml` asset names and layouts line up (`dora-cli-<target>` nested unix tarball via `--strip-components 1`, flat windows zip).

One thing worth confirming in CI: the `quick-xml` 0.38 → 0.41 bump in `dora-ros2-bridge-msg-gen` crosses a couple of minor versions with known API churn (reader/attribute APIs) — make sure the msg-gen build and ros2-bridge codegen still compile and produce identical output.

---

phil-opp:
🤖 _Automated review by Claude — this is a fully automated review with no human in the loop; treat it as advisory._

The bundled fixes look correct against this PR's base, and I agree with the earlier review's read of them. One new point that the prior review predates: this branch is now `mergeable_state: dirty` and, more importantly, `main` has diverged in exactly the file this PR refactors.

Current `main`'s `check_dataflow` (`libraries/core/src/descriptor/validate.rs`) validates more timing fields than this PR's `check_dataflow_static` split accounts for — specifically the per-input `input_timeout` check and the dataflow-level `health_check_interval` check (both feed `Duration::from_secs_f64` in the daemon). This PR's `check_dataflow_static` only lifts `check_timing_fields` + `check_wiring` + the log-config resolution.

When you resolve the conflict, please make sure those two newer checks land inside `check_dataflow_static` (not just in `check_dataflow`). Otherwise a naive "keep this PR's version of `check_dataflow`" resolution would silently drop `input_timeout`/`health_check_interval` validation from `dora run` altogether, and even a careful resolution that keeps them only in `check_dataflow` would leave `dora validate` unable to catch them — which is the opposite of this PR's stated goal of making `validate` match `run`.

Nothing else new to flag beyond what the earlier review already covered.

---

phil-opp:

The `origin/main` merge resolves the concern from the previous review: `check_dataflow_static` in `libraries/core/src/descriptor/validate.rs` now contains the `input_timeout` and `health_check_interval` second-field guards, and `check_dataflow` delegates to `check_dataflow_static` rather than keeping those checks only in itself — so `dora validate` and `dora run` both enforce them now. No new issues from my read of the merged diff.

---

PR #2517, opened 2026-07-08
ci: gate find_package smoke on the merge queue instead of every PR

phil-opp:
The C/C++ find_package smoke matrix (3 OSes x 2 release cargo builds x
4 cmake builds) ran on every PR touching apis/c{,++}, xtask, or any of
the transitive Rust deps it wraps -- 100 of the last 313 commits to main,
though only 13 touched apis/c{,++} and 2 touched xtask. It gated nothing:
it is absent from both .trunk/trunk.yaml and branch protection, so a red
run never blocked a merge.

Move it to the expensive tier alongside test/e2e/contract-tests/bench:
the matrix now runs on trunk-merge/* batches, pushes to main, and manual
dispatch, so one run covers a batch of up to 5 PRs. Add the aggregator
`find_package smoke all green` to the queue's required statuses so a
broken find_package contract now actually blocks the merge, rather than
surfacing at release time when publish-c-cpp-libraries.yml runs.

Making it a required status forces dropping the workflow-level `paths:`
filter: a workflow skipped by path filtering emits no check run at all,
so the required name would stay Pending and stall the batch forever. (A
job skipped by `if:` does emit a check run, so `if:` gating is safe.) The
filter therefore moves into a `changes` job that resolves the changed
files via `gh api` -- pulls/files for PRs, compare for pushes -- and the
aggregator runs unconditionally. Same shape as `pip-release all green`.

Matching is bash prefix-stripping, not globbing. Every entry is a
directory prefix or one exact path, so this is exact, and it sidesteps
the `+`-escaping trap that broke the `apis/c++/**` glob in #1918. Each
endpoint truncates its file list at a different size (3000 vs 300) and a
truncated list makes a "no match" answer unsound, so both caps -- plus
API errors and base-less pushes -- fail open toward running the matrix.

### DIFF
--- a/.github/workflows/test-c-cpp-libraries.yml
+++ b/.github/workflows/test-c-cpp-libraries.yml
@@ -6,49 +6,158 @@ name: Test C/C++ find_package
 # dora-operator-api-cxx. Each crate path: cargo build → xtask stage →
 # cmake configure → cmake build of a checked-in smoke fixture.
 # Guards #1756's find_package contract.
+#
+# Tiering, mirroring .github/workflows/ci.yml:
+#
+#   The smoke matrix is expensive (3 OSes × 2 release cargo builds ×
+#   4 cmake builds; macOS bills at 10x, Windows at 2x). It is skipped on
+#   regular PRs, and on Trunk merge-queue batches / pushes to main it runs
+#   only when the changed files touch the find_package surface — so one run
+#   covers a batch of up to 5 PRs, and quiet batches cost one short runner.
+#   Manual dispatch always runs it.
+#
+#   `find_package smoke all green` is a merge-queue required status
+#   (.trunk/trunk.yaml). That is why there is no workflow-level `paths:`
+#   filter: a workflow skipped by path filtering emits *no* check run at
+#   all, so the required name would stay Pending and stall the batch.
+#   (A job skipped by `if:` does emit a check run, which is why the `if:`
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found.

I checked the merge-queue gating logic for a false-green risk and it holds up:

- The `find-package-all-green` aggregator has `if: always()` and `needs: [changes, find-package-smoke]`, so the required status `find_package smoke all green` is always produced on every batch branch (no Pending stall), and it fails on `failure`/`cancelled` while treating a `skipped` matrix as green — which is the intended "batch touched nothing relevant" case.
- The `changes` job fails open everywhere (API error, truncation at the per-endpoint caps of 3000/300, base-less push, unknown event) toward `relevant=true`, so an undeterminable diff runs the matrix rather than silently skipping a gated check.
- Prefix matching uses quoted parameter expansion (`${file#"$prefix"}`), so entries like `apis/c++/` are exact string strips, not globs — correctly sidestepping the `+`-metacharacter trap that broke the old `paths:` filter.
- Gating is consistent across events: `push`/`workflow_dispatch` satisfy `github.event_name != 'pull_request'`; the merge-queue batch PR has `head_ref` starting with `trunk-merge/`. Regular PRs skip the matrix but still emit a green aggregator, which is fine since the status is only required by the Trunk queue.

---
🤖 *This is a fully automated review by Claude. It may contain mistakes — please verify before acting.*

---

PR #2576, opened 2026-07-09
fix(cli): gate local_working_dir on daemon pinning (machine or labels)

phil-opp:
## Issue

`local_working_dir` (in `binaries/cli/src/common.rs`) decides whether the CLI should send a base working directory to the coordinator. It does so by checking that **no node pins a target machine**:

```rust
.all(|n| n.deploy.as_ref().map(|d| d.machine.as_ref()).is_none())
```

`n.deploy.as_ref().map(|d| d.machine.as_ref())` has type `Option<Option<&String>>`, so `.is_none()` is `true` **only when the node has no `deploy` block at all**. It never inspects `d.machine`. Consequences:

- A node with a `deploy` block that only sets `working_dir`/`labels` (and leaves `machine: None`) is misclassified as pinning a machine → `local_working_dir` returns `None` → `dora start`/`dora build` send no base working dir → the daemon cannot resolve the node's relative path and spawn fails, even though every node runs locally.
- Symmetrically, a node with `deploy: { machine: "gpu" }` would be treated as *not* pinning a machine.

## Fix

Extract the intent into a small, testable helper using `and_then` so only `deploy.machine` disables the local fast path:

```rust
fn deploy_pins_machine(deploy: Option<&Deploy>) -> bool {
    deploy.and_then(|d| d.machine.as_ref()).is_some()
}
```

`local_working_dir` now returns the local dir iff **no node pins a machine** and the CLI and daemon are on the same machine.

## Validation

- New unit test `deploy_pins_machine_only_when_machine_set` covers the three deploy shapes (no deploy block, deploy with `machine: None`, deploy pinning a machine).
- `cargo test -p dora-cli --lib common::` → 7 passed; `cargo fmt`/`clippy` clean.

---

---

### DIFF
--- a/binaries/cli/src/common.rs
+++ b/binaries/cli/src/common.rs
@@ -7,6 +7,7 @@ use dora_download::download_file;
 use dora_message::{
     cli_to_coordinator::ControlRequest,
     coordinator_to_cli::{ControlRequestReply, DataflowList, DataflowResult},
+    descriptor::Deploy,
 };
 use eyre::{Context, ContextCompat, bail};
 use std::{
@@ -267,6 +268,25 @@ pub(crate) fn canonicalize_working_dir(
     }
 }
 
+/// Returns `true` if the node's deploy config routes it to a specific
+/// (possibly remote) daemon rather than the default/unnamed one.
+///
+/// This mirrors the coordinator's `resolve_daemon` priority
+/// (`machine` > `labels` > unnamed): a node is pinned when it either names a
+/// `machine` or carries a non-empty `labels` set, since label-based
+/// scheduling can match a remote daemon just as `machine` can. Only nodes
+/// that pin nothing land on the local/default daemon and may safely reuse the
+/// CLI-side [`local_working_dir`] fast path.
+///
+/// A `deploy` block that merely sets `working_dir` (no `machine`, no `labels`)
+/// does *not* pin a daemon. The previous `map(|d| d.machine.as_ref())`
+/// produced an `Option<Option<_>>` whose `is_none()` was true only when the
+/// whole `deploy` block was absent, so it both disabled the fast path for
+/// harmless `working_dir`-only blocks and ignored `labels` entirely.
+fn deploy_pins_daemon(deploy: Option<&Deploy>) -> bool {
+    deploy.is_some_and(|d| d.machine.is_some() || !d.labels.is_empty())
+}
+
 pub(crate) fn local_working_dir(
     dataflow_path: &Path,
     dataflow_descriptor: &Descriptor,
@@ -276,7 +296,7 @@ pub(crate) fn local_working_dir(
         if dataflow_descriptor
             .nodes
             .iter()
-            .all(|n| n.deploy.as_ref().map(|d| d.machine.as_ref()).is_none())
+            .all(|n| !deploy_pins_daemon(n.deploy.as_ref()))
             && cli_and_daemon_on_same_machine(coordinator_session)?
         {
             Some(
@@ -313,6 +333,7 @@ pub(crate) fn write_events_to() -> Option<PathBuf> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::BTreeMap;
 
     #[test]
     fn working_dir_or_parent_prefers_override() {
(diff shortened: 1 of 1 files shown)


phil-opp:
## Codebase review — run summary

This PR is part of a batch from an automated codebase review (correctness, optimization, documentation, simplification). Each PR is small, independently reviewed against current `origin/main`, and machine-generated by Claude (an AI agent) — please review before merging.

### PRs opened this run

**Correctness / security**
- #2576 — `fix(cli)`: `local_working_dir` used `Option<Option<_>>::is_none()`, so a `deploy` block without a `machine` wrongly disabled the local-working-dir path (breaking local spawn). Switched to `and_then`.
- #2577 — `fix(message)`: `dora/timer/hz/<huge>` rounded `1/hz` below 1 ns → `Duration::ZERO` → busy-loop timer. Reject zero-length intervals.
- #2578 — `fix(core)`: git-source pin denylist missed `~` and `@{`, so `main~5` / `HEAD@{yesterday}` resolved to a different commit than the literal ref.
- #2579 — `fix(tracing)`: trace-context codec split on `;`/`:`, both legal in W3C `tracestate` values, corrupting propagated context. Switched to a newline separator + first-`:` split.
- #2581 — `fix(daemon)`: node stdout/stderr `read_until` was unbounded (OOM via a single multi-GB line, defeating `truncate_log_line`); stderr ring buffer kept the untruncated line; stdout busy-spun on a persistent read error. Added a bounded `read_capped_line` and stopped the reader on error.

**Simplification**
- #2580 — `refactor(core)`: removed the always-`None` `TypeSuggestion.line` field and its unreachable `Display` branch.

### Validation
- Per-crate `fmt` + `clippy -D warnings` + tests pass for every affected crate; each PR adds regression tests.
- Combined-diff gates on all six changes together: `/review` and `/simplify` produced **no findings**.
- Full workspace suite (`cargo test --all --lib --bins --tests`, Python/examples excluded): **91 test binaries pass, 1 fails** — `dora-node-api`'s `zenoh_schema_cache::advanced_pub_cache_serves_late_joining_subscriber`, which fails deterministically in this sandbox at a second `zenoh::open()` (networking) and is in a crate **none of these PRs touch**. It is a pre-existing environment limitation, not a regression.

### Findings considered but not opened as PRs (rejected)
- **ros2-bridge action feedback threads leak** — real, but a correct fix needs cancellation plumbing and can't be verified without a live ROS2 runtime; too risky to ship blind.
- **ros2-bridge dead `messages` param** — valid cleanup, but building `ros2-bridge-node` pulls heavy ROS deps; low ROI, deferred.
- **mavlink2-bridge rebuilds the Arrow schema per frame** — real hot-path allocation, but caching needs a trait-signature change across ~17 impls; broad and modest ROI, deferred.
- **ros2-bridge service-client blocks up to 30 s and ignores `Stop`** — real responsiveness issue, but needs an event-loop restructure; risk outweighs the benefit here.
- **operator FFI drops `Event::Error`** — real, but the fix adds a variant to the public `Event` enum (breaking API + design change), out of scope for a small PR.
- **node `zenoh_publish` over-alloc guard inconsistency** — latent only; masked by zenoh's equal-length guarantee, not an active bug.
- **node scheduler `flush` drops not counted in `drain_drop_counts`** — arguable-by-design observability nit.
- **coordinator `retrieve_logs` resolves daemon by `deploy.machine` instead of `node_to_daemon`** — real for label-deployed nodes, but the fix touches label-deploy resolution; medium risk, hard to verify in-sandbox, deferred.
- **coordinator `GetParams` silently drops undecodable params** (while `GetParam` errors) — low-severity consistency nit.
- **record-node `timestamp_offset` derived from `now()` rather than the message timestamp** — arguably by design.

*Machine-generated by Claude (an AI agent).*

---

phil-opp:

No issues found.

The fix is correct. The old `n.deploy.as_ref().map(|d| d.machine.as_ref()).is_none()` yields `Option<Option<&String>>`, whose `is_none()` is true only when the whole `deploy` block is absent — so a `deploy` block that set only `working_dir`/`labels` (with `machine: None`) made the `.all(...)` predicate false and wrongly disabled the local-working-dir fast path. `deploy.and_then(|d| d.machine.as_ref()).is_some()` flattens correctly: `true` only when a machine is actually pinned, so `!deploy_pins_machine(...)` is true for both "no deploy block" and "deploy block without a machine". The added test covers all three cases.

(The batch-summary comment above is a run summary, not a prior code review, so this is a first review of the diff.)

---

phil-opp:
🤖 **Automated review by Claude** — _fully automated (Claude Code); no human has vetted this. Treat it as a suggestion, not an approval._

The core fix is correct: the old `n.deploy.as_ref().map(|d| d.machine.as_ref()).is_none()` is an `Option<Option<_>>::is_none()` that was true only when the whole `deploy` block was absent, so a `working_dir`/`labels`-only block wrongly disabled the `local_working_dir` fast path. Switching to `deploy_pins_machine` via `and_then` fixes that specific case. One gap:

**`deploy_pins_machine` ignores label-based scheduling, which can place a node on a *remote* daemon.**

The coordinator resolves a node to a daemon in `resolve_daemon` (`binaries/coordinator/src/run/mod.rs:138-159`) with priority **`machine` > `labels` > `unnamed`**. A node with `machine: None` but a non-empty `deploy.labels` is scheduled onto a label-matched daemon (`get_matching_daemon_by_labels`, line 150-153), which may live on a different machine.

Because `deploy_pins_machine` (`binaries/cli/src/common.rs:273`) only inspects `deploy.machine`, a labels-only node is treated as *not* pinned and no longer disables the local fast path. `cli_and_daemon_on_same_machine` only compares the CLI against the default/unnamed daemon, so it can still return `true` in a cluster that also has remote labeled daemons. In that situation `local_working_dir` returns the CLI-local canonical dir, and that single path is broadcast to *every* daemon — including the remote label-matched one, where the directory generally won't exist (or, worse, resolves relative node paths against the wrong dir if it coincidentally does).

Under the previous code, any present `deploy` block (including labels-only) forced `local_working_dir = None`, so remote labeled daemons safely fell back to their own work dir. The fix over-corrects. Suggest treating a node as pinned when **either** `machine` is set **or** `labels` is non-empty, mirroring `resolve_daemon`:

```rust
fn deploy_pins_machine(deploy: Option<&dora_message::descriptor::Deploy>) -> bool {
    deploy.is_some_and(|d| d.machine.is_some() || !d.labels.is_empty())
}
```

and updating the helper's name/doc, since `labels` also pins a node to a specific (possibly remote) daemon. A regression test with a labels-only node would guard this.

---

phil-opp:

The current commit addresses the labels gap raised earlier: the helper is now `deploy_pins_daemon` and treats a node as pinned when it sets `machine` **or** a non-empty `labels` map (`deploy.is_some_and(|d| d.machine.is_some() || !d.labels.is_empty())`), mirroring the coordinator's `resolve_daemon` priority (machine > labels > unnamed), with a labels-only regression test added. Looks correct now — no new findings.

---

PR #2577, opened 2026-07-09
fix(message): reject zero-length hz timer intervals

phil-opp:
## Issue

The per-interval timer task builds `tokio::time::interval(interval)` (`binaries/daemon/src/running_dataflow.rs:315`), which **panics** on a zero-length interval (`` `period` must be non-zero ``). Several descriptor inputs parsed to a zero interval and reached that panic:

- `dora/timer/hz/<huge>` — for a very large `hz`, `1/hz` rounds below 1 ns and `Duration::try_from_secs_f64` returns `Ok(Duration::ZERO)`.
- `dora/timer/secs/0`, `millis/0`, `micros/0`, `nanos/0` — `Duration::from_*(0).is_zero()` is `true`.

Parsing accepted all of these, so a bad descriptor panicked a daemon task at runtime instead of failing at load.

## Fix

Hoist a single `interval.is_zero()` guard to just after the `match unit { … }` block, so **all five units** are covered in one place, and simplify the `hz` arm. A zero interval now fails at parse time with a clear message (surfaced at YAML load via `Deserialize for InputMapping`) instead of panicking the daemon later.

## Compatibility

Not a compat break for previously-valid input: `dora/timer/hz/0` was already rejected on `main` by the existing `hz <= 0.0` guard. The only newly-rejected inputs are `<unit>/0` and `hz` values large enough to round `1/hz` below 1 ns — all of which previously produced a zero interval that panicked the daemon. This trades a runtime panic for a parse-time error. No in-repo dataflow uses a zero-interval timer.

## Validation

- `parse_timer_rejects_zero_interval_for_every_unit` covers `secs/0`, `millis/0`, `micros/0`, `nanos/0`, and the huge-`hz` case.
- `cargo test -p dora-message --lib config::tests::parse_timer` → 8 passed; `cargo fmt`/`clippy` clean.

---
*This is a machine-generated pull request authored by Claude (an AI agent), updated to address review feedback. Please review carefully before merging.*

### DIFF
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -280,9 +280,22 @@ impl FromStr for InputMapping {
                                     "hz must be a positive finite number (got `{value}`)"
                                 ));
                             }
-                            Duration::try_from_secs_f64(1.0 / hz).map_err(|e| {
+                            let interval = Duration::try_from_secs_f64(1.0 / hz).map_err(|e| {
                                 format!("hz `{value}` produces an out-of-range interval: {e}")
-                            })?
+                            })?;
+                            // A very large hz makes `1/hz` round below 1ns, so
+                            // `try_from_secs_f64` returns `Ok(Duration::ZERO)`
+                            // rather than an error. A zero-length interval means
+                            // the timer fires continuously (busy loop), so reject
+                            // it explicitly -- mirroring the tiny-hz overflow guard
+                            // above.
+                            if interval.is_zero() {
+                                return Err(format!(
+                                    "hz `{value}` is too large: it produces a zero-length \
+                                     timer interval"
+                                ));
+                            }
+                            interval
                         }
                         other => {
                             return Err(format!(
@@ -746,6 +759,23 @@ mod tests {
             .parse::<InputMapping>()
             .unwrap_err();
         assert!(err.contains("hz"), "error should mention hz: {err}");
+    }
+
+    #[test]
+    fn parse_timer_hz_rejects_zero_interval() {
+        // A pathologically large hz makes 1/hz round below 1ns, so
+        // `try_from_secs_f64` returns `Ok(Duration::ZERO)` instead of an
+        // error. That would be a zero-length (busy-loop) timer, so parsing
+        // must reject it rather than silently accept it.
+        let err = "dora/timer/hz/1000000000000"
+            .parse::<InputMapping>()
+            .unwrap_err();
+        assert!(err.contains("hz"), "error should mention hz: {err}");
+        // Sanity check: the raw conversion really does round to zero here.
+        assert_eq!(
+            Duration::try_from_secs_f64(1.0 / 1_000_000_000_000.0),
+            Ok(Duration::ZERO)
+        );
     }
 
     /// Regression test for dora-rs#2031: the `Display` impl previously emitted
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -280,22 +280,13 @@ impl FromStr for InputMapping {
                                     "hz must be a positive finite number (got `{value}`)"
                                 ));
                             }
-                            let interval = Duration::try_from_secs_f64(1.0 / hz).map_err(|e| {
-                                format!("hz `{value}` produces an out-of-range interval: {e}")
-                            })?;
                             // A very large hz makes `1/hz` round below 1ns, so
-                            // `try_from_secs_f64` returns `Ok(Duration::ZERO)`
-                            // rather than an error. A zero-length interval means
-                            // the timer fires continuously (busy loop), so reject
-                            // it explicitly -- mirroring the tiny-hz overflow guard
-                            // above.
-                            if interval.is_zero() {
-                                return Err(format!(
-                                    "hz `{value}` is too large: it produces a zero-length \
-                                     timer interval"
-                                ));
-                            }
-                            interval
+                            // `try_from_secs_f64` returns `Ok(Duration::ZERO)`;
+                            // that zero interval is caught by the shared guard
+                            // after this `match`, together with `<unit>/0`.
+                            Duration::try_from_secs_f64(1.0 / hz).map_err(|e| {
+                                format!("hz `{value}` produces an out-of-range interval: {e}")
+                            })?
                         }
                         other => {
                             return Err(format!(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Verified: for a very large `hz`, `1.0 / hz` rounds below 1ns and `Duration::try_from_secs_f64` returns `Ok(Duration::ZERO)` (the test's sanity assertion confirms this for `hz = 1e12`), so the pre-existing `try_from_secs_f64` guard only caught the overflow (tiny-`hz`) direction. The new `interval.is_zero()` check closes the underflow direction and rejects the zero-length (busy-loop) timer, mirroring the existing overflow guard. The `> 0` / finiteness pre-check upstream already rejects `hz <= 0`/NaN/inf, so `interval` here is only ever zero via rounding, which the new branch handles. The regression test exercises the exact case.

---

phil-opp:
🤖 **Automated review by Claude** — this review was fully generated by an AI agent with no human in the loop. Advisory only; please verify before acting.

The `hz` fix is correct: `interval.is_zero()` closes the rounds-to-zero case that `Duration::try_from_secs_f64` returns as `Ok(Duration::ZERO)`, it sits on the only hz construction path (`FromStr`, which `Deserialize for InputMapping` also routes through), and no legitimate ≥1 ns interval is rejected.

One thing worth flagging beyond the hz path, though: **the same zero-length-interval hazard is still reachable via the sibling integer units.** In the same `match unit` block, `dora/timer/secs/0`, `dora/timer/millis/0`, `dora/timer/micros/0`, and `dora/timer/nanos/0` build `Duration::from_secs(0)` / `from_millis(0)` / … (`libraries/message/src/config.rs:250`–`272`) with no zero-check, and the new `is_zero()` guard is placed *inside the `hz` arm* rather than after the match — so it doesn't cover them. `check_input` in `libraries/core/src/descriptor/validate.rs` (`InputMapping::Timer { .. } => {}`) doesn't inspect the interval either.

Downstream, the daemon does:

```rust
// binaries/daemon/src/running_dataflow.rs:310
let mut interval_stream = tokio::time::interval(interval);
```

and `tokio::time::interval` **panics when the period is zero**. So a descriptor containing e.g. `dora/timer/millis/0` is accepted at parse/validate time but panics the daemon timer task at runtime — the same failure mode this PR fixes for `hz`, just through a different unit.

Consider centralizing the rejection so it covers every unit — e.g. hoist the `interval.is_zero()` check to just before `Self::Timer { interval }` (after the `match`), or add it in `check_input`. That closes the whole class rather than only the `hz` corner. Not a blocker for the `hz` fix as written.

---

phil-opp:
## Fix is correct but covers 1 of 5 timer units — and the premise is off

`tokio::time::interval(Duration::ZERO)` does not busy-loop, it **panics**: `` `period` must be non-zero. `` That call is `binaries/daemon/src/running_dataflow.rs:315`, inside the spawned per-interval timer task. So a zero interval panics a daemon task rather than spinning.

More importantly: `Duration::from_millis(0).is_zero() == true`, and the `secs`/`millis`/`micros`/`nanos` arms at `config.rs:250-269` have **no zero guard**. `dora/timer/millis/0` still parses fine on this branch and still reaches the panic. The guard is inside the `hz` arm (`config.rs:283-295`); hoisting a single `if interval.is_zero()` to after the `match unit { ... }` block covers all five units in fewer lines than this PR adds. Please extend the test to match.

**Backward-compat is a non-issue, contrary to the description:** `dora/timer/hz/0` was already rejected on main by the `hz <= 0.0` guard at `config.rs:277-281` (existing test `parse_timer_hz_rejects_zero:723`). The only newly-rejected inputs are `hz` > ~2e9, which previously parsed to a zero interval that panicked the daemon anyway — so this trades a runtime panic for a parse-time error. It does fail at YAML load with a good message via `Deserialize for InputMapping` (`config.rs:412-420`). No in-repo YAML uses a zero-interval timer.

One call worth making explicitly: rejecting `millis/0` is more plausible in the wild than `hz/3000000000`, and it currently panics the daemon. If that is considered too aggressive a compat break, say so deliberately rather than leaving it unguarded.

phil-opp:
Thanks — you're right on both counts. Pushed 90b6c8b:

- **Hoisted the guard to cover all five units.** A single `if interval.is_zero()` now sits right after the `match unit { … }` block, so `secs/0`, `millis/0`, `micros/0`, `nanos/0`, and the huge-`hz` case are all rejected in one place; the `hz` arm is simplified accordingly. Replaced the hz-only test with `parse_timer_rejects_zero_interval_for_every_unit` covering all five.
- **Corrected the premise.** It's a panic, not a busy-loop — `tokio::time::interval(Duration::ZERO)` panics with `` `period` must be non-zero `` in the timer task at `running_dataflow.rs:315`. Updated the code comment and the PR description.
- **Fixed the compat note** in the description: `hz/0` was already rejected by the `hz <= 0.0` guard; the only newly-rejected inputs are `<unit>/0` and `hz` > ~2e9, which previously panicked the daemon — so this trades a runtime panic for a parse-time error.

On your explicit question: I went with rejecting `millis/0` (and the other `<unit>/0`) deliberately, since it panics the daemon today and no in-repo dataflow uses a zero-interval timer. Happy to relax it to a warning-and-clamp instead if you'd prefer to keep those parsing.

---

phil-opp:

The new commit (90b6c8b) that hoists the `interval.is_zero()` guard to just after the `match unit` block correctly closes the whole class — `secs/0`, `millis/0`, `micros/0`, `nanos/0`, and the huge-`hz` rounds-to-zero case are all rejected in one place, and the test covers each. This addresses the earlier feedback and doesn't introduce new issues — looks safe to merge from my side.

---

phil-opp:
Heads-up: the red `ci/circleci: audit` on the latest commit is **not** from this change. This PR touches only `config.rs` and leaves `Cargo.lock` untouched, and `audit` passed on the parent commit (`be9b100`).

I reproduced it locally — `cargo audit` passes; `cargo deny check` fails because **`spin` 0.9.8 and 0.10.0 were yanked from crates.io** and `deny.toml` sets `yanked = "deny"`:

```
error[yanked]: detected yanked crate (try `cargo update -p spin`)
  spin 0.9.8  — yanked version
  spin 0.10.0 — yanked version
advisories FAILED
```

Both versions are transitive; the latest published is 0.12.2. This is an upstream yank against the unchanged lockfile, so it fails `main` and every open PR re-run right now — a repo-wide lockfile update (`cargo update -p spin@0.9.8` / `@0.10.0` to non-yanked releases), not something belonging in this timer fix. Happy to open a dedicated lockfile PR for it if that's useful; otherwise the timer change here can be evaluated on its own merits.

---

PR #2578, opened 2026-07-09
fix(core): reject ancestry and reflog git rev-specs in checkout_tree

phil-opp:
## Issue

`checkout_tree` (in `libraries/core/src/build/git.rs`) guards a git-source pin against rev-spec expressions so it resolves only to a literal commit hash or branch/tag name. The denylist covered `..`, `:`, and `^`, but not the `~` (ancestor) or `@{...}` (reflog / date) navigation operators:

```rust
if commit_hash.contains("..") || commit_hash.contains(':') || commit_hash.contains('^') { ... }
```

So a descriptor with `commit_hash: "main~5"` or `"HEAD@{yesterday}"` passed validation, and `revparse_ext` resolved it to a **different** commit than any literal ref — defeating the reproducibility the guard is meant to enforce.

## Fix

Extract the check into `contains_revspec_navigation` and add `~` and `@{` to the rejected set. A valid git ref name can never contain `~`, `^`, `:`, `..`, or `@{` (per `git-check-ref-format`), and neither can a hex commit hash, so no legitimate pin is newly rejected. All three `checkout_tree` call sites pass a user-supplied `commit_hash`, so every path is hardened uniformly.

## Validation

- New unit tests: `plain_refs_and_hashes_are_allowed` (main, `release-1.2`, `v1.0.0`, `feature/foo`, a 40-char hash, `HEAD`) and `revspec_navigation_is_rejected` (`main~5`, `HEAD@{yesterday}`, `HEAD@{1}`, `v1.0..v2.0`, `tag^2`, `main^`, `HEAD:src/main.rs`).
- `cargo test -p dora-core --features build build::git::tests` → 2 passed; `cargo fmt`/`clippy --features build` clean.

---

---

### DIFF
--- a/libraries/core/src/build/git.rs
+++ b/libraries/core/src/build/git.rs
@@ -350,9 +350,31 @@ async fn fetch_changes(
     Ok(repository)
 }
 
+/// Returns `true` if `commit_hash` contains git rev-spec *navigation* syntax.
+///
+/// A pinned commit reference is only meant to be a hex commit hash or a plain
+/// branch/tag name. Rev-spec operators instead resolve to a *different* commit
+/// than any literal ref, defeating the reproducibility of the pin:
+///
+/// * `a..b` / `a...b` — commit range
+/// * `ref:path`       — `<tree-ish>:<path>` blob/tree lookup
+/// * `ref^`, `ref^2`  — parent navigation
+/// * `ref~5`          — ancestor navigation
+/// * `ref@{...}`      — reflog / date lookup (e.g. `HEAD@{yesterday}`)
+///
+/// A legitimate git ref name can never contain `~`, `^`, `:`, `..`, or `@{`
+/// (see `git-check-ref-format`), so rejecting them cannot block a valid pin.
+fn contains_revspec_navigation(commit_hash: &str) -> bool {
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Verified: `checkout_tree` now routes through `contains_revspec_navigation`, which extends the previous `..`/`:`/`^` denylist with `~` (ancestor) and `@{` (reflog/date) navigation. Per `git-check-ref-format`, a valid branch/tag name cannot contain `~`, `^`, `:`, or the `@{` sequence, and a hex commit hash contains none of these, so the added rejections cannot block a legitimate pin — while `main~5` / `HEAD@{yesterday}` (which `revparse_ext` would otherwise resolve to a non-literal commit, defeating the pin) are now rejected. Tests cover both the accepted and rejected sets.

Minor note (non-blocking): this stays a denylist, so exotic aliases that don't contain any listed token — e.g. a bare `@` (resolves to `HEAD`) — still pass through. That's an edge case with no clear reproducibility impact, and an allowlist (hex-hash-or-valid-ref-name) would be the stricter long-term shape if this surface ever matters more.

---

phil-opp:
Closing as duplicate of #2490 (which is stricter)

PR #2579, opened 2026-07-09
fix(tracing): make trace-context codec robust to W3C tracestate values

phil-opp:
## Issue

`serialize_context` / `deserialize_to_hashmap` (in `libraries/extensions/telemetry/tracing/src/telemetry.rs`) joined context entries with `;` and split each key/value on `:`. Both bytes are legal inside a **W3C `tracestate` value** — the spec permits the byte ranges `0x20-0x2B`, `0x2D-0x3C`, `0x3E-0x7E`, which include `:` (`0x3A`) and `;` (`0x3B`).

So a propagated context carrying a vendor `tracestate` such as `vendor=a;b` was decoded into `{"tracestate": "vendor=a", "b": ""}` — the trace state truncated and a bogus key injected, corrupting distributed traces. An empty input also produced a spurious `{"": ""}` entry.

## Fix

- Use newline (`0x0A`, a control byte that can never appear in a `traceparent`/`tracestate` value) as the entry separator.
- Split key/value on the **first** `:` only (`split_once`), so a value containing `:` survives intact. The keys (`traceparent`/`tracestate`, after `baggage` is stripped) never contain `:`.
- Skip empty entries, fixing the spurious `{"": ""}`.

Both encode and decode sides change together. The codec is used symmetrically, in-process, between nodes of the same build (node/runtime/daemon), so there is no persisted/cross-version wire-format concern.

## Validation

- New unit tests: `:`/`;`-in-`tracestate` round-trips with no spurious entries, empty input yields an empty map, and a value-less entry defaults to `""`.
- `cargo test -p dora-tracing --lib telemetry::tests` → 3 passed; `cargo fmt`/`clippy` clean.

---

---

### DIFF
--- a/libraries/extensions/telemetry/tracing/src/telemetry.rs
+++ b/libraries/extensions/telemetry/tracing/src/telemetry.rs
@@ -52,10 +52,6 @@ pub fn init_tracing(_name: &str, endpoint: &str) -> eyre::Result<sdktrace::SdkTr
         .build())
 }
 
-/// Serialize the trace context (trace ID, span ID) into a compact string.
-/// Only W3C TraceContext keys (`traceparent`, `tracestate`) are included.
-/// OTel Baggage keys are stripped to prevent sensitive data from leaking
-/// across node boundaries in the dataflow.
 /// Separator between entries in the serialized trace context.
 ///
 /// The previous format joined entries with `;` and split key/value on `:`,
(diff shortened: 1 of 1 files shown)


phil-opp:

The fix looks correct. The old `;`-join / `:`-split codec did corrupt a `tracestate` value containing `:` or `;` (both are legal W3C `tracestate` value bytes, `0x3A`/`0x3B`), and newline (`0x0A`) is a control byte the spec disallows in `traceparent`/`tracestate` values, so it's a safe entry separator. `split_once(':')` keeps a colon-bearing value intact, and the empty-entry skip removes the spurious `{"": ""}` the old `split(';')` produced. Both encode and decode sides are updated together and the added tests cover the `:`/`;`-in-value, empty, and value-less cases.

One heads-up, not a blocker: this is a wire-format change for the serialized trace context passed between nodes. During a rolling upgrade, an old node emitting `;`-separated context to a new node (or vice versa) will mis-parse it — e.g. a new decoder reading an old `a:b;c:d` string treats the whole thing as one entry `key="a"`, `value="b;c:d"`. Since this only affects distributed-tracing metadata (best-effort telemetry, not dataflow correctness) and self-corrects once all nodes are on the new version, it's likely acceptable — just worth being aware of if mixed-version deployments are a concern.

---

PR #2581, opened 2026-07-09
fix(daemon): bound node log-line reads and stop stdout reader on error

phil-opp:
## Issue

The node stdout/stderr log readers (in `binaries/daemon/src/spawn/prepared.rs`) called `read_until(b'\n', &mut raw)`, which is **unbounded**. A malicious or buggy node that writes gigabytes with no newline grows `raw` until the daemon exhausts memory. `truncate_log_line` runs only *after* the whole line is materialized, so it cannot prevent this — defeating the OOM protection its own comment claims. Related defects in the same path:

- The stderr crash-diagnostics ring buffer (`node_stderr_most_recent`) stored the **raw, untruncated** line, retaining the full multi-GB string.
- The stdout reader set `finished = false` on a read error and looped again, so a persistent error (e.g. `EIO`) would **busy-spin** the task and flood the log channel. (The stderr reader already stopped on error.)

## Fix

- Add `read_capped_line`, which buffers at most `MAX_LOG_LINE_BYTES + 1` bytes per line (via `Take`) and discards the remainder of an over-long line through the reader's own buffer (`fill_buf`/`consume`, O(1) extra memory), so the next line is read intact.
- Use it for both readers, removing the duplicated read/decision blocks.
- Stop the stdout reader on a persistent read error instead of spinning, matching stderr.
- Store the **truncated** line in `node_stderr_most_recent`.

Behavior is otherwise preserved: still one `LogLine` per line, and the trailing empty `LogLine` at EOF is unchanged.

## Validation

- New unit tests: plain lines, a trailing newline-less line, and the cap-and-discard path (asserts an over-long line buffers ≤ `MAX_LOG_LINE_BYTES + 1` bytes and the following line is read intact).
- `cargo test -p dora-daemon --lib spawn::prepared::tests` → 3 passed; `cargo fmt`/`clippy` clean.

---

---

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -31,7 +31,7 @@ use std::{
 };
 use tokio::{
     fs::File,
-    io::{AsyncBufReadExt, AsyncWriteExt},
+    io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt},
     sync::{mpsc, oneshot},
 };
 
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

I traced `read_capped_line` and both listener tasks:
- `set_limit(MAX_LOG_LINE_BYTES + 1)` per call bounds each line's buffering; the over-limit discard loop uses `fill_buf`/`consume` on the `Take` with `limit = u64::MAX`, draining to the next `\n` in O(1) extra memory, then the next call resets the cap. Boundary cases (newline landing exactly at the cap → `raw.last() == Some(b'\n')`, EOF mid-line) are handled.
- `caps_and_discards_overlong_line` correctly asserts the over-long line is not split unbounded and the following line is read intact.
- The stderr ring buffer now stores the truncated `content` rather than the raw line, so it can't retain a multi-MB line.
- Returning `Ok(true)` on a persistent read error stops the busy-spin/log-flood; acceptable tradeoff (listener stops rather than looping).

The empty `LogLine` emitted on the final EOF iteration is pre-existing behavior, not introduced here.

---

phil-opp:
The merge-queue `E2E Tests` failure looks like an unrelated flake — a **re-queue should clear it**.

From the job log ([run 29762310566](https://github.com/dora-rs/dora/actions/runs/29762310566/job/88419662413)):
- `fault-tolerance-e2e`: **all 11 tests passed** (the suite that exercises this PR's node stdout/stderr path).
- The failure is in a different binary, `daemon-reconnect-e2e::node_survives_daemon_watchdog_disconnect`:
  ```
  coordinator: ERROR control request failed: no unnamed daemon connections  (run/mod.rs:158)
  daemon:      Connected to dora-coordinator at ws://127.0.0.1:33995/api/daemon
  panicked at daemon-reconnect-e2e.rs:456: dora start failed
  ```

That's a **daemon-registration race**: `dora start` reached the coordinator in the window after the daemon's WS connect but before it was registered as an unnamed daemon. It fails *before any node is spawned*, whereas this PR only changes how the daemon reads a node's stdout/stderr *after* spawn — so it can't produce "no unnamed daemon connections." Unrelated to this change; safe to re-run.

(Separately, the `ci/circleci: audit` red across all current PRs is the upstream `spin` yank noted above, also not from this change.)

---

PR #2587, opened 2026-07-10
fix(node): bound _zenoh_schema_subscribers teardown on shutdown

### LINKED ISSUE #2583: Node shutdown can still hang on a wedged zenoh runtime: `_zenoh_schema_subscribers` teardown is not bounded (#2439 fix left incomplete by #2366)
## Summary

The shutdown-hang fix in **#2439** (`fd1f050`) bounded the teardown of one per-input zenoh entity, but **#2366** (`81e5efe`, "Arrow IPC data plane"), merged around the same time, added a *second* per-input zenoh entity on the shared session — `_zenoh_schema_subscribers` — that is **not** covered by the timeout guard. As a result the exact node-shutdown hang tracked in #2425 remains reachable via the schema-subscriber undeclare path.

## Details

`EventStream` holds two per-input subscriber vectors on the shared zenoh session:

```rust
// apis/rust/node/src/event_stream/mod.rs
82:    _zenoh_subscribers: Vec<zenoh::pubsub::Subscriber<()>>,
85:    _zenoh_schema_subscribers: Vec<zenoh_ext::AdvancedSubscriber<()>>,
```

`EventStream::Drop` only tears down the first one under the deadline:

```rust
// apis/rust/node/src/event_stream/mod.rs:1543
let subscribers = std::mem::take(&mut self._zenoh_subscribers);
if !subscribers.is_empty() {
    let completed = teardown_with_timeout("zenoh-subscribers", ZENOH_TEARDOWN_TIMEOUT, move || {
        drop(subscribers);
    });
    ...
}
```

`_zenoh_schema_subscribers` is never `mem::take`n, so after the `Drop` body returns it is dropped **inline, unbounded**, during Rust's implicit field-drop phase (it is declared immediately after `_zenoh_subscribers`). `zenoh_ext::AdvancedSubscriber::drop` undeclares the subscription on the shared session — precisely the operation #2439's own code comment says "blocks indefinitely when zenoh's net runtime is wedged (e.g. stuck retrying an unreachable scouted peer)."

This is inconsistent with the **publisher** side, which #2366 got right. `DoraNode::Drop` takes *both* publisher maps and drops them together inside the guard:

```rust
// apis/rust/node/src/node/mod.rs:1809
let publishers = std::mem::take(&mut self.zenoh_publishers);
let schema_publishers = std::mem::take(&mut self.zenoh_schema_publishers);
...
let completed = teardown_with_timeout("zenoh", ZENOH_TEARDOWN_TIMEOUT, move || {
    drop(publishers);
    drop(schema_publishers);   // node/mod.rs:1826
});
```

The `EventStream` subscriber side was simply not updated to match when the schema plane was added.

## Risk / affected code paths

- Any node with **user-mapped inputs** (which declare `@schema` AdvancedSubscribers) shutting down while the zenoh net runtime is wedged can hang in `_zenoh_schema_subscribers`' field-drop — *before* `DoraNode::Drop` (which is bounded) runs — so the daemon never observes the node finishing and the dataflow stalls until an outer timeout.
- This is the same failure signature as #2425 (still open): record-replay `dora record` never returns; multiple-daemons "dataflow not finished after N retries". #2439 reduced the probability (one of the two subscriber sets is now bounded) but did not close the hang.
- Nodes with only timer inputs declare no subscribers and are unaffected.

## Suggested fix

Tear down `_zenoh_schema_subscribers` under the same guard as `_zenoh_subscribers`, mirroring the publisher side — e.g. `mem::take` both and drop them inside a single `teardown_with_timeout("zenoh-subscribers", ...)` closure.

## Verification

`teardown_with_timeout` (`apis/rust/node/src/node/mod.rs:1774`) runs the closure on a dedicated OS thread and returns on `recv_timeout`, so a wedged tokio/zenoh runtime does not defeat the deadline — the guard is effective for whatever it wraps; the schema subscribers are simply outside it. `std::mem::take` leaves an empty `Vec`, so extending the same pattern introduces no double-undeclare.

---

phil-opp:
## Summary

Fixes #2583.

The node-shutdown-hang guard added in #2439 only wrapped `_zenoh_subscribers` in `EventStream::Drop`. The Arrow IPC data plane (#2366) then added a **second** per-input zenoh entity on the shared session — `_zenoh_schema_subscribers` (the `@schema` `AdvancedSubscriber`s) — which was left **outside** the `teardown_with_timeout` deadline. It was dropped inline during Rust's implicit field-drop phase, after the `Drop` body returns. Its `undeclare` on the shared session blocks indefinitely when zenoh's net runtime is wedged (stuck retrying an unreachable scouted peer), reopening the exact node-shutdown hang #2439 set out to fix — the #2425 signature (record-replay `dora record` never returns; multi-daemon "dataflow not finished after N retries").

This was inconsistent with the **publisher** side, which #2366 got right: `DoraNode::Drop` takes *both* publisher maps and drops them together inside one guard.

## Change

Take both subscriber vectors and drop them together inside a single `teardown_with_timeout("zenoh-subscribers", ...)` closure in `EventStream::Drop`, mirroring the publisher side. `std::mem::take` leaves empty `Vec`s behind, so the later implicit field-drop is a no-op and there is no double-undeclare.

## Testing

- `cargo check -p dora-node-api`
- `cargo clippy -p dora-node-api -- -D warnings`

Nodes with only timer inputs declare no subscribers and are unaffected (the guard is skipped when both vectors are empty).

---

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -1540,11 +1540,20 @@ impl Drop for EventStream {
         // finish and the dataflow stalls until an outer timeout (dora-rs/dora#2425).
         // The callbacks use `try_send`, so undeclaring before `receiver` drops
         // cannot deadlock on a blocked callback.
+        //
+        // The `@schema` `AdvancedSubscriber`s (added with the Arrow IPC data
+        // plane in #2366) undeclare on the same shared session and can wedge
+        // the same way, so they must be torn down under the same deadline —
+        // mirroring `DoraNode::Drop`, which drops both publisher maps inside
+        // one guard. Left out, they would otherwise drop unbounded in the
+        // implicit field-drop phase after this `Drop` body returns (#2583).
         let subscribers = std::mem::take(&mut self._zenoh_subscribers);
-        if !subscribers.is_empty() {
+        let schema_subscribers = std::mem::take(&mut self._zenoh_schema_subscribers);
+        if !subscribers.is_empty() || !schema_subscribers.is_empty() {
             let completed =
                 teardown_with_timeout("zenoh-subscribers", ZENOH_TEARDOWN_TIMEOUT, move || {
                     drop(subscribers);
+                    drop(schema_subscribers);
                 });
             if !completed {
                 tracing::warn!(


phil-opp:

No issues found.

I verified `_zenoh_schema_subscribers` exists on `EventStream` (`event_stream/mod.rs:85`), and the change `mem::take`s both subscriber vectors and drops them together inside the single `teardown_with_timeout` closure, so the `@schema` `AdvancedSubscriber`s are now bounded by `ZENOH_TEARDOWN_TIMEOUT` like the data subscribers. `mem::take` leaves empty `Vec`s, so the implicit field-drop after `Drop` returns is a no-op — no double-undeclare. The guard now fires when either vector is non-empty. This correctly mirrors `DoraNode::Drop`'s handling of both publisher maps and closes the reopened #2425/#2583 hang.

---

PR #2588, opened 2026-07-10
fix(node): bound the resulting Arrow IPC stream length in encode guards

### LINKED ISSUE #2586: Arrow IPC encode/decode size guards are asymmetric: large UInt8 payloads encode but are rejected on receive (silent drop)
## Summary

In the Arrow IPC data plane added by **#2366** (`81e5efe`), the **encode** guard bounds the *input `data_len`* against `MAX_IPC_BYTES`, but the emitted IPC stream is ~1.125× larger, and every **receiver** bounds the *whole stream* against the same `MAX_IPC_BYTES`. A narrow band of large payloads therefore encodes and is sent successfully, then is rejected by all receivers — silently dropped on the zenoh path.

## Details

Encode guards only `data_len`:

```rust
// apis/rust/node/src/node/arrow_utils/ipc_encode.rs:468
fn uint8_layout(data_len: usize) -> eyre::Result<Uint8Layout> {
    if data_len > super::MAX_IPC_BYTES {           // guards the INPUT length
        bail!("UInt8 payload too large: {data_len} bytes ...");
    }
    ...
    let total = schema_block + record_batch_block + body_len + PREFIX_LEN;  // strictly > data_len
```

`total` includes the validity bitmap (`data_len/8`), 64-byte alignment padding, and the schema + record-batch message blocks — roughly `1.125 * data_len`. But receivers guard the **entire stream**:

```rust
// apis/rust/node/src/node/arrow_utils.rs:112  decode_arrow_ipc_zero_copy
if buffer.len() > MAX_IPC_BYTES {
    bail!("Arrow IPC payload too large: {} bytes (max {MAX_IPC_BYTES})", buffer.len());
}
// ipc_encode.rs:723  check_ipc_size(len)  — same whole-stream bound
```

With `MAX_IPC_BYTES = 256 MiB`, a `send_output_raw` `UInt8` payload with `data_len` in roughly **(227.5 MiB, 256 MiB]** passes the encode guard (the producer allocates the buffer and sends it), but the resulting stream exceeds 256 MiB and every receiver bails with "Arrow IPC payload too large." On the zenoh receive path this becomes a warn-and-drop (`decode_zenoh_sample` → `Err`), i.e. the input is **silently lost**; on the daemon path it errors. `ipc_fast_path_len` can likewise return a value `> MAX_IPC_BYTES`.

## Risk / scope

Niche — single messages that large also tend to collide with SHM-pool / `MAX_MESSAGE_BYTES` limits in most configurations — but it is a genuine internal inconsistency: the encode limit and the decode limit disagree, and the failure mode on the distributed path is a silent drop rather than a producer-side error.

## Suggested fix

Bound the *resulting stream length* (`total` / the fast-path length) against `MAX_IPC_BYTES` in the encode guards, not `data_len`/the body alone, so that anything the producer accepts is decodable by every receiver (or, conversely, fails loudly at the producer).

---

phil-opp:
## Summary

Fixes #2586.

The Arrow IPC data plane (#2366) bounds the **input** `data_len` against `MAX_IPC_BYTES` on encode, but the emitted IPC stream is ~1.125× larger (validity bitmap, 64-byte alignment padding, schema + record-batch message blocks), while every **receiver** bounds the **whole stream** against the same `MAX_IPC_BYTES`. A narrow band of `data_len` just under 256 MiB therefore encoded and was sent successfully, then was rejected by every receiver:

- on the zenoh path → warn-and-drop (`decode_zenoh_sample` → `Err`): the input is **silently lost**;
- on the daemon path → error.

The encode limit and the decode limit disagreed.

## Change

Bound the **resulting stream length** rather than the input:

- **`uint8_layout`** (the `send_output_raw` / Python zero-copy `UInt8` "construct in place" path) now rejects when the computed `total` exceeds `MAX_IPC_BYTES`. The cheap `data_len` pre-filter is kept only to bound the layout arithmetic (`total` is always ≥ `data_len`).
- **`prepare`** (the generic fast path) declines (`None`) when `total` exceeds the limit, routing to the fallback encoder — which bails loudly at the producer — instead of emitting a stream every decoder rejects. `push_buffer` previously only bounded the running *body* length, so a body just under the limit could still yield an over-limit stream once the header blocks are added.

Now anything a producer accepts is decodable by every receiver, or it fails loudly at the producer.

## Testing

- New regression test `uint8_ipc_len_bounds_resulting_stream_not_input` — binary-searches the accept/reject boundary and asserts (a) `data_len == MAX_IPC_BYTES` is rejected, (b) the largest accepted payload encodes to a stream `<= MAX_IPC_BYTES`, and (c) the boundary sits *below* the input limit (i.e. it is set by the stream size, not `data_len`). Allocation-free — `uint8_layout` computes purely from `data_len`.
- `cargo test -p dora-node-api` (arrow_utils suite), `cargo clippy -p dora-node-api -- -D warnings`, `cargo fmt --all -- --check`.

---

### DIFF
--- a/apis/rust/node/src/node/arrow_utils/ipc_encode.rs
+++ b/apis/rust/node/src/node/arrow_utils/ipc_encode.rs
@@ -309,6 +309,15 @@ fn prepare(array: &ArrayData) -> Option<Prepared> {
     let record_batch_block = round_up(PREFIX_LEN + record_batch_message.len(), ALIGN);
     // schema block + record-batch header block + body + end-of-stream (8 bytes).
     let total = schema_block + record_batch_block + layout.body_len + PREFIX_LEN;
+    // `push_buffer` only bounds the running body length; the schema and
+    // record-batch header blocks push `total` slightly higher, so a body just
+    // under MAX_IPC_BYTES can still yield a stream over it. Bound the whole
+    // stream here so the fast path declines (routing to the fallback encoder,
+    // which bails loudly) rather than emitting a stream every receiver rejects
+    // — matching the whole-stream limit the decoders enforce (#2586).
+    if total > super::MAX_IPC_BYTES {
+        return None;
+    }
     Some(Prepared {
         layout,
         schema_message,
@@ -466,6 +475,10 @@ struct Uint8Layout {
 }
 
 fn uint8_layout(data_len: usize) -> eyre::Result<Uint8Layout> {
+    // Cheap pre-filter: the emitted stream is always >= `data_len`, so anything
+    // over the limit is rejected below regardless, and bounding `data_len` here
+    // keeps the round_up/add arithmetic that computes `total` well within
+    // `usize` range.
     if data_len > super::MAX_IPC_BYTES {
         bail!(
             "UInt8 payload too large: {data_len} bytes (max {})",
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed the diff and traced the encode/send paths in `apis/rust/node/src`.

The **UInt8 path** change in `uint8_layout` is correct: it computes `total` and `bail!`s loudly when the resulting stream exceeds `MAX_IPC_BYTES`, so `send_output_raw`/the Python zero-copy path fail at the producer. Good.

The **generic `prepare` path** change does **not** achieve what its comment claims, though. The comment says returning `None` routes to "the fallback encoder, which bails loudly". But the fallback is:

- `send_output_array` (`node/mod.rs:894`) → `None` branch → `ipc_encode::encode_ipc_to_vec` (`ipc_encode.rs:451`) → `arrow_utils::encode_arrow_ipc` (`arrow_utils.rs:26`).
- `encode_arrow_ipc` has **no `MAX_IPC_BYTES` guard** — only `decode_arrow_ipc` (`arrow_utils.rs:65`) does. It happily encodes the oversized stream via the official `StreamWriter`, and `send_output_array` then allocates a sample of `bytes.len()` and sends it.

So for a non-UInt8 primitive array in the target band (body ≤ limit but `total` > limit), the stream is still produced and sent — silently dropped on the zenoh path, exactly the #2586 symptom — rather than bailing loudly. Behavior for that band is unchanged by this PR (it merely moves from the fast path to the 2-copy fallback), and #2586 remains unfixed for non-UInt8 outputs.

Suggestion: add the whole-stream `MAX_IPC_BYTES` check to the fallback branch of `send_output_array` (or to `encode_arrow_ipc`/`encode_ipc_to_vec`) so the generic path fails loudly too, and update the comment which currently overstates the fix.

---

PR #2589, opened 2026-07-10
fix(daemon,cli): report true on-wire size in `dora topic info`

### LINKED ISSUE #2584: `dora topic info` over-reports bandwidth for schema-once outputs (per-frame schema block double-counted)
## Summary

After **#2366** (`81e5efe`, "Arrow IPC data plane"), `dora topic info` reports inflated bandwidth and total-bytes for small/primitive ("schema-once") outputs, because the daemon prepends the full Arrow schema block to *every* debug frame before the CLI measures it — even though on the wire that schema block travels only once.

## Details

Small node→node outputs use the schema-once path: the schema is published a single time on the `@schema` subtopic and each data sample carries only the schema-less record batch. For inspection, the daemon rebuilds a full self-describing stream by prepending the retained schema to each schema-once batch:

```rust
// binaries/daemon/src/lib.rs:5644  rebuild_debug_topic_stream(...)
let (_, schema) = cache.iter().find(|(h, _)| *h == hash as u64)?;
let mut full = Vec::with_capacity(schema.len() + payload.len());
full.extend_from_slice(schema);   // schema block prepended to EVERY frame
full.extend_from_slice(payload);
Some(full)
```

`dora topic info` then measures the length of that rebuilt stream:

```rust
// binaries/cli/src/command/topic/info.rs:181
let data_size = data.as_ref().map(|d| d.len()).unwrap_or(0);
...
stats_clone.record(data_size, data_type, Instant::now());
```

So for, say, an 8-byte `UInt64` output emitted at high rate, each measured frame is `schema_block (~200+ B) + batch (tens of B)`, and the ~200 B schema is counted on every message. Reported bandwidth / total bytes come out several-to-many× the true network usage. Under the pre-#2366 raw-buffer format `data_size` was ≈ the payload bytes, so this is an accuracy regression.

## Risk / scope

Inspection/diagnostics only — no data-plane impact. But `dora topic info` is a bandwidth-measurement tool, and it now materially misreports for exactly the case (small frequent primitive messages) where the schema-once optimization matters most, which can mislead capacity/latency investigations.

## Suggested fix

Measure the on-wire size rather than the rebuilt stream — e.g. subtract the prepended schema length in `rebuild_debug_topic_stream`'s consumer, have the daemon forward the original batch length alongside the rebuilt stream, or record `payload.len()` (plus a one-time schema cost) instead of the full rebuilt buffer length.

---

phil-opp:
## Summary

Fixes #2584.

After the Arrow IPC data plane (#2366), `dora topic info` over-reports bandwidth and total-bytes for schema-once (small/primitive) outputs. The daemon rebuilds a self-describing Arrow IPC stream for inspection by prepending the retained schema block to **every** debug frame (`rebuild_debug_topic_stream`), but on the wire that schema block travels only **once** (published on the `@schema` subtopic); each data sample carries only the schema-less record batch. `dora topic info` measured the length of the *rebuilt* stream, so the ~200+ B schema was counted on every message — reported bandwidth / total bytes came out several-to-many× the true network usage for exactly the small, frequent primitive messages the schema-once optimization matters most for. Under the pre-#2366 raw-buffer format the measured size was ≈ the payload bytes, so this is an accuracy regression.

## Change

Measure the real on-wire size instead of the rebuilt stream:

- **daemon**: stamps the original data-sample length on the debug frame's metadata under a new debug-only `_wire_size` key (`dora_message::metadata::WIRE_SIZE`). For a full self-describing frame this equals `data.len()`; for a schema-once batch it excludes the prepended schema block.
- **cli**: `dora topic info` records `_wire_size` when present, falling back to the buffer length otherwise (robust to older daemons / non-debug frames).

The size is carried in **metadata** (set only on the inspection path) rather than added as a field to the shared `InterDaemonEvent::Output` wire struct — so the real data-forwarding hot path gains **no** per-message overhead, and the change stays confined to the debug/inspection path.

## Testing

- New `dora-daemon` test `on_wire_size_excludes_prepended_schema` — asserts the stamped wire size (the data-sample `payload.len()`) excludes the schema block prepended by `rebuild_debug_topic_stream` for a schema-once frame, and equals the full length for a self-describing frame.
- `cargo test -p dora-daemon --lib debug_topic_tests` (8 passed), `cargo clippy -p dora-message -p dora-cli -- -D warnings`, `cargo fmt --all -- --check`.

---

### DIFF
--- a/binaries/cli/src/command/topic/info.rs
+++ b/binaries/cli/src/command/topic/info.rs
@@ -178,19 +178,14 @@ fn info(
                     };
                     match event.inner {
                         InterDaemonEvent::Output { data, metadata, .. } => {
-                            // Prefer the daemon-stamped on-wire size: the debug
-                            // frame's `data` is a rebuilt self-describing stream
-                            // with the schema prepended to every frame, but a
-                            // schema-once output ships that schema only once, so
-                            // `data.len()` over-reports bandwidth. Fall back to the
-                            // buffer length when the size isn't present (#2584).
-                            let data_size = dora_message::metadata::get_integer_param(
+                            // Charge the real on-wire size, not the rebuilt
+                            // self-describing `data` (which re-prepends the schema
+                            // a schema-once output ships only once) — see
+                            // `debug_frame_wire_size` (#2584).
+                            let data_size = dora_message::metadata::debug_frame_wire_size(
                                 &metadata.parameters,
-                                dora_message::metadata::WIRE_SIZE,
-                            )
-                            .and_then(|n| usize::try_from(n).ok())
-                            .or_else(|| data.as_ref().map(|d| d.len()))
-                            .unwrap_or(0);
+                                data.as_deref(),
+                            );
                             // The payload is a self-describing Arrow IPC stream;
                             // read its data type from the decoded array (best
                             // effort — a malformed stream just leaves it unknown).
--- a/binaries/cli/src/command/topic/info.rs
+++ b/binaries/cli/src/command/topic/info.rs
@@ -177,8 +177,20 @@ fn info(
                         }
                     };
                     match event.inner {
-                        InterDaemonEvent::Output { data, .. } => {
-                            let data_size = data.as_ref().map(|d| d.len()).unwrap_or(0);
+                        InterDaemonEvent::Output { data, metadata, .. } => {
+                            // Prefer the daemon-stamped on-wire size: the debug
+                            // frame's `data` is a rebuilt self-describing stream
+                            // with the schema prepended to every frame, but a
+                            // schema-once output ships that schema only once, so
+                            // `data.len()` over-reports bandwidth. Fall back to the
+                            // buffer length when the size isn't present (#2584).
+                            let data_size = dora_message::metadata::get_integer_param(
+                                &metadata.parameters,
+                                dora_message::metadata::WIRE_SIZE,
+                            )
+                            .and_then(|n| usize::try_from(n).ok())
+                            .or_else(|| data.as_ref().map(|d| d.len()))
+                            .unwrap_or(0);
                             // The payload is a self-describing Arrow IPC stream;
                             // read its data type from the decoded array (best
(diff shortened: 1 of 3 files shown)


phil-opp:

No issues found.

I verified:
- The daemon stamps `WIRE_SIZE` from `payload.len()` (`daemon/src/lib.rs:~3012`), which is the actual on-wire `sample.payload()` (the schema-less batch for schema-once, the full stream otherwise) — so it is the true on-wire size, and the insertion happens on the `DebugTopicData` inspection path only, after `strip_internal_parameters`, so it never touches the real node→node hot path.
- `WIRE_SIZE` (`_wire_size`) is not removed by `strip_internal_parameters` (which only drops `SCHEMA_HASH`/`FRAMING`), so it survives to the CLI.
- The CLI (`topic/info.rs`) reads `get_integer_param(WIRE_SIZE)`, `usize::try_from`, then falls back to `data.len()` when absent — matching pre-#2366 accounting.

Minor (non-blocking): `_wire_size` will appear among the frame's metadata parameters if `dora topic info` ever prints them verbatim; harmless but internal.

---

phil-opp:

Reviewed the newest commit (`0dfb8ac`, the test/refactor follow-up). It extracts the CLI's read+fallback into a shared `debug_frame_wire_size` helper in `dora-message` (colocated with `WIRE_SIZE`) and reworks the daemon test to replicate the real stamp sequence (rebuild → strip internal keys → stamp the input payload length) instead of only re-asserting `rebuild_debug_topic_stream`'s existing invariant. The stamp still records `payload.len()` (the schema-less on-wire sample) and the CLI charges `debug_frame_wire_size(&metadata.parameters, data.as_deref())` with a correct precedence (stamped value → `usize::try_from` → buffer length → 0). The unit tests genuinely exercise the fix — the out-of-range/absent/present cases would fail if the fallback logic were broken. No new issues, and nothing outstanding from the earlier review. The latest commits look safe.

---

PR #2592, opened 2026-07-10
fix: correct signal-name mapping for signals 22 and 23 in NodeError Display

### LINKED ISSUE #2459: Incorrect signal-name mapping in NodeError Display: signal 22 shown as "SIGABRT", signal 23 as "NSIG"

## Summary

The signal-number → signal-name lookup used in `NodeError`'s `Display` implementation maps two signal numbers to the wrong names. When a dora node is killed by signal 22 or 23, the error message shown to the user reports a misleading signal name.

## Location

`libraries/message/src/common.rs`, in `impl std::fmt::Display for NodeError` (the `NodeExitStatus::Signal` arm):

```rust
NodeExitStatus::Signal(signal) => {
    let signal_str: Cow<_> = match signal {
        1 => "SIGHUP".into(),
        2 => "SIGINT".into(),
        3 => "SIGQUIT".into(),
        4 => "SIGILL".into(),
        6 => "SIGABRT".into(),
        8 => "SIGFPE".into(),
        9 => "SIGKILL".into(),
        11 => "SIGSEGV".into(),
        13 => "SIGPIPE".into(),
        14 => "SIGALRM".into(),
        15 => "SIGTERM".into(),
        22 => "SIGABRT".into(),   // <-- wrong
        23 => "NSIG".into(),      // <-- wrong
        other => other.to_string().into(),
    };
    ...
```

## The problem

On Linux (the primary platform / CI target where the daemon spawns and reaps nodes):

- **Signal 22 is `SIGTTOU`**, not `SIGABRT`. `SIGABRT` is signal **6**, which is already correctly mapped on the line above (`6 => "SIGABRT"`). So `SIGABRT` is now emitted for two different signal numbers, and a real `SIGTTOU` termination is mislabeled.
- **Signal 23 is `SIGURG`**, not `NSIG`. `NSIG` is not a deliverable signal at all — it is the libc constant for "number of signals" (a count/upper bound), so it can never legitimately appear as a termination cause.

These two entries look like a transcription/copy-paste error: the surrounding table lists the common low-numbered signals correctly, then appends `22`/`23` with incorrect names.

## Impact

Diagnostic-only (no control-flow effect), but user-facing and misleading. A user debugging why their node died sees, e.g.:

```
exited because of signal SIGABRT
```

when the node was actually terminated by `SIGTTOU` (signal 22), or:

```
exited because of signal NSIG
```

for signal 23 (`SIGURG`). This sends anyone debugging a crash down the wrong path.

## Suggested fix

Correct the two entries to the proper Linux names:

```rust
22 => "SIGTTOU".into(),
23 => "SIGURG".into(),
```

Optionally, the table could be filled out further (signals 5, 7, 10, 12, 16–21, 24+ are currently unmapped and fall through to the numeric `other` arm, which is acceptable), or a battle-tested crate could be used for the mapping. Note also that signal numbers above ~15 differ between Linux and macOS/BSD (e.g. macOS `SIGURG` is 16, and 23 is `SIGIO`), so a fully correct implementation would ideally be platform-aware — but at minimum the two entries above should not report `SIGABRT`/`NSIG`.
</body>

mvanhorn:
## Summary
Edit the `NodeExitStatus::Signal` match arm in `impl std::fmt::Display for NodeError` in libraries/message/src/common.rs (lines 143-144), changing `22 => "SIGABRT".into()` to `22 => "SIGTTOU".into()` and `23 => "NSIG".into()` to `23 => "SIGURG".into()`, matching the Linux signal numbering used by the rest of the table. Keep the change minimal per the issue's "at minimum" guidance - do not expand the table to other signal numbers or introduce platform-conditional mapping, since the maintainer flagged those as optional extras. Add unit tests in the existing `mod tests` block at the bottom of the same file asserting the Display output for `NodeExitStatus::Signal(22)`, `Signal(23)`, and an unmapped number.

## Why this matters
The `Display` implementation for `NodeError` in `libraries/message/src/common.rs` contains a signal-number to signal-name lookup table used when a dora node is killed by a signal. Two entries are wrong: signal 22 is mapped to "SIGABRT" (which is actually signal 6, already correctly mapped) and signal 23 is mapped to "NSIG" (a libc count constant, not a deliverable signal). On Linux, signal 22 is SIGTTOU and signal 23 is SIGURG, so users debugging a node crash see a misleading signal name. The issue was filed by the repo's primary maintainer (phil-opp) with the exact fix spelled out, has bug/rust/daemon labels, no assignee, no comments, and no prior PR attempts.

See https://github.com/dora-rs/dora/issues/2459.

## Testing
- Display of a NodeError with `NodeExitStatus::Signal(22)` contains "SIGTTOU" (happy path for the first fixed entry). - Display of a NodeError with `NodeExitStatus::Signal(23)` contains "SIGURG" (happy path for the second fixed entry). - Display of a NodeError with `NodeExitStatus::Signal(6)` still contains "SIGABRT" (regression guard: the correct entry is untouched and no longer duplicated). - Display of a NodeError with an unmapped signal number (e.g.

Fixes #2459

### DIFF
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -140,8 +140,8 @@ impl std::fmt::Display for NodeError {
                     13 => "SIGPIPE".into(),
                     14 => "SIGALRM".into(),
                     15 => "SIGTERM".into(),
-                    22 => "SIGABRT".into(),
-                    23 => "NSIG".into(),
+                    22 => "SIGTTOU".into(),
+                    23 => "SIGURG".into(),
                     other => other.to_string().into(),
                 };
                 if matches!(self.cause, NodeErrorCause::GraceDuration) {
@@ -387,6 +387,35 @@ pub struct BinaryPin {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    fn node_error_with_signal(signal: i32) -> NodeError {
+        NodeError {
+            timestamp: uhlc::HLC::default().new_timestamp(),
+            cause: NodeErrorCause::Other {
+                stderr: String::new(),
+            },
+            exit_status: NodeExitStatus::Signal(signal),
+        }
+    }
+
+    #[test]
+    fn node_error_signal_display_uses_linux_signal_names() {
+        for (signal, name) in [(6, "SIGABRT"), (22, "SIGTTOU"), (23, "SIGURG")] {
+            assert_eq!(
+                node_error_with_signal(signal).to_string(),
+                format!("exited because of signal {name}")
+            );
+        }
+    }
+
+    #[test]
+    fn node_error_signal_display_falls_back_to_signal_number() {
+        assert_eq!(
+            node_error_with_signal(40).to_string(),
+            "exited because of signal 40"
+        );
+    }
+
     #[test]
     fn test_log_message_serialization() {
         let log_message = LogMessage {


phil-opp:
No issues found.

I verified the signal-number mapping against Linux signal names. Signal 6 is already correctly mapped to `SIGABRT` (further up the table), so the old `22 => "SIGABRT"` and `23 => "NSIG"` entries were both wrong. On Linux, signal 22 is `SIGTTOU` and 23 is `SIGURG`, which is exactly what this change now emits, and `NSIG` is a signal *count* rather than an actual deliverable signal, so removing it is correct. The added tests cover the corrected names plus the numeric fallback path (`other => number`).

This is a display-only, best-effort mapping and the table is already Linux-oriented, so I don't see any regression risk here.

---
*🤖 Automated review by Claude. This review was generated fully automatically and may contain mistakes — please verify independently before relying on it.*

---

PR #2593, opened 2026-07-11
fix(core): resolve single-`operator:` output types in pre-build type check

phil-opp:
## Summary

`check_type_annotations_full` (run by `dora validate` and `dora build`) operates on the **module-expanded but not alias-resolved** descriptor — the pre-build path never calls `resolve_aliases_and_set_defaults`. For a single-`operator:` node it registered the output type under the `op/`-prefixed key `(node, "op/result")`, but a consumer references that output in **short form** in the source YAML (`reading: producer/result`), yielding the lookup key `(node, "result")`. The keys never matched.

**Effect:** for an edge from a single-`operator:` producer, the upstream type was invisible:
- genuine type mismatches went **unreported**,
- no type was **inferred** across the edge,
- under `strict_types`, a spurious `"upstream ... has no type annotation"` warning fired even though the upstream *was* annotated.

Custom-node producers (registered under the bare `output_id`) and multi-`operators:` producers (consumers write the full `node/op/out`) were unaffected — only the singular `operator:` form, which had no test coverage.

## Fix

Register the operator output under **both** the `op/`-prefixed key and the bare `output_id`, matching how consumers reference single-operator outputs before alias resolution.

## Validation

- Adds 3 regression tests (inference, mismatch detection, strict-mode false-positive) that **fail without the fix** and pass with it.
- `cargo test -p dora-core`, `cargo clippy -p dora-core -- -D warnings`, `cargo fmt --all --check` all pass.

---

---

### DIFF
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -893,8 +893,18 @@ pub fn check_type_annotations_full(
             );
             for (output_id, urn) in &op.config.output_types {
                 if registry.resolve(urn).is_some() {
+                    // A consumer of a single-`operator:` node references its
+                    // output in short form (`node/output`) in the source YAML;
+                    // the `op/` prefix is only injected later by
+                    // `resolve_aliases_and_set_defaults`, which this pre-build
+                    // check never runs. Register both the prefixed key and the
+                    // bare `output_id` so short-form edges resolve here too —
+                    // otherwise the upstream type is invisible, silently
+                    // dropping genuine mismatches and, under `strict`, emitting
+                    // a false "upstream has no type annotation" warning.
                     output_type_map
                         .insert((nid.clone(), format!("{op_id}/{output_id}")), urn.clone());
+                    output_type_map.insert((nid.clone(), output_id.to_string()), urn.clone());
                 }
             }
             check_port_types(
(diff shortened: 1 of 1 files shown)


phil-opp:

This PR is one of a batch opened by an automated Claude code-review pass over `origin/main`. Each finding was verified against current `main`, tested (RED→GREEN where applicable), and checked against prior runs' open PRs to avoid duplicates. Per-branch verification: `fmt` + `clippy -D warnings` + affected-crate tests. Combined-diff verification: full workspace `cargo test` (one unrelated pre-existing failure — a zenoh test needing IPv6, which the sandbox lacks), `cargo clippy --all -D warnings`, `cargo check --examples`, plus the `/review` and `/simplify` gates (both clean).

### PRs opened this run

**Correctness**
| PR | Area | Issue |
|----|------|-------|
| #2593 | `dora-core` | Pre-build type check missed single-`operator:` producer edges (short-form refs) — mismatches silently unreported, false strict-mode warnings |
| #2594 | `dora-node-api` | Fallback Arrow IPC encoder had no size cap → producer sends >256 MB payloads the receiver silently rejects (data loss) |
| #2595 | `dora-runtime` | Unsupported operator source (WASM / Python-without-feature) returned `Ok(())`, dropping `init_done` → misleading "channel closed" error |
| #2596 | `dora-message` | `queue_size: 0` + `DropOldest` → cap 0 → 100% event drop, operator starves / dataflow hangs |
| #2597 | `dora-coordinator` | `dora logs` re-derived daemon from `deploy.machine` for running dataflows → fails with >1 unnamed daemon; now uses authoritative `node_to_daemon` |

**Consistency / usability**
| PR | Area | Issue |
|----|------|-------|
| #2598 | `dora-node-api-c` | `dora_send_output` rejected the `(NULL, 0)` empty-payload idiom that the operator FFI accepts (also avoids `from_raw_parts(NULL,0)` UB) |

### Findings investigated but **not** turned into PRs (rejected)

- **`topics.rs` `connect/endpoints` clobber** (`open_zenoh_session_with_listen`): the coordinator-address insert would overwrite the `DORA_ZENOH_CONNECT` / inter-daemon endpoints, but **every** call site passes `coordinator_addr = None`, so it's latent/unreachable today — not a live defect.
- **`zenoh_publish` unchecked `copy_from_slice`** (`node/mod.rs`): a latent panic only if a future zenoh version over-allocates the SHM buffer; safe under zenoh 1.x, and the sibling `allocate_data_sample` already guards it. Defensive-only, no live bug.
- **Runtime early `?` skips `main_task.join()`** (`runtime/lib.rs`): real but minor — the orphaned thread self-terminates as the process exits; join-restructure adds control-flow risk for little gain. Folded conceptually into #2595's scope but not fixed.
- **Coordinator daemon-reconnect shares the CLI per-IP rate limiter**: plausible lockout of a flapping remote daemon, but the fix touches security-sensitive rate-limiting policy — needs a maintainer decision, not an automated patch.
- **`build_dataflow` partial multi-daemon failure orphans in-progress builds**: same class as the already-known stop-partial issue; fix is non-trivial (cancel/finalize semantics) and better scoped as a deliberate change.
- **`runtime` `queue_size: 0` in the node scheduler** (`event_stream/scheduler.rs`): same root cause as #2596 and fixed centrally by that PR's `effective_cap` clamp — not a separate PR.

Each PR is independent and can be reviewed/merged on its own. All are machine-generated; please review before merging.

*— Claude Code (automated review)*

---

phil-opp:

No issues found.

Registering a single-`operator:` output under both `(nid, "op_id/output_id")` and the bare `(nid, "output_id")` key matches how consumers reference the output in the pre-alias-resolution descriptor, so short-form edges now resolve. The extra bare key is harmless for multi-`operators:` producers (consumers there use the full `op_id/output` form, so the bare key is never looked up). Inference, mismatch, and strict-mode false-positive paths are covered by the three added tests.

---

PR #2594, opened 2026-07-11
fix(node): cap fallback Arrow IPC encode to fail loudly on oversized output

phil-opp:
## Summary

The send and receive paths disagreed on the 256 MB `MAX_IPC_BYTES` limit, causing **silent data loss**.

- Every **receive** path rejects oversized payloads: `decode_arrow_ipc`, `decode_arrow_ipc_zero_copy`, and the streaming `InputDecoder`'s `check_ipc_size` all bail above the cap.
- The **fast-path encoder** refuses oversized arrays too — `Layout::push_buffer` returns `None`, routing them to the fallback.
- But the **fallback** `encode_arrow_ipc` (reached via `encode_ipc_to_vec` for oversized or non-fast-path arrays) had **no size check**, so `send_output`/`send_output_array` happily produced and sent a >256 MB stream.

**Result:** the send succeeds, but the consumer unconditionally rejects the payload with `"Arrow IPC payload too large"` and drops the input — silent data loss with **no error on the producer**. This is inconsistent with `send_output_raw`, whose `uint8_layout` already returns a clear producer-side error for the same size.

## Fix

Enforce `MAX_IPC_BYTES` in `encode_arrow_ipc`, so any oversized output fails at the source with an actionable message instead of being silently undecodable downstream. All callers propagate `eyre::Result`, so normal-size payloads are unaffected.

## Validation

- Adds a regression test (`ipc_encode_rejects_oversized_payload`) that **fails without the guard**.
- All 41 `arrow_utils` tests pass; `cargo clippy -p dora-node-api -- -D warnings` and `cargo fmt --all --check` pass.

---

---

### DIFF
--- a/apis/rust/node/src/node/arrow_utils.rs
+++ b/apis/rust/node/src/node/arrow_utils.rs
@@ -51,6 +51,22 @@ pub fn encode_arrow_ipc(arrow_array: &ArrayData) -> eyre::Result<Vec<u8>> {
             .finish()
             .context("failed to finish Arrow IPC stream")?;
     }
+
+    // Fail loudly at the producer instead of emitting a stream that every
+    // receive path will unconditionally reject. `decode_arrow_ipc`,
+    // `decode_arrow_ipc_zero_copy`, and the streaming `InputDecoder` all bail
+    // on payloads over `MAX_IPC_BYTES`, and the fast-path encoder refuses
+    // oversized arrays too (routing them here). Without this check an
+    // oversized array would encode successfully, get sent, and then be
+    // silently dropped as undecodable on the consumer with no error on the
+    // sending side — see the matching guard in `uint8_layout`.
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Enforcing `MAX_IPC_BYTES` on the encoded `buf` in the fallback `encode_arrow_ipc` closes the send/receive asymmetry: `decode_arrow_ipc`, `decode_arrow_ipc_zero_copy`, and the streaming `InputDecoder` all bail above the cap, so producing an oversized stream previously meant silent data loss with no producer-side error. Failing at encode time with an actionable message is the right behavior and matches `send_output_raw`'s `uint8_layout` guard. The check is on the framed stream length (what the consumer actually receives), so the two sides agree. Regression test covers the just-over-cap case.

---

PR #2595, opened 2026-07-11
fix(runtime): return a descriptive error for unsupported operator sources

phil-opp:
## Summary

When `run_operator` received an `OperatorSource::Wasm` (a valid descriptor variant) — or a Python source in a build without the `python` feature — it only logged via `tracing::error!` and returned `Ok(())`. That dropped the `init_done` oneshot sender without ever signalling it, so the runtime task blocked in `init_done.await` woke with a `RecvError` and failed with the misleading:

> the `init_done` channel was closed unexpectedly / failed to init an operator

instead of the real reason. No clean shutdown path ran.

## Fix

Return a descriptive error from these arms (`"... uses a WASM source, which is not supported yet"` / `"... built without the `python` feature"`) so the accurate message propagates through the caller's `?` at `lib.rs:107`.

## Validation

- Adds a regression test (`wasm_source_returns_descriptive_error`) asserting `run_operator` returns a descriptive error **and** that `init_done` is never signalled for an unsupported source.
- `cargo test -p dora-runtime`, `cargo clippy -p dora-runtime -- -D warnings`, `cargo fmt --all --check` all pass.

---

---

### DIFF
--- a/binaries/runtime/src/operator/mod.rs
+++ b/binaries/runtime/src/operator/mod.rs
@@ -57,12 +57,17 @@ pub fn run_operator(
                 )
             })?;
             #[cfg(not(feature = "python"))]
-            tracing::error!(
-                "Dora runtime tried spawning Python Operator outside of python environment."
+            eyre::bail!(
+                "operator `{}` uses a Python source, but this dora-runtime was \
+                 built without the `python` feature",
+                operator_definition.id
             );
         }
         OperatorSource::Wasm(_) => {
-            tracing::error!("WASM operators are not supported yet");
+            eyre::bail!(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Replacing the `tracing::error!` + implicit `Ok(())` with `eyre::bail!` for the WASM arm and the `#[cfg(not(feature = "python"))]` Python arm is correct: it now propagates a descriptive error through the caller's `?` instead of dropping the `init_done` sender and stalling the runtime with the misleading "channel was closed unexpectedly". The `#[cfg]` gating is preserved, so the Python-enabled build path is unaffected. The regression test also asserts `init_done` never signals.

---

PR #2596, opened 2026-07-11
fix(message): clamp DropOldest queue capacity to at least 1

phil-opp:
## Summary

`QueuePolicy::DropOldest::effective_cap` returned `queue_size` verbatim, so an input configured with `queue_size: 0` yielded an effective cap of **0**. In the runtime operator channel's `drop_oldest_inputs`, the newest event (iterated first) then matches the `Some(0)` arm and is set to `None` on **every** `add_event` — the operator never receives a single message for that input and the dataflow **silently hangs**, with only a generic "dropped … queue too full" warning. The node event-stream scheduler is likewise degraded (constant dropping, and a `cap == 0` would run eviction against an empty queue).

No descriptor validation rejects `queue_size: 0`, so this is reachable from user config.

## Fix

Clamp the `DropOldest` cap to `queue_size.max(1)`, so a zero size behaves as **latest-only** instead of a dead port. `Backpressure` already had a floor of 100 and is unchanged.

## Validation

- Adds unit tests for both policies' cap flooring (`drop_oldest_cap_is_never_zero`, `backpressure_cap_has_floor`).
- `cargo test -p dora-message`, `cargo clippy -p dora-message -- -D warnings`, `cargo fmt --all --check` all pass.

---

---

### DIFF
--- a/libraries/message/src/config.rs
+++ b/libraries/message/src/config.rs
@@ -38,11 +38,17 @@ pub enum QueuePolicy {
 impl QueuePolicy {
     /// Returns the effective capacity for a given configured queue size.
     ///
-    /// - `DropOldest`: returns `queue_size` as-is.
+    /// - `DropOldest`: returns `queue_size`, but never less than 1.
     /// - `Backpressure`: returns `10 * queue_size` (min 100) as a hard safety cap.
+    ///
+    /// A `DropOldest` cap of 0 would drop 100% of the input's events — the
+    /// runtime operator channel sets the just-queued event to `None` on every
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Clamping `DropOldest`'s effective cap to `queue_size.max(1)` correctly avoids a cap of 0, which would otherwise starve the input entirely. The change is confined to the `DropOldest` arm; `Backpressure` retains its floor of 100 and non-zero sizes are unchanged. Tests cover both policies' flooring.

---

PR #2598, opened 2026-07-11
fix(c-node): accept the (NULL, 0) empty-payload idiom in dora_send_output

phil-opp:
## Summary

`dora_send_output`'s `try_send_output` rejected any null `data_ptr` unconditionally, so a C node emitting a zero-length output with the standard idiom `dora_send_output(ctx, id, id_len, NULL, 0)` got `-1`. The parallel operator FFI `dora_send_operator_output` already accepts `(NULL, 0)` (it branches on `data_len == 0` first), so the node surface was inconsistent and empty/tick-style outputs from C nodes were impossible via the natural call.

## Fix

Extract a `data_slice` helper that:
- treats `data_len == 0` as an empty slice (accepting the `(NULL, 0)` idiom),
- rejects a null pointer only with a **non-zero** length,
- otherwise builds the slice.

This also avoids `slice::from_raw_parts(NULL, 0)`, which is UB in Rust even for length 0 (the pointer must be non-null and aligned regardless of length).

## Validation

- Adds 3 unit tests: `(NULL, 0)` accepted, null-with-length rejected, valid pointer read.
- `cargo test -p dora-node-api-c`, `cargo clippy -p dora-node-api-c -- -D warnings`, `cargo fmt --all --check` all pass.

---

---

### DIFF
--- a/apis/c/node/src/lib.rs
+++ b/apis/c/node/src/lib.rs
@@ -284,18 +284,42 @@ unsafe fn try_send_output(
     data_ptr: *const u8,
     data_len: usize,
 ) -> eyre::Result<()> {
-    if context.is_null() || id_ptr.is_null() || data_ptr.is_null() {
+    if context.is_null() || id_ptr.is_null() {
         eyre::bail!("null pointer passed to dora_send_output");
     }
     let context: &mut DoraContext = unsafe { &mut *context.cast() };
     let id = std::str::from_utf8(unsafe { slice::from_raw_parts(id_ptr, id_len) })?;
     let output_id = id.to_owned().into();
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

The `data_slice` helper correctly branches on `data_len == 0` before dereferencing, avoiding `slice::from_raw_parts(NULL, 0)` UB, and only rejects a null pointer when `data_len > 0` — matching the operator FFI's `(NULL, 0)` acceptance. The removal of the unconditional `data_ptr.is_null()` check from `try_send_output` is sound because the null case is now handled inside the helper. Tests cover all three cases.

---

PR #2603, opened 2026-07-11
fix(replay-node): honor each node's initial timestamp offset on replay

### LINKED ISSUE #2602: replay-node: first message emitted immediately drops each node's initial timestamp offset, losing cross-node alignment on replay
> ⚠️ **This issue was created by an automated scheduled Claude code-review check.** Verified against source; please confirm before acting on it.

## Summary

`dora-replay-node` paces output by sleeping the delta between *consecutive entries of the same node*. Because the delta accumulator starts at `None`, the **first** message of every replayed node is emitted immediately, discarding that node's initial `timestamp_offset_nanos` (the delay from recording-start to the node's first output). Per-node inter-message spacing is preserved, but inter-node alignment is not.

## Location

`binaries/replay-node/src/main.rs`, lines ~32 and ~41–50:

```rust
let mut prev_offset: Option<u64> = None;
...
// Sleep to maintain timing
if speed > 0.0
    && let Some(prev) = prev_offset
{
    let delta_nanos = entry.timestamp_offset_nanos.saturating_sub(prev);
    let sleep_nanos = (delta_nanos as f64 / speed) as u64;
    if sleep_nanos > 0 {
        thread::sleep(Duration::from_nanos(sleep_nanos));
    }
}
prev_offset = Some(entry.timestamp_offset_nanos);
```

On the first matching entry, `prev_offset` is `None`, so no sleep happens and `entry.timestamp_offset_nanos` (which may be large) is simply adopted as the new baseline.

## Concrete scenario

Record a dataflow where sensor `A` first emits at recording-offset 0 ms and sensor `B` first emits at 1000 ms (B starts late). Each is replayed by a separate `dora-replay-node` instance. On replay, both `A` and `B` emit their first message at ~t=0, so a downstream fusion node sees B's stream ~1 s earlier relative to A than it did live. This contradicts the intent that "downstream nodes receive replayed data identically to live data."

## Suggested fix

Seed the first sleep from the entry's own offset, i.e. treat the pre-loop baseline as 0 so the initial gap is honored:

```rust
let mut prev_offset: u64 = 0;
...
if speed > 0.0 {
    let delta_nanos = entry.timestamp_offset_nanos.saturating_sub(prev_offset);
    let sleep_nanos = (delta_nanos as f64 / speed) as u64;
    if sleep_nanos > 0 {
        thread::sleep(Duration::from_nanos(sleep_nanos));
    }
}
prev_offset = entry.timestamp_offset_nanos;
```

(Independent node spawn timing already limits perfect wall-clock sync, but not honoring the initial offset makes cross-node alignment strictly and unnecessarily worse.)

*Confidence: medium — the timing loss is real and reproducible; rated medium because some cross-node skew is inherent to independent replay-node spawn.*

phil-opp:
## Summary

Fixes #2602.

`dora-replay-node` paced output by sleeping the delta between *consecutive entries of the same node*, with the delta accumulator (`prev_offset`) starting at `None`. Because of that, the **first** message of every replayed node was emitted immediately, discarding the node's initial `timestamp_offset_nanos` (the delay from recording-start to its first output). Per-node inter-message spacing was preserved, but **inter-node alignment was not**: two sensors that first emitted 1 s apart when recorded would both replay their first message at ~t=0, so a downstream fusion node sees a different relative timing than it did live.

## Fix

Seed the pacing baseline at `0` (recording start) instead of the first entry's own offset, so the first entry sleeps for its own `timestamp_offset_nanos` and the initial gap is honored. The pacing delta is extracted into a small pure function `pacing_sleep_nanos(prev_offset, entry_offset, speed)` and covered by unit tests (initial offset honored, inter-message delta, speed scaling, non-monotonic saturation).

## Validation

- `cargo test -p dora-replay-node` — 4 new tests ✅
- `cargo clippy -p dora-replay-node -- -D warnings` ✅
- `cargo fmt -p dora-replay-node -- --check` ✅

---

---

### DIFF
--- a/binaries/replay-node/src/main.rs
+++ b/binaries/replay-node/src/main.rs
@@ -9,6 +9,22 @@ use dora_node_api::{
 use dora_recording::RecordingReader;
 use eyre::Context;
 
+/// Nanoseconds to sleep before emitting an entry, given the previous entry's
+/// recording offset, this entry's offset, and the replay `speed`.
+///
+/// `prev_offset` starts at 0 (recording start), so the very first entry sleeps
+/// for its own `timestamp_offset_nanos` — the delay from recording-start to the
+/// node's first output. Dropping that initial gap (by starting the baseline at
+/// the first entry's own offset) would emit every node's first message at ~t=0
+/// and destroy cross-node alignment on replay (dora-rs/dora#2602).
+fn pacing_sleep_nanos(prev_offset: u64, entry_offset: u64, speed: f64) -> u64 {
+    if speed <= 0.0 {
+        return 0;
+    }
+    let delta_nanos = entry_offset.saturating_sub(prev_offset);
+    (delta_nanos as f64 / speed) as u64
+}
+
 fn main() -> eyre::Result<()> {
     let replay_file =
         std::env::var("DORA_REPLAY_FILE").wrap_err("DORA_REPLAY_FILE env var not set")?;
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Seeding `prev_offset` at `0` so the first entry sleeps for its own `timestamp_offset_nanos` correctly restores cross-node alignment, and `pacing_sleep_nanos` uses `saturating_sub` to guard non-monotonic offsets plus a `speed <= 0.0` short-circuit. The `f64` division for pacing is acceptable here. The extracted helper and its four tests cover the relevant cases.

---

PR #2604, opened 2026-07-11
fix(coordinator): treat a fully-drained state log as pruned for a lagging daemon

### LINKED ISSUE #2601: Coordinator: state_log_delta empty-log branch returns Some([]) instead of None, silently skipping a lagging daemon's param catch-up
> ⚠️ **This issue was created by an automated scheduled Claude code-review check.** The reachability argument was corroborated against the source (`daemon_ack_sequence` is only seeded at construction and updated on ack — the relink path does not seed it), but please confirm before acting on it.

## Summary

`RunningDataflow::state_log_delta` returns `Some(Vec::new())` ("you're up to date") whenever the state log is empty, even when the querying daemon is provably behind. Its documented contract says it should return `None` (→ full param replay) when the log was pruned past `last_ack`. For the sole production caller, an empty log while the daemon is behind always means "pruned past `last_ack`", so the correct answer there is `None`. As written, a lagging daemon can permanently miss param mutations with no fallback replay and no log line.

## Location

`binaries/coordinator/src/state.rs`, lines ~391–410:

```rust
/// Return the entries a daemon has missed since `last_ack`.
/// Returns `None` if the log has been pruned past `last_ack` (caller
/// should fall back to a full param replay).
pub(crate) fn state_log_delta(&self, last_ack: u64) -> Option<Vec<StateCatchUpEntry>> {
    if self.state_log.is_empty() {
        return Some(Vec::new());        // <-- unconditional "up to date"
    }
    let oldest = self.state_log[0].sequence;
    if last_ack.saturating_add(1) < oldest {
        return None;                    // gap -> full replay
    }
    Some(self.state_log.iter().filter(|e| e.sequence > last_ack).cloned().collect())
}
```

The non-empty branch honors the "pruned → `None`" contract via the `oldest` gap check; the empty-log branch does not.

## Caller

`binaries/coordinator/src/lib.rs` ~2611–2657:

```rust
let last_ack = df.daemon_ack_sequence.get(&daemon_id).copied().unwrap_or(0);
if last_ack >= df.state_log_sequence { continue; }   // up to date -> skip
match df.state_log_delta(last_ack) {
    Some(entries) if entries.is_empty() => {}         // <-- does nothing
    Some(entries) => { /* send StateCatchUp */ }
    None => { /* full param replay fallback */ }
}
```

So `state_log_delta` is only ever called when `last_ack < state_log_sequence`. If the log is empty in that state, the entries `last_ack+1 ..= state_log_sequence` were pruned — the daemon must get the `None` fallback. Instead it gets `Some([])` → the `entries.is_empty() => {}` arm → nothing happens.

## Why an empty log implies "pruned past last_ack" here

`prune_state_log` (state.rs ~381) removes entries with `sequence <= min(daemon_ack_sequence.values())`. If the log is fully drained while `state_log_sequence > 0`, then `min_ack >= state_log_sequence`, i.e. every daemon *in the ack map* has acked past the end. For the querying daemon to still satisfy `last_ack < state_log_sequence`, it must **not** be represented in `daemon_ack_sequence` (so `unwrap_or(0)` yields 0). That is exactly the relink situation below — the ack map is seeded only at construction (`recovered` state.rs:462, `handlers.rs:640`) and updated on ack (lib.rs:2675); no relink path inserts a reconnecting daemon into it.

## Concrete reachable scenario (multi-daemon + coordinator restart)

1. Multi-daemon dataflow on d1, d2. Coordinator restarts / orphan-reclaim rebuilds `running_dataflows` from the store.
2. d1 reconnects first → `RunningDataflow::recovered` seeds `daemon_ack_sequence = {d1: 0}`, `state_log_sequence = 0`, `state_log = []`.
3. d2 reconnects → relink (`reestablish_running_dataflow`, lib.rs ~3085) does `df.daemons.insert(d2)` but **never** seeds `daemon_ack_sequence` for d2.
4. d2 disconnects again (removed from `df.daemons`; ack map still `{d1: …}`).
5. Params are set while d2 is down → `state_log_sequence` climbs to 3, entries forwarded live to d1 only; d1 acks 3 → `prune_state_log` uses `min({d1: 3}) = 3` → **log drained empty**, `state_log_sequence` stays 3.
6. d2 reconnects → relink re-adds it to `df.daemons`; catch-up loop runs `last_ack = unwrap_or(0) = 0`, `0 < 3` passes the guard, `state_log_delta(0)` sees an empty log → `Some([])` → **nothing sent**. d2 permanently misses param mutations seq 1–3, no fallback, no log.

Had d2 carried an ack entry (even 0), `prune` would keep `min = 0` and never fully drain the log, so `state_log_delta(0)` would return the correct full delta. The existing test `state_log_delta_returns_none_when_pruned` only exercises a *partially* pruned log, never the fully-drained case.

## Suggested fix (either or both)

- Make the empty-log branch prune-aware: pass `state_log_sequence` in and return `None` when `last_ack < state_log_sequence` (only return `Some([])` when the caller can prove up-to-dateness).
- And/or seed `daemon_ack_sequence` for a relinked daemon in `reestablish_running_dataflow` so `prune_state_log` cannot drain the log out from under a lagging member.

*Confidence: medium — the logic error is unambiguous for the sole caller; reachability requires the multi-daemon recovery/relink/disconnect sequence above, but every step is a real code path.*

phil-opp:
## Summary

Fixes #2601.

`RunningDataflow::state_log_delta` returned `Some(Vec::new())` ("you're up to date") whenever the state log was empty — even when the querying daemon was provably behind. Its documented contract is to return `None` (→ full param replay) when the log has been pruned past `last_ack`; only the *non-empty* branch honored that via the `oldest` gap check.

The sole production caller (`binaries/coordinator/src/lib.rs`) only calls `state_log_delta` when `last_ack < state_log_sequence`, so an empty log in that state always means the entries the daemon still needs (`last_ack+1..=state_log_sequence`) were pruned away. As written, that daemon got the `Some([])` → `entries.is_empty() => {}` arm and silently missed every param mutation — no fallback replay, no log line.

### Reachability

On multi-daemon coordinator restart, a relinked daemon is re-added to `df.daemons` but never seeded into `daemon_ack_sequence` (the ack map is only seeded at construction and updated on ack). So it queries with `last_ack == 0` (`unwrap_or(0)`), while its peers ack and `prune_state_log` drains the log entirely (`state_log_sequence` stays > 0). It then permanently misses param mutations.

## Fix

Make the empty-log branch prune-aware: return `None` when `last_ack < self.state_log_sequence`, and `Some(vec[])` only when the daemon is genuinely current. No signature change — the method already has `self.state_log_sequence`. Adds a regression test (`state_log_delta_returns_none_when_log_fully_drained_but_daemon_behind`) exercising the fully-drained case the existing `..._returns_none_when_pruned` test (partial prune) never covered.

## Validation

- `cargo test -p dora-coordinator state_log` — 5 tests incl. new regression ✅
- `cargo clippy -p dora-coordinator -- -D warnings` ✅
- `cargo fmt -p dora-coordinator -- --check` ✅

---

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -5928,6 +5928,42 @@ mod tests {
     }
 
     #[test]
+    fn state_log_delta_returns_none_when_log_fully_drained_but_daemon_behind() {
+        // Regression for #2601: a relinked daemon that was never seeded into
+        // `daemon_ack_sequence` queries with last_ack == 0. If peers have
+        // acked and `prune_state_log` drained the log entirely (while
+        // `state_log_sequence` stays > 0), the empty-log branch must return
+        // `None` (→ full param replay), not `Some([])` ("up to date").
+        let dataflow_id = DataflowId::from(Uuid::new_v4());
+        let d1 = DaemonId::new(Some("m1".to_string()));
+        let node_id: dora_core::config::NodeId = "camera".to_string().into();
+
+        let mut df = test_running_dataflow(dataflow_id, d1.clone(), node_id.clone());
+        for i in 0..3 {
+            df.append_state_log(StateCatchUpOperation::SetParam {
+                node_id: node_id.clone(),
+                key: format!("key_{i}"),
+                value: serde_json::json!(i),
+            });
+        }
+
+        // d1 (the only member in the ack map) acks all 3 → prune drains the
+        // whole log, but state_log_sequence stays at 3.
+        df.daemon_ack_sequence.insert(d1, 3);
+        df.prune_state_log();
+        assert!(df.state_log.is_empty());
+        assert_eq!(df.state_log_sequence, 3);
+
+        // A relinked daemon not in the ack map is provably behind (0 < 3) yet
+        // sees an empty log → must get the None fallback, not Some([]).
+        assert!(df.state_log_delta(0).is_none());
+
+        // A member already at the head is genuinely up to date → empty, non-None.
+        let delta = df.state_log_delta(3).expect("head daemon is up to date");
+        assert!(delta.is_empty());
+    }
+
+    #[test]
     fn set_param_forward_reply_reports_daemon_rejection() {
         let reply = serde_json::to_vec(&DaemonCoordinatorReply::SetParamResult(Err(
             "node `camera` channel full".to_string(),
--- a/binaries/coordinator/src/state.rs
+++ b/binaries/coordinator/src/state.rs
@@ -393,6 +393,17 @@ impl RunningDataflow {
     /// should fall back to a full param replay).
     pub(crate) fn state_log_delta(&self, last_ack: u64) -> Option<Vec<StateCatchUpEntry>> {
         if self.state_log.is_empty() {
+            // An empty log only means "up to date" when nothing newer than
+            // `last_ack` ever existed. If `last_ack < state_log_sequence`, the
+            // entries the daemon still needs (`last_ack+1..=state_log_sequence`)
+            // were pruned away — the daemon must fall back to a full replay,
+            // not be told it is current. This happens on relink: a daemon that
+            // was never seeded into `daemon_ack_sequence` queries with
+            // `last_ack == 0` after peers acked and drained the log
+            // (dora-rs/dora#2601).
+            if last_ack < self.state_log_sequence {
+                return None;
+            }
             return Some(Vec::new());
         }
         let oldest = self.state_log[0].sequence;


phil-opp:
The failing `ci/circleci: contract-tests` job appears to be the known flake tracked in #2064, not a real failure from this change:

- #2064 documents this job intermittently hanging/failing (`no_output_timeout`) on PRs that touch nothing in its Python / ROS2 / example-dataflow surface, and explicitly lists "daemon/coordinator + test-only" PRs among the false-negative victims.
- This PR changes only `binaries/coordinator/src/state.rs` (the `state_log_delta` empty-log branch) plus a coordinator unit test. `state_log_delta` is only ever called from the daemon-reconnect state-catch-up loop, which the `contract_*` dataflow-semantics tests never exercise — so this change cannot affect them.

All the fast hard gates (fmt, clippy, typos, audit, unwrap-budget, license, cross-checks) are green. A re-run of the contract-tests job should clear it.

---

phil-opp:

No issues found.

The empty-log branch in `state_log_delta` now returns `None` (→ full replay) when `last_ack < state_log_sequence` and `Some(vec[])` only when `last_ack == state_log_sequence` (genuinely current). That is consistent with the non-empty branch's pruned-past-`last_ack` contract, and the boundary case (`last_ack == state_log_sequence`) is covered by the test's `state_log_delta(3)` assertion. Logic looks correct.

---

PR #2605, opened 2026-07-11
fix(core): register the admitted type body for a duplicated manifest URN

### LINKED ISSUE #2599: Manifest type injection registers the wrong (possibly body-invalid) TypeDef when two manifests ship the same URN
> ⚠️ **This issue was created by an automated scheduled Claude code-review check.** The finding was verified against the source by hand, but please confirm before acting on it.

## Summary

In `libraries/core/src/manifest/inject.rs`, the loop that registers shipped types into the **shared** registry can write a *body-invalid* definition into the registry when two matched manifests ship the same type URN. This defeats the carefully-built body-resolvability fixpoint that the surrounding code (and its tests) rely on.

## Location

`libraries/core/src/manifest/inject.rs`, lines ~122–128:

```rust
// Register the survivors into the shared registry, skipping URNs already
// present so preserved definitions win.
for &(urn, def) in &candidates {
    if scratch.resolve(urn).is_some() && registry.resolve(urn).is_none() {
        let _ = registry.add_user_type(urn, def.clone());
    }
}
```

## The problem

`candidates` (built at lines ~92–99) is a flat `Vec<(&urn, &def)>` accumulated across **all** matched manifests and is **not** deduplicated by URN. The fixpoint above (lines ~100–121) decides admission *per URN* into `scratch`, inserting only a **body-valid** definition for each URN.

The registration loop then guards each candidate with `scratch.resolve(urn).is_some()`. But `scratch.resolve(urn)` being `Some` only proves that *some* definition for that URN was admitted — **not** that *this* candidate's `def` is the admitted one. So the loop registers `def` from the **first-listed candidate** for that URN via `add_user_type(urn, def.clone())`, which may be a body-**invalid** definition. `add_user_type`/`insert_type` just store whatever `def` is passed (see `libraries/core/src/types.rs:621` / `:503`).

## Concrete failure scenario

Two `path:` nodes under the same namespace ship the same URN, invalid one first:

- **Node A** (`namespace: acme`, listed first) ships `acme/foo/v1/X` = `arrow: Struct` with field `bad: NotARealType` → **body-invalid** (unresolvable field type).
- **Node B** (`namespace: acme`, listed second) ships `acme/foo/v1/X` = `arrow: Struct` with field `x: Float32` → **body-valid**.
- **Node C** has an output annotated `type: acme/foo/v1/X`.

Trace:
1. `candidates = [(X, A_invalid), (X, B_valid)]`.
2. Fixpoint: `A_invalid` fails `body_valid`; `B_valid` passes → `scratch` gets `X = B_valid`. `scratch.resolve("acme/foo/v1/X")` is now `Some`.
3. Registration loop, first iteration `(X, A_invalid)`: `scratch.resolve(X).is_some()` (true, from B) `&& registry.resolve(X).is_none()` (true) → `registry.add_user_type("acme/foo/v1/X", A_invalid.clone())`. **The shared registry now holds the body-invalid definition.** Second iteration `(X, B_valid)`: `registry.resolve(X)` is now `Some` → skipped.
4. Pass 2: Node C's port check `check_port_type("acme/foo/v1/X")` resolves against the registry → passes, contracts injected. Later Arrow materialization of `X` fails on `NotARealType`.

This directly violates the invariant the module documents (inject.rs lines ~76–90) and the tests `unresolvable_sibling_body_cannot_contaminate_the_registry` / `type_referencing_a_body_invalid_sibling_is_not_registered` are meant to protect: a body-invalid sibling must never make a referencing node's port check pass against a type that can't be materialized into an Arrow schema. The existing tests don't catch it because none exercise two manifests shipping the **same** URN with differing bodies.

## Reachability

Requires two matched manifests under the same namespace shipping the same type URN, with the body-invalid one ordered first (e.g. two nodes from the same org where one carries a broken/WIP type definition). Cross-namespace duplicates are already rejected, so both must be same-namespace. Narrow, but a realistic local-development configuration.

## Suggested fix

Register the **admitted** definition from `scratch` rather than the candidate's own `def`, e.g.:

```rust
for &(urn, _def) in &candidates {
    if registry.resolve(urn).is_none()
        && let Some(admitted) = scratch.resolve(urn)
    {
        let _ = registry.add_user_type(urn, admitted.clone());
    }
}
```

(or deduplicate `candidates` by URN, selecting the body-valid entry). A regression test with two same-namespace manifests shipping the same URN — body-invalid first — would lock this down.

*Confidence: high (logic verified directly against source).*

phil-opp:
## Summary

Fixes #2599.

In `libraries/core/src/manifest/inject.rs`, the loop that registers shipped types into the **shared** registry could write a *body-invalid* definition when two matched manifests ship the same type URN.

`candidates` is a flat `Vec<(&urn, &def)>` accumulated across all matched manifests and is **not** deduplicated by URN. The admission fixpoint decides admission *per URN* into `scratch`, inserting only a body-valid definition. But the registration loop then wrote each candidate's own `def`, guarded only by `scratch.resolve(urn).is_some()` — which proves *some* definition for that URN was admitted, **not** that *this* candidate's `def` is the admitted one. So when a body-invalid duplicate is listed first, the registry ended up holding the body-invalid definition, and a referencing node's port check passed against a type that can never be materialized into an Arrow schema — defeating the body-resolvability invariant the module documents and its tests protect.

### Failure scenario

Two same-namespace `path:` nodes ship `acme/foo/v1/X`, the invalid one first:
- Node A (listed first) ships `X` with field `bad: NotARealType` → body-invalid.
- Node B (listed second) ships `X` with field `x: Float32` → body-valid.

The fixpoint admits B into `scratch`, but the registration loop's first iteration `(X, A_invalid)` sees `scratch.resolve(X).is_some()` (from B) and registers **A_invalid** into the shared registry.

## Fix

Register the *admitted* definition from `scratch` (the body-valid one the fixpoint chose) rather than the candidate's own `def`:

```rust
for &(urn, _def) in &candidates {
    if registry.resolve(urn).is_none()
        && let Some(admitted) = scratch.resolve(urn)
    {
        let _ = registry.add_user_type(urn, admitted.clone());
    }
}
```

Adds a regression test (`duplicate_urn_registers_the_body_valid_definition_not_the_first_listed`) with two same-namespace manifests shipping the same URN, body-invalid first. Verified it **fails** against the old code and passes with the fix.

## Validation

- `cargo test -p dora-core --lib manifest::inject` — 18 tests incl. new regression ✅ (confirmed red without the fix)
- `cargo clippy -p dora-core -- -D warnings` ✅
- `cargo fmt -p dora-core -- --check` ✅

---

---

### DIFF
--- a/libraries/core/src/manifest/inject.rs
+++ b/libraries/core/src/manifest/inject.rs
@@ -120,10 +120,19 @@ pub fn inject_adjacent_manifests(
         }
     }
     // Register the survivors into the shared registry, skipping URNs already
-    // present so preserved definitions win.
-    for &(urn, def) in &candidates {
-        if scratch.resolve(urn).is_some() && registry.resolve(urn).is_none() {
-            let _ = registry.add_user_type(urn, def.clone());
+    // present so preserved definitions win. `candidates` is a flat list across
+    // all manifests and is NOT deduplicated by URN, so two matched manifests
+    // may ship the same URN with differing bodies. Register the *admitted*
+    // definition from `scratch` — the body-valid one the fixpoint chose — not
+    // this candidate's own `def`, which may be a body-invalid duplicate listed
+    // first. Registering `def` here would write an unmaterializable definition
+    // into the shared registry and let a referencing node's port check pass
+    // against a type that can never become an Arrow schema (dora-rs/dora#2599).
+    for &(urn, _def) in &candidates {
+        if registry.resolve(urn).is_none()
+            && let Some(admitted) = scratch.resolve(urn)
+        {
+            let _ = registry.add_user_type(urn, admitted.clone());
         }
     }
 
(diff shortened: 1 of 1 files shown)


phil-opp:
The failing `ci/circleci: contract-tests` job appears to be the known flake tracked in #2064, not a real failure from this change:

- #2064 documents this job intermittently hanging/failing (`no_output_timeout`) on PRs that touch nothing in its Python / ROS2 / example-dataflow surface — it even lists a one-file `libraries/core` change (#2048) among the false-negative victims.
- This PR changes only `libraries/core/src/manifest/inject.rs`, a compose-time (`dora build` / `dora validate`) type-injection path with no Python or dataflow-runtime interaction, plus a unit test in the same module. It cannot affect the `contract_*` dataflow-semantics tests.

All the fast hard gates (fmt, clippy, typos, audit, unwrap-budget, license, cross-checks) are green, and `cargo test -p dora-core --lib manifest::inject` passes locally (18 tests, including the new regression). A re-run of the contract-tests job should clear it.

---

phil-opp:

No issues found.

I traced the change in `inject_adjacent_manifests` (`libraries/core/src/manifest/inject.rs`): registering the `scratch`-admitted (body-valid) definition instead of the first-listed candidate's own `def` is correct, and the `registry.resolve(urn).is_none()` guard still makes the loop idempotent for repeated URNs. The regression test exercises the body-invalid-first case. Behavior looks sound.

---

PR #2606, opened 2026-07-11
fix(daemon): re-check disable_restart before respawn on every path

phil-opp:
## Summary

`restart_loop` in `binaries/daemon/src/spawn/prepared.rs` can respawn a node **after** an explicit stop, leaving an orphan process that survives `dora stop`.

## The issue

Each loop iteration reads `disable_restart` once, near the top (`prepared.rs:247`), and then:
1. emits the `SpawnedNodeResult` event via `daemon_tx.send(event).await` (an `.await` point), and
2. optionally sleeps for the restart backoff (`tokio::time::sleep(backoff).await`, another `.await` point),

before actually calling `spawn_inner` to respawn. During those await points a concurrent `StopDataflow` runs `RunningDataflow::stop_all`, which sets `disable_restart = true` on **every** node.

The guard that re-checks `disable_restart` after the wait lived **inside** the `if let Some(base_delay) = config.restart_delay { … }` block. But `restart_delay` defaults to `None` (`restart_config`, `prepared.rs:194`), so a node using the default configuration had **no** final gate. A stop that races an in-progress restart was therefore ignored, and the node was respawned after the stop.

Nothing reaps the respawned process: `stop_all` already ran and had nothing to kill for this node (its process slot was empty because it had just exited), and `finish_stragglers` is gated on `stop_sent`. The result is an orphan process that outlives `dora stop`.

### Failing scenario
1. A node with `restart_policy: Always` and no `restart_delay` exits.
2. `restart_loop` decides `restart = true` and awaits on the event send / logging.
3. Concurrently the coordinator delivers `StopDataflow`; `stop_all` sets `disable_restart = true` for the node (whose slot is already empty).
4. `restart_loop` resumes. With `restart_delay == None` the re-check is skipped, so it respawns the node — after the stop.

## The fix

Move the `disable_restart` re-check out of the `restart_delay.is_some()` branch so it runs **unconditionally**, immediately before committing to the respawn — on both the delayed and the immediate (default) restart paths. This is a one-line control-flow change plus the relocated guard; the `break` was already the behavior used by the old delay-branch check, so no new exit path is introduced.

## Validation

- `cargo fmt -p dora-daemon -- --check` ✓
- `cargo clippy -p dora-daemon -- -D warnings` ✓ (matches CI invocation)
- `cargo test -p dora-daemon` ✓ (96 passed)

`restart_loop` takes a fully-constructed `PreparedNode` plus live spawn machinery, so there is no isolated seam to unit-test the race directly; the fix is a minimal, self-contained control-flow correction.

---

🤖 This is a **machine-generated** pull request proposed by Claude (an AI agent) as part of an automated code-review pass. It has been verified locally as described above, but a human should review it carefully before merging.

https://claude.ai/code/session_01WYwkkfs2mPg1Gj4s26HB1q

---
_Generated by [Claude Code](https://claude.ai/code/session_01WYwkkfs2mPg1Gj4s26HB1q)_

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -329,18 +329,28 @@ impl PreparedNode {
                         )
                         .await;
                     tokio::time::sleep(backoff).await;
+                }
 
-                    // Re-check disable_restart after sleep (may have changed)
-                    if disable_restart.load(atomic::Ordering::Acquire) {
-                        logger
-                            .log(
-                                LogLevel::Info,
-                                Some("daemon".into()),
-                                "restart cancelled: inputs closed during backoff wait".to_string(),
-                            )
-                            .await;
-                        break;
-                    }
+                // Re-check `disable_restart` before committing to a respawn. It
+                // may have flipped to `true` after the `load` at the top of this
+                // iteration: emitting the `SpawnedNodeResult` event and any
+                // backoff sleep above are `.await` points, and a concurrent
+                // `StopDataflow` (`RunningDataflow::stop_all`) sets
+                // `disable_restart` on every node. Keeping this gate inside the
+                // `restart_delay.is_some()` branch above meant that a node with
+                // the default (unset) `restart_delay` had no final check, so a
+                // stop racing an in-progress restart was ignored and the node
+                // respawned *after* the stop — an orphan process that `stop_all`
+                // has already passed over.
+                if disable_restart.load(atomic::Ordering::Acquire) {
+                    logger
+                        .log(
+                            LogLevel::Info,
+                            Some("daemon".into()),
+                            "restart cancelled: inputs closed before respawn".to_string(),
+                        )
+                        .await;
+                    break;
                 }
 
                 restart_count += 1;


phil-opp:

🤖 _Machine-generated by Claude (an AI agent). This comment summarizes one automated review pass over the repo. Each PR was verified against a freshly-fetched `origin/main`; the combined diff passed the `/review` and `/simplify` gates._

### PRs opened this run

**Correctness (2)**
- #2606 — `fix(daemon): re-check disable_restart before respawn on every path` — a `StopDataflow` racing an in-progress restart is ignored when `restart_delay` is unset (the default), respawning a node **after** stop and leaving an orphan process. The `disable_restart` re-check was nested inside the `restart_delay.is_some()` branch; hoisted to run unconditionally before respawn.
- #2607 — `fix(cli): correct byte-offset tracking in dora logs --follow` — the follow loop advanced its per-file offset to the pre-read `metadata()` size instead of the bytes actually consumed, causing **duplicated output** under active logging, **aborting the whole session** on a read that ends mid-multibyte-UTF-8 (`read_to_string` → `InvalidData`), and **silently dropping** lines split across flushes. Extracted into a tested helper that consumes only through the last newline, decodes lossily, and advances by bytes consumed.

_Optimization / Documentation: no findings met the bar this run (see rejected below)._

### Findings investigated but rejected (with reason)

| Area | Finding | Why rejected |
|------|---------|--------------|
| `libraries/message` (`common.rs`) | Signal-name mapping for signals 22/23 is wrong (`SIGABRT`/`NSIG` instead of `SIGTTOU`/`SIGURG`) | **Duplicate** — already covered by open PR #2592. |
| `binaries/mavlink2-bridge-node` (`main.rs`) | Telemetry forwarding stalls up to `POLL_INTERVAL` (100 ms) because `rx` is only drained at the top of the loop, which then blocks on the dora event stream | Real latency issue, but the clean fix is either a debatable magic-number tweak or a risky inversion of carefully-commented shutdown logic (dora-rs/dora#2027, #2034). Below the risk/confidence bar for an autonomous PR. |
| `apis/rust/node` (`daemon_connection/json_to_arrow.rs`) | `batch.column(0)` can panic on a zero-column batch | Only reachable from `DORA_TEST_WITH_INPUTS` integration-test tooling with a degenerate inferred schema; not on the production receive path. Too narrow to warrant a PR. |
| `binaries/cli` (`logs.rs`) | `--since`/`--until` time filters not applied to lines streamed in `--follow` mode (only `--until` observable) | Arguable UX (following live logs past `--until` may be intended); left out to keep #2607 a tight, unambiguous correctness fix. |

### Notes
- Full-workspace `cargo test` could not run to completion in this ephemeral environment (fixed per-session disk allowance is exhausted by the ~30 GB of build artifacts a full test build requires). Each PR was instead verified with **fmt + clippy + the full test suite of its affected crate**; both changes are confined to a single binary crate and touch no shared library, so no other crate's behavior can change.
- Areas already covered by the ~40 open automated PRs (ros2-bridge, type-inference removal, IPC size guards, coordinator lifecycle, etc.) were deliberately excluded from this pass.

---

phil-opp:

No issues found in this diff.

I reviewed `binaries/daemon/src/spawn/prepared.rs` at `4da6096`. The change hoists the `disable_restart` re-check out of the `if let Some(base_delay) = config.restart_delay` block so it runs unconditionally right before `spawn_inner`. The reasoning checks out: the load at the top of the loop is followed by two `.await` points (the `SpawnedNodeResult` send and the optional backoff sleep) during which `stop_all` can set `disable_restart = true`, and the default `restart_delay: None` path previously skipped the re-check entirely. The hoisted gate is still nested under the outer `if restart`, so it only fires on the respawn path, and when `disable_restart` is false the flow is unchanged. No regression or new race introduced.

---

phil-opp:
🤖 **Automated review by Claude** — this review was generated by an AI agent with no human in the loop. Advisory only; please verify before acting.

Hoisting the `disable_restart` re-check out of the `if let Some(base_delay) = config.restart_delay` block is correct and a clear improvement: it closes the large window on the default `restart_delay: None` path (the `SpawnedNodeResult` send plus the potentially multi-second backoff `sleep`), which is where the orphan was most likely to occur.

One caveat before treating the orphan bug as fully closed on *every* path: the hoisted gate is still a **pre-spawn** check. Between it and the point where the respawned process becomes reachable for stop/kill (`ProcessHandleReplaced`, `spawn/prepared.rs:414`) there are still `.await` points:

- the `"restarting node after ..."` log (`prepared.rs:356`–`374`), and
- the whole `spawn_inner(...).await` (`prepared.rs:393`) that actually creates the child process.

If a `StopDataflow` sets `disable_restart` during that residual window — most importantly *while `spawn_inner` is creating the OS process* — the same orphan results:

- `stop_all` (`running_dataflow.rs:380`–`396`) sets `disable_restart` and `take()`s each node's process handle into the grace-kill snapshot. The mid-restart node's *old* process is already gone, and its *new* incarnation isn't in `running_nodes` yet, so the snapshot kills nothing for it.
- `stop_all` also drains `subscribe_channels` and sends `NodeEvent::Stop` (`running_dataflow.rs:384`) before the new process subscribes, so the new incarnation never receives the stop either.
- `finish_stragglers` is gated on `stop_sent`/idle and won't reap a freshly-spawned, active process.

So the fix shrinks the race dramatically (seconds → spawn latency) but doesn't eliminate it. Fully closing it needs a **post-spawn** guard in addition to the pre-spawn one: after `spawn_inner` succeeds and the handle is installed, re-check `disable_restart` and stop/kill the just-spawned incarnation if it's now set. That can be a follow-up — flagging it so the residual window is a conscious decision rather than an assumed-closed gap.

---

PR #2607, opened 2026-07-11
fix(cli): correct byte-offset tracking in `dora logs --follow`

phil-opp:
## Summary

`dora logs --follow` (the local `follow_local_logs` loop in `binaries/cli/src/command/logs.rs`) tracks how far it has read into each log file with a byte offset, but advances that offset incorrectly. This causes duplicated output, can abort the whole follow session on non-ASCII logs, and can silently drop lines.

## The issue

Each poll:
1. samples the file size with `metadata()` — call it `current_size` (time T1),
2. reads from the saved offset to EOF with `read_to_string` (time T2 ≥ T1),
3. stores the offset as **`current_size`** (the T1 size), not the number of bytes actually read.

Three defects follow:

1. **Duplicated output.** Under active logging the file grows between T1 and T2, so `read_to_string` consumes past `current_size` and prints those bytes — but the offset only advances to `current_size`. On the next poll the reader seeks back and re-reads/re-prints every byte written in the (T1, T2) window.
2. **Follow session aborts on non-ASCII logs.** `read_to_string` returns `InvalidData` if the byte range ends inside a multibyte UTF-8 sequence (any `é`, CJK char, emoji written across two flushes). The `?` propagates out of `follow_local_logs`, terminating the whole `dora logs --follow` command with *"stream did not contain valid UTF-8"*.
3. **Silent log loss.** A line split across two flushes is parsed as a broken JSON line, dropped, and the offset still advances past it, so the completed line is never re-read.

## The fix

Extract the read into a small, testable helper `read_appended_log_lines(path, pos) -> (Vec<LogMessage>, u64)` that:

- reads raw bytes (`read_to_end`) instead of `read_to_string`,
- consumes only **through the last newline**, leaving any trailing partial line for the next poll,
- decodes lossily (`from_utf8_lossy`) so a read ending mid-multibyte-char can never abort the session,
- advances the offset by exactly the **bytes consumed** (`pos + consumed`), never by the racily-larger file size.

The caller's cheap `current_size <= pos` early-out is preserved. Behavior for the common ASCII, newline-terminated case is unchanged aside from removing the duplication.

## Validation

- `cargo clippy -p dora-cli -- -D warnings` ✓ (matches CI invocation)
- `cargo test -p dora-cli` ✓ (251 passed, incl. 3 new regression tests)

New unit tests cover: complete-lines read + no re-read on the next poll (duplication guard), a trailing partial line left for the next poll, and a read ending on a split multibyte UTF-8 byte not aborting.

---

🤖 This is a **machine-generated** pull request proposed by Claude (an AI agent) as part of an automated code-review pass. It has been verified locally as described above, but a human should review it carefully before merging.

https://claude.ai/code/session_01WYwkkfs2mPg1Gj4s26HB1q

---
_Generated by [Claude Code](https://claude.ai/code/session_01WYwkkfs2mPg1Gj4s26HB1q)_

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -313,16 +313,9 @@ fn follow_local_logs(args: &LogsArgs) -> Result<()> {
             if current_size <= pos {
                 continue;
             }
-            let mut file = std::fs::File::open(path)?;
-            file.seek(std::io::SeekFrom::Start(pos))?;
-            let mut buf = String::new();
-            file.read_to_string(&mut buf)?;
-            for line in buf.lines() {
-                if let Some(msg) = parse_jsonl_line(line) {
-                    new_messages.push(msg);
-                }
-            }
-            file_positions.insert(path.clone(), current_size);
+            let (msgs, new_pos) = read_appended_log_lines(path, pos)?;
+            new_messages.extend(msgs);
+            file_positions.insert(path.clone(), new_pos);
         }
 
         new_messages.sort_by_key(|a| a.timestamp);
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found in this diff.

I reviewed `read_appended_log_lines` (`binaries/cli/src/command/logs.rs`) and its caller in `follow_local_logs`. The offset logic is correct: it consumes only through the last `\n` (`idx + 1`), returns `(empty, pos)` when no newline is present so a mid-write partial line is re-read rather than dropped, and advances the saved offset by exactly the bytes consumed (never past the racily-larger `metadata()` size) — this eliminates both the duplication and the silent-line-loss bugs. `from_utf8_lossy` on the consumed prefix removes the `read_to_string` UTF-8 abort; any truncated multibyte sequence can only land in the unconsumed tail after the last newline, so lossy decoding of the prefix is safe. Truncation/empty-read cases leave the offset unchanged. The three added tests exercise the complete-lines, trailing-partial-line, and split-multibyte-UTF-8 paths accurately.

---

Issue #2600, opened 2026-07-11
download_file: URL-derived fallback filename leaks query string / fragment (breaks .py extension detection and Windows writes)


phil-opp:
> ⚠️ **This issue was created by an automated scheduled Claude code-review check.** The finding was verified with a small standalone reproducer, but please confirm before acting on it.

## Summary

When a download response has no `Content-Disposition` header, `dora-download` derives the target filename from the URL using `Path::new(url).file_name()`. Because `Path` doesn't understand URL syntax, the query string and fragment (and sometimes the host) are captured into the filename verbatim. This produces incorrect on-disk names, corrupts extension-based dispatch, and fails outright on Windows.

## Location

`libraries/extensions/download/src/lib.rs`, `get_filename`, lines ~35–40:

```rust
// If Content-Disposition header is not available, extract from URL
let raw_name = raw_name.or_else(|| {
    Path::new(response.url().as_str())
        .file_name()
        .and_then(|n| n.to_str())
        .map(|s| s.to_string())
});
```

## Reproducer

```rust
use std::path::Path;
for u in [
    "https://example.com/path/model.bin?version=2&token=abc",
    "https://example.com/model.so#frag",
    "https://example.com/download?file=model.bin",
    "https://example.com/",
] {
    println!("{u:?} -> {:?}", Path::new(u).file_name().and_then(|n| n.to_str()));
}
```

Output:

```
"https://example.com/path/model.bin?version=2&token=abc" -> Some("model.bin?version=2&token=abc")
"https://example.com/model.so#frag"                       -> Some("model.so#frag")
"https://example.com/download?file=model.bin"             -> Some("download?file=model.bin")
"https://example.com/"                                    -> Some("example.com")   // host, not a filename
```

The `file_name()`-based sanitizer that follows (lines ~44–53) does not strip `?`/`#`, so these names reach `target_dir.join(filename)` and `File::create`.

## Impact

1. **Broken extension detection.** `binaries/daemon/src/spawn/command.rs` downloads URL nodes via `download_file(source, target_dir, None)` and then dispatches on `resolved_path.extension()`. A Python node served from e.g. `https://host/node.py?token=x` is saved as `node.py?token=x`; `Path::extension()` yields `"py?token=x"`, so the `Some("py")` arm is missed and dora tries to `exec` the script as a native binary → failure.
[...]

PR #2610, opened 2026-07-12
fix(download): match Content-Disposition filename case-insensitively

### LINKED ISSUE #2609: download_file: Content-Disposition filename parsing is case-sensitive, silently dropping RFC-legal `Filename=` headers
> ⚠️ **This issue was created by an automated scheduled Claude code-review check.** The finding was verified with a small standalone reproducer, but please confirm before acting on it.

## Summary

`parse_content_disposition_filename` in `dora-download` matches the header parameter name with a **case-sensitive** `split("filename=")`. Per [RFC 6266](https://www.rfc-editor.org/rfc/rfc6266) / RFC 2616, `Content-Disposition` parameter names are **case-insensitive**, so a spec-compliant server that sends `filename` in any other case (`Filename`, `FILENAME`, `FileName`) has its filename **silently ignored**. The download then falls through to the URL-derived fallback path — which has its own bug (#2600) — so the two together can produce a wrong on-disk name or an outright "Could not find a filename" error even though the server told us the correct name.

This is a **different code path** from #2600: that issue is about the URL fallback in `get_filename`; this is about the header parse itself in `parse_content_disposition_filename`.

## Location

`libraries/extensions/download/src/lib.rs`, `parse_content_disposition_filename`, lines 14–25:

```rust
fn parse_content_disposition_filename(header: &str) -> Option<String> {
    let rest = header.split("filename=").nth(1)?.trim_start();   // <- case-sensitive
    let name = if let Some(after_quote) = rest.strip_prefix('"') {
        after_quote.split('"').next().unwrap_or(after_quote)     // <- no `\"` unescaping
    } else {
        rest.split(';').next().unwrap_or(rest).trim()
    };
    (!name.is_empty()).then(|| name.to_string())
}
```

## Reproducer (verified)

A standalone copy of the function, compiled and run:

```rust
fn parse_content_disposition_filename(header: &str) -> Option<String> {
    let rest = header.split("filename=").nth(1)?.trim_start();
    let name = if let Some(after_quote) = rest.strip_prefix('"') {
        after_quote.split('"').next().unwrap_or(after_quote)
    } else {
        rest.split(';').next().unwrap_or(rest).trim()
    };
    (!name.is_empty()).then(|| name.to_string())
}
```

Output:

```
"attachment; filename=\"model.bin\""     -> Some("model.bin")   // lowercase works
"attachment; Filename=\"model.bin\""     -> None                // RFC-legal, DROPPED
"attachment; FILENAME=\"model.bin\""     -> None                // DROPPED
"inline; FileName=node.py"               -> None                // DROPPED
"attachment; filename=\"a\\\"b.bin\""    -> Some("a\\")         // escaped-quote truncation
```

## Impact

1. **Correct filename silently discarded.** Any server sending a non-lowercase `filename` parameter (common from some Java/.NET/IIS stacks) is treated as if it sent no `Content-Disposition` filename at all.
2. **Compounds with #2600.** With the header ignored, `get_filename` falls to the URL-derived path. For endpoints like `https://host/download?token=…` that rely on `Content-Disposition` for the real name, the result is a wrong name (query string / host captured) or a hard `"Could not find a filename"` failure. Because URL nodes are downloaded and then dispatched by extension in the daemon spawn path, a wrong name can also break `.py`/native detection.
3. **Secondary: no `\"` unescaping.** RFC 6266 quoted-strings allow a backslash-escaped quote (`filename="a\"b.bin"`); the current `split('"')` stops at the escaped quote and yields `a\`. Rare in practice, but worth handling if the quoted branch is revisited.

## Suggested fix

Match the parameter name case-insensitively while still skipping the RFC 5987 extended `filename*=` form (which is intentionally not decoded here):

```rust
fn parse_content_disposition_filename(header: &str) -> Option<String> {
    // RFC 6266: parameter names are case-insensitive.
    let lower = header.to_ascii_lowercase();
    let mut from = 0;
    let start = loop {
        let idx = from + lower[from..].find("filename=")?;
        // skip the RFC 5987 extended `filename*=` form (byte before `=` is `*`)
        if header.as_bytes().get(idx + "filename".len() - 1) == Some(&b'*') {
            from = idx + "filename=".len();
            continue;
        }
        break idx + "filename=".len();
    };
    let rest = header[start..].trim_start();
    // ... existing quoted / token handling unchanged ...
}
```

(An `\"` unescape in the quoted branch would be a small independent follow-up.)

A regression test in the same `#[cfg(test)] mod tests` block — e.g. asserting `Filename="model.bin"` and `FILENAME=model.bin` both parse to `model.bin` — would lock this in.

*Confidence: high — the case-sensitivity behavior is reproduced directly from the function's own logic. Severity is moderate (correctness on a non-hot path), and it interacts with the already-filed #2600.*

phil-opp:
## Summary

Fixes #2609.

`parse_content_disposition_filename` in `dora-download` matched the header parameter name with a **case-sensitive** `split("filename=")`. Per [RFC 6266](https://www.rfc-editor.org/rfc/rfc6266) / RFC 2616, `Content-Disposition` parameter names are **case-insensitive**, so a spec-compliant server sending `filename` in any other case (`Filename`, `FILENAME`, `FileName` — common from some Java/.NET/IIS stacks) had its filename **silently ignored**. The download then fell through to the URL-derived fallback path (which has its own bug, #2600), so the two together could produce a wrong on-disk name or an outright `Could not find a filename` error even though the server told us the correct name. Because URL nodes are dispatched by extension in the daemon spawn path, a wrong name can also break `.py`/native detection.

## Fix

Match the parameter name case-insensitively via an ASCII-lowercased copy of the header. `to_ascii_lowercase` maps only ASCII bytes 1:1 and preserves byte length, so the found offset stays a valid index into the original `header` (the value bytes are returned unchanged, preserving case). The RFC 5987 extended `filename*=` form is still skipped, because searching for the literal `filename=` (with the `=`) does not match `filename*=`.

```rust
let lower = header.to_ascii_lowercase();
let start = lower.find("filename=")? + "filename=".len();
let rest = header[start..].trim_start();
```

The escaped-quote (`\"`) unescaping noted as a secondary nit in the issue is intentionally left as an independent follow-up to keep this change focused on the reported case-sensitivity bug.

## Tests

- `filename_parameter_name_is_case_insensitive` — `Filename=`, `FILENAME=`, and `FileName=` all parse to `model.bin`.
- `extended_form_before_plain_filename_is_skipped` — a header carrying both `filename*=…` and a plain `filename="plain.bin"` still resolves the plain value.
- All pre-existing tests (lowercase, trailing params, unquoted, `;`-in-quotes, empty/missing, `filename*=`-only) continue to pass.

## Validation

- `cargo test -p dora-download` — 7 passed ✅
- `cargo clippy -p dora-download -- -D warnings` ✅

---

⚠️ **Machine-generated PR.** Authored by Claude (an AI agent) while working through automatically-filed code-review issues. Verified as described, but please review before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_019gSmVnGaEqXnzMScnFXdxV)_

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -9,10 +9,17 @@ use tokio::io::AsyncWriteExt;
 ///
 /// Handles headers that carry additional parameters after the filename, e.g.
 /// `attachment; filename="model.bin"; size=1000`, and both quoted and unquoted
-/// forms. Returns `None` when no non-empty `filename=` value is present (the
-/// RFC 5987 extended `filename*=` form is not decoded and falls through).
+/// forms. Per RFC 6266 / RFC 2616 the parameter name is case-insensitive, so
+/// `Filename`, `FILENAME`, etc. are matched too. Returns `None` when no
+/// non-empty `filename=` value is present (the RFC 5987 extended `filename*=`
+/// form is not decoded and falls through — searching for the literal
+/// `filename=` skips it, since `filename*=` does not contain that substring).
 fn parse_content_disposition_filename(header: &str) -> Option<String> {
-    let rest = header.split("filename=").nth(1)?.trim_start();
+    // RFC 6266: parameter names are case-insensitive. Lower-casing only maps
+    // ASCII bytes 1:1, so byte offsets stay valid indices into `header`.
+    let lower = header.to_ascii_lowercase();
+    let start = lower.find("filename=")? + "filename=".len();
+    let rest = header[start..].trim_start();
     let name = if let Some(after_quote) = rest.strip_prefix('"') {
         // Quoted form: the value runs up to the closing quote, so a trailing
         // `; param=...` (or a `;` inside the quotes) is handled correctly.
@@ -156,6 +163,37 @@ mod tests {
         assert_eq!(
             parse_content_disposition_filename("attachment; filename=\"a;b.bin\""),
             Some("a;b.bin".to_string())
+        );
+    }
+
+    #[test]
+    fn filename_parameter_name_is_case_insensitive() {
+        // RFC 6266: `Content-Disposition` parameter names are case-insensitive,
+        // so a spec-compliant server sending a non-lowercase `filename` must not
+        // have its filename silently dropped.
+        for header in [
+            "attachment; Filename=\"model.bin\"",
+            "attachment; FILENAME=\"model.bin\"",
+            "attachment; FileName=model.bin",
+        ] {
+            assert_eq!(
+                parse_content_disposition_filename(header),
+                Some("model.bin".to_string()),
+                "header {header:?} should parse case-insensitively"
+            );
+        }
+    }
+
+    #[test]
+    fn extended_form_before_plain_filename_is_skipped() {
+        // A header may carry both the RFC 5987 extended form and a plain one;
+        // the plain `filename=` value must still be found (and the `filename*=`
+        // form skipped, not mistaken for it).
+        assert_eq!(
+            parse_content_disposition_filename(
+                "attachment; filename*=UTF-8''extended.bin; filename=\"plain.bin\""
+            ),
+            Some("plain.bin".to_string())
         );
     }
 


phil-opp:
The failing `ci/circleci: contract-tests` check is the known-flaky job tracked in #2064 (it intermittently hangs to a ~50 min `no_output_timeout` on PRs that touch nothing in its Python / ROS2 / example-dataflow surface), not a regression from this PR:

- This diff changes only `libraries/extensions/download/src/lib.rs` (`Content-Disposition` filename parsing) — no Python, ROS2, dataflow, or example code that the contract tests exercise.
- Every hard gate is green on this commit: **Format, Clippy, Check, Audit, Typos, License, Unwrap budget** (plus CircleCI `fmt`/`clippy`/`audit`). Clippy compiles the whole workspace, so this is not a build break.

A re-run of that CircleCI job should clear it. Flagging per #2064 so the red check isn't mistaken for a real failure.

---

phil-opp:

No issues found in this diff.

I reviewed commit `d71e37d`. The switch to case-insensitive matching via `to_ascii_lowercase()` + `find("filename=")` is sound:

- ASCII lowercasing is length-preserving, so the byte offset from the lowered copy stays a valid index into the original `header` — no panic or misalignment on non-ASCII bytes, matching the code comment.
- The filename value is extracted from the original (case-preserving) `header[start..]`, so the returned value's case is unchanged — only the parameter *name* match is case-insensitive, which is the intended fix.
- `find("filename=")` does not match `filename*=` (the byte after `filename` is `*`, not `=`), so the RFC 5987 extended form is still skipped as documented; `extended_form_before_plain_filename_is_skipped` genuinely covers this.
- The added tests assert the correct expected values for the mixed/upper-case quoted and unquoted forms.

No new correctness, overflow, or resource concerns.

---

phil-opp:
The `ci/circleci: test-linux` failure on the merge-queue commit is a Linux-specific flake in the shared workspace test suite, not a regression from this change — the same `test-linux` flake hit the unrelated PR #2611 (a Rust IPC-decoder change) on its own merge-queue commit at the same time, where `test-macos` passed the identical `cargo test --all` suite and every other gate was green.

This PR changes only `libraries/extensions/download/src/lib.rs` (a pure string-parsing function + unit tests) — nothing in it can affect the coordinator/daemon integration tests that flake under `test-linux`. A re-queue / re-run of `test-linux` should clear it.

---

PR #2612, opened 2026-07-12
fix(core): validate health_check_interval and input_timeout as seconds

phil-opp:
## Issue

The daemon converts several second-valued `f64` descriptor fields to `Duration` via `Duration::from_secs_f64`, which **panics** on a negative, NaN, or infinite input.

`check_dataflow` (in `libraries/core`) already guards the per-node timing fields — `finish_grace_secs`, `health_check_timeout`, `restart_delay`, `max_restart_delay`, `restart_window` — for exactly this reason (`check_timing_fields`). But two more fields fed to the same conversion were missed:

- **dataflow-level `health_check_interval`** — `binaries/daemon/src/lib.rs`: `descriptor.health_check_interval.map(Duration::from_secs_f64)`, converted right after `check_dataflow` runs.
- **per-input `input_timeout`** — `binaries/daemon/src/lib.rs`: `Duration::from_secs_f64(timeout)` on both the initial-spawn and the reconnect paths.

A descriptor such as:

```yaml
health_check_interval: -1.0
# or an input with:
#   input_timeout: .inf
```

passes `dora validate` and then **crashes the daemon** the moment the dataflow is started, instead of surfacing a clean descriptor error.

## Fix

- Extract the shared bound check into `check_seconds_field(owner, field, value)`.
- Reuse it from `check_timing_fields` (same error message, no behavior change).
- Apply it to `input_timeout` for every node's inputs (custom **and** runtime operators, via a new `node_inputs` helper) and to the dataflow-level `health_check_interval`.

## Validation

- New unit tests for `check_seconds_field` (accepts `None`/`0`/positive; rejects negative + NaN + ±inf).
- New `check_dataflow` tests: negative `health_check_interval` and non-finite `input_timeout` are rejected with a clear message; valid values are accepted.
- `cargo test -p dora-core`, `cargo fmt --all -- --check`, and `cargo clippy -p dora-core -- -D warnings` pass. On a combined branch with all three sibling PRs, `dora-core` + `dora-cli` + `dora-coordinator` suites pass together (0 failures) and a high-effort `/code-review` of the combined diff surfaced no correctness findings.

---

### DIFF
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -132,7 +132,24 @@ pub fn check_dataflow(dataflow: &Descriptor, working_dir: &Path) -> eyre::Result
         if let descriptor::CoreNodeKind::Custom(custom) = &node.kind {
             check_timing_fields(&node.id, custom)?;
         }
+        // `input_timeout` is a second-valued `f64` that the daemon also feeds
+        // to `Duration::from_secs_f64`, on both the initial-spawn and the
+        // reconnect paths.
+        for (input_id, input) in node_inputs(node) {
+            check_seconds_field(
+                &format!("input `{input_id}` of node `{}`", node.id),
+                "input_timeout",
+                input.input_timeout,
+            )?;
+        }
     }
+    // dataflow-level `health_check_interval` reaches `Duration::from_secs_f64`
+    // in the same way (`binaries/daemon/src/lib.rs`).
+    check_seconds_field(
+        "dataflow",
+        "health_check_interval",
+        dataflow.health_check_interval,
+    )?;
 
     // check that all inputs mappings point to an existing output
     check_wiring(dataflow)?;
@@ -167,6 +184,7 @@ fn check_timing_fields(
     node_id: &NodeId,
     custom: &dora_message::descriptor::CustomNode,
 ) -> eyre::Result<()> {
+    let owner = format!("node `{node_id}`");
     for (field, value) in [
         ("finish_grace_secs", custom.finish_grace_secs),
         ("health_check_timeout", custom.health_check_timeout),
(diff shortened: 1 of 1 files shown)


phil-opp:

This PR is part of a batch opened by an automated Claude code-review pass over `origin/main`. All three are independent, each on its own branch off a freshly-fetched `origin/main`, and each is machine-generated — please review carefully.

### PRs opened (all Correctness)

| PR | Area | Fix |
|----|------|-----|
| **#2612** (this) | `dora-core` | Validate `health_check_interval` + `input_timeout` so a negative/NaN/inf value fails `dora validate` instead of panicking the daemon at `Duration::from_secs_f64`. |
| **#2613** | `dora-cli` | Guard `dora topic hz` window-cutoff with `Instant::checked_sub` — a large `--window` (unbounded) or near-boot uptime otherwise panics on the first message (mirrors the existing `topic info` guard). |
| **#2614** | `dora-coordinator` | Don't treat a partially-finished multi-daemon dataflow as stopped: `Stop`/`StopByName`/`List` now apply the same `running_dataflows` guard the `Clean` handler uses, so `dora stop` actually stops the remaining daemons and `dora list` stops double-listing. |

No Optimization / Documentation / Simplification-only PRs met the bar this run.

### Findings investigated but **not** turned into PRs

| Finding | Reason skipped |
|---------|----------------|
| daemon stdout log-reader busy-loops on a persistent (non-EOF) read error (asymmetric with stderr) | **Duplicate** — already fixed by open PR #2581. |
| runtime/message `queue_size: 0` silently drops all inputs for that id | **Already covered** by open PR #2596 (DropOldest capacity clamp); also consistent with the node-side scheduler. |
| `dora self update --check-only` uses string `!=` instead of a semver comparison (mislabels a newer-than-release build as "update available") | Weak — cosmetic, check-only branch only; the real update path is already semver-aware. |
| mavlink2-bridge `parse_baud` accepts `baud=0` | Weak — fails gracefully as `BridgeError::Config` (no panic), only an opaque message. |
| runtime `InputClosed` handler omits the "all channels closed → break" the `Finished` arm has | Latent only — benign under the current single-operator invariant; not user-visible. |
| coordinator `broadcast_all_nodes_ready` propagates a WS-send `?` that is treated as fatal | Plausible but timing-dependent and reachability unconfirmed; deferred rather than risk a wrong fix. |
| coordinator `?` on `serde_json::to_vec` inside per-request handlers can tear down the event loop | Low — serialization of these well-typed structs effectively never fails. |
| mavlink2-bridge rejects sliced/offset top-level `StructArray` inputs | Not a defect — fail-safe (`Err`, never a panic), at worst over-strict. |

### Validation (combined diff)
- Per-crate: `fmt` + `clippy -D warnings` + tests green for each changed crate; the three changed crates' suites also pass **together** (0 failures).
- `/code-review` (high effort) on the combined diff: no correctness findings (2 minor cleanup notes consciously declined as they mirror existing codebase idioms); `/simplify`: clean.
- A full `cargo test --all` could not complete in the review sandbox — it exhausted the fixed per-session disk allowance during linking (ENOSPC in crates unrelated to these changes). No real compile/test failures were seen, and none of the three PRs change any public API or signatures, so downstream crates are unaffected. CI runs the full matrix.

_Generated by Claude (`claude-opus`) — an AI agent. Machine-generated; review before merging._

---

phil-opp:

The change factors the existing per-node bound check into `check_seconds_field` and extends it to two more second-valued fields that reach `Duration::from_secs_f64` in the daemon: per-input `input_timeout` (covered for both custom nodes and every operator of a runtime node via `node_inputs`) and dataflow-level `health_check_interval`. The predicate (`!is_finite() || < 0.0`) matches the pre-existing timing-field guard, and `None`/zero/positive values are still accepted. Tests cover negative, NaN, and infinite rejection plus the valid cases.

---

---

phil-opp:
**Automated review by Claude** — this is a fully automated review, not verified by a human.

I found one gap that's directly on-topic for this PR's goal (keeping these second-valued fields from panicking the daemon at `Duration::from_secs_f64`):

`check_seconds_field` rejects negative and non-finite values (`!value.is_finite() || value < 0.0`), but `Duration::from_secs_f64` *also* panics on **finite values that overflow `Duration`** (roughly `secs >= u64::MAX`, i.e. `>= ~1.8446744e19`). A descriptor with e.g. `health_check_interval: 1e20` or `input_timeout: 1e20` is finite and non-negative, so it passes the new guard and still panics the daemon at `Duration::from_secs_f64` (`binaries/daemon/src/lib.rs`) — the exact failure class this PR is meant to close.

The same gap applies to the five pre-existing timing fields now routed through this helper, so it's a shared limitation rather than a regression introduced here. Rejecting `value >= u64::MAX as f64` too, or switching to `Duration::try_from_secs_f64` (which returns an error for every panic case, including overflow), would close the whole class in one place.

Otherwise the change looks correct: the units are unambiguously seconds, `0`/positive are accepted, and the input traversal mirrors the daemon's own usage.

---

PR #2613, opened 2026-07-12
fix(cli): avoid Instant underflow panic in `dora topic hz`

phil-opp:
## Issue

`HzStats::record` computed the sliding-window cutoff as `now - self.window_duration`. `Instant - Duration` **panics** with `"overflow when subtracting duration from instant"` whenever `window_duration` exceeds the current monotonic-clock value.

This is reachable from unvalidated CLI input: `parse_window` only rejects `0`, imposing no upper bound. So:

```
dora topic hz --window 100000000000 -d <flow>
```

builds a multi-thousand-year window and panics on the **first received message**. Because the panic happens while the `timestamps` mutex guard is held, the mutex is poisoned and the interactive UI/reader threads then panic on their own `.lock().unwrap()`. The default 10 s window is also affected within the first 10 seconds of monotonic uptime (freshly-booted host / container).

The sibling `dora topic info` path (`info::calculate_hz`) already guards this exact case with `Instant::checked_sub` and an explanatory comment — `hz.rs` was simply not given the same treatment.

## Fix

Apply `Instant::checked_sub` to both cutoff computations in `hz.rs`:

- the per-message prune in `record`, and
- the fixed 1 s sub-window in the interactive UI loop (`run_hz`).

When the cutoff would predate the monotonic epoch, every recorded timestamp is within the window, so all are kept/counted.

## Validation

- Added `record_does_not_underflow_on_huge_window` — panics before the fix, passes after.
- `cargo test -p dora-cli` (`command::topic::hz`), `cargo fmt --all -- --check`, and `cargo clippy -p dora-cli -- -D warnings` pass. On a combined branch with all three sibling PRs, `dora-core` + `dora-cli` + `dora-coordinator` suites pass together (0 failures) and a high-effort `/code-review` of the combined diff surfaced no correctness findings.

---

### DIFF
--- a/binaries/cli/src/command/topic/hz.rs
+++ b/binaries/cli/src/command/topic/hz.rs
@@ -185,12 +185,19 @@ impl HzStats {
     fn record(&self, now: Instant) {
         let mut timestamps = self.timestamps.lock().unwrap_or_else(|e| e.into_inner());
         timestamps.push_back(now);
-        let cutoff = now - self.window_duration;
-        while let Some(&first) = timestamps.front() {
-            if first < cutoff {
-                timestamps.pop_front();
-            } else {
-                break;
+        // `now - window_duration` panics ("overflow when subtracting duration
+        // from instant") when the window exceeds the monotonic-clock value --
+        // reachable via a large `--window` (which `parse_window` leaves
+        // unbounded) or within `window` seconds of the monotonic epoch. When
+        // the cutoff would predate the epoch, every timestamp is within the
+        // window, so there is nothing to prune. Mirrors `info::calculate_hz`.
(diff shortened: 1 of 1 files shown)


phil-opp:

Both cutoff computations now use `Instant::checked_sub`, closing the `Instant - Duration` underflow panic reachable from an unvalidated large `--window` (`parse_window` only rejects 0) and within the first `window` seconds of monotonic uptime. When the cutoff would predate the epoch the guard yields `None`, and both sites then treat every recorded timestamp as within the window (nothing pruned / everything counted), which is the right behavior for a window exceeding uptime. This matches the existing `info::calculate_hz` guard, and the `record_does_not_underflow_on_huge_window` test exercises the previously-panicking path.

---

---

PR #2614, opened 2026-07-12
fix(coordinator): don't treat a partially-finished multi-daemon dataflow as stopped

phil-opp:
## Issue

`dataflow_results` is filled **incrementally — one entry per daemon** — as each daemon reports `DataflowFinishedOnDaemon`, while the dataflow stays in `running_dataflows` until *every* daemon has finished. The `Clean` handler already accounts for this (it skips uuids still in `running_dataflows`, with a comment explaining why), but three other consumers of `dataflow_results` did not:

- **`Stop` / `StopByName`** early-returned `DataflowStopped` on *any* `dataflow_results` entry. On a multi-daemon dataflow where one daemon has finished but others are still running, `dora stop <uuid>` / `--name <name>` replied `"stopped"` and **skipped `stop_dataflow` entirely** — so the still-running daemons' nodes were never told to stop and kept running until they exited on their own.
- **`List`** yielded every `dataflow_results` entry as `Finished`/`Failed` with no `running_dataflows` filter, so a partially-finished dataflow appeared **twice** in `dora list` — once `Running`, once `Finished`/`Failed` — with contradictory statuses.

### Failure scenario

Two-daemon dataflow on A and B. A's nodes all exit → `DataflowFinishedOnDaemon(A)` inserts a partial `dataflow_results[uuid] = {A: ...}` and removes A from `dataflow.daemons`; the dataflow is still running on B. `dora stop <uuid>` finds the partial entry, replies `DataflowStopped`, and `continue`s — B's nodes are never stopped.

## Fix

Gate all three consumers on the dataflow being truly absent from `running_dataflows`, mirroring the `Clean` handler:

```rust
if !running_dataflows.contains_key(&dataflow_uuid)
    && let Some(result) = dataflow_results.get(&dataflow_uuid)
{ /* fast path */ }
```

and, in `List`, `.filter(|(uuid, _)| !running_dataflows.contains_key(uuid))`.

Fully-finished single- and multi-daemon dataflows keep the existing fast paths unchanged.

## Validation

- Added `partial_multi_daemon_dataflow_is_not_stopped_early_or_listed_twice`, exercising the shared guard predicate for both the Stop fast path and the List projection (partial dataflow falls through; fully-finished keeps the fast path and appears once).
- `cargo test -p dora-coordinator`, `cargo fmt --all -- --check`, and `cargo clippy -p dora-coordinator -- -D warnings` pass. On a combined branch with all three sibling PRs, `dora-core` + `dora-cli` + `dora-coordinator` suites pass together (0 failures) and a high-effort `/code-review` of the combined diff surfaced no correctness findings.

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -893,7 +893,18 @@ async fn start_inner(
                             grace_duration,
                             force,
                         } => {
-                            if let Some(result) = dataflow_results.get(&dataflow_uuid) {
+                            // `dataflow_results` is filled incrementally, one
+                            // entry per daemon, while a multi-daemon dataflow is
+                            // still running on the others (see
+                            // `DataflowFinishedOnDaemon`). Only take the
+                            // already-stopped fast path when the dataflow is
+                            // truly gone from `running_dataflows`; otherwise fall
+                            // through to `stop_dataflow` so the daemons that are
+                            // still running actually get told to stop. Mirrors
+                            // the `Clean` handler's guard.
+                            if !running_dataflows.contains_key(&dataflow_uuid)
+                                && let Some(result) = dataflow_results.get(&dataflow_uuid)
+                            {
                                 let reply = ControlRequestReply::DataflowStopped {
                                     uuid: dataflow_uuid,
                                     result: dataflow_result(result, dataflow_uuid, &clock),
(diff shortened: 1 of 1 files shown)


phil-opp:

The `Stop`, `StopByName`, and `List` handlers now consult `dataflow_results` only when the uuid is truly gone from `running_dataflows`, matching the guard the `Clean` handler already uses. This correctly fixes the partial multi-daemon case: a dataflow that has a per-daemon `dataflow_results` entry but is still in `running_dataflows` now falls through to `stop_dataflow` (rather than replying `DataflowStopped` and leaving the remaining daemons' nodes running), and `List` yields it once as `Running` instead of also as `Finished`/`Failed`. Fully-finished dataflows are unaffected since they're absent from `running_dataflows`. The guard predicate and the added regression test line up.

---

---

PR #2616, opened 2026-07-13
fix(ros2-bridge): request action results explicitly

DGHX12345:
## Issue

The native ROS2 action-client examples could hang while waiting for a terminal Fibonacci result.

In the Rust example, result handling used `async_request_result` through a dynamically merged external stream. On cold starts this result future could wedge and never yield, so the node kept running until the outer timeout instead of making progress through Dora timer ticks. The old code also merged feedback/status/result streams into the Dora event stream, which made the example harder to reason about and harder to fail fast when the action server returned a terminal non-success status.

In the C++ example, result requests were tied to goal submission in a way that did not reliably retry/poll result reception. If a result status was not terminal-success, the loop could keep running with the same `goal_id` and no useful state transition. Aborted or canceled goals were logged but did not fail the example, so the runner could continue waiting instead of reporting the real terminal action state.

The Rust action-client runner also killed the ROS2 action server only after the dataflow thread joined successfully. If the dataflow returned an error, cleanup could be skipped or the original error could be obscured by cleanup behavior.

## Fix

- Rework the Rust action-client example to drive action progress from Dora timer ticks instead of merging ROS action streams into the Dora event stream.
- Send one Fibonacci goal, store the accepted `GoalId`, and request the result explicitly with `request_result`.
- Track pending result request IDs and poll `result_client().receive_response()` so unrelated responses are ignored.
- Treat terminal states explicitly:
  - `Succeeded` prints the result and exits the loop.
  - `Canceled` and `Aborted` return errors.
  - `Unknown` and other non-terminal statuses keep polling.
- Preserve the 60-second fail-fast timeout so the example does not hang indefinitely.
- Update the C++ action-client example to:
  - track whether a result request is already outstanding,
  - request results after goal submission,
  - reset result-request state after completion,
  - fail immediately on aborted/canceled terminal states.
- Make the Rust runner collect the dataflow result and ROS cleanup result separately so the action server is still stopped and cleanup errors remain visible.

## Validation

- Ran:
  - `cargo check -p rust-ros2-example-node --features ros2 --bins`
  - `cargo clippy -p rust-ros2-example-node --features ros2 --bins -- -D warnings`
- Also covered by the full branch validation run:
  - `cargo check --examples`

### DIFF
--- a/examples/ros2-bridge/c++/action-client/main.cc
+++ b/examples/ros2-bridge/c++/action-client/main.cc
@@ -35,6 +35,7 @@ int main()
 
     client->wait_for_action(node);
     std::unique_ptr<::rust::Box<::ActionGoalId>> goal_id;
+    bool result_requested = false;
     for(int count = 0; count < 3;) {
         auto event = merged_events.next();
         if (event.is_dora()) { // Event from Dora
@@ -47,8 +48,12 @@ int main()
                     auto goal = example_interfaces::Fibonacci_Goal{ .order = 10 + count };
                     goal_id = std::make_unique<::rust::Box<::ActionGoalId>>(
                         client->send_goal(goal));
+                    result_requested = false;
                     std::cout << "Goal have been sent" << std::endl;
+                }
+                if (!result_requested) {
                     client->request_result(*goal_id);
+                    result_requested = true;
                 }
             } else if (ty == DoraEventType::Stop || ty == DoraEventType::AllInputsClosed) {
                 std::cout << "Received stop event" << std::endl;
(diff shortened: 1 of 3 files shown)


phil-opp:

C++ `main.cc`: `result_requested` is reset whenever a new goal is sent (guarded by `!goal_id`) and after a non-terminal status, and `request_result(*goal_id)` only runs after a block that guarantees `goal_id` is non-null (either it was already set, or the `!goal_id` send-block just set it), so there's no null dereference. Handling `Aborted`/`Canceled` with `return 1` is a clear improvement over the previous silent loop.

Rust `action_client.rs`: the rewrite to explicit `request_result` + `receive_result` is sound — `receive_result` drains `result_client().receive_response()` without blocking, matches by `RmwRequestId`, and returns `None` on an empty queue; the `pending_results.is_empty()` gate keeps a single outstanding request and re-requests after each non-terminal (Unknown) response, matching the old retry semantics. The 60s deadline keeps the fail-fast behavior, and terminal statuses map correctly (Succeeded breaks, Canceled/Aborted bail).

---

---

PR #2617, opened 2026-07-13
fix(ros2-bridge): wait for service readiness in turtle examples

DGHX12345:
## Issue

The ROS2 turtle/service examples had discovery races around `/add_two_ints`.

The C++ and Rust turtle runners started `dora run` immediately after spawning the ROS2 `examples_rclcpp_minimal_service` process. ROS2 service discovery is asynchronous, so the service process can be running while `/add_two_ints` is not yet visible to clients. In that window, the Dora dataflow starts, the turtle node begins sending service requests, and the example can fail or hang depending on discovery timing.

Inside the Rust turtle node, service readiness was checked up front with a blocking wait. That made startup fragile because Dora input processing was blocked while waiting for ROS2 discovery. It also meant a slow service discovery path could prevent the node from handling normal Dora events until the wait completed or failed.

The generated C++ service client wait loop also only retried 10 times. On slower ROS2 environments this can be too short, producing intermittent failures even though the service eventually appears.

There was an additional compile-time issue in the Rust turtle node: the code referenced `add_two_ints_service_ready` without defining it, and `rand::random()` did not infer the intended `i64` type reliably in the service request path.

## Fix

- Update both turtle runners to wait for `/add_two_ints` to appear in `ros2 service list` before starting the Dora dataflow.
- Keep ROS child cleanup explicit:
  - collect the service wait result,
  - collect the dataflow result,
  - kill the ROS service and turtlesim processes,
  - then report any wait/dataflow/cleanup errors.
- Change the Rust turtle node to poll service readiness from the `service_timer` input path with a short timeout instead of blocking startup.
- Add `add_two_ints_service_ready` using `wait_for_service` plus a 100 ms timeout.
- Add a 60-second service discovery deadline so the example fails with `add_two_ints service not available` instead of hanging indefinitely.
- Explicitly type generated service operands as `i64` and compare against `wrapping_add`.
- Increase generated C++ service-client readiness retries from 10 to 30.

## Validation

- Reproduced the compile failure with:
  - `cargo check -p rust-ros2-example-node --features ros2 --bin rust-ros2-example-node-turtle`
- Verified the fix with:
  - `cargo check -p rust-ros2-example-node --features ros2 --bins`
  - `cargo check -p dora-ros2-bridge-msg-gen`
  - `cargo clippy -p rust-ros2-example-node --features ros2 --bins -- -D warnings`
  - `cargo clippy -p dora-ros2-bridge-msg-gen -- -D warnings`
  - `cargo test -p dora-ros2-bridge-msg-gen`

### DIFF
--- a/examples/ros2-bridge/c++/turtle/run.rs
+++ b/examples/ros2-bridge/c++/turtle/run.rs
@@ -1,5 +1,5 @@
 use eyre::{Context, bail};
-use std::{env::consts::EXE_SUFFIX, path::Path};
+use std::{env::consts::EXE_SUFFIX, path::Path, process::Command, time::Duration};
 
 use process_wrap::std::{StdChildWrapper as ChildWrapper, StdCommandWrap as CommandWrap};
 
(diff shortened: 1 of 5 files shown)


phil-opp:

Points I checked:
- In both `run.rs` files the two `kill()` calls run unconditionally (results bound to locals) and are propagated only afterward via `?`, so both ROS2 child processes are always reaped even when the dataflow or the service wait fails — no leaked processes on the error path.
- `turtle.rs` replaces the up-front blocking `wait_for_service` loop with a non-blocking readiness poll on each `service_timer` tick, bounded by a 60s deadline that `bail!`s rather than hanging; once ready the flag latches so requests proceed.
- The retry-count increases (10→30) and the `&ros_node`→`ros_node` argument tweak are behavior-preserving.

---

---

PR #2620, opened 2026-07-13
fix(python): reject non-finite/negative timeouts instead of panicking

phil-opp:
## Problem

`Node.next(timeout=...)`, `Node.recv_async(timeout=...)`, and the `run(..., stop_after=...)` entry points (in both `dora-node-api-python` and `dora-cli-api-python`) accept a user-supplied float number of seconds and feed it straight into `Duration::from_secs_f32` / `Duration::from_secs_f64`. Both **panic** when the value is negative, NaN, or infinite. From Python that surfaces as an opaque `PanicException` (and, for `recv_async`, a panic inside a future) rather than a clean, catchable error.

Concrete triggers before this change:

- `node.next(timeout=-1.0)` / `timeout=float('nan')` / `timeout=float('inf')`
- `await node.recv_async(timeout=-1.0)`
- `dora.run(path, stop_after=-1.0)`

## Fix

Switch to the non-panicking `Duration::try_from_secs_f32` / `try_from_secs_f64` and map the error to a `ValueError` (node methods) or an `eyre` error (`run`). A small `timeout_to_duration` helper is shared by the two `PyResult`-returning node methods. This mirrors the pattern already used in `dora-operator-api-python` for datetime timestamp conversion (`try_from_secs_f64`).

## Validation

The affected crates are PyO3 extension modules not built by `cargo test` (they require maturin/Python linkage), so verification is `cargo check -p dora-node-api-python -p dora-cli-api-python` (clean) plus `cargo fmt --all -- --check`. The combined change also passes a workspace-wide `cargo check --all` and `cargo check --examples`.

---

---

### DIFF
--- a/apis/python/cli/src/lib.rs
+++ b/apis/python/cli/src/lib.rs
@@ -47,7 +47,10 @@ pub fn build(
 pub fn run(dataflow_path: String, uv: Option<bool>, stop_after: Option<f64>) -> eyre::Result<()> {
     use dora_cli::Executable;
 
-    let stop_after_duration = stop_after.map(std::time::Duration::from_secs_f64);
+    let stop_after_duration = stop_after
+        .map(std::time::Duration::try_from_secs_f64)
+        .transpose()
+        .map_err(|err| eyre::eyre!("invalid stop_after of {stop_after:?} seconds: {err}"))?;
     let mut run = dora_cli::RunCommand::new(dataflow_path);
     if let Some(uv) = uv {
         run.uv = uv;
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -34,6 +34,21 @@ static RUNTIME: LazyLock<Runtime> = LazyLock::new(|| {
         .expect("Failed to create Tokio runtime")
 });
 
+/// Convert a user-supplied timeout in seconds into a [`Duration`], rejecting
+/// negative, NaN, or infinite values with a clean `ValueError` instead of the
+/// panic that [`Duration::from_secs_f32`] raises on such inputs.
+fn timeout_to_duration(timeout: Option<f32>) -> PyResult<Option<Duration>> {
+    timeout
+        .map(|secs| {
+            Duration::try_from_secs_f32(secs).map_err(|err| {
+                pyo3::exceptions::PyValueError::new_err(format!(
+                    "invalid timeout of {secs} seconds: {err}"
+                ))
+            })
+        })
+        .transpose()
+}
+
 fn runtime() -> PyResult<&'static Runtime> {
     // Access the LazyLock; if the builder panicked, this will propagate.
     // In normal operation the expect above fires only if thread/memory
@@ -604,7 +619,8 @@ impl Node {
     #[pyo3(signature = (timeout=None))]
     #[allow(clippy::should_implement_trait)]
     pub fn next(&self, py: Python, timeout: Option<f32>) -> PyResult<Option<Py<PyDict>>> {
-        let event = py.detach(|| self.events.recv(timeout.map(Duration::from_secs_f32)));
+        let timeout = timeout_to_duration(timeout)?;
+        let event = py.detach(|| self.events.recv(timeout));
         if let Some(event) = event {
             let dict = event
                 .to_py_dict(py)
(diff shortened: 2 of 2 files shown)


phil-opp:

This PR is one of a batch opened from an automated review of the repository against current `origin/main`. Each finding was verified by reading the current code and constructing a concrete failure scenario; each PR is independent and branches off a fresh `origin/main`.

**These changes are machine-generated by Claude (an AI agent). Please review carefully before merging.**

### PRs opened, by category

**Correctness — reachable panics on user input**
- #2620 — `fix(python)`: `Node.next` / `recv_async` / `run` fed a user float straight into `Duration::from_secs_f32/f64`, which panics on negative/NaN/inf; switched to the non-panicking `try_from_secs_*`.
- #2621 — `fix(cli)`: `dora node remove --grace` was a bare `f64` → `from_secs_f64` panic on `-1`/`nan`/`1e30`; now uses the same `parse_duration` validator as `node stop`/`node restart`.

**Correctness — silent data corruption**
- #2623 — `fix(python)`: a tuple of strings in operator metadata (`{"k": ("a","b")}`) fell through to `Parameter::String("('a', 'b')")` because the string-list branch checked `PyList` only while the int/float branches accept `PyTuple` too.

**Correctness — misleading diagnostics**
- #2624 — `fix(message)`: node exit-status `Display` mislabeled signal 22 as `SIGABRT` (a copy of signal 6) and 23 as `NSIG` (the signal *count*); corrected to `SIGTTOU`/`SIGURG` with a regression test.

**Robustness (daemon)**
- #2622 — `fix(daemon)`: node stdout reader could busy-spin at 100% CPU on a persistent read error (stderr already terminated correctly); and the stderr "most recent" ring retained *untruncated* lines, bypassing the documented per-line 1 MiB heap-exhaustion guard. Both fixed.

### Validation

Each branch was verified with fmt + clippy + tests for its affected crate (`dora-daemon`: 96 passed; `dora-message`: incl. new test; `dora-cli`: pass; the PyO3 crates aren't `cargo test`-able and were checked with `cargo check`). The combined diff additionally passes a workspace-wide `cargo check --all`, `cargo check --examples`, and the `/review` + `/simplify` gates. (The full `cargo test --all` build exceeds this environment's disk allowance, so it was run per-affected-crate plus a workspace-wide compile check.)

### Note on the failing `audit` check

The `Audit (cargo-audit + cargo-deny)` / `ci/circleci: audit` failure is **pre-existing and unrelated to these diffs** — it fails identically on every PR in this batch, including ones that touch no pyo3 and no dependencies (#2621, #2622), and none of these PRs modify `Cargo.toml`/`Cargo.lock`. cargo-deny reports `advisories FAILED` and the `deny.toml` ignores for `RUSTSEC-2026-0176`/`0177` now "match no crate", i.e. the advisory landscape shifted (likely after the #2445 dependency upgrade). That belongs in a separate `deny.toml` maintenance PR, not these focused fixes.

- **`host_log` drops Python log records with `levelno` > 255** (`apis/python/node`): real but niche (only custom levels above 255), and widening the extract risks perturbing the existing level ladder — low value, deferred.
- **`dora logs --local` ignores `--grep`/`--tail`/`--since`/`--until` for raw (non-JSON) legacy `.txt` files**: genuine, but the clean fix requires threading filter args into `read_log_file` and reshaping how raw lines enter the pipeline — more surface area than warranted for a legacy path; deferred.
- **`rotate_log_files` with `max_rotated_files: 0` keeps one file instead of zero**: cosmetic, no failure scenario.
- Several heavily-guarded areas were reviewed and found correct (memory-pool DORADMA header parsing, IPC framing/offset arithmetic, coordinator state-log pruning, `sample_handler` buffer-protocol state machine, TCP length-prefix bounds) — no action.

---

phil-opp:

No issues found. Swapping `Duration::from_secs_f32/f64` for the `try_from_secs_*` variants correctly eliminates the panic path on negative/NaN/infinite user-supplied timeouts, surfacing a clean `ValueError` (node methods) or `eyre` error (`run`) instead of a `PanicException`. The `.map(..).transpose()?` handling of the `Option` is correct, and the `timeout_to_duration` helper de-duplicates the two node call sites cleanly.

---

phil-opp:
The merge-queue `E2E Tests` failure looks like an **unrelated flake**, not something in this diff — a re-queue should clear it.

The failure is a startup race in `daemon-reconnect-e2e.rs::node_survives_daemon_watchdog_disconnect`:

```
ERROR dora_coordinator::ws_control: control request failed: no unnamed daemon connections
  binaries/coordinator/src/run/mod.rs:158
--- daemon log ---
INFO  Connected to dora-coordinator at ws://127.0.0.1:.../api/daemon
panicked at tests/daemon-reconnect-e2e.rs:456: dora start failed
```

The timestamps show `dora start` reaching the coordinator ~0.4 ms *before* the daemon finished registering (`.743707` vs `.744089`), so the control request hit "no unnamed daemon connections". The sibling suites in the same job passed (`ws-cli-e2e`: 11 passed; `fault-tolerance-e2e`).

This PR only touches the PyO3 bindings (`apis/python/node`, `apis/python/cli`) and can't affect a Rust daemon↔coordinator reconnect test, so I don't think any change here is warranted — flagging in case it's useful when re-queuing.

---

PR #2621, opened 2026-07-13
fix(cli): validate `dora node remove --grace` like sibling node commands

phil-opp:
## Problem

`--grace` on `dora node remove` was parsed as a bare `f64` with clap's default parser and then passed to `Duration::from_secs_f64`, which **panics** on negative, NaN, infinite, or overflowing values:

```
dora node remove mynode --grace=-1     # panic: "value is negative"
dora node remove mynode --grace nan    # panic: "value is not finite"
dora node remove mynode --grace 1e30   # panic: "value is too big"
```

So the CLI aborted with a Rust panic/backtrace instead of a clean error.

## Fix

The sibling commands `dora node stop` and `dora node restart` already parse `--grace` with `value_parser = crate::common::parse_duration`, which rejects negatives and garbage and accepts human-friendly forms (`30s`, `500ms`, `1m`, and bare seconds like `5`). Switch `remove` to the same parser so `grace_duration` is a validated `Duration` and the three cases above become clean "invalid value" errors. This also makes the three `dora node` grace flags consistent.

## Validation

- `cargo clippy -p dora-cli -- -D warnings` (clean)
- `cargo test -p dora-cli` (pass)

---

---

### DIFF
--- a/binaries/cli/src/command/node/remove.rs
+++ b/binaries/cli/src/command/node/remove.rs
@@ -16,9 +16,9 @@ pub struct Remove {
     #[clap(long, short = 'd', value_name = "NAME_OR_UUID")]
     dataflow: Option<String>,
     node: String,
-    /// Grace period in seconds before force-killing the node.
-    #[clap(long)]
-    grace: Option<f64>,
+    /// Grace period before force-killing the node (e.g. `30s`, `500ms`, `1m`).
+    #[clap(long, value_name = "DURATION", value_parser = crate::common::parse_duration)]
+    grace: Option<std::time::Duration>,
 }
 
 impl Executable for Remove {
@@ -37,7 +37,7 @@ impl Executable for Remove {
             &ControlRequest::RemoveNode {
                 dataflow_id,
                 node_id: node_id.clone(),
-                grace_duration: self.grace.map(std::time::Duration::from_secs_f64),
+                grace_duration: self.grace,
             },
         )?;
 


phil-opp:

No issues found.

Switching `--grace` from a bare `f64` (fed into `Duration::from_secs_f64`, which panics on negative / NaN / infinite / overflowing input) to `value_parser = crate::common::parse_duration` producing an `Option<Duration>` is a clean fix: the parser rejects those cases as a normal clap "invalid value" error instead of aborting with a Rust panic, and it matches the `dora node stop` / `dora node restart` siblings. `grace_duration: self.grace` now passes the validated `Duration` straight through, and since `parse_duration` still accepts bare seconds (e.g. `5`) alongside `30s`/`500ms`/`1m`, existing `--grace <seconds>` invocations remain valid — so no user-facing regression.

---

phil-opp:

Switching `--grace` to `parse_duration` is the right fix for the panic-on-garbage cases, but there's a backward-compat wrinkle worth calling out. The old parser was a clap-default `f64` fed to `Duration::from_secs_f64`, so **fractional bare seconds** like `--grace 0.5` or `--grace 2.5` were valid (500 ms / 2.5 s). `parse_duration` delegates to `duration_str`, whose numeric grammar is integer-digits-only (`digit1` → `u64`, no decimal point), so those inputs now fail with an "invalid value" error rather than being accepted.

Integer bare seconds (`--grace 5`) and unit forms (`500ms`, `30s`, `1m`) all still work, and sub-second is still expressible via `500ms`, so there is a migration path. But any existing script passing a fractional `--grace` value will break. This matches the `stop`/`restart` siblings, so it may well be intentional consistency — but it's a user-visible behavior change that the new help text (`e.g. 30s, 500ms, 1m`) doesn't make obvious. Worth a note in the PR description (or a changelog entry) that fractional bare seconds are intentionally dropped. Not a blocker.

---

PR #2622, opened 2026-07-13
fix(daemon): harden node stdout/stderr log readers

phil-opp:
## Problem

Two independent robustness issues in the per-node log readers spawned by `PreparedNode` (`binaries/daemon/src/spawn/prepared.rs`):

1. **stdout reader busy-loops on a persistent read error.** On a non-EOF read error the stdout task set `finished = false` and re-entered the loop. A broken/errored stdout fd keeps returning the same (non-`WouldBlock`) error, so the reader spins at 100% CPU, emitting a warning and pushing a `LogLine` every iteration for the rest of the node's life. The sibling stderr reader already terminates on error (`finished = true`) — the two were inconsistent.

2. **stderr "most recent" ring retained untruncated lines.** Each stderr line was `force_push`ed into `node_stderr_most_recent` (a bounded ring of `STDERR_LOG_LINES_MAX = 500` lines, drained on node failure by `extract_err_from_stderr`) *before* truncation; only the copy sent to the log channel passed through `truncate_log_line`. That bypasses the per-line `MAX_LOG_LINE_BYTES` (1 MiB) guard whose documented purpose is to stop "a malicious or buggy node from causing heap exhaustion via a single multi-GB stdout/stderr line" — up to ~500 multi-MB/GB lines could be retained at once.

## Fix

1. Make the stdout `Err` arm terminate the reader (`finished = true`), mirroring the stderr reader.
2. Truncate the stderr line before pushing it into `node_stderr_most_recent`.

Both are small and align the two readers' behavior; comments explain the intent at each site.

## Validation

- `cargo clippy -p dora-daemon -- -D warnings` (clean)
- `cargo test -p dora-daemon` (96 passed)

---

---

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -588,10 +588,15 @@ impl PreparedNode {
                     Ok(0) => true,
                     Ok(_) => false,
                     Err(err) => {
+                        // Terminate the reader on a persistent read error
+                        // instead of re-entering the loop: a broken stdout fd
+                        // keeps returning the same (non-`WouldBlock`) error, so
+                        // `finished = false` would busy-spin at 100% CPU while
+                        // flooding logs. Mirrors the stderr reader below.
                         logger_c
                             .log(LogLevel::Warn, Some("daemon".into()), format!("{err:?}"))
                             .await;
-                        false
+                        true
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Both fixes check out against the surrounding code:

1. The stdout reader on a non-EOF error now sets `finished = true`, matching the sibling stderr reader (line ~656). Since `read_until` is tokio's async read, a returned `Err` is a genuine terminal error (not `WouldBlock`, which tokio handles internally), so a persistent broken-fd error would indeed have busy-spun at 100% CPU before this change. Terminating is the right behavior. The post-loop handling of the empty `raw` on the terminating iteration matches the existing EOF (`Ok(0)`) path.

2. `node_stderr_most_recent.force_push` previously retained the untruncated line, bypassing the per-line `MAX_LOG_LINE_BYTES` guard that the channel copy already goes through — up to `STDERR_LOG_LINES_MAX` oversized lines could be held alive. Truncating before `force_push` closes that gap.

---

phil-opp:
Closing in favor of #2581

PR #2624, opened 2026-07-13
fix(message): correct SIGTTOU/SIGURG names in exit-status display

phil-opp:
## Problem

`NodeError`'s `Display` maps the POSIX signal that killed a node to a human-readable name. Two entries were wrong on Linux:

- signal **22** was labeled `SIGABRT` — a copy of the entry already used for signal 6. Signal 22 is `SIGTTOU`.
- signal **23** was labeled `NSIG` — `NSIG` is the *count* of signals (~65), never a signal delivered to a process. Signal 23 is `SIGURG`.

So a node killed by `SIGTTOU` (e.g. a background node writing to a controlling terminal with `TOSTOP` set) was reported as "exited because of signal SIGABRT", pointing debugging at an abort that never happened, and `SIGURG` was reported as the nonsensical `NSIG`.

Severity is low — only the human-readable diagnostic string is affected, not control flow — but the string was misleading.

## Fix

Correct the two names and add a regression test covering signals 6/15/22/23 and the raw-number fallback.

## Validation

- `cargo test -p dora-message` (pass, incl. the new test)
- `cargo clippy -p dora-message -- -D warnings` (clean)

---

---

### DIFF
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -475,6 +475,36 @@ mod tests {
         assert_eq!(DaemonId::from_display_str(""), None);
     }
 
+    /// `NodeExitStatus::Signal` is populated from the real POSIX signal
+    /// number that killed the node. Signal 22 is `SIGTTOU` and 23 is `SIGURG`
+    /// on Linux; they were previously mislabeled `SIGABRT` (a copy of the
+    /// entry for signal 6) and `NSIG` (the signal *count*, never delivered),
+    /// which pointed debugging at the wrong failure.
+    #[test]
+    fn signal_exit_status_uses_correct_signal_names() {
+        let timestamp = uhlc::HLC::default().new_timestamp();
+        let display_for = |signal: i32| {
+            NodeError {
+                timestamp,
+                cause: NodeErrorCause::Other {
+                    stderr: String::new(),
+                },
+                exit_status: NodeExitStatus::Signal(signal),
+            }
+            .to_string()
+        };
+
+        assert!(display_for(15).contains("SIGTERM"));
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. On Linux, signal 22 is `SIGTTOU` and 23 is `SIGURG`, so the previous `SIGABRT` (a duplicate of the signal-6 entry) and `NSIG` (the signal *count*, never delivered) labels were genuinely wrong; the correction and regression test are right.

Minor, non-blocking: this whole table is Linux signal-number-specific (already true before this PR — e.g. 22/23 differ on macOS/BSD, where 23 is `SIGIO`), so the labels can still be misleading on non-Linux nodes. Out of scope for this fix, just noting the limitation isn't newly introduced here.

---

phil-opp:

Looking at the current diff, this PR no longer contains the signal-name fix it describes — the only remaining change is the new `signal_exit_status_uses_correct_signal_names` test. The actual correction is already on `main`: `libraries/message/src/common.rs` already maps `22 => "SIGTTOU"` and `23 => "SIGURG"`, and there are already regression tests for it (`node_error_signal_display_uses_linux_signal_names` covers 6/22/23, and `node_error_signal_display_falls_back_to_signal_number` covers the raw-number fallback).

The added test overlaps that existing coverage aside from also asserting signal 15 (`SIGTERM`), the 99 fallback, and the negative cases (`!SIGABRT` for 22, `!NSIG` for 23). So I don't see what this PR still adds over `main` in its current form — it looks like the fix landed separately and this could be closed, or trimmed down to just the incremental assertions if those are considered worth keeping.

---

phil-opp:
Already fixed in #2592

PR #2297, opened 2026-06-22
refactor(message): remove unused ResolvedNodeSource descriptor enum

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and opened via a maintainer's account because the bot cannot author PRs directly. It has **not** been hand-reviewed by a human yet — please treat it as a suggestion and review before merging.

## What

Removes the unused `ResolvedNodeSource` enum from `dora-message`'s descriptor module:

```rust
#[allow(missing_docs)]
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
pub enum ResolvedNodeSource {
    Local,
    GitCommit { repo: String, commit_hash: String },
}
```

## Why

`ResolvedNodeSource` has **zero references anywhere in the workspace** — a repo-wide search returns only the definition.

It looks like a "resolved" counterpart to the live [`NodeSource`](https://github.com/dora-rs/dora/blob/main/libraries/message/src/descriptor.rs) enum (branch → pinned commit), but nothing constructs it, returns it, or embeds it: it is not part of the `Descriptor` tree and — despite deriving `JsonSchema` — it is **not reachable from the schema root**, so it is absent from the checked-in `dora-schema.json`. It is orphaned scaffolding.

## Verification

- `cargo clippy --all -- -D warnings` (excluding the Python packages) ✅
- `cargo test -p dora-message` ✅
- `cargo test -p dora-core checked_in_schema_is_current` ✅ — schema unchanged
- `cargo fmt --all -- --check` ✅

## Note

`dora-message` is a published library, so this is technically a public-API change. Of the three dead-code candidates in this batch this is the least certain: if `ResolvedNodeSource` is intended as scaffolding for in-progress git-source resolution (branch-pin → commit), feel free to close and I'll leave it be. As it stands it is entirely unreferenced.

---
_Generated by [Claude Code](https://claude.ai/code/session_018wPn3gy8cweX8VPbgorn55)_

PR #2296, opened 2026-06-22
refactor(message): remove unused PythonOperatorConfig descriptor struct

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and opened via a maintainer's account because the bot cannot author PRs directly. It has **not** been hand-reviewed by a human yet — please treat it as a suggestion and review before merging.

## What

Removes the unused `PythonOperatorConfig` struct from `dora-message`'s descriptor module (plus the now-unused `InputMapping` import):

```rust
#[allow(missing_docs)]
#[derive(Debug, Serialize, Deserialize, Clone)]
#[serde(deny_unknown_fields)]
pub struct PythonOperatorConfig {
    pub path: PathBuf,
    pub inputs: BTreeMap<DataId, InputMapping>,
    pub outputs: BTreeSet<DataId>,
    pub output_framing: BTreeMap<DataId, OutputFraming>,
}
```

## Why

`PythonOperatorConfig` has **zero references anywhere in the workspace** — a repo-wide search returns only the definition.

The actual operator descriptor types use a different config struct: `OperatorDefinition` and `SingleOperatorDefinition` both carry an [`OperatorConfig`](https://github.com/dora-rs/dora/blob/main/libraries/message/src/descriptor.rs) (which holds an `OperatorSource`). `PythonOperatorConfig` is **not** embedded in the `Descriptor` tree, is **not** reachable from the JSON schema (it doesn't even derive `JsonSchema`, and it's absent from `dora-schema.json`), and cannot be deserialized from any real dataflow YAML. It's orphaned leftover scaffolding.

## Verification

- `cargo clippy --all -- -D warnings` (excluding the Python packages) ✅
- `cargo test -p dora-message` ✅
- `cargo check --all` (excluding the Python packages) ✅ — nothing else references it
- `cargo test -p dora-core checked_in_schema_is_current` ✅ — schema unchanged
- `cargo fmt --all -- --check` ✅

## Note

`dora-message` is a published library, so this is technically a public-API change. If `PythonOperatorConfig` is reserved for a planned python-operator descriptor path that isn't wired up yet, feel free to close — but as it stands it is entirely unreferenced and unreachable.

---
_Generated by [Claude Code](https://claude.ai/code/session_018wPn3gy8cweX8VPbgorn55)_

PR #2625, opened 2026-07-14
refactor(runtime): remove unused OperatorEvent::AllocateOutputSample variant

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and is opened under a maintainer's account only because the automation authenticates with that token — the change was written by Claude, not by the account owner. It has **not** been hand-reviewed by a human. Please review before merging.

## What

Removes the `AllocateOutputSample` variant from `OperatorEvent` in `binaries/runtime` (crate `dora-runtime`):

```rust
AllocateOutputSample {
    len: usize,
    sample: oneshot::Sender<eyre::Result<DataSample>>,
},
```

## Why it's dead

The variant has **no construction site anywhere in the workspace** — nothing ever sends `OperatorEvent::AllocateOutputSample`, so the match arm handling it in the runtime event loop (`binaries/runtime/src/lib.rs`) was unreachable. A workspace-wide search for the variant returns only its definition and that single unreachable consumer:

```
$ rg -n 'AllocateOutputSample' --glob '!target'
binaries/runtime/src/operator/mod.rs:74:    AllocateOutputSample {          # definition
binaries/runtime/src/lib.rs:216:   OperatorEvent::AllocateOutputSample ...  # unreachable arm
```

The deadness was masked by the `#[allow(dead_code, clippy::large_enum_variant)]` on the enum. The other variants are all live — `Output` (operator output), `Error`, `Panic`, and `Finished` are constructed in `python.rs` / `shared_lib.rs`. Output-sample allocation is **not** routed through this event; the live path allocates directly via `DoraNode::allocate_data_sample`.

`dora-runtime` is a binary crate, so `OperatorEvent` is not a published-library API surface.

## Changes (cascade)

- Delete the `AllocateOutputSample` variant (`operator/mod.rs`)
- Delete the unreachable match arm (`lib.rs`) — required, since matching a non-existent variant won't compile
- Drop the now-unused `DataSample` import (it was referenced only by the removed field)
- Drop the `dead_code` portion of the enum's `#[allow(...)]`, keeping `clippy::large_enum_variant`

`DoraNode::allocate_data_sample` in `dora-node-api` is left untouched (it's public node API, used directly by node authors).

## Verification

- `cargo fmt -p dora-runtime -- --check` — clean
- `cargo clippy -p dora-runtime -- -D warnings` — clean (no orphaned imports; `oneshot` is still used by `run_operator`'s `init_done` param)
[...]

PR #2626, opened 2026-07-14
refactor(node-api): remove never-constructed NodeError::Connection variant

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and is opened under a maintainer's account only because the automation authenticates with that token — the change was written by Claude, not by the account owner. It has **not** been hand-reviewed by a human. Please review before merging, and close it if the variant is being reserved for planned work (see the note below).

## What

Removes the `Connection(String)` variant from `NodeError` in `apis/rust/node` (crate `dora-node-api`):

```rust
/// Could not connect to the daemon or lost connection.
#[error("connection error: {0}")]
Connection(String),
```

## Why it's dead

`NodeError::Connection` has **no construction site anywhere in the workspace** — the library never returns it. A workspace-wide search finds only the definition:

```
$ rg -n 'NodeError::Connection' --glob '!target'
apis/rust/node/src/error.rs:11:    Connection(String),   # definition only
```

Every other variant is actually produced: `Init` (config/env/daemon handshake, ~12 sites), `Output` (~8 sites), `Data` (descriptor/allocation), and `Internal` via the `#[from] eyre::Report]` blanket `?` conversion. The connection-failure paths that exist today surface as `Init(..)` during the daemon handshake or as `Internal(..)`.

Error-enum variants are **produced by the library and matched by consumers** — they are not something a downstream user *calls*. Because the crate never constructs `Connection`, no consumer can ever observe or match it, so removing it drops no capability. This is unlike the maintainer-rejected removals of callable extension points (`TracingBuilder::add_layer`, `HEADER_OPERATOR_API`); it's closer to the already-merged orphaned-scaffolding removals (#2296 `PythonOperatorConfig`, #2297 `ResolvedNodeSource`).

## Note for maintainers

`dora-node-api` is published and `NodeError` is not `#[non_exhaustive]`, so dropping a variant is technically a semver-breaking change — worth folding into the next version bump. It cannot break any consumer at runtime, since the variant is never produced. **If `Connection` is intentionally reserved for the daemon-reconnect work** (a dedicated connection-lost error would be its natural home), feel free to close this — as it stands it is entirely unconstructed.

## Verification

- `cargo fmt -p dora-node-api -- --check` — clean
- `cargo build -p dora-node-api` + `cargo clippy -p dora-node-api --lib -- -D warnings` — clean
- Diff is a 3-line deletion in `error.rs` only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KFp6v3cwJhCDFEQtPV1dNc) — automated dead-code sweep. No human has verified this; please double-check before merging._

---
_Generated by [Claude Code](https://claude.ai/code/session_01KFp6v3cwJhCDFEQtPV1dNc)_

### DIFF
--- a/apis/rust/node/src/error.rs
+++ b/apis/rust/node/src/error.rs
@@ -6,9 +6,6 @@ pub enum NodeError {
     /// Initialization failed (config parsing, env vars, daemon handshake).
     #[error("initialization failed: {0}")]
     Init(String),
-    /// Could not connect to the daemon or lost connection.
-    #[error("connection error: {0}")]
-    Connection(String),
     /// Sending an output or closing outputs failed.
     #[error("output error: {0}")]
     Output(String),


phil-opp:
Heads-up: the `Audit (cargo-audit + cargo-deny)` / `ci/circleci: audit` gate will fail on this PR, but it is **pre-existing and unrelated to this diff** — this PR only deletes a dead enum variant and does not touch `Cargo.lock`.

The gate fails on a yanked transitive crate, not an advisory in the changed code:

```
error[yanked]: detected yanked crate (try `cargo update -p spin`)
  spin 0.9.8   (buddy_system_allocator 0.10 → zenoh-shm → zenoh)
advisories FAILED, bans ok, licenses ok, sources ok
```

`spin 0.9.8` was yanked from crates.io upstream, so `cargo-deny` fails advisories for the whole workspace — this reproduces on `main` and every open PR identically. Resolving it is dependency maintenance (`cargo update -p spin`, or a `deny.toml` entry), out of scope for a pure dead-code deletion, so I've left it out of this PR.

---
🤖 _Automated note from Claude (dead-code sweep) — no human has verified this; please double-check before relying on it._

---

phil-opp:
The `ci/circleci: contract-tests` failure here is a flake, not a defect in this diff:

- This PR only deletes `NodeError::Connection`, a variant the library **never constructs** — no code creates or matches it, so runtime behavior is unchanged and no dataflow-execution contract test can be affected.
- All GitHub Actions gates on this SHA are green — `Clippy`, `Check`, `Format`, `License check`, `Typos`, `Unwrap budget` — so the change compiles and lints cleanly workspace-wide. Only the pre-existing `spin 0.9.8` audit gate (noted above) and this networked suite are red.
- The contract suite spins up `dora up`/`start`/`stop` with Python venvs and timing-sensitive polling; it flaked on the concurrent sibling PR #2625 at the same time, which points to suite-wide flakiness rather than anything in either diff.

A re-run should clear it. The CircleCI job page requires auth so I couldn't inspect the exact assertion — flagging rather than pushing a speculative change.

---
🤖 _Automated note from Claude (dead-code sweep) — no human has verified this; please double-check before relying on it._

---

phil-opp:

No correctness issues — I confirmed `NodeError::Connection` has no construction site in the workspace, so this is a safe dead-code removal.

One thing worth flagging: `NodeError` is a `pub enum` in the published `dora-node-api` crate and is **not** `#[non_exhaustive]`. Removing a variant is technically a SemVer-breaking change — any downstream code doing an exhaustive `match` on `NodeError` would fail to compile after upgrading (code using a `_` arm is unaffected). Given the variant was never constructed the practical impact is small, but this likely warrants a minor-version bump / changelog note rather than being treated as a pure no-op.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review, not verified by a human.

The removal itself looks safe: `NodeError::Connection` has no construction site anywhere, so nothing breaks at runtime. (Removing a variant from the published `dora-node-api` `NodeError` is a semver-minor break for downstream exhaustive `match`es, which the PR body already acknowledges.)

One thing the diff misses: two docs still list the removed variant, so they'll be inaccurate once this merges:
- `docs/api-rust.md:421` — `Connection(String),  // daemon connection lost`
- `guide/src/languages/rust.md:421` — same line

Worth dropping those lines in this PR so the documented `NodeError` matches the enum.

---

phil-opp:
Status update on the rebased commit (`402d7d5`):

- **All GitHub Actions checks are green** — including `Audit`, now that the rebase onto current `main` picked up the fix for the `spin 0.9.8` yank (my earlier audit note no longer applies). The GHA `Test` job is *skipped* here by path-filtering.
- The **only** red check is `ci/circleci: test-linux`. A compile break is ruled out (GHA `Check`/`Clippy` green), and no test references the removed `NodeError::Connection` variant — the change is a reviewed no-op.
- The same `ci/circleci: test-linux` job is failing concurrently on the sibling PR #2625 (an unrelated, equally-trivial dead-code deletion, also just rebased onto `main`). Two independent no-op diffs failing the same job at the same time points to a **`main`-level `test-linux` issue** rather than either diff.

I can't name the exact failing test — the CircleCI log is auth-gated and I can't faithfully reproduce `test-linux` locally (its zenoh tests need networking my sandbox restricts) — so I'm flagging rather than guessing at a fix. A re-run, or checking whether `test-linux` is red on `main`, is the natural next step.

---
🤖 _Automated note from Claude (dead-code sweep) — no human has verified this; please double-check before relying on it._

---

phil-opp:

Re-reviewed after the rebase onto `main` (commit `402d7d5`, committed 2026-07-17). The diff is unchanged — still just the deletion of `NodeError::Connection` from `apis/rust/node/src/error.rs`, which is safe (no construction site).

The previously-flagged docs issue is still outstanding: both `docs/api-rust.md:421` and `guide/src/languages/rust.md:421` still list `Connection(String),  // daemon connection lost`, so they'll be inaccurate once this merges. Worth dropping those two lines here.

---

PR #2628, opened 2026-07-14
fix(daemon): treat Windows STATUS_CONTROL_C_EXIT as a clean planned stop

phil-opp:
## Problem

Recurring Windows nightly failures tracked in #2425 (July 10–13, consistently on the `cli-tests` and `examples` jobs) report:

```
Node `webcam` failed: exited with code -1073741510
```

`-1073741510` is `0xC000013A` = **`STATUS_CONTROL_C_EXIT`** — the exit code Windows assigns to a process terminated by an unhandled console `CTRL_C` / `CTRL_BREAK` event.

## Root cause

When the daemon stops a node it issues a `SoftKill`. On Windows that is implemented as `GenerateConsoleCtrlEvent(CTRL_BREAK_EVENT, pid)` (`binaries/daemon/src/running_dataflow.rs`). A node that doesn't install its own console handler is then terminated by the OS default handler with `STATUS_CONTROL_C_EXIT`, which `ExitStatus::code()` surfaces as `-1073741510`.

The daemon's planned-stop detection in `handle_event` (the `SpawnedNodeResult` arm) only recognized the **Unix** shapes of a stop-induced exit:

```rust
NodeExitStatus::Signal(15) | NodeExitStatus::Signal(2)      // SIGTERM / SIGINT
| NodeExitStatus::ExitCode(143) | NodeExitStatus::ExitCode(130)  // wrapper (128 + signo)
```

There was no Windows equivalent, so a node the daemon *deliberately* stopped was misclassified as a failure, failing the whole dataflow and the CI job — even though shutdown happened exactly as requested. This is the direct Windows analog of the Unix `Signal(15)` / `143` handling added in #1882.

## Fix

- Extract the classification into a pure helper `is_sigterm_like_exit()` and add the Windows case `ExitCode(STATUS_CONTROL_C_EXIT)` (with a named, documented constant).
- The new arm only takes effect when the daemon actually initiated the stop — the call site is still gated on `grace_duration_kills` (populated solely by the daemon's own SoftKill/Kill path) plus `caused_by_node.is_none()` and `!finish_escalated`. A node that produces `0xC000013A` on its own, without a preceding daemon SoftKill, is still reported as a failure. `STATUS_CONTROL_C_EXIT` cannot occur as a Unix exit code, so the change is effectively Windows-only.
- Add unit tests for the predicate (recognized stop shapes, the new Windows code, and genuine failures incl. `Signal(9)`/SIGKILL which must keep flowing through the `GraceDuration` branch).

## Testing

- `cargo test -p dora-daemon planned_stop_exit` — 3 new tests pass.
- `cargo test -p dora-daemon fault_tolerance_tests` — 32 existing tests pass (surrounding stop logic unaffected).
- The change is behavior-preserving for all previously-handled cases (the extracted helper is identical to the old inline `matches!` plus the one new Windows arm).

I don't have a Windows runner in this environment to reproduce the nightly job end-to-end, so the fix is verified via the unit tests and static analysis of the exit-code path rather than a live CI repro.

Addresses the Windows shutdown-exit-code portion of #2425.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RMR3VBfwmc6evC5Rr3ZVjm)_


phil-opp:

No issues found. The refactor into `is_sigterm_like_exit` is behavior-preserving for the existing Unix cases, and the new Windows `STATUS_CONTROL_C_EXIT` arm is correct: `0xC000013A as u32 as i32 == -1073741510`, which the added test pins down. The predicate remains gated on `grace_duration_kill` at the call site, so a node that emits this code without a preceding SoftKill is still reported as a failure. Tests cover the positive shapes and the important negatives (SIGKILL/9, exit 1, Success/Unknown).

---

PR #2629, opened 2026-07-14
Update yanked spin dependency

phil-opp:


phil-opp:
Merging manually to unblock CI of other PRs.

PR #2643, opened 2026-07-14
ci: stop cache thrashing (gate rust-cache to main, kill sccache flood)

phil-opp:
The repo Actions cache sat at ~12 GB / 2,660 entries against GitHub's
10 GB cap, so GitHub continuously LRU-evicted the shared main-branch
workspace caches that check/clippy/test restore from — every PR started
cold and recompiled the whole dependency graph.

- Gate every Swatinem/rust-cache save in ci.yml and test-c-cpp to
  `main` (save-if). PRs restore the shared cache but no longer write
  per-PR ~800 MB / ~1.2 GB copies that churn the budget and evict each
  other. Matches the pattern already used in pip-release/release.
- Disable the sccache backend in pip-release.yml. Its per-object GitHub
  cache entries were 2,648 of the 2,660 caches (~4 GB) and give the PR
  loop no benefit.
- Add cache-cleanup.yml to drop a PR's caches when it closes.

### DIFF
--- a/.github/workflows/cache-cleanup.yml
+++ b/.github/workflows/cache-cleanup.yml
@@ -1,0 +1,44 @@
+# Delete a PR's Actions caches when the PR closes.
+#
+# GitHub's Actions cache is capped at 10 GB per repo and LRU-evicts once
+# full. Caches created on a PR branch (`refs/pull/<n>/merge`) are scoped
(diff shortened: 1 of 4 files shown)


phil-opp:
Merging manually

PR #2666, opened 2026-07-14
fix(node): add zenoh data-plane readiness barrier to stop startup message loss

phil-opp:
Since #1787 node outputs are published directly node-to-node over zenoh, bypassing the daemon. The daemon's "all nodes ready" barrier only synchronizes the control plane: it guarantees every subscriber is *declared* before any node runs, but not that each subscription has *propagated across zenoh to the producer's session*. A fast producer — e.g. a `dora/timer`-driven source — could start publishing before that route was established, and zenoh drops samples for a not-yet-wired subscriber. Those early messages were silently lost, which made the record/replay and validated-pipeline contract tests flaky (`sink got only 0 messages`; the recorder captured 10 messages instead of 20).

Each subscriber now declares a zenoh liveliness token once its data subscriber is wired (dynamic nodes excluded — they connect at arbitrary times). At startup, after the daemon barrier and before handing control to user code, each producer eagerly declares its output publishers and waits (bounded) until every expected subscriber for an output is present — counted from the descriptor via `DescriptorExt::output_subscriber_count` — and zenoh reports a matching subscription. Only then is that output on the direct zenoh path, so the first `send_output` is guaranteed to hit an established route. The wait runs after the barrier, so all tokens are already declared and it only waits out route propagation (normally milliseconds); it cannot deadlock, since a producer only ever waits on its subscribers, never on another producer's wait.

Fail-safe: an output whose subscribers don't connect within the timeout (an unreachable remote subscriber, or a zenoh data plane that can't connect while the control plane can) starts on the reliable daemon path and upgrades to direct zenoh on a later send once its subscribers appear — the dataflow is never blocked from starting and nothing is dropped.

The readiness liveliness key hex-encodes the output name so a producer's wildcard subscription cannot collide across nested output names (an output `cmd` must not count tokens of a distinct output `cmd/vel`, which would otherwise let `cmd` switch to the direct path before all of its own subscribers were wired).

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -83,6 +83,12 @@ pub struct EventStream {
     /// Per-input `@schema` subscribers (zenoh-ext AdvancedSubscriber) that prime
     /// the data subscribers' decoders. Kept alive like `_zenoh_subscribers`.
     _zenoh_schema_subscribers: Vec<zenoh_ext::AdvancedSubscriber<()>>,
+    /// Per-input liveliness tokens announcing "my data subscriber is declared"
+    /// so producers can count subscribers and only switch an output to the direct
+    /// zenoh data plane once every subscriber is wired (startup no-loss barrier;
+    /// see [`dora_core::topics::zenoh_input_ready_liveliness_topic`]). Kept alive
+    /// for the EventStream's lifetime; dropping a token undeclares it.
+    _zenoh_liveliness_tokens: Vec<zenoh::liveliness::LivelinessToken>,
     close_channel: DaemonChannel,
     clock: Arc<uhlc::HLC>,
     scheduler: Scheduler,
@@ -119,6 +125,7 @@ impl EventStream {
         clock: Arc<uhlc::HLC>,
         write_events_to: Option<PathBuf>,
         zenoh_session: Option<&zenoh::Session>,
+        dynamic: bool,
     ) -> eyre::Result<Self> {
         let channel = match daemon_communication {
             DaemonCommunicationWrapper::Standard(daemon_communication) => {
@@ -263,6 +270,7 @@ impl EventStream {
             total_queue_capacity,
             zenoh_session,
             &input_config,
+            dynamic,
         )
     }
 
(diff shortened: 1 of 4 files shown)


phil-opp:

Reviewed the full diff at `5395d8b`. **No issues found.** The readiness barrier looks correct on the points that matter for message loss:

- **Count/token symmetry (the dangerous direction is covered).** The descriptor-derived `expected` count (`output_subscriber_count`) and the subscriber-declared liveliness tokens are computed the same way — both count non-dynamic `User`-mapping subscriber links to the output, and both exclude dynamic nodes. So `expected` can't *under*count the tokens that will legitimately appear, which is the only direction that could switch to the direct zenoh path too early and drop startup samples. Over-counting only keeps the output on the lossless daemon path.
- **The startup race is actually closed.** The liveliness token is declared immediately after the data subscriber on the same session, and zenoh propagates declarations FIFO per link, so a token becoming visible at the producer implies the subscription route is already established. The extra `publisher.matching_status()` gate is belt-and-suspenders, and `history(true)` replays tokens from early subscribers.
- **Every failure path is fail-safe.** No zenoh session, invalid key, poisoned mutex, failed subscriber declaration, and unresolved remote tokens all keep the output on the lossless daemon path rather than silently dropping.
- **Teardown ordering is preserved.** The new liveliness tokens and readiness subscriber are dropped before the session under the existing `teardown_with_timeout` deadline, consistent with the documented publisher/session drop order (the #2425 wedged-runtime guard).

I couldn't find a correctness or message-loss concern in the diff.

---

phil-opp:
I found the following issue:

**Duplicate `#[allow(clippy::too_many_arguments)]` on `EventStream::init` will fail the clippy `-D warnings` gate.**

In `apis/rust/node/src/event_stream/mod.rs`, `init` already carried an `#[allow(clippy::too_many_arguments)]` (the line just above `#[tracing::instrument(...)]`). This diff adds a *second*, identical one right below the `instrument` attribute:

```rust
    #[allow(clippy::too_many_arguments)]
    #[tracing::instrument(level = "trace", skip(clock, zenoh_session))]
    #[allow(clippy::too_many_arguments)]   // <- added here, now duplicated
    pub(crate) fn init(
```

Two identical attributes on the same item trip clippy's `duplicated_attributes` lint (warn-by-default, `suspicious` group). CI runs `cargo clippy --all -- -D warnings`, so this is promoted to an error and fails the clippy job. Dropping the newly-added duplicate (keeping the pre-existing one) resolves it.

The readiness-barrier logic itself looks sound on the message-loss front — this is purely the lint regression introduced by the added attribute.

---
🤖 _Automated review by Claude — no human reviewed this before posting. May contain mistakes._

---

PR #2676, opened 2026-07-14
fix(node-api-cxx): correct pre-epoch timestamps in Metadata::set_timestamp

phil-opp:

## Issue

`Metadata::set_timestamp` in `apis/c++/node/src/lib.rs` splits an `i64` nanosecond value using truncating division:

```rust
let secs = value / 1_000_000_000;
let subsec_nanos = (value % 1_000_000_000) as u32;
```

For any **negative** `value` (an instant before the Unix epoch), `value % 1_000_000_000` is a *negative* remainder, and casting it to `u32` wraps it to a huge number. `DateTime::from_timestamp(secs, subsec_nanos)` then interprets `subsec_nanos` as an out-of-range nanosecond count and returns `None`, so:

- `set_timestamp(key, -500_000_000)` — a perfectly valid instant 0.5 s before the epoch — **fails** with "Invalid timestamp: out of range".
- Every pre-epoch instant is unrepresentable, and values in `-1_000_000_000 < value < 0` are silently corrupted.

This does not round-trip with `get_timestamp`, which reads the value back via `DateTime::timestamp_nanos_opt`.

## Fix

Use Euclidean division/remainder so the subsecond part stays in `0..1_000_000_000`, matching how `from_timestamp` interprets the `(secs, subsec_nanos)` pair:

```rust
let secs = value.div_euclid(1_000_000_000);
let subsec_nanos = value.rem_euclid(1_000_000_000) as u32;
```

For example, `-500_000_000` now becomes `secs = -1, subsec_nanos = 500_000_000` → `-1e9 + 0.5e9 = -0.5e9`, which round-trips back to `-500_000_000`.

## Validation

- Added `set_timestamp_roundtrips_negative_values`, asserting `set_timestamp` → `get_timestamp` round-trips for negative, epoch-boundary, and positive nanosecond values. The test fails on `main` and passes with this change.
- `cargo test -p dora-node-api-cxx`, `cargo clippy -p dora-node-api-cxx -- -D warnings`, and `cargo fmt --check` all pass.

---

### DIFF
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -770,9 +770,18 @@ impl Metadata {
     }
 
     pub fn set_timestamp(&mut self, key: &str, value: i64) -> EyreResult<()> {
-        // Convert nanoseconds since Unix epoch to chrono::DateTime<Utc>
-        let secs = value / 1_000_000_000;
-        let subsec_nanos = (value % 1_000_000_000) as u32;
+        // Convert nanoseconds since Unix epoch to chrono::DateTime<Utc>.
+        //
+        // Use Euclidean division so pre-epoch (negative) timestamps split
+        // correctly. Plain `/` and `%` truncate toward zero, yielding a
+        // *negative* remainder for negative inputs; casting that to `u32`
+        // wraps it to a huge value and makes `from_timestamp` reject the
+        // (otherwise valid) instant. `div_euclid`/`rem_euclid` keep the
+        // remainder in `0..1_000_000_000`, matching how `from_timestamp`
+        // interprets `(secs, subsec_nanos)` and round-tripping with
+        // `get_timestamp` for negative values.
+        let secs = value.div_euclid(1_000_000_000);
+        let subsec_nanos = value.rem_euclid(1_000_000_000) as u32;
 
         let dt = DateTime::from_timestamp(secs, subsec_nanos)
             .ok_or_else(|| eyre!("Invalid timestamp: out of range (nanos: {value})"))?;
@@ -1044,5 +1053,32 @@ mod tests {
 
         let result = event_as_input(event);
         assert!(result.is_err(), "expected Err for non-UInt8 input, got Ok");
+    }
+
+    /// Regression test: `set_timestamp` must accept and correctly represent
+    /// pre-epoch (negative) nanosecond timestamps. The previous truncating
+    /// `value % 1_000_000_000` produced a negative remainder that wrapped when
+    /// cast to `u32`, making `from_timestamp` reject valid instants. The value
+    /// must also round-trip through `get_timestamp`.
+    #[test]
+    fn set_timestamp_roundtrips_negative_values() {
+        for value in [
+            -1_i64,
+            -500_000_000,
+            -1_000_000_000,
+            -1_500_000_000,
+            -1_000_000_001,
+            0,
+            1,
+            1_500_000_000,
+        ] {
+            let mut meta = Metadata::empty();
+            meta.set_timestamp("ts", value)
+                .unwrap_or_else(|e| panic!("set_timestamp({value}) failed: {e}"));
+            let got = meta
+                .get_timestamp("ts")
+                .unwrap_or_else(|e| panic!("get_timestamp after set({value}) failed: {e}"));
+            assert_eq!(got, value, "timestamp {value} did not round-trip");
+        }
     }
 }


phil-opp:

This PR is one of a small batch opened by an automated (Claude / AI-agent) review pass over the codebase. Everything here is **machine-generated and reviewed for correctness**; please review before merging. Posting the run summary on this, the first PR of the batch.

### PRs opened, by category

**Correctness**
- **#2676** — `fix(node-api-cxx)`: `Metadata::set_timestamp` used truncating `%`, wrapping the negative remainder to `u32` and rejecting/corrupting every pre-epoch (negative) nanosecond timestamp. Switched to `div_euclid`/`rem_euclid`; added a `set_timestamp`→`get_timestamp` round-trip test.

**Soundness (undefined behavior)**
- **#2677** — `fix(ros2-bridge)`: `OwnedFFISeq::as_slice` / `RefFFISeq::as_slice` called `slice::from_raw_parts(self.data, 0)` on the **null** pointer `from_rust` stores for an empty sequence — UB even at length 0. Added the same `size == 0` guard that `FFISeq::deref` already has.

**Robustness / DoS-hardening**
- **#2678** — `fix(ros2-bridge-arrow)`: the single-`wstring` field deserializer pushed `u16` code units in an unbounded loop, unlike every sequence path in `sequence.rs` which caps at `MAX_SEQUENCE_ELEMENTS` (16M). Applied the same cap (sharing the constant).

### Validation
- Per-PR: `cargo fmt --check`, `cargo clippy -- -D warnings`, and the affected-crate tests pass (each PR adds a regression test).
- Combined diff: `/code-review` and `/simplify` gates run once over the union — no actionable findings; full-workspace `cargo test` run in progress.

### Findings examined but **rejected** (not turned into PRs)
- **c-node `(NULL, 0)` empty-payload idiom** — duplicate of already-open **#2598**.
- **ros2-bridge-node service client exits on a single response timeout** — not a clear bug: the action goal-send path and the service-response path both `bail!` by design, so this is a subjective fail-fast-vs-resilient choice, not an inconsistency.
- **replay-node: tiny `DORA_REPLAY_SPEED` → multi-century sleep** — an absurdly small speed legitimately means "replay absurdly slowly"; the huge sleep is directionally correct and only reachable via an operator-supplied pathological value.
- **ros2-bridge-node `oldest_insert` staleness** — effectively benign (only causes extra harmless eviction scans; never wrong eviction).
- **`read_json_value_as_arrow` panics on a 0-column schema** — requires a caller-supplied empty schema; not reachable on realistic node data paths.
- **`OwnedFFISeq::Drop` calls `Vec::from_raw_parts(null, 0, 0)`** (same UB class as #2677) — pre-existing, benign in practice, and a fix would alter FFI-adjacent `Drop` semantics; intentionally left out of #2677's minimal scope.

<sub>Generated with [Claude Code](https://claude.com/claude-code)</sub>

---

phil-opp:
No issues found. Switching `Metadata::set_timestamp` from truncating `/` and `%` to `value.div_euclid(1_000_000_000)` / `value.rem_euclid(1_000_000_000)` is the correct fix: `rem_euclid` keeps the sub-second part in `0..1_000_000_000` for negative inputs too, so `DateTime::from_timestamp(secs, subsec_nanos)` no longer receives a wrapped/negative nanosecond value and no longer corrupts or rejects pre-epoch instants. For any `value`, `secs * 1_000_000_000 + subsec_nanos == value`, so it round-trips through `get_timestamp`, which the added test confirms across negative, epoch, and positive values.

---
🤖 _Automated review by Claude — no human reviewed this before posting. May contain mistakes._

---

phil-opp:
The `ci/circleci: test-linux` failure was not from this change — it was a stale base. Diagnosis:

- This PR's only delta from its base is the single `apis/c++/node/src/lib.rs` file; it compiles and its test (`set_timestamp_roundtrips_negative_values`) passes locally. A one-crate change can't flip another crate's test under `cargo test --all`, so the failure lived in the base, not the diff.
- The branch was sitting on `07ed3be`, which predates **#2740 (`fix(core): don't fail ipv6_endpoints_are_bracketed on IPv6-less hosts`)** — a `dora-core` test (in `test-linux` scope) that fails on runners without IPv6. That matches the symptom exactly.
- Main has also since removed the CircleCI config (**#2743**), so the CircleCI signal itself is deprecated.

Rebased onto latest `main` (`acc2209`), which picks up the #2740 fix and drops the dead CircleCI config. Verified on the new base: `ipv6_endpoints_are_bracketed` passes, the cxx `set_timestamp` test passes, and `fmt` + `clippy -D warnings` are clean. The diff is unchanged (still +39/−3 in the one file).

---

phil-opp:
## 🤖 Automated review by Claude

_Fully automated review — no human has verified this. It may be incomplete or mistaken; treat it as advisory only._

Re-reviewed after the rebase onto `main` (`acc2209`). The code diff is unchanged from the earlier review — still just the `div_euclid`/`rem_euclid` fix in `apis/c++/node/src/lib.rs::set_timestamp` plus the negative-value round-trip test. The rebase only moves the base; it introduces no new logic. The fix remains correct: `rem_euclid(1_000_000_000)` keeps `subsec_nanos` in `0..1_000_000_000` for negative inputs, and `secs * 1e9 + subsec_nanos == value` holds for all `i64`, so pre-epoch timestamps round-trip through `get_timestamp`. Latest commit looks safe to merge; no new issues.


---

PR #2677, opened 2026-07-14
fix(ros2-bridge): avoid null-pointer slice UB in Owned/RefFFISeq::as_slice

phil-opp:

## Issue

For an **empty** sequence, `FFIFromRust::from_rust` deliberately stores a **null** `data` pointer (the `vec.is_empty()` branches set `data: std::ptr::null_mut()`, as the rosidl C ABI requires). Two code paths in `libraries/extensions/ros2-bridge/src/_core/sequence.rs` then use that null pointer in ways that are undefined behavior:

1. **`OwnedFFISeq::as_slice` and `RefFFISeq::as_slice`** build a slice unconditionally:

   ```rust
   pub fn as_slice(&self) -> &[T] {
       unsafe { std::slice::from_raw_parts(self.data, self.len()) }
   }
   ```

   `std::slice::from_raw_parts` requires a non-null, aligned pointer *even for a zero length*, so passing the null `data` pointer is UB. The sibling `FFISeq::deref` in the same file already guards exactly this case (`if self.size == 0 { return &[]; }`); the two `as_slice` accessors simply omitted the same guard.

2. **`OwnedFFISeq::drop`** reconstructs a `Vec` unconditionally:

   ```rust
   fn drop(&mut self) {
       unsafe { Vec::from_raw_parts(self.data, self.size, self.capacity) };
   }
   ```

   `Vec::from_raw_parts(null, 0, 0)` feeds the null pointer to `Unique::new_unchecked`, which aborts under debug assertions (and is UB in release). This is the same empty-sequence null-pointer path, and `OwnedFFISeq` is what msg-gen emits for every non-primitive sequence field.

## Fix

- Add the `size == 0` early-return to both `OwnedFFISeq::as_slice` and `RefFFISeq::as_slice`, mirroring `FFISeq::deref`.
- Add a `data.is_null()` early-return to `OwnedFFISeq::drop`, so an empty (null) sequence is not reconstructed.

Both are minimal soundness fixes on the same empty-sequence path. Neither changes the FFI representation — `from_rust` still stores a null pointer for the empty case.

## Validation

- Added regression tests covering the empty and non-empty paths of both `as_slice` and `Drop`. `RefFFISeq` is used for the empty `as_slice` case because it borrows (no `Drop`) and its `from_rust` is unconstrained in `T`, so the null-data representation can be built safely; the `OwnedFFISeq` empty representation is built directly to also exercise the `Drop` guard.
- `cargo test -p dora-ros2-bridge --lib sequence`, `cargo clippy -p dora-ros2-bridge --lib`, and `cargo fmt --check` all pass.

---

### DIFF
--- a/libraries/extensions/ros2-bridge/src/_core/sequence.rs
+++ b/libraries/extensions/ros2-bridge/src/_core/sequence.rs
@@ -90,6 +90,12 @@ pub struct OwnedFFISeq<T> {
 impl<T> OwnedFFISeq<T> {
     /// Extracts a slice.
     pub fn as_slice(&self) -> &[T] {
+        // `from_rust` stores a null `data` pointer for an empty sequence, and
+        // `slice::from_raw_parts` is UB with a null pointer even for length 0.
+        // Guard the empty case, mirroring `FFISeq::deref`.
+        if self.size == 0 {
+            return &[];
+        }
         unsafe { std::slice::from_raw_parts(self.data, self.len()) }
     }
 
@@ -152,6 +158,12 @@ pub struct RefFFISeq<T> {
 impl<T> RefFFISeq<T> {
     /// Extracts a slice.
     pub fn as_slice(&self) -> &[T] {
+        // `from_rust` stores a null `data` pointer for an empty sequence, and
+        // `slice::from_raw_parts` is UB with a null pointer even for length 0.
+        // Guard the empty case, mirroring `FFISeq::deref`.
+        if self.size == 0 {
+            return &[];
+        }
         unsafe { std::slice::from_raw_parts(self.data, self.len()) }
     }
 
@@ -183,5 +195,52 @@ impl<T> FFIFromRust for RefFFISeq<T> {
                 capacity: vec.len(),
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // `from_rust` deliberately stores a null `data` pointer for an empty
+    // sequence (see the `vec.is_empty()` branches). `slice::from_raw_parts`
+    // is undefined behavior when given a null pointer even for a zero
+    // length, so `as_slice` must special-case the empty sequence — exactly
+    // as `FFISeq::deref` already does. These tests exercise the empty and
+    // non-empty paths of that guard.
+    //
+    // `RefFFISeq` is used because it borrows (no `Drop`) and its `from_rust`
+    // is unconstrained in `T`, so the empty case can be built safely.
+
+    #[test]
+    fn ref_seq_as_slice_empty_is_safe() {
+        let empty: Vec<i32> = Vec::new();
+        let seq = unsafe { RefFFISeq::from_rust(&empty) };
+        assert!(seq.is_empty());
+        assert_eq!(seq.as_slice(), &[] as &[i32]);
+    }
+
+    #[test]
+    fn ref_seq_as_slice_non_empty_roundtrips() {
+        let data = vec![1_i32, 2, 3];
+        let seq = unsafe { RefFFISeq::from_rust(&data) };
+        assert_eq!(seq.as_slice(), data.as_slice());
+    }
+
+    #[test]
+    fn owned_seq_as_slice_empty_is_safe() {
+        // Build the empty representation `from_rust` produces (null data,
+        // size 0) directly, since `OwnedFFISeq::from_rust` requires
+        // `T: FFIFromRust`. `forget` avoids the unrelated `Drop` path
+        // (`Vec::from_raw_parts` on the null pointer); this test only
+        // covers the `as_slice` empty-guard.
+        let seq: OwnedFFISeq<i32> = OwnedFFISeq {
+            data: std::ptr::null_mut(),
+            size: 0,
+            capacity: 0,
+        };
+        assert!(seq.is_empty());
+        assert_eq!(seq.as_slice(), &[] as &[i32]);
+        std::mem::forget(seq);
     }
 }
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Adding the `if self.size == 0 { return &[]; }` guard to `OwnedFFISeq::as_slice` and `RefFFISeq::as_slice` correctly avoids the `slice::from_raw_parts` null-pointer UB that `from_rust` produces for empty sequences, matching the guard already present in `FFISeq::deref`. The regression tests cover the empty and non-empty paths, and the `OwnedFFISeq` empty test appropriately uses `mem::forget` to avoid the unrelated `Drop` path.

---

phil-opp:
## The fix is right, but it leaves the same UB live in `Drop`

The `as_slice` guard is correct, and returning `&[]` rather than `NonNull::dangling()` is the right call — `from_rust` must keep storing NULL for the empty case because rosidl expects NULL for empty sequences. The tests are genuine: `slice::from_raw_parts(null, 0)` does trip a debug precondition and hard-abort, so they fail pre-fix.

But the same root cause is live 30 lines down, at `libraries/extensions/ros2-bridge/src/_core/sequence.rs:137-141`:

```rust
impl<T> Drop for OwnedFFISeq<T> {
    fn drop(&mut self) {
        unsafe { Vec::from_raw_parts(self.data, self.size, self.capacity) };
    }
}
```

`Vec::from_raw_parts(null, 0, 0)` -> `Unique::new_unchecked(null)` -> UB. Reproduced standalone: with this PR applied, an empty `OwnedFFISeq` built by the real `from_rust` path and dropped normally still hard-aborts on a plain debug `cargo test`:

```
unsafe precondition(s) violated: NonNull::new_unchecked requires that the pointer is non-null
signal: 6, SIGABRT
```

Same `from_rust` empty branch (`sequence.rs:113-119`), same file, and **more** reachable than the `as_slice` case — `OwnedFFISeq` is what msg-gen emits for every non-primitive sequence (`msg-gen/src/types/sequences.rs:58`), so any empty `nav_msgs/Path.poses`-style field hits it.

The PR sidesteps it with `std::mem::forget` and the comment calls it "the unrelated `Drop` path". It is not unrelated, and that comment will mislead the next reader. One-line fix — guard `Drop` on `self.data.is_null()` / `capacity == 0` — and it belongs in this PR.

Checked and fine: `_core/string.rs:96,185` already guard with `is_empty()`. Pre-existing and out of scope: `RefFFISeq<T>` has no lifetime parameter, so safe `as_slice` can hand out a slice outliving the borrowed `Vec`.

Aside: `scripts/qa/all.sh:319` skips miri ("no miri-runnable unsafe target"). `_core/sequence.rs` is exactly the FFI-free, allocation-only unsafe miri covers — pointing the nightly gate at `dora-ros2-bridge --lib _core` would have caught both this bug and the `Drop` one.

phil-opp:
You're right on all three counts — fixed in b71b47c.

- **`Drop` guard added.** `OwnedFFISeq::drop` now returns early when `self.data.is_null()`, mirroring the `as_slice` empty-case guard, so `Vec::from_raw_parts(null, 0, 0)` no longer reaches `Unique::new_unchecked`. You're right that this is the more reachable path — msg-gen emits `OwnedFFISeq` for every non-primitive sequence field, so a normal drop of an empty `nav_msgs/Path.poses`-style field hits it.
- **Misleading comment removed.** Dropped the "unrelated `Drop` path" wording and the `mem::forget` workaround from the empty `as_slice` test — it now drops normally and exercises the guard. Added `owned_seq_empty_drop_is_safe` (guarded null path) and `owned_seq_non_empty_drop_frees` (confirms the guard only skips the null case and a real allocation is still reconstructed/freed). All 5 tests green under a plain debug `cargo test`.

On the two side notes: I've left the `RefFFISeq<T>` missing-lifetime issue out of scope for this PR as you flagged — it's a distinct API-signature change worth its own PR. And the miri aside is a good one: `_core/sequence.rs` is exactly the FFI-free, allocation-only unsafe miri covers, and pointing the nightly gate at `dora-ros2-bridge --lib _core` would have caught both this and the `as_slice` bug. I didn't touch `scripts/qa/all.sh` here to keep this PR to the fix, but happy to open a follow-up wiring that in if you'd like.

---

phil-opp:

The `Drop` guard added in b71b47c is correct: returning early when `self.data.is_null()` mirrors the `as_slice` `size == 0` guard and stops `Vec::from_raw_parts(null, 0, 0)` from reaching `Unique::new_unchecked`. The two added drop tests cover the guarded-empty and still-frees-non-empty paths. The commit since my last review doesn't introduce new issues — looks safe to merge from my side.

---

phil-opp:
The `ci/circleci: contract-tests` failure here is unrelated to this change — it's the flaky Python-dataflow smoke suite (`contract_*` in `example-smoke.rs`, which the tests themselves document as timing-flaky under CI load), and it was running on a stale base. This change only touches `_core/sequence.rs` (ROS2 FFI), which has no path to the Python dataflow tests; the same suite passed for the sibling #2678 on that base.

Rebased both commits onto latest `main` (`00abfb3`), same as #2676 — this drops the now-removed CircleCI config (#2743) so the real gating CI runs instead. No conflicts (no main commit since the base touched `_core/sequence.rs`), diff unchanged. Re-verified on the new base: all 5 `_core::sequence` tests pass (including the two new `Drop` tests), `fmt` + `clippy -D warnings` clean.

---

phil-opp:

The latest commit (`00abfb3`) rebases the two existing commits onto current `main`; the diff is unchanged from the previous review — the `as_slice` `size == 0` guard plus the `Drop` `self.data.is_null()` guard. Both correctly avoid the empty-sequence null-pointer UB (`slice::from_raw_parts(null, 0)` and `Vec::from_raw_parts(null, 0, 0)` → `Unique::new_unchecked(null)`) while still reconstructing and freeing non-empty sequences, and the `_core::sequence` tests exercise the empty/non-empty slice and drop paths. No new issues; the rebase looks safe to merge.

---

PR #2678, opened 2026-07-14
fix(ros2-bridge-arrow): cap element count when deserializing a wstring field

phil-opp:

## Issue

Every sequence path in `libraries/extensions/ros2-bridge/arrow/src/deserialize/sequence.rs` bounds its element count at `MAX_SEQUENCE_ELEMENTS` (16M) to stop a malformed CDR length prefix from driving an unbounded allocation — e.g. the bool, string, wstring-in-sequence, struct, and primitive paths all count elements and bail past the cap.

The single-`wstring` field deserializer in `deserialize/string.rs` (`WStringVisitor::visit_seq`) was the one path missing that guard:

```rust
let mut u16_buf: Vec<u16> = Vec::new();
while let Some(value) = seq.next_element::<u16>()? {
    u16_buf.push(value);   // unbounded
}
```

So a single `wstring`/`wstring<=N>` field could grow `u16_buf` without the cap that every neighbouring collection path enforces — a defense-in-depth inconsistency against untrusted ROS2 wire data.

## Fix

Apply the same element-count cap in `WStringVisitor::visit_seq`, sharing the existing constant (promoted from module-private to `pub(super)`) so the two paths cannot diverge. A wstring past the cap is rejected with a clear error instead of allocating without bound.

## Validation

- Added regression tests: a normal round-trip (`"hi"`) and an oversized case driven by a lazy iterator, so the guard fires without allocating past the cap.
- `cargo test -p dora-ros2-bridge-arrow --lib deserialize::string`, `cargo clippy -p dora-ros2-bridge-arrow --lib --tests -- -D warnings`, and `cargo fmt --check` all pass.

---

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/deserialize/sequence.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/deserialize/sequence.rs
@@ -17,7 +17,7 @@ use super::{StructDeserializer, error};
 /// Maximum number of elements allowed in a deserialized sequence.
 /// This prevents unbounded allocations from malformed CDR packets.
 /// 16 million elements is ~128 MB for f64 or ~16 MB for u8.
-const MAX_SEQUENCE_ELEMENTS: usize = 16_000_000;
+pub(super) const MAX_SEQUENCE_ELEMENTS: usize = 16_000_000;
 
 pub struct SequenceDeserializer<'a> {
     pub item_type: &'a NestableType,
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. `WStringVisitor::visit_seq` now caps its `u16` accumulation at `MAX_SEQUENCE_ELEMENTS` (checked with `>=` before each push, so the buffer never exceeds the cap), mirroring the existing guard in `sequence.rs` and sharing the now-`pub(super)` constant so the two paths can't diverge. The lazy `U16Seq` test correctly exercises the oversized rejection without materializing past the cap.

---

Issue #2496, opened 2026-07-07
`dora topic hz`: unchecked `Instant - Duration` panics on large `--window` (same bug class #2378 fixed in `dora topic info`)


phil-opp:
## Summary

`HzStats::record` in `dora topic hz` computes a sliding-window cutoff with an **unchecked** `Instant - Duration` subtraction. When the window (in seconds) exceeds the monotonic clock's current value, this panics with *"overflow when subtracting duration from instant"*, crashing the command.

This is the **exact bug class** that PR #2378 (`a198a95`) already fixed in the sibling `dora topic info` command — but the same fix was never applied to `dora topic hz`.

## Affected code

`binaries/cli/src/command/topic/hz.rs:185-196`:

```rust
fn record(&self, now: Instant) {
    let mut timestamps = self.timestamps.lock().unwrap_or_else(|e| e.into_inner());
    timestamps.push_back(now);
    let cutoff = now - self.window_duration;   // <-- line 188: panics on underflow
    while let Some(&first) = timestamps.front() {
        if first < cutoff {
            timestamps.pop_front();
        } else {
            break;
        }
    }
}
```

`window_duration` comes straight from the user-supplied `--window` flag with no upper bound:

```rust
// hz.rs:50-56 — parse_window only rejects 0, accepts any usize
fn parse_window(s: &str) -> Result<usize, String> {
    let val: usize = s.parse().map_err(|e| format!("{e}"))?;
    if val == 0 {
        return Err("window must be at least 1".to_string());
    }
    Ok(val)
}
// hz.rs:65 — #[clap(long, default_value_t = 10, value_parser = parse_window)]
// hz.rs:178-183 — HzStats::new: window_duration = Duration::from_secs(window_secs as u64)
```

On Linux `Instant` is `CLOCK_MONOTONIC` (time since boot). `now - Duration::from_secs(window)` therefore underflows and panics whenever `window` (seconds) is larger than the machine's uptime. The panic fires inside `record`, i.e. on the **first** received message (`timestamps` is empty, so it panics before the drain loop even runs).

## How to reproduce

Guaranteed on any host — pick a window larger than the machine's uptime in seconds:

```bash
# non-interactive path — panics the CLI process
dora topic hz -d my-dataflow some_node/some_output --duration 5 --window 999999999
```

The first message received triggers:

```
thread 'main' panicked at 'overflow when subtracting duration from instant'
```

[...]

PR #2378, opened 2026-06-26
fix(cli): compute dora topic info bandwidth over actual collection window

phil-opp:
> 🤖 This PR is **machine-generated by Claude** (automated repository review run). Please review carefully before merging.

## Issue

`dora topic info` (`binaries/cli/src/command/topic/info.rs`) collects stats in a thread that **breaks early** when the publisher closes (`InterDaemonEvent::OutputClosed => break`), but always divided the byte total by the full configured `--duration`:

```rust
let bandwidth_mbps = (total_bytes as f64 * 8.0) / (duration_secs as f64 * 1_000_000.0);
```

So a topic that stopped publishing partway through reported a bandwidth that was too low — e.g. ~10× low for a topic that closed after 1 s of a 10 s window. The trailing `else` branch was also dead: `duration` is parsed with `value_parser(range(1..))`, so `duration_secs > 0` is always true.

## Fix

Divide by the **actual** elapsed collection time (`start_time.elapsed()` captured after the thread joins), and extract the computation into a small pure `bandwidth_mbps` helper that returns `None` (rendered `<unknown>`) when no messages were observed or the window is zero. (The `hz` figure already uses inter-message intervals, so it was unaffected.)

## Validation

- New unit tests `bandwidth_uses_actual_elapsed_window` and `bandwidth_unknown_without_messages_or_time`.
- `cargo test -p dora-cli --lib bandwidth` — 2 passed.
- `cargo clippy -p dora-cli -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BbJBzWqUQmwFTy7gsmYmpg)_

PR #2682, opened 2026-07-15
fix(cli): avoid Instant underflow panic on 'r' key in `dora inspect top`

### LINKED ISSUE #2681: `dora inspect top`: unchecked `Instant - Duration` on 'r' keypress panics with a large `--refresh-interval` (same bug class as #2378 / #2496)
> ⚠️ **This issue was created by an automated, scheduled Claude code-review check.** The panic path was traced by hand against the current `main`; please confirm before acting on it.

## Summary

`run_app` in `dora inspect top` handles the **`r` (force refresh)** key by subtracting the user-supplied refresh interval from `Instant::now()` with an **unchecked** `Instant - Duration`:

```rust
// binaries/cli/src/command/inspect/top.rs:328-331
KeyCode::Char('r') => {
    // Force refresh by resetting last_update
    last_update = Instant::now() - refresh_duration;   // <-- panics on underflow
}
```

When `refresh_duration` exceeds the current monotonic-clock value, this panics with *"overflow when subtracting duration from instant"*, crashing the TUI **with the terminal left in raw mode / alternate screen** (see Impact).

This is the **exact bug class** already fixed in `dora topic info` (PR #2378) and filed for `dora topic hz` (#2496). The same hazard was never addressed in `dora inspect top`.

## Why it's reachable

`refresh_interval` is parsed with a lower bound only — **no upper bound**:

```rust
// top.rs:41-42
#[clap(long, value_name = "SECONDS", default_value_t = 2,
       value_parser = clap::value_parser!(u64).range(1..))]
pub refresh_interval: u64,
```

```rust
// top.rs:68
let refresh_duration = Duration::from_secs(self.refresh_interval);
```

On Linux `Instant` is `CLOCK_MONOTONIC` (time since boot). `Instant::now() - refresh_duration` therefore underflows whenever `refresh_interval` (seconds) is larger than the machine's uptime. Two realistic triggers:

1. **Any host, unbounded flag:** `dora inspect top --refresh-interval 999999999` then press `r` → guaranteed panic.
2. **Low-uptime host, ordinary flag:** e.g. `--refresh-interval 3600` (refresh hourly) on a freshly-started container / CI runner / VM with uptime < 1 h, then press `r` → panic. Fresh containers routinely have single- or double-digit-second uptimes.

Note the loop's *other* subtraction is already guarded — line 301-303 uses `checked_sub(...).unwrap_or(...)` — so the author was aware of the hazard; line 330 is simply an inconsistent raw subtraction that slipped through.

## Impact

The panic fires on the **main thread** inside `run_app`. Because it unwinds *past* the terminal-restore block in `Executable::execute` (top.rs:75-82, which runs only after `run_app` returns), the restore never happens:

```rust
let res = run_app(&mut terminal, ...);   // panic unwinds here
// unreachable on panic:
disable_raw_mode()?;
execute!(terminal.backend_mut(), LeaveAlternateScreen, DisableMouseCapture)?;
terminal.show_cursor()?;
```

So the user is dropped back to a **corrupted terminal** (raw mode still on, alternate screen still active, cursor hidden) in addition to the crash — strictly worse than a clean error.

## Suggested fix

The intent of the `r` handler is only *"force the next loop iteration to refresh now"*. Rather than back-date `last_update` (which is what causes the underflow), use an explicit flag so no `Instant` arithmetic is needed:

```rust
let mut force_refresh = false;
// ...
KeyCode::Char('r') => {
    force_refresh = true;
}
// ...
if force_refresh || last_update.elapsed() >= refresh_duration {
    let reply = send_control_request(&session, &ControlRequest::GetNodeInfo)?;
    node_infos = expect_reply!(reply, NodeInfoList(infos))?;
    app.update_stats(node_infos.clone());
    last_update = Instant::now();
    force_refresh = false;
}
```

(A minimal alternative — `Instant::now().checked_sub(refresh_duration)` — is unsafe here because on the `None` path you'd fall back to "now", which *skips* the very refresh the key is supposed to force. The flag is the clean fix.)

A regression test analogous to `info.rs`'s `calculate_hz_does_not_panic_on_oversized_window` — exercising the `r` path with `refresh_interval` larger than any plausible uptime — would lock this in.

## Scope note

I grepped every `Instant::now() - …` in the tree: all other occurrences are in `#[cfg(test)]` code with fixed constant durations (`binaries/daemon/src/lib.rs`, `binaries/coordinator/src/lib.rs`, `binaries/daemon/src/running_dataflow.rs`). `binaries/cli/src/command/inspect/top.rs:330` is the only remaining **production** path with a user-controlled duration, and it is not covered by #2378 or #2496.

---
*Confidence: high — the flag parsing (no upper bound), the `refresh_duration` construction, and the raw subtraction on the documented `r` keypath were all read directly from the cited lines on current `main`.*

phil-opp:
## Summary

Fixes #2681.

The `r` (force refresh) key handler in `dora inspect top` back-dated `last_update` with an **unchecked** `Instant::now() - refresh_duration`:

```rust
KeyCode::Char('r') => {
    // Force refresh by resetting last_update
    last_update = Instant::now() - refresh_duration;   // <-- panics on underflow
}
```

`refresh_interval` is parsed with a lower bound only (no upper bound), so when `refresh_duration` exceeds the monotonic clock value this underflows and panics with *"overflow when subtracting duration from instant"*. Because the panic unwinds past the terminal-restore block, the user is dropped back into a corrupted terminal (raw mode + alternate screen still active).

Two realistic triggers (both from the issue):
1. `dora inspect top --refresh-interval 999999999`, then press `r`.
2. `--refresh-interval 3600` on a freshly-started container / CI runner / VM with uptime < 1h, then press `r`.

This is the same bug class already fixed for `dora topic info` (#2378) and filed for `dora topic hz` (#2496); the loop's *other* subtraction was already guarded with `checked_sub`.

## Fix

Replace the back-dating with an explicit `force_refresh` flag so the refresh decision uses only `Duration` comparisons and can never underflow. The decision is extracted into a small `should_refresh(force_refresh, since_last_update, refresh_duration)` helper.

## Testing

- Added unit tests `force_refresh_triggers_regardless_of_elapsed` (asserts a forced refresh fires with `Duration::from_secs(u64::MAX)` and zero elapsed — the previous panic scenario) and `refresh_waits_for_interval_when_not_forced`.
- `cargo test -p dora-cli --lib` — passing

---

### DIFF
--- a/binaries/cli/src/command/inspect/top.rs
+++ b/binaries/cli/src/command/inspect/top.rs
@@ -284,6 +284,12 @@ fn run_app<B: Backend>(
 ) -> eyre::Result<()> {
     let mut app = App::new();
     let mut last_update = Instant::now();
+    // Set by the `r` key to force the next loop iteration to refresh. Using an
+    // explicit flag avoids back-dating `last_update` with an unchecked
+    // `Instant - Duration`, which panics on underflow when `refresh_duration`
+    // exceeds the monotonic clock (e.g. a large `--refresh-interval` on a
+    // low-uptime host).
+    let mut force_refresh = false;
 
     // Reuse coordinator connection
     let session = connect_to_coordinator(coordinator_addr)?;
@@ -326,15 +332,15 @@ fn run_app<B: Backend>(
                     app.toggle_sort(SortColumn::Memory);
                 }
                 KeyCode::Char('r') => {
-                    // Force refresh by resetting last_update
-                    last_update = Instant::now() - refresh_duration;
+                    // Force the next iteration to refresh immediately.
+                    force_refresh = true;
                 }
                 _ => {}
             }
         }
 
-        // Update data if refresh interval has passed
-        if last_update.elapsed() >= refresh_duration {
+        // Update data if a refresh was forced or the refresh interval elapsed.
+        if should_refresh(force_refresh, last_update.elapsed(), refresh_duration) {
             // Query node info every refresh interval to get updated metrics
             let reply = send_control_request(&session, &ControlRequest::GetNodeInfo)?;
             node_infos = expect_reply!(reply, NodeInfoList(infos))?;
@@ -342,8 +348,23 @@ fn run_app<B: Backend>(
             // Update stats with current node info
             app.update_stats(node_infos.clone());
             last_update = Instant::now();
+            force_refresh = false;
         }
     }
+}
+
+/// Decides whether node stats should be refreshed this loop iteration.
+///
+/// A refresh happens either because the user forced one (`r` key) or because
+/// `refresh_duration` has elapsed since the last update. This deliberately uses
+/// only `Duration` comparisons (no `Instant - Duration`) so it can never panic
+/// on underflow, regardless of how large `refresh_duration` is.
+fn should_refresh(
+    force_refresh: bool,
+    since_last_update: Duration,
+    refresh_duration: Duration,
+) -> bool {
+    force_refresh || since_last_update >= refresh_duration
 }
 
 fn ui(f: &mut Frame, app: &mut App, refresh_duration: Duration) {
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Replacing the back-dated `last_update = Instant::now() - refresh_duration` with an explicit `force_refresh` flag and the `should_refresh` helper removes the `Instant - Duration` underflow panic, since the refresh decision now uses only `Duration` comparisons. The `force_refresh` flag is correctly reset after a refresh, and the unit tests cover the forced path (with `Duration::from_secs(u64::MAX)`) and the interval-elapsed boundaries.

---

PR #2683, opened 2026-07-15
test(runtime): guard the #2483 compact() fix with an integration-level test

### LINKED ISSUE #2680: test(runtime): #2483 regression test does not guard the regression (compact() call site survives mutation)
## Summary

_This issue was filed by an automated, scheduled Claude code-review check. Please verify before acting._

PR #2483 (`fix(runtime): compact dropped operator inputs to bound queue memory`, commit a6dbc20) fixes a real unbounded-memory leak in `binaries/runtime/src/operator/channel.rs`: when a downstream operator stalls, `drop_oldest_inputs` converts over-cap inputs to `None` tombstones (`*event = None`) that are only ever cleared by `pop_front` in `send_next_queued` — which never runs while the outgoing rendezvous send is pending. The fix adds `compact()` (`queue.retain(Option::is_some)`) and calls it from `drop_oldest_inputs`:

```rust
if dropped > 0 {
    self.compact();   // <-- the actual fix
    tracing::warn!("dropped {dropped} operator inputs because event queue was too full");
}
```

**The fix itself is correct.** The concern is the accompanying regression test.

## The problem

The added test `compact_removes_tombstones_preserving_order` exercises `compact()` **in isolation** — it manually assigns `buffer.queue` and calls `buffer.compact()` directly, never routing through `add_event` → `drop_oldest_inputs`:

```rust
let mut buffer = InputBuffer::new(BTreeMap::new());
buffer.queue = VecDeque::from([None, Some(...), None, None, Some(...), None]);
buffer.compact();
assert_eq!(buffer.queue.len(), 2, ...);
```

Because the test bypasses `drop_oldest_inputs` entirely, **deleting the `self.compact();` call at the fix site has no effect on the test — it still passes.** In mutation-testing terms this is a surviving mutant: the exact line that fixes the leak is unguarded, so the regression can be silently reintroduced (e.g. by a future refactor of `drop_oldest_inputs`) and CI stays green.

The PR description acknowledges the isolation as intentional (to avoid depending on the `Metadata`/`ArrowData` construction API), but the side effect is that the *regression path* — "a stalled consumer accumulates one tombstone per dropped input without bound" — is never actually exercised. The test verifies the primitive works, not that the leak is fixed.

This matches a class of issue the project has flagged before (#1669: "test claims to cover X but doesn't"), and it's exactly what `make qa-mutants` / `qa-mutation-audit` are meant to catch on this crate.

## Verification

By inspection: the test never calls `add_event`/`drop_oldest_inputs`, so removing line `self.compact();` in `drop_oldest_inputs` cannot change the assertion outcome. (I applied that one-line deletion locally to confirm the diff is isolated; a full `cargo test -p dora-runtime` compile was too slow to complete in the check window, but the survival follows directly from the test structure.)

## Affected code

- `binaries/runtime/src/operator/channel.rs` — `drop_oldest_inputs` (fix site) and `mod tests::compact_removes_tombstones_preserving_order` (test).

## Suggested fix

Add an integration-level test that drives the real path and asserts the queue length stays bounded, so the `compact()` call site is covered. `Event::Input` requires `Metadata`, but the drop logic keys only on `Event::Input { id, .. }`, so a minimal constructor helper (or reusing an existing test builder) is enough. Sketch:

```rust
#[test]
fn drop_oldest_inputs_keeps_deque_bounded_under_stall() {
    let mut caps = BTreeMap::new();
    caps.insert(DataId::from("x".to_string()), (2, QueuePolicy::Drop)); // effective cap
    let mut buffer = InputBuffer::new(caps);
    for _ in 0..1000 {
        buffer.add_event(input_event("x")); // helper building Event::Input
    }
    // Without compact() at the call site this grows to ~1000; with it, it stays ~cap.
    assert!(buffer.queue.len() <= SMALL_BOUND);
}
```

Keeping the existing unit test for the primitive is fine; the point is to add coverage that fails if the fix is removed.

Severity: low — the shipped behavior in #2483 is correct; this is a test-coverage / regression-protection gap, not a live bug.

phil-opp:
## Summary

Fixes #2680.

PR #2483 fixed a real unbounded-memory leak in `binaries/runtime/src/operator/channel.rs`: when a downstream operator stalls, `drop_oldest_inputs` converts over-cap inputs to `None` tombstones that are only cleared by `pop_front` in `send_next_queued` — which never runs while the outgoing rendezvous send is pending. The fix adds `compact()` (`queue.retain(Option::is_some)`) and calls it from `drop_oldest_inputs`.

The accompanying regression test `compact_removes_tombstones_preserving_order` only exercises `compact()` **in isolation** — it assigns `buffer.queue` directly and calls `compact()`, never routing through `add_event` → `drop_oldest_inputs`. As the issue notes, deleting the `self.compact();` call at the fix site therefore has **no effect** on that test — a surviving mutant on the exact line that bounds queue memory.

## Fix

Add `drop_oldest_inputs_keeps_deque_bounded_under_stall`, which drives the real path:
- feeds 50 inputs for a `DropOldest`-capped id (cap 2) via `add_event`, never draining the queue (models a stalled consumer — exactly when tombstones would otherwise accumulate);
- asserts the deque stays bounded to the cap and holds no `None` tombstones.

With `compact()` present the deque length is `cap`; without it, every drop past the cap leaves a `None` behind and the length grows roughly linearly with the input count.

## Verification

- `cargo test -p dora-runtime --lib operator::channel::tests` — both tests pass.
- Temporarily removed the `self.compact();` call and confirmed the new test **FAILS** (`0 passed; 1 failed`), then restored it. The previous isolated test stayed green under the same mutation.
- `cargo clippy -p dora-runtime` and `cargo fmt --all -- --check` — clean.

---

### DIFF
--- a/binaries/runtime/src/operator/channel.rs
+++ b/binaries/runtime/src/operator/channel.rs
@@ -160,10 +160,22 @@ impl InputBuffer {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow::array::new_empty_array;
+    use arrow::datatypes::DataType;
+    use dora_message::metadata::Metadata;
+    use dora_node_api::ArrowData;
 
     fn closed_event(id: &str) -> Event {
         Event::InputClosed {
             id: DataId::from(id.to_string()),
+        }
+    }
+
+    fn input_event(id: &str) -> Event {
+        Event::Input {
+            id: DataId::from(id.to_string()),
+            metadata: Metadata::new(dora_core::uhlc::HLC::default().new_timestamp()),
+            data: ArrowData(new_empty_array(&DataType::Null)),
         }
     }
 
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. This is a test-only change adding `drop_oldest_inputs_keeps_deque_bounded_under_stall`, which drives the real `add_event` → `drop_oldest_inputs` → `compact` path (modelling a stalled consumer by never draining the queue) and asserts the deque stays bounded to the cap with no `None` tombstones. It correctly guards the `#2483` fix site that the isolated `compact_removes_tombstones_preserving_order` test bypassed.

---

Issue #2021, opened 2026-06-03
bug(fault-tolerance): input_timeout deadlines never fire when zenoh drops backpressure messages (regression from #1787)


phil-opp:
## Summary

Since PR #1787 moved data routing off the daemon, `input_timeout` deadlines no longer fire when a receiver node is too slow to consume its event channel and zenoh silently drops messages. The daemon's `note_output_sent_to_local_receivers` function unconditionally updates `last_received` whenever an `OutputSent` notification arrives, even when zenoh has already discarded the payload.

## Root cause

**Before #1787** — data was routed through the daemon. In `send_output_to_local_receivers`, `last_received` was updated **only when `channel.try_send()` succeeded**:

```rust
// binaries/daemon/src/lib.rs  (old path)
match channel.try_send(Timestamped { inner: NodeEvent::Input { .. }, .. }) {
    Ok(()) => {
        // deadline updated only on successful delivery
        if let Some(deadline) = dataflow.input_deadlines.get_mut(&..) {
            deadline.last_received = Some(Instant::now());
        }
    }
    Err(TrySendError::Full(_)) => {
        // channel full → deadline NOT updated → watchdog can fire
    }
}
```

**After #1787** — nodes publish directly via zenoh. The source node calls `report_output_sent`, and the daemon calls `note_output_sent_to_local_receivers`, which updates `last_received` for **every receiver unconditionally**:

```rust
// binaries/daemon/src/lib.rs:4447  (new path)
fn note_output_sent_to_local_receivers(..) {
    for (receiver_id, input_id) in local_receivers {
        if let Some(deadline) = dataflow.input_deadlines.get_mut(&..) {
            deadline.last_received = Some(now);  // ← ALWAYS, even if zenoh dropped it
        }
        // …circuit-breaker recovery…
    }
}
```

Meanwhile, the zenoh subscriber callback uses `try_send`:

```rust
// apis/rust/node/src/event_stream/mod.rs  (callback, can't block zenoh IO worker)
if let Err(TrySendError::Full(_)) = tx_cb.try_send(EventItem::ZenohInput { .. }) {
    tracing::warn!("event channel full; dropping zenoh input");
    // ← no signal back to daemon; OutputSent was already sent
}
```

Because `OutputSent` reaches the daemon before (or regardless of) the callback's `try_send` result, the daemon always marks the deadline as satisfied. The `input_timeout` watchdog never fires for a slow-consuming node on the zenoh path.

## Impact

[...]

PR #2684, opened 2026-07-15
fix(daemon): gate circuit-breaker recovery on receiver backpressure

### LINKED ISSUE #2627: daemon: circuit-breaker recovery on `OutputSent` is not gated on receiver backpressure, so a broken input flaps recovered↔broken for a persistently-slow consumer
> ⚠️ **This issue was created by an automated scheduled Claude code-review check.** The finding was verified by reading the source directly (both code paths + the surrounding tests); please confirm before acting on it.

## Summary

In `note_output_sent_to_local_receivers` (the `OutputSent` handler), the **deadline refresh** is correctly gated on whether the receiver is keeping up (`receiver_keeping_up`), but the **circuit-breaker recovery** a few lines below it runs **unconditionally**. As a result, a bare `OutputSent` notification re-opens a broken input even when the receiver is still saturated and no data was actually delivered — defeating the circuit breaker for a persistently-slow consumer and causing the input to flap `broken ↔ recovered` roughly every `input_timeout`.

This is the exact anti-pattern the #2021 gating was written to prevent ("the bare `OutputSent` notification is *not* a delivery confirmation"). The fix was applied to the deadline-refresh path but the recovery path immediately below it was left ungated. The parallel recovery in `send_output_to_local_receivers` **is** correctly gated (it lives inside the `Ok(())` arm of `channel.try_send(..)`, i.e. only on real delivery), which makes the asymmetry stand out.

## Location

`binaries/daemon/src/lib.rs`, `note_output_sent_to_local_receivers`, lines ~5099–5138:

```rust
let receiver_keeping_up = dataflow
    .subscribe_channels
    .get(receiver_id)
    .is_some_and(|channel| channel.capacity() >= CONTROL_EVENT_HEADROOM);
if receiver_keeping_up
    && let Some(deadline) = dataflow
        .input_deadlines
        .get_mut(&(receiver_id.clone(), input_id.clone()))
{
    deadline.last_received = Some(now);          // <- gated on keeping-up (correct, #2021)
}

let Some(timeout) = dataflow
    .broken_inputs
    .remove(&(receiver_id.clone(), input_id.clone()))   // <- NOT gated on receiver_keeping_up
else {
    continue;
};
// ...re-inserts open_inputs, re-arms input_deadlines with last_received = Some(now),
//    bumps circuit_breaker_recoveries, and sends NodeEvent::InputRecovered...
```

Compare with the correctly-gated recovery in `send_output_to_local_receivers` (~line 5197–5234), which only recovers **inside** the successful `channel.try_send(..) => Ok(())` arm (which itself is only reached when `channel.capacity() >= CONTROL_EVENT_HEADROOM`).

## Concrete failing scenario

1. Node `R` has input `x` with an `input_timeout`; consumer `R` is slow and its daemon-side event channel saturates (`capacity() < CONTROL_EVENT_HEADROOM`).
2. `check_input_timeouts` fires → `x` is inserted into `broken_inputs` and `break_input` sends `InputClosed` (circuit breaker armed open).
3. Producer `P` keeps producing. Since #1787 the data payload goes directly over zenoh (the node-side callback `try_send`s and drops when the node channel is full), so no recovery happens through `send_output_to_local_receivers`. But `P` still emits `OutputSent`, which reaches `note_output_sent_to_local_receivers`.
4. There `receiver_keeping_up == false` (channel still saturated), so the deadline is **not** refreshed — but `broken_inputs.remove(x)` still succeeds: `x` is re-opened, the deadline is re-armed with `last_received = Some(now)`, `circuit_breaker_recoveries` is incremented, and `InputRecovered` is queued.
5. `R` is still overwhelmed, so after `timeout` elapses `check_input_timeouts` breaks `x` again, and the next `OutputSent` "recovers" it again — the input flaps `broken ↔ recovered` every ~`timeout` instead of staying open while the consumer is behind.

## Impact

- **Circuit breaker defeated for a slow consumer** — the mechanism that is supposed to stop hammering an overwhelmed/dead receiver keeps re-opening the input on a signal that is explicitly *not* a delivery confirmation.
- **Control-channel churn** — each re-break re-sends `InputClosed` and each phantom recovery queues `InputRecovered`, consuming the very `CONTROL_EVENT_HEADROOM` the breaker reserves to keep control events flowing.
- **Misleading fault-tolerance stats** — `FaultToleranceStats::circuit_breaker_recoveries` is inflated with phantom recoveries, so `dora top` / coordinator metrics report recoveries that never delivered anything.

## Suggested fix

Gate the recovery on `receiver_keeping_up` as well, mirroring the deadline refresh directly above it (and the delivery-gated recovery in `send_output_to_local_receivers`). Only recover a broken input when the receiver actually has headroom:

```rust
let Some(timeout) = (receiver_keeping_up)
    .then(|| dataflow
        .broken_inputs
        .remove(&(receiver_id.clone(), input_id.clone())))
    .flatten()
else {
    continue;
};
```

A regression test in the same `#[cfg(test)] mod tests` block — asserting that a saturated receiver (`receiver_keeping_up == false`) does **not** clear `broken_inputs` on `OutputSent`, mirroring the existing #2021 deadline-gating tests around lines ~6590–6730 — would lock this in.

## Confidence

Medium. The code asymmetry is unambiguous (deadline refresh gated, recovery not; the delivery-path recovery is gated on real delivery), and no test currently pins recovery-via-`OutputSent`, so this reads as an oversight rather than intent. It's conceivable the authors want recovery to be eager (producer resumed → re-open regardless of receiver state); if so, the deadline-refresh gating and this recovery path should be reconciled either way.

phil-opp:
## Summary

Fixes #2627.

In `note_output_sent_to_local_receivers` (the `OutputSent` handler), the **deadline refresh** is correctly gated on `receiver_keeping_up` (#2021), but the **circuit-breaker recovery** a few lines below ran **unconditionally**:

```rust
let receiver_keeping_up = dataflow
    .subscribe_channels
    .get(receiver_id)
    .is_some_and(|channel| channel.capacity() >= CONTROL_EVENT_HEADROOM);
if receiver_keeping_up && let Some(deadline) = ... {
    deadline.last_received = Some(now);      // gated (correct)
}

let Some(timeout) = dataflow
    .broken_inputs
    .remove(&(receiver_id.clone(), input_id.clone()))   // NOT gated
else { continue; };
// ...re-opens the input, re-arms the deadline, bumps circuit_breaker_recoveries,
//    and sends InputRecovered...
```

A bare `OutputSent` is **not** a delivery confirmation: since #1787 the data payload goes directly over zenoh and the node-side callback drops the input with `try_send` when its event channel is full. So for a persistently-slow consumer whose daemon-side channel is saturated, each `OutputSent` "recovered" the input even though nothing was delivered — the input then broke again on the next `input_timeout`, flapping `broken ↔ recovered` indefinitely and defeating the circuit breaker.

The parallel recovery in `send_output_to_local_receivers` **is** correctly gated (it lives inside the `Ok(())` arm of `channel.try_send(..)`), which makes the asymmetry stand out.

## Fix

Gate the recovery on the same `receiver_keeping_up` signal, leaving the `broken_inputs` entry in place until the receiver drains enough to keep up.

## Testing

Added two tests in `fault_tolerance_tests`:
- `output_sent_does_not_recover_broken_input_when_receiver_saturated` — a saturated receiver (channel `capacity() < CONTROL_EVENT_HEADROOM`) keeps the input broken: `broken_inputs` retained, `circuit_breaker_recoveries == 0`, no `InputRecovered` emitted.
- `output_sent_recovers_broken_input_when_receiver_keeping_up` — a receiver with headroom recovers as before.

Verified that removing the new gating makes the saturated test **FAIL** (`0 passed; 1 failed`), then restored it. `cargo clippy -p dora-daemon` and `cargo fmt --all -- --check` are clean.

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -5304,6 +5304,19 @@ fn note_output_sent_to_local_receivers(
             deadline.last_received = Some(now);
         }
 
+        // Circuit-breaker recovery must be gated on the same backpressure
+        // signal as the deadline refresh above. A bare `OutputSent` is not a
+        // delivery confirmation (see the comment above), so re-opening a broken
+        // input while the receiver is still saturated only makes it flap
+        // `broken ↔ recovered` every `input_timeout` without any data actually
+        // getting through (#2627). Leave the `broken_inputs` entry in place
+        // until the receiver drains enough to keep up — matching the recovery
+        // in `send_output_to_local_receivers`, which only fires inside the
+        // successful `try_send` arm.
+        if !receiver_keeping_up {
+            continue;
+        }
+
         let Some(timeout) = dataflow
             .broken_inputs
             .remove(&(receiver_id.clone(), input_id.clone()))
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The fix gates the circuit-breaker recovery in `note_output_sent_to_local_receivers` on the same `receiver_keeping_up` backpressure signal already used for the deadline refresh, via an early `continue` placed before the `broken_inputs.remove(...)`. The `continue` only skips the recovery logic for a saturated receiver (leaving the `broken_inputs` entry in place), which matches the recovery gating in `send_output_to_local_receivers`, and the two added tests cover both the saturated (stays broken, no `InputRecovered`, counter 0) and keeping-up (recovers, counter 1) cases.

---

Issue #2131, opened 2026-06-10
coordinator: stale PendingRestart leaks reply_sender and hangs caller if a daemon dies mid-restart


phil-opp:
## Background

PR #2085 (merged) reworked dataflow restart into a two-phase, event-driven flow to fix the cross-daemon Zenoh race in #2082:

- **Phase 1** (`initiate_restart`) sends `StopDataflow` to all daemons, persists `Stopping`, and parks a `PendingRestart` (descriptor + name + uv + the caller's `reply_sender`) in `pending_restarts`.
- **Phase 2** completes inside the `DataflowFinishedOnDaemon` handler when the *last* daemon reports finish, calling `start_dataflow` and firing the parked reply.

The core fix is correct. This issue tracks the follow-up gaps that PR #2085 left behind (one of them already flagged as an inline TODO).

## Problem 1 — caller hangs forever if a daemon dies mid-restart (primary)

If a daemon crashes/disconnects after receiving `StopDataflow` but before sending `DataflowFinishedOnDaemon`, the `PendingRestart` entry is never removed and its `reply_sender` never fires. The restart caller (CLI) blocks indefinitely with no error.

This is a behavioral regression relative to the old synchronous `restart_dataflow`, which surfaced stop/start failures immediately. The two-phase flow trades a visible error for a silent hang.

The PR documents this as an inline TODO in `initiate_restart`:

```rust
// TODO: If a daemon crashes after receiving StopDataflow but before
// sending DataflowFinishedOnDaemon, this entry (and its reply_sender)
// will never be cleaned up, causing the restart caller to hang
// indefinitely. A timeout or a daemon-disconnect cleanup path should
// be added to evict stale PendingRestart entries.
```

**Proposed fix:** evict stale `PendingRestart` entries and reply `Err` when their restart can no longer complete — either via a timeout, or (preferred) by draining matching `pending_restarts` entries in the daemon-disconnect cleanup path so the caller gets a clear error.

## Problem 2 — double-restart guard runs after `StopDataflow` (minor)

In `initiate_restart`, the `pending_restarts.contains_key(...)` guard is checked *after* `stop_dataflow` has already broadcast `StopDataflow` to every daemon. A duplicate `Restart` for an already-restarting dataflow therefore incurs a redundant stop round-trip before being rejected. The guard should be moved ahead of the `stop_dataflow` call (right after extracting the descriptor) so duplicates are rejected before any daemon I/O.

[...]

PR #2685, opened 2026-07-15
test(coordinator): cover pending-restart cleanup on daemon disconnect (#2131)

phil-opp:
## Summary

Refs #2131.

While investigating #2131, I found that **Problem 1** (caller hangs forever if a daemon dies mid-restart) and **Problem 2** (double-restart guard running after `StopDataflow`) are already fixed on `main`:

- `cleanup_disconnected_daemons_from_running_dataflows` now drains matching `pending_restarts` entries and replies `Err` when a daemon disconnects (`binaries/coordinator/src/lib.rs`), resolving the inline TODO that Problem 1 quoted.
- `initiate_restart` now checks the `pending_restarts.contains_key(...)` duplicate guard **first**, before extracting the descriptor or calling `stop_dataflow`.

What was still missing is **Problem 3** for the Problem-1 cleanup path: the existing tests cover the duplicate-restart guard (`initiate_restart_rejects_duplicate_request`) and orphan reclaim (`disconnect_reclaims_orphaned_running_dataflow`), but nothing asserts that a disconnect **drains a parked restart and unblocks its caller with an error**. Without coverage, that cleanup could silently regress back into the original hang.

## Change

Adds `disconnect_drains_pending_restart_and_replies_err`:
- parks a `PendingRestart` (with the caller's reply channel) for a running dataflow;
- runs `cleanup_disconnected_daemons_from_running_dataflows` with the dataflow's daemon in the disconnected set;
- asserts `pending_restarts` is drained and the caller receives an `Err` mentioning "daemon disconnected while restart was pending" (rather than being left hanging).

Test-only change; no production behavior is modified.

## Verification

- `cargo test -p dora-coordinator --lib disconnect_drains_pending_restart_and_replies_err` — passes.
- Temporarily removed the drain loop and confirmed the test **FAILS** (`0 passed; 1 failed`), then restored it.
- `cargo clippy -p dora-coordinator` and `cargo fmt --all -- --check` — clean.

Once merged, this closes the remaining Problem 3 gap; the primary fixes for Problems 1 & 2 are already on `main`, so #2131 can be closed.

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -5017,6 +5017,66 @@ mod tests {
         );
     }
 
+    // Regression test for #2131 (Problem 1): if a daemon disconnects while a
+    // two-phase restart is parked in `pending_restarts`, that daemon will never
+    // send `DataflowFinishedOnDaemon`, so the deferred restart can never
+    // complete. The disconnect cleanup must evict the stale entry and reply
+    // `Err` so the restart caller (CLI) gets a clear error instead of hanging
+    // forever. This locks in the cleanup path added for the inline TODO.
+    #[test]
+    fn disconnect_drains_pending_restart_and_replies_err() {
+        let dataflow_id = DataflowId::from(Uuid::new_v4());
+        let daemon_id = DaemonId::new(Some("gone".to_string()));
+        let node_id: dora_core::config::NodeId = "sender".to_string().into();
+
+        let mut df = test_running_dataflow(dataflow_id, daemon_id.clone(), node_id);
+        df.spawn_result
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. This is a test-only change adding `disconnect_drains_pending_restart_and_replies_err`, which parks a `PendingRestart`, runs `cleanup_disconnected_daemons_from_running_dataflows` with the daemon disconnected, and asserts the stale entry is drained and the caller receives an error rather than hanging. The setup and assertions match the two-phase-restart cleanup path it locks in.

---

phil-opp:
Already resolved by #2085

PR #2688, opened 2026-07-15
ci: stop nightly from evicting PR caches + stabilize cache key

phil-opp:
Follow-up to the cache-thrashing fix. Two remaining cache-miss sources:

1. nightly.yml saved ~15-25 GB of Windows/macOS/examples/cli-tests
   caches per run (five of them 1-2.7 GB each) into the shared 10 GB
   repo budget, LRU-evicting the Linux PR-CI caches every night. Gate
   all 23 rust-cache steps with `save-if: false` (restore-only) —
   nightly is a non-blocking daily job, so it trades a slower nightly
   for a warm PR loop.

2. rust-cache folds every installed rustup toolchain into its cache
   key, so the runner image's preinstalled stable (e.g. 1.97.0) sitting
   next to our pinned 1.92.0 flips the key on every runner-image Rust
   bump and misses all caches at once. Add a "Pin toolchain set" step to
   each cached ci.yml job that uninstalls all but RUST_VERSION before
   rust-cache runs. test-c-cpp is intentionally left on `stable`, so its
   key still tracks the runner image by design.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,14 @@
 # cache steps gated the same way, and do NOT enable the sccache backend
 # in any workflow (it writes one cache entry per object and floods the
 # same 10 GB budget — see pip-release.yml).
+#
+# Cache-key stability: rust-cache folds *every* installed rustup
+# toolchain into its cache key. The GitHub runner image ships a default
+# stable toolchain (e.g. 1.97.0) alongside our pinned RUST_VERSION, so
+# when GitHub bumps the image's Rust the key silently changes and every
+# cache misses at once. Each cached job therefore runs a "Pin toolchain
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. The "Pin toolchain set" step sets `$RUST_VERSION` as the default before uninstalling toolchains, and since `grep -vF "$RUST_VERSION"` filters the pinned toolchain out of the uninstall list, the default is never removed; the `|| true` guards keep the step from failing when the filter matches nothing. The `save-if: false` additions are restore-only and merge cleanly into the existing `with:` blocks (including the one that also carries `shared-key: workspace-test`), so they don't disturb which jobs write the shared cache. This is CI-only YAML/bash with no runtime code path affected.

---

PR #2691, opened 2026-07-15
ci: drop full debuginfo in workspace-build jobs to avoid disk exhaustion

phil-opp:
The Test/E2E/contract-tests jobs build the whole workspace in the
dev/test profile with default full debug info. Measured on a workspace
artifact, debug info is ~78% of each binary (781 MB -> 175 MB after
`strip --strip-debug`), so `cargo build` + `cargo test` overrun the
runner disk even after the existing "Free disk space" step, failing
with "No space left on device".

Set CARGO_PROFILE_DEV_DEBUG / CARGO_PROFILE_TEST_DEBUG to
`line-tables-only` on those three jobs. This keeps file:line in test
backtraces while dropping the bulk of target/. The env is applied on
both main (cache save) and PRs (restore) so the shared workspace-test
rust-cache fingerprint stays coherent, and identically on `test` and
`contract-tests`, which reuse the same cache key.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,7 +159,12 @@ jobs:
         run: |
           set -eux
           df -h
-          sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt/ghc /opt/hostedtoolcache/CodeQL || true
+          # None of these toolchains are used by any dora CI job. Not the
+          # whole /opt/hostedtoolcache — contract-tests' setup-python reads
+          # /opt/hostedtoolcache/Python (only the CodeQL bundle is removed).
+          sudo rm -rf /usr/local/lib/android /usr/share/dotnet /opt/ghc \
+            /usr/local/.ghcup /usr/local/share/boost /usr/lib/jvm \
+            /opt/hostedtoolcache/CodeQL || true
           docker system prune -af || true
           df -h
       - name: Build (excluding Python packages)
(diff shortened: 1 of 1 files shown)


phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. May contain mistakes.

I reviewed this CI change (debug-info reduction + expanded disk cleanup) and found no blocking issues. What I checked:

- **Cache-key coherence.** `Swatinem/rust-cache` folds `CARGO_*` env vars into its cache key, so setting `CARGO_PROFILE_DEV_DEBUG`/`CARGO_PROFILE_TEST_DEBUG` changes the fingerprint. The only jobs sharing a key are `test` and `contract-tests` (both `shared-key: workspace-test`), and both set the two vars identically, so they stay coherent. `e2e` sets them but uses its own default-keyed cache; `check` (`workspace-check`), `clippy`, `bench`, and `check-license` don't set them and don't share `workspace-test`, so nothing is cross-contaminated.
- **`line-tables-only`** is a valid value for both the `dev` and `test` profiles on the pinned toolchain (1.92), and it leaves the `release` profile untouched — so the `e2e` job's `cargo build -p dora-cli --release` is unaffected, as the comment claims.
- **Disk cleanup.** The expanded `rm -rf` (ghcup, boost, JVM) targets toolchains no dora CI job uses, and it deliberately keeps `/opt/hostedtoolcache/Python`, which `contract-tests`' `setup-python` depends on. `|| true` keeps a missing path from failing the step.

The first `main` run after merge will rebuild cold once (new cache key) and re-save, which is expected rather than a regression. No correctness, coherence, or CI-breakage concern from my read — no issues found. · fully automated review, no human reviewed this before posting.

---

PR #2705, opened 2026-07-15
fix(daemon): send OutputClosed for remote-only outputs on node finish

phil-opp:
## Summary

When a node finishes, the daemon's `handle_outputs_done` derived the set of
closed outputs solely from `dataflow.mappings`, which only records outputs that
have a **local** consumer. Outputs consumed exclusively by a node on another
daemon live in `open_external_mappings` and were therefore omitted, so
`send_output_closed_events` never emitted the `InterDaemonEvent::OutputClosed`
for them.

## Issue

Consider a distributed dataflow where daemon **A** runs a source node `sensor`
whose output `frame` is consumed **only** by node `proc` on daemon **B**
(nothing on A subscribes to `sensor/frame`):

- `sensor/frame` is registered in B's `mappings` (local consumer `proc`) and in
  A's `open_external_mappings` (remote consumer), but **not** in A's `mappings`.
- When `sensor` finishes, A calls `handle_outputs_done`, which collects
  `outputs` from `mappings` only → `sensor/frame` is missing.
- `send_output_closed_events` filters the remote-close list by
  `outputs.contains(&output_id.1)`, so it never sends `OutputClosed` to B.
- On B, the `OutputClosed` handler that would `close_input` for `proc` never
  fires. `proc`'s input stays open forever: it never receives
  `InputClosed`/`AllInputsClosed`, never drains cleanly, and is ultimately
  SIGKILLed by the finish-straggler watchdog after the grace period instead of
  shutting down gracefully.

The explicit `CloseOutputs` path is unaffected — there the node supplies its own
output list — so only the implicit finish path (`OutputsDone`) was broken.

## Fix

Add `RunningDataflow::node_output_ids`, which unions a node's outputs across both
`mappings` and `open_external_mappings`, and use it in `handle_outputs_done`.

## Validation

- New unit tests on the pure helper (`node_output_ids_unions_local_and_remote_outputs`,
  `node_output_ids_ignores_other_producers`).
- `cargo test -p dora-daemon`, `cargo clippy -p dora-daemon -- -D warnings`, and
  `cargo fmt --check` all pass.
- The change is `pub(crate)`-internal (no public API change). CI will exercise
  the full multi-daemon E2E suite.

---
carefully before merging.

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4035,13 +4035,12 @@ impl Daemon {
             .get_mut(&dataflow_id)
             .ok_or_else(|| eyre!("no running dataflow with ID `{dataflow_id}`"))?;
 
-        let outputs = dataflow
-            .mappings
-            .keys()
-            .filter(|m| &m.0 == node_id)
-            .map(|m| &m.1)
-            .cloned()
-            .collect();
+        // Include outputs consumed only by nodes on other daemons
+        // (`open_external_mappings`), not just those with a local consumer
+        // (`mappings`). Otherwise a remote-only output never triggers an
+        // `OutputClosed` event when the producing node finishes, leaving the
+        // remote consumer's input open until it is force-killed.
+        let outputs = dataflow.node_output_ids(node_id).into_iter().collect();
 
         if might_restart {
             self.logger
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -592,6 +592,18 @@ impl RunningDataflow {
         self.open_inputs.get(node_id).unwrap_or(&self.empty_set)
     }
 
+    /// All output ids produced by `node_id`, whether they are consumed by a
+    /// local node (recorded in `mappings`) or only by nodes on another daemon
+    /// (recorded in `open_external_mappings`).
+    ///
+    /// Deriving the set from `mappings` alone misses outputs that have no local
+    /// consumer, so their remote consumers would never receive the
+    /// `OutputClosed` event when the producing node finishes (dora-rs/dora#2152
+    /// region — graceful cross-daemon shutdown).
+    pub(crate) fn node_output_ids(&self, node_id: &NodeId) -> BTreeSet<DataId> {
+        node_output_ids(&self.mappings, &self.open_external_mappings, node_id)
+    }
+
     /// Nodes blocking an otherwise-finished dataflow (dora-rs/dora#2152).
     ///
     /// Returns nodes that should be force-stopped because the dataflow is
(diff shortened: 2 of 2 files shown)


phil-opp:

This is the summary for a batch of PRs opened by an automated code-review pass
over the repository (machine-generated by Claude). Each PR is independent and
branches off the same freshly-fetched `origin/main`.

### PRs opened, by category

**Correctness**
- #2705 — `fix(daemon)`: send `OutputClosed` for remote-only outputs on node finish (fixes a cross-daemon graceful-shutdown hang).
- #2706 — `fix(coordinator)`: don't let a disconnected daemon's frozen ack block state-log pruning (unbounded log growth).
- #2708 — `fix(core)`: reject module `params` keys that collide case-insensitively (silent param drop via `PARAM_<KEY>`).

**Optimization**
- #2707 — `perf(runtime)`: reuse the already-held GIL token in the operator `on_event` path (drops two redundant `Python::attach` re-entries per event).

**Documentation / Simplification** — none met the bar this run.

### Validation
Per-branch: `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test`
for each affected crate all pass, with new unit tests for the three correctness
fixes. The combined diff passed the `/review` (high-effort) and `/simplify`
gates with no findings. All four changes are `pub(crate)`-internal with no
public API change. (Note: a full-workspace `cargo test` run was constrained by
the session's disk allowance; one unrelated, timing-sensitive zenoh loopback
test — `dora-node-api::zenoh_schema_cache`, untouched by any of these diffs —
flaked under that pressure. CI will run the full suite.)

### Findings considered but not turned into PRs

_Duplicates of already-open automated PRs:_
- Download fallback filename leaking the URL query/fragment — already covered by an open PR (derive filename from URL path).
- Trace-context codec corrupting `tracestate` values containing `:` — already covered by an open PR (robust W3C tracestate codec).
- C++ `Metadata::set_timestamp` mishandling pre-epoch negative nanoseconds — already covered by an open PR.
- Python `next()`/`recv_async()` panicking on negative/non-finite `timeout` — already covered by an open PR (reject non-finite/negative timeouts).
- `OperatorEvent::AllocateOutputSample` dead variant — already covered by an open PR.

_Rejected on merit:_
- `runtime` `drop_oldest_inputs` allocating a map and rescanning the whole queue per event — real inefficiency, but the incremental-count refactor risks the tombstone/compaction invariants for only a modest gain on already-bounded queues. Not worth the risk.
- `coordinator` `reload_dataflow` applying partially across daemons on mid-loop failure — plausible but medium-confidence and the same class as an existing partial-apply fix; deferred rather than shipped half-verified.
- `retrieve_logs` dead `else` branch / undocumented `run_operator`/`StopReason` — too minor to justify a PR.
- mavlink `SET_MODE.base_mode` stored as `UInt8` vs. the "all enums are `UInt32`" module doc — doc/consistency only, no runtime impact.

---
🤖 Machine-generated. Please review each PR carefully before merging.

---

phil-opp:
No issues found.

I traced the finish path: `send_output_closed_events` (lib.rs:3931-3936) only emits `InterDaemonEvent::OutputClosed` for entries of `open_external_mappings` whose output id appears in the `outputs` argument. Before this change `handle_outputs_done` built `outputs` from `mappings.keys()` alone (local consumers only), so a remote-only output was never in the list and its `OutputClosed` was dropped — matching the described cross-daemon hang. Unioning `mappings` and `open_external_mappings` via `node_output_ids` is correct, and it does not perturb the local-input-closing branch (lib.rs:3920-3929), since remote-only output ids aren't present in `mappings`. The pure helper and its two unit tests cover the union and the other-producer-filtering behavior.

---

---

PR #2706, opened 2026-07-15
fix(coordinator): don't let a disconnected daemon block state-log pruning

phil-opp:
## Summary

`RunningDataflow::prune_state_log` computed `min_ack` over **every** entry in
`daemon_ack_sequence`, including daemons that have permanently disconnected. A
disconnected daemon's stale ack pinned `min_ack` forever, preventing the
state-replication log from ever being pruned.

## Issue

- On disconnect, the cleanup loop removes a daemon from `self.daemons` but
  **retains** its `daemon_ack_sequence` entry (deliberately, for a possible
  reclaim). The ack handler also refuses to advance the ack of any daemon no
  longer in `daemons`, so that entry is frozen at its last value.
- `prune_state_log` took the `min` over all entries, so the frozen ack pins
  `min_ack` at the disconnected daemon's last acknowledged sequence.

**Failure scenario:** a multi-daemon dataflow runs on A and B (B last acked
sequence 5). B dies permanently while A keeps mutating replicated state (params,
add/remove nodes). B's stale `daemon_ack_sequence[B] = 5` pins `min_ack = 5`
forever, so `state_log.retain(|e| e.sequence > 5)` can never prune past
sequence 5. The log grows without bound until it hits the
`MAX_STATE_LOG_ENTRIES` (10 000) hard cap, which drops the oldest entries and
forces a full param replay for **all** daemons — including the healthy A — on
their next status report.

## Fix

Gate pruning on the acks of **live** daemons only (those still in
`self.daemons`). If a disconnected daemon reconnects after its missed entries
were pruned, `state_log_delta` already returns `None` and the caller falls back
to a full param replay for it.

## Validation

- New unit test `state_log_prune_ignores_disconnected_daemon_ack`; existing
  `state_log_prune_removes_acked_entries` / `state_log_delta_returns_none_when_pruned`
  still pass.
- `cargo test -p dora-coordinator`, `cargo clippy -p dora-coordinator -- -D warnings`,
  and `cargo fmt --check` all pass. The change is `pub(crate)`-internal.

---
carefully before merging.

---


phil-opp:
No issues found.

I checked the pruning change in `prune_state_log`: gating `min_ack` on `self.daemons` (live daemons) instead of every entry in `daemon_ack_sequence` is correct. A live daemon with no ack entry maps to `0` via `unwrap_or(0)`, so pruning is still blocked while any live daemon is behind — no premature prune. The retained frozen ack of a disconnected daemon no longer pins `min_ack`. The reconnect path is consistent: a reconnecting daemon is re-added to `df.daemons` before catch-up (lib.rs:2608), and if the log was pruned past its ack, `state_log_delta` returns `None` and the code falls back to a full param replay (lib.rs:2639-2656). The added unit test exercises exactly this scenario.

---

---

phil-opp:
🤖 **Automated review by Claude** (fully automated, no human in the loop).

Re-reviewed after the new commits since my last review (the pruned-to-empty reconnect full-replay handling in `state_log_delta`, plus the extended prune test). No new issues found — the prune gate on live daemons never over-prunes (a live daemon that is behind, or has no ack entry, still blocks pruning via `unwrap_or(0)`), and the pruned-to-empty reconnect case correctly falls back to a full param replay. The added test fails if either fix is reverted. Latest commits look safe to merge.

---

PR #2708, opened 2026-07-15
fix(core): reject module param keys that collide case-insensitively

phil-opp:
## Summary

Module `params` keys that differ only in case (e.g. `mode` and `Mode`) both map
to the same `PARAM_<KEY>` env var, so one silently overwrote the other. This PR
rejects such collisions with a clear error.

## Issue

`expand_module_node` injects each module param into the inner node's env as
`PARAM_<KEY>` with the key upper-cased (`substitute_params_in_node`). The key
validation only enforced `[A-Za-z0-9_]`, so `mode`, `Mode`, and `MODE` are all
valid, distinct param keys that all collapse to `PARAM_MODE`. Because the env
map is built by inserting in `BTreeMap` iteration order, only one survives — the
rest are silently dropped, with the winner determined by ASCII sort order.

```yaml
params:
  mode: safe
  Mode: turbo
```

injects only `PARAM_MODE` = `safe` (`Mode` sorts before `mode`, then `mode`
overwrites), silently losing `turbo`.

## Fix

During param validation, track the upper-cased form of each key and `bail!` on a
collision, naming both keys and the colliding env var. A param can no longer be
silently dropped.

## Validation

- New unit test `reject_case_colliding_param_keys`; existing
  `reject_invalid_param_key` still passes.
- `cargo test -p dora-core`, `cargo clippy -p dora-core -- -D warnings`, and
  `cargo fmt --check` all pass.

---
carefully before merging.

---

### DIFF
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -352,6 +352,7 @@ fn expand_module_node(
         .collect();
 
     // Validate and collect params
+    let mut seen_upper: BTreeMap<String, &String> = BTreeMap::new();
     for key in node.params.keys() {
         if !key.chars().all(|c| c.is_ascii_alphanumeric() || c == '_') || key.is_empty() {
             bail!(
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found.

The collision check uses `key.to_uppercase()`, which matches the env-var construction in `substitute_params_in_node` exactly (`format!("PARAM_{}", key.to_uppercase())`, expand.rs:507), so the detection is precisely aligned with the real overwrite condition. Rejecting the collision at validation time is the right behavior versus the prior silent, iteration-order-dependent overwrite. The error message names both keys and the colliding env var. Test covers the `mode`/`Mode` case.

---

---

PR #2679, opened 2026-07-15
refactor(mavlink2-bridge): remove never-constructed BridgeError::Transport variant

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and opened via a maintainer's account only because the bot cannot author PRs directly (the commit itself is authored as `Claude <noreply@anthropic.com>`). It has **not** been hand-reviewed by a human yet — please treat it as a suggestion and review before merging.

## What

Removes the never-constructed `Transport` variant (and its `From<std::io::Error>` impl) from `BridgeError` in `libraries/extensions/mavlink2-bridge/src/error.rs`:

```rust
#[error("transport error: {0}")]
Transport(#[from] std::io::Error),
```

## Why

`BridgeError::Transport` has **no constructor anywhere in the workspace**, and its `#[from] std::io::Error` conversion is **never triggered** — there is no `?`-on-`io::Error` call site in a `BridgeResult` context.

The transport layer deliberately maps connection/parse failures to `BridgeError::Config` with added context rather than surfacing a bare `io::Error`, e.g.:

```rust
// transport.rs
.map_err(|e| BridgeError::Config(format!("failed to connect mavlink to '{addr}': {e}")))?;
```

So the `Transport` variant and its auto-derived `From` impl are dead code. The only other repo-wide match for the identifier is an unrelated doc-comment mentioning "Transport builders" (the `transport` module), not the variant.

The sibling variants stay live: `Arrow` (constructed via `#[from]` on `?` in `arrow_convert.rs`), `Config`, and `Mavlink`.

## Verification

- `cargo fmt -p dora-mavlink2-bridge -- --check` ✅
- `cargo clippy -p dora-mavlink2-bridge --all-targets -- -D warnings` ✅ (confirms no now-orphaned imports)
- `cargo test -p dora-mavlink2-bridge -p dora-mavlink2-bridge-node` ✅ (16 tests + doctest pass)

## Note

`dora-mavlink2-bridge` inherits the workspace version and is publishable (no `publish = false`), so removing a public error-enum variant is technically a semver-relevant change — worth catching in the next version bump. If the `Transport` variant is intentionally reserved as part of the public error surface (e.g. for a planned code path that surfaces raw I/O errors instead of wrapping them in `Config`), feel free to close — as it stands it is entirely unconstructed. This is in the same spirit as the in-flight #2626 (removing the never-constructed `NodeError::Connection` variant).

---
[...]

PR #2709, opened 2026-07-16
refactor(mavlink2-bridge): remove unused BridgeError::Transport variant

phil-opp:
> [!IMPORTANT]
> **🤖 Auto-generated by Claude (Claude Code).** This PR was created by an automated dead-code sweep and is opened under a maintainer's account only because the automation authenticates with that token — the change was written by Claude, **not** by the account owner. It has **not** been hand-reviewed by a human. Please review before merging.

## What

Removes the `Transport` variant from `BridgeError` in `dora-mavlink2-bridge` (`libraries/extensions/mavlink2-bridge/src/error.rs`):

```rust
#[error("transport error: {0}")]
Transport(#[from] std::io::Error),
```

## Why it's dead

The variant is **never constructed anywhere in the workspace**, and its auto-generated `From<std::io::Error>` impl is never triggered either:

- Nothing produces `BridgeError::Transport(..)` explicitly. A workspace-wide search for `Transport` in the bridge lib + node crate finds only the definition and an unrelated doc-comment ("Transport builders for TCP, UDP, and serial links").
- The `#[from] std::io::Error` conversion has no implicit call site. Every fallible path in the crate maps I/O failures to `BridgeError::Config` instead. In `transport.rs`, `open_mavlink` wraps `mavlink::connect(..)`'s error via `.map_err(|e| BridgeError::Config(..))`, and the other `connect_*` helpers only ever return `BridgeError::Config`. No function returning `BridgeResult` uses `?` on a `std::io::Error`, so the `From` impl is unreachable.

The live variants — `Arrow` (`#[from] ArrowError`), `Config`, and `Mavlink` — are all still constructed and are left untouched.

## Note

`dora-mavlink2-bridge` is a `1.0.0-rc1` crate without `publish = false`, so removing a `pub` enum variant is technically a pre-1.0 public-API change. It is an internal error-representation variant (not an external extension point), and pruning an impossible-to-construct error before the 1.0 freeze seems like the right time — but flagging it so the call is yours.

## Verification

- `cargo fmt -p dora-mavlink2-bridge -- --check` — clean
- `cargo clippy -p dora-mavlink2-bridge -p dora-mavlink2-bridge-node --all-targets -- -D warnings` — clean
- `cargo test -p dora-mavlink2-bridge -p dora-mavlink2-bridge-node` — all pass, including the TCP/UDP loopback transport tests that exercise the real connect path

---
_Generated by [Claude Code](https://claude.ai/code). Commit authored as `Claude <noreply@anthropic.com>`._

---
_Generated by [Claude Code](https://claude.ai/code/session_01X6EQTcEpBcx1pMHbEhD1kb)_

### DIFF
--- a/libraries/extensions/mavlink2-bridge/src/error.rs
+++ b/libraries/extensions/mavlink2-bridge/src/error.rs
@@ -5,9 +5,6 @@ pub enum BridgeError {
     #[error("Arrow error: {0}")]
     Arrow(#[from] arrow::error::ArrowError),
 
-    #[error("transport error: {0}")]
-    Transport(#[from] std::io::Error),
-
     #[error("config error: {0}")]
     Config(String),
 


phil-opp:
No issues found.

I grepped the crate for `Transport` and `io::Error`: the only occurrences are the variant definition itself and an unrelated doc-comment mention of "Transport builders". No code constructs `BridgeError::Transport(..)`, and no function returning `BridgeResult` applies `?` to a `std::io::Error`, so the generated `From<std::io::Error>` impl has no call site. Removing the variant is a safe pure deletion with no behavior change.

---

---

phil-opp:
Duplicate of #2679 — the diffs are byte-identical. Closing in favour of #2679.

Issue #2715, opened 2026-07-16
zenoh 1.9 silently disabled peer routing (linkstate removed), plus two config hazards


phil-opp:
Investigating startup message loss (#2666) surfaced several problems with our zenoh 1.9.0 config. They're independent but share a root: **zenoh 1.9 changed behavior under us and our config still describes the old world.**

## 1. `routing/peer: linkstate` is a silent no-op — we lost peer relaying

`libraries/core/src/topics.rs` sets:

```rust
// Linkstate make it possible to connect two daemons on different network through a public daemon
zenoh_config.insert_json5("routing/peer", r#"{ mode: "linkstate" }"#)
```

zenoh 1.9 **removed** this. From the [1.9.x "Longwang" release notes](https://zenoh.io/blog/2026-04-16-zenoh-longwang/): *"`routing.peer` is removed—all peers now operate in peer-to-peer mode."* In 1.8 the field was functional (`PeerRoutingConf`, with `linkstate_peer/` and `p2p_peer/` routing hats); in 1.9 both hats are gone and the config field only logs a deprecation warning.

Two consequences:

- **The failure is invisible to us.** `insert_json5` returns `Ok`, so our `warn!("failed to set zenoh routing/peer to linkstate")` branch never fires. The only signal is zenoh's own log line, on every session open:
  `routing.peer.mode` and `routing.peer.linkstate` are deprecated and have no effect`
- **`peer_to_peer` was already the default in 1.8.** Our linkstate line was a deliberate *opt-out* of that default. 1.9 silently reverted us *to* the thing we were avoiding. This is a behavior regression from the upgrade, not just log noise.

## 2. Peers no longer relay → routes can permanently fail to establish

zenoh's [deployment docs](https://zenoh.io/docs/getting-started/deployment/): *"In peer to peer regions, Zenoh nodes must be deployed with `peer` mode and must also be deployed in a clique topology where every node is connected to all other nodes of the region."* The 1.9 notes also remove the last loophole: *"Removed peer failover brokering."*

Relevant facts about our setup:

- **We have no router anywhere.** We never set `mode`, so every session (daemon, coordinator, node) defaults to `peer`.
- Node↔node links therefore depend entirely on gossip + autoconnect forming a clique opportunistically. That is asynchronous and best-effort.
- When that race is lost there is **no fallback**: delivery is permanently zero for that pair, not degraded.

[...]

PR #2716, opened 2026-07-16
fix(zenoh): restore reliable node links and large cross-host messages under zenoh 1.9

phil-opp:
Addresses part of #2715. Three independent zenoh-1.9 defects, all silent, all found while investigating startup message loss (#2666).

> **Correction to an earlier revision of this description and of #2715:** it quoted release notes claiming *"routing.peer is removed—all peers now operate in peer-to-peer mode"* and that `peers_failover_brokering` was removed. **That quote is not in the 1.9.0 release notes and should be disregarded** — it came from an unverified source. The conclusion is unchanged, but the evidence below is now first-hand. The "must form a clique" claim is likewise withdrawn pending verification.

## 1. `routing/peer: linkstate` became a silent no-op → node links are racy

| source | evidence |
|---|---|
| [1.8.0 `linkstate_peer/mod.rs`](https://github.com/eclipse-zenoh/zenoh/blob/1.8.0/zenoh/src/net/routing/hat/linkstate_peer/mod.rs) | `peer_full_linkstate = unwrap_or_default!(config.routing().peer().mode()) == *"linkstate"` — our setting **worked** |
| [1.9.0 `peer/mod.rs`](https://github.com/eclipse-zenoh/zenoh/blob/1.9.0/zenoh/src/net/routing/hat/peer/mod.rs) | the `linkstate_peer` hat is gone; the `peer` hat passes **`full_linkstate: false`**, hardcoded |
| [1.9.0 release notes](https://github.com/eclipse-zenoh/zenoh/releases/tag/1.9.0) | *"Disable `full_linkstate` in `peer::Hat::Network`"* — listed under **Bug fixes**, not breaking changes |
| runtime | `routing.peer.mode` and `routing.peer.linkstate` are deprecated and have no effect` on every session open |

`insert_json5` still returns `Ok`, so our own `warn!` branch never fired. We lost transitive peer routing and nothing told us.

Consequence: node↔node links depend entirely on gossip's autoconnect, which is asynchronous and best-effort. When it loses, there is no relay to fall back on — delivery for that pair is permanently zero, not degraded, which is why it's per-pair independent.

Measured on `examples/rust-dataflow` (dev container, no multicast — the explicit-endpoint path, representative of CI):

| | route failures |
|---|---|
| before (gossip autoconnect) | **5 / 20** |
| after | **0 / 20** |

In a failing run `rust-node/random` published 2066 samples over 10s and its consumer received **zero**, while another pair in the same run delivered 1949/1949.

**Fix:** the nodes a daemon spawns are all on one machine and therefore loopback-addressable, so we say who dials whom instead of hoping discovery converges. The daemon reserves a listener per local node before spawning and injects `DORA_ZENOH_LISTEN`, plus a `DORA_ZENOH_CONNECT` list naming the daemon and the nodes this one consumes from.

Each consumer dials only its **producers** — a zenoh transport is bidirectional, so the consumer's dial carries the producer's data back. That is `|edges|` links rather than `N²`, and exactly the set the dataflow requires. Traffic stays direct peer-to-peer: **no relay hop, no change to zero-copy**.

## 2. `lowlatency` made >64 KiB cross-host messages undeliverable, silently

LowLatency cannot fragment, so a message must fit one batch — and `batch_size` is capped at 64 KiB by `pub type BatchSize = u16` (65535 is the *maximum*, not just the default; no feature widens it).

Shared memory hid this: an SHM payload travels as a ~16-byte descriptor and never fragments. But **SHM is per-host by construction, so it cannot negotiate between machines** — meaning a >64 KiB message to another host had no working path at all. Not "if SHM fails"; for cross-host it is guaranteed.

The failure is silent and misattributed: [`lowlatency/link.rs`](https://github.com/eclipse-zenoh/zenoh/blob/1.9.0/io/zenoh-transport/src/unicast/lowlatency/link.rs) writes a 4-byte u32 length prefix with **no size check on TX**, so `put()` returns `Ok` and the bytes reach the socket; the receiver's buffer is bounded by the u16 `batch_size` and rejects the frame with `Batch len is invalid`. The publisher believes it succeeded.

Reproduced on `examples/benchmark` (release, ascending sizes, otherwise identical runs), forcing SHM off to stand in for cross-host:

| condition | 40KB | 400KB | 3MB |
|---|---|---|---|
| SHM on + lowlatency on (old default) | ok | ok | ok |
| SHM **off** + lowlatency **on** | ok | **LOST** | **LOST** |
| SHM **off** + lowlatency **off** | ok | ok | ok |
| **SHM off + this PR** | ok | **ok** | **ok** |

Row 3 isolates the cause; row 1 shows why we never saw it. With the fix the SHM-off case delivers 400KB and 3MB again (**2 of 2, from 0 of 2**) at 255 µs and 1.7 ms — i.e. genuinely fragmented rather than riding a descriptor.

**Fix:** drop `transport/unicast/lowlatency`. `qos/enabled: false` goes with it — it existed *only* because the low-latency transport is negotiated without QoS.

Not a latency regression (release, `examples/benchmark`, vs the old config):

| size | p50 before | p50 after | throughput |
|---|---|---|---|
| 64B | 65µs | **55µs** | −28% |
| 512B | 66µs | **58µs** | +8% |
| 4KB | 72µs | **62µs** | **+51%** |
| 16KB | 73µs | **63µs** | **+41%** |
| 400KB | 69µs | **62µs** | +7% |

The two must be removed **together**: dropping `lowlatency` alone cost ~25 µs p50 on small messages, and restoring QoS is what recovers it — the publishers' `Priority::RealTime` was silently inert the whole time QoS was disabled. `express(true)` on publishers is untouched; that is the setting actually carrying small-message latency.

## 3. The shared-memory comment was inverted

It claimed we *"deliberately do NOT enable"* `transport/shared_memory/enabled`. That setting **defaults to `true`** ([defaults.rs](https://github.com/eclipse-zenoh/zenoh/blob/1.9.0/commons/zenoh-config/src/defaults.rs), so `Config::default()` agrees — not just the shipped file), and has since 1.0. We get zero-copy *by default*, not by opting out. Anyone "correcting" the code to match the comment would silently turn every SHM buffer into a copy.

## Scope

- **Only local nodes are planned.** `nodes` spans the whole dataflow including other daemons' nodes, whose endpoints are not on this host's loopback — dialing one would hit nothing, or an unrelated local process. A local consumer of a remote producer gets no dial for it and keeps existing cross-machine behavior. Covered by a regression test.
- **Dynamic nodes** keep daemon-endpoint-only + gossip, since they join at arbitrary times and aren't in the spawn set.
- **Multi-machine/NAT is untouched** — those deployments inject their own config via `ZENOH_CONFIG_PATH`, which bypasses this path entirely.

## Still open in #2715

- **daemon↔daemon across networks.** `inter_daemon_peer`'s hub only provides *discovery*; peers no longer relay, so two daemons behind NAT can discover each other and still not exchange data. Needs a router or 1.9 regions.
- **Dynamic nodes** remain exposed to the gossip race.
- The upstream transport bug itself: `put()` returning `Ok` for a message the peer must reject is zenoh's to fix.

## Verification

- 4 unit tests for the peering plan: graph mapping, cycle (`a→b→a`), dynamic-node exclusion, remote-node exclusion.
- 20-run before/after route measurement; 3-condition (+fix) large-message matrix; release latency/throughput comparison — all above.
- Full workspace tests, clippy `-D warnings`, `cargo fmt --check`, `cargo check --examples`, example dataflow end-to-end.

**Caveat on the numbers:** measured in a shared dev container, one run per condition for the benchmark. Directions are consistent and the deltas large, but the exact percentages want confirmation on quiet hardware. The route-failure rate (5/20) is specific to a no-multicast environment — representative of CI and dev containers, though the fix removes the race rather than tuning it.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1941,6 +1941,7 @@ impl Daemon {
                         ft_stats: self.ft_stats.clone(),
                         shutdown: dataflow.listener_shutdown_rx.clone(),
                         zenoh_connect_endpoint: self.zenoh_listen_endpoint.clone(),
+                        zenoh_peering: dataflow.zenoh_peering.clone(),
                     };
                     let mut logger = self
                         .logger
@@ -2780,11 +2781,20 @@ impl Daemon {
             .try_clone()
             .await
             .context("failed to clone logger")?;
-        let dataflow = RunningDataflow::new(
+        let mut dataflow = RunningDataflow::new(
             dataflow_id,
             self.daemon_id.clone(),
             dataflow_descriptor.clone(),
         );
+        // Decide who dials whom before anything spawns: zenoh 1.9 peers do not
+        // relay, so a producer/consumer pair that never forms a direct link can
+        // never exchange data. Assign the links explicitly instead of leaving
+        // them to gossip's best-effort autoconnect.
+        dataflow.zenoh_peering = Arc::new(crate::spawn::plan_zenoh_peering(
+            &nodes,
+            &spawn_nodes,
+            self.zenoh_listen_endpoint.as_deref(),
+        ));
         let dataflow = match self.running.entry(dataflow_id) {
             std::collections::hash_map::Entry::Vacant(entry) => {
                 self.working_dir
@@ -3032,6 +3042,7 @@ impl Daemon {
             ft_stats: self.ft_stats.clone(),
             shutdown: dataflow.listener_shutdown_rx.clone(),
             zenoh_connect_endpoint: self.zenoh_listen_endpoint.clone(),
+            zenoh_peering: dataflow.zenoh_peering.clone(),
         };
 
         let mut tasks = Vec::new();
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -183,6 +183,10 @@ pub struct LogSubscriber {
 pub struct RunningDataflow {
     pub(crate) id: uuid::Uuid,
     pub(crate) descriptor: Descriptor,
+    /// Per-node zenoh listener + dial-list, so the node↔node links this dataflow
+    /// needs are established deterministically rather than left to gossip.
+    /// Populated when the dataflow is spawned; see `plan_zenoh_peering`.
+    pub(crate) zenoh_peering: Arc<BTreeMap<NodeId, crate::spawn::NodeZenohPeering>>,
     pub(crate) pending_nodes: PendingNodes,
     pub(crate) dataflow_started: bool,
     pub(crate) subscribe_channels: HashMap<NodeId, Sender<Timestamped<NodeEvent>>>,
@@ -257,6 +261,7 @@ impl RunningDataflow {
         let (listener_shutdown_tx, listener_shutdown_rx) = tokio::sync::watch::channel(false);
         Self {
             id: dataflow_id,
+            zenoh_peering: Arc::new(BTreeMap::new()),
             pending_nodes: PendingNodes::new(dataflow_id, daemon_id),
             dataflow_started: false,
             subscribe_channels: HashMap::new(),
--- a/binaries/daemon/src/spawn/mod.rs
+++ b/binaries/daemon/src/spawn/mod.rs
@@ -1,5 +1,5 @@
 pub use prepared::PreparedNode;
-pub use spawner::Spawner;
+pub use spawner::{NodeZenohPeering, Spawner, plan_zenoh_peering};
 
 mod command;
 mod prepared;
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -116,17 +116,20 @@ pub struct NodeZenohPeering {
 /// Assign every node a loopback listener and dial-list so the links the dataflow
 /// needs are established *by construction*.
 ///
-/// Since zenoh 1.9 peers no longer relay for each other (`routing.peer` and
-/// `peers_failover_brokering` were both removed): a peer-to-peer region has to be
-/// a clique, and a producer/consumer pair that never forms a direct link simply
-/// cannot exchange data — there is no relay to fall back on and no amount of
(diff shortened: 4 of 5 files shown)


phil-opp:
🤖 **Automated review by Claude** — generated fully automatically, with no human verification. Treat accordingly.

**No issues found.** Reviewed the full diff plus the surrounding daemon/topics code; the peering plan is correct:

- `plan_zenoh_peering(&nodes, &spawn_nodes, ...)` is passed `spawn_nodes` — the set this daemon actually spawns (used as `local` in the mappings loop). Remote nodes therefore get no loopback listener, and a local consumer of a remote producer gets no dial for it, so cross-machine behavior is preserved (covered by `remote_nodes_are_never_dialled_on_local_loopback`).
- Each consumer dials only its `User`-mapped producers; self-references and timer/log mappings are filtered by `input_sources` + the `source == node_id` guard, so intra-operator links and cyclic graphs (`cyclic_dataflow_links_both_directions`) resolve — listeners are reserved for every node before any dial-list is built.
- The daemon endpoint stays first in every connect list, so nodes still reach the daemon and the #2666 readiness/daemon-forward path is untouched; this only *adds* direct node↔node links. `connect.dedup()` is a harmless no-op here since the entries are already distinct.
- `open_zenoh_session_with_listen` prefers an explicitly-passed `listen_endpoint` over `DORA_ZENOH_LISTEN` (`.or(env_listen_endpoint)`), so the daemon's own session is unaffected — only spawned nodes pick up the injected listener. `split_endpoints` correctly parses the comma-joined `DORA_ZENOH_CONNECT`.
- The `routing/peer: linkstate` deletion and the SHM-comment correction match zenoh 1.9 behavior (peers no longer relay; SHM stays enabled by default), and both are documented in-code.

The endpoint reservation keeps a small TOCTOU window (bind port 0 → drop → node rebinds), but that is pre-existing, documented, and defended by `listen/exit_on_failure: false` plus the `info().locators()` bind check; a rare miss degrades to the daemon-forwarded path rather than failing.

---

PR #2717, opened 2026-07-16
ci: trigger merge-queue checks on push to trunk-merge branches

phil-opp:
Trunk's queue currently runs in draft-PR mode, so every batch, retry and
bisection opens a draft PR against main. Over the last two weeks that was
289 draft PRs across 133 source PRs — 33% of them bisection branches.

Point the required checks at the batch branch itself instead. Trunk pushes
`trunk-merge/...`, the push trigger picks it up, and the checks attach to
the batch head SHA, which is what the queue reads. pip-release.yml has
worked this way since it went push-only and `pip-release all green` is a
required status, so the mechanism is already proven on this repo.

The expensive tier now gates on `github.event_name != 'pull_request'`
alone; the `startsWith(github.head_ref, 'trunk-merge/')` half only existed
to catch the draft-PR event. This reports correctly in either mode, so it
can merge before the dashboard setting is flipped.

`test-c-cpp-libraries.yml` needs more care: its path filter diffs against
`github.event.before`, which is the all-zero SHA on a freshly pushed batch
branch and a real SHA on a re-pushed one. The first fails open and would
run the 3-OS matrix on every batch; the second narrows the diff to the
delta between attempts and would wrongly skip it. Batch branches are now
keyed on the ref name and diffed against main, which is correct for both.

Requires flipping Draft Pull Request Creation to Disabled in the Trunk
dashboard after this lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,16 @@
 #     These gate the merge itself. Running them only on Trunk's
 #     batched queue branch means 5 PRs share one expensive CI run.
 #
-# Detection of merge-queue runs relies on Trunk's draft-PR queue
-# mode: the queue branch is named `trunk-merge/...` and the
-# `pull_request` event fires with `github.head_ref` pointing at it.
+# Merge-queue runs are push-triggered: Trunk pushes a `trunk-merge/...`
+# batch branch and the `push` trigger below picks it up, so the expensive
+# tier gates on `github.event_name != 'pull_request'` alone. Checks attach
+# to the batch head SHA, which is what Trunk reads — pip-release.yml has
+# relied on exactly this since it went push-only, and `pip-release all
+# green` is a required status today.
+#
+# This reports correctly whether or not Trunk also opens a draft PR per
+# batch, so the two can overlap during a cutover — see .trunk/trunk.yaml
+# for that setting and why it stays off.
 #
 # Moved to .github/workflows/nightly.yml (runs once daily, ~3-4h):
 #   - test on macos + windows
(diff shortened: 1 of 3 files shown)


phil-opp:
🤖 **Automated review by Claude** — generated fully automatically, with no human verification. Treat accordingly.

**No issues found.** Reviewed the CI trigger changes against the workflows and `.trunk/trunk.yaml`:

- Every merge-queue `required_status` still reports on a `trunk-merge/**` batch branch. The expensive tier (`Test`, `E2E Tests`, `Semantic contract tests`, `Benchmark regression check`) now gates on `github.event_name != 'pull_request'`, which holds for the `push` event, so it runs and reports on the batch head SHA. The cheap tier (`Format`/`Clippy`/`Check`/`Typos`/`Audit`/`Unwrap budget`/`License check`) has no `pull_request`-only `if:`, so it also fires on the new `push: trunk-merge/**` trigger. `find_package smoke all green` is an `if: always()` aggregator, so it always emits a check — nothing the queue waits on can hang permanently Pending.
- `test-c-cpp-libraries.yml`: keying `BASE` on `REF_NAME` (`trunk-merge/* → main`, else `$BEFORE`) is correct for both a fresh batch (all-zero `before`, which would otherwise fail open into the 3-OS matrix) and a re-pushed batch (real `before`, which would otherwise narrow the diff and wrongly skip). `main...$SHA` yields exactly the batch's delta.
- Concurrency is keyed on `github.ref`, so a batch `push` (`refs/heads/trunk-merge/...`) and any overlapping draft-PR `pull_request` (`refs/pull/N/merge`) land in different groups and don't cancel each other during the cutover.

The double-run of the cheap tier while draft-PR mode is still on is acknowledged in the PR and resolves once the dashboard setting is flipped. No correctness concerns.

---

PR #2719, opened 2026-07-16
fix(ci): make the release pipelines work for the v1.0.0 rc tags

haixuanTao:
## Why

The `v1.0.0-rc.2` release attempt on 2026-07-15 failed three independent ways:

1. **release.yml never triggered** — its tag glob `v[0-9]+.[0-9]+.[0-9]+` doesn't match `-rc.N` suffixes, so the tag push was silently ignored (no CLI binaries, no release assets, no crates). Same for `v1.0.0-rc.1` in April.
2. **Cargo Release failed** ([run 29433155502](https://github.com/dora-rs/dora/actions/runs/29433155502)) — `dora-core` embeds the stdlib type YAMLs from the repo-root `types/` directory, which `cargo package` cannot include (outside the crate dir), so the publish verify build dies with `couldn't read src/../../../types/std/core/v1.yml`.
3. **pip-release "succeeded" but published the wrong version** — the workspace version was still `1.0.0-rc1`, so PyPI got `1.0.0rc1` from the rc.2 release.

## What

- **release.yml**: add a `v[0-9]+.[0-9]+.[0-9]+-*` tag pattern for prerelease tags.
- **Move `types/` → `libraries/core/types/`** and update the `include_str!` paths. Verified with a full `cargo package -p dora-core` (the exact step that failed). Side benefit: running `dora record`/`replay` from the workspace root no longer mis-loads the stdlib YAMLs as user types.
- **Fix the same latent bug in `dora-node-api-cxx` / `dora-operator-api-cxx`**: their build.rs `include_str!`'d cmake templates from `apis/c/node/`. Each crate now carries local copies, and a new workspace test (`tests/cmake-template-sync.rs`) keeps the three copies in sync.
- **Fix both crate publish lists** (release.yml + cargo-release.yml), validated against `cargo metadata`:
  - added `dora-hub-client` and `dora-memory-pool` (deps of dora-cli / dora-daemon — publishing would have failed without them)
  - added `dora-operator-api-python` to release.yml (optional dep of dora-runtime)
  - removed the `publish = false` binaries (record/replay/ros2-bridge-node) which `cargo publish` rejects
  - re-ordered topologically including dev-dep edges: `dora-ros2-bridge` dev-depends on `dora-daemon`/`dora-cli` (version reqs survive packaging and are resolved by the verify build), and the cxx APIs optionally depend on `dora-ros2-bridge`
- **release.yml publish loop**: fail on real errors instead of downgrading everything to a warning (only "already uploaded" is tolerated, keeping re-runs idempotent), and drop the `sleep 5` (cargo ≥1.66 blocks until the index has the crate).
- **Version bump `1.0.0-rc1` → `1.0.0-rc.2`** so the next tag publishes consistent versions everywhere.
- Updated stale doc/comment references to the old `types/` location.

## Validation

- `cargo package -p dora-core` (full verify build) — the exact failing step from yesterday, now green
- `cargo package --list` confirms the YAMLs / cmake templates land in the tarballs
- publish-list order machine-checked against `cargo metadata` (all dep kinds)
- `cargo test -p dora-core --lib` (239 tests), new sync test, fmt, clippy on touched crates
- workflow YAML parse + `bash -n` on touched script

## Known follow-ups (out of scope)

- The release flow double-publishes by design: release.yml (tag) publishes crates+wheels, then its GitHub-release step triggers cargo-release.yml + pip-release.yml which redo the same work idempotently. Worth consolidating — `cargo publish --workspace` (stable since 1.90; CI pins 1.92) could replace both hand-ordered lists.
- `dora record`/`replay`'s `cargo install dora-record-node` fallback can never succeed while those crates are `publish = false` (pre-existing; the release tarball only ships the `dora` binary).


phil-opp:
Just FYI: Merges should go through the trunk comment, which runs the full test pipeline before merging. PRs only run a small (fast) subset, so if we merge directly we might end up with a broken `main` branch. Also, merging directly resets the merge queue and forces a rebuild/retest of all PRs in the trunk queue.

phil-opp:
Resubmitted to the trunk queue after your no-op push

> 🚫 This pull request was removed from the merge queue because it was pushed to by @haixuanTao. Please re-submit it in order to merge. See more details [here](https://app.trunk.io/dora-rs/merge-queue/0ab727ff-7713-4487-988a-07f8b3b98e52/2719).

phil-opp:

No issues found. I focused on the parts most likely to break a release:

- The crate lists in `release.yml` and `cargo-release.yml` are consistent with each other, and I walked the normal-dependency edges across the published crates without finding an ordering inversion. The dev-dep ordering (daemon/cli before `dora-ros2-bridge`) does matter for `cargo-release.yml`, which publishes with the verify build and so must resolve dev-deps; for `release.yml` under `--no-verify` it's harmless.
- `dora-hub-client` and `dora-memory-pool` are the right additions — they're path-deps of cli/daemon that would otherwise break those publishes. `dora-log-utils` and `dora-mavlink2-bridge` are correctly left out (`publish = false` / only consumed by `publish = false` binaries).
- The `types/` → `libraries/core/types/` move updates every reference I could find (the `include_str!`s, `hub/init.rs`, docs, and both nightly workflows), and the relative include path from `libraries/core/src/types.rs` resolves correctly to the new location.
- The crate-local cmake copies for the two C++ crates are kept honest by the new `tests/cmake-template-sync.rs`.

These pipelines only run on tags, so nothing here can regress `main`; the change looks correct for the rc release it targets.

---

phil-opp:
Please stop pushing to the branch once it's in the trunk queue. It will remove it every time and require reapproval.

PR #2720, opened 2026-07-16
ci: build the Python extension in dev profile for contract tests

phil-opp:
maturin's PEP 517 backend forces `--profile release` unless told
otherwise, so `uv pip install -e apis/python/node` release-built
dora-node-api-python's entire dep closure with thin LTO on every run:
15m21s ("Prepared 3 packages in 15m 21s"), sharing nothing with the
dev-profile rust-cache the job already restores.

The step looks like it hangs on "Downloading numpy" — that's a red
herring. numpy downloads in ~1s; the silence after it is the cargo
build, which prints nothing until it finishes.

Set MATURIN_PEP517_ARGS="--profile dev" so the build reuses the
restored dev cache. Caching target/release instead would cost ~1 GB
against the repo's 10 GB Actions cache cap.

Crates whose features differ under the extension build (dora-runtime/
python, dora-cli/python, pyo3/extension-module) still rebuild; the
expected win is cache hits on the leaf tier (arrow, tokio, zenoh) plus
dropping thin LTO. The extension is now a debug build — the daemon it
drives already is, so watch contract_streaming_example and the
cross-language tests for timeout regressions.

nightly.yml has six copies of the same install; left alone pending the
measured delta here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -426,6 +426,18 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name != 'pull_request'
     timeout-minutes: 30
+    env:
+      # maturin's PEP 517 backend forces `--profile release` unless told
+      # otherwise (`ensure_release_profile` in maturin's commands/pep517.rs),
+      # so the `uv pip install -e apis/python/node` step below release-built
+      # dora-node-api-python's whole dep closure — arrow, tokio, zenoh,
+      # dora-cli — with thin LTO: 15m21s, sharing nothing with the
+      # dev-profile `shared-key: workspace` cache this job restores. Dev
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

`MATURIN_PEP517_ARGS` is the correct env var for maturin's PEP 517 backend, `--profile dev` is a valid cargo profile, and the variable is job-scoped so it only affects the maturin / `uv pip install` build (the other cargo steps ignore it). The one thing worth keeping an eye on is the one the commit message already calls out: the extension is now a debug build, so if `contract_streaming_example` or the cross-language contract tests start flaking on timeouts, this is the change to revisit. CI-only; nothing blocking.

---

PR #2729, opened 2026-07-17
fix(download): derive URL fallback filename without query string

phil-opp:

## Issue

When a download has no `Content-Disposition` header, `get_filename` fell back to deriving the name from the URL:

```rust
Path::new(response.url().as_str()).file_name()
```

`response.url().as_str()` is the **serialized** URL — query string and fragment included — and `Path::file_name` only splits on `/`. So for a URL like `https://.../model.bin?X-Amz-Signature=…&expires=…` the derived name became `model.bin?X-Amz-Signature=…`.

This is the common shape of presigned **S3 / GitHub-release** redirects (no `Content-Disposition`, long signed query). Concrete failures:

- **Download aborts**: a presigned query easily exceeds the 255-char sanitize limit, so `get_filename` returned `None` and the download failed with *"Could not find a filename"* — even though `model.bin` was right there.
- **Wrong name on success**: shorter queries produced `model.bin?x=1` on disk.
- **`File::create` fails on Windows**, where `?` is an illegal filename character.

## Fix

Derive the name from the parsed URL's last non-empty **path segment**, which is inherently free of the query and fragment:

```rust
fn filename_from_url(url: &reqwest::Url) -> Option<String> {
    url.path_segments()?
        .rfind(|segment| !segment.is_empty())
        .map(|segment| segment.to_string())
}
```

`reqwest` already re-exports `url::Url`, so no new dependency. Behavior for a trailing slash is unchanged (`/dir/` → `dir`, matching the old `Path::file_name`).

## Validation

- New unit tests cover query strings, fragments, plain paths, and trailing slashes (the URL-fallback path had no prior coverage).
- `cargo test -p dora-download` — 9 passed.
- `cargo clippy -p dora-download -- -D warnings` — clean.
- `cargo fmt -p dora-download -- --check` — clean.

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -24,6 +24,18 @@ fn parse_content_disposition_filename(header: &str) -> Option<String> {
     (!name.is_empty()).then(|| name.to_string())
 }
 
+/// Derive a candidate filename from the *last path segment* of a URL.
+///
+/// Uses the parsed URL structure rather than the serialized string, so the
+/// query string and fragment (e.g. the long presigned-URL parameters that S3
+/// and GitHub-release redirects append) never leak into the name. Returns
+/// `None` when the URL has no non-empty path segment.
+fn filename_from_url(url: &reqwest::Url) -> Option<String> {
+    url.path_segments()?
+        .rfind(|segment| !segment.is_empty())
+        .map(|segment| segment.to_string())
+}
+
 fn get_filename(response: &reqwest::Response) -> Option<String> {
     let raw_name = response
         .headers()
@@ -31,13 +43,8 @@ fn get_filename(response: &reqwest::Response) -> Option<String> {
         .and_then(|value| value.to_str().ok())
         .and_then(parse_content_disposition_filename);
 
-    // If Content-Disposition header is not available, extract from URL
-    let raw_name = raw_name.or_else(|| {
-        Path::new(response.url().as_str())
-            .file_name()
-            .and_then(|n| n.to_str())
-            .map(|s| s.to_string())
-    });
+    // If Content-Disposition header is not available, extract from the URL.
+    let raw_name = raw_name.or_else(|| filename_from_url(response.url()));
 
     // Sanitize: strip path components to prevent traversal,
     // reject null bytes and overly long names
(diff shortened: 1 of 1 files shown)


phil-opp:

This PR is one of **6** opened by an automated review sweep of `origin/main` (session `dreamy-bardeen-e4bfsv`). All changes are machine-generated by Claude (an AI agent) and should be reviewed carefully. Each was verified per-branch (fmt + clippy + affected-crate tests), and the combined diff passed the full affected-crate suite plus a `/review` + `/simplify` gate.

### PRs by category

**Correctness / panic-safety**
- **#2729** `fix(download)` — URL-fallback filename included the query string/fragment, breaking (or misnaming) presigned S3 / GitHub-release downloads. *(this PR)*
- **#2730** `fix(coordinator)` — `dora list` double-lists a partially-finished multi-daemon dataflow and can report it `Finished` while still running; adds the guard the `Clean` handler already has.
- **#2731** `fix(daemon)` — a `serde_json` error in the metrics loop bypassed the `System` restore, permanently emptying the shared sysinfo `System` (a single `NaN` CPU value poisons all later readings).
- **#2732** `fix(core)` — an operator `id` with a char illegal in a `DataId` panicked during descriptor validation (`dora check`/`run`/`graph`); now a clean error. Includes a regression test proven to panic pre-fix.

**Optimization (hot path)**
- **#2733** `perf(daemon)` — `note_output_sent_to_local_receivers` cloned `(NodeId, DataId)` keys per message even when the deadline/circuit-breaker maps are empty; adds the `is_empty()` guards its sibling already has.
- **#2734** `perf(daemon)` — `LogBroadcast` rebuilt an identical `Metadata`/`Arc` per subscriber; hoisted to one build, matching the sibling Timer/Logs loops.

### Rejected findings (surfaced but not turned into PRs)

- **Signal-number→name table is Linux-specific** (`message/common.rs`) — diagnostic text only, and overlaps prior signal-name work (#2624).
- **`dora inspect top` selection not re-clamped when the node list shrinks** — cosmetic; ratatui clamps rendering, no panic.
- **`FormatDataflowError` `hidden > 1` gate / stale comment** — the `> 1` threshold is a deliberate grammar choice ("1 consequential errors"); subjective.
- **`LogMessageHelper` builds `NodeId` without validation** — fails soft (a malformed id simply never matches a filter); not worth a change.
- **`DoraOutputSender` / `Daemon::run_dataflow` missing doc comments** — trivial doc-only nits; deferred.
- **`working_dir` / `dataflow_node_results` grow unbounded on long-lived daemons** — `working_dir` is legitimately read post-finish (Logs handler), so a correct fix needs a retention-policy design rather than a plain `remove`; deferred.
- **Coordinator dynamic-topology handlers block the event loop on unbounded `send_and_receive`** — real, but the fix (bounded timeout / offload) is an architectural change too large for an automated PR.
- **`NodeMetrics` per-node clone behind `#[cfg(feature = "metrics")]`** — only compiles under a non-default feature; negligible.

### Note for reviewers/CI

`cargo test -p dora-core` in this environment shows one **unrelated, pre-existing** failure — `topics::tests::ipv6_endpoints_are_bracketed` — because the sandbox has no IPv6 (`reserve_zenoh_endpoint(::1)` cannot bind). `topics.rs` is untouched by any of these 6 PRs.

---

phil-opp:

No issues found.

`filename_from_url` deriving the name from the parsed URL's last non-empty path segment (via `path_segments().rfind(|s| !s.is_empty())`) correctly strips the query string and fragment that the old `Path::new(url.as_str()).file_name()` embedded — fixing both the 255-char sanitize-limit abort and the Windows `?`-in-filename failure for presigned S3/GitHub-release redirects. Edge cases are handled: trailing slash falls back to the last non-empty segment, bare root returns `None`, matching the prior `Path::file_name` semantics. Tests cover query, fragment, and trailing-slash cases.

---

PR #2732, opened 2026-07-17
fix(core): reject invalid operator ids instead of panicking during resolve

phil-opp:

## Issue

`resolve_aliases_and_set_defaults` rewrites an input mapping that targets a single-operator node to the fully-qualified output `"{operator_id}/{output}"`, built via `DataId::from(String)`:

```rust
mapping.output = DataId::from(format!("{op_name}/{}", mapping.output));
```

Unlike `NodeId` and `DataId`, `OperatorId` performs **no charset validation** on construction (`FromStr` is `Infallible`, `Deserialize` is derived), so an operator `id` may contain characters that are illegal in a `DataId` (e.g. a space, or an empty id producing a leading slash). Because `DataId::from(String)` **panics** on an invalid id, resolving such a well-formed-but-invalid descriptor aborts the process instead of returning an error.

This is reachable from every validation entry point that calls `resolve_aliases_and_set_defaults` — `check_wiring`, `check_dataflow` (`dora check` / `dora run`), and `visualize` (`dora graph`). Example that panics today:

```yaml
nodes:
  - id: producer
    operator:
      id: "bad id"          # legal OperatorId, illegal DataId component
      python: op.py
      outputs: [out]
  - id: consumer
    path: ./consumer
    inputs:
      x: producer/out
```

→ `panicked at id.rs: invalid DataId 'bad id/out'`.

## Fix

Parse the concatenation fallibly via a small `qualified_operator_output` helper and propagate a descriptive `eyre` error (the function already returns `Result`):

```rust
fn qualified_operator_output(op_name: &OperatorId, output: &DataId) -> Result<DataId> {
    format!("{op_name}/{output}").parse::<DataId>().wrap_err_with(|| { ... })
}
```

## Validation

- New regression test `resolve_rejects_operator_id_that_is_invalid_in_a_data_id`. I confirmed it **panics** on the pre-fix code (`invalid DataId 'bad id/out'`) and returns a clean `Err` with the fix.
- `cargo test -p dora-core` — passes (one unrelated pre-existing failure, `topics::tests::ipv6_endpoints_are_bracketed`, is environmental — the sandbox has no IPv6 and it does not touch this diff).
- `cargo clippy -p dora-core -- -D warnings` and `cargo fmt` — clean.

---

### DIFF
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -106,6 +106,21 @@ fn node_is_dynamic(node: &ResolvedNode) -> bool {
         if matches!(n.source, NodeSource::Local) && n.path == DYNAMIC_SOURCE)
 }
 
+/// Build the fully-qualified output id `"{operator}/{output}"` for an input
+/// mapping that references a single-operator node.
+///
+/// `OperatorId` is not charset-validated on construction (unlike `NodeId` and
+/// `DataId`), so an operator `id` may contain characters that are illegal in a
+/// `DataId`. Parsing the concatenation fallibly turns such an id into a clean
+/// validation error instead of a panic — `DataId::from(String)` would
+/// otherwise abort descriptor validation (`dora check`, `dora run`, `dora
+/// graph`, ...) on well-formed-but-invalid YAML.
+fn qualified_operator_output(op_name: &OperatorId, output: &DataId) -> Result<DataId> {
+    format!("{op_name}/{output}").parse::<DataId>().wrap_err_with(|| {
+        format!("operator id `{op_name}` is not a valid data-id component (referenced in an input mapping as `{op_name}/{output}`)")
+    })
+}
+
 impl DescriptorExt for Descriptor {
     fn resolve_aliases_and_set_defaults(&self) -> eyre::Result<BTreeMap<NodeId, ResolvedNode>> {
         let default_op_id = OperatorId::from(SINGLE_OPERATOR_DEFAULT_ID.to_string());
@@ -128,7 +143,7 @@ impl DescriptorExt for Descriptor {
                     if let InputMapping::User(m) = &mut input.mapping
                         && let Some(op_name) = single_operator_nodes.get(&m.source).copied()
                     {
-                        m.output = DataId::from(format!("{op_name}/{}", m.output));
+                        m.output = qualified_operator_output(op_name, &m.output)?;
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

`OperatorId` is not charset-validated on construction, so an operator id containing a `DataId`-illegal character previously reached the panicking `DataId::from(String)` during `resolve_aliases_and_set_defaults`, aborting `dora check`/`run`/`graph`. Routing both call sites through `qualified_operator_output`, which uses fallible `parse::<DataId>()` + `wrap_err_with` and propagates via the existing `?`, turns this into a clean, descriptive error. The regression test exercises the previously-panicking path and asserts the offending id is named.

---

phil-opp:

I looked at the diff on its own terms (not the description). The change itself is correct: routing the two single-operator alias-rewrite sites through a fallible `parse::<DataId>()` turns a panic into a clean `eyre` error, and the reuse of `DataId` parsing means it rejects exactly the ids `DataId` considers invalid. The regression test exercises the real path and would fail without the fix.

One issue: **the panic isn't actually eliminated for all invalid operator ids** — only for the case where the operator's output is referenced by another node's input mapping (the only path that goes through the alias rewrite).

`OperatorId` is not charset-validated anywhere (`FromStr` is `Infallible`, `From<String>` is unchecked, and there's no operator-id check in `validate.rs`), so an operator id containing an invalid char (e.g. a space) still passes descriptor validation / `dora check`. If that operator's outputs are *not* referenced by any consumer, the alias rewrite never builds the qualified id, so this PR's new error path is never hit — and the daemon then panics at spawn when it builds the operator's own qualified `DataId`s unconditionally:

- `binaries/daemon/src/lib.rs:5781` — `DataId::from(format!("{}/{input_id}", operator.id))`
- `binaries/daemon/src/lib.rs:5797` — `DataId::from(format!("{}/{output_id}", operator.id))`
- `binaries/runtime/src/lib.rs:352` — `operator_output_id` → `DataId::from(format!("{operator_id}/{output_id}"))`

`DataId::from(String)` is the documented-`# Panics` constructor, so each of these crashes the daemon on an invalid operator id.

Repro shape: a runtime node with `operators: [{ id: "bad id", outputs: [x] }]` where `bad id/x` is not wired into any other node. `dora check` passes; the daemon panics at spawn.

This isn't a regression — it's a pre-existing latent panic, and the PR is a net improvement — so I wouldn't block on it. But the PR's stated goal ("reject invalid operator ids instead of panicking") isn't fully met. The complete fix is to charset-validate the operator id once at descriptor-validation time (mirroring `NodeId`/`DataId`), which closes all four sites at once rather than the referenced-output path only.

---

phil-opp:
Superseded by #2467, which landed the same fix in main: `prefix_output_with_operator_id`
parses `"{operator}/{output}"` fallibly at both alias-resolution call sites
(libraries/core/src/descriptor/mod.rs), with an equivalent regression test
(`invalid_operator_id_prefix_errors_instead_of_panicking`). Nothing in this
branch remains to apply, so closing.

PR #2739, opened 2026-07-17
fix(cli): report hidden failures in all-cascading DataflowError case

### LINKED ISSUE #2736: CLI: `FormatDataflowError` silently drops a failure when exactly one error is hidden (all-cascading case)
> **Note:** This issue was created by an automated, scheduled Claude code-review check. It was found by reading `binaries/cli/src/formatting.rs`; it has not been reproduced against a live dataflow. Please verify before acting. Severity: **low** (user-facing error reporting, not a crash or data corruption).

## Summary

`FormatDataflowError::fmt` can display **only one** of several failed nodes while giving the user **no indication** that other failures occurred and **no pointer** to the `out/<uuid>` folder — specifically in the "all errors are cascading" branch when exactly one error ends up hidden.

## Location

`binaries/cli/src/formatting.rs:37-62`

```rust
let hidden = if !non_cascading.is_empty() {
    let printed = non_cascading.len();
    for (id, err) in non_cascading {
        writeln!(f, "Node `{id}` failed: {err}")?;
    }
    total_failed - printed
} else {
    // no non-cascading errors -> print earliest cascading
    let mut all: Vec<_> = failed.collect();
    all.sort_by_key(|(_, e)| e.timestamp);
    if let Some((id, err)) = all.first() {
        write!(f, "Node `{id}` failed: {err}")?;
        total_failed - 1          // <-- only ONE cascading error is printed
    } else {
        write!(f, "unknown error")?;
        0
    }
};

if hidden > 1 {                   // <-- summary suppressed when exactly 1 is hidden
    write!(
        f,
        "\n\nThere are {hidden} consequential errors. Check the `out/{}` folder for full details.",
        self.0.uuid
    )?;
}
```

## Problem

In the `else` branch (no non-cascading / root-cause errors — every failure is classified `NodeErrorCause::Cascading`), the code prints **only the single earliest** cascading error and sets `hidden = total_failed - 1`.

The trailing summary that would tell the user "there are more errors, look in `out/<uuid>`" is gated on `hidden > 1`. So when **exactly two** nodes fail and **both** are cascading:

- `total_failed = 2`
- earliest cascading error is printed (1 shown)
- `hidden = 2 - 1 = 1`
- `1 > 1` is `false` → the summary line is **not** printed

Result: the user sees exactly one `Node \`X\` failed: ...` line and has **no way to know** a second node also failed, and **no pointer** to the `out/` folder where the full details live. The second failure is silently swallowed.

For 3+ all-cascading failures the summary *does* appear (`hidden >= 2`), so the boundary is inconsistent: 2 failures → 1 shown + silence; 3 failures → 1 shown + "There are 2 consequential errors...".

The `> 1` threshold looks intended to avoid the ungrammatical "There are 1 consequential errors", but it does so by suppressing the message entirely rather than by handling the singular case — which loses information instead of just fixing grammar.

## Note on the non-cascading branch

In the *first* branch (`!non_cascading.is_empty()`), all non-cascading root-cause errors are printed and `hidden` counts only the cascading *consequences*. There, suppressing the summary when `hidden == 1` is arguably fine (the root cause is already shown). The information-loss bug is specific to the **all-cascading** branch, where only one error is ever printed regardless of how many exist.

Separately, the comment at line 36 (`// try to print earliest non-cascading error`) is now stale — the loop below it prints **all** non-cascading errors, not just the earliest.

## Suggested fix

Show the summary whenever anything is hidden, with singular/plural-aware wording, e.g.:

```rust
if hidden >= 1 {
    let noun = if hidden == 1 { "error" } else { "errors" };
    write!(
        f,
        "\n\n... {hidden} more consequential {noun}. Check the `out/{}` folder for full details.",
        self.0.uuid
    )?;
}
```

(and refresh the stale comment at line 36).

## Suggested test

Add a unit test building a `DataflowResult` with two nodes, both failing with `NodeErrorCause::Cascading`, and assert the formatted output mentions both / points to the `out/` folder rather than showing only one node.

phil-opp:
Fixes #2736

## Problem

`FormatDataflowError::fmt` (`binaries/cli/src/formatting.rs`) can display **only one** failed node while giving the user no indication that other failures occurred and no pointer to the `out/` folder.

In the "all errors are cascading" branch (no non-cascading root-cause error), the code prints only the single earliest cascading error and sets `hidden = total_failed - 1`. The trailing "there are N more, check `out/`" summary is gated on `hidden > 1`, so when **exactly two** nodes fail and both are cascading:

- `total_failed = 2` → 1 shown → `hidden = 1`
- `1 > 1` is `false` → summary suppressed

The user sees one failure line and has no way to know a second node failed. For 3+ all-cascading failures the summary *does* appear, so the boundary was inconsistent (2 → silence, 3 → "There are 2 …").

## Fix

- Lower the threshold to `hidden >= 1` and make the wording singular/plural aware ("There is 1 more consequential error" / "There are N more consequential errors"), so nothing is silently dropped and the grammar stays correct.
- Refresh the stale comment at line 36 that claimed only the earliest non-cascading error is printed (the loop prints all of them).

## Tests

Adds unit tests in `formatting.rs`:
- two all-cascading failures → exactly one inline failure **plus** a singular summary line pointing at `out/`
- three all-cascading failures → plural summary line

Local `cargo test -p dora-cli --lib formatting`, `cargo clippy -p dora-cli -- -D warnings`, and `cargo fmt` all pass.

---

### DIFF
--- a/binaries/cli/src/formatting.rs
+++ b/binaries/cli/src/formatting.rs
@@ -33,7 +33,8 @@ impl std::fmt::Display for FormatDataflowError<'_> {
             .filter(|(_, e)| !matches!(e.cause, NodeErrorCause::Cascading { .. }))
             .collect();
         non_cascading.sort_by_key(|(_, e)| e.timestamp);
-        // try to print earliest non-cascading error
+        // print every non-cascading (root-cause) error; if there are none, fall
+        // back to the earliest cascading error
         let hidden = if !non_cascading.is_empty() {
             let printed = non_cascading.len();
             for (id, err) in non_cascading {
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

I verified against `binaries/cli/src/formatting.rs`. In both branches `hidden` is the count of failures not printed inline (`total_failed - printed`), so lowering the gate to `hidden >= 1` correctly surfaces the previously-dropped two-node all-cascading case (`hidden == 1`) instead of silently swallowing it, and the singular/plural wording (`is`/`are`, `error`/`errors`) matches the boundary. The added tests cover the two- and three-failure cases. The word "more" is accurate since the summary counts errors beyond the one(s) already printed.

---

PR #2740, opened 2026-07-17
fix(core): don't fail ipv6_endpoints_are_bracketed on IPv6-less hosts

### LINKED ISSUE #2737: Test `ipv6_endpoints_are_bracketed` (added in #2722) panics on hosts without IPv6 loopback, breaking `cargo test -p dora-core`

## Summary

The unit test `topics::tests::ipv6_endpoints_are_bracketed`, added two days ago by **PR #2722 / commit `1650eed`** ("fix(daemon): bind zenoh listener on a routable address for multi-machine setups"), **panics on any host whose loopback interface has no IPv6 address** — a very common configuration for Docker containers, CI runners, and locked-down hosts. This makes `cargo test -p dora-core` (and therefore the CI `cargo test --all` job) fail in exactly the "dev containers, locked-down hosts, many CI runners" environments the surrounding code repeatedly says it cares about.

`libraries/core/src/topics.rs`

## Reproduction

On a host with IPv6 disabled on `lo` (e.g. `sysctl net.ipv6.conf.all.disable_ipv6=1`, or many containers):

```
$ cargo test -p dora-core --lib ipv6_endpoints_are_bracketed

thread 'topics::tests::ipv6_endpoints_are_bracketed' panicked at libraries/core/src/topics.rs:660:
reservation on ::1 succeeds: Os { code: 97, kind: Uncategorized,
message: "Address family not supported by protocol" }

test result: FAILED. 243 passed; 1 failed; 0 ignored
```

(`EAFNOSUPPORT` / errno 97 — the kernel has no `AF_INET6` on this host.) Reproduced in a standard IPv6-disabled Linux container.

## Root cause

```rust
// libraries/core/src/topics.rs
#[test]
fn ipv6_endpoints_are_bracketed() {
    let endpoint = reserve_zenoh_endpoint(IpAddr::V6(std::net::Ipv6Addr::LOCALHOST))
        .expect("reservation on ::1 succeeds");   // <-- panics when ::1 can't be bound
    assert!(
        endpoint.starts_with("tcp/[::1]:"),
        "expected a bracketed IPv6 endpoint, got {endpoint}"
    );
}
```

`reserve_zenoh_endpoint` calls `TcpListener::bind((bind, 0))`, which requires the OS to actually support binding `::1`. The test unconditionally `.expect()`s success, so a host without IPv6 loopback turns the intended assertion (that the endpoint string is *bracketed*, i.e. `tcp/[::1]:<port>`) into a hard panic before the assertion is even reached.

This slipped through PR CI because GitHub's `ubuntu-latest` runners do have `::1` on `lo`; the failure only appears on IPv6-disabled hosts.

## Why it matters

- `cargo test -p dora-core` is a documented developer command (see `CLAUDE.md`) and part of the CI `cargo test --all` gate. It now fails for any contributor or alternative/self-hosted CI on an IPv6-disabled host — a false negative that has nothing to do with their change.
- The production code path is **not** affected: `zenoh_bind_address_for` yields an IPv4 `LOCALHOST` for the loopback case and only produces an IPv6 bind address when the coordinator is genuinely reached over IPv6 (which can't happen on an IPv6-less host), and an explicit `--zenoh-listen ::1` is caught by the up-front bindability probe with a clear error. So this is a **test-portability regression only**, not a runtime bug.

## Suggested fix

The test only wants to verify the *string formatting* (bracketing) of an IPv6 endpoint, so it should tolerate a host that can't bind `::1` instead of asserting the bind succeeds. For example:

```rust
#[test]
fn ipv6_endpoints_are_bracketed() {
    match reserve_zenoh_endpoint(IpAddr::V6(std::net::Ipv6Addr::LOCALHOST)) {
        Ok(endpoint) => assert!(
            endpoint.starts_with("tcp/[::1]:"),
            "expected a bracketed IPv6 endpoint, got {endpoint}"
        ),
        // No IPv6 loopback on this host (containers, CI, locked-down hosts):
        // the formatting is exercised elsewhere; nothing to assert here.
        Err(e) if matches!(
            e.kind(),
            std::io::ErrorKind::AddrNotAvailable | std::io::ErrorKind::Unsupported
        ) || e.raw_os_error() == Some(97) => {}
        Err(e) => panic!("unexpected error reserving ::1 endpoint: {e}"),
    }
}
```

Alternatively, split the formatting logic (`format!("tcp/{}", SocketAddr::new(bind, port))`) out of the socket-binding step so the bracketing can be asserted with no bind at all.

## Scope / provenance

- Introduced by: PR #2722 / commit `1650eed` (the `ipv6_endpoints_are_bracketed` test).
- Impact: test-only; breaks `cargo test -p dora-core` on IPv6-disabled hosts. No production data path affected.

---
_Filed automatically by a scheduled Claude review pass. If this is noise or already understood, feel free to close._

phil-opp:
Fixes #2737

## Problem

The unit test `topics::tests::ipv6_endpoints_are_bracketed`, added in #2722, unconditionally `.expect()`s that reserving an endpoint on `::1` succeeds:

```rust
let endpoint = reserve_zenoh_endpoint(IpAddr::V6(std::net::Ipv6Addr::LOCALHOST))
    .expect("reservation on ::1 succeeds");   // <-- panics when ::1 can't be bound
```

`reserve_zenoh_endpoint` calls `TcpListener::bind((bind, 0))`, which requires the OS to actually support binding `::1`. On any host whose loopback has no IPv6 address (many Docker containers, CI runners, `sysctl net.ipv6.conf.all.disable_ipv6=1` hosts), the bind fails with `EAFNOSUPPORT` (errno 97) and the test panics before it even reaches the assertion it cares about. This breaks `cargo test -p dora-core` and the CI `cargo test --all` gate on IPv6-disabled hosts — a false negative unrelated to any contributor's change. It slips past GitHub's `ubuntu-latest` runners because they *do* have `::1` on `lo`.

The production data path is **not** affected (the loopback case binds IPv4, and an explicit `--zenoh-listen ::1` is caught by the up-front bindability probe), so this is a test-portability regression only.

## Fix

Tolerate a host that cannot bind `::1` by matching on the result: assert the `tcp/[::1]:` bracketing whenever `::1` is bindable, and skip (rather than panic) when the bind fails with `AddrNotAvailable` / `Unsupported` / errno 97. Any other error still fails the test.

Local `cargo test -p dora-core --lib ipv6_endpoints_are_bracketed`, `cargo clippy -p dora-core -- -D warnings`, and `cargo fmt` all pass.

---

### DIFF
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -656,12 +656,23 @@ mod tests {
     // (`tcp/::1:7447` is ambiguous).
     #[test]
     fn ipv6_endpoints_are_bracketed() {
-        let endpoint = reserve_zenoh_endpoint(IpAddr::V6(std::net::Ipv6Addr::LOCALHOST))
-            .expect("reservation on ::1 succeeds");
-        assert!(
-            endpoint.starts_with("tcp/[::1]:"),
-            "expected a bracketed IPv6 endpoint, got {endpoint}"
-        );
+        match reserve_zenoh_endpoint(IpAddr::V6(std::net::Ipv6Addr::LOCALHOST)) {
+            Ok(endpoint) => assert!(
+                endpoint.starts_with("tcp/[::1]:"),
+                "expected a bracketed IPv6 endpoint, got {endpoint}"
+            ),
+            // Many CI runners, Docker containers, and locked-down hosts have no
+            // IPv6 address on `lo`, so binding `::1` fails with EAFNOSUPPORT
+            // (`Unsupported`) or `AddrNotAvailable`. That is a property of the
+            // host, not of the endpoint formatting under test, so skip rather
+            // than fail — the production path never binds `::1` on such hosts.
+            Err(e)
+                if matches!(
+                    e.kind(),
+                    std::io::ErrorKind::AddrNotAvailable | std::io::ErrorKind::Unsupported
+                ) || e.raw_os_error() == Some(97) => {}
+            Err(e) => panic!("unexpected error reserving ::1 endpoint: {e}"),
+        }
     }
 
     // The filter behind the routing lookup, tested directly rather than through


phil-opp:

No issues found.

The change is test-only. It converts the `ipv6_endpoints_are_bracketed` test's unconditional `.expect()` into a `match` that still asserts bracketing when `::1` is bindable but skips when the host has no IPv6 loopback (`AddrNotAvailable` / `Unsupported` / raw errno 97 = `EAFNOSUPPORT` on Linux). The production path is untouched. This correctly fixes the portability regression on IPv6-less CI/containers without weakening the assertion where it can run.

---

PR #2741, opened 2026-07-17
fix(node): only count same-machine subscribers in the startup readiness barrier

### LINKED ISSUE #2738: Readiness barrier (#2666) counts unreachable remote subscribers: fixed 5 s per-node startup stall + lost direct-zenoh fast path for local subscribers in multi-machine dataflows

## Summary

The startup readiness barrier added by **PR #2666 / commit `b325512`** ("fix(node): add zenoh data-plane readiness barrier to stop startup message loss") counts **cross-daemon (remote) subscribers** in its "expected subscribers" total, but a remote subscriber's readiness token can **never reach the producer** in a default multi-machine deployment. As a result, in any distributed dataflow:

1. **Every producer node feeding a remote consumer blocks its own `init()` for the full 5 s `ZENOH_OUTPUT_CONNECT_TIMEOUT`** (a synchronous `std::thread::sleep` loop), on every startup.
2. **A local subscriber loses the direct-zenoh fast path** (#1787) for any output that *also* has a remote subscriber — that output is pinned to the daemon-forwarded path permanently.

This is **not** data loss (staying on the daemon path is exactly what correctly delivers to remote subscribers), but it is a real, fixed startup-latency regression and a fast-path regression, and the code's own justifying comments assert the opposite of what happens.

Files: `apis/rust/node/src/node/mod.rs`, `libraries/core/src/descriptor/mod.rs`

## Mechanism (verified by reading the code)

- `DescriptorExt::output_subscriber_count` (`libraries/core/src/descriptor/mod.rs`) iterates **every** node in the resolved descriptor and excludes only *dynamic* nodes — there is **no locality filter**. So `expected` includes subscribers that live on other daemons/machines.

- A subscriber declares its readiness liveliness token on its **own** zenoh session (`apis/rust/node/src/event_stream/mod.rs`). Since zenoh 1.9 peers do not relay (the premise of #2716), and dora only wires **loopback, same-machine** node↔node links (`plan_zenoh_peering` plans local nodes only) with multicast scouting disabled, a remote subscriber's token is known only to its own daemon and never propagates to the producer. Cross-machine **data** is delivered by explicit daemon-to-daemon forwarding (`send_to_remote_receivers`), not by zenoh peer gossip — so the liveliness tokens and the direct data path never span machines.

- `warm_up_direct_outputs` (`apis/rust/node/src/node/mod.rs:1365`) loops until **all** outputs are ready **or** the 5 s deadline:
  ```rust
  const ZENOH_OUTPUT_CONNECT_TIMEOUT: Duration = Duration::from_secs(5); // node/mod.rs:129
  let deadline = Instant::now() + ZENOH_OUTPUT_CONNECT_TIMEOUT;
  loop {
      let all_ready = outputs.iter().all(|o| self.ensure_output_ready(o));
      if all_ready { break; }
      if Instant::now() >= deadline { … break; }
      std::thread::sleep(ZENOH_OUTPUT_CONNECT_POLL_INTERVAL);
  }
  ```
  Because any output with a remote subscriber can never reach `count == expected`, `all_ready` is never true and the loop always runs the **full 5 s**.

- `ensure_output_ready` returns `false` while `count < expected`, so `zenoh_publish` returns `NotPublished` and every send for that output takes the daemon path — including for co-located local subscribers of that same output.

## Failure scenario

Three machines, pipeline `A → B → C`, one node per daemon:
- A's output `A/out` has one subscriber (B) on another daemon → `expected = 1`, remote token never arrives → A sleeps the full **5 s** in `init()` before user code runs. Same for B.
- Mixed fan-out variant: `A/out` feeds both a local node `A2` (same daemon) and remote `B`. `expected = 2`, only A2's token ever arrives → the output never switches to direct zenoh, so **A2 is served over the slower daemon path forever**, defeating #1787 for the local link.

Startup latency grows by ~5 s per producer stage; before #2666 the corresponding `ZENOH_FIRST_PUBLISH_MATCH_TIMEOUT` was 200 ms.

## Inaccurate justifying comments

These describe a single-machine world and are false for the multi-machine case the surrounding code is otherwise careful about:

- `node/mod.rs:1356-1357`: "merely waits out zenoh route propagation between already-declared endpoints — **normally milliseconds**." Remote tokens are never delivered, so it is a hard 5 s.
- `node/mod.rs:122-125`: "this only waits out the residual zenoh propagation between two already-declared endpoints, so it is **normally sub-second**." Same issue.

(`node/mod.rs:1360` does list "an unreachable remote subscriber" as a timeout cause, but frames it as a rare edge rather than *every* cross-daemon edge.)

## Suggested fix

Make the barrier only wait for subscribers it can actually observe:

- Restrict `output_subscriber_count` (or add a local-only variant) to **local, non-dynamic** subscribers — the ones whose readiness tokens can reach this producer over the same-machine zenoh links dora actually wires. Remote subscribers are served by the daemon path regardless, so they should not gate the local fast-path switch. This removes both the 5 s stall and the local-fast-path denial in one change.
- Alternatively, track expected/observed **per subscriber locality**, so an output flips to direct zenoh for its local subscribers as soon as those are wired while remote subscribers keep using the daemon path independently.
- Either way, correct the "normally milliseconds / sub-second" comments.

## Scope / provenance

- Introduced by: PR #2666 / commit `b325512` (readiness barrier + `output_subscriber_count`).
- Impact: multi-machine deployments only; fixed +5 s startup per producer stage and loss of the direct-zenoh fast path for local subscribers of mixed local/remote outputs. **Not** data loss.
- Interaction note: unrelated to the #2722 routable-endpoint change, which integrates cleanly with #2716.

---
_Filed automatically by a scheduled Claude review pass. If this is an accepted tradeoff, feel free to close — but the "normally milliseconds" comments are worth correcting regardless._

phil-opp:
Fixes #2738

## Problem

The startup readiness barrier added in #2666 counts **cross-daemon (remote) subscribers** in its "expected subscribers" total, but a remote subscriber's readiness token can never reach the producer in a default multi-machine deployment:

- `DescriptorExt::output_subscriber_count` iterated every node with no locality filter, so `expected` included subscribers on other daemons/machines.
- A subscriber declares its readiness liveliness token on its **own** daemon's zenoh session. Since zenoh 1.9 peers don't relay (#2716) and dora only wires loopback, same-machine node↔node links, a remote token never propagates to the producer. Cross-machine data is delivered by explicit daemon-to-daemon forwarding, not zenoh peer gossip.

Consequences in any distributed dataflow:

1. **Every producer feeding a remote consumer blocked its `init()` for the full 5s `ZENOH_OUTPUT_CONNECT_TIMEOUT`** on every startup, because `all_ready` could never become true for an output whose `count` can never reach `expected`.
2. **A local subscriber lost the direct-zenoh fast path (#1787)** for any output that *also* has a remote subscriber — that output stayed pinned to the daemon-forwarded path forever, even for its co-located subscriber.

This is not data loss (the daemon path delivers correctly), but it is a fixed +5s startup-latency regression per producer stage and a fast-path regression, and the code's own "normally milliseconds / sub-second" comments asserted the opposite of what happened.

## Fix

- Rename `output_subscriber_count` → `local_output_subscriber_count` and skip subscribers whose deploy machine differs from the producer's (new `node_machine` helper comparing `deploy.machine`).
- Same-machine subscribers — including the single-machine case where every machine is `None` — are still counted, so #2666's "don't drop startup messages" guarantee is preserved for exactly the links whose readiness can actually be observed. Remote subscribers keep using the daemon path independently.
- Update the sole caller (`declare_output_readiness`) and correct the stale "normally milliseconds / sub-second" comments and the "unreachable remote subscriber" timeout example in `warm_up_direct_outputs`.

## Tests

- Existing `output_subscriber_count_*` tests (single-machine, all machines `None`) are unchanged in behavior — they still count 2 / 1 / 0 — confirming no single-machine regression.
- New `local_output_subscriber_count_excludes_remote_subscribers`: a `source` on machine `a` with one co-located subscriber (machine `a`) and one remote subscriber (machine `b`) counts **1**, not 2.

Local `cargo test -p dora-core --lib`, `cargo clippy -p dora-core -p dora-node-api -- -D warnings`, and `cargo fmt` all pass.

---

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -170,6 +170,12 @@ pub struct DoraNode {
     /// Per-output startup readiness for the direct zenoh data plane (lazily
     /// created on first send). See [`ZenohOutputReadiness`].
     zenoh_output_readiness: HashMap<DataId, ZenohOutputReadiness>,
+    /// Outputs with at least one cross-machine subscriber. These are pinned to
+    /// the reliable daemon path (which forwards them to remote daemons) and never
+    /// switch to the direct node-to-node zenoh path, whose publishes the daemon
+    /// does not forward — see [`DoraNode::zenoh_publish`] and
+    /// [`DescriptorExt::outputs_with_remote_subscribers`] (dora-rs/dora#2738).
+    zenoh_daemon_only_outputs: BTreeSet<DataId>,
 
     dataflow_descriptor: serde_yaml::Result<Descriptor>,
     warned_unknown_output: BTreeSet<DataId>,
@@ -744,6 +750,7 @@ impl DoraNode {
             zenoh_schema_state: HashMap::new(),
             zenoh_zero_copy_threshold,
             zenoh_output_readiness: HashMap::new(),
+            zenoh_daemon_only_outputs: BTreeSet::new(),
             dataflow_descriptor: serde_yaml::from_value(dataflow_descriptor),
             warned_unknown_output: BTreeSet::new(),
             interactive: false,
@@ -777,6 +784,23 @@ impl DoraNode {
                         );
                     }
                 }
+            }
+        }
+
+        // Pin any output with a cross-machine subscriber to the reliable daemon
+        // path. Cross-machine data is delivered by daemon-to-daemon forwarding,
+        // which is fed only by the daemon-path send; a direct node-to-node zenoh
+        // publish bypasses the daemon and never reaches a remote consumer. Without
+        // this, an output whose co-located subscriber releases the readiness
+        // barrier would switch to direct zenoh and silently drop every message to
+        // its remote consumers (dora-rs/dora#2738).
+        if let Ok(descriptor) = &node.dataflow_descriptor {
+            match descriptor.outputs_with_remote_subscribers(&node.id) {
+                Ok(remote) => node.zenoh_daemon_only_outputs = remote,
+                Err(e) => tracing::warn!(
+                    "failed to determine cross-machine outputs ({e}); \
+                     outputs feeding remote consumers may use the direct zenoh path"
+                ),
             }
         }
 
@@ -1358,7 +1382,10 @@ impl DoraNode {
     /// The barrier only waits for *same-machine* subscribers, whose readiness
     /// tokens can actually reach this producer; remote subscribers are served by
     /// the daemon path and never gate the switch (see
-    /// [`DescriptorExt::local_output_subscriber_count`]).
+    /// [`DescriptorExt::local_output_subscriber_count`]). Outputs that have any
+    /// cross-machine subscriber are excluded entirely — they stay pinned to the
+    /// daemon path (see [`Self::zenoh_daemon_only_outputs`]), so the barrier
+    /// neither waits for them nor stalls startup on a token that can never arrive.
     ///
     /// Fail-safe: outputs that don't connect within
     /// [`ZENOH_OUTPUT_CONNECT_TIMEOUT`] (e.g. a local zenoh data plane that can't
@@ -1369,7 +1396,16 @@ impl DoraNode {
         if self.zenoh_session.is_none() {
             return;
         }
-        let outputs: Vec<DataId> = self.node_config.outputs.iter().cloned().collect();
+        // Skip outputs pinned to the daemon path: they never switch to direct
+        // zenoh, so declaring publishers or waiting on their (unreachable) remote
+        // readiness tokens would only stall startup for no benefit.
+        let outputs: Vec<DataId> = self
+            .node_config
+            .outputs
+            .iter()
+            .filter(|o| !self.zenoh_daemon_only_outputs.contains(*o))
+            .cloned()
+            .collect();
         if outputs.is_empty() {
             return;
         }
@@ -1420,6 +1456,16 @@ impl DoraNode {
         finalized: FinalizedSample,
     ) -> eyre::Result<PublishOutcome> {
         use zenoh::Wait;
+
+        // Outputs with a cross-machine subscriber must never take the direct
+        // node-to-node zenoh path: a remote consumer is reached only by daemon-to-
+        // daemon forwarding, which the daemon feeds from this daemon-path send. A
+        // direct zenoh publish bypasses the daemon and would be silently dropped
+        // for every remote consumer (dora-rs/dora#2738). Co-located subscribers of
+        // the same output are served correctly by the daemon path too.
+        if self.zenoh_daemon_only_outputs.contains(output_id) {
+            return Ok(PublishOutcome::NotPublished(finalized));
+        }
 
         // Every failure *before* the payload is moved into `put` returns the
         // sample as `NotPublished` so the caller can still deliver it via the
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -1236,7 +1236,7 @@ impl DoraNode {
         // which just means "no barrier" — the pre-existing behaviour.
         let expected = match &self.dataflow_descriptor {
             Ok(descriptor) => descriptor
-                .output_subscriber_count(&self.id, output_id)
+                .local_output_subscriber_count(&self.id, output_id)
                 .unwrap_or_else(|e| {
                     tracing::warn!(
                         output = %output_id,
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found.

I traced the logic in `libraries/core/src/descriptor/mod.rs`. `local_output_subscriber_count` now computes `source_machine` via `node_machine(source_node)` and skips subscribers whose `node_machine` differs, which correctly preserves the single-machine case (all `None`, so all count) and #2666's guarantee for co-located links. The rename is complete — the only two references (`descriptor/mod.rs` and `apis/rust/node/src/node/mod.rs:1239`) are both updated, so nothing dangles. The added regression test covers the mixed local/remote case. Fallback when `source_node` isn't found yields `source_machine = None`, which is a sensible default.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review; no human has vetted it. Treat it as advisory and verify before acting.

Tracing the data-plane consequences of this change, I think it introduces **silent cross-machine data loss** for any output that has a remote subscriber, under the default daemon-managed multi-machine deployment.

`local_output_subscriber_count` (now filtering out subscribers where `node_machine(node) != source_machine`) feeds `expected` in the readiness barrier — `declare_output_readiness` stores it as `ZenohOutputReadiness.expected` (`apis/rust/node/src/node/mod.rs:1237`, `:1287`), which `ensure_output_ready` reads (`:1175`). That barrier gates the switch from the daemon path to the direct-zenoh fast path. The problem is what "switched to zenoh" does to *remote* delivery:

- Once ready, `zenoh_publish` puts over zenoh and returns `Delivery::Zenoh`, whose arm calls **only** `report_output_sent(output_id, metadata)` — metadata, **no payload** to the daemon (`node/mod.rs:1064`, `:1081-1089`). The payload-bearing `send_message` runs only on `Delivery::Daemon` (`:1107`).
- The daemon's `OutputSent` handler → `output_sent()` calls only `note_output_sent_to_local_receivers` (`binaries/daemon/src/lib.rs:3967`, `:5262`); it takes no payload and never calls `send_to_remote_receivers`.
- Cross-machine forwarding (`send_to_remote_receivers`, `:3987`) is reached **only** from `send_out()` (`:3898`), i.e. the `SendOut`/`Delivery::Daemon` path.
- Node→node zenoh links are loopback/same-machine only: `plan_zenoh_peering` reserves loopback listeners for local nodes only (`spawn/spawner.rs:144`, `:160-163`), and the daemon subscribes to node zenoh output topics only under `enable_debug_inspection`, which is relayed to coordinator watchers and never re-delivered to receivers (`lib.rs:3080-3092`).

So after this change, for an output with a remote subscriber:
- **remote subscriber only** → `expected == 0` → `mark_output_ready` fires immediately → zenoh path → the remote node receives nothing from the first message on.
- **local + remote subscriber** → `expected == N_local`, satisfied by the local sub's token → latches ready → zenoh path → the remote node silently stops receiving.

Before this change, the remote subscriber was counted; its readiness token never arrives (as the PR notes), so the output stayed on the `Delivery::Daemon` path and the daemon forwarded to the remote machine (via `open_external_mappings` → `send_to_remote_receivers`). That "never switches" behavior is the startup-latency bug #2738 describes, but it's also what keeps cross-machine delivery working, because the daemon can only forward payloads it receives via `SendOut`. The description's premise — "cross-machine data is delivered by explicit daemon-to-daemon forwarding regardless" — doesn't hold once the fast-path switch stops routing the payload through the daemon.

(The one case where this is safe is a user-supplied zenoh router config via `ZENOH_CONFIG_PATH`, where node→node zenoh can actually cross machines. But the change keys on `_unstable_deploy: machine` — the default daemon-managed deployment — where node listeners are loopback and cross-machine delivery depends on daemon forwarding, so it isn't restricted to the router case.)

A correct fix for #2738 probably needs the fast-path predicate to be "ready **and** the output has no remote subscribers" (keeping mixed/remote outputs on the daemon path), or to have the node publish to both zenoh (local) and the daemon (remote). The added test only asserts the count, so it wouldn't catch this — an end-to-end test with a producer feeding a subscriber on another machine would be worth adding.

---

phil-opp:

Re-review of the new commit `d9475eb` ("keep outputs with a remote subscriber on the daemon path"). This directly addresses the silent cross-machine data-loss concern I raised on the previous revision, and the fix looks correct:

- `DescriptorExt::outputs_with_remote_subscribers` computes, for a producer, the set of its outputs that have at least one subscriber on a different machine (skipping dynamic nodes, mirroring the barrier count).
- `DoraNode` stores that set in `zenoh_daemon_only_outputs` at construction and `zenoh_publish` now returns `NotPublished` for any output in it, so it stays on the reliable daemon path (which forwards to remote daemons) and never takes the direct node↔node zenoh path.
- `warm_up_direct_outputs` filters those outputs out, so `init()` no longer stalls on a readiness token that can never arrive.

Remote-only and mixed local+remote outputs are now pinned to the daemon path, so cross-machine delivery is preserved. The one accepted tradeoff (a mixed output's co-located subscriber also stays on the daemon path rather than the fast path) is acknowledged in the commit message as a follow-up and is not a correctness issue. The added `outputs_with_remote_subscribers_*` tests cover the mixed/remote-only/local-only and single-machine cases.

No new issues found.

---

phil-opp:
Superseded by #2806: the daemon-computed output routing subsumes the remote pin, and the counting half targets code #2806 removes.

PR #2744, opened 2026-07-17
fix(download): reject error statuses instead of saving the error page

phil-opp:
The Content-Disposition fallback ran Path::file_name() over the whole URL
string, which includes the query and fragment, so a URL like
  https://example.com/model.bin?token=abc
produced the filename `model.bin?token=abc`. Derive it from the parsed URL
path segments instead.

Also reject 4xx/5xx before reading the body -- previously an error page was
written out as if it were the requested artifact.

Supersedes #2454 and #2465, which both predate the #2485 get_filename rewrite.

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -88,6 +88,12 @@ where
         .await
         .wrap_err_with(|| format!("failed to request operator from `{url}`"))?;
 
+    // Reject 4xx/5xx before reading the body: otherwise the error page is
+    // written out as if it were the requested artifact.
+    let response = response
+        .error_for_status()
+        .wrap_err_with(|| format!("server returned an error status for `{url}`"))?;
+
     let filename = get_filename(&response).context("Could not find a filename")?;
     let bytes = response
         .bytes()
@@ -160,6 +166,16 @@ mod tests {
         assert_eq!(
             name_from("https://example.com/path/model.bin#section"),
             Some("model.bin".to_string())
+        );
+    }
+
+    #[test]
+    fn url_filename_normalises_traversal() {
+        // The URL parser resolves `..` segments before we ever see them, so a
+        // traversal attempt cannot escape via the fallback path.
+        assert_eq!(
+            name_from("https://example.com/a/../../etc/passwd"),
+            Some("passwd".to_string())
         );
     }
 


phil-opp:

No issues found.

I traced `libraries/extensions/download/src/lib.rs`. `filename_from_url` correctly uses `path_segments().next_back()` so the query/fragment are excluded (the whole point vs. the old `Path::file_name(url.as_str())`), and the empty-segment filter handles trailing-slash URLs by falling through to the "Could not find a filename" error rather than producing an empty name. `error_for_status()` is taken by value before `response.bytes()`, so a 4xx/5xx page is rejected before it can be written out as the artifact. Redirects resolve via `response.url()` to the final URL, so the fallback name is derived from the right place. The added tests cover query/fragment stripping, trailing slash, and `..` normalization.

---

phil-opp:
@claude fix formatting

phil-opp:

The two commits pushed since my earlier review (`7164835`, `c9efed9`) are formatting-only: they re-wrap/collapse the `url_fallback_tests` `assert_eq!` lines to satisfy `cargo fmt --check`. No change to `filename_from_url` or the `error_for_status()` logic. No new issues.

---

phil-opp:

The diff itself checks out — `filename_from_url` correctly strips query/fragment (it operates on parsed path segments), returns `None` on a trailing slash, and `..` is normalized at parse time plus re-sanitized downstream by `get_filename`; taking `error_for_status()` before `bytes()` correctly rejects 4xx/5xx before an error body can be written as the artifact. No new security or correctness issue.

One merge-context heads-up though: the URL-fallback half overlaps with #2729 (`02888c0`, merged Jul 22), which already derives the fallback via `url.path_segments()` on `main` — that's likely the source of the `dirty` state. The genuinely unique piece not on `main` is the `error_for_status()` rejection, which still seems worth having.

Watch out on rebase: the two implementations differ on trailing-slash URLs. This PR's `filter(|seg| !seg.is_empty())` returns `None` for `https://host/dir/`, whereas the merged #2729 uses `rfind(|s| !s.is_empty())` and returns `Some("dir")`. Worth deciding which behavior is intended so the resolution doesn't quietly flip what just shipped — might be simplest to rebase down to just the `error_for_status()` change on top of #2729.

---

phil-opp:
Rebased onto main (c2dc8cec1). The URL-fallback-filename half of this PR
landed independently as #2729, so it has been dropped here — what remains is
the `error_for_status()` fix (a 404/500 error page was written to disk and then
dlopen-ed/exec-ed as if it were the requested artifact) plus one
fallback-filename traversal test that #2729 left uncovered.

phil-opp:
The Jul 27 rebase dropped the `filename_from_url` half (now landed as #2729), so the diff is reduced to the `error_for_status()` fix plus one fallback-filename test. `error_for_status()` is taken by value before `response.bytes()`, so a 4xx/5xx page is rejected before it can be written out as the artifact, and the `url_filename_normalises_traversal` test correctly asserts the URL parser collapses `..` so the fallback can't escape. The trailing-slash divergence flagged in the earlier review no longer applies. No new issues — the latest state looks safe.

---
🤖 _This is a fully automated review by Claude — no human has reviewed these findings. Please verify before relying on them._

---

PR #2745, opened 2026-07-17
fix(examples): bump rust-dataflow-git pin past the Arrow IPC wire break

phil-opp:
#2366 dropped the `ArrowTypeInfo` sidecar from `Metadata` and took the
payload wire format to v1. The example still pinned 1.0.0-rc1 (May 29) —
a v0 node — so its nodes desynced against a current daemon and died with
`tag for enum is not valid, found 150`.

The example is working as designed: its header comment asks for a rev
bump when a format-breaking change lands. No commit since #2366 has
touched the wire format, so main's HEAD is format-current.

This fixes the Examples (windows-latest) nightly job, red since
2026-07-09. Worth noting the desync fires on all three platforms — but
Linux/macOS still report green, because the wedged node is killed by
SIGTERM at the grace period and the daemon treats Signal(15) during a
planned stop as clean. Windows has no signals, so the hard kill surfaces
as ExitCode(1) and fails the job.

Refs #2742

### DIFF
--- a/examples/rust-dataflow-git/dataflow.yml
+++ b/examples/rust-dataflow-git/dataflow.yml
@@ -3,10 +3,15 @@
 # without needing a release. Update `rev:` when a message-format-breaking
 # change lands on main — otherwise the CI job catches the mismatch and
 # signals a compatibility break, which is the whole point of this test.
+#
+# Currently pinned past #2366, which dropped the `ArrowTypeInfo` sidecar
+# from `Metadata` and took the payload wire format to v1. The previous pin
+# (1.0.0-rc1) predated that, so its nodes desynced against a current daemon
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

I verified the rev bump against the local repo: the new pin `07ed3be` is a real ancestor of `main`, it lands after `81e5efe` (#2366, the commit that dropped the `ArrowTypeInfo` sidecar and moved the payload wire format to v1), so it is format-current. The only four commits between the new pin and current `main` HEAD (#2587, #2707, #2743, #2740) touch zenoh teardown, a GIL token, CircleCI config, and an IPv6 test — none touch the wire format, so the pinned node stays compatible with a HEAD-built daemon. Both consumer nodes and the sink are bumped consistently to the same rev.

---

PR #2747, opened 2026-07-17
fix(node-api): stop wait_for_correlation livelocking on buffered events

phil-opp:
## Issue

The pattern-aware wait loop behind `EventStream::recv_service_response` / `recv_action_result` (`wait_for_correlation`) pumps the event stream while it searches for the correlated response, buffering every *non-matching* event into `pending_passthrough` so the caller's own event loop can still see those events later.

The loop pumped the stream via `recv_async` — but `recv_async` drains `pending_passthrough` **first**:

```rust
pub async fn recv_async(&mut self) -> Option<Event> {
    if let Some(event) = self.pending_passthrough.pop_front() {
        return Some(event);   // <-- served before any new stream read
    }
    ...
}
```

So the moment the wait loop buffered one non-matching event, the next `recv_async` handed that **same** event straight back, the classifier re-buffered it, and the loop spun on it forever — never reading the awaited response off the receiver.

**Failure scenario:** a client calls `recv_service_response(...)`; a regular input `U` arrives before the matching response `M` (a very common interleaving — normal inputs and service replies share one stream).
- Iter 1: read `U` → classify `Passthrough` → `push_back(U)`.
- Iter 2: `recv_async` pops `U` back out → `Passthrough` → `push_back(U)`.
- Iter 3+: same, forever — the loop never awaits, pins a CPU core, never reads `M`, and returns `PatternError::Timeout` once the deadline elapses.

Every service/action wait preceded by even one unrelated event times out.

## Fix

Split the stream-reading path into a private `recv_from_stream`, which reads from the scheduler/receiver **without** draining `pending_passthrough`, and pump the wait loop through it. `recv_async` keeps its passthrough drain for the caller (`pop_front` early-return, then delegates to `recv_from_stream`). Non-matching events are still buffered and replayed to the caller after the wait returns — that behavior is unchanged.

## Validation

- Added a regression test `recv_service_response_matches_after_non_matching_event` using the integration-testing FIFO path: it delivers a non-matching `sensor` input before the correlated `response`, then asserts the wait returns the response and that `sensor` is still replayed afterward.
- Verified **RED/GREEN**: the test times out (5.37s) on the pre-fix code and passes instantly on the fix.
- `cargo test -p dora-node-api --lib` — 142 passed.
- `cargo clippy -p dora-node-api --lib -- -D warnings` — clean. `cargo fmt --all -- --check` — clean.

---

---

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -716,6 +716,23 @@ impl EventStream {
         if let Some(event) = self.pending_passthrough.pop_front() {
             return Some(event);
         }
+        self.recv_from_stream().await
+    }
+
+    /// Receive the next event straight from the scheduler/receiver,
+    /// **without** draining `pending_passthrough` first.
+    ///
+    /// The pattern-aware wait loop ([`wait_for_correlation`](Self::wait_for_correlation))
+    /// buffers every non-matching event into `pending_passthrough` itself. If
+    /// it pumped the stream through [`recv_async`](Self::recv_async), that
+    /// drain would immediately hand the just-buffered event straight back, the
+    /// classifier would re-buffer it, and the loop would spin on the same event
+    /// forever — never reading the awaited response off the receiver — until it
+    /// hit its deadline and wrongly reported a timeout (while pinning a CPU
+    /// core). Reading through this bypass keeps the buffered events reserved for
+    /// the caller's own `recv`/`recv_async` while the wait loop makes real
+    /// progress.
+    async fn recv_from_stream(&mut self) -> Option<Event> {
         // Close the stream after a Stop event: the daemon thread has
         // already dropped its sender, but zenoh subscriber threads
         // hold clones that would otherwise keep `receiver` open.
@@ -1134,7 +1151,11 @@ impl EventStream {
             if remaining.is_zero() {
                 return Err(PatternError::Timeout);
             }
-            let event = match select(Delay::new(remaining), pin!(self.recv_async())).await {
+            // Pump the stream via `recv_from_stream`, NOT `recv_async`: the
+            // latter drains `pending_passthrough` first, which would re-hand us
+            // the very events we buffer below and livelock the loop (see
+            // `recv_from_stream`'s docs).
+            let event = match select(Delay::new(remaining), pin!(self.recv_from_stream())).await {
                 Either::Left((_elapsed, _)) => return Err(PatternError::Timeout),
                 Either::Right((None, _)) => return Err(PatternError::StreamEnded),
                 Either::Right((Some(e), _)) => e,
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Splitting the stream read into `recv_from_stream` (which does **not** drain `pending_passthrough`) for the wait loop, while `recv_async` keeps draining the buffer for normal callers, correctly breaks the re-buffer/re-serve livelock: the wait loop now makes real progress while buffered non-matching events stay reserved for the caller. The regression test covers both the correlation match and the later replay of the buffered event.

---

phil-opp:
Re-reviewed the fix against the code on `refs/pull/2747/head`. The livelock fix itself is correct — the wait loop making progress via `recv_from_stream` instead of `recv_async` breaks the re-buffer/re-serve spin. But bypassing `pending_passthrough` entirely introduces a new gap that I think is worth addressing before merge:

**A correlated response that is already sitting in `pending_passthrough` will never be matched by a later wait, causing a spurious `Timeout`.**

`wait_for_correlation` (`apis/rust/node/src/event_stream/mod.rs:1138`) now reads *only* through `recv_from_stream` (`:1158`), which never inspects `pending_passthrough`. That buffer is only drained on the caller's `recv`/`recv_async` path (`:716`). So any event already buffered there is invisible to a subsequent correlation wait.

Concrete scenario with two overlapping/pipelined service calls sharing one stream:
```rust
let ra = events.recv_service_response("A", &srv, t).await?; // during this wait,
                                                            // B's response arrives ->
                                                            // classified Passthrough ->
                                                            // push_back into pending_passthrough
let rb = events.recv_service_response("B", &srv, t).await?; // B's response is in the buffer,
                                                            // wait reads only from the stream ->
                                                            // never sees it -> Timeout
```
B's buffered response is later handed to the caller's *main* `recv()` loop as an ordinary input, not to `recv_service_response("B")`, so the correlation call fails even though the response did arrive.

This is a behavioral regression relative to the pre-fix code for that case: the old `recv_async`-based loop drained `pending_passthrough` first, so it *would* pop a buffered matching response and classify it `Match`. The livelock it suffered only occurred when the buffered event was non-matching. The current patch trades "livelock on a buffered non-match" for "silently miss a buffered match."

A robust fix is to have the wait loop first scan `pending_passthrough` for a match (returning it, and leaving non-matching entries in place / re-buffered in order) before falling through to `recv_from_stream`, rather than ignoring the buffer entirely. If out-of-order / pipelined correlation waits are considered out of scope, it'd be worth a doc note that a correlation wait only matches events read after it starts.

Happy to be corrected if there's an invariant guaranteeing `pending_passthrough` is always empty at the start of a wait — I didn't find one.

---
🤖 _Automated review by Claude — fully automated, no human in the loop. May contain mistakes; please verify before acting._

---

phil-opp:

The new commit `27dcb3f` addresses the buffered-match gap I raised on 2026-07-19. `wait_for_correlation` now scans `pending_passthrough` for a match before entering the read loop, extracts it in place (preserving the order of the remaining buffered events for the caller's own `recv()`), and only pulls an actual `Match` — a buffered `Stop`/`NodeRestarted` is correctly left untouched. A single pre-loop scan is sufficient here, since any event buffered *during* the current wait is non-matching to the current needle by construction.

The added `recv_service_response_matches_buffered_response_from_prior_wait` test genuinely exercises the pipelined case (`resp-2` arriving before `resp-1`, then matched from the buffer by the later wait) and would fail on the pre-`27dcb3f` code. No new issues found; the latest commit is safe.

---

PR #2748, opened 2026-07-17
fix(cli): reject `dora daemon --worker-threads 0` instead of panicking

phil-opp:
## Issue

`dora daemon --worker-threads 0` aborts the daemon with a raw Rust panic:

```
if let Some(threads) = self.worker_threads {
    builder.worker_threads(threads);   // tokio asserts val > 0
}
```

`tokio::runtime::Builder::worker_threads` does `assert!(val > 0, "Worker threads cannot be set to 0")`, so a user typo produces an ugly panic + backtrace instead of a clean CLI usage error. The flag had no range validation, unlike every other bounded numeric flag in the CLI (`--refresh-interval`, `--duration`, `--count`, …), which consistently use `range(1..)`.

## Fix

`usize` is not one of the ranged integer types clap's `value_parser!` macro supports (the sibling flags use `u64`/`u32`), so `.range(1..)` doesn't compile for it. This adds a small `parse_worker_threads` value parser that rejects `0` with a normal usage error and keeps the field type `usize`, so the downstream `builder.worker_threads(threads)` call is unchanged.

```
fn parse_worker_threads(s: &str) -> Result<usize, String> {
    match s.parse::<usize>() {
        Ok(0) => Err("worker threads must be at least 1".to_string()),
        Ok(n) => Ok(n),
        Err(err) => Err(err.to_string()),
    }
}
```

## Validation

- Added parser tests: `--worker-threads 4` / `1` parse; `--worker-threads 0` is rejected.
- `cargo test -p dora-cli --lib` — 253 passed.
- `cargo clippy -p dora-cli --lib -- -D warnings` — clean. `cargo fmt --all -- --check` — clean.

---

---

### DIFF
--- a/binaries/cli/src/command/daemon.rs
+++ b/binaries/cli/src/command/daemon.rs
@@ -18,6 +18,19 @@ use std::{
 use tokio::runtime::Builder;
 use tracing::level_filters::LevelFilter;
 
+/// Parse `--worker-threads`, rejecting 0.
+///
+/// `tokio::runtime::Builder::worker_threads` asserts `val > 0` and would
+/// otherwise abort the daemon with a raw panic ("Worker threads cannot be set
+/// to 0"). Validating here turns that into a normal clap usage error.
+fn parse_worker_threads(s: &str) -> Result<usize, String> {
+    match s.parse::<usize>() {
+        Ok(0) => Err("worker threads must be at least 1".to_string()),
+        Ok(n) => Ok(n),
+        Err(err) => Err(err.to_string()),
+    }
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. The `value_parser` is applied to the inner `usize` and the flag stays optional, so `dora daemon --worker-threads 0` now returns a clean clap usage error instead of tripping tokio's `Builder::worker_threads` `val > 0` assertion (raw panic), and non-numeric input still surfaces as a parse error. Tests match the behavior.

---

PR #2751, opened 2026-07-17
fix(tracing): don't truncate trace-context values at the second colon

phil-opp:
## Issue

`deserialize_to_hashmap` reconstructs the propagated OpenTelemetry trace context on the receiving node. It parsed each `key:value` entry like this:

```rust
for s in string_context.split(';') {
    let mut values = s.split(':');
    let key = values.next().unwrap();
    let value = values.next().unwrap_or("");   // everything after the 2nd ':' is dropped
    map.insert(key, value);
}
```

`serialize_context` only ever emits the W3C keys `traceparent` and `tracestate`. `traceparent` values contain no `:`, so the common path is fine — but a **`tracestate`** value may legitimately contain `:` (the W3C spec allows 0x20–0x7E except `,` and `=`). For a value like `vendor=a:b`, deserialization kept only `vendor=a`, silently corrupting the propagated sampling/vendor state across node boundaries.

## Fix

Split on the **first** `:` only:

```rust
let mut values = s.splitn(2, ':');
```

The value now retains everything after the first colon. `traceparent` is unaffected. This is backward-compatible: previously-serialized contexts without problematic characters parse identically (and ones *with* them were already broken).

## Validation

- Added `deserialize_keeps_colons_in_value` (a colon-bearing `tracestate` round-trips intact) and `deserialize_plain_traceparent_round_trips`.
- `cargo test -p dora-tracing --lib` — 5 passed.
- `cargo clippy -p dora-tracing --all-targets -- -D warnings` — clean. `cargo fmt --all -- --check` — clean.

### Note on remaining scope

This fixes the `:`-in-value truncation, the realistic case. The entry delimiter is still `;`, which is also technically allowed in a `tracestate` value; a fully delimiter-safe encoding (percent-encoding/JSON) would be a larger, wire-format-changing change and is deliberately out of scope here.

---

---

### DIFF
--- a/libraries/extensions/telemetry/tracing/src/telemetry.rs
+++ b/libraries/extensions/telemetry/tracing/src/telemetry.rs
@@ -79,10 +79,34 @@ pub fn deserialize_context(string_context: &str) -> Context {
 pub fn deserialize_to_hashmap(string_context: &str) -> HashMap<&str, &str> {
     let mut map = HashMap::new();
     for s in string_context.split(';') {
-        let mut values = s.split(':');
+        // Split on the FIRST ':' only. The value (e.g. a W3C `tracestate`
+        // like `vendor=a:b`) may legitimately contain ':', and a plain
+        // `split(':')` would drop everything after the second colon, silently
+        // corrupting the propagated context.
+        let mut values = s.splitn(2, ':');
         let key = values.next().unwrap();
         let value = values.next().unwrap_or("");
         map.insert(key, value);
     }
     map
+}
+
+#[cfg(test)]
+mod tests {
+    use super::deserialize_to_hashmap;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found in the change itself — switching the value split to `splitn(2, ':')` correctly stops truncating W3C `tracestate` values at the second colon, and the tests cover it.

One follow-up (pre-existing, out of scope for this PR): the **entry** separator is still `split(';')`, and a `tracestate` value may legitimately contain `;` (allowed value chars are `0x20`–`0x7E` except `,` and `=`). So a value containing `;` would still be split across entries and corrupted — the same class of bug this PR fixes for `:`. Worth a separate fix, and on the serialize side too if that's under your control.

---

phil-opp:
No longer needed after #2579

PR #2754, opened 2026-07-18
fix(core): reject zero `health_check_interval` before it panics the daemon

### LINKED ISSUE #2752: Daemon panics on `health_check_interval: 0.0` — validation gap left by #2612 (`tokio::time::interval(Duration::ZERO)`)
## Summary

A dataflow descriptor with `health_check_interval: 0.0` passes `dora validate` and then **panics the daemon** the moment the dataflow starts. This is the exact failure class that #2612 (commit a96ed62… — actually 47fa510, "fix(core): validate health_check_interval and input_timeout as seconds") set out to eliminate, but the added validation only guards against negative / NaN / infinite values, not zero — and zero is specifically dangerous for this field because it is fed into `tokio::time::interval`, which panics on a zero period.

## Root cause / affected code paths

1. **Validation permits `0.0`.** `check_seconds_field` (introduced/generalized by #2612) rejects only non-finite and negative values:

   `libraries/core/src/descriptor/validate.rs` (~L203):
   ```rust
   fn check_seconds_field(owner: &str, field: &str, value: Option<f64>) -> eyre::Result<()> {
       if let Some(value) = value
           && (!value.is_finite() || value < 0.0)   // 0.0 passes
       {
           bail!(...);
       }
       Ok(())
   }
   ```
   So `health_check_interval: 0.0` clears `check_dataflow`.

2. **`0.0` becomes `Duration::ZERO`.** `binaries/daemon/src/lib.rs` (~L1002):
   ```rust
   let health_check_interval = descriptor
       .health_check_interval
       .map(Duration::from_secs_f64);   // Some(0.0) -> Some(Duration::ZERO)
   ```

3. **`Duration::ZERO` panics `tokio::time::interval`.** `binaries/daemon/src/lib.rs` (~L1376):
   ```rust
   let health_check_interval = tokio_stream::wrappers::IntervalStream::new(
       tokio::time::interval(health_check_interval_duration.unwrap_or(Duration::from_secs(5))),
   )
   ```
   `tokio::time::interval` panics with *"`interval` called with a period of zero"* when the period is `Duration::ZERO`. Note the `unwrap_or(Duration::from_secs(5))` default only guards the `None` case — a `Some(Duration::ZERO)` sails straight through.

## Reproduction

Add `health_check_interval: 0.0` at the dataflow top level of any descriptor and start it. `dora validate` succeeds; the daemon panics as soon as the dataflow's event loop is built.

## Risk

Daemon crash triggered by unvalidated descriptor input — a denial-of-availability from a single mistyped/zeroed config value. A zero health-check period is meaningless anyway, so it should never be accepted.

## Suggested fix

Reject `0.0` specifically for `health_check_interval` (a zero interval has no valid meaning), e.g. add a dedicated positive-value check for this field, or clamp/guard the interval construction so `Duration::ZERO` falls back to the default. The simplest scoped fix is to tighten the check for this field to require a strictly positive value.

Note: `input_timeout: 0.0` and the `restart_*` fields are benign at zero (they feed deadline comparisons / `sleep`, which tolerate zero), so this is specific to `health_check_interval` (and any other field that reaches `tokio::time::interval`). `health_check_timeout` is worth a second look for the same reason if it ever reaches an interval constructor.

---

*This issue was created by a scheduled automated Claude code-review check. The panic chain above was verified against the current tree but not reproduced end-to-end at runtime; please confirm before fixing.*

phil-opp:
## Summary

Fixes #2752.

A dataflow descriptor with `health_check_interval: 0.0` passed `dora validate` and then **panicked the daemon** the moment the dataflow started. The value is fed to `tokio::time::interval`, which panics with *"`interval` called with a period of zero"* on a `Duration::ZERO`. The validation added in #2612 (`check_seconds_field`) only rejected negative / non-finite values, so `0.0` slipped through, and the daemon's `unwrap_or(Duration::from_secs(5))` default only guarded the `None` case — `Some(Duration::ZERO)` sailed through.

## Changes

- **`libraries/core/src/descriptor/validate.rs`**: added an `allow_zero` parameter to `check_seconds_field`. `health_check_interval` now requires a strictly positive value; fields that tolerate zero (`input_timeout`, `finish_grace_secs`, `restart_*`, …) keep `allow_zero: true`. The rejection message reads `positive` vs. `non-negative` accordingly.
- **`binaries/daemon/src/lib.rs`**: defense-in-depth — the health-check interval construction now falls back to the 5 s default when the configured duration is zero, so a stray `Duration::ZERO` can never panic the daemon.

## Tests

- New `seconds_field_rejects_zero_when_positive_required` unit test.
- New `check_dataflow_rejects_zero_health_check_interval` regression test.
- Updated `check_dataflow_rejects_negative_health_check_interval` for the `positive` constraint wording.

Local `cargo test -p dora-core`, `cargo fmt --all -- --check`, and `cargo clippy -p dora-core -p dora-daemon -- -D warnings` all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CYmpFfw3ZcSyeo5XBAS7EE)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1375,13 +1375,19 @@ impl Daemon {
         });
 
         let health_check_clock = self.clock.clone();
-        let health_check_interval = tokio_stream::wrappers::IntervalStream::new(
-            tokio::time::interval(health_check_interval_duration.unwrap_or(Duration::from_secs(5))),
-        )
-        .map(move |_| Timestamped {
-            inner: Event::NodeHealthCheckInterval,
-            timestamp: health_check_clock.new_timestamp(),
-        });
+        // `tokio::time::interval` panics on a zero period. Descriptor validation
+        // already rejects a zero `health_check_interval` (see `check_dataflow`),
+        // but guard here as well so a stray `Duration::ZERO` falls back to the
+        // default rather than crashing the daemon (defense-in-depth for #2752).
+        let health_check_period = health_check_interval_duration
+            .filter(|d| !d.is_zero())
+            .unwrap_or(Duration::from_secs(5));
+        let health_check_interval =
+            tokio_stream::wrappers::IntervalStream::new(tokio::time::interval(health_check_period))
+                .map(move |_| Timestamped {
+                    inner: Event::NodeHealthCheckInterval,
+                    timestamp: health_check_clock.new_timestamp(),
+                });
 
         let mut events = (
             external_events,
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -81,15 +81,19 @@ pub fn check_dataflow_static(dataflow: &Descriptor) -> eyre::Result<()> {
                 &format!("input `{input_id}` of node `{}`", node.id),
                 "input_timeout",
                 input.input_timeout,
+                true,
             )?;
         }
     }
     // dataflow-level `health_check_interval` reaches `Duration::from_secs_f64`
-    // in the same way (`binaries/daemon/src/lib.rs`).
+    // in the same way (`binaries/daemon/src/lib.rs`). A zero interval must also
+    // be rejected: it is fed to `tokio::time::interval`, which panics on a zero
+    // period.
     check_seconds_field(
         "dataflow",
         "health_check_interval",
         dataflow.health_check_interval,
+        false,
     )?;
 
     // check that all inputs mappings point to an existing output
@@ -208,7 +212,7 @@ fn check_timing_fields(
         ("max_restart_delay", custom.max_restart_delay),
         ("restart_window", custom.restart_window),
     ] {
-        check_seconds_field(&owner, field, value)?;
+        check_seconds_field(&owner, field, value, true)?;
     }
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


phil-opp:

No issues found. The `allow_zero` split is correct: `health_check_interval` now rejects `0.0` (and `-0.0`, since `value == 0.0` matches both) before it can reach `tokio::time::interval` and panic on a zero period, while the zero-tolerant fields (`input_timeout`, `restart_*`) keep `allow_zero: true`. The daemon-side `.filter(|d| !d.is_zero())` is a reasonable defense-in-depth guard, and negative / non-finite values remain covered by the existing checks.

---

phil-opp:

Nice catch on the zero-interval panic. Two things worth reconciling before this lands:

**1. Rebase onto current `main` — `check_seconds_field` was refactored underneath this branch.** Since this PR's base, `main` (`0573360`) changed the predicate to `Duration::try_from_secs_f64(value).is_err()`, which also rejects finite-but-overflowing values (> `Duration::MAX`) that would otherwise panic `Duration::from_secs_f64` at spawn. This branch still edits the older `!value.is_finite() || value < 0.0` form, hence the merge conflict. On rebase, the `allow_zero` logic needs to layer on top of the new form, e.g.:
```rust
if let Some(value) = value
    && (Duration::try_from_secs_f64(value).is_err() || (!allow_zero && value == 0.0))
```
so the overflow guard isn't lost.

**2. Consider checking the resulting `Duration` rather than `value == 0.0`.** `tokio::time::interval` panics on `Duration::ZERO`, and `Duration::from_secs_f64` rounds sub-nanosecond positive values down to zero — e.g. `from_secs_f64(1e-10)` is `0ns`, but `1e-10 == 0.0` is `false`. So a tiny positive `health_check_interval` still yields a zero interval; today the daemon-side `.filter(|d| !d.is_zero())` keeps it from panicking, but validation silently accepts it and swaps in the 5s default instead of erroring. The timer path already handles this the robust way in `libraries/message/src/config.rs` (checks `interval.is_zero()` on the constructed `Duration`) — mirroring that here would make the validation match the actual panic condition.

---

phil-opp:

The 2026-07-28 rebase onto current `main` resolves point 1 from my earlier review: `allow_zero` now layers on top of the `Duration::try_from_secs_f64(value).is_err()` check, so the overflow guard is preserved and `health_check_interval` still rejects `0.0`/`-0.0`. No new issues in the reject logic.

Point 2 still stands (non-blocking): validation compares `value == 0.0`, but `Duration::from_secs_f64` rounds a sub-nanosecond positive like `1e-10` down to `Duration::ZERO`. Such a value passes validation and is then silently swapped for the 5s default via the daemon-side `.filter(|d| !d.is_zero())` rather than erroring — no panic, just a validation/UX inconsistency. Checking the constructed `Duration::is_zero()` (as the timer path in `libraries/message/src/config.rs` does) would make validation match the actual panic condition.

---

PR #2756, opened 2026-07-18
fix(core): don't delete a git clone another in-flight build is writing

### LINKED ISSUE #2711: git clone reuse (#2482): cross-session HEAD-verify can `remove_dir_all` a directory a concurrent build session is still cloning into
## Summary

#2482 (`a96ed62`, "stop reusing a wrong-commit git clone after a failed build") added a HEAD-verification step to the `Reuse` arm: before reusing a clone pinned to a full commit hash, it checks `HEAD == commit`, and if not it calls `cleanup_failed_clone()` (`remove_dir_all`) and bails.

The PR's third commit explicitly fixed a race where this delete could hit a directory a *sibling node in the same build session* was still cloning into, by gating verification on `planned_this_session`:

```rust
// libraries/core/src/build/git.rs:70-78
let planned_this_session = self
    .prepared_builds
    .get(&session_id)
    .map(|p| p.planned_clone_dirs.contains(&clone_dir))
    .unwrap_or(false);
ReuseOptions::Reuse {
    dir: clone_dir.clone(),
    verify_commit: (!planned_this_session).then_some(commit_hash),
}
```

`planned_this_session` only consults **the current session's** planned set. But in the daemon the exact same "delete a dir a sibling is still cloning into" race remains open **across concurrent build sessions**, because the daemon shares one `GitManager` and runs build prepares concurrently.

## Why it's reachable (daemon)

- The daemon holds a single long-lived `GitManager` (`binaries/daemon/src/lib.rs:290`), shared by all build sessions.
- `build_dataflow` runs the synchronous `choose_clone_dir` phase (which registers `planned_clone_dirs` and needs `&mut git_manager`), then returns a future that performs the actual clones. That future is **`tokio::spawn`ed** at `binaries/daemon/src/lib.rs:1562`, so the event loop immediately goes back to processing events — including the next `Build`.
- `choose_clone_dir`'s readiness check is `dir.exists()` OR planned-by-this-session (`git.rs:128-134`).

Race (two dataflows built with overlapping timing, both referencing the same repo at the same **full** commit hash):

1. Session **A** choose-phase → `NewClone` into dir `D`; `D` registered in A's planned set; A's build task spawned and starts cloning into `D` (so `D` now exists on disk, mid-clone).
2. Event loop handles Session **B**'s `Build`. `choose_clone_dir` sees `D.exists()` → `Reuse`. `D` is **not** in B's planned set → `verify_commit = Some(hash)`.
3. B's spawned task runs the `Reuse` arm's HEAD check (`git.rs:~324`) while A is still cloning: `Repository::open`/`head()`/`peel_to_commit` fails or HEAD isn't at the commit yet → treated as mismatch → `cleanup_failed_clone(&D)` **deletes the directory A is actively cloning into**.

A's clone then fails; B bails too. Both are **spurious build failures under concurrency** — flaky, timing-dependent, and only when two concurrent builds share a repo@commit.

## Contributing factor

The HEAD check treats **any** `Err` from HEAD resolution as "wrong commit → delete + bail" (`git.rs:~324-336`): not-a-git-repo, empty repo (no HEAD yet), a transient `Repository::open` failure, or a lock/permission hiccup all collapse to a delete. A dir that is merely *mid-clone* reads as "wrong commit."

## Severity / impact

Medium. Not a silent-wrong-artifact bug — both builds fail loudly rather than building stale source. But #2482 **introduced** this failure mode: before the PR, `Reuse` was a plain read that never deleted, so concurrent same-repo builds couldn't corrupt each other. The added self-healing delete reintroduces the exact race the third commit set out to close, for the cross-session case.

Related same-session case (also uncovered): when a **prior finished** build left `D` on the wrong commit and two nodes of the new session reference it, the first-chosen node gets `Reuse{verify_commit: Some}` and the second gets `Reuse{verify_commit: None}` (D now planned). Under the CLI's concurrent `JoinSet` prepare path the first can delete `D` while the second plain-reuses/builds in it.

## Affected code
- `libraries/core/src/build/git.rs:70-78` (per-session guard), `:~295-337` (Reuse/HEAD check + delete), `:388-404` (`cleanup_failed_clone`).
- `binaries/daemon/src/lib.rs:290` (shared `GitManager`), `:1562` (concurrent spawn of build task), `:2665` (`clear_planned_builds` clears only one session).

## Suggested fix

Don't verify/delete a clone that any in-flight build could be writing. Options:
- Track planned/in-progress clone dirs **process-wide** (across sessions) in the `GitManager`, and skip verification (or block) for any dir currently owned by an in-flight build, not just `session_id`'s.
- Or take a per-clone-dir lock so a `Reuse` HEAD check cannot run concurrently with a `NewClone`/fetch into the same dir.
- Additionally, distinguish "dir mid-clone / not yet a valid repo" from "valid repo on the wrong commit" so a transient/partial state doesn't trigger `remove_dir_all` (e.g. only delete when HEAD resolves to a concrete, different commit).

A regression test should drive two overlapping sessions (or the concurrent `JoinSet` path) against one repo@commit and assert no clone is deleted out from under an in-flight build — the current tests exercise only sequential `prepare()` calls, so this race is invisible to them.

## Introduced by
#2482 (`a96ed62`).

---
*This issue was created by a scheduled, automated Claude code-review check. The daemon spawn/shared-`GitManager` claims were verified directly against the source (`lib.rs:290/1562/2665`, `git.rs:70-134`), but the race is timing-dependent and was reasoned about rather than reproduced — please confirm before acting.*

phil-opp:
## Summary

Fixes #2711.

#2482 added a HEAD-verification step to the `Reuse` arm of `GitFolder::prepare` that deletes a clone (`remove_dir_all`) when its HEAD isn't at the pinned commit. Its third commit gated verification on `planned_this_session` to avoid deleting a dir a *sibling node in the same build session* was still cloning into. But the daemon holds a **single `GitManager` shared across concurrent build sessions**, and spawns each build task, so the same "delete a dir a sibling is still cloning into" race stayed open **across sessions**:

1. Session A → `NewClone` into `D`; A's build task starts cloning (so `D` now exists mid-clone).
2. Session B's `Build` sees `D.exists()` → `Reuse`; `D` is not in B's planned set → `verify_commit = Some(hash)`.
3. B runs the HEAD check while A is still cloning: `Repository::open`/`head()` fails → treated as mismatch → `cleanup_failed_clone(&D)` deletes the dir A is actively cloning into. Both builds fail.

## Changes (`libraries/core/src/build/git.rs`)

1. **Process-wide in-flight guard.** `choose_clone_dir` now consults *every* in-flight build session's `planned_clone_dirs` (not just `session_id`'s) when deciding whether to hand down a commit to verify. A dir any running build owns is reused with `verify_commit: None` — a plain read, no HEAD check, no delete. Once every session that touched the dir clears its planned set, a genuinely stale leftover is HEAD-verified again.

2. **No delete on a transient/partial HEAD state.** The HEAD check now only deletes when HEAD resolves to a concrete, *different* commit (`Ok(other)`). A HEAD-resolution error (`Err` — mid-clone, empty/partial repo with no HEAD yet, or a transient open/lock failure) no longer collapses to "wrong commit → delete"; it fails loudly and **leaves the dir untouched**, so it can't yank a dir out from under an in-flight clone.

## Tests

- `choose_clone_dir_skips_verify_for_dir_planned_by_another_session` — drives two overlapping sessions against one repo@commit and asserts the second reuses without verification, then that a genuinely stale leftover is verified again after both clear.
- `reuse_leaves_dir_when_head_unresolvable` — a dir whose HEAD can't be resolved must fail loudly without being deleted.
- Existing `Reuse` tests (wrong-commit delete, HEAD-match reuse, branch-ref skip, sibling-clone untouched) still pass.

Local `cargo test -p dora-core --features build`, `cargo fmt --all -- --check`, and `cargo clippy -p dora-core --features build --all-targets -- -D warnings` all pass.

## Note

The reported same-session `JoinSet` variant is also covered by fix (1), since it consults all in-flight sessions including the current one. A full per-clone-dir lock (the heavier alternative in the issue) was deliberately not taken — this targeted change removes the spurious delete without restructuring the concurrent build path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CYmpFfw3ZcSyeo5XBAS7EE)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1561,6 +1561,16 @@ impl Daemon {
                     session_id,
                     result,
                 } => {
+                    // The build future (every clone/checkout task) has finished, so
+                    // this session no longer owns any in-flight clone dir. Drop its
+                    // planned set now that it's done: `choose_clone_dir` consults
+                    // *every* session's planned set to skip the Reuse HEAD check for
+                    // a dir a concurrent build is still writing (#2711), so "planned"
+                    // must mean "in flight". A finished build's dirs left here would
+                    // permanently suppress that check for later reuses (reintroducing
+                    // #2482) and would accumulate for the daemon's whole lifetime.
+                    self.git_manager.clear_planned_builds(session_id);
+
                     let (build_info, result) = match result {
                         Ok(build_info) => (Some(build_info), Ok(())),
                         Err(err) => (None, Err(err)),
--- a/libraries/core/src/build/git.rs
+++ b/libraries/core/src/build/git.rs
@@ -299,12 +299,14 @@ impl GitFolder {
             ReuseOptions::Reuse { dir, verify_commit } => {
                 // Belt and braces for #2480: even with the cleanup above, a stale
                 // dir could still slip through if remove_dir_all itself failed or
-                // we got killed mid-checkout. So for a clone left by a prior build
-                // (verify_commit is Some) that's pinned to a full commit hash, I
-                // check that HEAD actually points there. verify_commit is None when
-                // a sibling node in this build owns the dir and may still be cloning
-                // into it, so I leave those completely untouched. Branch and tag
-                // pins I can't verify offline, so those pass through too.
+                // we got killed mid-checkout. So for a clone that no in-flight build
+                // owns (verify_commit is Some) and that's pinned to a full commit
+                // hash, I check that HEAD actually points there. verify_commit is
+                // None when some running build session has the dir in its planned
+                // set — a sibling node here, or a concurrent build in the daemon's
+                // shared GitManager (#2711) — and may still be cloning into it, so I
+                // leave those completely untouched. Branch and tag pins I can't
+                // verify offline, so those pass through too.
                 if let Some(commit_hash) = verify_commit
                     && dir.exists()
                     && is_full_commit_hash(&commit_hash)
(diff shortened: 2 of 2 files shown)


phil-opp:

No blocking issues found. Both changes look correct: `choose_clone_dir` now reuses a clone dir owned by *any* in-flight build session (`prepared_builds.values().any(...)`) rather than only the current session's, and a HEAD-resolution **error** in `GitFolder::prepare` no longer collapses into "wrong commit → `remove_dir_all`", so a concurrent writer's clone can't be deleted out from under it.

One behavior change worth a conscious decision: the new `Err(err) => bail!(...)` arm never deletes the dir, so a genuinely corrupt / un-openable clone that is *not* mid-write (e.g. a `.git` left truncated by a crashed process) now fails every retry instead of self-healing via a fresh re-clone as it did before. That's the safer default given the race this fixes, but it does drop the old auto-recovery path — fine to keep as-is if intentional.

---

phil-opp:

I found one issue worth a conscious decision before merging (distinct from the `Err`-arm self-heal tradeoff already discussed).

**The broadened in-flight guard keys on a *persistent* set, so it can permanently disable the #2480 HEAD verification.**

The `verify_commit` gate changes from "planned by *this* session" to consulting *every* session's `planned_clone_dirs`:

```rust
let planned_by_in_flight_build = self
    .prepared_builds
    .values()
    .any(|p| p.planned_clone_dirs.contains(&clone_dir));
```

The comment justifies this as "a dir that *any* still-running build session planned." But `planned_clone_dirs` is not an *in-flight* signal — entries in `prepared_builds` are only ever removed by `clear_planned_builds(session_id)` (`libraries/core/src/build/git.rs:158`), which is called from exactly one place: the **start** of that same session's next `build_dataflow` (`binaries/daemon/src/lib.rs:2847`). There is no removal on build completion, dataflow stop, or session/CLI disconnect.

Consequence: a session that builds once and exits — the common `dora build` / `dora start` case — leaves its planned dirs in `prepared_builds` for the daemon's entire lifetime. Because clone dirs are deterministic per repo@commit (as the new `choose_clone_dir_skips_verify_for_dir_planned_by_another_session` test itself relies on), that stale entry now forces `verify_commit: None` for that repo@commit across **all future sessions** on that daemon. The "belt and braces for #2480" HEAD check is then silently skipped for any clone dir once the session that first planned it goes away — reopening exactly the stale/wrong-commit reuse #2480's verification was added to catch.

The race this PR targets is specifically "a dir some build is *actively cloning into right now*." Scoping the guard to genuinely in-flight clones — a set that entries are added to when the `NewClone` task starts and removed when it finishes/fails — would close the cross-session race (#2711) without permanently suppressing verification for every dir after its originating session exits. As written, the guard over-approximates "in-flight" with "was ever planned by any session that never rebuilt."

The added cross-session test asserts the intended skip, but it can't catch this because it manually calls `clear_planned_builds` for both sessions; in the real one-shot-build flow that clear never happens.

_(Reviewed the diff at head `e9465c1`; no code changes since the prior automated review, so I've limited this to the one new point above.)_

---

phil-opp:
Closing in favor of 2795

PR #2757, opened 2026-07-18
fix(daemon): purge removed node's per-node bookkeeping on RemoveNode

phil-opp:
## Summary

The `RemoveNode` coordinator handler in `binaries/daemon/src/lib.rs` cleans up routing tables (`mappings`, `timers`, `log_subscribers`) and most per-node maps (`running_nodes`, `open_inputs`, `subscribe_channels`, `pending_messages`, `all_inputs_closed_at`, `connected_nodes`, `finish_escalated`, descriptor entry) — but leaves three maps keyed by the removed node untouched:

- `input_deadlines: HashMap<(NodeId, DataId), InputDeadline>`
- `broken_inputs: HashMap<(NodeId, DataId), Duration>`
- `node_stderr_most_recent: BTreeMap<NodeId, Arc<ArrayQueue<String>>>`

## The bug

A `input_deadlines` entry for an input that never received traffic is "not yet armed", so `InputDeadline::is_timed_out()` returns `false` and it is **never** removed (entries are only dropped when they time out, at `lib.rs:2641`). `check_input_timeouts` (`lib.rs:2607`) therefore re-scans it on **every tick** for the lifetime of the dataflow.

A `broken_inputs` entry can never recover once the node is gone — recovery only happens on message receipt (`send_output_to_local_receivers`), which a removed node never sees.

Across repeated dynamic add/remove cycles (a supported reconfiguration pattern, cf. the `#2270` comments in the same handler) these stale entries accumulate without bound, together with the dangling per-node stderr queue, wasting per-tick work indefinitely.

## The fix

Add `RunningDataflow::forget_node_bookkeeping(&mut self, node_id)` which `retain`s away the removed node's `input_deadlines`/`broken_inputs` entries and drops its `node_stderr_most_recent` queue, and call it from the `RemoveNode` cleanup block. Other nodes' entries are preserved.

## Validation

- New unit test `forget_node_bookkeeping_purges_only_that_node` populates all three maps for two nodes, removes one, and asserts only that node's entries are dropped. **Passes.**
- `cargo fmt -p dora-daemon -- --check`, `cargo clippy -p dora-daemon -- -D warnings`, `cargo test -p dora-daemon` all clean.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2342,6 +2342,12 @@ impl Daemon {
                     // subscribes and could be selected mid-startup (dora#2270).
                     dataflow.connected_nodes.remove(&node_id);
                     dataflow.finish_escalated.remove(&node_id);
+                    // Purge per-node bookkeeping keyed by node id that the
+                    // routing cleanup above doesn't touch. Otherwise stale
+                    // input_deadlines/broken_inputs entries are re-scanned
+                    // every tick forever and the stderr queue leaks across
+                    // repeated dynamic add/remove cycles.
+                    dataflow.forget_node_bookkeeping(&node_id);
 
                     // Remove from stored descriptor (inverse of AddNode
                     // push) so descriptor-based lookups stay consistent.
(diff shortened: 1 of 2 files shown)


phil-opp:

This PR is one of a batch opened by an automated (machine-generated) code-review pass over `origin/main`. Each finding was independently verified against current `main` before a PR was opened, and prior open `claude/*` PRs were checked to avoid duplicates. Every branch was verified with `cargo fmt`/`clippy`/tests for its affected crate; the combined diff additionally passed `cargo fmt --all -- --check`, combined `clippy -D warnings`, and the unit-test suites of all affected crates (`dora-core` 278, `dora-daemon` 104, `dora-node-api` 141, `dora-operator-api` 7 — all green), plus the `/review` and `/simplify` gates (no findings).

### PRs by category

**Correctness / resource-leak fixes**
- #2757 — `fix(daemon)`: purge a removed node's `input_deadlines`/`broken_inputs`/`node_stderr_most_recent` on `RemoveNode` (stale entries otherwise re-scanned every tick, unbounded across add/remove cycles).
- #2758 — `fix(daemon)`: record force-stopped nodes in `grace_duration_kills` so `dora stop --force` doesn't report daemon-killed nodes as genuine failures.
- #2760 — `fix(core)`: trim field-type strings in `resolve_field_type` so a padded type accepted by the admission gate doesn't silently fail schema materialization.

**Optimization (hot path)**
- #2761 — `perf(daemon)`: skip the per-message `Metadata` deep-clone + `Arc` alloc when an output has no local receiver (cross-daemon-only / debug-inspection outputs).

**Documentation**
- #2759 — `docs(apis)`: fix the copy-pasted `DoraEvent` alias doc and document the undocumented `DoraOutputSender` operator type.

### Rejected findings (verified, then dropped)

- **`download_file` doesn't reject HTTP 4xx/5xx** (error page written to disk as the artifact) — already fixed by open PR #2744.
- **`download_file` derives filename from full URL incl. query string** — already fixed by open PR #2744.
- **Python operator metadata: a list of `bool`s coerces to the string `"[True, False]"`** — adjacent to already-covered Python metadata work, and there is no `ListBool` parameter variant to represent it faithfully; not a clear-cut, independently-fixable defect.
- **`download_file` sets `0o700` on every downloaded file** (incl. YAML/`.py`) — permission-hygiene wart, not a correctness bug.
- **`types_match` treats disjoint parameter sets as compatible** — documented intended "shared keys must agree" semantics.

---

⚠️ *All PRs in this batch are machine-generated by Claude (an AI assistant) and should be reviewed carefully before merging.*

---

phil-opp:

Reviewed the diff — no issues found. `forget_node_bookkeeping` correctly purges only the removed node's entries (`retain(|(n,_),_| n != node_id)` on `input_deadlines`/`broken_inputs`, plus `node_stderr_most_recent.remove`), and the regression test confirms a sibling node's entries survive. The call is placed alongside the existing `connected_nodes`/`finish_escalated` cleanup in the `RemoveNode` handler.

---

PR #2761, opened 2026-07-18
perf(daemon): skip per-message metadata clone when output has no local receiver

phil-opp:
## Summary

`send_output_to_local_receivers` (`binaries/daemon/src/lib.rs`) is on the per-message output hot path. It unconditionally built an Arc-wrapped copy of the message metadata:

```rust
let local_receivers = dataflow.mappings.get(&output_id).unwrap_or(&empty_set);
let metadata = Arc::new(metadata.clone());   // always runs
```

but that Arc is only ever consumed inside the `for (receiver_id, input_id) in local_receivers` fan-out loop.

## The waste

When an output has **no local receiver** — a purely cross-daemon output, or any output while `enable_debug_inspection` forces `need_data_bytes = true` with no local subscriber — the loop body never executes, yet the full `Metadata` deep-clone (including its `parameters: BTreeMap<String, Parameter>`) plus an `Arc` allocation was performed and immediately dropped, **on every message**. The remote-forwarding path uses the original `&metadata` (moved into `InterDaemonEvent` by the caller), not this clone.

## The fix

Wrap the metadata only when there is at least one local receiver:

```rust
let metadata = (!local_receivers.is_empty()).then(|| Arc::new(metadata.clone()));
```

and unwrap it inside the loop (reached only when `local_receivers` is non-empty). `data` is left wrapped unconditionally on purpose — the `need_data_bytes` path below moves the payload out of a uniquely-owned `Arc` without copying, which is the existing zero-copy remote optimization.

## Validation

- The existing `data_bytes_returned_with_and_without_local_receivers` test already exercises both the no-receiver path (returns the payload, now with `metadata = None`) and the with-receiver path (delivers an `Input` event). Extended its with-receiver branch to assert the delivered metadata timestamp still matches the input — guarding metadata delivery through the `Option`/`expect` refactor.
- `cargo fmt -p dora-daemon -- --check`, `cargo clippy -p dora-daemon -- -D warnings`, `cargo test -p dora-daemon` (103 tests) all pass.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -5372,8 +5372,15 @@ async fn send_output_to_local_receivers(
     let empty_set = BTreeSet::new();
     let output_id = OutputId(node_id, output_id);
     let local_receivers = dataflow.mappings.get(&output_id).unwrap_or(&empty_set);
-    // Wrap in Arc once; fan-out clones are O(1) atomic ref bumps instead of O(payload_size) memcpy
-    let metadata = Arc::new(metadata.clone());
+    // Wrap in Arc once; fan-out clones are O(1) atomic ref bumps instead of O(payload_size) memcpy.
+    // Only clone the metadata when there is at least one local receiver to hand
+    // it to: for outputs consumed only across daemons — or while debug
+    // inspection forces `need_data_bytes` with no local subscriber — the loop
+    // below never runs, so this per-message `Metadata` deep-clone + allocation
+    // would otherwise be pure waste. (`data` still wraps unconditionally: the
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff — no issues found. `metadata` becomes `Some(Arc::new(..))` exactly when `local_receivers` is non-empty, which is the only condition under which the `for` loop body (and thus the `.expect(...)`) is reached, so the expect can never panic. `metadata.clone()` on the `Option<Arc>` is an O(1) refcount bump. `data` is still wrapped unconditionally as required by the `need_data_bytes` move-out path.

---

phil-opp:
Superseded by #2837

Issue #2762, opened 2026-07-19
zenoh per-node listener bind race (#2716) reintroduces the #1858 silent-partition class on a path the #1858 fix does not cover


phil-opp:
> **Automated review notice:** This issue was created by a **scheduled automated Claude review** of recent commits/PRs. It is a code-path analysis (not observed at runtime), so please sanity-check the reasoning before acting. Probability is low; impact (silent data loss) is high.

## Summary

Closed issue **#1858** ("zenoh listener bind race: open Ok ≠ actually bound") documented a silent-partition race for the daemon's zenoh listener and was fixed by verifying the bind post-open via `session.info().locators()` and only advertising an endpoint that actually bound.

Commit **fd8af08** (PR **#2716**, "restore reliable node links … under zenoh 1.9", merged 2026-07-16) added a **new per-node listener mechanism** (`plan_zenoh_peering` + `DORA_ZENOH_LISTEN`/`DORA_ZENOH_CONNECT`) that reintroduces the same failure class on a path the #1858 fix structurally cannot cover. The daemon reserves each local node's loopback port and commits it into that node's **consumers'** dial list **at plan time — before the producing node process has bound the port** — and the node then discards its own bind-verification result. If the reserved port is lost in the reserve→bind window, the producer runs listener-less, its consumers hold a hardcoded dead endpoint, and **multicast fallback has been explicitly disabled** for them, so there is no recovery.

## Why #1858's fix does not protect this path

The #1858 fix works because the daemon does **bind-then-advertise**: `open_zenoh_session_with_listen` opens the session, confirms the requested endpoint is in `info().locators()`, and only then returns `effective_listen_endpoint = Some(ep)` for the daemon to inject (`libraries/core/src/topics.rs` ~330). The per-node path from #2716 inverts this to **advertise-then-bind**:

1. `plan_zenoh_peering` reserves each local node's port with `reserve_loopback_zenoh_endpoint()` (binds `127.0.0.1:0`, reads the port, **drops the socket immediately** — `libraries/core/src/topics.rs:401`) and records it in `binaries/daemon/src/spawn/spawner.rs:163`.
2. The plan is committed into `dataflow.zenoh_peering` and each consumer's `DORA_ZENOH_CONNECT` list already contains the producer's `tcp/127.0.0.1:P` (`maybe_inject_zenoh_connect`, `spawner.rs` ~258) — this happens as the nodes are spawned, i.e. **before the producer binds `P`**.
[...]

PR #2764, opened 2026-07-19
fix(daemon,core): correct misleading zenoh bind-failure warning and document per-node advertise-before-bind race (#2762)

phil-opp:
Fixes the "at minimum" part of #2762.

## Background

The per-node zenoh listener mechanism added in #2716 (`plan_zenoh_peering` + `DORA_ZENOH_LISTEN`/`DORA_ZENOH_CONNECT`) reserves each local node's loopback port with `reserve_loopback_zenoh_endpoint()` (bind `127.0.0.1:0`, read the port, drop the socket) and commits that port into every **consumer's** dial list at plan time — **before the producing node process binds it**. Those consumers also set explicit `connect/endpoints`, which disables multicast scouting. If the reserve→bind window is lost (another process grabs the port), the producer runs listener-less, the consumers hold a dead endpoint, and there is **no multicast fallback** — that edge is silently partitioned.

Unlike the daemon's own listener, this path is **not** covered by the #1858 bind-verification: the node opens with `listen/exit_on_failure: false` and `open_zenoh_session` discards its own `effective_listen_endpoint`, so nothing propagates the failed bind back to the consumers that were already told to dial it.

## What this PR changes

This is the low-risk, explicitly-endorsed "at minimum" fix from the issue (suggested fix #3). It does **not** attempt to close the race itself.

- **`libraries/core/src/topics.rs`** — the bind-verification warning claimed *"spawned nodes will use multicast scouting only"* unconditionally. But when explicit `connect/endpoints` are set (`connect_inserted`), multicast scouting was just disabled, so that fallback does **not** exist. The old message pointed a debugger the wrong way when the race actually fired. The warning is now split:
  - explicit connect endpoints set → no fallback path; the message says so and references #2762;
  - otherwise → genuinely falls back to multicast scouting.

- **`binaries/daemon/src/spawn/spawner.rs`** — document on `plan_zenoh_peering` that the per-node endpoints are advertised to consumers **before** the node binds, and are therefore **not** covered by the #1858 daemon-listener bind verification, so future readers don't assume they are.

## Deliberately out of scope (needs maintainer design input)

The two structural fixes from the issue — (1) holding each port reservation until the child confirms its bind, or (2) propagating each node's `effective_listen_endpoint` back to the daemon and re-planning/keeping multicast on failure — are larger changes to the fault-tolerance-sensitive spawn path and are left for maintainer direction. The race probability is low (fresh ephemeral ports each call); the value here is removing the misleading diagnostic and recording the gap.

## Validation

- `cargo clippy -p dora-core -p dora-daemon -- -D warnings` — clean
- `cargo test -p dora-core -p dora-daemon` — pass (105 tests)

Changes are a logging-message refactor and a doc comment; no control-flow logic changed.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01VVAiDbEZ7Uuy3K4yH7vwJz)_

### DIFF
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -135,6 +135,24 @@ pub struct NodeZenohPeering {
 /// Dynamic nodes are omitted: they are not in the descriptor's spawn set and join
 /// at arbitrary times, so they keep the daemon-endpoint-only behavior.
 ///
+/// # Caveat: per-node endpoints are advertised before they bind (#2762)
+///
+/// This reserves each local node's loopback port with
+/// `dora_core::topics::reserve_loopback_zenoh_endpoint`
+/// — which binds `127.0.0.1:0`, reads the port, and drops the socket — and commits
+/// that port into every *consumer's* dial list here at plan time, i.e. **before the
+/// producing node process has bound it**. Unlike the daemon's own listener, these
+/// per-node endpoints are therefore *not* covered by the bind-verification in
+/// `open_zenoh_session_with_listen` (#1858): the node opens with
+/// `listen/exit_on_failure: false` and discards its own `effective_listen_endpoint`,
+/// so a lost reserve→bind race (another process grabs the port in the window) leaves
+/// the producer listener-less while its consumers hold a dead endpoint. Because those
+/// consumers also have explicit `connect/endpoints`, multicast scouting is disabled
+/// for them, so there is no fallback and that edge is silently partitioned. The
+/// probability is low (the OS keeps handing out fresh ephemeral ports) but the impact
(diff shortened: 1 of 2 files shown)


phil-opp:

Reviewed the diff — no issues found. The warning split correctly distinguishes the two cases: when `connect_inserted` is true, multicast scouting was disabled earlier in this function (the `scouting/multicast/enabled=false` insert), so the "no fallback path" message is accurate; the original "falling back to multicast" wording is retained only for the `else` (no explicit connect endpoints) branch. `connect_inserted` and `bound_locators` are both in scope at the warning sites. The `plan_zenoh_peering` doc caveat is an accurate description of the advertise-before-bind window and does not change behavior.

---

PR #2767, opened 2026-07-19
fix(coordinator-store): reject over-limit records at write time to prevent silent data loss

phil-opp:
## Issue

`RedbStore::decode` deliberately caps deserialization at 64 MiB (to bound allocation from a malformed varint length prefix):

```rust
let config = bincode::config::standard().with_limit::<{ 64 * 1024 * 1024 }>();
```

…but `encode` and the `put_*` write paths had **no matching bound**. The two limits were asymmetric, so a `put` → `get` round-trip was not guaranteed to be lossless for the store's own data:

- A `DataflowRecord` whose `descriptor_json` / `node_to_daemon` map serialized to more than 64 MiB was accepted by `put_dataflow` and written to disk.
- Every later `get_dataflow` then returned a decode error, and `list_dataflows` treated that as a **corrupt row and silently skipped it** (the skip logic added for #2027).
- After a coordinator restart, the dataflow silently vanished from recovery even though its on-disk bytes were perfectly valid — a write that can never be read back, mislabeled as corruption.

## Fix

Share a single `MAX_RECORD_BYTES` constant between `decode`'s `with_limit` and a new length check in `encode`, so an over-limit record is rejected **loudly at write time** with a clear error instead of becoming silently unreadable. This also removes the duplicated `64 * 1024 * 1024` literal.

## Validation

- `cargo clippy -p dora-coordinator-store --all-targets -- -D warnings` — clean
- `cargo test -p dora-coordinator-store --features redb-backend` — 34 pass, including a new regression test `encode_rejects_record_larger_than_decode_limit`.

---

🤖 This PR is **machine-generated by Claude** (an AI agent) as part of an automated code-review pass. Please review carefully before merging.

https://claude.ai/code/session_01H9EsyZqeYbHZjjgpdMazTu

---
_Generated by [Claude Code](https://claude.ai/code/session_01H9EsyZqeYbHZjjgpdMazTu)_

### DIFF
--- a/libraries/coordinator-store/src/redb_store.rs
+++ b/libraries/coordinator-store/src/redb_store.rs
@@ -109,18 +109,30 @@ impl RedbStore {
 // Serialisation helpers
 // ---------------------------------------------------------------------------
 
+/// Maximum size of a single serialized record. `decode` refuses to *read*
+/// beyond this (capping allocation from a malformed varint length prefix), and
+/// `encode` refuses to *write* beyond it. The two limits must match: a record
+/// that encodes larger than `decode` will accept is silently unreadable —
+/// `put_dataflow` would persist it, but every later `get_dataflow` returns a
+/// decode error and `list_dataflows` skips it as "corrupt" (dora-rs/dora#2027),
+/// dropping a perfectly valid dataflow from startup recovery. Guarding at write
+/// time turns that silent data loss into an explicit error at the source.
+const MAX_RECORD_BYTES: usize = 64 * 1024 * 1024;
+
 fn encode<T: serde::Serialize>(value: &T) -> Result<Vec<u8>> {
-    bincode::serde::encode_to_vec(value, bincode::config::standard())
-        .map_err(|e| eyre!("encode error: {e}"))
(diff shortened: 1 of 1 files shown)


phil-opp:
> 🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Treat findings as suggestions to verify.

No issues found. Making `encode` reject records over `MAX_RECORD_BYTES` closes a real asymmetry: `decode` capped reads at 64 MiB while `put_*` had no matching write bound, so an over-limit record could be persisted yet be unreadable by every later `get_dataflow`/`list_dataflows` (silent recovery loss). The shared `const MAX_RECORD_BYTES` keeps the write check (`encode`) and the read limit (`decode`'s `with_limit::<{ MAX_RECORD_BYTES }>()`) in lockstep in `libraries/coordinator-store/src/redb_store.rs`; the const-in-const-generic form is valid, and the two limits are symmetric at the boundary (a record encoding to exactly the limit still round-trips; limit+1 is rejected at write time, matching what `decode` would refuse). The regression test covers both the reject and round-trip paths.

---

PR #2772, opened 2026-07-20
fix(hub-client): confine version_entry_path writes through symlinked parents

### LINKED ISSUE #2771: hub-client: `confine()` doesn't enforce `version_entry_path`'s documented "write can't land outside catalog root" guarantee for non-existent files
> **Note:** This issue was created by a scheduled automated Claude code-review check. It reports a *latent* correctness gap (not a currently-reachable exploit); please validate before acting.

## Summary

`IndexCatalog::confine()` in `libraries/hub-client/src/index.rs` treats "path does not resolve" as "path is safe". Because `std::fs::canonicalize` fails on the **whole path** when the final component doesn't exist yet, `confine()` returns `Ok(())` for any not-yet-created file **without resolving intermediate symlinks**. As a result, `version_entry_path()` does not actually enforce the safety guarantee its doc comment promises:

> Errors if the path escapes the catalog root via a symlink, so a write can't land outside it.

## Code

`confine()` (index.rs:196-205):

```rust
fn confine(&self, path: &Path) -> eyre::Result<()> {
    match path.canonicalize() {
        Ok(real) if real.starts_with(&self.canonical_root) => Ok(()),
        Ok(_) => eyre::bail!(
            "index path `{}` escapes the catalog root (symlink?)",
            path.display()
        ),
        Err(_) => Ok(()),   // <-- non-existent leaf => treated as safe, symlinks NOT resolved
    }
}
```

`version_entry_path()` (index.rs:279-290):

```rust
/// The confined on-disk path of a version entry file, for callers that
/// read or rewrite it in place (e.g. `dora hub yank`). Errors if the path
/// escapes the catalog root via a symlink, so a write can't land outside it.
pub fn version_entry_path(
    &self,
    namespace: &str,
    name: &str,
    version: &Version,
) -> eyre::Result<PathBuf> {
    let path = self
        .package_dir(namespace, name)?
        .join(format!("{version}.yml"));
    self.confine(&path)?;
    Ok(path)
}
```

## Failure scenario

Catalog root `/idx`. The catalog is a local (git-cloned) directory, and a crafted/compromised index repo contains a symlinked package directory:

- `/idx/acme/lidar` → `/etc/evil` (a symlink pointing outside the root)

Call `version_entry_path("acme", "lidar", &"1.0.0")`:

1. `path = /idx/acme/lidar/1.0.0.yml`.
2. `1.0.0.yml` does not exist yet, so `path.canonicalize()` returns `Err(NotFound)` — even though the `lidar` component is a symlink out of the root.
3. `confine()` hits the `Err(_) => Ok(())` arm and returns success.
4. A caller that then **writes a new entry** to the returned path follows the `lidar` symlink and writes to `/etc/evil/1.0.0.yml` — outside the catalog root, contradicting the method's doc.

## Why it isn't a live exploit today

The only current caller, `binaries/cli/src/command/hub/yank.rs:91`, guards the write with `if !entry_path.is_file() { bail!(...) }`, so it only ever rewrites a **pre-existing** file. For an existing file, `canonicalize` succeeds and the escape is caught by the `Ok(_)` arm. The gap only bites a future caller that creates a *new* entry file through `version_entry_path` (a natural use given the method's stated purpose "for callers that read or rewrite it").

The read paths (`versions`, `entry`, `package_meta`) are safe: a non-existent leaf that slips past `confine` simply fails the subsequent read, and an existing symlinked target is caught by the `Ok(_)` escape arm.

## Suggested fix

Make `confine()` enforce containment structurally rather than relying on the leaf existing. Options:

- Canonicalize the **parent directory** (which must exist for any write) and verify it starts with `canonical_root`, then re-join the final component; or
- Lexically normalize the path (reject `..`, then check `starts_with(root)`) *and* canonicalize whatever prefix does exist and confirm it stays within the root.

Either approach closes the "non-existent leaf ⇒ unchecked symlinks" hole so `version_entry_path` actually upholds its documented invariant. Adding a regression test with a symlinked package directory pointing outside the root (writing a not-yet-existing entry) would pin the behavior.

## References
- `libraries/hub-client/src/index.rs:196-205` (`confine`)
- `libraries/hub-client/src/index.rs:279-290` (`version_entry_path`)
- `binaries/cli/src/command/hub/yank.rs:91` (the guard that currently masks the gap)

phil-opp:
## Summary

Fixes #2771.

`IndexCatalog::confine()` in `libraries/hub-client/src/index.rs` treated any `canonicalize()` failure as "path is safe", returning `Ok(())` on the `Err(_)` arm. Because `std::fs::canonicalize` fails on the **whole path** when the final component doesn't exist yet, `confine()` never resolved intermediate symlinks for a not-yet-created leaf.

As a result, `version_entry_path()` did not actually enforce its documented guarantee:

> Errors if the path escapes the catalog root via a symlink, so a write can't land outside it.

A crafted/compromised index repo with a symlinked package directory (e.g. `/idx/acme/lidar` → `/etc/evil`) would cause `version_entry_path("acme", "lidar", "1.0.0")` to return `Ok` — the leaf `1.0.0.yml` doesn't exist, so `canonicalize` failed and the symlinked `lidar` component was never resolved. A caller that then writes the entry follows the symlink out of the root.

## Change

When the leaf is absent, `confine()` now resolves the path **component by component** under the canonical root:

- an existing component that redirects (symlink) outside the root is rejected;
- a dangling symlink in the tail is rejected (its target can't be verified);
- genuinely new, validated leaf names still resolve to an in-root writable path.

The fast path (fully existing path) is unchanged in behavior. Existing callers (`versions`, `entry`, `package_meta`) behave identically for existing and non-existent paths.

## Tests

- `version_entry_path_rejects_symlinked_dir_for_a_nonexistent_leaf` — reproduces the escape (fails under the old `Err(_) => Ok(())`).
- `version_entry_path_allows_a_new_leaf_under_the_root` — a not-yet-existing entry under a legitimate package still resolves.
- Existing `reads_reject_symlinked_package_dir_escaping_the_catalog` still passes.

`cargo test -p dora-hub-client`, `cargo clippy -p dora-hub-client -- -D warnings`, and `cargo fmt --check` all pass.

https://claude.ai/code/session_013BXYFyne1xCkXHL55jwnA1

---
_Generated by [Claude Code](https://claude.ai/code/session_013BXYFyne1xCkXHL55jwnA1)_

### DIFF
--- a/libraries/hub-client/src/index.rs
+++ b/libraries/hub-client/src/index.rs
@@ -5,7 +5,7 @@
 //! verbatim and a source pointer; resolution picks the highest non-yanked
 //! version satisfying a requirement.
 
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 
 use dora_core::{manifest::NodeManifest, types::edit_distance};
 use eyre::Context;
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found. The component-by-component resolution closes the gap where a missing leaf made `canonicalize` fail on the whole path and skip intermediate-symlink checks: existing symlinked components are canonicalized and rejected if they escape `canonical_root`, non-`Normal` components are rejected defensively, and genuinely new leaf names still resolve in-root. Fast path and existing-caller behavior are unchanged, and the added tests cover both the escape and the legitimate-new-leaf cases.

---
*🤖 This is a fully automated review by Claude. It may be incorrect.*

---

PR #2773, opened 2026-07-20
fix(daemon): reject explicit loopback --zenoh-listen under a remote coordinator

### LINKED ISSUE #2770: zenoh-listen (#2722): explicit `--zenoh-listen <loopback>` under a remote coordinator is accepted and runs silently undialable
## Summary

#2722 (`1650eed`, "bind zenoh listener on a routable address for multi-machine setups") added `--zenoh-listen <IP>` and the `ZenohBind` type, whose `Explicit` variant is documented as: *"An address the operator named must bind or the daemon must exit … they are on a network where they had to say the address out loud, which is precisely a network where multicast is not going to rescue us."*

That contract is **not enforced for a loopback address**. `--zenoh-listen 127.0.0.1` (or `::1`) while `--coordinator-addr` points at a remote host is accepted, the daemon starts, and it advertises `127.0.0.1` to remote peers — who dial their own loopback and reach nothing. Under zenoh 1.9's no-relay peer behavior (the very reason this PR exists) that pair is dead with no fallback. This is exactly the silent-undialable failure the PR set out to prevent, for an address the operator *named explicitly*.

## Why it slips through

- `validate_zenoh_listen` (`libraries/core/src/topics.rs:455-465`) rejects only `is_unspecified()` (the wildcard), **not** `is_loopback()`:
  ```rust
  pub fn validate_zenoh_listen(bind: IpAddr) -> eyre::Result<()> {
      if bind.is_unspecified() { eyre::bail!(...); }
      Ok(())
  }
  ```
  So a wildcard (unusable for multi-machine) is a hard error, but loopback (equally unusable for multi-machine) passes. That is an inconsistency: both have "nothing to advertise" to a remote peer.
- The early bind probe (`binaries/daemon/src/lib.rs:716-724`) only catches an address that *doesn't exist on this host* (`EADDRNOTAVAIL`). Loopback always binds, so the probe passes.
- The only remaining signal is a **warning**, not an exit (`binaries/daemon/src/lib.rs:726-737`):
  ```rust
  if zenoh_bind.addr().is_loopback() && !coordinator_ws_addr.ip().is_loopback() {
      tracing::warn!(
          "coordinator at {coordinator_ws_addr} is remote, but no routable local \
           address toward it was found; zenoh will bind loopback and other daemons \
           will not be able to reach this one. Pass --zenoh-listen <IP> explicitly."
      );
  }
  ```
  For the `ZenohBind::Explicit` case this text is **factually wrong**: nothing was "searched for," and the operator *did* pass `--zenoh-listen` explicitly — it was just loopback. The message is written for the `Derived` fallback and reused verbatim for the explicit path.

## Impact

Low–medium. It requires operator misconfiguration (`--zenoh-listen 127.0.0.1` with a remote coordinator), but the result is a daemon that looks healthy (binds fine, connects to the coordinator) yet is unreachable by every other daemon, with only a warning that misdescribes the cause — the precise "silently dead daemon" outcome this mechanism was built to eliminate. Debugging a silent multi-machine dataflow stall from a warning is exactly the cost the routable-bind change was meant to remove.

## Affected code
- `libraries/core/src/topics.rs:455-465` (`validate_zenoh_listen` — loopback not rejected)
- `binaries/daemon/src/lib.rs:700-737` (`run_inner_with_builds` — explicit path warns instead of failing; message wrong for the explicit case)

## Suggested fix

Make the named-address contract match its own docstring and the wildcard rejection. Either:
- In `validate_zenoh_listen`, also reject `is_loopback()` **when the daemon has a remote coordinator** (the check would need the coordinator address, or a separate `validate_explicit_zenoh_listen(addr, coordinator)` used only on the `Explicit` path), so `--zenoh-listen 127.0.0.1` under a remote coordinator is a hard error like the wildcard is; **or**
- Keep it a warning but split the message so the explicit-loopback case says something true (e.g. "you passed `--zenoh-listen <loopback>`, which peers cannot reach; pass this host's LAN/VPN address instead"). A single-host user who deliberately wants loopback with a remote-looking coordinator would still be served.

## Minor (same commit, same function)

The `wrap_err` string at `binaries/daemon/src/lib.rs:703` contains a stray run of literal spaces from an unescaped source line-wrap, so it renders as `"...derive it&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; from --coordinator-addr"`:
```rust
"invalid --zenoh-listen address (omit the flag to derive it                      from --coordinator-addr)",
```
Collapse the run to a single space.

---
*This issue was created by a scheduled, automated Claude code-review check. The code paths were verified directly against the source at the lines cited; the failure scenario was reasoned about, not reproduced on a real multi-machine setup — please confirm before acting.*

phil-opp:
## Summary

Fixes #2770.

`--zenoh-listen 127.0.0.1` (or `::1`) while `--coordinator-addr` points at a remote host was accepted: the daemon started, advertised loopback to remote peers, and those peers dialed their own loopback and reached nothing. Under zenoh 1.9's no-relay peer behavior that pair is dead with no fallback — the exact silent-undialable failure the routable-bind mechanism (#2722) exists to prevent, for an address the operator named **explicitly**.

## Why it slipped through

- `validate_zenoh_listen` (`libraries/core/src/topics.rs`) rejects only `is_unspecified()` (the wildcard), not `is_loopback()` — yet both have "nothing to advertise" to a remote peer.
- The early bind probe only catches an address that doesn't exist on this host (`EADDRNOTAVAIL`); loopback always binds.
- The only remaining signal was a `warn!`, and its text ("no routable local address toward it was found; Pass --zenoh-listen explicitly") is written for the *derived* fallback — factually wrong for the explicit path, where the operator *did* pass `--zenoh-listen`.

## Change

The `ZenohBind::Explicit` variant is documented as "bind a routable address or exit". So for an **explicit** loopback address under a **remote** coordinator, `announce_zenoh_bind` now returns a hard error with accurate, actionable text, instead of starting a silently-undialable daemon.

- **Derived** loopback under a remote coordinator still only warns (a best-effort fallback the operator can override with `--zenoh-listen`).
- **Single-machine** setups (loopback coordinator address, e.g. `dora run`) are unaffected — the guard is gated on `!coordinator_ws_addr.ip().is_loopback()`.
- A routable bind is still announced at info level.

The loopback/announce logic is extracted from `run_inner_with_builds` into a free function `announce_zenoh_bind` so the decision is unit-testable directly.

## Tests

New `announce_zenoh_bind_tests` module covers all four branches:
- explicit loopback + remote coordinator → error (the regression);
- derived loopback + remote coordinator → allowed (warn only);
- explicit loopback + local coordinator → allowed (single-machine);
- explicit routable address + remote coordinator → allowed.

`cargo test -p dora-daemon`, `cargo clippy -p dora-daemon -- -D warnings`, and `cargo fmt --check` all pass.

## Note

This does not overlap with in-flight #2764 (which touches `spawner.rs` and `open_zenoh_session_with_listen`); the two edit disjoint code.

https://claude.ai/code/session_013BXYFyne1xCkXHL55jwnA1

---
_Generated by [Claude Code](https://claude.ai/code/session_013BXYFyne1xCkXHL55jwnA1)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -614,6 +614,59 @@ impl ZenohBind {
     }
 }
 
+/// Report whether `zenoh_bind` can be reached by remote daemons, and reject a
+/// configuration that would run silently undialable.
+///
+/// A loopback listener advertises `127.0.0.1` to peers, who dial their own
+/// loopback and reach nothing — and since zenoh 1.9 peers do not relay, that
+/// pair is dead with no fallback. When the coordinator is remote (so other
+/// daemons are expected):
+///
+/// * an *explicit* loopback address is a hard error — the operator named it, and
+///   the [`ZenohBind::Explicit`] contract is "bind a routable address or exit",
+///   the same "nothing to advertise" reason [`validate_zenoh_listen`] already
+///   rejects the wildcard for;
+/// * a *derived* loopback only warns — it is a best-effort fallback the operator
+///   can override with `--zenoh-listen`.
+///
+/// A routable bind is announced at info level; a single-machine (loopback
+/// coordinator) setup is silent.
+fn announce_zenoh_bind(zenoh_bind: ZenohBind, coordinator_ws_addr: SocketAddr) -> eyre::Result<()> {
+    if zenoh_bind.addr().is_loopback() && !coordinator_ws_addr.ip().is_loopback() {
+        match zenoh_bind {
+            ZenohBind::Explicit(addr) => {
+                eyre::bail!(
+                    "--zenoh-listen {addr} is a loopback address, but the coordinator \
+                     at {coordinator_ws_addr} is remote, so other daemons must be able \
+                     to reach this one. A loopback listener advertises {addr} to peers, \
+                     who would dial their own loopback and reach nothing. Pass the \
+                     address other daemons should use to reach this host (e.g. its LAN \
+                     or VPN address)."
+                );
+            }
+            ZenohBind::Derived(_) => {
+                tracing::warn!(
+                    "coordinator at {coordinator_ws_addr} is remote, but no routable local \
+                     address toward it was found; zenoh will bind loopback and other daemons \
+                     will not be able to reach this one. Pass --zenoh-listen <IP> explicitly."
+                );
+            }
+        }
+    } else if !zenoh_bind.addr().is_loopback() {
+        tracing::info!(
+            "zenoh listener binding {} ({}); this port accepts connections from \
+             other hosts",
+            zenoh_bind.addr(),
+            match zenoh_bind {
+                ZenohBind::Explicit(_) => "given via --zenoh-listen".to_string(),
+                ZenohBind::Derived(_) =>
+                    format!("derived from coordinator address {coordinator_ws_addr}"),
+            }
+        );
+    }
+    Ok(())
+}
+
 impl Daemon {
     /// Derives the zenoh listen address from `coordinator_ws_addr`; see
     /// [`Daemon::run_with_zenoh_listen`] to override it.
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found. The extraction of `announce_zenoh_bind` preserves the original derived-loopback/routable behavior and correctly turns an explicit loopback `--zenoh-listen` under a remote coordinator into a hard error; the four unit tests cover each branch.

---
*🤖 This is a fully automated review by Claude. It may be incorrect.*

---

PR #2774, opened 2026-07-20
fix(message): reject metadata wire-format skew at node register

phil-opp:
The `dora_version` semver check on `NodeRegisterRequest` is too coarse to catch a message-format break within a release series: two builds that both report `1.0.0-rc` pass as compatible even when the payload layout differs. #2366 dropped a `Metadata` field without a semver bump, so a stale git-pinned node registered successfully and only desynced later, mid-stream, with a cryptic `tag for enum is not valid` bincode error (#2742).

Carry the node's `Metadata::CURRENT_VERSION` in the register request and reject a mismatch against the daemon's own version with a legible message naming both sides. The check rides inside the existing `check_version()`, so no daemon change is needed.

Note the wire-format tradeoff, acceptable at 1.0.0-rc: appending the field is itself a one-time break of the register message. A node built before this change registering against a newer daemon fails to deserialize ("unexpected end of file") rather than getting the new legible error — the clean message only appears once both ends carry the field. From here on, every future format break that bumps CURRENT_VERSION is caught at register.

Refs #2742

### DIFF
--- a/libraries/message/src/node_to_daemon.rs
+++ b/libraries/message/src/node_to_daemon.rs
@@ -84,6 +84,17 @@ pub struct NodeRegisterRequest {
     pub dataflow_id: DataflowId,
     pub node_id: NodeId,
     dora_version: semver::Version,
+    /// The metadata wire-format version this node was built against
+    /// ([`Metadata::CURRENT_VERSION`]).
+    ///
+    /// The `dora_version` semver check above is too coarse to catch a
+    /// message-format break within a release series: two builds that both
+    /// report `1.0.0-rc` are "compatible" by semver even when the payload
+    /// layout differs (as it did across #2366). Carrying the format version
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. `NodeRegisterRequest` now carries `metadata_version: u16` (stamped from `Metadata::CURRENT_VERSION`, `pub const CURRENT_VERSION: u16 = 1` in `libraries/message/src/metadata.rs`), and `check_version()` returns a legible error when it diverges from the daemon's own version. The field type matches, the semver gate still runs first, and the rejection only fires on a genuine wire-format mismatch. The one-time wire-format break from appending the field (an old node against a new daemon deserializes as "unexpected end of file" rather than the new message) is acknowledged in the commit and acceptable at 1.0.0-rc. Tests cover the stamp, the accept path, and the mismatch-rejection path.

---

PR #2778, opened 2026-07-20
fix(node): bound zenoh teardown below the daemon force-kill grace

phil-opp:
The nightly `CLI Tests (windows-latest)` job (`python-operator-dataflow`) has failed every night, distinct from the Arrow IPC wire break that #2745 already fixed in the examples job.

On shutdown a node tears down its zenoh subscribers, liveliness tokens, and session under `ZENOH_TEARDOWN_TIMEOUT` — which was **15s, exactly the daemon's `DEFAULT_STOP_GRACE * 1.5 = 15s` force-kill deadline**. When the zenoh net runtime is wedged (unreachable scouted peers on a headless CI runner), each `undeclare`/close blocks to its full deadline, so the daemon `TerminateProcess`es the node ~1s before it gives up. On Windows that surfaces as `ExitCode(1)` and reddens the job; on Unix the identical kill is masked as `Signal(15)`, which is why only Windows shows it.

Lower the timeout to **3s** so a node with a wedged runtime abandons teardown and exits cleanly before the force-kill on all platforms (three sequential teardown phases → 9s worst case, comfortably under 15s). This deliberately no longer waits out zenoh's internal 10s session-close: a semi-wedged close is abandoned and peers fall back to liveliness expiry — the right trade for a shutting-down node.

Adds a guard test asserting the worst-case teardown stays under the grace, plus a back-reference on the daemon const so the two can't silently drift.

Refs #2742.

Follow-ups intentionally out of scope: hoisting the stop-grace into a shared crate so the guard tracks the real daemon value instead of a hard-coded copy, and eliminating the zenoh teardown wedge at its source (separate issue).

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -130,9 +130,25 @@ const ZENOH_OUTPUT_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
 /// Poll interval while waiting for output routes to connect at startup.
 const ZENOH_OUTPUT_CONNECT_POLL_INTERVAL: Duration = Duration::from_millis(5);
 
-/// Must exceed zenoh's internal 10s session close timeout, which is not
-/// enforceable when zenoh's net runtime is wedged.
-pub(crate) const ZENOH_TEARDOWN_TIMEOUT: Duration = Duration::from_secs(15);
+/// Per-phase deadline for tearing down zenoh state on node shutdown (subscribers,
+/// liveliness tokens, and the session/publishers — see [`DoraNode::drop`] and
+/// `EventStream::drop`). Each `undeclare`/close blocks indefinitely when zenoh's net
+/// runtime is wedged (e.g. retrying an unreachable scouted peer on a headless CI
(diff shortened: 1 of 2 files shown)


phil-opp:
🤖 **Automated review by Claude** — this review is fully automated and has not been verified by a human.

I reviewed the diff against the surrounding teardown / force-kill logic. The change is sound: lowering `ZENOH_TEARDOWN_TIMEOUT` to 3s keeps the worst-case sequential teardown (3 phases × 3s = 9s) comfortably under the daemon's 15s force-kill grace (`DEFAULT_STOP_GRACE + DEFAULT_STOP_GRACE / 2`), and the back-reference comment plus the `zenoh_teardown_fits_within_daemon_force_kill_grace` guard test make the invariant explicit. Trading a semi-wedged ~10s close for a clean exit with liveliness fallback is reasonable for a shutting-down node. No correctness issues found.

Minor, non-blocking: the guard test hardcodes `TEARDOWN_PHASES = 3`, so it can't detect a future 4th sequential teardown phase that would silently erode the margin back toward the 15s grace. Worth keeping that constant in sync if another bounded phase is ever added (the comment already calls this out).

---

PR #2780, opened 2026-07-20
test(ros2-bridge): don't reconstruct from a null pointer for empty FFI sequences

phil-opp:
## Issue

`OwnedFFISeq::from_rust` and `RefFFISeq::from_rust` (`libraries/extensions/ros2-bridge/src/_core/sequence.rs`) deliberately store a **null** `data` pointer for an empty input `Vec`:

```rust
if vec.is_empty() {
    Self { data: std::ptr::null_mut(), size: 0, capacity: 0 }
}
```

Reconstructing a `Vec`/slice from that null pointer is **library-level undefined behavior**, even though a zero-length / zero-capacity reconstruction performs no allocation or deallocation:

- `Vec`'s internal pointer is a `NonNull`, so `Vec::from_raw_parts(null, 0, 0)` violates its non-null invariant.
- `std::slice::from_raw_parts` requires a non-null, aligned pointer *even for length 0*.

Miri flags both. Three sites hit this on an empty sequence:

- `OwnedFFISeq::drop` → `Vec::from_raw_parts(self.data, 0, 0)`
- `OwnedFFISeq::as_slice` → `slice::from_raw_parts(self.data, 0)`
- `RefFFISeq::as_slice` → `slice::from_raw_parts(self.data, 0)`

Only `FFISeq::deref` already guarded the empty case (`if self.size == 0 { return &[]; }`), so the codebase already knows the hazard exists — the guard was just applied inconsistently. Empty ROS2 sequences (empty arrays in messages) are common, and `OwnedFFISeq` is the Rust→C owning-conversion type, so any empty owned/ref sequence that is sliced or dropped hits this.

## Fix

Guard all three reconstruction sites, matching the existing `FFISeq::deref` guard. This keeps the null-for-empty representation intact (so the C/ROS2 ABI is unchanged) and simply avoids ever reconstructing from the null pointer.

## Validation

- Added regression tests (`empty_owned_seq_slice_and_drop_without_ub`, `empty_ref_seq_slice_without_ub`) that build an empty sequence via `from_rust`, then slice and drop it. These pass under `cargo test` and are meaningful under `cargo +nightly miri test`, which detects the pointer-invariant violation the guards prevent.
- `cargo fmt -p dora-ros2-bridge -- --check`, `cargo clippy -p dora-ros2-bridge -- -D warnings`, and `cargo test -p dora-ros2-bridge` all pass.

---

---

### DIFF
--- a/libraries/extensions/ros2-bridge/src/_core/sequence.rs
+++ b/libraries/extensions/ros2-bridge/src/_core/sequence.rs
@@ -208,6 +208,7 @@ impl<T> FFIFromRust for RefFFISeq<T> {
 
 #[cfg(test)]
 mod tests {
+    use super::super::string::OwnedFFIString;
     use super::*;
 
     // `from_rust` deliberately stores a null `data` pointer for an empty
@@ -249,6 +250,25 @@ mod tests {
         };
         assert!(seq.is_empty());
         assert_eq!(seq.as_slice(), &[] as &[i32]);
+    }
+
+    // The `OwnedFFISeq` tests above build the null/empty representation by hand,
+    // because `OwnedFFISeq::from_rust` requires `T: FFIFromRust`. `OwnedFFIString`
+    // satisfies that bound, so this drives the *real* constructor and would catch a
+    // `from_rust` that stopped storing null for the empty case. Run under
+    // `cargo +nightly miri test` to exercise the pointer-validity checks.
+    #[test]
+    fn owned_seq_from_rust_empty_slice_and_drop_without_ub() {
+        let empty: Vec<String> = Vec::new();
+        // SAFETY: `empty` is a valid `Vec<String>` (the `From` type of
+        // `OwnedFFIString`); the produced sequence borrows nothing from it.
+        let seq: OwnedFFISeq<OwnedFFIString> = unsafe { FFIFromRust::from_rust(&empty) };
+        assert!(seq.is_empty());
+        assert_eq!(seq.len(), 0);
+        // `as_slice` on the empty (null-backed) sequence must not deref null.
+        assert!(seq.as_slice().is_empty());
+        // Dropping here must not hit `Vec::from_raw_parts(null, 0, 0)`.
+        drop(seq);
     }
 
     #[test]


phil-opp:

This PR is one of a batch opened by an automated Claude codebase-review pass. All findings were verified against `origin/main` (`c9a3699`), de-duplicated against currently-open review PRs, and each branch was cut fresh from `origin/main`. Machine-generated — please review carefully.

### PRs opened this run

**Correctness**
- #2780 — `fix(ros2-bridge)`: don't reconstruct a `Vec`/slice from the null pointer stored for empty FFI sequences (`Drop` + both `as_slice` sites; Miri-flagged UB).
- #2784 — `fix(coordinator)`: surface a daemon's WS-level error instead of returning a `{"ws_error":…}` envelope that every caller then fails to deserialize (dropping the real error text).

**Optimization**
- #2781 — `perf(core)`: resolve a wiring input's source node via `nodes.get(source)` instead of an O(N) scan, removing the O(nodes²·inputs) blowup in `check_wiring`.
- #2782 — `perf(daemon)`: drop the per-log-line `std::env::var("DORA_QUIET")` read and the per-line `content.clone()` in the stdout/stderr reader tasks.

**Documentation**
- #2783 — `docs(c-api)`: fix `read_dora_input_timestamp`'s misplaced `## Safety` note (missing the pointer-validity contract) and document that `read_dora_input_data` also returns `(NULL, 0)` for unsupported payload types.

**Simplification** — none met the bar this run.

### Verification
Each branch passed `cargo fmt`/`clippy -D warnings`/tests for its affected crate. The combined diff also passed the full workspace test suite, `cargo check --examples`, and a 4-angle `/simplify` review (the altitude pass prompted widening #2780 from the `Drop`-only fix to all three null-reconstruction sites). One unrelated pre-existing failure was observed — `dora-node-api`'s `zenoh_schema_cache` test cannot bind an IPv6 listener in this sandbox (`Address family not supported by protocol`); it is environmental and untouched by any of these diffs.

### Findings considered but rejected
- **download fallback filename keeps URL query string** (`libraries/extensions/download`) — already covered by open PRs #2744 / #2729. Duplicate.
- **Duplicate byte-size parsers** (`parse_byte_size` in core vs `ByteSize::from_str` in dora-message) — real duplication, but delegating is not strictly behavior-preserving (`is_ascii_alphabetic` vs `is_alphabetic`, `u64` vs `usize`, differing error text); risk outweighed the simplification benefit.
- **C++ `Metadata::to_json` `.expect` can abort across the cxx bridge** — near-unreachable (serde_json does not error for this value shape) and the clean fix changes the exported cxx signature (a C++ API change); low value.
- **`dataflow_result` clones each daemon's `node_results` map** (`coordinator/handlers.rs`) — runs once per dataflow completion, not a hot path; too minor to warrant a standalone PR.

---

phil-opp:

No issues found.

I verified the guards against the unchanged `from_rust` implementations: both `OwnedFFISeq::from_rust` and `RefFFISeq::from_rust` set `data = null`, `size = 0`, `capacity = 0` for an empty input, and the non-empty path always yields `size == capacity` with a non-null pointer. So `as_slice`'s `size == 0` guard and `OwnedFFISeq::drop`'s `capacity == 0` guard both fire exactly on the null-pointer case and never skip a real allocation — this removes the `slice::from_raw_parts(null, 0)` / `Vec::from_raw_parts(null, 0, 0)` library-level UB that Miri flags, matching the existing `FFISeq::deref` guard.

---

phil-opp:

This change appears to have already landed on `main`, which is likely why the branch now shows as conflicting. #2677 (`fix(ros2-bridge): avoid null-pointer slice UB in Owned/RefFFISeq::as_slice`) merged on 2026-07-22 and added the same guards in `libraries/extensions/ros2-bridge/src/_core/sequence.rs`:

- `OwnedFFISeq::as_slice` and `RefFFISeq::as_slice` both early-return `&[]` on `self.size == 0`.
- `OwnedFFISeq::drop` early-returns on `self.data.is_null()` (equivalent to this PR's `capacity == 0` check, since `from_rust` sets `data = null` and `capacity = 0` together for the empty case).

It also carries a superset of the regression tests, including a non-empty `Drop`-frees case. As far as I can tell this PR is now redundant and could be closed.

---

phil-opp:
Re-reviewed after the recent rebase — flagging that the diff has changed materially since the earlier "this looks redundant, could be closed" note, so that verdict no longer applies.

The null-pointer guards this PR originally added have since landed on `main` via #2677, and this branch was rebased on top, dropping those hunks. The current diff is now **test-only**: it adds `owned_seq_from_rust_empty_slice_and_drop_without_ub` (plus its `OwnedFFIString` import).

That remaining test is worth keeping and is not redundant with #2677's coverage: the existing empty-case tests either build the `OwnedFFISeq` struct literal by hand or use `RefFFISeq<i32>`, whereas this one drives the *real* `OwnedFFISeq::<OwnedFFIString>::from_rust` constructor on an empty `Vec<String>`, then slices and drops it — i.e. it exercises the exact path that stores the null `data` pointer, so it would catch a future regression where `from_rust` stopped null-guarding the empty case (meaningful under Miri).

No correctness concerns with the diff as it stands. Might make sense to retarget it as a test-only follow-up to #2677.

---
_🤖 This is a fully automated review by Claude. Generated by [Claude Code](https://claude.ai/code); no human has reviewed these findings._

---

PR #2782, opened 2026-07-21
perf(daemon): drop per-log-line env read and content clones in the log path

phil-opp:
## Issue

The per-node stdout/stderr logging pipeline in `binaries/daemon/src/spawn/prepared.rs` does redundant work on every log line — the hot path for any chatty node:

1. **Per-line `std::env::var("DORA_QUIET")`.** The file-logging loop (`while let Some(log_line) = rx.recv().await`) called `std::env::var("DORA_QUIET")` on every forwarded line. `std::env::var` takes the process-global environment lock and allocates a `String` on each call — and the value cannot change during the task's lifetime. (Ironic given the adjacent comment that deliberately avoids per-line `fsync` for the same reason.)

2. **Per-line `content.clone()` in both reader tasks.** The stdout and stderr reader tasks each cloned `content` on every line solely so the original stayed available for a `tracing::warn!("Could not log: …")` that only fires when the log channel is already closed (a rare, terminal condition).

## Fix

1. Read `DORA_QUIET` once before the loop: `let quiet = std::env::var_os("DORA_QUIET").is_some();` (`var_os` also skips UTF-8 validation/allocation), and test the captured `quiet` bool in the loop.

2. Move `content` into the `LogLine` instead of cloning; on the rare `SendError` (closed channel), recover the value from the returned `mpsc::error::SendError` for the warning. The stderr path keeps its single, unavoidable clone for the crash-diagnostics ring buffer (`force_push`).

No behavior change on the success path.

## Validation

- `cargo fmt -p dora-daemon -- --check`, `cargo clippy -p dora-daemon -- -D warnings`, and `cargo test -p dora-daemon --lib` (106 tests) all pass.

---

---

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -662,14 +662,17 @@ impl PreparedNode {
                     }
                 };
                 truncate_log_line(&mut content);
-                let sent = stdout_tx
+                // Move `content` into the `LogLine`; on the rare closed-channel
+                // error, recover it from the returned `SendError` for the
+                // warning instead of cloning on every line.
+                if let Err(mpsc::error::SendError(unsent)) = stdout_tx
                     .send(LogLine {
-                        content: content.clone(),
+                        content,
                         stream: LogStream::Stdout,
                     })
-                    .await;
-                if sent.is_err() {
-                    tracing::warn!("Could not log: {content}");
+                    .await
+                {
+                    tracing::warn!("Could not log: {}", unsent.content);
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The three optimizations are behavior-preserving on the relevant paths:

- Moving `content` into the `LogLine` and recovering it from `mpsc::error::SendError(unsent)` for the warning is correct — tokio's `SendError` returns the un-sent value, so `unsent.content` is exactly the line that would have been logged. The stderr side correctly keeps its one unavoidable `force_push` clone for the crash-diagnostics ring buffer.
- Hoisting the `DORA_QUIET` read out of the per-line loop is sound: the var can't change during the task and the captured bool is read-only in the closure.

One minor semantic nuance, non-blocking: `env::var("DORA_QUIET").is_err()` was true both when the var was absent *and* when it held non-UTF-8 bytes, whereas `env::var_os(...).is_some()` treats a non-UTF-8 value as "set" (quiet). This only differs for a `DORA_QUIET` set to invalid Unicode — an implausible configuration, and arguably the new behavior is the more intuitive one — so not a real concern.

---

PR #2783, opened 2026-07-21
docs(c-api): fix input-event doc/safety comments in the C node API

phil-opp:
## Issue

Two documentation defects in the C node API (`apis/c/node/src/lib.rs`):

1. **`read_dora_input_timestamp` has a wrong `## Safety` section.** It put a return-value note under `## Safety`:

   ```rust
   /// ## Safety
   ///
   /// Return `0` if the given event is not an input event.
   pub unsafe extern "C" fn read_dora_input_timestamp(event: *const ()) -> ... {
       let event: &Event = unsafe { &*event.cast() };
   ```

   The function dereferences the caller-supplied raw pointer, so the real safety contract is the pointer-validity rule that **every** sibling accessor documents (`read_dora_input_id`, `read_dora_input_data`, `read_dora_event_type`). This one omitted it — the one piece of documentation that governs UB for the `unsafe extern "C"` function.

2. **`read_dora_input_data` doc omits the unsupported-type behavior.** The doc says null/`0` is written "if the given event is not an input event or when an input event has no associated data." But the implementation *also* writes null/`0` (and logs an error) for any Arrow payload type other than `UInt8`/`Null` (the `other =>` arm). A C caller couldn't tell "metadata-only message" from "payload exists but is an unsupported type", and the documented contract never mentioned that path.

## Fix

1. Move the return-value note into `read_dora_input_timestamp`'s description and add the standard pointer-validity safety paragraph.
2. Document that a non-`UInt8` payload also yields `(NULL, 0)` (with a logged error), so a null result is not proof of "no data".

Documentation only; no behavior change.

## Validation

- `cargo fmt -p dora-node-api-c -- --check` and `cargo clippy -p dora-node-api-c -- -D warnings` pass. The doc changes reuse the same intra-doc link style (`[dora_next_event]`) already used by the sibling functions in this file.

---

---

### DIFF
--- a/apis/c/node/src/lib.rs
+++ b/apis/c/node/src/lib.rs
@@ -161,6 +161,12 @@ pub unsafe extern "C" fn read_dora_input_id(
 /// Writes a null pointer and length `0` if the given event is not an input event
 /// or when an input event has no associated data.
 ///
+/// The raw-byte C API can only expose `UInt8` (and empty `Null`) payloads.
+/// A payload of any other Arrow type (for example an `Int32`/`Float` array
+/// from another node) also yields a null pointer and length `0` (and logs an
+/// error), so a null result is not by itself proof that the message carried no
+/// data.
+///
 /// ## Safety
 ///
 /// The `event` argument must be a dora event received through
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Documentation-only change, and both edits are accurate against the implementation:

- `read_dora_input_data`: the added note that a null result also occurs (with a logged error) for any non-`UInt8`/non-`Null` Arrow payload matches the body, which only handles `UInt8`/`Null` and returns `(NULL, 0)` for other types. Correctly warns callers that a null result is not proof of "no data".
- `read_dora_input_timestamp`: moving the "returns `0` if not an input event" note out of `## Safety` into the description, and adding the standard pointer-validity contract, is correct — the body does `&*event.cast()`, so the pointer-validity precondition genuinely belonged in `## Safety`.

No behavior change.

---

PR #2784, opened 2026-07-21
fix(coordinator): surface daemon WS errors instead of an undeserializable envelope

phil-opp:
## Issue

When a daemon replies with a WS-level `error`, `handle_daemon_response` (`binaries/coordinator/src/ws_daemon.rs`) flattens it into a JSON string and pushes it through the pending-reply channel:

```rust
} else if let Some(err) = response.error {
    tracing::warn!("daemon returned error for request {}: {err}", response.id);
    serde_json::json!({"ws_error": err}).to_string()
```

`DaemonConnection::send_and_receive` returned those bytes as `Ok(...)`, and every caller (`stop_dataflow`, `reload_dataflow`, `build_dataflow`, node commands, logs, destroy) then does `serde_json::from_slice::<DaemonCoordinatorReply>(&reply_raw)`. A `{"ws_error": "..."}` object matches **no** `DaemonCoordinatorReply` variant, so the caller fails with a generic *"failed to deserialize \<op\> reply from daemon"* and the daemon's real error text is dropped from the returned error chain — it survives only in a `warn!` log line. The operator running `dora stop`/`reload`/`build` sees an opaque deserialization message instead of the actual cause.

This is reachable today: the daemon emits `WsResponse::err(request_id, …)` when it fails to serialize a reply (`binaries/daemon/src/coordinator.rs`).

## Fix

Detect the `{"ws_error": ...}` envelope in `send_and_receive` and return it as an `Err` carrying the daemon's text, so **all** callers propagate the real cause. A dedicated `#[serde(deny_unknown_fields)]` single-field struct makes the match exact, so it can never collide with a real `DaemonCoordinatorReply` payload (whose variants serialize with their own tags). The success path is unchanged — the raw JSON bytes still flow through (preserving the u128-fidelity rationale for raw-JSON passthrough).

### Note on depth

A deeper fix would change the pending-reply channel payload from `oneshot::Sender<String>` to `Result<String, String>`, carrying the error *as* an error across the boundary and removing the need for the sentinel envelope entirely. That change threads through `state.rs`, `ws_daemon.rs`, `run/mod.rs`, and ~15 test construction sites, so it's left as a follow-up; this localized fix is behavior-safe and self-contained.

## Validation

- Added `ws_error_envelope_surfaces_as_err`, which drives the full `send_and_receive` path with an injected `ws_error` reply and asserts the returned error carries the daemon's message.
- `cargo fmt -p dora-coordinator -- --check`, `cargo clippy -p dora-coordinator -- -D warnings`, and `cargo test -p dora-coordinator` (incl. the new test) all pass.

---

---

### DIFF
--- a/binaries/coordinator/src/state.rs
+++ b/binaries/coordinator/src/state.rs
@@ -113,6 +113,15 @@ pub(crate) struct DaemonConnection {
     pub(crate) connection_id: Uuid,
 }
 
+/// The envelope `handle_daemon_response` (see `ws_daemon.rs`) produces when a
+/// daemon replies with a WS-level `error` instead of a result. `deny_unknown_fields`
+/// keeps it from ever matching a real `DaemonCoordinatorReply` payload.
+#[derive(serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+struct WsErrorEnvelope {
+    ws_error: String,
+}
+
 impl DaemonConnection {
     pub(crate) fn new(
         sender: mpsc::Sender<String>,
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Detecting the `{"ws_error": ...}` envelope in `send_and_receive` and returning it as an `Err` carrying the daemon's text is a strict improvement: callers previously hit an opaque "failed to deserialize … reply" (the envelope matches no `DaemonCoordinatorReply` variant), losing the real cause. The dedicated `#[serde(deny_unknown_fields)]` `WsErrorEnvelope` cannot collide with a real reply, which is enum-tagged and never a bare single-field `ws_error` object, so there is no false-positive risk. No behavioral regression for callers, since the prior path already failed.

---

PR #2786, opened 2026-07-21
fix(download): match Content-Disposition filename param case-insensitively

### LINKED ISSUE #2609: download_file: Content-Disposition filename parsing is case-sensitive, silently dropping RFC-legal `Filename=` headers

phil-opp:
## Summary

Fixes #2609.

`parse_content_disposition_filename` in `dora-download` matched the header parameter name with a **case-sensitive** `split("filename=")`. Per [RFC 6266](https://www.rfc-editor.org/rfc/rfc6266), `Content-Disposition` parameter names are **case-insensitive**, so a spec-compliant server sending `filename` in any other case (`Filename`, `FILENAME`, `FileName` — common from some Java/.NET/IIS stacks) had its filename **silently dropped**. The download then fell through to the URL-derived fallback path (which has its own separate issue, #2600), producing a wrong on-disk name or an outright `"Could not find a filename"` error even though the server supplied the correct name.

## Changes

- Match the `filename=` parameter name **case-insensitively** by locating it in an ASCII-lowercased copy of the header, then slicing the value from the original (case-preserving) header.
- The literal `filename=` substring never matches the RFC 5987 extended `filename*=` spelling, so that form is still skipped and falls through to `None` (existing behavior, covered by an existing test).
- Honor `\`-escaped quotes inside the quoted-string form (RFC 6266 quoted-string), so `filename="a\"b.bin"` yields `a"b.bin` instead of truncating at the escaped quote.

## Tests

Added two regression tests in the existing `#[cfg(test)]` block:
- `parameter_name_is_case_insensitive` — `Filename=`, `FILENAME=`, `FileName=` all parse.
- `escaped_quote_inside_quoted_value` — `filename="a\"b.bin"` → `a"b.bin`.

All 7 unit tests pass; the pre-existing cases (quoted/unquoted, trailing params, semicolon-inside-quotes, empty/`filename*=` → `None`) are unchanged. `cargo fmt --check` and `cargo clippy -p dora-download -- -D warnings` are clean.

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -9,19 +9,37 @@ use tokio::io::AsyncWriteExt;
 ///
 /// Handles headers that carry additional parameters after the filename, e.g.
 /// `attachment; filename="model.bin"; size=1000`, and both quoted and unquoted
-/// forms. Returns `None` when no non-empty `filename=` value is present (the
-/// RFC 5987 extended `filename*=` form is not decoded and falls through).
+/// forms. Per RFC 6266 the parameter name is matched case-insensitively, so
+/// `Filename=`/`FILENAME=` are accepted too. Returns `None` when no non-empty
+/// `filename=` value is present (the RFC 5987 extended `filename*=` form is not
+/// decoded and falls through — the literal `filename=` substring never matches
+/// the `filename*=` spelling, so it is skipped automatically).
 fn parse_content_disposition_filename(header: &str) -> Option<String> {
-    let rest = header.split("filename=").nth(1)?.trim_start();
+    // RFC 6266: parameter names are case-insensitive. Locate `filename=`
+    // regardless of case without copying the value bytes.
+    let lower = header.to_ascii_lowercase();
+    let value_start = lower.find("filename=")? + "filename=".len();
+    let rest = header[value_start..].trim_start();
+
     let name = if let Some(after_quote) = rest.strip_prefix('"') {
-        // Quoted form: the value runs up to the closing quote, so a trailing
-        // `; param=...` (or a `;` inside the quotes) is handled correctly.
-        after_quote.split('"').next().unwrap_or(after_quote)
+        // Quoted form: read up to the closing quote, honoring RFC 6266
+        // `\`-escapes (`filename="a\"b"` -> `a"b`), so a trailing
+        // `; param=...` (or a `;`/`"` inside the quotes) is handled correctly.
+        let mut name = String::new();
+        let mut chars = after_quote.chars();
+        while let Some(c) = chars.next() {
+            match c {
+                '"' => break,
+                '\\' => name.extend(chars.next()),
+                _ => name.push(c),
+            }
+        }
+        name
     } else {
         // Token form: the value ends at the next parameter separator.
-        rest.split(';').next().unwrap_or(rest).trim()
+        rest.split(';').next().unwrap_or(rest).trim().to_string()
     };
-    (!name.is_empty()).then(|| name.to_string())
+    (!name.is_empty()).then_some(name)
 }
 
 fn get_filename(response: &reqwest::Response) -> Option<String> {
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

`parse_content_disposition_filename` now locates the parameter case-insensitively via a lowercased copy and indexes back into the original header. This is byte-safe: `to_ascii_lowercase` is a 1:1 byte-length-preserving mapping, so the offset from `lower.find("filename=")` lands on a valid boundary in `header`, and the following byte after the ASCII `filename=` is always a char boundary. The `filename*=` extended form still falls through (the literal `filename=` substring never matches `filename*=`), and the added `\`-escape handling in the quoted form is correct.

---

phil-opp:

The case-insensitive matching here looks correct and byte-safe (`to_ascii_lowercase` preserves byte length so the `find("filename=")` offset stays a valid char boundary, and `filename*=` still correctly falls through since the literal `filename=` substring never matches it). The `\`-escape handling in the quoted-string branch is a nice RFC 6266 improvement.

One heads-up: #2610 (`2580969`) already landed a case-insensitive fix on `main` using the same `to_ascii_lowercase()` + `find("filename=")` shape — but *without* the quoted-string `\`-escape handling. So the case-insensitivity part of this PR is now redundant (and conflicts), and after a rebase the unique remaining value is just the escaped-quote support (`filename="a\"b.bin"` → `a"b.bin`). Might be worth reducing this to just that delta, or confirming it isn't superseded.

Minor, pre-existing: per RFC 6266 §4.3 recipients should prefer `filename*` (extended) over `filename` when both are present; this still only reads `filename=`, so a non-ASCII name carried in `filename*` gets dropped for the ASCII fallback — the doc comment notes this is intentional, just flagging it.

---

phil-opp:
Superseded by #2823

PR #2789, opened 2026-07-21
fix(core): register the admitted TypeDef, not a body-invalid same-URN duplicate

### LINKED ISSUE #2599: Manifest type injection registers the wrong (possibly body-invalid) TypeDef when two manifests ship the same URN

phil-opp:
## Summary

Fixes #2599.

In `libraries/core/src/manifest/inject.rs`, the loop that registers shipped types into the **shared** registry could store a **body-invalid** definition when two matched manifests ship the same type URN.

## The bug

`candidates` is a flat `Vec<(&urn, &def)>` accumulated across **all** matched manifests and is **not** deduplicated by URN. The admission fixpoint decides, per URN, which **body-valid** definition enters the scratch registry. But the registration loop then wrote each candidate's **own** `def`, guarded only by `scratch.resolve(urn).is_some()` — which only proves that *some* definition for that URN was admitted, not that *this* candidate's `def` is the admitted one.

So with two same-namespace manifests shipping the same URN and the **body-invalid one listed first**, the loop registered the invalid `def`; the second (valid) candidate was then skipped because `registry.resolve(urn)` was already `Some`. A referencing node's port check passed against a type that can never be materialized into an Arrow schema — exactly the invariant the surrounding fixpoint and its tests exist to protect.

## Fix

Register the **admitted** definition from `scratch` rather than the candidate's own `def`:

```rust
for &(urn, _def) in &candidates {
    if registry.resolve(urn).is_none()
        && let Some(admitted) = scratch.resolve(urn)
    {
        let _ = registry.add_user_type(urn, admitted.clone());
    }
}
```

`scratch` only ever holds the body-valid definition for each URN (the fixpoint inserts nothing else), so this guarantees the shared registry receives the body-valid one regardless of candidate ordering.

## Tests

Added `duplicate_urn_registers_the_body_valid_definition`: two same-namespace manifests ship `acme/foo/v1/X`, body-invalid (`broken: NotARealType`) listed first, body-valid (`x: Float32`) second, plus a referencing node. The test asserts the registry holds the valid body (field `x`, not `NotARealType`) and the referencing node's output type materializes. It fails against the old code (which registered the invalid duplicate). All 18 `manifest::inject` tests pass; `cargo fmt --check` and `cargo clippy -p dora-core -- -D warnings` are clean.

---

### DIFF
--- a/libraries/core/src/manifest/inject.rs
+++ b/libraries/core/src/manifest/inject.rs
@@ -120,10 +120,17 @@ pub fn inject_adjacent_manifests(
         }
     }
     // Register the survivors into the shared registry, skipping URNs already
-    // present so preserved definitions win.
-    for &(urn, def) in &candidates {
-        if scratch.resolve(urn).is_some() && registry.resolve(urn).is_none() {
-            let _ = registry.add_user_type(urn, def.clone());
+    // present so preserved definitions win. Register the *admitted* definition
+    // from `scratch`, not the candidate's own `def`: `candidates` is not
+    // deduplicated by URN, so when two same-namespace manifests ship the same
+    // URN, `scratch.resolve(urn).is_some()` only proves that *some* body-valid
+    // definition was admitted — the first-listed candidate's `def` may be the
+    // body-invalid duplicate, which must not reach the shared registry (#2599).
+    for &(urn, _def) in &candidates {
+        if registry.resolve(urn).is_none()
+            && let Some(admitted) = scratch.resolve(urn)
+        {
+            let _ = registry.add_user_type(urn, admitted.clone());
         }
     }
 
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

I checked the fix in `libraries/core/src/manifest/inject.rs`: the injection loop now registers the fixpoint-admitted definition from `scratch.resolve(urn)` rather than the candidate's own `def`. Since `candidates` is not deduplicated by URN, `scratch.resolve(urn).is_some()` alone never guaranteed the first-listed candidate's `def` was the body-valid one; registering `scratch`'s admitted def is the correct fix. The `registry.resolve(urn).is_none()` guard still preserves already-registered types, and the regression test correctly exercises the body-invalid-first ordering.

---

phil-opp:

The fix in this PR appears to have **already landed on `main`**, so relative to `main` this change is effectively a no-op.

- `libraries/core/src/manifest/inject.rs` on current `main` already contains the exact registration loop this PR adds — byte-for-byte identical logic:
  ```rust
  for &(urn, _def) in &candidates {
      if registry.resolve(urn).is_none()
          && let Some(admitted) = scratch.resolve(urn)
      {
          let _ = registry.add_user_type(urn, admitted.clone());
      }
  }
  ```
  It landed via `bd3d565` — *"fix(core): register the admitted type body for a duplicated manifest URN (#2605)"*.
- `main` also already carries a regression test for this (`duplicate_urn_registers_the_body_valid_definition_not_the_first_listed`), distinct from this PR's `duplicate_urn_registers_the_body_valid_definition`.

The logic itself is correct — I verified the URN-dedup path and the body-invalid-first ordering the test exercises, and it matches what's on `main`. The only issue is redundancy: this PR makes no production change on top of `main` and only adds a near-duplicate test, and because it rewrites the same `inject_adjacent_manifests` hunk that `main` already rewrote it will conflict on rebase. Suggest rebasing onto latest `main` to confirm #2605 already covers #2599 — if so, this can likely be closed.

---

phil-opp:
Superseded by #2605, which landed the same fix in main: the registration loop now registers the *admitted* `TypeDef` from `scratch` rather than the candidate's own (possibly body-invalid, same-URN duplicate) `def`. The code is identical and the regression test is equivalent. Rebasing this branch onto main produces an empty diff, so closing.

PR #2790, opened 2026-07-21
feat(ros2-bridge): add native rmw_zenoh_cpp transport

### LINKED ISSUE #2735: ROS2 bridge: support rmw_zenoh_cpp (Zenoh transport), not just RTPS/DDS
Summary

    dora's built-in ROS2 bridge (Ros2Context / Ros2Node / Ros2Subscription / Ros2Publisher) speaks RTPS/DDS via its Rust stack (rustdds / ros2-client). This means it can
    only interoperate with ROS 2 nodes running a DDS-based RMW (rmw_fastrtps_cpp, rmw_cyclonedds_cpp). It cannot see topics whose transport is Zenoh (rmw_zenoh_cpp),
    because Zenoh is a different wire protocol, not RTPS.

    Why this matters

    rmw_zenoh_cpp has become an official tier-1 RMW and is increasingly the default on real robots — it's used across many mobile/navigation stacks today (e.g.
    nav2-based deployments) because of its discovery and multi-robot/WAN behavior. On those robots, a dora node using the built-in ROS2 bridge silently receives nothing:
    the topics exist, discovery just never matches, so it fails quietly rather than erroring.

    Concretely, we hit this trying to bridge a localization odometry topic (nav_msgs/Odometry) into a dora dataflow. The stack defaults to:

    export RMW_IMPLEMENTATION="${RMW_IMPLEMENTATION:-rmw_zenoh_cpp}"

    and dora's ROS2 bridge can't observe it. The only robust workaround today is to run an rclpy subscriber inside a dora node and source the ROS 2 install so rclpy
    loads the matching RMW — which reintroduces the ROS 2 / rclpy dependency the native bridge was meant to avoid.

    Request

    Add a Zenoh transport option to dora's ROS2 bridge so it can interoperate with rmw_zenoh_cpp peers — ideally selectable per context/node (e.g. an RTPS-vs-Zenoh
    transport choice or QoS/transport config on Ros2NodeOptions), matching the RMW the rest of the ROS 2 graph is using.

    Notes / possible directions

    - Zenoh already has a Rust implementation (zenoh crate) and the rmw_zenoh project defines the topic/key-expression mapping and attachment/QoS conventions ROS 2 uses
    over Zenoh — dora could reuse that mapping rather than reimplement it.
    - Interop must follow rmw_zenoh's key-expression naming and message-attachment format so discovery and type negotiation line up with real ROS 2 nodes.
    - Even a subscribe-only Zenoh path would unblock the most common use case (pulling robot state/telemetry into a dora dataflow).

    Environment

    - dora ROS2 bridge (Rust/DDS, RTPS-only)
    - ROS 2 Humble, rmw_zenoh_cpp
    - Topic: nav_msgs/Odometry

bobdingAI:
## Summary

Adds native Zenoh transport support to Dora's ROS 2 bridge, allowing Dora
nodes to interoperate directly with ROS 2 nodes using `rmw_zenoh_cpp`.

This addresses the silent interoperability failure described in #2735, where
Dora's DDS/RTPS-only bridge could not discover or receive data from ROS 2
graphs using Zenoh.

## What changed

- Add selectable DDS and Zenoh ROS 2 transport backends.
- Implement `rmw_zenoh_cpp`-compatible discovery, graph queries, topics,
  services, actions, key expressions, attachments, and QoS behavior.
- Add transport selection to the Rust, Python, C++, YAML, and bridge-node
  surfaces.
- Add ROS type-description parsing for Zenoh type negotiation.
- Harden the bridge-node / action-server / Python paths against malformed or
  expired peer requests, and fix the QoS liveliness wire encoding (see the
  review-round hardening under Validation).
- Add Zenoh dataflow examples and documentation.
- Add protocol fixtures and integration tests.
- Add native Humble and Kilted `rmw_zenoh_cpp` interoperability matrices.
- Support architecture-specific ROS package revisions on AMD64 and ARM64.
- Make the interop harness respect `CARGO_TARGET_DIR`, select Python 3.11+,
  and provision PyArrow in its isolated test environment.

## Motivation

`rmw_zenoh_cpp` uses Zenoh rather than RTPS/DDS, so the existing bridge cannot
communicate with it at the wire-protocol level. This affects real robot
deployments, including the Humble localization/odometry use case reported in
#2735.

## Validation

**Class**: C

All Class C gates complete (Linux; branch rebased onto latest `main`):

- `make qa-fast` (fmt / clippy `-D warnings` / cargo-audit + cargo-deny /
  unwrap-budget / typos): ✅
- Full workspace tests (`cargo test --workspace`, Python crates excluded): ✅
- Fault-tolerance E2E (`cargo test --test fault-tolerance-e2e -- --test-threads=1`):
  ✅ 11 passed
- Semantic contract tests (`example-smoke contract_*`): ✅ 8/8
- `cargo check --all` + `cargo check --examples` (#1680): ✅
- ROS 2 Humble `rmw_zenoh_cpp` interoperability: ✅ 10/10 cases
- ROS 2 Kilted `rmw_zenoh_cpp` interoperability: ✅ 10/10 cases
- GitHub Actions CI on this head: ✅ cheap PR gate **and** full tier
  (`Test`, `E2E`, semantic contracts, benchmark) all green

Interop coverage for both distributions: topic publisher/subscriber, service
client/server, action client/server, graph discovery, domain isolation,
namespace handling, transient-local QoS.

### Review-round hardening

Addressing the automated review findings on untrusted-network handling and
`rmw_zenoh_cpp` parity:

- Bridge-node reply/decode paths no longer `?`-propagate a single malformed or
  expired peer request out of the event loop (remote-DoS hardening) — malformed
  request payloads are rejected and the loop continues, matching the existing
  goal-handler pattern. Regression test included.
- Standalone Zenoh action server retires goals after result delivery and
  reclaims completed goals when at capacity, so it no longer stops accepting
  work after a lifetime of `MAX_CONCURRENT_GOALS` goals nor retains results and
  status entries indefinitely.
- QoS key expression encodes the liveliness kind in the first subfield of the
  trailing group (`<kind>,<lease_sec>,<lease_nsec>`), matching `rmw_zenoh_cpp`;
  a manual-liveliness regression test guards it.
- Python Zenoh action server skips a malformed `get_result` query and keeps
  draining instead of aborting a legitimate pending result.

### Adversarial review

The untrusted-input parsers (attachment / key-expression / graph / pub-sub
ingest with pre-decode size caps) were reviewed for malformed-input handling;
the findings above were fixed. The crate contains no `unsafe`.

## Notes

The Zenoh transport follows `rmw_zenoh_cpp` key-expression, discovery,
attachment, and QoS conventions. DDS remains available as the existing/default
transport path.

Closes #2735


phil-opp:

I reviewed the new Zenoh transport, focusing on how untrusted network payloads are handled. The Zenoh keyexpr / attachment / pubsub / service parsing paths all handle malformed input gracefully (size-checked decodes, deserialize errors mapped to `Malformed`, subscriber callbacks swallow errors). But the standalone bridge node has a set of decode sites that `?`-propagate errors out of the main event loop, terminating the whole bridge process on a single malformed, peer-controlled request:

**`binaries/ros2-bridge-node/src/main.rs` — a malformed request payload from any peer kills the bridge (remote DoS):**
- `run_zenoh_service_mode`, Server role: `let data = deserialize_raw_cdr(&request.payload, …)?;` — the `?` escapes the `block_on_stream` loop and returns out of the fn (which is `-> eyre::Result<()>`), tearing down the service-server bridge.
- `run_zenoh_action_server`: `let decoded: GetResultRequest = deserialize_cdr(&request.payload)?;` (~`main.rs:718`) and `let decoded: CancelGoalRequestWire = deserialize_cdr(&request.payload)?;` (~`main.rs:733`) — same, for `get_result` / `cancel_goal` requests.

`request.payload` is attacker-controlled (only the attachment is validated, never the payload bytes), so any non-CDR / truncated / trailing-byte payload sent to these keys terminates the bridge. The sibling **Goal** handler in the same function (~`main.rs:682`) already does the right thing: on a decode error it calls `server.send_goal.reject(request.id, …)` and `continue`s. The three sites above should follow that reject-and-continue pattern rather than `?`.

Lower-severity, same file: the action-client loop (`run_zenoh_action_client`, ~`main.rs:561`–`564`) also `?`-propagates `client.send_goal.call(...)?` and `SendGoalResponse = deserialize_cdr(&response)?`, so a single goal-send timeout or a malformed server response tears down the whole client bridge instead of dropping that one goal — inconsistent with the graceful feedback/result handling elsewhere.

The decoders themselves (`deserialize_raw_cdr` / `deserialize_cdr`) are bounded and return `Result` (no panic/OOM), so this is purely the callers' error propagation.

---

phil-opp:

Re-review of the force-pushed revision (`66a9e4a`). The remote-DoS issues I raised earlier are **still unfixed** in `binaries/ros2-bridge-node/src/main.rs`: a single malformed, peer-controlled request payload `?`-propagates out of the `block_on_stream` event loop and tears down the whole bridge task (each of these fns returns `eyre::Result<()>`).

Still-outstanding sites:

- `run_zenoh_service_mode`, **Server** role (main.rs:350) — `let data = deserialize_raw_cdr(&request.payload, request_info.clone(), 64 * 1024 * 1024)?;`. `request.payload` is attacker-controlled; a non-CDR/truncated payload kills the service-server bridge.
- `run_zenoh_action_server`, **ResultRequest** handler (main.rs:718) — `let decoded: GetResultRequest = deserialize_cdr(&request.payload)?;`
- `run_zenoh_action_server`, **Cancel** handler (main.rs:733) — `let decoded: CancelGoalRequestWire = deserialize_cdr(&request.payload)?;`

For contrast, the sibling **Goal** handler (main.rs:680-692) in the same function already does the right thing — on a decode error it calls `server.send_goal.reject(request.id, ...)` and `continue`s. The three sites above should follow that same reject-and-continue pattern instead of `?`.

Lower severity, same file — `run_zenoh_action_client` (main.rs:561-564): `client.send_goal.call(request, ACTION_GOAL_TIMEOUT)?` and `let response: SendGoalResponse = deserialize_cdr(&response)?;` still tear down the entire client bridge on a single goal-send timeout or malformed server response. This is inconsistent with the graceful `tracing::warn!`+`return`/`continue` handling used for the get-result path in the same function (main.rs:585-600).

---

phil-opp:

One new finding on `66a9e4a` beyond what the earlier automated reviews raised:

**Standalone Zenoh action server never releases completed goals (`binaries/ros2-bridge-node/src/main.rs`, `run_zenoh_action_server`).** Goals are stored in `GoalSlots`, and acceptance is gated by `goals.len() < MAX_CONCURRENT_GOALS`, but the function only ever calls `goals.insert` — there is no `remove`/`retain`/eviction anywhere in the body. Completed goals (result stored, `get_result` already answered) stay in the map for the server's whole lifetime. Consequences:
- After `MAX_CONCURRENT_GOALS` (8) goals *over the entire lifetime* — not concurrently — `goals.len()` hits the cap and every subsequent goal is silently rejected (`accepted = false`), so a long-running action server stops accepting work.
- Each retained goal keeps its serialized `result: Vec<u8>` alive (count-bounded, but unbounded-per-goal retention).
- `publish_zenoh_status` re-publishes all accumulated terminal goals indefinitely.

This is a regression relative to the two paths that get it right: the DDS `run_action_server` retires goals on completion, and the Python Zenoh server (`python/src/lib.rs`) does `self.executing.remove(goal_id)` in `send_result` plus an evict-oldest loop in `take_goal`. The standalone Zenoh server has neither.

Separately, the remote-DoS `?`-propagation sites the previous two automated reviews flagged (service-server payload decode, action get_result/cancel decode, action-client goal-send/response) are **still present unchanged** on this SHA. The service-reply-after-expiry path (`run_zenoh_service_mode` server, `server.reply(id, &response)?`) is the same class — a `RequestExpired`/`RequestAbsent` from a slow local handler propagates out of the loop and tears down the bridge. All of these want the reject-and-`continue` pattern the Goal handler already uses.

Untrusted-input parsers (`attachment.rs`, `keyexpr.rs`/`graph.rs`, `pubsub.rs` ingest with its pre-decode size cap) and the atomic sequence/limit accounting look sound; no memory-safety issues (crate has no `unsafe`).

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review, with no human in the loop. Please verify before acting on it.

One new correctness finding, not covered by the earlier automated reviews:

**QoS liveliness is encoded into the wrong keyexpr subfield** — `libraries/extensions/ros2-bridge/src/transport/zenoh/qos.rs`, `ZenohQosMapping::from_ros_qos`.

The mapping builds the keyexpr with:

```rust
format!("{reliability}:{durability}:{history},{depth}:,:,:,{liveliness},")
```

For a manual-liveliness endpoint (`Liveliness::ManualByTopic` / `ManualByParticipant` → `"3"`) with otherwise-default QoS this yields `::,1:,:,:,3,`. Splitting on `:` gives 6 groups, and the trailing group is `,3,` — i.e. `<empty>,3,<empty>`.

Upstream `rmw_zenoh_cpp` encodes that final group as `<liveliness>,<lease_sec>,<lease_nsec>`, with the liveliness *kind* in the **first** subfield. So the code puts the liveliness kind (`3`) into the liveliness-lease-**seconds** slot and leaves the liveliness-kind slot empty. A real `rmw_zenoh` peer parsing this token reads `liveliness = <empty → automatic/default>` and `liveliness_lease_duration.sec = 3`, misreporting the endpoint's QoS and defeating any liveliness-based QoS-compatibility handling. Since native-mode wire parity with `rmw_zenoh_cpp` is the whole point of the feature, this is worth fixing before merge. The correct trailing group is `{liveliness},,` (kind first, empty lease sec/nsec).

It's also untested: `qos_default_and_non_default_match_rmw_components` only varies reliability and history depth — never durability or liveliness — so it passes with the bug present. A manual-liveliness case (expected trailing group `3,,`) would catch it. Separately, the QoS/attachment golden fixtures are generated by the same `encode()` / `from_ros_qos()` code under test rather than captured from a real `rmw_zenoh_cpp` peer, so despite names like `attachment_matches_upstream_serializer` they guard against regressions of this code's own output, not against actual upstream parity — a fixture captured from a pinned `rmw_zenoh_cpp` build would make the parity claim real.

---

phil-opp:
I found one issue not covered by the earlier automated reviews — it's in the Python bindings, which the previous reviews didn't audit.

**A single malformed `get_result` query from any peer aborts a legitimate result delivery** — `libraries/extensions/ros2-bridge/python/src/lib.rs`, `Ros2ActionServer::send_result` (Zenoh backend).

When the result is ready before the client has polled `get_result`, the code drains the `zenoh_result_requests` channel looking for the matching request:

```rust
let request = receiver.recv_timeout(remaining).ok()?;
let decoded: GetResultRequest = deserialize_cdr(&request.payload).ok()?;
let key = decoded.goal_id.uuid.to_string();
if decoded.goal_id == *id { return Some(request.id); }
self.pending_result_requests.insert(key, request.id);
```

`request.payload` is attacker-controlled, and `GetResultRequest` is a tiny fixed message. If any peer sends a non-CDR/truncated query to the `get_result` key while this loop is waiting, `deserialize_cdr(...).ok()?` returns `None` out of the whole `or_else` closure, so `send_result` fails with a misleading `"get-result request timed out"` error — even though a valid result was pending. Note the inconsistency: a well-formed request for a *different* goal is stashed in `pending_result_requests` and the loop continues, but a malformed one is dropped and aborts the call. Under sustained malformed traffic this denies result delivery to legitimate goals (it is retryable, since the goal isn't removed from `executing` on the error path, so it's a transient rather than permanent wedge).

Decoding should skip the bad request and continue the loop (like the sibling valid-but-non-matching branch already does) — e.g. on `Err`, `warn` + `continue` rather than `.ok()?`.

---
🤖 _Automated review by Claude — no human was in the loop for this comment. It may contain mistakes, so please verify the findings before acting on them._

---

bobdingAI:
Thanks,  so the summary for these findings is

1. Original DoS: 5 ?-propagation sites in main.rs
2. Zenoh action-server never evicts completed goals → stops accepting after 8 lifetime goals
3. server.reply(id,&response)? on request-expiry tears down service bridge
4. QoS liveliness in wrong keyexpr subfield (qos.rs:27) 
5.  Python send_result aborts on malformed get_result (.ok()? → misleading timeout)

phil-opp:

Re-reviewed the commits since the previous review rounds (`07a5a3c`, `9d348af`). They address the concerns raised earlier — verified on the current head:

- The five `?`-propagation decode/reply sites in `binaries/ros2-bridge-node/src/main.rs` now go through `peer_value_or_warn` (reject-and-`continue`), and the service `reply` path warns-and-continues on `RequestExpired`/`RequestAbsent` instead of `?`-propagating out of the loop.
- The standalone Zenoh action server now retires goals on result delivery and reclaims finished goals when at capacity, so it no longer wedges after `MAX_CONCURRENT_GOALS` lifetime goals.
- The QoS keyexpr now encodes the liveliness kind in the first subfield of the trailing group (`…:{liveliness},,`), matching `rmw_zenoh_cpp`, guarded by the manual-liveliness regression test.
- The Python `send_result` drain skips a malformed `get_result` query and continues rather than aborting a pending result.

No new issues found. One minor, non-blocking observation: `deserialize_cdr(&request.payload)?` remains in the Python `take_goal` / `cancel` Zenoh paths (`python/src/lib.rs` ~1896, ~2099), but those are single-shot user-invoked calls that raise an exception to the caller rather than tearing down a long-lived loop, so they aren't in the same remote-DoS class as the bridge-node sites.

---

phil-opp:
**Automated review by Claude** — fully automated, no human has vetted this. Verify before acting.

I reviewed the current head (`9d348af`) against the diff directly. The untrusted-input hardening from the earlier rounds is real: the decode sites now go through `peer_value_or_warn`, and the **reply** paths in the Zenoh service/action event loops were converted to warn-and-continue (`binaries/ros2-bridge-node/src/main.rs:395`, `760`, `798`, `862`), each with a comment noting a `RequestExpired`/`RequestAbsent` must not tear down the bridge.

I found the following issue, which I don't think any prior round raised.

**Remote DoS: the `reject(...)?` calls in the Zenoh service- and action-server loops still `?`-propagate and kill the bridge on malformed peer input.**

The **reject** siblings of those hardened reply calls were left as `block_on(server...reject(...))?`:
- `main.rs:376` — service `reject(request.id, "invalid request payload")`
- `main.rs:730` — action `send_goal.reject(request.id, ...)`
- `main.rs:791` — action `get_result.reject(request.id, "invalid request payload")`
- `main.rs:815` — action `get_result.reject(request.id, "unknown goal")`
- `main.rs:835` — action `cancel_goal.reject(request.id, "invalid request payload")`

`RawServiceServer::reject` (`libraries/extensions/ros2-bridge/src/transport/zenoh/service.rs`) is fallible in two ways:
```rust
let query = self.pending.lock()...take(id, elapsed(self.origin))?;   // RequestAbsent / RequestExpired
query.reply_err(ZBytes::from(message.to_owned())).await.map_err(...) // Err on a finalized query
```
and `PendingRequests::insert` **overwrites** an existing entry with the same `RequestId`, while `take` returns `RequestAbsent` once the id has been removed.

The `RequestId` comes straight from the attacker-controlled attachment (`sequence_number` + `client_gid`, a trivially forgeable `ZSerialized i64,i64,[u8;16]`). So the following is a deterministic, unauthenticated availability attack:

1. Peer sends **two** requests carrying the **same attachment** (identical seq+gid) and a ROS-level-malformed payload.
2. The queryable callback inserts `id → q1`, sends request 1; then inserts `id → q2` (overwriting q1's entry), sends request 2.
3. Event loop handles request 1: payload fails to decode → `reject` → `take(id)` returns `q2` and removes the entry → `reply_err` ok.
4. Event loop handles request 2: `reject` → `take(id)` → `RequestAbsent` → `Err` → `?` → the `for event in block_on_stream(merged)` loop returns `Err` → `run_zenoh_service_mode` / `run_zenoh_action_server` returns `Err` → the bridge process exits.

(The `RequestExpired` path — a request that ages past the pending expiry before the loop reaches it — and the `reply_err`-on-finalized-query path are additional, less deterministic triggers of the same `?`.)

This is the same remote-DoS class the earlier reviews were about; the fix just didn't cover the reject branch. Suggest giving these five `reject` calls the same treatment as their reply siblings — `if let Err(e) = block_on(...reject(...)) { warn!(...); }` then `continue` — so a malformed or duplicated peer request can't tear down the bridge.

Secondary (already noted in the 2026-07-22 review, still true after the latest commits): the `attachment_matches_upstream_serializer` / QoS-component tests and the `tests/fixtures/rmw_zenoh/*` goldens are generated by the same `Attachment::encode` / `from_ros_qos` code under test, so they guard against regressions of this code's own output but don't prove wire parity with a real `rmw_zenoh_cpp` peer (e.g. the raw-16-byte gid with no length prefix, and default-QoS-as-empty-string encoding remain unconfirmed against upstream). The "native parity" claim rests on the interop harness, not these unit fixtures.

---

phil-opp:

Took a pass over the latest hardening commit (`2b87e37`) — the five `reject(...)` sites in the service/action loops now warn-and-continue like their reply siblings, so the earlier duplicate-attachment tear-down path looks closed, and the added map bounds (graph cache, service `pending`, `pending_result_requests`, `ActionState::accept` reclaim) are a clear improvement over the prior unbounded growth. A few low-severity residuals in the new bounding code, none blocking:

- **Graph cache cap is refuse-new, not LRU** (`transport/zenoh/graph.rs`): once `MAX_GRAPH_ENTITIES` (16384) is hit, every new liveliness token is dropped — including the local node's own observed entities and real ROS peers. An unauthenticated peer that announces 16384 distinct tokens and keeps them alive permanently stalls local graph discovery until those tokens are undeclared. That swaps the old memory-exhaustion DoS for a bounded-memory discovery one; an LRU / own-entity-priority policy (or at least a comment) would help.
- **`pending_result_requests` overflow evicts an arbitrary entry** (`python/src/lib.rs`): `keys().next()` on a HashMap isn't the oldest, so a flood of non-matching get-result queries can evict a legitimate in-flight request. Retryable, but eviction-oldest would be more predictable.
- **`ActionState::accept` reclaim can drop an unfetched result** (`transport/action.rs`): at `goals.len() >= 2*limit` it retains only non-terminal goals, so a slow-but-legitimate client that finished a goal but hasn't polled `get_result` yet gets "unknown goal" under load.

Separately (raised in an earlier round, just noting it's still open): the `attachment_matches_upstream_serializer` / QoS goldens are produced by the same `encode()` / `from_ros_qos()` code they test, so they lock in this code's own output rather than proving wire parity against a captured `rmw_zenoh_cpp` build — the parity guarantee really rests on the interop harness.

---

bobdingAI:
Thanks for the thorough automated reviews across the rounds — **all findings are now addressed** on the current head (`e66abda8`), rebased onto latest `main`. Mapping each finding to its fix:

### Remote-DoS / untrusted-input hardening (`binaries/ros2-bridge-node`, Zenoh transport)
- The five `?`-propagating decode/reply sites that let a single malformed peer request tear down the bridge → reject/warn-and-continue (`32bd1588`, `07a5a3c4`).
- The sibling `reject(...)?` calls (duplicate forged attachment → `RequestAbsent`/`RequestExpired` → crash) → warn-and-continue (`2b87e378`).
- Standalone Zenoh action server now retires goals on result delivery and reclaims completed goals at capacity — no longer wedges after `MAX_CONCURRENT_GOALS` lifetime goals (`07a5a3c4`).
- Python Zenoh `send_result` skips a malformed `get_result` query instead of aborting a pending legitimate result (`07a5a3c4`).

### QoS liveliness keyexpr
- The liveliness kind is now encoded in the first subfield of the trailing group (`{kind},,`) to match `rmw_zenoh_cpp`, with a manual-liveliness regression test (`07a5a3c4`).

### Wire-parity fixtures (goldens were generated by the code under test)
- Added fixtures **captured from a live ROS 2 Humble `rmw_zenoh_cpp` 0.1.8 peer** (sniffed with an independent zenoh client), plus tests that decode+re-encode the real attachment bytes byte-identically, parse a real publisher liveliness token, and confirm `from_ros_qos` reproduces the peer's QoS component byte-for-byte (`7d58bcc6`). Capture procedure documented in `tests/fixtures/rmw_zenoh/README.md`.

### Additional issues found in a deeper security pass (also fixed)
- `GraphCache` liveliness-entity map was unbounded (peer-announced tokens → memory exhaustion) → capped (`2b87e378`).
- Standalone service-mode `pending` map was unbounded → TTL-evicted + capped like the DDS path (`2b87e378`).
- `ActionState` retained terminal goals → reclaimed (`2b87e378`).
- Low-severity: service-client skips a malformed reply instead of cancelling `call`; keyexpr rejects over-long tokens before allocating; get-result log throttled; pubsub cap scope documented (`70481a82`).

### Validation
- CI green: fmt, clippy `-D warnings`, workspace `Test`, E2E, semantic-contract, and benchmark jobs.
- Humble + Kilted `rmw_zenoh_cpp` interop matrices: 10/10 each.
- Two graph-snapshot tests that raced on async liveliness-token propagation were deflaked with an event-driven wait (`9d348afa`, `e66abda8`).

Decoders (`Attachment::decode`, `LivelinessKey::parse`) were also reviewed and confirmed bounds-checked / fail-closed. Happy to drill into any specific item.

bobdingAI:
Thanks for the close read — addressed all three low-severity residuals in `daf0446b`:

- **Graph cache** now evicts the oldest *remote* entity to admit a new token (was refuse-new), and locally-declared entities are tracked via `apply_put_local` and never evicted — so a remote token flood can neither stall discovery of new legitimate entities nor evict the node's own state.
- **`pending_result_requests`** now evicts by insertion sequence (oldest), not `keys().next()`, so a flood of non-matching get-result queries can't drop a recently-stashed legitimate request.
- **`ActionState::accept`** now evicts the oldest *terminal* goals to stay under `2*limit`, instead of purging all terminal goals — a recently finished-but-not-yet-fetched result survives under load.

Regression tests added for each: local-entity protection + remote LRU eviction (`graph.rs`) and oldest-terminal-first reclaim (`action.rs`).

On the goldens/parity note (you were reviewing `2b87e37`): that's since been handled on the current head by `7d58bcc6` — I added fixtures **captured from a live ROS 2 Humble `rmw_zenoh_cpp` 0.1.8 peer** (sniffed with an independent zenoh client) plus tests that round-trip the real attachment bytes, parse a real publisher liveliness token, and confirm `from_ros_qos` matches the peer's QoS component byte-for-byte. So parity is now pinned to captured upstream data rather than resting on the interop harness alone; the earlier self-generated `attachment_matches_upstream_serializer` check stays alongside as a regression guard. Capture procedure is documented in `tests/fixtures/rmw_zenoh/README.md`.

phil-opp:
The two commits since the last review (`e66abda8`, `daf0446b`) look safe and address the three eviction-policy residuals raised in the previous round:

- **Graph cache** (`transport/zenoh/graph.rs`): a new remote token is now admitted by evicting the oldest *remote* entity (`min_by_key(seq)` over non-local entries) rather than refusing new entries, and locally-declared entities are tracked via the new `apply_put_local` (wired in `mod.rs` for both node and endpoint declaration), so a remote flood can neither stall discovery nor evict the node's own state.
- **`pending_result_requests`** (`python/src/lib.rs`): the stash is now keyed with a monotonic `seq` and overflow evicts the oldest entry, instead of an arbitrary `keys().next()` HashMap entry.
- **`ActionState::accept`** (`transport/action.rs`): terminal goals are evicted oldest-first down to `2*limit` rather than purged all at once, so a recently finished-but-unfetched result survives under load. The arithmetic holds — active goals are capped at `limit`, so there are always enough terminal goals to bring the map below `2*limit`.

Each change has a matching regression test (`apply_put_never_evicts_local_entities`, `apply_put_caps_entities_and_evicts_oldest_remote`, the oldest-seq eviction test, `accept_evicts_oldest_terminal_first_keeping_recent_results`). `e66abda8` is a test-timing deflake with no production impact. No new issues found in these commits.

---
🤖 This is a fully automated review by Claude — no human was in the loop. It may be wrong; please use your own judgement.

---

PR #2791, opened 2026-07-21
fix(examples): bump rust-dataflow-git pin past the node-register format break (#2742)

phil-opp:
The `Examples` job fails on all three platforms — the git-pinned `rust-dataflow-git` nodes can't register against a current daemon: `failed to deserialize DaemonRequest: unexpected end of file`.

Cause: #2774 added a `metadata_version` field to `NodeRegisterRequest` (to reject metadata-format skew at registration) — a change to the register wire format. The example's pin was still `07ed3be` (set by #2745, predates #2774), so the pinned nodes send a register request without the new field and the daemon fails to deserialize it. #2775's backstop guard then makes the run fail loudly on every platform instead of silently.

Bumps the pin to `c9a3699` (current `main`, format-current) across all three nodes. Same operation #2745 performed for #2366's earlier payload-format break — the example working as designed (bump `rev:` when a message-format break lands on main).

Refs #2742.

### DIFF
--- a/examples/rust-dataflow-git/dataflow.yml
+++ b/examples/rust-dataflow-git/dataflow.yml
@@ -4,14 +4,15 @@
 # change lands on main — otherwise the CI job catches the mismatch and
 # signals a compatibility break, which is the whole point of this test.
 #
-# Currently pinned past #2366, which dropped the `ArrowTypeInfo` sidecar
-# from `Metadata` and took the payload wire format to v1. The previous pin
-# (1.0.0-rc1) predated that, so its nodes desynced against a current daemon
-# and died with `tag for enum is not valid` (#2742).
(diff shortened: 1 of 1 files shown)


phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Please verify before acting on it.

No issues found. This bumps all three `rust-dataflow-git` node pins (`rust-node`, `rust-status-node`, `rust-sink`) from `07ed3be` to `c9a3699` so the git-built example nodes speak the current `NodeRegisterRequest` format (post-#2774 `metadata_version`) and register cleanly against a current daemon. All three `rev:` entries are updated consistently and the explanatory comment matches the change.

---

PR #2792, opened 2026-07-21
fix(examples): make webcam stop within grace on a camera-less CI runner (#2742)

phil-opp:
The `CLI Tests (windows)` nightly runs `python-operator-dataflow`, whose `webcam` operator is force-killed at the daemon's grace period on headless runners (`ExitCode(1)` → red). With no camera, `cv2.VideoCapture.read()` on the Windows MSMF/obsensor backend can block for seconds per call, so the operator never observes STOP within grace — and never even reaches node teardown (which is why the earlier teardown-timeout change in #2778 couldn't fix it).

Detect the missing camera up front via `isOpened()` and skip the blocking `read()` entirely: emit a placeholder frame each tick (keeping the operator responsive) and, under CI, self-stop after a short window well inside the daemon's `--stop-after` (20s), so the node exits cleanly before any force-kill. Also honors a STOP event immediately, and logs a slow read plus the `camera_available` result so the next run's log *shows* where the time goes instead of us reasoning from constants.

Control flow verified with a mock-based test across the headless / within-window / STOP / camera-frame / transient-failure / opened-but-broken branches.

Refs #2742.

### DIFF
--- a/examples/python-operator-dataflow/webcam.py
+++ b/examples/python-operator-dataflow/webcam.py
@@ -13,6 +13,15 @@ CAMERA_HEIGHT = 480
 CAMERA_INDEX = int(os.getenv("CAMERA_INDEX", 0))
 CI = os.environ.get("CI")
 
+# When no real camera is available (e.g. a headless CI runner), run only this
+# long before stopping. `read()` is skipped in that case so the operator stays
+# responsive, and the window keeps us well inside the daemon's `--stop-after`
+# (20s) so the node self-exits cleanly instead of being force-killed at the
+# grace period (ExitCode(1) on Windows, reddening the nightly — #2742). Long
+# enough for the downstream YOLO node to warm up and run inference on a few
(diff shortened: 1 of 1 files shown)


phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop. Please verify before acting on it.

No issues found. Verified against the full file: `import time` and `self.start_time = time.time()` are both present, so the new elapsed-time checks won't raise. Gating the blocking `read()` behind `self.camera_available` (from `isOpened()`) is the right fix for the headless-runner force-kill — it avoids the multi-second `read()` on a device-less backend that kept the operator from observing `STOP` within the grace period. The headless / repeated-failure paths all converge correctly on the placeholder frame, and the CI headless self-stop at `CI_HEADLESS_RUN_SECS` (10s) sits well inside the daemon's 20s `--stop-after`. Logic is sound.

---

PR #2796, opened 2026-07-21
fix(daemon): emit a terminal result when a node restart is abandoned

phil-opp:
## Summary

`PreparedNode::restart_loop` (`binaries/daemon/src/spawn/prepared.rs`) can leave a node **wedged in `running_nodes` forever**, hanging the whole dataflow, when a restart is announced but then abandoned.

The loop reports a `SpawnedNodeResult { restart: true }` **before** it attempts to respawn the node. On that event the daemon deliberately keeps the node in `running_nodes` (it only notifies downstream nodes of `NodeRestarted`) and waits for a follow-up result — node-stop bookkeeping (`handle_node_stop`) and dataflow finish-detection (`finish_dataflow`) run **only** on a `restart: false` result (`lib.rs`, the `if restart { … } else { … }` split in the `SpawnedNodeResult` handler).

`restart_loop` is the **sole** emitter of `SpawnedNodeResult` for every node, but three of its exit paths `break` out *after* announcing `restart: true` without ever sending a follow-up:

1. **Respawn fails** — `spawn_inner` returns `Err` (e.g. the node binary is transiently unavailable, the OS refuses the fork, or the per-incarnation log file can't be created).
2. **Unexpected dynamic result** — `spawn_inner` returns `Ok(NodeKind::Dynamic)` ("cannot restart dynamic node").
3. **Restart cancelled during backoff** — the node's inputs close (upstream finishes) during the `restart_delay` backoff wait, so `disable_restart` flips and the loop breaks.

## Impact

In all three cases the node stays in `running_nodes` with a dead process handle:

- `finish_dataflow` never runs, so the coordinator never receives `AllNodesFinished`; the dataflow shows `Running` indefinitely (`dora list`) and `dora run` / attach never returns.
- The node's `subscribe_channels`, pending messages, zenoh publishers, and memory pools leak until the daemon is killed.
- The finish-straggler watchdog cannot rescue it: the `restart: true` event already cleared `connected_nodes`, so the node hits the `connected == false` arm and **vetoes** escalation for the entire dataflow.

The most reachable trigger is (1): a node with `restart_policy: on-failure` (or `always`) that crashes once, where a transient condition prevents the respawn.

## Fix

Emit a terminal `SpawnedNodeResult { restart: false }` — preserving the original exit status — on each abandon path, via a small shared helper `emit_restart_abandoned`. This lets the daemon run its normal node-stop bookkeeping so the node is removed and the dataflow can finish. Because `restart_loop` is the only emitter for the node, there is no competing cleanup path and thus no double-handling.

## Validation

- New unit test `abandoned_restart_emits_terminal_result` drives the abandon path (`command: None` → `Ok(NodeKind::Dynamic)`, `restart_policy: always`, failure exit) and asserts the loop emits exactly `[restart: true, restart: false]`. Without the fix only `restart: true` is emitted and the assertion fails.
- `cargo test -p dora-daemon abandoned_restart_emits_terminal_result` — passes.
- `cargo clippy -p dora-daemon -- -D warnings` (CI's invocation) — clean.
- Full workspace test suite run as a final check.

## Note

🤖 This change is **machine-generated by Claude** (an AI agent) and was reviewed automatically. A human maintainer should review it before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EbTHdwv1463UPyn8YrdBhH)_

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -347,7 +347,10 @@ impl PreparedNode {
             let event = DoraEvent::SpawnedNodeResult {
                 dataflow_id: self.dataflow_id,
                 node_id: self.node.id.clone(),
-                exit_status,
+                // Cloned (not moved) so the same status can be re-reported as a
+                // terminal `restart: false` result if the restart is later
+                // abandoned (see `emit_restart_abandoned`).
+                exit_status: exit_status.clone(),
                 dynamic_node: self.node.kind.dynamic(),
                 restart,
                 restart_count,
@@ -398,6 +401,12 @@ impl PreparedNode {
                             Some("daemon".into()),
                             "restart cancelled: inputs closed before respawn".to_string(),
                         )
+                        .await;
+                    // A `restart: true` result was already reported above, so the
+                    // daemon kept this node in `running_nodes` expecting a
+                    // follow-up. Emit a terminal result so it is cleaned up and
+                    // the dataflow can finish.
+                    self.emit_restart_abandoned(exit_status, restart_count)
                         .await;
                     break;
                 }
@@ -480,6 +489,8 @@ impl PreparedNode {
                                 "cannot restart dynamic node".to_string(),
                             )
                             .await;
+                        self.emit_restart_abandoned(exit_status, restart_count)
+                            .await;
                         break;
                     }
                     Err(err) => {
@@ -490,6 +501,8 @@ impl PreparedNode {
                                 format!("failed to restart node: {err:?}"),
                             )
                             .await;
+                        self.emit_restart_abandoned(exit_status, restart_count)
+                            .await;
                         break;
                     }
                 }
@@ -497,6 +510,41 @@ impl PreparedNode {
                 break;
             }
         }
+    }
+
+    /// Report a terminal `SpawnedNodeResult` (with `restart: false`) after a
+    /// restart was announced but then abandoned.
+    ///
+    /// `restart_loop` emits a `restart: true` result *before* attempting a
+    /// respawn. On that event the daemon deliberately keeps the node in
+    /// `running_nodes` (only notifying downstream of `NodeRestarted`) and waits
+    /// for a follow-up result — node-stop bookkeeping and dataflow
+    /// finish-detection run only on a `restart: false` result. `restart_loop`
+    /// is the sole emitter of `SpawnedNodeResult`, so if it breaks out after a
+    /// failed/abandoned respawn (respawn error, an unexpected dynamic-node
+    /// result, or the restart being cancelled during the backoff wait) without
+    /// this call, the node stays wedged in `running_nodes` forever — its
+    /// channels, memory pools and zenoh publishers leak and the dataflow never
+    /// finishes (`dora list` shows it `Running`, `dora run`/attach never
+    /// returns). Emitting a terminal result closes that gap.
+    async fn emit_restart_abandoned(&self, exit_status: NodeExitStatus, restart_count: u32) {
+        let event = DoraEvent::SpawnedNodeResult {
+            dataflow_id: self.dataflow_id,
+            node_id: self.node.id.clone(),
+            exit_status,
+            dynamic_node: self.node.kind.dynamic(),
+            restart: false,
+            restart_count,
+        }
+        .into();
+        let _ = self
+            .daemon_tx
+            .clone()
+            .send(Timestamped {
+                inner: event,
+                timestamp: self.clock.clone().new_timestamp(),
+            })
+            .await;
     }
 
     async fn spawn_inner(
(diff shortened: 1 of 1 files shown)


phil-opp:
## 🤖 Automated review run summary

This PR is the output of an automated, machine-generated codebase-review pass by Claude (an AI agent). This comment summarizes the full run for traceability. **All findings were verified against `origin/main` @ `55eab7a` and de-duplicated against the ~74 open PRs from prior automated runs.** A human maintainer should review before merging.

### PRs opened this run

**Correctness**
- **#2796** (this PR) — `fix(daemon): emit a terminal result when a node restart is abandoned`. `restart_loop` announces `SpawnedNodeResult { restart: true }` before respawning, but three exit paths (respawn error, unexpected dynamic result, restart cancelled during backoff) break without a follow-up terminal result — leaving the node wedged in `running_nodes` forever, hanging the dataflow and leaking its channels/pools/publishers. Since `restart_loop` is the sole emitter of `SpawnedNodeResult`, nothing else recovers it.

_Optimization / Documentation / Simplification: no PR-worthy findings this run._

### Rejected findings (verified, not acted on)

| Finding | Area | Reason rejected |
|---|---|---|
| SHM length guard missing in `zenoh_publish` vs `allocate_data_sample` | `apis/rust/node/src/node/mod.rs` | Latent panic only under a hypothetical zenoh SHM provider that over-allocates the logical length; zenoh 1.8 guarantees `alloc(n)` length == `n`, so it cannot trigger on current input. |
| Undecodable inbound frame is fatal to the reader thread | `binaries/mavlink2-bridge-node` | Theoretical only — `to_record_batch` builds from generated, schema-matched columns and cannot fail for any real inbound frame. |
| `dora up` daemon-timeout message off-by-one | `binaries/cli/src/command/up.rs` | Cosmetic: reports `2.1s` when `2.0s` actually elapsed. Wrong number in a rarely-seen error string, no functional impact. |
| ROS2 type format requires exactly 2 `/`-parts | `libraries/core` validate | Intended — dora's convention is uniformly 2-part (`sensor_msgs/Image`); verified against `examples/ros2-bridge/**`. |
| `ByteSize::from_str` uses `is_alphabetic` vs `is_ascii_alphabetic` elsewhere | `libraries/message` / `libraries/core` | Cosmetic inconsistency; both reject malformed input without panic. |
| Missing multi-runtime informational warning for `send_logs_as`/`min_log_level` | `libraries/core` validate | A missing warning, not incorrect behavior. |
| `env_default` int-overflow reported as type mismatch | `libraries/core` validate | Rejection is correct (value doesn't fit `int`); only the message is slightly imprecise. |

### Validation for #2796
- `cargo test -p dora-daemon` — **109 passed, 0 failed** (incl. new `abandoned_restart_emits_terminal_result`).
- `cargo clippy -p dora-daemon -- -D warnings` (CI's invocation) — clean.
- Full-workspace `cargo test --all` could not run to completion in this environment due to a hard disk-capacity ceiling (compiling every example bin crate's test harness exhausts the session disk allowance — `No space left on device`); the change is isolated to a single private method + test in `dora-daemon`, whose complete suite passes.

---

phil-opp:
🤖 **Automated review by Claude** — fully automated review, no human oversight; verify before acting.

No issues found.

I reviewed the full diff (`binaries/daemon/src/spawn/prepared.rs`). The fix is sound:

- `restart_loop` announces `SpawnedNodeResult { restart: true }` before respawning and then, on the three abandon paths (respawn `Err`, unexpected dynamic-node result, restart cancelled during backoff), broke out without a follow-up terminal result — leaving the node wedged in `running_nodes` and the dataflow unable to finish. `emit_restart_abandoned` now sends a terminal `restart: false` on each of those paths, which is what triggers the daemon's node-stop bookkeeping and finish-detection.
- The announce site is correctly switched to `exit_status.clone()` so the original status remains available for the later abandon call. Each abandon path `break`s immediately, so at most one terminal result is emitted per loop termination — no double-emission, and no use-after-move across iterations.
- Reporting the original triggering `exit_status` (rather than the respawn error, which is already surfaced via `report_error`) is fine, since the terminal result's role here is cleanup; the `restart_count` is preserved.
- The unit test drives the `command: None → Ok(NodeKind::Dynamic)` abandon path and asserts the emitted flags are `[true, false]`, which pins the regression.

---

phil-opp:
I checked this against current `main`. The terminal-result approach and the three abandonment paths covered here look correct, and the test pins the regression well.

One concern before merge: this branch is stale against `main` and collides with #2606 (`fix(daemon): re-check disable_restart before respawn on every path`, commit `bc64ab3`), which merged after this PR opened. #2606 moved the `disable_restart` re-check **out** of the `if let Some(restart_delay)` block so it now runs unconditionally before every respawn (`binaries/daemon/src/spawn/prepared.rs:394`, log `"restart cancelled: inputs closed before respawn"`). On `main` that relocated cancel break has **no** terminal-result emit — in fact `emit_restart_abandoned` doesn't exist on `main` at all — and it is now reachable for the default `restart_delay: None` config, i.e. exactly the wedge this PR is meant to eliminate.

Because this PR attaches `emit_restart_abandoned` to the *old* in-backoff-block check that #2606 deleted, rebasing will conflict in this hunk, and a naive resolution can silently drop the emit on the cancel path (leaving the cancel-during-restart abandonment uncovered on real `main`). Please rebase onto latest `main` and make sure `emit_restart_abandoned` fires on the **unconditional** pre-respawn cancel break, not just the in-backoff one.

Minor: the doc comment enumerates "three exit paths," but there's a fourth — after a successful respawn (which already announced `restart: true`), a top-of-loop `finished_rx.await` resolving to `Err` breaks with the last announced state still `restart: true`, leaving the node wedged. Likely unreachable in practice, but worth a note or guard given the stated goal of covering all abandonment paths.

---
🤖 This is a fully automated review by Claude — no human has reviewed or verified it. Treat it as a suggestion to double-check, not authority.

---

phil-opp:
**Automated review — re-checked after the latest force-push.**

The rebase onto current `main` resolves the concern from the earlier automated review (2026-07-23) about the #2606 collision: `emit_restart_abandoned` is now attached to the **unconditional** pre-respawn `disable_restart` check (`binaries/daemon/src/node_communication/prepared.rs:408`, outside the `restart_delay` block), matching the #2606 restructure, and the diff now applies cleanly (`mergeable_state: clean`).

The fix looks correct: exactly one terminal `SpawnedNodeResult { restart: false }` is emitted on each of the three announced-then-abandoned paths (cancel-during-backoff, unexpected dynamic result, respawn error), each immediately followed by `break`, so there's no double-emit, and the original `exit_status`/`restart_count` are preserved for downstream node-stop bookkeeping. The added unit test genuinely drives the abandon path and asserts the emitted flags are exactly `[true, false]` — it fails without the fix. No new issues.

Minor, pre-existing (out of scope, not introduced here): the loop-top `finished_rx.await → Err` break (`prepared.rs:269`) also exits without a terminal result — including after a successful respawn that already announced `restart: true`. It's likely unreachable under normal process monitoring, but if the intent is to cover *all* abandonment paths, it's worth either a guard or a comment marking it intentionally out of scope.

---
🤖 _This is a fully automated review by Claude ([Claude Code](https://claude.ai/code)). No human has vetted these findings — please confirm before acting on them._

---

phil-opp:
**Automated review — re-check after `main` advanced** (this is a fully automated review by Claude; no human has vetted it)

No change to this diff since my previous review, and the fix itself still looks correct. One new issue, though, from `main` moving underneath the branch:

**#2893 (`eec31a4`, merged 2026-07-30) added a required `pid: u32` field to `DoraEvent::SpawnedNodeResult`** (`binaries/daemon/src/event_types.rs:182`), and the existing announce site now sets `pid: exited_pid` (`prepared.rs:358`). The new `emit_restart_abandoned` helper on this branch builds `SpawnedNodeResult { … restart_count }` **without `pid`**, so it will **fail to compile once rebased onto current `main`** (missing field `pid`). GitHub still reports the merge as clean because that check is textual, not a compile.

`exited_pid` is already in scope in `restart_loop` (bound at `prepared.rs:272`), so the fix is to thread it into `emit_restart_abandoned` and set `pid: exited_pid` on the terminal result, matching the announce site.

---

phil-opp:
Superseded: main already emits a terminal `SpawnedNodeResult` at all three abandoned-restart sites via `send_terminal_exit`, which also carries `generation` and `pid`. Rebasing onto main leaves zero commits.

PR #2799, opened 2026-07-22
fix(ros2-bridge): saturate seconds decrement in builtin_interfaces__Time conversion

### LINKED ISSUE #2798: ROS2 bridge: integer overflow in `builtin_interfaces__Time` conversion (`quot_sat - 1` when saturated to `i32::MIN`)
> ⚙️ **This issue was created by an automated, scheduled Claude code-review check.** It was verified against the source but has **not** been triaged by a human. Please double-check before acting on it.

## Summary

The generated `From<ros2_client::builtin_interfaces::Time> for ffi::builtin_interfaces__Time` conversion has an integer-overflow bug in its negative-time branch. When the whole-seconds value saturates to `i32::MIN`, the code then unconditionally computes `i32::MIN - 1`, which **panics in debug builds** (`attempt to subtract with overflow`) and **silently wraps to `i32::MAX` in release builds** — turning a large *negative* timestamp into a large *positive* one (sign flip).

## Location

`libraries/extensions/ros2-bridge/msg-gen/src/types/message.rs:218` (inside the `"builtin_interfaces__Time"` arm of a `quote! { ... }` macro, so the bug is emitted into every generated ROS2 bridge build).

```rust
// t = time.to_nanos();  // i64 nanoseconds
// quot = t / 1_000_000_000;  rem = t % 1_000_000_000;
// ... negative-time, non-zero-remainder branch (rem < 0):

// saturate whole seconds
let quot_sat = if quot >= (i32::MIN as i64) {
    quot as i32
} else {
    tracing::warn!("rcl_interfaces::Time conversion underflow");
    i32::MIN
};

Self {
    sec: quot_sat - 1, // note -1   <-- overflows when quot_sat == i32::MIN
    nanosec: (1_000_000_000 + rem) as u32,
}
```

`sec` is an `i32` (matching the `i32::MAX`/`i32::MIN` saturation used in the sibling `rem >= 0` branch a few lines above).

## The bug

`quot_sat` is carefully saturated so it never goes *below* `i32::MIN`. But the very next line decrements it unconditionally: `quot_sat - 1`. When `quot_sat == i32::MIN` (which happens whenever `quot <= i32::MIN`), this evaluates `i32::MIN - 1`:

- **Debug build:** panics with `attempt to subtract with overflow`.
- **Release build:** wraps to `i32::MAX`, corrupting the timestamp — a ~`-2.15e9`-second time is reported as ~`+2.15e9` seconds.

The saturation branch's own existence (`else { warn!("...underflow"); i32::MIN }`) shows the authors expect `quot` to reach and exceed the `i32`-second range; the `- 1` then breaks the saturation guarantee they intended to provide.

## Trigger

Any `Time` whose `to_nanos()` value `t` satisfies both:
1. `t / 1_000_000_000 <= i32::MIN` (i.e. roughly ≥ 68 years before the epoch — well within `i64` range), **and**
2. `t % 1_000_000_000 != 0` (non-zero negative sub-second remainder, so the negative branch is taken).

Concrete example: `t = i32::MIN as i64 * 1_000_000_000 - 1 = -2_147_483_648_000_000_001`
→ `quot = -2_147_483_648` (== `i32::MIN`), `rem = -1`
→ `quot_sat = i32::MIN`
→ `sec = i32::MIN - 1` → panic (debug) / `i32::MAX` (release).

## Suggested fix

Use saturating subtraction so the decrement respects the same saturation the branch already intends:

```rust
sec: quot_sat.saturating_sub(1),
```

(or only decrement when `quot_sat > i32::MIN`). This keeps `sec` clamped at `i32::MIN` for these extreme timestamps instead of overflowing.

## Severity

**Low.** It requires an extreme (but valid `i64`) timestamp and only affects the ROS2 bridge's outbound `Time` → FFI conversion. Impact is a debug-build panic or a silently sign-flipped seconds value in release. A regression test covering `t == i32::MIN as i64 * 1_000_000_000 - 1` (and a value strictly below it) would lock in the intended saturating behavior.

phil-opp:
## Summary

Fixes #2798.

The generated `From<Time> for ffi::builtin_interfaces__Time` conversion in `libraries/extensions/ros2-bridge/msg-gen/src/types/message.rs` carefully saturates the whole-seconds value to `i32::MIN`, but the negative sub-second branch then decremented it unconditionally:

```rust
sec: quot_sat - 1, // note -1
```

When `quot_sat == i32::MIN` (whenever `quot <= i32::MIN`), this evaluates `i32::MIN - 1`:

- **Debug builds:** panic with `attempt to subtract with overflow`.
- **Release builds:** wrap to `i32::MAX`, corrupting the timestamp — a `~-2.15e9`-second time is reported as `~+2.15e9` seconds (sign flip).

This bug is emitted into every generated ROS2 bridge build.

## Fix

Use `saturating_sub(1)` so the value stays pinned at `i32::MIN`, preserving the saturation guarantee the surrounding branches already establish. One-token behavioral change plus an explanatory comment.

## Trigger

Any `Time` whose `to_nanos()` value `t` satisfies both `t / 1_000_000_000 <= i32::MIN` (≈ 68 years before the epoch, well within `i64` range) and a non-zero negative sub-second remainder.

## Validation

- `cargo check -p dora-ros2-bridge-msg-gen` ✅
- `cargo clippy -p dora-ros2-bridge-msg-gen -- -D warnings` ✅

The changed code is inside a `quote! { ... }` codegen macro (it is emitted into the generated bridge crate rather than compiled here), so there is no in-crate unit test to add; the fix is a localized saturation change verified by inspection against the sibling `rem >= 0` saturation branch.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01V54sCK4Gp5JR4dwv7KvNRE)_

### DIFF
--- a/libraries/extensions/ros2-bridge/msg-gen/src/types/message.rs
+++ b/libraries/extensions/ros2-bridge/msg-gen/src/types/message.rs
@@ -215,7 +215,16 @@ impl Message {
                             // * -1.0           sec = -1 sec and 0 nanosec
                             // * -1.00000000001 sec = -2 sec and 999_999_999 nanosec
                             Self {
-                                sec: quot_sat - 1, // note -1
+                                // Borrow one second for the negative fractional
+                                // part. Saturate the decrement: when `quot_sat`
+                                // is already `i32::MIN` (either `quot == i32::MIN`
+                                // or the underflow branch above clamped it), a
+                                // plain `- 1` computes `i32::MIN - 1`, which
+                                // panics in debug builds and wraps to `i32::MAX`
+                                // (a sign flip) in release builds. `saturating_sub`
+                                // keeps the value pinned at `i32::MIN`, preserving
+                                // the saturation guarantee.
+                                sec: quot_sat.saturating_sub(1),
                                 nanosec: (1_000_000_000 + rem) as u32,
                             }
                         }


phil-opp:

Reviewed the codegen change in `msg-gen` (`From<Time> for ffi::builtin_interfaces__Time`). The `quot_sat.saturating_sub(1)` correctly clamps the borrowed-second value at `i32::MIN` for the `quot <= i32::MIN` case, where the old `quot_sat - 1` would overflow (debug panic / release wraparound to `i32::MAX`, a sign flip). For every other input it is bit-identical to `- 1`, and the sibling `nanosec = (1_000_000_000 + rem) as u32` stays in range for `rem ∈ [-999_999_999, -1]`, so there's no borrow/overflow interaction with the changed line.

No issues found.

---

PR #2800, opened 2026-07-22
fix(cli): reset follow offset on log rotation in `dora logs --local --follow`

### LINKED ISSUE #2785: `dora logs --local --follow` drops and corrupts output when the log file rotates

## Summary

When following local logs with `dora logs --local --follow`, log **rotation** during the follow session causes new log lines to be **silently dropped** and, once the fresh file grows past the stale offset, **misaligned/garbled** output. The follow loop tracks a byte offset against a fixed path string, but rotation replaces the file behind that path with a fresh one that starts at offset 0.

This is a pre-existing bug in the follow loop; it was **not** introduced by the recent byte-offset fix (`e792309`), whose inner `read_appended_log_lines` helper is correct. The defect is in how `follow_local_logs` manages offsets across a file-replacement boundary.

## Location

`binaries/cli/src/command/logs.rs` — `follow_local_logs` (offsets captured once at lines ~298-303; poll loop at ~306-319).

```rust
// Track file byte offsets (start after existing content)
let mut file_positions: HashMap<PathBuf, u64> = HashMap::new();
for path in &files {
    let len = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
    file_positions.insert(path.clone(), len);
}

loop {
    std::thread::sleep(Duration::from_millis(200));
    for path in &files {
        let pos = file_positions.get(path).copied().unwrap_or(0);
        let current_size = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
        if current_size <= pos {
            continue;                      // <-- new (smaller) file: all lines skipped
        }
        let (msgs, new_pos) = read_appended_log_lines(path, pos)?; // <-- seeks to stale offset in NEW file
        ...
    }
}
```

The daemon rotates node logs when `max_log_size` is configured (see `binaries/daemon/src/spawn/prepared.rs:961-999`), which calls `log::rotate_log_files` (`binaries/daemon/src/log.rs:43-72`):

```rust
// Rename current -> .1
let current = log_path(working_dir, dataflow_id, node_id);   // out/<id>/log_<node>.jsonl
let first   = log_path_rotated(working_dir, dataflow_id, node_id, 1); // ...log_<node>.1.jsonl
if current.exists() { std::fs::rename(&current, &first)?; }
```

…and then the writer re-creates a fresh `log_<node>.jsonl` starting at 0. The follow loop keeps polling the same path string `log_<node>.jsonl`, and `std::fs::metadata` / `File::open` re-resolve that name to the **new** file every poll.

## Concrete failing scenario

Precondition: a node with `max_log_size` configured (a documented option, e.g. `max_log_size: "100MB"`), and `dora logs --local --follow` running against it.

1. Pre-rotation file reaches 20000 bytes → follow loop has `pos = 20000`.
2. The node logs past `max_log_size`; the daemon renames the file to `log_<node>.1.jsonl` and creates a fresh `log_<node>.jsonl` (size 0), then keeps writing.
3. While the fresh file is `<= 20000` bytes, `current_size <= pos` is true → **every line in the new file is silently dropped**.
4. Once the fresh file grows past 20000 bytes, `read_appended_log_lines(path, 20000)` seeks to offset 20000 in the *new* file, **discarding everything before it and almost certainly starting mid-line** → the resumed output is garbled.
5. The rotated content (now in `.1.jsonl`) is never followed either — that path was never in the `files` list.

Net effect: a chunk of logs is lost across every rotation, and output resumes mid-line.

## Suggested fix direction

Detect rotation per poll and reset the offset when the file has been replaced. Options:
- If `current_size < pos`, treat the file as rotated: reset `pos = 0` (and optionally pick up the just-rotated `.1.jsonl` tail so nothing is lost).
- Track the file's identity (inode/`dev`+`ino` on Unix, or creation time) alongside the offset and reset when it changes, rather than relying solely on size going backwards.
- Optionally re-glob `files` each poll so newly-created rotated siblings can be picked up.

## Related lower-severity observation (same function, same offset-boundary class)

The initial offset is `metadata().len()` (line ~301), which is **not** guaranteed to be a newline boundary. If the daemon is caught mid-write (a partial, non-newline-terminated line) at the instant follow starts, the initial dump drops that partial line (fails JSON parse), sets `pos` to end-of-partial, and when the line is later completed the next poll seeks mid-line and reads only the tail — which again fails to parse and is dropped. Net: exactly one log line can be lost at the moment follow begins. Very narrow (requires starting follow during an in-progress write) and low severity, but it shares the same root cause: offsets are advanced/initialized off line boundaries. A fix that snaps the initial offset back to the last newline boundary would close it.

## Notes

- `read_appended_log_lines` itself (the `e792309` change) is correct — verified for the newline-consume offset, the no-newline case, UTF-8 split handling, and overflow. This issue is specifically about the surrounding rotation/offset lifecycle.
- Impact is scoped to `--local --follow` with `max_log_size` configured; the coordinator streaming path (`stream_logs_from_coordinator`) is unaffected.

phil-opp:
## Summary

Fixes #2785.

`dora logs --local --follow` tracks a byte offset per log file and, on each poll, reads only the bytes appended past that offset. But it never handled the file being **replaced** behind its path. When the daemon rotates a node log (`max_log_size` configured — see `binaries/daemon/src/log.rs`), it renames `log_<node>.jsonl` to `log_<node>.1.jsonl` and re-creates a fresh file starting at offset 0. The follow loop kept polling the same path string, which now resolves to the fresh file, while still comparing against the stale (larger) offset:

- while `current_size <= pos`, every new line was **silently dropped** (`continue`);
- once the fresh file grew past `pos`, the reader seeked to the stale offset, **dropping the fresh file's leading bytes and garbling** the remainder.

This is a pre-existing defect in the follow loop's offset management, independent of the inner `read_appended_log_lines` helper (which is correct).

## Fix

Extract the resume-offset decision into a pure helper:

```rust
fn follow_read_offset(pos: u64, current_size: u64) -> Option<u64> {
    let start = if current_size < pos { 0 } else { pos };
    if current_size <= start { None } else { Some(start) }
}
```

A file whose size shrank below the tracked offset was rotated or truncated, so we resume from offset 0 instead of seeking to a now-invalid stale offset. The follow loop calls this helper and skips the poll when it returns `None`.

## Tests

Added four unit tests for `follow_read_offset` covering: normal append, no-change skip, reset-after-rotation, and skip-when-fresh-file-empty. All pass:

```
test result: ok. 4 passed; 0 failed
```

## Validation

- `cargo test -p dora-cli follow_read_offset` ✅ (4 passed)
- `cargo clippy -p dora-cli -- -D warnings` ✅

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01V54sCK4Gp5JR4dwv7KvNRE)_

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -315,10 +315,10 @@ fn follow_local_logs(args: &LogsArgs) -> Result<()> {
         for path in &files {
             let pos = file_positions.get(path).copied().unwrap_or(0);
             let current_size = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
-            if current_size <= pos {
+            let Some(start) = follow_read_offset(pos, current_size) else {
                 continue;
-            }
-            let (msgs, new_pos) = read_appended_log_lines(path, pos)?;
+            };
+            let (msgs, new_pos) = read_appended_log_lines(path, start)?;
             new_messages.extend(msgs);
             file_positions.insert(path.clone(), new_pos);
         }
@@ -329,6 +329,25 @@ fn follow_local_logs(args: &LogsArgs) -> Result<()> {
                 print_log_message(msg, &config);
             }
         }
+    }
+}
+
+/// Decide where to resume reading a followed log file given the last-read
+/// offset `pos` and the file's `current_size`, returning `None` when there is
+/// nothing new to read.
+///
+/// A file whose size shrank below `pos` was rotated or truncated behind its
+/// path (the daemon renames `log_<node>.jsonl` to `log_<node>.1.jsonl` and
+/// re-creates a fresh file at offset 0 when `max_log_size` is configured — see
+/// `binaries/daemon/src/log.rs`). Seeking to the stale `pos` in the fresh file
+/// would silently drop its leading bytes and garble the remainder once it grows
+/// past `pos`, so we resume from offset 0 instead.
+fn follow_read_offset(pos: u64, current_size: u64) -> Option<u64> {
+    let start = if current_size < pos { 0 } else { pos };
+    if current_size <= start {
+        None
+    } else {
+        Some(start)
     }
 }
 
@@ -1078,6 +1097,33 @@ mod tests {
 
         assert_eq!(files.len(), 1);
         assert!(files[0].file_name().unwrap() == "log_cam.jsonl");
+    }
+
+    // --- follow_read_offset (rotation / truncation handling) ---
+
+    #[test]
+    fn follow_read_offset_reads_appended_bytes() {
+        // File grew past the tracked offset: resume from `pos`.
+        assert_eq!(follow_read_offset(10, 25), Some(10));
+    }
+
+    #[test]
+    fn follow_read_offset_skips_when_unchanged() {
+        // No new bytes: nothing to read.
+        assert_eq!(follow_read_offset(25, 25), None);
+    }
+
+    #[test]
+    fn follow_read_offset_resets_after_rotation() {
+        // File shrank (rotated behind the path to a fresh 0-based file): the
+        // stale offset would drop leading bytes, so resume from 0.
+        assert_eq!(follow_read_offset(500, 30), Some(0));
+    }
+
+    #[test]
+    fn follow_read_offset_skips_empty_after_rotation() {
+        // Rotated but the fresh file has no content yet: skip this poll.
+        assert_eq!(follow_read_offset(500, 0), None);
     }
 
     // --- read_appended_log_lines (follow-mode offset tracking) ---


phil-opp:

The `follow_read_offset` helper and its unit tests look correct, and this is a strict improvement over comparing a freshly-rotated file against a stale offset. Two limitations of the size-based rotation detection are worth flagging (neither is a regression):

1. **Rotation is missed when the fresh file outgrows the stale offset within one poll interval.** Rotation is detected only via `current_size < pos`. If the daemon rotates the file and the new file grows *past* the old offset within one ~200 ms poll (e.g. `pos = 500`, fresh file already at 600 bytes), then `current_size >= pos`, so the helper returns `Some(500)` and the loop seeks to byte 500 of the *new* file — dropping its first 500 bytes and garbling the remainder. That is the same failure mode this PR fixes, for the case where the new file outran the offset between polls. A fully robust fix would key on file identity (inode/dev) rather than size.

2. **Unread tail of the pre-rotation file is never read.** `follow_local_logs` captures the file list once and only polls those paths, so bytes written to the current file between the last poll and the rename (moved into `log_<node>.1.jsonl`) are never read if that rotated path wasn't in the initial set. Pre-existing design, but it means "no lines dropped on rotation" is only approximately true.

Otherwise no issues found.

---

PR #2802, opened 2026-07-22
ci(nightly): accept the reworded "up to date" self-update message

phil-opp:
#2512 reworded `dora self update --check-only`'s up-to-date output to "Dora CLI is up to date (current: X, latest release: Y)", but the topic-and-top-smoke assertion still only accepted "latest version" / "update is available". The new wording matched neither, so it fell through to `ERROR: --check-only produced no recognized outcome message` and failed the nightly deterministically whenever the GitHub releases API was reachable (surfaced once #2742's Examples fix let the run reach this job).

Add "up to date" to the valid-outcome grep in both the workflow step (.github/workflows/nightly.yml) and the local repro (scripts/qa/ci-nightly-jobs.sh). Verified the exact failing CI output now matches.

Caught in https://github.com/dora-rs/dora/issues/2742#issuecomment-5044526458

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -662,10 +662,11 @@ jobs:
           # command didn't even start the check path.
           echo "$out" | grep -q "Checking for updates" \
             || { echo "ERROR: --check-only didn't enter update-check path"; exit 1; }
-          # One of the two valid outcomes must print. If neither appears,
-          # either GitHub rate-limited us (tolerated) or the output format
-          # regressed (fail).
-          if echo "$out" | grep -qE "latest version|update is available"; then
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. I verified against the current source that `binaries/cli/src/command/self_.rs` prints `"Dora CLI is up to date (current: X, latest release: Y)"` for the up-to-date case, so adding `up to date` to the valid-outcome grep is the correct fix. The change only extends the existing alternation (`latest version|update is available` → `up to date|latest version|update is available`) in both the workflow step and the local repro script, without removing the prior patterns or altering the rate-limit/fail branches, so it's a safe, minimal fix.

---

PR #2803, opened 2026-07-22
chore(deps): move off yanked spin 0.9.8/0.10.0 to 0.9.9/0.10.1

phil-opp:
## Summary

`spin` **0.9.8** and **0.10.0** (transitive dependencies via the zenoh / ring stack) have been **yanked** from crates.io. `deny.toml` sets `yanked = "deny"`, so `cargo deny check advisories` — the supply-chain **`audit`** CI job — now fails on **every branch, including `main`**, independent of the diff under test. This is currently blocking the Trunk merge queue for all in-flight PRs.

This is not caused by any code change — it's a live-advisory-DB gate reacting to an upstream yank.

## Change

Bump both to their non-yanked patch successors — the remedy `cargo deny` itself suggests (`try cargo update -p spin`):

- `spin 0.9.8 → 0.9.9`
- `spin 0.10.0 → 0.10.1`

**`Cargo.lock` only** — no manifest or source changes, and no transitive cascade (the other 193 locked dependencies are unchanged).

## Verification

- `cargo deny check` → **advisories ok, bans ok, licenses ok, sources ok**
- `cargo check -p dora-daemon -p dora-cli` → builds clean

## Notes

Kept deliberately separate from the feature PRs (#2588, #2589) so their scope stays clean. Merging this unblocks the `audit` gate for those and any other open PR.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -700,7 +700,7 @@ version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a7913f22349ffcfc6ca0ca9a656ec26cfbba538ed49c31a273dff2c5d1ea83d9"
 dependencies = [
- "spin 0.9.8",
+ "spin 0.9.9",
 ]
 
 [[package]]
@@ -2747,7 +2747,7 @@ dependencies = [
  "futures-sink",
  "nanorand",
  "pin-project",
- "spin 0.9.8",
+ "spin 0.9.9",
 ]
 
 [[package]]
@@ -2759,7 +2759,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "nanorand",
- "spin 0.9.8",
+ "spin 0.9.9",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

This is a `Cargo.lock`-only change that bumps the yanked `spin 0.9.8 → 0.9.9` and `spin 0.10.0 → 0.10.1`. I independently verified against the crates.io index that:

- The two new checksums in the diff (`3763264f…` for 0.9.9 and `023a211c…` for 0.10.1) are the genuine registry checksums for those versions.
- `0.9.8` and `0.10.0` are indeed yanked, while `0.9.9` and `0.10.1` are not — so this resolves the `yanked = "deny"` advisory failure as described.

I also confirmed the update is internally consistent at the PR head: no references to the old versions remain, both `[[package]]` entries were updated, and every dependent edge was re-pointed (4 references to `spin 0.9.9`, 1 to `spin 0.10.1`). The bump stays within the same minor series (patch-level), so no transitive cascade is expected. This is the standard `cargo update -p spin` remedy.

---

phil-opp:

I don't think this change is necessary anymore: the same spin bump already landed on `main` via #2629 ("Update yanked spin dependency", `b4b1c28`). Current `main` already carries `spin 0.9.9` / `0.10.1` with the same checksums this diff produces, and no remaining `0.9.8` / `0.10.0` references — the diff only shows `0.9.8 → 0.9.9` because the branch was cut from an older `main`; rebased onto today's `main` it would be empty. So the `audit` gate should already be green on `main`. Unless there's a fresh yank of `0.9.9` / `0.10.1` I'm missing, this can probably be closed as superseded rather than merged. The lock contents themselves check out — checksums match the registry and the edge re-pointing is consistent.

---

PR #2806, opened 2026-07-22
fix(node): prove zenoh startup routes with a marker/ack handshake, switching paths producer-side

### LINKED ISSUE #2738: Readiness barrier (#2666) counts unreachable remote subscribers: fixed 5 s per-node startup stall + lost direct-zenoh fast path for local subscribers in multi-machine dataflows

phil-opp:
Reworks the startup-marker approach after review: the markers' end-to-end evidence stays, but the actor changes. The previous revision had the consumer block on receiving a marker per input and hard-fail init after 10s — yet only the producer chooses the delivery path. That mismatch is what made a static consumer of a dynamic producer abort startup, broke every multi-machine dataflow (markers can't cross the loopback-only node mesh), and silently starved a consumer whose zenoh subscriber failed to declare.

Now the handshake is two-way and the producer decides. Every output starts on the lossless daemon path. The producer publishes markers on the real output topic; each consumer answers every marker with an ack on the output's exact-key `@ack` topic (identity in the attachment — no wildcards, no hex-encoded keys). Once all *required ackers* of an output have acked, the producer switches it to direct zenoh — in practice within one 5ms marker round-trip during init, before the first real send, so messages don't straddle the path switch. Consumers never block and nothing hard-fails: a route that never proves itself keeps its output on the daemon path (warning at a 10s deadline that only arms once the barrier releases, so slow-starting consumers can't burn it).

Required-acker sets and daemon-path pins come from the daemon's **actual placement** at spawn time via a new `NodeConfig.output_routing` field, not from descriptor `deploy` intent — so label-scheduled placement can't be guessed wrong. Any output with a consumer under another daemon is pinned daemon-only, which keeps inter-daemon forwarding fed and fixes #2738 for this branch; #2741 is superseded (its counting half targets code this PR deletes, its pin half is subsumed by the daemon-computed routing). Outputs of `dora node add`-spawned nodes with no current receivers are also pinned: their future consumers can only be wired via `dora node connect`, which delivers on the daemon path only.

Dynamic and restarted producers run the same handshake at join time — consumers ack markers for their whole lifetime — closing the silent first-send loss both had under the previous revision, and giving them the fast path (which the old subscriber counting gave, and the marker revision lost).

The `examples/multiple-daemons` sink now consumes `rust-node/random` cross-machine (an output that also has a same-machine consumer) and fails if forwarding starves, so nightly exercises the mixed local+remote pin end-to-end. The action contract test asserts no output silently froze on the daemon path in a healthy local dataflow.

Known limitations, stated:

- An output with any remote consumer rides the daemon path entirely (local co-consumers included) until direct cross-machine node-to-node zenoh lands — that follow-up is a policy flip in `compute_output_routing` plus endpoint wiring, tracked in a separate issue.
- Mixed-version: a new consumer under an old (previous-revision) producer reintroduces the startup race, since the consumer no longer gates the barrier; and a pre-marker consumer binary under a new producer sees the markers as undecodable empty-payload frames (warn-per-marker for up to 10s, then correct daemon-path delivery). Both bounded by the `Register` version check.
- Pre-existing, unchanged: `dora node connect` edges from a producer already on the direct zenoh path starve (the daemon-side `AddMapping` delivers only on the daemon path, and the consumer has no zenoh subscriber for an undeclared source). This held for the counting design too; the added-node pin above closes it for freshly added producers.


phil-opp:
conflicts with https://github.com/dora-rs/dora/pull/2741

phil-opp:

I read the full diff and the surrounding code on `f9b48e9`. The marker mechanism itself is sound (publishers declared eagerly, marker thread started before any wait so cycles resolve, global barrier keeps every producer emitting until all consumers have proven their routes, `Arc::try_unwrap` after join, correct drop ordering). Two things I'd want resolved before this lands:

**1. Consumers now block on markers from a *dynamic* producer, which can hang/fail startup (regression).**

In `apis/rust/node/src/event_stream/mod.rs:654`, a `RouteProbe` is registered for every user-mapped input, guarded only by `if !dynamic` — where `dynamic` is whether *this* (consumer) node is dynamic. It does **not** consider whether the input's *source* node is dynamic. So a static node consuming `some_dynamic_node/output` registers a probe and blocks in `wait_for_input_routes`, which `eyre::bail!`s after `ZENOH_ROUTE_PROBE_TIMEOUT` (10s) → init fails → dataflow aborts.

A dynamic producer connects at an arbitrary later time (that is the whole point of `path: dynamic`), so while the static consumer is in its pre-`Subscribe` wait, the producer isn't running and emits no markers. This is exactly the case the code being deleted here handled explicitly: `output_subscriber_count` skipped `node_is_dynamic` nodes precisely because "they connect at arbitrary times (or never), so the startup barrier must not wait for them." That exclusion is gone, and the new consumer-side wait has no equivalent guard for dynamic *sources* — the `Input`/`input_config` passed to `EventStream::init` carries no source-dynamic-ness, and the descriptor (which does, via the now-removed `node_is_dynamic`) isn't consulted.

Worse, the mechanism doesn't actually protect this route even when the producer *does* connect: a dynamic node's `EventStream::init` returns almost immediately (the global barrier released long ago), so `StartupMarkers::stop` fires right after it connects — likely before its route to the static consumer has propagated — so its first real sends can still be dropped. Net effect for static-consumer ← dynamic-producer: the old code started fine and never lost data (producer stayed on the daemon path until wired); the new code makes the consumer hard-fail after 10s. The existing dynamic examples only use dynamic nodes as *sinks*, so this direction isn't exercised by the test suite or smoke tests.

Suggest skipping probe registration for inputs whose source is a dynamic node (mirroring the removed `node_is_dynamic` logic, using `dataflow_descriptor`).

**2. Timeout is now a hard failure instead of graceful fallback — please confirm this narrowing is intended.**

The removed `warm_up_direct_outputs` documented that on timeout "the un-connected outputs simply start on the reliable daemon path … the dataflow is never blocked, and nothing is dropped." `wait_for_input_routes` replaces that fail-safe with `eyre::bail!`. That means any deployment where the direct zenoh data plane can't connect but the daemon/coordinator control plane can (e.g. cross-machine with the zenoh data port firewalled, while the coordinator TCP port is open) now fails init entirely rather than running over the daemon path. The PR description frames the hard-fail as the goal, but it's a real behavioral narrowing versus the prior always-degrade-gracefully contract — worth calling out explicitly since the daemon path is still fully functional in those cases.

Everything else (the metadata marker plumbing, the `is_startup_marker` filtering before payload/decoder, the reserved `__dora_` key, the removed hex-key/liveliness machinery) looks correct, and the metadata round-trip tests are meaningful.

---

phil-opp:
Looking again at this revision, I think the earlier "firewall edge case" framing understates the problem. I found the following issue.

**The marker-based route probe breaks every multi-machine dataflow, not just firewalled ones.** A consumer registers a `RouteProbe` for each user input (guarded only by whether *this* consumer node is dynamic), and `wait_for_input_routes` `bail!`s init after `ZENOH_ROUTE_PROBE_TIMEOUT` (10s) if the producer's marker never arrives. But `StartupMarkers` publishes markers only over the direct node-to-node zenoh publishers, and node-to-node zenoh is loopback-only by design: `plan_zenoh_peering` reserves loopback listeners for local nodes only and gives "a local consumer of a remote producer ... no dial for it" (`binaries/daemon/src/spawn/spawner.rs:152-157`). So a static consumer on machine B of a producer on machine A can never receive a marker, and its `init()` hard-fails after 10s, aborting the dataflow.

This is the *default* cross-machine deployment (delivery via daemon-to-daemon forwarding), not a firewall edge case — the marker never crosses the machine boundary at all, firewall or not. The removed `output_subscriber_count` mechanism tolerated this: the producer simply stayed on the daemon path and forwarding kept working. The new hard `bail!` removes that graceful fallback.

This is also the same multi-machine breakage that #2741 addresses from the opposite direction (pinning any output with a remote subscriber to the daemon path). The two changes are mutually exclusive, and this one reintroduces the cross-machine loss that #2741 is trying to prevent — worth reconciling the two approaches before either lands.

---
🤖 _Automated review by Claude — no human was in the loop for this comment. It may contain mistakes, so please verify the findings before acting on them._

---

phil-opp:
---
🤖 **Automated review by Claude** — _this is a fully automated review with no human verification; treat findings as advisory._
---

I reviewed the reworked handshake on `584b5a7` (the design changed substantially since the earlier automated reviews, so this is a fresh pass over the new code, not a re-check).

The two findings from the marker-only revision are genuinely resolved:

- **Static consumer of a dynamic producer no longer hard-fails init.** The consumer side no longer blocks on receiving a marker: `wait_for_input_routes`/`RouteProbe` and the pre-`Subscribe` 10s bail are gone; the data callback just acks markers via `spawn_startup_acker` and never gates init. A route that never proves itself simply keeps the output on the daemon path.
- **Multi-machine no longer breaks.** `compute_output_routing` pins any output with a consumer under another daemon to `daemon_only`, so it never leaves the daemon path and inter-daemon forwarding stays fed (#2738). The `multiple-daemons` example now exercises the mixed local+remote case and fails if forwarding starves.

Spot-checks that hold up: `AckState::record` only counts identities in the required set (dupes/strangers ignored) and flips `ready` exactly when `received == required`; `daemon_only` outputs get no publisher and no ack state, so `required_ackers` on them are correctly inert; the marker thread is joined in `Drop` before the publishers `Arc` is released, so publishers are undeclared inside the bounded teardown; markers are filtered by `is_startup_marker()` at both the node data callback and the `dora topic` inspection path before the decoder; exact-key `@ack` topics avoid the `cmd`/`cmd/vel` collision without the old hex-wildcard machinery. The added unit tests (ack-set coverage, exact-identity match, routing table tests, marker/ack wire round-trips) are meaningful, not tautological.

One thing I could not confirm from the diff, worth a sanity check before merge: the required-acker identity for a **runtime/operator** consumer is `operator_id/input_id` (asserted in `runtime_operator_inputs_are_operator_prefixed`), and correctness depends on the runtime's consumer-side `EventStream` acking under that exact same `op/input` key. The runtime binary isn't in this diff, so I couldn't verify the two sides agree — if they don't, a dataflow with an operator consumer would never complete that output's handshake and would silently stay on the daemon path (correct, but the fast path is lost). Worth a targeted check or a smoke test with an operator consumer.

No blocking issues found in the new design.

---

phil-opp:
**Automated review** — this comment was generated fully automatically by Claude; no human has reviewed the diff.

I reviewed the latest diff and found one shutdown-robustness issue that the earlier reviews didn't cover:

In `DoraNode::drop` (`apis/rust/node/src/node/mod.rs`), the startup handshake is stopped at the very top of the function, *before and outside* the `teardown_with_timeout("zenoh", ZENOH_TEARDOWN_TIMEOUT, …)` block that follows:

```rust
let ack_subscribers = match self.startup_handshake.take() {
    Some(mut handshake) => {
        handshake.shutdown();   // sets stop flag, then handle.join() on the marker thread
        std::mem::take(&mut handshake.ack_subscribers)
    }
    None => Vec::new(),
};
```

`handshake.shutdown()` sets the stop flag and then unconditionally `handle.join()`s the marker thread. The marker thread's loop only checks the stop flag at the top of each iteration, then blocks in `publisher.put(&[][..]).wait()`. Setting `stop = true` cannot interrupt an in-progress `.wait()`, so on a wedged zenoh net runtime that `put().wait()` can block indefinitely and the unbounded `join()` stalls node shutdown — and with it the daemon, which waits for `InputClosed`. That is exactly the failure mode #2425 and the `teardown_with_timeout` guard were introduced to bound. It's reachable when a node process exits while its handshake is still in flight (route not yet proven, within the 10s marker window) and the session is unhealthy.

The consumer side already handles this correctly: `EventStream::drop` joins the `dora-startup-acker` thread *inside* `teardown_with_timeout("zenoh-subscribers", …)`. Making the producer symmetric — running `handshake.shutdown()` / the marker-thread join inside a bounded `teardown_with_timeout` (still before the publishers are dropped, so the join keeps its "release the last publisher `Arc`" property) — would close the gap.

Aside from this, I found no issues in the diff.

---

phil-opp:

I looked at the commits pushed since the last review (the #2883 marker-thread-join fix, the `main` merge, and the unwrap-budget reconcile). The shutdown-robustness issue from my earlier review is resolved: in `impl Drop for DoraNode` (`apis/rust/node/src/node/mod.rs`), `handshake.shutdown()` and the ack-subscriber teardown now run **inside** the `teardown_with_timeout("zenoh", ZENOH_TEARDOWN_TIMEOUT, …)` closure and still before `drop(publishers)`, so the marker-thread join is bounded by the deadline (symmetric with the consumer-side acker join) while preserving the documented publisher → subscriber → session drop order. The no-zenoh branch drops the handshake inline. The unwrap-budget change is a baseline reconcile, not a code change. No new issues found in these commits.

---

PR #2809, opened 2026-07-23
fix(core/build): atomically promote git clones so a crashed build can self-heal

### LINKED ISSUE #2808: core/build: #2795 leaves a crashed build's half-written git clone permanently un-recoverable (Reuse HEAD-check bails without cleanup)

## Summary

PR #2795 (`996741c`, *"don't let a concurrent session's HEAD-check delete a git clone that's still being written"*) changed the `Reuse`-arm HEAD verification so that when a clone's HEAD **fails to resolve**, it now `bail!`s and **leaves the directory in place** instead of deleting it. That correctly closes #2711 (deleting a clone a concurrent process is still writing), but it introduces a new regression: a **half-written clone left behind by a crashed build (SIGKILL / power loss) can never self-heal** — every subsequent build of that `repo@<full-commit-hash>` re-enters the same failing path and bails, so the build is permanently wedged until someone manually `rm -rf`s the directory.

The commit's own rationale — *"An unresolvable HEAD fails loudly and leaves the dir where it is, so a retry or the owning build can recover it"* — does **not** hold for a crash leftover: there is no owning build (the process died), and a retry deterministically hits the same bail.

## Introduced by

- PR #2795 / commit `996741c9e442c0ee4b6d0cdea4cd7c4b67e7647` (specifically its second sub-commit, *"only delete a clone whose HEAD resolves to a different commit"*), merged 2026-07-22.
- Pre-#2795 behavior (from #2482): the `Err` arm called `cleanup_failed_clone()` and re-cloned — i.e. it **self-healed** this exact case.

## Affected code paths (`libraries/core/src/build/git.rs`)

1. **A leftover dir on disk is treated as reusable.** `clone_dir_ready` returns `true` on `dir.exists()`:
   ```rust
   pub fn clone_dir_ready(&self, session_id: SessionId, dir: &Path) -> bool {
       self.prepared_builds
           .get(&session_id)
           .map(|p| p.planned_clone_dirs.contains(dir))
           .unwrap_or(false)
           || dir.exists()                      // git.rs:196-201
   }
   ```
   So `choose_clone_dir` picks the `Reuse` arm for a half-written leftover.

2. **A crashed process leaves nothing in the in-memory in-progress set,** so verification is armed:
   ```rust
   let in_progress = lock_in_progress(&self.clones_in_progress).contains_key(&clone_dir);
   ReuseOptions::Reuse {
       dir: clone_dir.clone(),
       verify_commit: (!in_progress).then_some(commit_hash),   // git.rs:120-124 → Some(hash)
   }
   ```
   `clones_in_progress` is an in-process `Arc<Mutex<..>>`; after a crash it is empty, so `in_progress == false` and `verify_commit == Some(hash)`.

3. **The `Err` arm bails without cleanup** (the regression):
   ```rust
   match head {
       Ok(h) if h.eq_ignore_ascii_case(&commit_hash) => {}
       Ok(h) => { cleanup_failed_clone(logger, &dir).await; bail!(...) }   // deletes on wrong-commit
       // HEAD didn't resolve → half-written repo → NO delete, bail forever:
       Err(err) => bail!(
           "couldn't verify clone dir {} is on commit {commit_hash}: {err:?}; \
            leaving it in place in case another build is writing it, please retry",
           dir.display()
       ),                                                                  // git.rs ~411-421
   }
   ```

4. **`NewClone` clones directly into the target path** (no temp-then-rename), so a half-written repo can exist at the final path in the first place:
   ```rust
   let repository = clone_into(repo_url.clone(), &clone_target)?;   // git.rs:262
   checkout_tree(&repository, &commit_hash)?;
   ```
   A SIGKILL between `clone_into` starting and completion leaves an unopenable `.git` at `clone_dir`.

## Failure scenario

1. `dora build` a dataflow with a git node pinned to a **full commit hash**; the daemon starts `clone_into(target_dir)`.
2. The daemon (or machine) is SIGKILLed / loses power mid-clone. `cleanup_failed_clone` (only reached via the in-build `Err` arms) never runs, so a half-written repo is left at `clone_dir`. `clones_in_progress` is gone with the process.
3. Restart the daemon and rebuild the same `repo@commit`.
4. `clone_dir_ready` → `true` (`dir.exists()`) → `Reuse` arm; `in_progress == false` → `verify_commit == Some(hash)`.
5. `Repository::open` / `head()` fails on the half-written repo → `Err` → **bail, no cleanup**.
6. Every subsequent rebuild repeats step 4–5. The build is **permanently wedged** with no automated recovery; the user must manually delete the clone directory.

(For a **branch/tag** pin the outcome differs but is also bad: `is_full_commit_hash` is `false`, so the HEAD check is skipped entirely and the half-written dir is silently returned as *"reusing up-to-date"*, and the build then fails later using a broken checkout.)

## Why the naive fix doesn't work, and the suggested fix

The reason #2795 stopped deleting on `Err` is that an unresolvable HEAD is ambiguous: it could be a crash leftover (safe to delete) **or** a clone another *process* is mid-writing (unsafe to delete — the in-progress set only covers this process's `GitManager`; the daemon and the CLI each have their own). You cannot distinguish these two from in-process state alone, so simply restoring the delete would reopen #2711 for the cross-process case.

The robust fix is to make a directory's *existence at the target path* imply a *complete* clone:

- **Clone into a temporary sibling dir and atomically `rename` into `clone_dir` only after `checkout_tree` succeeds.** Then a half-written clone only ever exists under a temp name; a crash leftover is a temp dir (sweepable by age, and never matched by `clone_dir_ready`/`Reuse`), and any dir at `clone_dir` is by construction openable — so the `Err` arm becomes genuinely stale-vs-live only for corruption that happens *after* a good clone, which is rare and can be handled by an mtime/age heuristic or an on-disk lock/heartbeat file that survives across processes.

A lighter-weight alternative: keep the current bail for a *fresh* leftover but add a cross-process staleness signal (e.g. only refuse to delete when a lock/heartbeat file in the dir was touched within the last N seconds), so an abandoned crash leftover is eventually reclaimed.

## Secondary, lower-confidence note (same function, same PR)

The `Ok(h)` *different-commit* arm still `remove_dir_all`s unconditionally. During a `RenameAndFetch`/`CopyAndFetch` window a live clone in **another process's** `GitManager` legitimately sits on the old commit; a concurrent `Reuse` in *this* process would resolve HEAD to that old commit and delete the live dir. This is only reachable if two processes target the *same on-disk clone path* (normally they use different `base_working_dir`s), so it is narrow — but it means the cross-process variant of #2711 that #2795 set out to close is not fully closed for the `Ok(different-commit)` path either. A cross-process claim signal (as above) would cover this too.

phil-opp:
## Summary

Fixes #2808.

`ReuseOptions::NewClone` cloned directly into the final `clone_dir`, so a crash (SIGKILL / power loss) *mid-clone* left a half-written repo at that path. Since #2795 the Reuse arm no longer deletes a clone whose HEAD fails to resolve (correctly, to avoid re-opening #2711 for a concurrently-written clone). But that turned a crash leftover into a **permanent wedge**: every subsequent build of the same `repo@commit` re-enters the Reuse arm, fails to resolve HEAD, and `bail!`s *without* cleanup — with no owning process left to recover it, only a manual `rm -rf` unblocks the build. For a branch/tag pin the outcome differs but is also bad: the HEAD check is skipped, so the broken checkout is silently reused.

## Fix

Fix the regression at its root rather than restoring the risky delete-on-`Err`:

- Clone + checkout into a **unique temporary sibling** dir and `rename` it into `clone_dir` **only after** the checkout fully succeeds. `rename` within the same parent directory is atomic, so a crash only ever leaves a *temp* dir — never a directory at `clone_dir`. Any directory that exists at `clone_dir` is therefore, by construction, a complete clone, and `clone_dir_ready` / the Reuse arm can trust it again.
- Temp names are `pid`+counter unique, so no build ever writes into or reclaims another live build's temp dir — this keeps the change **cross-process safe** and does not re-open #2711.
- A best-effort, **age-gated** sweep reclaims temp dirs abandoned by crashed builds (only dirs older than 1h, which no live clone can be) so leftovers don't accumulate on disk.
- The `git2::Repository` handle is dropped before the rename so Windows (which refuses to rename a dir with open handles) is not broken.

This also fixes the branch/tag-pin variant, since a half-written checkout can no longer exist at `clone_dir` in the first place.

## Tests

New unit tests in `libraries/core/src/build/git.rs`:
- `new_clone_promotes_temp_into_target_and_cleans_up` — clone lands at the target atomically, is a valid repo on the requested commit, and no temp sibling survives.
- `new_clone_failure_leaves_no_target_dir` — a failed clone leaves nothing at the target path (the property that prevents the wedge).
- `partial_clone_path_is_a_unique_sibling` — temp paths are distinct siblings, never the target.
- `sweep_removes_abandoned_partial_dirs_only` — the sweep removes stale temps and leaves the target / unrelated siblings alone.

All 16 `git::` tests pass; `cargo fmt` and `cargo clippy -p dora-core --features build --all-targets -- -D warnings` are clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AYVWUwXULA9QoNbYJcLsF4)_

### DIFF
--- a/libraries/core/src/build/git.rs
+++ b/libraries/core/src/build/git.rs
@@ -6,7 +6,11 @@ use itertools::Itertools;
 use std::{
     collections::{BTreeMap, BTreeSet},
     path::{Path, PathBuf},
-    sync::{Arc, Mutex},
+    sync::{
+        Arc, Mutex,
+        atomic::{AtomicU64, Ordering},
+    },
+    time::Duration,
 };
 use url::Url;
 
@@ -267,12 +271,34 @@ impl GitFolder {
                         ),
                     )
                     .await;
-                let clone_target = target_dir.clone();
+
+                // Clone into a temporary sibling dir and only atomically
+                // `rename` it into `target_dir` once the checkout has fully
+                // succeeded. That way a crash (SIGKILL / power loss) mid-clone
+                // leaves a half-written repo under the temp name, *never* at
+                // `target_dir` -- so a later build's `clone_dir_ready` check
+                // (which trusts `dir.exists()`) can't mistake a broken leftover
+                // for a good clone and wedge itself permanently on the
+                // un-resolvable HEAD in the Reuse arm (#2808). Any directory
+                // that *does* exist at `target_dir` is, by construction, a
+                // complete checkout.
+                let tmp_dir = partial_clone_path(&target_dir);
+
+                // Best-effort: reclaim temp dirs abandoned by earlier crashed
+                // builds so they don't accumulate. Only clearly-stale ones are
+                // swept (see helper), never a temp another build may still be
+                // writing right now.
+                sweep_stale_partial_clones(logger, &target_dir, PARTIAL_CLONE_MAX_AGE).await;
+
+                let clone_target = tmp_dir.clone();
                 let checkout_result = match tokio::task::spawn_blocking(move || {
                     let repository = clone_into(repo_url.clone(), &clone_target)
                         .with_context(|| format!("failed to clone git repo from `{repo_url}`"))?;
                     checkout_tree(&repository, &commit_hash)
                         .with_context(|| format!("failed to checkout commit `{commit_hash}`"))
+                    // `repository` is dropped here, before the rename below, so
+                    // no git2 handles remain open on the temp dir (Windows
+                    // refuses to rename a dir with open handles).
                 })
                 .await
                 {
@@ -287,12 +313,35 @@ impl GitFolder {
                 };
 
                 match checkout_result {
-                    Ok(()) => target_dir,
+                    Ok(()) => {
+                        // Promote the finished clone into place atomically.
+                        match tokio::fs::rename(&tmp_dir, &target_dir).await {
+                            Ok(()) => target_dir,
+                            // Another build won the race and already put a
+                            // complete clone at `target_dir` (rename onto a
+                            // populated dir fails). Drop ours and reuse theirs.
+                            Err(_) if target_dir.exists() => {
+                                cleanup_failed_clone(logger, &tmp_dir).await;
+                                target_dir
+                            }
+                            Err(err) => {
+                                logger
+                                    .log_message(LogLevel::Error, format!("{err:?}"))
+                                    .await;
+                                cleanup_failed_clone(logger, &tmp_dir).await;
+                                bail!(
+                                    "failed to move finished clone from {} into {}: {err}",
+                                    tmp_dir.display(),
+                                    target_dir.display()
+                                )
+                            }
+                        }
+                    }
                     Err(err) => {
                         logger
                             .log_message(LogLevel::Error, format!("{err:?}"))
                             .await;
-                        cleanup_failed_clone(logger, &target_dir).await;
+                        cleanup_failed_clone(logger, &tmp_dir).await;
                         bail!(err)
                     }
                 }
@@ -504,6 +553,74 @@ async fn cleanup_failed_clone(logger: &mut impl BuildLogger, dir: &Path) {
                     ),
                 )
                 .await;
+        }
+    }
+}
+
+/// How long an abandoned `NewClone` temp dir must have sat untouched before a
+/// later build reclaims it. A live clone is younger than a fresh checkout, so
+/// this comfortably exceeds any real in-progress clone while still bounding how
+/// long a crash leftover lingers on disk (#2808).
+const PARTIAL_CLONE_MAX_AGE: Duration = Duration::from_secs(60 * 60);
+
+/// Filename prefix shared by every in-flight `NewClone` temp dir for `target`.
+/// Sits alongside `target` (whose final component is the commit hash), begins
+/// with a dot so it's visually distinct from real clone dirs, and can never
+/// collide with a sibling commit-hash dir.
+fn partial_clone_prefix(target: &Path) -> String {
+    let name = target
+        .file_name()
+        .map(|n| n.to_string_lossy())
+        .unwrap_or_default();
+    format!(".{name}.partial-")
+}
+
+/// A unique temp path (sibling of `target`) to clone into before the atomic
+/// rename into place. The name is `<prefix><pid>-<counter>`: the pid keeps it
+/// distinct across processes sharing a working dir, and the process-wide
+/// counter keeps concurrent clones of the same commit apart. Uniqueness (rather
+/// than a fixed `.partial` name) is what keeps this cross-process safe -- no
+/// build ever writes into, or reclaims, another live build's temp dir.
+fn partial_clone_path(target: &Path) -> PathBuf {
+    static COUNTER: AtomicU64 = AtomicU64::new(0);
+    let n = COUNTER.fetch_add(1, Ordering::Relaxed);
+    let pid = std::process::id();
+    target.with_file_name(format!("{}{pid}-{n}", partial_clone_prefix(target)))
+}
+
+/// Best-effort removal of `NewClone` temp dirs for `target` abandoned by
+/// crashed builds. Only dirs whose mtime is at least `max_age` old are removed,
+/// so a temp another build is *currently* cloning into -- necessarily younger
+/// than a completed checkout -- is never touched. Combined with the unique temp
+/// names, this reclaims clearly-dead leftovers without ever racing a live clone
+/// in another process. Any error (unreadable dir, missing entry) is ignored:
+/// this is opportunistic housekeeping, not a correctness requirement.
+async fn sweep_stale_partial_clones(
+    logger: &mut impl BuildLogger,
+    target: &Path,
+    max_age: Duration,
+) {
+    let Some(parent) = target.parent() else {
+        return;
+    };
+    let prefix = partial_clone_prefix(target);
+    let Ok(mut entries) = tokio::fs::read_dir(parent).await else {
+        return;
+    };
+    while let Ok(Some(entry)) = entries.next_entry().await {
+        let name = entry.file_name();
+        let Some(name) = name.to_str() else { continue };
+        if !name.starts_with(&prefix) {
+            continue;
+        }
+        let stale = match entry.metadata().await.and_then(|m| m.modified()) {
+            // `elapsed()` errors if the mtime is in the future (clock skew);
+            // treat that as "not yet stale" and leave the dir alone.
+            Ok(mtime) => mtime.elapsed().map(|age| age >= max_age).unwrap_or(false),
+            Err(_) => false,
+        };
+        if stale {
+            cleanup_failed_clone(logger, &entry.path()).await;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Cloning into a unique `<pid>-<counter>` temp sibling and renaming into place only after a successful checkout is a sound way to guarantee that any directory present at `clone_dir` is a complete clone, so a crashed build can self-heal on the next run. The concurrent-winner race (`rename` onto an already-populated target) is handled by dropping the local temp and reusing the existing clone, the `git2::Repository` is dropped before the rename (Windows handle safety), and the age-gated sweep won't touch a live clone because the in-progress temp dir's mtime stays fresh.

The tests cover the key invariants: `new_clone_failure_leaves_no_target_dir` (wedge prevention), `sweep_removes_abandoned_partial_dirs_only` (age gate in both directions), and `partial_clone_path_is_a_unique_sibling` (uniqueness / same-parent).

---

phil-opp:

The temp-dir + atomic-rename approach looks like the right root-cause fix — any dir at `target_dir` being complete-by-construction cleanly closes the #2808 wedge. Two non-blocking edge cases on the housekeeping side that might be worth a note:

**1. The stale-sweep's age gate can misjudge a very long live clone.** The comment on `sweep_stale_partial_clones` reasons that a temp being actively cloned into is "necessarily younger than a completed checkout." That holds for the temp dir's *own* mtime only while its top-level entries are being written — but `git clone` spends the bulk of its time writing into `.git/objects/…` (a subdir), which doesn't bump the top-level dir's mtime. For a large repo / slow network whose fetch exceeds `PARTIAL_CLONE_MAX_AGE` (1h), the top-level temp mtime can be >1h old while the clone is still live, and since the sweep matches every pid's `.<commit>.partial-*`, a second process cloning the same commit could `remove_dir_all` a live temp. It's self-healing (the interrupted clone just fails and retries, nothing lands at `target_dir`), but it does undercut the invariant the comment states. A pid-liveness check, or touching the temp dir periodically, would make the gate match the claim.

**2. The sweep only runs in the `NewClone` arm.** Once a good clone exists at `target_dir`, every later build of that `repo@commit` takes the `Reuse` arm and never sweeps — so a `.<commit>.partial-*` leftover from a *concurrent* build that crashed (while another finished the clone) never gets reclaimed for that commit. Minor, but it narrows the "so they don't accumulate on disk" guarantee to commits that still have no successful clone.

Neither blocks the fix.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review, not vetted by a human. I looked at the diff and the surrounding code, not the PR description.

I found one correctness gap: the atomic-promotion invariant only holds for the `NewClone` path, but the change documents it as a global property.

The temp-dir + atomic-`rename` approach is sound for `NewClone` — the temp is a sibling of `target_dir` (same filesystem, so `rename` is genuinely atomic), the concurrent-winner `ENOTEMPTY` reuse is handled, and the `Repository` is dropped before the rename for Windows. The tests cover that path well.

But the comment the change adds asserts:

> Any directory that *does* exist at `target_dir` is, by construction, a complete checkout.

That isn't true for the other two arms that write `clone_dir`, and they're untouched here:

- **`CopyAndFetch`** calls `std::fs::create_dir_all(&to)` with `to = target_dir` and then copies content + fetches into that same directory — non-atomically. If the build is SIGKILLed / loses power **during the copy phase** (the exact threat model this PR targets), a partial directory with an incomplete `.git` is left sitting at `target_dir`. The next build then hits `clone_dir_ready` → `dir.exists()` → the `Reuse` arm, where `git2::Repository::open` / HEAD resolution fails, landing on the `Err(err) => bail!(...)` branch that **intentionally leaves the dir in place** (the #2711 tradeoff). That's precisely the permanent #2808 wedge, repeating on every rebuild until someone manually removes the cache dir. (A crash during the *fetch/checkout* phase instead leaves a resolvable HEAD on the old commit, which the `Reuse` arm self-heals via the `Ok(h)`-mismatch cleanup — so the vulnerable window is specifically the copy phase.)
- **`RenameAndFetch`** is mostly self-healing for the same reason (its leftover is a complete clone at the previous commit), but a crash mid-`checkout_tree` can still corrupt HEAD into the same leave-in-place bail.

So the fix closes the wedge for `NewClone` but leaves `CopyAndFetch` exposed to the identical failure it means to eliminate "at the root," and the documented invariant over-claims. Consider routing `CopyAndFetch` (and the fetch/checkout mutation in `RenameAndFetch`) through the same temp-dir + atomic-rename promotion so the "complete-by-construction" property actually holds for every path that writes `clone_dir`.

Two prior points from earlier reviews still stand and aren't addressed by the current diff: the stale-sweep age gate can misjudge a very long live clone (the top-level temp mtime isn't bumped by writes into `.git/objects/…`), and the sweep only runs in the `NewClone` arm so `.partial-*` leftovers are never reclaimed once a good clone exists at `target_dir`.

---

phil-opp:
The mechanics are right: the temp dir is a sibling so the rename is same-device and atomic, `git2::Repository` is dropped inside the `spawn_blocking` closure before the rename (Windows-safe), the concurrent-winner race is handled, and every error path calls `cleanup_failed_clone`.

The gap is scope, not mechanism. The invariant this writes into the code — "any directory that exists at `clone_dir` is, by construction, a complete clone" — only holds for `NewClone`. `CopyAndFetch` still does `create_dir_all` and copies in place (`libraries/core/src/build/git.rs:314`), and `RenameAndFetch` renames the old clone straight onto `target_dir` (`git.rs:357`). SIGKILL mid-`fs_extra::dir::copy` still leaves a partial `.git` at `clone_dir`, the Reuse arm's HEAD read fails, and you get the exact #2808 permanent wedge. Either extend temp-then-promote to those two arms, or narrow the claim in the comment.

Two smaller things:
- None of the four new tests goes RED without the fix — `new_clone_failure_leaves_no_target_dir` and `new_clone_promotes_temp_into_target_and_cleans_up` both pass pre-fix, because the old code already called `cleanup_failed_clone(&target_dir)`. The actual property (crash mid-clone) is untested.
- `sweep_stale_partial_clones` reasons that "a live clone is necessarily younger" from the temp dir's *top-level* mtime, which git does not touch during the fetch phase (objects land under `.git/`). So a clone running longer than an hour can be `remove_dir_all`'d out from under a concurrent build of the same commit.

phil-opp:
Thanks — all three are fair. Pushed a follow-up commit that addresses them:

**1. Scope — extended temp-then-promote to all write arms.** Factored the atomic promotion into a shared `promote_clone(tmp, target)` and routed `CopyAndFetch` and `RenameAndFetch` through a temp sibling too:
- `CopyAndFetch` now `create_dir_all` + copies + fetches + checks out in the temp dir, then promotes.
- `RenameAndFetch` renames the old clone into the temp dir (not straight onto `target_dir`), fetches + checks out there, then promotes.

So the "any dir at `clone_dir` is a complete clone" invariant now holds for every arm — a SIGKILL mid-`fs_extra::dir::copy` or mid-fetch leaves a temp dir, never a broken `.git` at `clone_dir`.

**2. Racy sweep — removed entirely.** You're right that the top-level mtime doesn't advance while git writes objects under `.git/`, so a long clone could be reclaimed out from under a concurrent build. I dropped `sweep_stale_partial_clones` rather than try to patch the heuristic. Orphaned temp dirs from a hard crash are harmless (unique names, never matched by `clone_dir_ready`/Reuse) and are left for external cleanup; a safe reclaimer would need a cross-process liveness signal (advisory lock / pid-liveness), which is out of scope here.

**3. Tests.** New tests that genuinely exercise the mechanism:
- `promote_clone` unit tests: absent-target move, and the concurrent-winner race (temp dropped, winner's clone reused untouched) — both red without the new code path.
- End-to-end promote tests for `CopyAndFetch` and `RenameAndFetch` via a local `origin` remote, asserting the checkout lands at `target_dir` and no temp sibling survives.
- Failure-path tests for all three arms asserting no `target_dir` **and** no temp leftover.

On the honest limit you flagged: a true *SIGKILL-mid-clone* can't be reproduced in an in-process unit test, because every in-process error path runs `cleanup_failed_clone`, which masks the difference between writing-in-place and writing-to-temp. That's exactly why the fix makes the invariant hold *structurally* — `target_dir` is only ever created by the final `rename` — rather than relying on cleanup. The `promote_clone` and per-arm tests lock that structural boundary.

The follow-up also merged cleanly with #2820 on the rebased base: a panic in the blocking clone/checkout task now routes through the temp-dir cleanup + `bail!` path.

---

phil-opp:
Heads up on the merge-queue failure (`error[E0063]: missing field \`ready_barrier_released\` in \`state::RunningDataflow\`` at `binaries/coordinator/src/handlers.rs:911`): that wasn't from this PR — it only touches `libraries/core/src/build/git.rs`. The branch had been rebased onto an older `main` snapshot (before `ready_barrier_released` was added to the coordinator), so the queue's merge with current `main` produced a stale-base semantic conflict in the coordinator.

Fixed by rebasing the branch onto the latest `origin/main` (`fb8c4e9`) and force-pushing. Verified locally on rustc 1.95.0: `dora-coordinator` compiles (E0063 gone) and the 22 `dora-core` `build::git` tests pass. No code changes beyond the rebase.

---

phil-opp:

The follow-up commit (`b8a8383`) resolves the scope gap from the earlier pass: `promote_clone` is now the shared tail of all three write arms, so `target_dir` is only ever created by the final atomic `rename` and the "any dir at `clone_dir` is a complete clone" invariant now holds for `CopyAndFetch` and `RenameAndFetch`, not just `NewClone`. The concurrent-winner reuse (`Err(_) if target.exists()`), the drop-`Repository`-before-rename ordering (Windows-safe), and the per-arm `cleanup_failed_clone` all look right, and the new `promote_clone` unit tests plus per-arm promote/failure tests exercise the real mechanism (the race test is genuinely red without the new path). The racy `sweep_stale_partial_clones` is gone from the net diff with no dangling caller.

Two small leftovers, neither blocking:
- Removing the sweep means temp dirs orphaned by a hard crash/SIGKILL now stay in the clone cache indefinitely (unique-named, never reused). That's a reasonable trade vs. the racy reclaimer, but it drops the earlier "so they don't accumulate on disk" guarantee — worth a note that external cleanup is now expected.
- The PR description still says "a best-effort, age-gated sweep reclaims temp dirs," which no longer matches the code now that the sweep is removed.

No new blocking issues.

---

Issue #2585, opened 2026-07-10
daemon: node added via `AddNode`/`dora node add` never receives a timer input for an interval no existing node uses (no timer task is ever spawned)


phil-opp:
> **Note:** This issue was created by an automated, scheduled Claude code-review check. Please verify before acting on it.

## Summary

When a node is added to a *running* dataflow via the dynamic-topology path (`DaemonCoordinatorEvent::AddNode`, surfaced as `dora node add`), and that node subscribes to a `dora/timer/...` interval that **no already-running node was subscribed to at dataflow start**, the node never receives any tick on that input. The timer subscription is registered in the routing table, but the background task that actually *generates* the ticks for that interval is never spawned, so no `DoraEvent::Timer` is ever produced for it.

## Location

**1. `AddNode` registers the timer subscriber but never spawns the timer task** — `binaries/daemon/src/lib.rs:2004`:

```rust
InputMapping::Timer { interval } => {
    dataflow
        .timers
        .entry(*interval)
        .or_default()
        .insert((node_id.clone(), input_id.clone()));
}
```

The `AddNode` handler closure ends at `lib.rs:2093` (`.await` at 2094) without ever calling `dataflow.start(...)`.

**2. Timer ticks are produced only by per-interval tasks spawned in `RunningDataflow::start()`** — `binaries/daemon/src/running_dataflow.rs:302`:

```rust
pub(crate) async fn start(&mut self, events_tx: ..., clock: ...) -> eyre::Result<()> {
    for interval in self.timers.keys().copied() {
        if self._timer_handles.contains_key(&interval) {
            continue;                       // idempotency guard (from #1149's fix)
        }
        // ... spawn tokio::time::interval task that emits DoraEvent::Timer ...
        self._timer_handles.insert(interval, handle);
    }
    Ok(())
}
```

**3. `start()` is only ever called from two sites, both gated behind the initial all-nodes-ready transition** — `binaries/daemon/src/lib.rs:1671` (coordinator `AllNodesReady`) and `binaries/daemon/src/lib.rs:3487`:

```rust
DataflowStatus::AllNodesReady if !dataflow.dataflow_started => {
    // ...
    dataflow.start(&self.events_tx, &self.clock).await?;
    dataflow.dataflow_started = true;      // never re-entered afterwards
}
```

Once the dataflow is running (`dataflow_started == true`), neither site fires again.

**4. Delivery reads the live subscriber set** — `binaries/daemon/src/lib.rs:4279`:

```rust
[...]

PR #2810, opened 2026-07-23
fix(daemon): cancel a timer task when RemoveNode drains its last subscriber

phil-opp:
## Summary

Addresses the lower-severity observation in #2585.

The **primary** bug in #2585 (a node added via `AddNode` to a running dataflow never receiving ticks for a brand-new interval) was already fixed on `main` by #2474, which re-invokes `dataflow.start()` from the `AddNode` handler. This PR closes out the issue's remaining, explicitly-noted **secondary** observation.

`RemoveNode` removed the departing node from every `timers` subscriber set but never touched `_timer_handles`. When the removed node was the *last* subscriber to an interval, that interval's per-interval task kept ticking and dispatching `DoraEvent::Timer`s to an empty subscriber set for the remaining life of the dataflow — a bounded but pointless stream of wakeups.

## Fix

Extract the timer-unsubscribe logic into `RunningDataflow::unsubscribe_node_from_timers`, which:
- drops the node from each interval's subscriber set, and
- for any interval left with no subscribers, cancels its timer task (dropping the `futures::future::RemoteHandle` stored in `_timer_handles` cancels the spawned future) and forgets the now-empty `timers` entry.

A later `AddNode` re-subscribing to the same interval re-spawns the task via the idempotent `start()` (its `_timer_handles.contains_key` guard makes re-invocation safe), so behavior for the re-subscribe case is preserved.

## Tests

`unsubscribe_last_subscriber_cancels_timer_task` — with one interval subscribed only by node `a` and another shared by `a` and `b`, removing `a` drops the solo interval's entry *and* its task, while the shared interval keeps its remaining subscriber and its task.

`cargo fmt --all -- --check` is clean and `cargo clippy -p dora-daemon --lib -- -D warnings` passes. (An unrelated pre-existing `match_like_matches_macro` lint in the `matches_event` test helper is untouched, per the repo's "don't fix unrelated warnings" convention.)

---
_Generated by [Claude Code](https://claude.ai/code/session_01AYVWUwXULA9QoNbYJcLsF4)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2314,9 +2314,9 @@ impl Daemon {
                     // both to stop delivering to a removed node and so a re-added
                     // ID is classified by its own inputs, not stale timer/log
                     // state (which would mark it never-finishing forever, #2270).
-                    for receivers in dataflow.timers.values_mut() {
-                        receivers.retain(|(nid, _)| nid != &node_id);
-                    }
+                    // Cancels the timer task of any interval left with no
+                    // subscribers (#2585); see the method for details.
+                    dataflow.unsubscribe_node_from_timers(&node_id);
                     dataflow
                         .log_subscribers
                         .retain(|sub| sub.node_id != node_id);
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. The extracted `unsubscribe_node_from_timers` correctly removes both the `timers` entry and the `_timer_handles` `RemoteHandle` only for intervals whose subscriber set becomes empty, and dropping that handle cancels the spawned task. Intervals that still have another subscriber keep both their entry and their task. The re-add case stays safe via `start()`'s `_timer_handles.contains_key` guard.

The regression test `unsubscribe_last_subscriber_cancels_timer_task` exercises the real spawn path (`start()`) and distinguishes the drained solo interval (entry + handle both gone) from the still-subscribed shared interval (subscriber + handle retained), so it would actually fail if the handle weren't cancelled — not a trivially-passing test.

---

PR #2818, opened 2026-07-24
fix(core): resolve module outputs produced by operator/custom inner nodes

### LINKED ISSUE #2817: Module expansion rejects a declared output produced by an operator/custom inner node (output-side counterpart of #2441)
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. It was found by randomly selecting a component (`libraries/core/src/descriptor/expand.rs`) and reviewing it for correctness issues, then verified against the current source and existing tests. Please treat the severity assessment as a starting point for human triage.

## Summary

In `libraries/core/src/descriptor/expand.rs`, module **output** resolution only looks at the node-level `Node::outputs` field. For a **runtime node** (`operators:` / `operator:`) or a legacy **`custom:`** inner node, the produced outputs live in `OperatorConfig::outputs` (`operators[].config.outputs`) / `NodeRunConfig::outputs` (`custom.run_config.outputs`), and the node-level `outputs` set is empty. As a result, a module that declares an output produced by an operator/custom inner node is wrongly rejected.

This is the **output-side counterpart** of #2441 (which fixed the same blind spot for *inputs* by routing operator/custom `config.inputs` through `node_input_maps`). The input side is now first-class, but the output side never got the parallel treatment — there is no `node_output_maps` equivalent.

## Where

**Phase 3 output resolution (real expansion), `expand_module_node`:**

```rust
// expand.rs:548-561
let mut output_map: BTreeMap<String, String> = BTreeMap::new();
for declared_output in &module_file.module.outputs {
    let producer = final_nodes
        .iter()
        .find(|n| n.outputs.contains(declared_output))   // <-- node-level outputs only
        .ok_or_else(|| {
            eyre::eyre!(
                "module `{}` declares output `{}` but no inner node produces it",
                module_file.module.name, declared_output,
            )
        })?;
    output_map.insert(declared_output.to_string(), producer.id.to_string());
}
```

**Isolated validator `check_module_file` has the same blind spot:**

```rust
// expand.rs:199-204
let mut inner_outputs: BTreeSet<String> = module_file
    .nodes
    .iter()
    .filter(|n| n.module.is_none())
    .flat_map(|n| n.outputs.iter().map(|o| o.to_string()))   // <-- node-level outputs only
    .collect();
// ... expand.rs:240-249: bails "declares output `X` but no inner node produces it"
```

For reference, the output fields that are ignored:
- `Node::outputs` — `libraries/message/src/descriptor.rs:437` (node-level, empty for operator/custom nodes)
- `OperatorConfig::outputs` — `descriptor.rs:957`
- `CustomNode::run_config` → `NodeRunConfig::outputs`

## Reproduction

Module file with an operator inner node producing the module's declared output:

```yaml
module:
  name: rt
  inputs: [data]
  outputs: [result]          # declared module output
nodes:
  - id: runtime
    operators:
      - id: proc
        shared-library: proc.so
        inputs: { x: _mod/data }
        outputs: [result]      # produced by the operator, NOT node-level
```

**Expected:** the module expands, and `result` maps to the operator's producer node.
**Actual:** both `expand_module_node` and `check_module_file` bail with *"module `rt` declares output `result` but no inner node produces it"* — even though this is a legitimate, fully-wired module (and its inputs already work post-#2441). Any downstream node wiring `<module_id>/result` then also fails to resolve.

## Impact

Any module whose declared outputs are produced by operator (runtime) or legacy `custom` inner nodes cannot be used. Since #2441 made operator/custom inner nodes first-class on the **input** side, this output-side gap is an asymmetric oversight rather than an intended limitation.

## Test gap

No test covers this path. The #2441 operator/custom regression tests (`expand_rewrites_operator_inputs`, `expand_rewrites_custom_node_inputs`, `check_module_file_rejects_invalid_mod_ref_in_operator_input`) use modules that declare no `outputs:` (or fail earlier on an input check), so output resolution for an operator/custom producer is never exercised.

## Suggested fix

Introduce a `node_output_maps(node)` helper mirroring `node_input_maps` (enumerating `node.outputs` + `operators[].config.outputs` + `operator.config.outputs` + `custom.run_config.outputs`), and use it in both the Phase-3 `find(|n| ...)` producer lookup and the `check_module_file` `inner_outputs` collection. Add a regression test for a module whose declared output is produced by an operator (and by a custom) inner node.

phil-opp:
## Summary

Fixes #2817.

Module **output** resolution in `libraries/core/src/descriptor/expand.rs` only consulted the node-level `Node::outputs` set. For a **runtime** node (`operators:` / `operator:`) or a legacy **`custom:`** inner node, the produced outputs live in `OperatorConfig::outputs` / `NodeRunConfig::outputs` and the node-level `outputs` set is empty. As a result, a module that declared an output produced by such a node was wrongly rejected with *"module `X` declares output `Y` but no inner node produces it"*, and any downstream node wiring that output then also failed to resolve.

This is the **output-side counterpart of #2441**, which fixed the same blind spot on the *input* side by routing operator/custom `config.inputs` through `node_input_maps`. The output side never got the parallel treatment.

## Changes

- Add a `node_output_maps(node)` helper mirroring `node_input_maps`, enumerating `node.outputs` + `operators[].config.outputs` + `operator.config.outputs` + `custom.run_config.outputs`.
- Use it in the Phase-3 producer lookup in `expand_module_node` (real expansion).
- Use it in the isolated `check_module_file` validator's `inner_outputs` collection.

## Tests

- `expand_resolves_operator_produced_module_output` — a module whose declared output is produced by an operator inner node now expands, and a downstream consumer of `m/result` resolves to the producer node (`m.runtime` / `result`). This failed before the fix.
- `check_module_file_accepts_operator_and_custom_produced_outputs` — `check_module_file` now accepts modules whose declared output is produced by an operator or a legacy `custom:` inner node.

`cargo test -p dora-core`, `cargo clippy -p dora-core`, and `cargo fmt --all -- --check` all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FK32zqkgDm6L44ucZb27RA)_

### DIFF
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -195,12 +195,16 @@ pub fn check_module_file(module_path: &Path) -> eyre::Result<()> {
     }
 
     // Check outputs: each declared output should be produced by some inner node.
+    // Runtime (operator) and legacy custom nodes declare their outputs through
+    // config.outputs / run_config.outputs rather than the node-level `outputs`
+    // set, so those need the same treatment as inputs (see #2817).
     // For nested module children, recursively load their declared outputs.
     let mut inner_outputs: BTreeSet<String> = module_file
         .nodes
         .iter()
         .filter(|n| n.module.is_none())
-        .flat_map(|n| n.outputs.iter().map(|o| o.to_string()))
+        .flat_map(|n| node_output_maps(n))
+        .flat_map(|outputs| outputs.iter().map(|o| o.to_string()))
         .collect();
 
     // Check nested module files exist and collect their declared outputs
@@ -298,6 +302,26 @@ fn node_input_maps(node: &Node) -> Vec<&BTreeMap<DataId, Input>> {
     }
     if let Some(ref custom) = node.custom {
         maps.push(&custom.run_config.inputs);
+    }
+    maps
+}
+
+/// Collect every output set a node produces. Runtime (`operators:` / `operator:`)
+/// and legacy `custom:` nodes declare their produced outputs in
+/// `config.outputs` / `run_config.outputs` rather than the node-level
+/// `outputs` set, which stays empty for them. Output resolution must consult
+/// all of these, mirroring [`node_input_maps`] on the input side (see #2817,
+/// the output-side counterpart of the #2441 input fix).
+fn node_output_maps(node: &Node) -> Vec<&BTreeSet<DataId>> {
+    let mut maps = vec![&node.outputs];
+    if let Some(ref operators) = node.operators {
+        maps.extend(operators.operators.iter().map(|op| &op.config.outputs));
+    }
+    if let Some(ref operator) = node.operator {
+        maps.push(&operator.config.outputs);
+    }
+    if let Some(ref custom) = node.custom {
+        maps.push(&custom.run_config.outputs);
     }
     maps
 }
@@ -549,7 +573,11 @@ fn expand_module_node(
     for declared_output in &module_file.module.outputs {
         let producer = final_nodes
             .iter()
-            .find(|n| n.outputs.contains(declared_output))
+            .find(|n| {
+                node_output_maps(n)
+                    .iter()
+                    .any(|outputs| outputs.contains(declared_output))
+            })
             .ok_or_else(|| {
                 eyre::eyre!(
                     "module `{}` declares output `{}` but no inner node produces it",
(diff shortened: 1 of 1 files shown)


phil-opp:
I reviewed the diff (`descriptor/expand.rs`). The new `node_output_maps` helper correctly mirrors `node_input_maps`, consulting `node.outputs` plus `operators[].config.outputs`, `operator.config.outputs`, and `custom.run_config.outputs`, and it's applied consistently to both the `check_module_file` `inner_outputs` collection and the Phase-3 producer lookup in `expand_module_node`. This is the output-side counterpart of the #2441 input fix and looks sound. Both regression tests are genuine — they assert an operator-produced module output resolves to the runtime node with output `result`, and that `check_module_file` accepts operator- and custom-produced outputs.

No issues found.

---
🤖 *This is a fully automated review by Claude ([Claude Code](https://claude.ai/code)). No human has reviewed these comments — please verify before acting.*

---

phil-opp:

I found the following issue: the fix resolves module outputs produced by **plural `operators:` (runtime) inner nodes** into wiring that later fails validation, and the added test doesn't catch it because it stops at the `expand_modules` stage.

For a module output produced by an `operators:` node, expansion rewrites the downstream consumer reference to `source = m.<node>`, `output = <declared_output>` — the bare output name, with no operator segment. But operator outputs are only auto-prefixed with the operator id at resolve time for **singular `operator:`** nodes: `single_operator_nodes` in `libraries/core/src/descriptor/mod.rs:133-141` filters on `n.operator` (singular), so `prefix_output_with_operator_id` (called at `mod.rs:151`/`:177`) never runs for a plural-`operators:` producer. The output therefore stays bare, and `libraries/core/src/descriptor/validate.rs:483-495` then rejects it:

```
input `y` references output `result` of node `m.runtime`, which is a runtime node;
runtime node outputs must include the operator id
(expected format: `m.runtime/<operator_id>/<output_id>`)
```

**Repro:** the exact module in this PR's own `expand_resolves_operator_produced_module_output` test (a `runtime` node with `operators: [{ id: proc, outputs: [result] }]`, consumer `y: m/result`). A full `Descriptor::check` (expand → `resolve_aliases_and_set_defaults` → `check_dataflow`) fails on it. The test only calls `expand_modules` and asserts `m.output == "result"` — but `"result"` is precisely the bare form that `check_dataflow` rejects, so the test passes while `dora check` / `dora run` would fail on the same input.

The `custom:` path is fine (custom outputs are validated directly against `run_config.outputs`, no operator segment required — `validate.rs:475-482`), and the singular `operator:` path is fine (auto-prefixed at resolve). Only the plural `operators:` case is broken, and that's the case the new tests exercise.

Suggested fix: when the producing inner node is a runtime (`operators:`) node, rewrite the resolved output to `<operator_id>/<output_id>` (find which operator's `config.outputs` declares it), and add a test that runs `check_dataflow` end-to-end rather than asserting only the intermediate expand mapping.

---

libraries/core/src/descriptor/expand.rs
```diff
@@ -302,6 +306,26 @@ fn node_input_maps(node: &Node) -> Vec<&BTreeMap<DataId, Input>> {
...
+        maps.extend(operators.operators.iter().map(|op| &op.config.outputs));
+    }
+    if let Some(ref operator) = node.operator {
+        maps.push(&operator.config.outputs);
+    }
+    if let Some(ref custom) = node.custom {
+        maps.push(&custom.run_config.outputs);
+    }
+    maps
+}
```
keirsalterego, line 327: it needs to return (output_name, output_ref) pairs where output_ref is \"{op.id}/{output}\... 
this drops the operator id for multi operator nodes else there would be no way to know which operator produced what if there is more than one..

phil-opp:
CLosed in favor of #2872

PR #2819, opened 2026-07-24
fix(ros2-bridge): enforce BoundedSequence max_size for Binary byte columns

### LINKED ISSUE #2816: ros2-bridge: `BoundedSequence<uint8>` max_size is not enforced when the byte field is supplied as an Arrow `Binary`/`LargeBinary` column
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. It was found by randomly selecting a component (the `dora-ros2-bridge-arrow` serialize path) and reviewing it for correctness issues, then verified against the current source and existing tests. Please treat the severity assessment as a starting point for human triage.

## Summary

On the **dora → ROS2** serialize path, the `BoundedSequence` capacity bound (`max_size`) is **silently not enforced** when a byte-sequence field (`sequence<uint8>` / `char` / `byte`, i.e. ROS2 `uint8[<=N]`) is provided as an Arrow `Binary`/`LargeBinary` array. The identical value supplied as an Arrow `List<UInt8>` **is** bounded correctly, so the two accepted Arrow representations of a byte vector diverge on validation.

This is a distinct defect from the (already-fixed) wire double-nesting bug #2507 — the CDR bytes emitted here are well-formed; the problem is that an over-capacity bounded sequence is accepted and transmitted instead of being rejected.

## Affected code

`libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs`

The top-level dispatch accepts a `Binary`/`LargeBinary` column and slices it to a **1-row** array (`entry`):

```rust
// sequence.rs:44-59
} else if let Some(list) = self.column.as_binary_opt::<i32>() {
    Arc::new(list.slice(0, 1)) as ArrayRef      // <-- entry.len() == 1 (row count)
} else if let Some(list) = self.column.as_binary_opt::<i64>() {
    Arc::new(list.slice(0, 1)) as ArrayRef
}
```

The bound check then compares against `entry.len()`:

```rust
// sequence.rs:66-75
if let Some(max) = self.max_size
    && entry.len() > max                        // Binary: 1 > N  -> always false for N >= 1
{
    return Err(error(format!(
        "sequence length {} exceeds BoundedSequence max_size {}",
        entry.len(), max
    )));
}
```

- For the **`List<UInt8>`** path, `entry = list.value(0)` is the inner element array, so `entry.len()` is the true element count and the bound is enforced.
- For the **`Binary`/`LargeBinary`** path, `entry` is the 1-row binary array, so `entry.len()` is always `1` (the number of rows), *not* the byte count. A 100-byte payload against `max_size = 10` evaluates `1 > 10 == false` and passes; `serialize_binary` (sequence.rs:281-311) then emits all 100 bytes.

`max_size: Some(v.max_size)` is set for every `BoundedSequence` field at `libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs:209-215`, so the bound is genuinely expected to hold on this path.

## Concrete failing scenario

A `BoundedSequence<uint8>` field (`uint8[<=10]`) whose column is supplied as an Arrow `Binary` array (the idiomatic Arrow encoding for bytes, e.g. what a `pyarrow` producer yields) with 100 bytes:

- **Expected:** serialize fails with `"sequence length 100 exceeds BoundedSequence max_size 10"`.
- **Actual:** serialize succeeds and emits all 100 bytes. A strict ROS2 subscriber whose generated code asserts the bounded-sequence capacity (`<= N`) may reject the message or overflow a fixed-capacity buffer.

## Test gap

The existing byte-sequence test `byte_sequence_binary_matches_list_through_real_cdr` (sequence.rs:538) exercises only an **unbounded** `Sequence`, never a `BoundedSequence` with a Binary column, so nothing asserts the bound holds for this representation.

## Suggested fix

Measure the actual element/byte count for the binary arms before comparing against `max` — e.g. use `binary.value(0).len()` (the byte count of the single row) for the `Binary`/`LargeBinary` paths — rather than the row count `entry.len()`. Add a regression test for a `BoundedSequence<uint8>` supplied as a `Binary` column that exceeds `max_size`, asserting it errors identically to the `List<UInt8>` representation.

phil-opp:
## Summary

Fixes #2816.

On the **dora → ROS2** serialize path, the `BoundedSequence` capacity bound (`max_size`) was **silently not enforced** when a byte-sequence field (`uint8[<=N]`) was supplied as an Arrow `Binary`/`LargeBinary` column. The identical value supplied as an Arrow `List<UInt8>` **was** bounded correctly, so the two accepted Arrow representations of a byte vector diverged on validation.

### Root cause

In `libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs`, the `Binary`/`LargeBinary` arms slice the column to a **1-row** array (`entry`). The bound check then compared `max_size` against `entry.len()`, which for those arms is the **row count** (always `1`), not the byte count. A 100-byte payload against `uint8[<=10]` evaluated `1 > 10 == false`, passed, and all 100 bytes were emitted. A strict ROS2 subscriber asserting the bounded-sequence capacity could reject the message or overflow a fixed-capacity buffer.

### Fix

Track the true element count (`seq_len`) alongside `entry`:
- `List`/`LargeList`: the inner element array length (unchanged behavior).
- `Binary`/`LargeBinary`: the single row's byte length (`list.value(0).len()`).

The bound check now uses `seq_len`, so both Arrow encodings enforce `max_size` identically.

## Tests

- `bounded_byte_sequence_binary_enforces_max_size` — a `uint8[<=10]` field with a 100-byte payload is now rejected identically whether supplied as a `List<UInt8>` or a `Binary` column, and a within-capacity `Binary` payload still serializes. This failed before the fix (the Binary case silently serialized).

`cargo test -p dora-ros2-bridge-arrow`, `cargo clippy -p dora-ros2-bridge-arrow`, and `cargo fmt --all -- --check` all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FK32zqkgDm6L44ucZb27RA)_

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs
@@ -25,14 +25,23 @@ impl serde::Serialize for SequenceSerializeWrapper<'_> {
     where
         S: serde::Serializer,
     {
-        let entry = if let Some(list) = self.column.as_list_opt::<i32>() {
+        // `entry` is the array passed to the element serializers; `seq_len` is
+        // the true number of sequence elements used for the BoundedSequence
+        // bound check. For `List`/`LargeList` these coincide (`entry` is the
+        // inner element array). For `Binary`/`LargeBinary`, `entry` is a 1-row
+        // binary array — its `len()` is the row count (always 1), so the
+        // element count must be read from the row's byte length instead
+        // (see #2816).
+        let (entry, seq_len) = if let Some(list) = self.column.as_list_opt::<i32>() {
             if list.len() != 1 {
                 return Err(error(format!(
                     "expected single-element list, got length {}",
                     list.len()
                 )));
             }
-            list.value(0)
+            let entry = list.value(0);
+            let seq_len = entry.len();
+            (entry, seq_len)
         } else if let Some(list) = self.column.as_list_opt::<i64>() {
             if list.len() != 1 {
                 return Err(error(format!(
@@ -40,7 +49,9 @@ impl serde::Serialize for SequenceSerializeWrapper<'_> {
                     list.len()
                 )));
             }
-            list.value(0)
+            let entry = list.value(0);
+            let seq_len = entry.len();
+            (entry, seq_len)
         } else if let Some(list) = self.column.as_binary_opt::<i32>() {
             if list.len() != 1 {
                 return Err(error(format!(
@@ -48,7 +59,8 @@ impl serde::Serialize for SequenceSerializeWrapper<'_> {
                     list.len()
                 )));
             }
-            Arc::new(list.slice(0, 1)) as ArrayRef
+            let seq_len = list.value(0).len();
+            (Arc::new(list.slice(0, 1)) as ArrayRef, seq_len)
         } else if let Some(list) = self.column.as_binary_opt::<i64>() {
             if list.len() != 1 {
                 return Err(error(format!(
@@ -56,7 +68,8 @@ impl serde::Serialize for SequenceSerializeWrapper<'_> {
                     list.len()
                 )));
             }
-            Arc::new(list.slice(0, 1)) as ArrayRef
+            let seq_len = list.value(0).len();
+            (Arc::new(list.slice(0, 1)) as ArrayRef, seq_len)
         } else {
             return Err(error(format!(
                 "value is not compatible with expected sequence type: {:?}",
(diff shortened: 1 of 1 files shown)


phil-opp:
I reviewed the diff (`sequence.rs`). The fix is correct: for the `Binary`/`LargeBinary` arms the old check compared `max_size` against `entry.len()`, which was the 1-row array length (always 1), so the bound never fired; reading the element count from the row byte length (`list.value(0).len()`) and using `seq_len` for the check restores parity with the `List<UInt8>` path. `list.value(0)` correctly accounts for array offsets, and the `List`/`LargeList` arms keep their existing `entry.len()` element count. The regression test is genuine — it exercises both over-capacity (rejected) and within-capacity (accepted) Binary payloads and asserts the specific error message.

No issues found.

---
🤖 *This is a fully automated review by Claude ([Claude Code](https://claude.ai/code)). No human has reviewed these comments — please verify before acting.*

---

PR #2820, opened 2026-07-24
fix(core): don't re-panic the build when a git clone/checkout task panics

phil-opp:
## Issue

In `libraries/core/src/build/git.rs`, the `NewClone` arm of `reuse_or_clone` ran the blocking git clone + checkout like this:

```rust
let checkout_result = tokio::task::spawn_blocking(move || { ... })
    .await
    .unwrap();   // <-- unwraps the JoinError
```

If the blocking task **panics** (e.g. a panic bubbling out of the `git2` FFI) or is aborted, the `JoinError` returned by `.await` is `.unwrap()`ed and **re-panics the whole build process**. That bypasses the `cleanup_failed_clone(...)` + `bail!` recovery arm immediately below it, leaving a half-written clone directory on disk that a later build could mistake for a good clone — the exact hazard that #2480 guards against for the copy-and-fetch path.

This arm is also the *only* `spawn_blocking` in the file that unwraps its join result. Every sibling propagates the `JoinError` cleanly (`.await??` at lines 320 and 580, `.await.context(...)?` at 403, `task.await??` in `build/mod.rs`).

## Fix

Convert a task panic into an `eyre` error and route it through the **same** cleanup + `bail!` arm as an ordinary clone/checkout failure, instead of unwrapping:

```rust
let checkout_result = match tokio::task::spawn_blocking(move || { ... }).await {
    Ok(result) => result,
    Err(join_err) => Err(eyre::Report::new(join_err))
        .context("git clone/checkout task panicked"),
};
```

Now a panicked clone/checkout task produces a clean build error *and* runs `cleanup_failed_clone`, so no half-written clone is left behind.

## Validation

- `cargo build -p dora-core --features build`
- `cargo clippy -p dora-core --features build -- -D warnings` — clean
- `cargo fmt -p dora-core -- --check` — clean
- `cargo test -p dora-core --features build` — 295 passed, 0 failed (plus doctests)

Note: the `git` module is behind the `build` feature, so verification requires `--features build`.

---

---

### DIFF
--- a/libraries/core/src/build/git.rs
+++ b/libraries/core/src/build/git.rs
@@ -268,14 +268,23 @@ impl GitFolder {
                     )
                     .await;
                 let clone_target = target_dir.clone();
-                let checkout_result = tokio::task::spawn_blocking(move || {
+                let checkout_result = match tokio::task::spawn_blocking(move || {
                     let repository = clone_into(repo_url.clone(), &clone_target)
                         .with_context(|| format!("failed to clone git repo from `{repo_url}`"))?;
                     checkout_tree(&repository, &commit_hash)
                         .with_context(|| format!("failed to checkout commit `{commit_hash}`"))
                 })
                 .await
-                .unwrap();
+                {
+                    Ok(result) => result,
+                    // A panic in the blocking clone/checkout task must not abort the
+                    // whole build: route it through the same cleanup + `bail!` arm as
+                    // an ordinary clone error so we don't leave a half-written clone
+                    // behind for the next build to reuse (#2480).
+                    Err(join_err) => {
+                        Err(eyre::Report::new(join_err)).context("git clone/checkout task panicked")
+                    }
+                };
 
                 match checkout_result {
                     Ok(()) => target_dir,


phil-opp:

This is the summary comment for an automated code-review pass over `origin/main` (`117dfd6`). Each finding was verified against current `main`, checked against the open PRs from prior runs to avoid duplicates, and shipped as its own focused PR off a fresh `origin/main`. All six PRs are **draft** and **machine-generated** (Claude / Anthropic's Claude Code) — a human should review before merging.

### PRs opened, by category

**Correctness / robustness**
- **#2820** `fix(core)` — `git.rs` ran the blocking clone/checkout as `spawn_blocking(...).await.unwrap()`; a panic in the blocking task re-panicked the whole build and bypassed the `cleanup_failed_clone` + `bail!` arm (leaving a half-written clone, the #2480 hazard). Now routes a `JoinError` through the same cleanup path, matching every sibling `spawn_blocking` in the file.
- **#2823** `fix(download)` — `parse_content_disposition_filename` used a raw `find("filename=")` substring search, which false-matched a longer token (`xfilename="evil"; filename="good.bin"` → `evil`). Now matches on a quote-aware parameter boundary.
- **#2824** `fix(hub-client)` — `read_capped` read into a `String` before its size check, so an oversize entry whose cap byte split a multi-byte char surfaced a confusing "invalid UTF-8" instead of "too large". Now reads bytes, checks the cap, then converts.
- **#2825** `fix(ros2-bridge)` — Arrow→CDR serialization of single-element scalar/string fields ignored the Arrow null validity bit, silently encoding a null as `0`/`false`/`""`. Now rejects nulls, matching the sibling `arrow-convert` and `mavlink2-bridge` layers.

**Security / untrusted-input hardening**
- **#2822** `fix(cli)` — hub `sanitize` filtered only Cc controls (`char::is_control`), letting bidi/format characters (CVE-2021-42574 Trojan-Source set: RLO, isolates, marks, line/paragraph separators) through to the terminal. Now strips them too; ordinary/CJK text is untouched.

**Documentation**
- **#2821** `docs(node-api)` — `Scheduler::next` carried a stale comment claiming "oldest timestamp using BTreeMap Ordering"; the type has no `BTreeMap` and does a plain `VecDeque` LRU rotation. Comment corrected (no behavior change).

### Validation

Each PR was verified independently (fmt + clippy + affected-crate tests). The union of all six was then cherry-picked onto a scratch branch off `origin/main` and validated once as a combined diff:
- `cargo clippy` on all affected crates (incl. `dora-core --features build`) `-- -D warnings` — clean
- `cargo test --all` (excluding Python/examples) — **passed**
- Manual review + simplification pass — no duplication introduced (shared `check_not_null` / `split_disposition_params` helpers), no dead code, each change isolated to its own crate.

### Verified findings that were **rejected** (not shipped), with reasons

- **hub-client `is_valid_key_part` doc** has a duplicated lead-in (doc-merge artifact) — cosmetic only; skipped to stay within the per-run PR budget.
- **hub `search.rs`** aborts the whole cross-namespace scan on one package's `versions()?` IO error (vs. per-entry tolerance elsewhere) — real inconsistency but low impact and carries behavior-change risk; deprioritized.
- **node `thread.rs`** keeps polling after `AllInputsClosed` and drops later events — could not confirm a user-visible bug; likely intentional (`AllInputsClosed`→`Stop`), so raised as a question rather than a fix.
- **node `json_to_arrow.rs`** indexes `batch.column(0)` unconditionally — only panics on a caller-supplied zero-field schema, not a normal input; confidence too low to act.
- **coordinator-store `DataflowStatus`** — four public variants lack doc comments — pure doc nit.
- **core `output_subscriber_count`** re-resolves/clones the full descriptor per call (potential O(n²)) — the hot caller is outside the verified scope; couldn't confirm it's invoked per-output on a hot path, so not verified enough to act on.
- **redb `get_*` vs `list_*`** corrupt-row handling asymmetry — verified **intentional** (pinned by an existing test); not a defect.

---

phil-opp:
**Automated review by Claude — this is a fully automated review with no human in the loop.**

Took a close look at this one against the current file. The change is correct and does what it says:

- Both arms of the new `match` yield `eyre::Result<()>` — the `Ok(result) => result` arm and the `Err(join_err) => Err(eyre::Report::new(join_err)).context(...)` arm — so `checkout_result` keeps its type and the existing `match checkout_result { Ok(()) => …, Err(err) => { cleanup_failed_clone(…); bail!(err) } }` arm below now also catches a panicked/aborted blocking task. `eyre::Report::new(join_err)` is valid because `JoinError: std::error::Error`.
- This makes the `NewClone` arm consistent with its siblings, which already propagate the `JoinError` rather than unwrapping it (`fetch_changes` and the copy path use `.await??`; the Reuse HEAD-read uses `.await.context("HEAD read task panicked")?`). Worth noting the Reuse arm deliberately does *not* clean up on a task failure (it leaves the dir, per the #2711 reasoning), so routing `NewClone`'s panic into cleanup is the right call specifically because a fresh `NewClone` dir has no concurrent-writer ambiguity to worry about.

One observation, not a blocker: the PR adds no regression test for the behavior it changes (the panic path). The sibling arms have cleanup tests, but nothing exercises "blocking clone/checkout task panics → dir is cleaned up." Inducing a real `git2` panic is awkward, but the arm could be made testable by factoring the `spawn_blocking` join + cleanup into a small helper that takes a closure, letting a test pass one that `panic!()`s. As-is the change is still a clear net improvement over `.unwrap()`.

No other issues found.

---

PR #2822, opened 2026-07-24
fix(cli): strip bidi/format characters in hub `sanitize`, not just Cc controls

phil-opp:
## Issue

`binaries/cli/src/command/hub/common.rs::sanitize` neutralizes untrusted, index-supplied strings (package name, description, keywords, examples) before the `dora hub info` / `search` / `list` / `outdated` commands print them — its stated purpose is to stop "a hostile index entry [from] inject[ing] terminal escapes."

```rust
pub fn sanitize(s: &str) -> String {
    s.chars().filter(|c| !c.is_control()).collect()
}
```

`char::is_control()` is `true` only for Unicode category **Cc** (C0/C1 controls, including ESC `0x1B`). It returns **false** for the bidirectional and separator format characters used in Trojan-Source-style spoofing ([CVE-2021-42574](https://www.cve.org/CVERecord?id=CVE-2021-42574)):

- U+202A–U+202E — bidi embeddings and overrides (e.g. RIGHT-TO-LEFT OVERRIDE)
- U+2066–U+2069 — bidi isolates
- U+200E / U+200F — left/right-to-left marks, U+061C — arabic letter mark
- U+2028 / U+2029 — line / paragraph separators

A description or keyword field carrying one of these survives `sanitize` and can visually reorder or split the rendered line on the terminal, even after Cc controls are stripped.

## Fix

Extend the filter to also drop that bidi/separator set. The change is deliberately scoped to the known display-reordering controls — ordinary text, including accented Latin and CJK characters, is untouched (verified by a passthrough test). Zero-width joiners/non-joiners (U+200C/U+200D), which are load-bearing in several scripts, are intentionally **not** stripped since they don't reorder output.

## Validation

- `cargo fmt -p dora-cli -- --check` — clean
- `cargo test -p dora-cli --lib command::hub::common` — 3 passed, 0 failed (new tests cover escape/control removal, bidi/isolate/separator removal, and untouched ordinary/CJK text)

---

---

### DIFF
--- a/binaries/cli/src/command/hub/common.rs
+++ b/binaries/cli/src/command/hub/common.rs
@@ -7,10 +7,29 @@ use dora_hub_client::{
 };
 use eyre::Context;
 
-/// Strip control characters from index-supplied strings before printing them
-/// (a hostile index entry could otherwise inject terminal escapes).
+/// Strip control and bidirectional/format characters from index-supplied
+/// strings before printing them (a hostile index entry could otherwise inject
+/// terminal escapes or visually reorder the rendered line).
+///
+/// `char::is_control` only covers Unicode category **Cc** (C0/C1 controls such
+/// as ESC). It does *not* cover the bidi and separator format characters used
+/// in Trojan-Source-style spoofing (CVE-2021-42574) — right-to-left overrides,
+/// isolates, and line/paragraph separators — which can reorder or split a line
+/// on the terminal even after the controls are gone. Drop those too.
 pub fn sanitize(s: &str) -> String {
-    s.chars().filter(|c| !c.is_control()).collect()
+    s.chars()
+        .filter(|c| {
+            !c.is_control()
+                && !matches!(
+                    *c,
+                    '\u{061C}'                // arabic letter mark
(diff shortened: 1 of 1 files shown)


phil-opp:
I reviewed the diff (`binaries/cli/src/command/hub/common.rs`). The change extends `sanitize` to drop the bidirectional/separator format characters that `char::is_control` (Unicode Cc only) misses. The stripped set — U+061C, U+200E–U+200F, U+2028–U+2029, U+202A–U+202E, U+2066–U+2069 — is exactly the Trojan-Source (CVE-2021-42574) reordering set, and the `matches!` ranges are correct and inclusive. Zero-width joiners/non-joiners (U+200C/U+200D) are deliberately left in, which is right since they don't reorder output, and the passthrough test confirms accented Latin and CJK are untouched.

The three unit tests are meaningful: they assert removal of ESC/BEL/C0 controls, removal of RLO/isolate/marks/separators, and preservation of ordinary text — the bidi-removal assertions would fail against the old Cc-only filter, so they actually pin the new behavior.

No issues found.

---

---

PR #2823, opened 2026-07-24
fix(download): match Content-Disposition `filename` on a parameter boundary

phil-opp:
## Issue

`libraries/extensions/download/src/lib.rs::parse_content_disposition_filename` found the filename with a raw substring search:

```rust
let start = lower.find("filename=")? + "filename=".len();
```

`find("filename=")` matches the literal `filename=` **anywhere** in the header, including inside a longer parameter token. So a header like:

```
attachment; xfilename="evil"; filename="good.bin"
```

resolved to `evil` — the value of the unrelated `xfilename` parameter — instead of the genuine `filename` parameter's `good.bin`. (The RFC 5987 extended `filename*=` form was already skipped, but only incidentally: it happens to contain a `*` before the `=`.)

This is on the download path used to fetch node/operator binaries. It requires an unusual or adversarial header, so severity is low, but the result is picking the wrong filename off an attacker-influenced header.

## Fix

Parse the header into its `;`-separated parameters and match the parameter whose **name before `=`** equals `filename` case-insensitively — a real parameter boundary rather than a substring:

- Splitting is **quote-aware**, so a `;` inside `filename="a;b.bin"` still yields a single parameter (preserving the existing behavior).
- `filename*` (extended form) and look-alikes such as `xfilename` no longer compare equal to `filename`.

All prior behaviors are retained: quoted and unquoted forms, trailing-parameter stripping, case-insensitive parameter name, semicolon-inside-quotes, `filename*=`/plain coexistence, and empty→`None`.

## Validation

- `cargo fmt -p dora-download -- --check` — clean
- `cargo clippy -p dora-download --all-targets -- -D warnings` — clean
- `cargo test -p dora-download` — 12 passed, 0 failed (adds a regression test asserting `xfilename="evil"; filename="good.bin"` → `good.bin`, and a bare `xfilename` → `None`)

---

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -11,24 +11,51 @@ use tokio::io::AsyncWriteExt;
 /// `attachment; filename="model.bin"; size=1000`, and both quoted and unquoted
 /// forms. Per RFC 6266 / RFC 2616 the parameter name is case-insensitive, so
 /// `Filename`, `FILENAME`, etc. are matched too. Returns `None` when no
-/// non-empty `filename=` value is present (the RFC 5987 extended `filename*=`
-/// form is not decoded and falls through — searching for the literal
-/// `filename=` skips it, since `filename*=` does not contain that substring).
+/// non-empty `filename` value is present.
+///
+/// The `filename` parameter is matched on a real parameter boundary — the
+/// name before its `=`, compared case-insensitively to exactly `filename` —
+/// not by a substring search. That distinguishes it from the RFC 5987 extended
+/// `filename*=` form (which this does not decode) and from an unrelated longer
+/// token such as `xfilename=` that merely ends in `filename`.
 fn parse_content_disposition_filename(header: &str) -> Option<String> {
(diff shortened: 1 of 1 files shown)


phil-opp:

The substring-match bug is real (`find("filename=")` matches inside `xfilename=`), and the rewrite fixes it by splitting into `;`-separated parameters and comparing the trimmed name before `=` case-insensitively to exactly `filename`. I walked the notable edge cases:

- `filename*=` extended form → name before `=` is `filename*`, so it is correctly excluded.
- `xfilename=` → excluded.
- `=` inside a quoted value (`filename="a=b"`) → `split_once('=')` splits at the first `=`, leaving the quoted value intact for the quote-strip step.
- quote-aware `;` splitter keeps `filename="a;b.bin"` as one parameter, preserving prior behavior.
- a parameter with no `=` yields `None` from `split_once` and is skipped.

`find_map` returns the first genuine `filename` parameter, and empty values still map to `None`. The added test asserts `xfilename="evil"; filename="good.bin"` → `good.bin` and a lone `xfilename` → `None`, which directly exercises the fix. Reads safe to merge.

---

---

PR #2824, opened 2026-07-24
fix(hub-client): keep read_capped's "too large" error from surfacing as "invalid UTF-8"

phil-opp:
## Issue

`libraries/hub-client/src/index.rs::read_capped` bounds catalog-entry reads by over-reading exactly one byte past the cap and then checking the length:

```rust
let mut raw = String::new();
file.take(MAX_ENTRY_SIZE + 1)
    .read_to_string(&mut raw)                       // validates UTF-8 first
    .with_context(|| format!("failed to read `{}`", path.display()))?;
if raw.len() as u64 > MAX_ENTRY_SIZE {
    eyre::bail!("index file `{}` too large", path.display());
}
```

`read_to_string` validates UTF-8 across all `MAX_ENTRY_SIZE + 1` bytes **before** the size check runs. For an oversize entry whose byte at the cap lands in the middle of a multi-byte UTF-8 sequence — easy to hit with a corrupt or hostile 2 MB+ file — the read fails with `stream did not contain valid UTF-8`, and the intended, clearer `index file ... too large` message never fires. `read_capped` is on every catalog entry/metadata read.

## Fix

Read raw bytes with `read_to_end`, enforce the size cap, then convert to `String` — so the size guard wins for oversize input, and genuinely non-UTF-8 sub-cap files get a dedicated `is not valid UTF-8` context instead of a bare read error.

## Validation

- `cargo fmt -p dora-hub-client -- --check` — clean
- `cargo clippy -p dora-hub-client --all-targets -- -D warnings` — clean
- `cargo test -p dora-hub-client` — 43 passed, 0 failed. Adds two regression tests:
  - an oversize file filled with a 2-byte char (cap splits a sequence) → asserts `too large`
  - a small non-UTF-8 binary file → asserts `not valid UTF-8`

---

---

### DIFF
--- a/libraries/hub-client/src/index.rs
+++ b/libraries/hub-client/src/index.rs
@@ -439,14 +439,19 @@ fn read_capped(path: &Path) -> eyre::Result<String> {
     use std::io::Read as _;
     let file = std::fs::File::open(path)
         .with_context(|| format!("no index entry at `{}`", path.display()))?;
-    let mut raw = String::new();
+    // Read raw bytes and check the size cap *before* validating UTF-8. Reading
+    // straight into a `String` validates UTF-8 over the `MAX_ENTRY_SIZE + 1`
+    // bytes we deliberately over-read to detect oversize files; if that extra
+    // byte lands mid-way through a multi-byte sequence, the caller would see a
+    // confusing "invalid UTF-8" error instead of the intended "too large" one.
+    let mut raw = Vec::new();
     file.take(MAX_ENTRY_SIZE + 1)
-        .read_to_string(&mut raw)
+        .read_to_end(&mut raw)
         .with_context(|| format!("failed to read `{}`", path.display()))?;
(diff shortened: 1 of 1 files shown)


phil-opp:

Switching from `read_to_string` to `read_to_end` + size check + `String::from_utf8` correctly reorders the guards: the `take(MAX_ENTRY_SIZE + 1)` over-read is preserved, so any oversize file now hits the `too large` bail before UTF-8 is validated, and a genuinely non-UTF-8 sub-cap file gets the dedicated `is not valid UTF-8` context instead of a bare read error. Behavior for valid, in-bounds files is unchanged.

Both regression tests are meaningful: one writes an oversize file of 2-byte chars so the cap splits a sequence and asserts `too large`, the other writes a small non-UTF-8 file and asserts the UTF-8 error. Reads safe to merge.

---

---

PR #2825, opened 2026-07-24
fix(ros2-bridge): reject null single-element scalar/string fields on serialize

phil-opp:
## Issue

The Arrow→CDR serializers for single-element **scalar** and **string** ROS2 message fields validate the element count but never check the Arrow **null validity bit**:

- `libraries/extensions/ros2-bridge/arrow/src/serialize/primitive.rs` — `as_single_primitive` (`array.value(0)`) and the `Bool` arm (`array.value(0)`)
- `libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs` — the `String`/`WString` single-field paths (`string_array.value(0)`)

`PrimitiveArray::value(0)` / `BooleanArray::value(0)` / `GenericStringArray::value(0)` return the physical buffer slot **regardless of the validity bit**. So a single-element field column built with a null — e.g. `Int32Array::from(vec[None])`, or a nullable bool/string — was serialized to CDR as `0` / `false` / `""`, silently inventing a value the producing node never sent.

This is inconsistent with the two sibling conversion layers in this repo, which were explicitly hardened for exactly this case:
- `arrow-convert` rejects nulls (`if array.null_count() != 0 { bail!(...) }`).
- `mavlink2-bridge`'s `read_primitive` rejects `is_null(0)` with the comment that `value` "returns the slot but ignores the null bit, which silently invents a value; reject it."

## Fix

Add a shared `check_not_null` guard (alongside the existing `check_single_element`) and apply it after the length check in `as_single_primitive`, the `Bool` arm, and the `String`/`WString` single-element paths. ROS2/CDR has no null concept, so a null field is now a clear serialization error instead of a substituted default. A null-free column (`null_count == 0`) is unaffected, and missing fields still fall back to computed defaults as before.

**Scope:** this covers the single-element scalar/string field paths. Null handling for `sequence<T>` / fixed-array **element** columns is a broader change and is intentionally left out of this PR.

## Validation

- `cargo fmt -p dora-ros2-bridge-arrow -- --check` — clean
- `cargo clippy -p dora-ros2-bridge-arrow --all-targets -- -D warnings` — clean
- `cargo test -p dora-ros2-bridge-arrow` — 15 passed, 0 failed (existing CDR round-trip tests still green; adds message-level tests that a null `int32` and a null `string` are rejected, with non-null positive controls)

---

---

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs
@@ -147,12 +147,14 @@ impl TypedValue<'_> {
                     GenericString::String | GenericString::BoundedString(_) => {
                         let string = if let Some(string_array) = column.as_string_opt::<i32>() {
                             check_single_element(string_array.len(), "StringArray")?;
+                            check_not_null(string_array, "StringArray")?;
                             string_array.value(0)
                         } else {
                             let string_array = column
                                 .as_string_opt::<i64>()
                                 .ok_or_else(|| error("expected string array"))?;
                             check_single_element(string_array.len(), "LargeStringArray")?;
+                            check_not_null(string_array, "LargeStringArray")?;
                             string_array.value(0)
                         };
                         if let GenericString::BoundedString(max) = t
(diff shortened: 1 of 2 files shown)


phil-opp:

The premise holds: `PrimitiveArray::value(0)` / `BooleanArray::value(0)` / `GenericStringArray::value(0)` return the physical slot regardless of the validity bit, so a single-element field built with a null was being encoded to CDR as `0` / `false` / `""`. The new `check_not_null` guard is applied after the existing single-element length check in each affected path (`as_single_primitive`, the `Bool` arm, and the four `String`/`WString` single-field branches), which is the right order.

I also confirmed the stated scope is accurate: `as_single_primitive` and the `String`/`WString` single-field branches are only reached for single scalar/string fields — fixed arrays and sequences serialize through separate code — so this correctly leaves element-level null handling for a follow-up. Rejecting (rather than substituting a default) is consistent with the `arrow-convert` and `mavlink2-bridge` siblings.

The tests are meaningful: each asserts a non-null value serializes and the null variant is rejected with a null-mentioning error. Reads safe to merge.

---

---

PR #2829, opened 2026-07-25
fix(daemon): increment pending counter when propagating NodeFailed downstream

### LINKED ISSUE #2827: daemon: `pending_messages` counter underflows to `u64::MAX` when propagating `NodeFailed` downstream (missing `inc_pending`)
> **Note:** This issue was created by a **scheduled automated Claude code-review check**. A component was selected for review and the finding below was verified by reading the full increment/decrement invariant in the current source. It is machine-generated — please verify before acting.

## Summary

When an upstream node exits with a non-zero code, the daemon propagates a `NodeEvent::NodeFailed` to each downstream (subscriber) node. It enqueues that event into the subscriber's channel **without** the matching `inc_pending()` that every other delivery site performs. The per-node `pending_messages` counter (an `Arc<AtomicU64>`) is still decremented unconditionally by the `Listener` when it drains the event, so the counter ends up one lower than the true in-flight count — and if the subscriber's channel was already empty (counter `== 0`), the `fetch_sub(1)` **wraps to `u64::MAX`**. The bogus value is then reported to the coordinator (metrics / `dora node list`) until the node re-subscribes.

## Where

`binaries/daemon/src/lib.rs`, NodeFailed propagation (lines ~4979-4991):

```rust
for (recv_id, affected_ids) in affected_by_receiver {
    if let Some(channel) = dataflow.subscribe_channels.get(&recv_id) {
        let _ = send_with_timestamp(                 // <-- no inc_pending, ignores Ok(true)
            channel,
            NodeEvent::NodeFailed {
                affected_input_ids: affected_ids,
                error: error_msg.clone(),
                source_node_id: node_id.clone(),
            },
            &self.clock,
        );
    }
}
```

The invariant everywhere else in the daemon: each `send_with_timestamp(channel, …)` that returns `Ok(true)` is paired with `dataflow.inc_pending(receiver_id)`. Examples in the same file: lib.rs:4507, 4559, 4660, 4901, 4934, 5109, 5142, 5381, 5480, 5565, 5593 (plus the Reload/Stop/AllInputsClosed sends and the `fetch_add` sites in `running_dataflow.rs`). This `NodeFailed` send is the **only** delivery to a `subscribe_channels` sender that omits it.

Decrement side — unconditional for every drained event, `NodeFailed` included — `binaries/daemon/src/node_communication/mod.rs:208` (also 310, 440):

```rust
while let Ok(event) = events.try_recv() {
    if let Some(counter) = &self.pending_counter {
        counter.fetch_sub(1, Ordering::Relaxed);     // <-- decrements NodeFailed too
    }
    self.queue.push_back(event);
}
```

Counter definition and increment:
- `pending_messages: HashMap<NodeId, Arc<AtomicU64>>` — `binaries/daemon/src/running_dataflow.rs:199`
- `inc_pending` does `fetch_add(1)` — `running_dataflow.rs:631-635`

## Failure scenario

Dataflow `A → B`, where `B` consumes `A`'s output and is written to survive upstream failure (handles `NodeFailed`).

1. `A` exits non-zero with restart policy `Never`, so the handler takes the `else` branch at lib.rs:4959.
2. At that instant `B` has drained its channel, so its `pending_messages` counter is `0`.
3. The daemon enqueues `NodeFailed` for `B` (`send_with_timestamp` returns `Ok(true)`) **without** calling `inc_pending`.
4. `B`'s `Listener` pulls the event and runs `counter.fetch_sub(1)`: `0 → 18446744073709551615`.
5. The next metrics push reports `B.pending_messages = 18446744073709551615` to the coordinator, so `dora node list` / doctor shows ~1.8×10¹⁹ pending messages for `B`. The bogus value persists until `B` re-subscribes (the counter is replaced at lib.rs:3664).

Even when `B`'s counter is `> 0` at drain time, the missing increment leaves the reported count permanently one too low; the empty-channel case is just where it becomes the dramatic `u64::MAX` symptom.

## Impact

Corrupted reporting metric via integer underflow — not a crash or data mis-routing, so severity is moderate. But it is a concrete `AtomicU64` underflow producing definitively wrong output, on the downstream-failure-propagation path with an idle receiver, which the existing tests (around lib.rs:7222, covering only `ParamUpdate` / `ParamDeleted` delivery) do not exercise.

## Suggested fix

Mirror the other call sites — increment on successful enqueue:

```rust
if send_with_timestamp(channel, NodeEvent::NodeFailed { … }, &self.clock) == Ok(true) {
    dataflow.inc_pending(&recv_id);
}
```

(reborrowing `dataflow` mutably as the surrounding code requires). A regression test asserting the receiver's `pending_messages` counter stays consistent after a `NodeFailed` propagation to an idle node would lock this in.

phil-opp:
## Summary

Fixes #2827.

When an upstream node exits with a non-zero code, the daemon propagates a `NodeEvent::NodeFailed` to each downstream (subscriber) node. It enqueued that event into the subscriber's channel **without** the matching `inc_pending()` that every other delivery site performs. The per-node `pending_messages` counter (an `Arc<AtomicU64>`) is still decremented unconditionally by the `Listener` when it drains the event, so the counter ended up one lower than the true in-flight count — and if the subscriber's channel was already empty (counter `== 0`), the `fetch_sub(1)` **wrapped to `u64::MAX`**. The bogus value was then reported to the coordinator (metrics / `dora node list`) until the node re-subscribed.

## Change

- Extract the `NodeFailed` propagation loop out of the `SpawnedNodeResult` handler in `lib.rs` into a new `RunningDataflow::propagate_node_failed(failed_node, error_msg, clock)` method.
- The method pairs each successful enqueue (`send_with_timestamp(...).ok() == Some(true)`) with `inc_pending(&recv_id)`, mirroring every other delivery site in the daemon.
- Add a regression test (`propagate_node_failed_keeps_idle_receiver_counter_consistent`) that drives a `source → sink` mapping with an idle receiver, propagates a `NodeFailed`, then drains the event the way the `Listener` does (one unconditional `fetch_sub`) and asserts the counter returns to `0` rather than underflowing to `u64::MAX`.

## Testing

- `cargo test -p dora-daemon` — all 113 tests pass, including the new one.
- `cargo fmt -p dora-daemon -- --check` and `cargo clippy -p dora-daemon -- -D warnings` are clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CQi7Jq6ewwkjjskR7nmNgK)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4963,32 +4963,7 @@ impl Daemon {
                     if let Err(e) = &node_result
                         && let Some(dataflow) = self.running.get(&dataflow_id)
                     {
-                        let error_msg = e.to_string();
-                        let mut affected_by_receiver: BTreeMap<NodeId, Vec<DataId>> =
-                            BTreeMap::new();
-                        for (output_id, receivers) in &dataflow.mappings {
-                            if output_id.0 == node_id {
-                                for (recv_id, input_id) in receivers {
-                                    affected_by_receiver
-                                        .entry(recv_id.clone())
-                                        .or_default()
-                                        .push(input_id.clone());
-                                }
-                            }
-                        }
-                        for (recv_id, affected_ids) in affected_by_receiver {
-                            if let Some(channel) = dataflow.subscribe_channels.get(&recv_id) {
-                                let _ = send_with_timestamp(
-                                    channel,
-                                    NodeEvent::NodeFailed {
-                                        affected_input_ids: affected_ids,
-                                        error: error_msg.clone(),
-                                        source_node_id: node_id.clone(),
-                                    },
-                                    &self.clock,
-                                );
-                            }
-                        }
+                        dataflow.propagate_node_failed(&node_id, &e.to_string(), &self.clock);
                     }
 
                     let exit_clean = node_result.is_ok();
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -634,6 +634,47 @@ impl RunningDataflow {
         }
     }
 
+    /// Propagate a `NodeFailed` event to every downstream subscriber of the
+    /// failed node's outputs.
+    ///
+    /// Each successful enqueue is paired with `inc_pending`, mirroring every
+    /// other delivery site: the `Listener` unconditionally decrements a node's
+    /// `pending_messages` counter for every event it drains (`NodeFailed`
+    /// included), so an enqueue without the matching increment would leave the
+    /// reported count one too low — and underflow it to `u64::MAX` when the
+    /// receiver's channel was already empty (dora-rs/dora#2827).
+    pub(crate) fn propagate_node_failed(&self, failed_node: &NodeId, error_msg: &str, clock: &HLC) {
+        let mut affected_by_receiver: BTreeMap<NodeId, Vec<DataId>> = BTreeMap::new();
+        for (output_id, receivers) in &self.mappings {
+            if output_id.0 == *failed_node {
+                for (recv_id, input_id) in receivers {
+                    affected_by_receiver
+                        .entry(recv_id.clone())
+                        .or_default()
+                        .push(input_id.clone());
+                }
+            }
+        }
+        for (recv_id, affected_ids) in affected_by_receiver {
+            if let Some(channel) = self.subscribe_channels.get(&recv_id) {
+                let delivered = send_with_timestamp(
+                    channel,
+                    NodeEvent::NodeFailed {
+                        affected_input_ids: affected_ids,
+                        error: error_msg.to_string(),
+                        source_node_id: failed_node.clone(),
+                    },
+                    clock,
+                )
+                .ok()
+                    == Some(true);
+                if delivered {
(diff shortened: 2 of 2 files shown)


phil-opp:

I confirmed the bug and the fix against the surrounding code:

- `send_with_timestamp` returns `Ok(true)` when the event is actually delivered, `Ok(false)` when dropped (channel full), and `Err` when closed (`event_types.rs`). The new `.ok() == Some(true)` gate therefore increments the counter only when an event truly enters the channel.
- The node-facing listener decrements `pending_messages` unconditionally for *every* drained event — including `NodeFailed` (`node_communication/mod.rs`, all three drain sites). So an enqueue without the matching `inc_pending` does leave the count one low, and `fetch_sub(1)` on a `0` counter wraps to `u64::MAX`, exactly as described.
- The extracted `propagate_node_failed` is behavior-preserving relative to the old inline loop and mirrors the existing `inc_pending` pattern used at the `NodeEvent::Stop` delivery sites (`running_dataflow.rs`).

The regression test is meaningful: it drives the real propagation, asserts the counter reaches `1`, then drains exactly as the listener does and asserts it returns to `0` rather than underflowing. Reads safe to merge.

---

---

PR #2834, opened 2026-07-25
fix(memory-pool): include OS error details during orphan cleanup

### LINKED ISSUE #2831: memory-pool: Include underlying OS error details during orphan cleanup logging
## Summary

While reviewing the memory-pool cleanup path, I noticed that cleanup failures appear to log generic messages without preserving the underlying operating system error.

Including the original OS error would make diagnosing permission or filesystem-related cleanup failures much easier.

## Suggested Improvement

When cleanup fails, include the original error returned by the operating system in the debug log instead of only emitting a generic message.

This is a small diagnostics improvement that should not affect runtime behavior.

I'd be happy to submit a PR if this sounds useful.

SaitejaKommi:
## Summary

Improve orphan cleanup diagnostics by surfacing `/dev/shm` directory read failures and providing additional file path context when shared memory cleanup encounters OS errors.

Fixes #2831

## Root Cause

`MemoryPoolManager::cleanup_orphans` previously used `if let Ok(entries) = std::fs::read_dir("/dev/shm")`. If reading the shared memory directory failed (e.g. due to permission or filesystem errors), the failure was silently ignored.

Additionally, when removing orphaned shared memory files failed, the debug log only included the shared memory object name and the OS error, making it harder to identify the affected file.

## Solution

- Replaced `if let Ok(...)` with `match std::fs::read_dir("/dev/shm")` to explicitly handle and log directory enumeration failures.
- Enhanced orphan cleanup logging to include the full file path (`entry.path().display()`) together with the shared memory object name and the underlying OS error.
- Added a unit test (`cleanup_orphans_runs_without_panic`) to verify cleanup executes safely without panicking.

## Testing

- ✅ `cargo fmt --all -- --check`
- ✅ `cargo check -p dora-memory-pool`
- ✅ `cargo test -p dora-memory-pool` (10 tests passed)

## Scope

This PR is intentionally limited to improving diagnostics in `libraries/extensions/memory-pool/src/lib.rs`. No functional behavior, public APIs, or cleanup logic were modified.

### DIFF
--- a/libraries/extensions/memory-pool/src/lib.rs
+++ b/libraries/extensions/memory-pool/src/lib.rs
@@ -225,16 +225,27 @@ impl MemoryPoolManager {
         #[cfg(target_os = "linux")]
         {
             let prefix = format!("dora_pool_{}_", dataflow_id);
-            if let Ok(entries) = std::fs::read_dir("/dev/shm") {
-                for entry in entries.flatten() {
-                    let name = entry.file_name();
-                    let name = name.to_string_lossy();
-                    if name.starts_with(&prefix)
-                        && let Err(err) = std::fs::remove_file(entry.path())
-                        && err.kind() != std::io::ErrorKind::NotFound
-                    {
-                        tracing::debug!("Orphan sweep: could not unlink {}: {}", name, err);
+            match std::fs::read_dir("/dev/shm") {
+                Ok(entries) => {
+                    for entry in entries.flatten() {
+                        let name = entry.file_name();
+                        let name = name.to_string_lossy();
+                        if name.starts_with(&prefix)
+                            && let Err(err) = std::fs::remove_file(entry.path())
+                            && err.kind() != std::io::ErrorKind::NotFound
+                        {
+                            tracing::debug!(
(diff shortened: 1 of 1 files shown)


SaitejaKommi:
Hi @phil-opp , ,could you please review this PR when you have a moment? All CI checks have passed.

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop.

No issues found. This is a diagnostic-only change to `MemoryPoolManager::cleanup_orphans` in `libraries/extensions/memory-pool/src/lib.rs`: the `read_dir("/dev/shm")` failure is now surfaced via a `match` (with `ErrorKind::NotFound` still silently ignored, so the non-Linux / no-shm case stays quiet), and the per-file unlink warning now includes the full path. The sweep behavior itself is unchanged and the `#[cfg(target_os = "linux")]` split is preserved.

One minor note on the added test: `cleanup_orphans_runs_without_panic` only asserts that the call doesn't panic — it doesn't exercise the new error-logging branch or verify that orphaned segments are actually removed. That's acceptable for a diagnostics-only change, but the test doesn't cover the code path this PR adds.

---

PR #2837, opened 2026-07-25
perf(daemon): skip the metadata deep-clone for pure-remote outputs

phil-opp:
## Issue

`send_output_to_local_receivers` unconditionally built `Arc::new(metadata.clone())` before fanning out to local receivers. That `Arc<Metadata>` is only consumed *inside* the fan-out loop, so when an output's subscribers all live on **other** daemons (empty `local_receivers`), the deep clone was built and immediately dropped on every message. `Metadata` owns a `BTreeMap<String, Parameter>` whose values include `String`/`Vec`, so this is a genuine heap-allocating deep clone, not a cheap copy. `send_out` still calls this function in the pure-remote case (with `need_data_bytes = true`) to move the payload bytes out.

## Fix

Build the `Arc` lazily via `get_or_insert_with`, so the clone happens once on the first actual delivery and is skipped entirely when there is no local receiver — or when every receiver is over its control-headroom capacity and is skipped anyway. The payload path was already optimal here (`Arc::try_unwrap` moves the bytes out without copying); this closes the same gap for metadata.

No new `.unwrap()` / `.expect()` is introduced (relevant because of the production unwrap-budget ratchet).

## Validation

- `cargo clippy -p dora-daemon --lib -- -D warnings` clean.
- `cargo test -p dora-daemon --lib` → 112 passed.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -5403,10 +5403,15 @@ async fn send_output_to_local_receivers(
     let empty_set = BTreeSet::new();
     let output_id = OutputId(node_id, output_id);
     let local_receivers = dataflow.mappings.get(&output_id).unwrap_or(&empty_set);
-    // Wrap in Arc once; fan-out clones are O(1) atomic ref bumps instead of O(payload_size) memcpy
-    let metadata = Arc::new(metadata.clone());
     let data = data.map(Arc::new);
     let mut closed = Vec::new();
+    // Clone the metadata into an `Arc` lazily, on the first actual delivery.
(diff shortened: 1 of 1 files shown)


phil-opp:

Deferring the `Arc<Metadata>` construction via `metadata_arc.get_or_insert_with(|| Arc::new(metadata.clone()))` is behavior-equivalent for every delivered receiver (built once on first delivery, shared thereafter) and correctly skips the deep clone when `local_receivers` is empty or all receivers are capacity-skipped, since no delivery occurs in those cases. `metadata` stays owned by the function (no use-after-move). No issues found.

---

PR #2842, opened 2026-07-26
fix(ros2-bridge): preserve DDS QoS durability and stop panics on bad QoS durations

### LINKED ISSUE #2841: ros2-bridge (Python): DDS QoS regression in #2790 — Transient/Persistent durability silently downgraded to Volatile, and NaN/negative QoS durations now panic
## Summary

PR #2790 (`feat(ros2-bridge): add native rmw_zenoh_cpp transport`, merged as 0532878) re-routed the Python bridge's **DDS** QoS conversion through a new backend-neutral `Ros2Qos` type. In doing so it introduced two regressions on the pre-existing, feature-independent DDS path in `libraries/extensions/ros2-bridge/python/src/qos.rs`. Both are unrelated to the new Zenoh transport — they affect every DDS node created from Python.

The relevant change is in `impl From<Ros2QosPolicies> for rustdds::QosPolicies`, which previously built the rustdds QoS directly and now does:

```rust
impl From<Ros2QosPolicies> for rustdds::QosPolicies {
    fn from(value: Ros2QosPolicies) -> Self {
        ::dora_ros2_bridge::transport::dds::to_rustdds_qos(&value.into()) // value.into(): Ros2QosPolicies -> Ros2Qos
    }
}
```

---

## Finding 1 — `Transient` / `Persistent` durability silently downgraded to `Volatile` (regression)

`python/src/qos.rs` (`impl From<Ros2QosPolicies> for Ros2Qos`):

```rust
durability: match value.durability {
    Ros2Durability::TransientLocal => Durability::TransientLocal,
    _ => Durability::Volatile,   // Transient AND Persistent collapse here
},
```

The neutral `Durability` enum has only two variants (`libraries/extensions/ros2-bridge/src/transport/types.rs:19-24`):

```rust
pub enum Durability { Volatile, TransientLocal }
```

and `to_rustdds_qos` maps them 1:1 (`src/transport/dds.rs:107-109`), so `Transient`/`Persistent` can never reach rustdds through this path.

**Before #2790**, the conversion used `.durability(value.durability.into())`, i.e. `From<Ros2Durability> for policy::Durability`, which faithfully mapped all four variants to the corresponding `rustdds::policy::Durability`:

```rust
Ros2Durability::Volatile        => policy::Durability::Volatile,
Ros2Durability::TransientLocal  => policy::Durability::TransientLocal,
Ros2Durability::Transient       => policy::Durability::Transient,
Ros2Durability::Persistent      => policy::Durability::Persistent,
```

That `From` impl still exists but is now **dead code** on the runtime path.

**Risk / affected paths.** The Python API still exposes `Ros2Durability.Transient` and `Ros2Durability.Persistent` (`qos.rs`), and these policies flow into `create_topic` / publisher / subscription / service / action QoS. A DDS node created with `Ros2QosPolicies(durability=Ros2Durability.Transient)` now advertises `Volatile` on the wire. A remote DDS participant requesting `Transient` durability sees an incompatible-QoS mismatch and fails to match — whereas it matched before #2790. The downgrade is silent (no error, no warning), which makes it hard to diagnose.

Severity: medium (silent behavior change on a public API), mitigated by the fact that `Transient`/`Persistent` are rarely used in real ROS2 setups (they require a DDS persistence service). The commonly used `Volatile` and `TransientLocal` are still mapped correctly.

## Finding 2 — NaN / negative QoS durations now panic instead of being accepted

`python/src/qos.rs` (same `impl From<Ros2QosPolicies> for Ros2Qos`):

```rust
let lease_duration = (!value.lease_duration.is_infinite())
    .then(|| std::time::Duration::from_secs_f64(value.lease_duration));   // qos.rs:62
...
max_blocking_time: std::time::Duration::from_secs_f64(value.max_blocking_time), // qos.rs:77 (reliable path)
```

`std::time::Duration::from_secs_f64` **panics** on a negative, `NaN`, or overflowing value. The pre-#2790 code used `rustdds::Duration::from_frac_seconds(...)`, which did not panic.

- `lease_duration`: guarded only by `!is_infinite()`. `-inf`/`+inf` are safely skipped, but a **finite negative** (e.g. `-1.0`) or **`NaN`** reaches `from_secs_f64` and panics.
- `max_blocking_time`: reached only when `reliable=True`; a negative/NaN value panics.

**Risk / affected paths.** `Ros2QosPolicies(lease_duration=float('nan'))` or `Ros2QosPolicies(reliable=True, max_blocking_time=-1.0)` now panics inside a `#[pymethods]` call (surfaced to Python as an exception / abort) instead of being accepted. Edge case — the defaults (`f64::INFINITY` / `0.0`) are safe — hence low severity, but it is a new crash path on the DDS QoS builder.

---

## Suggested fix

Preserve DDS QoS fidelity when lowering from the Python type, rather than forcing DDS through the two-variant neutral `Ros2Qos`:

- **Durability**: keep the four-variant mapping for the DDS path. Either give the neutral `Durability` enum `Transient`/`Persistent` variants (and map them in `to_rustdds_qos`), or bypass the neutral type for DDS and build rustdds QoS directly via the still-present `From<Ros2Durability> for policy::Durability`.
- **Durations**: use a non-panicking conversion (e.g. `Duration::try_from_secs_f64`, available in recent Rust, or clamp/validate negative/NaN to a sane default or a typed error) in place of `from_secs_f64`, matching the previous non-panicking `from_frac_seconds` behavior.

Both findings share one root cause (the new lossy/ungrarded neutral-QoS funnel) and one fix location (`python/src/qos.rs`), so they're reported together.

---

_This issue was created by an automated, scheduled Claude code-review check (not a human-authored report). The two findings above were verified against the actual `git diff` of PR #2790 and the surrounding code; please review before acting._

phil-opp:
## Summary

Fixes #2841.

PR #2790 re-routed the Python bridge's **DDS** QoS conversion through the new backend-neutral `Ros2Qos` type (`libraries/extensions/ros2-bridge/python/src/qos.rs`). That introduced two regressions on the feature-independent DDS path — both affecting every DDS node created from Python:

1. **Durability downgrade.** The neutral `Durability` enum has only two variants (`Volatile`, `TransientLocal`), so the conversion collapsed `Transient` and `Persistent` to `Volatile`. A DDS node created with `Ros2Durability.Transient`/`.Persistent` silently advertised `Volatile` on the wire, causing incompatible-QoS mismatches with remote participants that previously matched.
2. **Panic on invalid QoS durations.** The neutral conversion used `std::time::Duration::from_secs_f64`, which **panics** on a negative, `NaN`, or overflowing value. `Ros2QosPolicies(lease_duration=float('nan'))` or `Ros2QosPolicies(reliable=True, max_blocking_time=-1.0)` panicked inside a `#[pymethods]` call instead of being accepted (the pre-#2790 `rustdds` path did not panic).

## Changes

- `From<Ros2QosPolicies> for rustdds::QosPolicies`: build reliability/history/liveliness through the neutral path, then restore the faithful four-variant DDS durability mapping via the still-present `From<Ros2Durability> for policy::Durability`. `Transient`/`Persistent` are DDS-only concepts (the Zenoh transport only supports two durability levels), so the fix is scoped to the DDS conversion and leaves the neutral type and the Zenoh path untouched.
- `From<Ros2QosPolicies> for Ros2Qos`: convert QoS durations with the non-panicking `Duration::try_from_secs_f64`, clamping an invalid value to zero.
- Added regression tests: all four durability variants survive the DDS conversion, and NaN/negative/overflow duration inputs no longer panic.

## Testing

- `cargo test -p dora-ros2-bridge-python --lib qos::` — both new tests pass.
- `cargo clippy -p dora-ros2-bridge-python -- -D warnings` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WSS8xjgq1q9ujn4mLPqGYK)_

### DIFF
--- a/libraries/extensions/ros2-bridge/python/src/qos.rs
+++ b/libraries/extensions/ros2-bridge/python/src/qos.rs
@@ -52,14 +52,37 @@ impl Ros2QosPolicies {
 
 impl From<Ros2QosPolicies> for rustdds::QosPolicies {
     fn from(value: Ros2QosPolicies) -> Self {
-        ::dora_ros2_bridge::transport::dds::to_rustdds_qos(&value.into())
+        // `Transient`/`Persistent` durability exists only on the DDS path: the
+        // backend-neutral `Ros2Qos` collapses both to `Volatile`, so routing DDS
+        // QoS through it would silently downgrade them on the wire. Build the
+        // shared policies through the neutral path (reliability/history/
+        // liveliness), then restore the faithful four-variant DDS durability.
+        let durability: policy::Durability = value.durability.into();
+        let neutral = ::dora_ros2_bridge::transport::dds::to_rustdds_qos(&value.into());
+        let mut builder = rustdds::QosPolicyBuilder::new().durability(durability);
+        if let Some(reliability) = neutral.reliability() {
+            builder = builder.reliability(reliability);
+        }
+        if let Some(history) = neutral.history() {
+            builder = builder.history(history);
+        }
+        if let Some(liveliness) = neutral.liveliness() {
+            builder = builder.liveliness(liveliness);
+        }
+        builder.build()
     }
 }
 
 impl From<Ros2QosPolicies> for Ros2Qos {
     fn from(value: Ros2QosPolicies) -> Self {
-        let lease_duration = (!value.lease_duration.is_infinite())
-            .then(|| std::time::Duration::from_secs_f64(value.lease_duration));
+        // `Duration::from_secs_f64` panics on a negative, NaN, or overflowing
+        // value; the pre-existing `rustdds` path did not. Convert without
+        // panicking, treating an invalid duration as zero.
+        let to_duration = |secs: f64| {
+            std::time::Duration::try_from_secs_f64(secs).unwrap_or(std::time::Duration::ZERO)
+        };
+        let lease_duration =
+            (!value.lease_duration.is_infinite()).then(|| to_duration(value.lease_duration));
         Self {
             durability: match value.durability {
                 Ros2Durability::TransientLocal => Durability::TransientLocal,
@@ -74,7 +97,7 @@ impl From<Ros2QosPolicies> for Ros2Qos {
             },
             reliability: if value.reliable {
                 Reliability::Reliable {
-                    max_blocking_time: std::time::Duration::from_secs_f64(value.max_blocking_time),
+                    max_blocking_time: to_duration(value.max_blocking_time),
                 }
             } else {
                 Reliability::BestEffort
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff against `libraries/extensions/ros2-bridge/src/transport/dds.rs`. The fix is correct and well-targeted.

- **Durability preservation:** `to_rustdds_qos` sets exactly four policies (reliability, durability, history, liveliness). The rebuilt `From<Ros2QosPolicies> for rustdds::QosPolicies` copies reliability/history/liveliness from the neutral path and overrides durability with the faithful four-variant mapping (`value.durability.into()`), so nothing the neutral path set is dropped, and `Transient`/`Persistent` are no longer collapsed to `Volatile` on the wire.
- **No move hazard:** `value.durability` is `Copy`, so reading it before `value.into()` (which resolves unambiguously to `Ros2Qos` via the `to_rustdds_qos(&..)` parameter type) is fine.
- **Panic removal:** `try_from_secs_f64(..).unwrap_or(ZERO)` correctly replaces `from_secs_f64`, which panics on NaN/negative/overflow; both the `lease_duration` and `max_blocking_time` paths are covered.

Both tests are genuine: `dds_durability_preserves_all_variants` asserts all four variants survive, and `invalid_qos_durations_do_not_panic` exercises NaN/−1/`f64::MAX` on both duration paths. No issues found.

---

PR #2843, opened 2026-07-26
fix: avoid bail macros in expression position

### LINKED ISSUE #2835: Build fails on newer Rust due to semicolon_in_expressions_from_macros
**Describe the bug**

Building Dora from source fails with nightly Rust toolchain because some `eyre::bail!` / `bail!` macro invocations are used in expression position. This triggers `semicolon_in_expressions_from_macros`, which is denied as part of `future_incompatible`, causing `dora-core` to fail compilation.

The affected call sites include `bail!` invocations at the end of blocks and directly in `match` arms.

**To Reproduce**

1. Clone the repository:
   ```bash
   git clone https://github.com/dora-rs/dora.git
   cd dora
   ```

2. Check `dora-core` with a newer Rust toolchain:
   ```bash
   cargo +nightly check -p dora-core
   ```

3. The build fails with errors like:

   ```text
   error: trailing semicolon in macro used in expression position
   --> libraries\core\src\build\git.rs:102:13
   |
   102 | /             eyre::bail!(
   103 | |                 "the build directory is still in use by the following
   104 | |                     dataflows, please stop them before rebuilding: {}",
   105 | |                 using.iter().join(", ")
   106 | |             )
   | |_____________^
   |
   = note: macro invocations at the end of a block are treated as expressions
   = note: to ignore the value produced by the macro, add a semicolon after the invocation of bail
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #79813 https://github.com/rust-lang/rust/issues/79813
   = note: #[deny(semicolon_in_expressions_from_macros)] (part of #[deny(future_incompatible)]) on by default
   = note: this error originates in the macro eyre::bail
   ```

   ```text
   error: trailing semicolon in macro used in expression position
   --> libraries\core\src\build\git.rs:287:25
   |
   287 |                         bail!(err)
   |                         ^^^^^^^^^^
   |
   = note: macro invocations at the end of a block are treated as expressions
   = note: to ignore the value produced by the macro, add a semicolon after the invocation of bail
   = note: #[deny(semicolon_in_expressions_from_macros)] (part of #[deny(future_incompatible)]) on by default
   = note: this error originates in the macro bail
   ```

   Similar errors also occur in other locations, for example:

   ```text
   libraries\core\src\build\git.rs:339:25
   libraries\core\src\build\git.rs:373:25
   ```

**Expected behavior**

`dora-core` should compile successfully on nightly Rust toolchain

The affected `bail!` calls could be moved from expression position to statement position. For example:

```rust
Err(err) => {
    bail!(err);
}
```

For `bail!` invocations at the end of a block, adding a trailing semicolon should also avoid using the macro invocation as the block expression.

This should not change runtime behavior because `bail!` returns early.

**Screenshots or Video**

Not applicable.

**Environments :**

- System info: Windows 11 25H2
- Rust version:
  ```text
  rustc 1.99.0-nightly
  ```
- Cargo version:
  ```text
  cargo 1.99.0-nightly
  ```
- Dora version: source build from current repository checkout
- `dora --version`: unavailable because the source build fails

**Additional context**

The issue seems related to `eyre::bail!` expanding with a trailing semicolon while being used in expression position. I would be happy to open a PR that updates the affected call sites to use `bail!` in statement position if this approach is acceptable.

Lioyae:
## Summary

Fixes `bail!` and `eyre::bail!` macro invocations that are used in expression position and trigger `semicolon_in_expressions_from_macros` on recent nightly Rust.

This lint is denied through `future_incompatible`, so recent nightly toolchains fail when checking or building `dora-cli`.

Closes #2835

## Background

This follows up on the reopened issue comment: https://github.com/dora-rs/dora/issues/2835#issuecomment-5082795113

There are two separate problems involved:

1. A rustc nightly ICE can occur while compiling `tokio` at higher release optimization levels.
2. Once the build proceeds past `tokio`, Dora still hits the original `bail!` macro compilation error.

This PR only addresses the Dora-specific `bail!` / `eyre::bail!` error. The `tokio` ICE appears to be a separate rustc nightly issue and is intentionally left out of scope.

## Validation

Tested locally on Windows MSVC:

- `cargo +nightly-2026-07-13 check -p dora-cli` on `main`: passes
- `cargo +nightly check -p dora-cli` on `main` with `rustc 1.99.0-nightly (da86f4d07 2026-07-24)`: fails with `semicolon_in_expressions_from_macros`
- `cargo +nightly check -p dora-cli` on this branch: passes
- `cargo +nightly check -p dora-cli --release` on this branch: passes
- `cargo +stable check -p dora-cli`: passes
- `cargo fmt --all -- --check`: passes
- `git diff --check`: passes

Additional note:

- `cargo +nightly build --release -p dora-cli` with `rustc 1.99.0-nightly (da86f4d07 2026-07-24)` still triggers a rustc ICE while compiling `tokio v1.53.1`.
- The same full release build passes with `nightly-2026-07-13`.
- This supports treating the `tokio` ICE as unrelated to this PR.

### DIFF
--- a/apis/rust/node/src/daemon_connection/interactive.rs
+++ b/apis/rust/node/src/daemon_connection/interactive.rs
@@ -83,7 +83,7 @@ impl InteractiveEvents {
             | DaemonRequest::ReadPinnedMemory { .. }
             | DaemonRequest::FreePinnedMemory { .. } => DaemonReply::Result(Ok(())),
             DaemonRequest::NodeConfig { .. } => {
-                eyre::bail!("unexpected NodeConfig in interactive mode")
+                eyre::bail!("unexpected NodeConfig in interactive mode");
             }
         };
         Ok(reply)
--- a/apis/rust/node/src/daemon_connection/json_to_arrow.rs
+++ b/apis/rust/node/src/daemon_connection/json_to_arrow.rs
@@ -13,7 +13,9 @@ pub fn read_json_bytes_as_arrow(data: &[u8]) -> eyre::Result<ArrayData> {
             // try again with quoting the input to treat it as a string
             match arrow_json::reader::infer_json_schema(wrapped_quoted(data), None) {
                 Ok((schema, _)) => read_from_json_with_schema(wrapped_quoted(data), schema),
-                Err(err) => eyre::bail!("failed to infer JSON schema: {err}"),
+                Err(err) => {
+                    eyre::bail!("failed to infer JSON schema: {err}");
+                }
             }
         }
     }
--- a/apis/rust/node/src/daemon_connection/mod.rs
+++ b/apis/rust/node/src/daemon_connection/mod.rs
@@ -53,7 +53,9 @@ impl DaemonChannel {
             DaemonReply::Result(result) => result
                 .map_err(|e| eyre!(e))
                 .wrap_err("failed to register node with dora-daemon")?,
-            other => bail!("unexpected register reply: {other:?}"),
+            other => {
+                bail!("unexpected register reply: {other:?}");
+            }
         }
         Ok(())
     }
--- a/apis/rust/node/src/daemon_connection/node_integration_testing.rs
+++ b/apis/rust/node/src/daemon_connection/node_integration_testing.rs
@@ -136,7 +136,7 @@ impl IntegrationTestingEvents {
             | DaemonRequest::ReadPinnedMemory { .. }
             | DaemonRequest::FreePinnedMemory { .. } => DaemonReply::Result(Ok(())),
             DaemonRequest::NodeConfig { .. } => {
-                eyre::bail!("unexpected NodeConfig in interactive mode")
+                eyre::bail!("unexpected NodeConfig in interactive mode");
             }
         };
         Ok(reply)
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -623,9 +623,11 @@ impl EventStream {
         match reply {
             DaemonReply::Result(Ok(())) => {}
             DaemonReply::Result(Err(err)) => {
-                eyre::bail!("subscribe failed: {err}")
+                eyre::bail!("subscribe failed: {err}");
             }
-            other => eyre::bail!("unexpected subscribe reply: {other:?}"),
+            other => {
+                eyre::bail!("unexpected subscribe reply: {other:?}");
+            }
         }
 
         close_channel.register(dataflow_id, node_id.clone(), clock.new_timestamp())?;
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -725,9 +725,11 @@ impl EventStream {
         match reply {
             DaemonReply::Result(Ok(())) => {}
             DaemonReply::Result(Err(err)) => {
-                eyre::bail!("subscribe failed: {err}")
+                eyre::bail!("subscribe failed: {err}");
             }
-            other => eyre::bail!("unexpected subscribe reply: {other:?}"),
+            other => {
+                eyre::bail!("unexpected subscribe reply: {other:?}");
+            }
         }
 
         close_channel.register(dataflow_id, node_id.clone(), clock.new_timestamp())?;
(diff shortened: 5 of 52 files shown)


Lioyae:
Hi @phil-opp, could you please review this PR when you have a moment? All CI checks have passed, and the linked issue is #2835.

phil-opp:
I went through all 49 files in this change. Every edit is the same mechanical transformation — adding a trailing `;` to a tail-position `bail!` / `eyre::bail!`, or rewriting a `arm => bail!(...)` match arm as `arm => { bail!(...); }`. Because `bail!` expands to a diverging `return Err(...)`, wrapping it in a `{ ...; }` block still diverges and coerces correctly in value positions (e.g. the `Ok(match ...)` in `binaries/cli/src/common.rs`), so no match-arm type mismatch is introduced and there is no control-flow or behavior change on any path.

No issues found.

---
🤖 This is a fully automated review by Claude (Claude Code) — no human has vetted these comments.

---

phil-opp:
Thanks for the PR! Let's try to fix this upstream first, I opened https://github.com/eyre-rs/eyre/pull/294 for that. The rustc lint was downgraded to a warning for now (see https://github.com/rust-lang/rust/issues/79813#issuecomment-5097085671), so we have a bit of time to wait for the `eyre` fix and release.

(If it isn't fixed there in time, we can proceed with this PR.)

Lioyae:
@phil-opp Thanks for taking a look and for opening the upstream "eyre" PR!

That makes sense to me. I'll leave this PR open as a fallback for now and wait for the "eyre" fix/release. If the upstream fix does not land in time, or if Dora still needs a local workaround afterwards, I'm happy to update this PR accordingly.

Lioyae:
Hi @phil-opp, thanks for the smaller-scoped #3086 — it makes sense as the minimal fix to unblock the build.

I did a nightly build on the latest `main` and counted the remaining `semicolon_in_expressions_from_non_local_macros` diagnostics: 129 sites across dora's crates (cli 57, core 20, coordinator 17, node-api 17, daemon 10, hub-client 4, runtime 4). One thing I noticed: #3086 covers 18 of the 20 `dora-core` sites, but two remain outside its 4 files — `libraries/core/src/build/mod.rs:277` and `libraries/core/src/topics.rs:510`. These are warnings today — the lint was recently downgraded from deny — but rust-lang/rust#159218 proposes making it a hard error, so these sites would fail the nightly build once it lands.

If it's useful, I can scope this PR down to exactly the sites #3086 doesn't cover, so the two would be fully complementary and mergeable in any order — happy to push that anytime.

phil-opp:

I re-reviewed the latest commits, which rescope the change to the sites not covered by #3086 and add the remaining `bail!` sites. Every edit is still the same mechanical transformation — appending a trailing `;` to a tail-position `bail!`/`eyre::bail!` or wrapping a `arm => bail!(...)` in `{ bail!(...); }`. Since `bail!` expands to a diverging `return Err(...)`, the block form still diverges and coerces in value positions, so there is no control-flow or behavior change. The latest commits look safe to merge — no new issues.

---

Lioyae:
Thanks for the re-review! Happy to adjust the scope further if you'd prefer a smaller change — otherwise I'll keep it as-is and wait for your call.

phil-opp:
Update: https://github.com/eyre-rs/eyre/pull/294 was merged and a PR to do a new release is up: https://github.com/eyre-rs/eyre/pull/296

phil-opp:
Heads up — this may no longer be needed. [eyre 0.6.14](https://crates.io/crates/eyre/0.6.14) has been released, and it removes the trailing semicolon from every arm of the `bail!` macro:

```diff
 macro_rules! bail {
     ($msg:literal $(,)?) => {
-        return $crate::private::Err($crate::eyre!($msg));
+        return $crate::private::Err($crate::eyre!($msg))
     };
     …
 }
```

That semicolon was the entire cause of `semicolon_in_expressions_from_macros` / `semicolon_in_expressions_from_non_local_macros`, so with 0.6.14 the lint has nothing left to fire on and the call sites can stay in expression position.

I opened #3126, which is a one-package `Cargo.lock` bump with no source changes. Measured on `rustc 1.99.0-nightly (12c36e253 2026-08-10)`, `cargo +nightly check --all` (excluding the PyO3 crates) goes from **154** `trailing semicolon in macro used in expression position` diagnostics across 17 crates to **0** — that covers all 48 files here plus the ones this PR doesn't reach.

Thank you for the original diagnosis in #2835 and for the careful validation work — separating the `tokio` nightly ICE from the actual dora problem is what made the root cause clear, and it's what made it easy to confirm the upstream release fixes it. Sorry the fix landed upstream after you'd already done the legwork here.

Deferring to the maintainers on whether to close this in favour of #3126, or land it anyway for the benefit of anyone pinned to an older eyre.

---

Issue #1465, opened 2026-03-19
fix(coordinator): fail pending build/spawn waiters on daemon disconnect


Bhanudahiyaa:
## Problem
  Coordinator watchdog currently removes dead daemon connections, but does not propagate that disconnect into pending lifecycle operations.

  When a daemon is disconnected (watchdog timeout or daemon exit), coordinator state may still contain:
  - pending build waiters (`running_builds.pending_build_results`)
  - pending spawn waiters (`running_dataflows.pending_spawn_results`)

  As a result, CLI calls like `wait_for_build` and `wait_for_spawn` can stay pending until RPC deadline instead of failing immediately.

  ## Why This Matters
  This causes poor failure recovery and unclear operator experience in distributed runs:
  - users see long hangs instead of deterministic failure
  - orchestration behavior is inconsistent with actual cluster health
  - retries and remediation are delayed unnecessarily

  This is especially relevant for distributed-state/system-reliability work because daemon↔coordinator consistency under failures is a core contract.

  ## Expected Behavior
  On daemon disconnect, coordinator should:
  1. Fail all pending build waiters that depended on that daemon.
  2. Fail all pending spawn waiters that depended on that daemon.
  3. Surface actionable error messages immediately.

  ## Actual Behavior
  Coordinator removes daemon connections but does not fail pending waiters, so lifecycle RPCs can hang for a long timeout window.

  ## Reproduction (Conceptual)
  1. Start build/start involving daemon A.
  2. Kill daemon A or let watchdog timeout.
  3. Observe that connection is removed.
  4. Observe `wait_for_build`/`wait_for_spawn` remain pending rather than failing immediately.

heyong4725:
## Status update: spawn half is done, build half is still real — rescoping

Verified against current main. The spawn half of this issue's ask was shipped via #1593 (rescue of #1592), but the **build half is still open and the bug is reproducible exactly as described**.

### What shipped (spawn side)

On daemon disconnect, the coordinator already:
- Drains the disconnected daemon's entry from `df.pending_spawn_results` (`binaries/coordinator/src/lib.rs:1996-2003`)
- Releases stuck `wait_for_spawn` waiters via a 60s timeout watchdog `check_spawn_timeouts()` (`lib.rs:2894`) — handles both stuck-daemon and disconnect-mid-spawn cases
- 4 dedicated unit tests at `lib.rs:5509-5673`

### What's still missing (build side)

[...]

PR #2844, opened 2026-07-26
fix(core): skip blank lines in multi-line build commands

phil-opp:
## Issue

Each line of a node's `build:` command is spawned as a separate process. A blank or whitespace-only line yields no tokens, so `split.next()` returned `None` and `run_build_command` aborted the **entire** build with `build command is empty`.

This made an ordinary multi-line block fail whenever it used a blank line for readability:

```yaml
build: |
  pip install -r requirements.txt

  maturin develop
```

The empty middle line triggered the bail, so the build never ran. Module-level builds (concatenated with `\n`) were affected the same way.

## Fix

Skip blank/whitespace-only lines before splitting, via a small `build_command_lines` helper in `libraries/core/src/build/build_command.rs`.

Comment (`#`) lines are intentionally left unchanged: build lines are spawned directly, not through a shell, so `#` was never a comment — treating it as one would be a behavior change, out of scope for this fix.

## Validation

- Added unit tests for `build_command_lines`: blank/whitespace skipping, single-line passthrough, all-blank input.
- `cargo test -p dora-core`, `cargo clippy ... -D warnings`, and `cargo fmt --all --check` pass.

---

---

### DIFF
--- a/libraries/core/src/build/build_command.rs
+++ b/libraries/core/src/build/build_command.rs
@@ -33,6 +33,18 @@ struct LocalDoraPythonSource {
     fingerprint: String,
 }
 
+/// Split a (possibly multi-line) `build:` command into the individual command
+/// lines to execute.
+///
+/// Each line is spawned as a separate command, so blank or whitespace-only
+/// lines carry no program to run. They are skipped here rather than passed on
+/// to the splitter, where an empty token stream would otherwise abort the whole
+/// build with `build command is empty` — a surprising failure for a multi-line
+/// block that merely uses blank lines for readability.
+fn build_command_lines(build: &str) -> impl Iterator<Item = &str> {
+    build.lines().filter(|line| !line.trim().is_empty())
+}
+
 pub async fn run_build_command(
     build: &str,
     working_dir: &Path,
@@ -43,8 +55,7 @@ pub async fn run_build_command(
 ) -> eyre::Result<()> {
     std::fs::create_dir_all(working_dir).context("failed to create working directory")?;
 
-    let lines = build.lines().collect::<Vec<_>>();
-    for build_line in lines {
+    for build_line in build_command_lines(build) {
         let mut split = splitty::split_unquoted_whitespace(build_line).unwrap_quotes(true);
 
         let program = split
@@ -767,7 +778,7 @@ async fn forward_build_output<R1, R2>(
 #[cfg(test)]
 mod tests {
     use super::{
-        cleanup_stale_local_dora_python_wheel_dirs, dora_runtime_install_args,
+        build_command_lines, cleanup_stale_local_dora_python_wheel_dirs, dora_runtime_install_args,
         find_local_dora_python_wheel, forward_build_output, local_dora_python_package_dir,
         local_dora_python_shared_wheel_dir, local_dora_python_source_fingerprint,
         local_dora_python_wheel_root, local_runtime_marker_matches,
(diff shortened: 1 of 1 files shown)


phil-opp:

This PR is part of a batch from an automated codebase-review pass (fan-out over CLI / daemon / core / node+message / coordinator+extensions crates). Each finding below was verified against `origin/main` and checked against prior open automated PRs to avoid duplicates. All six are small, independent, and land on their own branch.

### PRs opened, by category

**Correctness**
- #2844 — `fix(core)`: skip blank/whitespace lines in multi-line `build:` commands (a blank line for readability aborted the whole build with `build command is empty`).
- #2847 — `fix(node)`: accept `off`/`0`/`false` as disable spellings for `DORA_RUNTIME_TYPE_CHECK` (setting `=false` warned "unknown value" despite doing the intended thing).
- #2849 — `fix(cli)`: validate node name in `dora new --kind node` (the dataflow path already rejects `/` and non-ASCII; the node path didn't).

**Performance (hot path)**
- #2845 — `perf(daemon)`: early-out in `note_output_sent_to_local_receivers` when no input deadlines / circuit breakers are configured — skips a clock read + per-receiver channel lookup on every `OutputSent` in the common case.

**Simplification / robustness**
- #2846 — `refactor(node)`: delegate the scheduler's correlation check to the canonical `carries_pattern_correlation`, removing a duplicated key list that could silently diverge and drop service/action messages.

**Documentation**
- #2848 — `docs(message)`: document the `Input` / `InputMapping` / `UserInputMapping` public API with a compile-checked `FromStr`↔`Display` doctest.

- **Coordinator — `DaemonExit` skips build-membership cleanup**: the cleanup fn deliberately defers finalization to the watchdog (documented design #1465); adding the call alone doesn't fix the described hang (last-reporter-first ordering), and the fuller fix contradicts the intentional single-chokepoint design.
- **Coordinator — OTel gauges recorded for finalized nodes**: low-severity, `metrics`-feature-gated cosmetic (a stray sample for a just-stopped node); not worth a standalone PR.
- **Core — `check_module_file` nested-module containment stricter than expansion**: the standalone validator has no project root available, so aligning it would weaken a path-escape guard — fix is ambiguous.
- **Core — `clone_dir_path` doesn't sanitize `commit_hash`**: robustness only; git ref names disallow `..`, and changing derived cache-dir paths risks churn — low value.
- **Core — `merge_env` returns `Some({})` instead of `None` when both sides are empty**: cosmetic serialization difference, no functional impact.
- **Node — `Stop`/`AllInputsClosed` evictable from an overfull non-input queue**: near-unreachable given the 1000-event margin; the fix needs an eviction-policy redesign — too risky for a latent low-severity issue.
- **Node — `read_json_value_as_arrow` indexes `column(0)` without a guard**: the panic is effectively unreachable via in-tree callers (the `{"inner": …}` wrapping always yields ≥1 column); pure defensive hardening on a `pub` fn — weak.
- **Daemon — pure-local per-message `Metadata` clone**: a genuine hot-path allocation, but it sits on the same send path being actively modified by open metadata-clone PRs (#2761 / #2837) — held to avoid conflicts.
- **Daemon — `dora node restart` no-ops past `max_restarts`**: narrow (only nodes with `max_restarts > 0` that already exhausted the budget); changing restart-gate semantics is design-sensitive — left for maintainers.
- **Daemon — `grace_duration_kills` exit-classification race**: an inherent race with no clean happens-before fix; flagged only.
- **Daemon — `AddNode` does blocking spawn work on the event loop**: real, but the fix (offload to a background task like the `Spawn` path) is a larger refactor beyond a small self-contained PR.

---

⚠️ These PRs are **machine-generated by Claude (Claude Code)**, an AI assistant, and should be reviewed carefully before merging.

---

phil-opp:

I reviewed the diff (single commit `987491e`). No issues found.

- `build_command_lines` (`libraries/core/src/build/build_command.rs`) correctly filters blank/whitespace-only lines with `line.trim().is_empty()`, which resolves the `build command is empty` abort on multi-line `build:` blocks that use blank lines for readability.
- The `#`-line behavior is left unchanged, which is correct since build lines are spawned directly rather than through a shell.
- The three added tests exercise the actual helper and assert meaningful output (blank/whitespace skipping, single-line passthrough, all-blank → no commands).

No correctness, security, concurrency, or resource concerns in this change.

---

PR #2846, opened 2026-07-26
refactor(node): delegate scheduler correlation check to the canonical predicate

phil-opp:
## Issue

The event-stream scheduler's `is_correlated` inlined the same three-key check (`request_id` / `goal_id` / `goal_status`) that `dora_message` already exposes as `carries_pattern_correlation`. That canonical function's own doc explicitly asks callers to keep it the single definition, so the send-side, node-receive-side, and daemon-debug layers can never disagree on what "pattern-correlated" means.

The scheduler is the one place that had drifted into a private copy. The two are identical today, so there is **no active bug** — but if a fourth correlation key were ever added to `carries_pattern_correlation`, the scheduler would keep treating those messages as non-correlated and could evict them under queue pressure, silently breaking the service/action request-response contract this eviction logic exists to protect (dora-rs/adora#145).

## Fix

Delegate `is_correlated` to `carries_pattern_correlation` and drop the duplicated key list. `REQUEST_ID`/`GOAL_ID`/`GOAL_STATUS` remain imported for the per-key structured fields in `log_correlation_drop`.

## Validation

- `cargo test -p dora-node-api` passes, including the existing scheduler eviction tests.
- `cargo clippy ... -D warnings` and `cargo fmt --all --check` pass.

---

---

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -7,7 +7,7 @@ use dora_message::{
     config::{DEFAULT_QUEUE_SIZE, QueuePolicy},
     daemon_to_node::NodeEvent,
     id::DataId,
-    metadata::{GOAL_ID, GOAL_STATUS, REQUEST_ID, get_string_param},
+    metadata::{GOAL_ID, GOAL_STATUS, REQUEST_ID, carries_pattern_correlation, get_string_param},
 };
 
 use super::thread::EventItem;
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. `carries_pattern_correlation` has the exact same three-key body (`REQUEST_ID` / `GOAL_ID` / `GOAL_STATUS`) that was inlined, so this is behavior-preserving, and the `REQUEST_ID`/`GOAL_ID`/`GOAL_STATUS` imports are still used by `log_correlation_drop`, so the change compiles cleanly with no unused imports. Delegating to the single canonical predicate is a reasonable way to keep the scheduler's eviction guard from drifting from the send/receive/daemon layers.

---

PR #2847, opened 2026-07-26
fix(node): accept off/0/false as disable spellings for DORA_RUNTIME_TYPE_CHECK

phil-opp:
## Issue

`RuntimeTypeCheck::from_env` accepted `1`/`warn`/`true` (→ warn) and `error`, and treated the empty string / unset as `Off`. But the natural "disable" spellings `0`, `false`, and `off` fell into the unknown-value arm, logging:

```
unknown DORA_RUNTIME_TYPE_CHECK value "false", expected "warn" or "error"; disabling runtime type check
```

So a user who set `DORA_RUNTIME_TYPE_CHECK=false` to turn the feature off got a warning implying they had mistyped — even though the resulting behavior (disabled) was exactly what they asked for.

## Fix

Treat `` / `0` / `false` / `off` as `Off` without a warning, and accept `on` alongside the existing enable spellings for symmetry. Genuinely unrecognized values still warn and disable, as before.

The env read is split into a pure `from_value(Option<&str>)` helper so parsing is unit-testable without mutating process-global env state.

## Validation

- Added unit tests covering enable spellings, disable spellings, unset (`None`), and an unknown value.
- `cargo test -p dora-node-api`, `cargo clippy ... -D warnings`, and `cargo fmt --all --check` pass.

---

---

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -61,11 +61,19 @@ enum RuntimeTypeCheck {
 
 impl RuntimeTypeCheck {
     fn from_env() -> Self {
-        match std::env::var("DORA_RUNTIME_TYPE_CHECK").as_deref() {
-            Ok("error") => Self::Error,
-            Ok("1" | "warn" | "true") => Self::Warn,
-            Ok("") | Err(_) => Self::Off,
-            Ok(other) => {
+        Self::from_value(std::env::var("DORA_RUNTIME_TYPE_CHECK").ok().as_deref())
+    }
+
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Splitting the env read from the pure `from_value(Option<&str>)` parser makes the arms unit-testable without touching process env, and the tests cover enable spellings, disable spellings, unset (`None`), and an unknown value.

Minor, non-blocking: the fallback warning still reads `expected "warn" or "error"` even though `on`/`off`/`0`/`1`/`true`/`false` are now accepted, so the message slightly understates the valid set — worth updating if you touch this again, but not important.

---

PR #2849, opened 2026-07-26
fix(cli): validate node name in `dora new --kind node`

phil-opp:
## Issue

`create_dataflow` rejects names containing `/` or non-ASCII characters, but the `--kind node` path (`create_custom_node`) had no such guard. It built directories straight from the raw name via the non-recursive `fs::create_dir`, so:

```
dora new --lang python --kind node a/b
```

failed with a confusing "failed to create root directory" (or, if `a` already existed, silently created a nested path the user didn't intend). Non-ASCII names were likewise unvalidated.

## Fix

Apply the same `contains('/')` / `is_ascii()` checks that `create_dataflow` already uses, with clear error messages. Spaces remain allowed — they're normalized to `-`/`_` — so the internal `create_dataflow` callers (`"talker 1"`, `"listener 1"`, …) are unaffected.

## Validation

- `cargo build -p dora-cli`, `cargo clippy ... -D warnings`, and `cargo fmt --all --check` pass.
- The `dora new --kind dataflow` template path still generates its custom nodes (its calls use space-separated ASCII names, which pass validation).

---

---

### DIFF
--- a/binaries/cli/src/template/python/mod.rs
+++ b/binaries/cli/src/template/python/mod.rs
@@ -39,6 +39,16 @@ fn create_custom_node(
     path: Option<PathBuf>,
     main: &str,
 ) -> Result<(), eyre::ErrReport> {
+    // Reject names that would turn into path separators or non-ASCII module
+    // directories, mirroring the validation in `create_dataflow`. Spaces are
+    // fine — they are normalized to `-`/`_` below.
+    if name.contains('/') {
+        bail!("node name must not contain `/` separators");
+    }
+    if !name.is_ascii() {
+        bail!("node name must be ASCII");
+    }
+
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The `contains('/')` and `!is_ascii()` guards match the existing `create_dataflow` validation in the same file, `bail` is already in scope, and spaces are still allowed (normalized to `-`/`_`), so the internal callers that pass names like `"talker 1"` are unaffected.

One optional, non-blocking note: like `create_dataflow`, this does not reject a Windows-style `\` separator, so `dora new --kind node a\b` would still create a nested path on Windows — but that matches the existing behavior you're mirroring, so leaving it out is reasonable.

---

PR #2850, opened 2026-07-27
fix(core): fall back to non-multicast zenoh discovery when scouting bind fails

bobdingAI:
## Summary

`open_zenoh_session_with_listen` (`libraries/core/src/topics.rs`, called from the
daemon at `binaries/daemon/src/lib.rs:1225`) opens the tuned zenoh session and,
on failure, retried with a plain `zenoh::Config::default()`. The default keeps
**multicast scouting enabled**, so when the failure's *cause* is the scouting
socket failing to bind its multicast group (`224.0.0.224:7446`), the retry
re-attempts the identical bind, fails again, and the dataflow aborts at startup.

This bites when dora shares one network namespace with a busy DDS/ROS 2 multicast
graph — i.e. **dev containers / CI / any co-located `docker`-netns setup** (the
existing comment in `topics.rs` already notes multicast scouting is "broken in
dev containers and many CI environments"). Reproduced in `docker/ros2dev` (dora +
ROS 2 in one netns) alongside a ~25-participant DDS graph.

## Fix

Disable multicast scouting on the **fallback retry** only. Gossip + explicit
connect endpoints (`DORA_ZENOH_CONNECT`) still cover discovery for co-located and
connect-configured setups, so degrading to non-multicast keeps dora running where
it would otherwise fail outright. The **primary open is unchanged** — normal
(uncontended) environments still get multicast by default and the fallback never
triggers. It's strictly safer than today: the fallback runs only *after* the
primary already failed to bind multicast, so a multicast-dependent setup wasn't
getting multicast on that attempt anyway.

## Scope — what this does and does NOT fix

This is the **dora daemon zenoh-transport** layer only. It does **not** touch the
ROS 2 bridge's DDS path (`Ros2Transport::Dds` / `Ros2Context` default), which
opens no zenoh session at all.

There is a **separate** discovery bug with a confusingly similar ~50% rate: the
ROS 2 bridge's **RustDDS discovery** going silently empty under DDS domain
contention (process healthy, subscription just receives nothing — no bind error,
no daemon required). **This PR does not address that**; it lives at the
`ros2_client`/`rustdds` layer, not in dora's zenoh session. The two share an
environmental cause (multicast contention) but are distinct code paths.

Host deployments where dora and ROS 2 do **not** share a netns are not exposed to
the failure this PR fixes (measured: 10/10 dataflow starts with and without
`ZENOH_CONFIG` against a live 32-node graph, zero bind errors) — for those, the
DDS-layer bug above is the relevant one.

## Validation

`rust-ros2-dataflow` bridge example vs a ~25-participant DDS graph, 12 runs each,
in `docker/ros2dev` (dora + ROS 2 share one netns):

| Condition | Dataflow starts |
|---|---|
| clean graph, multicast on (main) | 11/12 |
| **busy graph, multicast on (main)** | **0/12** — all fail on `224.0.0.224:7446 Address already in use` |
| **busy graph, this PR** | **12/12** — primary fails, fallback logs `retrying with multicast scouting disabled`, dataflow finishes |
| clean graph, this PR | unchanged (fallback never triggers) |

`cargo fmt` / `cargo clippy -p dora-core -- -D warnings` clean.

A deterministic unit test isn't practical (triggering the fallback needs a
contended-multicast environment where the primary bind fails), so this is
validated by the reproduction above.

## Workaround (no rebuild)

Set `ZENOH_CONFIG` to a config with `scouting: { multicast: { enabled: false } }`,
or set `DORA_ZENOH_CONNECT`, to opt out of multicast scouting directly.

### DIFF
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -342,12 +342,29 @@ pub async fn open_zenoh_session_with_listen(
                 }
                 Err(err) => {
                     warn!(
-                        "failed to open tuned zenoh session ({err}), retrying with default config"
+                        "failed to open tuned zenoh session ({err}), retrying with \
+                         multicast scouting disabled"
                     );
                     // Default fallback has no listener; `effective_listen_endpoint`
                     // stays `None` so peers don't try to reach a bind that isn't
                     // there (#1856).
-                    let zenoh_config = zenoh::Config::default();
+                    //
+                    // Disable multicast scouting on the retry. The most common open
+                    // failure is the scouting socket failing to bind its multicast
+                    // group (`224.0.0.224:7446`) — e.g. when a busy DDS/ROS2
+                    // multicast graph (a full TurtleBot3 / nav2 stack) saturates the
+                    // multicast networking. A plain `Config::default()` retry keeps
+                    // multicast enabled, re-attempts the same failing bind, and takes
+                    // the whole dataflow down. Gossip + explicit connect endpoints
+                    // still cover discovery for co-located and connect-configured
+                    // setups, so degrading to non-multicast here keeps dora running
+                    // where it would otherwise fail outright.
+                    let mut zenoh_config = zenoh::Config::default();
+                    if let Err(err) =
+                        zenoh_config.insert_json5("scouting/multicast/enabled", "false")
+                    {
+                        warn!("failed to disable zenoh scouting/multicast on fallback: {err}");
+                    }
                     zenoh::open(zenoh_config)
                         .await
                         .map_err(|e| eyre!(e))


phil-opp:

No issues found. Disabling multicast scouting only on the fallback retry is a strictly-safer degradation: the fallback runs only after the primary tuned open already failed to bind, so a multicast-dependent setup wasn't getting multicast on that attempt anyway, and gossip / explicit `DORA_ZENOH_CONNECT` endpoints still cover discovery. `insert_json5("scouting/multicast/enabled", "false")` passes a valid JSON5 boolean, and its error path only warns rather than aborting the retry, so a future zenoh key rename degrades gracefully. The primary open is untouched, so uncontended environments are unaffected. Agreed that a deterministic unit test isn't practical here given it requires a contended-multicast bind failure.

---

phil-opp:
**Automated review.**

The fallback direction is reasonable for the co-located / dev-container multicast-contention case this targets, and the `insert_json5("scouting/multicast/enabled", "false")` call matches the existing usage at `topics.rs:347`. One correctness gap is worth tightening, though:

The retry disables multicast **unconditionally**, but the `Err(err)` arm (`libraries/core/src/topics.rs:446`) is reached on *any* `zenoh::open` failure — not only a multicast-bind failure. So the "strictly safer / wasn't getting multicast anyway" rationale only holds when the multicast bind was the actual cause. If the primary open fails for an unrelated reason in a multi-daemon deployment that genuinely relies on multicast scouting, the old `Config::default()` fallback would still recover via multicast, whereas this one disables it. And because the fallback is a bare `Config::default()`, it also carries no listener and none of the `coordinator_addr` connect endpoints that the tuned config set — so the result is a silently-partitioned session rather than the loud startup failure it replaces, which is harder to debug.

Gating the multicast-disable on the error actually being the `224.0.0.224:7446` bind failure — or re-applying the connect endpoints on the fallback so gossip has a seed — would close the gap. Not blocking, and the change is still a net improvement for the target scenario.

---
🤖 _This is a fully automated review by Claude ([Claude Code](https://claude.ai/code)). No human has vetted these findings — please confirm before acting on them._

---

phil-opp:
The `dora run` case should be fixed by #2871. It also adds a `dora daemon --zenoh-no-multicast` flag (and a corresponding env var) to opt out of multicast.

Could you check whether this fixes your issues?

phil-opp:

Thanks for the detailed reproduction — the 0/12 vs 12/12 table is exactly the evidence that makes a change like this easy to reason about.

First, a question that changes the framing: **#2871 merged on 2026-07-28**, the day after you opened this, and added `dora daemon --zenoh-no-multicast`, `DORA_ZENOH_MULTICAST`, and `dora run` opting out on its own. Could you check whether that already covers your `docker/ros2dev` setup? If it does, what is left for this PR is the narrower case of a user who has not found the flag — still worth having, but it should be rebased onto that world and re-described.

On the fallback itself, two things need to change.

**The retry config has no discovery path left.** The fallback builds a fresh `zenoh::Config::default()`, which discards everything the tuned config assembled: the operator's `connect_endpoints`, the `DORA_ZENOH_CONNECT` endpoints merged at `topics.rs:581`, the coordinator endpoints at `:760`, the listener, and `DORA_ZENOH_CONFIG_OVERLAY`. With multicast also off, there is nothing to dial, nothing to be dialled on, and gossip has no seed link — the silent partition the guard at `topics.rs:729-757` exists to prevent ("Drop multicast scouting only once this session is reachable some other way"). So "gossip + explicit connect endpoints still cover discovery" is not true of the config the retry actually opens. Deriving the retry from the tuned config with only the failing pieces removed, and disabling multicast only once the session has something to dial, would fix both.

**The disable is unconditional, but the `Err` arm catches every `zenoh::open` failure**, not just a multicast bind failure. Gating it on `multicast_disabled(matches!(multicast, MulticastScouting::Disabled))` (the helper at `topics.rs:744`) and/or on the error naming the scouting bind would be better. That gating is worth having on its own merits: today the fallback re-enables multicast even for an operator who passed `--zenoh-no-multicast` or `DORA_ZENOH_MULTICAST=off`, which is a real inconsistency.

Related: the change converts a hard startup failure into a running-but-partitioned daemon, with `warn!` as the only trace, and the message asserts a cause it never checked. Saying what was actually lost, and raising it to `error!` when the retry ends up with neither endpoints nor multicast, would keep a genuinely misconfigured deployment obvious in the log rather than surfacing later as "my node receives nothing".

On testing: "a deterministic unit test isn't practical" is only half true. You cannot force the bind failure, but you can test the config the retry *builds*. Factoring it into something like `fallback_zenoh_config(&params) -> zenoh::Config` makes it a plain unit test in the existing `mod tests` in `topics.rs`, which already builds configs and asserts on their JSON: multicast off when the caller asked for it off, multicast on when it did not, and connect endpoints surviving into the retry.

One coordination note: #3340 bounds `connect/timeout_ms`, `scouting/timeout` and `open_timeout` on the tuned config only, so a bare `Config::default()` retry would still open with zenoh's unbounded defaults. No textual conflict, but another argument for deriving the retry from the tuned config.

phil-opp:

Thanks for the detailed reproduction — the 0/12 vs 12/12 table is exactly the evidence that makes a change like this easy to reason about.

First, a question that changes the framing: **#2871 merged on 2026-07-28**, the day after you opened this, and added `dora daemon --zenoh-no-multicast`, `DORA_ZENOH_MULTICAST`, and `dora run` opting out on its own. Could you check whether that already covers your `docker/ros2dev` setup? If it does, what is left for this PR is the narrower case of a user who has not found the flag — still worth having, but it should be rebased onto that world and re-described.

On the fallback itself, two things need to change.

**The retry config has no discovery path left.** The fallback builds a fresh `zenoh::Config::default()`, which discards everything the tuned config assembled: the operator's `connect_endpoints`, the `DORA_ZENOH_CONNECT` endpoints merged at `topics.rs:581`, the coordinator endpoints at `:760`, the listener, and `DORA_ZENOH_CONFIG_OVERLAY`. With multicast also off, there is nothing to dial, nothing to be dialled on, and gossip has no seed link — the silent partition the guard at `topics.rs:729-757` exists to prevent ("Drop multicast scouting only once this session is reachable some other way"). So "gossip + explicit connect endpoints still cover discovery" is not true of the config the retry actually opens. Deriving the retry from the tuned config with only the failing pieces removed, and disabling multicast only once the session has something to dial, would fix both.

**The disable is unconditional, but the `Err` arm catches every `zenoh::open` failure**, not just a multicast bind failure. Gating it on `multicast_disabled(matches!(multicast, MulticastScouting::Disabled))` (the helper at `topics.rs:744`) and/or on the error naming the scouting bind would be better. That gating is worth having on its own merits: today the fallback re-enables multicast even for an operator who passed `--zenoh-no-multicast` or `DORA_ZENOH_MULTICAST=off`, which is a real inconsistency.

Related: the change converts a hard startup failure into a running-but-partitioned daemon, with `warn!` as the only trace, and the message asserts a cause it never checked. Saying what was actually lost, and raising it to `error!` when the retry ends up with neither endpoints nor multicast, would keep a genuinely misconfigured deployment obvious in the log rather than surfacing later as "my node receives nothing".

On testing: "a deterministic unit test isn't practical" is only half true. You cannot force the bind failure, but you can test the config the retry *builds*. Factoring it into something like `fallback_zenoh_config(&params) -> zenoh::Config` makes it a plain unit test in the existing `mod tests` in `topics.rs`, which already builds configs and asserts on their JSON: multicast off when the caller asked for it off, multicast on when it did not, and connect endpoints surviving into the retry.

One coordination note: #3340 bounds `connect/timeout_ms`, `scouting/timeout` and `open_timeout` on the tuned config only, so a bare `Config::default()` retry would still open with zenoh's unbounded defaults. No textual conflict, but another argument for deriving the retry from the tuned config.

phil-opp:

Thanks for the detailed reproduction — the 0/12 vs 12/12 table is exactly the evidence that makes a change like this easy to reason about.

First, a question that changes the framing: **#2871 merged on 2026-07-28**, the day after you opened this, and added `dora daemon --zenoh-no-multicast`, `DORA_ZENOH_MULTICAST`, and `dora run` opting out on its own. Could you check whether that already covers your `docker/ros2dev` setup? If it does, what is left for this PR is the narrower case of a user who has not found the flag — still worth having, but it should be rebased onto that world and re-described.

On the fallback itself, two things need to change.

**The retry config has no discovery path left.** The fallback builds a fresh `zenoh::Config::default()`, which discards everything the tuned config assembled: the operator's `connect_endpoints`, the `DORA_ZENOH_CONNECT` endpoints merged at `topics.rs:581`, the coordinator endpoints at `:760`, the listener, and `DORA_ZENOH_CONFIG_OVERLAY`. With multicast also off, there is nothing to dial, nothing to be dialled on, and gossip has no seed link — the silent partition the guard at `topics.rs:729-757` exists to prevent ("Drop multicast scouting only once this session is reachable some other way"). So "gossip + explicit connect endpoints still cover discovery" is not true of the config the retry actually opens. Deriving the retry from the tuned config with only the failing pieces removed, and disabling multicast only once the session has something to dial, would fix both.

**The disable is unconditional, but the `Err` arm catches every `zenoh::open` failure**, not just a multicast bind failure. Gating it on `multicast_disabled(matches!(multicast, MulticastScouting::Disabled))` (the helper at `topics.rs:744`) and/or on the error naming the scouting bind would be better. That gating is worth having on its own merits: today the fallback re-enables multicast even for an operator who passed `--zenoh-no-multicast` or `DORA_ZENOH_MULTICAST=off`, which is a real inconsistency.

Related: the change converts a hard startup failure into a running-but-partitioned daemon, with `warn!` as the only trace, and the message asserts a cause it never checked. Saying what was actually lost, and raising it to `error!` when the retry ends up with neither endpoints nor multicast, would keep a genuinely misconfigured deployment obvious in the log rather than surfacing later as "my node receives nothing".

On testing: "a deterministic unit test isn't practical" is only half true. You cannot force the bind failure, but you can test the config the retry *builds*. Factoring it into something like `fallback_zenoh_config(&params) -> zenoh::Config` makes it a plain unit test in the existing `mod tests` in `topics.rs`, which already builds configs and asserts on their JSON: multicast off when the caller asked for it off, multicast on when it did not, and connect endpoints surviving into the retry.

One coordination note: #3340 bounds `connect/timeout_ms`, `scouting/timeout` and `open_timeout` on the tuned config only, so a bare `Config::default()` retry would still open with zenoh's unbounded defaults. No textual conflict, but another argument for deriving the retry from the tuned config.

phil-opp:
🤖 **Automated review** — fully automated Claude Code review; no human vetted this. Advisory only, not an approval.

New finding on the current head (`bdb2327`): I don't think the fallback is safe as written, and its stated mitigation doesn't actually hold.

The retry arm fires on **any** `zenoh::open` error (the `Err(err)` arm of the tuned open in `libraries/core/src/topics.rs`), not only a multicast-scouting-bind failure. But the fallback config is a bare `zenoh::Config::default()` — it does **not** re-run `build_zenoh_config`, so it carries no `connect/endpoints` (neither `DORA_ZENOH_CONNECT` nor the per-node connect plan) and no listener. On `main`, that fallback at least still had multicast scouting, so discovery could still function. With this change, when the fallback fires for a cause **unrelated to multicast**, the session is left with **no discovery mechanism at all**: multicast disabled, no connect endpoints, no listener → gossip has no seed peer to bootstrap from. The dataflow "starts" (the process no longer aborts), but inter-node / inter-daemon data can silently never flow — which is worse than the current abort.

The commit message says *"gossip + explicit connect endpoints (`DORA_ZENOH_CONNECT`) still cover discovery"*, but the fallback `Config::default()` contains neither, so that mitigation isn't present in the retry config. (The repro counts sessions that *started*, which wouldn't catch a silent no-delivery partition.)

Suggested direction: either disable multicast only when the open error is specifically the scouting-bind failure, or rebuild the fallback from `build_zenoh_config` with multicast forced off (so the connect/listen endpoints survive) instead of starting from a bare `Config::default()`.

---

phil-opp:
🤖 Automated review (fully automated Claude Code review — no human vetted this; advisory only, posted like an outside contributor, I can't approve).

New angle on top of the standing dropped-endpoints note: the multicast-disable retry (`libraries/core/src/topics.rs`, `open_zenoh_session_with_listen` `Err` arm) fires on **any** tuned-open error, not just the `224.0.0.224:7446` multicast-scout-bind failure it targets. So an unrelated or transient tuned-open failure silently downgrades the whole session to no-multicast (and, per the standing finding, no configured listener) for its entire lifetime — multicast-only multi-machine discovery can then stop working after an unrelated hiccup, and the single `warn!("retrying with multicast scouting disabled")` doesn't let an operator distinguish "disabled on purpose" from "degraded for an unrelated reason." Scoping the multicast-disable to the actual bind-failure error (or at least surfacing that discovery is degraded) would avoid the silent regression. The fallback JSON5 itself is correct.

---

phil-opp:
I think this should be fixed by #2871 and there was no reply since [my comment above](https://github.com/dora-rs/dora/pull/2850#issuecomment-5119811995). So this seems to be fixed on `main`.

PR #2852, opened 2026-07-27
fix(core): stop module linter over-rejecting cross-directory nested refs

### LINKED ISSUE #2851: Module validation: `check_module_file` rejects cross-directory nested module refs that real expansion accepts
> **Note:** This issue was created by an automated scheduled Claude code-review check. It reviews a randomly selected component each run. Please verify the analysis before acting on it — this is low severity and may be intentional.

## Summary

In `libraries/core/src/descriptor/expand.rs`, the standalone module linter `check_module_file` and the real expansion path `expand_module_node` disagree on the **containment boundary** applied to a *nested* module reference:

- **`check_module_file`** confines a nested module path to the **module file's own directory** (`module_dir`).
- **`expand_module_node`** (used by `dora run` / `dora build`) confines it to the **project root** (`canonical_base`, threaded through recursion).

As a result, a module file that references a nested module in a *sibling directory still inside the project* (e.g. `module: ../shared/base.yml`) is **expanded successfully** by real expansion, but **spuriously rejected** by `dora expand --module <file>` with `nested module path ... escapes the module directory`.

This is a false-rejection in the validator's contract: `dora expand --module` reports a valid, runnable module file as invalid.

## Locations

**Stricter boundary — `check_module_file` (nested module check), lines ~217–231:**

```rust
let nested = module_dir.join(mod_path);
let nested_canonical = nested.canonicalize().with_context(|| { /* not found */ })?;
let base_canonical = module_dir.canonicalize()?;          // <-- boundary = module's own dir
if !nested_canonical.starts_with(&base_canonical) {
    bail!(
        "module `{}`: nested module path `{}` escapes the module directory",
        module_file.module.name,
        mod_path,
    );
}
```

**Looser boundary — `expand_module_node`, lines ~359–370:**

```rust
let module_path = base_dir.join(module_path_str);
let canonical = module_path.canonicalize()
    .with_context(|| format!("module file not found: {}", module_path.display()))?;

if !canonical.starts_with(canonical_base) {                // <-- boundary = project root
    bail!(
        "module path `{}` escapes the project directory (node `{}`)",
        module_path_str,
        node.id
    );
}
```

Nested expansion recurses with `base_dir = module_dir` but keeps `canonical_base` as the project root (see the `expand_module_node(&inner_node, module_dir, canonical_base, depth + 1, seen)` call, line ~523), so anything inside the project root is accepted there.

## Reproduction (conceptual)

Project layout:

```
project/
  dataflow.yml            # references modules/a/mod.yml
  modules/a/mod.yml       # has a node with `module: ../shared/base.yml`
  modules/shared/base.yml
```

- `dora run dataflow.yml` / `dora build` — nested `../shared/base.yml` is inside `project/`, passes the `canonical_base` check, expands fine.
- `dora expand --module modules/a/mod.yml` — `check_module_file` uses `module_dir = modules/a` as the boundary; `../shared/base.yml` canonicalizes to `modules/shared/base.yml`, which is **not** under `modules/a`, so it bails with `nested module path ... escapes the module directory`.

Same file, two contradictory verdicts.

## Impact

- Low severity: the divergence is an **over-rejection** in the `dora expand --module` lint command only. It never accepts an invalid file, and it does not affect running or building dataflows.
- The concern is the broken validator contract (printing an error for a file that is actually valid and runnable) and the general-principle inconsistency between the two code paths.

## Possible resolutions

Either is reasonable — the maintainers should pick based on intended semantics:

1. If cross-directory nested modules within a project are supported (as real expansion implies), relax `check_module_file` so it does not reject nested refs that merely leave the module's own directory. In isolation there is no project root, so the safest change is to drop the `module_dir` containment check (keep the existing absolute-path rejection and the `canonicalize()` existence check), or accept any nested module found on disk.
2. If the isolation-scoped strictness is intentional, document the limitation (that standalone `--module` validation cannot resolve cross-directory nested references) and, ideally, downgrade the hard error to a warning so the command does not report a runnable module as invalid.

A regression test covering a module whose nested reference points to a sibling directory inside the project (accepted by `expand_modules`, and whatever behavior is chosen for `check_module_file`) would pin the decision.

phil-opp:
Fixes #2851

## Problem

`check_module_file` (the `dora expand --module` linter) confined a nested module reference to the module file's **own directory** (`module_dir`), while the real expansion path `expand_module_node` (used by `dora run` / `dora build`) confines nested modules to the **project root** (`canonical_base`).

As a result a module referencing a sibling module inside the same project — e.g. `module: ../shared/base.yml` — **expands and runs fine** but was **spuriously rejected** by `dora expand --module` with `nested module path ... escapes the module directory`. Two contradictory verdicts for the same, runnable file.

## Fix

The linter runs on a module file in isolation, with no project root to bound against, so the `module_dir` containment check was both stricter than real expansion and unable to enforce a meaningful boundary. This follows resolution #1 from the issue: drop the containment check, keeping:

- the absolute-path rejection (nested paths must be relative), and
- the `canonicalize()` existence check (nested module must exist on disk).

The real containment boundary is still enforced where it matters — the actual expansion path (`expand_module_node`), which bounds nested modules to the project root.

## Test

Adds `check_module_file_accepts_cross_directory_nested_ref`, which lays out `modules/a/mod.yml` referencing `../shared/base.yml` and asserts the linter accepts it (matching real expansion).

## Validation

- `cargo test -p dora-core --lib check_module_file` — 5 passing (incl. new regression test)
- `cargo clippy -p dora-core -- -D warnings` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_015wt7e5qHsQ9MqtvCBJ4BaM)_

### DIFF
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -221,15 +221,21 @@ pub fn check_module_file(module_path: &Path) -> eyre::Result<()> {
                     module_file.module.name, mod_path, node.id,
                 )
             })?;
-            let base_canonical = module_dir.canonicalize()?;
-            if !nested_canonical.starts_with(&base_canonical) {
-                bail!(
-                    "module `{}`: nested module path `{}` escapes the module directory",
-                    module_file.module.name,
-                    mod_path,
-                );
-            }
-            // Load nested module to collect its declared outputs
+            // Note: unlike `expand_module_node`, we intentionally do NOT reject
+            // a nested reference that leaves `module_dir`. The real expansion
+            // path confines nested modules to the *project root*
+            // (`canonical_base`, threaded through recursion), which routinely
+            // sits above an individual module's directory -- a module in
+            // `modules/a/` may reference a sibling module in `modules/shared/`
+            // via `../shared/base.yml`. This linter runs on a module file in
+            // isolation, with no project root to bound against, so a
+            // `module_dir` containment check would spuriously reject
+            // cross-directory references that `dora run` / `dora build` accept
+            // and run fine (see #2851). We keep the absolute-path rejection and
+            // the `canonicalize()` existence check above; the containment
+            // boundary is enforced by the real expansion path, not this lint.
+            //
+            // Load nested module to collect its declared outputs.
             let nested_module = load_module_file(&nested_canonical)?;
             for output in &nested_module.module.outputs {
                 inner_outputs.insert(output.to_string());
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The linter was stricter than the real expansion path (`module_dir` vs `canonical_base`/project-root), so dropping the containment check in `check_module_file` correctly removes the contradictory verdict described in #2851. Since this is a lint (`dora expand --module`) and the actual containment boundary is still enforced in `expand_module_node`, making the linter more permissive can only under-reject — it does not weaken any runtime/security guarantee. The new `check_module_file_accepts_cross_directory_nested_ref` test asserts the real behavior (sibling `../shared/base.yml` accepted), and the absolute-path rejection plus `canonicalize()` existence check are retained.

---

PR #2858, opened 2026-07-27
fix(runtime): keep a shared-library operator's .so mapped until the event loop joins

phil-opp:
A shared-library operator's outputs are Arrow arrays exported over the C data
interface; `arrow::ffi::from_ffi` (in `send_output_closure`) imports them with a
`release` callback that points **into** the operator's `.so`. The runtime
unloaded that library (dlclose on `Library` drop) as soon as the operator's
`on_event` loop returned — but the main event loop runs on another thread and can
still hold in-flight arrays at that point (`OperatorEvent::Finished` only
guarantees the channel slot was taken, not that every prior output has been sent
and dropped). Freeing those arrays after the unload jumps into unmapped code:
`Signal(11)` / SIGSEGV, reliably reproducible under a high output rate at
shutdown.

**Fix:** hand the loaded `Library` back through `run_operator` to `main`, which
holds it until *after* the main loop has joined — by which point every exported
array has been released. Deterministic; zero-copy preserved (no leak, no
per-output copy).

**Regression test:** a shared-library operator floods large outputs into a sink
so a burst is in flight at `--stop-after`; the smoke test asserts `dora run`
exits cleanly. Verified RED→GREEN — it fails with `Node source failed: exited
because of signal SIGSEGV` on the pre-fix runtime and passes with the fix.

Found while diagnosing #2742 (independent of that Windows wedge).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6374,6 +6374,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "rust-operator-teardown-example-operator"
+version = "1.0.0-rc.4"
+dependencies = [
+ "dora-operator-api",
+]
+
+[[package]]
 name = "rust-ros2-example-node"
 version = "1.0.0-rc.4"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ members = [
     "examples/multiple-daemons/node",
     "examples/multiple-daemons/operator",
     "examples/multiple-daemons/sink",
+    "examples/rust-operator-teardown/operator",
     "examples/cpu-affinity-probe",
     "examples/log-sink-tcp",
     "examples/log-sink-file",
--- a/binaries/runtime/src/lib.rs
+++ b/binaries/runtime/src/lib.rs
@@ -95,7 +95,12 @@ pub fn main() -> eyre::Result<()> {
     });
 
     let operator_id = operator_definition.id.clone();
-    run_operator(
+    // Keep the operator's shared library mapped until *after* the main event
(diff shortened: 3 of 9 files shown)


phil-opp:
Reviewed the diff and surrounding code (`binaries/runtime/src/lib.rs`, `operator/mod.rs`, `operator/shared_lib.rs`). No issues found.

The fix is correct: `shared_lib::run` runs the operator's `on_event` loop to completion and now returns the loaded `libloading::Library` instead of dropping it. All borrows into the library (`Bindings`/`Symbol<'lib>`, `OperatorContext::drop_fn`) live inside the `catch_unwind` closure and are released before `Ok(library)`, so moving the library out is sound. `run_operator` threads it up as `Option<Library>` — the shared-lib arm yields `Some`, the Python arm yields `None`, and the Wasm/no-python arms diverge via `bail!`, so the match type-checks. `main` holds it in `_operator_library` until after `main_task.join()`, i.e. until the async `run` loop has sent and dropped every exported Arrow array (whose FFI `release` callbacks point into the `.so`). Dropping the library only after the join therefore can't dangle those callbacks — this closes the shutdown SIGSEGV.

The regression fixture is genuine: the operator floods 20×512KB outputs per 5ms tick so a burst is in flight at `--stop-after`, and `run_smoke_test_local` asserts a clean exit, which the pre-fix runtime fails with `Signal(11)`. Both nodes loading the same library is fine (separate runtime processes, separate maps).

---
🤖 _This is a fully automated review by Claude — no human has reviewed these findings. Please verify before relying on them._

---

PR #2860, opened 2026-07-27
fix(cli): don't drop compact JSONL log lines that omit the `level` key

phil-opp:
## Issue

`parse_jsonl_line` in `binaries/cli/src/output.rs` read a compact log line's level with:

```rust
let level_str = v.get("level")?.as_str().unwrap_or("stdout");
```

The `?` short-circuits the **whole function** to `None` when the `level` key is absent, so a compact log line carrying `ts` + `msg` but no `level` was silently **discarded** instead of shown. That also made the trailing `.unwrap_or("stdout")` dead code — it can only ever run when the key is present but non-string, never for the missing-key case the default was clearly written to cover.

The first-party daemon writer currently always emits `level`, so this is latent for dora's own logs, but it bites any legacy / hand-written / third-party compact JSONL the CLI is asked to read, and the intent of the code (default to stdout) was already contradicted by its own control flow.

## Fix

Read the key without the short-circuit:

```rust
let level_str = v.get("level").and_then(|l| l.as_str()).unwrap_or("stdout");
```

so a missing (or non-string) level falls back to stdout-level, matching the intent and the `.get().and_then()` shape already used for the adjacent `node` / `msg` / `target` fields.

## Validation

- Added regression test `parse_jsonl_compact_without_level_defaults_to_stdout`, asserting a compact line without `level` is parsed (not dropped) and defaults to `Stdout`.
- `cargo test -p dora-cli`, `cargo clippy -p dora-cli --all-targets -- -D warnings`, and `cargo fmt --all -- --check` all pass.

---

🤖 This is a machine-generated pull request opened by Claude Code. A human maintainer should review before merging.

https://claude.ai/code/session_0142J7B2578ixrh77kNDNJry

---
_Generated by [Claude Code](https://claude.ai/code/session_0142J7B2578ixrh77kNDNJry)_

### DIFF
--- a/binaries/cli/src/output.rs
+++ b/binaries/cli/src/output.rs
@@ -183,7 +183,11 @@ pub fn parse_jsonl_line(line: &str) -> Option<LogMessage> {
     let v: serde_json::Value = serde_json::from_str(line).ok()?;
     let ts = v.get("ts")?.as_str()?;
     let timestamp = chrono::DateTime::parse_from_rfc3339(ts).ok()?.to_utc();
-    let level_str = v.get("level")?.as_str().unwrap_or("stdout");
+    // A missing `level` key defaults to stdout-level rather than dropping the
+    // line. The previous `v.get("level")?` short-circuited the whole function
+    // to `None` when the key was absent, which both discarded the line and made
+    // the trailing `.unwrap_or("stdout")` dead code.
+    let level_str = v.get("level").and_then(|l| l.as_str()).unwrap_or("stdout");
     let level = match level_str {
         "error" => LogLevelOrStdout::LogLevel(log::Level::Error),
         "warn" => LogLevelOrStdout::LogLevel(log::Level::Warn),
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The fix is correct: `v.get("level")?` short-circuited the entire function to `None` whenever the `level` key was absent, so a compact line carrying `ts` + `msg` but no `level` was dropped rather than shown — and that same `?` made the trailing `.unwrap_or("stdout")` unreachable for the missing-key case it was written for. Switching to `v.get("level").and_then(|l| l.as_str()).unwrap_or("stdout")` restores the intended stdout-level fallback and matches the `.get().and_then()` shape already used for the adjacent `node` / `msg` / `target` fields.

The regression test exercises the changed path directly (a compact line with no `level`) and asserts both that the line is parsed and that it defaults to `Stdout`, so it would have failed before the fix.

---

phil-opp:
Closing as a duplicate of #3082, which fixes the same bug and goes further: it routes the level through `parse_log_level_str` (the single source of truth) instead of a second hand-rolled match, so a capitalized `INFO` is classified rather than silently downgraded to stdout. It also handles a non-string `level` and covers the behavior with four tests to this PR's one.

PR #2861, opened 2026-07-27
fix(cli): validate cluster machine id and labels before they reach the remote shell

phil-opp:
## Issue

`dora cluster up` interpolates the remote-command fields of the cluster config **verbatim** into the command it runs over SSH (`binaries/cli/src/command/cluster/up.rs`, via `format_labels_arg` / `format_zenoh_peer_arg` in `mod.rs`):

```
nohup dora daemon --machine-id {id} --labels {k}={v},... --zenoh-peer {ep} \
    > /tmp/dora-daemon-{id}.log 2>&1 &
```

`ClusterConfig::validate()` only checked that the id was non-empty and unique, so a value containing whitespace or shell metacharacters would corrupt that command and the log path:

- `id: "a b"` word-splits the daemon arguments.
- a label value like `gpu: "x;rm -rf /"`, or a `zenoh_peer: "tcp/1.2.3.4:7447;rm -rf /"` — or a `$(...)`/backtick in the id — executes arbitrary commands on the remote host.

The `cluster.yml` is operator-authored, so this is a robustness / footgun issue rather than a remote-exploit path, but a malformed value fails confusingly (or dangerously) instead of being rejected up front.

## Fix

Validate every field that lands in the remote command string, in `ClusterConfig::validate()`:

- **`machine.id`** and every **label key/value** → the safe identifier charset dora already enforces for node and data ids, `[a-zA-Z0-9_.-]` (see `libraries/message/src/id.rs`), via a shared `validate_shell_safe` helper.
- **`zenoh_peer`** → a sibling `validate_endpoint_shell_safe` helper that additionally permits the characters a Zenoh endpoint carries (`/`, `:`, and `[`/`]` for IPv6 literals like `tcp/[::1]:7447`) while still rejecting whitespace and shell metacharacters. It can't reuse the id charset, which would reject a legitimate `tcp/1.2.3.4:7447`.

`host` / `user` are deliberately **left unrestricted**: they are network addresses that legitimately carry characters outside these sets and are folded into the ssh *target* argv rather than the remote command string. Hostname-style ids and normal labels/endpoints (`jetson-01`, `gpu=true`, `arch=arm64`, `tcp/10.0.0.1:5456`) remain valid.

## Validation

- Added tests: rejected id / label value / `zenoh_peer` with shell metacharacters, plus accepted conventional ids and an IPv6 `zenoh_peer`. All `cluster::config` tests pass (18 total).
- `cargo test -p dora-cli`, `cargo clippy -p dora-cli --all-targets -- -D warnings`, and `cargo fmt --all -- --check` all pass.

---

🤖 This is a machine-generated pull request opened by Claude Code. A human maintainer should review before merging.

https://claude.ai/code/session_0142J7B2578ixrh77kNDNJry

### DIFF
--- a/binaries/cli/src/command/cluster/config.rs
+++ b/binaries/cli/src/command/cluster/config.rs
@@ -75,6 +75,22 @@ impl ClusterConfig {
             if m.id.is_empty() {
                 bail!("machine id must not be empty");
             }
+            // `machine.id` and every label key/value are interpolated verbatim
+            // into the shell command that `dora cluster up` runs over SSH
+            // (`dora daemon --machine-id {id} --labels {k}={v},... >
+            // /tmp/dora-daemon-{id}.log`). A value containing whitespace or shell
+            // metacharacters would corrupt that command (or, for the id, the log
+            // path), so restrict these fields to the same safe identifier charset
+            // dora uses for node/data ids (`libraries/message/src/id.rs`).
+            // `host`/`user` are deliberately left unrestricted: they are network
+            // addresses that legitimately carry characters outside this set
+            // (e.g. IPv6 literals), and are folded into the ssh *target* rather
+            // than the remote command string.
+            validate_shell_safe(&format!("machine id `{}`", m.id), &m.id)?;
+            for (k, v) in &m.labels {
+                validate_shell_safe(&format!("label key `{k}` on machine `{}`", m.id), k)?;
+                validate_shell_safe(&format!("label value `{v}` on machine `{}`", m.id), v)?;
+            }
             if !seen.insert(&m.id) {
                 bail!("duplicate machine id: `{}`", m.id);
             }
@@ -90,6 +106,19 @@ impl ClusterConfig {
         }
         Ok(())
     }
+}
+
+/// Reject a value that will be interpolated into the remote SSH command unless
+/// it consists only of the safe identifier charset `[a-zA-Z0-9_.-]`. `what`
+/// names the field for the error message.
+fn validate_shell_safe(what: &str, value: &str) -> eyre::Result<()> {
+    if let Some(ch) = value
+        .chars()
+        .find(|c| !c.is_ascii_alphanumeric() && *c != '_' && *c != '-' && *c != '.')
+    {
+        bail!("{what} contains invalid character `{ch}` -- only [a-zA-Z0-9_.-] are allowed");
+    }
+    Ok(())
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Restricting `machine.id` and label keys/values to `[a-zA-Z0-9_.-]` closes the unquoted-interpolation footgun in the SSH command `cluster up` builds, and matches the charset dora already enforces for node/data ids. Leaving `host`/`user` unrestricted is reasonable since they form the ssh target rather than the remote command string and legitimately carry IPv6 literals. The added tests exercise the real `ClusterConfig::load` path for both rejection and acceptance cases.

---

phil-opp:

The `machine.id` / label validation itself looks correct, and the added tests drive the real `ClusterConfig::load` path for both the rejected and accepted cases. But I think the fix is incomplete:

- **`zenoh_peer` is left unvalidated even though it is interpolated into the same remote shell command.** `zenoh_peer` (a free-form `Option<String>` in `config.rs:23`) is folded into the daemon command via `format_zenoh_peer_arg` → `format!(" --zenoh-peer {ep}")` (`mod.rs:86`), and that string is embedded **unquoted** into exactly the command this PR is hardening — `up.rs:70` (`... {zenoh_peer_arg}{labels} ... > /tmp/dora-daemon-{id}.log`) and the systemd `ExecStart` line at `install.rs:50`. So a config such as `zenoh_peer: "tcp/1.2.3.4:5456 ; rm -rf /"` (or a `$(...)`/backtick) reproduces the very command-injection / word-splitting this PR closes for `id` and `labels`. The PR description calls out only `host`/`user` as the intentionally-unrestricted fields, so `zenoh_peer` appears to have been overlooked rather than deliberately excluded. It runs the same `validate_shell_safe` gauntlet — though note the current charset `[a-zA-Z0-9_.-]` rejects `/` and `:`, which a zenoh endpoint like `tcp/10.0.0.1:5456` legitimately needs, so it would need its own slightly wider (but still whitespace-/metacharacter-free) charset rather than the identifier one.

Minor, pre-existing and out of scope for this PR: `host` is passed to `ssh` as a separate argv rather than through a shell, so shell metacharacters there are inert — but a `host` value beginning with `-` could still be parsed by `ssh` as an option. Noting only for completeness.

---

phil-opp:
The charset restriction is right and nothing in `docs/distributed-deployment.md` or `guide/src/operations/distributed.md` regresses — every documented id and label is inside the new set.

But the same threat model is still open in the same file: `zenoh_peer` is interpolated verbatim into the identical remote command (`binaries/cli/src/command/cluster/up.rs:62,66`, via `format_zenoh_peer_arg` / `cluster/mod.rs:84-88`) and `validate()` never touches it. So `zenoh_peer: "tcp/1.2.3.4:7447;rm -rf /"` still reaches the remote shell, and the PR's stated goal is not met until that field is validated too.

No conflict with #3061 (it touches `command/daemon.rs` only). The one interaction is that #3061's whitespace-trim becomes unreachable for labels arriving via `dora cluster up`, since this PR rejects them earlier.

phil-opp:
Good catch — fixed in b7a821a.

`zenoh_peer` is now validated in `ClusterConfig::validate()`. It can't reuse the id/label charset, since a legitimate endpoint (`tcp/1.2.3.4:7447`, or an IPv6 literal `tcp/[::1]:7447`) carries `/`, `:`, and `[`/`]`. So there's a sibling `validate_endpoint_shell_safe` helper that permits exactly those endpoint characters on top of the safe id set while still rejecting whitespace and shell metacharacters — `zenoh_peer: "tcp/1.2.3.4:7447;rm -rf /"` now fails validation up front. Added a rejection test for that payload and an accept test for `tcp/[::1]:7447`.

On the `host`/`user` distinction: I left those unvalidated deliberately (they're folded into the ssh *target* argv, not the remote command string, and legitimately carry IPv6 literals) — noted in the code comment next to the id/label check so the asymmetry is intentional rather than an oversight. Happy to tighten those too if you'd prefer a uniform rule.

Re #3061: agreed, its whitespace-trim becomes unreachable for labels arriving via `dora cluster up` — this PR rejects a whitespace-bearing label before the trim would run. That seems like the right precedence (reject at config-load rather than silently trim), but flagging it in case you'd rather the trim win; no code interaction either way since #3061 only touches `command/daemon.rs`.

---

phil-opp:
The merge-queue failure here isn't from this PR — it's a compile error in a **co-batched** PR.

The failing job (`Test (ubuntu-latest)`) died on:

```
```

That's in `dora-coordinator`, which this PR never touches — #2861's entire diff is `binaries/cli/src/command/cluster/config.rs`. A couple of confirming facts:

- `handlers.rs` is 867 lines on this branch and under 900 on current `main`, so **line 911 doesn't exist in either** — the extra lines come from another change in the batch. The merge-queue test branch is named `trunk-merge/pr-2910/...`, and it merged commit `87ab2c3` (not this PR's head), so the queue is testing #2861 batched behind another PR whose coordinator change left a `RunningDataflow` initializer without the new `ready_barrier_released` field.
- `main` didn't modify any file this PR touches, and the CLI crate compiles cleanly on this branch (`cargo test -p dora-cli`, `clippy -D warnings`, all green).

So #2861 itself is green; it's just "waiting for other pull requests to finish testing" behind the broken batch member. Once that PR is fixed or dropped, the queue should re-test #2861 clean — no change needed here.

This branch is a fair way behind `main` (~135 commits, though with no overlap on the files it changes, so trunk built the merge commit without conflict). Happy to rebase it onto latest `main` if you'd prefer, but I've held off so as not to eject the PR from the queue you just submitted it to.

---

PR #2864, opened 2026-07-28
fix(ros2-bridge): don't panic on zero-field message types (std_msgs/Empty)

### LINKED ISSUE #2804: ros2-bridge: empty message types (std_msgs/Empty) panic the CDR->Arrow deserializer and kill the DDS participant
## Summary

Deserializing a ROS 2 message type with **zero fields** (e.g. `std_msgs/msg/Empty`) panics in the ros2-bridge Arrow conversion:

```
thread '<unnamed>' panicked at arrow-array-59.0.0/src/array/struct_array.rs:91:46:
called `Result::unwrap()` on an `Err` value: InvalidArgumentError("use StructArray::try_new_with_length or StructArray::new_empty_fields to create a struct array with no fields so that the length can be set correctly")
```

Because the deserializer runs inside the subscription take path while a RustDDS cache lock is held, the panic **poisons the DDS cache mutex**, after which the participant event loop dies too:

```
thread 'RustDDS Participant 2 event loop' panicked at rustdds-0.13.1/src/structure/dds_cache.rs:103:30:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```

From that point **every subscription on the participant is silently dead** (0 Hz on all topics) — one stray `std_msgs/Empty` message takes down the whole bridge.

## Reproduction

Subscribe to any `std_msgs/msg/Empty` topic through `Ros2Context`/`create_subscription` (observed at commit de261f77, with the workspace's rustdds bumped to 0.13.1; the panic site is version-independent) and publish one message, e.g. `ros2 topic pub --once /rosbag/stop std_msgs/msg/Empty`.

## Root cause

`libraries/extensions/ros2-bridge/arrow/src/deserialize/mod.rs` builds the result with `let struct_array: StructArray = fields.into();` — `StructArray::from(Vec<...>)` panics when the vec is empty because the length would be ambiguous. Same pattern in `serialize/defaults.rs` (`default_struct_values`). Notably the test code in `serialize/defaults.rs` already uses `StructArray::new_empty_fields(1, None)`; the production paths don't.

## Fix

Guard both sites:

```rust
let struct_array: StructArray = if fields.is_empty() {
    StructArray::new_empty_fields(1, None)
} else {
    fields.into()
};
```

We are carrying exactly this as a build-time patch and verified it live: with the guard, 17 `std_msgs/Empty` messages deserialize cleanly (0 panics) in a 43-topic bridge run that previously wedged to all-topic 0 Hz. Happy to open a PR if useful.

Related: #2801 (same deployment context).

phil-opp:
## Summary

Fixes #2804.

Deserializing a ROS 2 message type with **zero fields** (e.g. `std_msgs/msg/Empty`) panicked in the ros2-bridge CDR→Arrow conversion:

```
called `Result::unwrap()` on an `Err` value: InvalidArgumentError("use StructArray::try_new_with_length or StructArray::new_empty_fields to create a struct array with no fields so that the length can be set correctly")
```

`StructArray::from(Vec<..>)` panics on an empty field vec because the array length is ambiguous with no columns. Since the deserializer runs inside the subscription take path **while a RustDDS cache lock is held**, the panic poisoned the DDS cache mutex and then killed the participant event loop — silently taking down *every* subscription on the participant (0 Hz on all topics) after a single stray `Empty` message.

## Changes

- `deserialize/mod.rs` (the take path) and `serialize/defaults.rs` (default-value construction) now build a length‑1 empty‑fields struct via `StructArray::new_empty_fields(1, None)` when there are no fields, instead of `fields.into()`. This mirrors the pattern already used in the crate's default-value test code.
- Added a regression test `empty_message_deserializes` that deserializes a zero‑field message and asserts it produces a single struct row with no columns.

## Testing

- `cargo test -p dora-ros2-bridge-arrow` — 16 passed, incl. the new regression test
- `cargo clippy -p dora-ros2-bridge-arrow -- -D warnings` — clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01PpY25UTvkyfgyTaJYcAQ7v)_

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/deserialize/mod.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/deserialize/mod.rs
@@ -217,7 +217,15 @@ impl<'de> serde::de::Visitor<'de> for StructVisitor<'_> {
             ));
         }
 
-        let struct_array: StructArray = fields.into();
+        // A message type with zero fields (e.g. `std_msgs/msg/Empty`) yields an
+        // empty `fields` vec. `StructArray::from(Vec<..>)` panics on an empty vec
+        // because the array length is ambiguous, so build a length-1 empty-fields
+        // struct explicitly instead.
+        let struct_array: StructArray = if fields.is_empty() {
+            StructArray::new_empty_fields(1, None)
+        } else {
+            fields.into()
+        };
 
         Ok(struct_array.into())
     }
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. `StructArray::new_empty_fields(1, None)` is the right way to build a zero-column struct with a defined length, and both the take-path (`deserialize/mod.rs`) and default-construction path (`serialize/defaults.rs`) are covered. The regression test exercises the real deserializer end-to-end on an empty CDR body and asserts a single zero-column row, so it genuinely guards the previously-panicking path.

---

PR #2872, opened 2026-07-28
fix(core): resolve module outputs produced by operator/custom inner nodes

### LINKED ISSUE #2817: Module expansion rejects a declared output produced by an operator/custom inner node (output-side counterpart of #2441)

keirsalterego:
## Summary

Fixes #2817.

Module output resolution in `expand_module_node` and `check_module_file` only consulted `Node::outputs`, the node-level output set. A runtime node (`operators:` / `operator:`) or a legacy `custom:` inner node declares its outputs in `OperatorConfig::outputs` / `NodeRunConfig::outputs` instead, which stays empty at the node level. A module whose declared output came from one of these was rejected with "no inner node produces it," and any downstream node wiring that output then failed to resolve too.

This is the output-side counterpart of #2441, which fixed the same gap on the input side by routing operator/custom `config.inputs` through `node_input_maps`.

## Changes

- Add `node_output_refs(node)`, mirroring `node_input_maps`, returning `(output_name, output_ref)` pairs for every output a node produces across all its forms.
- Use it in the Phase-3 producer lookup in `expand_module_node`, and in `check_module_file`'s output-declaration check.
- `docs/modules.md`: document that an inner node can produce a declared output from node-level `outputs:`, an `operator:`/`operators:` block, or a legacy `custom:` block.

## Why not just node id -> output name

An earlier attempt at this fix (#2818) maps `declared_output -> producer_node_id` and always uses the bare declared name as the consumer-facing reference. That works for a plain node or a single `operator:` node, but not for a multi-operator `operators:` node: `check_wiring` requires those to be addressed as `<node>/<operator_id>/<output_id>`, and a map with no operator id can't distinguish two operators on the same node that happen to produce differently-scoped outputs. Verified this concretely — building #2818's own `operators:` test case and running `check_wiring` on the result fails with:

```
input `sink/y` references output `result` of node `m.runtime`, which is a runtime
node; runtime node outputs must include the operator id (expected format:
`m.runtime/<operator_id>/<output_id>`)
```

`node_output_refs` avoids this by keeping the `<operator_id>/` segment in the `output_ref` for each `operators:` entry, while leaving node-level, `custom:`, and single-`operator:` outputs bare (matching how `resolve_aliases_and_set_defaults` injects the `op/` prefix for the single-operator case afterward).

## Tests

- `expand_resolves_operator_produced_module_output` — multi-operator `operators:` producer; asserts the consumer's rewritten reference is `<node>/<operator_id>/<output>` and passes `check_wiring`.
- `expand_resolves_single_operator_produced_module_output` — single `operator:` producer; reference stays bare.
- `expand_resolves_custom_produced_module_output` — legacy `custom:` producer.
- `check_module_file_accepts_operator_and_custom_produced_outputs` — the isolated linter accepts outputs from both, and still rejects a declared output nothing produces.

`cargo test -p dora-core` (36 tests in the `descriptor::expand` module), `cargo clippy -p dora-core -- -D warnings`, and `cargo fmt --all -- --check` all pass.

### DIFF
--- a/docs/modules.md
+++ b/docs/modules.md
@@ -151,7 +151,7 @@ Parameters are also injected as environment variables (`PARAM_SPEED`, `PARAM_MOD
 2. Prefix all internal node IDs with `{module_id}.` (e.g., `nav_stack.planner`)
 3. Replace `_mod/port_name` references with the actual sources from the parent's input map
 4. Rewrite internal cross-references (e.g., `planner/path` becomes `nav_stack.planner/path`)
-5. Map module-declared outputs to internal node outputs, so `nav_stack/cmd_vel` resolves to `nav_stack.controller/cmd_vel`
+5. Map module-declared outputs to internal node outputs, so `nav_stack/cmd_vel` resolves to `nav_stack.controller/cmd_vel`. An inner node may produce a declared output from its node-level `outputs:`, from an `operator:`/`operators:` block, or from a legacy `custom:` block. For an output produced by one operator of a multi-operator `operators:` node, the resolved reference keeps the operator segment that runtime nodes require: `nav_stack/cmd_vel` resolves to `nav_stack.runtime/controller/cmd_vel`
 6. Replace the module node with the expanded flat nodes
 7. Substitute `params:` values in `args:` fields and inject as env vars
 
@@ -231,6 +231,7 @@ This checks:
 - Valid YAML structure
 - Module header is present with `name`, `inputs`, `outputs`
 - All `_mod/` references correspond to declared inputs or optional inputs
+- Every declared output is produced by some inner node (counting `operator:`/`operators:` and legacy `custom:` outputs)
 - No duplicate node IDs
 - Internal wiring is consistent
 
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -46,6 +46,12 @@ const MAX_MODULE_FILE_SIZE: u64 = 1_048_576;
 /// Usage in module YAML: `_mod/port_name`
 const MODULE_INPUT_SOURCE: &str = "_mod";
 
+/// Where the outputs a module declares end up after expansion: maps each
+/// declared output name to the mapping a consumer must use to read it — the
+/// prefixed inner node ID plus the output name as that node addresses it
+/// (see [`node_output_refs`]).
+type ModuleOutputMap = BTreeMap<String, UserInputMapping>;
+
 /// Header section of a module definition file.
 #[derive(Debug, Clone, Deserialize)]
 struct ModuleHeader {
@@ -112,7 +118,7 @@ pub fn expand_modules_with_boundaries(
         .with_context(|| format!("failed to resolve base directory: {}", base_dir.display()))?;
     let mut seen = HashSet::new();
     let mut flat_nodes = Vec::new();
-    let mut output_maps: BTreeMap<String, BTreeMap<String, String>> = BTreeMap::new();
+    let mut output_maps: BTreeMap<String, ModuleOutputMap> = BTreeMap::new();
     let mut boundaries = ModuleBoundaries::default();
 
     for node in &descriptor.nodes {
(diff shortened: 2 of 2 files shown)


phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review (no human in the loop), posted as an outside contributor would. Advisory only; may contain mistakes.

No issues found.

`node_output_refs` cleanly mirrors the existing `node_input_maps` helper: it enumerates node-level `outputs`, `operators[].config.outputs`, single `operator.config.outputs`, and legacy `custom.run_config.outputs`. The only case that keeps an operator segment (`<op_id>/<output>`) is the multi-operator runtime node — which is exactly what `check_wiring` requires — while single-`operator:`, `custom:`, and node-level outputs stay bare so `resolve_aliases_and_set_defaults` doesn't double the `op/` prefix. The field accesses line up with the descriptor types, and every consumer of the widened `ModuleOutputMap` is updated.

The added tests are meaningful rather than vacuous: each runs `check_wiring` on the *expanded* dataflow and asserts the exact rewritten consumer mapping (including the multi-operator case keeping the operator segment), and the linter test still asserts the negative case (a declared output nothing produces is rejected). This also resolves the multi-operator resolution gap that was left open in #2818.

---

phil-opp:
Thanks!

PR #2873, opened 2026-07-28
fix(replay-node): skip an undecodable payload instead of aborting the whole replay

phil-opp:
## Summary

> ⚠️ **Machine-generated PR.** This change was written by Claude (Claude Code) as part of an automated code-review pass. Please review carefully before merging.

The replay loop in `binaries/replay-node/src/main.rs` deliberately logs-and-continues when a record's bincode `InterDaemonEvent` fails to deserialize:

```rust
let timestamped = match bincode::deserialize(&entry.event_bytes) {
    Ok(event) => event,
    Err(e) => { eprintln!("warning: failed to deserialize event ..."); continue; }
};
```

…but the *next* step decoded the inner Arrow IPC payload with `?`:

```rust
let array = match &data {
    Some(bytes) => decode_arrow_ipc(bytes).wrap_err("failed to decode ...")?, // aborts everything
    None => NullArray::new(0).into(),
};
```

So a single record whose framing and bincode survived, yet whose inner Arrow IPC body bytes are corrupt, terminated replay of **every remaining valid message**.

## Fix

Mirror the bincode arm: warn and `continue` on a decode failure, so one bad payload skips only its own message.

## Why this is the right behavior

- **Consistency with the sibling arm.** The bincode-deserialize step directly above already skips-and-continues; the IPC-decode step is the only one that aborted.
- **Consistency with the reader.** `RecordingReader::next_entry` (`libraries/recording/src/lib.rs`) goes out of its way to tolerate a torn trailing record and stop gracefully rather than error, precisely so that every fully-written record still replays. Aborting on one decodable-framing/undecodable-body record undoes that intent.

`send_output(...)?` on the next line remains a `?` — a broken daemon connection is a legitimate hard failure, unlike a single corrupt payload.

## Validation

- `cargo fmt -p dora-replay-node`, `cargo clippy -p dora-replay-node -- -D warnings` — clean.
- `cargo test -p dora-replay-node` — 4 passed.
- Reviewed against `origin/main` (`ffa6fd1`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01U4UYiwoYzdxW5ay5UV9kZe)_

### DIFF
--- a/binaries/replay-node/src/main.rs
+++ b/binaries/replay-node/src/main.rs
@@ -88,9 +88,24 @@ fn main() -> eyre::Result<()> {
                     // stream (or absent for metadata-only messages). Decode it
                     // back to an array and re-send; `send_output` re-encodes it
                     // into a fresh IPC stream on the wire.
+                    //
+                    // Skip-and-continue on a single undecodable payload rather
+                    // than aborting the whole replay, matching the bincode arm
+                    // above and the reader's own torn-record tolerance
+                    // (`RecordingReader::next_entry` stops gracefully at a torn
+                    // trailing record instead of erroring). Otherwise one corrupt
+                    // record would drop every remaining valid message.
                     let array = match &data {
-                        Some(bytes) => decode_arrow_ipc(bytes)
-                            .wrap_err("failed to decode recorded Arrow IPC payload")?,
+                        Some(bytes) => match decode_arrow_ipc(bytes) {
+                            Ok(array) => array,
+                            Err(e) => {
+                                eprintln!(
+                                    "warning: failed to decode recorded Arrow IPC payload for {}/{}: {e}",
+                                    entry.node_id, entry.output_id
+                                );
+                                continue;
+                            }
+                        },
                         None => NullArray::new(0).into(),
                     };
                     node.send_output(output_id, metadata.parameters, make_array(array))


phil-opp:

> ⚠️ Machine-generated. This is an autonomous code-review run by Claude (Claude Code) against `origin/main` (`ffa6fd1`). Each PR is a draft; please review carefully.

This codebase is exceptionally hardened — three independent correctness sweeps (core/message, daemon/coordinator/cli, node/operator/extensions) surfaced essentially no reachable bugs, so this run is deliberately small. Every finding below was re-verified against current `origin/main`; the combined diff passed `/review`, `/simplify`, the affected-crate test suites (`dora-core` 269 + 3 doctests, `dora-replay-node` 4, `dora-record-node` build), and `fmt`/`clippy -D warnings`.

### PRs opened by category

**Correctness / robustness**
- #2873 — `fix(replay-node)`: a single record whose bincode survived but whose inner Arrow IPC body was corrupt aborted the *entire* replay via `?`, while the sibling bincode arm and `RecordingReader` both skip-and-continue. Now warns and skips just that message.

**Optimization**
- #2874 — `perf(record-node)`: every recorded message went through the official IPC writer (~2 payload copies) + a 3rd copy into the aligned `AVec`. Switched to the 1-copy `ipc_encode` fast path (with official-writer fallback for non-eligible types), mirroring the live `DoraNode::send_output_array` path. Wire format unchanged.

**Documentation**
- #2875 — `docs(core)`: added a compile-checked doctest to `parse_urn` pinning the success shape and the silent `None` rejection cases (unclosed `[`, empty `[]`, empty key/value).

**Simplification** — none this run.

### Rejected findings (verified, not acted on)

| Finding | Why rejected |
|---|---|
| Delete unused `pub fn types_match` (dora-core `types.rs`) | It's a documented mutation-testing case study referenced in `.cargo/mutants.toml` + several QA docs; deletion would break those. Its overlap with `is_compatible` is an intentional parse-once optimization, not accidental duplication. |
| `types_match` returns "compatible" for same-base URNs with disjoint param keys | Likely the intended "each side constrains only what it names" semantics; affects only a type-suggestion path, not memory/data integrity — needs maintainer confirmation, not a unilateral change. |
| Daemon per-node metrics process-tree walk has no visited-set (`lib.rs`) | `children_map` is a forest under any real OS snapshot (acyclic); no reachable cycle exists, so no unbounded loop. Not a real-world bug. |
| Reduce record-node's 4× per-message `String` clones of the source ids | Negligible (small ids), and folding an unrelated micro-opt would dilute the focused encode PR. |
| Extract a shared `encode_ipc_aligned` helper (record-node vs `send_output_array`) | The two sites use different sinks (SHM `DataSample` vs heap `AVec`) so can't share one helper; adding a new `pub` API to the published `dora-node-api` crate for a single binary consumer isn't warranted. |
| record-node double `prepare()` + zero-fill of the `AVec` | Mirrors the established `send_output_array` / `allocate_data_sample` pattern; the flatbuffer rebuild is cheap next to the payload copy the change already removes. Net win stands. |

---

phil-opp:

The fix looks correct. The Arrow-IPC decode arm at `binaries/replay-node/src/main.rs:97-107` now mirrors the bincode arm directly above it (warn to stderr + `continue`), so a single record with intact framing but a corrupt IPC body only skips its own message instead of aborting the remaining valid records — and `send_output(...)?` on the next line stays fatal, which is the right split. The `continue` is in the entry-iteration loop and no partial output or accumulator state is left behind by a skip.

One suggestion: there's no regression test pinning this behavior change. A test that replays two entries — one with a deliberately corrupt Arrow-IPC payload followed by a valid one — and asserts the valid record still replays would lock in the fix; the tests referenced in the description are pre-existing and don't exercise this path, so a future refactor could silently reintroduce the `?`-abort.

---

phil-opp:
Closing as a duplicate of #3050, which makes the same skip-instead-of-abort fix and is the stronger version: it extracts the decode into a testable `decode_recorded_payload` helper, prints the full error chain (`{e:#}`) in the skip warning, and adds three unit tests where this PR has none.

PR #2874, opened 2026-07-28
perf(record-node): encode recorded payloads via the 1-copy Arrow IPC fast path

phil-opp:
## Summary

> ⚠️ **Machine-generated PR.** This change was written by Claude (Claude Code) as part of an automated code-review pass. Please review carefully before merging.

Every recorded message in `binaries/record-node/src/main.rs` was Arrow-IPC-encoded through the official writer (`encode_arrow_ipc`), which stages the record-batch body in an internal `Vec` — at least two payload copies — and the result was then copied a **third** time into the aligned `AVec`:

```rust
let ipc_bytes = arrow_utils::encode_arrow_ipc(&arrow_data)?; // official writer, ~2 copies
Some(AVec::from_slice(128, &ipc_bytes))                       // + a 3rd copy
```

This runs on **every recorded message** (e.g. every camera frame), so the extra copies add up on high-rate recordings.

## Fix

Use the hand-rolled 1-copy fast path (`ipc_fast_path_len` + `encode_ipc_into`) when the array type is eligible — it copies each buffer straight into the aligned target — and fall back to the official writer only for non-fast-path types (dictionary / `*View` / union / map):

```rust
let encoded: AVec<u8, ConstAlign<128>> =
    match arrow_utils::ipc_encode::ipc_fast_path_len(&arrow_data) {
        Some(len) => {
            let mut buf = AVec::__from_elem(128, 0, len);
            arrow_utils::ipc_encode::encode_ipc_into(&arrow_data, &mut buf)?;
            buf
        }
        None => AVec::from_slice(128, &arrow_utils::ipc_encode::encode_ipc_to_vec(&arrow_data)?),
    };
```

This mirrors `DoraNode::send_output_array`, which already uses exactly this fast-path/fallback dispatch on the live send path.

## Wire format is unchanged

The fast path emits a normal Arrow IPC stream. Its `ipc_encode` test suite proves round-trip equality against both the official `StreamReader` and the zero-copy decoder, and `replay-node` decodes recorded payloads via the same official `decode_arrow_ipc` path — so recordings written by the new code decode identically. Non-eligible types still go through the official writer, so nothing regresses for them.

## Validation

- `cargo fmt -p dora-record-node`, `cargo clippy -p dora-record-node -- -D warnings` — clean.
- `cargo test -p dora-record-node` — builds (thin binary, no unit tests).
- Full workspace test suite run on the combined review branch.
- Reviewed against `origin/main` (`ffa6fd1`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01U4UYiwoYzdxW5ay5UV9kZe)_

### DIFF
--- a/binaries/record-node/src/main.rs
+++ b/binaries/record-node/src/main.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, fs::File, time::SystemTime};
 
-use aligned_vec::AVec;
+use aligned_vec::{AVec, ConstAlign};
 use dora_message::{
     common::Timestamped,
     daemon_to_daemon::InterDaemonEvent,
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The new fast-path/fallback dispatch mirrors `DoraNode::send_output_array` (`apis/rust/node/src/node/mod.rs:1001`) exactly: `ipc_fast_path_len` returns the required length, `AVec::__from_elem(128, 0, len)` allocates precisely that, and `encode_ipc_into` enforces `dst.len() == ipc_fast_path_len(array)`. The `None` fallback (`encode_ipc_to_vec`) is a thin wrapper over the same `encode_arrow_ipc` the old code used, so non-eligible types (dictionary / `*View` / union / map) stay byte-identical to before, and the `MAX_IPC_BYTES` guard still fires (oversize eligible arrays fall through `prepare` → `None` → `encode_arrow_ipc`, which errors). Wire format and the `DataType::Null`-empty `None` special case are unchanged.

---

PR #2854, opened 2026-07-27
feat(ros2-bridge): port to ros2-client 0.10.1 and RustDDS 0.14

bobdingAI:
Ports `dora-ros2-bridge` from **ros2-client 0.8.0 / rustdds `=0.11.4`** to
**ros2-client 0.10.1 / rustdds 0.14**, removing the `=` pin on RustDDS. One
upgrade, not two: ros2-client 0.10.1 declares `rustdds = "0.14"`.

No source changes were required. Every ros2-client symbol the bridge uses (~60
items) survives 0.8 → 0.10.1 — the public-surface diff is additions only
(`distributions`, `pub use rustdds`, `Node::stop_spinner`). The APIs that *were*
removed, `QosPolicyBuilder::property()` and the `rustdds::CdrEncodingSize`
re-export, were never referenced here.

`ros2-client` and `cdr-encoding` move into `[workspace.dependencies]` so the
bridge and its arrow helper cannot drift apart on the ROS distro.
`cdr-encoding` goes 0.10.2 → 0.11 to match what RustDDS depends on, leaving a
single copy in the lockfile.

Kept as two commits: the first is the original port (ros2-client 0.10.0 /
rustdds 0.13.1, with a Windows CI exclusion); the second moves it to
rustdds 0.14 and deletes that exclusion. Read the second commit alone to see
what changed.

### No Windows workaround needed

The `=0.11.4` pin was load-bearing: rustdds 0.11.5–0.13.x depended
unconditionally on `pnet`, whose `pnet_datalink` declares
`#[link(name = "Packet")]` under `#[cfg(windows)]`, so Windows linking needed
`Packet.lib` from the Npcap SDK ([RustDDS#375](https://github.com/Atostek/RustDDS/issues/375)).

That is fixed upstream. RustDDS [`92c9b117c`](https://github.com/Atostek/RustDDS/commit/92c9b117c)
replaced `pnet` with `netdev` (which uses `windows-sys` and needs no SDK); it
shipped in **rustdds 0.14.0**, #375 is closed, and **ros2-client 0.10.1**
followed with `rustdds ^0.14`. So this PR needs no CI exclusions and no Windows
caveat in the docs — `cargo tree -e normal --workspace --target
x86_64-pc-windows-msvc` resolves **zero `pnet*` crates**. That matters beyond
tidiness: excluding the bridge from Windows would have cancelled out #3092, the
`mio 0.6.23` debug-assertions fix that stopped the bridge test binary aborting
there.

[...]

PR #2878, opened 2026-07-29
fix(ros2-bridge): restore DDS service and action discovery

SunSunSun689:
## Summary

Fix DDS ROS2 service/action bridge behavior across ROS domains:

- read `ROS_DOMAIN_ID` when creating the ROS2 DDS context
- use RELIABLE KeepLast(10) defaults for service/action bridge endpoints while preserving explicit compatible QoS overrides
- improve service/action discovery diagnostics without treating locally-created client topics as proof that a server exists
- correct the YAML Fibonacci action examples to use the `sequence` feedback field
- add YAML service/action bridge example targets to release QA coverage

## Validation

- cargo check -p dora-ros2-bridge-node --no-default-features
- cargo test -p dora-ros2-bridge-node peer_failure_tests -- --nocapture
- cargo check -p dora-ros2-bridge --example yaml-bridge-service --example yaml-bridge-action --features ros2-examples
- cargo test -p yaml-bridge-action-goal -p yaml-bridge-action-server-handler -p yaml-bridge-service-requester -p yaml-bridge-service-handler
- cargo fmt --all -- --check
- cargo clippy -p dora-ros2-bridge-node -- -D warnings
- cargo clippy -p dora-ros2-bridge --example yaml-bridge-service --example yaml-bridge-action --features ros2-examples -- -D warnings
- cargo test -p dora-ros2-bridge-node --lib --tests
- cargo test -p dora-ros2-bridge --lib --tests


phil-opp:

The `ROS_DOMAIN_ID` parsing fix and the reliable service/action QoS default look correct and are covered by real unit tests. My main concern is making `pre-iron-gid` a **default** feature.

`rmw_dds_common/Gid` is 24 bytes pre-Iron (Foxy/Galactic/Humble) and 16 bytes on Iron and newer (Iron/Jazzy/Rolling/Kilted), and this is a compile-time selection. So turning `pre-iron-gid` on by default in `libraries/extensions/ros2-bridge/Cargo.toml` doesn't add Humble support alongside the existing behavior — it swaps it: default builds now emit 24-byte GIDs, which fixes Humble but reintroduces the same `sequence size exceeds remaining buffer` discovery mismatch for anyone on Iron or newer building with defaults. If Humble is the only officially supported distro that's fine, but if newer distros are meant to keep working out of the box, this probably needs to stay opt-in (or be selected some other way) rather than becoming the default.

Two smaller points:

- **`wait_for_action_goal_service` changes startup behavior.** The action bridge now `?`-propagates and hard-fails at startup if the goal service isn't discoverable within the wait loop, where before it created the client and just waited for goals. That's a stricter contract — a dataflow whose ROS action server comes up shortly after the bridge would now error at startup instead of waiting. Worth confirming the wait window is generous enough for that ordering.

- **The `default_ros2_bridge_build_uses_humble_compatible_gid_length` test is tautological.** It only asserts `cfg!(feature = "pre-iron-gid")`, which is always true under the new default features — it doesn't exercise any actual GID encoding, so it won't catch a real regression, only removal of the default flag.

(Also noticed the PR bundles a ~1.5k-line internal planning/diagnostics document under `docs/superpowers/plans/` that seems unrelated to the fix.)

---

phil-opp:

Thanks for the follow-up commit (`e393045`). It's mostly additive — expanded service/action discovery diagnostics, example-node cleanups, unit tests, plus new docs and loop scripts — and I don't see new correctness bugs introduced by it. But the three substantive points from the earlier review look **still outstanding**, since the new commit didn't touch the relevant files:

- **`pre-iron-gid` is still a default feature.** `libraries/extensions/ros2-bridge/Cargo.toml` still has `default = ["generate-messages", "pre-iron-gid"]`, and `binaries/ros2-bridge-node/Cargo.toml` now hard-enables it. GID length is a compile-time either/or, so default builds still emit 24-byte GIDs — which fixes Humble but reintroduces the `sequence size exceeds remaining buffer` mismatch for Iron/Jazzy/Rolling built with defaults. If newer distros are meant to keep working out of the box, this should stay opt-in.
- **The GID test is still tautological.** `default_ros2_bridge_build_uses_humble_compatible_gid_length` in `dds.rs` only asserts `cfg!(feature = "pre-iron-gid")`, which is always true under the new defaults; it exercises no actual GID encoding and would only catch removal of the default flag.
- **Stricter startup is now broader.** `wait_for_action_client_services` now `?`-propagates a hard failure at startup if the goal, result, *and* cancel services aren't all discovered within the retry window. That's an even stricter startup contract than before for action clients whose ROS server comes up slightly late.

Two smaller notes on the new commit:
- `docs/superpowers/plans/2026-07-28-ros2-bridge-strict-diagnostics.md` (~2.3k lines, an internal debugging log) plus the new manual-test-checklist and two 300+-line loop scripts look like out-of-scope artifacts for this fix.
- In the `goal-node` example, `GoalSendState` only clears `in_flight` on `on_result`; if a goal is rejected or produces no result, the node stays permanently stuck on "Goal still in flight, skipping tick". Minor since it's example code. Also worth confirming the `partial_sequence` → `sequence` feedback field rename matches the real `example_interfaces/action/Fibonacci` interface (both dora sides were changed consistently, so the intra-example test passes regardless).

---

phil-opp:
🤖 **Fully automated review by Claude** — no human in the loop.

The two latest commits (`dc8e949`, `b5aacaa`) address the main concern from the earlier reviews: `pre-iron-gid` is now a non-default opt-in feature in both `libraries/extensions/ros2-bridge/Cargo.toml` and `binaries/ros2-bridge-node/Cargo.toml`, so default builds no longer switch to the 24-byte Humble GID layout that would break Iron/Jazzy service discovery. The previously-flagged tautological GID test was also reworked to assert the feature stays opt-in, backed by a manifest-grep test. No new issues found in these commits.

The only still-open item from earlier reviews is the strict action-client startup (`wait_for_action_client_services` in `binaries/ros2-bridge-node/src/main.rs` hard-failing within ~20s if the goal/result/cancel services aren't all discovered), which now appears to be a deliberate, documented design choice rather than an oversight.

---

phil-opp:

The QoS half of this is a genuine root-cause fix and worth landing. Services were being created BEST_EFFORT, which cannot RxO-match rclcpp's RELIABLE service endpoints — that is the real reason discovery failed, and it would fail identically on Iron and Jazzy, so the "Humble-specific" framing in the title is not quite right.

The rest needs work, and one item is now a hard blocker that did not exist when this was opened.

## 1. This will not compile against current `main`

`#2854` has merged. `main` now pins:

```toml
ros2-client = { version = "0.10.1", default-features = false, features = ["humble"] }
```

and **ros2-client 0.10 removed `pre-iron-gid`** — distro selection became explicit Cargo features (`galactic`/`humble`/`iron`/`jazzy`/…). This PR still adds:

- `pre-iron-gid = ["ros2-client/pre-iron-gid"]` (`libraries/extensions/ros2-bridge/Cargo.toml`)
- `pre-iron-gid = ["dora-ros2-bridge/pre-iron-gid"]` (`binaries/ros2-bridge-node/Cargo.toml`)

GitHub reports this PR as MERGEABLE because the two sides touch different lines of the same files, so git sees no conflict — but cargo will fail immediately with `feature 'pre-iron-gid' in dependency 'ros2-client' does not exist`. The two tests that assert the feature stays opt-in (`libraries/extensions/ros2-bridge/src/transport/dds.rs`, and the `Cargo.toml`-string assertion in `binaries/ros2-bridge-node/src/main.rs`) break with it.

Since `#2854` already selects `humble` (24-byte Gid) workspace-wide, the GID contribution here is also redundant. **The whole GID portion should be removed** — the feature declarations, both tests, and the `.gitignore` line. That is the single change that unblocks this.

Worth noting separately: `#2854` hardcodes `humble` with no override, so the workspace is now Humble-pinned rather than distro-selectable. The feature-passthrough idea in this PR was the better design and is worth revisiting as its own follow-up — just expressed in 0.10's vocabulary (`ros-jazzy` / `ros-humble` passthrough) rather than `pre-iron-gid`.

## 2. The discovery fallback makes `wait_for_service` a no-op

`binaries/ros2-bridge-node/src/main.rs` (~:2122, and again ~:2195) — after the first 2 s timeout the code calls `discovered_topic_names(ros_context)` and, if `rq/<svc>Request` and `rr/<svc>Reply` are both present, prints `ROS2 service connected` and returns `Ok(())`.

Those two topics are created **by the bridge's own client**. RustDDS stores locally-defined topics in its discovery DB (`DiscoveryDB::update_topic_data_p` inserts them with `DiscoveredVia::SelfDefined` under `self.my_guid`), and neither `update_publication` nor `update_subscription` filters the local GUID prefix — the code explicitly tolerates self-reflection. `Context::discovered_topics()` only filters `DCPS*`.

Failure scenario: start a service-client dataflow with **no ROS 2 server running at all**. After 2 s the fallback fires, prints `ROS2 service connected: /add_two_ints`, and `run_service_client` proceeds; the first request is written to a topic with no matched reader and is silently dropped until the 30 s response timeout. So the retry window is dead code in exactly the case it exists for — a slow server — and the startup gate the docs advertise never gates anything. The same applies to all three endpoints in `wait_for_action_client_services`.

`Client::wait_for_service` already does the right check (`wait_for_reader` on the request writer's GUID plus `wait_for_writer` on the reply reader's GUID). Keep the discovered-topic list for the **error message** — that part is genuinely useful — and drop the early return.

## 3. `RosContextCreation::DefaultDomain` rests on a false premise

`main.rs` (~:2021-2046) — the comment states that "ros2-client's default constructor is not equivalent to `ContextOptions::domain_id(0)`". In ros2-client 0.8.1, `Context::new()` is `from_domain_participant(DomainParticipant::new(0))`, and rustdds 0.11.4's `DomainParticipant::new(id)` is literally `DomainParticipantBuilder::new(id).build()` — byte-for-byte what `with_options(...domain_id(0))` does.

So the enum, `ros_context_creation()`, and the test `default_domain_uses_ros2_client_default_context_constructor` encode an invariant that does not exist. It also contradicts the rest of the repo: `libraries/extensions/ros2-bridge/src/transport/dds.rs:47` already exposes `dds::Context::new(domain_id)` using `with_options` unconditionally, and `transport::Context::new(&config.transport, domain_id)` is the existing abstraction the YAML DDS path should use.

Relatedly, `parse_ros_domain_id` is the **fourth** copy of `env::var("ROS_DOMAIN_ID").ok().and_then(parse).unwrap_or(0)` in this one file (existing copies around `main.rs:106`, `:150`, `:273`, `:473`). Worth collapsing to one helper. It also maps garbage to 0 silently where rcl errors out; the repo's own Python binding does this properly (`libraries/extensions/ros2-bridge/python/src/lib.rs:144-163` — trims whitespace, errors on unparsable).

## 4. QoS override is all-or-nothing, and silently restores the bug

`main.rs` (~:2066, ~:2491) — `service_or_action_qos` is only applied when `is_default_topic_qos()` holds, which requires *every* field to be unset. So `qos: {keep_last: 20}` on a service node falls through to `config.to_rustdds_qos()` → `reliable: false` (the field is a plain `bool`, so "unset" and "explicitly best-effort" are indistinguishable) → BEST_EFFORT request writer → no RxO match → **the exact bug this PR fixes comes back**, silently, for any user who touches QoS at all.

Better to start from the service defaults and let explicit fields override, or make `reliable` an `Option<bool>`.

The status-QoS work is correct, for the record: `transient_local` + reliable matches `rcl_action_qos_profile_status_default` on both publisher and subscription, and feedback/goal/result/cancel reliable + KeepLast(10) matches `rmw_qos_profile_services_default`. Service/action topic mangling is also right for this stack.

## 5. Scope creep

Still in the diff:

- `docs/ros2-bridge-manual-test-checklist.md` — the title is literally `# ROS2 Bridge Manual Test Checklist[PASSED]` and it hardcodes `cd ~/Desktop/dora` in **9 places**. Not publishable docs.
- `.gitignore` — adds an ignore rule for one specific file under `docs/superpowers/plans/`, which is a **tracked** directory. Repo-wide config changed to hide one working note.
- `scripts/ros2-bridge-{service,action}-loop.sh` — 654 lines, competently written, but ROS/bash-only, not wired into `scripts/smoke-all.sh` or nightly, and unrunnable in CI.

## 6. Tests

The +1711 is hand-written, not generated bindings, and the unit tests do run — but **every one of them tests a helper this PR introduces**, so none could have failed before it. Nothing asserts the QoS actually handed to `create_client` / `create_action_client` is reliable, which is the bug being fixed. A cheap real regression test: assert `service_or_action_qos(&Ros2QosConfig::default())` yields `policy::Reliability::Reliable{..}` from the *rustdds* policies, not the intermediate config. That would have failed before the fix.

(The `partial_sequence` → `sequence` rename is correct — `example_interfaces/action/Fibonacci` uses `sequence` on both humble and rolling.)

## Suggested path

Rebase onto current `main`, drop the GID portion entirely (it is both broken and redundant after #2854), drop the discovery fallback's early return, and keep the QoS fix broadened per §4 plus the `ROS_DOMAIN_ID` plumbing routed through the existing `transport::Context::new`. That leaves a focused, landable PR. The distro-passthrough idea deserves its own follow-up against 0.10's feature names.

SunSunSun689:
Rebased onto the latest origin/main and pushed the updated feat branch. I removed the obsolete pre-iron-gid feature from both the binary manifest and libraries/extensions/ros2-bridge/Cargo.toml, and dropped the scope-creep files from the PR (manual checklist, loop scripts, and the README section that referenced them). Verified with rustfmt and targeted tests: cargo fmt --all -- --check passed, cargo test -p dora-ros2-bridge-node --lib --tests passed, and cargo test -p dora-ros2-bridge --lib --tests passed.

phil-opp:
**Re-review after the 2026-08-20 rebase**

The rebase resolved the compile blocker (`pre-iron-gid` is gone from both manifests) and the scope-creep files from the earlier review — thanks. Two substantive issues from that review are still present in the rebased diff:

**1. The discovery fallback in `wait_for_service` / `wait_for_action_client_services` returns `Ok(())` even when no server exists.** After the 2 s timeout the code calls `discovered_topic_names(ros_context)` and, if `rq/<svc>Request` and `rr/<svc>Reply` are both present, declares the service connected and returns. But those two topics are created by the bridge's *own* client endpoints (the request writer and the reply reader), and RustDDS records locally-defined topics in its discovery DB — so `Context::discovered_topics()` lists them regardless of whether any ROS server is running. Start a service-client dataflow with no server: after 2 s the fallback fires, prints `ROS2 service connected`, and `run_service_client` proceeds to write requests to a topic with no matched reader (silently dropped until the 30 s timeout). That makes the startup gate a no-op in exactly the slow-server case it exists to cover. `Client::wait_for_service` already does the correct reader/writer GUID match — I'd keep the discovered-topic list for the *error message* only and drop the early `Ok(())`.

**2. `service_or_action_qos` reverts to BEST_EFFORT the moment any QoS field is set.** The reliable-service default is applied only when `is_default_topic_qos()` holds, which requires *every* field to be unset. So `qos: {keep_last: 20}` (or any single field) on a service/action node falls through to `config.to_rustdds_qos()`, and since `reliable` is a plain `bool`, "unset" is indistinguishable from "explicitly best-effort" → a `Reliability::BestEffort` request writer → no RxO match with an rclcpp RELIABLE service → the discovery failure this PR fixes comes back silently for any user who touches QoS. I'd start from the service defaults and let explicit fields override, or make `reliable` an `Option<bool>`.

Minor (not a nit-only point since it undercuts the PR's own guarantee): none of the added unit tests assert the QoS actually handed to `create_client`/`create_action_client` is `Reliability::Reliable` — they all exercise newly-introduced string/helper functions, so none could have failed before this change. A test asserting `service_or_action_qos(&Ros2QosConfig::default())` yields a reliable rustdds policy would guard the actual fix.

---
🤖 _This is a fully automated review by Claude. It may contain mistakes — please verify before acting on it._

---

SunSunSun689:
Implemented the review follow-up on this branch:\n\n- removed the fallback that treated discovered rq/rr topics as a successful service/action readiness signal\n- changed service/action QoS to start from the bridge's default reliable service QoS and then apply explicit overrides, so partial QoS configs no longer fall back to best effort\n- added a regression test covering the reliability case\n\nValidated with  and .

SunSunSun689:
Implemented the review follow-up on this branch:

- removed the fallback that treated discovered rq/rr topics as a successful service/action readiness signal
- changed service/action QoS to start from the bridge's default reliable service QoS and then apply explicit overrides, so partial QoS configs no longer fall back to best effort
- added a regression test covering the reliability case

Validated with `cargo test -p dora-ros2-bridge-node --lib --tests` and `cargo fmt --all -- --check`.

phil-opp (commented):

The latest commit (`d833bd54`) resolves the two blocking issues raised in the earlier reviews:

- **Service/action QoS is now reliable by default.** `service_or_action_qos` starts from `default_service_qos_config()` (`reliable: true`, `keep_last: Some(10)`) and applies only the explicit overrides, never flipping `reliable` back to BEST_EFFORT — verified by `service_qos_keeps_reliable_when_overriding_other_fields`, which asserts `Reliability::Reliable` survives a `keep_last` override. This is what lets the bridge RxO-match rclcpp's RELIABLE endpoints on Humble.
- **`wait_for_service` no longer early-returns on the client's own topics.** `discovered_topic_names(ros_context)` is used only to build the `bail!` diagnostic; the sole `Ok(())` return is the real `client.wait_for_service` GUID match.

`pre-iron-gid` is dropped from the manifest and the scope is back to the bridge. No new issues found in this diff. The only remaining points are the previously-noted non-blocking cleanups (the `RosContextCreation` Default-vs-`domain_id(0)` distinction and the duplicated `parse_ros_domain_id` helper).

---

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human reviewed this comment.

Re-checked after the follow-up commit. Both blocking items from the 2026-08-20 review look addressed in the current diff:

- `wait_for_service` / `wait_for_action_client_services` no longer treat locally-discovered `rq/`/`rr/` topics as a successful readiness signal — `discovered_topic_names(ros_context)` is now used only inside the `eyre::bail!` error message after MAX_RETRIES.
- `service_or_action_qos` starts from `default_service_qos_config()` (`reliable: true, keep_last: 10`) and applies each explicit YAML field via `if let Some(...)`, so partial QoS configs like `{keep_last: 20}` keep RELIABLE instead of silently falling back to BEST_EFFORT. The new `service_qos_keeps_reliable_when_overriding_other_fields` test asserts against the actual `rustdds::policy::Reliability::Reliable` enum, so it would fail before the fix.

One small side effect worth noting (not a blocker): because `service_or_action_qos` never reads `config.reliable`, a user who explicitly sets `qos: {reliable: false}` on a service/action node will still get RELIABLE. That's the right ROS-service default, but silently ignoring an explicit `reliable: false` may deserve a doc line or a warning. Same applies to the `ROS_DOMAIN_ID` parser silently mapping unparseable values to 0.

---

phil-opp:
**Automated review by Claude — fully automated, no human in the loop.**

Reviewed the current head (`d833bd54`). The previously-flagged blockers (the ros2-client 0.10 compile break, the discovery-fallback no-op, the all-or-nothing QoS revert, scope creep) look resolved. One leftover from the `pre-iron-gid` removal slipped through, though — the rebase deleted the feature from the manifests but left the two tests that guard it, and both are now vacuous:

1. `libraries/extensions/ros2-bridge/src/transport/dds.rs`:
   ```rust
   #[test]
   fn default_ros2_bridge_build_keeps_pre_iron_gid_opt_in() {
       assert!(!cfg!(feature = "pre-iron-gid"));
   }
   ```
   `dora-ros2-bridge` no longer declares a `pre-iron-gid` feature, so `cfg!(feature = "pre-iron-gid")` is always `false` and `!false` can never fail. It also references an undeclared feature, which produces an `unexpected_cfgs` warning on `cargo test -p dora-ros2-bridge`.

2. `binaries/ros2-bridge-node/src/main.rs` — `binary_manifest_keeps_pre_iron_gid_opt_in`: asserts the `dora-ros2-bridge =` manifest line does **not** contain `"pre-iron-gid"`. After the rebase that dependency is `{ path = "...", default-features = false }`, so the string is gone and the assertion is trivially true forever.

Both tests are named `..._keeps_pre_iron_gid_opt_in` but the opt-in they describe no longer exists, so they assert nothing about real behavior and are misleading to a future reader. Suggest deleting both (the `dds.rs` `mod tests` block and this one test in `main.rs`), which is what the earlier review round asked for alongside the manifest cleanup.

For what it's worth, the rest of the added tests do look genuine — e.g. `service_qos_keeps_reliable_when_overriding_other_fields` asserts against the real `Reliability::Reliable` output of `service_or_action_qos` and would have failed before the QoS fix, and the `GoalSendState` / `int32_list_values` tests exercise real logic. Two smaller behaviors already noted by the prior review still stand (an explicit `qos: {reliable: false}` on a service/action node is silently forced back to reliable; a non-numeric `ROS_DOMAIN_ID` silently maps to 0) — worth at least a warn, but non-blocking.

---

SunSunSun689:
Removed the two leftover `pre-iron-gid` guard tests from the latest review:

- `binary_manifest_keeps_pre_iron_gid_opt_in`
- `default_ros2_bridge_build_keeps_pre_iron_gid_opt_in`

The `pre-iron-gid` feature no longer exists, so these checks were vacuous and could trigger stale `cfg`/manifest assumptions.

Validation run locally:

```text
cargo fmt --all -- --check
cargo test -p dora-ros2-bridge-node --lib --tests
cargo test -p dora-ros2-bridge --lib --tests
git diff --check
```

Pushed as `7d7fe40bc test(ros2-bridge): remove obsolete pre-iron gid guards`.

phil-opp:

Re-checked at head `7d7fe40`. The two leftover `pre-iron-gid` guard tests flagged last round are now removed (`default_ros2_bridge_build_keeps_pre_iron_gid_opt_in` in `dds.rs` and `binary_manifest_keeps_pre_iron_gid_opt_in` in `main.rs`), so that item is resolved. The new commit is test-only; the rest of the diff (reliable service/action QoS defaults, discovered topics used only for the failure diagnostic, `ROS_DOMAIN_ID` plumbing) is unchanged from the previously-verified state, and I found no new issues.

The two remaining non-blocking notes still stand and are optional: an explicit `qos: {reliable: false}` on a service/action node is silently forced back to reliable, and a non-numeric `ROS_DOMAIN_ID` silently maps to `0` — a warn line for each would be nice, but neither blocks.

---

SunSunSun689:
Addressed the two remaining optional review notes in `e4c7b9b3d`:

- Added a warning when `ROS_DOMAIN_ID` is set but cannot be parsed; the bridge still preserves the existing fallback to DDS domain `0`.
- Added a warning when a service/action bridge explicitly configures `qos: { reliable: false }`; ROS2 services/actions still keep RELIABLE QoS as required. The Zenoh transport paths now use the same `ROS_DOMAIN_ID` parser, so they no longer silently bypass the warning.

Validation run locally:

```text
rustup run 1.97.1 cargo fmt --all -- --check
RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc /home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/cargo test -p dora-ros2-bridge-node peer_failure_tests
git diff --check
```

I also ran:

```text
RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc /home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/cargo clippy -p dora-ros2-bridge-node -- -D warnings
```

phil-opp:
**Automated review (Claude)** — re-reviewed after the new commit `e4c7b9b` ("warn on ignored service qos and domain fallback").

No new issues found. The commit is warning-only and low-risk: it adds a warning when an explicit `qos: { reliable: false }` is overridden to reliable for service/action endpoints, and a warning (guarded by a one-shot atomic) when `ROS_DOMAIN_ID` is unparseable and falls back to 0 — directly addressing the two non-blocking notes from the previous review. It doesn't change discovery or QoS behavior.

Two minor, non-blocking observations on the new warning: `warn_explicit_service_action_best_effort` runs transport-agnostically in `main()`, so it also fires on the default `rmw-zenoh` path where the zenoh QoS mapping may actually honor `reliable: false` (slightly inaccurate message there); and a node configured with e.g. `qos: { reliable: false, keep_last: 20 }` will emit both this startup warning and the one inside `service_or_action_qos` (double-warn). Neither affects correctness.

Latest commit looks safe to merge.

_This is a fully automated review; a maintainer should still confirm before merging._

---

phil-opp:

Thanks for sticking with this through so many rounds — the current head is in good shape, and the two root causes you found are real. `Context::new()` really does hardcode domain 0 (it is `from_domain_participant(DomainParticipant::new(0))` in ros2-client 0.10.1), so any bridge under a non-zero `ROS_DOMAIN_ID` could never see a real ROS2 node; and BEST_EFFORT service endpoints cannot RxO-match rclcpp's RELIABLE ones. Both hit every distro, not just Humble. The two hard blockers from earlier rounds are gone from the current head.

One thing I would like fixed before merge: **the new best-effort warning misfires** (`binaries/ros2-bridge-node/src/main.rs:2099`). `if !config.reliable && !config.is_default_topic_qos()` is true for `qos: {keep_last: 20}`, because `Ros2QosConfig::reliable` is a plain `bool` defaulting to `false`. A user who only set a history depth gets told their "descriptor QoS uses topic-style best effort", which they never asked for. The genuinely explicit case is already caught by `warn_explicit_service_action_best_effort` on the raw JSON in `main()`, so the cleanest fix is to delete the warning block from `service_or_action_qos` entirely.

Two cleanups worth doing while you are in there:

- **`RosContextCreation` guards an invariant that does not exist.** The comment says "ros2-client's default constructor is not equivalent to `ContextOptions::domain_id(0)`", but I checked both: `Context::new()` is `from_domain_participant(DomainParticipant::new(0))`, `with_options` is `DomainParticipantBuilder::new(opt.domain_id).build()`, and rustdds' `DomainParticipant::new(id)` is literally `DomainParticipantBuilder::new(id).build()`. The security branch is inert here (`default-features = false`, no `security`). So the enum, `ros_context_creation()`, and the `default_domain_uses_ros2_client_default_context_constructor` test can all go, and `create_ros_context` collapses to one `with_options(...domain_id(domain_id))` call.
- **The new `rmw-zenoh` default feature in `binaries/ros2-bridge-node/Cargo.toml:12-16` gates no code** — there is no `#[cfg(feature = ...)]` anywhere in the crate — so `--no-default-features` becomes a configuration that cannot compile, since the zenoh transport is `#[cfg(feature = "rmw-zenoh")]` in the library. Nothing in CI builds that combination (`cargo hack --ignore-private` skips this `publish = false` crate), so it just sits there broken. I would revert to the inline `features = ["rmw-zenoh"]`.

Smaller: `service_topics_discovered` (`:2306`) is `#[cfg(test)]`-only and exists solely for its own test — either use it in the diagnostic path or drop both. Four call sites still spell out `parse_ros_domain_id(std::env::var("ROS_DOMAIN_ID").ok().as_deref())` even though `current_ros_domain_id()` exists at `:2269`. And `:2155`/`:2165`/`:2215`/`:2222` use `println!` where the rest of the binary is on `tracing`.

The last thing, and the reason this has been hard to land: **nothing in CI or release QA touches the YAML bridge's DDS path.** PR CI does `cargo check` + msg-gen with no ROS distro; the nightly `ros2-zenoh-humble`/`ros2-zenoh-kilted` jobs run `scripts/ros2-zenoh-interop.sh`, i.e. the zenoh transport, not the rustdds path this PR changes; and `scripts/ros2dev.sh qa` runs only the native code API examples. Could you add `yaml-bridge-service` and `yaml-bridge-action` as `[[example]]` targets in `libraries/extensions/ros2-bridge/Cargo.toml` and to the `EXAMPLES` array in `scripts/ros2dev.sh` (~line 115), so this is at least exercised at release QA? Failing that, a manual `ROS_DOMAIN_ID=23` run against `ros2 run examples_rclcpp_action_server fibonacci_action_server` pasted into the PR would do.

Kilted/Jazzy risk looks low now that the GID hunks are gone — the remaining changes are distro-neutral and the workspace is Humble-pinned by #2854. One thing to watch: `action_status_qos` forcing `transient_local` on the status subscription matches `rcl_action_qos_profile_status_default`, so it matches an rcl server, but it would stop matching a non-standard server publishing status VOLATILE. The `partial_sequence` to `sequence` rename in the examples is correct.

Also worth updating the PR title and body — they still describe enabling `pre-iron-gid`, which is no longer in the diff, and "Humble" is narrower than what this actually fixes.

phil-opp:

Thanks for sticking with this through so many rounds — the current head is in good shape, and the two root causes you found are real. `Context::new()` really does hardcode domain 0 (it is `from_domain_participant(DomainParticipant::new(0))` in ros2-client 0.10.1), so any bridge under a non-zero `ROS_DOMAIN_ID` could never see a real ROS2 node; and BEST_EFFORT service endpoints cannot RxO-match rclcpp's RELIABLE ones. Both hit every distro, not just Humble. The two hard blockers from earlier rounds are gone from the current head.

One thing I would like fixed before merge: **the new best-effort warning misfires** (`binaries/ros2-bridge-node/src/main.rs:2099`). `if !config.reliable && !config.is_default_topic_qos()` is true for `qos: {keep_last: 20}`, because `Ros2QosConfig::reliable` is a plain `bool` defaulting to `false`. A user who only set a history depth gets told their "descriptor QoS uses topic-style best effort", which they never asked for. The genuinely explicit case is already caught by `warn_explicit_service_action_best_effort` on the raw JSON in `main()`, so the cleanest fix is to delete the warning block from `service_or_action_qos` entirely.

Two cleanups worth doing while you are in there:

- **`RosContextCreation` guards an invariant that does not exist.** The comment says "ros2-client's default constructor is not equivalent to `ContextOptions::domain_id(0)`", but I checked both: `Context::new()` is `from_domain_participant(DomainParticipant::new(0))`, `with_options` is `DomainParticipantBuilder::new(opt.domain_id).build()`, and rustdds' `DomainParticipant::new(id)` is literally `DomainParticipantBuilder::new(id).build()`. The security branch is inert here (`default-features = false`, no `security`). So the enum, `ros_context_creation()`, and the `default_domain_uses_ros2_client_default_context_constructor` test can all go, and `create_ros_context` collapses to one `with_options(...domain_id(domain_id))` call.
- **The new `rmw-zenoh` default feature in `binaries/ros2-bridge-node/Cargo.toml:12-16` gates no code** — there is no `#[cfg(feature = ...)]` anywhere in the crate — so `--no-default-features` becomes a configuration that cannot compile, since the zenoh transport is `#[cfg(feature = "rmw-zenoh")]` in the library. Nothing in CI builds that combination (`cargo hack --ignore-private` skips this `publish = false` crate), so it just sits there broken. I would revert to the inline `features = ["rmw-zenoh"]`.

Smaller: `service_topics_discovered` (`:2306`) is `#[cfg(test)]`-only and exists solely for its own test — either use it in the diagnostic path or drop both. Four call sites still spell out `parse_ros_domain_id(std::env::var("ROS_DOMAIN_ID").ok().as_deref())` even though `current_ros_domain_id()` exists at `:2269`. And `:2155`/`:2165`/`:2215`/`:2222` use `println!` where the rest of the binary is on `tracing`.

The last thing, and the reason this has been hard to land: **nothing in CI or release QA touches the YAML bridge's DDS path.** PR CI does `cargo check` + msg-gen with no ROS distro; the nightly `ros2-zenoh-humble`/`ros2-zenoh-kilted` jobs run `scripts/ros2-zenoh-interop.sh`, i.e. the zenoh transport, not the rustdds path this PR changes; and `scripts/ros2dev.sh qa` runs only the native code API examples. Could you add `yaml-bridge-service` and `yaml-bridge-action` as `[[example]]` targets in `libraries/extensions/ros2-bridge/Cargo.toml` and to the `EXAMPLES` array in `scripts/ros2dev.sh` (~line 115), so this is at least exercised at release QA? Failing that, a manual `ROS_DOMAIN_ID=23` run against `ros2 run examples_rclcpp_action_server fibonacci_action_server` pasted into the PR would do.

Kilted/Jazzy risk looks low now that the GID hunks are gone — the remaining changes are distro-neutral and the workspace is Humble-pinned by #2854. One thing to watch: `action_status_qos` forcing `transient_local` on the status subscription matches `rcl_action_qos_profile_status_default`, so it matches an rcl server, but it would stop matching a non-standard server publishing status VOLATILE. The `partial_sequence` to `sequence` rename in the examples is correct.

Also worth updating the PR title and body — they still describe enabling `pre-iron-gid`, which is no longer in the diff, and "Humble" is narrower than what this actually fixes.

phil-opp:

Thanks for sticking with this through so many rounds — the current head is in good shape, and the two root causes you found are real. `Context::new()` really does hardcode domain 0 (it is `from_domain_participant(DomainParticipant::new(0))` in ros2-client 0.10.1), so any bridge under a non-zero `ROS_DOMAIN_ID` could never see a real ROS2 node; and BEST_EFFORT service endpoints cannot RxO-match rclcpp's RELIABLE ones. Both hit every distro, not just Humble. The two hard blockers from earlier rounds are gone from the current head.

One thing I would like fixed before merge: **the new best-effort warning misfires** (`binaries/ros2-bridge-node/src/main.rs:2099`). `if !config.reliable && !config.is_default_topic_qos()` is true for `qos: {keep_last: 20}`, because `Ros2QosConfig::reliable` is a plain `bool` defaulting to `false`. A user who only set a history depth gets told their "descriptor QoS uses topic-style best effort", which they never asked for. The genuinely explicit case is already caught by `warn_explicit_service_action_best_effort` on the raw JSON in `main()`, so the cleanest fix is to delete the warning block from `service_or_action_qos` entirely.

Two cleanups worth doing while you are in there:

- **`RosContextCreation` guards an invariant that does not exist.** The comment says "ros2-client's default constructor is not equivalent to `ContextOptions::domain_id(0)`", but I checked both: `Context::new()` is `from_domain_participant(DomainParticipant::new(0))`, `with_options` is `DomainParticipantBuilder::new(opt.domain_id).build()`, and rustdds' `DomainParticipant::new(id)` is literally `DomainParticipantBuilder::new(id).build()`. The security branch is inert here (`default-features = false`, no `security`). So the enum, `ros_context_creation()`, and the `default_domain_uses_ros2_client_default_context_constructor` test can all go, and `create_ros_context` collapses to one `with_options(...domain_id(domain_id))` call.
- **The new `rmw-zenoh` default feature in `binaries/ros2-bridge-node/Cargo.toml:12-16` gates no code** — there is no `#[cfg(feature = ...)]` anywhere in the crate — so `--no-default-features` becomes a configuration that cannot compile, since the zenoh transport is `#[cfg(feature = "rmw-zenoh")]` in the library. Nothing in CI builds that combination (`cargo hack --ignore-private` skips this `publish = false` crate), so it just sits there broken. I would revert to the inline `features = ["rmw-zenoh"]`.

Smaller: `service_topics_discovered` (`:2306`) is `#[cfg(test)]`-only and exists solely for its own test — either use it in the diagnostic path or drop both. Four call sites still spell out `parse_ros_domain_id(std::env::var("ROS_DOMAIN_ID").ok().as_deref())` even though `current_ros_domain_id()` exists at `:2269`. And `:2155`/`:2165`/`:2215`/`:2222` use `println!` where the rest of the binary is on `tracing`.

The last thing, and the reason this has been hard to land: **nothing in CI or release QA touches the YAML bridge's DDS path.** PR CI does `cargo check` + msg-gen with no ROS distro; the nightly `ros2-zenoh-humble`/`ros2-zenoh-kilted` jobs run `scripts/ros2-zenoh-interop.sh`, i.e. the zenoh transport, not the rustdds path this PR changes; and `scripts/ros2dev.sh qa` runs only the native code API examples. Could you add `yaml-bridge-service` and `yaml-bridge-action` as `[[example]]` targets in `libraries/extensions/ros2-bridge/Cargo.toml` and to the `EXAMPLES` array in `scripts/ros2dev.sh` (~line 115), so this is at least exercised at release QA? Failing that, a manual `ROS_DOMAIN_ID=23` run against `ros2 run examples_rclcpp_action_server fibonacci_action_server` pasted into the PR would do.

Kilted/Jazzy risk looks low now that the GID hunks are gone — the remaining changes are distro-neutral and the workspace is Humble-pinned by #2854. One thing to watch: `action_status_qos` forcing `transient_local` on the status subscription matches `rcl_action_qos_profile_status_default`, so it matches an rcl server, but it would stop matching a non-standard server publishing status VOLATILE. The `partial_sequence` to `sequence` rename in the examples is correct.

Also worth updating the PR title and body — they still describe enabling `pre-iron-gid`, which is no longer in the diff, and "Humble" is narrower than what this actually fixes.

SunSunSun689:
Addressed the latest review feedback in `279801045`:

- Removed the misleading best-effort warning from `service_or_action_qos`; explicit `qos: { reliable: false }` is still handled by the raw-config warning path.
- Collapsed `create_ros_context` to always use `Context::with_options(ContextOptions::new().domain_id(domain_id))`, and removed the `RosContextCreation` helper/test.
- Reverted the binary crate `rmw-zenoh` feature shim back to an inline dependency feature, so `cargo check -p dora-ros2-bridge-node --no-default-features` compiles again.
- Removed the test-only `service_topics_discovered` helper/test.
- Reused `current_ros_domain_id()` in DDS node creation.
- Added `yaml-bridge-service` and `yaml-bridge-action` as ros2-bridge example targets and wired them into `scripts/ros2dev.sh` release QA.
- Updated the PR title/body to match the current distro-neutral DDS service/action scope.

Local validation run:

```bash
cargo check -p dora-ros2-bridge-node --no-default-features
cargo test -p dora-ros2-bridge-node peer_failure_tests -- --nocapture
cargo check -p dora-ros2-bridge --example yaml-bridge-service --example yaml-bridge-action --features ros2-examples
cargo test -p yaml-bridge-action-goal -p yaml-bridge-action-server-handler -p yaml-bridge-service-requester -p yaml-bridge-service-handler
cargo fmt --all -- --check
cargo clippy -p dora-ros2-bridge-node -- -D warnings
cargo clippy -p dora-ros2-bridge --example yaml-bridge-service --example yaml-bridge-action --features ros2-examples -- -D warnings
cargo test -p dora-ros2-bridge-node --lib --tests
cargo test -p dora-ros2-bridge --lib --tests
```

All passed locally. The only warnings were the existing ROS build-script warnings from running without `AMENT_PREFIX_PATH` set.

phil-opp:

Re-reviewed after `2798010` ("address dds bridge review cleanup"). Both issues raised in the earlier review are now resolved:

1. **Misfiring best-effort warning — fixed.** The blanket warning is gone from `service_or_action_qos`; the retained `warn_explicit_service_action_best_effort` reads the raw config JSON and only fires when `qos.reliable` is *explicitly* `false` on a service/action bridge, so it no longer trips on the topic-default `reliable: false`.
2. **`default = ["rmw-zenoh"]` feature concern — not present.** The node still depends on the lib with `default-features = false, features = ["rmw-zenoh"]` and the lib's `default = ["generate-messages"]` is unchanged; `--no-default-features` is not newly broken by this PR.

No new issues found in this commit. The core change is a genuine, correctly-scoped post-1.0 bug fix — `create_ros_context` now uses `Context::with_options(...domain_id(...))` instead of the hardcoded domain-0 `Context::new()`, and service/action QoS is forced RELIABLE — restoring DDS service/action discovery, and it's covered by unit tests.

---

phil-opp:

Thanks for the follow-up in `2798010`: the misfiring best-effort warning is gone, `create_ros_context` is down to a single `with_options` call, the feature shim is reverted, and the title/body match what the PR does. The current head applies cleanly onto today's `main`, clippy is clean, `cargo check --examples` passes, and the unit tests pass. The core fix looks ready.

What still needs work is the release-QA wiring from last time. As it stands, the two new examples report green without exercising the bridge:

- **Wrong server package in `examples/ros2-bridge/yaml-bridge-action/run.rs:12`.** There is no `examples_rclcpp_action_server` package in ros2/examples (humble). The Fibonacci server is `examples_rclcpp_minimal_action_server` / `action_server_member_functions`, which is what the ros2dev image installs and what `rust/action-client/run.rs` uses. (That name came from the existing `dataflow.yml` header comment and our last review repeated it, sorry. Worth fixing the comment too.)
- **The examples can't fail.** `stop_after` is 12 s for the service and 20 s for the action, but the bridge waits up to 20 s (service) or up to 60 s (three action endpoints) before giving up, inside `block_on`, so it never sees Stop. The requester and goal nodes also exit `Ok` on Stop whether or not they got a reply, so with no server running at all the run is stopped and reported as passing. Please have the requester/goal node exit after N responses (and fail if Stop arrives first), with `stop_after` dropped or set above the retry window, like the existing `rust-ros2-dataflow-*` examples.
- **Is `dora-ros2-bridge-node` on PATH in the ros2dev container?** A `ros2:` node resolves to that bare binary name, and as far as I can see neither `cargo run -p dora-ros2-bridge --example …` nor the `qa` script builds it. If so, the run.rs files (or the `qa` script) need a `cargo build -p dora-ros2-bridge-node` with `target/debug` prepended to PATH.

Smaller, take or leave:

- `docs/ros2-bridge.md` still shows `partial_sequence` for Fibonacci feedback in a few places (~298, 592–599, 884, 910).
- Document that `qos.reliable: false` is ignored for service/action bridges.
- The four zenoh-mode call sites could use `current_ros_domain_id()` too.
- An empty `ROS_DOMAIN_ID=""` now logs "invalid"; rcl treats empty as unset.
- Please drop the whitespace-only change to `binaries/ros2-bridge-node/Cargo.toml` (`workspace = true ,`); it sits next to where #3560 edits that manifest and will cause a pointless conflict.

A pasted log of `yaml-bridge-action` passing against the real minimal action server under e.g. `ROS_DOMAIN_ID=23`, plus the same run failing with the server stopped, would settle it.

phil-opp:
Re-reviewed at `2798010` (unchanged). The core fix is correct — domain-aware `Context::with_options(ContextOptions::new().domain_id(...))` replacing the hardcoded domain-0 `Context::new()`, plus RELIABLE service/action QoS via `service_or_action_qos`. Two test-quality gaps worth flagging (non-blocking), since they aren't yet on the thread:

1. **The headline fix has no automated regression coverage.** The domain-id change lives in `create_ros_context` (`binaries/ros2-bridge-node/src/main.rs:~2021`), but only the pure helper `parse_ros_domain_id` is unit-tested — nothing asserts the parsed id actually reaches context creation. Reverting `create_ros_context` back to `Context::new()` would leave every unit test green. The only artifacts that exercise the fixed DDS discovery path are the two new `run.rs` examples, which require a live ROS2 install and run solely under release QA (`ros2dev.sh`) — not PR CI or nightly — so the fix has no CI safety net.

2. **`handler-node`'s `fibonacci_feedback_uses_sequence_field` is tautological.** It calls `make_list_struct("sequence", &[...])` and then asserts the produced column is named `"sequence"` and not `"partial_sequence"` (`handler-node/src/main.rs:104-108`). Since `make_list_struct(field_name, …)` names the column after its literal argument, this passes by construction regardless of whether `main()` (lines 60/66) uses the correct field — so it doesn't guard the rename it appears to protect.

The `service_qos_keeps_reliable_when_overriding_other_fields` guard, by contrast, is genuine and valuable. Nothing here blocks; a minimal assertion that the domain id flows into `create_ros_context` would close the main gap.

---
🤖 Fully automated Claude Code review — no human has vetted this; advisory only, not an approval, and I can't merge. Reported like an outside contributor.


---

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human vetted this; advisory only, not an approval, and I can't approve or merge.

Re-reviewed at head `2798010` (diff unchanged since the last pass). One **new** finding that also corrects an earlier automated verdict, plus a resolution of the open feedback-field question.

**New — `warn_explicit_service_action_best_effort` misfires on the *default* config, and its test locks in the misfire.** A [prior review](https://github.com/dora-rs/dora/pull/2878#issuecomment-5711489948) concluded this warning "only fires when `qos.reliable` is *explicitly* `false`… no longer trips on the topic-default `reliable: false`." Reading the raw JSON doesn't actually achieve that, because the JSON it reads is not the user's YAML — it's `serde_json::to_string(&config)` of the already-resolved `Ros2BridgeConfig` (`libraries/core/src/descriptor/mod.rs:293`). In `dora-message/src/descriptor.rs`, both `Ros2BridgeConfig::qos` (line 1669, `#[serde(default)]`, no `skip_serializing_if`) and `Ros2QosConfig::reliable` (line 1863, plain `bool`, `#[serde(default)]`, no `skip_serializing_if`) are serialized **unconditionally**, and `reliable` defaults to `false`. So *every* service/action bridge that doesn't set `qos.reliable: true` — the default and by far the most common configuration — serializes `"qos":{"reliable":false,…}`, and `main()` unconditionally emits `service/action QoS ignores explicit \`reliable: false\``, wrongly telling the user they explicitly asked for best-effort. Because `reliable` is a `bool` (not `Option<bool>`), explicit-false and default-false are byte-identical post-serialization, so "explicit" is not detectable as written. The test `service_qos_warns_when_explicit_best_effort_is_ignored` passes precisely because it feeds `{"reliable": false}` — the exact shape a default config produces — so it asserts the misfire rather than guarding against it. Fix options: make `reliable` an `Option<bool>` so default vs. explicit are distinguishable, or simply drop the warning (services are silently forced RELIABLE anyway, which is the correct ROS 2 behavior). This is low-severity (a spurious startup log, not a crash), but it's misleading on the common path.

**Resolved (the previously-open feedback-field question): the `partial_sequence` → `sequence` rename is correct, not a regression.** Both action dataflows declare `action_type: example_interfaces/Fibonacci` (unchanged by this PR), and the bridge derives the Arrow struct field names from the *declared* action type, not from whichever package the running server links. `example_interfaces/action/Fibonacci` feedback is `int32[] sequence` (only `action_tutorials_interfaces` uses `partial_sequence`), so the old `partial_sequence` lookup could never have matched — the pre-existing example would have panicked at `.expect("missing 'partial_sequence' field")` on the first feedback. The rename fixes a real pre-existing bug on both the goal-node and handler-node sides; the installed-package ambiguity from earlier doesn't affect the field name. So this one is settled — no run-log needed.

The already-open items from prior reviews still stand (no CI coverage for the `create_ros_context` domain-id fix; the false-green `run.rs` examples + wrong server package; the tautological `handler-node` test); I'm not re-listing them in detail.



---
_This is a fully automated review by Claude. It may contain mistakes — please verify before acting on it._

---

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human vetted this; advisory only, not an approval, and I can't approve or merge. Reviewed from the diff/code, not the description.

Diff unchanged since the last review, but an independent re-read surfaced findings the earlier passes missed. (The previously-reported items — the `warn_explicit_service_action_best_effort` false-positive on the default config, and the scope-bloat/split suggestion — are unchanged and still stand; not re-arguing them here.)

**1. The core fix has no regression coverage — reverting either load-bearing line leaves every test green.** All the new unit tests in `binaries/ros2-bridge-node/src/main.rs` exercise *pure helpers* in isolation: `parse_ros_domain_id`, the QoS builders (`default_service_qos_config` / `service_or_action_qos` / `action_status_qos_config`), and the `*_message` format strings. Nothing tests the two behaviors this PR actually fixes:
- that `create_ros_node` passes the parsed domain id into the `Context` (the real bug — `Context::new()` hardcodes `DomainParticipant::new(0)`), and
- that `run_service_mode` (`main.rs:1132`) / `run_action_mode` (`main.rs:1358`) call `service_or_action_qos(...)` rather than the old best-effort path.

So the headline claims (domain id applied, discovery restored via RELIABLE QoS) have zero unit coverage; the only end-to-end check is the ROS2-dependent `run.rs` examples, which run Linux-only in nightly. A test that asserts the domain id reaches the participant, and one that asserts the service/action modes select the reliable QoS, would make these genuine regression tests.

**2. Windows compile break in the new action example harness.** `examples/ros2-bridge/yaml-bridge-action/run.rs:5` imports `process_wrap::std::ProcessGroup` and `:33` calls `ProcessGroup::leader()` **unconditionally**. In `process-wrap`, `ProcessGroup` is `#[cfg(all(unix, feature = "process-group"))]`, so this fails to compile on Windows (unresolved import + unknown type). The sibling `yaml-bridge-service/run.rs:31-34` deliberately cfg-guards exactly this (`#[cfg(unix)] ProcessGroup` / `#[cfg(windows)] JobObject`). It's latent because the target carries `required-features = ["ros2-examples"]` (built only in the Linux-only ROS2 nightly), but it's a real portability defect and an unexplained divergence from the portable sibling — anyone running `--features ros2-examples --examples` on Windows hits it.

**3. (scope note) The RELIABLE-QoS fix is applied only to the native DDS path.** `run_zenoh_service_mode` (`main.rs:280`) and `run_zenoh_action_mode` (`main.rs:518`) still use `config.qos.to_neutral_qos()` (best-effort/volatile default), while only the DDS `run_service_mode`/`run_action_mode` got `service_or_action_qos`. Consistent with the "DDS" title, but if the same reliability requirement applies over `rmw_zenoh`, that transport's service/action discovery remains unfixed — worth an explicit scope statement either way.



---
_This is a fully automated review by Claude. It may contain mistakes — please verify before acting on it._

---

phil-opp:
🤖 **Automated review** (fully automated Claude Code review; no human vetted this — advisory only).

I re-read `2798010`; the head is unchanged since the last review. I found one new issue that the earlier passes didn't cover.

**The "discovered topics" startup diagnostic always reports both service topics as found, even when no server exists** (`binaries/ros2-bridge-node/src/main.rs`, `discovered_topic_names`, used in the service and action unavailable errors).

- `Context::discovered_topics()` resolves to `DomainParticipant::discovered_topics()` in rustdds 0.14.3, which calls `DiscoveryDB::all_user_topics()`. That function filters out only topics starting with `DCPS`.
- Whenever a *local* writer or reader is created, rustdds also registers its topic in that DB as `DiscoveredVia::SelfDefined` (`update_topic_data_p`, called from `pubsub.rs` for both writers and readers).
- The bridge's own client creates the writer on `rq/<svc>Request` and the reader on `rr/<svc>Reply`, so both names are always in the list.

With no server running, or with the server on a different `ROS_DOMAIN_ID`, the error after the retry window therefore always reports both the request and response topics as `found`. The action endpoints behave the same way. That wording suggests the server was seen but didn't match, for example because of a QoS mismatch, which points users the wrong way in exactly the cases this diagnostic exists for. The unit tests only format hand-built topic lists, so they can't catch this.

A possible fix: use `domain_participant().discovered_readers()` / `discovered_writers()`, which only return external endpoints, and report whether a remote reader exists on the `rq/` topic and a remote writer on the `rr/` topic. Alternatively, drop the topic part of the message.

The items from earlier reviews at this head are still open.

---

phil-opp:
**Automated review by Claude (Claude Code)**: this is a fully automated review, not a review by a human maintainer.

The head is still `2798010`, the same commit as the last automated review. I re-read the diff and found one issue that the earlier reviews didn't cover.

**An out-of-range `ROS_DOMAIN_ID` now panics the DDS bridge in debug builds, or silently breaks it in release builds** (`binaries/ros2-bridge-node/src/main.rs:2026`, `parse_ros_domain_id`, used by `create_ros_node`).

`parse_ros_domain_id` accepts any `u16`. It only falls back to 0 when the value doesn't parse. That value now goes straight into `ContextOptions::new().domain_id(..)`. rustdds (0.14.x) computes the RTPS ports with plain `u16` arithmetic and no range check: `PB + DG * domain_id + ...`, where `PB = 7400` and `DG = 250`. So any `domain_id >= 233` overflows:
- In a debug build (overflow checks on), the bridge panics with "attempt to add/multiply with overflow" while creating the participant.
- In a release build, the port wraps. For example, domain 233 gives SPDP multicast port 114 instead of about 65650. The bridge starts, never discovers anything, and then fails with the "service not available" diagnostic, which doesn't point at the real cause.

Before this PR, the DDS path always used `Context::new()` (domain 0), so this is a regression the PR introduces. ROS 2 only defines domain IDs 0–232, and other RMWs reject larger values at startup. I'd suggest treating values above 232 like unparseable ones: warn and fall back to 0, or preferably return an error. A unit test for `Some("233")` would lock this in. (I checked the arithmetic with a standalone copy of rustdds' port formula. I couldn't run it against a real ROS 2 setup here.)

The open items from the earlier reviews at this head still stand.

---

phil-opp:
🤖 **Automated review by Claude.** This review was fully automated (Claude Code) and no human checked it before posting, so treat it as advisory.

I re-reviewed head `2798010`. The earlier open items still apply. I found one new issue. The stricter action-client startup was raised before; this is a different consequence of it.

**The new action-client wait runs before the node registers with Dora, so it holds up the startup of the whole dataflow. It can also deadlock against an action server running in the same dataflow.**

- **Where it happens:**
  - In `run_action_mode`, `wait_for_action_client_services(...)` runs at `binaries/ros2-bridge-node/src/main.rs:1382`.
  - `DoraNode::init_from_env()` is only called afterwards, inside `run_action_client` (`:1424`).
- **Why that matters:**
  - The daemon parks every node's subscribe call until all nodes of the dataflow have subscribed (`binaries/daemon/src/pending.rs`: "Subscribe requests block the node until all other nodes are ready too").
  - So while the bridge waits (up to 3 endpoints × 10 retries × 2 s), no other node gets past `init_from_env()`, and no timers or inputs flow.
- **Deadlock scenario:**
  1. One dataflow contains a YAML `role: client` action bridge and a user's native action-server node. `docs/ros2-bridge.md` recommends pairing a Dora-hosted server with a Dora client.
  2. The server node calls `init_from_env()` / `Node()` before `create_action_server`, so it parks in the startup barrier.
  3. The bridge waits for endpoints that the server only creates after the barrier releases.
  4. The bridge gives up before ever subscribing, and the dataflow fails to start.
- **On `main`:** the action client did not wait at all, so this setup works there.
- **Scope:** the in-repo example servers create their ROS server before init, which is why nothing in the repo hits this. User code that inits first does hit it.
- **Possible fix:** call `init_from_env()` first, then do the wait. Better still, do the wait while also watching Dora events, so that a `Stop` during the wait is also honoured.

I found this by reading the code (the call order, plus the daemon barrier in `pending.rs`). I did not run it against ROS 2.

---

PR #2879, opened 2026-07-29
fix(coordinator): apply single-operator output prefixing to AddNode inputs

### LINKED ISSUE #2877: Dynamic `AddNode` skips single-operator output prefixing, silently mis-wiring inputs that reference a legacy `operator:` producer
## Summary

When a node is added at runtime via the dynamic-topology `AddNode` control request, its input mappings are resolved **in isolation** (against a temporary single-node descriptor). As a result, the single-operator output-prefixing that `resolve_aliases_and_set_defaults` normally applies is **not** applied for inputs that reference an *existing* producer declared with the legacy single-`operator:` syntax. The dynamically-added consumer then subscribes to the unqualified output name and **silently receives no data** — there is no error, the node just never gets messages on that input.

_This issue was filed by an automated, scheduled Claude code-review check. A maintainer should confirm the intended support level of the `AddNode` + single-`operator:` producer combination before acting._

## Mechanism / evidence chain

1. **Runtime nodes publish operator outputs under the `<operator_id>/<output>` name.**
   `binaries/runtime/src/lib.rs:421`
   ```rust
   fn operator_output_id(operator_id: &OperatorId, output_id: &DataId) -> DataId {
       DataId::from(format!("{operator_id}/{output_id}"))
   }
   ```

2. **Normal descriptor resolution rewrites consumer inputs to the qualified name.**
   In `resolve_aliases_and_set_defaults`, `single_operator_nodes` is built from **all** nodes in the descriptor, and any input whose `source` is a single-operator node has its `output` rewritten (`result` → `op/result`):
   `libraries/core/src/descriptor/mod.rs:134-142` and `:177-179`
   ```rust
   let single_operator_nodes: HashMap<_, _> = self
       .nodes
       .iter()
       .filter_map(|n| n.operator.as_ref().map(|op| (&n.id, op.id.as_ref().unwrap_or(&default_op_id))))
       .collect();
   // ...
   if let Some(op_name) = single_operator_nodes.get(&mapping.source).copied() {
       mapping.output = prefix_output_with_operator_id(op_name, &mapping.output)?;
   }
   ```

3. **The coordinator's `AddNode` path resolves the new node in isolation.**
   `binaries/coordinator/src/lib.rs:1586-1598` builds a one-node `tmp_desc` and calls `resolve_aliases_and_set_defaults()` on it. Because `tmp_desc.nodes` contains only the *new* node, `single_operator_nodes` never contains the existing producer, so the rewrite in step 2 is skipped — even though the coordinator has the full topology available (`RunningDataflow.descriptor`, `binaries/coordinator/src/state.rs:246`).
   ```rust
   let tmp_desc = dora_message::descriptor::Descriptor {
       nodes: vec[node],
       // ...
   };
   match tmp_desc.resolve_aliases_and_set_defaults() { /* ... */ }
   ```

4. **The daemon inserts the resolved mapping verbatim.**
   `binaries/daemon/src/lib.rs:2212-2216` keys the routing table by `OutputId(mapping.source, mapping.output)` with no re-prefixing against the full node set:
   ```rust
   dataflow
       .mappings
       .entry(OutputId(mapping.source.clone(), mapping.output.clone()))
       .or_default()
       .insert((node_id.clone(), input_id.clone()));
   ```

**Net effect:** the dynamically-added consumer registers a subscription to `OutputId("producer", "result")`, while the producer runtime emits `OutputId("producer", "op/result")`. The keys never match, so the input is silently starved.

## Concrete scenario

1. Start a dataflow containing a producer defined with the legacy single-operator syntax, e.g.:
   ```yaml
   - id: producer
     operator:
       # no explicit id -> defaults to `op`
       outputs: [result]
       # ...
   ```
   At runtime this publishes `producer/op/result`.
2. At runtime, issue `AddNode` for a consumer whose input references the unqualified output:
   ```yaml
   - id: consumer
     inputs:
       data: producer/result
   ```
3. **Expected:** the consumer's `data` input is wired to `producer/op/result` (as it would be for a statically-declared consumer) and receives messages.
   **Actual:** the mapping stays `producer/result`, never matches the producer's `producer/op/result` output, and the consumer receives nothing — with no error surfaced.

The ROS2-bridge input-adjustment branch (`libraries/core/src/descriptor/mod.rs:147-155`) shares the same root cause and is affected identically.

## Suggested direction

Resolve the added node against the existing dataflow's node set so single-operator producers are visible during prefixing — e.g. build the `tmp_desc` (or at least the `single_operator_nodes` map) from `RunningDataflow.descriptor.nodes` plus the new node, rather than from the new node alone. The coordinator already holds the full `descriptor` at this call site.

## Severity

Medium — silent data loss (no crash, no error message) on a supported feature combination (dynamic node addition + legacy single-`operator:` producer). Statically-declared dataflows are unaffected.

phil-opp:
Fixes #2877.

A node joining a *running* dataflow via `dora node add` / `dora node replace` was resolved against a temporary single-node descriptor. Whole-descriptor resolution rewrites an input that references a single-`operator:` producer from the bare output name to the operator-qualified one (`result` → `op/result`), but with only the new node present that producer is invisible, so the rewrite was skipped. The node then subscribed to the unqualified output name that nobody publishes and **silently received no data**.

## Changes

- **`libraries/core`**: add `resolve_aliases_and_set_defaults_in_topology(desc, topology_nodes)`, which resolves a node while sourcing the single-operator-producer lookup from the surrounding topology as well as `desc`. A free function rather than a `DescriptorExt` method so the published trait keeps its shape, taking `&[Node]` to make explicit that the surrounding nodes feed *only* that lookup — nothing else on the running descriptor is consulted. `resolve_aliases_and_set_defaults` becomes a thin `&[]` call into it, so the no-topology behavior is unchanged by construction. `desc`'s own nodes win over a same-id topology entry.
- **`binaries/coordinator`**: `resolve_single_node` now takes the running dataflow's descriptor instead of just its `env:`, so the topology lookup and the existing env carry-through (#2919) travel together. The `AddNode` and `ReplaceNode` arms already share that helper, so `dora node replace` — which had the same defect — is fixed with it.

## Behavior note

`DataId` permits `/`, so anyone who worked around this by writing the qualified name themselves (`inputs: {x: producer/op/result}`) now gets it rewritten to `op/op/result`. Those descriptors have to drop the manual prefix — which is what a statically declared consumer already requires.

## Tests

- `dora-core`: `add_node_prefixes_single_operator_producer_input_via_topology` (prefixing applies with the topology, not without) and `topology_lookup_prefers_the_node_being_added_over_a_same_id_topology_entry` (lookup precedence).
- `dora-coordinator`: `dynamic_node_prefixes_input_referencing_a_single_operator_producer` and `dynamic_node_inherits_the_running_dataflow_env`, pinning the call site itself — previously the resolution was inlined in the control-request match with no coverage of its own. The first fails with `left: "result", right: "op/result"` if the topology argument is dropped.

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -25,6 +25,7 @@ use dora_message::{
         DaemonCoordinatorEvent, RegisterResult, StateCatchUpOperation, Timestamped,
     },
     daemon_to_coordinator::{DaemonCoordinatorReply, DataflowDaemonResult},
+    descriptor::{Descriptor, Node, ResolvedNode},
     id::NodeId,
 };
 pub use events::{DaemonRequest, DataflowEvent, Event};
@@ -1583,13 +1584,12 @@ async fn start_inner(
                                         // node).
                                         let original_node = node.clone();
 
-                                        // See `resolve_single_node` for the
-                                        // env carry-through rationale
-                                        // (dora-rs/dora#2919).
-                                        match resolve_single_node(
-                                            node,
-                                            dataflow.descriptor.env.clone(),
-                                        ) {
+                                        // See `resolve_single_node` for what
+                                        // the running descriptor contributes
+                                        // (env carry-through #2919,
+                                        // single-operator output prefixing
+                                        // #2877).
+                                        match resolve_single_node(node, &dataflow.descriptor) {
                                             Ok((node_id, resolved_node)) => {
                                                 // Pick the first daemon (single-daemon case)
                                                 // TODO: use machine label or load balancing for multi-daemon
@@ -1789,7 +1789,12 @@ async fn start_inner(
                                 // Route to the daemon that OWNS the id — a
                                 // replace targets an existing node, unlike
                                 // AddNode's first-daemon placement.
-                                let (daemon_id, uv, dataflow_env) = {
+                                let original_node = node.clone();
+                                // Resolve inside the borrow so the running
+                                // descriptor can be passed by reference; the
+                                // borrow ends before `daemon_connections` is
+                                // taken mutably below.
+                                let (daemon_id, uv, node_id, resolved_node) = {
                                     let dataflow =
                                         running_dataflows.get(&dataflow_id).ok_or_else(|| {
                                             eyre!("no running dataflow with ID {dataflow_id}")
@@ -1805,11 +1810,10 @@ async fn start_inner(
                                                 node.id
                                             )
                                         })?;
-                                    (daemon_id, dataflow.uv, dataflow.descriptor.env.clone())
+                                    let (node_id, resolved_node) =
+                                        resolve_single_node(node, &dataflow.descriptor)?;
+                                    (daemon_id, dataflow.uv, node_id, resolved_node)
                                 };
-                                let original_node = node.clone();
-                                let (node_id, resolved_node) =
-                                    resolve_single_node(node, dataflow_env)?;
                                 let msg = serde_json::to_vec(&Timestamped {
                                     inner: DaemonCoordinatorEvent::ReplaceNode {
                                         dataflow_id,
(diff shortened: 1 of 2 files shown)


phil-opp:

I reviewed this closely and it looks correct. The extraction into `resolve_aliases_and_set_defaults_impl` preserves the existing single-argument behavior exactly (the public entry point calls it with `&[]`), and the surrounding topology nodes are used only to populate the single-operator prefix lookup — they're never resolved or emitted (the loop still iterates `desc.nodes` only), and `merge_env` correctly reads from `desc`. The coordinator passes the running dataflow's descriptor so the existing single `operator:` producer becomes visible and the `op/` prefix is applied to AddNode inputs. The added test pins the real behavior (`producer/result` → `op/result` with the topology present, unchanged without it), so a regression would fail it. No issues found.

---

PR #2880, opened 2026-07-29
fix(cli): follow log rotation by file identity in `dora logs --local --follow`

### LINKED ISSUE #2876: `dora logs --local --follow`: rotation fix in #2785 (a68ee18) is partial — fresh-file-past-offset race still garbles, and rotated `.1.jsonl` tail is still dropped

## Summary

Commit **a68ee18** (`fix(cli): reset follow offset on log rotation in dora logs --local --follow`, #2800) closed **#2785** as *completed*. It genuinely fixes the **common** rotation case, but it implements only the weakest of the three remedies that #2785 itself proposed, and **two of the failure scenarios documented in #2785 remain reproducible**. The new doc comment on `follow_read_offset` now claims a guarantee the code does not actually provide.

## What a68ee18 actually does

`binaries/cli/src/command/logs.rs:345`

```rust
fn follow_read_offset(pos: u64, current_size: u64) -> Option<u64> {
    let start = if current_size < pos { 0 } else { pos };
    if current_size <= start { None } else { Some(start) }
}
```

This resets to offset `0` **only while the freshly-rotated file is still smaller than the stale `pos`**. That covers the typical case (a 200 ms poll usually catches the new file well below `max_log_size ≈ pos`), which is a real improvement over the old `current_size <= pos { continue }` that dropped everything.

## Residual gap 1 — the exact scenario the new doc comment claims to prevent

The doc comment at `binaries/cli/src/command/logs.rs:339-344` states:

> Seeking to the stale `pos` in the fresh file would silently drop its leading bytes and garble the remainder **once it grows past `pos`**, so we resume from offset 0 instead.

But the detection is `current_size < pos`. If a rotation happens and the fresh `log_<node>.jsonl` **grows past `pos` within a single 200 ms poll interval**, then on the next poll `current_size >= pos`, so `follow_read_offset` returns `Some(pos)` and `read_appended_log_lines` seeks to the stale offset **in the new file** — dropping its leading bytes and starting mid-line. This is precisely the "grows past `pos`" case the comment says is handled; it is not.

- Reachability: requires the node to write ~`pos` (≈ `max_log_size`) bytes within one 200 ms poll. Narrow with a large `max_log_size` (e.g. the documented `"100MB"` example), but plausible for a chatty node configured with a small `max_log_size` (e.g. `"1MB"` at 5 MB/s). **Low severity, but real.**

## Residual gap 2 — the just-rotated `.1.jsonl` tail is never followed

`files` is captured **once** before the poll loop (`logs.rs:275-278`) and the loop iterates that fixed list (`logs.rs:315`), never re-globbing. When the daemon rotates (`binaries/daemon/src/log.rs:43-72`), it renames the current `log_<node>.jsonl` → `log_<node>.1.jsonl` and creates a fresh current file. Any bytes written to the old file **between the follow loop's last read (`pos`) and the rename** now live only in the newly-created `.1.jsonl`, which was not in `files` at startup and is never picked up. Those lines (up to ~one poll interval of logs) are lost on every rotation.

This is the same gap #2785 explicitly called out: *"The rotated content (now in `.1.jsonl`) is never followed either — that path was never in the `files` list."* It was not addressed by a68ee18.

## Concrete failing scenario (gap 1)

1. Node configured with a small `max_log_size` and logging at high rate; `dora logs --local --follow` running, `pos = 1_000_000`.
2. Daemon rotates: renames current → `.1.jsonl`, creates fresh `log_<node>.jsonl` at 0, keeps writing.
3. Within the next 200 ms the fresh file exceeds 1_000_000 bytes.
4. Next poll: `current_size (>1_000_000) >= pos (1_000_000)` → `follow_read_offset` returns `Some(1_000_000)` → seek to offset 1_000_000 in the **new** file → leading ~1 MB dropped, output resumes mid-line (garbled).

## Suggested fix direction

Track the file's **identity**, not just its size:
- On Unix, record `dev`+`ino` (`std::os::unix::fs::MetadataExt`) alongside the offset; when it changes, treat the file as rotated and reset `pos = 0` regardless of size. On Windows, fall back to creation-time or a size-decrease heuristic.
- Re-glob the log directory each poll (or on a detected rotation) so the freshly-created `.1.jsonl` is picked up and its unread `pos..EOF` tail is drained before switching to the new current file.

Either change closes both residual gaps and lets the `follow_read_offset` doc comment's guarantee hold. At minimum, the doc comment should be corrected to reflect that only the size-shrink case is detected.

## Scope / notes

- Impact is limited to `dora logs --local --follow` with `max_log_size` configured (rotation only occurs then). The coordinator streaming path is unaffected.
- The inner `read_appended_log_lines` helper is correct; this is about the rotation/offset lifecycle around it.
- Introduced-by / related: **a68ee18** (#2800), which closed **#2785**.

phil-opp:
## Summary

Fixes #2876.

The rotation fix in a68ee18 (#2800) closed the common case of `dora logs --local --follow` across a log rotation, but two documented failure scenarios remained:

1. **Fresh-file-past-offset garble.** `follow_read_offset` detected rotation only via `current_size < pos`. If the freshly-rotated `log_<node>.jsonl` grew *past* the stale offset within a single 200 ms poll, the next poll seeked to the stale offset in the new file, dropping its leading bytes and starting mid-line.
2. **Dropped `.1.jsonl` tail.** The followed file list was globbed once before the poll loop and never refreshed, so the just-rotated `log_<node>.1.jsonl` (and any file that appeared after the follow started) was never read — its unread tail was lost.

## Approach

The daemon rotates by *renaming* (`log_<node>.jsonl` → `.1.jsonl`, shifting older ones up) and creating a fresh current file (`binaries/daemon/src/log.rs`). Tracking files by path alone can't tell "the current file rotated away" from "the current file grew". The fix tracks each followed file by a **head-byte fingerprint** (first ≤256 bytes, made distinct per file by the JSONL timestamp on line 1) in addition to its offset, and **re-globs the directory every poll**:

- **Same file at a path** (recorded head is a prefix of the current head — a followed file only grows): resume from the tracked offset.
- **Content that moved to a new path** (rotation rename — the fingerprint now appears at a different path): carry the offset there and drain the unread tail exactly once.
- **A genuinely new file** (fingerprint matches nothing tracked): read from 0. A fresh current file that grew past the stale offset falls here — its differing fingerprint means it is read from the start, not seeked into mid-line (fixes gap 1).

This is portable (no inode / platform-specific file-identity APIs) and the decision logic (`plan_follow_reads`) is a pure function, unit-tested independently of the poll loop. A transient failure to fingerprint (empty head) keeps the tracked offset rather than re-reading from 0, so a flaky `stat`/`open` cannot duplicate a whole file.

## Testing

New unit tests in `binaries/cli/src/command/logs.rs` (all passing; 36 in the logs module):
- `same_log_file_matches_growing_prefix_only`
- `plan_resumes_from_offset_on_plain_append`
- `plan_reads_brand_new_file_from_start`
- `plan_rotation_drains_tail_and_reads_fresh_file_from_zero`
- `plan_fresh_file_grown_past_offset_is_not_garbled` (gap 1)
- `plan_unreadable_head_keeps_offset_instead_of_rereading`
- `plan_already_read_rotated_file_is_not_reread_on_shift`

Plus `cargo clippy -p dora-cli -- -D warnings` and `cargo fmt --all -- --check` — both clean.

https://claude.ai/code/session_01N44E4D59JxczmnyyRe6vBS

---
_Generated by [Claude Code](https://claude.ai/code/session_01N44E4D59JxczmnyyRe6vBS)_

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -286,10 +286,17 @@ fn follow_local_logs(args: &LogsArgs) -> Result<()> {
         );
     }
 
-    // Print existing content with filters
+    // Print existing content with filters, tracking per-file follow state
+    // (offset already consumed + a head fingerprint to recognise the same file
+    // across a rotation rename) as we go. The state comes from the *same* read
+    // as the printed content, so a rotation in between cannot leave the rotated
+    // file untracked and make the first poll re-print it from the start.
     let mut all_messages: Vec<LogMessage> = Vec::new();
+    let mut follow_state: HashMap<PathBuf, FollowedFile> = HashMap::new();
     for path in &files {
-        all_messages.extend(read_log_file(path)?);
+        let (messages, state) = read_log_file_tracked(path)?;
+        all_messages.extend(messages);
+        follow_state.insert(path.clone(), state);
     }
     all_messages.sort_by_key(|a| a.timestamp);
     let filtered = apply_time_filters(all_messages, args.since, args.until, now);
@@ -298,21 +305,6 @@ fn follow_local_logs(args: &LogsArgs) -> Result<()> {
 
     for msg in display {
         print_log_message(msg, &config);
-    }
-
-    // Track per-file follow state (offset already consumed + a head fingerprint
-    // to recognise the same file across a rotation rename). Start after the
-    // existing content printed above.
-    let mut follow_state: HashMap<PathBuf, FollowedFile> = HashMap::new();
-    for path in &files {
-        let size = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
-        follow_state.insert(
-            path.clone(),
-            FollowedFile {
-                offset: size,
-                head: read_log_head(path),
-            },
-        );
     }
 
     // Follow loop: poll for new content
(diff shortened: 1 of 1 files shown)


phil-opp:

I traced `plan_follow_reads` through the rename-based rotation scenarios — rotation with an unread tail, a fresh file that grew past the stale offset within one poll, a fully-read rotated file shifting up, in-place truncation, and a transient unreadable head — and the fingerprint-following logic (`same_log_file` head-prefix match plus the `.min(snap.size)` truncation guard) drains each tail exactly once without re-reading or seeking mid-line. The unit tests assert the concrete resume offsets for each case, so they'd catch a regression to the old size-only behavior. No issues found.

One very minor, non-blocking note: the poll loop reads with `read_appended_log_lines(...)?`, so a file vanishing between the per-poll snapshot and the read would end the follow session via the propagated error. It looks effectively unreachable (rotation renames rather than deletes the file being read, and the pruned oldest file has already been fully consumed so no read is attempted), so just flagging it rather than asking for a change.

---

phil-opp:
🤖 **Automated review by Claude** (fully automated — not vetted by a human).

I found one issue: the "a transient failure to fingerprint keeps the tracked offset" invariant doesn't hold when the `stat` and the head-read fail *together*.

In `plan_follow_reads`, the empty-head arm:

```rust
// Couldn't re-fingerprint (transient read error): assume the same file
// and keep the recorded head ... never re-read from 0 and duplicate.
Some(state) if snap.head.is_empty() => {
    (state.offset.min(snap.size), state.head.clone())
}
```

`snap.size` is `std::fs::metadata(path).map(|m| m.len()).unwrap_or(0)` and `snap.head` is empty exactly when `read_log_head` failed. Both target the same path in the same poll, so the failure is correlated: if the file is momentarily unavailable (e.g. a rotation rename landing between the per-poll glob and these `stat`/`open` calls, or any transient stat error), `metadata` returns `Err` → `snap.size == 0` **and** `snap.head` is empty. The arm then computes `resume = state.offset.min(0) = 0` and records `offset = 0` for the next poll.

That poll reads nothing (`size 0`), but the reset offset persists. On the next poll the file is readable again and unchanged, so `same_log_file(OLDHEAD, OLDHEAD)` is true and the second arm computes `resume = 0.min(size) = 0` → `read_appended_log_lines(path, 0)` re-reads the whole file and re-prints every already-emitted line. If the transient failure coincided with a rotation, the `0` offset also propagates to the renamed `.1.jsonl` through the fingerprint-carry in the `_` branch, duplicating the rotated tail as well.

This contradicts the commit's stated guarantee that *"a flaky stat/open cannot duplicate a file"* — the `.min(snap.size)` clamp defeats it precisely when the stat also fails.

Suggestion: in this arm keep `state.offset` unchanged rather than clamping to a size that was itself read as `0`/untrusted. The `.min(snap.size)` truncation guard is only sound in the arm where the head was successfully re-read and the file is known to be present.

The added `plan_unreadable_head_keeps_offset_instead_of_rereading` test doesn't catch this because it uses a non-zero size (`80`) with the empty head. A case like `snap("log_n.jsonl", 0, b"")` with a `prev` offset of `80` should assert `resume == 80`, and currently returns `0`.

---

PR #2885, opened 2026-07-29
chore(node): log send_output hop entry to pin the #2742 Windows wedge

phil-opp:
Follow-on to #2857. Refs #2742.

## What the #2857 diagnostics showed

They fired in the 2026-07-29 nightly ([run 30436246412](https://github.com/dora-rs/dora/actions/runs/30436246412), job `CLI Tests (windows-latest)`) and pin the wedge:

| time (UTC) | event |
|---|---|
| 10:03:12.92 | `webcam: camera_available=False` — the headless gate works, `cv2.read()` is never called |
| 10:03:26.45 | `webcam-diag: t=13.54s send_output START` → `send_output DONE` (**0.3 ms**) |
| ~10:03:26.45 | runtime main loop enters `OperatorEvent::Output` — and never leaves |
| 10:03:30.99 → 47.00 | `runtime 'webcam' main loop stalled for 5s … 21s while handling an operator output` |
| 10:03:32.03 | stop-after fires; `object_detection` and `plot` log `received Stop` → `forwarded Stop` within 1 ms and exit `Success`. **webcam logs no `received Stop`, ever** |
| 10:03:47.05 | `webcam was killed due to not stopping within the 15s grace period` → `ExitCode(1)` → job red |

So the Python operator is not stuck — it even returned `DoraStatus.STOP` on its own at t=13.54s (`webcam.py:115-120`, `CI_HEADLESS_RUN_SECS=10`), i.e. 5.6 s before `--stop-after` fired. The wedge is `node.send_output(...)` in `dora-runtime`'s main loop (`binaries/runtime/src/lib.rs:293-299`) for the 921,600-byte `op/image` payload. Corroborating: the `event queue was too full` warnings — emitted by that same loop when it forwards inputs — stop dead at 10:03:26.45.

## Why this patch

The runtime watchdog resolves only to `"an operator output"`, one level too coarse. Inside `DoraNode::send_output_sample` there are two candidate blocking hops for a payload this size and nothing in the log distinguishes them:

1. `provider.alloc(...).wait()` / `publisher.put(sbuf).wait()` in `zenoh_publish`
2. `control_channel.send_message(...)` — a blocking `write_all` of ~900 KB to the daemon's local TCP socket (`daemon_connection/tcp.rs:68-82`), with no timeout anywhere on the path

#2806 makes both reachable: an output whose startup handshake has not completed now takes hop (2) where it previously took hop (1).

## Changes

`apis/rust/node/src/node/mod.rs`:

- Log on **entry** to each hop, not after it. A post-hoc "took Ns" log can never fire here — the process is killed while still inside the call. The last line printed before the stall warning names the blocked call outright.
- Covered hops: chosen delivery path (zenoh direct vs daemon), SHM alloc, zenoh put (both the moved-SHM and copied-SHM arms), `control_channel.send_message`, `report_output_sent`.
- Capped at the first `LARGE_SEND_DIAG_LIMIT = 3` sends at or above the zero-copy threshold (the wedge is on the first), via a `u32` counter on `DoraNode`. A healthy run pays one comparison per send and prints a handful of lines; sub-threshold sends are never traced.
- `warn!` level so the lines survive the runtime's default stdout filter and reach the nightly log, matching #2857.

Temporary observability, to be removed together with the runtime's stall watchdog once #2742 is understood.

## Validation

Class A (diagnostics only, no behavior change on the data path).

- `cargo clippy -p dora-node-api --all-targets -- -D warnings` — clean
- `cargo check -p dora-node-api` — clean
- Linux smoke: `DORA_ZERO_COPY_THRESHOLD=1 cargo run --example rust-dataflow`, which forces every send onto the traced path, to confirm the hop lines print and the dataflow still completes normally

## Not addressed here

Even once the stalling hop is fixed, the runtime's single event loop remains a single point of failure: *any* slow output send leaves a runtime node unable to observe `Stop` and unable to act on its operator's own `Finished`. Handling those concurrently with an in-flight send would have turned this nightly green on its own, but it changes runtime stop semantics and deserves a deliberate maintainer call rather than a drive-by.

https://claude.ai/code/session_013ZE91SPs2b1FKoUgdVEgF8

---
_Generated by [Claude Code](https://claude.ai/code/session_013ZE91SPs2b1FKoUgdVEgF8)_

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -136,6 +136,14 @@ mod runtime_type_check_tests {
 /// TCP.
 pub const ZERO_COPY_THRESHOLD: usize = 4096;
 
+/// How many large outbound sends are traced hop-by-hop
+/// (dora-rs/dora#2742 diagnostic; see [`DoraNode::send_output_sample`]).
+///
+/// The Windows nightly wedge happens on a node's *first* large output, so a
+/// handful of traced sends is enough to name the blocked call while keeping a
+/// healthy run's logs clean.
+const LARGE_SEND_DIAG_LIMIT: u32 = 3;
+
 /// How often a starting node re-publishes its startup route-probe markers.
 ///
 /// See [`StartupHandshake`]. Markers stop per output as soon as that output's
@@ -710,6 +718,13 @@ pub struct DoraNode {
     zenoh_schema_state: HashMap<DataId, SchemaOnceState>,
     /// Threshold for using zenoh SHM vs inline bytes (default 4096).
     zenoh_zero_copy_threshold: usize,
+
+    /// Diagnostic (dora-rs/dora#2742): how many large sends have already been
+    /// traced hop-by-hop. The Windows nightly wedges the *runtime's* main loop
+    /// inside `send_output` on the very first large output, so tracing only the
+    /// first few large sends pins the stuck hop without spamming a healthy run.
+    /// Remove together with the runtime's stall watchdog once #2742 is closed.
+    large_send_diag_count: u32,
 
     dataflow_descriptor: serde_yaml::Result<Descriptor>,
     warned_unknown_output: BTreeSet<DataId>,
@@ -1326,6 +1341,7 @@ impl DoraNode {
             zenoh_schema_publishers: HashMap::new(),
             zenoh_schema_state: HashMap::new(),
             zenoh_zero_copy_threshold,
+            large_send_diag_count: 0,
             dataflow_descriptor: serde_yaml::from_value(dataflow_descriptor),
             warned_unknown_output: BTreeSet::new(),
             interactive: false,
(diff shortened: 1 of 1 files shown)


phil-opp:
The Linux smoke listed as pending in the description has now completed. `DORA_ZERO_COPY_THRESHOLD=1 cargo run --example rust-dataflow` (threshold forced to 1 B so every send is treated as "large" and takes the traced path):

```
WARN rust-node: dora_node_api::node  output `random`: 456 B -> zenoh direct path (dora-rs/dora#2742 diagnostic)
WARN rust-node: dora_node_api::node  output `random`: entering zenoh put of an SHM buffer (dora-rs/dora#2742 diagnostic)
WARN rust-node: dora_node_api::node  output `random`: entering report_output_sent (dora-rs/dora#2742 diagnostic)
   … ×3, then silent for the rest of the run
```

- Exactly **9 lines** — the `LARGE_SEND_DIAG_LIMIT = 3` cap holds; the remaining sends are untraced.
- Each traced send emits the full hop chain, so the last line before a stall names the blocked call.
- Dataflow completed normally: all three nodes `finished successfully`, exit 0.

Also: `cargo test -p dora-node-api --lib` — 154 passed, 0 failed; `cargo fmt --all -- --check` and `cargo clippy -p dora-node-api --all-targets -- -D warnings` clean.

Note that the run above only exercises the `zenoh direct path` → `SHM put` → `report_output_sent` arm. The other traced arms (SHM alloc, copied-SHM put, `control_channel.send_message`) are entry-log-only additions on the same pattern and are the ones the Windows nightly is expected to land in — they can't be reached on a healthy local run precisely because nothing stalls there.


---

PR #2886, opened 2026-07-29
fix(coordinator): don't disconnect a live daemon on a heartbeat-send backpressure timeout

phil-opp:
## Issue

The per-daemon heartbeat is written to the daemon's **bounded command channel** — the same channel every coordinator→daemon command (spawn, stop, add/remove node, …) uses. `send_heartbeat_with_timeout` wraps the send in a 500 ms `timeout`, and the watchdog disconnected the daemon on **any** error, including that timeout:

```rust
let result = tokio::time::timeout(Duration::from_millis(500), send_heartbeat_message(..))
    .await.wrap_err("timeout").and_then(|r| r) ...;
// caller:
if let Err(err) = result { disconnected.insert(machine_id); }
```

A disconnected daemon is removed, unregistered from the store, and **all of its running dataflows are torn down**. A `timeout` here only means the bounded command channel was momentarily full — backpressure from a large spawn or a burst of control commands — while the daemon is provably alive (its *inbound* heartbeat was checked against the 30 s threshold immediately above). So a single slow WS-write burst could falsely disconnect a healthy daemon and fail all of its dataflows.

## Fix

Classify the send outcome (`Delivered` / `SendFailed` / `TimedOut`) and only disconnect *immediately* on `SendFailed`. A transient timeout is tolerated; a persistent streak escalates to a disconnect after `MAX_CONSECUTIVE_HEARTBEAT_SEND_TIMEOUTS` (10 ≈ 30 s at the 3 s interval). A genuinely dead daemon is still caught by (a) the next send returning `Err` once the WS writer's receiver drops, and (b) the existing 30 s inbound-heartbeat threshold.

## Scope of the escalation backstop (reviewer-raised, important)

The escalation bounds only the **no-traffic wedge**: a daemon whose command channel stays full while *no* command is routed to it is disconnected after ~30 s of heartbeat ticks. It does **not** bound the case where a command is routed to a wedged daemon first — `DaemonConnection::send_and_receive` awaits the same channel with **no** timeout, so that send blocks the single-threaded coordinator event loop before any further heartbeat tick can advance the counter. Pre-fix, one 500 ms timeout disconnected the daemon and stopped routing there, so this change narrows that immediate backstop. Fully closing it means bounding the `send_and_receive` enqueue, which changes command-dispatch failure semantics and is left as a **separate follow-up** (see thread). The const's doc and the code comments state this limitation explicitly rather than claiming the backstop is unconditional.

Net effect: the common, impactful failure (a transient spawn burst tearing down all of a live daemon's dataflows) is fixed; the rare wedged-daemon-plus-routed-command head-of-line window is documented and left to the follow-up.

## Validation

- `heartbeat_disconnect_decision` is a pure helper with unit tests (tolerate transient → escalate persistent → reset on delivery → disconnect on `SendFailed`).
- `send_heartbeat_with_timeout` is covered by two **wiring** tests so re-flattening the outcome back into "disconnect on any error" can't pass silently: a dropped-receiver channel → disconnect; a full capacity-1 channel under `#[tokio::test(start_paused = true)]` → a tolerated timeout that only bumps the streak.
- `cargo test -p dora-coordinator` (148 passed), `clippy --all-targets -D warnings`, and `fmt --check` all green on rustc 1.97.1, rebased on current `main`.

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -293,6 +293,40 @@ async fn start_with_events(
     Ok((port, future))
 }
 
+/// Outcome of the coordinator's periodic heartbeat send to one daemon.
+///
+/// The heartbeat is written to the daemon's bounded command channel — the
+/// *same* channel that carries every coordinator→daemon command (spawn, stop,
+/// add/remove node, …), not a dedicated heartbeat channel. That makes the two
+/// failure shapes mean very different things, and the disconnect decision must
+/// tell them apart (see [`HeartbeatSendOutcome::should_disconnect`]).
+enum HeartbeatSendOutcome {
+    /// The heartbeat was enqueued for delivery.
+    Delivered,
+    /// `send()` returned an error, i.e. the daemon's WS writer task has
+    /// dropped its receiver. The connection is genuinely gone.
+    ChannelClosed,
+    /// The send did not complete within the timeout: the bounded command
+    /// channel is momentarily full (backpressure from a large spawn or a
+    /// burst of control commands), *not* a dead daemon.
+    TimedOut,
+}
+
+impl HeartbeatSendOutcome {
+    /// Whether this outcome should tear the daemon down.
+    ///
+    /// Only a closed channel does. A [`TimedOut`](Self::TimedOut) send means
+    /// the command channel is backed up, which says nothing about daemon
+    /// liveness — the daemon reports its own health through *inbound*
+    /// heartbeats (`last_heartbeat`), and the 30 s inbound threshold already
+    /// catches a genuinely dead daemon. Treating a transient send timeout as
+    /// death would let one slow WS write burst falsely disconnect a live
+    /// daemon and tear down all of its dataflows.
+    fn should_disconnect(&self) -> bool {
+        matches!(self, HeartbeatSendOutcome::ChannelClosed)
+    }
+}
+
 async fn start_inner(
     events: impl Stream<Item = Event> + Unpin,
     clock: Arc<HLC>,
@@ -2026,18 +2060,38 @@ async fn start_inner(
                         disconnected.insert(machine_id.clone());
                         continue;
                     }
-                    let result: eyre::Result<()> = tokio::time::timeout(
+                    let outcome = match tokio::time::timeout(
                         Duration::from_millis(500),
                         send_heartbeat_message(connection, clock.new_timestamp()),
                     )
                     .await
-                    .wrap_err("timeout")
-                    .and_then(|r| r)
-                    .wrap_err_with(|| {
-                        format!("failed to send heartbeat message to daemon at `{machine_id}`")
-                    });
-                    if let Err(err) = result {
-                        tracing::warn!("{err:?}");
+                    {
+                        Ok(Ok(())) => HeartbeatSendOutcome::Delivered,
+                        Ok(Err(err)) => {
+                            tracing::warn!(
+                                "heartbeat send channel to daemon at `{machine_id}` is closed: {err:?}"
+                            );
+                            HeartbeatSendOutcome::ChannelClosed
+                        }
+                        Err(_elapsed) => {
+                            // The 500 ms budget elapsed while the *bounded*
+                            // command channel was full — backpressure from a
+                            // large spawn or a burst of control commands, not a
+                            // dead daemon. Do not tear the daemon down: it is
+                            // proven alive by its recent inbound heartbeat
+                            // (checked just above against the 30 s threshold),
+                            // and disconnecting here would fail all of its
+                            // dataflows over transient congestion.
+                            tracing::warn!(
+                                "heartbeat send to daemon at `{machine_id}` timed out \
+                                 (command channel full); not disconnecting a live daemon \
+                                 (last inbound heartbeat {:?} ago)",
+                                connection.last_heartbeat.elapsed()
+                            );
+                            HeartbeatSendOutcome::TimedOut
+                        }
+                    };
+                    if outcome.should_disconnect() {
                         disconnected.insert(machine_id.clone());
                     }
                 }
@@ -4706,6 +4760,29 @@ mod tests {
     use std::collections::HashMap;
     use tokio::time::{Duration as TokioDuration, timeout};
     use uuid::Uuid;
+
+    // A heartbeat that could not be delivered because the *bounded command
+    // channel* was momentarily full (a `TimedOut` outcome) must NOT disconnect
+    // the daemon: the daemon reports its own liveness through inbound
+    // heartbeats, and the 30 s inbound threshold already catches a genuinely
+    // dead one. Only a closed send channel is a real death signal. Regression
+    // for a transient control-command burst falsely tearing down a live
+    // daemon's dataflows.
+    #[test]
+    fn heartbeat_timeout_does_not_disconnect_a_live_daemon() {
+        assert!(
+            HeartbeatSendOutcome::ChannelClosed.should_disconnect(),
+            "a closed heartbeat send channel must disconnect the daemon"
+        );
+        assert!(
+            !HeartbeatSendOutcome::TimedOut.should_disconnect(),
+            "a transient command-channel backpressure timeout must not disconnect a live daemon"
+        );
+        assert!(
+            !HeartbeatSendOutcome::Delivered.should_disconnect(),
+            "a delivered heartbeat must not disconnect the daemon"
+        );
+    }
 
     fn test_running_dataflow(
         dataflow_id: DataflowId,
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2193,9 +2193,8 @@ async fn start_inner(
                         clock.new_timestamp(),
                     ));
                 }
-                for (machine_id, result) in join_all(heartbeats).await {
-                    if let Err(err) = result {
-                        tracing::warn!("{err:?}");
+                for (machine_id, disconnect) in join_all(heartbeats).await {
+                    if disconnect {
                         disconnected.insert(machine_id);
                     }
                 }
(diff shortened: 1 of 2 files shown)


phil-opp:

This is the lead PR of an automated review sweep of `origin/main`. Each finding below was verified against current `main` before acting; branches were cut fresh off `origin/main`. Machine-generated — please review carefully.

### PRs opened (by category)

**Correctness / robustness**
- **#2886** `fix(coordinator)` — a heartbeat-send timeout on the *bounded command channel* (backpressure) no longer disconnects a live daemon and tears down its dataflows; only a genuinely closed channel does.
- **#2887** `fix(node)` — the `Stop` shutdown signal is never evicted from the full non-input event queue, so a busy node still observes shutdown instead of being force-killed.
- **#2888** `fix(record-node)` — invalid `DORA_RECORD_TOPICS` ids fail cleanly at startup (`FromStr`) instead of panicking mid-recording via the `NodeId::from(String)` / `DataId::from(String)` panicking conversions.
- **#2889** `fix(core)` — `merge_env` normalizes an empty merged env to `None` uniformly, honoring its documented contract for the `env: {}` / empty-global cases.

### Validation

- Per-branch `fmt` + `clippy -D warnings` + affected-crate tests: all green.
- Combined-diff `/review` (correctness) and `/simplify` (reuse / simplification / efficiency / altitude) gates: clean. The one `/simplify` note (the `HeartbeatSendOutcome` enum being "one bit") was considered and kept — it provides an isolated regression-test seam for "a timeout must not disconnect" inside an otherwise-untestable 2000-line event loop, matching this repo's regression-test culture.
- Combined-diff test run over the affected-crate closure (`dora-coordinator`, `dora-node-api`, `dora-record-node`, `dora-core`, `dora-daemon`): **735 passed, 0 failed**. The full-workspace run additionally hit the pre-existing, environment-only failure `zenoh_schema_cache::advanced_pub_cache_serves_late_joining_subscriber` (`Address family not supported by protocol` — the sandbox lacks IPv6 for zenoh's TCP listener bind); unrelated to these changes.

### Findings reviewed but **not** turned into PRs (with reason)

- **Coordinator heartbeat handler head-of-line blocking** — awaits each daemon's send + `list_dataflows` recovery scan serially on the hot tick. Real, but a perf/architecture refactor of the event loop; higher risk than value, deferred.
- **`enable_debug_inspection` forces serialize + Zenoh publish per output** — extra work on an *unstable* debug feature with no subscriber; acceptable cost, not clearly a bug.
- **`break_input` sends a spurious `InputClosed` when the node has no `open_inputs` entry** — genuine guard asymmetry vs `close_input`, but not reachable from current callers (the node is always present in `open_inputs`); latent only.
- **`resolve_daemon` places all deploy-less nodes on the first unnamed daemon** — likely the intended single-daemon default; changing placement needs maintainer intent.
- **Non-input lifecycle events other than `Stop` (`InputClosed`/`Error`/…) can still be dropped on overflow** — #2887 fixes the highest-value case (`Stop`); broadening immunity to all lifecycle events risks unbounded queue growth and is a separate design question.
- **Replay pacing saturates to a ~584-year sleep for a tiny `DORA_REPLAY_SPEED`** — self-inflicted extreme misconfiguration ("replay very slowly" is the requested behavior).
- **Replay timing uses record-side wall-clock receipt time, not the stored `uhlc` message timestamp** — a fidelity/design tradeoff; using the stored clock is a larger change needing a design decision.
- **record-node random header `dataflow_id` + literal `version: 1`** — metadata-only; `FORMAT_VERSION` isn't re-exported, low value.
- **ros2-bridge fixed-array default offsets cast `usize as i32`** — only reachable with >2 GB array sizes from message defs; defensive-only, very low.
- **CLI `calculate_char_sum` normalizes by UTF-8 byte length, not char count** — cosmetic log-colorizer skew, no correctness impact.
- **`dora graph` mislabels an unprefixed runtime-node output as "missing"** — display-only.

---

phil-opp:

Reviewed the diff — the fix looks correct. Walking the disconnect path in `binaries/coordinator/src/lib.rs`:

- A genuinely dead daemon is still caught: the 30 s inbound-heartbeat threshold (`last_heartbeat.elapsed() > Duration::from_secs(30)` with `continue`) runs *before* the send, so liveness detection doesn't depend on the send outcome.
- No hang or leak introduced: the send is still bounded by the 500 ms `tokio::time::timeout`, and the channel is a `tokio::sync::mpsc::Sender`, whose `send().await` is cancel-safe — a timed-out (dropped) send future enqueues nothing and consumes no permit, so repeated `TimedOut` ticks don't accumulate buffered heartbeats.
- The `Ok(Err(_)) → ChannelClosed` classification is sound: on the heartbeat path the send can only error on a closed channel (serializing the `Heartbeat` enum + timestamp and the `str::from_utf8` on serde_json output cannot realistically fail).

Two small notes, neither blocking:

- Behavioral tradeoff worth a line in the doc comment: if a daemon keeps emitting *inbound* heartbeats but its command channel stays wedged indefinitely, it now remains "connected" forever while unable to receive any command (spawn/stop/etc.). That's a strict improvement over disconnecting on a single 500 ms hiccup, but it shifts the only backpressure-death signal onto `ChannelClosed`; a persistently-full command channel never proactively tears the daemon down.
- `heartbeat_timeout_does_not_disconnect_a_live_daemon` is close to tautological — it `matches!`-checks the three variants, mirroring the function's own body. It pins the mapping as a regression seam (the stated intent), but its blast radius is narrow: it won't catch a regression in the surrounding classification logic in the event loop (the `match` arms mapping timeout/err/ok → outcome), which is the part that actually had the bug.

---

phil-opp:
The reasoning holds and I checked the two ways this could have been unsafe, both clean: a genuinely dead daemon is still disconnected, because the inbound check at `binaries/coordinator/src/lib.rs:2096` runs *before* the send and the daemon emits its own heartbeat on an independent timer; and dropping the timed-out `send` future cannot leave a torn WebSocket frame, since `sender` is a `tokio::sync::mpsc::Sender<String>` and the writer task owns the socket. Timeout math is monotonic and saturating.

The residual gap: `TimedOut` now only logs, with no consecutive-failure counter. A command channel that stays full produces a warning every heartbeat tick forever — and `DaemonConnection::send_and_receive` (`binaries/coordinator/src/state.rs:174`) awaits `self.sender.send(json)` with **no** timeout, so the next `dora stop` / `dora start` routed to that daemon blocks the coordinator's single event loop indefinitely. Disconnect was previously the only thing that stopped new commands being routed there. Escalating after N consecutive timeouts would preserve the intent without giving up the backstop.

Also worth stating plainly in the body: the test is a pure enum-classification assertion and exercises none of the event-loop path it protects, so it cannot go RED against the old code.

phil-opp:
Thanks — the residual gap is real and I've addressed it in `0c78de5`.

You're right that never disconnecting on timeout removed the only backstop against a wedged command channel: `send_and_receive` (`state.rs`) awaits the send with no timeout, so a persistently-full channel would block the event loop on the next command routed to that daemon (and warn every tick forever).

Fix: a per-connection consecutive-timeout counter (`consecutive_heartbeat_send_timeouts`), reset on a successful send, that escalates to a disconnect after `MAX_CONSECUTIVE_HEARTBEAT_SEND_TIMEOUTS = 10` (~30 s at the 3 s interval, matching the inbound liveness horizon). Transient bursty backpressure is still tolerated; a genuinely wedged channel is torn down.

I also folded the decision into a pure `heartbeat_disconnect_decision(outcome, streak) -> (streak, disconnect)` helper, so the unit test now exercises the actual policy (tolerate transient → escalate persistent → reset on delivery → immediate disconnect on channel-closed) instead of the earlier static-classification assertion you flagged as unable to go RED. The PR body is updated to state this.

---

phil-opp:
Rebased onto the latest `main` (`de2724e`). `main` had moved ~110 commits since this branch was cut, so I rebased to avoid a stale-base semantic conflict in the merge queue (the same kind that failed the earlier batch this PR was grouped with — that failure was an unrelated PR's `RunningDataflow` field, not this change). The only conflict was in the coordinator test module, where a concurrently-merged PR added tests next to this PR's heartbeat test; resolved by keeping both.

Re-verified on the fresh base: `cargo test -p dora-coordinator` (126 passed), `clippy -D warnings`, and `fmt --check` all green.

---

phil-opp:

Re-reviewed after the escalation commit (`0c78de5`, now on `de2724e`). The consecutive-timeout backstop closes the gap from the prior round: `heartbeat_disconnect_decision` disconnects on `ChannelClosed`, resets the streak on `Delivered`, and escalates a persistent `TimedOut` streak to a disconnect at `MAX_CONSECUTIVE_HEARTBEAT_SEND_TIMEOUTS` (~30s at the 3s interval). A wedged command channel can therefore no longer block the event loop indefinitely on the next routed command, while bursty backpressure is still tolerated. `saturating_add` is overflow-safe and the streak resets on both `Delivered` and `ChannelClosed`.

The reworked unit tests now drive the actual policy (tolerate transient → escalate at threshold → reset on delivery → disconnect on close) rather than a static classification, so they can go RED against the pre-fix behavior. The only untested seam is the event-loop `match` mapping the raw `timeout()`/`Ok(Err)`/`Ok(Ok)` results into `HeartbeatSendOutcome` — where the original bug lived — but that mapping is simple; a note, not a request. No new issues found.

---

phil-opp:
🤖 **Fully automated review by Claude** (no human in the loop — treat as advisory). The escalation logic itself was already validated in the 2026-08-11 round and I have no new concern there; this note is only about a rebase hazard that appeared since, now that the PR is `needs-rebase`.

The rebase is not mechanical. Since this branch was last based on `de2724e`, `main` refactored the coordinator heartbeat watchdog: the serial inline `match tokio::time::timeout(...)` loop this diff edits no longer exists — it's now the async helper `send_heartbeat_with_timeout(...) -> (DaemonId, eyre::Result<()>)` (`lib.rs:3208`) dispatched concurrently via `join_all` (`lib.rs:2146-2168`).

The trap: that helper flattens the send outcome with `.wrap_err("timeout").and_then(|r| r)` into a single `eyre::Result<()>`, and the caller disconnects on any `Err`. That flattening erases exactly the `Ok(Err) → ChannelClosed` vs `Err(_elapsed) → TimedOut` distinction this PR relies on. Re-applying the fix means re-introducing the `Delivered`/`ChannelClosed`/`TimedOut` split and threading the `consecutive_heartbeat_send_timeouts` streak through the new helper — and if the two error cases get collapsed during the rebase, the original "any send error disconnects a live daemon" bug comes back silently. Worth a small test over that raw-result → `HeartbeatSendOutcome` mapping while rebasing, since that seam isn't covered by the existing unit tests.

The disconnect behavior once re-applied still looks correct (independent inbound 30s liveness check, immediate disconnect on `ChannelClosed`, streak escalation bounding a wedged channel to ~30s) — genuinely-dead daemons stay detectable.

---

phil-opp:

_Fully automated review — no human in the loop; treat as advisory._

Re-reviewed after the 2026-08-20 rebase that resolved the `needs-rebase` flagged in my previous pass. The rebase hazard I warned about (the new `send_heartbeat_with_timeout` helper on main flattens the send outcome via `.wrap_err("timeout").and_then(|r| r)`, erasing the `Ok(Err) → ChannelClosed` vs `Err(_elapsed) → TimedOut` distinction) was handled correctly: the helper now returns `HeartbeatSendOutcome` directly, the `match` arms preserve the distinction explicitly, and the `consecutive_heartbeat_send_timeouts` streak is threaded through the `join_all` caller with the same escalation semantics. Diff since my last review LGTM — no new issues; the earlier issues raised on this PR were already addressed prior to the rebase.

---

phil-opp:

The classification split and escalation counter look correct, and the genuinely-dead-daemon paths still hold: the 30s inbound-heartbeat check runs before the send (independent of the send outcome), `ChannelClosed` disconnects immediately, and the send only errors when the WS-writer receiver is dropped.

One caveat on the new escalation backstop that I don't think earlier reviews captured. The `TimedOut` streak only advances when a `DaemonHeartbeatInterval` tick is actually processed, but `send_and_receive` awaits the bounded command channel with no timeout. So if a control command (`dora stop`/`start`/etc.) is routed to a wedged daemon *before* the streak reaches `MAX_CONSECUTIVE_HEARTBEAT_SEND_TIMEOUTS`, the single event loop blocks inside that send — no further heartbeat ticks are processed, and the escalation counter can never reach the threshold. So the counter reliably tears down a wedged channel only when no command happens to be routed to that daemon during the ~30s window; in the "command routed first" case the loop still blocks until the channel drains (or the daemon's WS writer exits and we finally observe `ChannelClosed`).

That's the residual reverse-failure surface for a half-dead-but-inbound-alive daemon (TCP stalled, socket not closed, still emitting inbound heartbeats). The real fix — a timeout on `send_and_receive` — is out of scope for this PR, so this isn't a blocker; but I'd frame the backstop as bounding the *no-traffic* case rather than "preserves that backstop" unconditionally, since a routed command can still wedge the loop indefinitely.

---

phil-opp:

Confirming this is **not superseded** — the bug is live on `main`. `send_heartbeat_with_timeout` (`binaries/coordinator/src/lib.rs:3345-3358`) still collapses both failure shapes into one error with `.wrap_err("timeout").and_then(|r| r)`, and the caller at `:2196-2201` still disconnects on either. #3189 only made the sends concurrent; it did not classify them.

The conflict is mechanical: `DaemonConnection` in `binaries/coordinator/src/state.rs` gained a `zenoh_listen_endpoint` field immediately after `connection_id`, exactly where this inserts `consecutive_heartbeat_send_timeouts`, plus the matching initializer in `DaemonConnection::new`. Second likely hunk is `binaries/coordinator/src/lib.rs:5613-5615`, where new tests landed directly above `fn test_running_dataflow`. Both are keep-both resolutions.

Two things to fix while rebasing.

**The escalation backstop only covers the no-traffic case**, but `MAX_CONSECUTIVE_HEARTBEAT_SEND_TIMEOUTS`'s doc claims it covers the case unconditionally. `send_and_receive` awaits `self.sender.send(json).await` with no timeout, and the streak only advances when a `DaemonHeartbeatInterval` tick is processed. The coordinator's event loop is single-threaded, so if any control command (`dora start`/`stop`/`logs`) is routed to the wedged daemon before the streak reaches 10, the loop blocks inside that send, no further ticks are processed, and the counter can never reach the threshold. Pre-fix, one 500 ms timeout disconnected the daemon and stopped routing there — so this opens a new unbounded head-of-line-blocking window that did not exist before. Either bound the enqueue in `send_and_receive` with a timeout (the real fix, probably out of scope here), or reword the const's doc to say it bounds the no-traffic case only and say so in the PR body rather than "preserves that backstop". This was raised in the last automated comment and has not been addressed in code.

**`ChannelClosed` is applied to every `Err`, not just a closed channel.** `send_heartbeat_message` (`binaries/coordinator/src/handlers.rs:213-217`) also returns `Err` on a `serde_json::to_vec` failure, so the new `"heartbeat send channel to daemon at `{machine_id}` is closed"` warning would be wrong there. The disconnect decision is unchanged from `main`, so this is cosmetic — but match on the actual cause or reword the log.

On tests: the two added tests exercise only `heartbeat_disconnect_decision`, a pure function this PR introduces, so they would pass against broken wiring. Nothing tests that `send_heartbeat_with_timeout` maps `Ok(Err(_))` to `ChannelClosed` and `Err(Elapsed)` to `TimedOut`, and nothing tests that the `join_all` loop applies the returned decision — re-flattening both outcomes back into "disconnect" at the helper would leave every test green. Two additions would close it: one driving `send_heartbeat_with_timeout` against a `DaemonConnection` whose receiver was dropped (expect `ChannelClosed`), and one against a pre-filled capacity-1 channel under `#[tokio::test(start_paused = true)]` (expect `TimedOut`) — the same technique `send_topic_frames_closes_subscriber_after_100_consecutive_timeouts` already uses.

Given this can keep a wedged daemon attached for up to 30 s while the event loop may block on a routed command, I would like to land it on its own rather than batched, so a nightly regression is attributable.

phil-opp:
Thanks for the thorough pass — all addressed in `d3fdc182` (rebuilt on current `main`, since the heartbeat path had been refactored into the concurrent `send_heartbeat_with_timeout`/`join_all` form and `DaemonConnection` gained `zenoh_listen_endpoint`; both were keep-both as you predicted).

**1. Backstop only covers the no-traffic case.** You're right, and I've stopped claiming otherwise rather than expanding scope. The const doc and code comments now state plainly that the escalation bounds only the no-traffic wedge, and that a command routed to a wedged daemon still blocks the event loop because `send_and_receive` awaits the channel with no timeout — narrowing the pre-fix immediate backstop. The PR body says the same and frames the `send_and_receive` timeout as a separate follow-up (it changes command-dispatch failure semantics; happy to open it on its own if you'd like — that's the real fix and deserves its own discussion). I kept the escalation because it still disconnects a send-wedged-but-inbound-alive daemon that the 30 s inbound threshold never catches.

**2. `ChannelClosed` applied to every `Err`.** Renamed the variant to `SendFailed` and reworded the log to "heartbeat send to daemon … failed: {err}", so a `serde_json::to_vec` failure is no longer mislabeled as a closed channel. Decision is unchanged (either way → immediate disconnect).

**3. Tests didn't exercise the wiring.** Added two, mirroring `send_topic_frames_closes_subscriber_after_100_consecutive_timeouts`:
- `send_heartbeat_with_timeout_disconnects_on_closed_channel` — dropped receiver → expect disconnect (`SendFailed`).
- `send_heartbeat_with_timeout_tolerates_a_full_channel` — pre-filled capacity-1 channel under `#[tokio::test(start_paused = true)]` → expect a tolerated timeout that bumps the streak to 1, no disconnect.

Re-flattening both outcomes back into "disconnect" at the helper now fails these, not just the pure-function tests.

**Landing on its own:** agreed — please don't batch it; I'll keep it rebased on `main` so a nightly regression stays attributable to this change alone.

---

PR #2887, opened 2026-07-29
fix(node): never evict the Stop event from a full non-input event queue

phil-opp:
## Issue

In the node event-stream scheduler, all non-input events — `Stop`, `InputClosed`, `InputRecovered`, `NodeRestarted`, `NodeFailed`, `Error`, `Reload`, param updates, … — are buffered in a **single** queue keyed by `NON_INPUT_EVENT`, capped at 1000 (`event_stream/mod.rs`). When that queue overflows, `select_eviction` dropped the oldest **non-correlated** event:

```rust
if let Some(idx) = queue.iter().position(|e| !is_correlated(e)) {
    return Eviction::RemoveAt(idx);
}
```

Every non-input event — including `Stop` — is non-correlated (`is_correlated` only inspects `Input`/`ZenohInput` metadata). So a node busy in user code (not draining its stream) while a flood of non-input events arrives could have its buffered `Stop` silently evicted. The node then never observes shutdown and runs until the daemon's force-kill grace deadline. The old loud-drop fallback (`pop_front`) had the same hazard for a front-positioned `Stop`.

## Fix

Make `Stop` eviction-immune, mirroring the existing correlation protection:
- `select_eviction` prefers an ordinary (non-correlated, non-`Stop`) victim.
- The loud correlation-drop fallback now removes the oldest **correlated** event *by index* (`DropCorrelatedLoud(idx)`) rather than blindly popping the front, so a `Stop` at the front is never lost.
- An incoming `Stop` is likewise admitted into a full queue instead of being dropped as the overflow victim.

This is bounded — at most one `Stop` is ever in flight — so it cannot make the queue grow without limit (every eviction outcome still removes exactly one element or drops the incoming, preserving the `len == cap` post-condition).

## Validation

- New regression tests: `stop_survives_non_input_queue_overflow` (buffered `Stop` survives a 100-event overflow) and `incoming_stop_is_admitted_into_a_full_non_input_queue`.
- All 18 scheduler tests pass; `cargo clippy -p dora-node-api -- -D warnings` and `cargo fmt --check` clean.

---

🤖 This is a machine-generated pull request opened by Claude Code as part of an automated codebase review. Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DeK3nXPz84sLjc8eTKPnbb)_

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -264,9 +264,14 @@ impl Scheduler {
         // `goal_status` correlations whose senders are waiting for them
         // (dora-rs/adora#146). Use the same correlation predicate that the
         // drop_oldest path uses and retain correlated events across the flush.
+        // Also retain the `Stop` shutdown signal (eviction-immune everywhere,
+        // see `is_stop`): flush normally targets a per-input queue and `Stop`
+        // lives under `NON_INPUT_EVENT_ID`, but the two collide if an input is
+        // literally named `dora.non_input_event`, which `validate_data_id`
+        // permits — so guard the flush path too rather than rely on that.
         if should_flush && let Some((_size, queue)) = self.event_queues.get_mut(event_id) {
             let before = queue.len();
-            queue.retain(is_correlated);
+            queue.retain(|e| is_correlated(e) || is_stop(e));
             let drained = before - queue.len();
             if drained > 0 {
                 tracing::debug!(
@@ -750,6 +755,28 @@ mod tests {
         assert!(
             non_input.iter().any(is_stop),
             "an incoming Stop must be admitted into a full non-input queue"
+        );
+    }
+
+    // The flush path (`retain`) is a second eviction site. It normally targets
+    // a per-input queue, but an input literally named `dora.non_input_event`
+    // (which `validate_data_id` permits) collides with the queue where `Stop`
+    // lives. Flush must still preserve the `Stop` shutdown signal there.
+    #[test]
+    fn flush_retains_stop_when_targeting_the_non_input_queue() {
+        let (mut sched, _id) = make_scheduler(10);
+
+        sched.add_event(make_stop());
+        sched.add_event(make_input_closed("x"));
+
+        let mut flush_params = MetadataParameters::new();
+        flush_params.insert(FLUSH.into(), Parameter::Bool(true));
+        sched.add_event(make_input(NON_INPUT_EVENT, flush_params));
+
+        let non_input = &sched.event_queues[&*NON_INPUT_EVENT_ID].1;
+        assert!(
+            non_input.iter().any(is_stop),
+            "flush must not evict the Stop shutdown signal"
         );
     }
 
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -37,31 +37,61 @@ fn is_correlated(event: &EventItem) -> bool {
     carries_pattern_correlation(params)
 }
 
+/// Returns `true` if the event is the daemon's [`Stop`][NodeEvent::Stop]
+/// shutdown signal.
+///
+/// `Stop` must never be silently evicted from a full queue: dropping it makes
+/// the node miss shutdown entirely and keep running until the daemon's
+/// force-kill grace deadline. Since at most one `Stop` is ever in flight,
+/// making it eviction-immune cannot let the queue grow unbounded.
+fn is_stop(event: &EventItem) -> bool {
+    matches!(
+        event,
+        EventItem::NodeEvent {
+            event: NodeEvent::Stop,
+            ..
+        }
+    )
+}
+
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The eviction rewrite stays bounded (exactly one removal per admission) and correctly makes the `Stop` event immune: `select_eviction` prefers an ordinary (non-correlated, non-`Stop`) victim, then an ordinary incoming event, and only as a last resort drops the *oldest correlated* event by index — never `Stop` — falling back to `DropIncoming` when a `Stop` is all that remains.

One observation, not blocking: the two new tests only exercise the branch where an ordinary victim exists (`RemoveAt`). The new `DropCorrelatedLoud(idx)` path at `scheduler.rs:300-304` — a full queue of correlated events plus a `Stop`, receiving another correlated event — isn't covered, so the "drop oldest correlated, keep `Stop`" behavior isn't pinned by a test. A case that fills the queue with correlated events and asserts the `Stop` still survives would close that gap.

---

phil-opp:
I reviewed the diff and found one issue with the added tests.

`incoming_stop_is_admitted_into_a_full_non_input_queue` does not actually guard the fix. It fills the non-input queue with 10 non-correlated `InputClosed` events and then adds a `Stop`. With this PR's change reverted, the old `select_eviction` still returns `RemoveAt(idx)` for the first non-correlated event (an `InputClosed` at index 0) and admits the incoming `Stop`, so `non_input.iter().any(is_stop)` is already `true` — the test passes on the unpatched code.

The case where the old code actually drops an incoming `Stop` is when the queue is *entirely correlated* (then `!is_correlated(incoming)` → `DropIncoming`), and that can't happen in the non-input queue: correlated `Input`/`ZenohInput` events route to their per-input queues, never here. So this test exercises a path that already behaved correctly before the change.

The sibling test `stop_survives_non_input_queue_overflow` is the one that genuinely pins the fix — with the change reverted, the old code evicts the front element (which is the `Stop`, since `Stop` is non-correlated) on the first overflow, so that test does fail on revert.

Suggestion: either drop the redundant test, or rework it to cover the genuinely new `DropCorrelatedLoud(idx)` branch, which is the only truly new behavior and is currently untested (also noted in the earlier review).

---
🤖 This is a fully automated review by Claude (Claude Code). No human has vetted this comment; please treat it as advisory.

---

phil-opp:

No issues found.

Traced the eviction/scheduler logic end to end:

- `is_stop` is excluded from the victim predicate in both eviction steps — step 1 (`!is_correlated(e) && !is_stop(e)`) and step 3 (`!is_stop(e)`) — so `Stop` is never selected as an eviction target. An incoming `Stop` is never dropped: the `DropIncoming` guard excludes it, and step 1 evicts an ordinary event to make room.
- The `DropCorrelatedLoud(idx)` handler uses `queue.remove(idx)` (skipping a front `Stop`) instead of the old blind `pop_front`, and control falls through to `push_back`, so the incoming event is still admitted after eviction.
- Bounded by design: at most one `Stop` is in flight.

`stop_survives_non_input_queue_overflow` is a real regression test — under the old logic the front `Stop` is the oldest non-correlated event and would be evicted first, so it fails pre-change. `incoming_stop_is_admitted_into_a_full_non_input_queue` passes under both old and new logic, so it's a weaker guard rather than a strict regression test, but it's not vacuous.

---

phil-opp:
Boundedness is verified — every branch either removes an element or drops the incoming one, including the degenerate all-`Stop` queue, so this cannot grow unbounded. Two notes:

**One eviction path is not covered.** `apis/rust/node/src/event_stream/scheduler.rs:239` — the flush path does `queue.retain(is_correlated)`, which would wipe a buffered `Stop`, since `is_correlated` is false for it. Unreachable in practice only because flush targets `event_queues[input_id]` while `Stop` lives under `NON_INPUT_EVENT_ID`, and the two collide iff an input is literally named `dora.non_input_event` (which `validate_data_id` allows). Contrived, but it is the one eviction site the new guard does not cover, and `retain(|e| is_correlated(e) || is_stop(e))` closes it for free.

**The second test is not RED.** `incoming_stop_is_admitted_into_a_full_non_input_queue` passes on the old code too: old `select_eviction` finds the `InputClosed` at index 0 and does `RemoveAt(0)`, then pushes the `Stop`. It is a valid non-regression guard, just not coverage of the fix. (`stop_survives_non_input_queue_overflow` genuinely is RED — old code evicts the `Stop` on the first overflow.)

phil-opp:
Thanks — both notes are right.

- **Flush path:** fixed in `ef011f6`. `retain` now keeps `Stop` too (`is_correlated(e) || is_stop(e)`), closing the `dora.non_input_event` collision you identified. Added a regression test that drives the collision and is RED against the prior `retain(is_correlated)`.
- **Second test:** agreed — `incoming_stop_is_admitted_into_a_full_non_input_queue` passes on the old code too, so it's a non-regression guard rather than fix coverage. `stop_survives_non_input_queue_overflow` (and now the flush test) are the ones that actually go RED against the old code. I've kept the admission test as a guard.

---

PR #2888, opened 2026-07-29
fix(record-node): validate topic ids at startup instead of panicking mid-run

phil-opp:
## Issue

The record node parses `DORA_RECORD_TOPICS` (`{ "input_id": "source_node/source_output" }`) into source `(node, output)` id strings, but only checked for the presence of a `/` — the parts themselves were never validated. The strings were later converted with `NodeId::from(String)` / `DataId::from(String)` **inside the event loop**:

```rust
node_id: NodeId::from(source_node.clone()),
output_id: DataId::from(source_output.clone()),
```

Those are the panicking `From<String>` impls (documented "Panics if `id` contains invalid characters"). A malformed topic — e.g. a source node id containing a space — therefore **panicked the node mid-recording**, when the first matching event arrived, rather than failing cleanly at startup.

## Fix

Parse the ids with `str::parse` (`FromStr`, which validates and returns an error) while building the reverse map, so an invalid topic aborts the node immediately with a clear message and the hot loop no longer touches a panicking conversion. The parsing is extracted into a testable `build_reverse_map`, and the map now stores typed `(NodeId, DataId)` so the hot loop clones them directly.

`FromStr` and `From<String>` call the *same* `validate_node_id`/`validate_data_id`, so the accepted set for valid inputs is byte-for-byte unchanged (including `DataId`'s `/`-containing ids); only the invalid-input behavior changes from panic to error.

## Validation

- New unit tests: `valid_topics_parse`, `missing_slash_is_an_error`, `invalid_node_id_is_an_error_not_a_panic`, `invalid_output_id_is_an_error_not_a_panic`.
- `cargo test -p dora-record-node`, `cargo clippy -p dora-record-node -- -D warnings`, `cargo fmt --check` all pass.

---

🤖 This is a machine-generated pull request opened by Claude Code as part of an automated codebase review. Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DeK3nXPz84sLjc8eTKPnbb)_

### DIFF
--- a/binaries/record-node/src/main.rs
+++ b/binaries/record-node/src/main.rs
@@ -10,6 +10,36 @@ use dora_node_api::{DoraNode, Event, arrow::datatypes::DataType, arrow_utils};
 use dora_recording::{RecordEntry, RecordingHeader, RecordingWriter};
 use eyre::Context;
 
+/// Parse `DORA_RECORD_TOPICS` (`{ "input_id": "source_node/source_output" }`)
+/// into a lookup from the record node's own input id to the validated
+/// `(source_node, source_output)` ids.
+///
+/// The node and output ids are parsed via [`str::parse`] (`FromStr`), which
+/// *validates* the character set and returns an error, rather than the
+/// panicking `NodeId::from`/`DataId::from` (`From<String>`) conversions. A
+/// malformed topic (e.g. a source node id containing a space) therefore fails
+/// the node cleanly at startup instead of panicking mid-run when the offending
+/// event is first recorded.
+fn build_reverse_map(topics_json: &str) -> eyre::Result<HashMap<String, (NodeId, DataId)>> {
+    let topic_map: HashMap<String, String> =
+        serde_json::from_str(topics_json).wrap_err("failed to parse DORA_RECORD_TOPICS")?;
+
+    let mut reverse_map: HashMap<String, (NodeId, DataId)> = HashMap::new();
+    for (input_id, source) in &topic_map {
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. `NodeId`/`DataId`'s `FromStr` validate the id charset and return `InvalidId` (`libraries/message/src/id.rs:94-101, 180-187`), whereas `From<String>` panics (`id.rs:114-118, 206-210`). Moving the topic-id parsing into `build_reverse_map` via `str::parse` therefore genuinely converts a mid-recording panic into a clean startup error, and the record hot loop no longer touches a panicking conversion. Valid inputs are unaffected.

The four unit tests cover the valid, missing-slash, invalid-node-id, and invalid-output-id cases, and each would fail if the panicking conversions were reintroduced or the validation dropped.

---

phil-opp:

No issues found.

Confirmed the core claim against the id types: `FromStr for NodeId`/`DataId` (`libraries/message/src/id.rs:94,180`) run `validate_node_id`/`validate_data_id` and return `Err`, whereas the previously-used `From<String>` panics on any out-of-charset character. Moving the parse into `build_reverse_map` with `str::parse` means a malformed topic now aborts the node cleanly at startup, and the event loop stores/uses already-validated ids (no more panicking conversions in the hot path).

The tests are sound: a space is outside both id charsets, so `"my node/image"` and `"camera/bad output"` genuinely hit the node-id and output-id validation branches respectively, and each asserts the specific error context — they'd fail if the old panicking conversion were still in place.

---

PR #2892, opened 2026-07-30
fix(cli): surface coordinator startup stderr in dora up

DGHX12345:
## Issue

`dora up` could hide the real coordinator startup failure when a persisted coordinator redb file was incompatible with the current binary.

The coordinator redb backend intentionally rejects old schema versions at startup. For example, a user upgrading from older rc builds can still have `~/.dora/coordinator.redb` stamped with schema v2, while the current binary expects schema v3. In that case `dora coordinator` exits immediately with `redb schema version mismatch`.

Before this change, `dora up` spawned the coordinator as a detached child with stderr redirected to null. The child could exit with the actionable schema mismatch error, but the parent CLI only kept retrying the WebSocket connection until it timed out. Users then saw a generic “timed out waiting for coordinator to start” message instead of the real reason.

## Fix

- Capture `dora coordinator` stderr in a temporary file while `dora up` waits for the coordinator to become ready.
- Poll the coordinator child process during startup so an early exit is detected directly.
- Include captured coordinator stderr when the coordinator exits before readiness.
- Include captured coordinator stderr on startup timeout if the child produced any output.
- Remove the temporary stderr capture file on success, early exit, and timeout paths.
- Keep the daemon startup path unchanged.
- Add a regression test that creates an isolated HOME with a legacy v2 `coordinator.redb` and verifies `dora up` reports `redb schema version mismatch`.

## Validation

- Reproduced the issue with:
  - `cargo test --test ws-cli-e2e e2e_up_reports_redb_schema_mismatch_from_coordinator -- --test-threads=1 --nocapture`
- Verified the fix with:
  - `cargo test --test ws-cli-e2e e2e_up_reports_redb_schema_mismatch_from_coordinator -- --test-threads=1 --nocapture`
  - `cargo test --test node-lifecycle-e2e lifecycle_rust_dynamic_add_remove -- --test-threads=1 --nocapture` with isolated `HOME`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2150,6 +2150,7 @@ dependencies = [
  "eyre",
  "futures",
  "libloading 0.9.0",
+ "redb",
  "serde_json",
  "serde_yaml",
  "tempfile",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -203,6 +203,7 @@ tracing = { workspace = true }
 futures = { workspace = true }
 tokio-stream = "0.1.18"
 tempfile = { workspace = true }
+redb = "2"
 assert2 = "0.4.0"
 tokio-tungstenite = "0.29"
 libloading = "0.9"
--- a/binaries/cli/src/command/up.rs
+++ b/binaries/cli/src/command/up.rs
@@ -10,6 +10,7 @@ use eyre::{Context, ContextCompat, bail};
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::{fs, net::SocketAddr, path::Path, process::Command, time::Duration};
+use uuid::Uuid;
 
 #[derive(Debug, clap::Args)]
 /// Spawn coordinator and daemon in local mode (with default config)
@@ -54,15 +55,8 @@ pub(crate) fn up(config_path: Option<&Path>, auth: bool) -> eyre::Result<()> {
             session
         }
         Err(_) => {
-            start_coordinator(auth).wrap_err("failed to start dora-coordinator")?;
-
-            connect_with_retry(coordinator_addr, Duration::from_secs(10)).map_err(|err| {
-                eyre::eyre!(
-                    "timed out waiting for coordinator to start at {coordinator_addr}: {err}\n\n  \
-                     hint: is port {port} already in use? Check with `dora status`\n  \
-                     or stop the existing coordinator with `dora down`"
-                )
-            })?
+            let startup = start_coordinator(auth)?;
+            wait_for_coordinator_start(coordinator_addr, port, startup)?
         }
     };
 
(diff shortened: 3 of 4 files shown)


phil-opp:

Took a close look at this change. The core fix is sound and the regression test is genuinely meaningful — I verified that `redb-backend` is a default CLI feature and `dora coordinator`'s default `--store` is `redb`, so `dora up` really does open `~/.dora/coordinator.redb`; the fixture's `meta` / `schema_version: u32` layout matches `libraries/coordinator-store/src/redb_store.rs` exactly, and `--quiet` doesn't suppress the propagated `redb schema version mismatch` error, so it reaches the captured stderr. Detachment semantics also look correct: `Child` isn't killed on drop so the coordinator survives, and `process_group(0)` is preserved.

Two minor points, neither blocking:

1. **Temp stderr file can leak on Windows.** On the success and timeout paths, `fs::remove_file(&startup.stderr_path)` runs while the coordinator child still holds the file open as its stderr. On Unix this is a harmless unlink of an open inode, but on Windows the delete typically fails with a sharing violation (the file wasn't opened with `FILE_SHARE_DELETE`) and the error is swallowed by `let _ =`. Since the coordinator is long-lived, that leaves one `dora-coordinator-<uuid>.stderr` file in `%TEMP%` per `dora up` for the whole session. Worth a cleanup-on-coordinator-exit strategy if Windows is a supported target.
2. **Duplication of `detach_process`.** `start_coordinator` re-inlines the body of `detach_process` (stdin/stdout `null` + `#[cfg(unix)] process_group(0)`) just to redirect stderr to a file, which also drops the doc comment explaining the process-group rationale. Consider a `detach_process_with_stderr` variant so the "new process group prevents terminal signals" logic stays in one place.

Otherwise this reads well — capturing the child and polling with `try_wait` is a clean way to surface early-exit causes the old retry-until-timeout loop hid.

---

phil-opp:

No issues found.

Reviewed the stderr-surfacing change with an eye for the classic pitfalls and it holds up:

- **No pipe deadlock.** Coordinator stderr is redirected to a temp file (`cmd.stderr(stderr_file)`), not an OS pipe, so there is no fixed-size pipe buffer that could fill and block the child while the parent isn't draining it. The parent only reads the file after the child exits or the connect deadline passes.
- **No hang.** `wait_for_coordinator_start` polls with non-blocking `child.try_wait()` plus a single-attempt connect, a 50 ms sleep, and a hard 10 s deadline. Every branch either returns, sleeps, or bails.
- **No Windows regression.** The inlined spawn setup (stdin/stdout null, `process_group(0)` under `#[cfg(unix)]`) matches `detach_process` except stderr now goes to the capture file.
- **Temp-file cleanup** happens on all three exit paths (success, early-exit, timeout).

The e2e test drives a real coordinator crash and asserts the crash cause reaches the user's output, which is exactly the behavior the old code dropped — a meaningful regression guard.

---

phil-opp:
Closing as superseded by #2962, which landed the same capability. The version on `main` is stronger on three points this branch is weak on: it keeps a named `~/.dora/coordinator-stderr.log` (this branch unlinks the capture file while the detached coordinator still holds it open, so all later stderr goes to an unlinked inode), it kills and reaps the child on the timeout path, and it handles the concurrent-`dora up` port race by attaching to the winner.

Thanks for the fix — the problem was real, it just got solved on the other thread.

PR #2893, opened 2026-07-30
fix(daemon): ignore stale dynamic node exits

DGHX12345:
## Issue

Dynamic node remove/re-add has a lifecycle race when the same node ID is reused before the old process has fully exited.

`dora node remove --grace ... <node>` removes the node from daemon lifecycle state immediately and asks the old node process to stop. If that process does not exit promptly, `dora node add` can start a new process with the same `node_id` while the old process is still alive. When the old process exits later, the daemon receives its terminal node result keyed only by `node_id`.

Because the exit path did not identify which process incarnation finished, the daemon could treat the stale exit from the old process as the current state of the newly re-added node. That incorrectly marks the new node as `Failed`, removes same-id daemon state, and reports a stale stop to the coordinator even though the replacement process is still running.

This affects hot replacement, online reconfiguration, and dynamic deployment flows that reuse node IDs, especially when the removed node is non-cooperative or slow to handle `Stop`.

## Fix

- Carry the exited child PID through the daemon-internal `NodeProcessFinished` and `DoraEvent::SpawnedNodeResult` path.
- Compare the exited PID against the currently tracked `RunningNode.pid` before classifying a non-restart exit.
- Ignore stale terminal exits when the current same-id node has a different PID.
- Clear the old grace-kill marker when ignoring a stale exit so the replacement incarnation is not polluted by prior stop state.
- Add an end-to-end regression that removes a non-cooperative dynamic node, immediately re-adds the same node ID, waits for the old process to exit, and asserts the re-added node remains `Running`.
- Refactor the node lifecycle E2E startup sequence into a shared helper so the existing lifecycle matrix and the regression use the same startup path.

## Validation

- Reproduced the failure before the fix with:
  - `cargo test --test node-lifecycle-e2e rust_dynamic_node_readd_same_id_ignores_stale_exit -- --nocapture`
- Verified the fix with:
  - `cargo test --test node-lifecycle-e2e rust_dynamic_node_readd_same_id_ignores_stale_exit -- --nocapture`
  - `cargo test --test node-lifecycle-e2e lifecycle_rust_dynamic_add_remove -- --nocapture`
  - `cargo test -p dora-daemon --lib`
  - `cargo clippy -p dora-daemon -- -D warnings`

Note: `cargo clippy -p dora-daemon --tests -- -D warnings` currently fails on an unrelated pre-existing `match_like_matches_macro` lint in daemon tests.

### DIFF
--- a/binaries/daemon/src/event_types.rs
+++ b/binaries/daemon/src/event_types.rs
@@ -179,6 +179,7 @@ pub enum DoraEvent {
         exit_status: NodeExitStatus,
         restart: bool,
         restart_count: u32,
+        pid: u32,
     },
     /// The per-node `restart_loop` spawned a fresh process after an exit
     /// and now wants the daemon to swap the tracked `ProcessHandle` in
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4740,6 +4740,7 @@ impl Daemon {
                 exit_status,
                 restart,
                 restart_count,
+                pid,
             } => {
                 let mut logger = self
                     .logger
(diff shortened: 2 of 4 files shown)


phil-opp:
**Reviewed the full diff — no issues found.** The fix is correct and the added e2e test genuinely pins the regression.

Verification:
- The stale-exit guard runs only for `!restart` results, and a new incarnation stores its PID into a fresh `Arc<AtomicU32>` before the `RunningNode` is inserted (`spawn/prepared.rs`), so `running_nodes[node_id].pid != exited_pid` reliably distinguishes an older same-ID incarnation's exit from the live one's.
- The early `return Ok(())` correctly skips `handle_node_stop`, which is keyed only by `node_id` and would otherwise tear down the *re-added* node's channels/subscriptions — that skip is the point of the fix, not a regression.
- `restart: true` results are unaffected (guard is gated on `!restart`), and an ordinary single-incarnation exit still matches (`current_pid == exited_pid`), so the restart loop and normal shutdown are unchanged. Nodes with no tracked `pid` fall through to unchanged behavior.
- Removing `grace_duration_kills[node_id]` on the stale path is correct cleanup of the *old* incarnation's kill marker.
- PID reuse can't defeat the check in the re-add scenario: the old process is still alive when the new one spawns (it is signal-killed later, after the grace period), so the OS can't hand the new incarnation the old PID.

The test is non-vacuous: `fire-and-forget-source-node` ignores the cooperative stop and is signal-killed only after the 3s grace — i.e. after re-add has installed the new PID — so without the guard the stale exit tears the re-added `filter` down and the `Running` assertion fails. I confirmed the `fire-and-forget-source-node` package and the `examples/rust-dynamic-add-remove` dataflow both exist in the workspace, so the test builds.

---
🤖 *This is a fully automated review by Claude (no human in the loop). It is not a maintainer approval — treat findings as suggestions to verify.*

---

PR #2894, opened 2026-07-30
fix(node): freeze un-acked outputs at the startup grace boundary (#2891)

### LINKED ISSUE #2891: Startup handshake can reorder a topic's messages when an output upgrades daemon→zenoh mid-stream (grace→deadline window)
## Summary

The startup route handshake introduced in **#2806** (commit `e918028`) can violate dora's per-input FIFO (topic) ordering guarantee. An output that is not yet route-proven when `init()` returns starts sending on the reliable **daemon path**, and can then flip to the **direct-zenoh path mid-stream** at any point up to 10 s later. Because the direct-zenoh path has fewer hops than the daemon-relay path and the consumer merges both paths into a single arrival-ordered channel, a message sent right after the switch can overtake an earlier message still in flight on the daemon path.

This is the exact reorder the design deliberately avoids at the 10 s deadline boundary — but the same reorder is reachable throughout the whole 500 ms → 10 s window.

## Why it happens

The producer chooses the transport **per message**, gated only on the output's `ready` flag, with no check that the output is quiescent on the daemon path:

- `apis/rust/node/src/node/mod.rs:1666-1703` — `send_output_sample`: `Some(finalized) if self.output_direct_ready(&output_id)` → direct zenoh, else daemon path. (`output_direct_ready` = `apis/rust/node/src/node/mod.rs:1783-1786`.)
- `AckState::record` flips `ready=true` the instant the required acks arrive (`apis/rust/node/src/node/mod.rs:311-322`), unconditionally.

`init()` returns to user code after only a **500 ms** grace wait, regardless of whether the output is ready:

- `apis/rust/node/src/node/mod.rs:1297-1300` — `arm_deadline()` then `wait_for_grace()`.
- `ZENOH_STARTUP_GRACE = 500ms` (`apis/rust/node/src/node/mod.rs:175`) vs `ZENOH_STARTUP_ACK_TIMEOUT = 10s` (`apis/rust/node/src/node/mod.rs:164`).

The marker thread keeps running and can flip `ready` at any time until the 10 s deadline (`apis/rust/node/src/node/mod.rs:474-497`).

The consumer merges daemon-path `NodeEvent::Input` and direct-zenoh `EventItem::ZenohInput` into the **same** `tokio::mpsc` channel in arrival order — there is no cross-path resequencing:

- Channel created once at `apis/rust/node/src/event_stream/mod.rs:355` (`let (tx, rx) = …`).
- Zenoh callback writes `EventItem::ZenohInput` via `tx_cb = tx.clone()` — `apis/rust/node/src/event_stream/mod.rs:468, 648-652`.
- Both event kinds are drained from the same `self.receiver` — `apis/rust/node/src/event_stream/mod.rs:849-851`.

## Self-inconsistency with the deadline-freeze logic

The code at `apis/rust/node/src/node/mod.rs:484-488` freezes an un-acked output on the daemon path **at the 10 s deadline**, with this exact rationale:

> `// Freeze: no more markers and no late upgrade — a route that produced no ack in 10s ... upgrading minutes in would trade a working (slower) path for a mid-run reorder.`

But user code has been sending since the 500 ms grace expiry, so **any** upgrade after grace is already "mid-run." The freeze correctly recognizes the reorder hazard at `t = 10s`, yet the identical hazard is left open for the entire `0.5s … 10s` window. The reorder-safety claim in the constant's doc comment (`apis/rust/node/src/node/mod.rs:170-172`, "switch-before-first-send ... so cross-path reordering cannot occur") only holds when the handshake completes within grace.

## Failure scenario

- A same-machine producer→consumer topic edge (this only affects direct-zenoh-eligible outputs; cross-daemon outputs are `daemon_only`-pinned and never upgrade).
- The consumer is slow to wire its zenoh subscriber / ack — e.g. heavy per-node init that finishes shortly after the barrier releases — so the marker→ack round trip lands at ~600 ms (past the 500 ms grace, well under 10 s).
- The producer emits a steady stream starting right after `init()` returns.
- Messages `M1…Mk` go daemon path (node→daemon→node); the output then flips `ready`, and `M(k+1)…` go direct zenoh (node→node). `M(k+1)` can arrive at the consumer before `Mk` is relayed → the consumer observes `M(k+1)` before `Mk`.

Severity: **medium** — no data loss or crash, but it silently breaks the per-input FIFO ordering that ordering-sensitive topic consumers rely on. Confidence that the mechanism exists is high; likelihood depends on route-establishment latency exceeding 500 ms, which is unusual locally but plausible under heavy node-startup load or many outputs.

## Suggested fix

Make the reorder-safety condition "has this output already sent on the daemon path since the barrier released?" rather than "is it before the 10 s deadline?". Concretely, one of:

1. **Freeze at the grace boundary, not the deadline.** If an output is still un-acked when `wait_for_grace()` returns, pin it to the daemon path for the rest of the run (never upgrade), matching the deadline-freeze rationale. Simplest; costs the fast path for slow-to-establish local routes.
2. **Gate the upgrade on quiescence.** Only allow the `ready` flip to take effect for the send path once the output has no daemon-path send outstanding (e.g. track a per-output "sent on daemon path" flag; once set, refuse the upgrade — equivalent to option 1 in practice).
3. At minimum, correct the doc comments at `apis/rust/node/src/node/mod.rs:170-172` and `484-488` to state that a mid-run reorder is possible in the grace→deadline window, so the guarantee is not overstated.

Option 1 is the most defensible: it preserves "never reorder" unconditionally and only sacrifices the direct-zenoh fast path for routes that fail to establish within 500 ms.

---

*This issue was filed automatically by a scheduled, automated Claude review of recent commits/PRs (no human authored it). Please verify the analysis before acting; the reasoning was validated against the source at the line references above, but timing-dependent behavior should be confirmed empirically.*

phil-opp:
Implements **option 1** of #2891.

## The bug

The startup route handshake (#2806) picks an output's transport *per message*, gated only on its `ready` flag, and `ready` could flip at any point up to the 10 s `ZENOH_STARTUP_ACK_TIMEOUT`. But `init()` returns to user code after only the 500 ms grace, so an output that acked late upgraded from the daemon path to direct zenoh **mid-stream**.

The direct-zenoh path has fewer hops than the daemon relay, and the consumer merges both into one arrival-ordered channel with no cross-path resequencing — so a message sent just after the switch can overtake an earlier one still in flight, silently breaking the per-input FIFO ordering that topic consumers rely on.

The code already recognised this hazard at the 10 s deadline:

> `// Freeze: … upgrading minutes in would trade a working (slower) path for a mid-run reorder.`

…yet the identical hazard was open across the whole 0.5 s → 10 s window, because user code has been sending since the grace expired.

## The fix

Move the freeze to the **grace boundary**: whatever has not proven its routes when `wait_for_grace` returns is pinned to the daemon path for the rest of the run. Each output's transport is then settled *before* user code sends its first message and never changes, so no message can straddle a path switch and FIFO order holds unconditionally.

- `AckState` gains a `frozen` flag. `record` reads it under the same `received` lock that `freeze` takes, so an ack landing exactly at the boundary either completes the handshake or is ignored — never a half-applied upgrade.
- The marker thread skips frozen outputs, so on a broken route it retires ~9.5 s earlier instead of publishing 5 ms markers until the deadline.
- `ZENOH_STARTUP_ACK_TIMEOUT` (and `arm_deadline`) are removed — the grace now *is* the whole handshake window, and `ZENOH_STARTUP_GRACE` is the single knob for the trade-off.
- `wait_for_grace` is a free function over the ack states, so the boundary is testable without a zenoh session.

**The cost, stated plainly:** a route that fails to establish within 500 ms now loses the direct-zenoh fast path for the whole run, where it previously had until 10 s to upgrade. That is the deliberate trade in option 1 — correctness of ordering over the fast path for slow-to-establish routes. Dynamic/restarted producers are most exposed (their barrier releases immediately, so the window starts a few ms after the zenoh session opens); this is now called out on the constant, and raising `ZENOH_STARTUP_GRACE` is the intended remedy if it shows up as a regression.

## Docs corrected

The protocol docs described the upgrade as unconditional on ack arrival, with no mention of a window — now false. Updated `STARTUP_MARKER_PARAM`, `STARTUP_ACK_PARAM`, `OutputRouting::required_ackers`, the daemon's routing policy, and the consumer-side acker (which claimed ack-route races "self-heal" because markers keep coming — they no longer do past the grace). This subsumes option 3 of the issue.

## Tests

New unit tests in `apis/rust/node/src/node/mod.rs`:

- `ack_state_freeze_blocks_a_late_upgrade` — the direct #2891 regression guard: a frozen output ignores a late ack and `ready` stays `false`.
- `ack_state_freeze_spares_an_output_that_acked_in_time` — an output that completed its handshake keeps direct zenoh.
- `grace_boundary_freezes_unacked_outputs_only` — end-to-end over the boundary: the acked output is untouched, the un-acked one is frozen, and its straggler ack is ignored.
- `grace_returns_early_once_every_output_is_acked` — a completed handshake does not wait out the grace (also covers the empty-states case).

## Validation

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | pass |
| `cargo clippy --all …` (CI-equivalent, `-D warnings`) | pass |
| `cargo test -p dora-node-api -p dora-message` | pass (158 + 140 + 5 + 1) |
| `cargo test -p dora-daemon` | pass (133) |
| `cargo check --examples` | pass |

Two notes on validation, so the gaps are explicit:

- `dora-node-api --test zenoh_schema_cache` fails in my sandbox (`Address family not supported by protocol` binding `tcp/[::]:0` — no IPv6). Verified it fails identically on the unmodified baseline commit, so it is environmental and unrelated to this change.
- The full `cargo test --all` did not complete: the workspace build exceeded this environment's disk allowance. The affected crates were tested individually instead; CI should be treated as the authority for the rest.

## Follow-up considered but not taken

Option 2 of the issue (latch the transport per output at its *first send*, rather than globally at the grace boundary) is strictly less restrictive — it would keep the fast path for outputs first published long after startup, e.g. a `metrics` output first sent 30 s in, whose route needed 600 ms. It gives the same unconditional FIFO guarantee. The issue named option 1 as the most defensible and that is what this PR implements; option 2 is a reasonable follow-up if the lost fast path proves to matter.

Closes #2891

https://claude.ai/code/session_01HbpiNTgsFoQmL6BWeeN8f7

---
_Generated by [Claude Code](https://claude.ai/code/session_01HbpiNTgsFoQmL6BWeeN8f7)_

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -121,11 +121,16 @@ pub struct EventStream {
 /// (node, input) in the attachment. The producer switches the output from the
 /// lossless daemon path to direct zenoh once all its required consumers acked.
 ///
-/// The thread acks *every* marker for the stream's whole lifetime: markers
-/// keep coming until the producer has this ack, so ack-route establishment
-/// races self-heal, and late producers (dynamic nodes, restarts) get their
-/// acks whenever they run their handshake. It exits when the data subscribers
-/// (the only senders) are dropped.
+/// The thread acks *every* marker for the stream's whole lifetime, so late
+/// producers (dynamic nodes, restarts) get their acks whenever they run their
+/// handshake. Note that the producer's markers are **not** unbounded: it stops
+/// marking an output at its startup grace boundary and pins any still-un-acked
+/// output to the daemon path for the rest of its run (dora-rs/dora#2891). So
+/// acking promptly is what preserves the fast path — an ack-route race that
+/// outlasts the producer's grace does not self-heal, and a dropped ack (the
+/// bounded-queue `try_send` fallback below) costs that output direct zenoh for
+/// the producer's whole run. The thread exits when the data subscribers (the
+/// only senders) are dropped.
 fn spawn_startup_acker(
     node_id: NodeId,
     ack_publishers: HashMap<DataId, zenoh::pubsub::Publisher<'static>>,
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -147,31 +147,32 @@ const LARGE_SEND_DIAG_LIMIT: u32 = 3;
 /// How often a starting node re-publishes its startup route-probe markers.
 ///
 /// See [`StartupHandshake`]. Markers stop per output as soon as that output's
-/// required acks have arrived (or the deadline froze it on the daemon path),
-/// so this rate only applies while the handshake is in flight.
+/// required acks have arrived (or the grace boundary froze it on the daemon
+/// path), so this rate only applies while the handshake is in flight.
 const ZENOH_STARTUP_MARKER_INTERVAL: Duration = Duration::from_millis(5);
 
-/// How long a producer keeps publishing markers for an un-acked output before
-/// freezing it on the reliable daemon path.
+/// The whole budget the startup handshake gets, measured from the daemon's
+/// "all nodes ready" barrier: `init` waits this long for the handshake, and
+/// whatever is still un-acked at the end is frozen on the daemon path for the
+/// run (see [`wait_for_grace`] for why the freeze is unconditional).
 ///
-/// Armed only once the daemon's "all nodes ready" barrier has released — i.e.
-/// once every static consumer has declared its subscribers and ack publishers
-/// — so a consumer that is merely slow to *start* (a Python node importing
-/// heavy libraries for a minute) cannot burn the deadline. From arming, this
-/// long with 5 ms markers and no ack means the route is genuinely broken, not
-/// slow. Nothing fails on the deadline: the output just keeps riding the
-/// lossless daemon path.
-const ZENOH_STARTUP_ACK_TIMEOUT: Duration = Duration::from_secs(10);
-
-/// How long `init` waits after the barrier for the handshake to complete
-/// before returning control to user code.
+/// The window starts at the barrier, not at spawn: by then every static
+/// consumer has declared its subscribers and ack publishers, so a consumer
+/// that is merely slow to *start* (a Python node importing heavy libraries for
+/// a minute) cannot burn it. The healthy case completes in one marker→ack
+/// round-trip (single-digit milliseconds); half a second of 5 ms markers with
+/// no ack means the route is broken, not slow. Nothing fails at the boundary —
+/// a frozen output just keeps riding the lossless daemon path, trading the
+/// fast path for ordering. This is the single knob for that trade: raising it
+/// gives slow-to-establish routes more chance at direct zenoh, at the cost of
+/// delaying every node whose routes are genuinely broken.
 ///
-/// The healthy case completes in one marker→ack round-trip (single-digit
-/// milliseconds), which makes switch-before-first-send the overwhelmingly
-/// common case — messages then never straddle the daemon→zenoh path switch,
-/// so cross-path reordering cannot occur. When a route is genuinely broken
-/// this costs half a second once and the affected outputs keep riding the
-/// daemon path (still upgrading later if acks arrive before the deadline).
+/// For a **dynamic or restarted** producer the barrier releases immediately —
+/// its consumers have long been running — so this window instead starts a few
+/// milliseconds after the zenoh session opens, while the peer links it needs
+/// may still be dialing. Such a producer is therefore the most likely to end
+/// up frozen on the daemon path; if that shows up as a measurable regression,
+/// this constant (not a late upgrade) is the thing to raise.
 const ZENOH_STARTUP_GRACE: Duration = Duration::from_millis(500);
 
 /// Poll interval for the post-barrier grace wait.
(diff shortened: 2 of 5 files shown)


phil-opp:
**Reviewed the diff — no issues found.**

The freeze correctly closes the mid-stream-upgrade hazard. `record` and `freeze` both take the `received` mutex, and `record` re-reads `frozen` under that lock before inserting or flipping `ready`, so an ack landing at the boundary either completes the set before the freeze or is ignored — never a half-applied upgrade. After `settle` returns, every `AckState` is either `ready` (freeze saw it and returned `false`) or `frozen`, and a frozen output can never flip back to `ready`, so each output's transport is fixed before user code's first send and FIFO order holds unconditionally. The marker thread skipping `ready || is_frozen()` outputs (unlocked `is_frozen`, where a stale `false` costs at most one extra marker) is safe, and `wait_for_grace` on an empty slice is vacuously done, matching the old early return. The added tests pin the freeze, late-ack-ignored, ready-spared, and early-return cases.

---
🤖 *This is a fully automated review by Claude (no human in the loop). It is not a maintainer approval — treat findings as suggestions to verify.*

---

PR #2895, opened 2026-07-30
fix(record-node): bound recorder crash-loss by record count and wall-clock time

phil-opp:
> 🤖 **Machine-generated PR.** Opened by Claude (Anthropic) during an automated code-review pass, and revised in response to review. Please review carefully before merging.

## Issue

`dora-record-node` wraps its output file in a `BufWriter` (via `RecordingWriter::new`) and calls `write_entry` in its event loop, but only ever flushes at the clean `Event::Stop` path through `finish()`. `RecordingWriter` does not flush on `Drop`, so on an abrupt **process death** that skips the graceful `Stop` — **SIGKILL, or SIGINT/Ctrl-C** — the buffered-but-unflushed records are lost.

(Scope note, from review: a **panic** unwinds — the workspace sets no `panic = "abort"` — so `BufWriter`'s `Drop` flushes best-effort and records are *not* lost that way. And `flush()` only reaches the OS page cache, so **power loss** still loses everything since the last writeback regardless; that would need `sync_data`, which this PR intentionally does not add. SIGKILL and SIGINT are the two cases this fixes.)

This contradicts the reader's crash-recovery model: `dora_recording::RecordingReader`'s torn-tail recovery assumes records are flushed as they are written, so only the *trailing* record can be torn. The node recorder was the odd one out — after a SIGKILL it lost many whole records, not just a torn tail.

## Fix

**Bound the loss window by `min(100 records, 1 s)`, enforced on the write path.** `FlushPolicy::after_write` flushes when *either* the count bound (`FLUSH_EVERY_N_RECORDS`) or the elapsed-time bound (`FLUSH_INTERVAL`) trips.

The earlier revision of this PR put the time bound only on the idle `_` arm, which fires only after a full second with **no** event — so on any stream faster than 1 Hz it never fired and the real bound was still count-only (a 2 Hz stream buffered ~99 records ≈ 50 s). Checking `last_flush.elapsed()` on the write path is what makes the time bound actually bind on an active stream. `on_idle` still flushes the tail of a stream that has gone quiet without closing.

**Timer cost:** the idle `recv_timeout` is only armed when records are actually buffered (`has_buffered()`); a caught-up recorder blocks on plain `recv()` and pays no per-event timer setup. (On a very high-rate stream that is continuously accumulating toward a flush, `recv_timeout` does allocate a short-lived timer per event — negligible beside the per-message Arrow-IPC encode, but noted.)

**Don't drop the post-`Stop` tail.** `Event::Stop` was a `break`, but the node API gives `Stop` priority over inputs already queued behind it and re-delivers those inputs on the following `recv()`s before closing (the `EventStream` docs note a manual `break` "is typically not needed"; pinned by `recv_drains_buffered_scheduler_inputs_after_stop`). Breaking discarded that tail of already-produced messages — exactly what a recorder must not do. `Stop` now falls through (flushing the buffered tail); the stream drains its queued inputs and returns `None`, ending the loop. This also future-proofs against the larger per-topic buffer in #3283.

## Tests

The flush *timing* lives in a small `FlushPolicy` state machine, unit-tested against a real `RecordingWriter<ProbeSink>` (an in-memory `Write` that records only flushed bytes):

- **`time_bound_flushes_on_the_write_path_below_the_count_bound`** — the core fix: with `last_flush` backdated past `FLUSH_INTERVAL`, a single write (well under the count bound) flushes. RED against the old count-only `after_write`.
- `count_bound_flushes_every_n_records` — no flush at `N-1`, flush on the `N`th (asserted on the internal counter, not only on buffered bytes, so it doesn't silently depend on the 8 KiB `BufWriter` default).
- `records_below_both_bounds_stay_buffered` / `idle_tick_flushes_records_the_count_bound_would_hold` — the lower edge and the quiet-stream tail flush.

**Coverage limitation (honest):** these cover the extracted `FlushPolicy` only. The loop wiring itself — `recv()`/`recv_timeout` selection and the `Stop` fall-through — can't be unit-tested here because `Event` is `#[non_exhaustive]` (synthetic events can't be constructed in this crate). The graceful record→replay path is exercised end-to-end by `contract_record_replay_reproduces_validated_pipeline` in `tests/example-smoke.rs`; a dedicated SIGKILL-mid-stream smoke test (kill the record node, assert `dora replay` recovers > 0 records) would be the ideal behavior test but is inherently racy/flaky, so it is deliberately not added here.

## Validation

- `cargo test -p dora-record-node` — 9 tests pass (Rust 1.95.0, the current MSRV).
- `cargo clippy -p dora-record-node --all-targets -- -D warnings` — clean.

The sibling `dora record --proxy` recorder in `dora-cli` has the same count-only gap; giving it the same write-path time bound is a reasonable follow-up, kept out of this PR to stay scoped to the node recorder.

https://claude.ai/code/session_01CCyEn8YqaYQYwh3RLjEZ7B


phil-opp:
## 🤖 Automated code-review pass — summary

This PR is one of a batch opened by an automated Claude (Anthropic) code-review pass over the repository. Each finding was verified against a freshly-fetched `origin/main`, checked against the currently-open review PRs to avoid duplicates, and lives on its own branch/PR. All are **draft** and **machine-generated** — please review carefully before merging.

### PRs opened, by category

**Correctness**
- **#2895** — `fix(record-node)`: flush the recorder's `BufWriter` every 100 records so an abrupt kill (SIGKILL/Ctrl-C) loses at most 100 records instead of a whole buffer, keeping the file within the reader's torn-tail recovery model.
- **#2896** — `fix(coordinator)`: make multi-daemon `stop_dataflow` best-effort — one unreachable daemon no longer aborts the stop and orphans nodes on the remaining healthy daemons (mirrors `rollback_spawned_daemons`).
- **#2897** — `fix(node)`: guard the JSON→Arrow decoders against a zero-column record batch (a caller-supplied empty-field schema would panic on `column(0)`); returns an error like the sibling IPC decoders.
- **#2898** — `fix(message)`: reserve the node id `dora` so the conflict with built-in input namespaces (`dora/timer/...`, `dora/logs`) is caught at validation time instead of surfacing as an opaque input-parse error.

- **#2899** — `docs(arrow-convert)`: add a runnable doctest for reading `ArrowData` back into Rust (`TryFrom<&ArrowData>`), the read counterpart to `IntoArrow`.
- **#2900** — `docs(coordinator-store)`: add a runnable doctest pinning `validate_param_limits`' accept/reject contract (size limits + null-byte rejection).

- **`logs --local` legacy `.txt` bypasses `--tail`/`--grep`/ordering** — real, but a correct fix requires synthesizing per-line timestamps (introducing its own `--since`/`--until` ambiguity) and the path is rarely reached since current logs are `.jsonl`.
- **`InputMapping::Logs` `Display` drops `node_filter` when `min_level` is `None`** — genuine round-trip data loss, but the grammar has no slot for "node without level"; the right fix (fail-loud / forbid at construction / extend grammar) is a maintainer judgment call.
- **`dora_read_data` returns `None` on non-`UInt8`/nullable input after consuming the FFI array** — real footgun, but a clean fix needs a breaking FFI signature/return-type change; too risky for an automated PR.
- **`dora record --proxy` output-path `PathBuf` round-trip is a no-op contradicting its comment** — cosmetic/dead-code only.
- **`node_binary::search` misses the `.exe` suffix in the `CARGO_TARGET_DIR` fallback on Windows** — niche (only when both exe-adjacent and PATH lookups miss), low confidence.
- **`dora replay --speed` accepts negative/NaN `f64` unvalidated** — no panic, effectively documented behavior, low value.
- **Python `Node.timestamp()` f64 microsecond can round off-by-one** — already documented as lossy, low value.

### Validation notes

Each PR was verified with `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, and tests for its affected crate(s). A full `cargo test --all` was attempted on the combined diff; it could not complete in this environment due to (a) a pre-existing `dora-ros2-bridge` C++ link failure unrelated to any of these diffs, and (b) the session's fixed disk allowance. Coverage was instead established per affected crate plus a repo-wide grep confirming no existing code/test/example uses a node named `dora`.

---

phil-opp:

No issues found. `if msg_count.is_multiple_of(100) { writer.flush()?; }` after `write_entry` / `msg_count += 1` flushes every 100 records, mirroring the `dora-cli` proxy recorder, and keeps the file within the reader's torn-tail recovery model. The clean `Event::Stop` path still flushes any remaining records via `finish()`. `is_multiple_of` is stable well below the pinned 1.97.1 toolchain.

No test was added; this replicates the already-tested proxy-recorder flush cadence exactly and a crash-recovery test is hard to author for a node binary, so that's acceptable here rather than a blocker.

---

phil-opp:
Throughput and deadlock are fine — `RecordingWriter::flush` is a plain `BufWriter::flush`, one `write(2)`, no `fsync`, so every 100 records is negligible.

The durability window is bounded by *count*, not time. A 1Hz topic with payloads under the 8 KiB `BufWriter` auto-flush threshold sits unflushed for roughly 100 seconds, so a SIGKILL loses ~100s of recording. "Loses at most 100 records" is literally true, but crash-recovery fixes usually want a wall-clock bound. The sibling CLI recorder (`binaries/cli/src/command/record.rs:487`) has the identical gap, so this is consistency-preserving rather than a new hazard — but neither recorder is actually time-safe.

Also no test: nothing asserts the node recorder flushes, and "mirrors the already-tested proxy recorder" is not a substitute.

phil-opp:
Good points — addressed both in 50c1e8c.

**Wall-clock bound.** You're right that count-only isn't time-safe. The event loop now polls with `events.recv_timeout(FLUSH_INTERVAL)` (1 s) and flushes any buffered records on each idle tick, so the loss window is bounded by wall-clock time as well as record count. `recv_timeout` returns `None` only on stream close and yields `Event::Error` on a timeout, so the idle tick falls through the loop's existing non-input arm — no change to the close/stop paths.

**Test.** `Event` is `#[non_exhaustive]`, so I can't construct synthetic `Event::Input`s in this crate to drive the loop end-to-end. Instead I extracted the flush timing into a small `FlushPolicy` and tested it directly against a real `RecordingWriter` over an in-memory sink that records only *flushed* bytes — so the assertions are about actual on-"disk" durability, not a mock of the cadence:
- records below the count bound stay buffered (the gap a SIGKILL would lose),
- an idle tick flushes them (and a second idle tick with nothing buffered is a no-op),
- the count bound flushes on the Nth record.

The event loop is now a thin wiring of `after_write` / `on_idle` around that tested policy.

On the sibling `dora record --proxy` recorder having the same count-only gap: agreed. I've left it out to keep this PR scoped to the node recorder, but happy to give it the same time bound in a follow-up.

---

phil-opp:

Re-reviewed after the wall-clock-bound rework. The `recv_timeout` wiring is sound: `recv_async_timeout` in `apis/rust/node/src/event_stream/mod.rs` returns `Some(Event::Error(..))` on timeout and `None` only on stream close, so an idle tick correctly falls through the `_ => flush_policy.on_idle(..)` arm and the loop exits only on a genuine close (where `finish()` still flushes). `on_idle` no-ops when nothing is buffered, and the Stop path is unchanged.

The `FlushPolicy` tests assert against a real `RecordingWriter` over a sink that records only *flushed* bytes, so they verify actual durability rather than a mocked cadence — that closes the "no test / time-unbounded flush" gap from the earlier pass. One residual note: since `Event` is `#[non_exhaustive]`, the loop wiring (`recv_timeout` → `Event::Error` → `_` arm) itself isn't covered — only the extracted `FlushPolicy` is — so a future change to that arm mapping wouldn't be caught. Correct today; not blocking. No new issues found.

---

phil-opp (commented):

Re-reviewed after the 2026-08-18 rebase/conflict-resolution merge (`28c2d17`). The reviewable content in `binaries/record-node/src/main.rs` is unchanged from the last review of the wall-clock-bound rework: the `FlushPolicy` (count bound via `after_write`, time bound via `on_idle`) and the `recv_timeout(FLUSH_INTERVAL)` loop wiring are intact, and the `Event::Stop` path still flushes via `finish()`. No new issues found from the rebase.

---

phil-opp:

One observation on the wall-clock flush bound, prompted by the newly added `recv_timeout` idle-flush. `recv_timeout` constructs a fresh `Delay` on each call (`apis/rust/node/src/event_stream/mod.rs:1172`), so its 1 s window measures the gap between consecutive events, not elapsed time since the last flush. That means the idle tick only fires when the stream actually goes quiet for ≥ `FLUSH_INTERVAL`.

For a stream that stays continuously active with inter-event gaps under 1 s — e.g. a steady 2 Hz sensor with sub-8 KiB payloads that never hits the `BufWriter` size threshold — `on_idle` never runs, and flushing falls back to the pure count bound (up to `FLUSH_EVERY_N_RECORDS` records, ~50 s of buffered data at 2 Hz). So for that case the loss window is still count-bounded rather than time-bounded, which is narrower than the "bounds loss by wall-clock time" framing in the description and the `FLUSH_INTERVAL` doc comment. This is exactly the low-rate class the earlier review round pointed at, so the idle-flush closes part but not all of that gap.

If a true wall-clock bound is desired, tracking the time of the last flush and flushing after a write once `FLUSH_INTERVAL` has elapsed (in addition to the idle tick) would cover the always-busy case; otherwise it may be worth softening the doc comment to say the time bound applies while the stream is idle. Not a crash or corruption issue — everything else looks solid: the `Stop`/close paths still flush via `finish()`, and the `FlushPolicy` tests assert against actually-flushed bytes.

---

phil-opp:

The direction is right and the base facts check out — `RecordingWriter` wraps a `BufWriter`, only `finish()` flushes, and the sibling count bound at `binaries/cli/src/command/record.rs:583-585` is real.

**The advertised time bound does not bind.** `FlushPolicy::on_idle` is only reachable from the `_` arm, which runs only when no event arrived for a full second. On any stream faster than 1 Hz the timeout never fires, so the only bound is still the count bound: a 2 Hz stream buffers 99 records ≈ **50 seconds**; a 1.1 Hz stream ≈ 90 s. The body's "this bounds loss by wall-clock time, not just record count" is false in exactly the regime it claims to fix.

Put the deadline on the write path, where it actually binds:

```rust
struct FlushPolicy { records_since_flush: u64, last_flush: std::time::Instant }

fn after_write<W: Write>(&mut self, w: &mut RecordingWriter<W>) -> eyre::Result<()> {
    self.records_since_flush += 1;
    if self.records_since_flush >= FLUSH_EVERY_N_RECORDS
        || self.last_flush.elapsed() >= FLUSH_INTERVAL
    { w.flush()?; self.records_since_flush = 0; self.last_flush = std::time::Instant::now(); }
    Ok(())
}
```

Keep `on_idle` for the fully-silent case; then `min(100 records, 1 s)` is genuinely the bound.

**Two of the four listed failure modes are wrong.** A panic unwinds (the workspace sets no `panic = "abort"`), so `BufWriter::Drop` flushes best-effort — records are *not* lost. And `flush()` reaches the page cache, not the platter, so power loss still loses everything since the last writeback; that needs `sync_data()`, which this PR does not add and arguably should not. SIGKILL and SIGINT are the two genuine cases.

**None of the three tests would fail without the change.** All three construct `FlushPolicy` and call `after_write`/`on_idle` — a type this PR introduces. `records_below_the_count_bound_stay_buffered` actually pins the *un-fixed* behavior. Nothing covers the two lines that change behavior (`recv()` → `recv_timeout(FLUSH_INTERVAL)`, and the `_ => on_idle`). There is an end-to-end seam at `tests/example-smoke.rs:1100`: a test that SIGKILLs the record node mid-stream and asserts `dora replay` recovers more than zero records would fail on `main`. Failing that, please say plainly in the body that the unit tests cover the extracted policy only.

Two smaller notes: `test_writer()` hard-codes `version: 1` instead of `FORMAT_VERSION`, and the count test's `flushed_len() == 0` silently depends on 99 × 30-byte entries fitting under `BufWriter`'s 8 KiB default. Also, `recv_timeout` allocates and drops a `Delay` per event where `recv()` did neither — at 1 kHz that is 1000 timer setups/sec on a throughput-critical node; worth a sentence, or a larger `FLUSH_INTERVAL`.

Finally, `Event::Stop(_) => break` at `binaries/record-node/src/main.rs:165` is left as-is, and it is the bigger hole. The node API deliberately re-delivers inputs held behind the prioritized `Stop` (`event_stream/mod.rs:877-907`, pinned by `recv_drains_buffered_scheduler_inputs_after_stop`), and this `break` opts out. Deleting the arm so `Stop` falls through to `_` would make the PR's own claim hold for the graceful path too. Heads-up that #3283 raises the per-topic cap to 1000, which makes that discard 100× larger — so whichever of the two lands, the `break` should go with it.

phil-opp:
Thanks — this caught a real hole. All addressed in `ca7145d`.

**Time bound didn't bind — fixed.** You're right that `on_idle` only fires after a full second of silence, so anything faster than 1 Hz was count-only. Moved the deadline onto the write path exactly as you sketched: `after_write` now flushes when `records_since_flush >= FLUSH_EVERY_N_RECORDS || last_flush.elapsed() >= FLUSH_INTERVAL`, with `flush()` resetting both. `on_idle` stays for the fully-silent tail. `min(100 records, 1 s)` now genuinely binds.

**Body's failure modes corrected.** Dropped "panic" and "power loss" — a panic unwinds (no `panic = "abort"`) so `BufWriter::Drop` flushes, and `flush()` only reaches the page cache so power loss needs `sync_data` (out of scope). Body now says SIGKILL/SIGINT only, with a scope note.

**Tests.** Added `time_bound_flushes_on_the_write_path_below_the_count_bound` — backdates `last_flush`, writes one record under the count bound, asserts the flush reaches the sink. RED against the old count-only `after_write`. I've also stated plainly in the body that the unit tests cover the extracted `FlushPolicy` only: the loop wiring can't be unit-tested here since `Event` is `#[non_exhaustive]`. I did not add the SIGKILL-mid-stream smoke test — it's inherently racy and I can't validate a networked smoke test reliably in this environment — so the body calls that out as the coverage gap rather than pretending otherwise.

**`Event::Stop => break` removed** — the bigger hole, agreed. It now falls through so the inputs the scheduler re-delivers after the prioritized `Stop` get recorded, then `None` ends the loop (the `EventStream` docs even say the manual `break` "is typically not needed"). Flagged #3283 in the commit since the larger cap makes the old discard worse.

**Smaller notes:** `test_writer` now uses `FORMAT_VERSION`; the count test asserts on `records_since_flush` (not just buffered bytes) and documents the 99×~30 B < 8 KiB assumption; and `recv_timeout` is now only armed when records are buffered (`has_buffered()`), so a caught-up recorder blocks on `recv()` with no per-event timer — I noted the residual cost on a continuously-accumulating high-rate stream in the body.

---

phil-opp:

Re-reviewed after `ca7145db`. The prior review's main concern is resolved: the wall-clock bound now binds on the write path — `FlushPolicy::after_write` flushes when `records_since_flush >= FLUSH_EVERY_N_RECORDS || last_flush.elapsed() >= FLUSH_INTERVAL` — so a stream faster than 1 Hz is genuinely `min(100 records, 1 s)`-bounded rather than count-only, and `on_idle` still handles the fully-quiet tail. `Event::Stop` now falls through to the flush arm instead of `break`ing, so the inputs the scheduler re-delivers after the prioritized `Stop` are recorded before the stream returns `None`. The new `time_bound_flushes_on_the_write_path_below_the_count_bound` test is a real RED against the old count-only `after_write` (it backdates `last_flush` and asserts a single sub-count-bound write reaches the sink), and the body no longer overstates the failure modes or test coverage. No new issues found.



---

PR #2897, opened 2026-07-30
fix(node): guard JSON-to-Arrow decoders against a zero-column record batch

phil-opp:

## Issue

In `apis/rust/node/src/daemon_connection/json_to_arrow.rs`, both `read_from_json_with_schema` and `read_json_value_as_arrow` end with `batch.column(0).to_data()` with no column-count check.

`read_json_value_as_arrow` takes a **caller-supplied** `Arc<Schema>` (used e.g. from `node_integration_testing.rs`). An empty-field schema produces a record batch with 0 columns, and `batch.column(0)` then indexes out of bounds → **panic**.

The sibling IPC decoders in this crate (`decode_arrow_ipc`, `ipc_encode::decode_one_batch`) already guard the column count before indexing; these two JSON paths did not.

## Fix

Add a `num_columns() == 0` check before `column(0)` in both functions, returning a descriptive error instead of panicking — consistent with the IPC decoders:

```rust
if batch.num_columns() == 0 {
    eyre::bail!("JSON record batch has no columns");
}
Ok(batch.column(0).to_data())
```

A regression test constructs an empty-field schema and asserts the function returns `Err` rather than panicking.

## Validation

- `cargo test -p dora-node-api --lib json_to_arrow` — new test passes

Severity is low (the reachable path is cold — interactive/test schema input), but it removes a real panic and aligns the JSON decoders with the crate's existing IPC guards.

---

### DIFF
--- a/apis/rust/node/src/daemon_connection/json_to_arrow.rs
+++ b/apis/rust/node/src/daemon_connection/json_to_arrow.rs
@@ -31,6 +31,9 @@ fn read_from_json_with_schema(
         .context("no record batch in JSON")?
         .context("failed to read record batch")?;
 
+    if batch.num_columns() == 0 {
+        eyre::bail!("JSON record batch has no columns");
+    }
     Ok(batch.column(0).to_data())
 }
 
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The `if batch.num_columns() == 0 { eyre::bail!(...) }` guard before `batch.column(0)` is correct and mirrors the crate's existing IPC decoders, and it's applied to both decode functions. The added test is genuine: `decoder.serialize(&[{}])` decodes one row against a 0-field schema, so `flush()` returns `Some(batch)` with 0 columns, which reaches the new guard (rather than the earlier "no record batch" path), and the test asserts `is_err()` instead of a panic.

---

phil-opp:
needs rebase

phil-opp:
Rebased onto the latest `main` (21568f1) and pushed. The conflict was only in the `#[cfg(test)] mod tests` block — `main` had added the plain-text/`json_quote` round-trip tests while this branch added `empty_field_schema_errors_instead_of_panicking`; I kept both import sets and all five tests. The `num_columns() == 0` guards in the non-test code merged cleanly. Verified locally: `cargo test -p dora-node-api --lib json_to_arrow` → 5/5 pass, fmt clean.

---

PR #2898, opened 2026-07-30
fix(message): reserve the node id `dora` to match built-in input parsing

phil-opp:

## Issue

`validate_node_id` (`libraries/message/src/id.rs`) accepts `dora` as a node id (it is plain alphanumeric). But `InputMapping::from_str` (`libraries/message/src/config.rs`) **unconditionally** treats any mapping whose source segment is `dora` as a built-in namespace (`dora/timer/...`, `dora/logs`).

So a user node literally named `dora` is silently unusable as an input source. For a consumer subscribing to `dora/some_output`:

- `source == "dora"`, so the built-in `"dora"` arm is taken;
- `some_output` is not `logs` and has no further `/`, so parsing falls through to `Err("dora input has invalid format")`.

Net effect: a validly-named node `dora` fails at input-mapping parse time with an opaque error, and nothing rejects the id up front.

## Fix

Reserve `dora` in `validate_node_id` so the conflict is caught at node definition with a clear message, rather than surfacing later as an opaque input-parse error:

```rust
if id == "dora" {
    return Err(InvalidId(
        "identifier 'dora' is reserved for built-in inputs \
         (dora/timer/..., dora/logs) and cannot be used as a node id".into(),
    ));
}
```

Only the exact id `dora` is rejected — names that merely contain or extend it (`dora-node`, `my-dora`, `Dora`) remain valid.

## Validation

- Added `node_id_rejects_reserved_dora` unit test.
- `cargo test -p dora-message` (263 + 141 …) and `-p dora-core` (all) pass.
- Repo-wide grep confirms **no** existing code, test, or example dataflow uses a node literally named `dora`, so nothing regresses.
- `cargo clippy` / `cargo fmt --all -- --check` — clean.

Note: this is a (small) behavior change — it newly rejects the name `dora` at validation time. Since such a node was already non-functional as an input source, rejecting it early is strictly clearer. Flagging it explicitly for maintainer judgment.

---


phil-opp:

No issues found. `validate_node_id` now rejects exactly `id == "dora"`, which correctly closes the gap where a node named `dora` produced outputs that `InputMapping::from_str` would misparse as the built-in `dora/...` namespace. The exact-match check is right — `dora-node`, `my-dora`, and `Dora` remain valid — and the added test guards both the reject case and those three near-miss accept cases, so it would catch an over- or under-broad implementation. The narrow tightening is justified since a node named `dora` was already effectively unusable.

---

phil-opp:

The branch was rebased/force-pushed since the last review, so I re-checked the current diff. It still looks correct and safe:

- Only the exact id `dora` is rejected (`if id == "dora"` in `libraries/message/src/id.rs`), which matches the unconditional built-in-namespace handling in `InputMapping::from_str` — so `dora-node`, `my-dora`, and `Dora` remain valid, as the test asserts.
- I confirmed no bundled example uses a node literally named `dora` (the only `dora` occurrences in `*.yml` are `dora-sender`/`dora-receiver` in `examples/ros2-comparison/dataflow.yml` plus unrelated CI config), so nothing existing regresses.

No new issues found. I couldn't tie the earlier trunk merge-queue failure to anything in this diff — it looks environmental rather than a code defect, so a re-run/rebase is likely all that's needed.

---

PR #2900, opened 2026-07-30
docs(coordinator-store): add a compile-checked doctest for validate_param_limits

phil-opp:

## Issue

`validate_param_limits` (`libraries/coordinator-store/src/lib.rs`) is the single source of truth both store backends (in-memory + redb) rely on for param key/value size limits and the subtle null-byte rejection rule (redb uses `\0` as an internal composite-key separator). It had only prose docs; its accept/reject contract was pinned only by `#[cfg(test)]` backend tests, not a compile-checked example on the exported function itself.

## Fix

Add a runnable `# Examples` doctest covering the function's contract:

- an ordinary key/value is accepted;
- a key exactly at `MAX_PARAM_KEY_BYTES` is accepted, one byte over is rejected;
- a key containing a null byte is rejected.

Pure documentation addition — no code behavior changes.

## Validation

- `cargo test -p dora-coordinator-store --doc` — the new doctest passes.

---

### DIFF
--- a/libraries/coordinator-store/src/lib.rs
+++ b/libraries/coordinator-store/src/lib.rs
@@ -18,6 +18,23 @@ pub const MAX_PARAM_VALUE_BYTES: usize = 65_536;
 /// internal composite-key separator (see `redb_store::KEY_SEPARATOR`); without
 /// this check `put_node_param` with a `\0`-containing key would succeed on the
 /// in-memory backend but fail on redb.
+///
+/// # Examples
+///
+/// ```
+/// use dora_coordinator_store::{validate_param_limits, MAX_PARAM_KEY_BYTES};
+///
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The doctest compiles against the real public API (`MAX_PARAM_KEY_BYTES` and `validate_param_limits` are both `pub` at the crate root and unconditionally compiled), and its assertions match the implementation exactly: the check is `key.len() > MAX_PARAM_KEY_BYTES`, so a key of length `MAX_PARAM_KEY_BYTES` is accepted and `+1` is rejected, and the `"a\0b"` case is rejected by the null-byte branch. It exercises both accept and reject paths plus the exact boundary, so it can genuinely fail if the contract regresses.

---

PR #2901, opened 2026-07-31
fix(core): expand documented module param args syntax

SunSunSun689:
Fix `dora expand` module parameter substitution for the documented `$PARAM_<UPPERCASE_KEY>` syntax in module `args`.
Previously, only `${_param.name}` was substituted, so examples like  `$PARAM_SPEED` stayed unresolved after expansion. This keeps the existing syntax and adds the documented one, with longest-name-first replacement to avoid prefix conflicts such as `speed` / `speed_limit`.


phil-opp:

No issues found. The `$PARAM_<KEY>` substitution runs after the existing `${_param.name}` pass, and patterns are sorted longest-first (`sort_by_key(|(p,_)| Reverse(p.len()))`) so `$PARAM_SPEED_LIMIT` is replaced before `$PARAM_SPEED`, correctly avoiding the prefix collision. Tests are meaningful: one asserts `$PARAM_SPEED`/`$PARAM_MODE` resolve correctly, and another specifically exercises the `speed`/`speed_limit` collision and would fail without the longest-first ordering.

One theoretical edge case (not a real bug): keys are uppercased, so two params differing only by case (e.g. `speed` vs `Speed`) would both map to `$PARAM_SPEED` and one value would win nondeterministically — an implausible input.

---

phil-opp:
🤖 **Automated review by Claude** — this comment is fully automated and has not been vetted by a human.

I took a fresh look at the diff and found one correctness issue I don't think has been raised yet:

The new `$PARAM_<KEY>` substitution in `substitute_params_in_str` (`expand.rs`) uses an unanchored `String::replace`, so it will match a prefix that isn't a parameter boundary. The longest-first sort (`Reverse(pattern.len())`) only disambiguates among the **declared** params — it does nothing when an arg token is a *superstring* of a param pattern but isn't itself a declared param.

Example: with a single param `speed: "2.0"` and `args: --flag $PARAM_SPEED_LIMIT`, the result is `--flag 2.0_LIMIT` rather than being left intact. A shell treats `$PARAM_SPEED_LIMIT` as one identifier and would not partially expand it. Since the expanded args then go through `shlex::split` and are passed straight to the process (`binaries/daemon/src/spawn/command.rs`), the mangled token reaches the node silently — a hard-to-diagnose failure.

The delimited `${_param.name}` form isn't affected because `}` terminates the match. Consider only replacing `$PARAM_<KEY>` when the following character isn't `[A-Za-z0-9_]` (matching shell identifier semantics), and adding a test for the "arg contains a superstring of a param name" case — the two added tests don't cover it (the longest-first test uses two *declared* params, which is the case the sort already handles).

---

phil-opp (commented):

Re-checked after the 2026-08-18 rebase merge. The substitution logic is unchanged by the merge, so the correctness issue flagged in my [earlier review](https://github.com/dora-rs/dora/pull/2901#issuecomment-5197097125) is still present and worth re-surfacing now that the PR is active again.

`substitute_params_in_str` (`libraries/core/src/descriptor/expand.rs:842-849`) substitutes `$PARAM_<KEY>` with an unanchored `String::replace`, so it matches a param name that is a prefix of a longer, undeclared token. The longest-first sort only disambiguates among *declared* params; it does nothing when an arg token is a superstring of a param pattern.

Example: with a single param `speed: "2.0"` and `args: --flag $PARAM_SPEED_LIMIT`, the result is `--flag 2.0_LIMIT` rather than being left intact. A shell treats `$PARAM_SPEED_LIMIT` as one identifier and would not partially expand it. The expanded args go through `shlex::split` and are passed straight to the process, so the mangled token reaches the node silently.

The delimited `${_param.name}` form is unaffected (`}` terminates the match). Fix by only replacing `$PARAM_<KEY>` when the following character isn't `[A-Za-z0-9_]`, and add a test for the "arg contains a superstring of a declared param name" case — the two added tests both use *declared* params, which is the case the sort already handles.

---

phil-opp:
Merged `main`, which conflicted semantically rather than textually: `main` had rewritten `substitute_params_in_str` as a single-pass scanner that never re-scans a substituted value (#3101), while this branch kept the old naive `String::replace` loop and added a second pass for the `$PARAM_<KEY>` form.

I resolved it by folding `$PARAM_<KEY>` into `main`'s scanner, which also closes the standing review comment: the loop now finds the next `$`, dispatches on `${_param.` vs `$PARAM_` vs a bare `$`, and for the undelimited form takes the key as the maximal `[A-Za-z0-9_]` run. So `$PARAM_SPEED_LIMIT` is one token and is left intact when only `speed` is declared, instead of becoming `2.0_LIMIT`.

Verified rather than assumed:

- Compiled the old branch implementation alongside the merged one and ran your exact repro. `--flag $PARAM_SPEED_LIMIT` → old `--flag 2.0_LIMIT`, new `--flag $PARAM_SPEED_LIMIT`; `$PARAM_SPEEDX` → old `2.0X`, new unchanged; `--flag $PARAM_SPEED --x` unchanged in both. So the new boundary test is genuinely RED-before/GREEN-after, not a tautology.
- Differential-fuzzed the new scanner against `main`'s over ~1.4M inputs (400k random + 579k exhaustive atom-sequences, alphabet biased toward `$ { } _ . P A R M` plus 2/3/4-byte chars, a combining mark, ZWJ, NBSP, and the Turkish dotted-I pair). Zero panics. Every single divergence contains the literal `$PARAM_` — on 400k `$PARAM_`-free inputs there were zero mismatches, so behavior for the existing `${_param.KEY}` syntax is unchanged.
- Termination: every branch consumes ≥1 byte before reassigning `rest`, including a bare `$` at end-of-string, `$$`, and an unterminated `${_param.`.
- Byte-index safety: every index feeding a slice is a char boundary by construction — `str::find` returns the start of a match, `strip_prefix` is boundary-aligned, and `}`/`$` are single-byte, so no multi-byte char can be split.
- The uppercase lookup can't silently drop a param: `substitute_params_in_str` has exactly one production caller chain, and the case-collision validation is unconditional and lexically precedes it in the same function with no early return between. `ModuleHeader` has no `params` field, so there's no alternative source that skips the check. The validator's identifier set and the scanner's terminator are exact complements, and all three uppercase sites use the identical `key.to_uppercase()`.

The merge also fixed a second latent bug in the branch that wasn't flagged: the old code re-expanded substituted values transitively — with `{a: "$PARAM_B", b: "x"}`, `$PARAM_A` yielded `x` rather than `$PARAM_B`. Pinned by a new test.

Two follow-ups pushed: `expand_params_in_env_style_args_replaces_longest_names_first` named a sort that no longer exists, so it's renamed (kept, because it covers the both-keys-declared half of the overlap problem where the new boundary test covers the only-shorter-declared half); and both doc comments described only `${_param.name}`, which is the form documented *nowhere* — `$PARAM_<UPPERCASE_KEY>` is what `docs/modules.md` and `docs/yaml-spec.md` teach, and is the whole point of this PR.

One thing for you: the PR body still says "with longest-name-first replacement to avoid prefix conflicts", which is the mechanism the merge removed. Worth rewording before merge, since the body becomes the squash commit message.

PR #2866, opened 2026-07-28
fix(python): use wrapping_add(1) in seqlock_begin_write for overflow consistency

SaitejaKommi:
## Summary

Use `wrapping_add(1)` in `seqlock_begin_write` to provide consistent overflow handling for the seqlock generation counter.

Fixes #2865

## Root Cause

In `apis/python/node/src/lib.rs`, `seqlock_begin_write` incremented the generation counter using `pre_write_gen + 1`, while `seqlock_end_write` already used `pre_write_gen.wrapping_add(2)`.

Using normal integer addition can panic on `u64` overflow in debug builds or when overflow checks are enabled, resulting in inconsistent overflow behavior between the begin and end write paths.

## Solution

- Replaced `pre_write_gen + 1` with `pre_write_gen.wrapping_add(1)` in `seqlock_begin_write`.
- Align the overflow behavior with `seqlock_end_write`, ensuring the generation counter wraps consistently without panicking.

## Testing

- ✅ `cargo fmt --all -- --check`
- ✅ `cargo check -p dora-node-api-python`

## Scope

This PR is intentionally limited to `apis/python/node/src/lib.rs` and changes a single arithmetic operation to ensure consistent overflow handling. No functional APIs or runtime behavior were otherwise modified.

Issue #2865, opened 2026-07-28
Use wrapping_add(1) in seqlock_begin_write for overflow consistency


SaitejaKommi:
## Summary

`seqlock_begin_write` increments the generation counter using normal integer addition, while `seqlock_end_write` already uses `wrapping_add`.

## Location

`apis/python/node/src/lib.rs`

## Current behavior

The write-begin path increments the generation counter using:

```rust
pre_write_gen + 1
```

The write-end path already uses:

```rust
wrapping_add(2)
```

This creates inconsistent overflow handling.

In debug builds, normal integer addition can panic on overflow, whereas `wrapping_add` preserves the intended wrapping semantics of the sequence counter.

## Suggested fix

Replace

```rust
pre_write_gen + 1
```

with

```rust
pre_write_gen.wrapping_add(1)
```

to make both write paths use consistent overflow behavior.

PR #2619, opened 2026-07-13
feat(memory-pool): GPU-to-GPU adaptive transport with P2P auto-selection and host-staging fallback

tang-canran:
## Summary

Extend the HeteroPool memory pool with device-aware GPU-to-GPU tensor transfer. The sender automatically detects the source and destination device topology via CUDA Runtime API, and selects the optimal transport path without any user configuration or hardcoded device logic.

### 1. Same-GPU IPC Handle (cuda_inner)
- `cudaIpcGetMemHandle` / `cudaIpcOpenMemHandle` for zero-copy cross-process tensor sharing on the same GPU.
- Pool reuse + DORADMA header fast-read eliminate per-frame `cudaMalloc` and daemon queries.

### 2. Cross-GPU P2P via cudaMemcpyPeer (cuda2cuda)
- `cudaDeviceCanAccessPeer` auto-detection → `cudaMemcpyPeer` when P2P is available.
- Transparently uses NVLink (if hardware present) or PCIe P2P.
- **48,362 MB/s** on A100 (PCIe P2P, 84% of same-GPU IPC bandwidth).

### 3. Cross-GPU Host-Staging Fallback (no-P2P)
- When P2P is unavailable (e.g., RTX 5090), cross-device IPC handle import is fundamentally unsupported — `cudaIpcOpenMemHandle` fails on a different device (error 217). The fallback uses a **CPU page-locked staging buffer** (`cudaHostAlloc`): source GPU → DtoH → pinned host buffer → HtoD → destination GPU.
- The staging path incurs two PCIe traversals (read + write), yielding ~40% of same-GPU IPC bandwidth.
- **20,307 MB/s** on RTX 5090 (CPU page-locked staging).

### 4. P2P Auto-Enablement
- `_ensure_p2p_enabled()` called at pool registration and IPC import time.
- Iterates all GPU pairs, calls `cudaDeviceEnablePeerAccess` for eligible pairs.
- Idempotent — subsequent calls are no-ops.

## Transport Auto-Selection Logic

```
1. Same GPU → CUDA IPC Handle (zero-copy)
2. Same CPU → /dev/shm shared memory (existing)
3. CPU↔GPU → Pinned DMA or pageable cudaMemcpy (existing, size-adaptive)
4. Different GPU, P2P OK → cudaMemcpyPeer (NVLink > PCIe P2P)
5. Different GPU, no P2P → CPU page-locked staging (DtoH→cudaHostAlloc buffer→HtoD)
[...]

PR #2902, opened 2026-07-31
fix(node): use wrapping_add for all inline seqlock generation increments

### LINKED ISSUE #2890: seqlock overflow-consistency fix (#2866) is incomplete: two inline CPU-copy end-write paths still use non-wrapping `old_gen + 1`

## Summary

```rust
```

## Location

**Fast path (~line 1800):**
```rust
unsafe {
}
```

**Slow path (~line 1945):**
```rust
unsafe {
}
```

## Why this matters

## Suggested fix

## How this was found

phil-opp:
Fixes #2890.

## Problem

PR #2866 (for issue #2865) set out to make the memory-pool seqlock generation counter in `apis/python/node/src/lib.rs` use *uniform* wrapping semantics, so that no increment of the counter can panic on overflow in debug builds while the others wrap silently. That PR only changed the `seqlock_begin_write` helper.

Four inline increments of the very same counter (`header[96]`, `gen_ptr`) were left using the non-wrapping `old_gen + 1`:

- `apis/python/node/src/lib.rs:1432` — begin-write of the top-level `write_memory_pool` path
- `apis/python/node/src/lib.rs:1511` — end-write of that same path
- `apis/python/node/src/lib.rs:1802` — inlined "end write (infallible CPU copy)" block (fast path)
- `apis/python/node/src/lib.rs:1947` — inlined "end write (infallible CPU copy)" block (slow path)

These were the only remaining non-wrapping increments of a counter whose helpers already use `wrapping_add(1)` / `wrapping_add(2)`. The issue named the two CPU-copy blocks (1802, 1947); the two begin/end increments at 1432/1511 are the same class, so this fixes all four for full consistency — which is what #2865 originally asked for.

## Fix

Replace all four `old_gen + 1` with `old_gen.wrapping_add(1)`, matching the semantics of the `seqlock_begin_write` / `seqlock_end_write` helpers.

## Severity

Low. `gen` is a `u64` advancing by 2 per write, so overflow requires ~2^63 writes and is not reachable in practice. This is a code-consistency / robustness cleanup that removes a debug-build panic divergence, not a live runtime bug.

## Verification

`cargo check -p dora-node-api-python` passes cleanly (only a pre-existing, unrelated `downcast_into` deprecation warning remains).

---
_Generated by [Claude Code](https://claude.ai/code/session_01QXQ8MhA2GNSbyzqEBam6Xx)_

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -1429,7 +1429,7 @@ impl Node {
         unsafe {
             let gen_ptr = shmem_ptr.add(96) as *mut u64;
             let old_gen = std::ptr::read_volatile(gen_ptr);
-            std::ptr::write_volatile(gen_ptr, old_gen + 1);
+            std::ptr::write_volatile(gen_ptr, old_gen.wrapping_add(1));
             std::sync::atomic::fence(std::sync::atomic::Ordering::Release);
         }
 
@@ -1508,7 +1508,7 @@ impl Node {
         unsafe {
             let gen_ptr = shmem_ptr.add(96) as *mut u64;
             let old_gen = std::ptr::read_volatile(gen_ptr);
-            std::ptr::write_volatile(gen_ptr, old_gen + 1);
+            std::ptr::write_volatile(gen_ptr, old_gen.wrapping_add(1));
             std::sync::atomic::fence(std::sync::atomic::Ordering::Release);
         }
 
@@ -1799,7 +1799,7 @@ impl Node {
                             // Seqlock: end write (infallible CPU copy, always advances)
                             unsafe {
                                 let old_gen = std::ptr::read_volatile(gen_ptr);
-                                std::ptr::write_volatile(gen_ptr, old_gen + 1);
+                                std::ptr::write_volatile(gen_ptr, old_gen.wrapping_add(1));
                                 std::sync::atomic::fence(std::sync::atomic::Ordering::Release);
                             }
                         }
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The four inline `old_gen.wrapping_add(1)` changes correctly align the seqlock generation increments with the wrapping semantics already used by the `seqlock_begin_write`/`seqlock_end_write` helpers on the same counter (`header[96]`), so the previously inconsistent debug-build overflow panics can no longer occur at only some increment sites. Readers only compare generation parity/equality, so wrapping is semantically safe. No behavior change to test beyond the existing seqlock protocol coverage.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop; treat it accordingly.

Re-checking this diff against current `main` (now at `#2619`, `e99ba21`), I found it's stale in a way that undercuts the change's stated goal:

1. **Two of the four hunks no longer apply.** Hunks 3 and 4 target the inline `// Seqlock: end write (infallible CPU copy, always advances)` blocks, but `#2619` refactored both end-write paths into the `seqlock_end` / `seqlock_begin_if_even` helpers, so those inline blocks no longer exist on `main`. The branch will conflict on rebase. (Hunks 1 and 2 are fine — the begin-write `old_gen + 1` sites are now at `apis/python/node/src/lib.rs:2048` and `:2317`.)

2. **A new non-wrapping increment of the same counter is left unfixed.** `#2619` added `seqlock_begin_if_even` (`apis/python/node/src/lib.rs:1052`), which does `std::ptr::write_volatile(gen_ptr, cur + 1)` on the same `header[96]` generation counter — a plain, non-wrapping increment. The PR's premise is that these were "the only remaining non-wrapping increments"; after `#2619` that's no longer true. To actually reach uniform wrapping semantics, this site should also become `cur.wrapping_add(1)`.

Net: this needs a rebase onto `#2619`, and the `seqlock_begin_if_even` site should be converted as well so the "uniform wrapping" goal actually holds. The two surviving inline edits (now at :2048 and :2317) are correct.

---

PR #2903, opened 2026-07-31
fix(core): validate module file internal wiring

SunSunSun689:
Fix standalone module validation for `dora expand --module`.

  The module docs say standalone module validation checks duplicate node IDs and internal wiring consistency, but `check_module_file` only validated `_mod/` references and declared outputs. This allowed invalid module files
  to pass as valid. Examples that previously passed incorrectly:

  ```yaml
  nodes:
    - id: proc
      path: a.py
    - id: proc
      path: b.py

  nodes:
    - id: preprocessor
      path: preprocessor.py
      outputs:
        - cleaned

    - id: producer
      path: producer.py
      inputs:
        data: preprocessor/cleaned
      outputs:
        - y

    - id: valid_consumer
      path: valid_consumer.py
      inputs:
        data: producer/y
      outputs:
        - z

    - id: consumer
      path: consumer.py
      inputs:
        data: producer/missing

    - id: independent
      path: independent.py
      outputs:
        - side
```
  The valid internal edges and unrelated nodes should be accepted, but the bad producer/missing edge should be rejected.
  This change makes check_module_file build a module-local output table,reject duplicate inner node IDs, and validate references to outputs produced by sibling inner nodes or nested modules.
  Standalone validation still does not reject unknown external sources, because a standalone module file has no parent dataflow context.


phil-opp:

No issues found. `collect_module_source_outputs` builds a node-id -> outputs map, rejects duplicate inner IDs, resolves nested-module outputs via canonicalize+load (rejecting absolute nested paths), and `check_internal_wiring` only enforces when a source matches a sibling node id — deliberately leaving unknown external sources permitted, consistent with the standalone "no parent context" rule. The added tests are real negative tests (duplicate inner node ids; an invalid `producer/missing` edge whose error is asserted to name that edge).

One non-blocking observation: since validation bails on the first bad edge, the invalid-wiring test doesn't independently prove the valid edges are *accepted* — a separate all-valid positive case would strengthen it. Not a defect.

---

phil-opp:

Re-reviewed after the merge from `main` and the follow-up repair commit (`104ec710`). No new issues. The repaired `collect_module_source_outputs` / `check_internal_wiring` pair is consistent with the merged `node_output_refs` API — it stores the `output_ref` element (so multi-operator `<op_id>/<output>` refs and bare single-operator/custom outputs both resolve correctly) and compares `mapping.output` against it. The duplicate-node-ID guard and the deliberate "unknown external source permitted" behavior are unchanged, and the added tests remain valid negative tests.

---

phil-opp:

Re-checked after `1a63cdd3` (and the `d15e97f` merge). The new commit only refines the regression test; the validation logic (`collect_module_source_outputs` / `check_internal_wiring`) is unchanged. The wiring check remains consistent with the addressing scheme in `node_output_refs` — produced outputs are stored as their `output_ref` (bare name for node-level `outputs`, single `operator:`, and legacy `custom:` outputs; `<op_id>/<output>` for multi-operator runtime nodes) and `check_internal_wiring` compares `mapping.output` against that same form, so valid sibling edges are not falsely rejected. The test now includes a `valid_consumer` edge (`producer/y`) accepted before the bad `producer/missing` edge, so it also exercises the accept path. No new issues found.

---

PR #2904, opened 2026-07-31
fix(core): resolve module inner sources relative to module file

SunSunSun689:
Fix module expansion so inner operator/custom source paths resolve relative
  to the module file.

  Module expansion already rewrites plain inner node `path:` values relative
  to the module file location. For example, a module at `modules/nested/
  source_paths.yml` with:

  ```yaml
  nodes:
    - id: worker
      path: worker.py

  expands to:

  path: modules/nested/worker.py

  However, equivalent source paths inside runtime operators, single
  operators, and legacy custom nodes were left unchanged:

  nodes:
    - id: runtime
      operators:
        - id: proc
          shared-library: libproc.so

    - id: single
      operator:
        python: single.py

    - id: runner
      custom:
        path: runner.py
        source: Local
```
  These stayed as libproc.so, single.py, and runner.py, so later validation/
  runtime resolution looked for them relative to the parent dataflow
  directory instead of the module directory.

  This change applies the same module-relative path resolution to:

  - operators[].shared-library
  - operators[].python
  - operators[].wasm
  - operator.shared-library
  - operator.python
  - operator.wasm
  - custom.path for local custom nodes

  URL and absolute paths are left unchanged. Relative paths that escape the
  project root through .. are rejected, matching the existing confinement
  behavior for plain inner node path:.

### CODE REVIEW

**Comment by phil-opp** [comment]

No issues found. The refactor extracts the existing path-confinement logic into `resolve_module_relative_path` and applies it to the previously-uncovered source kinds (`operators[].{shared-library,python,wasm}`, single `operator.*`, and `custom.path`). The URL/absolute short-circuit and the `strip_prefix(canonical_base)` traversal rejection preserve the prior semantics, and custom-node resolution is correctly gated on `NodeSource::Local` so remote sources are left alone. The added tests are meaningful: a negative test asserts a `../../etc/evil-operator` operator source is rejected with the containment message, and a positive test asserts the three source kinds are each rewritten relative to the module file.

---

**Comment by phil-opp** [comment]
**Automated review by Claude (not vetted by a human).**

One correctness issue in the new `custom.path` handling.

`resolve_module_relative_path` only skips resolution for URLs and absolute paths:

```rust
if super::source_is_url(path) || Path::new(path).is_absolute() {
    return Ok(());
}
```

It does not exempt the sentinel path values `DYNAMIC_SOURCE` (`"dynamic"`) and `SHELL_SOURCE` (`"shell"`). Because `resolve_inner_node_paths` now resolves `custom.path` for every `NodeSource::Local` custom node, a module whose inner node is a dynamic or shell node — e.g.

```yaml
# modules/nested/foo.yml
nodes:
  - id: dyn
    custom:
      source: Local
      path: dynamic
      inputs: { x: _mod/data }
      outputs: [y]
```

has its path silently rewritten to `modules/nested/dynamic`. `expand_modules` runs before `resolve_aliases_and_set_defaults`, so the sentinel is still live at this point, and both the validator (`descriptor/validate.rs:155-157`) and the daemon spawner (`daemon/src/spawn/command.rs:25-49`) detect dynamic/shell nodes by **exact string match** on `custom.path`. After the rewrite:

- `validate.rs` falls into the `source =>` branch and calls `resolve_path("modules/nested/dynamic", working_dir)`, failing with `Could not find source path`; and
- if validation is bypassed, the daemon no longer returns `Ok(None)` for the dynamic node and instead tries to exec a nonexistent `modules/nested/dynamic` binary.

The same latent gap exists in the pre-existing top-level `node.path` handling (a module inner node written as top-level `path: dynamic`), so the cleanest fix is to exempt the sentinels centrally in `resolve_module_relative_path`, e.g. short-circuit when `path == DYNAMIC_SOURCE || path == SHELL_SOURCE` alongside the existing URL/absolute checks. Operator sources (`shared-library`/`python`/`wasm`) have no such sentinels, so they're unaffected.

Suggested test: add a module inner custom node with `path: dynamic` and assert the expanded node's `custom.path` is still exactly `"dynamic"`.

---

phil-opp:

The latest commits (`db7b3d4` merge + `0fd6e93` "close module path resolver helper") don't change `resolve_module_relative_path`'s short-circuit logic, so the issue from my [previous review](https://github.com/dora-rs/dora/pull/2904#issuecomment-5269648437) is **still outstanding**:

`resolve_module_relative_path` skips only URLs and absolute paths, not the sentinel `custom.path` values `DYNAMIC_SOURCE` (`"dynamic"`) and `SHELL_SOURCE` (`"shell"`). Since `resolve_inner_node_paths` rewrites `custom.path` for every `NodeSource::Local` inner node, a module inner node that is a dynamic/shell node has its path rewritten to e.g. `modules/nested/dynamic`, which no longer matches the exact-string checks in `validate.rs` and daemon `spawn/command.rs` (`DYNAMIC_SOURCE => return Ok(None)`). Fix by exempting the two sentinels alongside the URL/absolute checks in `resolve_module_relative_path`, and add a test asserting an inner `path: dynamic` node stays exactly `"dynamic"` after expansion.

---

SunSunSun689:
Updated the PR to address the latest review feedback.

  Changes:
  - `resolve_module_relative_path` now preserves the `DYNAMIC_SOURCE` / `SHELL_SOURCE` sentinel values instead of rewriting them as module-relative
  paths.
  - Added regression coverage for module inner custom nodes using:
    - `custom.path: dynamic`
    - `custom.path: shell`

phil-opp:

The latest commits (`41b07922`, `2277ce21`) address the sentinel issue flagged in the earlier review. `resolve_module_relative_path` now short-circuits on the `DYNAMIC_SOURCE` (`"dynamic"`) and `SHELL_SOURCE` (`"shell"`) sentinels alongside the existing URL/absolute-path exemption, so an inner node written as `path: dynamic`/`path: shell` is no longer rewritten to `modules/<dir>/dynamic`. Because the exemption lives in the shared helper, it also covers the top-level `node.path` case, not just `custom.path`. The two added tests (`expand_preserves_dynamic_custom_source_sentinel`, `expand_preserves_shell_custom_source_sentinel`) are genuine regressions — they would fail without the exemption. No new issues found.

---

phil-opp:

The rebase onto current `main` (merge `490d1642`) materially changed this diff from the version I last reviewed: the `custom.path` handling is gone. That's correct rather than a regression — `Node` no longer has a `custom` field on `main`, so there's nothing to resolve there. What remains — routing `node.path`, `operators[].config.source`, and `operator.config.source` through the shared `resolve_module_relative_path`, with `DYNAMIC_SOURCE` / `SHELL_SOURCE` / URL / absolute exemptions — is sound and compiles against the current `OperatorSource` variants (`SharedLibrary` / `Wasm` / `Python`). Because the exemption now lives in the shared helper, it also covers plain `node.path`, which `main` previously rewrote unconditionally (a latent `path: dynamic` → `modules/<dir>/dynamic` bug); the two sentinel tests and the `..`-traversal test exercise the new paths. No new issues.

---

phil-opp:
Merged `main` (clean). The premise checks out — `validate.rs` resolves operator sources against the *dataflow* directory and the daemon spawns runtime nodes with that as cwd, so before this PR a module in `modules/` declaring `shared-library: proc` produced the bare dlopen name `libproc.so`, which `dlopen` looks for in the system paths and never in the module directory. Genuinely broken. All four new tests are real regression tests. 292 tests pass, clippy clean.

Pushed three fixes:

- **The new assertions would have gone red on Windows nightly.** `normalize_path` builds the rewritten path with `MAIN_SEPARATOR` and `strip_prefix` returns a sub-slice of it, so on Windows these come back as `modules\nested\libproc.so` while the assertions compare against `"modules/nested/…"` literals. PR CI is Linux-only, so this is invisible here and would have failed `test-cross-platform` on `windows-latest` and auto-filed a `nightly-regression` issue. Now compared on a normalized form.
- **`Path::new(path).is_absolute()` should be `is_absolute_any_platform`.** The file defines that helper specifically because `is_absolute()` returns false for a Unix-style `/foo` on Windows, and uses it for the `module:` path — but the new shared helper doesn't. On Windows, `shared-library: /opt/robotics/op` would fall through to `module_dir.join("/opt/…")`, which `PathBuf::push` resolves against the drive prefix, and the confinement check would then reject it with a misleading "resolves outside the project directory" instead of leaving the absolute path alone. Carried over from the moved code, but newly applied to operator sources.
- **Changelog + docs entry.** This changes user-visible path resolution for three descriptor fields and adds a new hard error, with neither. Documented in `docs/modules.md` (mirrored to the guide) and `Changelog.md`, including the footgun below.

Two things I documented rather than changed, because they're design calls:

- **`build:` still runs with the dataflow directory as cwd.** So a module declaring `build: cargo build -p my-op` with `shared-library: target/debug/my-op` gets its source rewritten to `modules/target/debug/my-op` while the artifact lands in `<dataflow>/target/debug/` — and `validate.rs` explicitly *skips* the existence check when `build.is_some()`, so `dora validate` passes and it surfaces as a raw dlopen error at spawn. The same footgun already exists for `path:`, so this PR is at least consistent, but "resolve relative to the module, build relative to the dataflow" isn't a principled split. Either skip resolution when `build.is_some()`, or run module inner-node builds with the module dir as cwd.
- **`../shared/op` now hard-fails.** A module sharing a library across sibling projects worked before; the confinement check now rejects it at expansion. This also means the repo's own `../../target/debug/<bin>` idiom can never be used from inside a module.

Nothing in-repo breaks — `examples/module-dataflow/` is the only module-using dataflow, it has no operators, and none of the 39 descriptors using `../../target/debug/…` are module files.

Left for you:

- **`git:`/`hub:` inner nodes still get their `path` rewritten.** For those the working directory is the *clone* dir, so `path: target/release/bar` must stay clone-relative — it becomes `modules/nested/target/release/bar` and isn't found. Pre-existing, not introduced here, but this PR creates the exact central seam where the exemption belongs and `resolve_inner_node_paths` has the whole `&mut Node` in hand, so it's a one-line guard.
- **No test at nesting depth ≥ 2.** The recursion's `module_dir`-as-child-`base_dir` handoff is the load-bearing line for depth-2 correctness, and `expand_nested_modules` puts both module files at `base/`, where the rewrite is a no-op at every level.
- **No idempotence test** — that a module sitting next to the dataflow leaves `shared-library: proc` as exactly `proc`, which is the path ~all existing users are on.
- **The PR body is pre-rebase**: it describes `custom.path`, `source: Local` and "legacy custom nodes", none of which exist any more. This repo squash-merges the body into the commit message.

phil-opp:

Reviewed the latest commit (`cf40d18`) since the previous review. The changes look correct and address the earlier concerns:

- The new test assertions are normalized for Windows path separators (they'd otherwise go red on the `windows-latest` nightly).
- The shared helper switched from `Path::new(path).is_absolute()` to `is_absolute_any_platform`, so a Unix-style `/foo` operator source isn't misrewritten into the module directory on Windows.
- Changelog / `docs` entries were added for the resolution change and the `build:`-cwd footgun.

No new correctness issues in the current diff.

One pre-existing item worth tracking separately (not introduced by this PR): `resolve_inner_node_paths` rewrites an inner `node.path` module-relative even when that node's source is `git:`/`hub:`, where the working directory is the clone dir and the path should stay clone-relative. The shared helper is now a natural seam for that exemption.

---

PR #2905, opened 2026-07-31
fix(core): propagate module node env to inner nodes

SunSunSun689:


phil-opp:

This PR propagates a module node's `env` map onto each expanded inner node (`propagate_module_node_env`), inserting the wrapper's entries so they override an inner node's own env on key conflicts, and does so *before* `substitute_params_in_node` injects the `PARAM_*` keys. Tracing `expand_module_node` and the recursive nested-module path: because propagation runs on every inner node — including nested module nodes — before recursion, an outer module node's env (and outer params materialized as `PARAM_*` entries) correctly flows down through multiple nesting levels.

The two added tests are meaningful, not tautological:
- `expand_module_node_env_propagates_to_inner_nodes` verifies both the union (inner-only + wrapper-only keys present) and wrapper-wins precedence (`SHARED == "wrapper"`).
- `expand_outer_params_reach_nested_module_inner_nodes_as_env` asserts `PARAM_SPEED` reaches a two-levels-deep `outer.inner.worker`, which would fail without the new propagation.

The one behavioral choice worth flagging for the maintainer (not a defect): the wrapper env *overrides* an inner node's explicit env on conflict. That's a defensible call-site-wins semantics, but it does mean an inner node cannot pin an env var against an outer override — worth confirming that's intended.

---

---

phil-opp:
Replaced by https://github.com/dora-rs/dora/pull/3122

PR #2906, opened 2026-07-31
fix(core): reject duplicate module header ports

SunSunSun689:


phil-opp:

`validate_module_header` rejects duplicate entries in `inputs`, `inputs_optional`, and `outputs`, plus the required/optional overlap, and is wired into both `check_module_file` and `expand_module_node` so it runs on the standalone `dora check` path and during expansion. The two added tests assert the error condition and the message substrings (`duplicate`, the field name, the port name) across all three port kinds — meaningful, not tautological.

One integration heads-up (not a defect in this diff): this PR overlaps with sibling PR #2907, which adds a `validate_module_header` function **with the same name** containing only the required/optional-overlap check that this PR already includes — plus a *nested-module* validation call site that this PR lacks. Since both branch from the same `main` commit and touch the same region, they will conflict textually on merge, and neither fully subsumes the other (this PR has duplicate-port rejection but not nested validation). Worth consolidating so nested modules also get the duplicate-port check.

---

---

PR #2907, opened 2026-07-31
fix(core): reject conflicting module input declarations

SunSunSun689:


phil-opp:

`validate_module_header` rejects an input declared in both `inputs` and `inputs_optional`, and is invoked at the top-level, nested-module, and expand sites. The nested case is genuinely exercised by `check_module_file_rejects_nested_module_conflicting_input_declarations`, which asserts the error names the offending nested module (`inner_bad`). Tests assert the `required and optional` message plus the port name — meaningful.

One integration heads-up (not a defect in this diff): sibling PR #2906 adds a `validate_module_header` function **of the same name** that already contains this required/optional-overlap check *plus* duplicate-port rejection, but wires it only at the top-level and expand sites (not the nested-module site this PR adds). Because both branch from the same `main` commit and touch the same region, the two PRs will conflict on merge and partially duplicate each other. The ideal end state is this PR's nested wiring combined with #2906's duplicate-port check — worth consolidating rather than merging both as-is.

---

---

phil-opp:
Closing as a strict subset of #2906. Both add a function named `validate_module_header` at the same two `load_module_file` call sites, but with different bodies — so they cannot both land (you get `error[E0428]: the name is defined multiple times` even after a clean git merge). #2906 covers everything here (duplicate `inputs`/`inputs_optional`/`outputs` entries *and* the required∩optional overlap) plus more.

The one thing worth carrying over is this PR's extra call on nested headers — please fold that into #2906.

PR #2908, opened 2026-07-31
fix(core): apply module build to inner node kinds

SunSunSun689:


phil-opp:

The refactor into `prepend_module_build_to_node` / `prepend_build` correctly extends module-build prepending to the actual inner-node kinds — `operators[].config.build`, single `operator.config.build`, and `custom.build` — which the previous code missed because it only touched the top-level `Node.build`. The test asserts the exact `pip install shared\n<inner>` result for all three kinds.

I checked the one thing that could have been a bug: `prepend_module_build_to_node` also unconditionally prepends to `node.build` even on explicit `custom`/`operator`/`operators` nodes. That is inert rather than a double-build — `resolve_aliases_and_set_defaults` only consumes top-level `Node.build` for `Standard` nodes and drops it for the explicit-kind arms, and `NodeExt::kind` keys off `path`/`operators`/`custom`/`operator`/`ros2`/`module`, not `build`. So no spurious second build step is produced.

---

---

PR #2912, opened 2026-07-31
fix(core): validate nested module required inputs

SunSunSun689:


phil-opp:

`check_nested_module_required_inputs` makes the standalone module linter (`check_module_file`) reject a nested-module reference that omits a required input of the nested module, mirroring the expand-time check. It correctly iterates only `nested_module.inputs` (required, not the optional set) and checks presence by key in the referencing node's `inputs`. The added test reproduces the gap and asserts on the module / input / node names. This is complementary to the sibling input-validation PRs rather than a duplicate.

---

---

phil-opp:
The check itself is fine. The thing to watch is merge ordering against #2907.

Both PRs insert at the same anchor in `check_module_file`, immediately after `let nested_module = load_module_file(&nested_canonical)?;`. If this one's call lands first, #2907's test `check_module_file_rejects_nested_module_conflicting_input_declarations` fails: its fixture wires `x: _mod/data` to a nested module declaring required input `data`, so this check bails with "declares required input `data`" and the assertion `msg.contains("required and optional")` never sees its message.

Whichever merges second will need to re-order the checks or re-fixture the test. Worth deciding now rather than discovering it as a red test after a rebase — note that #2907 is a strict subset of #2906, so folding it into #2906 and closing it would make this moot.

PR #2914, opened 2026-07-31
fix(cli,coordinator): normalize topic outputs across inspection and replay

DGHX12345:
## Issue

Topic inspection and output discovery treated node outputs inconsistently across descriptor kinds. Single-operator nodes exposed public topic names like `single/image`, but the coordinator debug path still needed runtime IDs like `op/image`, and several CLI paths only enumerated `node.outputs`.

That meant `topic echo`, `topic hz`, `topic list`, record/replay discovery, and coordinator-side debug subscriptions could disagree on which outputs existed for the same node.

## Fix

- Add a shared descriptor-aware output enumerator in `binaries/cli/src/command/topic/selector.rs`.
- Reuse it in `topic selector` resolution so empty selection and explicit `node/output` validation agree on public output names.
- Reuse the same helper in `topic list` so listing matches `echo`/`hz`.
- Normalize coordinator debug subscriptions to runtime-facing IDs before registering daemon watchers.
- Keep record/replay output discovery and replacement aligned with the same descriptor forms.

## Validation

- `cargo test -p dora-cli topic:: -- --nocapture`
- `cargo test -p dora-cli discovers_recordable_outputs_from_all_descriptor_node_kinds -- --nocapture`
- `cargo test -p dora-cli replay_replacement_preserves_outputs_from_all_descriptor_node_kinds -- --nocapture`
- `cargo test -p dora-coordinator topic_outputs_by_daemon_normalizes_single_operator_outputs -- --nocapture`


phil-opp:

I traced the topic-normalization changes across all four surfaces and found them consistent and correct; no issues found.

The main thing I checked was the bare-vs-operator-qualified output naming, since it differs deliberately by context:

- User-facing discovery (`topic/selector.rs::node_topic_outputs`, `record.rs::discover_descriptor_outputs`, `replay.rs::replay_node_outputs`) emits the bare output name for single-`operator:` nodes (`single/image`) and the operator-qualified name only for multi-`operators:` nodes (`runtime/op/status`). This matches how users reference these outputs in a descriptor, where the `op/` prefix for single-operator nodes is injected later during `resolve_aliases_and_set_defaults`.
- The coordinator path (`topic_outputs_by_daemon` → `resolved_topic_output_id`) correctly translates the bare form to the wire form `op/image` (`SINGLE_OPERATOR_DEFAULT_ID = "op"`), because that subscription routes directly to the daemon's resolved output ids. The added test asserts exactly `single` → `op/image` and would fail if the resolution were dropped.

I also confirmed the record/replay round-trip stays in bare form end-to-end: the record node stores `source_output` taken from the descriptor topic map (bare `image`, not the wire id), so the replay node re-emits `send_output("image", ...)` matching the bare `outputs: [image]` it now declares. The previous code declared no outputs at all for operator nodes, so this is a real fix.

---

phil-opp:

Traced the output-normalization across all five touched surfaces against `SINGLE_OPERATOR_DEFAULT_ID = "op"` and the expand.rs convention (single-`operator:` outputs referenced bare, `op/` injected only at resolution time). The scheme is internally consistent: the CLI surfaces use the bare user-facing name for single-operator nodes and the `<op_id>/` prefix only for multi-operator nodes, and the coordinator's `resolved_topic_output_id` maps the bare single-operator name to its `op/<output>` wire id for routing — so `topic echo`/subscription resolves correctly end-to-end. `record` and `replay` share the same bare-vs-prefixed rule, keeping them symmetric. The `Runtime`-with->1-operator branch correctly rejects an ambiguous bare name. Tests meaningfully cover all four node kinds on each surface.

---

---

phil-opp:
🤖 **Automated review by Claude** — this review was generated fully automatically, with no human in the loop.

Thanks for tightening output-id normalization across `topic`, record, and replay — the live-subscription and non-proxy record/replay paths look internally consistent. I found the following issue in the **proxy record** path for single-`operator:` nodes.

For a node declared with the `operator:` shorthand (e.g. `operator: { outputs: [image] }`, resolved operator id `op`), the recorded output id differs between the two record modes:

- Non-proxy `dora record` (record-node) embeds the **bare** id `image`: `discover_descriptor_outputs` yields `single/image` → `source_output = "image"`, and `binaries/record-node/src/main.rs` writes `output_id: image` into both the `RecordEntry` and the event bytes.
- `dora record --proxy` (`run_record_proxy` in `binaries/cli/src/command/record.rs`) copies `InterDaemonEvent::Output.output_id` **verbatim** into the entry and event bytes. With the new `resolved_topic_output_id` normalization, the daemon forwards the **wire** id `op/image`, so the proxy recording stores `op/image`.

On replay, `replay_node_outputs` declares the replacement node's `outputs:` as the bare `[image]` (via `append_outputs` from the operator block), but the replay node re-emits the `output_id` read from the recorded event bytes. So replaying a `--proxy` recording calls `send_output("op/image", …)` on a node that only declares `image`.

`send_output` → `validate_output` (`apis/rust/node/src/node/mod.rs`) returns `false` for the undeclared id, and `send_output` then returns `Ok(())` — the output is **silently dropped** (one-time "Ignoring output … not in node's output list" warning), not an error. Net effect: every replayed output from a `--proxy` single-operator recording is silently discarded and never reaches downstream subscribers, while the identical recording taken with plain `dora record` replays fine. The absence of a crash makes this easy to miss.

This combination is newly reachable in this PR: previously the proxy subscription bailed for single-`operator:` nodes (empty pre-resolution `node.outputs`). Multi-operator `operators:` nodes are unaffected (their ids are already `op/x` on both the record and replay sides).

Two ways to close it: normalize the `output_id` back to the bare public form in `run_record_proxy` before writing the entry (matching the non-proxy record node and the replay declaration), or have replay declare/emit the same wire id the recording carries. A round-trip test through `--proxy` for a single-operator node would pin this down.

---

phil-opp:

Re-checked the current diff after the 2026-08-18 rebase. The CLI live-subscription, `topic list`, and non-proxy record/replay paths are consistent. The `--proxy` record gap flagged in an earlier review is still unaddressed, so re-surfacing it since the PR is active again:

For a single-`operator:` node, the two record modes disagree on the stored `output_id`:
- Plain `dora record` (record-node) stores the bare public id `image`.
- `dora record --proxy` (`run_record_proxy`, `binaries/cli/src/command/record.rs:466-482`) copies `InterDaemonEvent::Output.output_id` verbatim. With the new `resolved_topic_output_id` normalization the daemon now forwards the wire id `op/image`, so the proxy recording stores `op/image`.

On replay, `replay_node_outputs`/`append_outputs` declare the replacement node's `outputs:` as the bare `[image]`, but `binaries/replay-node/src/main.rs:132` re-emits the recorded id via `node.send_output(output_id, …)` — i.e. `send_output("op/image", …)` on a node that only declares `image`. `validate_output` returns `false` and `send_output` returns `Ok(())`, so every replayed output from a `--proxy` single-operator recording is silently dropped (a one-time "Ignoring output … not in node's output list" warning) and never reaches subscribers — while the identical plain-`dora record` recording replays fine. No crash, so it's easy to miss. Multi-operator `operators:` nodes are unaffected.

Two ways to close it: normalize `output_id` back to the bare public form in `run_record_proxy` before writing the entry (matching record-node and the replay declaration), or have replay declare/emit the same wire id the recording carries. A `--proxy` round-trip test for a single-operator node would pin it down.

---

phil-opp:
Merged `main` and pushed fixes. The conflict in `binaries/coordinator/src/lib.rs` was textual only — both sides added new functions at the same spot (this PR's `resolved_topic_output_id`/`resolved_node_outputs`/`runtime_node_outputs`, main's `handle_dataflow_spawn_result` from #3134) — so I kept both; `handle_dataflow_spawn_result` is byte-identical to main's and nothing is duplicated or dropped.

**The root cause behind the comment I raised on 2026-08-12 and again on 08-18.** The PR normalizes only the *request* direction. Frames come back carrying the **wire** id — the daemon builds `InterDaemonEvent::Output { output_id }` from the resolved id and the coordinator forwards the payload opaquely — while every CLI consumer still keys, labels and stores by the public name it asked for. For a single-`operator:` node the two never meet. That's three bugs, not one:

1. **`topic hz` reported 0 samples for every single-`operator:` topic.** The index was built from public ids and frames matched by wire ids, so the `<all>` aggregate row showed real traffic while the per-topic row read zero. Before this PR the subscription failed loudly; silently reporting a wrong number is the worse failure.
2. **`record --proxy` + `replay` silently discarded every message.** The recording stored `op/image` while `replay_node_outputs` declared `image`, so replay-node called `send_output("op/image", …)`, `validate_output` returned false, and `send_output` still returned `Ok(())`. One warning line, no data. A plain `dora record` recording of the same dataflow replays fine, because record-node stores the bare `source_output`. Newly reachable here: before the PR the topic wasn't recorded at all, so "not recorded" became "recorded but silently unreplayable".
3. **`topic echo` labelled frames with the wire id**, printing `webcam/op/image` for what the user typed as `webcam/image` and what `list` and `info` both display as `webcam/image`.

Fixed with `public_topic_output_id` and its inverse next to `node_topic_outputs`. `hz` keys its index by the wire id so the translation is paid once at setup rather than per frame; `echo` labels with the public id.

**One design call you should sanity-check.** For `--proxy` I rewrite the id in *both* the entry and the embedded event, rather than the cheaper alternative of leaving the payload alone and having replay-node prefer `entry.output_id`. Reason: the id inside the event is what replay-node re-emits, so an untouched payload carries the wire id no matter what the entry says, and any other `.drec` consumer would see it too. Rewriting both makes a `--proxy` recording shaped exactly like one from the record node, which stores the bare id in both places. The cost is one re-serialize per frame — the loop already fully deserializes every frame just to read two ids, so it's one extra copy. If you'd rather keep the recorded bytes verbatim as a forensic record of what the daemon sent, say so and I'll switch it. This needs the descriptor typed, so record parses it a second time, deliberately non-fatally: a descriptor that fails the typed parse still records, just without the translation.

Also in the same push:

- **`topic list`/`info` subscriber columns missed every operator-node consumer.** Both iterated `&node.inputs`, which is always empty for `operator:`/`operators:` nodes — their inputs live in `operator.config.inputs`. The PR fixed the output half of this asymmetry in `list.rs` and left the symmetric input half in the same function, so on `python-operator-dataflow` all three outputs listed with an empty Subscribers cell.
- **Coordinator errors.** `unknown output 'image' for custom node` dropped the node id, the dataflow id and the available outputs the old message carried, and gave the 2+-operator ambiguity case the same text as a plain typo. Rejecting a bare id on a multi-operator node is the right call — nothing in a descriptor can reference one — so the message now lists the qualified names, which is the hint the user actually needs.
- **`topic_outputs_by_daemon` re-resolved a descriptor the coordinator already has resolved.** `RunningDataflow.nodes` is that exact result; I checked every mutation site and `spawn_dataflow`, `AddNode` and `ReplaceNode` all write it together with the descriptor. Using it drops a per-subscribe clone-and-rewalk and a `?` on a path that otherwise can't fail. That exposed `test_running_dataflow` handing out a real descriptor with an empty resolved map — a state production never produces — so the helper now populates both, and the two `"path"` fixture additions are no longer load-bearing for that reason.
- **Dead `custom:` branches** in `record.rs` and `replay.rs`, plus the test fixtures built on them: `Node` is `deny_unknown_fields` and has no `custom` field, so such a descriptor can't deserialize — you'd already made this substitution in the coordinator test with a comment saying so.
- **Documented the contract** in `docs/cli.md` and `docs/websocket-topic-data-channel.md`. This PR makes operator-node topics user-visible for the first time and establishes a two-form contract; three call sites got it wrong, so it belongs in writing.

`proxy_recorded_output_id_is_declared_by_the_replay_node` is the regression test for the record/replay half — verified RED without the fix (`recorded 'op/image' … is not in the replay node's declared outputs ["image"]`). The round-trip and subscriber-enumeration tests cover the new helpers, but note they test new functions, so they can't go RED against the old code the way that one does.

`cargo test -p dora-cli` (360) and `-p dora-coordinator` (155) pass; clippy `--all-targets -D warnings` clean.

Round-tripping an already-qualified `op/image` worked before and still does — that part of the PR was right.

phil-opp:

Re-checked after the 2026-08-19 rebase and fix commit. The `--proxy` record gap flagged in the prior reviews is now closed: `run_record_proxy` rewrites the wire id back to the public id via `public_topic_output_id` in both the entry and the re-serialized event, so a `--proxy` recording is shaped like a plain `dora record` one and replays. The `public_topic_output_id`/`wire_topic_output_id` round-trip, the coordinator's `resolved_topic_output_id` (bare→`op/<output>` for single-operator, reject-with-hint for multi-operator), and the `node_topic_inputs`/`node_topic_outputs` symmetry all check out. No new issues.

---

Issue #2916, opened 2026-07-31
Bug: `dora node remove` -> `node add` (same id) race — old process's SIGKILL is attributed to the freshly added node, marking it failed


heyong4725:
**Versions:** dora-cli 1.0.0-rc.4; daemon/coordinator + Python node API built from git rev `7eb4a5f8b` (CLI cargo-installed from the same rev). macOS arm64 (Darwin 25.5.0).

## What happens

Hot-swapping a Python node on a live dataflow via `dora node remove -d <df> <id>` immediately followed by `dora node add -d <df> --from-yaml <same-id node>` succeeds at the CLI level, but the daemon then attributes the REMOVED process's kill to the node identity, ~15 ms AFTER the add:

```
18:29:02.565671  INFO  removing node from running dataflow   node_id=grasp-planner-topdown
18:29:02.577049  INFO  adding node to running dataflow       node_id=grasp-planner-topdown
18:29:02.592507  INFO  node added successfully               node_id=grasp-planner-topdown  dynamic=false
18:29:02.592585  WARN  process was killed on drop because it was still running
18:29:02.593590  ERROR node exited with error: Signal(9)     node_id=grasp-planner-topdown
```

From that point the dataflow keeps running but the node is dead/marked failed: in our
pipeline every later episode failed with the planner silent, and `dora stop` reported
`Dataflow <uuid> failed: Node grasp-planner-topdown failed: exited because of signal SIGKILL`
even though the SIGKILL belonged to the OLD instance.

## Repro

1. `dora up`; `dora start graph.yaml --uv --name df --detach` (any graph with a Python
   node `n` consuming a timer works).
2. `dora node remove -d df n && dora node add -d df --from-yaml n.yaml` (back-to-back,
   as any scripted caller does; `n.yaml` = the same node entry).
3. Watch the daemon log: "node added successfully" then "node exited with error: Signal(9)".

**Workaround we ship:** sleep ~2 s between remove and add — with the gap, the old exit is
accounted first and the replacement runs correctly (verified over repeated live swaps).

## Asks

- `node remove` should return only after the removed process is reaped AND its exit is
  accounted — or process accounting should be keyed by process instance/generation,
  not by node id, so a late exit of the old instance cannot poison the new one.
- Bigger picture: an atomic **`dora node replace -d <df> <id> --from-yaml <node>`** is the
  primitive live-iteration workflows actually want (remove+add with no observable gap and
  no identity confusion).

heyong4725:
[...]

PR #2925, opened 2026-07-31
test(lifecycle): cover the same-id node hot-swap race, and actually run the suite in CI

heyong4725:
## What

Regression coverage for the `dora node remove` → `dora node add` (same id) race reported in dora-rs/dora#2916, plus a CI fix without which none of it — or the existing coverage — runs.

No daemon changes: the reported bug is already fixed on `main`.

## Why

The reporter built from `7eb4a5f8b` (Jul 26). The stale-exit pid guard in dora-rs/dora#2893 landed Jul 30, four days later, and addresses it.

While confirming that, two things turned up:

**1. The reported ordering was untested.** `rust_dynamic_node_readd_same_id_ignores_stale_exit` covers a node that ignores `Stop` and dies to the grace SIGTERM seconds after the re-add. The issue's node honored `Stop` and exited on its own — just slowly enough (Python interpreter teardown) that the exit still landed *after* the re-add. That is the window where the exit gets accounted against the live successor: `handle_node_stop` removes the new `running_nodes` entry, and dropping its `ProcessHandle` SIGKILLs the fresh process.

My first attempt at this test used the existing `filter` fixture and **passed with the pid guard reverted** — it never entered the race, because that node exits ~1 ms after `Stop`, always ahead of the next CLI round trip. Hence the new fixture: `stop-delay-node` takes its shutdown latency and exit code from the descriptor's `env:` block, so the ordering is pinned explicitly instead of inherited from how slow a language runtime happens to be.

**2. The existing regression test never runs in CI.** The e2e job filtered by name (`lifecycle_rust lifecycle_cxx`), which silently drops every test not starting with `lifecycle_` — including the one #2893 shipped with its own fix. Inverting to `--skip lifecycle_python` takes selection from 2 tests to 5 and closes the file by default.

## Test plan

- [x] `hot_swap_survives_predecessor_exit_after_readd` verified **RED** with #2893's pid guard reverted — the re-added node goes `Failed` immediately, matching the reported log — and green with it restored.
- [x] `hot_swap_survives_predecessor_exit_before_readd` passes on unmodified `main`. Noted inline that it holds only incidentally: the removed incarnation's failure *is* recorded under the bare node id and stays there until the successor's own clean exit overwrites it. The assertion guards against that becoming a real contradiction.
- [x] `cargo test -p dora-examples --test node-lifecycle-e2e -- --test-threads=1 --skip lifecycle_python` — the newly selected set passes locally except `lifecycle_cxx_dynamic_add_remove`, which needs a cmake toolchain this machine lacks and was already in CI's selection before this PR.
- [x] `cargo fmt --all -- --check`, `cargo clippy` on the new fixture and test target.
- [x] `lifecycle_python_dynamic_add_remove` fails on this machine, but identically on clean `main` with these changes stashed — pre-existing and environmental, and unaffected by this PR (that test runs in the contract-tests job).

## Notes for reviewers

Two id-keyed paths I traced but did not touch, neither reproducing today:

- `ProcessHandleReplaced` is still keyed by node id alone, so a restart loop from a dead incarnation can overwrite — and thereby drop, hence kill — a live successor's process handle.
- The stale `dataflow_node_results[node_id]` entry above survives if the successor never reports its own result, e.g. a dynamic-node successor.

Happy to file either separately if they're worth tracking.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -396,7 +396,12 @@ jobs:
           --
           --test-threads=1
           --skip lifecycle_python
-        timeout-minutes: 20
+        # 5 tests, each driving a real coordinator + daemon + node
+        # processes, plus nested `cargo build` calls for the fixtures.
+        # ~4 min on a warm dev machine; a cold 2-vCPU runner is several
+        # times that, and a step killed at the cap leaves a detached
+        # coordinator holding port 6013 for the next step.
+        timeout-minutes: 30
       # Finish-straggler watchdog regression (#2270): a deterministic wedge
       # (source exits, downstream node hangs) must be escalated/killed by the
       # watchdog rather than hanging until stop_after. Lives in `dora-examples`
@@ -500,8 +505,14 @@ jobs:
           --test-threads=1
           --nocapture
       # Python variant of the lifecycle test piggybacks on this job
-      # because it needs uv (already set up above). See the e2e job
-      # for the Rust + C++ variants.
+      # because it needs uv (already set up above). The e2e job runs
+      # every OTHER test in the file via `--skip lifecycle_python`, so
+      # the two jobs partition it by that one substring: any test in
+      # node-lifecycle-e2e.rs that needs uv MUST be named
+      # `lifecycle_python*`, or it lands in the uv-less e2e job instead.
+      # Note libtest exits 0 when a filter matches nothing, so renaming
+      # the Python test without updating this filter would make the step
+      # silently run zero tests.
       - name: Run dora node lifecycle E2E (Python)
         run: >
           cargo test -p dora-examples --test node-lifecycle-e2e
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -379,18 +379,24 @@ jobs:
       # `dora node` lifecycle E2E (issue #1703). The shared test
       # binary lives under the `dora-examples` package which the bulk
       # `cargo test --all` step excludes, so it needs an explicit
-      # invocation here. Filter to the Rust + C++ variants because
-      # the Python variant requires uv (handled in `contract-tests`
-      # below where uv is already installed).
-      - name: Run dora node lifecycle E2E (Rust + C++)
-        # Cargo's TESTNAME positional is single-valued; multiple
-        # substring filters go after `--` for libtest.
+      # invocation here.
+      #
+      # Excludes rather than enumerates: the Python variant is the only
+      # test in the file this job can't run (it needs uv — handled in
+      # `contract-tests` below where uv is already installed), so
+      # `--skip` closes the file by default. The previous positive
+      # filters (`lifecycle_rust lifecycle_cxx`) silently dropped every
+      # test whose name didn't start with `lifecycle_` — including
+      # `rust_dynamic_node_readd_same_id_ignores_stale_exit`, the
+      # regression test that shipped with the #2893 stale-exit fix,
+      # which therefore never ran here.
+      - name: Run dora node lifecycle E2E (non-Python)
         run: >
           cargo test -p dora-examples --test node-lifecycle-e2e
           --
           --test-threads=1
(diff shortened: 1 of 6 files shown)


heyong4725:
Pushed a follow-up that hardens the new test against a vacuous pass.

`hot_swap_survives_predecessor_exit_after_readd` only covers the reported ordering if the predecessor outlives the `dora node add` round trip. On a loaded runner that can silently invert — the exit lands before the re-add, the test exercises the *other* ordering, and still passes. That is exactly how the first draft of this test came out green against a reverted pid guard, so it seemed worth closing rather than leaving to timing.

It now asserts the predecessor is still alive at the moment the re-add returns, and fails with a message naming the cause when it isn't. Verified by setting the fixture delay to 0: the assertion fires rather than the test passing. Delay raised to 4s for margin, still well inside the 10s stop grace.

One note for whoever runs this in the merge queue: the expensive tier is gated `if: github.event_name != 'pull_request'`, so the lifecycle step this PR touches doesn't run on the PR itself — the filter change is exercised for the first time in the queue. I verified the selection locally instead:

```
$ cargo test -p dora-examples --test node-lifecycle-e2e -- --skip lifecycle_python --list
hot_swap_survives_predecessor_exit_after_readd: test
hot_swap_survives_predecessor_exit_before_readd: test
lifecycle_cxx_dynamic_add_remove: test
lifecycle_rust_dynamic_add_remove: test
rust_dynamic_node_readd_same_id_ignores_stale_exit: test
5 tests, 0 benchmarks
```

against 2 tests for the previous `lifecycle_rust lifecycle_cxx` filter. Say the word if you'd rather I trigger a `workflow_dispatch` run on this branch first — I held off on spending the runner time unasked.

heyong4725:
Ran the exact command this PR installs in CI, on a clean machine:

```
$ cargo test -p dora-examples --test node-lifecycle-e2e -- --test-threads=1 --skip lifecycle_python
test hot_swap_survives_predecessor_exit_after_readd ... ok
test hot_swap_survives_predecessor_exit_before_readd ... ok
test lifecycle_rust_dynamic_add_remove ... ok
test rust_dynamic_node_readd_same_id_ignores_stale_exit ... ok
test lifecycle_cxx_dynamic_add_remove ... FAILED
test result: FAILED. 4 passed; 1 failed; finished in 203.31s
```

The one failure is the C++ fixture failing to link locally on macOS (`clang++: error: linker command failed`, in `ensure_cxx_built`'s cmake step) — a local toolchain limitation, not a dataflow assertion. That test was already selected by the previous `lifecycle_rust lifecycle_cxx` filter, so this PR doesn't change its risk on the Linux runner either way.

The three tests the new filter *adds* to CI all pass, including `rust_dynamic_node_readd_same_id_ignores_stale_exit`, which had never run there.

Given that, I don't think a `workflow_dispatch` dry run is worth the runner time — happy to trigger one anyway if you'd prefer belt-and-braces before it hits the queue.

heyong4725:
Fixed in 10d61cb48 — good catch, and it's the exact mirror of the vacuous-pass problem already fixed in the sibling test.

One correction on the suggested signal: `filter` never becomes `Failed`. I probed a live daemon across the remove → re-add window, polling every 250 ms:

```
--- before remove ---
{"node":"filter","status":"Running","pid":"11525",...}
t+0.25s : (absent)
t+0.50s : (absent)
...
t+3.00s : (absent)
```

The row disappears the instant the node is removed and never reappears to report the later exit, so there's no accounting state to poll on. Process liveness is the observable that exists, so the test now waits for the predecessor to exit before re-adding. The exit event is then queued before the replacement's CLI round trip even begins, and a slower runner only widens that margin rather than inverting it.

Validated the wait isn't decorative by giving the predecessor a 2s lingering exit — precisely the case that previously would have flipped the ordering — and confirming the test still passes while covering the intended path. Restored to 0 afterwards; `git diff` against the pre-experiment file is clean.

Both hot-swap tests green locally: `2 passed; 0 failed; finished in 85.44s`.

PR #2928, opened 2026-07-31
fix(node): JSON-escape bare-string input in read_json_bytes_as_arrow

phil-opp:
> [!NOTE]

## Issue

`read_json_bytes_as_arrow` (`apis/rust/node/src/daemon_connection/json_to_arrow.rs`) has a string fallback used when the raw input is not valid JSON: it wraps the input bytes in literal double quotes and re-tries schema inference. The old `wrapped_quoted` did this **without JSON-escaping** the bytes:

```rust
fn wrapped_quoted(data: impl BufRead) -> impl BufRead {
    let quoted = "\"".as_bytes().chain(data).chain("\"".as_bytes());
    wrapped(quoted)
}
```

Any input containing a `"`, `\`, a newline, or another control character therefore produced **malformed JSON**, so `infer_json_schema` failed and the whole call bailed.

This is the path that backs the interactive `Data` prompt's advertised bare-string input. Typing `he said "hi"` produced the wrapper `{ "inner":"he said "hi""}` (invalid JSON), so the input was silently rejected. Same for any multi-line paste. The bare-string path only worked for strings with no JSON-special characters.

## Fix

JSON-encode the bytes as a proper string value instead of raw-quoting them:

```rust
fn wrapped_quoted(data: &[u8]) -> impl BufRead {
    let s = String::from_utf8_lossy(data);
    let escaped = serde_json::to_string(s.as_ref()).unwrap_or_else(|_| "\"\"".to_string());
    wrapped(std::io::Cursor::new(escaped.into_bytes()))
}
```

`from_utf8_lossy` keeps non-UTF-8 input producing valid JSON rather than failing outright.

## Validation

- New regression tests cover embedded quotes, backslash + newline, and a plain string (round-trip through `read_json_bytes_as_arrow` to a `StringArray`).
- `cargo test -p dora-node-api`, `cargo clippy -p dora-node-api -- -D warnings`, `cargo fmt --all -- --check` all pass.

---

### DIFF
--- a/apis/rust/node/src/daemon_connection/json_to_arrow.rs
+++ b/apis/rust/node/src/daemon_connection/json_to_arrow.rs
@@ -39,10 +39,56 @@ fn wrapped(data: impl BufRead) -> impl BufRead {
     "{ \"inner\":".as_bytes().chain(data).chain("}".as_bytes())
 }
 
-// wrap data into JSON object to also allow bare JSON values
-fn wrapped_quoted(data: impl BufRead) -> impl BufRead {
-    let quoted = "\"".as_bytes().chain(data).chain("\"".as_bytes());
-    wrapped(quoted)
+// wrap data into a JSON object, treating the input as a bare string value.
+//
+// The raw bytes are JSON-encoded (not merely surrounded by quotes) so that
+// strings containing `"`, `\`, newlines, or other control characters still
+// yield *valid* JSON. Previously the bytes were wrapped in literal quotes
+// without escaping, so any such character produced malformed JSON and made
+// schema inference fail — meaning the advertised "bare string" input path
+// only worked for strings with no JSON-special characters.
+fn wrapped_quoted(data: &[u8]) -> impl BufRead {
+    // `from_utf8_lossy` keeps non-UTF-8 input producing valid JSON (via the
+    // replacement character) instead of failing outright. `to_string` on a
+    // string value cannot fail, but fall back defensively just in case.
+    let s = String::from_utf8_lossy(data);
+    let escaped = serde_json::to_string(s.as_ref()).unwrap_or_else(|_| "\"\"".to_string());
+    wrapped(std::io::Cursor::new(escaped.into_bytes()))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::read_json_bytes_as_arrow;
+    use arrow::array::{Array, StringArray};
+
+    fn parse_string(input: &str) -> String {
+        let data = read_json_bytes_as_arrow(input.as_bytes())
+            .unwrap_or_else(|e| panic!("failed to parse {input:?}: {e}"));
+        let array = arrow::array::make_array(data);
+        let strings = array
+            .as_any()
+            .downcast_ref::<StringArray>()
+            .expect("expected a string array for bare-string input");
+        assert_eq!(strings.len(), 1);
+        strings.value(0).to_string()
+    }
+
+    #[test]
+    fn bare_string_with_embedded_quotes() {
+        // Regression: the string fallback used to wrap the raw bytes in literal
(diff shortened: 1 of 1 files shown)


phil-opp:

This is the summary comment for a batch of machine-generated (Claude) PRs from one automated in-depth review pass over the repository. Each PR was verified against a freshly-fetched `origin/main`, branched independently, and validated per affected crate (fmt + clippy + tests); the combined diff additionally passed the `/review` and `/simplify` gates and the full workspace test suite.

### PRs opened, by category

**Correctness**
- #2928 — `fix(node)`: JSON-escape bare-string input in `read_json_bytes_as_arrow` (embedded `"`/`\`/newlines produced malformed JSON and rejected the advertised bare-string input path).
- #2929 — `fix(cli)`: apply the log level filter before `--tail`, not after (`dora logs --tail N --level error` could show zero lines even when many errors exist).
- #2930 — `fix(cli)`: flush buffered logs before `dora run` returns (detached printer thread was never joined, so a short dataflow's final log lines could be lost).
- #2931 — `fix(daemon)`: free per-dataflow node results on a persistent daemon (`dataflow_node_results` grew unbounded over a coordinator-managed daemon's lifetime).

**Documentation**
- #2932 — `docs(arrow-convert)`: add a compile-checked doctest for `into_vec` (documents null / unsupported-type / out-of-range-cast failure modes).

### Findings considered but rejected (with reason)

- **`dora cluster upgrade` may report success against the old daemon** — plausible race, but not verifiable without a live cluster and no clean minimal fix; deferred.
- **`dora up` "daemon not connected after Ns" off-by-one** — the message over-reports elapsed time by one 100 ms tick; cosmetic only.
- **`rotate_log_files` with `max_rotated_files == 0` keeps one `.1` file** — minor off-by-one, never panics, doesn't leak unboundedly; low value.
- **`CompatibilityGraph::is_compatible` universal `Bytes` sink uses a raw-string compare** — real inconsistency, but only yields a spurious *warning* and `std/core/v1/Bytes` is not parameterized in the std library; marginal impact.
- **`types_match` / `QueuePolicy::effective_cap` lack doctests** — valid, but shipped a single representative doctest (#2932) rather than several near-identical low-value doctest PRs.
- **`NodeError` signal-name table is x86/ARM-Linux-specific** — cosmetic; degrades gracefully to the raw number on other ABIs.
- **MAVLink TCP/UDP builders mis-format IPv6 hosts (bracketed literal)** — real but low-impact (IPv6 MAVLink is uncommon) and arguably out of the module's documented `host:port` scope.

---

phil-opp:

Reviewed the change. The escaping fix is correct: `serde_json::to_string` on a `Cow<str>` produces a properly-escaped JSON string value, and `read_json_bytes_as_arrow` already calls `wrapped_quoted(data)` twice against the same `&[u8]` (once for schema inference, once for reading), which the new `&[u8]` signature supports fine since each call builds a fresh `Cursor`. `from_utf8_lossy` keeps non-UTF-8 input producing valid JSON. The regression tests cover the embedded-quote / backslash / newline cases that previously produced malformed JSON. No issues found.

---

phil-opp:

Confirmed the escaping is correct and complete — `serde_json::to_string` on the `&str` covers `"`, `\`, and the full C0 control range, and leaves valid UTF-8 untouched, all valid JSON. The `&[u8]` signature change is fine since each call site gets its own owned `Cursor`. Tests round-trip embedded quotes, backslash+newline, and the plain case through the actual parser and would fail on the old code.

---

---

phil-opp:
Closing as a duplicate of #3080, which fixes the same bug and is the stronger version: it escapes once at the call site instead of re-escaping for both the infer and the read pass, drops a dead `unwrap_or_else` fallback, and adds a regression test that a bare JSON number still parses as a number rather than as a string.

PR #2929, opened 2026-07-31
fix(cli): apply the log level filter before --tail, not after

phil-opp:
> [!NOTE]

## Issue

`dora logs --tail N` trimmed to the last `N` lines **before** the minimum-level (`--level`) and per-node (`--log-filter`) filters ran. Those filters are only enforced inside `print_log_message` → `should_display`, which executes *after* `apply_tail`.

As a result, `dora logs --tail 20 --level error` could display **zero** lines even when many error lines exist: `apply_tail` kept the last 20 lines of *all* levels (mostly stdout/info in a healthy run), and the level filter then dropped them all. The local (`--local`), all-nodes, and follow-initial-batch paths had the same defect.

The single-node coordinator path was worse: the level filter was not part of the "fetch everything and tail client-side" decision, so the coordinator itself was asked to pre-tail — discarding older matching lines before the CLI ever saw them.

## Fix

- Add `apply_level_filter` and drop level-filtered messages **before** `apply_tail` in all four log paths (local, follow-local initial batch, all-nodes, single-node).
- Treat an active level filter like the other client-side filters (`since`/`until`/`grep`) when deciding whether to fetch all logs from the coordinator, via a new `level_filter_is_active` helper (`min_level != Stdout` or per-node filters set; the default level is `stdout`, which drops nothing).

## Validation

- New unit tests: `tail_counts_only_level_filtered_lines` (the early-errors-then-many-infos case) and `level_filter_active_predicate`.
- `cargo test -p dora-cli`, `cargo clippy -p dora-cli -- -D warnings`, `cargo fmt --all -- --check` all pass.

---

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -12,8 +12,8 @@ use crate::{
         send_control_request,
     },
     output::{
-        LogFormat, LogOutputConfig, parse_jsonl_line, parse_log_filter, parse_log_level_str,
-        print_log_message,
+        LogFormat, LogOutputConfig, level_filter_is_active, message_passes_level_filter,
+        parse_jsonl_line, parse_log_filter, parse_log_level_str, print_log_message,
     },
     ws_client::WsSession,
 };
@@ -249,7 +249,8 @@ fn read_local_logs(args: &LogsArgs) -> Result<()> {
     all_messages.sort_by_key(|a| a.timestamp);
     let filtered = apply_time_filters(all_messages, args.since, args.until, now);
     let grepped = apply_grep(filtered, args.grep.as_deref());
-    let display = apply_tail(grepped, args.tail);
+    let leveled = apply_level_filter(grepped, &config);
+    let display = apply_tail(leveled, args.tail);
 
     for msg in display {
         print_log_message(msg, &config);
@@ -294,7 +295,8 @@ fn follow_local_logs(args: &LogsArgs) -> Result<()> {
     all_messages.sort_by_key(|a| a.timestamp);
     let filtered = apply_time_filters(all_messages, args.since, args.until, now);
     let grepped = apply_grep(filtered, args.grep.as_deref());
-    let display = apply_tail(grepped, args.tail);
+    let leveled = apply_level_filter(grepped, &config);
+    let display = apply_tail(leveled, args.tail);
 
     for msg in display {
         print_log_message(msg, &config);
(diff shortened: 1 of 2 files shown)


phil-opp:

Reviewed the change. `message_passes_level_filter` (output.rs) delegates to the same `should_display` that `print_log_message` uses, so the pre-tail filter and the final print filter are guaranteed consistent — the re-filter in `print_log_message` is harmless and idempotent. Inserting `apply_level_filter` before `apply_tail` in all four paths is correct, and treating `level_filter_is_active` like the other client-side filters when deciding whether to fetch all logs from the coordinator fixes the pre-tail-at-coordinator truncation. `level_filter_is_active` is conservative (a permissive per-node filter still forces a full fetch), which is never incorrect. The regression test captures the buggy "tail then filter" ordering. No issues found.

---

phil-opp:

Verified the fix against `should_display`/`passes` semantics: the default config (`min_level = Stdout`, no node filters) drops nothing, so `level_filter_is_active` correctly returns false and behavior is unchanged when no level filter is set; `message_passes_level_filter` shares `should_display` with `print_log_message`, so pre-tail filtering matches print-time filtering exactly. The single-node fetch-all decision now correctly includes the level filter alongside `--since/--until/--grep`. The regression test reproduces the real failure and pins the correct ordering.

---

---

phil-opp:
Correctness checks out across the board: all four display pipelines are patched, the `-f` follow loop and `stream_logs_from_coordinator` correctly need no change since `--tail` never applies to live lines, and `level_filter_is_active` is right because `LogLevelOrStdout::Stdout` is both the permissive minimum and the clap default — so the coordinator-side pre-tail optimization survives for the default invocation. Re-filtering in `print_log_message` is idempotent, so no double-print.

One coverage note: the new test at `binaries/cli/src/command/logs.rs:1249` exercises `apply_level_filter` / `apply_tail` composition in isolation, not the production call order. If someone reverts the four call sites back to `apply_tail(grepped, ...)`, the test still passes. It goes RED before this change only because the helper does not exist yet, not because it guards the ordering — so it would not catch the regression it is nominally protecting against.

phil-opp:
Good catch — the isolated-composition test wouldn't have caught a revert of the four call sites. Fixed:

- Extracted the shared pipeline (time filter → grep → level filter → tail) into a single `filter_and_tail` helper, and routed all four non-follow log paths (local, follow-local initial batch, all-nodes, single-node) through it — also removing the four-way duplication.
- The ordering test now drives `filter_and_tail` directly, so it goes RED if the level filter is ever moved after the tail in the production path, rather than only exercising the helpers in isolation.

Verified with `cargo test`/`clippy`/`fmt` for `dora-cli`.

---

phil-opp:
> 🤖 **Automated review by Claude** — fully automated; not vetted by a human.

Reviewed the latest commit (`4cdcf15`). The refactor addresses the coverage concern raised earlier: all four non-follow log paths now route through a single `filter_and_tail` helper (time filter → grep → level filter → tail), and the regression test drives that production helper directly, so it goes RED if the level filter is ever moved after `--tail`. The ordering is correct and the four-way duplication is gone. `level_filter_is_active` still returns false for the permissive default, preserving the coordinator pre-tail optimization for the default invocation. No issues found.

---

PR #2930, opened 2026-07-31
fix(cli): flush buffered logs before `dora run` returns

phil-opp:
> [!NOTE]

## Issue

`dora run` printed log messages on a detached `std::thread` whose `JoinHandle` was discarded:

```rust
let (log_tx, log_rx) = flume::bounded(100);
std::thread::spawn(move || {
    for message in log_rx { print_log_message(message, &log_config); }
});
```

When `Daemon::run_dataflow` finished, up to 100 `LogMessage`s could still be queued in the channel. Nothing joined the printer thread before `execute` returned and the process exited, so the **final log lines of a short dataflow** — often the node output or error the user ran `dora run` to see — could be lost. `rt.shutdown_timeout` does not wait on this thread because it is not a tokio worker.

## Fix

Keep the printer's `JoinHandle` and join it after `block_on` and the runtime shutdown. At that point the daemon task has completed and the runtime is torn down, so every clone of the log-channel sender has been dropped and the channel is closed — the join drains and prints all remaining messages. The result is now flushed on both the success and daemon-error paths (a daemon error's buffered logs are printed before we bail). A printer panic is propagated via `resume_unwind`.

## Validation

- `cargo build -p dora-cli`, `cargo clippy -p dora-cli -- -D warnings`, `cargo fmt --all -- --check` all pass.
- This is a thread-lifecycle/timing fix not amenable to a deterministic unit test (it would require driving a real dataflow); the existing `dora run` smoke tests exercise the path.

---

### DIFF
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -195,7 +195,7 @@ impl Executable for Run {
         };
 
         let (log_tx, log_rx) = flume::bounded(100);
-        std::thread::spawn(move || {
+        let printer = std::thread::spawn(move || {
             for message in log_rx {
                 print_log_message(message, &log_config);
             }
@@ -237,14 +237,23 @@ impl Executable for Run {
             )
             .await
         });
-        let result = rt
-            .block_on(handle)
-            .context("dora-run daemon task panicked")??;
+        let result = rt.block_on(handle);
         // Bound runtime shutdown to prevent hanging on blocking Drop impls
         // (e.g. zenoh::Session::drop blocks tokio workers on macOS during
         // TCP teardown). Without this, `rt` drops implicitly at end of scope
         // and waits indefinitely for all worker threads to exit (#2287).
         rt.shutdown_timeout(Duration::from_secs(10));
+        // The daemon task has finished and the runtime is shut down, so every
+        // clone of the log-channel sender has been dropped and the channel is
+        // closed. Join the printer thread so the last messages still buffered
+        // in the channel are flushed before we return. Without this, the
+        // detached printer could be killed by process exit mid-drain and the
+        // final log lines of a short dataflow (often the ones the user cares
+        // about) would be lost.
+        if let Err(panic) = printer.join() {
+            std::panic::resume_unwind(panic);
+        }
+        let result = result.context("dora-run daemon task panicked")??;
         handle_dataflow_result(result, None)
     }
 }
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -263,10 +263,16 @@ impl Executable for Run {
         };
 
         let (log_tx, log_rx) = flume::bounded(100);
-        std::thread::spawn(move || {
+        // `printer_done` lets us wait for the printer to finish draining with a
+        // timeout (see the flush after `shutdown_timeout` below) instead of
+        // blocking on its `JoinHandle`, which could hang forever if a leaked
+        // tokio worker still holds a `Sender<LogMessage>` clone.
+        let (printer_done_tx, printer_done_rx) = std::sync::mpsc::channel();
+        let printer = std::thread::spawn(move || {
             for message in log_rx {
                 print_log_message(message, &log_config);
             }
+            let _ = printer_done_tx.send(());
         });
 
         // Drive `Daemon::run_dataflow` on a tokio worker thread, not the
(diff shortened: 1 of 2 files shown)


phil-opp:

The fix is correct in the normal case: once `block_on(handle)` returns and `rt.shutdown_timeout` completes, the sender clones held by `run_dataflow` are dropped, `log_rx` closes, and `printer.join()` drains and returns.

One low-probability concern at `binaries/cli/src/command/run.rs:250-256`: `printer.join()` is now an **unbounded** wait, whereas the immediately preceding `rt.shutdown_timeout(Duration::from_secs(10))` deliberately does *not* wait forever — it returns after 10s and leaks (detaches) any worker still stuck in a blocking `Drop` (the `zenoh::Session::drop` case in #2287 that this very code guards against). If a leaked worker still holds a clone of the `LogDestination::Channel` sender (e.g. a detached logging task spawned inside `run_dataflow` that outlives the awaited task), `log_rx` never closes and `printer.join()` blocks `dora run` forever. Before this PR the detached printer meant the process would still exit in that scenario.

This requires a leaked worker that both holds a log-sender clone and is stuck past the 10s timeout, so it's unlikely in practice — but since the surrounding code exists precisely because that teardown hang is real, consider bounding the join (or ensuring all sender clones are provably confined to the awaited task) so a leaked worker can't turn lost-logs into a hung command.

---

phil-opp:

The flush fix is sound: `log_tx` is moved into the spawned daemon task and no clone is kept on the calling thread, so once `block_on(handle)` returns and `rt.shutdown_timeout` drops the workers, every sender is dropped, the channel closes, and the joined printer drains cleanly. Unwrapping the daemon result after the join is also correct (a daemon error still flushes logs first).

Minor, non-blocking: the join relies on all sender clones being dropped; the only way it could stall is a worker leaked by `shutdown_timeout` while still holding a sender — not reachable here since the sender drops when the daemon future completes before the join.

---

---

phil-opp:

The bug is real — the printer thread is detached, `flume::bounded(100)` can hold up to 100 messages, and nothing drains them before `execute` returns.

But the fix trades a rare loss of log lines for a rare CLI that never exits. `printer.join()` returns only once every clone of the `Sender<LogMessage>` is dropped, and those clones live in detached tasks (`Logger::try_clone` at `binaries/daemon/src/log.rs:463`, moved into `tokio::spawn`ed node-log tasks at `binaries/daemon/src/spawn/prepared.rs:753`, `:949`). Normally `rt.shutdown_timeout` drops them. But `shutdown_timeout` exists in this file precisely *because* a worker can wedge in a blocking `Drop` — the comment at `run.rs:315-318` cites `zenoh::Session::drop` blocking tokio workers on macOS (#2287). After the timeout tokio **leaks** the wedged thread, its task is never dropped, its sender clone never drops, and `printer.join()` blocks forever.

Bound the flush the same way the runtime shutdown is bounded:

```rust
let (done_tx, done_rx) = std::sync::mpsc::channel();
let printer = std::thread::spawn(move || {
    for message in log_rx { print_log_message(message, &log_config); }
    let _ = done_tx.send(());
});
// … after rt.shutdown_timeout(…) …
if done_rx.recv_timeout(Duration::from_secs(5)).is_ok()
    && let Err(panic) = printer.join()
{
    std::panic::resume_unwind(panic);
}
```

Separately, the branch is stale: the pre-image blob for `run.rs` is `1eb4a2283` while `main` is `a96c0c094`, so the hunks are ~70 lines off. It still applies, but please rebase and re-run the `dora run` smoke tests.

The body's "not amenable to a deterministic unit test" is fair for the thread-lifecycle part, but the bounded-join variant above is testable, and CLAUDE.md's tier table asks for a smoke test for CLI behavior changes.

phil-opp:
You're right on all counts — thank you, this would have traded a rare lost-line for a rare hang. Fixed:

- **Bounded flush.** The printer now signals completion over an `mpsc` channel; after `rt.shutdown_timeout(…)` we `recv_timeout(Duration::from_secs(5))` and only then `join()` (which returns immediately). If a worker wedged in a blocking `Drop` was leaked past `shutdown_timeout` (the #2287 case), its `Sender<LogMessage>` clone never drops, the channel never closes, and the drain never signals — so we give up the last few lines after 5s instead of hanging the CLI forever. Essentially your suggested shape.
- **Rebased.** The branch was stale (unrelated-history artifact on my side, not a clean fast-forward), so I rebuilt it on current `main` (`9f7c7e49`) and force-pushed; the hunks now apply against the current `run.rs`.
- **Smoke test.** Added `smoke_rust_dataflow_run_flushes_logs` (per the CLAUDE.md tier table): it runs the rust-dataflow example under `dora run --stop-after` and asserts the sink's `sink received message:` line reaches stdout before the command returns — a hang there (rather than the missing-line assertion) would catch a regression back to an unbounded join.

Verified with `cargo build`/`clippy -D warnings`/`fmt` for `dora-cli` and the new smoke test, all on rustc 1.95.0.

---

phil-opp:

Re-reviewed after `ca6c8ddc`. The unbounded-join hang raised in the prior review is addressed: the printer signals completion over an `mpsc` channel, and after `rt.shutdown_timeout(...)` the code waits `printer_done_rx.recv_timeout(Duration::from_secs(5))` before joining. So if `shutdown_timeout` leaked a tokio worker wedged in a blocking `Drop` (the #2287 case), its surviving `Sender<LogMessage>` clone can no longer keep the channel open and wedge `dora run` forever — the CLI gives up the last few lines after 5s instead. Flushing still occurs on both the success and daemon-error paths (the daemon result is unwrapped only after the flush), and a printer panic is propagated via `resume_unwind`. No new issues found.



---

PR #2931, opened 2026-07-31
fix(daemon): free per-dataflow node results on a persistent daemon

phil-opp:
> [!NOTE]

## Issue

In `finish_dataflow` (`binaries/daemon/src/lib.rs`), a finished dataflow's entry was read out of `dataflow_node_results` with `.get().cloned()` but **never removed**, and `self.running.remove()` right next to it left the results map untouched:

```rust
node_results: self.dataflow_node_results.get(&dataflow_id).cloned().unwrap_or_default(),
...
```

In the common case, every finished dataflow's node results (inserted for each non-dynamic node on stop, `Ok` and `Err` alike) stayed resident. Unlike every sibling per-dataflow map, this one is only drained at daemon shutdown (`std::mem::take` in `run_inner`), so a long-lived **coordinator-managed daemon accumulated one entry per finished dataflow** — steady growth over its lifetime, affecting normal dataflows, not just failed ones.

## Why the naive fix is wrong

Simply switching `.get().cloned()` to `.remove()` breaks `dora run`: that run-once path reads the map via `run_inner`'s `std::mem::take` **after** `finish_dataflow` returns, so removing early yields *"no node results for dataflow_id"*.

The two paths are distinguished by `exit_when_done`: `Some(..)` for the run-once path (set in `run_dataflow`), `None` for the coordinator-managed daemon (`build_daemon(..., None, ...)`).

## Fix

Extract an `extract_node_results` helper that:
- **removes** the entry when the daemon is coordinator-managed (`exit_when_done.is_none()`), because nothing reads it again after the result is reported; and
- **clones and keeps** it for the run-once path, where `run_inner`'s later `mem::take` still needs it.

## Scope / limitation

This **reduces** the growth; it does not fully close it. As the code documents around the straggler race, a late `SpawnedNodeResult` can insert into `dataflow_node_results` *after* `finish_dataflow` has run (the insert happens before `handle_node_stop`, which then early-returns because the dataflow already left `self.running`). Since `finish_dataflow` never runs again for that id, such a re-created entry still leaks. That path is much rarer than the per-dataflow accumulation this PR eliminates, and closing it fully means guarding the straggler inserts — a deeper change to the node-stop path that's better as a focused follow-up. This PR is strictly better than the status quo without touching that path.

## Validation

- New unit tests: `persistent_daemon_removes_the_entry`, `run_once_daemon_keeps_the_entry`, `missing_entry_yields_empty_results`.
- `cargo test -p dora-daemon`, `cargo clippy -p dora-daemon -- -D warnings`, `cargo fmt --all -- --check` all pass.


phil-opp:

Reviewed the change. `finish_dataflow` is the only reader of a dataflow's `dataflow_node_results` entry (it reads it once to build the `DataflowDaemonResult`), and it runs after all node-stop inserts and immediately before `running.remove`, so removing the entry on the persistent path (`exit_when_done.is_none()`) is safe — nothing reads it again. The run-once path keeps the entry so `run_inner`'s later `std::mem::take` still returns it. The `keep_entry = exit_when_done.is_some()` mapping matches the two paths correctly, and the unit tests exercise both plus the missing-entry case. No issues found.

---

phil-opp:

The polarity is correct (`exit_when_done.is_some()` → keep for the `dora run` `mem::take` drain; `None` → remove on the coordinator-managed daemon), and removal is safe: all three `finish_dataflow` call sites gate on `self.running` still containing the dataflow, and `finish_dataflow` removes that running entry, so a duplicate finish (which would now report empty `node_results`) cannot occur. The results are already delivered to the coordinator via `AllNodesFinished` and the map isn't read again for a finished dataflow. Tests cover both branches meaningfully.

---

---

phil-opp:
Polarity and safety check out: `exit_when_done.is_some()` is exactly the `dora run` path, which consumes the map and hard-errors if the entry is missing, so keeping it there is required; the persistent path discards `run_inner`'s return and has no other reader; and keying by `Uuid` makes dataflow-*name* reuse a non-issue.

One correction to the body: this reduces the growth rather than closing it. The straggler race the code documents at `binaries/daemon/src/lib.rs:5321-5337` re-creates the entry *after* `finish_dataflow` has run — `lib.rs:6025` and `lib.rs:1850` insert into `dataflow_node_results` before calling `handle_node_stop`, which then early-returns because the dataflow already left `self.running`. Since `finish_dataflow` never runs again for that id, that entry leaks permanently. Still strictly better than today, but "grows without bound over the daemon's lifetime" is overstated.

The merge conflict is pure line drift — all three anchors are byte-identical on `main`.

phil-opp:
Thanks — both points addressed:

- **Body corrected.** Updated the description to state this *reduces* rather than *closes* the growth, and added a "Scope / limitation" section documenting the straggler-race path (a late `SpawnedNodeResult` re-inserting after `finish_dataflow`, which then can't run again for that id) as a known residual, with fully closing it called out as a focused follow-up on the node-stop path.
- **Merge conflict resolved.** It was pure line-drift at EOF: my `node_results_cleanup_tests` module and `main`'s new `health_check_tests` module were both appended after `announce_zenoh_bind_tests`. Merged `main` and kept both modules; the `finish_dataflow`/`extract_node_results` anchors merged unchanged. Verified with `cargo test`/`clippy`/`fmt` for `dora-daemon` on the merged branch (all 8 tests green — my 3 plus the 5 health-check tests from `main`).

---

Issue #2917, opened 2026-07-31
Bug: a failing dynamic node's error is delivered into ANOTHER node's `Node()` init (cross-talk between concurrent dynamic-node initializations)


heyong4725:
**Versions:** dora-cli 1.0.0-rc.4; daemon + Python node API from git rev `7eb4a5f8b`. macOS arm64.

## What happens

Two dynamic nodes added in quick succession: node A (`dora node add`, a Python node whose
init takes a couple of seconds — heavy imports), then ~120 ms later node B (a short-lived
probe that crashed at spawn, before registering). Node B's failure was delivered as the
error result of node A's `Node()` initialization:

```
(node A stderr)
  File ".../grasp_topdown.py", line 216, in main
    node = Node()
RuntimeError: Could not initiate node from environment variable. For dynamic node, please add a node id in the initialization function.
Caused by:
   0: internal error
   1: failed to init event stream
   2: subscribe failed: Node probe-e9bcdc2f exited before initializing dora. ...
Location:
    apis/python/node/src/lib.rs:665:39
```

`probe-e9bcdc2f` is node B — a different node entirely. Node A then exited (ExitCode(1)),
so one crashing dynamic node took down an unrelated healthy one that happened to be
mid-init.

Daemon timeline:

```
18:46:01.317 INFO  adding node ...                node_id=grasp-planner-topdown
18:46:01.319 INFO  node added successfully        node_id=grasp-planner-topdown
18:46:01.442 INFO  adding node ...                node_id=probe-e9bcdc2f
18:46:01.444 INFO  node added successfully        node_id=probe-e9bcdc2f
18:46:01.502 ERROR node exited with error: ExitCode(1)  node_id=probe-e9bcdc2f
18:46:01.518 ERROR node exited with error: ExitCode(1)  node_id=grasp-planner-topdown
```

## Repro

1. Live dataflow; `dora node add` a Python node with slow imports (so its `Node()` call
   is still pending), then immediately `dora node add` a second node that exits before
   registering (e.g., a script that crashes at import).
2. The first node's `Node()` raises with the SECOND node's failure in the cause chain.

**Workaround we ship:** sequence dynamic adds — wait for the earlier node's init to finish
before adding another. With sequencing, both nodes work.

## Ask

Scope daemon init/subscribe replies per requesting node so one dynamic node's failure can
never be delivered as the outcome of another node's initialization.

heyong4725:
[...]

PR #2933, opened 2026-08-01
fix(daemon): scope startup-barrier failures to the startup cohort

heyong4725:
## What

A node that exits before subscribing lands in `exited_before_subscribe`, and `answer_subscribe_requests` hands every waiting subscriber one shared `Err` naming it. That is correct while the dataflow is starting — if a node dies then, nothing can start.

Nodes added at runtime (`dora node add`) were enrolled in that same cohort by `AddNode`, so the error reached nodes that had no relationship to it. This scopes the barrier to the descriptor-declared cohort.

Addresses #2917. Deliberately no closing keyword — leaving it to the reporter to close after retesting on their own pipeline.

## Two reproduced consequences

**1. A healthy node inherits an unrelated node's crash.** Added after an unrelated node crashed, its own `Node()` init fails with:

```
filter: Caused by:
filter:    2: subscribe failed: Node crasher exited before initializing dora.
```

and it exits. `exited_before_subscribe` is never cleared, so this sticks: one crashed node poisons **every later `dora node add` for the life of the dataflow**. Sequencing the adds — the workaround in the issue — does not help this variant; the test sequences them fully and it still fails.

**2. A crash during the startup window kills the dataflow.** A runtime addition that crashes while the dataflow's own nodes are still subscribing is broadcast to that cohort and takes it down — a dataflow that was running fine disappears. Found while writing the first test, which initially failed with a misleading "no dataflow is running".

## The change

Runtime additions register via `insert_runtime_added` instead of `local_nodes`. They still **wait** on the barrier — a node added while the dataflow is coming up should not start producing before its consumers are listening — but they:

- cannot fail it (`handle_node_stop` only records `local_nodes` members into `exited_before_subscribe`),
- cannot inherit its failures (`answer_subscribe_requests` scopes the reply per waiter),
- cannot stall it by never subscribing (previously they held `local_nodes` non-empty and blocked `AllNodesReady` indefinitely).

Descriptor-declared nodes are untouched — the cohort is still populated from the descriptor at spawn time.

## Test plan

Both tests verified RED by stashing **only** the daemon change, then green with it:

| Test | Without fix | With fix |
|---|---|---|
| `dynamic_add_survives_earlier_crash_before_subscribe` | healthy node `Stopped` | ok |
| `crash_before_subscribe_does_not_kill_the_startup_cohort` | `sender=None receiver=None` | ok |

The second exists because a partial revert — scoping the subscribe reply but still enrolling runtime additions in the cohort — leaves the first green while the dataflow-killing variant stays broken.

The crasher is `false(1)` rather than a new fixture crate: it needs to do nothing at all, and another workspace member would add another nested `cargo build` to an already-slow CI job.

Regression sweep, all with the fix applied:

- [x] `cargo test -p dora-daemon` — 136 passed
- [x] `fault-tolerance-e2e` — 11/11
- [x] `ws-cli-e2e` — 25/25
- [x] `node-lifecycle-e2e --skip lifecycle_python` — 6/7; the failure is `lifecycle_cxx_dynamic_add_remove` at its cmake step, a pre-existing macOS linker issue that fails identically on a clean tree
- [x] `cargo fmt --all -- --check`, `cargo clippy -p dora-daemon -- -D warnings`

## For reviewers

**This is a behavior change beyond the bug:** a runtime addition that never subscribes no longer blocks `AllNodesReady`. I read that as correct for what the barrier is for, but it deserves a maintainer's eye.

Unrelated, noticed while linting: `binaries/daemon/src/lib.rs:6513` (`matches_event`) trips `clippy::match_like_matches_macro` under `--all-targets`. CI lints without that flag so it isn't gating; left alone.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -396,11 +396,12 @@ jobs:
           --
           --test-threads=1
           --skip lifecycle_python
-        # 5 tests, each driving a real coordinator + daemon + node
-        # processes, plus nested `cargo build` calls for the fixtures.
-        # ~4 min on a warm dev machine; a cold 2-vCPU runner is several
-        # times that, and a step killed at the cap leaves a detached
-        # coordinator holding port 6013 for the next step.
+        # Every non-Python test in the file, each driving a real
+        # coordinator + daemon + node processes, plus nested `cargo
+        # build` calls for the fixtures. ~5 min on a warm dev machine; a
+        # cold 2-vCPU runner is several times that, and a step killed at
+        # the cap leaves a detached coordinator holding port 6013 for
+        # the next step.
         timeout-minutes: 30
       # Finish-straggler watchdog regression (#2270): a deterministic wedge
       # (source exits, downstream node hangs) must be escalated/killed by the
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2298,16 +2298,14 @@ impl Daemon {
                     if is_dynamic {
                         dataflow.dynamic_nodes.insert(node_id.clone());
                     }
-                    // Register as a runtime addition rather than a
-                    // member of the startup cohort. This node was not in
-                    // the descriptor the dataflow started from, so it
-                    // must neither gate that cohort's barrier nor
-                    // inherit its failures — previously it joined
-                    // `local_nodes`, which meant a crash here was
-                    // broadcast as the subscribe result of unrelated
-                    // nodes, and a node that never subscribed could
-                    // stall startup outright (dora-rs/dora#2917).
-                    dataflow.pending_nodes.insert_runtime_added(node_id.clone());
+                    // Deliberately NOT enrolled in `pending_nodes`: this
+                    // node is not part of the descriptor the dataflow
+                    // started from, so it must neither gate that
+                    // cohort's barrier nor inherit its failures.
+                    // Enrolling it meant a crash here was broadcast as
+                    // the subscribe result of unrelated nodes, and a
+                    // node that never subscribed could stall startup
+                    // outright (dora-rs/dora#2917).
 
                     // Insert the running node
                     dataflow.running_nodes.insert(node_id.clone(), running_node);
@@ -2459,6 +2457,13 @@ impl Daemon {
                     // subscribes and could be selected mid-startup (dora#2270).
                     dataflow.connected_nodes.remove(&node_id);
                     dataflow.finish_escalated.remove(&node_id);
+                    // Drop the node from the startup barrier too. A node
+                    // removed before it subscribed would otherwise keep
+                    // gating startup until it exits, and its id would
+                    // still reach `exited_before_subscribe` — which
+                    // nothing clears — poisoning every later subscribe
+                    // on this dataflow (dora-rs/dora#2917).
+                    dataflow.pending_nodes.remove_node(&node_id);
                     // Purge per-node bookkeeping keyed by node id that the
                     // routing cleanup above doesn't touch. Otherwise stale
                     // input_deadlines/broken_inputs entries are re-scanned
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2457,13 +2457,6 @@ impl Daemon {
                     // subscribes and could be selected mid-startup (dora#2270).
                     dataflow.connected_nodes.remove(&node_id);
                     dataflow.finish_escalated.remove(&node_id);
-                    // Drop the node from the startup barrier too. A node
-                    // removed before it subscribed would otherwise keep
-                    // gating startup until it exits, and its id would
-                    // still reach `exited_before_subscribe` — which
-                    // nothing clears — poisoning every later subscribe
-                    // on this dataflow (dora-rs/dora#2917).
-                    dataflow.pending_nodes.remove_node(&node_id);
                     // Purge per-node bookkeeping keyed by node id that the
                     // routing cleanup above doesn't touch. Otherwise stale
                     // input_deadlines/broken_inputs entries are re-scanned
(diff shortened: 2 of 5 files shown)


heyong4725:
Pushed `ab17241f` after an adversarial review pass. Two of its findings were right and material.

**The exemption was keyed on the wrong predicate.** It tested "was added at runtime", which misses descriptor-declared dynamic nodes — those are never enrolled in `local_nodes` either, so they were still handed the cohort's shared failure. Two spellings of the same node kind behaved differently. The predicate is now "was in the startup cohort", recorded once in `insert()`, which covers both by construction and drops the extra set.

**My stated invariant was false.** The comment claimed runtime additions could never reach `exited_before_subscribe`. `RemoveNode` purges nine per-node maps but never touched `pending_nodes`, so removing a node that hadn't yet subscribed left it in `local_nodes`; a re-add under the same id sat in the cohort again, and its exit poisoned the never-cleared list — naming an id that is alive again by then. `remove_node` now scrubs `local_nodes`, `cohort`, any parked subscriber, and that node's entry in `exited_before_subscribe`.

**Unit tests.** `pending.rs` had no test module despite being a pure state machine; the whole scoping decision rested on two Unix-only e2e tests with 8s sleeps. Added 7 tests that pin it in milliseconds: cohort members still inherit failures, non-cohort and dynamic nodes do not, cascading errors are recorded for cohort members only, runtime additions do not gate the barrier, `remove_node` scrubs every field, and remove-then-re-add comes up clean.

Also caught a flaw in my own verification, worth stating: my first "RED without the fix" re-check stashed only the *uncommitted* changes, but `HEAD` already contained v1 of the fix, so the tests passed and appeared to show the tests were vacuous. Re-ran properly against `origin/main`'s daemon:

```
test crash_before_subscribe_does_not_kill_the_startup_cohort ... FAILED
  the dataflow's own nodes must survive ...; got sender=None receiver=None
test dynamic_add_survives_earlier_crash_before_subscribe ... FAILED
test result: FAILED. 0 passed; 2 failed
```

and green with the change: `dora-daemon` 143 unit tests, `node-lifecycle-e2e` 6/7 (`lifecycle_cxx` fails at its cmake step, pre-existing macOS linker issue).

Two findings I did **not** act on, with reasons:

- **Distributed dataflows hang.** With external nodes, `update_dataflow_status` parks waiters and returns `Pending` without answering; the only path that answers them is a one-shot at dataflow start, so anything subscribing later waits forever. Real, entirely pre-existing, and it means `dora node add` is broken on multi-machine deployments regardless of this PR. I documented it next to `cohort` rather than fixing it blind — no multi-daemon test to verify against, and it deserves its own change.
- **Missing "exited before initializing dora connection" warn for non-cohort nodes.** The specific warn is gated on cohort membership, but the failure is not silent: the daemon still logs `node exited with error: ExitCode(1)` with the node id, and `dora node list` shows `Failed`. I confirmed both while reproducing the bug. Not worth a third set to recover one log line.

heyong4725:
Fixed in `418ee6eb`. The P1 was correct and it was a regression I introduced in `ab17241f` — worth being explicit that it made things **worse than before this PR**, not merely incomplete.

Before: a removed-but-unsubscribed node stayed in `local_nodes`, so its later exit released the barrier via `handle_node_stop` — with the wrong error attached, which is exactly what #2917 is about, but it did release. After my scrub: the id was gone from `local_nodes`, so that exit path became a no-op, and removing the last pending member emptied the barrier without ever firing it. Parked subscribers stayed parked and `dataflow.start()` was never called. I traded a wrong error for a hang.

`handle_node_removal` now scrubs and then drives the same transition a subscription does, and the `RemoveNode` handler acts on the returned status exactly as the `Subscribe` handler does — starting the dataflow when the removal is what completed the cohort. It moved outside that handler's sync closure since it is async.

Regression test as requested: `removing_the_last_pending_member_releases_parked_subscribers`. One cohort member is parked waiting, the other never subscribes and is removed; the test asserts the removal reports `AllNodesReady` and that the parked subscriber is actually answered.

I verified it catches the defect rather than assuming it does — with the re-evaluation removed it fails, and it is the only one that does:

```
test pending::tests::removing_the_last_pending_member_releases_parked_subscribers ... FAILED
  removing the last pending member must report the cohort ready
test result: FAILED. 8 passed; 1 failed
```

Sweep on the fix: `dora-daemon` 144 unit tests, `fault-tolerance-e2e` 11/11, `node-lifecycle-e2e` 6/7 (`lifecycle_cxx` fails at its cmake step, pre-existing macOS linker issue). The lifecycle suite exercises `dora node remove` in three separate tests, which is the coverage that matters for this change.

Also renamed a stale doc reference to `remove_node` that the rename left behind.

heyong4725:
Ran the testing and maintainability specialists. Pushed `d267af4f`. The testing pass was the more valuable of the two — it mutation-verified its claims, and three of my tests could not fail.

**Deleted as vacuous.** `runtime_additions_do_not_gate_the_barrier` asserted `HashSet::is_empty` after a hand-removal; the behaviour it named lives in the `AddNode` arm and was never invoked, so restoring the enrolment left it green. `dynamic_nodes_are_not_blamed_for_startup_failures` was a line-for-line duplicate of the non-cohort test — `PendingNodes` has no notion of dynamic-ness, so they could never diverge.

**Fixed an inert assertion.** `removal_scrubs_barrier_state` checked `!waiting_subscribers.contains_key(..)`, which cannot fail: `answer_subscribe_requests` drains that map with `mem::take` regardless. Confirmed by the reviewer deleting the scrub and watching it still pass. It now asserts the removed node's receiver was *dropped* rather than answered.

**Covered the paths that had none.** `handle_node_stop` is the only production writer of `exited_before_subscribe` — every test pushed to it by hand, so the cohort-only invariant the whole fix rests on was unverified. Two tests now cover it, and I checked they discriminate: making the guard unconditional fails exactly one. Also added the `external_nodes` branch and `handle_external_all_nodes_ready`.

**The `RemoveNode` wiring had zero coverage** — deleting that whole block left every test green, including the unit test for the method it calls. That is the P1 from the last round, guarded only by a comment. There is now an e2e for it.

Worth recording how that e2e went, because it repeated the mistake this PR keeps hitting. My first version asserted `dora node list` showed the parked node `Running` — and it **passed with the wiring deleted**. `Running` only means the process is in `running_nodes`, not that its `Node()` init ever returned. The fixture now writes `DORA_TEST_READY_FILE` once `init_from_env` returns, and the test waits on that file. Verified both directions:

```
with the wiring:    ok
without the wiring: FAILED
  removing the last pending node did not release the barrier: `quick` never got
  past its `Node()` init, so the dataflow never started.
```

Maintainability findings applied: the removal-triggered start now emits the same Info log the other two start sites do (a dataflow starting because of a removal previously left no trace); the unreachable `None` arm reports an error instead of silently succeeding where the closure above would have errored; the call-site comment no longer restates the method doc; the external-nodes caveat moved off the `cohort` field onto `update_dataflow_status` and cites #2938; three inlined `node add` blocks now use the existing helper.

Corrected in the e2e comments: stale `file:line` citations that had drifted, RED-phase text still describing the pre-fix outcome as current, duplicate step numbers, and a claim of mine that a partial revert would fail one test but not its sibling — under that revert both fail, so the stated discriminator was wrong.

One reviewer claim I did not accept: that `crash_before_subscribe_does_not_kill_the_startup_cohort` passes pre-fix. It does not — against `origin/main`'s daemon it fails with `sender=None receiver=None`. The underlying concern (nothing observes that the barrier was still open) is fair, and its doc now records both the verified discrimination and the residual timing dependence.

Sweep: `dora-daemon` 146 unit tests, `node-lifecycle-e2e` 7/8 (`lifecycle_cxx` at its cmake step, pre-existing macOS linker issue).

phil-opp:

No issues found. I traced the core logic and it holds up:

- **Scoping** (`answer_subscribe_requests`): `scoped_result` is `result.clone()` only for `cohort` members and `Ok(())` otherwise, and the cascading-error report is now gated on `scoped_result.is_err()`, so a non-cohort waiter (runtime `dora node add` or descriptor-declared dynamic node) is neither blamed nor recorded as a victim. Keying on `cohort` membership rather than "added at runtime" correctly covers both spellings, since `PendingNodes` has no notion of dynamic-ness.
- **Removal** (`handle_node_removal`): scrubs `local_nodes`, `cohort`, the parked subscriber, and the `exited_before_subscribe` entry, then re-evaluates via `update_dataflow_status`. Driving the barrier transition here (and acting on `AllNodesReady` in the `RemoveNode` handler exactly as the `Subscribe` handler does) correctly avoids the hang where removing the last pending member would otherwise strand parked subscribers — the removed process's own exit can't release the barrier since its id is already gone from `local_nodes`.

The test suite is genuinely discriminating rather than vacuous: it covers the production writer (`handle_node_stop`) of `exited_before_subscribe`, the `RemoveNode` wiring via the `DORA_TEST_READY_FILE` observable (rather than the misleading `Running` status), the external-nodes path, and the last-pending-member release. The known distributed-dataflow limitation (#2938) is documented rather than papered over. Nice work.

---

PR #2939, opened 2026-08-01
fix(daemon): gate health-check watchdog on node connection (#2937)

### LINKED ISSUE #2937: daemon: `check_node_health` has no "connected" guard (its `last == 0` guard is dead code), so a node with a `health_check_timeout` shorter than its cold-start time is SIGKILLed mid-startup — asymmetric with the finish-straggler watchdog
> **This issue was created by a scheduled automated Claude check** (an unattended code-review routine). It is a **lower-confidence** finding — the mechanism is verified from the code, but whether `health_check_timeout` is *intended* to cover the pre-connect startup window is a design call for the maintainers. Flagging it mainly because of a concrete asymmetry with a sibling watchdog that was explicitly fixed for the same trap.

## Summary

`Daemon::check_node_health` (`binaries/daemon/src/lib.rs:2694`) kills any node whose `last_activity` is older than its configured `health_check_timeout`. Its only startup guard is `if last == 0 { continue }`. But `last_activity` is **seeded to `current_millis()` at spawn time**, not `0`, so that guard is effectively dead code — `last` is never `0` in production. There is no `connected` check. As a result the timeout clock starts at **spawn**, before the node has had any chance to communicate, and a node whose cold start legitimately exceeds `health_check_timeout` is SIGKILLed before it ever connects.

The sibling finish-straggler watchdog was given an explicit `connected` guard for *exactly* this seeding trap; `check_node_health` was not.

## The code

`binaries/daemon/src/lib.rs:2694`:

```rust
fn check_node_health(&self) {
    let now_millis = node_communication::current_millis();
    for dataflow in self.running.values() {
        for (node_id, node) in &dataflow.running_nodes {
            let Some(timeout) = node.health_check_timeout else { continue };
            let last = node.last_activity.load(atomic::Ordering::Acquire);
            if last == 0 {
                continue; // "not yet connected"  <-- never true: seeded non-zero at spawn
            }
            let elapsed_ms = now_millis.saturating_sub(last);
            if elapsed_ms > timeout.as_millis() as u64 {
                // ... process.submit(ProcessOperation::Kill);
            }
        }
    }
}
```

Seeding and refresh:

- Spawn seeds it non-zero — `binaries/daemon/src/spawn/spawner.rs:330`:
  ```rust
  let last_activity = Arc::new(AtomicU64::new(crate::node_communication::current_millis()));
  ```
- Restart re-seeds it non-zero — `binaries/daemon/src/spawn/prepared.rs:446`.
- It is only refreshed once the node sends a `DaemonRequest` (i.e. after it registers/subscribes) — `binaries/daemon/src/node_communication/mod.rs:222` (`handle_message` → `self.last_activity.store(current_millis(), ...)`).

So between spawn and the node's first daemon message, `last_activity` holds the spawn timestamp and `check_node_health` (invoked periodically from the health-check interval, `lib.rs:1570`) measures elapsed time from spawn.

## The asymmetry (why this looks like an oversight, not a design choice)

The finish-straggler watchdog reads the *same* `last_activity` and was explicitly given a `connected` gate to avoid killing a slow-starting node. `binaries/daemon/src/running_dataflow.rs:756-764`:

```rust
// A node is only a silence candidate once it has subscribed
// — i.e. has connected ... Until then it is still starting (a slow
// model-load node), not a wedge, even though `last_activity` is seeded
// at spawn time and would otherwise read as long-silent. ...
connected: self.connected_nodes.contains(id),
```

`connected_nodes` is exactly the signal `check_node_health` is missing. (See also the comments at `running_dataflow.rs:815` and `:1142` calling out the seed-at-spawn hazard.)

## Concrete trigger

1. A node sets `health_check_timeout: 5.0` and has a heavy cold start — Python imports + model-weight load that takes, say, ~10s before it opens its dora node connection and sends its first `DaemonRequest`.
2. The daemon's health-check interval fires (~every 5s, `lib.rs:1453`). At ~5s post-spawn, `elapsed_ms > 5000` while `last` still equals the spawn seed (the node hasn't communicated yet).
3. `check_node_health` SIGKILLs the node mid-startup and increments `ft_stats.health_check_kills`.
4. Per the field doc ("it is killed and the restart policy is evaluated", `descriptor.rs:788`), under `restart_policy: Always`/`OnFailure` this becomes a restart→kill→restart loop the node can never escape, because every incarnation is killed before it finishes loading.

## Suggested direction

Gate the kill on the node having connected, mirroring the straggler watchdog — e.g. `if !dataflow.connected_nodes.contains(node_id) { continue; }` before the timeout comparison (and drop the now-redundant `last == 0` check, or keep it as belt-and-suspenders). That restricts `health_check_timeout` to *post-connection* liveness, which matches the "monitors this node for activity" wording and the sibling watchdog's behavior.

If instead the intended semantic is deliberately "must connect within `health_check_timeout` of spawn," then the `last == 0` dead-code guard should be removed and the field docs updated to make the from-spawn window explicit — but the straggler-watchdog precedent suggests post-connect was the intent.

phil-opp:
Fixes #2937.

## Problem

`Daemon::check_node_health` kills any node whose `last_activity` is older than
its configured `health_check_timeout`. Its only startup guard was
`if last == 0 { continue }` — but `last_activity` is seeded to the spawn
timestamp (`current_millis()`), never `0`, so that guard was dead code. There
was no `connected` check, so the timeout clock started ticking at **spawn**,
before the node had any chance to communicate. A node whose legitimate cold
start (e.g. Python imports + model-weight load) exceeds `health_check_timeout`
was SIGKILLed mid-startup — and under `restart_policy: Always`/`OnFailure` this
became an unescapable restart→kill→restart loop.

This was asymmetric with the sibling finish-straggler watchdog
(`running_dataflow.rs`), which reads the same `last_activity` and was
explicitly given a `connected` gate for exactly this seed-at-spawn trap.

## Fix

Gate the kill on the node having connected (`dataflow.connected_nodes`),
mirroring the straggler watchdog, and drop the dead `last == 0` check. The
decision is extracted into a pure `health_check_should_kill(connected,
last_activity, now, timeout)` helper so it can be unit-tested without a full
`Daemon`. `health_check_timeout` now covers **post-connection** liveness, which
matches the "monitors this node for activity" field wording.

## Tests

Added `health_check_tests` covering: a not-yet-connected node is never killed
even when long-silent past the timeout; a connected+silent node past the
timeout is killed; a connected+recently-active node is not; the boundary
(exactly-at-timeout) is not killed; and clock skew does not underflow.

All 5 pass; `cargo clippy -p dora-daemon -- -D warnings` is clean.

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2699,22 +2699,28 @@ impl Daemon {
                     continue;
                 };
                 let last = node.last_activity.load(atomic::Ordering::Acquire);
-                if last == 0 {
-                    continue; // not yet connected
+                // The health-check watchdog only monitors *post-connection*
+                // liveness. `last_activity` is seeded to the spawn timestamp
+                // (not 0), so a node that has not yet sent its first
+                // `DaemonRequest` would be measured as silent from spawn and
+                // SIGKILLed mid-startup — e.g. a node whose cold start (Python
+                // imports + model-weight load) legitimately exceeds
+                // `health_check_timeout`. Gate the kill on the node having
+                // connected, mirroring the finish-straggler watchdog (#2937).
+                let connected = dataflow.connected_nodes.contains(node_id);
+                if !health_check_should_kill(connected, last, now_millis, timeout) {
+                    continue;
                 }
                 let elapsed_ms = now_millis.saturating_sub(last);
-                let timeout_ms = timeout.as_millis() as u64;
-                if elapsed_ms > timeout_ms {
-                    tracing::warn!(
-                        "node `{node_id}` unresponsive for {}ms (timeout: {timeout:?}), killing",
-                        elapsed_ms,
-                    );
-                    self.ft_stats
-                        .health_check_kills
-                        .fetch_add(1, atomic::Ordering::Relaxed);
-                    if let Some(process) = &node.process {
-                        process.submit(ProcessOperation::Kill);
-                    }
+                tracing::warn!(
+                    "node `{node_id}` unresponsive for {}ms (timeout: {timeout:?}), killing",
+                    elapsed_ms,
+                );
+                self.ft_stats
+                    .health_check_kills
+                    .fetch_add(1, atomic::Ordering::Relaxed);
+                if let Some(process) = &node.process {
+                    process.submit(ProcessOperation::Kill);
                 }
             }
         }
@@ -5859,6 +5865,31 @@ fn is_sigterm_like_exit(exit_status: &NodeExitStatus) -> bool {
     )
 }
 
+/// Decide whether the health-check watchdog should kill a node.
+///
+/// Returns `true` only once the node has **connected** at least once (sent its
+/// first `DaemonRequest`) and has then been silent for longer than `timeout`.
+///
+/// A node that has not yet connected is still starting up and must never be
+/// killed by this watchdog: `last_activity` is seeded to the spawn timestamp
+/// (see `spawner.rs` / `prepared.rs`), so without the `connected` gate the
+/// timeout clock would start ticking at spawn — before the node has had any
+/// chance to communicate — and a node whose legitimate cold start exceeds
+/// `health_check_timeout` would be SIGKILLed mid-startup, mirroring the trap
+/// the finish-straggler watchdog was explicitly fixed for (#2937).
+fn health_check_should_kill(
+    connected: bool,
+    last_activity_millis: u64,
+    now_millis: u64,
+    timeout: Duration,
+) -> bool {
+    if !connected {
+        return false;
+    }
+    let elapsed_ms = now_millis.saturating_sub(last_activity_millis);
+    elapsed_ms > timeout.as_millis() as u64
+}
+
 /// Grace period before the finish-straggler watchdog escalates a stuck node, or
 /// `None` if the watchdog has been explicitly **disabled**.
 ///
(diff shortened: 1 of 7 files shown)


phil-opp:

No issues found. The fix correctly replaces the dead `if last == 0` guard (dead because `last_activity` is seeded to the spawn timestamp, never 0) with a `connected_nodes` gate, so the health-check watchdog only measures post-connection silence — mirroring the finish-straggler watchdog. Extracting the decision into the pure `health_check_should_kill` helper is clean, and it uses `saturating_sub` so clock skew can't underflow.

The unit tests are meaningful and non-vacuous: they cover the not-yet-connected case (never killed even when long-silent past timeout), connected-and-silent (killed), connected-and-recent (not killed), the strict `>` boundary at exactly the timeout, and the clock-skew case. The `>` (not `>=`) comparison is preserved from the original.

---

phil-opp:
> [!NOTE]

One behavior observation not covered by the earlier automated review (diff unchanged since; no new commits):

- **Pre-connection hangs now escape every watchdog.** After this PR, both liveness watchdogs gate on `connected`: `check_node_health` (this diff) and `finish_stragglers` (`running_dataflow.rs:899`). A node only enters `connected_nodes` when it calls `subscribe` (`lib.rs:4480`). A node that hangs/deadlocks *before* ever subscribing (e.g. a bug in import/init code that never reaches `Node::init`) therefore stays `connected == false` forever, and `health_check_should_kill` returns `false` unconditionally — so it is never reaped. I could not find any separate node-connection/init timeout on the daemon side that would catch it (`REGISTER_TIMEOUT` in `coordinator.rs:21` is daemon↔coordinator, not node↔daemon).
- This is the intended tradeoff and is symmetric with the #2270 straggler gate, and post-connection hangs are still detected correctly (`last_activity` stops advancing → kill). But it is a genuine reduction vs. prior behavior: when `health_check_timeout` was set, the old spawn-seeded clock *did* reap a node stuck in startup. Users who set `health_check_timeout` specifically to bound total startup will silently lose that. Worth a one-line note in the field docs that it covers **post-connection** liveness only, and (optionally) a follow-up for a distinct startup/connection deadline, since slow-legit-startup and hung-startup are otherwise indistinguishable pre-connection.

Helper and tests are otherwise sound: the unit tests drive the pure `health_check_should_kill` directly with `connected` true/false, are non-vacuous, and use no sleeps (not timing-flaky). The strict `>` boundary and `saturating_sub` skew handling are preserved and tested.

---

phil-opp:

Re-review after the docs commit — no new issues.

The added doc comments address the earlier suggestion to document the post-connection-liveness semantics; the code is unchanged and still correct (the watchdog gate arms exactly at first subscribe via `connected_nodes`, and the `health_check_should_kill` unit tests are sleep-free and non-vacuous). For maintainers merging this: the intended behavior change stands — a node that hangs *before* it ever connects now escapes this watchdog, which is symmetric with the #2270 straggler gate and is now documented.

---

phil-opp:
The merge-queue failure here is not attributable to this PR — no code change is needed.

- All of this PR's own required checks are green: Format, Clippy, Check, Typos, Audit, License, Unwrap-budget, find_package smoke.
- Every `ci.yml` run on this PR's merge-queue batch (`trunk-merge/pr-3009/*`) has conclusion `cancelled` (Trunk re-batching) — none failed. The genuine test failures were on batches of unrelated PRs it was grouped behind (#2970, #3018, #2959, #2975).
- `Trunk Merge Queue (main)` is currently back to `in_progress`, i.e. re-testing this PR after the batch re-formed.

Leaving the diff as-is; watching for the next queue outcome.

---

PR #2940, opened 2026-08-01
fix(memory-pool): unregister host with host_base, not device VA, in broadcast-free cleanup (#2934)

### LINKED ISSUE #2934: memory-pool: broadcast-free receiver cleanup calls `_unregister_host` with the device VA instead of the host pointer
> **This issue was created by a scheduled, automated Claude code-review check** that looks at recently merged PRs. The finding was verified against the source, but please double-check before acting on it.

## Summary

The cross-process cleanup path added in **PR #2619** (`feat(memory-pool): GPU-to-GPU adaptive transport…`, commit `e99ba21`) calls `cudaHostUnregister` with the **device virtual address** rather than the **original host pointer** that was registered. `cudaHostUnregister` then fails, the host registration is never removed, and the shmem is `munmap`'d while still pinned in the CUDA driver — leaking the pin and leaving a dangling pinned registration over an address range that can be reused.

## Location / affected code path

`apis/python/node/src/lib.rs`, in `process_pending_memory_pool_frees` (the handler for the daemon-broadcast `NodeEvent::FreeMemoryPool`):

```rust
} else if slot.gpu_va != 0 {
        let bound = helpers.bind(py);
        let _ = bound.call_method1("_unregister_host", (slot.gpu_va,));  // BUG
    }
}
// Drop slot → munmap
```

`RecvGpuSlot` documents the two fields explicitly:

```rust
gpu_va: u64,     // device VA from cudaHostGetDevicePointer, 0 if IPC path
host_base: u64,  // original host ptr passed to cudaHostRegister
```

The **sibling** cleanup in `free_memory_pool` (same file, ~line 3290) does it correctly and even carries a warning comment describing exactly this pitfall:

```rust
} else if slot.gpu_va != 0 {
    // Host-registered mapping (effective_as_cuda branch):
    // must cudaHostUnregister before munmap.  _unregister_host
    // requires the original host pointer (shmem base), not the
    // device VA returned by cudaHostGetDevicePointer.
    let _ = bound.call_method1("_unregister_host", (slot.host_base,));
}
```

The broadcast-free path (added in the same PR) diverged from this and passes `slot.gpu_va`.

## Risk / failure scenario

1. A sender registers a CUDA pool via the pinned-host / device-VA path (`ipc_present != 1`, non-CPU source → the `else if effective_as_cuda` branch).
2. A receiver reads it, creating a `RecvGpuSlot` with `gpu_va != 0` and `gpu_buf == 0`.
3. Any node calls `free_memory_pool`; the daemon broadcasts `FreeMemoryPool` to the receiver (it is in `touched_by` because it read the pool).
4. The receiver runs the buggy branch: `_unregister_host(gpu_va)` fails with `cudaErrorHostMemoryNotRegistered`, the error is swallowed by `let _ = …`, and the slot then drops, `munmap`ping memory that is still registered.

The result is a pinned-memory leak plus a dangling pinned registration. On UVA systems where `gpu_va == host_base` this is coincidentally masked, so the failure is **intermittent and hardware-dependent**, which makes it easy to miss in testing (and the two new GPU smoke tests are `#[ignore]`-gated, so nothing exercises this in CI).

## Suggested fix

Pass `slot.host_base` (mirroring the correct sibling at ~line 3290):

```rust
let _ = bound.call_method1("_unregister_host", (slot.host_base,));
```

## Related (same PR, minor / latent)

In `register_memory_pool` (~line 2077), the DtoH-failure cleanup gates the host-unregister on `!is_pinned`, while the corresponding `_register_host` call (~line 2016) is **unconditional**. It is currently harmless only because `should_pin(is_cuda=true, …)` is always `false` for a CUDA source, so `!is_pinned` is always true here. If `should_pin` is ever tuned to pin CUDA sources, a large-tensor DtoH failure would skip the unregister and leak a pin. Consider making the unregister unconditional to match the register call.

phil-opp:
Fixes #2934.

## Problem

The cross-process cleanup path added in #2619 (`process_pending_memory_pool_frees`,
the handler for the daemon-broadcast `NodeEvent::FreeMemoryPool`) called
`cudaHostUnregister` (`_unregister_host`) with the **device virtual address**
(`slot.gpu_va`) instead of the **original host pointer** (`slot.host_base`) that
was registered with `cudaHostRegister`.

`cudaHostUnregister` then fails with `cudaErrorHostMemoryNotRegistered`, the
error is swallowed by `let _ = …`, and the slot drops — `munmap`ping memory
that is still pinned in the CUDA driver. The result is a pinned-memory leak
plus a dangling pinned registration over an address range that can be reused.
On UVA systems where `gpu_va == host_base` this is coincidentally masked, so the
failure is intermittent and hardware-dependent.

The sibling cleanup in `free_memory_pool` does this correctly (passes
`slot.host_base`) and even carries a comment describing exactly this pitfall;
the broadcast-free path diverged.

## Fix

- Pass `slot.host_base` in the broadcast-free path, mirroring the correct
  sibling, with the same explanatory comment.
- Related latent fix flagged in the report: in `register_memory_pool`, the
  DtoH-failure cleanup gated `_unregister_host` on `!is_pinned` while the
  corresponding `_register_host` call is unconditional. The unregister is now
  unconditional to match, so a future change to `should_pin` (to pin CUDA
  sources) cannot leak a pin on a large-tensor DtoH failure.

## Testing

The GPU memory-pool paths are `#[ignore]`-gated (require CUDA hardware), so this
is a targeted correctness fix verified against the working sibling path; no new
automated test exercises the CUDA branch in CI. `cargo check -p
dora-node-api-python` compiles cleanly (the pre-existing warnings are unrelated
to this change).

---

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -1287,9 +1287,15 @@ impl Node {
                             let _ = bound.call_method1("_ipc_close", (slot.gpu_buf,));
                         }
                     } else if slot.gpu_va != 0 {
+                        // Host-registered mapping (effective_as_cuda branch):
+                        // must cudaHostUnregister before munmap.  _unregister_host
+                        // requires the original host pointer (shmem base), not the
+                        // device VA returned by cudaHostGetDevicePointer — passing
+                        // the device VA makes cudaHostUnregister fail and leaks the
+                        // pin over an address range that then gets munmap'd.
                         if let Ok(helpers) = get_cuda_helpers(py) {
                             let bound = helpers.bind(py);
-                            let _ = bound.call_method1("_unregister_host", (slot.gpu_va,));
+                            let _ = bound.call_method1("_unregister_host", (slot.host_base,));
                         }
                     }
                     // Drop slot → munmap
@@ -2074,11 +2080,13 @@ impl Node {
                     dtoh_copy_ok = false;
                 }
                 if !dtoh_copy_ok {
-                    if !is_pinned {
-                        if let Ok(helpers) = get_cuda_helpers(py) {
-                            let bound = helpers.bind(py);
-                            let _ = bound.call_method1("_unregister_host", (shmem_ptr as u64,));
-                        }
+                    // The matching `_register_host` above is unconditional (it
+                    // runs whenever `!receiver_is_cuda`), so the unregister must
+                    // be too — gating it on `!is_pinned` would leak the pin if
+                    // `should_pin` is ever tuned to pin CUDA sources.
+                    if let Ok(helpers) = get_cuda_helpers(py) {
+                        let bound = helpers.bind(py);
+                        let _ = bound.call_method1("_unregister_host", (shmem_ptr as u64,));
                     }
                     shmem.set_owner(true);
                     eyre::bail!(


phil-opp:

No issues found. I verified both hunks against the source:

- **Primary fix (line ~1289):** the slot struct documents `host_base` as "original host ptr passed to cudaHostRegister" and `gpu_va` as the device VA from `cudaHostGetDevicePointer`. `cudaHostUnregister` requires the registered host pointer, so passing `slot.host_base` is correct. This now matches the already-correct sibling cleanup in `free_memory_pool` (which carries the identical comment and already uses `host_base`); the broadcast-free path was the one that diverged.
- **Secondary fix (DtoH-failure cleanup):** `_register_host` runs unconditionally for `shmem_ptr` whenever `!receiver_is_cuda`, so gating the matching `_unregister_host(shmem_ptr)` on the source tensor's `!is_pinned` was a mismatch. Making the unregister unconditional is the right symmetry.

CUDA paths are hardware-gated so this isn't covered by CI, but the fix is a straightforward alignment with the working sibling path.

---

PR #2941, opened 2026-08-01
fix(cli): emit `dora node list --format json` as a single JSON array (#2922)

### LINKED ISSUE #2922: Docs/UX: `dora node list --format json` emits JSON Lines, not a JSON array
**Versions:** dora-cli 1.0.0-rc.4 (git rev `7eb4a5f8b`).

`dora node list -d <df> --format json` prints one JSON object per line:

```
{"node":"budget-guard","status":"Running","pid":"54353",...}
{"node":"dora-genesis","status":"Running","pid":"54375",...}
```

Callers expecting `--format json` to mean "the output parses as one JSON document"
(a common convention) fail with a single `json.loads`. Ask: either document the JSONL
shape where the flag is described, or add `--format json-array` (or make `json` an array
and `jsonl` the line-oriented form). No strong preference — just make the contract
explicit; we now parse line-wise.

phil-opp:
Fixes #2922.

## Problem

`dora node list --format json` printed one JSON object per line (JSON Lines),
not a single JSON document. Callers that reasonably expect `--format json` to
mean "the output parses as one JSON document" (a common convention) fail with a
single `json.loads`.

## Fix

Emit the entries as a single pretty-printed JSON **array**, matching the sibling
`dora node info --format json` (which already serializes its whole output as one
document via `serde_json::to_string_pretty`). An empty result now prints `[]`
rather than nothing, so the output always parses. The command help is updated to
state the array contract explicitly.

## Tests

Added unit tests asserting the JSON output parses as a single JSON array (and
round-trips the entries), and that an empty result serializes to `[]`.

Both pass; `cargo clippy -p dora-cli -- -D warnings` and `cargo fmt` are clean.

---

### DIFF
--- a/binaries/cli/src/command/node/list.rs
+++ b/binaries/cli/src/command/node/list.rs
@@ -23,7 +23,7 @@ use dora_message::{cli_to_coordinator::ControlRequest, coordinator_to_cli::NodeI
 /// List nodes in a specific dataflow:
 ///   dora node list --dataflow my-dataflow
 ///
-/// List nodes as JSON:
+/// List nodes as a JSON array (parses as a single JSON document):
 ///   dora node list --format json
 #[derive(Debug, Args)]
 #[clap(verbatim_doc_comment)]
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The change correctly replaces the JSON Lines output with a single pretty-printed array via `serde_json::to_string_pretty(&entries)`, matching the sibling `dora node info --format json`, and an empty result now yields `[]`. The two added tests assert the real contract (parses as one array, round-trips entries, empty → `[]`) rather than being tautological.

One minor note (not blocking): the tests exercise `serde_json::to_string_pretty` directly rather than the `list()` function's output path, so they pin the serialization contract but not the `println!` call site itself. Given `list()` requires a live coordinator connection, that's a reasonable tradeoff.

---

phil-opp:
I found the following issue.

**The output format change breaks an existing consumer in the test suite: `tests/node-lifecycle-e2e.rs::parse_node_list`.**

That helper parses `dora node list --format json` **line by line**:

```rust
// tests/node-lifecycle-e2e.rs:193
fn parse_node_list(stdout: &str) -> BTreeMap<String, (String, String, String)> {
    let mut out = BTreeMap::new();
    for line in stdout.lines() {
        // ...
        let v: serde_json::Value = match serde_json::from_str(line) {
            Ok(v) => v,
            Err(_) => continue,
        };
        // ... reads v["node"], v["status"], ...
    }
    out
}
```

It worked with the old JSON-Lines output (each line was one complete `to_string(&entry)` object). With the new `to_string_pretty(&entries)` array, **no individual line is a parseable JSON object** — each line is `[`, `{`, `  "node": "a",`, `}`, `]`, etc. — so every `from_str(line)` hits the `Err(_) => continue` arm and `parse_node_list` **always returns an empty map**.

Consequence: every `wait_for_list(...)` predicate that keys off node status (e.g. `start_lifecycle` at line ~325, and the many callers throughout the file) can never be satisfied, so each spins its full timeout (up to 30s) and then fails with a "did not reach Running" assertion. Effectively the whole `node-lifecycle-e2e.rs` suite regresses.

This binary runs in the nightly `cli-tests` matrix rather than the PR-blocking gate, so green CI on this PR does not clear it.

Suggested fix: update `parse_node_list` to parse the whole document once, e.g.

```rust
let entries: Vec<serde_json::Value> = serde_json::from_str(stdout).unwrap_or_default();
for v in entries { /* read v["node"], ... */ }
```

The two added unit tests exercise `to_string_pretty` directly, which is why they don't catch this — the break is only visible through the command's real stdout consumer.

---
_This is a fully automated review by Claude. No human has reviewed or verified these comments._

---

phil-opp:
Superseded by #2922

phil-opp:
Superseded: #2922 was resolved by documenting the JSON Lines shape (`e8d23aab6`, `b2c8c7f34`) rather than switching to an array, so this change would now contradict the documented contract. Closing — reopen if the array shape is worth revisiting before 1.0 freezes the CLI surface.

Issue #2918, opened 2026-07-31
Bug/Feature: dynamically added nodes lose the Python environment — no `--uv` inheritance, and interpreter symlink resolution bypasses the venv


heyong4725:
**Versions:** dora-cli 1.0.0-rc.4; daemon from git rev `7eb4a5f8b`. macOS arm64, uv-managed venvs.

## What happens

Two related problems for `dora node add` on a dataflow started with `dora start ... --uv`:

1. The dynamically added node is spawned WITHOUT the dataflow's `--uv` wrapping, so a
   plain `path: my_node.py` runs under whatever bare python the daemon finds — imports
   like pyarrow/dora fail (`ExitCode(1)` before register).
2. Pointing `path:` at a venv interpreter does not help: the daemon appears to resolve
   the interpreter SYMLINK before exec, so argv[0] is the uv base interpreter and
   `pyvenv.cfg` discovery never happens. Observed with the node yaml literally containing
   the venv path:

```
path: /.../wt-h4/.venv/bin/python
args: /.../trace_recorder.py
```
```
ModuleNotFoundError: No module named 'numpy'
```

while `/.../wt-h4/.venv/bin/python -c "import numpy"` succeeds in a shell.

**Workaround we ship:** set `env: {PYTHONPATH: <venv site-packages>}` on the dynamic
node — packages then resolve regardless of how the interpreter was exec'd.

## Asks

- `dora node add --uv` (or better: inherit the dataflow's exec/build mode by default) so
  dynamic nodes run in the same environment as their static peers.
- Don't canonicalize the node executable path before exec (or exec via the given path so
  venv discovery works).

heyong4725:
Thanks — both observations reproduced from this report, and one of them ran deeper than described.

## Ask 2 (symlink canonicalization) — confirmed, fix up at #2942

`resolve_path` (`libraries/core/src/descriptor/mod.rs:400`) called `canonicalize()` on the node path before exec. A venv's `bin/python` is a symlink, and CPython locates `pyvenv.cfg` relative to the path it was *invoked* as — that indirection is the entire venv mechanism. Canonicalizing execs the base interpreter with the venv gone. Directly observable with the same interpreter binary:

```
$ .venv/bin/python -c "import sys; print(sys.prefix)"       → /…/.venv
$ <canonicalized target> -c "import sys; print(sys.prefix)" → <base prefix>
```

Two notes beyond the report:

[...]

PR #2942, opened 2026-08-01
fix(core): don't resolve symlinks in node paths, so venv interpreters keep their venv

heyong4725:
## What

`resolve_path` canonicalized the node executable path before exec. A virtualenv's `bin/python` is a symlink, and CPython finds `pyvenv.cfg` relative to the path it was *invoked* as — that indirection is the entire venv mechanism. Canonicalizing resolved the symlink and ran the base interpreter with no venv, so `path: .venv/bin/python` silently lost its environment:

```
$ .venv/bin/python -c "import numpy"          # works in a shell
# under dora:  ModuleNotFoundError: No module named 'numpy'
```

Addresses the second ask in #2918 (no closing keyword — the first ask is discussed below and the reporter should confirm on their pipeline before closing). Not specific to dynamic adds: every spawn goes through `resolve_path`, so a *static* node pointing at a venv interpreter was equally broken.

## The fix

`std::path::absolute` instead of `canonicalize()`: prepends the cwd, strips `.`, and leaves symlinks and `..` for the kernel to resolve at exec time — how a shell would invoke the same path. The existence check gating the fall-through to uv/`$PATH` resolution is preserved.

Deliberately untouched: `resolve_path_confined` (hub nodes), where canonicalization is a security measure against symlink escapes; and the post-download `canonicalize()` calls, where no venv symlink semantics exist.

## On the issue's first ask (`--uv` inheritance)

Traced and partially disproved: the coordinator already forwards the dataflow's `--uv` flag on `AddNode` (`uv: dataflow.uv`), and did at the reporter's rev. What remains true is that a build-less dynamic `.py` node gets `uv run python` resolved from the dataflow working directory, which is not necessarily the environment its static peers use — there is no single "dataflow env" to inherit, since managed envs are per-node (keyed by `build:`). The reliable way to hand a dynamic node a specific environment is exactly `path: <venv>/bin/python` — which this PR makes work.

## Test plan

Both tests verified in both directions by stashing only the `dora-core` change:

| Test | Without fix | With fix |
|---|---|---|
| `resolve_path_preserves_symlinks` (unit, relative + absolute shapes) | FAILED | ok |
| `venv_interpreter_path_runs_inside_the_venv` (e2e, real daemon spawn) | `sys.prefix is "/opt/anaconda3/envs/aisle"` | ok |

The e2e creates a scratch venv with `python3 -m venv --without-pip` (no network, no pip), adds a node whose `path:` is the venv symlink with the script in `args:` — the exact shape from the issue — and asserts the marker file the probe writes contains the venv's `sys.prefix`. The probe is deliberately not a dora node (that would need dora-rs installed into the venv, hitting the #1710 PyPI/workspace drift); post-#2933 a pre-register exit is harmless to the dataflow.

- [x] `cargo test -p dora-core` — 264 passed
- [x] `node-lifecycle-e2e --skip lifecycle_python` — 8/9 (`lifecycle_cxx` at its cmake step, pre-existing macOS linker issue)
- [x] `fault-tolerance-e2e` — 11/11 (spawn-heavy, exercises the changed path on every node)
- [x] fmt + clippy clean

## Behavior note

Spawn log lines may now show un-normalized paths (`examples/foo/../../target/debug/bin`) where they previously showed the canonical form. Exec behavior is unchanged.

### DIFF
--- a/binaries/cli/src/command/start/attach.rs
+++ b/binaries/cli/src/command/start/attach.rs
@@ -51,6 +51,15 @@ pub fn attach_dataflow(
                             .wrap_err_with(|| {
                                 format!("failed to resolve node source `{}`", python_source.source)
                             })?;
+                        // Canonicalize for the WATCH MAP only: `notify`
+                        // backends (FSEvents in particular) report real
+                        // paths, and the lookup below is a plain `==` on
+                        // the event path. Since #2918 `resolve_path` no
+                        // longer canonicalizes (exec must preserve venv
+                        // symlinks), so a symlinked or `..`-containing
+                        // source would silently never match its own
+                        // modification events without this.
+                        let path = path.canonicalize().unwrap_or(path);
                         node_path_lookup
                             .insert(path, (dataflow_id, node.id.clone(), Some(op.id.clone())));
                     }
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -396,9 +396,17 @@ pub fn resolve_path(source: &str, working_dir: &Path) -> Result<PathBuf> {
         path.to_owned()
     };
 
-    // Search path within current working directory
-    if let Ok(abs_path) = working_dir.join(&path).canonicalize() {
-        Ok(abs_path)
+    // Search path within current working directory. Deliberately NOT
+    // `canonicalize()`: that follows symlinks, and a virtualenv's
+    // `bin/python` is a symlink whose *location* is what CPython uses to
+    // discover `pyvenv.cfg`. Resolving it before exec runs the base
+    // interpreter with no venv, so imports that work in a shell fail
+    // under dora (dora-rs/dora#2918). `path::absolute` only prepends the
+    // cwd and drops `.` components; symlinks and `..` are left for the
+    // kernel to resolve at exec time, which matches shell behavior.
+    let joined = working_dir.join(&path);
+    if joined.exists() {
+        std::path::absolute(&joined).map_err(Into::into)
     // Otherwise resolve against the `uv`-managed environment first (when `uv`
     // is available), then fall back to the system `$PATH`.
     } else if which::which("uv").is_ok() {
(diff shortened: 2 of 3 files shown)


phil-opp:
**Automated review** — I looked at the diff as an outside contributor. No blocking issues found.

The fix is correct: swapping `canonicalize()` for `exists()` + `std::path::absolute()` in the non-confined `resolve_path` preserves the symlink of a venv `bin/python`, so CPython's `pyvenv.cfg` discovery (keyed off argv[0]) still works, while `..` and symlink resolution are left to the kernel at exec time. The fall-through to uv/`$PATH` is preserved: `exists()` follows symlinks and returns false for a broken link, matching the old canonicalize-failure behavior. `resolve_path_confined` correctly keeps its canonicalization as a symlink-escape guard, so the security model is untouched. Both the unit test (relative + absolute shapes) and the e2e test (real venv, asserts `sys.prefix`) exercise the actual regression.

One minor, non-blocking note: `venv_interpreter_path_runs_inside_the_venv` calls `python3 -m venv` with `.expect(...)`, so on a runner without `python3` the test panics rather than skipping. It's a `#[cfg(unix)]` e2e test outside the PR-CI gate, so impact is low, but a `which python3` guard that skips would make it more robust across environments.

---
🤖 _This is a fully automated review by Claude ([Claude Code](https://claude.ai/code)) — no human was in the loop. Treat it as one more set of eyes, not a maintainer sign-off._

---

heyong4725:
Review pass (adversarial + testing + maintainability specialists) done; fixes pushed as `99614044`. Two findings were real and material — both verified in the code before acting.

**The fix was incomplete: `resolve_path_via_uv` still canonicalized.** Same venv bypass through the fallback door — a `path: python` that misses the working dir but resolves via `uv run which` to a venv interpreter got symlink-resolved to the base interpreter. Two reviewers found this independently (one also caught that the function's doc claimed "canonicalization parity with the other branches", which this PR had just inverted). Both fall-through branches now share one `absolutize_preserving_symlinks` helper (existence check + `path::absolute`), so the contract can't regress in one branch while the tests exercise another. Verified separately: plain `which::which` (the no-uv fallback) does **not** canonicalize, so that branch was already fine.

**A silent hot-reload regression in `attach.rs`.** The watch map keys on `resolve_path` output and does a plain `==` against `notify` event paths, which backends (FSEvents notably) report as real paths. Canonical keys used to match; post-fix, a symlinked or `..`-containing Python operator source would silently never match its own modification events. The watch key — a comparison key, not an exec path — is now canonicalized explicitly, with a comment on why it deliberately differs from the exec path.

**Coverage added from the review:** the `exists()` branch boundary (dangling venv symlink falls through and errors — a switch to `symlink_metadata()` would exec a guaranteed-ENOENT path unnoticed), and `..` preservation in the unit test (kernel-side `..` resolution through symlinked dirs differs from lexical collapse — the exact semantic the code comment promises). Plus e2e hygiene: stale pid-suffixed scratch venvs are reclaimed at test start, and the marker deadline went 20s→30s for cold runners.

The testing specialist independently mutation-verified both original tests (reverted the fix → both fail with the expected diagnostics), and confirmed the CI fit: the e2e needs only bare `python3` (`--without-pip` sidesteps the Debian ensurepip gap) and its name keeps it in the uv-less e2e job.

**Declined, with reasons:**
- Extending `write_node_yml` to express this test's `args:`/`env:` shape — one caller, and the hand-written yaml deliberately mirrors the issue report line for line.
- A `cfg(windows)` dotted-stem case — Windows venvs copy `python.exe` rather than symlinking, and the `EXE_EXTENSION` behavior (`python3.12` never getting `.exe` appended) is pre-existing and untouched by this diff. Verified empirically that `with_extension` does not mangle dotted paths on the changed branch.
- Canonicalizing hub-confined venv-interpreter entrypoints — confinement canonicalizes as a security measure (symlink-escape rejection); hub Python nodes get their interpreter from the managed env, not argv[0]. Noted, not changed.

Sweep: `dora-core` 265 unit tests, venv e2e green post-refactor, fmt/clippy clean.

heyong4725 (commented):
Reviewed current head `99614044` against `main`: no blocking findings.

The non-confined resolver now preserves the executable spelling for both working-directory and successful `uv run which` resolution, while confined hub entrypoints and downloaded artifacts retain their canonicalization boundaries. The explicit canonicalization in the hot-reload watch map also preserves the prior `notify` comparison behavior without changing the exec path.

Local validation on this exact head:
- `cargo test -p dora-core resolve_path` — 6 passed
- `cargo test --test node-lifecycle-e2e venv_interpreter_path_runs_inside_the_venv -- --exact --nocapture` — 1 passed; the spawned probe reported the scratch venv as `sys.prefix`

Scope note: this fixes the interpreter-symlink ask in #2918. The separate dynamic-node `--uv` inheritance ask remains unresolved, so #2918 should stay open unless that remainder is split into another issue.

Issue #2919, opened 2026-07-31
Feature: dataflow-scoped env injection for `dora start` (CLI env does not reach daemon-spawned nodes)


heyong4725:
**Versions:** dora-cli 1.0.0-rc.4; daemon from git rev `7eb4a5f8b`.

## Problem

With `dora run`, nodes are children of the CLI process and inherit its environment, so
`MYVAR=x dora run graph.yaml` configures nodes. With `dora start`, nodes are spawned by
the daemon and inherit the DAEMON's environment — the CLI invocation's env silently does
not apply. The only channel is a per-node `env:` block in the graph YAML, which forces
callers to rewrite/copy graphs just to parameterize a run (seeds, output paths, etc.),
and makes `dora run` -> `dora start` migration subtly breaking.

## Ask

`dora start graph.yaml --env KEY=VAL [--env ...]` applying to every node of that dataflow
(node-level `env:` still wins). Alternatively/additionally: document loudly that CLI env
does not propagate under `start`.

heyong4725:
Implemented and landed on `main` as `397bfd225` (#2943).

## What you get

```
dora start graph.yaml --env SEED=42 --env OUT_DIR=/tmp/run7
dora run   graph.yaml --env SEED=42
```

Repeatable, applies to every node of the dataflow, and the same flag exists on both commands so a `dora run` → `dora start` migration keeps working. Precedence, most specific wins:

```
node env:  >  --env  >  dataflow-level env: in the YAML
```

It also reaches nodes added later with `dora node add` — worth calling out separately, because that turned out to be broken independently of this feature (details below).

## How it works, and one thing it deliberately doesn't do

`--env` merges into the dataflow-level `env:` block of the descriptor the CLI already sends, so there is no protocol or daemon change and the daemon's existing env denylist still applies. It applies at **spawn time only**: neither command re-runs `build:`, so build-time environment still comes from the YAML at `dora build` time. `dora build --env` is a separate lifecycle and isn't part of this — say so if you need it and it can be its own issue.

## One restriction worth knowing before you use it

**Values may not contain `$`.** This isn't fussiness — `EnvValue` deserializes through env-expansion, and the descriptor is deserialized *again* in the coordinator and *again* in the daemon. A `$` that reaches the wire is therefore expanded against **that host's** environment, not yours:

```
$ dora start flow.yml --env 'X=${HOME}'      # would resolve to the DAEMON's $HOME
[...]

PR #2943, opened 2026-08-01
feat(cli): --env KEY=VALUE for dora start and dora run

heyong4725:
## What

`dora start graph.yaml --env KEY=VAL [--env ...]` — and the same flag on `dora run` — setting an environment variable for every node of the dataflow without editing the YAML.

Implements #2919 (no closing keyword — leaving confirmation to the reporter). With `dora run` nodes inherit the CLI process environment, but with `dora start` they inherit the **daemon's**, so `MYVAR=x dora start …` silently configured nothing and the only channel was rewriting per-node `env:` blocks.

## Design

**Pure CLI-side change.** The dataflow-level `env:` block already exists in the descriptor with node-wins merge semantics (`merge_env` in dora-core), and `ControlRequest::Start` already ships the parsed descriptor — so `--env` merges into `descriptor.env` before send. No protocol, coordinator, or daemon changes; the daemon-side env denylist still applies to the merged result.

**Precedence** (most specific wins), matching docker/compose convention and the issue's "node-level still wins":

```
node env:   >   --env   >   dataflow-level env: in the YAML
```

**Two placement traps avoided in `dora start`** — the merge happens *after* the hub `fingerprint_source` comparison and after `invalidate_if_build_inputs_changed`, both of which hash the descriptor (env included) against `dora build`-time state. Merging earlier would reject hub dataflows as "changed since build" and invalidate the cached build id on every `--env` change. The rationale is documented at the call site and on the helper.

**`dora run` gets the identical flag** via the existing `descriptor_override` parameter of `Daemon::run_dataflow` (previously used only for hub desugaring) — identical precedence, no daemon API change. Without it, run's only mechanism (process env inheritance) sits at a *different* precedence level than start's flag, which is exactly the run→start subtle-difference trap the issue complains about.

**Spawn-time only.** Neither command re-runs `build:`, so `--env` deliberately doesn't reach build commands; the flag help on both commands says so, which also carries the issue's "document loudly" ask.

## Test plan

- [x] 9 unit tests on the parse/merge module (`KEY=VAL` shapes, value-with-`=`, empty value, missing-`=` rejected, repeated-key-last-wins, no-op leaves `env: None` untouched, override-vs-yaml, and the full precedence chain through the same `merge_env` the daemon spawn uses)
- [x] E2e in `ws-cli-e2e.rs`: one python probe pins the whole chain in a single networked `dora start --detach --env …` — the YAML sets `E2919_VAR: from-yaml` (`--env` must beat it), the node sets `E2919_NODE: node-wins` (`--env` must lose to it)
- [x] **Mutation-verified**: with `apply_env_overrides` replaced by a no-op, the e2e fails with `--env must override the dataflow-level env: block; got "from-yaml\nnode-wins\n"` — and only then
- [x] Manual smoke of both `dora run --env` and `dora start --env` against a live cluster, same precedence observed
- [x] `cargo test -p dora-cli` — 277 passed; full `ws-cli-e2e` — 26/26; fmt + clippy clean

## Out of scope, deliberately

- `dora build --env` — build-time env is a different lifecycle and interacts with the build-inputs fingerprint; follow-up if anyone asks.
- Dynamic adds: `dora node add` does not merge the stored dataflow-level env into added nodes today (pre-existing; noticed while designing the e2e). The added node's own `env:` block remains the mechanism there.

### DIFF
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -21,6 +21,7 @@ use crate::{
     session::DataflowSession,
 };
 use dora_core::build::LogLevelOrStdout;
+use dora_core::descriptor::{Descriptor, DescriptorExt};
 use dora_daemon::{Daemon, LogDestination, flume};
 use eyre::Context;
 use std::{path::PathBuf, time::Duration};
@@ -114,6 +115,9 @@ pub struct Run {
     /// under `dora start` (where nodes inherit the DAEMON's environment
     /// instead). Applies at spawn time only; `build:` commands are
     /// unaffected.
+    /// Values must survive the descriptor encoding verbatim: a literal
+    /// `$` is refused (the receiving process would expand it) and so are
+    /// numeric-looking values that would be coerced.
     #[clap(long = "env", value_name = "KEY=VALUE")]
     pub env: Vec<String>,
 }
@@ -178,6 +182,10 @@ impl Executable for Run {
 
         let dataflow_path =
             resolve_dataflow(self.dataflow.clone()).context("could not resolve dataflow")?;
+        // Validate `--env` BEFORE building: a typo should fail in
+        // milliseconds, not after a full dataflow build.
+        let env_overrides = crate::env_overrides::parse_env_overrides(&self.env)?;
+
         build_dataflow(BuildConfig {
             dataflow: dataflow_path.to_string_lossy().into_owned(),
             uv: self.uv,
@@ -192,6 +200,23 @@ impl Executable for Run {
         .context("failed to build dataflow before run")?;
         let dataflow_session = DataflowSession::read_session(&dataflow_path)
             .context("failed to read DataflowSession")?;
+
+        // `--env` merges into the dataflow-level `env:` of the descriptor
+        // handed to the in-process daemon. The hub-resolved descriptor
+        // (when present) is the mandatory base — the on-disk YAML still
+        // has unresolved `hub:` references. Without `--env`, pass the
+        // session state through unchanged.
+        let descriptor_override = if env_overrides.is_empty() {
+            dataflow_session.resolved_dataflow.clone()
+        } else {
+            let mut descriptor = match dataflow_session.resolved_dataflow.clone() {
+                Some(resolved) => resolved,
+                None => Descriptor::blocking_read(&dataflow_path)
+                    .context("failed to read dataflow descriptor for --env")?,
+            };
+            crate::env_overrides::apply_env_overrides(&mut descriptor, env_overrides);
+            Some(descriptor)
+        };
 
         let node_filters = match &self.log_filter {
             Some(filter) => parse_log_filter(filter).map_err(|e| eyre::eyre!(e))?,
@@ -226,24 +251,6 @@ impl Executable for Run {
         // `run_dataflow` takes `&Path`, so the returned future borrows
         // non-`'static`; move an owned `PathBuf` (plus the other Run
         // fields) into the spawned task so the future is `'static`.
-        // `--env` merges into the dataflow-level `env:` of the descriptor
-        // handed to the in-process daemon. The hub-resolved descriptor
-        // (when present) is the mandatory base — the on-disk YAML still
-        // has unresolved `hub:` references. Without `--env`, pass the
-        // session state through unchanged.
-        let env_overrides = crate::env_overrides::parse_env_overrides(&self.env)?;
-        let descriptor_override = if env_overrides.is_empty() {
-            dataflow_session.resolved_dataflow.clone()
-        } else {
-            let mut descriptor = match dataflow_session.resolved_dataflow.clone() {
-                Some(resolved) => resolved,
-                None => <dora_message::descriptor::Descriptor as dora_core::descriptor::DescriptorExt>::blocking_read(&dataflow_path)
-                    .context("failed to read dataflow descriptor for --env")?,
-            };
-            crate::env_overrides::apply_env_overrides(&mut descriptor, env_overrides);
-            Some(descriptor)
-        };
-
         let dataflow_path_for_daemon = dataflow_path.clone();
         let uv = self.uv;
         let stop_after = self.stop_after;
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -105,6 +105,17 @@ pub struct Run {
     /// --hub-override`; `dora run` is always local, so it applies directly.
     #[clap(long = "hub-override", value_name = "PKG=PATH")]
     pub hub_override: Vec<String>,
+    /// Set an environment variable for every node of this dataflow
+    /// (repeatable). Merges into the dataflow-level `env:` block;
+    /// node-level `env:` entries still win on conflict.
+    ///
+    /// Under `dora run` nodes also inherit this process's environment,
+    /// but `--env` is the portable spelling that behaves identically
+    /// under `dora start` (where nodes inherit the DAEMON's environment
+    /// instead). Applies at spawn time only; `build:` commands are
+    /// unaffected.
+    #[clap(long = "env", value_name = "KEY=VALUE")]
+    pub env: Vec<String>,
 }
 
 impl Run {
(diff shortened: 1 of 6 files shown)


heyong4725:
Review pass (adversarial + testing + maintainability) done; fixes pushed as `6694a8ba`. The adversarial pass found a **security hole I introduced**, and it was right to block.

## `--env` could exfiltrate host secrets

`EnvValue` is an untagged enum whose variants deserialize through `with_expand_envs`, so a value is re-interpreted **every time the descriptor is deserialized** — in the coordinator, then again in the daemon. I verified this against `dora-message` directly rather than taking the report on faith:

```
wire "${DORA_AUTH_TOKEN}"  ->  s3cr3t-daemon-token     ← expands in the RECEIVING process
wire "p$ssw0rd"            ->  ERROR: no matching variant
wire "1.10"                ->  1.1                     ← silent coercion
```

So `dora start flow.yml --env 'X=${DORA_AUTH_TOKEN}'` would have handed a node the **coordinator's or daemon's** token. `strip_denied_env` cannot stop it — that guard matches key *names*, and the key here is innocuous. This was new with the flag: YAML `env:` values are expanded CLI-side at descriptor read, so they reach the wire already substituted; `--env` was the only way to put an unexpanded `$` on it.

Escaping doesn't fix it (two deserialization hops — a `$$` surviving one is expanded on the next), so `--env` now refuses any value that doesn't survive a serialize/deserialize round-trip verbatim. One check covers all three shapes above, and the error names the key and points at the node `env:` block. Confirmed end to end:

```
$ dora start flow.yml --env 'LEAK=${DORA_AUTH_TOKEN}'
[ERROR] --env `LEAK=${DORA_AUTH_TOKEN}` cannot be represented in the dataflow descriptor…
  hint: a literal `$` cannot be carried through `--env` … Set this one in the node's `env:` block instead.
```

I chose refusal over silent delivery of a different value: rejecting `VERSION=1.10` is annoying, but handing a node `1.1` when the operator typed `1.10` is worse. Four new unit tests pin all of it, including the branch where expansion *succeeds* (the real exfiltration shape, using `PATH` as a stand-in for the secret).

## Other findings applied

- **`dora run --env` had zero coverage** — the entire `descriptor_override` branch was untested. Added an e2e, mutation-verified by passing the session descriptor instead of the merged one (fails with `got "from-yaml"`). It runs local-only, so it can't contend for port 6013.
- **`dora run` validated `--env` after a full build** — a typo cost a compile. Now parses first, like `start`.
- **Marker flush race** in the `start` e2e: the poll checks existence, but Python file objects aren't line-buffered, so it could read a created-but-unflushed file. Now write-then-`os.replace`.
- **Merge-placement premise** now pinned by a unit test asserting `--env` changes the descriptor fingerprint — the ordering constraint in `start_dataflow` was previously guarded only by a comment.
- Flag help documents the wire restrictions plus persistence in the coordinator store / visibility in `ps`.
- Maintainability: idiomatic `Descriptor::blocking_read`, proper imports, comment placement restored around the tokio spawn block, private `merge_env` no longer linked as public API, unix-only rationale on both e2es.

## Flagged, not fixed

`DORA_NODE_CONFIG`, `DORA_ZENOH_*` and `DORA_RUNTIME_CONFIG` are set **before** the user-env loop in the spawner and are absent from `ENV_DENYLIST`, so dataflow-level env can override them — `--env DORA_NODE_CONFIG=…` hands a node a forged config. Pre-existing and already reachable from YAML `env:` today, so not an escalation this PR creates, but the denylist exists as defense-in-depth for spawn integrity. Worth hardening separately rather than inside a CLI-flag PR — happy to file it.

Sweep: `dora-cli` 282 unit tests, `ws-cli-e2e` 27/27, fmt/clippy clean.

heyong4725 (commented):
Two blocking findings on `6694a8ba`:

**[P1] Reject `$` independently of the CLI environment — `binaries/cli/src/env_overrides.rs:72-76`**

`ensure_wire_safe` accepts a value whenever one local deserialize happens to reproduce the input. That is environment-dependent and does not prove the next process will see the same value. A self-referential local variable bypasses the check:

```sh
env DORA_TEST_2919_SELF='$DORA_TEST_2919_SELF' \
  dora start flow.yml --env 'LEAK=$DORA_TEST_2919_SELF'
```

I reproduced this end to end with the coordinator/daemon started under `DORA_TEST_2919_SELF=daemon-secret-2943`: the CLI accepted the flag and the spawned node received `daemon-secret-2943`. Thus the receiver-side secret-disclosure path described in this PR is still open. Since the CLI contract already says literal `$` is refused, reject `value.contains('$')` before the round-trip/coercion check and add a regression for the self-reference shape.

**[P2] Apply the stored dataflow env when resolving `AddNode` — `binaries/cli/src/command/start/mod.rs:50` / `binaries/coordinator/src/lib.rs:1588-1597`**

The flag help promises every node in the dataflow receives the value, and the override is persisted in `dataflow.descriptor.env`, but `AddNode` resolves its node through a temporary descriptor with `env: None`. Reproduced on this head: after `dora start --env GLOBAL_2943=from-cli`, a successful `dora node add` probe observed `GLOBAL_2943=<unset>`. Use the running descriptor's global env when constructing the temporary descriptor (with node-level env still winning), and cover this path in the dynamic-node E2E.

Validation performed on the exact head:
- `cargo test -p dora-cli env_overrides` — 14 passed
- `cargo test --test ws-cli-e2e env_flag_reaches -- --nocapture --test-threads=1` — 2 passed
- Separate real coordinator/daemon reproductions confirmed both findings.

Current lightweight CI checks are green; the full Test/E2E lanes are skipped by PR policy.

heyong4725:
Both findings confirmed and fixed in `91b34f22`. I reproduced each before touching code — the P1 bypass in particular is a good catch, because my guard tested the wrong property and I'd have defended it otherwise.

## P1 — the round-trip check had a fixed-point bypass

`ensure_wire_safe` asked *"does this value survive a serialize/deserialize round-trip in the CLI's process?"*. That is not a sound test of the security property: expansion runs against **whichever** environment decodes the descriptor. A CLI-local variable that refers to itself makes the CLI-side round-trip an identity:

```
CLI    (DORA_AUTH_TOKEN='${DORA_AUTH_TOKEN}')   round-trip identical  -> ACCEPTED
daemon (DORA_AUTH_TOKEN='s3cr3t-daemon-token')  decodes as            -> s3cr3t-daemon-token
```

The `$` rule is now **syntactic** — no value containing `$` reaches the wire, so no expandable value exists for any environment to resolve. Comparison-based validation can't work here in principle, since the attacker controls one side of the comparison. The round-trip check stays only for the numeric-coercion case (`1.10` → `1.1`), where no environment lookup is involved and it *is* sound.

The regression test drives a scoped child process with the self-referential variable set (`set_var` is unsafe in edition 2024 and would race sibling tests). It fails against the round-trip-only version and passes against the syntactic one — verified both ways.

## P2 — added nodes inherited no dataflow env

Root cause is one line in the coordinator: `dora node add` resolves the incoming node through a temporary single-node descriptor built with `env: None`, so `resolve_aliases_and_set_defaults` — which performs the dataflow-into-node merge — had nothing to merge. It now carries the running dataflow's `env`; node-level keys still win.

Worth noting this is **pre-existing for YAML-declared `env:`** too, not just `--env`. I'd called it out as out-of-scope in the PR description, and that was the wrong call: once `--env` is documented as applying to "every node of that dataflow", a node added later silently missing them is a correctness gap in the feature, not adjacent behavior. Thanks for pushing back on it.

New e2e covers it end to end, and pins that node-level `env:` still wins for added nodes. Mutation-verified: reverting the coordinator line fails it with `got "<unset>"`.

Sweep: `dora-cli` 284 + `dora-coordinator` 107/22 unit tests, `ws-cli-e2e` 28/28, fmt/clippy clean.

One thing I did **not** change, for the record: `DORA_NODE_CONFIG` / `DORA_ZENOH_*` / `DORA_RUNTIME_CONFIG` remain overridable via dataflow env (set before the user-env loop in the spawner, absent from `ENV_DENYLIST`). Pre-existing and reachable from YAML today, so this PR doesn't create it — but P2's fix does widen the reach slightly, since dataflow env now flows to added nodes as well. Happy to file it as daemon spawn hardening if you'd like it tracked.

PR #2945, opened 2026-08-01
perf(node): prepare the Arrow IPC fast path once per send, not twice

phil-opp:
## Summary

The per-output Arrow IPC send path computed the message layout and both flatbuffer headers **twice** per send:

- `send_output_array` called `ipc_fast_path_len` (→ `prepare`) to size the shared-memory sample, then `encode_ipc_into` (→ `prepare` **again**) to fill it.
- `send_output_raw` and the Python buffer-protocol send (`sample_handler.rs`) did the same via `uint8_ipc_len` → `uint8_layout` then `encode_uint8_ipc_header` → `uint8_layout`.

Each `prepare` / `uint8_layout` walks the array to build the `FieldNode` / `IpcBuffer` layout `Vec`s and allocates two flatbuffer headers (schema + record-batch). The first call ran to completion, had its length pulled out, was dropped, and was then recomputed identically to encode. That redundant work is independent of payload size, so it is a real fraction of send cost for high-frequency small/medium messages.

## Fix

Add prepare-once handles `PreparedIpc` and `PreparedUint8Ipc` that hold the computed layout + headers and expose:
- `byte_len()` — to size the destination sample, and
- `encode_into` / `encode_header_into` — to fill it.

All three send paths (Rust array, Rust raw, Python buffer-protocol) now prepare once and reuse the result. The one-shot `ipc_fast_path_len` / `encode_ipc_into` / `uint8_ipc_len` / `encode_uint8_ipc_header` functions are retained (used by benches and tests) and now delegate to the same shared inner encoders, so there is no duplicated encoding logic.

**No behavior change** — the emitted bytes are byte-identical to before.

## Validation

- `cargo test -p dora-node-api --lib` — 158 tests pass (incl. the `ipc_encode` byte-for-byte regression suite, which confirms identical output)
- `cargo test -p dora-node-api --test copy_count` — passes (no extra intermediate allocation)
- `cargo check -p dora-node-api-python` — clean

## Disclaimer

⚠️ This PR is **machine-generated by Claude** (an AI assistant) as part of an automated code-review pass. A human should review before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01U5MBnacdwSwDRdKFComqWU)_

### DIFF
--- a/apis/python/node/src/sample_handler.rs
+++ b/apis/python/node/src/sample_handler.rs
@@ -323,9 +323,11 @@ impl SampleHandler {
         // the data region, and expose only that region to Python — so the user
         // writes their bytes straight into the wire payload with zero copies.
         use dora_node_api::arrow_utils::ipc_encode;
-        let total = ipc_encode::uint8_ipc_len(data_length)?;
-        let mut sample = node.get_mut().allocate_data_sample(total)?;
-        let data_offset = ipc_encode::encode_uint8_ipc_header(&mut sample, data_length)?;
+        // Prepare the UInt8 IPC header once, then size and fill the sample from
+        // it — avoids rebuilding the layout + IPC headers for the length query.
+        let prepared = ipc_encode::PreparedUint8Ipc::new(data_length)?;
+        let mut sample = node.get_mut().allocate_data_sample(prepared.byte_len())?;
+        let data_offset = prepared.encode_header_into(&mut sample)?;
         Ok(Self {
             node,
             meta: Some(SampleMeta {
--- a/apis/rust/node/src/node/arrow_utils/ipc_encode.rs
+++ b/apis/rust/node/src/node/arrow_utils/ipc_encode.rs
@@ -332,8 +332,47 @@ fn prepare(array: &ArrayData) -> Option<Prepared> {
 /// `None` if `array` is not fast-path eligible (use [`encode_ipc_to_vec`]).
 ///
 /// Lets a caller size the (shared-memory) sample before encoding into it.
+///
+/// Sizing and encoding both call [`prepare`] internally, so a caller that does
+/// `ipc_fast_path_len(a)` then `encode_ipc_into(a, dst)` builds the layout and
+/// both flatbuffer headers twice. On the hot send path prefer [`PreparedIpc`],
+/// which prepares once and reuses the result for both steps.
 pub fn ipc_fast_path_len(array: &ArrayData) -> Option<usize> {
     prepare(array).map(|p| p.total)
+}
+
+/// A prepared Arrow IPC fast-path encode.
+///
+/// Holds the computed layout and both framed message headers so the byte
+/// length ([`byte_len`](Self::byte_len)) and the encode
+/// ([`encode_into`](Self::encode_into)) share the work. A separate
+/// [`ipc_fast_path_len`] + [`encode_ipc_into`] pair, by contrast, runs the
+/// whole layout + flatbuffer build twice — once to size the sample and once to
+/// fill it. On the per-output send path that redundant work is independent of
+/// payload size, so it is a real fraction of the cost for high-frequency
+/// small/medium messages.
+///
+/// The emitted bytes are identical to [`encode_ipc_into`].
+pub struct PreparedIpc(Prepared);
+
+impl PreparedIpc {
+    /// Prepare the fast-path encode for `array`, or `None` if `array` is not
+    /// fast-path eligible (fall back to [`encode_ipc_to_vec`]).
+    pub fn new(array: &ArrayData) -> Option<Self> {
+        prepare(array).map(Self)
+    }
+
+    /// Exact byte length of the IPC stream [`encode_into`](Self::encode_into)
+    /// writes. Use it to size the (shared-memory) destination sample.
+    pub fn byte_len(&self) -> usize {
+        self.0.total
+    }
+
+    /// Encode the complete IPC stream into `dst`, which must be exactly
+    /// [`byte_len`](Self::byte_len) bytes. Copies each array buffer once.
+    pub fn encode_into(&self, dst: &mut [u8]) -> eyre::Result<()> {
+        encode_prepared_into(&self.0, dst)
+    }
 }
 
 /// Frame one IPC message into `dst[at..]`: continuation marker, i32-LE metadata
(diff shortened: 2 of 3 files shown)


phil-opp:

This PR is one of a batch opened by an automated Claude code-review pass over `origin/main` (all branched from a fresh fetch, each verified independently, and the combined diff run through `fmt` + `clippy` + affected-crate tests + the `/simplify` gate). Posting the run summary here as the first PR of the batch.

### PRs opened, by category

**Optimization**
- **#2945** — `perf(node)`: prepare the Arrow IPC fast path once per send, not twice. Removes a double layout + flatbuffer-header build on the per-output send path (Rust array, Rust raw, and Python buffer-protocol sends). No behavior change; byte-identical output.

**Simplification**
- **#2946** — `refactor(node)`: dedup the copy-pasted `EventItem`→parameters match in the scheduler's `is_correlated` / `log_correlation_drop` into one `event_parameters` helper (single source of truth for which variants carry metadata).

- **#2947** — `docs(message)`: add doctests for `NodeId` / `DataId` — the previously undocumented structs — covering the `parse()`-is-fallible vs `From`/`.into()`-panics footgun and the `/`-allowed asymmetry, with a `should_panic` block that asserts the panic.
- **#2948** — `docs(node)`: add an executing doctest for `StreamSegment` demonstrating the subtle `chunk` / `next_segment` / `flush` seq/segment/`fin` state transitions.

- **Correctness sweep of `daemon` / `coordinator` / `core` / `message`** — no PR. The hot spots hunted (indexing, arithmetic overflow, `unwrap`/`expect`, char-boundary truncation, barrier/eviction logic) are already guarded, each with targeted regression tests. Two apparent gaps (`pending.rs` external-node waiter parking, the zenoh endpoint reserve→bind race) are documented known-limitations already tracked as dora#2938 / dora#2762.
- **Correctness sweep of `node` / `operator` / `runtime` / `cli` / `arrow-convert`** — no PR. Same picture: the IPC fast path, JSON/Arrow decoders, log rotation, and CLI topic rendering all carry explicit guards + regression tests. The one reachable-looking edge (`RawData::Vec` on an empty `AVec`) degrades to an `Err`, not a panic, and isn't produced by any sender.
- **`node_communication/mod.rs:303` `VecDeque`→`Vec` re-collection** — skipped. Cosmetic only (`Vec::from(mem::take(..))` would reuse the ring buffer), negligible impact; not worth the churn.
- **Move `data_len` into `Uint8Layout` to make `PreparedUint8Ipc` a symmetric newtype** (raised by the `/simplify` pass on #2945) — skipped. It's a relocation, not a reduction of state, and the altitude review argued keeping `data_len` out of the layout keeps the one-shot `encode_uint8_ipc_header` cleaner. Left as-is.
- **Extra metadata-helper doctests** (`get_*_param` no-coercion behavior, `carries_pattern_correlation`, `fnv1a` vectors) — skipped. These functions already carry accurate prose docs; the marginal value didn't clear the bar for this run, unlike the fully-undocumented `NodeId`/`DataId`.
- **`merged.rs` `MergeExternal` / `MergeExternalSend` near-identical impls** — skipped. Differ only by `Send + Sync` bounds; not cleanly dedupable without macros, so not a net simplification.

⚠️ All four PRs are **machine-generated by Claude** (an AI assistant). Each should be reviewed by a human before merging.

---

phil-opp:

No issues found. I checked the three send paths (`DoraNode::send_output_raw`, `send_output_array`, and the Python buffer-protocol path in `sample_handler.rs`) and the new `PreparedIpc` / `PreparedUint8Ipc` handles in `ipc_encode.rs`.

Key correctness points verified against the diff:
- `Prepared` (and `Uint8Layout`) are fully owned — `Layout.descs` hold Arc-backed `Buffer` clones and the flatbuffer headers are `Vec<u8>` — so neither prepared handle borrows from the source `ArrayData`. Holding `PreparedIpc` / `PreparedUint8Ipc` across `allocate_data_sample(...)` before encoding is therefore sound.
- `byte_len()` (used for sizing the sample) and the `dst.len() != prepared.total` / `layout.total` check inside the encode both read from the *same* prepared instance, so the allocated sample size and the encode's size assertion cannot diverge. The old code recomputed `prepare` / `uint8_layout` independently for sizing and for encoding — same result, just built twice.
- `encode_prepared_into` / `encode_uint8_prepared_into` are exact extractions of the previous one-shot bodies, and the retained one-shot `encode_ipc_into` / `encode_uint8_ipc_header` now delegate to them, so the emitted bytes are unchanged.

The optimization is straightforward and behavior-preserving, and the byte-for-byte IPC regression tests cover output equivalence.

---

phil-opp:
needs rebase

phil-opp:
Rebased onto the latest `main`. Since the arrow send path was refactored on main (`send_output_array`'s inline encoding moved into `SampleAllocator::encode_arrow`), I re-applied the prepare-once change at the new location rather than force-fitting the old diff — `encode_arrow` and `send_output_raw` now use `PreparedIpc` / `PreparedUint8Ipc`. No behavior change; output is byte-identical.

Re-verified on the rebased branch: `clippy -p dora-node-api -- -D warnings` clean, `test -p dora-node-api --lib` (169 tests) green, `check -p dora-node-api-python` clean, `fmt --check` clean, unwrap budget 162/163.

---

phil-opp:
This `Test (ubuntu-latest)` failure is again a merge-queue batch artifact, not this PR. The failing compile error is in the **coordinator**:

```
```

This PR touches only `apis/rust/node` + `apis/python/node` — nothing in `binaries/coordinator`. The error is a semantic conflict between two **sibling PRs batched together**: one adds the `ready_barrier_released` field to `RunningDataflow`, another adds a new initializer (at `handlers.rs:911`, a line that doesn't exist on this branch) that doesn't set it. On the rebased branch, `cargo check -p dora-coordinator --tests` compiles cleanly and every `RunningDataflow` initializer already sets that field.

The queue has already recognized this — it reports the batch ahead failed and is re-testing this PR on a fresh SHA. No change needed here.

---

phil-opp:

Reviewed the rebased commit (`0ece813`), which re-applies the prepare-once change at the new `SampleAllocator::encode_arrow` location plus `send_output_raw` and the Python buffer-protocol path. No issues found — looks safe to merge.

The rebased diff is a clean extraction: `PreparedIpc` / `PreparedUint8Ipc` size the sample (`byte_len()`) and encode from the *same* prepared instance whose `total` the encode's `dst.len()` assertion checks, so sizing and encoding cannot diverge; the handles are fully owned (no borrow from the source `ArrayData`); and the retained one-shot `encode_ipc_into` / `encode_uint8_ipc_header` delegate to the shared `encode_prepared_into` / `encode_uint8_prepared_into`, keeping output byte-identical.

---

PR #2946, opened 2026-08-01
refactor(node): dedup the EventItem-parameters match in the scheduler

phil-opp:
## Summary

`is_correlated` and `log_correlation_drop` in the event-stream scheduler each opened with the **identical** match extracting `&metadata.parameters` from an `EventItem` (`NodeEvent::Input` / `ZenohInput`, else bail):

```rust
let params = match event {
    EventItem::NodeEvent { event: NodeEvent::Input { metadata, .. }, .. } => &metadata.parameters,
    EventItem::ZenohInput { metadata, .. } => &metadata.parameters,
    _ => return ...,
};
```

Two copies of "which `EventItem` variants carry metadata parameters" can silently drift — e.g. a new parameter-carrying variant added to only one copy would make the eviction guard and its diagnostics disagree about the same event.

## Fix

Extract one `event_parameters(&EventItem) -> Option<&MetadataParameters>` helper as the single source of truth for that classification, and route both callers through it:
- `is_correlated` → `event_parameters(event).is_some_and(carries_pattern_correlation)`
- `log_correlation_drop` → a `let ... else` guard.

**No behavior change.**

## Validation

- `cargo test -p dora-node-api --lib scheduler` — 17 tests pass

## Disclaimer

⚠️ This PR is **machine-generated by Claude** (an AI assistant) as part of an automated code-review pass. A human should review before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01U5MBnacdwSwDRdKFComqWU)_

### DIFF
--- a/apis/rust/node/src/event_stream/scheduler.rs
+++ b/apis/rust/node/src/event_stream/scheduler.rs
@@ -7,10 +7,30 @@ use dora_message::{
     config::{DEFAULT_QUEUE_SIZE, QueuePolicy},
     daemon_to_node::NodeEvent,
     id::DataId,
-    metadata::{GOAL_ID, GOAL_STATUS, REQUEST_ID, carries_pattern_correlation, get_string_param},
+    metadata::{
+        GOAL_ID, GOAL_STATUS, MetadataParameters, REQUEST_ID, carries_pattern_correlation,
+        get_string_param,
+    },
 };
 
 use super::thread::EventItem;
+
+/// The metadata parameters carried by an input-bearing event, or `None` for
+/// events with no metadata (Stop, reload, input-closed, ...).
+///
+/// This is the single place that decides which `EventItem` variants carry
+/// parameters, so the eviction guard (`is_correlated`) and its diagnostics
+/// (`log_correlation_drop`) can never disagree on that classification.
+fn event_parameters(event: &EventItem) -> Option<&MetadataParameters> {
+    match event {
+        EventItem::NodeEvent {
+            event: NodeEvent::Input { metadata, .. },
+            ..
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. This is a pure refactor: the new `event_parameters(&EventItem) -> Option<&MetadataParameters>` helper matches exactly the same variants (`NodeEvent::Input` / `ZenohInput`) the two inlined matches did, and both call sites are equivalent to before — `is_correlated` becomes `event_parameters(event).is_some_and(carries_pattern_correlation)` (returns `false` for non-metadata variants, same as the old `_ => return false`), and `log_correlation_drop`'s `let ... else { return }` guard mirrors the old early return. No behavior change, and it removes the drift risk between the eviction guard and its diagnostics.

---

phil-opp:
Same as before: this `Unwrap budget` failure is not caused by this PR. The Trunk merge-queue tested it at batch SHA [`d8530d0`](https://github.com/dora-rs/dora/commit/d8530d07dd014143837688ce63f7618cad026849), which merges a **sibling** queued PR (`ed8e5f8`) modifying `binaries/coordinator/src/lib.rs` (+9) on top of this branch — that's where the extra `.unwrap()`/`.expect(` come from.

This PR only touches `apis/rust/node/src/event_stream/scheduler.rs` and adds **zero** `.unwrap()`/`.expect(` (it's a pure refactor that replaces two copy-pasted matches with one helper). On `origin/main` the count is `158`; this branch alone keeps it at `158`, 5 under the `163` budget.

No change is needed here; it should pass once the queue re-tests it without the coordinator PR in the same batch.

---

Issue #2920, opened 2026-07-31
Feature: dataflow completion semantics for timer-driven graphs + process-group cleanup when the CLI is killed


heyong4725:
**Versions:** dora-cli 1.0.0-rc.4; daemon + API from git rev `7eb4a5f8b`. macOS arm64.

## Problem 1: a timer keeps the dataflow alive forever

A graph whose worker nodes all finish cleanly never exits if any node consumes a
`dora/timer/...` input: the timer keeps ticking and `dora run` blocks indefinitely.
Observed: our client node logged "all N episodes done" and exited successfully; the
dataflow then sat for 9+ minutes (until an external kill), with the sim node discarding
timer events the whole time. (Verified with `dora run`, headless and windowed; under
`dora start` staying alive until `dora stop` is presumably intended, so this ask is
about `run`'s batch-style use.)

## Problem 2: killing `dora run` leaks every node process

When the blocked `dora run` is killed externally (`timeout N dora run ...`, Ctrl-C storm,
CI reaper), the node processes survive as orphans. We have measured leaked nodes at high
CPU skewing later timing-sensitive runs three separate times; each ad-hoc smoke needs a
manual `pkill` sweep afterwards.

## Asks

- An opt-in completion policy: e.g. `dora run/start --exit-when-nodes-finish` (timers
  don't count as keep-alive), or `--exit-on <node-id>` (dataflow ends when the named node
  exits). Either makes "run N episodes and exit" scriptable without wrapping in timeouts.
- `dora run` should place nodes in its own process group / kill children on abnormal
  exit, so an external kill of the CLI cannot leak simulator processes.

heyong4725:
## Architecture note: the remaining gap is `dora run`-only

Worth recording, because the second ask here reads broader than what is actually left.

**`dora up` + `dora start` already behave the way you would want**, and this is by design rather than accident. The CLI is a WebSocket client on 6013; `dora up` spawns the coordinator and daemon with `process_group(0)` (`binaries/cli/src/command/up.rs`), deliberately putting them in their own process groups so a signal aimed at the CLI's group cannot reach them.

Measured, with the CLI attached and then SIGKILLed (the unhandleable worst case):

| | |
|---|---|
| attached CLI | killed |
| coordinator + daemon | **both survive** |
| dataflow | still `Running`, 2 nodes |
| a new CLI | reconnects and sees it in `dora list` |

[...]

PR #2949, opened 2026-08-02
fix(daemon): reduce node-process leaks when `dora run` is killed (SIGTERM/SIGHUP)

heyong4725:
## What

On `main`, killing `dora run` leaks its node processes. This **reduces** that — it does not eliminate it. Read the "What this does not fix" section before relying on it.

Addresses part of the second ask in #2920 (no closing keyword — the issue stays open for the remaining work, and the timer/completion half is untouched).

## Why the leak happens

`dora run`'s only signal handling is `ctrlc::set_handler`, which on default features catches **SIGINT only**. SIGTERM and SIGHUP terminate the process outright, so no teardown runs.

That matters because nodes are spawned as process-group **leaders** — deliberate, so a terminal Ctrl-C cannot kill them out from under the daemon. An orphan therefore ends up with `ppid=1` and its own pgid: unreachable by both inherited signal delivery and a group-kill of the CLI. The daemon's teardown is the only thing that can reap it, and it never ran.

## The change

A SIGTERM/SIGHUP handler scoped to `Daemon::run_dataflow`, which serves exactly the two "run one dataflow and exit" modes (`dora run`, `dora daemon --run-dataflow`). It feeds the existing, already-tested `Event::CtrlC` path.

**Deliberately not** `ctrlc`'s `termination` feature. `dora run`, `dora daemon` and `dora coordinator` are subcommands of one binary and cargo unifies features, so that flag changes signal handling process-wide. An earlier revision of this PR did exactly that, and review caught the collateral: it overrides the `SIG_IGN` that `nohup` installs (breaking `dora cluster up`'s remote daemons), makes `pkill -TERM` of a coordinator gracefully destroy its daemons rather than crash, and reaches `dora record`'s handler, which has no escalation at all (#2952).

Escalation mirrors the ctrl-c handler exactly: first signal stops the dataflow, second takes the `SecondCtrlC` early-exit path (still unwinding, so `RunningDataflow` and every `ProcessHandle` drop), third aborts. Delivery uses `try_send` so a wedged event loop cannot swallow the escalation.

## What this does not fix

`Event::CtrlC` routes to `trigger_manual_stop`, which *requests* a graceful stop. The process exits once nodes report done, so **a node slow to answer can still be orphaned**. Measured 0 leaks in 6 runs after the escalation fix, and 1 in 6 before it — that is a sample, not a guarantee.

Making it deterministic needs real teardown: a shared process group for nodes plus `killpg`, or waiting on child exit before returning. That is tracked in #2920, which stays open.

SIGKILL remains unfixable in-process (needs `PR_SET_PDEATHSIG` on Linux or a kqueue watchdog on macOS).

## Verification

No automated test. The leak reproduces reliably from a shell but **not** under the cargo-test harness — four fixture/invocation shapes (a polling node, a producing node, a receive-only node in its own process group, and the exact shell invocation via `bash -c`) all had the node reaped at t+0 regardless of this change. A test there would pass with or without the fix, which is worse than none.

Verified by hand on the final state:

| check | result |
|---|---|
| `dora run` + SIGTERM reaps nodes | 0 leaks / 6 runs |
| doubled SIGTERM | 0 survivors (no mid-teardown abort) |
| `dora run` + SIGHUP | 0 survivors |
| `dora run` + SIGINT | unchanged |
| `nohup dora daemon` + SIGHUP | survives, as on `main` |
| `pkill -TERM` coordinator → daemon orphaned | 5/5, identical to `main` |
| `dora up` / `list` / `destroy` | unaffected |

- [x] `dora-daemon` 146 unit tests
- [x] `fault-tolerance-e2e` 11/11
- [x] fmt + clippy clean

## Related

Filed from this review, independent of this change: #2951 (no `timeout -k` at any CI call site) and #2952 (`dora record` has no signal escalation).

### DIFF
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -50,7 +50,7 @@ uuid = { workspace = true }
 inquire = { workspace = true }
 tokio-tungstenite = "0.29"
 notify = "8.2.0"
-ctrlc = "3.5.2"
+ctrlc = { version = "3.5.2", features = ["termination"] }
 tracing = { workspace = true }
 tracing-log = "0.2.0"
 dora-tracing = { workspace = true, optional = true }
--- a/binaries/cli/Cargo.toml
+++ b/binaries/cli/Cargo.toml
@@ -50,7 +50,7 @@ uuid = { workspace = true }
 inquire = { workspace = true }
 tokio-tungstenite = "0.29"
 notify = "8.2.0"
-ctrlc = { version = "3.5.2", features = ["termination"] }
+ctrlc = "3.5.2"
 tracing = { workspace = true }
 tracing-log = "0.2.0"
 dora-tracing = { workspace = true, optional = true }
--- a/binaries/coordinator/Cargo.toml
+++ b/binaries/coordinator/Cargo.toml
@@ -29,7 +29,7 @@ futures-concurrency = "7.7.1"
 serde = { workspace = true }
 serde_json = { workspace = true }
 petname = "3.1.0"
-ctrlc = { version = "3.5.2", features = ["termination"] }
+ctrlc = "3.5.2"
 log = { version = "0.4.33", features = ["serde"] }
 dora-message = { workspace = true }
 dora-coordinator-store = { workspace = true }
--- a/binaries/coordinator/Cargo.toml
+++ b/binaries/coordinator/Cargo.toml
@@ -29,7 +29,7 @@ futures-concurrency = "7.7.1"
 serde = { workspace = true }
 serde_json = { workspace = true }
 petname = "3.1.0"
-ctrlc = "3.5.2"
+ctrlc = { version = "3.5.2", features = ["termination"] }
 log = { version = "0.4.33", features = ["serde"] }
 dora-message = { workspace = true }
 dora-coordinator-store = { workspace = true }
--- a/binaries/daemon/Cargo.toml
+++ b/binaries/daemon/Cargo.toml
@@ -42,7 +42,7 @@ futures = { workspace = true }
 shared_memory_extended = "0.13.0"
 bincode = { workspace = true }
 aligned-vec = "0.6.4"
-ctrlc = "3.5.2"
+ctrlc = { version = "3.5.2", features = ["termination"] }
 which = "8"
 crossbeam = "0.8.4"
 crossbeam-skiplist = "0.1.3"
--- a/binaries/daemon/Cargo.toml
+++ b/binaries/daemon/Cargo.toml
@@ -42,7 +42,7 @@ futures = { workspace = true }
 shared_memory_extended = "0.13.0"
 bincode = { workspace = true }
 aligned-vec = "0.6.4"
-ctrlc = { version = "3.5.2", features = ["termination"] }
+ctrlc = "3.5.2"
 which = "8"
 crossbeam = "0.8.4"
 crossbeam-skiplist = "0.1.3"
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1095,6 +1095,7 @@ impl Daemon {
         let clock = Arc::new(HLC::default());
 
         let ctrlc_events = ReceiverStream::new(set_up_ctrlc_handler(clock.clone())?);
+        let termination_events = ReceiverStream::new(set_up_termination_handler(clock.clone()));
 
         // Set up optional timeout for --stop-after
         let timeout_events = if let Some(duration) = stop_after {
@@ -1137,6 +1138,7 @@ impl Daemon {
         let events = (
             coordinator_events,
             ctrlc_events,
+            termination_events,
             timeout_events,
             dynamic_node_events,
         )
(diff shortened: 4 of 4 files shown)


heyong4725:
**Converted to draft — this is not a complete fix, and the original mechanism was wrong.** Review found both, and measuring changed my conclusions twice.

## The Cargo.toml approach was the wrong mechanism

`dora run`, `dora daemon` and `dora coordinator` are subcommands of **one binary**, and cargo unifies features — so `ctrlc = { features = ["termination"] }` changes signal handling process-wide no matter which crate declares it. It cannot be scoped that way at all. Real collateral from it:

- **`nohup` is silently overridden.** `ctrlc::set_handler` registers with `overwrite=true`, so it installs over the `SIG_IGN` that `nohup` sets. `dora cluster up` runs `nohup dora daemon … &` over ssh (`cluster/up.rs:70`) and the guide documents that pattern.
- **`dora record`'s handler has no escalation** (`record.rs:391` just sends on a channel), so SIGTERM would be swallowed permanently — only SIGKILL would stop it.

`bf0629ae` replaces it with handling scoped to `run_dataflow`, which serves exactly the two "run one dataflow and exit" modes. Long-lived services keep their current semantics. Second signal aborts; delivery uses `try_send` rather than an await so that escape hatch survives a wedged event loop.

## But it is still not a fix — 1 run in 6 leaks

`Event::CtrlC` routes to `trigger_manual_stop`, which *requests* a graceful stop. The process exits once nodes report done, so a node slow to answer outlives it. Measured 6 consecutive runs: 5 clean, 1 orphan. Closing that gap needs deterministic teardown — a shared process group for nodes plus `killpg`, or waiting on child exit before returning — which is a different and larger change than signal routing. I'd rather leave this in draft than land an 83% fix described as a fix.

## One review finding was misattributed, and it matters

The review flagged that this PR neuters nightly's crash-recovery jobs, which do `pkill -TERM -f "dora coordinator"` with the comment *"leave daemon orphaned (simulates a coord crash)"* (`nightly.yml:905`, `:1088`).

I measured the **baseline on clean `main`**: 2 processes → 0. The daemon already dies when the coordinator is SIGTERMed, so that premise is **already false today**, independent of this PR. Worth its own issue — `redb-backend-smoke` and `state-reconstruction` are not testing what their comments claim, and they pass anyway, so it fails silently. `pkill -KILL` would restore the intended crash semantics.

Two other findings stand on their own merits, also independent of this PR:

- **No `timeout -k` / `--kill-after` anywhere** in `.github/workflows/` or `scripts/` — 21 bare `timeout N dora …` call sites. Every CI time bound is soft against a process that declines to die.
- **`dora record` has no signal escalation**, unlike the daemon (aborts on 3rd), coordinator (2nd) and attach (2nd).

Happy to file those three separately, and to take the deterministic-teardown work as its own piece if wanted.

heyong4725 (commented):
Found two blocking issues on bf0629ae8:

1. [P1] Remove the process-wide ctrlc termination feature (binaries/cli/Cargo.toml:53; binaries/coordinator/Cargo.toml:32; binaries/daemon/Cargo.toml:45).

The latest daemon code says this feature must not be enabled, but all three manifest changes from cd9d78846 are still present. cargo tree -e features -i ctrlc confirms that termination is unified into the single dora executable through cli, coordinator, and daemon. ctrlc 3.5.2 then installs SIGTERM/SIGHUP handlers for every ctrlc::set_handler call and overwrites existing Unix dispositions. Consequently the current head still overrides nohup SIGHUP handling for detached daemons, turns coordinator SIGTERM into its graceful CtrlC path, and makes dora record swallow TERM through its one-shot CtrlC handler. In run_dataflow it also overlaps the new Tokio SIGTERM/SIGHUP listener, so a termination signal can enter two independent counters/event paths. Revert all three Cargo.toml feature changes so the new handler is actually scoped as documented.

2. [P1] Do not abort before tracked node processes have been killed/reaped (binaries/daemon/src/lib.rs:6065).

On the first termination event, trigger_manual_stop calls stop_all with force=false. stop_all only spawns a background escalation task: SIGTERM after 10 seconds and kill after another 5 seconds. If a second TERM/HUP arrives during that interval, lines 6066-6068 abort the daemon immediately. Nodes are deliberately separate process-group leaders, so aborting the parent also cancels the only escalation task and leaves any still-running nodes orphaned. A timeout/reaper that sends TERM more than once therefore deterministically recreates the exact leak this PR is intended to fix; the documented 1-in-6 single-signal leak shows the first path is not deterministic either. Escalation needs to kill the tracked process groups and wait for their exits before the CLI exits, including the repeated-signal path.

Validation: diff check passed and the PR revision compiled. cargo test -p dora-daemon --lib ran 146 tests: 142 passed; four existing loopback-endpoint spawner tests failed only from the temporary review worktree sandbox and all four pass on the identical main base from the repository worktree. No signal/child-cleanup regression test is added in this PR; please add one that sends SIGTERM (and repeated SIGTERM) and asserts the CLI exits with no node process group left behind.

heyong4725:
Both P1s confirmed and fixed in `26680b59`. Both were mine, and one of them had propagated into a bad conclusion I published elsewhere.

## P1a — the feature flag really was still there

Correct. The previous commit claimed to replace it, but reverted the three `Cargo.toml`s with `git checkout --`, which restores to **HEAD** — and HEAD was the commit that *added* the flag. So the branch carried both mechanisms and every bit of process-wide collateral the scoped handler exists to avoid. Reverted against `origin/main` this time; `git diff origin/main --stat` is now `binaries/daemon/src/lib.rs` alone.

**That mistake also invalidated a measurement I used to push back on this review.** I claimed nightly's coordinator-crash simulation was already broken on `main`, citing a baseline of 2 processes → 0. That baseline was taken with the flag still enabled — which is precisely what causes the coordinator to gracefully destroy its daemons. Re-measured properly, 5 runs each:

```
this branch (flag genuinely gone): daemon survived 5 / 5
clean main:                        daemon survived 5 / 5
```

So the crash simulation works, the earlier review's collateral finding was right, and my "correction" of it was wrong. #2950 is closed as invalid with that explanation.

## P1b — aborting on the second signal orphaned what it was cleaning up

Also correct. The ladder now mirrors the ctrl-c handler exactly: first signal stops the dataflow, second takes the `SecondCtrlC` early-exit path — which still unwinds, dropping the `RunningDataflow` and with it every `ProcessHandle` — and only a third aborts. The channel holds 2 so a queued first event cannot make the second undeliverable, and an undeliverable *first* signal now warns rather than aborting, since teardown may still be in flight.

## Verified on the final state

| check | result |
|---|---|
| `dora run` + SIGTERM reaps nodes | 0 leaks / 6 runs (was 1 / 6 before the ladder fix) |
| doubled SIGTERM | 0 survivors |
| `nohup dora daemon` + SIGHUP | survives |
| `pkill -TERM` coordinator | daemon survives 5/5, identical to main |
| `dora-daemon` unit tests | 146 pass, fmt/clippy clean |

On your 142/146: I see 146/146 locally on the final state. If four are still failing for you, they may be the sandbox-sensitive ones you mention passing on main — worth a look either way, so tell me which and I'll dig.

**Still draft.** The residual race described earlier stands: `Event::CtrlC` requests a graceful stop and the process can exit before a slow node is reaped. 6/6 clean runs is better than the 5/6 I measured before, but that is a sample, not a guarantee — the deterministic fix is a shared process group plus `killpg`, or waiting on child exit. I would rather land that than claim this is finished.

The other two issues I filed from this review — #2951 (no `timeout -k` anywhere) and #2952 (`dora record` has no signal escalation) — are unaffected by any of the above; both are plain code/grep facts.

PR #2953, opened 2026-08-02
fix(cli): escalate `dora record` on a repeated Ctrl-C

### LINKED ISSUE #2952: cli: `dora record`'s signal handler has no escalation — a repeated Ctrl-C does nothing
Found while reviewing #2949. Pre-existing and independent of that PR.

## The inconsistency

Every other signal handler in the tree escalates on a repeated signal, so an operator who presses Ctrl-C again gets out:

| Handler | Escalation |
|---|---|
| `binaries/daemon/src/lib.rs` `set_up_ctrlc_handler` | 2nd → `SecondCtrlC` (immediate exit), 3rd → `std::process::abort()` |
| `binaries/coordinator/src/events.rs` `set_up_ctrlc_handler` | 2nd → `abort()` |
| `binaries/cli/src/command/start/attach.rs:116` | 2nd → force-stop + `exit(1)` |
| `binaries/cli/src/command/record.rs:391` | **none** |

`dora record`'s is:

```rust
ctrlc::set_handler(move || {
    let _ = stop_tx.send(());
})
```

Once the first signal is delivered, every subsequent Ctrl-C is swallowed. The handler is installed for the life of the process, so the default "SIGINT terminates" disposition is gone and repeated presses do nothing at all.

## Why it can actually hang

The poll loop checks `stop_rx.try_recv()` every 100 ms, so the happy path is responsive. But the work between polls is blocking file I/O — `writer.write_entry`, and the `flush`/`finish` on the way out. If a write blocks (full disk, stalled network mount, a very large flush), the loop never reaches the next poll and there is no second-signal path to break out. The only escape is SIGKILL from another terminal.

`nightly.yml:385` and `scripts/qa/ci-nightly-jobs.sh:415` wrap it in `timeout 300s` with no `-k` (see the companion issue), so in CI that state would hang to the job cap rather than failing at 300s.

## Suggested fix

Mirror the pattern the other three already use — count signals in the closure and escalate:

```rust
let mut signalled = false;
ctrlc::set_handler(move || {
    if signalled {
        eprintln!("received second Ctrl-C -> exiting immediately (recording may be truncated)");
        std::process::exit(130);
    }
    signalled = true;
    let _ = stop_tx.send(());
})
```

`exit(130)` rather than `abort()` seems right here: a truncated recording is a legitimate outcome of an impatient operator, and 130 is the conventional SIGINT status. Whether the partial file is left in place or removed is worth deciding explicitly in the fix — `dora replay` should not be handed a half-written file without the user knowing.

phil-opp:
## Summary

Fixes #2952.

`dora record`'s SIGINT handler (`binaries/cli/src/command/record.rs`) only sent a stop request and then swallowed every subsequent Ctrl-C:

```rust
ctrlc::set_handler(move || {
    let _ = stop_tx.send(());
})
```

The poll loop checks the stop channel every 100 ms, so the happy path is responsive — but the work between polls is blocking file I/O (`write_entry`, `flush`, `finish`). If a write wedges (full disk, stalled network mount, a very large flush), the loop never reaches the next poll and there is no second-signal path to break out. Because the handler stays installed for the life of the process, the default "SIGINT terminates" disposition is gone too, so repeated Ctrl-C does nothing at all and the only escape is SIGKILL from another terminal.

Every other signal handler in the tree already escalates on a repeated signal:

| Handler | Escalation |
|---|---|
| `binaries/daemon/src/lib.rs` `set_up_ctrlc_handler` | 2nd → `SecondCtrlC`, 3rd → `abort()` |
| `binaries/coordinator/src/events.rs` `set_up_ctrlc_handler` | 2nd → `abort()` |
| `binaries/cli/src/command/start/attach.rs` | 2nd → force-stop + `exit(1)` |
| `binaries/cli/src/command/record.rs` | **none** (this PR) |

## Change

Count signals in the closure and escalate. The second Ctrl-C exits immediately with the conventional SIGINT status (`130`) and prints a message warning that the recording may be truncated — so a partial `.drec` is not handed to `dora replay` silently.

`exit(130)` rather than `abort()`: a truncated recording is a legitimate outcome of an impatient operator, and 130 is the conventional SIGINT status. The partial file is left in place (removing it from inside the async-unsafe handler would be riskier than the visible warning).

## Testing

- `cargo clippy -p dora-cli -- -D warnings` (clean)

A signal handler that ends in `process::exit` isn't unit-testable in-process; the change mirrors the already-established, exercised pattern in the daemon/coordinator/attach handlers.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EpwSeg7gZkeNEWq11apYNF)_

### DIFF
--- a/binaries/cli/src/command/record.rs
+++ b/binaries/cli/src/command/record.rs
@@ -414,15 +414,26 @@ fn run_record_proxy(args: Record) -> eyre::Result<()> {
     // Set up Ctrl-C handler.
     //
     // The first Ctrl-C requests a graceful stop (finalize the recording and
-    // exit). Escalate on a repeated signal — mirroring the daemon, coordinator,
-    // and `dora start` attach handlers — so an operator is never stuck: the
+    // exit). Escalate on a repeated signal — as the daemon, coordinator, and
+    // `dora start` attach handlers all do — so an operator is never stuck: the
     // recording loop does blocking file I/O between poll ticks, and a write that
     // wedges (full disk, stalled network mount, a large flush) would otherwise
     // swallow every subsequent Ctrl-C with no way out but SIGKILL from another
     // terminal. The second signal exits immediately with the conventional SIGINT
     // status; the partial file is left in place, but the message warns that it
-    // may be truncated so `dora replay` is not handed a half-written file
-    // silently.
+    // may be truncated so the operator knows `dora replay` will get a short
+    // recording (a missing footer is tolerated by `RecordingReader`, and the
+    // flush before `finish()` below bounds the loss to the footer).
+    //
+    // Two deliberate departures from those siblings, both because what wedges
+    // here is our own teardown rather than a child process:
+    //   * `process::exit` rather than the daemon's unwinding `bail!` — the
+    //     thing being escaped IS `writer.finish()`, so unwinding through it
+    //     would re-enter the wedge. Nothing here owns child processes to
+    //     orphan.
+    //   * exit 130 (128 + SIGINT) rather than the attach handler's 1, so a
+    //     supervisor can tell an operator abort from a recording error. This
+    //     is the only place in the CLI that returns it; see docs/cli.md.
     let (stop_tx, stop_rx) = std::sync::mpsc::channel();
     let mut signalled = false;
     ctrlc::set_handler(move || match record_ctrlc_action(&mut signalled) {
@@ -482,6 +493,17 @@ fn run_record_proxy(args: Record) -> eyre::Result<()> {
             Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => break,
         }
     }
+
+    // Flush before the blocking finalize, not as part of it. `finish()` writes
+    // the buffered tail AND the footer, and it is the call most likely to wedge
+    // (full disk, stalled network mount) — which is exactly when an operator
+    // reaches for the second Ctrl-C. Escalating out of an unflushed `finish()`
+    // discards every entry since the last 100-message flush: with fewer than
+    // 100 messages recorded that is the whole file, and `dora replay` then
+    // fails with "failed to read recording header" rather than replaying a
+    // short recording. Flushing here bounds what escalation can cost to the
+    // 24-byte footer, which readers already tolerate.
+    writer.flush()?;
 
     let footer = writer.finish()?;
     eprintln!(
--- a/binaries/cli/src/command/record.rs
+++ b/binaries/cli/src/command/record.rs
@@ -386,9 +386,26 @@ fn run_record_proxy(args: Record) -> eyre::Result<()> {
 
     eprintln!("Recording... (press Ctrl-C to stop)");
 
-    // Set up Ctrl-C handler
+    // Set up Ctrl-C handler.
+    //
+    // The first Ctrl-C requests a graceful stop (finalize the recording and
+    // exit). Escalate on a repeated signal — mirroring the daemon, coordinator,
+    // and `dora start` attach handlers — so an operator is never stuck: the
+    // recording loop does blocking file I/O between poll ticks, and a write that
+    // wedges (full disk, stalled network mount, a large flush) would otherwise
+    // swallow every subsequent Ctrl-C with no way out but SIGKILL from another
+    // terminal. The second signal exits immediately with the conventional SIGINT
+    // status; the partial file is left in place, but the message warns that it
+    // may be truncated so `dora replay` is not handed a half-written file
+    // silently.
     let (stop_tx, stop_rx) = std::sync::mpsc::channel();
+    let mut signalled = false;
     ctrlc::set_handler(move || {
+        if signalled {
+            eprintln!("received second Ctrl-C -> exiting immediately (recording may be truncated)");
+            std::process::exit(130);
+        }
+        signalled = true;
         let _ = stop_tx.send(());
     })
     .wrap_err("failed to set ctrl-c handler")?;
--- a/binaries/cli/src/command/record.rs
+++ b/binaries/cli/src/command/record.rs
@@ -253,6 +253,31 @@ fn run_record(args: Record) -> eyre::Result<()> {
         .execute()
 }
 
+/// What a Ctrl-C on the recording loop should do, given whether one was already
+/// seen. Split out from the signal handler so the escalation state machine is
+/// unit-testable without delivering a real signal (which also exits the process
+/// via `process::exit`, and races the handler-installation window).
+#[derive(Debug, PartialEq, Eq)]
+enum CtrlCAction {
+    /// First Ctrl-C: request a graceful stop (finalize the recording).
+    RequestStop,
+    /// Repeated Ctrl-C: give up and exit immediately, even if a write is wedged.
+    ExitImmediately,
+}
+
+/// Advance the Ctrl-C escalation state: the first call requests a stop, every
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. The repeated-Ctrl-C escalation is implemented correctly: `ctrlc::set_handler` takes an `FnMut`, and the `move` closure captures `mut signalled` so its state persists across signals — the first Ctrl-C sends the stop request, the second exits with status 130. This matches the escalation pattern used by the daemon/coordinator/`dora start` handlers.

---

heyong4725:
Reviewed. The change is right and the scope is right — one clarification worth adding to the record, and one gap.

## Scope is correct, and narrower than the issue implies

`run_record_proxy` is only reached with `--proxy` (`record.rs:71`). The default `dora record <yml>` goes through `run_record`, which has no handler of its own — it builds a rewritten descriptor and delegates:

```rust
Run::new(tmp_path.to_string_lossy().to_string())
    .execute()
```

so it inherits the daemon's ctrl-c handler, which already escalates (`SecondCtrlC` at `daemon/lib.rs:1597`, `abort()` at `:6026`). That makes the proxy handler the **only** unescalated one in the tree, which is exactly what this patches. I mention it because the issue body cites `record.rs:391` without noting the mode split, and a reader could reasonably think the common path was affected.

## Implementation

`ctrlc::set_handler` takes `FnMut`, so the captured `signalled` persists across invocations and the second signal takes the escalation branch. The crate dispatches from a single dedicated thread, so there is no race on the flag. `exit(130)` skips the flush deliberately, and the message says the recording may be truncated — which answers the "decide explicitly" point in the issue rather than leaving it implicit.

## Gap: no automated coverage

Nothing exercises this. It is awkward to test but not impossible: spawn `dora record --proxy` against a running dataflow, wait for the `press Ctrl-C to stop` banner, send SIGINT twice, assert exit 130 and the message. Worth it because the failure mode is silent — if a future refactor moves `set_handler` or drops the `mut`, nothing fails.

Two traps I hit trying to verify by hand, in case anyone else tries:

- Signalling before the handler is installed exits **130 via the default SIGINT disposition** — the same code the escalation uses. I nearly recorded a false pass from that.
- `dora record <yml>` without `--proxy` never reaches this code, and its output also contains the word "Recording", so a naive `grep` gate matches the wrong mode.

Both are reasons the test should wait on the exact `press Ctrl-C to stop` banner and assert the escalation **message**, not just the exit code.

CI is green (0 failures). No objection to merging as-is; the test would be a good follow-up.

phil-opp:
Thanks for the thorough review, and for confirming the mode split — you're right that `run_record` (the non-`--proxy` path) delegates to `Run` and inherits the daemon's already-escalating handler, so `run_record_proxy` is the only unescalated one, which is what this patches.

Added coverage in `ad687c7`. Rather than the signal-timing integration test — which, as you found, has real false-pass traps (pre-install SIGINT exits 130 via the default disposition; the non-proxy path also prints "Recording") — I extracted the escalation decision into a small helper and unit-tested it:

```rust
fn record_ctrlc_action(signalled: &mut bool) -> CtrlCAction {
    if *signalled { CtrlCAction::ExitImmediately }
    else { *signalled = true; CtrlCAction::RequestStop }
}
```

The handler is now a thin `match` over it. The test pins first-signal → `RequestStop`, every later signal → `ExitImmediately`, deterministically and with no process/signal machinery. It also directly guards your "drops the `mut`" concern: the helper takes `&mut bool`, so removing `mut` from the handler's `signalled` fails to compile rather than silently disabling escalation.

This doesn't assert the exact banner/exit wiring end-to-end, so a real `--proxy` SIGINT integration test (waiting on the `press Ctrl-C to stop` banner and asserting the escalation message, per your design) is still a reasonable further follow-up if we want that last mile.

---

phil-opp:
The new commit `ad687c7` addresses the earlier "no automated coverage" feedback by extracting the escalation decision into `record_ctrlc_action(&mut bool) -> CtrlCAction` and unit-testing it: first signal → `RequestStop`, every subsequent → `ExitImmediately`. The test is non-vacuous (it would fail if the first/repeat branches were swapped or if escalation were disabled), and taking `&mut bool` makes a future drop of the handler's `mut` a compile error rather than a silent regression. No issues found in the new commit.

---

---

heyong4725:
Reviewed with three independent passes (adversarial, consistency/testing, Codex). One finding was worth acting on and is fixed in `0ea3b5d10`; the rest are recorded below rather than changed.

## Fixed: the escalation could destroy a recording the old code always completed

Before this PR every Ctrl-C after the first was a no-op, so `writer.finish()` always ran. `finish()` writes the buffered tail **and** the footer, and the loop only flushes every 100 messages — so escalating out of a wedged `finish()` discards everything since the last flush. With fewer than 100 messages recorded that is the whole file: verified empirically at **0 bytes on disk**, after which `dora replay` fails with `failed to read recording header` rather than replaying a short recording. Worse where the output path already held a good `.drec`, since `File::create` truncated it at startup. The handler's "may be truncated" is then optimistic — the real outcome is "gone".

`writer.flush()` now runs before `finish()`, so escalation costs at most the 24-byte footer, which `RecordingReader` already tolerates (`truncated_file_no_footer`).

## Also changed: the sibling comment

The comment claimed to mirror the daemon, coordinator and attach handlers. It diverges from all three: they unwind (`bail!`) or `abort()`, this uses a non-unwinding `process::exit`, and 130 appears nowhere else in the CLI — `attach.rs:125`, same binary, same double-Ctrl-C, returns 1. Both departures look right to me (the wedge being escaped *is* our teardown, so unwinding would re-enter it; and a supervisor should be able to tell an operator abort from a recording error), so the comment now states them as deliberate. `docs/cli.md` gained the ladder, as it already documents for `dora start --attach`.

## Not changed, for you to judge

**The unit test is tautological.** `record_ctrlc_action` is five lines with one call site, and the test asserts its two branches. The regression most likely to actually ship — moving `let mut signalled = false` inside the closure, which silently disables escalation forever — leaves it green, as do changing the exit code, dropping the message, or installing the handler too late. All three siblings inline this same `if ctrlc_sent` with no helper and no test, so the extraction is a divergence introduced to make the test possible. Two honest options: drop the extraction and the test (matching the siblings, losing nothing the four-line closure doesn't already show), or spend an e2e — `tests/node-lifecycle-e2e.rs::run_killed_by_sigterm_terminates_nodes_and_exits` is the pattern, and would pin exit 130 plus a still-readable `.drec`. I did not pick for you.

**`eprintln!` before `exit(130)` can itself block.** It takes the stderr lock, so `dora record --proxy 2>&1 | tee /mnt/stalled/log` re-wedges the escape hatch. A raw `write(2)` to fd 2 avoids it; `libc` is an optional dep of `dora-cli` today, so it is not a one-liner.

**SIGTERM is uncovered.** `ctrlc` is declared without the `termination` feature, so a supervisor's `SIGTERM` still kills `dora record --proxy` with no `finish()` and no message — the same footerless file this PR warns about, silently. Deliberate per #2949, but worth its own issue now that the interactive path is handled.

**The branch predates `main`.** It is cut from `397bfd225`, so it misses #2949 (`set_up_termination_handler`, the newest escalation precedent, whose doc argues *for* unwinding) and #2961 (the e2e that signals a spawned CLI, which is exactly the tier the test discussion needs). Worth a rebase before merge.

Validation: `cargo fmt`, `clippy -p dora-cli` clean, `cargo test -p dora-cli --lib record::` 6/6.

PR #2954, opened 2026-08-02
fix(daemon): deny descriptor override of control-plane env vars

### LINKED ISSUE #2944: daemon: node `env:` can override the daemon's own control-plane variables (DORA_ZENOH_*, DORA_RUNTIME_CONFIG)
Found while reviewing #2943. Not a privilege boundary crossing — anyone who can write a dataflow descriptor already controls `path:`/`args:` — so this is a footgun and a defense-in-depth gap rather than an escalation. Filing because the current protection is accidental and therefore fragile.

## What happens

In both spawn paths (`binaries/daemon/src/spawn/spawner.rs`), the daemon injects its own control-plane variables **before** applying the user's `env:` entries, and `is_denied_env` does not cover them:

```rust
command = strip_denied_env(command);
command = command.env("DORA_NODE_CONFIG", …);      // or DORA_RUNTIME_CONFIG
command = self.maybe_inject_zenoh_connect(command, &node.id);
// then, later:
if let Some(envs) = &node.env {
    for (key, value) in envs {
        if !is_denied_env(key) { command = command.env(key, value.to_string()); }
    }
}
```

So a later `.env()` from the descriptor wins. Verified with a probe node that dumps what it received:

```yaml
nodes:
  - id: probe
    path: probe.py
    env:
      DORA_ZENOH_MULTICAST: FORGED
```
```
DORA_ZENOH_CONNECT=tcp/127.0.0.1:52868
DORA_ZENOH_LISTEN=tcp/127.0.0.1:52869
DORA_ZENOH_MULTICAST=FORGED        <- descriptor value won
```

`ENV_DENYLIST` (`spawner.rs:38`) covers the loader variables and `DORA_AUTH_TOKEN` / `DORA_ALLOW_SHELL_NODES`, but none of the wiring variables.

## The part that makes it worth fixing

`DORA_NODE_CONFIG` — the most sensitive of the three, since it carries `dataflow_id`, `node_id` and the daemon-communication config — is **not** currently forgeable. I tried, and the node received the real config.

But it is protected by accident. `spawn_inner` (`binaries/daemon/src/spawn/prepared.rs:515-527`) re-serializes `DORA_NODE_CONFIG` from the authoritative `node_config` *after* the command was built:

```rust
// Re-serialize DORA_NODE_CONFIG from the current node_config. The command
// was built once at initial spawn time with the initial config; the
// restart_loop updates self.node_config.restart_count between restarts but
// the command's baked-in env var was stale.
```

That exists to fix a restart-count staleness bug, not as a security measure, and it applies to `DORA_NODE_CONFIG` only. So:

- the one variable that is safe is safe for an unrelated reason, and a future refactor of that re-serialization would silently make it forgeable;
- its siblings (`DORA_ZENOH_LISTEN` / `DORA_ZENOH_CONNECT` / `DORA_ZENOH_MULTICAST`, and `DORA_RUNTIME_CONFIG` by inspection — same position, no re-application) are forgeable today.

## Impact

Mostly self-inflicted: `DORA_ZENOH_CONNECT` pointed somewhere wrong cuts a node off from its peers, which surfaces as a dataflow that starts cleanly and then silently exchanges nothing — an unpleasant debugging session, and one where the descriptor looks innocent because the var is spelled like an ordinary setting. A forged `DORA_RUNTIME_CONFIG` would hand an operator runtime the wrong operator set.

The reach widened slightly in #2943: dataflow-level `env:` now also flows to nodes added with `dora node add`, so these variables can be set for a node that was not in the original descriptor.

## Suggested fix

Either:

1. **Apply user env first, daemon-injected vars last** in both spawn paths, so the daemon's wiring always wins by construction; or
2. **Add the control-plane keys to `ENV_DENYLIST`** (`DORA_NODE_CONFIG`, `DORA_RUNTIME_CONFIG`, `DORA_ZENOH_LISTEN`, `DORA_ZENOH_CONNECT`, `DORA_ZENOH_MULTICAST`) so an attempt is dropped with the existing warning rather than silently honored.

(1) is stronger and needs no list maintenance; (2) is more discoverable, because `is_denied_env` already logs a warning naming the key, which turns a silent misconfiguration into a visible one. Doing both would also make `DORA_NODE_CONFIG`'s safety intentional instead of incidental.

Worth a test either way: nothing currently pins that a descriptor cannot override the daemon's wiring.

phil-opp:
## Summary

Fixes #2944.

In both spawn paths (`binaries/daemon/src/spawn/spawner.rs`), the daemon injects its own control-plane variables from the authoritative node/runtime config **before** applying the descriptor's `env:` entries, and `is_denied_env` did not cover them — so a later `.env()` from the descriptor silently won:

```yaml
nodes:
  - id: probe
    path: probe.py
    env:
      DORA_ZENOH_MULTICAST: FORGED   # descriptor value overrode the daemon's
```

### Why it's worth fixing

- A wrong `DORA_ZENOH_CONNECT` / `DORA_ZENOH_LISTEN` cuts a node off from its peers — a dataflow that starts cleanly and then silently exchanges nothing, with a descriptor that looks innocent because the var is spelled like an ordinary setting.
- A forged `DORA_RUNTIME_CONFIG` would hand an operator runtime the wrong operator set.
- `DORA_NODE_CONFIG` (the most sensitive — it carries `dataflow_id`, `node_id`, daemon-comm config) was safe only **by accident**: `spawn_inner` re-serializes it later for an unrelated restart-count fix. A future refactor of that re-serialization would silently make it forgeable.

Not a privilege-boundary crossing (anyone who can write a descriptor already controls `path:`/`args:`), so this is a footgun / defense-in-depth gap — but the current protection is accidental and therefore fragile.

## Change

Add the five control-plane keys — `DORA_NODE_CONFIG`, `DORA_RUNTIME_CONFIG`, `DORA_ZENOH_LISTEN`, `DORA_ZENOH_CONNECT`, `DORA_ZENOH_MULTICAST` — to `ENV_DENYLIST` (option 2 in the issue, using the existing zenoh env-name constants). The daemon still injects each via its own direct `.env()` calls (`maybe_inject_zenoh_connect`, `DORA_NODE_CONFIG`/`DORA_RUNTIME_CONFIG`), which don't go through `is_denied_env`, so the daemon's wiring is unaffected — only the descriptor's user-env loops now drop an override attempt. This makes `DORA_NODE_CONFIG`'s safety intentional rather than incidental and turns a would-be silent override into the visible warning that already names the key. The warning text is generalized since the denylist now covers more than shared-library injection.

## Testing

- New unit test `control_plane_env_vars_are_denied` pins that every control-plane key is denied while an ordinary node variable still passes through (nothing pinned this before).
- `cargo test -p dora-daemon` (new test passes), `cargo clippy -p dora-daemon -- -D warnings` (clean), `cargo fmt --all -- --check`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EpwSeg7gZkeNEWq11apYNF)_

### DIFF
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -35,6 +35,18 @@ use std::{
 use tokio::sync::mpsc;
 
 /// Environment variable names that must never be passed to spawned nodes.
+///
+/// Two groups:
+/// - dynamic-loader / auth variables that could inject shared libraries or leak
+///   the daemon's own secrets to child nodes;
+/// - the daemon's control-plane wiring variables. These are injected by the
+///   daemon from the authoritative node/runtime config *before* the descriptor's
+///   `env:` is applied, so without a denylist a later `.env()` from the
+///   descriptor would silently win — pointing a node's zenoh connect/listen at
+///   the wrong endpoint (a dataflow that starts cleanly and then exchanges
+///   nothing), or handing an operator runtime the wrong operator set. Denying
+///   them keeps the daemon's wiring authoritative and turns a would-be silent
+///   override into the visible warning below (#2944).
 const ENV_DENYLIST: &[&str] = &[
     "LD_PRELOAD",
     "LD_AUDIT",
@@ -43,13 +55,20 @@ const ENV_DENYLIST: &[&str] = &[
     "LD_LIBRARY_PATH",
     "DORA_AUTH_TOKEN",
     "DORA_ALLOW_SHELL_NODES",
+    // Control-plane wiring — owned by the daemon, never by the descriptor.
+    "DORA_NODE_CONFIG",
+    "DORA_RUNTIME_CONFIG",
+    DORA_ZENOH_LISTEN_ENV,
+    DORA_ZENOH_CONNECT_ENV,
+    DORA_ZENOH_MULTICAST_ENV,
 ];
 
 /// Returns true if the env var key is denied, logging a warning if so.
 fn is_denied_env(key: &str) -> bool {
     if ENV_DENYLIST.contains(&key) {
         tracing::warn!(
-            "skipping denied environment variable '{key}' (security: could inject shared libraries)"
+            "skipping denied environment variable '{key}': it is reserved for the daemon \
+             (dynamic-loader/auth hardening or control-plane wiring) and cannot be set by a node"
         );
         true
     } else {
@@ -721,6 +740,33 @@ mod tests {
             None,
             "a daemon still scouting must not disable it for its nodes"
         );
+    }
+
+    /// The daemon injects its control-plane wiring from the authoritative
+    /// node/runtime config before the descriptor's `env:` is applied, so a
+    /// descriptor must not be able to override it via a same-named `env:` entry.
+    /// `is_denied_env` is the guard that drops such an attempt (with a warning)
+    /// in the user-env loops, so pin that every control-plane key is denied
+    /// while an ordinary variable still passes through (#2944).
+    #[test]
+    fn control_plane_env_vars_are_denied() {
+        for key in [
+            "DORA_NODE_CONFIG",
+            "DORA_RUNTIME_CONFIG",
+            DORA_ZENOH_LISTEN_ENV,
+            DORA_ZENOH_CONNECT_ENV,
+            DORA_ZENOH_MULTICAST_ENV,
+        ] {
+            assert!(
+                is_denied_env(key),
+                "control-plane var {key} must not be overridable by a descriptor env:"
+            );
+        }
+        // The loader/auth hardening keys are still denied...
+        assert!(is_denied_env("LD_PRELOAD"));
+        assert!(is_denied_env("DORA_AUTH_TOKEN"));
+        // ...and an ordinary node-defined variable is still allowed through.
+        assert!(!is_denied_env("MY_NODE_SETTING"));
     }
 
     fn plan_for(yaml: &str, daemon: Option<&str>) -> BTreeMap<NodeId, NodeZenohPeering> {


phil-opp:

No issues found. I checked the main risk — that denylisting the control-plane keys might also block the daemon's own injection of them — and it does not: in `binaries/daemon/src/spawn/spawner.rs` the daemon sets `DORA_NODE_CONFIG` / `DORA_RUNTIME_CONFIG` and the zenoh vars directly (via `command.env(...)` and `maybe_inject_zenoh_connect`) after `strip_denied_env`, while `is_denied_env` gates only the descriptor/node `env:` loops. So the daemon's wiring remains authoritative and only a descriptor override is dropped (with the warning). The added test pins each control-plane key as denied while an ordinary var still passes.

---

phil-opp:
> [!NOTE]

The five keys added to `ENV_DENYLIST` (`binaries/daemon/src/spawn/spawner.rs:58-62`) look correct, and the unit test genuinely exercises the guard the spawn loops call (`is_denied_env` → drops a descriptor `env:` entry with a warning, while `MY_NODE_SETTING` passes). One possible gap in **completeness** of the control-plane set:

**`DORA_DAEMON_LOCAL_LISTEN_PORT` is a control-plane wiring var that is *not* in the denylist.** The daemon exports it into its own process environment specifically so children can find the daemon:

- `binaries/cli/src/command/daemon.rs:125-133` — `set_var(DORA_DAEMON_LOCAL_LISTEN_PORT_ENV, …)` with the comment *"Export the listen port so dynamic nodes (and spawned child processes) can discover it via env var."*
- Spawned nodes inherit that value, and the node builder consumes it to pick the daemon address: `apis/rust/node/src/node/mod.rs:2419-2430` (`DORA_DAEMON_LOCAL_LISTEN_PORT` → `daemon_address = (LOCALHOST, port)`).

Because the descriptor `env:` loop (`spawner.rs:419-420`, `626-627`) applies `command.env(key, value)` gated only by `is_denied_env`, a descriptor entry `env: { DORA_DAEMON_LOCAL_LISTEN_PORT: <wrong> }` overrides the inherited value for that node's process. A statically-spawned node itself uses `DORA_NODE_CONFIG` (so it is unaffected), but any *dynamic* child node it launches via `NodeBuilder::node_id(...).build()` then reads the forged port and connects to the wrong daemon — the same "starts cleanly, then exchanges nothing" failure mode this PR closes #2944 against, and the same category (`cluster/config.rs:84-89` already treats `DORA_DAEMON_LOCAL_LISTEN_PORT=0` as a connect-breaking misconfiguration). It seems within scope of "control-plane wiring — owned by the daemon, never by the descriptor" to deny this key too (and pin it in `control_plane_env_vars_are_denied`).

Note this is an edge case — it only bites when a spawned node in turn launches dynamic sub-nodes — so it may be intentionally out of scope; flagging for a conscious decision rather than as a blocker.

Minor, non-blocking: the match is case-sensitive (`ENV_DENYLIST.contains(&key)`), which is correct on Linux where the node reads these keys case-sensitively; worth keeping in mind only if Windows spawn support is ever hardened, since Windows env names are case-insensitive.

---

phil-opp:
already fixed by #2991

phil-opp:
Superseded: #2944 was fixed on main by #2991 (`137208adb`, "stop descriptor `env:` from overriding the daemon's control-plane variables"). Closing as moot — reopen if this draft covers something #2991 missed.

PR #2955, opened 2026-08-02
ci: add `timeout --kill-after` to bounded CI call sites

### LINKED ISSUE #2951: ci: no `timeout` call site uses --kill-after, so every time bound is a request rather than a limit
Found while reviewing #2949. Independent of that PR — a pre-existing robustness gap, filed because that PR made me look.

## What's there today

`grep` for `timeout -k` / `--kill-after` across `.github/workflows/` and `scripts/`: **zero hits**. There are ~45 bare `timeout N …` call sites, ~21 of them wrapping `dora` commands, e.g.:

```
.github/workflows/nightly.yml:385   timeout 300s dora record examples/rust-dataflow/dataflow.yml …
.github/workflows/nightly.yml:718   timeout 15  dora topic echo -d tui-smoke ticker/count --count 5 …
scripts/qa/ci-nightly-jobs.sh:415   timeout 300s dora record … 
scripts/qa/ci-nightly-jobs.sh:425   timeout 120s dora replay "$DREC"
scripts/ros2dev.sh:141              timeout 600 …
```

`timeout N cmd` sends SIGTERM and then **waits indefinitely** for the child to die. Without `-k`, the stated bound is only honored by processes that choose to honor SIGTERM.

## Why it matters here specifically

Several dora entry points install SIGINT handlers via `ctrlc`, and the question of whether SIGTERM should also be caught is live (#2920 asks for it, so that killing `dora run` stops leaking node processes). The moment any of them starts handling SIGTERM, every one of these call sites becomes a soft bound: a wedged teardown turns a `timeout 300s` into a hang that runs to the GitHub Actions job cap instead of failing at 300s with exit 124.

`dora record` is the sharpest example. Its handler (`binaries/cli/src/command/record.rs:391`) is:

```rust
ctrlc::set_handler(move || { let _ = stop_tx.send(()); })
```

No escalation on a repeated signal (unlike the daemon, coordinator and attach handlers, which abort or exit on the 2nd/3rd) — see the companion issue on that. It is wrapped in `timeout 300s` with no `-k` at `nightly.yml:385`.

This is not currently broken — SIGTERM is still fatal for these processes on `main` — but the CI's only defence against a hang is a bound that stops working the instant the product becomes better-behaved about SIGTERM. That is a bad dependency to leave implicit, especially on a repo that has been chasing nightly hangs (#2143, #2152, and the wall-clock diagnostic work).

## Suggested change

Add a kill-after to the call sites that wrap long-running dora commands:

```
-  timeout 300s dora record …
+  timeout -k 30s 300s dora record …
```

30s is enough for a graceful teardown to complete (the daemon's stop ladder is 10s grace → SIGTERM → 5s → SIGKILL) while still guaranteeing the process dies. Worth doing as one sweep so the guarantee is uniform rather than per-site judgement.

Note `timeout -k` is GNU coreutils and BSD/macOS `timeout` supports it too, so this is portable across the runners in use. The `scripts/` sweep matters as much as the workflows, since `qa-nightly` runs those locally.

phil-opp:
## Summary

Fixes #2951.

`timeout N cmd` sends SIGTERM and then **waits indefinitely** for the child to die — without `-k`, the stated bound is only honored by processes that choose to honor SIGTERM. A `grep` for `timeout -k` / `--kill-after` across `.github/workflows/` and `scripts/` previously returned **zero hits** across ~45 bare `timeout` call sites.

### Why it matters

Several dora entry points install SIGINT handlers via `ctrlc`, and whether they should also catch SIGTERM is live (#2920 asks for it so that killing `dora run` stops leaking node processes). The moment any of them starts handling SIGTERM, every one of these call sites becomes a soft bound: a wedged teardown turns a `timeout 300s` into a hang that runs to the GitHub Actions job cap instead of failing at 300s with exit 124. `dora record` is the sharpest example (its handler had no escalation either — see #2952 / #2953). This isn't broken on `main` today — SIGTERM is still fatal — but the CI's only defence against a hang is a bound that stops working the instant the product becomes better-behaved about SIGTERM.

## Change

- Add `-k 30s` to every bounded `timeout` call site across five files:
  - `.github/workflows/nightly.yml` (3)
  - `scripts/qa/ci-nightly-jobs.sh` (40)
  - `scripts/ros2dev.sh` (1)
  - `scripts/ros2-zenoh-interop.sh` (1)
  - `tests/ros2-zenoh/fixtures/run-case.sh` (3) — the callee invoked by `ros2-zenoh-interop.sh:57`, so the bound the caller guarantees is not re-softened one level down. Line 36 there runs a dora Python node, exactly the SIGTERM-handling class this issue is about. (Added after review — thanks @heyong4725.)

  30s clears the daemon's stop ladder (10s grace → SIGTERM → 5s → SIGKILL) while still guaranteeing the process dies. Done as one uniform sweep rather than per-site/per-file judgement, as the issue requested.
- Teach the portable `timeout` fallback shim in `ci-nightly-jobs.sh` (used on hosts without coreutils `timeout`/`gtimeout`) to strip an optional `-k <dur>` prefix, so the kill-after flag isn't left in place and executed as a command.

The substitution used a negative lookbehind so `--timeout`, `gtimeout`, and path substrings like `queue_size_and_timeout_python` are untouched — only the leading coreutils `timeout` command gets the flag.

## Testing

- No bare `timeout <digit>` sites remain **in the five files touched** (all now carry `-k 30s`).
- `bash -n` clean on all four scripts; `nightly.yml` parses as valid YAML.
- Functional test of the fallback shim: it correctly strips both `-k 30s 300s cmd` and `300s cmd` down to `cmd`.

`timeout -k` is GNU coreutils and is supported by BSD/macOS `timeout` too, so it stays portable across the runners in use.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -382,7 +382,7 @@ jobs:
           # #1674 for record). (Historically the workspace root also had
           # a types/ dir that replay wrongly loaded as user types; it moved
           # to libraries/core/types/ but the placement rule stands.)
-          timeout 300s dora record examples/rust-dataflow/dataflow.yml \
+          timeout -k 30s 300s dora record examples/rust-dataflow/dataflow.yml \
             -o examples/rust-dataflow/run.drec
           test -f examples/rust-dataflow/run.drec || { echo "record did not produce run.drec"; exit 1; }
           echo "recording size: $(wc -c < examples/rust-dataflow/run.drec) bytes"
@@ -656,7 +656,7 @@ jobs:
         # update is available. Catches regressions in CLI flag parsing,
         # the self_update crate integration, and the API-response contract.
         run: |
-          out=$(timeout 30 dora self update --check-only 2>&1) || exit_code=$?
+          out=$(timeout -k 30s 30 dora self update --check-only 2>&1) || exit_code=$?
           echo "$out"
           # Must always print the entry banner — if this is missing, the
           # command didn't even start the check path.
@@ -715,7 +715,7 @@ jobs:
           # Tests the CLI-side WebSocket subscription path end-to-end (#238).
           # Fixture emits at 10 Hz, so 5 frames arrive in ~500ms — generous
           # 15s timeout protects against daemon startup delay.
-          timeout 15 dora topic echo -d tui-smoke ticker/count --count 5 --format json \
+          timeout -k 30s 15 dora topic echo -d tui-smoke ticker/count --count 5 --format json \
             2>&1 | tee echo.out
           lines=$(grep -c '"name":"ticker/count"' echo.out || echo 0)
           if [ "$lines" -lt 5 ]; then
--- a/scripts/qa/ci-nightly-jobs.sh
+++ b/scripts/qa/ci-nightly-jobs.sh
@@ -196,7 +196,12 @@ if ! command -v timeout > /dev/null 2>&1; then
     echo "      Jobs will run UNBOUNDED until they complete or you Ctrl-C."
     echo
     timeout() {
-      # Silently drop the duration arg and run the command unbounded.
+      # Silently drop the leading duration args and run the command unbounded.
+      # Handle the optional `-k <kill-after>` prefix the call sites use, so the
+      # kill-after flag isn't left in place and executed as a command.
+      if [ "$1" = "-k" ]; then
+        shift 2
+      fi
       shift
       "$@"
     }
@@ -412,7 +417,7 @@ job_record_replay() {
   # record-replay job (.github/workflows/nightly.yml:261, tuned in
   # #1705 for cold-cache rebuild variance). Replay is normally ~2s;
   # 120s is generous but still bounded.
-  if ! timeout 300s dora record examples/rust-dataflow/dataflow.yml -o "$DREC"; then
+  if ! timeout -k 30s 300s dora record examples/rust-dataflow/dataflow.yml -o "$DREC"; then
     echo "ERROR: dora record failed or exceeded 300s"
     return 1
   fi
@@ -422,7 +427,7 @@ job_record_replay() {
   fi
   echo "OK: recording size: $(wc -c < "$DREC") bytes"
 
-  if ! timeout 120s dora replay "$DREC"; then
+  if ! timeout -k 30s 120s dora replay "$DREC"; then
     echo "ERROR: dora replay failed or exceeded 120s"
     rm -f "$DREC"
     return 1
@@ -1173,7 +1178,7 @@ EOF
   # $WORK/out/.
   echo "=== dora replay $DREC (local single-daemon run) ==="
   rm -rf "$WORK/out"
-  if ! timeout 120s dora replay "$DREC" --speed 0; then
+  if ! timeout -k 30s 120s dora replay "$DREC" --speed 0; then
     echo "ERROR: dora replay failed or exceeded 120s"
     rm -rf "$WORK"
     return 1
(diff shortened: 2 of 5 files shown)


phil-opp:

No issues found. This is a CI-only change. The portable `timeout` fallback shim in `scripts/qa/ci-nightly-jobs.sh` correctly handles the new `-k <dur>` prefix (`shift 2` for the flag, then `shift` for the duration, then run the command), and it still handles the bare `timeout N cmd` form. All converted call sites keep valid argument ordering.

---

heyong4725:
Verified the sweep and the shim — both check out. One file was missed, and it is reachable from a file this PR already touches.

## Verified

- `.github/workflows/nightly.yml` (3), `scripts/qa/ci-nightly-jobs.sh` (40), `scripts/ros2dev.sh` (1), `scripts/ros2-zenoh-interop.sh` (1) — 45 sites, all now `timeout -k 30s`, zero bare `timeout <digit>` left in those four.
- The shim change is right: `if [ "$1" = "-k" ]; then shift 2` strips the prefix before the fallback runs, so `-k` is not executed as a command on hosts without coreutils `timeout`.
- The negative-lookbehind claim holds — `--timeout`, `gtimeout` and `queue_size_and_timeout_python` are untouched.

## Missed: `tests/ros2-zenoh/fixtures/run-case.sh`

Three bare sites remain there:

```
26: timeout 60 "${compose[@]}" exec -T "$SERVICE" bash -lc "... python3 .../peers/$peer"
31: timeout 20 bash -c 'until grep -q ... done' _ "$log" "$peer_pid"
36: timeout 60 "$DORA_ROS2_ZENOH_PYTHON_EXECUTABLE" .../dora_peer.py "$CASE"
```

It is not an unrelated file — `scripts/ros2-zenoh-interop.sh:57` invokes it:

```
"$ROOT/tests/ros2-zenoh/fixtures/run-case.sh" "$DISTRO" "$name" "$PROJECT" "$service"
```

so the sweep covered the caller but not the callee, and the bound the caller now guarantees is re-softened one level down.

Line 36 is the one that matters for #2951's actual argument: it runs a **dora Python node**, which is precisely the class of process that becomes SIGTERM-handling as #2920 lands. Line 26 runs a peer inside the container and plausibly the same. Line 31 is a bare `bash` polling loop — harmless either way, but worth including so the file has no exceptions to reason about later.

Suggest the same `-k 30s` on all three, so the guarantee stays uniform rather than becoming per-file judgement — which is what the issue asked for.

Minor: the PR body's "NONE bare `timeout <digit>` sites remain" reads repo-wide but is scoped to the four files touched. Worth narrowing the wording, or widening the sweep — I would widen it.

phil-opp:
Good catch — fixed in `443a73e`. The three sites in `tests/ros2-zenoh/fixtures/run-case.sh` (26, 31, 36) now carry `-k 30s`, applied with the same lookbehind sweep; the file has no local `timeout` shim, so nothing else needed touching there. `bash -n` clean, and no bare `timeout <digit>` remains in it.

You're right that line 36 is the one that matters for this issue's argument (it runs a dora Python node via `$DORA_ROS2_ZENOH_PYTHON_EXECUTABLE`), and swept it uniformly rather than picking sites.

Also narrowed the PR body: the "no bare sites remain" claim is now explicitly scoped to the files touched (which now includes `run-case.sh`), and the file list is updated. Thanks for widening the sweep by one level.

---

phil-opp:
The new commit `443a73e` extends the `-k 30s` sweep to the three bare `timeout` sites in `tests/ros2-zenoh/fixtures/run-case.sh` (lines 26, 31, 36), which is the callee of the already-swept `scripts/ros2-zenoh-interop.sh`. The change is safe: those sites use coreutils `timeout` directly (no local shim), so `timeout -k 30s <N> ...` is valid ordering, and line 36 (the dora Python node peer) is exactly the SIGTERM-handling process class this PR targets. No issues found in the new commit.

---

---

heyong4725 (approved):
Verified the complete timeout sweep against current main: no executable bare timeout-duration call sites remain under workflows, scripts, or tests. The portable fallback strips the optional -k duration correctly, and a forced-kill probe confirms a SIGTERM-ignoring child is killed after the grace period. Local validation: bash -n, workflow YAML parse, git diff --check, fmt, workspace clippy, cargo-audit/cargo-deny, unwrap budget, and typos all pass.

PR #2957, opened 2026-08-02
feat(cli): add `dora run --exit-when-nodes-finish` (#2920)

heyong4725:
Second half of #2920. The [first half](https://github.com/dora-rs/dora/pull/2949) stopped `dora run` from leaking node processes when it is killed; this half removes the reason people were killing it.

## The problem

A timer input is registered in the daemon's `open_inputs` map like any other input, but it never closes — there is no upstream node that can finish it. So a node consuming `dora/timer/...` never reaches "all inputs closed", is never sent `AllInputsClosed`, and never exits. One timer anywhere makes a whole graph unable to terminate on its own, even after every node doing real work has finished. `dora run` then hangs until you interrupt it.

## The change

`dora run --exit-when-nodes-finish` makes the drain rule count only inputs that *can* close: a node finishes once its **data** inputs have closed, even while its clock keeps ticking.

Off by default, and deliberately so — for a long-lived graph a timer is precisely what keeps it alive. Only `dora run` can set it; a daemon serving a coordinator keeps the default, since such a dataflow is stopped explicitly rather than by running out of work.

Two structural decisions make it work, both learned the hard way (see the review history below):

- the daemon records each node's data inputs **where it registers them**, so the rule sees the same resolved view as `open_inputs` — including `node_inputs()`, which flattens a runtime node's `operators[].config.inputs`;
- both drain sites share one predicate (`is_finished`, and `is_finished_non_source` for the source rule), because a decision spelled out at a call site cannot be unit-tested.

Guards, so the opt-in cannot stop a graph early:

- a node with no data inputs at all — timer-only, logs-only, or no inputs — is never drained, so a pure generator is not stopped the moment it starts;
- a circuit-broken input is recoverable, so it does not count as finished;
- restart is disabled when a node has nothing open **or** is drained, so a node told to finish is not restarted into a finish/exit loop;
- the flag is per-run.

## Public API

`Daemon::run_dataflow` keeps its exact signature. `dora-daemon` is published, so adding a required parameter would break every downstream caller. New options travel in a `#[non_exhaustive]` `RunDataflowOptions` passed to `run_dataflow_with`, so the next option will not break anyone either.

A welcome consequence: this PR no longer touches `tests/fault-tolerance-e2e.rs`, `tests/finish-straggler-e2e.rs`, or the CLI's `daemon.rs`. The 13 call-site edits that an added parameter had forced are gone.

## Review history — what earlier cuts got wrong

Every one of these made the flag cause the very hang it exists to prevent, and **CI was green for all of them.** None is reachable by fmt, clippy, or a test suite that does not exercise operator nodes, restart policies, or circuit breakers.

Round one:

1. **Operator dataflows never drained.** The source check asked the descriptor, whose top-level `inputs` map is empty for a runtime node. Five shipped examples are operator-based.
2. **Source nodes stopped having restart disabled**, so a `restart_policy: always` source respawned forever and its outputs never closed.
3. **The finish-straggler watchdog was vetoed** for exactly the nodes the flag drains.

Round two:

4. **The same operator misclassification survived at the sibling drain site.** Fixing `is_drained` was not enough: `Daemon::subscribe` still consulted the descriptor, so an operator node that finished draining before it subscribed still hung.
5. **The subscribe site lacked the `!has_broken` guard** its sibling had, so a node whose only data input was circuit-broken read as finished and exited for good instead of awaiting recovery.
6. **The watchdog guard keyed on a static property** (`has_data_input`) rather than on having actually drained, arming escalation for healthy timer-fed nodes whose data inputs were still open — a slow upstream plus a long timer interval can read as silent past the grace period and earn a SIGKILL.
7. **`AddMapping` did not record runtime-added edges** in `data_inputs`, so a node could be reported drained while a freshly connected input was still open.
8. **Two tests proved less than they claimed.** The `AddMapping` test reached the invariant through the test module's own helper, which re-implements the recording — deleting the production line left all 157 tests green. The e2e fixture counted every `Event::Input`, including timer ticks, so "received data" was satisfiable by the clock alone.

## Testing

The e2e asserts three things, not one: the graph must **hang without** the flag, **terminate with** it, and the consumer must have **received its data input and exited via `AllInputsClosed`**. Asserting only "exited 0" cannot tell correct draining from draining at startup — both exit cleanly.

Every guard is mutation-checked against the **production** code. That distinction is the lesson of finding 8: mutating a *test* only shows the assertion is sensitive to its fixture. Reverting any one of the guards now fails exactly the test written for it, and the premature-drain mutation reproduces as `record: "0 true"` — drained having received nothing.

The premise run is torn down with SIGTERM rather than `Child::kill()`; SIGKILL is unhandleable and verifiably leaked a node process on every run.

## Docs

`--exit-when-nodes-finish` is listed in both `dora run` flag tables (`docs/cli.md`, `guide/src/operations/cli.md`) with a worked example explaining the timer-never-closes gotcha.

Refs #2920

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2602,6 +2602,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "drain-recording-node"
+version = "1.0.0-rc.4"
+dependencies = [
+ "dora-node-api",
+ "eyre",
+]
+
+[[package]]
 name = "dtoa"
 version = "1.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ members = [
     "tests/fault_tolerance/always_crash_node",
     "tests/fault_tolerance/clean_exit_node",
     "tests/fault_tolerance/delayed_crash_node",
+    "tests/fault_tolerance/drain_recording_node",
     "tests/fault_tolerance/emit_then_exit_source_node",
     "tests/fault_tolerance/event_log_observer_node",
     "tests/fault_tolerance/fire_and_forget_source_node",
--- a/binaries/cli/src/command/daemon.rs
+++ b/binaries/cli/src/command/daemon.rs
@@ -281,6 +281,9 @@ impl Executable for Daemon {
                             dataflow_session.build_id, dataflow_session.local_build, dataflow_session.session_id, false,
                             LogDestination::Tracing, None, None, false, None,
                             descriptor_override,
+                            // `dora daemon --run-dataflow` keeps the default
+                            // completion semantics; the opt-in lives on `dora run`.
+                            false,
                         ).await?;
                         handle_dataflow_result(result, None)
                     }
--- a/binaries/cli/src/command/daemon.rs
+++ b/binaries/cli/src/command/daemon.rs
@@ -281,9 +281,6 @@ impl Executable for Daemon {
                             dataflow_session.build_id, dataflow_session.local_build, dataflow_session.session_id, false,
                             LogDestination::Tracing, None, None, false, None,
                             descriptor_override,
-                            // `dora daemon --run-dataflow` keeps the default
-                            // completion semantics; the opt-in lives on `dora run`.
-                            false,
                         ).await?;
                         handle_dataflow_result(result, None)
                     }
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -22,7 +22,7 @@ use crate::{
 };
 use dora_core::build::LogLevelOrStdout;
 use dora_core::descriptor::{Descriptor, DescriptorExt};
-use dora_daemon::{Daemon, LogDestination, flume};
+use dora_daemon::{Daemon, LogDestination, RunDataflowOptions, flume};
 use eyre::Context;
 use std::{path::PathBuf, time::Duration};
 use tokio::runtime::Builder;
@@ -274,7 +274,7 @@ impl Executable for Run {
         let working_dir_override = self.working_dir.clone();
         let exit_when_nodes_finish = self.exit_when_nodes_finish;
         let handle = rt.spawn(async move {
-            Daemon::run_dataflow(
+            Daemon::run_dataflow_with(
                 &dataflow_path_for_daemon,
                 dataflow_session.build_id,
                 dataflow_session.local_build,
@@ -287,7 +287,7 @@ impl Executable for Run {
                 working_dir_override,
                 // hub-resolved descriptor and/or `--env` merge — see above
                 descriptor_override,
-                exit_when_nodes_finish,
+                RunDataflowOptions::default().exit_when_nodes_finish(exit_when_nodes_finish),
             )
             .await
         });
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -120,6 +120,21 @@ pub struct Run {
     /// numeric-looking values that would be coerced.
     #[clap(long = "env", value_name = "KEY=VALUE")]
     pub env: Vec<String>,
+    /// Exit once every node has finished, treating `dora/timer/...`
+    /// inputs as a clock rather than as work.
+    ///
+    /// By default a node is only told its inputs are closed when ALL of
+    /// them are, and a timer input never closes — so a graph where any
+    /// node consumes a timer cannot finish on its own, even after every
+    /// worker has done its work. With this flag a node is finished once
+    /// its DATA inputs have closed, which makes "run N items and exit"
+    /// scriptable without wrapping the command in a timeout.
+    ///
+    /// Nodes whose inputs are all timers are unaffected: they have no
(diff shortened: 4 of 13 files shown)


phil-opp:

No issues found. The default behavior is preserved exactly (`is_drained` reduces to the previous `open_inputs(...).is_empty()` check when `timers_gate_drain` is on, at both `binaries/daemon/src/lib.rs` call sites), the opt-in correctly treats timer-only and input-less nodes as sources, and the added unit tests plus the bounded e2e test (which asserts the graph hangs *without* the flag before asserting the fix) exercise the change meaningfully.

---

phil-opp:
I re-reviewed after the follow-up commit (`43418ad`) that reworks the drain path for operators, sources, and the watchdog. The rework looks correct in general — `is_drained` collapses to the previous `open_inputs(...).is_empty()` in the default mode at both call sites, the restart-disable OR-condition and the `node_never_finishes` opt-in carve-out are each separately tested, and recording `data_inputs` at registration is the right way to see the flattened operator inputs.

I found one gap, though, on a path the follow-up did not touch:

The subscribe-time drain in the `subscribe` fn (`binaries/daemon/src/lib.rs`, the block around L4482, which reports inputs that were already closed when a node connects) now gates on the drain-aware `is_drained`, but still keeps the descriptor-based source check inside it:

```rust
if dataflow.is_drained(&node_id)
    && let Some(node) = dataflow.descriptor.nodes.iter().find(|n| n.id == node_id)
{
    if node.inputs.is_empty() {
        // do not send AllInputsClosed for source nodes
    } else if send_with_timestamp(&event_sender, NodeEvent::AllInputsClosed, clock).ok() == Some(true)
    { ... }
}
```

For an `operators:` node the top-level `inputs` map is empty (inputs live under `operators[].config.inputs`) — the same fact `operator_nodes_drain_under_the_opt_in` asserts. So under `--exit-when-nodes-finish`, an operator node whose data inputs all close **before** its runtime connects (e.g. a produce-once-and-exit upstream racing a slow-starting operator) reaches this block with `is_drained() == true`, but `node.inputs.is_empty()` is also true, so it is treated as a source and never told `AllInputsClosed`. No later `close_input` fires because the inputs are already closed, so the node hangs — the exact failure this flag exists to prevent. The sibling `signal_all_inputs_closed_if_drained` was correctly changed to drop this guard and rely on `is_drained` alone; the subscribe path was not.

Because this block also has to keep skipping genuine sources in the default mode (where `is_drained` does not distinguish them), the clean fix looks like replacing `node.inputs.is_empty()` with `!dataflow.has_data_input(&node_id)` — the helper this PR already adds — which classifies true sources correctly in both modes while letting operator nodes drain. It's a timing-dependent path, so it may not show up often, but it's worth closing since the whole point of the flag is operator/timer graphs terminating.

Two smaller notes, not blockers:
- The new e2e (`e2e_run_exit_when_nodes_finish_terminates_timer_driven_graph`) and the unit tests are genuinely load-bearing — they parse real YAML, assert the consumer drained *after* receiving input (via the `drain_recording_node` count + `AllInputsClosed` flag), and pin the premise that the graph hangs without the flag. Good coverage. The one uncovered case is the operator subscribe-time path above (the e2e uses only `custom`/`path` nodes).
- `node_never_finishes` keeps its documented pre-existing operator misclassification; you've called it out in a comment as out of scope, which is reasonable.

---
_This is an automated review by Claude (fully automated — no human in the loop). Please treat it as advisory._

---

heyong4725 (commented):
Found one behavioral blocker and one compatibility issue on 43418adc4:

1. [P1] Use the registered input index for the subscribe-time source check (binaries/daemon/src/lib.rs:4470-4475).

This branch correctly teaches is_drained to use data_inputs so runtime operator inputs are visible, but the late-subscribe path immediately reclassifies the node through descriptor.nodes[].inputs. A runtime node keeps inputs under operators[].config.inputs, so the top-level inputs map is empty. If a fast upstream closes before the operator runtime subscribes, is_drained returns true, then line 4473 treats the operator as a source and suppresses AllInputsClosed. No later close_input call can repair this because the input was already closed, so dora run --exit-when-nodes-finish hangs.

I reproduced this on the published head with a disposable unit test that constructs an operators node, records its flattened data input, closes it before Daemon::subscribe, and expects AllInputsClosed. The test fails with try_recv: Empty. The source guard should use !dataflow.has_data_input(&node_id), which shares the same registered view as is_drained and still skips genuine sources in the default mode. Please retain that timing case as a regression test.

2. [P2] Preserve the published Daemon::run_dataflow entry point (binaries/daemon/src/lib.rs:1014-1027).

dora-daemon is a published workspace crate, and this public function now requires a twelfth positional argument. Updating in-repo call sites does not protect downstream Rust callers: every existing call fails to compile after upgrading. Keep the current signature as a false/default wrapper and add an options-bearing sibling (or introduce a backward-compatible options API) for the new completion policy. The feature should not require an unrelated public API break.

Validation: git diff --check passed; all 15 drain-filtered daemon unit tests passed; all 284 dora-cli library tests passed. The new operator test above is the only targeted failure. Current CI format, check, clippy, audit, license, typo, and unwrap jobs are green; the Test and E2E jobs are skipped for this PR head.

PR #2959, opened 2026-08-02
fix(node): prevent init_testing Drop deadlock

### LINKED ISSUE #2855: init_testing(): intermittent deadlock from detached daemon thread in DoraNode::drop()
## Problem

`DoraNode::init_testing()` spawns a daemon simulation thread via `std::thread::spawn` without storing the `JoinHandle` (detached thread). When `DoraNode::drop()` runs, it performs a synchronous request-reply handshake through a bounded tokio mpsc channel (capacity 5). If the daemon thread happens to be inside `next_event()` → `rx.blocking_recv()` (waiting for test input) at that moment, the Drop handshake deadlocks — the daemon is stuck waiting for input, and Drop is stuck waiting for the daemon's reply.

## Deadlock mechanism

Three interacting design choices create the deadlock:

### 1. Detached daemon thread

`apis/rust/node/src/node/mod.rs` (init_with_options, testing branch):

```rust
let (sender, mut receiver) = tokio::sync::mpsc::channel(5); // capacity = 5
let new_communication = DaemonCommunicationWrapper::Testing { channel: sender };
let mut events = IntegrationTestingEvents::new(input, output, options)?;
std::thread::spawn(move || {                          // JoinHandle not stored!
    while let Some((request, reply_sender)) = receiver.blocking_recv() {
        let reply = events.request(&request);
        // ...
    }
});
```

### 2. Blocking request-reply handshake

`apis/rust/node/src/daemon_connection/mod.rs` (request method, IntegrationTestChannel arm):

```rust
DaemonChannel::IntegrationTestChannel(channel) => {
    let (reply_tx, reply) = oneshot::channel();
    channel
        .blocking_send((request.clone(), reply_tx))  // blocks if channel full
        .expect("failed to send request to IntegrationTestChannel");
    reply
        .blocking_recv()                              // blocks until daemon replies
        .context("failed to receive oneshot reply")
}
```

### 3. The deadlock

```
Main thread                           Daemon thread
-----------                           -------------
test ends, harness dropped
  DoraNode::drop()
    report_closed_outputs()
      channel.blocking_send() ----->  ...but daemon is stuck in:
                                        next_event()
                                          rx.blocking_recv()
                                          (waiting for test input
                                           that will never arrive)
      reply.blocking_recv() ------->  NEVER REPLIES -> DEADLOCK
```

The daemon is blocked on the **input channel** (waiting for next test event), while the main thread is blocked on the **daemon communication channel** (waiting for daemon to process CloseOutputs). Neither can make progress.

### 4. Why it's intermittent

The deadlock only triggers if the daemon thread is inside `next_event()` when Drop runs. This depends on thread scheduling timing — rare on fast machines (~10% locally) but common on CI runners with limited CPU cores.

## Impact

- Any downstream user of `init_testing()` can hit this
- CI pipelines using `init_testing()` suffer random 30-minute timeouts
- Workaround `--test-threads=1` reduces but doesn't eliminate the problem
- Adding `timeout` wrappers in CI limits blast radius but doesn't fix root cause

Real-world example: a 30-minute CI timeout at https://github.com/dora-rs/gsoc2026-dora-test-utils/actions/runs/30205472256

## Suggested fix

**Option A (minimal):** Store the `JoinHandle` returned by `std::thread::spawn` in `DoraNode`, and join it in Drop *before* sending CloseOutputs/OutputsDone. Also close the input channel first to unblock the daemon from `next_event()`.

**Option B (defense-in-depth):** Increase channel capacity from 5 to a larger value (e.g. 256) and/or change `blocking_send` to `try_send` with a timeout in the Drop path.

**Option C (both):** Apply both A and B for robustness.

## Environment

- DORA version: v1.0.0-rc.4 (commit 0f4082f7)
- Also reproduced on 45436aad (v1.0.0-rc.1 + 212 commits)

trippyogi:
﻿## Summary
- Fix intermittent `init_testing()` deadlock in `DoraNode::drop()` when the testing daemon thread is blocked in `next_event()`.
- Store the daemon `JoinHandle`, interrupt scheduled waits on shutdown, bump the request channel capacity, and join cleanly on Drop.
- Add a regression test that Drop during a long scheduled wait returns promptly.

Fixes #2855

## Test plan
- [x] `cargo test -p dora-node-api`
- [x] New test `init_testing_drop_during_scheduled_wait_does_not_hang`
- [x] `cargo fmt --check`
- [ ] CI green on this PR


trippyogi:
## Follow-up: node-first Drop no longer hits the 2s join timeout

Addressing review feedback on the testing-daemon teardown path for #2855.

### Approach
- After answering `OutputsDone` while the shutdown flag is set, the testing-daemon request loop now `break`s even if `EventStream` still holds a `DaemonChannel` sender clone (drops the receiver so later EventStream close sends fail cleanly).
- Removed the helper-thread + `TESTING_DAEMON_JOIN_TIMEOUT` join workaround; `DoraNode::drop` joins the daemon thread directly.
- Kept interruptible scheduled sleeps in `next_event`.
- Added regression coverage for **both** destruction orders with a `< 500ms` budget (well below the old 2s internal timeout).

### Test results
`cargo test -p dora-node-api` — all passed.
Drop-order filter (serial): **0.11s** total for both tests (includes ~50ms setup sleep each), so both paths stay comfortably under 500ms:
- `init_testing_drop_events_then_node_during_scheduled_wait_does_not_hang`
- `init_testing_drop_node_then_events_during_scheduled_wait_does_not_hang`

Commit: `b865ee50`

phil-opp:
> [!NOTE]

No issues found.

I reviewed the deadlock fix against both destruction orders and the daemon lock/blocking semantics:

- The single `Arc<AtomicBool>` shutdown flag is cloned consistently into `IntegrationTestingEvents`, `EventStream.testing_shutdown`, and `DoraNode.testing_shutdown` — all the same cell.
- Both `Drop` impls set the flag *before* their blocking handshake, so an in-flight `next_event` sleep is interrupted regardless of which of node/event-stream drops first. `Ordering::Relaxed` is fine here: the mpsc+oneshot channel provides the happens-before edge for the `OutputsDone`-triggered break, and the ≤10ms sleep-slice loop observes the store promptly for the interrupt path.
- The `outputs_done && shutdown` loop break correctly lets the daemon exit while `EventStream` still holds a sender clone; dropping the receiver is what makes the later `EventStream` close send fail cleanly, matching the `expect(...)` → `.context(...)?` change in `daemon_connection/mod.rs`.
- Removing the join timeout is safe: the break is guaranteed once `OutputsDone` is delivered, and if that send fails the daemon has already exited, so `join()` cannot hang indefinitely.
- The two regression tests are meaningful — the 30s scheduled event guarantees the daemon is parked in `next_event`'s sleep, so an un-fixed build would blow far past the 500ms budget in both drop orders; there's no vacuous-pass path that would hide a regression.

---

phil-opp:

The Drop-deadlock fix itself looks correct: a single shared `Arc<AtomicBool>` shutdown flag, cloned into `IntegrationTestingEvents`/`EventStream`/`DoraNode` and set before each blocking handshake, lets the sliced `next_event` sleep abort regardless of drop order, and both regression tests are non-vacuous (the 30s scheduled event guarantees the daemon is parked in the sleep, so an unfixed build blows past the 500ms budget in either drop order).

One cross-PR concern: the new `init_testing_node_mid_scheduled_wait` test constructs `TestingOutput::ToChannel(tx)` from `flume::unbounded()`, while the open PR #2972 retypes `ToChannel` to `tokio::sync::mpsc::UnboundedSender`. Whichever merges second won't compile until this test is updated (both PRs also touch the same three files: `node_integration_testing.rs`, `event_stream/mod.rs`, `node/mod.rs`). Worth coordinating the merge order with #2972.

---

phil-opp:
> 🤖 **Automated review by Claude** — this review was generated fully automatically, with no human in the loop.

The latest commit (`401fdf4`, "use tokio ToChannel in init_testing drop regression") addresses the cross-PR concern raised earlier: the `init_testing_node_mid_scheduled_wait` helper now builds its output pair via `crate::integration_testing::unbounded_channel()` and `TestingOutput::ToChannel(tx)`, matching the `tokio::sync::mpsc::UnboundedSender<OutputJson>` API that `ToChannel` now carries on `main`. That type and re-export exist (`apis/rust/node/src/integration_testing.rs:173`, `:295`), so the regression test compiles against the current API and the earlier flume/tokio mismatch is resolved.

The change is test-only and does not touch the deadlock-fix logic. The core fix (single shared `Arc<AtomicBool>` shutdown flag set before each blocking handshake, sliced interruptible `next_event` sleep, and the `OutputsDone && shutdown` loop break for node-first Drop) is unchanged from the previously-reviewed diff. The latest commit looks safe to merge; the only remaining consideration is merge ordering relative to #2972, which is moot now that its `ToChannel` retype is present on `main`.

---

PR #2960, opened 2026-08-02
fix(core): use non-verbatim /_schema key for zenoh liveliness

### LINKED ISSUE #2923: Noise: zenoh_ext "Received malformed liveliness token key expression" warnings flood node logs
**Versions:** dora-cli 1.0.0-rc.4; API from git rev `7eb4a5f8b`; zenoh via bundled deps.

Every node subscribing to another node's output logs warnings like:

```
WARN zenoh_ext::advanced_subscriber AdvancedSubscriber{}: Received malformed liveliness
token key expression: dora/default/<uuid>/output/<node>/<topic>/@schema/@adv/pub/<hash>/39/_
```

one per (subscriber, publisher) pair at startup — a 9-node graph produces dozens, and
with reconnects/dynamic nodes they recur. They appear harmless (dataflows work), but they
dominate log tails and hide real warnings. Ask: either emit liveliness tokens in the
format `zenoh_ext` expects, or downgrade/silence this specific mismatch if it is expected
with dora's key layout.

trippyogi:
﻿## Summary
- Zenoh AdvancedPublisher liveliness tokens under `/@schema/@adv/...` failed zenoh_ext parsing because `@` chunks are verbatim and `**` cannot cross them, flooding WARN logs.
- Rename the schema side-channel key from `/@schema` to `/_schema` so liveliness tokens parse correctly.
- Add unit coverage that the schema topic has no verbatim `@` chunks, plus publisher-after-subscriber schema cache test.

Fixes #2923

## Compatibility note
Mixed-version graphs (old `/@schema` publishers + new `/_schema` subscribers, or vice versa) will not share the schema side-channel until both sides upgrade. Data-plane topics are unchanged.

## Test plan
- [x] `cargo test -p dora-core schema_topic_has_no_verbatim_chunks --features zenoh`
- [x] `cargo test -p dora-node-api --test zenoh_schema_cache`
- [x] `cargo fmt --all`
- [ ] CI green on this PR

### DIFF
--- a/apis/rust/node/tests/zenoh_schema_cache.rs
+++ b/apis/rust/node/tests/zenoh_schema_cache.rs
@@ -59,10 +59,13 @@ fn free_endpoint() -> String {
     format!("tcp/127.0.0.1:{port}")
 }
 
-/// Schema side-channel key. Must not end in an `@…` verbatim chunk: zenoh_ext
+/// Schema side-channel key. Lives under the dedicated `/schema/` plane (not
+/// under `/output/…`) and must not contain `@…` verbatim chunks: zenoh_ext
 /// liveliness tokens are `${remaining:**}/@adv/...` and `**` cannot cross
-/// verbatim chunks (#2923). Mirrors `dora_core::topics::zenoh_output_schema_topic`.
-const SCHEMA_KEY: &str = "dora/test/schema-cache/output/_schema";
+/// verbatim chunks (#2923). Layout mirrors
+/// `dora_core::topics::zenoh_output_schema_topic`
+/// (`…/schema/{node}/{hex(output)}`; hex("out") == `6f7574`).
+const SCHEMA_KEY: &str = "dora/test/schema-cache/schema/node/6f7574";
 
 #[test]
 fn advanced_pub_cache_serves_late_joining_subscriber() {
@@ -117,8 +120,8 @@ fn advanced_pub_cache_serves_late_joining_subscriber() {
 
 /// Subscriber is up first; the publisher appears later. Recovery requires
 /// `detect_late_publishers` to parse the publisher's liveliness token
-/// (`…/_schema/@adv/pub/…`). With the old `/@schema` key that parse failed
-/// (#2923) and this path silently never queried history.
+/// (`…/schema/{node}/{hex}/@adv/pub/…`). With the old `/@schema` key that
+/// parse failed (#2923) and this path silently never queried history.
 #[test]
 fn advanced_pub_cache_serves_when_publisher_joins_after_subscriber() {
     let endpoint = free_endpoint();
--- a/apis/rust/node/tests/zenoh_schema_cache.rs
+++ b/apis/rust/node/tests/zenoh_schema_cache.rs
@@ -6,6 +6,10 @@
 //! SequenceNumber sequencing via `sample_miss_detection` so no session
 //! timestamping is needed, `publisher_detection` + `detect_late_publishers`).
 //!
+//! Also covers the complementary case (publisher joins *after* the subscriber):
+//! that path depends on liveliness-token parsing, which fails when the schema
+//! key ends in an `@…` verbatim chunk (#2923).
+//!
 //! Runs in the default suite: it forms a loopback peer link on an OS-assigned
 //! port (a transient `TcpListener` bind reserves a free port, so there is no
 //! fixed-port clash) with multicast disabled. The cold-start path is covered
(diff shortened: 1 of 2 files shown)


trippyogi:
## Follow-up: schema namespace redesign

Review feedback was right: nesting `/_schema` under the user-controlled output path collided with a valid nested `DataId` (`cmd` schema vs `cmd/_schema` data) and made schema traffic visible to wildcard data subscriptions. `@` was previously safe only because it is forbidden in `DataId`; `_` is not.

**New layout** (outside `output/{node}/{data_id}/…`):

```text
dora/{network}/{dataflow_id}/schema/{node_id}/{hex(output_id)}
```

- Hex-encodes the output id into a single chunk (same approach as the old #2666 readiness keys), so `/` in DataIds cannot reshape the key.
- No verbatim `@` chunks before `/@adv/…`, so AdvancedPublisher `publisher_detection` liveliness still parses (#2923 WARN flood stays fixed).
- Schema plane is not under `/output/`, so wildcard data subscribers do not see schema samples.

Regression coverage: `schema_topic_does_not_collide_with_nested_data_id` asserts schema(`cmd`) ≠ data(`cmd/_schema`). Leaving this PR as draft.

phil-opp:
> [!NOTE]

No important issues found. The rename moves the schema side-channel off the user-controlled output path to `dora/{network}/{dataflow}/schema/{node}/{hex(output)}`, and both the producer (`apis/rust/node/src/node/mod.rs:2847`) and all consumers (`apis/rust/node/src/event_stream/mod.rs:1516`, `binaries/daemon/src/lib.rs:3376`) route through the single `zenoh_output_schema_topic` helper, so there is no intra-version key mismatch. The mixed-version incompatibility is real but correctly documented in the PR body. The two unit tests (`schema_topic_has_no_verbatim_chunks`, `schema_topic_does_not_collide_with_nested_data_id`) exercise the changed function and guard the actual regressions.

Two minor observations (non-blocking):

- The new integration test `advanced_pub_cache_serves_when_publisher_joins_after_subscriber` asserts the late-publisher path against a **hardcoded** `SCHEMA_KEY` constant rather than a key derived from `zenoh_output_schema_topic`. It faithfully characterizes the zenoh-ext liveliness behavior, but it would not fail if the production key function regressed back to a verbatim (`@…`) chunk — that guarantee rests entirely on `schema_topic_has_no_verbatim_chunks`. The combined coverage is adequate; just noting the integration test is decoupled from the function under change.
- Numerous doc comments across `node/mod.rs`, `event_stream/mod.rs`, and `daemon/src/lib.rs` still refer to the "`@schema` subtopic", which is now stale since the key no longer contains `@schema`. Purely cosmetic.

---

heyong4725:
Reviewed and verified the mechanism independently. This is the right fix.

## Root cause confirmed

Probed zenoh's matching directly, since the warning itself is hard to reproduce (`zenoh_ext` logs through the *node's* tracing subscriber, so a fixture that installs none shows nothing regardless of the bug):

```
KeyExpr("**/@adv/pub/abc/39/_").intersects(...)

  dora/default/0/output/n/t/@schema/@adv/pub/abc/39/_   ->  false
  dora/default/0/schema/n/6f7574/@adv/pub/abc/39/_      ->  true
```

Exactly as described: `@`-prefixed chunks are verbatim, `**` cannot cross them, so every liveliness token under `/@schema/` failed `ke_liveliness::parse` — one per (subscriber, publisher) pair, which is the flood in #2923. The new key parses.

## Two things I checked that could have gone wrong

**`@ack` is not also affected.** `zenoh_output_ack_topic` still ends in a verbatim `/@ack`, but it uses a plain **exact-key** subscriber rather than an `AdvancedPublisher`, so it declares no liveliness tokens and emits no warnings. Correctly left alone — and its `@` is still doing useful work.

**Dropping the `@` does not cause cross-delivery.** The old doc comment noted the `@` also kept the key from matching wildcard subscriptions on the data namespace. That protection is not lost: every `declare_subscriber` in the tree uses a concrete key, and the new key lives outside `/output/` entirely. Moving to a dedicated `/schema/` plane is a stronger guarantee than the `@` was, and the hex-encoded output id plus `schema_topic_does_not_collide_with_nested_data_id` close the `cmd/_schema` DataId collision that a plain rename would have opened.

## One note for next time

The PR body is stale: it still describes renaming `/@schema` to `/_schema`, and a compatibility note about mixed-version graphs sharing the side-channel. The implementation has moved on to the `/schema/` plane, which is the better fix. Since a squash merge takes the PR body as the commit message, I have substituted an accurate description at merge rather than record the outdated one in history.

Thanks — good catch and a clean fix.

PR #2961, opened 2026-08-02
test(e2e): guard that SIGTERM to `dora run` tears the dataflow down

heyong4725:
Test-only. Covers a gap in the second half of #2920: killing `dora run` already tears its nodes down, but nothing verified it.

## What already works

The teardown chain exists, assembled from several pieces that all have to stay in step:

| Piece | Where |
|---|---|
| Nodes spawned as process-group leaders | `spawn/prepared.rs:587` |
| `ProcessOperation::Kill` is a `killpg` over the whole group | via `process_wrap` |
| Group reaped before the node reports finished | `spawn/prepared.rs:761` → `782` |
| `run_dataflow` returns only once every node reported | `lib.rs:1680` / `1827` |
| SIGTERM/SIGHUP routed into that chain | #2949 |

Measured on a node that ignores both `Stop` and SIGTERM: `dora run` exits after ~15s (`grace + grace/2`), node gone, nothing stranded.

## What this adds

A regression test, plus a fixture that honors **neither** the cooperative `Stop` **nor** SIGTERM — so only the ladder's final SIGKILL rung can end it. A node that dies to SIGTERM gets cleaned up whether or not that rung ever runs, and so cannot tell a working ladder from a broken one.

## What it catches, and how I know

**A teardown wedge, not an orphan.** Because the CLI waits for every node to report, it *cannot* exit while a node is still alive — so broken teardown surfaces as a run that never returns, which is the original shape of the complaint in #2920.

Both directions are checked:

- clean tree → passes
- **both** kill submissions disabled → fails on the exit deadline
- **only** the ladder's SIGKILL rung disabled → still passes

That last line is stated rather than hidden, because someone will try it. `stop_all` moves the `ProcessHandle`s into the ladder task (`n.process.take()`), so when that task ends they drop and `ProcessHandle::drop` submits one more `Kill`. It is a second kill in the **same task**, not an independent path — so a single-rung mutation is uninformative here. An earlier revision of this PR described that as "defence in depth"; it isn't, and the code comment now says so.

## One thing worth knowing before editing the fixture

Its 300s lifetime is **load-bearing**. Shortening it so a stranded process self-limits looks tidier and silently guts the test: once the node can exit on its own inside the deadline, a wedged teardown completes anyway and the mutation that should fail passes. An earlier revision of this PR did exactly that, and the mutation went green.

Cleanup is the harness's job instead — an RAII guard kills the CLI and the node's process group on every exit path, so no failure strands a process the suite cannot signal away.

## Robustness

- `--target-dir` pinned on the fixture build, matching the sibling helpers; the descriptor hard-codes the workspace-relative path, so a `CARGO_TARGET_DIR` override would otherwise build it where the daemon never looks and fail 60s later with the wrong diagnosis.
- `kill` and `libc::signal` results are checked: an undelivered signal would blame production for a wedge that never happened, and a failed `SIG_IGN` install would leave a node that dies to SIGTERM — still green, but never exercising the rung under test.
- the surviving-node check confirms identity with `ps -o comm=`, so a recycled pid cannot report a false orphan or get an unrelated process SIGKILLed.
- deadline 120s → 60s, so a regression that merely inflates teardown is caught rather than passing silently.
- stderr captured and tailed into every failure message; the wedge this guards would otherwise be reported with no evidence on a remote runner.
- takes `LIFECYCLE_LOCK` like every sibling, sweeps stale artifacts by prefix, writes the pid file atomically, quotes interpolated paths.

## Scope

Marginal CI cost is ~18-20s, and the job that runs this file is merge-queue-only, so PR turnaround is unaffected.

A genuine orphan needs the CLI itself to be SIGKILLed, which is unhandleable in-process and would need parent-death signals (`PR_SET_PDEATHSIG` on Linux, a kqueue watcher on macOS). Deliberately out of scope; the surviving-node assertion is kept as a cheap backstop.

No production change.

Refs #2920

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7127,6 +7127,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "sigterm-ignoring-node"
+version = "1.0.0-rc.4"
+dependencies = [
+ "dora-node-api",
+ "eyre",
+ "libc",
+]
+
+[[package]]
 name = "silent-source-node"
 version = "1.0.0-rc.4"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,6 +81,7 @@ members = [
     "tests/fault_tolerance/hang_after_init_node",
     "tests/fault_tolerance/input_closed_observer_node",
     "tests/fault_tolerance/reconnect_survivor_node",
+    "tests/fault_tolerance/sigterm_ignoring_node",
     "tests/fault_tolerance/silent_source_node",
     "tests/fault_tolerance/stop_delay_node",
     "tests/fault_tolerance/timed_burst_source_node",
--- a/tests/fault_tolerance/sigterm_ignoring_node/Cargo.toml
+++ b/tests/fault_tolerance/sigterm_ignoring_node/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "sigterm-ignoring-node"
+edition.workspace = true
+rust-version.workspace = true
+version.workspace = true
+description.workspace = true
+documentation.workspace = true
+license.workspace = true
+publish = false
+
+# Test-only node that honors NEITHER the cooperative `Stop` nor SIGTERM,
+# so the daemon's only remaining lever is SIGKILL. Used to prove the
+# stop ladder actually reaches its hard-kill rung before the CLI exits
+# (dora-rs/dora#2920).
+
+[[bin]]
+name = "sigterm-ignoring-node"
+path = "src/main.rs"
+
+[dependencies]
+dora-node-api = { workspace = true, default-features = false }
+eyre = { workspace = true }
+libc = "0.2"
--- a/tests/fault_tolerance/sigterm_ignoring_node/src/main.rs
+++ b/tests/fault_tolerance/sigterm_ignoring_node/src/main.rs
@@ -1,0 +1,59 @@
+//! A node that can only be stopped by SIGKILL.
+//!
+//! It ignores SIGTERM and never polls its event stream, so neither the
+//! cooperative `Stop` event nor the daemon's SIGTERM rung can end it.
+//! That makes it the only way to observe whether the stop ladder
+//! actually reaches its final SIGKILL rung: a node that dies to SIGTERM
+//! would be cleaned up either way and cannot distinguish the two
+//! (dora-rs/dora#2920).
+
+use dora_node_api::DoraNode;
+use eyre::ensure;
+use std::{thread, time::Duration};
+
+fn main() -> eyre::Result<()> {
+    // Installed BEFORE `init_from_env`, which spawns the event-stream
+    // threads, the node's tokio runtime and zenoh's threads: `signal` is
+    // only well specified in a single-threaded process.
+    //
+    // SAFETY: setting a disposition to `SIG_IGN` runs no handler, and no
+    // other thread exists yet to observe the change.
+    let previous = unsafe { libc::signal(libc::SIGTERM, libc::SIG_IGN) };
+    // Checked, because a silent failure here would leave a node that
+    // dies to SIGTERM — which still passes the test, but via the wrong
+    // rung, quietly costing the fixture the only property it has.
+    ensure!(
+        previous != libc::SIG_ERR,
+        "failed to ignore SIGTERM: {}",
+        std::io::Error::last_os_error()
+    );
+
+    let (_node, _events) = DoraNode::init_from_env()?;
+
+    // Publish our pid so the test can assert on THIS process rather than
+    // pattern-matching the process table, which would also catch a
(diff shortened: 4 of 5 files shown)


phil-opp:
> [!NOTE]

Test-only PR (new `sigterm-ignoring-node` fixture + one Unix e2e test). I reviewed it for correctness and, importantly, whether the test is meaningful. It holds up well:

- **The test is honest about what it catches.** The PR body and commit message state plainly that disabling only the SIGKILL rung does *not* fail the test (kill-on-drop still reaps), and that the real signal is a teardown *wedge* surfacing as a run that never returns. That matches the code: `run.try_wait()` returns `Some` only after `run_dataflow` has waited for every node to be reaped, so the post-exit liveness check on the fixture's own pid has correct ordering. Not tautological.
- **`LIFECYCLE_LOCK` omission is correct here** (not a bug): every other test in this file takes the lock, but those drive `dora node`/the coordinator on port 6013. `dora run` is explicitly isolated and binds no coordinator port (`binaries/cli/src/command/run.rs:1-11`), so there's no shared state to serialize.
- **Fixture and cleanup look sound**: asserts on the pid the node writes (not process-table pattern matching), `process_alive` treats zombies as dead, 120s bound vs 15s ladder gives margin, and the orphan branch kills the survivor so it can't poison later tests.

One minor, non-blocking observation:

- `tests/node-lifecycle-e2e.rs:1601` builds the fixture with `cargo build -p sigterm-ignoring-node` (default target dir) but derives the node path from `CARGO_MANIFEST_DIR/target` (`:1607`), ignoring `CARGO_TARGET_DIR`. The file's own `dora_bin()` (`:40-42`) deliberately honors `CARGO_TARGET_DIR` (per #1701). In a checkout with `CARGO_TARGET_DIR` set, the spawn path won't match the build output and the test would fail at the "node never reported its pid" deadline rather than for a real regression. Aligning it with the `dora_bin()` pattern would remove that environment-specific footgun.

No correctness or resource issues found in the changed code.

---

heyong4725 (commented):
[P1] Disarm cleanup before signalling stale PIDs

The success path reaps `dora run` through `try_wait()` and then proves `node_is_fixture(node_pid)` is false, but both numeric identifiers remain populated when this `Drop` runs. It therefore sends SIGKILL to the already-reaped CLI PID, the negative node PID, and the node PID. At that point neither identifier is owned by this test. If either PID is recycled in the gap, a passing test can kill an unrelated process; the negative PID can kill its entire process group. The identity check at line 1745 does not protect this destructor. The same hazard exists on the precondition panic if the fixture exits after writing its PID.

Please make both tracked PIDs optional and disarm the CLI as soon as `try_wait()` reaps it, then disarm the node after it is confirmed gone. Before any cleanup group kill, re-confirm that the PID still belongs to this fixture; never signal a stale bare PID. The targeted test passed locally, and its cleanup printed `No such process` for all three signals, confirming that the success path currently attempts these stale kills.

phil-opp:
The latest revision addresses the earlier feedback (the fixture build now pins `--target-dir`, the exit deadline is bounded, the `kill`/`libc::signal` results are checked, and stderr is captured to a log). Reviewing the current diff, I found one new correctness issue:

**`node_is_fixture()` / `still_our_fixture()` will never match on Linux.** Both identify the fixture with `ps -o comm=` and `.contains("sigterm-ignoring-node")`. On Linux a process's `comm` (`/proc/<pid>/comm`, which is what `ps -o comm=` reports) is truncated to 15 characters (`TASK_COMM_LEN` − 1), but `sigterm-ignoring-node` is 21 characters — so `comm` is `sigterm-ignorin` and the `.contains("sigterm-ignoring-node")` check is always `false`. Verified locally:

```
$ cp /bin/sleep sigterm-ignoring-node && ./sigterm-ignoring-node 5 &
$ ps -o comm= -p <pid>
sigterm-ignorin
```

Two consequences on the (Linux) e2e runner:

1. The pre-signal guard `assert!(node_is_fixture(node_pid), "precondition: ...")` fails, so the test aborts before it ever drives the stop ladder — it cannot pass. The final `assert!(!node_is_fixture(...))` also becomes vacuously true.
2. `still_our_fixture()` in the `Teardown` drop also always returns `false`, so a genuinely wedged/stranded fixture would never be SIGKILLed and would leak its process group into the rest of the suite — the exact leak the RAII guard is meant to prevent.

(Note this is Linux-specific: macOS `ps -o comm=` reports the full executable path, so the substring match happens to work there — which can mask the problem in local macOS testing.)

Simplest fixes: rename the fixture binary to ≤15 chars, match on the truncated `comm`, or switch the identity check to `ps -o args=` / `/proc/<pid>/cmdline`, neither of which is truncated.

---
*🤖 This is a fully automated review by Claude (Claude Code). No human has reviewed these findings — please verify before acting. Generated by [Claude Code](https://claude.ai/code)*

---

PR #2962, opened 2026-08-02
fix(cli): recover from stale coordinator store

### LINKED ISSUE #2921: UX: `dora up` fails on stale `coordinator.redb` schema — offer auto-recreate
**Versions:** dora-cli 1.0.0-rc.4 (git rev `7eb4a5f8b`), upgrading from an older local install.

`dora up` prints "started dora coordinator" but the coordinator dies immediately when
`~/.dora/coordinator.redb` predates the current schema:

```
redb schema version mismatch: database at `/Users/<u>/.dora/coordinator.redb` has v2,
but this binary expects v3. Delete the file and restart to create a fresh database,
or use `--store memory` to bypass persistence.
```

The message is good — but `dora up` itself only reports a connect timeout, so the real
error is buried in `out/dora-coordinator.txt`, and every user who upgrades across the
schema bump hits it. Ask: surface the coordinator's startup error in `dora up`'s output,
and/or offer `dora up --recreate-store` (or auto-recreate with a warning, since the file
is a cache).

heyong4725:
## Summary

- capture coordinator stderr during `dora up` startup and report early process exits directly
- add `dora up --recreate-store` to archive the default redb store before starting fresh
- serialize recreation with an OS file lock and recheck coordinator state after acquiring it, preventing overlapping commands from overwriting the original backup
- keep the recreate hint in the default-store `dora up` path so custom `redb:/path` backends receive valid recovery guidance
- preserve prior stores as `coordinator.redb.backup` or a numbered variant instead of deleting persisted state
- document the recovery flow and add unit and full-stack regression coverage

Automatic deletion is intentionally avoided because the coordinator store contains persisted dataflow, build, and daemon records.

## Validation

- `cargo check --examples`
- `cargo test -p dora-cli --lib coordinator_store_backup_path_skips_existing_backups`
- targeted `ws-cli-e2e` coverage for stale-store reporting, recreation, and concurrent recreation

Related: #2892
Closes #2921

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2150,6 +2150,7 @@ dependencies = [
  "eyre",
  "futures",
  "libloading 0.9.0",
+ "redb",
  "serde_json",
  "serde_yaml",
  "tempfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1947,6 +1947,7 @@ dependencies = [
  "duration-str",
  "env_logger",
  "eyre",
+ "fs2",
  "futures",
  "git2",
  "inquire",
@@ -2148,6 +2149,7 @@ dependencies = [
  "dora-tracing",
  "dunce",
  "eyre",
+ "fs2",
  "futures",
  "libloading 0.9.0",
  "redb",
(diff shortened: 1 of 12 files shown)


phil-opp:

I reviewed the diff. The recovery logic itself looks correct: staleness detection matches the real error string (`redb_store.rs:91` emits `"redb schema version mismatch: ..."`, and the hint check in `wait_for_coordinator_start` looks for exactly that), archiving is only triggered by the explicit `--recreate-store` flag (so no valid state is auto-discarded), there's no unbounded loop, and the file-lock + post-lock re-check in `up()` correctly serializes concurrent recreations and preserves the original backup.

One resource concern worth a look:

- `binaries/cli/src/command/up.rs` (`start_coordinator` / `detach_process_with_stderr`): the coordinator's stderr is now redirected to `tempfile::tempfile()` (an unlinked temp file). The detached coordinator keeps that fd open for its **entire lifetime**, but the file is only ever read during the ~10s startup window and is never truncated or closed afterward. Previously stderr went to `Stdio::null()`. So all coordinator stderr/tracing output now accumulates in an anonymous temp file for the life of a long-running coordinator — and on Linux `tempfile` typically lands in tmpfs (RAM), so this grows unbounded in memory for busy/long-lived deployments. This affects every `dora up` that starts a coordinator, not just `--recreate-store`. Consider closing/replacing the child's stderr once readiness is confirmed, or capturing only a bounded prefix.

---

Issue #2031, opened 2026-06-04
config: sub-millisecond (>1kHz) timer interval corrupts to a 0ms busy-loop on wire round-trip


heyong4725:
## Summary

A sub-millisecond timer interval (`> 1 kHz`) silently corrupts to a 0 ms busy-loop after one descriptor round-trip on the wire.

`libraries/message/src/config.rs:364-371` — the `Display` impl emits `secs/0` for any interval `< 1 ms`, because both `as_secs() == 0` and `subsec_millis() == 0`. Descriptors are serialized/round-tripped between CLI/coordinator/daemon, so a configured high-rate timer (e.g. `dora/timer/hz/3000`) comes back as `dora/timer/secs/0` → a 0 ms interval → a busy-loop pinning a CPU on the daemon.

Note: the **parse-side** panic for a pathologically tiny `hz` was fixed in #2027 (`try_from_secs_f64`). This issue is the separate **Display round-trip corruption** for valid sub-millisecond intervals, still open.

## Suggested fix

In the `Display` impl, emit `millis` only when the interval is a whole number of milliseconds; otherwise emit microseconds or nanoseconds (and add the matching unit(s) to the `FromStr` parser). Test: `"dora/timer/hz/3000"` survives `Display` → `parse` with the same `Duration`.

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3). Adversarially re-verified against source._

PR #2963, opened 2026-08-02
fix(core): bound the module-file read to prevent unbounded reads

phil-opp:
## Issue

`load_module_file` (`libraries/core/src/descriptor/expand.rs`) guarded oversized module files with a `std::fs::metadata().len()` pre-check, then read the file with an **unbounded** `std::fs::read`:

```rust
let metadata = std::fs::metadata(path)?;
if metadata.len() > MAX_MODULE_FILE_SIZE { bail!(...) }
let buf = std::fs::read(path)?; // unbounded
```

The `MAX_MODULE_FILE_SIZE` doc comment states this *"prevents DoS from huge or infinite files"*, but the metadata pre-check does not achieve that:

- **FIFOs / character devices** (e.g. `/dev/zero`) report a length of `0`, so the check passes and the unbounded read then never terminates — exactly the "infinite files" case the limit claims to prevent.
- The check is **racy** (TOCTOU): a file appended to between the `stat` and the `read` bypasses the cap.

The sibling manifest loader (`libraries/core/src/manifest/mod.rs::read`) was already hardened against this class of bug, with a comment noting *"a size check on metadata alone would miss FIFOs/devices and growing files."*

## Fix

Bound the read itself with `take(MAX_MODULE_FILE_SIZE + 1)` and reject when the cap is hit, mirroring the manifest loader. This drops the separate `stat` syscall (a single open+read instead of stat-then-read).

## Validation

- Added regression tests for the just-over-cap (rejected) and exactly-at-cap (accepted) boundaries.
- `cargo test -p dora-core`, `cargo clippy -p dora-core -- -D warnings`, `cargo fmt --check` all pass.

---

⚠️ **This is a machine-generated pull request authored by Claude (Claude Code).** It should be reviewed by a maintainer before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_015PhdgdGBxVnQSk6ekLELQk)_

### DIFF
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -818,18 +818,27 @@ fn rewrite_inputs_map(
 }
 
 fn load_module_file(path: &Path) -> eyre::Result<ModuleFile> {
-    let metadata = std::fs::metadata(path)
-        .with_context(|| format!("failed to stat module file: {}", path.display()))?;
-    if metadata.len() > MAX_MODULE_FILE_SIZE {
+    use std::io::Read as _;
+    // Bound the read itself, not just a `metadata().len()` pre-check: that
+    // check reports 0 for FIFOs/character devices (e.g. `/dev/zero`) and is
+    // racy against a file being appended to, so it would sail past the cap and
+    // then `std::fs::read` unboundedly — exactly the "infinite files" DoS the
+    // limit is meant to prevent. `take(MAX + 1)` caps the read regardless of
+    // what the file claims its size is; if we hit the cap we reject. Mirrors
+    // the manifest loader in `crate::manifest`.
+    let file = std::fs::File::open(path)
+        .with_context(|| format!("failed to read module file: {}", path.display()))?;
+    let mut buf = Vec::new();
+    file.take(MAX_MODULE_FILE_SIZE + 1)
+        .read_to_end(&mut buf)
+        .with_context(|| format!("failed to read module file: {}", path.display()))?;
+    if buf.len() as u64 > MAX_MODULE_FILE_SIZE {
         bail!(
-            "module file too large ({} bytes, limit {}): {}",
-            metadata.len(),
+            "module file too large (limit {} bytes): {}",
             MAX_MODULE_FILE_SIZE,
             path.display()
         );
     }
-    let buf = std::fs::read(path)
-        .with_context(|| format!("failed to read module file: {}", path.display()))?;
     serde_yaml::from_slice(&buf)
         .with_context(|| format!("failed to parse module file: {}", path.display()))
 }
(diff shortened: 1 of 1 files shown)


phil-opp:

This is the summary for a machine-generated review run by Claude (Claude Code). Each PR was verified against a freshly-fetched `origin/main`; all are opened as **drafts** for maintainer review.

### PRs opened, by category

**Correctness**
- #2963 — `fix(core)`: bound the module-file read. `load_module_file` gated on `metadata().len()` then did an *unbounded* `std::fs::read`, so a FIFO/`/dev/zero` (metadata len 0) or a concurrently-growing file bypassed the `MAX_MODULE_FILE_SIZE` DoS guard the comment claims to enforce. Now bounds the read with `take(MAX+1)`, mirroring the already-hardened manifest loader.
- #2964 — `fix(mavlink2-bridge-node)`: stop warning on routine recv read timeouts. Routine `WouldBlock`/`TimedOut`/`Interrupted` (which mavlink-core 0.18 emits every ~100 ms on a quiet link) were classified `Transient` and logged every 5 s forever on a *healthy* link. Adds a `Silent` disposition; genuine anomalies still warn.
- #2966 — `fix(runtime)`: report the correct callback name. A shared-lib operator's `dora_on_event` failure was reported as `on_input failed` — a callback that doesn't exist in the operator FFI.

**Documentation**
- #2965 — `docs(message)`: compile-checked doctest for `format_duration`/`FormattedDuration`, pinning the "coarsest exact unit" round-trip contract (the #2031 fix) that was previously undocumented public API.

**Optimization** — none opened as standalone (see rejected). #2964 incidentally removes per-timeout `Instant::now()`/rate-limit bookkeeping from the reader's hot path.

**Simplification** — none opened as standalone (see rejected).

### Validation

Per branch: `cargo fmt --check` + `cargo clippy -- -D warnings` + full `cargo test` for each affected crate (dora-core, dora-message, dora-runtime, dora-mavlink2-bridge-node) — all green. `/review`-style and `/simplify` gates were run once over the combined diff (clean; the only flag was bounded-read duplication — see rejected #9).

> Note: a workspace-wide `cargo test --all` could not complete in the review sandbox for **environmental** reasons unrelated to these diffs — the `dora-ros2-bridge` link exceeds the disk allowance, and `dora-node-api`'s `zenoh_schema_cache` test fails on `EAFNOSUPPORT` (the sandbox has no IPv6, so `tcp/[::]:0` can't bind). Neither touches any changed crate.

### Findings reviewed but **rejected** (not turned into PRs)

1. **mavlink per-message `Schema` reallocation** (perf) — real, but caching 18+ `schema()` impls is high churn for a 5 Hz default path; poor value-to-risk.
2. **mavlink inbound conversion error tears down the node** — latent only; `to_record_batch` is infallible for these fixed schemas.
3. **runtime `drop_oldest_inputs` per-event `BTreeMap` alloc + O(n) scan** (perf) — a clean fix needs a risky incremental-counter refactor; cost is bounded by cap sizes.
4. **runtime: Python `Event::Reload` dispatched to `on_event`** — intended behavior; the post-reload error-leniency logic is deliberately built around it. Not a bug.
5. **memory-pool: sender-side free-notification drain-stealing across nodes in one process** — plausible, but the fix is a substantial refactor in GPU/CUDA code that can't be verified/tested on a GPU-less host. Deferred to maintainers.
6. **memory-pool: `_unregister_host` passed device VA instead of host base** — currently unreachable given the `effective_as_cuda` predicate; latent, low value.
7. **memory-pool: CUDA current device not restored after GPU-source registration** — low impact (callers re-set device per op); untestable without a GPU.
8. **`validate_data_id` allows `.`/`..` segments** — no sink joins a DataId into a filesystem path today; exploitability unproven.
9. **bounded-read duplication → extract a shared `read_file_capped`** (reuse/altitude) — deliberately skipped: folding the already-tested manifest loader into a focused security fix broadens scope/risk; the code matches the existing per-call-site convention.
10. **operator API `register_operator!`/`DoraStatus` doctest** — `dora-operator-api`'s `lib.rs` already ships a full compile-checked trait doctest; marginal added value.

---

phil-opp:
> [!NOTE]

I looked over this change and did not find blocking issues. Notes for the maintainer:

- The bound is correct at both edges. `file.take(MAX_MODULE_FILE_SIZE + 1)` (`expand.rs:833`) followed by `buf.len() as u64 > MAX_MODULE_FILE_SIZE` (`expand.rs:837`) accepts a file of exactly `MAX_MODULE_FILE_SIZE` bytes and rejects anything larger — no off-by-one, boundary inclusive.
- No regression for legitimate files: the 1 MiB threshold is unchanged from the previous `metadata().len()` check, so files that passed before still pass; the cap is now just enforced against the real read instead of a spoofable/racy stat.
- It faithfully mirrors the already-hardened `NodeManifest::read` in `libraries/core/src/manifest/mod.rs:289`, which is a reasonable convention to follow.
- The two added tests exercise both sides of the limit (over-cap rejected via the "too large" message, exactly-at-cap accepted), and the at-limit test asserts `content.len() == MAX_MODULE_FILE_SIZE`, so the inclusive boundary is pinned.

One non-blocking observation: `load_module_file_rejects_oversized_file` uses a regular finite file, so it confirms rejection but does not actually demonstrate the read is *bounded* — the FIFO/`/dev/zero` case that motivates the fix (a stat-len-0 source that would read forever) isn't covered. That's hard to test portably and the boundary tests are still valuable, so this is just a note, not a request.

Also minor: the error message no longer reports the file's actual size (`expand.rs:840`), since the read is now capped at `MAX + 1`. That's an intentional and acceptable trade-off.

---

phil-opp:

The switch from a `metadata().len()` pre-check to a bounded `take(MAX_MODULE_FILE_SIZE + 1).read_to_end` correctly closes the DoS gap: FIFOs/char devices reporting length 0 and TOCTOU growth can no longer slip past the cap into an unbounded read. The boundary math is right (`u64` constant, inclusive at exactly MAX, rejected at MAX+1, no overflow), and the two regression tests pin both the just-over and exactly-at-cap edges.

---

---

PR #2969, opened 2026-08-03
fix(daemon): drop a closed input's stale deadline so it can't orphan a broken record (#2968)

phil-opp:
Fixes #2968.

## Problem

`close_input` removed an input from `open_inputs` and `broken_inputs` but left its `input_deadlines` entry behind. If the input had an armed deadline (`input_timeout` configured and at least one message received), the stale deadline kept counting after the input closed:

1. `check_input_timeouts` fires on the already-closed input, inserts a `broken_inputs[(node,input)]` record, and calls `break_input`.
2. `break_input` early-returns because the input is no longer in `open_inputs`, leaving the just-inserted `broken_inputs` record **orphaned**.
3. Circuit-breaker recovery only clears `broken_inputs` on message receipt — impossible once the producer has exited.
4. `has_broken_input(node)` is now permanently `true`, so `is_finished` stays false, the node is never sent `AllInputsClosed`, hangs the finish grace period, and is escalated to SIGKILL and reported as a **failure** instead of a clean exit.

This defeats `dora run --exit-when-nodes-finish` (#2957) on exactly the fault-tolerant configs it targets: a drained node with an `input_timeout` input whose producer exits early.

## Fix

- **`close_input`** now drops the input's `input_deadlines` entry when the input closes (mirroring the `forget_node_bookkeeping` cleanup), so a closed input can never time out — the orphan can never form. This is suggestion #1 in the issue and is the root-cause fix.
- **`break_input`**'s early-return path now rolls back the `broken_inputs` record its caller inserted (suggestion #3), so a no-op break can never leave an orphan even if it is ever reached on an already-closed input.

## Tests

- `close_input_drops_armed_deadline` — closing an input with an armed deadline removes the `input_deadlines` entry.
- `drained_node_finishes_despite_stale_deadline_timeout` — a two-input node with one `input_timeout` input: close the timed input, fire the stale-deadline timeout path, then close the other input, and assert `AllInputsClosed` is delivered (not left orphaned).

Both new tests plus the existing `close_input`/`break_input` suite pass; `cargo fmt --all -- --check` and `cargo clippy -p dora-daemon -- -D warnings` are clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NSmdswXU8QEnQjskE8Tf45)_

Issue #2956, opened 2026-08-02
TestingOutput::ToChannel still uses flume after EventStream migration (#1603)


SunSunSun689:
## Summary

`TestingOutput::ToChannel` still uses `flume::Sender`, even though #1603 migrated `EventStream` from flume to `tokio::sync::mpsc`. This is an incomplete migration — the integration testing output path was left behind.

## Current state (main)

```rust
// apis/rust/node/src/integration_testing.rs
pub enum TestingOutput {
    ToFile(std::path::PathBuf),
    ToWriter(Box<dyn std::io::Write + Send>),
    ToChannel(flume::Sender<serde_json::Map<String, serde_json::Value>>), // ← still flume
}
```

## Why it matters

1. **Consistency**: #1603 replaced flume with tokio mpsc across the event stream to fix a spinlock deadlock on low-CPU runners (2-vCPU CI). The same `flume 0.10.14` spinlock lives in `ToChannel`.

2. **Upstream completeness**: Every remaining `flume` dependency is a latent deadlock risk. The output path has lower probability than the input path, but the root cause is identical.

3. **Downstream impact**: Projects like [dora-test-utils](https://github.com/dora-rs/gsoc2026-dora-test-utils) need `ToChannel` for capturing node outputs in unit tests. Keeping flume forces downstream crates to depend on both `flume` and `tokio` simultaneously.

## Proposed fix

```rust
ToChannel(tokio::sync::mpsc::Sender<serde_json::Map<String, serde_json::Value>>),
```

The daemon simulation thread runs outside a tokio runtime, so `send()` must be replaced with `blocking_send()` in `handle_output()` in `node_integration_testing.rs`.

A patch is ready at [SunSunSun689/dora:testing-output-tokio-mpsc](https://github.com/SunSunSun689/dora/tree/testing-output-tokio-mpsc) — happy to open a PR.

Refs: #1603, dora-rs/gsoc2026-dora-test-utils discussions #28

PR #2970, opened 2026-08-03
fix(daemon): start the dataflow when a node death completes the startup barrier (#2967)

### LINKED ISSUE #2967: daemon: startup barrier completed by a node death (not a subscribe) never calls start(), so a non-dynamic node added during startup hangs — #2933 regression
> **Created by an automated, scheduled Claude review** of recent merges (last 2 days). Read from the code, not reproduced end-to-end — the severity is narrow and the reachability window is small, so please confirm before acting.

## Summary

`PendingNodes::handle_node_stop` **discards** the `DataflowStatus` returned by `update_dataflow_status`, whereas the two other callers of the barrier (`Subscribe` and `RemoveNode`) act on it and call `dataflow.start()` on `AllNodesReady`. As a result, when the startup barrier's *final* transition is a cohort member **dying before subscribing** (rather than subscribing), `start()` is never called. Since #2933, non-cohort subscribers are answered `Ok(())`, so a **non-dynamic** node added via `dora node add` during the pre-`dataflow_started` window can end up subscribed-and-alive on a dataflow that never starts and never tears down.

## Where

- `binaries/daemon/src/pending.rs:146-159` — `handle_node_stop` calls `update_dataflow_status(...).await?;` and drops the returned status, returning `Ok(())`.
- Compare the two callers that *do* act on the status:
  - `Subscribe` path: `binaries/daemon/src/lib.rs:3967` — `DataflowStatus::AllNodesReady if !dataflow.dataflow_started => { ... dataflow.start(...) }`.
  - `RemoveNode` path (added in #2933): `binaries/daemon/src/lib.rs:2596` — same handling.
- `answer_subscribe_requests` scoping (from #2933): `binaries/daemon/src/pending.rs:265-275` — non-cohort subscribers get `Ok(())`.
- `should_finish` teardown gate: `binaries/daemon/src/lib.rs:4708-4712` — requires `!local_nodes_pending() && all running_nodes are dynamic`.
- `AddNode` deliberately does **not** enrol the node in the cohort: `binaries/daemon/src/lib.rs:2394-2407`.

## Why it hangs

In the normal cohort-only case the discard is harmless: when a death completes the barrier, `exited_before_subscribe` is non-empty, so `result` is `Err`, and we correctly *don't* want to start — cohort waiters get the `Err` and exit. The discard only bites once there is a **non-cohort survivor that received `Ok(())`** and needs the dataflow to either start or tear down. That survivor path is exactly what #2933 introduced (before #2933 the runtime-added node was enrolled in `local_nodes`/the cohort and would have received the `Err` and died — the #2917 cross-talk this PR fixed).

### Event sequence

1. Cohort = `{A, B}`; neither has subscribed yet; `dataflow_started == false`.
2. `dora node add X` with `X` **non-dynamic** (reachable — `AddNode` only requires the dataflow present in `self.running`, which holds before startup completes; `X` is not enrolled in the cohort, `lib.rs:2394-2407`).
3. `X` subscribes → parked in `waiting_subscribers` as a non-cohort node.
4. `A` dies before subscribing → `handle_node_stop`: `local_nodes = {B}`, status `Pending` (discarded).
5. `B` dies before subscribing → `handle_node_stop`: `local_nodes = {}`, `exited_before_subscribe = [A, B]`, `answer_subscribe_requests` runs. Cohort is dead; `X` (non-cohort) is answered `Ok(())`. The resulting `AllNodesReady` status is **discarded**, so `start()` is never called.
6. `X` is non-dynamic and still in `running_nodes`, so `should_finish` (`lib.rs:4708`, "all remaining nodes dynamic") stays `false` → no teardown. `dataflow_started` stays `false` → timers never start.

**Outcome:** `X` sits subscribed with `Ok(())`, receives no inputs, and the dataflow neither starts nor tears down.

## Scope / severity

Narrow: requires a **non-dynamic** node added via `dora node add` during the pre-`dataflow_started` window **and** the descriptor cohort all crashing-before-subscribe. Dynamic nodes are unaffected (they're covered by `should_finish`'s all-dynamic teardown). This is a *trade* introduced by #2933, not a pure regression — it replaces the #2917 cross-talk (wrong node blamed and killed) with a hang in this corner. Distinct from #2938, which is the multi-daemon/external subscribe-after-startup hang.

## Suggested fix

Make `handle_node_stop` act on the barrier status the way `handle_node_removal` and `Subscribe` do — i.e. when the death completes the cohort and there are non-cohort survivors, drive the same start-or-teardown transition (start if the result is `Ok` and `!dataflow_started`; otherwise ensure the surviving non-cohort node is torn down rather than left parked). Alternatively, decline to answer non-cohort waiters with `Ok(())` while the barrier resolved to a failure, and tear them down as part of the failed-startup path.

## Regression coverage gap

No test drives "barrier completed by a death with a non-cohort survivor parked" — `pending.rs`'s new suite covers removal-completing-the-barrier but not death-completing-it with a survivor.

phil-opp:
Fixes #2967.

## Problem

`PendingNodes::handle_node_stop` discarded the `DataflowStatus` returned by `update_dataflow_status`, whereas the two other callers of the startup barrier act on it and call `dataflow.start()` on `AllNodesReady`:

- `Subscribe` path — `lib.rs` `DataflowStatus::AllNodesReady if !dataflow.dataflow_started => { … dataflow.start(…) }`
- `RemoveNode` path (added in #2933) — same handling.

So when the barrier's **final** transition was a cohort member **dying before subscribing** (rather than subscribing), `start()` was never called.

Since #2933, a non-cohort subscriber is answered `Ok(())` even when the cohort fails (to avoid the #2917 cross-talk). A **non-dynamic** node added via `dora node add` during the pre-`dataflow_started` window is exactly such a survivor. Combined with the discarded status, it could end up subscribed-and-alive on a dataflow that:

- never starts — `dataflow_started` stays `false`, so timers never spawn, and
- never tears down — `should_finish` requires all remaining nodes to be dynamic, and the survivor is non-dynamic.

### Event sequence

1. Cohort `{A, B}`, neither subscribed, `dataflow_started == false`.
2. `dora node add X` (non-dynamic, not enrolled in the cohort); `X` subscribes → parked as a non-cohort waiter.
3. `A` dies before subscribing → barrier still `Pending` (status discarded).
4. `B` dies before subscribing → barrier completes; `X` is answered `Ok(())`; the resulting `AllNodesReady` is **discarded**, so `start()` is never called.
5. `X` sits subscribed, receives nothing, and the dataflow neither starts nor tears down.

## Fix

`handle_node_stop` now **returns** the barrier `DataflowStatus`, and `handle_node_stop_inner` drives the same start transition as the other two callers — starting the dataflow on `AllNodesReady` while `!dataflow_started`. `start()` only spawns timer tasks and sets the flag; it is idempotent and guarded by `dataflow_started`, so this is a no-op on the common path.

This is suggestion #1 in the issue ("make `handle_node_stop` act on the barrier status the way `handle_node_removal` and `Subscribe` do"), and makes all three barrier callers consistent.

## Tests

Fills the coverage gap the issue names ("no test drives barrier-completed-by-a-death with a non-cohort survivor parked"):

- `death_completing_the_barrier_reports_all_nodes_ready` — a death that empties the cohort reports `AllNodesReady` and releases a parked non-cohort survivor.
- `death_leaving_the_barrier_incomplete_stays_pending` — a death with another member still pending stays `Pending` (no early start).

Existing `pending` suite (13 tests) passes; `cargo fmt --all -- --check` and `cargo clippy -p dora-daemon -- -D warnings` are clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NSmdswXU8QEnQjskE8Tf45)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -4655,7 +4655,7 @@ impl Daemon {
             ),
         };
 
-        dataflow
+        let status = dataflow
             .pending_nodes
             .handle_node_stop(
                 node_id,
@@ -4665,6 +4665,25 @@ impl Daemon {
                 &mut logger,
             )
             .await?;
+
+        // If this death completed the startup barrier (the dying node was the
+        // last cohort member yet to subscribe), the barrier resolves to
+        // `AllNodesReady` just as a final subscribe or a `RemoveNode` would —
+        // and both of those callers start the dataflow here. Do the same, so a
+        // non-cohort survivor that was answered `Ok(())` (since #2933) isn't
+        // left parked on a dataflow that never starts and never tears down
+        // (#2967). `start()` is idempotent and guarded by `dataflow_started`.
+        if matches!(status, DataflowStatus::AllNodesReady) && !dataflow.dataflow_started {
+            logger
+                .log(
+                    LogLevel::Info,
+                    None,
+                    Some("daemon".into()),
+                    "startup barrier completed by a node exit, starting dataflow",
+                )
+                .await;
+            dataflow.start(&self.events_tx, &self.clock).await?;
+        }
 
         // node only reaches here if it will not be restarted
         let might_restart = false;
--- a/binaries/daemon/src/pending.rs
+++ b/binaries/daemon/src/pending.rs
@@ -142,7 +142,7 @@ impl PendingNodes {
         clock: &HLC,
         cascading_errors: &mut CascadingErrorCauses,
         logger: &mut DataflowLogger<'_>,
-    ) -> eyre::Result<()> {
+    ) -> eyre::Result<DataflowStatus> {
         if self.local_nodes.remove(node_id) {
             logger
                 .log(
@@ -153,10 +153,19 @@ impl PendingNodes {
                 )
                 .await;
             self.exited_before_subscribe.push(node_id.clone());
-            self.update_dataflow_status(coordinator_sender, clock, cascading_errors, logger)
-                .await?;
+            // Return the barrier status so the caller can drive the same
+            // start-or-teardown transition that `Subscribe` and `RemoveNode`
+            // do. A death can complete the barrier just as a subscribe can;
+            // dropping the status here left a non-cohort survivor (answered
+            // `Ok(())` since #2933) parked on a dataflow that never starts
+            // and never tears down (#2967).
+            return self
+                .update_dataflow_status(coordinator_sender, clock, cascading_errors, logger)
+                .await;
         }
-        Ok(())
+        // The dying node was not a pending cohort member (it had already
+        // subscribed), so the barrier state is unchanged.
+        Ok(DataflowStatus::Pending)
     }
 
     pub async fn handle_dataflow_stop(
@@ -603,6 +612,70 @@ mod tests {
             reply_of(&mut rx),
             Ok(()),
             "the parked non-cohort node inherited a cohort member's death"
+        );
+    }
+
+    /// #2967: when a death is the *final* transition that completes the
+    /// startup barrier, `handle_node_stop` must report `AllNodesReady` — the
+    /// signal the daemon acts on to call `dataflow.start()`. Dropping it left
+    /// a non-cohort survivor parked on a dataflow that never started.
+    #[tokio::test]
+    async fn death_completing_the_barrier_reports_all_nodes_ready() {
+        let (last, survivor) = (node("last"), node("survivor"));
+        let mut p = pending();
+        // `last` is the only cohort member still pending; `survivor` joined
+        // later (not enrolled) and is parked waiting to run.
+        p.insert(last.clone());
+        let mut rx = park(&mut p, &survivor);
+
+        let mut daemon_logger = test_logger();
+        let status = p
+            .handle_node_stop(
+                &last,
+                &mut None,
+                &HLC::default(),
+                &mut CascadingErrorCauses::default(),
+                &mut daemon_logger.for_dataflow(uuid::Uuid::nil()),
+            )
+            .await
+            .expect("stop should succeed");
+
+        assert!(
+            matches!(status, DataflowStatus::AllNodesReady),
+            "a death that empties the cohort must report the barrier ready so \
+             the dataflow can start"
+        );
+        assert_eq!(
+            reply_of(&mut rx),
+            Ok(()),
+            "the non-cohort survivor should be released, not stranded"
+        );
+    }
+
+    /// A death that does *not* complete the barrier (other members still
+    /// pending) leaves it `Pending`, so the daemon does not start early.
+    #[tokio::test]
+    async fn death_leaving_the_barrier_incomplete_stays_pending() {
+        let (dead, still_pending) = (node("dead"), node("still-pending"));
+        let mut p = pending();
+        p.insert(dead.clone());
+        p.insert(still_pending.clone());
+
+        let mut daemon_logger = test_logger();
+        let status = p
+            .handle_node_stop(
+                &dead,
+                &mut None,
+                &HLC::default(),
+                &mut CascadingErrorCauses::default(),
+                &mut daemon_logger.for_dataflow(uuid::Uuid::nil()),
+            )
+            .await
+            .expect("stop should succeed");
+
+        assert!(
+            matches!(status, DataflowStatus::Pending),
+            "the barrier must stay pending while a cohort member has not resolved"
         );
     }
 


phil-opp:

I reviewed the diff and found no issues.

- Barrier accounting is correct: `handle_node_stop` (`pending.rs:146`) removes the dying node exactly once via the `local_nodes.remove(node_id)` HashSet guard before re-evaluating, so no double- or under-count; a node that had already subscribed (not in `local_nodes`) returns `Pending` unchanged.
- No premature start: `update_dataflow_status` returns `AllNodesReady` only when `local_nodes` is empty *and* there are no external nodes. The multi-daemon (`external_nodes`) branch returns `Pending` and reports upward, so a local death cannot start a distributed dataflow early.
- The new start transition in `handle_node_stop_inner` (`lib.rs`) is identical to the two other barrier callers — `handle_node_subscription` (`lib.rs:3967`) and `handle_node_removal` (`lib.rs:2595`) — both of which guard on `AllNodesReady && !dataflow_started` and call `dataflow.start()`. `start()` is idempotent, so this is consistent and cannot double-start or hang.
- The regression tests assert meaningfully (status value plus the survivor's released reply), and `DataflowStatus` is already imported in `lib.rs`.

---

phil-opp:
The `Benchmark regression check` failure on the merge-queue batch ([SHA 4fbf433](https://github.com/dora-rs/dora/commit/4fbf4331e22972a315b3d886f4f0fe8d9b5e0a8b)) is **not caused by this PR**. It's a compile error (`E0599`, a uuid `now_v7`/`new_v7` method-not-found) in `dora-message` — a crate this PR doesn't touch.

Verified:
- `cargo bench -p dora-message --no-run` (the exact command the job runs) compiles cleanly on the base branch tip (313ca4c) and on this PR's branch.
- This PR only changes `binaries/daemon/src/{lib.rs,pending.rs}`; it touches neither `dora-message` nor any `uuid` usage or `Cargo.toml`.

The error came from another PR batched into the same optimistic merge-queue test group (this PR and #2969 were tested together on that SHA and failed identically). No change is needed here — it should go green once the queue bisects the offending sibling PR out of the batch.

---

phil-opp:
Same root cause as the failure on #2969 (correcting my earlier note — full [job log](https://github.com/dora-rs/dora/actions/runs/30939301951/job/92093301340) read):

The merge-queue `Benchmark regression check` runs `cargo bench -p dora-message --no-run`, which fails with 7× `E0599: no associated function named 'new_v4' found for uuid::Uuid` in `libraries/message/src/ws_protocol.rs`. The workspace enables only uuid `v7` (not `v4`); `dora-message` uses `uuid = { workspace = true }` but calls `Uuid::new_v4()`, which normally compiles only via workspace feature-unification (uuid `v4` comes from `dora-coordinator`/`dora-record-node`/`dora-tracing`). The isolated benchmark build doesn't get that unification.

This is unrelated to this PR (only `binaries/daemon/src/{lib.rs,pending.rs}`; no `dora-message`/uuid/manifest changes), and this check is `skipped` on the PR's own path-filtered CI. Suggested fix lives in `libraries/message/Cargo.toml` (`uuid = { workspace = true, features = ["v4"] }`), a separate shared-manifest change I've left out of this PR. Details in [my comment on #2969](https://github.com/dora-rs/dora/pull/2969#issuecomment-5182546647).

---

phil-opp:
The `Test (ubuntu-latest)` failure on [SHA 080a0f7](https://github.com/dora-rs/dora/commit/080a0f7bc032b14e95c2c692d11660f0124c5acb) was a **transient merge-queue stacking artifact**, not this PR (which is daemon-only):

`E0308` in `apis/rust/node/src/node/mod.rs` — `TestingOutput::ToChannel(tx)` expected `UnboundedSender<OutputJson>` but got a `flume::Sender`. This came from two fixes for #2956 landing close together: #2972 (bounded `tokio::sync::mpsc::Sender`, merged first) and a follow-up that switched `ToChannel` to `UnboundedSender`. At that intermediate stacked SHA the enum was already `UnboundedSender` while a `flume::unbounded()` call site wasn't yet updated.

Current `main` (91c2184) is internally consistent (enum + all call sites use `unbounded_channel()`), and I verified `cargo test -p dora-node-api --no-run` compiles cleanly on it. This PR touches only `binaries/daemon/src/{lib.rs,pending.rs}` and doesn't go near that code, so the bisection re-test against current `main` should be green.

---

phil-opp:
The `Test (ubuntu-latest)` failure on [SHA 391f38f](https://github.com/dora-rs/dora/commit/391f38ff5504f1e04972563fa9b5a6bafc6e0d9b) is an **unrelated flaky test**, not this PR:

```
shutdown::tests::a_wrapper_that_exits_does_not_hide_its_surviving_child
  panicked at binaries/daemon/src/shutdown.rs:289
  assertion `left == right` failed: the wrapper is alive, so this is an ordinary wait
    left: Done   right: Waiting
(199 passed; 1 failed)
```

That test (from #3004) spawns `sh -c "sleep 300 & exit 0"` and then immediately asserts the wrapper is still alive (`Waiting`). It's racy: `exit 0` can be reaped before the first `poll`, so the group reads `Done`. It's `#[tokio::test(start_paused = true)]`, but the real `sh` subprocess exit isn't governed by paused tokio time.

It's unrelated to this PR:
- This PR changes only `binaries/daemon/src/{lib.rs,pending.rs}` — not `shutdown.rs`, and doesn't touch `DestroyWait`/process-group logic.
- The same test **passed in this PR's isolated bisection** ([SHA d0417dc](https://github.com/dora-rs/dora/commit/d0417dc8a8ba4a6cd4f5a03cb1dfdc43cacd9af3)) minutes earlier.

So a re-run should clear it. Flagging the flake separately in case it's worth hardening (e.g. spawn a wrapper that stays alive until the first assert, rather than `exit 0`).

---

PR #2971, opened 2026-08-03
fix(daemon): settle the restart announcement when a respawn returns dynamic (#2936)

### LINKED ISSUE #2936: daemon: restart_loop that breaks after announcing `SpawnedNodeResult{restart:true}` leaks the node in `running_nodes` → dataflow never finishes (stop-vs-restart race and respawn-failure both hit it)
> **This issue was created by a scheduled automated Claude check** (an unattended code-review routine). The finding was verified by reading the full event chain across `prepared.rs`, `lib.rs`, and `running_dataflow.rs`, but has **not** been reproduced with a running dataflow — treat the trigger scenarios as read-from-code, and please sanity-check before acting.

## Summary

`RawSpawner::restart_loop` emits `DoraEvent::SpawnedNodeResult { restart, .. }` to the daemon **before** the backoff sleep and the final `disable_restart` re-check. On the `restart == true` branch the daemon treats that event as "a respawn is coming" — it only notifies downstream (`NodeRestarted`) and deliberately does **not** remove the node from `running_nodes` or record a terminal result. It then relies on the `restart_loop` to follow up with either a `ProcessHandleReplaced` (successful respawn) or a later `SpawnedNodeResult { restart:false }` (gave up).

But there are `break` paths that exit `restart_loop` **after** `SpawnedNodeResult{restart:true}` has already been sent, and none of them send a corrective terminal event. The node is then stuck in `running_nodes` forever, and because it is non-dynamic, the dataflow can never satisfy the finish condition — `dora stop` hangs until an external timeout.

## The event `restart:true` is sent, then the loop can break

`binaries/daemon/src/spawn/prepared.rs`, `restart_loop`:

```rust
// line 351-365: emitted unconditionally, restart may be `true`
let event = DoraEvent::SpawnedNodeResult {
    dataflow_id: self.dataflow_id,
    node_id: self.node.id.clone(),
    exit_status,
    dynamic_node: self.node.kind.dynamic(),
    restart,          // <-- true
    restart_count,
    pid: exited_pid,
}.into();
// ...
let _ = self.daemon_tx.clone().send(event).await;   // daemon now expects a follow-up

if restart {
    if let Some(base_delay) = config.restart_delay {
        // ... line 385:
        tokio::time::sleep(backoff).await;           // .await point — stop can race in here
    }

    // line 399-408: final re-check
    if disable_restart.load(atomic::Ordering::Acquire) {
        logger.log(/* "restart cancelled: inputs closed before respawn" */).await;
        break;                                       // (A) leaves WITHOUT any corrective event
    }

    restart_count += 1;
    // ...
    let result = self.clone().spawn_inner(&mut logger, op_rx_new, finished_tx).await;
    match result {
        Ok(NodeKind::Spawned { pid }) => { /* sends ProcessHandleReplaced */ }
        Ok(NodeKind::Dynamic) => { /* log */ break; }
        Err(err) => {
            logger.log(/* "failed to restart node: {err}" */).await;
            break;                                    // (B) also leaves WITHOUT any corrective event
        }
    }
}
```

The task is `tokio::spawn`'d and its `JoinHandle` is dropped (`prepared.rs:206`), so when the task ends via `break` nothing observes it. There is **no** post-loop code that emits a terminal `SpawnedNodeResult`.

## Why the leak is permanent

Daemon handling in `binaries/daemon/src/lib.rs` (`DoraEvent::SpawnedNodeResult`, from `:4839`):

- On `restart == true` (`:5057`) it only fans out `NodeEvent::NodeRestarted` to downstream subscribers. It does **not** call `handle_node_stop`, so the node stays in `running_nodes`.
- Only the `restart == false` branch (`:5146`, ending at `:5162`) calls `handle_node_stop(...)`, which is what removes the node and can trigger `finish_dataflow`.

`handle_node_stop_inner` (`:4535`) gates finishing on every remaining running node being dynamic:

```rust
let should_finish = !dataflow.pending_nodes.local_nodes_pending()
    && dataflow.running_nodes.iter().all(|(_id, n)| n.node_config.dynamic);
```

A leaked **non-dynamic** node keeps this `false` for every *other* node's stop, so `finish_dataflow` is never called.

`stop_all` (`binaries/daemon/src/running_dataflow.rs:430`) sets `disable_restart` on all nodes (`:448-450`) and `stop_sent = true` (`:509`), then `should_finish_immediately` (`:514`) returns `WaitForNodes` because the leaked node is non-dynamic. And the finish-straggler watchdog cannot rescue it — `finish_stragglers` (`:745`) early-returns an empty list whenever `stop_sent` is true:

```rust
pub(crate) fn finish_stragglers(&self, grace: Duration, now_millis: u64) -> Vec<NodeId> {
    if self.stop_sent || !self.open_external_mappings.is_empty() {
        return Vec::new();
    }
    // ...
}
```

So after a `dora stop`, nothing removes the leaked node and the dataflow hangs. The only terminal signal to the CLI (`DataflowDaemonResult`) is sent from `finish_dataflow`, which never runs.

## Concrete triggers

**(A) Stop-vs-restart race** — needs `restart_delay` to widen the window:
1. A node with `restart_policy: Always` (or `OnFailure`) **and** `restart_delay` set crashes.
2. `restart_loop` sends `SpawnedNodeResult{restart:true}`, downstream gets `NodeRestarted`, and it enters the backoff `sleep` (`prepared.rs:385`).
3. During that sleep the operator runs `dora stop` → `stop_all` sets `disable_restart` on every node.
4. The loop wakes, the re-check at `:399` sees `disable_restart == true`, logs `"restart cancelled: inputs closed before respawn"`, and `break`s at `:407`.
5. The node is never removed; `dora stop` blocks on `DataflowDaemonResult` until an external timeout.

**(B) Respawn failure** — does not even need the stop race:
1. A node with `restart_policy: Always`/`OnFailure` crashes.
2. `restart_loop` sends `SpawnedNodeResult{restart:true}`, passes the `disable_restart` re-check, and calls `spawn_inner`.
3. `spawn_inner` returns `Err(..)` (e.g. the executable was removed/replaced, `fork`/exec fails under resource pressure, working dir gone), so the loop logs `"failed to restart node"` and `break`s at `:498`.
4. The daemon was told `restart:true` and never hears back; the node lingers in `running_nodes` and the dataflow can't finish.

## Suggested direction

When `restart_loop` abandons a restart it has **already announced** as `restart:true`, it must deliver a corrective terminal signal so the daemon removes the node and can finish the dataflow — e.g. re-send `SpawnedNodeResult { restart:false, .. }` (with the same `exit_status`/`pid`) on the `break` paths at `prepared.rs:407` and `:498`, or route those cases through the normal stop path. Note the interaction with the pid-staleness guard added in #2893 (`lib.rs:4875`): the corrective event should carry the original `exited_pid` so it isn't discarded as a stale exit.

## Notes

- Distinct from #2926 (incarnation identity on `ProcessHandleReplaced` / `dataflow_node_results` during `remove`→`add` races) — that is about a dead incarnation acting on a live one; this is about a `restart:true` announcement with no terminal follow-up.
- Not covered by existing tests: `tests/node-lifecycle-e2e.rs` pins the remove→re-add orderings, not a stop (or respawn failure) racing an in-progress restart-with-backoff.

phil-opp:
Fixes the one abort path #2978 left open on #2936. Rebased onto `main` and reduced to that branch — the rest of the original diff landed with #2978.

`restart_loop` announces `SpawnedNodeResult { restart: true, .. }` *before* it attempts the respawn. The daemon's restart arm keeps the node's `running_nodes` entry and waits for a follow-up, so every path that leaves the loop after that announcement owes it a terminal event — that is the contract `send_terminal_exit`'s doc comment states. #2978 added the helper and wired it into the cancel and spawn-failure branches, but `Ok(NodeKind::Dynamic)` still `break`s silently, leaking the entry (a non-dynamic leak keeps `should_finish` false, so the dataflow never finishes and `dora stop` hangs).

This calls `send_terminal_exit` there too, with the announcing generation and the exited pid: no `ProcessHandleReplaced` was sent on this path, so that is what the daemon's entry still holds and what `matches_generation` accepts.

To be clear about the severity: that arm is not reachable in production today — `command: None` implies a dynamic node, and a dynamic node's `finished_tx` is dropped before any announcement, so its loop exits at the "failed to receive finished signal" break. This closes the invariant rather than a live hang.

`dynamic_respawn_settles_terminal_exit` fails without the fix (no terminal event arrives) and passes with it; the two abort-path tests now share `announce_restart` / `expect_terminal_settlement` scaffolding.

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -560,6 +560,11 @@ impl PreparedNode {
                                 "cannot restart dynamic node".to_string(),
                             )
                             .await;
+                        // Same restart debt as the branches below and above:
+                        // the respawn produced no process to hand back, so
+                        // nothing else will ever settle the announcement.
+                        self.send_terminal_exit(generation, exit_status, restart_count, exited_pid)
+                            .await;
                         break;
                     }
                     Err(err) => {
@@ -1249,15 +1254,24 @@ mod tests {
         );
     }
 
-    /// A stop that lands during restart backoff (after `restart: true` was
-    /// announced) must settle the daemon's restart debt with a terminal
-    /// `SpawnedNodeResult` — otherwise the node entry is never removed and
-    /// the dataflow can never finish (dora-rs/dora#2926 review).
-    #[tokio::test]
-    async fn cancelled_restart_settles_terminal_exit() {
+    /// The incarnation that crashes in the abort-path tests below.
+    const ANNOUNCING_GENERATION: u64 = 7;
+    const EXITED_PID: u32 = 42;
+
+    /// Drive a restart loop up to and including its `restart: true`
+    /// announcement: start it, open the registration gate, report a failed
+    /// exit, then consume and check the announcement. Returns the daemon
+    /// receiver, the loop's handle, and the shared `disable_restart` flag so
+    /// the caller can steer the abort path it exercises.
+    async fn announce_restart(
+        restart_delay_secs: f64,
+    ) -> (
+        tokio::sync::mpsc::Receiver<Timestamped<Event>>,
+        tokio::task::JoinHandle<()>,
+        Arc<AtomicBool>,
+    ) {
         let (daemon_tx, mut daemon_rx) = tokio::sync::mpsc::channel(8);
-        // Long backoff so the disable flag flips deterministically mid-sleep.
-        let node = test_prepared_node(daemon_tx, 7, 0.5);
+        let node = test_prepared_node(daemon_tx, ANNOUNCING_GENERATION, restart_delay_secs);
         let (finished_tx, finished_rx) = oneshot::channel();
         let (registered_tx, registered_rx) = oneshot::channel();
         let disable_restart = Arc::new(AtomicBool::new(false));
@@ -1279,7 +1293,7 @@ mod tests {
             finished_tx
                 .send(NodeProcessFinished {
                     exit_status: NodeExitStatus::ExitCode(1),
-                    pid: 42,
+                    pid: EXITED_PID,
                 })
                 .is_ok(),
             "report node exit"
@@ -1292,24 +1306,40 @@ mod tests {
                 restart,
                 ..
             }) => {
-                assert_eq!(generation, 7);
+                assert_eq!(generation, ANNOUNCING_GENERATION);
                 assert!(restart, "first event must announce the restart");
             }
             other => panic!("unexpected event: {other:?}"),
         }
 
-        // Stop lands while the loop sleeps out its restart backoff.
-        disable_restart.store(true, atomic::Ordering::Release);
+        (daemon_rx, loop_task, disable_restart)
+    }
 
+    /// Assert that the announced restart was settled by exactly one terminal
+    /// `SpawnedNodeResult`, and that the loop then exited. The terminal event
+    /// must carry the announcing generation and the exited pid: no
+    /// `ProcessHandleReplaced` was sent on these paths, so that is what the
+    /// daemon's `running_nodes` entry still holds — anything else is dropped
+    /// as a stale exit and the entry leaks.
+    async fn expect_terminal_settlement(
+        daemon_rx: &mut tokio::sync::mpsc::Receiver<Timestamped<Event>>,
+        loop_task: tokio::task::JoinHandle<()>,
+        what: &str,
+    ) {
         let second = daemon_rx.recv().await.expect("terminal settlement");
         match second.inner {
             Event::Dora(DoraEvent::SpawnedNodeResult {
                 generation,
                 restart,
+                pid,
                 ..
             }) => {
-                assert_eq!(generation, 7, "terminal event must match the entry");
-                assert!(!restart, "cancelled restart must settle terminally");
+                assert_eq!(
+                    generation, ANNOUNCING_GENERATION,
+                    "terminal event must match the entry"
+                );
+                assert!(!restart, "{what} must settle terminally");
+                assert_eq!(pid, EXITED_PID, "terminal event must carry the exited pid");
             }
             other => panic!("unexpected event: {other:?}"),
         }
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I traced the full event chain and the fix looks correct:

- `send_abandoned_restart_result` (prepared.rs) sets all required `SpawnedNodeResult` fields, including `pid`, and emits exactly one terminal `restart: false` per abandoned path — each call is immediately followed by `break`, and the three `break` sites (disable_restart re-check, `Ok(Dynamic)`, `Err`) are mutually exclusive within an iteration.
- The `exit_status.clone()` at the announcement keeps the owned `exit_status` available for the corrective event; `NodeExitStatus` derives `Clone` and there's no use-after-move, so it compiles.
- The reused `exited_pid` passes the daemon's pid-staleness guard (lib.rs:5033) as claimed: none of these paths sent a `ProcessHandleReplaced`, so the tracked pid still equals `exited_pid`, and the guard only fires on `!restart && current_pid != pid`. The daemon keeps the node on `restart: true` and removes it on `restart: false`, so there's no double-removal.

Minor (non-blocking, cosmetic only): the `Ok(Dynamic)`/`Err` paths run after `restart_count += 1`, so the corrective result carries a count one higher than the announcement, while the `disable_restart` path uses the pre-increment value. `restart_count` in the terminal result is log-only, so no functional impact.

No automated test accompanies the fix, but the commit message documents why (`restart_loop` drives real child processes; e2e coverage is the appropriate tier per #2936).

---

phil-opp:

I don't think this can land as-is — it looks superseded by #2978 (`fix(daemon): isolate node lifecycle generations`), which merged to `main` after this PR was opened and fixes the same #2936 leak. Two concrete problems against current `main`:

1. **Won't compile.** #2978 added a mandatory `generation: u64` field to `DoraEvent::SpawnedNodeResult` (`binaries/daemon/src/event_types.rs:178`). The new `send_abandoned_restart_result` helper constructs `SpawnedNodeResult { … }` without it → `error[E0063]: missing field generation`.

2. **Two of the three abandoned-restart paths are already fixed on `main`.** #2978 added `send_terminal_exit(generation, exit_status, restart_count, exited_pid)` and already calls it from the `disable_restart` re-check branch (`prepared.rs:454`) and the `Err(err)` respawn-failure branch (`prepared.rs:554`) — the same two paths this PR patches. The doc comment's pid-staleness rationale is also obsolete: main's stale-exit guard is now generation-based (`matches_generation`, `lib.rs:5077`), not a pid comparison, so reusing `exited_pid` no longer accomplishes anything.

The one path this PR covers that #2978 did **not**: the `Ok(NodeKind::Dynamic)` respawn branch (`prepared.rs:542` on main) still `break`s without a terminal event, so a node that announced `restart: true` and then respawns as dynamic can still leak in `running_nodes`. If anything here is worth keeping, that's it — as a rebase that calls the existing `send_terminal_exit(generation, …)` rather than adding a new helper.

Recommend rebasing onto current `main` (or closing in favor of #2978 plus a small follow-up for the dynamic-node branch).

---

phil-opp:

Re-review after the 08-04 rebase — no new issues, and the earlier compile concern is resolved.

`send_terminal_exit(generation, exit_status, restart_count, exited_pid)` now exists on `main` (added by #2978) and the new call matches its signature, so the rebased diff compiles. `exit_status.clone()` at the announcement keeps the owned value movable into `send_terminal_exit`, and `generation` is still the announcing generation in the `Dynamic` branch (it's only reassigned in the `Ok(Spawned)` arm), so `matches_generation` stays correct. The new `dynamic_respawn_settles_terminal_exit` test is non-vacuous — a `command: None` prepared node returns `Ok(Dynamic)`, and without the fix the terminal settlement event never arrives.

---

PR #2972, opened 2026-08-03
fix(node): migrate TestingOutput::ToChannel from flume to tokio mpsc (#2956)

### LINKED ISSUE #2956: TestingOutput::ToChannel still uses flume after EventStream migration (#1603)

phil-opp:
Fixes #2956.

## Problem

#1603 migrated `EventStream` from `flume` to `tokio::sync::mpsc`, but the
integration-testing **output** path was left behind: `TestingOutput::ToChannel`
still carried a `flume::Sender`, so the same spinlock lived on in the output
path and downstream test-util crates had to depend on both `flume` and `tokio`
just to capture node outputs.

## Fix

`ToChannel` now takes a `tokio::sync::mpsc::UnboundedSender`.

**Unbounded, not bounded.** The issue proposed the bounded `Sender`, but that
deadlocks. The daemon simulation thread parks in `blocking_send` on a full
output channel while the node thread is parked in `blocking_recv` awaiting that
same output's reply (`daemon_connection/mod.rs`) — the test hangs forever, with
no timeout and no diagnostic. Reproduced directly: a capacity-2 channel with
five outputs wedges at 0% CPU indefinitely.

Nothing consumes the channel while the node runs — every call site in the repo
drains only *after* the node finishes, and five of them keep the receiver alive
but never read it — so no caller can supply the backpressure a bounded channel
needs. `UnboundedSender` is an exact match for the `flume::unbounded()`
semantics all of these sites already had, and `send` needs no `blocking_send`,
so the output path no longer cares what thread or runtime it runs on.

## Changes

- `integration_testing.rs`: `ToChannel(UnboundedSender<OutputJson>)`, plus
  `unbounded_channel`, `drain_outputs` and the `OutputJson` alias so tests need
  no `tokio` dependency of their own. Only the unbounded API is re-exported, so
  the bounded constructor is not reachable through `dora-node-api`.
- `daemon_connection/node_integration_testing.rs`: `OutputWriter::Channel` type
  and a plain non-blocking `send`.
- All ten in-workspace example test crates migrated — these were the build break
  flagged in review; they are workspace members, so `cargo test --all` covered
  them and would have failed.
- The four guide/docs snippets that still showed the flume pattern.

## Tests

`to_channel_delivers_outputs_in_order` covers end-to-end delivery and ordering
through `ToChannel`. This path was previously untested — every existing
`ToChannel` test dropped the receiver (`let (tx, _rx) = …`).

`dora-node-api` 159/159, all ten example crates green, `cargo check --all
--tests`, CI-equivalent `cargo clippy --all -- -D warnings`, and
`cargo fmt --all -- --check` all clean.

### DIFF
--- a/apis/rust/node/src/daemon_connection/node_integration_testing.rs
+++ b/apis/rust/node/src/daemon_connection/node_integration_testing.rs
@@ -166,8 +166,12 @@ impl IntegrationTestingEvents {
                 writeln!(writer.as_mut()).context("failed to write newline to output file")?;
             }
             OutputWriter::Channel(sender) => {
+                // The testing daemon simulation runs on a dedicated
+                // `std::thread` outside any tokio runtime (see
+                // `DoraNode::init_testing`), so `blocking_send` is safe here
+                // and gives flume's original blocking-on-full semantics.
                 sender
-                    .send(output)
+                    .blocking_send(output)
                     .context("failed to send output to channel")?;
             }
         }
@@ -234,7 +238,7 @@ impl IntegrationTestingEvents {
 
 enum OutputWriter {
     Writer(Box<dyn Write + Send>),
-    Channel(flume::Sender<serde_json::Map<String, serde_json::Value>>),
+    Channel(tokio::sync::mpsc::Sender<serde_json::Map<String, serde_json::Value>>),
 }
 
 pub fn convert_output_to_json(
--- a/apis/rust/node/src/daemon_connection/node_integration_testing.rs
+++ b/apis/rust/node/src/daemon_connection/node_integration_testing.rs
@@ -166,12 +166,9 @@ impl IntegrationTestingEvents {
                 writeln!(writer.as_mut()).context("failed to write newline to output file")?;
             }
             OutputWriter::Channel(sender) => {
-                // The testing daemon simulation runs on a dedicated
-                // `std::thread` outside any tokio runtime (see
-                // `DoraNode::init_testing`), so `blocking_send` is safe here
-                // and gives flume's original blocking-on-full semantics.
+                // Must not block: the node is waiting for this request's reply.
                 sender
-                    .blocking_send(output)
+                    .send(output)
                     .context("failed to send output to channel")?;
             }
         }
@@ -238,7 +235,7 @@ impl IntegrationTestingEvents {
 
 enum OutputWriter {
     Writer(Box<dyn Write + Send>),
-    Channel(tokio::sync::mpsc::Sender<serde_json::Map<String, serde_json::Value>>),
+    Channel(tokio::sync::mpsc::UnboundedSender<serde_json::Map<String, serde_json::Value>>),
 }
 
 pub fn convert_output_to_json(
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -2314,12 +2314,54 @@ mod tests {
             "test-node".parse().unwrap(),
             events,
         ));
-        let (tx, _rx) = flume::unbounded();
+        let (tx, _rx) = tokio::sync::mpsc::channel(16);
         let outputs = TestingOutput::ToChannel(tx);
         let options = TestingOptions {
             skip_output_time_offsets: true,
         };
         crate::DoraNode::init_testing(inputs, outputs, options).unwrap()
+    }
+
+    /// #2956: an output sent through `TestingOutput::ToChannel` must actually
+    /// arrive on the receiver. This is the end-to-end path that the flume →
+    /// `tokio::sync::mpsc` migration touched (`handle_output`'s
+    /// `blocking_send`). Plain `#[test]` on purpose: the testing daemon
+    /// bridge uses `blocking_send`/`blocking_recv`, which panic inside a tokio
+    /// runtime.
+    #[test]
+    fn to_channel_delivers_outputs() {
+        use arrow::array::Int32Array;
+
+        let events = vec![TimedIncomingEvent {
+            time_offset_secs: 0.0,
+            event: IncomingEvent::Stop,
+        }];
+        let inputs = TestingInput::Input(IntegrationTestInput::new(
+            "test-node".parse().unwrap(),
+            events,
+        ));
+        let (tx, mut rx) = tokio::sync::mpsc::channel(16);
+        let outputs = TestingOutput::ToChannel(tx);
+        let options = TestingOptions {
+            skip_output_time_offsets: true,
+        };
+        let (mut node, _events) = crate::DoraNode::init_testing(inputs, outputs, options).unwrap();
+
+        node.send_output(
+            "out".parse().unwrap(),
+            Default::default(),
+            Int32Array::from(vec![1, 2, 3]),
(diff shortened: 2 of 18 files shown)


phil-opp:

No blocking issues found. I traced the migration and the semantics hold up:

- `handle_output`'s new `blocking_send` (`daemon_connection/node_integration_testing.rs`) runs on the dedicated `std::thread` daemon-sim loop (`node/mod.rs:1159-1171`, driven by `blocking_recv`), outside any tokio runtime — so `blocking_send` won't panic, and it stays on the sim thread regardless of the caller's context. The inline comment is accurate.
- Behavior is preserved vs. flume: `blocking_send` blocks when the bounded channel is full and returns `Err` on a closed (receiver-dropped) channel; the error is still propagated via `.context(...)?`.
- `to_channel_delivers_outputs` is a genuine end-to-end assertion — `send_output` blocks on the reply, so the sim thread's `blocking_send` has completed before `try_recv` runs, and it checks the delivered `id`. Correctly a plain `#[test]`.

One minor observation (non-blocking): the enum now only accepts a *bounded* `Sender` (tokio has no unbounded variant behind this type), whereas the old `flume::Sender` could be unbounded. The documented "collect outputs after `crate::main()` returns" pattern (`integration_testing.rs` module docs, capacity `16`) could therefore deadlock if a node emits more outputs than the channel capacity before anything drains — previously impossible with `flume::unbounded()`. It's an `ignore`d illustrative example and the doc note flags the block-when-full behavior, so this is just a heads-up on the capacity chosen in the example.

---

phil-opp:

I found the following issue:

**Build break: the `ToChannel` retype isn't applied to the in-workspace example crates.** This PR changes the public `TestingOutput::ToChannel` variant to take a `tokio::sync::mpsc::Sender`, but only migrates the call sites inside `apis/rust/node`. Ten example test crates still construct it from `flume::unbounded()`, so they fail to compile under `cargo test --all` / `cargo build --all` (the mandatory QA gate — PR CI's `cargo test --all` only excludes the three Python packages, not these). Each is a hard `flume::Sender` vs `tokio::sync::mpsc::Sender` type mismatch at the `ToChannel(tx)` call:

- `examples/cross-language/rust-sender/src/tests.rs:26`
- `examples/cross-language/rust-receiver/src/tests.rs:29`
- `examples/rust-dataflow/node/src/tests.rs:101`
- `examples/rust-dataflow/status-node/src/tests.rs:122` and `:169`
- `examples/rust-dataflow/sink/src/tests.rs:29`
- `examples/validated-pipeline/source/src/tests.rs:26`
- `examples/validated-pipeline/transform/src/tests.rs:29`
- `examples/validated-pipeline/sink/src/tests.rs:29`
- `examples/multiple-daemons/node/src/tests.rs:26`
- `examples/multiple-daemons/sink/src/tests.rs:29`

These files also `use dora_node_api::flume` and drain via `rx.try_iter().collect()`, so they need the same migration you applied to the in-crate helpers (a tokio `channel(...)` + a `try_recv()` drain loop).

Separately (non-blocking): the guide/docs still show the old `flume::unbounded()` + `try_iter()` `ToChannel` pattern (`guide/src/languages/rust.md`, `guide/src/development/testing.md`, `docs/testing-guide.md`, `docs/api-rust.md`).

The new `to_channel_delivers_outputs` test is a good addition — it covers the previously-untested end-to-end delivery path.

---

phil-opp:

Re-reviewed after `867bbc9` (the new commit since my last review). The build break I raised earlier is fixed: all ten in-workspace example test crates and the four guide/docs snippets are migrated off `flume::unbounded()` to `integration_testing::unbounded_channel()` + `drain_outputs()`, so `cargo test --all` should compile again.

The switch from a bounded `tokio::sync::mpsc::Sender` to `UnboundedSender` also resolves the deadlock concern from the first review: `OutputWriter::Channel` now calls the synchronous `UnboundedSender::send` (`apis/rust/node/src/daemon_connection/node_integration_testing.rs:169`), which cannot block the node while it awaits its own output's reply. The reasoning in the `ToChannel` doc comment (`apis/rust/node/src/integration_testing.rs:288-295`) matches the actual failure mode. Only the unbounded constructor is re-exported, so the bounded (deadlock-prone) path is unreachable through `dora-node-api`.

The new `to_channel_delivers_outputs_in_order` test meaningfully exercises the migrated path — three outputs sent, drained, and asserted on both count and send-order (`apis/rust/node/src/event_stream/mod.rs:2334-2372`). No issues found in the latest commit.

---

phil-opp:

One integration concern with another open PR.

In isolation the change looks sound: switching `TestingOutput::ToChannel` to a `tokio::sync::mpsc::UnboundedSender` removes the bounded-channel deadlock (a node blocking on its own output's reply while nothing drains), and only the unbounded constructor is re-exported so the deadlock-prone bounded path is unreachable. `to_channel_delivers_outputs_in_order` is a genuine end-to-end test.

However, this collides with the open PR #2959, which adds a test (`init_testing_node_mid_scheduled_wait`) that constructs `TestingOutput::ToChannel(tx)` from `flume::unbounded()`. Once this PR retypes `ToChannel`, that construction becomes a hard `flume::Sender` vs `tokio` `UnboundedSender` type mismatch, so whichever of the two merges second will fail `cargo test --all` compilation. The two also overlap textually in `node_integration_testing.rs`, `event_stream/mod.rs`, and `node/mod.rs`. Worth sequencing the merges and updating #2959's test to the new `unbounded_channel()` constructor.

---

PR #2973, opened 2026-08-03
fix(node): use tokio mpsc for testing output channel

### LINKED ISSUE #2956: TestingOutput::ToChannel still uses flume after EventStream migration (#1603)

SunSunSun689:
## Summary

  - migrate `TestingOutput::ToChannel` from `flume::Sender` to
  `tokio::sync::mpsc::Sender`
  - use `blocking_send` when integration-test output is emitted from the
  synchronous testing daemon path
  - update node API tests and docs examples to use tokio mpsc channels

  ## Why

  `EventStream` already moved from `flume` to `tokio::sync::mpsc` to avoid
  the flume/PyO3 async waker deadlock class tracked in #1603. The
  integration testing output channel was a remaining node API testing path
  that still exposed `flume::Sender`.

  Closes #2956.

### DIFF
--- a/apis/rust/node/src/daemon_connection/node_integration_testing.rs
+++ b/apis/rust/node/src/daemon_connection/node_integration_testing.rs
@@ -167,7 +167,7 @@ impl IntegrationTestingEvents {
             }
             OutputWriter::Channel(sender) => {
                 sender
-                    .send(output)
+                    .blocking_send(output)
                     .context("failed to send output to channel")?;
             }
         }
@@ -234,7 +234,7 @@ impl IntegrationTestingEvents {
 
 enum OutputWriter {
     Writer(Box<dyn Write + Send>),
-    Channel(flume::Sender<serde_json::Map<String, serde_json::Value>>),
+    Channel(tokio::sync::mpsc::Sender<serde_json::Map<String, serde_json::Value>>),
 }
 
 pub fn convert_output_to_json(
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -2314,7 +2314,7 @@ mod tests {
             "test-node".parse().unwrap(),
             events,
         ));
-        let (tx, _rx) = flume::unbounded();
+        let (tx, _rx) = tokio::sync::mpsc::channel(16);
         let outputs = TestingOutput::ToChannel(tx);
         let options = TestingOptions {
             skip_output_time_offsets: true,
@@ -2430,7 +2430,7 @@ mod tests {
             "test-node".parse().unwrap(),
             events,
         ));
-        let (tx, _rx) = flume::unbounded();
+        let (tx, _rx) = tokio::sync::mpsc::channel(16);
         let outputs = TestingOutput::ToChannel(tx);
         let options = TestingOptions {
             skip_output_time_offsets: true,
@@ -2497,7 +2497,7 @@ mod tests {
             "test-node".parse().unwrap(),
             events,
         ));
-        let (tx, _rx) = flume::unbounded();
+        let (tx, _rx) = tokio::sync::mpsc::channel(16);
         let outputs = TestingOutput::ToChannel(tx);
         let options = TestingOptions {
             skip_output_time_offsets: true,
--- a/apis/rust/node/src/integration_testing.rs
+++ b/apis/rust/node/src/integration_testing.rs
@@ -49,7 +49,7 @@
 //!    );
 //!
 //!    // send the node's outputs to a channel so we can verify them later
-//!    let (tx, rx) = flume::unbounded();
+//!    let (tx, mut rx) = tokio::sync::mpsc::channel(16);
 //!    let outputs = dora_node_api::integration_testing::TestingOutput::ToChannel(tx);
 //!
 //!    // don't include time offsets in the outputs to make them deterministic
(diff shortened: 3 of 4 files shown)


phil-opp:

No issues found. I traced the semantics of the flume → `tokio::sync::mpsc` switch:

- The one production-path change, `sender.blocking_send(output)` in `daemon_connection/node_integration_testing.rs:170`, runs on the plain `std::thread::spawn` worker in `node/mod.rs:1159` (driven by `receiver.blocking_recv()`), not inside a tokio task — so `blocking_send` won't hit the "cannot block within a runtime" panic. The full route is `send_message` → `IntegrationTestChannel` `blocking_send` → that worker thread → `events.request` → `handle_output`.
- Blocking-on-full and error-on-closed-receiver semantics are preserved (channel is now bounded at the caller-chosen capacity, matching the doc note).
- Test-side `rx.try_iter()` → `std::iter::from_fn(|| rx.try_recv().ok())` is equivalent: `try_recv().ok()` yields buffered items then returns `None` on `Empty`/`Disconnected`, terminating `from_fn` just as `try_iter` stops at momentary emptiness. The tests `drop(node)`/`drop(events)` first, so the sender is closed and all buffered outputs drain.

Consistent with the parallel change in #2972.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review with no human in the loop; please verify before acting.

I found the following issue:

**Build break: the `TestingOutput::ToChannel` retype isn't applied to the in-workspace example crates.** This PR changes the public `TestingOutput::ToChannel` variant to hold a `tokio::sync::mpsc::Sender` (`apis/rust/node/src/integration_testing.rs`), but only migrates the call sites *inside* `apis/rust/node`. Ten example crates — all workspace members, so compiled by the mandatory `cargo test --all` gate (PR CI excludes only the three Python packages) — still construct the variant from `flume::unbounded()`, which is now a hard `flume::Sender` vs `tokio::sync::mpsc::Sender` type mismatch at each `ToChannel(tx)`, plus the paired `rx.try_iter()` drains that no longer exist on a tokio receiver:

- `examples/rust-dataflow/node/src/tests.rs` (`ToChannel(tx)` + `try_iter()`)
- `examples/rust-dataflow/status-node/src/tests.rs`
- `examples/rust-dataflow/sink/src/tests.rs`
- `examples/multiple-daemons/node/src/tests.rs`
- `examples/multiple-daemons/sink/src/tests.rs`
- `examples/cross-language/rust-sender/src/tests.rs`
- `examples/cross-language/rust-receiver/src/tests.rs`
- `examples/validated-pipeline/source/src/tests.rs`
- `examples/validated-pipeline/transform/src/tests.rs`
- `examples/validated-pipeline/sink/src/tests.rs`

Each needs the same migration applied to its in-crate test helper: a bounded `tokio::sync::mpsc::channel(...)` plus a `try_recv()` drain loop (e.g. `std::iter::from_fn(|| rx.try_recv().ok())`) in place of `flume::unbounded()` + `try_iter().collect()`.

Separately (non-blocking): the guide docs still show the old `flume::unbounded()` + `try_iter()` `ToChannel` pattern in `guide/src/languages/rust.md` and `guide/src/development/testing.md`.

Note: PR #2972 targets the same issue (#2956) and carries the identical unfixed example-crate break — its most recent review already flagged this there.

---

phil-opp:

One point neither prior review has raised, on the test added in `apis/rust/node/src/integration_testing.rs`:

The new `testing_output_to_channel_accepts_tokio_mpsc_sender` test has no assertions — it constructs `TestingOutput::ToChannel(tx)` from a tokio sender and immediately drops it. It can therefore only ever fail by failing to *compile*, and that exact construction is already exercised (with real assertions on the received outputs) by the migrated helpers in `node/mod.rs` and `event_stream/mod.rs`, so it's fully redundant with them. As a runtime `#[test]` it passes regardless of the code under test and would not catch a regression. Consider dropping it, or making it observe something — e.g. send an output through the channel and assert it arrives via `try_recv`.

(The example-crate build break from `ToChannel`'s type change is already tracked in the most recent review, so I won't restate it here.)

---

phil-opp:
Closed in favor of https://github.com/dora-rs/dora/pull/2972

PR #2974, opened 2026-08-03
feat(cli): `--exit-when-nodes-finish` for `dora start`, via the descriptor

heyong4725:
Completes the first ask of #2920. [#2957](https://github.com/dora-rs/dora/pull/2957) shipped `dora run --exit-when-nodes-finish`; the issue asks for `run/start`, and `start` reaches the daemon by a different route, so the flag did not exist there.

## Why the descriptor, not the wire

`start` goes CLI → coordinator → daemon, and the setting has to survive three things `run` never faced: **auto-recovery re-spawn**, **coordinator restart** with state reconstruction, and **`dora restart`**.

Carrying it on the wire meant remembering it in five places — `ControlRequest::Start`, `SpawnDataflowNodes`, the coordinator's `RunningDataflow`, its persisted record plus reconstruction, and `PendingRestart`. Miss one and the setting silently vanishes. That is the same drop-one-copy shape that already cost this issue two rounds (the operator-node check in #2957, the pid guard in #2961).

The descriptor is already carried through every one of those. Putting it there means:

- **no protocol change at all** — the `Start` / `Spawn` fields I first added are reverted
- **essentially no coordinator change** — one single-node temp descriptor literal
- `dora run` folds its `RunDataflowOptions` onto the same field, so both entry points share one source of truth instead of a daemon-scoped flag plus a wire field
- it becomes settable in YAML, so a batch dataflow can declare its own policy

## Semantics

The flag overrides the descriptor in **both** directions, and is tri-state:

```bash
dora start flow.yml --exit-when-nodes-finish          # force on
dora start flow.yml --exit-when-nodes-finish=false    # force off
dora start flow.yml                                   # the YAML decides
```

Absence deliberately does **not** mean `false`. `run` and `start` are the only ways to start a dataflow, so "absent means off" would override a YAML `true` on every invocation and the descriptor field could never take effect. A test pins both halves, so a later "simplification" back to a plain bool fails with an explanation.

`DescriptorExt::apply_exit_when_nodes_finish` is the single place the field is written.

## Verification

All four combinations, end to end:

| YAML | flag | result |
|---|---|---|
| `true` | omitted | finishes — descriptor decides |
| `true` | `=false` | hangs — flag overrides off |
| unset | bare | finishes — flag overrides on |
| unset | omitted | hangs — default |

Mutation-checked against **production** code, not tests:

- daemon ignores the descriptor field → **both** the `run` and `start` e2e fail, each with its own diagnostic
- `expand()` drops the field → propagation test fails
- override reverted to "only turns on" → semantics test fails
- override changed to "absent means false" → semantics test fails

The propagation test carries a precondition assert: without a module in the fixture, `expand_modules` short-circuits to a whole-struct clone and never reaches the field-by-field rebuild, so an earlier version of that test passed under mutation and proved nothing.

Gate: fmt, clippy `-D warnings`, workspace tests, `cargo check --examples`, timer e2e (`run` + `start`), fault-tolerance (11), finish-straggler (1), node-lifecycle (10). No leftover processes.

## Docs and schema

- both CLI flag tables — the `dora run` row claimed a `false` default, which is no longer true
- `docs/yaml-spec.md`: root-level field plus a **Completion** section on why a timer stops a graph ending
- `libraries/core/dora-schema.json` regenerated

The root `dora-schema.json` is deliberately untouched: the generator does not write it and it was already 86 lines stale before this change. Worth a separate look.

## Note

`RunDataflowOptions::exit_when_nodes_finish` becomes `Option<bool>` to carry the third state. It shipped in #2957 today with the workspace version unchanged, so nothing is published with the old shape; the builder setter keeps its `bool` signature.

Refs #2920

### DIFF
--- a/binaries/cli/src/command/run.rs
+++ b/binaries/cli/src/command/run.rs
@@ -133,8 +133,19 @@ pub struct Run {
     /// Nodes whose inputs are all timers are unaffected: they have no
     /// data dependency that could finish, so they are treated as
     /// sources, exactly as a node with no inputs is.
-    #[clap(long, action)]
-    pub exit_when_nodes_finish: bool,
+    ///
+    /// Overrides `exit_when_nodes_finish:` in the dataflow YAML in
+    /// either direction: pass the flag to force it on, or
+    /// `--exit-when-nodes-finish=false` to force it off for a descriptor
+    /// that asks for it. Omit it entirely and the descriptor decides.
+    #[clap(
+        long,
+        num_args = 0..=1,
+        require_equals = true,
+        default_missing_value = "true",
+        value_name = "BOOL"
+    )]
+    pub exit_when_nodes_finish: Option<bool>,
 }
 
 impl Run {
@@ -154,7 +165,7 @@ impl Run {
             working_dir: None,
             hub_override: Vec::new(),
             env: Vec::new(),
-            exit_when_nodes_finish: false,
+            exit_when_nodes_finish: None,
         }
     }
 
@@ -287,7 +298,11 @@ impl Executable for Run {
                 working_dir_override,
                 // hub-resolved descriptor and/or `--env` merge — see above
                 descriptor_override,
-                RunDataflowOptions::default().exit_when_nodes_finish(exit_when_nodes_finish),
+                // Left unset the descriptor decides; given, it overrides.
+                match exit_when_nodes_finish {
+                    Some(v) => RunDataflowOptions::default().exit_when_nodes_finish(v),
+                    None => RunDataflowOptions::default(),
+                },
             )
             .await
         });
--- a/binaries/cli/src/command/start/mod.rs
+++ b/binaries/cli/src/command/start/mod.rs
@@ -60,6 +60,30 @@ pub struct Start {
     /// numeric-looking values that would be coerced. They are also
     /// persisted with the dataflow in the coordinator's state store and
     /// visible in `ps` — prefer a node `env:` block for secrets.
+    /// Exit once every node has finished, treating `dora/timer/...`
+    /// inputs as a clock rather than as work.
+    ///
+    /// A timer input never closes, so by default a dataflow in which any
+    /// node consumes one cannot end on its own, even after every node
+    /// doing real work has exited. With this flag a node is finished once
+    /// its DATA inputs have closed.
+    ///
+    /// Note this is usually what you want only for batch-style runs: a
+    /// long-lived dataflow is normally ended with `dora stop`, and there
+    /// the timer is exactly what keeps it alive.
+    ///
+    /// Overrides `exit_when_nodes_finish:` in the dataflow YAML in
+    /// either direction: pass the flag to force it on, or
+    /// `--exit-when-nodes-finish=false` to force it off for a descriptor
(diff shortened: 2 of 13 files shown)


phil-opp:

I reviewed the diff. The core mechanism looks correct: the tri-state `Option<bool>` on the descriptor, `apply_exit_when_nodes_finish` override semantics, and the daemon deriving `timers_gate_drain = !descriptor.exit_when_nodes_finish.unwrap_or(false)` at spawn all line up, and both `run` and `start` fold the flag onto the same descriptor field before spawn. One issue:

- **`binaries/cli/src/command/start/mod.rs`** — the new `exit_when_nodes_finish` field is inserted *between* the `env` field's doc comment (the block ending `…prefer a node \`env:\` block for secrets.`) and the `#[clap(long = "env")]` attribute it documents. Doc comments attach to the item that follows, so that whole `env` help block now attaches to `exit_when_nodes_finish` instead: the `--exit-when-nodes-finish` `--help` text will open with the env-secrets paragraph, and `--env` is left with no help text at all. Placing the new field (with its own doc) *after* the `env` field — or moving the env doc block back directly above `env` — fixes both help strings.

Not a runtime bug, but it does corrupt the generated `--help` for two flags.

---

heyong4725 (commented):
Reviewed `34789cbad`. I found two issues beyond the existing automated `--help` finding:

1. [P2] Apply the CLI override after hub fingerprinting and resolution (`binaries/cli/src/command/start/mod.rs:177`).

`DataflowSession::fingerprint_source` serializes the entire expanded descriptor, and the stored fingerprint was created by `dora build` before any transient CLI override. Applying `exit_when_nodes_finish` at line 177 therefore makes a real override look like an on-disk edit for every built `hub:` dataflow: a YAML with the field unset plus the bare flag changes `None` to `Some(true)`, while `=false` against YAML `true` changes the opposite way. The comparison at lines 190-191 then fails with `this dataflow changed since the last dora build`, so the advertised override cannot be used. Also, line 197 replaces the descriptor with the cached resolved copy, which would discard any early override if fingerprinting did not reject it. Move `apply_exit_when_nodes_finish` after the hub block, analogous to the existing rule that spawn-time `--env` is applied only after source fingerprint validation and resolved-descriptor substitution. Please add a regression test with a cached hub descriptor where the flag differs from the YAML value.

2. [P3] Update the schema that the documentation actually publishes (`dora-schema.json`).

This PR documents the new root field and says editor users should load `https://raw.githubusercontent.com/dora-rs/dora/main/dora-schema.json`, but only `libraries/core/dora-schema.json` contains `exit_when_nodes_finish`. The generator and schema-sync workflow also write/check only the library copy, so merging this PR will not repair the root schema later. The root schema allows extra properties, so YAML is not rejected, but users following the documented editor setup get no completion, type information, or description for this new setting. Update the root schema too, or consolidate the generator/docs so the published URL stays in sync.

The automated comment about the `env` doc block attaching to `exit_when_nodes_finish` is also correct and remains present at this head; I have not duplicated it as a third review finding.

Validation: `git diff --check` passed, and the targeted dora-core tests passed (3 passed): override semantics plus both descriptor-expansion cases. CI format/check/clippy are green, while the PR Test and E2E jobs are skipped.

PR #2975, opened 2026-08-03
test(e2e): guard that a dynamically added node gets ticks on a fresh timer interval (#2585)

phil-opp:
Both code fixes for #2585 already landed — #2474 made `AddNode` re-invoke
`RunningDataflow::start()`, #2810 made `RemoveNode` cancel a timer task whose
last subscriber went away. The coverage gap the issue itself called out did
not: nothing adds a node with a fresh timer interval and checks that ticks
actually arrive. `lifecycle_rust_dynamic_add_remove` adds a node wired to the
sender's output and asserts only that it reaches `Running` — which is exactly
what the bug looked like from the daemon side.

Adds `rust_dynamic_node_receives_ticks_on_fresh_timer_interval`: starts
`examples/rust-dynamic-add-remove` (timers at 200ms and 500ms), `dora node
add`s a node whose only input is `dora/timer/millis/300`, then asserts at
least 3 ticks reach it within 15s. The interval has to be one no running node
subscribes to — reusing an existing one passes even with the bug, because that
interval's task is already ticking and the delivery path re-reads the
subscriber set every tick.

Observing delivery needs a node that reports what it received, so this adds
the `timer-tick-recorder-node` fixture (appends each input id to
`DORA_TEST_MARKER_FILE`, same convention as `stop-delay-node`).

Checked against a revert of the `AddNode` → `start()` call: the test fails
with `got 0 ticks on the previously-unused interval 300ms`, and passes on
current `main`.

Also documents that a dynamic node's YAML can declare `inputs:` inline,
including timers — the guide only mentioned wiring them afterwards with `dora
node connect`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7754,6 +7754,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "timer-tick-recorder-node"
+version = "1.0.0-rc.4"
+dependencies = [
+ "dora-node-api",
+ "eyre",
+]
+
+[[package]]
 name = "tiny-keccak"
 version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,7 @@ members = [
     "tests/fault_tolerance/silent_source_node",
     "tests/fault_tolerance/stop_delay_node",
     "tests/fault_tolerance/timed_burst_source_node",
+    "tests/fault_tolerance/timer_tick_recorder_node",
     "xtask",
 ]
 
--- a/guide/src/operations/dynamic-topology.md
+++ b/guide/src/operations/dynamic-topology.md
@@ -29,9 +29,14 @@ id: filter
 path: filter.py
 outputs:
   - output
+inputs:
+  tick: dora/timer/millis/300
 ```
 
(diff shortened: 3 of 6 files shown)


phil-opp:

No issues found. This is a test-only change (a new `timer-tick-recorder-node` fixture plus a `#2585` regression test) and a docs tweak; it touches no production code path.

- The fixture node correctly opens the marker file *before* the event loop, so file-exists ("connected") is distinguishable from file-empty ("connected but no delivery") — the exact distinction the test relies on. It flushes per line, which is necessary since the test polls the file while the node runs, and the `MAX_LIFETIME` watchdog thread bounds an orphaned incarnation.
- The test asserts meaningful behavior, not a tautology: it picks a timer interval (300ms) that no node in `examples/rust-dynamic-add-remove` already subscribes to, so the added node is genuinely the first subscriber and a fresh tick task must be spawned — the case that reproduces `#2585`. It requires `>= 3` ticks (a single stray delivery can't pass) within a 15s deadline, and separately asserts the marker file exists so a never-connected node fails with a distinct message rather than a misleading zero-tick failure. The count filters lines equal to `"tick"`, matching the input id the recorder writes.
- Rerun safety is handled (marker removed before the run), and the build helper uses a `Once` guard consistent with the existing fixtures.

---

PR #2978, opened 2026-08-03
fix(daemon): isolate node lifecycle generations

### LINKED ISSUE #2926: daemon: two node-lifecycle paths still keyed by node id alone, so a dead incarnation can act on a live one
Latent follow-up from the #2916 triage. Neither of these reproduces today — both are read from the code, not observed — but they share the root cause and the kill mechanism that #2916 turned out to be, so they seem worth tracking rather than rediscovering.

## Background

#2916 was a `dora node remove` → `dora node add` (same id) race: the removed process's exit event was accounted against the freshly added incarnation, `handle_node_stop` removed the *new* `running_nodes` entry, and dropping that `ProcessHandle` SIGKILLed the live successor. `eec31a40b` (#2893) fixed it by carrying the exited pid through `NodeProcessFinished` → `SpawnedNodeResult` and discarding an exit whose pid no longer matches `running_nodes[node_id]` (`binaries/daemon/src/lib.rs:4875`).

That guard covers the `SpawnedNodeResult` path. Two neighbouring paths are still keyed by node id alone.

## 1. `ProcessHandleReplaced` can overwrite a live incarnation's handle

`binaries/daemon/src/lib.rs:5166`, installing at `:5182`:

```rust
node.process = Some(new_handle);
```

The event carries `dataflow_id` + `node_id` and no incarnation identity, so a `restart_loop` belonging to a dead incarnation can install its handle over whatever entry currently owns that id. Overwriting drops the previous `ProcessHandle`, and `ProcessHandle::drop` submits `Kill` (`binaries/daemon/src/running_dataflow.rs`) — the exact mechanism behind #2916, reached by a different route.

The existing comment there reasons that the dropped handle's `Kill` hits an already-closed `op_rx` and is a no-op. That holds when the overwritten handle belongs to the *same* node's dead predecessor. It does not hold if the entry has since been replaced by a `remove` + `add` of the same id, where the handle being dropped is live.

## 2. A stale `dataflow_node_results` entry can outlive the id

`binaries/daemon/src/lib.rs:5160`:

```rust
.insert(node_id.clone(), node_result);
```

When a removed incarnation's exit lands *before* the re-add, the pid guard cannot fire — `RemoveNode` took the `running_nodes` entry out and `AddNode` hasn't arrived, so there is nothing to compare against — and the failure is recorded under the bare node id. Nothing clears that entry when the id is re-added.

Today this is benign only incidentally: the successor's own exit overwrites the map entry at teardown. It stops being benign if the successor never reports a result — a dynamic node, which doesn't emit `SpawnedNodeResult`. The removed incarnation's failure would then stand as the node's recorded result, and `dora stop` would report the dataflow failed while `dora node list` shows the node running.

Confirmed empirically during triage that the stale `Err` *is* written (daemon log shows `node exited with error: ExitCode(1)` accounted with no live node), and that `dora node list` drops the row on removal and never reappears to report the later exit — so there is no observable for it from the CLI side.

## Suggested direction

A monotonic incarnation/generation id on `RunningNode`, stamped into both events and compared on receipt, would close both plus the pid-reuse weakness in the current pid comparison. The pid guard is a narrower version of the same idea and could fold into it.

## Coverage

`tests/node-lifecycle-e2e.rs` (#2925) now pins both orderings of the remove→re-add race, including a marker-file assertion that the removed incarnation really exited the way each test assumes. Neither of the paths above is covered — the `ProcessHandleReplaced` one needs a restart racing a re-add, and the `dataflow_node_results` one needs a dynamic-node successor.

heyong4725:
## Summary

- assign a monotonic generation to each node process incarnation and reject stale lifecycle events
- scope planned-stop markers and stored node results so removed or restarted nodes cannot affect a live successor
- gate restart-loop events until node registration and make teardown win concurrent respawn races
- add regression coverage for stale exits, stale handle replacement, PID reuse, registration ordering, and result cleanup

## Validation

- `cargo check --examples`
- `cargo test -p dora-daemon fault_tolerance_tests:: -- --test-threads=1` (47 passed)

The targeted `node-lifecycle-e2e` invocation could not start because the existing default local coordinator store uses schema v2 while this checkout expects schema v3. The store was left untouched.

Closes #2926

### DIFF
--- a/binaries/daemon/src/event_types.rs
+++ b/binaries/daemon/src/event_types.rs
@@ -175,6 +175,7 @@ pub enum DoraEvent {
     SpawnedNodeResult {
         dataflow_id: DataflowId,
         node_id: NodeId,
+        generation: u64,
         dynamic_node: bool,
         exit_status: NodeExitStatus,
         restart: bool,
@@ -191,6 +192,8 @@ pub enum DoraEvent {
     ProcessHandleReplaced {
         dataflow_id: DataflowId,
         node_id: NodeId,
+        previous_generation: u64,
+        new_generation: u64,
         new_handle: crate::ProcessHandle,
     },
 }
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1712,14 +1712,14 @@ impl Daemon {
                     dynamic_node,
                     result,
                 } => match result {
-                    Ok(running_node) => {
+                    Ok(mut running_node) => {
                         if let Some(dataflow) = self.running.get_mut(&dataflow_id) {
+                            // Open the restart-loop gate before the insert:
+                            // the loop's first events are queued behind this
+                            // handler on the same event loop, so they cannot
+                            // be processed before the entry is registered.
+                            running_node.mark_registered();
                             dataflow.running_nodes.insert(node_id.clone(), running_node);
-                            dataflow
-                                .running_nodes
-                                .get_mut(&node_id)
-                                .expect("running node was just inserted")
-                                .mark_registered();
                         } else {
                             tracing::error!(
                                 "failed to handle SpawnNodeResult: no running dataflow with ID {dataflow_id}"
@@ -2344,7 +2344,7 @@ impl Daemon {
                         .await
                         .wrap_err("failed to prepare node")?;
                     let prepared = task.await.wrap_err("failed to build node")?;
-                    let running_node = prepared
+                    let mut running_node = prepared
                         .spawn(logger)
                         .await
                         .wrap_err("failed to spawn node")?;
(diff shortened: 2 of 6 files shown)


phil-opp:
I reviewed the diff and found no issues.

Things I checked specifically (the failure modes a per-incarnation generation is meant to fix):

- **Stale exit vs. handle-replacement ordering.** `SpawnedNodeResult` and `ProcessHandleReplaced` are emitted from the same `restart_loop` task on the same ordered `daemon_tx` channel, so a restart exit carrying `generation=G` is handled while `node.generation` is still `G` (the `G→G+1` replacement hasn't been processed yet). Dropping the old `!restart` guard on the staleness check therefore doesn't discard legitimate restart exits.
- **`grace_duration_kills` key migration.** All insert/contains/remove sites now use the generation-scoped `(NodeId, u64)` key; no `NodeId`-only lookup survives that could match a stale incarnation.
- **Registration gate is leak-free.** Both production `running_nodes.insert` sites call `mark_registered()`; if the dataflow is gone, the `RunningNode` (holding `registered_tx`) drops, `registered.await` errors, the loop returns, and the spawned process's `ProcessHandle` drops → Kill. An early exit before registration is buffered in the `finished` oneshot and consumed once the gate releases.
- **`replace_process_handle` rejection paths.** On generation mismatch or `restarts_disabled()`, the moved `new_handle` drops → Kill to the orphan.
- **`clear_node_result` on AddNode** removes only the re-added node's stale terminal result and prunes now-empty dataflow maps.

The 8 added tests exercise real behavior (PID-reuse rejection, the registration-gate ordering, generation-scoped kill markers, `clear_node_result` selectivity, and the three `replace_process_handle` outcomes with concrete assertions on `Kill` delivery) — not vacuous.

---
_This is a fully automated review by Claude. No human has reviewed or verified these comments._

---

Issue #2980, opened 2026-08-03
Coordinator: `handle_destroy` destroys daemons before the stop it requested can take effect, orphaning wedged nodes


heyong4725:
## Summary

`handle_destroy` tears the daemons down before the stop it just requested can take effect, so any node that does not exit on the cooperative `Stop` is orphaned. `dora down` documents "Stops any running dataflows first", which holds only for well-behaved nodes.

## Where

`binaries/coordinator/src/handlers.rs`, `handle_destroy`:

```rust
for dataflow_uuid in running_dataflows.keys().cloned().collect::<Vec<_>>() {
    stop_dataflow(running_dataflows, dataflow_uuid, daemon_connections,
                  clock.new_timestamp(), None /* grace */, false /* force */).await
}
destroy_daemons(daemon_connections, clock.new_timestamp()).await?;
tokio::time::sleep(Duration::from_millis(200)).await;
```

Two things combine:

1. `stop_dataflow` awaits the daemon's **acknowledgement of the request**, not the dataflow finishing. The escalation ladder (grace → SIGTERM → SIGKILL) then runs asynchronously inside the daemon.
2. It is called with `grace_duration: None` and `force: false`, i.e. the default **10s** grace — and `destroy_daemons` runs ~200ms later.

So the daemon is destroyed roughly 10s before the rung that would have killed a wedged node.

## Reproduction

With a node that honors neither the cooperative `Stop` nor SIGTERM (`tests/fault_tolerance/sigterm_ignoring_node`, added in #2961 for exactly this class of thing):

```
$ dora up && dora start hung.yml --detach --name hung
$ dora down --force        # pre-#2924-guard behavior
Coordinator and daemons destroyed successfully     <-- returns in <1s

node alive=1  coordinator=0  daemon=0
```

The node survives with `ppid 1`. By contrast `dora stop --all` on the same dataflow does clean up, taking ~10s for the ladder to reach SIGKILL — so the machinery works, `destroy` just does not wait for it.

## Why it matters

- It is the difference between the documented and actual contract of `dora down`.
- The orphan holds whatever the node held: shared-memory segments, GPU contexts, ports, CPU. #2924 describes leaked nodes at high CPU skewing later timing-sensitive runs.
- It is invisible: the command reports success.

## Scope note

#2924's fix makes **`dora down --force`** stop each dataflow through the path that does wait, so that one route is covered. Everything else that reaches `Destroy` is not — notably `dora cluster down`, and any direct control-protocol client.

[...]

Issue #2924, opened 2026-07-31
Safety/UX: `dora destroy` silently destroys an UNRELATED instance's coordinator (shared default port, no confirmation, active dataflows ignored)


heyong4725:
**Versions:** observed between dora-cli 1.0.0-rc.4 (git rev `7eb4a5f8b`, release install) and a debug build from a development checkout, same machine. macOS arm64.

## What happens

Two independent dora "instances" on one machine — a release CLI running a
long experiment, and a development checkout whose test/dev loop runs
`target/debug/dora destroy` — share the default coordinator port
`127.0.0.1:6013`. The dev checkout's `destroy` connects to whatever
coordinator owns the port and destroys it, killing the OTHER project's
daemon and every running dataflow, with no confirmation and no warning
that dataflows are active:

```
20:21:06.674 INFO Received destroy command
20:21:06.674 INFO successfully destroyed daemon `019fb9d6-...`
20:21:06.876 INFO stopped
```

(caught via a log tripwire that captured the sender:
`/Users/<u>/src/dora-rs/dora/target/debug/dora destroy`.)

From the victim's side this is indistinguishable from a coordinator
crash: subsequent `dora start` fails with connection refused, and a
multi-hour experiment dies mid-run. It took us three such kills and a
process-capture tripwire to identify the cause.

## Asks (any subset helps)

1. `dora destroy` should refuse (or prompt / require `--force`) when the
   target coordinator reports RUNNING dataflows.
2. First-class multi-instance isolation: a `--context`/profile concept, or
   at minimum prominent docs that `DORA_COORDINATOR_PORT` is the isolation
   mechanism and that all lifecycle commands are port-global.
3. `dora destroy` could print WHICH coordinator (pid, uptime, dataflow
   count) it is about to destroy.

`DORA_COORDINATOR_PORT` works as a workaround (we now run experiments on
a private port), but nothing warns the default-port user that their
`destroy` has machine-wide blast radius.

heyong4725:
Fixed on `main` in #2981.

## Ask 1 — refuse when dataflows are running

`dora down` (and `destroy`) now asks what it is about to destroy and stops:

```
$ dora down
[ERROR]
refusing to destroy the coordinator at 127.0.0.1:6013: 1 dataflow(s) still running
  - [victim] 019fc838-d8b4-7587-b7d5-adad0299df7e

  This tears down the coordinator, its daemons and every dataflow above.
  Lifecycle commands target whatever coordinator owns the port, so this may
[...]

PR #2981, opened 2026-08-03
fix(cli): guard `dora down`, and make `--force` actually clean up

heyong4725:
Addresses #2924.

## The problem

Every lifecycle command reaches whatever coordinator owns the port, and two checkouts on one machine share `127.0.0.1:6013` by default. A `dora destroy` in a dev loop therefore tears down an unrelated project's coordinator, its daemons and every dataflow on it — no confirmation, no indication anything was running. From the victim's side it looks exactly like a coordinator crash. The reporter needed three kills and a process-capture tripwire to identify the cause.

## The guard

```
$ dora down
[ERROR]
refusing to destroy the coordinator at 127.0.0.1:6013: 1 dataflow(s) still running
  - [victim] 019fc838-d8b4-7587-b7d5-adad0299df7e

  This tears down the coordinator, its daemons and every dataflow above.
  Lifecycle commands target whatever coordinator owns the port, so this may
  not be the instance you meant — a different checkout or another project on
  this machine shares 127.0.0.1:6013 by default.

  hint: stop the dataflows first (`dora stop --all`), or re-run with `--force`
        if you do mean to kill them
  hint: to keep instances isolated, give each one its own port via
        `DORA_COORDINATOR_PORT` (or `--coordinator-port`)
```

It names the cause, not just the symptom: the surprise is not that destroy destroys, it is that it reached an instance you did not have in mind. The idle case is untouched, deliberately — `up`/`down` loops stay frictionless so nobody learns to type `--force` reflexively.

## `--force` had to be fixed before it could be recommended

It did not clean up. `Destroy` sends each dataflow a stop and then tears the daemons down **immediately**, without waiting for the stop ladder (grace → SIGTERM → SIGKILL). A node that exits promptly on `Stop` is reaped; a wedged one outlives its daemon and is orphaned.

Measured on a node that ignores **both** the cooperative stop and SIGTERM:

| | before | after |
|---|---|---|
| `dora down --force` | "destroyed successfully" in <1s | stops each dataflow first |
| wedged node | **still running** | SIGKILLed and reported |
| coordinator / daemon | gone | gone |

So `dora down`'s own documentation — *"Stops any running dataflows first"* — held only for well-behaved nodes. A guard whose escape hatch doesn't work would have been worse than no guard.

`--force` now stops each dataflow through the path that does wait, then destroys. I also verified the other route the error suggests: `dora stop --all` does work on a wedged dataflow (~10s via the grace ladder). Both ways out are real.

## Against the three asks

1. **Refuse when dataflows are running** — done, `--force` as the escape hatch, and it now genuinely cleans up.
2. **Isolation** — docs, not a new concept. Both CLI copies get a "machine-wide, not project-wide" callout with a working side-by-side recipe, and `DORA_COORDINATOR_PORT` is documented as *the* isolation mechanism rather than just "Coordinator WebSocket port". A first-class `--context`/profile is a design project, not a bug fix.
3. **Say which coordinator** — the refusal names the address and every running dataflow; the success path announces what it is destroying. PID and uptime aren't reachable over the control protocol today and would need a new request type.

## Verification

`down_decision` is a separate function rather than an inline check so the rule is unit-testable. Mutation-checked against production code both ways: dropping the guard fails `down_refuses_when_dataflows_are_running`; ignoring `--force` fails `force_destroys_despite_running_dataflows`.

End to end: idle `down` works; running `down` refuses and **the victim survives**; `--force` cleans up a node that ignores stop and SIGTERM; the `destroy` alias behaves identically.

Gate: fmt and clippy `-D warnings` pass locally. The full local sweep (workspace tests, `cargo check --examples`, `ws-cli-e2e`, `node-lifecycle-e2e`) was still running when this was opened — those suites call `up`/`down` constantly, which is how a teardown guard would break something unit tests can't see, so CI's run of them is the gate that matters here.

## Known limits

- **The underlying `handle_destroy` bug is untouched.** A direct `Destroy` still orphans wedged nodes; only the CLI's `--force` path compensates. Filed separately as #2980, since fixing it in the coordinator changes behavior for every caller and the right waiting policy is a judgment call.
- **`dora cluster down` keeps its current unconditional behavior**, with an explicit argument and a comment. It is an explicit cluster teardown and was not the reported footgun; a guard would break existing teardown scripts.
- **A failed pre-flight `List` warns and proceeds.** Refusing on a transport hiccup would make teardown unusable — a worse failure than the one being fixed.

Refs #2924

### DIFF
--- a/binaries/cli/src/command/cluster/down.rs
+++ b/binaries/cli/src/command/cluster/down.rs
@@ -20,6 +20,10 @@ pub struct Down {
 impl Executable for Down {
     fn execute(self) -> eyre::Result<()> {
         default_tracing()?;
-        up::down(None, self.coordinator.socket_addr())
+        // `dora cluster down` keeps its existing unconditional behavior: it
+        // is an explicit cluster teardown, and the #2924 report is about
+        // `dora down`/`destroy` hitting an unrelated instance on the shared
+        // default port. Revisit if the same footgun shows up here.
+        up::down(None, self.coordinator.socket_addr(), true)
     }
 }
--- a/binaries/cli/src/command/down.rs
+++ b/binaries/cli/src/command/down.rs
@@ -10,11 +10,30 @@ pub struct Down {
     config: Option<PathBuf>,
     #[clap(flatten)]
     coordinator: CoordinatorOptions,
+    /// Destroy even if the coordinator still has running dataflows,
+    /// terminating them immediately first.
+    ///
+    /// Without this, `dora down` refuses when the target coordinator
+    /// reports running dataflows. Lifecycle commands reach whatever
+    /// coordinator owns the port, so the instance you hit may belong to
+    /// another checkout or project on the same machine
+    /// (dora-rs/dora#2924). Give each instance its own
+    /// `DORA_COORDINATOR_PORT` to keep them apart.
+    ///
+    /// This stops each dataflow before tearing the daemons down. `Destroy`
+    /// on its own does not wait for the stop to take effect, so a wedged
+    /// node would outlive its daemon and be orphaned.
+    #[clap(long, action)]
+    force: bool,
 }
 
 impl Executable for Down {
     fn execute(self) -> eyre::Result<()> {
         default_tracing()?;
-        up::down(self.config.as_deref(), self.coordinator.socket_addr())
+        up::down(
+            self.config.as_deref(),
+            self.coordinator.socket_addr(),
+            self.force,
+        )
     }
 }
--- a/binaries/cli/src/command/stop.rs
+++ b/binaries/cli/src/command/stop.rs
@@ -121,7 +121,7 @@ fn stop_dataflow_interactive(
     Ok(())
 }
 
-fn stop_dataflow(
+pub(crate) fn stop_dataflow(
     uuid: Uuid,
     grace_duration: Option<Duration>,
     force: bool,
--- a/binaries/cli/src/command/up.rs
+++ b/binaries/cli/src/command/up.rs
@@ -5,7 +5,10 @@ use crate::{
     common::{connect_to_coordinator, connect_with_retry, send_control_request},
 };
 use dora_core::topics::DORA_COORDINATOR_PORT_WS_DEFAULT;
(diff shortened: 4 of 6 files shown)


heyong4725 (commented):
Reviewed `a2babcb0e`. I found two blocking lifecycle issues:

1. [P1] Include recovering dataflows in the destroy guard (`binaries/cli/src/command/up.rs:424`).

`List` only exposes entries in the coordinator in-memory `running_dataflows` map. On coordinator startup, however, persisted Pending/Running/Stopping records are changed to `Recovering` and deliberately left out of that map until a daemon reconnects and sends its status report; the recovery window is 60 seconds. During that window a plain `dora down` receives an empty active list and proceeds to `Destroy` even though the daemon and its nodes can still be alive. If the daemon has not reconnected, `daemon_connections` is empty too, so the coordinator exits without touching it; if it reconnects between the preflight and destroy, this falls into the known non-waiting destroy behavior from #2980. That recreates the unmanaged-workload failure this guard is intended to prevent. Make the authoritative destroy decision include persisted Pending/Running/Stopping/Recovering records, preferably atomically in the coordinator rather than as a client-side List/Destroy check, and add a restart/recovery regression test.

2. [P1] Do not continue to unsafe Destroy after a force-stop error (`binaries/cli/src/command/up.rs:458`).

The code warns and continues whenever `stop_dataflow(..., force=true)` fails, but the next operation is exactly the `Destroy` path that this PR documents as orphaning nodes. A concrete case is a multi-daemon flow with one missing daemon connection: the coordinator stop RPC fails before that daemon receives the force kill, the CLI ignores the error, `handle_destroy` retries with `force=false` and also only logs its error, then destroys the remaining connected daemons and can report success while the disconnected daemon and nodes survive. Either abort teardown and return the aggregated stop failures while the coordinator remains available, or first fix `handle_destroy` so this fallback itself guarantees process cleanup. Continuing is incompatible with the new `--force` cleanup contract.

One smaller documentation mismatch also remains: `binaries/cli/src/command/down.rs:6` still says every `dora down` stops running dataflows first, but the new default behavior refuses and only `--force` stops them.

Validation: `git diff --check` passed. All three new decision tests passed (2 through Cargo plus the force case directly from the compiled test binary). CI format/check/clippy are green; the PR Test and E2E jobs are skipped.

PR #2982, opened 2026-08-03
fix(tests): make the sigterm-ignoring-node fixture compile on Windows (#2742)

phil-opp:
`tests/fault_tolerance/sigterm_ignoring_node` does not compile for any Windows
target:

```
error[E0308]: mismatched types      previous != libc::SIG_ERR
error[E0277]: can't compare `usize` with `i32`
```

`libc::signal` returns a `sighandler_t` (`usize`) everywhere, but `libc::SIG_ERR`
is a `c_int` on Windows and a `sighandler_t` on Unix, so the guard added in #2961
(`1a947a3c7`, merged 2026-08-02) is a type error there. PR CI is ubuntu-only, so
it merged green and broke two nightly jobs the same night:

- `Test (windows-latest) [nightly]`
- `Cross-check (x86_64-pc-windows-gnu)`

both in [run 30802384898](https://github.com/dora-rs/dora/actions/runs/30802384898).

The fixture is Unix-only in substance — it fakes a SIGTERM disposition, and its
only caller (`run_killed_by_sigterm_terminates_nodes_and_exits`) is `#[cfg(unix)]`
— but cargo has no cfg-conditional `members`, so it stays a workspace member on
every platform and must still compile. So: scope `libc` to `cfg(unix)` and give
Windows a stub `main` that exits non-zero.

Widening the comparison with `libc::SIG_ERR as usize` was rejected — the cast is
a no-op on Unix (and trips `clippy::unnecessary_cast`) and would keep dead
signal-handling code alive on a platform with no signals.

Reproduced and verified locally against `x86_64-pc-windows-gnu`: the pre-fix code
fails with the two errors above, the post-fix code checks clean for both
windows-gnu and the host.

Refs #2742

### DIFF
--- a/tests/fault_tolerance/sigterm_ignoring_node/Cargo.toml
+++ b/tests/fault_tolerance/sigterm_ignoring_node/Cargo.toml
@@ -21,11 +21,15 @@ path = "src/main.rs"
 dora-node-api = { workspace = true, default-features = false }
 eyre = { workspace = true }
 
-# Unix-only: on Windows `libc::signal` returns a `sighandler_t` (`usize`)
-# while `libc::SIG_ERR` is a `c_int`, so the check in `main` does not even
-# compile there. The fixture has no Windows build to guard, so the
-# dependency is scoped rather than the comparison widened with a cast —
-# a cast would be a no-op on Unix (and trip `clippy::unnecessary_cast`)
-# while keeping dead signal code alive on a platform without signals.
+# Scoped to unix, matching the `#[cfg(unix)]` on `main`: it keeps libc
+# out of the Windows dependency graph entirely, and makes "this fixture's
+# signal handling is Unix-only" enforced by the build rather than by
+# convention — uncfg'd `libc::` code added later fails to compile on
+# Windows instead of silently becoming part of the stub's build.
+#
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found.

The fix is correct. Scoping `libc` to `[target.'cfg(unix)'.dependencies]` while gating the real `fn main` on `#[cfg(unix)]` and providing a `#[cfg(not(unix))] fn main` stub resolves the Windows build break cleanly — the crate is a workspace member (cargo has no cfg-conditional `members`), so it must still compile under `cargo check --all` / `cargo test --all` on Windows, and every platform now has exactly one `main` for the `[[bin]]` target. Choosing a stub that exits non-zero over widening the guard with a `SIG_ERR as usize` cast is the better call: the cast would be a no-op on Unix (tripping `clippy::unnecessary_cast`) and would keep dead signal-handling code compiling on a platform with no signals.

The moved `use` statements are confined to the `#[cfg(unix)]` body, so the non-unix build pulls in nothing it doesn't use; `eyre` remains used on the Unix path. No test accompanies this, which is expected — it is a build-portability fix to a test fixture, and the fixture's Unix-only caller (`run_killed_by_sigterm_terminates_nodes_and_exits`) is already `#[cfg(unix)]`.

---

PR #2983, opened 2026-08-03
fix(daemon): survive an output that arrives after its dataflow finished (#2742)

phil-opp:
The daemon dies when a node's output arrives after its dataflow finished.

`handle_node_event` returns into the daemon's main loop, which treats any `Err`
as fatal — it unwinds `run()` and drops the coordinator connection. `SendOut` and
`OutputSent` handed it one whenever the dataflow was already gone. They are the
only two arms that can: neither carries a reply channel, so unlike every sibling
arm they cannot report the error back to the node.

That state is normal, not a bug:

- a node's exit and its already-transmitted outputs reach the daemon on
  independent paths, so a burst sent just before exit can be queued behind the
  `SpawnedNodeResult` that finished the dataflow;
- `should_finish` deliberately ignores still-running **dynamic** nodes, so a
  dynamic node is expected to outlive `finish_dataflow` entirely.

Under `dora run` (an in-process daemon) the first shape fails the whole run,
which is the intermittent nightly `smoke-suite` failure in #2742 — red on 07-31
and 08-02, green on 08-01 and 08-03:

```
[ERROR] failed to send out
Caused by: send out failed: no running dataflow with ID `019fc1a4-...`
```

Now the daemon warns and drops the message. It warns once per (dataflow, node),
then falls to `debug` — the dynamic-node case is unbounded, and a 1 kHz node
would otherwise emit 1000 warn lines a second for as long as it lived.

I reproduced the original race directly (1 failure in 12 runs of
`examples/rust-operator-teardown`, matching CI's rate), but the regression test
reaches the same state structurally instead of by winning it: a dynamic node
keeps sending while its anchor — the dataflow's only non-dynamic node — exits on
the sender's first message, so every later send is a post-finish one. No sleeps
to tune. The anchor is the existing `emit-then-exit-source-node` fixture, wired
to the sender's output rather than to a timer.

Verified RED (the daemon logs the fatal line and drops its coordinator
connection) → GREEN.

### Not fixed here

`handle_node_stop_inner` still `bail!`s on the same "dataflow already gone"
condition when a **non-dynamic** node's exit lands after `finish_dataflow`
(`binaries/daemon/src/lib.rs`, the `failed to get downstream nodes` arm). Same
root cause, narrower race, and it needs its own repro and test — worth a
follow-up issue rather than widening this change.

Also worth recording: dropping a late message is not always free. `should_finish`
only inspects the local daemon's non-dynamic nodes, so a still-running local
dynamic consumer, or one on another daemon via `open_external_mappings`, can
still have been waiting for it. That is a pre-existing finish-ordering gap and is
strictly better than taking the daemon — and every other dataflow's connection —
down for it. It is why this warns rather than dropping silently.

Refs #2742


phil-opp:

No issues found.

The core fix is correct: `handle_node_event` treats any `Err` as fatal (it unwinds `run()` and drops the coordinator connection), and `SendOut`/`OutputSent` are the only two arms with no reply channel, so returning `Err` on a missing dataflow was the wrong lever — the `let Some(dataflow) = ... else { log_late_node_output(...); return Ok(()) }` short-circuit is the right one. When the dataflow is already gone there are no live receivers on this daemon to route to, so dropping is safe on the local path (the doc correctly notes the residual remote/dynamic-consumer finish-ordering gap, which is pre-existing and strictly better than taking the daemon down).

Rate-limiting and its bound are sound: warn-once-per-`(dataflow_id, node_id)` then `debug!`, with the cap-reached case clearing the dedup set rather than freezing it — freezing would permanently silence the warning (including the stale-node case) after ~1024 dataflows on a long-lived daemon. All mutation happens under `&mut self` on the single-threaded event loop, so the plain `HashSet` is fine, and the struct-literal init guarantees the new field can't be silently omitted from another constructor.

On the new test (`late_output_from_a_finished_dataflow_does_not_kill_the_daemon`): it is meaningful and does not pass vacuously. Wiring the anchor's `tick` input to the dynamic sender's `late` output (so send #0 finishes the dataflow and every later send is post-finish) reaches the state structurally rather than by racing sleeps; waiting for *either* `LATE_OUTPUT_WARNING` or `FATAL_LATE_OUTPUT` means a regression fails on the fatal line it actually logs instead of on a 60s timeout, and a timeout correctly signals the fixture stopped reproducing the state. Matching `FATAL_LATE_OUTPUT` on the substring shared by both `send out failed:` and `output sent failed:` covers both delivery paths, and asserting survival via `try_wait()` (not `kill -0`, which succeeds on an unreaped zombie) is the right check.

---

PR #2984, opened 2026-08-03
ci(nightly): give the ROS2 Zenoh interop jobs the maturin they need (#2742)

phil-opp:
`scripts/ros2-zenoh-interop.sh` builds the ROS2 bridge wheel with `maturin build`
and installs it with pip, but the two jobs only set up a Rust toolchain. Every
run since they were added (#2790, 2026-07-21) has died at the same line, after
spending ~2 minutes pulling and booting the ROS container:

```
scripts/ros2-zenoh-interop.sh: line 49: maturin: command not found
##[error]Process completed with exit code 127
```

Both jobs have failed identically every night for two weeks and have never
passed. Nobody noticed because they are `continue-on-error: true`, which makes
`needs.<job>.result` read `success` — so `file-issue-on-failure` never listed
them and the nightly issue (#2742) never mentioned them.

Add `actions/setup-python@v6` + `pip install maturin`, mirroring the
`ros2-bridge` job's Python setup, and preflight `docker`/`maturin` in the script
so a missing tool is named up front with a fix hint instead of surfacing as a
bare 127 after the container pull.

Left advisory for now — these drive real ROS 2 containers, and a job that has
never run green should not start blocking nightlies on its first day. There is a
comment recording that `continue-on-error` is what hid the failure and that it
should go once the jobs are trustworthy.

Verified the preflight fires with the expected message and exit code, and that
argument validation still runs first.

Refs #2742

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1786,6 +1786,13 @@ jobs:
         run: cargo test -p dora-ros2-bridge-python
 
   # ===== MSRV check =====
+  # Advisory (`continue-on-error`) until these have run green for a
+  # stretch: they drive real ROS 2 containers, so a docker/apt hiccup
+  # should not redden a nightly. NOTE that this also hides them from
+  # `file-issue-on-failure` — `needs.<job>.result` reads `success` for a
+  # continue-on-error job — which is why they failed silently from the
+  # day they were added (#2790, 2026-07-21) until #2742 went looking.
+  # Drop `continue-on-error` once they are trustworthy.
   ros2-zenoh-humble:
     name: ROS2 Zenoh Interop (Humble)
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found.

The change correctly addresses the exit-127 failure: `ros2-zenoh-interop.sh` builds the bridge wheel with `maturin` and installs it with pip, but the two jobs only set up a Rust toolchain — so adding `actions/setup-python@v6` + `pip install maturin` (mirroring the existing `ros2-bridge` job) is the right fix. The script-side preflight is sound: `command -v "$tool" >/dev/null || missing+=(...)` won't trip `set -e`, the `((${#missing[@]}))` guard reads the array only when it is non-empty, and it runs after the CASE validation but before the ~2-min container pull, so a missing tool is named up front instead of surfacing as a bare 127 mid-run.

Two observations, neither blocking:
- The jobs remain `continue-on-error: true`, so their `needs.<job>.result` still reads `success` and they stay invisible to `file-issue-on-failure`. The added comment acknowledges this and flags it for removal once the jobs run green for a stretch — so this is intentional, but the maturin fix does not by itself restore failure visibility.
- No automated test accompanies this (expected — it is a CI-config/shell change); the preflight itself is not exercised by a test, but it is straightforward.

---

PR #2985, opened 2026-08-03
ci(nightly): split the Python half out of cli-tests so Windows fits its budget (#2742)

phil-opp:
`CLI Tests (windows-latest)` has not completed since 2026-07-16. Since 07-30 it
has been killed by its own `timeout-minutes: 100` partway through "Test Python
Operator example" — the step #2742 is actually about — so the #2885 diagnostics
added to pin that wedge have never had a chance to report.

Step times from the 08-02 nightly, with 07-29 (the last run that finished) in
brackets:

| step | 08-02 | 07-29 |
|---|---|---|
| Build CLI | 28.8 min | 21.7 |
| Set up Python venv | 25.2 min | 17.8 |
| Test Python template project | 13.0 min | 9.9 |
| Test Python Dynamic Node example | 12.9 min | 10.1 |
| Test Dynamic Rust Dataflow | 8.3 min | 6.2 |
| Test Rust template project | 5.2 min | 4.0 |
| Test Python Operator example | 5.6 min — **killed at the job cap** | 10.7 |

Every step is ~30% slower than it was, and the fixed cost ahead of the operator
step now eats 94 of the 100 minutes.

Two things made this worse than a slow job. A timed-out job reports `cancelled`,
not `failure`: `file-issue-on-failure` only lists `failure`, so the nightly issue
never named it, and `close-issue-on-success` refuses to close on any `cancelled`
need — so #2742 could not auto-close no matter what else went green.

Split rather than raise the cap, which the previous comment explicitly ruled out
("Do not raise further — investigate per-step times instead"). `cli-tests` keeps
the CLI build and the native steps (~43 min on Windows) and publishes the CLI it
already built; `cli-tests-python` downloads that artifact and runs the Python
steps (~63 min) without repeating the 29-minute build, so the split costs no
extra build time. Both halves are wired into the issue-filing and issue-closing
gates, and the nightly job count (now 23) is refreshed in `CLAUDE.md` and
`ci-nightly-jobs.sh`.

The second commit fixes three breakages a review of the first one caught, all of
which would have shipped:

- `Install CLI to PATH` ran `dora --version` in the same step that wrote
  `$GITHUB_PATH`, which only takes effect for *later* steps — exit 127 on all
  three platforms, every night.
- Both PATH writes used `$PWD` from a `shell: bash` step. Under git-bash on
  Windows that is `/d/a/dora/dora`, and GITHUB_PATH entries are taken verbatim,
  so the nine steps in `cli-tests-python` that invoke `dora` without an explicit
  `shell:` (they default to pwsh) would have resolved it drive-relative.
  `${{ github.workspace }}` is Windows-form on Windows.
- The artifact upload sat at the end of `cli-tests`, so any later native failure
  — including the flaky Windows leg this change is about — would have skipped
  the Python half on *every* platform.

This one can't be run locally, so verification is a YAML parse plus `bash -n` on
all 130 embedded run-blocks, and an audit of which steps invoke `dora` under
which shell.

Refs #2742


phil-opp:

The split itself preserves coverage on a green run (every step lands in exactly one half; C/C++ moved into `cli-tests`, Python steps into `cli-tests-python`; both aggregator `needs:` lists and the local-repro maps are updated). One issue on the failure path:

**`cli-tests-python` is silently skipped on all platforms whenever any `cli-tests` leg fails.** The job has `needs: cli-tests` and no job-level `if:`, so it defaults to `if: success()`. Because `cli-tests` is a `fail-fast: false` matrix, if any single leg (Windows/macOS/Linux) ends in `failure`, `success()` on the dependent job is false and `cli-tests-python` is **skipped on every platform**. That directly contradicts the `Publish CLI for the Python half` step's own comment — *"a later native test failing here must not skip the Python half on every platform"*: publishing the artifact early makes it *available*, but availability is moot when the downstream job is gated out by `needs` + `success()`.

The consequence is exactly the regression this PR is meant to avoid: on any nightly where the (now native) `cli-tests` half fails — including the flaky Windows leg that motivated #2742 — the whole Python CLI suite doesn't run, and a `skipped` job is invisible to `file-issue-on-failure` (which lists only `failure`), so the nightly issue won't even name it. To get the isolation the comment describes, gate the job on `if: ${{ !cancelled() }}` instead of the implicit `success()` (the early artifact upload means the binary is present whenever the build step itself succeeded).

Minor: the CLAUDE.md edit bumps the job count to **23**, but the surrounding prose is left stale — "the **17 remaining**" still sums to 21 with the 4 example-backed jobs, and the trailing "The **22nd**, `memory-pool-smoke`…" sentence no longer lines up.

---

phil-opp:

The new commit (`2a0188718`) fixes both issues I raised earlier: `cli-tests-python` is now gated on `if: ${{ !cancelled() }}` instead of the implicit `success()`, so a single red `cli-tests` leg no longer skips the Python half on every platform, and the CLAUDE.md breakdown now sums correctly (4 example-backed + 18 driven by `ci-nightly-jobs.sh` + `memory-pool-smoke` = 23). No new issues.

---

PR #2988, opened 2026-08-03
feat(cli): atomic `dora node replace` for live node swaps

### LINKED ISSUE #2927: feat(cli): atomic `dora node replace` for live node swaps
Split out from #2916, where it was the third ask. The bug reported there is fixed (`eec31a40b`); this is the workflow request that outlives it.

## What

```
dora node replace --dataflow <df> <id> --from-yaml <node.yml>
```

An atomic swap of a running node for a new definition under the same id, with no observable gap and no identity confusion between the outgoing and incoming incarnations.

## Why

Live iteration on a running dataflow is the actual workflow. Today it's spelled `dora node remove` + `dora node add`, which is two round trips through CLI → coordinator → daemon with a window in between where:

- the id exists in no `running_nodes` entry, so lifecycle bookkeeping for the outgoing incarnation has nowhere consistent to land (this is what #2916 was, and what #2926 tracks the remainder of);
- downstream consumers see the output close and reopen rather than a swap;
- a scripted caller has to decide how long to sleep between the two commands. The report in #2916 shipped a ~2s settle for exactly this reason.

An atomic primitive removes the window rather than making callers time around it.

## Sketch

A single `ReplaceNode { dataflow_id, node_id, node, uv }` coordinator → daemon request that, in one pass of the daemon event loop:

1. spawns the replacement first and only commits if the spawn succeeds (matching `AddNode`'s existing deferred-state-mutation shape at `binaries/daemon/src/lib.rs`, where routing state is only touched after a successful spawn);
2. swaps the `running_nodes` entry and re-points routing/mappings/timers in the same step;
3. stops the outgoing incarnation, with its later exit attributed to the incarnation that produced it rather than to the id.

Step 3 is the part that needs the incarnation identity discussed in #2926 — worth landing that first so `replace` isn't built on the same id-keyed accounting.

## Open questions

- Should inbound messages queue during the swap, or is a drop acceptable? Queueing is what makes it a true hot swap; dropping is far simpler.
- Does `replace` require the new descriptor to have compatible inputs/outputs, or may it change the node's edges? Changing edges makes it a topology edit, not a swap.
- Behaviour when the replacement fails to spawn: keep the old incarnation running (safest), or fail the dataflow?

heyong4725:
## Summary

Implements the atomic node swap from #2927: `dora node replace <id> --from-yaml <file> [--grace <dur>]` — one command instead of `remove` + `add`, no caller-side sleep, no window where the id is unaccounted for.

**Depends on #2978** (node lifecycle generations, #2926) — this branch stacks on it and stays a draft until it lands. The outgoing incarnation's exit is dropped by the generation guard instead of being attributed to its replacement, which is exactly the property #2926 was landed to provide.

## Design decisions (the three open questions in #2927)

- **Drop, not queue**, inputs during the swap window — the simpler semantics the issue offered; `pending_messages` is a counter, not a queue, so true queueing would be new machinery.
- **Same edges required**: identical input mappings, and declared outputs must cover every output downstream consumers are mapped to. Edge changes are rejected with a pointer at `remove`/`add` and the mapping commands — a replace is a swap, not a topology edit.
- **Spawn failure keeps the old incarnation running**: the daemon prepares and spawns the replacement before touching any state.

## Implementation

- `ControlRequest::ReplaceNode` → coordinator routes to the daemon **owning** the id (`node_to_daemon`, unlike AddNode's first-daemon placement), commits its descriptor swap only on the specific `ReplaceNodeResult` reply (#1682 contract)
- Daemon handler does the swap in one event-loop pass: validate edges → spawn replacement → stop outgoing incarnation (graceful `Stop` while its subscribe channel is still installed, then the grace-kill schedule) → reset per-incarnation state, keep edge registrations (validated identical) → register replacement → `clear_node_result`
- Startup guard is `PendingNodes::is_pending` (the barrier itself), **not** `dataflow_started` — that flag lags the all-daemons-ready roundtrip by seconds under load; the first e2e run caught the coarse guard spuriously rejecting replaces on a visibly running dataflow

## Known limitation

The brief overlap window shares the pre-existing generation-blind `Event::Node` hazard (old connection's stream-drop racing the successor's subscribe) flagged in the #2978 review — in practice the outgoing socket closes at kill time, well before the replacement finishes exec+connect. Generation-stamping `Event::Node` is the tracked follow-up.

## Tests

- CLI parse tests (`parse_node_replace`, `reject_node_replace_without_yaml`)
- e2e `replace_swaps_running_node_without_blaming_successor`: outgoing incarnation exits 1 on stop; the successor must be Running with a new pid and `dora stop` must report a clean dataflow
- e2e `replace_failure_keeps_current_incarnation_running`: bad-path spawn failure and edge-change rejection both error and leave the original pid Running

## Validation

- `cargo clippy -D warnings` on dora-cli, dora-daemon, dora-coordinator, dora-message + the e2e target
- unwrap budget 159/163
- unit suites: 291 cli / 107+140 message+coordinator / 168 daemon — all green
- both new e2e tests green locally (isolated HOME)
- `cargo check --examples`

Closes #2927


phil-opp:

The single-daemon path looks solid and is well covered by the new e2e tests (`replace_swaps_running_node_without_blaming_successor`, `replace_failure_keeps_current_incarnation_running`): spawn-first ordering leaves the current incarnation running on any build/spawn failure, and the outgoing exit is dropped by the generation guard so `dora stop` stays clean. One correctness gap worth addressing:

**Output-coverage validation ignores remote consumers (multi-daemon).** The same-edges check validates dropped outputs only against the local `mappings`:

```rust
let new_outputs = node.kind.run_config().outputs;
for output_id in dataflow.mappings.keys().filter(|oid| oid.0 == node_id) {
    eyre::ensure!(new_outputs.contains(&output_id.1), ...);
}
```

`RunningDataflow` tracks remote-only consumers separately in `open_external_mappings`, which this loop never consults. In a multi-daemon dataflow where node `A` (daemon-1) produces output `X` consumed only by node `B` (daemon-2), a `dora node replace A` whose YAML drops `X` passes validation, succeeds, and silently breaks `B`'s input — violating the advertised atomic same-edges guarantee. The codebase already has `RunningDataflow::node_output_ids(node_id)` (`running_dataflow.rs:977`), which unions `mappings` and `open_external_mappings`; using it here instead of `mappings.keys()` would close the gap. (The single-daemon path the e2e tests exercise is fully validated — this is scoped to distributed dataflows.)

Minor / lower-confidence: in the `replace_process_handle` teardown-race path, the entry's generation is advanced to `new_generation` while the `grace_duration_kills` markers were inserted under the old generation, so the replacement's SIGKILL exit could be classified as unclean rather than a planned stop. It's a narrow window (`disable_restart` flipping after the loop's pre-send check) and belongs to the generation work rather than the replace feature itself — noting for completeness.

---

heyong4725:
## Round-2 review applied (`e27b8ffa1`)

A 7-pass review (2 specialists, Claude adversarial, Codex adversarial + structured, Red Team) found the v1 swap deterministically corrupted downstream state. All approved fixes are in:

**Generation-attributed node events** — per-spawn listeners stamp each `Event::Node` with the connection's incarnation (snapshotted at register time; the restart loop publishes the successor's generation before spawning). The daemon drops events from strictly older incarnations, so the outgoing process's graceful teardown can no longer close the replacement's consumers or remove its subscription. Also closes the same-shape pre-existing holes for `dora node restart` and re-added ids.

**Validator/state fixes** — output coverage includes `open_external_mappings`; input edges validated against live topology (fixes operator-node/provenance mismatch); `open_inputs` no longer reset (fixes the replaced-sink-after-finished-upstream hang); stale cascading causes cleared; v1 scope restricted to spawned custom nodes with actionable rejections.

**Tests** — consumer-survival e2e (sole-input sink outlives its producer's replacement: red/green for the stamp), edge/output rejection matrix + identical-edges success with a live timer input, replace-the-replacement, `ensure_replace_node_applied` trio, `is_pending` unit test.

**Pre-existing issues surfaced by the review, out of scope here (tracked for follow-up):** spawn_inner failure window can leak an unregistered child (shared with AddNode); coordinator 30s reply timeout can produce applied-but-reported-failed divergence for slow builds (protocol-level, shared with AddNode); per-incarnation TCP listeners live until dataflow end (shared with every spawn); "spawned ≠ subscribed" commit semantics (a replacement that crashes immediately still stops the old node — its restart policy then owns recovery).

All local gates green: fmt, clippy -D warnings across 4 crates + e2e target, unwrap budget 159/163, 711 unit tests, 4/4 replace e2e (isolated HOME).

heyong4725:
Review findings for this replacement flow:

1. **P1 — Preserve successor subscription events during respawn** (`binaries/daemon/src/lib.rs:1638-1643`)

   A restarted process can connect before `ProcessHandleReplaced` updates `running_nodes`. Its listener emits events with the successor generation, but the daemon still compares against the predecessor generation and drops the successor's one-time `Subscribe` event. The restarted incarnation can then remain disconnected from daemon events indefinitely. Accept the pending successor generation or install it before the child can connect.

2. **P2 — Pass the replacement descriptor to the new process** (`binaries/daemon/src/lib.rs:2835-2839`)

   `DoraNode::dataflow_descriptor()` is cloned and serialized into `DORA_NODE_CONFIG` before the target entry is updated, so the replacement receives the outgoing node's path, environment, restart policy, and other definition fields. Construct the descriptor with the replacement definition before creating the `Spawner`.

3. **P2 — Shut down the outgoing incarnation's listener** (`binaries/daemon/src/lib.rs:2857-2868`)

   Each successful replacement binds a fresh TCP listener, while the outgoing listener remains alive until the entire dataflow shuts down. Repeated live replacements therefore accumulate listener tasks and socket file descriptors. Retain a per-incarnation listener shutdown handle and close it when committing the swap.

heyong4725:
Fixed in `5c0c362dc`. All three findings confirmed and fixed in the follow-up commit.

**1 (P1) — confirmed, and a sharp catch.** The gate was written as an exact-generation match while the round-2 design notes (and the counter-published-before-spawn ordering) required strictly-older. Your scenario is exactly the hole: successor connects → stamped `G+1` → entry still `G` → its one-shot `Subscribe` dropped → incarnation permanently deaf. Fixed by extracting the predicate (`event_generation_is_stale`: `event < entry`) and gating on that; `node_event_gate_drops_only_strictly_older_generations` pins all three orderings (older dropped, equal passes, **newer passes**).

**2 (P2) — confirmed.** Added `patch_descriptor_entry` (path/args/env/edges/restart config/log routing from the resolved replacement) and applied it in both places: the spawn-time descriptor clone before the `Spawner` is built (so `DORA_NODE_CONFIG` describes the replacement) and the committed entry (so later spawns/respawns of other nodes serialize it too).

**3 (P2) — confirmed.** Listeners now have per-node lifetime: `spawn_node` creates a `watch` channel, the accept loop selects on it alongside the dataflow-wide shutdown, and the sender lives in the `RunningNode` entry plus the restart loop. When the node is replaced/removed/torn down and its restart loop exits, the last sender drops and the listener stops accepting — no accumulation across repeated swaps. Existing connections are unaffected (independent tasks, generation-gated).

Validation: fmt/clippy `-D warnings` clean, unwrap budget 159/163, 171 daemon unit tests (incl. the new gate regression), 4/4 replace e2e green (isolated HOME).

PR #2989, opened 2026-08-03
fix(test): identify the fixture by `args`, not `comm` — comm truncates on Linux

heyong4725:
**Main's `E2E Tests` job is red and has been since #2961 merged.** This fixes it.

## What is failing

`run_killed_by_sigterm_terminates_nodes_and_exits` fails at its own precondition on every Linux CI run:

```
panicked at tests/node-lifecycle-e2e.rs:
  precondition: node <pid> should be running our fixture before we signal
```

## Why, and why it was not caught

The identity check came out of review on #2961 — a real pid-reuse hazard, where a recycled pid could be mistaken for a surviving node and get SIGKILLed. The guard was right; the implementation was not portable:

- on **Linux**, `ps -o comm=` reads `/proc/<pid>/comm`, which the kernel caps at `TASK_COMM_LEN - 1` = **15** characters
- `sigterm-ignoring-node` is **21**, so Linux sees `sigterm-ignorin`
- `.contains("sigterm-ignoring-node")` is therefore always false

On macOS `comm` returns the full path, so it matched. Every local run passed, including mutation checks — those proved the test *discriminates*, which says nothing about whether it *ports*. The test was written, reviewed, mutation-verified and merged without ever running on the platform CI uses.

## The fix

`ps -ww -o args=` carries the full command line on both platforms. Both identity checks now route through a single `ps_args` helper so they cannot drift apart — the same consolidation lesson that produced the bug they guard (a correct check added in one place, missed in its sibling).

The pid-reuse guard is unaffected: it still confirms the process is ours before asserting on it or signalling its group.

## Verification

Passes on macOS (607s, no leftover processes). **Linux CI on this PR is the real check** — that is the platform the bug is specific to, and the one I could not exercise locally.

Refs #2961

### DIFF
--- a/tests/node-lifecycle-e2e.rs
+++ b/tests/node-lifecycle-e2e.rs
@@ -1899,14 +1899,38 @@ fn run_killed_by_sigterm_terminates_nodes_and_exits() {
 /// mistaken for a surviving node.
 #[cfg(unix)]
 fn node_is_fixture(pid: u32) -> bool {
-    if !process_alive(&pid.to_string()) {
-        return false;
-    }
-    let out = Command::new("ps")
-        .args(["-o", "comm=", "-p", &pid.to_string()])
+    // `expect`, not a silent false: this feeds the assertions, and the
+    // final one is `!node_is_fixture(..)`. If a failure to RUN `ps` were
+    // folded into "not our process", an unavailable `ps` would satisfy
+    // that assertion and the test would report success having verified
+    // nothing.
+    let args = ps_args(pid).expect("failed to run `ps -o args=`");
+    process_alive(&pid.to_string()) && args.contains(FIXTURE_BIN)
+}
+
+/// Matched with a leading `/` so it hits the spawned binary's path and not
+/// a process that merely *names* it in its arguments — `cargo build -p
+/// sigterm-ignoring-node`, which this test itself runs.
+const FIXTURE_BIN: &str = "/sigterm-ignoring-node";
+
+/// The full command line of `pid`. `None` if `ps` could not be run at all;
+/// `Some("")` if it ran and the pid is gone.
+///
+/// `args=`, NOT `comm=`: on Linux `comm` is `/proc/<pid>/comm`, which the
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

The switch from `ps -o comm=` to `ps -ww -o args=` is correct: `comm` is capped at `TASK_COMM_LEN-1` (15 chars) on Linux via `/proc/<pid>/comm`, so the 21-char `sigterm-ignoring-node` never matched there, while `args=` carries the full command line on both Linux and macOS. Routing both `node_is_fixture` and `still_our_fixture` through the single `ps_args` helper removes the drift risk between the two checks.

On test validity: the change preserves the pid-reuse guard the check exists for. `node_is_fixture` still requires `process_alive(pid)` before the name match (the `&&` short-circuits), and `ps_args` returning empty on a failed/absent `ps` invocation means `.contains(FIXTURE_BIN)` is `false` — so `still_our_fixture` stays conservative (it will decline to signal rather than risk hitting the wrong target), matching the old `out.status.success()` gate's intent. The precondition assertion in `run_killed_by_sigterm_terminates_nodes_and_exits` continues to exercise real behavior rather than passing vacuously.

---

heyong4725 (commented):
No findings on `af45db942`.

The change fixes the Linux-specific `comm` truncation without weakening the test intent: `ps -ww -o args=` preserves the full absolute fixture path, the leading-slash match avoids the concurrent Cargo package-name argument, the live precondition prevents an unsupported `ps` format from making the final negative assertion vacuously pass, and the Drop path remains conservative on inspection failure.

Validation:
- `git diff --check` passed.
- `cargo test --test node-lifecycle-e2e run_killed_by_sigterm_terminates_nodes_and_exits -- --exact --nocapture --test-threads=1` passed: 1 passed in 255.63s.
- No `sigterm-ignoring-node` process remained afterward.
- CI format/check/clippy and static gates are green.

Residual limitation: this host run is macOS, while the regression is Linux-specific, and the PR E2E job is currently skipped. The merge-queue Linux E2E remains the definitive portability verification.

PR #2991, opened 2026-08-03
fix(daemon): stop descriptor `env:` from overriding the daemon's control-plane variables

### LINKED ISSUE #2944: daemon: node `env:` can override the daemon's own control-plane variables (DORA_ZENOH_*, DORA_RUNTIME_CONFIG)

heyong4725:
Closes #2944.

## Problem

In both spawn paths the daemon injected `DORA_NODE_CONFIG` / `DORA_RUNTIME_CONFIG` and the `DORA_ZENOH_*` wiring **before** applying descriptor `env:` entries, and `is_denied_env` did not cover those keys — so a later `.env()` from the descriptor won. A forged `DORA_ZENOH_CONNECT` cuts a node off from its peers: the dataflow starts cleanly and silently exchanges nothing. `DORA_NODE_CONFIG` was safe only because `spawn_inner` re-serializes it after command build (a restart-count fix, not a security measure).

## Fix

Both defenses suggested in the issue, so the protection is layered and intentional:

1. **Deny with a warning** — new `CONTROL_PLANE_ENV` list (`DORA_NODE_CONFIG`, `DORA_RUNTIME_CONFIG`, `DORA_ZENOH_LISTEN`, `DORA_ZENOH_CONNECT`, `DORA_ZENOH_MULTICAST`) checked by `is_denied_env` with a wiring-specific message, turning a silent misconfiguration into a visible one. Matching is now **case-insensitive** for both lists: Windows env vars are case-insensitive (and last-insert-wins in the std `Command` env map), so `dora_zenoh_connect:` would otherwise bypass both defenses there.
2. **Inject last** — both spawn paths now apply descriptor env first (`apply_descriptor_env`, which also dedups the three copy-pasted env loops) and inject the daemon's variables afterwards (`inject_control_plane_env`), so the daemon's wiring wins by construction even for a control-plane key the list misses.

`strip_denied_env` intentionally still strips only the original `ENV_DENYLIST`: ambient `DORA_ZENOH_*` inheritance from the daemon's environment (e.g. an externally configured router) is untouched.

## Tests

- `reserved_control_plane_keys_are_denied` — all five keys plus case variants are denied; ordinary vars pass.
- `descriptor_env_cannot_override_the_daemons_wiring` — pins both defenses independently: forged descriptor entries are dropped, and a pre-seeded forged `DORA_NODE_CONFIG` (bypassing the denylist) is still overwritten by inject-last ordering; a wiring variable the daemon did not inject does not appear at all.

## Validation

- `make qa-fast` (fmt, clippy `-D warnings`, audit, unwrap budget, typos) clean
- `cargo test -p dora-daemon --lib`: 173 passed
- full workspace tests + `cargo check --examples` + `replace_swaps_running_node_without_blaming_successor` e2e (real spawn through the modified path) green

## Follow-up candidate (not this PR)

The deny warning surfaces in the daemon log at spawn time; descriptor validation (`libraries/core/src/descriptor/validate.rs`) is env-blind, so `dora build`/`dora start` give no CLI-visible feedback for a reserved key in `env:`. A validation-time warning would surface it where the author is looking — same follow-up shape as the existing `ENV_DENYLIST` behavior.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1413,6 +1413,14 @@ impl Daemon {
         zenoh_bind: ZenohBind,
         disable_multicast: bool,
     ) -> eyre::Result<(Self, mpsc::Receiver<Timestamped<Event>>)> {
+        // Fold in `DORA_ZENOH_MULTICAST` so this is the daemon's *effective*
+        // decision, not just its flag. The zenoh session honors the variable on
+        // its own, but the spawner forwards this field to nodes — and nodes no
+        // longer inherit the variable — so a daemon started with the variable
+        // instead of `--zenoh-no-multicast` would otherwise leave its nodes
+        // scouting by multicast (#2991 review).
+        let disable_multicast = dora_core::topics::multicast_disabled(disable_multicast);
+
         // Reserve a port and have zenoh listen on it. The endpoint is injected
         // into spawned nodes via `DORA_ZENOH_CONNECT` so peer discovery works
         // without multicast (#1778).
--- a/binaries/daemon/src/spawn/spawner.rs
+++ b/binaries/daemon/src/spawn/spawner.rs
@@ -45,16 +45,64 @@ const ENV_DENYLIST: &[&str] = &[
     "DORA_ALLOW_SHELL_NODES",
 ];
 
+/// Control-plane variables the daemon injects into every node it spawns.
+/// Descriptor `env:` / `envs:` entries must not override them: they configure
+/// how the node reaches its daemon and peers, and a forged value produces a
+/// dataflow that starts cleanly but silently exchanges nothing (#2944).
+///
+/// `DORA_NODE_CONFIG` was previously safe only because `spawn_inner`
+/// re-serializes it after the command is built (a restart-count fix, not a
+/// security measure); listing it here makes that safety intentional.
+const CONTROL_PLANE_ENV: &[&str] = &[
+    "DORA_NODE_CONFIG",
+    "DORA_RUNTIME_CONFIG",
+    DORA_ZENOH_LISTEN_ENV,
+    DORA_ZENOH_CONNECT_ENV,
+    DORA_ZENOH_MULTICAST_ENV,
+];
+
 /// Returns true if the env var key is denied, logging a warning if so.
+///
+/// Matched case-insensitively: Windows environment variables are
+/// case-insensitive (and last-insert-wins in the std `Command` env map), so
+/// a `dora_zenoh_connect` descriptor entry would otherwise bypass both the
+/// denylist and the inject-last ordering there.
 fn is_denied_env(key: &str) -> bool {
-    if ENV_DENYLIST.contains(&key) {
+    if ENV_DENYLIST.iter().any(|d| key.eq_ignore_ascii_case(d)) {
         tracing::warn!(
             "skipping denied environment variable '{key}' (security: could inject shared libraries)"
+        );
+        true
+    } else if CONTROL_PLANE_ENV
+        .iter()
+        .any(|d| key.eq_ignore_ascii_case(d))
+    {
+        tracing::warn!(
+            "ignoring `{key}` from the descriptor env: it is control-plane wiring \
+             set by the daemon and cannot be overridden"
         );
         true
     } else {
         false
     }
+}
+
+/// Apply descriptor-provided env (`env:` / `envs:`) to the command, dropping
+/// denied keys with a warning. Must run BEFORE the daemon injects its
+/// control-plane variables so the daemon's wiring wins by construction even
+/// for a key the denylist misses (#2944).
+fn apply_descriptor_env(
+    mut command: Command,
+    envs: Option<&BTreeMap<String, EnvValue>>,
+) -> Command {
+    if let Some(envs) = envs {
+        for (key, value) in envs {
+            if !is_denied_env(key) {
+                command = command.env(key, value.to_string());
+            }
+        }
+    }
+    command
 }
 
 /// Strip all denied env vars from the inherited process environment.
@@ -305,6 +353,23 @@ impl Spawner {
         }
     }
 
+    /// Inject the daemon's control-plane variables (`config_key` +
+    /// `DORA_ZENOH_*`). Must run AFTER [`apply_descriptor_env`] — a later
+    /// `.env()` overwrites an earlier one — so the daemon's wiring cannot be
+    /// overridden from the descriptor (#2944). The denylist in
+    /// [`is_denied_env`] additionally turns an override attempt into a
+    /// visible warning instead of a silent no-op.
+    fn inject_control_plane_env(
+        &self,
+        command: Command,
+        node_id: &NodeId,
+        config_key: &str,
+        config_yaml: String,
+    ) -> Command {
+        let command = command.env(config_key, config_yaml);
+        self.maybe_inject_zenoh_connect(command, node_id)
+    }
+
     #[allow(clippy::too_many_arguments)]
     pub async fn spawn_node(
         self,
@@ -424,29 +489,16 @@ impl Spawner {
                     command = command.stdin(Stdio::Null);
                     command = strip_denied_env(command);
 
-                    command = command.env(
+                    // Descriptor env first, daemon control-plane wiring last (#2944).
+                    command = apply_descriptor_env(command, node.env.as_ref());
+                    command = apply_descriptor_env(command, n.envs.as_ref());
+                    command = self.inject_control_plane_env(
+                        command,
+                        &node.id,
                         "DORA_NODE_CONFIG",
-                        serde_yaml::to_string(&node_config.clone())
+                        serde_yaml::to_string(&node_config)
                             .wrap_err("failed to serialize node config")?,
                     );
-                    command = self.maybe_inject_zenoh_connect(command, &node.id);
-                    // Injecting the env variable defined in the `yaml` into
-                    // the node runtime.
-                    if let Some(envs) = &node.env {
-                        for (key, value) in envs {
-                            if !is_denied_env(key) {
-                                command = command.env(key, value.to_string());
-                            }
-                        }
-                    }
-                    if let Some(envs) = &n.envs {
-                        // node has some inner env variables -> add them too
-                        for (key, value) in envs {
-                            if !is_denied_env(key) {
-                                command = command.env(key, value.to_string());
-                            }
-                        }
-                    }
 
                     // For managed Python custom nodes, also set VIRTUAL_ENV and
                     // prepend the env's bin dir to PATH so subprocesses, console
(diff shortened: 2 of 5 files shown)


heyong4725:
Multi-model review (2 specialists + adversarial Claude + adversarial Codex) found the first commit incomplete and a neighboring function inert. Both verified against the dependency source and a runtime probe, both fixed in `1b032db48`.

**The fix had a hole: `ZENOH_CONFIG`.** `open_zenoh_session_with_listen` (`libraries/core/src/topics.rs:157`) branches on zenoh's own config-file variable **first** and builds the entire session from that file, never reading `DORA_ZENOH_CONNECT` / `LISTEN` / `MULTICAST`. So `env: { ZENOH_CONFIG: ./mine.json5 }` reproduced #2944's silent-disconnection symptom without touching a single reserved key — and could point a node's session at an arbitrary remote router. Now refused from descriptor `env:`, still inherited from the daemon's environment (the documented deployment-wide override). Exposed as `dora_core::topics::ZENOH_CONFIG_PATH_ENV`, aliased to zenoh's own const so the two cannot drift.

**`strip_denied_env` was a no-op.** `clonable_command::Command::env_remove` calls `self.environment.remove(key)`, but `impl From<&Command> for StdCommand` only emits `env_remove` for entries whose value is `None` — a deleted entry produces nothing, and `inherit_environment` stays true. Verified in the crate source (`clonable-command-0.2.0/src/lib.rs:232,259`). Every spawned node was inheriting the daemon's `DORA_AUTH_TOKEN` (readable from `/proc/<pid>/environ`) and any ambient loader variables, which is exactly what the function's doc comment claimed to prevent. It now inserts the `None` sentinel directly, and covers the control-plane keys too, so a stale `DORA_ZENOH_CONNECT` in the daemon's shell can't wire a node the daemon deliberately left unwired.

**Malformed key names bypassed the denylist.** A key containing `=` goes to `execve` verbatim. Probed on macOS: `.env("LD_PRELOAD=/tmp/evil.so ", "")` lands in the child as `LD_PRELOAD=/tmp/evil.so =`, i.e. a real `LD_PRELOAD` whose value the loader splits on whitespace — while `is_denied_env` saw an unfamiliar name and let it through. Empty, whitespace-bearing, and NUL-bearing names are now rejected too.

**Case folding is Windows-only now.** Codex was right that folding on Unix is a regression: `ld_preload` there is a different, harmless variable, and the first commit silently dropped it. Windows now uses the OS uppercase mapping instead of an ASCII fold, so Unicode lookalikes collapse the same way the platform collapses them.

**Tests pin the ordering for real.** The previous regression test performed the correct order inside the test body, so reverting either call site would have kept it green. Both spawn paths now compose through one `compose_node_env` helper that the tests drive directly, which also gives the operator-runtime path (`DORA_RUNTIME_CONFIG`, the arm with no accidental backstop) its first coverage.

### Behavior change worth calling out

Making the scrub real means the loader denylist finally takes effect. A node that relied on inheriting `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH` from the daemon's environment must now set it in its own `env:`. Nothing in this repo does (checked examples, scripts, CI); external users could be affected. Reserved names are now documented in `docs/yaml-spec.md`.

### Validation

`make qa-fast` clean, 176 daemon unit tests, full workspace tests, `cargo check --examples`, plus real spawns: `smoke_rust_dataflow{,_dynamic,_url}` (3/3) and the `replace_swaps` node-lifecycle e2e.

### Not fixed here

- Descriptor env **values** are re-expanded against the daemon's environment (`with_expand_envs`), so `env: { X: "$${DORA_AUTH_TOKEN}" }` can exfiltrate the token under an innocuous key. The CLI's `--env` already rejects `$` for this reason; the YAML path does not. Separate issue — different mechanism, different fix.
- Descriptor validation is env-blind, so the deny warnings only reach the daemon log, not `dora build` / `dora start`. The hub *manifest* validator already hard-rejects `DORA_*`; mirroring that in `descriptor/validate.rs` would surface this where the author is looking.

heyong4725:
Review findings for the inherited-environment scrub:

1. **P1 — Preserve runtime library search paths for spawned nodes** (`binaries/daemon/src/spawn/spawner.rs:155-158`)

   When the daemon starts in an environment that supplies `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH` (for example, sourced ROS/CUDA or custom-library setups), the new scrub removes those variables from every child. They cannot be restored through node `env:` because `is_denied_env` drops them there too, so dynamically linked node executables can fail before starting. Retain inherited search paths or allow explicitly configured values after scrubbing.

2. **P2 — Forward ambient multicast disablement to child nodes** (`binaries/daemon/src/spawn/spawner.rs:62-64`)

   If the daemon is launched with `DORA_ZENOH_MULTICAST=off` instead of `--zenoh-no-multicast`, `open_zenoh_session_with_listen` honors the environment variable for the daemon but `self.disable_multicast` remains false. Because this key is now scrubbed, spawned nodes neither inherit the effective setting nor receive a replacement; they still use multicast scouting and can hit the bind failures the setting was meant to prevent. Propagate the daemon's effective multicast mode before removing the inherited value.

heyong4725:
Both confirmed and fixed in `0ea9fc96c`. These are regressions the scrub introduced, not pre-existing gaps — thanks for catching them before this landed.

**1 (P1) — library search paths.** Right, and it was a dead end in both directions: scrubbed from what a node inherits *and* refused from `env:`, so a daemon started from a sourced ROS or CUDA environment left a dynamically linked node with no way to find its libraries, failing in the loader before the node can report anything. I went with "retain inherited search paths" of your two options, because that is exactly the pre-PR behavior (`strip_denied_env` was inert, so they were always inherited, and `env:` always refused them) — letting `env:` set them would loosen a denylist entry that predates this PR.

The list is now split by what the variable is *for*:

- `ENV_DENYLIST` — loader-injection vectors and daemon secrets (`LD_PRELOAD`, `LD_AUDIT`, `DYLD_INSERT_LIBRARIES`, `DORA_AUTH_TOKEN`, `DORA_ALLOW_SHELL_NODES`): scrubbed from inheritance and refused from a descriptor.
- `SEARCH_PATH_ENV` — `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`: refused from a descriptor, inherited. Its own warning tells the author where the value is expected to come from.

`library_search_paths_survive_the_scrub` pins both halves. This also retires the behavior-change caveat from my earlier comment: nothing that worked before this PR stops working now.

**2 (P2) — multicast.** Confirmed at `libraries/core/src/topics.rs:357`: the session computes `request || DORA_ZENOH_MULTICAST`, but `Daemon::disable_multicast` carried only the `--zenoh-no-multicast` flag and that is what the spawner forwards. Nodes used to inherit the variable directly, so setting it instead of the flag worked by accident until the scrub removed that path — and it failed in precisely the environments the setting exists for, where the scouting bind is the problem.

Rather than special-case the spawner, the fold now lives in one place: `topics::multicast_disabled(requested_off)` is used both for the session's own decision and by `build_daemon` to normalize its field, so what the daemon does and what it forwards cannot disagree. `an_explicit_multicast_disable_is_never_lost` pins the direction that is testable without mutating the process environment (the value-parsing half is already covered by the existing tests).

**Validation:** `make qa-fast` clean, 177 daemon + 317 core unit tests, full workspace tests, `cargo check --examples`, and `smoke_rust_dataflow{,_dynamic,_url}` 3/3 — real spawns through the modified path, from a clean target dir.

PR #2993, opened 2026-08-03
fix(cli): remove spurious double spaces in default `pretty` log output

phil-opp:
## Issue

The `pretty` log formatter (`binaries/cli/src/output.rs`) built its final line as:

```rust
"{time} {level_str} {dataflow} {node}{target} {message}"
```

But the `dataflow` and `target` fields already carry their **own** trailing space when present, and are empty strings when absent. The literal spaces placed *around* them therefore doubled up whenever those fields were empty — which is the default for essentially every line (`print_dataflow_id` is off by default and most messages have no `target`).

The result was `INFO    node:  message` instead of `INFO   node: message` on ordinary `dora logs` / `dora run` / `dora start` output.

## Fix

Rely on each field's own trailing space, so present values stay single-spaced and absent ones contribute nothing:

```rust
"{time} {level_str} {dataflow}{node}{target}{message}"
```

To make the exact column spacing testable without capturing stdout, the line construction is extracted into `format_pretty_line` (returns the `String`); `print_pretty` keeps the lifecycle blank-line handling and the `println!`.

## Validation

- Three new unit tests pin the default, with-target, and system (`[dora]`) cases and assert no double space appears (`cargo test -p dora-cli --lib output::` → 20 passed).

All pass. Also verified in a combined full-workspace `cargo test` run.

---

⚠️ **This pull request was generated autonomously by Claude (an AI agent).** It is machine-generated; please review carefully before merging.

https://claude.ai/code/session_01Ltyvv3SpFoRwR3ckEDUpjQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ltyvv3SpFoRwR3ckEDUpjQ)_

### DIFF
--- a/binaries/cli/src/output.rs
+++ b/binaries/cli/src/output.rs
@@ -60,6 +60,23 @@ pub fn print_log_message(log_message: LogMessage, config: &LogOutputConfig) {
 }
 
 fn print_pretty(log_message: LogMessage, config: &LogOutputConfig) {
+    let is_system = log_message.node_id.is_none();
+    let is_lifecycle = is_system && is_lifecycle_message(&log_message.message);
+    let line = format_pretty_line(&log_message, config);
+
+    if is_lifecycle {
+        println!();
+    }
+    println!("{line}");
+    if is_lifecycle {
+        println!();
+    }
+}
+
+/// Render a single log message as a colored one-line string for the `pretty`
+/// format. Kept separate from [`print_pretty`] so the exact column spacing is
+/// unit-testable without capturing stdout.
+fn format_pretty_line(log_message: &LogMessage, config: &LogOutputConfig) -> String {
     let LogMessage {
         build_id: _,
         dataflow_id,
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found.

The double-space fix is correct: the `dataflow` and `target` fields already carry their own trailing space when present and are empty otherwise, so the old `"{dataflow} {node}{target} {message}"` produced a stray double space on the common path where those fields are empty. The new `"{dataflow}{node}{target}{message}"` relies on each field's own trailing space, keeping present values single-spaced and absent ones contributing nothing. Extracting `format_pretty_line` (returning `String`) while `print_pretty` keeps the lifecycle blank-line handling makes the spacing unit-testable without capturing stdout, and the three tests assert the default, with-target, and system (`[dora]`) cases with explicit no-double-space checks — they would fail if the separator logic regressed.

---

---

phil-opp:

The fix is correct: `dataflow`/`node`/`target` each already carry their own trailing space when present and are empty otherwise, so the previous literal spaces around them produced a double space on the default (empty-field) path. The new format string single-spaces every case (default, with-target, with-dataflow, and `[dora]` system lines), lifecycle blank-line handling is preserved, and the extracted `format_pretty_line` makes the spacing unit-testable. The three tests assert exact rendered strings and would fail under the old format.

---

---

PR #2994, opened 2026-08-03
fix(mavlink2-bridge): reject an unrecognized `?proto=` instead of defaulting to V2

phil-opp:
## Issue

In `libraries/extensions/mavlink2-bridge/src/transport.rs`, `parse_proto_query` returned `Option<MavlinkVersion>` and the callers did:

```rust
let version = parse_proto_query(url).unwrap_or(MavlinkVersion::V2);
```

That collapsed two different situations into the same V2 default: a genuinely **absent** `?proto=` parameter, and a **present-but-unrecognized** one such as `?proto=v3` or a typo. A misconfigured endpoint therefore connected on the wrong protocol with no diagnostic.

## Fix

Make `parse_proto_query` return `BridgeResult<MavlinkVersion>`:

- absent → the `MavlinkVersion::V2` default, quietly;
- `v1`/`1`/`1.0` or `v2`/`2`/`2.0` (case-insensitive) → that version;
- anything else → a `BridgeError::Config` naming the offending token and the supported set (`v1`, `v2`).

`connect_tcp` / `connect_udp` propagate it with `?`. `parse_proto_query` is private, so this is not an API break.

## Validation

- New unit tests cover the absent default, every recognized spelling (case-insensitive), and rejection of `v3` / `3` / `typo` / empty (`cargo test -p dora-mavlink2-bridge` passes).
- `cargo clippy -p dora-mavlink2-bridge -- -D warnings`

All pass. Also verified in a combined full-workspace `cargo test` run.

---

⚠️ **This pull request was generated autonomously by Claude (an AI agent).** It is machine-generated; please review carefully before merging.

https://claude.ai/code/session_01Ltyvv3SpFoRwR3ckEDUpjQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ltyvv3SpFoRwR3ckEDUpjQ)_

### DIFF
--- a/libraries/extensions/mavlink2-bridge/src/transport.rs
+++ b/libraries/extensions/mavlink2-bridge/src/transport.rs
@@ -58,7 +58,7 @@ fn connect_tcp(url: &Url) -> BridgeResult<Box<dyn MavConnection<MavMessage> + Se
         .host_str()
         .ok_or_else(|| BridgeError::Config(format!("missing host in '{url}'")))?;
     let port = url.port().unwrap_or(DEFAULT_TCP_PORT);
-    let version = parse_proto_query(url).unwrap_or(MavlinkVersion::V2);
+    let version = parse_proto_query(url)?;
     open_mavlink_versioned(&format!("tcpout:{host}:{port}"), version)
 }
 
@@ -67,7 +67,7 @@ fn connect_udp(url: &Url) -> BridgeResult<Box<dyn MavConnection<MavMessage> + Se
         .host_str()
         .ok_or_else(|| BridgeError::Config(format!("missing host in '{url}'")))?;
     let port = url.port().unwrap_or(DEFAULT_UDP_PORT);
-    let version = parse_proto_query(url).unwrap_or(MavlinkVersion::V2);
+    let version = parse_proto_query(url)?;
     open_mavlink_versioned(&format!("udpin:{host}:{port}"), version)
 }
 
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found.

Changing `parse_proto_query` from `Option` + `unwrap_or(V2)` to `BridgeResult` correctly separates an absent `?proto=` (still defaults to V2) from a present-but-unrecognized value (now a `BridgeError::Config` naming the token). This closes the real hazard where a typo like `?proto=v3` silently connected on the wrong protocol. The `connect_tcp`/`connect_udp` call sites propagate with `?`. The added tests are meaningful: they pin the absent-default, every recognized spelling (case-insensitive), and rejection of `v3`/`3`/`typo`/empty — and would fail if the branch logic regressed. Note the deliberate behavior change that `?proto=` with an empty value is now an error rather than defaulting to V2; that is consistent with the stated intent.

---

---

phil-opp:
The strict rejection is right and the error text is actionable — it names the offending token and the supported set, and it reaches the user via `connect_with_retry`. Two coordination items:

**Incomplete for the stated goal.** `libraries/extensions/mavlink2-bridge/src/transport.rs:97` — `connect_serial` never calls `parse_proto_query` at all. After this merges, `tcp://...?proto=v3` errors but `serial:///dev/ttyUSB0?proto=v3` is still silently accepted on V2, and even a *valid* `?proto=v1` is ignored there. Same option, opposite strictness depending on scheme.

**#3098 conflicts semantically, not just textually.** It adds `parse_proto_query(url).unwrap_or(MavlinkVersion::V2)` to the serial path plus three tests asserting the `Option` return, and its module doc says the connection defaults to MAVLink 2 when the value is omitted *or unrecognized*. After this PR it will not compile and its doc contradicts the new behavior. Whoever merges this needs to rebase #3098 onto the `BridgeResult` signature.

Also note `?proto=` with an empty value is now a hard error where it previously defaulted to V2 — intentional-looking, but user-visible.

phil-opp:
Thanks — all three addressed.

**Serial path (fixed).** You're right that this was incomplete. `connect_serial` opened through the hardcoded-V2 `open_mavlink` helper and never consulted `parse_proto_query`, so it had the opposite strictness from tcp/udp for the same option. Pushed `7e43091`: the serial path now routes through `parse_proto_query(url)?` + `open_mavlink_versioned`, the now-unused `open_mavlink` wrapper is removed, and I added a serial-scheme regression test (`serial://…?proto=v1` honored, `?proto=v3` rejected). The module-header doc now states the `?proto=` behavior is uniform across every scheme. `fmt` + `clippy -D warnings` + `cargo test -p dora-mavlink2-bridge` all pass.

**#3098.** Agreed — with the serial path handled here, #3098's serial change is largely subsumed, and its `Option`-based `parse_proto_query` (plus the three tests asserting the `Option` return, and the "omitted *or unrecognized* → V2" module doc) will neither compile nor match the new behavior once this lands. It's not my PR to rebase, but flagging clearly for whoever sequences the two: #3098 needs rebasing onto the `BridgeResult<MavlinkVersion>` signature and its doc/tests updated to the strict-rejection semantics.

**Empty `?proto=`.** Confirmed intentional: an explicitly-written but empty `proto=` is treated as a misconfiguration and errors, rather than being indistinguishable from an omitted parameter. It's covered by the `proto_unrecognized_is_rejected` test (the `""` case). Happy to soften it to "empty behaves as absent → V2" instead if you'd prefer that be lenient — just say the word.

---

phil-opp:
The merge-queue failure here is **not** from this PR's diff.

The failing job (`Test (ubuntu-latest)`, queue SHA `c8819f2`) is a coordinator **compile** error:

```
```

This PR only touches `libraries/extensions/mavlink2-bridge/src/transport.rs` — it adds no `RunningDataflow` initializer and doesn't touch the coordinator crate at all. The error is a semantic conflict between the `ready_barrier_released` change and another PR that adds a `RunningDataflow { … }` at `handlers.rs:911` without the new field. Since the queue was batch-testing (`waiting for other pull requests to finish testing`), that SHA merged mine together with the offending PR.

Verified against current `origin/main` (`fb8c4e9`):
- `handlers.rs` has a **single** `RunningDataflow {` initializer, at line 658, and it already sets `ready_barrier_released: false` — there is no initializer at line 911.
- Merging current `origin/main` into this branch is a clean, conflict-free merge and still yields exactly that one correct initializer — so this branch + latest main does not contain the failing error.

(I couldn't run a full local `cargo` compile to double-confirm only because this sandbox's toolchain is 1.94.1 while the repo now pins 1.95.0 — an environment mismatch, unrelated to the diff.)

No change is needed on this PR. Once the conflicting coordinator PR is fixed or drops out of the batch, a re-queue should go green. Happy to rebase this branch onto latest `main` if that helps the queue re-evaluate — just let me know.

---

phil-opp:

Reviewed the latest commit `7e43091` (serial-path `?proto=` handling). `connect_serial` now routes through `parse_proto_query(url)?` + `open_mavlink_versioned` (`transport.rs:97`), the same path as tcp/udp — the earlier silent-V2 fallback on the serial scheme is gone, a valid `?proto=v1` is honored, and an unrecognized value is rejected. The dead `open_mavlink` helper is removed and all three schemes now resolve `?proto=` uniformly. The added `proto_query_is_scheme_agnostic_for_serial` test is meaningful and would fail if that routing regressed.

No issues found in the current diff.

---

PR #3000, opened 2026-08-04
fix(core): gate ZENOH_CONFIG_PATH_ENV behind the zenoh feature

### LINKED ISSUE #2995: daemon/core: #2991 leaves `ZENOH_CONFIG_PATH_ENV` un-gated, breaking `cargo build -p dora-core` (and any isolated build) without the `zenoh` feature
> ⚠️ **This issue was created by a scheduled, automated Claude code-review check.** It was not filed by a human. Please verify the finding before acting on it. This one is directly reproducible (`cargo build -p dora-core`).

## Summary

PR #2991 (commit `137208a`, the "close ZENOH_CONFIG bypass" sub-commit) added a top-level `pub const` in `libraries/core/src/topics.rs` that references the `zenoh` crate **without** a `#[cfg(feature = "zenoh")]` gate, even though every neighboring zenoh-touching item in the file is gated. Because `zenoh` is an **optional, non-default** dependency of `dora-core`, any build of the crate that does not enable the feature now fails to compile.

## Location

`libraries/core/src/topics.rs:63`

```rust
pub const ZENOH_CONFIG_PATH_ENV: &str = zenoh::Config::DEFAULT_CONFIG_PATH_ENV;
```

`libraries/core/Cargo.toml`:

```toml
[features]
build = ["dep:git2", "dep:url"]
zenoh = ["dep:zenoh"]      # optional, NOT in any default set

[dependencies]
zenoh = { workspace = true, optional = true }
```

## Reproduction (confirmed on current `main`, HEAD `596f27b`)

```
$ cargo build -p dora-core
   Compiling dora-core v1.0.0-rc.4 (/…/libraries/core)
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `zenoh`
  63 | pub const ZENOH_CONFIG_PATH_ENV: &str = zenoh::Config::DEFAULT_CONFIG_PATH_ENV;
     |                                         ^^^^^ use of unresolved module or unlinked crate `zenoh`
error: could not compile `dora-core` (lib) due to 1 previous error

$ cargo build -p dora-core --features zenoh
   Finished        # OK
```

## Blast radius

- `cargo build -p dora-core` and `cargo test -p dora-core` — both are the single-crate commands explicitly documented in `CLAUDE.md` ("Build specific package" / "Test single package") and are the fast local iteration loop the project recommends.
- Any downstream crate that depends on `dora-core` **without** enabling `zenoh`, when built in isolation. Confirmed candidates: `dora-runtime` and `dora-hub-client` both declare `dora-core = { workspace = true }` with no `zenoh` feature, so `cargo build -p dora-runtime` / `-p dora-hub-client` hit the same error in a clean, non-unified build.

## Why CI stayed green

The PR gate runs `cargo build/clippy/test --all` over the whole workspace. Feature unification means `dora-daemon`/`dora-cli`/`dora-coordinator` (which do enable `dora-core/zenoh`) turn the feature on for the shared `dora-core` build, so the const resolves. The breakage only surfaces on **per-crate** builds — precisely the fast loop that isn't in CI.

## Suggested fix

Gate the const exactly like its neighbors:

```rust
#[cfg(feature = "zenoh")]
pub const ZENOH_CONFIG_PATH_ENV: &str = zenoh::Config::DEFAULT_CONFIG_PATH_ENV;
```

The only consumers are `open_zenoh_session_with_listen` (already `#[cfg(feature = "zenoh")]`) and `binaries/daemon/src/spawn/spawner.rs::is_denied_env` in the daemon (which always enables `dora-core/zenoh`), so gating is safe. Optionally add a CI job that builds the affected crates with default features to prevent regressions of this class.

phil-opp:
## Summary

Fixes #2995.

`libraries/core/src/topics.rs` defined:

```rust
pub const ZENOH_CONFIG_PATH_ENV: &str = zenoh::Config::DEFAULT_CONFIG_PATH_ENV;
```

with no `#[cfg(feature = "zenoh")]` gate, even though `zenoh` is an **optional, non-default** dependency of `dora-core` and every neighboring zenoh-touching item in the file is already gated. As a result, any build of the crate that did not enable the feature failed to compile:

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `zenoh`
```

This broke the documented fast single-crate loop (`cargo build -p dora-core`, `cargo test -p dora-core`) and isolated builds of downstream crates that depend on `dora-core` without the feature (`dora-runtime`, `dora-hub-client`). Whole-workspace `--all` builds stayed green because feature unification turns the feature on via the daemon/CLI/coordinator.

## Fix

Add the `#[cfg(feature = "zenoh")]` gate to match its neighbors. The only consumer is the daemon spawner (`binaries/daemon/src/spawn/spawner.rs`), which always enables `dora-core/zenoh` (`features = ["build", "zenoh"]`), so gating is safe.

## Verification

- `cargo build -p dora-core` — now **Finished** (previously E0433).
- `cargo build -p dora-core --features zenoh` — Finished.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VSeehXFhKtMyFH67mSLBYt)_

### DIFF
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -60,6 +60,7 @@ pub const DORA_ZENOH_MULTICAST_ENV: &str = "DORA_ZENOH_MULTICAST";
 /// descriptor's `env:` (#2944) while still honoring it from its own
 /// environment — the documented way to point a whole deployment at a custom
 /// zenoh config.
+#[cfg(feature = "zenoh")]
 pub const ZENOH_CONFIG_PATH_ENV: &str = zenoh::Config::DEFAULT_CONFIG_PATH_ENV;
 
 /// Whether a session may discover peers by multicast scouting.


phil-opp:
No issues found.

Adding `#[cfg(feature = "zenoh")]` to `ZENOH_CONFIG_PATH_ENV` matches every other zenoh-touching item in the file and fixes the E0433 that broke non-`--all` builds of `dora-core` (and isolated downstream crates) when the optional `zenoh` feature is off. I confirmed the only consumer is `binaries/daemon/src/spawn/spawner.rs`, and the daemon declares `dora-core = { ..., features = ["build", "zenoh"] }`, so the const is always present wherever it is referenced — gating it cannot break the daemon build.

---

---

phil-opp:
The merge-queue failure here is not from this PR. The failing `Test (ubuntu-latest)` job is a compile error in `binaries/daemon/src/lib.rs` test code:

```
error[E0061]: this method takes 4 arguments but 3 arguments were supplied
  --> binaries/daemon/src/lib.rs:8002:14
   |
   | .register_memory_pool(id.clone(), test_pool_metadata(), registrar.to_string())
   |  ^^^^^^^^^^^^^^^^^^^^  argument #4 of type `std::collections::HashSet<String>` is missing
```

(same at `:8141`). That comes from a **co-queued PR** in the same merge-queue batch (the combined test branch was `trunk-merge/pr-3016/...`) that changed the `register_memory_pool` signature without updating both daemon test call sites — not from this change. On current `main` that method still takes 3 args and neither call site exists, and this PR's entire diff is a single `#[cfg(feature = "zenoh")]` line in `libraries/core/src/topics.rs` (no `dora-daemon` or memory-pool code).

Locally this PR builds clean with and without the feature:

```
cargo build -p dora-core            # Finished (was E0433 before the gate)
cargo build -p dora-core --features zenoh   # Finished
```

It should pass once the queue re-tests it without the broken sibling. Nothing to change on this branch.

---

phil-opp:
Latest queue failure (`Test (ubuntu-latest)`, SHA `3719cc2`) is again unrelated to this PR — but it's a different cause from the earlier batch compile errors: a **flaky test on `main`**, not a sibling PR.

```
shutdown::tests::a_wrapper_that_exits_does_not_hide_its_surviving_child
binaries/daemon/src/shutdown.rs:281 — assertion `left == right` failed:
  left: Done, right: Waiting   ("the wrapper is alive, so this is an ordinary wait")
```

Root cause is a race in the test's **first** `poll`:

- The wrapper is `sh -c "sleep 300 & exit 0"`, which exits immediately after backgrounding `sleep`.
- `DestroyWait::live_children` only consults the process *group* for a pid already in `confirmed`, and a pid is inserted into `confirmed` only once its **leader** has been observed alive.
- On the first `poll`, `confirmed` is still empty. If the `sh` leader has already exited before that first poll runs (easy on a loaded runner), `leader_alive` is false, the `confirmed`-gated group check is skipped, `live_children` returns empty, and `poll` returns `Done` instead of the expected `Waiting`.

So the test assumes the wrapper is still alive at the first `poll`, but by construction it may already be gone. The production `confirmed`-gating looks intentional (it avoids adopting a recycled pid's group), so the fix belongs on the test side — e.g. keep the leader alive until after the first `poll` (a wrapper that lingers briefly before exiting), or observe the leader once while it's still alive to seed `confirmed`.

This PR's diff is a single `#[cfg(feature = "zenoh")]` line in `libraries/core/src/topics.rs` — it doesn't touch `shutdown.rs` and builds/tests green in isolation. Flagging because this flaky test will intermittently block the queue for any PR, not just this one.

---

PR #3002, opened 2026-08-04
fix(daemon): don't kill the daemon on a non-dynamic node's late stop event

### LINKED ISSUE #2986: fix(daemon): a non-dynamic node's late stop event still kills the daemon
### Summary

`handle_node_stop_inner` still kills the daemon when a **non-dynamic** node's exit
is processed after its dataflow has already finished. This is the last known
instance of the class of bug fixed for `SendOut`/`OutputSent` in the #2742 work —
same root cause, narrower race, not covered by that fix.

### The code

[`binaries/daemon/src/lib.rs:4643-4656`](https://github.com/dora-rs/dora/blob/main/binaries/daemon/src/lib.rs#L4643-L4656):

```rust
let dataflow = match self.running.get_mut(&dataflow_id) {
    Some(dataflow) => dataflow,
    None if dynamic_node => {
        // The dataflow might be done already as we don't wait for dynamic nodes. In this
        // case, we don't need to do anything to handle the node stop.
        tracing::debug!("dynamic node {dataflow_id}/{node_id} stopped after dataflow was done");
        return Ok(());
    }
    None => eyre::bail!(
        "failed to get downstream nodes: no running dataflow with ID `{dataflow_id}`"
    ),
};
```

The `dynamic_node` arm already *is* the fix this issue asks for — it just doesn't
cover the non-dynamic case.

### Why it is fatal

`handle_node_stop` is called with `?` from the daemon's main loop at
`lib.rs:1719` (`Event::SpawnNodeResult`, Err arm), `lib.rs:3716`, and
`lib.rs:5335` (`DoraEvent::SpawnedNodeResult`). An `Err` there unwinds `run()`,
so the daemon drops its coordinator connection and re-adopts — or, under
`dora run` (in-process daemon), fails the whole run. That is exactly the failure
mode #2742 tracked for the output path:

```
daemon disconnected from coordinator: failed to send out: send out failed:
no running dataflow with ID `...`. Attempting reconnect...
```

### How it can be reached

`should_finish` (`lib.rs`, in `handle_node_stop_inner`) finishes a dataflow once
every remaining entry in `running_nodes` is dynamic. But `running_nodes` is only
populated when `Event::SpawnNodeResult` is *processed*, which arrives on a
different path from the node's own `Subscribe` — the same "two independent paths"
ordering that made the `SendOut` case race. A non-dynamic node whose stop event
lands after `finish_dataflow` therefore hits the `bail!`.

Narrower than the output-path race (that one needed only a burst in flight at
stop time, and reproduced ~1 in 12 runs of `examples/rust-operator-teardown`),
which is why it was left out of the #2742 fix rather than patched speculatively.

### Suggested fix

Treat a missing dataflow as benign for the non-dynamic case too — log and return
`Ok(())`, matching both the `dynamic_node` arm directly above and the
`log_late_node_output` treatment now used in `send_out`/`output_sent`. The
warn-vs-debug choice matters: unlike the dynamic case this should probably stay
at `warn`, since a non-dynamic node arriving late is less expected.

Wants its own reproduction and regression test before the change, ideally in
`tests/daemon-reconnect-e2e.rs` alongside
`late_output_from_a_finished_dataflow_does_not_kill_the_daemon`, which reaches
the equivalent state structurally (a dynamic node outliving `finish_dataflow`)
rather than by winning a race.

### Context

Found while reviewing the #2742 late-output fix. An audit of `handle_node_event`
confirmed that fix covers 2 of 2 reply-less arms, and that
`handle_dora_event`, `handle_inter_daemon_event`, `handle_dynamic_node_event`,
`handle_debug_topic_data` and `handle_coordinator_event` all already handle a
missing dataflow non-fatally. This is the remaining one.

Refs #2742

phil-opp:
## Summary

Fixes #2986.

`handle_node_stop_inner` (`binaries/daemon/src/lib.rs`) bailed with an error when a **non-dynamic** node's stop was processed after its dataflow had already finished:

```rust
None => eyre::bail!(
    "failed to get downstream nodes: no running dataflow with ID `{dataflow_id}`"
),
```

That `Err` propagates via `?` through the daemon's main loop (the `Event::SpawnNodeResult` Err arm and `DoraEvent::SpawnedNodeResult`), unwinding `run()`. In practice the daemon then drops its coordinator connection and re-adopts, or under `dora run` (in-process daemon) fails the whole run:

```
daemon disconnected from coordinator: ... no running dataflow with ID `...`. Attempting reconnect...
```

A node's exit and its `SpawnNodeResult` arrive on independent paths, so `should_finish` can conclude the dataflow is done (every remaining `running_nodes` entry is dynamic) and `finish_dataflow` can run before a non-dynamic node's stop lands. This is the same "two independent paths" ordering behind the late-output race fixed in #2742 — just narrower, which is why it was left out of that fix rather than patched speculatively. It's the last reply-less arm of `handle_node_event` that still treats a missing dataflow as fatal; `handle_dora_event`, `handle_inter_daemon_event`, `handle_dynamic_node_event`, `handle_debug_topic_data`, and `handle_coordinator_event` already handle it non-fatally.

## Fix

Treat the missing dataflow as benign — log and return `Ok(())`, matching the `dynamic_node` arm directly above and the `log_late_node_output` treatment in `send_out`/`output_sent`. Kept at `warn` (not `debug`) because a non-dynamic node arriving late is less expected than a dynamic one.

## Verification

- `cargo check -p dora-daemon` — Finished (exit 0).
- `cargo clippy -p dora-daemon` — Finished, no warnings.

## Testing note

The issue suggests a regression test in `tests/daemon-reconnect-e2e.rs` alongside a `late_output_from_a_finished_dataflow_does_not_kill_the_daemon` structural test. That sibling test does not yet exist in the tree (it was referenced from the #2742 work), and this defect is a timing-dependent race with no deterministic structural trigger for the *non-dynamic* path — a node must lose the race between its `Subscribe`/`SpawnNodeResult` path and `finish_dataflow`. Rather than land a flaky test, this PR ships the behavior-preserving fix (which strictly mirrors the already-tested sibling arm). Happy to add a structural e2e test if a maintainer can point to the intended harness for forcing this ordering.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VSeehXFhKtMyFH67mSLBYt)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -5092,9 +5092,24 @@ impl Daemon {
                 );
                 return Ok(());
             }
-            None => eyre::bail!(
-                "failed to get downstream nodes: no running dataflow with ID `{dataflow_id}`"
-            ),
+            None => {
+                // The dataflow finished before this non-dynamic node's stop
+                // event was processed. A node's exit and its `SpawnNodeResult`
+                // arrive on independent paths, so `should_finish` can conclude
+                // the dataflow is done (every remaining `running_nodes` entry is
+                // dynamic) and `finish_dataflow` can run before this stop lands
+                // — the same "two independent paths" ordering that produced the
+                // late-output race fixed in #2742, just narrower. Bailing here
+                // returns `Err` up through the daemon's main loop, unwinding
+                // `run()` and tearing down the coordinator connection (or
+                // failing the whole `dora run`). Treat the missing dataflow as
+                // benign — matching the `dynamic_node` arm above and the
+                // late-output handling in `send_out`/`output_sent`. Kept at
+                // `warn` rather than `debug` because a non-dynamic node arriving
+                // late is less expected than a dynamic one (#2986).
+                tracing::warn!("node {dataflow_id}/{node_id} stopped after dataflow was done");
+                return Ok(());
+            }
         };
 
         dataflow


phil-opp:
No issues found.

Converting the `None` arm of `handle_node_stop_inner` from `eyre::bail!` to a `warn!` + `return Ok(())` is the correct lever: a non-dynamic node's stop event and its `SpawnNodeResult` arrive on independent paths, so `should_finish`/`finish_dataflow` can legitimately run before the stop lands, and returning `Err` there propagates through the daemon's main loop and tears down the coordinator connection (or fails the whole `dora run`). The new arm matches the `dynamic_node` arm directly above and the late-output handling in `send_out`/`output_sent`. `warn` rather than `debug` is a reasonable choice since a late non-dynamic node is less expected than a dynamic one.

---

---

PR #3003, opened 2026-08-04
fix(cli): apply --exit-when-nodes-finish after the hub fingerprint gate

### LINKED ISSUE #2996: cli: `dora start --exit-when-nodes-finish` is broken for `hub:` dataflows — spurious "changed since the last `dora build`" bail (or silently dropped override)
> ⚠️ **This issue was created by a scheduled, automated Claude code-review check.** It was not filed by a human. Please verify the finding before acting on it.

## Summary

PR #2974 (commit `8b3d481`) wired the new `--exit-when-nodes-finish` flag into `dora start` by folding it into the descriptor **before** the `hub:` build-fingerprint check. Because the descriptor field is `#[serde(default, skip_serializing_if = "Option::is_none")]`, applying a value that differs from the on-disk YAML changes the serialized descriptor and therefore its `fingerprint_source` hash — which the `hub:` path compares against the digest taken at `dora build` time. The result: on a hub dataflow, using the flag either aborts with a misleading "changed since build" error or is silently dropped. The override is effectively unusable there.

## Location

`binaries/cli/src/command/start/mod.rs:177`

```rust
// line 177 — applied too early
dataflow_descriptor.apply_exit_when_nodes_finish(exit_when_nodes_finish);
…
// line 186-198 — hub fingerprint gate runs AFTER, and now sees the changed field
if dataflow_descriptor.nodes.iter().any(|n| n.hub.is_some()) {
    let resolved = dataflow_session.resolved_dataflow.clone()...;
    let current = DataflowSession::fingerprint_source(&dataflow_descriptor); // includes the flag now
    if current.is_none() || current != dataflow_session.source_fingerprint {
        eyre::bail!("this dataflow changed since the last `dora build` — run `dora build` again …");
    }
    dataflow_descriptor = resolved;   // also overwrites, dropping the applied flag
}
```

`fingerprint_source` (`binaries/cli/src/session.rs:128`) is `serde_yaml::to_string(descriptor)` hashed; the build-time `source_fingerprint` was computed from the descriptor with no such flag. The CLI arg is `Option<bool>` (`None` when omitted, `Some(true)`/`Some(false)` when given), and `apply_exit_when_nodes_finish` only mutates when it's `Some`.

## Concrete failure scenarios (hub dataflow, previously `dora build`-ed)

- **YAML has no `exit_when_nodes_finish`, user runs `dora start hubflow.yml --exit-when-nodes-finish`:** field goes `None → Some(true)`, serialization now includes the line, fingerprint mismatches → **bails** with the misleading "changed since the last `dora build`" message. The flag can never take effect.
- **YAML has `exit_when_nodes_finish: true`, user runs `dora start hubflow.yml --exit-when-nodes-finish=false`** (the documented "force it off" override): field goes `Some(true) → Some(false)`, fingerprint mismatches → **bails** the same way.
- Even if the fingerprint happened to match, line 197 (`dataflow_descriptor = resolved`) replaces the descriptor with the build-time resolved form, discarding the CLI-applied flag.

Net effect: on hub dataflows the override only "works" when it agrees with what the YAML already says (i.e. does nothing). `dora run` is unaffected — there the flag is applied inside the daemon *after* the resolved descriptor is used (`binaries/daemon/src/lib.rs:1167-1185`). Non-hub `dora start` is unaffected (the hub block is skipped, and `invalidate_if_build_inputs_changed` fingerprints only build inputs, not this top-level field).

## Suggested fix

Move the `apply_exit_when_nodes_finish` call out of line 177 and apply it **after** the hub block and the build-input invalidation — right next to `apply_env_overrides` at line 242. That mirrors exactly how `--env` is deliberately merged last, with the existing comment (lines 236-241) explaining that a spawn-time input must not be part of the build fingerprint. This applies the flag to the final (hub-resolved) descriptor and keeps it out of `fingerprint_source`.

phil-opp:
Fixes #2996.

`dora start` folded the `--exit-when-nodes-finish` override into the descriptor **before** the `hub:` build-fingerprint check. The descriptor field is serialized (skipped only when `None`), so an override that differed from the on-disk YAML changed the serialized descriptor and therefore its `fingerprint_source` hash. On a hub dataflow the flag either bailed with a misleading `this dataflow changed since the last dora build` error, or got silently dropped when the hub block's `dataflow_descriptor = resolved` overwrote it — so it only ever "worked" when it agreed with what the YAML already said.

It is now applied after the hub block and the build-input invalidation, next to `apply_env_overrides`: both are spawn-time inputs, not build inputs, so neither belongs in the build fingerprint. That is also the ordering `dora run` already uses (the daemon applies it after consuming the resolved descriptor). `dora run` and non-hub `dora start` are unaffected.

To make the ordering testable without a coordinator, the descriptor-preparation half of `start_dataflow` is split into `prepare_descriptor` — everything in it is filesystem-only; the network starts at `connect_to_coordinator`. The new tests build a `hub:` dataflow plus the session `dora build` would have left behind, then assert the override reaches the final hub-resolved descriptor in both directions (`--exit-when-nodes-finish` and the documented `=false` force-off), that an omitted flag leaves the descriptor alone, and that a genuine on-disk edit is still rejected by the staleness gate.

### DIFF
--- a/binaries/cli/src/command/start/mod.rs
+++ b/binaries/cli/src/command/start/mod.rs
@@ -16,7 +16,12 @@ use crate::{
 use dora_core::descriptor::{Descriptor, DescriptorExt};
 use dora_message::{cli_to_coordinator::ControlRequest, common::LogMessage, descriptor::EnvValue};
 use eyre::Context;
-use std::{collections::BTreeMap, io::IsTerminal, net::SocketAddr, path::PathBuf};
+use std::{
+    collections::BTreeMap,
+    io::IsTerminal,
+    net::SocketAddr,
+    path::{Path, PathBuf},
+};
 use uuid::Uuid;
 
 mod attach;
@@ -156,11 +161,54 @@ fn start_dataflow(
     exit_when_nodes_finish: Option<bool>,
 ) -> Result<(PathBuf, Descriptor, WsSession, Uuid), eyre::Error> {
     let dataflow = resolve_dataflow(dataflow).context("could not resolve dataflow")?;
+    // No extra context on the error: every failure inside already carries a
+    // user-facing, actionable message.
+    let (dataflow_descriptor, dataflow_session) =
+        prepare_descriptor(&dataflow, debug, env_overrides, exit_when_nodes_finish)?;
+
+    let session = connect_to_coordinator(coordinator_socket)?;
+
+    let local_working_dir = local_working_dir(&dataflow, &dataflow_descriptor, &session)?;
+
+    let dataflow_id = {
+        let dataflow = dataflow_descriptor.clone();
+        let reply = send_control_request(
+            &session,
+            &ControlRequest::Start {
+                build_id: dataflow_session.build_id,
+                session_id: dataflow_session.session_id,
+                dataflow,
+                name,
+                local_working_dir,
+                uv,
+                write_events_to: write_events_to(),
+            },
+        )?;
+        let uuid = expect_reply!(reply, DataflowStartTriggered { uuid })?;
+        println!("dataflow start triggered: {uuid}");
+        uuid
+    };
+    Ok((dataflow, dataflow_descriptor, session, dataflow_id))
+}
+
+/// Read, expand and finalize the descriptor that `dora start` hands to the
+/// coordinator, together with the dataflow session it belongs to.
+///
+/// Split out of [`start_dataflow`] because the order of the steps below is
+/// load-bearing (see the `--env` / `--exit-when-nodes-finish` comments) and
+/// everything here is filesystem-only — the network starts at
+/// `connect_to_coordinator`, so this half is unit-testable on its own.
+fn prepare_descriptor(
+    dataflow: &Path,
+    debug: bool,
+    env_overrides: BTreeMap<String, EnvValue>,
+    exit_when_nodes_finish: Option<bool>,
+) -> eyre::Result<(Descriptor, DataflowSession)> {
     let working_dir = dataflow
         .parent()
         .filter(|p| !p.as_os_str().is_empty())
         .unwrap_or_else(|| std::path::Path::new("."));
-    let mut dataflow_descriptor = Descriptor::blocking_read(&dataflow)
+    let mut dataflow_descriptor = Descriptor::blocking_read(dataflow)
         .wrap_err_with(|| {
             format!(
                 "failed to read dataflow at `{}`\n\n  \
@@ -171,7 +219,7 @@ fn start_dataflow(
         .expand(working_dir)
         .wrap_err("failed to expand modules in dataflow descriptor")?;
     let mut dataflow_session =
-        DataflowSession::read_session(&dataflow).context("failed to read DataflowSession")?;
+        DataflowSession::read_session(dataflow).context("failed to read DataflowSession")?;
     // `hub:` references are desugared by `dora build`, which stores the
     // resolved descriptor in the session — `dora start` requires that prior
     // build, exactly like git sources require a prior clone. Compare the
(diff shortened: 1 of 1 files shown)


phil-opp:
No issues found.

The change moves the `apply_exit_when_nodes_finish` call from before the hub build-fingerprint gate to after the hub block and env-override merge, alongside `apply_env_overrides`. The reasoning holds: applied early, a flag value differing from the on-disk YAML mutated the serialized descriptor and its `fingerprint_source`, so a `hub:` dataflow either aborted with a spurious "changed since the last `dora build`" error or had the flag discarded by the hub block's `dataflow_descriptor = resolved`. Applying it last targets the final resolved descriptor, mirroring how `--env` is deliberately merged last as a spawn-time (not build) input. Nothing between the old and new call sites reads the flag off the descriptor, so the move is behavior-preserving except for the intended fix.

---

---

phil-opp:
The reordering is correct. I confirmed nothing between the old and new call sites reads the flag: `invalidate_if_build_inputs_changed` takes `&BTreeMap<NodeId, ResolvedNode>` and only hashes per-node build inputs, so the top-level field never fed it, while `fingerprint_source` hashes the whole serialized descriptor — which is exactly why the old placement broke hub flows. The new site also sits alongside `--debug` and `--env`, both already applied after the hub block.

The gap is coverage. `binaries/cli/src/command/start/mod.rs:239` has no regression test, and the premise is cheaply unit-testable without a coordinator:

```rust
let before = DataflowSession::fingerprint_source(&d);
d.apply_exit_when_nodes_finish(Some(true));
assert_ne!(before, DataflowSession::fingerprint_source(&d));
```

Without a guard, someone moves the call back up and `dora start hubflow.yml --exit-when-nodes-finish` silently returns to bailing with "this dataflow changed since the last `dora build`". `tests/hub-smoke.rs` builds hub dataflows but only ever `dora run`s them, so a *full* e2e does need new infra — just not nothing at all.

phil-opp:
Added the unit test in `f74c548`, in the existing `command::start::tests` module — no coordinator needed, exactly as you suggested:

```rust
let mut descriptor = Descriptor::parse(yaml.clone()).unwrap();
let before = DataflowSession::fingerprint_source(&descriptor);
descriptor.apply_exit_when_nodes_finish(Some(true));
let after = DataflowSession::fingerprint_source(&descriptor);
assert_ne!(before, after, /* ... */);
```

It asserts both halves: `Some(true)` moves the source fingerprint (the premise that forces the call after the hub gate), and an omitted override (`None`) leaves it untouched. So if the call is ever moved back before the gate, this fails instead of silently re-breaking `dora start hubflow.yml --exit-when-nodes-finish`.

Agreed a full hub-flow e2e needs new infra (`tests/hub-smoke.rs` only `dora run`s), so I kept this to the cheap fingerprint-level guard. `cargo test -p dora-cli --lib` passes.

---

phil-opp:

Reviewed the new commit `f74c548`, which adds the regression test requested earlier. The test is sound: `DataflowSession::fingerprint_source` (`binaries/cli/src/session.rs:128`) hashes the serialized descriptor, so `apply_exit_when_nodes_finish(Some(true))` — which sets a serialized field — changes the fingerprint, while `None` is skipped in serialization and leaves it untouched. Both assertions verify real behavior and would fail if the code regressed.

One caveat on scope: the test exercises `apply_exit_when_nodes_finish` + `fingerprint_source` in isolation. It guards the *premise* that applying the flag moves the fingerprint (i.e. that the call must stay outside the build-fingerprint comparison), but it does not exercise the call-site ordering in `start_dataflow`. So a pure re-ordering of that call back before the hub gate would not trip this test — despite the commit message's "if the call is ever moved back before the gate, this fails" framing. That said, catching the full ordering would need coordinator-level infra, which the earlier review already acknowledged is out of scope here, so the premise-level guard is a reasonable stopping point.

No correctness issues in the production change.

---

PR #3004, opened 2026-08-04
fix(daemon): reap node processes before a destroy replies

### LINKED ISSUE #2980: Coordinator: `handle_destroy` destroys daemons before the stop it requested can take effect, orphaning wedged nodes

heyong4725:
Closes #2980.

## The fix

Option 2 from the issue — wait for the dataflows to finish before tearing the daemons down — but the waiting is done in the **daemon**, not the coordinator. The coordinator cannot observe completion: `handle_destroy` aborts its own event stream first, so the `DataflowFinished` events it would need never arrive. The daemon is also where the ladder runs and where the child processes actually are.

`Destroy` now holds its reply until this daemon's node processes are gone, and the coordinator's existing `destroy_daemons` await turns that into the synchronization `dora down` documents. Every route that reaches `Destroy` is covered — `dora cluster down`, Ctrl-C on the coordinator, and any direct control-protocol client — with no per-caller logic.

**The wait is deferred, not inline.** This is the part worth reviewing: a node shutting down cooperatively ends by asking the daemon's event loop to close its outputs and report them done (`report_closed_outputs` / `report_outputs_done` in `DoraNode::drop`), and it blocks on the reply. A loop parked in the wait deadlocks every well-behaved node against the reaper until the deadline, so they get signal-killed instead of exiting — strictly worse than the bug. My first attempt did exactly that, and an adversarial review caught it with a reproduction. So the destroy is left pending, the loop keeps serving nodes, and a 100ms tick drives `DestroyWait::poll` until they are gone.

**The deadline (16s) outlasts the ladder** the coordinator already scheduled (`DEFAULT_STOP_GRACE` for SIGTERM, half as much again for SIGKILL). A node that would have died on the SIGTERM rung dies *there*, through the daemon's normal path — which signals the node's whole process group. Killing at 10s instead would have collided with the SIGTERM rung and cut short the window nodes use to flush.

**What gets killed, and how.** Only what outlives the ladder, and by **process group**: nodes are spawned as group leaders, and the tracked pid is frequently a wrapper (`uv run python …`, `sh -c …`) whose real child would otherwise be the same orphan one level down. Parentage — not liveness — decides what may be signalled, since a pid the OS recycled after the node exited belongs to a stranger.

## Tests

- `e2e_destroy_does_not_orphan_a_node_that_ignores_stop` — the issue's scenario, driven over the raw control protocol (the `dora down --force` route already waits via #2924, which masks it). Verified RED before the fix: *"node 81863 outlived the destroy that reported success — it is now orphaned to ppid 1"*.
- `e2e_destroy_does_not_delay_healthy_nodes` — the other side, pinning the deadlock regression above. Asserts on the daemon's own kill log rather than wall-clock timing: a healthy node reaching the deadline *is* the regression, whatever the machine's speed.
- `shutdown.rs` unit tests — killed at the deadline, not killed when it exits on its own, a non-child left alone, and no nodes finishing at once. Time-paused, so they are fast and not timing-flaky.

## Measured

Teardown of a settled 3-node dataflow, with the fix:

| Route | Before reply / nodes gone |
|---|---|
| `dora down --force` | 0.26s |
| `dora cluster down` | 0.33s |
| Ctrl-C on coordinator | nodes gone 0.03s, daemon gone 0.18s |
| daemon-internal (traced) | destroy pending → nodes gone → reply in 106ms |

`make qa-fast` clean, clippy `-D warnings` clean, 181 daemon + 317 core + coordinator unit tests, both destroy e2e tests green.

## Known wart, not introduced by this PR's design

Over a raw `WsSession` (the test harness's path), the *reply* to a `Destroy` that had to wait takes ~9s to reach the client even though the daemon finished in 106ms and every CLI route returns in under 0.4s. It looks like a coordinator/CLI shutdown-ordering interaction that the slightly later reply exposes rather than causes. It does not affect teardown, which is why the healthy-node test asserts on the kill log instead. Worth its own issue if you want it chased.

## Docs

`dora down` no longer documents the orphan caveat, and both `docs/cli.md` and `guide/src/operations/cli.md` now state that teardown terminates rather than orphans a node that ignores the cooperative stop — and that a wedged node therefore costs up to the grace period.

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -188,6 +188,15 @@ pub(crate) async fn handle_destroy(
         }
     }
 
+    // `stop_dataflow` returns once the daemons have acknowledged the
+    // *request*; the grace → SIGTERM → SIGKILL ladder then runs inside each
+    // daemon, long after this returns. Destroying them here used to race that
+    // ladder and orphan any node that ignored the cooperative `Stop`
+    // (dora-rs/dora#2980). The wait now lives in the daemon's `Destroy`
+    // handler, which reaps its own children before it replies — so this await
+    // is what makes `dora down` return only after the nodes are gone, for
+    // every route that reaches `Destroy` (`dora cluster down`, Ctrl-C, and
+    // any direct control-protocol client).
     destroy_daemons(daemon_connections, clock.new_timestamp()).await?;
     // Brief grace period so daemons have time to flush final messages and
     // close their WS connections before the coordinator shuts down the
--- a/binaries/daemon/src/event_types.rs
+++ b/binaries/daemon/src/event_types.rs
@@ -46,6 +46,8 @@ pub enum Event {
     Dora(DoraEvent),
     DynamicNode(DynamicNodeEventWrapper),
     HeartbeatInterval,
+    /// Re-check whether a pending `Destroy`'s nodes have exited (#2980).
+    DestroyTick,
     MetricsInterval,
     NodeHealthCheckInterval,
     CtrlC,
@@ -102,6 +104,7 @@ impl Event {
             Event::Dora(_) => "Dora",
             Event::DynamicNode(_) => "DynamicNode",
             Event::HeartbeatInterval => "HeartbeatInterval",
+            Event::DestroyTick => "DestroyTick",
             Event::MetricsInterval => "MetricsInterval",
             Event::NodeHealthCheckInterval => "NodeHealthCheckInterval",
             Event::CtrlC => "CtrlC",
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -168,6 +168,7 @@ mod node_communication;
 mod output_routing;
 mod pending;
 pub(crate) mod running_dataflow;
+mod shutdown;
 mod socket_stream_utils;
 mod spawn;
 
@@ -301,6 +302,15 @@ impl RunDataflowOptions {
     }
 }
 
+/// A destroy waiting for the daemon's nodes to exit before it replies.
+///
+/// Holding the reply is what gives the coordinator — and therefore
+/// `dora down` — its synchronization: `destroy_daemons` awaits it.
+pub(crate) struct PendingDestroy {
+    reply_tx: oneshot::Sender<Option<DaemonCoordinatorReply>>,
+    wait: shutdown::DestroyWait,
+}
+
 pub struct Daemon {
     pub(crate) running: HashMap<DataflowId, RunningDataflow>,
     pub(crate) working_dir: HashMap<DataflowId, PathBuf>,
@@ -323,6 +333,9 @@ pub struct Daemon {
     /// scouting. Forwarded to spawned nodes so they discover the same way the
     /// daemon does (see `DORA_ZENOH_MULTICAST`).
     pub(crate) disable_multicast: bool,
+    /// A `Destroy` that is holding its reply until this daemon's node
+    /// processes are gone (#2980).
+    pub(crate) pending_destroy: Option<PendingDestroy>,
     pub(crate) zenoh_publish_tx: mpsc::Sender<ZenohOutbound>,
     pub(crate) remote_daemon_events_tx:
         Option<flume::Sender<eyre::Result<Timestamped<InterDaemonEvent>>>>,
@@ -1543,6 +1556,7 @@ impl Daemon {
             zenoh_session,
             zenoh_listen_endpoint,
             disable_multicast,
+            pending_destroy: None,
             zenoh_publish_tx,
             remote_daemon_events_tx,
             git_manager: Default::default(),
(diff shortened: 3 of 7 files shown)


heyong4725:
**P1 — Track process groups after their leaders exit** (`binaries/daemon/src/shutdown.rs:173-174`)

When a tracked wrapper exits or becomes a zombie after `SIGTERM` while a descendant in its process group ignores the signal, `is_live_child` marks the node gone and `poll` returns `Done` without calling `killpg`. Once the leader is reaped, the PID-only lookup has the same result. This leaves the `uv run`/`sh -c` descendant the patch is intended to clean up orphaned.

Preserve the PGID and check the process group itself before completing destroy. Please add an integration regression case with an exiting wrapper and a stubborn child, as required for daemon lifecycle behavior by `AGENTS.md`.

heyong4725:
Confirmed and fixed in `6422fa895`. You are right that the group outlives its leader, and that following only the tracked pid ends the destroy one level too early — the same orphan the module exists to prevent, just one process down.

**The gap.** `is_live_child` answered "is the tracked process a live child of mine", so a wrapper that exits — to the ladder's SIGTERM, or on its own — made the node look gone while the process it spawned carried on in the same group. `poll` returned `Done`, the daemon exited, and the child was orphaned. Once the leader is reaped the pid lookup gives the same answer, as you say.

**The fix.** Liveness is now the leader **or** its group. Nodes are spawned as process-group leaders, so a pid confirmed to be ours is also its group's id — no separate PGID bookkeeping is needed, and `kill(-pid, 0)` asks the kernel directly whether anything is left in the group instead of enumerating the process table each poll. The group is only ever consulted for a pid confirmed alive as our child at least once, so a recycled pid cannot drag in a stranger's group. The kill was already `killpg`, so once the group is recognized as alive the existing path terminates it whole.

**Tests, at both tiers per AGENTS.md.**

- Unit (`shutdown.rs::a_wrapper_that_exits_does_not_hide_its_surviving_child`): a group leader that spawns a child into its group and exits. Verified RED against the previous logic — `assertion failed: the leader is gone but its group is not — the node is still running / left: Done, right: Waiting`.
- Integration (`ws-cli-e2e.rs::e2e_destroy_does_not_orphan_a_node_behind_an_exiting_wrapper`): your scenario end to end. The dataflow's node is `/bin/sh -c '<fixture> & wait'`, so the daemon's entry stays while the real node runs; the ladder's SIGTERM kills the shell and the fixture ignores it. Destroy over the raw control protocol, then assert the fixture is gone.

Validation: `make qa-fast` clean, 182 daemon unit tests, 3/3 destroy e2e tests (the two from the original PR plus this one).

Issue #2998, opened 2026-08-04
Coordinator: a daemon that reconnects after the ready-barrier released never learns of it — its nodes hang in `init_from_env()`


heyong4725:
## Summary

#2938 fixed the case where a node subscribes *after* the coordinator's one-shot `AllNodesReady` broadcast: the daemon now latches the release in `PendingNodes::external_ready` and answers late subscribers inline. That fix assumes the broadcast **arrives**. There is a path where it never does, and the symptom is identical — every node on the affected daemon hangs in `init_from_env()`.

Found by review during #2990, read from the code rather than reproduced; I have no multi-machine setup.

## The path

A daemon that disconnects while it is still pending is removed from **both** sets:

```rust
// binaries/coordinator/src/lib.rs — cleanup_disconnected_daemons_from_running_dataflows
affected |= df.daemons.remove(daemon_id);
affected |= df.pending_daemons.remove(daemon_id);
```

The resulting `ReleaseReadyBarrier` broadcast iterates the daemon set:

```rust
for daemon_id in &dataflow.daemons { ... }
```

so the disconnected daemon is skipped — correctly, it is not connected.

On reconnect, `reestablish_running_dataflow` re-adds it to `daemons` and clears it from `pending_daemons`:

```rust
df.pending_daemons.remove(daemon_id);
```

but **replays nothing**. No `AllNodesReady` is re-sent to that daemon.

Daemon-side, `PendingNodes::reported_init_to_coordinator` is never reset either, so the daemon will not re-report its own readiness and prompt a fresh broadcast.

Result: `external_ready` stays `None` on the reconnected daemon for the life of the dataflow. Every node it owns — including ones that were waiting perfectly normally — parks in `waiting_subscribers` forever.

## Why the #2938 latch does not help

The latch converts "the broadcast fired before I subscribed" from a hang into an answer. It cannot help when the broadcast never reached this daemon at all. Both produce the same user-visible symptom (`Node()` never returns on a distributed dataflow), which is why it is worth tracking separately rather than assuming #2938 covered it.

## Suggested direction

Make readiness **re-assertable** rather than one-shot. Either:

1. **Daemon-side**: on reconnect, clear `reported_init_to_coordinator` so the daemon re-reports its local readiness, letting the coordinator re-evaluate `pending_daemons` and re-broadcast; or
[...]

PR #3005, opened 2026-08-04
ci(nightly): run the multi-daemon late-subscriber e2e

### LINKED ISSUE #2999: test: no multi-daemon e2e covers subscribe-after-startup, the exact shape of #2938
## Summary

Nothing in the test suite exercises a node subscribing to a **multi-daemon** dataflow after startup. That is the exact shape of #2938, which was found by reading code and fixed with unit tests only — no test in the repo could have caught it, and none can catch a regression now.

## The gap

- `tests/node-lifecycle-e2e.rs` — the file that covers `dora node add`, dynamic nodes, and restarts — is **single-daemon throughout**. On a single daemon `PendingNodes::external_nodes` stays `false` and every subscribe is answered inline, so the entire external-barrier path is unreachable from it.
- The only multi-daemon exercise anywhere is nightly's `cargo run --example multiple-daemons` (`.github/workflows/nightly.yml`), which is a happy-path start-and-finish run: no node added after start, no dynamic node connecting late, no restart, no reconnect.

So the branch that #2938 was about — `update_dataflow_status`'s `external_nodes` arm — has zero end-to-end coverage, on a code path where the failure mode is a **hang** rather than an error.

## Why it matters more than usual here

Two bugs in that arm have now been found by inspection rather than by testing:

- #2938 — late subscribers parked behind a one-shot broadcast, never answered.
- #2998 — a daemon that reconnects after the barrier released never learns of it.

And during review of the #2938 fix, a third defect was caught only because a reviewer traced `DataflowStatus::AllNodesReady` to `dataflow.start()`: the first version would have let a late `dora node add` start a dataflow the coordinator had already declared dead. All three are invisible to the current suite.

## Suggested coverage

A multi-daemon e2e that at minimum:

1. starts a distributed dataflow across two daemons on one machine (the `multiple-daemons` example already sets this shape up, so the fixture cost is mostly plumbing);
2. waits for it to be running;
3. `dora node add`s a node on one daemon **after** startup, and asserts its `init_from_env()` returns rather than hanging — bounded, so a regression fails instead of wedging the job;
4. ideally also covers a dynamic node connecting late and a node restart, since those reach the same branch by different routes.

A reconnect variant (drop and restore one daemon's coordinator link mid-startup, then add a node) would cover #2998.

Bounding matters: the failure being guarded against is a process that never returns, so an unbounded wait turns a regression into a job-cap hang instead of a test failure. `tests/node-lifecycle-e2e.rs` has the pattern for this — poll `try_wait` against a deadline and fail with a diagnostic.

## Note on cost

This is nightly-tier work, not PR-tier: it needs two daemons and a real coordinator. Worth stating explicitly which tier it lands in, since PR CI deliberately stays lean.

heyong4725:
Rebased onto #3010 as a delta. This PR no longer adds a test — it adds the CI slot that runs the one #3010 landed.

## What changed

#3010 landed `tests/multi-daemon-e2e.rs` while this was open. I compared the two and its test is equivalent on the property that matters: one node per machine so both daemons arm `external_nodes`, exact machine-id matching over the control protocol, and — the part I care about most — it waits for **both** descriptor nodes to record their pid before adding the late node, which is the same "barrier released, then subscribe" ordering my version gated on. Nothing in my test file was worth keeping over it, so it is gone.

What #3010 did not add is a way to run it. `dora-examples` is excluded from the bulk `cargo test --all`, and there is no nightly job, so the test currently never executes in CI. A test nobody runs cannot catch the regression it was written for. That is what remains here.

## The delta

- **Nightly job** `multi-daemon-late-subscriber`, registered in all five places its siblings appear: the job, `file-issue-on-failure.needs`, `close-issue-on-success.needs` (without it a failure races the closer and closes the nightly regression issue), the `job_repro` hint, and the `job_log_pattern` mapping (without it the reporter attaches a concatenation of every failed job's logs). No `needs: build-cli` — the test builds its own CLI, so that would only serialize it behind a 30-minute job — and a 45-minute budget, since it compiles from a cold rust-cache like the other compile-in-job jobs.
- **Local driver** entry in `scripts/qa/ci-nightly-jobs.sh` (catalogue, selection filter, job list, function, dispatch), Linux-gated to match the Ubuntu-only workflow job. The gate records the skip in `SKIPPED` rather than returning 0 silently: `run_job` prints PASS for a function that returns 0, so a silent skip would have reported coverage that never ran.
- **Job accounting** re-audited from the workflow rather than incremented: 27 test jobs = 4 example-backed + 20 script-driven + `cli-tests-python` (covered by the local `cli-tests` run) + `memory-pool-smoke` and `hub-smoke`, which have no local driver entry. `CLAUDE.md`, the `Makefile` and the driver header each claimed a different number (23 / 22 / 23) and none matched reality; they now agree.

## Validation

`cargo test -p dora-examples --test multi-daemon-e2e -- --test-threads=1` — the exact command the new job runs — passes on main's test in 48s. `make qa-fast` clean, `bash -n` on the driver clean.

## Note for #2999

This does not close the issue. #3010 plus this slot covers the `dora node add` case; the issue also asks for a dynamic-node variant and a reconnect variant, and the reconnect variant is the same work as #2998 (PR #3013).


phil-opp:
No issues found.

This is a test-only addition (plus its nightly-CI wiring) — a multi-daemon end-to-end test for a node that subscribes after startup, covering the cross-daemon barrier path (`PendingNodes::external_nodes`) that no existing single-daemon test exercises. I checked the load-bearing fixture details: `stop-delay-node` (built by `ensure_built`) does write `DORA_TEST_READY_FILE` immediately after `init_from_env()` returns (`tests/fault_tolerance/stop_delay_node/src/main.rs`), so the ready-file assertion genuinely gates on the subscribe being answered rather than on mere process spawn — the reasoning about `dora node list` reporting `Running` for a still-parked node is correct. Every wait is bounded, so a regression fails the test instead of wedging the CI job. The `nightly.yml` job and its `ci-nightly-jobs.sh` local mirror are kept in lockstep as the repo requires.

---

---

heyong4725:
Review findings for the multi-daemon nightly coverage:

1. **P1 — Parse exact daemon IDs before starting the dataflow** (`tests/multi-daemon-e2e.rs:239`)

   The readiness predicate searches all `dora doctor` output for the characters `A` and `B`. `PASS` supplies `A`, so one registered daemon named B satisfies the predicate. I reproduced this locally: the test proceeded while daemon A was still reconnecting and then failed with `no matching daemon for machine id A`. Parse the connected-machine entries and require exact IDs A and B.

2. **P1 — Prove the startup barrier released before adding the late node** (`tests/multi-daemon-e2e.rs:372`)

   Three nodes reporting `Running` does not establish that `init_from_env()` returned or that the cross-daemon barrier released. Daemon metrics mark a live process Running even while it is parked in initialization, as this test itself notes at lines 318-321. The late node can therefore join the initial cohort and let the regression test pass vacuously. Use ready markers from the initial nodes or another exact post-barrier observable.

3. **P1 — Include the new job in the green-nightly closer** (`.github/workflows/nightly.yml:2303`)

   `file-issue-on-failure.needs` includes `multi-daemon-late-subscriber`, but `close-issue-on-success.needs` does not. Its green decision only examines `needs.*.result`, so a failing new job can race with the closer and have the nightly regression issue closed. Add the new job to this dependency list.

4. **P2 — Bound each spawned client command** (`tests/multi-daemon-e2e.rs:111`)

   `Command::output()` waits indefinitely. Commands invoked inside `wait_until`, plus start/add/stop, can therefore exceed their stated deadlines. The GitHub job eventually stops at 20 minutes, while the local fallback can run unbounded. Spawn each command, poll it to a deadline, then kill and reap it on timeout.

5. **P2 — Add the job's failure-log mapping** (`.github/workflows/nightly.yml:2023`)

   `job_log_pattern()` has no mapping for the `multi-daemon late subscriber end-to-end` display name. On concurrent failures, the reporter falls back to a truncated concatenation of every failed job's logs and can attach unrelated output to this job. Add the exact display-name pattern.

6. **P2 — Keep the local parity job portable** (`scripts/qa/ci-nightly-jobs.sh:1539`)

   The workflow job is Ubuntu-only, but the local driver runs this job on every platform. The test writes executable and ready-file paths into double-quoted YAML scalars; ordinary Windows paths such as `C:\Users\...` contain YAML escape sequences and fail to parse. Gate this local job to Linux or serialize the paths safely.

7. **P3 — Reconcile nightly job accounting** (`CLAUDE.md:162`, `scripts/qa/ci-nightly-jobs.sh:4-10`)

   The edited documentation now says 4 example-backed jobs + 18 script-driven jobs + a separately covered memory-pool job while retaining a total of 22; that partition totals 23. The driver header also still says 21 total and 17 script-driven. Re-audit and update both descriptions together.

Validation: `git diff --check` and `bash -n scripts/qa/ci-nightly-jobs.sh` pass. The targeted E2E was run locally and failed at the daemon-readiness race described in finding 1.

heyong4725:
All seven fixed in `97bc901f9`. The two P1s were the difference between a test that guards the path and one that only looks like it does — thanks for running it rather than reading it.

**1 (P1) — readiness matched `A` inside `PASS`.** Exactly right, and a substring scan of human-readable output was the wrong instrument to begin with. Both machines are now confirmed over the control protocol by exact id (`ControlRequest::ConnectedMachines` + `daemon_id.matches_machine_id`), so the start cannot race a still-connecting daemon.

**2 (P1) — three nodes `Running` did not prove the barrier released.** This is the sharper of the two: it is the same trap the test's own assertion exists to avoid, and I walked into it in the setup phase. If the late node joins the initial cohort, the test passes without exercising the late path at all — the exact vacuity the issue was filed to prevent. The gate is now the **coordinator** marking the dataflow `Running`, which it does on `AllNodesReady`: that transition *is* the barrier release, so "released, then subscribe" is now ordered by construction rather than by hope.

**3 (P1) — closer `needs`.** Added. A failing new job racing the closer and closing the nightly regression issue is a nasty failure mode, since the next green run then reports nothing.

**4 (P2) — unbounded client commands.** Each `dora` invocation is now spawned and polled to a 120s deadline, then killed and reaped. Output goes to files rather than pipes, since reading a pipe to EOF is itself an unbounded wait — bounding `try_wait` while blocking on a pipe would have been theatre.

**5 (P2) — `job_log_pattern`.** Added the exact display-name pattern.

**6 (P2) — local parity portability.** The driver job is now gated to Linux, matching the Ubuntu-only workflow job — keeping the platforms matched is the point of that script. Descriptor paths are also single-quoted YAML scalars now, so a Windows path's backslashes stay literal rather than being read as escapes.

**7 (P3) — job accounting.** Re-audited from the workflow instead of incrementing: **26** test jobs = 4 example-backed + 20 script-driven + `memory-pool-smoke` and `hub-smoke`, which have no local driver entry. Both `CLAUDE.md` and the driver header were stale before this PR (they predated `hub-smoke` and the two `ros2-zenoh-*` jobs) and now agree with each other and with reality.

**Re-verified anti-vacuity after the gating changes**, since tightening the setup could have moved the node add to before the release: with `PendingNodes::external_ready` neutralized the test still fails with *"the node added after startup never got past `init_from_env()`"*, and passes with the latch restored. `make qa-fast` clean, fmt clean, test green in ~55s.

phil-opp:

Re-reviewed the four commits added since the last automated review (through `79cc8d0`). These addressed a reviewer's seven findings, and the current diff looks sound:

- The daemon-readiness gate no longer substring-matches `dora doctor` text (where `PASS` supplied an `A`); both machines are now confirmed by exact id over the control protocol (`ConnectedMachines` + `matches_machine_id`, `tests/multi-daemon-e2e.rs:352-369`).
- The "barrier released before the late node is added" ordering is now proven by per-machine `gate` nodes writing a ready file the instant `init_from_env()` returns, rather than by a `List` poll that reports `Running` before `AllNodesReady` (`tests/multi-daemon-e2e.rs:498-503`).
- Every client command and control round trip is bounded (`COMMAND_TIMEOUT`, `REQUEST_TIMEOUT`), so a regression (a hang) fails the test rather than wedging the job.
- The local driver now records the non-Linux skip in `SKIPPED` instead of letting `run_job` print PASS for a skipped run (`scripts/qa/ci-nightly-jobs.sh:1543-1548`), and the new job is wired into both `file-issue-on-failure.needs` and `close-issue-on-success.needs` plus the `job_log_pattern` mapping.

This is test + CI wiring only, no production code path. No issues found in the new commits.

---

heyong4725:
Rebased onto #3010 as a delta: the test file is dropped (#3010's is equivalent, and gates on both nodes' pid files before the late add, same ordering), leaving only the nightly CI slot that runs it, the Linux-gated local driver entry, and the job-count reconciliation. Title and description updated. Verified with the job's exact command against main's test: 48s, green.

phil-opp:
Heads-up: #3010 landed as `7f9b0b784` covering the same subscribe-after-startup multi-daemon case. This looks superseded — closing unless it covers something #3010 missed.

PR #3009, opened 2026-08-04
fix(qa): unwrap-budget.sh silently passes on Windows checkouts with CRLF .unwrap-budget

### LINKED ISSUE #3008: scripts/qa/unwrap-budget.sh silently passes (exit 0) regardless of budget on any Windows checkout with core.autocrlf=true
**Describe the bug**
`scripts/qa/unwrap-budget.sh` (part of the mandatory `make qa-fast` pre-commit gate documented in CLAUDE.md/AGENTS.md) reads the numeric budget from `.unwrap-budget` via:

    BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \n')

This strips spaces and LF but not CR. On any Windows checkout with `core.autocrlf=true` (the Git-for-Windows installer's recommended default option), `.unwrap-budget` gets CRLF line endings, so BUDGET becomes "163\r" instead of "163". The subsequent

    if [[ "$CURRENT" -gt "$BUDGET" ]]; then ...
    if [[ "$CURRENT" -lt "$BUDGET" ]]; then ...

then throw `syntax error: invalid arithmetic operator` — and because these are `if` conditions (exempt from `set -e`), the error is swallowed and the script exits 0 regardless of the actual comparison. This isn't just noisy output but a genuine silent gate bypass: with a CRLF-terminated budget file set to "1" and the real unwrap/expect count at 158 (158 >> 1), the unpatched script still exits 0.

Verified three ways:
1. Live repo checkout with `core.autocrlf=true` (default on this machine).
2. A completely fresh `git clone --depth 1` with `-c core.autocrlf=true` explicitly set — same CRLF bytes, same crash.
3. Adversarial test: budget=1 (CRLF) vs actual count=158 → original script exits 0 (silent false pass).

**To Reproduce**
1. On Windows (or any machine with `git config core.autocrlf true`), clone dora-rs/dora
2. `xxd .unwrap-budget`  — confirm trailing `0d 0a` (CRLF)
3. `bash scripts/qa/unwrap-budget.sh`
4. Observe: two "syntax error: invalid arithmetic operator" lines printed, but the script's own exit code is 0 even when the count exceeds budget

**Expected behavior**
The unwrap-budget ratchet should fail loudly (or at minimum, exit non-zero) whenever the actual unwrap/expect count exceeds the committed budget — regardless of the checkout's line-ending configuration.

**Environments (please complete the following information):**
- System info: Windows 11, any Git-for-Windows install with the installer's recommended `core.autocrlf=true`
- Dora version: main @ `4612ca86af6f783c235d5ed0fb0ce28d3d203178`

**Additional context**
Fix (verified — patched script now correctly exits 0 when under budget and exits 1 with a clear message when over budget, even with CRLF-terminated `.unwrap-budget`):

```diff
-BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \n')
+BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \r\n')

GuTS805:
## Problem
`scripts/qa/unwrap-budget.sh` — part of the mandatory `make qa-fast` pre-commit gate documented in CLAUDE.md/AGENTS.md — strips spaces/LF from `.unwrap-budget` but not CR:

    BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \n')

On any checkout with `core.autocrlf=true` (the Git-for-Windows installer's recommended default), `.unwrap-budget` gets CRLF line endings, so `BUDGET` becomes `"163\r"`. The subsequent

    if [[ "$CURRENT" -gt "$BUDGET" ]]; then ...
    if [[ "$CURRENT" -lt "$BUDGET" ]]; then ...

then throw a bash arithmetic syntax error that — being inside an `if` condition, exempt from `set -e` — is silently swallowed, and the script exits 0 regardless of whether the budget is actually exceeded.

Verified this is a genuine silent gate bypass, not just noisy output: with a CRLF-terminated budget file set to `1` and the real unwrap/expect count at 158 (158 >> 1), the unpatched script still exits 0.

## Fix
Strip `\r` too:

    -BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \n')
    +BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \r\n')

## Validation
- `bash scripts/qa/unwrap-budget.sh` now passes cleanly on a CRLF-terminated `.unwrap-budget` (verified on a fresh checkout with `core.autocrlf=true`)
- Confirmed the fix still correctly fails (exit 1, clear message) when the budget is genuinely exceeded, even with a CRLF-terminated `.unwrap-budget`
- Checked sibling QA scripts (`typos.sh`, `audit.sh`, `semver.sh`) for the same class of bug — none read a checked-in single-value file the same way, so this appears isolated to `unwrap-budget.sh`

**Class A** (mechanical one-line shell fix, no behavior change beyond correcting the bug) — `make qa-fast` is sufficient.

Not a duplicate of prior `unwrap-budget.sh` fixes (#2002 / #2012 / #2261 — those addressed awk truncation and test-code miscounting, not line endings).

Fixes #3008

### DIFF
--- a/scripts/qa/unwrap-budget.sh
+++ b/scripts/qa/unwrap-budget.sh
@@ -120,7 +120,7 @@ fi
 
 CURRENT=$(count_unwraps)
 
-BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \n')
+BUDGET=$(cat "$BUDGET_FILE" | tr -d ' \r\n')
 
 echo "unwrap/expect count: $CURRENT (budget: $BUDGET)"
 


phil-opp:
> 🤖 **Automated review by Claude — fully automated, no human in the loop.**

No issues found. `tr -d ' \r\n'` correctly extends the strip set to include CR: `tr` interprets the `\r`/`\n` escapes in the single-quoted set, so a CRLF `.unwrap-budget` now yields a clean integer instead of `163\r`, which was tripping the swallowed bash arithmetic error and letting the check exit 0 regardless. Escape handling for `\r` is portable across GNU/BSD/macOS `tr`. The fix is complete and the commit message's reproduction matches the observed behavior.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3010, opened 2026-08-04
test(e2e): cover subscribe-after-startup on a multi-daemon dataflow (#2938)

### LINKED ISSUE #2938: daemon: on multi-daemon dataflows, any subscribe after startup is never answered — `dora node add`, dynamic nodes and restarts hang in `Node()` init
Found while reviewing the fix for #2917. Pre-existing, unrelated to that fix, and read from the code rather than reproduced — I have no multi-machine setup to hand, so the severity claim below deserves confirmation by someone who does.

## The path

`PendingNodes::update_dataflow_status` (`binaries/daemon/src/pending.rs`) has two branches once `local_nodes` is empty:

```rust
if self.local_nodes.is_empty() {
    if self.external_nodes {
        if !self.reported_init_to_coordinator {
            self.report_nodes_ready(...).await?;      // one-shot
            self.reported_init_to_coordinator = true;
        }
        Ok(DataflowStatus::Pending)                   // <-- waiter parked, NOT answered
    } else {
        self.answer_subscribe_requests(Vec::new(), cascading_errors).await;
        Ok(DataflowStatus::AllNodesReady)
    }
}
```

In the local-only case the subscriber is answered inline. In the external case it is parked in `waiting_subscribers` and the only other thing that ever answers it is `handle_external_all_nodes_ready` (`binaries/daemon/src/lib.rs:1919`), driven by `DaemonCoordinatorEvent::AllNodesReady`.

The coordinator sends that exactly once per dataflow: `broadcast_all_nodes_ready` (`binaries/coordinator/src/lib.rs:4363`) is called from `ReadyOnDaemon` when `dataflow.pending_daemons` becomes empty (`:459-460`), or from the disconnect path when the last pending daemon goes away (`:3280`, #2028). `pending_daemons` stays empty afterwards, so it never fires again.

`external_nodes` is set whenever any descriptor node is assigned to another daemon (`binaries/daemon/src/lib.rs:3501`, the non-local branch) — i.e. exactly the multi-machine case.

## Consequence

On a multi-daemon dataflow, any node that subscribes **after** that one-shot has fired parks forever. Its `Node()` / `init_from_env()` never returns. That covers at least:

1. **`dora node add`** — a runtime addition subscribes long after startup, so it hangs at init. This is the deployment mode where dynamic topology is most useful.
2. **Dynamic nodes (`path: dynamic`)** — these are *designed* to connect on their own schedule. One that connects after startup hangs.
3. **Node restarts** — a restarted node re-subscribes (`binaries/daemon/src/lib.rs`, the restart path re-runs subscribe), so any restart on a distributed dataflow would hang the restarted node rather than recovering it.

Single-machine dataflows are unaffected: `external_nodes` stays false and every subscribe is answered inline. That is consistent with this never having been reported — the local path is the one everyone exercises.

## Suggested direction

The external barrier needs a "already resolved" latch, not just a one-shot event. Something like an `external_ready: bool` set in `handle_external_all_nodes_ready`, with the external branch of `update_dataflow_status` answering immediately when it is already true:

```rust
if self.external_nodes {
    if self.external_ready {
        self.answer_subscribe_requests(Vec::new(), cascading_errors).await;
        return Ok(DataflowStatus::AllNodesReady);
    }
    ...
}
```

That mirrors what the local branch already does and keeps the startup semantics intact — the barrier still holds the initial cohort until every daemon reports ready.

Worth pairing with a multi-daemon e2e that adds a node to a running distributed dataflow, since nothing in the current suite covers subscribe-after-startup on the external path. `tests/node-lifecycle-e2e.rs` is single-daemon throughout.

## Notes

The `#2917` fix (#2933) documents this next to `PendingNodes::cohort` but deliberately does not change it: that fix scopes *which* nodes inherit a startup failure, and this is about waiters never being answered at all. Different failure, different change.

phil-opp:
Follow-up to #2990, which fixed #2938 but could only be verified by unit test — its own "known limit" section asks for exactly this.

A real coordinator plus daemons `A` and `B` on loopback (all ephemeral ports), a dataflow with a node on each machine, then `dora node add` against the running distributed dataflow. The added node records its pid only after `init_from_env()` returns, so the assertion is exactly "the late subscribe was answered": on the parent of #2990 the pid never appears and the test fails, with it the test passes.

No cross-daemon data delivery is involved — the barrier travels over the coordinator's WebSocket and the added node talks to its own local daemon — so it does not depend on runner multicast, which is what let it run reliably here. Reuses the existing `reconnect-survivor-node` fixture rather than adding one. Wired into the existing E2E job in `ci.yml` (~60s).

While reproducing the original hang I also confirmed the other two symptoms from #2938 by hand on pre-#2990 binaries: a `path: dynamic` node sat in `init_from_node_id` until killed, and streamed data across daemons once the latch was in place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -376,6 +376,13 @@ jobs:
         run: >
           cargo test -p dora-examples --test daemon-reconnect-e2e -- --test-threads=1
         timeout-minutes: 15
+      # Regression for #2938: a node added to a running *distributed*
+      # dataflow used to hang forever in `Node()` init. Spawns a real
+      # coordinator + two named daemons. Same `dora-examples` exclusion.
+      - name: Run multi-daemon late-join E2E
+        run: >
+          cargo test -p dora-examples --test multi-daemon-e2e -- --test-threads=1
+        timeout-minutes: 10
       # `dora node` lifecycle E2E (issue #1703). The shared test
       # binary lives under the `dora-examples` package which the bulk
       # `cargo test --all` step excludes, so it needs an explicit
--- a/tests/multi-daemon-e2e.rs
+++ b/tests/multi-daemon-e2e.rs
@@ -1,0 +1,361 @@
+//! End-to-end regression test for dora-rs/dora#2938.
+//!
+//! On a multi-daemon dataflow the start barrier is resolved by the
+//! coordinator, which broadcasts `AllNodesReady` at most once per
+//! dataflow. The daemon used to treat that broadcast as the *only* thing
+//! that ever answers a parked subscribe request, so every node that
+//! subscribed afterwards — a runtime `dora node add`, a dynamic node,
+//! a restarted node — parked forever and never returned from `Node()`
+//! init. Single-machine dataflows were unaffected (they answer each
+//! subscribe inline), which is why this went unnoticed.
+//!
+//! The latch that fixed it landed in #2990, verified by unit test only —
+//! nothing in the suite covered subscribe-after-startup on the external
+//! path, and `tests/node-lifecycle-e2e.rs` is single-daemon throughout.
+//! This is that missing coverage: it fails on the parent of #2990 and
+//! passes with it.
+//!
+//! The test runs a real two-daemon deployment (coordinator + daemons `A`
+//! and `B` on loopback, all on ephemeral ports), starts a dataflow with
+//! a node on each machine, and then adds a node at runtime. The added
+//! node records its pid *after* `init_from_env()` returns, so that pid
+//! appearing is proof that its subscribe was answered.
+//!
+//! No cross-daemon *data* delivery is needed: the barrier travels over
+//! the coordinator's WebSocket, and the added node talks to its own
+//! local daemon. That keeps the test independent of the multicast/zenoh
+//! peering quirks of CI runners and dev containers.
+//!
+//! Heavyweight (spawns processes, ~30-60s). It lives in the
+//! `dora-examples` crate, which the fast PR `cargo test --all` step
+//! excludes; run it explicitly with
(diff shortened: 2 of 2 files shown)


phil-opp:

No issues found.

I checked the new `tests/multi-daemon-e2e.rs` against the code it drives:

- The `reconnect-survivor-node` fixture writes its pid to `$DORA_TEST_PID_FILE` **after** `init_from_env()` returns (`tests/fault_tolerance/reconnect_survivor_node/src/main.rs`), so a recorded pid genuinely proves the late subscribe was answered — the assertion is exactly what the PR claims.
- All CLI surfaces the test uses exist and match: `dora daemon --machine-id/--local-listen-port/--coordinator-port/--zenoh-peer`, `dora coordinator --port/--store redb:…`, `dora node add --from-yaml --dataflow`, and `dora destroy`.
- `WsSession::connect`/`.request`, `ControlRequest::ConnectedMachines`, and `ControlRequestReply::ConnectedDaemons(Vec<DaemonInfo>)` all exist as used; `daemon_id.machine_id()` is the right accessor.
- The failure mode is correct: on buggy (pre-#2990) binaries the added node parks in init, `pid_late` never appears, and the test fails at the final `wait_until`; the two descriptor nodes coming up first establishes the "AllNodesReady already fired" precondition.
- `tests/multi-daemon-e2e.rs` is auto-discovered as a `dora-examples` test target (same as the existing `daemon-reconnect-e2e`, which is not explicitly listed either), so `cargo test -p dora-examples --test multi-daemon-e2e` resolves.

Ephemeral-port selection and the shared-`zenoh-peer` rendezvous match the daemon's documented behavior; no correctness concerns.

---

PR #3011, opened 2026-08-04
fix(python): drain pending memory-pool frees in \`recv_async\`

### LINKED ISSUE #2958: python-node: `recv_async()` skips memory-pool free draining, so async nodes never release freed GPU/shmem resources
`Node.recv_async()` is the only one of the four Python receive methods that
doesn't drain the pending memory-pool free set, so async nodes never release
GPU IPC handles, transit buffers, or shmem mappings for pools other nodes
have freed.

Call sites in `apis/python/node/src/lib.rs` (at `7a15da24`):

```
1259  self.process_pending_memory_pool_frees(py);   // next()
1347  self.process_pending_memory_pool_frees(py);   // drain()
1377  self.process_pending_memory_pool_frees(py);   // try_recv()
      (missing)                                     // recv_async()  :1411
```

`__next__` delegates to `next()`, so it's covered. `PENDING_FREES` is only
emptied by `drain_freed_pools()`, so a loop that is exclusively
`await node.recv_async()` never cleans up. Fails silently the symptom
shows up later as a CUDA OOM elsewhere.

This contradicts the contract in
`apis/rust/node/src/event_stream/memory_pool.rs:1-9`: *"The language binding
drains this set before yielding the next user-visible event."*

Introduced in #2619 (`e99ba212`), which wired three of the four paths;
`recv_async` predates it (#909), so it looks like a missed call site:

```
$ git log --oneline -S "process_pending_memory_pool_frees" -- apis/python/node/src/lib.rs
e99ba212 feat(memory-pool): GPU-to-GPU adaptive transport ... (#2619)
```

### Fix

Drain before the await, so it also runs when `recv_async` returns `None`:

```rust
pub async fn recv_async(&self, timeout: Option<f32>) -> PyResult<Option<Py<PyDict>>> {
    Python::attach(|py| self.process_pending_memory_pool_frees(py));
    let timeout = timeout_to_duration(timeout)?;
    ...
```

The scoped `attach` drops the GIL before the suspend point. Regression test
would fit alongside `test_mock_node_recv_async_loop`.

Distinct from #2935 (daemon never sends the broadcast), #2934/#2940 (wrong
pointer in the cleanup itself), and #2881 (crash/removal path).

phil-opp:
`recv_async` was the one of the Python node's four receive paths that skipped
`process_pending_memory_pool_frees`, so a node whose loop is exclusively
`await node.recv_async()` never released its GPU IPC handles, transit buffers,
or shmem mappings for pools that another node freed. `PENDING_FREES` is only
emptied by `drain_freed_pools()`, so the set just grew — the leak is silent and
surfaces later as a CUDA OOM somewhere unrelated.

`next`/`drain`/`try_recv` were wired up in #2619; `recv_async` predates it
(#909) and was missed. The drain runs before the await so it also covers the
`None` (stream closed) path, and the scoped `attach` drops the GIL before the
suspend point.

Pinned by a guard test over the four receive methods — `Node` needs a live
daemon connection to construct, so the invariant can't be asserted
behaviourally in a unit test.

Fixes #2958

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -1409,6 +1409,13 @@ impl Node {
     #[pyo3(signature = (timeout=None))]
     #[allow(clippy::should_implement_trait)]
     pub async fn recv_async(&self, timeout: Option<f32>) -> PyResult<Option<Py<PyDict>>> {
+        // Same cleanup contract as `next`/`drain`/`try_recv`: release the
+        // per-process resources of pools that other nodes freed before
+        // yielding the next user-visible event.  Done *before* the await so
+        // that it also runs on the `None` (stream closed) path, and so the
+        // scoped `attach` drops the GIL before the suspend point.
+        Python::attach(|py| self.process_pending_memory_pool_frees(py));
+
         let timeout = timeout_to_duration(timeout)?;
         let event = self.events.recv_async_timeout(timeout).await;
         if let Some(event) = event {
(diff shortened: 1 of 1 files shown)


phil-opp:

- The one-line fix (`apis/python/node/src/lib.rs:1418`) correctly runs `process_pending_memory_pool_frees` before the `.await`, so it also fires on the `None`/stream-closed path, and the scoped `Python::attach` releases the GIL before the suspend point — matching the reasoning in the commit.
- It brings `recv_async` in line with the `next`/`drain`/`try_recv` paths, closing the `await`-only leak.
- The guard test genuinely exercises the invariant: it would fail if the drain call were removed from any of the four methods, and the `__next__`-delegates-to-`next` escape hatch is explicitly checked. Asserting on source is a fair compromise given `Node` requires a live daemon connection to construct.

Minor, non-blocking: `method_body`'s brace matching is naive (it would miscount a `{`/`}` inside a string or char literal in one of those method bodies). It's self-guarding — a miscount fails the test loudly rather than passing silently — so it can't hide a real regression, but a future edit that introduces a braced literal into one of these bodies could produce a confusing failure.

---

Issue #2935, opened 2026-08-01
memory-pool: cross-process `FreeMemoryPool` cleanup can be silently skipped (full-channel drop + read-with-free path)


phil-opp:
> **This issue was created by a scheduled, automated Claude code-review check** that looks at recently merged PRs. The findings were verified against the source, but please double-check before acting on them.

## Summary

**PR #2619** (`feat(memory-pool): GPU-to-GPU adaptive transport…`, commit `e99ba21`) added cross-process cleanup: when a node calls `free_memory_pool`, the daemon broadcasts `NodeEvent::FreeMemoryPool` to every other node that registered or read the pool, and each recipient releases its per-process GPU/host/IPC/shmem resources.

Two paths defeat that guarantee and reintroduce the per-process leak the feature is meant to prevent.

## Bug 1 — broadcast is silently dropped when the receiver's event channel is full

`binaries/daemon/src/lib.rs`, `FreePinnedMemory` handler (~line 4025):

```rust
for (node, channel) in &dataflow.subscribe_channels {
    if touched.contains(node.as_ref())
        && node.as_ref() != node_id.as_ref()
        && let Err(e) = send_with_timestamp(channel, event.clone(), &self.clock)
    {
        tracing::warn!(/* failed to deliver FreeMemoryPool */);
    }
}
```

`send_with_timestamp` (`binaries/daemon/src/event_types.rs:213`) returns **`Ok(false)`** — not `Err` — when the channel is full (it `try_send`s and logs *"CRITICAL: control event dropped despite headroom reservation"*). The loop above only inspects the `Err` (channel-closed) case, so an `Ok(false)` drop is neither retried nor surfaced by this site. When it happens, the recipient never runs `process_pending_memory_pool_frees`, and its GPU buffer / pinned+transit host buffer / IPC handle / shmem mapping for that pool leak for the life of the process.

Note the sibling `send_reload` loop (~line 4060) explicitly handles the `Ok(false)` "event dropped (channel full)" case; this new site does not.

**Suggested fix:** match on the `bool`, e.g. warn (and/or retry) on `Ok(false)` as well as `Err`.

## Bug 2 — the read-with-free path frees the pool but never broadcasts

`binaries/daemon/src/lib.rs`, `ReadPinnedMemory { free: true }` handler (~line 3972):

```rust
if free
    && let Err(err) = self.memory_pool.free_memory_pool(&id, node_id.as_ref())
{
[...]

PR #3012, opened 2026-08-04
fix(daemon): broadcast memory-pool cleanup from both free paths (#2935)

phil-opp:
`free_memory_pool` tells every other node that touched a pool to drop its
GPU buffer, pinned host buffer, IPC handle and shmem mapping. Two paths
skipped that notification:

- `ReadPinnedMemory { free: true }` freed the pool without notifying
  anyone, so every other node held its per-process resources for life —
  the leak the broadcast exists to prevent, via the other free entry point.
- `send_with_timestamp` reports a full channel as `Ok(false)`, not `Err`.
  The broadcast matched only on `Err`, so a backed-up receiver silently
  lost its cleanup notification.

Both entry points now go through `free_pool_and_notify`. A full channel is
reported and logged with the affected nodes instead of being swallowed; a
closed channel stays benign (the node is gone, so the OS reclaimed its
resources). Delivery is not retried — this runs on the daemon main loop
with the freeing node blocked on its reply, so awaiting a backed-up
receiver would stall every dataflow on the daemon.

Also pairs each enqueue with `inc_pending`. The listener decrements
`pending_messages` for every event it drains, so the broadcast was
underflowing the reported count to `u64::MAX` (#2827).

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -41,7 +41,7 @@ use local_listener::DynamicNodeEventWrapper;
 use log::{CoordinatorLogTarget, DaemonLogger, DataflowLogger, Logger};
 use spawn::Spawner;
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
     env::current_dir,
     future::Future,
     io,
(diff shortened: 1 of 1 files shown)


phil-opp:

Checks:
- Both free paths now funnel through `free_pool_and_notify` (`binaries/daemon/src/lib.rs:4786`, `:4829`), closing the `ReadPinnedMemory { free: true }` leak where the pool was freed without notifying touched nodes.
- The `send_with_timestamp` result is handled on all three arms: `Ok(true)`→deliver+`inc_pending`, `Ok(false)`→report as undelivered (the previously-swallowed full-channel case), `Err`→benign skip (node gone). The broadcast is not aborted by one full or closed channel.
- The `inc_pending` pairing is correct rather than double-counting: `notify_pool_freed_reports_nodes_with_full_channels` fills a channel via `send_with_timestamp` directly and asserts the pending counter stays `0`, proving `send_with_timestamp` doesn't itself increment — so the single `inc_pending` on `Ok(true)` is the sole matching increment for the listener's unconditional decrement (#2827).
- `free_pool_and_notify` propagates the manager's free error before any notification and treats a `None` dataflow as "release the entry, nobody to notify" — both covered by dedicated tests.
- Initiator and untouched-subscriber exclusion are asserted directly.

The one behavioral nuance worth a maintainer's awareness (not a defect): a node in `touched` but absent from `subscribe_channels` is silently not notified, but that matches the pre-existing iteration model and such a node isn't reachable via the daemon event path anyway.

---

PR #3013, opened 2026-08-04
fix(coordinator): replay the ready barrier to a daemon that reconnects after it released

### LINKED ISSUE #2998: Coordinator: a daemon that reconnects after the ready-barrier released never learns of it — its nodes hang in `init_from_env()`

heyong4725:
Fixes #2998.

## The bug

A daemon that disconnects while still pending is dropped from **both**
`daemons` and `pending_daemons`, so the `AllNodesReady` broadcast — which
iterates `daemons` — correctly skips it. On reconnect,
`reestablish_running_dataflow` relinked it but replayed nothing, and the
daemon never re-reports readiness (`reported_init_to_coordinator` is not
reset). Its `PendingNodes::external_ready` stays `None` for the life of the
dataflow and every node it owns parks in `init_from_env()` forever.

The #2938 latch cannot help: it converts "the broadcast fired before I
subscribed" into an answer, but this daemon never received the broadcast at all.

## The fix

Coordinator-side option from the issue — remember that the barrier released,
and re-send it to just that daemon on reconnect.

- `RunningDataflow::ready_barrier_released` records the release. In-memory
  only: after a coordinator restart `recovered` rebuilds with an empty
  `pending_daemons`, so the next `ReadyOnDaemon` broadcasts to everyone anyway.
- `release_barrier_message` builds the frame **and** sets the flag, so a
  broadcast that failed to record itself is a test failure rather than a
  convention.
- `all_nodes_ready_message` is shared by broadcast and replay so a replay
  cannot drift from the original — in particular it carries the same
  `exited_before_subscribe`. A blank success there would start a dataflow the
  coordinator already declared failed, the daemon-side trap #2990 hit.
- `reestablish_running_dataflow` is now `#[must_use] -> bool`, reporting
  whether the reconnecting daemon is owed a replay; wired at both the
  `Recovering` and `Running` reconnect paths.

The replay is **best-effort** (warn, not `?`). It runs inside `start_inner`,
the main event loop, where propagating would take down the coordinator and
every other dataflow because one just-reconnected daemon socket send failed —
which is exactly the flaky-link scenario this fix targets.

## Duplicate delivery is safe

A daemon connected at broadcast time, disconnected after, then reconnecting
receives a second `AllNodesReady`. That path calls `dataflow.start()`
(`daemon/src/lib.rs:2114`) with no `dataflow_started` guard. Verified benign:
`start()` skips intervals already in `_timer_handles`
(`running_dataflow.rs:456`) and `answer_subscribe_requests` drains
`waiting_subscribers` via `std::mem::take`, so the second call is a no-op.

## Tests

- `reconnecting_daemon_is_owed_a_barrier_replay_only_after_release` — the
  obligation is false before release, true after
- `broadcasting_the_barrier_records_it_even_with_no_daemon_reachable` — drives
  the real broadcast with zero connections, the #2998 shape exactly
- `releasing_the_barrier_records_it_on_the_dataflow` — the mark/message pairing
- `replayed_barrier_carries_the_same_failure_verdict_as_the_broadcast` — a
  replay must not carry a blank success

Mutation-verified: reverting the obligation, dropping the mark, blanking the
verdict, and swapping the marking builder for the plain one each fail a test.

**Coverage limit, stated plainly:** the socket send itself is not asserted —
there is no fake daemon connection, so no-oping `replay_all_nodes_ready` still
passes the suite. Multi-daemon reconnect coverage is tracked in #2999.

## Validation

- `cargo fmt --check`, `cargo clippy --all -D warnings`, `cargo check --examples` — clean
- Full workspace suite — green
- Linux CI dispatched on the branch: all 11 jobs green, including
  `Test (ubuntu-latest)`, `E2E Tests`, and `Semantic contract tests`
  (run 30885613740)

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -656,6 +656,7 @@ pub(crate) async fn start_dataflow(
             BTreeSet::new()
         },
         exited_before_subscribe: Default::default(),
+        ready_barrier_released: false,
         daemons: daemons.clone(),
         nodes,
         node_to_daemon,
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -6018,7 +6018,12 @@ mod tests {
         assert_eq!(summary.attempted, 1);
         assert_eq!(summary.failed, 0);
 
-        daemon_task.await.unwrap();
+        // Bounded: if the coordinator stops sending, this must fail fast
+        // rather than block forever and burn the CI job timeout.
+        tokio::time::timeout(Duration::from_secs(10), daemon_task)
+            .await
+            .expect("daemon task did not receive the expected message")
+            .unwrap();
     }
 
     #[tokio::test]
@@ -6106,7 +6111,12 @@ mod tests {
         assert_eq!(summary.attempted, 1);
         assert_eq!(summary.failed, 1);
         assert!(summary.failed > 0);
-        daemon_task.await.unwrap();
+        // Bounded: if the coordinator stops sending, this must fail fast
+        // rather than block forever and burn the CI job timeout.
+        tokio::time::timeout(Duration::from_secs(10), daemon_task)
+            .await
+            .expect("daemon task did not receive the expected message")
+            .unwrap();
     }
 
     #[tokio::test]
@@ -6241,7 +6251,12 @@ mod tests {
             Arc::new(HLC::default()),
         );
 
-        daemon_task.await.unwrap();
+        // Bounded: if the coordinator stops sending, this must fail fast
+        // rather than block forever and burn the CI job timeout.
+        tokio::time::timeout(Duration::from_secs(10), daemon_task)
+            .await
+            .expect("daemon task did not receive the expected message")
+            .unwrap();
     }
 
     #[tokio::test]
@@ -6319,7 +6334,12 @@ mod tests {
         .await
         .expect("subscription should succeed");
 
-        daemon_task.await.unwrap();
+        // Bounded: if the coordinator stops sending, this must fail fast
+        // rather than block forever and burn the CI job timeout.
+        tokio::time::timeout(Duration::from_secs(10), daemon_task)
+            .await
+            .expect("daemon task did not receive the expected message")
+            .unwrap();
         assert_eq!(Some(subscription_id), *seen_subscription.lock().await);
         assert_eq!(
             running_dataflows[&dataflow_id]
@@ -6524,7 +6544,12 @@ mod tests {
         )
         .await;
 
-        daemon_task.await.unwrap();
+        // Bounded: if the coordinator stops sending, this must fail fast
+        // rather than block forever and burn the CI job timeout.
+        tokio::time::timeout(Duration::from_secs(10), daemon_task)
+            .await
+            .expect("daemon task did not receive the expected message")
+            .unwrap();
         assert_eq!(
             *seen.lock().await,
             Some((subscription_id, dataflow_id)),
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -3256,7 +3256,10 @@ fn reestablish_running_dataflow(
         "re-established running dataflow {} in live coordinator state after daemon {daemon_id} reconnect",
         record.uuid
     );
-    false
+    // The rebuilt entry carries the persisted release, so this path owes the
+    // replay exactly as the relink path does. Returning a flat `false` here is
+    // what left orphan-reclaim and coordinator-restart reconnects hanging.
+    record.ready_barrier_released
 }
 
 /// Tell a single daemon to stop a dataflow the coordinator has terminally given
@@ -4605,6 +4608,20 @@ async fn broadcast_all_nodes_ready(
     tracing::debug!("sending all nodes ready message to daemons");
     let message = release_barrier_message(uuid, dataflow, clock)?;
 
+    // Persist the release. The in-memory entry is destroyed by orphan reclaim
+    // and by coordinator restart; without a durable record of it, a daemon that
+    // missed this broadcast and reconnects afterwards is never replayed it and
+    // hangs for the life of the dataflow (#2998). Only on a *successful*
+    // barrier: a failed one is persisted by the failure path with its verdict,
+    // and must not be promoted to `Running` here.
+    if dataflow.exited_before_subscribe.is_empty()
+        && let Err(e) = dataflow
+            .make_record(StoreDataflowStatus::Running)
+            .and_then(|r| store.put_dataflow(&r))
+    {
+        tracing::warn!(dataflow = %uuid, "failed to persist ready-barrier release: {e}");
+    }
+
     // notify all machines that run parts of the dataflow
     for daemon_id in &dataflow.daemons {
         let Some(connection) = daemon_connections.get_mut(daemon_id) else {
@@ -5008,6 +5025,8 @@ mod tests {
             daemon_ids: vec![daemon_id.clone()],
             node_to_daemon: BTreeMap::new(),
             uv: false,
+            ready_barrier_released: false,
+            barrier_exited_before_subscribe: Vec::new(),
             generation: 3,
             created_at: 7,
             updated_at: 7,
@@ -5075,6 +5094,8 @@ mod tests {
             daemon_ids: vec![daemon_id.clone()],
             node_to_daemon: BTreeMap::new(),
             uv: false,
+            ready_barrier_released: false,
+            barrier_exited_before_subscribe: Vec::new(),
             generation: 3,
             created_at: 7,
             updated_at: 7,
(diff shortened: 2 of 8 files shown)


phil-opp:

What I verified:
- The `ready_barrier_released` flag is set inside `release_barrier_message` (`binaries/coordinator/src/lib.rs:4557`), which is the sole builder used by the real broadcast, so a broadcast can never fire without recording itself — `broadcasting_the_barrier_records_it_even_with_no_daemon_reachable` locks that in for the exact #2998 shape (no reachable daemon at broadcast time).
- The replay reuses `all_nodes_ready_message`, so it carries the same `exited_before_subscribe`; the `replayed_barrier_carries_the_same_failure_verdict_as_the_broadcast` test meaningfully guards against a blank-success replay that would start a dataflow already declared failed.
- `reestablish_running_dataflow` returns `ready_barrier_released` only in the relink branch and `false` when reconstructing from the record — consistent with the documented in-memory-only design (a reconstructed entry legitimately has no barrier history and relies on a fresh broadcast).
- Borrow flow across the `.await` is fine: the `&mut` from `reestablish_running_dataflow` ends before the immutable `get`, and `daemon_connections`/`clock` are borrowed independently. The `get_mut` change in `apply_disconnect_actions` is the necessary consequence of `broadcast_all_nodes_ready` now taking `&mut`.
- Replay targets only the reconnecting `daemon_id`, and a missing connection is handled as a warn+`Ok` rather than failing the whole reconciliation.

Consistent with the commit's own note, the actual socket send is not asserted (no fake daemon connection) and is deferred to #2999 — that's a coverage gap, not a defect in this diff.

---

heyong4725:
Confirmed, and fixed in 8732c00d1. The finding was right on every link:

- `PendingNodes::reported_init_to_coordinator` (`binaries/daemon/src/pending.rs:262`) is set once and never reset, so the daemon genuinely cannot prompt a fresh broadcast.
- `begin_orphaned_dataflow_reclaim` destroys the live entry whenever *every* daemon running the dataflow disconnects, so this is reachable in an ordinary multi-daemon dataflow, not only on coordinator restart.
- `recovered` hardcoded `ready_barrier_released: false`, so the reconstruct path reported no replay owed.

My doc comment claiming "the next `ReadyOnDaemon` broadcasts to every daemon anyway" was simply wrong, for exactly that first reason. That was worse than the gap itself, since it asserted coverage that did not exist.

**Fix:** the release is now persisted and restored.

- `DataflowRecord` gains `ready_barrier_released` and `barrier_exited_before_subscribe`.
- `make_record` mirrors both. It is the single snapshot point every store write goes through, so there is no second write to drift.
- `recovered` restores both; `reestablish_running_dataflow` returns the restored value on the reconstruct path instead of a flat `false`.
- `broadcast_all_nodes_ready` persists at release, which was necessary: the existing `Running` persist fires on spawn success, *before* the barrier releases, so a restart would otherwise still lose it. Guarded to successful barriers so a failed one is not promoted to `Running`.

The verdict had to travel with the flag. Restoring the release but blanking `exited_before_subscribe` would replay a bare success and start a dataflow the coordinator had already given up on.

**One thing the review did not reach, worth flagging.** The redb backend encodes with **bincode**, which is positional, so `#[serde(default)]` does nothing for it — the added fields shift the byte layout and an existing database would misdecode into structurally valid but wrong records. `SCHEMA_VERSION` is bumped 3 → 4; `open()` rejects a mismatch, the same remedy #1854 used. Without that bump this change would have corrupted live redb state on upgrade.

**Coverage.** Three new tests: the persist/rebuild round trip, the reconstruct path owing the replay, and the reconstructed failure verdict. Mutation-verified — dropping the flag or the verdict on either side of the round trip, and flattening the reconstruct return, each fail a test. Two of my earlier tests turned out to prove nothing (they asserted on `reestablish`'s return, which reads the record directly and bypasses both `make_record` and `recovered`); the round-trip test closes that.

Still true, and unchanged from the PR body: the socket send itself is not asserted, since there is no fake daemon connection. Multi-daemon reconnect coverage is tracked in #2999.

Local: fmt, `clippy --all -D warnings`, and `clippy -p dora-coordinator-store --features redb-backend` all clean; 117 coordinator + 37 redb-backend tests green, including the four schema-rejection tests. Full Linux CI dispatched on the branch.

heyong4725 (commented):
Two blocking findings remain on the current head (`8732c00d1`):

1. **[P1] Persist a failed barrier at the release boundary** (`binaries/coordinator/src/lib.rs:4611-4623`). The new durable latch is written only when `exited_before_subscribe` is empty. For a failed barrier, persistence is deferred to the later daemon-finish path. That leaves a real crash window: daemon B is disconnected, daemon A reports an exited-before-subscribe node, the coordinator broadcasts the failed barrier to the surviving set, then the coordinator restarts before A's finish event is stored. The record still says `ready_barrier_released = false` with an empty verdict. When B reconnects, the reconstruct path returns `false`, so B never receives the failed release and its pending nodes hang in `init_from_env()`—the same #2998 failure this PR is intended to make restart-safe. The release flag and verdict need to be stored regardless of success, without incorrectly promoting the status; a regression test should read the store immediately after a failed `broadcast_all_nodes_ready` and rebuild from that record.

2. **[P1] Replay the barrier's persisted-parameter side effect to the reconnecting daemon** (`binaries/coordinator/src/lib.rs:4574-4594`). The initial release deliberately couples `AllNodesReady` with `schedule_param_replay_for_ready_dataflow`, and a daemon absent from `dataflow.daemons` misses both. The reconnect path sends only `AllNodesReady`, so the daemon starts but never receives persisted desired params. The later state-catch-up loop does not cover the general case: when `last_ack == state_log_sequence == 0` it skips entirely, which includes params already present in the store before this running incarnation/restart. Please add a single-daemon persisted-param replay alongside the targeted barrier replay (or factor the existing scheduler) and cover it with the reconnect test.

Validation on a clean merge with current `origin/main` (`313ca4ccf`):

- Merge: clean, including overlapping coordinator files
- `cargo fmt --all -- --check`: pass
- `cargo test -p dora-coordinator-store --features redb-backend`: 37 passed
- `cargo test -p dora-coordinator`: 145 passed (119 unit, 22 control, 4 daemon)
- `cargo clippy -p dora-coordinator -- -D warnings`: pass

The schema-version bump and the persisted round-trip itself look correct. The two findings are at the boundary around that state: making every released outcome durable, and replaying all release-time state to the daemon that missed it.

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human in the loop. Findings are advisory.

Re-reviewed after the persistence commit (`8732c00`), which landed after the previous automated review. No new issues found.

Spot-checks on the added persistence path:
- The release is now durably recorded: `broadcast_all_nodes_ready` (`binaries/coordinator/src/lib.rs`) persists via `make_record(Running)` at release, and `make_record`/`recovered` mirror `ready_barrier_released` + `barrier_exited_before_subscribe` symmetrically, so the flag and its verdict survive orphan reclaim and coordinator restart. Round-trip is covered by `the_release_survives_a_persist_and_rebuild_round_trip`.
- The success-only persist guard (`exited_before_subscribe.is_empty()`) does not open a hang: a failed barrier terminates the dataflow, and a reconnecting daemon then takes the `Failed { terminal: true }` branch (StopDataflow), not the `Running`/`Recovering` replay branches — so it is stopped rather than parked.
- `SCHEMA_VERSION` 3→4 is correct given redb's positional bincode encoding; `#[serde(default)]` alone would misdecode old rows.
- Borrow flow across the `.await` in the reconnect handler is fine: the `&mut` from `reestablish_running_dataflow` ends before the immutable `get`, and `daemon_connections`/`clock` are independent borrows.

As the author notes, the socket send itself is still unasserted (no fake daemon connection; tracked in #2999) — a coverage gap, not a defect in this diff.

---

heyong4725:
Both findings confirmed and fixed in 1c3052811.

**1. Failed barriers were not persisted at release.** Correct. `broadcast_all_nodes_ready` is reached with a non-empty `exited_before_subscribe` (`binaries/coordinator/src/lib.rs:457`) and nothing in that handler persists afterwards, so the release lived only in memory until some later failure path happened to write. A restart in that window leaves the record saying "not released" and reproduces #2998.

My guard was defensive about the wrong thing: I skipped the entire write to avoid promoting a failed dataflow to `Running`, when only the *status* needed guarding. The release now always persists; a failed barrier keeps whatever status the record already had, so nothing is promoted.

**2. The replay dropped the persisted-parameter half.** Correct, and the sharper of the two. The initial release is two halves — `AllNodesReady` plus `schedule_param_replay_for_ready_dataflow`. A daemon disconnected at broadcast time missed both, and my replay restored only the barrier, so its nodes would have been released with default state instead of the parameters the operator set. Silent wrong state rather than a hang, which is worse to diagnose.

`replay_all_nodes_ready` now replays that daemon's parameters as well. While fixing it I found the node-selection rule duplicated between the broadcast path and the replay path, so it is extracted as `nodes_on_daemon` and shared — the replay cannot select a different set than the broadcast would have.

**The coverage gap this PR kept deferring to #2999 is now closed.** I had been asserting the sends were untestable because there was no fake daemon connection. That was wrong — `ready_boundary_schedules_replay_for_daemon_nodes` in this same file already builds one (an mpsc channel plus a `pending_replies` map) and asserts on the exact bytes sent. `replaying_the_barrier_sends_it_and_the_params_to_that_daemon` uses that harness to drive the real replay and assert the reconnecting daemon receives both the `AllNodesReady` frame and the `SetParam` replay. It is bounded by a 10s timeout so a regression fails fast rather than wedging the job.

On that last point: I found this because a mutation stubbing the node selection to return nothing made `ready_boundary_schedules_replay_for_daemon_nodes` **hang** rather than fail. Worth noting independently of this PR — under a real regression that test would burn the CI job timeout instead of reporting a clean failure.

**Verification:** 120 coordinator lib + 22 + 4 tests green, fmt clean, `clippy --all-targets -D warnings` clean. Mutation-verified: reverting to a success-only persist, promoting a failed barrier to `Running`, and breaking the daemon filter in the node selection each fail a test. Full Linux CI dispatched on the new head.

heyong4725 (commented):
Re-reviewed current head `738c21b68`. The two prior blocking findings are resolved, and I found no new issues.

- Failed barriers are now persisted at the release boundary with their verdict while preserving the store's existing status. `a_failed_barrier_is_persisted_as_durably_as_a_successful_one` covers both durability and the no-promotion invariant.
- The reconnect path now replays the targeted daemon's persisted parameters after `AllNodesReady`, using the same extracted node-selection rule as the initial broadcast. `replaying_the_barrier_sends_it_and_the_params_to_that_daemon` exercises the real fake-connection transport and asserts both frames and the parameter reply.
- Bounding the five daemon-task awaits is appropriate; these are in-memory channel assertions, and the timeout converts a missing send into a named test failure instead of a job-level hang.

Validation on a non-committing merge with current `origin/main` (`313ca4ccf`):

- Integration merge and `git diff --check`: clean
- `cargo fmt --all -- --check`: pass
- `cargo test -p dora-coordinator`: 148 passed (122 unit, 22 control, 4 daemon). The socket-binding tests were rerun outside the sandbox after the sandbox correctly denied local listener creation.
- `cargo test -p dora-coordinator-store --features redb-backend`: 37 passed
- `cargo clippy -p dora-coordinator -- -D warnings`: pass

No remaining review blockers from this pass.

PR #3014, opened 2026-08-04
fix(daemon): reclaim memory pools once no live node can reach them

### LINKED ISSUE #2881: Memory pools may not be released when a node crashes or is dynamically removed
## Summary

While reviewing the daemon lifecycle, I noticed that memory pools registered by a node appear to remain tracked when that node is removed from a running dataflow or disconnects unexpectedly.

If this observation is correct, repeated node crashes or dynamic node removal could leave stale memory pool registrations until the entire dataflow is destroyed.

## Evidence

During `RemoveNode` cleanup, the daemon removes the node's runtime state and bookkeeping, but I could not find corresponding cleanup for memory pools registered by that node.

Relevant locations:

- `binaries/daemon/src/lib.rs`
  - `DaemonCoordinatorEvent::RemoveNode`
  - node cleanup path

- `MemoryPoolManager`
  - memory pool lifecycle

`cleanup_orphans` appears to operate at the dataflow level rather than the individual node level.

## Possible Impact

If memory pools remain registered after node removal:

- `/dev/shm` segments may accumulate
- stale pool registrations may persist
- long-running daemons could eventually exhaust the configured pool limit

## Suggested Direction

If this behavior is unintended, node removal and unexpected disconnect paths could free any memory pools registered by that node before the remaining bookkeeping is discarded.

I'm happy to investigate further and prepare a PR if this is considered a bug.

phil-opp:
Fixes #2881.

Pools registered by a node stayed in the daemon's table until the daemon itself
exited: neither `RemoveNode` nor a node crash released them, and `finish_dataflow`
did not either — so a long-running daemon (`dora up`) accumulated table entries and
`/dev/shm` segments across dataflows, walking toward the 512-entry pool cap.

Eagerly freeing a node's pools when it goes away would be wrong, as @SaitejaKommi
pointed out on the issue: a pool deliberately outlives its registrar. The normal
lifecycle is that a sender registers a pool and a *receiver* reads and frees it,
possibly well after the sender exited.

So reclaim by reachability instead. Each entry records who can still reach it:

- `touched_by` — nodes that have opened the pool, and
- `potential_readers` — the registrar's transitive downstream consumers as of
  registration time, which can still learn the pool id from a message already in
  flight (the daemon cannot see pool ids inside payloads, so the whole downstream
  closure counts).

A pool is released once none of those is still running locally. Every path that ends
an incarnation now checks: `RemoveNode`, `ReplaceNode` (whose outgoing exit event the
generation guard drops), and a crash, clean exit or restart. `finish_dataflow`
releases the remainder unconditionally — a finished dataflow has no node left to
serve, and its routing, channels and listeners are discarded in the same breath.

Nodes on other daemons never count as live: pools are host-local shared memory and
each daemon keeps its own table, so a remote consumer can never open this one's
segments.

Because a reclaimed pool has no live node in `touched_by`, there is nobody to send a
`FreeMemoryPool` notification to — unlike `free_memory_pool`, which must tell the
other holders to drop their per-process buffers.

`finish_dataflow` also sweeps the `/dev/shm` segments no table entry covers — a node
that died between creating a segment and registering it. That sweep is now scoped to
the nodes of *this* daemon, which it was not before: `/dev/shm` is host-wide, the
segment name embeds only the dataflow id, and `finish_dataflow` is a per-daemon
finish. With two daemons serving one dataflow on one machine (a `machine:` split),
whichever finished first would have unlinked the other's *live* segments, leaving a
consumer over there with `ENOENT`. `RunningDataflow` tracks its local nodes for this,
and the pre-existing spawn-time sweep — same hazard, narrower window — is scoped the
same way.

`register_memory_pool` gains a `potential_readers` parameter, `MemoryPoolEntry` a
field, and `cleanup_orphans` an `is_local_node` predicate, so this is a breaking
change to the `dora-memory-pool` crate surface. Its only consumer is `dora-daemon`.

Reachability cannot be a registration-time snapshot, because a running dataflow can
be rewired. `dora graph connect S/out C/in` after `S` registered a pool gives `C` a
path to it, and `S` is in its own reader set, so reclamation on `S`'s exit would have
unlinked the segment while `C` still had the id in flight. `AddMapping` and `AddNode`
therefore widen the reader set of every pool the edge's source can reach, by the new
target's downstream closure. Recomputing the whole set at reclaim time is not an
option — `RemoveNode` tears the node's mappings down first — but widening on the two
paths that create edges covers the same ground.

The closure also follows edges whose receiver is on another daemon. `mappings` holds
only what this daemon delivers, so a `local -> remote -> local` chain used to stop at
the first hop and drop the local node at its end, even though a pool id can travel
the whole way. A remote node never counts as live, so following through one costs
nothing.

Tests: reachability rules and dataflow-scoped cleanup in `dora-memory-pool`; the
downstream closure (transitive, cycle-safe) on `RunningDataflow`; and the daemon
wiring — a pool survives an exited sender while its consumer runs, and is reclaimed
once the last consumer is gone, when the exiting node is its only reference, and
regardless of unrelated long-lived nodes. The finish path is covered too: it releases
a pool a live consumer could still reach, and its `/dev/shm` sweep unlinks this
daemon's orphan while leaving a co-located daemon's segment alone. Plus the two
reachability gaps: a consumer connected after registration keeps the pool alive until
it too is gone, and the closure reaches a local node whose only path runs through
another daemon.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -694,6 +694,37 @@ fn reclaim_memory_pools_after_exit(
     }
 }
 
+/// Release everything this daemon still holds for a dataflow that just
+/// finished here: its pool table entries, plus the `/dev/shm` segments of
+/// `owned_nodes` that no table entry covers.
+///
+/// A daemon that outlives the dataflow (`dora up`) would otherwise carry
+/// every unfreed pool — table entry and segment alike — until it exits
+/// (dora-rs/dora#2881). Releasing the table entries is unconditional,
+/// deliberately: a dynamic node can outlive the finish (`should_finish`
+/// ignores them), but the dataflow's routing, channels and listeners are gone
+/// by now, so it can no longer send or receive anything a pool would serve,
+/// and no later event would ever reach [`reclaim_memory_pools_after_exit`]
+/// for this dataflow.
+///
+/// This is a *per-daemon* finish and `/dev/shm` is host-wide, so the sweep is
+/// restricted to this daemon's own nodes — see
+/// [`MemoryPoolManager::cleanup_orphans`].
+///
+/// A free function rather than a method so it can be exercised without a
+/// whole `Daemon`.
+fn release_memory_pools_of_finished_dataflow(
+    memory_pool: &MemoryPoolManager,
+    dataflow_id: Uuid,
+    owned_nodes: &BTreeSet<NodeId>,
+) {
+    let dataflow_id = dataflow_id.to_string();
+    memory_pool.cleanup_dataflow(&dataflow_id);
+    MemoryPoolManager::cleanup_orphans(&dataflow_id, |node| {
+        owned_nodes.iter().any(|id| id.as_ref() == node)
+    });
+}
+
 /// Convert MemoryPoolMetadata into daemon-protocol MetadataParameters.
 fn pool_metadata_to_params(meta: &MemoryPoolMetadata) -> MetadataParameters {
     use dora_message::metadata::Parameter;
@@ -2835,6 +2866,9 @@ impl Daemon {
                     // entry is registered.
                     running_node.mark_registered();
                     dataflow.running_nodes.insert(node_id.clone(), running_node);
+                    // Added after the spawn cohort, so it is not in
+                    // `spawn_nodes` — record it as ours for the orphan sweep.
+                    dataflow.owned_nodes.insert(node_id.clone());
 
                     // Update the daemon's stored descriptor so
                     // descriptor-based lookups (e.g. AllInputsClosed
@@ -3995,10 +4029,13 @@ impl Daemon {
         uv: bool,
         write_events_to: Option<PathBuf>,
     ) -> eyre::Result<impl Future<Output = eyre::Result<()>> + use<>> {
-        // Sweep orphaned /dev/shm segments from a previous crash of the
-        // same dataflow (keyed by dataflow_id, a UUID — safe in multi-
-        // daemon setups).
-        MemoryPoolManager::cleanup_orphans(&dataflow_id.to_string());
+        // Sweep orphaned /dev/shm segments left by a previous crash of this
+        // dataflow's nodes, scoped to the nodes this daemon spawns — a
+        // co-located daemon may be spawning the other half of the same
+        // dataflow right now (see `cleanup_orphans`).
+        MemoryPoolManager::cleanup_orphans(&dataflow_id.to_string(), |node| {
+            spawn_nodes.iter().any(|id| id.as_ref() == node)
+        });
 
         let mut logger = self
             .logger
@@ -4011,6 +4048,7 @@ impl Daemon {
             self.daemon_id.clone(),
             dataflow_descriptor.clone(),
         );
+        dataflow.owned_nodes = spawn_nodes.clone();
         // Read from the descriptor, which is the one copy that survives
         // everything a dataflow outlives: auto-recovery re-spawn,
         // coordinator restart with state reconstruction, and `dora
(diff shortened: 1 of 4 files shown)


phil-opp:

What I traced:
- Reachability is `touched_by ∪ potential_readers` (`libraries/extensions/memory-pool/src/lib.rs:64-70`), and `potential_readers` is the *full transitive* downstream closure (`binaries/daemon/src/running_dataflow.rs:1021-1039`), so a pool id forwarded through any chain of edges keeps the pool alive — the closure follows every queued node's outputs, not just the registrar's direct consumers, and the cycle guard via `reachable.insert` terminates.
- The exited node is correctly excluded from the live set in `reclaim_memory_pools_after_exit` (`binaries/daemon/src/lib.rs:658-662`), which matters because the `SpawnedNodeResult` path runs before the node leaves `running_nodes`; otherwise a consumer-less sender could never be reclaimed.
- `release_matching` drops the table lock before the `remove_file` unlinks, matching the existing `free_memory_pool` locking discipline, and skipping `FreeMemoryPool` notifications is justified — reclamation only fires when no live node remains in either set, so there is nobody to notify.
- Local-only liveness is correct: pools are host-local shared memory, so a remote consumer in the closure never pins one, and unlinking a segment a live node already mapped is safe (existing mappings survive `shm_unlink`).
- `finish_dataflow`'s unconditional `cleanup_dataflow` + `cleanup_orphans` is safe for dynamic nodes that outlive the finish, since their routing/channels are already torn down.
- The tests exercise the actual change, including the leak scenario (`unrelated node running`), the in-flight-transfer scenario (`consumer still running`), dataflow scoping, and the late-reader-kept-via-`touched_by` case.

---

phil-opp:
The reclamation itself is safe — freeing is `HashMap::remove` + `shm_unlink`, which never invalidates a live consumer's existing mapping, and the call-site coverage is complete. Two things:

**The finish-time sweep crosses daemon boundaries.** `finish_dataflow` is *per-daemon* local finish (`should_finish` is computed from this daemon's `running_nodes`; the log line says "dataflow finished on machine"). But `cleanup_orphans(dataflow_id)` unlinks **every** `/dev/shm/dora_pool_{dataflow_id}_*` on the host, unfiltered by table membership. Two daemons on one host running one dataflow (`examples/multiple-daemons`, a nightly job): daemon A's nodes finish first, A sweeps, and daemon B's *live* nodes lose their segments — B's table still lists them, so a B-side consumer that has not yet mmap'd gets ENOENT. The comment at `libraries/extensions/memory-pool/src/lib.rs:232-234` justifying this ("no two daemons run the same dataflow id concurrently") is false for multi-daemon dataflows. The spawn-time call site has a narrow window; this one can be arbitrarily long.

**The "known limitation" argument does not cover the case it is meant to.** The body says a node wired downstream after registration "can only learn the pool id from a node that is in `potential_readers`, and while any of those is alive the pool is retained" — but the registrar is itself in `potential_readers`, and reclamation fires on the registrar's own exit. So: `S` registers pool `P`, `dora graph connect S/out C/in`, `S` sends the id and exits, `P` is unlinked while `C` still has the id queued. Before this PR the pool leaked but the transfer succeeded, so this is a regression rather than "unchanged in spirit". Narrow, but the comment should say so.

Test coverage is thin where the new behavior actually lives: the four daemon cases exercise only the extracted helper, nothing covers the `finish_dataflow` additions.

phil-opp:
Rebased onto main and addressed all three points.

**The finish-time sweep crossing daemon boundaries.** Correct, and the spawn-time
sweep had the same defect with a narrower window. `cleanup_orphans` now takes an
`is_local_node` predicate and attributes each `/dev/shm` file to a node before
unlinking it — the segment name's last component is always the counter, so the node
id splits off unambiguously even when it contains `_`. A name that does not fit the
pattern is left alone: leaving an orphan is recoverable, unlinking a peer's live
segment is not. `RunningDataflow` tracks the nodes this daemon owns (`spawn_nodes`
plus anything `AddNode` adds later) to scope both call sites. The false justification
on `cleanup_orphans` is gone.

**The "known limitation" not covering its own case.** Also correct, and it was a
regression rather than a limitation, so it is fixed rather than documented.
`AddMapping` and `AddNode` — the only two paths that add an edge to a running
dataflow; `ReplaceNode` rejects any input change — now widen the reader set of every
pool the edge's source can reach by the new target's downstream closure. Recomputing
the whole set at reclaim time is indeed impossible, but updating where edges are
created covers the same ground.

While in there: `downstream_closure` walked `mappings`, which holds only edges this
daemon delivers, so a `local -> remote -> local` chain stopped at the first hop and
dropped the local node at its end — even though a pool id travels the whole chain.
Remote edges are recorded at spawn and followed now. A remote node never counts as
live, so following through one only widens what the closure reaches.

**Test coverage of the finish path.** The finish-time release is a free function now,
so it can be exercised without a whole `Daemon`: one test that it drops the table
entries a finished dataflow still holds, one that its sweep spares a co-located
daemon's segments. Plus the late-connected consumer (both halves: pinned while it
runs, reclaimed once it is gone), the reader-set widening not leaking into unrelated
pools, the segment-name parsing edge cases, and the closure's remote hop.

Two residual gaps, both erring toward retention rather than premature release:

- Runtime rewiring only reaches the target node's daemon, since that is where the
  coordinator routes `AddMapping`. A `local -> remote -> local` chain built at
  runtime through two `dora graph connect` calls therefore still misses the source
  daemon's reader sets.
- `potential_readers` only grows: `RemoveMapping` drops the edge but nothing removes
  the node, so repeated connect/disconnect cycles widen every pool's reader set and
  decay reclamation back toward teardown-only.

phil-opp:

The two commits since my last review address both issues raised on 2026-08-11, and I traced them against the diff:

- **Cross-daemon finish/spawn sweep** — `cleanup_orphans` now takes an `is_local_node` predicate and attributes each `/dev/shm/dora_pool_*` segment to a node (`rsplit_once('_')` + `counter.parse::<u64>()` guard) before unlinking, skipping any name that doesn't fit the pattern. Both call sites pass a per-daemon owner set, so a co-located daemon's live segments are no longer destroyed by a peer's finish.
- **Rewiring regression** — `AddMapping`/`AddNode` now call `extend_pool_readers_for_new_edge` after recording the edge, widening every reachable pool's readers by the target's `downstream_closure`; and `downstream_closure` now follows `remote_edges`, so a `local -> remote -> local` chain keeps the trailing local reader. A consumer connected after registration now pins the pool it can reach.

No new issues. The two residual gaps you documented (runtime `local -> remote -> local` rewiring reaching only the target daemon; `potential_readers` growing monotonically across connect/disconnect cycles) both err toward retention rather than premature free, so they're safe as follow-ups.

---

PR #3016, opened 2026-08-04
chore(deps): upgrade uhlc 0.5 -> 0.9 (#2446)

### LINKED ISSUE #2446: chore(deps): upgrade uhlc 0.5 -> 0.9 (wire-format break, needs protocol migration)
`dora-message` pins `uhlc = "0.5.1"`. Upgrading to 0.9 is **not** a drop-in bump: uhlc 0.9 changed the `ID` representation from `NonZeroU128` to `[u8; 16]`, and under bincode the serialized ID comes out **byte-reversed** relative to 0.5 (verified empirically: same timestamp, same 24-byte encoding length, ID bytes in opposite order). `uhlc::Timestamp` is embedded in every bincode-encoded daemon↔node and inter-daemon message, so a mixed-version deployment would silently corrupt HLC clock identities.

Upgrading therefore requires either a coordinated wire-format/protocol-version bump or a custom serde impl on our side that preserves the 0.5 byte order. Note zenoh vendors its own uhlc (0.8.x), so the two copies never unify regardless.

Deferred from the dependency-upgrade PR #2445.

phil-opp:
`dora-message` pinned `uhlc = "0.5.1"`. #2446 deferred the bump to 0.9 on the grounds that 0.9 serializes the HLC `ID` byte-reversed under bincode, so it needed a coordinated protocol migration or a custom serde impl.

That premise is wrong. 0.9 changes the in-memory representation of `ID` from `NonZeroU128` to `[u8; 16]`, but both serialize to the same 16 little-endian bytes — bincode writes a `u128` little-endian, and 0.5 built that `u128` from the id bytes in little-endian order to begin with. Verified by running the new golden-vector tests against both versions with the real `dora-message` types: every bincode vector is byte-identical, and a 0.5-encoded `Timestamp` cross-deserializes into 0.9 (and back) with its id bytes intact. What actually reversed in 0.9 is `ID`'s `Display`/`FromStr`, which dora never uses.

So the binary plane is untouched: daemon↔node, inter-daemon zenoh, and `dora record` files all keep their encoding, and existing recordings replay unchanged. No protocol version bump is needed — bumping `Metadata::CURRENT_VERSION` would have broken node↔daemon compatibility for nothing.

The JSON plane does change shape: on the CLI↔coordinator and coordinator↔daemon WebSocket links the timestamp `id` goes from a bare `u128` number to a 16-element byte array. Both links are already gated by the `dora_version` semver handshake, a mismatch now fails at parse time with a serde type error rather than being misread, and nothing persists a uhlc timestamp as JSON on disk.

That reshape is a net win. Under 0.5 a real HLC id used all 16 bytes, so `serde_json::to_value` failed outright with "number out of range" for *every* timestamp — which is why the coordinator, daemon and CLI all pass pre-serialized JSON fragments around. The `[u8; 16]` form round-trips through `serde_json::Value` losslessly, so this PR corrects the comments documenting that constraint (several still cited `NonZeroU128`) and turns the one place that skipped a check because of it — the coordinator's mock daemon — into a real assertion that daemon commands deserialize.

No source changes were required for the bump itself: `update_with_timestamp`'s error type went from `String` to `ExceedingDeltaError`, but all eight call sites only `Display` it. The dependency tree shrinks — uhlc 0.9's `rand`/`spin` were already present via zenoh's uhlc 0.8, while the 0.5 copy's `hex` and `uuid` deps drop out. MSRV 1.88 still builds.

New `libraries/message/tests/uhlc_wire_format.rs` pins both encodings so a future bump cannot move them unnoticed.

Fixes #2446.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2252,7 +2252,7 @@ dependencies = [
  "serde_yaml",
  "tempfile",
  "tokio",
- "uhlc 0.5.2",
+ "uhlc 0.9.0",
  "uuid",
 ]
 
@@ -8258,24 +8258,23 @@ checksum = "2896d95c02a80c6d6a5d6e953d479f5ddf2dfdb6a244441010e373ac0fb88971"
 
 [[package]]
 name = "uhlc"
-version = "0.5.2"
+version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8d291a7454d390b753ef68df8145da18367e32883ec2fa863959f0aefb915cdb"
+checksum = "b62a645e3e4e6c85b7abe49b086aa3204119431f42b6123b0070419fb6e9d24e"
 dependencies = [
- "hex",
  "humantime",
  "lazy_static",
  "log",
+ "rand 0.8.7",
  "serde",
- "spin 0.9.9",
- "uuid",
+ "spin 0.10.1",
 ]
 
 [[package]]
 name = "uhlc"
-version = "0.8.2"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b62a645e3e4e6c85b7abe49b086aa3204119431f42b6123b0070419fb6e9d24e"
+checksum = "55bb25449f66ea76693b51f631fdd190821683c65e65e68d990b36c2d1ae9dd2"
 dependencies = [
  "humantime",
  "lazy_static",
--- a/binaries/cli/src/ws_client.rs
+++ b/binaries/cli/src/ws_client.rs
@@ -12,8 +12,9 @@ use tokio_tungstenite::tungstenite::Message;
 use uuid::Uuid;
 
 /// Helper for deserializing incoming WS frames without going through
-/// `serde_json::Value` for the result/payload fields. This preserves
-/// u128 fidelity for uhlc::ID inside timestamps.
+/// `serde_json::Value` for the result/payload fields: they are handed to the
+/// caller as raw bytes to be parsed into their real type, so they are never
+/// re-serialized from an intermediate `Value`.
 #[derive(serde::Deserialize)]
 struct IncomingFrame {
     #[serde(default)]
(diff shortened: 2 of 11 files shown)


phil-opp:

I reviewed the current diff. No issues found in the code itself. The bincode-invariance golden vectors (`libraries/message/tests/uhlc_wire_format.rs`) are well-designed — distinct id bytes so any reordering surfaces, `Metadata`/`Timestamped` composites pinned, and `a_freshly_generated_hlc_id_uses_all_sixteen_bytes` guards the premise of the `serde_json::Value` round-trip test so it can't pass for a non-generalizing reason. The `mock_daemon` change (`binaries/coordinator/src/run/mod.rs:301-304`) is a real improvement: it converts a previously-skipped deserialization into an actual `Timestamped<DaemonCoordinatorEvent>` assertion.

One thing I'd flag for a maintainer's explicit confirmation, not as a defect: the whole "not a wire break" argument rests on uhlc 0.5's `Timestamp` bincode output being byte-identical to 0.9's. The committed test can only pin 0.9's encoding (the hardcoded `TIMESTAMP_HEX`) — it cannot prove the 0.5 side inside CI, since only one uhlc version is linked per build. The commit states this was cross-checked against 0.5.2 by hand. If that manual check is trusted, the binary plane (daemon↔node, inter-daemon zenoh, `dora record` files) is safe and no `Metadata::CURRENT_VERSION` bump is needed. The JSON-plane reshape (`id` number → 16-byte array on the WS links) is a real shape change, correctly relied on being gated by the `dora_version` handshake so a mismatch fails at parse rather than being misread.

---

phil-opp:

Follow-up on the latest commit since my previous review: declaring the dev-only `uuid/v4` feature for `dora-message` and `dora-coordinator-store` looks safe to merge. It only restores, in `[dev-dependencies]`, a feature that `Uuid::new_v4()` in the `#[cfg(test)]` modules relied on and that uhlc 0.5 used to supply through feature unification; no production code is affected and no new crates enter the lockfile. No new issues found.

---

PR #3017, opened 2026-08-04
chore(deps): redb 2 -> 4 and sysinfo 0.38 -> 0.39, MSRV 1.88 -> 1.95 (#2449)

### LINKED ISSUE #2449: chore(deps): redb 4 + sysinfo 0.39 blocked on MSRV (need Rust 1.89 / 1.95)

phil-opp:
Closes #2449.

Both deps were held back only by the workspace MSRV: every redb 3.x/4.x needs Rust 1.89, and sysinfo 0.39 needs 1.95. This raises `rust-version` to **1.95.0** — the lowest floor that admits both — and takes redb 2.6.3 → 4.1.0 and sysinfo 0.38.4 → 0.39.6. The lockfile diff is exactly those two crates plus one macOS-only transitive (`objc2-open-directory`).

**The store-format break.** redb 3 dropped the v2 on-disk format, so a coordinator store written by an older dora cannot be opened after this bump. redb reports that as a bare `Manual upgrade required. Expected file format version 3, but file is version 2`, which gives the user nothing to act on. `RedbStore::open` now maps `DatabaseError::UpgradeRequired` onto the existing `SCHEMA_MISMATCH_MARKER`, routing it into the recovery path a bincode schema bump already uses: `dora up` prints the `--recreate-store` hint, which *archives* the old file rather than deleting it.

I deliberately did not carry a redb 2.6 dependency to do an in-place `Database::upgrade()`. The store holds coordinator state that daemons re-register on reconnect, `--recreate-store` already preserves the old file, and a permanent second redb major in the tree is a poor trade for a one-time pre-1.0 migration. Happy to revisit if the persisted state turns out to be worth more than that.

**API migration** was small: `begin_read` moved to the new `ReadableDatabase` trait (coordinator-store + the e2e test). The rest of the surface we use is `&str`/`&[u8]`/`u32` tables, so redb 3's variable-width tuple re-encoding and the `Legacy` removal don't apply. sysinfo 0.39 needed no code changes at all.

One incidental fix: raising the MSRV let clippy's MSRV-aware `missing_const_for_fn` reach `Package::new`, which is now `const`.

### Verification

- `cargo fmt --all -- --check`, `cargo clippy --all -- -D warnings` — clean
- `cargo test --all` — 1537 passed. The 2 failures are the container-local `rmw_zenoh_pubsub` discovery timeouts (no multicast), which fail on `main` here too
- `cargo check --examples` — clean
- `cargo test --test ws-cli-e2e -- store` — 5/5, including `e2e_up_recreate_store_archives_incompatible_store`
- `cargo hack check --rust-version --workspace --ignore-private --locked` — 29/29 crates check at 1.95.0
- `make qa-fast` — audit (advisories/bans/**licenses**/sources), typos, unwrap-budget all pass

New regression test `redb_v2_file_format_is_reported_as_a_store_mismatch` stamps a real store's header down to the v2 format and asserts the marked error, so the recovery hint can't silently regress.

### Two things to know

1. **MSRV 1.95.0 is a real jump** (from 1.88.0). Rust 1.95 shipped 2026-04-14, so it's ~4 months old at time of writing. Downstream users on 1.88–1.94 are cut off. If that's too aggressive, a redb-only variant at MSRV 1.89 is a one-line change — sysinfo 0.39 would go back on the blocked list.
2. **The nightly `msrv` job is already failing on `main`**, unrelated to this PR: `dora-core` (and therefore `dora-hub-client`) references `zenoh::Config` at `libraries/core/src/topics.rs:63` while `zenoh` is an optional feature-gated dep, so neither crate builds standalone with default features. It reproduces on stable 1.97.1. Those two are the only crates excluded from the `cargo hack` run above; worth a separate issue.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,7 +28,7 @@ Key communication paths:
 ## Workspace Facts
 
 - Rust edition: 2024
-- MSRV: 1.88.0
+- MSRV: 1.95.0
 - Shared workspace versioning
 - Python packages are built with `maturin`, not normal `cargo` flows
 - The repository is in an `adora` -> `dora` consolidation transition. Prefer `dora` names for new code, but preserve documented compatibility paths unless the change is explicitly a breaking cleanup.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1963,7 +1963,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "sha2 0.11.0",
- "sysinfo 0.38.4",
+ "sysinfo 0.39.6",
  "tabwriter",
  "tempfile",
  "termcolor",
@@ -2107,7 +2107,7 @@ dependencies = [
  "shared_memory_extended",
  "shellexpand 3.1.2",
  "shlex",
- "sysinfo 0.38.4",
+ "sysinfo 0.39.6",
  "tempfile",
  "tokio",
  "tokio-stream",
@@ -4980,6 +4980,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2a180dd8642fa45cdb7dd721cd4c11b1cadd4929ce112ebd8b9f5803cc79d536"
 dependencies = [
  "bitflags 2.13.1",
+ "dispatch2",
+ "objc2",
 ]
 
 [[package]]
@@ -5006,6 +5008,17 @@ checksum = "33fafba39597d6dc1fb709123dfa8289d39406734be322956a69f0931c73bb15"
 dependencies = [
  "libc",
  "objc2-core-foundation",
+]
+
+[[package]]
+name = "objc2-open-directory"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bb82bed227edf5201dfedf072bba4015a33d3d4a98519837295a90f0a23f676d"
+dependencies = [
+ "objc2",
+ "objc2-core-foundation",
+ "objc2-foundation",
 ]
 
 [[package]]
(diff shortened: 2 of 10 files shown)


phil-opp:

Checks I did:
- The redb 4 API move is fully accounted for: `ReadableDatabase` is the only trait relocation used (`begin_read`), and grepping the workspace confirms `libraries/coordinator-store/src/redb_store.rs` and `tests/ws-cli-e2e.rs` are the only redb consumers — both updated here, so no call site is left to break the build.
- `libraries/coordinator-store/src/redb_store.rs:65-84` correctly narrows the new arm to `DatabaseError::UpgradeRequired(version)` and routes everything else through the existing `wrap_err`, so no other open failure is misclassified. Reusing `SCHEMA_MISMATCH_MARKER` keeps the `dora up` → `--recreate-store` recovery path intact (the marker is asserted end-to-end).
- The v2-format test (`redb_v2_file_format_is_reported_as_a_store_mismatch`) meaningfully exercises the new arm rather than asserting on nothing: it stamps both commit-slot version bytes to force a real `UpgradeRequired`, and the extra `msg.contains("file format")` assertion guards against the header hand-stamp silently ceasing to trigger an upgrade error if redb's layout shifts.
- MSRV 1.95.0 stays under the CI-pinned 1.97.1 toolchain; the `Package::new` → `const fn` change is an unrelated but harmless clippy consequence of the bump.

One thing worth a maintainer's explicit sign-off (not a code defect): the decision to recover a v2 store by archive-and-recreate rather than carrying redb 2.6 for an in-place `Database::upgrade()`. It's documented in the commit and justified by coordinator state being re-registered on reconnect, so this is a judgment call rather than a bug.

---

Issue #2856, opened 2026-07-27
`dora run` orphan containment: remaining gaps after #3018 — pre-`init` window, `path: shell` nodes, Windows Job Object


heyong4725:
## Summary

When the `dora run` process is killed abnormally (SIGKILL, supervisor timeout, `kill` to its process group), the dataflow's **node processes survive it** and busy-spin at 100–170% CPU indefinitely. They never exit on their own — we have observed an orphaned node accumulate **59 CPU-hours** before being found.

## Environment

- dora-cli 0.5.0, `dora run <graph>.yaml --uv`
- macOS 26 (arm64), Python 3.13 nodes (`from dora import Node`, pyarrow payloads)
- Nodes are plain Python processes spawned by the runtime (`python3 -u <node>.py`)

## Reproduction

1. `dora run dataflow.yaml --uv` with any dataflow of a few Python nodes (ours have timer-driven inputs, e.g. `dora/timer/millis/10`).
2. Kill the `dora run` process hard while the dataflow is running — any of:
   - `kill -9 <pid of dora run>`
   - `timeout 30 dora run ...` (the pattern that first bit us: a test harness wrapping runs in `timeout`)
   - `os.killpg(pid, SIGKILL)` on the process group `dora run` was started in (`start_new_session=True`)
3. Observe surviving `python3 -u <node>.py` processes: `ps aux | awk '$3 > 50'`.

The killpg case is notable: the node processes are evidently not (all) in the `dora run` process group, so even a group kill misses them.

## Expected

Node processes exit (or are reaped) when their runtime/daemon dies — ideally via a parent-death signal (`PR_SET_PDEATHSIG` equivalent; on macOS e.g. a watchdog on the daemon connection) or a bounded read-timeout → exit path.

## Actual

Orphaned nodes spin in their event loop at 100–170% CPU each, apparently busy-polling the dead dataflow's channels, forever. On a laptop-class machine a handful of orphans dominate the cores and silently corrupt any timing-sensitive workload that runs afterwards (this is how we usually notice — unexplained perf regressions, then `ps` shows orphans from hours or days earlier).

## Relationship to existing issues

- #2152 describes nodes failing to exit after a **clean dataflow stop** with the daemon spinning on TCP read timeouts — plausibly the same underlying no-exit-on-dead-peer loop, but the trigger here is the runtime being killed, not a stop.
- #1996 (closed) covered lingering **daemons** when the coordinator is gone; this is the node-level analogue.

## Workaround we use

[...]

PR #3018, opened 2026-08-04
fix(daemon): end `dora run` nodes when the CLI is killed outright (#2856)

phil-opp:
Closes the residual half of #2856. The busy-spin is gone and #2949/#2961 fixed
`SIGTERM`/`SIGHUP` to `dora run`, but `SIGKILL` still orphaned every node that
does not poll its event stream: nothing of ours runs after `SIGKILL`, so the
stop ladder and the destroy reaper are unavailable by construction. And because
nodes are spawned as process-group leaders on purpose (so a terminal `Ctrl-C`
cannot kill them out from under the daemon), the orphan ends up with `ppid 1`
in a group of its own — reachable by neither inherited signal delivery nor a
group-kill of the CLI.

The containment therefore has to live in the node. `Daemon::run_dataflow` — the
in-process shape used by `dora run`, `dora daemon --run-dataflow` and
embedders — injects `DORA_RUN_PARENT_PID` into the nodes it spawns. The node
API arms a watchdog thread from it at `DoraNode::init`, and once that pid is
gone it `SIGKILL`s its own process group. The group, not the process, because
the tracked child is often a wrapper (`uv run python …`) with the real node one
level down — the reporter's shape, verified by hand.

`dora up` + `dora start` deliberately does not set the variable, and it is
scrubbed from the inherited environment there, so a node keeps outliving
coordinator drops, reconnects and watchdog disconnects (#2029).

`PR_SET_PDEATHSIG` was rejected as the mechanism: it reaches only the direct
child (missing the `uv` wrapper), it fires on parent *thread* exit while the
daemon spawns from a tokio worker, and it is Linux-only while the report is
from macOS.

Not covered, deliberately: a node killed before it reaches `init`, a node that
never calls `init` (`path: shell`), and Windows — where the equivalent is
`JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` on the job object nodes already run in,
which `process-wrap` couples to `Command::kill_on_drop` and which cannot be
verified from this workspace.

`tests/node-lifecycle-e2e.rs` gains a sibling of the `SIGTERM` test: it starts
`dora run` in its own process group, `SIGKILL`s the CLI, and asserts the
fixture node is gone within 30s. The fixture ignores `SIGTERM` and never polls
its event stream, so a containment scheme that only delivers `SIGTERM` fails
it. Verified failing before the fix (node survived the full 30s) and passing
after. The two orphan tests now share one `StubbornRun` harness instead of a
duplicated teardown.

Refs #2856.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2292,6 +2292,7 @@ dependencies = [
  "futures-timer",
  "inquire",
  "is-terminal",
+ "libc",
  "opentelemetry",
  "serde_json",
  "serde_yaml",
--- a/apis/rust/node/Cargo.toml
+++ b/apis/rust/node/Cargo.toml
@@ -49,6 +49,12 @@ is-terminal = "0.4.17"
 thiserror = { workspace = true }
 uuid = { workspace = true }
 
+# Process-group signalling for the `dora run` orphan guard (`orphan_guard.rs`).
+# Unix-only: the mechanism is `getppid`/`killpg`, which Windows has no
+# equivalent of.
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [dev-dependencies]
 criterion = { version = "0.5", features = ["html_reports"] }
 
--- a/apis/rust/node/src/lib.rs
+++ b/apis/rust/node/src/lib.rs
@@ -126,6 +126,7 @@ mod error;
 pub mod event_stream;
 pub mod integration_testing;
 mod node;
+mod orphan_guard;
 
 pub use error::{NodeError, NodeResult, PatternError};
 
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -1130,6 +1130,11 @@ impl DoraNode {
         node_config: NodeConfig,
         testing_communication: Option<TestingCommunication>,
     ) -> NodeResult<(Self, EventStream)> {
+        // Before anything that can fail or block: a node spawned by `dora run`
+        // must not outlive the CLI even if the rest of this initialization
+        // stalls (dora-rs/dora#2856). A no-op on every other spawn path.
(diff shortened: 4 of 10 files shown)


phil-opp:

Things I specifically checked:
- `apis/rust/node/src/orphan_guard.rs:159-166` — the direct-child branch uses `getppid()` against the recorded pid, which reparenting makes an exact liveness test no pid recycling can fool; the wrapped-node branch (`kill(pid, 0)` + `ESRCH`) is correctly one-sided (only `ESRCH` counts as gone, `EPERM`/`0` read as alive), so it can only fail to kill, never kill the wrong process.
- `apis/rust/node/src/orphan_guard.rs:188` — `killpg(getpgrp(), SIGKILL)` targets the node's own group, which is safe only because the var is injected exclusively by the daemon on group-leader spawns; `DORA_RUN_PARENT_PID` is added to `CONTROL_PLANE_ENV` (`binaries/daemon/src/spawn/spawner.rs:73`) so a descriptor- or shell-supplied value can't arm the guard against a stranger.
- `apis/rust/node/src/orphan_guard.rs:118-122` — the pre-thread liveness check closes the fork/arm race, and `_exit` (not `process::exit`) on the fallback avoids running other threads' destructors concurrently.
- Scoping: `bind_nodes_to_parent` is `true` only on the `run_dataflow` path and `false` for the standalone `dora up` daemon (`binaries/daemon/src/lib.rs:1027,1397`), preserving the #2029 property that daemon-spawned nodes outlive coordinator drops/reconnects.
- The new `run_killed_by_sigkill_does_not_orphan_nodes` test meaningfully exercises the fix: the fixture ignores `SIGTERM` and never polls, so a scheme that only delivered `SIGTERM` (or relied on cooperative teardown) would fail it.

---

PR #3028, opened 2026-08-04
fix(daemon): stop the finish-straggler watchdog misclassifying operator nodes as sources

### LINKED ISSUE #3021: daemon: finish-straggler watchdog misclassifies operator nodes as sources, disarming it for the whole dataflow
`node_never_finishes` (`running_dataflow.rs:1069`) checks a node's
top-level `inputs` map to decide if it's a source. That's wrong for a
runtime (`operators:`) node its inputs live under
`operators[].config.inputs`, so the top level is always empty. Every
operator node with real inputs gets misclassified as a source.

Impact is dataflow-wide, not per-node: `select_finish_stragglers` bails
out entirely (no escalation for anyone) the instant it sees one
`never_finishes` node. So one operator node permanently disarms the
watchdog for the whole graph  if any other node later wedges, nothing
catches it.

Already found and knowingly deferred in #2957:
```
// NOTE: this misclassifies a runtime (`operators:`) node as a
// source ... Pre-existing and left alone here: correcting it
// would newly arm the straggler watchdog for operator dataflows,
// which is a behavior change well outside #2920.
```
Same PR fixed the identical mismatch elsewhere in this file
(`is_drained`/`has_data_input`) by reading registered inputs instead of
the descriptor. This spot never got the same fix.

### Fix
```rust
let is_source = !self.has_data_input(node_id);
```

CapThunder19:
Fixes #3021 

`node_never_finishes` derived "is this a source" from the descriptor's
top-level `inputs` map, which is always empty for a runtime
(`operators:`) node its inputs live under `operators[].config.inputs`.
Every operator node with real inputs was misclassified as a source.

Impact was dataflow-wide: `select_finish_stragglers` bails out entirely
(no escalation for anyone) on the first `never_finishes` node it sees,
so one operator node permanently disarmed the finish-straggler watchdog
for the whole dataflow.

Already found and knowingly deferred in #2957. The sibling case in this
file (`is_drained`/`has_data_input`) was already fixed the same way;
`node_never_finishes` just never got it.

Reuses `has_data_input`, which reads `data_inputs` recorded at
registration time instead of the raw descriptor.

**Test:** a node with an empty top-level `inputs` map but a registered
data input (the operator-node shape) must not be classified as a source.

`cargo test -p dora-daemon`, `cargo clippy -p dora-daemon -- -D warnings`,
`cargo fmt --all -- --check` all pass.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -7857,6 +7857,18 @@ mod fault_tolerance_tests {
         df.connected_nodes.insert(node.clone());
     }
 
+    /// Record a real `User` data input for `node`, exactly as the spawn and
+    /// `AddNode` paths do. `node_never_finishes` classifies "is this a source"
+    /// from `data_inputs`, so a fixture that skips this models a *source*, and
+    /// a source vetoes the watchdog for the whole dataflow — every assertion
+    /// below about a user-input node needs the input recorded to mean anything.
+    fn register_data_input(df: &mut RunningDataflow, node: &NodeId) {
+        df.data_inputs
+            .entry(node.clone())
+            .or_default()
+            .insert("value".to_string().into());
+    }
+
     #[test]
     fn timer_fed_node_is_not_a_finish_straggler() {
         // A long-running timer-only node never drains and sends no daemon
@@ -7885,6 +7897,7 @@ mod fault_tolerance_tests {
         let mut df = test_dataflow();
         let stuck: NodeId = "stuck".to_string().into();
         insert_silent_node(&mut df, &stuck);
+        register_data_input(&mut df, &stuck);
 
         let now = node_communication::current_millis();
         let selected = df.finish_stragglers(Duration::from_millis(1), now);
@@ -7901,6 +7914,7 @@ mod fault_tolerance_tests {
         let running = test_running_node();
         running.last_activity.store(1, atomic::Ordering::Release);
         df.running_nodes.insert(starting.clone(), running);
+        register_data_input(&mut df, &starting);
         // NOTE: deliberately NOT added to subscribe_channels (not connected).
 
         let now = node_communication::current_millis();
@@ -7922,6 +7936,7 @@ mod fault_tolerance_tests {
         running.last_activity.store(1, atomic::Ordering::Release);
         df.running_nodes.insert(stuck.clone(), running);
         df.connected_nodes.insert(stuck.clone());
+        register_data_input(&mut df, &stuck);
         // NOTE: no subscribe_channels entry — the event stream was dropped.
 
         let now = node_communication::current_millis();
@@ -7942,6 +7957,7 @@ mod fault_tolerance_tests {
         let running = test_running_node();
         running.last_activity.store(1, atomic::Ordering::Release);
         df.running_nodes.insert(node_a.clone(), running);
+        register_data_input(&mut df, &node_a);
 
         let now = node_communication::current_millis();
         let selected = df.finish_stragglers(Duration::from_millis(1), now);
@@ -7959,6 +7975,7 @@ mod fault_tolerance_tests {
         let mut df = test_dataflow();
         let node: NodeId = "reused".to_string().into();
         insert_silent_node(&mut df, &node);
+        register_data_input(&mut df, &node);
         df.timers
             .entry(Duration::from_millis(100))
             .or_default()
@@ -7998,6 +8015,7 @@ mod fault_tolerance_tests {
         );
         df.running_nodes.insert(node.clone(), running);
         df.connected_nodes.insert(node.clone());
+        register_data_input(&mut df, &node);
         df.all_inputs_closed_at.insert(
             node.clone(),
             std::time::Instant::now() - Duration::from_secs(1),
--- a/binaries/daemon/src/running_dataflow.rs
+++ b/binaries/daemon/src/running_dataflow.rs
@@ -1082,17 +1082,16 @@ impl RunningDataflow {
         if !self.timers_gate_drain && self.is_drained(node_id) {
             return false;
         }
-        // NOTE: this misclassifies a runtime (`operators:`) node as a
-        // source, because its inputs live under `operators[].config.inputs`
-        // and the top-level map is empty. Pre-existing and left alone here:
-        // correcting it would newly arm the straggler watchdog for operator
-        // dataflows, which is a behavior change well outside #2920.
-        let is_source = self
-            .descriptor
-            .nodes
-            .iter()
-            .find(|n| &n.id == node_id)
-            .is_some_and(|n| n.inputs.is_empty());
+        // `has_data_input` reads `data_inputs`, recorded at registration
+        // time, rather than the raw descriptor: a runtime (`operators:`)
+        // node keeps its inputs under `operators[].config.inputs` and has
+        // an empty top-level `inputs` map, so a descriptor-derived check
+        // reports every operator node as a source and permanently disarms
+        // the straggler watchdog for the whole dataflow the moment one is
+        // running (`select_finish_stragglers` returns early on the first
+        // `never_finishes` node it sees). Same fix `is_finished_non_source`
+        // already applies for the same reason.
+        let is_source = !self.has_data_input(node_id);
         let has_timer = self
             .timers
             .values()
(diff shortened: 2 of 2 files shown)


phil-opp:
> 🤖 **Automated review by Claude** — this review was generated fully automatically, with no human in the loop.

No issues found.

I traced the change in `RunningDataflow::node_never_finishes` (`binaries/daemon/src/running_dataflow.rs`):

- Replacing the descriptor-derived `is_source` with `!self.has_data_input(node_id)` is correct and makes this method consistent with `is_finished_non_source` and `is_drained`, both of which already read the registration-time `data_inputs` for the same operator-node reason. A real source (no inputs) still reads as a source (`data_inputs` empty → `has_data_input` false), and a timer-only/logs-only node is unchanged because `has_timer`/`has_logs` still carry it. The only behavior change is the intended one: an `operators:` node with real data inputs is no longer misclassified as a source, so it stops permanently vetoing the finish-straggler watchdog for the whole dataflow.
- The new test `operator_shaped_node_with_data_input_is_not_a_source` is a genuine (non-vacuous) regression test: `timers_gate_drain` defaults to `true`, so the early `is_drained` return is not taken and the assertion actually reaches the changed `is_source` line. It fails under the old descriptor-derived logic (empty top-level `inputs` → `is_source == true`) and passes under the new one, so it exercises exactly the code that changed.

---

phil-opp:
The merge-queue failure is real and reproduces on the PR head alone (`491147c`), not just on the merge commit — `cargo test -p dora-daemon --lib` fails 4 tests in `fault_tolerance_tests`:

```
cleared_timer_state_makes_reused_id_escalatable
dropped_stream_node_is_still_a_finish_straggler
reopened_input_clears_stale_drain_timestamp
wedged_user_input_node_is_a_finish_straggler
```

All four assert a node **is** selected by `finish_stragglers`, and all four now get `[]`.

**Cause.** These fixtures build their nodes straight into `running_nodes` / `connected_nodes` and never touch the descriptor or `data_inputs`. Under the old descriptor-derived check that was fine — a node absent from `descriptor.nodes` made `find(...)` return `None`, so `is_some_and` gave `is_source == false` and the node stayed escalatable. Under `!self.has_data_input(node_id)` a node with no recorded `data_inputs` is now a source, `node_never_finishes` returns true, and `select_finish_stragglers` returns early with `Vec::new()` for the whole dataflow.

So it's the fixtures that are stale, not the production change: a running node with no registered data inputs never happens in production (the spawn path at `lib.rs:4005` and the `AddNode` path at `lib.rs:2714` both record every `InputMapping::User` input), and each of these tests says in its own comment that it is modelling a *user-input* node. The fix is to record the input the fixture was implicitly assuming.

Patch below — adds a `register_data_input` helper next to `insert_silent_node` and calls it in the four failing tests plus `unconnected_slow_starting_node_is_not_a_finish_straggler` and `removed_node_id_is_not_connected_on_reuse`. Those last two still pass today, but only because the source veto short-circuits before their intended `connected` check — without the input they'd be silently vacuous.

<details>
<summary>Patch</summary>

```diff
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -7857,6 +7857,18 @@ mod fault_tolerance_tests {
         df.connected_nodes.insert(node.clone());
     }
 
+    /// Record a real `User` data input for `node`, exactly as the spawn and
+    /// `AddNode` paths do. `node_never_finishes` classifies "is this a source"
+    /// from `data_inputs`, so a fixture that skips this models a *source*, and
+    /// a source vetoes the watchdog for the whole dataflow — every assertion
+    /// below about a user-input node needs the input recorded to mean anything.
+    fn register_data_input(df: &mut RunningDataflow, node: &NodeId) {
+        df.data_inputs
+            .entry(node.clone())
+            .or_default()
+            .insert("value".to_string().into());
+    }
+
     #[test]
     fn timer_fed_node_is_not_a_finish_straggler() {
@@ -7885,6 +7897,7 @@ mod fault_tolerance_tests {
         let stuck: NodeId = "stuck".to_string().into();
         insert_silent_node(&mut df, &stuck);
+        register_data_input(&mut df, &stuck);
 
@@ -7901,6 +7914,7 @@ mod fault_tolerance_tests {
         df.running_nodes.insert(starting.clone(), running);
+        register_data_input(&mut df, &starting);
         // NOTE: deliberately NOT added to subscribe_channels (not connected).
 
@@ -7922,6 +7936,7 @@ mod fault_tolerance_tests {
         df.connected_nodes.insert(stuck.clone());
+        register_data_input(&mut df, &stuck);
         // NOTE: no subscribe_channels entry — the event stream was dropped.
 
@@ -7942,6 +7957,7 @@ mod fault_tolerance_tests {
         df.running_nodes.insert(node_a.clone(), running);
+        register_data_input(&mut df, &node_a);
 
@@ -7959,6 +7975,7 @@ mod fault_tolerance_tests {
         insert_silent_node(&mut df, &node);
+        register_data_input(&mut df, &node);
         df.timers
 
@@ -7998,6 +8015,7 @@ mod fault_tolerance_tests {
         df.connected_nodes.insert(node.clone());
+        register_data_input(&mut df, &node);
         df.all_inputs_closed_at.insert(
```

</details>

`timer_fed_node_is_not_a_finish_straggler` is left alone on purpose: a timer-only node genuinely has no user inputs, so source-shaped is the right model for it, and `cleared_timer_state_makes_reused_id_escalatable` still covers the timer arm on its own (timer + data input → vetoed; clear the timer → escalated).

**Verified locally:**
- with the patch: `cargo test -p dora-daemon` → 188 passed, 0 failed
- with the patch but `is_source` reverted to the old descriptor form: 187 passed, and the only failure is this PR's own `operator_shaped_node_with_data_input_is_not_a_source` — so the fixture change is a faithful model of both behaviors rather than something that masks the regression, and the new test is genuinely RED against the old logic
- `cargo fmt --all -- --check` and `cargo clippy -p dora-daemon -- -D warnings` clean

phil-opp:

The follow-up commit (`047fc63`) lines up with the fix: `register_data_input` records the `User` input that the spawn and `AddNode` paths always record in production (`lib.rs` ~2714 / ~4005), so the six `fault_tolerance_tests` that assert a node **is** a finish-straggler now model a real running node rather than an accidental source. Confirmed non-vacuous — reverting `is_source` back to the descriptor-derived form fails only `operator_shaped_node_with_data_input_is_not_a_source`, and `timer_fed_node_is_not_a_finish_straggler` is correctly left with no data input.

One observation on the (unchanged) production change, just for the record: `!self.has_data_input(node_id)` now defines "source" purely from registration-time `data_inputs`, so any path that reaches `node_never_finishes` for a node *before* its inputs are registered would transiently read it as a source. The current spawn/`AddNode` ordering doesn't open that window, so no action needed — noting it only so a future reorder of registration vs. watchdog arming keeps it in view. No blocking issues.

---

phil-opp:
Root cause of the `Test (ubuntu-latest)` failure, in case it helps — the same four tests fail on both batch SHAs this PR has been in (`fc73f4df` and the earlier one), always `202 passed; 4 failed`:

```
fault_tolerance_tests::cleared_timer_state_makes_reused_id_escalatable
fault_tolerance_tests::dropped_stream_node_is_still_a_finish_straggler
fault_tolerance_tests::reopened_input_clears_stale_drain_timestamp
fault_tolerance_tests::wedged_user_input_node_is_a_finish_straggler
```

all with the same shape — an empty straggler list where one was expected:

```
assertion `left == right` failed
  left: []
 right: [NodeId("stuck")]
```

These four pass on current `main` (verified: 204/204 in `cargo test -p dora-daemon --lib` at `a8d3bbf01`), so it is this change rather than a pre-existing flake.

The mechanism looks like a fixture/logic mismatch. The PR swaps the source check from the descriptor:

```rust
let is_source = /* descriptor lookup */ n.inputs.is_empty();
```

to registration-time state:

```rust
let is_source = !self.has_data_input(node_id);
```

and `has_data_input` returns `false` when `data_inputs` has no entry at all:

```rust
self.data_inputs.get(node_id).is_some_and(|inputs| !inputs.is_empty())
```

The failing tests build their nodes with helpers like `insert_silent_node`, which populate `running_nodes`, `subscribe_channels` and `connected_nodes` but never call `register_inputs` — so `data_inputs` is empty, `is_source` is now `true`, the node counts as never-finishing, and it drops out of the straggler list.

Adding `register_inputs` to those fixtures would make them green. But it may be worth checking the non-test side of the same question first: is there a window where a real node is running and connected but has not yet registered its inputs? If so, `has_data_input` reads `false` there too and the watchdog would be disarmed for that node during the window — the same class of bug this PR is fixing, just moved. If registration is guaranteed to precede any straggler evaluation then it is purely a fixture update and this is a non-issue.

_Posted by an automated repo watch; please verify before acting._

CapThunder19:
Checked non-issue. Both `data_inputs` registration sites (`spawn_dataflow`,
`AddNode`) complete synchronously before a node can be marked connected, and
`self.running` is owned by a single `tokio::select!` loop with no concurrent path in.

PR #3031, opened 2026-08-04
fix(daemon,node): exit dora run nodes when the run process dies

### LINKED ISSUE #2856: `dora run` orphan containment: remaining gaps after #3018 — pre-`init` window, `path: shell` nodes, Windows Job Object

heyong4725:
Fixes #2856.

## The gap

SIGKILL of `dora run` vanishes the in-process coordinator+daemon without teardown. Nodes are process-group leaders by design (Ctrl-C isolation, per-node group kills in the stop ladder), so no inherited signal reaches them — and a node that is not polling its event stream never observes the daemon-socket EOF either: its event thread can be parked in `blocking_send` on a full channel, **upstream of the read that would see the EOF**. Such nodes are orphaned indefinitely (`ppid 1`, own pgid). Reproduced on macOS with current main, matching the Linux measurement on the issue.

## The containment

One mechanism for both Unix platforms, scoped to `dora run` only, per the direction on the issue:

- **Daemon**: the in-process local-run daemon injects its own pid into every node it spawns (`DORA_LOCAL_RUN_PID`). Derived from `exit_when_done` — the marker only the single-shot `dora run` path sets — so a coordinator-attached daemon injects nothing and its nodes keep surviving daemon restarts (#2029, `tests/daemon-reconnect-e2e.rs` semantics preserved).
- **Denylist**: the variable is control-plane-reserved. A descriptor cannot forge it (that would tie every node's lifetime to an arbitrary process) and a stale shell export cannot leak it into coordinator-attached spawns. Covered by the existing `reserved_control_plane_keys_are_denied` test plus a new injection test.
- **Node API**: a watchdog thread spawned from `init_with_options` — the choke point all init paths share, so custom nodes, the operator runtime, and the C/C++/Python bindings are all covered. Liveness is `kill(pid, 0)` at 1 Hz; on death: cooperative SIGTERM to self, 5 s grace, then hard exit — mirroring the stop ladder's escalation. Windows keeps its existing Job Object containment; the watchdog is Unix-only.

## The e2e caught a real defect before this PR was opened

The first version of the watchdog orphaned nodes anyway: its `eprintln!` panicked on EPIPE — the node's stderr is a pipe whose reader (the daemon) had just died — and the panic unwound the thread **before `process::exit` ran**. The new e2e (`run_killed_by_sigkill_does_not_orphan_nodes`) failed on exactly this, which is as organic a RED→GREEN as it gets: the test has demonstrably failed on a real containment bug and passes now. The death path is panic-proof: raw `write_all` with the error dropped, `catch_unwind` around logging.

## Measured (macOS 26 arm64)

| Scenario | Before | After |
|---|---|---|
| Non-polling, SIGTERM-ignoring node; `kill -9` CLI | alive at t+20 s and beyond (orphan) | **gone at t+6 s** (1 s detect + 5 s grace + exit) |
| Idiomatic polling node; `kill -9` CLI | gone at t+3 s (EOF path) | unchanged |

## Tests

- `run_killed_by_sigkill_does_not_orphan_nodes` (e2e): SIGKILLs the CLI, asserts the node is gone within 30 s. Uses the existing `sigterm-ignoring-node` fixture, which both ignores SIGTERM and never polls — so only the watchdog's hard-exit rung can end it.
- `the_local_run_pid_reaches_nodes_only_in_run_mode` (unit): the pid reaches nodes in run mode, a descriptor-forged value never lands, and coordinator-attached mode injects nothing.
- `reserved_control_plane_keys_are_denied` (existing): auto-covers the new denylist entry.

## Validation

- fmt clean, `clippy -D warnings` clean on touched crates (repo-standard targets)
- e2e + unit green locally; full workspace suite running; Linux CI dispatched on the branch
- Changelog entry under Unreleased → Fixed


heyong4725:
Closing as superseded by #3018 (`4eeb652ec`), which merged **before this work started** — I built against a stale checkout and never searched for PRs referencing #2856, so the whole implementation raced a fix that had already landed. The repro I posted on the issue was real but measured against pre-#3018 code; on actual current main the non-polling orphan is already contained.

Where the two differ, #3018 is stronger on every point:

- **`SIGKILL` to the node's own process group**, not `exit(1)` of the API process — under `--uv` the tracked process is the wrapper, and my version would have left the interpreter orphaned one level down, the exact shape of the original report.
- **No SIGTERM grace rung**, with the right argument: the first rung kills the process that would deliver the second, so grace is self-defeating once the daemon is gone.
- **Arms before anything in `init` can block**, rather than after the daemon handshake.
- The pre-`init` / `path: shell` / Windows gaps are **documented as gaps** instead of implicitly claimed covered.

One piece of this PR survives: #3018 shipped without a Changelog entry, so I'll send that separately, describing the merged mechanism.

PR #3036, opened 2026-08-05
perf(runtime): drop the per-message map alloc in the operator input buffer

phil-opp:
## Summary

`drop_oldest_inputs` runs on **every** event an operator receives. It rebuilt a `BTreeMap` keyed by every configured input (`effective_caps.len()` entries) on each call, regardless of how many events were actually queued. In steady state — consumer keeping up, deque holding 0–1 events, nothing dropped — an operator declaring N inputs still paid an **O(N) allocation plus N map inserts per message** for a scan that could never drop anything. For a fan-in node with many inputs at a high message rate this is pure per-message overhead.

## Fix

Only the just-pushed event's input can newly exceed its cap: every other input was already within cap, and one push can't change their counts. So the method now inspects just the newest input — look up its cap, count its live events in the (near-empty) deque, and return early when within cap. **The common path allocates nothing.**

The eviction path (over cap) is unchanged in outcome: drop the oldest surplus occurrences, keep the newest `cap` in FIFO order, then compact tombstones. `effective_cap` is always ≥ 1, so a single push can only put one input one over cap. The `id.clone()` needed to release the deque borrow before mutation runs **only on that rare over-cap path**, never in steady state.

## Validation

- New regression test `drop_oldest_inputs_caps_each_input_independently` asserts inputs are capped independently (pushing many of one input never drops another's) with FIFO order preserved.
- Existing `drop_oldest_inputs_keeps_deque_bounded_under_stall` and `compact_removes_tombstones_preserving_order` tests still pass.
- `cargo fmt` / `cargo clippy --all-targets -D warnings` / `cargo test -p dora-runtime` (channel tests) all pass.

## Note

⚠️ This is a **machine-generated** change, authored by Claude (an AI assistant) as part of an automated code-review pass. It has been reviewed for correctness, but please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FtaLn7JWcYzmYfb4ppgYYN)_

### DIFF
--- a/binaries/runtime/src/operator/channel.rs
+++ b/binaries/runtime/src/operator/channel.rs
@@ -100,47 +100,68 @@ impl InputBuffer {
     fn add_event(&mut self, event: Event) {
         self.queue.push_back(Some(event));
 
-        // drop oldest input events to maintain max queue length queue
+        // Cap the queue after *every* single push. `drop_oldest_inputs` relies
+        // on this cadence — it only re-checks the just-pushed input — so exactly
+        // one event must be enqueued between calls.
         self.drop_oldest_inputs();
     }
 
     fn drop_oldest_inputs(&mut self) {
-        let mut remaining: BTreeMap<&DataId, usize> = self
-            .effective_caps
-            .iter()
-            .map(|(k, (cap, _))| (k, *cap))
-            .collect();
-        let mut dropped = 0;
+        // Only the input of the just-pushed event can now exceed its cap: every
+        // other input was already within cap before this push (the invariant
+        // this method maintains on every `add_event`), and pushing a single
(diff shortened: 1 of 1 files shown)


phil-opp:

I verified the correctness argument the optimization rests on: `add_event` is the only caller of `drop_oldest_inputs` and enqueues exactly one event per call, so only the just-pushed input's count can newly exceed its cap — every other input was already within cap and one push can't change that. The deque is tombstone-free between calls (the drop path is the only tombstone source and it compacts immediately), so the single-input `live` count is accurate, and FIFO order is preserved. The steady-state path allocates nothing, as claimed. The new independent-capping test is non-vacuous and pins the intended per-input behavior.

---

---

Issue #3032, opened 2026-08-04
Memory leak: FREED_POOL_IDS grows unbounded in the Python node GPU memory-pool fast path


CapThunder19:
## Summary
FREED_POOL_IDS (a process-global HashSet<String> in apis/python/node/src/lib.rs) is used by the zero-copy GPU memory-pool fast path to detect read-after-free. Every call to free_memory_pool inserts an entry into this set, but nothing ever removes entries in normal operation the set grows for the entire lifetime of the process.

## Where
apis/python/node/src/lib.rs

Declaration: ~line 140-145
Insertion (never pruned in practice): ~lines 3320-3329, in free_memory_pool
The only removal path exists at ~lines 3866-3875, but it only fires when a receiver re-reads the exact same buffer_id later and buffer_id is format!("pool_{}_{}", node_id, pool_counter), where pool_counter comes from a monotonic counter that never repeats within a process. So the removal path essentially never triggers during normal streaming.
The struct's own doc comment already admits this:

/// Tracks freed pool buffer IDs so the DORADMA fast path can detect
/// read-after-free. Entries are inserted on free_memory_pool and never
/// pruned bounded in practice by the total number of registrations
/// over the process lifetime.
That "bounded in practice" claim doesn't hold for the workload this feature targets.

## Impact
For a node doing register  write  free once per frame (e.g. a camera or tensor-producing node at 30-60Hz, the exact zero-copy GPU use case this feature was built for), every frame permanently adds a new entry. At 30fps that's ~2.6M entries/day, growing memory usage by roughly 100+ MB/day with no upper bound a real risk of OOM in long-running production robotics/AI pipelines, which is the deployment scenario this feature and dora's fault-tolerance work are meant to support.

Notably, every other per-pool cache in the same function (PINNED_POOL, TRANSIT_META, RECV_GPU_VA, RECV_CPU_SHMEM, GPU_BUF_SIZES) is explicitly cleaned up on free, with comments calling out leak avoidance this one was just missed.

## Suggested fix
Bound FREED_POOL_IDS instead of letting it grow forever, e.g. replace it with an LRU/ring-buffer structure (indexmap::IndexSet + evict-oldest past a size threshold like a few thousand entries). The set only needs to catch reads shortly after a free, so unbounded retention isn't required for correctness.

CapThunder19:
@dora-bot assign me

phil-opp:
Closed by 3039

PR #3039, opened 2026-08-05
fix(python): bound FREED_POOL_IDS to stop unbounded memory growth in …

CapThunder19:
## Summary

`FREED_POOL_IDS` (a process-global set in `apis/python/node/src/lib.rs`) tracks freed GPU memory-pool buffer IDs so the DORADMA fast path can detect read-after-free. Every `free_memory_pool` call inserted into this set, but nothing ever removed entries in normal operation  the set grew for the entire lifetime of the process.

Because `buffer_id` is derived from a monotonic per-process counter that never repeats, a node doing register  write free once per frame (the exact zero-copy GPU streaming use case this feature targets, e.g. a camera node at 30-60Hz) added one permanent entry per frame, with no upper bound. At 30fps that's ~2.6M entries/day an unbounded memory leak in exactly the long-running production pipelines dora is meant to support.

This PR bounds `FREED_POOL_IDS` with a small FIFO-eviction wrapper (`FreedPoolIds`), capped at `FREED_POOL_IDS_CAP` (4096) entries, since only recently-freed buffers are relevant for read-after-free detection.

## Root cause

- `apis/python/node/src/lib.rs`, previously `FREED_POOL_IDS: LazyLock<Mutex<HashSet<String>>>`  insert-only, no eviction.
- The one removal path (clearing the tombstone on a successful fast-path read) only fires when a receiver re-reads the *exact same* `buffer_id`, which requires the same `node_id + counter` pair  effectively impossible during normal streaming since the counter never repeats.
- Every other per-pool cache in `free_memory_pool` (`PINNED_POOL`, `TRANSIT_META`, `RECV_GPU_VA`, `RECV_CPU_SHMEM`, `GPU_BUF_SIZES`) is explicitly cleaned up with comments calling out leak-avoidance this one was missed.

## Fix

Replaced the plain `HashSet<String>` with `FreedPoolIds`, a small struct pairing a `HashSet<String>` (O(1) membership) with a `VecDeque<String>` (insertion order), evicting the oldest entry once the set exceeds `FREED_POOL_IDS_CAP`. No new dependencies, no protocol/API changes scoped entirely to this one file.

## Testing

Added `freed_pool_ids_tests` (unit tests, co-located with the struct, matching this file's existing `seqlock_tests` convention):
- `insert_past_cap_does_not_grow_unbounded` regression test for the leak: inserting 4x the cap keeps `len()` bounded at exactly the cap.
- `insert_past_cap_evicts_oldest_first` confirms FIFO eviction (oldest gone, newest retained).
- `duplicate_insert_is_idempotent`  re-inserting a tracked ID doesn't inflate the count or eviction queue.
- `remove_drops_membership_and_order_entry`  covers the existing tombstone-clear path and re-insertion after removal.

#3032

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -137,12 +137,114 @@ static CUDA_HELPERS: LazyLock<std::sync::Mutex<Option<Py<PyModule>>>> =
 /// Counter to make pinned memory buffer IDs unique across registrations.
 static PINNED_COUNTER: LazyLock<std::sync::Mutex<u64>> = LazyLock::new(|| std::sync::Mutex::new(0));
 
+/// Maximum number of freed pool buffer IDs to remember at once. Only
+/// recently-freed buffers are useful for read-after-free detection, so
+/// once the cap is exceeded the oldest entries are evicted rather than
+/// keeping every ID for the life of the process.
+const FREED_POOL_IDS_CAP: usize = 4096;
+
 /// Tracks freed pool buffer IDs so the DORADMA fast path can detect
-/// read-after-free. Entries are inserted on free_memory_pool and never
-/// pruned — bounded in practice by the total number of registrations
-/// over the process lifetime.
-static FREED_POOL_IDS: LazyLock<std::sync::Mutex<HashSet<String>>> =
-    LazyLock::new(|| std::sync::Mutex::new(HashSet::new()));
+/// read-after-free. Bounded to `FREED_POOL_IDS_CAP` entries (oldest evicted
+/// first) so long-running nodes doing register->write->free every frame
+/// don't leak memory indefinitely.
+static FREED_POOL_IDS: LazyLock<std::sync::Mutex<FreedPoolIds>> =
+    LazyLock::new(|| std::sync::Mutex::new(FreedPoolIds::default()));
+
+/// Bounded, insertion-ordered set of freed pool buffer IDs. `set` gives
+/// O(1) membership checks; `order` tracks insertion order so the oldest
+/// entry can be evicted once `FREED_POOL_IDS_CAP` is exceeded.
+#[derive(Default)]
+struct FreedPoolIds {
+    set: HashSet<String>,
+    order: std::collections::VecDeque<String>,
+}
+
+impl FreedPoolIds {
+    fn insert(&mut self, id: String) {
+        if self.set.insert(id.clone()) {
+            self.order.push_back(id);
+            while self.order.len() > FREED_POOL_IDS_CAP {
+                if let Some(oldest) = self.order.pop_front() {
+                    self.set.remove(&oldest);
+                }
+            }
+        }
+    }
+
+    fn contains(&self, id: &str) -> bool {
+        self.set.contains(id)
+    }
+
+    fn remove(&mut self, id: &str) {
+        if self.set.remove(id) {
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed the bounded-tracking change. The `FreedPoolIds` wrapper preserves the exact method surface (`insert` / `contains` / `remove` / `len`) used by all four call sites, so the substitution is transparent and no caller depended on the old `HashSet::insert` bool return. Eviction is safe given buffer IDs are made unique per registration (counter suffix): an evicted old tombstone can never be re-queried by a later registration, so FIFO eviction of the oldest entries cannot cause a false-negative read-after-free on any live path. The four tests are meaningful — in particular `remove_drops_membership_and_order_entry` guards the stale-order-queue bug that a naive `remove` would introduce, and `insert_past_cap_does_not_grow_unbounded` is a real regression test for the original leak.

Minor observation only (not blocking): `remove`'s `order.retain(...)` is O(n), but it's gated on the ID actually being present, so it's rare and off the hot path.

---

---

phil-opp:
The eviction policy is sound and neither obvious hazard materialises: `FREED_POOL_IDS` is a read-after-free tombstone set rather than a free-list, so evicting an entry cannot cause a double free, and an evicted tombstone degrades to the same user-visible `warn_missing_memory_pool` path (the write fast path uses `ShmemConf::open`, not `create`). `remove`'s O(n) `order.retain` is gated behind `set.remove(id)` returning true, so the hot call stays O(1). Two notes:

1. **The four regression tests never execute.** `dora-node-api-python` is excluded from `cargo test` in `.github/workflows/ci.yml:229` and `nightly.yml:1192/1833`, and from clippy at `ci.yml:144`. The only job touching the crate is `cargo check --all`, which does not build `#[cfg(test)]` modules — so `freed_pool_ids_tests` is neither compiled nor run anywhere, and a compile error in it would ship. Pre-existing (the file's `seqlock_tests` are in the same position), but it means the regression test carries no enforcement.

2. **`FREED_POOL_IDS_CAP` is one global budget shared across every peer the process reads from**, so a 60Hz sender evicts a 1Hz sender's tombstones in about 68s. Consequence is benign per the above, but the doc comment implies a per-stream recency window it does not actually provide.

No collision with #3056 beyond a trivial textual one (adjacent lines); semantically #3056 strengthens this. No interaction with #3014.

phil-opp:

Re-checked at `09644ae`, the one docs-only commit since the last review: the `FREED_POOL_IDS` doc now states the cap is a single cross-peer budget rather than a per-stream recency window, closing the doc/behaviour mismatch noted earlier. No code change, and the `FreedPoolIds` logic still holds up — `set`/`order` stay in sync on insert (one eviction per insert, guarded by the `set.insert` bool) and on remove, and evicting a read-after-free tombstone only degrades a stale fast-path read to the existing `warn_missing_memory_pool`/daemon fallback, never a false-negative on a live buffer id (ids are counter-unique and never reused).

The one point still open is structural and pre-existing: `dora-node-api-python` is excluded from `cargo test` in `ci.yml`/`nightly.yml`, and `cargo check --all` doesn't build `#[cfg(test)]` modules, so `freed_pool_ids_tests` is neither compiled nor run in CI — a compile error in it would ship. Not introduced by this PR, but it means these regression tests currently carry no enforcement. No new issues found.

---

PR #3001, opened 2026-08-04
fix(coordinator): box CachedResult::Cached to fix large_enum_variant on Windows

phil-opp:
## Summary

Fixes #2979.

`CachedResult` in `binaries/coordinator/src/state.rs` had one large variant and one small one:

```rust
pub(crate) enum CachedResult {
    Pending { result_senders: Vec<oneshot::Sender<eyre::Result<ControlRequestReply>>> },
    Cached  { result: eyre::Result<ControlRequestReply> },  // large
}
```

`ControlRequestReply` is large, so `Cached` dwarfs `Pending`. Clippy's `large_enum_variant` heuristic flags this **on some targets (Windows) but not others (Linux)** for identical code, toolchain (1.97.1), and lockfile. Because PR CI only runs `clippy` on ubuntu-latest, `cargo clippy --all -- -D warnings` — the documented cross-platform pre-commit/pre-push gate — passed in CI yet failed for Windows contributors running the exact command the docs tell them to.

## Fix

Box the `result` field so `Cached` becomes a single pointer, making the two variants comparable in size:

```rust
    Cached { result: Box<eyre::Result<ControlRequestReply>> },
```

The two `matches!` arms that destructured the field (`is_terminal_error`, `is_cached_ok`) are rewritten with `.is_err()` / `.is_ok()` guards, since box deref-patterns are unstable. All construction sites now `Box::new(...)` the result. `send_result_to` is unchanged — `&Box<T>` deref-coerces to `&T` at the call sites.

## Verification

- `cargo clippy -p dora-coordinator --all-targets -- -D warnings` — passes (exit 0).
- `cargo test -p dora-coordinator --lib` — `110 passed; 0 failed`.

Note: because the lint only fires on Windows, the fix can't be observed to remove the diagnostic on this Linux runner, but it is a strict size reduction of the flagged variant and the behavior-preserving change is covered by the existing lib tests (including the reconstruction test that asserts on `CachedResult::Cached { result } if result.is_ok()`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01VSeehXFhKtMyFH67mSLBYt)_

PR #3040, opened 2026-08-05
fix(cli): normalize workspace path to forward slashes for CMake templates

### LINKED ISSUE #3038: dora new --lang c/cxx --internal-create-with-path-dependencies generates an unparseable CMakeLists.txt on Windows (backslash path breaks CMake string escaping)
## Problem
`workspace_dir()` in `binaries/cli/src/template/mod.rs:12-20` resolves the dora
workspace root from `env!("CARGO_MANIFEST_DIR")` and returns it as a raw OS path
string. On Windows that's backslash-separated (e.g. `C:\Users\...`). It's
substituted verbatim — no escaping — into the `__DORA_PATH__` placeholder in
the CMake templates:

- `binaries/cli/src/template/c/mod.rs:83,139`
- `binaries/cli/src/template/cxx/mod.rs:82,135`

producing a generated `CMakeLists.txt` line like:

    set(DORA_ROOT_DIR "C:\Users\alok0\Desktop\GSOC\dora\dora" CACHE FILEPATH "Path to the root of dora")

CMake treats `\` inside a quoted string as an escape character. `\U`, `\D`,
etc. are not recognized escapes, so `cmake -B build` fails immediately with a
parse error, before any compiler/generator logic runs. This only fires on the
`--internal-create-with-path-dependencies` path — the default `dora new` flow
substitutes an empty string and git-clones instead, so it never hits this.

## Evidence
Reproduced directly by feeding the exact substituted string to `cmake -P`:

    CMake Error ... Invalid character escape '\U'.

Fixed by using forward slashes (accepted by both CMake and Windows), which
parses cleanly. This matches `.github/workflows/nightly.yml:1602-1616`, which
exercises this exact `dora new --lang c/cxx --internal-create-with-path-dependencies`
+ `cmake -B` flow but is gated `if: runner.os == 'Linux'` — which is exactly
why this has never been caught (Linux paths are already forward-slash).

`docs/testing-matrix.md:191-209` explicitly acknowledges this class of
coverage gap in principle, but no concrete regression inside it appears to
be filed yet.

## Why it matters
Hits any Windows contributor smoke-testing the C/C++ node templates against
a local checkout with an opaque, unattributed CMake error. Also silently
blocks the day the Linux-only CI gate is lifted.

## Proposed fix
Normalize `workspace_dir()`'s return value to forward slashes before
substitution (2-file, ~6-line change in `template/mod.rs`, `c/mod.rs`,
`cxx/mod.rs`), plus a regression test — either asserting `workspace_dir()`
never contains `\`, or lifting the Windows gate on the nightly CMake step.

## Caveats
I verified the CMake parse failure directly (string substitution → `cmake -P`)
rather than running the literal `dora` binary end-to-end on Windows with an
MSVC toolchain — worth a quick real `dora new --lang c --internal-create-with-path-dependencies`
+ `cmake -B` sanity check to fully close the loop.

GuTS805:
## Problem

`workspace_dir()` in `binaries/cli/src/template/mod.rs` resolved the dora
workspace root from `env!("CARGO_MANIFEST_DIR")` and returned it as a raw OS
path. On Windows that's backslash-separated (e.g. `C:\Users\...`). This
string was substituted verbatim — no escaping — into the `__DORA_PATH__`
placeholder in the C/C++ CMake templates
(`binaries/cli/src/template/c/mod.rs`, `binaries/cli/src/template/cxx/mod.rs`),
producing a generated `CMakeLists.txt` line like:

    set(DORA_ROOT_DIR "C:\Users\alok0\...\dora" CACHE FILEPATH "Path to the root of dora")

CMake treats `\` inside a quoted string as an escape character. `\U`, `\D`,
etc. are not recognized escapes, so `cmake -B build` failed immediately with
a parse error (`Invalid character escape '\U'`), before any compiler/generator
logic ran. This only affects `dora new --lang c/cxx --internal-create-with-path-dependencies`
(the flag used to build templates against a local checkout) — the default
`dora new` flow substitutes an empty string and git-clones instead, so it
never hit this.

This is why it was never caught in CI: `.github/workflows/nightly.yml`
exercises this exact `dora new --lang c/cxx --internal-create-with-path-dependencies`
+ `cmake -B` flow, but is gated `if: runner.os == 'Linux'`, where paths are
already forward-slash.

## Fix

`workspace_dir()` now normalizes `\` to `/` before returning, since both
CMake and Windows accept forward slashes. Added a regression test
(`workspace_dir_has_no_backslashes`) asserting the returned path never
contains a backslash.

## Verification

- Built the actual `dora` CLI and ran
  `dora new test_c_project --lang c --internal-create-with-path-dependencies`
  — the generated `CMakeLists.txt` now contains
  `set(DORA_ROOT_DIR "C:/Users/.../dora" ...)` (forward slashes).
- `cmake -B build` on the generated project no longer hits the
  backslash-escape parse error (progresses past it to the `project()` step;
  the only remaining error is an unrelated missing MSVC/nmake toolchain on
  this machine).
- `cargo test -p dora-cli workspace_dir_has_no_backslashes` — pass

Fixes #3038

### DIFF
--- a/binaries/cli/src/template/c/mod.rs
+++ b/binaries/cli/src/template/c/mod.rs
@@ -80,7 +80,7 @@ fn create_cmakefile(root: PathBuf, use_path_deps: bool) -> Result<(), eyre::ErrR
     const CMAKEFILE: &str = include_str!("cmake-template.txt");
 
     let cmake_file = if use_path_deps {
-        CMAKEFILE.replace("__DORA_PATH__", super::workspace_dir()?)
+        CMAKEFILE.replace("__DORA_PATH__", &super::workspace_dir()?)
     } else {
         CMAKEFILE.replace("__DORA_PATH__", "")
     };
@@ -136,7 +136,7 @@ fn create_node_cmakefile(
     let cmake_content = if use_path_deps {
         NODE_CMAKE
             .replace("___name___", name)
-            .replace("__DORA_PATH__", super::workspace_dir()?)
+            .replace("__DORA_PATH__", &super::workspace_dir()?)
     } else {
         NODE_CMAKE
             .replace("___name___", name)
--- a/binaries/cli/src/template/cxx/mod.rs
+++ b/binaries/cli/src/template/cxx/mod.rs
@@ -79,7 +79,7 @@ fn create_cmakefile(root: PathBuf, use_path_deps: bool) -> Result<(), eyre::ErrR
     const CMAKEFILE: &str = include_str!("cmake-template.txt");
 
     let cmake_file = if use_path_deps {
-        CMAKEFILE.replace("__DORA_PATH__", super::workspace_dir()?)
+        CMAKEFILE.replace("__DORA_PATH__", &super::workspace_dir()?)
     } else {
         CMAKEFILE.replace("__DORA_PATH__", "")
     };
@@ -132,7 +132,7 @@ fn create_node_cmakefile(
     let cmake_content = if use_path_deps {
         NODE_CMAKE
             .replace("___name___", name)
-            .replace("__DORA_PATH__", super::workspace_dir()?)
+            .replace("__DORA_PATH__", &super::workspace_dir()?)
     } else {
         NODE_CMAKE
             .replace("___name___", name)
--- a/binaries/cli/src/template/mod.rs
+++ b/binaries/cli/src/template/mod.rs
@@ -17,23 +17,29 @@ fn workspace_dir() -> eyre::Result<String> {
         .context("Could not get manifest grandparent folder")?
         .to_str()
         .context("dora workspace path is not valid UTF-8")?;
-    // CMake treats `\` as a string escape character, so a raw Windows path
-    // (e.g. `C:\Users\...`) substituted into CMakeLists.txt fails to parse.
-    // Both CMake and Windows accept `/`, so normalize before substitution.
-    Ok(dir.replace('\\', "/"))
+    Ok(normalize_for_cmake(dir))
+}
+
+// CMake treats `\` as a string escape character, so a raw Windows path
+// (e.g. `C:\Users\...`) substituted into CMakeLists.txt fails to parse.
+// Both CMake and Windows accept `/`, so normalize before substitution.
+fn normalize_for_cmake(path: &str) -> String {
+    path.replace('\\', "/")
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    // Uses a hardcoded Windows-style sample path rather than the live
+    // `workspace_dir()` output, since on Linux/macOS CI runners the real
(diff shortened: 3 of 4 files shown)


phil-opp:
I reviewed both parts of this diff.

The CMake path normalization is correct: replacing `\` with `/` yields a path that CMake (which treats `\` as a string-escape character) and Windows both accept, and the C/C++ template callers are updated to borrow the now-owned `String`.

Two things worth raising:

1. **An unrelated change is bundled in.** Beyond the CMake templates, this PR also boxes `CachedResult::Cached` in `binaries/coordinator/src/state.rs` to silence a `clippy::large_enum_variant` warning. That is the *same* change already proposed on its own in **#3001**, so the two PRs overlap and will conflict, and per the repo convention ("Don't fix unrelated warnings in PRs") it doesn't belong in a CMake-path fix. I'd drop it here and let #3001 carry it (or close whichever is redundant). The boxing itself is correct and behavior-preserving — all `CachedResult::Cached` sites still compile (the `send_result_to(&result, …)` call happens before the value is boxed, and the `is_ok()`/`is_err()` guards work through the `Box`).

2. **The added test can't catch the regression on PR CI.** `workspace_dir_has_no_backslashes` is trivially true on Linux (Linux paths contain no `\`), so the Linux-only PR CI never exercises the `replace('\\', "/")` path — the fix is only really guarded on the Windows nightly job. Asserting the replacement against a constructed sample path that contains backslashes would give the test teeth on every platform.

---

---

GuTS805:
Thanks for the review. Addressed both points:

1. **Dropped the unrelated coordinator change.** Reverted the `CachedResult::Cached`
   boxing in `binaries/coordinator/src/state.rs` — you're right that it duplicates
   #3001, so I'll let that PR own it and keep this one scoped to the CMake path fix.

2. **Made the test OS-independent.** Pulled the normalization logic into a
   `normalize_for_cmake()` helper and changed the test to assert against a
   hardcoded Windows-style sample path (`C:\Users\example\dora`) instead of the
   live `workspace_dir()` output. It now actually exercises the `\` → `/`
   replacement on every platform instead of trivially passing on Linux CI.

Pushed both fixes. Diff is back to just the 3 template files.

phil-opp:
🤖 **Automated review by Claude** — fully automated review, not verified by a human.

Reviewed the latest commit (`b45176f`). Both points from the prior review are addressed: the unrelated `CachedResult::Cached` boxing in `binaries/coordinator/src/state.rs` is dropped, and the regression test now runs `normalize_for_cmake(r"C:\Users\example\dora")` against a hardcoded Windows-style path, so it exercises the `\` → `/` replacement on every platform instead of passing trivially on Linux CI. The diff is scoped to the three template files, and the `&super::workspace_dir()?` borrow updates match the new `String` return type. No new issues found.

---

PR #3042, opened 2026-08-05
fix(message): declare uuid/v4 feature used by dora-message

### LINKED ISSUE #3020: `dora-message` uses `Uuid::new_v4()` without declaring the `uuid/v4` feature, breaking `cargo bench -p dora-message`
`dora-message` uses `Uuid::new_v4()` without declaring the `uuid/v4` feature, so `cargo bench -p dora-message` fails to compile

## Summary

`dora-message`'s test/bench code calls `Uuid::new_v4()`, but neither the
workspace nor the crate declares the `uuid` `v4` feature. It compiles today only
because three *other* workspace crates enable `v4` and cargo unifies features
across a whole-workspace build. Any narrower selection loses that unification and
the crate stops compiling.

This is the same bug class as #2995 (`ZENOH_CONFIG_PATH_ENV` compiling only
because the `zenoh` feature happened to be on): code that builds by accident of
feature unification rather than by declaration.

## Evidence

Workspace declares `v4`-less features:

```toml
# Cargo.toml:165
uuid = { version = "1.23", features = ["serde", "v7"] }
```

Only these three crates opt in:

```
libraries/extensions/telemetry/tracing/Cargo.toml:27   features = ["v4"]
binaries/record-node/Cargo.toml:23                     features = ["v4"]
binaries/coordinator/Cargo.toml:24                     features = ["v4"]
```

But `dora-message` takes the bare workspace dep (`libraries/message/Cargo.toml:24`,
`uuid = { workspace = true }`) while using `new_v4()` in at least five places
(`libraries/message/src/common.rs:422,423,491,507`, `ws_protocol.rs:95`).

The CI benchmark job builds it in isolation:

```yaml
# .github/workflows/ci.yml — Benchmark regression check
cargo bench -p dora-message --no-run
cargo bench -p dora-daemon --features bench --no-run
```

`-p dora-message` excludes coordinator/tracing/record-node, so `v4` is never
activated and the build fails with 11 instances of:

```
error[E0599]: no associated function or constant named `new_v4` found
              for struct `uuid::Uuid` in the current scope
error: could not compile `dora-message` (lib test) due to 11 previous errors
```

Observed on the `trunk-merge/pr-2969` batch, job
https://github.com/dora-rs/dora/actions/runs/30932884643/job/92071666687 —
which failed the whole batch (#2969, #2970, #3018) under the misleading name
**"Benchmark regression check"**. It is a compile error, not a perf regression.

## Not caused by the batch

None of #2969, #2970, #3018 touch `libraries/message`, and the `new_v4()` call
sites date back to #2072 / #2043. The landmine was already there; the batch just
stepped on it.

**Open question:** why it started failing now rather than on earlier batches
today (the `pr-2940` batch passed this same check around 16:00). A `uuid`
1.23 → 1.24 resolution change is a candidate — the lockfile currently pins
1.24.0 — but that has not been confirmed. Worth pinning down, since it
determines whether anything else is newly exposed.

## Fix

One line — declare the feature where it is used:

```toml
# libraries/message/Cargo.toml
uuid = { workspace = true, features = ["v4"] }
```

Add a regression guard so the isolated build is actually gated: `cargo check -p
dora-message --tests` on its own would have caught this, and is cheap enough for
the PR tier. More broadly, the workspace has now produced two of these in a week
(#2995 and this) — a `cargo hack --each-feature check` or a
per-crate-isolated check would catch the class rather than the instance.

---
_Filed from an automated repo watch; verified against the failing job log and the
workspace manifests on 2026-08-04._

phil-opp:
## Summary

Fixes #3020.

`dora-message` calls `Uuid::new_v4()` in `libraries/message/src/common.rs` and `libraries/message/src/ws_protocol.rs`, but declared the bare workspace `uuid` dependency, which enables only `v7` and `serde`:

```toml
uuid = { workspace = true }
```

This PR declares the `v4` feature explicitly where it is used:

```toml
uuid = { workspace = true, features = ["v4"] }
```

## Why

This is the "compiles by accident of feature unification" bug class described in the issue. `new_v4()` is only available because some *other* crate in the build set enables `uuid/v4` and cargo unifies features across the whole build.

Note on reproduction: on the current `main`, `cargo bench -p dora-message --no-run` actually **passes**, because `uhlc` (a direct dependency of `dora-message`) transitively enables `uuid/v4`:

```
uuid v1.24.0
├── dora-message
└── uhlc v0.5.2   # uhlc's Cargo.toml: uuid = { features = ["v4"] }
    └── dora-message
```

So the crate depends on `uhlc` continuing to pull in `v4`. Anything that drops that transitive activation — e.g. the pending `uhlc` upgrade in #3016, or a narrower build selection — re-exposes the original failure:

```
error[E0599]: no associated function or constant named `new_v4` found for struct `uuid::Uuid`
```

Declaring `v4` directly removes that hidden dependency and makes the intent explicit, which is the correct fix regardless of which crate happens to activate the feature today.

## Validation

- `cargo +1.97.1 check -p dora-message --tests` — passes
- `cargo +1.97.1 bench -p dora-message --no-run` — passes
- No `Cargo.lock` change (the `v4` feature was already resolved).

---

### DIFF
--- a/libraries/message/Cargo.toml
+++ b/libraries/message/Cargo.toml
@@ -21,7 +21,11 @@ serde = { workspace = true }
 eyre = { workspace = true }
 arrow-schema = { workspace = true, features = ["serde"] }
 tokio = { workspace = true, features = ["sync"] }
-uuid = { workspace = true }
+# `v4` is used directly by this crate (see `Uuid::new_v4()` in `common.rs` /
+# `ws_protocol.rs`). Declared explicitly rather than relying on feature
+# unification from another crate, so an isolated build (`cargo bench -p
+# dora-message`) compiles on its own — see #3020.
+uuid = { workspace = true, features = ["v4"] }
 log = { version = "0.4.33", features = ["serde"] }
 aligned-vec = { version = "0.6.4", features = ["serde"] }
 semver = { workspace = true }


phil-opp:

Declaring `uuid`'s `v4` feature explicitly is correct dependency hygiene: `dora-message` calls `Uuid::new_v4()` directly (in `common.rs` / `ws_protocol.rs`), and relying on `uhlc`'s transitive activation plus cargo feature unification is fragile — it breaks an isolated build such as `cargo bench -p dora-message`. The feature is purely additive, so there is no semver or behavioral risk.

---

---

phil-opp:
The merge-queue `E2E Tests` failure here is a **10-minute wall-clock timeout, not a test failure**, and it is not caused by this PR's diff.

**What the log shows** ([job 93239514140](https://github.com/dora-rs/dora/actions/runs/31311339134/job/93239514140)):
- Every E2E test that ran printed `... ok`; nothing asserted false. The job ended with `##[error]The action 'Run E2E tests' has timed out after 10 minutes` right after `e2e_param_set_json_types_running_dataflow ... ok`.
- The `E2E Tests` step is `cargo test --test ws-cli-e2e -- --test-threads=1` with `timeout-minutes: 10` (`ci.yml:362-365`), and that step *includes* compilation. On this batch branch the build was cold and consumed ~1m32s of the 10-minute budget, so the serial suite ran out of time before finishing.

**Why it isn't this PR:** this change only adds `features = ["v4"]` to `dora-message`'s `uuid` dependency — a compile-time feature declaration with no runtime effect on the dataflow, so it cannot change E2E timing. The failing SHA (`1424de8f`, branch `trunk-merge/pr-3016/...`) is a **batch** that also bundles #3016 (uhlc 0.5→0.9). Notably, with uhlc 0.9 the transitive `uuid/v4` activation this crate relied on may be gone — so this PR's explicit declaration is what let `dora-message` compile in that batch (`Compiling uhlc v0.9.0` / `dora-message` both succeed in the log).

Nothing in this diff to fix, so I'm not pushing a change; leaving it for the merge queue to re-test (it typically re-tests PRs individually after a batch fails). I'll follow up if it recurs on a build isolated to this PR.

---

Issue #1856, opened 2026-05-18
zenoh session: required-config inserts now warn-and-continue, can silently partition (follow-up to #1855)


heyong4725:
## Context

Surfaced by Codex adversarial review on #1855 (rescue of #1608). Recording here so it doesn't get lost — out of scope for the rescue itself, which faithfully implemented PR 1608's panel-faithful panic→warn conversion.

## The shape of the regression

After #1855, `libraries/core/src/topics.rs::open_zenoh_session_with_listen` no longer panics when any `zenoh_config.insert_json5(...)` call fails. Each failure now logs a `warn!` and execution continues. For inserts that affect required transport topology, this turns a loud crash into a silent partition.

Concrete example — `DORA_ZENOH_CONNECT_ENV` injection path (`topics.rs:70-81`):

```rust
if let Ok(ep) = std::env::var(DORA_ZENOH_CONNECT_ENV) {
    if let Err(err) =
        zenoh_config.insert_json5("connect/endpoints", &format!(r#"["{ep}"]"#))
    {
        warn!("failed to set zenoh connect/endpoints from DORA_ZENOH_CONNECT: {err}");
    }
    if let Err(err) = zenoh_config.insert_json5("scouting/multicast/enabled", "false") {
        warn!("failed to disable zenoh scouting/multicast: {err}");
    }
}
```

If the first insert fails but the second succeeds (e.g., a malformed `ep` from a future daemon-injection bug), the node opens a Zenoh session with **no peer endpoint AND no multicast scouting** — it will keep running but the `>=4 KiB` data path is dead. The pre-#1855 `.unwrap()` would have crashed the daemon loudly.

The same shape exists for `listen/endpoints` (`topics.rs:83-87`) — the daemon could proceed without binding the listener it later injects into spawned nodes.

## Why we deferred fixing this in #1855

PR #1855 was scoped to faithfully rescue @LeonRust's #1608, which proposed a 1:1 `.unwrap()` → `warn!` conversion. Classifying each insert as required-vs-optional, and propagating `Err` for required-topology failures, is a broader policy decision that belongs in its own PR.

## Probability

Low today:
- `DORA_ZENOH_CONNECT_ENV` is set by the daemon itself, not by user input (see `topics.rs:65-69`) — a malformed value implies a daemon bug.
- `insert_json5("scouting/multicast/enabled", "false")` and `insert_json5("listen/exit_on_failure", "false")` are static strings — they cannot fail.
[...]

PR #3044, opened 2026-08-05
fix(core): bracket IPv6 coordinator address in zenoh connect/endpoints

### LINKED ISSUE #3041: IPv6 coordinator address is not bracketed when building zenoh `connect/endpoints`, producing a malformed peer endpoint
> It randomly selects a component and reviews it for correctness issues. It is machine-generated — please verify before acting. No human has confirmed this finding.

## Summary

In `open_zenoh_session_with_listen`, when a `coordinator_addr` is supplied, the zenoh `connect/endpoints` peer entry is built by interpolating the address with a bare `{}`. For an `IpAddr::V6`, `Display` emits the address **without brackets**, so the resulting endpoint string is a syntactically invalid zenoh TCP locator where the port colon is indistinguishable from the address colons (`tcp/::1:5456`). `insert_json5` then rejects it, the code falls through to the `warn!` branch, and the coordinator peer connect-endpoint is silently never configured.

The bug is IPv6-specific — IPv4 coordinator addresses format correctly.

**File:** `libraries/core/src/topics.rs` (lines ~401–411)

## The code

```rust
if let Some(addr) = coordinator_addr
    && let Err(err) = zenoh_config.insert_json5(
        "connect/endpoints",
        &format!(
            r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{}:5456"] }}"#,
            addr        // <-- IpAddr::V6 Display is unbracketed: "tcp/::1:5456"
        ),
    )
{
    warn!("failed to set zenoh connect/endpoints for coordinator {addr}: {err}");
}
```

`coordinator_addr` has type `Option<IpAddr>` (see the signature at `topics.rs:193`), so an IPv6 value is representable and reaches this branch unchanged.

Note the *router* entry on the same line — `tcp/[::]:7447` — **is** correctly bracketed, which shows the bracketing was simply omitted for the dynamic `peer` address.

## Verification

`IpAddr` vs `SocketAddr` `Display` differ exactly here (standalone Rust program, real output):

```
v6 bare:   tcp/::1:5456                    <-- malformed (what this code produces)
v4 bare:   tcp/127.0.0.1:5456              <-- fine
v6 socket: tcp/[::1]:5456                  <-- correct, via SocketAddr::new
v4 socket: tcp/127.0.0.1:5456
real bare: tcp/fd7a:1234:5678::1:5456      <-- malformed (e.g. a Tailscale ULA)
```

## Why this is the wrong behavior (same-file evidence)

The rest of the file already knows IPv6 must be bracketed and does it correctly by formatting through a `SocketAddr`:

- `reserve_zenoh_endpoint` (`topics.rs:559`): `format!("tcp/{}", SocketAddr::new(bind, port))` — brackets IPv6 correctly.
- The unit test `ipv6_endpoints_are_bracketed` (added in PR #2722) explicitly asserts `tcp/[::1]:` bracketing for that path.

Only this `connect/endpoints` call site formats the bare `IpAddr`, so it is inconsistent with the file's own convention and its own test's intent.

## Impact / severity

This is currently a **latent** bug: every in-tree caller of `open_zenoh_session` / `open_zenoh_session_with_listen` passes `None` for `coordinator_addr` today (`binaries/daemon/src/lib.rs`, `binaries/coordinator/src/ws_control.rs`, `apis/rust/node/src/node/mod.rs`), so the buggy branch is not exercised by the current binaries. However:

- `open_zenoh_session` is a **public API** of the `dora-core` crate (`pub async fn`), accepting `Option<IpAddr>`. Any external or future caller that passes an IPv6 coordinator address hits this and loses its explicit peer connect endpoint (falling back to multicast scouting, which may be disabled — cf. #1778).
- It's a silent failure: the only signal is a `warn!` about a rejected `insert_json5`, not an error.

No IPv4 deployment is affected.

## Suggested fix

Format the peer address through a `SocketAddr` (as `reserve_zenoh_endpoint` already does) so IPv6 is bracketed:

```rust
use std::net::SocketAddr;

let peer = SocketAddr::new(addr, 5456); // Display => "[::1]:5456" for IPv6, "127.0.0.1:5456" for IPv4
let endpoints = format!(
    r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{peer}"] }}"#
);
if let Err(err) = zenoh_config.insert_json5("connect/endpoints", &endpoints) {
    warn!("failed to set zenoh connect/endpoints for coordinator {addr}: {err}");
}
```

A small regression test covering an IPv6 `coordinator_addr` (asserting the generated `connect/endpoints` string contains `tcp/[::1]:5456`, not `tcp/::1:5456`) would pin it. This can be tested purely on the string formatting without opening a real session.

phil-opp:
## Summary

Fixes #3041.

In `open_zenoh_session_with_listen` (`libraries/core/src/topics.rs`), the zenoh `connect/endpoints` peer entry was built by interpolating the coordinator address with a bare `{}`:

```rust
format!(r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{}:5456"] }}"#, addr)
```

For an `IpAddr::V6`, `Display` emits the address **without** brackets, so the result is a malformed zenoh TCP locator — `tcp/::1:5456` — where the port colon is indistinguishable from the address colons. `insert_json5` rejects it, the code falls into the `warn!` branch, and the coordinator peer connect-endpoint is silently never configured. IPv4 addresses were unaffected. (Note the *router* entry on the same line, `tcp/[::]:7447`, was already bracketed — the bracketing was simply omitted for the dynamic `peer`.)

## Fix

Format the peer address through a `SocketAddr`, exactly as `reserve_zenoh_endpoint` already does — `SocketAddr`'s `Display` brackets IPv6 and leaves IPv4 bare. The endpoint-string construction is extracted into a small `coordinator_connect_endpoints(addr)` helper so it is unit-testable without opening a real session:

```rust
fn coordinator_connect_endpoints(addr: IpAddr) -> String {
    let peer = SocketAddr::new(addr, 5456);
    format!(r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{peer}"] }}"#)
}
```

The helper and its test are gated on the `zenoh` feature (matching the only call site, which is `#[cfg(feature = "zenoh")]`).

## Impact

Currently latent: every in-tree caller passes `None` for `coordinator_addr` today. But `open_zenoh_session` is public `dora-core` API accepting an `IpAddr`, so any external/future caller passing an IPv6 coordinator address hits this and silently loses its peer connect endpoint. Consistent now with the file's own `ipv6_endpoints_are_bracketed` convention.

## Validation (toolchain 1.97.1, `--features zenoh`)

- New test `topics::tests::coordinator_connect_endpoints_bracket_ipv6` — passes (asserts `tcp/[::1]:5456`, not `tcp/::1:5456`, for IPv6 and `tcp/127.0.0.1:5456` for IPv4)
- `cargo test -p dora-core --features zenoh` (existing `ipv6_endpoints_are_bracketed` etc.) — passes
- `cargo clippy -p dora-core --features zenoh --all-targets -- -D warnings` — clean

---

### DIFF
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -184,6 +184,20 @@ pub async fn open_zenoh_session(coordinator_addr: Option<IpAddr>) -> eyre::Resul
 /// shared by the caller (e.g. `dora cluster up`), not per-daemon
 /// state to advertise back to nodes.
 ///
+/// Builds the zenoh `connect/endpoints` JSON5 for a coordinator peer.
+///
+/// The peer address is formatted through a [`SocketAddr`] so that IPv6
+/// addresses are bracketed (`tcp/[::1]:5456`), matching zenoh's TCP locator
+/// grammar. Interpolating a bare [`IpAddr`] instead would emit `tcp/::1:5456`
+/// for IPv6 — a malformed locator where the port colon is indistinguishable
+/// from the address colons, which `insert_json5` rejects (#3041). This is the
+/// same bracketing [`reserve_zenoh_endpoint`] already relies on.
+#[cfg(feature = "zenoh")]
+fn coordinator_connect_endpoints(addr: IpAddr) -> String {
+    let peer = SocketAddr::new(addr, 5456);
+    format!(r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{peer}"] }}"#)
+}
+
 /// `multicast` lets a caller that establishes every link explicitly opt out of
 /// scouting — see [`DORA_ZENOH_MULTICAST_ENV`], honored in addition to this
 /// argument. It is a request, not a command: it is ignored unless this session
@@ -399,13 +413,8 @@ pub async fn open_zenoh_session_with_listen(
             }
 
             if let Some(addr) = coordinator_addr
-                && let Err(err) = zenoh_config.insert_json5(
-                    "connect/endpoints",
-                    &format!(
-                        r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{}:5456"] }}"#,
-                        addr
-                    ),
-                )
+                && let Err(err) = zenoh_config
+                    .insert_json5("connect/endpoints", &coordinator_connect_endpoints(addr))
             {
                 warn!("failed to set zenoh connect/endpoints for coordinator {addr}: {err}");
             }
@@ -838,6 +847,29 @@ mod tests {
                 ) || e.raw_os_error() == Some(97) => {}
             Err(e) => panic!("unexpected error reserving ::1 endpoint: {e}"),
         }
+    }
+
+    // The coordinator peer endpoint must bracket IPv6 too, or `insert_json5`
+    // rejects the malformed locator and the peer connect-endpoint is silently
+    // dropped (#3041). Pure string formatting — no session is opened.
+    #[cfg(feature = "zenoh")]
+    #[test]
+    fn coordinator_connect_endpoints_bracket_ipv6() {
+        let v6 = coordinator_connect_endpoints(IpAddr::V6(std::net::Ipv6Addr::LOCALHOST));
+        assert!(
+            v6.contains(r#"peer: ["tcp/[::1]:5456"]"#),
+            "IPv6 coordinator peer must be bracketed, got {v6}"
+        );
+        assert!(
+            !v6.contains("tcp/::1:5456"),
+            "unbracketed IPv6 peer locator is malformed, got {v6}"
+        );
+
+        let v4 = coordinator_connect_endpoints(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)));
+        assert!(
+            v4.contains(r#"peer: ["tcp/127.0.0.1:5456"]"#),
+            "IPv4 coordinator peer must be unbracketed, got {v4}"
+        );
     }
 
     // The filter behind the routing lookup, tested directly rather than through
--- a/libraries/core/src/topics.rs
+++ b/libraries/core/src/topics.rs
@@ -154,6 +154,20 @@ pub async fn open_zenoh_session(coordinator_addr: Option<IpAddr>) -> eyre::Resul
     Ok(session)
 }
 
+/// Builds the zenoh `connect/endpoints` JSON5 for a coordinator peer.
+///
+/// The peer address is formatted through a [`SocketAddr`] so that IPv6
+/// addresses are bracketed (`tcp/[::1]:5456`), matching zenoh's TCP locator
+/// grammar. Interpolating a bare [`IpAddr`] instead would emit `tcp/::1:5456`
+/// for IPv6 — a malformed locator where the port colon is indistinguishable
+/// from the address colons, which `insert_json5` rejects (#3041). This is the
+/// same bracketing [`reserve_zenoh_endpoint`] already relies on.
+#[cfg(feature = "zenoh")]
+fn coordinator_connect_endpoints(addr: IpAddr) -> String {
+    let peer = SocketAddr::new(addr, 5456);
+    format!(r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{peer}"] }}"#)
+}
+
 /// Like [`open_zenoh_session`], but also configures the session to listen on
 /// the given endpoint (e.g. `tcp/127.0.0.1:43217`, or a routable address such as
 /// `tcp/10.0.2.100:43217` for a daemon in a cluster). The daemon uses this so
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed the IPv6 bracketing fix. Routing the peer string through `SocketAddr::new(addr, 5456)` is the right approach — `SocketAddr`'s `Display` brackets IPv6 (`tcp/[::1]:5456`) and leaves IPv4 bare, which matches the zenoh TCP locator grammar that previously silently rejected the malformed `tcp/::1:5456`. The required imports (`IpAddr`, `Ipv4Addr`, `SocketAddr`) are already in scope, the port and router endpoint are unchanged, and the `#[cfg(feature = "zenoh")]` gating is consistent between the function and its test. The test asserts both the presence of the bracketed form and the absence of the malformed one, so it would catch a regression to bare interpolation.

---

---

phil-opp:
**Automated review by Claude — this is a fully automated review.**

The IPv6 bracketing fix itself is correct, and the added `coordinator_connect_endpoints_bracket_ipv6` test genuinely pins it (it would fail if the peer string went back to bare interpolation). I did find one issue this change introduces, though:

The new `coordinator_connect_endpoints` helper (with its own doc comment) is inserted **in the middle of the doc-comment block for the public `open_zenoh_session_with_listen`** — between the `/// … state to advertise back to nodes` line and the `/// `multicast` lets a caller …` paragraph in `libraries/core/src/topics.rs`. Because rustdoc attaches a run of preceding `///` lines to the next item, the first (and largest) part of that block — the summary, the `inter_daemon_peer` rendezvous explanation, and the `(session, effective_listen_endpoint)` return-value semantics (#1856/#1858) — now documents the private helper instead. `open_zenoh_session_with_listen` is left documented only by the trailing `multicast` paragraph.

The code compiles and behaves correctly; this is purely a docs regression on a public function. Moving the helper (with its doc) above the `open_zenoh_session_with_listen` doc block (before line 157) or below the function body would keep both docs intact.

---

libraries/core/src/topics.rs
```diff
@@ -184,6 +184,20 @@ pub async fn open_zenoh_session(coordinator_addr: Option<IpAddr>) -> eyre::Resul
...
+/// grammar. Interpolating a bare [`IpAddr`] instead would emit `tcp/::1:5456`
+/// for IPv6 — a malformed locator where the port colon is indistinguishable
+/// from the address colons, which `insert_json5` rejects (#3041). This is the
+/// same bracketing [`reserve_zenoh_endpoint`] already relies on.
+#[cfg(feature = "zenoh")]
+fn coordinator_connect_endpoints(addr: IpAddr) -> String {
+    let peer = SocketAddr::new(addr, 5456);
+    format!(r#"{{ router: ["tcp/[::]:7447"], peer: ["tcp/{peer}"] }}"#)
+}
+
```
phil-opp: This is inserted within a doc comment, must be fixed

phil-opp: Good catch — thanks. Fixed in 74744d5: moved `coordinator_connect_endpoints` (with its own doc comment) to sit *before* `open_zenoh_session_with_listen`'s doc comment instead of splitting it, so that function's `/// Like [open_zenoh_session], but also configures…` doc is whole and attached to the right item again. No behavior change; `cargo fmt --check` and the `coordinator_connect_endpoints_bracket_ipv6` test still pass.

---

phil-opp:

Re-reviewed the new commit (`74744d51`), which post-dates the previous review. It resolves the doc-comment regression flagged earlier: `coordinator_connect_endpoints` (with its own doc) now sits after `open_zenoh_session`'s closing brace and ahead of `open_zenoh_session_with_listen`'s doc block, so that public function's full doc comment (summary, `inter_daemon_peer` rendezvous, `(session, effective_listen_endpoint)` return semantics) is attached to it again instead of to the private helper. The IPv6-bracketing fix and its `coordinator_connect_endpoints_bracket_ipv6` test are unchanged and still correct. No new issues.

---

phil-opp:
The merge-queue `Test (ubuntu-latest)` failure is a **cross-PR compile conflict in the batch, not caused by this PR**.

**What the log shows** ([job 93806983920](https://github.com/dora-rs/dora/actions/runs/31499935493/job/93806983920)):
```
```
This is in `binaries/coordinator` — one batched PR added the `ready_barrier_released` field to `RunningDataflow` while another batch member has a `RunningDataflow { … }` initializer at `handlers.rs:911` that doesn't set it. That mismatch only exists when those PRs are combined.

**Why it isn't this PR:** #3044 changes exactly one file, `libraries/core/src/topics.rs` (`git diff --name-only origin/main...HEAD` → just that file); it doesn't touch `binaries/coordinator` at all. On this branch the coordinator is byte-identical to `main`: there's a single `RunningDataflow` initializer (`handlers.rs:658`) and it *does* set `ready_barrier_released`. The failing line `handlers.rs:911` doesn't even exist on this branch — it's from another batch member's version of the file. The batch was tested on SHA `8cb4317` (`trunk-merge/pr-2931/…`), not this PR's head.

Nothing in this diff to fix. The queue is already re-sequencing #3044 behind the failed batch, so it'll be re-tested on its own — expecting green there.

---

PR #3045, opened 2026-08-05
test(daemon): deflake a_wrapper_that_exits_does_not_hide_its_surviving_child

### LINKED ISSUE #3023: test(daemon): `a_wrapper_that_exits_does_not_hide_its_surviving_child` races its own wrapper and flakes under load
`shutdown::tests::a_wrapper_that_exits_does_not_hide_its_surviving_child` (added by #3004, `binaries/daemon/src/shutdown.rs`) fails intermittently under load. It passes reliably in isolation, so it does not show up in a focused `cargo test -p dora-daemon <name>` loop — it shows up in `cargo test --all`, and can hit CI.

```
assertion `left == right` failed: the wrapper is alive, so this is an ordinary wait
  left: Done
 right: Waiting
binaries/daemon/src/shutdown.rs:281
```

## Root cause

The test spawns a wrapper that is *designed* to exit immediately:

```rust
let mut wrapper = Command::new("sh")
    .arg("-c")
    .arg("sleep 300 & exit 0")
    ...
let mut wait = DestroyWait::new();
assert_eq!(wait.poll(&[pid]), DestroyProgress::Waiting,
           "the wrapper is alive, so this is an ordinary wait");
```

Nothing orders the first `poll` before the wrapper exits — the assertion just assumes it wins the race. When it loses:

- `sh` has exited and is an unreaped zombie (the test only calls `wrapper.wait()` later).
- `is_live_child` returns `false` for `ProcessStatus::Zombie` (`shutdown.rs:214-219`).
- This is the *first* poll, so `self.confirmed` is still empty and the `else if self.confirmed.contains(&pid) && process_group_has_members(pid)` fallback in `live_children` does not apply.
- `alive` is empty → `poll` returns `Done` instead of `Waiting`.

So it is purely a test-side race. The production behaviour under test — a group outliving its leader — is fine; the test just fails to establish the precondition it asserts.

## Reproduction

Not reproducible by running the test alone (0/25 full-suite runs on an idle machine), because the first poll normally wins. Run the daemon lib test binary concurrently with itself to starve the test thread:

```bash
cargo test -p dora-daemon --lib          # build
BIN=$(ls -t target/debug/deps/dora_daemon-* | grep -v '\.d$' | head -1)
for round in 1 2 3; do
  for j in $(seq 1 12); do ( $BIN 2>&1 | grep -E "^test .* FAILED" ) & done
  wait
done
```

Observed **9 failures out of 36 runs** (~25%), always this test, always `left: Done, right: Waiting`.

## Suggested fix

Give the wrapper a deterministic live window: block it on stdin so it cannot exit until the test says so, then close the pipe where the test currently expects the exit.

```rust
 let mut wrapper = Command::new("sh")
     .arg("-c")
-    .arg("sleep 300 & exit 0")
+    .arg("sleep 300 & read line")
+    .stdin(Stdio::piped())
     .stdout(Stdio::null())
     ...

 // The wrapper exits and is reaped; its child lives on.
+drop(wrapper.stdin.take());
 wrapper.wait().expect("failed to wait for wrapper");
```

The wrapper is now provably alive for the first `poll` (which is also what lets `confirmed` record the pid, so the group fallback covers the second poll as intended), and closing stdin makes `read` hit EOF so `sh` exits exactly where the test wants it to.

Verified locally against the reproduction above: **0 failures out of 36 runs** with this change, vs 9/36 without.

Found while running the pre-push gate for #2971; not filed there because it is unrelated to that diff.

phil-opp:
## Summary

Fixes #3023.

`shutdown::tests::a_wrapper_that_exits_does_not_hide_its_surviving_child` (`binaries/daemon/src/shutdown.rs`) fails intermittently under load — it passes in isolation but flakes in `cargo test --all` and can hit CI:

```
assertion `left == right` failed: the wrapper is alive, so this is an ordinary wait
  left: Done
 right: Waiting
```

## Root cause

The test spawns a wrapper *designed* to exit immediately (`sleep 300 & exit 0`), then asserts the first `DestroyWait::poll` returns `Waiting`. Nothing orders that poll before the wrapper exits. When the poll loses the race:

- `sh` has exited and is an unreaped zombie (the test only calls `wrapper.wait()` later).
- `is_live_child` returns `false` for a zombie.
- It's the *first* poll, so `self.confirmed` is empty and the `confirmed && process_group_has_members` fallback in `live_children` does not apply.
- `alive` is empty → `poll` returns `Done` instead of `Waiting`.

Pure test-side race; the production behavior under test (a group outliving its leader) is fine.

## Fix

Give the wrapper a deterministic live window: block it on stdin (`read`) instead of `exit 0`, and close the pipe where the test previously expected the exit. The wrapper is now provably alive for the first poll (which also records the pid so the group fallback covers the second poll), and closing stdin makes `read` hit EOF so `sh` exits exactly on cue.

## Validation

- `cargo test -p dora-daemon --lib a_wrapper_that_exits` — passes.
- Stress reproduction from the issue (12-way concurrent runs × 3 rounds of the daemon lib test binary): **0 failures / 36 runs** on this branch (issue reports ~9/36 without the fix).

---

### DIFF
--- a/binaries/daemon/src/shutdown.rs
+++ b/binaries/daemon/src/shutdown.rs
@@ -275,9 +275,19 @@ mod tests {
 
         // A group leader that spawns a child into its group and exits, like
         // a wrapper dying while the interpreter it launched keeps running.
+        //
+        // The wrapper blocks on `read` rather than `exit 0` so its lifetime is
+        // controlled by the test, not by a race: nothing ordered the first
+        // `poll` before a bare `exit 0`, so on a loaded machine the wrapper
+        // could already be a reaped-later zombie by then, `is_live_child`
+        // returns false, and — being the first poll — the `confirmed`-group
+        // fallback does not yet apply, so `poll` returned `Done` and the
+        // "wrapper is alive" assertion flaked (#3023). Closing stdin below makes
+        // `read` hit EOF, so the wrapper exits exactly where the test wants it.
         let mut wrapper = Command::new("sh")
             .arg("-c")
-            .arg("sleep 300 & exit 0")
+            .arg("sleep 300 & read line")
+            .stdin(Stdio::piped())
             .stdout(Stdio::null())
             .stderr(Stdio::null())
             .process_group(0)
@@ -292,7 +302,9 @@ mod tests {
             "the wrapper is alive, so this is an ordinary wait"
         );
 
-        // The wrapper exits and is reaped; its child lives on.
+        // Close stdin so the wrapper's `read` hits EOF and it exits; its child
+        // lives on. The wrapper is reaped by the `wait()` below.
+        drop(wrapper.stdin.take());
         wrapper.wait().expect("failed to wait for wrapper");
         assert_eq!(
             wait.poll(&[pid]),


phil-opp:

I reviewed the deflake change. The fix is correct: `Stdio::piped()` guarantees `wrapper.stdin` is populated, dropping the taken `ChildStdin` closes the write end so the wrapper's `read` hits EOF and exits on cue, and the backgrounded `sleep 300` remains in the same process group for the second poll. This removes the race where the wrapper could already be a zombie before the first poll, so the "wrapper is alive" first-poll assertion now deterministically exercises the live-child path.

---

---

PR #3047, opened 2026-08-05
feat(node-api): non-blocking correlated receives for service and action clients

### LINKED ISSUE #3046: Make the Service/Action client usable from an async, multiplexing C++ node
**Component:** `apis/c++/node` (cxx bridge), `apis/rust/node` (`EventStream`)
**Baseline:** `main` @ `cd597e70`

## Summary

#2853 exposed the Service and Action patterns to the C++ node binding. The **server
side** is complete and works well. The **client side** is not usable from a C++ node
that (a) cannot block and (b) has more than one request in flight.

Three independent gaps:

1. **The only client-side receive is blocking.** Rust has `pub async fn
   recv_service_response`, which an async client can compose. The cxx bridge flattens
   it to a blocking call, and there is no `try_recv` / poll variant for service or
   action in **either** API.
2. **`send_service_request` cannot take a caller-supplied `request_id`.** It always
   mints a fresh one, so one logical request fanned out to N nodes cannot share a
   correlation id.
3. **Correlation is bound to exactly one expected server.**
   `wait_for_correlation(timeout, expected_server: &NodeId, …)` cannot await
   "whichever of these N nodes answers first".

Together these mean a C++ client must re-implement request correlation, timeouts, and
failure handling by hand — which is exactly what the pattern is meant to provide.

## Why blocking is not workable for this class of client

The pattern currently assumes a client that can afford to stop and wait. A common
shape cannot:

- **Single-threaded.** `dora::Node` is documented as not thread-safe, so all node I/O
  happens on one thread. A blocking wait cannot be moved off it.
- **Tick-driven.** The node has its own periodic work — draining an external
  transport, bridging telemetry, emitting heartbeats — that must keep running.
- **Multiplexing.** Several requests can be outstanding at once, targeting different
  nodes.

For such a node, `recv_service_response` stalls everything else for the duration of
the wait. On the happy path that is roughly one tick period of the destination node;
on the failure path — a node that is wedged rather than restarted — it burns the
entire timeout. Buffering non-matching events (which the API does correctly) prevents
data *loss*, but not the stall.

The workaround is a hand-rolled pending map keyed by correlation id, with per-request
deadlines swept on the tick and cancellation driven by `NodeFailed`.

### 1. No non-blocking client receive

- `apis/rust/node/src/event_stream/mod.rs:1178` — `pub async fn recv_service_response`;
  `:1209` — `pub async fn recv_action_result`. Async, therefore composable in Rust.
- `apis/c++/node/src/lib.rs:1237` / `:1252` — the cxx equivalents. Blocking; the
  async-ness is gone.
- `EventStream::try_recv` exists for plain events (`mod.rs:1086`), but there is **no**
  `try_recv_service_response` / `try_recv_action_result` in either API.

**The C++ binding already establishes the convention this would follow.**
`try_next_event` (`apis/c++/node/src/lib.rs:496`) is a non-blocking single poll
returning `DoraEventType::Empty` when nothing is ready, explicitly distinguished from
a real timeout. This request is not asking for a new concept — only to extend an
existing one from plain events to the correlated receives.

### 2. `send_service_request` always mints its own id

```rust
fn send_service_request(sender, output_id, data, metadata) -> DoraRequestId {
    let mut parameters = (*metadata).into_parameters();
    let request_id = insert_request_id(&mut parameters);   // always fresh; overwrites
    ...
}
```

A request fanned out to several nodes — where the first reply resolves it — needs all
publishes to carry the **same** `request_id`. With this API each call mints a
different one.

The workaround is to bypass the helper: `new_request_id()` →
`Metadata::set_request_id()` → `send_output_with_metadata()` per destination. It
works, but it makes the ergonomic entry point unusable for any fan-out, and requires
knowing that the two paths are interchangeable except for id handling.

### 3. Correlation is pinned to a single server

```rust
async fn wait_for_correlation<F>(&mut self, timeout: Duration, expected_server: &NodeId, …)
```

(`apis/rust/node/src/event_stream/mod.rs:1248`)

For the fan-out case the responder is "whichever node answers first"; a single
`expected_server` cannot express that. This makes the receive API unusable for any
broadcast / any-of pattern, independent of the blocking issue.

## What we are asking for

Ordered by impact.

1. **A non-blocking client receive**, following the existing `try_next_event`
   convention:

   ```
   try_recv_service_response(events, request_id, server) -> Option<Event>
   try_recv_action_result(events, goal_id, server)       -> Option<Event>
   ```

   Equally good alternatives: exposing the Rust `async` form to C++, or a
   completion-callback registration (`on_service_response(request_id, cb)`) driven by
   the existing event loop.

2. **Let the caller supply the `request_id`** — either a
   `send_service_request_with_id(…)` variant, or have `send_service_request` respect a
   `request_id` already present in the passed `Metadata` instead of overwriting it.
   The doc comment currently documents the overwrite as intentional, so this would be
   a deliberate change, not a bug fix.

3. **Allow correlation without a single fixed server** — accept a set of acceptable
   responders, or allow "any". If the `expected_server` check exists to prevent
   cross-talk, a set preserves that intent.

Item 1 alone is enough to drop caller-side deadline handling and adopt the framework's
timeout and restart behavior. Items 2 and 3 additionally remove fan-out bookkeeping.

## Non-goals

- Not asking for threads or an async runtime in the C++ binding. A poll that returns
  "not ready" is sufficient for a fixed-rate event loop.
- Not asking to change the server side. It works.

## Acceptance criteria

- [ ] A single-threaded C++ node can have N service requests outstanding and keep
      making progress on its own event loop without blocking on any of them.
- [ ] A C++ node can issue one logical request to several destination nodes under one
      `request_id` and correlate whichever reply arrives first.
- [ ] Timeout and server-restart handling come from the framework rather than
      caller-side bookkeeping.
- [ ] The existing blocking `recv_service_response` keeps working unchanged for simple
      clients.

## References

- #2853 — exposed Service and Action to the C++ binding (merged; the work this builds on)
- `docs/patterns.md` §2 (Service), §3 (Action)
- `examples/c++-service-action/` — both nodes are simple request/reply; neither is a
  multiplexing client
- Related C++/Rust parity gap: the pinned/CUDA memory-pool transport (#2168, #2386) is
  also Rust/Python-only

---

Happy to implement any or all of this if the shape sounds right. Item 1 is where we
would start.

caothu2k1:
Closes #3046.

## What

#2853 gave the C++ binding the Service and Action patterns. The server side
was complete; the client side was only usable from a node that can afford to
stop and wait. This adds the three pieces #3046 asked for, all additive — no
existing signature changes.

### 1. Non-blocking correlated receive

| Rust | C++ |
|------|-----|
| `EventStream::try_recv_service_response` | `try_recv_service_response(...)` |
| `EventStream::try_recv_action_result` | `try_recv_action_result(...)` |

Returns `Ok(None)` / `DoraPatternStatus::NotReady` instead of waiting.
Buffering, restart detection, correlation **and the deadline** are all
handled as in the blocking form; only the waiting is gone.

The framework owns the timeout. The first poll carrying one registers a
deadline against the correlation id, and a later poll past it returns
`Timeout` exactly once — the registration is dropped as it expires, and also
on a match or terminal error. A caller passes the same timeout every
iteration and reacts to `Timeout` like any other outcome; it does not sweep
deadlines itself. Rust takes `Option<Duration>`; C++ spells "no deadline" as
`timeout_ms == 0`.

Decisions worth a look:

- **The clock starts at the first deadline-carrying poll, not at send time.**
  `EventStream` does not see the send, and threading registration through
  `DoraNode` would couple the two for no practical gain — a client polls in
  the same iteration it sends.
- **The first deadline registered for an id wins.** Otherwise a caller
  recomputing a relative timeout each iteration would push its own deadline
  forward forever and never expire.
- **A reply that arrived before the deadline always wins over it**, whether it
  was already buffered or still unread on the stream — the deadline is
  consulted only after everything available has been examined. So `Timeout`
  means the reply had genuinely not arrived, not that the caller polled late.
- `cancel_correlation` (Rust **and** C++) drops a registration for a request
  that will never be polled again — a peer died, the operator cancelled, the
  reply stopped mattering. Without it that entry lives until the stream is
  dropped, which on a node running for days is an unbounded slow leak. It is
  a no-op for an unknown id, so a cancel racing a reply is harmless.
- **`timeout_ms == 0` means "no deadline", not "return immediately."** That
  is right here — a poll never blocks, so there is nothing to time out — but
  it inverts the usual convention for a timeout argument, so both docs now
  say so explicitly. Flagging it in case a different spelling is preferred.
- For actions the timeout bounds the *whole goal*, not the gap between
  feedback messages.

`NotReady` follows the convention `try_next_event` already established for
plain events, and stays distinct from `Timeout`: "not yet" and "too late"
are different answers, and only the second is terminal.

### 2. Caller-supplied `request_id`

| Rust | C++ |
|------|-----|
| `DoraNode::send_service_request_with_id` | `send_service_request_with_id(...)` |

`send_service_request` mints a fresh id per call, so it cannot express one
logical request fanned out to several nodes. It now delegates to the new
variant and its warning points at it. Existing behaviour is unchanged.

### 3. Correlation without a single fixed server

New `ExpectedServers` enum — `One(&NodeId)` / `AnyOf(&[NodeId])` / `Any` —
plus `recv_service_response_from`, `recv_action_result_from` and the `_from`
polls. In C++ these take a `Vec<String>`; empty means "any".

The existing `recv_service_response(request_id, &NodeId, timeout)` delegates
to `ExpectedServers::One`, so nothing breaks.

Two semantics worth checking in review:

- **The server set governs only restart detection.** Which reply matches is
  decided by `request_id` alone, so a fan-out reply is correlated even from a
  node the caller never listed. There is a test pinning this.
- **`AnyOf` reports a restart of any listed node**, naming that node (taken
  from the event, not the caller's expectation). For a first-reply-wins
  fan-out that is a notification rather than a verdict — the other candidates
  may still answer — so the caller can simply wait again. Buffered events,
  the correlation **and its deadline** survive across calls.
- **What counts as `One` is decided by candidate count, not spelling.** The
  C++ single-server polls wrap their lone id in a slice, and a one-element
  `Vec` passed to a `_from` variant is the same thing — both resolve to
  `One`, so their restart releases the deadline. Routing them through a
  one-element `AnyOf` (as an earlier revision did) kept the entry alive and
  leaked it on the give-up path.
- **A restart therefore affects the deadline differently per variant.** For
  `One` it is terminal: the request is orphaned and its deadline is dropped
  with it. For `AnyOf` the deadline keeps running on its *original* clock. An
  earlier revision dropped it in both cases, which meant a caller following
  the documented "just wait again" advice re-registered a fresh deadline
  every poll — so a flapping node could hold the correlation open forever and
  `Timeout` would never fire. Caught in review; both branches are now pinned
  by tests.

## One thing that came out of building the example

The polls and the caller's own `next_event` read the **same** stream, so
whichever runs first consumes what is there. A poll correlates the reply it
wants and buffers everything else, so polling first loses nothing. The
reverse is not true: a reply handed to `next_event` is gone, and no later
poll can see it.

My first version of the example got this wrong and silently timed out every
request. That ordering rule is now documented in `docs/patterns.md`, the C++
README, and prominently in the example itself. It may be worth a maintainer's
eye — it is inherent to polling a single shared stream rather than something
this PR introduces, but it is a sharp edge.

## Why the deadline part is here

The first revision of this PR punted timeouts to the caller. Re-reading
#3046's acceptance criteria — *"timeout and server-restart handling for those
requests come from the framework rather than from caller-side bookkeeping"* —
that only got half way: restarts came from the framework, deadlines did not,
and the issue explicitly expected to be able to delete its deadline sweep.
The example proves the difference — `polling-client.cc` no longer does any
clock arithmetic.

## `DoraEventType::NodeRestarted`

`Event::NodeRestarted` had no C++ variant and no accessor, so it fell through
`event_type`'s catch-all and arrived as `Unknown` — delivered but
unidentifiable. Added the variant plus `event_as_node_restarted`, following
`NodeFailed` / `Reload`, whose own doc comment gives this same reason.

This bears on the `ServerRestarted` contract. The correlation reports it for a
restart that lands *while a wait or poll is running*, since it owns the stream
then — verified by driving a blocking `recv_service_response` against a restart
on the stream. A restart landing between calls belongs to the node's own loop,
and that route used to drop it silently, leaving the next request to wait out
its full deadline against a server the node had already been told about. The
bridge doc and README now scope which restarts go where.

## Two confirmed issues left open for a maintainer's call

Both reproduced, both raised in the PR discussion rather than fixed here.

**A `NodeRestarted` reaches only the first-polled correlation.** Less severe
since `DoraEventType::NodeRestarted` landed: the restart now reaches the node's
own loop, so a client can cancel and resend the others itself (the C++ README
shows the loop). The framework still does not do it for you.

Originally: With N requests
in flight to one server, one gets `ServerRestarted` and the other N-1 wait out
their deadlines; the restart is buffered and never re-classified. Fixing it
means new per-correlation state and making registration unconditional, which
widens the surface `cancel_correlation` manages. Polling-only, so
`wait_for_correlation` is unaffected.

**`pending_passthrough` is unbounded and escapes the per-input queue policy** (now tracked in #3197)
applied in `Scheduler::add_event`, so a poll-only node with a high-rate input
grows where it previously dropped oldest. I started on a cap and backed out: the
buffer is written by `wait_for_correlation` too, so any cap changes behaviour for
API shipped in #2853, for users who never touch the polls.

## Known limitation: late replies stay buffered

A reply that arrives after its request timed out or was cancelled matches no
live correlation, so the next poll buffers it for the caller's event loop.
`cancel_correlation` releases the deadline, not that buffered event. Measured:
three unclaimed replies leave the buffer at three, three `cancel_correlation`
calls leave it at three, three `recv()` calls take it to zero.

Not a correctness issue — nothing is lost or misrouted — and only reachable by
a client that polls and never reads events at all; any node with inputs of its
own drains as a matter of course. Growth is one buffered event per unclaimed
reply, so it tracks the failure path rather than steady-state traffic.

Worth flagging because the obvious mitigation is wrong: `try_recv` returns
buffered events first but then falls through to the live stream, so a loop
reading until empty can consume and discard a reply a later poll was going to
correlate. Confirmed by driving it. The docs now say so in both APIs and in the
example.

A buffer-only drain would close this properly. That is new API surface and felt
like the wrong thing to add here, so it is documented rather than fixed — happy
to do it in a follow-up if you would rather have it.

## Cost note

One poll drains *all* currently-ready events, so it is O(ready events), not
O(1). That is exactly what lets it find a reply queued behind unrelated
traffic. It never waits, but a large ready burst is still a large amount of
work in one call. Documented on the method.

## Testing

**New example node** — `examples/c++-service-action/nodes/polling-client.cc`:
a client with several requests in flight that never blocks. Already covered
by the nightly `examples` job, which runs `cargo run --example
cxx-service-action` and builds every node in the directory. Verified against
a live daemon:

```
[polling-client] sent 019fd0f3-5a24-...: 0 + 10 (1 in flight)
[polling-client] sent 019fd0f3-5a25-...: 1 + 11 (2 in flight)
[polling-client] sent 019fd0f3-5a25-...: 2 + 12 (3 in flight)
[polling-client] 019fd0f3-5a24-...: 0 + 10 = 10
[polling-client] 019fd0f3-5a25-...: 1 + 11 = 12
[polling-client] 019fd0f3-5a25-...: 2 + 12 = 14
...
[polling-client] completed 6 request(s), several in flight at a time, never blocking
```

All five nodes exit `finished successfully`.

The failure path is verified too, since the happy path is all CI walks. The
service server already skips requests whose payload is under two bytes
without replying, so a one-byte payload plus a short timeout makes every
request time out. The client prints `completed 0 request(s), 6 unanswered`
and exits immediately. An earlier revision hung there instead — it counted
only answered requests toward its budget, so an unanswered one left it
spinning with nothing in `pending` to poll, which also meant it never saw
`StreamEnded` and held up dataflow teardown.

**44 new unit/integration tests** — `ExpectedServers` membership, each
`classify_correlation_event` outcome under `One`/`AnyOf`/`Any`, the
restarted-node name coming from the event, match-wins-regardless-of-server,
polling with buffered passthrough, feedback passing through an action poll,
the fan-out correlation, `NotReady` mapping, server-list parsing (empty,
`""`, malformed, empty-entry-inside-a-list), and the caller-id contract.

For deadlines: expiry reported exactly once, first-registration-wins, a match
clearing its deadline, a buffered reply beating an expired deadline,
per-correlation-id isolation, `cancel_correlation`, `u64::MAX` not panicking,
and `timeout_ms == 0` meaning no deadline.

Both new C++ entry points were confirmed to actually execute across the FFI
boundary, not merely compile, using temporary probes in the example that were
removed before commit: `Timeout` fired for a correlation nobody answers, and
`cancel_correlation` ran repeatedly (including for an unregistered id)
without disturbing the run.

The polling integration tests are bounded (2000 attempts), so an
implementation that blocked or never returned `Ok(None)` fails rather than
hangs.

**Gates:** `cargo fmt --all -- --check`, `cargo clippy --all -- -D warnings`,
`cargo test -p dora-node-api -p dora-node-api-cxx`, `cargo check --examples`,
`g++ -Wall -Wextra` on all example sources including with `-include X11/X.h`
(the `None`/`Success` macro trap from #2853), and a compile check of every
C++ snippet added to the README.

**The full `cargo test --all`** does not fit on my disk locally, and I had
previously written here that CI covers it. That was wrong for a PR: `Test
(ubuntu-latest)`, `E2E Tests`, `Semantic contract tests` and `Benchmark
regression check` all carry `if: github.event_name != 'pull_request'`, so the
green checks on this PR do not include any of them — they run on push to
`main`, in the Trunk merge queue, or on `workflow_dispatch`.

So I dispatched the same workflow against this branch on my fork, where the
condition holds and those jobs actually run. Result linked in the comments.

## Notes

- `#[allow(clippy::ptr_arg)]` on the four `&Vec<String>` functions: cxx maps
  `Vec<String>` to `rust::Vec<rust::String>` and has no slice-of-String
  equivalent, so the signature is dictated by the bridge — same situation as
  the existing `borrowed_box` allows in this file.
- An empty entry inside a non-empty server list is rejected rather than
  silently widening the set to "any", so a stray `""` cannot quietly disable
  restart detection.

### DIFF
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -343,7 +343,14 @@ auto reply = recv_service_response_from(
 The set governs only restart detection; which reply matches is decided by
 `request_id` alone. A restart of any listed node is reported as
 `ServerRestarted` naming that node — a notification, not a verdict, since
-the other candidates may still answer.
+the other candidates may still answer, so the request's deadline keeps
+running on its original clock.
+
+A list holding exactly one id is treated as the single-server case, not
+as a one-element set: with nobody else to answer, that restart *is* the
+verdict, and the orphaned request's deadline is released with it. The
+single-server polls above take the same path, so they behave
+identically.
 
 The server must echo the request's `request_id` back, which is why it needs
 `event_as_input_with_metadata`:
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -282,9 +282,25 @@ The framework owns the deadline: the first poll carrying a `timeout_ms`
 registers it against that `request_id`, and a later poll past it returns
 `Timeout` exactly once. A caller passes the same `timeout_ms` every
 iteration and reacts to `Timeout` like any other status — no deadline
-bookkeeping of its own. Pass `0` for no deadline. The clock starts at
-that first poll rather than at send time, and the first deadline
-registered for an id wins.
+bookkeeping of its own. The clock starts at that first poll rather than
+at send time, and the first deadline registered for an id wins.
+
+`timeout_ms == 0` means **no deadline**, not "return immediately" — a
+poll never blocks anyway, so there is nothing to time out. This inverts
+the usual convention for a `timeout` argument, so it is worth a second
+look when reading calling code.
+
+A poll drops its own registration on a match, an error or expiry. If the
+node abandons a request it will never poll again — the peer died, the
+operator cancelled, the reply stopped mattering — release it explicitly,
+or that one entry lives until the event stream is dropped:
+
+```c++
+cancel_correlation(dora_node.events, request_id);
+```
+
+Safe to call for an id that was never registered, so a cancel racing a
+reply is harmless.
 
 > **Ordering matters.** The polls and your own `next_event` read the same
 > event stream, so whichever runs first consumes what is there. A poll
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -334,8 +334,8 @@ acceptable responders — an empty vector means any node:
 
 ```c++
 rust::Vec<rust::String> candidates;
-candidates.push_back("cam-eo");
-candidates.push_back("cam-ir");
+candidates.push_back("a");
+candidates.push_back("b");
 auto reply = recv_service_response_from(
     dora_node.events, request_id, candidates, /* timeout_ms */ 5000);
 ```
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -250,6 +250,76 @@ The `event` field also carries a matching `DoraEventType` (`Timeout` /
 `server_node_id` yields `DoraPatternStatus::InvalidArgument` rather than
 aborting the process.
 
+#### Clients that cannot block
+
+`recv_service_response` waits. A single-threaded node with its own
+schedule to keep, or several requests outstanding at once, cannot afford
+that. `try_recv_service_response` polls instead and returns
+`DoraPatternStatus::NotReady` when the reply has not arrived — the same
+convention `try_next_event` uses for plain events. Buffering, restart
+detection and correlation are unchanged; only the waiting is gone.
+
+```c++
+// per loop iteration, for each outstanding request
+auto poll = try_recv_service_response(dora_node.events, request_id, "server-node-id");
+switch (poll.status)
+{
+case DoraPatternStatus::Matched:
+    complete(event_as_input(std::move(poll.event)));
+    break;
+case DoraPatternStatus::NotReady:
+    break; // nothing to do, and no time spent
+default:
+    std::cerr << std::string(poll.error) << std::endl;
+}
+```
+
+There is no deadline, so `Timeout` never comes back from a poll — keeping
+deadlines is the caller's job.
+
+> **Ordering matters.** The polls and your own `next_event` read the same
+> event stream, so whichever runs first consumes what is there. A poll
+> correlates the reply it wants and buffers everything else for a later
+> `next_event`, so polling first loses nothing. The reverse is not true:
+> a reply handed to `next_event` is gone, and no later poll can see it.
+> Poll first, then drain your own events with `try_next_event`.
+
+`try_recv_action_result` is the equivalent for actions. See
+`examples/c++-service-action/nodes/polling-client.cc` for a client with
+several requests in flight that never blocks.
+
+#### Fanning one request out to several servers
+
+`send_service_request` mints a fresh `request_id` per call, so it cannot
+express one logical request sent to several nodes. Mint the id once and
+send each copy with `send_service_request_with_id`:
+
+```c++
+auto request_id = std::string(new_request_id());
+for (const auto &server : servers)
+{
+    send_service_request_with_id(
+        dora_node.send_output, server.output, payload, new_metadata(), request_id);
+}
+```
+
+Then await whichever answers first with `recv_service_response_from` (or
+`try_recv_service_response_from`), which take a `Vec<String>` of
+acceptable responders — an empty vector means any node:
+
+```c++
+rust::Vec<rust::String> candidates;
+candidates.push_back("cam-eo");
+candidates.push_back("cam-ir");
+auto reply = recv_service_response_from(
+    dora_node.events, request_id, candidates, /* timeout_ms */ 5000);
+```
+
+The set governs only restart detection; which reply matches is decided by
+`request_id` alone. A restart of any listed node is reported as
+`ServerRestarted` naming that node — a notification, not a verdict, since
+the other candidates may still answer.
+
 The server must echo the request's `request_id` back, which is why it needs
 `event_as_input_with_metadata`:
 
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -261,7 +261,8 @@ detection and correlation are unchanged; only the waiting is gone.
 
 ```c++
 // per loop iteration, for each outstanding request
-auto poll = try_recv_service_response(dora_node.events, request_id, "server-node-id");
+auto poll = try_recv_service_response(
+    dora_node.events, request_id, "server-node-id", /* timeout_ms */ 5000);
 switch (poll.status)
 {
 case DoraPatternStatus::Matched:
@@ -269,13 +270,21 @@ case DoraPatternStatus::Matched:
     break;
 case DoraPatternStatus::NotReady:
     break; // nothing to do, and no time spent
+case DoraPatternStatus::Timeout:
+    give_up(); // the registered deadline lapsed; reported once
+    break;
 default:
     std::cerr << std::string(poll.error) << std::endl;
 }
 ```
 
-There is no deadline, so `Timeout` never comes back from a poll — keeping
-deadlines is the caller's job.
+The framework owns the deadline: the first poll carrying a `timeout_ms`
+registers it against that `request_id`, and a later poll past it returns
+`Timeout` exactly once. A caller passes the same `timeout_ms` every
+iteration and reacts to `Timeout` like any other status — no deadline
+bookkeeping of its own. Pass `0` for no deadline. The clock starts at
+that first poll rather than at send time, and the first deadline
+registered for an id wins.
 
 > **Ordering matters.** The polls and your own `next_event` read the same
 > event stream, so whichever runs first consumes what is there. A poll
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -257,6 +257,31 @@ The `event` field also carries a matching `DoraEventType` (`Timeout` /
 `server_node_id` yields `DoraPatternStatus::InvalidArgument` rather than
 aborting the process.
 
+#### Reacting to a restart with several requests in flight
+
+A restart is reported to the *one* correlation that consumes it — the wait or
+poll that happened to be reading the stream. Others outstanding against the
+same server are not told, and would otherwise sit until their own deadlines.
+Handling the event in your own loop covers them:
+
+```c++
+case DoraEventType::NodeRestarted:
+{
+    const auto who = std::string(event_as_node_restarted(std::move(event)));
+    // Everything still outstanding against that node is orphaned: the
+    // instance that would have answered is gone.
+    for (const auto &entry : pending_for(who))
+    {
+        cancel_correlation(dora_node.events, entry.request_id);
+    }
+    resend_against_new_instance(who);
+    break;
+}
+```
+
+`cancel_correlation` releases the framework's deadline for those requests;
+resending is yours to do, under a fresh `request_id` from `new_request_id()`.
+
 #### Clients that cannot block
 
 `recv_service_response` waits. A single-threaded node with its own
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -245,6 +245,13 @@ default:
 
 Events that arrive during the wait but do not match are buffered and
 replayed by later `next_event` calls, so your main event loop loses nothing.
+
+`ServerRestarted` is reported for a restart that lands while a wait or poll
+is running, because the correlation is reading the stream then. One that
+lands between calls reaches your own loop as `DoraEventType::NodeRestarted`
+(use `event_as_node_restarted` for the id). Ignore it and the next request
+will wait out its whole deadline against a server you had already been told
+had restarted.
 The `event` field also carries a matching `DoraEventType` (`Timeout` /
 `AllInputsClosed` / `Empty`), so you can branch on either field. A malformed
 `server_node_id` yields `DoraPatternStatus::InvalidArgument` rather than
--- a/apis/c++/node/README.md
+++ b/apis/c++/node/README.md
@@ -317,10 +317,11 @@ iteration and reacts to `Timeout` like any other status — no deadline
 bookkeeping of its own. The clock starts at that first poll rather than
 at send time, and the first deadline registered for an id wins.
 
-`timeout_ms == 0` means **no deadline**, not "return immediately" — a
-poll never blocks anyway, so there is nothing to time out. This inverts
-the usual convention for a `timeout` argument, so it is worth a second
-look when reading calling code.
+`timeout_ms` carries one meaning across the whole header: `0` expires
+immediately, exactly as it does for the blocking `recv_service_response`,
+and `UINT64_MAX` is the spelling for **no practical deadline**. Moving a
+request from a blocking wait to a poll by changing the function name
+therefore never silently drops its deadline.
 
 A poll drops its own registration on a match, an error or expiry. If the
 node abandons a request it will never poll again — the peer died, the
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -2089,14 +2089,14 @@ mod tests {
 
     #[test]
     fn server_list_parses_every_entry() {
-        let Ok(parsed) = parse_server_list(&["cam-eo", "cam-ir"]) else {
+        let Ok(parsed) = parse_server_list(&["a", "b"]) else {
             panic!("both ids are valid");
         };
         match parsed {
             OwnedServers::Some(ids) => {
                 assert_eq!(ids.len(), 2);
-                assert_eq!(ids[0].as_ref(), "cam-eo");
-                assert_eq!(ids[1].as_ref(), "cam-ir");
+                assert_eq!(ids[0].as_ref(), "a");
+                assert_eq!(ids[1].as_ref(), "b");
             }
             OwnedServers::Any => panic!("a non-empty list must not widen to Any"),
         }
@@ -2106,7 +2106,7 @@ mod tests {
     fn empty_entry_inside_a_list_is_rejected() {
         // Silently treating this as "any" would let one stray "" switch
         // off restart detection for the whole set.
-        let result = parse_server_list(&["cam-eo", ""]);
+        let result = parse_server_list(&["a", ""]);
         match result {
             Err(failure) => assert!(matches!(
                 failure.status,
@@ -2151,13 +2151,13 @@ mod tests {
         let timeout = try_pattern_result(Err(PatternError::Timeout));
         assert!(matches!(timeout.status, ffi::DoraPatternStatus::Timeout));
 
-        let restarted = try_pattern_result(Err(PatternError::ServerRestarted("cam-ir".into())));
+        let restarted = try_pattern_result(Err(PatternError::ServerRestarted("b".into())));
         assert!(matches!(
             restarted.status,
             ffi::DoraPatternStatus::ServerRestarted
         ));
         assert!(
-            restarted.error.contains("cam-ir"),
+            restarted.error.contains("b"),
             "the restarted node's id must reach the C++ caller: {}",
             restarted.error
         );
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -366,25 +366,40 @@ mod ffi {
         /// state carries across calls — so polling in a loop is
         /// equivalent to waiting, minus the blocking.
         ///
-        /// There is no deadline, so `Timeout` is never returned; keeping
-        /// deadlines is the caller's job. Pass an empty
-        /// `server_node_id` to accept a reply from any node and skip
-        /// restart correlation.
+        /// `timeout_ms` is owned by the framework: the first poll
+        /// carrying one registers a deadline for `request_id`, and a
+        /// later poll past it returns `Timeout` exactly once. So a
+        /// caller does not sweep deadlines itself — it passes the same
+        /// `timeout_ms` every iteration and reacts to `Timeout` like
+        /// any other status. Pass `0` for no deadline.
+        ///
+        /// The clock starts at that first poll rather than at send
+        /// time, and the first deadline registered for an id wins.
+        ///
+        /// Pass an empty `server_node_id` to accept a reply from any
+        /// node and skip restart correlation.
         fn try_recv_service_response(
             events: &mut Box<Events>,
             request_id: &str,
             server_node_id: &str,
+            timeout_ms: u64,
         ) -> DoraPatternResult;
 
         /// Non-blocking poll for a *terminal* result for `goal_id`.
         ///
         /// Returns `NotReady` while the goal is still running. Feedback
-        /// messages stay buffered for the caller's own event loop. Pass
-        /// an empty `server_node_id` to accept any responder.
+        /// messages stay buffered for the caller's own event loop.
+        ///
+        /// `timeout_ms` behaves as in `try_recv_service_response`, but
+        /// bounds the *whole goal* rather than the gap between feedback
+        /// messages: a long goal that is making visible progress will
+        /// still expire. Pass `0` for no deadline, and an empty
+        /// `server_node_id` to accept any responder.
         fn try_recv_action_result(
             events: &mut Box<Events>,
             goal_id: &str,
             server_node_id: &str,
+            timeout_ms: u64,
         ) -> DoraPatternResult;
 
         /// `recv_service_response` over a set of acceptable responders,
@@ -418,6 +433,7 @@ mod ffi {
             events: &mut Box<Events>,
             request_id: &str,
             server_node_ids: &Vec<String>,
+            timeout_ms: u64,
         ) -> DoraPatternResult;
 
         /// Non-blocking `recv_action_result_from`.
@@ -425,6 +441,7 @@ mod ffi {
             events: &mut Box<Events>,
             goal_id: &str,
             server_node_ids: &Vec<String>,
+            timeout_ms: u64,
         ) -> DoraPatternResult;
 
         /// Send a service request under a caller-supplied `request_id`.
@@ -1361,28 +1378,34 @@ fn try_recv_service_response(
     events: &mut Box<Events>,
     request_id: &str,
     server_node_id: &str,
+    timeout_ms: u64,
 ) -> ffi::DoraPatternResult {
     let servers = match parse_server_list(std::slice::from_ref(&server_node_id)) {
         Ok(servers) => servers,
         Err(result) => return result,
     };
-    try_pattern_result(
-        events
-            .0
-            .try_recv_service_response(request_id, servers.as_ref()),
-    )
+    try_pattern_result(events.0.try_recv_service_response(
+        request_id,
+        servers.as_ref(),
+        poll_timeout(timeout_ms),
+    ))
 }
 
 fn try_recv_action_result(
     events: &mut Box<Events>,
     goal_id: &str,
     server_node_id: &str,
+    timeout_ms: u64,
 ) -> ffi::DoraPatternResult {
     let servers = match parse_server_list(std::slice::from_ref(&server_node_id)) {
         Ok(servers) => servers,
         Err(result) => return result,
     };
-    try_pattern_result(events.0.try_recv_action_result(goal_id, servers.as_ref()))
+    try_pattern_result(events.0.try_recv_action_result(
+        goal_id,
+        servers.as_ref(),
+        poll_timeout(timeout_ms),
+    ))
 }
 
 /// `&Vec<String>` rather than `&[String]`: cxx maps `Vec<String>` to
@@ -1442,16 +1465,17 @@ fn try_recv_service_response_from(
     events: &mut Box<Events>,
     request_id: &str,
     server_node_ids: &Vec<String>,
+    timeout_ms: u64,
 ) -> ffi::DoraPatternResult {
     let servers = match parse_server_strings(server_node_ids) {
         Ok(servers) => servers,
         Err(result) => return result,
     };
-    try_pattern_result(
-        events
-            .0
-            .try_recv_service_response(request_id, servers.as_ref()),
-    )
+    try_pattern_result(events.0.try_recv_service_response(
+        request_id,
+        servers.as_ref(),
+        poll_timeout(timeout_ms),
+    ))
 }
 
 /// `&Vec<String>` rather than `&[String]`: cxx maps `Vec<String>` to
@@ -1463,12 +1487,17 @@ fn try_recv_action_result_from(
     events: &mut Box<Events>,
     goal_id: &str,
     server_node_ids: &Vec<String>,
+    timeout_ms: u64,
 ) -> ffi::DoraPatternResult {
     let servers = match parse_server_strings(server_node_ids) {
         Ok(servers) => servers,
         Err(result) => return result,
     };
-    try_pattern_result(events.0.try_recv_action_result(goal_id, servers.as_ref()))
+    try_pattern_result(events.0.try_recv_action_result(
+        goal_id,
+        servers.as_ref(),
+        poll_timeout(timeout_ms),
+    ))
 }
 
 /// Owned form of [`ExpectedServers`], which borrows.
@@ -1512,6 +1541,21 @@ fn parse_server_list(ids: &[&str]) -> Result<OwnedServers, ffi::DoraPatternResul
         parsed.push(parse_server_node_id(id)?);
     }
     Ok(OwnedServers::Some(parsed))
+}
+
+/// Map a C++ `timeout_ms` onto the Rust poll's optional deadline.
+///
+/// `0` means "no deadline" — the poll then behaves exactly as it did
+/// before deadlines existed. Any other value is clamped the same way
+/// `clamp_pattern_timeout` clamps the blocking waits, because it
+/// crosses the bridge as an unbounded `u64` and ends up in an
+/// `Instant + Duration`.
+fn poll_timeout(timeout_ms: u64) -> Option<Duration> {
+    if timeout_ms == 0 {
+        None
+    } else {
+        Some(clamp_pattern_timeout(timeout_ms))
+    }
 }
 
 /// Map a non-blocking correlated receive onto the C++ result struct.
@@ -1996,6 +2040,22 @@ mod tests {
             pinned.get(dora_node_api::REQUEST_ID),
             Some(&DoraParameter::String("req-fixed".into()))
         );
+    }
+
+    #[test]
+    fn zero_timeout_means_no_deadline() {
+        // `0` has to keep meaning "poll forever", or a caller that
+        // never wanted a deadline would start getting Timeout.
+        assert!(poll_timeout(0).is_none());
+    }
+
+    #[test]
+    fn nonzero_timeout_is_clamped_like_the_blocking_waits() {
+        let clamped = poll_timeout(u64::MAX).expect("a non-zero timeout is a deadline");
+        assert_eq!(clamped, clamp_pattern_timeout(u64::MAX));
+
+        let ordinary = poll_timeout(5_000).expect("a non-zero timeout is a deadline");
+        assert_eq!(ordinary, Duration::from_millis(5_000));
     }
 
     #[test]
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -444,6 +444,18 @@ mod ffi {
             timeout_ms: u64,
         ) -> DoraPatternResult;
 
+        /// Forget an outstanding correlation's deadline.
+        ///
+        /// A poll drops its own registration on a match, an error or
+        /// expiry, so this is only needed when a node abandons a
+        /// request it will never poll again — a peer died, the operator
+        /// cancelled, the reply stopped mattering. Without it that one
+        /// entry lives until the event stream is dropped, which for a
+        /// long-running node is an unbounded slow leak.
+        ///
+        /// Safe to call for an id that was never registered.
+        fn cancel_correlation(events: &mut Box<Events>, correlation_id: &str);
+
         /// Send a service request under a caller-supplied `request_id`.
         ///
         /// `send_service_request` mints a fresh id per call, so it
@@ -1543,6 +1555,10 @@ fn parse_server_list(ids: &[&str]) -> Result<OwnedServers, ffi::DoraPatternResul
     Ok(OwnedServers::Some(parsed))
 }
 
+fn cancel_correlation(events: &mut Box<Events>, correlation_id: &str) {
+    events.0.cancel_correlation(correlation_id);
+}
+
 /// Map a C++ `timeout_ms` onto the Rust poll's optional deadline.
 ///
 /// `0` means "no deadline" — the poll then behaves exactly as it did
@@ -2001,10 +2017,6 @@ mod tests {
         dora_node_api::uuid::Uuid::parse_str(&new_goal_id()).expect("goal id should be a UUID");
     }
 
-    /// A caller-supplied `request_id` must be replaced, matching
-    /// `DoraNode::send_service_request`. Silently honouring it would let
-    /// two in-flight requests share a correlation key.
-    #[test]
     // ---- dora-rs/dora#3046 ----
 
     /// The whole point of the `_with_id` variant: the caller's id must
@@ -2151,6 +2163,9 @@ mod tests {
         );
     }
 
+    /// A caller-supplied `request_id` must be replaced, matching
+    /// `DoraNode::send_service_request`. Silently honouring it would let
+    /// two in-flight requests share a correlation key.
     #[test]
     fn insert_request_id_overwrites_caller_value() {
         let mut parameters = DoraMetadataParameters::default();
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -1527,7 +1527,24 @@ impl OwnedServers {
     fn as_ref(&self) -> ExpectedServers<'_> {
         match self {
             Self::Any => ExpectedServers::Any,
-            Self::Some(ids) => ExpectedServers::AnyOf(ids),
+            // Exactly one candidate is `One`, not a one-element `AnyOf`.
+            //
+            // The two differ in what a restart *means*: with a set,
+            // a restart is a notification because the others may still
+            // answer, so the deadline keeps running. With a single
+            // server there is nobody else, so the restart is the
+            // verdict and the deadline must go with the orphaned
+            // request.
+            //
+            // This is the path every C++ single-server poll takes —
+            // `try_recv_service_response(.., server_node_id, ..)` wraps
+            // its lone id in a slice — so without this it would leak one
+            // deadline entry per abandoned request, which is exactly
+            // what `cancel_correlation` was added to avoid needing.
+            Self::Some(ids) => match ids.as_slice() {
+                [only] => ExpectedServers::One(only),
+                many => ExpectedServers::AnyOf(many),
+            },
         }
     }
 }
@@ -2085,6 +2102,45 @@ mod tests {
             panic!("an empty id is the 'any' spelling and must be accepted");
         };
         assert!(matches!(parsed, OwnedServers::Any));
+    }
+
+    /// The C++ single-server polls wrap their lone id in a slice, so
+    /// this is the mapping every one of them goes through. It has to
+    /// land on `One`: with a single candidate a restart is terminal, and
+    /// `One` is what drops the orphaned request's deadline. A
+    /// one-element `AnyOf` would keep it, leaking one entry per
+    /// abandoned request.
+    ///
+    /// Paired with `a_one_server_restart_clears_the_deadline` in
+    /// `dora-node-api`, which pins the clearing behaviour itself.
+    #[test]
+    fn a_single_server_maps_to_one_not_a_one_element_any_of() {
+        let Ok(parsed) = parse_server_list(&["srv"]) else {
+            panic!("a single valid id must parse");
+        };
+        assert!(
+            matches!(parsed.as_ref(), ExpectedServers::One(id) if id.as_ref() == "srv"),
+            "a lone server must become One, or its restart will not clear the deadline"
+        );
+    }
+
+    #[test]
+    fn several_servers_still_map_to_any_of() {
+        let Ok(parsed) = parse_server_list(&["a", "b"]) else {
+            panic!("both ids are valid");
+        };
+        assert!(
+            matches!(parsed.as_ref(), ExpectedServers::AnyOf(ids) if ids.len() == 2),
+            "a real fan-out must keep AnyOf semantics"
+        );
+    }
+
+    #[test]
+    fn no_servers_still_maps_to_any() {
+        let Ok(parsed) = parse_server_list(&[]) else {
+            panic!("an empty list is valid");
+        };
+        assert!(matches!(parsed.as_ref(), ExpectedServers::Any));
     }
 
     #[test]
--- a/apis/c++/node/src/lib.rs
+++ b/apis/c++/node/src/lib.rs
@@ -11,7 +11,7 @@ use chrono::DateTime;
 #[cfg(any(feature = "ros2-bridge", test))]
 use dora_node_api::merged::MergeExternal;
 use dora_node_api::{
-    self, Event, EventStream, Metadata as DoraMetadata,
+    self, Event, EventStream, ExpectedServers, Metadata as DoraMetadata,
     MetadataParameters as DoraMetadataParameters, Parameter as DoraParameter, PatternError,
     TryRecvError,
     arrow_v59::array::{AsArray, UInt8Array},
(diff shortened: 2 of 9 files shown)


phil-opp:

I reviewed this PR — large but additive, no existing signatures change. The implementation and tests look solid. `try_correlation` drains ready events one at a time via `recv_from_stream` + `now_or_never` (bypassing `pending_passthrough` so buffered events aren't re-handed), buffers non-matches, and checks the already-buffered reply before the deadline so a reply that arrived in time isn't discarded. The `ExpectedServers` refactor preserves the existing `recv_service_response` / `recv_action_result` behavior — for `One`, `classify_correlation_event` only reports `ServerRestarted` when the event's node equals the expected one, so `restarted_node_id(&event)` yields the same string the old `expected_server.to_string()` did. The added tests are real (bounded `poll_until` loops that fail rather than hang, deterministic `Duration::ZERO` expiry, per-id deadline isolation, fan-out correlation), not vacuous.

One thing worth a look:

- In the polling path, a `ServerRestarted` outcome removes the correlation's deadline (`correlation_deadlines.remove(needle)` in `try_correlation`) before returning `Err(ServerRestarted)`. For `ExpectedServers::AnyOf`, the docs describe a restart of a listed node as "a notification, not a verdict … the caller can simply wait again." But when the caller re-polls with the same timeout, `correlation_expired` re-registers a fresh deadline on a new clock, so a listed server that keeps restarting resets the framework-owned deadline every iteration and the caller may never observe `Timeout` — which runs counter to the stated "the framework owns the deadline / pass the same timeout every iteration and react to `Timeout`" contract for the fan-out case. For `One` this is harmless since the restart is terminal. Consider preserving the deadline across an `AnyOf` restart notification (only clearing it on match/expiry), or documenting that a restart resets the clock.

Otherwise no issues found.

---

phil-opp:
**Automated review by Claude — this is a fully automated review and has not been checked by a human.**

I looked at the current diff (head `6bf4b81`). The core `try_correlation` / `ExpectedServers` / deadline logic looks fine to me beyond the `AnyOf`-restart deadline-reset point already raised in the earlier review, which I won't repeat.

One new issue, in the C++ node test module (`apis/c++/node/src/lib.rs`, `#[cfg(test)] mod tests`): a botched insertion left `send_service_request_with_id_preserves_the_caller_id` with two `#[test]` attributes and the wrong doc comment. The new test was spliced in *between* the doc comment and the `#[test]` that belonged to `insert_request_id_overwrites_caller_value`:

```rust
    /// A caller-supplied `request_id` must be replaced, matching
    /// `DoraNode::send_service_request`. Silently honouring it would let
    /// two in-flight requests share a correlation key.
    #[test]                       // <-- stray first #[test]
    // ---- dora-rs/dora#3046 ----

    /// The whole point of the `_with_id` variant: the caller's id must
    /// survive, or a fan-out cannot share one correlation.
    #[test]
    fn send_service_request_with_id_preserves_the_caller_id() {
```

Consequences:
1. `send_service_request_with_id_preserves_the_caller_id` now carries **two `#[test]` attributes** (a duplicate-attribute warning). It still compiles — which is why CI stayed green, since the test module isn't built under the `cargo clippy --all -- -D warnings` gate (no `--all-targets`) and `cargo test` doesn't treat warnings as errors.
2. The `/// A caller-supplied request_id must be replaced …` doc — which describes the *overwrite* behavior — is now attached to the *preserve* test (the opposite behavior), while `insert_request_id_overwrites_caller_value` is left undocumented.

Removing the stray `#[test]` and moving that doc block back down onto `insert_request_id_overwrites_caller_value` fixes both.

---

phil-opp:

Re-reviewed at head `a53366`:

- The mis-spliced test I flagged last time is fixed: `send_service_request_with_id_preserves_the_caller_id` now carries a single `#[test]` with its correct doc, and the "must be replaced" doc is back on `insert_request_id_overwrites_caller_value`.
- `cancel_correlation` is correctly exposed across the cxx bridge (`apis/c++/node/src/lib.rs:429`) and the Rust `EventStream` (`apis/rust/node/src/event_stream/mod.rs:1013`); the abandoned-registration leak the commit targets is closed.
- Still outstanding (from my first review, not addressed here): in `try_correlation`, a `ServerRestarted` outcome removes the correlation's deadline (`event_stream/mod.rs:948`) before returning. For `ExpectedServers::AnyOf`, where a restart is documented as a notification the caller can wait past, the next poll re-registers a fresh deadline on a new clock via `correlation_expired`, so a listed node that keeps restarting can reset the framework-owned deadline indefinitely and the caller may never observe `Timeout`. Consider preserving the deadline across an `AnyOf` restart, or documenting that a restart resets the clock.

No new issues.

---

phil-opp:
> 🤖 **This is a fully automated review by Claude. No human has vetted these findings — please double-check before acting.**

Re-reviewed after the latest commit (`9e3101564`), which post-dates the previous automated review. It resolves the `AnyOf`-restart deadline-reset point raised earlier: in `try_correlation` (`apis/rust/node/src/event_stream/mod.rs`) the deadline is now preserved on a `ServerRestarted` outcome for `ExpectedServers::AnyOf` and only dropped for `One`, and `correlation_expired` registers the deadline with `or_insert_with`, so re-polling after a restart keeps the original clock instead of resetting it — a flapping listed node can no longer hold the correlation open past the requested timeout. Both branches are pinned by `an_any_of_restart_preserves_the_deadline` and `a_one_server_restart_clears_the_deadline`, which drive a real `NodeRestarted` through the scheduler rather than the passthrough buffer. No new issues found in the latest commits.

---

phil-opp:
**Automated review by Claude — this is a fully automated review.**

Re-reviewed at the latest commit. The previously flagged `AnyOf`-restart deadline reset and the mis-spliced C++ test module are both resolved, and the Rust-side tests (`an_any_of_restart_preserves_the_deadline`, `a_one_server_restart_clears_the_deadline`, `first_registered_deadline_wins`, etc.) drive real scheduler events and would fail if their branch regressed — they're not vacuous.

One minor, previously-unflagged point:

The C++ single-server poll `try_recv_service_response(req, "server", timeout)` routes a lone server through `parse_server_list(std::slice::from_ref(&server_node_id))` → `OwnedServers::Some([id])` → `ExpectedServers::AnyOf(&[id])` (`apis/c++/node/src/lib.rs`). In `recv_*_from`, the `ServerRestarted` branch keeps the deadline for *every* `AnyOf`:

```rust
if !matches!(expected, ExpectedServers::AnyOf(_)) {
    self.correlation_deadlines.remove(needle);
}
```

So when that single server restarts, the C++ caller gets `ServerRestarted` (which is terminal when there's only one server) but the `correlation_deadlines` entry is **not** cleared — unlike the Rust `One` path, which drops it. A caller that abandons the request without calling `cancel_correlation` leaks that one entry until the stream is dropped; it won't self-heal via expiry unless the same correlation id is polled again.

It's bounded (one entry per abandoned request) and error-path-only, so low severity — but it's exactly the leak class `cancel_correlation` was added to prevent, and it isn't covered by a test (the deadline-clearing tests all use the Rust `One`/`AnyOf` API directly, never the C++ single-server bridge path). Routing a single-element server list to `ExpectedServers::One`, or documenting that C++ single-server polls should `cancel_correlation` on `ServerRestarted`, would close it.

---

phil-opp:

Re-reviewed the latest commit (`3e98095b`), which post-dates my previous review. It fixes the C++ single-server deadline leak I flagged: `OwnedServers::as_ref` now resolves by candidate count — a one-element `Some` maps to `ExpectedServers::One` (so a terminal restart releases the orphaned request's deadline) and a multi-element list stays `AnyOf`. The three new mapping tests (`a_single_server_maps_to_one_not_a_one_element_any_of`, `several_servers_still_map_to_any_of`, `no_servers_still_maps_to_any`) pin all three cases, and the single-server one fails if the fix is reverted. No new issues.

---

phil-opp:
**Automated review by Claude — fully automated, not vetted by a human.**

Re-reviewed at HEAD (`3e98095b`). The library side looks solid: `try_correlation` checks the buffered reply before the deadline, the `One`/`AnyOf` deadline asymmetry is correct (`One` clears on restart, `AnyOf` keeps its original clock), `correlation_expired` is first-registration-wins with a saturating add, and the cxx single-server→`One` mapping is right. The 39 tests are non-vacuous (bounded poll loops, deterministic `Duration::ZERO` expiry, real `NodeRestarted` injection). The issues from the earlier reviews all appear resolved. I found one new issue — in the example, not the library:

**`examples/c++-service-action/nodes/polling-client.cc` hangs on any request that does not complete (timeout / server restart / error).**

In `sweep()`, the `Timeout`, `ServerRestarted`, and `default` (error) arms drop the entry from `pending` but do **not** increment `out.completed`. In `main()` the loop is `while (completed < REQUEST_BUDGET || !pending.empty())`, and new requests are only issued while `issued < REQUEST_BUDGET`. So once all `REQUEST_BUDGET` requests have been issued, a single non-completing request means `completed` can never reach `REQUEST_BUDGET`; `pending` drains to empty, and the loop then spins forever at `POLL_INTERVAL`. Worse, `sweep()` only calls `try_recv_service_response` for entries in `pending`, so with `pending` empty the node never polls again and therefore never observes `StreamEnded` — it can't shut down, and it blocks clean dataflow teardown.

This is precisely the failure path the example is built to showcase — `REQUEST_TIMEOUT_MS = 5000` exists to survive a wedged server — yet actually hitting a timeout deadlocks the node. The nightly `examples` job only exercises the happy path (all 6 replies arrive → clean exit), which is why CI stays green and this isn't caught.

Suggested fix: count terminal-but-not-matched requests toward loop termination (e.g. track `finished = completed + failed` against `REQUEST_BUDGET`), or drive the loop off `issued`/`pending` rather than `completed`.

---

phil-opp:

Re-reviewed at `669b4fd`, which adds one commit since the last review.

The `polling-client.cc` hang flagged earlier is fixed. The loop now terminates on requests *settled* rather than answered — `while (completed + failed < REQUEST_BUDGET || !pending.empty())` — the `Timeout`/`ServerRestarted`/error arms increment `failed`, and a send that never reaches the wire counts as settled too (`failed++; issued++`). So a timed-out or orphaned request no longer leaves `completed` permanently short with an empty `pending`, and the node reaches `StreamEnded` instead of spinning. The accounting is consistent (`pending` is non-empty exactly while `completed + failed < issued <= REQUEST_BUDGET`).

I re-checked the library side and have no new findings: `try_correlation` takes `&mut self` while draining so the `now_or_never` poll retains any received event (flume recv is cancel-safe) — a non-blocking poll can't drop or misroute a reply; non-matches buffer into `pending_passthrough` and are checked before the deadline; the `One`-clears / `AnyOf`-keeps deadline asymmetry and the cxx single-server→`One` mapping are correct.

One pre-existing, low-severity note (not introduced here): a reply that arrives *after* its request timed out or was cancelled is buffered in `pending_passthrough`, and `cancel_correlation` clears only the deadline map, not that buffer — so a long-running client that never drains via `next_event` accumulates late replies. Bounded here (≤6), just worth a caller's awareness on a long-lived fan-out node.

---

phil-opp:

Re-reviewed the one new commit since the last pass (`93d3b7ce`, "docs(node-api): say what cancel_correlation does not do"). It is documentation-only — 53 additions, 0 deletions across `cancel_correlation`'s doc in both the Rust and C++ APIs, `docs/patterns.md`, and the polling example's header comment. No code or behavior change.

It addresses the low-severity note the previous review raised: that `cancel_correlation` releases only the deadline and a late-arriving reply stays buffered for `next_event`, and it correctly warns against the wrong mitigation (a read-until-empty drain can swallow a reply a later poll would correlate, since `try_recv`/`try_next_event` fall through to the live stream after the buffer). Accurate, and no new issues in this delta.

---

phil-opp:
The correlation machinery is sound — no event loss (`recv_from_stream` takes `&mut self`, and `tokio::sync::mpsc::Receiver::recv` is cancel-safe), no deadlock, and the deadline map is correctly keyed and cleaned up. Three notes:

1. **Expiry is checked before the stream is drained** (`apis/rust/node/src/event_stream/mod.rs:1427`). The buffered scan above it only inspects `pending_passthrough`, so a reply that landed in the receiver at t=4.9s but has not been drained yet is reported `Timeout` if the caller polls at t=5.001s. The doc claim "a reply that did arrive in time is never discarded because the caller polled late" only holds for already-buffered replies. Moving `correlation_expired` below the drain loop makes it unconditionally true.

2. **`NodeRestarted` reaches only the first-polled correlation** (`:1465`). It is buffered afterwards, and the buffered scan uses `is_match`, which only matches `Event::Input`, so it is never re-classified. With N requests in flight to one server — exactly `polling-client.cc`'s shape — only the first gets `ServerRestarted` and the other N-1 wait out their deadlines. Inherited from `wait_for_correlation`, but multi-in-flight is this PR's headline use case, so it deserves a doc line.

3. **`pending_passthrough` is an unbounded `VecDeque`** (`:1481`) and a poll drains the whole scheduler into it, bypassing the per-input `queue_size`/`drop_oldest` enforcement applied at `add_event` time (`scheduler.rs:207`). The blocking form does this too, but only for one wait; a permanent poll loop makes it steady state, so a polling node with a high-rate input grows unboundedly where it previously dropped oldest.

Also worth knowing: the C++ timeout/never-answered path has no CI coverage (only the mapping helpers are unit-tested), and the Python/C bindings have no correlated receive at all — pre-existing, not introduced here.

caothu2k1:
Thanks — all three reproduce. I drove each one rather than reasoning from the
code, and the measurements changed what I did about them.

## 1. Expiry before the drain — fixed

Reproduced: with a reply sitting unread in the receiver and a deadline that had
just lapsed, the poll returned `Err(Timeout)` and never looked at the reply.

`correlation_expired` now runs *after* the drain loop: the loop `break`s when
the stream would block, and the deadline is only consulted once everything
already available has been examined. The doc claim you quoted is now
unconditionally true rather than true-for-buffered-replies-only.

The regression test pushes the reply straight onto the stream via the scheduler
rather than through `pending_passthrough`, so it exercises the drain path rather
than the buffer scan, and it fails without the fix.

## 2. Restart reaches only the first-polled correlation — confirmed, not yet fixed

Reproduced: three correlations against one server, one `NodeRestarted`
injected. `req-1` got `ServerRestarted("calc")`; `req-2` and `req-3` got
nothing, and the restart stayed buffered without being re-classified.

Fixing it properly means tracking which servers each live correlation expects
and replaying a restart against all of them. That is new per-correlation state,
and registration currently only happens when a timeout is supplied, so it would
also have to become unconditional — which widens the abandoned-registration
surface `cancel_correlation` exists to manage. It is polling-only, so it does
not disturb `wait_for_correlation`.

I would rather land that as a follow-up than bolt it on here. Say the word if
you would prefer it in this PR.

## 3. Unbounded `pending_passthrough` — confirmed, and I stopped short

Confirmed: the buffer is a plain `VecDeque` with no policy, and `queue_size` /
`drop_oldest` are enforced in `Scheduler::add_event`, so an event that moves
scheduler → passthrough leaves that policy behind. Your framing is right that a
permanent poll loop makes this steady state rather than per-wait.

I started on a cap and backed out, because the buffer is written by **both**
paths — `wait_for_correlation` (blocking, shipped in #2853) and
`try_correlation`. Any cap therefore changes what the existing blocking API
retains, for users who never touch the polls. That is a behaviour change to
shipped API in a PR about adding new ones, so it did not feel like mine to make
unilaterally.

Two shapes I can see, both yours to choose:

- A global cap with drop-oldest and a warn — simple, bounds memory, but silently
  changes blocking behaviour and does not honour per-input policy.
- Mirror the per-input policy, reusing `select_eviction` so correlated events
  keep their preferential treatment — faithful, more code, still a behaviour
  change for the blocking path.

Documented for now: `cancel_correlation`, `docs/patterns.md` and the example all
say that a poll-only node that never reads events retains unclaimed traffic, and
that the obvious mitigation is wrong (`try_recv` empties the buffer then falls
through to the live stream, so a drain loop can swallow a reply a later poll
wanted — also confirmed by driving it).

## Your two closing notes

Both fair, and neither introduced here.

The C++ timeout / never-answered path has no CI coverage — correct. I verified it
by hand (the service server already skips sub-two-byte payloads without
replying, which forces every request to time out) and confirmed the client now
exits instead of hanging, but that is not automation. Adding a deliberately
failing request to the example would cover it at the cost of teaching the wrong
thing and paying a timeout cycle per nightly run; a separate smoke dataflow
would be cleaner if you want it covered.

Python and C bindings having no correlated receive at all is real and predates
this PR — #3046 was scoped to C++. Happy to open a tracking issue if that is
useful.

phil-opp:

Re-reviewed `a6aa3e5`, the one commit since the last review. It resolves the expiry-before-drain ordering: in `try_correlation` (`apis/rust/node/src/event_stream/mod.rs`) the drain loop now `break`s when the stream would block and `correlation_expired` runs only afterwards, so a reply that reached the receiver before a just-lapsed deadline is classified (`Match`) before the deadline is ever consulted — `Timeout` now means the reply had genuinely not arrived. The regression test pushes the reply through the scheduler rather than `pending_passthrough`, so it exercises the drain path and fails without the fix.

No new findings in the surrounding logic: the `One`-clears / `AnyOf`-keeps deadline asymmetry on `ServerRestarted` is correct, `correlation_expired` is first-registration-wins with a saturating `checked_add`, and the cxx single-server→`One` mapping in `OwnedServers::as_ref` is right. The two deferred items (a `NodeRestarted` reaching only the first-polled correlation, and the unbounded `pending_passthrough` escaping per-input queue policy) remain as discussed — both polling-only / error-path-bounded rather than correctness regressions, so a scoped follow-up seems reasonable. Minor non-blocking: the `Some(None)` stream-ended arm returns without removing the deadline entry (unlike the `Event::Stop` arm), but the stream is dead there so it's inconsequential.

---

caothu2k1:
Thanks. Took the minor point too — `89e77f55` clears the deadline on the closed-receiver route as well, so both ways of reaching `StreamEnded` leave the same state behind. You're right that it costs nothing today; it seemed worth closing so the invariant does not depend on knowing which path you arrived by.

The match now binds the polled future to a local first — matching on `pin!(self.recv_from_stream()).now_or_never()` inline keeps the borrow of `self` alive across the arms.

Writing the test turned up an assumption worth recording: an empty scripted stream does not close immediately, so an early poll reports `Ok(None)` rather than `StreamEnded`. The test polls until the stream ends and asserts every intermediate result is `Ok(None)`, which also pins that a poll on an about-to-close stream produces nothing else.

Agreed on the two deferred items being follow-up scope. I'll open tracking issues for the restart broadcast and the `pending_passthrough` bound unless you would rather they stayed on this PR.

phil-opp:

Re-reviewed only the new delta since my last review (`a6aa3e5` → `89e77f55`, `apis/rust/node/src/event_stream/mod.rs`). **The new commit looks safe.**

- It adds `self.correlation_deadlines.remove(needle)` on the closed-receiver `Some(None)` route to `StreamEnded`, matching the buffered-`Event::Stop` route (`CorrelationOutcome::StreamEnded`) that already did this — so both routes to the same error leave the same state.
- Binding `pin!(self.recv_from_stream()).now_or_never()` to a local `polled` before the `match` is correct and necessary: the temporary pinned future borrows `self` for the whole match scrutinee's lifetime, so removing from `correlation_deadlines` in an arm would otherwise fail to borrow-check. `now_or_never()` yields an owned `Option<Option<Event>>`, so the local holds no borrow.
- The test is non-vacuous: it registers a deadline, closes the stream through the real scheduler (not `pending_passthrough`, which would be found by the buffer scan ahead of this path and pass regardless), and asserts the deadline is cleared. Without the fix the `Some(None)` arm leaves the entry, so the final assertion fails. The 60s timeout cannot expire within the ~2s poll loop, so `correlation_expired` isn't what clears it — only the closed-receiver route can. It also pins that intermediate polls return `Ok(None)`.

The two deferred items (NodeRestarted reaching only the first-polled correlation; unbounded `pending_passthrough`) are untouched here and already agreed as follow-up.

---

caothu2k1:
One correction to my own earlier claim, since it affects how the green checks on this PR should be read.

I had written that `cargo test --all` was not run locally (workspace debug build does not fit on my disk) but that CI covers it. That is not true for a pull request. `Test (ubuntu-latest)`, `E2E Tests`, `Semantic contract tests` and `Benchmark regression check` are all guarded by `if: github.event_name != 'pull_request'`, so none of them ran here — the 10 green checks are fmt, clippy, check, typos, audit, licence, unwrap-budget and friends. The test suite runs on push to `main`, in the Trunk merge queue, or via `workflow_dispatch`.

So nothing had actually exercised the test suite against this branch: not me, and not this PR's CI. I have dispatched the same workflow against the branch on my fork, where `github.event_name` is `workflow_dispatch` and those jobs do run:

https://github.com/caothu2k1/dora/actions/runs/31575389029

I will report the result here. If it is red I would rather find out now than in the merge queue.

Incidentally the workflow sets `CARGO_PROFILE_DEV_DEBUG: line-tables-only` for exactly the reason I hit locally — the comment notes debuginfo is ~78% of every artifact and full-debuginfo `cargo test` overruns the runner disk. That is worth knowing for anyone else building this workspace on a small disk.

caothu2k1:
Results from dispatching the workflow against this branch on my fork, since the PR's own CI skips these jobs.

**`cargo test --all` passes on this branch.** 131 `test result: ok` lines across the workspace. That is the first time the suite has actually been run against these changes — it was not covered locally (disk) or by this PR's CI.

Two jobs went red, neither from this PR:

**`Test (ubuntu-latest)` — flaky, `dora-daemon`.** First run failed on `shutdown::tests::a_wrapper_that_exits_does_not_hide_its_surviving_child` (`binaries/daemon/src/shutdown.rs:289`, `left: Done, right: Waiting`). Re-running the *same commit* passed, which settles it, and three independent lines agree:

- No code path. `shutdown.rs` imports only `sysinfo` and `tokio::time`; zero references to `EventStream` or correlation. This PR does not touch `binaries/daemon`.
- Reproducible on demand: injecting a 50 ms sleep before the first `poll` reproduces the exact assertion locally, on this branch, with nothing else changed.
- The mechanism fits. The test spawns `sh -c 'sleep 300 & exit 0'`, which exits immediately, then asserts the wrapper is still alive. `live_children` only consults the process group for a pid in `self.confirmed`, and `confirmed` is populated solely by having seen the leader alive. If the first poll loses the race, `confirmed` is empty, the group fallback is correctly skipped, and `poll` returns `Done`.

Worth saying that guard is doing its job — it is the documented anti-pid-reuse check, not a bug. It is the test's first assertion that assumes it wins a race against an immediate `exit 0`. This will bite any PR entering the merge queue, not just this one; happy to open a separate issue if that is useful.

**`E2E Tests` — cold-cache timeout, not a test failure.** Timed out at 10 minutes with no test output at all, only `Compiling` lines. It fails identically on my fork's `main` at `cd597e70` (the unmodified base of this branch), so it is my fork's empty `rust-cache`, not the change.

Runs: branch https://github.com/caothu2k1/dora/actions/runs/31575389029 · control on `main` https://github.com/caothu2k1/dora/actions/runs/31577602686

phil-opp:

Re-reviewed the one new commit since the last pass (`89e77f5` → `d1f01b4`, "let a C++ node see that a node restarted"). Looks safe.

- The `NodeRestarted` variant is appended at the **end** of the `DoraEventType` cxx enum (after `Reload`), so existing variants keep their discriminants — no ABI shift for already-compiled C++ callers. Verified the commit is purely additive (76 additions, 0 deletions).
- `event_type` maps `Event::NodeRestarted { .. } => DoraEventType::NodeRestarted` ahead of the `_ => Unknown` catch-all, so the event that previously surfaced as `Unknown` is now identifiable.
- `event_as_node_restarted` mirrors `event_as_node_failed` exactly — pattern-matches and `bail!`s on a mismatch rather than returning a plausible id.
- The two new C++ tests are non-vacuous (one pins both the `event_type` mapping and the id round-trip; the other pins that a non-restart event is rejected).

The two deferred items from earlier passes (a `NodeRestarted` reaching only the first-polled correlation; unbounded `pending_passthrough`) are untouched here and remain follow-ups as agreed.

---

phil-opp:

Re-reviewed only the new delta since my last review (`d1f01b4` → `e421cd5`). **The latest commit looks safe.**

It is documentation-only: 25 additions to `apis/c++/node/README.md`, no code change. It adds a "Reacting to a restart with several requests in flight" section showing the caller-side loop for a `DoraEventType::NodeRestarted` event — take the id via `event_as_node_restarted`, `cancel_correlation` each request outstanding against that node, then resend under a fresh `request_id`. This is consistent with the API added in the previous commit and with the agreed follow-up scope (framework-side broadcast of a restart to every live correlation remains the open item, not addressed or misdescribed here). The snippet matches the generated bridge surface (`event_as_node_restarted`, `cancel_correlation`, `new_request_id`), and `pending_for` / `resend_against_new_instance` are clearly caller placeholders. Nothing outstanding from the earlier reviews regresses.

---

phil-opp:

The three commits added since the last review look safe to me:

- `event_as_node_restarted` correctly gates on `Event::NodeRestarted { id }` and errors on any other event, and the new `DoraEventType::NodeRestarted` mapping is straightforward. The two added C++ tests assert both the happy path and the wrong-event rejection.
- I traced the "used to be silently dropped" claim: a `NodeRestarted` for a node not in `expected` classifies as `Passthrough` and is buffered into `pending_passthrough`, so it now reaches the node's own `next_event` and surfaces as `NodeRestarted` rather than `Unknown`.
- The `StreamEnded` deadline-clearing commit removes the correlation deadline symmetrically on both the `Some(None)` and `Event::Stop` routes — no leak, and it matches the minor note from the earlier review.

No new issues found.

---

phil-opp:
**Automated review from Claude — this is a fully automated review with no human in the loop.**

This branch was rebased/force-pushed onto `main` on 2026-08-20, after the last automated review (2026-08-17). I re-checked the post-rebase state: the correlation/polling logic (`try_correlation`, `correlation_expired`, the `One`/`AnyOf` deadline handling, and the C++ `OwnedServers`/`poll_timeout` mapping) is unchanged and I found no new issues — the latest state looks safe to merge. The previously documented limitations (`NodeRestarted` reaching only the first-polled correlation, unbounded `pending_passthrough` tracked in #3197, and late replies staying buffered) remain as acknowledged follow-ups.

---

phil-opp:

This is careful work, and the thread shows it — every point raised got reproduced and fixed rather than argued away. Two things I want settled before we freeze any of it, since every symbol here is frozen for the life of 1.x the moment it lands.

**The poll functions are named for one shape and take another.** The blocking pair is `recv_service_response(request_id, &NodeId, Duration)` and `recv_service_response_from(request_id, ExpectedServers, Duration)`. The new polls are `try_recv_service_response(request_id, ExpectedServers, Option<Duration>)` — the `_from` shape under the non-`_from` name. The C++ side ships both spellings, so the Rust and C++ names of the same shape now disagree, and `docs/patterns.md:449` pairs the Rust `try_recv_service_response` with the C++ one, which is the wrong counterpart. Renaming the Rust polls to `try_recv_service_response_from` / `try_recv_action_result_from` and adding thin `try_recv_service_response(&str, &NodeId, Option<Duration>)` delegates would mirror the blocking pair exactly. About 20 lines now, unfixable after release.

**`timeout_ms == 0` means opposite things in the same header.** `recv_service_response(..., 0)` gives `Duration::ZERO` and expires immediately; `try_recv_service_response(..., 0)` gives `None` and never expires. A caller moving a request from the blocking form to the polling form by changing the function name silently loses its deadline. You flagged this yourself and asked for a maintainer call — here it is: drop `poll_timeout` and always pass `Some(clamp_pattern_timeout(timeout_ms))`. `clamp_pattern_timeout` already halves `UINT64_MAX` to a representable-but-enormous `Duration`, so `UINT64_MAX` is the "no practical deadline" spelling and `0` keeps one meaning across the whole header.

Also, there is a mis-spliced doc comment in `event_stream/mod.rs` around line 2271 — the same shape as the two-`#[test]`-attributes splice you already fixed in the C++ test module. `push_scheduler_input_for_testing`'s `#2027` block is now attached to `push_scheduler_node_restarted_for_testing`, leaving the former undocumented.

Finally, please split this up. 2348 lines carrying four independently reviewable features plus a permanent API freeze is too much to land as one unit. Suggested order:

1. `DoraEventType::NodeRestarted` + `event_as_node_restarted` + the README restart section (~90 lines). This one changes what existing C++ nodes see — restarts used to arrive as `Unknown`, and a `switch` without a `default` stops compiling under `-Werror=switch` — so it wants its own PR and its own release note.
2. `send_service_request_with_id` (~60 lines), no interaction with anything else.
3. `ExpectedServers` + `recv_*_from` + the cxx `_from` wrappers + `OwnedServers`/`parse_server_list` — the fan-out feature, reviewable without any deadline machinery.
4. `try_correlation` + `correlation_deadlines` + `cancel_correlation` + the polling example — the part that actually needs the scrutiny, now at a quarter the size.

Smaller notes for whichever PR carries them: `correlation_expired` calls `needle.to_owned()` unconditionally, so a hot poll loop allocates a `String` per in-flight request per iteration — a `get()` first and `entry()` only on the miss avoids it. The cxx entry points have no automated coverage at all (only the pure helpers are unit-tested), and `apis/c++/node/src/lib.rs` already constructs a test `EventStream` for the `merged` tests, so the harness exists. And the "a `NodeRestarted` reaches only the first-polled correlation" limitation is documented in the C++ README but not in the Rust docs or `docs/patterns.md`, which is where most readers will look.

phil-opp:

This is careful work, and the thread shows it — every point raised got reproduced and fixed rather than argued away. Two things I want settled before we freeze any of it, since every symbol here is frozen for the life of 1.x the moment it lands.

**The poll functions are named for one shape and take another.** The blocking pair is `recv_service_response(request_id, &NodeId, Duration)` and `recv_service_response_from(request_id, ExpectedServers, Duration)`. The new polls are `try_recv_service_response(request_id, ExpectedServers, Option<Duration>)` — the `_from` shape under the non-`_from` name. The C++ side ships both spellings, so the Rust and C++ names of the same shape now disagree, and `docs/patterns.md:449` pairs the Rust `try_recv_service_response` with the C++ one, which is the wrong counterpart. Renaming the Rust polls to `try_recv_service_response_from` / `try_recv_action_result_from` and adding thin `try_recv_service_response(&str, &NodeId, Option<Duration>)` delegates would mirror the blocking pair exactly. About 20 lines now, unfixable after release.

**`timeout_ms == 0` means opposite things in the same header.** `recv_service_response(..., 0)` gives `Duration::ZERO` and expires immediately; `try_recv_service_response(..., 0)` gives `None` and never expires. A caller moving a request from the blocking form to the polling form by changing the function name silently loses its deadline. You flagged this yourself and asked for a maintainer call — here it is: drop `poll_timeout` and always pass `Some(clamp_pattern_timeout(timeout_ms))`. `clamp_pattern_timeout` already halves `UINT64_MAX` to a representable-but-enormous `Duration`, so `UINT64_MAX` is the "no practical deadline" spelling and `0` keeps one meaning across the whole header.

Also, there is a mis-spliced doc comment in `event_stream/mod.rs` around line 2271 — the same shape as the two-`#[test]`-attributes splice you already fixed in the C++ test module. `push_scheduler_input_for_testing`'s `#2027` block is now attached to `push_scheduler_node_restarted_for_testing`, leaving the former undocumented.

Finally, please split this up. 2348 lines carrying four independently reviewable features plus a permanent API freeze is too much to land as one unit. Suggested order:

1. `DoraEventType::NodeRestarted` + `event_as_node_restarted` + the README restart section (~90 lines). This one changes what existing C++ nodes see — restarts used to arrive as `Unknown`, and a `switch` without a `default` stops compiling under `-Werror=switch` — so it wants its own PR and its own release note.
2. `send_service_request_with_id` (~60 lines), no interaction with anything else.
3. `ExpectedServers` + `recv_*_from` + the cxx `_from` wrappers + `OwnedServers`/`parse_server_list` — the fan-out feature, reviewable without any deadline machinery.
4. `try_correlation` + `correlation_deadlines` + `cancel_correlation` + the polling example — the part that actually needs the scrutiny, now at a quarter the size.

Smaller notes for whichever PR carries them: `correlation_expired` calls `needle.to_owned()` unconditionally, so a hot poll loop allocates a `String` per in-flight request per iteration — a `get()` first and `entry()` only on the miss avoids it. The cxx entry points have no automated coverage at all (only the pure helpers are unit-tested), and `apis/c++/node/src/lib.rs` already constructs a test `EventStream` for the `merged` tests, so the harness exists. And the "a `NodeRestarted` reaches only the first-polled correlation" limitation is documented in the C++ README but not in the Rust docs or `docs/patterns.md`, which is where most readers will look.

phil-opp:

This is careful work, and the thread shows it — every point raised got reproduced and fixed rather than argued away. Two things I want settled before we freeze any of it, since every symbol here is frozen for the life of 1.x the moment it lands.

**The poll functions are named for one shape and take another.** The blocking pair is `recv_service_response(request_id, &NodeId, Duration)` and `recv_service_response_from(request_id, ExpectedServers, Duration)`. The new polls are `try_recv_service_response(request_id, ExpectedServers, Option<Duration>)` — the `_from` shape under the non-`_from` name. The C++ side ships both spellings, so the Rust and C++ names of the same shape now disagree, and `docs/patterns.md:449` pairs the Rust `try_recv_service_response` with the C++ one, which is the wrong counterpart. Renaming the Rust polls to `try_recv_service_response_from` / `try_recv_action_result_from` and adding thin `try_recv_service_response(&str, &NodeId, Option<Duration>)` delegates would mirror the blocking pair exactly. About 20 lines now, unfixable after release.

**`timeout_ms == 0` means opposite things in the same header.** `recv_service_response(..., 0)` gives `Duration::ZERO` and expires immediately; `try_recv_service_response(..., 0)` gives `None` and never expires. A caller moving a request from the blocking form to the polling form by changing the function name silently loses its deadline. You flagged this yourself and asked for a maintainer call — here it is: drop `poll_timeout` and always pass `Some(clamp_pattern_timeout(timeout_ms))`. `clamp_pattern_timeout` already halves `UINT64_MAX` to a representable-but-enormous `Duration`, so `UINT64_MAX` is the "no practical deadline" spelling and `0` keeps one meaning across the whole header.

Also, there is a mis-spliced doc comment in `event_stream/mod.rs` around line 2271 — the same shape as the two-`#[test]`-attributes splice you already fixed in the C++ test module. `push_scheduler_input_for_testing`'s `#2027` block is now attached to `push_scheduler_node_restarted_for_testing`, leaving the former undocumented.

Finally, please split this up. 2348 lines carrying four independently reviewable features plus a permanent API freeze is too much to land as one unit. Suggested order:

1. `DoraEventType::NodeRestarted` + `event_as_node_restarted` + the README restart section (~90 lines). This one changes what existing C++ nodes see — restarts used to arrive as `Unknown`, and a `switch` without a `default` stops compiling under `-Werror=switch` — so it wants its own PR and its own release note.
2. `send_service_request_with_id` (~60 lines), no interaction with anything else.
3. `ExpectedServers` + `recv_*_from` + the cxx `_from` wrappers + `OwnedServers`/`parse_server_list` — the fan-out feature, reviewable without any deadline machinery.
4. `try_correlation` + `correlation_deadlines` + `cancel_correlation` + the polling example — the part that actually needs the scrutiny, now at a quarter the size.

Smaller notes for whichever PR carries them: `correlation_expired` calls `needle.to_owned()` unconditionally, so a hot poll loop allocates a `String` per in-flight request per iteration — a `get()` first and `entry()` only on the miss avoids it. The cxx entry points have no automated coverage at all (only the pure helpers are unit-tested), and `apis/c++/node/src/lib.rs` already constructs a test `EventStream` for the `merged` tests, so the harness exists. And the "a `NodeRestarted` reaches only the first-polled correlation" limitation is documented in the C++ README but not in the Rust docs or `docs/patterns.md`, which is where most readers will look.

caothu2k1:
Split as asked, in the order you gave. Every point from the review is addressed; the two API-freeze items are fixed in the parts that carry them.

| | PR | Content |
|---|---|---|
| 1 | #3478 | `DoraEventType::NodeRestarted` + `event_as_node_restarted` + README. Independent, lands on its own. Carries its own breaking-change note in `Changelog.md`. |
| 2 | #3479 | `send_service_request_with_id`. |
| 3 | #3480 | `ExpectedServers` + `recv_*_from` + the cxx `_from` wrappers + `OwnedServers`/`parse_server_list`. |
| 4 | #3481 | `try_correlation` + `correlation_deadlines` + `cancel_correlation` + the polling example. |

PRs from a fork can only target `main`, so they are stacked rather than chained: each shows its ancestors' commits until those land, and each body says which commit is actually new. Rebuilt on current `main` (`30f929cc`), not rebased — `send_output_internal` changed signature upstream since the old base, so the old tree would have applied textually and then failed to compile.

**On the two frozen-surface items.** Both confirmed against the code and fixed:

- The polls now take the shape their names claim: `try_recv_service_response(&str, &NodeId, _)` and `try_recv_service_response_from(&str, ExpectedServers, _)`, mirroring the blocking pair. The `docs/patterns.md` table pairs the right counterparts and gained the four rows it was missing.
- `poll_timeout` is gone. `timeout_ms` means one thing across the whole header — `0` expires immediately, `UINT64_MAX` is "no practical deadline" — so moving a request between the blocking and polling forms cannot silently drop its deadline. Six doc sites updated to match, and a test pins the invariant.

**Smaller notes.** `correlation_expired` reads with `get` before `entry`, so a poll loop no longer allocates per in-flight request per iteration. The restart limitation is now in the Rust docs and `docs/patterns.md`, not only the C++ README.

**One correction on the mis-spliced doc comment.** Real, and fixed — but it was doc-only this time. `push_scheduler_input_for_testing` was still present and still called by two tests; only its `#2027` comment block had drifted onto the neighbouring function. I checked specifically because of the earlier two-`#[test]` splice, where a function genuinely was lost.

**One correction on the test-harness note.** The cxx entry points do have no coverage — 35 tests, none touching an `EventStream`. But `merged_events_assigns_ids_to_external_streams` builds a `MergedEvents` over `stream::empty()`, not an `EventStream`, so that harness does not exist yet. `setup_integration_testing` is `pub`, so it is still reachable — just not for free. Left for a follow-up rather than bolted onto part 4.

This PR is superseded by the four above; happy to close it whenever you say.

phil-opp:
🤖 **Fully automated review by Claude — no human has vetted this.**

Short confirmation on the latest commit (`2b1d6726`), since it landed after my previous review here: no new findings, safe from my side.

I re-read it and it is exactly the content that is now carried by the four split PRs — #3478 (`DoraEventType::NodeRestarted`), #3479 (`send_service_request_with_id`), #3480 (`ExpectedServers` + `recv_*_from`), and #3481 (`try_correlation` + framework-owned deadlines). I reviewed all four at their current heads in this run and found no correctness bugs, compile breaks, unjustified 1.0 breaks, or vacuous tests. The two API-freeze items are resolved (polls mirror the blocking-pair names; `poll_timeout` dropped so `timeout_ms` has one meaning across the header).

As you and the author both note, this PR is superseded by that split and looks safe to close in favour of #3478–#3481.

---

PR #3048, opened 2026-08-05
fix(message): recover build_id from its Display form in log fields

phil-opp:
## Summary

`LogMessage::from(LogMessageHelper)` back-fills each optional identity field from the untrusted structured-logging `fields` map. Every other field is recovered with a parser matched to its type's `Display`:

- `daemon_id` → `DaemonId::from_display_str`
- `node_id` → `NodeId::from_str`
- `dataflow_id` (a bare `Uuid`) → `Uuid::parse_str`

`build_id` was the odd one out. Its `Display` wraps the UUID as `BuildId(<uuid>)`, but recovery used `Uuid::parse_str`, which rejects that wrapper. So a `build_id` emitted the idiomatic way — e.g. the coordinator's `tracing::warn!(build_id = %build_id, ...)` build warnings, which render `BuildId(<uuid>)` — was **silently dropped to `None`** when reconstructed from the `fields` map.

## Fix

Add `BuildId::from_display_str`, mirroring the existing `DaemonId::from_display_str`: it strips the `BuildId(...)` wrapper before parsing and still accepts a bare UUID for backward compatibility. Also expose `BuildId::uuid()` for symmetry with `SessionId`.

## Validation

- New regression test `build_id_recovered_from_fields_survives_display_round_trip` pins the `Display` round trip, the bare-UUID path, and rejection of garbage.
- New compile-checked doctest on `BuildId::from_display_str`.
- `cargo test -p dora-message` (141 unit + doctests), `cargo clippy -p dora-message -- -D warnings`, `cargo fmt --check` — all green.

---

---

### DIFF
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -50,7 +50,7 @@ impl From<LogMessageHelper> for LogMessage {
         LogMessage {
             build_id: helper.build_id.or(fields
                 .and_then(|f| f.get("build_id").cloned())
-                .and_then(|id| Uuid::parse_str(&id).ok().map(BuildId))),
+                .and_then(|id| BuildId::from_display_str(&id))),
             dataflow_id: helper.dataflow_id.or(fields
                 .and_then(|f| f.get("dataflow_id").cloned())
                 .and_then(|id| Uuid::parse_str(&id).ok())),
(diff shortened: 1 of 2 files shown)


phil-opp:

This PR is one of a batch opened by an automated code-review pass (Claude) over the repository. Each finding was verified against a freshly-fetched `origin/main`, checked for duplication against existing open PRs, and validated per-branch (fmt + clippy + affected-crate tests), with the full workspace suite and `/review` + `/simplify` gates run once on the combined diff. All PRs are drafts and clearly attributed as machine-generated.

### PRs opened by category

**Correctness**
- #3048 — `fix(message)`: recover `build_id` from its `Display` form (`BuildId(<uuid>)`) in log-field reconstruction; it was silently dropped because recovery used `Uuid::parse_str`.
- #3049 — `fix(coordinator)`: make the `AllNodesReady` broadcast best-effort per daemon; a single `send` failure previously tore down the whole coordinator via `?`.

**Robustness**
- #3050 — `fix(replay-node)`: skip a corrupt recorded payload (warn + continue) instead of aborting the entire replay, matching the sibling bincode branch and the recording layer's torn-record resilience.

**Optimization**
- #3051 — `perf(node)`: build the `UInt8` IPC layout once per raw send instead of rebuilding the schema/record-batch flatbuffers twice (Rust `send_output_raw` + Python raw path).

- **`InputMapping::Logs` `Display` drops `node_filter` when `min_level` is `None`** — unreachable: `FromStr` never produces that state and nothing in-repo constructs it.
- **Duplicated byte-size parser (`config.rs` `ByteSize::from_str` vs `core` `parse_byte_size`)** — low value; cross-crate dedup risks error-message churn, only a cosmetic `is_alphabetic` vs `is_ascii_alphabetic` drift.
- **`zenoh_publish` eagerly serializes `metadata_bytes` even when the schema-once path discards it** — a real waste on the streaming hot path, but a clean fix is intricate in the single hottest, most heavily-tuned function; risk/reward not worth an autonomous change.
- **cxx `Metadata::to_json` uses `.expect()` across the FFI boundary** — not triggerable today (serde_json encodes NaN/Inf as `null`, keys are `String`); latent only.
- **`ReadyOnDaemon` not idempotent against a duplicate report** — low; defensive only, the daemon self-guards duplicates; tied to #3049.
- **DDS action client spawns one feedback-reader thread per goal** (`ros2-bridge-node`) — plausibly a per-goal thread leak, but medium-confidence (couldn't verify `ros2_client` stream-termination semantics) and untestable here (no ROS distro in CI); deferred as a larger refactor.
- **`init_meter_provider` panics via `.expect()` where its sibling returns `Result`** — cold setup path (deprioritized), low value.
- **`record-node` clones `node_id`/`output_id` per message; `SystemTime::duration_since().unwrap()` in the record loop** — micro-optimization / panics only with a pre-1970 clock; negligible.
- **`download_file` sets the executable bit on every download and buffers the whole body** — behavior-change risk (exec bit likely intentional for fetched binaries), low confidence.
- **`CoordinatorStore` trait has no doctests** — internal crate, low value.

---

phil-opp:
**Automated review by Claude** — this is a fully automated review, with no human in the loop. Flagging that up front so maintainers can weigh it accordingly.

No issues found. The fix correctly recovers `build_id` from its `Display` form: `BuildId`'s `Display` emits `BuildId(<uuid>)` (`libraries/message/src/lib.rs:77-81`), which the previous `Uuid::parse_str` recovery rejected — silently dropping the id to `None` whenever a log field carried the idiomatic `%build_id` rendering. `from_display_str` strips the `BuildId(...)` wrapper before parsing and still accepts a bare UUID for backward compatibility, mirroring how `daemon_id` is recovered. Malformed wrappers fall through to a failed UUID parse and yield `None` rather than a bogus id.

The regression test is meaningful: it pins the `Display` round-trip (which would fail under the old `Uuid::parse_str` path), the bare-UUID path, and rejection of garbage — so it would fail if the fix were reverted.

---

PR #3049, opened 2026-08-05
fix(coordinator): make the AllNodesReady broadcast best-effort per daemon

phil-opp:
## Summary

`broadcast_all_nodes_ready` notifies every daemon running part of a dataflow that the start barrier has been released. The send loop used `?`, so a single `connection.send(...)` failure propagated out of the main event loop (`ReadyOnDaemon` handler → `start_inner().await?`) and **shut down the entire coordinator — every dataflow and every daemon — over one transient per-daemon error**.

A daemon that is fully gone already takes the `None` branch (log and continue); the fatal path was only hit by a *present-but-broken* connection, exactly the kind of transient condition coordination must survive. The `?` also aborted mid-loop, so daemons after the failing one never received `AllNodesReady` and their nodes stayed parked.

## Fix

Make the loop best-effort, mirroring the missing-connection branch and the existing multi-daemon *stop* path: log the failure and keep notifying the rest. This is safe because the release was already persisted and `ready_barrier_released` set *before* the loop (`release_barrier_message`), so any daemon that misses the frame is replayed on reconnect (#2998). The sibling `replay_all_nodes_ready` already uses this same best-effort posture for the missing-connection case.

## Validation

- New test `ready_broadcast_is_best_effort_when_a_daemon_send_fails` drives `broadcast_all_nodes_ready` with a daemon connection whose channel is closed (so `send` errors) and asserts the broadcast still returns `Ok` and still records the barrier as released. Without the fix the error propagates and the assertion fails.
- `cargo test -p dora-coordinator` (123 tests), `cargo clippy -p dora-coordinator -- -D warnings`, `cargo fmt --check` — all green.

---

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -4679,15 +4679,28 @@ async fn broadcast_all_nodes_ready(
         tracing::warn!(dataflow = %uuid, "failed to persist ready-barrier release: {e}");
     }
 
-    // notify all machines that run parts of the dataflow
+    // notify all machines that run parts of the dataflow.
+    //
+    // This is best-effort per daemon: a broken (but not yet dropped)
+    // connection makes `send` fail, but that must not tear down coordination
+    // for every other dataflow and daemon. `ready_barrier_released` was
+    // already set and persisted above, so a daemon that misses this frame is
+    // replayed on reconnect (#2998) -- a dropped send is recoverable. Log and
+    // move on instead of aborting the coordinator's event loop (the same
+    // best-effort posture the multi-daemon stop path already takes), and keep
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the change making the `AllNodesReady` per-daemon broadcast best-effort: a single `connection.send` failure is now logged instead of propagating `?` out of the coordinator event loop and tearing down every dataflow/daemon over one transient per-daemon error. This is safe because `ready_barrier_released` is persisted before the loop, so a daemon that misses the frame is replayed on reconnect (#2998), and the loop no longer aborts mid-iteration so later daemons still get notified. The test drives it with a closed-channel connection and asserts `Ok` plus `ready_barrier_released`. No important issues found.

---

PR #3050, opened 2026-08-05
fix(replay-node): skip a corrupt recorded payload instead of aborting replay

phil-opp:
## Summary

The replay loop deserializes each recorded entry in two steps. A `bincode::deserialize` failure on the event envelope already warns and `continue`s to the next entry, but a `decode_arrow_ipc` failure on the payload propagated with `?` out of `main`, **terminating the replay of every remaining message** (and, under `DORA_REPLAY_LOOP`, the whole loop) over a single corrupt-but-complete record.

That is inconsistent with the sibling bincode branch and with the recording layer, which deliberately drops torn trailing records rather than failing the replay (see `truncated_record_body_stops_gracefully`).

## Fix

Extract the payload decode into `decode_recorded_payload` and, on error, warn and `continue` — matching the bincode branch. A genuinely fatal condition (failure to send to the daemon) still aborts.

## Validation

- Tests cover the extracted helper: a metadata-only (`None`) payload decodes to an empty array, a valid IPC stream round-trips, and a corrupt payload returns `Err` (the signal `main` uses to skip the record).
- `cargo test -p dora-replay-node` (7 tests), `cargo clippy -p dora-replay-node`, `cargo fmt --check` — all green.

---

---

### DIFF
--- a/binaries/replay-node/src/main.rs
+++ b/binaries/replay-node/src/main.rs
@@ -24,6 +24,15 @@ fn decode_recorded_payload(data: Option<&[u8]>) -> eyre::Result<ArrayData> {
     }
 }
 
+/// Whether a replay pass emitted nothing usable: every matching record was
+/// skipped as undecodable. This is a hard failure (a systematically corrupt or
+/// format-drifted recording), distinct from a pass that legitimately matched no
+/// records at all (`replayed == 0 && skipped == 0` -- the node simply had no
+/// recorded output), which is not a failure.
+fn replay_emitted_nothing_usable(replayed: u64, skipped: u64) -> bool {
+    replayed == 0 && skipped > 0
+}
+
 /// Nanoseconds to sleep before emitting an entry, given the previous entry's
 /// recording offset, this entry's offset, and the replay `speed`.
 ///
@@ -66,6 +75,7 @@ fn main() -> eyre::Result<()> {
         // output — is honored, preserving cross-node alignment on replay.
         let mut prev_offset: u64 = 0;
         let mut replayed = 0u64;
+        let mut skipped = 0u64;
 
         while let Some(entry) = reader.next_entry()? {
             if entry.node_id != replay_node {
@@ -88,6 +98,7 @@ fn main() -> eyre::Result<()> {
                             "warning: failed to deserialize event for {}/{}: {e}",
                             entry.node_id, entry.output_id
                         );
+                        skipped += 1;
                         continue;
                     }
                 };
@@ -115,6 +126,7 @@ fn main() -> eyre::Result<()> {
                                 "warning: skipping undecodable payload for {}/{output_id}: {e:#}",
                                 entry.node_id
                             );
+                            skipped += 1;
                             continue;
                         }
                     };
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the change that extracts `decode_recorded_payload` and, on a decode error, warns and `continue`s instead of propagating `?` out of `main` — so one corrupt-but-complete record no longer aborts the rest of the replay (and, under `DORA_REPLAY_LOOP`, the whole loop). I checked the loop: `prev_offset` is updated before the decode, so skipping a record via `continue` does not corrupt inter-message pacing, and this matches the sibling bincode branch. The three helper tests assert meaningful behavior (None → empty array, valid IPC round-trips, corrupt → `Err`). No important issues found.

---

phil-opp:
Correct on the specifics — pacing is unharmed (`prev_offset` is updated before the `continue`), and "corrupt" cannot mis-fire on an empty or zero-row batch, since `encode_arrow_ipc` always writes schema + one batch + EOS.

The gap is the systematic case. A recording that is undecodable end-to-end — format drift of the kind #2366 caused, or a truncated file — now prints per-record warnings, reports `replayed 0 messages` at `binaries/replay-node/src/main.rs:106`, and **exits 0**. The dataflow then completes "successfully" having emitted nothing, which turns a data-integrity failure into a silent wrong answer. Suggest tracking a `skipped` counter, including it in the summary line, and failing when `skipped > 0 && replayed == 0`.

phil-opp:
Good catch — that silent-empty-success case is exactly the failure mode this should not create. Fixed in `369a789`:

- Track a `skipped` counter across **both** skip paths (the bincode-envelope skip and the Arrow-payload skip), so systematic format drift / truncation is fully accounted for.
- Surface it in the summary line: `replayed N messages (M skipped)`.
- Fail the replay (non-zero exit) when a pass matched records but decoded none of them — `replayed == 0 && skipped > 0`. A pass that matched no records at all (the node simply had no recorded output) stays a clean success.

The decision is factored into `replay_emitted_nothing_usable(replayed, skipped)` with unit tests covering all four quadrants.

---

phil-opp:
The merge-queue failure is **not** from this PR. The failing `Test (ubuntu-latest)` job is a compile error in `dora-coordinator`:

```
```

This PR only modifies `binaries/replay-node/src/main.rs` (no coordinator changes), so it can't produce this error. It's a semantic merge conflict from another PR in the same merge-queue batch (`trunk-merge/pr-2910`) that adds a second `RunningDataflow` construction omitting the `ready_barrier_released` field. On current `origin/main` the only such construction (`handlers.rs:658`) sets the field correctly, so `main` itself builds — the batch-mate was written against an older base.

No action needed here; the queue's batch-bisection should eject the offending PR and re-test this one. I'll keep watching.

---

phil-opp:

Reviewed the latest commit `369a789`, which addresses the systematic-empty-success gap raised earlier. It fixes the issue: a `skipped` counter now covers both skip paths (bincode envelope at `binaries/replay-node/src/main.rs:101` and Arrow payload at `:118`), the summary line surfaces it, and `replay_emitted_nothing_usable(replayed, skipped)` bails when a pass matched records but decoded none of them (`replayed == 0 && skipped > 0`) — while a pass that matched no records at all (`0, 0`, node simply had no recorded output) stays a clean success. `replay_file`/`replay_node` in the bail message are in scope, `skipped` resets per pass, and the four unit tests assert real behavior (they would fail if the boundary were wrong). No new issues found.

---

PR #3054, opened 2026-08-06
fix(daemon): don't start a stopping dataflow when a node death completes the barrier

### LINKED ISSUE #3053: daemon: #2970 can call `dataflow.start()` on an already-stopping dataflow (barrier-completed-by-node-death start isn't gated on `stop_sent`)
**This issue was filed by an automated, scheduled Claude review of recent commits/PRs. Please verify before acting.**

## Summary

**#2970** (`6ccb374`, "start the dataflow when a node death completes the startup barrier") added a new call to `dataflow.start()` inside `handle_node_stop_inner`, gated only on `!dataflow.dataflow_started`. Unlike the other three `start()`-on-`AllNodesReady` call sites, this one can fire **after the dataflow has already been torn down by `stop_all`**, spawning timer tasks and flipping `dataflow_started = true` on a stopping dataflow.

## Why this path is unique

The other `start()` triggers are all driven by a node *subscribing* or being *added*:
- coordinator `AllNodesReady` (`lib.rs:~2370`)
- node `Subscribe` (`lib.rs:~4680`)
- `AddNode` (`lib.rs:~2841`)

After `stop_all`, `subscribe_channels` has been `.drain()`ed (`running_dataflow.rs:586`), so none of those can fire post-stop. But `PendingNodes::handle_dataflow_stop` (`pending.rs:209-215`, called from `stop_all`) only removes **dynamic** nodes from the barrier cohort — a **non-dynamic** pending cohort member is left in `local_nodes`. #2970 introduces a start trigger driven by a *node death*, which is reachable after `stop_all` because the stop ladder itself kills that still-pending node.

## Affected code

`binaries/daemon/src/lib.rs:5454-5464` (`handle_node_stop_inner`):

```rust
if matches!(status, DataflowStatus::AllNodesReady) && !dataflow.dataflow_started {
    // ... log ...
    dataflow.start(&self.events_tx, &self.clock).await?;
}
```

`stop_all` sets `self.stop_sent = true` (`running_dataflow.rs:604`) and drains `subscribe_channels`, but does **not** clear the barrier for non-dynamic pending nodes.

## Triggering sequence

1. Dataflow starting; cohort `[A, B]` non-dynamic, not yet subscribed, `dataflow_started = false`.
2. `A` subscribes; `B` still pending in `local_nodes`.
3. `dora stop` → `stop_all`: `disable_restart` on all nodes, `subscribe_channels` drained, processes taken and kills scheduled, `stop_sent = true`. The barrier is left with `B` (non-dynamic) still pending.
4. `B` is SIGKILLed by the stop ladder → its `SpawnedNodeResult` (generation matches) → `handle_node_stop_inner` → `pending_nodes.handle_node_stop(B)` empties the cohort → returns `AllNodesReady`.
5. #2970's guard `matches!(AllNodesReady) && !dataflow_started` is true → `dataflow.start()` spawns timer tasks and sets `dataflow_started = true` on a dataflow that was already fully torn down.

## Wrong outcome

Benign in the steady state (the spawned `Timer` events find no `subscribe_channels`, and `finish_dataflow` later drops the `RunningDataflow`, cancelling the `_timer_handles`), but there are real observable effects during teardown:
- A brief burst of no-op timer tasks is spawned after stop.
- `dataflow_started` flips to `true` during teardown, so a racing `dora list` / metrics read can momentarily report the dataflow as `Running`.
- A racing `AddNode` (`lib.rs:~2841`, which also checks `dataflow_started`) would spawn real timers on the stopping dataflow.

## Suggested fix

Gate the #2970 start on `!dataflow.stop_sent` (the other callers are implicitly protected by the drained `subscribe_channels`; this death-driven one is not):

```rust
if matches!(status, DataflowStatus::AllNodesReady)
    && !dataflow.dataflow_started
    && !dataflow.stop_sent
{
    // ...
    dataflow.start(&self.events_tx, &self.clock).await?;
}
```

## Confidence

Confirmed reachable at the code level; low severity (no case found where the dataflow fails to finish, since every remaining non-dynamic node still has a scheduled kill). Worth the one-line guard to avoid the `Running`-during-teardown flip and the `AddNode` race.

phil-opp:
Fixes #3053.

A startup-barrier completion can arrive *during* teardown, and every trigger that
turns one into `dataflow.start()` was reachable that way:

- **node death** (`handle_node_stop_inner`, added in #2970) — `stop_all` SIGKILLs a
  still-pending non-dynamic cohort member when the grace window expires, and that
  death completes the barrier;
- **subscribe** — `stop_all` drains the *existing* `subscribe_channels` but does not
  close the node listener, so a pending cohort member can still subscribe inside the
  grace window. `Daemon::subscribe` has a dedicated `stop_sent` branch for exactly
  that node; if it is the last pending member, its subscription completes the barrier;
- **`RemoveNode`** — removing the last pending member completes the barrier, and the
  request can land mid-teardown;
- **coordinator `AllNodesReady`** — had no guard at all. The release can legitimately
  land after a local stop: the daemon can stop itself (`trigger_manual_stop`), and the
  coordinator replays `AllNodesReady` to a reconnecting daemon.

Starting a stopping dataflow spawns timer tasks that tick into the teardown until the
dataflow is dropped, and flips `dataflow_started` on — which is the flag a racing
`AddNode` reads to decide it may spawn timer tasks of its own.

## Change

- Add `RunningDataflow::should_start_on_barrier_completion`, which gates on
  `AllNodesReady && !dataflow_started && !stop_sent`, and route the node-death,
  subscribe, and `RemoveNode` call sites through it.
- Gate the coordinator `AllNodesReady` path on `!stop_sent`.
- Unit test drives `stop_all` (rather than assigning `stop_sent`) so it tracks the
  real teardown entry point.

Nothing is re-stranded by suppressing the start: parked subscribers are answered by
`answer_subscribe_requests` inside `handle_node_stop`, independently of the start
decision, so the #2967 guarantee still holds.

## Testing

- `cargo test -p dora-daemon` — 234 passed
- `cargo test --all` (Python + examples excluded) — green apart from the two
  `rmw_zenoh_pubsub` tests that need multicast
- `cargo fmt --all -- --check`, `cargo clippy -p dora-daemon -- -D warnings`,
  `cargo check --examples` — clean

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -7860,7 +7860,13 @@ impl Daemon {
         // non-cohort survivor that was answered `Ok(())` (since #2933) isn't
         // left parked on a dataflow that never starts and never tears down
         // (#2967). `start()` is idempotent and guarded by `dataflow_started`.
-        if matches!(status, DataflowStatus::AllNodesReady) && !dataflow.dataflow_started {
+        //
+        // `should_start_on_barrier_completion` also gates on `!stop_sent`
+        // (dora-rs/dora#3053): unlike the subscribe / `AddNode` callers — which
+        // can no longer fire after `stop_all` drained `subscribe_channels` —
+        // this death-driven trigger is reachable during teardown, because the
+        // stop ladder itself kills a still-pending non-dynamic cohort member.
+        if dataflow.should_start_on_barrier_completion(&status) {
             logger
                 .log(
                     LogLevel::Info,
@@ -10368,6 +10374,34 @@ mod fault_tolerance_tests {
         assert!(!df.dataflow_started);
         df.start(&events_tx, &clock).await.unwrap();
         assert!(df.dataflow_started);
+    }
+
+    // dora-rs/dora#3053: the death-driven start trigger in
+    // `handle_node_stop_inner` (added in #2970) can fire during teardown,
+    // because `stop_all` kills a still-pending non-dynamic cohort member and
+    // that death completes the startup barrier. `should_start_on_barrier_completion`
+    // gates the start on `!stop_sent` so a stopping dataflow is never (re)started.
+    #[test]
+    fn barrier_completion_does_not_start_a_stopping_dataflow() {
+        let mut df = test_dataflow();
+
+        // A fresh dataflow whose barrier just completed should start.
+        assert!(df.should_start_on_barrier_completion(&DataflowStatus::AllNodesReady));
+
+        // Once stop has been sent, the same completion must not start it.
+        df.stop_sent = true;
+        assert!(
+            !df.should_start_on_barrier_completion(&DataflowStatus::AllNodesReady),
+            "a stopping dataflow must not be started by a barrier-completing node death"
+        );
+
+        // A `Pending` status never starts, stopping or not.
+        df.stop_sent = false;
+        assert!(!df.should_start_on_barrier_completion(&DataflowStatus::Pending));
+
+        // An already-started dataflow is not started again either.
+        df.dataflow_started = true;
+        assert!(!df.should_start_on_barrier_completion(&DataflowStatus::AllNodesReady));
     }
 
     fn test_running_node() -> RunningNode {
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -186,7 +186,6 @@ pub(crate) use running_dataflow::{
 use crate::{
     extension_table::{ExtensionKey, ExtensionTable},
     extract_err_from_stderr::extract_err_from_stderr,
-    pending::DataflowStatus,
 };
 use dora_tensor_pool::TensorPoolManager;
 use shared_memory_extended::ShmemConf;
@@ -3710,7 +3709,14 @@ impl Daemon {
                                 &mut dataflow.cascading_error_causes,
                             )
                             .await?;
-                        if ready {
+                        // Not gated on `dataflow_started` — `start()` is
+                        // idempotent — but it must not resurrect a dataflow that
+                        // is already tearing down (dora-rs/dora#3053). The
+                        // release can legitimately arrive after a local stop: the
+                        // daemon can stop on its own (`trigger_manual_stop`), and
+                        // the coordinator replays `AllNodesReady` to a daemon
+                        // that reconnects.
+                        if ready && !dataflow.stop_sent {
                             logger.log(LogLevel::Info, None,
                                 Some("daemon".into()),
                                 "coordinator reported that all nodes are ready, starting dataflow",
@@ -4310,8 +4316,13 @@ impl Daemon {
                                     )
                                     .await;
                                 match status {
-                                    Ok(DataflowStatus::AllNodesReady)
-                                        if !dataflow.dataflow_started =>
+                                    // Removing the last pending cohort member
+                                    // completes the barrier, and a `RemoveNode`
+                                    // can land mid-teardown, so this shares the
+                                    // `!stop_sent` gate with the subscribe and
+                                    // node-death triggers (dora-rs/dora#3053).
+                                    Ok(status)
+                                        if dataflow.should_start_on_barrier_completion(&status) =>
                                     {
                                         logger
                                             .log(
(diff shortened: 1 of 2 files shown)


phil-opp:

Reviewed the extraction of the barrier-completion start decision into `RunningDataflow::should_start_on_barrier_completion`, which additionally gates on `!stop_sent`. This correctly suppresses the death-driven start trigger during teardown, where `stop_all` SIGKILLs a still-pending non-dynamic cohort member and that death completes the startup barrier — avoiding the momentary `dataflow_started` flip and no-op timer spawns. The unit test covers the `AllNodesReady`/`stop_sent`/`dataflow_started`/`Pending` combinations. No important issues found.

---

phil-opp:
I found an issue: the `!stop_sent` gate is applied only to the death-driven start site, but the subscribe-driven site has the same shape and is reachable during teardown, so the class of bug this PR targets can still occur.

The new guard goes through `should_start_on_barrier_completion` at the death path (`handle_node_stop_inner`, lib.rs:5460). But the subscribe start site still gates only on `!dataflow.dataflow_started`:

```rust
// lib.rs:4680
DataflowStatus::AllNodesReady if !dataflow.dataflow_started => {
    …
    dataflow.start(&self.events_tx, &self.clock).await?;
}
```

and the `RemoveNode` start site (lib.rs:2965) has the identical `!dataflow_started`-only pattern.

This is reachable. `handle_dataflow_stop` (pending.rs:201) removes only *dynamic* nodes from the barrier set (`local_nodes`); a still-pending **non-dynamic** cohort member stays in `local_nodes` after `stop_all`. `Self::subscribe` explicitly anticipates a node subscribing *after* the stop was sent — the `if dataflow.stop_sent { … send Stop … }` branch at lib.rs:5311 — and still inserts the channel and runs `handle_node_subscription`, which removes the node from `local_nodes` and can return `AllNodesReady`. When that late subscribe completes the barrier, the unguarded site above fires `start()` on the stopping dataflow — the same spurious start / `dataflow_started` flip this PR is meant to prevent. (Equivalently, framed as an event race: for a killed-while-pending non-dynamic node, if its `Subscribe` is dequeued before its death event, the barrier completes via subscribe rather than death.)

The rationale that subscribe "can no longer fire after `stop_all` drained `subscribe_channels`" doesn't hold: the drain only sends `Stop` to *already-subscribed* nodes; it neither removes the pending node from `local_nodes` nor blocks a late `Subscribe`.

Suggest routing the subscribe site — and the `RemoveNode` site — through `should_start_on_barrier_completion` as well, so all three barrier-completion start paths gate on `!stop_sent` consistently. A regression test that completes the barrier via a post-`stop_sent` subscribe (not a death) would lock this down.

---

---

phil-opp:

I found an issue: the fix only closes one of the three barrier-completion start paths, so the bug class this PR targets is still reachable during teardown.

The PR routes the death-driven start site (`handle_node_stop_inner`, `lib.rs:5460`) through `should_start_on_barrier_completion`, which correctly adds the `!stop_sent` gate. But the other two `AllNodesReady` start sites are left with the old `!dataflow_started`-only check:

- subscribe-driven start (`lib.rs:4680`): `DataflowStatus::AllNodesReady if !dataflow.dataflow_started => { … dataflow.start(...) }`
- node-removal-driven start (`lib.rs:2965`): identical `!dataflow_started`-only pattern.

The PR's rationale is that these "can no longer fire after `stop_all` drained `subscribe_channels`", but that doesn't hold for the subscribe path. `Self::subscribe` explicitly anticipates a node connecting *after* stop was sent — the `if dataflow.stop_sent { … send Stop … }` branch at `lib.rs:5316` still inserts the channel, and the caller then runs `handle_node_subscription`, which removes the node from `local_nodes` and can return `AllNodesReady`. `handle_dataflow_stop` removes only *dynamic* nodes from the barrier set, so a still-pending **non-dynamic** cohort member remains in `local_nodes` after `stop_all`. If that node was SIGKILLed while pending and its `Subscribe` is dequeued before its death event, the barrier completes via the subscribe site at `lib.rs:4680` — firing `start()` on the stopping dataflow, i.e. exactly the spurious-start / `dataflow_started`-flip this PR is meant to prevent, just through a sibling code path.

Suggest routing both the subscribe site (4680) and the node-removal site (2965) through `should_start_on_barrier_completion` so all three barrier-completion start paths gate on `!stop_sent` consistently, and adding a regression test that completes the barrier via a post-`stop_sent` subscribe rather than a death.

---

PR #3055, opened 2026-08-06
fix(daemon): fold local exit-before-subscribe into the latched barrier start verdict

### LINKED ISSUE #3052: daemon: latched cross-daemon barrier starts the dataflow while simultaneously failing a surviving cohort node's `Node()` init (local `exited_before_subscribe` ignored by the start verdict)
**This issue was filed by an automated, scheduled Claude review of recent commits/PRs. Please verify before acting.**

## Summary

In `PendingNodes::update_dataflow_status`, the latch branch added by **#2990** (`596f27b`, "latch the cross-daemon barrier so late subscribers are answered") decides whether to report `AllNodesReady` (which makes the daemon call `dataflow.start()`) using **only the coordinator's external failure list**, while the reply actually handed to each subscriber is computed by `answer_subscribe_requests`, which **prioritizes the daemon's *local* `exited_before_subscribe`**. The two halves use different sources, so a daemon can **start the dataflow and at the same moment fail a healthy surviving cohort node's `Node()` init** with `"Node <X> exited before initializing dora"`.

This is a new inconsistency created by the interaction of three same-day PRs: **#2990** (introduced the latch), **#3013** (`2ce3a97`, coordinator now releases/replays the ready barrier as *success* to a daemon that disconnected while still pending — so the coordinator's list can legitimately be empty even though the daemon locally lost a node), and **#2970** (`6ccb374`, the returned `AllNodesReady` status now actively drives `dataflow.start()`).

## Affected code

`binaries/daemon/src/pending.rs:273-290` (latch branch):

```rust
if let Some(exited_before_subscribe) = self.external_ready.clone() {
    let barrier_succeeded = exited_before_subscribe.is_empty();   // <-- external list ONLY
    self.answer_subscribe_requests(exited_before_subscribe, cascading_errors).await;
    return Ok(if barrier_succeeded {
        DataflowStatus::AllNodesReady   // <-- caller calls dataflow.start()
    } else {
        DataflowStatus::Pending
    });
}
```

`binaries/daemon/src/pending.rs:308-314` (`answer_subscribe_requests` — local list wins):

```rust
let node_exited_before_subscribe = match self.exited_before_subscribe.as_slice() {
    [first, ..] => Some(first),                              // local failure takes priority
    [] => match exited_before_subscribe_external.as_slice() {
        [first, ..] => Some(first),
        [] => None,
    },
};
```

So when `self.exited_before_subscribe = [X]` but `external_ready = Some([])`:
- `barrier_succeeded = true` → returns `AllNodesReady` → `dataflow.start()` (timers spawn, `dataflow_started = true`).
- `answer_subscribe_requests` names local node `X` → every **cohort** subscriber (e.g. a healthy node `Y`) gets `Err("Node X exited before initializing dora ...")` and fails its own `Node()` init.

Every other path that returns `AllNodesReady` does so only when there is no failure; the latch branch is the sole place the start verdict and the subscriber replies can disagree.

## How the empty-external / non-empty-local state is reached (reconnect during startup)

`self.exited_before_subscribe` holds **cohort members only** and is normally reported to the coordinator (`report_nodes_ready` sends `self.exited_before_subscribe`), so under normal broadcast the coordinator's list would *include* `X`. The divergence requires the daemon's report to be skipped — exactly the reconnect path #3013 handles:

1. Multi-daemon dataflow. Daemon `D` has local cohort `{A, X, Y}` (+ external nodes on other daemons). `A` subscribes; `X`, `Y` still pending, so `D` has **not** reported ready (`local_nodes` non-empty, `reported_init_to_coordinator = false`).
2. `D` briefly disconnects from the coordinator (network blip). Local nodes keep running (reconnect feature, #2029).
3. Coordinator removes `D` from `pending_daemons`; if `D` was the last pending daemon and the spawn was OK, `on_daemon_disconnect` pushes `DisconnectAction::ReleaseReadyBarrier` (`binaries/coordinator/src/lib.rs:3180`) → `broadcast_all_nodes_ready` with `exited_before_subscribe = dataflow.exited_before_subscribe.clone()` (`lib.rs:4564`) = **`[]`** (the coordinator never learned about `X`). `ready_barrier_released = true`, persisted.
4. Still disconnected, `D`'s node `X` exits before subscribing → `handle_node_stop` pushes `X` into `self.exited_before_subscribe`.
5. `D` reconnects → coordinator replays the release (`replay_all_nodes_ready`, `lib.rs:2571/2608`) carrying `[]` → `D.external_ready = Some([])`.
6. `Y` subscribes → `update_dataflow_status` latch branch: `barrier_succeeded = true` → **dataflow starts**, yet `Y` (in cohort) is answered `Err("Node X exited before initializing dora")` and dies at init.

## Wrong outcome

A partially-started dataflow: timers and already-subscribed nodes (e.g. `A`) run, while a healthy surviving cohort node (`Y`) is killed at `Node()` init and a genuinely-dead node (`X`) is silently missing. Either the dataflow should not start (local failure is real) or the surviving nodes should start cleanly (trusting the global success verdict) — doing both is incorrect. This is the same "survivor left on an inconsistently started dataflow" class #2970/#2967 set out to remove, here re-introduced in the reconnect path.

## Suggested fix

Make the start verdict consistent with the replies the subscribers actually receive by folding in the local list:

```rust
let barrier_succeeded =
    exited_before_subscribe.is_empty() && self.exited_before_subscribe.is_empty();
```

With this, a local cohort loss keeps the dataflow from starting (returns `Pending`) and the cohort subscribers still receive the failure — the two halves agree. A regression test in the `mod tests` of `pending.rs` seeding `self.exited_before_subscribe = [X]` + `external_ready = Some(vec[])` and asserting the returned status is `Pending` (not `AllNodesReady`) while a parked cohort subscriber gets an `Err` would pin it.

## Confidence

Medium-high. The code-level inconsistency between the start decision and the subscriber replies is unambiguous and localized. The reachability path depends on the coordinator's disconnect-release (#3013) firing with the daemon's own `exited_before_subscribe` not yet reported, which is the specific scenario that PR targets.

phil-opp:
## Summary

Fixes #3052.

In `PendingNodes::update_dataflow_status`, the latch branch added by #2990 decides whether to report `AllNodesReady` (which drives `dataflow.start()`) using **only** the coordinator's external failure list:

```rust
let barrier_succeeded = exited_before_subscribe.is_empty(); // external list ONLY
```

But the reply actually handed to each subscriber is computed by `answer_subscribe_requests`, which **prioritizes this daemon's *local* `exited_before_subscribe`**. The two halves used different sources, so a daemon could **start the dataflow and at the same moment fail a healthy surviving cohort node's `Node()` init** with `"Node X exited before initializing dora"`.

The divergent state (empty external list, non-empty local list) is reachable via the #3013 reconnect path: a daemon disconnects during startup before reporting its own loss, the coordinator releases the ready barrier as success with an empty list, a local cohort node then dies before subscribing, and on reconnect the replayed release carries `[]`. A later cohort subscribe then hits the latch branch, starts the dataflow, yet answers the surviving cohort member with an error.

This re-introduces the "survivor left on an inconsistently started dataflow" class that #2970/#2967 set out to remove.

## Change

- Make the start verdict consistent with the replies subscribers actually receive by additionally requiring `self.exited_before_subscribe.is_empty()`. A local cohort loss now keeps the dataflow `Pending` and the cohort subscribers still receive the failure — the two halves agree.
- Add a regression test (`latch_verdict_respects_a_local_exit_the_coordinator_never_saw`) that seeds an empty external release + a local `exited_before_subscribe`, and asserts the status is `Pending` (not `AllNodesReady`) while the parked cohort subscriber gets an `Err` naming the dead node.

## Testing

- `cargo test -p dora-daemon --lib pending` — 20 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01KwYrfJpHX6PSwoewwKyA9z)_

### DIFF
--- a/binaries/daemon/src/pending.rs
+++ b/binaries/daemon/src/pending.rs
@@ -273,7 +273,19 @@ impl PendingNodes {
                 if let Some(exited_before_subscribe) = self.external_ready.clone() {
                     // The barrier is already down; a late subscriber must
                     // not wait for an event that has been and gone.
-                    let barrier_succeeded = exited_before_subscribe.is_empty();
+                    //
+                    // The start verdict must agree with the reply each
+                    // subscriber actually receives. `answer_subscribe_requests`
+                    // prioritizes this daemon's LOCAL `exited_before_subscribe`,
+                    // so folding it in here keeps the two halves consistent
+                    // (dora-rs/dora#3052): a local cohort loss the coordinator
+                    // never learned about (e.g. a node that died before
+                    // subscribing while the daemon was disconnected, so the
+                    // released barrier carried an empty external list — #3013)
+                    // must keep the dataflow from starting, not start it while
+                    // simultaneously failing a surviving cohort node's init.
+                    let barrier_succeeded = exited_before_subscribe.is_empty()
+                        && self.exited_before_subscribe.is_empty();
                     self.answer_subscribe_requests(exited_before_subscribe, cascading_errors)
                         .await;
                     // Only report ready if the barrier actually succeeded.
@@ -976,6 +988,46 @@ mod tests {
         assert!(
             !matches!(status, DataflowStatus::AllNodesReady),
             "adding a node must not start a dataflow whose barrier failed; got {status:?}"
+        );
+    }
+
+    /// dora-rs/dora#3052: on the latched barrier the START verdict must agree
+    /// with the reply each subscriber actually receives.
+    /// `answer_subscribe_requests` prioritizes this daemon's LOCAL
+    /// `exited_before_subscribe`, so when the coordinator released the barrier
+    /// as success (empty external list — the #3013 reconnect path, where this
+    /// daemon's own loss was never reported) but a local cohort member died
+    /// before subscribing, the dataflow must report `Pending`, not
+    /// `AllNodesReady`. Otherwise it would start while simultaneously failing a
+    /// surviving cohort node's `Node()` init — the exact "survivor left on an
+    /// inconsistently started dataflow" class #2970/#2967 set out to remove.
+    #[tokio::test]
+    async fn latch_verdict_respects_a_local_exit_the_coordinator_never_saw() {
+        let (dead, survivor) = (node("locally-dead"), node("survivor"));
+        let mut p = external_barrier();
+        // Both are cohort members that have already left `local_nodes` (one
+        // died, one subscribed), so the latch branch is the one exercised.
+        p.cohort.insert(dead.clone());
+        p.cohort.insert(survivor.clone());
+        // The coordinator released the barrier as SUCCESS: its external list is
+        // empty because this daemon's local loss was never reported upward.
+        release(&mut p, Vec::new()).await;
+        // ...but locally a cohort member died before subscribing.
+        p.exited_before_subscribe.push(dead.clone());
+
+        let mut rx = park(&mut p, &survivor);
+        let status = settle(&mut p).await;
+
+        assert!(
+            matches!(status, DataflowStatus::Pending),
+            "a local cohort loss must keep the dataflow from starting even when \
+             the coordinator's external verdict was success; got {status:?}"
+        );
+        let err = reply_of(&mut rx)
+            .expect_err("the surviving cohort member must inherit the local startup failure");
+        assert!(
+            err.contains("locally-dead"),
+            "the failure must name the locally dead node: {err}"
         );
     }
 }


phil-opp:

Reviewed the change that folds `self.exited_before_subscribe.is_empty()` into the latched-barrier start verdict so it agrees with the reply `answer_subscribe_requests` actually hands each subscriber (which prioritizes the local list). The verdict and the per-subscriber replies now use the same source, so a local cohort loss the coordinator never learned about (the #3013 empty-external-list reconnect path) keeps the dataflow `Pending` instead of starting it while failing a surviving node's init. The regression test exercises exactly that path. No important issues found.

---

PR #3056, opened 2026-08-06
fix(tensor-pool): seed the pool-id counter randomly so restarts don't collide

### LINKED ISSUE #3015: Memory-pool ids collide across node restarts: a restarted node cannot re-register its pool
## Summary

Memory-pool shared-memory names are derived from a **process-local** counter, so a
node that crashes and restarts regenerates the exact same name as its previous
incarnation. If the old pool has not been reclaimed by then, the restarted node
cannot register a pool at all — `ShmemConf::create()` fails on the leftover segment,
and the daemon would independently reject the registration as a duplicate.

Split out of #2881, which fixed *when* pools are reclaimed. This is the separate
question of pool ids not being unique across incarnations.

## Evidence

The counter is a process-global static seeded at `0`
([`apis/python/node/src/lib.rs:137-138`](https://github.com/dora-rs/dora/blob/main/apis/python/node/src/lib.rs#L137-L138)):

```rust
/// Counter to make pinned memory buffer IDs unique across registrations.
static PINNED_COUNTER: LazyLock<std::sync::Mutex<u64>> = LazyLock::new(|| std::sync::Mutex::new(0));
```

It is unique across registrations *within one process*, but a fresh process always
starts at `1`. The segment name mixes it with two values that do not change across a
restart (`apis/python/node/src/lib.rs:1965-1968`):

```rust
let shmem_name = format!("dora_pool_{}_{}_{}", self.dataflow_id, self.node_id, pool_counter);
```

So incarnation 2 of a node re-derives `dora_pool_{dataflow}_{node}_1` and calls
`ShmemConf::new().os_id(&shmem_name)…create()`. That fails while the old segment
exists — confirmed against `shared_memory_extended` 0.13.0:

```
first create: Ok
second create: ERR <-- collision: Shared memory OS specific ID already exists
```

which is exactly what the call site's own error message anticipates: *"name collision
with another node or leftover segment"*. Independently, the daemon would reject the
registration too — `MemoryPoolManager::register_memory_pool` returns
`"Memory pool with ID {} already registered"` for the still-present entry.

## Impact

A node using the memory-pool transport under `restart_policy: Always` fails on its
first registration after any crash, for the remaining life of the dataflow. Since
`register_memory_pool` raises, the restarted node typically dies immediately — a
crash-restart loop that never recovers.

## Why #2881 does not cover it

After #2881, a pool is reclaimed on node exit once no live node can still reach it,
which frees both the table entry and the `/dev/shm` segment — so the restart succeeds
whenever the crashed node's consumers are gone too. The collision remains in the
common case: sender crashes while its receiver keeps running. The pool is then
deliberately retained (the receiver may still read it), and the restarted sender walks
straight into its own leftover name.

Reclaiming it eagerly is not an option, for the reason established in #2881 — it would
cut off the receiver's in-flight transfer.

## Suggested direction

Make the name unique per incarnation rather than per registration. The cheapest form
that preserves the on-wire id format is to seed `PINNED_COUNTER` with a random `u64`
instead of `0`: the component stays a `u64`, so both existing parsers keep working —
the writer fast path (`buffer_id.rsplit_once('_')` then `parse::<u64>()`,
`apis/python/node/src/lib.rs:2459-2462`) and `try_doradma_read`, which reconstructs
`dora_pool_{dataflow_id}_{node_id}_{counter}` from `pool_{node_id}_{counter}`
(`apis/python/node/src/lib.rs:3481-3512`). Any scheme that adds a *new* component to
the name would need both parsers updated.

An alternative — letting a re-registration by the same node id take over a stale entry
whose registrar has exited — fixes the daemon table but not the `/dev/shm` collision,
which happens first, node-side. It would also weaken the duplicate-registration check
for the Rust API, where the caller supplies `shared_memory_id` directly.

Happy to prepare a PR if the random-seed direction looks right.

phil-opp:
## Summary

Fixes #3015. Now targets the **tensor-pool extension** — `libraries/extensions/tensor-pool/python/src/transport.rs` — after #3152 relocated the pool transport out of `apis/python/node`.

Memory-pool shared-memory names are `dora_pool_{dataflow_id}_{node_id}_{counter}`, where `counter` comes from a **process-local** static (`PINNED_COUNTER`) seeded at `0`. Both `dataflow_id` and `node_id` are stable across a crash-restart, so a restarted node re-derives the *exact* name its previous incarnation used. When the old pool is still live — the sender crashes while its receiver keeps reading, so #2881's reclaim deliberately retains it — `ShmemConf::create()` collides on the leftover `/dev/shm` segment and the daemon rejects the registration as a duplicate. Under `restart_policy: Always` this is a crash-restart loop that never recovers.

## Change

- Seed `PINNED_COUNTER` with a random `u64` per process (via the standard library's OS-seeded `RandomState` — no new dependency) so each incarnation derives a distinct name. The counter stays a plain `u64`, so every name parser keeps working unchanged. `wrapping_add` guards the now-possible overflow.
- Extract `pool_shmem_name` / `parse_pool_counter` as the single source of truth for the on-wire id format, replacing ~7 duplicated inline `format!`/`rsplit_once`+`parse` sites. The round-trip tests exercise those production functions (RED on a format regression), covering underscore-in-node-id and non-numeric-tail cases.
- Document the leak trade-off (with the collision gone, a pathological crash loop whose receiver never frees leaks a pool per restart up to the registry cap, instead of failing fast) in the `PINNED_COUNTER` doc and the extension README's defect table; the complete fix is an owner-death reclaim, tracked as follow-up.

## Testing

- `cargo test -p dora-tensor-pool-python pool_id_tests` — 4 passed
- `cargo clippy -p dora-tensor-pool-python -- -D warnings` — clean


phil-opp:

Reviewed the change that seeds `PINNED_COUNTER` with a random per-process `u64` (via `RandomState`) and switches the increment to `wrapping_add`, so a crash-restarted node no longer re-derives its previous incarnation's still-live `dora_pool_{dataflow}_{node}_{counter}` shmem name. I confirmed the real reader fast path (`try_doradma_read`) parses the counter with `buffer_id.rsplit_once('_')`, so a large random-seeded counter and node_ids containing underscores both round-trip correctly — the on-wire id format is unchanged. No important issues found.

(Minor, non-blocking: `large_counter_round_trips_through_the_name_parser` reimplements the `rsplit_once('_')` parse inline rather than calling the production parser, so it pins the format but wouldn't catch drift in the real parser. Not worth changing.)

---

phil-opp:
The hazards check out — every parser is `parse::<u64>()`, a collision is detected loudly at `ShmemConf::create()` rather than silently corrupting, and fork behavior is no worse than before. One consequence worth noting:

**With the collision gone, nothing reclaims the dead incarnation's pool.** `MemoryPoolManager::cleanup_orphans` runs only at dataflow spawn (`binaries/daemon/src/lib.rs:3805`), and the daemon registry is hard-capped at 512 entries (`lib.rs:4686`). So with `restart_policy: Always`, a node that crashes repeatedly, and a receiver that never calls `free_memory_pool`, each incarnation now registers a *fresh* id and leaves the old entry plus its `/dev/shm` segment (data region up to 1 GiB) behind. After 512 registrations every pool registration on that daemon fails, including unrelated dataflows. Pre-fix this loop failed fast on the first restart. An owner-death reclaim, or at least a note in the issue, would cover it.

Also: neither new test can go RED (one exercises a function this PR introduces, the other re-implements the parser inline instead of calling it), and `dora-node-api-python` is excluded from both PR CI and nightly `cargo test` — so nothing in CI exercises this change at all.

phil-opp:
Thanks — both points are fair. Pushed `0029b8e` addressing them, and left the reclaim itself for you to weigh.

**Pool leak with the collision gone.** You're right that this converts fail-fast into fail-after-N: a repeatedly-crashing node whose receiver never frees now leaks a registry entry + `/dev/shm` segment per restart, and `cleanup_orphans` only runs at spawn against a capped registry. I've documented the trade-off in the `PINNED_COUNTER` doc comment and added a note on #3015, and I've **not** folded an owner-death reclaim into this PR — that touches `MemoryPoolManager` lifecycle and wants its own integration coverage, so it's better as a separate change you can scope. My read is that this is still net-positive because it fixes the *common* case (a transient crash under `restart_policy: Always` now recovers instead of crash-looping on the first restart), while the regressed case is a node that is already crash-looping — but if you'd rather hold this until the reclaim lands, that's reasonable too.

**Tests.** Agreed the originals were tautological. I extracted `pool_shmem_name` / `parse_pool_counter` as the single source of truth for the on-wire id format (also dedups ~7 inline copies), and the round-trip test now exercises those production functions, so a format regression in either turns it RED — including the underscore-in-node-id and non-numeric-tail cases. The structural gap you flagged is real and I can't close it here: `dora-node-api-python` is excluded from both PR and nightly `cargo test`, so none of this runs in CI. If it'd help, I'm happy to open a follow-up to add a minimal `cargo test -p dora-node-api-python` job (it builds against the system Python, no maturin needed) so at least these pure-logic tests get exercised.

---

phil-opp:

Reviewed the latest commit `0029b8e`. The extraction of `pool_shmem_name` / `parse_pool_counter` is a faithful refactor of the ~7 inline `format!` / `rsplit_once('_')+parse` sites — the last-segment-as-counter logic and every consumer (`try_doradma_read`, the fast/slow write paths) are preserved, so the on-wire id format is unchanged. The round-trip test now exercises the real production functions (would go RED on format drift, including the underscore-in-node-id and non-numeric-tail cases), which closes the tautological-test gap noted earlier. The leak trade-off is now documented rather than fixed; that remains a deliberate follow-up as discussed. No new issues found.

---

phil-opp:
Removed from the 1.0 milestone: the memory-pool transport is being parked out of the tree in https://github.com/dora-rs/dora/pull/3152, so #3015 no longer blocks 1.0.

This fix is not in the parked copy — #3015 is recorded in `external/dora-pool/README.md` as a re-entry criterion. Worth rebasing onto the external package to carry it forward.

phil-opp:
Makes sense — happy to carry this into the parked copy. I checked #3152: `external/dora-pool/python-binding/node_binding.rs` is a verbatim copy with `PINNED_COUNTER` still seeded at `0` (line 33) and the same duplicated `dora_pool_{}_{}_{}` / `rsplit_once('_')` sites, so the port is mechanical — the same random seed + `wrapping_add`, plus the `pool_shmem_name`/`parse_pool_counter` extraction and round-trip tests applied there instead.

One sequencing question, since `external/dora-pool` only exists on `pool-extract` today: do you want me to

1. **stack it on `pool-extract` now** (retarget #3056's base to `pool-extract`, re-apply the fix to `node_binding.rs`) so it merges together with / right after #3152, or
2. **wait for #3152 to land on `main`**, then port and retarget #3056 to `main`?

I'd lean toward (2) to avoid basing on a branch that's still moving, and I'll keep #3056 as the in-tree fix in the meantime in case #3152 doesn't land — but I'll do whichever you prefer. Either way I'll flip the README re-entry row from "PR #3056 open" to the fix being present in the parked copy.

---

phil-opp:
**Correction to my earlier note on this issue.** The memory-pool transport is *not* being moved out of the repository after all.

It is now an opt-in extension at `libraries/extensions/memory-pool`, outside the 1.0 guarantees (https://github.com/dora-rs/dora/pull/3152), so #3015 does not block 1.0 — but the fix is still wanted. The transport code moved wholesale; this should rebase onto `libraries/extensions/memory-pool/python/src/transport.rs` with little change.

phil-opp:
Rebased onto `main` now that #3152 has landed. The fix moved to the relocated home — `libraries/extensions/tensor-pool/python/src/transport.rs` (`PINNED_COUNTER`) — with the same random seed + `wrapping_add`, the `pool_shmem_name`/`parse_pool_counter` dedup, and the round-trip tests. I also flipped the `tensor-pool/README.md` "Known open defects" row for #3015 from open to fixed, with the leak trade-off recorded as the remaining follow-up.

Note `dora-tensor-pool-python` is a normal workspace member (not in the CI python-exclude list), so unlike the old `apis/python/node` location these tests and clippy now actually run under `cargo test --all` / `cargo clippy --all` in PR CI. Green locally: `cargo test -p dora-tensor-pool-python pool_id_tests` (4 passed), `cargo clippy -p dora-tensor-pool-python -- -D warnings`, `cargo fmt --all -- --check`.

---

phil-opp:
Re-reviewed the Aug 18 rebase (`f4372d76`) that moves this fix onto the relocated `libraries/extensions/tensor-pool/python/src/transport.rs` now that #3152 has landed. The port is faithful: `PINNED_COUNTER` is still seeded via the OS-seeded `RandomState` with `wrapping_add` guarding overflow, and every previously-inline `format!("dora_pool_{}_{}_{}")` / `rsplit_once('_')+parse` site now routes through the extracted `pool_shmem_name` / `parse_pool_counter` pair, so the on-wire id format is unchanged and `try_doradma_read` plus the fast/slow write paths keep parsing the counter as the last `_`-segment (correct for node ids that themselves contain `_`).

The two points from the earlier reviews remain addressed: the pool-leak trade-off (a crash-looping node whose receiver never frees leaks one registry entry + `/dev/shm` segment per restart, up to the daemon cap) is documented in both the `PINNED_COUNTER` doc and the README, and the round-trip tests exercise the production functions (underscore-in-node-id and non-numeric-tail cases included) rather than reimplementing the parse. Also worth noting: the relocated crate `dora-tensor-pool-python` is not in the CI python-exclude list, so unlike the old `apis/python/node` home these `pool_id_tests` now actually run under `cargo test --all` in PR CI — closing the "nothing in CI exercises this" gap raised earlier.

No new issues found in this diff.

---

---

phil-opp:
No issues found. Re-checked after the Aug 18 merge commit (`7aac346`) that resolved the `transport.rs` conflict, since it post-dates the last automated review of `f4372d7`. The conflict resolution is clean and the fix is intact: `PINNED_COUNTER` is still seeded via the OS-seeded `RandomState` (with `wrapping_add` guarding the now-possible overflow), and every pool-name / counter site still routes through the extracted `pool_shmem_name` / `parse_pool_counter` pair, so the on-wire id format is unchanged. No new findings relative to the earlier reviews.

---
_🤖 Fully automated review by Claude — no human has verified these findings. Reply if anything looks off._

---

phil-opp:
Merged `main` (clean). All 11 prior comments reconcile — the round-trip test now exercises the production `pool_shmem_name`/`parse_pool_counter` pair, the leak trade-off is documented rather than fixed (as agreed), and the relocation to `dora-tensor-pool-python` means these tests actually run under `cargo test --all`, which was the point. 44 tests pass, clippy clean.

Two things worth recording about the fix's shape, since they came up while checking it:

- **It fails closed.** `shared_memory_extended` opens with `O_CREAT | O_EXCL` and maps `EEXIST` to `MappingIdExists`, and the single call site bails on it. So a residual random collision degrades to today's loud failure, never to a silent map into a live foreign segment.
- **`wrapping_add` matters.** Without it a seed landing on `u64::MAX` would panic in debug builds.

I pushed three doc corrections, because the fix invalidated claims the surrounding comments were making:

- `owned_pool_counter`'s rationale said "every process's counters restart at 1" — this PR is precisely what makes that false. The #3168 guard is still needed, now as defense against an improbable rather than a certain collision. Left as-is a future reader checks the premise, finds it untrue, and concludes the guard is removable.
- The `PINNED_COUNTER` doc credited #2881's reclaim with deliberately keeping the stale segment. `TensorPoolManager::reclaim_unreachable` has no production caller — only tests; the daemon wires sweeps at daemon start, dataflow spawn and dataflow finish only. So a crash-restart inside a live dataflow unlinks nothing, and the bug this fixes is *broader* than the comment described, not narrower.
- `pool_shmem_name` claims to be the single source of truth for the format; it is for this crate, but the daemon half derives the same names in `cross_pool_shmem_name`/`cleanup_orphans` — which also disagree on the empty-machine-id edge (`dora_pool__{df}_…` vs `dora_pool_{df}_…`). That asymmetry is pre-existing and out of scope here, but the claim shouldn't overreach.

Also noted in the test's own comment that `random_seed_is_not_constant` is a weaker guard than it looks: `RandomState::new()` bumps its cached key on every call within a thread, so successive draws differ even if the OS seeding were removed. It only goes RED against a literal-constant implementation; a real cross-process check needs a spawned subprocess.

Ready to merge from my side.

phil-opp:
Thanks for the thorough pass — reviewed your three doc corrections and they're all right, nothing further from my side.

The `reclaim_unreachable`-has-no-production-caller point is the one that matters most: I'd written the `PINNED_COUNTER` doc as if #2881's reclaim were deliberately *retaining* the live segment, when in fact nothing unlinks it on a crash-restart inside a running dataflow — so the collision was certain, not "retained on purpose," and the bug is broader than I described. Your rewording is the accurate framing. Agreed too that `pool_shmem_name` is only the single source of truth for this crate's half (the daemon derives the same names separately, incl. the pre-existing empty-machine-id asymmetry), and that `random_seed_is_not_constant` only catches a literal-constant regression. Good to merge.

---

PR #3059, opened 2026-08-06
docs(message): correct LogLevelOrStdout::passes filter semantics

phil-opp:
## Summary

The doc comment on the public `LogLevelOrStdout::passes` (`libraries/message/src/common.rs`) described the wrong behavior for the `Stdout` case.

It claimed a single total order `Stdout < Error < Warn < Info < Debug < Trace` and that "a message passes if its level is at or above (`<=`) the minimum". Taken literally that makes `Stdout` the smallest element, so a `Stdout` message would pass **every** filter. The implementation does the opposite:

```rust
(LogLevelOrStdout::Stdout, LogLevelOrStdout::Stdout) => true,
(LogLevelOrStdout::Stdout, _) => false,   // Stdout does NOT pass a severity filter
```

and the existing `stdout_does_not_pass_log_level_filters` test pins exactly that. A reader relying on the comment would get the `Stdout` case backwards.

## Fix

Reword the doc comment to describe the actual rule: `Stdout` is a distinct channel, not a severity.

- A `Stdout` message passes only a `Stdout` filter.
- A `LogLevel` message always passes a `Stdout` filter (the most permissive), and passes a `LogLevel` filter when it is at least as severe.

Documentation only — no behavior change.

## Validation

- `cargo test -p dora-message` (existing tests already encode the described behavior)
- `cargo clippy -p dora-message -- -D warnings`
- `cargo fmt -p dora-message -- --check`

---

⚠️ **Machine-generated change.** This PR was authored by Claude (Claude Code), an AI assistant, as part of an automated code-review pass. It has been reviewed against the existing tests and compiled locally, but please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RoWA8SwvteRHxQ7npJsCyi)_

### DIFF
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -93,8 +93,15 @@ pub enum LogLevelOrStdout {
 impl LogLevelOrStdout {
     /// Returns true if a message at this level passes the given minimum level filter.
     ///
-    /// Ordering: Stdout < Error < Warn < Info < Debug < Trace.
-    /// A message passes if its level is "at or above" (i.e. <=) the minimum.
+    /// `Stdout` is treated as a distinct channel rather than a severity, so it
+    /// does not participate in the `Error < Warn < Info < Debug < Trace`
+    /// ordering (where `Error` is the most severe and `Trace` the least):
+    ///
+    /// - A `Stdout` message passes only a `Stdout` filter; it is never
+    ///   delivered to a severity filter.
+    /// - A `LogLevel` message always passes a `Stdout` filter (the most
+    ///   permissive), and passes a `LogLevel` filter `min` when its severity is
+    ///   at least as severe as `min` (`msg <= min` in log-crate ordering).
     pub fn passes(&self, min: &LogLevelOrStdout) -> bool {
         match (self, min) {
             (LogLevelOrStdout::Stdout, LogLevelOrStdout::Stdout) => true,


phil-opp:

This is the summary comment for an automated review pass (machine-generated by Claude / Claude Code). Six PRs were opened off freshly-fetched `origin/main`, each independently verified (fmt + clippy + affected-crate tests). The combined diff was additionally run through the full workspace test suite and the `/review` + `/simplify` gates once.

### PRs opened, by category

**Correctness**
- **#3063** — `fix(coordinator)`: tolerate transient log-subscriber send timeouts. A single 100 ms send timeout permanently dropped a `dora logs -f` subscriber; now mirrors the topic path's consecutive-timeout streak (100 in a row before eviction). The `/review` gate caught a follow-on bug in the first draft — a level-filtered message counted as a successful send and reset the streak, so below-filter traffic could keep a stuck subscriber alive forever — now fixed and regression-tested. `/simplify` prompted extracting the shared `TimeoutStreak` counter used by both subscriber types.
- **#3061** — `fix(cli)`: trim whitespace around `dora daemon --labels` keys/values, so `--labels "gpu = true"` no longer produces an unschedulable `"gpu "` label.
- **#3064** — `fix(ros2-bridge)`: validate transient-local subscription history depth. The publisher rejected a non-positive `KeepLast` depth but the subscriber cast it straight to `usize` (a negative `i32` wrapped to a near-`usize::MAX` sample count); both paths now share one validated helper.

**Documentation**
- **#3059** — `docs(message)`: correct `LogLevelOrStdout::passes` — the comment described a single `<=` total order that contradicted the actual `Stdout`-is-a-separate-channel behavior (which the tests pin).
- **#3062** — `docs(cli)`: split the merged `--env` / `--exit-when-nodes-finish` help on `dora start` (the `--env` help had run into `exit_when_nodes_finish`'s doc; `--env` had none).

**Simplification**
- **#3060** — `refactor(core)`: share one lexical `normalize_path` between `descriptor::expand` and `manifest::inject` (two byte-identical copies of a security-sensitive path-escape normalizer).

### Notable findings considered but **not** turned into PRs (this run)

- `message`: `InputMapping::Display` silently drops `node_filter` when `min_level` is `None` — latent, only reachable by constructing an invalid filter programmatically; needs a structural change to be worth it.
- `message`: `validate::parse_byte_size` duplicates `ByteSize::from_str` — real duplication, but the two intentionally return different integer widths (`u64` vs `usize`); a safe merge is a larger cross-crate change.
- `cli`: raw legacy `.txt` local logs bypass `--since/--until/--grep/--tail` and timestamp ordering in `dora logs` — genuine, but the fix changes output behavior and carries regression risk beyond this pass.
- `apis/c++` node/operator: dead `_key` param in `parameter_to_json`; a catch-all comment naming `Event` variants that don't exist; undocumented silent-`None` in `dora_read_data`; `Metadata::to_json` `.expect` that could unwind across the cxx bridge — all low-value or low-probability C++/cxx-crate nits, skipped to avoid churn disproportionate to impact.
- `coordinator`: `send_log_message` serializes the same `LogMessage` to JSON once per subscriber — a real optimization, but pre-existing (not introduced by these changes), so out of scope here.

### Verification notes

- Per-branch: `cargo fmt --check` + `cargo clippy -- -D warnings` + affected-crate `cargo test` all green.
- Combined diff: full workspace test suite (`--exclude` Python/examples) green, **except** two `apis/rust/node` zenoh tests (`advanced_pub_cache_serves_*`) that fail on `tcp/[::]:0` IPv6 binding in this sandbox — an environment limitation unrelated to these changes (no file in the diff touches that crate; they fail identically on `origin/main`).
- `/review` and `/simplify` both run on the combined diff; their findings are folded into #3063 as described above.

---

phil-opp:
**Automated review by Claude** — fully automated, may contain mistakes; please verify before acting.

No issues found. I checked the rewritten doc comment against the actual `passes` match arms:

- `(Stdout, Stdout) => true` / `(Stdout, _) => false` matches "a `Stdout` message passes only a `Stdout` filter; it is never delivered to a severity filter."
- `(LogLevel(_), Stdout) => true` matches "a `LogLevel` message always passes a `Stdout` filter (the most permissive)."
- `(LogLevel(msg), LogLevel(max)) => msg <= max` matches "passes a `LogLevel` filter `min` when its severity is at least as severe as `min` (`msg <= min` in log-crate ordering)."

The `Error < Warn < Info < Debug < Trace` ordering and "Error most severe / Trace least" note agree with `log::Level`'s ordering, so `msg <= min` correctly means "at least as severe." The new text also matches the behavior pinned by the existing tests (`stdout_fails_non_stdout_filters`, `any_log_level_passes_stdout_filter`, `error_passes_less_verbose_filters`, `debug_fails_info_filter`). Documentation-only; no behavior change.

---

PR #3061, opened 2026-08-06
fix(cli): trim whitespace around `dora daemon --labels` keys and values

phil-opp:
## Summary

`parse_labels` (in `binaries/cli/src/command/daemon.rs`) trimmed each comma-separated pair as a whole, but not the two halves of `key=value`:

```rust
let pair = pair.trim();
let (k, v) = pair.split_once('=')?;
map.insert(k.to_string(), v.to_string());   // "gpu " => " true"
```

So `--labels "gpu = true"` produces key `"gpu "` and value `" true"`. Those never match a node's `gpu: true` deploy requirement, so the daemon silently advertises a label that cannot be scheduled against — with no error surfaced to the user. The sibling `parse_log_filter` in `output.rs` already trims both `node` and `level`; this is the one label path where surrounding whitespace leaks in.

## Fix

- Trim both halves after the split.
- Reject an empty key so a stray `=value` is a clear error rather than a blank label.
- Add unit tests covering trimming, empty-pair skipping, and the missing-value / empty-key error cases.

## Validation

- `cargo test -p dora-cli --lib parse_labels` — 3 new tests pass
- `cargo clippy -p dora-cli --lib -- -D warnings`
- `cargo fmt -p dora-cli -- --check`

---

⚠️ **Machine-generated change.** This PR was authored by Claude (Claude Code), an AI assistant, as part of an automated code-review pass. It compiles and passes the new tests locally, but please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RoWA8SwvteRHxQ7npJsCyi)_

### DIFF
--- a/binaries/cli/src/command/daemon.rs
+++ b/binaries/cli/src/command/daemon.rs
@@ -303,7 +303,40 @@ fn parse_labels(s: &str) -> Result<BTreeMap<String, String>, String> {
         let (k, v) = pair
             .split_once('=')
             .ok_or_else(|| format!("invalid label `{pair}`, expected key=value"))?;
-        map.insert(k.to_string(), v.to_string());
+        // Trim each half, not just the whole pair: `--labels "gpu = true"`
+        // must yield key `gpu` / value `true`, otherwise the surrounding
+        // whitespace leaks into the label and it silently fails to match a
+        // node's `gpu: true` requirement at scheduling time.
+        let k = k.trim();
+        if k.is_empty() {
+            return Err(format!("invalid label `{pair}`, key must not be empty"));
+        }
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the change that trims both halves of each `key=value` label and rejects an empty key. The whole-pair `trim()` + empty-pair skip already existed, so `--labels "gpu = true"` now correctly yields key `gpu` / value `true`, and a stray `=value` becomes a clear usage error instead of a blank label. The three unit tests each assert meaningful behavior (trimming, empty-pair skipping, and both error paths). Empty *values* remain allowed (e.g. `gpu=`), which matches the prior behavior. No issues found.

---

PR #3063, opened 2026-08-06
fix(coordinator): tolerate transient log-subscriber send timeouts

phil-opp:
## Summary

`send_log_message` (in `binaries/coordinator/src/handlers.rs`) closed a log subscriber on a **single** 100 ms send timeout:

```rust
let send_result = tokio::time::timeout(Duration::from_millis(100), subscriber.send_message(message));
if send_result.await.is_err() {
    subscriber.close();   // one hiccup ends the subscription
}
```

`send_message` writes to a bounded `mpsc` channel feeding a `dora logs -f` CLI. Whenever that reader stalls long enough to keep the channel full for 100 ms during a log burst, this permanently closes the subscriber and silently ends the log stream.

The sibling `send_topic_frames` in the *same file* was deliberately hardened against exactly this failure mode: it keeps a per-subscriber consecutive-timeout counter, resets it on any successful send, and only closes after `100` **consecutive** timeouts. `LogSubscriber` had no such tolerance, so the two structurally-parallel "stream to a CLI" paths degraded differently.

## Fix

- Give `LogSubscriber` the same tolerance and rewrite `send_log_message` to mirror the topic path: reset the streak on success, close on an actually-closed channel, and close on a timeout only after 100 consecutive ones.
- `send_message` now reports whether the message was **actually enqueued** vs **dropped by the level filter**. A filtered message is not a delivery attempt, so it leaves the streak untouched — otherwise a stuck subscriber that keeps seeing below-filter traffic (e.g. a Warn-level reader during Info/Debug chatter) would reset the streak on every filtered message and never be evicted.
- Both subscribers now **share the streak counter**: a small `TimeoutStreak` (`record`/`reset`) is extracted into its own module and embedded in `LogSubscriber` and `TopicSubscriber`, so the counter mechanism has one definition instead of two byte-identical copies. The send loops themselves stay separate (the log path's level filter makes them genuinely different).

## Validation

- `cargo test -p dora-coordinator` — all pass, including:
  - `timeout_streak::tests::record_is_monotonic_and_reset_returns_to_zero`
  - `handlers::tests::send_log_message_resets_timeout_streak_on_successful_send`
  - `handlers::tests::send_log_message_closes_subscriber_after_100_consecutive_timeouts` (uses `start_paused` to auto-advance the 100 timeouts)
  - `handlers::tests::send_log_message_filtered_message_does_not_reset_streak`
  - the existing `send_topic_frames_*` tests still pass unchanged
- `cargo clippy -p dora-coordinator --all-targets -- -D warnings`
- `cargo fmt -p dora-coordinator -- --check`

---

⚠️ **Machine-generated change.** This PR was authored by Claude (Claude Code), an AI assistant, as part of an automated code-review pass. It compiles and passes the crate's test suite locally, but please review carefully before merging.

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -71,12 +71,43 @@ pub(crate) async fn send_log_message(
     log_subscribers: &mut Vec<LogSubscriber>,
     message: &LogMessage,
 ) {
+    // Tolerate transient slowness the same way `send_topic_frames` does: a
+    // single 100 ms timeout — which happens whenever a `dora logs -f` reader
+    // stalls long enough to keep its bounded channel full during a log burst —
+    // must not permanently drop the subscription. Only close after the channel
+    // is actually gone, or after a sustained streak of timeouts.
+    const MAX_CONSECUTIVE_LOG_SEND_TIMEOUTS: usize = 100;
     for subscriber in log_subscribers.iter_mut() {
         let send_result =
-            tokio::time::timeout(Duration::from_millis(100), subscriber.send_message(message));
-
-        if send_result.await.is_err() {
-            subscriber.close();
+            tokio::time::timeout(Duration::from_millis(100), subscriber.send_message(message))
+                .await;
+        match send_result {
+            // Actually enqueued — the subscriber is keeping up, clear the streak.
+            Ok(Ok(true)) => {
+                subscriber.reset_timeout_streak();
+            }
+            // Dropped by the level filter — not a delivery attempt, so leave the
+            // streak untouched (resetting here would let below-filter traffic
+            // keep a stuck subscriber alive forever).
+            Ok(Ok(false)) => {}
+            Ok(Err(_)) => {
+                subscriber.close();
+            }
+            Err(_) => {
+                let timeouts = subscriber.record_timeout();
+                if timeouts >= MAX_CONSECUTIVE_LOG_SEND_TIMEOUTS {
+                    tracing::warn!(
+                        timeouts,
+                        "closing slow log subscriber after repeated send timeouts"
+                    );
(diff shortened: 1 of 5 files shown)


phil-opp:

No issues found. The change correctly brings the log-subscriber path to parity with `send_topic_frames`: reset on a real enqueue (`Ok(Ok(true))`), close on an actually-closed channel (`Ok(Err(_))`), and close on timeout only after 100 consecutive occurrences. Making `send_message` return `Ok(false)` for level-filtered drops — and leaving the streak untouched in that arm — is the right call, since resetting on below-filter traffic would let a stuck subscriber evade eviction. Extracting the shared `TimeoutStreak` de-duplicates the counter without changing either path's semantics. The three integration tests assert the meaningful behaviors (reset-on-success, eviction on the 100th consecutive timeout, filtered-message-does-not-reset), not just happy paths.

---

phil-opp:
🤖 **Automated review by Claude** — this review is fully automated, with no human in the loop. Treat it as advisory.

Re-checked after the branch was rebased/force-pushed onto `main` since the previous automated review. The change is unchanged in substance and still looks correct: `send_log_message` mirrors `send_topic_frames` — reset the streak only on a real enqueue (`Ok(Ok(true))`), leave it untouched on a level-filtered drop (`Ok(Ok(false))`) so below-filter traffic can't keep a stuck subscriber alive, close immediately on an actually-closed channel (`Ok(Err(_))`), and evict on timeout only after 100 consecutive ones. The extracted `TimeoutStreak` is a faithful de-dup of the two previously-identical counters, and the three integration tests exercise the meaningful arms (reset-on-success, eviction on the 100th consecutive timeout via `start_paused`, filtered-message-does-not-reset).

No new issues found; latest commits look safe.

---

PR #3068, opened 2026-08-07
fix(ros2-bridge): accept Binary/LargeBinary for fixed-size byte arrays

### LINKED ISSUE #3065: ros2-bridge: fixed-size array field (`type[N]`) rejects Arrow `Binary`/`LargeBinary` columns, unlike the sequence path
> **Note:** This issue was created by an automated, scheduled Claude code-review check. It has been verified by reading the code but has not been reproduced against a live ROS 2 peer. Please treat the severity assessment accordingly.

## Summary

The serializer for **fixed-size array** message fields (`type[N]`, e.g. `uint8[16] uuid`) only accepts an Arrow `List`/`LargeList` column. The **variable-length sequence** serializer, by contrast, was deliberately extended in #2507 / #2816 to also accept an Arrow `Binary`/`LargeBinary` column — the representation a `pyarrow` producer naturally uses for a byte array. As a result, a byte-valued fixed-size array supplied as a `Binary` column fails serialization with `value is not compatible with expected array type`, while the identical value in a `sequence<uint8>`/`uint8[]` field serializes fine.

This fails **safe** (a clean error, not silent wire corruption), so it is a functional/consistency gap rather than a data-integrity bug.

## Location

`libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs`, lines 29–49:

```rust
let entry = if let Some(list) = self.column.as_list_opt::<i32>() {
    list.value(0)
} else {
    let list = self
        .column
        .as_list_opt::<i64>()
        .ok_or_else(|| error("value is not compatible with expected array type"))?;
    list.value(0)
};
```

There is no `as_binary_opt::<i32>()` / `as_binary_opt::<i64>()` branch here.

Compare the sequence path, `libraries/extensions/ros2-bridge/arrow/src/serialize/sequence.rs`, lines 35–78, which handles `List`, `LargeList`, `Binary`, **and** `LargeBinary` (with `seq_len` read from the row's byte length), plus the `ByteSequence` element serializer at lines 269–290.

## Concrete scenario

A common ROS 2 message, `unique_identifier_msgs/msg/UUID`:

```
uint8[16] uuid
```

If a Python/`pyarrow` node produces the `uuid` field as an Arrow `Binary` (or `FixedSizeBinary`) column — the idiomatic representation for a byte buffer — serialization fails with:

```
value is not compatible with expected array type
```

The same 16 bytes supplied as a `uint8[]` (variable sequence) field serialize correctly, because the sequence path accepts `Binary`. The dora-internal round trip is unaffected because the deserializer always emits `List<UInt8>`; the gap only affects external Arrow producers that use `Binary` for byte fields.

## Suggested fix

Mirror the sequence path in `ArraySerializeWrapper::serialize`: add `as_binary_opt::<i32>()` / `as_binary_opt::<i64>()` branches for byte-typed fixed-size arrays (`BasicType::U8 | Char | Byte`), slicing the single row and feeding its bytes to the `UInt8` tuple serializer, and keep the existing `check_array_len` (`== N`) enforcement so a wrong-length buffer still errors. A regression test analogous to the `#[test] fn fixed_string_array_length_is_checked` cases (but for `uint8[N]` supplied as `Binary`) would pin the behavior.

phil-opp:
## Summary

Fixes #3065.

The serializer for **fixed-size array** message fields (`type[N]`, e.g. `uint8[16] uuid`) only accepted an Arrow `List`/`LargeList` column. The **variable-length sequence** serializer, by contrast, was extended in #2507 / #2816 to also accept an Arrow `Binary`/`LargeBinary` column — the representation a `pyarrow` producer naturally uses for a byte buffer. As a result, a byte-valued fixed-size array supplied as a `Binary` column failed serialization with `value is not compatible with expected array type`, while the identical value in a `sequence`/`uint8[]` field serialized fine.

This affected common messages such as `unique_identifier_msgs/msg/UUID` (`uint8[16] uuid`) when the field is produced as an Arrow `Binary` column.

## Changes

`libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs`:

- `ArraySerializeWrapper::serialize` now handles a single-row `Binary`/`LargeBinary` column before the generic `List` extraction, mirroring the sequence path.
- New `serialize_binary_fixed_array` helper serializes the row's bytes directly into the fixed-size tuple, keeping the existing fixed-size length check (`check_array_len`, #2027) — now applied to the byte count.
- A `Binary` column supplied for a **non-byte** element type is rejected as a genuine type mismatch, rather than silently mis-serialized.

## Behavior

- `uint8[N]` / `char[N]` / `byte[N]` supplied as `Binary`/`LargeBinary` now serializes, and encodes byte-for-byte identically to the same value supplied as `List<UInt8>`.
- A wrong-length `Binary` buffer still errors (fixed arrays have no CDR length prefix, so length must match exactly).
- The dora-internal round trip is unaffected — the deserializer still emits `List`; this only closes the gap for external Arrow producers that use `Binary` for byte fields.

## Tests

Added regression tests in the same file:

- `fixed_byte_array_binary_matches_list` — a `uint8[3]` field as `Binary` encodes identically to `List<UInt8>`, with a trailing `int32` field decoded via a real CDR reader to verify wire alignment.
- `fixed_byte_array_binary_length_is_checked` — a wrong-length byte buffer errors; a correct-length one serializes.

All 21 tests in the crate pass; `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_013CswsuhodunaWKhA77nuXa)_

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs
@@ -26,6 +26,30 @@ impl serde::Serialize for ArraySerializeWrapper<'_> {
     where
         S: serde::Serializer,
     {
+        // A byte-typed fixed array (`uint8[N]`, `char[N]`, `byte[N]`) may be
+        // supplied as an Arrow `Binary`/`LargeBinary` column — the idiomatic
+        // representation a `pyarrow` producer uses for a byte buffer. The
+        // variable-length sequence path already accepts this (#2507/#2816); the
+        // fixed-array path must be consistent, otherwise the same 16 bytes that
+        // serialize fine as `uint8[]` fail as `uint8[16]`. The row's *bytes* are
+        // the array elements, so handle it here before the generic `List`
+        // extraction below.
+        if let Some(binary) = self.column.as_binary_opt::<i32>() {
+            return serialize_binary_fixed_array(
+                serializer,
+                binary,
+                &self.array_info.value_type,
+                self.array_info.size,
+            );
+        } else if let Some(binary) = self.column.as_binary_opt::<i64>() {
+            return serialize_binary_fixed_array(
+                serializer,
+                binary,
+                &self.array_info.value_type,
+                self.array_info.size,
+            );
+        }
+
         let entry = if let Some(list) = self.column.as_list_opt::<i32>() {
             if list.len() != 1 {
                 return Err(error(format!(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Accepting a single-row `Binary`/`LargeBinary` column for byte-typed fixed arrays mirrors the sequence path correctly:

- The `value_type` guard restricts the Binary shortcut to `U8 | Char | Byte` and rejects any other element type as a genuine mismatch, so a non-byte fixed array can't silently take this path.
- `check_array_len` is still applied to the byte count, and the `binary.len() != 1` guard rejects multi-row columns.
- The tests are non-vacuous: `fixed_byte_array_binary_matches_list` asserts byte-for-byte equality against the `List<UInt8>` encoding *and* round-trips through a real CDR reader to confirm the trailing `int32` stays aligned (a fixed array has no length prefix, so this is the meaningful check); `fixed_byte_array_binary_length_is_checked` asserts the wrong-length buffer errors.

---

PR #3070, opened 2026-08-07
feat(core)：add node field whitelist classifier to prevent silent field drops

SunSunSun689:
### Problem addressed

In Dora’s YAML dataflow descriptor, the `Node` struct has 40+ fields that serve 6 node types (Standard / Custom / Runtime / Operator / ROS2 Bridge / Module）Each node type consumes only a subset of those fields, but previously there was **no validation** — any field that did not belong to the current node type was **silently dropped** (no effect, no error).
As a result:
- Users mistakenly believed the configuration was effective, while the actual dataflow behaviour differed from their expectations.

### Solution
Introduce a centralised **whitelist checker** (`classify.rs`) that runs before node resolution:
- Each of the 6 node types has an explicit whitelist (the set of allowed fields).
- Any field not in the whitelist is reported as an error all at once, with hints pointing to the correct location.
- Node type classification (`NodeClass`) and field checking are performed together, replacing the previously scattered validation logic.
- The check is read-only (takes `&Node`, not `&mut Node`).

### Architecture change
```
Before: YAML → kind() → node_kind_mut → validate_* (scattered) → resolution
 After: YAML → classify() → NodeClass → resolution
                  ↑
         Type detection + whitelist check + hints for Custom
         Contained in a single file
```

### File changes

| File | Change | Description |
|------|--------|-------------|
| `libraries/core/src/descriptor/classify.rs` | +380 lines | New: whitelist constants, `classify()`, `check_module_fields()` |
| `libraries/core/src/descriptor/mod.rs` | +81/-130 | Wiring: replaces `node_kind_mut` / `NodeKindMut` |
| `libraries/core/src/descriptor/expand.rs` | +5 lines | Calls `check_module_fields()` before module expansion |
| `tests/descriptor-validation.rs` | new | 22 integration test functions |
| `tests/descriptor-validation/cases/` | 22 YAML files | Covers valid + invalid combinations for all 6 types |


phil-opp:

No issues found. The `node_kind_mut` → `classify()` refactor preserves the prior resolution behavior for every kind:

- The per-kind whitelists cover the fields each kind actually consumes during resolution (Standard's list matches every field read into `CustomNode`/`NodeRunConfig`; Operator/Runtime/Custom keep their config inside their own block so the outer-node fields correctly stay unset).
- `standard_source()` reproduces the old git/branch/tag/rev validation verbatim, and the `Module` arm still errors at resolution time exactly as `node_kind_mut` did.
- The `is_set` default-detection (`restart_policy != Never`, `max_restarts != 0`, empty-collection checks) means default-valued fields aren't spuriously flagged.

One intentional behavior change worth calling out (not a bug): fields that were previously silently ignored on a given kind (e.g. `params` on a Standard node) are now hard-rejected. That's the stated purpose of the whitelist, and the per-kind lists look complete, so I don't expect valid dataflows to regress.

---

phil-opp:
**Automated review by Claude — this is a fully automated review with no human in the loop; please verify before acting on it.**

One inconsistency in the per-kind whitelists in `libraries/core/src/descriptor/classify.rs`: `STANDARD_ALLOWED` includes `output_metadata` and `pattern`, but `ROS2_ALLOWED` omits both.

ROS2 bridge nodes carry their `outputs` at the top level, and `validate.rs::check_metadata_annotations` (validate.rs:964) validates top-level `node.output_metadata` / `node.pattern` against `node.outputs` uniformly for every node kind. So a ROS2 bridge node that set top-level `output_metadata:` (or the `pattern:` shorthand) was previously accepted and processed by the validation pass, but `classify()` now hard-rejects it before validation runs, e.g. `fields that are not allowed on ROS2 bridge nodes: output_metadata`.

`output_metadata` in particular looks legitimately usable on a ROS2 node (it just annotates the node's own outputs), so this reads like an accidental omission rather than intent. Worth confirming whether top-level `output_metadata`/`pattern` are meant to be valid on ROS2 bridge nodes — if so, adding them to `ROS2_ALLOWED` would avoid rejecting descriptors that used to work. Adding a test that a ROS2 node with `output_metadata` still resolves would also pin the intended behavior either way.

---

phil-opp:

Two things not raised in the earlier reviews:

**1. The tests described in the PR body are not in this PR.** The "File changes" table claims a new `tests/descriptor-validation.rs` (22 test functions) plus 22 YAML case files, but the actual diff touches only 3 files (`classify.rs`, `expand.rs`, `mod.rs`) — no test file, no fixtures. So the new whitelist logic ships with zero test coverage. Given that this change *hard-rejects* configurations that previously parsed (any stray field is now an error), the behavior it introduces is exactly the kind that needs regression tests: a valid dataflow per kind, and at least one rejected-field case per kind. Please add the tests the description references (or correct the description).

**2. The whitelist is three hand-maintained string lists decoupled from the struct, which re-opens the silent-drop hole it's meant to close.** `is_set()` (`classify.rs:96`), the `all_checkable` array (`classify.rs:135`), and the per-kind `*_ALLOWED` consts each independently enumerate `Node`'s fields as string literals, with no compile-time link to `dora_message::descriptor::Node`. The doc comment tells future authors to update "the appropriate whitelist(s)", but the load-bearing list is actually `all_checkable`: a new `Node` field that isn't added there is **never checked at all** — silently accepted on every kind, i.e. the same silent-drop this PR exists to prevent, just relocated. Nothing guards against that drift (no exhaustive-match, no test). Worth at least a test that every `Node` field is either in `all_checkable` or intentionally shared, so a future field addition fails loudly rather than slipping through.

The ROS2 `output_metadata`/`pattern` omission noted in the prior review still stands (confirmed: `validate.rs:963` runs `check_metadata_annotations` on top-level `node.output_metadata`/`node.pattern` for every kind, and ROS2 nodes carry `outputs` at top level, so `classify()` now rejects a descriptor the validator would have accepted).

---

phil-opp:

The Aug 10 commit addresses the earlier review points well — `output_metadata`/`pattern` are now allowed on ROS2 bridge nodes (with a `valid-ros2-metadata.yml` case), the promised tests landed, and `all_node_fields_are_classified_or_marked_shared` uses the generated schema to guard against a new `Node` field being left out of the lists. A few things still worth a look:

1. **The `is_set()` catch-all reopens a narrow version of the silent-drop hole the schema test was meant to close.** The new test guards membership in `ALL_CHECKABLE_FIELDS`, but detection actually flows through `is_set()` (`classify.rs`), whose final arm is `_ => false`. If a future `Node` field is added to `ALL_CHECKABLE_FIELDS` (so the schema test passes) but nobody adds a matching arm to `is_set()`, that field silently returns `false` → never flagged → accepted on every kind. That is exactly the silent-drop this PR exists to prevent, moved one layer down, and the schema test does not catch it. Consider driving `is_set` from the same source of truth, or adding a test that a value in every checkable field is actually detected as set.

2. **`classify_inner()` duplicates `NodeExt::kind()`.** The 6-way discriminator match in `classify_inner` is a verbatim copy of the one in `kind()`, and the comment even says "This mirrors the match in `NodeExt::kind()`." Two independent copies of the exclusive-field logic now have to be kept in sync. Since `kind()` is still `pub`, could `classify` call `kind()` and switch on `NodeKind` instead of re-deriving the discriminator?

3. **Hub error regression + stale comment.** The resolve path used to go through `node_kind_mut` → `node.kind()`, whose pre-check emits a specific "unresolved `hub:` reference — run `dora build` first" message. The new path calls `classify()` directly and never calls `kind()`, so a hub-only node reaching resolution now falls into `classify_inner`'s first arm and gets the generic "requires a `path`, `custom`, …" error instead. The doc comment in `classify_inner` still says hub validation "lives in `Node::kind()`'s pre-check … not here," which no longer holds on this path. Worth either restoring the hub-specific message or updating the comment.

Minor context, not a blocker: because `Node` already has `#[serde(deny_unknown_fields)]`, genuinely-unknown/typo fields are already rejected at parse time — the real gap this PR closes is narrower (a *known* field placed on the wrong node kind, e.g. top-level `outputs` on a `custom:` node). Might be worth reflecting that in the PR description so the scope is clear.

---

phil-opp:

The latest commit (`3ed3d65`, Aug 11) addresses the three points from the previous review, and the change looks safe:

- `is_set` detection is now a per-field function-pointer table in `ALL_CHECKABLE_FIELDS`, and `every_checkable_field_has_set_detector` requires a matching case for each field — the previous `_ => false` catch-all that could silently swallow a newly-added field is gone.
- `classify_inner` now switches on `node.kind()?` instead of re-deriving the discriminator, so it no longer duplicates `NodeExt::kind()`.
- Routing through `kind()` restores the hub-specific "unresolved `hub:` reference — run `dora build`" error, now pinned by the `hub_only_node_keeps_resolution_error` test.

Re-checked the resolution arms against the deleted `node_kind_mut` machinery: Standard still requires `path`, and `standard_source` reproduces the git/branch/tag/rev validation verbatim, so there's no behavior regression beyond the intended hard-rejection of misplaced fields. The schema-driven tests (`all_node_fields_are_classified_or_marked_shared`) will fail loudly if a future `Node` field is left unclassified. No new issues found.

---

phil-opp:
**Automated code review**

Re-reviewed after the Aug 13 merge-from-main (`bab7e4dd`), which is the only new commit since the previous automated review of `3ed3d65`. It's a plain merge — no functional code changed, only the base moved — so the three points from that review remain addressed:

1. `is_set` detection is a per-field function-pointer table (`ALL_CHECKABLE_FIELDS`) guarded by `every_checkable_field_has_set_detector`; the old `_ => false` catch-all is gone.
2. `classify_inner` switches on `node.kind()?` instead of re-deriving the kind discriminator.
3. The hub-specific "unresolved `hub:` reference — run `dora build`" error is restored on the resolution path and pinned by `hub_only_node_keeps_resolution_error`.

I also confirmed the merge introduced no field drift, which is the failure mode that would reopen the silent-drop hole: `dora_message::descriptor::Node` currently has 41 fields, and `SHARED_FIELDS` (5) + the 5 kind discriminators (`custom`/`operators`/`operator`/`ros2`/`module`) + `ALL_CHECKABLE_FIELDS` (31) = 41, so every field is classified and `all_node_fields_are_classified_or_marked_shared` still holds. The per-kind whitelist assignments and resolution arms match the prior (deleted) `node_kind_mut` behavior, so the only behavior change is the intended hard-rejection of misplaced fields.

---

---

phil-opp:

Re-reviewed after the 2026-08-18 merge-from-main (`f109c06`), the only new commit since the last review. It's non-functional for the classifier: the conflict resolution in `libraries/core/src/descriptor/mod.rs` keeps every `NodeClass` arm and the `ok_or_eyre` guards intact, and no classifier logic changed.

Re-verified the failure mode that would reopen the silent-drop hole — field drift between `Node` and the classifier lists — and it still holds: `Node` has 40 fields = `SHARED_FIELDS` (5) + 4 kind discriminators + `ALL_CHECKABLE_FIELDS` (31), so `all_node_fields_are_classified_or_marked_shared` still guards every field. No new findings.

---

phil-opp:
Merged `main` (clean) and audited the whitelist field-by-field against the serde structs. **The whitelist itself is complete**: `Node` has 40 fields, partitioned exactly into 5 shared + 4 kind discriminators + 31 checkable, with no `rename`/`alias`/`flatten` anywhere on `Node` to complicate it, and `all_node_fields_are_classified_or_marked_shared` pins that against the schemars-generated schema. I also classified every node in all 137 in-repo YAML files and 179 fenced YAML blocks in the docs — zero violations outside the intentional `invalid-*` fixtures.

But I found and fixed one regression the review history missed, because every prior pass checked the whitelists against *hand-written* YAML and never against nodes dora synthesizes itself:

**The classifier rejected a field dora writes.** `prepend_module_build_to_node` set `node.build` unconditionally, so a module file with a top-level `build:` put it on *every* inner node — including runtime/operator/ROS2 nodes, where it was a harmless dead value before and is now a hard error. That breaks the documented module-level-build feature (`docs/modules.md`) across `dora check`, `build`, `start`, `run` and `validate`. Reproduced with a CLI built from this branch on all three inner kinds:

```
node `m.single`  has fields that are not allowed on Operator nodes: build
node `m.runtime` has fields that are not allowed on Runtime nodes: build
node `m.bridge`  has fields that are not allowed on ROS2 bridge nodes: build
```

The exact shape is already in the suite as `expand_module_build_prepended_to_operator_and_custom_builds` — it passes only because it stops at `expand_modules()` and never resolves, and the classifier runs at resolution. The build now lands only on kinds that consume it (`path:`, nested `module:`) plus the operator configs, which is where `build/mod.rs` reads it from for runtime nodes anyway. Added a test that crosses the expand→resolve seam; it's RED without the fix.

Two more:

- **The hint told users to delete their wiring.** It was one hardcoded string — "remove them from the node" — for every kind. Top-level `inputs:`/`outputs:` on an `operator:` node is the flagship mistake this check exists to catch, and the correct fix there is to *move* them into the operator block. Hints are now per-kind.
- **The new integration test target never ran.** `tests/descriptor-validation.rs` is a target of the root `dora-examples` package, which every CI job and QA script excludes and no job invokes explicitly — so the only two tests exercising the real expand+resolve pipeline, i.e. exactly the seam the regression lived in, were dead. Moved to `libraries/core/tests/` where `cargo test --all` picks them up, rather than growing CI.

Also dropped `check_module_fields`, a pure passthrough to `check_module`.

Left for you, not fixed:

- **`build:` is rejected on a top-level `module:` node but honored on a nested one** (`MODULE_ALLOWED` omits it; `expand_module_node` reads `inner_node.build`). Verified: the nested form builds and runs. Rejecting the top-level form is defensible — it was a silent no-op — but refactoring a working nested module up one level now breaks it, with an error saying the field "is not consumed by this node kind", which is only true at that one level.
- **Nested `module:` nodes inside module files are never field-checked** — `check_module` is called only from the top-level loop, not from the recursive `expand_module_node`.
- **`cpu_affinity` on a module node is an error while `deploy` is propagated.** Internally consistent, but a user who sees `deploy:` work will expect `cpu_affinity:` to.
- **The PR description doesn't match the diff**: `classify.rs` is 680 lines not 380, there are 2 integration tests not 22 and 11 fixtures not 22, and the architecture diagram still shows a `Custom` node type that no longer exists. The 2026-08-10 note about `deny_unknown_fields` already covering typos is also still unreflected — the real gap this closes is a *known* field on the wrong node kind, which is narrower and worth stating plainly.

PR #3071, opened 2026-08-07
Fix/reliable inter daemon output close

SunSunSun689:
## Summary

  Fix inter-daemon messages being silently dropped when the Zenoh publish drain channel is full.
  `send_to_remote_receivers` now retries full-channel enqueue attempts briefly and returns an error if the message still cannot be queued, instead of logging a warning and returning `Ok(())`. This prevents control events like  `OutputClosed` from being lost without the caller knowing.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -191,14 +191,6 @@ const METRICS_INTERVAL_SECS: f64 = METRICS_INTERVAL.as_secs_f64();
 /// patterns; a full channel is recorded as publish backpressure and dropped so
 /// the daemon event loop does not stall.
 const ZENOH_PUBLISH_CHANNEL_CAPACITY: usize = 256;
-/// Control-plane queue for lifecycle events such as `OutputClosed`. This queue
-/// is separate from the data-plane publish channel so regular output bursts do
-/// not starve close notifications.
-const ZENOH_CONTROL_CHANNEL_CAPACITY: usize = 1024;
-/// Pending control events admitted by the daemon event loop. A background
-/// forwarder waits for capacity in the control publish queue; the event loop
-/// itself never waits.
-const ZENOH_CONTROL_PENDING_CHANNEL_CAPACITY: usize = 4096;
 /// How long the daemon keeps trying to (re)connect to the coordinator before
 /// giving up and exiting. Bounds the orphan-daemon window when the coordinator
 /// is permanently gone (dora-rs/dora#1996); a reachable coordinator connects
@@ -353,7 +345,6 @@ pub struct Daemon {
     /// processes are gone (#2980).
     pub(crate) pending_destroy: Option<PendingDestroy>,
     pub(crate) zenoh_publish_tx: mpsc::Sender<ZenohOutbound>,
-    pub(crate) zenoh_control_tx: mpsc::Sender<ZenohOutbound>,
     pub(crate) remote_daemon_events_tx:
         Option<flume::Sender<eyre::Result<Timestamped<InterDaemonEvent>>>>,
     pub(crate) logger: DaemonLogger,
@@ -1637,44 +1628,14 @@ impl Daemon {
         // Use a large channel capacity to prevent deadlock
         let (dora_events_tx, dora_events_rx) = mpsc::channel(1000);
 
-        // Zenoh publish drain tasks: offload `.put().await` from the main event
-        // loop. Regular outputs use a best-effort data queue; lifecycle control
-        // events first enter a bounded pending queue, then a background
-        // forwarder waits for capacity in a dedicated control publish queue.
+        // Zenoh publish drain task: offload `.put().await` from the main event
+        // loop. All events for a given output, including `OutputClosed`, share
+        // this FIFO so close cannot pass data that was produced earlier.
         let (zenoh_publish_tx, mut zenoh_publish_rx) =
             mpsc::channel::<ZenohOutbound>(ZENOH_PUBLISH_CHANNEL_CAPACITY);
-        let (zenoh_control_publish_tx, mut zenoh_control_publish_rx) =
-            mpsc::channel::<ZenohOutbound>(ZENOH_CONTROL_CHANNEL_CAPACITY);
-        let (zenoh_control_tx, mut zenoh_control_rx) =
-            mpsc::channel::<ZenohOutbound>(ZENOH_CONTROL_PENDING_CHANNEL_CAPACITY);
-        let _zenoh_control_forwarder_handle = tokio::spawn(async move {
-            while let Some(msg) = zenoh_control_rx.recv().await {
-                if zenoh_control_publish_tx.send(msg).await.is_err() {
-                    tracing::debug!("zenoh control publish queue closed");
-                    break;
-                }
-            }
-            tracing::debug!("zenoh control forwarder task exiting");
-        });
         let _zenoh_drain_handle = tokio::spawn(async move {
-            let mut control_open = true;
-            let mut data_open = true;
-            while control_open || data_open {
-                tokio::select! {
-                    biased;
-                    msg = zenoh_control_publish_rx.recv(), if control_open => {
-                        match msg {
-                            Some(msg) => publish_zenoh_outbound(msg).await,
-                            None => control_open = false,
-                        }
-                    }
-                    msg = zenoh_publish_rx.recv(), if data_open => {
-                        match msg {
-                            Some(msg) => publish_zenoh_outbound(msg).await,
-                            None => data_open = false,
-                        }
-                    }
-                }
+            while let Some(msg) = zenoh_publish_rx.recv().await {
+                publish_zenoh_outbound(msg).await;
             }
             tracing::debug!("zenoh publish drain task exiting");
         });
@@ -1704,7 +1665,6 @@ impl Daemon {
             bind_nodes_to_parent,
             pending_destroy: None,
             zenoh_publish_tx,
-            zenoh_control_tx,
             remote_daemon_events_tx,
             git_manager: Default::default(),
             memory_pool: MemoryPoolManager::new(),
@@ -5144,10 +5104,13 @@ impl Daemon {
             net_messages_sent: dataflow.net_messages_sent.clone(),
             net_publish_failures: dataflow.net_publish_failures.clone(),
         };
-        handle_control_enqueue_result(
-            dataflow,
-            try_enqueue_zenoh_outbound(&self.zenoh_control_tx, outbound),
-        )
+        enqueue_required_zenoh_outbound(
+            &self.zenoh_publish_tx,
+            outbound,
+            dataflow.net_publish_failures.clone(),
+            "inter-daemon output-closed event",
+        );
+        Ok(())
     }
 
     async fn try_send_to_remote_receivers(
@@ -5191,24 +5154,11 @@ impl Daemon {
             net_messages_sent: dataflow.net_messages_sent.clone(),
             net_publish_failures: dataflow.net_publish_failures.clone(),
         };
-        match try_enqueue_zenoh_outbound(&self.zenoh_publish_tx, outbound) {
-            ZenohEnqueueResult::Queued => {}
-            ZenohEnqueueResult::Full => {
-                dataflow
-                    .net_publish_failures
-                    .fetch_add(1, atomic::Ordering::Relaxed);
-                tracing::warn!(
-                    "zenoh publish channel full ({ZENOH_PUBLISH_CHANNEL_CAPACITY}); \
-                     dropping regular inter-daemon output"
-                );
-            }
-            ZenohEnqueueResult::Closed => {
-                dataflow
-                    .net_publish_failures
-                    .fetch_add(1, atomic::Ordering::Relaxed);
-                tracing::error!("zenoh drain task is gone; inter-daemon publish channel closed");
-            }
-        }
+        handle_publish_enqueue_result(
+            dataflow,
+            try_enqueue_zenoh_outbound(&self.zenoh_publish_tx, outbound),
+            "regular inter-daemon output",
+        );
         Ok(())
     }
 
@@ -7333,25 +7283,51 @@ async fn publish_zenoh_outbound(msg: ZenohOutbound) {
         .fetch_add(1, atomic::Ordering::Relaxed);
 }
 
-fn handle_control_enqueue_result(
+fn handle_publish_enqueue_result(
     dataflow: &RunningDataflow,
     result: ZenohEnqueueResult,
-) -> eyre::Result<()> {
+    event_kind: &'static str,
+) {
     match result {
-        ZenohEnqueueResult::Queued => Ok(()),
+        ZenohEnqueueResult::Queued => {}
         ZenohEnqueueResult::Full => {
             dataflow
                 .net_publish_failures
                 .fetch_add(1, atomic::Ordering::Relaxed);
             tracing::warn!(
-                "zenoh control pending queue full ({ZENOH_CONTROL_PENDING_CHANNEL_CAPACITY}); \
-                 dropping inter-daemon control event"
+                "zenoh publish channel full ({ZENOH_PUBLISH_CHANNEL_CAPACITY}); dropping {event_kind}"
             );
-            Ok(())
         }
-        ZenohEnqueueResult::Closed => Err(eyre!(
-            "zenoh control forwarder is gone — inter-daemon control channel closed"
-        )),
+        ZenohEnqueueResult::Closed => {
+            dataflow
+                .net_publish_failures
+                .fetch_add(1, atomic::Ordering::Relaxed);
+            tracing::error!("zenoh drain task is gone; dropping {event_kind}");
+        }
+    }
+}
+
+fn enqueue_required_zenoh_outbound<T: Send + 'static>(
+    tx: &mpsc::Sender<T>,
+    outbound: T,
+    net_publish_failures: Arc<AtomicU64>,
+    event_kind: &'static str,
+) {
+    match tx.try_send(outbound) {
+        Ok(()) => {}
+        Err(mpsc::error::TrySendError::Full(outbound)) => {
+            let tx = tx.clone();
+            tokio::spawn(async move {
+                if tx.send(outbound).await.is_err() {
+                    net_publish_failures.fetch_add(1, atomic::Ordering::Relaxed);
+                    tracing::error!("zenoh drain task is gone; dropping {event_kind}");
+                }
+            });
+        }
+        Err(mpsc::error::TrySendError::Closed(_)) => {
+            net_publish_failures.fetch_add(1, atomic::Ordering::Relaxed);
+            tracing::error!("zenoh drain task is gone; dropping {event_kind}");
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


phil-opp:
**Automated review by Claude** — fully automated, may contain mistakes; please verify before acting.

I found the following issue(s):

**Making `send_to_remote_receivers` fatal can tear down the entire daemon on the regular output path.** By returning `Err` when the channel stays full, the failure now propagates fatally for normal `Output` events:

- `binaries/daemon/src/lib.rs:5046` — `send_out()` calls `send_to_remote_receivers(...).await?`.
- `binaries/daemon/src/lib.rs:4738` — the `SendOut` event arm calls `send_out(...).await.context("failed to send out")?` (unlike the `CloseOutputs` arm at ~4714, which swallows the error into a `reply_sender` and does **not** crash the daemon).
- `binaries/daemon/src/lib.rs:1846` — `handle_node_event(...).await?` runs inside the top-level `while let Some(event) = events.next().await` loop (starts at line 1781), so the error exits the daemon's main event loop, taking down **all** local dataflows.

Net effect: sustained back-pressure to a remote daemon that keeps the 256-slot channel full for more than ~30 ms (3 × 10 ms) now converts a single dropped inter-daemon message into a fatal daemon exit. That is a strictly more severe failure mode than the previous warn-and-drop for regular outputs. Note the PR's motivating case — `OutputClosed` via `send_output_closed_events` — actually goes through the `CloseOutputs` reply path and is *not* fatal; it's the generic `Output` path that is.

**Secondary: the retry sleep blocks the single-threaded event loop.** `enqueue_zenoh_outbound_reliably` awaits `tokio::time::sleep` (up to 3 × 10 ms) inline within `send_out`, which is awaited directly in the main event loop (line 1846). While retrying one back-pressured remote publish, the daemon cannot process any other event (other nodes' outputs, heartbeats, coordinator events) for up to ~30 ms.

Worth deciding whether channel-full for the regular output path should be fatal at all, or whether the retry should avoid stalling the event loop (e.g. handled off the main loop / bounded without an inline sleep).

---

phil-opp:
**Automated review by Claude — this is a fully automated review with no human in the loop; please verify before acting on it.**

One scope note in addition to the earlier review. The stated change here is the inter-daemon publish retry (`enqueue_zenoh_outbound_reliably` in `binaries/daemon/src/lib.rs`), but the diff against `main` also contains the entire ~385-line descriptor field classifier — `libraries/core/src/descriptor/classify.rs` plus the rewrite of `resolve_aliases_and_set_defaults_in_topology` and removal of `node_kind_mut`/`NodeKindMut`.

That refactor is the whole content of #3070 and also appears verbatim in #3072, so the same change is currently duplicated across three PRs and is hard to review independently here. If these are meant to be stacked on #3070, it would help to note the dependency (or rebase once #3070 lands); otherwise the classifier is orthogonal to the Zenoh fix and would be much easier to review on its own. It also changes descriptor acceptance — `validate_against_whitelist` now hard-errors on top-level fields that resolution previously ignored — which is a compatibility change worth landing separately with its own tests and a changelog note.

---

phil-opp:

Thanks for splitting the data-plane path off — routing regular `Output` events through the non-fatal `try_send_to_remote_receivers` resolves the earlier concern about backpressure crashing the daemon on the normal send path.

I checked whether the reliable path (`send_output_closed_events` → `send_to_remote_receivers`, which now returns `Err` on sustained-full) is fatal on node-finish. It does propagate via `?` up through `handle_dora_event`, but it's caught at the `run_inner` call site — the `match result { Err(e) => warn!("daemon disconnected from coordinator… reconnect") }` loop — so per the #2029 design the running nodes are preserved and the daemon reconnects rather than crashing. On `dora run` the path is unreachable (no remote receivers, so the `closed`-outputs loop never calls it). So this is **not** a daemon-crash path. Two smaller things do seem worth a look:

1. On that sustained-full `OutputClosed` case you now get a spurious `"daemon disconnected from coordinator… reconnecting"` cycle where the old code just logged a warning and returned `Ok(())`. The reconnect doesn't itself redeliver the `OutputClosed`, so the "reliable" intent isn't actually achieved on this path — the net effect is a misleading disconnect log plus an unnecessary reconnect. It might be cleaner to treat sustained-full here as non-fatal (log + a failure counter, like the data path) or to drive an explicit redelivery, rather than surfacing it as a coordinator-connection error.

2. (Carried over from the earlier review, still unaddressed) `enqueue_zenoh_outbound_reliably` awaits `tokio::time::sleep` (up to ~3×10 ms) inline on the single-threaded event loop, so during the retries the daemon can't process other node/heartbeat/coordinator events.

Also, as previously noted, this diff still bundles the unrelated ~385-line node-classifier refactor from #3070, which is orthogonal to the close-handling fix.

---

phil-opp:

Reviewed the latest push: the unrelated ~385-line node-classifier refactor (flagged in the prior reviews as bundled from #3070) is now gone, so this PR is scoped to just the Zenoh publish-enqueue change — that resolves the earlier scope note. No new issues in the current diff, and the data-plane split (regular `Output` → non-fatal `try_send_to_remote_receivers`) still looks correct; the added tests are meaningful.

Two earlier points remain outstanding on the reliable path and are unchanged by this push:

1. `enqueue_zenoh_outbound_reliably` still awaits `tokio::time::sleep` (up to ~3×10 ms) inline, so while retrying one back-pressured remote publish the single-threaded daemon event loop can't process other node/heartbeat/coordinator events.
2. On the `send_output_closed_events` → `send_to_remote_receivers` path, sustained-full now returns `Err` and surfaces as a spurious `"daemon disconnected from coordinator … reconnecting"` cycle (non-fatal, but the reconnect doesn't redeliver the `OutputClosed`, so the "reliable" intent isn't actually achieved). A non-fatal log-plus-counter, like the data path, or an explicit redelivery would be cleaner.

---

phil-opp:

The latest commit (`d5265b7`, Aug 11) replaces the earlier sleep-based retry design with a two-channel priority split, and it resolves both previously-outstanding concerns: there's no longer an inline `tokio::time::sleep` retry on the event loop, and a sustained-full control queue is now non-fatal in `handle_control_enqueue_result` (only a *gone* forwarder is fatal), so the spurious "daemon disconnected … reconnecting" cycle on the `OutputClosed` path is gone.

I found one new issue introduced by the two-channel prioritization, though:

**`OutputClosed`-vs-`Output` reordering / data-loss window.** Regular `Output` and `OutputClosed` for the same output are published on the **same** zenoh topic via the **same** cached publisher (`zenoh_daemon_control_topic`, `dataflow.publishers` keyed by `output_id`), but they now travel through **separate** channels — data via `zenoh_publish_tx`, `OutputClosed` via `zenoh_control_tx` → forwarder → control publish queue. The drain task's `tokio::select! { biased; … }` drains all ready control messages before data, so a later-produced `OutputClosed` can `publisher.put()` ahead of earlier-produced `Output` messages still sitting in the data channel. On the receiver, `InterDaemonEvent::OutputClosed` runs `close_input(...)`, so those trailing `Output` events then arrive at an already-closed input and get dropped.

That reorders *close* ahead of *data* on a single logical stream — a data-loss window that didn't exist when both events shared one FIFO channel on `main`. It triggers whenever the data channel merely has a backlog at the moment the producing node finishes, not only when the queue is full. Prioritizing control traffic is reasonable, but `OutputClosed` needs to stay ordered behind the data on its own output — e.g. route the close through the same data channel, or drain any pending data for that publisher before publishing its close.

The added backpressure unit tests (`try_enqueue_zenoh_outbound` Queued/Full/Closed, `data_queue_full_does_not_block_control_enqueue`, `control_pending_enqueue_full_is_recorded_without_error`) are meaningful, but none cover cross-channel ordering, which is where the new risk lives.

---

phil-opp:

Re-reviewed after the latest push (`a20a6c2`, "preserve inter-daemon close ordering"), which is new since the previous automated review.

The prior review's blocking concern — the two-channel priority split letting a later `OutputClosed` be published ahead of earlier `Output` messages on the same stream — is resolved. This commit reverts to a **single FIFO** (`zenoh_publish_tx`) shared by both paths:

- Regular `Output` → `try_send_to_remote_receivers` → `try_enqueue_zenoh_outbound` (`binaries/daemon/src/lib.rs:5116`): non-fatal, drops on Full and records `net_publish_failures`, returns `Ok`. Resolves the original daemon-teardown concern.
- `OutputClosed` → `send_to_remote_receivers` → `enqueue_required_zenoh_outbound` (`binaries/daemon/src/lib.rs:5107`, `:7310`): `try_send`, and on Full spawns a detached task that `send().await`s for capacity instead of sleeping inline or dropping. It never returns `Err` for backpressure, so the earlier spurious "daemon disconnected … reconnecting" cycle on the close path is also gone.

Ordering within a single output holds: both events share one FIFO with one consumer (the drain task), there is no post-close data on that output, and tokio's mpsc hands the next freed permit to the queued `send().await` waiter, so `OutputClosed` cannot pass earlier data for its output. No missed/double close and no deadlock in the enqueue path. The `output_closed_waits_for_publish_capacity_without_dropping` and `required_publish_channel_closed_is_recorded` tests meaningfully exercise the reliable path.

---

phil-opp:

First, the good news: this does **not** repeat the #3072 defect. The diff never touches `open_external_mappings`, the finish-straggler veto, or any node-liveness computation, so no healthy node can be SIGKILLed by it. There is also no event-loop deadlock — `enqueue_required_zenoh_outbound` hands the blocking `send().await` to a `tokio::spawn`.

**1. The stated guarantee is not delivered.** `binaries/daemon/src/lib.rs:7484-7495` — the new `try_send_to_remote_receivers` looks the publisher up in the *same* `dataflow.publishers` map, so the "required" `OutputClosed` and the droppable data messages share one publisher declared `.congestion_control(CongestionControl::Drop)`. Making the in-process channel enqueue reliable only moves the drop one layer down.

Failure scenario: the remote link is saturated, node A finishes, the close event now survives the 256-slot channel and reaches `put()` — and zenoh discards it. `publish_zenoh_outbound` bumps `net_publish_failures` and returns. Downstream node B on daemon 2 never receives `InputClosed` and hangs exactly as before. A real fix needs a separate `CongestionControl::Block` publisher (or a retry/ack) for control events.

**2. None of the eight added tests exercise the fix.** `zenoh_publish_drain_preserves_output_closed_ordering` calls only `try_send`/`try_recv` on a `mpsc::channel::<&'static str>` — it contains no dora code and passes identically on unmodified main. `output_closed_uses_the_same_fifo_as_regular_output` is the same test with a wrapper call, and `regular_output_enqueue_full_is_nonfatal` duplicates `zenoh_publish_enqueue_reports_full_without_waiting` line for line. The rest assert tokio's documented `try_send` semantics.

Concretely: revert the `enqueue_required_zenoh_outbound` call at the `OutputClosed` site to a plain `try_send`, and every test except the two that name the new helper still passes. What needs covering is `close_outputs` → `send_to_remote_receivers` with a saturated channel.

**3. `output_closed_waits_for_publish_capacity_without_dropping` is order-fragile.** It relies on a single `yield_now().await` being enough for the spawned task to take the permit freed by the preceding `try_recv`. That holds on the current-thread runtime today but breaks intermittently under `flavor = "multi_thread"` or if the helper gains an await before its `send`. `publish_rx.recv().await` with a timeout would be deterministic.

**4. A reported metric changes meaning silently.** `handle_publish_enqueue_result` now bumps `net_publish_failures` on channel-full drops, where it previously counted only zenoh `put()` failures. That is surfaced as `NetworkMetrics.publish_failures` (`binaries/daemon/src/lib.rs:2074`).

**5.** `try_send_to_remote_receivers` duplicates ~38 lines of publisher lookup and `ZenohOutbound` construction from `send_to_remote_receivers`. One lookup helper plus two enqueue policies would stop the two copies drifting.

SunSunSun689:
Addressed the review follow-ups on this branch:

- split network metrics so `publish_failures` stays reserved for actual zenoh `put()` failures, and enqueue/channel-full drops are counted separately as `publish_enqueue_failures`
- removed the `yield_now()` timing dependency from the close-path regression test by waiting on the receiver with a timeout
- deduplicated the Zenoh outbound publisher/outbound setup into a shared helper so the close and regular paths can't drift

Verified with:
- `cargo test -p dora-daemon output_closed_waits_for_publish_capacity_without_dropping -- --nocapture`
- `cargo test -p dora-daemon regular_publish_queue_full_is_recorded_without_error -- --nocapture`
- `cargo test -p dora-daemon required_publish_channel_closed_is_recorded -- --nocapture`

phil-opp (commented):

The latest commit (`bbb7096`) addresses the earlier test-determinism, metric-meaning, and duplication points (the `yield_now` test is now a `tokio::time::timeout`, `net_publish_enqueue_failures` is split out from `net_publish_failures`, and the shared `prepare_zenoh_outbound` helper removes the duplication). Two items from the earlier review are still open:

**1. The "reliable close" guarantee isn't actually delivered.** Both the regular and `OutputClosed` paths resolve the same per-output publisher from `dataflow.publishers`, which is declared with `.congestion_control(CongestionControl::Drop)` (`binaries/daemon/src/lib.rs:7623`). Making the in-process mpsc enqueue block-until-capacity (`enqueue_required_zenoh_outbound`) only guarantees the event reaches `publisher.put()`; under a saturated remote link Zenoh's `Drop` congestion control discards the sample at the transport, and `put()` still returns `Ok`, so the drop isn't even counted in `net_publish_failures`. Net effect: the `OutputClosed` can still be silently lost and the downstream node on the other daemon never receives `InputClosed` — the exact hang this PR targets. Reliable close delivery needs a `Block` (or acked/dedicated) publisher for the control/close events, not just a reliable in-process enqueue.

**2. A couple of the added tests don't exercise dora code.** `output_closed_uses_the_same_fifo_as_regular_output` and `zenoh_publish_drain_preserves_output_closed_ordering` only `try_send`/`try_recv` on a bare `mpsc::channel::<&'static str>`, so they pass identically on unmodified `main` and don't test the close-ordering guarantee against the real publish path. `regular_output_enqueue_full_is_nonfatal` also duplicates `zenoh_publish_enqueue_reports_full_without_waiting`. The genuinely useful new tests are `output_closed_waits_for_publish_capacity_without_dropping` (now deterministic) and the two metric-split tests.

---

phil-opp:
**Automated review by Claude** — _fully automated, no human in the loop; please verify before acting._

Re-checked after `bbb7096e` ("split zenoh publish failure metrics"), the one non-merge commit since the previous review.

Resolved from the prior review:
- **(#4) metric semantics** — `publish_failures` is now reserved for actual zenoh `put()` failures, and the new `publish_enqueue_failures` counter (added to `NetworkMetrics` in `libraries/message/src/daemon_to_coordinator.rs`) tracks channel-full/closed drops separately. `handle_publish_enqueue_result` bumps only the new counter.
- **(#5) duplication** — `prepare_zenoh_outbound` now shares publisher lookup and `ZenohOutbound` construction between the two call sites; the two enqueue policies live in `enqueue_required_zenoh_outbound` and `handle_publish_enqueue_result`.
- **(#3) test timing fragility** — `output_closed_waits_for_publish_capacity_without_dropping` now waits with `tokio::time::timeout(Duration::from_secs(1), publish_rx.recv())` instead of relying on a `yield_now()`, so it no longer breaks under a multi-thread runtime.

Still open (unchanged by this commit) — flagging again since there was new activity:

**(#1) The "reliable" close event can still be silently discarded at the zenoh layer.** `prepare_zenoh_outbound` looks the publisher up in `dataflow.publishers` for both paths, so `try_send_to_remote_receivers` (regular data) and `send_to_remote_receivers` (`OutputClosed`) share the same publisher — declared with `.congestion_control(CongestionControl::Drop)` in `binaries/daemon/src/lib.rs:5386`. Making the in-process channel enqueue reliable only moves the drop one layer down: when the remote link is saturated, `enqueue_required_zenoh_outbound` waits for a permit and `publish_zenoh_outbound` calls `publisher.put(...)`, and zenoh then discards the close for the same backpressure reason. `publish_zenoh_outbound` bumps `net_publish_failures` and returns, the downstream daemon's node never receives `InputClosed`, and the dataflow hangs exactly as before. A real fix needs a separate `CongestionControl::Block` publisher (or a retry/ack) for control events.

**(#2, partial)** The newly added tests do exercise the enqueue helpers directly (an improvement), but none exercise the actual `close_outputs` → `send_to_remote_receivers` path with a saturated channel, which is where the promised guarantee lives.

---

SunSunSun689:
Addressed the remaining review follow-ups in `ee883cf52`:

- `OutputClosed` now uses a separate cached Zenoh publisher with `CongestionControl::Block`, while regular data output keeps `CongestionControl::Drop`. This keeps the data path non-blocking but avoids silently dropping required close/control events at the Zenoh publisher layer.
- Replaced the bare `mpsc` ordering tests with a daemon-path regression test that goes through `try_send_to_remote_receivers` and `send_output_closed_events`, verifies queued data still uses the drop publisher, and verifies the close event waits for capacity and uses the block publisher.

Verified with:
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo test -p dora-daemon fault_tolerance_tests -- --nocapture`
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo fmt --all -- --check`

phil-opp:
**Automated review by Claude** — this is a fully automated review, with no human in the loop; please verify before acting.

Re-reviewed after `ee883cf` ("use blocking publisher for output closes"), the one non-merge commit since the last review.

**Resolved:** the prior concern — `OutputClosed` sharing the `Drop` publisher and so still being silently discardable at the zenoh transport — is fixed. `prepare_zenoh_outbound` now resolves control events from a separate `control_publishers` map whose publisher is declared `CongestionControl::Block`, while data keeps `Drop`. The new `output_closed_uses_block_publisher_without_passing_queued_data` test drives the real `try_send_to_remote_receivers` / `send_output_closed_events` paths and asserts the two distinct publishers and congestion settings, so it is no longer a bare-`mpsc` test. Good.

Two things introduced by the `Block` approach are worth a look:

**1. Head-of-line blocking on the shared drain task (potential daemon-wide publish stall).** The `Block` publisher's `put().await` executes inside the *single* per-daemon drain task (`_zenoh_drain_handle`), which drains `zenoh_publish_rx` sequentially and awaits each `publish_zenoh_outbound`. If one remote link is saturated — or a remote peer stalls while its session stays alive — the close's `put()` blocks, and that stalls **all** subsequent inter-daemon publishing (data *and* control, across every dataflow on the daemon) until the link drains, potentially indefinitely. This swaps the old silent-drop for a possible whole-daemon outbound hang triggered by a single slow/stuck downstream. Consider a bounded/timeout'd block, a dedicated drain task per control stream, or an ack-based scheme, so one bad receiver cannot wedge the shared drain.

**2. Cross-publisher ordering.** Data and close now go through two different zenoh `Publisher` instances on the same topic. The shared FIFO channel guarantees `put(data)` is awaited before `put(close)`, but zenoh's ordering guarantee is per-publisher, not across two publishers on the same key-expression — so on a saturated link a close could still be delivered ahead of a trailing data sample, closing the receiver's input before the last data arrives. This is narrower than the earlier two-channel reordering issue, but the single-FIFO argument that closed that one no longer strictly holds once two publishers are in play. Worth confirming zenoh preserves inter-publisher order here, or routing both through one publisher with per-kind congestion handled another way.

Neither blocks the core fix, but (1) is a real availability trade-off worth a decision before merge.

---

SunSunSun689:
Addressed the follow-up availability/ordering concerns in `aa261e323`:

- Removed the separate `control_publishers` map. Regular data and `OutputClosed` for the same output now share one cached per-output Zenoh publisher again, so the single FIFO drain no longer relies on cross-publisher ordering on the same topic.
- The shared publisher is declared with `CongestionControl::Block`, so required close events do not go through a dropping publisher. Regular data remains non-fatal at the daemon enqueue layer: if the bounded publish queue is full, the data enqueue is recorded as `publish_enqueue_failures` and dropped instead of stalling the event loop.
- Bounded the actual Zenoh `put().await` in the drain task with `ZENOH_REQUIRED_PUBLISH_TIMEOUT` so a saturated/stuck remote link cannot block the shared drain indefinitely. Timeout/failure increments `net_publish_failures` and the drain continues.
- Updated the regression test to drive the real `try_send_to_remote_receivers` / `send_output_closed_events` paths and assert that queued data and close use the same publisher with `CongestionControl::Block`.

Verified with:
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo test -p dora-daemon fault_tolerance_tests -- --nocapture`
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo fmt --all -- --check`
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo clippy -p dora-daemon -- -D warnings`

phil-opp:
**Automated review by Claude** — this is a fully automated review, with no human in the loop; please verify before acting.

Re-reviewed after `aa261e32` ("preserve close ordering without publisher stalls"), the one non-merge commit since the last review.

**Both prior concerns look resolved:**
- *Cross-publisher ordering* — the separate `control_publishers` map is gone. Data and `OutputClosed` for an output share one cached publisher and one FIFO (`zenoh_publish_tx`) again, so a close can't be published on a different `Publisher` than its data. `output_closed_shares_blocking_stream_with_queued_data` asserts `Arc::ptr_eq` of the two publishers on the real daemon path, so it's a genuine check rather than a bare-`mpsc` test.
- *Unbounded drain stall* — the drain's `put()` is now wrapped in `timeout(ZENOH_REQUIRED_PUBLISH_TIMEOUT = 1s)`, so a stuck remote link can no longer wedge the shared drain indefinitely; on timeout it drops and bumps `net_publish_failures`.

**One residual tradeoff this commit introduces (new vs. earlier reviews).** Because ordering forces data and close onto the *same* publisher, that shared publisher is now declared `CongestionControl::Block` — so regular *data* `put()` now also blocks (up to the 1s timeout) inside the single per-daemon drain task (`_zenoh_drain_handle`). A single slow/absent remote consumer can therefore delay the drain — data *and* close, across every dataflow on the daemon — by up to ~1s per queued message, whereas the pre-PR data path (`Drop`) never blocked the drain at all. The main event loop stays safe (it only `try_send`s / spawns), and the stall is bounded rather than indefinite, so this is a latency tradeoff, not a hang. Worth a conscious decision that a bounded per-message drain stall on the data plane is acceptable; otherwise the close event needs its own drain/publisher rather than dragging data onto `Block`.

Relatedly, it's worth being explicit that the "required" close is now *best-effort-with-wait*, not guaranteed: after the 1s `put` timeout it is dropped (counted in `net_publish_failures`). That matches the bounded-block direction from the last review — just note the guarantee is "wait up to ~1s, then drop", not "never drop".

---

phil-opp:

Re-reviewed after `ee883cf` + `aa261e3`.

- The earlier concern that the two added tests didn't exercise dora code is resolved: `output_closed_shares_blocking_stream_with_queued_data` now builds a real daemon and drives the actual publish path.
- The earlier "close silently dropped under congestion" concern is mostly addressed: the publisher is now `Block`, and a timed-out/failed `put()` is counted in `net_publish_failures`, so a lost close is no longer silent (though a close under >1s sustained saturation can still time out and drop — best-effort, which seems acceptable).

One new concern from *how* that was fixed. The fix flips the **shared** cached per-output publisher to `CongestionControl::Block`, so regular data outputs now also publish with `Block` (the new test even asserts this on the data sample). Combined with the single daemon-wide serial drain task:

```rust
while let Some(msg) = zenoh_publish_rx.recv().await {
    publish_zenoh_outbound(msg).await; // Block put, 1s timeout
}
```

a single saturated remote consumer can now head-of-line-block that serial task for up to `ZENOH_REQUIRED_PUBLISH_TIMEOUT` (1s) per message, delaying inter-daemon delivery to *every other* consumer/dataflow on the daemon — and once the 256-slot channel backs up, healthy consumers' data gets dropped at `try_send`. Under the previous `Drop` behavior `put()` returned immediately, so this looks like a newly-introduced cross-consumer isolation/throughput regression.

A **dedicated** publisher for the close/control events would give the close its `Block` reliability while leaving regular data on `Drop`, avoiding the interference. If the shared-`Block` design is intentional, it'd be worth a comment acknowledging that one slow remote link now throttles all inter-daemon output.

---

SunSunSun689:
This draft is now superseded by #3215.

Both PRs target the same underlying inter-daemon `OutputClosed` reliability/ordering issue. This draft preserved data-before-close ordering by keeping `OutputClosed` on the shared drain, but that shape can still create daemon-wide head-of-line stalls.

#3215 now uses a per-output FIFO instead: regular `Output` and `OutputClosed` share ordering for the same `OutputId`, while different outputs/dataflows do not share one blocking drain. That keeps the ordering property from this draft without reintroducing the shared-drain stall. The latest #3215 also includes regression coverage for `Output` before `OutputClosed` ordering plus the related reload/close bookkeeping fixes, so continuing the fix there should avoid two conflicting implementations.

phil-opp:
Closing in favor of #3215

PR #3051, opened 2026-08-05
perf(node): build the UInt8 IPC layout once per raw send

phil-opp:
## Summary

`send_output_raw` (the raw-bytes send path used by the C API, C++ `send_output`, and the Python raw path) sized its sample with `uint8_ipc_len(data_len)` and then wrote the header with `encode_uint8_ipc_header(dst, data_len)`. Both call `uint8_layout` internally, which builds the Arrow **schema and record-batch flatbuffer messages** — so every raw byte send constructed those two flatbuffers **twice**. The Python raw send site (`sample_handler.rs`) had the same double-build.

## Fix

Expose `uint8_layout` (returning a public `Uint8Layout`) with `total()` (to size the sample) and `write_header()` (to write into it), so a caller builds the layout once and reuses it. The former two functions remain as thin convenience wrappers for callers that only need one half, and `encode_uint8_ipc_header` now delegates to `write_header` so there is a single implementation of the writing logic. Both hot call sites now build the layout once.

The change is purely additive to the public API (no breaking changes) and the emitted bytes are unchanged.

## Validation

- New test `uint8_layout_reused_matches_two_call_path` asserts the compute-once path is **byte-identical** to the two-call path across several lengths (0, 1, 7, 8, 9, 1000).
- Existing zero-copy `copy_count` integration test still passes (no new payload copies).
- `cargo test -p dora-node-api` (lib + ipc_encode + copy_count), `cargo clippy -p dora-node-api -- -D warnings`, `cargo fmt --check` — all green. The Python crate (`dora-node-api-python`) compiles with the updated call site.

---

⚠️ **This is a machine-generated pull request authored by Claude (an AI agent).** Please review carefully before merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01HZRfXvMfoesmRj3M8Fdpmz)_

PR #3073, opened 2026-08-08
fix(node): guard the SHM copy-fallback against an over-allocated buffer

phil-opp:
## Issue

The heap→SHM copy path in `zenoh_publish` (`apis/rust/node/src/node/mod.rs`) does:

```rust
Ok(mut sbuf) => {
    sbuf.as_mut().copy_from_slice(&avec);   // no length check
```

on the buffer returned by `provider.alloc(avec.len())`. `<[u8]>::copy_from_slice`
**panics** when the source and destination lengths differ, and this runs on the
node's send thread — so a length mismatch crashes the node rather than falling
back.

The sibling allocation site `allocate_data_sample` already guards exactly this
case:

```rust
if sbuf.as_ref().len() == data_len {
    return Ok(DataSample { storage: SampleStorage::Shm(sbuf) });
}
// "If a future provider ever over-allocates, fall back to heap rather than
//  expose or publish an oversized slice."
```

So the codebase already recognizes that `provider.alloc(n)` may return a buffer
whose `.len()` exceeds `n` and guards one call site — but not this one. The two
paths were inconsistent: one falls back safely, the other panics.

## Fix

Mirror the blessed guard: only `copy_from_slice` when
`sbuf.as_mut().len() == avec.len()`. On a mismatch, log at `debug` and fall
through to the reliable daemon path (the `avec` payload is still owned, so the
existing `NotPublished(Vec(avec))` fallthrough at the bottom of the arm routes
it correctly).

zenoh 1.8 guarantees the logical length matches the request today, so this is
defensive hardening that closes a latent panic and makes the two SHM-allocation
sites consistent. No behavior change on the normal (equal-length) path.

## Validation

- `cargo build -p dora-node-api` — clean
- `cargo test -p dora-node-api --lib` — 162 passed
- `cargo clippy -p dora-node-api --lib -- -D warnings` — clean
- fmt clean

(The 2 failing `zenoh_schema_cache` integration tests fail identically on
unmodified `origin/main` — they need zenoh networking unavailable in the CI
sandbox — so they are pre-existing and unrelated to this change.)

---

🤖 This is a machine-generated pull request authored by Claude (Claude Code), part of an automated codebase review. Each finding was verified against `origin/main` before opening. Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PzFF4tZy8gAEh9sBeXCJcf)_

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -1936,24 +1936,43 @@ impl DoraNode {
                         .wait()
                     {
                         Ok(mut sbuf) => {
-                            sbuf.as_mut().copy_from_slice(&avec);
-                            if diag {
-                                tracing::warn!(
-                                    "output `{output_id}`: entering zenoh put of a \
-                                     copied SHM buffer (dora-rs/dora#2742 diagnostic)"
-                                );
-                            }
-                            return match publisher.put(sbuf).attachment(&metadata_bytes[..]).wait()
-                            {
-                                Ok(()) => Ok(PublishOutcome::Published),
-                                Err(e) => {
+                            // Mirror the guard in `allocate_data_sample`: only
+                            // copy into the SHM buffer when it is exactly the
+                            // requested size. zenoh 1.8 guarantees the logical
+                            // length matches the request, but `copy_from_slice`
+                            // requires equal lengths and would panic on the
+                            // node's send thread if a future provider ever
+                            // over-allocated. Fall through to the reliable
+                            // daemon path instead of risking that panic.
+                            if sbuf.as_mut().len() == avec.len() {
+                                sbuf.as_mut().copy_from_slice(&avec);
+                                if diag {
                                     tracing::warn!(
-                                        "zenoh SHM publish failed ({e}); \
-                                         falling back to daemon path"
+                                        "output `{output_id}`: entering zenoh put of a \
+                                         copied SHM buffer (dora-rs/dora#2742 diagnostic)"
                                     );
-                                    Ok(PublishOutcome::NotPublished(FinalizedSample::Vec(avec)))
                                 }
-                            };
+                                return match publisher
+                                    .put(sbuf)
+                                    .attachment(&metadata_bytes[..])
+                                    .wait()
+                                {
+                                    Ok(()) => Ok(PublishOutcome::Published),
+                                    Err(e) => {
+                                        tracing::warn!(
+                                            "zenoh SHM publish failed ({e}); \
+                                             falling back to daemon path"
+                                        );
+                                        Ok(PublishOutcome::NotPublished(FinalizedSample::Vec(avec)))
+                                    }
+                                };
+                            }
+                            tracing::debug!(
+                                "zenoh SHM alloc returned {} bytes for a {}-byte \
+                                 request; using daemon path",
+                                sbuf.as_ref().len(),
+                                avec.len()
+                            );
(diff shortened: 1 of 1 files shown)


phil-opp:

This PR is part of an automated review sweep of the repository. Findings were verified against `origin/main` before opening; each PR is small, self-contained, branched off fresh `origin/main`, and machine-generated (please review carefully).

### PRs opened this run

**Correctness**
- #3073 — `fix(node)`: guard the SHM copy-fallback against an over-allocated buffer (mirror the `allocate_data_sample` length guard so `copy_from_slice` can't panic on the send thread)
- #3074 — `fix(coordinator)`: skip finalized nodes when exporting OTEL metrics (apply the same `node_finalized` filter the in-memory table already uses)
- #3075 — `fix(node)`: measure the *serialized* JSON size for the log-fields budget, not the raw key/value byte sum (escaping/structure can blow the 64 KB downstream limit a raw sum passes)

_No optimization / documentation / simplification PR met the bar this run (see rejected below)._

- **hub-client `resolve()` — tolerate a corrupt version entry**: adversarial review showed this trades *fail-closed* corruption handling for a *silent version downgrade* (a downgrade-attack surface), and the crate has no logging facility to surface the skip. Policy call for maintainers — dropped.
- **mavlink2-bridge telemetry latency (~50 ms avg / up to 100 ms per frame)**: real and arguably the highest-impact finding, but the fix restructures the delicate shutdown/reconnect event loop (#2027 / #2034). Too architecturally significant for an automated PR without maintainer review.
- **mavlink2-bridge per-frame schema rebuild + `format!` loops**: legitimate micro-opt, but negligible at 5–50 Hz telemetry rates and needs ~17 per-callsite edits — churn/risk out of proportion to the gain.
- **node `zenoh_publish` redundant `bincode::serialize(metadata)` on the schema-once fast path**: real, but wasted on only one sub-path, and deferring it cleanly touches the hottest send path (already the subject of #3051 / #2945). Too risky for a marginal gain.
- **mavlink2-bridge transient UDP recv has no backoff**: a potential busy-spin, but in practice the queued socket error (SO_ERROR / ICMP) is consumed once per occurrence, so it doesn't sustain. Largely theoretical.
- **daemon `rotate_log_files` keeps one file when `max_rotated_files == 0`**: unreachable — descriptor validation already rejects `0` (`must be at least 1`).
- **message `InputMapping` Display drops `node_filter` when `min_level` is `None`**: latent only; the state isn't constructible via `FromStr` today.
- **node `log_with_fields` double-serialization**: addressed inline in #3075 (zero-alloc byte counter), not a separate PR.

### Validation notes
Each affected crate was verified individually with fmt + clippy (`-D warnings`) + tests (dora-node-api: 162 lib tests; dora-coordinator: metrics-feature tests). Changes are internal-only (no public API surface changed), so cross-crate breakage is precluded by construction. The full-workspace debug build could not run to completion in this environment — a complete build exceeds the session's writable-disk allowance and fails linking *unrelated* crates on `No space left on device`, not on any change here.

---

phil-opp:

I reviewed the diff and found no issues. The `if sbuf.as_mut().len() == avec.len()` guard before `copy_from_slice` correctly mirrors the existing `allocate_data_sample` length check, and on a mismatch the code logs a debug line and falls through to the reliable daemon path rather than panicking on the node's send thread. This is defensive (zenoh currently guarantees the length), so the absence of a dedicated test is reasonable.

---

PR #3075, opened 2026-08-08
fix(node): measure serialized JSON size for the log-fields budget

phil-opp:
## Issue

`DoraNode::log_with_fields` (`apis/rust/node/src/node/mod.rs`) guards a
downstream 64 KB log-line parse limit by dropping oversized `fields`, but it
measured the **raw** key/value byte sum:

```rust
let total: usize = fields.iter().map(|(k, v)| k.len() + v.len()).sum();
if total <= Self::MAX_LOG_FIELDS_BYTES {   // 60 KB
    entry["fields"] = serde_json::json!(fields);
```

The emitted `"fields":{...}` object is larger than that raw sum: JSON adds
structural bytes (quotes, colons, commas) and, more importantly, escaping — a
value full of `"`/`\` doubles, and control characters expand ~6× via `\uXXXX`.
A fields map whose raw content is ~20 KB but serializes to well over 60 KB
passes the gate and is then dropped or truncated **whole** by the downstream
parser — silently losing the very log line the guard exists to protect.

## Fix

Measure the **serialized** size instead, via a small testable helper
`log_fields_within_budget`. It counts the serialized bytes with a zero-alloc
`io::Write` byte counter (no throwaway string) and builds the JSON value only
when it fits.

## Validation

- Adds a regression test (`log_fields_budget_measures_serialized_json_not_raw_bytes`)
  with a control-character value that is under budget by raw bytes but over it
  once escaped. Confirmed **RED** on the old raw-sum logic, **GREEN** with the fix.
- `cargo test -p dora-node-api --lib` — 162 passed
- `cargo clippy -p dora-node-api --lib -- -D warnings` — clean
- fmt clean

---

🤖 This is a machine-generated pull request authored by Claude (Claude Code), part of an automated codebase review. Each finding was verified against `origin/main` before opening. Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PzFF4tZy8gAEh9sBeXCJcf)_

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -2130,8 +2130,10 @@ impl DoraNode {
         self.log_with_fields(level, message, target, None);
     }
 
-    /// Maximum total size of log fields before they are dropped (60 KB).
-    /// Matches the downstream 64 KB parse limit with headroom for the message envelope.
+    /// Maximum serialized size of the log `fields` object before it is
+    /// dropped (60 KB). Matches the downstream 64 KB parse limit with headroom
+    /// for the message envelope. Measured on the serialized JSON (see
+    /// [`log_fields_within_budget`]), not the raw key/value byte sum.
     const MAX_LOG_FIELDS_BYTES: usize = 60 * 1024;
 
     /// Send a structured log message with optional key-value fields.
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed the diff and found no issues. Measuring the actual serialized JSON size (via the zero-alloc `io::Write` byte counter, building the JSON value only if it fits) is the correct fix — JSON escaping can make the serialized form far exceed the raw `k.len() + v.len()` sum the old gate used, which is what the downstream 64 KB parser actually cares about. The regression test is non-trivial: a ~20 KB control-character value that is under budget by raw bytes but ~120 KB serialized is correctly dropped, while a small map passes.

---

phil-opp:
The `Unwrap budget` failure here comes from a **batch-mate in the Trunk merge queue, not from this PR**. The queue tested a combined branch (`trunk-merge/pr-2946`, SHA `d8530d07`) that batches several queued PRs together, and the combined count hit `164 > 163`.

Running the exact check (`scripts/qa/unwrap-budget.sh`) on this branch in isolation gives **161** — identical to `origin/main` (161) and well under the 163 budget. This PR adds no new `.unwrap()`/`.expect(` (its only additions are a `?`-based helper and a `#[cfg(test)]` test, which the script excludes). It should pass once the queue re-tests it without the offending PR in the batch.

---

PR #3076, opened 2026-08-08
fix(daemon): don't let a recycled leader pid drag a stranger's group into the destroy reaper (#3067)

### LINKED ISSUE #3067: daemon: destroy reaper can SIGKILL an unrelated process group after PID reuse — `live_children` group fallback skips the parentage re-check its comment claims (#3004)
> ⚠️ **This issue was created by a scheduled, automated Claude code-review check.** It was not filed by a human. Please verify the finding before acting on it. Severity is **low** — the trigger window is extremely narrow — but the group-kill path is a real gap and the code comment asserting it is closed is incorrect.

## Summary

The destroy reaper added in #3004 (`4612ca8`, *"reap node processes before a destroy replies"*) tracks a node's liveness by its **process group** once the leader pid is gone. The leader-liveness path re-verifies parentage (`is_live_child`), but the **group-fallback path does not** — it trusts the `confirmed` set (populated while the *original* node was alive) plus a bare `kill(-pid, 0)` existence probe. If that leader pid is recycled by the OS into an **unrelated** process that becomes a group leader, the reaper can issue `killpg(pid, SIGKILL)` against a **stranger's process group**.

The doc comment claims this cannot happen:

```rust
// binaries/daemon/src/shutdown.rs:124-126
// A group is only ever consulted for a pid confirmed to be ours while it
// was alive, so a recycled pid cannot drag in a stranger's group.
```

That reasoning only holds for the *leader* check. Being "confirmed ours while alive" at time T1 does **not** prevent the pid being recycled at T2, and the group check at T2 performs no parentage re-verification.

## Affected code

`binaries/daemon/src/shutdown.rs`

- `live_children` — the group fallback (no parentage check):
  ```rust
  // :140-149
  let leader_alive = self.system.process(Pid::from_u32(pid))
      .is_some_and(|process| is_live_child(process, self.own_pid));
  if leader_alive {
      self.confirmed.insert(pid);
      alive.push(pid);
  } else if self.confirmed.contains(&pid) && process_group_has_members(pid) {
      alive.push(pid);   // <-- confirmed-once + bare kill(-pid,0); no re-check of ownership
  }
  ```
- `process_group_has_members` (`:160-165`) is a pure `kill(-pid, 0) == 0` existence probe — no ownership check.
- `kill_process_group` (`:182-195`) then `killpg(pid, SIGKILL)`; its own comment (`:186-187`) also overstates the guarantee: *"A pid the OS recycled is already excluded by the parentage check in `live_children`."* — true only for the leader path.

`confirmed` (`:57`) is never cleared, so a pid stays "ours" for the whole `DestroyWait` even after the original leader has exited.

## Failure scenario

1. Node `X` is spawned as a process-group leader (pid `P`) — typically a wrapper such as `uv run python node.py`. It is seen alive-as-our-child, so `P` is inserted into `confirmed`.
2. A `Destroy` is in progress (we are inside the `DestroyWait::poll` loop).
3. `X`'s leader and its whole group exit, but `X`'s `running_nodes` entry is still present — its terminal `SpawnedNodeResult` is queued behind the `DestroyTick` in the daemon event loop, so `P` is still returned by `running_node_pids()`.
4. Within the poll/settle window, the OS reuses pid `P` for a new, unrelated process that calls `setsid`/`setpgid` and becomes a group leader with members.
5. `live_children`: `is_live_child(P)` is `false` (the stranger's parent ≠ the daemon), so it falls to `else if confirmed.contains(P) && process_group_has_members(P)` → **true** → `P` added to `alive`.
6. After the deadline, `kill_process_group(P)` issues `killpg(P, SIGKILL)` against the **stranger's** group.

## Why the window is narrow (but not empty)

Because the daemon is the node's parent, `P` is not reusable while the child is an un-reaped zombie (`is_live_child` returns `false` for `Zombie`, and the group is empty, so it isn't selected). The exploitable window is only between the daemon **reaping** the child (freeing `P`) and the daemon **processing** the queued terminal event that removes `P` from `running_nodes`. On a normal Linux host (`pid_max` in the millions) hitting exactly `P` in that sub-millisecond window is astronomically unlikely; on a container/embedded host with a small `pid_max` under heavy process churn it is merely very unlikely. Consequence if it does fire — SIGKILL of an unrelated process group — is worse than the orphan-leak the reaper exists to prevent, which is why it seems worth recording.

## Suggested fix

In the group-fallback branch, do not rely on `confirmed` alone once the leader pid is gone. Options:

- Re-verify ownership before killing: confirm at least one member of group `P` has the daemon in its ancestry (walk `parent()` up to `own_pid`), rather than trusting a bare `kill(-P, 0)`.
- Or track the node's actual child pids observed while alive and check those, instead of the recycled leader pid.
- Or, cheapest partial mitigation: drop the pid from `confirmed` (and stop consulting its group) once its `running_nodes` entry is gone, closing the "entry lingers after reap" window — and correct the two comments (`shutdown.rs:124-126`, `:186-187`) so they no longer assert a guarantee the group path does not provide.

## Related secondary edge case (same subsystem, also #3004)

`begin_pending_destroy` treats "no non-zero pids" as "nothing to reap, exit now":

```rust
// binaries/daemon/src/lib.rs:2087
if self.running_node_pids().is_empty() {
    return Some(reply_tx);   // daemon exits immediately
}
```

but `running_node_pids()` (`lib.rs:2156-2165`) filters out `pid == 0`, which per its own comment means *"the entry exists but its process has not reported a pid yet (`prepared.rs` stores it right after spawn)."* If a `Destroy` arrives in the sub-millisecond window after a node's process is spawned but before its pid Arc is populated, `begin_pending_destroy` concludes there is nothing to reap and lets the daemon exit — orphaning that just-spawned process, the exact leak #2980/#3004 exists to close. Also very narrow, but worth folding into the same fix (e.g. treat a present-but-pid-0 entry as "still starting, keep waiting" rather than "empty").

## How this was found

A scheduled automated review examined the last ~2 days of merged commits, traced the #3004 destroy-reaper lifecycle code (`DestroyWait`, `live_children`, `kill_process_group`, `begin_pending_destroy`), and cross-checked the parentage guarantees each path actually enforces against the guarantees its comments claim.

phil-opp:
Fixes #3067.

## Problem

The destroy reaper added in #3004 tracks a node's liveness by its **process group** once the leader pid is gone. The leader-liveness path re-verifies parentage (`is_live_child`), but the **group-fallback path did not** — it trusted the `confirmed` set (populated while the *original* node was alive) plus a bare `kill(-pid, 0)` existence probe.

A process group can outlive its leader, and the leader pid is then free for the OS to reuse. If it is recycled into an **unrelated live process** that becomes a group leader, the reaper could issue `killpg(pid, SIGKILL)` against a **stranger's process group**. The doc comments at `shutdown.rs:124-126` and `:186-187` asserted this could not happen; that reasoning only held for the leader path.

The window is narrow (only between the daemon reaping the child — freeing the pid — and processing the queued terminal event that removes it from `running_nodes`, and only if the pid is recycled in that sub-millisecond gap), but the consequence — SIGKILL of an unrelated process group — is worse than the orphan leak the reaper exists to prevent.

## Fix

**`live_children` group-fallback guard (`shutdown.rs`).** In the group-fallback branch, exclude a pid that has been recycled into a stranger. The branch is only reached when the leader is *not* a live child of ours, so a live **non-zombie** process still occupying that pid is parented elsewhere → its group must not be attributed to us. A pid that is simply gone (no process), or is our own not-yet-reaped **zombie** leader, is not a recycle, so the group remnant stays trusted — the legitimate "wrapper died, its child lives on in the group" case (#3004) is preserved. The two overstated comments are corrected.

**Secondary window in `begin_pending_destroy` (`lib.rs`).** `running_node_pids()` filters out entries whose pid `Arc` still reads `0` — a node that was spawned but has not reported its pid yet. A `Destroy` arriving in that window concluded "nothing to reap" and let the daemon exit, orphaning the just-spawned process — the exact leak #2980/#3004 exist to close. A present-but-pid-0 entry is now treated as "still starting, keep waiting" (new `has_starting_node` helper, checked both at destroy-begin and when `poll` reports `Done`).

## Tests

Added `a_recycled_leader_pid_does_not_drag_in_a_strangers_group` in `shutdown.rs`: a confirmed leader pid that later presents as a live non-child (as a recycled pid would) is kept out of the live set even though the bare group probe still passes. The existing `a_wrapper_that_exits_does_not_hide_its_surviving_child` guards the legitimate orphaned-group case, which still passes.

- `cargo test -p dora-daemon --lib shutdown::` — 6 passed, 0 failed

> The `--all-targets` clippy run surfaces one pre-existing `match_like_matches_macro` lint in unrelated test code (present on `main`, at the `NodeEvent`/`&str` match helper); left untouched per the "don't fix unrelated warnings" convention, and CI's clippy invocation does not lint test targets.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WvtaXNUmfXMCZ8fuFKJwCj)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2084,7 +2084,7 @@ impl Daemon {
             tokio::spawn(Self::finish_destroy(reply_tx));
             return None;
         }
-        if self.running_node_pids().is_empty() {
+        if self.running_node_pids().is_empty() && !self.has_starting_node() {
             return Some(reply_tx);
         }
 
@@ -2127,7 +2127,18 @@ impl Daemon {
 
         let survivors = match progress {
             shutdown::DestroyProgress::Waiting => return false,
-            shutdown::DestroyProgress::Done => Vec::new(),
+            shutdown::DestroyProgress::Done => {
+                // A node whose entry exists but whose pid Arc still reads 0 was
+                // spawned but has not reported its pid yet. `running_node_pids`
+                // filters those out, so `poll` sees "nothing to reap" and would
+                // let the daemon exit, orphaning the just-spawned process — the
+                // exact leak the reaper exists to close (#3067). Keep waiting
+                // until every entry has either reported a pid or been removed.
+                if self.has_starting_node() {
+                    return false;
+                }
+                Vec::new()
+            }
             shutdown::DestroyProgress::Abandoned(survivors) => survivors,
         };
 
@@ -2163,6 +2174,21 @@ impl Daemon {
             .map(|pid| pid.load(atomic::Ordering::Acquire))
             .filter(|pid| *pid != 0)
             .collect()
+    }
+
+    /// Whether any running node was spawned but has not reported its pid yet
+    /// (its pid Arc still reads 0, set right after spawn in `prepared.rs`).
+    ///
+    /// Such a node is mid-spawn: `running_node_pids` filters it out, so a
+    /// destroy landing in that window must not conclude there is nothing to
+    /// reap. Dynamic nodes carry no pid Arc and are not counted — the daemon
+    /// did not spawn them and cannot reap them.
+    fn has_starting_node(&self) -> bool {
+        self.running
+            .values()
+            .flat_map(|dataflow| dataflow.running_nodes.values())
+            .filter_map(|node| node.pid.as_ref())
+            .any(|pid| pid.load(atomic::Ordering::Acquire) == 0)
     }
 
     /// Send the destroy reply and wait for it to go out.
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2084,7 +2084,7 @@ impl Daemon {
             tokio::spawn(Self::finish_destroy(reply_tx));
             return None;
         }
-        if self.running_node_pids().is_empty() && !self.has_starting_node() {
+        if self.running_node_pids().is_empty() {
             return Some(reply_tx);
         }
 
@@ -2127,18 +2127,7 @@ impl Daemon {
 
         let survivors = match progress {
             shutdown::DestroyProgress::Waiting => return false,
-            shutdown::DestroyProgress::Done => {
-                // A node whose entry exists but whose pid Arc still reads 0 was
-                // spawned but has not reported its pid yet. `running_node_pids`
-                // filters those out, so `poll` sees "nothing to reap" and would
-                // let the daemon exit, orphaning the just-spawned process — the
-                // exact leak the reaper exists to close (#3067). Keep waiting
-                // until every entry has either reported a pid or been removed.
-                if self.has_starting_node() {
-                    return false;
-                }
-                Vec::new()
-            }
+            shutdown::DestroyProgress::Done => Vec::new(),
             shutdown::DestroyProgress::Abandoned(survivors) => survivors,
         };
 
@@ -2174,21 +2163,6 @@ impl Daemon {
             .map(|pid| pid.load(atomic::Ordering::Acquire))
             .filter(|pid| *pid != 0)
             .collect()
-    }
-
-    /// Whether any running node was spawned but has not reported its pid yet
-    /// (its pid Arc still reads 0, set right after spawn in `prepared.rs`).
-    ///
-    /// Such a node is mid-spawn: `running_node_pids` filters it out, so a
-    /// destroy landing in that window must not conclude there is nothing to
-    /// reap. Dynamic nodes carry no pid Arc and are not counted — the daemon
-    /// did not spawn them and cannot reap them.
-    fn has_starting_node(&self) -> bool {
-        self.running
-            .values()
-            .flat_map(|dataflow| dataflow.running_nodes.values())
-            .filter_map(|node| node.pid.as_ref())
-            .any(|pid| pid.load(atomic::Ordering::Acquire) == 0)
     }
 
     /// Send the destroy reply and wait for it to go out.
--- a/binaries/daemon/src/shutdown.rs
+++ b/binaries/daemon/src/shutdown.rs
@@ -124,12 +124,13 @@ impl DestroyWait {
     /// one level down that this module exists to prevent (#3004 review).
     ///
     /// The group is only consulted for a pid confirmed ours while alive, and
-    /// only while that pid is *not* currently a live foreign process. Once a
+    /// only while that pid has not been recycled out from under us. Once a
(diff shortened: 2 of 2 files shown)


phil-opp:

I reviewed the diff and found no issues. The recycled-pid guard on the group-fallback path (requiring `!recycled_into_stranger` before falling back to `killpg`) and the new `has_starting_node` check that keeps the destroy poll from reporting `Done` while a spawned-but-pid-not-yet-reported node exists both look correct. The new `#[cfg(unix)]` test meaningfully exercises the recycle scenario (spawns a real group leader, simulates pid recycle by changing `own_pid`, and asserts the stranger group is excluded).

---

phil-opp:
**Automated review by Claude** (fully automated — no human has vetted this comment; please verify before acting).

The recycled-stranger guard in `live_children` (and its new test) looks correct — it fails on base code and passes with the fix, and it captures the real recycle case.

On the second half of the PR, though, I don't think `has_starting_node()` actually does anything:

- `has_starting_node()` returns true only if a *registered* `running_nodes` entry has its pid Arc reading 0 (`lib.rs`, iterating `dataflow.running_nodes.values()`).
- But every `RunningNode` is produced by `PreparedNode::spawn()`, which creates the pid Arc as `AtomicU32::new(0)` (`spawn/prepared.rs:194`) and then stores the real pid into it (`prepared.rs:212`, and `:536` on restart) *before* the `RunningNode` struct is built. The entry is inserted into `running_nodes` only after `spawn()` returns (`lib.rs:1946`, `:2773`, `:3302`), so at every insert site the Arc already holds a non-zero pid. The restart loop likewise overwrites with `new_pid` and never resets to 0.

So a registered entry never holds pid 0, `has_starting_node()` is always false, and the `begin_pending_destroy` condition (`running_node_pids().is_empty() && !has_starting_node()`) is unchanged in effect. The orphan window described in #3067 — process spawned, `SpawnNodeResult` still in flight, *no* `running_nodes` entry yet — is exactly the state `has_starting_node()` cannot observe, because it only scans registered entries. The code comments (both the new doc and the `Done`-branch comment) assert the pid Arc "still reads 0" for a mid-spawn registered node, but that state doesn't appear to ever be registered.

Worth confirming whether this half exercises anything. Separately, the `Done`-branch `return false` bypasses `poll`'s deadline/kill logic, so if a registered pid-0 entry ever *did* persist (a future regression), the destroy would loop forever and skip the `DESTROY_WAIT` backstop that otherwise bounds every destroy.

---

PR #3078, opened 2026-08-08
feat(memory-pool): cross-machine zero-copy tensor transport for end-edge-cloud collaboration

tang-canran:
### (1) Design — new architecture for multi-daemon and cross-machine deployment

This PR extends the memory-pool transport to the deployment topologies real edge-cloud robotics actually run: **multiple daemons on one host** (sensor, perception, and inference pipelines as separate daemons on a robot or edge box) and **daemon clusters across machines** (edge ↔ cloud). The additions:

**Same-host multi-daemon — direct read across daemons**
- The receiver locates the sender's pool segment **directly by machine-qualified name** (`pool_{node_id}_{machine}_{counter}`) and reads it in place, bypassing the daemon relay and even the mirror push entirely (`same-host direct detection`). Native dora has no such path: outputs to a consumer on another daemon are pinned to the daemon relay (#2738), so the same topology costs 89 MB/s instead of ~5.8 GB/s (65×).
- For GPU pools, the sender exports the CUDA IPC handle into the pool header at registration; a same-host cross-daemon receiver imports it once and reads device memory zero-copy (32 GB/s measured).
- Control-plane notifications (RegisterPool/RegisterPoolAck/FreePool) between daemons ride a zenoh SHM payload for same-host delivery.

**Cross-machine — reliable large-frame data plane with GPU staging**
- The data plane uses the daemon's zenoh channel with **blocking congestion control**: large frames queue and drain over slow WAN links instead of being dropped. Native dora's relay uses Drop + express and silently loses any payload above ~8 MiB on a slow link (queue limit of 16 batches), leaving **no working native path for WAN large frames** — the pool is the only option.
- GPU endpoints get **CPU staging pools**: a pageable transit pool on the send side (GPU_A → DtoH → CPU_A) and a pinned pool on the receive side (CPU_B → HtoD → GPU_B), both registered and reused per frame; the full path GPU→CPU→zenoh TCP→CPU→GPU runs at WAN bandwidth (~38 MB/s).
- Machine-qualified pool names and the daemon-side memory manager (`touched_by` / targeted cleanup) prevent cross-machine ID aliasing and clean up pools on every daemon involved.

Existing single-daemon behavior is unchanged; the event stream, zenoh, and daemon relay code are untouched.

### (2) Performance: memory pool vs native dora (3 scenarios × 4 device pairs)

Benchmark: 100 frames of 10000×512 int64 (40.96 MB), turn-based per-frame handshake, `Average transfer throughput` (`data_bytes / (t_received − t_send)`), same payload and cadence on both sides. Native dora runs with the 256 MiB zenoh SHM pool configured (its best zero-copy configuration).

**Memory pool (MB/s, this PR):**

| Scenario | cpu2cpu | cpu2cuda | cuda2cpu | cuda2cuda |
|----------|---------|----------|----------|-----------|
| Single daemon | 6157.9 | 3197.1 | 6792.7 | 44145.8 (same-GPU IPC) |
| Same-host cross-daemon | 5787.5 | 2616.6 | 6977.6 | 32691.6 (IPC direct read) |
| Cross-machine WAN | 38.0 | 37.5 | 38.3 | 37.5 (GPU_A→CPU→TCP→CPU→GPU_B) |

**Native dora (MB/s, same benchmark):**

| Scenario | cpu2cpu | note |
|----------|---------|------|
| Single daemon | 2409.6 | zenoh SHM zero-copy active |
| Same-host cross-daemon | 89.3 | pinned to the daemon relay (#2738); node-to-node zenoh mesh is same-machine-only |
| Cross-machine WAN | **unusable at this size** | relay uses Drop + express; >~8 MiB payloads silently dropped on slow links (16-batch queue limit), producer blocks on its ack forever |

Speedup: **2.6×** single-daemon, **65×** same-host cross-daemon, and **no working native path** for WAN large frames.

### (3) Purpose and significance — enabling end-edge-cloud tensor pipelines

VLA models and world models are driving explosive demand for on-device inference [1–4]; end-edge co-computing is becoming the mainstream deployment paradigm, keeping data transport inside the LAN at 1–6 ms round trips versus tens-to-hundreds of milliseconds over the WAN to the cloud [10,11]. A robot pipeline in this paradigm moves tens-of-MB GPU tensors (camera frames, point clouds, VLA features) through several processes — sensor → preprocessing → inference — on the same host, and then to the edge cluster or the cloud.

This PR makes that pipeline zero-copy at every hop:

- **On the end device**: tensors stay on the GPU across processes (same-GPU CUDA IPC at 42–44 GB/s; CPU↔GPU at 3–7 GB/s), and the zero-copy read frees the device's limited CPU for inference instead of serialization and relay copies.
- **At the edge (multi-daemon, LAN)**: separate daemons per pipeline stage no longer pay the daemon-relay tax — direct cross-daemon reads deliver 65× over native dora, so multi-process edge pipelines scale without the message-passing overhead.
- **Edge ↔ cloud (WAN)**: native dora cannot carry large frames across machines at all; the pool's blocking-control relay delivers them reliably at WAN bandwidth — closing the gap where distributed deployments currently have no viable data path for training-data uploads, fleet telemetry, or cloud-assisted inference.

One API (`write_memory_pool` / `read_memory_pool`) covers end, edge, and cloud with automatic device- and topology-aware path selection, turning message-passing deployments into shared-state pipelines without changing the dataflow description.


tang-canran:
Reopened from the fork's memory-pool branch per our PR convention (content identical; branch re-pointed to the same state).

PR #3080, opened 2026-08-08
fix(node): JSON-escape plain-text input in read_json_bytes_as_arrow

phil-opp:
## Issue

`read_json_bytes_as_arrow` (`apis/rust/node/src/daemon_connection/json_to_arrow.rs`) first tries to parse the input as JSON, and on failure retries by treating it as a plain string. The retry (`wrapped_quoted`) wrapped the **raw** bytes in `"` quotes:

```rust
fn wrapped_quoted(data: impl BufRead) -> impl BufRead {
    let quoted = "\"".as_bytes().chain(data).chain("\"".as_bytes());
    wrapped(quoted)
}
```

If the input itself contains a `"`, a backslash, or a raw control character (newline/tab), the wrapped result is **invalid JSON**, so both the JSON attempt and the string-fallback attempt fail and the whole call errors with `failed to infer JSON schema` instead of yielding a string array.

This is reachable from `dora run` **interactive mode**, whose Data prompt advertises "String/JSON …" (`apis/rust/node/src/daemon_connection/interactive.rs:140`). Concretely, these all failed before this change:

- typing `say "hi"` → produced `{ "inner":"say "hi""}` (invalid JSON)
- entering a Windows path `C:\Users\node`
- pasting a multi-line value

## Fix

JSON-escape the bytes via `serde_json` before wrapping, so any plain-text input becomes a valid JSON string literal. `wrapped_quoted` is replaced by a small `json_quote` helper. Behavior for bare JSON values (numbers, objects, arrays) is unchanged — they still parse natively on the first attempt.

## Validation

- Added regression tests in `json_to_arrow.rs` covering double quotes, backslashes, and control characters round-tripping to the expected string, plus a guard that a bare JSON number still parses as a non-string array.
- `cargo test -p dora-node-api --lib json_to_arrow` — 4 passed.
- `cargo fmt --all -- --check` and `cargo clippy -p dora-node-api -- -D warnings` clean.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01SD4dBVimzepwKSh2b9F8aG)_

### DIFF
--- a/apis/rust/node/src/daemon_connection/json_to_arrow.rs
+++ b/apis/rust/node/src/daemon_connection/json_to_arrow.rs
@@ -10,9 +10,14 @@ pub fn read_json_bytes_as_arrow(data: &[u8]) -> eyre::Result<ArrayData> {
     match arrow_json::reader::infer_json_schema(wrapped(data), None) {
         Ok((schema, _)) => read_from_json_with_schema(wrapped(data), schema),
         Err(_) => {
-            // try again with quoting the input to treat it as a string
-            match arrow_json::reader::infer_json_schema(wrapped_quoted(data), None) {
-                Ok((schema, _)) => read_from_json_with_schema(wrapped_quoted(data), schema),
+            // Try again by treating the input as a plain string. The bytes may
+            // contain characters that are not valid on their own inside a JSON
+            // string literal (`"`, `\`, raw control characters), so JSON-escape
+            // them first instead of wrapping the raw bytes in quotes — otherwise
+            // any such input produces invalid JSON and is rejected.
+            let quoted = json_quote(data);
+            match arrow_json::reader::infer_json_schema(wrapped(quoted.as_bytes()), None) {
+                Ok((schema, _)) => read_from_json_with_schema(wrapped(quoted.as_bytes()), schema),
                 Err(err) => eyre::bail!("failed to infer JSON schema: {err}"),
             }
         }
@@ -39,10 +44,12 @@ fn wrapped(data: impl BufRead) -> impl BufRead {
     "{ \"inner\":".as_bytes().chain(data).chain("}".as_bytes())
 }
 
-// wrap data into JSON object to also allow bare JSON values
-fn wrapped_quoted(data: impl BufRead) -> impl BufRead {
-    let quoted = "\"".as_bytes().chain(data).chain("\"".as_bytes());
-    wrapped(quoted)
+/// JSON-encode the given bytes as a quoted string literal, escaping any
+/// characters that are not valid inside a JSON string (`"`, `\`, control
+/// characters). Non-UTF-8 bytes are replaced lossily.
+fn json_quote(data: &[u8]) -> String {
+    // Serializing a `String` into JSON never fails.
+    serde_json::Value::String(String::from_utf8_lossy(data).into_owned()).to_string()
 }
 
 /// convert the given JSON object to the closed arrow representation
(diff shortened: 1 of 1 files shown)


phil-opp:

This is the tracking comment for a machine-generated code-review pass over the repository (branches suffixed off freshly-fetched `origin/main`). Each finding was verified against current `origin/main` and checked against open PRs from prior passes to avoid duplicates.

### PRs opened, by category

**Correctness**
- #3080 — `fix(node)`: JSON-escape plain-text input in `read_json_bytes_as_arrow`. Plain text containing `"`, `\`, or control chars (e.g. `dora run` interactive `say "hi"`, Windows paths, multi-line pastes) was rejected instead of becoming a string.
- #3081 — `fix(daemon)`: honor `max_rotated_files == 0` in `rotate_log_files`. A `0` config kept a rotated log forever instead of none; now discards the current log and prunes pre-existing rotated files.
- #3082 — `fix(cli)`: parse JSONL log lines that omit or capitalize the `level` field. Lines missing `level` were silently dropped; capitalized levels were misclassified as stdout.

**Simplification / Optimization**
- #3083 — `refactor(core)`: resolve the descriptor once in `check_dataflow` (was resolving — and cloning the whole node topology — 3× per validation). Public signatures and error ordering preserved.

**Documentation**
- #3084 — `docs(arrow-convert)`: document and add a compile-checked doctest for the `ArrowData` `TryFrom<&ArrowData>` read-back API (scalar vs. slice/`Vec` length + null contracts).

### Rejected findings (with one-line reasons)

- **`OperatorId` skips the validation `NodeId`/`DataId` enforce** — latent only (the parser never builds an invalid id), and adding validation risks rejecting operator ids in existing dataflows (breaking change).
- **`InputMapping::Logs` `Display` drops `node_filter` when `min_level` is `None`** — unreachable; `FromStr` never produces that combination.
- **mavlink `wake_udp_recv` keeps a non-loopback bind host** — low confidence; the `READER_SHUTDOWN_GRACE` detach path already bounds the worst case.
- **`local_listener` `_` arm neither replies nor closes on an unexpected request** — low reachability (socket only expects `NodeConfig`), and a behavior change to a protocol/concurrency path is hard to cover with a regression test.
- **`PendingNodes::handle_dataflow_stop` discards `DataflowStatus` without a rationale comment** — not a demonstrated defect; maintainability-only.
- **arrow-convert missing `TryFrom<&ArrowData> for Vec<String>`** — API-symmetry/ergonomics gap (a feature), not a defect.
- **`decode_arrow_ipc` decodes only the first batch** — single-batch is the internal producer contract; no demonstrated harm.
- **`node_inputs` per-node `Vec` alloc / `json_quote` cold-path alloc** — cold or pre-existing; negligible.

### Validation

- **Per branch**: `cargo fmt --all --check` + `cargo clippy -- -D warnings` + `cargo test` for the affected crate(s) — all green.
- **Combined diff**: full workspace suite (`cargo test --all`, Python/examples excluded) green **except** the two `apis/rust/node/tests/zenoh_schema_cache.rs` Zenoh loopback integration tests, which fail only in this network-restricted sandbox (they need a loopback peer link) and are unrelated to any change here — none of the diff touches Zenoh code.
- `/review` + `/simplify` gates were run once over the combined diff; their findings (case-sensitive level match in #3082; pruning pre-existing rotated files in #3081) were folded into the respective PRs.

> ⚠️ These pull requests are **machine-generated** by Claude Code. Please review each carefully before merging.

---

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human in the loop. Findings are suggestions to verify.

Reviewed the diff against the surrounding code. The fix is correct: `json_quote` builds the string literal via `serde_json::Value::String(...).to_string()`, which escapes `"`, `\`, and control characters, so the subsequent `wrapped(...)` produces valid JSON for any plain-text input. Behavior for bare JSON values is unchanged — they still satisfy `infer_json_schema` on the first `wrapped(data)` attempt and never reach the fallback.

The added tests are meaningful and would fail against the old `wrapped_quoted` logic:
- `plain_text_with_double_quotes_round_trips` / `_with_backslash_` / `_with_control_characters_` each exercise exactly the inputs the old raw-quote path turned into invalid JSON, and assert the round-tripped string value.
- `bare_json_values_still_parse_natively` guards the no-regression case (a bare number stays a non-string array).

No issues found. (Minor, non-blocking: `String::from_utf8_lossy` means non-UTF-8 input is replaced lossily rather than errored — which is reasonable and matches the documented intent.)

---

phil-opp:
This `Unwrap budget` failure isn't caused by this PR. The ratchet counts **production** `.unwrap()`/`.expect()`; the only ones #3080 adds are in test code (`#[cfg(test)]` — the `.unwrap()`/`.expect()` in the new `read_json_bytes_as_arrow` tests), which the check ignores.

The failing job ran against the Trunk merge-queue **batch** branch (`d8530d0`, `trunk-merge/pr-2946`), which bundles this PR with others ahead of it in the queue. The batch reports:

```
unwrap/expect count: 164 (budget: 163)
```

The `+1` comes from a co-batched sibling PR that added a production unwrap without updating `.unwrap-budget`.

Verified locally on this branch alone:

```
unwrap/expect count: 161 (budget: 163)
```

It passes — 2 under budget. Nothing to fix here; #3080 should re-test clean once the queue re-batches it without the offending sibling. (I've deliberately not bumped `.unwrap-budget` in this PR, since that would mask another PR's unaudited unwrap — that budget change belongs to whoever added the unwrap.)

---

PR #3086, opened 2026-08-09
fix(core): move newer `bail!` sites out of expression position (partial #2835, complements #2843)

### LINKED ISSUE #2835: Build fails on newer Rust due to semicolon_in_expressions_from_macros

phil-opp:
## Summary

Part of #2835. **Complements the pre-existing #2843** — please land both.

`eyre::bail!` / `bail!` expand to a block ending in a semicolon (`{ return Err(..); }`). Used as the **tail expression of a block** or as a **bare `match` arm**, that trailing semicolon is in expression position, which trips `semicolon_in_expressions_from_macros` / `semicolon_in_expressions_from_non_local_macros` — denied under `future_incompatible` on recent nightlies, failing `dora-core` compilation.

## Relationship to #2843

#2843 (opened 2026-07-26, `Closes #2835`) fixes this lint class across the whole workspace (48 files: node API, CLI, etc.). Since then, newer code has merged into `main` that introduced **additional** expression-position `bail!` sites in `dora-core` which #2843 does not cover. This PR fixes those.

The two PRs are **disjoint — no overlapping files** — so both can land independently, and `dora-core` is only fully clean once both do:

| dora-core file | fixed by |
|---|---|
| `build/git.rs`, `descriptor/mod.rs`, `descriptor/validate.rs`, `descriptor/expand.rs` | **this PR** |
| `topics.rs`, `build/mod.rs` | #2843 |

## What changed here

Moved every affected call in the files above to statement position:

- **Block-tail calls** (`{ …; bail!(..) }`): added a trailing `;`.
- **Bare `match` arms** (`_ => bail!(..)`): wrapped in a `{ bail!(..); }` block.

18 sites total. Every one diverges (`bail!` returns early), so it still coerces to the surrounding arm/expression type via the never type — **no runtime behavior change**.

## Verification

Toolchain: `rustc 1.99.0-nightly (771916f90 2026-08-08)`.

- `cargo +nightly check -p dora-core` — the 13 diagnostics that were reported at these sites are gone. (Note: which sites a given nightly flags for the non-local `eyre::bail!` variant varies by toolchain build; the `topics.rs` / `build/mod.rs` sites owned by #2843 are the remaining ones needed for full cross-toolchain coverage.)
- `cargo +1.97.1 check -p dora-core` — clean on the pinned toolchain.
- `cargo +1.97.1 clippy -p dora-core -- -D warnings` — clean.
- `cargo +1.97.1 fmt -p dora-core -- --check` — clean.
- `cargo +1.97.1 test -p dora-core` — 272 passing.

### DIFF
--- a/libraries/core/src/build/git.rs
+++ b/libraries/core/src/build/git.rs
@@ -103,7 +103,7 @@ impl GitManager {
                 "the build directory is still in use by the following \
                     dataflows, please stop them before rebuilding: {}",
                 using.iter().join(", ")
-            )
+            );
         }
 
         let reuse = if self.clone_dir_ready(session_id, &clone_dir) {
@@ -293,7 +293,7 @@ impl GitFolder {
                             .log_message(LogLevel::Error, format!("{err:?}"))
                             .await;
                         cleanup_failed_clone(logger, &target_dir).await;
-                        bail!(err)
+                        bail!(err);
                     }
                 }
             }
@@ -345,7 +345,7 @@ impl GitFolder {
                     Ok(()) => target_dir,
                     Err(err) => {
                         cleanup_failed_clone(logger, &target_dir).await;
-                        bail!(err)
+                        bail!(err);
                     }
                 }
             }
@@ -379,7 +379,7 @@ impl GitFolder {
                     Ok(()) => target_dir,
                     Err(err) => {
                         cleanup_failed_clone(logger, &target_dir).await;
-                        bail!(err)
+                        bail!(err);
                     }
                 }
             }
@@ -432,12 +432,14 @@ impl GitFolder {
                         // in-progress set only covers one GitManager and the CLI
                         // builds with its own, so deleting here is how #2711
                         // wipes out a live build. Fail loudly, leave the dir.
-                        Err(err) => bail!(
-                            "couldn't verify clone dir {} is on commit {commit_hash}: {err:?}; \
-                             leaving it in place in case another build is writing it, \
-                             please retry",
-                            dir.display()
-                        ),
+                        Err(err) => {
+                            bail!(
+                                "couldn't verify clone dir {} is on commit {commit_hash}: {err:?}; \
+                                 leaving it in place in case another build is writing it, \
+                                 please retry",
+                                dir.display()
+                            );
+                        }
                     }
                 }
 
--- a/libraries/core/src/descriptor/expand.rs
+++ b/libraries/core/src/descriptor/expand.rs
@@ -693,10 +693,12 @@ fn rewrite_module_input(
                         }))
                     }
                     None if optional_inputs.contains(&port_name) => Ok(None),
-                    None => bail!(
-                        "module input reference `_mod/{}` not found in module node inputs",
-                        port_name,
-                    ),
+                    None => {
+                        bail!(
+                            "module input reference `_mod/{}` not found in module node inputs",
+                            port_name,
+                        );
+                    }
                 }
             } else if inner_node_ids.contains(&source_str) {
                 // Internal cross-reference: prefix with module_id
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -355,7 +355,7 @@ fn node_kind_mut(node: &mut Node) -> eyre::Result<NodeKindMut<'_>> {
                 "module node `{}` must be expanded before resolution — \
                  call expand_modules() first",
                 node.id
-            )
+            );
         }
         NodeKind::Standard(_) => {
             let source = match (&node.git, &node.branch, &node.tag, &node.rev) {
@@ -369,7 +369,7 @@ fn node_kind_mut(node: &mut Node) -> eyre::Result<NodeKindMut<'_>> {
                         other @ (_, _, _) => {
                             eyre::bail!(
                                 "only one of `branch`, `tag`, and `rev` are allowed (got {other:?})"
-                            )
+                            );
                         }
                     };
                     NodeSource::GitBranch {
@@ -378,7 +378,7 @@ fn node_kind_mut(node: &mut Node) -> eyre::Result<NodeKindMut<'_>> {
                     }
                 }
                 (None, _, _, _) => {
-                    eyre::bail!("`git` source required when using branch, tag, or rev")
+                    eyre::bail!("`git` source required when using branch, tag, or rev");
                 }
             };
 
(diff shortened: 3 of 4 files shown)


phil-opp:

I reviewed this diff and found no issues. The changes are mechanical: adding a trailing `;` to block-tail `bail!` / `eyre::bail!` calls and wrapping bare match-arm `bail!` calls in `{ …; }` blocks across `libraries/core/src/build/git.rs`, `descriptor/expand.rs`, `descriptor/mod.rs`, and `descriptor/validate.rs`. Every touched call diverges (early return), so it still coerces to the surrounding arm/expression type and runtime behavior is unchanged — the change only silences the `semicolon_in_expressions_from_macros` nightly lint.

One thing worth flagging for the maintainer: this overlaps with the still-open #2843 ("avoid bail macros in expression position"), which applies the same fix more broadly. You may want to reconcile the two so the lint fix lands once.

---

phil-opp:
Update: https://github.com/eyre-rs/eyre/pull/294 was merged and a PR to do a new release is up: https://github.com/eyre-rs/eyre/pull/296

phil-opp:
The 18 edited sites are all correct — every one diverges, so `!`-coercion keeps them type-checking and there is no runtime change.

One correction to the body: it states that `cargo +nightly check -p dora-core` has no remaining `trailing semicolon in macro` diagnostics. That is not the case. Two expression-position sites survive in `dora-core`: `libraries/core/src/topics.rs:486` (`Err(VarError::NotUnicode(_)) => eyre::bail!(...)`, a bare match arm) and `libraries/core/src/build/mod.rs:277` (block-tail `eyre::bail!` in the `else` of `validate_subdir`). Both are covered by #2843.

So this PR alone does not unblock the nightly build of `dora-core` — it does so only once #2843 lands too. Worth either correcting the body or merging the two as a pair. (They are otherwise disjoint: no file overlap, so both should land.)

phil-opp:
Thanks — you're right, and I've corrected the body.

Confirmed against the current tree: within `dora-core`, #2843 touches only `topics.rs` and `build/mod.rs`, and this PR touches only `build/git.rs` + the `descriptor/{mod,validate,expand}.rs` sites — **no file overlap**, so the two are complementary. My PR covers the expression-position `bail!` sites in code that merged into `main` after #2843 was opened; `dora-core` is fully nightly-clean only once both land. The title/body now scope the claim to a partial fix rather than a standalone unblock.

(On `rustc 1.99.0-nightly (771916f90 2026-08-08)` I don't reproduce the diagnostic at the two `#2843` sites, whereas #2843 was validated on `da86f4d07 2026-07-24` — the non-local `eyre::bail!` variant seems to be flagged inconsistently across nightly builds. Landing both makes the crate robust regardless of which nightly a contributor runs.)

---

phil-opp:
Heads up — this may no longer be needed. [eyre 0.6.14](https://crates.io/crates/eyre/0.6.14) has been released, and it removes the trailing semicolon from every arm of the `bail!` macro:

```diff
 macro_rules! bail {
     ($msg:literal $(,)?) => {
-        return $crate::private::Err($crate::eyre!($msg));
+        return $crate::private::Err($crate::eyre!($msg))
     };
     …
 }
```

That semicolon was the entire cause of the lint, so both shapes this PR rewrites — block-tail calls and bare `match` arms — are fine as-is under 0.6.14.

I opened #3126, a one-package `Cargo.lock` bump with no source changes. On `rustc 1.99.0-nightly (12c36e253 2026-08-10)`, `cargo +nightly check --all` (excluding the PyO3 crates) drops from **154** `trailing semicolon in macro used in expression position` diagnostics across 17 crates to **0**, which subsumes both this PR's 4 `dora-core` files and #2843's 48.

Worth noting your observation here — that which sites a given nightly flags for the non-local `eyre::bail!` variant varies by toolchain build — held up in testing: on the nightly above the diagnostics surface as future-incompat warnings rather than the hard errors reported in #2835. That variability is a good argument for fixing it at the dependency rather than chasing call sites per toolchain.

Deferring to the maintainers on whether to close this in favour of #3126.

---

PR #3089, opened 2026-08-09
fix(core): regenerate the repo-root `dora-schema.json` (#3088)

### LINKED ISSUE #3088: Root `dora-schema.json` is never regenerated — but it is the URL the guide gives users
The repo has two copies of the dataflow JSON Schema. Only one is maintained, and
the docs point users at the other one.

## Nothing writes the root copy

`generate_schema` writes to `CARGO_MANIFEST_DIR/dora-schema.json`, which for
`dora-core` is `libraries/core/` (`libraries/core/src/bin/generate_schema.rs:39`).
The sync workflow then diffs and commits only `libraries/core/dora-schema.json`
and `libraries/core/dora-node-schema.json`
(`.github/workflows/regenerate-schemas.yml:28` and `:36`). The root
`dora-schema.json` is not produced, updated, or checked by anything in the repo.

## But the root copy is the documented one

`docs/yaml-spec.md:5` and `guide/src/concepts/dataflow-yaml.md:5` both say "A JSON
Schema is available at the repo root (`dora-schema.json`)", and both hand out:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/dora-rs/dora/main/dora-schema.json
```

`libraries/core/README.md:27` points at `libraries/core/dora-schema.json` instead,
so the two documented URLs disagree about which file is authoritative.

## Drift

The root copy last changed in #2512; the maintained copy last changed in #2974.
They now differ by 94 diff lines, and these properties exist in the maintained
schema but are absent from the root one:

`compatibility`, `config_uri`, `exit_when_nodes_finish`, `kind`, `transport`

Descriptions are stale too — e.g. the `health_check_timeout` clarification that
#3030 is currently syncing lands only in the `libraries/core` copy.

## Impact is editor UX, not spurious errors

`generate_schema` rewrites every `additionalProperties` to `true`, and the root
file has 13 such occurrences and zero `false`, so unknown keys are not flagged as
errors. Users following the guide simply get no completion, hover text, or type
checking for anything added since #2512 — `exit_when_nodes_finish` is a top-level
descriptor field and is missing entirely.

## Fix

1. Make the root copy generated: have the workflow copy
   `libraries/core/dora-schema.json` to `dora-schema.json` and add it to the
   `git diff` / `git add` lists.
2. Or delete the root copy and repoint `docs/yaml-spec.md`,
   `guide/src/concepts/dataflow-yaml.md` and the `guide/po/*` translations at the
   `libraries/core` URL.

(1) is the safer default: the root URL is already published in the guide and
likely pasted into users' dataflow YAML, so removing it would 404 their `$schema`
line with no warning.

phil-opp:
Fixes #3088.

The repo has two copies of the descriptor schema, but `generate_schema` wrote
only `libraries/core/dora-schema.json` and the sync workflow diffed and
committed only that one. The root `dora-schema.json` therefore stopped tracking
the descriptor after #2512 — while `docs/yaml-spec.md` and
`guide/src/concepts/dataflow-yaml.md` hand users *its*
`raw.githubusercontent.com` URL as their `# yaml-language-server: $schema=`
line. Editors loading it got no completion or hover for `compatibility`,
`config_uri`, `exit_when_nodes_finish`, `kind` or `transport`.

Kept the root copy rather than deleting it (option 1 in the issue): that URL is
already published in the guide and pasted into users' YAML, so removing it would
404 their `$schema` line with no warning.

- `generate_schema` writes the root copy in the same run. It resolves the root
  as `CARGO_MANIFEST_DIR/../..` but verifies the `[workspace]` marker first —
  the binary ships inside the published `dora-core` crate, where `../..` is a
  registry directory it must not write to.
- `regenerate-schemas.yml` diffs and commits all three generated files from one
  `SCHEMAS` list, so a fourth cannot be added to one place and forgotten in the
  other.
- A unit test next to the generator fails if the two committed copies drift
  again, and a second one pins the registry guard.
- Both copies are regenerated here, so the root one gains the five missing
  properties. That includes the `health_check_timeout` description the open
  auto-sync PR #3030 carries — the two copies have to be identical for the new
  test to hold, and #3030 stays mergeable as a no-op.

### DIFF
--- a/.github/workflows/regenerate-schemas.yml
+++ b/.github/workflows/regenerate-schemas.yml
@@ -25,7 +25,11 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          if git diff --exit-code -- libraries/core/dora-schema.json libraries/core/dora-node-schema.json; then
+          # Every generated file, including the repo-root `dora-schema.json` —
+          # that is the copy the guide publishes as a `$schema` URL, so leaving
+          # it out of these lists is what let it drift (#3088).
+          SCHEMAS="dora-schema.json libraries/core/dora-schema.json libraries/core/dora-node-schema.json"
+          if git diff --exit-code -- $SCHEMAS; then
               echo "Schema file is up to date"
           else
               # Reuse ONE branch + PR: force-push so an already-open sync PR is
@@ -33,7 +37,7 @@ jobs:
               # main (which otherwise piles up dozens of duplicates while the
               # schema stays drifted).
               git checkout -B auto/schema-update
-              git add libraries/core/dora-schema.json libraries/core/dora-node-schema.json
+              git add $SCHEMAS
               git config user.email "dora-bot@phil-opp.com"
               git config user.name "Dora Bot"
               git commit -m "Update JSON schema (${{ github.sha }})"
--- a/dora-schema.json
+++ b/dora-schema.json
@@ -14,6 +14,13 @@
         "$ref": "#/$defs/EnvValue"
       }
     },
+    "exit_when_nodes_finish": {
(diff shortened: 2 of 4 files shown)


phil-opp:

No issues found. The generator now writes the repo-root `dora-schema.json` alongside the `libraries/core` copy, correctly guarded by `workspace_root` so a build from the published crate can't write into a registry directory, and the two new tests (drift check + the package-vs-workspace guard, exercised both ways) actually cover those paths. Looks safe to merge.

---

PR #3093, opened 2026-08-09
fix(ci): keep the Kani proofs running after the MSRV 1.95 bump (#3090)

### LINKED ISSUE #3090: nightly: `Kani formal-verification proofs` broken by the MSRV 1.95 bump — Kani 0.67.0 ships rustc 1.93.0-nightly
nightly: `Kani formal-verification proofs` broken by the MSRV 1.95 bump — Kani 0.67.0 ships rustc 1.93.0-nightly

## Summary

#3017 (`chore(deps): redb 2 -> 4 and sysinfo 0.38 -> 0.39, MSRV 1.88 -> 1.95`,
commit `bd6d178e8`) raised the workspace `rust-version` to 1.95. Kani 0.67.0
bundles its own toolchain — rustc **1.93.0-nightly** — which is older than that,
so `cargo kani` refuses to build `dora-message` before running a single proof.

The `kani-proofs` nightly job has failed every night since, deterministically.

## Failure

```
==> cargo kani -p dora-message
Kani Rust Verifier 0.67.0 (cargo plugin)
error: rustc 1.93.0-nightly is not supported by the following package:
  dora-message@1.0.0-rc.4 requires rustc 1.95.0

error: Failed to execute cargo (exit status: 101). Found 0 compilation errors.
```

Note "Found 0 compilation errors" — nothing was verified. The formal-verification
coverage described in `docs/formal-verification.md` (the `constant_time_eq` auth
proofs) is currently **not running at all**, which is worse than a failing proof
because the job's red state looks like a known-broken CI job rather than lost
coverage.

Observed on nightly runs 08-05 through 08-09 (e.g. run 31299869856, job
93210812005).

## Options

1. **Wait for a Kani release on a ≥1.95 toolchain** and pin the job to it. Kani's
   bundled toolchain moves independently of the release number, so this needs
   checking against the Kani changelog rather than assuming the next patch does it.
2. **Lower `rust-version` for `dora-message` specifically.** It is the only crate
   with `#[kani::proof]` harnesses. If nothing in it actually needs 1.95 —
   the redb 4 / sysinfo requirement is in `coordinator-store` / the CLI, not the
   message crate — a per-crate `rust-version = "1.88"` would let Kani build it
   again while the workspace stays on 1.95. Cheapest fix if it holds.
3. **Skip the job with an explicit reason** until (1) lands. `make qa-kani`
   already skips when Kani is not installed; this would be a different, louder
   skip so the lost coverage is visible rather than silent.

(2) looks best if `dora-message` genuinely does not need 1.95 — worth checking
before reaching for (3).

## Related

- #3017 introduced it.
- #2742 (nightly regression tracker).
- `docs/formal-verification.md`, `make qa-kani`.

---
_Filed from an automated repo watch on 2026-08-09; root-caused from the job log._

phil-opp:
Fixes #3090.

`cargo kani` does not build with the workspace toolchain — it ships its own, and
0.67.0 (still the newest release) bundles `nightly-2025-11-21`, rustc
1.93.0-nightly. Since #3017 raised the workspace `rust-version` to 1.95, cargo
refused to compile `dora-message` under it, so the nightly `kani-proofs` job died
before codegen every night with "Found 0 compilation errors" — the
`constant_time_eq` proofs were not running at all.

`dora-message` now pins `rust-version = "1.88.0"` instead of inheriting the
workspace floor. It is the only crate with `#[kani::proof]` harnesses, and
nothing in it needs 1.95: the sysinfo 0.39 / redb 4 requirement that raised the
floor belongs to the CLI, daemon, and coordinator store, and this crate's whole
dependency closure declares at most 1.85. Kani exposes no way to forward
`--ignore-rust-version` to cargo, and cargo has no config key or env var for it,
so a per-crate pin is the only fix that keeps verification switched on.

Option 1 from the issue (wait for a Kani release on a ≥1.95 toolchain) has
nothing to pin to yet: 0.67.0 from January is still the newest release. Kani
`main` has since moved to `nightly-2026-02-18` (rustc 1.95.0-nightly), which
would satisfy the current floor — cargo compares release numbers and ignores the
`-nightly` pre-release tag — but no release carries it.

The pin is enforced, not asserted: the nightly `msrv` job runs `cargo hack check
--rust-version`, which checks `dora-message` at 1.88 and fails if the crate stops
building there, while `kani-proofs` fails if the pin ever climbs past Kani's
rustc. Both sites that spring the trap now say so — a comment on the workspace
`rust-version`, and a section in `docs/formal-verification.md` covering why proof
crates must stay below Kani's bundled toolchain and what to do when adding a new
one.

Verified with the same Kani build CI installs (0.67.0 / `nightly-2025-11-21`):
`scripts/qa/kani.sh` reproduces the failure on `main` and now reports
`VERIFICATION:- SUCCESSFUL` — 133 checks, 0 failed. `cargo hack check
--rust-version --workspace --ignore-private --locked` passes all 31 packages,
checking `dora-message` at 1.88. `cargo update --dry-run` resolves identically
with and without the pin, so the lower floor holds no dependency back.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,9 @@ members = [
 
 [workspace.package]
 edition = "2024"
+# Raising this does not raise `dora-message`, which pins its own lower
+# `rust-version` so Kani's bundled toolchain can still build the proof
+# harnesses (#3090). Bump that one only against Kani's rustc.
 rust-version = "1.95.0"
 # Python packages derive version from CARGO_PKG_VERSION automatically.
 version = "1.0.0-rc.4"
--- a/docs/formal-verification.md
+++ b/docs/formal-verification.md
@@ -92,6 +92,30 @@ nightly workflow runs the proofs daily (`kani-proofs` job in
 `scripts/qa/ci-nightly-jobs.sh kani-proofs`), so a broken harness is
 caught within a day and auto-filed as a `nightly-regression` issue.
 
+### Proof crates must stay below Kani's bundled toolchain
+
+Kani does not build with the workspace toolchain — it ships its own
+nightly (0.67.0 → `nightly-2025-11-21`, rustc 1.93.0-nightly). Cargo
+refuses to compile a package whose `rust-version` exceeds the running
+rustc, so **a proof crate whose MSRV outruns Kani's rustc stops being
+verified at all**: `cargo kani` fails before codegen, reporting "Found 0
+compilation errors" rather than a failed proof. That is how #3090
+happened — the workspace MSRV went to 1.95 in #3017 and every nightly
+run after it was red with zero proofs run.
+
+So `dora-message` pins its own `rust-version` (currently 1.88.0) instead
+of inheriting `[workspace.package]`, with a comment in its `Cargo.toml`
+saying why. Any crate that gains a `#[kani::proof]` harness needs the
+same treatment. Two guards keep the pin honest: the nightly `msrv` job
+runs `cargo hack check --rust-version`, which fails if the crate no
+longer builds on the version it advertises, and the `kani-proofs` job
+fails if the pin ever climbs past Kani's rustc.
+
+Cargo compares only the release numbers here, ignoring the pre-release
+tag, so `rust-version = "1.93.0"` would be accepted by rustc
+1.93.0-nightly. The pin sits a few releases lower anyway, to leave room
+for a workspace MSRV bump landing before the matching Kani release.
+
 ## Adding a new proof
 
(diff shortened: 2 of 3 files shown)


phil-opp:

No issues found. Pinning `dora-message` to `rust-version = "1.88.0"` (below the workspace 1.95 floor) is the right fix to keep `cargo kani` — which builds with its own 1.93.0-nightly toolchain — from bailing before codegen, and the nightly `msrv`/`kani-proofs` jobs enforce the pin from both sides. The lower floor holds no dependency back.

---

PR #3095, opened 2026-08-09
fix(runtime): stop freeing operator-owned payloads on the runtime thread (#2742)

phil-opp:
Fixes the last surviving `cli-tests-python (windows)` failure in #2742.

## The bug

`webcam.py` builds its payload with `pa.array(frame.ravel())`. That is **zero-copy over numpy** — the Arrow buffer *is* the numpy allocation — so releasing it acquires the GIL (pyarrow's `NumPyBuffer` destructor does `PyAcquireGIL`). Verified locally against the same pyarrow 25.0.0 / numpy 2.5.1 the nightly installs: arrow buffer address == numpy data pointer, `pa.total_allocated_bytes()` delta 0.

That array was handed to the runtime's event loop, which dropped it on a tokio blocking thread holding no GIL. Measured with a pyo3 + arrow probe: **15.6 µs** to drop with the GIL free, **4.5 s** while another thread held it.

The GIL holder in the nightly is `webcam.py`'s `__del__` → `cv2.VideoCapture.release()`, run under `Python::attach` on the STOP tick. That is exactly the tick every observed wedge started on (`t=12.21s` on 08-09, `t=13.54s` on 07-29, both past `CI_HEADLESS_RUN_SECS=10`). The loop then observed neither `Stop` nor the operator's own `Finished`, and the daemon force-killed the node — `ExitCode(1)` on Windows, masked as `Signal(15)` on Unix.

The `#2885` hop diagnostics pointed at `report_output_sent` because that is the last instrumented hop; the stall is in the un-instrumented array drop immediately after it.

## The fix

**Commit 1 — operators encode their own outputs.** `OperatorEvent::Output` now carries an `EncodedSample` instead of an `ArrayData`, so foreign-owned memory structurally cannot cross the channel. `RuntimeHandle::send_output` is the single place the event is built; it borrows the array, so both backends drop it on their own thread — for Python, with the GIL held, where the release is free.

No extra payload copy: `encode_arrow` is the same `encode_ipc_into` the node already ran in `send_output_array`, moved to the other side of the channel. The runtime path actually loses a `make_array` + `to_data()` per output. The genuinely zero-copy `send_output_raw` path is untouched.

**Commit 2 — the loop stays responsive during a send.** Even with the above, any slow send (zenoh, daemon socket) made the node deaf to `Stop`. `await_send_watching_for_stop` keeps consuming the event stream: `Stop` is forwarded to the operators immediately, everything else is buffered and handled in order once the send completes.

It deliberately stops there. An earlier draft also abandoned a wedged send and returned early — that turned out to be inert: the `DoraNode` lives inside the blocking task and `Runtime::drop` "waits forever" for `spawn_blocking` work, so returning early just moves the wedge one frame up into the runtime's own shutdown. A real escape needs `shutdown_background()` at the `main()` layer; left as a follow-up, since after commit 1 the wedge this issue is about is gone at the source.

## Trade-offs

- The IPC encode now runs inside the operator's `send_output` rather than after the channel hand-off. With channel capacity 1 there was little pipelining to lose, but it is on the critical path of user callbacks now (~100 µs for a 1 MB frame).
- An SHM sample is allocated earlier and lives across the channel hop, so a producer can hold two chunks instead of one — ~12.5% more pressure on the default 8 MB pool at 1 MB payloads. Worth a throughput comparison before/after on a large-payload dataflow.
- The loop now pulls up to 4 events off the stream during a send instead of leaving them in the daemon channel, so a few input payloads are pinned slightly longer.
- `close_outputs` on the `Finished` path still awaits its `spawn_blocking` inline; the same treatment would apply there, but it is outside what this change needed.

## Not included

- The `#2857`/`#2885` diagnostics (`webcam.py` prints, the stall watchdog, the hop logs) are left in place; removing them belongs in a follow-up once a nightly confirms the fix.
- `webcam.py` still opens a `cv2.VideoCapture` when headless and only releases it in `__del__`. Worth fixing on its own, but it is no longer load-bearing.
- `ipc_encode::prepare()` runs twice per message (once for the length, once for the encode) — pre-existing, and `encode_arrow` is now the natural place to fix it.

Refs #2742

### DIFF
--- a/apis/rust/node/src/lib.rs
+++ b/apis/rust/node/src/lib.rs
@@ -113,7 +113,8 @@ pub use futures;
 #[cfg(feature = "tracing")]
 pub use node::init_tracing;
 pub use node::{
-    DataSample, DoraNode, DoraNodeBuilder, StreamSegment, ZERO_COPY_THRESHOLD, arrow_utils,
+    DataSample, DoraNode, DoraNodeBuilder, EncodedSample, SampleAllocator, StreamSegment,
+    ZERO_COPY_THRESHOLD, arrow_utils,
 };
 pub use uuid;
 
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -724,7 +724,10 @@ pub struct DoraNode {
     /// `None` in interactive/testing mode.
     zenoh_session: Option<zenoh::Session>,
     /// Zenoh shared memory provider for zero-copy publishing.
-    zenoh_shm_provider: Option<zenoh::shm::ShmProvider<zenoh::shm::PosixShmProviderBackend>>,
+    /// Owns the SHM provider and the zero-copy threshold. Cloned out by
+    /// [`DoraNode::sample_allocator`] so an operator thread can build output
+    /// samples without reaching into the node (dora-rs/dora#2742).
+    sample_allocator: SampleAllocator,
     /// Per-output zenoh publishers with their handshake state, declared eagerly
     /// at init (see [`declare_output_publishers`]) so zenoh wires routes
     /// immediately and [`StartupHandshake`] can probe them before the first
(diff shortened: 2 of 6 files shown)


phil-opp:

I reviewed this diff and found no issues. The ownership invariant that fixes #2742 holds where it matters — `RuntimeHandle::send_output` only borrows the array, encodes into a dora-owned sample, and the operator thread drops the source payload itself — and `OperatorEvent::Output` now carrying `EncodedSample` instead of `ArrayData` enforces it at the type level. `await_send_watching_for_stop` preserves event ordering, bounds buffering, and forwards `Stop` without losing any of the old main-loop handling. The added tests are substantive (foreign-buffer release via a custom allocation, IPC round-trip, bounded/ordered buffering, mid-send stop delivery).

---

PR #3097, opened 2026-08-09
fix(qa): resolve the cargo target dir instead of hardcoding `target/`

phil-opp:
The QA and smoke scripts locate prebuilt binaries at `<repo>/target/...`, so they break whenever the target dir moves — a `CARGO_TARGET_DIR` override, or a `build.target-dir` in a `.cargo/config.toml`. The motivating case is one target dir shared across git worktrees, so third-party deps are compiled once instead of once per worktree.

Symptoms were `dora start` failing with `Could not find source path ../../target/debug/<bin>`, and `dora replay` silently falling through to `cargo install` from crates.io.

Adds `scripts/cargo-target-dir.sh`, which asks `cargo metadata` for `target_directory` — authoritative, since it already accounts for `CARGO_TARGET_DIR` and every config source cargo honors. It falls back to `<root>/target` only when cargo cannot answer at all (cargo or python3 missing, no reachable manifest), which reproduces the previous behavior exactly.

Converts all 29 hardcoded sites across `smoke-all.sh`, `qa/all.sh`, `qa/pgo.sh`, `qa/ci-nightly-jobs.sh` and `ros2-zenoh-interop.sh`. Two heredocs that generate dataflow descriptors are unquoted so the path interpolates, and two `REPO_ROOT` locals that became dead are dropped.

Verified with `./scripts/smoke-all.sh --rust-only` against a relocated target dir, plus `make qa-fast` (fmt/clippy/audit/unwrap-budget/typos all pass). The helper's four resolution paths — config-based, `CARGO_TARGET_DIR` override, no reachable manifest, no `python3` — were each exercised directly.

Two things deliberately left out of scope:

- `dora_cli::command::node_binary::search` honors `CARGO_TARGET_DIR` but not `build.target-dir`, so `dora replay` can still miss `dora-replay-node` under a relocated target dir. That is production code in `dora-cli` and wants its own change with a test.
- 42 example dataflow descriptors hardcode `path: ../../target/debug/<bin>`, which dora resolves relative to the descriptor. Those relative paths are what users copy into their own projects, so they should stay; a `target` symlink is the right local workaround.

### DIFF
--- a/scripts/cargo-target-dir.sh
+++ b/scripts/cargo-target-dir.sh
@@ -1,0 +1,31 @@
+# shellcheck shell=bash
+#
+# Resolve the cargo target directory for the workspace.
+#
+# The QA and smoke scripts locate prebuilt binaries by path, so hardcoding
+# `<repo>/target` breaks them whenever the target dir moves -- a
+# CARGO_TARGET_DIR override, or a `build.target-dir` / `build.build-dir` in
+# some .cargo/config.toml (e.g. one target dir shared across git worktrees, so
+# that deps are not rebuilt per worktree).
+#
+# `cargo metadata` is authoritative: it already accounts for CARGO_TARGET_DIR
+# and every config source cargo itself honors. We only fall back to
+# `<root>/target` when cargo cannot answer at all (cargo or python3 missing, no
+# manifest reachable), which reproduces the previous hardcoded behavior.
+#
+# Usage:
+#   source "$SCRIPT_DIR/cargo-target-dir.sh"
+#   TARGET_DIR="$(cargo_target_dir "$ROOT")"
+cargo_target_dir() {
+    local root="${1:-$PWD}"
+    local dir
+
+    # A failure anywhere in this chain -- no cargo, no python3, no reachable
+    # manifest -- leaves $dir empty, and the fallback below takes over.
+    dir="$(cd "$root" 2> /dev/null \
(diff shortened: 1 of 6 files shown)


phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review with no human in the loop; please verify before acting.

I reviewed this diff. It centralizes cargo-target-dir resolution into `scripts/cargo-target-dir.sh` and threads `$TARGET_DIR` through the QA/smoke scripts in place of a hardcoded `target/`.

- `cargo_target_dir` falls back to `${dir:-${CARGO_TARGET_DIR:-$root/target}}`, so when `cargo metadata` can't answer (no cargo/python3/manifest) it reproduces the previous hardcoded behavior.
- The two heredocs in `scripts/qa/ci-nightly-jobs.sh` that write `redb-smoke.yml` and `long-running.yml` were switched from quoted (`<<'EOF'`) to unquoted (`<<EOF`). I checked both bodies in full — each is 7 lines of YAML whose only `$` is the interpolated `$TARGET_DIR` path, with no other `$`, backticks, or command substitutions, so the unquoting expands only what's intended.
- Those two fixtures move from a relative `../../target/debug/...` node path to an absolute `$TARGET_DIR/debug/...`; since `cargo metadata` returns an absolute `target_directory` (and the fallback is `$root/target` from the repo root), this resolves to the same location as before.

No issues found.

---

PR #3098, opened 2026-08-09
fix(mavlink2-bridge): honor `?proto=` query on the serial transport

phil-opp:
## Issue

The MAVLink transport builder (`libraries/extensions/mavlink2-bridge/src/transport.rs`) accepts an optional `?proto=` query parameter to select the MAVLink wire version. `connect_tcp` and `connect_udp` both read it via `parse_proto_query(url).unwrap_or(MavlinkVersion::V2)`, but `connect_serial` routed through the `open_mavlink` wrapper, which **hardcoded `MavlinkVersion::V2`**:

```rust
fn connect_serial(url: &Url) -> BridgeResult<...> {
    ...
    open_mavlink(&format!("serial:{device}:{baud}"))   // always V2
}

fn open_mavlink(addr: &str) -> BridgeResult<...> {
    open_mavlink_versioned(addr, MavlinkVersion::V2)   // proto ignored
}
```

So a user connecting with `serial:///dev/ttyUSB0?baud=57600&proto=v1` silently got a MAVLink **V2** connection — no error, no warning — while the exact same override works on `tcp://` and `udp://`. The `proto` parameter was also completely undocumented (the module docs cover `baud` but never mention `proto`).

## Fix

- Route `connect_serial` through `parse_proto_query(url).unwrap_or(MavlinkVersion::V2)` and `open_mavlink_versioned`, exactly like the tcp/udp paths.
- Delete the now-dead `open_mavlink` wrapper (its only caller was `connect_serial`).
- Document the `proto` query parameter (accepted spellings + default) in the module docs.

No behavior change for tcp/udp or for serial connections that omit `proto`; the default remains MAVLink V2.

## Validation

- Added unit tests: `parses_proto_query_variants`, `parses_proto_query_missing_or_garbage`, and `serial_url_honors_proto_query` (regression guard for the serial branch).
- `cargo +1.97.1 fmt -p dora-mavlink2-bridge -- --check` — clean.
- `cargo +1.97.1 clippy -p dora-mavlink2-bridge -- -D warnings` — clean.
- `cargo +1.97.1 test -p dora-mavlink2-bridge` — all pass (incl. doctests).

---

---

### DIFF
--- a/libraries/extensions/mavlink2-bridge/src/transport.rs
+++ b/libraries/extensions/mavlink2-bridge/src/transport.rs
@@ -10,6 +10,15 @@
 //! | `serial:///dev/path?baud=N` (Unix)    | `serial:/dev/path:N`         | n/a       |
 //! | `serial://COM1?baud=N` (Windows)      | `serial:COM1:N`              | n/a       |
 //!
+//! ## Protocol version
+//!
+//! All three schemes accept an optional `?proto=` query parameter to
+//! select the MAVLink wire version: `proto=v1` / `proto=1` / `proto=1.0`
+//! for MAVLink 1, and `proto=v2` / `proto=2` / `proto=2.0` for MAVLink 2.
+//! When omitted (or unrecognized) the connection defaults to MAVLink 2.
+//! It combines with the other query parameters, e.g.
+//! `serial:///dev/ttyUSB0?baud=57600&proto=v1`.
+//!
 //! ## Known limitations (intentional, not bugs)
 //!
 //! * **`tcp://` is always `tcpout:` (client mode).** There is no way
@@ -85,17 +94,14 @@ fn connect_serial(url: &Url) -> BridgeResult<Box<dyn MavConnection<MavMessage> +
         )));
     }
     let baud = parse_baud(url).unwrap_or(DEFAULT_SERIAL_BAUD);
-    open_mavlink(&format!("serial:{device}:{baud}"))
+    let version = parse_proto_query(url).unwrap_or(MavlinkVersion::V2);
+    open_mavlink_versioned(&format!("serial:{device}:{baud}"), version)
 }
 
 fn parse_baud(url: &Url) -> Option<u32> {
     url.query_pairs()
         .find(|(k, _)| k == "baud")
         .and_then(|(_, v)| v.parse::<u32>().ok())
-}
-
-fn open_mavlink(addr: &str) -> BridgeResult<Box<dyn MavConnection<MavMessage> + Send + Sync>> {
-    open_mavlink_versioned(addr, MavlinkVersion::V2)
 }
 
 fn open_mavlink_versioned(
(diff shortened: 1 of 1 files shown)


phil-opp:

This is the summary for a batch of machine-generated PRs opened by Claude (Claude Code) from an in-depth review of the repo against a freshly-fetched `origin/main`. Each PR is standalone, branches off `origin/main`, and is clearly attributed + marked machine-generated. Per-branch verification: `fmt` + `clippy -D warnings` + affected-crate tests (toolchain 1.97.1). The combined diff additionally passed the full workspace test suite (`cargo test --all` minus the Python/examples excludes) and the `/review` + `/simplify` gates.

### PRs by category

**Correctness**
- **#3098** — `fix(mavlink2-bridge)`: honor the `?proto=` query on the **serial** transport (tcp/udp already did; serial silently forced V2). *(this PR)*
- **#3100** — `fix(cli)`: `dora up` now rejects a malformed `DORA_COORDINATOR_ADDR`/`PORT` instead of silently falling back to the default (diverged from every other command's clap `env=`).
- **#3101** — `fix(core)`: `${_param.…}` substitution is now single-pass and order-independent (was `BTreeMap`-order-dependent and transitively expanded param values).
- **#3103** — `fix(node)`: record zenoh-delivered inputs in the `write_events_to` log (they surface as `Event::Input` but were dropped from recordings). Includes a timestamp-underflow guard surfaced by the `/review` gate (a remote HLC timestamp predating node start would panic/wrap `get_diff_duration`).

**Documentation**
- **#3099** — `docs(cli)`: attach the `--env` help text to the right field in `dora start` (it was mis-attributed to `--exit-when-nodes-finish`, leaving `--env` undocumented).

**Optimization**
- **#3102** — `perf(daemon)`: drop the per-log-line `String` reallocation in the log task (hot path: one alloc+copy per log line per node).

### Rejected / not-pursued findings (one-line reasons)
- **`InputMapping::Logs` `Display` drops `node_filter` when `min_level` is `None`** — unreachable: `FromStr` never constructs that state and no other construction site exists, so no demonstrable failure.
- **`dora logs --follow` degrades to a one-shot dump when the coordinator is unreachable** — real but niche; deferred (per-run PR cap).
- **`restart.rs` request/reply duplication** — valid simplification (`send_control_request`+`expect_reply!`); deferred (per-run PR cap).
- **`core::resolve_path` missing doctest** — genuine doc gap, but a deterministic doctest depends on `uv`/`$PATH` availability; deferred.
- **`daemon::truncate_log_line` off-by-one at an exact 1 MB line** — cosmetic and only at the exact boundary; too weak.
- **`message::auth::generate_token` per-byte `format!`** — cold startup path; too weak to justify a PR.
- **`Metadata::open_telemetry_context` always allocates** — fix is an API-signature change needing a call-site audit; minor.
- **`descriptor.rs` `pattern_metadata_keys()` doc reference** — the function does exist (in `dora-core`, cross-crate); weak.
- **`event_stream` `Stream::poll_next` bypasses recording** — plausibly intentional (recording may be scheduler-path-only); medium-confidence, not pursued.
- **Dead `coordinator_addr` branch in `open_zenoh_session_with_listen`** — effectively dead (all in-repo callers pass `None`); removing it touches a `pub` signature; deferred.
- **mavlink `MavlinkArrow::schema()` rebuilt (with `format!`) per message** — real per-frame optimization, but the clean fix needs a trait-signature change (`SchemaRef`); larger scope, deferred.
- **Altitude (`/simplify`): flatten `CoordinatorOptions` into `dora up`** — the deeper fix for #3100, but it adds new `--coordinator-addr/--coordinator-port` flags (a CLI-surface change beyond the bug fix); skipped as out-of-scope, noted here as a follow-up. The in-scope `/simplify` cleanup (fold the env read into the helper) was applied to #3100.

---

phil-opp:

No blocking issues found. The fix is correct: `connect_serial` now reads `parse_proto_query(url).unwrap_or(MavlinkVersion::V2)` and routes through `open_mavlink_versioned`, matching the tcp/udp paths exactly; the dead `open_mavlink` wrapper is removed along with its only caller. Behavior is unchanged when `proto` is omitted (defaults to V2).

Two smaller notes, neither blocking:
- The `connect()` rustdoc still asserts "The returned connection is pre-configured for MAVLink V2," which is now only the default rather than a guarantee — worth softening now that all three schemes can select V1.
- `serial_url_honors_proto_query` exercises `parse_baud`/`parse_proto_query` on the serial URL rather than `connect_serial` itself (which needs real hardware), so it guards the parsing but not the wiring into `open_mavlink_versioned`. Acceptable given the constraint, and it does prevent a regression to the hardcoded-V2 parse.

---

phil-opp:
Superseded: main already honors `?proto=` uniformly across all three transports (including serial) via `parse_proto_query`, and rejects an unrecognized value outright instead of falling back to V2 — strictly stricter than this PR. Rebasing onto main leaves zero commits.

PR #3100, opened 2026-08-09
fix(cli): reject a malformed coordinator address/port in `dora up`

phil-opp:
## Issue

`dora up` resolved the coordinator location from `DORA_COORDINATOR_ADDR` / `DORA_COORDINATOR_PORT` like this (`binaries/cli/src/command/up.rs`):

```rust
let addr: std::net::IpAddr = std::env::var("DORA_COORDINATOR_ADDR")
    .ok().and_then(|s| s.parse().ok()).unwrap_or(LOCALHOST);
let port: u16 = std::env::var("DORA_COORDINATOR_PORT")
    .ok().and_then(|s| s.parse().ok()).unwrap_or(DORA_COORDINATOR_PORT_WS_DEFAULT);
```

`.and_then(|s| s.parse().ok())` throws away a parse **error**, so a present-but-unparseable value silently falls back to the default. Every *other* lifecycle command reads the same two env vars through clap's typed `env=` on `CoordinatorOptions` (`common.rs`), which **errors** on an unparseable value.

The result is a silent divergence: `DORA_COORDINATOR_PORT=abc dora up` starts a coordinator on the **default** port `53290`, while `dora stop` / `dora list` / `dora down` abort with `invalid value for '--coordinator-port'` — so they can no longer find the coordinator that `up` just started.

## Fix

Add a small `coordinator_env_value` helper that returns the default **only** when the variable is unset, and returns an error when it is set but unparseable — mirroring the strictness of `CoordinatorOptions`. Route both reads through it. Behavior is unchanged for the unset case and for valid values.

## Validation

- Added unit tests (`coordinator_env_tests`): unset → default, valid → parsed, malformed → error (not defaulted).
- `cargo +1.97.1 fmt -p dora-cli -- --check` — clean.
- `cargo +1.97.1 clippy -p dora-cli -- -D warnings` — clean.
- `cargo +1.97.1 test -p dora-cli coordinator_env` — 3 passed.

---

---

### DIFF
--- a/binaries/cli/src/command/up.rs
+++ b/binaries/cli/src/command/up.rs
@@ -49,14 +49,21 @@ struct UpConfig {}
 
 pub(crate) fn up(config_path: Option<&Path>, auth: bool, recreate_store: bool) -> eyre::Result<()> {
     let UpConfig {} = parse_dora_config(config_path)?;
-    let addr: std::net::IpAddr = std::env::var("DORA_COORDINATOR_ADDR")
-        .ok()
-        .and_then(|s| s.parse().ok())
-        .unwrap_or(LOCALHOST);
-    let port: u16 = std::env::var("DORA_COORDINATOR_PORT")
-        .ok()
-        .and_then(|s| s.parse().ok())
-        .unwrap_or(DORA_COORDINATOR_PORT_WS_DEFAULT);
+    // Surface a malformed value instead of silently falling back to the
+    // default: every other lifecycle command reads these env vars through
+    // clap's typed `env=` on `CoordinatorOptions`, which errors on an
+    // unparseable value. If `dora up` quietly ignored a typo it would bind a
+    // different port than `dora stop`/`list`/`down` then look for.
+    let addr: std::net::IpAddr = coordinator_env_value(
+        "DORA_COORDINATOR_ADDR",
+        std::env::var("DORA_COORDINATOR_ADDR").ok().as_deref(),
+        LOCALHOST,
+    )?;
+    let port: u16 = coordinator_env_value(
+        "DORA_COORDINATOR_PORT",
+        std::env::var("DORA_COORDINATOR_PORT").ok().as_deref(),
+        DORA_COORDINATOR_PORT_WS_DEFAULT,
+    )?;
     let coordinator_addr = (addr, port).into();
     if recreate_store && !addr.is_loopback() {
         bail!(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The change replaces the silent `.and_then(|s| s.parse().ok()).unwrap_or(default)` fallback with a helper that returns the default only when the var is unset and errors on a present-but-unparseable value, matching the strictness of clap's typed `env=` on `CoordinatorOptions`. The `parse_coordinator_env` seam is a clean way to unit-test without touching the process environment, and the three tests (unset / valid / malformed) exercise meaningful behavior rather than being tautological.

One minor behavioral note, not a blocker: an env var set to an empty string (e.g. `DORA_COORDINATOR_PORT=`) now errors where it previously fell back to the default, since `"".parse::<u16>()`/`IpAddr` fails. That is consistent with the stated goal of agreeing with the other lifecycle commands (which route through clap and would also reject empty), so it looks intentional.

---

phil-opp:
The merge-queue failure on this PR is **not** caused by this change, and no fix is needed here.

The failing job (`Test (ubuntu-latest)`) failed on a doctest in a different crate:

```
apis/rust/node/src/node/arrow_utils/ipc_encode.rs - PreparedIpc
error[E0433]: cannot find `arrow` in `dora_node_api`
  use dora_node_api::arrow::array::{Array, UInt64Array};
```

This PR's diff is confined to `binaries/cli/src/command/up.rs` (env-var parsing) and touches nothing in `dora-node-api`, so it cannot affect that doctest. It's a batch interaction: the failed batch reported the doctest at line 360, but on current `origin/main` it lives at line 359 — i.e. the batch also contained another PR that modified `ipc_encode.rs`. I ran that doctest against current `origin/main` and it passes (`pub use arrow;` is present at `apis/rust/node/src/lib.rs:89`), so `main` is green.

The bisection should exonerate this PR; it's safe to re-queue.

---

phil-opp:
Follow-up: this PR passed the bisection alone (SHA `137a37a`), re-entered the queue, and then failed **again** on the next batch (SHA `61d591d`) with the *same* unrelated doctest error:

```
apis/rust/node/src/node/arrow_utils/ipc_encode.rs - PreparedIpc (line 360)
error[E0433]: cannot find `arrow` in `dora_node_api`
```

Note the doctest is at **line 360** in the failing batch but **line 359** on `origin/main` — so a batch-mate PR that modifies `ipc_encode.rs` (and is out of sync with `main`'s `pub use arrow;` at `lib.rs:89`) is poisoning every batch this PR lands in. Re-queuing #3100 on its own won't help; the offending batch-mate needs to be dropped or rebased first. This PR's diff is `up.rs`-only and is green against `main` — nothing to change here.

---

PR #3102, opened 2026-08-09
perf(daemon): drop the per-log-line String reallocation in the log task

phil-opp:
## Issue

The daemon's per-node file-logging task ran this for **every** log line of **every** spawned node (`binaries/daemon/src/spawn/prepared.rs`):

```rust
let formatted = content.lines().fold(String::default(), |mut output, line| {
    output.push_str(line);
    output
});
```

`content` is an owned `String` moved out of the `LogLine` and not used afterward. It holds a single logical line — `read_capped_line` stops at the first `\n` — so this fold does nothing but strip the trailing newline, yet it heap-allocates a brand-new `String` and copies the entire line byte-for-byte. On a node logging at high frequency that's an allocation + full copy per line, on the log hot path, for no benefit.

## Fix

Introduce a small `strip_trailing_newline(String) -> String` helper that pops a trailing `\n` (and a preceding `\r`) from the already-owned buffer in place, and use it instead of the fold. No allocation, no copy.

For the single-line inputs the reader produces, the result is byte-for-byte identical to `content.lines().next()`.

## Validation

- Added `strip_trailing_newline_matches_lines_for_single_line`, which diffs the helper against the **old** `.lines().fold(...)` implementation across `\n`, `\r\n`, bare `\r`, empty, mid-line `\r`, trailing-spaces, and truncated-marker cases.
- `cargo +1.97.1 fmt -p dora-daemon -- --check` — clean.
- `cargo +1.97.1 clippy -p dora-daemon -- -D warnings` — clean.
- `cargo +1.97.1 test -p dora-daemon strip_trailing` — passes.

---

---

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -88,6 +88,20 @@ fn truncate_log_line(content: &mut String) {
     }
 }
 
+/// Drop a single trailing line terminator (`\n`, or `\r\n`) from an owned
+/// log line, in place. `content` holds one logical line (the reader stops at
+/// the first `\n`), so this reproduces what `content.lines().next()` would
+/// yield without allocating a new `String`.
+fn strip_trailing_newline(mut content: String) -> String {
+    if content.ends_with('\n') {
+        content.pop();
+        if content.ends_with('\r') {
+            content.pop();
+        }
+    }
+    content
+}
+
 /// Read a single log line from `reader`, buffering at most
 /// `MAX_LOG_LINE_BYTES + 1` bytes into `raw`. Returns `Ok(true)` once the
 /// stream reaches EOF (nothing more to read).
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

I verified the optimization against the code. `read_capped_line` reads with `read_until(b'\n', raw)` (`binaries/daemon/src/spawn/prepared.rs`), so `content` is a single logical line holding at most one trailing `\n` (and possibly a preceding `\r`), with no embedded newlines. Under that invariant the old `content.lines().fold(...)` and the new `strip_trailing_newline` are byte-for-byte identical:

- The fold only concatenated lines and dropped terminators; with a single line it merely stripped the trailing `\n`/`\r\n`.
- A lone `\r` (truncated line) and no-newline input are preserved by both.
- Empty, `"\n"`, and `"\r\n"` all collapse to `""` in both.

`content` is moved out of `LogLine` at the use site and not read afterward, so consuming it in place is safe. The differential unit test (`strip_trailing_newline_matches_lines_for_single_line`) checks the helper against the old fold across representative cases, which is a meaningful guard rather than a tautology.

---

PR #3079, opened 2026-08-08
feat(memory-pool): cross-machine zero-copy tensor transport for end-edge-cloud collaboration

tang-canran:
### (1) Design — new architecture for multi-daemon and cross-machine deployment

This PR extends the memory-pool transport to the deployment topologies real edge-cloud robotics actually run: **multiple daemons on one host** (sensor, perception, and inference pipelines as separate daemons on a robot or edge box) and **daemon clusters across machines** (edge ↔ cloud). The additions:

**Same-host multi-daemon — direct read across daemons**
- The receiver locates the sender's pool segment **directly by machine-qualified name** (`pool_{node_id}_{machine}_{counter}`) and reads it in place, bypassing the daemon relay and even the mirror push entirely (`same-host direct detection`). Native dora has no such path: outputs to a consumer on another daemon are pinned to the daemon relay (#2738), so the same topology costs 89 MB/s instead of ~5.8 GB/s (65×).
- For GPU pools, the sender exports the CUDA IPC handle into the pool header at registration; a same-host cross-daemon receiver imports it once and reads device memory zero-copy (32 GB/s measured).
- Control-plane notifications (RegisterPool/RegisterPoolAck/FreePool) between daemons ride a zenoh SHM payload for same-host delivery.

**Cross-machine — reliable large-frame data plane with GPU staging**
- The data plane uses the daemon's zenoh channel with **blocking congestion control**: large frames queue and drain over slow WAN links instead of being dropped. Native dora's relay uses Drop + express: fragments are silently dropped when the TX queue (16-batch cap) backs up on a slow link, and the daemon↔daemon data plane hangs at *any* frame size once the RTT crosses ~200 ms — reproducible on loopback with `tc qdisc add dev lo root netem delay 100ms` (a 0.5 MiB frame hangs identically to 40 MiB). There is **no working native path for WAN large frames**.
- GPU endpoints get **CPU staging pools**: a pageable transit pool on the send side (GPU_A → DtoH → CPU_A) and a pinned pool on the receive side (CPU_B → HtoD → GPU_B), both registered and reused per frame; the full path GPU→CPU→zenoh TCP→CPU→GPU runs at link bandwidth (~38 MB/s on the 1-Gbps LAN link; GPU staging adds no per-frame allocation).
[...]

PR #3106, opened 2026-08-10
feat(node-cxx): memory-pool transport with a CUDA-IPC-free path for integrated GPUs

caothu2k1:
Closes the remaining half of #2686. The Service/Action half landed in #2853.

### The problem

`apis/python/node/src/lib.rs:2308` fails registration when a CUDA receiver's IPC
export fails:

```rust
if receiver_is_cuda && !ipc_written {
    eyre::bail!("register_memory_pool: failed to set up GPU pool buffer / IPC handle …");
}
```

On an integrated GPU `cudaIpcGetMemHandle` is unsupported, so that branch always
fires. Today **no language binding can use a CUDA receiver on such a device** —
the memory-pool transport is unavailable there entirely, not merely slow.

### What this adds

**A third transport, `unified`.** Full shared-memory data region, `ipc_flag = 0`;
the receiver takes a device alias of the same pages with
`cudaHostRegister(..., cudaHostRegisterMapped)` + `cudaHostGetDevicePointer`. It
needs no protocol change because it reuses the existing segment layout, and it
works on every platform — on a discrete GPU it is merely slower than `ipc`.
`auto` resolves to `unified` when the receiver is CUDA and `shmem` otherwise, so
it can never resolve to a mode that fails at registration.
`DORA_MEMORY_POOL_TRANSPORT` overrides `auto` and nothing else.

**The `DORADMA` data plane gains a definition in `dora-memory-pool`.** The header
layout, the seqlock and the segment lifecycle previously existed only as
hand-written byte offsets inside the Python binding. This adds a tested
implementation the C++ binding builds on — it does **not** consolidate: the
Python binding still carries its own hand-written copy, and making it the single
definition is follow-up work.

**The C++ binding gains the pool surface**: `register_memory_pool`,
`write_memory_pool`, `read_memory_pool`, `free_memory_pool`, plus the pointers a
producer needs to fill a pool in place instead of paying a second copy.

**Two optional header-only C++ helpers.** `dora/memory_pool.hpp` provides
`PoolWriteGuard` and `PoolReadGuard`; `dora/cuda_pool.hpp` provides the mapping
helper. No Rust crate links CUDA — the language side owns it, matching how the
Python binding already works.

### Design notes a reviewer may want

**Invariants live in the types, not in doc comments.** FFI is where Rust's
enforcement stops, so several review rounds moved guarantees across that line:

- Payload access is `pool_payload(&mut ptr, &mut len) -> bool`, an out-param
  predicate returning false for an `ipc` pool. An earlier shape exported a
  pointer and a length that did not bound each other: on an `ipc` pool the
  offset points one byte past the mapping while the declared size reports
  megabytes, so `memcpy(host_ptr(), src, size())` — the natural C++ line — wrote
  megabytes out of bounds. There is no declared-size accessor on the C++
  surface at all.
- No generation token crosses the bridge. It lives in the segment, so a
  fabricated value cannot publish a torn frame as complete.
- `begin_read` returns an opaque `OpeningSample` that is deliberately not
  `PartialEq`, and the closing comparison lives inside `read_completed`. Closing
  a read with a second opening sample loads the same value with the wrong fence,
  so it passes every test while admitting torn frames on a weakly-ordered CPU —
  the type system is the only thing that can catch it.
- `PoolWriteGuard` is the documented write path. A caller that returns early or
  throws between `begin_write` and `end_write` leaves the generation odd
  permanently and the pool unreadable to every consumer; there is no `Drop` on
  the Rust side to recover it.
- `dtype` and `shape` are advisory metadata, never a bound. An unrecognized
  dtype is assumed one byte, so a shape-derived product under-estimates.
  `payload_len` is the only bound.

**Seqlock ordering.** A reader needs two different edges — `gen load → payload
loads` when opening, `payload loads → gen load` when closing. An acquire load
gives only the first. `read_completed` therefore pairs an acquire *fence* before
a relaxed load. Accesses go through `AtomicU64::from_ptr` rather than
`read_volatile` plus a bystander fence, so the code is correct under Rust's
memory model rather than only under inspection of the emitted barriers; the
shared-memory representation is unchanged.

One deliberate divergence to flag, since the extracted code is meant to match
what it was extracted from: the writer fences **before** the generation store,
while `seqlock_end` in the Python binding (`apis/python/node/src/lib.rs:1070`)
has its only `fence(Release)` *after* it. The pre-store fence is the one that
orders the payload writes ahead of the publish, so without it a weakly-ordered
target can publish an even generation whose payload is not yet visible. The
Rust version is stronger on purpose. The Python path is untouched here — this
change does not modify that binding — but it is worth a separate look.

### Verification

**Unit** — 85 tests in `dora-memory-pool`, 53 in `dora-node-api-cxx`. Every
`DORADMA` constant is pinned against the Python binding's own values. Each check
was mutation-tested: the guard was deleted, the targeted test confirmed red for
the right reason, the guard restored. Where a mutation could not be killed it is
recorded in a comment rather than left implied.

**In CI** — `examples/c++-memory-pool` runs the whole path over a live daemon on
the CPU: a 16-slot ring filled in place through the guard, both read paths, an
abandoned write cycle rejected and then recovered, and a freed pool going
permanently unreadable while a second pool still reads. Registered in
`tests/example-smoke.rs`, `scripts/smoke-all.sh` and the nightly `examples` job.
The runner fails if any `dora_pool_*` segment survives.

Its `run.rs` passes `-l dora_node_api_cxx` before the system libraries, unlike
the neighbouring C++ example. That order is required: GNU ld resolves an archive
only against what follows it, so the other order leaves `dlopen`/`dlsym`
undefined on any glibc older than 2.34. It happens to link on `ubuntu-latest`
because libdl is folded into libc there.

**On device** — a Jetson (JetPack 5, CUDA 11.4), integrated GPU:

```
[cuda-receiver] transport=unified ipc_present=false integrated_gpu=true
[cuda-receiver] registered 16896 bytes of `frames`: host=0xffff… device=0x2031…, payload at +512
[cuda-receiver] 21 frames verified through the device alias, 1 torn, 1 unavailable
```

A kernel reads bytes the host wrote, through the device alias, with no
intervening copy — the only two `cudaMemcpy` calls move a 12-byte verdict, never
the payload. The mapping is registered once at startup, before any frame exists,
and 21 frames of changing content then arrive through that same never-refreshed
alias, so it cannot be a snapshot taken at registration time. Substituting the
host pointer for the device alias faults rather than passing, because the two
are genuinely different addresses on this hardware.

**Cross-language** — a pool registered by the **untouched** Python binding is
read by a C++ node, on the CPU path and from a kernel through the device alias.
Note the precise claim: the two are **not** byte-identical. A Python CPU pool
writes no `transport` key, writes `pinned_type: "cpu"` where a C++ `unified`
pool writes `"cuda"`, uses `json.dumps` spacing so `json_len` differs, and lands
at `write_gen = 2` where `create` leaves `0`. They interoperate because the
consumer reads every one of those fields *out of the header* instead of
assuming them — which is the real argument, and a better one.

### Not in this PR

- **The Python binding is unchanged.** A Python sender with
  `receiver_device="cuda"` still fails on an integrated GPU. Follow-up: teach
  its auto-select about `unified`, and move its data plane onto the shared
  crate.
- **The C++ binding does not produce `ipc` pools** — that would require CUDA
  inside the binding. It reads one and hands out the handle.
- **The pool's daemon parameters are still hand-mirrored.**
  `MemoryPoolMetadata` lives in `dora-memory-pool`, but its `MetadataParameters`
  conversion is private to `dora-daemon`, so the binding transcribes nine keys
  and pins them with a test against a copy of the daemon's reader. Moving that
  conversion next to the type would delete the transcription at the cost of a
  `dora-message` dependency.
- The CUDA example and the Python-interop CUDA dataflow are manual, hardware-
  gated artifacts; only the CPU paths run in CI.


phil-opp:

No issues found.

I reviewed the data-plane crate (`seqlock.rs`, `doradma.rs`, `segment.rs`, `naming.rs`), the C++ binding FFI, and the shipped C++ headers, focusing on the parts a seqlock/FFI transport tends to get wrong:

- **Seqlock ordering** — the closing read correctly uses an acquire fence before a relaxed load (not a second acquire load), `begin_write` fences after marking the generation odd, and the leftover-odd recovery path lands back on an even baseline. `OpeningSample`/`DoraPoolRead` are opaque across the bridge so a caller can't forge or mis-fence a token.
- **Bounds** — `header_region_len` caps the header slice at `data_offset` (never at `segment_len`), the `ipc_flag`/`transport` contradiction is rejected before the size check can be skipped, and offset arithmetic saturates/checks. `payload()` returns `None` for `ipc` rather than a one-past-the-end ptr with a large length.
- **Free/liveness race** — `mark_freed` flips the flags under the registry mutex with Release/Acquire pairing before the ids reach C++, read tokens are serial-bound to their issuing view, and the mapping is deliberately not unmapped on free, so a racing read is not a use-after-free.
- **Naming** — component charset + explicit `..` check on construction, and `validate_segment_name` re-guards prefix/`/`/`..` at the unlink point for names arriving from a peer.

The tests exercise real behavior (specific error substrings, adversarial headers, cross-language segment shapes) rather than tautologies. Given the size and the `unsafe`/FFI surface, a human pass over the memory-ordering and the C++ header lifetimes is still worthwhile before merge.

---

phil-opp:
Two things worth fixing before this lands, neither blocking the design.

**Data race on the seqlock generation.** `segment.rs:360` / `doradma.rs:386` form a `&[u8]` over the header region (`HEADER_SIZE + json_len`, so >=256 bytes). That span *includes* byte offset 96 — the seqlock generation a peer writer stores to via `AtomicU64` — and it is read with `read_u64`. That is exactly the race the payload path takes care to avoid, and the comment at `segment.rs:346-348` ("written once, before the pool is registered") is not true for those 8 bytes. `ParsedHeader.write_gen` is never read anywhere in the workspace, so dropping the field removes the race for free (otherwise load it through `AtomicU64::from_ptr`).

**`free_memory_pool` can strand a segment.** `apis/c++/node/src/lib.rs:1790` refuses when a write cycle is open, but the `Box<DoraMemoryPool>` is consumed regardless — so after the refusal the caller has no handle left to close the cycle or retry. Net effect of the "safe" path is a segment left in `/dev/shm` and in the daemon's table for the rest of the dataflow. Returning the box on refusal would make it actionable.

Minor: `DORA_MEMORY_POOL_TRANSPORT=""` hits the `Some(other)` arm and hard-fails instead of being treated as unset (`lib.rs:1488`); and the nightly `uv venv` step writes `VIRTUAL_ENV` to `$GITHUB_ENV`, which applies to every subsequent step in the `examples` job, not just this one.

For the record, I read the memory ordering closely and it is correct — relaxed odd-store plus `fence(Release)` is the right formulation, and making `OpeningSample` opaque and non-`PartialEq` genuinely prevents the close-with-a-second-`begin_read` bug.

caothu2k1:
Thanks — all four verified at the source and fixed in 4354a259. Three are fully closed; the first is closed only in part and I'd rather say so than let it read as done.

**Seqlock read race.** Confirmed: `parse_header` decoded `write_gen` with a plain `read_u64` out of a span a peer writer stores to atomically, and the field's only consumer in the workspace was a test assertion. Dropped the field and the read. Every generation access already went through `gen_ptr()` as an `AtomicU64`, so nothing else changed.

You were also right that the justification comment was false for those 8 bytes, and that is the part still open. Dropping the field removes the *read*, but the `&[u8]` still spans offset 96, and by the standard that comment itself sets — UB whether or not anything reads through the reference — that is not yet clean. Narrowing it properly means passing the fixed header and the JSON as two slices; conveniently every fixed field `parse_header` decodes lives below offset 96, so `0..96` plus the JSON region covers it with nothing left aliasing the generation word. That touches `parse_header`'s signature and 19 test call sites, so I did not fold it into a PR you are mid-review on. The comment now states the residual plainly instead of claiming the span is write-once. Happy to do the split here or as a follow-up — your call.

**Strandable free.** This one was a regression I introduced, and your reading of it is exactly right: the refusal landed in a late review round, and I added the guard without following what happens after it fires. The box was consumed regardless, so the "safe" path guaranteed a leak where the unguarded path merely risked one. It borrows now, so a refusal leaves the caller the handle to close the cycle and free properly. The example's `release()` was the only call site.

**Empty env var.** Fixed. Worth noting the test that covered it asserted the defect — it listed `""` among the values that must be rejected — so correcting the behaviour turned it red. It now asserts that clearing the variable restores the default.

**`$GITHUB_ENV`.** Scoped to the step that needs it.

Gates green after the change: fmt, workspace clippy `-D warnings`, 87 + 56 unit tests, `cargo check --examples`, and the CPU example end to end on a live daemon (`11 copied, 10 zero-copy, 1 torn, 1 unavailable`, `/dev/shm` clean).

And thank you for reading the ordering closely — that is the part I most wanted a second pair of eyes on, and knowing the relaxed-store-plus-release-fence formulation and the `OpeningSample` opacity hold up is worth more than the rest of the review combined.

phil-opp:

Reviewed the single new commit since the last review, `4354a259`. Three of the four items are cleanly fixed and genuinely tested:

- **Strandable free** — `free_memory_pool` now takes `&mut Box<DoraMemoryPool>` instead of consuming it, so a refusal leaves the caller the handle. Correct at the library level.
- **Empty transport env var** — `resolve_transport_with` now does `env_override.filter(|v| !v.is_empty())`, so `DORA_MEMORY_POOL_TRANSPORT=` restores the default instead of hard-failing. `an_empty_environment_value_means_unset` asserts the real behavior, and the old test that had listed `""` among the rejected values was corrected rather than left asserting the defect.
- **`$GITHUB_ENV`** — scoped to the one step via `env:` instead of exported job-wide.

Two things I'd still flag:

**1. (Outstanding, previously flagged — still partially open) The header-region `&[u8]` still aliases the seqlock generation word.** `PoolSegment::open` builds `header_region = slice::from_raw_parts(base, region_len)` where `region_len >= HEADER_SIZE`, so the span still covers byte offset 96 — the generation a peer writer stores to via `AtomicU64` on every frame. This commit removed the *read* (`ParsedHeader::write_gen` and its `read_u64`), but the shared reference over those 8 bytes remains, which is UB by the very standard the payload two-step in this function is built to uphold ("UB whether or not anything reads through the reference"). The author acknowledges this in the thread and the updated comment, and defers the proper narrowing (pass the fixed `0..96` header and the JSON as two slices) to a follow-up. Noting it here so it isn't lost: this is not fully closed.

**2. (Minor, example only) The shipped example re-strands on the refusal path the library fix just made recoverable.** In `examples/c++-memory-pool/nodes/pool-sender.cc`, `release()` calls `free_memory_pool(sender, *pool)` and then `pool.reset()` **unconditionally**. On a refusal (a write cycle still open) the RPC does not free, but `reset()` drops the box anyway — and `PoolSegment` drop unmaps without unlinking (`segment.rs:7`), while the daemon-side unlink happens only through the `free_pinned_memory` RPC. So the refusal branch discards the only handle that could close the cycle and free, re-stranding the segment in `/dev/shm` + the daemon table — exactly the outcome the by-reference change was meant to make avoidable. In practice this only triggers if a cycle is left open at release time, which the example's `PoolWriteGuard` normally prevents, so impact is low; but the shipped pattern doesn't demonstrate the recovery the fix enables (e.g. `reset()` only when `result.error.is_empty()`).

---

caothu2k1:
Fixed in a3698293. You were right, and this one stings a little: the by-reference change made a refusal recoverable, and then I updated the call site just enough to compile without following the semantics through. The example rebuilt the exact trap the library fix was meant to remove — in the file people copy, two lines under a comment stating the rule it broke.

`release` now resets only after a successful free, and names the recovery at the refusal.

I verified it by mutation rather than by reading, since the refusal branch is unreachable in the normal script. Opening a write cycle immediately before `release`:

```
[sender] failed to free pool `frames`: refusing to free memory pool `frames` while a
         write cycle is open on it; close the cycle first (let the `dora::PoolWriteGuard`
         go out of scope, or call `pool_end_write`) and free the pool after that
[sender] failed to free pool `frames`: ...        <- second call, handle still there
```

That second line is the part that matters: the handle survived the refusal, so `release` could be tried again. Under the old code the second call hit the `if (!pool)` early return and reported success. Unmutated, the example is unchanged — 11 copied reads, 10 zero-copy, 1 torn, 1 unavailable, `/dev/shm` clean.

On item 1: agreed it is not closed, and thank you for restating it so it does not get lost in the thread. The read is gone; the reference over byte 96 is not. I have the narrowing mapped out — pass the fixed header (`0..96`, which covers every field `parse_header` decodes) and the JSON region as two slices — and it is mechanical apart from 19 test call sites. Still happy to land it here rather than as a follow-up; say the word and I will.

Two notes for anyone reading later. Both bugs on this path were the same mistake: adding a refusal without tracing what happens after it fires. And both survived because the example stays green — the refusal branch never runs in the normal script, so a passing example proved only the happy path. That is what pushed me to mutation-test this fix instead of eyeballing it.

phil-opp:
🤖 **Automated review by Claude** — this review was generated fully automatically, with no human in the loop.

I re-reviewed the one commit since the last automated pass (`a3698293`). Item 2 from that review is cleanly fixed: `release()` in `pool-sender.cc` now calls `pool.reset()` only after `free_memory_pool` returns without error, so a refused free (write cycle still open) leaves the caller holding the handle to close the cycle and retry, instead of dropping the box and re-stranding the segment.

No new correctness issues. Two notes:

- Item 1 remains open, as discussed earlier in the thread — flagging again only because a commit has landed since. In `PoolSegment::open` the header-region `&[u8]` is built spanning `HEADER_SIZE + json_len` (≥256) bytes, so it still covers byte offset 96, the seqlock generation word a peer writer stores to via `AtomicU64` on every frame. The *read* of that word was removed, but the aliasing shared reference remains, which is a data race / UB. The two-slice narrowing (fixed `0..96` header + separate JSON region) would close it.
- Minor: the comment at `free_memory_pool` (`apis/c++/node/src/lib.rs:1821`, "The mapping is released when `pool` drops at the end of this call") is now stale — after the by-reference change `pool` is borrowed, not consumed, so the local unmap happens when the C++ side drops the handle, not here. Behavior is correct; only the comment describes the pre-fix ownership.

---

phil-opp:
Thanks — the C++ memory-pool surface is genuinely wanted, and the review round on the seqlock read race and the strandable free landed well. Two pieces of direction before this goes further.

## This is a 1.1 feature, not a 1.0 blocker

I checked whether anything here forces a pre-1.0 landing, and it doesn't:

- the C++ memory-pool API does not exist on `main` at all (`apis/c++/node/src/lib.rs` has no `memory_pool` symbols), so the `free_memory_pool` signature change from by-value to `&mut rust::Box<T>` breaks nothing that has shipped;
- `dora-memory-pool`'s existing public items (`MemoryPoolId`, `MemoryPoolMetadata`, `MemoryPoolEntry`, `CleanupSummary`, `MemoryPoolManager`) are untouched — this only *adds* `pub mod doradma / naming / segment`.

So it is additive on every published surface and can land after 1.0.0 without costing a major version. I'm moving it to the **1.1 milestone**.

## Please split it

The diff is ~9,700 added lines, and it breaks down as roughly 35% C++ binding, 33% examples, 28% the memory-pool crate itself. That last 28% is the part with real architectural weight, and it is currently gated behind review of the other 72%.

Suggested split:

1. **`dora-memory-pool`: extract `doradma.rs` / `seqlock.rs` / `naming.rs` / `segment.rs`.** This is the DORADMA format — the header, the seqlock word at offset 96, segment lifecycle, name validation. It is worth landing on its own and soon, because the format is currently duplicated: the Python binding carries a hand-written copy of the constants and offsets, and #3079 open-codes a *third* copy in the daemon. One canonical definition is what stops that spreading.
2. **The C++ binding** (`apis/c++/node`), on top of (1).
3. **The examples** (`examples/c++-memory-pool`), last.

Landing (1) alone also unblocks #3079's third-copy problem without it having to wait on the C++ surface at all.

## Please drop the CUDA-specific parts from the first PR

`cuda_pool.hpp` and the `unified` transport (`cudaHostRegister` / `cudaHostGetDevicePointer`) exist to work around `cudaIpcGetMemHandle` being unsupported on integrated GPUs. That is a narrow platform accommodation, it is the only thing in the workspace that pulls in CUDA, and it is opt-in already. It does not need to be entangled with the format extraction or the base binding — please keep it for a later PR (or an out-of-tree header) so the first two pieces can be reviewed on their own merits.

## Still open from the last round

- **The header aliasing is not fixed, only narrowed in scope.** Dropping `ParsedHeader::write_gen` removed the *read*, but `segment.rs:366` still forms `&[u8]` over `0..HEADER_SIZE` and `:370` over `0..region_len`, both spanning `OFFSET_WRITE_GEN = 96`, while a peer writer stores there via `AtomicU64` on every frame — and `open()` is exactly what a consumer calls against a live producing pool. Forming a shared reference over concurrently-mutated bytes is UB regardless of which bytes you then read; that was imprecise in my earlier comment and I apologise for sending you down the half-fix. The fix is contained: every fixed field parsed lives below 96 (`ipc_handle` is 32..96), so `header_region_len` / `parse_header` want a `0..OFFSET_WRITE_GEN` fixed slice plus a separate `HEADER_SIZE..HEADER_SIZE+json_len` JSON slice, with the `len() >= HEADER_SIZE` guards relaxed to match.
- **`guide/src/languages/cpp.md:810` and `:849`** still show `free_memory_pool(sender, std::move(pool))`, which no longer compiles now that the bridge takes `rust::Box<T>&` — an rvalue cannot bind to it. Both snippets were added by this PR.
- **The PR body** still says the example is "registered in `tests/example-smoke.rs`, `scripts/smoke-all.sh` and the nightly `examples` job". In `example-smoke.rs` it is a ledger comment and in `smoke-all.sh` it is a `log_skip`; only the nightly job runs it end to end.


<generated with claude>

phil-opp:
Heads-up: the memory-pool transport is being parked out of `dora-rs/dora` in https://github.com/dora-rs/dora/pull/3152, so this PR's in-tree target is going away.

Worth knowing regardless of that decision: as written this adds a **third** independent implementation of the DORADMA wire format. It introduces typed `doradma.rs` / `seqlock.rs` / `segment.rs` modules in `dora-memory-pool` — which is the right consolidation — but does not migrate the Python binding onto them (the only change to `apis/python/node/Cargo.toml` is the `shared_memory_extended` workspace-dep switch, and the Python `lib.rs` has no references to the new modules). The C++ and Python sides would hand-implement the same format independently, with one shared CI-tested path between them.

The typed-module work is genuinely valuable and would port well onto the external package. See the seam contract in `external/dora-pool/README.md`.

phil-opp:
Heads-up: the memory-pool transport has moved to `libraries/extensions/memory-pool` and is now an opt-in feature outside the 1.0 guarantees (https://github.com/dora-rs/dora/pull/3152). Its daemon integration is gone — it reaches dora only through a generic extension channel (`docs/extensions.md`).

That changes the target for this PR, and worth knowing regardless: as written it adds a **third** independent implementation of the DORADMA wire format. The typed `doradma.rs` / `seqlock.rs` / `segment.rs` modules are the right consolidation, but the Python binding is not migrated onto them, so the C++ and Python sides would hand-implement the same format independently.

Porting those typed modules into the extension crate — and moving the Python side onto them — would be a clear improvement over either state.

caothu2k1:
Closing this — #3152 removed the ground it was built on, and a rebase is not
what it needs.

Concretely, checked against `main`: `libraries/extensions/memory-pool` is now
`tensor-pool`, and the `RegisterPinnedMemory` / `ReadPinnedMemory` /
`FreePinnedMemory` variants this PR's control plane calls are gone from
`node_to_daemon.rs`. So the crate it modifies and the daemon API it depends on
both no longer exist. That is a redesign on the extension channel, not a
conflict to resolve, and it deserves its own PR rather than 9,700 lines of
history dragged onto a different foundation.

For the record, #3152 looks right to me, and its reasoning is the same one that
motivated the typed modules here: ~40 `unsafe` sites hand-parsing a 256-byte
header at hardcoded offsets, in memory another process writes, is exactly the
shape that keeps producing #2881 / #2935 / #3015.

## What was verified, so it is not lost

On a Jetson (JetPack 5, CUDA 11.4), integrated GPU, over a live daemon:

- A CUDA kernel read bytes the host wrote through the device alias of a pool
  segment — `transport=unified`, `ipc_present=false`, no CUDA IPC anywhere. The
  mapping was registered once before any frame existed, and 21 frames of
  changing content then arrived through that same never-refreshed alias, so it
  could not be a snapshot. Substituting the host pointer for the device alias
  faulted rather than passing, because the two are genuinely different
  addresses there.
- A pool registered by the then-current Python binding was read by a C++ node,
  on the CPU path and from a kernel through the device alias.
- 87 + 56 unit tests, each guard mutation-checked: the guard deleted, the
  targeted test confirmed red for the right reason, the guard restored.

## Offer

You said twice that the typed modules would port well, and that is the piece I
would like to contribute: `doradma` / `seqlock` / `naming` / `segment` into
`tensor-pool`, CUDA-free, as its own PR — with the header-region aliasing you
flagged closed properly rather than carried across. Happy to open an issue first if you would rather
scope it before any code, given #1872's history on exactly that point.

Thank you for the review. The seqlock read race and the strandable free were
both real, both mine, and neither would have been found by the tests I had.

PR #3107, opened 2026-08-10
test(message): deflake HLC timestamp serde_json::Value round-trip test

### LINKED ISSUE #3104: Flaky test: `a_freshly_generated_hlc_id_uses_all_sixteen_bytes` fails ~1/256 runs under uhlc 0.9
## Summary

The test `a_freshly_generated_hlc_id_uses_all_sixteen_bytes` in `libraries/message/tests/uhlc_wire_format.rs` is **non-deterministic** and will fail on roughly **1 in 256** `cargo test` runs. Because `dora-message` is exercised by `cargo test --all` in both the PR gate and nightly, this introduces an intermittent, hard-to-reproduce CI failure.

## Where it came from

Introduced by **PR #3016 / commit `4a24101`** ("chore(deps): upgrade uhlc 0.5 -> 0.9"), which added the `uhlc_wire_format.rs` test file. The flakiness is a direct consequence of the uhlc 0.9 behavior change, so it did not exist before this commit.

## The bug

```rust
// libraries/message/tests/uhlc_wire_format.rs:150
fn a_freshly_generated_hlc_id_uses_all_sixteen_bytes() {
    let timestamp = dora_message::uhlc::HLC::default().new_timestamp();
    assert_eq!(timestamp.get_id().size(), 16);   // <-- probabilistic
    ...
}
```

Under uhlc 0.9:

- `HLC::default()` builds its id with `ID::rand()` (`uhlc-0.9.0/src/lib.rs:179`).
- `ID::rand()` is `OsRng.gen_range(1..u128::MAX)` (`uhlc-0.9.0/src/id.rs:87`) — a uniformly random 128-bit value.
- `ID::size()` returns the number of **significant** little-endian bytes: `MAX_SIZE - (u128::from_le_bytes(self.0).leading_zeros() / 8)` (`uhlc-0.9.0/src/id.rs:64`).

`size()` therefore equals 16 only when the most-significant byte of the random id is non-zero. The top byte is zero with probability `1/256`, so `size() == 16` fails ~`1/256 ≈ 0.39%` of the time.

This is a *new* failure mode relative to uhlc 0.5: there `HLC::default()` seeded its id from `Uuid::new_v4()`, whose most-significant bits are the fixed UUID version/variant nibbles, so `size()` was effectively always 16. The assertion was implicitly safe under 0.5 and became flaky under 0.9.

## Empirical evidence

Running `HLC::default().new_timestamp()` 2,000,000 times against `uhlc = "=0.9.0"`:

```
short (size<16): 7919 / 2000000 = 0.3960%
theoretical 1/256 = 0.3906%
```

The measured rate matches the theoretical `1/256` exactly.

## Risk / impact

- Intermittent red CI on unrelated PRs (~1 run in 253), which is corrosive to trust in the gate and wastes reruns.
- The failure is confined to this one test — no production code path depends on `size() == 16`; the JSON round-trip below the assertion works for any non-zero id, including short ones. So this is a test-quality defect, not a runtime bug.
- Mildly ironic: the assertion's own comment says it "guards the premise" of the neighboring `serde_json::Value` round-trip test, but the guard itself is the non-deterministic part. The round-trip it guards is actually valid for *any* id, short or not.

## Suggested fix

Don't assert a probabilistic property of a random id. Options, in rough order of preference:

1. **Test the round-trip on a deterministic full-width id** instead of a random one. The file already has `golden_timestamp()` whose id is `[0x01 ..= 0x10]` (top byte `0x10`, so `size() == 16` deterministically). Use that (or any fixed 16-significant-byte id) for the `to_value`/`from_value` assertion and drop the random `HLC::default()` entirely from this test.
2. If a real `HLC::default()` timestamp is desired, weaken the assertion to what actually matters — e.g. `assert!(timestamp.get_id().size() >= 1)` (id is always non-zero) — and keep the round-trip assertion, which holds regardless of width.

Option 1 keeps the original intent (prove a production-shaped, all-16-bytes id survives `serde_json::Value`) while making it deterministic.

---

*This issue was created by a scheduled, automated Claude code-review check. The finding above was verified empirically (uhlc 0.9.0 source inspection + a 2,000,000-iteration measurement), but a maintainer should confirm before acting.*

phil-opp:
## Summary

Fixes #3104.

The test `a_freshly_generated_hlc_id_uses_all_sixteen_bytes` in `libraries/message/tests/uhlc_wire_format.rs` was non-deterministic and failed on roughly **1 in 256** `cargo test` runs. Because `dora-message` is exercised by `cargo test --all` in both the PR gate and nightly, this introduced an intermittent, hard-to-reproduce CI failure.

## Root cause

Under uhlc 0.9, `HLC::default()` seeds its id via `ID::rand()` — a uniformly random 128-bit value. `ID::size()` returns the number of *significant* little-endian bytes, so `size() == 16` only holds when the most-significant byte is non-zero. That byte is zero with probability `1/256`, so the assertion `assert_eq!(timestamp.get_id().size(), 16)` failed ~0.39% of the time. Under uhlc 0.5 the id was seeded from a UUID whose top bits were fixed version/variant nibbles, so it was effectively always 16 bytes — the assertion was implicitly safe then and became flaky after the 0.5 → 0.9 bump (#3016).

No production path depends on `size() == 16`; the JSON round-trip works for any non-zero id. This is a test-quality defect, not a runtime bug.

## Fix

The full-width (`> u64`) property is already pinned **deterministically** by `timestamp_survives_a_serde_json_value_round_trip`, which uses `golden_timestamp` (top byte `0x10`). This test's remaining value is confirming that a *real*, randomly-generated `HLC::default()` timestamp survives the `serde_json::Value` round trip — which is lossless for an id of any width.

So the probabilistic `size()` assertion is dropped and the real-HLC round-trip check is kept (and the test renamed accordingly). This removes the flakiness while preserving the real-HLC coverage the original test added, on top of the deterministic full-width coverage that already exists.

## Testing

- `cargo test -p dora-message --test uhlc_wire_format` — all 7 tests pass.
- `cargo fmt -p dora-message -- --check` — clean.

https://claude.ai/code/session_01EZXbbi2V3Dw4174Xn3AgMX

---
_Generated by [Claude Code](https://claude.ai/code/session_01EZXbbi2V3Dw4174Xn3AgMX)_

### DIFF
--- a/libraries/message/tests/uhlc_wire_format.rs
+++ b/libraries/message/tests/uhlc_wire_format.rs
@@ -147,12 +147,20 @@ fn timestamp_survives_a_serde_json_value_round_trip() {
 }
 
 #[test]
-fn a_freshly_generated_hlc_id_uses_all_sixteen_bytes() {
-    // Guards the premise of the test above: if `HLC::default()` ever produced a
-    // short id, the `serde_json::Value` round trip would pass for reasons that
-    // do not generalize to production timestamps.
+fn a_freshly_generated_hlc_timestamp_survives_a_serde_json_value_round_trip() {
+    // Companion to `timestamp_survives_a_serde_json_value_round_trip`, which pins
+    // the property on the deterministic full-width `golden_timestamp` (top byte
+    // `0x10`, so it exceeds `u64`): confirm a *real* `HLC::default()` timestamp
+    // also survives `serde_json::Value`.
+    //
+    // We deliberately do NOT assert that the id uses all sixteen bytes. Under
+    // uhlc 0.9 `HLC::default()` seeds its id with a uniformly random 128-bit
+    // value (`ID::rand()`), whose most-significant byte is zero ~1/256 of the
+    // time, so `size() == 16` is probabilistic and would flake ~1 run in 256
+    // (dora-rs/dora#3104). The round trip below is lossless for an id of any
+    // width — that, not the byte count of a random id, is the property that
+    // matters here.
     let timestamp = dora_message::uhlc::HLC::default().new_timestamp();
-    assert_eq!(timestamp.get_id().size(), 16);
 
     let value = serde_json::to_value(Timestamped {
         inner: (),


phil-opp:

No issues found.

This is a clean test-quality fix. The dropped assertion (`timestamp.get_id().size() == 16`) was genuinely probabilistic — under uhlc 0.9, `HLC::default()` seeds its id via `ID::rand()`, and `ID::size()` counts significant little-endian bytes, so a zero most-significant byte (~1/256) yields `size() == 15`.

The retained round-trip check still exercises meaningful behavior: it builds a real `HLC::default()` timestamp, runs it through `serde_json::to_value` → `from_value`, and asserts `decoded.timestamp == timestamp`, so it is not vacuous. Deterministic full-width (`> u64`) coverage is still pinned separately by `timestamp_survives_a_serde_json_value_round_trip` (golden timestamp), so nothing is lost.

---

PR #3109, opened 2026-08-10
fix(daemon): compute `dora top` disk-I/O rate from the real sampling window

phil-opp:
`collect_and_send_metrics_bg` (`binaries/daemon/src/lib.rs`) turned sysinfo's per-refresh disk byte deltas into the `dora top` "I/O READ" / "I/O WRITE" rates by dividing by a hard-coded constant:

```rust
disk_read_bytes: Some((disk_read as f64 / METRICS_INTERVAL_SECS) as u64),   // 2.0
```

Both halves of that division were wrong in practice.

**The divisor was not the real window.** `process.disk_usage()` reports bytes since the previous `refresh_processes_specifics` call, not "over the last 2 s". When a cycle is skipped — `try_lock` fails because a previous collection is still running — the next refresh covers ~two intervals but was still divided by one, roughly doubling the reported rate.

**The numerator was not always a delta.** For a process sysinfo is seeing for the first time there is no previous refresh, so `read_bytes` is its *total since start*. A node that read a 4 GB model file during startup reported ~2 GB/s on its first sample. This hits every freshly spawned node — a second dataflow on a long-lived daemon, or a restart with a fresh PID — not just the daemon's very first refresh.

## Fix

The `System` and the facts needed to interpret its counters now live together in one `MetricsState` behind a single mutex:

```rust
struct MetricsState {
    system: sysinfo::System,
    last_refresh: Option<Instant>,  // the divisor
    processes: HashMap<Pid, u64>,   // pid -> start_time: who had a baseline
}
```

A collection takes the whole state at once, so the measured window and the "already observed" set always describe the very deltas that refresh produced. Rates come from the measured window (`disk_rate_bytes_per_sec`), and only processes present in the previous refresh contribute bytes — anything else would contribute total-since-start. `start_time` is compared alongside the PID, so a recycled PID correctly counts as a new process. When nothing in a node's process tree has a baseline there is no measurement to report: `disk_read_bytes` / `disk_write_bytes` stay `None` and `dora top` shows `N/A`.

Keeping the two together also closes a hole a separate baseline field would have had. `try_lock` releases the mutex immediately after `mem::take`, so a concurrent collection does not actually skip — it gets an *empty* `System`, whose deltas are all total-since-start. Pairing that with a populated baseline would report exactly the spike this PR removes; taking the state as a unit means such a collection starts from `Default` and reports no rate.

A node whose descendant is new in a given window under-reports by that descendant's bytes for that one sample, rather than spiking or blanking the row — the descendant is counted from the next refresh on. A cycle with no running nodes leaves the state untouched instead of clearing it: nothing refreshed, so the stored baseline still describes the stored `System`, and the next refresh's delta and window both simply span the longer gap.

The now-unused `METRICS_INTERVAL_SECS` is removed; `METRICS_INTERVAL` still drives the interval timer.

## Validation

`cargo test -p dora-daemon` — 244 passed, including 6 in `disk_rate_tests`: no baseline / near-zero window / division by the measured window rather than the constant (8 MB over 4 s = 2 MB/s, where the old divisor said 4 MB/s) / only processes seen in the previous refresh count as baselined, with a recycled PID rejected / a node with nothing summed gets no window even when the refresh window is valid / `snapshot_processes` and `has_disk_baseline` agreeing on a real process from a real `System`.

`cargo clippy -p dora-daemon -- -D warnings` and `cargo fmt --all -- --check` clean.

---

⚠️ **This is a machine-generated PR** authored by Claude (Claude Code) as part of an automated codebase review. Please review carefully before merging.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -195,7 +195,11 @@ use zenoh::shm::{PosixShmProviderBackend, ShmProvider, ShmProviderBuilder};
 
 const STDERR_LOG_LINES_MAX: usize = 500;
 const METRICS_INTERVAL: Duration = Duration::from_secs(2);
-const METRICS_INTERVAL_SECS: f64 = METRICS_INTERVAL.as_secs_f64();
+/// Shortest sampling window (seconds) we trust for a disk-I/O rate. A window
+/// below this either has no prior baseline (first sample) or is so short that
+/// dividing by it would blow the rate up into a meaningless spike, so we report
+/// no rate for that sample instead.
+const MIN_METRICS_WINDOW_SECS: f64 = 0.1;
 /// Capacity of the Zenoh publish drain channel. Large enough for burst
 /// patterns; messages are dropped with a warning when full.
 const ZENOH_PUBLISH_CHANNEL_CAPACITY: usize = 256;
@@ -1783,6 +1787,9 @@ pub struct Daemon {
     pub(crate) builds: BTreeMap<BuildId, BuildInfo>,
     pub(crate) git_manager: GitManager,
     pub(crate) metrics_system: Arc<std::sync::Mutex<sysinfo::System>>,
+    /// Wall-clock instant of the previous successful metrics refresh, used to
+    /// compute the real disk-I/O sampling window. `None` until the first refresh.
+    pub(crate) metrics_last_refresh: Arc<std::sync::Mutex<Option<Instant>>>,
     /// Opaque, dataflow-scoped store for out-of-tree extensions. See
     /// `extension_table` and `docs/extensions.md`.
     pub(crate) extensions: ExtensionTable,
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff — no issues found. The new `metrics_last_refresh` timestamp and the sysinfo `System` baseline stay in sync: the timestamp is set to `Some` only on the successful-refresh path that also restores the refreshed `System` (carrying the per-process baseline), and the try-lock-skip / no-running-nodes / panic paths all leave it consistent — the last two clear it to `None` so the next sample is suppressed rather than producing a spike. First sample correctly yields `None`, and windows shorter than 0.1s are rejected. The unit tests exercise the no-baseline, near-zero-window, and real-window division cases (8 MB / 4 s = 2 MB/s) with concrete assertions that distinguish the measured window from the old constant divisor.

---

phil-opp:

The rate math is safe: `Instant` is monotonic so there is no clock-jump path, `saturating_duration_since` guards the subtraction, the `MIN_METRICS_WINDOW_SECS = 0.1` floor prevents a divide-by-tiny, and returning `None` on the first sample / after a `spawn_blocking` panic / after a no-nodes cycle correctly suppresses the total-since-start spike. The `try_lock` early return leaves the timestamp untouched, which is right.

There is a gap between the rationale and the implementation, though. The doc comment on `disk_rate_bytes_per_sec` correctly identifies the spike as **per-process** ("sysinfo then reports *the process's* total-since-start I/O"), but the guard is **System-level** (`binaries/daemon/src/lib.rs:1985-1990`). A node spawned into an already-warm `System` — a second dataflow on a long-lived daemon, or any node restart, which gets a fresh PID — is seen by sysinfo for the first time, so its `disk_usage().read_bytes` is total-since-process-start and is now divided by a perfectly valid 2 s window and reported as a real rate.

Concretely: `dora up`, run dataflow A, stop it, run dataflow B whose node has already read 4 GB from a model file → the first `dora top` sample shows roughly 2 GB/s. Closing this needs a per-PID "seen" set rather than a single global timestamp. Not a regression — main is equally wrong — so not blocking, but the headline claim is not fully delivered yet.

phil-opp:
Agreed on both points — the rate math is the intended scope here, and the per-process first-observation spike is a real but separate limitation, which the PR body's scope note already calls out (a node newly seen by a warm `System` — a second dataflow or a restart with a fresh PID — still reports total-since-start for one sample). Closing that fully needs a per-PID "seen" set, and it's complicated further by the descendant aggregation (a freshly spawned child folded into the parent's total would spike the aggregate even for an established PID), so it's deliberately out of scope for this change rather than a claim left half-delivered. As you note, `main` is equally affected, so this is a strict improvement for the divisor/first-sample/panic-reset cases without regressing that one.

Rebased onto current `main` (resolved a conflict where the `Daemon` struct fields were reorganized) and re-verified: `cargo clippy -p dora-daemon -- -D warnings` clean, `disk_rate` unit tests pass.

---

phil-opp:
The Aug 18 rebase is conflict-resolution only (the `Daemon` struct field reorg) — the disk-rate logic is unchanged, so the divisor / first-sample / panic-reset fixes remain safe.

Re-mentioning the still-open point from the previous review, now that the rebase touched the PR but didn't change this: the guard is still System-level (a single global `metrics_last_refresh: Arc<Mutex<Option<Instant>>>`), not per-PID. So a node newly seen by an already-warm `System` — a second dataflow on a long-lived daemon, or a restart with a fresh PID — still gets its total-since-start I/O divided by a valid ~2 s window and reported as a one-sample spike. `main` is equally affected and a full fix needs a per-PID "seen" set (complicated by descendant aggregation), so this is a strict improvement rather than a regression — flagging only that the "compute from the real sampling window" claim isn't fully delivered for the first-observation case.

---

---

phil-opp:
Fixed rather than deferred — the first-observation case is now handled per-PID, so the "compute from the real sampling window" claim holds for it too.

`MetricsState` now holds the sysinfo `System` together with the two facts needed to interpret its counters: the instant of the refresh that produced it, and a `pid -> start_time` snapshot of the processes it observed. A collection takes the whole state at once, so the window and the observed-process set always describe the very deltas that refresh produced. `DiskDelta` sums only processes that had a baseline; when none did — the node's own first observation — there is no measurement and the columns stay `N/A` rather than showing a spike. `start_time` is compared alongside the PID, so a recycled PID counts as a new process.

Two things fell out of doing it this way that are worth flagging:

**Descendant aggregation didn't turn out to be the blocker.** My first cut required *every* process in the tree to have a baseline, which is wrong in the other direction: a node that spawns a short-lived child each tick would show `N/A` forever. Summing only the baselined processes instead under-reports by the newcomer's bytes for that one window — it's counted from the next refresh on — which is the right trade against both the spike and a permanently blank row.

**Holding the `System` and its baseline in one mutex is load-bearing, not just tidy.** `try_lock` releases the guard immediately after `mem::take`, so a concurrent collection doesn't actually skip — it gets an *empty* `System`, whose deltas are all total-since-start. With the baseline in a separate field it would have paired those with a populated snapshot and reported precisely the spike this PR removes. Taking the state as a unit means such a collection starts from `Default` and reports nothing.

One deliberate behavior change from the earlier version: a cycle with no running nodes now leaves the state untouched instead of clearing it. Nothing refreshed, so the stored baseline still describes the stored `System`, and the next refresh's delta and window both simply span the longer gap — clearing it discarded a baseline that was still valid.

I also looked at deriving "first observation" from sysinfo directly (`disk_usage().read_bytes == total_read_bytes` iff `old_read_bytes == 0`) instead of keeping the map. It doesn't work: a process that has been observed for several refreshes but has done no I/O yet satisfies the same condition, so an idle node would show `N/A` permanently rather than `0.0`. The explicit snapshot distinguishes the two.

`cargo test -p dora-daemon` — 244 passed, 6 in `disk_rate_tests`. `cargo clippy -p dora-daemon -- -D warnings` and `cargo fmt --all -- --check` clean.

PR #3110, opened 2026-08-10
fix(message): don't mislabel signal 23 as SIGURG on macOS/BSD daemon hosts

phil-opp:
## Issue

`NodeError`'s `Display` impl (`libraries/message/src/common.rs`) maps `NodeExitStatus::Signal(n)` to a signal name from a hard-coded table:

```rust
22 => "SIGTTOU".into(),
23 => "SIGURG".into(),
other => other.to_string().into(),
```

Those are the Linux/glibc numbers. But `NodeExitStatus::Signal` is produced from `ExitStatus::signal()` under `#[cfg(unix)]` (same file), which also covers macOS and the BSDs — and signal numbers past the portable POSIX range diverge there:

| number | Linux | macOS / BSD |
|-------:|-------|-------------|
| 23 | `SIGURG` | **`SIGIO`** |

So a node killed by signal 23 on a macOS/BSD daemon host was reported as `SIGURG` when it was actually `SIGIO`. (Severity is low — it only affects the human-readable message, not control flow — but it is a genuine cross-platform mislabel, and the test was even named `..._uses_linux_signal_names`, baking in the assumption silently.)

## Fix

Signals `1..=22` in the table share the same number on every Unix dora targets, so they stay unconditional. Name signal 23 per target OS and fall back to the raw number on any other platform rather than printing a wrong name:

```rust
22 => "SIGTTOU".into(),
#[cfg(any(target_os = "linux", target_os = "android"))]
23 => "SIGURG".into(),
#[cfg(any(target_os = "macos", target_os = "ios", target_os = "freebsd",
          target_os = "openbsd", target_os = "netbsd", target_os = "dragonfly"))]
23 => "SIGIO".into(),
other => other.to_string().into(),
```

The existing test is renamed to `node_error_signal_display_uses_platform_signal_names` and made platform-aware, so the **macOS nightly CI job actually exercises the `SIGIO` mapping**.

## Validation

```
cargo test -p dora-message      # 140 passed (incl. the updated signal test)
cargo clippy -p dora-message -- -D warnings
cargo fmt --all -- --check
```

Verified against 1.97.1 (the CI-pinned toolchain).

---

---

### DIFF
--- a/libraries/message/src/common.rs
+++ b/libraries/message/src/common.rs
@@ -140,8 +140,23 @@ impl std::fmt::Display for NodeError {
                     13 => "SIGPIPE".into(),
                     14 => "SIGALRM".into(),
                     15 => "SIGTERM".into(),
+                    // Signals 1..=22 above share the same number on every Unix
+                    // dora targets. Numbers past that diverge — signal 23 is
+                    // SIGURG on Linux but SIGIO on macOS/BSD — so name 23 per
+                    // target OS and fall back to the raw number on any other
+                    // platform rather than printing a wrong name.
                     22 => "SIGTTOU".into(),
+                    #[cfg(any(target_os = "linux", target_os = "android"))]
                     23 => "SIGURG".into(),
+                    #[cfg(any(
+                        target_os = "macos",
+                        target_os = "ios",
+                        target_os = "freebsd",
+                        target_os = "openbsd",
+                        target_os = "netbsd",
+                        target_os = "dragonfly"
+                    ))]
+                    23 => "SIGIO".into(),
                     other => other.to_string().into(),
                 };
                 if matches!(self.cause, NodeErrorCause::GraceDuration) {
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff — no issues found. The unconditional signal-name table only contains numbers whose name is identical on Linux and macOS/BSD (the divergent low numbers 7/10/12/16–21 are correctly left to the numeric fallback rather than mislabeled), and signal 23 is the only added divergent case, cfg-gated correctly to `SIGURG` (Linux/Android) vs `SIGIO` (macOS/BSD) with a raw-number fallback elsewhere. The renamed test asserts the real per-OS `Display` output, and the `signal 40 → "signal 40"` fallback test is a genuine assertion.

---

phil-opp:

The mapping itself is right, and no other number is mislabeled: the table at `libraries/message/src/common.rs:140-160` lists only 1, 2, 3, 4, 6, 8, 9, 11, 13, 14, 15, 22, all identical on Linux and macOS/BSD. The three that actually diverge below 23 (7 SIGBUS/SIGEMT, 10 SIGUSR1/SIGBUS, 12 SIGUSR2/SIGSYS) are absent, so the "1..=22 agree" premise holds for what is in the table. Windows falls through to the raw number.

The caveat is that the `#[cfg]` keys off the **formatting binary's** target OS, but `NodeError` is serialized on the daemon and rendered in the CLI (`binaries/cli/src/formatting.rs:33,75`). So in a mixed cluster a Linux CLI still prints `SIGURG` for a node killed on a macOS daemon — the exact case the title names — and a Windows CLI now prints a bare `23` where it previously printed `SIGURG`.

A correct fix needs the producing daemon's OS carried in the message, or the name resolved daemon-side, rather than `cfg!` at the display site.

phil-opp:
Good catch — this is the wrong layer, and I'm closing it.

I verified the objection: `NodeExitStatus::Signal` is produced on the daemon (under `#[cfg(unix)]`) but rendered in the CLI, and the `#[cfg(target_os = ...)]` here keys off the **rendering** binary's OS, not the daemon's. So:

- A Linux CLI viewing a node killed on a macOS daemon still prints `SIGURG` — the exact mixed-cluster case in the title is **not** fixed.
- A Windows CLI now prints a bare `23` where it previously printed `SIGURG` — a small regression for the common Windows-CLI/Linux-daemon case.

A signal number is only unambiguous alongside the OS that produced it, so a correct fix has to resolve the name daemon-side (where the producing OS is known) or carry the producer OS in the message — a deliberate message-format change that's out of scope for this automated pass and shouldn't ship as a display-site `cfg!`. Closing rather than layering another band-aid. Thanks.

---

PR #3111, opened 2026-08-10
docs(cli): correct the `dora top` note about what each metric column means

phil-opp:
## Issue

The `dora top` help note (`binaries/cli/src/command/inspect/top.rs`) states:

```
/// Note:
/// - Values are averaged over the last refresh period
```

That is inaccurate for most of the table. Verified against the data source:

- **NET TX / NET RX** — `net_bytes_sent` / `net_bytes_received` are monotonic counters that are only ever `fetch_add`-ed and never reset (`running_dataflow.rs`, incremented in `daemon/src/lib.rs`). They are **lifetime cumulative totals**, not per-period averages.
- **RESTARTS** is a cumulative count; **QUEUE** (pending messages) and **MEMORY (MB)** are point-in-time snapshots; **CPU%** is an instantaneous sample.
- Only **I/O READ / I/O WRITE** are a per-refresh-interval rate.

Under the old wording, a steadily growing cumulative "NET TX" reads as an ever-rising throughput.

## Fix

Reword the note to state per-column semantics accurately (instantaneous vs. per-interval rate vs. cumulative/point-in-time). Documentation only — **no behavior change**.

## Validation

```
cargo fmt --all -- --check
cargo clippy -p dora-cli -- -D warnings
```

(Doc-comment-only change to a clap `Args` help string.)

---

---

### DIFF
--- a/binaries/cli/src/command/inspect/top.rs
+++ b/binaries/cli/src/command/inspect/top.rs
@@ -30,7 +30,10 @@ use super::super::{Executable, default_tracing};
 /// so this works for distributed dataflows across multiple machines.
 ///
 /// Note:
-/// - Values are averaged over the last refresh period
+/// - Columns report over different time bases: CPU% is averaged over the last
+///   refresh interval and I/O READ and I/O WRITE are per-interval rates; MEMORY
+///   and QUEUE are instantaneous snapshots; NET TX, NET RX and RESTARTS are
+///   cumulative totals since the dataflow started
 /// - CPU percentage is of a single core (values can add to more than 100% if multiple cores are used)
 /// - Nodes can run on different machines with potentially different CPUs, so percentages are not comparable across machines
 #[derive(Debug, Args)]


phil-opp:

Reviewed the diff — documentation-only change to the `dora top` header note, and the new per-column semantics (instantaneous CPU/MEMORY, per-interval I/O rates, cumulative NET TX/RX/RESTARTS, current QUEUE) match how those columns are actually computed. No issues found.

---

phil-opp:
**Automated review (Claude)**

I found one issue with the reworded note: it groups `CPU%` and `MEMORY` together as "instantaneous samples," but CPU% is not instantaneous — it's an interval-averaged rate, the same category as the I/O columns.

In `binaries/daemon/src/lib.rs:512` the value comes from `process.cpu_usage()`. sysinfo defines `cpu_usage()` as the fraction of CPU time the process consumed *between the two most recent refreshes*, and the metrics loop refreshes every `METRICS_INTERVAL = 2s` (`lib.rs:188-189`). So CPU% is an average over the ~2s refresh window — a short CPU spike inside that window gets smoothed out, which is exactly what the old "averaged over the last refresh period" wording described. It's computed in the same spirit as I/O READ / I/O WRITE (`lib.rs:548-552`, byte delta ÷ `METRICS_INTERVAL_SECS`), not as a point-in-time reading.

Only MEMORY (`process.memory()`, `lib.rs:513`) is genuinely point-in-time. Suggest moving CPU% out of the instantaneous group, e.g.:
- MEMORY (and QUEUE) are point-in-time snapshots;
- CPU%, I/O READ and I/O WRITE are per-refresh-interval rates.

---
🤖 _This is a fully automated review by Claude — no human has verified these findings. Treat them as advisory._

---

phil-opp:

Four of the five corrections check out against `binaries/cli/src/command/inspect/top.rs:388-441`: NET TX/RX are `format_bytes(net_tx_bytes/net_rx_bytes)` (cumulative), I/O READ/WRITE are `disk_read_mb_s`/`disk_write_mb_s` (rates), QUEUE is `pending_messages` (current), RESTARTS cumulative, MEMORY instantaneous.

The CPU% claim goes the wrong way. The daemon sends `sysinfo::Process::cpu_usage()` (`binaries/daemon/src/lib.rs:1985`), which is the average over the interval between the last two refreshes — which is what the *old* wording ("averaged over the last refresh period") said. Demoting it to "instantaneous sample" makes that column less accurate while fixing the other four.

phil-opp:
You're right about CPU% — `sysinfo::Process::cpu_usage()` is the average over the interval since the last refresh, not an instantaneous sample, so grouping it with MEMORY was wrong. Fixed: the note now says CPU% is averaged over the last refresh interval (alongside the I/O rates), and only MEMORY and QUEUE are instantaneous snapshots. The four column classifications you confirmed are unchanged. Rebased onto current `main` and force-pushed.

---

phil-opp:
The latest force-push addresses the earlier point about CPU%. The reworded note now reads "CPU% is averaged over the last refresh interval and I/O READ and I/O WRITE are per-interval rates; MEMORY and QUEUE are instantaneous snapshots; NET TX, NET RX and RESTARTS are cumulative totals" — CPU% (`sysinfo::Process::cpu_usage()`, an interval-averaged rate) is now correctly grouped with the I/O rates rather than with the instantaneous MEMORY/QUEUE columns. No remaining issues found; documentation-only change and all column classifications match how they're computed.

---

---

PR #3112, opened 2026-08-10
fix(node): return an error instead of panicking on a zero-field JSON schema

phil-opp:
## Issue

`read_json_value_as_arrow` (`apis/rust/node/src/daemon_connection/json_to_arrow.rs`) ended with:

```rust
Ok(batch.column(0).to_data())
```

`RecordBatch::column(0)` indexes out of bounds and **panics** when the batch has no columns. This is a `pub fn` whose `schema` is caller-supplied, and a zero-field schema is reachable: the node integration-testing harness (`node_integration_testing.rs`) builds the schema via `infer_json_schema_from_iterator(...)` when no `data_type` is given, and an empty JSON array infers a schema with no fields. That degenerate-but-valid input turns into a panic instead of a recoverable error. (The sibling `read_from_json_with_schema` is unaffected — its `wrapped()` always injects an `"inner"` field, guaranteeing ≥1 column.)

## Fix

Use `batch.columns().first().context(...)?` so a zero-column batch returns an `eyre::Result` error rather than panicking:

```rust
Ok(batch
    .columns()
    .first()
    .context("JSON record batch has no columns")?
    .to_data())
```

## Validation

```
cargo test -p dora-node-api --lib json_to_arrow    # 2 passed (new)
cargo clippy -p dora-node-api -- -D warnings
cargo fmt --all -- --check
```

New unit tests cover: a zero-field schema returns `Err` (no panic), and a normal single-field schema still round-trips.

---

---

### DIFF
--- a/apis/rust/node/src/daemon_connection/json_to_arrow.rs
+++ b/apis/rust/node/src/daemon_connection/json_to_arrow.rs
@@ -3,8 +3,21 @@ use std::{
     sync::Arc,
 };
 
-use arrow::array::{Array, ArrayData};
+use arrow::array::{Array, ArrayData, RecordBatch};
 use eyre::{Context, ContextCompat};
+
+/// Extract the first column of `batch` as [`ArrayData`].
+///
+/// `RecordBatch::column(0)` panics on a batch with no columns; both JSON
+/// readers can be handed (or infer) a zero-field schema, so return an error
+/// instead of panicking.
+fn first_column_data(batch: &RecordBatch) -> eyre::Result<ArrayData> {
+    Ok(batch
+        .columns()
+        .first()
+        .context("JSON record batch has no columns")?
+        .to_data())
+}
 
 pub fn read_json_bytes_as_arrow(data: &[u8]) -> eyre::Result<ArrayData> {
     match arrow_json::reader::infer_json_schema(wrapped(data), None) {
@@ -36,10 +49,7 @@ fn read_from_json_with_schema(
         .context("no record batch in JSON")?
         .context("failed to read record batch")?;
 
-    if batch.num_columns() == 0 {
-        eyre::bail!("JSON record batch has no columns");
-    }
-    Ok(batch.column(0).to_data())
+    first_column_data(&batch)
 }
 
 // wrap data into JSON object to also allow bare JSON values
@@ -70,10 +80,7 @@ pub fn read_json_value_as_arrow(
         .flush()
         .context("failed to read record batch")?
         .context("no record batch in JSON")?;
-    if batch.num_columns() == 0 {
-        eyre::bail!("JSON record batch has no columns");
-    }
-    Ok(batch.column(0).to_data())
+    first_column_data(&batch)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff — no issues found. Swapping `batch.column(0)` for a shared `first_column_data` helper backed by `columns().first()` correctly removes the out-of-bounds panic on a zero-column batch, and both JSON→Arrow call sites are updated.

Minor test note: `empty_schema_returns_err_not_panic` asserts only `is_err()`, and a zero-field schema may already short-circuit earlier at the `"no record batch in JSON"` context before ever reaching `first_column_data` — so it validates the overall no-panic contract but doesn't guarantee the new helper is the branch actually exercised. If the intent is to pin the new helper specifically, constructing a batch that has a record but zero columns would target it directly. The single-field round-trip test is solid.

---

phil-opp:
**Automated code review**

I found the following issue (which a prior automated review missed):

1. `apis/rust/node/src/daemon_connection/json_to_arrow.rs` — I don't see why this change is necessary. The PR is described as fixing a reachable panic where the code "ended with `Ok(batch.column(0).to_data())`", but the current base (`0eebb32`) already guards **both** call sites: `read_from_json_with_schema` and `read_json_value_as_arrow` each contain `if batch.num_columns() == 0 { eyre::bail!("JSON record batch has no columns"); }` immediately before `batch.column(0).to_data()`. So `column(0)` is never reached on a zero-column batch, and no panic is reachable. This change extracts a `first_column_data` helper that is functionally equivalent to the existing guarded code (both return `Err` on zero columns, both take the first column otherwise) — it is a refactor, not a bug fix.

2. The base already contains a regression test for the degenerate case, `empty_field_schema_errors_instead_of_panicking`, which builds an empty schema and asserts `read_json_value_as_arrow(...).is_err()` — and it passes on `main` today. The PR description's claim that it adds a "zero-field schema returns `Err`" test is therefore inaccurate; the only genuinely new test here is `single_field_schema_reads_first_column` (a happy-path round-trip). The new code is correct and harmless, but it does not close a gap that is currently open.

---

---

phil-opp:

This looks superseded — recommend closing, or re-scoping to the refactor alone.

The panic was already fixed by #2897 (`91d5ecb1a`, landed 2026-08-11 15:30 UTC), and this branch was rebased onto it about five minutes later. So `gh pr diff` now shows it *deleting* the two existing `if batch.num_columns() == 0 { bail! }` guards at `apis/rust/node/src/daemon_connection/json_to_arrow.rs:39` and `:78` and re-adding them as a shared `first_column_data` helper — identical behaviour, identical error string.

The one net-new test, `single_field_schema_reads_first_column`, is a happy-path assertion that passes both before and after, so it cannot go RED. The actual zero-column regression test (`empty_field_schema_errors_instead_of_panicking`) is already on main.

The PR body still describes the panic as live and cites "2 passed (new)"; if the refactor is kept, that description should not land as the commit message.

phil-opp:
Confirmed — this is superseded. I verified against current `origin/main`: `read_from_json_with_schema` and `read_json_value_as_arrow` both already carry the `if batch.num_columns() == 0 { bail!("JSON record batch has no columns") }` guard (landed in #2897), and a zero-column regression test is already on main. So the panic this PR set out to fix is already fixed, and the diff now only reshapes the two existing guards into a shared `first_column_data` helper — identical behavior, identical error string. The remaining refactor is too minor to justify a standalone PR, so I'm closing this. Thanks for the catch.

---

PR #3117, opened 2026-08-11
fix(coordinator): persist ready-barrier release even when the status read fails

### LINKED ISSUE #3115: coordinator: failed ready-barrier release is not persisted when the store status read fails, re-opening the #2998 reconnect-hang window
## Summary

In `broadcast_all_nodes_ready` (`binaries/coordinator/src/lib.rs`), the ready-barrier release is recorded **in memory unconditionally** but **persisted only conditionally** on the failed-barrier path. When the status lookup used to build the persisted record fails, the durable write is skipped entirely — even though the in-memory `ready_barrier_released` flag has already been flipped to `true`. This re-opens exactly the reconnect-hang window that #2998 (PR #3013, commit `2ce3a97`) set out to close, and it directly contradicts the invariant stated in that function's own comment.

## Where

`binaries/coordinator/src/lib.rs`, introduced by commit `2ce3a97` / PR #3013 ("fix(coordinator): replay the ready barrier to a daemon that reconnects after it released").

`release_barrier_message` sets the in-memory flag unconditionally:

```rust
fn release_barrier_message(uuid, dataflow, clock) -> eyre::Result<Vec<u8>> {
    dataflow.ready_barrier_released = true;   // lib.rs:4582 — always
    all_nodes_ready_message(uuid, dataflow, clock)
}
```

But in `broadcast_all_nodes_ready` the persist is gated on resolving a status, and on the failure path a store read error (or `Ok(None)`) skips the write:

```rust
let message = release_barrier_message(uuid, dataflow, clock)?;   // lib.rs:4649 — flag now true

let persist_status = if dataflow.exited_before_subscribe.is_empty() {
    Some(StoreDataflowStatus::Running)
} else {
    match store.get_dataflow(&uuid) {
        Ok(Some(existing)) => Some(existing.status),
        Ok(None) => None,                                        // → skip persist
        Err(e) => {
            tracing::warn!(dataflow = %uuid, "cannot read status to persist failed barrier: {e}");
            None                                                 // → skip persist
        }
    }
};
if let Some(status) = persist_status
    && let Err(e) = dataflow.make_record(status).and_then(|r| store.put_dataflow(&r))
{
    tracing::warn!(dataflow = %uuid, "failed to persist ready-barrier release: {e}");
}
```

The comment immediately above this block states the intended invariant explicitly:

> This must happen on a *failed* barrier too. [...] deferring the write to whatever failure path runs next leaves a restart window that reproduces #2998. Only the status is conditional [...]

The code makes the *entire write* conditional, not just the status.

## Risk / affected code path

On a **failed** barrier (`exited_before_subscribe` non-empty), if `store.get_dataflow(&uuid)` returns `Err(_)` (transient store I/O error) the barrier release + verdict are never persisted, yet the in-memory entry already reports `ready_barrier_released = true`. If the in-memory `RunningDataflow` is then destroyed before any later status-persist captures the flag — coordinator restart, or orphan reclaim — a daemon that was disconnected when the broadcast fired reconnects, `reestablish_running_dataflow` reads `record.ready_barrier_released == false` from the (stale) persisted record, `replay_all_nodes_ready` is never called, and that daemon's nodes park in `init_from_env()` for the life of the dataflow. That is the #2998 symptom.

Reachability is **narrow / low severity**, to be clear:
- The `Ok(None)` branch is nearly unreachable in practice — the record is normally persisted as `Pending` at dataflow start, so a subsequent `None` would require the record to have been deleted.
- The realistic trigger is the `Err(_)` store-read branch (transient DB/disk error) coinciding with a failed barrier, a disconnected-then-reconnecting daemon, and entry destruction before the next status-persist re-captures the flag.

It is a genuine latent gap rather than a routinely-hit bug, but it undermines the durability guarantee the PR was specifically written to provide, on the one path (store I/O trouble) where durability matters most.

## Suggested fix

Decouple *persisting the release + verdict* from *resolving the status*. The `ready_barrier_released` flag and the `exited_before_subscribe` verdict must be written whenever the barrier fires; only the status field should stay conditional. When the current status cannot be read, persist the release with a non-promoting fallback status (i.e. never `Running` for a failed barrier) rather than skipping the write entirely — e.g. fall back to persisting with the record's last-known/`Pending` status, or capture the status to preserve at an earlier reliable point. The key property: a failed-barrier release should still leave `record.ready_barrier_released == true` on disk even when the status read fails.

---

*This issue was created by a scheduled, automated Claude code-review check (not a human). It reviewed the last few days of merged commits/PRs in this repo. The finding above was verified against the code, but please confirm before acting on it. The rest of the reviewed window (the daemon lifecycle fixes #2967/#2936/#2937, the memory-pool/input fixes #3011/#2935/#2968, the flume→tokio migration #2956, the uhlc 0.5→0.9 bump #2446, the runtime operator-payload fix #2742, and the CLI/CI/schema changes) reviewed clean.*

phil-opp:
## Summary

Fixes #3115.

`broadcast_all_nodes_ready` flips the in-memory `ready_barrier_released` flag **unconditionally** (via `release_barrier_message`) but, on the *failed*-barrier path, only persisted the durable record when it could resolve a status. A `store.get_dataflow` **error** (or `Ok(None)`) skipped the **entire write**, leaving the flag `true` in memory but stale on disk.

If the in-memory `RunningDataflow` is then destroyed (coordinator restart or orphan reclaim) before a later status-persist re-captures the flag, a daemon that was disconnected when the broadcast fired reconnects, reads `ready_barrier_released == false` from the stale record, is never replayed the release, and parks its nodes for the life of the dataflow — exactly the #2998 reconnect-hang window that #3013 set out to close, on the store-I/O-trouble path where durability matters most.

## Fix

Decouple *persisting the release + verdict* from *resolving the status*, as the issue suggested. Extract `persist_ready_barrier_release`, which:

- **always** writes the record (the `ready_barrier_released` flag + `exited_before_subscribe` verdict come from `make_record`), and
- makes only the **status** conditional. A successful barrier promotes to `Running`; a failed barrier preserves the record's current status when it can be read, and otherwise falls back to `Pending` — a non-promoting status the reconcile path can still advance — so it never promotes to `Running`.

The key property restored: a failed-barrier release leaves `record.ready_barrier_released == true` on disk **even when the status read fails**.

## Tests

New regression tests in `binaries/coordinator/src/lib.rs`:

- `failed_barrier_release_persists_when_status_read_fails` — a `FailingReadStore` whose `get_dataflow` always errors; asserts the release is still persisted (fails on the old code, which skipped the write).
- `failed_barrier_release_persists_when_record_absent` — the `Ok(None)` companion path.
- `successful_barrier_release_persists_running` — confirms a successful barrier still promotes to `Running`.

Full `dora-coordinator` suite passes (125 lib + 22 + 4); `cargo fmt --check` and `cargo clippy -- -D warnings` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PCXs8nkFNV7e4AhQLRVSSH)_


phil-opp:

Reviewed the diff — no issues found. Decoupling the (now unconditional) durable record write from the (conditional) status resolution correctly closes the reconnect-hang window on the store-read-error path — previously a read error or `Ok(None)` skipped the whole write, leaving the in-memory `ready_barrier_released` flag `true` but the durable record stale. The `Pending` fallback on a read error is safe, since a failed-barrier dataflow (non-empty `exited_before_subscribe`) never reached `Running`, so there is no more-advanced status to regress. `FailingReadStore` forces a `get_dataflow` error while still delegating writes and proves the release is persisted with `ready_barrier_released == true` and never promoted to `Running`; the `Ok(None)` and success paths are covered too.

---

phil-opp:
🤖 **Automated review by Claude** (fully automated, no human in the loop — treat as advisory).

I found one edge worth a look on the read-error fallback in `persist_ready_barrier_release` (the `Err(_) => StoreDataflowStatus::Pending` arm):

On a *successful* read this correctly preserves `existing.status`, so an on-disk terminal `Failed { terminal: true }` survives. But on a read *error* the code falls back to `Pending`, and `put_dataflow` rewrites the whole record via `make_record`, dropping the `terminal` marker. The reconcile path (`lib.rs:2577-2588`) then treats `Pending` as promotable and can flip it back to `Running` on the next daemon report — exactly the resurrection the `terminal` flag is meant to prevent (`coordinator-store/src/lib.rs:135-147`).

This is a different direction than the earlier reasoning that a barrier-failed dataflow "never reached `Running`, so there's no more-advanced status to regress": the hazard here isn't regressing *from `Running`*, it's clobbering a terminal `Failed` on the read-error path. It's arguably a regression from the old skip-the-write behavior, which left the terminal record intact. It is a narrow race (a terminal-`Failed` writer — the spawn-timeout watchdog, or `Recovering -> Failed` — landing just before the final `ReadyOnDaemon` triggers the broadcast, plus a transient store read error), so it's worth first confirming whether a terminal `Failed` can even be on disk at barrier-release time. If so, consider not lowering an unreadable status all the way to `Pending`. The `Ok(None)` branch is fine (nothing to regress).

Minor, separate note: the three new tests set `ready_barrier_released = true` by hand and call `persist_ready_barrier_release` directly, so they don't lock in the `release_barrier_message`-before-persist ordering in `broadcast_all_nodes_ready` — a reversal there would resurrect #3115 with the tests still green. The ordering is correct today, so this is only a coverage gap.

---

phil-opp:
Merged `main` (clean) and verified the fix. The two regression tests are genuine RED-without-it: restoring the pre-PR conditional inside `persist_ready_barrier_release` makes `failed_barrier_release_persists_when_status_read_fails` and `..._when_record_absent` both fail. 157 tests pass, clippy clean.

Addressed the one open item from the 2026-08-16 review — the `Err(_) => Pending` fallback rewrites the whole record, so an on-disk status is lowered to one the reconcile path can promote back to `Running`. The concern is real but narrower than blocking, and I found a *more* reachable variant than the `Failed { terminal: true }` the comment names: `dora stop` persists `Stopping` and deliberately keeps the in-memory entry alive, so a `ReadyOnDaemon` arriving in that window triggers a broadcast with a non-empty `exited_before_subscribe`; one store read error and the record goes `Stopping → Pending`, which *is* in the reconcile promote arm.

The trade is still right — a lost release parks a daemon's nodes forever, while a wrong status is re-derived from the next daemon report, and the success arm already writes `Running` with no read at all — so I documented it in the doc comment rather than changing behavior.

Also tightened the two assertions: `!matches!(status, Running)` would have passed if the fallback silently became `Recovering` or `Stopping`. They now assert `Pending`, which is what the doc comment promises.

The other item from that review — "the new tests don't lock in the `release_barrier_message`-before-persist ordering" — is already satisfied, just not by the new tests: the pre-existing `a_failed_barrier_is_persisted_as_durably_as_a_successful_one` drives the real `broadcast_all_nodes_ready` and fails if the two lines are swapped.

Ready to merge from my side.

PR #3118, opened 2026-08-11
fix(daemon): retry finished dataflow reports after reconnect

SunSunSun689:
# Summary
  Fixes a daemon cleanup ordering bug where finish_dataflow could leave a completed dataflow in the daemon’s local running map if reporting  AllNodesFinished to the coordinator failed during a disconnect.
  The fix makes local teardown independent from coordinator delivery:
  - signal listener shutdown and remove the dataflow from self.running before
    reporting completion

  - keep the existing error return when coordinator reporting fails, so
    reconnect behavior is preserved

  - store failed finish reports in pending_finished_dataflows
  - retry pending AllNodesFinished reports on the next coordinator connection
    before sending StatusReport

  This prevents the daemon from reporting an already-finished dataflow as still running after reconnect, while still giving the coordinator another chance to  receive the terminal result.

  # Notes

  This fixes the direct failure path where the coordinator send channel is  already closed and send_event() returns an error. It does not change the  broader fire-and-forget semantics of CoordinatorSender; if a message is  accepted into the local send channel but the background websocket writer later fails, there is still no coordinator-level ack.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -8222,6 +8222,13 @@ mod fault_tolerance_tests {
         let mut listener_shutdown = dataflow.listener_shutdown_tx.subscribe();
         daemon.running.insert(dataflow_id, dataflow);
 
+        #[cfg(feature = "tensor-pool")]
+        let subscriber = tokio::spawn(std::future::pending::<()>());
+        #[cfg(feature = "tensor-pool")]
+        daemon
+            .memory_pool_subscribers
+            .insert(dataflow_id, subscriber);
+
         let result = daemon.finish_dataflow(dataflow_id).await;
 
         assert!(
@@ -8241,6 +8248,11 @@ mod fault_tolerance_tests {
             .await
             .expect("finish_dataflow should signal listener shutdown");
         assert!(*listener_shutdown.borrow());
+        #[cfg(feature = "tensor-pool")]
+        assert!(
+            !daemon.memory_pool_subscribers.contains_key(&dataflow_id),
+            "memory-pool subscriber must be removed even when coordinator reporting fails"
+        );
 
         let (coordinator_sender, mut coordinator_rx) = coordinator::CoordinatorSender::for_test();
         daemon.coordinator_sender = Some(coordinator_sender);
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -8223,11 +8223,15 @@ mod fault_tolerance_tests {
         daemon.running.insert(dataflow_id, dataflow);
 
         #[cfg(feature = "tensor-pool")]
-        let subscriber = tokio::spawn(std::future::pending::<()>());
-        #[cfg(feature = "tensor-pool")]
-        daemon
-            .memory_pool_subscribers
-            .insert(dataflow_id, subscriber);
+        {
+            unsafe { std::env::set_var("DORA_MEMORY_POOL_CROSS_MACHINE", "1") };
+            daemon.pool_subscribe_dataflow(dataflow_id);
+            unsafe { std::env::remove_var("DORA_MEMORY_POOL_CROSS_MACHINE") };
+            assert!(
+                daemon.pool.has_subscriber(&dataflow_id),
+                "test setup must create a memory-pool subscriber"
+            );
+        }
 
         let result = daemon.finish_dataflow(dataflow_id).await;
 
@@ -8250,7 +8254,7 @@ mod fault_tolerance_tests {
         assert!(*listener_shutdown.borrow());
         #[cfg(feature = "tensor-pool")]
         assert!(
-            !daemon.memory_pool_subscribers.contains_key(&dataflow_id),
+            !daemon.pool.has_subscriber(&dataflow_id),
             "memory-pool subscriber must be removed even when coordinator reporting fails"
         );
 
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -433,6 +433,7 @@ pub struct Daemon {
     pub(crate) exit_when_done: Option<BTreeSet<(Uuid, NodeId)>>,
     pub(crate) exit_when_all_finished: bool,
     pub(crate) dataflow_node_results: BTreeMap<Uuid, BTreeMap<NodeId, Result<(), NodeError>>>,
+    pub(crate) pending_finished_dataflows: BTreeMap<Uuid, DataflowDaemonResult>,
     pub(crate) clock: Arc<uhlc::HLC>,
     pub(crate) ft_stats: Arc<FaultToleranceStats>,
     pub(crate) zenoh_session: zenoh::Session,
@@ -1933,6 +1934,7 @@ impl Daemon {
             exit_when_done,
             exit_when_all_finished: false,
             dataflow_node_results: BTreeMap::new(),
+            pending_finished_dataflows: BTreeMap::new(),
             warned_late_outputs: HashSet::new(),
             clock,
             ft_stats: Default::default(),
@@ -2031,6 +2033,9 @@ impl Daemon {
             .merge();
 
         // Send status report to coordinator so it can reconcile dataflow state.
+        if self.coordinator_sender.is_some() {
+            self.report_pending_finished_dataflows().await?;
+        }
         if let Some(sender) = &self.coordinator_sender {
             let running_dataflows: Vec<_> = self
                 .running
@@ -2356,6 +2361,48 @@ impl Daemon {
         // `run_inner` borrows `&mut self`, so move the accumulated results out
         // (the daemon may be reused for a reconnect, where these are ignored).
         Ok(std::mem::take(&mut self.dataflow_node_results))
+    }
+
+    async fn report_pending_finished_dataflows(&mut self) -> eyre::Result<()> {
+        let Some(sender) = &self.coordinator_sender else {
+            return Ok(());
+        };
+
+        let pending: Vec<_> = self
+            .pending_finished_dataflows
+            .iter()
+            .map(|(dataflow_id, result)| (*dataflow_id, result.clone()))
+            .collect();
+        for (dataflow_id, result) in pending {
+            self.send_all_nodes_finished(sender, dataflow_id, &result)
+                .await
+                .wrap_err("failed to retry dataflow finish report to dora-coordinator")?;
+            self.pending_finished_dataflows.remove(&dataflow_id);
+        }
+
+        Ok(())
+    }
+
+    async fn send_all_nodes_finished(
+        &self,
+        sender: &coordinator::CoordinatorSender,
+        dataflow_id: Uuid,
+        result: &DataflowDaemonResult,
+    ) -> eyre::Result<()> {
+        let msg = serde_json::to_vec(&Timestamped {
+            inner: CoordinatorRequest::Event {
+                daemon_id: self.daemon_id.clone(),
+                event: DaemonEvent::AllNodesFinished {
+                    dataflow_id,
+                    result: result.clone(),
+                },
+            },
+            timestamp: self.clock.new_timestamp(),
+        })?;
+        sender
+            .send_event(&msg)
+            .await
+            .wrap_err("failed to report dataflow finish to dora-coordinator")
     }
 
     /// Hold a destroy until this daemon's nodes are gone, or hand the reply
@@ -6002,22 +6049,6 @@ impl Daemon {
             )
             .await;
 
-        if let Some(sender) = &self.coordinator_sender {
-            let msg = serde_json::to_vec(&Timestamped {
-                inner: CoordinatorRequest::Event {
-                    daemon_id: self.daemon_id.clone(),
-                    event: DaemonEvent::AllNodesFinished {
-                        dataflow_id,
-                        result,
-                    },
-                },
-                timestamp: self.clock.new_timestamp(),
-            })?;
-            sender
-                .send_event(&msg)
-                .await
-                .wrap_err("failed to report dataflow finish to dora-coordinator")?;
-        }
         // Signal all listener loops for this dataflow to shut down
         if let Some(df) = self.running.get(&dataflow_id) {
             let _ = df.listener_shutdown_tx.send(true);
@@ -6030,6 +6061,16 @@ impl Daemon {
         // tasks and can create duplicate consumers.
         #[cfg(feature = "tensor-pool")]
         self.pool_cleanup_dataflow(dataflow_id).await;
+
+        if let Some(sender) = &self.coordinator_sender
+            && let Err(err) = self
+                .send_all_nodes_finished(sender, dataflow_id, &result)
+                .await
+        {
+            self.pending_finished_dataflows.insert(dataflow_id, result);
+            return Err(err);
+        }
+
         Ok(())
     }
 
@@ -8152,6 +8193,72 @@ mod fault_tolerance_tests {
         let mut df = test_dataflow();
         df.dataflow_started = true;
         assert!(!df.should_start_on_barrier_completion(&DataflowStatus::AllNodesReady));
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn finish_dataflow_cleans_local_state_when_coordinator_send_fails() {
+        let (coordinator_sender, coordinator_rx) = coordinator::CoordinatorSender::for_test();
+        drop(coordinator_rx);
+        let (mut daemon, _events_rx) = Daemon::build_daemon(
+            None,
+            Some(coordinator_sender),
+            DaemonId::new(None),
+            None,
+            Arc::new(HLC::default()),
+            None,
+            BTreeMap::new(),
+            LogDestination::Tracing,
+            None,
+            Vec::new(),
+            ZenohBind::Derived(LOCALHOST),
+            false,
+            false,
+        )
+        .await
+        .expect("daemon should build");
+
+        let dataflow_id = Uuid::new_v4();
+        let dataflow = test_dataflow();
+        let mut listener_shutdown = dataflow.listener_shutdown_tx.subscribe();
+        daemon.running.insert(dataflow_id, dataflow);
+
+        let result = daemon.finish_dataflow(dataflow_id).await;
+
+        assert!(
+            result.is_err(),
+            "closed coordinator sender should still report the send failure"
+        );
+        assert!(
+            !daemon.running.contains_key(&dataflow_id),
+            "finished dataflow must be removed locally even if coordinator reporting fails"
+        );
+        assert!(
+            daemon.pending_finished_dataflows.contains_key(&dataflow_id),
+            "failed finish report must be retained for reconnect retry"
+        );
+        listener_shutdown
+            .changed()
+            .await
+            .expect("finish_dataflow should signal listener shutdown");
+        assert!(*listener_shutdown.borrow());
+
+        let (coordinator_sender, mut coordinator_rx) = coordinator::CoordinatorSender::for_test();
+        daemon.coordinator_sender = Some(coordinator_sender);
+        daemon
+            .report_pending_finished_dataflows()
+            .await
+            .expect("retrying pending finish report should succeed");
+        assert!(
+            !daemon.pending_finished_dataflows.contains_key(&dataflow_id),
+            "pending finish report should be removed after successful retry"
+        );
+
+        let retried = coordinator_rx
+            .recv()
+            .await
+            .expect("retry should send an event to the coordinator");
+        assert!(retried.contains("AllNodesFinished"), "{retried}");
+        assert!(retried.contains(&dataflow_id.to_string()), "{retried}");
     }
 
     fn test_running_node() -> RunningNode {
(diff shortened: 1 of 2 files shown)


phil-opp:

Reviewed the diff — no issues found. The reordering so that local cleanup (removing the dataflow from `running` + signalling listener shutdown) happens *before* the coordinator report correctly decouples local cleanup from the send result. The reconnect retry path only re-sends entries that were actually stashed on a send failure and removes each entry only on success, so there's no duplicate-report or lost-entry risk. `finish_dataflow_cleans_local_state_when_coordinator_send_fails` is a real regression test: it drops the receiver to force a genuine send failure, then asserts error propagation, `running` cleared, the pending entry retained, listener shutdown fired, and a fresh sender draining the retried `AllNodesFinished` event.

One minor, non-blocking note: on reconnect, if a stashed report fails to send *again*, the `?` aborts the rest of the reconnect status report. That's a benign transient edge case, not a defect.

---

SunSunSun689:
Resolved the merge conflict by preserving both sides' intended behavior.

  The final `finish_dataflow` flow now:
  1. builds the `DataflowDaemonResult`
  2. performs the local cleanup first:
     - signals listener shutdown
     - removes the dataflow from `self.running`
  3. then attempts to send `AllNodesFinished` to the coordinator
  4. if that send fails, stores the result in `pending_finished_dataflows` so it
  can be retried after reconnect

  This keeps the newer local cleanup behavior from `main` while preserving this
  PR's retry-after-reconnect behavior.

phil-opp:
The Aug 14 merge-conflict resolution (`a50f2fe`, force-pushed today) looks safe. `finish_dataflow` now performs local cleanup first (signal listener shutdown + remove the dataflow from `self.running`), then sends `AllNodesFinished` to the coordinator, and on send failure stashes the `DataflowDaemonResult` into `pending_finished_dataflows` and returns `Err`. That keeps `main`'s newer local-cleanup ordering while preserving this PR's retry-after-reconnect behavior, with no duplicate-report or lost-entry risk (a dataflow is removed from `running` on finish, so it is reported exactly once — inline, or once via the reconnect retry drain). The `finish_dataflow_cleans_local_state_when_coordinator_send_fails` regression test still forces a real send failure and asserts error propagation, local cleanup, pending retention, and a successful retry drain.

No new issues.

---
🤖 This is a **fully automated** review by Claude (Claude Code) — no human reviewed it before posting. Treat it as advisory.

---

phil-opp:

Took another look after the 2026-08-18 merge with `main`, since the base picked up new teardown steps in `finish_dataflow` (`drain_cross_write_pending`, the `cross_data_endpoints` retain, etc.) that now interact with this PR's error path.

I found one issue worth addressing before merge: on the coordinator-send-failure path, `finish_dataflow` returns early, which skips the rest of the local teardown:

```rust

if let Some(sender) = &self.coordinator_sender
    && let Err(err) = self.send_all_nodes_finished(sender, dataflow_id, &result).await
{
    self.pending_finished_dataflows.insert(dataflow_id, result);
    return Err(err);   // everything below is skipped
}

// memory_pool_subscribers.remove(..).shutdown()
// CROSS_POOL_WRITE_LOCKS(_ASYNC) drains
// drain_cross_write_pending(dataflow_id)
// CROSS_WRITE_SEQ / CROSS_DIRECT_DEGRADED / cross_data_endpoints retains
// tensor_pool.cleanup_dataflow / cleanup_cross_pools
```

The `Err` case is exactly the disconnect this PR targets. When it fires, the dataflow is already removed from `self.running` (good) and its `dataflow_node_results` entry has been extracted, so `finish_dataflow` won't run again for this id — and the reconnect retry (`report_pending_finished_dataflows`) only re-sends the `AllNodesFinished` message; it does no reclamation. So the memory-pool subscriber task, the cross-machine tensor pools, and the write-sequence counters are leaked for the daemon's lifetime, and any in-flight cross-machine write hangs until the 120s safety-net timeout. The `CROSS_WRITE_SEQ` block's own comment calls out the monotonicity break and lifetime leak if it isn't drained.

Since the PR's goal is to decouple local teardown from coordinator delivery, it'd be good to complete that decoupling: run the full teardown regardless of the send result and only stash + return the error at the very end (e.g. move the send/stash block below `tensor_pool.cleanup_cross_pools`, or capture the send `Result` and defer the `return Err` until after reclamation). A test asserting the pools/subscriber are reclaimed on the send-failure path would lock this in — the current test drives `report_pending_finished_dataflows` directly and doesn't check reclamation.

The stash/retry logic itself looks correct: reports are stashed only on `Err` and removed only on a successful retry, so there's no duplicate or lost-report risk.

---

phil-opp:

The reordering correctly guarantees listener-shutdown and `running.remove` run even when the coordinator send fails, and the stash/retry path (`pending_finished_dataflows` → `report_pending_finished_dataflows`) is well tested.

One gap I found: in `finish_dataflow` the tensor-pool cleanup stays *after* the coordinator send and is now skipped by the new `return Err(err)` on send failure:

```rust

if let Some(sender) = &self.coordinator_sender
    && let Err(err) = self.send_all_nodes_finished(sender, dataflow_id, &result).await
{
    self.pending_finished_dataflows.insert(dataflow_id, result);
    return Err(err);   // <-- returns before pool_cleanup_dataflow
}

// The memory-pool subscriber task has no shutdown branch of its own — terminate it...
#[cfg(feature = "tensor-pool")]
self.pool_cleanup_dataflow(dataflow_id).await;
```

Because the dataflow is already removed from `running`, `finish_dataflow` won't run again for it, and the retry path `report_pending_finished_dataflows()` only re-sends the `AllNodesFinished` event — it never calls `pool_cleanup_dataflow`. So with the `tensor-pool` feature enabled, a coordinator-send failure at finish time permanently leaks that dataflow's memory-pool subscriber task, its zenoh session clone, and its event sender. That's the same "local cleanup skipped when the send fails" class of bug this PR sets out to fix, just missed for the pool subscriber.

Suggestion: move the `pool_cleanup_dataflow` call above the send (with the other local cleanup), or call it before the `return Err(err)`.

The added test is meaningful — it drives the real fail → stash → retry-succeeds path and asserts local state (running removal, listener shutdown, pending retention) — but it exercises a daemon built without the `tensor-pool` feature, so it wouldn't catch this leak.

---

SunSunSun689:
Addressed the latest automated review feedback.

The coordinator-send failure path in `finish_dataflow` no longer returns before local teardown. The function now completes the full local cleanup first, including:

- listener shutdown and `running` removal
- memory-pool subscriber removal/abort
- cross-write lock cleanup
- pending cross-write drain
- cross-write seq/degraded cleanup
- `cross_data_endpoints` cleanup
- tensor-pool dataflow/cross-pool cleanup

Only after that does it attempt to send `AllNodesFinished` to the coordinator. If the send still fails, the result is stashed in `pending_finished_dataflows` and the error is returned as before, so the reconnect retry behavior is preserved.

I also expanded `finish_dataflow_cleans_local_state_when_coordinator_send_fails` to cover the skipped-cleanup regression directly. The test now injects local memory-pool/cross-write state, forces the coordinator send failure, and asserts the local state is reclaimed while the failed report remains pending for retry.

Local validation run:

```text
cargo test -p dora-daemon finish_dataflow_cleans_local_state_when_coordinator_send_fails -- --nocapture
cargo test -p dora-daemon
cargo clippy -p dora-daemon -- -D warnings
cargo fmt --all -- --check
git diff --check
```

All passed locally.

phil-opp:
**Automated review by Claude — this is a fully automated review; no human has vetted it.**

Re-reviewed after `4377ed32`, the commit since the last review — it resolves the previously-reported blocker. The coordinator `AllNodesFinished` send is now the last step in `finish_dataflow`, after listener shutdown, `running.remove`, and `pool_cleanup_dataflow`, so a send failure no longer skips local teardown (which previously leaked the memory-pool subscriber): it stashes into `pending_finished_dataflows` and returns `Err` for the reconnect retry. The stash/retry logic is correct — stash only on `Err`, remove only on a successful retry, and the dataflow is already removed from `running`, so it's reported exactly once (no duplicate or lost report).

One nit: the new reclamation assertion (memory-pool subscriber removed on send failure) is under `#[cfg(feature = "tensor-pool")]`, so it only runs when that feature is enabled in the test build; the non-gated assertions (running removal, listener shutdown, pending retention, retry drain) do cover the core fix.

---

SunSunSun689:
Follow-up on the tensor-pool test coverage nit in `8a667c0db`.

I ran the targeted test with `--features tensor-pool`. That exposed a gated-test issue: the test still referenced the old `Daemon::memory_pool_subscribers` field, which no longer exists now that tensor-pool state lives under `daemon.pool`.

Updated the test to exercise the real path instead:

- create the subscriber through `daemon.pool_subscribe_dataflow(dataflow_id)`
- assert it exists through a new read-only `PoolState::has_subscriber`
- force the coordinator-send failure
- assert `finish_dataflow` removes the subscriber before stashing the failed report

Validation run locally with Rust 1.97.1 and `PYO3_PYTHON=/home/dora/.local/bin/python3.12`:

```text
cargo test -p dora-daemon --features tensor-pool finish_dataflow_cleans_local_state_when_coordinator_send_fails -- --nocapture
cargo clippy -p dora-daemon --features tensor-pool -- -D warnings
cargo fmt --all -- --check
cargo test -p dora-daemon finish_dataflow_cleans_local_state_when_coordinator_send_fails -- --nocapture
git diff --check
```

The `tensor-pool` target test now passes and the non-feature target test still passes.

phil-opp:

Re-reviewed after `8a667c0d` ("exercise tensor-pool finish cleanup"), the commit since the last review — a test-only change that addresses the earlier gating nit. Verified against the head: `finish_dataflow` runs all local teardown (listener shutdown, `running.remove`, and `pool_cleanup_dataflow`) before `send_all_nodes_finished`, which is now the final step; on send failure it stashes into `pending_finished_dataflows` and returns `Err` for the reconnect retry. The gated test now drives the real path (`pool_subscribe_dataflow` + the new read-only `has_subscriber`) and asserts the subscriber is reclaimed on the send-failure path, so the `#[cfg(feature = "tensor-pool")]` coverage is no longer vacuous. Stash-only-on-`Err` / remove-only-on-success retry logic remains correct. No new issues.

---

phil-opp:

Thanks for sticking with this through several rounds — the fix is in good shape and the test genuinely reproduces the bug. I confirmed the problem is still live on `main`: `finish_dataflow` at `binaries/daemon/src/lib.rs:6165` still does the fallible `send_event(...)?` before `self.running.remove(...)`, so a coordinator blip at exactly the moment a dataflow finishes strands it in `self.running` forever.

Two things before it can land.

**The branch is semantically stale even though GitHub reports MERGEABLE.** #3364 added two parameters to `Daemon::build_daemon` — `requested_listen_endpoint: Option<String>` and `zenoh_discovered_connect: Vec<String>` — so it now takes 15 arguments, and your test's 13-argument call will not compile on current `main`. The two sides touch different files, which is why the merge looks clean. A rebase passing `None` and `Vec::new()` should be all it needs; I checked the other helpers you use (`CoordinatorSender::for_test`, `test_dataflow`, `Daemon::pool`) and they are unchanged.

**Please drop the `?` on `report_pending_finished_dataflows()` in `run_inner` and log a warning instead.** As written, a retry that fails returns before the `StatusReport` and the `ZenohListenEndpoint` re-send just below it, and that endpoint confirmation is load-bearing for peer discovery after a coordinator restart (#3364). The neighbouring `StatusReport` send deliberately only warns, so mirroring it keeps the sequence intact:

```rust
if let Err(err) = self.report_pending_finished_dataflows().await {
    tracing::warn!("failed to retry pending dataflow finish reports: {err:#}");
}
```

Since the current code only removes entries on success, leaving the entry in place means the next connect picks it up anyway.

Everything else checks out. No duplicate reports (the dataflow is removed from `self.running` and its results extracted before the send, so `finish_dataflow` cannot run twice for the same id). No retry storm (`run_inner` is re-entered per reconnect and the retry drains once per connect). The ordering is right, since the retry runs before the `StatusReport` that would otherwise let the coordinator reconcile the dataflow as still-unknown. And a coordinator that restarted and never knew the dataflow just logs "dataflow not running on DataflowFinishedOnDaemon", which is harmless.

Smaller notes: the `if self.coordinator_sender.is_some()` guard is redundant since `report_pending_finished_dataflows` already returns early on `None`; `std::mem::take` plus re-insert on failure would avoid cloning every pending entry per retry; and the `unsafe { std::env::set_var(...) }` in the test is process-global, which Rust 2024 made unsafe precisely because it races with other threads reading the environment while `cargo test` runs them in parallel — there is no other `set_var` in `binaries/daemon/src/`, so driving `pool_subscribe_dataflow` directly or serializing behind a mutex would be safer. Also worth a comment that `tensor-pool` is not a default feature, so the gated reclamation assertions never run under CI's `cargo test --all`.

One heads-up on ordering: #3409 touched the same block and has since been closed as superseded by this PR, so nothing there should force you to rebase. It did also fix a batch abort in `trigger_manual_stop` that this PR does not cover — the first failing `finish_dataflow` returns `Err` and the remaining dataflows never get cleaned up — so we will reopen just that hunk as a follow-up on top of yours. No action needed from you.

phil-opp:

Thanks for sticking with this through several rounds — the fix is in good shape and the test genuinely reproduces the bug. I confirmed the problem is still live on `main`: `finish_dataflow` at `binaries/daemon/src/lib.rs:6165` still does the fallible `send_event(...)?` before `self.running.remove(...)`, so a coordinator blip at exactly the moment a dataflow finishes strands it in `self.running` forever.

Two things before it can land.

**The branch is semantically stale even though GitHub reports MERGEABLE.** #3364 added two parameters to `Daemon::build_daemon` — `requested_listen_endpoint: Option<String>` and `zenoh_discovered_connect: Vec<String>` — so it now takes 15 arguments, and your test's 13-argument call will not compile on current `main`. The two sides touch different files, which is why the merge looks clean. A rebase passing `None` and `Vec::new()` should be all it needs; I checked the other helpers you use (`CoordinatorSender::for_test`, `test_dataflow`, `Daemon::pool`) and they are unchanged.

**Please drop the `?` on `report_pending_finished_dataflows()` in `run_inner` and log a warning instead.** As written, a retry that fails returns before the `StatusReport` and the `ZenohListenEndpoint` re-send just below it, and that endpoint confirmation is load-bearing for peer discovery after a coordinator restart (#3364). The neighbouring `StatusReport` send deliberately only warns, so mirroring it keeps the sequence intact:

```rust
if let Err(err) = self.report_pending_finished_dataflows().await {
    tracing::warn!("failed to retry pending dataflow finish reports: {err:#}");
}
```

Since the current code only removes entries on success, leaving the entry in place means the next connect picks it up anyway.

Everything else checks out. No duplicate reports (the dataflow is removed from `self.running` and its results extracted before the send, so `finish_dataflow` cannot run twice for the same id). No retry storm (`run_inner` is re-entered per reconnect and the retry drains once per connect). The ordering is right, since the retry runs before the `StatusReport` that would otherwise let the coordinator reconcile the dataflow as still-unknown. And a coordinator that restarted and never knew the dataflow just logs "dataflow not running on DataflowFinishedOnDaemon", which is harmless.

Smaller notes: the `if self.coordinator_sender.is_some()` guard is redundant since `report_pending_finished_dataflows` already returns early on `None`; `std::mem::take` plus re-insert on failure would avoid cloning every pending entry per retry; and the `unsafe { std::env::set_var(...) }` in the test is process-global, which Rust 2024 made unsafe precisely because it races with other threads reading the environment while `cargo test` runs them in parallel — there is no other `set_var` in `binaries/daemon/src/`, so driving `pool_subscribe_dataflow` directly or serializing behind a mutex would be safer. Also worth a comment that `tensor-pool` is not a default feature, so the gated reclamation assertions never run under CI's `cargo test --all`.

One heads-up on ordering: #3409 touched the same block and has since been closed as superseded by this PR, so nothing there should force you to rebase. It did also fix a batch abort in `trigger_manual_stop` that this PR does not cover — the first failing `finish_dataflow` returns `Err` and the remaining dataflows never get cleaned up — so we will reopen just that hunk as a follow-up on top of yours. No action needed from you.

phil-opp:

Thanks for sticking with this through several rounds — the fix is in good shape and the test genuinely reproduces the bug. I confirmed the problem is still live on `main`: `finish_dataflow` at `binaries/daemon/src/lib.rs:6165` still does the fallible `send_event(...)?` before `self.running.remove(...)`, so a coordinator blip at exactly the moment a dataflow finishes strands it in `self.running` forever.

Two things before it can land.

**The branch is semantically stale even though GitHub reports MERGEABLE.** #3364 added two parameters to `Daemon::build_daemon` — `requested_listen_endpoint: Option<String>` and `zenoh_discovered_connect: Vec<String>` — so it now takes 15 arguments, and your test's 13-argument call will not compile on current `main`. The two sides touch different files, which is why the merge looks clean. A rebase passing `None` and `Vec::new()` should be all it needs; I checked the other helpers you use (`CoordinatorSender::for_test`, `test_dataflow`, `Daemon::pool`) and they are unchanged.

**Please drop the `?` on `report_pending_finished_dataflows()` in `run_inner` and log a warning instead.** As written, a retry that fails returns before the `StatusReport` and the `ZenohListenEndpoint` re-send just below it, and that endpoint confirmation is load-bearing for peer discovery after a coordinator restart (#3364). The neighbouring `StatusReport` send deliberately only warns, so mirroring it keeps the sequence intact:

```rust
if let Err(err) = self.report_pending_finished_dataflows().await {
    tracing::warn!("failed to retry pending dataflow finish reports: {err:#}");
}
```

Since the current code only removes entries on success, leaving the entry in place means the next connect picks it up anyway.

Everything else checks out. No duplicate reports (the dataflow is removed from `self.running` and its results extracted before the send, so `finish_dataflow` cannot run twice for the same id). No retry storm (`run_inner` is re-entered per reconnect and the retry drains once per connect). The ordering is right, since the retry runs before the `StatusReport` that would otherwise let the coordinator reconcile the dataflow as still-unknown. And a coordinator that restarted and never knew the dataflow just logs "dataflow not running on DataflowFinishedOnDaemon", which is harmless.

Smaller notes: the `if self.coordinator_sender.is_some()` guard is redundant since `report_pending_finished_dataflows` already returns early on `None`; `std::mem::take` plus re-insert on failure would avoid cloning every pending entry per retry; and the `unsafe { std::env::set_var(...) }` in the test is process-global, which Rust 2024 made unsafe precisely because it races with other threads reading the environment while `cargo test` runs them in parallel — there is no other `set_var` in `binaries/daemon/src/`, so driving `pool_subscribe_dataflow` directly or serializing behind a mutex would be safer. Also worth a comment that `tensor-pool` is not a default feature, so the gated reclamation assertions never run under CI's `cargo test --all`.

One heads-up on ordering: #3409 touched the same block and has since been closed as superseded by this PR, so nothing there should force you to rebase. It did also fix a batch abort in `trigger_manual_stop` that this PR does not cover — the first failing `finish_dataflow` returns `Err` and the remaining dataflows never get cleaned up — so we will reopen just that hunk as a follow-up on top of yours. No action needed from you.

phil-opp:

Brief note after the base-merge (`43ea1ca9`): the previously-reported compile blocker is now resolved — the test calls the current 15-argument `build_daemon` signature, and the removed coordinator-send block correctly targets `finish_dataflow` in its current home (`dora_events.rs`). The retry logic and its regression test are sound.

One substantive item from the earlier review is **still unfixed**: `run_inner` calls `self.report_pending_finished_dataflows().await?`, so a failed retry returns early — *before* the `StatusReport` and `ZenohListenEndpoint` re-send below it. That endpoint re-send is load-bearing for peer discovery after a coordinator restart, so a transient report failure shouldn't short-circuit it; log-and-warn would be safer than `?`. (The redundant `if self.coordinator_sender.is_some()` guard also remains.)

---

SunSunSun689:
Addressed the latest review note in `7a554d7b0`.

`run_inner` no longer lets a failed `report_pending_finished_dataflows()` retry abort the reconnect cycle before the `StatusReport` and `ZenohListenEndpoint` resend. It now logs a warning and leaves the pending finish report in place for the next reconnect, preserving the existing remove-only-on-success retry behavior. I also removed the redundant `coordinator_sender.is_some()` guard.

Added `failed_pending_finish_retry_does_not_abort_reconnect_cycle` to cover the regression directly. I verified the new test fails before the code change with the old early-return behavior, then passes after the fix.

Validation run locally:

```text
cargo test -p dora-daemon failed_pending_finish_retry_does_not_abort_reconnect_cycle -- --nocapture
cargo test -p dora-daemon finish_dataflow_cleans_local_state_when_coordinator_send_fails -- --nocapture
cargo fmt --all -- --check
cargo clippy -p dora-daemon -- -D warnings
git diff --check
```

All passed locally.

phil-opp:
🤖 **Automated review (Claude Code)** — fully automated; no human vetted this. Advisory only, not an approval; I'm posting like an outside contributor and can't approve or merge.

Confirmed the one open item from the last pass is fixed at `7a554d7b`: `run_inner` now logs-and-continues instead of `report_pending_finished_dataflows().await?`, so a failed retry no longer short-circuits the `StatusReport` / `ZenohListenEndpoint` re-send below it, and the redundant `coordinator_sender.is_some()` guard is gone. The new `failed_pending_finish_retry_does_not_abort_reconnect_cycle` test is a genuine regression test — it drives `run_inner` with a closed sender and asserts `Ok`, which would have failed on the old `?` path that returned `Err`, and it also checks the report stays pending for the next reconnect. Stash-only-on-`Err` / remove-only-on-success is intact. No new issues at this head.

---

phil-opp:
Thanks!

PR #3119, opened 2026-08-11
fix(coordinator): stop succeeded dataflow orphans on status report

SunSunSun689:
## Summary

  Fixes a coordinator reconciliation gap where a daemon could report a dataflow  as still running even though the coordinator store already marked it as  `Succeeded`.

  Previously, the `DaemonStatusReport` handler handled:

  - `Pending` / `Recovering` / non-terminal `Failed`: reconcile back to  `Running`
  - `Running`: re-establish live coordinator state
  - terminal `Failed`: preserve the terminal verdict and stop orphaned daemon-  side nodes
  - `Succeeded`: fell through to `_ => {}`  after the coordinator had already recorded the dataflow as successfully  finished.

  This change treats `Succeeded` as a terminal state for status-report reconciliation and reuses the existing orphan cleanup path.

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2623,13 +2623,13 @@ async fn start_inner(
                                     }
                                 }
                             }
-                            StoreDataflowStatus::Failed { terminal: true, .. } => {
-                                // Terminal failure (watchdog or equivalent
-                                // coordinator-side verdict). Daemon's report
-                                // is ignored to preserve the verdict the
-                                // user already received via wait_for_spawn.
+                            status if status_report_should_stop_orphan(&status) => {
+                                // Terminal state (success, watchdog failure, or
+                                // equivalent coordinator-side verdict). Daemon's
+                                // report is ignored to preserve the verdict the
+                                // user already received.
                                 tracing::warn!(
-                                    "daemon {daemon_id} reports terminally-failed \
+                                    "daemon {daemon_id} reports terminal \
                                      dataflow {df_id} as running; skipping reconcile",
                                 );
                                 // Stop the orphaned nodes the daemon kept alive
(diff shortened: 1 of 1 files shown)


phil-opp:
Reviewed the diff — no issues found, looks correct and safe to merge.

The new `status if status_report_should_stop_orphan(&status)` guard arm in the `DaemonStatusReport` reconciliation is only reachable for `Failed { terminal: true }` and `Succeeded`, because the earlier arms already consume `Failed { terminal: false } | Pending | Recovering` and `Running` — so arm ordering is sound and `Stopping` still falls through to the unchanged `_ => {}`. `Succeeded` is a genuinely terminal state (set only once all node results are collected error-free), so stopping a daemon that still reports such a dataflow as running is the right call and consistent with the existing terminal-failure orphan cleanup. The added unit test directly asserts the helper returns `true` for `Succeeded`, exercising the fixed path.

---

---

phil-opp:
The logic checks out on every axis I looked at: arm ordering is sound (`Failed{terminal:false} | Pending | Recovering` and `Running` are consumed earlier, `Stopping` still falls through), there is no cross-dataflow kill (ids are fresh UUIDv7 per start, so a restarted dataflow with the same *name* has a different id), it is not racy against a finishing node (`Succeeded` is only persisted once `dataflow.daemons.is_empty()`), and the fire-and-forget send matches its two sibling call sites, so no deadlock.

Two notes rather than blockers:

1. **Reachability is narrower than the description suggests.** Every finished dataflow is inserted into `archived_dataflows`, and the pre-check at `binaries/coordinator/src/lib.rs:2504-2521` already orphan-stops anything archived. This new arm therefore only fires when the in-memory archive is gone — after a coordinator restart (startup recovery deliberately does not reload `Succeeded` rows) or after `MAX_ARCHIVED_DATAFLOWS` eviction. A real gap, worth stating precisely.

2. **The test is a tautology.** It asserts the helper this PR introduces, so it cannot fail before the fix, and it exercises neither the arm ordering nor the negative cases. Adding `assert!(!status_report_should_stop_orphan(&Running))` plus `Pending` / `Recovering` / `Failed{terminal:false}` / `Stopping` would pin the property that actually regresses if the arm is ever moved.

PR #3121, opened 2026-08-11
chore: bump workspace version to 1.0.0 for the stable release

haixuanTao:
Rebased onto current `main` and brought up to date for the GA tag (#1626). The branch was cut at `1.0.0-rc.4` in August and had drifted: it conflicted with `main`, and its publish fixes had since landed elsewhere.

**The bump.** `[workspace.package].version` plus the 29 exact intra-workspace pins in `Cargo.toml`, and the 89 matching workspace entries in `Cargo.lock` — now from `1.0.0-rc.5` rather than `-rc.4`. Version-string only; the lock diff is `version = ` lines and nothing else. `release.yml`'s `check-version` job fails a tag run on tag/version skew, so this has to be on `main` before `v1.0.0` is tagged.

**The Changelog.** `Unreleased` becomes `v1.0.0 (2026-09-02)`, and the six breaking changes of the rc window that never reached the Changelog are added: the `_unstable_` prefix drop (#3220), Arrow leaving the frozen `dora-node-api` surface (#3213), the narrowed `dora_core` re-export (#3221), the tensor-pool transport moving behind the extension seam (#3219), the removal of the deprecated descriptor surface (#3158), and the wire-protocol enums going `#[non_exhaustive]` (#3151). The mixed-version warning is also sharpened — "rejected at register with a clear error" only holds between two postcard peers; an rc.4-or-earlier peer speaks bincode, so its register frame fails to decode before the version check runs, and `VersionReq::parse("1.0.0-rc.4")` is `^1.0.0-rc.4`, which matches `1.0.0`.

**Dropped in the rebase.** Publishing `dora-tensor-pool` and adding it to both publish lists landed separately in #3317, and the earlier `Unreleased` module bullet was superseded on `main` by #3281's. `main`'s wording is kept in both cases.

Verified at `1.0.0`: `cargo check --all --locked`, `cargo clippy --all -- -D warnings`, `cargo check --examples`, `cargo fmt --check`, `typos`, `make qa-publish-graph`, and the two version-sensitive crates' tests (`dora-message`, `dora-core`). Nothing in the tree keys off the `-rc.N` suffix — the only occurrences outside the manifests are usage examples in `scripts/release/verify-release.py` and `CONTRIBUTING.md`, and that script's `pep440()` maps a suffix-free `1.0.0` to `1.0.0`, so `verify-release` reads back the right PyPI version. `softprops/action-gh-release` auto-detects the semver prerelease suffix — which is why rc.5 came out `prerelease: true` with no input passed — so a bare `v1.0.0` publishes as a full release with no workflow change.


phil-opp:

No issues found.

The diff is a clean workspace version bump from `1.0.0-rc.4` to `1.0.0`:
- `Cargo.toml`: `workspace.package.version` and every `workspace.dependencies` path entry bumped consistently to `1.0.0`.
- `Cargo.lock`: all workspace crates updated `1.0.0-rc.4` → `1.0.0`, matching the manifest.
- `Changelog.md`: `## Unreleased` retitled to `## v1.0.0 (2026-08-11)`.

No code, dependency-source, or behavioral changes are present. Versions are internally consistent and the changelog date matches the release.

---

phil-opp:

The bump within the diff itself is internally consistent (`Cargo.toml` workspace version + every workspace-dependency entry + `Cargo.lock` all go `1.0.0-rc.4` → `1.0.0`, and the two independently-pinned crates — `dora-operator-api-types` at `0.1.13` and the `examples/error-propagation/*` samples at `0.1.0` — are intentionally excluded). But this branch is stale against current `main`, and the staleness produces a missed version reference:

- Since this branch's base, `#3152` replaced the `dora-memory-pool` workspace dependency with `dora-tensor-pool` + `dora-tensor-pool-python`. On current `main`, `Cargo.toml:140-141` declares both at `1.0.0-rc.4`, and `Cargo.lock` has matching `dora-tensor-pool` / `dora-tensor-pool-python` entries at `1.0.0-rc.4`.
- This PR's diff instead bumps a `dora-memory-pool` entry that no longer exists on `main`. When rebased/merged, `dora-tensor-pool` and `dora-tensor-pool-python` (and their `Cargo.lock` entries) will be left at `1.0.0-rc.4` while everything else is `1.0.0` — i.e., a 1.0.0 release shipping two workspace crates still at an rc version.

Recommend rebasing onto current `main` and re-running the bump so the two `dora-tensor-pool*` entries (manifest + lockfile) are included. No issue with the version-bump logic otherwise.

---

phil-opp:

Re-checked after the 2026-08-18 merge of `main` into this branch. The issue flagged on the earlier revision — `dora-tensor-pool` and `dora-tensor-pool-python` left at `1.0.0-rc.4` after #3152 renamed `dora-memory-pool` — is now resolved: both crates are bumped to `1.0.0` in `Cargo.toml` (`workspace.dependencies`) and in `Cargo.lock`, and no `dora-memory-pool` entry remains.

The bump is now internally consistent: `[workspace.package] version`, all 26 `workspace.dependencies` pins, and every workspace-member entry in `Cargo.lock` are at `1.0.0`. The remaining `1.0.0-rc` strings in the tree are all non-live references (doc comments, a legacy JSON test fixture, `smoke-all.sh` error text, and `release.yml`/`docs` tag-pattern documentation). No further version-consistency concerns.

---

phil-opp:
Merged `main` and resolved the `Cargo.toml` conflict: `main` had switched the intra-workspace deps to exact pins (`=1.0.0-rc.4`, #3233) and added four new entries, so I took that structure wholesale and re-applied the bump — `=1.0.0` on all 29, plus `[workspace.package] version = "1.0.0"`. `Cargo.lock` has no `rc.4` left and `cargo check --workspace --locked` passes.

**Do not push the `v1.0.0` tag before the publish fix in this PR lands.** `dora-tensor-pool` was `publish = false`, while `dora-daemon` — which *is* in both publish lists — carries `dora-tensor-pool = { workspace = true, optional = true }`, now pinned `=1.0.0`. `cargo publish` resolves every path dependency against crates.io, optional ones included, and `--no-verify` does not skip that resolution. Confirmed two ways:

- a scratch crate whose only dep is an optional `dora-tensor-pool = "=1.0.0"` fails `cargo package --no-verify` with `no matching package named 'dora-tensor-pool' found`;
- `dora-tensor-pool` does not exist on crates.io at any version. Its predecessor `dora-memory-pool` was published through `1.0.0-rc.4`; #3152 renamed it and marked it unpublished, so the rc.4 release never exercised this path.

The run would abort at `dora-daemon`, position 17 of 27, with the 16 crates before it already live and unrecoverable — crates.io versions can be yanked but never replaced. Fixed by publishing `dora-tensor-pool` (it stays non-public-API) and ordering it before `dora-daemon` in both lists.

Two more things fixed here:

- The heading was `## v1.0.0 (2026-08-11)`, but the tag has not been pushed and two entries already under it were merged after that date. Renaming `## Unreleased` also freezes everything under it as shipped — and six breaking changes merged since the heading was written had no entry at all: #3220 (the `_unstable_deploy`/`_unstable_debug` rename, which breaks existing YAML), #3213, #3221, #3219, #3158, #3151. Re-dated and backfilled.
- The mixed-version bullet claimed an rc.4 peer is "rejected at register with a clear version-mismatch error". It isn't. `VersionReq::parse("1.0.0-rc.4")` yields `^1.0.0-rc.4`, which matches `1.0.0`, so the semver gate passes; and rc.4 predates the bincode→postcard switch, so its register frame fails to decode *before* the `metadata_version` check runs. The daemon logs the error at `info` and drops the connection with no reply; the node sees a 30s register timeout. Corrected to describe what actually happens.

Left for a follow-up, not fixed here:

- `DaemonRegisterRequest` has no wire-format gate at all — that leg is JSON, so an rc.4 daemon's `Register` decodes fine and `versions_compatible` admits it, after which it speaks bincode on every binary plane. `NodeRegisterRequest` has `metadata_version` for exactly this; the daemon leg needs the same.
- `dora-mavlink2-bridge` and `dora-mavlink2-bridge-node` are publishable but in neither publish list, so they silently never reach crates.io. Either mark them `publish = false` (as `dora-ros2-bridge-node` is) or add them.
- `binaries/cli/pyproject.toml` still floors `dora-rs >= 0.3.9`, and the `dora new` Python template floors `dora-rs >= 0.2.0`. Both are wrong across a hard wire break.

phil-opp:

Re-reviewed the changes added since the last automated pass — the `dora-tensor-pool` publish fix and the changelog re-date/backfill. No new issues found. The publish ordering is correct in both `cargo-release.yml` and `release.yml` (`dora-tensor-pool` sits after `dora-coordinator` and before `dora-daemon`, and its own workspace deps `dora-message`/`dora-core` are published earlier, so its verify build resolves); removing `publish = false` is the right fix for `dora-daemon`'s optional `=1.0.0` path dep, and the crate still inherits `version.workspace = true`. The version bump stays internally consistent and the only non-version change is release mechanics.

---

phil-opp:

No issues found. The version bump is consistent across the workspace — `[workspace.package].version`, every `=1.0.0-rc.4` dependency pin in `[workspace.dependencies]`, and all matching `Cargo.lock` entries move to `1.0.0` in lockstep, and no code changes are bundled in beyond the version/release plumbing and Changelog backfill.

The one non-obvious change checks out: making `dora-tensor-pool` publishable (dropping `publish = false`) and ordering it before `dora-daemon` in both publish workflows is correct and necessary — `dora-daemon` carries an optional `=1.0.0` path dep on it, and `cargo publish` resolves optional path deps against crates.io, so leaving it unpublished would abort the release run at `dora-daemon` with the earlier crates already live.

---

phil-opp:
🤖 **Automated review by Claude** — fully automated, no human in the loop; verify before relying on it.

Re-checked after the rebase onto the newer base (the bump is now `1.0.0-rc.5` → `1.0.0`, and the Changelog heading is re-dated to `v1.0.0 (2026-09-02)`). Still internally consistent: `[workspace.package].version`, every `=1.0.0-rc.5` pin in `[workspace.dependencies]`, and all matching `Cargo.lock` workspace entries move to `1.0.0` together — no stray rc string left in the manifests, and no code/behavior changes bundled in. No new issues.

---

PR #3122, opened 2026-08-11
Fix module node env propagation

phil-opp:


phil-opp:

No issues found.

Reviewed the change to `libraries/core/src/descriptor/expand.rs`. `propagate_module_node_env` gap-fills an inner node's `env` from the referencing module node's `env` via `entry().or_insert_with()`, so the inner node's own entries win on conflict — consistent with `merge_env` ("per-node entries override global ones") and the adjacent `deploy` propagation. The empty/`None` module-env early return correctly avoids turning an inner `None` into `Some({})`. Because it runs for every inner node (including nested module-type ones) before Phase-2 recursion, the module env — and the `PARAM_*` entries later injected by `substitute_params_in_node` — propagate down through nested modules, mirroring `deploy`/`build`. Ordering relative to `substitute_params_in_node` is safe (`PARAM_*` is a separate key namespace).

The three added tests assert meaningful behavior and would fail if the logic were wrong: inner-wins-on-conflict (`expand_module_node_env_propagates_to_inner_nodes`), empty-env-leaves-`None` (`expand_module_node_with_empty_env_leaves_inner_env_unset`), and outer params reaching nested-module inner nodes as env (`expand_outer_params_reach_nested_module_inner_nodes_as_env`).

---

PR #3127, opened 2026-08-12
fix(cli): resolve `dora topic <node-id>` to all of a node's outputs

phil-opp:
## Issue

`dora topic echo/hz/info <node-id>` — passing a **bare node id**, which is meant to select *every output of that node* — always fails with a confusing error:

```
Invalid output id `camera/`: identifier must not be empty
```

`TopicSelector::resolve_with_descriptor` appended a trailing `/` to any argument without a slash and then relied on `InputMapping` parsing to yield an *empty* output id, dispatching through the `user.output.is_empty()` branch to expand to all outputs. But `InputMapping::from_str` (`libraries/message/src/config.rs`) splits on the first `/` and parses the remainder as a `DataId`, whose validator (`libraries/message/src/id.rs`) rejects the empty string. So `"camera/".parse::<InputMapping>()` returns `Err`, control falls to `Err(e) => bail!("Invalid output id ...")`, and the `is_empty()` branch was **unreachable dead code**.

## Fix

Handle the bare-node case explicitly, *before* touching `InputMapping` parsing: a slash-less selector is parsed as a `NodeId` and expanded to every declared output. The fragile append-`/` hack is removed. The topic-resolution logic is extracted into a pure `resolve_topics(data, descriptor)` function so it can be unit-tested without a live coordinator session, and the "unknown node" hint is deduplicated into a small closure shared by both the bare-node and qualified-output paths.

## Validation

- New regression tests (5) cover bare-node expansion, qualified-output selection, the empty-selector case, and the unknown-node / unknown-output error hints — all passing (`cargo test -p dora-cli --lib command::topic::selector`).
- `cargo test -p dora-cli --lib` green (325 tests); `cargo fmt --all -- --check` and `cargo clippy -p dora-cli -- -D warnings` clean; whole-workspace `cargo check --all` clean.
- Verified the explicit-trailing-slash case (`dora topic camera/`) still errors as before — no behavior regression.

---

---

### DIFF
--- a/binaries/cli/src/command/topic/selector.rs
+++ b/binaries/cli/src/command/topic/selector.rs
@@ -1,5 +1,4 @@
 use std::{
-    borrow::Cow,
     collections::{BTreeSet, HashMap},
     fmt,
 };
@@ -54,7 +53,7 @@ pub struct TopicSelector {
     pub data: Vec<String>,
 }
 
-#[derive(Clone, PartialOrd, Ord, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq)]
 pub struct TopicIdentifier {
     pub node_id: NodeId,
     pub data_id: DataId,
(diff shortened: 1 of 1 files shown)


phil-opp:

This is the summary for an automated review pass over the repository (branch prefix `claude/dreamy-bardeen-568p4j-*`). Each finding was verified against the current `origin/main` and opened as its own single-file draft PR. Prior automated runs' open PRs were checked to avoid duplicates.

### PRs opened, by category

**Correctness**
- #3127 — `fix(cli)`: `dora topic <node-id>` (bare node id) always errored (`Invalid output id \`camera/\``) because the "all outputs of a node" branch was unreachable dead code; handled explicitly + regression tests.

**Optimization**
- #3128 — `perf(node)`: coalesce the node→daemon TCP request header+body into one write (avoids a runt TCP segment under `TCP_NODELAY`; mirrors the daemon-side send).
- #3129 — `perf(daemon)`: build the `OutputId` routing key once per message in `send_out` instead of cloning `node_id`/`output_id` to build it twice (−2 `String` allocs/msg on the daemon path).

**Simplification**
- #3130 — `refactor(core)`: dedupe the type-parameter agreement rule (duplicated verbatim in `types_match` and `CompatibilityGraph::is_compatible`) into one `params_agree` helper.

**Documentation**
- #3131 — `docs(message)`: document the `Parameter` type-mismatch contract with a compile-checked doctest + per-variant/constructor docs.
- #3132 — `docs(operator)`: document the public unsafe FFI entry points in `raw.rs` (module doc + `# Safety` sections).

- **perf — node `zenoh_publishers` double hashmap lookup** (`node/mod.rs`): the map holds one entry per output, so the absolute cost of the redundant lookup is negligible; not worth a PR.
- **docs — `get_python_path` / `get_uv_path` `//`→`///`** (`core/src/lib.rs`): promoting non-doc comments to rustdoc is low-value; skipped in favor of higher-value doc work.
- **correctness — several suspected panics/underflows**: `extract_err_from_stderr.rs` `saturating_sub`, `read_last_n_lines` backward read, `topic hz/info` `Instant - Duration` (guarded with `checked_sub`), and timer/`ByteSize` parsing — all verified to be already guarded and regression-tested. No change.
- **dead code**: candidates `is_known_arrow_type`, `current_crate_version`, `urn_short_name`, `send_service_response`, `Event::InputRecovered`, `EncodedSample::as_bytes`, etc. — all verified still referenced in production/tests. Not dead.
- **reuse — node/daemon TCP framing duplication** (raised by `/review` on #3128): the two length-prefix implementations live in separate crates (a lib vs a binary), so there is no importable shared helper; consolidating into `dora-message` is a larger cross-crate change out of scope for that fix.

### Validation

Per-branch: `cargo fmt --all --check` + `cargo clippy -- -D warnings` + tests for the affected crate. On the combined diff, once: `/review` (no correctness bugs) and `/simplify` (clean on all four angles). Affected-crate test results — cli 325, node-api 208, daemon 173, message 280 + 9 doctests, core 146 + doctests, operator + doctest — all green.

> **Note on the full-workspace suite:** a full `cargo test --all` build exceeds this ephemeral environment's fixed disk allowance (it ran the disk to 100% and failed with linker bus errors during unrelated heavy targets, not test failures). It was substituted with full test runs of every affected crate plus a whole-workspace `cargo check --all` and `cargo check --examples` (both clean), which together cover behavior for the changed crates and cross-crate compilation integrity.

Each PR is a draft and clearly marked machine-generated; a human maintainer should review before merging.

---

phil-opp:

No issues found. I traced the fix and the tests in `binaries/cli/src/command/topic/selector.rs`.

The root cause is correctly diagnosed: the old code appended `/` to a slash-less argument and routed it through `InputMapping` parsing, whose `DataId` validator rejects the empty tail, so the `user.output.is_empty()` "all outputs of a node" branch was unreachable — a bare node id always errored. The new `resolve_topics` handles the no-`/` case before touching `InputMapping`, parsing the argument as a `NodeId`, looking it up, and expanding to `node.outputs`. The qualified `node/output` path preserves the prior semantics (unknown-node context error, membership check against `node.outputs`, `bail!` on unknown output / reserved mapping / parse error), and the empty-selector and final "no outputs found" branches are unchanged. I did not find a behavioral regression.

The regression tests are meaningful (not vacuous): each asserts a concrete expected `BTreeSet` or a specific error substring against a descriptor where `camera` declares `frame`+`status` and `sink` declares none — so `empty_selector_returns_every_output` and `bare_node_id_expands_to_all_outputs` genuinely exercise the expansion, and the two error tests use `{err:#}` to check the hint text. The YAML is parsed with `.expect`, so a descriptor-schema mismatch would fail the test rather than pass silently.

Minor, non-blocking: an explicitly malformed argument containing a slash (e.g. a typed trailing-slash `camera/`) still falls into the `InputMapping` path and surfaces `Invalid output id \`camera/\``; that matches the prior behavior for such input and is a reasonable edge case, not a regression.

---

phil-opp:

No issues found.

I traced the change in `binaries/cli/src/command/topic/selector.rs`. The root cause is diagnosed correctly: the old code appended `/` to a slash-less argument and routed it through `InputMapping` parsing, whose `DataId` validator rejects the empty tail, so the `user.output.is_empty()` "all outputs of a node" branch was unreachable — a bare node id always errored. The new `resolve_topics` handles the no-`/` case first, parsing the argument as a `NodeId`, looking it up, and expanding to `node.outputs`; the qualified `node/output` path preserves the prior semantics (unknown-node hint, membership check, `bail!` on unknown output / reserved mapping / parse error). The empty-selector early return and the final "no outputs found" bail are unchanged from the original. No behavioral regression.

The five regression tests are meaningful (not vacuous): each asserts a concrete `BTreeSet` or a specific error substring against a descriptor where `camera` declares `frame`+`status`, so `bare_node_id_expands_to_all_outputs` and `empty_selector_returns_every_output` genuinely exercise the expansion, and the error tests check the hint text via `{err:#}`. The descriptor YAML is parsed with `.expect`, so a schema mismatch would fail rather than pass silently.

---

PR #3128, opened 2026-08-12
perf(node): send daemon TCP request header+body in a single write

phil-opp:
## Issue

On the node side of the node↔daemon TCP transport, `tcp_send` (`apis/rust/node/src/daemon_connection/tcp.rs`) wrote the 8-byte length header and the payload with two separate `write_all` calls before flushing:

```rust
connection.write_all(&len_raw)?;
connection.write_all(message)?;
connection.flush()?;
```

The connection has `TCP_NODELAY` enabled (`daemon_connection::connect`), so the header is flushed as its own tiny TCP segment ahead of the body — an extra syscall and a runt segment on **every** request. This path runs once per `DoraNode::send_output_sample` (both the daemon `send_message` and the zenoh `report_output_sent` control sync).

## Fix

Coalesce the header and payload into a single buffer and write it once — exactly the fix already applied on the daemon side in `binaries/daemon/src/socket_stream_utils.rs::socket_stream_send`. The wire format (8-byte little-endian length prefix + body) is byte-for-byte unchanged.

The payload here is small: messages larger than 4 KB travel via shared memory, so this control path carries only small frames, and the single copy is negligible while removing the extra syscall + runt segment.

## Validation

- Existing framing round-trip tests (`framing_roundtrip`, `empty_payload_roundtrips`, and the oversize/truncation guards) continue to pass — they assert the exact `8 + payload.len()` wire layout.
- `cargo test -p dora-node-api --lib` green (208 tests); `cargo clippy -p dora-node-api -- -D warnings` and `cargo fmt --all -- --check` clean; whole-workspace `cargo check --all` clean.

---

---

### DIFF
--- a/apis/rust/node/src/daemon_connection/tcp.rs
+++ b/apis/rust/node/src/daemon_connection/tcp.rs
@@ -77,9 +77,17 @@ fn tcp_send(connection: &mut (impl Write + Unpin), message: &[u8]) -> std::io::R
             ),
         ));
     }
+    // Concatenate the 8-byte length header and the payload into a single buffer
+    // and write it in one call. The connection has `TCP_NODELAY` enabled (see
+    // `daemon_connection::connect`), so writing the header separately would flush
+    // it as its own tiny TCP segment before the body — an extra syscall and a
+    // runt segment on every request. This mirrors the daemon-side send
+    // (`socket_stream_utils::socket_stream_send`). The wire format is unchanged.
     let len_raw = (message.len() as u64).to_le_bytes();
-    connection.write_all(&len_raw)?;
-    connection.write_all(message)?;
+    let mut buf = Vec::with_capacity(8 + message.len());
+    buf.extend_from_slice(&len_raw);
+    buf.extend_from_slice(message);
+    connection.write_all(&buf)?;
     connection.flush()?;
     Ok(())
 }
--- a/apis/rust/node/src/daemon_connection/tcp.rs
+++ b/apis/rust/node/src/daemon_connection/tcp.rs
@@ -66,6 +66,12 @@ fn receive_reply(
     }
 }
 
+/// Payload size (excluding the 8-byte header) at or below which `tcp_send`
+/// coalesces the header and body into one buffer. Chosen to cover normal
+/// control messages and small data outputs while keeping the extra copy
+/// trivial; larger payloads are written header-then-body without copying.
+const COALESCE_THRESHOLD: usize = 16 * 1024;
+
 fn tcp_send(connection: &mut (impl Write + Unpin), message: &[u8]) -> std::io::Result<()> {
     if message.len() > dora_message::MAX_MESSAGE_BYTES {
         return Err(std::io::Error::new(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The change in `apis/rust/node/src/daemon_connection/tcp.rs` writes the exact same bytes on the wire — the 8-byte little-endian length prefix followed by the payload — but assembles them into one buffer and issues a single `write_all` instead of two. The receiver's framing (read length, then read that many body bytes) is unaffected, so the wire format is genuinely unchanged and there is no correctness regression.

The only tradeoff is one heap allocation (`Vec::with_capacity(8 + message.len())`) per send in exchange for eliminating the extra syscall and the runt TCP_NODELAY segment; that is a reasonable perf call and matches the daemon-side `socket_stream_send` pattern referenced in the comment.

---

phil-opp:

No issues found. `tcp_send` now assembles the 8-byte little-endian length prefix and the payload into one buffer and issues a single `write_all`, replacing two separate writes. The bytes on the wire are identical, so the receiver's length-then-body framing is unaffected and there is no correctness regression. The tradeoff is one small heap allocation per send in exchange for removing the extra syscall and the runt `TCP_NODELAY` segment, matching the daemon-side `socket_stream_send` pattern.

---

phil-opp:

Framing is safe — `write_all` on the combined buffer loops internally, so a short socket write resumes mid-buffer rather than desyncing header from body, and the receiver (`tcp_receive`, unchanged) still reads 8 bytes then `reply_len`. The pattern already exists in-tree at `binaries/daemon/src/socket_stream_utils.rs:17-25`.

The one concern is that the optimisation inverts at the large end. `apis/rust/node/src/daemon_connection/tcp.rs:80-90` copies the whole payload into a second buffer, and this path carries `DaemonRequest::SendMessage { data }` and `WritePinnedMemory { tensor_data }`, capped at `MAX_MESSAGE_BYTES` = 64 MiB. So a node publishing a 64 MiB output to a non-shared-memory daemon now allocates and memcpys an extra 64 MiB to save one syscall.

Worth bounding — only merging below some size threshold, or reserving the 8 header bytes inside `encode_presized` so no second buffer is needed. Not blocking, since the daemon side already made the same trade.

phil-opp:
Good catch — verified and fixed in db57810.

You're right that the large end inverts the trade. I confirmed the node control channel is TCP-only (`DaemonChannel` has a single `Tcp(TcpStream)` variant), and that an output taking the daemon path instead of direct zenoh sends its **full payload inline** via `DaemonRequest::SendMessage { data }` — capped at `MAX_MESSAGE_BYTES` = 64 MiB (`node/mod.rs` `Delivery::Daemon`, which even has an explicit `> MAX_MESSAGE_BYTES` guard right before the send). So the unconditional copy could have memcpy'd tens of MiB to save one header segment. (My original PR description's "≤4 KiB only" claim was wrong for this path.)

The fix bounds the coalescing to payloads ≤ 16 KiB — which covers normal control messages and small data outputs, where avoiding the runt segment under `TCP_NODELAY` is the actual win. Above the threshold it writes the 8-byte header and the body separately with no copy; the runt header segment is negligible relative to a multi-MiB body. The wire format (8-byte LE length prefix + body) is identical on both paths, and a new `large_payload_roundtrips_uncoalesced` test exercises the uncoalesced branch.

I went with the threshold rather than reserving 8 bytes inside the serializer because it keeps the change local to `tcp_send` and its framing tests; happy to switch to the reserved-header approach (eliminating the copy for all sizes) if you'd prefer that instead.

---

phil-opp:
The new commit `db57810` addresses the large-payload concern raised in the previous review. `tcp_send` now coalesces only when `message.len() <= COALESCE_THRESHOLD` (16 KiB); larger payloads take the header-then-body path with no extra copy, so the up-to-64 MiB `SendMessage`/`WritePinnedMemory` outputs no longer get memcpy'd just to avoid one runt header segment.

I verified the branch is behaviorally sound: both paths emit the identical wire format (8-byte LE length prefix + body), `write_all` on the combined buffer loops internally so a short write can't desync header from body, and the receiver (`tcp_receive`) is unchanged. The added `large_payload_roundtrips_uncoalesced` test genuinely exercises the uncoalesced branch (`COALESCE_THRESHOLD + 1` bytes) and asserts both the on-wire length and a full round-trip. No issues found in the updated diff.

---

---

PR #3129, opened 2026-08-12
perf(daemon): build the output routing key once per message in send_out

phil-opp:
## Issue

`Daemon::send_out` (`binaries/daemon/src/lib.rs`) runs for **every** message on the daemon path. It cloned `node_id` and `output_id` to build an `OutputId` key for two map lookups (`open_external_mappings`, `debug_topic_watchers`), then moved the originals into `send_output_to_local_receivers`, which immediately rebuilt the *identical* `OutputId` from them:

```rust
let output_id_key = OutputId(node_id.clone(), output_id.clone()); // 2 String allocations
...
send_output_to_local_receivers(node_id, output_id, ...);          // rebuilds OutputId(node_id, output_id)
```

`NodeId`/`DataId` are `String`-backed, so this is two heap allocations per message purely to construct a key that is then reconstructed.

## Fix

Change `send_output_to_local_receivers` to take `output_id: &OutputId`. `send_out` now builds the key **once** by moving `node_id`/`output_id` in (neither is read afterwards) and passes it by reference; the key is reused for both `contains` checks, local delivery, and the inter-daemon path. The inter-daemon (`handle_inter_daemon_event`) call site keeps a single `node_id.clone()` because `node_id` is still needed there by the error logger. All call sites (production + the four routing/circuit-breaker unit tests + the routing bench helper) are updated.

Net effect: 2 `String` allocations per daemon-path message → 0 (main path) / 1 (inter-daemon error-logging path). Behavior is unchanged.

## Validation

- `cargo test -p dora-daemon --lib` green (173 tests, incl. the circuit-breaker recovery tests that exercise `send_output_to_local_receivers` directly).
- `cargo clippy -p dora-daemon -- -D warnings` and `cargo fmt --all -- --check` clean; whole-workspace `cargo check --all` clean.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -144,9 +144,9 @@ pub mod bench_support {
 
     /// Run one iteration of the routing hot path using pre-built fixture.
     pub async fn route_message(df: &mut RunningDataflow, fixture: &RoutingFixture, clock: &HLC) {
+        let output_id = OutputId(fixture.sender_id.clone(), fixture.output_id.clone());
         let _ = send_output_to_local_receivers(
-            fixture.sender_id.clone(),
-            fixture.output_id.clone(),
+            &output_id,
             df,
             &fixture.metadata,
             Some(fixture.data_msg.clone()),
@@ -3741,9 +3741,11 @@ impl Daemon {
                     let dataflow = self.running.get_mut(&dataflow_id).wrap_err_with(|| {
                         format!("send out failed: no running dataflow with ID `{dataflow_id}`")
                     })?;
+                    // `node_id` is still needed below for the error logger, so
+                    // clone it into the key; `output_id` is moved.
+                    let output_id_key = OutputId(node_id.clone(), output_id);
                     send_output_to_local_receivers(
-                        node_id.clone(),
-                        output_id.clone(),
+                        &output_id_key,
                         dataflow,
                         &metadata,
                         data.map(DataMessage::Vec),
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I checked the perf change for behavioral equivalence.

`send_output_to_local_receivers` previously took `node_id: NodeId, output_id: DataId` and rebuilt `OutputId(node_id, output_id)` internally; it now takes `&OutputId` and looks up `dataflow.mappings.get(output_id)` directly — the same key, same lookup.

Call sites all construct the key in the original field order `OutputId(node_id/sender, output_id/output)`, so the routing key is unchanged:
- Main hot path (`lib.rs:5016`): key is now built once by moving `node_id`/`output_id` in (neither is read afterward — they were previously moved into the callee, and the post-call inter-daemon block at `lib.rs:5040` uses `output_id_key`, which survives because the callee now borrows it). Saves the two per-message `String` clones. Correct.
- Inter-daemon forward path (`lib.rs:~3745`): `node_id` is still cloned (needed for `.for_node(node_id)` in the error logger), `output_id` is moved. Correct.
- Bench fixture and the four `fault_tolerance_tests` call sites construct the key with the same operands.

No correctness regression; the change only removes redundant heap allocations on the key.

---

phil-opp:

No issues found. `send_output_to_local_receivers` now takes `&OutputId` and looks up `dataflow.mappings.get(output_id)` directly instead of rebuilding the key from moved `NodeId`/`DataId` — same key, same lookup. All call sites construct the key in the original field order `OutputId(node_id, output_id)`, so routing is unchanged.

On the hot path (`send_out`), the key is built once by moving `node_id`/`output_id` in; because they are moved, any later read would be a compile error, and the post-call inter-daemon block uses `output_id_key` instead — so passing it by reference is sound and drops the two per-message `String` clones. The inter-daemon forward path keeps a single `node_id.clone()` (still needed by the error logger). Correct.

---

PR #3130, opened 2026-08-12
refactor(core): dedupe the type-parameter agreement rule into one helper

phil-opp:
## Issue

`libraries/core/src/types.rs` implemented the same parameter-agreement contract **twice**, independently:

- `types_match` (URN string comparison), and
- `CompatibilityGraph::is_compatible` (the implicit-conversion graph check).

Both encode: *same base URN, and for every key present in **both** parameter maps the values must agree; an empty map on either side is a wildcard that matches anything.* This rule gates the type system in two distinct code paths, so a future edit to one copy that misses the other would be a silent correctness bug.

## Fix

Extract the rule into a single private helper:

```rust
fn params_agree(a: &BTreeMap<String, String>, b: &BTreeMap<String, String>) -> bool {
    if a.is_empty() || b.is_empty() { return true; }
    a.iter().all(|(k, va)| b.get(k).is_none_or(|vb| vb == va))
}
```

Both call sites now delegate to it. The behavior is identical, including the wildcard-on-empty and unparsed-URN (wildcard) cases in `is_compatible`.

## Validation

- `cargo test -p dora-core` green (146 unit tests + 3 doctests), including the existing `types_match_*` and `compat_*` coverage that exercises the wildcard and disagreeing-param cases.
- `cargo clippy -p dora-core -- -D warnings` and `cargo fmt --all -- --check` clean; whole-workspace `cargo check --all` clean.

---

🤖 **This is a machine-generated pull request.** It was produced by Claude (an AI agent, model Opus 4.8) during an automated code-review pass over the repository. Please review carefully before merging; a human maintainer should confirm the analysis and the change.

---

### DIFF
--- a/libraries/core/src/types.rs
+++ b/libraries/core/src/types.rs
@@ -254,6 +254,20 @@ pub fn parse_urn(urn: &str) -> Option<ParsedUrn> {
     Some(ParsedUrn { base, params })
 }
 
+/// Parameter-agreement rule shared by [`types_match`] and
+/// [`CompatibilityGraph::is_compatible`]: two parameter maps agree when every
+/// key present in *both* maps has an equal value. An empty map on either side is
+/// a wildcard that matches anything.
+///
+/// Keeping this in one place ensures the two type-compatibility code paths can
+/// never silently diverge on the parameter contract.
+fn params_agree(a: &BTreeMap<String, String>, b: &BTreeMap<String, String>) -> bool {
+    if a.is_empty() || b.is_empty() {
+        return true;
+    }
+    a.iter().all(|(k, va)| b.get(k).is_none_or(|vb| vb == va))
+}
+
 /// Check if two type URNs are compatible (considering parameters).
 ///
 /// Rules:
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I checked the refactor for behavioral equivalence and it holds in every case.

`params_agree(a, b)` returns `true` if either map is empty, else `a.iter().all(|(k,va)| b.get(k).is_none_or(|vb| vb == va))` — i.e. every key present in *both* must agree, keys unique to either side are ignored. That is exactly the old inline rule.

- `types_match` (`types.rs:285`): old loop over `pa.params` returning `false` on a shared-key mismatch, with the empty-map early return, is identical to `params_agree(&pa.params, &pb.params)`.
- `CompatibilityGraph::is_compatible` (`types.rs:369`), case by case:
  - both parsed, both non-empty → old checks shared keys / returns `false` on mismatch; new `params_agree` does the same.
  - both parsed, at least one empty → old match guard fails, falls to `return true`; new `params_agree` short-circuits to `true`.
  - either side unparsed (`None`) → old `_ => {}` then `return true`; new `if let (Some,Some)` fails, then `return true`.

All three branches match the prior behavior. `Option::is_none_or` is stable well below the pinned toolchain. The existing `types_match_*`/`compat_*` tests cover these paths.

---

phil-opp:

No issues found. The extracted `params_agree(a, b)` returns `true` when either map is empty, else `a.iter().all(|(k, va)| b.get(k).is_none_or(|vb| vb == va))` — i.e. every shared key must agree, keys unique to either side are ignored. That is exactly the two inline rules it replaces.

Checked `CompatibilityGraph::is_compatible` branch by branch:
- both parsed, both non-empty → old loop returns `false` on a shared-key mismatch; `params_agree` does the same.
- both parsed, at least one empty → old guard falls through to `return true`; `params_agree` short-circuits to `true`.
- either side unparsed (`None`) → old `_ => {}` then `return true`; new `if let (Some, Some)` fails, then `return true`.

`types_match` reduces identically. Behavior is preserved.

---

PR #3135, opened 2026-08-12
fix(cli): reject cluster host/user starting with `-` (ssh arg injection)

### LINKED ISSUE #3133: cli(cluster): #2861 leaves ssh argument-injection open — a `machine.host`/`user` starting with `-` runs an arbitrary command on the local machine (ProxyCommand)
> **This issue was created by a scheduled, automated Claude review** of recent commits/PRs. It was verified by reading the code directly, but please double-check before acting on it.

## Summary

PR #2861 (commit `9b288f5`, *"fix(cli): validate cluster machine id and labels before they reach the remote shell"*) hardens `dora cluster` against remote-shell injection by validating `machine.id`, label keys/values, and `zenoh_peer`. It **deliberately** leaves `machine.host` and `machine.user` unvalidated, with this stated rationale (`binaries/cli/src/command/cluster/config.rs:92-96`):

> `host`/`user` are deliberately left unrestricted: they are network addresses that legitimately carry characters outside this set (e.g. IPv6 literals), and are folded into the ssh **target** rather than the remote command string.

That reasoning only rules out injection into the **remote** shell. It overlooks injection into the **local `ssh` client's own argument parser**, which is a separate and arguably more severe vector: it executes on the operator's machine, before any network connection.

## The vulnerability

`ssh_target` interpolates `host`/`user` verbatim (`binaries/cli/src/command/cluster/mod.rs:58-63`):

```rust
pub(super) fn ssh_target(machine: &MachineConfig) -> String {
    match &machine.user {
        Some(user) => format!("{user}@{}", machine.host),
        None => machine.host.clone(),
    }
}
```

and `run_ssh` passes that target to `ssh` with **no `--` end-of-options separator** (`binaries/cli/src/command/cluster/mod.rs:92-105`):

```rust
pub(super) fn run_ssh(target: &str, port: Option<u16>, cmd: &str) -> eyre::Result<bool> {
    let mut command = std::process::Command::new("ssh");
    command.args([ "-o", "BatchMode=yes", /* … */ ]);
    if let Some(p) = port { command.args(["-p", &p.to_string()]); }
    command.args([target, cmd]);   // <-- target is unguarded; no "--" before it
    // …
}
```

If `host` (or `user`) begins with `-`, the local `ssh` binary parses `target` as an **option**, not a hostname. This is the classic ssh/git argument-injection class (cf. CVE-2017-1000117).

### Concrete exploit

A cluster descriptor is the untrusted input the rest of #2861 is defending against. Given:

```yaml
machines:
  - id: a
    host: "-oProxyCommand=touch /tmp/pwned;false"
```

the argv becomes:

```
ssh -o BatchMode=yes … -oProxyCommand=touch /tmp/pwned;false <remote_cmd>
```

`ssh` interprets `-oProxyCommand=…` and runs `touch /tmp/pwned;false` **on the machine running `dora cluster up`** (via the local shell that `ProxyCommand` uses) before/instead of connecting. Arbitrary local command execution.

`user` is an equivalent vector: `ssh_target` yields `{user}@{host}`, so a `user` beginning with `-` produces a target string that also starts with `-`.

## Affected code paths

`run_ssh` / `ssh_target` are the shared SSH primitives for the whole `dora cluster` surface, so every subcommand that shells out is affected:

- `binaries/cli/src/command/cluster/up.rs` (`ssh_target` at ~:66, ~:78)
- `install.rs`, `uninstall.rs`, `upgrade.rs` (all route through `run_ssh`)

The `host`/`user` values originate from `MachineConfig` parsed out of the cluster YAML and are never checked for a leading `-`.

## Suggested fix

Two complementary defenses (the codebase already treats the cluster descriptor as an injection boundary, so validation is consistent with its own posture):

1. **Reject `host`/`user` values that begin with `-`** in `ClusterConfig::validate()` (git's portable mitigation for exactly this class). This keeps IPv6 literals and normal `user@host` values valid while closing the option-injection hole. It can't be folded into `validate_shell_safe` (which would reject the `:`/`[`/`]` in IPv6), so it needs a dedicated leading-dash check.
2. **Insert a literal `"--"` before `target`** in `run_ssh` — `command.args(["--", target, cmd])`. OpenSSH accepts `--` as an end-of-options marker; this is defense-in-depth for the local argv, and because all subcommands share `run_ssh`, one change covers them all. (Note `--` support is version-dependent on very old OpenSSH, so prefer #1 as the primary fix and treat #2 as belt-and-suspenders.)

## Notes / non-issues found in the same review

- A secondary, low-severity observation on the same PR: `validate_endpoint_shell_safe` permits `[`/`]` in `zenoh_peer`, which is interpolated **unquoted** into the remote command. Under default `sh`/`bash` (no `nullglob`/`failglob`) an unmatched `[::1]` glob is left literal, so it normally works; it is not command injection (no way to start a new command). Single-quoting the remote interpolations, or requiring the un-bracketed IPv6 form, would remove the glob ambiguity. Low priority.

phil-opp:
## Summary

Closes #3133.

`dora cluster` builds its ssh target by folding `machine.host`/`machine.user` verbatim (`{user}@{host}`) and passes it to the local `ssh` binary **with no `--` end-of-options separator** (`run_ssh`). If `host` (or `user`) begins with `-`, the local `ssh` client parses the target as an **option** rather than a hostname — the classic ssh/git argument-injection class (cf. CVE-2017-1000117).

### Exploit

```yaml
machines:
  - id: a
    host: "-oProxyCommand=touch /tmp/pwned;false"
```

produces `ssh … -oProxyCommand=touch /tmp/pwned;false <cmd>`, so `ssh` runs `touch /tmp/pwned;false` **on the machine running `dora cluster up`**, before/instead of connecting. `user` is an equivalent vector since `ssh_target` yields `{user}@{host}`.

#2861 deliberately left `host`/`user` unvalidated because they legitimately carry characters the id charset rejects (IPv6 literals, `@`), but that reasoning only ruled out injection into the *remote* shell — it overlooked injection into the *local* ssh argv, which executes before any network connection.

## Changes

Two complementary defenses (matching the issue's suggested fix):

1. **`ClusterConfig::validate` rejects a `host`/`user` beginning with `-`** via a dedicated `validate_no_leading_dash` check. This can't be folded into `validate_shell_safe` (whose `[a-zA-Z0-9_.-]` charset would reject the `:`/`[`/`]` in an IPv6 literal), so it's a separate leading-dash-only check that keeps IPv6 literals and normal `user@host` values valid.
2. **`run_ssh` inserts a literal `--` before the target** (`command.args(["--", target, cmd])`) as defense-in-depth — OpenSSH treats `--` as an end-of-options marker, and because every `dora cluster` subcommand routes through `run_ssh`, one change covers them all.

## Tests

Added to `binaries/cli/src/command/cluster/config.rs`:
- `reject_host_with_leading_dash` — the `-oProxyCommand=…` payload is rejected.
- `reject_user_with_leading_dash` — the equivalent `user` vector is rejected.
- `accept_ipv6_host` — `::1`, `2001:db8::1`, and a plain IPv4 host stay valid (regression guard so the new check doesn't over-reject).

`cargo test -p dora-cli` (21 cluster tests pass), `cargo clippy -p dora-cli -- -D warnings`, and `cargo fmt --check` are all green.

## Note on a related low-severity observation

The issue also mentions `validate_endpoint_shell_safe` permitting `[`/`]` in `zenoh_peer` (interpolated unquoted into the remote command). As the issue itself notes, that is not command injection under default `sh`/`bash` — it's left out of this PR to keep the change focused on the security-relevant argument-injection fix.

https://claude.ai/code/session_01Xtgsx6GZ58s8Hm4G6AECrF

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xtgsx6GZ58s8Hm4G6AECrF)_

### DIFF
--- a/binaries/cli/src/command/cluster/config.rs
+++ b/binaries/cli/src/command/cluster/config.rs
@@ -105,6 +105,20 @@ impl ClusterConfig {
             if m.host.is_empty() {
                 bail!("machine `{}` host must not be empty", m.id);
             }
+            // `host`/`user` are folded into the ssh *target* (`{user}@{host}`)
+            // and passed to the local `ssh` binary. They are deliberately not
+            // run through `validate_shell_safe` (a target legitimately carries
+            // `:`/`[`/`]` for IPv6 literals, `@`, etc.), but a value that begins
+            // with `-` is parsed by `ssh`'s own argument parser as an *option*
+            // rather than a hostname — e.g. `-oProxyCommand=...` executes an
+            // arbitrary command on the local machine before connecting (the
+            // classic ssh/git argument-injection class, cf. CVE-2017-1000117).
+            // Reject a leading dash on both fields; `run_ssh` additionally
+            // passes `--` before the target as defense-in-depth.
+            validate_no_leading_dash(&format!("machine `{}` host", m.id), &m.host)?;
+            if let Some(user) = &m.user {
+                validate_no_leading_dash(&format!("machine `{}` user", m.id), user)?;
+            }
             // Reject daemon_port = 0: the daemon would bind an ephemeral port
             // but exports DORA_DAEMON_LOCAL_LISTEN_PORT=0 to spawned nodes
             // (binaries/cli/src/command/daemon.rs), so they fail to connect.
@@ -125,6 +139,17 @@ fn validate_shell_safe(what: &str, value: &str) -> eyre::Result<()> {
         .find(|c| !c.is_ascii_alphanumeric() && *c != '_' && *c != '-' && *c != '.')
     {
         bail!("{what} contains invalid character `{ch}` -- only [a-zA-Z0-9_.-] are allowed");
+    }
+    Ok(())
+}
+
+/// Reject a value that begins with `-`, which the local `ssh` binary would
+/// otherwise parse as an option rather than a hostname/user (argument
+/// injection). Unlike [`validate_shell_safe`] this places no charset
+/// restriction, so IPv6 literals and normal `user@host` values stay valid.
+fn validate_no_leading_dash(what: &str, value: &str) -> eyre::Result<()> {
+    if value.starts_with('-') {
+        bail!("{what} must not start with `-` (would be parsed as an ssh option)");
     }
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


phil-opp:

I reviewed the security fix at commit `9db007b` and did not find a hole in the guard. Notes on what I checked:

- **`--option=` / `-o` forms both caught.** `validate_no_leading_dash` rejects on `value.starts_with('-')`, which covers both the short `-oProxyCommand=…` form from the report *and* GNU-style `--option=…` forms, since both begin with `-`. Good.
- **Combined target can't start with `-`.** `ssh_target` builds `{user}@{host}`; host and user are validated independently. When `user` is `Some`, the target starts with the first char of `user`; when `None`, with `host`. Rejecting a leading `-` on each field means the concatenation can never present a leading `-` to ssh. Correct.
- **No local-shell split risk.** host/user are passed as a single argv element via `Command::args` (no shell), so embedded spaces/metachars like `host = "foo -oProxyCommand=x"` stay one hostname token and aren't re-parsed — only a *leading* dash matters for the local argv, which is exactly what's guarded.
- **Remote command path unchanged.** host/user are folded only into the ssh target, not the remote command string (which is separately `validate_shell_safe`'d), so this doesn't reopen the #2861 remote-shell vector.
- Tests exercise both rejected fields and accepted IPv6 literals — not vacuous.

One residual defense-in-depth asymmetry worth a look (not an exploitable bug given the validation): `binaries/cli/src/command/cluster/upgrade.rs:61` folds the target into the scp destination `format!("{target}:/usr/local/bin/dora")` **without** an `--` end-of-options separator, unlike the `--` you added to `run_ssh`. Today the config-level leading-dash check makes the target safe, so scp is protected transitively; but if that validation ever regresses or a target reaches scp by another path, scp would have no second line of defense. Adding `scp.args(["--", …])` (or passing the destination after `--`) would make the two paths symmetric. Purely defense-in-depth — the current fix closes the reported vector.

---

PR #3136, opened 2026-08-12
fix(coordinator): roll back and tear down on async partial spawn failure

### LINKED ISSUE #3134: coordinator: async partial spawn failure (one daemon Errs after another already spawned) never rolls back the succeeded daemon — orphaned nodes + `running_dataflows`/store divergence (`dora list` stuck "Running")
> **Note:** This issue was created by a scheduled, automated Claude code-review check (unattended run). The finding below was verified against the source at commit `f3c032f` (branch `main`), but a human should confirm before acting on it.

## Summary

On a multi-daemon dataflow, when the daemons report their spawn results **asynchronously** and one daemon reports success while another reports failure, `handle_spawn_result_err` marks the dataflow terminally `Failed` and caches the spawn error — but it does **not** roll back the daemon(s) that already spawned successfully, and it does **not** remove the dataflow from `running_dataflows`. The result is:

1. **Orphaned node processes** on the succeeded daemon(s) that keep running indefinitely, unmanaged by the coordinator.
2. **A durable in-memory vs. persisted-store divergence:** the store record says `Failed { terminal: true }`, but the dataflow stays in `running_dataflows`, so `dora list` reports it as **`Running`** — forever, until a manual `dora stop`.

This is the *asynchronous* sibling of the partial-spawn-failure cases that the codebase already handles correctly (see "Why this is a gap, not by design" below); the async path is simply missing the equivalent rollback + teardown.

## Affected code

- `binaries/coordinator/src/lib.rs` — `handle_spawn_result_err` (lines ~3046–3077)
- dispatched from the `Event::DataflowSpawnResult` handler (lines ~2461–2493)

```rust
fn handle_spawn_result_err(
    dataflow_id: DataflowId,
    daemon_id: &DaemonId,
    err: eyre::Report,
    store: &dyn CoordinatorStore,
) {
    if !dataflow.spawn_result.is_pending() { /* ignore late ... */ return; }

    // persists Failed { terminal: true } ...
    // <-- no rollback of already-succeeded daemons,
    //     no removal/teardown of the running_dataflows entry
}
```

## Concrete failure scenario

Dataflow spawned across daemons **A** and **B**. Each daemon *accepts* the spawn request synchronously (`run::spawn_dataflow` → `TriggerSpawnResult(Ok)` only means "queued"); the actual node spawn happens asynchronously and is reported later via `DaemonEvent::SpawnResult` → `Event::DataflowSpawnResult`.

**Ordering 1 — A succeeds first, then B fails:**
1. A's `Ok` arrives → `pending_spawn_results` drops A; `spawn_result` stays `Pending` (B still pending). **A's node processes are now running.**
2. B's `Err` arrives (e.g. missing binary/env on machine B) → `handle_spawn_result_err` runs while `spawn_result.is_pending()` is still true → persists `Failed { terminal: true }`, sets `spawn_result = Err`. **A is never told to stop.**

**Ordering 2 — B fails first, then A succeeds:**
1. B's `Err` arrives → terminal `Failed`, `spawn_result = Err`.
2. A's `Ok` arrives → `handle_spawn_result_ok` hits the `!is_pending()` guard and **ignores the late success** — but A's spawn genuinely succeeded and **A's nodes are running**. A is never rolled back.

Either way: the succeeded daemon's nodes are orphaned and the dataflow stays in `running_dataflows`.

## Why no recovery path catches it

- **Spawn-timeout watchdog** (`check_spawn_timeouts`, line ~3551) filters on `df.spawn_result.is_pending()`. After `handle_spawn_result_err`, the result is `Cached(Err)` — **not pending** — so the watchdog (which *does* do `fire_and_forget_rollback` + full teardown) never touches this dataflow.
- **`DaemonStatusReport` reconcile / orphan-stop** (lines ~2511–2654) can stop orphans, but:
  - the `archived_dataflows` orphan-stop branch (line ~2513) is skipped because this path never archives the dataflow; and
  - the terminal-store orphan-stop branch (`status_report_should_stop_orphan`, line ~2635) only fires **when a daemon actually sends a `DaemonStatusReport`**. The daemon emits `DaemonEvent::StatusReport` **exactly once**, immediately before its main event loop (`binaries/daemon/src/lib.rs:1787`, before `while let Some(event) = events.next().await`) — i.e. at startup/reconnect, **not periodically**. The healthy daemon A never disconnects here, so it never re-reports, and this cleanup never runs.
- **`ControlRequest::List`** (lines ~1100–1110) unconditionally classifies every entry still in `running_dataflows` as `DataflowStatus::Running`, so the divergence is user-visible.

Net effect: leaked/orphaned node processes on the healthy daemon **plus** a dataflow permanently shown as `Running` despite a terminal `Failed` store verdict — cleared only by a manual `dora stop` (or if daemon A happens to disconnect and reconnect, which re-emits a StatusReport and triggers the terminal orphan-stop).

## Why this is a gap, not by design

The two sibling partial-spawn-failure paths both roll back the succeeded daemons and reconcile in-memory state; only the async-report path is missing it:

- `run::spawn_dataflow` calls `rollback_spawned_daemons` on a **synchronous** partial failure (`binaries/coordinator/src/run/mod.rs` ~88–118).
- `check_spawn_timeouts` calls `fire_and_forget_rollback` and performs full teardown (archive, drain `stop_reply_senders`, synthesize `dataflow_results`, remove from `running_dataflows`) on a **timed-out** spawn (`lib.rs` ~3568–3742).

`handle_spawn_result_err` should do the equivalent: roll back the daemons that already succeeded (the `daemons.difference(&pending_spawn_results)` set) and perform the in-memory teardown so state doesn't diverge — rather than only recording the failure and caching the error.

## Suggested direction

In `handle_spawn_result_err`, when transitioning the dataflow to terminal `Failed`, additionally:
1. Compute the succeeded-daemon set (`dataflow.daemons.difference(&dataflow.pending_spawn_results)`) and issue `fire_and_forget_rollback` (`StopDataflow { force: true }`) to them, and
2. Perform the same teardown the timeout watchdog does (archive + remove from `running_dataflows` + synthesize results), so `dora list` and the store agree.

Because `handle_spawn_result_err` currently takes `&mut RunningDataflow` (not the maps / `daemon_connections`), realizing this likely means moving the rollback/teardown up into the `Event::DataflowSpawnResult` handler (which has access to `running_dataflows`, `daemon_connections`, `archived_dataflows`, `dataflow_results`, and the clock), mirroring how `check_spawn_timeouts` is structured. A regression test analogous to the existing spawn-timeout tests (assert the succeeded daemon receives a `StopDataflow` and the dataflow leaves `running_dataflows`) would pin the behavior.

phil-opp:
## Summary

Closes #3134.

On a multi-daemon dataflow whose daemons report spawn results **asynchronously**, when one daemon reports success and another reports a spawn failure, the `Event::DataflowSpawnResult` Err arm (via `handle_spawn_result_err`) marked the dataflow terminally `Failed` and cached the error — but it did **not** roll back the daemon(s) that already spawned, and did **not** remove the dataflow from `running_dataflows`. The result:

1. **Orphaned node processes** on the succeeded daemon(s), unmanaged by the coordinator forever.
2. **A durable in-memory vs. persisted-store divergence:** the store says `Failed { terminal: true }` but the dataflow stays in `running_dataflows`, so `dora list` reports it as **`Running`** — indefinitely, until a manual `dora stop`.

No recovery path catches it: the spawn-timeout watchdog (`check_spawn_timeouts`) filters on `spawn_result.is_pending()` (now `Cached(Err)`, so it's skipped), and the terminal-store orphan-stop only fires on a `DaemonStatusReport`, which each daemon emits exactly once at startup/reconnect — the healthy daemon never re-reports.

This is the **asynchronous sibling** of the partial-spawn-failure cases the codebase already handles for the synchronous (`run::rollback_spawned_daemons`) and timed-out (`check_spawn_timeouts`) paths; only the async-report path was missing the equivalent rollback + teardown.

## Changes

- Extract the spawn-timeout watchdog's rollback + in-memory teardown into a shared **`rollback_and_teardown_failed_spawn`** helper: fire-and-forget `StopDataflow{force}` to the daemons no longer pending a spawn result, final log line, close topic subscribers, synthesize per-node `FailedToSpawn` results, drain `stop_reply_senders`, archive, and cap. The detailed round-6/7/8 invariant comments move with the code.
- `check_spawn_timeouts` keeps its own terminal set-result/persist (with its timeout message) and then delegates the teardown to the helper — behavior is unchanged (all existing watchdog tests still pass).
- The `Event::DataflowSpawnResult` Err arm now persists via `handle_spawn_result_err` (unchanged, keeping its late-arrival guard and its store-error message), and **only when the failure is the terminal transition** (the dataflow was still pending) rolls back the not-still-pending daemons (the succeeded ones, plus the failing daemon in case it partially started nodes) and tears the dataflow down. Late failures on an already-terminal dataflow keep the previous "ignore, don't overwrite" behavior.

The rollback/teardown decision is computed inside the `get_mut` borrow and executed after it ends, so the `running_dataflows.remove` is borrow-safe.

## Ordering coverage

- **A succeeds, then B fails** (the primary case): B's failure is the terminal transition → roll back A (and B) + tear down. ✅ covered by the new test.
- **B fails, then A succeeds:** B's failure tears the dataflow down; A's later `Ok` finds no entry and is logged. A's rollback in this ordering is a known residual the original issue calls out (A had not reported success at teardown time, so it isn't in the rollback set) — out of scope for this change, which closes the verified orphan/divergence on the succeeded-first ordering.

## Tests

Added `async_partial_spawn_failure_rolls_back_succeeded_daemon_and_tears_down`, which replays the exact statements the handler runs for the "A succeeded, B fails" ordering and asserts: the succeeded daemon receives a `StopDataflow`, the dataflow leaves `running_dataflows` (the core regression), it is persisted `Failed` + archived, and in-flight `dora stop` / spawn waiters are released rather than hung.

`cargo test -p dora-coordinator` (126 lib + integration tests, incl. all pre-existing spawn/watchdog tests), `cargo clippy -p dora-coordinator -- -D warnings`, and `cargo fmt --check` are all green.

https://claude.ai/code/session_01Xtgsx6GZ58s8Hm4G6AECrF

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xtgsx6GZ58s8Hm4G6AECrF)_

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2462,35 +2462,88 @@ async fn start_inner(
                 dataflow_id,
                 daemon_id,
                 result,
-            } => match running_dataflows.get_mut(&dataflow_id) {
-                Some(dataflow) => {
-                    dataflow.pending_spawn_results.remove(&daemon_id);
-                    match result {
-                        Ok(()) => {
-                            handle_spawn_result_ok(
-                                dataflow,
-                                dataflow_id,
-                                &daemon_id,
-                                store.as_ref(),
-                            );
-                        }
-                        Err(err) => {
-                            handle_spawn_result_err(
-                                dataflow,
-                                dataflow_id,
-                                &daemon_id,
-                                err,
-                                store.as_ref(),
-                            );
-                        }
-                    };
+            } => {
+                // On an async partial spawn failure — one daemon reports a
+                // spawn error while others have already reported success — we
+                // must roll back the succeeded daemons and tear the dataflow
+                // down. Otherwise their node processes are orphaned and the
+                // dataflow lingers in `running_dataflows`, shown as `Running`
+                // forever despite a terminal `Failed` store record
+                // (dora-rs/dora#3134). `handle_spawn_result_err` already
+                // persists `Failed` + sets `spawn_result`; the rollback and
+                // in-memory teardown are decided here (inside the `get_mut`
+                // borrow) and performed once that borrow ends.
+                let mut teardown: Option<(BTreeSet<DaemonId>, String)> = None;
+                match running_dataflows.get_mut(&dataflow_id) {
+                    Some(dataflow) => {
+                        dataflow.pending_spawn_results.remove(&daemon_id);
+                        match result {
+                            Ok(()) => {
+                                handle_spawn_result_ok(
+                                    dataflow,
+                                    dataflow_id,
+                                    &daemon_id,
+                                    store.as_ref(),
+                                );
+                            }
+                            Err(err) => {
+                                // A failure only tears the dataflow down if it
+                                // transitions it to terminal *now*. If it was
+                                // already terminal (watchdog fired, or an
+                                // earlier daemon already failed it),
+                                // `handle_spawn_result_err`'s guard makes the
+                                // persist a no-op and we must not tear down a
+                                // second time.
+                                let was_pending = dataflow.spawn_result.is_pending();
+                                let err_msg =
+                                    format!("spawn failed on daemon `{daemon_id}`: {err}");
+                                handle_spawn_result_err(
+                                    dataflow,
+                                    dataflow_id,
+                                    &daemon_id,
+                                    err,
+                                    store.as_ref(),
+                                );
+                                if was_pending {
+                                    // Daemons no longer pending a spawn result:
+                                    // those that already succeeded, plus this
+                                    // one (which may have started some nodes
+                                    // before failing). Force-stop them all.
+                                    let rollback_daemons: BTreeSet<DaemonId> = dataflow
+                                        .daemons
+                                        .difference(&dataflow.pending_spawn_results)
+                                        .cloned()
+                                        .collect();
+                                    teardown = Some((rollback_daemons, err_msg));
+                                }
+                            }
+                        };
+                    }
+                    None => {
+                        tracing::warn!(
+                            "received DataflowSpawnResult, but no matching dataflow in `running_dataflows` map"
+                        );
+                    }
                 }
-                None => {
-                    tracing::warn!(
-                        "received DataflowSpawnResult, but no matching dataflow in `running_dataflows` map"
-                    );
+                if let Some((rollback_daemons, err_msg)) = teardown {
+                    // The `get_mut` borrow above has ended; safe to remove and
+                    // hand ownership to the shared rollback/teardown helper.
+                    if let Some(df) = running_dataflows.remove(&dataflow_id) {
+                        rollback_and_teardown_failed_spawn(
+                            dataflow_id,
+                            df,
+                            &rollback_daemons,
+                            err_msg,
+                            &mut running_dataflows,
+                            &mut archived_dataflows,
+                            &mut dataflow_results,
+                            &mut daemon_connections,
+                            &clock,
+                        )
+                        .await;
+                    }
                 }
-            },
+            }
             Event::DaemonStatusReport {
                 daemon_id,
                 running_dataflows: reported_dataflows,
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed commit `6086504` (the async partial-spawn-failure rollback + the extraction of `rollback_and_teardown_failed_spawn`). **No blocking issues found.**

What I checked:

- **`was_pending` guard is captured correctly** — read before `handle_spawn_result_err` moves `err`, and `err_msg` is `format!`-ed before the move, so no use-after-move. `handle_spawn_result_err` (lib.rs:3046) only sets `spawn_result`/persists and does no removal or rollback of its own, so there's no double-teardown when the helper then runs.
- **Rollback set is right** — `daemons.difference(&pending_spawn_results)` is computed after `pending_spawn_results.remove(&daemon_id)`, so it correctly includes the succeeded daemon(s) and the just-failed one (which may have partially started nodes), while daemons still pending are excluded.
- **Borrow safety** — the teardown decision is computed inside the `get_mut` borrow and the `running_dataflows.remove` + helper call happen after it ends; single-threaded coordinator loop means no race with `check_spawn_timeouts`.
- **Refactor is behavior-preserving for the watchdog path** — the only observable reordering is that `fire_and_forget_rollback` now runs after the `Failed` persist instead of before; both are independent side effects (fire-and-forget mpsc enqueue vs. store write), so no functional change. All the round-6/7/8 invariants (iterate `node_to_daemon` not `daemons`, cap last, insert-before-drain) moved intact.
- **Test is non-vacuous** — `test_running_dataflow` starts with a pending `CachedResult`, and the assertions (StopDataflow dispatched to the mocked succeeded daemon, removal from `running_dataflows`, persisted `Failed` + archived, stop/spawn waiters released) each genuinely exercise the helper. The one caveat is that it replays the handler statements by hand rather than driving `start_inner`, so a future divergence between the handler and the replayed sequence wouldn't be caught — acceptable given the helper itself is exercised for real.

One observation, not a blocker: the "not-yet-reported daemon orphans on later success" residual the PR scopes out for the two-daemon "B fails then A succeeds" ordering generalizes to any daemon still in `pending_spawn_results` at teardown time (e.g. a 3rd daemon C): it's excluded from the rollback set, and once the dataflow leaves `running_dataflows`, C's later `Ok` hits the `None` warn arm and its nodes orphan. This is the same documented residual, just noting the general shape for whoever picks up the follow-up.

---

keirsalterego:
Was looking at #3134 too and hit something on this one.

The B-fails-then-A-succeeds case you mentioned as out of scope, I think it's actually worse after this patch than before it. On main the dataflow is still in `running_dataflows, so dora stop <uuid> reaches daemon A and kills the leftover nodes. Once this tears down, the synthesized dataflow_results entry makes Stop early-return at lib.rs:896 without talking to any daemon, so A just keeps running forever.

Doesn't feel like an edge case either. A spawn error usually comes back fast (missing binary, bad path) while the healthy daemon is still starting its nodes, so B failing first seems pretty normal.

The None arm has enough to handle it :

if result.is_ok() && archived_dataflows.contains_key(&dataflow_id) {
    fire_and_forget_rollback(dataflow_id, &BTreeSet::from([daemon_id]), &mut daemon_connections, &clock).await;
}

keirsalterego:
one more thing, the test copies the match arm instead of calling it, so if someone deletes the teardown block it still passes. Pulling the arm out into handle_dataflow_spawn_result() would fix that and give the late-Ok path somewhere to test.

i've both working locally if you want me to push it.

phil-opp:
**Automated review by Claude** 🤖 _(fully automated review — findings have not been vetted by a human)_

Following up on my earlier review, which downplayed the "later-`Ok` daemon orphans" case as a documented residual. Having re-examined it (and agreeing with @keirsalterego's analysis above), I now think the **B-fails-then-A-succeeds** ordering is a genuine merge-blocking regression, not just a residual — this patch makes it *worse* than `main`, and it's a common ordering (a spawn error like a missing binary/bad path typically returns fast, while the healthy daemon is still starting its nodes).

Trace, in `start_inner` (`binaries/coordinator/src/lib.rs`):

1. B reports `Err` first. A is still pending, so `pending_spawn_results` still contains A after `remove(&daemon_b)`. `rollback_daemons = daemons.difference(&pending_spawn_results)` therefore **excludes A**.
2. `was_pending` is true, so `rollback_and_teardown_failed_spawn` runs: it removes the dataflow from `running_dataflows` and inserts a synthesized `dataflow_results` entry.
3. A later reports `Ok` → the dataflow is gone from `running_dataflows` → the handler hits the `None` warn arm. A's node processes are now orphaned.
4. The user runs `dora stop <uuid>` to clean up. But the `Stop` handler (`lib.rs:927-937`) early-returns `DataflowStopped` whenever `!running_dataflows.contains_key(uuid) && dataflow_results.get(uuid).is_some()` — which the synthesized entry from step 2 guarantees. So `stop_dataflow` is never called and **no `StopDataflow` is ever dispatched to A**. A's nodes run forever with no recovery path.

On `main` (before this patch) the async-failure arm left the dataflow in `running_dataflows`, so `dora stop <uuid>` fell through to `stop_dataflow` and did reach A. The synthesized-results + teardown here removes that recovery path, which is the regression.

The narrow fix @keirsalterego suggested — rolling back in the `None`/late-`Ok` arm when the dataflow is already archived/terminal — closes the orphan directly at the point A reports success, and is worth doing regardless of the `Stop` early-return, since a user shouldn't have to run `dora stop` at all for an already-failed dataflow.

Separately, the regression test replays the handler's statements by hand rather than driving the real `Event::DataflowSpawnResult` arm (as I noted before, and @keirsalterego reiterates), so deleting the teardown from the handler would not fail it, and the late-`Ok` ordering is untested. Extracting the arm into a `handle_dataflow_spawn_result()` the test can call would fix both.

---

phil-opp:
Superseded by #3180

PR #3140, opened 2026-08-12
refactor(coordinator): resolve archived log daemon by direct node lookup

phil-opp:
## Summary

`resolve_log_daemon_id` (in `binaries/coordinator/src/handlers.rs`) resolves which daemon holds a node's logs. For an **archived** dataflow it scanned every entry of the `BTreeMap<NodeId, ResolvedNode>` with a `filter(|node| node.id == *node_id)`, collected into a `Vec`, then matched on the slice:

```rust
let machine_id = match &machine_ids[..] {
    [machine_id] => machine_id.clone(),
    [] => bail!("No machine contains {}/{}", dataflow_id, node_id),
    _ => bail!("More than one machine contains {}/{}. ...", ...),
};
```

## Issue

That map is keyed by `NodeId`, and every `ResolvedNode.id` equals its map key — this is established in `resolve_aliases_and_set_defaults_in_topology` (`libraries/core/src/descriptor/mod.rs`), which inserts `resolved.insert(node.id.clone(), ResolvedNode { id: node.id, ... })` and additionally `bail!`s on a duplicate id. `ArchivedDataflow` is built by cloning that same map (`coordinator/src/state.rs`).

Consequently the filter can never match more than one entry:
- The `_ =>` ("More than one machine contains …") arm is **unreachable dead code**.
- The O(n) `values().filter(...)` scan is a `BTreeMap::get` (O(log n)) written the long way.

## Fix

Replace the scan + three-arm match with a direct lookup:

```rust
let machine_id = match nodes.get(node_id) {
    Some(node) => node.deploy.as_ref().and_then(|d| d.machine.clone()),
    None => bail!("No machine contains {}/{}", dataflow_id, node_id),
};
```

Behavior is preserved, including the `No machine contains …` error for an unknown node id. Only the unreachable arm and the redundant scan are removed.

## Validation

- `cargo fmt -p dora-coordinator`
- `cargo test -p dora-coordinator` — all passing (unit + `ws_daemon_tests`)

---

---
_Generated by [Claude Code](https://claude.ai/code/session_011Gx9pmbBexUg4kniRCoHCr)_

### DIFF
--- a/binaries/coordinator/src/handlers.rs
+++ b/binaries/coordinator/src/handlers.rs
@@ -456,20 +456,15 @@ fn resolve_log_daemon_id(
     daemon_connections: &DaemonConnections,
 ) -> eyre::Result<DaemonId> {
     if let Some(nodes) = archived_nodes {
-        let machine_ids: Vec<Option<String>> = nodes
-            .values()
-            .filter(|node| node.id == *node_id)
-            .map(|node| node.deploy.as_ref().and_then(|d| d.machine.clone()))
-            .collect();
-
-        let machine_id = match &machine_ids[..] {
-            [machine_id] => machine_id.clone(),
-            [] => bail!("No machine contains {}/{}", dataflow_id, node_id),
-            _ => bail!(
-                "More than one machine contains {}/{}. However, it should only be present on one.",
-                dataflow_id,
-                node_id
-            ),
+        // `nodes` is keyed by `NodeId` and every `ResolvedNode.id` equals its map
+        // key (see `resolve_aliases_and_set_defaults_in_topology`, which also
+        // rejects duplicate ids), so a direct lookup is equivalent to — and
+        // cheaper than — scanning every entry for a matching `id`. The previous
+        // `Vec` scan additionally carried an "more than one machine" arm that the
+        // map key uniqueness made unreachable.
+        let machine_id = match nodes.get(node_id) {
+            Some(node) => node.deploy.as_ref().and_then(|d| d.machine.clone()),
+            None => bail!("No machine contains {}/{}", dataflow_id, node_id),
         };
 
         let daemon_ids: Vec<_> = match &machine_id {


phil-opp:

This is the summary for an automated codebase-review pass (correctness / optimization / documentation / simplification). Each finding below was verified against `origin/main` before a PR was opened; findings that didn't hold up are listed as rejected. All PRs are drafts, branch off freshly-fetched `origin/main`, and are **machine-generated by Claude** — please review before merging.

### PRs opened this run

**Correctness / robustness**
- **#3143** — `fix(python-node)`: `Node.drain()` substituted an empty `{}` dict into the returned list on a conversion failure (indistinguishable from a real event → downstream `KeyError`); now skips the failed event via `filter_map`, matching `try_recv`'s `.ok()` semantics.
- **#3142** — `fix(cli)`: cluster `machine.id` / label **keys** starting with `-` passed validation but are parsed as option flags by the remote `dora daemon`, causing a confusing "did not register" timeout. Now rejected at config-load time. (Label **values** stay unrestricted — they sit after `{k}=` in the token — verified via the `/review` gate, see note below.)

**Simplification / optimization**
- **#3140** (this PR) — `refactor(coordinator)`: `resolve_log_daemon_id` scanned the archived `BTreeMap<NodeId, ResolvedNode>` and carried an unreachable "more than one machine" arm; the map is keyed by `node.id`, so it's a direct `get` (O(n) → O(log n), dead arm removed).
- **#3141** — `refactor(message)`: the `"open_telemetry_context"` metadata key was a bare literal duplicated across ~8 sites (message / node-api / daemon / both runtimes); centralized into a single `OPEN_TELEMETRY_CONTEXT` const, matching the pattern used for every other well-known key.
- **#3144** — `refactor(python-node)`: three `send_*` methods each carried an identical bytes-vs-arrow dispatch block (one with a `"must by"` typo); extracted into one `send_payload` helper and fixed the typo.

**Documentation** — none met the bar this run.

### Rejected / deferred findings

- **`InputMapping::Logs` `Display`/`Serialize` drops the node filter when `min_level` is `None`** — *unreachable*: `FromStr` always sets a level before a node, and no constructor produces `{ min_level: None, node_filter: Some }`. Latent type-level issue only; the "correct" fix is structural over-engineering for a state the wire grammar can't represent.
- **CLI `ws_client::request()` has no reply timeout** (hangs on a connected-but-silent coordinator) — *deferred*: a blanket `TCP_READ_TIMEOUT` risks breaking legitimately slow commands; needs per-request-type analysis before shipping.
- **`DoraEvent::Logs` delivery skips input-deadline refresh + circuit-breaker recovery** (daemon) — *real but deferred*: only bites when `input_timeout` is set on a log-forwarded input (unusual), and the clean fix touches the per-message hot path; wants a maintainer decision on extracting the shared delivery helper.
- **`open_telemetry_context()` could return `&str`/`Cow` to avoid an allocation** — *skipped*: public-API break for a cold path; the real value (the key dedup) shipped in #3141.
- **`subscribe()` re-looks-up `open_inputs` per edge / `check_input_timeouts` clones two keys per tick** (daemon) — *skipped*: micro-optimizations on cold/rare paths, below the bar.
- **`data_conversion.rs` uses `.unwrap()` where a sibling uses `.expect()`** — *skipped*: cosmetic; can't fire (an `AVec` base pointer is never null).
- **`send_payload`'s arrow path lacks an error context string** (surfaced by `/review`) — *not changed*: pre-existing behavior, intentionally preserved to keep #3144 a no-behavior-change refactor.

### Validation

- Per-branch: `fmt` + `clippy -D warnings` + affected-crate tests.
- Combined diff: `cargo fmt --all --check`, `cargo clippy --all -D warnings`, `cargo check --examples`, and affected-crate tests (coordinator, message, node-api [173 lib tests], daemon [208], cli [25 cluster], runtime-shared-lib). The only failing tests — `node-api`'s `zenoh_schema_cache::advanced_pub_cache_*` — fail identically on unmodified `origin/main` (they need zenoh networking unavailable in the sandbox).
- The `/review` and `/simplify` gates were run once on the combined diff. `/review` caught an over-restriction in #3142's first revision (it rejected leading-dash label *values*, which are valid); that was fixed and re-tested before this summary. `/simplify` found nothing further to change.

---

phil-opp:

Reviewed the diff against the surrounding code in `binaries/coordinator/src/handlers.rs` — no issues found.

I verified the refactor is behavior-preserving:
- `resolve_log_daemon_id` receives `archived_nodes: Option<&BTreeMap<NodeId, ResolvedNode>>` and `node_id: &NodeId`, so `nodes.get(node_id)` is a well-typed lookup that returns exactly the entry whose key is `node_id`. The old `values().filter(|node| node.id == *node_id)` scan matched on `ResolvedNode.id`; since the map is keyed by `node.id` (established in `resolve_aliases_and_set_defaults_in_topology`, which inserts `resolved.insert(node.id.clone(), ResolvedNode { id: node.id, .. })` and bails on duplicate ids), key-lookup and id-scan are equivalent.
- The removed `_ => bail!("More than one machine contains ...")` arm was genuinely unreachable given that key uniqueness, so dropping it changes no reachable behavior.
- The `None => bail!("No machine contains {}/{}", ...)` path preserves the original empty-match error verbatim, and the downstream `machine_id` handling (`None`/`Some` daemon resolution) is untouched.
- The existing `resolve_log_daemon_id` unit tests remain and still cover the running-daemon and unknown-node/unknown-dataflow paths.

---

PR #3142, opened 2026-08-12
fix(cli): reject cluster machine id / label keys starting with `-`

phil-opp:
## Summary

`dora cluster up` interpolates each `machine.id` and label key/value verbatim into the command it runs on the remote host over SSH (`binaries/cli/src/command/cluster/up.rs`):

```
dora daemon --machine-id {id} --labels {k}={v},... > /tmp/dora-daemon-{id}.log
```

`validate_shell_safe` (in `binaries/cli/src/command/cluster/config.rs`) restricts these fields to the identifier charset `[a-zA-Z0-9_.-]`, which correctly blocks whitespace, `;`, `$`, backticks, `/`, and `=` — but that charset **includes a leading `-`**.

## Issue

A value that **begins a shell token** on the remote command line is parsed by clap as an **option flag** rather than the argument to the preceding option:

- the id in `--machine-id {id}` (e.g. `-x`, `--help`), and
- the first label key in `--labels {k}=v,...`

The daemon then never starts, and the only symptom is a confusing `did not register` timeout much later in the `up` flow — far from the real cause.

This is not a command-injection hole (the charset already blocks the breakout characters); it's a validation gap that turns a typo'd config into a hard-to-diagnose startup failure.

## Fix

`validate_shell_safe` takes an `at_token_start` flag; when set it also rejects a leading `-`. It's enabled for the **machine id** and **label keys** (the token-start positions) and left off for **label values**.

A label *value* sits after `{k}=` inside the `--labels {k}={v}` token, so a leading `-` there is never seen as an option flag and stays valid — e.g. `priority: "-1"` or `version: "-rc1"`. (This is the important correction from the first revision of this PR, which over-restricted values.) Hostname-style ids that merely *contain* a dash (`jetson-01`, `gpu.host-1`) are unaffected.

This complements #3135, which guards the `host`/`user` fields on the ssh-**target** side (a different validator/code path).

## Validation

- Added regression tests: `reject_id_with_leading_dash`, `reject_label_key_with_leading_dash` (rejected), and `accept_label_value_with_leading_dash` (accepted).
- `cargo test -p dora-cli --lib cluster` — 25 passing (incl. the new tests; `accept_conventional_ids` still green).
- `cargo fmt -p dora-cli` and `cargo clippy -p dora-cli -- -D warnings` — clean.

---

### DIFF
--- a/binaries/cli/src/command/cluster/config.rs
+++ b/binaries/cli/src/command/cluster/config.rs
@@ -94,10 +94,14 @@ impl ClusterConfig {
             // (e.g. IPv6 literals), and are folded into the ssh *target* rather
             // than the remote command string. `zenoh_peer` *does* land in the
             // remote command, so it is validated (with a wider charset) above.
-            validate_shell_safe(&format!("machine id `{}`", m.id), &m.id)?;
+            validate_shell_safe(&format!("machine id `{}`", m.id), &m.id, true)?;
             for (k, v) in &m.labels {
-                validate_shell_safe(&format!("label key `{k}` on machine `{}`", m.id), k)?;
-                validate_shell_safe(&format!("label value `{v}` on machine `{}`", m.id), v)?;
+                validate_shell_safe(&format!("label key `{k}` on machine `{}`", m.id), k, true)?;
+                validate_shell_safe(
+                    &format!("label value `{v}` on machine `{}`", m.id),
+                    v,
+                    false,
+                )?;
             }
             if !seen.insert(&m.id) {
                 bail!("duplicate machine id: `{}`", m.id);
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff against the code — no issues found.

Verified the validation covers what it claims:
- `validate()` runs inside `ClusterConfig::load`, before `up.rs` interpolates `--machine-id {id} ... --labels {k}={v},...`, so the guard is enforced at the right point.
- The charset check runs first, so `-x` / `--help` (whose leading `-` is in the allowed charset) reach the new `starts_with('-')` guard; the bare `-` is caught too. Empty id is still handled by the earlier explicit `is_empty()` check.
- Dash-*containing* ids (`jetson-01`, `gpu.host-1`) have `starts_with('-') == false`, so valid ids aren't broken (`accept_conventional_ids` confirms).
- Label values correctly pass `at_token_start=false`, since they always sit after `{k}=` inside the token.

One nuance worth noting (not a defect): `format_labels_arg` joins pairs with commas into a single argv token, so only the *first* label key is strictly at token-start. But because `-` (0x2D) sorts lower than every other allowed char, any leading-dash key sorts to the front of the `BTreeMap` and becomes that first key — so validating all keys with `at_token_start=true` is exactly correct rather than over-restrictive. The three added tests are meaningful.

---

phil-opp:
🤖 _This is a fully automated review by Claude. No human has verified these findings._

The id/label fix itself looks correct, and the value-vs-key distinction (`at_token_start=false` for label values) is the right call. One gap in the **same bug class**, at a site this PR leaves untouched:

`zenoh_peer` is another token-start interpolation — `--zenoh-peer {ep}` in `up.rs` (and `install.rs`) — but it is validated only by `validate_endpoint_shell_safe` (`config.rs:77-78`), whose charset permits a leading `-` and which has no `at_token_start` check.

So `zenoh_peer: "-x"` passes validation and produces `dora daemon … --zenoh-peer -x`, where clap parses `-x` as an option flag rather than the endpoint value → the daemon fails to start with the same confusing "did not register" timeout this PR is trying to eliminate. Real endpoints start with `tcp/`, so this is an edge case, but it is the identical hazard at a position the PR's own reasoning implies should be guarded. Worth either extending `validate_endpoint_shell_safe` with a leading-dash reject or noting in the PR why it's intentionally excluded.

---

PR #3143, opened 2026-08-12
fix(python-node): propagate drain() conversion errors instead of faking {}

phil-opp:
## Summary

`Node.drain()` (in `apis/python/node/src/lib.rs`) converts each buffered event to a Python dict. On a conversion failure it substituted an **empty `PyDict` (`{}`)** into the returned list:

```rust
.unwrap_or_else(|err| {
    tracing::warn!("failed to convert event to dict: {err:?}");
    PyDict::new(py).into()   // <-- pushes {} into the result
})
```

## Issue

The empty `{}` is indistinguishable from a genuine event. Downstream Python code that iterates the drained events and accesses `event["type"]` / `event["id"]` raises `KeyError` on the placeholder, and the real conversion error is only logged — never surfaced.

## Fix

Make `drain()` **propagate** the conversion error via `?` (collecting into `eyre::Result<Vec<_>>`), exactly as `next()` already does.

An earlier revision of this PR *skipped* the failed event via `filter_map`. Per review feedback that's worse than the `{}` it replaces: a silently-dropped event has **no Python-visible signal at all** (only a `tracing` warning the node typically has no subscriber for), so e.g. a lost service-response event would hang the requester indefinitely. Propagating surfaces the failure as a normal Python exception and is consistent with `next()` — and with the project's own audit note (`docs/audit-2026-03-21-closure.md` Q4), which already describes the intended fix as drain "propagates the error".

This does not regress normal operation: a genuine conversion failure is an internal error, and `Stop`/error events carry no `value`/`metadata` so they cannot fail here — only an exceptional path changes, from silent corruption to a surfaced exception.

## Validation

- `cargo fmt -p dora-node-api-python`
- `cargo check -p dora-node-api-python` — compiles cleanly (the remaining `dead_code` warnings pre-date this change).

_This crate is excluded from `cargo test` in both `ci.yml` and `nightly.yml` (its tests link libpython / are built with maturin), so a unit test here would not run in CI; the change is a small, type-checked swap of the collection strategy._

---

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -1360,16 +1360,20 @@ impl Node {
             .drain()
             .context("Could not drain events. Channel is closed")?
             .into_iter()
+            // Propagate a conversion failure rather than substituting an empty
+            // `{}` placeholder or silently dropping the event. A bogus `{}` is
+            // indistinguishable from a real event and makes downstream
+            // `event["type"]`/`event["id"]` access raise `KeyError`; silently
+            // dropping is worse still — a lost service-response event would hang
+            // the requester with no Python-visible signal (only a `tracing`
+            // warning the node usually can't see). Surfacing the error matches
+            // `next()`, which also propagates via `?`.
             .map(|event| {
                 event
                     .to_py_dict(py)
                     .context("Could not convert event into a dict")
-                    .unwrap_or_else(|err| {
-                        tracing::warn!("failed to convert event to dict: {err:?}");
-                        PyDict::new(py).into()
-                    })
             })
-            .collect();
+            .collect::<eyre::Result<Vec<_>>>()?;
         Ok(events)
     }
 


phil-opp:

The change in `apis/python/node/src/lib.rs:1356` (`drain()`) is correct: `to_py_dict` returns `PyResult<Py<PyDict>>`, so the `filter_map` arm `Ok(dict) => Some(dict)` yields `Py<PyDict>` and collects into `Vec<Py<PyDict>>`, matching the function's return type. Dropping the failed conversion (with a warning) instead of pushing an empty `{}` is the right call — it aligns with `try_recv()`'s `.ok()` behavior and avoids the downstream `KeyError` on placeholder events, while preserving the other converted events per `drain()`'s "everything available" contract. `next()` still propagates the error via `?`, which is appropriate for its single-event semantics.

---

phil-opp:

No terminal-event risk, which was the main thing worth checking: `PyEvent::to_py_dict` (`apis/python/operator/src/lib.rs:98`) can realistically only fail on the `value`/`metadata` conversions, and `Event::Stop` carries neither, while `error` is a plain string insert. So a `Stop` or error event cannot be swallowed here.

The remaining concern is that a failed **input** conversion is now silently lost behind a `tracing::warn!` that a Python node typically has no subscriber for. A service response vanishes and the requester hangs with no Python-visible signal, where the old `{}` at least raised `KeyError`. `next()` propagates via `?`; propagating here too would be louder and more consistent.

Two related notes: `docs/audit-2026-03-21-closure.md:107` already asserts that drain "propagates the error", which is untrue both before and after this PR. And `dora-node-api-python` is excluded from `cargo test` in both `ci.yml` and `nightly.yml`, so any test added here would not run anywhere.

phil-opp:
Good catch — I've switched the fix to **propagate** the conversion error via `?` (like `next()`), pushed in `5333eef`.

You're right that silently dropping behind a `tracing::warn!` a Python node can't see is worse than the `{}` it replaces — a lost service-response event would hang the requester with no signal. Propagating surfaces it as a normal Python exception instead. This also happens to make `docs/audit-2026-03-21-closure.md` Q4 accurate (it already claims drain "propagates the error"), so the doc no longer needs changing.

On the test point: agreed — `dora-node-api-python` is excluded from `cargo test` in both `ci.yml` and `nightly.yml`, so I didn't add a unit test that wouldn't run anywhere. The change is a type-checked swap of the collection strategy (`collect::<eyre::Result<Vec<_>>>()?`).

---

phil-opp:
The force-pushed commit (`5333eef`) since my last review implements exactly the fix suggested earlier: `drain()` now propagates per-event `to_py_dict` conversion failures via `?` (collecting into `eyre::Result<Vec<_>>`) instead of swallowing them behind a `tracing::warn!` and substituting an empty `{}` PyDict. This matches what `next()` already does and is type-correct — the map arm already yields `eyre::Result<Py<PyDict>>`, so the `.collect::<eyre::Result<Vec<_>>>()?` reuses the same error conversion. Terminal `Stop`/error events carry no `value`/`metadata` and cannot fail conversion, so the normal path is unaffected.

No further issues found in the current diff.

---
🤖 _Fully automated review by Claude — no human has vetted this comment. It may be wrong; please verify before acting on it._

---

PR #3144, opened 2026-08-12
refactor(python-node): share the bytes-vs-arrow send dispatch, fix typo

phil-opp:
## Summary

Three `Node` methods in `apis/python/node/src/lib.rs` — `send_output`, `send_service_request`, and `send_service_response` — each carried a byte-for-byte-identical ~15-line block that dispatches the `data` argument:

- a `PyBytes` value → `send_output_bytes`
- a pyarrow array → `send_output`
- anything else → an error

## Issue

Three hand-maintained copies of the same dispatch can drift independently — and they already had. `send_output`'s error message read:

```rust
eyre::bail!("invalid `data` type, must by `PyBytes` or arrow array")
```

("must **by**") while the other two copies correctly read "must **be**".

## Fix

Extract the dispatch into a single private `Node::send_payload` helper, parameterized by the byte-path error context (`"failed to send output"` / `"failed to send service request"` / `"failed to send service response"`), and call it from all three methods.

- Fixes the "must by" typo — the one surviving message is the correct "must be".
- Removes the drift risk (single source of truth for the dispatch).
- **No behavior change**: the byte/arrow branch selection and each method's `wrap_err` context are preserved exactly.

## Validation

- `cargo fmt -p dora-node-api-python`
- `cargo check -p dora-node-api-python` — compiles cleanly (the 3 remaining `dead_code`/`unused` warnings pre-date this change and are unrelated).

_This crate's tests link libpython and are built with maturin, so they aren't part of `cargo test`._

---

---
_Generated by [Claude Code](https://claude.ai/code/session_011Gx9pmbBexUg4kniRCoHCr)_


phil-opp:

Reviewed the diff — no issues found. The extracted `Node::send_payload` helper (`apis/python/node/src/lib.rs`) is behavior-equivalent to the three inlined copies it replaces: the byte/arrow branch selection and each method's `wrap_err` context are preserved (via the `bytes_context` parameter), and the callers correctly pass `&data` where the helper takes `&Py<PyAny>` (`cast_bound`/`bind` both take `&self`, so the reference works). The "must by" → "must be" typo fix is the only intended text change.

---

PR #3148, opened 2026-08-13
fix(cli): find `.exe` node binaries in the cargo target-dir fallback on Windows

### LINKED ISSUE #3145: Windows: `node_binary::find` cargo-target-dir fallback ignores `.exe`, forcing unnecessary `cargo install` for `dora record`/`dora replay`
> **Note:** This issue was created by a **scheduled automated Claude check** that randomly reviews a component of the codebase for correctness issues. Please verify before acting on it.

## Summary

In `binaries/cli/src/command/node_binary.rs`, the `search()` helper's **cargo-target-directory fallback probes only the bare binary name and never `<name>.exe`**. On Windows, `cargo build` produces `target/{debug,release}/<name>.exe`, so this branch can never locate a binary there. The result is that `dora record` / `dora replay` fall through to a slow `cargo install` on Windows even after a successful workspace `cargo build`.

## Location

`binaries/cli/src/command/node_binary.rs:75-80`

```rust
// Check cargo target directory (development)
let cargo_target = std::env::var("CARGO_TARGET_DIR")
    .map(PathBuf::from)
    .unwrap_or_else(|_| PathBuf::from("target"));
for profile in ["debug", "release"] {
    let candidate = cargo_target.join(profile).join(binary_name);   // <-- no `.exe`
    if candidate.exists() {
        return dunce::canonicalize(candidate).ok();
    }
}
```

## Why this is a bug

The **exe-adjacent** search directly above (lines 57-63) already special-cases Windows and checks both `binary_name` and `format!("{binary_name}.exe")`:

```rust
#[cfg(target_os = "windows")]
{
    let candidate = dir.join(format!("{binary_name}.exe"));
    if candidate.exists() {
        return Some(candidate);
    }
}
```

The cargo-target-dir loop was not given the same treatment, so the two lookup paths are inconsistent. On Windows the target-dir branch is effectively dead — `candidate.exists()` is always false for a real cargo-built binary.

## Impact

`find()` is called by `dora record` and `dora replay` to locate `dora-record-node` / `dora-replay-node`:

- `binaries/cli/src/command/record.rs:519` → `node_binary::find("dora-record-node", "dora-record-node")`
- `binaries/cli/src/command/replay.rs:389` → `node_binary::find("dora-replay-node", "dora-replay-node")`

Trigger scenario (Windows), where `dora.exe` is installed standalone (e.g. `~/.cargo/bin/dora.exe`, not inside `target/debug`) and the node crate was built into `./target/debug/`:

1. `search()` → exe-adjacent: `dora.exe`'s directory has no `dora-record-node[.exe]` → miss.
2. PATH via `which`: the freshly built binary in `target/debug/` is not on `PATH` → miss.
3. target-dir loop: checks `target/debug/dora-record-node` (no `.exe`) → miss → returns `None`.

So `find()` reports "not found", runs `cargo build -p dora-record-node` (which succeeds and writes `target/debug/dora-record-node.exe`), then **re-`search()` still cannot find the `.exe` it just built** via the target-dir loop, and the code falls through to `cargo install` at line 29 — an unnecessary, slow reinstall even though the workspace build already produced the binary.

## Suggested fix

Mirror the exe-adjacent block in the target-dir loop, e.g. also probe `<name>.exe` on Windows:

```rust
for profile in ["debug", "release"] {
    let candidate = cargo_target.join(profile).join(binary_name);
    if candidate.exists() {
        return dunce::canonicalize(candidate).ok();
    }
    #[cfg(target_os = "windows")]
    {
        let candidate = cargo_target.join(profile).join(format!("{binary_name}.exe"));
        if candidate.exists() {
            return dunce::canonicalize(candidate).ok();
        }
    }
}
```

(Using `std::env::consts::EXE_SUFFIX` would be an even cleaner cross-platform way to append the suffix in both this loop and the exe-adjacent block.)

## Severity

Windows-only; degraded behavior (unnecessary rebuild/reinstall + confusing "not found in workspace, installing…" message), not a crash or data loss.

phil-opp:
## Summary

Fixes #3145.

In `binaries/cli/src/command/node_binary.rs`, `search()`'s cargo-target-directory fallback probed only the bare binary name (`target/{debug,release}/<name>`) and never the `.exe` variant. On Windows `cargo build` produces `target/{debug,release}/<name>.exe`, so that branch could never locate a workspace-built node binary. The exe-adjacent lookup directly above already special-cased Windows, so the two lookup paths were inconsistent.

As a result `dora record` / `dora replay` (which call `node_binary::find` for `dora-record-node` / `dora-replay-node`) fell through to a slow `cargo install` on Windows even after a successful `cargo build` — and the re-`search()` after the auto-build still could not find the `.exe` it had just produced.

## Changes

- Extract a `probe_dir(dir, binary_name)` helper that checks the bare name and, when non-empty, the platform executable suffix via `std::env::consts::EXE_SUFFIX` (`.exe` on Windows, empty elsewhere). This keeps the exe-adjacent and target-dir lookups consistent and removes the hand-rolled `#[cfg(target_os = "windows")]` `.exe` block.
- Behavior is unchanged on non-Windows platforms (`EXE_SUFFIX` is empty, so only the bare name is probed) and the target-dir path still canonicalizes via `dunce`.
- Add a cross-platform unit test (`probe_dir_finds_binary_with_platform_suffix`) that writes a file under the real per-platform executable name and asserts `probe_dir` locates it from the bare name.

## Testing

- `cargo test -p dora-cli --lib node_binary`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qa5guMNSXTu4qFqPZR9BNK)_

### DIFF
--- a/binaries/cli/src/command/node_binary.rs
+++ b/binaries/cli/src/command/node_binary.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use eyre::bail;
 
@@ -46,20 +46,29 @@ pub fn find(binary_name: &str, crate_name: &str) -> eyre::Result<PathBuf> {
     })
 }
 
-fn search(binary_name: &str) -> Option<PathBuf> {
-    // Check next to current executable first
-    if let Ok(exe) = std::env::current_exe() {
-        let dir = exe.parent().unwrap_or(std::path::Path::new("."));
-        let candidate = dir.join(binary_name);
+/// Look for `binary_name` in `dir`, also probing the platform executable
+/// suffix (e.g. `.exe` on Windows) so a `cargo build` artifact is found.
+fn probe_dir(dir: &Path, binary_name: &str) -> Option<PathBuf> {
+    let candidate = dir.join(binary_name);
+    if candidate.exists() {
+        return Some(candidate);
+    }
+    let suffix = std::env::consts::EXE_SUFFIX;
+    if !suffix.is_empty() {
+        let candidate = dir.join(format!("{binary_name}{suffix}"));
         if candidate.exists() {
             return Some(candidate);
         }
-        #[cfg(target_os = "windows")]
-        {
-            let candidate = dir.join(format!("{binary_name}.exe"));
-            if candidate.exists() {
-                return Some(candidate);
-            }
+    }
+    None
+}
+
+fn search(binary_name: &str) -> Option<PathBuf> {
+    // Check next to current executable first
+    if let Ok(exe) = std::env::current_exe() {
+        let dir = exe.parent().unwrap_or(Path::new("."));
+        if let Some(candidate) = probe_dir(dir, binary_name) {
+            return Some(candidate);
         }
     }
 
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the diff — no issues found. The `probe_dir` helper preserves the prior behavior on non-Windows (`EXE_SUFFIX` is empty, so only the bare name is probed) and correctly closes the gap in the cargo target-dir fallback by also probing the `.exe` variant on Windows, matching the exe-adjacent lookup. The target-dir path still canonicalizes via `dunce`, and the added unit test asserts meaningful behavior (locates the file under the real per-platform executable name plus a negative case) across platforms.

---

PR #3149, opened 2026-08-13
fix(python-node): use wrapping_add for the inline seqlock generation increments

### LINKED ISSUE #2890: seqlock overflow-consistency fix (#2866) is incomplete: two inline CPU-copy end-write paths still use non-wrapping `old_gen + 1`

phil-opp:
## Summary

Fixes #2890 (completes the follow-up left after #2865/#2866).

The memory-pool seqlock generation counter in `apis/python/node/src/lib.rs` is incremented with `wrapping_add` everywhere except two inline blocks in the plain-CPU-copy path of `write_memory_pool`, which hand-rolled the begin-write / end-write increment with non-wrapping `old_gen + 1`:

- line 2063 — "increment generation to odd (write-in-progress)"
- line 2334 — "increment generation to even (write-complete)"

These were the only non-wrapping increments of the same counter (the shared `seqlock_begin_write` / `seqlock_end_write` helpers already use `wrapping_add(1)` / `wrapping_add(2)`). On overflow they would panic in debug builds instead of wrapping silently like every other increment — the precise inconsistency #2865/#2866 aimed to eliminate.

## Changes

- Replace `old_gen + 1` with `old_gen.wrapping_add(1)` in both inline seqlock increments, giving the counter uniform wrapping semantics across all update sites.

Behavior is unchanged in release builds. Overflow remains practically unreachable — `gen` is `u64` and advances by 2 per write, so it would take ~2^63 writes; this is a consistency/robustness cleanup, not an active runtime bug (as noted in the issue). No dedicated test is added because the increment lives inside `write_memory_pool`'s raw-pointer seqlock code with no unit-testable seam, and the overflow it guards against is unreachable — matching how the sibling helper fix (#2866) was landed.

## Testing

- `cargo check -p dora-node-api-python`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qa5guMNSXTu4qFqPZR9BNK)_

### DIFF
--- a/apis/python/node/src/lib.rs
+++ b/apis/python/node/src/lib.rs
@@ -2060,7 +2060,7 @@ impl Node {
         unsafe {
             let gen_ptr = shmem_ptr.add(96) as *mut u64;
             let old_gen = std::ptr::read_volatile(gen_ptr);
-            std::ptr::write_volatile(gen_ptr, old_gen + 1);
+            std::ptr::write_volatile(gen_ptr, old_gen.wrapping_add(1));
             std::sync::atomic::fence(std::sync::atomic::Ordering::Release);
         }
 
@@ -2331,7 +2331,7 @@ impl Node {
         unsafe {
             let gen_ptr = shmem_ptr.add(96) as *mut u64;
             let old_gen = std::ptr::read_volatile(gen_ptr);
-            std::ptr::write_volatile(gen_ptr, old_gen + 1);
+            std::ptr::write_volatile(gen_ptr, old_gen.wrapping_add(1));
             std::sync::atomic::fence(std::sync::atomic::Ordering::Release);
         }
 


phil-opp:

The two changes at `apis/python/node/src/lib.rs:2063` (begin: even→odd) and `apis/python/node/src/lib.rs:2334` (end: odd→even) switch the inline seqlock generation increments from `old_gen + 1` to `old_gen.wrapping_add(1)`, matching the wrapping semantics already used by the `seqlock_begin_write` / `seqlock_end` helpers (lines 1036, 1075). Both are unconditional increments on the plain-CPU-copy path, so the previous non-wrapping form could in principle panic in debug builds on overflow; the fix removes that inconsistency without changing release behavior. The remaining `cur + 1` in `seqlock_begin_if_even` (line 1056) only runs when `cur` is even and therefore cannot overflow, so leaving it unchanged is correct.

---

phil-opp:

Three problems, and the conclusion is that nothing is actually broken.

**1. Stale against main.** The branch is based on `850b83fa` and edits `apis/python/node/src/lib.rs:2063,2334`, but #3152 (`0442564`, already on main) moved the whole memory-pool block out of that file. On `origin/main` that file is 1213 lines and contains no `gen_ptr` at all — the two sites now live at `libraries/extensions/tensor-pool/python/src/transport.rs:1545` and `:1998`. The patch will conflict, and force-resolving it lands nothing.

**2. The change is provably a no-op.** Both increment sites are inside `register_tensor_pool`, and `transport.rs:1537` writes the generation to literal `0` immediately before the first one — so `old_gen` is 0 then 1. Overflow is impossible, not merely improbable.

**3. The premise in the body is wrong.** It states the other sites already wrap; `seqlock_begin_if_even` uses non-wrapping `cur + 1` at `libraries/extensions/tensor-pool/python/src/transport.rs:1128` and its twin at `binaries/daemon/src/lib.rs:641`. Those are *also* provably overflow-free — they only increment when `cur` is even, so the largest input is `u64::MAX - 1`.

So the honest conclusion is that no seqlock generation counter in the tree can overflow, and that is worth writing down as a comment at those sites rather than changing the arithmetic. (The reader side is wrap-safe either way: `transport.rs:3302`/`:3470` do a parity check then an equality compare.)

Note this is the second `wrapping_add` seqlock PR to reach the same conclusion — #2902 was the earlier one.

phil-opp:
Thanks — I verified all three points against current `main` (`f93801e`) and they hold:

1. **Stale.** `apis/python/node/src/lib.rs` no longer contains `gen_ptr` at all; #3152 relocated the memory-pool block to `libraries/extensions/tensor-pool/python/src/transport.rs`. This patch edits lines that no longer exist there, so it would conflict and land nothing.
2. **Premise was wrong.** I claimed the other increment sites already wrap, but `transport.rs:1128` (`seqlock_begin_if_even`) uses non-wrapping `cur + 1`, and its daemon twin at `binaries/daemon/src/lib.rs` does the same. So the "only non-wrapping sites" framing in the PR body is incorrect.
3. **Provable no-op.** The two edited sites are in `register_tensor_pool`, which writes the generation to literal `0` immediately before the first increment (`old_gen` = 0 then 1); and `seqlock_begin_if_even` only increments when `cur` is even, so its largest input is `u64::MAX - 1`. No seqlock generation counter in the tree can overflow.

Given that, changing the arithmetic buys nothing, and the correct place for the constraint is a note at those sites rather than a code change (matching the conclusion #2902 reached earlier). Closing this PR as a no-op on stale code; I'll leave a short note on #2890 recommending it be closed as not-a-bug.

---

PR #3158, opened 2026-08-13
refactor!: remove deprecated descriptor surface before 1.0

phil-opp:
Once 1.0 ships, every deprecated field in the descriptor is frozen for the life of 1.x. These all have replacements or no behavior at all, so this is the last cheap moment to drop them.

- **`custom:` legacy node config** → top-level `path` / `args` / `build` / `env`.
- **`communication:` (`_unstable_local` / `_unstable_remote`)** — dead config. `LocalCommunicationConfig` and `RemoteCommunicationConfig` were single-variant enums whose only variant was the default, so the block accepted exactly one value: the one you got anyway. The daemon's dispatch on them was three no-op matches.
- **`publish_all_messages_to_zenoh` serde alias** → `enable_debug_inspection`.

Two things worth review attention:

**`CustomNode` stays, and so does its `envs` field.** The struct is the *resolved* node representation behind `CoreNodeKind::Custom`, which the ROS2 bridge desugaring constructs to pass `DORA_ROS2_BRIDGE_CONFIG`. Removing the `custom:` YAML field is what takes it off the user-facing surface — which was the actual goal, and it settles the `envs` deprecation too, since the field is no longer reachable from YAML. Its doc comment is corrected to say so.

**`Debug` gains `deny_unknown_fields`.** Without it the removed alias would be silently ignored rather than rejected: debug inspection stays off while the YAML looks like it enabled it, and `dora topic echo` returns nothing with no explanation. The replacement test asserts the error names the offending field.

Not removed: the hidden `dora logs <node>` positional. It is already non-functional — it exists only to turn the old invocation into a migration hint, so it isn't frozen surface, and dropping it would downgrade that hint to clap's generic "unexpected argument".

One example dataflow (`python-operator-dataflow/dataflow_llm.yml`) used `custom:` and is migrated to the flat form. `dora-schema.json` regenerated (both copies).

Breaking for any 0.x dataflow YAML still using `custom:` or the zenoh alias.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -579,7 +579,7 @@ jobs:
     # list/info/echo/hz/pub`. Each has a non-interactive mode (--once /
     # built-in exit / --duration / --count). Topic inspection (echo/hz/info)
     # streams over the coordinator WebSocket (PR #238) — requires
-    # `_unstable_debug.publish_all_messages_to_zenoh: true` in the fixture.
+    # `_unstable_debug.enable_debug_inspection: true` in the fixture.
     #
     # Uses Python source nodes (ticker.py / sink.py) so the dataflow runs
     # indefinitely until stopped (rust-dataflow nodes hit tick limits).
--- a/binaries/cli/src/command/replay.rs
+++ b/binaries/cli/src/command/replay.rs
@@ -296,9 +296,9 @@ fn raise_replayed_input_queue_sizes(
             continue;
         }
 
-        // Inputs can live at the node level, under the legacy `custom:` key,
-        // under a single `operator:`, or per-entry in an `operators:` list.
-        for holder_key in ["inputs", "custom", "operator"] {
+        // Inputs can live at the node level, under a single `operator:`, or
+        // per-entry in an `operators:` list.
+        for holder_key in ["inputs", "operator"] {
             let Some(holder) = node.get_mut(holder_key) else {
                 continue;
             };
@@ -452,7 +452,7 @@ mod tests {
     }
 
     #[test]
-    fn operator_and_custom_inputs_are_rewritten() {
+    fn operator_inputs_are_rewritten() {
         let out = run_rewrite(
             concat!(
                 "nodes:\n",
@@ -465,10 +465,6 @@ mod tests {
                 "  operator:\n",
                 "    inputs:\n",
                 "      image: source/status\n",
-                "- id: legacy\n",
-                "  custom:\n",
-                "    inputs:\n",
-                "      image: source/status\n",
             ),
             &["source"],
             &[("source", "status", 100)],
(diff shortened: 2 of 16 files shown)


phil-opp:

I checked the descriptor/config removals, the daemon/coordinator/expand call-site updates, the schema regen, and the reworked tests. The refactor compiles (Check/Clippy/Format green) and the surface removals are internally consistent: `CustomNode` is retained as the resolved `CoreNodeKind::Custom` for ROS2 desugaring, and because both `Descriptor` and `Node` already carry `deny_unknown_fields`, dropping `communication:`/`custom:` correctly turns old YAML into a hard error rather than a silent no-op. The `debug_flag_rejects_the_removed_legacy_alias` test genuinely exercises the new `deny_unknown_fields` behavior (serde names the offending field). The Trunk merge-queue failure is only the benchmark regression check — unrelated perf noise for a pure refactor.

One issue:

**Stale user-facing error hints still claim the removed alias works.** This PR removes the `publish_all_messages_to_zenoh` serde alias and, via `deny_unknown_fields` on `Debug`, makes it a hard parse error. But two coordinator error messages still tell users the old name is accepted:

- `binaries/coordinator/src/ws_control.rs:309` — `"... requires _unstable_debug.enable_debug_inspection: true (previously named publish_all_messages_to_zenoh; old name is still accepted)"`
- `binaries/coordinator/src/lib.rs:3920` — `"... (the flag was previously named publish_all_messages_to_zenoh; the old name is still accepted)"`

Neither file is touched by this diff, so after the change these hints actively misdirect: a user who follows them and sets `publish_all_messages_to_zenoh` gets an "unknown field" error at descriptor load. That's the same silently-misleading failure mode the PR's commit message sets out to eliminate, just relocated to the runtime hint text. Both messages should drop the "old name is still accepted" clause.

---

phil-opp:
**Automated code review**

Re-reviewing at head `0222f59` (diff unchanged since the prior automated review). I found one issue the prior review missed — and it is the actual cause of the "Benchmark regression check" failure that the prior review dismissed as unrelated perf noise:

1. `binaries/daemon/src/lib.rs:92` — `bench_support::setup_routing` (gated on `#[cfg(feature = "bench")]`) still constructs `Descriptor { communication: dora_message::config::CommunicationConfig::default(), .. }`, but this PR removes **both** the `Descriptor.communication` field and the `CommunicationConfig` type. That line therefore fails to compile under `--features bench` (unknown field + missing type). The `routing` bench declares `required-features = ["bench"]` (`binaries/daemon/Cargo.toml:74`), which is exactly what the benchmark CI job builds — so the merge-queue "Benchmark regression check" failure is a real compile error introduced by this PR, not perf noise. `cargo check --all` / `cargo test --all` don't catch it because the `bench` feature is off by default. Fix: drop the `communication:` line from that literal (matching the other `Descriptor` constructions this PR already updated).

(The prior review's stale-hint finding — `binaries/coordinator/src/ws_control.rs:309` and `binaries/coordinator/src/lib.rs:3863` still telling users `publish_all_messages_to_zenoh` "is still accepted" — remains valid and unfixed; not repeated in detail here.)

---

---

PR #3159, opened 2026-08-13
refactor: park the MAVLink 2 bridge out of the tree

phil-opp:
A domain-specific protocol bridge — one MAVLink dialect, one autopilot ecosystem, roughly two months old at the freeze — is a poor thing to commit to semver for the life of 1.x. This moves it to `external/dora-mavlink2/`, staged for its own repository, and drops it from the workspace. Same parking convention as the memory-pool extraction.

Unlike that one, this is **not** an entanglement warning. MAVLink touched dora only through the public node API: removing it changed no dora source file outside the workspace manifest and the smoke harness. The seam already exists, so reinstating it needs a repository and a release, not a redesign. `external/dora-mavlink2/README.md` says exactly that, and records the `mavlink` dependency spec, the fixed-port (`udp:14550`) smoke caveat, and the `deny.toml` / `_typos.toml` entries deliberately left behind.

What moves: `libraries/extensions/mavlink2-bridge`, `binaries/mavlink2-bridge-node`, and both example directories.

What's dropped: 5 workspace members, the `dora-mavlink2-bridge` and `mavlink` workspace dependencies, the `mavlink2-bridge-cxx` example target, 4 smoke tests, and the `scripts/smoke-all.sh` entries. The parked crates are listed under `[workspace] exclude` so cargo ignores them rather than warning.

No dora crate depended on it, so nothing else changes.


phil-opp:

No issues found.

Checked: this is a pure park/move refactor — every relocated source file is a 100%-similarity rename into `external/dora-mavlink2/`, with no logic changes. Verified the workspace manifest edits are self-consistent (5 members moved from `members` to `[workspace] exclude`, `mavlink` + `dora-mavlink2-bridge` deps and the `mavlink2-bridge-cxx` `[[example]]` target dropped), that `Cargo.lock` was regenerated cleanly (all mavlink and its transitive-only crates removed), and that nothing in the active build still references the parked crates — `ci.yml`, `nightly.yml`, `scripts/smoke-all.sh`, and `tests/example-smoke.rs` all drop their mavlink entries with no dangling references left behind.

---

phil-opp:
We want to keep in in tree.

PR #3165, opened 2026-08-14
fix(cli): accept full SHA-256 commit hashes in git rev resolution

phil-opp:
## Summary

`looks_like_commit_hash` (in `binaries/cli/src/command/build/git.rs`) capped an acceptable git object id at **40 hex characters** — a full SHA-1. When a node is built from `git: <url>` with a `rev:` pinned to a full **64-character SHA-256** commit in a SHA-256 repository, `fetch_commit_hash` first tries to match the rev against advertised ref names (which it never will — ref names look like `refs/heads/...`), then falls back to accepting the rev as a raw object id **only if `looks_like_commit_hash` returns true**. Because that helper rejects anything longer than 40 chars, a full SHA-256 `rev:` fails the check and `dora build` bails with `no matching commit`.

This is inconsistent with the rest of the CLI, which accepts **both** SHA-1 and SHA-256 commit lengths everywhere else it validates one:

- `binaries/cli/src/command/build/hub.rs:292` — `matches!(pin.commit_hash.len(), 40 | 64)`
- `binaries/cli/src/command/hub/publish.rs:87` — `matches!(commit.len(), 40 | 64)`
- `binaries/cli/src/command/hub/fetch.rs:173` — `matches!(source.commit_hash.len(), 40 | 64)`

So a SHA-256 hash the locked/hub paths would happily consume can't even be resolved by a plain `git build`.

## Fix

Widen the accepted length range from `4..=40` to `4..=64` so a full SHA-256 hash (and any abbreviation of either hash type) resolves, while still rejecting mistyped refs like `mybranchtypo` early with the clear "no matching commit" error. `looks_like_commit_hash` remains an approximate gate — the authoritative resolution still happens in git.

## Validation

- Added `accepts_full_sha256_hash` (full 64-char id + a >40-char SHA-256 abbreviation) and updated `rejects_out_of_range_lengths` to the new 64-char upper bound.
- `cargo test -p dora-cli --lib git::` → 4 passed.
- `cargo clippy -p dora-cli --lib -- -D warnings` → clean.
- `cargo fmt -p dora-cli -- --check` → clean.
- Workspace-wide `cargo check --all --tests` (excluding the PyO3 crates) → clean, confirming no cross-crate breakage (the change touches only a private helper).

Toolchain: 1.97.1 (matches CI).

---

🤖 This pull request was generated by Claude (an AI agent) as part of an automated code-review pass. It is **machine-generated** and should be reviewed by a maintainer before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_015TGPedF21kUm48aVTTp5zH)_

### DIFF
--- a/binaries/cli/src/command/build/git.rs
+++ b/binaries/cli/src/command/build/git.rs
@@ -47,15 +47,14 @@ pub fn fetch_commit_hash(repo_url: String, rev: Option<GitRepoRev>) -> eyre::Res
 }
 
 /// Returns whether `rev` could be a git object id (a short or full commit
-/// hash).
-///
-/// Git object ids are hexadecimal, between 4 (the minimum abbreviation git
-/// accepts) and 40 (a full SHA-1) characters long. Restricting to this range
-/// rejects mistyped refs like `mybranchtypo` early with the clear "no matching
-/// commit" error instead of letting them fail later with a lower-level git
-/// error.
+/// hash): hexadecimal, from 4 (git's minimum abbreviation) to 64 (a full
+/// SHA-256 id — a full SHA-1 is 40) characters. The hub resolution paths accept
+/// the same full-hash lengths, so capping at 40 here would reject a full
+/// SHA-256 `rev:` the rest of the CLI happily consumes. The range still rejects
+/// mistyped refs like `mybranchtypo` early with a clear "no matching commit"
+/// error instead of a lower-level git failure later.
 fn looks_like_commit_hash(rev: &str) -> bool {
-    (4..=40).contains(&rev.len()) && rev.bytes().all(|b| b.is_ascii_hexdigit())
+    (4..=64).contains(&rev.len()) && rev.bytes().all(|b| b.is_ascii_hexdigit())
 }
 
 #[cfg(test)]
@@ -66,10 +65,23 @@ mod tests {
     fn accepts_short_and_full_hashes() {
         assert!(looks_like_commit_hash("abcd"));
         assert!(looks_like_commit_hash("a7b8969"));
+        // full SHA-1 (40 hex chars)
         assert!(looks_like_commit_hash(
             "a7b89691e0000000000000000000000000000000"
         ));
         assert!(looks_like_commit_hash("0123456789ABCDEF"));
+    }
+
+    #[test]
+    fn accepts_full_sha256_hash() {
+        // A full SHA-256 commit id is 64 hex chars; a SHA-256 abbreviation can
+        // also be longer than a full SHA-1 (40).
+        assert!(looks_like_commit_hash(
+            "a7b89691e0000000000000000000000000000000000000000000000000000000"
+        ));
+        assert!(looks_like_commit_hash(
+            "a7b89691e00000000000000000000000000000000"
+        ));
     }
 
     #[test]
@@ -83,9 +95,9 @@ mod tests {
     fn rejects_out_of_range_lengths() {
         assert!(!looks_like_commit_hash(""));
         assert!(!looks_like_commit_hash("abc"));
-        // longer than a full SHA-1
+        // longer than a full SHA-256 (64 hex chars)
         assert!(!looks_like_commit_hash(
-            "a7b89691e00000000000000000000000000000000"
+            "a7b89691e00000000000000000000000000000000000000000000000000000000"
         ));
     }
 }


phil-opp:

This PR is one of **2** opened by an automated Claude review pass over `origin/main` (`0442564`). Each finding was verified against current `main` and cross-checked against prior automated runs' open PRs to avoid duplicates.

### PRs opened this run

**Correctness**
- **#3165** — `fix(cli)`: accept full SHA-256 commit hashes in git `rev:` resolution (a full 64-char SHA-256 pin was rejected by the 40-char cap, inconsistent with the hub/locked-git paths that accept `40 | 64`).
- **#3166** — `fix(core)`: reject ROS2 QoS `max_blocking_time` / `lease_duration` values that overflow `Duration`, before they panic the bridge's `Duration::from_secs_f64` conversion (validation only guarded NaN/inf/negative, not finite-but-too-large).

_Optimization / Documentation / Simplification: no PR this run — see rejected findings and the note below._

### Findings examined and rejected (with one-line reasons)

- **cli `git.rs` — annotated tags record the tag-object OID, not the peeled commit.** Real reproducibility wart, but masked in practice (`git checkout` dereferences) and fixing/validating needs a live annotated-tag remote I can't exercise here. Deferred.
- **message `config.rs` — `InputMapping::Logs` `Display` drops `node_filter` when `min_level` is `None`.** Latent only: the lossy `{None, Some(node)}` state is never constructed in-tree (reachable only via the public fields) and the string grammar can't represent it, so there's no clean round-trip fix.
- **message `config.rs` — a node whose `id` is `dora` can't be used as an input source.** Judgment call on whether `dora` should be a reserved id; changing it is a behavior/error-message decision with unclear intent, not a clear-cut bug.
- **daemon `lib.rs` — deferred `pending_messages` increment after an `.await` can transiently `u64`-underflow.** Self-correcting cosmetic metrics glitch, low confidence on the race window; not worth the change risk.
- **coordinator `lib.rs` — `expire_stopped_nodes` drops the `node_finalized` guard after the 60s grace.** Couldn't construct a concrete trigger; low confidence.
- **mavlink2-bridge `main.rs` — telemetry forwarding gated behind the 100 ms event poll adds up to 100 ms latency for telemetry-only bridges.** Genuine hot-path latency, but a clean fix requires restructuring the event loop around a `select` over both the reader channel and the `EventStream` (which exposes no flume receiver), entangling the carefully-built 3-layer shutdown — not low-risk or verifiable without MAVLink hardware.
- **mavlink2-bridge `main.rs` — a `send_output` error while draining telemetry bypasses graceful shutdown via `?`.** Low severity (the process is exiting anyway) and entangled with the same event-loop restructuring.

Several crates were also read in depth and found already well-hardened with no new findings (`recording`, `log-utils`, `hub-client`, `arrow-convert`, `download`, `tensor-pool`, `coordinator-store`, `apis/c`, `runtime-shared-lib`).

### Validation note

Each PR's affected crate passed its own `cargo test` + `cargo clippy -D warnings` + `cargo fmt --check`, and a workspace-wide `cargo check --all --tests` (excluding the PyO3 crates) passed on the combined diff. The full `cargo test --all` binary run could not complete in this session because linking every workspace test + example binary exceeds the session's fixed writable-disk allowance (it failed only with `No space left on device` / linker bus-errors in unrelated crates, **0 actual test failures**); both changes are API-preserving, so cross-crate breakage is covered by the workspace `cargo check`. Toolchain 1.97.1 (matches CI).

---

phil-opp:

No issues found.

I reviewed the widening of `looks_like_commit_hash` from `4..=40` to `4..=64`.

- The change is consistent with the rest of the CLI, which already accepts both SHA-1 and SHA-256 commit lengths (`matches!(len, 40 | 64)` in `build/hub.rs`, `hub/publish.rs`, `hub/fetch.rs`); the old 40-char cap was the outlier that rejected a full 64-char SHA-256 `rev:`.
- The heuristic is still only a hex-and-length gate to distinguish a plausible object id from a mistyped ref before the raw-object-id fallback, so accepting abbreviations of any length in `4..=64` (rather than exactly `40 | 64`) is appropriate and does not weaken the "no matching commit" path for genuine typos.
- The added tests are non-vacuous: `accepts_full_sha256_hash` uses a 64-char id (now accepted) and a 41-char abbreviation, and `rejects_out_of_range_lengths` uses a 65-char id (still rejected). Both would fail against the old 40-char bound.

---

PR #3169, opened 2026-08-14
fix(tensor-pool): guard sender-side pool cleanup on ownership in free_tensor_pool

### LINKED ISSUE #3168: tensor-pool: free_tensor_pool destroys the caller's own pool when freeing a peer's pool (missing owner guard on PINNED_POOL cleanup)
> **Note:** This issue was created by a **scheduled automated Claude check** that randomly selects a component and reviews it for correctness bugs. Please verify before acting on it.

## Summary

In the opt-in `tensor-pool` extension, `TensorPoolTransport::free_tensor_pool` performs its **sender-side** cleanup of the process-global `PINNED_POOL` (and `TRANSIT_META`, plus the CUDA GPU/transit buffers) by extracting only the **bare counter** from the pool id, with **no check that the pool is owned by the calling node**. Because `PINNED_POOL` is keyed by a per-process monotonic counter (not namespaced by node id), a node that frees a *peer's* pool ends up freeing **its own** same-counter pool instead — freeing a live GPU buffer whose IPC handle is already exported to a downstream node.

The sibling cleanup path, `process_pending_tensor_pool_frees`, does the identical work but **correctly guards on ownership**, with a comment explicitly explaining why. `free_tensor_pool` is missing that guard.

`libraries/extensions/tensor-pool/python/src/transport.rs`

## The two paths, side by side

**`process_pending_tensor_pool_frees` — correct (has the owner guard):** (lines ~1198-1219)

```rust
// Guard against cross-process counter aliasing: buffer ids are
// pool_{node_id}_{counter}.  Extract the owner segment ... and require
// an exact equality match ...
if let Some(owner_and_counter) = buffer_id.strip_prefix("pool_")
    && let Some((owner, counter_str)) = owner_and_counter.rsplit_once('_')
    && owner == self.node_id.as_ref()          // <-- owner check
    && let Ok(c) = counter_str.parse::<u64>()
    && let Some(slot) = PINNED_POOL.lock()...remove(&c)
{
    // _unregister_host(slot.base); _free_gpu_buf(c); _free_transit(...); TRANSIT_META.remove(&c);
}
```

**`free_tensor_pool` — buggy (no owner check):** (lines ~2554-2581)

```rust
// PINNED_POOL is sender-side (per-process), so bare counter is sufficient.   // <-- wrong assumption
{
    let counter = buffer_id
        .strip_prefix("pool_")
        .and_then(|s| s.rsplit_once('_').map(|(_, c)| c))   // owner segment discarded
        .and_then(|c| c.parse::<u64>().ok());
    if let Some(c) = counter
        && let Some(slot) = PINNED_POOL.lock()...remove(&c)  // <-- no owner == self check
    {
            let bound = helpers.bind(py);
            let _ = bound.call_method1("_unregister_host", (slot.base,));
            let _ = bound.call_method1("_free_gpu_buf", (c,));
            if slot.transit_ptr != 0 {
                let _ = bound.call_method1("_free_transit", (slot.transit_ptr,));
            }
        }
        TRANSIT_META.lock()...remove(&c);
    }
    // PoolSlot dropped here -> Shmem unmapped
}
```

The comment *"PINNED_POOL is sender-side (per-process), so bare counter is sufficient"* only holds if the freeing node is always the owner. But the public API explicitly supports a **receiver** freeing a pool it received (docstring on `register_tensor_pool`, lines ~1229-1231: *"The returned pool ID can be shared across nodes ... so that a receiver can call `read_tensor_pool` and `free_tensor_pool` on it"*).

## Why the counters collide

- `PINNED_COUNTER` is a process-global `Mutex<u64>` starting at `0`, incremented before each registration (line ~1298). Every process's first pool is counter `1`, the second is `2`, and so on.
- `PINNED_POOL` is `HashMap<u64, PoolSlot>` keyed by that bare counter (insert at line ~1662), **not** namespaced by node id.
- The pool id / `buffer_id` is `format!("pool_{}_{}", self.node_id, pool_counter)` (line ~1684).

So counter-`1` collisions across nodes are the norm, not a corner case.

## Concrete failure scenario (a standard A → B → C pipeline)

1. Node B registers its own output pool → `pool_B_1`; in B's process `PINNED_POOL[1]` = B's live slot (GPU buffer + exported IPC handle).
2. Node A registers its output pool → `pool_A_1` and sends the id to B.
3. B reads `pool_A_1`, then calls `free_tensor_pool("pool_A_1")` to release it promptly.
4. In B's process: `counter = 1`, so `PINNED_POOL.remove(&1)` returns **B's own `pool_B_1` slot**, not anything belonging to A.
5. B then runs `_unregister_host(slot.base)`, `_free_gpu_buf(1)` (`cudaFree`s B's own pooled GPU buffer), `_free_transit(...)`, removes `TRANSIT_META[1]`, and drops the `PoolSlot` (munmapping B's own shmem) — all against its **own still-live** output pool.
6. Downstream node C holds an IPC handle to B's now-freed GPU buffer → reads freed-then-reallocated GPU memory (silent data corruption / use-after-free). On B's next `write_tensor_pool("pool_B_1")`, the cache miss reallocates a *new* GPU buffer, so C's imported handle is permanently stale.

Note that the **receiver-side** caches in the same function (`RECV_GPU_VA`, `RECV_CPU_SHMEM`, `GPU_BUF_SIZES`, lines ~2585-2620) are correctly keyed by the full namespaced `buffer_id`, so they target `pool_A_1` correctly. It is specifically the **sender-side `PINNED_POOL`/`TRANSIT_META` block** that mis-fires on the wrong pool.

## Suggested fix

Mirror the guard already used in `process_pending_tensor_pool_frees`: only touch `PINNED_POOL`/`TRANSIT_META`/GPU buffers when the pool's owner segment equals `self.node_id`. Extract `(owner, counter)` from `buffer_id`, require `owner == self.node_id.as_ref()`, and only then `remove(&counter)`.

## Severity

`tensor-pool` is an opt-in extension outside the 1.0 compatibility guarantees (#3152), so this only affects builds/wheels compiled with the `tensor-pool` feature. Within that feature it is a real, reachable data-corruption bug in the middle-node topology of any multi-stage pipeline that frees received pools.

## Confidence

Medium-high. The defect is verified by direct code reading; the divergence from the sibling path (whose own comment documents exactly this "cross-process counter aliasing" hazard) is strong corroboration. It requires the freeing node to also own a pool at a colliding counter — which is the common case, since counters restart at 1 per process.

phil-opp:
Fixes #3168

## Problem

In the opt-in `tensor-pool` extension, `TensorPoolTransport::free_tensor_pool` (`libraries/extensions/tensor-pool/python/src/transport.rs`) performed its sender-side cleanup of the process-global `PINNED_POOL` / `TRANSIT_META` maps (and the CUDA GPU/transit buffers) by extracting only the **bare counter** from the pool id, with **no check that the calling node owns the pool**.

`PINNED_POOL` is keyed by a per-process monotonic counter that restarts at `1` in every process — it is *not* namespaced by node id. The public API explicitly supports a **receiver** freeing a pool it received (`register_tensor_pool` docstring). So when node B frees a peer's pool `pool_A_1`, `PINNED_POOL.remove(&1)` returns **B's own `pool_B_1` slot** and frees a live GPU buffer whose IPC handle is already exported to a downstream node C — a use-after-free / silent GPU-memory corruption in any A→B→C pipeline that frees received pools.

The sibling path `process_pending_tensor_pool_frees` already guarded this correctly (with a comment documenting the exact "cross-process counter aliasing" hazard); `free_tensor_pool` was missing the guard.

## Fix

- Extract the owner check into a shared `owned_pool_counter(buffer_id, node_id)` helper that yields the bare counter **only** when the pool's owner segment exactly equals the calling node id. The match is exact (not a `starts_with` prefix check) so node ids that are prefixes of each other (e.g. `cam` / `cam_left`) do not alias.
- Use the helper at **both** the `free_tensor_pool` and `process_pending_tensor_pool_frees` sites, so the two paths can no longer diverge (which is how this bug arose).
- The receiver-side caches (`RECV_GPU_VA`, `RECV_CPU_SHMEM`, `GPU_BUF_SIZES`) are already keyed by the full namespaced `buffer_id` and are unchanged.

## Tests

Added a `owner_guard_tests` unit module covering the owned, peer-owned, prefix-aliasing, and malformed-id cases.

## Validation

- `cargo test -p dora-tensor-pool-python owner_guard` — 4 passed
- `cargo clippy -p dora-tensor-pool-python -- -D warnings` — clean

## Scope

`tensor-pool` is opt-in (`--features tensor-pool`), off by default, and outside the 1.0 compatibility guarantees, so default builds/wheels are unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KUcz6HGAvRRnibU5R8Vpzo)_

### DIFF
--- a/libraries/extensions/tensor-pool/python/src/transport.rs
+++ b/libraries/extensions/tensor-pool/python/src/transport.rs
@@ -1020,6 +1020,59 @@ fn warn_missing_tensor_pool(node_id: &NodeId, action: &str, buffer_id: &str) {
     );
 }
 
+/// Extracts the per-process pool counter from a namespaced
+/// `pool_{node_id}_{counter}` buffer id, but only when `node_id` owns the pool.
+///
+/// The sender-side `PINNED_POOL`/`TRANSIT_META` maps are keyed by the bare
+/// per-process counter, *not* namespaced by node id, and every process's
+/// counters restart at 1. So a node that frees a *peer's* pool (a receiver
+/// releasing a pool it read) would otherwise reclaim its *own* same-counter
+/// slot — freeing a live GPU buffer whose IPC handle is still exported to a
+/// downstream node (dora-rs/dora#3168). Guarding on the owner segment keeps the
+/// sender-side cleanup pinned to pools this node actually registered.
+///
+/// The owner match is exact (not a `starts_with` prefix check) so node ids that
+/// are prefixes of each other (e.g. `cam` / `cam_left`) do not alias.
+fn owned_pool_counter(buffer_id: &str, node_id: &str) -> Option<u64> {
+    let (owner, counter) = buffer_id.strip_prefix("pool_")?.rsplit_once('_')?;
+    if owner != node_id {
+        return None;
+    }
+    counter.parse::<u64>().ok()
+}
+
+#[cfg(test)]
+mod owner_guard_tests {
+    use super::owned_pool_counter;
+
+    #[test]
+    fn extracts_counter_for_owned_pool() {
+        assert_eq!(owned_pool_counter("pool_camera_7", "camera"), Some(7));
+    }
+
+    /// A receiver freeing a peer's pool must not resolve to a bare counter —
+    /// otherwise it would reclaim its own same-counter `PINNED_POOL` slot
+    /// (dora-rs/dora#3168).
+    #[test]
+    fn rejects_peer_owned_pool() {
+        assert_eq!(owned_pool_counter("pool_A_1", "B"), None);
+    }
+
+    /// Owner ids that are prefixes of each other must not alias.
+    #[test]
+    fn does_not_alias_prefix_node_ids() {
+        assert_eq!(owned_pool_counter("pool_cam_left_3", "cam"), None);
+        assert_eq!(owned_pool_counter("pool_cam_left_3", "cam_left"), Some(3));
+    }
+
+    #[test]
+    fn rejects_malformed_ids() {
+        assert_eq!(owned_pool_counter("notapool_x_1", "x"), None);
+        assert_eq!(owned_pool_counter("pool_x_notanumber", "x"), None);
+        assert_eq!(owned_pool_counter("pool_x", "x"), None);
+    }
+}
+
 // ==================== seqlock ====================
 /// Begins a tensor-pool seqlock write at `gen_ptr` (header offset 96)
 /// **if the generation is even**.  Returns the **even** pre-write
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The `owned_pool_counter` helper is correct, including the underscore-in-node-id case: `rsplit_once('_')` peels only the trailing counter, so `pool_cam_left_3` resolves to owner `cam_left`, and the exact-equality check (not `starts_with`) stops `cam` from aliasing `cam_left`. Routing both `free_tensor_pool` and `process_pending_tensor_pool_frees` through the one helper closes the divergence that produced the use-after-free, and the receiver-side caches keyed by the full `buffer_id` are correctly left alone.

The unit tests exercise the guard directly (owned / peer-owned / prefix-alias in both directions / malformed ids) and would fail if it were reverted to a bare-counter or `starts_with` implementation.

---

Issue #3161, opened 2026-08-13
ci: `pip-release` wheel jobs intermittently fail with `self-signed certificate` in `maturin-action`'s versions-manifest lookup


phil-opp:
ci: `pip-release` wheel jobs intermittently fail with `self-signed certificate` while `maturin-action` downloads maturin, ejecting PRs from the merge queue

## Summary

Wheel-build jobs in `pip-release.yml` intermittently fail with a Node.js TLS
error before any build work happens:

```
Error: self-signed certificate; if the root CA is installed locally,
try running Node.js with --use-system-ca
##[error]self-signed certificate; ...
```

It occurs while `maturin-action` downloads the maturin release asset. No compile
error, no test failure — the job dies during setup.

Because `pip-release all green` is a **required merge-queue status**
(`.trunk/trunk.yaml`), any batch containing an affected job is ejected and has to
be re-submitted. That is a real throughput cost while working through the v1.0
blocker set.

## Occurrences (2026-08-11)

| Time (UTC) | Batch | Job(s) |
|---|---|---|
| — | `trunk-merge/pr-2859` | `linux (ubuntu-22.04, x86, binaries/cli, dora-rs-cli)` |
| — | `trunk-merge/pr-3039` | `macos (macos-15, aarch64, binaries/cli, dora-rs-cli)` |
| 18:19 | `trunk-merge/pr-3123` | `musllinux (ubuntu-22.04, aarch64, …)` **and** `windows (windows-latest, x64, …)` |
| 17:43 | `main` | `musllinux (ubuntu-22.04, x86, binaries/cli, dora-rs-cli)` ([run 31518988305](https://github.com/dora-rs/dora/actions/runs/31518988305/job/93871294375)) |
| 21:53 | `trunk-merge/pr-3123` | `musleabi (armv7-unknown-linux-musleabihf, …, binaries/cli, dora-rs-cli)` ([run 31539935556](https://github.com/dora-rs/dora/actions/runs/31539935556/job/93939804237)) |

The `pr-3123` pair failed at the **same second** (`18:19:50`) on two different
runner OSes, which points at a network/registry event rather than anything
runner-local or PR-specific.

Three of the affected PRs are unrelated to packaging (#2859 is a `dora node info`
rendering fix; #3039 bounds a Python-side set), so the failures are clearly not
caused by the code under test.

## Not the buildjet thing

The `cache-provider buildjet is not valid` annotation that appeared alongside the
[...]

PR #3170, opened 2026-08-14
fix(daemon): avoid coordinator WS deadlock under outbound backpressure

### LINKED ISSUE #3164: fix(daemon): avoid deadlock between WebSocket control loop and main event loop under outbound backpressure
## Problem Description

The daemon can experience a permanent deadlock when the outbound WebSocket channel (`send_tx`) becomes saturated under heavy log or event traffic.

### Root Cause Analysis

In `binaries/daemon/src/coordinator.rs`, the WebSocket handling task uses a `tokio::select!` loop to handle both incoming coordinator messages (`ws_rx`) and outgoing messages (`send_rx`):

https://github.com/dora-rs/dora/blob/main/binaries/daemon/src/coordinator.rs#L210-L292

When a coordinator command (e.g., `DaemonCoordinatorEvent::AllNodesReady`) arrives via `ws_rx`:
1. The WebSocket task forwards the event to the daemon's main event loop over `tx`.
2. Inside the `ws_rx.next()` branch of `tokio::select!`, the task synchronously awaits `reply_rx.await`:
   ```rust
   let Ok(reply) = reply_rx.await else { ... };

phil-opp:
Fixes #3164

## Problem

The daemon's coordinator WebSocket task ran a single `tokio::select!` handling both inbound commands (`ws_rx`) and the outbound queue (`send_rx`). The `ws_rx` branch awaited the daemon's main event loop **inline** — but that branch is the only drain for `send_rx`.

That closes a cycle: the main loop needs `send_tx` capacity (64) to finish handling a command, `send_tx` only frees when `send_rx` drains, and `send_rx` only drains when the `ws_rx` branch yields — which it won't while it is waiting on the main loop. Under heavy log/event traffic the daemon wedges permanently.

## Fix

The connection is serviced by two cooperating tasks:

- a **writer** (`run_coordinator_ws_writer`), sole owner of `ws_tx`, draining both `send_rx` (external fire-and-forget events) and `internal_rx` (reader-produced command responses and pongs);
- a **reader** (`run_coordinator_ws_reader`), owning `ws_rx`, which forwards each command to the main loop, awaits its reply, and hands responses/pongs to the writer.

Because the outbound drain runs in its own task, nothing the reader does — awaiting a reply, or blocking on a full `tx` — can stall it, so the internal cycle cannot form; only genuine peer backpressure can slow transmission. The reader keeps one command in flight at a time, unchanged from before.

`DestroyResult`'s completion notify is queued to the writer as a trailing `OutboundFrame::DestroyNotify` on the same FIFO channel, so it still fires only after the response has been flushed.

Both halves are generic over the WS sink/stream, so they can be driven with in-memory channels.

## Tests

- `ws_writer_drains_outbound_while_command_reply_is_in_flight` — a mock main loop emits a 500-message burst against the production 64-slot `send_tx` before replying; it completes only if the writer drains concurrently. The pre-fix single loop deadlocks and hits the 10s timeout.
- `ws_reader_routes_replies_to_pending_daemon_requests` — drives `resolve_machine` against the reader, pinning the `ReplyRouteRaw` / `COORDINATOR_PENDING` dispatch ahead of the typed command parse. Deleting that block makes `resolve_machine` fall through and time out returning `None`.

## Validation

`cargo test -p dora-daemon` (244 passed), `cargo clippy -p dora-daemon -- -D warnings`, `cargo fmt --all -- --check` — all clean.

### DIFF
--- a/binaries/daemon/src/coordinator.rs
+++ b/binaries/daemon/src/coordinator.rs
@@ -755,4 +755,77 @@ mod tests {
         .await
         .expect("WS router deadlocked under outbound backpressure (#3164)");
     }
+
+    /// Regression test for the reply routing `resolve_machine` depends on
+    /// (dora-rs/dora#3079): the coordinator answers a daemon->coordinator
+    /// request inside the same `daemon_event` envelope as an inbound command,
+    /// but with a params type the typed `CoordinatorCommandRaw` parse rejects.
+    /// The reader must therefore route it to the pending caller by id, *before*
+    /// that parse.
+    ///
+    /// Pinned because the routing block sits mid-loop in the reader and was
+    /// dropped once already while restructuring this connection: without it
+    /// every cross-machine `resolve_machine` falls through to the command
+    /// parse, logs a parse failure, and times out returning `None`.
+    #[tokio::test]
+    async fn ws_reader_routes_replies_to_pending_daemon_requests() {
+        let clock = Arc::new(HLC::default());
+
+        // Inbound frames are fed by the test, so the reply can echo the request
+        // id `resolve_machine` picks at call time.
+        let (ws_in_tx, ws_in_rx) = mpsc::channel::<Result<Message, std::io::Error>>(1);
+        let (tx, mut rx) = mpsc::channel::<Timestamped<CoordinatorEvent>>(1);
+        let (internal_tx, mut internal_rx) = mpsc::channel::<OutboundFrame>(1);
+        let reader = tokio::spawn(run_coordinator_ws_reader(
+            ReceiverStream::new(ws_in_rx),
+            tx,
+            internal_tx,
+            clock.clone(),
+        ));
+
+        let (sender, mut outbound_rx) = CoordinatorSender::for_test();
+        let resolve = tokio::spawn({
+            let clock = clock.clone();
+            async move { resolve_machine(&sender, &clock, "machine-a").await }
+        });
+
+        // Answer the request the daemon just sent, echoing its id.
+        let request = outbound_rx
+            .recv()
+            .await
+            .expect("resolve_machine must send a request");
+        let request_id = serde_json::from_str::<serde_json::Value>(&request).unwrap()["id"]
+            .as_str()
+            .expect("the envelope carries the request id")
+            .to_owned();
+        let reply = format!(
+            r#"{{"id":"{request_id}","method":"daemon_event","params":{{"inner":{{"ResolveMachineResult":{{"found":true,"address":"127.0.0.1:6021"}}}}}}}}"#
+        );
+        ws_in_tx
+            .send(Ok(Message::Text(reply.into())))
+            .await
+            .unwrap();
+
+        let resolved = tokio::time::timeout(Duration::from_secs(5), resolve)
+            .await
+            .expect(
+                "the reply must reach the pending caller, not fall through to the command parse",
+            )
+            .unwrap();
+        assert_eq!(resolved, Some("127.0.0.1:6021".parse().unwrap()));
+
+        // A routed reply is consumed by the routing block: it must neither be
+        // dispatched to the main loop nor answered on the wire.
+        assert!(
+            rx.try_recv().is_err(),
+            "a routed reply must not reach the daemon main loop"
+        );
+        assert!(
+            internal_rx.try_recv().is_err(),
+            "a routed reply must not produce an outbound response frame"
+        );
+
+        drop(ws_in_tx);
+        reader.await.unwrap();
+    }
 }
--- a/binaries/daemon/src/coordinator.rs
+++ b/binaries/daemon/src/coordinator.rs
@@ -7,7 +7,7 @@ use dora_message::{
     ws_protocol::WsResponse,
 };
 use eyre::eyre;
-use futures::{SinkExt, StreamExt};
+use futures::{Sink, SinkExt, StreamExt};
 use std::{net::SocketAddr, sync::Arc, time::Duration};
 use tokio::sync::{mpsc, oneshot};
 use tokio_stream::{Stream, wrappers::ReceiverStream};
@@ -178,8 +178,9 @@ pub async fn register(
     let (mut ws_tx, mut ws_rx) = ws_stream.split();
 
     // Channel for outgoing messages (daemon events + command replies).
-    // The coordinator sender writes to this, and the spawned task reads and forwards to WS.
-    let (send_tx, mut send_rx) = mpsc::channel::<String>(64);
+    // The coordinator sender writes to this, and the writer task below reads
+    // and forwards to WS.
+    let (send_tx, send_rx) = mpsc::channel::<String>(64);
 
     // Send Register request.
     // Serialize params via to_string (not to_value) to preserve u128 fidelity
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I traced the deadlock fix end to end:

- `tx` is capacity 1 (`mpsc::channel(1)` in `register`), so command *delivery* stays ordered — a new `ws_rx` command isn't read until the main loop accepts the previous one, and at most two replies are ever pending. Only reply *delivery* interleaves, which is safe since the coordinator correlates by `request_id`.
- The one remaining inline await in the `ws_rx` branch (`tx.send(next_cmd)`) only blocks until the main loop *receives* the buffered command, which happens before it produces its outbound burst — so it can't re-wedge the loop the way the old inline `reply_rx.await` did.
- `DestroyResult` still terminates the loop, and the empty-`FuturesUnordered` arm stays correctly disabled (`Ready(None)` doesn't match `Some(..)`).

The regression test is non-vacuous: the mock loop bursts 500 messages (far above the 64-slot capacity) before replying, so it only completes if `send_rx` keeps draining concurrently; reverting to the inline await makes it hit the timeout.

---

phil-opp:
🤖 **Automated review by Claude** — fully automated; not yet vetted by a human maintainer. This corrects an earlier automated comment on this PR that said "no issues found"; on a closer pass I believe a narrower instance of the same deadlock class survives.

Moving the reply await into `FuturesUnordered` does fix the single-command case, and the regression test demonstrates that. But the `ws_rx` arm still awaits `tx.send(...)` **inline**, and `tx` is capacity 1 — so I don't think the ordering guarantee in the new doc comment actually holds:

> Command *delivery* stays ordered: `tx` has capacity 1, so a new `ws_rx` command is not read until the main loop has accepted the previous one.

A capacity-1 channel still buffers **one** item: the router can read a command, have `tx.send(...)` return immediately into the free slot (before the main loop accepts it), and then read a *second* command whose `tx.send(...).await` blocks. Because that await parks the whole `select!`, the `send_rx` drain arm stops running while it's pending — structurally the same hazard as the old inline `reply_rx.await`, just one command deeper.

Concrete interleaving (`tx` cap 1, `send_tx` cap 64):

1. Main loop receives cmdA and, while handling it, blocks on `CoordinatorSender::send_event().await` because `send_tx` is full under outbound backpressure.
2. Router reads cmdB; `tx.send(cmdB)` buffers into the now-free slot (main loop already took cmdA) and returns; its reply future goes into `pending_replies`.
3. Router reads cmdC; `tx.send(cmdC).await` blocks — `tx` is full because cmdB is still buffered (the main loop is parked in step 1 and hasn't `recv`'d it).
4. Parked at that await, the router no longer drains `send_rx`. So the main loop's `send_event()` never completes → it never receives cmdB → `tx.send(cmdC)` never returns. Permanent deadlock.

Preconditions: outbound backpressure (`send_tx` full via a handler that uses the blocking `send_event`, which is the #3164 premise) **plus** at least two more coordinator frames already queued in `ws_rx` behind the one being processed. Since `select!` polls ready arms in random order, once cmdB and cmdC are both queued there's a real chance the router advances `ws_rx` twice before `send_rx`, and reaching the blocking `tx.send(cmdC)` is then terminal.

The regression test only drives a single in-flight command (`stream::iter(vec[one cmd]).chain(stream::pending())`), so it wouldn't catch this. A test that queues ≥2 inbound commands while the mock main loop backpressures on `send_tx` would exercise it.

Possible direction: keep the inbound arm from ever awaiting while parked — e.g. gate reading the next `ws_rx` frame on a `tx.reserve()` permit acquired *before* `ws_rx.next()`, or hand the accepted command off through the same off-to-the-side mechanism as the replies — so that no single `.await` in the loop can stop `send_rx` from draining. I may be over-weighting the "≥2 queued commands" precondition, so flagging for a maintainer to confirm the interleaving rather than asserting it's a hard blocker.

---

phil-opp:

The deadlock is narrowed rather than removed, and the branch is missing a block that main has since added.

**1. The deadlock is still reachable.** `binaries/daemon/src/coordinator.rs:283` still does an inline, untimed `tx.send(...).await` inside the same `select!` that owns `send_rx`, and `tx` has capacity 1 (`coordinator.rs:234` on main).

Sequence: the main loop is inside a coordinator-command handler emitting a long outbound burst and blocks on the full 64-slot `send_tx`. The router reads inbound frame #1 → `tx.send` buffers it. It reads frame #2 → `tx.send(...).await` blocks, because the main loop has not returned to `rx.recv()`. `send_rx` stops draining → the main loop stays blocked → permanent deadlock, as before.

This needs no client action to trigger: the coordinator pushes a fire-and-forget heartbeat every 3 s (`binaries/coordinator/src/lib.rs:317`), so any handler that stays in its outbound burst for ~6 s reproduces it.

The new test cannot catch this — its `ws_in` yields exactly one command and then `stream::pending()`, so the second `tx.send` never happens. A structural fix (moving the outbound `send_rx` → `ws_tx` drain into its own task, and routing command responses through `send_tx`) would make the inline `tx.send` harmless.

**2. Stale base drops #3079's reply routing.** The branch is based on `044256482`, which predates the cross-machine work. `run_coordinator_ws_router` reproduces the old loop body and omits the `COORDINATOR_PENDING` / `ReplyRouteRaw` block main added at `binaries/daemon/src/coordinator.rs:257-277`, which routes `ResolveMachine` replies *before* the `CoordinatorCommandRaw` parse. `git merge-tree` conflicts on this file so it will not land silently — but the rebase has to reinstate that block inside the new router, or every cross-machine `resolve_machine` falls through to the typed parse, logs a parse failure, and times out after 5 s returning `None`.

For the record, the parts that are fine: the reply reordering this introduces is safe (the coordinator correlates by `request_id` and has a `TCP_READ_TIMEOUT`), and the added test is genuinely RED against the old code.

phil-opp:
Both findings were correct — thanks. Pushed a rework (`a293c81`, rebased onto current `main`) that addresses them:

**1. Deadlock was still reachable → reworked to a two-task split.**

You're right that `FuturesUnordered` only moved the block: it read the next frame while a reply was pending, so the *second* `tx.send(...).await` (capacity-1 `tx`) could still stall the loop that owns `send_rx`. The real problem is that a single loop both awaited replies and drained `send_rx`.

The connection is now serviced by two cooperating tasks:
- a **writer** task that is the sole owner of `ws_tx` and continuously drains both `send_rx` (external events) and an internal channel of reader-produced frames (command responses + pongs);
- a **reader** task that owns `ws_rx`, forwards each command to the main loop, awaits its reply inline, and hands responses/pongs to the writer.

Because the outbound drain now runs independently of command/reply processing, nothing the reader does — awaiting a reply *or* blocking on a full `tx` — can stall it, so the internal cycle can't form. I reverted `FuturesUnordered`: the reader is back to one command in flight at a time (unchanged from the original), which also removes the read-ahead that exposed the second `tx.send`. Only one task can own the `SplitSink`, so responses/pongs route through the writer via an internal `OutboundFrame` channel; `DestroyResult`'s notify is sent as a trailing `OutboundFrame::DestroyNotify` so it still fires only after the response is flushed.

The regression test was updated accordingly (`ws_writer_drains_outbound_while_command_reply_is_in_flight`): a handler emits a 500-message burst against a 64-slot `send_tx` before replying, which only completes if the writer keeps draining concurrently.

**2. Stale base / missing #3079 reply routing → rebased and preserved.**

Rebased off `0442564` onto current `main`. The new `run_coordinator_ws_reader` keeps the `ReplyRouteRaw` / `COORDINATOR_PENDING` block ahead of the `CoordinatorCommandRaw` parse, so `resolve_machine` replies are still routed by id before the typed command parse.

Validation: `cargo test -p dora-daemon --lib coordinator::` (5 passed), `cargo clippy -p dora-daemon -- -D warnings` clean, `cargo fmt --all -- --check` clean.

---

phil-opp:
The rework (`a293c81`) addresses both points from my previous review.

**1. Deadlock — fixed structurally.** Splitting the connection into `run_coordinator_ws_writer` (sole owner/drainer of `ws_tx` + `send_rx`) and `run_coordinator_ws_reader` removes the earlier hazard: because the writer drains `send_rx` in an independent task, neither the inline `tx.send(...).await` nor the inline `reply_rx.await` in the reader can stall the outbound drain, so the capacity-1-`tx` "second queued command" interleaving I described can no longer wedge the main loop. `FuturesUnordered` is gone and the reader is back to one command in flight. The new test (`ws_writer_drains_outbound_while_command_reply_is_in_flight`) bursts 500 messages against a 64-slot `send_tx` before replying, so it genuinely exercises the concurrent drain.

**2. Missing reply routing — restored.** After the rebase, `run_coordinator_ws_reader` keeps the `ReplyRouteRaw` / `COORDINATOR_PENDING` block ahead of the `CoordinatorCommandRaw` parse, so `ResolveMachine` replies are still routed by id.

I also checked the reader/writer split for new issues and found none: `DestroyNotify` is enqueued after the `Ws` response on the same FIFO `internal_rx`, so the notify still fires only after the response is flushed; the `None`/`is_err()` arms give clean shutdown (dropped receivers make senders return `Err`, so the main loop can't hang on a dead writer). No issues found in this diff.

---

---

phil-opp:
The merge-queue failure is **not from this PR** — it was a batched run (batch branch `trunk-merge/pr-3100/…`) and the failing job was a `dora-node-api` **doctest**, unrelated to this daemon-only change:

```
apis/rust/node/src/node/arrow_utils/ipc_encode.rs - node::arrow_utils::ipc_encode::PreparedIpc (line 360)
error[E0433]: cannot find `arrow` in `dora_node_api`
  363 | use dora_node_api::arrow::array::{Array, UInt64Array};
```

This PR only touches `binaries/daemon/src/coordinator.rs`; it doesn't touch `dora-node-api` or its `pub use arrow;` re-export (`apis/rust/node/src/lib.rs:89`). I ran the full doctest suite on this branch (current `main` + this change) and it's green:

```
cargo test -p dora-node-api --doc
test result: ok. 10 passed; 0 failed; 4 ignored; 0 measured; 0 filtered out
```

So the failure comes from another PR co-batched in the queue, not from #3170. Trunk's bisection (`trunk-merge/pr-3170/…-bisection`) should clear this PR and re-queue it. No change needed here; I'll keep watching.

---

phil-opp:
Confirmed the blocker is a **pre-existing merged-doctest failure on `main`**, not this PR. The failing job is a `dora-node-api` doctest added by #3191 (`88c0883`):

```
apis/rust/node/src/node/arrow_utils/ipc_encode.rs - PreparedIpc (line 359/360)
error[E0433]: cannot find `arrow` in `dora_node_api`
  use dora_node_api::arrow::array::{Array, UInt64Array};
```

Evidence it is neither #3170's fault nor a deterministic break:

- **This PR is daemon-only** — it doesn't touch `dora-node-api` or its `pub use arrow;` re-export (`apis/rust/node/src/lib.rs:89`, ungated; `arrow` is a required dep).
- **It passes locally** with the exact CI command and the pinned toolchain (`rustc 1.97.1`): `cargo test -p dora-node-api --doc` and `cargo test --all --doc --exclude <python> --exclude dora-examples` both green (`PreparedIpc … ok`, 11 passed / 0 failed), across isolated and full-workspace builds.
- **#3169 merged through this same queue minutes earlier** with #3191's doctest already on `main` (#3191 is older than #3169 in the history), so its batch passed the very same doctest.

So it reproduces only in certain merge-queue batch compositions — the signature of a flaky Rust-2024 **merged-doctest** bug resolving the `pub use`'d `arrow` crate (`dora_node_api::arrow`), not a real regression in the code under test. I can't reproduce it locally, so I'm deliberately **not** pushing a speculative doctest change here (it wouldn't belong in a daemon PR anyway, and I can't validate it).

Suggested paths, in order:
1. **Re-queue #3170** — it may land on a batch composition that doesn't trip the flake.
2. If it recurs, harden the #3191 doctest against the merged-doctest resolver in a separate PR (e.g. force it to compile standalone, or avoid the `dora_node_api::arrow` re-export path). Happy to take that on if you'd like — it would unblock every PR in the queue, not just this one.

#3170 itself is ready on its merits (reworked per the earlier review; all its own PR-level checks are green).

---

phil-opp:
**Correction to my previous comment — that diagnosis was wrong.** The doctest failure is not a flaky merged-doctest resolver bug. It is deterministic, and it comes from **#3213**, co-batched in the queue.

**Cause.** #3191 merged a `PreparedIpc` doctest to `main` at 12:28 (`88c0883`) that names `dora_node_api::arrow`. #3213 (`arrow-version-decoupling`) removes `pub use arrow;`, replacing it with a feature-gated `pub use arrow as arrow_v59;` where `arrow-v59` is **not** a default feature. #3213 was branched before #3191 landed, so it never saw that doctest — and the two merge with no textual conflict, producing a tree whose doctest names a symbol that no longer exists.

**Evidence.** The batch that failed here is `trunk-merge/pr-3100/00f7c5e5…` = `cfabd48d0`, whose members are #3100, #3170 (`a293c81`), #3213 (`970a949`), #3217 (`0c4307e`), and one more. `git show cfabd48d0:apis/rust/node/src/lib.rs` contains no `pub use arrow;` — only the gated alias. Rebasing #3213 onto current `main` locally reproduces the exact CI error:

```
error[E0433]: cannot find `arrow` in `dora_node_api`
```

and the doctest passes again once fixed. That also accounts for everything I previously misread as flakiness: it passes on this branch (no #3213 here), #3169 merged minutes earlier (its batch had no #3213), and it only appears in "certain batch compositions" — precisely those containing #3213. Trunk's own bisection agreed: #3100's and #3217's bisection runs went green, while #3213 picked up the bisection label and then its own queue failure.

**The fix belongs in #3213, not here**, and it is larger than the import: `PreparedIpc::new` now takes `&DoraArray` rather than `&ArrayData`, `decode_arrow_ipc` returns `DoraArray`, and `DoraArray` derives no `PartialEq`, so the closing `assert_eq!` could not compile either. Rewritten there through `IntoArrow` / `into_vec`, which keeps the example on the Arrow-free path that PR exists to establish.

Two things fixed on this PR while I was here:

- **Restored `ws_reader_routes_replies_to_pending_daemon_requests`.** The rework dropped it, leaving nothing guarding the `ReplyRouteRaw` / `COORDINATOR_PENDING` block that `resolve_machine` (#3079) depends on — the same block this PR had already lost once. The test drives `resolve_machine` against the reader and fails, exactly as the bug presents, when the block is removed.
- **Refreshed the description**, which still described the reverted `FuturesUnordered` design and three tests that no longer exist.

One residual issue this PR does not address, and shouldn't: under genuine peer backpressure the writer parks on `ws_tx.send()`, `send_rx` fills, and the daemon's main loop then blocks on `send_tx.send().await`. A wedged coordinator socket can still stall the main loop. That is unchanged from before this PR — worth a separate issue.

---

phil-opp:
This queue failure is the `pip-release` wheel-build flake tracked in #3161, not the daemon change — 20 of the 22 wheel jobs are green and the two remaining `linux` CLI jobs (armv7/aarch64) are already retrying. A daemon-only Rust change can't affect the Python wheel builds. Nothing to fix here; it should clear on the retry/next queue attempt.

---

PR #3171, opened 2026-08-14
fix(daemon): forget cascading-error cause when a node is removed

phil-opp:
## Issue

`ReplaceNode` scrubs the per-node cascading-error record via `cascading_error_causes.forget()` (dora-rs/dora#2927) so that a replacement incarnation of a node id is not blamed for its predecessor's cohort-era failure. The `RemoveNode` cleanup path did **not**: it called `forget_node_bookkeeping()` — which only touches `input_deadlines`, `broken_inputs`, and `node_stderr_most_recent` — but never `cascading_error_causes.forget()`.

A `caused_by` entry is otherwise never removed. So after a `dora node remove X` followed by `dora node add X` (both first-class dynamic-reconfiguration flows), a stale `caused_by[X] = Y` survives the removal.

## Impact

When the **new** incarnation of `X` later exits with a genuine error, the classifier (`cascading_error_causes.error_caused_by(&node_id)` in `binaries/daemon/src/lib.rs`) finds the stale entry and tags the failure as `NodeErrorCause::Cascading { caused_by_node: Y }`. This suppresses the real stderr capture (`NodeErrorCause::Other`) and blames a node that has nothing to do with the new failure — exactly the stale-attribution class that `CascadingErrorCauses::forget` was introduced to prevent.

## Fix

Move the `cascading_error_causes.forget()` call **into** `forget_node_bookkeeping()`, which both `RemoveNode` and `ReplaceNode` already call, and drop the now-redundant explicit call in `ReplaceNode`. This closes the `RemoveNode` gap without changing `ReplaceNode` behavior.

- `binaries/daemon/src/running_dataflow.rs` — `forget_node_bookkeeping` now also calls `self.cascading_error_causes.forget(node_id)`; doc comment updated.
- `binaries/daemon/src/lib.rs` — removed the redundant `cascading_error_causes.forget(&node_id)` from the `ReplaceNode` handler (behavior preserved).

## Validation

- Extended the existing `forget_node_bookkeeping_purges_only_that_node` regression test to seed a cascading cause for two nodes and assert the removed node's cause is dropped while the other node's is untouched.
- `cargo test -p dora-daemon` — 214 passed.

---

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -3162,7 +3162,8 @@ impl Daemon {
                     dataflow.all_inputs_closed_at.remove(&node_id);
                     dataflow.connected_nodes.remove(&node_id);
                     dataflow.finish_escalated.remove(&node_id);
-                    dataflow.cascading_error_causes.forget(&node_id);
+                    // `forget_node_bookkeeping` also drops the recorded
+                    // cascading-error cause for this id (see its doc comment).
                     dataflow.forget_node_bookkeeping(&node_id);
                     dataflow
                         .node_stderr_most_recent
@@ -8246,6 +8247,7 @@ mod fault_tolerance_tests {
         let input_x: DataId = "input_x".to_string().into();
         let timeout = Duration::from_secs(1);
 
+        let upstream: NodeId = "upstream".to_string().into();
         for node in [&node_a, &node_b] {
             df.input_deadlines.insert(
                 (node.clone(), input_x.clone()),
@@ -8258,6 +8260,9 @@ mod fault_tolerance_tests {
                 .insert((node.clone(), input_x.clone()), timeout);
             df.node_stderr_most_recent
                 .insert(node.clone(), Arc::new(ArrayQueue::new(4)));
+            // Both nodes were recorded as cascading victims of `upstream`.
+            df.cascading_error_causes
+                .report_cascading_error(upstream.clone(), node.clone());
         }
 
         df.forget_node_bookkeeping(&node_a);
@@ -8272,6 +8277,10 @@ mod fault_tolerance_tests {
                 .contains_key(&(node_a.clone(), input_x.clone()))
         );
         assert!(!df.node_stderr_most_recent.contains_key(&node_a));
+        // … including the stale cascading-error cause, so a re-added/replaced
+        // `node_a` incarnation is classified by its own failure, not a stale
+        // upstream cause (dora-rs/dora#2927).
+        assert_eq!(df.cascading_error_causes.error_caused_by(&node_a), None);
 
         // … while node_b's are untouched.
         assert!(
@@ -8283,6 +8292,10 @@ mod fault_tolerance_tests {
                 .contains_key(&(node_b.clone(), input_x.clone()))
         );
         assert!(df.node_stderr_most_recent.contains_key(&node_b));
+        assert_eq!(
+            df.cascading_error_causes.error_caused_by(&node_b),
+            Some(&upstream)
+        );
     }
 
     // -- Test 3: close_input defers AllInputsClosed when broken_inputs exist --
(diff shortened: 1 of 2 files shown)


phil-opp:
## Autonomous code-review run — summary

This is the lead PR of a machine-driven review pass over `origin/main` (`0442564`). Each finding was verified against current `main`, checked against prior open autonomous-run PRs to avoid duplicates, and gated with fmt + clippy + affected-crate tests per branch, plus `fmt --all` / `clippy --all` and a `/review` + `/simplify` pass on the combined diff.

### Opened PRs

**Correctness**
- #3171 — `fix(daemon)`: forget the cascading-error cause when a node is removed. `RemoveNode` never cleared `cascading_error_causes`, so `remove` + re-`add` of a node id misattributes the new incarnation's genuine failure as `Cascading` and suppresses its stderr. Moved the `forget` into the shared `forget_node_bookkeeping` (this PR).
- #3173 — `fix(cli)`: don't panic the hot-reload watcher thread when the control loop's channel is already closed during shutdown (`.unwrap()` → `let _ =`, matching the sibling senders).
- #3174 — `fix(record-node)`: don't panic the recorder on a pre-epoch system clock (`duration_since(UNIX_EPOCH).unwrap()`), relevant to RTC-less embedded boots; saturates to 0.

**Documentation**
- #3172 — `docs(cli)`: correct the `dora stop --grace-duration` kill-ladder timing — the help said SIGKILL fires at `DURATION * 0.5` (before the `DURATION` SIGTERM); the real ladder is SIGTERM at `DURATION`, SIGKILL at `DURATION * 1.5`.

### Rejected / deferred findings (one line each)

- **replay-node: skip a corrupt container record instead of aborting** — opened as #3175, then **withdrawn**: the original `next_entry()?` already replayed the good prefix and failed *loudly*; the change only turned that into a silent exit-0 (and a loop-mode spin), a regression. The recording layer intentionally signals mid-file corruption as `Err` (vs a benign torn-tail `Ok(None)`).
- **core/message byte-size + log-level parser duplication** — the `dora-core` `u64` `parse_byte_size` (for `max_log_size`) is a deliberate, mutation-test-hardened parser distinct from `dora-message`'s `usize` `ByteSize`; consolidating fights an intentional choice and changes user-facing error text.
- **daemon `ExtensionTable::store` O(n) full-table scan on insert** — real, but a side-counter map must be kept in sync across `drop_key`/`reclaim_owner`/`reclaim_dataflow`; correctness risk outweighs the win without evidence it's a measured hot path.
- **cli `dora logs`: legacy `.txt` files bypass `--since/--grep/--level/--tail` and sort order** — genuine, but the fix is a non-trivial refactor of `read_log_file` to route raw text through the filter/sort/tail pipeline; deferred as medium-impact (legacy fallback path).
- **replay-node: tiny-positive `DORA_REPLAY_SPEED` → effectively-infinite sleep** — low severity (user misconfiguration); a sleep ceiling would be arbitrary.
- **cli `dora record` dead/misleading output-path comment** — cosmetic-only; not worth a standalone PR.

### Validation note

Per-branch and combined-diff `fmt --all` + `clippy --all -D warnings` are clean, and all affected crates pass (`dora-daemon` 214, `dora-record-node` 5, `dora-cli` doctests). The full `cargo test --all` build exceeded this ephemeral session's fixed disk allowance (~26 GB of test artifacts), so it was run as the affected-crate set on the combined branch rather than the whole workspace; the changes touch disjoint files across three crates with no cross-crate interaction.

⚠️ All four PRs are machine-generated and marked draft — please review before merging.

---

phil-opp:

No issues found. `forget_node_bookkeeping` has only the `RemoveNode` and `ReplaceNode` callers, both of which want the cascading cause purged, so consolidating the `cascading_error_causes.forget()` there (and dropping the now-redundant explicit call in `ReplaceNode`) is behavior-preserving for `ReplaceNode` and closes the real `RemoveNode` gap where a stale `caused_by[X]` survived a remove+re-add and would misclassify the new incarnation's genuine failure as `Cascading`. The extended regression test seeds cascading causes for two nodes and asserts the removed node's cause becomes `None` while the untouched node's stays intact — it would fail without the added `forget`.

---

PR #3174, opened 2026-08-14
fix(record-node): don't panic the recorder on a pre-epoch system clock

phil-opp:
## Issue

`binaries/record-node/src/main.rs` computes timestamps with:

```rust
let now_nanos = SystemTime::now()
    .duration_since(SystemTime::UNIX_EPOCH)
    .unwrap()          // panics if now < 1970
    .as_nanos() as u64;
```

`duration_since(UNIX_EPOCH)` returns `Err` whenever the wall clock is set before 1970-01-01, and `.unwrap()` then panics. The line-142 call runs for **every recorded message**, so this aborts the recorder mid-capture (there is an identical call at startup for `start_nanos`).

## Impact

This is directly relevant to the robotics/embedded target: devices without a battery-backed RTC frequently boot with the clock at or before the epoch until NTP/GPS sync lands. The offset computation immediately after already guards the clock going *backward* with `saturating_sub(start_nanos)`, so the unwrap was an inconsistency in an otherwise-defensive path.

## Fix

Extract a small helper that saturates to `0` on a pre-epoch clock (matching the existing `saturating_sub` philosophy) and use it at both call sites:

```rust
fn unix_nanos(now: SystemTime) -> u64 {
    now.duration_since(SystemTime::UNIX_EPOCH)
        .map(|d| d.as_nanos() as u64)
        .unwrap_or(0)
}
```

## Validation

- New unit test `unix_nanos_saturates_on_pre_epoch_clock` (pre-epoch → 0, epoch → 0, post-epoch → offset).
- `cargo test -p dora-record-node` — 5 passed.
- `cargo clippy -p dora-record-node -- -D warnings` — clean.

---

---

### DIFF
--- a/binaries/record-node/src/main.rs
+++ b/binaries/record-node/src/main.rs
@@ -40,6 +40,21 @@ fn build_reverse_map(topics_json: &str) -> eyre::Result<HashMap<String, (NodeId,
     Ok(reverse_map)
 }
 
+/// Nanoseconds since the Unix epoch, saturating to 0 when the wall clock reads
+/// a pre-epoch time.
+///
+/// `SystemTime::duration_since(UNIX_EPOCH)` returns `Err` whenever the clock is
+/// set before 1970 — common on battery-less embedded/robotics hardware that
+/// boots at (or before) the epoch until NTP/GPS sync lands. This runs once per
+/// recorded message, so an `.unwrap()` here would abort the recorder mid-capture.
+/// Saturating to 0 matches the `saturating_sub` already used when computing the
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The `unix_nanos` helper saturates to 0 on a pre-epoch `Err`, consistent with the existing `saturating_sub(start_nanos)` in the offset path, and it replaces both `.unwrap()` call sites. The new unit test covers the pre-epoch (0), epoch (0), and post-epoch-offset cases, and would fail if the saturation were reverted to `unwrap()`.

---

PR #3175, opened 2026-08-14
fix(replay-node): don't abort the whole replay on one corrupt record

phil-opp:
## Issue

The replay loop in `binaries/replay-node/src/main.rs` read container records with:

```rust
while let Some(entry) = reader.next_entry()? {
```

The `?` propagates any `Err` from `RecordingReader::next_entry`, aborting the process.

But the recording layer distinguishes two cases **deliberately**:
- `Ok(None)` for EOF, the footer, or a *torn trailing* record (a crash that flushed a record's length prefix but only part of the body). `truncated_record_body_stops_gracefully` documents this.
- `Err` for a *complete* record whose internal length fields are corrupt — the bounds checks in `read_slice` / `read_array` (dora-rs/dora#2027), covered by the recording crate's `corrupt_node_id_len_returns_err_not_panic` / `corrupt_event_bytes_len_returns_err_not_panic` tests.

So a single mid-file corrupted length field (bit rot, a partial disk write, or a crafted `DORA_REPLAY_FILE`) aborted the **entire** replay and discarded every subsequent fully-intact record — the exact opposite of the skip-and-continue resilience the payload-decode and event-decode branches in the same loop already implement (each `skipped += 1; continue`).

## Fix

Route the container read through a `next_replay_entry` helper that logs a warning and ends iteration on `Err`, mirroring the recording layer's graceful torn-tail handling:

```rust
fn next_replay_entry<R: Read>(reader: &mut RecordingReader<R>, replay_file: &str) -> Option<RecordEntry> {
    match reader.next_entry() {
        Ok(entry) => entry,
        Err(e) => {
            eprintln!("warning: stopping replay at a corrupt record in {replay_file}: {e:#}");
            None
        }
    }
}
```

Every record **before** the corruption still replays, and the process no longer exits with an error. Stopping gracefully (rather than skipping onward through possibly-misaligned bytes) avoids re-interpreting corrupt data as fabricated replay outputs — the safer failure mode for a robotics pipeline.

## Validation

- New unit test `corrupt_record_stops_replay_gracefully_after_good_records`: builds a recording with one intact record followed by a complete-but-corrupt record, asserts the good record replays and the corrupt one ends iteration without error/panic.
- `cargo test -p dora-replay-node` — 9 passed.
- `cargo clippy -p dora-replay-node -- -D warnings` — clean.

---

---

### DIFF
--- a/binaries/replay-node/src/main.rs
+++ b/binaries/replay-node/src/main.rs
@@ -1,4 +1,4 @@
-use std::{fs::File, thread, time::Duration};
+use std::{fs::File, io::Read, thread, time::Duration};
 
 use dora_message::{common::Timestamped, daemon_to_daemon::InterDaemonEvent};
 use dora_node_api::{
@@ -6,7 +6,7 @@ use dora_node_api::{
     arrow::array::{ArrayData, NullArray, make_array},
     arrow_utils::decode_arrow_ipc,
 };
-use dora_recording::RecordingReader;
+use dora_recording::{RecordEntry, RecordingReader};
 use eyre::Context;
 
 /// Decode a recorded output payload back into Arrow [`ArrayData`].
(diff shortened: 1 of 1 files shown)


phil-opp:
Withdrawing this PR after a closer look during the combined-diff self-review.

The premise — that propagating the `Err` from `next_entry` "discards every subsequent fully-intact record" — doesn't hold: once a *complete* record's internal length fields are corrupt, the container stream is misaligned, so records after it are unrecoverable **either way**. The original `while let Some(entry) = reader.next_entry()? { … }` already replayed every good record *before* the corruption and only then exited non-zero.

So this change's only real effect was to turn a loud, non-zero failure into a **silent exit-0** (and, with `DORA_REPLAY_LOOP=true`, an infinite spin) when the corruption precedes the target node's records — because the new stop path never feeds the `replay_emitted_nothing_usable` guard. That's a regression, not a fix.

The recording layer deliberately distinguishes a benign torn tail (`Ok(None)`) from genuine mid-file corruption (`Err`); treating the latter as a hard error is the intended behavior. Closing.

---

PR #3178, opened 2026-08-15
fix(daemon): reclaim extension-table entries on RemoveNode/ReplaceNode

### LINKED ISSUE #3177: daemon: RemoveNode/ReplaceNode no longer reclaim a node's ExtensionTable entries (leak + missed drop-notifications) — #3152 dropped the #3014 call sites
## Summary

When #3152 (`0442564`, *"refactor(tensor-pool): move behind a generic extension seam"*) replaced the daemon's memory-pool manager with the generic `ExtensionTable`, it carried over **only one** of #3014's three reclamation call sites. The `RemoveNode` and `ReplaceNode` handlers no longer reclaim the exiting node's entries, and — because of the daemon's generation guard — the surviving `SpawnedNodeResult` call site is **unreachable** for a removed or replaced node. The result is that an entry owned by a dynamically removed/replaced node **leaks until dataflow finish**, and every node that touched that key **never receives the drop notification**.

This is a distinct failure mode from #3167 (please keep both open): #3167 is about `reclaim_owner` firing **too eagerly** on a *clean exit* (premature free). This issue is the opposite — reclaim **not firing at all** on `RemoveNode`/`ReplaceNode` (leak + missed notification). #3167's analysis assumes reclaim runs on "every node-result path (crash, clean exit, restart)"; for a *removed* or *replaced* node it does not, precisely because the exit event is swallowed by the generation guard.

## Evidence

**The documented contract.** `binaries/daemon/src/extension_table.rs:9-13` states the daemon "drops its entries **on exit** and on dataflow finish", and the `Entry.owner` field doc (`extension_table.rs:34`) says *"Its exit reclaims the entry."* `docs/extensions.md` further promises everyone who stored or loaded a key is notified when it is dropped. A dynamic `RemoveNode` is an owner exit under this contract.

**Only two reclaim call sites exist in the daemon:**
- `reclaim_extensions_of_exited_node` → `ExtensionTable::reclaim_owner`, called once, in the `SpawnedNodeResult` handler at `binaries/daemon/src/lib.rs:5883`.
- `ExtensionTable::reclaim_dataflow`, called once, at dataflow finish (`binaries/daemon/src/lib.rs:5424`).

**The `SpawnedNodeResult` site is unreachable for removed/replaced nodes.** The generation guard at `binaries/daemon/src/lib.rs:~5699-5717` returns `Ok(())` *before* reaching line 5883 whenever the node is no longer present in `running_nodes` with a matching generation:

```rust
// an exit arriving after `RemoveNode` took the entry out is dropped
// here WITHOUT running the finish accounting below.
if !current_generation.is_some_and(|node| node.matches_generation(generation)) {
    // ... log "ignoring stale exit ..." ...
    return Ok(());   // <-- returns before the reclaim at :5883
}
```

**`RemoveNode` removes the node but never reclaims.** The handler at `binaries/daemon/src/lib.rs:2776-2903` does `dataflow.running_nodes.remove(&node_id)` (`:2821`) and cleans up mappings, timers, log subscribers, `open_inputs`, `data_inputs`, `subscribe_channels`, `pending_messages`, `connected_nodes`, `finish_escalated`, `forget_node_bookkeeping`, and the stored descriptor — but makes **no** `extensions.reclaim_owner` / `reclaim_extensions_of_exited_node` call. `ReplaceNode` (`:2904-3233`) has the same gap.

**This is a regression from #3014.** In #3014 (`1ccbc00`), both handlers reclaimed explicitly, and the `ReplaceNode` comment documented exactly the generation-guard reasoning that #3152 then removed:

```rust
// RemoveNode (in 1ccbc00):
// Without this, repeated add/remove cycles walk the daemon
// into its pool cap (dora-rs/dora#2881).
if result.is_ok() { reclaim_memory_pools_after_exit(&self.memory_pool, &self.running, dataflow_id, &node_id); }

// ReplaceNode (in 1ccbc00):
// The outgoing incarnation's exit event is dropped by
// the generation guard, so this is the only place its
// pools can be reclaimed (dora-rs/dora#2881).
if result.is_ok() { reclaim_memory_pools_after_exit(...); }
```

#3152 kept the equivalent of the third site (`SpawnedNodeResult`) but dropped these two.

## Risk / impact

- **Leak + cap exhaustion.** Entries owned by a dynamically removed/replaced node are not released until the dataflow finishes (`reclaim_dataflow`). Repeated add/remove cycles of pool-owning nodes in a long-lived `dora up` dataflow accumulate toward `MAX_ENTRIES_PER_DATAFLOW = 8192` (`extension_table.rs:51`), after which `store()` fails for **every** node in that dataflow (`extension_table.rs:86-90`) — the same "repeated add/remove cycles walk the daemon into its pool cap" accumulation that #2881/#3014 set out to fix, and which #3014's now-deleted `RemoveNode` comment named verbatim.
- **Missed drop-notifications.** Because the entries are never reclaimed on removal, `reclaim_owner`'s returned `touched_by` set is never produced, so `drain_dropped_keys`/the drop notification never fires for these keys. Readers that mapped per-process resources keyed on them hold those resources until they themselves exit, instead of releasing promptly — the notification guarantee `docs/extensions.md` advertises.
- **Scope is bounded.** Only the opt-in `tensor-pool` extension (or another out-of-tree extension) writes to the `ExtensionTable`, so default builds are unaffected; the leak is per-dataflow and cleared at finish. Filed as a latent regression against the daemon's documented extension-lifetime contract rather than a crash-every-time bug.

## Suggested fix

Re-add the reclaim call to both handlers, mirroring #3014's placement: in `RemoveNode` after the successful-removal branch, and in `ReplaceNode` after a successful replace (`clear_node_result` is already there). These are the only reachable points for a removed/replaced node, since the `SpawnedNodeResult` reclaim is gated out by the generation guard. A regression test at the daemon tier — dynamically remove an entry-owning node and assert both that the table entry is gone and that a reader receives the drop notification — would pin it.

---

*This issue was created by a scheduled automated Claude code-review check (reviewing the last 2 days of merged PRs/commits). It reflects static analysis of the interaction between #3014 and #3152 and has not been reproduced live; the runtime manifestation requires the opt-in `tensor-pool` feature and dynamic node removal/replacement. Please verify before acting.*

phil-opp:
Closes #3177.

When #3152 (`0442564`) moved the daemon's memory-pool manager behind the generic `ExtensionTable`, it carried over only **one** of #3014's three reclamation call sites (`SpawnedNodeResult`). That site is **unreachable** for a dynamically removed or replaced node: the generation guard drops the node's later exit event *before* the reclaim runs, because the node is already gone from `running_nodes` (`RemoveNode`) or superseded by a new incarnation (`ReplaceNode`).

As a result, an extension-table entry owned by a removed/replaced node was not released until dataflow finish (`reclaim_dataflow`). Repeated add/remove cycles of an entry-owning node accumulate toward `MAX_ENTRIES_PER_DATAFLOW = 8192`, after which `store()` fails for *every* node in the dataflow — the exact #2881 accumulation that #3014 set out to fix. Readers keyed on those entries also never received the drop notification the extension contract (`docs/extensions.md`) advertises.

## Fix

- **`RemoveNode`** — reclaim after a successful removal. This is also the only release for a node that never reports an exit at all (a dynamic node has no process handle to wait on).
- **`ReplaceNode`** — reclaim after a successful replace, next to the existing `clear_node_result`. The replacement's higher generation is installed first and the event loop is serial, so this drops only the outgoing incarnation's entries and its later stores are dropped as superseded (#2926/#2927).
- **`SpawnedNodeResult`, generation-guard branch** — a node removed while alive keeps its daemon connection for the whole stop grace window (10s, hard kill at 15s), and the node-event guard deliberately lets its events through once its `running_nodes` entry is gone. So it can still `store()` after `RemoveNode`'s eager reclaim, and those entries have no other release point. Reclaim again when the process actually dies — but only when nothing owns the id, since ownership is keyed by node id alone and a replacement or re-add would otherwise lose its entries.

The known residual: a node removed and re-added under the same id inherits its predecessor's grace-window entries until dataflow finish. Closing that needs generation-aware ownership in `ExtensionTable`, which needs the storing incarnation's generation threaded down to `store` — documented on `Entry::owner` rather than fixed here.

## Scope / risk

Bounded to the opt-in `tensor-pool` extension (or another out-of-tree extension) — nothing in a default build writes to the `ExtensionTable`, so default builds are unaffected. Latent regression against the daemon's documented extension-lifetime contract (Class A).

Not covered, and filed separately: after the cross-machine memory-pool work in `c1b225ea9`, peer daemons hold mirror descriptors owned by the *remote* sender node. `RemoveNode` reaches only the hosting daemon and node removal publishes no `FreePool`, so the same accumulation remains on the peer.

## Testing

`cargo fmt --all -- --check`, `cargo clippy --all -- -D warnings`, `cargo test --all` (workspace, Python crates excluded), `cargo check --examples` — all clean. New unit test covers the daemon-side half of reclamation that `extension_table.rs` cannot reach: the `ExtensionDropped` notification arriving on a reader's subscribe channel.

### DIFF
--- a/binaries/daemon/src/extension_table.rs
+++ b/binaries/daemon/src/extension_table.rs
@@ -32,6 +32,15 @@ pub struct ExtensionKey {
 struct Entry {
     value: Vec<u8>,
     /// The node that stored it. Its exit reclaims the entry.
+    ///
+    /// A bare id, not (id, incarnation): a node that is removed and re-added
+    /// under the same id inherits whatever its predecessor stored, because
+    /// the daemon cannot tell the two apart here. That costs the removed
+    /// incarnation's stop-grace-window entries their prompt release — they
+    /// live until dataflow finish (dora-rs/dora#3177). Making ownership
+    /// generation-aware would close it; it needs the storing incarnation's
+    /// generation threaded down to `store`, which the daemon drops before
+    /// the `ExtensionStore` arm today.
     owner: NodeId,
     /// Everyone who stored or loaded it, and so must be told when it is
     /// dropped. Includes the owner.
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2842,6 +2842,25 @@ impl Daemon {
                     Ok(())
                 })();
 
+                // Reclaim the removed node's extension-table entries. Its
+                // later `SpawnedNodeResult` exit event is swallowed by the
+                // generation guard (the node is already gone from
+                // `running_nodes`), so the reclaim at that call site is
+                // unreachable for a removed node — this is the only reachable
+                // point. Without it, repeated dynamic add/remove cycles of an
+                // entry-owning node leak entries toward the per-dataflow cap
+                // and readers never receive the drop notification
+                // (dora-rs/dora#3177, originally #2881/#3014).
+                if result.is_ok() {
+                    reclaim_extensions_of_exited_node(
+                        &mut self.extensions,
+                        self.running.get(&dataflow_id),
+                        dataflow_id,
+                        &node_id,
+                        &self.clock,
+                    );
+                }
+
                 // Outside the closure because it is async. Why removal
                 // has to drive the barrier at all: see
                 // `PendingNodes::handle_node_removal`.
@@ -3225,6 +3244,20 @@ impl Daemon {
                     // The outgoing (or an even earlier) incarnation's stale
                     // result must not be attributed to the replacement.
                     clear_node_result(&mut self.dataflow_node_results, dataflow_id, &node_id);
+                    // Reclaim the outgoing incarnation's extension-table
+                    // entries. Its exit event is dropped by the generation
+                    // guard, so this is the only reachable point to release
+                    // them. The freshly spawned replacement shares the node id
+                    // but has not stored anything yet, so this drops only the
+                    // outgoing incarnation's entries (dora-rs/dora#3177,
+                    // originally #2881/#3014).
+                    reclaim_extensions_of_exited_node(
+                        &mut self.extensions,
+                        self.running.get(&dataflow_id),
+                        dataflow_id,
+                        &node_id,
+                        &self.clock,
+                    );
                 }
                 let reply = DaemonCoordinatorReply::ReplaceNodeResult(
                     result.map_err(|err| format!("{err:?}")),
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -850,6 +850,21 @@ fn extract_node_results(
 }
 
 impl Daemon {
+    /// Release every extension-table entry owned by `node_id`, notifying the
+    /// nodes that read them. Thin wrapper over the free function so the call
+    /// sites do not each repeat the `running` lookup — the exception is the
+    /// clean-exit site, which holds a `&mut self.logger` borrow and so must
+    /// name the disjoint fields itself.
+    fn reclaim_extensions(&mut self, dataflow_id: DataflowId, node_id: &NodeId) {
+        reclaim_extensions_of_exited_node(
+            &mut self.extensions,
+            self.running.get(&dataflow_id),
+            dataflow_id,
+            node_id,
+            &self.clock,
+        );
+    }
+
     /// Runs the daemon. `zenoh_listen_addr` overrides the address this
     /// daemon's zenoh listener binds, and therefore the locator its peers are
     /// told to dial.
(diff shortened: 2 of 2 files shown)


phil-opp:

No issues found. I traced both call sites:

- The `RemoveNode` reclaim uses the closure's `result` (before it is shadowed) and runs after the node has left `running_nodes` while the dataflow is still present in `self.running`, so reader drop-notifications are still delivered, and a later `SpawnedNodeResult` for the removed id is dropped by the generation guard (so there is no double reclaim; `reclaim_owner` is idempotent anyway).
- The `ReplaceNode` reclaim sits inside `if result.is_ok()` after `clear_node_result`. Because the daemon processes coordinator events sequentially, the freshly-spawned same-id incarnation cannot have issued a `store` yet, so `reclaim_owner(node_id)` only releases the outgoing incarnation's entries.

Scope is bounded to the opt-in `tensor-pool` extension. Note this overlaps the `ReplaceNode` region touched by #3171, but the two edit disjoint lines and don't conflict semantically.

---

phil-opp:

The fix is correct and the premise checks out: the generation guard in `DoraEvent::SpawnedNodeResult` (`binaries/daemon/src/lib.rs:8314-8330`) returns before the existing reclaim at `:8497`, so a removed or replaced node's extension-table entries really do leak today. No use-after-free — `reclaim_owner` only drops entries whose `owner` is the departing node, and readers get the standard drop notification. The `ReplaceNode` ordering claim also holds: the replacement's `ExtensionStore` request cannot be processed before the reclaim, because the daemon event loop is blocked inside the awaited `ReplaceNode` async block for the whole spawn.

The gap is **no test**, for precisely the kind of leak that regresses invisibly — a future refactor that moves the reclaim would be undetectable. The existing `RemoveNode`-cleanup unit tests around `binaries/daemon/src/lib.rs:11100` poke `RunningDataflow` state directly and cannot reach a handler-level call, so the options are a `tests/`-level dynamic add/remove loop asserting `ExtensionTable::len()` stays flat, or an explicit note that this is untestable at unit granularity.

Minor side effect: `subscribe_channels.remove(&node_id)` runs before the reclaim, so the departing node now appears in the `undelivered` warn list on every removal.

phil-opp:
Re-reviewing after the new commit `55db4b06` ("close the stop-grace-window half of the extension leak"), which adds the second reclaim on the dropped exit event plus a unit test. This part looks safe to me:

- In the `SpawnedNodeResult` entry-absent branch, `node_id_unowned = current_generation.is_none()` correctly gates the reclaim: it's `true` once `RemoveNode` has taken the id out of `running_nodes` (so the grace-window stores get released), and `false` when a replacement or re-add now owns the id — so a live incarnation's entries are not swept away. Good call keying the guard off `current_generation` rather than re-looking-up.
- The extra reclaim is idempotent — `reclaim_owner` only drops entries whose `owner == node_id`, so a normal removal with no grace-window `store()` is a no-op and there's no double `ExtensionDropped` for the same key.
- This also closes the "no test" gap noted earlier: `reclaim_drops_owned_entries_and_notifies_readers` asserts both that the owner's entry is gone and that the `ExtensionDropped` notification reaches the reader's subscribe channel, and it exercises the shared `reclaim_extensions_of_exited_node` path all call sites use.

No new issues found in the new commit.

---

---

PR #3180, opened 2026-08-15
fix(coordinator): roll back succeeded daemons on async partial spawn failure

### LINKED ISSUE #3134: coordinator: async partial spawn failure (one daemon Errs after another already spawned) never rolls back the succeeded daemon — orphaned nodes + `running_dataflows`/store divergence (`dora list` stuck "Running")

keirsalterego:
Closes #3134.

On a multi-daemon dataflow the daemons report spawn results independently, so one can fail well after another has already started its nodes. The Err arm of `DataflowSpawnResult` only recorded that much: it cached the error and persisted a terminal `Failed` record. It never stopped the daemons that had already spawned and never dropped the dataflow from `running_dataflows`.

Nothing else covers for it. The spawn-timeout watchdog skips the dataflow once its spawn result is no longer pending, and the orphan-stop from #3119 only runs on a daemon status report, which each daemon sends once at startup. So the healthy daemon keeps running nodes nobody manages, and `dora list` says Running while the store says Failed, until someone stops it by hand.

The other two partial-failure paths already do the right thing here: `run::spawn_dataflow` rolls back a synchronous partial failure, and `check_spawn_timeouts` rolls back on a timeout. So rather than write a third variant, this reuses the watchdog's:

- Pulled the watchdog's rollback + in-memory teardown into `teardown_failed_spawn` (stop the started daemons, cancel a parked restart, final log line, close topic subscribers, synthesize per-node `FailedToSpawn` results, drain stop waiters, archive, cap). The watchdog keeps its own timeout message and delegates the rest, so nothing about its behaviour changes.
- Moved the event arm into `handle_dataflow_spawn_result` so a test can call the real thing instead of replaying it.
- #3114's pending-restart cancel moved into the shared helper too. Otherwise the async path tears the dataflow down and leaves a `PendingRestart` keyed to a dead UUID, which hangs the parked restart caller and makes every later Stop for that UUID fail.

Both orderings end up rolled back:

- **A succeeds, then B fails.** B's failure is the terminal transition, so everything no longer waiting on a spawn result gets stopped: A, plus B in case it started some nodes before erroring.
- **B fails, then A succeeds.** The teardown already happened, so A's Ok finds nothing in `running_dataflows`. It is in `archived_dataflows`, which tells us we failed it, so A gets stopped on its own. This one is worth handling, not just for symmetry: once the teardown has synthesized a `dataflow_results` entry, `dora stop <uuid>` early-returns without contacting any daemon, so the manual cleanup path is gone and the orphan would be permanent. A fail-fast spawn error (missing binary, bad path) also tends to come back before a daemon that has to start every node, so the ordering isn't rare.

The test drives `handle_dataflow_spawn_result` through both orderings and checks the started daemon gets a `StopDataflow`, the dataflow leaves `running_dataflows` and is archived, the store record is `Failed`, in-flight `dora start` / `dora stop` waiters are released instead of hanging, a parked restart is cancelled with the spawn error, and the late Ok stops that daemon too. Confirmed it fails without the fix.

`cargo test -p dora-coordinator` (126 lib + 26 integration), `cargo clippy -p dora-coordinator --all-targets -- -D warnings` and `cargo fmt --all -- --check` are green, plus the full workspace `cargo test`.

One note: #3136 fixes the same issue and I only found it after writing this. Same idea for the shared helper. The differences are the late-Ok rollback (out of scope there) and having the test call the handler rather than re-implement its statements. No strong feelings about which lands — happy to close this in favour of #3136 plus a follow-up if that's simpler for you.


phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review, no human in the loop. Treat it as advisory.

No issues found. I traced the new `handle_dataflow_spawn_result` / `teardown_failed_spawn` split against the unchanged helpers:

- In the terminal-Err arm, `is_terminal_transition` is captured from `spawn_result.is_pending()` *before* `handle_spawn_result_err` resolves it, so only the first failure owns the rollback. `handle_spawn_result_err` sets `spawn_result` and persists the `Failed` record; `teardown_failed_spawn` deliberately does not re-set `spawn_result` (the watchdog path sets it before calling), so there is no double-set or store clobber.
- The `started` rollback set = `daemons.difference(pending_spawn_results)` after the failing daemon was already removed from `pending_spawn_results`, so it correctly covers the already-succeeded daemons (live nodes to stop) plus the failing daemon (may have partially spawned). Daemons still pending are excluded.
- The mirror ordering (success after teardown) lands in the `None` arm and force-stops that daemon via `fire_and_forget_rollback`, since teardown always removed the dataflow from `running_dataflows` and archived it.
- Watchdog behavior is preserved: it keeps its own timeout wording, sets `spawn_result`/persists `Failed` itself, then delegates the rest. The rollback being fire-and-forget mpsc `send` makes its reordering relative to the store write immaterial.

The added test is not tautological: it drives both orderings and would fail if the rollback were reverted — it asserts the started daemon receives a `StopDataflow`, the dataflow leaves `running_dataflows` and is archived, the spawn/stop/restart waiters all resolve (rather than hang), and the parked `PendingRestart` is cancelled with a reason.

---

PR #3181, opened 2026-08-15
fix(mavlink2-bridge): reject an invalid `?baud=` instead of silently using 115200

phil-opp:
## Issue

`connect_serial` in `libraries/extensions/mavlink2-bridge/src/transport.rs` resolved the serial baud with:

```rust
let baud = parse_baud(url).unwrap_or(DEFAULT_SERIAL_BAUD);
```

where `parse_baud` returned `Option<u32>` and mapped **both** an absent `?baud=` **and** a present-but-invalid one to `None`. So a typo such as `serial:///dev/ttyUSB0?baud=57600x` (or a stray `?baud=0`) silently fell back to 115200 and the endpoint connected at the wrong baud **with no diagnostic** — producing a garbled or dead serial link.

This is exactly the silent-misconfiguration failure mode the sibling `parse_proto_query` was deliberately rewritten to prevent. Its own doc comment says:

> The explicit error on an unrecognized value is deliberate: a typo such as `?proto=v3` previously fell back to V2 silently, so a misconfigured endpoint connected on the wrong protocol with no diagnostic.

The two query parsers were inconsistent: `proto` distinguished *absent* (default) from *invalid* (hard error), while `baud` conflated them.

## Fix

`parse_baud` now returns `BridgeResult<u32>`, mirroring `parse_proto_query`:

- absent `?baud=` → defaults quietly to `DEFAULT_SERIAL_BAUD` (115200);
- a positive integer → that rate;
- anything else (unparseable, or `0`) → a `BridgeError::Config` naming the bad value and URL.

`connect_serial` propagates the error with `?`.

## Validation

- The old `parses_baud_garbage` test (which asserted the silent-`None` behavior) is replaced by `rejects_baud_garbage`; a new `rejects_baud_zero` test covers the `?baud=0` case; `parses_baud_missing` now asserts the default is returned.
- `cargo fmt -p dora-mavlink2-bridge -- --check`, `cargo clippy -p dora-mavlink2-bridge -- -D warnings`, and `cargo test -p dora-mavlink2-bridge` all pass.

---

---

### DIFF
--- a/libraries/extensions/mavlink2-bridge/src/transport.rs
+++ b/libraries/extensions/mavlink2-bridge/src/transport.rs
@@ -89,15 +89,36 @@ fn connect_serial(url: &Url) -> BridgeResult<Box<dyn MavConnection<MavMessage> +
             "missing device path in '{url}'"
         )));
     }
-    let baud = parse_baud(url).unwrap_or(DEFAULT_SERIAL_BAUD);
+    let baud = parse_baud(url)?;
     let version = parse_proto_query(url)?;
     open_mavlink_versioned(&format!("serial:{device}:{baud}"), version)
 }
 
-fn parse_baud(url: &Url) -> Option<u32> {
-    url.query_pairs()
-        .find(|(k, _)| k == "baud")
-        .and_then(|(_, v)| v.parse::<u32>().ok())
+/// Resolve the serial baud rate from an optional `?baud=` query parameter.
+///
+/// - absent → the [`DEFAULT_SERIAL_BAUD`] default;
+/// - a positive integer → that rate;
+/// - anything else (unparseable, or `0`) → a [`BridgeError::Config`].
+///
+/// The explicit error on an invalid value is deliberate and mirrors
+/// [`parse_proto_query`]: a typo such as `?baud=57600x` (or a stray `?baud=0`)
+/// previously fell back to `115_200` silently, so a misconfigured endpoint
+/// connected at the wrong baud with no diagnostic — producing a garbled or
+/// dead serial link. An absent parameter is distinct from an invalid one and
+/// still defaults quietly.
+fn parse_baud(url: &Url) -> BridgeResult<u32> {
+    match url.query_pairs().find(|(k, _)| k == "baud") {
+        None => Ok(DEFAULT_SERIAL_BAUD),
+        Some((_, v)) => v
+            .parse::<u32>()
+            .ok()
+            .filter(|&baud| baud > 0)
+            .ok_or_else(|| {
+                BridgeError::Config(format!(
+                    "invalid serial baud '{v}' in '{url}' (expected a positive integer)"
+                ))
+            }),
+    }
 }
 
 fn open_mavlink_versioned(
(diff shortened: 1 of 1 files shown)


phil-opp:

This PR is one of a batch opened by an automated Claude Code pass over the repository, branched off `origin/main` at `0442564`. Every PR is **machine-generated and needs human review**. Each was independently verified (`cargo fmt` + `clippy -D warnings` + affected-crate tests), and the combined diff passed the `/review` and `/simplify` quality gates plus a combined test run of all affected crates (dora-core 341+ tests, dora-daemon 216, mavlink lib/node — all green, 0 failures). The full `cargo test --all` could not finish in this environment because the unrelated `dora-ros2-bridge` **examples** exhausted the sandbox disk allowance; ros2-bridge is untouched by this batch and no other crate references the changed behavior, so the affected-crate run stands in for it.

### PRs opened this run

**Correctness / diagnostics**
- **#3181** — `fix(mavlink2-bridge)`: reject an invalid `?baud=` instead of silently falling back to 115200 (mirrors the deliberately-strict `?proto=` parser).
- **#3182** — `fix(core)`: stop reporting an annotated-but-**typoed** upstream output as having "no type annotation" in strict type-checking.

**Robustness**
- **#3184** — `fix(mavlink2-bridge-node)`: `#[serde(deny_unknown_fields)]` on `MAVLINK_BRIDGE_CONFIG` so a typo'd key is a loud error, not a silent wrong default.

**Performance / hardening**
- **#3183** — `perf(daemon)`: make the extension-table per-dataflow cap check O(1) instead of an O(total-entries) scan (O(N²) in exactly its adversarial store-per-frame case).

### Findings considered but deliberately **not** turned into PRs

- **mavlink `MavlinkArrow::schema()` rebuilt per telemetry frame → cache in `LazyLock`**: a genuine per-frame allocation on a hot path, but the fix touches all 18 message impls; the review burden outweighs the modest bounded-rate gain. *Deferred.*
- **runtime-shared-lib: `arrow::ffi::to_ffi` error tears down the operator**: a very-low-probability path (nearly all Arrow types are C-Data-Interface-representable) and the candidate "fix" (silently dropping the input) is arguably worse than failing loudly. *Rejected.*
- **runtime-shared-lib: unreachable `Event::Error` match arm**: the runtime layer bails on `Event::Error` before it can reach the operator, but reconciling it changes error semantics — too speculative for a mechanical cleanup. *Rejected.*
- **runtime-shared-lib: redundant nested `#[cfg(feature = "telemetry")]`**: correct but a one-line cosmetic cleanup, too trivial for a standalone PR. *Rejected.*
- **runtime-api: per-event watchdog mutex traffic**: two uncontended locks per event — negligible, not worth the churn. *Rejected.*
- **node: `decode_arrow_ipc` doesn't enforce its "exactly one record batch" doc**: no encoder emits multiple, so latent only; not worth a behavior change. *Rejected.*
- **core: `send_logs_as` lacks the single-operator informational warning `send_stdout_as` emits**: plausibly an oversight, but possibly intentional — not confident enough. *Rejected.*
- **core: universal `*→Bytes` sink bypassed by a *parameterized* `Bytes` target**: only bites a user who writes a (meaningless) parameter on `Bytes`. Low value. *Rejected.*
- **cli: `trace/list.rs` raw `&trace_id[..12]` byte-slice without a char-boundary guard**: trace IDs are ASCII hex, so unreachable in practice; defensive-only. *Rejected.*
- **Assorted public-API doc gaps** (mavlink `MavlinkArrow` methods, operator-types FFI helpers): minor, not defects. *Rejected.*

### Subsystems audited and found already well-hardened (no action)

hub-client, coordinator-store (redb + in-memory), recording (`.drec`), arrow-convert, telemetry span store, ros2-bridge Arrow (de)serialize, `dora-message`, node data-conversion / IPC paths, cli command surface, and the daemon/coordinator hot paths.

---

phil-opp:
**Automated review by Claude** 🤖 _(fully automated review — findings have not been vetted by a human)_

No issues found. I reviewed the `parse_baud` change in `libraries/extensions/mavlink2-bridge/src/transport.rs` against its call site and the sibling `parse_proto_query`:

- The three cases are handled correctly and distinctly: an absent `?baud=` returns `Ok(DEFAULT_SERIAL_BAUD)` (omission is not a misconfiguration), a present value that parses as a positive `u32` returns that rate, and anything else — unparseable or `0` — returns `BridgeError::Config`. The `.filter(|&baud| baud > 0)` correctly rejects `?baud=0` before it could be formatted into `serial:/dev/...:0` and handed to the mavlink layer.
- `connect_serial` now propagates the error via `?` instead of `unwrap_or(DEFAULT_SERIAL_BAUD)`, so the silent wrong-baud fallback is eliminated and the behavior now mirrors the deliberately-strict `?proto=` parser, as intended.
- The tests were updated consistently: `parses_baud_missing` asserts the default, `rejects_baud_garbage` asserts a `Config` error (matched via `matches!`), and the new `rejects_baud_zero` covers the `0` case. These exercise the real branches and are not tautological.

No correctness, overflow, or compatibility concerns.

---

PR #3182, opened 2026-08-15
fix(core): don't report an annotated-but-typoed upstream as having "no type annotation"

phil-opp:
## Issue

In strict type-checking mode (`strict_types: true`), an upstream output whose declared type URN does **not resolve** — i.e. a typo — is reported to the user as if it carried *no type annotation at all*.

Given:

```yaml
nodes:
  - id: sender
    outputs: [image]
    output_types:
      image: std/media/v1/Imag      # typo of ...Image
  - id: receiver
    inputs:
      image: sender/image
    input_types:
      image: std/media/v1/Image
strict_types: true
```

the user gets **two** warnings:

```
unknown type "std/media/v1/Imag" on output "image" (did you mean "std/media/v1/Image"?)   ✅ correct
input "image" expects type "std/media/v1/Image" but upstream sender/image has no type annotation   ❌ false
```

The second is factually wrong — the upstream *is* annotated, just misspelled — and it sends the user chasing a nonexistent missing-annotation instead of fixing the typo the first warning already points at.

**Root cause:** in `check_type_annotations_full`, an output is only inserted into `output_type_map` when `registry.resolve(urn).is_some()`. So an unresolvable annotation never lands in the map and is indistinguishable, at the strict-mode edge check, from a genuinely absent one — which then falls into the `(None, Some(in_urn)) if strict` arm that emits "has no type annotation".

## Fix

Track every output port that carries *some* annotation — regardless of whether its URN resolves — in a new `annotated_output_ports: BTreeSet<(String, String)>` (populated alongside `output_type_map` at all three registration sites: node-level, single-`operator:`, and `operators:` runtime, using the same key forms). The strict-mode "no annotation" arm now also requires `!annotated_output_ports.contains(&key)`, so an annotated-but-unresolvable upstream falls through with no warning here. The unknown-URN warning from `check_port_types` still fires, and genuinely unannotated upstreams still warn exactly as before.

## Validation

- New regression test `strict_mode_no_missing_annotation_warning_for_typoed_upstream`: **fails before the fix** (asserts the misleading "no type annotation" warning is absent — both warnings were emitted) and **passes after**. Verified the RED state directly by reverting only the guard.
- Full `cargo test -p dora-core` passes (282 lib tests + doctests); `cargo fmt -p dora-core -- --check` and `cargo clippy -p dora-core -- -D warnings` clean.

---

---

### DIFF
--- a/libraries/core/src/descriptor/validate.rs
+++ b/libraries/core/src/descriptor/validate.rs
@@ -920,6 +920,26 @@ pub struct TypeCheckResult {
 /// Timer nodes auto-inject this type.
 const TIMER_TYPE: &str = "std/core/v1/UInt64";
 
+/// Register one declared output type under `key` (a `(node_id, output_ref)`
+/// pair). The port is *always* recorded in `annotated` so a typo'd or
+/// otherwise unresolvable URN still suppresses the strict-mode "upstream has
+/// no type annotation" warning; it is additionally recorded in
+/// `output_type_map` (the map used for cross-edge type matching) only when its
+/// URN resolves. Keeping both writes behind one call makes the two structures
+/// impossible to desync — the invariant the strict-mode check relies on.
+fn register_output_type(
+    output_type_map: &mut BTreeMap<(String, String), String>,
+    annotated: &mut BTreeSet<(String, String)>,
+    key: (String, String),
+    urn: &str,
+    registry: &crate::types::TypeRegistry,
+) {
+    annotated.insert(key.clone());
+    if registry.resolve(urn).is_some() {
+        output_type_map.insert(key, urn.to_string());
+    }
+}
+
 /// Check type annotations in a dataflow, with strict mode support.
 ///
 /// Returns the collected warnings plus any inferred edge types.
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I reviewed the diff against the surrounding code in `libraries/core/src/descriptor/validate.rs`.

- The new `annotated_output_ports` set is populated through the single `register_output_type` helper with the *same* keys as `output_type_map` at every registration site (node-level, single-`operator:` node — both the `op/output` prefixed key and the bare `output` key — and `operators:` runtime). The strict-mode suppression guard `!annotated_output_ports.contains(&key)` uses the identical `key = (mapping.source, mapping.output)` that indexes `output_type_map`, so the two structures stay aligned across all edge-reference forms.
- The premise holds: `check_port_types` still reports the unresolvable URN (`unknown type "..." ... did you mean ...?`) from the sender's `output_types`, so the typo is not silently swallowed — only the misleading second "no type annotation" warning is suppressed. A genuinely unannotated upstream is absent from `annotated_output_ports`, so it still warns exactly as before.
- The regression test `strict_mode_no_missing_annotation_warning_for_typoed_upstream` exercises the real strict `(None, Some)` branch and asserts both the presence of the unknown-type warning and the absence of the "no type annotation" warning — not a tautology.

The change is well-scoped and I don't see correctness, concurrency, or compatibility concerns.

---

PR #3183, opened 2026-08-15
perf(daemon): make the extension-table per-dataflow cap check O(1)

phil-opp:
## Issue

`ExtensionTable::store` (in `binaries/daemon/src/extension_table.rs`) enforces a per-dataflow entry cap (`MAX_ENTRIES_PER_DATAFLOW = 8192`) so a looping node cannot exhaust the daemon's memory with a store-per-frame. The cap check computed the current count by scanning **every** key in the whole `entries` map — which is shared across *all* dataflows on the daemon — and string-comparing each 36-char UUID `dataflow_id`:

```rust
let count = self.entries.keys()
    .filter(|k| k.dataflow_id == key.dataflow_id).count();
```

So a single new-key `store` is O(total-entries-across-all-dataflows), and filling one dataflow toward its cap is **O(N²)** (~67M string comparisons to reach 8192, more when several dataflows coexist). The guard exists precisely for the adversarial store-per-frame pattern — which is exactly the pattern under which this scan is worst — so the bound meant to contain a runaway is itself quadratic in the runaway's size.

## Fix

Maintain a `per_dataflow_count: HashMap<String, usize>` in step with `entries`:

- incremented on insert in `store`;
- decremented (via a small `decrement` helper) in `drop_key`, `reclaim_owner`, and `reclaim_dataflow`, dropping the counter key at zero so an idle dataflow leaves no residue.

The cap check becomes an O(1) `HashMap` lookup. Observable behavior is unchanged — this is a pure internal accounting change.

## Validation

- The existing `the_per_dataflow_cap_is_enforced_and_scoped` and `overwriting_at_cap_still_works` tests still pass, confirming the O(1) count matches the previous scan.
- New tests:
  - `dropping_an_entry_at_cap_frees_a_slot` — at cap, a fresh key is rejected; after `drop_key` frees one, a store succeeds (exercises the decrement).
  - `reclaim_paths_keep_the_count_in_step` — cross-checks `per_dataflow_count` against a recomputed scan of `entries` after a mix of store/drop/reclaim, and asserts no zero-valued residue remains.
- `cargo fmt`/`cargo clippy -p dora-daemon -- -D warnings` clean; `cargo test -p dora-daemon extension_table` → 16 passed.

Note: this is complementary to (and does not conflict with) the open PR #3178, which adds new `reclaim` call sites in `daemon/src/lib.rs`; those go through `reclaim_owner`, which this change already keeps in step with the counter.

---

---

### DIFF
--- a/binaries/daemon/src/extension_table.rs
+++ b/binaries/daemon/src/extension_table.rs
@@ -43,6 +43,12 @@ struct Entry {
 #[derive(Debug, Default)]
 pub struct ExtensionTable {
     entries: HashMap<ExtensionKey, Entry>,
+    /// Live entry count per `dataflow_id`, kept in step with `entries` so the
+    /// per-dataflow cap check in [`store`](Self::store) is O(1) instead of an
+    /// O(total-entries) scan of every key across every dataflow. Invariant: a
+    /// dataflow's value equals the number of `entries` keys with that
+    /// `dataflow_id`, and the key is absent once that count reaches zero.
+    per_dataflow_count: HashMap<String, usize>,
 }
 
 /// Upper bound on entries per dataflow, so a looping node cannot exhaust the
@@ -79,10 +85,10 @@ impl ExtensionTable {
             }
             None => {
                 let count = self
-                    .entries
-                    .keys()
-                    .filter(|k| k.dataflow_id == key.dataflow_id)
-                    .count();
+                    .per_dataflow_count
+                    .get(&key.dataflow_id)
+                    .copied()
+                    .unwrap_or(0);
                 if count >= MAX_ENTRIES_PER_DATAFLOW {
                     return Err(format!(
                         "extension table full for this dataflow ({MAX_ENTRIES_PER_DATAFLOW} entries); \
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I traced every path that mutates `entries` against the new `per_dataflow_count` counter in `binaries/daemon/src/extension_table.rs`:

- `store` new-key branch increments only after the cap check passes, then inserts — count stays in step. The overwrite branch correctly leaves the counter untouched, preserving `overwriting_at_cap_still_works`.
- `drop_key` and `reclaim_owner` call `decrement` exactly once per successful `entries.remove`, so `*count -= 1` cannot underflow given the stated invariant.
- `reclaim_dataflow` drops the whole counter key, matching the wholesale removal of that dataflow's entries and leaving other dataflows alone.
- `decrement` removes the key at zero, so idle dataflows leave no residue.

The two new tests exercise the decrement path (`dropping_an_entry_at_cap_frees_a_slot`) and independently recompute the invariant after a store/drop/reclaim mix (`reclaim_paths_keep_the_count_in_step` via `assert_count_invariant`) — real checks, not tautologies. Observable behavior is unchanged and the O(N²) scan is now an O(1) lookup.

---

PR #3184, opened 2026-08-15
fix(mavlink2-bridge-node): reject unknown keys in MAVLINK_BRIDGE_CONFIG

phil-opp:
## Issue

The bridge node's `Config` (in `binaries/mavlink2-bridge-node/src/main.rs`) is deserialized from the `MAVLINK_BRIDGE_CONFIG` env var but did **not** set `#[serde(deny_unknown_fields)]`. So a mistyped key was silently ignored and the corresponding field kept its default:

```yaml
endpoint: udp://127.0.0.1:14550
systemid: 10        # typo of system_id
```

parses successfully, the `systemid` line is dropped, and the node quietly runs with `system_id = 255` — no error, no warning. This is the same silent-misconfiguration footgun that the transport-layer query parsers (`?proto=`, and now `?baud=`) explicitly guard against.

## Fix

Add `#[serde(deny_unknown_fields)]` to `Config` so an unknown/typo'd key becomes a hard parse error, surfaced through the existing `.context("parsing MAVLINK_BRIDGE_CONFIG")`. The three real keys (`endpoint`, `system_id`, `component_id`) are unaffected, and this matches the convention already used by other config structs in the workspace (e.g. the dataflow descriptor).

### Note on YAML merge keys (`<<: *anchor`)

`deny_unknown_fields` also rejects a top-level YAML merge key. I verified this does **not** remove working behavior: `serde_yaml` 0.9 does not apply merge keys when deserializing directly into a struct via `from_str`, so a merge key was already silently dropped — e.g. `<<: {system_id: 7}` + `endpoint: ...` deserialized to `system_id = 0` (the default), not 7. So merge keys never functioned in this path; this change turns that silently-wrong config into a loud error, consistent with the PR's intent, rather than removing a capability.

## Validation

- `config_rejects_unknown_field` — a `systemid` typo now produces a parse error naming the unknown field.
- `config_parses_minimal_and_applies_defaults` — a minimal valid config still parses and applies the `system_id`/`component_id` defaults.
- `cargo fmt`/`cargo clippy -p dora-mavlink2-bridge-node -- -D warnings` clean; full bin unit suite → 17 passed.

---

### DIFF
--- a/binaries/mavlink2-bridge-node/src/main.rs
+++ b/binaries/mavlink2-bridge-node/src/main.rs
@@ -168,7 +168,13 @@ fn classify_recv_error(err: &MessageReadError, datagram: bool) -> RecvDispositio
     }
 }
 
+/// `deny_unknown_fields` turns a mistyped key into a hard parse error rather
+/// than silently ignoring it and falling back to the default. Without it,
+/// `systemid: 10` (missing underscore) is dropped and the node quietly runs
+/// with `system_id = 255`; a `MAVLINK_BRIDGE_CONFIG` typo would otherwise
+/// misconfigure the bridge with no diagnostic.
 #[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
 struct Config {
     endpoint: String,
     #[serde(default = "default_system_id")]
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. Adding `#[serde(deny_unknown_fields)]` to `Config` is correct and composes cleanly with the `#[serde(default = ...)]` fields — defaulted keys, when present, still parse, and only genuinely unrecognized keys now error (surfaced through the existing `.context("parsing MAVLINK_BRIDGE_CONFIG")`). Both new tests are meaningful rather than tautological: `config_rejects_unknown_field` exercises the new failure path (a `systemid` typo now hard-errors) and `config_parses_minimal_and_applies_defaults` confirms the defaults still apply. The YAML merge-key caveat in the description is accurate for `serde_yaml` 0.9 direct-struct deserialization. This is a low-risk, well-targeted hardening change.

---

PR #3186, opened 2026-08-16
fix(daemon): reject a stranger's higher generation in the node-event gate (#2997)

### LINKED ISSUE #2997: daemon: `dora node replace` racing a restart of the same node misattributes the restart-zombie's events (higher generation) to the replacement — silent hang / misrouted outputs
> ⚠️ **This issue was created by a scheduled, automated Claude code-review check.** It was not filed by a human. Please verify the finding before acting on it. This is a race, so it is intermittent, but the analysis below is concrete and the consequences are severe.

## Summary

The node-lifecycle generation guard introduced with `dora node replace` (#2988, commit `79afcc8`) drops **strictly-older** node events only:

```rust
// binaries/daemon/src/lib.rs:347
fn event_generation_is_stale(entry_generation: u64, event_generation: u64) -> bool {
    event_generation < entry_generation
}
```

Its justification (doc comment at 339-346 and the call-site comment at 1685-1695) is that "a newer-than-entry generation is always the entry's own racing restart successor," so a newer generation must be let through or the successor's one-shot `Subscribe` is lost. That invariant holds for the **restart** mechanism — but **`dora node replace` breaks it**, because a replacement is a brand-new incarnation with its **own** generation that can be **lower** than a concurrent restart of the outgoing incarnation. When that happens, the outgoing incarnation's restart-zombie carries a *higher* generation than the replacement, so its `Subscribe`/`SendOut`/`OutputsDone` events pass the gate and are applied to the replacement.

## Why the inversion is possible

Both the replacement spawn and the restart loop draw from the **same global monotonic counter**, and the replacement's generation is minted *early* (before its slow build), while restart is disabled *late*:

1. `spawn_node` mints the replacement's generation from the global counter **before** the build:
   `binaries/daemon/src/spawn/spawner.rs:465` → `let generation = crate::running_dataflow::next_node_generation();`
2. The restart loop mints its successor's generation from the **same** global counter:
   `binaries/daemon/src/spawn/prepared.rs:525` → `let new_generation = crate::running_dataflow::next_node_generation();`
3. In the replace handler the outgoing incarnation's restart is only disabled **after** the replacement is fully spawned:
   `binaries/daemon/src/lib.rs:2956` → `outgoing.disable_restart();` (spawn is at 2937-2941; commit/entry-removal at 2952-2953).
4. The daemon event loop is **blocked** on the inline build+spawn the whole time — `handle_coordinator_event` is awaited inline at `lib.rs:1656`, and the `ReplaceNode` arm awaits `task.await` (build, 2937) and `prepared.spawn().await` (2941) — so the outgoing node's exit, its restart's `ProcessHandleReplaced`, and the zombie's connection events all queue up behind the commit.

## Concrete sequence

Node `X` has `restart_policy: always` (or `on-failure`), current entry generation = 9.

1. `dora node replace X …` arrives. Handler calls `spawn_node`, minting `G_replace = 10` (global → 11), then blocks the event loop on the build + process spawn.
2. During that window `X`'s process exits (a clean exit for an `always` node, or a crash). `X`'s restart loop is a **separate task**, still enabled (`disable_restart` not yet reached). It mints `new_generation = 11` (> 10), stores it into `X`'s listener's shared `generation_counter`, and spawns a zombie process on the still-open **old** listener.
3. The zombie connects and its `Subscribe{11}` / `SendOut{11}` / `OutputsDone{11}` events are stamped generation 11 and queued (event loop still blocked).
4. Build finishes; replace commits: old entry (gen 9) removed, replacement inserted at **gen 10**, handler returns.
5. Event loop drains the queue:
   - Zombie's `ProcessHandleReplaced{prev=9,new=11}` → entry gen is 10 ≠ 9 → `RejectedStale` → drop-kills the zombie. Good.
   - But the zombie's already-queued `Event::Node{generation=11}` hit the gate: `event_generation_is_stale(10, 11) = (11 < 10) = false` → **not stale → applied to the replacement's entry.**

## Impact

- `Subscribe{11}` reinstalls `subscribe_channels[X]` pointing at the (dying) zombie. Landing after the replacement's own `Subscribe{10}`, the replacement receives **no inputs** and silently hangs.
- `OutputsDone{11}`/`CloseOutputs{11}` runs `handle_outputs_done` for `X`, propagating output-closed downstream — consumers stop reading the **replacement's** outputs.
- `SendOut{11}` injects spurious/duplicate output under `X`'s id.

The **exit** path is already safe because it uses exact `matches_generation` (`running_dataflow.rs:125`), so the zombie's terminal `SpawnedNodeResult{11}` is correctly dropped. That asymmetry is the tell: terminal events are exact-match (sound against a higher-generation orphan), but control/data `Event::Node`s are strictly-older (unsound against one).

## Suggested fixes (either closes it)

- **Targeted:** disable the outgoing incarnation's restart loop at the **top** of the replace handler, before `spawn_node` mints the replacement generation and before the build — re-enabling it only if a pre-commit step fails (to preserve the "outgoing left running on failure" guarantee).
- **Robust:** stop trusting "any newer generation is my successor." Track the specific successor generation the entry's *own* restart loop announced (e.g. a `pending_successor_generation` on `RunningNode`, set when the loop advances the counter and cleared when `ProcessHandleReplaced` lands), and accept an `Event::Node` only when its generation equals the entry generation *or* that announced successor — dropping every other newer generation.

Note the inline build+spawn on the event loop (`lib.rs:2937-2941`) both blocks the whole daemon for the build duration and widens this race window.

phil-opp:
Fixes #2997

## Problem

The `Event::Node` staleness gate in the daemon event loop dropped **only strictly-older** generations:

```rust
fn event_generation_is_stale(entry_generation: u64, event_generation: u64) -> bool {
    event_generation < entry_generation
}
```

This is deliberately permissive so a fast restart successor's early `Subscribe` — published to the listener *before* the daemon processes the matching `ProcessHandleReplaced` — is not lost.

But `dora node replace` breaks the underlying assumption that *"any newer-than-entry generation is the entry's own racing restart successor"*. A replacement is a brand-new incarnation whose generation is minted **early** (before its slow build), while a concurrent restart of the *outgoing* node mints a **higher** generation from the same global counter. The outgoing incarnation's restart-zombie then carries a generation *above* the replacement's, so its `Subscribe` / `SendOut` / `OutputsDone` events pass the `<` gate and are applied to the replacement:

- `Subscribe` reinstalls the subscribe channel pointing at the dying zombie → the replacement receives no inputs and silently hangs.
- `OutputsDone` / `CloseOutputs` propagates output-closed downstream → consumers stop reading the replacement's outputs.
- `SendOut` injects spurious/duplicate output under the node id.

The terminal-exit path was already safe because it uses exact `matches_generation`; that asymmetry (terminal = exact-match, control/data = strictly-older) was the tell.

## Fix

Track each entry's **own lineage** instead of trusting "newer = mine". `RunningNode` now carries a clone of the shared `generation_counter` that its *own* restart loop publishes the successor generation into (before spawning the successor, so the value is visible by the time the successor connects). The gate accepts an event only when its generation equals:

- the entry's current generation, **or**
- the successor the entry's own restart loop announced (`generation_counter`).

A stranger's higher generation — minted into a **different** lineage's counter (the outgoing node's, not the replacement's) — is now rejected, while the real successor's early events still pass. This also brings the control/data path in line with the exact-match terminal path.

Because generations are globally monotonic and the replacement's counter only advances via its own restart loop, a replace-race zombie's generation can never coincide with the replacement's announced-successor value, so the gate rejects it reliably.

## Tests

- Rewrote the gate unit test (`node_event_gate_accepts_only_own_lineage`) to cover all four cases: current generation accepted, older dropped, a fast **announced** successor accepted, and a **stranger's** higher generation dropped (the #2997 case).

## Validation

- `cargo build -p dora-daemon` ✓
- `cargo test -p dora-daemon` ✓ (214 passed)
- `cargo clippy -p dora-daemon -- -D warnings` ✓

---

_This PR was created by an automated Claude Code session working through auto-filed issues._

---
_Generated by [Claude Code](https://claude.ai/code/session_012FqbGAM7vQCCMYUuSTCfVP)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -465,14 +465,37 @@ type DaemonRunResult = BTreeMap<Uuid, BTreeMap<NodeId, Result<(), NodeError>>>;
 
 /// Whether a node-connection event belongs to a superseded incarnation.
 ///
-/// Strictly-older comparison, deliberately not equality: generations are
-/// globally monotonic, and the restart loop publishes a successor's
-/// generation to the listener BEFORE spawning it, so a fast successor's
-/// events can carry a generation NEWER than the entry until its
-/// `ProcessHandleReplaced` is processed. Dropping those would eat the
-/// successor's one-shot `Subscribe` (dora-rs/dora#2988 review, finding 1).
-fn event_generation_is_stale(entry_generation: u64, event_generation: u64) -> bool {
-    event_generation < entry_generation
+/// An event is fresh only when its generation belongs to this entry's own
+/// incarnation lineage: either the currently-registered incarnation
+/// (`entry_generation`) or the successor this entry's OWN restart loop
+/// announced before spawning it (`successor_generation`, published into the
+/// entry's shared `generation_counter` before the successor connects).
+///
+/// The successor exception is why this is not a plain `== entry_generation`
+/// check: the restart loop publishes the successor's generation to the
+/// listener and spawns it BEFORE the daemon processes the matching
+/// `ProcessHandleReplaced` (which advances `entry_generation`), so a fast
+/// successor's `Subscribe`/`SendOut` can arrive while the entry still holds
+/// the predecessor's generation. Dropping those would eat the successor's
+/// one-shot `Subscribe` (dora-rs/dora#2988 review, finding 1).
+///
+/// The reason it is not a plain `event < entry` check either: that wrongly
+/// accepts a *stranger's* higher generation. When `dora node replace` races a
+/// restart of the outgoing node, the outgoing incarnation's restart loop mints
+/// a generation ABOVE the replacement's (both draw from the same global
+/// counter, but the replacement's is minted first, before its slow build). Its
+/// zombie's events would then pass a `<` gate and be applied to the
+/// replacement — reinstalling a dead subscribe channel, closing the
+/// replacement's outputs, injecting spurious output (dora-rs/dora#2997).
+/// Matching on the entry's OWN announced successor rejects that stranger — the
+/// stranger's generation lives in the outgoing lineage's counter, never the
+/// replacement's — while still letting the real successor through.
+fn event_generation_is_stale(
+    entry_generation: u64,
+    successor_generation: u64,
+    event_generation: u64,
+) -> bool {
+    event_generation != entry_generation && event_generation != successor_generation
 }
 
 /// Patch a stored descriptor entry with a replacement node's definition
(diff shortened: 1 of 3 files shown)


phil-opp:

No issues found. I traced the mechanism the fix relies on and it holds together:

- Generations are globally unique (`NEXT_NODE_GENERATION`, `running_dataflow.rs:190`), and each spawn mints its own `generation_counter` `Arc<AtomicU64>` (`spawner.rs:480`) that only *this* lineage's restart loop advances (`prepared.rs:525-527`, `Release` store before the successor spawns). So the new gate `event == entry.generation || event == generation_counter` accepts exactly this lineage's current incarnation plus its announced successor, and a stranger's higher generation — living in a different lineage's counter — can never equal either value. That correctly rejects the #2997 replace-vs-restart zombie while preserving the #2988 fast-successor `Subscribe` case.
- `replace_process_handle` (`running_dataflow.rs:135`) updates the entry in place, keeping the same shared `generation_counter` Arc, so `entry.generation` and the announced-successor value stay coherent across restarts.
- The `Release`/`Acquire` pairing between the restart-loop store and the gate load is correct.
- The rewritten unit test asserts all four cases with distinct, non-tautological assertions.

One minor behavioral note, not a defect: in a rapid double-restart where the daemon hasn't yet processed the first `ProcessHandleReplaced`, the counter advances past an intermediate successor, so that intermediate incarnation's late in-flight events would now be dropped (the old `<` gate let them through). That incarnation is already superseded by the newer one, which re-Subscribes under its own generation, so no permanent disconnection results.

---

PR #3187, opened 2026-08-16
fix(cli): don't fail `dora logs --follow` on an already-finished dataflow

phil-opp:
## Issue

`dora logs <uuid> --follow` prints the historical logs correctly and then exits **non-zero** when the target dataflow has already finished.

The `logs()` command runs in two steps:
1. Fetch and print historical logs via a `Logs` request — this succeeds for finished/archived dataflows.
2. If `--follow` was passed, open a live subscription via `stream_logs_from_coordinator` → `session.subscribe_logs(...)`.

For a dataflow that has already finished, the coordinator has dropped it from its running-dataflows table, so `LogSubscribe` is rejected with `"no running dataflow with id X"` (`binaries/coordinator/src/ws_control.rs:165`). In `stream_logs_from_coordinator` that error was propagated with `?`, so the command emitted the full history and then failed.

This is exactly the sub-second/finished race that the CLI already recognizes elsewhere: `common.rs::error_indicates_dataflow_finished` exists for it, and `start/mod.rs::wait_until_dataflow_started` deliberately tolerates it. The `--follow` path did not. It is reachable for `dora logs <finished-uuid> --follow`, `--node <n> --follow`, and `--all-nodes --follow`, because `resolve_logs_dataflow_identifier` accepts a UUID without requiring the dataflow to still be running.

## Fix

In `stream_logs_from_coordinator`, match on the result of `subscribe_logs`. When the error indicates a finished dataflow, return `Ok(())` — a clean end of stream after the history has already been printed — mirroring `wait_until_dataflow_started`. Any other subscription error is still surfaced (now with a clearer `failed to subscribe to logs` context).

```rust
let log_rx = match session.subscribe_logs(&request) {
    Ok(rx) => rx,
    Err(err) if error_indicates_dataflow_finished(&err.to_string()) => return Ok(()),
    Err(err) => return Err(err).wrap_err("failed to subscribe to logs"),
};
```

## Validation

- `cargo fmt -p dora-cli -- --check` — clean
- `cargo test -p dora-cli` — 320 passed, 0 failed
- Manual: start a dataflow that exits immediately, then `dora logs <uuid> --follow`. Before: prints logs, exits non-zero. After: prints logs, exits 0.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01JnVKJTwF81rQJ6eRbbX5QX)_

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -8,8 +8,8 @@ use super::{Executable, default_tracing};
 use crate::common::parse_duration;
 use crate::{
     common::{
-        CoordinatorOptions, expect_reply, resolve_dataflow_identifier_interactive,
-        send_control_request,
+        CoordinatorOptions, error_indicates_dataflow_finished, expect_reply,
+        resolve_dataflow_identifier_interactive, send_control_request,
     },
     output::{
         LogFormat, LogOutputConfig, level_filter_is_active, message_passes_level_filter,
@@ -787,13 +787,21 @@ fn stream_logs_from_coordinator(
         until.and_then(|d| chrono::TimeDelta::from_std(d).ok().map(|td| now - td));
     let want_node = node.map(|n| n.as_ref());
 
-    let log_rx = session.subscribe_logs(
-        &serde_json::to_vec(&ControlRequest::LogSubscribe {
-            dataflow_id: uuid,
-            level: log_level,
-        })
-        .wrap_err("failed to serialize message")?,
-    )?;
+    let request = serde_json::to_vec(&ControlRequest::LogSubscribe {
+        dataflow_id: uuid,
+        level: log_level,
+    })
+    .wrap_err("failed to serialize message")?;
+    let log_rx = match session.subscribe_logs(&request) {
+        Ok(rx) => rx,
+        // The dataflow finished (and was dropped from the running table)
+        // before the live subscription was accepted. The historical logs
+        // have already been printed by `logs`, so this is a clean end of
+        // stream, not an error — mirror `wait_until_dataflow_started`, which
+        // tolerates the same race for sub-second dataflows.
+        Err(err) if error_indicates_dataflow_finished(&err.to_string()) => return Ok(()),
+        Err(err) => return Err(err).wrap_err("failed to subscribe to logs"),
+    };
 
     while let Ok(raw) = log_rx.recv() {
         let raw = match raw {


phil-opp:

This is the summary comment for a batch of machine-generated PRs from an in-depth review pass over the codebase (branched off `origin/main`). Each PR is small, self-contained, individually verified (fmt + clippy + affected-crate tests), and opened as a draft. The combined diff additionally passed the full workspace test suite and the `/review` + `/simplify` gates.

### PRs opened, by category

**Correctness**
- #3187 — `fix(cli)`: `dora logs --follow` on an already-finished dataflow printed the history and then exited non-zero; treat the finished-dataflow rejection of the live subscription as a clean end-of-stream (reusing the existing `error_indicates_dataflow_finished` helper).
- #3188 — `fix(cli)`: `dora doctor` counted a `Restarting` (crash-looping) node as *healthy* → `PASS`/exit 0, contradicting `dora top` (which renders `Restarting` red). Count it as degraded so a flapping node warns; mapping extracted into a testable pure helper.
- #3190 — `fix(download)`: `Content-Disposition` `filename` parsing stopped at the first raw `"`, truncating a value with an escaped quote (`filename="my \"weird\" name.bin"`). Decode the `\"` escape (only `\"`; literal backslashes / Windows paths are preserved) in both the parameter splitter and the value decoder so they agree on the value boundary.

**Optimization**
- #3189 — `perf(coordinator)`: the heartbeat watchdog awaited each 500 ms-bounded daemon send *serially* on the control loop, so N backpressured daemons stalled it for up to N×500 ms per tick. Send them concurrently with `join_all` (mirroring `destroy_daemons`), bounding the tick at ~500 ms.

**Simplification**
- #3192 — `fix(coordinator)`: the "log buffer full" warning re-fired for *every* message once a dataflow's (or build's) 10k buffer filled with no subscriber attached. Warn once on the filling message, then drop silently; both the dataflow-log and build-log branches share a new `buffer_log_message` helper.

**Documentation**
- #3191 — `docs(node)`: add a compile-checked doctest to the public `arrow_utils::ipc_encode::PreparedIpc` fast-path API (sizes from `byte_len()`, encodes, round-trips through `decode_arrow_ipc`), matching the existing `encode_arrow_ipc`/`decode_arrow_ipc` examples.

### Review-driven refinements applied before finalizing
- `/review` caught that the log-buffer fix (#3192) was applied only to the dataflow-log branch — the identical build-log branch still flooded. Both are now fixed via the shared helper.
- `/review` caught that a full quoted-pair unescaper in #3190 would drop literal backslashes in common non-RFC `filename="C:\Users\file.bin"` headers. Narrowed to only decode `\"`, preserving all other backslashes; added a regression test.

- **Byte-size parser duplicated** across `dora-core` (`parse_byte_size` → `u64`) and `dora-message` (`ByteSize::from_str` → `usize`): real duplication, but a safe shared implementation must reconcile the `u64` vs `usize` range, and the naive delegation changes the 32-bit overflow boundary in a 1.0-stable crate — deferred.
- **`dora_read_data` silently discards non-`UInt8` payloads** (and consumes the FFI array): real robustness gap, but the correct fix is a breaking FFI signature change (typed error return) — deferred.
- **Per-message `DataId` clone on the zenoh send fast path**: marginal — one short-string allocation, and the daemon request needs an owned copy anyway, so a clean removal only trades it for awkward error-context handling.
- **`enable_debug_inspection` forces a per-message serialize + Zenoh publish**: plausible redundant work, but I could not rule out an intended reason to keep the control-topic stream warm — needs author input rather than a speculative change.
- **Module param substitution can re-expand a value that embeds another param's placeholder**: low-confidence edge case (`{a: "${_param.b}", b: "hello"}`); only bites unusual descriptors.
- **Legacy `.txt` raw logs bypass sort/`--since`/`--grep`/`--tail` in local mode**: real, but the fix restructures `read_log_file`'s output to flow through the filter pipeline — larger change deferred.
- **`log_with_fields` per-call `to_lowercase()` allocation**: trivial; dominated by the surrounding `json!` + `to_rfc3339` + `println!`.
- **Undocumented internal `pub fn`s in the daemon** (`node_communication`, `socket_stream_utils`): low value — binary-crate internals, not user-facing API.

---

*All PRs in this batch are machine-generated by Claude (Claude Code) and marked draft. They have been verified locally as described in each PR, but please review carefully before merging.*

---

phil-opp:

No issues found. The race fix is correct and reuses existing infrastructure:
- The historical logs are already fetched and printed before `subscribe_logs` is reached, so treating a finished-dataflow rejection of the *live* subscription as a clean end-of-stream (`return Ok(())`) loses nothing.
- It gates on the existing `error_indicates_dataflow_finished` helper (`binaries/cli/src/common.rs:39-41`, matching `"no running dataflow with id"` / `"unknown dataflow"`), matching how `wait_until_dataflow_started` tolerates the same sub-second race. Every other subscription error is still surfaced via `wrap_err("failed to subscribe to logs")`.

---

PR #3189, opened 2026-08-16
perf(coordinator): send watchdog heartbeats concurrently, not serially

phil-opp:
## Issue

On every `DaemonHeartbeatInterval` tick, the coordinator's watchdog iterated the connected daemons and awaited a 500 ms-bounded `send_heartbeat_message` for each one **sequentially**, on the single control-plane event loop:

```rust
for (machine_id, connection) in daemon_connections.iter_mut() {
    ...
    let result = tokio::time::timeout(
        Duration::from_millis(500),
        send_heartbeat_message(connection, clock.new_timestamp()),
    ).await ...;   // <-- awaited before the next daemon is even tried
}
```

`send_heartbeat_message` → `DaemonConnection::send` awaits on a bounded `mpsc::Sender`. A daemon whose WS write side is backpressured (slow or half-dead, but not yet 30 s-stale, so not yet disconnected) consumes up to the full 500 ms timeout before the loop moves on. With N such daemons, a single tick stalls the event loop for up to **N × 500 ms**, during which every spawn/stop/logs/topic control request is delayed. The stall scales with deployment size and with the number of momentarily-slow daemons.

## Fix

Send the per-daemon heartbeats **concurrently** with `join_all`, so the whole tick is bounded at ~500 ms regardless of daemon count. This mirrors the existing `destroy_daemons` helper, which already fans out per-daemon sends this way.

- The classification is unchanged: `> 15 s` still warns, `> 30 s` still marks the daemon disconnected and skips its heartbeat.
- The send-failure bookkeeping is unchanged: a failed/timed-out heartbeat still adds the daemon to `disconnected`, driving the same downstream cleanup.
- Only the sends now overlap instead of running back-to-back. Each connection has its own mpsc, so overlapping the awaits is safe — that is exactly the wait being parallelized.

The per-daemon send + result tagging is factored into a small `send_heartbeat_with_timeout(machine_id, connection, timestamp)` helper so the futures collected for `join_all` share one type.

## Validation

- `cargo test -p dora-coordinator` — 151 passed, 0 failed

The disconnect/cleanup semantics are exercised by the existing coordinator test suite; this change only alters the scheduling of the sends within a tick, not their outcomes.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01JnVKJTwF81rQJ6eRbbX5QX)_

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2135,30 +2135,36 @@ async fn start_inner(
                     expire_stopped_nodes(dataflow);
                 }
                 let mut disconnected = BTreeSet::new();
+                // Send the per-daemon heartbeats concurrently rather than
+                // awaiting each 500 ms-bounded send in turn. Each send awaits a
+                // bounded WS mpsc, so a single backpressured (slow/half-dead but
+                // not yet 30 s-stale) daemon would otherwise stall this control
+                // loop for up to 500 ms before the next daemon is even tried —
+                // and N such daemons stall it for N × 500 ms, delaying every
+                // spawn/stop/logs request. `join_all` bounds the whole tick at
+                // ~500 ms regardless of daemon count. Mirrors `destroy_daemons`.
+                let mut heartbeats = Vec::new();
                 for (machine_id, connection) in daemon_connections.iter_mut() {
-                    if connection.last_heartbeat.elapsed() > Duration::from_secs(15) {
+                    let elapsed = connection.last_heartbeat.elapsed();
+                    if elapsed > Duration::from_secs(15) {
                         tracing::warn!(
-                            "no heartbeat message from machine `{machine_id}` since {:?}",
-                            connection.last_heartbeat.elapsed()
+                            "no heartbeat message from machine `{machine_id}` since {elapsed:?}"
                         )
                     }
-                    if connection.last_heartbeat.elapsed() > Duration::from_secs(30) {
+                    if elapsed > Duration::from_secs(30) {
                         disconnected.insert(machine_id.clone());
                         continue;
                     }
-                    let result: eyre::Result<()> = tokio::time::timeout(
-                        Duration::from_millis(500),
-                        send_heartbeat_message(connection, clock.new_timestamp()),
-                    )
-                    .await
-                    .wrap_err("timeout")
-                    .and_then(|r| r)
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. The concurrency refactor is sound:
- The per-daemon `&mut DaemonConnection` borrows collected into `heartbeats` come from `daemon_connections.iter_mut()` and are disjoint, so holding them across `join_all(...).await` is well-formed; `join_all` is already imported (`binaries/coordinator/src/lib.rs:33`) and this mirrors the existing `start_requests`/`stop_requests` patterns.
- Classification is preserved exactly: the >15 s warn and >30 s `continue`-into-`disconnected` branches still run serially in the build loop using a single cached `elapsed`, and only the sends now overlap. `send_heartbeat_with_timeout` tags each result with its `DaemonId` so the post-await failure bookkeeping into `disconnected` is unchanged.
- Because the futures aren't spawned they still run on the control-loop task, but cooperatively — a single backpressured daemon no longer serializes the others, bounding the tick at ~500 ms as claimed.

---

phil-opp:
The merge-queue conflict here was not with `main` (this branch merged cleanly against it) but with sibling PR #3192 — both inserted a new module-level helper immediately before `handle_spawn_result_ok` in `binaries/coordinator/src/lib.rs`, so the two diffs overlapped in the queue's batch.

Resolved by rebasing onto the latest `main` and **relocating** `send_heartbeat_with_timeout` to sit before `handle_spawn_result_err` instead — a different anchor that #3192 (whose helper stays before `handle_spawn_result_ok`) does not touch, so the two can no longer collide regardless of merge order. No behavior change; pure code move.

Re-verified on the rebased branch: `cargo fmt`/`clippy -D warnings` clean, `cargo test -p dora-coordinator` green (153 passed). Ready to re-queue.

---

phil-opp:
Since my last review the branch was rebased and `send_heartbeat_with_timeout` was relocated to sit before `handle_spawn_result_err` (to avoid the queue conflict with #3192). This is a pure code move with no behavior change — the helper body and the `join_all`-based concurrent-send loop are unchanged. The disjoint `&mut DaemonConnection` borrows collected from `iter_mut()` are held across `join_all(...).await` correctly, and the >15s-warn / >30s-disconnect classification plus the failure bookkeeping into `disconnected` are preserved. The relocated diff still looks safe. No issues found.

---

---

PR #3190, opened 2026-08-16
fix(download): honor an escaped quote in Content-Disposition filenames

phil-opp:
## Issue

`parse_content_disposition_filename` extracted the quoted `filename` value as everything up to the first **raw** `"`:

```rust
let name = if let Some(after_quote) = value.strip_prefix('"') {
    after_quote.split('"').next().unwrap_or(after_quote)  // stops at first `"`
} else { value };
```

RFC 6266 / RFC 2616 `quoted-string` values allow an escaped quote (`\"`) inside the value. So a header such as

```
Content-Disposition: attachment; filename="my \"weird\" name.bin"
```

was truncated at the first inner `\"` to `my \`, instead of decoding to `my "weird" name.bin`. This is a wrong-filename bug (not a security issue — the result still passes through the existing `file_name()` traversal sanitizer), triggered only by the uncommon server that emits escaped quotes, hence low severity.

## Fix

Decode the quoted value with a small `unquote_disposition_value` helper that walks the body, collapses a `\"` escape to a literal `"`, and stops at the first *unescaped* `"`.

**Only `\"` is treated as an escape.** A backslash that is not followed by a quote is kept verbatim, because real-world (non-RFC-compliant) servers routinely send unescaped Windows paths such as `filename="C:\Users\file.bin"`. Decoding every backslash as a quoted-pair (as a naive full-RFC unescaper would) would silently corrupt those names — so this deliberately handles only the escaped-quote case and preserves all other backslashes. `split_disposition_params` is left unchanged: the value decoder already terminates at the correct closing quote on its own.

The value is still fed through the existing `file_name()` / length / null-byte sanitizer in `get_filename`, so the traversal and length guards are unchanged.

## Validation

- `cargo fmt -p dora-download -- --check` — clean
- `cargo clippy -p dora-download --all-targets -- -D warnings` — clean
- `cargo test -p dora-download` — 16 passed, 0 failed

New regression tests cover: an escaped quote inside a quoted filename, an escaped quote before a trailing parameter, and unescaped backslashes kept verbatim (Windows path). The existing `semicolon_inside_quotes_is_preserved` and empty/missing cases still pass.

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -28,24 +28,70 @@ fn parse_content_disposition_filename(header: &str) -> Option<String> {
         }
         let value = value.trim();
         let name = if let Some(after_quote) = value.strip_prefix('"') {
-            // Quoted form: the value runs up to the closing quote.
-            after_quote.split('"').next().unwrap_or(after_quote)
+            // Quoted form: the value runs up to the first quote that is not a
+            // `\"` escape.
+            unquote_disposition_value(after_quote)
         } else {
-            value
+            value.to_string()
         };
-        (!name.is_empty()).then(|| name.to_string())
+        (!name.is_empty()).then_some(name)
     })
+}
+
+/// Decode a `Content-Disposition` quoted-string body (the text after the
+/// opening `"`): return everything up to the first *unescaped* closing quote,
+/// with a `\"` escape collapsed to a literal `"`. Without honoring the escape,
+/// a header like `filename="my \"weird\" name.bin"` is truncated at the first
+/// inner `\"` to `my \` instead of `my "weird" name.bin`.
+///
+/// Only `\"` is treated as an escape. A backslash that is *not* followed by a
+/// quote is kept verbatim, because real-world (non-RFC-compliant) servers
+/// routinely send unescaped Windows paths such as
+/// `filename="C:\dir\file.bin"`, and decoding those backslashes as quoted-pairs
+/// would silently corrupt the name. If no closing quote is present the whole
+/// remainder is returned (best-effort).
+fn unquote_disposition_value(body: &str) -> String {
+    let mut out = String::with_capacity(body.len());
+    let mut chars = body.chars().peekable();
+    while let Some(c) = chars.next() {
+        match c {
+            // `\"` -> literal `"`. Any other backslash stays literal.
+            '\\' if chars.peek() == Some(&'"') => {
+                out.push('"');
+                chars.next();
+            }
+            '"' => break,
+            _ => out.push(c),
+        }
+    }
+    out
 }
 
 /// Split a `Content-Disposition` header value into its `;`-separated
 /// parameters, treating a `;` inside a double-quoted value as literal so that
 /// `filename="a;b.bin"` stays a single parameter.
+///
+/// An escaped quote (`\"`) does not open or close a quoted value, so the split
+/// agrees with [`unquote_disposition_value`] on where a value ends. Without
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. I traced the `Content-Disposition` decoding change end-to-end:
- `unquote_disposition_value` (`libraries/extensions/download/src/lib.rs:53-67`) collapses only `\"` to `"`, stops at the first unescaped `"`, and keeps every other backslash verbatim — so unescaped Windows paths (`filename="C:\Users\file.bin"`) survive intact, as the added test asserts.
- `split_disposition_params` now carries an `escaped` flag that only arms on `\` while `in_quotes`, so a `\"` no longer flips the quote state. This keeps the splitter and the value decoder in agreement on where a value ends; the `escaped_quote_in_earlier_param_does_not_swallow_filename` test confirms a `\"` in a preceding parameter no longer merges the real `filename=` into it.
- The four regression tests are genuine (distinct inputs, non-tautological expectations). Output still flows through the existing `file_name()` sanitizer, so traversal/length guards are unchanged.

---

phil-opp:

Fixes the stated bug and the accompanying splitter desync — the `title="a \"b"; filename=...` case genuinely returns the wrong value on main. No panic path, no unbounded read (`with_capacity(body.len())`, header-bounded), and no traversal, since `get_filename` still applies `Path::file_name` plus the null/255-byte checks afterwards.

One non-blocking inconsistency: the splitter and the unquoter implement *different* escape policies. `split_disposition_params` (`libraries/extensions/download/src/lib.rs:~78`, the `'\\' if in_quotes => escaped = true` arm) treats `\x` as an escape for any `x`, while `unquote_disposition_value` only treats `\"` as one.

They disagree on a doubled backslash: for `filename="a\\"` the splitter closes the quote at the final `"` but the unquoter consumes it as an escape, yielding `a\"` instead of `a\`. The worse case is a Windows path ending in a backslash (`filename="C:\temp\"`), where the splitter leaves `in_quotes` true for the remainder of the header and a trailing `; size=…` gets merged into the filename — main returns `C:\temp\` there today. Both are cosmetically wrong filenames rather than a security issue, and both are exotic, so this is follow-up material.

phil-opp:
Thanks — good catch on the split/unquote escape-policy mismatch. Pushed a fix (force-update on this branch): `split_disposition_params` now uses the **same** policy as `unquote_disposition_value` — *only* `\"` is an escape, every other backslash is literal. The splitter peeks the next char and consumes only a `\"`, so the two stages can no longer disagree on where a quoted value ends (your `filename="a\\"` desync case now resolves the same way in both).

The genuinely ambiguous input — a value ending in a backslash immediately before the closing quote (`filename="C:\temp\"`) — is inherently undecidable without `\\` escaping, so it stays imperfect; but the two stages now agree on that interpretation rather than diverging. Verified with `cargo fmt`/`clippy -D warnings`/`test -p dora-download` (17 passed), and the common cases (unescaped Windows paths, escaped quote in an earlier parameter) keep their regression tests.

---

phil-opp:
The force-push resolves the split/unquote escape-policy mismatch I raised. `split_disposition_params` now peeks the next char and consumes a backslash only when it precedes a `"` (`'\\' if in_quotes && chars.peek().is_some_and(|&(_, c)| c == '"')`), identical to `unquote_disposition_value`'s "only `\"` is an escape" rule — so the two stages can no longer disagree on where a quoted value ends. The remaining trailing-backslash-before-quote case (`filename="C:\temp\"`) is genuinely undecidable without `\\` escaping, but both stages now interpret it the same way rather than diverging, which was the concern. No further issues found.

---

---

PR #3192, opened 2026-08-16
fix(coordinator): warn once when a log buffer fills, not per message

phil-opp:
## Issue

When a running dataflow (or an in-progress build) has no attached log subscriber, the coordinator buffers its log messages up to `MAX_BUFFERED_LOG_MESSAGES` (10,000):

```rust
if buffered_log_messages.len() < MAX_BUFFERED_LOG_MESSAGES {
    buffered_log_messages.push(message);
} else if buffered_log_messages.len() == MAX_BUFFERED_LOG_MESSAGES {
    tracing::warn!("log buffer full for … dropping new messages");
}
```

Once the buffer is full, `len` stays pinned at `MAX_BUFFERED_LOG_MESSAGES` (nothing is pushed), so the `else if` branch runs for **every subsequent message**, re-emitting the "log buffer full … dropping" warning each time. A chatty node/build with nobody watching its logs floods the coordinator log with one warning per dropped message. This was present in **both** the dataflow-log and build-log branches of the `Event::Log` handler.

## Fix

Factor the push-or-drop logic into a `buffer_log_message` helper that returns `true` exactly once — on the message that fills the buffer — so the caller emits a single "buffer full" warning and then drops subsequent messages silently:

```rust
fn buffer_log_message(buffer: &mut Vec<LogMessage>, message: LogMessage) -> bool {
    if buffer.len() < MAX_BUFFERED_LOG_MESSAGES {
        let now_full = buffer.len() == MAX_BUFFERED_LOG_MESSAGES - 1;
        buffer.push(message);
        now_full
    } else {
        false
    }
}
```

Both the dataflow-log and build-log branches now share it, so the buffering/warn policy lives in one place. The `dataflow_id`/`build_id` are matched by value (both are `Copy`), leaving `message` free to move into the helper.

## Validation

- `cargo test -p dora-coordinator` — 151 passed, 0 failed

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2292,32 +2292,31 @@ async fn start_inner(
                 }
             }
             Event::Log(message) => {
-                const MAX_BUFFERED_LOG_MESSAGES: usize = 10_000;
-                if let Some(dataflow_id) = &message.dataflow_id {
-                    if let Some(dataflow) = running_dataflows.get_mut(dataflow_id) {
+                // `dataflow_id`/`build_id` are `Copy`, so match them by value to
+                // leave `message` free to move into `buffer_log_message`.
+                if let Some(dataflow_id) = message.dataflow_id {
+                    if let Some(dataflow) = running_dataflows.get_mut(&dataflow_id) {
                         if dataflow.log_subscribers.is_empty() {
-                            if dataflow.buffered_log_messages.len() < MAX_BUFFERED_LOG_MESSAGES {
-                                dataflow.buffered_log_messages.push(message);
-                            } else if dataflow.buffered_log_messages.len()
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the `buffer_log_message` extraction (warn-once-on-full).
- Off-by-one is correct: `now_full` is `true` only when `buffer.len() == MAX - 1` immediately before the push that reaches `MAX`, so the warning fires exactly once; subsequent messages hit the `else` and drop silently. This fixes the original `else if len == MAX` branch that re-warned on every dropped message (the buffer never grows past `MAX`, so that condition stayed true forever).
- Buffer still caps at `MAX_BUFFERED_LOG_MESSAGES` and keeps the oldest messages. Matching `dataflow_id`/`build_id` by value (both `Copy`) to free `message` for the move is correct.

No issues found.

---

PR #3198, opened 2026-08-17
perf(message): serialize MemoryPoolWrite tensor_data as bulk bytes

### LINKED ISSUE #3195: Cross-machine MemoryPoolWrite tensor payload is serialized byte-by-byte (bypasses the #3156 bulk-bytes fast path)
## Summary

**PR #3156** ("Serialize bulk payloads as bytes, not byte-by-byte") routed `DataMessage::Vec` and `InterDaemonEvent::Output.data` through `serialize_bytes`/`deserialize_bytes` because serde's default byte-container impl emits one `serialize_u8` per element, which postcard turns into a per-byte `try_push` — 40-50x slower on payload-sized buffers.

Three days later **PR #3079** (commit c1b225e) added a new large-payload wire field that reintroduces exactly that slow path: `InterDaemonEvent::MemoryPoolWrite.tensor_data` is a plain `Vec<u8>`, not a bulk-bytes field.

## Affected code

```rust
// libraries/message/src/daemon_to_daemon.rs
Output {
    …
    #[serde(with = "crate::bulk_bytes::option")]        // fast path (#3156)
    data: Option<AVec<u8, ConstAlign<128>>>,
},
MemoryPoolWrite {
    …
    tensor_data: Vec<u8>,                               // ← plain Vec<u8>, byte-by-byte
    size: usize,
    seq: u64,
},
```

This event is serialized with postcard in `publish_memory_pool_event`, and the author already measured the cost — the function's own doc comment states:

```rust
// binaries/daemon/src/lib.rs ~1516-1518
/// declare and put timing — postcard::serialize of the 61.44MB
/// WriteMemoryPool payload takes hundreds of ms (3s+ in debug builds),
```

For a 1 MB payload #3156 measured encode dropping from ~932 µs to ~21 µs (and decode similarly) once the bulk-bytes path was used; the same ~40x applies here on both encode and decode of every relayed frame.

## Scope / severity

This is the **zenoh-relay fallback** path, not the steady state: the primary cross-machine data plane is the dedicated direct-TCP channel (`send_cross_data_frame`), which writes `&tensor_data` raw and is unaffected. The relay is used when no direct endpoint is known or the direct send fails — i.e. exactly on degraded/flaky links, where multi-MB tensors then pay a 40x serialize/deserialize tax on top of an already-degraded link. The node→daemon `DaemonRequest::WritePinnedMemory.tensor_data` (also plain `Vec<u8>`) is empty on the segment-backed path, so it is not a concern in the common case.

So: a real, self-documented performance regression on the fallback path, contradicting the pattern #3156 established — but not a correctness bug.

## Suggested fix

Give `MemoryPoolWrite.tensor_data` the same `serialize_bytes`/`deserialize_bytes` treatment as `Output.data`. The existing `bulk_bytes` module is typed for `AVec<u8, ConstAlign<128>>`; either switch `tensor_data` to that type (which also gets the 128-byte Arrow alignment the receiver's zero-copy decode wants) or add a plain-`Vec<u8>` variant of the helper / use `serde_bytes`. `encode_size_hint` already accounts for `tensor_data.len()`, so only the per-element serialization needs changing.

---

phil-opp:
## Summary

Fixes #3195.

`InterDaemonEvent::MemoryPoolWrite::tensor_data` is a plain `Vec<u8>`, so serde's default byte-container impl emits one `serialize_u8` per element, which postcard turns into a per-byte `try_push` — 40–50x slower than a bulk copy on payload-sized buffers. This is the **zenoh-relay fallback** path for the cross-machine tensor transport added in #3079 (commit `c1b225e`); it is used exactly on degraded/flaky links, where multi-MB tensors then pay that serialize/deserialize tax on top of an already-degraded link.

PR #3156 already routed `DataMessage::Vec` and `InterDaemonEvent::Output.data` through `serialize_bytes`/`deserialize_bytes` for the same reason. This PR applies the same treatment to the field #3079 reintroduced on the slow path.

## Changes

- Add a plain-`Vec<u8>` submodule (`crate::bulk_bytes::vec`) to the existing `bulk_bytes` module. The existing helper is typed for `AVec<u8, ConstAlign<128>>`; the mirror-segment receiver copies the bytes into its segment rather than decoding them zero-copy, so it does not need the 128-byte Arrow alignment. The visitor accepts both `visit_bytes` and `visit_seq` so self-describing formats (JSON) keep working.
- Annotate `MemoryPoolWrite::tensor_data` with `#[serde(with = "crate::bulk_bytes::vec")]`. The field type is unchanged (`Vec<u8>`), so no daemon producer/consumer code changes.

## Wire compatibility

The encoding is **byte-identical** to serde's default (postcard: varint length prefix + raw bytes), so this is a pure speedup — no `Metadata::CURRENT_VERSION` bump, no `.drec`/store schema change. A byte-identity test pins that across postcard's 127/128 varint length-prefix boundary, alongside postcard and JSON round-trip tests.

## Testing

- `cargo test -p dora-message bulk_bytes` — 7 passing, including the 3 new tests (`vec_encoding_is_unchanged_from_the_seq_form`, `vec_round_trips`, `vec_survives_a_json_round_trip`).
- `cargo clippy -p dora-message -- -D warnings` — clean.
- `cargo check -p dora-daemon` — clean (consumer type unchanged).

---

### DIFF
--- a/libraries/message/src/bulk_bytes.rs
+++ b/libraries/message/src/bulk_bytes.rs
@@ -127,6 +127,62 @@ pub mod option {
     }
 }
 
+/// The same bulk-bytes treatment for a plain `Vec<u8>` field.
+///
+/// Used by cross-machine payloads that do not need the 128-byte Arrow
+/// alignment (`InterDaemonEvent::MemoryPoolWrite::tensor_data`), where the
+/// receiver copies the bytes into a mirror segment rather than decoding them
+/// zero-copy. The wire encoding is identical to serde's default `Vec<u8>`
+/// (postcard: varint length + raw bytes), so this is a pure speedup with no
+/// version bump — `vec_encoding_is_unchanged_from_the_seq_form` pins it.
+pub mod vec {
+    use serde::{
+        Deserializer, Serializer,
+        de::{self, Visitor},
+    };
+    use std::fmt;
+
+    pub fn serialize<S: Serializer>(value: &[u8], serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_bytes(value)
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<Vec<u8>, D::Error> {
+        deserializer.deserialize_bytes(VecPayloadVisitor)
+    }
+
+    struct VecPayloadVisitor;
+
+    impl<'de> Visitor<'de> for VecPayloadVisitor {
+        type Value = Vec<u8>;
+
+        fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            f.write_str("a byte array")
+        }
+
+        fn visit_bytes<E: de::Error>(self, v: &[u8]) -> Result<Self::Value, E> {
+            Ok(v.to_vec())
+        }
+
+        fn visit_borrowed_bytes<E: de::Error>(self, v: &'de [u8]) -> Result<Self::Value, E> {
+            Ok(v.to_vec())
+        }
+
+        fn visit_byte_buf<E: de::Error>(self, v: Vec<u8>) -> Result<Self::Value, E> {
+            Ok(v)
+        }
+
+        /// Self-describing formats (notably JSON) render bytes as a sequence of
+        /// numbers and hand that back here rather than to `visit_bytes`.
+        fn visit_seq<A: de::SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
+            let mut out = Vec::with_capacity(seq.size_hint().unwrap_or(0));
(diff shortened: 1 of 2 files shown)


phil-opp:

Reviewed the `bulk_bytes::vec` serde helper applied to `MemoryPoolWrite::tensor_data`.
- `serialize_bytes` / `deserialize_bytes` with a visitor covering `visit_bytes`, `visit_borrowed_bytes`, `visit_byte_buf`, and `visit_seq` is correct and mirrors the existing `option` helper.
- Wire-compatibility claim is backed by a real byte-identity test across postcard's 127/128 varint boundary, plus postcard and JSON round-trips, including the `visit_seq` (JSON) path — so the "no version bump" assertion is actually pinned.

No issues found.

---

PR #3199, opened 2026-08-17
fix(daemon): key the cross-write safety-net timeout to the effective seq

### LINKED ISSUE #3193: Cross-machine memory-pool write can hang the node forever after a direct-TCP → zenoh failover (safety-net timeout keyed to the wrong seq)
## Summary

In the cross-machine tensor-transport feature added by **PR #3079** (commit c1b225e), the ack-timeout "safety net" that is meant to unblock a node's `WritePinnedMemory` when a remote commit ack never arrives is keyed to the **pre-failover** write sequence. When the direct-TCP data plane fails and the write is re-keyed to the zenoh relay, the safety net can no longer see the pending entry — so a lost relay ack leaves the node's write blocked **forever** and leaks the `CROSS_WRITE_PENDING` entry for the daemon's lifetime.

## Affected code

`binaries/daemon/src/lib.rs`, in the `DaemonNodeEvent::WriteMemoryPool` handler:

- The pending reply is inserted under the original `seq` and `pending_key` captures that original seq:
  ```rust
  CROSS_WRITE_PENDING.lock()… .insert((dataflow_id, shared_memory_id.clone(), seq), reply_sender);   // ~6766
  let pending_key = (dataflow_id, shared_memory_id.clone(), seq);                                     // 6770
  ```
- On a direct-TCP send failure the entry is **re-keyed** to a fresh `relay_seq` (this part is deliberate and correct — it drops the stale `ok:false` via first-wins):
  ```rust
  let rekeyed = CROSS_WRITE_PENDING.lock()… .remove(&(dataflow_id, shared_memory_id.clone(), seq));   // 6872-6875
  if let Some(tx) = rekeyed {
      …relay_seq = *counter;…
      CROSS_WRITE_PENDING.lock()… .insert((dataflow_id, shared_memory_id.clone(), relay_seq), tx);    // 6885-6891
  }
  ```
- The relay-publish error path correctly removes `relay_seq` (6915-6923). **But** the safety-net task removes the original `pending_key`:
  ```rust
  tokio::spawn(async move {
      tokio::time::sleep(CROSS_WRITE_ACK_TIMEOUT).await;                 // 6928-6929
      if let Some(tx) = CROSS_WRITE_PENDING.lock()… .remove(&pending_key) {   // 6933  ← original seq
          let _ = tx.send(DaemonReply::Result(Err("…timed out…")));
      }
  });
  ```
- `finish_dataflow` drains the sibling maps (`CROSS_POOL_WRITE_LOCKS`, `CROSS_WRITE_SEQ`, `CROSS_DIRECT_DEGRADED`, `cross_data_endpoints`) but **not** `CROSS_WRITE_PENDING` (lib.rs ~8034-8082), so nothing else reclaims the orphaned `relay_seq` entry.

## Failure scenario

1. A cross-machine pool has a direct-TCP data endpoint (the steady-state fast path).
2. The direct send fails mid-frame (flaky WAN — the exact scenario this feature targets), so the handler re-keys the pending entry from `seq` to `relay_seq` and falls back to the zenoh relay.
3. The relay `MemoryPoolWrite` publish **succeeds** (so the 6915 error path does not fire).
4. The relay's `MemoryPoolWriteAck` never arrives (peer daemon restart or a dropped zenoh sample — precisely the case the safety net exists for, per the comment at 6926-6927).
5. The safety net fires after `CROSS_WRITE_ACK_TIMEOUT` and removes `pending_key` = `(df, pool, seq)` — already gone since step 2, so it's a no-op. The entry under `relay_seq` is never removed.

Result: the node's `WritePinnedMemory` (`reply.await` in `binaries/daemon/src/node_communication/mod.rs`) blocks forever, and `CROSS_WRITE_PENDING` leaks one entry per occurrence. The safety net is defeated in exactly the failover case it was written to protect.

## Suggested fix

Make the safety net follow the re-key. Options:
- Have the timeout task watch `relay_seq` instead of the captured original — e.g. move the timeout `spawn` into the write task after `relay_seq` is finalized, or pass the effective seq through a shared cell / re-arm the timeout on re-key.
- Or add a `CROSS_WRITE_PENDING` drain to `finish_dataflow` (keyed by dataflow id) as a backstop, so at least the leak and hang are bounded to the dataflow's lifetime rather than the daemon's.

The first fully fixes the hang; the second bounds the blast radius and matches how the sibling maps are already handled.

---

phil-opp:
## Summary

Fixes #3193.

The cross-machine memory-pool write (added in #3079, commit `c1b225e`) withholds a node's `WritePinnedMemory` reply until the mirror daemon's `MemoryPoolWriteAck` arrives, with a 120s safety-net timeout that fails the write if the ack never comes. On a direct-TCP → zenoh failover the pending entry is re-keyed from the original `seq` to a fresh `relay_seq` (deliberate — it drops the stale `ok:false` via first-wins), but the timeout task captured the **original** `seq`.

After a re-key, the timeout's `remove(&pending_key)` targets a key that is already gone — a silent no-op — so a lost relay ack (peer restart, dropped zenoh sample) leaves the node's write blocked **forever** and leaks the `CROSS_WRITE_PENDING` entry for the daemon's lifetime. The safety net is defeated in exactly the failover case it was written to protect.

## Failure scenario

1. A cross-machine pool has a direct-TCP data endpoint (steady state).
2. The direct send fails mid-frame (flaky WAN); the handler re-keys the pending entry from `seq` to `relay_seq` and falls back to the zenoh relay.
3. The relay `MemoryPoolWrite` publish **succeeds** (so the publish-error path doesn't fire).
4. The relay's `MemoryPoolWriteAck` never arrives.
5. The safety net fires and removes `(df, pool, seq)` — already gone since step 2 — so it's a no-op. The `relay_seq` entry is never removed → hang + leak.

## Fix

- Share the write's **effective seq** between the write task and the timeout task via an `AtomicU64`. The write task advances it to `relay_seq` when it re-keys; the timeout task reads it *after* sleeping and removes the entry under whatever seq is actually live. The removal is extracted into `fire_cross_write_timeout`.
- Drain `CROSS_WRITE_PENDING` in `finish_dataflow` (via `drain_cross_write_pending`), matching how the sibling cross-write maps (`CROSS_WRITE_SEQ`, `CROSS_POOL_WRITE_LOCKS`, `CROSS_DIRECT_DEGRADED`, `cross_data_endpoints`) are already reclaimed there. A dataflow that finishes with a write still pending (node crash, `dora stop`, or a relay entry orphaned by failover) now unblocks the node immediately instead of waiting out the 120s timeout, and cannot leak the entry for the daemon's lifetime.

The first fully fixes the hang; the second bounds the blast radius and closes the `finish_dataflow` gap the issue calls out.

## Testing

Adds two regression tests to `cross_write_failover_tests`:
- `safety_net_timeout_follows_the_relay_rekey` — after a re-key, timing out on the stale seq is a no-op (leaves the write pending); timing out on the effective seq fails the node's write cleanly.
- `finish_dataflow_drain_fails_pending_cross_writes` — the drain fails a finishing dataflow's pending write while leaving an unrelated dataflow's untouched.

```
cargo test -p dora-daemon cross_write   # 4 passed
cargo clippy -p dora-daemon -- -D warnings   # clean
cargo fmt --all -- --check   # clean
```

A full multi-daemon e2e reproduction (real coordinator + two daemons + a mid-frame direct disconnect) is out of scope for this unit-level fix; the tests exercise the seq-keying logic directly against the shared pending map.

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -1735,26 +1735,36 @@ fn resolve_cross_write_ack(
 }
 
 /// Fire the safety-net timeout for a pending cross-machine write: remove the
-/// entry under `seq` and fail the node's blocked `WritePinnedMemory`. Returns
-/// whether an entry existed.
+/// entry under the write's *effective* seq and fail the node's blocked
+/// `WritePinnedMemory`. Returns whether an entry existed.
 ///
-/// The caller passes the write's *effective* seq — the one the entry is live
-/// under now, which advances from the original `seq` to a fresh `relay_seq` on
-/// a direct-TCP → zenoh failover. Keying the timeout to the pre-failover seq
-/// makes it a no-op after a re-key, leaving the write hung forever (#3193).
-fn fire_cross_write_timeout(dataflow_id: Uuid, shared_memory_id: String, seq: u64) -> bool {
-    if let Some(tx) = CROSS_WRITE_PENDING
+/// `effective_seq` is the seq the entry is live under now, which advances from
+/// the original `seq` to a fresh `relay_seq` on a direct-TCP → zenoh failover.
+/// Keying the timeout to the pre-failover seq makes it a no-op after a re-key,
+/// leaving the write hung forever (#3193).
+///
+/// The load happens *under the pending-map guard*, the same one
+/// `rekey_cross_write_to_relay` holds while it re-keys and stores. Loading
+/// outside it would reopen #3193 as a race: a failover landing between the load
+/// and the `remove` would leave the removal targeting a key that has already
+/// moved, and nothing retries.
+fn fire_cross_write_timeout(
+    dataflow_id: Uuid,
+    shared_memory_id: &str,
+    effective_seq: &std::sync::atomic::AtomicU64,
+) -> bool {
+    let mut pending = CROSS_WRITE_PENDING
         .lock()
-        .unwrap_or_else(|e| e.into_inner())
-        .remove(&(dataflow_id, shared_memory_id, seq))
-    {
-        let _ = tx.send(DaemonReply::Result(Err(
-            "cross-machine write timed out waiting for the remote commit ack".to_string(),
-        )));
-        true
-    } else {
-        false
-    }
+        .unwrap_or_else(|e| e.into_inner());
+    let seq = effective_seq.load(std::sync::atomic::Ordering::SeqCst);
+    let Some(tx) = pending.remove(&(dataflow_id, shared_memory_id.to_string(), seq)) else {
+        return false;
+    };
+    drop(pending);
+    let _ = tx.send(DaemonReply::Result(Err(
+        "cross-machine write timed out waiting for the remote commit ack".to_string(),
+    )));
+    true
 }
 
 /// Re-key a pending cross-machine write from `seq` to a fresh relay seq for the
@@ -1773,10 +1783,18 @@ fn rekey_cross_write_to_relay(
     seq: u64,
     effective_seq: &std::sync::atomic::AtomicU64,
 ) -> Option<u64> {
-    let tx = CROSS_WRITE_PENDING
+    // Remove, re-key and store under a *single* pending-map guard. Releasing
+    // it between the remove and the insert would leave the entry in no map at
+    // all, so a concurrent `drain_cross_write_pending` walks past it and the
+    // re-insert then outlives `finish_dataflow` — the node stays blocked until
+    // the 120s safety net and the entry leaks past the dataflow it belongs to.
+    // The counter bump nests `CROSS_WRITE_SEQ` inside this guard; that is the
+    // only nesting of the two (every other site drops the seq lock before
+    // touching the pending map), so the order cannot deadlock.
+    let mut pending = CROSS_WRITE_PENDING
         .lock()
-        .unwrap_or_else(|e| e.into_inner())
-        .remove(&(dataflow_id, shared_memory_id.to_string(), seq))?;
+        .unwrap_or_else(|e| e.into_inner());
+    let tx = pending.remove(&(dataflow_id, shared_memory_id.to_string(), seq))?;
     let relay_seq = {
         let mut seqs = CROSS_WRITE_SEQ.lock().unwrap_or_else(|e| e.into_inner());
         let counter = seqs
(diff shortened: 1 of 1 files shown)


phil-opp:

Reviewed the safety-net timeout re-key fix and the finish-time drain.
- Sharing the effective seq via `AtomicU64` correctly follows the direct-TCP → zenoh re-key: the write task stores `relay_seq` only inside the successful `if let Some(tx)` re-key branch (`lib.rs:6876-6892`), and the timeout reads it after the 120s sleep, so it removes whatever seq is actually live. The old `pending_key` capture was indeed a silent no-op after a re-key.
- No double-send / double-free: entry removal is atomic under the map lock, and the oneshot is consumed by whichever of {relay resolver, timeout, drain} wins.
- Concurrent-drain safety: if `finish_dataflow`'s `drain_cross_write_pending` removes the entry before the failover arm re-keys, the `remove(seq)` in the re-key returns `None`, so no re-insert and no `store`; the timeout then loads the stale `seq` and no-ops. No hang, no double-send.
- Tests genuinely exercise the behavior (stale-seq no-op vs. effective-seq resolution; drain isolation across dataflows).

No blocking issues found.

---

phil-opp:

The bug and the fix are both real: the `pending_key` captured pre-failover was a guaranteed no-op after the re-key at `:6875`, and the `AtomicU64` handoff plus the `finish_dataflow` drain closes it. No lock is held across an await — `drain_cross_write_pending` collects under the lock and sends after releasing.

The problem is that **neither new test touches the fix**. `safety_net_timeout_follows_the_relay_rekey` hand-rolls an insert/remove/insert against `CROSS_WRITE_PENDING` and then asserts `fire_cross_write_timeout(df, pool, 1) == false` and `(df, pool, 2) == true` — that is a `HashMap::remove` tautology.

Concretely: delete the actual fix (`effective_seq.store(relay_seq, SeqCst)` at `:6958`) and both tests still pass, while the write still hangs forever. What needs pinning is the seq handoff itself — drive the atomic through the re-key and assert the timeout task reads `relay_seq`.

phil-opp:
You're right — the test was a `HashMap::remove` tautology that didn't touch the actual fix. Fixed in `5ec06e8`.

I extracted the direct-TCP → zenoh re-key (including the `effective_seq.store(relay_seq)`) into `rekey_cross_write_to_relay`, which the `WriteMemoryPool` handler now calls, and rewrote the test to drive that helper and assert the shared `effective_seq` the timeout task reads was advanced to the re-keyed seq.

Verified it now pins the fix by mutation: deleting the `store` inside the helper makes the test fail on

```
assertion `left == right` failed: re-key must advance the effective seq the timeout follows
```

and, with the atomic left at the stale seq, the subsequent `fire_cross_write_timeout(effective_seq.load())` no longer resolves the entry — i.e. the test fails exactly in the state where the write hangs forever. Restored, it passes. Behavior is unchanged: the helper runs the same remove / monotonic-counter bump / re-insert / atomic-store the handler did inline.

---

phil-opp:
The new commit `5ec06e8` addresses my earlier point that the test didn't touch the fix. The re-key is now factored into `rekey_cross_write_to_relay` (called by the `WriteMemoryPool` handler), and `rekey_advances_effective_seq_so_the_timeout_follows_it` drives that helper directly and asserts `effective_seq.load() == relay_seq` — so deleting the `effective_seq.store(relay_seq)` now fails the test rather than passing silently. It also confirms the timeout keyed to the stale seq is a no-op while the effective-seq timeout resolves the write. The extraction is behavior-preserving (same remove / counter-bump / re-insert / store). No further issues found.

---

---

PR #3200, opened 2026-08-17
fix(daemon): unlink cross-machine mirror segments on dataflow finish

### LINKED ISSUE #3194: Cross-machine mirror /dev/shm segments (up to 1 GiB) leak permanently on abnormal dataflow termination
## Summary

The cross-machine tensor-transport feature added by **PR #3079** (commit c1b225e) mirrors each remote pool into a local `/dev/shm` segment on the receiver daemon. That mirror segment is only ever unlinked in response to an explicit `FreePool` (originating from a node's `FreePinnedMemory`). Any cross-machine dataflow that ends **without** an explicit free reaching the mirror — node crash, dataflow abort, `dora stop` killing the node, origin-daemon crash, or a dropped `FreePool` publish — leaves the mirror segment resident **forever**. On a long-lived daemon (`dora up`) cycling many cross-machine dataflows, this grows unbounded until the host exhausts shared memory.

## Affected code

Three cleanup paths each fail to reclaim mirror segments:

1. **`finish_dataflow` does not unlink and does not publish `FreePool`.** It calls `cleanup_cross_pools`, which is documented as bookkeeping-only:
   ```rust
   // libraries/extensions/tensor-pool/src/lib.rs:365-368
   /// Unlinks are deliberately not done here: freeing a cross-machine pool
   /// publishes `FreePool` (the peer unlinks its side); this is only the local
   /// bookkeeping drain …
   pub fn cleanup_cross_pools(&self, dataflow_id: &str) {
       self.lock_cross_pools().retain(|(df, _), _| df != dataflow_id);
   }
   ```
   `finish_dataflow` (binaries/daemon/src/lib.rs ~8080-8082) calls only this — no `FreePool` is emitted, so the mirror machine never unlinks.

2. **The orphan sweep can't match the mirror name.** Mirror segments are named with `machine_id` **before** `dataflow_id`:
   ```rust
   // libraries/extensions/tensor-pool/src/lib.rs:575
   "dora_pool_{machine_id}_{dataflow_id}_{node_id}_{counter}"
   ```
   but `cleanup_orphans` strips the prefix `dora_pool_{dataflow_id}_`:
   ```rust
   // libraries/extensions/tensor-pool/src/lib.rs:423
   let prefix = format!("dora_pool_{}_", dataflow_id);
   ```
   The `machine_id` component sits before `dataflow_id`, so the prefix never matches a mirror segment. (Even if it did, the sweep filters by `is_local_node(node_id)`, and the mirror's owner is the *sender* node, which never runs on the mirror machine.)

3. **Daemon-shutdown `cleanup_all` only drains the main pool table**, which mirrors never enter.

The mirror segment is also created owner-less (`set_owner(false)`, see the comment at binaries/daemon/src/lib.rs ~1478), so nothing unlinks it implicitly on drop.

## Failure scenario

Run a cross-machine dataflow whose sender node is killed by `dora stop` (or crashes) rather than explicitly calling `free`. No `FreePinnedMemory` → no `FreePool` → the mirror segment (sized up to the 1 GiB registration cap) stays in the receiver machine's `/dev/shm` after the dataflow is gone. Repeat across many dataflows on a persistent daemon → unbounded `/dev/shm` growth until the host runs out of shared memory.

## Suggested fix

`finish_dataflow` already has the deterministic mirror segment name available (via `cross_pool_shmem_name`) and, by the time it runs, local readers are done — so it is safe to unlink the mirror there. Either:
- Unlink each of the dataflow's mirror segments in `finish_dataflow` before draining `cross_pools`, or
- Fix `cleanup_orphans` to also match the `dora_pool_{machine_id}_{dataflow_id}_…` shape (and reclaim mirror-owned segments host-safely), or
- Publish the targeted `FreePool` from the dataflow-finish path so the mirror side unlinks via the existing handler.

Related but distinct: #2881 (local pools on node crash, now closed) and #2935 (single-machine cross-process free skip). This one is specific to the remote **mirror** segment introduced by #3079.

---

phil-opp:
## Summary

Fixes #3194.

The cross-machine tensor transport (#3079, commit `c1b225e`) mirrors each remote pool into a local `/dev/shm` segment on the receiver daemon. That mirror segment was only ever unlinked in response to an explicit `FreePool` (originating from a node's `FreePinnedMemory`). Any cross-machine dataflow that ends **without** an explicit free reaching the mirror — node crash, dataflow abort, `dora stop` killing the node, origin-daemon crash, or a dropped `FreePool` publish — left the mirror segment (sized up to the 1 GiB registration cap) resident **forever**. On a long-lived daemon (`dora up`) cycling many cross-machine dataflows, this grows unbounded until the host exhausts shared memory.

The issue traced three cleanup paths that each miss the mirror:
1. `finish_dataflow` drained the sibling cross-pool bookkeeping (`cleanup_cross_pools`) but never unlinked and published no `FreePool`.
2. `cleanup_orphans` strips `dora_pool_{dataflow_id}_`, but the mirror name is `dora_pool_{machine_id}_{dataflow_id}_…` — the prefix never matches.
3. Daemon-shutdown `cleanup_all` only drains the main pool table, which mirrors never enter.

## Fix

Unlink this daemon's mirror segments in `finish_dataflow`, before draining the cross-pool table, using the same machine-qualified `cross_pool_shmem_name` + `remove_cross_pool_shmem` the `FreePool` handler already uses. This is the issue's first suggested fix direction.

- By the time `finish_dataflow` runs, the local receiver nodes are done, so the unlink is safe (same invariant the `FreePool` unlink relies on).
- The name is `machine_id`-qualified, so it only ever matches a segment **this** daemon created as the mirror. An origin-side `cross_pools` entry resolves to a name that does not exist locally — a harmless `NotFound` no-op (`remove_cross_pool_shmem` already swallows `NotFound`). That also keeps it host-safe when two daemons share a host: each has a distinct `machine_id`, so one daemon's finish can never unlink another's live mirror.

Adds `TensorPoolManager::cross_pool_ids` to enumerate a dataflow's tracked cross-machine pools.

## Testing

- New unit test `cross_pool_ids_scoped_to_dataflow_and_cleared_on_cleanup` — pins that the accessor is dataflow-scoped (never returns another dataflow's pools, whose mirrors are still live) and is cleared by `cleanup_cross_pools`.

```
cargo test -p dora-tensor-pool cross_pool   # 3 passed
cargo clippy -p dora-daemon -p dora-tensor-pool -- -D warnings   # clean
cargo fmt --all -- --check   # clean
```

The actual `remove_file` unlink is the same call the `FreePool` handler already exercises; this PR only changes *when* it runs. A full multi-machine reproduction (kill a sender mid-dataflow, assert the receiver's `/dev/shm` segment is gone) is out of scope for a unit-level change.

## Out of scope

The `cleanup_orphans` prefix mismatch (defect #2, which only bites when a mirror's creator dies between `shm_open` and registration, so there is no `cross_pools` entry) is left for a follow-up — this PR fixes the dominant node-crash / `dora stop` / abort path, where the entry does exist and `finish_dataflow` now reclaims it.

---

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -50,7 +50,7 @@ use std::{
     pin::pin,
     sync::{
         Arc,
-        atomic::{self, AtomicU32, AtomicU64},
+        atomic::{self, AtomicBool, AtomicU32, AtomicU64},
     },
     time::{Duration, Instant},
 };
@@ -1506,8 +1506,28 @@ fn remove_cross_pool_shmem(shmem_name: &str) {
     }
     #[cfg(not(target_os = "linux"))]
     {
-        tracing::warn!("memory pool: shmem removal only implemented on Linux");
+        // Not an anomaly worth a warning per pool per free/finish: the
+        // whole cross-machine mirror path is Linux-only.
+        tracing::debug!("memory pool: shmem removal only implemented on Linux");
     }
+}
+
+/// Unlink the `/dev/shm` segment a cross-machine pool resolves to under
+/// `machine_id`, returning whether the name could be derived.
+///
+/// Used by the free paths that only know the pool id. Paths that hold
+/// the name the segment was created under should unlink that instead —
+/// see `cross_pool_shmem_name`'s own warning: "create, write, and the
+/// free paths must agree or a mirror leaks under a name nobody unlinks."
+fn unlink_cross_pool_segment(machine_id: &str, dataflow_id: &str, shared_memory_id: &str) -> bool {
+    let Some(shmem_name) =
+        TensorPoolManager::cross_pool_shmem_name(machine_id, dataflow_id, shared_memory_id)
+    else {
+        tracing::warn!("memory pool: invalid pool id {shared_memory_id}, cannot unlink mirror");
+        return false;
+    };
+    remove_cross_pool_shmem(&shmem_name);
+    true
 }
 
 /// Publish an inter-daemon memory-pool event over the dataflow topic.
@@ -1604,15 +1624,9 @@ async fn release_cross_pool(
     shared_memory_id: &str,
     shm_provider: Option<&ShmProvider<PosixShmProviderBackend>>,
 ) {
-    let Some(shmem_name) = TensorPoolManager::cross_pool_shmem_name(
-        machine_id,
-        &dataflow_id.to_string(),
-        shared_memory_id,
-    ) else {
-        tracing::warn!("memory pool: invalid pool id {shared_memory_id}, cannot unlink mirror");
+    if !unlink_cross_pool_segment(machine_id, &dataflow_id.to_string(), shared_memory_id) {
         return;
-    };
-    remove_cross_pool_shmem(&shmem_name);
+    }
     tracing::info!("memory pool: forwarding free of {shared_memory_id} to peer {peer_machine_id}");
     if let Err(e) = publish_memory_pool_event(
         session,
@@ -1741,6 +1755,39 @@ fn resolve_cross_write_ack(
 /// on the plain path and never allocates from here.
 const MEMORY_POOL_SHM_PROVIDER_SIZE: usize = 8 * 1024 * 1024;
 
+/// A dataflow's memory-pool zenoh subscriber, plus the liveness flag the
+/// detached mirror-creation tasks it feeds check before they publish a
+/// mirror into the tensor pool.
+///
+/// Mirror creation runs off the event loop, so a `finish_dataflow` can
+/// land between the handler's liveness check and the task's
+/// `register_cross_pool`. Without the flag that task would register a
+/// segment *behind* the finish-time drain, leaking it for the daemon's
+/// lifetime (dora-rs/dora#3194).
+pub(crate) struct MemoryPoolSubscriber {
+    /// The receive loop has no shutdown branch of its own, so a dropped
+    /// handle would leak the task, its session clone, and its event
+    /// sender for the daemon's lifetime.
+    handle: tokio::task::JoinHandle<()>,
+    /// Cloned into every mirror task; cleared by [`Self::shutdown`].
+    dataflow_live: Arc<AtomicBool>,
+}
+
+impl MemoryPoolSubscriber {
+    /// Stop feeding this dataflow: clear the liveness flag, then abort
+    /// the receive loop.
+    ///
+    /// The flag is cleared *before* the caller drains the cross-pool
+    /// table, which is what makes the drain authoritative: a mirror task
+    /// that still reads `true` after its `register_cross_pool` is
+    /// guaranteed to be visible to that drain, and one that reads `false`
+    /// rolls its own registration back.
+    fn shutdown(self) {
+        self.dataflow_live.store(false, atomic::Ordering::SeqCst);
+        self.handle.abort();
+    }
+}
+
 pub struct Daemon {
     /// This machine's id (as registered with the coordinator), if any.
     /// Used to gate which daemon mirrors a cross-machine pool and to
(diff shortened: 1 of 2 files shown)


phil-opp:

Reviewed the mirror-segment unlink on dataflow finish. Logic is sound:
- `cross_pool_ids` is dataflow-scoped and the finish path unlinks each pool's segment before draining the table, closing the leak in #3194 for the crash / `dora stop` / dropped-`FreePool` cases.
- The unlink reuses the same `machine_id`-qualified `cross_pool_shmem_name` + `remove_cross_pool_shmem` as the existing `FreePool` handler (`lib.rs:1240`, `1483`), and `remove_cross_pool_shmem` is a NotFound-tolerant no-op, so origin-side entries and the two-daemons-per-host case are safe as the comment claims.
- Idempotent vs. a prior `FreePool` unlink (already-removed name is a no-op), and ordering (unlink before `cleanup_cross_pools`) is correct.
- The added `cross_pool_ids` test meaningfully pins scoping and the post-cleanup empty result.

No blocking issues found.

---

phil-opp:

The safety premise does not hold — this can unlink a **live local** segment.

The loop iterates every `cross_pools` entry, but the origin daemon registers those too (`binaries/daemon/src/lib.rs:7140`). The name it then computes for an origin-side entry is not a nonexistent one: `TensorPoolManager::cross_pool_shmem_name` (`libraries/extensions/tensor-pool/src/lib.rs:574`) produces `dora_pool_{machine}_{df}_{node}_{counter}`, which is byte-for-byte the format a node uses for its **own local** pool segment when `DORA_MACHINE_ID` is set (`libraries/extensions/tensor-pool/python/src/transport.rs:1457-1460`). Pool ids are `pool_{node}_{counter}`, so the strip-and-rejoin round-trips exactly.

Failure scenario — a same-host two-daemon cluster, which the cluster-e2e nightly exercises: daemon A hosts the sender, daemon B the receiver, and B's mirror carries `sender_shmem` so B reads A's segment directly. A's nodes finish first → A's `finish_dataflow` unlinks `dora_pool_A_{df}_sender_0`, its own live segment, while B's receiver is still draining. B's next `try_doradma_read` reopen gets ENOENT.

Main's `FreePool` handler gates exactly this with `machine_id != self.machine_id → return` (`:5418`) so only the mirror side ever unlinks; this PR drops that gate. The unlink needs to be restricted to entries where this daemon actually created a mirror — `cross_peer` can't distinguish, since both sides store "the other machine", so it likely needs a mirror flag on the entry.

Separately, `.filter(|m| !m.is_empty())` diverges from every other site, which uses `unwrap_or_default()`. A daemon with `machine_id: None` still mirrors pools (the `RegisterPool` gate at `:5248` compares against `unwrap_or("")`) and still unlinks them on `FreePool`, but is skipped here — so the leak this PR targets survives untouched in exactly that configuration.

phil-opp:
Good catch — both points were correct. Fixed in `ee9f0f8`.

**Live-segment unlink.** Verified the claim: with `DORA_MACHINE_ID` set, a node names its own local pool segment `dora_pool_{machine}_{df}_{node}_{counter}` (`transport.rs:1450-1461`), byte-for-byte what `cross_pool_shmem_name` produces, and the origin daemon does register a `cross_pools` entry (`lib.rs:7153`). So iterating all entries and unlinking would have destroyed the sender's live local segment on a same-host cluster — exactly as described.

The unlink now runs only for entries this daemon created as a **mirror**. Since both sides store "the other machine" and can't be told apart by peer id, I added an `is_mirror` flag to the cross-pool entry: `true` at the mirror-creation site (`lib.rs:5383`), `false` on the origin register-ack path (`lib.rs:7159`). `finish_dataflow` unlinks only `cross_pool_mirror_ids(df)`, so origin-side entries (whose name is the live local segment) are never touched — the same invariant the FreePool handler's `machine_id` gate enforces.

**`machine_id` divergence.** Also fixed: the finish path now uses `self.machine_id.as_deref().unwrap_or_default()`, matching the FreePool handler (`:5520`) and the RegisterPool gate (`:5248`), so a daemon with `machine_id: None` still reclaims its mirrors rather than skipping them.

New test `cross_pool_mirror_ids_returns_only_mirror_entries_scoped_to_dataflow` asserts origin entries are excluded. (Two pre-existing `cross_pool_write_tests` fail in my sandbox because a second peering `zenoh::open` can't bind here, but they pass in CI — unrelated to this change.)

---

phil-opp:
The new commit `ee9f0f8` addresses both points from my earlier review, and the fix looks correct:

- The `is_mirror` flag is set `false` on the origin register-ack path (`lib.rs:~7159`) and `true` only at the mirror-creation site (`lib.rs:~5383`), and `finish_dataflow` now unlinks only `cross_pool_mirror_ids(df)`. Origin-side entries — whose machine-qualified name is the sender node's own live local segment — are no longer touched, so the same-host two-daemon race I described is closed.
- The finish path now uses `self.machine_id.as_deref().unwrap_or_default()`, matching the `FreePool` handler and the `RegisterPool` gate, so a `machine_id: None` daemon still reclaims its mirrors.

The new `cross_pool_mirror_ids_returns_only_mirror_entries_scoped_to_dataflow` test genuinely pins that origin (`is_mirror == false`) entries are excluded and that the result stays dataflow-scoped. No further issues found.

---

---

PR #3202, opened 2026-08-17
fix(ci): repair the two nightly regressions from the 2026-08-13 batch (#3150)

phil-opp:
Fixes the two nightly regressions tracked in #3150 (both started with the 2026-08-13 batch, both reproduce deterministically on Linux).

**`examples` (all 3 platforms)** — `rust-dataflow-git` pins `rev: c9a3699` (2026-07-20), which predates #3153's bincode→postcard switch on the node↔daemon wire. The pinned nodes fail at registration with `Serde Deserialization Error` / `server disconnected unexpectedly`. That is the break this example is designed to catch, so the fix is the pin bump — to `c1b225ea` (current main).

**`cross-check` (linux-gnu + both darwin targets)** — #3152 added `dora-tensor-pool-python`, which links pyo3 but was not added to the job's python-crate exclude list. The job builds with `PYO3_NO_PYTHON=1`, so pyo3-ffi's build script fails with *"attempted to link to Python shared library but config does not contain lib_name"*. Not a macOS runner problem — it reproduces on `x86_64-unknown-linux-gnu`.

The local driver `scripts/qa/ci-nightly-jobs.sh cross-check` did not set `PYO3_NO_PYTHON`, so it found the host interpreter and passed where CI could not; it now mirrors the GHA job env (per-command, so it doesn't leak into later jobs).

Verified locally: `cargo run --example rust-dataflow-git` runs to completion and self-terminates in ~2s; `scripts/qa/ci-nightly-jobs.sh cross-check` passes with the new exclude and fails without it.

The remaining `ROS2 Zenoh Interop (Humble)` failure on the same issue predates this batch (failing since at least 2026-08-11) and is untouched here.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1792,6 +1792,7 @@ jobs:
           --exclude dora-ros2-bridge-python
           --exclude dora-runtime-python
           --exclude dora-cli-api-python
+          --exclude dora-tensor-pool-python
       - name: Check (cross)
         if: matrix.cross
         run: >
@@ -1801,6 +1802,7 @@ jobs:
           --exclude dora-ros2-bridge-python
           --exclude dora-runtime-python
           --exclude dora-cli-api-python
+          --exclude dora-tensor-pool-python
 
   # ===== ROS2 bridge basic checks =====
   # Keep nightly coverage for ROS2 bridge code that does not require a full ROS
--- a/examples/rust-dataflow-git/dataflow.yml
+++ b/examples/rust-dataflow-git/dataflow.yml
@@ -4,15 +4,16 @@
 # change lands on main — otherwise the CI job catches the mismatch and
 # signals a compatibility break, which is the whole point of this test.
 #
-# Currently pinned past #2774, which added a `metadata_version` field to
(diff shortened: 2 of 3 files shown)


phil-opp:
**Automated review (Claude)**

Reviewed the nightly CI fix and the `rust-dataflow-git` pin bump — the diff/code, not the description.

- Adding `--exclude dora-tensor-pool-python` to the `Check` and `Check (cross)` steps in `nightly.yml` matches the existing pyo3-crate exclude pattern: that crate links pyo3 without `extension-module`, so it can't build under `PYO3_NO_PYTHON=1`.
- Prefixing each `cargo check` in `scripts/qa/ci-nightly-jobs.sh::job_cross_check` with a per-command `PYO3_NO_PYTHON=1` mirrors the GHA job-level env, stays scoped so it doesn't leak into later jobs, and makes the local driver reproduce the CI failure instead of silently passing via the host interpreter. All three OS arms are updated consistently.
- The example pin bump (`c9a3699` → `c1b225e`) moves `rust-dataflow-git` past the #3153 bincode→postcard wire break, which is what this example exists to exercise.

No issues found.

---
🤖 _This is a fully automated review by Claude — no human has verified these findings. Treat them as advisory._

---

PR #3203, opened 2026-08-17
fix(telemetry): set service.name to the node id on OTLP traces and metrics

DGHX12345:
## Problem
init_tracing accepts a `name` argument (documented as "Service name
for the traces") but ignores it, so every dora node exports with the
OTel SDK default `unknown_service:<binary>`. Backends (Jaeger, Tempo,
Grafana, Studio) cannot attribute spans or metrics per node —
per-node flame graphs and service filtering are unusable.

## Fix
Attach a service.name resource to both the tracer and meter providers.
dora nodes already pass their node id as the name
(TracingBuilder::new(node_id)).

## Test plan
- cargo test -p dora-tracing: 8 unit + 1 doc tests pass
  (added service_resource_sets_service_name)
- Manual: run a dataflow with DORA_OTLP_ENDPOINT pointed at an OTLP
  backend; spans now carry the node id as service.name

### DIFF
--- a/libraries/extensions/telemetry/tracing/src/lib.rs
+++ b/libraries/extensions/telemetry/tracing/src/lib.rs
@@ -181,7 +181,7 @@ impl TracingBuilder {
         // Initialize OTLP tracing - this returns a tracer and sets the global provider
         let sdk_tracer_provider = crate::telemetry::init_tracing(&self.name, &endpoint)
             .wrap_err("failed to initialize OTLP tracing exporter")?;
-        let meter_provider = metrics::init_meter_provider(&endpoint)
+        let meter_provider = metrics::init_meter_provider(&self.name, &endpoint)
             .wrap_err("failed to initialize OTLP metrics exporter")?;
 
         // TODO: Maybe this needs to be removed in favor of application level global.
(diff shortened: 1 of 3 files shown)


phil-opp:

I reviewed this change (attaching a `service.name` resource to the OTLP tracer and meter providers). No issues found:
- `init_tracing` previously accepted `_name` and discarded it, so nodes exported as the SDK default `unknown_service:<binary>`; it now builds a `service_resource(name)` and both `init_tracing` and `init_meter_provider` attach it, so spans and metrics are attributed per node id.
- The caller in `tracing/src/lib.rs` passes `&self.name` to `init_meter_provider`, matching the trace side.
- The added `service_resource_sets_service_name` test is meaningful — it reads back the `service.name` key and asserts the exact value, so it would fail if the resource were not set.

---

phil-opp:

Two notes, neither blocking. No start-up hazard: `Resource::builder().with_service_name(..).build()` is infallible, and `with_otlp_tracing` already bails before this point when `DORA_OTLP_ENDPOINT` is unset — so a misconfigured endpoint fails exactly as it does today.

**Semver, worth catching before 1.0.** `libraries/extensions/telemetry/tracing/src/metrics.rs:14` adds a parameter to `pub fn init_meter_provider`, which is a breaking signature change on a published crate while #3121 is preparing the 1.0.0 bump. Making it `pub(crate)`, or having it take the resource, avoids spending a major version on it.

**The test does not pin the wiring.** It asserts the helper builds the right `Resource`, but nothing asserts that `init_tracing` / `init_meter_provider` actually call it — deleting either `.with_resource(...)` line leaves the suite green.

PR #3204, opened 2026-08-17
fix(core): don't warn on timer/log inputs during manifest injection

phil-opp:
## Summary

`apply_manifest_contracts` (in `libraries/core/src/manifest/inject.rs`) checks every entry of `node.inputs` against a node manifest's declared ports and emits an `input `X` is not declared` warning for any that is missing. It iterated over **all** inputs, including built-in `dora/timer/*` and `dora/logs/*` subscriptions — which are dataflow-level wiring, not node contract ports, and are never listed in a `dora-node.yml` manifest.

As a result, any manifest-backed node that also subscribes to a timer or log stream — a very common pattern — produced a bogus warning:

```yaml
nodes:
  - id: detector
    path: yolo/target/release/dora-yolo   # has an adjacent dora-node.yml
    inputs:
      image: camera/image                 # declared data port
      tick: dora/timer/millis/100         # timer — NOT a manifest port
```

→ `node "detector": input `tick` is not declared in …/dora-node.yml`

Per `InjectionResult`'s own doc, these warnings are printed and **fatal under `strict_types`**. So a dataflow with `strict_types: true`, a manifest-backed node, and any timer/log input would **fail to build** on a non-existent contract violation.

## Fix

Only check `InputMapping::User` subscriptions against the manifest, skipping `Timer`/`Logs`. This mirrors how the analogous edge checker in `descriptor/validate.rs` already special-cases those variants, and matches the established `matches!(input.mapping, InputMapping::User(_))` idiom used across the daemon and core. Type injection itself is unaffected — it iterates `manifest.inputs`, not `node.inputs`.

## Validation

- Added a regression test (`timer_and_log_inputs_do_not_warn_as_undeclared`) with a node subscribing to both a timer and the log stream alongside a declared data port; asserts no warnings. Confirmed it fails before the fix (emits the two spurious warnings) and passes after.
- `cargo fmt -p dora-core -- --check`, `cargo clippy -p dora-core --all-targets -- -D warnings`, and the full `cargo test -p dora-core` suite (283 tests) pass locally on rustc 1.97.1.

---

⚠️ **This is a machine-generated pull request**, authored by Claude (Claude Code). It has been verified locally as described above but should receive human review before merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DY5kNaoUXqmLD3VYGLrSv4)_

### DIFF
--- a/libraries/core/src/manifest/inject.rs
+++ b/libraries/core/src/manifest/inject.rs
@@ -16,7 +16,7 @@ use std::path::{Path, PathBuf};
 
 use crate::descriptor::normalize_path;
 
-use dora_message::{descriptor::Descriptor, id::NodeId};
+use dora_message::{config::InputMapping, descriptor::Descriptor, id::NodeId};
 
 use super::{
     MANIFEST_FILENAME, NodeManifest,
@@ -261,8 +261,15 @@ pub fn apply_manifest_contracts(
         return;
     }
 
-    // the dataflow's wiring must stay within the declared ports (spec §10.1)
-    for input in node.inputs.keys() {
+    // the dataflow's wiring must stay within the declared ports (spec §10.1).
+    // Only user data-port subscriptions are governed by the manifest; built-in
+    // `dora/timer/*` and `dora/logs/*` subscriptions are dataflow-level wiring,
+    // not node contract ports, so they must not be flagged as "not declared"
+    // (which is fatal under `strict_types`).
+    for (input, def) in &node.inputs {
+        if !matches!(def.mapping, InputMapping::User(_)) {
+            continue;
+        }
         if !manifest.inputs.contains_key(input.as_str()) {
             result.warnings.push(sanitize(&format!(
                 "node \"{}\": input `{input}` is not declared in {}",
@@ -446,6 +453,38 @@ nodes:
         assert!(all.contains("input `depth` is not declared"), "{all}");
         assert!(all.contains("output `points` is not declared"), "{all}");
         assert!(all.contains("required input `image` is not wired"), "{all}");
+    }
+
+    #[test]
+    fn timer_and_log_inputs_do_not_warn_as_undeclared() {
+        // A node commonly subscribes to a built-in timer/log stream alongside
+        // its manifest data ports. Those subscriptions are dataflow wiring, not
+        // node contract ports, so they must not be reported as "not declared"
+        // (which is fatal under `strict_types`).
+        let tmp = tempfile::tempdir().unwrap();
+        write_manifest(&tmp.path().join("yolo"), MANIFEST);
+        let mut df = dataflow(
+            r#"
+nodes:
+  - id: camera
+    path: cam
+    outputs: [image]
+  - id: detector
+    path: yolo/target/release/dora-yolo
+    inputs:
+      image: camera/image
+      tick: dora/timer/millis/100
+      logs: dora/logs
+    outputs: [bbox]
+"#,
+        );
+        let mut registry = TypeRegistry::new();
+        let result = inject_adjacent_manifests(&mut df, tmp.path(), &mut registry);
+        assert_eq!(
+            result.warnings,
+            Vec::<String>::new(),
+            "timer/log subscriptions must not be flagged as undeclared inputs"
+        );
     }
 
     #[test]


phil-opp:
## Automated codebase review — run summary

This is the summary for an automated, machine-generated review pass over the repository (branched off `origin/main` @ `8ff6f3b`). Findings were each re-verified against current `main` before acting, and checked against the open PRs from prior runs to avoid duplicates.

### PRs opened this run

**Correctness**
- #3204 — `fix(core): don't warn on timer/log inputs during manifest injection`. `apply_manifest_contracts` flagged built-in `dora/timer/*` and `dora/logs/*` subscriptions as "input not declared" against the node manifest; those warnings are fatal under `strict_types`, so a manifest-backed node with a timer input could fail to build.
- #3205 — `fix(cli): reject a malformed DORA_COORDINATOR_ADDR/PORT in `dora up``. `dora up` was the only subcommand parsing these env vars by hand and silently swallowing a parse error, binding the default endpoint while every clap-parsed subcommand hard-errors on the same value.

No optimization / documentation / simplification PRs met the bar this run.

### Findings examined and rejected (with reasons)

- **mavlink `?baud=` invalid value silently defaults** — duplicate of existing open PR #3181.
- **tensor-pool `ipc_written` shadow** (`.../tensor-pool/python/src/transport.rs`) — a second `let mut ipc_written` shadows the first, so the same-host cross-machine GPU-IPC path reports `ipc_present = false` to the daemon while the shmem header byte says `1`. Real code/comment inconsistency, but a CUDA + cross-machine + same-host-only path whose user-visible impact can't be verified without GPU hardware; skipped as too delicate to change blind.
- **mavlink IPv6 UDP shutdown-wake socket** (`mavlink2-bridge-node/src/main.rs`) — the wake socket is hard-bound to IPv4 (`0.0.0.0:0`), so it can't wake an IPv6 reader. But IPv6 UDP isn't cleanly supported at the connect layer either (`udpin:{host}:{port}` is malformed for bracketed IPv6), and the wake failure is non-fatal (≤500 ms slower shutdown via the grace window). Low value.
- **mavlink outbound sequence-number gap during UDP no-peer retries** — the sequence counter advances for frames that are never transmitted; a one-time, self-correcting jump. Not worth a fix.
- **`DaemonReply::encode_size_hint` unit mismatch** (`daemon_to_node.rs:145`) — the `PinnedMemoryMetadata` arm returns `parameters.len()` (a count) where every other arm returns a byte count. Real bug, but the value only pre-sizes a buffer that `postcard` reallocs transparently — perf-only, negligible.
- **`dora record` / topic selector don't discover per-operator outputs** — a node with `operators: [{ outputs: [...] }]` yields no topics. The same blind spot exists uniformly across the debug surface, which suggests a deliberate scope limitation of the operator API rather than a regression.
- **runtime-api known limitations** — input-path `Stop` responsiveness, `dora_read_data` consuming the array before a fallible convert, and a stale `open_operator_inputs` entry after `Finished`. All already acknowledged in-code (#2742), not new defects.
- **core `find_manifest_for` first-match-wins shadowing** — a stray `dora-node.yml` closer to the binary can silently shadow the real manifest. Plausibly deliberate ("a manifest governs its subtree"); not clearly a bug.

### Validation

Each changed crate was verified in full individually — `dora-core` (fmt + clippy `--all-targets` + all 283 tests) and `dora-cli` (fmt + the CI-style clippy gate + the new unit tests). The combined diff passed the `/review` and `/simplify` gates and a workspace-wide `cargo check --all` (all crates compile). The full `cargo test --all` run could not complete in this environment: it exhausted the session's fixed disk allowance during compilation (ENOSPC), which is an infrastructure limit rather than a test failure — the changes are isolated to two crates with no public-API change, so downstream crates are unaffected.

---

phil-opp:

No issues found. `InputMapping` has exactly the `User`/`Timer`/`Logs` variants, and gating the manifest-declaration check on `InputMapping::User(_)` correctly excludes built-in `dora/timer/*` and `dora/logs` subscriptions, which are dataflow wiring and never appear in a node manifest. This matches the existing special-casing in `descriptor/validate.rs` (the `Timer`/`Logs` arm is a no-op there). The change touches only the "input is not declared" direction, so the separate required-input-wiring check is unaffected, and there's no regression for the degenerate case where a manifest port happens to share a name with a timer input (timers carry a fixed type, so skipping the manifest check for them is the correct behavior anyway). The regression test wires a timer, a log stream, and a declared data port together and asserts an empty warnings vector, so it genuinely exercises the changed branch rather than a mock.

---

PR #3209, opened 2026-08-18
fix(python-node): reject negative shape dims in tensor_from_info GPU guard

### LINKED ISSUE #3208: cuda.py: the new tensor_from_info GPU size guard is bypassed by a negative shape dimension
## Summary

#3079 (c1b225e) added a size-validation guard to `tensor_from_info` to stop a peer-controlled header from building an out-of-bounds GPU tensor. The guard computes `prod(shape) * itemsize` and rejects it when it exceeds the registered `size`. Because it uses a plain running product, a **negative** shape dimension makes the product negative, so the `> size` comparison is false and the guard is skipped — the exact class of malformed header the guard was written to reject.

## Where

`apis/python/node/dora/cuda.py:386-393` (added by #3079):

```python
expected_bytes = np.dtype(np_dtype).itemsize
for dim in shape:
    expected_bytes *= dim
if expected_bytes > size:
    raise ValueError(
        f"tensor shape {shape} * {np_dtype} itemsize = {expected_bytes} bytes "
        f"exceeds registered size {size} bytes — header may be corrupted"
    )
```

`shape` originates from the pool's header JSON and is carried as a `Vec<i64>` (`transport.rs` extracts it as signed ints, so negative values survive), i.e. it is peer/header-controlled — which is precisely the input the guard's own comment names as the threat ("a peer-controlled header that claims a large shape over a small buffer would produce an out-of-bounds GPU tensor").

- A large **positive** product is caught (Python ints are unbounded, no overflow).
- A **negative** dim (e.g. `[-1, 1<<40]` or a single `-1`) yields a negative `expected_bytes`, so `expected_bytes > size` is `False` and the check passes. Execution proceeds to `_CudaArrayInterface(ptr, shape, None, typestr)` → `torch.as_tensor(..., device="cuda")` with an unvalidated shape.

## Severity / confidence

Defense-in-depth. In practice `torch.as_tensor` over a `__cuda_array_interface__` with a negative dimension most likely raises rather than materializing an OOB view, so I have **not** confirmed an actual out-of-bounds read — but relying on torch's downstream behavior means the guard is not doing the job it was added to do, and torch's handling is not a contract. Reachable only on the cross-machine GPU read path (opt-in `DORA_MEMORY_POOL_CROSS_MACHINE`).

## Suggested fix

Reject non-positive dimensions before (or as part of) the size check, e.g.:

```python
if any(d < 0 for d in shape):
    raise ValueError(f"tensor shape {shape} has a negative dimension — header may be corrupted")
```

(A zero dimension is legitimately empty and would pass the size check as `0 <= size`; only negatives need the explicit reject.)

---

_This issue was filed by a **scheduled automated Claude code-review check** (reviewing merged commits from the last two days). Framed as defense-in-depth: the logic hole is verified, but a concrete OOB was not demonstrated — please treat accordingly._

phil-opp:
## Summary

Fixes #3208.

The GPU size guard added to `tensor_from_info` in #3079 (`apis/python/node/dora/cuda.py`) computes `prod(shape) * itemsize` with a plain running product and rejects the header when it exceeds the registered pool `size`. Because `shape` is peer/header-controlled and carried as signed ints, a **negative** dimension makes the running product negative, so `expected_bytes > size` evaluates to `False` and the guard is silently bypassed — exactly the malformed-header class the guard was written to reject.

## Change

Reject negative dimensions explicitly before the size check:

```python
if any(dim < 0 for dim in shape):
    raise ValueError(
        f"tensor shape {shape} has a negative dimension — header may be corrupted"
    )
```

A zero dimension is a legitimately empty tensor and still passes the size check as `0 <= size`, so only negatives are rejected. This matches the fix suggested in the issue.

## Scope / testing

Defense-in-depth, reachable only on the opt-in cross-machine GPU read path (`DORA_MEMORY_POOL_CROSS_MACHINE`). The change is a pure-Python guard that runs before `torch.as_tensor`. There is no existing Python unit-test harness for `cuda.py` (it requires torch + a CUDA device, which is why #3079 shipped the original guard without one), so no automated test is added here; `python3 -m py_compile` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SYaXAByMZaV36WJfHbfZ1N)_

### DIFF
--- a/apis/python/node/dora/cuda.py
+++ b/apis/python/node/dora/cuda.py
@@ -383,6 +383,17 @@ def tensor_from_info(tensor_info: dict) -> torch.Tensor:
         # claims a large shape over a small buffer would produce an
         # out-of-bounds GPU tensor (the CPU path is saved by numpy
         # reshape, but the GPU path has no equivalent backstop).
+        #
+        # Reject negative dimensions first: the size check below uses a
+        # plain running product, and a negative dim makes that product
+        # negative, so `expected_bytes > size` would be False and the
+        # guard would be silently bypassed — exactly the malformed
+        # (peer-controlled) header it exists to reject. A zero dim is a
+        # legitimate empty tensor and passes the size check as 0 <= size.
+        if any(dim < 0 for dim in shape):
+            raise ValueError(
+                f"tensor shape {shape} has a negative dimension — header may be corrupted"
+            )
         expected_bytes = np.dtype(np_dtype).itemsize
         for dim in shape:
             expected_bytes *= dim


phil-opp:

No issues found. The fix is correct: `expected_bytes` is a running product seeded with the dtype itemsize, so a single negative dimension makes the product negative and `expected_bytes > size` becomes False, silently bypassing the guard on the GPU path and constructing a `_CudaArrayInterface` with a negative/out-of-bounds extent. Rejecting negative dims before the size loop closes that hole, and a zero dim still passes as `0 <= size`, preserving legitimate empty tensors. The CPU branch is correctly left alone since its `reshape` provides an independent backstop.

Minor, non-blocking: no regression test was added for the negative-dim header case, so nothing guards against this guard being refactored away later.

---

phil-opp:
The merge-queue failure here is the **Unwrap budget** check (`unwrap/expect count: 162 (budget: 161)`), and it is not caused by this PR.

- This PR changes exactly one file — `apis/python/node/dora/cuda.py` (+11 lines, Python only) — so it adds **zero** Rust `.unwrap()`/`.expect(` and cannot move the unwrap budget.
- On current `origin/main` the count is **159** (budget 161) and `scripts/qa/unwrap-budget.sh` passes locally. Merging this Python-only change on top leaves it at 159.
- The failing run tested a **batched** merge branch (`trunk-merge/pr-3241`, "waiting for other pull requests to finish testing"); the `162` comes from a co-queued PR that added Rust unwraps, not from this change.

No change is warranted on this PR — it should pass once re-tested outside that batch. Happy to rebase if that helps the queue pick it up cleanly.

---

PR #3210, opened 2026-08-18
fix(tensor-pool): reopen explicit-name pool segment on write cache-miss

### LINKED ISSUE #3207: tensor-pool: writes to an explicit-`name=` pool are silently dropped after a PINNED_POOL cache-miss (write fast path reconstructs the auto name)
## Summary

`register_memory_pool(..., name=...)` (added in #3079 / c1b225e) lets a node register a pool under an explicit `/dev/shm` segment name, used **verbatim**. But the per-frame **write** fast path does not remember that name: on a `PINNED_POOL` cache-miss it reconstructs the **auto-generated** name (`dora_pool_{df}_{node}_{counter}`) instead. For an explicit-name pool that segment does not exist, so `ShmemConf::open()` fails and the write is silently dropped (or, if a same-counter auto segment happens to exist, writes land in the wrong segment).

This is a correctness defect on a **legitimate, non-adversarial** user path (the documented `name=` parameter) and is independent of the cross-machine threat model.

## Where

Introduced by **#3079** (commit `c1b225e`), the "register_memory_pool name param + machine-qualified auto names" change.

**Register** stores the explicit name verbatim as the segment name — `libraries/extensions/tensor-pool/python/src/transport.rs:1433-1463`:

```rust
let shmem_name = match &name {
    Some(name) => { /* validated, then */ name.clone() }
    None => { /* format!("dora_pool_{df}_{node}_{counter}") ... */ }
};
```

**Write** fast path, cache-miss branch — `transport.rs:2164-2180` — ignores the explicit name and rebuilds the auto name from the counter:

```rust
} else {
    // Cache miss: open via ShmemConf ...
    let shmem_name = {
        let machine = std::env::var("DORA_MACHINE_ID").unwrap_or_default();
        if machine.is_empty() {
            format!("dora_pool_{}_{}_{}", self.dataflow_id, self.node_id, counter)
        } else {
            format!("dora_pool_{}_{}_{}_{}", machine, self.dataflow_id, self.node_id, counter)
        }
    };
    match ShmemConf::new().os_id(&shmem_name).open() { ... }
```

## Why the cache-miss branch is actually reachable

The write cache-miss path is not dead code — it exists to recover from `PINNED_POOL` slot eviction. The `TRANSIT_META` static was added for exactly this reason (`transport.rs:184-191`):

```rust
/// Persistent transit-buffer metadata for GPU pools.
/// Survives `PINNED_POOL` cache-miss so the write fast path can recover
/// `transit_ptr` and `pool_device` even when the `PoolSlot` has been evicted.
```

So the code already treats "write happens while the `PoolSlot` has been evicted" as an expected scenario and installs a side-table to recover GPU transit metadata across it — but there is **no** analogous recovery for the explicit segment name, so the name silently falls back to the auto format.

Note the **read** side does not have this problem: it resolves the registered segment name from the daemon once per pool. Only the write fast path reconstructs the name locally, which is the asymmetry that produces the bug.

## Impact

- A pool registered with `name="dora_pool_myapp_frames"` works while its `PINNED_POOL` slot is warm, then starts dropping writes after the slot is evicted — a confusing, data-dependent failure.
- Consumers of that pool see the last stable frame indefinitely (no new data), with no error surfaced to the producer.

## Suggested fix

Persist the explicit name at registration in a counter-keyed side table (mirroring `TRANSIT_META`), and have the write cache-miss branch prefer it before falling back to the auto-name format. Equivalently, store the resolved `shmem_name` in the `PoolSlot`/a recovery map so both the warm and cold write paths open the same segment the register created. Consider asserting/logging when `ShmemConf::open()` fails on the write path rather than dropping the frame silently.

---

_This issue was filed by a **scheduled automated Claude code-review check** (reviewing merged commits from the last two days). The finding was verified against the source at `c1b225e`, but please confirm the exact `PINNED_POOL` eviction trigger and reproduce before treating it as high priority — the analysis establishes the code path is by-design reachable (via `TRANSIT_META`) but did not pin down every eviction scenario end to end._

phil-opp:
## Summary

Fixes #3207.

`register_memory_pool(..., name=...)` (added in #3079) lets a node register a pool under an explicit `/dev/shm` segment name, used verbatim. But the per-frame **write** fast path did not remember that name: on a `PINNED_POOL` cache-miss (slot eviction — an expected scenario, which is why `TRANSIT_META` already exists to recover GPU transit metadata across it) the write branch reconstructed the **auto-generated** name `dora_pool_{df}_{node}_{counter}` instead. For an explicit-name pool that segment does not exist, so `ShmemConf::open()` failed and the write was **silently dropped** — the producer sees success, the consumer sees the last stable frame forever.

The read side is unaffected because it resolves the registered segment name from the daemon; only the write fast path reconstructed the name locally, which is the asymmetry that produced the bug.

## Change

- Add a counter-keyed `POOL_NAMES` side table (mirroring `TRANSIT_META`), populated with the resolved segment name in `register_tensor_pool` and cleared in both `free_tensor_pool` (normal free) and `rollback_local_pool` (registration failure).
- The write cache-miss path now resolves the segment name via `write_pool_shmem_name`, which prefers the recorded name and only falls back to the auto-name format when none is on record.
- The cache-miss open now emits a `warn!` when the segment is missing instead of silently dropping the frame — a missing segment on that path is a real routing fault.
- Name construction is factored into two pure helpers — `auto_pool_shmem_name` and `write_pool_shmem_name` — shared by the registration and write paths so they cannot drift apart. The cache-miss open body is extracted into `open_pool_segment_on_miss`.

## Testing

- New unit tests (`pool_name_tests`): auto-name with/without machine id, explicit-name-preferred on cache-miss, and auto-name fallback when unregistered.
- `cargo test -p dora-tensor-pool-python`: 40 passed.
- `cargo fmt --all -- --check` and `cargo clippy -p dora-tensor-pool-python -- -D warnings`: clean.

Reachable on the sender-side write path for any pool registered with an explicit `name=`; the tensor-pool extension is opt-in and outside the 1.0 guarantees (per #3152).

---
_Generated by [Claude Code](https://claude.ai/code/session_01SYaXAByMZaV36WJfHbfZ1N)_

### DIFF
--- a/libraries/extensions/tensor-pool/python/src/transport.rs
+++ b/libraries/extensions/tensor-pool/python/src/transport.rs
@@ -267,6 +267,32 @@ static PINNED_POOL: LazyLock<std::sync::Mutex<HashMap<u64, PoolSlot>>> =
 static TRANSIT_META: LazyLock<std::sync::Mutex<HashMap<u64, (u64, i32)>>> =
     LazyLock::new(|| std::sync::Mutex::new(HashMap::new()));
 
+/// Persistent map of pool counter → the exact `/dev/shm` segment name chosen
+/// at registration.
+///
+/// The read side resolves each pool's segment name from the daemon, but the
+/// per-frame write fast path reconstructs the name locally on a `PINNED_POOL`
+/// cache-miss (slot eviction). Reconstructing via `pool_shmem_name` is correct
+/// for the auto-named case, but wrong for a pool registered with an explicit
+/// `name=` (added in #3079), where the name is used verbatim and does not
+/// follow the auto format. Without this record the cache-miss reopen would
+/// target a segment that does not exist, `ShmemConf::open()` would fail, and
+/// the write would be silently dropped (#3207).
+///
+/// Keyed by the bare per-process counter (matching `PINNED_POOL` /
+/// `TRANSIT_META`); populated in `register_tensor_pool`, cleared in every free
+/// path (`rollback_local_pool`, `free_tensor_pool`, and the GC drain in
+/// `process_pending_tensor_pool_frees`). The free-time removal is
+/// ownership-guarded via `owned_pool_counter`, and the write-side lookup runs
+/// under the same guard by construction: it fires only after
+/// `PINNED_POOL.remove(&counter)` returns `None`, i.e. only for a counter this
+/// process itself registered (nothing else could have inserted into
+/// `PINNED_POOL` under that counter). A cross-process pool id with an aliased
+/// bare counter has no `PINNED_POOL` entry here, so this path is not reached
+/// for it.
+static POOL_NAMES: LazyLock<std::sync::Mutex<HashMap<u64, String>>> =
+    LazyLock::new(|| std::sync::Mutex::new(HashMap::new()));
+
 /// Receiver-side GPU cache per pool.
 /// Keeps Shmem alive to prevent munmap, preserving stable mmap addresses
 /// and valid GPU VAs for zero-copy reads across iterations.
@@ -390,6 +416,10 @@ fn rollback_local_pool(
         }
     }
     TRANSIT_META
+        .lock()
+        .unwrap_or_else(|e| e.into_inner())
+        .remove(&pool_counter);
+    POOL_NAMES
         .lock()
         .unwrap_or_else(|e| e.into_inner())
         .remove(&pool_counter);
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

I traced the counter key through all three sites: `register_tensor_pool` stores the exact `shmem_name` in `POOL_NAMES` keyed by `pool_counter`, and both the write fast path (`buffer_id.rsplit_once('_')`) and `free_tensor_pool` (`strip_prefix("pool_")` + `rsplit_once('_')`) recover the same counter, so the lookup and cleanup key are consistent. Counters come from the monotonic `PINNED_COUNTER`, so a removed entry is never re-keyed to a different pool — the fallback-to-auto-name branch is only reachable for a segment this process never registered (older process), which is the documented degradation. `POOL_NAMES` is cleared in both `free_tensor_pool` and `rollback_local_pool`, so no entry leaks. The unit tests exercise the actual helpers with distinct large counters and clean up, so they're meaningful and parallel-safe.

---

phil-opp (commented):

The two merge commits from `main` since my last review only resolved the conflict in `transport.rs`; the net change is unchanged in substance. Re-traced the counter-keyed `POOL_NAMES` lifecycle on the merged diff: insert in `register_tensor_pool`, removal in both `rollback_local_pool` and `free_tensor_pool` (ownership-guarded via `owned_pool_counter`, matching `PINNED_POOL`), lookup preferred over the auto-name fallback in `write_pool_shmem_name`. No leak, no wrong-segment reopen, and `POOL_NAMES` is never locked while holding another pool lock, so no deadlock. Latest commits look safe to merge, no new issues found.

---

phil-opp:
**Automated review by Claude** — this comment is fully automated and has not been vetted by a human.

The core fix (preferring the registered explicit name on a `PINNED_POOL` cache-miss) looks right, and the reopen path is adequately guarded downstream (`DORADMA_MAGIC` + size-vs-capacity checks, and the `FREED_POOL_IDS` early-return). One gap in the new `POOL_NAMES` lifecycle, though:

`free_tensor_pool` and `rollback_local_pool` now clear the `POOL_NAMES` entry, but `process_pending_tensor_pool_frees` — the drop/GC-based free path — does not. That function already removes the sibling `PINNED_POOL` slot and `TRANSIT_META` entry (ownership-guarded via `owned_pool_counter`) around `transport.rs:1359`/`1372`, and a `POOL_NAMES.remove(&c)` seems to belong there too, right next to the `TRANSIT_META` removal.

As written, a sender that registers explicit-name pools and lets the Python handles get garbage-collected (rather than calling `free_tensor_pool` explicitly) leaks one `POOL_NAMES` String per pool for the process lifetime — the same leak the new comment in `free_tensor_pool` describes preventing, just via the other free path. Counters are monotonic so it can't cause a wrong-segment reopen; it's purely a slow leak. A small assertion that `POOL_NAMES` is empty after a drop-path free would keep the third removal site from silently drifting again.

---

phil-opp:
review by claude:

The underlying bug is real and still live on main, but the branch predates #3056/#3219 and the merge is not mechanical. write_pool_shmem_name (transport.rs:232) looks up POOL_NAMES by a bare unqualified counter with no ownership guard, while the file's own comment at :3139 documents exactly that hazard and guards the free site — so a node receiving a peer's pool_A_5 can write into its own live pool #5. The branch also duplicates main's pool_shmem_name helper (added in #3056 as "the single source of truth"), still uses the pre-#3056 constant counter seed (:36), misses a POOL_NAMES removal in the GC free path (unbounded growth), and has one genuinely dead cleanup call. Rebase must delete the duplicate helper and adopt pool_shmem_name/parse_pool_counter.

phil-opp:
Rebased onto current `main` (`2f888cb`) and re-implemented on top of the #3056/#3169/#3219 helpers. The branch now carries a single new commit; force-pushed.

Addressed in the rebase:

- **Ownership guard on the `POOL_NAMES` lookup.** The write cache-miss path now consults `POOL_NAMES` only when `owned_pool_counter(&buffer_id, self.node_id)` yields the same counter — matching `free_tensor_pool` / `process_pending_tensor_pool_frees`. A peer pool id with an aliased bare counter falls straight through to `pool_shmem_name`, so this fix cannot cause a node to write into its own live pool #N under a peer's `pool_A_N`.
- **Duplicate helper deleted.** No `auto_pool_shmem_name` on this branch — the name reconstruction is `pool_shmem_name` from #3056 (the single source of truth), and the ownership check is `owned_pool_counter` from #3169.
- **GC-path `POOL_NAMES` removal added.** `process_pending_tensor_pool_frees` now clears `POOL_NAMES` on every owned counter — outside the `PINNED_POOL`-hit arm, so it runs even when the slot was evicted between register and free. Same placement in `free_tensor_pool` and `rollback_local_pool`.
- **`open_pool_segment_on_miss` helper** replaces my earlier inline expansion of the write cache-miss branch, keeping the hot-path structure identical to `main`.
- **Pre-#3056 constant counter seed:** N/A — I'm on top of #3056, which uses `random_u64_seed()`.

`cargo test -p dora-tensor-pool-python`: 46 passed (main's 44 + 2 new: `pool_names_stores_the_registered_segment_name`, `pool_names_removal_is_unconditional`). `cargo fmt --all -- --check` and `cargo clippy -p dora-tensor-pool-python -- -D warnings`: clean.

---

PR #3213, opened 2026-08-18
feat(node-api)!: remove Arrow from the frozen public API before 1.0

phil-opp:
At 1.0 everything reachable from `dora-node-api` is frozen for the life of 1.x, and Arrow is reachable three ways: `pub use arrow;`, `ArrowData(pub ArrayRef)` with its `Deref`, and `IntoArrow::A: Array`. Arrow ships a major roughly every 6–8 weeks, so freezing it pins dora 1.x to Arrow 59 and forces users to choose between dora and a current polars / datafusion / parquet.

The wire format was never the problem — dora serializes with Arrow IPC, which is version-independent, so two nodes built against different Arrow majors already interoperate. The coupling is purely the in-process Rust type boundary.

**Design** (full rationale in `docs/plan-arrow-version-decoupling.md`, added in the first commit):

- `DoraArray`, a dora-owned newtype with a private field, replaces `ArrowData` in every signature. `IntoArrow` drops its `type A: Array` bound and returns `DoraArray`.
- `pub use arrow;` becomes `arrow_v59` / `arrow_v58`, each feature-gated, `default = []`. A bare `arrow` re-export changes meaning silently when dora bumps; `arrow_v59` either exists and means Arrow 59, or it is visibly gone. That makes a future bump **additive** rather than mutating.
- `arrow-v59` pulls no extra dependency — it re-exports the copy dora already links. Only majors other than the internal one get an aliased optional dep, so you compile only what you ask for.
- Zero-copy `TryFrom`/`TryInto` between majors via the Arrow C Data Interface, which the operator API already uses in-tree. The fallible traits, not `From`/`Into`: importing through the C Data Interface can fail on a malformed schema or array, and an infallible `From` impl would have to panic internally on bad data.

The payoff is that dora can bump its internal Arrow in a **minor** release.

Worth noting: **the `dora new` templates needed no change**. `value.into_arrow()` and `(&data).try_into()?` never name an Arrow type, so the common node path needs no feature at all.

### The FFI bridge

`libraries/arrow-convert/src/ffi_bridge.rs` is the one `unsafe` module. `arrow58::ffi::FFI_ArrowArray` and `arrow::ffi::FFI_ArrowArray` are distinct Rust types over the same frozen C ABI, so the hop is a `transmute`. Four `const _: () = assert!(...)` blocks reject a size or alignment mismatch at compile time — they do not prove field-for-field identity, which is the stated uncovered case. The transmute *moves* the owned struct, so the `release` duty transfers exactly once.

Tested specifically for that: a test builds an Arrow 58 array over an allocation carrying a `Drop` counter, imports to 59, drops every Arrow 58 handle and asserts the counter is still 0, then drops the 59 array **without reading it** and asserts the counter is exactly 1. Mirror test for the other direction, plus nested/nullable `StructArray` and empty-array round trips. All five pass under `cargo +nightly miri` with no UB.

### One judgement call for reviewers

`dora_arrow_convert::internal` remains Arrow-typed and `pub`. `DoraArray` lives in a different crate from the code that builds and unwraps it, and Rust has no cross-crate `pub(crate)`. Routing it through the `arrow-v59` feature actively backfires: `dora-node-api` would have to enable it unconditionally, and cargo feature unification would then hand the ungated accessor to every downstream user — the exact silent drift the gate exists to prevent.

So `internal` is `pub`, deliberately not re-exported from `dora-node-api`, and documented semver-exempt. **`dora-node-api`'s frozen surface is Arrow-free**, which is the surface that matters, but the workspace's public surface is not — modulo this seam. The alternative is folding `dora-arrow-convert` into `dora-node-api` so the accessors become genuinely `pub(crate)`; that is a larger, more invasive change and was not taken unilaterally. Worth a reviewer's opinion.

### Other deviations from the design doc

- `data_type()` is feature-gated rather than returning a type URN. `TypeRegistry` has no URNs for nested lists, dictionaries, unions, or timestamps-with-timezone — precisely the types whose identity a caller most needs, so a URN accessor would be lossy exactly where it matters. An ungated `type_name() -> String` covers logging.
- The IPC encoders are **retyped, not hidden**. They have real cross-crate consumers (record-node, daemon, the Python `sample_handler`, a bench, `tests/copy_count.rs`), so `pub(crate)` would break them and `#[doc(hidden)]` would leave Arrow in the frozen API. They now take `DoraArray`; `pub(crate)` `*_data` twins keep the internal hot path from paying an extra `to_data()`.
- Gated `IntoArrow` impls are listed per concrete Arrow 59 array type, because a blanket `impl<A: Array> IntoArrow for A` is impossible (E0119 against `impl IntoArrow for u8`).

### Verification

fmt, clippy `-D warnings`, `cargo check --all --all-targets`, `cargo check --examples`, and the full feature matrix (none / v58 / v59 / both) all clean. `cargo test --all`: 1943 passed, 2 failed — the known `rmw_zenoh_pubsub` pair that needs multicast and fails on `main` in the same container. Miri on the FFI tests as above.

**Not verified:** `cargo check -p rust-ros2-example-node --features ros2`. No ROS distro in the dev container, so msg-gen emits nothing. `git diff` shows that crate byte-identical, but it was not observed compiling — nightly CI's ros2-bridge job covers it.

Breaking for every Rust node: `Event::Input`'s payload type changes, and code naming Arrow types now needs `features = ["arrow-v59"]`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,6 +190,19 @@ jobs:
           -p dora-operator-api-python
           -p dora-ros2-bridge-python
           -p dora-runtime-python
+      # The C/C++ bindings name Arrow types directly, so they depend on
+      # `dora-node-api`'s `arrow-v59` feature. `cargo check --all` above does
+      # NOT catch a missing opt-in: another workspace member (the C++ binding)
+      # enables the feature, and cargo unifies features across a whole-workspace
+      # build, so the C binding compiles by accident. Built per-crate the
+      # unification is gone and the break appears. Without this, the first place
+      # it surfaces is `find_package smoke`, which is skipped on `pull_request`
+      # and only runs on a merge-queue batch — one wasted batch cycle per
+      # attempt (#3213).
+      - name: Check C/C++ bindings compile standalone
+        run: |
+          cargo check -p dora-node-api-c
+          cargo check -p dora-node-api-cxx
 
   test:
     # Linux-only on PR CI (#1716). macOS + Windows coverage runs in nightly.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6,7 +6,7 @@ version = 4
 name = "action-example-client"
 version = "1.0.0-rc.4"
 dependencies = [
- "arrow",
+ "arrow 59.0.0",
  "dora-node-api",
  "eyre",
 ]
@@ -15,7 +15,7 @@ dependencies = [
 name = "action-example-server"
 version = "1.0.0-rc.4"
 dependencies = [
- "arrow",
+ "arrow 59.0.0",
  "dora-node-api",
  "eyre",
 ]
@@ -215,22 +215,54 @@ checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "arrow"
+version = "58.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6cfdd0833e32a9874d2b55089333ad310c0be208aafa277385ce2461dec90be3"
+dependencies = [
+ "arrow-arith 58.4.0",
+ "arrow-array 58.4.0",
+ "arrow-buffer 58.4.0",
+ "arrow-cast 58.4.0",
+ "arrow-data 58.4.0",
+ "arrow-ord 58.4.0",
+ "arrow-row 58.4.0",
+ "arrow-schema 58.4.0",
+ "arrow-select 58.4.0",
+ "arrow-string 58.4.0",
+]
+
+[[package]]
+name = "arrow"
 version = "59.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ffaaa3e009861fd829d0a24dd6f115aa8e4634324bb092147d43baafe69ca4a7"
 dependencies = [
- "arrow-arith",
- "arrow-array",
- "arrow-buffer",
- "arrow-cast",
- "arrow-data",
+ "arrow-arith 59.2.0",
+ "arrow-array 59.2.0",
+ "arrow-buffer 59.2.0",
+ "arrow-cast 59.2.0",
+ "arrow-data 59.2.0",
  "arrow-ipc",
- "arrow-ord",
+ "arrow-ord 59.2.0",
  "arrow-pyarrow",
- "arrow-row",
- "arrow-schema",
- "arrow-select",
- "arrow-string",
+ "arrow-row 59.2.0",
+ "arrow-schema 59.2.0",
+ "arrow-select 59.2.0",
+ "arrow-string 59.2.0",
+]
+
+[[package]]
+name = "arrow-arith"
+version = "58.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0a41203398f0eaa6f7ec8e62c0da742a21abf282c148fc157f6c35c90e29981a"
+dependencies = [
+ "arrow-array 58.4.0",
+ "arrow-buffer 58.4.0",
+ "arrow-data 58.4.0",
+ "arrow-schema 58.4.0",
+ "chrono",
+ "num-traits",
 ]
 
 [[package]]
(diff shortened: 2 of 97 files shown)


phil-opp:

No correctness issues found in the risk-bearing code. I focused on the new `dora-arrow-convert` core (`DoraArray`, the `IntoArrow` associated-type removal, the `internal` seam) and the `unsafe` FFI bridge.

The arrow58↔arrow59 `FFI_ArrowArray` / `FFI_ArrowSchema` transmute in `libraries/arrow-convert/src/ffi_bridge.rs` is sound: it's guarded by compile-time size/align asserts, ownership is transferred exactly once (the owned `to_ffi` result is consumed by the `transmute`, `from_ffi` takes it by value, and the schema is released via its `Drop`), and the release-callback path is exercised in both directions by drop-counter tests plus nested/nullable/empty round-trips (miri-clean per the description). The one uncovered case — a future arrow-rs layout change that preserves size and alignment — is explicitly documented and is equivalent to breaking arrow-rs's own C interop, so that's a fair scope call.

On the flagged judgment call: keeping `dora_arrow_convert::internal` `pub` but semver-exempt and not re-exporting it from `dora-node-api` is a reasonable substitute for a cross-crate `pub(crate)` — the frozen `dora-node-api` surface does stay Arrow-free, which is the surface that matters for the 1.0 freeze.

Caveat on scope: I did not read all 92 changed files; the bulk are mechanical `Cargo.toml` / example retypings, which are low-risk given the reported clean fmt / clippy / feature-matrix and test run. The added tests are meaningful — the FFI drop-counter tests would catch a layout mistake as a use-after-free / double-free, and the round-trips assert values + type + `to_data()` equality across the hop.

---

phil-opp:
The one commit since my last review (`970a949`) is a safe build/CI fix, not a behavior change: it adds `features = ["arrow-v59"]` to `apis/c/node/Cargo.toml` so `dora-node-api-c` (which names `arrow_v59::array::AsArray` / `DoraArray::as_array` directly) still compiles after this PR moves those behind the feature gate, and adds a per-crate `cargo check` of both bindings to PR CI so whole-workspace feature unification can't mask a break. No correctness concerns, and it doesn't touch the frozen `dora-node-api` public surface. My earlier assessment of the core change stands — no new issues found.

---

---

phil-opp:
The one non-merge commit since my last review (`4de30f0`) is a Cargo-feature completeness fix, not a behavior change: it adds the `arrow-v59` opt-in to `dora-record-node`/`dora-replay-node` (whose `main.rs` references `arrow_v59::…` and `DoraArray::as_array` / `IntoArrow for ArrayRef`), and gives `dora-operator-api` passthrough `arrow-v59`/`arrow-v58` features plus a gated `arrow_v59` re-export. Without these, those crates compiled only under whole-workspace feature unification and would have broken on a release build, `cargo install --path`, or any downstream consumer — so this closes a real gap in the migration. The gated `arrow_v59` re-export keeps the frozen public surface Arrow-free and fails visibly on a major bump, and the added `typed_array_round_trips_through_dora_array` test meaningfully exercises the send/receive/`from_array` paths the feature gates.

---
🤖 _Fully automated review by Claude — no human has vetted this comment. It may be wrong; please verify before acting on it._

---

phil-opp:

The one non-merge commit since my last review (`0b1124d`, "docs(node): keep the PreparedIpc doctest Arrow-free") is documentation-only: it rewrites the `PreparedIpc` doctest off the removed `dora_node_api::arrow` path onto `IntoArrow` / `into_vec`. It's internally consistent with the new API (`into_arrow()` → `DoraArray`, `PreparedIpc::new(&DoraArray)`, `decode_arrow_ipc` → `DoraArray`, `into_vec::<u64>(&decoded)?`) and keeps the example on the Arrow-free path. No new findings — safe from my side.

---

phil-opp:
**Automated review by Claude** — this comment is fully automated and has not been vetted by a human.

The decoupling itself looks solid — the private-field `DoraArray`, the version-gated `arrow_v59`/`arrow_v58` re-exports, and the C-Data-Interface `unsafe` bridge with its bidirectional release-callback drop tests all check out. One small, non-blocking consistency point:

`apis/python/node` calls `send_output(…, arrow::array::make_array(arrow_array))` (`apis/python/node/src/lib.rs:1099-1102`), passing an `ArrayRef`, which relies on the `arrow-v59`-gated `impl IntoArrow for ArrayRef`. But that crate's `Cargo.toml` depends on `dora-node-api` without `features = ["arrow-v59"]`. It compiles today only because its sibling deps `dora-operator-api-python` and `dora-runtime-python` both enable `arrow-v59` on the shared `dora-node-api`, and Cargo unifies that feature into the wheel's build closure. That's the same "green only under feature unification" fragility you removed explicitly for `dora-record-node`, `dora-replay-node`, and the C binding by giving them their own `features = ["arrow-v59"]`. Declaring it directly on `apis/python/node`'s `dora-node-api` dependency too would make it robust to a future dependency change and consistent with the invariant the rest of the PR enforces.

(The matching whitespace-only edit on `libraries/extensions/tensor-pool/python` looks incidental — that crate uses the `arrow` crate directly and never calls `send_output`, so it genuinely needs no feature.)

---

PR #3043, opened 2026-08-05
ci(nightly): add concurrency group so stale nightly runs are cancelled

phil-opp:
## Summary

Fixes #3019.

`.github/workflows/nightly.yml` had no `concurrency:` block, so nothing cancels a previous nightly when a new one starts. On days with several `nightly.yml` edits (its `push` trigger is path-filtered to the workflow itself) or overlapping scheduled/dispatch runs, multiple ~30-job matrices across Linux/macOS/Windows could queue simultaneously and compete with each other — and with the Trunk merge queue — for the shared runner pool. The older runs have no value once a newer one exists.

## Change

Adds the concurrency block suggested in the issue, right after the `on:` triggers:

```yaml
concurrency:
  group: nightly-${{ github.ref }}
  cancel-in-progress: true
```

Scheduled runs share `github.ref` (`refs/heads/main`), so a fresh nightly supersedes an in-flight/queued earlier one instead of stacking.

## Risk

Low. Nightly is not a merge-queue required status, so cancelling one of its runs cannot wedge the queue.

## Validation

- YAML parses; `concurrency` key present; all 30 jobs still parse.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xrz6gC6syG2G6RZ2K1dUC8)_

PR #3214, opened 2026-08-18
ci(nightly): size the topic-and-top budget for its maturin build, and report cancelled runs

phil-opp:
The nightly has been substantively green for days but cannot conclude
`success`: `topic-and-top-smoke` keeps blowing its 15-minute cap, and a
timed-out job reports as `cancelled`, which makes the whole run
`cancelled` and skips *both* nightly automations. Six of the eleven
nightlies between 08-06 and 08-18 ended that way, including 32107803630,
where every other job — Examples on all 3 platforms, CLI Tests: Python on
all 3, Test (windows), Kani, all 8 cross-check targets — was green. The
Phase 5b exit criterion (7 consecutive clean nightlies) is unreachable
while that holds.

Two fixes.

**Give the job a budget that matches what it does.** The job is ~99% one
step: a from-source maturin build of `dora-node-api-python`, which took
11m23 / 12m29 / 13m29 in the three runs that finished, while the CLI
download, `dora up` and all twenty inspection assertions after it took
16-20 seconds. `rust-cache` runs `save-if: false` here, so the job never
populates a cache and only gets a warm `target/` by luck. The build moves
into its own step with a 25-minute cap (~1.9x the slowest build that has
actually completed, the same headroom #3094 gave the E2E step for the
same reason); the job cap goes to 30. An overrun now says which half
broke.

**Report cancellations instead of going silent.** `file-issue-on-failure`
also fires on `cancelled`, and the issue body gets a "Timed out" section
that names the step the job was in and how long it ran — a blown cap is a
different diagnosis from a test failure, and "cancelled" reads like
somebody stopped the run.

The catch is #3043's `cancel-in-progress`: a superseded run marks every
in-flight job `cancelled`, and firing on that would file an issue on every
push to main that lands during a nightly. The step classifies each
cancellation by whether the run carried on afterwards — a `timeout-minutes`
cap kills one job while the rest of the matrix works for many more minutes,
a supersede kills everything in one wave and the run ends there. Checked
against real data: run 32107803630 (cap blown at 07:09:27, other jobs ran
until ~08:00) files; the 12:07:49-12:08:03 wave that ended run 32022246649
does not. A run with a single cancelled job is a timeout by construction —
a supersede would have taken the reporter down too.

`close-issue-on-success` is unchanged and still refuses to close on a
cancelled run.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -529,7 +529,11 @@ jobs:
     name: Topic & top inspection smoke
     runs-on: ubuntu-latest
     needs: build-cli
-    timeout-minutes: 15
+    # Sized as "the capped build step below, plus slack for everything else".
+    # Setup actions take ~20s and the actual inspection smokes take ~20s (see
+    # the successful runs on 2026-08-08/09/10), so this cap only bites if the
+    # build step's own 25-minute cap somehow does not.
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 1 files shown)


phil-opp:

No blocking issues — this is workflow-only, no product impact. Sizing the maturin build in its own 25-minute step and raising the job cap to 30 is sensible, and firing `file-issue-on-failure` on `cancelled` with the timeout-vs-supersede classifier matches the stated heuristic.

One minor edge in the classifier: it infers "the run carried on afterwards" from the max `completed_at` across completed jobs, so a run with two or more cancellations where a genuinely timed-out job happens to be the last to finish could be misclassified as a supersede and skip filing an issue. It only affects whether an issue is filed (self-correcting on the next nightly) and the single-cancellation shortcut covers the common case, so not blocking.

---

PR #3215, opened 2026-08-18
fix(daemon): report required event delivery failures

SunSunSun689:
Fixed the shared root cause for the silent required-event delivery failures.

This branch handles the same class of issue across the relevant daemon paths:

  - `Reload` to local node/operator channels now returns an error on full/closed channels instead of reporting success.
  - `InputClosed` / `AllInputsClosed` local node events now use strict delivery semantics where the caller can observe failures; closed channels are cleaned up.
  - `break_input` reuses the same strict delivery path and logs/cleans up when there is no direct caller response path.
  - Inter-daemon `OutputClosed` is now treated as a required lifecycle event: if the Zenoh publish queue is full, it is retained and sent from a background task when capacity is available.
  - Regular inter-daemon `Output` remains best-effort, but full/closed enqueue failures are recorded in `net_publish_failures`.
  - `OutputClosed` stays in the same FIFO as earlier `Output` messages for that output, so close cannot overtake already queued data.

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -207,6 +207,10 @@ const MIN_METRICS_WINDOW_SECS: f64 = 0.1;
 /// patterns; regular outputs are dropped with a warning when full, while
 /// required lifecycle events wait for capacity in a background task.
 const ZENOH_PUBLISH_CHANNEL_CAPACITY: usize = 256;
+/// Upper bound for a single Zenoh publish attempt. Required lifecycle events use
+/// blocking publishers, so this keeps a gone or congested peer from pinning a
+/// per-output drain task forever.
+const ZENOH_PUBLISH_TIMEOUT: Duration = Duration::from_secs(1);
 /// How long the daemon keeps trying to (re)connect to the coordinator before
 /// giving up and exiting. Bounds the orphan-daemon window when the coordinator
 /// is permanently gone (dora-rs/dora#1996); a reachable coordinator connects
@@ -7799,6 +7803,10 @@ fn remote_output_publish_queue(
             let output_id = output_id.clone();
             tokio::spawn(async move {
                 while let Some(msg) = rx.recv().await {
+                    // Regular Output and its OutputClosed use distinct Drop/Block
+                    // publishers on the same Zenoh key, but this per-output FIFO
+                    // is the ordering boundary: each publish is awaited before the
+                    // next one is submitted, so a close cannot pass earlier data.
                     publish_zenoh_outbound(msg).await;
                 }
                 tracing::debug!(?output_id, "per-output zenoh publish drain task exiting");
@@ -7839,12 +7847,41 @@ fn try_enqueue_zenoh_outbound<T>(tx: &mpsc::Sender<T>, outbound: T) -> ZenohEnqu
     }
 }
 
+#[derive(Debug)]
+enum ZenohPublishOutcome<E> {
+    Published,
+    Failed(E),
+    TimedOut,
+}
+
+async fn await_zenoh_put_with_timeout<F, E>(publish: F, timeout: Duration) -> ZenohPublishOutcome<E>
+where
+    F: std::future::IntoFuture<Output = Result<(), E>>,
+{
+    match tokio::time::timeout(timeout, publish.into_future()).await {
+        Ok(Ok(())) => ZenohPublishOutcome::Published,
+        Ok(Err(err)) => ZenohPublishOutcome::Failed(err),
+        Err(_) => ZenohPublishOutcome::TimedOut,
+    }
+}
+
 async fn publish_zenoh_outbound(msg: ZenohOutbound) {
-    if let Err(e) = msg.publisher.put(msg.serialized).await {
-        tracing::error!("zenoh publish failed: {e}");
-        msg.net_publish_failures
-            .fetch_add(1, atomic::Ordering::Relaxed);
-        return;
+    match await_zenoh_put_with_timeout(msg.publisher.put(msg.serialized), ZENOH_PUBLISH_TIMEOUT)
+        .await
+    {
+        ZenohPublishOutcome::Published => {}
+        ZenohPublishOutcome::Failed(e) => {
+            tracing::error!("zenoh publish failed: {e}");
+            msg.net_publish_failures
+                .fetch_add(1, atomic::Ordering::Relaxed);
+            return;
+        }
+        ZenohPublishOutcome::TimedOut => {
+            tracing::error!("zenoh publish timed out after {ZENOH_PUBLISH_TIMEOUT:?}");
+            msg.net_publish_failures
+                .fetch_add(1, atomic::Ordering::Relaxed);
+            return;
+        }
     }
     // Relaxed ordering is correct: counters are read-only for metrics
     // reporting and never used as synchronization guards.
@@ -9046,6 +9083,24 @@ mod fault_tolerance_tests {
         );
         assert_eq!(data.serialized, b"data");
         assert_ne!(close.serialized, b"data");
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn zenoh_publish_wait_is_bounded() {
+        let publish = std::future::pending::<Result<(), eyre::Report>>();
+        let handle = tokio::spawn(await_zenoh_put_with_timeout(publish, ZENOH_PUBLISH_TIMEOUT));
+
+        tokio::task::yield_now().await;
+        assert!(
+            !handle.is_finished(),
+            "the helper must not finish before the timeout elapses"
+        );
+
+        tokio::time::advance(ZENOH_PUBLISH_TIMEOUT).await;
+        assert!(matches!(
+            handle.await.unwrap(),
+            ZenohPublishOutcome::TimedOut
+        ));
     }
 
     #[test]
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -3719,8 +3719,16 @@ impl Daemon {
                         .map(|r| r.remove(&(target_node.clone(), target_input.clone())))
                         .unwrap_or(false);
                     if removed {
-                        close_input(dataflow, &target_node, &target_input, &self.clock)
-                            .map_err(|err| format!("{err:?}"))
+                        if let Err(err) =
+                            close_input(dataflow, &target_node, &target_input, &self.clock)
+                        {
+                            tracing::warn!(
+                                %target_node,
+                                %target_input,
+                                "failed to deliver InputClosed after removing mapping: {err:?}"
+                            );
+                        }
+                        Ok(())
                     } else {
                         Err(format!(
                             "mapping `{source_node}/{source_output}` -> \
@@ -5302,7 +5310,7 @@ impl Daemon {
             format!("Reload failed: no running dataflow with ID `{dataflow_id}`")
         })?;
         let Some(channel) = dataflow.subscribe_channels.get(&node_id) else {
-            return Err(eyre!("node `{node_id}` not connected"));
+            return Ok(());
         };
         match send_with_timestamp(channel, NodeEvent::Reload { operator_id }, &self.clock) {
             Ok(true) => {
@@ -5487,38 +5495,8 @@ impl Daemon {
             format!("send out failed: no running dataflow with ID `{dataflow_id}`")
         })?;
 
-        // Get or create publisher (lazy, cached per output)
-        let publisher = match dataflow.publishers.entry(output_id.clone()) {
-            std::collections::btree_map::Entry::Occupied(e) => e.get().clone(),
-            std::collections::btree_map::Entry::Vacant(e) => {
-                let publish_topic =
-                    zenoh_daemon_control_topic(dataflow.id, &output_id.0, &output_id.1);
-                tracing::debug!("declaring control publisher on {publish_topic}");
-                let publisher = self
-                    .zenoh_session
-                    .declare_publisher(publish_topic)
-                    .congestion_control(CongestionControl::Drop)
-                    .express(true)
-                    .priority(Priority::RealTime)
-                    .await
-                    .map_err(|err| eyre!(err))
-                    .context("failed to create zenoh publisher")?;
-                let arc = Arc::new(publisher);
-                e.insert(arc.clone());
-                arc
-            }
-        };
-        let payload_len = serialized_event.len() as u64;
-
-        // Offload Zenoh I/O to the drain task — never blocks the event loop.
-        let outbound = ZenohOutbound {
-            publisher,
-            serialized: serialized_event,
-            payload_len,
-            net_bytes_sent: dataflow.net_bytes_sent.clone(),
-            net_messages_sent: dataflow.net_messages_sent.clone(),
-            net_publish_failures: dataflow.net_publish_failures.clone(),
-        };
+        let publisher = remote_receiver_publisher(&self.zenoh_session, dataflow, output_id).await?;
+        let outbound = zenoh_outbound(dataflow, publisher, serialized_event);
         enqueue_required_zenoh_outbound(
             &self.zenoh_publish_tx,
             outbound,
@@ -5538,37 +5516,8 @@ impl Daemon {
             format!("send out failed: no running dataflow with ID `{dataflow_id}`")
         })?;
 
-        // Get or create publisher (lazy, cached per output)
-        let publisher = match dataflow.publishers.entry(output_id.clone()) {
-            std::collections::btree_map::Entry::Occupied(e) => e.get().clone(),
-            std::collections::btree_map::Entry::Vacant(e) => {
-                let publish_topic =
-                    zenoh_daemon_control_topic(dataflow.id, &output_id.0, &output_id.1);
-                tracing::debug!("declaring control publisher on {publish_topic}");
-                let publisher = self
-                    .zenoh_session
-                    .declare_publisher(publish_topic)
-                    .congestion_control(CongestionControl::Drop)
-                    .express(true)
-                    .priority(Priority::RealTime)
-                    .await
-                    .map_err(|err| eyre!(err))
-                    .context("failed to create zenoh publisher")?;
-                let arc = Arc::new(publisher);
-                e.insert(arc.clone());
-                arc
-            }
-        };
-        let payload_len = serialized_event.len() as u64;
-
-        let outbound = ZenohOutbound {
-            publisher,
-            serialized: serialized_event,
-            payload_len,
-            net_bytes_sent: dataflow.net_bytes_sent.clone(),
-            net_messages_sent: dataflow.net_messages_sent.clone(),
-            net_publish_failures: dataflow.net_publish_failures.clone(),
-        };
+        let publisher = remote_receiver_publisher(&self.zenoh_session, dataflow, output_id).await?;
+        let outbound = zenoh_outbound(dataflow, publisher, serialized_event);
         handle_publish_enqueue_result(
             dataflow,
             try_enqueue_zenoh_outbound(&self.zenoh_publish_tx, outbound),
@@ -7357,24 +7306,27 @@ fn send_input_closed_strict(
 /// while a timer keeps ticking.
 ///
 /// Shared drain-completion tail of [`close_input`] and [`break_input`]; a
-/// no-op if the node still has open/broken inputs or has no subscribe channel.
+/// no-op if the node still has open/broken inputs. Restart bookkeeping is
+/// independent from the node event channel; only the `AllInputsClosed` send
+/// requires a live channel.
 fn signal_all_inputs_closed_if_drained(
     dataflow: &mut RunningDataflow,
     receiver_id: &NodeId,
     clock: &HLC,
 ) -> eyre::Result<()> {
-    let Some(channel) = dataflow.subscribe_channels.get(receiver_id) else {
-        return Ok(());
-    };
     // As at the subscribe site: either "nothing left open" (pre-existing
     // behavior, and true for a source) or "drained" (the node we are about
     // to tell to finish) disables restart.
-    if (dataflow.open_inputs(receiver_id).is_empty() || dataflow.is_drained(receiver_id))
-        && !dataflow.has_broken_input(receiver_id)
-        && let Some(node) = dataflow.running_nodes.get_mut(receiver_id)
-    {
+    let should_disable_restart = (dataflow.open_inputs(receiver_id).is_empty()
+        || dataflow.is_drained(receiver_id))
+        && !dataflow.has_broken_input(receiver_id);
+    if should_disable_restart && let Some(node) = dataflow.running_nodes.get_mut(receiver_id) {
         node.disable_restart();
     }
+
+    let Some(channel) = dataflow.subscribe_channels.get(receiver_id) else {
+        return Ok(());
+    };
     if dataflow.is_finished(receiver_id)
         && match send_with_timestamp(channel, NodeEvent::AllInputsClosed, clock) {
             Ok(true) => true,
@@ -7789,6 +7741,50 @@ impl CoreNodeKindExt for CoreNodeKind {
                 matches!(&n.source, NodeSource::Local) && n.path == DYNAMIC_SOURCE
             }
         }
+    }
+}
+
+async fn remote_receiver_publisher(
+    zenoh_session: &zenoh::Session,
+    dataflow: &mut RunningDataflow,
+    output_id: &OutputId,
+) -> eyre::Result<Arc<zenoh::pubsub::Publisher<'static>>> {
+    match dataflow.publishers.entry(output_id.clone()) {
+        std::collections::btree_map::Entry::Occupied(e) => Ok(e.get().clone()),
+        std::collections::btree_map::Entry::Vacant(e) => {
+            let publish_topic = zenoh_daemon_control_topic(dataflow.id, &output_id.0, &output_id.1);
+            tracing::debug!("declaring control publisher on {publish_topic}");
+            let publisher = zenoh_session
+                .declare_publisher(publish_topic)
+                // Local enqueue policy decides which events are best-effort
+                // (`Output`) and which must wait for capacity (`OutputClosed`).
+                // Once an event reaches Zenoh, it must not be dropped by the
+                // transport congestion policy.
+                .congestion_control(CongestionControl::Block)
+                .express(true)
+                .priority(Priority::RealTime)
+                .await
+                .map_err(|err| eyre!(err))
+                .context("failed to create zenoh publisher")?;
+            let publisher = Arc::new(publisher);
+            e.insert(publisher.clone());
+            Ok(publisher)
+        }
+    }
+}
+
+fn zenoh_outbound(
+    dataflow: &RunningDataflow,
+    publisher: Arc<zenoh::pubsub::Publisher<'static>>,
+    serialized: Vec<u8>,
+) -> ZenohOutbound {
+    ZenohOutbound {
+        publisher,
+        payload_len: serialized.len() as u64,
+        serialized,
+        net_bytes_sent: dataflow.net_bytes_sent.clone(),
+        net_messages_sent: dataflow.net_messages_sent.clone(),
+        net_publish_failures: dataflow.net_publish_failures.clone(),
     }
 }
 
@@ -8829,6 +8825,70 @@ mod fault_tolerance_tests {
             !df.open_inputs(&node_a).contains(&input_x),
             "the input is still removed from the open-input set"
         );
+    }
+
+    #[test]
+    fn close_input_still_disables_restart_when_receiver_channel_is_closed() {
+        let mut df = test_dataflow();
+        let clock = test_clock();
+        let node_a: NodeId = "node_a".to_string().into();
+        let input_x: DataId = "input_x".to_string().into();
+
+        df.open_inputs
+            .entry(node_a.clone())
+            .or_default()
+            .insert(input_x.clone());
+
+        let running = test_running_node();
+        let disable_restart = running.disable_restart.clone();
+        df.running_nodes.insert(node_a.clone(), running);
+
+        let (tx, rx) = mpsc::channel(NODE_EVENT_CHANNEL_CAPACITY);
+        df.subscribe_channels.insert(node_a.clone(), tx);
+        drop(rx);
+
+        let result = close_input(&mut df, &node_a, &input_x, &clock);
+        assert!(
+            result.is_err(),
+            "closed channels must still report an error"
+        );
+        assert!(
+            disable_restart.load(atomic::Ordering::Acquire),
+            "drain bookkeeping must disable restart even when the receiver channel is closed"
+        );
+        assert!(
+            !df.subscribe_channels.contains_key(&node_a),
+            "closed channels must still be cleaned up"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn reload_missing_node_channel_is_noop_for_non_host_daemons() {
+        let clock = Arc::new(HLC::default());
+        let (mut daemon, _events_rx) = Daemon::build_daemon(
+            Some("non-host".to_string()),
+            None,
+            DaemonId::new(Some("non-host".to_string())),
+            None,
+            clock,
+            None,
+            BTreeMap::new(),
+            LogDestination::Tracing,
+            None,
+            ZenohBind::Derived(LOCALHOST),
+            true,
+            false,
+        )
+        .await
+        .unwrap();
+
+        let dataflow_id = Uuid::new_v4();
+        daemon.running.insert(dataflow_id, test_dataflow());
+
+        daemon
+            .send_reload(dataflow_id, "node_on_other_daemon".to_string().into(), None)
+            .await
+            .expect("a daemon that does not host the reloaded node must not fail reload");
     }
 
     #[test]
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -463,7 +463,6 @@ pub struct Daemon {
     /// A `Destroy` that is holding its reply until this daemon's node
     /// processes are gone (#2980).
     pub(crate) pending_destroy: Option<PendingDestroy>,
-    pub(crate) zenoh_publish_tx: mpsc::Sender<ZenohOutbound>,
     pub(crate) remote_daemon_events_tx:
         Option<flume::Sender<eyre::Result<Timestamped<InterDaemonEvent>>>>,
     pub(crate) logger: DaemonLogger,
@@ -1894,18 +1893,6 @@ impl Daemon {
         // Use a large channel capacity to prevent deadlock
         let (dora_events_tx, dora_events_rx) = mpsc::channel(1000);
 
-        // Zenoh publish drain task: offload `.put().await` from the main event
-        // loop. All events for a given output, including `OutputClosed`, share
-        // this FIFO so close cannot pass data that was produced earlier.
-        let (zenoh_publish_tx, mut zenoh_publish_rx) =
-            mpsc::channel::<ZenohOutbound>(ZENOH_PUBLISH_CHANNEL_CAPACITY);
-        let _zenoh_drain_handle = tokio::spawn(async move {
-            while let Some(msg) = zenoh_publish_rx.recv().await {
-                publish_zenoh_outbound(msg).await;
-            }
-            tracing::debug!("zenoh publish drain task exiting");
-        });
-
         let daemon = Self {
             machine_id,
             logger: Logger {
@@ -1934,7 +1921,6 @@ impl Daemon {
             disable_multicast,
             bind_nodes_to_parent,
             pending_destroy: None,
-            zenoh_publish_tx,
             remote_daemon_events_tx,
             git_manager: Default::default(),
             extensions: ExtensionTable::new(),
@@ -5497,8 +5483,9 @@ impl Daemon {
 
         let publisher =
             remote_receiver_control_publisher(&self.zenoh_session, dataflow, output_id).await?;
+        let queue = remote_output_publish_queue(dataflow, output_id);
         let outbound = zenoh_outbound(dataflow, publisher, serialized_event);
-        spawn_required_zenoh_outbound(outbound, "inter-daemon output-closed event");
+        enqueue_required_zenoh_outbound(queue, outbound, "inter-daemon output-closed event");
         Ok(())
     }
 
@@ -5514,10 +5501,11 @@ impl Daemon {
 
         let publisher =
             remote_receiver_data_publisher(&self.zenoh_session, dataflow, output_id).await?;
+        let queue = remote_output_publish_queue(dataflow, output_id);
         let outbound = zenoh_outbound(dataflow, publisher, serialized_event);
         handle_publish_enqueue_result(
             dataflow,
-            try_enqueue_zenoh_outbound(&self.zenoh_publish_tx, outbound),
+            try_enqueue_zenoh_outbound(&queue, outbound),
             "regular inter-daemon output",
         );
         Ok(())
@@ -7800,6 +7788,27 @@ async fn remote_receiver_publisher(
     }
 }
 
+fn remote_output_publish_queue(
+    dataflow: &mut RunningDataflow,
+    output_id: &OutputId,
+) -> mpsc::Sender<ZenohOutbound> {
+    match dataflow.remote_output_queues.entry(output_id.clone()) {
+        std::collections::btree_map::Entry::Occupied(e) => e.get().clone(),
+        std::collections::btree_map::Entry::Vacant(e) => {
+            let (tx, mut rx) = mpsc::channel::<ZenohOutbound>(ZENOH_PUBLISH_CHANNEL_CAPACITY);
+            let output_id = output_id.clone();
+            tokio::spawn(async move {
+                while let Some(msg) = rx.recv().await {
+                    publish_zenoh_outbound(msg).await;
+                }
+                tracing::debug!(?output_id, "per-output zenoh publish drain task exiting");
+            });
+            e.insert(tx.clone());
+            tx
+        }
+    }
+}
+
 fn zenoh_outbound(
     dataflow: &RunningDataflow,
     publisher: Arc<zenoh::pubsub::Publisher<'static>>,
@@ -7869,11 +7878,34 @@ fn handle_publish_enqueue_result(
     }
 }
 
-fn spawn_required_zenoh_outbound(outbound: ZenohOutbound, event_kind: &'static str) {
-    tokio::spawn(async move {
-        tracing::debug!("publishing {event_kind} outside the shared zenoh data drain");
-        publish_zenoh_outbound(outbound).await;
-    });
+fn enqueue_required_zenoh_outbound(
+    tx: mpsc::Sender<ZenohOutbound>,
+    outbound: ZenohOutbound,
+    event_kind: &'static str,
+) {
+    match tx.try_send(outbound) {
+        Ok(()) => {}
+        Err(mpsc::error::TrySendError::Full(outbound)) => {
+            tokio::spawn(async move {
+                tracing::debug!(
+                    "zenoh publish channel full ({ZENOH_PUBLISH_CHANNEL_CAPACITY}); \
+                     waiting to enqueue {event_kind}"
+                );
+                if let Err(mpsc::error::SendError(outbound)) = tx.send(outbound).await {
+                    outbound
+                        .net_publish_failures
+                        .fetch_add(1, atomic::Ordering::Relaxed);
+                    tracing::error!("zenoh drain task is gone; dropping {event_kind}");
+                }
+            });
+        }
+        Err(mpsc::error::TrySendError::Closed(outbound)) => {
+            outbound
+                .net_publish_failures
+                .fetch_add(1, atomic::Ordering::Relaxed);
+            tracing::error!("zenoh drain task is gone; dropping {event_kind}");
+        }
+    }
 }
 
 /// Cached `@schema` bytes (keyed by their FNV-1a hash, most-recent last) for
@@ -8885,6 +8917,7 @@ mod fault_tolerance_tests {
             BTreeMap::new(),
             LogDestination::Tracing,
             None,
+            Vec::new(),
             ZenohBind::Derived(LOCALHOST),
             true,
             false,
@@ -8902,7 +8935,7 @@ mod fault_tolerance_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-    async fn output_closed_uses_separate_block_publisher_from_regular_output() {
+    async fn output_closed_uses_block_publisher_but_shares_output_fifo_with_regular_output() {
         let clock = Arc::new(HLC::default());
         let (mut daemon, _events_rx) = Daemon::build_daemon(
             Some("daemon".to_string()),
@@ -8914,21 +8947,24 @@ mod fault_tolerance_tests {
             BTreeMap::new(),
             LogDestination::Tracing,
             None,
+            Vec::new(),
             ZenohBind::Derived(LOCALHOST),
             true,
             false,
         )
         .await
         .unwrap();
-        let (publish_tx, mut publish_rx) = mpsc::channel(1);
-        daemon.zenoh_publish_tx = publish_tx;
 
         let dataflow_id = Uuid::new_v4();
         let node_id = NodeId::from("source".to_string());
         let output_id = DataId::from("value".to_string());
         let output = OutputId(node_id.clone(), output_id.clone());
+        let (publish_tx, mut publish_rx) = mpsc::channel(2);
         let mut dataflow = test_dataflow();
         dataflow.open_external_mappings.insert(output.clone());
+        dataflow
+            .remote_output_queues
+            .insert(output.clone(), publish_tx);
         daemon.running.insert(dataflow_id, dataflow);
 
         daemon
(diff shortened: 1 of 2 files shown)


phil-opp:

I found one significant issue.

Making `close_input` return a `Result` and propagating it with `?` turns routine backpressure / shutdown races into a daemon-wide failure. In `send_output_closed_events` (`binaries/daemon/src/lib.rs:7606`) the local `close_input(dataflow, receiver_id, input_id, &self.clock)?` now goes through `send_input_closed_strict`, which returns `Err` when a downstream node's subscribe channel is full (`Ok(false)`) or closed (`Err(_)`).

When that fires during node-stop handling:
- the loop aborts, so the remaining receivers in `local_node_inputs` get no `InputClosed`;
- the inter-daemon `OutputClosed` sends just below (`lib.rs:7616-7629`) are skipped entirely, so remote consumers' inputs stay open until force-kill — the exact hang the surrounding code warns about;
- the `Err` propagates `send_output_closed_events().await?` (7757) → `handle_outputs_done().await?` (7868) → `handle_node_stop().await?` (8646) → `handle_dora_event().await?` (3184), i.e. out of `run_inner`, tearing down the daemon's main event loop for *every* dataflow on that daemon.

Both trigger conditions look realistic rather than rare:
- `Ok(false)` (channel full): the subscribe channel capacity is `NODE_EVENT_CHANNEL_CAPACITY` (1000), so a merely-slow-but-alive consumer that hasn't drained its queue trips it.
- `Err(_)` (channel closed): during coordinated / crash shutdown a downstream node's receiver is dropped while its `subscribe_channels` entry lingers until its own `NodeStop` is processed; if the producer's stop is handled first, this path fires.

Before this PR these cases were best-effort (drop on full; clean up the stale channel and continue on closed). The PR already keeps `break_input` best-effort — it logs the `send_input_closed_strict` error and continues (`lib.rs` ~9397) — so I'd apply the same treatment to the node-stop broadcast in `send_output_closed_events`: collect and log per-receiver failures but continue the loop, and still emit the remote `OutputClosed`, rather than aborting and propagating up to `run_inner`.

Separately, the added/changed tests only convert existing `close_input(...)` calls to `.unwrap()` on live, unfilled channels — they assert the pre-existing happy path and don't exercise the new full/closed `Err` semantics or the abort-remaining-receivers / skipped-remote-`OutputClosed` behavior, so the most consequential change here is untested (and `.unwrap()` would turn a filled channel into a test panic).

For what it's worth, the inter-daemon `OutputClosed` "required" enqueue path (`enqueue_required_zenoh_outbound`) and its FIFO-ordering reasoning look correct.

---

phil-opp:

Turning per-receiver best-effort delivery into a `?` that aborts a multi-receiver loop introduces three regressions. The key point: `send_with_timestamp` (`binaries/daemon/src/event_types.rs:261`) returns `Err` whenever a receiver's channel has been dropped — i.e. any downstream node that already exited but whose `subscribe_channels` entry has not been reaped yet. That is a routine state, not an exceptional one.

**1. A remote consumer can hang.** `binaries/daemon/src/lib.rs:7642` (`send_output_closed_events`) — `for (receiver_id, input_id) in &local_node_inputs { close_input(...)? }`. Take `A -> {B, C}` plus a remote consumer. B exits first, then A. `local_node_inputs` is a `BTreeSet`, so B is visited first, `send_input_closed_strict` returns `Err`, and `?` returns — C never receives `InputClosed`, and the loop below over `open_external_mappings` never runs, so the inter-daemon `OutputClosed` is never published. The remote consumer hangs until force-kill. That is exactly the failure the surrounding comment was written to prevent. Same shape at `:5102` and at `:4223` (`RemoveNode`, where the abort also skips the receiver-side mapping cleanup and `subscribe_channels.remove`).

**2. A node that should stop gets restarted.** `binaries/daemon/src/lib.rs:9297` (`close_input`) — `send_input_closed_strict(...)?` now short-circuits before `signal_all_inputs_closed_if_drained`. Previously both were attempted independently. A receiver that misses `InputClosed` now also loses `node.disable_restart()` and the `all_inputs_closed_at` bookkeeping.

**3. No new tests.** The only test changes are `.unwrap()` appended to eight existing `close_input` calls. Nothing exercises Reload-on-full-channel, `InputClosed` on a full channel, the `net_publish_failures` counting, or the deferred `OutputClosed` send — every claim in the PR body is untested, and the existing tests pass identically with or without the strict semantics.

Also worth noting: `handle_publish_enqueue_result` now increments `net_publish_failures` for *dropped* (never-attempted) sends, changing that metric's meaning from "zenoh publish errored" to "errored or dropped". It is surfaced to the CLI, so dashboards change meaning silently.

SunSunSun689:
Addressed the review by keeping the input-close broadcast paths best effort instead of aborting the whole loop on a single full/closed channel. In particular, , , and  handling now log and continue, so one failing receiver no longer blocks later receivers or remote close propagation.\n\nAdded regressions for:\n-  returning an error on a full channel\n- closed receiver channels being removed\n- best-effort input closing continuing after the first failure\n\nValidation:\n- \n-

SunSunSun689:
Addressed the review by keeping the input-close broadcast paths best effort instead of aborting the whole loop on a single full/closed channel. In particular, `send_output_closed_events`, `RemoveNode`, and `OutputClosed` handling now log and continue, so one failing receiver no longer blocks later receivers or remote close propagation.

Added regressions for:
- `close_input` returning an error on a full channel
- closed receiver channels being removed
- best-effort input closing continuing after the first failure

Validation:
- `cargo test -p dora-daemon close_input_ -- --nocapture`

phil-opp:
Re-reviewed after the latest commit (`441eb32`). The best-effort wrappers (`close_inputs_best_effort` at the `RemoveNode` / `OutputClosed` / `send_output_closed_events` sites) resolve the earlier daemon-teardown concern — a failed close no longer aborts the event loop — and the three new `close_input_*` tests are a solid, non-vacuous addition. Those earlier findings look fixed.

One residual from the earlier review still appears open, in `close_input`:

```rust
if was_open {
    send_input_closed_strict(dataflow, receiver_id, input_id, clock)?;
}
signal_all_inputs_closed_if_drained(dataflow, receiver_id, clock)
```

When the receiver's subscribe channel is *full* (not closed), `send_input_closed_strict` returns `Err`, and the `?` returns before `signal_all_inputs_closed_if_drained` runs. Before this PR that function ran unconditionally after the InputClosed send, and it does channel-independent bookkeeping — `node.disable_restart()` and the `all_inputs_closed_at` insert. So a slow-but-alive node (the channel holds up to `NODE_EVENT_CHANNEL_CAPACITY` events) whose last input just closed no longer gets `disable_restart()`, which can leave its restart policy active.

`break_input` already avoids this by calling `send_input_closed_strict` and `signal_all_inputs_closed_if_drained` independently (logging each rather than `?`-propagating). Applying the same decoupling in `close_input` — still attempt the drained-signal when the InputClosed send failed — would preserve the pre-PR behavior. None of the new tests covers this (they assert input removal, not that `disable_restart` still fires on a full channel), so a test pinning that would help.

Separately, and more minor: `handle_publish_enqueue_result` / `enqueue_required_zenoh_outbound` now bump `net_publish_failures` on a full/closed *enqueue* (the message was never handed to zenoh), which shifts that counter's meaning from "publish errored" to "errored or dropped before send." Worth a note if anything downstream reads that metric.

---

---

SunSunSun689:
Addressed the remaining close_input regression from the latest review.  now keeps drain bookkeeping best effort: if delivering  hits a full or closed channel, it still runs  so restart disabling and drained-node bookkeeping are not skipped. Added a regression test covering the full-channel case.\n\nValidation:\n- \n- 
running 8 tests
test fault_tolerance_tests::close_input_returns_error_when_receiver_channel_is_full ... ok
test fault_tolerance_tests::close_input_removes_closed_receiver_channel ... ok
test fault_tolerance_tests::close_input_drops_armed_deadline ... ok
test fault_tolerance_tests::close_input_sends_all_inputs_closed ... ok
test fault_tolerance_tests::close_input_removes_from_open_inputs ... ok
test fault_tolerance_tests::close_input_on_already_broken_input ... ok
test fault_tolerance_tests::close_input_still_disables_restart_when_input_closed_notification_is_full ... ok
test fault_tolerance_tests::close_input_deferred_by_broken_inputs ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 230 filtered out; finished in 0.00s

SunSunSun689:
Addressed the remaining `close_input` regression from the latest review.

`close_input` now keeps drain bookkeeping best effort: if delivering `InputClosed` hits a full or closed channel, it still runs `signal_all_inputs_closed_if_drained` so restart disabling and drained-node bookkeeping are not skipped. Added a regression test covering the full-channel case.

Validation:
- `rustup run 1.97.1 cargo fmt --all -- --check`
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc /home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/cargo test -p dora-daemon close_input_ -- --nocapture`

phil-opp:
**Follow-up on the latest commit**

The new commit (07:47Z) resolves the residual issue from the previous review: `close_input` now always runs `signal_all_inputs_closed_if_drained` even when the input-closed notification channel is full (it accumulates the send result instead of returning early with `?`), so `disable_restart` and the drain bookkeeping still fire — matching the pre-PR behavior. The added test `close_input_still_disables_restart_when_input_closed_notification_is_full` pre-fills a capacity-1 channel to force that path, so it genuinely covers the fix. No new issues found in the latest commit.

---
🤖 _This is a fully automated review by Claude. It may contain mistakes — please verify before acting on it._

---

phil-opp:
issues found by claude:

Breaks dora reload on every multi-daemon dataflow. Verified end to end:

binaries/daemon/src/lib.rs:7298 (PR head) replaces main's silent if let Some(channel) = … (lib.rs:5209) with else { return Err(eyre!("node {node_id} not connected")) }.
binaries/coordinator/src/handlers.rs:316 broadcasts ReloadDataflow to every daemon in dataflow.daemons and ?s on the first ReloadResult(Err) (:328-330).
With N≥2 daemons only one hosts the node; the rest now reply Err → reload fails deterministically. Same regression single-daemon once the node has finished and its channel was reaped. No test covers it.
Second high finding: lib.rs:9367+:9387 — the new subscribe_channels.remove() on a closed channel makes signal_all_inputs_closed_if_drained hit its else { return } guard and skip disable_restart(), so a node whose only producer is gone gets respawned with no inputs. break_input has the same hole. This is the same class of bug a previous review round fixed for the full-channel case — only that half was fixed.

Also: benign shutdown races now log as errors; RemoveMapping reports failure for an operation that already succeeded with no rollback; ~120 of the 426 lines are avoidable duplication (try_send_to_remote_receivers is 41 byte-identical lines of send_to_remote_receivers).

phil-opp:
**🤖 Automated review by Claude — this review is fully automated, with no human in the loop.**

I found one additional issue not covered by the previous review. (The earlier findings — multi-daemon `Reload` returning `Err` on non-hosting daemons, and the closed-channel path skipping `disable_restart()` — still apply to the current head, since no commits have landed since that review.)

**New point — the "required" inter-daemon `OutputClosed` delivery guarantee stops at the daemon's local publish channel and does not extend to the wire.** `send_to_remote_receivers` (and the shared per-output publisher cache) declares the publisher with `CongestionControl::Drop` + `express(true)` (`binaries/daemon/src/lib.rs` ~5500). `enqueue_required_zenoh_outbound` only retries when the local `zenoh_publish_tx` mpsc is full; once a message reaches `publisher.put().await`, zenoh's `Drop` congestion control can discard the sample under transport-queue congestion while `put()` still returns `Ok` (and `net_messages_sent` is incremented as if delivered).

Failure scenario: a remote consumer is slow/behind, the zenoh link congests, the producing node closes its output; the daemon's local channel has room so the retry path never triggers, `put()` returns `Ok`, but the sample is dropped on the wire — the remote input stays open until force-kill, i.e. exactly the hang this PR set out to fix.

If `OutputClosed` is genuinely a required lifecycle event, it likely needs a reliable/blocking publisher (or a separate reliably-configured publisher for close events) rather than reusing the drop-configured data-plane publisher; otherwise the added local-retry machinery closes only the "local channel full" gap, not the end-to-end one.

---

SunSunSun689:
Addressed the latest review findings in `7399de780`.

Changes:
- Restored reload as a no-op on daemons that do not host the target node, so multi-daemon reload no longer fails on `node not connected` from non-host daemons. Added `reload_missing_node_channel_is_noop_for_non_host_daemons`.
- Decoupled drain bookkeeping from subscribe-channel liveness. `signal_all_inputs_closed_if_drained` now still disables restart when the receiver channel was closed and removed. Added `close_input_still_disables_restart_when_receiver_channel_is_closed`.
- Kept `RemoveMapping` success independent from best-effort `InputClosed` delivery; a removed mapping now returns `Ok(())` even if the close notification cannot be delivered.
- Changed the shared remote receiver publisher to `CongestionControl::Block`, so required `OutputClosed` events that reach the Zenoh drain task are not sent through a drop-configured publisher.
- Removed the duplicated publisher/outbound construction in the regular and required remote-send paths.

Validation:
- `rustup run 1.97.1 cargo fmt --all -- --check`
- `cargo clippy -p dora-daemon -- -D warnings`
- `cargo test -p dora-daemon` (240 unit tests + doctests)
- `cargo test --test fault-tolerance-e2e -- --test-threads=1` (11 passed)
- `cargo test -p dora-examples --test example-smoke contract_ -- --test-threads=1` (8 passed, with scratch venv containing `pyarrow` and `-e apis/python/node`)
- workspace baseline excluding the Python/runtime/example packages listed in `CLAUDE.md`: passed

phil-opp:
**Automated review by Claude** — this is a fully automated review, with no human in the loop; please verify before acting.

Re-reviewed after `7399de780`, the non-merge commit since the last review.

**Resolved from earlier rounds:** reload is now a no-op on non-host daemons (with a covering test), `signal_all_inputs_closed_if_drained` runs `disable_restart()` before the channel guard so it fires even on closed/removed channels, `RemoveMapping` returns `Ok(())` independent of delivery, and the send-path duplication is largely gone. Thanks for those.

**One new concern with the fix for the wire-drop finding.** To stop `OutputClosed` being dropped on the wire, `remote_receiver_publisher` now declares the publisher with `CongestionControl::Block`. But that publisher is cached per output and shared by both `send_to_remote_receivers` (required `OutputClosed`) and `try_send_to_remote_receivers` (best-effort regular `Output`), and every message funnels through the single shared drain task (`_zenoh_drain_handle`) — one `zenoh_publish_tx` channel (cap 256) per daemon, covering every output of every dataflow — which calls `publisher.put(...).await` serially.

With `Block`, a single slow / absent / congested remote peer can make that `put().await` block indefinitely, head-of-line-stalling the whole drain task. Two consequences:

1. Regular `Output` is no longer best-effort once it is dequeued — it blocks on the wire (only masked by the local 256-channel `try_send` drops), which contradicts the "regular output remains best-effort" intent.
2. `OutputClosed` for *other* outputs/dataflows queued behind the stalled `put` never gets sent, and the retry tasks spawned by `enqueue_required_zenoh_outbound` pile up awaiting the now-perpetually-full channel — recreating the "remote input stays open until force-kill" hang this PR set out to fix, now triggerable by an unrelated congested output.

This is why a *separate* reliably-configured publisher for close events (rather than making the shared data-plane publisher blocking) is the safer shape: e.g. give `OutputClosed` its own `Block` publisher while keeping the data-plane publisher `Drop`, or move the blocking `put` off the single shared drain task (per-output/per-peer draining) so one congested peer can't stall everything. It is not a main-event-loop deadlock (the loop only `try_send`s), but it is a systemic drain-task stall worth addressing before merge.

---

SunSunSun689:
Addressed the latest drain-task stall concern in `bf109e870`:

- Regular inter-daemon `Output` now uses the data-plane publisher cache with `CongestionControl::Drop`, preserving the best-effort wire behavior for normal data.
- Required `OutputClosed` now uses a separate `control_publishers` cache with `CongestionControl::Block`, so close events are not sent through the dropping data publisher.
- `OutputClosed` publishes no longer go through the daemon-wide shared `zenoh_publish_tx` drain. They are spawned onto their own publish task, so a slow/stuck blocking close publish cannot head-of-line stall regular output or close events for unrelated outputs/dataflows in the shared drain.
- Added `output_closed_uses_separate_block_publisher_from_regular_output` to exercise the real `try_send_to_remote_receivers` / `send_output_closed_events` paths and assert that data uses `Drop`, close uses a distinct `Block` publisher, and close does not enter the shared daemon-wide data drain.

Verified with:
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo test -p dora-daemon fault_tolerance_tests -- --nocapture`
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo fmt --all -- --check`
- `RUSTC=/home/dora/.rustup/toolchains/1.97.1-x86_64-unknown-linux-gnu/bin/rustc rustup run 1.97.1 cargo clippy -p dora-daemon -- -D warnings`

phil-opp:
**Automated review by Claude** — this is a fully automated review, with no human in the loop; please verify before acting.

Re-reviewed after the new commit `bf109e8` ("isolate reliable output-close publishes"), which lands after my last review. It correctly resolves the shared-drain head-of-line stall from the previous revision, but in doing so it reintroduces an ordering hazard between regular `Output` and `OutputClosed` — the exact FIFO property the PR set out to preserve.

**The two paths no longer share a queue:**

- Regular data — `send_output` → `try_send_to_remote_receivers` → `remote_receiver_data_publisher` (`CongestionControl::Drop`, cached in `dataflow.publishers`) → `try_enqueue_zenoh_outbound(&self.zenoh_publish_tx, ...)`. The message sits in the shared FIFO drain (`zenoh_publish_tx`, cap 256) and is flushed **serially** by the single `_zenoh_drain_handle` task.
- `OutputClosed` — `send_output_closed_events` → `send_to_remote_receivers` → `remote_receiver_control_publisher` (`CongestionControl::Block`, cached in the *separate* `dataflow.control_publishers` map) → `spawn_required_zenoh_outbound(...)`, which is a bare `tokio::spawn` that calls `publish_zenoh_outbound(...).await` **immediately, bypassing the shared drain**.

Both publishers declare the **same** key (`zenoh_daemon_control_topic(dataflow_id, output.0, output.1)`), but they are two independent `Publisher` objects, and the close task runs concurrently with the drain task. The daemon's main loop enqueues the final data output (into the FIFO) before it handles the output-closed event (which spawns the close task) — so enqueue order is correct — but if the drain has any backlog (slow/absent remote consumer, which is the whole reason this queue exists), the spawned close `put()` can complete before the drain flushes the trailing data. On the receiving daemon that surfaces as `InputClosed` arriving ahead of the producer's last outputs, so a consumer that stops reading on close drops trailing data.

This contradicts the diff's own new comment on the drain (`All events for a given output, including OutputClosed, share this FIFO so close cannot pass data that was produced earlier`) — which is no longer true, since `OutputClosed` no longer goes through `zenoh_publish_tx`. And `output_closed_uses_separate_block_publisher_from_regular_output` asserts the close does *not* enter the shared drain, so the test locks in the ordering-breaking design rather than guarding against it; there is no test covering the data-then-close ordering invariant.

Note #3071 (same author, same underlying fix) takes the opposite approach: it keeps `OutputClosed` on the *same* cached publisher and the *same* FIFO, accepting a bounded (~1s timeout) per-message drain stall to preserve ordering. Only one of the two shapes can be right; the two open PRs currently disagree on the tradeoff. If ordering must hold, the close event needs to stay behind the earlier data for its output (e.g. a per-output/per-peer drain rather than a global one, so one slow peer can't stall everything *and* close can't overtake its own data).

Separately (lower severity): `spawn_required_zenoh_outbound` spawns one task per `OutputClosed` doing a `Block` `put().await` with no timeout — against a gone/congested peer these tasks can block indefinitely and accumulate. Bounded by the number of close events (not per-message), so much milder than the stall it replaced, but still an unbounded-in-principle wait with no upper bound.

The rest of the revision looks sound: the `close_input` / `send_input_closed_strict` / `signal_all_inputs_closed_if_drained` split correctly runs `disable_restart()` bookkeeping before the channel-liveness guard, the non-host reload no-op is right, and the added channel-full / channel-closed tests exercise real paths.

---

SunSunSun689:
Updated this PR to address the review ordering issue.

The new shape uses a per-output publish FIFO. Regular `Output` still uses the `Drop` publisher and best-effort enqueue, while `OutputClosed` uses the separate `Block` publisher but is enqueued into the same FIFO for that `OutputId`. This means `OutputClosed` can no longer overtake regular data that was queued earlier for the same output, while a slow/blocking close only stalls that output instead of a daemon-wide drain.

Also replaced the previous test that locked in the bypass path with coverage for the intended invariant:

- `output_closed_uses_block_publisher_but_shares_output_fifo_with_regular_output`
- `output_closed_stays_behind_queued_regular_output`

Validation run locally with Rust 1.97.1 and `PYO3_PYTHON=/home/dora/.local/bin/python3.12`:

- `cargo test -p dora-daemon`
- `cargo test -p dora-daemon fault_tolerance_tests -- --nocapture`
- `cargo test --all --exclude dora-node-api-python --exclude dora-operator-api-python --exclude dora-ros2-bridge-python --exclude dora-runtime-python --exclude dora-cli-api-python --exclude dora-examples`
- `cargo check --examples`
- `cargo test --test fault-tolerance-e2e -- --test-threads=1`
- `cargo test --test ws-cli-e2e -- --test-threads=1` using an isolated temp `HOME` because the real `/home/dora/.dora/coordinator.redb` is an incompatible old redb v2 store
- `cargo test --test multi-daemon-e2e -- --test-threads=1`
- `cargo test --test daemon-reconnect-e2e -- --test-threads=1`
- `cargo test --test example-smoke -- --test-threads=1` (`65 passed; 9 ignored`, ignored tests are the existing torch/CUDA-gated memory-pool cases)
- `git diff --check`

SunSunSun689:
Follow-up on the relationship with #3071:

#3071 and this PR were both trying to address the same underlying ordering/reliability problem around `OutputClosed`, but they ended up taking different tradeoffs:

- #3071 keeps `OutputClosed` on the shared drain path, which preserves ordering but can still let one slow/blocked remote output delay unrelated outputs.
- The current version of this PR keeps the ordering guarantee without going back to a single global drain: regular `Output` events and the corresponding `OutputClosed` event now share a FIFO per output, while different outputs can drain independently.

So the intended direction is to use the #3215 design as the replacement for #3071, rather than updating #3071 to match independently. I left a note on #3071 as well that it is superseded by this PR.

This also addresses the review concern that the previous #3215 revision let `OutputClosed` bypass queued data: close events are now enqueued behind earlier data for the same output, using the block/control publisher but the same per-output publish queue.

phil-opp:
**Automated review by Claude — this is a fully automated review; no human has vetted it.**

Re-reviewed after `ea29d8a` (per-output publish FIFO), the commit since the last review. This resolves the ordering hazard from the previous revision: regular `Output` (Drop publisher) and `OutputClosed` (Block publisher) for the same `OutputId` now share one per-output FIFO drained serially, so a close can't overtake earlier data for that output, and a slow/blocking close only stalls its own output's drain rather than a daemon-wide one. The two new tests genuinely guard data-then-close ordering and the distinct Drop/Block publishers (note they assert enqueue routing rather than wire delivery, since the injected channel means no drain task actually runs). Earlier-resolved items (non-host reload no-op, `disable_restart` bookkeeping before the channel-liveness guard, best-effort close loops that log-and-continue) still hold.

One residual, now milder than before: the per-output drain calls `control_publisher.put().await` with `CongestionControl::Block` and no timeout, so an `OutputClosed` to a gone/congested peer blocks that output's drain task (and any `enqueue_required_zenoh_outbound` retry task) indefinitely, leaking until the dataflow is removed. It's now bounded to the closing output, so not a blocker — a bounded timeout would make it fail-stop. Also worth a code comment: the ordering guarantee relies on zenoh preserving submission order across the two separate publisher objects on the same key expression; that holds with the serial FIFO + express puts, but it's load-bearing and currently undocumented.

---

SunSunSun689:
Addressed the latest follow-up in `6dd2d0c2b`.

Changes:
- Added a bounded `ZENOH_PUBLISH_TIMEOUT` around each Zenoh `put()` so a gone/congested peer cannot pin a per-output drain task indefinitely.
- Timeout failures are counted in `net_publish_failures`, matching other publish failures, and the per-output drain can continue after the bounded failure.
- Added a code comment documenting the ordering assumption: regular `Output` and `OutputClosed` use distinct Drop/Block publishers, but the per-output FIFO is the ordering boundary because each publish is awaited before the next one is submitted on the same key.
- Added `zenoh_publish_wait_is_bounded`, using paused Tokio time to cover the pending-publish timeout path without sleeping in real time.

Validation run locally with Rust 1.97.1 and `PYO3_PYTHON=/home/dora/.local/bin/python3.12`:

```text
cargo test -p dora-daemon zenoh_publish_wait_is_bounded -- --nocapture
cargo test -p dora-daemon fault_tolerance_tests -- --nocapture
cargo fmt --all -- --check
cargo clippy -p dora-daemon -- -D warnings
cargo test -p dora-daemon
git diff --check
```

phil-opp:

Re-reviewed after `6dd2d0c` ("bound zenoh publish waits"), the commit since the last review. It resolves the two residual points from the previous round:

- Each per-output `put()` is now wrapped in `await_zenoh_put_with_timeout(..., ZENOH_PUBLISH_TIMEOUT)` (1s), so a gone/congested peer can no longer pin an output's drain task indefinitely on the `Block` publisher. Timeouts are counted in `net_publish_failures` like other publish failures and the drain continues to the next message, so it's fail-stop rather than a leak.
- The load-bearing ordering assumption (regular `Output` via the `Drop` publisher and `OutputClosed` via the `Block` publisher share one per-output FIFO, each publish awaited before the next is submitted on the same key) is now documented in a code comment.

`zenoh_publish_wait_is_bounded` uses paused Tokio time to cover the timeout path without real sleeping. The removal of the daemon-wide `zenoh_publish_tx` field has no dangling references — both send paths now route through `remote_output_queues`. No new issues; the latest commit looks safe.

---

phil-opp:
🤖 **Automated review by Claude** — this is a fully automated review; no human has verified these findings, so please treat them as suggestions to check rather than authority.

Re-reviewed the current head (`6dd2d0c`, unchanged since my last review). The previously-resolved items still hold — I'm not reopening those. One point I don't think any earlier round raised:

**The two new per-output maps and their drain tasks are never cleaned up on node removal.** This PR replaces the single daemon-wide zenoh drain task with:
- `control_publishers` and `remote_output_queues` on `RunningDataflow` (`binaries/daemon/src/running_dataflow.rs:370-371`), and
- one drain task spawned per `OutputId` in `remote_output_publish_queue` (`binaries/daemon/src/lib.rs`), whose `Sender` is stored in `remote_output_queues`.

Both maps are only ever inserted into (via the `Entry::Vacant` arms of `remote_receiver_publisher` / `remote_output_publish_queue`); there is no `remove`/`retain` for either on `RemoveNode`, `ReplaceNode`, or any `close_input` path. Because the stored `Sender` is what keeps each spawned task alive (`while let Some(msg) = rx.recv().await`), that task also lives until the whole dataflow is torn down. So the change goes from one daemon-wide task to one retained task + `Block` publisher + channel **per distinct `OutputId` ever seen**.

For a static dataflow this is bounded by the number of outputs and harmless. The concern is dynamic churn — dynamic nodes joining/leaving with fresh node/output ids inside a long-running dataflow — where these accumulate for the dataflow's lifetime.

Worth flagging because this is the same retention pattern the pre-existing `publishers` map has, which sibling PR **#3351** is separately fixing in `forget_node_bookkeeping`. That PR only purges `publishers`/`debug_topic_watchers` and has no knowledge of these two new maps, so if both land the new surfaces still leak. Extending the same node-removal cleanup to `control_publishers` and `remote_output_queues` — and dropping the per-output `Sender` so its drain task exits — would close it. Not a blocker for the common static case.

Everything else in the current revision looks consistent with the resolved state from the earlier review rounds (per-output FIFO ordering, the Drop/Block publisher split, the bounded `ZENOH_PUBLISH_TIMEOUT`, and the `close_input` / `signal_all_inputs_closed_if_drained` bookkeeping split).

---

phil-opp:

This has come a long way over the review rounds and the per-output FIFO is the right shape. **The conflict is drift, not a sign the fix landed elsewhere** — I checked `main` directly: the daemon-wide `zenoh_publish_tx` field and its single drain task are still there (`binaries/daemon/src/lib.rs:456`, `:2029`, `:5663`), `close_input` still returns `()` and swallows send failures (`:7401-7448`), and `send_reload` still has the `Ok(false) => { /* event dropped (channel full) */ }` arm (`:5448-5457`). So please rebase rather than abandon it.

The conflict comes from three merged commits: #3277 rewrote the exact `send_output_closed_events` close loop this patches into a keyed `for output in &outputs` lookup merged into a `BTreeSet`, #3364 changed `build_daemon`'s signature (13 args to 15 — same fix as your #3118 needs), and #3429 edited neighbouring regions. `close_inputs_best_effort` takes `IntoIterator<Item = (NodeId, DataId)>`, so `local_node_inputs` should drop into the keyed version unchanged.

Three things I would like changed on the way through.

**The 1 s publish timeout is applied to regular `Output` too.** `ZENOH_PUBLISH_TIMEOUT` wraps every publish on the per-output drain, including `InterDaemonEvent::Output`, which carries the full payload. A multi-megabyte payload on a slow cross-machine link that needs more than a second to hand off is now dropped and counted in `net_publish_failures` — a new data-loss mode on a path that previously had no deadline. Scoping the timeout to the `Block` (control) publish keeps the fix without it; the constant's own doc comment says the reason is "required lifecycle events use blocking publishers", so the narrower application matches the stated intent.

**The per-output queues each get the full 256-slot capacity** that used to be the daemon-wide budget. Worst-case in-flight buffering goes from 256 serialized messages per daemon to 256 x (number of remote outputs), each holding a payload. `ZENOH_PUBLISH_CHANNEL_CAPACITY`'s doc comment was written for a single daemon-wide queue and was not re-tuned when it became per-output. A much smaller per-output cap (8-32 for a FIFO whose only job is ordering) with 256 kept as the daemon-wide budget would do, or state explicitly why per-output 256 is affordable.

**`control_publishers` and `remote_output_queues` are never removed from.** `forget_node_bookkeeping` (`binaries/daemon/src/running_dataflow.rs:471`) is not extended, and the stored `Sender` is what keeps each spawned drain task alive — so a removed or replaced node leaves behind a `Block` publisher, a channel, and a live tokio task per `OutputId` for the dataflow's lifetime. This was raised last round and is the one finding not yet addressed. #3351 is adding exactly this cleanup for `publishers`, so it is probably easiest to let that land first and then extend `forget_node_bookkeeping` to cover your two maps in the same place.

One smaller thing worth a look: moving `disable_restart()` above the `subscribe_channels` guard was the right response to the earlier review, but that guard also covered "node has not subscribed yet" and "node is mid-restart, its channel was removed on a failed send". `disable_restart` is one-way outside `restart_single_node`, so an upstream close during a restart window can now permanently pin a node's `restart_policy` off. A short test pinning what should happen there would be good.

On tests: the added unit tests do fail without the fix for the error-returning paths, which is solid. What is missing is that `output_closed_uses_block_publisher_but_shares_output_fifo_with_regular_output` and `output_closed_stays_behind_queued_regular_output` inject a `remote_output_queues` entry with no drain task, so they assert enqueue routing rather than wire ordering — nothing exercises the spawned task or the "each publish awaited before the next" claim the ordering guarantee rests on. Nothing covers `net_publish_failures` incrementing either, and the headline symptom (a slow remote peer no longer strands `OutputClosed`) has no coverage in `tests/multi-daemon-e2e.rs`.

If you would rather split the zenoh restructure into its own PR, the reload and input-close halves look ready to go on their own — that would get two thirds of this landed this week.

phil-opp:

This has come a long way over the review rounds and the per-output FIFO is the right shape. **The conflict is drift, not a sign the fix landed elsewhere** — I checked `main` directly: the daemon-wide `zenoh_publish_tx` field and its single drain task are still there (`binaries/daemon/src/lib.rs:456`, `:2029`, `:5663`), `close_input` still returns `()` and swallows send failures (`:7401-7448`), and `send_reload` still has the `Ok(false) => { /* event dropped (channel full) */ }` arm (`:5448-5457`). So please rebase rather than abandon it.

The conflict comes from three merged commits: #3277 rewrote the exact `send_output_closed_events` close loop this patches into a keyed `for output in &outputs` lookup merged into a `BTreeSet`, #3364 changed `build_daemon`'s signature (13 args to 15 — same fix as your #3118 needs), and #3429 edited neighbouring regions. `close_inputs_best_effort` takes `IntoIterator<Item = (NodeId, DataId)>`, so `local_node_inputs` should drop into the keyed version unchanged.

Three things I would like changed on the way through.

**The 1 s publish timeout is applied to regular `Output` too.** `ZENOH_PUBLISH_TIMEOUT` wraps every publish on the per-output drain, including `InterDaemonEvent::Output`, which carries the full payload. A multi-megabyte payload on a slow cross-machine link that needs more than a second to hand off is now dropped and counted in `net_publish_failures` — a new data-loss mode on a path that previously had no deadline. Scoping the timeout to the `Block` (control) publish keeps the fix without it; the constant's own doc comment says the reason is "required lifecycle events use blocking publishers", so the narrower application matches the stated intent.

**The per-output queues each get the full 256-slot capacity** that used to be the daemon-wide budget. Worst-case in-flight buffering goes from 256 serialized messages per daemon to 256 x (number of remote outputs), each holding a payload. `ZENOH_PUBLISH_CHANNEL_CAPACITY`'s doc comment was written for a single daemon-wide queue and was not re-tuned when it became per-output. A much smaller per-output cap (8-32 for a FIFO whose only job is ordering) with 256 kept as the daemon-wide budget would do, or state explicitly why per-output 256 is affordable.

**`control_publishers` and `remote_output_queues` are never removed from.** `forget_node_bookkeeping` (`binaries/daemon/src/running_dataflow.rs:471`) is not extended, and the stored `Sender` is what keeps each spawned drain task alive — so a removed or replaced node leaves behind a `Block` publisher, a channel, and a live tokio task per `OutputId` for the dataflow's lifetime. This was raised last round and is the one finding not yet addressed. #3351 is adding exactly this cleanup for `publishers`, so it is probably easiest to let that land first and then extend `forget_node_bookkeeping` to cover your two maps in the same place.

One smaller thing worth a look: moving `disable_restart()` above the `subscribe_channels` guard was the right response to the earlier review, but that guard also covered "node has not subscribed yet" and "node is mid-restart, its channel was removed on a failed send". `disable_restart` is one-way outside `restart_single_node`, so an upstream close during a restart window can now permanently pin a node's `restart_policy` off. A short test pinning what should happen there would be good.

On tests: the added unit tests do fail without the fix for the error-returning paths, which is solid. What is missing is that `output_closed_uses_block_publisher_but_shares_output_fifo_with_regular_output` and `output_closed_stays_behind_queued_regular_output` inject a `remote_output_queues` entry with no drain task, so they assert enqueue routing rather than wire ordering — nothing exercises the spawned task or the "each publish awaited before the next" claim the ordering guarantee rests on. Nothing covers `net_publish_failures` incrementing either, and the headline symptom (a slow remote peer no longer strands `OutputClosed`) has no coverage in `tests/multi-daemon-e2e.rs`.

If you would rather split the zenoh restructure into its own PR, the reload and input-close halves look ready to go on their own — that would get two thirds of this landed this week.

phil-opp:
🤖 **Automated review** — *this is a fully automated Claude Code review; no human has vetted it.*

One new finding, not raised in earlier rounds (the diff is unchanged; this surfaces from re-checking the branch against current `main`): the new tests call `build_daemon(...)` with **13** arguments, but `build_daemon` on current `main` (`lib.rs:1930`) takes **15** parameters — it gained `requested_listen_endpoint: Option<String>` and `zenoh_discovered_connect: Vec<String>`. The new tests (`reload_missing_node_channel_is_noop_for_non_host_daemons`, `output_closed_uses_block_publisher…`, `output_closed_stays_behind_queued_regular_output`) also put a `ZenohBind` value in the `requested_listen_endpoint: Option<String>` slot, so it's both an arity and a type mismatch — the test module will not compile against current `main`.

The branch hasn't been rebased since ~Aug 24, so it needs a rebase + test-signature fix before it can build/be evaluated by CI.



---

phil-opp:
needs rebase

SunSunSun689:
Resolved the rebase conflicts against current main and pushed commit 142fbc16f.

What changed:
- kept the new daemon module split from main and ported the PR logic into local_delivery/node_events/running_dataflow
- removed the daemon-wide Zenoh publish queue; remote outputs now use per-output FIFO queues with capacity 32
- regular Output uses Drop publishers without a publish timeout; OutputClosed uses a Block publisher with the bounded timeout and shares the same per-output FIFO so it cannot pass earlier data
- close_input now reports control-event delivery failures, while batch close paths use best-effort handling
- forget_node_bookkeeping now clears data/control publishers and remote output queues for removed node outputs
- added regression coverage for close_input failures/best-effort close, OutputClosed FIFO/publisher behavior, publish timeout, and remote output queue cleanup

Validation run locally:
- cargo test -p dora-daemon -- --nocapture
- cargo fmt --all -- --check
- cargo clippy -p dora-daemon -- -D warnings

phil-opp:

Re-reviewed after the `142fbc16` rebase. This resolves the previously-reported blockers:

- **Compiles against current `main` now.** The test passes the full 15-argument `build_daemon` signature (incl. `requested_listen_endpoint` / `zenoh_discovered_connect`) in the correct slots; the diff applies cleanly against main's current `node_events.rs` / `coordinator_events.rs` / `local_delivery.rs`.
- **All three earlier findings resolved:** (1) the 1s timeout is now scoped to the control channel only (`try_send_to_remote_receivers` passes `timeout=None` for regular data; only `send_to_remote_receivers` uses `Some(ZENOH_PUBLISH_TIMEOUT)`); (2) the per-output queue cap is reduced to `ZENOH_OUTPUT_QUEUE_CAPACITY = 32`; (3) `forget_node_bookkeeping` now `retain`s other nodes on both `control_publishers` and `remote_output_queues`, with a covering test assertion.

The core (Block publisher + per-output FIFO + bounded put + required-enqueue retry) is sound and well-tested. No blocking issues found at this head.

One minor behavior note (not a blocker): `send_reload` now returns `Err` when the channel is full (`Ok(false)`), turning main's best-effort silent reload-drop into a hard RPC error. That only bites under a rare backlog, but it is a behavior change worth a deliberate decision.

---

phil-opp:
🤖 _This is a fully automated review by Claude. No human has reviewed these findings._

Re-reviewed after the rebase to `142fbc16`. The one open item from the last pass — `control_publishers` and `remote_output_queues` (and their per-output drain tasks) never being reclaimed on node removal — now looks resolved: `forget_node_bookkeeping` runs `retain_other_nodes` over both maps, dropping the stored `Sender` so the drain task exits, and `forget_node_bookkeeping_purges_only_that_node` covers `remote_output_queues`. The per-output FIFO ordering (Drop data vs Block close on one queue), the bounded `ZENOH_PUBLISH_TIMEOUT`, and the `disable_restart`-before-channel-guard split all still hold. No new issues found.

---

phil-opp:

Thanks for the rebase and for working through the last round. The branch builds and the daemon tests pass on current `main`, and the earlier points are addressed: the timeout is now limited to the control publish, the per-output queue is down to 32, and `forget_node_bookkeeping` clears `control_publishers` and `remote_output_queues` with a test for it. The `send_reload` full-channel error goes back through the RPC reply rather than the event loop, which is fine.

One problem with the timeout: it can't fire. In zenoh 1.10.1, `PublicationBuilder<&Publisher, Put>` implements `IntoFuture` as `std::future::ready(self.wait())` (`zenoh/src/api/builders/publisher.rs`), so in `await_zenoh_put_with_timeout` the put runs synchronously inside `publish.into_future()`, before `tokio::time::timeout` has a future to time. With the `Block` publisher, the drain task then holds a tokio worker thread for up to zenoh's `wait_before_close` (5 s by default), after which zenoh closes the transport to that peer (see the note in `libraries/extensions/tensor-pool/examples/README.md`). On a daemon started with `--worker-threads 1`, or when several remote outputs close at once, that can stall the event loop. `zenoh_publish_wait_is_bounded` passes only because `std::future::pending` is a real async future, which a zenoh put isn't.

Two ways to fix it: run the blocking put in `tokio::task::spawn_blocking` and apply the timeout to the join handle (nothing is published after `OutputClosed` on that output, so a timed-out put can't reorder anything), or drop `ZENOH_PUBLISH_TIMEOUT`, document that the bound is zenoh's `wait_before_close`, and still move the `Block` put off the async worker. Either way, replace the `pending()`-based test with one that exercises the real publish path, or remove it.

Smaller points still open from last time, not blocking: `output_closed_uses_block_publisher_but_shares_output_fifo_with_regular_output` injects a queue with no drain task, so it checks queue routing rather than wire order; nothing covers `net_publish_failures` being incremented; and running `disable_restart()` before the `subscribe_channels` check changes what happens to a node that is mid-restart when its last upstream closes, which could use a test.

phil-opp:
Re-reviewed at `142fbc16` (unchanged since the last review). I'm correcting a conclusion from that earlier pass: I now believe **`ZENOH_PUBLISH_TIMEOUT` can never fire against a real zenoh publish**, so the "bounded put" it is meant to guarantee doesn't actually exist.

In the pinned zenoh 1.10.1, the put builder's `IntoFuture` is eager:

```rust
// zenoh/src/api/builders/publisher.rs:565
fn into_future(self) -> Ready<...> { std::future::ready(self.wait()) }
```

`wait()` calls the **synchronous** `Session::resolve_put(...) -> ZResult<()>` (`session.rs:2503`, not an `async fn`), which under `CongestionControl::Block` blocks the calling thread on a full transmit queue.

In `await_zenoh_put_with_timeout` (`binaries/daemon/src/node_events.rs:1146`):

```rust
tokio::time::timeout(timeout, publish.into_future()).await
```

the `publish.into_future()` argument is evaluated *first* — running (or blocking on) the entire put synchronously — before `tokio::time::timeout` is even constructed. `timeout` then wraps an already-resolved `Ready`, so it always returns `Ok(...)` and never yields `Elapsed`. A stuck `Block` put therefore still holds the per-output drain task forever, which is exactly the failure the timeout is documented to prevent.

The regression test `zenoh_publish_wait_is_bounded` (`node_events.rs:1319`) masks this: it feeds `std::future::pending()`, whose `IntoFuture` is genuinely `Pending`, so the timeout fires — but a real put's future is `Ready`, not `Pending`. The test exercises a future shaped nothing like a zenoh put, so it passes trivially while the real mechanism is inert.

To actually bound a synchronous put you'd need to run it off the async executor (e.g. `tokio::task::spawn_blocking` with the timeout on the join handle) rather than wrapping its `IntoFuture`. As written, the `timeout` field and its test could be deleted with zero behavioral change — which is itself a good indicator the mechanism doesn't do anything.

---
🤖 Fully automated Claude Code review — no human has vetted this; advisory only, not an approval, and I can't merge. Reported like an outside contributor.


---

phil-opp:
🤖 **Automated review** (fully automated Claude Code review; no human vetted this — advisory only).

The head is still `142fbc16`, unchanged since the last reviews. I found one new issue.

**The rebase dropped 5 of this PR's regression tests.** They exist in the pre-rebase commit `6dd2d0c2`, but none of them is in `142fbc16`, under any name:

- `close_input_still_disables_restart_when_input_closed_notification_is_full`
- `close_input_still_disables_restart_when_receiver_channel_is_closed`
- `close_input_removes_closed_receiver_channel`
- `reload_missing_node_channel_is_noop_for_non_host_daemons`
- `output_closed_stays_behind_queued_regular_output`

The behavior they pinned is still in the diff, but nothing tests it any more:
- `close_input` still runs `disable_restart()` when the `InputClosed` send hits a full or closed channel.
- `send_input_closed_strict` still removes a closed channel.
- `send_reload` is still a no-op on a daemon that doesn't host the node.
- `OutputClosed` is still ordered behind queued regular output.

Each of these was fixed after an earlier review finding. Without the tests, a later change could bring those bugs back without CI failing. Please restore the tests in the pending rebase. The only conflict against current `main` is in `binaries/daemon/src/node_events.rs`.

The earlier open points still apply:
- The zenoh put's `IntoFuture` runs eagerly, so the timeout can never fire.
- The `pending()`-based test only passes because of that input.
- The split suggestion.

---

phil-opp:
🤖 *This is a fully automated review by a scheduled Claude check. No human vetted it, and it may contain mistakes.*

I found two issues that I don't think earlier rounds covered. The head is still `142fbc16`.

**1. Nothing acts on the new `Result` from `close_input`, so a local `InputClosed` on a full channel is still lost.**
- Every production caller discards the error or only logs it:
  - `coordinator_events.rs:847` and `:1385` use `let _ = close_input(...)`.
  - `close_inputs_best_effort` (`local_delivery.rs:430`) and `break_input` only call `tracing::warn!`.
- Scenario: a consumer's event channel is full when its upstream finishes. `InputClosed` is dropped, `AllInputsClosed` is dropped for the same reason, and the node waits forever on an input that never closes. That is the same outcome as on `main`; the only difference is a warning in the log.
- The remote `OutputClosed` path got a retained send (`enqueue_required_zenoh_outbound`), but the local path did not. So the `Result` plumbing in `local_delivery.rs` currently has the same effect as a `warn!` in the existing `Ok(false)` arm.
- Suggestion: either give local `InputClosed` / `AllInputsClosed` a real retained delivery, or drop the plumbing and just log. The claim in the PR description that "the caller can observe failures" doesn't match the code.

**2. The regular remote-output hot path now clones `OutputId` several times per message.**
- On `main`, `send_to_remote_receivers` deliberately looks up `publishers.get(output_id)` by reference, so the key isn't cloned on every message (the code has a comment saying so).
- With this PR, every inter-daemon `Output` goes through `publishers.entry(output_id.clone())` (`node_events.rs:1055`) and `remote_output_queues.entry(output_id.clone())` (`:1079`), plus a `Sender` clone.
- Looking both maps up with `get` first, and cloning only on a miss, would restore the behaviour on `main`.

The points from the last round are still unaddressed: the 5 dropped regression tests, and zenoh's eager `IntoFuture` making `ZENOH_PUBLISH_TIMEOUT` a no-op. The PR also needs a rebase.

---

phil-opp:
🤖 **Automated review by Claude.** This was fully automated (Claude Code); no human read it before posting. It is advisory only.

I re-reviewed head `142fbc1` and found one new issue.

**The rebase merge (`142fbc16f`) dropped five regression tests that were added in earlier review rounds.** Main has since moved the daemon tests out of `lib.rs`. The tests that lived in `binaries/daemon/src/lib.rs` at `6dd2d0c2b` were not carried over into the new layout, and none of these names exists anywhere at the PR head:
- `reload_missing_node_channel_is_noop_for_non_host_daemons`
- `close_input_still_disables_restart_when_receiver_channel_is_closed`
- `close_input_still_disables_restart_when_input_closed_notification_is_full`
- `close_input_removes_closed_receiver_channel`
- `output_closed_stays_behind_queued_regular_output`

The behaviour these tests covered is still in the diff, but nothing tests it any more:
- `send_reload` returns `Ok(())` for a node with no channel. `send_reload` now has no test at all.
- `disable_restart()` runs before the channel check. Every remaining `close_input` test registers a live channel, so none reaches the missing or closed channel path.

Two regressions could therefore come back with a green test suite:
- The multi-daemon reload failure: every non-host daemon returns an error.
- Nodes with no producers being restarted.

The tests should be restored in the new test module.

The earlier findings are also still open: the zenoh publish timeout can never fire, the ordering and metric tests are missing, `node_events.rs` still conflicts with main, and the two subsystems are still bundled in one PR.

---

phil-opp:
🤖 **Automated review by Claude.** This review was fully automated (Claude Code) and no human checked it before posting, so treat it as advisory.

I re-reviewed head `142fbc1`. The earlier open items still apply, and the merge conflict in `binaries/daemon/src/node_events.rs` is still there. I found two new issues.

**1. Messages on different outputs of the same node can now reach a remote daemon in a different order than they were sent.**
- **Before:** on `main`, every inter-daemon `Output` goes through one drain task for the whole daemon, one `put()` at a time. So if node A sends `a/x` and then `a/y`, a remote consumer receives them in that order.
- **After:** `remote_output_publish_queue` (`node_events.rs:1075`) starts a separate drain task per `OutputId`. On the daemon's multi-thread runtime these tasks run in parallel, so `a/x` and `a/y` race to the transport and can arrive swapped.
- **Why it matters:**
  - Local delivery still keeps the cross-output order, because each receiver has a single FIFO channel.
  - So a consumer that works on one machine can behave differently when it is moved to another.
  - The ordering guarantee in the PR covers only one output's data followed by its own close.
- **Related case, `ReplaceNode`:** `forget_node_bookkeeping` drops the old queue's sender, but that queue's drain task keeps running. The replacement gets a new queue for the same key. So the old incarnation's last messages can arrive after the new incarnation's first ones.
- **How I checked:** by reading the code on `main` and on this PR. I did not reproduce the reordering at runtime.
- **Possible directions:** document this as an intentional change, or keep a single drain task per sending node rather than per output.

**2. The rebase onto current `main` will silently undo #3590's backpressure accounting and cut burst capacity.**
- **Accounting:**
  - `main` now counts drops on the remote publish path with `ft_stats.record_drop(1, dataflow.remote_backpressured_outputs.contains(output_id))` (`node_events.rs:603` on `main`).
  - That call also feeds the lost-backpressure counter.
  - This PR's `handle_publish_enqueue_result` has no `ft_stats`, so resolving the conflict mechanically loses that accounting.
- **Capacity:**
  - On `main`, a remote output that needs lossless backpressure shares 256 queue slots with the rest of the daemon.
  - Here it gets its own 32 slots (`ZENOH_OUTPUT_QUEUE_CAPACITY`).
  - That leaves 8x less room for a burst before a lossless output starts dropping messages.
- **Worth doing:** check both points while resolving the conflict.

---

PR #3216, opened 2026-08-18
ci: auto-manage a needs-rebase label for conflicting PRs

phil-opp:
Labels a PR `needs-rebase` when it conflicts with `main`, and removes the label once it merges cleanly again — whether the author rebased or someone else's merge into main resolved the overlap.

Triggers:

- `pull_request_target` (opened / reopened / synchronize) — checks just that PR
- `push` to `main` — rescans every open PR; this is what clears the label when a *different* PR's merge resolves the conflict
- daily cron + `workflow_dispatch` — safety net for runs lost to API errors or missed events

GitHub computes mergeability lazily, so `UNKNOWN` is re-polled (7 × 12s) rather than read as "no conflict"; a PR still undecided after that keeps its current label instead of being flipped on a guess. A push to main also doesn't invalidate mergeability atomically with the ref update, so the sweep waits 15s before its first query — otherwise it can read the pre-push answer and clear a label the push just earned. Re-polls on a sweep reuse the paginated bulk query instead of one request per undecided PR.

API-only: no checkout, no build, `pull-requests: write` and nothing else. No PR code is fetched or executed, which is what makes `pull_request_target` safe here (it's needed so fork PRs get labelled too). `needs-rebase` isn't in the repo's label list yet, so the job creates it on first use.

Validated with `actionlint`, plus the embedded script run against a mocked GitHub API: add / remove / no-op, late-arriving mergeability, never-resolved PRs, PRs targeting a non-default branch, single-PR events, closed-mid-run, first-use label creation, and a 404 race on removal.

`waiting-for-author` overlaps a little in meaning, but a dedicated label keeps the automation from ever fighting a human-set one.

### DIFF
--- a/.github/workflows/needs-rebase.yml
+++ b/.github/workflows/needs-rebase.yml
@@ -1,0 +1,208 @@
+# Keeps a `needs-rebase` label in sync with GitHub's merge-conflict state.
+#
+# The label is added when a PR conflicts with the default branch and removed
+# as soon as it stops conflicting — whether the author rebased (handled by the
+# `pull_request_target` trigger) or someone else's merge into main resolved the
+# overlap (handled by the `push` trigger, which rescans every open PR).
+#
+# GitHub computes mergeability lazily: a freshly invalidated PR reports
+# `UNKNOWN` until a background job finishes, so the script re-polls the
+# undecided PRs instead of treating `UNKNOWN` as "no conflict". PRs that are
+# still undecided when the retry budget runs out keep whatever label state
+# they had — a stale label is better than a wrong one, and the next push or
+# the daily sweep will settle it.
+name: Needs rebase
+
+on:
+  push:
+    branches: [main]
+  pull_request_target:
+    types: [opened, reopened, synchronize]
+  # Safety net for runs lost to API errors or missed events.
+  schedule:
+    - cron: "17 5 * * *"
+  workflow_dispatch:
+
+# `pull_request_target` runs in the context of the base repo, so the token has
+# write access even for PRs from forks. No code from the PR is checked out.
+permissions:
+  pull-requests: write
+
+concurrency:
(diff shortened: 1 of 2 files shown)


phil-opp:

No issues found. The workflow is API-only (no PR checkout, `pull-requests: write` only), so the `pull_request_target` trigger is safe here. The lazy-mergeability handling looks correct: `UNKNOWN` is re-polled rather than treated as "no conflict", the push sweep waits out the non-atomic mergeability invalidation before its first query, and `removeLabel` tolerates the 404 race. Concurrent per-PR and repo-sweep runs can both write one PR's label, but the operations are idempotent and 404-tolerant and self-correct on the next event.

One thing worth confirming (not clearly a bug): the `schedule` / `workflow_dispatch` paths read `context.payload.repository.default_branch`. That field is normally populated on scheduled runs, but it's the one path not covered by the mocked-API validation, so worth a sanity check that the sweep doesn't silently no-op there.

---

phil-opp:

The workflow will fail on its first run.

`.github/workflows/needs-rebase.yml:33` declares `permissions: pull-requests: write`, but `ensureLabelExists()` (`:157-171`) calls `github.rest.issues.getLabel` / `createLabel` — those are repository-label endpoints (`GET`/`POST /repos/{owner}/{repo}/labels`) and need the **Issues** permission. `pull-requests: write` only covers the PR-scoped `issues/{n}/labels` endpoints.

Since the PR body says the `needs-rebase` label does not exist yet, the first conflicting PR after merge hits `getLabel` → **403**, and the `if (error.status !== 404) throw error` guard rethrows rather than creating it. Every subsequent run fails the same way.

Fix: add `issues: write` to the `permissions` block. The repo already does this in `labeler.yml:10` and `nightly.yml:60` for the same reason.

PR #3217, opened 2026-08-18
fix(ci): grant `issues: write` to the needs-rebase label workflow

phil-opp:
Follow-up to #3216. `ensureLabelExists` calls the repository-label endpoints (`GET`/`POST /repos/{owner}/{repo}/labels`), which need the Issues permission — `pull-requests: write` only covers the PR-scoped `issues/{number}/labels` endpoints. Since the `needs-rebase` label does not exist yet, the first conflicting PR gets a 403 from `getLabel` rather than a 404, and the `status !== 404` guard rethrows instead of creating it. The job then fails on that run and every run after.

### DIFF
--- a/.github/workflows/needs-rebase.yml
+++ b/.github/workflows/needs-rebase.yml
@@ -26,6 +26,14 @@ on:
 # `pull_request_target` runs in the context of the base repo, so the token has
 # write access even for PRs from forks. No code from the PR is checked out.
 permissions:
+  # `issues: write` is required in addition to `pull-requests: write`:
+  # `ensureLabelExists` calls `issues.getLabel` / `issues.createLabel`, which are
+  # the repository-label endpoints (`GET`/`POST /repos/{owner}/{repo}/labels`).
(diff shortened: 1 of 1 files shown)


phil-opp:
**No issues found.** I verified the fix against the workflow script: `ensureLabelExists` calls `github.rest.issues.getLabel` and `issues.createLabel` (`.github/workflows/needs-rebase.yml:155,158`), which are the repository-label endpoints (`GET`/`POST /repos/{owner}/{repo}/labels`) governed by the Issues permission — `pull-requests: write` alone does not cover them. Since the `needs-rebase` label does not exist yet, the first `getLabel` call would return 403 (not 404) without this scope, and the `if (error.status !== 404) throw error` guard at line 157 would rethrow rather than fall through to `createLabel`. Adding `issues: write` is the correct fix. No repository code is checked out in this `pull_request_target` workflow, so the added scope doesn't widen fork-PR exposure.

---

---

PR #3218, opened 2026-08-18
ci: keep one live trunk merge-failure notification per PR

phil-opp:
The `trunk-merge-failure-notify` workflow posts a new comment each time Trunk's sticky merge-status comment enters the failed state, so authors get an email that a comment edit would not send. Nothing ever retired those comments, so a PR requeued several times accumulates one per attempt — #3213 collected seven, all but the last stale.

This collapses the notifications from earlier runs with `minimizeComment` (classifier `OUTDATED`) before posting a new one, leaving exactly one visible. Notifications posted before this change are matched on their wording, so the pile-ups already on open PRs get cleaned up on the next status flip.

Re-notification is gated on the **run identity** Trunk embeds in the render (test-branch UUID + head SHA) rather than on the failed-state transition:

| Sticky-comment edit | Behavior |
|---|---|
| → failed, new run id | hide previous, post fresh |
| failed → failed, same run id | no-op |
| failed → failed, different run id | hide previous, post fresh |
| failed → failed, no identifiable run | hide previous, post fresh |
| → not failed | hide all, post nothing |

In practice Trunk re-renders to "Running tests" between attempts, so a transition test would already catch each new run — but that's undocumented behavior. Comparing run ids also covers a new run that never leaves the failed state, while still suppressing re-renders of a single failure. A render with no identifiable run falls through to notifying: a duplicate email beats a silently missed failure.

Verified by simulating nine sticky-comment transitions against the real render strings observed on this repo, with the `isFailed`/`runIds` helpers extracted from the workflow so the test can't drift from the implementation. That simulation caught one bug during development: the constant queue UUID in the `app.trunk.io` details link made the run-id set non-empty for *any* render, which silently reclassified an unidentifiable new failure as a re-render. It's now stripped before comparing.

Two known limits, both documented in the file: the state test is a substring match on Trunk's prose (`failed`), which is empirically specific today — the unqueued boilerplate says "fails"/"failure" and the merged render says "Merged successfully" — but is the line that breaks if Trunk rewords. And no failed render survives in the API to copy verbatim, since the sticky comment is edited in place and GitHub exposes no comment history, so the assumption that it embeds the attempt UUID/SHA is unverified; if it doesn't, behavior degrades to notifying on every failed edit rather than going silent.

### DIFF
--- a/.github/workflows/trunk-merge-failure-notify.yml
+++ b/.github/workflows/trunk-merge-failure-notify.yml
@@ -1,41 +1,155 @@
-# Posts a *new* comment on a PR when Trunk's sticky merge-status comment
-# transitions into a failed state. Trunk edits one sticky comment as the
-# batch moves through queued/testing/failed/merged states; GitHub does
-# not send email on comment edits, so authors silently miss merge
-# failures. Reacting to the edit and posting a fresh @-mention restores
-# the email notification.
+# Keeps exactly one visible "Trunk merge queue failed" notification per PR,
+# and refreshes it once per failing run.
+#
+# Trunk edits a single sticky comment as the batch moves through
+# queued/testing/failed/merged. GitHub sends no email on comment edits, so
+# authors silently miss merge failures — hence this workflow posts a fresh
+# @-mention (which does email) for each failing run, and collapses the
+# notification it posted for the previous run so only the newest is visible.
+#
+# Observed sticky-comment renders on this repo (2026-08-18):
+#
+#   unqueued  "Merging to `main` in this repository is managed by Trunk. ...
+#              If the PR fails, failure details will also be posted here"
+#   testing   "🧪 Running tests on this pull request ([test branch](.../
+#              trunk-merge/pr-3213/605d4cad-...) and [its SHA](.../commit/
+#              9cb743b4...)) - [details](https://app.trunk.io/...)"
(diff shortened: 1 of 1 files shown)


phil-opp (commented):

No issues found.

I traced the workflow's state machine against the documented table and it holds:
- Same run re-rendered while failed → early-return no-op (guarded by `ids !== '' && ids === runIds(prev)`), so no duplicate email.
- Failed → failed with a different/absent run id, or a fresh transition into failed → hide prior notifications, then post one.
- Transition out of failed → hide all, post nothing.

The supporting details are sound: `viewerDidAuthor` scopes the collapse to this workflow's own comments (a human quoting the text is never minimized); the `app.trunk.io` queue-UUID strip before `runIds` correctly prevents the constant queue id from masking an unidentifiable run (the fallback-to-notify path stays reachable); and the empty-id-set fallthrough favors a duplicate email over a missed failure. The `if:` gate intentionally duplicating the `failed` substring test to avoid spinning a runner per edit is reasonable and called out.

The residual risks are the ones the PR already documents (the `failed` substring depends on Trunk's prose, and the per-attempt UUID/SHA embedding is unverifiable from the API). One minor robustness note, not blocking: the GraphQL query fetches `comments(last: 100)`, so on a PR with >100 comments an older stale notification could escape collapsing.

---

PR #3219, opened 2026-08-18
refactor(tensor-pool)!: put the cross-machine transport back behind the extension seam

phil-opp:
#3079 landed the cross-machine memory-pool work directly in dora's core, three
days after #3152 had moved the tensor pool behind an extension seam whose
contract was that no transport's architecture gets frozen into dora. It undid
that contract point by point: 2.6k lines and 47 `unsafe` sites into
`daemon/src/lib.rs`, twelve pool-named variants onto the 1.0 wire protocol, the
`tensor-pool` feature deleted and the dependency made unconditional, and the
Python wheel's flag flipped default-on. The tensor-pool README went on claiming
the transport reached dora "through the public extension channel and nothing
else", and documenting a `--features tensor-pool` build that no longer existed.

This restores the seam without giving up the feature. The transport keeps its
measured behaviour; what changes is where it lives and what dora has to know.

| | main | here |
|---|---|---|
| `daemon/src/lib.rs` | 14,201 lines | 10,072 |
| `unsafe` in `dora-daemon` | 54 | 7 (the pre-#3079 platform sites) |
| pool-named wire variants | 12 | 0 |
| `dora-tensor-pool` in a default `cargo tree` | yes | no |

Five commits, each self-contained:

1. **Extract** the transport out of `lib.rs` into modules, split by plane. Pure
   code motion.
2. **Delete** three dead wire variants. `RegisterPinnedMemory` was never sent
   and never dispatched; `ReadPinnedMemory` was sent by a method with no callers
   and could only ever come back as "unsupported request".
3. **Collapse** the protocol onto three opaque carriers — `ExtensionMessage`,
   `ExtensionRequest`, `ExtensionReply` — plus a per-namespace topic. The pool's
   own messages become `dora_tensor_pool::protocol`, postcard-encoded into the
   payload. dora now reads two fields: `namespace`, to pick an extension, and
   `target_machine`, to drop a broadcast on daemons it does not name. That gate
   was duplicated in two handlers before; it is one generic check now.
4. **Gate** it behind `tensor-pool`, off by default, with the six `Daemon` pool
   fields collapsed into one `PoolState` so the gate sits at module granularity.
   Reverts the Python default flip; adds a `dora-cli` pass-through, since the CLI
   hosts the in-process daemon and the feature was otherwise unreachable from
   `dora run`.
5. **Move** the daemon half into `libraries/extensions/tensor-pool/src/daemon/`,
   reaching dora through a seven-member `DaemonServices` trait. Two members are
   why an extension needs a daemon at all — a crashed node cannot withdraw its
   own descriptor (#2881); the rest is plumbing it cannot reach itself. dora's
   side is one file, `binaries/daemon/src/pool_extension.rs`.

Two incidental fixes fall out of commit 3: a cross-machine tensor frame is no
longer serialized one `u8` at a time (the #3156 treatment, which
`MemoryPoolWrite.tensor_data` never got) — commit 6 applies it at both encode
layers, since the frame is encoded into `PeerMessage::Write` and again into the
envelope; and `NodeRequest::Write` drops `tensor_data`, whose only callers
passed `Vec::new()` — the daemon reads the sender's segment itself, so the
payload-carrying branch was unreachable.

Commit 6 is the outcome of reviewing the first five, and also adds the two
wire-compatibility tests the new serde form was missing: byte-identical to the
sequence encoding across postcard's varint boundaries, and a JSON round trip.
Without them the "pure speedup, no wire break" claim was untested for the form
now carrying three protocol variants.

Because `cargo test --all` does not enable opt-in features, the extension's 18
daemon-half tests would have silently stopped compiling. Two targeted CI steps
keep them honest (~5s each on a warm target dir). The test step covers the
daemon's feature config too, so the `ws_reader_routes_replies_to_pending_daemon_requests`
pin added by #3170 keeps running: the routing block it guards is generic and
ungated, but `resolve_machine` — the only way to reach it — is gated, so the
test is gated the same way. The GPU paths remain
uncovered, as before — dora's CI has no GPU runners.

Local: fmt, workspace clippy `-D warnings`, workspace tests, `cargo check
--examples`, and clippy/tests in both feature configurations. 41 tests in
`dora-tensor-pool --features daemon`, 20 without; 232 daemon tests with the
feature, 231 without (the gated pin above is the difference).
The only failures are `rmw_zenoh_pubsub`'s two peer-discovery tests, which time
out for lack of multicast in this dev container and live in a crate this PR does
not touch.

The design questions #1872 raised about this architecture are still unanswered,
and this PR does not try to answer them. It only puts them back on the
extension's side of the line, where #3152 had them.


phil-opp (commented):

No blocking issues found. This is a large refactor but I focused on the substantive (non-code-motion) parts and they hold up:

- **Protocol collapse** (`daemon_to_daemon`/`node_to_daemon`/`daemon_to_node`): the twelve pool-named variants reduce cleanly to `ExtensionMessage`/`ExtensionRequest`/`ExtensionReply`, `ExtensionReply` is appended last so `ExtensionValue`/`Empty` indices don't shift for version-skewed Python nodes, and the `size_hint`/`bulk` arms are updated consistently.
- **`bulk_bytes::vec`** matches the existing `AVec` helper and the two new tests assert byte-identical postcard encoding across varint boundaries plus a JSON `visit_seq` round trip — so the "pure speedup, no wire break" claim is now actually covered.
- **Feature-off paths don't hang**: a peer `ExtensionMessage` is debug-logged and dropped; a node `ExtensionRequest` gets `DaemonReply::Result(Err("no extension registered…"))`, which the node's `extension_request` maps to `bail!` via the existing `Result(Err)` arm — so the node fails loudly rather than awaiting a reply that never comes.
- **`target_machine` routing faithfully reproduces the original**: `Register`/`Free` are published with `Some(machine)` and gated by the generic check; `Write`/`WriteAck`/`RegisterAck` are broadcast with `None` and filtered by `is_cross` / the pending-seq / pending-register registries — same selection the old per-variant handlers did, just centralized.

The two dead variants removed in commit 2 (`RegisterPinnedMemory`, `ReadPinnedMemory`) were indeed unreachable (no sender / no dispatch arm). I did not exercise the GPU/cross-machine transport at runtime (no GPU CI), and the head is currently marked `needs-rebase`.

---

phil-opp:

Reviewed this — no blocking issues found.

The bulk of the diff is a verbatim move of the cross-machine pool code out of `daemon/src/lib.rs` into `libraries/extensions/tensor-pool/src/daemon/*` behind the extension's `daemon` feature, reached through the new `DaemonServices` trait. I focused on the parts that aren't a pure move:

- **Wire protocol** — the 12 pool-named variants + 4 dead pinned-memory variants collapse to 3 opaque carriers (`ExtensionRequest` / `ExtensionReply` / `ExtensionMessage`); new variants are appended last where postcard index stability matters. Intentional pre-1.0 wire break, marked `refactor!`.
- **`bulk_bytes::vec` serde helper** mirrors the existing `AVec` form and has real round-trip tests across the postcard varint boundaries (0/1/127/128/129/300/4096/70000).
- **Dispatch seam** — both the `#[cfg(feature = "tensor-pool")]` and the `not(...)` arm resolve to valid values; the feature-off path drops a peer `ExtensionMessage` and answers a node `ExtensionRequest` with a "no extension registered" error.
- **Safety-critical logic survived the move** — the 1 GiB mirror-allocation cap on untrusted remote `size`, invalid-pool-id rejection (avoids the `NodeId::from` panic), the `target_machine`/machine-id gate, the `is_cross` drop, and the DORADMA magic + seqlock are all present in the new files. `NodeRequest::Write` dropping `tensor_data` is safe — every caller passed `Vec::new()`.

The one thing I can't exhaustively verify by reading is byte-for-byte behavioral equivalence of the ~2.6k moved lines; that rests on the daemon + extension test suites (feature-on and feature-off). So it looks sound, contingent on green CI.

---

PR #3220, opened 2026-08-18
refactor!: drop the `_unstable_` prefix from the `deploy` and `debug` keys

phil-opp:
The `_unstable_` key prefix was one of two ways the tree signalled instability, and it was applied inconsistently: `deploy` and `debug` carried it, while `hub:` and `ros2:` — both documented as unstable — did not.

With the decision to signal instability through documentation rather than naming, the prefix is the odd one out, so it goes. Renames `_unstable_deploy` → `deploy` and `_unstable_debug` → `debug` across 38 files.

The Rust fields were already named `deploy` and `debug`, so this drops the `#[serde(rename)]` attributes rather than renaming anything in code.

### Why the rejection test matters

`Descriptor` has `deny_unknown_fields`, so a dataflow still using `_unstable_deploy:` fails with an error naming the field rather than having the block silently ignored. That distinction is not cosmetic: silent acceptance would run every node on the local daemon while the YAML looks like it pinned them to specific machines, and a stale `_unstable_debug:` would leave `dora topic echo` returning nothing with no explanation.

`removed_unstable_key_prefix_is_rejected_not_ignored` pins that for both keys, so it guards the `deny_unknown_fields` attribute as much as the rename.

No schema change — both fields are `#[schemars(skip)]`.

Breaking for any dataflow using the old keys. Pre-1.0 is when this is free.


phil-opp (commented):

No issues found.

Verified the core of the change in `libraries/message/src/descriptor.rs`: dropping the `#[serde(rename = "_unstable_deploy")]` / `#[serde(rename = "_unstable_debug")]` attributes is safe because the Rust field names are already `deploy`/`debug`. Both the `Descriptor` **and** `Node` structs carry `#[serde(deny_unknown_fields)]`, so a stale `_unstable_deploy`/`_unstable_debug` at either the top level or the per-node level is rejected with an error naming the field rather than silently ignored — so the "silent acceptance would run every node locally" risk is covered at both levels, not only the one the new `removed_unstable_key_prefix_is_rejected_not_ignored` test exercises.

The rename appears applied consistently across docs, examples, fixtures, the coordinator error strings, and the nightly-CI awk/grep injectors (`^  deploy:$`). This is a clean intended breaking change.

---

PR #3221, opened 2026-08-18
refactor!: narrow the `dora_core` re-export out of the 1.0 surface

phil-opp:
`pub use dora_core::{self, uhlc};` re-exported the entire crate, so `dora_core`'s descriptor, topics, build, manifest and type-registry modules were all reachable as `dora_node_api::dora_core::*` — and would have been frozen by the 1.0 guarantee. A node needs none of them.

An audit of every crate reaching `dora_core` through `dora-node-api` without declaring its own dependency found the complete set to be exactly `config::{DataId, NodeId, OperatorId}` and `uhlc::HLC` — 30 files across 26 crates, none touching the other modules. The replacement shim keeps that set and nothing else, and **the whole workspace compiles unchanged**.

The `config` path is preserved verbatim rather than flattened, because `dora_node_api::dora_core::config::DataId` is what `dora new` scaffolds (`binaries/cli/src/template/rust/talker/main-template.rs`) and what `docs/api-rust.md` documents. Verified the scaffolded template still compiles.

Breaking only for code reaching `dora_node_api::dora_core::{descriptor, topics, build, manifest, types}`. Nothing in-repo does, and the fix is a direct `dora-core` dependency.

Part of the pre-1.0 re-export cleanup alongside #3213, which removed Arrow from the same surface.

### DIFF
--- a/apis/rust/node/src/lib.rs
+++ b/apis/rust/node/src/lib.rs
@@ -112,7 +112,26 @@ pub use arrow58 as arrow_v58;
 // Arrow-typed, semver-exempt seam for dora's own crates and must not become
 // reachable through the frozen `dora-node-api` surface.
 pub use dora_arrow_convert::{DoraArray, IntoArrow, into_vec};
-pub use dora_core::{self, uhlc};
+pub use dora_core::uhlc;
+
+/// The subset of [`dora_core`] that is part of dora's public API.
+///
+/// Deliberately not `pub use dora_core::self`. Re-exporting the whole crate
+/// froze `dora_core`'s descriptor, topics, build, manifest and type-registry
+/// surface into dora's 1.0 guarantee, none of which a node needs — the only
+/// items reached through this path anywhere are the three id types below and
+/// `uhlc`, which is re-exported at the crate root.
+///
+/// The `config` path is preserved verbatim because
+/// `dora_node_api::dora_core::config::DataId` is what `dora new` scaffolds and
+/// what the Rust API reference documents.
+pub mod dora_core {
+    /// Identifier types used in node and dataflow configuration.
+    pub mod config {
+        pub use dora_core::config::{DataId, NodeId, OperatorId};
+    }
+    pub use dora_core::uhlc;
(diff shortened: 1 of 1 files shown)


phil-opp (commented):

No issues found.

The one non-obvious risk in this change is name resolution: the new `pub mod dora_core` shadows the extern crate `dora_core` at the crate root, so the inner `pub use dora_core::config::{...}` and `pub use dora_core::uhlc` could in principle be ambiguous or resolve back into the local module (a cycle). I reproduced the exact shim pattern against a stand-in crate in edition 2024 and it compiles cleanly — the `use` paths resolve to the extern crate, not the local module — matching the PR's claim that the workspace builds unchanged.

The `config` sub-path is preserved verbatim and `uhlc` remains reachable at both `dora_node_api::uhlc` and `dora_node_api::dora_core::uhlc`, so the only surface removed is `dora_core::{descriptor, topics, build, manifest, types}`, which is the intended breaking change.

---

phil-opp:

Reviewed the change — no issues found.

I audited every in-repo consumer of the old `pub use dora_core::{self, uhlc}` re-export. They only reach `dora_node_api::dora_core::config::{DataId, NodeId, OperatorId}` and `dora_node_api::dora_core::uhlc::HLC` (tensor-pool python transport, `examples/multiple-daemons/sink`, the C/C++/Python node APIs, and the `node/mod.rs` doctests). The hand-written `pub mod dora_core` shim preserves all of those, and the crate-root `pub use dora_core::uhlc` keeps the `uhlc` path working, so the workspace compiles unchanged. No name-shadowing hazard with the extern crate either: a crate-root `pub mod dora_core` doesn't enter submodules' unqualified scope, and `use dora_core::…` inside the new module resolves via the extern prelude.

The breaking part (external code reaching `descriptor`/`topics`/`build`/`types` through the re-export) is intentional and correctly marked `refactor!`.

One practical note: the PR currently conflicts with `main` and will need a rebase before it can merge.

---

PR #3222, opened 2026-08-19
fix(daemon): send the original YAML-shape Node with ReplaceNode and assign it wholesale

phil-opp:

## Issue

`patch_descriptor_entry` refreshes the stored descriptor entry with a `ReplaceNode` replacement's definition so the new incarnation does not inherit the outgoing one's config through `DORA_NODE_CONFIG` / `DoraNode::dataflow_descriptor()` (dora-rs/dora#2988 review, finding 2).

The copy was open-ended and set `path`, `inputs` and `outputs`, but left several paired fields at the **outgoing** incarnation's values:

- `path_sha256` and `build` — a replacement with a different URL/checksum or build command kept the predecessor's `path_sha256`, now mispaired with the new `path`.
- the `run_config` type/framing/pool maps `output_types`, `output_framing`, `input_types` and `shared_memory_pool_size` — part of the same descriptor surface as the already-copied `inputs`/`outputs`.

All of these are serialized into the child's `DORA_NODE_CONFIG` and exposed via `DoraNode::dataflow_descriptor()`, i.e. exactly the stale-config leak class this function exists to close.

## Fix

- Copy the missing fields onto `entry`.
- Destructure `CustomNode`/`NodeRunConfig` **exhaustively** (no `..` rest pattern), so a field added upstream fails to compile here instead of silently leaking. Every field is now either copied or explicitly bound-and-skipped with a rationale (`source`, `envs`, and the log-rotation fields `max_log_size`/`max_rotated_files` are the deliberate skips). This turns the recurrence of the original bug into a build error.

## Validation

- Extended the `patch_descriptor_entry` unit test to assert the patched entry tracks the replacement's `path_sha256`, `build`, and `output_types` (rather than retaining the outgoing values).
- `cargo test -p dora-daemon --lib`, `cargo clippy -p dora-daemon --lib`, `cargo fmt --all -- --check` pass locally. (Two pre-existing `cross_pool_write_tests` zenoh tests fail only in the sandbox because it cannot bind `tcp/[::]:0` — unrelated to this change.)

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -1867,6 +1867,12 @@ async fn start_inner(
                                     inner: DaemonCoordinatorEvent::ReplaceNode {
                                         dataflow_id,
                                         node: resolved_node.clone(),
+                                        // Ship the original YAML-shape node so
+                                        // the daemon can assign the descriptor
+                                        // entry wholesale, mirroring the
+                                        // `*existing = original_node` commit
+                                        // this arm does below.
+                                        unresolved_node: original_node.clone(),
                                         uv,
                                         grace_duration,
                                     },
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2033,36 +2033,6 @@ fn event_generation_is_stale(
     event_generation != entry_generation && event_generation != successor_generation
 }
 
-/// Patch a stored descriptor entry with a replacement node's definition
-/// (dora-rs/dora#2988 review, finding 2). The stored descriptor is
-/// serialized into every spawned process's `DORA_NODE_CONFIG`, so the
-/// replacement must not receive the outgoing incarnation's path, env, or
-/// restart configuration through `DoraNode::dataflow_descriptor()`.
-/// Only `CoreNodeKind::Custom` reaches here — ReplaceNode's v1 scope
-/// rejects other kinds up front.
-fn patch_descriptor_entry(
-    entry: &mut dora_message::descriptor::Node,
-    resolved: &dora_core::descriptor::ResolvedNode,
-) {
-    if let CoreNodeKind::Custom(custom) = &resolved.kind {
-        entry.path = Some(custom.path.clone());
-        entry.args = custom.args.clone();
-        entry.env = resolved.env.clone();
-        entry.inputs = custom.run_config.inputs.clone();
-        entry.outputs = custom.run_config.outputs.clone();
-        entry.restart_policy = custom.restart_policy;
-        entry.max_restarts = custom.max_restarts;
-        entry.restart_delay = custom.restart_delay;
-        entry.max_restart_delay = custom.max_restart_delay;
-        entry.restart_window = custom.restart_window;
-        entry.health_check_timeout = custom.health_check_timeout;
-        entry.finish_grace_secs = custom.finish_grace_secs;
-        entry.send_stdout_as = custom.send_stdout_as.clone();
-        entry.send_logs_as = custom.send_logs_as.clone();
-        entry.min_log_level = custom.min_log_level.clone();
-    }
-}
-
 fn clear_node_result(results: &mut DaemonRunResult, dataflow_id: Uuid, node_id: &NodeId) {
     let remove_dataflow = results.get_mut(&dataflow_id).is_some_and(|node_results| {
         node_results.remove(node_id);
@@ -4833,6 +4803,7 @@ impl Daemon {
             DaemonCoordinatorEvent::ReplaceNode {
                 dataflow_id,
                 node,
+                unresolved_node,
                 uv,
                 grace_duration,
             } => {
@@ -4988,15 +4959,20 @@ impl Daemon {
                     // Fresh stderr buffer for the new incarnation; installed
                     // into the map only after the spawn succeeds.
                     let node_stderr = Arc::new(ArrayQueue::new(STDERR_LOG_LINES_MAX));
-                    // Patch the descriptor CLONE handed to the spawner with
+                    // Replace the descriptor CLONE handed to the spawner with
                     // the replacement's definition before it is serialized
                     // into the child's DORA_NODE_CONFIG — the stored
                     // descriptor still describes the outgoing incarnation
                     // at this point (state mutations are deferred until the
-                    // spawn succeeds).
+                    // spawn succeeds). Assigning the coordinator-supplied
+                    // original YAML-shape node wholesale keeps every field
+                    // (path/checksum/build/type maps/cpu_affinity/deploy/
+                    // git-source/log-rotation/…) paired with the replacement,
+                    // with no field list to drift as new fields are added to
+                    // `Node` (dora-rs/dora#2988 review, finding 2).
                     let mut descriptor = dataflow.descriptor.clone();
                     if let Some(entry) = descriptor.nodes.iter_mut().find(|n| n.id == node_id) {
-                        patch_descriptor_entry(entry, &node);
+                        *entry = unresolved_node.clone();
                     }
                     let spawner = Spawner {
                         dataflow_id,
@@ -5121,18 +5097,18 @@ impl Daemon {
                         }
                     }
 
-                    // Update the stored descriptor entry to the
-                    // replacement's definition — the same patch applied to
-                    // the spawn-time clone, so later spawns (and restart
-                    // respawns of OTHER nodes) serialize the replacement,
-                    // not the outgoing incarnation.
+                    // Commit the same wholesale assignment on the stored
+                    // descriptor so later spawns (and restart respawns of
+                    // OTHER nodes) serialize the replacement, not the
+                    // outgoing incarnation. Mirrors the coordinator's own
+                    // `*existing = original_node` commit.
                     if let Some(entry) = dataflow
                         .descriptor
                         .nodes
                         .iter_mut()
                         .find(|n| n.id == node_id)
                     {
-                        patch_descriptor_entry(entry, &node);
+                        *entry = unresolved_node.clone();
                     }
 
                     running_node.mark_registered();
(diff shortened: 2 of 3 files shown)


phil-opp:
## Automated code-review sweep — summary

This PR is part of a machine-generated (Claude Code) review sweep of the repository against `origin/main`. Each finding was independently verified before a PR was opened. PRs opened this run, by category:

**Correctness**
- #3222 — `daemon`: `patch_descriptor_entry` left `path_sha256`, `build` and the `run_config` type/framing/pool maps at the outgoing incarnation's values on `ReplaceNode` (stale-config leak); also hardened with an exhaustive destructure so future fields can't silently leak.
- #3223 — `cli`: stray space before the colon in pretty logs for a node on the default (unnamed) daemon.
- #3224 — `download`: `get_filename` returned `None` (aborting the download) when a degenerate `Content-Disposition` filename sanitized away, instead of falling back to the URL name.

**Documentation**
- #3225 — `message`: the flagship `Descriptor` doc example was malformed YAML; fixed and converted to a compile-checked doctest.
- #3226 — `node-api`: `ZERO_COPY_THRESHOLD` doc wrongly said sub-threshold payloads go over TCP (they take a heap-buffered zenoh `put`).
- #3227 — `cli`: added compile-checked doctests for the public `parse_log_filter` / `parse_log_level_str` helpers.

**Optimization / Simplification (standalone):** none met the bar this run. Two simplifications were folded into the correctness PRs above (#3222's exhaustive-match backstop, #3223's de-duplicated default-daemon label).

### Rejected findings (verified myself, not acted on)
- **mavlink2-bridge rebuilds the Arrow `Schema` per message** (opt) — genuine per-message allocation churn on a telemetry stream, but the fix is a trait-signature change across 17 `MavlinkArrow` impls; too broad/risky for a single autonomous sweep PR. Worth a dedicated PR.
- **`download_file` buffers the whole body in memory before hashing** (opt) — streaming would weaken the current "no partial file written on hash mismatch" guarantee; behavior-change tradeoff, not a clear win.
- **node metadata-only sends bypass the zenoh path** (correctness, #2891 reordering class) — architecturally significant transport-plane change; too subtle to land autonomously without deep integration validation.
- **node `@schema` callback lacks the data path's empty-payload guard** (correctness) — could not confirm the claimed panic: an empty borrowed slice's `as_ptr()` is non-null, so `NonNull::new(..).expect()` does not fire; the secondary "primes an empty schema" concern is an already-handled/warned path.
- **`core::parse_byte_size` duplicates `ByteSize::from_str`** (simplification) — the mirroring is deliberate per the code comment and delegating risks subtle behavior drift (`is_ascii_alphabetic` vs `is_alphabetic`); not worth the risk.
- **mavlink `from_record_batch` O(n²) `column_by_name` scans** (opt) — inbound-command path only (lower rate); would be subsumed by the schema-caching change above if that is done.

Local gates run on the combined diff: `cargo fmt --all --check` and `cargo clippy --all -- -D warnings` clean; affected-crate tests pass (`dora-message`, `dora-download`, `dora-node-api`, `dora-cli`, `dora-daemon`). The only failing tests are pre-existing zenoh integration tests that can't bind `tcp/[::]:0` in the sandbox — unrelated to these changes.

---

phil-opp:
🤖 **Fully automated review by Claude** (no human in the loop — treat as advisory).

The exhaustive destructure is a nice compile-time backstop, and the refresh looks correct against the current `CustomNode` / `NodeRunConfig` field sets. I do have one question about the deliberate skips.

The block skips `max_log_size` and `max_rotated_files`, with the rationale that log-rotation settings are "intentionally preserved across a replace." But as far as I can tell the stored `entry` is never consulted to configure a node's rotation — `spawn/prepared.rs` resolves both `max_log_size` and `max_rotated_files` from the *resolved replacement* node (`self.node.max_log_size()` / `.max_rotated_files()`), and a `ReplaceNode` re-runs that spawn path. So if a replacement changes its log-rotation settings, the node actually rotates at the **new** thresholds while `DoraNode::dataflow_descriptor()` / `DORA_NODE_CONFIG` (the stored `entry`) keeps reporting the **outgoing** values — which is the same stale-descriptor divergence this function exists to close.

It's also a little inconsistent within the log-config family: `send_stdout_as`, `send_logs_as`, and `min_log_level` are refreshed here, but these two are not.

Could you double-check whether rotation is genuinely preserved across a replace at runtime? If it isn't, these two fields probably want to be copied like the other log fields. `entry` already carries both, so it's a straightforward copy.

Minor: the added test covers `path_sha256`, `build`, and `output_types`, but not the other newly-refreshed `run_config` fields (`output_framing`, `input_types`, `shared_memory_pool_size`) — worth an assertion or two for completeness, and a small assertion locking in whatever behavior you settle on for the two log-rotation fields.

---

phil-opp (commented):

I found one issue not covered by the earlier review.

`patch_descriptor_entry` now refreshes the `CustomNode`/`NodeRunConfig` fields (exhaustively destructured) plus `resolved.env`, but it never copies the other `ResolvedNode`-level fields — `cpu_affinity`, `deploy`, `name`, `description` — onto the stored `entry`. On a `ReplaceNode` the node is spawned from the resolved replacement (e.g. `cpu_affinity` is applied at spawn from `self.node.cpu_affinity`, `binaries/daemon/src/spawn/prepared.rs:632`), but the stored descriptor entry keeps the outgoing incarnation's `cpu_affinity`/`deploy`. That entry is serialized into the child's `DORA_NODE_CONFIG` and exposed via `DoraNode::dataflow_descriptor()`, so after a replace that changes `cpu_affinity` (or `deploy`) the reported descriptor diverges from what the node was actually spawned with — the same stale-config-leak class this function exists to close (and the twin of the `path_sha256`/`build` miss it fixes).

Worth flagging: the exhaustive-destructure backstop the PR adds only covers `CustomNode`/`NodeRunConfig`, so it does **not** protect these `ResolvedNode`-level fields — neither the current miss nor a future field added to `ResolvedNode` would be caught at compile time, contrary to the impression the commit message gives. Consider copying `cpu_affinity`/`deploy` (and `name`/`description` if they're meant to track a replace) alongside `env`, and destructuring `ResolvedNode` exhaustively as well.

The added test covers `path_sha256`/`build`/`output_types` but not `cpu_affinity`/`deploy`.

---

phil-opp:
Good catch — addressed in the latest push (force-update `a849337`).

`patch_descriptor_entry` now also copies the `ResolvedNode`-level fields `cpu_affinity`, `deploy`, `name` and `description` (alongside `env`), and `ResolvedNode` is now destructured exhaustively too — so the compile-time backstop covers `ResolvedNode` in addition to `CustomNode`/`NodeRunConfig`, and a future field added to any of the three fails to build here rather than silently leaking. `id` is the one deliberate skip (a replace keeps the same id). The unit test now also asserts `cpu_affinity` tracks the replacement.

---

phil-opp:

The force-pushed head (`a849337`) resolves the earlier finding about missing replacement fields: the exhaustive `ResolvedNode` / `CustomNode` / `NodeRunConfig` destructures now cover `cpu_affinity`, `deploy`, `name`, and `description`, and they're complete and correct against current `main`.

One item is still outstanding from the earlier round — the `max_log_size` / `max_rotated_files` skip. The comment says log-rotation settings are "intentionally preserved across a replace," but I don't think that matches runtime behavior. `spawn/prepared.rs` resolves both from the *resolved replacement* node (`self.node.max_log_size()` / `.max_rotated_files()`, prepared.rs:906-914), and a `ReplaceNode` re-runs that spawn path, so the new incarnation actually rotates at the replacement's thresholds — the stored `entry` is never consulted for rotation. Its only role here is what gets serialized into the child's `DORA_NODE_CONFIG` and reported via `DoraNode::dataflow_descriptor()`. Skipping these two therefore leaves the reported descriptor diverging from what the node was spawned with — the exact stale-config-leak class this function is closing — and is inconsistent with `send_stdout_as` / `send_logs_as` / `min_log_level`, which are refreshed. `entry` already carries both fields, so this looks like a straightforward copy (plus a matching assertion). If rotation genuinely is preserved across a replace somewhere, a pointer in the skip comment would help, since the current wording reads as an assumption rather than verified behavior.

---

phil-opp:
issue found by claude:

The exhaustive destructure is a good backstop but two "deliberate skips" are contradicted by the code: source is the resolved form of git/branch/tag/rev (classify.rs:52), and rotation settings are resolved from the replacement at spawn time (prepared.rs:906), not preserved. The structural fix is to carry the original Node in ReplaceNode — the coordinator already holds it (coordinator/src/lib.rs:1851) — and assign the whole entry, so no field list exists to rot.

phil-opp:
Both corrections stand and the structural fix is the right call — pushed as `24dbfd4`.

Confirmed against the code:
- `Node` has `git`/`branch`/`tag`/`rev` fields (the resolved form of `NodeSource`), so my "source is spawn-time only" rationale for skipping was wrong.
- `prepared.rs:906` resolves `max_log_size`/`max_rotated_files` from the replacement at spawn, so my "log rotation is preserved" rationale was wrong too.

Rather than hand-writing a `ResolvedNode`→`Node` back-conversion that would keep drifting, `DaemonCoordinatorEvent::ReplaceNode` now carries an extra `unresolved_node: Node` field (the coordinator already had it as `original_node`), and both daemon call sites assign the entry wholesale (`*entry = unresolved_node.clone()`), mirroring the coordinator's own `*existing = original_node` commit. `patch_descriptor_entry` and its unit test are removed — the type-level identity of the assignment is the correctness guarantee, so there is no per-field enumeration left to test. Net -158 lines.

Local: `cargo check`/`clippy`/`fmt` clean; `cargo test -p dora-daemon --lib` passes 249/251 (the 2 failures are pre-existing sandbox-only zenoh tests that can't bind `tcp/[::]:0`, unrelated to this change).

---

phil-opp:
Thanks — both corrections were right (I verified: `git`/`branch`/`tag`/`rev` are the flat form of the resolved `NodeSource`, and `max_log_size`/`max_rotated_files` are resolved from the replacement at `prepared.rs:906`, so neither was a legitimate skip). Your structural commit (`24dbfd4`) carrying the unresolved `Node` on the wire and assigning the entry wholesale — mirroring the coordinator's own `*existing = original_node` — is the right depth and removes the field-list-rot entirely. I'll stand down on this branch and won't push over your commit.

---

PR #3223, opened 2026-08-19
fix(cli): drop stray space before the colon for the default daemon in pretty logs

phil-opp:

## Issue

In `format_pretty_line`, the daemon label for a `DaemonId` with no `machine_id` was `"on default daemon "` **with a trailing space**, while both the named-daemon arm and the node-less default-daemon arm have none. For a node-scoped message on the unnamed/default daemon with `print_daemon_name` enabled, the line builder appends the colon directly after the label, producing:

```
sensor on default daemon : hello
```

— a stray space before the colon that the named-daemon path (`sensor on daemon `m`: hello`) does not have.

## Fix

- Remove the trailing space.
- Collapse the two divergent `"on default daemon"` literals (the `Some(id)`-no-machine-id arm and the `None` arm) into a single arm by hoisting the `print_daemon_name` guard and matching on `daemon_id.as_ref().and_then(|id| id.machine_id())`, so the two copies cannot drift apart again (this near-duplication is what let the spacing diverge in the first place).

## Validation

- Added a `format_pretty_line` regression test for the default-daemon + `print_daemon_name` case, which was previously untested (the test helper always set `daemon_id: None`).
- `cargo test -p dora-cli --lib`, `cargo clippy -p dora-cli --lib`, `cargo fmt --all -- --check` pass locally.

---

### DIFF
--- a/binaries/cli/src/output.rs
+++ b/binaries/cli/src/output.rs
@@ -125,13 +125,16 @@ fn format_pretty_line(log_message: &LogMessage, config: &LogOutputConfig) -> Str
         }
         _ => String::new().cyan(),
     };
-    let daemon = match daemon_id {
-        Some(id) if config.print_daemon_name => match id.machine_id() {
+    let daemon = if config.print_daemon_name {
+        // A daemon with no machine id (or no daemon at all) is the default
+        // daemon; both render the same, with no trailing space so the colon
+        // that follows a node-scoped label abuts it directly.
+        match daemon_id.as_ref().and_then(|id| id.machine_id()) {
             Some(machine_id) => format!("on daemon `{machine_id}`"),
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed this change and found no issues. The refactor in `binaries/cli/src/output.rs:128-137` is behavior-equivalent to the original match apart from the intended fix: hoisting the `print_daemon_name` guard and matching on `daemon_id.as_ref().and_then(|id| id.machine_id())` collapses the `Some(id)`-with-no-machine-id arm and the `None` arm into one, both now rendering `"on default daemon"` with no trailing space, while the named-daemon and disabled cases are unchanged. The added regression test asserts the exact `"...sensor on default daemon: hello"` rendering and guards against the stray-space form the previous helper never exercised.

---

PR #3224, opened 2026-08-19
fix(download): fall back to the URL name when Content-Disposition sanitizes away

phil-opp:

## Issue

`get_filename` chose the `Content-Disposition` filename **or** the URL name *before* sanitizing:

```rust
let raw_name = cd_name.or_else(|| filename_from_url(response.url()));
raw_name.and_then(|name| /* Path::file_name + NUL/length checks */)
```

So a non-empty but degenerate header value — e.g. a hostile/broken `Content-Disposition: attachment; filename=".."` (or `"."`) — was selected, then `Path::file_name` mapped it to `None`, and the whole function returned `None`. The download then aborted with *"Could not find a filename"* even though the URL's last path segment would have named it fine. A broken or malicious server can thus deny a download the URL alone could satisfy.

## Fix

Sanitize each source **independently** and chain the fallback, so a Content-Disposition value that sanitizes away no longer suppresses the URL name:

```rust
cd_name.and_then(|n| sanitize_filename(&n))
    .or_else(|| filename_from_url(url).and_then(|n| sanitize_filename(&n)))
```

Factored the sanitize step into `sanitize_filename` and the selection into a pure, unit-testable `resolve_filename(content_disposition, url)` helper; `get_filename` is now a thin wrapper.

## Validation

- Added `resolve_filename` unit tests: prefers Content-Disposition; falls back to the URL when no header; **falls back to the URL when the header sanitizes away** (`".."`, `"."`); returns `None` only when both sources are unusable.
- `cargo test -p dora-download`, `cargo clippy -p dora-download --all-targets`, `cargo fmt --all -- --check` pass locally.

---

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -116,28 +116,43 @@ fn filename_from_url(url: &reqwest::Url) -> Option<String> {
         .map(|segment| segment.to_string())
 }
 
+/// Sanitize a candidate filename: strip path components to prevent traversal,
+/// reject null bytes and overly long names. Returns `None` when nothing usable
+/// survives — e.g. `".."` or `"."` (which `Path::file_name` maps to `None`),
+/// or a name over 255 bytes / containing a NUL. (A trailing slash such as
+/// `"dir/"` keeps its last component: `Path::file_name` returns `"dir"`.)
+fn sanitize_filename(name: &str) -> Option<String> {
+    let sanitized = Path::new(name)
+        .file_name()
+        .and_then(|n| n.to_str())
+        .map(|s| s.to_string())?;
+    if sanitized.contains('\0') || sanitized.len() > 255 {
+        return None;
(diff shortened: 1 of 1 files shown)


phil-opp:

I reviewed this change and found no issues. The diagnosis is accurate: the old `raw_name = cd_name.or_else(url_name)` selected a degenerate `Content-Disposition` value (e.g. `filename=".."`) before sanitizing, and `Path::file_name` then mapped it to `None`, aborting a download the URL alone could have named. Sanitizing each source independently and chaining the fallback in `resolve_filename` (`libraries/extensions/download/src/lib.rs:139-145`) fixes that, and `get_filename` stays a faithful thin wrapper. The new `resolve_*` tests cover the header-preferred, no-header, header-sanitizes-away, and both-unusable cases, and the `len() > 255` check operates on the sanitized name. No regression for the previously-working paths.

---

PR #3229, opened 2026-08-19
fix(daemon): don't forward the spurious empty EOF line from node log readers

### LINKED ISSUE #3228: Bug: spurious empty log line emitted at every stdout/stderr EOF in daemon node reader
> **Note:** This issue was created by a **scheduled automated Claude check** that randomly selects a component and reviews it for correctness bugs. The finding below was independently verified against the source, but please double-check before acting on it.

## Summary

The daemon's per-node stdout/stderr reader tasks in `binaries/daemon/src/spawn/prepared.rs` send one **spurious empty log line** every time a stream reaches EOF (i.e. once per stdout stream and once per stderr stream, on every node exit). Downstream handling does not filter it, so it produces:

1. an empty JSONL entry (`"msg": ""`) appended to the node's log file, and
2. an empty `LogBroadcast` to `dora/logs` subscribers, and
3. **most consequentially** — if the node has `send_stdout_as` / `send_logs_as` configured, a spurious empty-string Arrow output message delivered to downstream dataflow consumers, i.e. an extra data item the node author never produced.

## Mechanism

`read_capped_line` signals EOF as `Ok(true)` **only** when `read_until` returns `n == 0`, and on that path leaves `raw` untouched (empty). Conversely `Ok(false)` is only returned after `n > 0`, so it always implies a non-empty `raw`. Therefore the only empty `raw` this function ever produces is the final EOF read.

`binaries/daemon/src/spawn/prepared.rs:116-152`:
```rust
async fn read_capped_line<R>(reader: &mut tokio::io::Take<R>, raw: &mut Vec<u8>) -> std::io::Result<bool>
where R: tokio::io::AsyncBufRead + Unpin {
    const CAP: u64 = MAX_LOG_LINE_BYTES as u64 + 1;
    reader.set_limit(CAP);
    let n = reader.read_until(b'\n', raw).await?;
    if n == 0 {
        return Ok(true);   // EOF — `raw` left empty
    }
    // ... (cap/discard handling) ...
    Ok(false)              // only reached after n > 0, so `raw` is non-empty
}
```

The stdout reader task (`prepared.rs:741-791`; the stderr task at `806-851` is identical) checks `finished` in the `while` condition, which runs **after** the body has already built and sent the `LogLine`:

```rust
let mut finished = false;
while !finished {
    let mut raw = Vec::new();
    finished = match read_capped_line(&mut child_stdout, &mut raw).await ... {
        Ok(eof) => eof,
        Err(err) => { /* log */ true }
    };
    let mut content = match String::from_utf8(raw) { ... };  // empty -> ""
    truncate_log_line(&mut content);                          // no-op on ""
    // sends unconditionally, even when content == "" on the EOF iteration:
    stdout_tx.send(LogLine { content, stream: LogStream::Stdout }).await ...
}
```

On the final EOF iteration, `raw` is empty → `content == ""` → an empty `LogLine` is sent, then `finished == true` ends the loop.

## Concrete trigger

Any node whose stdout ends in a newline (the normal case), e.g. `"line1\nline2\n"`, yields three reads: `"line1\n"`, `"line2\n"`, then an empty EOF read. The third is sent as an empty log line. A node that prints nothing at all still emits one empty line per stream.

## Downstream impact (no emptiness filter anywhere)

In the log-consuming task (`prepared.rs:929-1131`), empty `content` is not filtered:

- `send_stdout_as` path (`939-963`): `content.as_str().into_arrow()` builds a valid single-element string array `[""]`; `ipc_log_payload` (`42-59`) Arrow-IPC-encodes it and returns `Some`, so a `DoraEvent::Logs` output is emitted to the dataflow. This block runs **before** the `min_log_level` filter, so it fires regardless of level config.
- Broadcast (`1005-1018`) and JSONL file write (`1048-1069`) both proceed with the empty message, producing an empty `dora/logs` broadcast and a `"msg": ""` log-file entry.

## Why this is an oversight, not intended

The test helper for this exact function, `read_all_lines` (`prepared.rs:1430-1444`), explicitly guards against the empty EOF read:

```rust
loop {
    let mut raw = Vec::new();
    let eof = read_capped_line(&mut reader, &mut raw).await.unwrap();
    if !raw.is_empty() {         // <-- discards the empty EOF read
        lines.push(raw);
    }
    if eof {
        break;
    }
}
```

The production reader tasks lack the equivalent guard.

## Suggested fix

Skip sending when the read produced empty `raw` (which only happens at EOF), mirroring the test helper — e.g. gate the `send` on non-empty `content`, or `break` before constructing/sending `content` when `finished && raw.is_empty()`. Apply to both the stdout (`741-791`) and stderr (`806-851`) tasks. A regression test asserting that an empty final line is not forwarded (and that no empty Arrow output is emitted under `send_stdout_as`) would lock in the behavior.

phil-opp:
## Summary

Fixes #3228.

The per-node **stdout** and **stderr** log reader tasks in `binaries/daemon/src/spawn/prepared.rs` emitted one **spurious empty log line** every time a stream reached EOF (once per stream, on every node exit).

`read_capped_line` signals EOF as `Ok(true)` and leaves `raw` **empty**, but the reader loop only checked `finished` in the `while` condition — *after* the loop body had already built and sent the `LogLine`. On the final EOF iteration `raw` was empty → `content == ""` → an empty `LogLine` was forwarded, then the loop ended.

Nothing downstream filters it, so every node exit produced:
- an empty JSONL entry (`"msg": ""`) in the node log file,
- an empty `LogBroadcast` to `dora/logs` subscribers, and
- **most consequentially** — under `send_stdout_as` / `send_logs_as`, a spurious empty-string Arrow output message delivered to downstream dataflow consumers (an extra data item the node author never produced).

## Fix

The forward-or-skip decision is now a single shared function:

```rust
fn line_to_forward(raw: Vec<u8>) -> Option<Vec<u8>> {
    (!raw.is_empty()).then_some(raw)
}
```

Both reader tasks call it — `let Some(raw) = line_to_forward(raw) else { continue };` — so there's exactly one place the guard lives. An empty `raw` is only ever the EOF read (a genuine blank line is `"\n"`, non-empty), so it maps to `None` and is skipped; every non-empty line is forwarded, preserving real blank lines and any partial bytes read before an I/O error.

## Test

`empty_eof_read_is_not_forwarded` drives its loop through the same `super::line_to_forward` function, not a private copy of the guard, so the test genuinely exercises the production code:

- output ending in a newline (`"line1\nline2\n"`) forwards exactly the two real lines,
- a node that prints nothing forwards no line,
- a genuine blank line (`"\n"`) is still preserved.

**Verified the coupling**: replacing `line_to_forward`'s body with `Some(raw)` makes the test fail (trailing empty line); restoring the guard passes it. Thanks @phil-opp for pointing that out — the previous version's private copy of the guard would have kept the test green after deleting the production guards.

## Validation

- `cargo test -p dora-daemon --lib spawn::prepared` — 8 passed (incl. the new test, verified failing when the guard is removed)
- `cargo fmt -p dora-daemon -- --check` — clean

### DIFF
--- a/binaries/daemon/src/spawn/prepared.rs
+++ b/binaries/daemon/src/spawn/prepared.rs
@@ -151,6 +151,20 @@ where
     Ok(false)
 }
 
+/// The forward-or-skip decision shared by the stdout/stderr reader tasks and
+/// their tests: given the `raw` bytes [`read_capped_line`] just produced,
+/// return the line to forward downstream, or `None` to skip it.
+///
+/// `read_capped_line` leaves `raw` empty **only** on the final EOF read — a
+/// real blank line is `"\n"`, which is non-empty — so an empty `raw` is that
+/// EOF marker and is dropped. Forwarding it would emit a spurious empty log
+/// line, and under `send_stdout_as` a spurious empty-string Arrow output the
+/// node never produced (dora-rs/dora#3228). Keeping this in one function means
+/// the tasks and the regression test exercise the same guard.
+fn line_to_forward(raw: Vec<u8>) -> Option<Vec<u8>> {
+    (!raw.is_empty()).then_some(raw)
+}
+
 #[derive(Clone, Default)]
 struct RestartConfig {
     max_restarts: u32,
@@ -758,6 +772,12 @@ impl PreparedNode {
                     }
                 };
 
+                // Drop the empty final EOF read rather than forwarding a
+                // spurious empty log line — see `line_to_forward`.
+                let Some(raw) = line_to_forward(raw) else {
+                    continue;
+                };
+
                 let mut content = match String::from_utf8(raw) {
                     Ok(s) => s,
                     Err(err) => {
@@ -817,6 +837,12 @@ impl PreparedNode {
                         tracing::warn!("{err:?}");
                         true
                     }
+                };
+
+                // Drop the empty final EOF read rather than forwarding a
+                // spurious empty log line — see `line_to_forward`.
+                let Some(raw) = line_to_forward(raw) else {
+                    continue;
                 };
 
                 let mut content = match String::from_utf8(raw) {
@@ -1457,6 +1483,39 @@ mod tests {
             read_all_lines(b"a\nno-newline").await,
             vec![b"a\n".to_vec(), b"no-newline".to_vec()]
         );
+    }
+
+    /// Drive the reader loop through the **same** `line_to_forward` guard the
+    /// production stdout/stderr tasks use, collecting whatever it forwards.
+    /// Because it calls the production function (not a private copy), removing
+    /// the guard there fails the assertions below.
+    async fn forwarded_lines(input: &[u8]) -> Vec<Vec<u8>> {
+        let mut reader = tokio::io::BufReader::new(input).take(MAX_LOG_LINE_BYTES as u64);
+        let mut forwarded = Vec::new();
+        let mut finished = false;
+        while !finished {
+            let mut raw = Vec::new();
+            finished = read_capped_line(&mut reader, &mut raw).await.unwrap();
+            if let Some(raw) = super::line_to_forward(raw) {
+                forwarded.push(raw);
+            }
+        }
+        forwarded
+    }
+
+    #[tokio::test]
+    async fn empty_eof_read_is_not_forwarded() {
+        // Normal case: output ends in a newline. The trailing EOF read is
+        // empty and must not be forwarded as a spurious empty log line
+        // (dora-rs/dora#3228).
+        assert_eq!(
+            forwarded_lines(b"line1\nline2\n").await,
+            vec![b"line1\n".to_vec(), b"line2\n".to_vec()]
+        );
+        // A node that prints nothing at all forwards no line.
+        assert!(forwarded_lines(b"").await.is_empty());
+        // A genuine blank line ("\n") is non-empty `raw`, so it is preserved.
+        assert_eq!(forwarded_lines(b"\n").await, vec![b"\n".to_vec()]);
     }
 
     #[tokio::test]


phil-opp:

I reviewed this change and found no issues. The `if raw.is_empty() { continue; }` guard sits after `finished` is assigned from `read_capped_line` (`prepared.rs:758/765` for stdout, `:831/833` for stderr) and the loop is `while !finished`, so the EOF read (`finished == true`, empty `raw`) exits the loop cleanly on the `continue` — no busy-loop risk. Gating on `raw` rather than the post-parse `content` is correct: a genuine blank line arrives as a non-empty `"\n"` and is still forwarded, while only the empty EOF read is dropped. The new `empty_eof_read_is_not_forwarded` test is meaningful — it drives the real `read_capped_line` and asserts all three cases (trailing newline forwards two lines, empty input forwards nothing, a bare `"\n"` is preserved).

---

phil-opp:

The fix itself looks correct — `read_capped_line` only leaves `raw` empty on the EOF read (`n == 0`), a blank line comes back as `"\n"` (non-empty, still forwarded), and the added `if raw.is_empty() { continue; }` in both reader tasks cleanly skips only the spurious EOF line.

I do have one issue with the added test, though. `empty_eof_read_is_not_forwarded` doesn't actually exercise the production guard. The `forwarded_lines` test helper reimplements the read loop and carries its **own** copy of the skip:

- `prepared.rs:1488` — `if raw.is_empty() { continue; }` (commented "Mirrors the `if raw.is_empty() { continue; }` guard in the tasks")

The real guards under test live in the two spawned reader tasks at `prepared.rs:765` (stdout) and `prepared.rs:833` (stderr). Because the test asserts against the helper's private copy of the loop, reverting **both** production guards would leave this test green — so it isn't a regression test for the bug this PR fixes.

It does lock in the load-bearing premise (that `read_capped_line` yields an empty `raw` only at EOF, and a blank line as a non-empty `"\n"`), which has value. But to actually guard the fix, the test would need to drive the reader-task code path — or factor the "read line → skip empty-at-EOF → forward" step into a shared function that both the tasks and the test call.

No other concerns.

---

phil-opp:
issue found by claude:

the test asserts against a private copy of the loop — delete both production guards and it still passes. Extract the shared step.

phil-opp:
Good catch — fixed in the latest push. The skip decision is now a single shared function, `line_to_forward(raw) -> Option<Vec<u8>>`, that both reader tasks and the test call. The two inline `if raw.is_empty() { continue; }` guards are gone; each task now does `let Some(raw) = line_to_forward(raw) else { continue };`, and `forwarded_lines` in the test drives its loop through `super::line_to_forward` instead of a private copy.

Verified the coupling holds: replacing the guard body with `Some(raw)` makes `empty_eof_read_is_not_forwarded` fail (trailing empty line), and restoring it passes. Behavior is otherwise unchanged — blank lines (`"\n"`) and any partial bytes read before an I/O error are still forwarded.

---

PR #3231, opened 2026-08-19
test(daemon): wait for the zombie with waitid, not a sysinfo poll

phil-opp:
`spawn_unreaped_wrapper` polled `sysinfo` for `ProcessStatus::Zombie` with a 5-second deadline. That made the test's *precondition* depend on sysinfo observing the zombie — a platform question, not the one under test. The poll timed out on the 2026-08-19 macOS nightly (#3150) while passing on Linux, so `an_unreaped_leader_of_ours_still_covers_its_surviving_child` and `a_strangers_zombie_at_a_recycled_pid_does_not_make_its_group_ours` both died at `the wrapper never became a zombie`, before reaching an assertion.

`waitid(P_PID, pid, WEXITED | WNOWAIT)` answers the question the helper is actually asking — has it exited yet — and `WNOWAIT` leaves the status pending, so the process stays an unreaped zombie for the assertions and for `kill_group_and_reap`. Measured on Linux: returns in ~0.9 ms, a second `waitid` still succeeds (i.e. still unreaped), `kill(-pid, 0)` still finds the group, and the later `Child::wait` still reaps.

Three details the call needs to be worth trusting. It retries on `EINTR`, so a signal delivered to the waiting thread is not reported as the child failing to exit — that is the flake class this change exists to remove. It reports `errno` on failure, because `ECHILD` (something reaped the wrapper out from under the test) and `EINVAL` (wrong options for this platform) are the two answers worth telling apart. And it runs on its own thread with a 30-second recv deadline, so a wrapper that never exits fails *here* with a message rather than blocking until the CI job's own timeout kills the run and reports nothing. Unlike the poll it replaces, that deadline cannot expire while the answer is already available.

The sysinfo check is kept, as an explicit assertion rather than a poll, because a disagreement between the two is worth surfacing. `live_children` treats a pid sysinfo cannot see as *not* recycled — `pid_recycled_into_stranger` returns `false` for `None` — so on a platform where zombies are invisible to sysinfo, a pid recycled into a **stranger's** zombie is attributed to us and its process group gets `killpg`'d. That is the #3067 case, fixed on Linux by #3076.

Reading sysinfo 0.39's `src/unix/apple/macos/process.rs` says which way macOS will go, and it is the unhappy one. For a pid not already in the map, `create_new_process` returns `Err(())` unless `sysctl(KERN_PROCARGS2)` or `proc_pidpath` succeeds — both read state the kernel no longer serves for a zombie — so `System::process()` yields `None` and the `SZOMB => Zombie` arm is never reached. That is precisely what the old poll was waiting for, which explains the nightly timeout. So expect this to leave the macOS job red, failing fast at the new assertion instead of after 5 seconds of nothing.

The blast radius is narrower than that assertion firing suggests, though, and its message now says so. The helper builds a *fresh* `System`; a long-lived `DestroyWait` keeps its `System` across polls and has already seen the pid alive, so it takes sysinfo's *update* path, which can still report `Zombie` where the create path returns nothing. The macOS follow-up is real but should be scoped to that difference rather than filed as "the guard does not work".

Full `dora-daemon` lib suite passes (253), shutdown tests green over 15 repeat runs at 0.02 s each, `cargo fmt --check` and `cargo clippy -p dora-daemon -- -D warnings` clean. (`--all-targets` clippy reports four pre-existing findings in `lib.rs`, untouched here and not covered by CI's `--all` invocation.)

Refs #3150

### DIFF
--- a/binaries/daemon/src/shutdown.rs
+++ b/binaries/daemon/src/shutdown.rs
@@ -463,29 +463,7 @@ mod tests {
         // without the reap it stays a zombie, which is the state under test.
         drop(wrapper.stdin.take());
 
-        // Block until the kernel says it has exited, *without* reaping it.
-        // `WNOWAIT` leaves the status pending, so the process stays a zombie
-        // for the assertions below and for `kill_group_and_reap` to clean up.
-        //
-        // This replaces a poll for `ProcessStatus::Zombie` (dora-rs/dora#3150):
-        // that made the precondition depend on `sysinfo` *observing* the
-        // zombie, which is a platform question, not the one under test. The
-        // poll timed out on macOS nightly while passing on Linux. `waitid` is
-        // POSIX and answers the actual question — has it exited yet — with no
-        // deadline to lose.
-        //
-        // SAFETY: `waitid` only reads the exit state of a child this test
-        // spawned; `WNOWAIT` means it does not consume it.
-        let rc = unsafe {
-            let mut info: libc::siginfo_t = std::mem::zeroed();
-            libc::waitid(
-                libc::P_PID,
-                pid as libc::id_t,
-                &mut info,
-                libc::WEXITED | libc::WNOWAIT,
-            )
-        };
-        assert_eq!(rc, 0, "waitid on the wrapper failed");
+        wait_for_exit_without_reaping(pid);
 
         // The wrapper is now definitively an unreaped zombie. If `sysinfo`
         // disagrees, that is a platform gap worth knowing about rather than a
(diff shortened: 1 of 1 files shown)


phil-opp:
🤖 **Fully automated review by Claude** (no human in the loop — treat as advisory).

I reviewed this change; no issues found.

A few things I checked specifically:
- `libc` is a `cfg(unix)` dependency (not dev-only), so `waitid`/`siginfo_t`/`P_PID`/`WEXITED`/`WNOWAIT` resolve in the test build, and the helper plus its callers are already `#[cfg(unix)]`.
- `WNOWAIT` leaves the zombie waitable, so the later `wrapper.wait()` in `kill_group_and_reap` still reaps it — no double-reap and no leak.
- The retained sysinfo check now asserts `Some(ProcessStatus::Zombie)`, which matches the production zombie test in `pid_recycled_into_stranger`/`is_live_child`, so a platform where sysinfo can't see the zombie fails at the mechanism it names rather than a bare timeout.

One minor, non-blocking observation: the new `waitid` blocks with no internal deadline, whereas the old poll gave up after 5s. Since dropping stdin reliably drives the wrapper to EOF-and-exit this should always return promptly, and it's the intended tradeoff — just noting that a hypothetical never-exiting wrapper would now hang the test instead of failing fast.

---

phil-opp:
🤖 **Automated review by Claude** — fully automated; no human has reviewed or verified these findings.

Re-reviewed the new commit `d679da9` added since my earlier review of `b433f82`. No new issues.

It directly addresses the one non-blocking observation from that review (the `waitid` block had no internal deadline): the wait now runs on its own thread behind a 30s `recv_timeout`, so a wrapper that never exits fails here with a clear message instead of hanging until the CI job timeout. The `EINTR` retry and the errno-preserving `panic!` (`ECHILD`/`EINVAL` vs. timeout) are correct hardening — a signal on the wait thread is retried rather than misreported as the child failing to exit, and `WNOWAIT` still leaves the zombie reapable for `kill_group_and_reap`. The narrowed sysinfo assertion message is consistent with the create-vs-update path distinction it describes.

---

PR #3235, opened 2026-08-19
fix(replay-node): honor the daemon's Stop event

### LINKED ISSUE #3234: replay-node ignores the Stop event, so `dora replay --loop` can't be stopped gracefully
### What's happening

`dora-replay-node` never reads its event stream, so it never sees the `Stop` event the daemon sends on shutdown. The only thing that ends its replay loop is running out of records, and with `--loop` that never happens. The daemon ends up force-killing the node 15 seconds later.

### Steps to reproduce

```bash
dora record dataflow.yml -o capture.drec   # Ctrl-C after a few seconds
dora replay capture.drec --loop
# Ctrl-C
```

I'd expect the replay node to pick up `Stop`, finish the entry it's on, and exit. Instead it keeps replaying for about 10 more seconds while downstream nodes keep receiving messages, and then the daemon prints:

```
<node> was killed due to not stopping within the 15s grace period
```

With `--speed 0`, those 10 seconds are the whole recording replayed over and over at full speed.

### Why it happens

In `binaries/replay-node/src/main.rs`:

```rust
let (mut node, _events) = DoraNode::init_from_env()?;
```

`_events` is never used again. There's no `Event::Stop` match and no `recv` or `try_recv` anywhere in the file. `init_from_env` does subscribe, so the daemon has a live channel and delivers `Stop` into it (`running_dataflow.rs:630`); the node just never reads it. The only way out of the loop is:

```rust
if !do_loop {
    break;
}
```

On the daemon side, `stop_all` sends `Stop`, waits `DEFAULT_STOP_GRACE` (10s), sends SIGTERM, waits another 5s, then SIGKILL. Outputs keep being routed that whole time, since there's no gate on the send path during teardown, which is deliberate so nodes can flush.

Worth noting that the sibling binary gets this right: `binaries/record-node/src/main.rs` has `Event::Stop(_) => break` in its loop. Nothing in the replay node mentions shutdown at all, so this reads more like an oversight than a decision.

### Why it matters

The main one is that `dora replay --loop` has no graceful way to stop. Beyond that, replay exists to feed recorded data into live nodes, so those extra 10 seconds mean replayed commands keep reaching a live control node after the operator already asked for a stop. And every `dora replay` teardown prints a force-kill warning, which makes that warning easy to start ignoring.

The kill is at least classified as planned (`grace_duration_kills`), so it doesn't surface as a failed dataflow.

### Possible fix

Swapping the pacing `thread::sleep` for `events.recv_timeout(...)` would cover both the pacing and the `Stop` check in one call, with no added latency. `--speed 0` never sleeps, so that path needs a cheap `try_recv` poll per iteration too, plus a check between loop passes. One thing to watch: a stop-interrupted pass should return `Ok(())` rather than hitting the `replay_emitted_nothing_usable` bail, otherwise a normal stop turns into an error.

CapThunder19:
Closes #3234 

### The problem

`dora-replay-node` bound its event stream to `_events` and never read it, so it never saw the `Stop` event the daemon sends on shutdown. The only thing that ended the replay loop was running out of records, and with `--loop` that never happens. So the daemon force-killed the node at the 15s grace deadline, and in the meantime replayed messages kept reaching live downstream nodes for ~10 seconds after the operator asked for a stop.

The sibling binary already gets this right: `record-node` has `Event::Stop(_) => break` in its loop.

### The fix

Wait out the pacing gap *on* the event stream instead of in `thread::sleep`. Pacing and stop-detection then happen in the same call, so there's no added latency a timeout costs what the sleep it replaced cost.

`--speed 0` has no gap to wait out, and neither does the boundary between two `--loop` passes, so those poll with `try_recv` instead.

One detail worth flagging for review: `recv_timeout` returns `Some(Event::Error(..))` on timeout and `None` only when the stream closes. `classify_event` leans on that distinction. If the two were conflated, every replay would die at its first pacing gap.

A stop-interrupted pass now exits `Ok` and skips the `replay_emitted_nothing_usable` check, since a partial pass is interrupted rather than defective — otherwise a normal `dora stop` would turn into a replay error.

Also fixes a smaller case: `--loop` over a recording holding nothing for this node used to spin reopening the file with no poll anywhere, so it couldn't be stopped at all.

### Testing

`cargo fmt --all -- --check`, `cargo clippy -p dora-replay-node --all-targets -- -D warnings`, and `cargo test -p dora-replay-node` (11 passed) are clean.

The three new tests cover `classify_event`, which is new, so they aren't a regression test that fails against the old code. The loop wiring itself isn't unit-testable without a live `EventStream` — every existing test in this file is a pure helper for the same reason. Happy to add a smoke test driving `dora replay --loop` if you'd prefer real coverage of the exit path.

### DIFF
--- a/binaries/replay-node/src/main.rs
+++ b/binaries/replay-node/src/main.rs
@@ -1,8 +1,12 @@
-use std::{fs::File, thread, time::Duration};
+use std::{
+    fs::File,
+    time::{Duration, Instant},
+};
 
 use dora_message::{common::Timestamped, daemon_to_daemon::InterDaemonEvent};
 use dora_node_api::{
-    DoraArray, DoraNode, IntoArrow, arrow_utils::decode_arrow_ipc, arrow_v59::array::NullArray,
+    DoraArray, DoraNode, Event, EventStream, IntoArrow, TryRecvError,
+    arrow_utils::decode_arrow_ipc, arrow_v59::array::NullArray,
 };
 use dora_recording::RecordingReader;
 use eyre::Context;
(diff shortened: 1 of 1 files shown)


phil-opp:
🤖 **Automated review by Claude** — fully automated; no human has reviewed or verified these findings.

Reviewed the diff at `159627e`. No issues found.

What I verified against the actual `dora-node-api` source rather than the doc comments:
- The whole design hinges on `EventStream::recv_timeout` surfacing a timeout as `Event::Error`, not `None`. Confirmed: `recv_async_timeout` returns `Some(Event::Error(...))` on the `Delay` branch (`event_stream/mod.rs:1174`) and `None` only once the stream is closed. So `classify_event` mapping a timeout to `Continue` and `None`/`Stop` to `Stop` is correct — a pacing timeout will not be mistaken for a closed stream.
- `wait_out_gap` tracks elapsed via `checked_sub` instead of an absolute `Instant::now() + gap` deadline, so a pathologically small `--speed` that saturates `sleep_nanos` to `u64::MAX` can't panic on the deadline addition. Pacing is preserved: a non-`Stop` event returned before the timeout just loops and waits out the remaining gap rather than returning early.
- `poll_stop` covers the two zero-gap paths — `--speed 0` (`pacing_sleep_nanos` returns 0 for `speed <= 0.0`) and the `--loop` pass that matches no entries — so a stop is still observable when there's no gap to wait on.
- `TryRecvError` / `EventStream` / `Event` / `StopCause` are all re-exported from `dora_node_api` (`apis/rust/node/src/lib.rs:150`), so the imports resolve.
- The added tests exercise the real `classify_event` and can genuinely fail (the timeout/`InputClosed` → `Continue` case in particular guards against the regression where every replay would stop at its first pacing gap).

One non-blocking note: events consumed inside `wait_out_gap`/`poll_stop` other than `Stop` are discarded. That's fine for a replay node (it has no inputs to lose), just flagging the behavior.

---

PR #3236, opened 2026-08-19
docs(migration): cover the RC-window breaking changes

phil-opp:
`docs/migration-from-0.x.md` predates this batch of breaking changes and covered none of them, while its own "Still to do" list asked for per-API before/after snippets. Every change below is one a 0.x user hits on the first build after upgrading, and #3213 in particular changes the type of `Event::Input`'s payload — the most-used thing in the API.

**Descriptor** — `custom:` removal (#3158), the `_unstable_` key-prefix drop (#3220), the `publish_all_messages_to_zenoh` alias removal (#3158), and the dead `communication:` block (#3158). Each with before/after YAML.

The section also records *why the errors are loud*: `Descriptor` and `Debug` both use `deny_unknown_fields`, so a stale key is rejected by name rather than ignored. That matters more than it sounds — a silently-dropped `deploy` block would run every node on the local daemon while the YAML looks like it pinned them to machines.

**Rust API** — `ArrowData` → `DoraArray` (private field, no `Deref`), `IntoArrow` losing its associated type, `pub use arrow` becoming the feature-gated `arrow_v59`, and the narrowed `dora_core` re-export (#3221). Each with before/after code.

The most useful line for most readers: **most nodes need no Arrow feature at all**, because `value.into_arrow()` and `(&data).try_into()?` never name an Arrow type — which is exactly why the `dora new` templates were unchanged by #3213.

Signatures were checked against the merged implementations rather than the PR descriptions.

Docs only.

### DIFF
--- a/docs/migration-from-0.x.md
+++ b/docs/migration-from-0.x.md
@@ -63,9 +63,139 @@ Features and code that existed in dora 0.x but are absent from 1.0. If you depen
 | Daemon `state.rs` module (172 lines) | Removed | Refactored into `running_dataflow.rs` + `event_types.rs`. Same state handling, clearer separation of concerns. |
 | Upstream `libraries/extensions/ros2-bridge/python/src/typed/deserialize/string.rs` | Moved | Functionality is in `libraries/extensions/ros2-bridge/arrow/src/deserialize/string.rs` and additionally supports `WStringDeserializer` (which was `todo!()` in 0.x). |
 
+## Descriptor changes (dataflow YAML)
+
+These landed during the 1.0 RC window. All are rejected with an error naming
+the offending field rather than being silently ignored — `Descriptor` and
+`Debug` both use `deny_unknown_fields`, which matters: a silently-dropped
+`deploy` block would run every node on the local daemon while the YAML looks
+like it pinned them to machines.
+
+### `custom:` removed (#3158)
+
+Use the top-level fields directly.
+
+```yaml
+# 0.x
+- id: my-node
+  custom:
+    path: node.py
+    args: --flag
+    inputs:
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found. This is a docs-only change to `docs/migration-from-0.x.md`; I verified each load-bearing technical claim against the code at the PR head:

- **Rust API** (`apis/rust/node/src/lib.rs`, `libraries/arrow-convert/src/lib.rs`): `ArrowData` is now `DoraArray` with a private `ArrayRef` field and no `Deref`; `DoraArray::as_array()` (returning `&arrow::array::ArrayRef`) is gated behind the `arrow-v59` feature; `IntoArrow::into_arrow(self) -> DoraArray` has no associated type; `arrow` is re-exported only as the feature-gated `arrow_v59`; the `arrow-v59` feature name and `default = []` (no Arrow feature on by default) match `Cargo.toml`.
- **`dora_core` re-export** narrowed to exactly `config::{DataId, NodeId, OperatorId}` and `uhlc`, with the `config` path preserved verbatim — matches the `pub mod dora_core` in lib.rs.
- **Descriptor** (`libraries/message/src/descriptor.rs`): both `struct Descriptor` and `struct Debug` carry `#[serde(deny_unknown_fields)]`, so the "rejected with a named error, not silently ignored" framing is accurate; the `publish_all_messages_to_zenoh` alias and the `_unstable_` key prefix are removed and explicitly rejected (covered by `removed_unstable_key_prefix_is_rejected_not_ignored` and the `publish_all_messages_to_zenoh` rejection test); `enable_debug_inspection` is the current field name.

Before/after snippets read correctly and the feature/dependency names line up with `Cargo.toml`.

---

PR #3237, opened 2026-08-19
refactor!: drop the `flume` re-export from `dora-node-api`

phil-opp:
`pub use flume;` and `pub use flume::Receiver;` were pure convenience — no flume type appears in any public signature of `dora-node-api`. Every use inside the crate is internal: a private field on a private struct, `RecvTimeoutError`, and `flume::bounded`, all in `event_stream/thread.rs`.

Removing them takes flume out of the frozen 1.0 surface, so a future flume major stops being a breaking change to dora.

Two consumers reached it through the re-export and now declare it themselves:

- `examples/ros2-bridge/rust/rust-ros2-example-node` (behind `required-features = ["ros2"]`)
- `binaries/mavlink2-bridge-node`

An earlier audit of this re-export reported only the first. Both write multi-line `use dora_node_api::{...}` blocks, which is what a single-line grep misses — the second surfaced only under `cargo clippy --all`, which is worth knowing for the remaining re-export cleanup.

### Verification caveat

The ros2 example cannot be compiled in the dev container: msg-gen emits nothing without a ROS distro, so that crate fails with 9 `cannot find <pkg> in messages` errors both before and after this change. No import errors appear among them, which is the available signal that the flume path resolves — a real break would show up alongside. Nightly CI's ros2-bridge job is the actual gate.

### Not doing the same for `uuid`

`pub type DataflowId = uuid::Uuid`, and `DoraNode::dataflow_id()` returns `&DataflowId`, so uuid stays in the frozen surface whether or not it is re-exported. Dropping that re-export would be cosmetic unless `DataflowId` becomes an opaque newtype — a larger, separately breaking change.


phil-opp:

No issues found.

Verified:
- The claim that no flume type appears in `dora-node-api`'s public signatures holds. Inside the crate, flume is used only internally: `event_stream/thread.rs` uses `flume::RecvTimeoutError`, `flume::bounded`, and a private `handle: flume::Receiver<...>` field; `event_stream/mod.rs` uses `tokio::sync::mpsc::Receiver`, not flume. Removing `pub use flume;` / `pub use flume::Receiver;` therefore does not remove any nameable public type.
- Both consumers that reached flume through the re-export are handled: `flume` is added as a direct dependency in `binaries/mavlink2-bridge-node/Cargo.toml` and `examples/ros2-bridge/rust/rust-ros2-example-node/Cargo.toml`, and Cargo.lock reflects both. Their usages are all path-qualified (`flume::bounded`, `flume::Sender`, `flume::unbounded`), which resolve via the extern prelude once flume is a direct dep — so dropping the `dora_node_api::flume` import in each still compiles.
- No other file in the ros2 example crate references `flume` or the re-exported `Receiver`; only `action_server.rs` did, and it is updated.

Caveat carried from the PR description and not independently checkable here: the ros2 example only builds with a ROS distro (msg-gen), so its compilation of the flume path relies on nightly CI's ros2-bridge job rather than local verification. The change to that crate is limited to the import line plus the Cargo.toml dep, which is low-risk.

---

PR #3238, opened 2026-08-19
fix(message): gate the daemon Metadata wire format at registration

phil-opp:
`InterDaemonEvent::Output` carries a `Metadata` between daemons, but nothing checked that two daemons agree on its layout.

`dora_version` does not cover this: #2366 dropped a `Metadata` field without changing the version, and the resulting mid-stream desync was #2742. The node→daemon path has gated it since then (`NodeRegisterRequest::check_version`); the daemon path never did.

### There is no daemon↔daemon handshake to add

That path is zenoh pub/sub — there is no connection to shake hands on. And semver compatibility between daemons is *already* guaranteed transitively, because registering with the coordinator is mandatory and `DaemonRegisterRequest::check_version()` rejects mismatches.

The coordinator is therefore the only chokepoint that sees every daemon, and gating there covers the daemon↔daemon wire transitively: any two daemons routing a dataflow have both passed the same check, so they agree on the `Metadata` layout.

So this adds `metadata_version` to `DaemonRegisterRequest` and checks it in `check_version()`, mirroring the node→daemon gate one hop out.

### On `#[serde(default)]`

A daemon built before the field omits it, `serde(default)` yields 0, and the check fails with a legible error rather than silently treating it as compatible.

That works because this frame is **JSON over the coordinator WebSocket**. In a non-self-describing encoding the peer would fail while decoding the very frame carrying the field and never reach the check — the same caveat the node→daemon gate already documents, and worth keeping in mind as more of the protocol moves to postcard.

Three tests: a same-version daemon with a stale layout is rejected, a pre-field daemon still deserializes but is rejected, and the existing direction-aware semver advice is unchanged.

### DIFF
--- a/libraries/message/src/daemon_to_coordinator.rs
+++ b/libraries/message/src/daemon_to_coordinator.rs
@@ -4,7 +4,8 @@ pub use crate::common::{
     DataMessage, LogLevel, LogMessage, NodeError, NodeErrorCause, NodeExitStatus, Timestamped,
 };
 use crate::{
-    BuildId, DataflowId, common::DaemonId, current_crate_version, id::NodeId, versions_compatible,
+    BuildId, DataflowId, common::DaemonId, current_crate_version, id::NodeId, metadata::Metadata,
+    versions_compatible,
 };
 
 /// Per-dataflow status reported by a daemon after (re-)registration.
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Verified:
- `metadata_version: u16` on `DaemonRegisterRequest` matches the type of `Metadata::CURRENT_VERSION` (`pub const CURRENT_VERSION: u16 = 2` in `libraries/message/src/metadata.rs`), and `DaemonRegisterRequest::new` stamps it with `Metadata::CURRENT_VERSION`.
- The new check lives inside the existing `check_version()`, and its `Result` flows through the identical `version_check_result` path the pre-existing semver check already used (`binaries/coordinator/src/ws_daemon.rs:182` → `DaemonRequest::Register { version_check_result, .. }`). So enforcement is unchanged in mechanism — only an additional rejection condition is added — and a mismatched daemon is rejected the same way an incompatible-semver daemon already was.
- The check is a strict `!=` against the coordinator's `CURRENT_VERSION`, which makes the "transitive agreement" argument hold: all daemons equal to the coordinator's version are equal to each other.
- Both new tests genuinely exercise the new branch: `same_version_daemon_with_a_different_metadata_layout_is_rejected` sets a compatible `dora_version` (via `new()`) but a `wrapping_add(1)` metadata version, so it reaches and fails the new check; `a_daemon_predating_the_field_is_rejected_...` strips the JSON field so `serde(default)` yields 0, then asserts rejection. Both would fail if the gate were removed.

One note (not a defect, just scoping): `metadata_version` gates only the `Metadata` layout, which is the specific desync #2366/#2742 addressed; it is not a general fingerprint of the entire `InterDaemonEvent` wire format. That matches the stated intent.

---

PR #3239, opened 2026-08-19
refactor!: hide tokio from the integration-testing API

phil-opp:
`tokio` was in `dora-node-api`'s public surface through exactly one line — `pub use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel}` in `integration_testing` — which would have frozen a tokio major into the 1.0 guarantee.

### Why this isn't the feature gate that was planned

The plan of record was to gate the whole `integration_testing` module behind a non-default feature. Investigating it turned up three costs:

- **The module is load-bearing on the *non*-test path.** `init_from_env` reads `DORA_TEST_WITH_INPUTS`, `DORA_TEST_WRITE_OUTPUTS_TO`, and `DORA_TEST_NO_OUTPUT_TIME_OFFSET`, so gating removes env-var testing from release builds.
- **It would have silently disabled a test tier.** `tests/example-tests.rs` drives that mode with `cargo run -p <crate>`, which builds *without* dev-dependencies — so a dev-dep-only feature leaves the harness green while testing nothing.
- **It contradicts the mode's rationale.** Env-var testing exists to exercise "the exact same executable as used in the dataflow"; a feature flag makes that untrue by construction.

None of it was necessary. Tokio's exposure was only those three items, and the env-var path never touches them — it uses `TestingOutput::ToFile`. Wrapping them achieves the goal with no gate at all:

| before | after |
|---|---|
| `unbounded_channel()` | `output_channel()` |
| `UnboundedSender<OutputJson>` | `OutputSender` |
| `UnboundedReceiver<OutputJson>` | `OutputReceiver` |

The newtypes serve the same purpose the re-export did — a test needs no channel dependency of its own, which is what the old doc comment gave as the reason for it — and additionally keep *which* channel library dora uses out of the frozen API, so it can be swapped later without a breaking change.

### Verification

`pub use tokio` and every `tokio::` path are now absent from all public signatures. **`cargo test --test example-tests` passes** — that is specifically the tier a feature gate would have broken, so it is the meaningful check here rather than a formality.

Also confirmed, since it was an open question blocking the original plan: the record/replay path does not touch `integration_testing_format`, so nothing there is affected.

Breaking for tests using the old names; the fix is the rename in the table.

### DIFF
--- a/apis/rust/node/src/daemon_connection/node_integration_testing.rs
+++ b/apis/rust/node/src/daemon_connection/node_integration_testing.rs
@@ -264,7 +264,7 @@ impl IntegrationTestingEvents {
 
 enum OutputWriter {
     Writer(Box<dyn Write + Send>),
-    Channel(tokio::sync::mpsc::UnboundedSender<serde_json::Map<String, serde_json::Value>>),
+    Channel(crate::integration_testing::OutputSender),
 }
 
 pub fn convert_output_to_json(
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -2357,7 +2357,7 @@ mod tests {
             "test-node".parse().unwrap(),
             events,
         ));
-        let (tx, _rx) = tokio::sync::mpsc::unbounded_channel();
+        let (tx, _rx) = crate::integration_testing::output_channel();
         let outputs = TestingOutput::ToChannel(tx);
         let options = TestingOptions {
             skip_output_time_offsets: true,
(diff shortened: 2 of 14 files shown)


phil-opp:

No issues found.

Verified:
- This is a pure rename/newtype refactor: `unbounded_channel()` → `output_channel()`, with `UnboundedSender<OutputJson>`/`UnboundedReceiver<OutputJson>` wrapped as opaque `OutputSender`/`OutputReceiver`. The goal (removing `tokio` from `dora-node-api`'s public surface) is achieved and no `tokio` type remains in the re-exported API.
- The one non-test call site, `handle_output` in `apis/rust/node/src/daemon_connection/node_integration_testing.rs`, calls `sender.send(output).context(...)`. `OutputSender::send` now returns `Result<(), eyre::Report>`, and `.context()` on a `Result<T, eyre::Report>` compiles fine (eyre's `WrapErr`/`Context` is implemented for `Report`), so this is correct.
- `drain_outputs` now reads through `rx.0.try_recv()`, unchanged semantics.
- All call sites (event_stream tests, node/mod.rs tests, and the seven example test files) are updated consistently; a missed rename would be a hard compile error caught by CI.
- `OutputSender` keeps `Clone`; `OutputReceiver` is not `Clone` (correct for a receiver).

---

Issue #2721, opened 2026-07-16
zenoh 1.9: dora is one flat peer region — remaining topology gaps after #2716


phil-opp:
#2715 bundled four defects. #2716 fixes three outright (dead `routing/peer` insert, `lowlatency` losing >64 KiB cross-host messages, inverted SHM comment) and the fourth — peers no longer relaying — for the case that actually bites: same-machine links between statically-spawned nodes, now established by construction from the dataflow graph.

Closing #2715 as superseded rather than editing it again: its framing predates the evidence below, and it still carries a mis-sourced quote. This issue tracks only what #2716 leaves open.

## Root cause, stated properly this time

#2715 said "zenoh 1.9 changed behavior under us." That's true but understates it. The precise statement, from [`DEFAULT_CONFIG.json5` @ 1.9.0](https://raw.githubusercontent.com/eclipse-zenoh/zenoh/1.9.0/DEFAULT_CONFIG.json5):

> The `auto` preset mode puts peers and clients south of routers and client south of peers. **This is the only available preset.**

Dora never sets a session `mode` anywhere in the repo, so every session — daemon, coordinator, node — is a default `peer`. No routers, no clients. Under `auto` that means **no gateway boundary materializes anywhere**: the whole dataflow is a single flat peer region, and per the [deployment docs](https://zenoh.io/docs/getting-started/deployment/):

> In peer to peer regions, Zenoh nodes must be deployed with `peer` mode and must also be deployed in a clique topology where every node is connected to all other nodes of the region.

So this isn't tuning we lost — it's a topology model dora no longer fits. `linkstate` *was* our exemption from the clique requirement, and 1.9 deleted the exemption without `auto` offering a replacement. The 5/20 route failures measured in #2716 are the predicted behavior of an incomplete clique, not bad luck.

zenoh's 1.9 migration guidance names our case directly: *"Users are expected to migrate non-clique peer-to-peer deployments from relying on router failover brokering to instead use multiple peer subregions in the router."*

#2716 satisfies the clique requirement the cheap way — over exactly the `|edges|` pairs the graph needs, rather than all N². That is the correct fix for local static nodes and should not be replaced by a relay-based shape: any gateway hop would undo the direct zero-copy path from #1787. The two gaps below are the ones that approach structurally cannot reach.

## 1. daemon↔daemon across networks

[...]

PR #3240, opened 2026-08-19
feat: cross-machine node-to-node zenoh links and an explicit daemon mesh

phil-opp:
Cross-machine dataflows currently relay every edge through two daemons, and the daemons themselves only find each other by multicast or a single gossip rendezvous — neither of which works on a mesh VPN, the deployment we want to support. This wires both tiers explicitly and takes the daemon off the cross-machine hot path.

**Daemon tier.** `--zenoh-listen` takes an optional port and `--zenoh-connect` names the peers to dial, so every daemon listens where its peers can predict and dials every other one — the clique zenoh 1.9 requires, with no multicast and no waiting for gossip. `dora cluster up` derives that mesh from `cluster.yml`, using `host` when it is an IP (the common tailnet case needs no new configuration) with per-machine `zenoh_addr` / `zenoh_port` overrides. Deriving nothing is preferred over deriving half a mesh: explicit connect endpoints disable multicast for the daemons that have them, so a partial mesh partitions the rest.

**Node tier.** A node consumed from another machine now also listens on a routable address, the daemons trade those endpoints over a zenoh queryable before spawning anything, and the consumer dials the producer directly. The trade has to happen up front because zenoh reads `connect/endpoints` once at startup and never re-reads it (verified in 1.9.0's `orchestrator.rs`). A remote static consumer becomes a required acker instead of pinning the output, so the switch is proven end-to-end before it is used — until the ack lands, and if it never lands, the output stays on the lossless daemon path. Edges that keep the daemon path: a dynamic consumer on another machine, a producer with no dialable address, an exchange that timed out, and a consumer declaring `input_timeout` (its deadline is refreshed only by its own daemon, so a direct send would leave liveness detection silently disarmed).

**Routers.** `--zenoh-config-overlay` layers a JSON5 file onto the config dora computes, appending to the endpoint lists rather than replacing them. `ZENOH_CONFIG` remains the wholesale escape hatch, but it discards the per-node plan — so a router named there relays same-machine traffic too, off the host and back if it runs elsewhere. Setting both is refused.

Also corrects `examples/multiple-daemons/README.md`, whose "run a `zenohd` per subnet" advice does not work for the NAT case it is offered for (a router relays 0/5 for non-clique peers under zenoh 1.9, measured in #2721), and the `plan_zenoh_peering` caveat claiming #2762 causes silent data loss — since #2891 an unproven route freezes on the daemon path, and holding the reservation as that issue proposes cannot work, because a held `TcpListener` is exactly what makes the node's own bind fail.

Design decisions here were made by @phil-opp; #2721 was blocked waiting on them.

## Validation

**Class**: C (daemon spawn wiring and the data path)

- `make qa-fast` equivalents: `cargo fmt --all -- --check` ✅, `cargo clippy --all -- -D warnings` ✅ (whole workspace, C++ crates included)
- `cargo test --all` (Python crates and `dora-examples` excluded per CLAUDE.md): ✅
- `cargo check --examples`: ✅
- `cargo test --test fault-tolerance-e2e -- --test-threads=1`: ✅ 11 passed
- Contract tests: 4 passed, 4 failed — all four failures are the Python/`uv` tests (`uv` is not installed in this dev container; the daemon log shows `spawning: uv run python -u ...` then a non-zero exit). Needs a re-run where `uv` is present.
- Targeted feature test: `tests/multi-daemon-e2e.rs::cross_machine_nodes_link_directly_over_an_explicit_mesh` — two daemons on an explicit mesh with multicast off, producer on one machine and consumer on the other. Delivery is asserted by construction (the sink fails its node if no cross-machine value arrives) and the daemon log proves the endpoint exchange carried it. Verified as a real guard: with `DORA_ZENOH_ENDPOINT_EXCHANGE_TIMEOUT_MS=0` the dataflow still delivers (daemon fallback intact) but the test fails on the direct-link assertion.
- `/code-review high`: ran; three findings, all fixed in `fix(daemon): keep liveness, mesh and event-loop guarantees intact` (the `input_timeout` regression, a mesh endpoint collision for two daemons on one host, and the exchange blocking the daemon event loop with an unbounded `declare_queryable`).
- `/simplify`: ran; three cleanups applied, no behavior change.

Closes part of #2721. Corrects the #2762 analysis without closing it (the reserve→bind window remains, now accurately described).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2223,6 +2223,7 @@ dependencies = [
  "fs_extra",
  "git2",
  "itertools 0.15.0",
+ "json5",
  "log",
  "proptest",
  "schemars 1.2.2",
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -195,10 +195,11 @@ type ZenohPublishers = HashMap<DataId, DirectOutput>;
 /// Declare a direct-zenoh data publisher for every output that may ever take
 /// the direct path, plus the per-output ack state the startup handshake needs.
 ///
-/// Outputs the daemon pinned `daemon_only` (some consumer runs under another
-/// daemon, so delivery must go through this daemon's inter-daemon forwarding —
-/// #2738) get no publisher and no markers: they stay on the daemon path for
-/// the node's lifetime. Every other output gets a publisher declared eagerly
+/// Outputs the daemon pinned `daemon_only` — a consumer only inter-daemon
+/// forwarding can reach (a dynamic node on another daemon, or a remote static
+/// one with no dialable endpoint for this node), and forwarding is fed solely
+/// by daemon-path sends (#2738) — get no publisher and no markers: they stay on
+/// the daemon path for the node's lifetime. Every other output gets a publisher declared eagerly
 /// at init (rather than on first send) for two reasons: zenoh starts wiring
 /// routes immediately, and [`StartupHandshake`] needs the publishers to probe
 /// those routes before the node's first real send. An output with no required
@@ -235,7 +236,8 @@ fn declare_output_publishers(
         if output_routing.daemon_only {
             debug!(
                 output = %output_id,
-                "output pinned to the daemon path (a consumer runs under another daemon)"
+                "output pinned to the daemon path (a consumer is reachable only by \
+                 inter-daemon forwarding)"
             );
             continue;
         }
@@ -630,7 +632,7 @@ fn wait_for_grace(ack_states: &[Arc<AckState>], grace: Duration) {
             .iter()
             .all(|state| state.ready.load(Ordering::Relaxed))
         {
-            return;
+            break;
         }
         if Instant::now() >= grace_deadline {
             break;
@@ -649,6 +651,15 @@ fn wait_for_grace(ack_states: &[Arc<AckState>], grace: Duration) {
                 "startup handshake incomplete after {}ms; output stays on the \
                  reliable daemon path for the rest of the run",
                 grace.as_millis()
+            );
+        } else {
+            // The positive half of the same decision, and the only signal that
+            // an output is *off* the daemon path — which for a consumer on
+            // another machine means its data no longer crosses two daemons.
+            // Logged per output, once, at the moment it is settled for the run.
+            debug!(
+                output = %state.output_id,
+                "startup handshake complete; output takes the direct zenoh path"
             );
         }
     }
@@ -1819,8 +1830,9 @@ impl DoraNode {
                     return Err(NodeError::Output(format!(
                         "output \"{output_id}\": IPC-encoded message is {} bytes, exceeding \
                          the {}-byte daemon transport limit (the output is on the daemon \
-                         path: pinned for a consumer on another daemon, its startup \
-                         handshake did not complete, or no zenoh route is available)",
+                         path: pinned for a consumer only forwarding can reach, its \
+                         startup handshake did not complete, or no zenoh route is \
+                         available)",
                         v.len(),
                         dora_message::MAX_MESSAGE_BYTES,
                     )));
--- a/binaries/cli/src/command/cluster/config.rs
+++ b/binaries/cli/src/command/cluster/config.rs
@@ -207,8 +207,7 @@ impl ClusterConfig {
     }
 
     /// Per-machine `--zenoh-listen`/`--zenoh-connect` arguments wiring every
-    /// daemon to dial every other one, or `None` when the mesh cannot be
-    /// derived.
+    /// daemon to dial every other one.
     ///
     /// Since zenoh 1.9, peers do not relay for each other: two daemons that
     /// never form a direct link exchange nothing, with no fallback to wait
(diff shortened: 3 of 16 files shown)


phil-opp:
🤖 **Automated review by Claude** — this review is fully automated, with no human in the loop. Treat it as advisory.

No issues found. This is a large change; I reviewed the logic-heavy paths against the diff:

- **`output_routing.rs`** — the dynamic/local/routable/liveness decision tree maps correctly to the documented policy. The new `input_timeout` guard (`watches_liveness`) correctly forces `daemon_only` on a remote input, since its deadline is refreshed only on the daemon path (`note_output_sent_to_local_receivers` walks local mappings only); a direct send would leave the deadline armed-but-unfired. `daemon_only` is sticky per output, so an output with a remote dynamic consumer stays pinned for all consumers, and a recorded acker on a pinned output is ignored — consistent with the docs.
- **`endpoint_exchange.rs`** — declare and collect share one bounded budget; `declare_queryable` has its own deadline; the queryable is kept alive for the dataflow lifetime and dropped in `finish_dataflow`; `ConsolidationMode::None` avoids silently dropping a daemon's endpoints; a panicked/cancelled task degrades to "no remote endpoints" and never fails the spawn. Key sanitization keeps daemons distinct via the uuid suffix.
- **`topics.rs` overlay** — `connect.endpoints`/`listen.endpoints` are removed from `rest` at parse time and merged into dora's lists, so `apply()` cannot overwrite them (confirmed by `an_overlay_splits_off_the_endpoint_lists` / `an_overlay_applies_onto_a_zenoh_config`). Deep-path insertion preserves siblings, with a whole-object fallback only for sections zenoh accepts atomically (#2721). Setting both `ZENOH_CONFIG` and the overlay is refused.
- **`open_zenoh_session_with_listen`** — the verification loop checks every requested listener but only ever returns the caller's own `listen_endpoint` as `effective_listen_endpoint`; env-supplied node listeners are diagnostic-only. Duplicate dials are removed via a seen-set (`retain`), not adjacency-only `dedup`.
- **`zenoh_mesh_args`** — all-or-nothing derivation, the two-daemons-on-one-host endpoint-collision check, and mutual exclusivity with `zenoh_peer` are correct and covered by tests.

The added tests exercise the changed behavior and would fail if the logic were wrong — notably the e2e test's `DORA_ZENOH_ENDPOINT_EXCHANGE_TIMEOUT_MS=0` control, which proves the direct-link assertion is a real guard rather than tautological. (The e2e test needs a routable local address and self-skips when none is found, so it may not run in every environment — test-only, no production risk.)

---

phil-opp:
review by claude:

The new e2e test will fail on rebase, silently. Verified: the branch writes _unstable_deploy: at tests/multi-daemon-e2e.rs:178,334,437,442; main renamed the key to deploy: in a1088ac30 (#3220) and descriptor.rs:1556 has a dedicated test asserting _unstable_deploy is rejected, not aliased. Git auto-merges the file cleanly (disjoint regions), so this lands as a semantic conflict that fails the blocking "multi-daemon late-join E2E" job.

Nothing but sync + GitGuardian has ever run on this branch. Other findings: dora cluster up now emits a fatal --zenoh-listen <host-ip>:5456 by default, giving unchanged cluster.yml files two new silent-death modes (NAT/VIP host address; port 5456 already held) with no opt-out; DORA_ZENOH_LISTEN becomes comma-separated with no version gate, and an older node parses it as one locator and loses both listeners — the #2742 failure class; endpoint_exchange.rs:234 has no inter-round delay, so the retry loop becomes a query storm; the 1.5 s exchange is awaited inline on the daemon event loop (lib.rs:6157).

phil-opp:
**Automated review by Claude** — _this review is fully automated, with no human in the loop. Treat it as advisory._

Re-reviewed after commit `2036249`. All four findings from my previous comment are addressed and no new issues surfaced:

- **`_unstable_deploy:` rebase clash** — `tests/multi-daemon-e2e.rs` now writes `deploy:` throughout (`node_yaml` at :171, cross-machine dataflow at :338/:342). No semantic conflict with #3220's rename remains.
- **`DORA_ZENOH_LISTEN` forward compat** — Split into `DORA_ZENOH_LISTEN` (loopback-only, single locator so an older 1.0.x node parser still accepts it) and the new `DORA_ZENOH_LISTEN_EXTRA` for the routable one (`libraries/core/src/topics.rs:42`, :50, :637-640; `binaries/daemon/src/spawn/spawner.rs` env-set block). Both variables are on `CONTROL_PLANE_ENV`, so a descriptor cannot forge them. The regression test `listen_endpoints_keep_loopback_first_and_merge_the_extra_var` at `topics.rs:1205` locks the order in.
- **Query storm** — `binaries/daemon/src/spawn/endpoint_exchange.rs:227-241` captures `round_started` at the top of each loop iteration and paces with `sleep(till_next_round.min(remaining))`, so a peer answering instantly no longer re-issues the wildcard `get` at reply-rate for the whole budget. The `.min(remaining)` also guarantees the pacing itself cannot push the exchange past `deadline`.
- **`cluster up` silent daemon death** — `binaries/cli/src/command/cluster/up.rs:158`, :210-236 tails `/tmp/dora-daemon-<id>.log` over ssh on timeout, so a NAT/VIP `host:` or an already-held port surfaces the actual bind error instead of just "timed out waiting for daemon(s)". Best-effort as described (an unreachable machine is skipped, not fatal).

The event-loop concern from that comment was actually already fixed one commit earlier in `a95f3e3`: `binaries/daemon/src/lib.rs:4369-4379` spawns the exchange as a `tokio::spawn` task before the build-metadata work, and only awaits it at :4655 just before the spawn loop — so the bounded wait overlaps real work rather than stalling the event loop on its own.

---

PR #3241, opened 2026-08-19
fix(cli): reject a leading dash in cluster `zenoh_peer`

phil-opp:
## Issue

`ClusterConfig::validate` (`binaries/cli/src/command/cluster/config.rs`) runs the `zenoh_peer` field through `validate_endpoint_shell_safe`. Its call-site comment states the field *"needs the same guard as the id/labels below"*, but the function only performed a charset check — it was **missing the leading-dash guard** that the sibling `validate_shell_safe` (used for `machine.id` and label keys, with `at_token_start = true`) applies.

`zenoh_peer` is interpolated as the whole argument of the remote command:

```
dora daemon ... --zenoh-peer {ep} ...        # binaries/cli/src/command/cluster/up.rs:70
```

So a value like `zenoh_peer: "-tcp/1.2.3.4:5456"` passes validation, but on the remote host the daemon's clap parser (`--zenoh-peer` has no `allow_hyphen_values`) treats `-tcp/...` as an unknown **option flag** rather than the endpoint. The daemon then fails to start, and the user only sees a confusing *"daemon(s) did not register"* timeout later — exactly the failure mode the id/labels leading-dash guard was added to prevent (see the existing `reject_id_with_leading_dash` / `reject_label_key_with_leading_dash` tests).

## Fix

Add the same leading-dash rejection to `validate_endpoint_shell_safe`, so `zenoh_peer` is now validated consistently with `machine.id` and label keys. No legitimate Zenoh endpoint (`tcp/…`, `udp/…`, `tls/…`, `quic/…`, `tcp/[::1]:7447`, …) begins with `-`, so nothing valid is rejected.

```rust
if value.starts_with('-') {
    bail!("{what} must not start with `-`");
}
```

A regression test (`reject_zenoh_peer_with_leading_dash`) covers `-tcp/1.2.3.4:5456`, `--zenoh-peer`, and `-`. The existing `accept_ipv6_zenoh_peer` test confirms `tcp/[::1]:7447` still validates.

## Validation

- `cargo test -p dora-cli --lib -- config::tests` → **25 passed, 0 failed** (incl. the new test)
- `cargo fmt -p dora-cli -- --check` → clean
- `cargo clippy -p dora-cli --all-targets -- -D warnings` → clean

Scoped to `binaries/cli/src/command/cluster/config.rs` (one guard clause + one test); no behavior change for any valid endpoint.

---

⚠️ *This is a machine-generated pull request authored by Claude (Claude Code). A human should review it before merging.*

---
_Generated by [Claude Code](https://claude.ai/code/session_01GBFpa44oHdDJxVTphFBTsV)_

### DIFF
--- a/binaries/cli/src/command/cluster/config.rs
+++ b/binaries/cli/src/command/cluster/config.rs
@@ -176,6 +176,14 @@ fn validate_no_leading_dash(what: &str, value: &str) -> eyre::Result<()> {
 /// `[`/`]` around an IPv6 literal like `tcp/[::1]:7447`. It still rejects
 /// whitespace and shell metacharacters (`;`, `|`, `$`, backtick, ...), so the
 /// value cannot break out of the remote command.
+///
+/// It also rejects a leading `-`, the same guard `validate_shell_safe` applies
+/// at a token start: `zenoh_peer` is interpolated as the whole argument of
+/// `dora daemon --zenoh-peer {ep}`, so a value like `-tcp/1.2.3.4:5456` is
+/// parsed by the remote daemon's clap as an option flag instead of the
+/// endpoint, and the daemon never starts — surfacing only later as a confusing
+/// "daemon(s) did not register" timeout. No legitimate Zenoh endpoint begins
+/// with `-`.
 fn validate_endpoint_shell_safe(what: &str, value: &str) -> eyre::Result<()> {
     if let Some(ch) = value.chars().find(|c| {
         !c.is_ascii_alphanumeric() && !matches!(c, '_' | '-' | '.' | '/' | ':' | '[' | ']')
@@ -183,6 +191,9 @@ fn validate_endpoint_shell_safe(what: &str, value: &str) -> eyre::Result<()> {
         bail!(
             "{what} contains invalid character `{ch}` -- only [a-zA-Z0-9_.:/] and `[`/`]` are allowed"
         );
+    }
+    if value.starts_with('-') {
+        bail!("{what} must not start with `-`");
     }
     Ok(())
 }
@@ -416,6 +427,24 @@ mod tests {
             err.contains("invalid character") && err.contains("zenoh_peer"),
             "malicious zenoh_peer should be rejected, got: {err}"
         );
+    }
+
+    #[test]
+    fn reject_zenoh_peer_with_leading_dash() {
+        // A leading `-` passes the endpoint charset check but makes the remote
+        // `dora daemon --zenoh-peer {ep}` parse the endpoint as an option flag,
+        // so the daemon silently fails to start. Reject it up front, matching
+        // the guard on the machine id / label key.
+        for bad in ["-tcp/1.2.3.4:5456", "--zenoh-peer", "-"] {
+            let f = write_yaml(&format!(
+                "coordinator:\n  addr: 10.0.0.1\nzenoh_peer: \"{bad}\"\nmachines:\n  - id: a\n    host: h\n"
+            ));
+            let err = ClusterConfig::load(f.path()).unwrap_err().to_string();
+            assert!(
+                err.contains("must not start with `-`") && err.contains("zenoh_peer"),
+                "zenoh_peer `{bad}` should be rejected, got: {err}"
+            );
+        }
     }
 
     #[test]


phil-opp:
## Automated repository review — run summary

This is the summary comment for a machine-generated (Claude Code) in-depth review pass over the repository at `origin/main` (`5938a21`). Each finding below was verified by hand against current `main`, and cross-checked against the existing open Claude PRs to avoid duplicating already-reported issues. **All PRs are drafts and should be reviewed by a human before merging.**

### PRs opened this run

**Correctness**
- **#3241** — `fix(cli): reject a leading dash in cluster zenoh_peer`. `validate_endpoint_shell_safe` checked the endpoint charset but, unlike the sibling guard for `machine.id`/label keys, omitted the leading-dash check its own comment promised. A `zenoh_peer` like `-tcp/…` is parsed by the remote daemon's clap as an option flag, so the daemon silently fails to start.
- **#3242** — `fix(metrics): honor DORA_OTLP_ENDPOINT for system/process metrics`. `dora_metrics::init_metrics` built its OTLP exporter without `.with_endpoint(...)`, so system/process metrics always went to `localhost:4317` even when `DORA_OTLP_ENDPOINT` pointed at a remote collector (traces already honored it). The sibling `dora_tracing` meter provider had already fixed this; `dora-metrics` — the crate actually wired into the runtime/node startup paths — had not.

### Rejected / not-pursued findings (with reasons)

- **`local_listener.rs` unexpected-request arm never replies** (`binaries/daemon`) — real inconsistency vs. `node_communication` (which replies with a loud error), but not reachable via the current node API, which only ever sends `NodeConfig` on that socket. Speculative; skipped.
- **`is_compatible` universal-`Bytes` sink matches the raw string, not the parsed base URN** (`libraries/core/src/types.rs`) — cosmetic; parameterized `Bytes` targets don't occur in practice.
- **`versions_compatible` uses caret semantics on pre-release versions** (`libraries/message/src/lib.rs`) — potential sharp edge, but backstopped by the exact `metadata_version` equality check, so no live desync.
- **mavlink `emit!` conflates a message-encode error with a fatal link loss** (`binaries/mavlink2-bridge-node`) — not triggerable: `to_record_batch` cannot fail for these fixed-schema messages.
- **`RawData::Vec` builds a `Buffer` from `data.as_ptr()` with `unwrap()`** (`apis/rust/node/.../data_conversion.rs`) — defensive-only; `AVec` never yields a null pointer even at length 0, so the `unwrap` is unreachable.
- **Doc-gap doctests** (`fnv1a`, `debug_frame_wire_size`, … in `dora-message`) — these already carry prose docs + unit tests with the same vectors; a doctest would add little, and the crate is documented as internal-only.
- **`metrics` gate ignores `DORA_JAEGER_TRACING`** — a `DORA_JAEGER_TRACING`-only user gets traces but no metrics monitor. Noted in #3242 as an out-of-scope follow-up to keep that fix focused.

### Method / scope note

The review fanned out across core/message, daemon/coordinator/CLI, node/operator/mavlink APIs, and the runtime/telemetry/shared-memory + hot-path code. The overwhelming majority of the codebase is already hardened (issue-referenced regression tests, guarded arithmetic/indexing, poison-recovery, whole-stream size checks), and prior automated runs have already landed many of the low-hanging fixes — hence a deliberately small, high-confidence PR set rather than a padded one. Per-branch validation: `fmt` + `clippy -D warnings` + affected-crate tests (metrics crates built/linted with the `metrics` feature enabled, since the changed code is feature-gated).

---

phil-opp:
Reviewed the diff. The added leading-dash guard in `validate_endpoint_shell_safe` is correctly ordered after the charset check (which permits `-`), and it matches the existing leading-dash guards on `machine.id`/label keys. No legitimate zenoh endpoint begins with `-` (tcp/udp/tls/quic, IPv6 literals all start with a letter or `[`), so there's no valid-input regression — the `accept_ipv6_zenoh_peer` test guards that direction. The new `reject_zenoh_peer_with_leading_dash` test drives the full `ClusterConfig::load` parse+validate path and asserts on both the message and field name.

No issues found.

---

---

phil-opp:
The merge-queue `Unwrap budget` failure is **not** from this PR — no change is needed here.

The unwrap-budget ratchet is a tree-wide count of production `.unwrap()`/`.expect(`. It failed on the merge queue's **batch** commit (`99da0dae`), which merged this PR *together with other queued PRs* — the batch diff modifies `binaries/cli/src/output.rs` (from a sibling PR, e.g. #3223), which this PR does not touch. On the batch the count was `162 > 161`; on this branch alone it is `159` (identical to `origin/main`, margin of 2). This PR's diff is a single validation guard clause plus a test, and adds zero production unwraps.

Trunk's status ("waiting for other pull requests to finish testing") indicates it will bisect the batch and re-test this PR individually, where the check passes. Leaving the clean diff as-is.

---

PR #3242, opened 2026-08-19
fix(metrics): honor `DORA_OTLP_ENDPOINT` for system/process metrics

phil-opp:
## Issue

`dora_metrics::init_metrics` (`libraries/extensions/telemetry/metrics/src/lib.rs`) built its OTLP `MetricExporter` **without** calling `.with_endpoint(...)`:

```rust
let exporter = MetricExporter::builder()
    .with_tonic()
    .build()?;          // no endpoint → OTLP gRPC default http://localhost:4317
```

Both hot callers gate the metrics monitor on `DORA_OTLP_ENDPOINT` being set, but never passed that value down:

- `binaries/runtime-api/src/lib.rs` — `if std::env::var("DORA_OTLP_ENDPOINT").is_ok() { … run_metrics_monitor(meter_id) … }`
- `apis/rust/node/src/node/mod.rs` — same gate, then `run_metrics_monitor(id)`

So with `DORA_OTLP_ENDPOINT=http://remote-collector:4317`, **traces** reached the remote collector (the trace exporter threads the endpoint through `dora_tracing::telemetry::init_tracing`), but **system/process metrics** (CPU/memory/disk) silently went to `localhost:4317` on the node/runtime host and were dropped if nothing listened there.

This is the exact defect the sibling crate already fixed — see the doc comment on `dora_tracing::telemetry::init_meter_provider` (`libraries/extensions/telemetry/tracing/src/metrics.rs`), which notes it passes the endpoint explicitly to avoid *"the previous behaviour where the metric exporter ignored `DORA_OTLP_ENDPOINT` and always fell back to the OTLP gRPC default"*. That fix landed in the tracing crate's meter provider but not in `dora-metrics`, which is the one wired into the runtime/node startup paths.

## Fix

Thread the endpoint through, mirroring the tracing crate:

- `init_metrics(endpoint: &str)` and `run_metrics_monitor(meter_id, endpoint: &str)` now call `.with_endpoint(endpoint)`.
- The two gated callers (`runtime-api`, `node`) already read `DORA_OTLP_ENDPOINT` — they now capture the value (`if let Ok(endpoint) = …`) and pass it through.
- The coordinator (`binaries/coordinator/src/lib.rs`) inits metrics **unconditionally** under the `metrics` feature; it now reads `DORA_OTLP_ENDPOINT` with a `http://localhost:4317` fallback, so its previous behaviour is preserved when the env var is unset while honoring it when set.

The exporter still honors the OTel-standard `OTEL_EXPORTER_OTLP_ENDPOINT` / `..._METRICS_ENDPOINT` env vars; this change only restores dora's own `DORA_OTLP_ENDPOINT` on the metrics path.

Out of scope (noted for a follow-up): the metrics gate checks only `DORA_OTLP_ENDPOINT`, whereas the trace path also accepts `DORA_JAEGER_TRACING` — a `DORA_JAEGER_TRACING`-only user gets traces but no metrics monitor. Left unchanged to keep this fix focused on the endpoint threading.

## Validation

Built and linted the metrics crate and every caller **with the `metrics` feature enabled** (the changed code is behind `#[cfg(feature = "metrics")]`):

- `cargo build -p dora-metrics -p dora-runtime-api --features dora-runtime-api/metrics -p dora-coordinator --features dora-coordinator/metrics` → OK
- `cargo build -p dora-node-api --features metrics` → OK
- `cargo clippy` on all four with the feature, `-D warnings` → clean
- `cargo fmt … -- --check` → clean

`dora_metrics` is documented as internal (not covered by the 1.0 stability guarantee), so the signature change to `init_metrics`/`run_metrics_monitor` is an internal-only change.

---

⚠️ *This is a machine-generated pull request authored by Claude (Claude Code). A human should review it before merging.*

---
_Generated by [Claude Code](https://claude.ai/code/session_01GBFpa44oHdDJxVTphFBTsV)_

### DIFF
--- a/apis/rust/node/src/node/mod.rs
+++ b/apis/rust/node/src/node/mod.rs
@@ -3236,12 +3236,12 @@ pub fn init_tracing(
     // attempt to connect to `localhost:4317` on every node startup. Mirrors
     // the gating applied to tracing above.
     #[cfg(feature = "metrics")]
-    if std::env::var("DORA_OTLP_ENDPOINT").is_ok() {
+    if let Ok(endpoint) = std::env::var("DORA_OTLP_ENDPOINT") {
         let id = format!("{dataflow_id}/{node_id}");
         let monitor_task = async move {
             use dora_metrics::run_metrics_monitor;
 
-            if let Err(e) = run_metrics_monitor(id.clone())
+            if let Err(e) = run_metrics_monitor(id.clone(), &endpoint)
                 .await
                 .wrap_err("metrics monitor exited unexpectedly")
             {
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -253,12 +253,12 @@ async fn start_with_events(
     // Start WS server
     #[cfg(feature = "metrics")]
     let _meter_provider = {
-        // Honor `DORA_OTLP_ENDPOINT` when set, falling back to the OTLP gRPC
-        // default so the previous (endpoint-less) behaviour is preserved when
-        // it is unset.
-        let endpoint = std::env::var("DORA_OTLP_ENDPOINT")
-            .unwrap_or_else(|_| "http://localhost:4317".to_string());
-        let provider = dora_metrics::init_metrics(&endpoint)?;
+        // Pin the exporter to `DORA_OTLP_ENDPOINT` when set; when unset, pass
+        // `None` so the previous (endpoint-less) behaviour is preserved — the
+        // exporter then resolves its target from the OTel-standard
+        // `OTEL_EXPORTER_OTLP_ENDPOINT` env vars (defaulting to localhost).
+        let endpoint = std::env::var("DORA_OTLP_ENDPOINT").ok();
+        let provider = dora_metrics::init_metrics(endpoint.as_deref())?;
         opentelemetry::global::set_meter_provider(provider.clone());
         provider
     };
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -253,7 +253,12 @@ async fn start_with_events(
     // Start WS server
     #[cfg(feature = "metrics")]
     let _meter_provider = {
-        let provider = dora_metrics::init_metrics()?;
+        // Honor `DORA_OTLP_ENDPOINT` when set, falling back to the OTLP gRPC
+        // default so the previous (endpoint-less) behaviour is preserved when
+        // it is unset.
+        let endpoint = std::env::var("DORA_OTLP_ENDPOINT")
+            .unwrap_or_else(|_| "http://localhost:4317".to_string());
+        let provider = dora_metrics::init_metrics(&endpoint)?;
         opentelemetry::global::set_meter_provider(provider.clone());
         provider
     };
--- a/binaries/runtime-api/src/lib.rs
+++ b/binaries/runtime-api/src/lib.rs
@@ -181,12 +181,12 @@ async fn run(
     // so it must be spawned rather than awaited inline. Mirrors the gating and
     // spawning used by the node API (`apis/rust/node/src/node/mod.rs`).
     #[cfg(feature = "metrics")]
-    if std::env::var("DORA_OTLP_ENDPOINT").is_ok() {
+    if let Ok(endpoint) = std::env::var("DORA_OTLP_ENDPOINT") {
         use dora_metrics::run_metrics_monitor;
 
         let meter_id = config.node_id.to_string();
         tokio::spawn(async move {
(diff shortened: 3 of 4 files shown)


phil-opp:
Reviewed the diff. All three call sites of `init_metrics`/`run_metrics_monitor` (node, runtime-api, coordinator) are updated, so no caller is left on the old signature. The coordinator keeps a `http://localhost:4317` fallback, so prior behavior is preserved when `DORA_OTLP_ENDPOINT` is unset, while the two gated callers only run when the var is set (so `&endpoint` is always a real value). The signature change is on the internal `dora-metrics` crate, so it's not a public-API break.

No issues found.

---

---

phil-opp (commented):
🤖 **Automated review by Claude** — this review was generated fully automatically; no human has vetted it.

I found one issue:

**The coordinator now overrides `OTEL_EXPORTER_OTLP_ENDPOINT` when `DORA_OTLP_ENDPOINT` is unset** — `binaries/coordinator/src/lib.rs:256-261`

The coordinator initializes metrics unconditionally and now always passes an endpoint to `init_metrics()`: either the `DORA_OTLP_ENDPOINT` value, or the hardcoded `http://localhost:4317` fallback when it's unset. `init_metrics()` forwards that to `MetricExporter::builder().with_endpoint(...)`.

Previously `init_metrics()` set no endpoint, so the OTLP exporter resolved its target from the OTel-standard `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` env vars, falling back to localhost only when those were also unset. A programmatic `.with_endpoint()` takes precedence over those env vars, so after this change a coordinator configured solely via `OTEL_EXPORTER_OTLP_ENDPOINT` (with `DORA_OTLP_ENDPOINT` unset) will silently send its system/process metrics to `http://localhost:4317` instead of the configured collector.

That's a regression from current behavior, and it contradicts the new doc comment's claim that "the previous (endpoint-less) behaviour is preserved when it is unset" — the endpoint-less behavior honored the OTel env vars, the new code does not.

Suggested fix: only call `.with_endpoint()` when `DORA_OTLP_ENDPOINT` is actually set (e.g. thread an `Option<&str>` and skip the builder call when `None`), so the OTel-standard env vars keep working as the fallback.

(The node and runtime-api paths are gated on `DORA_OTLP_ENDPOINT` being set, so they aren't affected by this.)

---

phil-opp:
Good catch — this is correct and I've fixed it in `ec7cbbf`.

The regression was exactly as described: a programmatic `.with_endpoint()` takes precedence over `OTEL_EXPORTER_OTLP_ENDPOINT` / `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`, so the hardcoded `http://localhost:4317` fallback would have overridden a coordinator configured solely via the OTel-standard vars — and my doc comment's "previous behaviour is preserved" claim was wrong, since the endpoint-less path resolved from those vars.

Fix: `init_metrics` now takes `Option<&str>` and only calls `.with_endpoint()` when `Some`. The coordinator passes `DORA_OTLP_ENDPOINT` as an `Option` (`.ok()`), so when it's unset no endpoint is pinned and the OTel-standard vars keep working as the fallback. The node/runtime-api paths are gated on `DORA_OTLP_ENDPOINT` being set, so they always pass `Some` and are unaffected.

---

phil-opp:

The latest commit (`ec7cbbf`) resolves the earlier concern about the coordinator overriding the OTel-standard `OTEL_EXPORTER_OTLP_ENDPOINT` variables. `init_metrics` now takes `Option<&str>` and only calls `.with_endpoint()` when `Some`, and the coordinator passes `DORA_OTLP_ENDPOINT` via `.ok()`, so when it's unset no endpoint is pinned and the OTel-standard vars keep resolving the target. The two gated callers (node, runtime-api) always pass a real endpoint value. All three call sites are updated and none is left on the old signature. No further issues found.

---

phil-opp:
The merge-queue `Unwrap budget` failure is **not** from this PR — no change is needed here.

The unwrap-budget ratchet is a tree-wide count of production `.unwrap()`/`.expect(`. It failed on the merge queue's **batch** commit (`99da0dae`), which merged this PR *together with other queued PRs* — the batch diff modifies `binaries/cli/src/output.rs` (from a sibling PR, e.g. #3223), which this PR does not touch. On the batch the count was `162 > 161`; on this branch alone it is `159` (identical to `origin/main`, margin of 2). This PR only threads an `Option<&str>` endpoint through `init_metrics`/`run_metrics_monitor` and reads an env var — it adds zero production unwraps.

Trunk's status ("waiting for other pull requests to finish testing") indicates it will bisect the batch and re-test this PR individually, where the check passes. Leaving the clean diff as-is.

---

PR #3244, opened 2026-08-20
fix(daemon): fold local exit into the release-path start verdict

### LINKED ISSUE #3243: daemon: #3052 fix is incomplete — the release-time path (`handle_external_all_nodes_ready` / `AllNodesReady` handler) still starts the dataflow while failing a parked cohort survivor (local `exited_before_subscribe` ignored by the start verdict)
**This issue was filed by an automated, scheduled Claude review of the codebase. Please verify before acting.**

## Summary

The fix for #3052 (PR #3055, `fold local exit-before-subscribe into the latched barrier start verdict`) closed the **late-subscriber** path (`PendingNodes::update_dataflow_status`) but left the **structurally-parallel release-time path** unfixed. When the coordinator's `AllNodesReady` is received/replayed and a cohort subscriber is **already parked**, the daemon still decides `dataflow.start()` from the coordinator's external failure list **only**, while `answer_subscribe_requests` fails the parked subscriber based on this daemon's **local** `exited_before_subscribe`. The two halves disagree — exactly the inconsistency #3052 set out to remove — so the dataflow can start while simultaneously failing a healthy surviving cohort node's `Node()` init.

This is reachable via **the very reproduction scenario in #3052**, one step earlier than the case #3055 fixed (it bites the already-parked subscriber at reconnect/replay time, before any late subscribe).

## Affected code

`binaries/daemon/src/pending.rs:220-243` — `handle_external_all_nodes_ready` answers parked subscribers but returns only `Ok(())`, carrying **no** local-failure signal to the caller:

```rust
pub async fn handle_external_all_nodes_ready(
    &mut self,
    exited_before_subscribe: Vec<NodeId>,
    cascading_errors: &mut CascadingErrorCauses,
) -> eyre::Result<()> {
    // ...
    self.external_ready = Some(exited_before_subscribe.clone());
    self.answer_subscribe_requests(exited_before_subscribe, cascading_errors).await; // local list wins here
    Ok(())   // <-- no barrier verdict returned
}
```

`binaries/daemon/src/lib.rs:2617-2640` — the `AllNodesReady` handler decides `start()` from the **external** list alone, never consulting `pending_nodes.exited_before_subscribe`:

```rust
Some(dataflow) => {
    let ready = exited_before_subscribe.is_empty();          // <-- external list ONLY
    dataflow.pending_nodes
        .handle_external_all_nodes_ready(exited_before_subscribe, &mut dataflow.cascading_error_causes)
        .await?;
    if ready && !dataflow.stop_sent {
        dataflow.start(&self.events_tx, &self.clock).await?; // <-- starts the dataflow
    }
}
```

`binaries/daemon/src/pending.rs:315-335` — `answer_subscribe_requests` prioritizes this daemon's **local** `exited_before_subscribe`, so a parked cohort member is answered with `Err("Node <X> exited before initializing dora ...")`:

```rust
let node_exited_before_subscribe = match self.exited_before_subscribe.as_slice() {
    [first, ..] => Some(first),                              // local failure takes priority
    [] => match exited_before_subscribe_external.as_slice() {
        [first, ..] => Some(first),
        [] => None,
    },
};
```

So when `self.exited_before_subscribe = [X]` but the incoming/replayed `exited_before_subscribe = []`:
- `lib.rs` sees `ready = true` → calls `dataflow.start()`.
- `answer_subscribe_requests` names local node `X` → every **parked cohort** subscriber (a healthy node `A`) gets `Err` and fails its own `Node()` init.

The sibling path was fixed to fold the local list into its verdict (`pending.rs:287-288`, added by #3055):

```rust
let barrier_succeeded =
    exited_before_subscribe.is_empty() && self.exited_before_subscribe.is_empty();
```

…but the release path above has no equivalent fold-in.

## How it is reached (the #3052 scenario, at the reconnect/replay step)

Same setup as #3052, but the bug fires on the **already-parked** subscriber `A` at reconnect (its step 5), which #3055 does not guard:

1. Multi-daemon dataflow. Daemon `D` local cohort `{A, X, Y}` (+ external nodes elsewhere). `A` subscribes; `X`, `Y` still pending. `A` parks in `waiting_subscribers`; `update_dataflow_status` returns `Pending` (local_nodes non-empty), so `A` is **not** answered. `D` has not reported ready.
2. `D` briefly disconnects from the coordinator (nodes keep running, #2029).
3. As the last pending daemon, `D`'s disconnect makes the coordinator release the ready barrier as **success** with `exited_before_subscribe = []` (it never learned of any `D` loss). Persisted.
4. Still disconnected, `D`'s node `X` exits before subscribing → `handle_node_stop` pushes `X` into `self.exited_before_subscribe`. `local_nodes = {Y}` still non-empty → `Pending`, `A` stays parked.
5. `D` reconnects → coordinator **replays** `AllNodesReady([])` → `lib.rs` `AllNodesReady` handler:
   - `ready = [].is_empty() = true`.
   - `handle_external_all_nodes_ready([])` → `answer_subscribe_requests([])` picks local `X` → **parked cohort member `A` is answered `Err("Node X exited before initializing dora ...")`** and dies at `Node()` init.
   - `ready && !stop_sent` → **`dataflow.start()` runs.**

Result: a partially-started dataflow — timers spawn and `A` is killed at init, at the same instant — the exact "survivor left on an inconsistently started dataflow" class #2970/#2967/#3052 aimed to eliminate. #3055's `update_dataflow_status` fold-in does not prevent this because the start decision and `A`'s reply both flow through the release path, not through `update_dataflow_status`. (A late subscriber `Y` arriving after step 5 is now correctly held `Pending` by #3055; but the inconsistent start already happened at step 5.)

## Suggested fix

Make the release path enforce the same combined verdict the late-subscriber path already does — keep the "start verdict == subscriber reply" invariant in one place for both paths. For example, have `handle_external_all_nodes_ready` return a `DataflowStatus` computed as

```rust
let barrier_succeeded =
    exited_before_subscribe.is_empty() && self.exited_before_subscribe.is_empty();
```

and gate the `dataflow.start()` in the `lib.rs` `AllNodesReady` handler on that status instead of the raw external-only `ready` flag. Minimally:

```rust
let ready = exited_before_subscribe.is_empty()
    && dataflow.pending_nodes.exited_before_subscribe_is_empty(); // new accessor
```

## Regression test

In `pending.rs` `mod tests`, mirror `latch_verdict_respects_a_local_exit_the_coordinator_never_saw` (added by #3055) but drive it through `handle_external_all_nodes_ready` with a **pre-parked** cohort subscriber: seed `self.exited_before_subscribe = [X]`, park a cohort member `A`, call `handle_external_all_nodes_ready(vec[])`, and assert the resulting start verdict is not "ready" while `A` receives the `Err`. (The current code answers `A` with `Err` yet still returns a "ready" verdict to the caller.)

## Confidence

Medium-high. The code-level inconsistency (external-only start verdict at `lib.rs:2619` vs. local-first subscriber replies in `answer_subscribe_requests`) is unambiguous and localized, and it is the same class #3052 documented. Reachability is the #3013/#3052 reconnect-release scenario, which the surrounding code comments (`pending.rs:280-286`) and the `lib.rs` handler comment ("the coordinator replays `AllNodesReady` to a daemon that reconnects") both treat as a real path; the only added requirement here is that a cohort subscriber is already parked at replay time, which holds in #3052's own step 1.

phil-opp:
## Summary

Fixes #3243.

The `#3052` fix (PR #3055) closed the **late-subscriber** path but left the structurally-parallel **release-time** path unguarded. When the coordinator's `AllNodesReady` is received/replayed and a cohort subscriber is already parked, the daemon decided `dataflow.start()` from the coordinator's **external** `exited_before_subscribe` list only, while `answer_subscribe_requests` fails the parked subscriber based on this daemon's **local** `exited_before_subscribe`. The two halves disagreed — exactly the inconsistency #3052 set out to remove — so the dataflow could start while simultaneously failing a healthy surviving cohort node's `Node()` init.

This is reachable via the #3052 reconnect/replay scenario, one step earlier than the case #3055 fixed: it bites the already-parked survivor at reconnect time. A cohort member that dies before subscribing while the daemon is disconnected is invisible to the coordinator, so the replayed barrier carries an empty external list — `ready` computed as `exited_before_subscribe.is_empty()` was `true`, and `start()` ran.

## Change

- `PendingNodes::handle_external_all_nodes_ready` now returns a `DataflowStatus` computed from the **combined** verdict `external.is_empty() && self.exited_before_subscribe.is_empty()`, mirroring the late-subscriber fold-in already present in `update_dataflow_status` (added by #3055). This keeps the "start verdict == subscriber reply" invariant in one place for both paths.
- The `lib.rs` `AllNodesReady` handler gates `start()` on the returned status instead of the raw external-only flag. The existing `!stop_sent` / idempotency semantics are preserved.

`start()` is now suppressed in exactly the one buggy case — external list empty but local list non-empty — where the dataflow's startup has actually failed and it should tear down rather than start.

## Test

Adds `release_verdict_respects_a_local_exit_with_a_pre_parked_survivor` in `pending.rs`, mirroring `latch_verdict_respects_a_local_exit_the_coordinator_never_saw` (from #3055) but driving the release path with a **pre-parked** cohort survivor and a local exit the coordinator never saw. It asserts the verdict is `Pending` (not `AllNodesReady`) and that the parked survivor is answered with the `Err` naming the locally-dead node.

## Validation

- `cargo test -p dora-daemon --lib pending::` — all 20 pass, including the new regression test
- `cargo clippy -p dora-daemon -- -D warnings` (CI toolchain 1.97.1) — clean

https://claude.ai/code/session_01Byrb4vD4xGfQkEZfCXkano

---
_Generated by [Claude Code](https://claude.ai/code/session_01Byrb4vD4xGfQkEZfCXkano)_

### DIFF
--- a/binaries/daemon/src/lib.rs
+++ b/binaries/daemon/src/lib.rs
@@ -2616,14 +2616,23 @@ impl Daemon {
                 )).await;
                 match self.running.get_mut(&dataflow_id) {
                     Some(dataflow) => {
-                        let ready = exited_before_subscribe.is_empty();
-                        dataflow
+                        // The verdict must fold in this daemon's local
+                        // `exited_before_subscribe`, not just the coordinator's
+                        // external list: a cohort member that died before
+                        // subscribing while this daemon was disconnected is
+                        // invisible to the replayed external list, but
+                        // `answer_subscribe_requests` still fails the parked
+                        // survivors on it. Deciding `start()` from the external
+                        // list alone would start the dataflow while failing a
+                        // healthy cohort node's init (dora-rs/dora#3243).
+                        let status = dataflow
                             .pending_nodes
                             .handle_external_all_nodes_ready(
                                 exited_before_subscribe,
                                 &mut dataflow.cascading_error_causes,
                             )
                             .await?;
+                        let ready = matches!(status, crate::pending::DataflowStatus::AllNodesReady);
                         // Not gated on `dataflow_started` — `start()` is
                         // idempotent — but it must not resurrect a dataflow that
                         // is already tearing down (dora-rs/dora#3053). The
--- a/binaries/daemon/src/pending.rs
+++ b/binaries/daemon/src/pending.rs
@@ -221,7 +221,7 @@ impl PendingNodes {
         &mut self,
         exited_before_subscribe: Vec<NodeId>,
         cascading_errors: &mut CascadingErrorCauses,
-    ) -> eyre::Result<()> {
+    ) -> eyre::Result<DataflowStatus> {
         if !self.local_nodes.is_empty() {
             tracing::warn!(
                 remaining = ?self.local_nodes,
@@ -229,6 +229,20 @@ impl PendingNodes {
                  proceeding anyway (coordinator may be ahead of this daemon)"
             );
         }
+
+        // The start verdict must agree with the reply each parked subscriber
+        // actually receives. `answer_subscribe_requests` (below) prioritizes
+        // this daemon's LOCAL `exited_before_subscribe`, so fold it into the
+        // verdict here — otherwise a local cohort loss the coordinator never
+        // learned about (a node that died before subscribing while this daemon
+        // was disconnected, so the replayed barrier carries an empty external
+        // list — #3013/#3052) would start the dataflow while simultaneously
+        // failing a surviving cohort node's `Node()` init. This mirrors the
+        // late-subscriber fold-in in `update_dataflow_status`, keeping the
+        // "start verdict == subscriber reply" invariant on the release path too
+        // (dora-rs/dora#3243).
+        let barrier_succeeded =
+            exited_before_subscribe.is_empty() && self.exited_before_subscribe.is_empty();
 
         // Latch before answering: the coordinator fires this once, so
         // everything that subscribes from here on has to be answered
@@ -239,7 +253,11 @@ impl PendingNodes {
         self.answer_subscribe_requests(exited_before_subscribe, cascading_errors)
             .await;
 
-        Ok(())
+        Ok(if barrier_succeeded {
+            DataflowStatus::AllNodesReady
+        } else {
+            DataflowStatus::Pending
+        })
     }
 
     /// Re-evaluate the barrier after something changed the pending set.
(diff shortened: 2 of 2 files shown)


phil-opp:
Reviewed the diff. Folding the local `exited_before_subscribe` check into the returned `DataflowStatus` mirrors the existing late-subscriber logic, and the three cases hold: healthy (both lists empty) → start unchanged; remote death (external non-empty) → no start unchanged; the bug case (external empty, local exit) → now `Pending`/no start while `answer_subscribe_requests` fails the parked survivor. The `ready && !stop_sent` gating at the call site is preserved.

The regression test genuinely exercises the changed path — it parks a survivor, records a local-only exit the coordinator never saw, replays `AllNodesReady([])`, and asserts the status is `Pending` (would be `AllNodesReady`/start without the fix) plus that the survivor gets an `Err`.

No issues found.

---

---

phil-opp:
The `Unwrap budget` failure is a merge-queue **batch** artifact, not a problem with this PR.

- The queue tested a combined branch (`trunk-merge/pr-3241/…`) and the CI log reports `unwrap/expect count: 162 (budget: 161)`.
- This PR standalone is at **159 ≤ 161** and adds **zero** production `.unwrap()`/`.expect()` — its only `.expect(` additions are inside a `#[cfg(test)]` test, which `scripts/qa/unwrap-budget.sh` excludes. `origin/main` is also at 159, so this branch changes the count by 0.
- The +3 over budget therefore comes from the other PR(s) batched ahead of this one; whichever adds the unwraps needs to bump `.unwrap-budget` in its own commit.

No change is warranted here — bumping the shared budget for unwraps this PR didn't add would defeat the ratchet. This PR should pass once the queue re-tests it on its own or without the offending batch member.

---

Issue #2030, opened 2026-06-04
C and C++ node APIs todo!()-abort on any non-UInt8 input


heyong4725:
## Summary

The C and C++ node APIs `todo!()`-abort (forced process abort) on any input whose Arrow type isn't UInt8 — a routine cross-language case (any Int32/Float/etc. payload from another node).

- **C**: `apis/c/node/src/lib.rs:195-198` — `read_dora_input_data` hits `todo!()` on any non-UInt8/Null input; there is no Arrow-FFI escape hatch.
- **C++**: `apis/c++/node/src/lib.rs:434-437` — `event_as_input` `todo!()`s via the cxx bridge; the function already returns a `Result`.

## Suggested fix

- C: write a null pointer + len 0 and `tracing::error!` the unsupported type instead of aborting. Test: non-UInt8 input → `out_ptr == NULL`, `out_len == 0`, no abort.
- C++: `bail!("unsupported input arrow type {:?}; use event_as_arrow_input for typed access", ...)`. Test: Int32 Input → `Err`.

Part of a broader "no panic across an `extern "C"`/cxx boundary" convention worth establishing across the C/C++/Python APIs (audit Deeper §6.3).

---
_From the 2026-06-04 soundness audit (`docs/audit-2026-06-04-soundness.md`, High §3). Adversarially re-verified against source._

PR #3246, opened 2026-08-20
fix(operator-api): log the specific failure in dora_read_data instead…

### LINKED ISSUE #3245: C/C++ operator FFI's dora_read_data silently swallows all failures with zero diagnostics
## Problem

`apis/rust/operator/types/src/lib.rs:162-169`:

    pub fn dora_read_data(input: &mut Input) -> Option<safer_ffi::Vec<u8>> {
        let data_array = input.data_array.take()?;
        let data = unsafe { arrow::ffi::from_ffi(data_array, &input.schema).ok()? };
        let array = ArrowData(arrow::array::make_array(data));
        let bytes: &[u8] = TryFrom::try_from(&array).ok()?;
        Some(bytes.to_owned().into())
    }

Three unrelated failure modes collapse into an indistinguishable `None`, with
**no logging at all**:

1. `input.data_array.take()?` — input already consumed (double-read)
2. `arrow::ffi::from_ffi(...).ok()?` — malformed/incompatible Arrow FFI schema
3. `TryFrom::try_from(&array).ok()?` — non-`UInt8` payload (e.g. a routine
   `Int32`/`Float32` array from an upstream Rust/Python node)

Compare `apis/c/node/src/lib.rs:202-217` (`read_dora_input_data`), the
equivalent function for C **nodes**, which explicitly `tracing::error!`s the
failing Arrow type before returning null. The operator-side twin never got
this treatment — every other FFI accessor across the C/C++ node/operator
surface already has it (tracked as #2030), this one was missed.

## Evidence

Reproduced by calling `dora_read_data` directly (out-of-tree scratch crate
path-depending on the real `dora-operator-api-types` + `arrow-convert`
crates) with an `Int32Array`-backed `Input` — the routine cross-language
case. Result: `None`, zero diagnostic output.

## Why it matters

C/C++ operators are a first-class dora node type. A real integration bug
("my C operator keeps getting empty payloads") becomes nearly undebuggable
without manually diffing Arrow schemas at the wire level.

## Not a duplicate

Searched `gh issue/pr list --state all --search` across `"dora_read_data"`,
variations of "returns null / error swallowed / unsupported type / silently
dropped", etc. — no matches. This item is documented internally in
`docs/audit-2026-06-04-soundness.md:97` as a known-but-never-filed Medium
finding (distinct from the already-fixed `examples/c-dataflow/operator.c`
buffer leak, and distinct from the 8 High findings already filed as
#2028-#2034).

## Proposed fix

Replace the `?`-chain with explicit branches, each with a `tracing::error!`
naming the specific failure (mirroring `read_dora_input_data`), plus a unit
test analogous to the node API's `read_dora_input_data_non_uint8_returns_null`.

GuTS805:
Three unrelated failure modes (input already consumed, malformed Arrow FFI schema, non-UInt8 payload type) collapsed into an indistinguishable None with zero diagnostics. The equivalent C node-API function (read_dora_input_data) already logs the failing Arrow type via tracing::error! - this brings the operator-side twin to parity.

A C/C++ operator receiving a routine non-byte input from an upstream Rust/Python node (not a corner case - every other FFI accessor in the C/C++ surface already handles and logs this, tracked as #2030) got a silent empty payload with no way to tell apart 'input was empty', 'already read this input', 'type mismatch', or 'malformed FFI schema'.

Documented as a known-but-unfiled Medium finding in the internal 2026-06-04 soundness audit; the sibling leak in examples/c-dataflow/operator.c was already fixed separately, but this diagnostics gap was not.

Rebased onto current upstream main (was accidentally branched off a 206-commits-stale fork main) and adapted to the DoraArray/internal::from_array_data rename from #3213.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2584,6 +2584,7 @@ dependencies = [
  "arrow 59.0.0",
  "dora-arrow-convert",
  "safer-ffi",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2584,7 +2584,6 @@ dependencies = [
  "arrow 59.0.0",
  "dora-arrow-convert",
  "safer-ffi",
- "tracing",
 ]
 
 [[package]]
--- a/apis/rust/operator/types/Cargo.toml
+++ b/apis/rust/operator/types/Cargo.toml
@@ -13,6 +13,7 @@ repository.workspace = true
 [dependencies]
 arrow = { workspace = true, features = ["ffi"] }
 dora-arrow-convert = { workspace = true }
+tracing = { workspace = true }
 
 [dependencies.safer-ffi]
 version = "0.1.13"
--- a/apis/rust/operator/types/Cargo.toml
+++ b/apis/rust/operator/types/Cargo.toml
@@ -13,7 +13,6 @@ repository.workspace = true
 [dependencies]
 arrow = { workspace = true, features = ["ffi"] }
 dora-arrow-convert = { workspace = true }
-tracing = { workspace = true }
 
 [dependencies.safer-ffi]
 version = "0.1.13"
--- a/apis/rust/operator/types/src/lib.rs
+++ b/apis/rust/operator/types/src/lib.rs
@@ -161,14 +161,23 @@ pub fn dora_free_input_id(_input_id: char_p_boxed) {}
 
 #[ffi_export]
 pub fn dora_read_data(input: &mut Input) -> Option<safer_ffi::Vec<u8>> {
+    // `eprintln!`, not `tracing::error!`: this function is compiled into the
+    // user's operator `.so`, which statically links its own copy of the
+    // `tracing` crate — a separate `GLOBAL_DISPATCH` from the one the dora
+    // binary initializes in its own process image. Without dynamic symbol
+    // export (which dora's build does not set up), events raised here would
+    // hit an uninitialized dispatcher and be silently dropped by tracing's
+    // no-subscriber default, defeating the point of this diagnostic.
+    // `eprintln!` writes straight to the process's stderr fd, unaffected by
+    // which copy of `tracing` compiled it in.
     let Some(data_array) = input.data_array.take() else {
-        tracing::error!("dora_read_data: input data already taken (double read?)");
+        eprintln!("dora_read_data: input data already taken (double read?)");
         return None;
     };
     let data = match unsafe { arrow::ffi::from_ffi(data_array, &input.schema) } {
         Ok(data) => data,
         Err(err) => {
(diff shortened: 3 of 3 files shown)


phil-opp:
Reviewed the diff. The refactor from the `?`/`.ok()?` chain into explicit `let-else`/`match` branches is behavior-equivalent — every failure path still returns `None`, so there's no change for existing C/C++ operators; the only addition is the `tracing::error!` diagnostics (and the `tracing` dependency, which is declared in both `Cargo.toml` and `Cargo.lock`). The `.unwrap()`s are confined to the test helper.

The two added tests assert on real behavior (`Int32` input → `None` via the `TryFrom` branch; double-read → first `Some`, second `None`). The Arrow-FFI-import-failure branch isn't directly covered, but that's a minor gap rather than a problem.

No issues found.

---

---

phil-opp:
Issue found by claude:

The new tracing::error! calls are statically linked into the user's operator .so, which carries its own GLOBAL_DISPATCH; the only set_global_default runs in the dora binary's separate copy, and Rust binaries don't --export-dynamic. So the logs likely reach no subscriber in the exact C/C++ deployment they target. eprintln! would work unconditionally. Worth one empirical examples/c-dataflow run with a non-UInt8 input to confirm before merging.

GuTS805:
@phil-opp Good catch — switched to `eprintln!`, which writes directly to the process's
stderr fd and isn't affected by the `.so`'s separate `tracing::GLOBAL_DISPATCH`
copy. Also dropped the now-unused `tracing` dependency.

I wasn't able to do the empirical `examples/c-dataflow` run with a non-UInt8
input in this environment (no C/C++ toolchain readily available here) — worth
a maintainer sanity-check before merge, but the structural fix (stderr instead
of a dispatcher-gated macro) should hold regardless of that verification.

phil-opp:
**🤖 Automated review by Claude — this review is fully automated, with no human in the loop.**

Re-reviewed after the switch from `tracing::error!` to `eprintln!` in `f90cf44`. This addresses the dispatcher concern raised in the earlier review: `eprintln!` writes directly to the process stderr fd and is unaffected by the operator `.so` linking its own `tracing` `GLOBAL_DISPATCH`, so the diagnostics will actually surface in the C/C++ deployment they target. The now-unused `tracing` dependency was correctly removed, leaving a net-zero dependency change.

The refactor remains behavior-equivalent — all failure paths still return `None`. Both added tests exercise real paths (the `Int32` case genuinely hits the `TryFrom` type-mismatch branch, since the `&[u8]` conversion is gated on `UInt8Type`; the double-read test covers the `take()` guard). No new issues found in the latest revision.

The one open item is verification rather than a code defect: an empirical `examples/c-dataflow` run with a non-`UInt8` input to confirm the messages surface as intended — worth a maintainer sanity-check.

---

PR #3249, opened 2026-08-20
refactor(python)!: move the CUDA helpers out of `dora` into the `dora_tensor_pool` extension

phil-opp:
`dora/cuda.py` was about to be frozen into the 1.0 Python API, but nothing in dora imports it. Its only consumers are `examples/cuda-benchmark/` and the tensor-pool extension's own examples, and neither runs in CI — both are on the skip list for want of a CUDA toolkit (`scripts/smoke-all.sh:630`, `tests/example-smoke.rs`). It is about ten weeks old, has been reworked twice in the last week (#3079, #3152), and carries an open defect against the guard #3079 added (#3208). Removing a documented symbol after 1.0 needs a major bump; adding one back in 1.1 is additive — so it moves now, behind the same seam #3152 put the rest of the tensor pool behind.

The helpers become a real package rather than a third parked file: `libraries/extensions/tensor-pool/python/dora_tensor_pool`, installable from that directory and importable as `dora_tensor_pool`. `tensor_info_helpers.py` — parked out of `dora/cuda.py` by #3152 and never given an import path, which is exactly why the extension README and its own examples still read `from dora.cuda import` — folds in as `tensor_info.py`.

`get_tensor_info` and `tensor_from_info` existed in both files, identical apart from one docstring word (`read_memory_pool` vs `read_tensor_pool`). The extension's copy wins and the `dora/cuda.py` duplicate is dropped. Both carried the #3208 negative-dimension bypass, so that defect now sits entirely outside the 1.0 guarantees rather than half in and half out.

Updated consumers: four files in `examples/cuda-benchmark/`, both tensor-pool examples, the extension README, `docs/api-python.md` and `guide/src/languages/python.md`. The two doc pages now state plainly that the module is not part of the 1.0 API and where to install it from.

`cargo check -p dora-node-api-python` passes — the Rust binding never referenced the module. The package's three files parse, and every name `__init__` re-exports is defined. Not runnable end-to-end from here: the examples need a CUDA toolkit, same as before the move.

### DIFF
--- a/docs/api-python.md
+++ b/docs/api-python.md
@@ -743,10 +743,12 @@ Return the dictionary representation for YAML serialization.
 
 ---
 
-## CUDA Module
+## CUDA Module (opt-in extension)
+
+> **Not part of the 1.0 API.** These helpers moved out of the `dora` package into the `dora_tensor_pool` extension before 1.0 — they exist to serve the tensor-pool transport, which sits behind the extension seam, and are not covered by dora's 1.0 compatibility guarantees. Install from `libraries/extensions/tensor-pool/python`.
 
 ```python
-from dora.cuda import torch_to_ipc_buffer, ipc_buffer_to_ipc_handle, open_ipc_handle
+from dora_tensor_pool import torch_to_ipc_buffer, ipc_buffer_to_ipc_handle, open_ipc_handle
 ```
 
 Utilities for zero-copy GPU tensor sharing between nodes via CUDA IPC. Requires PyTorch with CUDA and Numba with CUDA support.
@@ -759,7 +761,7 @@ Convert a PyTorch CUDA tensor into an Arrow array containing the CUDA IPC handle
 import torch
 import pyarrow as pa
 from dora import Node
-from dora.cuda import torch_to_ipc_buffer
+from dora_tensor_pool import torch_to_ipc_buffer
 
 node = Node()
 tensor = torch.randn(1024, 768, device="cuda")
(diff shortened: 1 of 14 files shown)


phil-opp:

_This is a fully-automated review posted by Claude Code on a schedule; a human hasn't vetted it. Skip anything that isn't useful._

`libraries/extensions/tensor-pool/python/dora_tensor_pool/tensor_info.py` has no imports — it references `torch`, `np`, `ctypes`, and `_CudaArrayInterface` at module level and inside `tensor_from_info`, but the file starts straight with the docstring and then `_DTYPE_MAP = { "<i8": torch.int64, ... }`. On `main`, `tensor_info_helpers.py` had the same latent defect but was orphaned (no import path, README/examples all read `from dora.cuda import`), so nothing ever loaded it. This PR wires it into `dora_tensor_pool/__init__.py`:

```python
from .tensor_info import get_tensor_info, tensor_from_info
```

so `import dora_tensor_pool` now executes `tensor_info.py` and dies at the `_DTYPE_MAP` dict with `NameError: name 'torch' is not defined` — every consumer the PR updates (`examples/cuda-benchmark/{demo_,}{sender,receiver}.py`, `libraries/extensions/tensor-pool/examples/{sender,receiver}.py`, README, `docs/api-python.md`, `guide/src/languages/python.md`) fails at import. Fixing the top-level `_DTYPE_MAP` / `_TORCH_TO_NUMPY_DTYPE_MAP` NameError is enough to make `import dora_tensor_pool` succeed, but calling `tensor_from_info` then hits the same problem on `np`, `ctypes`, and `_CudaArrayInterface` (which the file's own docstring notes lives in `cuda.py` — needs `from .cuda import _CudaArrayInterface`).

The PR description says "the package's three files parse, and every name `__init__` re-exports is defined." Parsing is fine — the failure is at import execution, not at parse time — but the second half doesn't hold: `torch`, `np`, `ctypes`, `_CudaArrayInterface` are all undefined at the point `_DTYPE_MAP` runs.

CI doesn't catch this because the CUDA examples are on the skip list, but a smoke check of just `python -c "import dora_tensor_pool"` from `libraries/extensions/tensor-pool/python` would.

Not commenting on PR #3253 — it's a one-line CLAUDE.md doc correction and I confirmed the claim (`xtask/Cargo.toml` pins `version = "0.1.0"`; `apis/rust/operator/types/Cargo.toml` uses `version.workspace = true`).


---

PR #3254, opened 2026-08-21
fix(ros2-bridge): reject null string elements in array/sequence serialization

phil-opp:
## Summary

The scalar-string serialization path in `libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs` deliberately **rejects** a null value via `check_not_null`, with a documented rationale:

> ROS2 messages (and the CDR wire format) have no null concept… Serializing that would invent a value the producer never sent, so reject it instead.

The **array** (`string[N]`) and **sequence** (`string[]`) string paths violated that same invariant: they encoded a null element as `""` via `s.unwrap_or_default()`.

## Issue

A producer emitting an Arrow `List<Utf8>` (or `LargeList`) for a ROS2 string-array field with a **null element** — e.g. a `pyarrow` node that leaves a slot unset — had that null silently turned into an empty string on the CDR wire, rather than surfacing an error. This is inconsistent with the module's own null-rejection policy for scalar strings, and invents data the producer never sent (an empty string is a *value*, not "absent", on the wire).

Four functions were affected:
- `serialize_arrow_string` / `serialize_arrow_wstring` in `serialize/array.rs` (fixed `T[N]`)
- `serialize_arrow_string` / `serialize_arrow_wstring` in `serialize/sequence.rs` (variable `T[]`)

## Fix

Route all four element loops through a shared `reject_null_string_element` helper added next to the existing `check_not_null`. It mirrors the scalar path — erroring on a null element (naming the offending index) instead of substituting a default — so scalar, array, and sequence string paths now enforce the identical invariant. No change to the all-present (happy) path: the error `format!` is lazy (`ok_or_else`), so it allocates only when a null is actually present (which aborts the message anyway).

## Validation

- Added regression tests for a null element in `string[N]`, `wstring[N]`, and `string[]` (each asserts the null case errors and the all-present case still serializes). The all-present array test helper was consolidated to delegate to the nullable one (no duplicate scaffolding).
- `cargo test -p dora-ros2-bridge-arrow` → 24 passed, 0 failed.
- `cargo fmt -p dora-ros2-bridge-arrow -- --check` clean; `cargo clippy -p dora-ros2-bridge-arrow --tests -- -D warnings` clean.
- Rebased on current `origin/main`; full-workspace fmt + clippy pass, and the full test suite passes except pre-existing zenoh/rmw networking tests that require IPv6 (unavailable in this CI sandbox) — unrelated to this change.

---

⚠️ **This is a machine-generated change**, authored by Claude (an AI agent) as part of an automated code-review pass and reviewed for correctness. Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_014Yn5Z4p2j3uCKvBRaRh241)_

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs
@@ -12,7 +12,7 @@ use serde::ser::SerializeTuple;
 
 use crate::TypeInfo;
 
-use super::{TypedValue, check_array_len, error};
+use super::{TypedValue, check_array_len, error, reject_null_string_element};
 
 /// Serialize an array with known size as tuple.
 pub struct ArraySerializeWrapper<'a> {
@@ -323,8 +323,8 @@ where
 {
     check_array_len(array.len(), array_len)?;
     let mut seq = serializer.serialize_tuple(array_len)?;
-    for s in array.iter() {
-        seq.serialize_element(s.unwrap_or_default())?;
+    for (i, s) in array.iter().enumerate() {
+        seq.serialize_element(reject_null_string_element(s, i)?)?;
     }
     seq.end()
 }
@@ -340,8 +340,8 @@ where
 {
     check_array_len(array.len(), array_len)?;
     let mut seq = serializer.serialize_tuple(array_len)?;
-    for s in array.iter() {
-        let utf16: Vec<u16> = s.unwrap_or_default().encode_utf16().collect();
+    for (i, s) in array.iter().enumerate() {
+        let utf16: Vec<u16> = reject_null_string_element(s, i)?.encode_utf16().collect();
         seq.serialize_element(&utf16)?;
     }
     seq.end()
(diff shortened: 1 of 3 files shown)


phil-opp:

This PR is one of three opened by an automated in-depth review pass over `origin/main`. Each finding was independently verified against the current tree; each PR is scoped to one crate and validated with fmt + clippy (`-D warnings`) + the affected crate's tests, plus one full-workspace suite + `/review` + `/simplify` gate run over the combined diff. All three are **draft** and **machine-generated** (authored by Claude, an AI agent) — please review carefully.

### PRs opened by category

**Correctness**
- #3254 — `fix(ros2-bridge)`: reject null string elements in `string[]` / `string[N]` serialization (this PR). The array/sequence string paths encoded a null as `""`, violating the null-rejection invariant the scalar path already documents and enforces.

**Documentation**
- #3255 — `docs(node-api)`: compile-checked doctests for `DoraNode::send_output` and `EventStream::recv` — the two primary send/receive verbs had no runnable example while the lower-level `send_output_raw` did.

**Simplification**
- #3256 — `refactor(node-api)`: dedup five near-identical control-event JSON arms in `record_event` into one `control_event_json` helper (byte-for-byte identical output under `preserve_order`).

- **Coordinator per-IP connection rate limit throttles authenticated daemon reconnects** (`binaries/coordinator/src/ws_server.rs`) — real availability concern under a reconnect storm / shared-NAT deployment, but the correct fix is a judgment call (exempt the authenticated daemon endpoint vs. key on `DaemonId` vs. count only failed-auth) and the change is in DoS-protection code. Left for a maintainer decision rather than a speculative PR.
- **Dead `pub type SharedMemoryId = String`** (`libraries/message/src/common.rs`) — genuinely unused, but removing a `pub` alias is a semver-breaking change on a published crate right before the 1.0 stable cut; not worth the churn for a transparent alias.
- **`broadcast_all_nodes_ready` not guarded against a repeat trigger** (`binaries/coordinator/src/lib.rs`) — likely unreachable given the daemon-side latch; a defensive one-liner, low confidence.
- **`/api/artifacts` downloads share the same per-IP cap** — latent only; every current caller passes `artifact_base_url: None`, so the HTTP artifact path isn't exercised.
- **`mangle`/`demangle` asymmetry on a literal `%`** (`ros2-bridge/.../keyexpr.rs`) — matches the rmw_zenoh convention; ROS2 names don't contain `%`. By design.
- **`DataMessage` single-variant enum** — collapsing to a newtype would touch many call sites and forfeits the reserved variant tag for wire forward-compat; needs a maintainer call.
- **Log-level string mapping duplicated across CLI printers** — cold path, low value.

---

phil-opp:

No issues found.

The fix routes all four array/sequence string-element paths (`string`/`wstring` × `array.rs`/`sequence.rs`) through `reject_null_string_element`, and the routing match arms confirm `BoundedString`/`BoundedWString` and both the Utf8 (i32) and LargeUtf8 (i64) offset types funnel through those same generic functions — so there is no remaining bypass. The error is propagated as a `serde::ser::Error` (no panic, no swallow), and the non-null path is byte-identical to before. The added tests build genuine null slots (`StringArray::from(vec[Some("a"), None, Some("c")])`) and assert both the error case and the still-serializes case.

---

PR #3256, opened 2026-08-21
refactor(node-api): dedup control-event JSON in record_event

phil-opp:
## Summary

Five arms of `EventStream::record_event` (`apis/rust/node/src/event_stream/mod.rs`) — `Stop`, `InputClosed`, `InputRecovered`, `NodeRestarted`, `AllInputsClosed` — each recomputed the node-relative time offset and hand-built a JSON object that differed only in the `"type"` string and whether an `"id"` field was present.

## Issue

~45 lines of near-identical copy-paste, with the event schema (the `time_offset_secs` computation and field layout) spread across five places to keep in sync.

## Fix

Extract a `control_event_json(clock, start_timestamp, ty, id)` free function that centralizes the timestamp computation and field layout; the five arms collapse to one-liners.

- It is a **free function** (not a `&self` method) by necessity: `record_event` holds a mutable borrow of `self.write_events_to` across the `match`, so a `&self` method would conflict. Passing the disjoint `clock` and `start_timestamp` fields by reference is the standard disjoint-borrow idiom, and preserves the existing laziness (JSON is only built when recording is active).
- The object is built via `serde_json::Map` (rather than the `json!` macro) so the key order — `type`, optional `id`, `time_offset_secs` — stays **byte-for-byte identical** to the previous per-arm literals under serde_json's `preserve_order` feature (which this crate enables). No behavior change; this is a cold, recording-only path.

## Validation

- Added a unit test (`control_event_json_shape_and_key_order`) asserting the exact key set **and order** for both the id-bearing and id-less cases.
- `cargo test -p dora-node-api` → 182 passed, 0 failed.
- `cargo fmt` clean; `cargo clippy -p dora-node-api -- -D warnings` clean.

---

⚠️ **This is a machine-generated change**, authored by Claude (an AI agent) as part of an automated code-review pass and reviewed for correctness. Please review carefully before merging.

---
_Generated by [Claude Code](https://claude.ai/code/session_014Yn5Z4p2j3uCKvBRaRh241)_

### DIFF
--- a/apis/rust/node/src/event_stream/mod.rs
+++ b/apis/rust/node/src/event_stream/mod.rs
@@ -1004,17 +1004,12 @@ impl EventStream {
         if let Some(write_events_to) = &mut self.write_events_to {
             let event_json = match event {
                 EventItem::NodeEvent { event, .. } => match event {
-                    NodeEvent::Stop => {
-                        let time_offset = self
-                            .clock
-                            .new_timestamp()
-                            .get_diff_duration(&self.start_timestamp);
-                        let event_json = serde_json::json!({
-                            "type": "Stop",
-                            "time_offset_secs": time_offset.as_secs_f64(),
-                        });
-                        Some(event_json)
-                    }
+                    NodeEvent::Stop => Some(control_event_json(
+                        &self.clock,
+                        &self.start_timestamp,
+                        "Stop",
+                        None,
+                    )),
                     NodeEvent::Reload { .. } => None,
                     NodeEvent::Input { id, metadata, data } => {
                         let mut event_json = convert_output_to_json(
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

This refactor is behavior-preserving. `control_event_json` reproduces the five former per-arm `json!` objects exactly: value types (`Value::String` for `type`/`id`, `Number` for `time_offset_secs`) and the `type` → optional `id` → `time_offset_secs` key order all match the originals, and `serde_json`'s `preserve_order` feature is enabled for this crate so that ordering is real rather than incidental. No event fields are dropped, and the set of recorded vs. skipped events (`Reload`/`_ => None`, the untouched `Input` arm) is unchanged. The added `control_event_json_shape_and_key_order` test genuinely pins the key set and order for both the id-bearing and id-less cases.

---

PR #3261, opened 2026-08-21
fix(download): reject Windows reserved names and trailing dots in `sanitize_filename`

### LINKED ISSUE #3251: download_file: sanitize_filename accepts Windows reserved device names and trailing dots
Found during review of #3224 (which is fine — this is a pre-existing gap on `main`, not something that PR introduces).

`sanitize_filename` in `libraries/extensions/download/src/lib.rs` strips path components, rejects NUL bytes, and caps the length at 255 — but nothing else:

```rust
fn sanitize_filename(name: &str) -> Option<String> {
    let sanitized = Path::new(name).file_name().and_then(|n| n.to_str()).map(|s| s.to_string())?;
    if sanitized.contains('\0') || sanitized.len() > 255 {
        return None;
    }
    Some(sanitized)
}
```

`Content-Disposition` is attacker-influenced, so a hostile server can pick the resulting on-disk name within that charset.

## Two gaps

**Windows reserved device names.** `Content-Disposition: attachment; filename="NUL"` (or `CON`, `AUX`, `COM1`, `LPT1`, and the `NUL.bin` form) passes. On Windows `tokio::fs::File::create(target_dir/NUL)` opens the *null device*: the write and `sync_all` both "succeed", `download_file` returns a path to a device rather than a file, and the caller — `binaries/daemon/src/spawn/command.rs:87` — goes on to spawn or `dlopen` something with no content. Not a traversal, but a silent-corruption path.

**Trailing dots and spaces.** `filename="evil."` passes; Windows silently strips the trailing dot on create, so the file lands at a different path than the one returned.

## Not affected

Traversal is genuinely handled: `Path::file_name` kills `..`, `.`, `/`, and `a/..`. There is no percent-decoding anywhere in the crate — `url::Url::path_segments()` yields percent-*encoded* segments — so `%2f`, `%00` and `%2e%2e` never decode into separators, NUL, or `..`. The decode-then-sanitize ordering hazard does not apply here.

## Suggested fix

Reject the reserved stems case-insensitively (with or without an extension) and trim trailing dots/spaces, then re-check for empty. Both branches of `resolve_filename` already share `sanitize_filename`, so one change covers the header and the URL fallback.

Worth adding tests for the adversarial inputs the doc comment implies are covered — `"../../etc/passwd"`, `"/etc/passwd"`, an embedded NUL, a 256-byte name — none of which are exercised today.

phil-opp:
## Summary

Fixes #3251.

The download filename can come from an attacker-influenced `Content-Disposition` header (`libraries/extensions/download/src/lib.rs`). `sanitize_filename` stripped path components, NUL bytes, and over-long names, but nothing else, so a hostile server could still pick a name that behaves unexpectedly on Windows:

- **Reserved device names.** `Content-Disposition: attachment; filename="NUL"` (or `CON`, `AUX`, `PRN`, `COM1`, `LPT1`, and the `NUL.bin` form) passed. On Windows `tokio::fs::File::create(target_dir/NUL)` opens the **null device**: the write and `sync_all` both "succeed", `download_file` returns a path to a device rather than a file, and the caller (`binaries/daemon/src/spawn/command.rs`) goes on to spawn or `dlopen` something with no content — a silent-corruption path.
- **Trailing dots and spaces.** `filename="evil."` passed; Windows silently strips the trailing dot on create, so the file lands at a different path than the one returned.

## Fix

In `sanitize_filename`:
- Trim trailing dots/spaces and reject a name that trims away to nothing.
- Reject Windows reserved device names (`CON`, `PRN`, `AUX`, `NUL`, `COM0`–`COM9`, `LPT0`–`LPT9`) case-insensitively, ignoring any extension and a trailing space in the stem (`NUL`, `NUL.bin`, `NUL.tar.gz`, `NUL .txt` are all rejected).

Both hardening steps run on **every** platform, not gated to Windows, so the returned name always matches the file that is actually created and the checks are exercised by the Linux PR gate. Rejecting a download literally named `NUL`/`CON`/… is acceptable — such names are exceedingly rare as real artifact filenames. Both the header and URL-fallback paths already share `sanitize_filename`, so one change covers both.

## Tests

Added tests (all run on Linux) for:
- Reserved-name rejection across case, extension, and trailing-space variants, plus non-reserved look-alikes (`NULls`, `console.log`, `COM10`, `LPT`, `com.bin`) that must stay valid.
- Trailing dot/space trimming, including names that trim to empty.
- The traversal / embedded-NUL / 255-vs-256-byte cases the doc comment implies were covered but weren't exercised.

## Validation

- `cargo test -p dora-download` — 24 passed.
- `cargo clippy -p dora-download --all-targets` — clean.
- `cargo fmt -p dora-download -- --check` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V7q1v9CUH5yvDwL1gWMbwG)_

### DIFF
--- a/libraries/extensions/download/src/lib.rs
+++ b/libraries/extensions/download/src/lib.rs
@@ -127,10 +127,20 @@ fn filename_from_url(url: &reqwest::Url) -> Option<String> {
 }
 
 /// Sanitize a candidate filename: strip path components to prevent traversal,
-/// reject null bytes and overly long names. Returns `None` when nothing usable
+/// reject null bytes and overly long names, trim trailing dots/spaces, and
+/// reject Windows reserved device names. Returns `None` when nothing usable
 /// survives — e.g. `".."` or `"."` (which `Path::file_name` maps to `None`),
-/// or a name over 255 bytes / containing a NUL. (A trailing slash such as
+/// a name over 255 bytes / containing a NUL, a name that is only dots and
+/// spaces, or a reserved device name such as `NUL`. (A trailing slash such as
 /// `"dir/"` keeps its last component: `Path::file_name` returns `"dir"`.)
+///
+/// The `Content-Disposition` header is attacker-influenced, so the returned
+/// name is a name a hostile server could pick. The two hardening steps below
+/// are applied on **every** platform — not gated to Windows — both so the
+/// returned name always matches the file that is actually created and so the
+/// checks are exercised by the Linux PR gate. Rejecting a download literally
+/// named `NUL`/`CON`/… is acceptable: such names are exceedingly rare as real
+/// artifact filenames.
 fn sanitize_filename(name: &str) -> Option<String> {
     let sanitized = Path::new(name)
         .file_name()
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Traced the sanitization logic and tests in `libraries/extensions/download/src/lib.rs`: trailing dot/space trimming runs before the reserved-name check, `is_windows_reserved_name` strips the extension via `split('.').next()` and trims a trailing space in the stem (so `NUL.bin`, `NUL .txt`, `CON `, `nul ` are all rejected), and the COM/LPT check requires exactly one trailing ASCII digit, so `COM10`/`LPT` correctly stay valid. Case-insensitivity, the 255-byte boundary, and the traversal/NUL cases are all covered by the added tests, and `resolve_filename` still falls back to the URL-derived name when a hostile `Content-Disposition` header sanitizes away — no regression.

---

PR #3262, opened 2026-08-21
fix(cli): tolerate a partial multibyte tail in the initial follow read

### LINKED ISSUE #3257: `dora logs --local --follow` aborts on invalid UTF-8 during the initial read, unlike every subsequent poll
> **Note:** This issue was filed automatically by a scheduled/automated Claude code-review check. It reports a single, verified correctness/robustness inconsistency. Please treat the analysis as a starting point and verify before acting.

## Summary

In the local `dora logs --follow` path, the **initial** read of each log file uses strict UTF-8 decoding (`std::fs::read_to_string`), while every **subsequent** poll of the same file deliberately uses lossy decoding (`String::from_utf8_lossy`) so that catching the daemon mid-write of a multi-byte UTF-8 character cannot abort the follow session. Because the two readers disagree, a follow session can fail to start with an `invalid utf-8` error in exactly the race the poll loop is written to survive.

## Location

`binaries/cli/src/command/logs.rs`

**Initial read — strict (`read_to_string`), aborts on invalid UTF-8:**

```rust
// binaries/cli/src/command/logs.rs:825-833
fn read_log_file_tracked(path: &Path) -> Result<(Vec<LogMessage>, FollowedFile)> {
    let content = std::fs::read_to_string(path)   // <-- errors on invalid UTF-8
        .wrap_err_with(|| format!("failed to read {}", path.display()))?;
    let bytes = content.as_bytes();
    let head = bytes[..bytes.len().min(LOG_HEAD_FINGERPRINT_LEN)].to_vec();
    let consumed = match bytes.iter().rposition(|&b| b == b'\n') {
        Some(idx) => idx + 1,
        None => 0,
    };
    // ...
```

**Poll-loop read — lossy by design, with an explicit comment stating why:**

```rust
// binaries/cli/src/command/logs.rs:667-679
    // ... Bytes are decoded lossily so a read that ends inside a multibyte
    // UTF-8 sequence cannot abort the follow session.
fn read_appended_log_lines(file: &mut std::fs::File, pos: u64) -> Result<(Vec<LogMessage>, u64)> {
    file.seek(std::io::SeekFrom::Start(pos))?;
    let mut buf = Vec::new();
    file.read_to_end(&mut buf)?;
    let consumed = match buf.iter().rposition(|&b| b == b'\n') {
        Some(idx) => idx + 1,
        None => return Ok((Vec::new(), pos)),
    };
    let text = String::from_utf8_lossy(&buf[..consumed]);   // <-- lossy: never aborts
    let messages = text.lines().filter_map(parse_jsonl_line).collect();
    Ok((messages, pos + consumed as u64))
}
```

`read_log_file_tracked` is called on the initial pass of `follow_local_logs`:

```rust
// binaries/cli/src/command/logs.rs:304-307
    for path in &files {
        let (messages, state) = read_log_file_tracked(path)?;   // <-- `?` aborts the whole session
        all_messages.extend(messages);
        follow_state.insert(path.clone(), state);
    }
```

## Trigger / failure scenario

1. A node emits a log message containing a non-ASCII / multi-byte UTF-8 character (an emoji, accented text, CJK, etc.) into its `log_*.jsonl` (or `.txt`) file.
2. The user runs `dora logs <dataflow> <node> --follow` locally, and the initial read races the daemon so the file's tail ends *inside* that multi-byte sequence (a partial character at EOF).
3. `read_to_string` validates the **entire** file buffer — including the trailing partial bytes past the last newline — and returns `Err` (`invalid utf-8 sequence...`).
4. The `?` at line 305 propagates the error and `dora logs --follow` fails to start, rather than tolerating the partial tail the way `read_appended_log_lines` does on every later poll.

It does not panic or produce wrong output; it just fails the command in precisely the mid-write race the poll loop is explicitly written to survive. Severity is low (requires a timing race plus non-ASCII log content), but the inconsistency is real.

## Suggested fix

Make the initial read decode lossily too, mirroring `read_appended_log_lines`. Read raw bytes and derive `head`, `consumed`, and the stored byte `offset` from the raw buffer (so the offset stays a correct byte offset into the file), then decode only the consumed prefix with `String::from_utf8_lossy` for parsing:

```rust
fn read_log_file_tracked(path: &Path) -> Result<(Vec<LogMessage>, FollowedFile)> {
    let bytes = std::fs::read(path)
        .wrap_err_with(|| format!("failed to read {}", path.display()))?;
    let head = bytes[..bytes.len().min(LOG_HEAD_FINGERPRINT_LEN)].to_vec();
    let consumed = match bytes.iter().rposition(|&b| b == b'\n') {
        Some(idx) => idx + 1,
        None => 0,
    };
    let content = String::from_utf8_lossy(&bytes[..consumed]);
    let messages = parse_log_content(path, &content)?;
    Ok((messages, FollowedFile { offset: consumed as u64, head, missed_polls: 0 }))
}
```

Deriving `consumed`/`offset`/`head` from the **raw** bytes (not from a lossy `String`, whose U+FFFD replacement chars would shift byte offsets and desync the seek offset used by later polls) keeps the follow-offset arithmetic correct while removing the abort. A regression test writing a file whose tail ends mid-multibyte-character and asserting `read_log_file_tracked` succeeds would lock this in.

phil-opp:
## Summary

Fixes #3257.

In the local `dora logs --follow` path (`binaries/cli/src/command/logs.rs`), the **initial** read of each log file used strict UTF-8 decoding (`read_to_string`), while every **subsequent** poll (`read_appended_log_lines`) deliberately decodes lossily so that catching the daemon mid-write of a multi-byte UTF-8 character cannot abort the follow session. Because the two readers disagreed, a follow session could fail to start with an `invalid utf-8` error in exactly the race the poll loop is written to survive.

### Failure scenario

1. A node emits a log line with a non-ASCII / multi-byte character (emoji, accented text, CJK) into `log_*.jsonl`.
2. `dora logs … --follow` runs locally and the initial read races the daemon so the file's tail ends *inside* that multi-byte sequence (a partial character at EOF).
3. `read_to_string` validates the **entire** buffer — including the partial trailing bytes past the last newline — and returns `Err`.
4. The `?` in `follow_local_logs` propagates the error and the command fails to start, rather than tolerating the partial tail the way `read_appended_log_lines` does on every later poll.

## Fix

Read raw bytes and decode only the newline-terminated prefix lossily, mirroring `read_appended_log_lines`. `head`, `consumed`, and the stored byte `offset` are derived from the **raw** buffer so the offset later polls seek to stays correct — deriving them from a lossy `String` would let its U+FFFD replacement chars shift byte offsets and desync the seek.

Only the complete-line prefix is parsed, so a line the daemon is part-way through writing is still left for a later poll (unchanged behavior).

## Tests

Added `read_log_file_tracked_tolerates_a_partial_multibyte_tail`: writes a file whose tail ends inside a 3-byte UTF-8 character, asserts `read_to_string` on the whole buffer fails (documenting the old abort) while `read_log_file_tracked` succeeds, returns the complete message, and reports the raw byte offset.

## Validation

- `cargo test -p dora-cli --lib read_log_file_tracked` — 2 passed.
- `cargo clippy -p dora-cli --lib` — clean.
- `cargo fmt -p dora-cli -- --check` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01V7q1v9CUH5yvDwL1gWMbwG)_

### DIFF
--- a/binaries/cli/src/command/logs.rs
+++ b/binaries/cli/src/command/logs.rs
@@ -823,16 +823,24 @@ fn read_log_file(path: &Path) -> Result<Vec<LogMessage>> {
 /// [`read_appended_log_lines`], so a line the daemon is part-way through
 /// writing is printed once — by a later poll — rather than twice or never.
 fn read_log_file_tracked(path: &Path) -> Result<(Vec<LogMessage>, FollowedFile)> {
-    let content = std::fs::read_to_string(path)
-        .wrap_err_with(|| format!("failed to read {}", path.display()))?;
-    let bytes = content.as_bytes();
+    // Read raw bytes and decode lossily, mirroring `read_appended_log_lines`:
+    // if this initial read races the daemon and the tail ends inside a
+    // multi-byte UTF-8 sequence, strict decoding (`read_to_string`) would abort
+    // the whole follow session — the very race the poll loop is written to
+    // survive. `head`, `consumed`, and the stored `offset` are derived from the
+    // raw buffer so the byte offset later polls seek to stays correct (a lossy
+    // `String`'s U+FFFD replacements would shift those offsets).
+    let bytes =
+        std::fs::read(path).wrap_err_with(|| format!("failed to read {}", path.display()))?;
     let head = bytes[..bytes.len().min(LOG_HEAD_FINGERPRINT_LEN)].to_vec();
     let consumed = match bytes.iter().rposition(|&b| b == b'\n') {
         Some(idx) => idx + 1,
         None => 0,
     };
-    // `consumed` sits just past a newline, so it is always a char boundary.
-    let messages = parse_log_content(path, &content[..consumed])?;
+    // Only the newline-terminated prefix is parsed; a partial trailing line is
+    // left for a later poll. Decoding lossily cannot abort here.
+    let content = String::from_utf8_lossy(&bytes[..consumed]);
+    let messages = parse_log_content(path, &content)?;
     Ok((
         messages,
         FollowedFile {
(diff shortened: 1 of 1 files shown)


phil-opp:

No issues found.

Verified the fix against the surrounding code: `read_log_file_tracked` now reads raw bytes and lossily decodes only the newline-terminated prefix, matching the existing `read_appended_log_lines` poll path. Since `\n` never occurs inside a multibyte UTF-8 sequence, `consumed` (from `rposition(b'\n')`) always lands on a char boundary, so the decoded prefix is fully valid UTF-8 (no `U+FFFD` substitutions) and the stored `offset`/`head` stay byte-accurate for subsequent polls. The regression test is meaningful — it truncates a file mid-multibyte-char, confirms the old strict `read_to_string` path fails, and confirms the new path returns the complete message with the correct raw byte offset.

---

PR #3264, opened 2026-08-21
fix(daemon): honor max_rotated_files: 0 in rotate_log_files

### LINKED ISSUE #3176: Daemon log rotation: `max_rotated_files: 0` still keeps one rotated file (off-by-one in `rotate_log_files`)
## Summary

`rotate_log_files` in `binaries/daemon/src/log.rs` does not honor a configured `max_rotated_files` of `0`. Setting `max_rotated_files: 0` (intent: "keep no rotated log files") still leaves one rotated file (`log_<node>.1.jsonl`) on disk, behaving identically to `max_rotated_files: 1`.

This is **low severity** — the disk effect is bounded (exactly one extra file, overwritten on every subsequent rotation, never unbounded) and there is no crash or data loss — but the observed behavior contradicts the configured value.

## Location

`binaries/daemon/src/log.rs:43-72`

```rust
pub fn rotate_log_files(
    working_dir: &Path,
    dataflow_id: &Uuid,
    node_id: &NodeId,
    max_files: u32,
) -> std::io::Result<()> {
    // Delete the oldest if it exists
    let oldest = log_path_rotated(working_dir, dataflow_id, node_id, max_files);
    if oldest.exists() {
        std::fs::remove_file(&oldest)?;
    }

    // Shift .N -> .N+1 (from max_files-1 down to 1)
    for i in (1..max_files).rev() {
        let from = log_path_rotated(working_dir, dataflow_id, node_id, i);
        let to = log_path_rotated(working_dir, dataflow_id, node_id, i + 1);
        if from.exists() {
            std::fs::rename(&from, &to)?;
        }
    }

    // Rename current -> .1
    let current = log_path(working_dir, dataflow_id, node_id);
    let first = log_path_rotated(working_dir, dataflow_id, node_id, 1);
    if current.exists() {
        std::fs::rename(&current, &first)?;   // <-- runs unconditionally, even when max_files == 0
    }

    Ok(())
}
```

## Why it happens

With `max_files == 0`:

1. `oldest` resolves to `log_<node>.0.jsonl`, which is never created (rotated indices start at `1`), so the delete is a no-op.
2. The shift loop `(1..0).rev()` is empty.
3. The final block still renames `current -> log_<node>.1.jsonl` — there is no guard for `max_files == 0`.

So a "keep 0 rotated files" configuration retains exactly one rotated file, i.e. it is indistinguishable from `max_rotated_files: 1`.

## Reachability

The value flows directly from user configuration and is only defaulted when unset:

`binaries/daemon/src/spawn/prepared.rs:895-899`
```rust
let max_rotated_files = self
    .node
    .max_rotated_files()
    .context("Could not resolve `max_rotated_files` configuration")?
    .unwrap_or(log::DEFAULT_MAX_ROTATED_FILES);
```

There is no lower-bound clamp, so a user setting `max_rotated_files: 0` reaches `rotate_log_files(..., 0)`.

## Suggested fix

Guard the final rename (and, for symmetry, treat `0` as "no rotated files kept"). For example:

```rust
    // Rename current -> .1 (unless the config asked for zero rotated files)
    if max_files >= 1 {
        let current = log_path(working_dir, dataflow_id, node_id);
        let first = log_path_rotated(working_dir, dataflow_id, node_id, 1);
        if current.exists() {
            std::fs::rename(&current, &first)?;
        }
    } else {
        // max_files == 0: keep no rotated files — drop the current one instead
        let current = log_path(working_dir, dataflow_id, node_id);
        if current.exists() {
            std::fs::remove_file(&current)?;
        }
    }
```

A regression test analogous to the existing `rotate_*` tests (e.g. `rotate_keeps_no_files_when_max_zero`) would pin the behavior.

---

*This issue was filed automatically by a scheduled, unattended Claude code-review check (no human was in the loop). It reflects a static reading of the code; please verify against intended semantics before acting. If `max_rotated_files: 0` is considered an unsupported/nonsensical configuration, feel free to close — clamping the config to a minimum of `1` at parse time would be an equally valid resolution.*

sanyyam0305:
`rotate_log_files` renamed the current log to `.1` unconditionally, so `max_rotated_files: 0` (intent: keep no rotated files) still left exactly one rotated file on disk -- indistinguishable from `max_rotated_files: 1`.

Guards the rename instead of clamping the config at parse time: clamping would mean a user who explicitly asks for zero retention still ends up with one file, which sidesteps the reported behavior rather than fixing it. With `max_files == 0` the current log is dropped instead of renamed.

Fixes #3176

### DIFF
--- a/binaries/daemon/src/log.rs
+++ b/binaries/daemon/src/log.rs
@@ -52,6 +52,21 @@ pub fn rotate_log_files(
         std::fs::remove_file(&oldest)?;
     }
 
+    // Prune anything above the current limit. Rotation only ever deletes index
+    // `max_files`, so lowering `max_rotated_files` between runs of the same
+    // dataflow id would otherwise strand every file above the new limit and
+    // break the documented `max_log_size * (1 + max_rotated_files)` bound.
+    // Indices are contiguous, so stopping at the first gap is sufficient.
+    let mut stale = max_files + 1;
+    loop {
+        let path = log_path_rotated(working_dir, dataflow_id, node_id, stale);
+        if !path.exists() {
+            break;
+        }
+        std::fs::remove_file(&path)?;
+        stale += 1;
+    }
+
     // Shift .N -> .N+1 (from max_files-1 down to 1)
     for i in (1..max_files).rev() {
         let from = log_path_rotated(working_dir, dataflow_id, node_id, i);
@@ -713,6 +728,62 @@ mod tests {
         assert!(!current.exists());
         let rotated = log_path_rotated(tmp.path(), &uuid, &node, 1);
         assert!(!rotated.exists());
+    }
+
+    #[test]
+    fn rotate_prunes_down_to_a_lowered_limit() {
+        let tmp = tempfile::tempdir().unwrap();
+        let uuid = Uuid::nil();
+        let node = NodeId::from("n".to_string());
+
+        let dataflow_dir = tmp.path().join("out").join(uuid.to_string());
+        std::fs::create_dir_all(&dataflow_dir).unwrap();
+
+        // A directory previously rotated under `max_rotated_files: 5`.
+        std::fs::write(log_path(tmp.path(), &uuid, &node), "current\n").unwrap();
+        for i in 1..=5 {
+            std::fs::write(log_path_rotated(tmp.path(), &uuid, &node, i), "old\n").unwrap();
+        }
+
+        rotate_log_files(tmp.path(), &uuid, &node, 2).unwrap();
+
+        // The documented bound is `max_log_size * (1 + max_rotated_files)`, so
+        // files above the new limit must not be stranded.
+        assert!(log_path_rotated(tmp.path(), &uuid, &node, 1).exists());
+        assert!(log_path_rotated(tmp.path(), &uuid, &node, 2).exists());
+        for i in 3..=5 {
+            assert!(
+                !log_path_rotated(tmp.path(), &uuid, &node, i).exists(),
+                ".{i} should have been pruned"
+            );
+        }
+    }
+
+    #[test]
+    fn rotate_at_zero_prunes_files_left_by_an_earlier_limit() {
+        let tmp = tempfile::tempdir().unwrap();
+        let uuid = Uuid::nil();
+        let node = NodeId::from("n".to_string());
+
+        let dataflow_dir = tmp.path().join("out").join(uuid.to_string());
+        std::fs::create_dir_all(&dataflow_dir).unwrap();
+
+        std::fs::write(log_path(tmp.path(), &uuid, &node), "current\n").unwrap();
+        for i in 1..=3 {
+            std::fs::write(log_path_rotated(tmp.path(), &uuid, &node, i), "old\n").unwrap();
+        }
+
+        rotate_log_files(tmp.path(), &uuid, &node, 0).unwrap();
+
+        // "keep no rotated files" has to mean none, including any left by an
+        // earlier non-zero configuration.
+        assert!(!log_path(tmp.path(), &uuid, &node).exists());
+        for i in 1..=3 {
+            assert!(
+                !log_path_rotated(tmp.path(), &uuid, &node, i).exists(),
+                ".{i} should have been pruned"
+            );
+        }
     }
 
     #[test]
--- a/binaries/daemon/src/log.rs
+++ b/binaries/daemon/src/log.rs
@@ -61,11 +61,17 @@ pub fn rotate_log_files(
         }
     }
 
-    // Rename current -> .1
+    // Rename current -> .1 (unless the config asked for zero rotated files)
     let current = log_path(working_dir, dataflow_id, node_id);
-    let first = log_path_rotated(working_dir, dataflow_id, node_id, 1);
-    if current.exists() {
-        std::fs::rename(&current, &first)?;
+    if max_files >= 1 {
+        let first = log_path_rotated(working_dir, dataflow_id, node_id, 1);
+        if current.exists() {
+            std::fs::rename(&current, &first)?;
+        }
+    } else if current.exists() {
+        // max_files == 0: keep no rotated files — drop the current one
+        // instead of renaming it to a .1 that would never get cleaned up.
+        std::fs::remove_file(&current)?;
     }
 
     Ok(())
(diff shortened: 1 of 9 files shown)


phil-opp:
The change to `rotate_log_files` is internally correct, but as written it doesn't change any user-observable behavior, so I don't think it actually fixes the reported issue.

At this head, `max_rotated_files: 0` is rejected *before* rotation ever runs:

- `libraries/core/src/descriptor/validate.rs:426-429` bails with `` `max_rotated_files` must be at least 1 `` whenever the resolved value is `0`.
- The daemon's rotation path resolves `max_files` through that same accessor: `binaries/daemon/src/spawn/prepared.rs` calls `self.node.max_rotated_files()?...`, and the `?` propagates the bail. So a `0` config errors out at spawn (and earlier during `check_dataflow` validation) and never reaches `rotate_log_files(..., 0)` — whose only production call site is in `prepared.rs`.

Net effect: a user who sets `max_rotated_files: 0` today gets a hard validation error, not "one leftover rotated file". The new `else { remove_file(...) }` branch is unreachable in production, and `rotate_keeps_no_files_when_max_zero` only exercises the helper directly, bypassing the validation that makes `0` impossible — so it green-lights a path production can't take.

To genuinely honor `max_rotated_files: 0`, this PR also needs to relax the lower bound in `validate.rs` (allow `0`, keep the `> 100` check) and update the schema/docs. Otherwise the `log.rs` change is inert, and the issue would instead be resolved by documenting that `0` is unsupported. Worth deciding which direction is intended before merging.

Note: the same limitation was raised on #3081, which attempts the same `log.rs`-only fix.

---
_This is an automated review by Claude Code (fully automated — no human in the loop). It may contain mistakes; please verify independently._

---

sanyyam0305:
Implemented direction (a). You were right that the log.rs change was inert - `validate.rs:427` rejected 0 before rotation ran, so the guard was unreachable and the test only passed by calling the helper directly.

`bd67527d` relaxes the lower bound so 0 means "keep the active log only", which is what the documented disk bound already implies (`max_log_size * (1 + max_rotated_files`) = one file at 0). Upper bound unchanged.

One thing surfaced along the way: the generated schema advertised `minimum: 0` (derived from `u32`) with no `maximum, so max_rotated_files: 500` passed schema validation and then failed at validate.rs. Added `#[schemars(range(max = 100))]` so the schema and validator agree in both directions, and regenerated both committed copies.

Docs updated across the 6 mirrored `docs/`+`guide/` sites. Verified on 1.97.1: fmt, clippy `-D warnings`, and 1,274 tests across the 5 affected crates.

Worth noting the bound has been in place since `145ccce0` (2026-04-17), four months before #3176 was filed - so the issue's "no lower-bound clamp" premise was already stale, and its suggested alternative (clamp at parse time) effectively already existed. If you'd rather keep 0 unsupported, say so and I'll drop `bd67527d` back to just closing this out. Same decision applies to #3081.

phil-opp:
**Automated review by Claude** (fully automated — no human in the loop).

The new commit `bd67527d` closes the reachability gap raised in the earlier review: `validate.rs` no longer rejects `0`, and since the daemon resolves `max_rotated_files` through that same validating accessor (`spawn/prepared.rs`), the `max_files == 0` branch in `rotate_log_files` is now reachable end-to-end. The generated schema (`maximum: 100`) and the doc pages are in sync, and the validator test now exercises a real descriptor path rather than calling the helper directly. No issues found in the delivered fix.

Two things worth a maintainer's eye:

- This is now a superset of #3081, which is `log.rs`-only and still carries the `n == 0` bail in `validate.rs`, so it targets the same issue (#3176) but can't actually honor `0`. #3081 looks superseded by this PR.
- Minor: at `max_rotated_files: 0` this deletes only the current log. Pre-existing `.1..N` files left behind by an earlier non-zero config in the same `out/<dataflow_id>/` directory would remain, so the documented "one file at 0" disk bound doesn't hold unconditionally in that transition. Dataflow IDs are normally per-run UUIDs so this is an edge case, but #3081's prune loop does handle it — worth folding in if this is the PR that lands.

---

sanyyam0305:
Folded in the prune fix from the second review (`71d82894`). `rotate_log_files` only ever deleted index `max_files`, so lowering `max_rotated_files` between runs of the same dataflow id stranded everything above the new limit - at 0 that meant `.1..N` from an earlier config survived. Now prunes above the limit before the shift; indices stay contiguous so it stops at the first gap, costing one extra stat per rotation.

Two new tests cover it: `rotate_prunes_down_to_a_lowered_limit (5 → 2)` and `rotate_at_zero_prunes_files_left_by_an_earlier_limit`. Full `dora-daemon` suite is 249 passing on 1.97.1.
On the superseding point - happy to defer if you'd rather land #3081 and I close this, or close #3081 if this one lands. Whichever you prefer; just flagging that both are open against #3176.

phil-opp:
🤖 **Fully automated review by Claude** — no human was in the loop; please verify before relying on it.

Re-reviewed after `71d82894` ("prune rotated logs above the configured limit"), the commit pushed since the last automated review. It folds in exactly the prune loop the earlier review asked for: `rotate_log_files` now removes rotated indices above `max_files` (starting at `max_files + 1`, stopping at the first gap since rotation keeps indices contiguous) *before* the shift. So lowering `max_rotated_files` between runs of the same dataflow id — `0` included — no longer strands `.1..N` left by an earlier, higher config, and the documented `max_log_size * (1 + max_rotated_files)` disk bound now holds unconditionally.

The two new tests (`rotate_prunes_down_to_a_lowered_limit`, `rotate_at_zero_prunes_files_left_by_an_earlier_limit`) create real files on disk and assert the above-limit ones are pruned while the in-limit ones survive, so they genuinely exercise the transition rather than just re-asserting an invariant. The `max_files + 1` cannot overflow (validation caps `max_rotated_files` at 100) and the loop is bounded by the files actually present.

No new issues found in the delivered fix — the earlier residual (files stranded by a lowered limit) is now resolved.

One note for maintainers: #3081 targets the same `max_rotated_files == 0` behavior but keeps the `validate.rs` floor at 1, which leaves its `rotate_log_files(.., 0)` branch unreachable in production. This PR is the more complete of the two (it relaxes the validation floor *and* prunes stranded files), so the two overlap and only one should be merged.

---

aryanraj45 (approved):

PR #3265, opened 2026-08-21
perf(daemon): cap coalescing copy in socket_stream_send

phil-opp:
> ⚠️ **Machine-generated PR.** This change was authored by Claude (Anthropic's Claude Code) during an automated codebase review. Please review carefully before merging.

## Issue

The daemon-side reply writer `socket_stream_send` (`binaries/daemon/src/socket_stream_utils.rs`) unconditionally copied the **entire** serialized message into a fresh `8 + message.len()` header+body buffer before writing:

```rust
let mut buf = Vec::with_capacity(8 + message.len());
buf.extend_from_slice(&len_raw);
buf.extend_from_slice(message);
connection.write_all(&buf).await?;
```

This is the daemon→node reply path (`node_communication/tcp.rs::send_reply`). A `DaemonReply::NextEvents` carries data-plane payloads up to `MAX_MESSAGE_BYTES` (**64 MiB**) whenever an output takes the daemon path instead of direct zenoh — so every such reply pays a full extra memcpy of the payload, on top of the copy `encode_presized` already made.

The **node-side** transport (`apis/rust/node/src/daemon_connection/tcp.rs`) already solved exactly this: it coalesces header+body into one buffer only for small payloads (`<= COALESCE_THRESHOLD` = 16 KiB, to avoid a runt TCP segment under `TCP_NODELAY`) and writes header-then-body separately above that threshold, with the comment *"copying tens of MiB to avoid one header segment is the wrong trade."* The daemon side never got that treatment.

## Fix

Mirror the node-side threshold on the daemon side: coalesce header+body into one write only for payloads `<= 16 KiB`; otherwise write the 8-byte length header and the body as two `write_all` calls with no large copy.

The wire format (8-byte little-endian length prefix + body) is **byte-identical** on both paths, so there is no behavior change — only the elimination of one full memcpy of every large daemon-path payload.

## Validation

- Added `large_payload_above_coalesce_threshold_roundtrips`, an above-threshold round-trip test proving the two-write path frames byte-identically to the coalesced path.
- `cargo test -p dora-daemon` (socket_stream + full lib/tests) — all pass (247 lib tests green)
- Reviewed with `/code-review` and `/simplify`: no correctness issues; call sites hold exclusive `&mut` to the connection, so the two-write path can't interleave.

https://claude.ai/code/session_01ND9WUgB4LmemR31SLpuXoF

---
_Generated by [Claude Code](https://claude.ai/code/session_01ND9WUgB4LmemR31SLpuXoF)_

### DIFF
--- a/binaries/daemon/src/socket_stream_utils.rs
+++ b/binaries/daemon/src/socket_stream_utils.rs
@@ -1,5 +1,13 @@
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
+/// Payload size (excluding the 8-byte header) at or below which
+/// [`socket_stream_send`] coalesces the header and body into one buffer.
+/// Chosen to cover normal control messages and small data outputs while
+/// keeping the extra copy trivial; larger payloads are written
+/// header-then-body without copying. Mirrors the node-side transport's
+/// `COALESCE_THRESHOLD`.
+const COALESCE_THRESHOLD: usize = 16 * 1024;
+
 pub async fn socket_stream_send(
     connection: &mut (impl AsyncWrite + Unpin),
     message: &[u8],
@@ -14,14 +22,31 @@ pub async fn socket_stream_send(
             ),
         ));
     }
-    // Concatenate length header + payload into a single buffer to avoid
-    // sending a tiny 8-byte TCP segment with TCP_NODELAY enabled.
-    // Previously used two separate write_all calls (2-3 syscalls per message).
     let len_raw = (message.len() as u64).to_le_bytes();
-    let mut buf = Vec::with_capacity(8 + message.len());
-    buf.extend_from_slice(&len_raw);
-    buf.extend_from_slice(message);
-    connection.write_all(&buf).await?;
+    // For a small message, concatenate the 8-byte length header and the payload
+    // into a single buffer and write it in one call. The connection has
+    // `TCP_NODELAY` enabled, so writing the header separately would flush it as
+    // its own tiny TCP segment before the body — an extra syscall and a runt
+    // segment on every message.
+    //
+    // Above the threshold we skip the coalescing copy and write the header and
+    // body separately: this is the daemon→node reply writer, and a
+    // `DaemonReply::NextEvents` carries data-plane payloads up to
+    // `MAX_MESSAGE_BYTES` (64 MiB) whenever an output takes the daemon path
+    // instead of direct zenoh. Copying tens of MiB just to avoid one runt
+    // header segment is the wrong trade — for a large body that header segment
+    // is negligible anyway. The wire format (8-byte LE length prefix + body) is
+    // identical on both paths. This mirrors the node-side transport
+    // (`apis/rust/node/src/daemon_connection/tcp.rs`).
+    if message.len() <= COALESCE_THRESHOLD {
+        let mut buf = Vec::with_capacity(8 + message.len());
+        buf.extend_from_slice(&len_raw);
+        buf.extend_from_slice(message);
+        connection.write_all(&buf).await?;
+    } else {
+        connection.write_all(&len_raw).await?;
+        connection.write_all(message).await?;
+    }
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


phil-opp:
## Automated codebase review — run summary

This PR is part of an automated review sweep of `dora` at `origin/main` (`70390f1`). Each finding was verified against current `main` before a PR was opened; all changes are behavior-preserving. Branches were cut from freshly-fetched `origin/main`.

### PRs opened (5), by category

**Optimization**
- #3265 — `perf(daemon)`: cap the coalescing copy in `socket_stream_send` so large daemon→node reply payloads (up to 64 MiB) aren't fully memcpy'd just to avoid a runt TCP header segment; mirrors the node-side 16 KiB threshold. *(this PR)*

**Documentation**
- #3266 — `docs(message)`: document the public `OperatorId` type (previously undocumented, unlike `NodeId`/`DataId`), with a compile-checked doctest.
- #3268 — `docs(node-api)`: remove a stale, misattributed doc comment that was wrongly prefixing `large_send_diag_count`'s docs.
- #3269 — `docs(node-api)`: correct the `Event::Reload.operator_id` doc, which claimed `None` never occurs though the runtime explicitly handles it.

**Simplification**
- #3267 — `refactor(node-api)`: drop dead `Result` wrapping in `data_to_arrow_array` (the wrapped conversion over the single-variant `DataMessage` is infallible).

**Correctness**
- None this run — the daemon and core/message crates are unusually well-defended (nearly every non-trivial path carries an issue-referencing regression test), and no high-confidence correctness bug survived verification.

### Rejected / skipped findings (one line each)

- **Duplicated circuit-breaker recovery block** (`daemon/src/lib.rs` ~7016 & ~7124) — the two sites' gating contexts differ subtly; not a safe blind merge.
- **`log.rs` `lock().unwrap()` poison panic** — only in a `#[cfg(test)]` test; both production lock sites already recover from poisoning.
- **ROS2 type format rejects 3-segment names** (`validate_ros2_type_format`) — consistent with the 2-part convention used throughout docs/examples; not a bug.
- **Node-exit generation-guard asymmetry** (`lib.rs:6277` vs `lib.rs:2122`) — deliberate and comment-justified; no concrete stranding race could be constructed.
- **Adding validation to `OperatorId`** — risks rejecting ids existing dataflows already use (behavior change); documented the current behavior instead (#3266).
- **Sharing `COALESCE_THRESHOLD` as one const** (raised during review) — would add a public const to `dora-message` and widen #3265's blast radius for marginal benefit; framing is identical regardless of the value, so kept as two locally-documented consts.

### Validation notes

Per-branch: `cargo fmt`, `clippy -D warnings`, and tests for the affected crate. On the combined diff: full-workspace `cargo check` (clean) + `cargo check --examples` (clean), affected-crate test suites (dora-daemon 247, dora-node-api 182 lib, dora-message doctests — all green), and the `/code-review` + `/simplify` gates (no correctness issues). Running the entire workspace test build *at once* exceeded this sandbox's fixed disk allowance, so it was split into the affected-crate suites plus the full-workspace check. Two `zenoh_schema_cache` integration tests fail identically on unmodified `origin/main` here (they need zenoh multicast/liveliness networking unavailable in the sandbox) and are unrelated to any change in this sweep.

---

phil-opp:

No issues found. I focused on the cap boundary and framing. The change is a faithful mirror of the node-side transport in `apis/rust/node/src/daemon_connection/tcp.rs` — same `COALESCE_THRESHOLD` value (16 KiB), same inclusive `<=` comparison, and the same header-then-body two-write structure above the threshold. Specifically verified:

- The two-write path writes the 8-byte LE length header then the body, and `socket_stream_receive_with_header_timeout` reads an 8-byte header then a body of exactly that length, so framing is byte-identical to the coalesced path — no dropped/duplicated bytes and no off-by-one at the boundary.
- `write_all` handles partial writes internally, the two calls are sequentially awaited so ordering is guaranteed, and the call sites hold an exclusive `&mut` to the connection, so there's no interleaving on the split path.
- `MAX_MESSAGE_BYTES` and the existing framing/timeout tests are untouched.

The added test exercises the new above-threshold branch (payload > threshold, non-constant fill) and asserts a byte-identical round-trip, so it's a real test of the split path. One optional thought: a case pinned at exactly `COALESCE_THRESHOLD` would document the inclusive boundary, though both branches frame identically so it's purely cosmetic.

---

PR #3273, opened 2026-08-22
fix(tensor-pool): add missing imports in tensor_info.py

### LINKED ISSUE #3270: tensor-pool: `import dora_tensor_pool` fails with NameError — `tensor_info.py` has no imports (regression from #3249)
## Summary

The `dora_tensor_pool` Python package is **completely unimportable**. `import dora_tensor_pool` raises `NameError: name 'torch' is not defined` before any function is called, so every consumer that the #3249 refactor rewired to the new package is broken.

- **Introduced by:** commit f9ba98b — `refactor(python)!: move the CUDA helpers out of `dora` into the `dora_tensor_pool` extension` (#3249)
- **File:** `libraries/extensions/tensor-pool/python/dora_tensor_pool/tensor_info.py`

## Root cause

`tensor_info.py` contains **zero import statements**, yet references `torch`, `np` (numpy), `ctypes`, and `_CudaArrayInterface` at module scope and in function bodies:

- `torch` at **module scope** — `tensor_info.py:14` (`_DTYPE_MAP = {"<i8": torch.int64, ...}`) and `:43` (`_TORCH_TO_NUMPY_DTYPE_MAP`)
- `np` — the dtype map and the function bodies (`:44`–`:53`, `:133`, `:138`, `:142`, `:157`)
- `ctypes` — `:151` (`(ctypes.c_byte * size).from_address(ptr)`)
- `_CudaArrayInterface` — `:143` (the class lives in `cuda.py` and is not imported here)

The new `dora_tensor_pool/__init__.py` eagerly runs `from .tensor_info import get_tensor_info, tensor_from_info`, so merely importing the package executes `tensor_info.py`'s module body and dies on the first line that dereferences `torch`.

This is a classic "promoted parked code" regression: the file was previously `tensor_info_helpers.py`, dead code that nothing imported, so its missing imports never mattered. #3249 renamed it to `tensor_info.py` and wired it into `__init__.py` **without adding the imports it always lacked**. `cuda.py` is fine — it imports `ctypes`, `numpy`, `pyarrow`, `torch` at its top; only `tensor_info.py` is broken.

There is also an internal contradiction worth noting: `cuda.py:34` deliberately resolves the CUDA runtime lazily so that "non-CUDA hosts (and macOS/Windows importers that never touch IPC) can `import dora_tensor_pool` without blowing up at module load." `tensor_info.py`'s module-scope `torch`/`np` dereferences defeat that goal on every platform — the import fails even where torch is installed, because the name is simply never bound in this module.

## Reproduction

```python
>>> import dora_tensor_pool
Traceback (most recent call last):
  ...
  File ".../dora_tensor_pool/tensor_info.py", line 14, in <module>
    "<i8": torch.int64,
NameError: name 'torch' is not defined
```

Confirmed by static analysis (`ast` reports 0 import statements) and by exec-ing the module body: it fails at line 14 regardless of whether torch/numpy are installed, because the names are never bound.

## Blast radius

Total — every rewired consumer fails at import time:

- `libraries/extensions/tensor-pool/examples/receiver.py`, `sender.py`
- `examples/cuda-benchmark/{sender,receiver,demo_sender,demo_receiver}.py`
- Documented usage in `docs/api-python.md`, `guide/src/languages/python.md`, and `libraries/extensions/tensor-pool/README.md`

A pure-CPU consumer that only wants `get_tensor_info` is broken too, since the failure is at package load, not at call.

## Why it shipped

No test anywhere imports `dora_tensor_pool` or references these helpers, and the example dataflows in this extension are not run in CI. A one-line `import dora_tensor_pool` smoke test would have caught it.

## Suggested fix

Add the missing imports at the top of `tensor_info.py`:

```python
import ctypes

import numpy as np
import torch

from .cuda import _CudaArrayInterface
```

And add a minimal import smoke test (e.g. `import dora_tensor_pool` + a `get_tensor_info` round-trip) so the package's import path is exercised in CI.

---

*This issue was filed by an automated Claude review job (scheduled task) that inspects recent commits/PRs for regressions. The finding above was reproduced directly against the current `main` (tip `70390f1`) before filing.*

phil-opp:
## Summary

`libraries/extensions/tensor-pool/python/dora_tensor_pool/tensor_info.py` referenced `torch`, `np`, `ctypes`, and `_CudaArrayInterface` at module scope and in function bodies but had **no** `import` statements. Because `dora_tensor_pool/__init__.py` re-exports `get_tensor_info` / `tensor_from_info` from that module, plain `import dora_tensor_pool` failed at module load time with `NameError: name 'torch' is not defined` (triggered by `_DTYPE_MAP` at module scope). Every consumer that #3249 rewired to the new package was broken.

## Fix

Add the four missing imports at the top of `tensor_info.py`:

- `import ctypes` — used by `tensor_from_info` (`ctypes.c_byte * size`)
- `import numpy as np` — used by `_TORCH_TO_NUMPY_DTYPE_MAP` and the size / shape / dtype validation
- `import torch` — used by both dtype maps and the tensor helpers
- `from .cuda import _CudaArrayInterface` — used in the CUDA branch of `tensor_from_info`. Kept internal to the package by importing the underscore-prefixed name directly from the sibling module (matches how `cuda.py` defines it).

No behavior change — this restores the module to a state where its own code can execute.

Closes #3270

---

### DIFF
--- a/libraries/extensions/tensor-pool/python/dora_tensor_pool/tensor_info.py
+++ b/libraries/extensions/tensor-pool/python/dora_tensor_pool/tensor_info.py
@@ -10,6 +10,13 @@ it. Reinstating these belongs on the dora-pool side of the seam, not in the
 `dora` package. See ../README.md.
 """
 
+import ctypes
+
+import numpy as np
+import torch
+
+from .cuda import _CudaArrayInterface
+
 _DTYPE_MAP = {
     "<i8": torch.int64,
     "<i4": torch.int32,


phil-opp:

No issues found. The four added imports all resolve to real uses in `tensor_info.py`:
- `torch` — `_DTYPE_MAP` / `_TORCH_TO_NUMPY_DTYPE_MAP` at module scope plus the tensor helpers
- `np` — `_TORCH_TO_NUMPY_DTYPE_MAP`, `np.dtype`, `np.frombuffer`
- `ctypes` — `ctypes.c_byte * size` in `tensor_from_info`
- `_CudaArrayInterface` — the CUDA branch of `tensor_from_info`; importing the underscore name directly from `.cuda` matches how `cuda.py` defines it

Since `_DTYPE_MAP` dereferences `torch` at module load, the missing imports really would have broken `import dora_tensor_pool` outright (via the `__init__.py` re-export), so the fix is on the mark. I checked the rest of the module too — no other unresolved names.

One unrelated observation while reading: `_ArrayInterface` (the CPU `__array_interface__` wrapper, ~line 62) appears to be defined but never used in this module — possibly worth removing in a follow-up, but it's outside the scope of this fix.

---

PR #3274, opened 2026-08-22
fix(ros2-bridge): reject null primitive/bool elements in array/sequence serialization

### LINKED ISSUE #3271: ros2-bridge: null element in a primitive/bool array or sequence field is silently encoded as 0/false (silent data corruption)
## Summary

In the ROS2-bridge Arrow→CDR serializer, a **null element inside a primitive or bool `array`/`sequence` field** is silently encoded as the underlying buffer default (`0` / `false`) instead of being rejected. This produces a structurally-valid ROS2 message that contains values the producer never sent — silent data corruption on the wire.

This is the primitive/bool analog of the string-element bug fixed in **#3254** (`reject null string elements in array/sequence serialization`). That fix covered `string[N]` / `string[]` / `wstring...`, but the primitive and bool array/sequence paths were left iterating `array.values()`, which ignores the Arrow validity bitmap.

## The invariant being violated

The serializer documents its own rule in `check_not_null` (`libraries/extensions/ros2-bridge/arrow/src/serialize/mod.rs:243-262`):

> ROS2 messages (and the CDR wire format) have no null concept, but `Array::value(0)` returns the physical buffer slot regardless of the validity bit — for a null it silently yields a default (`0`/`false`/`""`). Serializing that would invent a value the producer never sent, so reject it instead …

This invariant is enforced for:
- scalar primitive (`as_single_primitive`, `serialize/primitive.rs:87`) ✅
- scalar bool (`serialize/primitive.rs:61`) ✅
- scalar string/wstring (`serialize/mod.rs:151,157,175,181`) ✅
- `string[N]` / `wstring[N]` fixed arrays (`serialize/array.rs:327,344`, via `reject_null_string_element`) ✅ (#3254)
- `sequence<string>` / `sequence<wstring>` (`serialize/sequence.rs:219,234`) ✅ (#3254)

…but **not** for the numeric/bool array and sequence paths, which is the bug.

## Affected code (all iterate `array.values()` with no null check)

- `BasicArrayAsTuple::serialize` — fixed `int8[N]`…`float64[N]` — `serialize/array.rs:282`
- `BoolArrayAsTuple::serialize` — fixed `bool[N]` — `serialize/array.rs:307`
- `BasicSequence::serialize` — `int8[]`…`float64[]` — `serialize/sequence.rs:261`
- `BoolArray::serialize` — `bool[]` — `serialize/sequence.rs:343`

For example:

```rust
// serialize/array.rs — BasicArrayAsTuple::serialize
check_array_len(array.len(), self.len)?;
for value in array.values() {          // <-- .values() ignores the null bitmap
    seq.serialize_element(value)?;      //     a null slot serializes as 0
}
```

`array.values()` returns the raw physical value buffer, including the slots at null positions (which hold `0`). There is no `check_not_null` / `reject_null_*` equivalent here, and no way for the current code path to error on a null.

## Reproduction (verified)

Producer sends an `int32[3]` field as an Arrow `List<Int32>` with a null middle element (`[1, null, 3]` — the kind of value pyarrow readily produces). A temporary test added to `serialize/array.rs`:

```rust
let messages = fixed_array_message(NestableType::BasicType(BasicType::I32)); // int32[3]
let item = Arc::new(Field::new("item", DataType::Int32, true));
let list = ListArray::new(
    item.clone(),
    OffsetBuffer::from_lengths([3usize]),
    Arc::new(Int32Array::from(vec[Some(1), None, Some(3)])), // <-- null element
    None,
);
let struct_array = StructArray::from(vec[(
    Arc::new(Field::new("names", DataType::List(item), false)),
    Arc::new(list) as ArrayRef,
)]);
let value = Arc::new(struct_array) as ArrayRef;
let bytes = cdr_encoding::to_vec::<_, LittleEndian>(&TypedValue { value: &value, type_info: &type_info });
assert!(bytes.is_err(), "expected the null to be rejected");
```

Observed result — serialization **succeeds** and emits:

```
[1, 0, 0, 0,  0, 0, 0, 0,  3, 0, 0, 0]   // CDR i32[3] little-endian = [1, 0, 3]
```

The producer sent `[1, null, 3]`; the ROS2 peer receives `[1, 0, 3]`. No error is raised. The scalar path already rejects the equivalent single null (`null_scalar_field_is_rejected` test in `serialize/mod.rs`), so the array/sequence paths are inconsistent with both the scalar path and the string array/sequence paths.

## Suggested fix

Mirror the string fix: reject (rather than substitute) a null element in the primitive/bool array and sequence paths. Concretely, in `BasicArrayAsTuple`, `BoolArrayAsTuple`, `BasicSequence`, and `BoolArray`, return an error if `array.null_count() > 0` (or iterate with the validity mask and error on the first `None`), matching `reject_null_string_element`'s message/behavior. The `ByteSequence`/`Binary` byte-buffer paths are unaffected (a `Binary` value has no per-element validity). Regression tests should cover a null element in `int32[N]`, `int32[]`, `bool[N]`, and `bool[]`, alongside the existing string cases.

---

*This issue was filed automatically by a scheduled, unattended Claude Code review of the dora codebase (a periodic bug-hunting routine). The finding above was verified by compiling and running the reproduction against the current `main` (commit 70df... / branch tip at time of run); no human has reviewed it yet. Please treat the analysis as a starting point and verify before acting.*

phil-opp:
## Summary

In the ROS2-bridge Arrow→CDR serializer, a null element inside a primitive or bool `array`/`sequence` field was silently encoded as the buffer default (`0` / `false`) instead of being rejected — a valid CDR message containing values the producer never sent (silent data corruption).

This is the primitive/bool analog of #3254, which covered the string element path. The invariant is documented in `check_not_null` (`serialize/mod.rs`).

**Repro:** producer sends `int32[3] = [1, null, 3]` as a `List<Int32>` → CDR bytes `[1, 0, 3]` reach the ROS2 peer with no error.

**Affected code** (all iterated `array.values()`, which returns the raw value buffer and ignores the validity bitmap):

- `BasicArrayAsTuple::serialize` — fixed `int8[N]`…`float64[N]` (`serialize/array.rs`)
- `BoolArrayAsTuple::serialize` — fixed `bool[N]` (`serialize/array.rs`)
- `BasicSequence::serialize` — `int8[]`…`float64[]` (`serialize/sequence.rs`)
- `BoolArray::serialize` — `bool[]` (`serialize/sequence.rs`)

## Fix

- Add `reject_null_element(array, type_name)` in `serialize/mod.rs`. Short-circuits when `array.null_count() == 0`; otherwise reports the first null index in the same format as the existing `reject_null_string_element`.
- Call it from the four sites above, right after the existing `check_array_len` guard (fixed) or before allocating the seq (sequence).
- No change to the `ByteSequence` / `Binary` byte-buffer paths — a `Binary` value has no per-element validity.

## Tests

New regression tests, added alongside the existing `null_scalar_field_is_rejected` / `*_string_array_null_element_is_rejected` cases:

- `fixed_int32_array_null_element_is_rejected` (`int32[3]`)
- `fixed_bool_array_null_element_is_rejected` (`bool[3]`)
- `int32_sequence_null_element_is_rejected` (`int32[]`)
- `bool_sequence_null_element_is_rejected` (`bool[]`)

Each test asserts (a) a `[Some, None, Some]` list is rejected and (b) an all-present list still serializes. `cargo test -p dora-ros2-bridge-arrow --lib`: 28 passed, 0 failed. `cargo fmt --all -- --check` and `cargo clippy -p dora-ros2-bridge-arrow --all-targets -- -D warnings` are clean.

Closes #3271

---

### DIFF
--- a/libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs
+++ b/libraries/extensions/ros2-bridge/arrow/src/serialize/array.rs
@@ -12,7 +12,7 @@ use serde::ser::SerializeTuple;
 
 use crate::TypeInfo;
 
-use super::{TypedValue, check_array_len, error, reject_null_string_element};
+use super::{TypedValue, check_array_len, error, reject_null_element, reject_null_string_element};
 
 /// Serialize an array with known size as tuple.
 pub struct ArraySerializeWrapper<'a> {
@@ -278,6 +278,7 @@ where
             .as_primitive_opt()
             .ok_or_else(|| error(format!("not a primitive {} array", type_name::<T>())))?;
         check_array_len(array.len(), self.len)?;
+        reject_null_element(array, type_name::<T>())?;
 
         for value in array.values() {
             seq.serialize_element(value)?;
@@ -303,6 +304,7 @@ impl serde::Serialize for BoolArrayAsTuple<'_> {
             .as_boolean_opt()
             .ok_or_else(|| error("not a boolean array"))?;
         check_array_len(array.len(), self.len)?;
+        reject_null_element(array, "bool")?;
 
         for value in array.values() {
             seq.serialize_element(&value)?;
@@ -353,7 +355,8 @@ mod tests {
 
     use arrow::{
         array::{
-            ArrayRef, BinaryArray, Int32Array, ListArray, StringArray, StructArray, UInt8Array,
+            ArrayRef, BinaryArray, BooleanArray, Int32Array, ListArray, StringArray, StructArray,
+            UInt8Array,
         },
         buffer::OffsetBuffer,
         datatypes::{DataType, Field},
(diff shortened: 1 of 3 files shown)


phil-opp:

No issues found. I traced this against the diff and the surrounding serializer:

- The four guarded call sites (`BasicArrayAsTuple`, `BoolArrayAsTuple`, `BasicSequence`, `BoolArray`) are exactly the four `.values()` value-buffer iterations in `serialize/`, so there's no primitive/bool path left unguarded. The byte-buffer paths are rightly excluded, and the `ByteSequence` uint8 fallback picks up the guard for free via `BasicSequence<UInt8Type>`.
- `reject_null_element` is sound: `null_count() == 0` fast path, first-null index for the message, and the guard runs before any element is emitted (after `check_array_len` for fixed arrays, before `serialize_seq` for sequences), so a null fails cleanly with no partial CDR output.
- The tests exercise the real path through `cdr_encoding::to_vec` and assert both rejection-on-null and success-when-all-present, so they'd fail on the pre-fix code and also catch accidental over-rejection.

Nice symmetry with the #3254 string-element fix. Nothing blocking from my read.

---

PR #3275, opened 2026-08-22
fix(coordinator): prune dead daemon from running builds on DaemonExit

### LINKED ISSUE #3272: coordinator: DaemonExit path doesn't prune disconnected daemons from running builds, hanging multi-daemon `dora build` until the 20-min watchdog
## Summary

The coordinator has two daemon-disconnect paths, and they handle in-progress **builds** asymmetrically:

- **Heartbeat-timeout path** (`binaries/coordinator/src/lib.rs:2191-2222`) calls `cleanup_disconnected_daemons_from_running_builds`, pruning the dead daemon from every running build's `pending_build_results`.
- **`Event::DaemonExit` path** (`lib.rs:2370-2411`) calls `cleanup_disconnected_daemons_from_running_dataflows` + `apply_disconnect_actions`, but **never** calls `cleanup_disconnected_daemons_from_running_builds`.

`DaemonExit` is the common, immediate disconnect path — it fires as soon as the daemon's WebSocket closes (i.e. right when its process dies), well before the 30-second heartbeat-staleness threshold at `lib.rs:2175`. And because `DaemonExit` removes the daemon from `daemon_connections` (`lib.rs:2381`), the heartbeat watchdog — which only iterates `daemon_connections` — will *never* subsequently fire for that daemon. So the build-pruning call is simply skipped for the overwhelmingly common case.

## Why it hangs the build

A distributed build finalizes only when `pending_build_results` becomes empty (`lib.rs:2501`):

```rust
Event::DataflowBuildResult { build_id, daemon_id, result } => match running_builds.get_mut(&build_id) {
    Some(build) => {
        build.pending_build_results.remove(&daemon_id);
        // ...
        if build.pending_build_results.is_empty() {   // <-- only finalizes here
            // ... move to finished_builds, release wait_for_build
        }
    }
```

`apply_disconnect_actions` (`lib.rs:3710-3731`) only handles dataflow actions (`ReleaseReadyBarrier` / `ReclaimOrphaned`); it never touches `running_builds`. So on the `DaemonExit` path a dead daemon stays in `pending_build_results` indefinitely.

**Scenario:** a multi-daemon `dora build` spans daemons A and B. A's process exits (or its WS drops) after accepting the build RPC but before reporting its `DataflowBuildResult` → coordinator gets `Event::DaemonExit{A}`, which removes A from `daemon_connections` but leaves A in the build's `pending_build_results`. B then finishes and sends its result; the handler removes B, but A is still present, so `pending_build_results.is_empty()` is false and the build is never finalized. `wait_for_build` blocks until the `check_build_timeouts` watchdog fires — **`BUILD_RESULT_TIMEOUT_DEFAULT` = 20 minutes** (`lib.rs:116`) — and only then reports the build as *failed*.

On the heartbeat-timeout path the identical disconnect prunes A from `pending_build_results`, so B's later report finalizes the build promptly.

## Affected code

- `Event::DaemonExit` handler missing the build-cleanup call: `binaries/coordinator/src/lib.rs:2370-2411`
- The call it should mirror (heartbeat path): `lib.rs:2212-2215`
- Finalize-only-on-empty: `lib.rs:2501`
- `cleanup_disconnected_daemons_from_running_builds`: `lib.rs:3741-3750`

## Suggested fix

Add the same call the heartbeat path makes to the `DaemonExit` handler (inside the `connection_id` match arm, after the dataflow cleanup):

```rust
cleanup_disconnected_daemons_from_running_builds(&mut running_builds, &disconnected);
```

Note the design nuance to preserve: `cleanup_disconnected_daemons_from_running_builds` deliberately does **not** itself re-check `is_empty()` / finalize (see its doc comment at `lib.rs:3733-3740`, #1465 — the watchdog is meant to be the chokepoint for the all-daemons-gone case). That is fine here: in the mixed case above, the *surviving* daemon's later `DataflowBuildResult` performs the finalization once the dead daemon has been pruned; the all-gone case still falls to the watchdog. Worth confirming while fixing: after a mid-build daemon loss, whether the surviving-daemons-succeed outcome should report success or a "daemon vanished mid-build" failure — the pruning makes the build resolve, but the dead daemon's contribution is simply dropped.

## Verification

Confirmed by code inspection of `main` (the asymmetry, the finalize-only-on-empty condition, that `apply_disconnect_actions` never touches `running_builds`, and that `DaemonExit` removes the daemon from `daemon_connections` so the heartbeat path can't later catch it). Not reproduced at runtime — a multi-daemon build harness would be needed to observe the 20-minute hang end-to-end, so please validate before landing a fix.

---

*This issue was filed automatically by a scheduled, unattended Claude Code review of the dora codebase (a periodic bug-hunting routine). No human has reviewed it yet; please verify the analysis before acting.*

phil-opp:
## Summary

The coordinator has two daemon-disconnect paths, and they handled in-progress **builds** asymmetrically:

- **Heartbeat-timeout path** (`binaries/coordinator/src/lib.rs:2191-2222`) called `cleanup_disconnected_daemons_from_running_builds`, pruning the dead daemon from every running build's `pending_build_results`.
- **`Event::DaemonExit` path** (`lib.rs:2370-2411`) called `cleanup_disconnected_daemons_from_running_dataflows` + `apply_disconnect_actions`, but never called `cleanup_disconnected_daemons_from_running_builds`.

`DaemonExit` fires on the common WebSocket-close path (i.e. as soon as the daemon process dies), well before the 30-second heartbeat-staleness threshold at `lib.rs:2175` — and because `DaemonExit` removes the daemon from `daemon_connections` (`lib.rs:2381`), the heartbeat watchdog (which only iterates `daemon_connections`) never subsequently fires for that daemon. So the build-pruning call was simply skipped for the overwhelmingly common case.

## Why it hung the build

A distributed build only finalizes when `pending_build_results.is_empty()` (`lib.rs:2501`). Scenario:

1. Multi-daemon `dora build` spans daemons A and B.
2. A's process exits after accepting the build RPC but before reporting its `DataflowBuildResult`.
3. Coordinator gets `Event::DaemonExit{A}` → removes A from `daemon_connections` but leaves A in `pending_build_results`.
4. B finishes and sends its result; the handler removes B, but A is still present, so `pending_build_results.is_empty()` is false and the build is never finalized.
5. `wait_for_build` blocks until the `check_build_timeouts` watchdog fires — `BUILD_RESULT_TIMEOUT_DEFAULT` = 20 min (`lib.rs:116`) — and only then reports the build as failed.

On the heartbeat-timeout path the identical disconnect prunes A from `pending_build_results`, so B's later report finalizes the build promptly.

## Fix

Mirror the heartbeat path from the `DaemonExit` handler — one added call, positioned like the heartbeat-path call it mirrors (after the dataflow cleanup, before `notify_daemons_about_disconnected_peers`):

```rust
cleanup_disconnected_daemons_from_running_builds(
    &mut running_builds,
    &disconnected,
);
```

Finalization stays owned by the `DataflowBuildResult` handler / `check_build_timeouts` watchdog — the helper's docstring (`lib.rs:3733-3740`, #1465) intentionally leaves that as the single chokepoint. In the mixed case above, the surviving daemon's later `DataflowBuildResult` now performs the finalization once the dead daemon has been pruned; the all-gone case still falls to the watchdog.

## Tests

New regression test `cleanup_disconnected_daemons_from_running_builds_prunes_only_listed_daemons` locks in the pruning invariant the `DaemonExit` handler now relies on:

- The pruned daemon is gone from `pending_build_results`.
- Surviving daemons are untouched.
- `build_result` remains `Pending` — the cleanup helper must never itself finalize.

`cargo test -p dora-coordinator --lib` — 133 passed. `cargo fmt --all -- --check` clean; `cargo clippy -p dora-coordinator --lib --tests -- -D warnings` clean (a pre-existing `needless_late_init` warning under the newer Rust toolchain fires in `dora-tracing`, unrelated to this diff).

Closes #3272

---

### DIFF
--- a/binaries/coordinator/src/lib.rs
+++ b/binaries/coordinator/src/lib.rs
@@ -2396,6 +2396,25 @@ async fn start_inner(
                         &clock,
                     )
                     .await?;
+                    // Prune the exited daemon from every running build's
+                    // `pending_build_results`. Without this, an in-progress
+                    // multi-daemon `dora build` would sit with a dead
+                    // daemon in `pending_build_results` until the 20-minute
+                    // `check_build_timeouts` watchdog fires, even after
+                    // every surviving daemon has reported. The heartbeat-
+                    // timeout path (`Event::HeartbeatInterval`) already
+                    // does this at lib.rs:2212; `DaemonExit` fires on the
+                    // common WebSocket-close path well before the 30 s
+                    // heartbeat staleness threshold and removes the daemon
+                    // from `daemon_connections`, so the heartbeat path
+                    // will never subsequently catch it. Mirror the
+                    // heartbeat path here. Finalization is left to
+                    // `check_build_timeouts` / a surviving daemon's later
+                    // `DataflowBuildResult` — see the fn docstring (#3272).
+                    cleanup_disconnected_daemons_from_running_builds(
+                        &mut running_builds,
+                        &disconnected,
+                    );
                     notify_daemons_about_disconnected_peers(
                         &disconnected,
                         &mut daemon_connections,
@@ -10032,6 +10051,57 @@ mod tests {
             pending_restarts.len(),
             1,
             "original PendingRestart must not be overwritten by duplicate"
+        );
+    }
+
+    /// #3272: `cleanup_disconnected_daemons_from_running_builds` must prune
+    /// the given daemon(s) from every running build's `pending_build_results`
+    /// so a mixed disconnect (one daemon dies mid-build, others still to
+    /// report) can be finalized by the survivor's later `DataflowBuildResult`
+    /// rather than waiting for the 20-minute `check_build_timeouts` watchdog.
+    ///
+    /// This locks in the invariant the `DaemonExit` handler now relies on:
+    /// after the call, the pruned daemon's entry is gone but the build is
+    /// still present with its remaining `pending_build_results` (finalization
+    /// stays owned by the `DataflowBuildResult` handler / watchdog).
+    #[test]
+    fn cleanup_disconnected_daemons_from_running_builds_prunes_only_listed_daemons() {
+        let build_id = BuildId::generate();
+        let dead = DaemonId::new(Some("dead".to_string()));
+        let alive = DaemonId::new(Some("alive".to_string()));
+        let mut pending = BTreeSet::new();
+        pending.insert(dead.clone());
+        pending.insert(alive.clone());
+
+        let build = RunningBuild {
+            errors: Vec::new(),
+            build_result: CachedResult::default(),
+            buffered_log_messages: Vec::new(),
+            log_subscribers: Vec::new(),
+            pending_build_results: pending,
+            build_started_at: Instant::now(),
+        };
+        let mut running_builds: HashMap<BuildId, RunningBuild> = HashMap::new();
+        running_builds.insert(build_id, build);
+
+        let disconnected = BTreeSet::from([dead.clone()]);
+        cleanup_disconnected_daemons_from_running_builds(&mut running_builds, &disconnected);
+
+        let build = running_builds
+            .get(&build_id)
+            .expect("build must still be in running_builds — finalization is owned elsewhere");
+        assert!(
+            !build.pending_build_results.contains(&dead),
+            "dead daemon must be pruned from pending_build_results"
+        );
+        assert!(
+            build.pending_build_results.contains(&alive),
+            "surviving daemons must NOT be pruned"
+        );
+        assert!(
+            build.build_result.is_pending(),
+            "cleanup must NOT itself resolve build_result — the watchdog / \
+             surviving daemon's DataflowBuildResult owns finalization"
         );
     }
 }


phil-opp:
**Automated review by Claude** (fully automated — no human in the loop).

The fix itself looks correct and minimal — it mirrors the heartbeat path (same helper, same ordering: after `apply_disconnect_actions` and before `notify_daemons_about_disconnected_peers`, inside the connection-id guard), and the `DataflowBuildResult` handler (lib.rs:2491-2521) does finalize once `pending_build_results` is empty, so the mixed-disconnect scenario is genuinely resolved.

I do have one concern about the regression test. The behavioral change in this PR is the *added call* to `cleanup_disconnected_daemons_from_running_builds` inside the `Event::DaemonExit` handler (~lib.rs:2411). The new test calls that helper directly, but the helper is unchanged by this PR — it already existed and is already invoked from the heartbeat path at lib.rs:2212. So the test passes identically whether or not the fix is present: deleting the newly-added call in the `DaemonExit` handler would leave the test green. As written it locks in an invariant of an unchanged helper rather than guarding the actual fix or reproducing #3272.

Would it be feasible to drive this through the `DaemonExit` handler instead — e.g. a `start_inner`/event-loop-level test that feeds a `DaemonExit` mid-build and asserts the build's `pending_build_results` no longer contains the dead daemon? That would fail on the pre-fix code. The `DaemonExit` path currently has no test coverage at all, so it would also be net-new value.

---

phil-opp:
**Automated review by Claude** (fully automated — no human in the loop.)

One concern about the finalization semantics this change newly exposes on the common disconnect path.

`cleanup_disconnected_daemons_from_running_builds` (lib.rs:3745-3749) prunes the dead daemon from `pending_build_results` without recording anything in `build.errors`. In the exact mixed scenario this PR targets — daemon A dies mid-build, daemon B still to report — once A is pruned and B later reports `Ok`, the `DataflowBuildResult` handler finalizes with `build.errors.is_empty()` → `Ok(())` (lib.rs:2507-2511). So `dora build` returns **success** even though daemon A vanished before finishing its portion of the build, and the control client is never told a daemon was lost (`notify_daemons_about_disconnected_peers` only notifies daemons, not the CLI).

Before this change, that scenario stayed in `pending_build_results` and was eventually surfaced as an explicit failure by `check_build_timeouts` (`"… N daemon(s) never reported build_result"`). This PR moves a mid-build daemon death on `DaemonExit` — the common WebSocket-close path — from "reported as a build failure" to "silently reported as a passing build."

I realize the false-success behavior already exists on the heartbeat-timeout path (the helper is shared, #1465), so the PR doesn't introduce it from scratch — but it does make it the common case. Is that the intended semantics, or should a mid-build disconnect push an error into the affected running builds so the build finalizes as `Failed`? Happy to hear this is acceptable if a subsequent `dora start` is expected to catch the missing nodes.

---

phil-opp:
🤖 **Fully automated review by Claude** — generated automatically, not vetted by a human.

I found one issue: the fix closes only one of the two disconnect orderings; the reverse ordering still hangs for ~20 min and then reports a spurious *failure*.

`cleanup_disconnected_daemons_from_running_builds` only removes the dead daemon from `pending_build_results` and deliberately never finalizes (per its own docstring). A build is finalized in only two places: the `DataflowBuildResult` handler (on an incoming daemon result, when `pending_build_results.is_empty()`) and the 20‑min `check_build_timeouts` watchdog. The `DaemonExit` handler calls the cleanup helper but does not re-check completeness afterward.

So for daemons A (dies) + B (survives), in the ordering where **B reports first and A is the last pending entry**:
- B's `DataflowBuildResult` leaves `pending = {A}` (not empty → no finalize).
- A's `DaemonExit` prunes A → `pending = {}`, but nothing re-checks emptiness, so the build stays in `running_builds` with `build_result` still Pending.
- No further `DataflowBuildResult` can arrive, so `wait_for_build` blocks until the watchdog fires, which then sets `Err` with the self-contradictory message `"build timed out after 1200s; 0 daemon(s) never reported build_result"` — a spurious failure even though every surviving daemon reported `Ok`.

The PR body notes that the all-gone case still falls to the watchdog, so the delay is known — but worth flagging that it degrades to a false *failure*, not just latency. The complete fix looks like re-checking `pending_build_results.is_empty()` right after the prune in the `DaemonExit` handler and finalizing immediately with the surviving daemons' results, mirroring the `DataflowBuildResult` path.

This isn't a regression (the same ordering hangs on `main`), but the PR's stated goal — avoid waiting on the 20‑min watchdog when a daemon dies mid-build — is only partially met.

---

phil-opp:
closed in favor of #3342

PR #3276, opened 2026-08-23
fix(operator-api): surface stream Event::Error to Rust operators

phil-opp:
> ⚠️ **Machine-generated PR.** This change was produced by Claude (Claude Code) during an automated code-review pass. Please review carefully before merging.

## Issue

The shared-library operator runtime translates each node-stream `Event` into an FFI `RawEvent`, and it already builds `RawEvent { error: Some(..) }` for an `Event::Error`:

```rust
// binaries/runtime-shared-lib/src/runner.rs:253
Event::Error(err) => dora_operator_api_types::RawEvent {
    error: Some(err.into()),
    input_closed: None, input: None, stop: false,
},
```

But the FFI dispatch on the operator side (`apis/rust/operator/src/raw.rs::dora_on_event`) only inspected `input`, `input_closed`, and `stop`. A `RawEvent` carrying only `error` fell through to the *"ignore unknown events"* arm and returned `Continue` with no error — so stream-level failures (payload-decode errors, unrecognized daemon events, fatal stream errors — see `dora_node_api::event_stream`) were dropped **silently**: the operator's `on_event` was never called and nothing was logged. The populated `RawEvent::error` field had no consumer.

## Fix

Add an `Event::Error { error }` variant to the operator-facing `Event` enum (which is `#[non_exhaustive]`, so this is backward compatible) and dispatch `RawEvent::error` to it, mirroring the existing `InputParseError` handling. The variant borrows `&str` from the `RawEvent`, so there's no added allocation on the runtime path.

Operators can now observe and react to stream errors; returning `Err` from `on_event` reports it as fatal, exactly as the runtime already intended when it populated the field. `Error` is kept deliberately distinct from `InputParseError` (whole-stream vs. tied to one input id), matching the upstream node-api `Event` model.

## Validation

- `cargo test -p dora-operator-api` — all 9 lib tests + doctest pass, including two new regression tests (`stream_error_dispatched_to_on_event`, `empty_raw_event_is_ignored`).
- `cargo clippy -p dora-operator-api -- -D warnings` — clean.
- Existing downstream consumers already carry catch-all arms, so nothing breaks: the C++ operator bridge (`apis/c++/operator/src/lib.rs`, whose comment already lists `Error`) and the `dora new` Rust template. The PyO3 operator crate matches on `dora_node_api::Event`, not this enum, so it is unaffected.
- Full-workspace `cargo fmt --all --check`, `cargo clippy --all -- -D warnings`, and `cargo check --examples` pass on the combined review branch.

https://claude.ai/code/session_01MMnr5LKUBWDnpw7VvsreRn

---
_Generated by [Claude Code](https://claude.ai/code/session_01MMnr5LKUBWDnpw7VvsreRn)_

### DIFF
--- a/apis/rust/operator/src/lib.rs
+++ b/apis/rust/operator/src/lib.rs
@@ -55,9 +55,10 @@ pub mod raw;
 /// An event delivered to an operator's [`DoraOperator::on_event`] callback.
 ///
 /// The dataflow runtime dispatches one `Event` per occurrence: a received
-/// input, a failed input decode, an input that will produce no more data, or a
-/// request to shut down. The enum is `#[non_exhaustive]`, so implementations
-/// must include a catch-all arm to stay forward-compatible with future variants.
+/// input, a failed input decode, an input that will produce no more data, a
+/// runtime error on the event stream, or a request to shut down. The enum is
+/// `#[non_exhaustive]`, so implementations must include a catch-all arm to stay
+/// forward-compatible with future variants.
 #[derive(Debug)]
 #[non_exhaustive]
 pub enum Event<'a> {
@@ -89,6 +90,19 @@ pub enum Event<'a> {
     /// The runtime requested a graceful shutdown. The operator should finish
     /// any pending work and return [`DoraStatus::Stop`].
     Stop,
+    /// The runtime reported an error on the event stream itself — for example a
+    /// payload that could not be deserialized, an unrecognized daemon event, or
+    /// a fatal stream failure.
+    ///
+    /// This is distinct from [`InputParseError`](Event::InputParseError), which
+    /// is tied to a specific input id; an `Error` concerns the stream as a
+    /// whole. The operator can log or surface it and decide how to proceed;
+    /// returning `Err` from `on_event` reports it as fatal and stops the
+    /// operator.
+    Error {
+        /// A human-readable description of the error.
+        error: &'a str,
+    },
 }
 
 /// Trait implemented by every dora operator.
@@ -211,6 +225,7 @@ mod tests {
     #[derive(Default)]
     struct RecordingOperator {
         parse_errors: Vec<(String, String)>,
+        stream_errors: Vec<String>,
     }
 
     impl DoraOperator for RecordingOperator {
@@ -219,8 +234,14 @@ mod tests {
             event: &Event,
             _output_sender: &mut DoraOutputSender,
         ) -> Result<DoraStatus, String> {
-            if let Event::InputParseError { id, error } = event {
-                self.parse_errors.push((id.to_string(), error.clone()));
+            match event {
+                Event::InputParseError { id, error } => {
+                    self.parse_errors.push((id.to_string(), error.clone()));
+                }
+                Event::Error { error } => {
+                    self.stream_errors.push(error.to_string());
+                }
+                _ => {}
             }
             Ok(DoraStatus::Continue)
         }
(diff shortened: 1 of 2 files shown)


phil-opp:

This is the summary for an automated repository review pass (Claude / Claude Code). All findings were verified against a freshly fetched `origin/main`; each branch is off `origin/main`, and per-branch `fmt` + `clippy` + affected-crate tests passed. The full non-ROS2 workspace test suite (982 tests across the affected crates + `dora-coordinator`), `cargo check --examples`, and the `/review` + `/simplify` gates were run once on the combined diff and are green. *(The literal `cargo test --all` could not run to completion in this environment: `dora-ros2-bridge` needs a ROS2 distro — `AMENT_PREFIX_PATH` — and the sandbox's per-session disk allowance can't hold the full-workspace test build. The earlier `--all` run compiled every non-ROS2 crate successfully before aborting at the ROS2-env-dependent tests.)*

### PRs opened

**Correctness**
- #3276 — `fix(operator-api)`: surface stream `Event::Error` to Rust operators. The runtime populates `RawEvent::error` for `Event::Error`, but the FFI dispatch never read it, so stream-level errors were silently dropped. Adds an `Event::Error` variant + dispatch (+ regression tests).

**Optimization**
- #3277 — `perf(daemon)`: keyed lookup in `send_output_closed_events` instead of an O(total-edges) full-map scan, matching the keyed-access pattern used by every other delivery site. Behavior-equivalent; 246 daemon lib tests unchanged.

- #3278 — `docs(core)`: document the descriptor public API surface (`DescriptorExt` + doctest, `NodeExt`/`NodeExt::kind`, `NodeKind`, `resolve_path`).
- #3279 — `docs(message)`: document the `SessionId` / `BuildId` identifier types (+ `SessionId` doctest).

### Findings considered and rejected (one-line reasons)

- **Dedupe `parse_byte_size` (dora-core) onto `ByteSize::from_str` (dora-message)** — intentional decoupling (`u64` vs `usize`, `is_ascii_alphabetic` vs `is_alphabetic`); heavy dedicated mutation-test coverage; delegating adds a cross-platform overflow subtlety. Low net value, non-trivial risk.
- **Documented `hub` / `path` mutual-exclusivity not enforced in `NodeExt::kind`** — low confidence; `hub:` is explicitly outside the 1.0 stability guarantee and the `dora build` desugar path wasn't fully traced to confirm the combination isn't rejected earlier.
- **Unbounded in-memory download in `dora-download` (`response.bytes()`, no size cap)** — real observation, but the fix is policy-laden (what maximum? risks breaking legitimate large artifacts such as model weights). Deferred rather than guessing a limit.
- **Executable bit (`0o700`) applied to every downloaded file** — low confidence and partly intentional per the existing code comment.
- **Operator "data already taken" returns an error with `DoraStatus::Continue`** — effectively unreachable with the current runtime (a fresh `data_array` is built per event); not worth a change.
- **`handle_dataflow_stop` re-runs `update_dataflow_status` per removed dynamic node** — low confidence; relies on subtle idempotency reasoning, and the observable transition only fires when `local_nodes` empties.

Several other subsystems (`arrow-convert`, `recording`, `replay-node`, `hub-client`, `ros2-bridge` serialize/deserialize, `telemetry`, `runtime-api`, coordinator internals) were reviewed and found clean — no qualifying findings.

---

phil-opp:

Reviewed the diff end to end — this looks correct and complete.

- `apis/rust/operator/src/raw.rs:93` — the new `else if let Some(error) = &event.error` branch is placed correctly (after `input`/`input_closed`/`stop`, before the catch-all), and it consumes exactly the field that `binaries/runtime-shared-lib/src/runner.rs:253` already populates for `Event::Error`. No other event variant's dispatch changes.
- Type-wise it checks out: `RawEvent::error` is `Option<safer_ffi::String>` (`apis/rust/operator/types/src/lib.rs:97`), and `&safer_ffi::String` derefs to the `&'a str` the new `Event::Error { error }` variant expects — the same coercion the adjacent `InputClosed { id }` arm already relies on, so no added allocation.
- The `#[non_exhaustive]` attribute means adding the variant is source-compatible for downstream matchers, which already carry catch-all arms.
- `stream_error_dispatched_to_on_event` is a real regression test: against the pre-fix dispatch an error-only `RawEvent` fell through to the "ignore unknown events" arm and returned `Continue` without calling `on_event`, so the asserted `stream_errors` vec would be empty and the test would fail.

No issues found.

---

PR #3278, opened 2026-08-23
docs(core): document the descriptor public API surface

phil-opp:
> ⚠️ **Machine-generated PR.** This change was produced by Claude (Claude Code) during an automated code-review pass. Please review carefully before merging.

## Issue

Several central, user-facing public items in `libraries/core/src/descriptor/mod.rs` had no doc comments, inconsistent with the rest of the crate (their siblings carry detailed docs):

- `DescriptorExt` — the extension trait that turns a parsed `Descriptor` into the resolved node map dora runs, validates it, or renders it as mermaid — had no trait-level doc, and `resolve_aliases_and_set_defaults`, `visualize_as_mermaid_with_boundaries`, `blocking_read`, `parse`, and `check` were undocumented.
- `NodeExt` / `NodeExt::kind` — the dispatch point enforcing the exactly-one-of `path`/`operators`/`operator`/`ros2`/`module` contract.
- `NodeKind` and its variants.
- `resolve_path` — undocumented, while its confined sibling `resolve_path_confined` is extensively documented.

## Fix

Documentation only — no behavior change:

- Trait-level doc on `DescriptorExt` plus a **compile-checked doctest** walking `Descriptor::parse` → `resolve_aliases_and_set_defaults`, and docs for each previously-undocumented method.
- Docs on `NodeExt` / `NodeExt::kind` describing the exactly-one-of contract and the unresolved-`hub:` rejection.
- Docs on `NodeKind` and all its variants.
- Doc on `resolve_path` describing the working-dir → uv env → system `$PATH` search order and the no-symlink-resolution / no-confinement contract, contrasting it with `resolve_path_confined`.

## Validation

- `cargo test --doc -p dora-core` — the new `DescriptorExt` doctest passes.
- All added intra-doc links resolve (verified via `cargo doc`; the only remaining rustdoc warnings are pre-existing and in an unrelated file, `topics.rs`, left untouched).
- `cargo fmt --all --check` and `cargo clippy --all -- -D warnings` pass on the combined review branch.

https://claude.ai/code/session_01MMnr5LKUBWDnpw7VvsreRn

---
_Generated by [Claude Code](https://claude.ai/code/session_01MMnr5LKUBWDnpw7VvsreRn)_

### DIFF
--- a/libraries/core/src/descriptor/mod.rs
+++ b/libraries/core/src/descriptor/mod.rs
@@ -51,8 +51,49 @@ pub use expand::{
     expand_modules_with_boundaries,
 };
 
+/// Operations on a parsed [`Descriptor`].
+///
+/// These live on an extension trait rather than inherent methods because
+/// [`Descriptor`] is defined in `dora-message` (the wire-format crate), while
+/// the parsing, validation, resolution, and visualization logic belongs to
+/// `dora-core`. Import the trait to turn a descriptor into the resolved node
+/// map dora runs, validate it, or render it as a mermaid graph.
+///
+/// # Example
+///
+/// ```
+/// use dora_core::descriptor::{Descriptor, DescriptorExt};
+///
+/// let yaml = b"
+/// nodes:
+///   - id: source
+///     path: ./source
+///     outputs: [data]
+///   - id: sink
+///     path: ./sink
+///     inputs:
+///       value: source/data
+/// ";
+///
+/// // Parse the YAML into a descriptor...
+/// let descriptor = Descriptor::parse(yaml.to_vec())?;
+/// assert_eq!(descriptor.nodes.len(), 2);
+///
+/// // ...then resolve aliases and fill in defaults to get the node map dora runs.
+/// let resolved = descriptor.resolve_aliases_and_set_defaults()?;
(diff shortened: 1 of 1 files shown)


phil-opp:

Docs-only change; I verified each claim against `libraries/core/src/descriptor/mod.rs` and they hold up (the `NodeExt::kind` contract, `NodeKind` variants, and the `DescriptorExt` method docs all match the implementations).

One small factual imprecision worth tightening:

- The `resolve_path` doc (the new doc comment above `libraries/core/src/descriptor/mod.rs:417`) lists the search as an ordered cascade: "1. under `working_dir`, 2. the `uv`-managed environment, when `uv` is on the host, 3. the ambient system `$PATH`." But in the implementation (`mod.rs:431-437`), `uv` and `$PATH` are mutually exclusive branches: `$PATH` (`which::which(&path)`) is only reached when `uv` is **not** installed. If `uv` is present but the binary isn't found in the uv env, resolution errors out — it does not fall back to `$PATH`. The numbered list reads as "try 3 if 2 fails," which isn't what happens. Consider rewording step 3 to something like "or, when `uv` is not present, the ambient system `$PATH`" to reflect the either/or behavior.

Everything else checks out.

---

PR #3281, opened 2026-08-23
fix(core): unify module node field validation on the whitelist (#3258)

### LINKED ISSUE #3258: Module node validation diverges between top-level and nested: `build:` is rejected at top level but accepted+propagated when nested; other per-node fields are silently dropped when nested
## Summary

Module nodes are validated by **two different, disagreeing validators** depending on nesting level, introduced by the interaction of #3070 (`7e551cd`, added the top-level whitelist) and #2911 (`a342bff`, added the nested blacklist + docs):

- **Top-level module nodes** go through the *whitelist* `classify::check_module` (`libraries/core/src/descriptor/classify.rs:398`), called from `expand_modules` at `libraries/core/src/descriptor/expand.rs:117`. Allowed = `SHARED_FIELDS` (`id, name, description, env, deploy`) + `MODULE_ALLOWED` (`module, inputs, params`).
- **Nested module nodes** (a module node *inside* a module file) go through the *blacklist* `validate_module_node_fields` (`expand.rs:566`), called from the expand recursion at `expand.rs:640` and from the linter at `expand.rs:257`. It only rejects `path, args, path_sha256, git, hub, branch, tag, rev, operators, operator, ros2`.

Because a whitelist and a blacklist are not complements, the same YAML field behaves differently at the two levels. This produces two concrete, user-visible defects.

## Defect 1 — `build:` on a module node: rejected at top level, accepted+propagated when nested, and the shipped docs contradict both

`build` is in neither `MODULE_ALLOWED` nor `SHARED_FIELDS`, so the top-level whitelist **rejects** it. This is pinned by the test at `classify.rs:656-664`:

```rust
// build: cargo build  on a module node
check_module(&module).expect_err("module build should be rejected");
// error contains "Module" and "build"
```

But a **nested** module node's `build:` passes the blacklist and is actively propagated to the nested leaf nodes (`expand.rs:802` `let accumulated_build = inner_node.build.clone();` → `expand.rs:826-833` `prepend_module_build_to_node(nested_node, outer_build)`).

Meanwhile the docs added in the very same PR (#2911) at `docs/modules.md:264` (and `Changelog.md`, and the `dora_message::descriptor::Node` rustdoc that feeds the JSON schema) state:

> `env`, `build`, `deploy`, and `params` are accepted -- they propagate into the module's inner nodes.

Of those four fields, `env`, `deploy` (both in `SHARED_FIELDS`) and `params` (in `MODULE_ALLOWED`) are indeed accepted at the top level — but **`build` is not**; it hard-errors. The #2911 commit message itself characterizes top-level `build` as *"dropped at the top level"*, which is also inaccurate — it is rejected, not silently dropped.

**Failure scenario:**
```yaml
nodes:
  - id: src
    path: src.py
    outputs: [val]
  - id: m
    module: mod.yml
    build: pip install foo    # documented as accepted + propagating
    inputs:
      data: src/val
```
`dora run` / `dora start` / `dora build` / `dora validate` / `dora expand` all fail at expansion with:
`node `m` has fields that are not allowed on Module nodes: build`
— even though the shipped documentation instructs users to write exactly this.

## Defect 2 — silent field-drop persists for nested module nodes (the class #3070 set out to eliminate)

Since nested module nodes are validated by a blacklist, a nested module node carrying per-node fields that are *not* in the blacklist — e.g. `outputs`, `output_types`, `cpu_affinity`, `restart_policy`, `send_stdout_as`, `max_restarts`, etc. — passes validation and is then **silently discarded** during expansion (`expand_module_node` never reads them). The equivalent field on a *top-level* module node is correctly rejected by the whitelist.

**Failure scenario:**
```yaml
# inside a module file, referencing another module:
nodes:
  - id: inner
    module: leaf.yml
    cpu_affinity: [0]    # silently dropped, no diagnostic
```
This is precisely the silent-drop bug #3070 was created to prevent — still reachable one nesting level down.

## Root cause

Two independent validators for the same node kind at different nesting levels. They must be reconciled to a single rule.

## Suggested fix

Route **both** levels through the same `classify::check_module` whitelist (replace the nested `validate_module_node_fields` calls at `expand.rs:257` and `expand.rs:640`, or make `validate_module_node_fields` delegate to the whitelist). Then decide `build`'s status once and make code + docs agree:

- If `build` on a module node *should* work (as the docs and the nested behavior imply), add `"build"` to `MODULE_ALLOWED` and propagate the top-level module node's own `build` the way nested ones are handled.
- If it should *not*, correct `docs/modules.md`, `Changelog.md`, and the `Node` rustdoc, and drop `build` from the nested propagation for consistency.

Given nested module nodes already accept + propagate `build`, adding it to the whitelist (making the top level consistent with the nested behavior and the docs) is the behavior-preserving choice.

---

_This issue was filed by an automated, scheduled Claude code-review check (reviewing merged commits from 2026-08-19 → 2026-08-21). The analysis was verified against the code and the crate's own tests, but a human should confirm the intended `build`-on-module-node semantics before implementing the fix._

phil-opp:
Fixes #3258.

## Problem

Module nodes were validated by **two disagreeing validators** depending on nesting level: top-level module nodes went through the `check_module` whitelist, while nested module nodes went through a separate blacklist (`validate_module_node_fields`) in `expand_module_node`. Because a whitelist and a blacklist are not complements, the same field behaved differently at the two levels, producing two defects — both contradicting the documented contract in `docs/modules.md` and the `Node::module` rustdoc.

**Defect 1 — `build:` on a module node.** Rejected at the top level (not in the whitelist), but accepted and propagated to inner nodes when nested. The shipped docs and rustdoc, however, state that `env`, `build`, `deploy`, and `params` *are* accepted and propagate into the module's inner nodes. A user following the docs (`build: pip install foo` on a module node) hit a hard error at expansion.

**Defect 2 — silent field drop, one level down.** Per-node runtime fields (`outputs`, `cpu_affinity`, `restart_policy`, ...) on a *nested* module node passed the blacklist and were then silently discarded during expansion — the exact silent-drop class the top-level whitelist was introduced to prevent. The equivalent field on a top-level module node was correctly rejected.

## Fix

Route **both** nesting levels through `classify::check_module` (the issue's recommended, docs-preserving direction):

- Add `build` to the module whitelist and propagate a top-level module node's own `build` to its expanded leaf nodes, mirroring the existing nested propagation in Phase 2 of `expand_module_node`.
- Validate nested module nodes (both the expansion recursion and the `check_module_file` linter) with the same whitelist, so disallowed fields are **rejected rather than silently dropped** at every nesting level.
- Remove the now-dead `validate_module_node_fields` blacklist.

Code, docs, rustdoc, and the regenerated JSON schema (`dora-schema.json`, `libraries/core/dora-schema.json`) now agree.

## Tests

- Updated the two tests that pinned the old top-level `build` rejection (`classify.rs` unit test and the `invalid-module-build.yml` descriptor-validation case). `build` on a module node is now asserted **accepted**; a genuinely-disallowed field (`outputs`) is asserted **rejected** instead.
- New regression tests: `expand_top_level_module_build_propagated` (defect 1) and `nested_module_node_rejects_disallowed_field` (defect 2), plus a `valid-module-build.yml` acceptance case.

## Validation

- `cargo test -p dora-core` and `-p dora-message` — pass
- `cargo clippy -p dora-core -p dora-message --lib --tests -- -D warnings` — clean
- `cargo build -p dora-cli` — builds

## Note for reviewers

The issue flags that the intended `build`-on-module-node semantics deserve a human confirmation. This PR implements the direction the issue recommends and that the **already-committed** `docs/modules.md` + `Node::module` rustdoc already promise (accept + propagate). The only artifacts reversed are the two internal tests that encoded the top-level rejection; if you instead want `build` rejected everywhere, the alternative is to correct the docs/rustdoc and drop the nested `build` propagation. Kept as a draft pending that call.

---
_Generated by [Claude Code](https://claude.ai/code/session_015Yh1uTnGAr3BCJjwp7s1sD)_


phil-opp:

Reviewed the full diff against the surrounding `expand.rs`/`classify.rs` context. The change is sound:

- **Single validation site, both levels.** Top-level and nested module nodes now both go through `classify::check_module` (via `expand_module_node`, and via `check_module_file_inner` for the `dora expand` linter path). The old `validate_module_node_fields` blacklist is fully removed with no dangling references, so the whitelist and blacklist can no longer disagree.
- **`build` propagation applies exactly once per level.** `expand_module_node` only consumes `module_file.build` and `node.env` internally; the module node's *own* `build` is applied by the caller — Phase 2 for nested nodes (`accumulated_build`) and the new top-level loop in `expand_modules_with_boundaries`. So a leaf coming from a nested module gets `outer && inner && leaf`, outer-first, with no double application. Adding `build` to `MODULE_ALLOWED` (alongside the already-shared `env`/`deploy`) makes the whitelist match the documented contract.
- **Both regression tests are load-bearing.** `expand_top_level_module_build_propagated` would panic on `.unwrap()` against pre-fix code (old top-level `check_module` rejected `build`), and `nested_module_node_rejects_disallowed_field` would panic on `.unwrap_err()` (old blacklist let `outputs` pass and silently dropped it). The `descriptor-validation` fixtures are well-formed and the whitelist check fires before the missing-input check, so `invalid-module-outputs.yml` reports `outputs` as asserted.

No correctness, robustness, or security concerns found. The only open question is the intended semantics of `build` on a module node — which you've already flagged for human confirmation and gated behind draft status; the shipped docs/rustdoc already promise accept-and-propagate, so the code now matches them.

---

phil-opp:
> 🤖 _This is a **fully automated** review by Claude (Claude Code). No human reviewed these comments before posting._

Nice consolidation onto a single validation site — the `build` propagation (ordering and once-per-level) and the two new regression tests look correct. One gap in the whitelist migration, though.

Removing `validate_module_node_fields` also drops the only rejection of `operators`, `operator`, and `ros2` on a module node. Those three are kind-discriminators that are intentionally **not** in `ALL_CHECKABLE_FIELDS` (see the exhaustiveness test in `classify.rs` that re-adds `["operators", "operator", "ros2", "module"]` by hand). `check_module` validates via `validate_against_whitelist`, which only iterates `ALL_CHECKABLE_FIELDS` — so it structurally cannot flag those three fields.

For ordinary nodes that exclusion is safe because `node.kind()` enforces "only one discriminator". But module nodes never go through `classify()`/`node.kind()` — `classify()` bails on a module node ("must be expanded before resolution"), so they are validated solely through `check_module` in the expansion path.

Net effect: a node with both `module:` and, say, `operator:` now passes validation, and the `operator:` block is silently discarded during module expansion — exactly the silent-drop class this PR sets out to eliminate, and it contradicts the updated docs/rustdoc/schema which still list `operators`/`operator`/`ros2` among the rejected fields. (`path`, `git`, `hub`, `branch`, `tag`, `rev`, `args`, `path_sha256` are still correctly rejected — they're in the checkable table; only the three discriminators leak.)

Suggested fix: add explicit `operators`/`operator`/`ros2` rejection to `check_module` (or run module nodes through the multi-discriminator arm of `node.kind()` first), plus a regression test with `module:` + `operator:` set.

---

phil-opp:
Good catch — confirmed and fixed in `8f87c7d`.

You're right that dropping `validate_module_node_fields` also dropped the rejection of the `operators`/`operator`/`ros2` discriminators, since those are outside `ALL_CHECKABLE_FIELDS` and a module node never reaches `node.kind()`. `check_module` now rejects them explicitly (before the whitelist pass), with a regression test (`check_module_rejects_conflicting_kind_discriminator`) covering `module:` combined with each of `operator:`, `operators:`, and `ros2:`. The `path`/`git`/`hub`/`branch`/`tag`/`rev`/`args`/`path_sha256` rejections keep coming from the whitelist as before, so the docs/rustdoc/schema list of rejected fields is accurate again.

`cargo test -p dora-core`, `fmt --check`, and `clippy -p dora-core --lib --tests -- -D warnings` all pass.

---

phil-opp:
🤖 **Automated review by Claude** — this review is fully automated (no human in the loop).

Reviewed the new commit `8f87c7d` since my last review — it addresses the discriminator gap I raised. `check_module` now explicitly rejects `operators`/`operator`/`ros2` up front (before the whitelist pass), so a `module:` node that also sets one of those is no longer silently accepted with the extra block dropped during expansion. The `path`/`git`/`hub`/`branch`/`tag`/`rev`/`args`/`path_sha256` rejections continue to come from the whitelist (they are in `ALL_CHECKABLE_FIELDS`), and `check_module_rejects_conflicting_kind_discriminator` covers `module:` combined with each of the three discriminators, so the docs/rustdoc/schema list of rejected fields is accurate again.

No new issues in the added commit.

---

